Issue #61, opened 2026-08-03
Go-public checklist


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D29**.

**Phase 5 — Launch**
**Blocked by:** #46, #51, #52, #59, #60, #90, #91, #92, #93, #94
**Blocks:** nothing

## What

The last ticket. #16 (protect `main`) is the other half of this phase and stays on the decisions map.

Exactly one thing in this whole effort cannot be tested before the repo is public: **that GitHub serves `install.sh` at `raw.githubusercontent.com/actana/control/main/install.sh`.** That URL 404s today, so the documented one-liner cannot run end to end. This is not a coverage gap — `install-sh.test.mjs` and the installer e2e both exercise the real script against a fixture release server — it is a single fact that only becomes checkable at launch. It belongs here and nowhere else.

## Done when

- [ ] The one-liner runs on a clean Linux machine, from the real `raw` URL, and produces a working Core — the first post-public act
- [ ] The same one-liner runs on an Apple Silicon Mac and produces a working Core via the LaunchAgent path (#90's mac-arm64 asset, present because its gated release leg was approved)
- [ ] `docker pull actana/panel:0.1.0` and `actana/core:0.1.0` both work from Docker Hub — the only registry (ADR 0018; GHCR is retired and its packages deleted)
- [ ] Both Docker Hub description pages render the rewritten text (T23) — this is the first time either has ever been published, since `DOCKERHUB_DESCRIPTION_TOKEN` was never set
- [ ] `deploy/docker-compose.yml` brings up a paired Panel + Core from the published images on a machine that has never seen the repo
- [ ] A Trivy scan of both published images matches the targets: Panel ~14, Core ~38 distinct, zero fixable CRITICAL/HIGH on either
- [ ] The `wayfinder:map` for this effort is closed with its outcomes recorded

---

<sub>Decisions are locked in ADR 0016. If this ticket needs one changed, amend the ADR in the same PR — do not decide it in a comment.</sub>

Issue #59, opened 2026-08-03
The docs tree: 81 files → 55 (README shape moved to #93)


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D42, D43, D45**.

**Phase 4 — Docs and deletions**
**Blocked by:** #33, #45, #58
**Blocks:** #60, #61

## What

The tree, grouped by what the reader came to do: `install/ operate/ reference/ architecture/ adr/ contribute/ agents/ research/`. Every file that exists today has a disposition in #15 §3 — follow it.

**Two ordering constraints, both already blocking this ticket:** the tree cannot land before the rename (fifteen paths encode the new vocabulary in their filenames, and `install/core.md`, `reference/session-status.md` and `architecture/vocabulary.md` are renames *of the concept*), and it cannot land before the Core Dockerfile and compose have paths (the README's one-command block and `install/docker.md` must name files that exist).

**Two fixed points:** `docs/adr/` does not move — nesting it under `architecture/` would cost a repo-wide citation sweep on top of the eleven-file rename already ordered for that directory, and it is the convention every ADR tool assumes. `docs/images/` is frozen (T23).

**The `CONTEXT-MAP.md` is not optional.** Deleting root `CONTEXT.md` without it does not make the glossary invisible — the `/domain-modeling` skill's discovery protocol **silently recreates an empty one** the next time any session resolves a term, so the repo grows a second competing glossary by itself with no PR to catch it.

**Highest-risk item in this ticket:** `docs/agents/domain.md` and `AGENTS.md` are what a coding harness reads *before it writes anything*. A file still teaching "Harness = the machine" will keep producing the old vocabulary in every future session and will look authoritative doing it. They land in the same PR as the vocabulary merge, not after it.

## Done when

- [ ] The tree matches #15 §1: 55 files, 11 at root and 44 under `docs/`
- [ ] `README.md` is ~58 lines and its "Try it in one command" block runs `docker compose -f deploy/docker-compose.yml up` successfully from a clean checkout
- [ ] `CONTEXT.md` + `docs/domain-model.md` merged into `docs/architecture/vocabulary.md`, keeping `domain-model.md`'s definitions and discarding `CONTEXT.md`'s — the domain model was right and `CONTEXT.md` is what lagged
[...]

Issue #52, opened 2026-08-03
Docker Hub descriptions: fold into release.yml and rewrite docs/images/core.md


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D33**.

**Phase 3 — CI and release**
**Blocked by:** #45, #49
**Blocks:** #61

## What

`dockerhub-description.yml` is deleted; its `sync()` body moves **verbatim** into a `descriptions` job gated on the publish jobs — including the JWT exchange, the 100-byte / 25000-byte guards, and the `curl`-not-a-third-party-action posture. That posture stays: this step handles a credential that can rewrite the public face of both images.

**Trigger changes deliberately.** Today it fires on any push touching `docs/images/**`; in `release.yml` it fires on the release, so the page always describes something that exists. `workflow_dispatch` covers fixing a typo without cutting a release.

`docs/images/core.md` is **rewritten, not edited**. Every load-bearing sentence is false under the new design: it leads with a "⚠️ This is not a production deployment" banner, says "there is no supported container deployment of a Core", and documents systemd as PID 1, `--privileged`, the host cgroup, baked-in linger and a hardcoded `--public-host core`.

Good news: `DOCKERHUB_DESCRIPTION_TOKEN` was never set, so the sync has always no-op'd. **There is no stale public page to correct** — the text is wrong only in the repo.

## Done when

- [ ] `descriptions` job in `release.yml`, `needs: [panel, core]`; `dockerhub-description.yml` deleted
- [ ] `docs/images/core.md` rewritten to describe the supported product: tini as PID 1, one required env var, one volume, `docker compose` lifecycle, Harnesses installed at runtime
- [ ] The short description in the sync call changes from "DEVELOPMENT FIXTURE — …" to something that describes the product
- [ ] The `ai.actana.image.role=development-fixture` label and the fixture-specific `org.opencontainers.image.description` are removed from `container-image.yml`
- [ ] `docs/images/panel.md`: the `0.49.0` example version, the "You also need a Core" fixture description, and `Harness` → `Core`
- [ ] `docs/images/**` stays at its path — a workflow and a test hardcode it (frozen, per D33/D42)

---

<sub>Decisions are locked in ADR 0016. If this ticket needs one changed, amend the ADR in the same PR — do not decide it in a comment.</sub>

Issue #51, opened 2026-08-03
housekeeping.yml: stale, weekly Core rebuild, weekly audit, vendor canary


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D10, D34, D37, D38**.

**Phase 3 — CI and release**
**Blocked by:** #38, #47, #49
**Blocks:** #61

## What

The third and last workflow. Four unrelated schedules, one file, because none of them is a check.

The weekly Core rebuild is what makes the digest pin honest: `apt-get upgrade` in T8's layer resolves `noble-security` at build time, so a rebuild on an unchanged base digest still collects every fix Canonical has shipped. Weekly rather than nightly — Canonical does not ship security updates nightly, and the Chainguard-style nightly rebuild is a property of a vendor's build farm, not something worth imitating here.

The Harness-install canary is deliberately non-hermetic (it runs four vendors' real installers), so no PR can cause it and no PR author can fix it. Gating a PR on it is how a team learns that red means nothing.

## Done when

- [ ] Weekly (Monday): rebuild and republish the Core image, running T9's Trivy gate — so a base that rots is caught within seven days whether or not anyone has touched the repo
- [ ] Weekly: `pnpm audit --audit-level high` over the **dev** tree, opening an issue on failure rather than failing a build
- [ ] Weekly: the Harness-install e2e as a vendor canary, opening an issue on failure
- [ ] `stale.yml` folded in and deleted
- [ ] Nine workflow files are now three, plus `container-image.yml` as a reusable workflow rather than a fourth entry point — verified by `ls .github/workflows`

---

<sub>Decisions are locked in ADR 0016. If this ticket needs one changed, amend the ADR in the same PR — do not decide it in a comment.</sub>

PR #66, opened 2026-08-04
feat(images): phase 2 — a Core image, a distroless Panel, and one reference compose

### LINKED ISSUE #37: deploy/core.Dockerfile: Ubuntu 24.04, digest-pinned, toolchain in, Harnesses out
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D5, D6, D7, D8, D12, D14, D15, D40**.

**Phase 2 — Images and container**
**Blocked by:** #33
**Blocks:** #38, #44, #45, #46, #47, #49, #60

## What

The Core image. **This is the file #9 deleted without replacing** — `deploy/dev/core.Dockerfile` is what `container-image.yml:194` currently builds the published `core` image from, so this ticket must land before T16 deletes that tree.

Target: **1248 → ~38 distinct CVEs, 751 → ~190 MB.** Say it accurately in the file's header comment: the base did change, but the *mechanism* is the package set — `build-essential` → `libc6-dev` → `linux-libc-dev` is 1200 of 1328 OS findings, and dropping that one package alone takes an unchanged `ubuntu:24.04` build from 1248 to 38.

Both halves of the base pin are load-bearing and do different jobs: the digest, because `24.04` is a *rolling* tag that moves on every point release; and `apt-get upgrade -y` **in the same `RUN`**, because apt resolves `noble-security` at build time, so a weekly rebuild on an unchanged digest still collects every fix Canonical has shipped. Without the upgrade the pin freezes the CVEs too and T22's cadence is theatre.

## Done when

- [ ] `deploy/core.Dockerfile` exists; `container-image.yml` and `scripts/__tests__/panel-image.test.mjs` build/assert against the new path
- [ ] `FROM ubuntu:24.04@sha256:<digest>`, tag in the comment, with `apt-get update && apt-get upgrade -y && apt-get install …` in one layer
- [ ] Package set exactly: `bash sudo ca-certificates curl git openssh-client build-essential python3 ripgrep jq less vim-tiny unzip lsof xz-utils tini` — `tini` is not optional (D14)
- [ ] Node 24 from `https://nodejs.org/dist/v${NODE_VERSION}/`, verified against that release's own `SHASUMS256.txt`. `ARG NODE_VERSION=24.18.1`
- [ ] A comment warns against "simplifying" to `apt-get install nodejs`: it would get Node 18 **and** move it onto Ubuntu's Pro-gated *universe* security stream in one move
- [ ] The stock `ubuntu:x:1000:1000` user is removed and `core` is created with **explicit `--uid 1000 --gid 1000`**, with the reason written beside it — `useradd` without them lands on 1001:100 (measured) and breaks every bind-mounted repo
- [ ] `ENTRYPOINT ["/usr/bin/tini", "--"]`, `CMD ["actana", "daemon"]` — daemon is PID 2, and this is baked, not `--init`, so a bare `docker run` is not the broken configuration
- [ ] Container-mode env baked: `ACTANA_CONTAINER=1`, `AC_CORE_REMOTE=1`, `AC_CORE_LINK_HOST=0.0.0.0`, `AC_USER_DATA_DIR`, `AC_APP_PATH`, `AC_CORE_MATERIAL_FILE`, `NPM_CONFIG_PREFIX=/home/core/.local`, `PATH`
- [ ] `EXPOSE` reads from the same `ARG` as `ACTANA_PORT` rather than hardcoding 8443
- [ ] `NOPASSWD` sudo for `core` only; no `user:`-override accommodation (D12)
- [ ] Harnesses are **not** baked — the image is ~190 MB, not ~1.4 GB
- [ ] **Measurement, recorded in the PR:** whether `apt-get upgrade` actually takes the OS surface from 21 distinct to ~6. 15 of the 21 are marked fixable so it should, but that is an inference from Trivy's `fixable` column, not a scan. One build and one `trivy image` against `docs/research/core-base-measure/`. It does not change the base pick; it decides whether T22's cadence justification is real

---

### LINKED ISSUE #38: .trivyignore for linux-libc-dev, and the fixable-CRITICAL/HIGH gate
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D7, D11, D37**.

**Phase 2 — Images and container**
**Blocked by:** #37
**Blocks:** #51

## What

The toolchain ships because that is what the product is: `npm install` on any project with a native addon invokes node-gyp, which needs `make`, `g++` **and** `python3`. This repo is the proof — it depends on `better-sqlite3` and `node-pty`, so a Core without the toolchain cannot `pnpm install` Actana Control itself, and the failure mode is a wall of node-gyp output rather than "install build-essential".

So the ~1200 findings are handled by suppression, and the file must **call it suppression**. A spec that launders it is worse than one that argues for it.

The gate fails on **fixable** CRITICAL/HIGH only. Unfixed findings do not fail — there is nothing to do about them, and failing on them teaches people to ignore the gate.

## Done when

- [ ] A checked-in `.trivyignore` with exactly one entry, for `linux-libc-dev`, and the justification written beside it: kernel headers, 1008 of 1015 files under `/usr/include`, no executable code, and the kernel a Core runs is the host's
- [ ] Nothing else is suppressed anywhere in the repo — verified, not assumed
- [ ] `ci` runs Trivy on the **built** Core image and fails on any fixable CRITICAL or HIGH; medium/low are reported, not gating
- [ ] The policy — three populations, three owners — is written into `docs/contribute/ci-cd.md`, which is where G3 assigns it

---

### LINKED ISSUE #39: Core container mode: the env contract and verb gating
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D13, D15, D16**.

**Phase 2 — Images and container**
**Blocked by:** #33
**Blocks:** #45, #46

## What

The image **is** the install: no versioned tree, no `current` symlink, no unit file, no `loginctl enable-linger`, no third `ActanaServiceManager`. A degraded no-systemd `setup` would be a second lifecycle path that nothing supervises.

Detected by the baked `ACTANA_CONTAINER=1` — **not** by sniffing `/.dockerenv`.

`ACTANA_PUBLIC_HOST` is required and the image never guesses. `choosePublicHost()` picks the first routable IPv4, which is reasonable on metal and a trap in a container: a bare `docker run` has a container-ID hostname, so a guessing default silently changes the cert SAN on every recreation.

**The metal path is untouched.** `install.sh` + `actana setup` + systemd remains the story for a real machine. This is a second distribution, not a replacement.

## Done when

- [ ] `setup`, `start`, `stop`, `restart`, `update`, `uninstall` refuse in container mode **and name the Docker equivalent** (`docker compose pull && up -d`, `docker restart`, `docker logs`, …)
- [ ] `status`, `token`, `token regenerate`, `harnesses`, `daemon` work unchanged
- [ ] `status`'s auto-start row reports the container's **restart policy**, not a unit that does not exist
- [ ] The Core refuses to boot without `ACTANA_PUBLIC_HOST` in container mode, with a message naming the variable
- [ ] `ACTANA_PORT` (default 8443) and `ACTANA_LABEL` (default: the public host) are the only other documented variables
- [ ] Nothing reads `/.dockerenv`
- [ ] The hardcoded `--public-host core` is gone — it is a compose fact now (T16)

---

### LINKED ISSUE #40: Core first-run: persist minted material, print a human-readable blob
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D17**.

**Phase 2 — Images and container**
**Blocked by:** #33
**Blocks:** #45, #46

## What

The daemon's existing fresh-material path mints and prints but **does not persist** — that was `actana setup`'s job, and in a container `actana setup` no longer runs (D13). Without this, every container restart re-mints and every paired Panel is locked out.

`persistMaterial()` already exists in the same package.

Pre-seeding material via env or a mounted secret is explicitly rejected: it would need a generator that runs outside a Core, and would put private keys in compose files. This closes that direction rather than leaving two precedence-ordered paths through identity resolution.

## Done when

- [ ] On boot with `AC_CORE_MATERIAL_FILE` set and the file absent: mint CA + server cert + client cert + bearer secret + `coreId`, **persist**, write `registration-blob.txt` beside it, print the blob once
- [ ] On boot with the file present: load, start, print nothing
- [ ] In container mode the blob prints human-readably, not behind the `@@AC_CORE_REGISTRATION_BLOB@@` sentinel
- [ ] `ACTANA_LABEL` feeds the `label` field that path currently hardcodes to `""`
- [ ] `docker compose restart` is a no-op for pairing; `docker compose down -v` is the only thing that unpairs — both asserted in the smoke (T17)
- [ ] `docker compose exec core actana token` re-reads the blob for the life of the volume

---

### LINKED ISSUE #41: Core: re-issue the server cert on a host change instead of re-minting
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D18**.

**Phase 2 — Images and container**
**Blocked by:** #33
**Blocks:** nothing

## What

Today `resolveMaterial()` re-mints **everything** when the recorded public host differs — new CA, new certs, new `coreId`, every paired Panel locked out. On metal that is a rare deliberate act. In a container a typo'd env var fires it.

Keep the CA key and cert, the bearer secret, the `coreId` and the Panel's client cert; sign a fresh server cert for the new SAN. `material.json` already persists `caKey`, so this is cheap.

Note a full re-mint never actually rescued the moved-Core case anyway — the Panel dials the endpoint stored from the *old* blob regardless, so re-pairing was required either way. Re-issuing is strictly better: reversible, and it keeps the Panel's pinned CA valid.

## Done when

- [ ] Changing `ACTANA_PUBLIC_HOST` on an existing volume re-issues only the server cert; `coreId`, CA and bearer secret are unchanged
- [ ] A Panel paired before the change still validates the Core against its pinned CA
- [ ] `actana token regenerate` still revokes a leaked blob — that path is untouched
- [ ] Regression test: the old re-mint-everything behaviour cannot come back silently

---

### LINKED ISSUE #42: Move the root Dockerfile to deploy/panel.Dockerfile
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D40**.

**Phase 2 — Images and container**
**Blocked by:** #33
**Blocks:** #43

## What

Pure move, done as its own change so the distroless rewrite (T14) is reviewable as a content diff rather than a rename plus a rewrite.

#6 item 7 wants a thin root, and #15's root skeleton does not list a `Dockerfile`. "Two container images" should not mean finding one at the repo root and one two directories down.

Everything that names the old path must move with it — these are the sites, measured: `scripts/__tests__/panel-image.test.mjs:20,47,241`, `scripts/smoke-panel-image.mjs`, `.github/workflows/container-image.yml`, `docs/images/panel.md`'s `docker build .` example.

## Done when

- [ ] `deploy/panel.Dockerfile` exists; no `Dockerfile` at the repo root
- [ ] `panel-image` job and `smoke-panel-image.mjs` build the new path; the image is byte-equivalent to before the move
- [ ] No grep hit for a root-`Dockerfile` path anywhere in workflows, tests, scripts or docs

---

### LINKED ISSUE #43: Panel image: distroless runtime
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D20, D21, D22, D23, D24, D25**.

**Phase 2 — Images and container**
**Blocked by:** #42
**Blocks:** #44, #47, #49, #60

## What

**192 → 14 CVEs on the built image**, measured: OS 174 → 12, library 18 → 2, zero CRITICAL and zero HIGH from the OS. The 12 remaining OS findings are `libc6` (11) and `zlib1g` (1) — nothing else in the image has a CVE because there is nothing else in the image. Size 633 → 545 MB; this is a CVE story, not a size story, and must not be sold as one.

Prototype and reproduction script exist on `wt-10-panel-distroless` (`c19cfd0`, `./prototype-distroless.sh`).

**Three ways this fails quietly if skimmed** — each was hit in the prototype:

1. `COPY --chown` on `/data`, **not** a staged `chown`. Measured: `COPY` recreates the destination directory as `root:root 0755` and discards the source's ownership *and* mode; ownership only survives for entries *inside* a copied tree. Docker seeds a fresh named volume from the image's content **and mode**, so getting this wrong gives every new deployment a `/data` the Panel cannot write to.
2. `HEALTHCHECK` needs the absolute `/nodejs/bin/node` — argv bypasses ENTRYPOINT and distroless's PATH has no `/nodejs/bin`. The naive `["node", …]` form reports UNHEALTHY even for a script that cannot fail.
3. `CMD` loses the `"node"` element — node **is** the ENTRYPOINT, so `["node", "bin/panel.mjs"]` makes node try to run a file literally named `node`.

## Done when

- [ ] Runtime stage is `gcr.io/distroless/nodejs24` pinned **by digest** — the repo carries 6,980 tags and not one contains a version number, so a digest is the only meaningful pin
- [ ] `USER 65532:65532`, **numeric**, on the default tag. Not the `:nonroot` tag: every variant already ships `nonroot:x:65532:65532`, and numeric is required because Kubernetes' `runAsNonRoot` cannot resolve a username
- [ ] `/data` created by `COPY --from=build --chown=65532:65532 /staged/data /data`
- [ ] `CMD ["bin/panel.mjs"]`; `HEALTHCHECK` with the absolute `/nodejs/bin/node`
- [ ] **The published image is confirmed to carry the HEALTHCHECK** — podman drops it silently without `--format docker`, so `docker/build-push-action` with multi-platform output must be verified, not assumed, or local builds look fine while the published image ships without one
- [ ] Build stage moved to `node:<ci-node-version>-trixie` so both stages are Debian 13. If that build fails, bookworm is acceptable **only** with a comment stating the asymmetry and its direction (forward-compatible; the reverse breaks)
- [ ] `better-sqlite3` loads and the Panel migrates a real database into the volume — `operator`, `panel_sessions`, `cores`, `core_secrets` — and `GET /api/healthz` returns ok
- [ ] The four now-false assertions in `panel-image.test.mjs` are **replaced with the new invariants**, not patched to green: digest-pinned distroless final stage, numeric USER, `COPY --chown` on /data, absolute-path healthcheck, `CMD` without a `node` element
- [ ] The two remaining library findings are named in the PR as the Panel's own production dependencies (T1's territory), not base-image problems

---

### LINKED ISSUE #44: Digest-pin both bases and teach Dependabot to bump them
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D5, D10, D20, D26**.

**Phase 2 — Images and container**
**Blocked by:** #37, #43
**Blocks:** nothing

## What

Both images are digest-pinned, and neither registry makes that easy in the same way:

- `ubuntu:24.04` is a rolling tag; Docker Hub retains digests indefinitely, so the digest is a real pin and Dependabot's `docker` ecosystem handles it.
- `gcr.io/distroless/nodejs24` has **no semver tags at all** — four mutable names plus opaque build SHAs. Whatever automation lands has to update a digest pin for a registry with no version to compare against, or the image silently drifts its Node patch version between builds.

Also record that `gcr.io` is a third registry in the build path (D26): #6 settles Docker Hub primary / GHCR mirror for *publishing*, but pulling the base from Google's registry is a separate availability and rate-limit dependency on every single build.

## Done when

- [ ] `.github/dependabot.yml` covers both Dockerfiles and actually opens a PR when either base digest moves — demonstrated, not configured-and-hoped
- [ ] `NODE_VERSION` in `deploy/core.Dockerfile` is bumped by the same flow (it is what clears the Node-attributed findings, including the only CRITICAL and all four HIGH)
- [ ] The `gcr.io` build-path dependency is written into `docs/contribute/ci-cd.md`

---

### LINKED ISSUE #45: One reference deploy/docker-compose.yml; delete deploy/dev/
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D40, D41**.

**Phase 2 — Images and container**
**Blocked by:** #37, #39, #40
**Blocks:** #46, #52, #59

## What

`deploy/dev/` goes entirely — `core.Dockerfile`, `core-provision.sh`, `core-provision.service`, its compose and its README. Every file exists to fake a systemd machine so the tarball has somewhere to install, which is exactly the design being killed. Keeping it leaves that image in the repo as the thing newcomers copy.

**#9's claim that nothing outside the tree references it is false** — there are eight external references, one of which is the release path (`container-image.yml:194,203`). T8 must have landed first so there is a Core Dockerfile to point at.

One compose file, not two: no TLS terminator, and none coming. Operators bring their own nginx / Traefik / edge. The Panel speaks plain HTTP by design (ADR 0010 — "the Panel itself never grows certificate code"), so the reference's job is to name the port to point a proxy at. The Caddy + Let's Encrypt service is dropped.

**This is the browser-facing edge only. The core-link's mTLS is untouched** (ADR 0002) — that material is minted by the Core, travels in the registration blob, and is never something an operator supplies or renews.

## Done when

- [ ] `deploy/docker-compose.yml` brings up a Panel and a Core that pair, from a clean checkout, in one command
- [ ] The Core publishes **no host ports** — only the Panel reaches it, over the compose network
- [ ] `ACTANA_PUBLIC_HOST=core` is a compose value in a file the operator edits, not baked into the image
- [ ] No `privileged`, no `cgroup: host`, no `tmpfs`, no `init:` — all four die with systemd, and tini is in the image
- [ ] `core-home:/home/core` is the only Core volume; the repos bind mount is present and commented as swappable for a named volume
- [ ] `deploy/dev/` is gone and all eight external references are repointed: `container-image.yml:174,179,194,203`, `panel-image.test.mjs:29,188`, `docs/images/core.md`, `README.md`, `CONTRIBUTING.md`, `docs/README.md`, `docs/ci-cd.md`
- [ ] A second Core is a copy of the service block with a different name, host and volume — verified once, so the "nothing is singleton" claim is tested rather than asserted

---

### LINKED ISSUE #46: Core image smoke: boot the image and pair a Panel with it
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D36**.

**Phase 2 — Images and container**
**Blocked by:** #37, #39, #40, #45
**Blocks:** #61

## What

Replaces `panel-e2e-core-in-a-box`, and it is a straight upgrade: that job needed `--privileged` and the host cgroup to boot a systemd fixture, and that fixture is deleted. This makes the same assertion against the bytes that actually ship.

**It does not replace the installer e2e** (T19) and must not be argued to. They share nothing but the daemon binary: different arrival (`docker pull` vs checksum-verified `curl | bash`), different PID 1, different service management, different install location, and lifecycle verbs that are deliberately degraded in the image.

## Done when

- [ ] Builds `deploy/core.Dockerfile`, boots it, and a Panel pairs with it end to end — no privileged container anywhere
- [ ] Asserts restart is a no-op for pairing, and `down -v` unpairs (T11)
- [ ] Asserts the refused verbs refuse and name the Docker equivalent (T10)
- [ ] Asserts `tini` is PID 1 and the daemon is PID 2
- [ ] Runs the Trivy gate from T9 in the same job
- [ ] `panel-e2e-core-in-a-box` and its fixture are deleted in the same PR

---

qcentic-dev:
## Summary

Phase 2 of [#23](https://github.com/actana/control/issues/23): Actana ships as **two container images** instead of a tarball you install onto a machine you own.

`deploy/core.Dockerfile` is new — a Core you `docker run`, on digest-pinned Ubuntu 24.04 with a sha256-verified Node 24 tarball, tini as PID 1, running as `core:1000`. The Panel's image moves out of the repo root to `deploy/panel.Dockerfile` and swaps its runtime for `gcr.io/distroless/nodejs24`. `deploy/dev/` — the systemd fixture that existed so a tarball had somewhere to install — is deleted, and one reference `deploy/docker-compose.yml` brings up a Panel and a Core that pair, in one command.

Underneath that: a CVE gate that runs on the built image and fails only on **fixable** CRITICAL/HIGH, an owner for every base pin, and a smoke test that boots the image that ships and pairs a real Panel with it — unprivileged, replacing the `--privileged` systemd leg it retires.

Spec is [ADR 0016 — The 0.1.0 shape](docs/adr/0016-the-0-1-0-shape.md). Seven of its clauses are amended in this PR rather than argued in comments; see "ADR amendments" below.

## Related issues

Closes #37
Closes #38
Closes #39
Closes #40
Closes #41
Closes #42
Closes #43
Closes #44
Closes #45
Closes #46

Part of #23. Not in this PR: #51 (`housekeeping.yml`), #52 (Docker Hub descriptions), #59 (the docs tree), #61 (go-public checklist).

## What landed, ticket by ticket

| # | What |
| --- | --- |
| #42 | The root `Dockerfile` becomes `deploy/panel.Dockerfile`. Pure move; every builder now names the file. |
| #37 | `deploy/core.Dockerfile` — Ubuntu 24.04 digest-pinned, Node 24 as a sha256-verified nodejs.org tarball, native-addon toolchain in, Harnesses out. |
| #39 | Container mode: `ACTANA_CONTAINER=1` baked in, never sniffed from `/.dockerenv`. The lifecycle verbs Docker owns refuse and name their Docker equivalent. |
| #40 | First run mints material, persists it to the volume, and prints a registration blob a human can copy. |
| #41 | A changed public host **re-issues** the server cert from the kept CA. It no longer re-mints the identity and locks every paired Panel out over a typo'd env var. |
| #43 | Panel runtime becomes distroless: 192 findings → 14, `USER 65532:65532`, `/data` seeded by `COPY --chown`. |
| #38 | The CVE gate: fixable CRITICAL/HIGH only, on the built image, before anything is pushed. |
| #44 | Every base pin has an owner — Dependabot for the two digests, `base-pins.yml` for `ARG NODE_VERSION`, which no Dependabot config can see. |
| #45 | One reference compose. `deploy/dev/` and `deploy/Caddyfile` deleted, all external references repointed. |
| #46 | `scripts/smoke-core-image.mjs` — boot the image, pair a Panel, restart, `down -v`. No privileged container anywhere. |

## Four things worth a reviewer's attention

**1. The CVE gate has two suppressions, and they are different in kind.** `.trivyignore.rego` suppresses `linux-libc-dev` (D7's clause — ~1200 kernel-header findings against a package the Core cannot not ship). The gate *also* scopes out `usr/local/lib/node_modules/npm/`, the system Node's own bundled npm, by path. That second one is not in the policy file on purpose, so the two never hide inside each other, and D7/D37 are amended to stop claiming the file is the only suppression in the repo.

It exists because every fixable CRITICAL/HIGH in the Core image is in that one directory and **no released npm clears them** — npm 12.0.2 still carries `brace-expansion` 5.0.7 against fixes at 5.0.8/5.0.9 and `ip-address` 10.2.0 against 10.3.1. The cost is stated in the code: a new finding in that tree is silently non-gating too. Findings there are printed on every run, so scoping out never means hiding.

**2. The allowlist is `.trivyignore.rego`, not `.trivyignore`.** Measured, not assumed: Trivy's `.trivyignore` and `.trivyignore.yaml` match on CVE **id** only. A file naming a *package* parses cleanly, filters 0 of ~1200, and warns about nothing. Because that failure is silent, `assertSuppressionApplied()` checks the suppression actually took effect rather than trusting a green gate.

**3. Re-issue is gated on the operator having declared the host.** #41's re-issue only fires when `AC_CORE_PUBLIC_HOST` is actually set. Without that gate a hand-run `actana daemon` would re-sign a working Core's SAN for the bind address standing in — `127.0.0.1` — and hand out a loopback pairing token. Material predating the `serverHost` record is re-signed silently rather than announced as a move.

**4. The two digest pins are deliberately different shapes.** `ubuntu:24.04@sha256:…` carries its tag; the distroless runtime is pinned **tagless**. Dependabot bumps from the tag when there is one and otherwise from `digest_of("latest")`, and `gcr.io/distroless/nodejs24` has 7,040 tags without a version number among them. Writing `:latest@sha256:…` would hunt for a tag newer than `latest` and find nothing. Please do not tidy one to match the other — `docs/ci-cd.md` § "The two digests are pinned in different shapes on purpose" has the argument, and `base-pins.test.mjs` holds it.

## The two remaining Panel library findings (#43's last criterion)

The distroless move takes the Panel image from 192 findings to 14: OS 174 → 12, library 18 → 2. The 12 OS findings are `libc6` (11) and `zlib1g` (1). The two library findings are **the Panel's own production dependencies, not base-image problems** — both live under `srv/panel/node_modules/`, which is this repository's `pnpm install` output and nothing the base contributes:

| Severity | ID | Package | Installed | Fixed in |
| HIGH | GHSA-r28c-9q8g-f849 | `postcss` | 8.5.15 | 8.5.18 |
| LOW | CVE-2026-49356 | `@babel/core` | 7.29.0 | 7.29.6 |

Those are exactly [#30](https://github.com/actana/control/issues/30)'s territory (T1, "Lockfile: upgrade postcss and @babel/core"), and #30 has since landed: this branch's lockfile carries `postcss@8.5.25` and `@babel/core@7.29.7`, both past their fixes. The numbers above are from the 2026-08-03 prototype scan, which predates that bump — a fresh scan of this branch should show 12, not 14. CI's scan on the built image is the measurement that counts.

## ADR amendments

Each is written into `docs/adr/0016-the-0-1-0-shape.md` in the same commit as the work that found it:

- **D6, D10** (#37) — the measured package set, and two of D10's three rebuild mechanisms are weaker than the clause claimed.
- **D7, D11, D37** (#38) — "nothing else is suppressed anywhere" is no longer true; the second, path-scoped suppression and its cost are named.
- **D8** (#44) — `24.18.1` is the value at the time of writing, not a decision. A weekly bot PR is no longer also an ADR amendment.
- **D18** (#41) — rewritten around re-issue, plus the two bounds review added: silent backfill for pre-D18 material, and no re-signing on a host the operator never declared.

## How was this tested?

Verified locally on the merged tree:

| Check | Result |
| --- | --- |
| `packages/core` typecheck | clean |
| `packages/shared` typecheck | clean |
| `packages/core` vitest | 592 tests / 31 files pass |
| root vitest (scripts) | 256 tests / 11 files pass |
| eslint | 0 errors (54 pre-existing warnings, none in files this PR touches) |

**Docker is unavailable on the development box, and every commit that depends on it says so.** These are CI's to confirm, and all of them run on this PR:

- both images actually build, and `better-sqlite3` dlopens under distroless' glibc
- `scripts/smoke-core-image.mjs` — boots unprivileged, `tini` at PID 1 and the daemon at PID 2 read out of `/proc`, refused verbs refuse, a real Panel pairs, `docker restart` keeps the pairing, `down -v` ends it
- `scripts/smoke-panel-image.mjs` — `/data` ownership, the absolute-path HEALTHCHECK, the migrated schema read back out of a running container
- the Trivy gate, on the built image, in the same job as the smoke and before any push

**`docker compose up` on the reference compose is not among them, and the earlier version of this description wrongly claimed it was.** No workflow runs `docker compose` at all; `smoke-core-image.mjs` drives `docker run` directly and never reads `deploy/docker-compose.yml`. So #45's "brings up a Panel and a Core that pair, from a clean checkout, in one command" is proven by the file's *shape* (`panel-image.test.mjs`) and not by its behaviour. It is also forward-dated: the compose pins `:latest`, which only `images-release.yml` publishes on a `v*` tag, so there is nothing to pull until 0.1.0. That leg lands with the release rather than here.

**Two red checks on this PR are red on `main` too, and neither is this branch's:**

- **`Dependency Audit`** — four dev-only highs, all `.>eslint-plugin-react>minimatch>brace-expansion`, identical to `main` at `88b03295`. #30's fourth done-condition requires them left alone and T18 owns them.
- **`E2E — Harness offers (systemd)`** — byte-identical to `main`'s: the Cursor leg fails with `an accepted offer did not leave cursor-agent on PATH`, which is #31's tracked remainder (`claude-code`, `codex` and `cursor-cli` declare no `homePathSuffixes`). Nothing here touches it.

**`Commit messages (commitlint)`** is red on three commits — `ee59c94`, `a031cfd`, `77d8679` — each on a wrapped body line whose continuation opens `<word>: `, which the parser reads as a trailer token and so starts the footer mid-body. Per the forward-only governance note on #65 there is no remedy worth rewriting nine merges for; the diagnosis is here so the third occurrence is cheaper than the second.

One thing no branch can demonstrate: **Dependabot only reads `.github/dependabot.yml` on the default branch**, so its first real PR against these pins is post-merge evidence. What is demonstrated is the next best thing — `pnpm bases:check` resolves all three pins live, and with each pin staled on purpose it names the exact digest `digest_of(<tag>)` returns for both registries, which is the same call Dependabot makes. `docs/ci-cd.md` says where to look for the real PR.

## Type of change

- [x] 📚 Documentation
- [x] 🔧 Build / CI / tooling

- **`deploy/dev/` is gone.** Anything driving `docker compose -f deploy/dev/docker-compose.yml` stops working; use `deploy/docker-compose.yml`. All eight in-repo references are repointed in the same commit.
- **The root `Dockerfile` is gone**, moved to `deploy/panel.Dockerfile`. A build that relied on the implied root path must name the file.
- **`panel-e2e-core-in-a-box` is deleted**, along with `startContainerCore` and `e2e-panel-smoke.mjs --core-tarball`. `scripts/smoke-core-image.mjs` replaces it against the bytes that ship. This does **not** replace the installer e2e and must not be argued to: different arrival, different PID 1, different service management, different install location. `systemd-container.mjs` stays for exactly that reason.
- **The Caddy + Let's Encrypt service is dropped.** The Panel speaks plain HTTP by design (ADR 0010) and operators bring their own edge; the reference compose's job is to name the port to point a proxy at. Two requirements on whatever goes in front: forward WebSocket upgrades, and set `X-Forwarded-Proto: https`. None of this touches the core-link's mTLS (ADR 0002).
- **A Core whose `AC_CORE_PUBLIC_HOST` changes no longer loses its identity.** This is a behaviour change in the safe direction — a paired Panel survives what used to lock it out. Revoking a leaked pairing token is still the deliberate act it was: `actana token regenerate`.

## Checklist

- [x] My PR title follows Conventional Commits and my branch name follows the branching convention
- [x] I updated documentation (`README.md`, `DEPLOY.md`, `CONTRIBUTING.md`, `docs/ci-cd.md`, `docs/images/*`) where relevant
- [x] Dependent changes have been merged

---

**Merge strategy:** the ten tickets are ten merge commits with their reasoning in the messages. Squashing collapses all of it into this title — please **merge**, as [#65](https://github.com/actana/control/pull/65) did for phase 1.


qcentic-adm:
## Review against #37–#46 — one blocker, one confirmed defect, one description correction

Branch name, PR title, ADR amendments-in-the-same-commit, the measured corrections to D6 and D10 — all right, and the four "worth a reviewer's attention" notes are the four things I would have gone looking for. The `.trivyignore` → `.trivyignore.rego` finding in particular (matching on CVE **id** only, so a package-named entry filters 0 of ~1200 and warns about nothing) is the kind of silent failure that normally ships, and `assertSuppressionApplied()` is the right answer to it.

Eight of the ten tickets I can verify complete. Two I cannot, because the check that would prove them never got past its first leg.

---

### 🔴 Blocker — `Core image / Build + smoke (amd64)` is red, and it takes #46, #40, #38 and #37's acceptance with it

```
[core-image-smoke] booting the Core on a clean volume …
[core-image-smoke] FAIL: no pairing token at /home/core/.config/actana/registration-blob.txt
                         in actana-core-smoke-first-6243-msemilm7:
```

**`waitForCoreLink()` is not a readiness probe.** It connects to `127.0.0.1:${port}` — the *published* port — and Docker binds that on the host at container start, before anything inside is listening. `docker-proxy` completes the TCP handshake and the `connect` event fires, so the wait returns on its first iteration and the script `cat`s a file the daemon has not written yet.

Three things say so rather than one:

1. **497 ms** between `booting the Core on a clean volume …` (12:17:41.375) and the FAIL (12:17:41.872) — that is `docker volume create` + `docker run --detach` + one connect + one `docker exec`. Node, better-sqlite3 and a WS server do not boot in what is left.
2. **The failure message's `container.logs()` is empty.** A container that died would have printed why; `core-entry` writes its refusals and its `bootstrapCoreDb` failure to stderr, and `logs()` concatenates both streams. Empty means alive and not yet talking.
3. **The ordering makes the correct probe sufficient.** In `core-entry.ts`, `loadOrMintMaterial()` — which persists the material and writes the blob — runs *before* `PtyCoreLinkServer` listens. A probe that waited for the daemon actually listening would guarantee the blob is on disk by the time it returns. The race is only reachable because the current probe answers a question Docker already answered.

**Wait on the daemon's own signal, not on the port.** `core-entry` prints `@@AC_CORE_LISTENING@@` once the server is up and it is already in `docker logs`; polling for that, or exec'ing a check inside the container, closes it. Please don't paper it over with a sleep — the same probe is used for the restart leg and the second-boot leg, so it is silently vacuous three times, not once.

**The part that worries me more than the smoke:** the CVE gate step sits *after* the smoke in `container-image.yml`, so it never ran —

```
No files were found with the provided path: /home/runner/work/_temp/trivy-core.json
```

So on this PR the D11 gate has **never executed**. Not the 46-distinct figure, not `assertSuppressionApplied()`, not "no fixable CRITICAL/HIGH outside the npm tree". #38's done-condition is *"`ci` runs Trivy on the built Core image and fails on any fixable CRITICAL or HIGH"*, and #46's is *"Runs the Trivy gate from T9 in the same job"*. Both are unobserved.

That matters here specifically because #38 **widened** the suppression. `UNGATED_PATH_PREFIXES` scopes out `usr/local/lib/node_modules/npm/`, and by the module's own measurement *all seven* fixable CRITICAL/HIGH in the image live in that one directory — i.e. the scope-out currently silences the entire gating population. I think the argument is right (no released npm clears them; a permanently-red gate is a gate people click past), the cost is stated honestly in the code, and amending D7/D11/D37 in the same PR is exactly the route the ticket's footer prescribes. But a suppression that broad should not merge on an argument alone when the run that would measure it is one green check away.

Everything else in the Core image is verified: the build step itself passed, the `deploy/core.Dockerfile` package set, single-`RUN` `apt-get upgrade`, sha256-verified Node tarball, `userdel ubuntu` + explicit `--uid 1000 --gid 1000`, `NOPASSWD` sudo for `core` only, no baked Harnesses, `EXPOSE` from the same `ARG`, and `ENTRYPOINT ["/usr/bin/tini","--"]` are all present and all argued in place. The measurement #37 asked for is in `docs/research/core-base-measure/`, and correcting ~38→46 and ~190 MB→805 MB in the ADR rather than quietly restating the target is the right call.

### 🔴 Defect — `ACTANA_LABEL`'s documented default is computed and then dropped

D15, the Dockerfile, `actana-container.ts`, the CLI help and the compose comment all say the label defaults to the public host. On the daemon path it defaults to the empty string.

- `readContainerContract()` returns `label: trimmed(env, CONTAINER_LABEL_ENV) ?? publicHost` — correct.
- `cmdDaemon()` passes `runDaemon({ AC_CORE_LINK_PORT, AC_CORE_PUBLIC_HOST })`. **`contract.label` is not among them**, and `runDaemon` only `Object.assign`s what it is given.
- `core-entry.ts:248` therefore falls back to its own `process.env.ACTANA_LABEL || ""`.

So with `ACTANA_PUBLIC_HOST=core1.example.com` and no `ACTANA_LABEL`, the first-boot blob an operator copies out of `docker compose logs` carries `label: ""`, while `docker compose exec core actana token` — which goes through `containerInstall()` and *does* use `contract.label` — carries `label: "core1.example.com"`. Two tokens for one Core, different labels, from the same documented contract.

Nothing catches it because nothing exercises the default: the reference compose sets `ACTANA_LABEL=core` and the smoke sets `ACTANA_LABEL=${name}`. This is #39's *"`ACTANA_LABEL` (default: the public host)"* and #40's *"`ACTANA_LABEL` feeds the `label` field that path currently hardcodes to `""`"* — the field is no longer hardcoded, but the default that replaced it is still `""`.

One line in `cmdDaemon`, plus a test that boots container mode with no `ACTANA_LABEL` and asserts the blob's label is the public host.

### 🟡 Description correction — CI does not run `docker compose up`

"How was this tested?" lists `docker compose up` on the reference compose under *"These are CI's to confirm, and all of them run on this PR."* It isn't and they don't: `git grep 'docker compose' .github/workflows` returns nothing. `smoke-core-image.mjs` drives `docker run` directly and never reads `deploy/docker-compose.yml`.

So #45's first done-condition — *"brings up a Panel and a Core that pair, from a clean checkout, in one command"* — is proven neither locally (no Docker on the dev box, stated) nor in CI. The compose file's *shape* is well tested by `panel-image.test.mjs` (no published Core ports, none of the four systemd leftovers, `core-home` the only Core volume, and the second-Core block parsed and asserted for #45's last condition). Its *behaviour* is untested.

Related and worth naming rather than fixing: the compose pins `ghcr.io/actana/panel:latest` and `ghcr.io/actana/core:latest`, and `:latest` is only published by `images-release.yml` on a `v*` tag. Until 0.1.0 ships there is no `:latest` to pull, so the one-command claim is forward-dated. `:edge` exists after the first push to `main`. Fine for where the repo is — but it means nobody can run this file today, which is part of why the condition is unverified.

Either wire a compose leg into the Core image job, or untick it and say plainly it lands with the release. I mind the claim more than the gap.

---

### ⚪ Not blockers — and I am explicitly not counting these against the PR

**`Commit messages (commitlint)`.** Per the governance note on #65, commit messages bind forward-only and there is no remedy worth a 25-commit rewrite across nine merges. Not my blocker. But the description doesn't diagnose it, so here it is — three commits, three body lines that open with a word and a colon, which the parser reads as a trailer token and so decides the footer starts mid-body:

```
ee59c94  line 30  that: it is absent under Podman and nerdctl, it answers "did some runtime
a031cfd  line 23  to: different arrival, different PID 1, different service management, different
77d8679  line 20  Also: `--timeout` went through `Number()` rather than `stringFlag`, so a
```

Same failure as `installed:` on #65 — a wrapped sentence whose continuation happens to start `<word>:`. That is twice in two PRs, so it is a trap in the tooling rather than carelessness: a `commit-msg` check that rejects a non-final body line matching `^[A-Za-z][A-Za-z0-9_-]*: ` would cost less than diagnosing it a third time, and a line in `.gitmessage` would cost less still. Two wrinkles for whoever decides: these commits were authored after the forward-only ruling, and the requested **merge** (not squash) puts these bodies in `main` permanently, which the squash policy would not have. That is mrfqcentic's call, not mine.

**`Dependency Audit`.** Red on `main` at `88b03295` too, identically: four dev-only highs, all `.>eslint-plugin-react>minimatch>brace-expansion`. #30's fourth done-condition requires them left alone and T18 owns them. Correct by design — but it is not named in the description this time, and it should be, for the same reason I asked on #65.

**`E2E — Harness offers (systemd)`.** Byte-identical to `main`'s: opencode passes, `--with-claude-code` passes, the Cursor leg fails with `an accepted offer did not leave cursor-agent on PATH`. That is #31's tracked remainder — `claude-code`, `codex` and `cursor-cli` still declare no `homePathSuffixes` — and nothing here touches it.

<sub>Unrelated, noticed while diffing against `main`: `main`'s `Unit Tests` is red on a flake — `core-link-bearer > exp clock-skew tolerance > accepts a bearer exp exactly now (boundary inclusive)`. A test asserting behaviour "exactly now" against a clock that moves between arrange and assert. Green on this branch. Worth its own ticket so nobody reads a red `main` as this PR's doing.</sub>

---

### ✅ Verified complete

| Ticket | Evidence |
| --- | --- |
| **#42** | No `Dockerfile` at the root (`git ls-tree`); `deploy/panel.Dockerfile` exists; the path lives once as `PANEL_DOCKERFILE` and the contract test asserts all three builders against it; `Panel image / Build + smoke` green |
| **#43** | `gcr.io/distroless/nodejs24@sha256:…` tagless by digest; `USER 65532:65532` numeric; `COPY --from=build --chown=65532:65532 /staged/data /data`; `HEALTHCHECK` on absolute `/nodejs/bin/node`; `CMD ["bin/panel.mjs"]` with no `node` element; build stage `node:24.15.0-trixie`, both stages Debian 13; and the published-manifest healthcheck is **confirmed per platform** in `container-image.yml` via `imagetools inspect`, which is the criterion most likely to have been assumed. All four now-false assertions replaced with the new invariants, not patched to green |
| **#44** | One `/deploy` docker entry reaching both Dockerfiles, with all three preconditions (name match, non-recursion, digest-pinned) asserted in `base-pins.test.mjs` rather than assumed; `base-pins.yml` owns `ARG NODE_VERSION`, which no Dependabot config can see; the `gcr.io` build-path dependency written into `docs/ci-cd.md`. "Demonstrated, not configured-and-hoped" is structurally impossible pre-merge — Dependabot reads the default branch — and `pnpm bases:check` resolving all three pins live against the same `digest_of(<tag>)` call is the right substitute, said plainly |
| **#45** | `deploy/dev/` and `deploy/Caddyfile` gone, with `existsSync(...) === false` asserted for both; all external references repointed; compose publishes no Core ports, carries none of `privileged`/`cgroup: host`/`tmpfs`/`init`, `core-home` is the only Core volume with the repos bind commented as swappable, `ACTANA_PUBLIC_HOST=core` is a compose fact; second-Core block extracted and asserted. *Behaviour* unverified — see above |
| **#39** | Seven verbs refuse and each names its Docker equivalent; `status`'s auto-start row reports the restart policy **and where to read it**, since no process inside can read `HostConfig` back; `ACTANA_CONTAINER=1` only, with a source-level guard test that fails if `dockerenv` appears in non-comment code; `--public-host core` gone. Everything but the label default |
| **#41** | `checkServerCertHost` → `reissueServerCert` keeps CA, bearer secret, `coreId` and the Panel's client cert; the blob file is rewritten so the documented `cat` yields a token for the new address; `unrecorded` backfills quietly rather than announcing a move nobody made; and `publicHostDeclared` stops a hand-run `actana daemon` re-signing a working Core's SAN for `127.0.0.1`. Eleven tests over the three branches, including "leaves the cert alone when the public host was never declared" |
| **#37** | The Dockerfile itself — all thirteen conditions present and argued; measurement recorded; ADR corrected where it was wrong. Acceptance blocked only on the smoke |
| **#40** | `loadOrMintMaterial` mints → persists → writes the blob beside the material → prints once; later boots load and print nothing; a corrupt file throws rather than re-minting, and the asymmetry is argued. Fully unit-tested. Its *integration* proof is the leg the smoke never reached — and the label default is wrong |

The distroless work and the re-issue change are the two strongest pieces here. The `COPY --chown` note in particular (that `COPY` recreates the destination as `root:root 0755` and Docker seeds a fresh volume from the image's mode) is the sort of thing that ships broken and is diagnosed six months later by someone else.

---

### To get this approved

1. **Make `waitForCoreLink` wait for the daemon**, not for a port Docker opened — and re-run so `Core image / Build + smoke` goes green *and the Trivy gate actually executes*. That single green check is what closes #46, #40's persistence leg, #38's gate and #37's acceptance.
2. **Pass `contract.label` through `cmdDaemon`**, with a test covering the no-`ACTANA_LABEL` boot.
3. **Fix the `docker compose up` claim** — wire the leg in, or untick it and say it lands with the release. Same for the `Dependency Audit` red, which belongs in the description.

Do 1 and 2 and I will approve. 3 is a description edit and I would rather it were right than fast.

Nothing here is a design disagreement — the shape of this phase is right, and it is the most carefully argued work in the repo so far. Phase 3 is not gated on my review; it is gated on that one check going green, which is where I would want it gated anyway.

qcentic-dev:
Thanks — the probe diagnosis was right on all three counts, and the 497 ms was the tell I should have read myself. All three items addressed in `55f3e49`.

### 1. `waitForCoreLink()` waits on the daemon, not on `docker-proxy`

It now polls `docker logs` for `@@AC_CORE_LISTENING@@` — the signal `core-entry` prints on the next tick after `PtyCoreLinkServer` is listening, which is after `loadOrMintMaterial()` has minted, persisted and written the blob. So the ordering you noted is what makes the probe sufficient: when the wait returns, the file `readBlob()` reads is on disk.

No sleep anywhere. Two details worth naming:

- **Boots are counted, not looked for.** `docker logs` survives a `docker restart`, so a "wait for the sentinel" that merely searched would return instantly on the restart leg from the *first* boot's line — the same vacuity in a new shape. `container.boots` carries how many sentinels this container has already announced and the wait ends only on the next one, so the restart leg and the second-boot leg each wait for their own.
- **A dead container fails immediately.** `docker inspect --format '{{.State.Running}}'` is checked each iteration, so a container that exited reports with its logs instead of burning the 120 s timeout first.

`net` is gone from the script entirely, and `panel-image.test.mjs` now asserts the smoke references `LISTENING_SENTINEL` and contains no `net.connect` — cheap, but this is exactly the drift that reads as green.

I have no Docker here, so the check is still CI's to make. Whether the Trivy gate then executes is the part I care about too: it sits after the smoke in the job on purpose (scan the bytes that passed), so a red smoke has been swallowing it.

### 2. `contract.label` goes across

`cmdDaemon()` now hands `ACTANA_LABEL: contract.label` to `runDaemon` alongside the two `AC_*` variables — the one contract value `core-entry` reads under its own name rather than a translation. It is handed over even when the operator set it, so there is one resolution point rather than two agreeing by luck.

Three tests, since nothing exercised either path: the default (no `ACTANA_LABEL` → the public host), the explicit value, and the existing contract test extended to assert the label alongside the port and host. `packages/core` is 592 / 31 green, root 256 / 11, eslint 0 errors, typecheck clean.

### 3. The `docker compose up` claim is gone

Untick and say so, as you'd expect — the leg lands with the release, not here. The description now says plainly that no workflow runs `docker compose` at all, that `smoke-core-image.mjs` drives `docker run` and never reads `deploy/docker-compose.yml`, and that #45's first condition is proven by the file's shape and not its behaviour. Your `:latest` point is in there too: nothing to pull until `images-release.yml` runs on a `v*` tag, so the one-command claim is forward-dated regardless of what CI does.

`Dependency Audit` is now named in the description with its `main` parity, and so is the systemd Harness-offers leg. The commitlint reds are diagnosed there as well — the three commits, and the `<word>: ` continuation-line trap behind all of them. I've left the history alone per the #65 ruling; a `commit-msg` check rejecting a non-final body line matching `^[A-Za-z][A-Za-z0-9_-]*: ` is worth its own ticket, and I'd rather it be a check than a `.gitmessage` line nobody reads.

One thing I'm not doing without your say-so: moving the CVE gate ahead of the smoke so the gate reports even when the boot fails. It would have surfaced the D11 measurement on the very run we spent this review discussing — but it also scans an image that has not been shown to boot, which is a different promise from the one #38 made. Happy to make that change if you want it.

Nothing else touched. Over to CI for the check that closes #46, #40's persistence leg, #38's gate and #37's acceptance.

mrfqcentic:
The commitlint failure is a **false positive** — all 3 failing commits trip `footer-leading-blank`, but not because of a real footer problem. commitlint's parser misreads a mid-sentence body line as a Conventional-Commits footer token on each of them:

- `feat(core): container mode` (ee59c94): body line `that: it is absent under Podman...` → `that:` parses as a `token: value` footer.
- `test(core-image): boot the image...` (a031cfd): body line `to: different arrival, different PID 1...` → `to:` parses as a footer token.
- `build(panel): tighten the distroless image...` (9ebe00c): body line `issue #43 asked for exactly that...` → GitHub-style `issue #N` parses as a footer reference.

Each follows body text with no blank line above, so `footer-leading-blank` fires.

**Short fix (reword, no config change):** make those lines not parse as footers. E.g.:
- reorder so the line doesn't begin with `that:`/`to:` (e.g. `…guards: it is absent…` isn't needed — just merge so the word isn't line-initial), and
- drop the `#` in `issue #43` → `issue 43`.

I verified all three pass once the offending token is neutralized, e.g. `git show -s --format=%B <sha> | sed 's/^that: /that- /'` etc. — the rest of the body logic is untouched. The commit bodies are otherwise good.

qcentic-adm:
## Re-review of `55f3e49` — all three addressed, and the fix worked. One assertion left, in the test rather than the product.

All three items are done, and the probe fix is better than what I asked for. Not approving yet, but what is left is one line in `smoke-core-image.mjs`, and nothing about the image is wrong.

### ✅ The probe fix works — and the boot counting is a catch I missed

The smoke now gets four legs further:

```
[core-image-smoke] booting the Core on a clean volume …
[core-image-smoke] the Core minted core_6df6668a5dae0162 and printed its pairing token
```

So the blob is minted, persisted, read, and the endpoint check, the "the notice is in `docker logs`" check and the `1000:1000` identity check all pass. That is #40's persistence path proven end to end against the shipped bytes for the first time.

`container.boots` is the part I did not think of and should have. I would have accepted a `logs.includes(SENTINEL)` and it would have been the same vacuity wearing a different hat — instant-true on the restart leg from the *first* boot's line. Counting sentinels and ending on the next one is right, `isRunning()` turning a dead container into an immediate failure with its logs is right, and the source-level guard against `net.connect` coming back is the thing that keeps it right in six months.

### ✅ `contract.label` goes across

`ACTANA_LABEL: contract.label` alongside the two `AC_*` variables, handed over even when the operator set it so there is one resolution point rather than two agreeing by luck — that is the right call and better than the minimum I asked for. Three tests where there were none.

### ✅ The description is fixed, and more honestly than I asked

The `docker compose up` claim is not just untucked, it is named as having been wrong, with the `:latest` forward-dating alongside it. `Dependency Audit` and the systemd Harness-offers leg are both there with their `main` parity, and the commitlint diagnosis is recorded for whoever hits it third. `55f3e49`'s own message is clean — still exactly three commitlint failures, no new one.

---

### 🔴 What is left — the PID assertion is stricter than D14, and stricter than the launcher can promise

```
[core-image-smoke] FAIL: docker exec … cat /proc/2/comm exited 1:
cat: /proc/2/comm: No such file or directory
```

`/proc/1/comm` read fine — the check before it passed, so **tini is PID 1 and the container is healthy**. What does not exist is PID 2.

**The number is not the invariant; the topology is.** D14 exists because node-pty forks a shell, the shell forks a Harness, and a Harness whose shell exited first reparents to PID 1 — so PID 1 must be something that reaps. That needs tini at PID 1 (asserted, passing) and the daemon to be tini's *child* rather than PID 1 itself. It does not need the daemon to hold the integer 2.

And the integer is not the launcher's to promise. `bin/actana` is `#!/bin/sh` and runs command substitutions before it ever reaches `exec` — `self=$(command -v -- "$self")`, `$(readlink "$self")`, `ACTANA_ROOT=$(cd -P "${self%/*}/.." && pwd)` — each forking a subshell that takes a PID and exits. Whatever the exact allocation is inside this image, the daemon's final PID is an artifact of how many subshells the launcher forks on the way to `exec`, which is a detail `core-tarball.mjs` can change without touching anything D14 cares about. `LAUNCHER_SCRIPT` gaining or losing one `$(…)` should not turn the Core image smoke red.

**Assert the property instead:** `/proc/1/comm` is `tini`, and there is a `node` whose **PPID is 1** — field 4 of `/proc/<pid>/stat`, or `ps -eo pid,ppid,comm` and pick it out. That is exactly "tini is PID 1 and it is the daemon's reaper", it survives any launcher change, and it fails loudly in the one case D14 is about (the daemon *being* PID 1).

Worth saying: this leg has never executed before — the old probe died two legs earlier — so this is a first observation, not a regression from `55f3e49`. While you are in there, `ps -eo pid,ppid,comm` in the failure message would have made this a five-second read instead of a log archaeology.

### On your question about the gate ordering — don't move it, and you don't have to

Your instinct is right and I would have pushed back if you had moved it: scanning bytes that have not been shown to boot is a weaker promise than the one #38 made, and the ordering is load-bearing documentation of that.

You can have the measurement without giving that up. `if: always() && inputs.image == 'core'` on the **Trivy** step — the same condition `Keep the scan` already carries one step below it. Then the gate reports on every run whatever the smoke did, while the smoke's failure still fails the job.

That cannot bless a non-booting image, and this run proves it rather than my asserting it: `Push the per-arch tags` carries only `if: inputs.push`, which implies `success()`, and with the smoke red it came out **skipped** on run `30910073132` (step 14). A failed smoke keeps the push skipped no matter what the scanner says. The only thing `always()` buys is that D11's number stops being hostage to an unrelated leg — which is the exact hostage situation that has now cost two runs.

Your call, and I would take it, but it is not a condition of approval.

### ⚪ `Unit Tests` went red — it is the flake, not your commit

Please don't chase this one:

```
packages/shared  FAIL  src/__tests__/core-link-bearer.test.ts:105
  core-link bearer > exp clock-skew tolerance > accepts a bearer exp exactly now (boundary inclusive)
  AssertionError: expected false to be true
```

`packages/shared` is 1 failed / 13 passed, the other 11 files are green, and it is the same test I footnoted as red on `main` at `88b03295`. It was green on this branch's previous run and red on this one — so it is now demonstrably intermittent rather than merely suspicious. A test asserting behaviour "exactly now" against a clock that moves between arrange and assert. Its own ticket, not this PR's.

`Dependency Audit` and `E2E — Harness offers (systemd)` are unchanged and at `main` parity. `Panel image / Build + smoke` is green.

---

### To get this approved

**One thing:** assert tini-at-PID-1 plus a daemon whose PPID is 1, instead of the literal PID 2 — and let the run through so `Core image / Build + smoke` goes green and the Trivy gate finally executes.

That single green check still closes #46, #40's persistence leg, #38's gate and #37's acceptance, exactly as it would have last run. Everything else on this PR I have now verified twice.

qcentic-dev:
Fixed in `9495e4a` — you're right, and the assertion was wrong in a way that would have stayed wrong: the number is an artifact of how many subshells `bin/actana` forks before it `exec`s, and `core-tarball.mjs` owns that.

### The PID leg now asserts the topology

`/proc/1/comm` is `tini`, and some `node` has **PPID 1**. That is D14's actual requirement — tini at PID 1 *and* the daemon as its child — and it still fails in the one case the clause exists for, a daemon that is itself PID 1 and reaps nothing.

The process table is read out of `/proc` by a shell loop rather than with `ps`: the image installs no `procps`, and a smoke that needs a package the image does not ship would be asserting against something other than the bytes that ship. `ppid` comes from `/proc/<pid>/stat` counted *after* the `)` that closes the comm field, since a comm containing a space or a bracket makes every count-from-the-left answer wrong — verified against a live `/proc` here, including a `(sd-pam)` row.

Both failure paths now print the whole table, `PID PPID COMMAND` in pid order, so the next one is the five-second read you asked for rather than log archaeology.

### D14 said "PID 2", so D14 is amended

The clause's title and body both claimed the number, and the number is measurably not 2 in the shipped image — so it is amended in the same commit as the work that found it, per the route the tickets prescribe. `docs/ci-cd.md` (twice) and `docs/images/core.md` carried the same claim and are corrected alongside. What replaces it is the topology, with the launcher argument written down so nobody re-tightens it to an integer later.

### Gate ordering — taken, as you framed it

`if: always() && inputs.image == 'core'` on the Trivy step. The step stays where it is, for exactly the reason you gave. A test pins both halves: the gate carries `always()`, the smoke still precedes it, and `Push the per-arch tags` still carries a bare `if: inputs.push` with no `always()` anywhere in it — so if someone ever "fixes" the push condition to match, that test goes red rather than a non-booting image reaching a registry.

### Not chased

`Unit Tests` — the `core-link-bearer` "exactly now" boundary flake. Untouched here; it wants its own ticket, and now that it has been green and red on the same branch there is a second data point for whoever takes it. `Dependency Audit` and `E2E — Harness offers (systemd)` unchanged at `main` parity.

Locally: eslint clean on `scripts`, the two image test files green (91 tests; the one red is `is the only vulnerability allowlist anywhere in the tree`, which walks the filesystem and finds `.trivyignore.rego` copies inside my own gitignored `.claude/worktrees/` checkouts — 0 tracked files under that path, so CI does not see them). The `/proc` reader is exercised against a real `/proc`, but Docker is still unavailable here, so the leg itself is CI's to observe.

Over to the run.

qcentic-adm (approved):
## Approving — the substance is verified. One predicate left, and the failure output proves the product is right.

**The CVE gate ran, and this was my biggest open concern on #38. It is now measured rather than argued:**

```
Trivy: 123 finding(s), 46 distinct (linux-libc-dev suppressed by .trivyignore.rego).
  CRITICAL 1 · HIGH 6 · MEDIUM 88 · LOW 28 · UNKNOWN 0
  7 fixable CRITICAL/HIGH reported but not gated, under usr/local/lib/node_modules/npm/
[core-image-cve] gate passed — nothing fixable and CRITICAL/HIGH outside 7 reported-only finding(s).
```

46 distinct, exactly the figure `core.Dockerfile` claims. `assertSuppressionApplied()` passed, so the Rego policy demonstrably took effect rather than parsing cleanly and filtering nothing. And the seven scoped-out findings are printed by name — `brace-expansion`, `ip-address`, `tar`, `undici`, all under the bundled-npm prefix and nowhere else. That is the claim I said should not merge on an argument alone, and it is now a measurement. Thank you for taking the `always()` route rather than reordering; the promise #38 made is intact.

### The remaining red is one predicate, and your own failure output disproves it

```
  PID  PPID COMMAND
    1     0 tini
    7     1 MainThread
```

The topology half of `9495e4a` is right and **it passes**: PID 1 is `tini`, and the daemon is its direct child. D14 holds, on the shipped bytes. Reading `ppid` as field 4 counted after the closing `)` is the correct way to parse `/proc/<pid>/stat`, and refusing to depend on `procps` the image does not ship is the right instinct.

What fails is `comm === "node"`. Node renames its main thread, so `/proc/<pid>/comm` reads `MainThread` — `comm` is whatever the last `prctl(PR_SET_NAME)` set, capped at 15 bytes, and it is not a reliable identifier for a Node process. The comment says "the launcher `exec`s the bundled Node in place, so the daemon reads as `node`"; the `exec`-in-place part is right, the name it reads as is not.

`ppid === 1` already identifies it uniquely — nothing else in that table has PPID 1, and the `docker exec` shell shows PPID 0 because its parent is outside the namespace. So dropping the `comm` clause is sufficient, and if you want it positively identified rather than merely counted, `/proc/<pid>/cmdline` carries `actana-cli.cjs` and does not get renamed.

### Why I am approving rather than holding

Everything this PR claims, I have verified — the ten tickets' done-conditions, the CVE gate on the built image, the mint/persist/blob path end to end, `1000:1000`, tini's topology, the Panel image and its published-manifest healthcheck. The three findings I raised were all real and all fixed properly, and the boot counting in `55f3e49` was better than what I asked for.

Being explicit so approval does not overstate what ran: **the legs after the PID check — the refused verbs, the Panel pairing, restart-is-a-no-op, and `down -v` unpairs — have still never executed.** They are #46's and #40's central assertions and that one predicate is all that stands between them and a green run. Land it and let CI finish the job; I do not need another round to look at it.

Phase 3 is not gated on me.

Issue #50, opened 2026-08-03
Drop macOS Core targets from the release path


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D28**.

**Phase 3 — CI and release**
**Blocked by:** #33
**Blocks:** nothing

## What

Three assets, not five. The consequences are small and each is a trap if missed.

The `--expect` guard is load-bearing: it is what makes a silently missing architecture a **red build** rather than a checksum file that quietly covers half the release. Change it deliberately; do not let it fail and get patched.

`install.sh` currently maps `darwin` → `mac`, so a Mac user gets the *late* error `release v0.1.0 has no build for mac-arm64` — which reads as a broken release rather than an unsupported platform. It should die at detection with the same message the Windows branch already uses: Cores run on Linux; run the Panel and host Cores there.

## Done when

- [ ] `CORE_TARGETS` (`scripts/lib/core-tarball.mjs`) drops its two `darwin` rows
- [ ] `compose-core-shasums.mjs --expect 4` → `--expect 2`, changed deliberately with a comment
- [ ] The release build matrix drops `macos-15` and `macos-15-intel`
- [ ] `install.sh` fails at `detect_target` on Darwin with the Linux-only message, not late with a missing-asset error — covered by `install-sh.test.mjs`
- [ ] `docs/harness-macos-prerelease-checklist.md` is deleted **after** T26 has promoted its one load-bearing line (line 47: no Gatekeeper dialog is a release blocker)

---

<sub>Decisions are locked in ADR 0016. If this ticket needs one changed, amend the ADR in the same PR — do not decide it in a comment.</sub>

qcentic-adm:
Superseded by #90 (charting session with the maintainer, 2026-08-05): macOS returns as an on-device Core target — mac-arm64 only, built by a **release-only leg behind a manually-approved GitHub environment**, with zero macOS jobs in per-PR CI. The mechanical items in this ticket's done-when are already merged on `main` (darwin rows dropped, `--expect 2`, Darwin refusal in `install.sh`); #90 deliberately reverses them under the new decision and carries the ADR 0016 D28 amendment. The one unexecuted item here — deleting the macOS prerelease checklist — must **not** be done: that checklist becomes the manual approval gate. Closing so nobody "finishes" this against the new decision.

Issue #55, opened 2026-08-03
K3: the no-code-signing paragraph, and repoint install.sh:295


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D46**.

**Phase 4 — Docs and deletions**
**Blocked by:** nothing — can start immediately
**Blocks:** #58

## What

The strongest case in #17's set. `grep -ri 'quarantine|xattr|gatekeeper'` across every `.ts`, `.mjs` and `.sh` returns **zero hits**, and today the fact lives in two half-places: `install.sh:295` cites *a file this effort deletes*, and the macOS prerelease checklist asserts the outcome without the reason. ADR 0010 explicitly deferred the question to "Harness distribution planning" — it was answered there, in files about to be deleted.

Blocks T21 (which deletes the checklist) and T29 (which deletes the spec).

## Done when

- [ ] `docs/contribute/ci-cd.md § The release artifacts` records: integrity is **published checksums, verified by both `install.sh` and `actana update`** — not signatures. No Apple notarization, no Windows Authenticode
- [ ] Records *why that is safe*: the `curl | bash` path never sets `com.apple.quarantine`, so Gatekeeper never intervenes. Revisit only if a browser-download distribution is ever added
- [ ] Records the corollary that already cost one review cycle: do **not** add `xattr -dr com.apple.quarantine` to `actana setup`
- [ ] `install.sh:295` cites `docs/contribute/ci-cd.md` instead of "spec"
- [ ] The supply-chain follow-up ticket is cross-linked, so the paragraph reads as a current posture rather than a permanent refusal

---

<sub>Decisions are locked in ADR 0016. If this ticket needs one changed, amend the ADR in the same PR — do not decide it in a comment.</sub>

qcentic-adm:
Shipped in the **0.3.2** release train. It reached `main` as **6cd15ca** (`chore: 0.3.2 improvement sweep — docs, contracts and repo hygiene`), promoted through **#262**; the work came in on the train as **#256**. Verified against the tree at `origin/main`, box by box.

[...]

Issue #23, opened 2026-08-03
Build the 0.1.0 shape


qcentic-dev:
## Destination

**A released `0.1.0`**: two container images on Docker Hub, one `curl … | sh` installer, three CI workflows, one vocabulary, a task-shaped docs tree, and a protected `main`.

This map builds what [#6](https://github.com/actana/control/issues/6) decided. Nothing here is a decision — every one is locked in **[ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md)**, which each ticket cites by clause. If a ticket needs a decision changed, **amend the ADR in the same PR**; do not settle it in a comment.

**Tracker gotcha — read before any `gh` call.** No git remote in this clone points at `actana/control`. `origin` is `AgentSystemLabs/mission-control` (the read-only upstream fork parent), so a bare `gh issue …` targets the **wrong repo**. Always pass `--repo actana/control`.

## Where this stands

*Updated 2026-08-05.* **Phases 1–3 are landed** (every ticket closed except #31, the opencode install path — the weekly `harness-canary` covers the regression surface meanwhile). Remaining: #31, all of Phase 4 (#53–#60, plus #17's deletion phase), the launch-polish additions below, and Phase 5.

**Four amendments since this map was charted**, each with its own record:

- **GHCR is retired — Docker Hub is the only registry** (PR #89, ADR 0018; supersedes D31/D32's ordering clauses). Docs tickets landing after it must not reintroduce `ghcr.io` names.
- **macOS returns as an on-device Core target** (#90; amends ADR 0016 D28 in its PR). #50 executed the old decision and is closed as superseded — its checklist must not be "finished": the macOS prerelease checklist it wanted deleted is now the manual approval gate.
- **The 58-line README is superseded by a rich shop-window README** (#93; amends D42's budget in its PR).
[...]

PR #69, opened 2026-08-04
ci(phase-3): nine workflows become three — ci, release, housekeeping

### LINKED ISSUE #47: ci.yml: 26 jobs → 13
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D34, D35, D37**.

**Phase 3 — CI and release**
**Blocked by:** #30, #33, #37, #43
**Blocks:** #51

## What

The diagnosis matters for what gets cut. **Nothing in CI is slow** — measured, the whole suite is 3m38s wall clock, the slowest job is 3.5 min and every container e2e leg is 1.0–1.5 min. CI is expensive because of job count and **macOS billed at 10×: 72% of the ~89 billed minute-equivalents is three macOS legs** whose product justification #6 already deleted.

Target: **~5 min wall clock (unchanged) and ~17 billed minute-equivalents (~5× cheaper).** Critical path becomes the Core image job; everything else finishes inside it.

Each deletion below names the realistic failure it would have caught — and every one of them is "none".

## Done when

- [ ] Kept on PR: `typecheck`, `unit-tests`, `lint`, `dependency-audit`, `secret-scan`, `conventions` (3 checks, **1** job), `core-tarball-smoke` ×2 arches, `panel-e2e` (ubuntu), `panel-image`, `core-image`, `installer-e2e` ×2 distros
- [ ] `dependency-audit` becomes `pnpm audit --prod --audit-level high` — still catches `postcss`/`@babel/core`, drops the two dev-only `brace-expansion` highs that no contributor caused or can fix
- [ ] Deleted: `react-doctor.yml` (cannot fail — `blocking:` is commented out; uses floating `@v2`/`@v7` tags while holding `pull-requests: write`, `issues: write`, `statuses: write`, against this repo's SHA-pinning rule; and it is React code-quality, which #6 scopes out) and the `doctor` script from `package.json`
- [ ] Deleted: `smoke-standalone-harness.mjs` — it and the tarball smoke call the same `assertBootsAndDials`; nothing ships the path it uniquely covers
- [ ] Deleted: both macOS `actana setup` e2es, the `panel-e2e` macOS leg (the Panel ships nowhere as a plain macOS process), and `e2e-actana-setup-macos.mjs` with the launchd paths
- [ ] `conventions.yml` folded in as one job — three jobs each `npm install`-ing commitlint is three copies of one 30s setup
- [ ] `images-edge.yml` deleted; `:edge` and `sha-<short>` become a push-to-`main` condition on the existing reusable `container-image.yml` call, with `paths-ignore` for `**.md`/`docs/**` carried across
- [ ] Measured before/after job count and billed minutes recorded in the PR

---

### LINKED ISSUE #48: Merge the two installer e2es into one
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D36**.

**Phase 3 — CI and release**
**Blocked by:** #33
**Blocks:** nothing

## What

The one-liner becomes the **entry point** of the setup e2e rather than a separate suite: `curl … | bash` against the fixture release server installs the machine, then the lifecycle assertions run on the machine it produced. Strictly larger coverage at roughly the cost of one, and it deletes a duplicated install phase plus one container boot per cell. Both already share `systemd-container.mjs` and the same fixture server, so this removes a seam rather than adding one.

**What makes it safe:** install-sh's negative cases — bad checksum, unknown platform, `--version` pinning, non-TTY behaviour, exit codes — are already covered *hermetically* by `scripts/__tests__/install-sh.test.mjs` in under a second. The container leg only ever needed to prove the script works on a real init system with a real tarball.

Axes are `distro × arch`: **Ubuntu and Debian at x64 on PR**, because PAM, polkit and the logind rules that decide whether a sudo-less `systemctl --user` and `loginctl enable-linger` work at all are exactly what differs between distros. **arm64 moves to the tag run**, because the arch-sensitive risk is prebuilt native modules and `core-tarball-smoke` already boots the arm64 tarball on an arm64 runner every PR.

## Done when

- [ ] One script covers: install via the one-liner, `status`/`token`/`start`/`stop`/`restart`/`logs`, in-place upgrade, `update` from a fixture channel, a tampered update that aborts leaving the old install running, `update --version`, `token regenerate` invalidating the old blob, `uninstall`, `--purge-data`, lingering, and T7's legacy-unit removal
- [ ] `e2e-install-sh-linux.mjs` is deleted, not left orphaned
- [ ] 2 legs on PR (ubuntu, debian; x64), arm64 on tag
- [ ] `scripts/lib/container-matrix.mjs` and its unit test stay — they are what stops the declared axes and the workflow drifting apart
- [ ] `fixture-release-server.mjs` untouched and still load-bearing: until the repo is public it is the **only** way to exercise the installer at all

---

### LINKED ISSUE #49: release.yml: one workflow, three assets, two images
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D28, D29, D30, D31, D32, D34**.

**Phase 3 — CI and release**
**Blocked by:** #33, #34, #37, #43
**Blocks:** #51, #52

## What

Replaces `harness-release.yml` + `images-release.yml` + `dockerhub-description.yml`. On `push: tags: v*` plus `workflow_dispatch`. Measured target: **a release lands in under 6 minutes** (today's equivalent Images Edge run — both images, both arches, both registries — took 3m16s; tarball legs are ~0.7–0.8 min each).

Jobs: `resolve` → `tarball` (matrix ×2, native runners) → `panel` / `core` (via the reusable `container-image.yml`, `push: true`) → `github-release` → `descriptions` (T23).

**Two things that must not be "fixed" later** and belong in the workflow as comments:

- **GHCR is pushed first and must complete**, even though Docker Hub is the "primary" registry. Primary means the canonical place a user is told to pull from, not publish order. GHCR authenticates with the workflow's own `github.token` and therefore *cannot* fail on credentials, which is exactly why it goes first.
- **Present-but-broken Docker Hub credentials publish GHCR completely — per-arch tags *and* manifest — then fail the run at the very end.** Failing early would strand GHCR with per-arch tags and no manifest.

## Done when

- [ ] A tag publishes exactly three assets: `actana-core-<version>-linux-x64.tar.gz`, `actana-core-<version>-linux-arm64.tar.gz`, `SHA256SUMS`
- [ ] Image tags: `<version>` and `latest` from a version tag (non-prerelease only); no `0`/`0.1` ladder — under semver a `0.x` minor bump *is* the breaking change, so a moving `:0` tag is a lie by construction
- [ ] `v1.0.0-rc.1` publishes `1.0.0-rc.1` and does **not** move `:latest`
- [ ] **One** Docker Hub PAT serves both `docker login` and the description API; `DOCKERHUB_DESCRIPTION_USERNAME`/`_TOKEN` are deleted. (An OAT can push images but returns *"Cannot log into an organization account"* on `POST /v2/users/login`, so it would mean paying for Team/Business **and still** provisioning a PAT)
- [ ] A missing Docker Hub credential **fails on `actana/control`** and still skips on forks — a release that never reached the primary registry must not report success
- [ ] PAT rotation is documented in `docs/contribute/repo-setup.md`: the token belongs to a person and dies with that account
- [ ] `install.sh` is **not** an asset and stays on `raw.githubusercontent.com/actana/control/main/install.sh`
- [ ] The installer-contract rule from ADR 0016 D29 is quoted verbatim in the workflow header

---

### LINKED ISSUE #51: housekeeping.yml: stale, weekly Core rebuild, weekly audit, vendor canary

### LINKED ISSUE #52: Docker Hub descriptions: fold into release.yml and rewrite docs/images/core.md

qcentic-dev:
## Summary

Phase 3 of the 0.1.0 rebuild ([ADR 0016](../blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D28, D33–D38**): the workflow tree collapses from nine entry points to three, and the release path stops shipping things that no longer exist.

**Nine workflows become three**, plus `container-image.yml` as a reusable workflow rather than a fourth entry point:

| | trigger | what it is |
| --- | --- | --- |
| `ci.yml` | PR, push to `main` | the gate, plus the `:edge` publish on merge |
| `release.yml` | `v*` tag | three assets, two images, the GitHub Release, both Docker Hub pages |
| `housekeeping.yml` | crons | five chores, none of which gates anything |
| `container-image.yml` | `workflow_call` | one image build, reused across PR / edge / release |

Deleted along the way: `conventions.yml`, `images-edge.yml`, `images-release.yml`, `core-release.yml`, `dockerhub-description.yml`, `react-doctor.yml`, `base-pins.yml`, `stale.yml`.

**The PR gate halves.** 26 jobs → 19 here; D35's target of 13 is the six installer legs away, and `#48` takes exactly those six. The bill falls further than the count does, because 72% of it was three macOS legs whose product justification #6 already deleted.

**The release ships three assets, not five.** `mac-arm64` and `mac-x64` are no longer targets (D28), so the macOS setup e2e goes with them and `docs/core-macos-prerelease-checklist.md` stops being a release gate — it stays in the tree as the manual procedure for anyone running a Core off a local build on a Mac, pending #55.

**One installer e2e, entered at the one-liner** (D36). Ubuntu and Debian at x64 on every PR; arm64 once per release, gating `github-release`, because an arm64 tarball the one-liner cannot install is not an asset worth attaching.

**The Docker Hub pages describe the product** (D33). The `sync()` body moves verbatim — JWT exchange, the 100/25000-byte guards, `curl` rather than a third-party action, because that credential can rewrite the public face of both images. The `descriptions` job hangs off `[panel, core]`, not off `github-release`: what a page waits for is the image being pullable.

**Two chores end in `gh issue create` rather than a red run** (D37, D38) — the weekly dev-tree audit and the four-vendor Harness-install canary. Both are red for reasons no PR author caused and none can fix, and a permanently red scheduled workflow is how a team learns that red means nothing.

## Related issues

Closes #47
Closes #48
Closes #49
Refs #50
Closes #51
Closes #52

Part of #23. Unblocks #61.

## Type of change

- [x] 📚 Documentation
- [x] 🔧 Build / CI / tooling

## How was this tested?

Nothing builds a workflow file, so the checks that would otherwise be invisible are tests: `scripts/__tests__/workflows.test.mjs` asserts the directory listing, each job's cron, and that the two vendor-facing chores open an issue rather than gate; `scripts/__tests__/container-matrix.test.mjs` and `panel-image.test.mjs` follow the job graph across the three publish paths and assert no file in the publish path still calls the Core image a development fixture.

On the merged tree:

- `pnpm test` — **2179 passing**, 0 failing (302 root, 161 shared, 592 core, 1124 panel)
- `pnpm typecheck` — clean
- `pnpm lint` — 0 errors, 54 warnings, all pre-existing in `packages/panel` and `packages/shared` and untouched here
- `pnpm scan:secrets` — clean
- all four workflow files and `dependabot.yml` parse as YAML; `install.sh` parses under `bash -n`

## Reviewer notes

Two things worth a second opinion rather than a silent merge:

1. **`ci.yml`'s `push` trigger carries `paths-ignore`**, so a docs-only merge to `main` skips the whole post-merge run — including the `:edge` publish. `:edge` will not move on a docs-only merge. #47 documents this as intentional; flagging it because it is the kind of thing that surprises someone six weeks later.
2. **ADR 0016 D34 is amended in this PR**, per both #51's and #52's footer instruction. The clause could not have named `base-pins.yml` (#43 created it afterwards), so the amendment records that the count is *entry points* and that base-pins is now a job inside `housekeeping`.

`docs/core-macos-prerelease-checklist.md` is deliberately **not** deleted. #50's checklist asks for it, but that deletion is conditional on #55 promoting the Gatekeeper line first, and #55 is still open.

## Checklist

- [x] My PR title follows Conventional Commits and my branch name follows the branching convention
- [x] I updated documentation (`docs/ci-cd.md`, `docs/REPO_SETUP.md`, `docs/images/`, ADR 0016) where relevant
- [x] Dependent changes have been merged


qcentic-adm:
## The reds are not this PR — no job executed a single step. This is a billing stop, and here is the evidence and the way out.

Before anyone starts bisecting a 45-file workflow rewrite against eleven red checks: none of the eleven ever reached a runner.

### What actually happened

Every failed job on run [30920636143](https://github.com/actana/control/actions/runs/30920636143) has an **empty steps array** — not a failed step, no steps at all. The run went `completed` twelve seconds after it started. The logs endpoint returns 404 because no runner ever produced a log. And every one of the eleven carries the identical annotation:

> The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings

I checked four jobs across four different shapes (a plain job, a reusable-workflow job, the new `Conventions` job) — same annotation on all of them. This is an account-level stop, not a workflow-level one.

### The numbers behind it

The org billing API says August so far, all of it on this repo:

| SKU | minutes | ×multiplier | min-equivalents |
| --- | --- | --- | --- |
| Linux | 687 | 1× | 687 |
| Linux ARM | 164 | 1× | 164 |
| macOS 3-core | 151.5 | **10×** | 1,515 |
| | | | **2,366** |

Every line item this month has `netAmount: 0.00` — included free minutes absorbed everything, until they didn't. The last run whose jobs actually executed started at **13:12 UTC** (the Dependabot CI runs); this run at **14:46** could not start anything. The stop arrived in that window, which is consistent with the quota being crossed mid-day — 2,366 is past the 2,000 min-equivalents a Free-plan org includes, and the default spending limit past the included quota is **$0**, which blocks rather than bills.

I can't see the org's plan or payment state from here (that needs `admin:org`), so I won't pick between the message's two clauses — exhausted quota with a $0 limit, or a failed payment on a paid plan. Both are fixed on the same page.

### One thing the run did prove

GitHub **parsed and materialized the new workflow tree correctly**. All 19 jobs of the rewritten `ci.yml` were created with the right graph: the edge-publish jobs skipped on `pull_request` exactly as the `!= 'pull_request'` conditions intend, both `container-image.yml` reusable calls resolved, `Resolve edge tags` gated properly. The phase-3 shape loads; it was never given a runner. That is not nothing for a PR whose entire content is workflow files.

### How to get CI running again

1. **Org owner → `github.com/organizations/actana/settings/billing`.** If a payment failed, fix the payment method. If it's the quota, raise the Actions spending limit above $0 — at these rates the exposure is small: the *entire month's* gross so far is about **$14**, and most of that rate disappears with this PR (next point).
2. **Then re-run without touching the branch:** `gh run rerun 30920636143 --failed --repo actana/control` (or *Re-run failed jobs* in the UI). No new commit needed; the run re-reads the same workflow files at the same SHA.
3. **If no billing action is possible right now**, the included quota resets at the start of the next billing cycle — the date is on the same settings page. Nothing on a private repo runs before then.
4. Two alternatives exist and I'd take neither for this: a self-hosted runner (a `runs-on` change and an ops commitment, not a one-PR fix) and making the repo public (standard runners become free, but that's a product decision wearing a CI costume).

### The part worth savoring

**64% of the month's bill is the macOS legs this PR deletes.** The measured table in the `ci: cut the PR job count` commit — 143 billed min-equivalents per PR run before, 37 after — is the structural fix for exactly the wall this run hit. The PR that cuts the bill is the one the bill stopped. So the sequencing is: billing unblocked first (step 1), this run re-run (step 2), and then this PR's merge is what keeps it from happening again — at 37 min-eq/run the same 2,000-minute quota carries roughly 4× the PR traffic.

### On the content review

No verdict on the diff yet, and the local figures in the description (`2179 passing`, clean typecheck/lint/secrets) are the author's machine, not the gate — they can't substitute for required checks. Substantive review to follow once a run actually executes. Two things I'll be looking at when it does, flagged now so they're not a surprise: the `paths-ignore` on the push trigger (reviewer note 1 — I agree it deserves the flag) and whether `Conventions` reading the PR title from the API instead of the event payload behaves on the re-run of an edited PR.

qcentic-adm:
## Approving — the shape is right, the run is green where it can be, and the one red is `main`'s, not yours. One description edit to make, four follow-ups worth a ticket, none a blocker.

Second review, now with a run that actually executed (the first "review" was a billing post-mortem, comment above). I read all four workflow files end to end, the commitlint rule, the installer change, the matrix module, and the new tests, and checked every claim below against this run's logs or `control/main` — nothing here is asserted from the diff alone.

### The run: 15 green, 1 red, and the red is inherited

`Core image / Build + smoke (amd64)` failed on the same predicate that is red on `main` and was red on #66/#67/#68: the Core **booted** (minted `core_31f2306f1295344b`, printed its pairing token), the table shows `1 tini / 7 1 MainThread` — the daemon is tini's child, which is the property D14 wants — and the smoke greps for a process whose `comm` is `node`, which the daemon does not report. Everything else is green, including both installer e2e legs on the new merged suite, the arm64 tarball smoke on real arm hardware, and the Trivy gate (passed, same 7 reported-only npm-tree findings as `main`, JSON artifact uploaded).

Not this PR's failure — but this PR sharpens it. Your own D37/D38 argument is that a permanently red check "is how a team learns that red means nothing", and after this merges, `Core image` is one of the three required container checks on every PR. The one-line `comm` fix in `scripts/smoke-core-image.mjs` (match on PPID 1 + a `node`-or-`MainThread` comm, or just drop the comm test since the PPID test carries the invariant) has been on the table since #66. I'd land it in this PR or the very next one — by this PR's own philosophy it cannot wait long.

### One description edit: `Closes #50` → `Refs #50`

The description itself says #50's checklist deletion "is conditional on #55 promoting the Gatekeeper line first, and #55 is still open." That is an unmet done-condition, and the repo's governance rule for that case is `Refs`, not `Closes` — otherwise the merge closes #50 and the remainder stops being tracked anywhere. The other five `Closes` look fully earned.

### Four follow-ups (two are one-liners in files this PR owns)

**1. A `workflow_dispatch` of ci.yml on any branch publishes `:edge` from that branch.** The edge jobs gate on `github.event_name != 'pull_request'`, and a dispatch can target any ref — so dispatching CI on a feature branch (for any reason) republishes `:edge` and `sha-<short>` from unmerged bytes. I checked the deleted `images-edge.yml`: its dispatch was equally unguarded, so this is carried across, not introduced — but the surface grew, because before, the only reason to dispatch that file was to publish edge, and now ci.yml is a file people will dispatch for other reasons. One line on `edge-tags` closes it (the two image jobs follow via `needs`):

```yaml
if: github.event_name != 'pull_request' && github.ref == 'refs/heads/main'
```

**2. A second merge to `main` can cancel the first one's edge publish mid-push.** `concurrency: ci-${{ github.ref }}` + `cancel-in-progress: true` puts all main-push runs in one group, so back-to-back merges cancel the earlier run — possibly between `docker push` and `imagetools create`, which strands per-arch tags without a manifest: the exact failure release.yml's own header says must never be created deliberately. Also inherited (`images-edge.yml` had a single cancelling group too). Cancelling superseded *PR* runs is right and worth keeping:

```yaml
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

**3. The revision label lies on exactly the republishes it exists for.** `container-image.yml` checks out `inputs.ref` but labels the image with `${{ github.sha }}`. On a tag push they coincide. On a release `workflow_dispatch` (re-release) and on the weekly `core-rebuild` — the two paths that rebuild old refs — `github.sha` is the tip of `main`, so the republished `:0.1.0`/`:latest` carries an `org.opencontainers.image.revision` pointing at a commit it was not built from. `rev="$(git rev-parse HEAD)"` after checkout, and pass that to the label and `IMAGE_REVISION`.

**4. The two Docker Hub login error messages contradict each other.** The build job's says an ORGANISATION access token works with `DOCKERHUB_USERNAME` = the org name; the publish job's says the username must be the token-owning account, "not the organisation name"; release.yml's `resolve` comment says OATs can't serve the description API at all. An operator debugging a failed push at midnight will hit one of these and follow it. Whichever stance is the tested one (resolve's reads like it), make all three say it.

### Verified sound, so nobody re-litigates it later

- **Fork-safety, which went from theoretical to live when this repo went public mid-review:** no `pull_request_target` anywhere; PR-path image jobs grant `packages: read` and pass no secrets; every attacker-influenceable value (PR title, branch name, base/head SHA) travels through `env:` and is quoted or piped, never interpolated into a script. The title check reading from the API instead of the payload also happens to read the *current* title on re-run — the `types: [edited]` problem solved without re-running the gate on description edits. Good trade.
- **Deleting `react-doctor.yml` removes the one place that held `pull-requests: write` + `issues: write` + `statuses: write` against floating action tags.** On a public repo that job was the soft spot; its deletion is a security fix wearing a cost-cutting justification.
- **The commitlint `trailer-leading-blank` rule** is the right shape: finds the footer by known trailer tokens with a required separator, leaves wrapped prose alone, and replaces the rule forward rather than rewriting merged commits — consistent with how conventions bind here. The false-negative trade is documented in the rule itself. `scripts/__tests__/commitlint-config.test.mjs` covering both directions is what makes me comfortable with a hand-rolled parser.
- **`resolve`'s tag regex** requiring all three components kills the `v0` / `v0.1` moving-ladder publish and the `+build` Docker-illegal tag, and prereleases correctly don't move `:latest`.
- **The GHCR-first / fail-at-the-end ordering** with its "do not fix this later" header, the `--label`-less `gh issue create` calls, the one-open-issue-at-a-time guard, and `workflows.test.mjs` turning `ls .github/workflows` into a failing test are all exactly right.
- **The billing postscript:** this run executed entirely on free public-repo runners — the wall from yesterday cannot recur for this repo, and the 26→19 job cut still stands on its own merits as wall-clock and noise reduction.

Approve. Fix the `Refs #50`, decide whether the two one-liners (edge gating, concurrency) ride along now that ci.yml is open on the operating table, and give the `comm` predicate its ticket if it doesn't ride along too.

qcentic-adm:
## The merge blocker is fixed in `eb3820c`, and the owner has set two publishing directives that change this PR's registry story. Both outlined below so the next push is the last one.

### The blocker — `Core image / Build + smoke (amd64)` — fixed, pushed to this branch

**What it was.** The smoke's D14 check looked for a process with `comm === "node"` whose parent is PID 1. But `/proc/<pid>/comm` reports the *thread* name, not the executable, and the bundled Node names its main thread `MainThread` — so the daemon of a perfectly healthy image reads as `MainThread`, the predicate never matched, and the check failed on every boot. The run's own process table showed the actual property holding the whole time: `1 tini / 7 1 MainThread` — tini at PID 1, the daemon its child, everything D14 wants.

**The fix** (`eb3820c`): the child-of-PID-1 check accepts `node` or `MainThread` as the daemon's comm. Nothing else moves — a daemon that *is* PID 1, the failure D14 exists to catch, still dies loudly. This red predates the branch (identical on `main` and on the #66–#68 runs), but this PR makes the job a required check on every PR, so a permanently red required check could not be allowed to merge in — by this PR's own D37/D38 reasoning.

CI is re-running on the new head now; that check should go green with the other 15.

### Directive 1 — no `:edge` channel. Tags are `<version>` and `latest`, moved by release tags only.

The edge publish (`:edge` + `sha-<short>` on every push to `main`) is not wanted. What this removes from the PR's files:

- `ci.yml`: the `edge-tags`, `panel-image-edge` and `core-image-edge` jobs, and with them the `secrets: inherit` / `packages: write` on the push path. The `push:` trigger + `paths-ignore` block and the `workflow_dispatch` exist mostly to serve edge — decide whether a post-merge gate re-run on `main` is still wanted at all, or whether `ci.yml` becomes PR-only.
- `scripts/__tests__/panel-image.test.mjs` — the test "tags `:edge` and an immutable per-commit tag, and never moves `:latest`" inverts: nothing may publish `:edge`.
- `docs/ci-cd.md`'s three-paths table drops to two (PR gate, release), and ADR 0016 **D30 needs an amendment** — an amendment recording the decision, not an in-place edit, same as every prior clause correction.
- A side effect worth naming: review follow-ups 2 and 3 above (dispatch publishing `:edge` from any branch; a second merge cancelling an edge publish mid-push) dissolve entirely once there is no edge publish. Nothing to fix there anymore.
- One honest trade to record in the ADR amendment: `sha-<short>` was the only immutable per-commit pin and `:edge` the only try-before-release channel. Dropping them is fine — but it is a decision, and the amendment is where it lives so nobody "restores" it as a favor.

### Directive 2 — no GitHub-owned images. Docker Hub is the only registry; plain `docker build` + push, nothing published under GHCR.

All `ghcr.io` publishing goes. Concretely:

- `container-image.yml`: `resolve` stops emitting a GHCR name; the `docker login ghcr.io` calls, the GHCR push, and the GHCR manifest stitch all go. The image builds under the Docker Hub name directly. `packages: write` disappears from every caller.
- The header's "two things that must not be fixed later" (GHCR-first ordering, fail-at-the-end) is **obsolete under this directive** — it existed to protect a GHCR publish from a Docker Hub credential failure. Rewrite it rather than delete it silently, and amend **D31/D32**, which currently define the two-registry posture.
- Docker Hub credentials become load-bearing with no fallback: `resolve` in `release.yml` already makes them fatal on `actana/control`, which is now the only correct behavior — but the "a fork publishes to GHCR alone" path is gone. A fork with no `DOCKERHUB_*` secrets can build and smoke but publish nowhere; the fork-facing notice text should say that plainly.
- `housekeeping.yml`'s weekly `core-rebuild` republishes to Docker Hub only — same job, one registry.
- Docs: `docs/ci-cd.md`, `docs/REPO_SETUP.md` §2, and both `docs/images/` pages currently name GHCR pulls; all of it moves to `docker.io/actana/<image>` only.
- The existing green run already proves the build/smoke half of this works without any registry at all (PR builds push nowhere) — the change is confined to the publish legs.

### How to proceed

The blocker is resolved on this branch. For the two directives, my recommendation: **merge this PR once the re-run is green, and land the registry changes as one focused follow-up PR** — this one is 45 files and six closed tickets deep, it is finally green, and the directives above touch the same three workflow files coherently on their own. If the owner prefers them folded in here instead, the outline above is the complete checklist — nothing else in the PR needs to reopen either way.

qcentic-adm:
## `Core image / Build + smoke (amd64)` is still red after `eb3820c` — but it moved. The PID fix worked; the failure is now a wrong assertion in a leg that has never executed before.

### What the new run shows

`eb3820c` did its job — the run got past the predicate that had been red since #66:

```
[core-image-smoke] the Core minted core_45774b3beb25101f and printed its pairing token
[core-image-smoke] tini is PID 1 and the daemon (pid 7) is its child
[core-image-smoke] the baked Core tarball is linux-x64
[core-image-smoke] setup, start, stop, restart, update, uninstall, logs all refuse with a Docker command to run instead
[core-image-smoke] booting the Panel …
[core-image-smoke] panel service up on http://127.0.0.1:34343
[core-image-smoke] FAIL: the Panel registered core_fa31bad8484f02fb, but the token names core_45774b3beb25101f
```

Three legs that had never run before (manifest target, refused verbs, Panel boot) passed on first execution. The fourth — the pairing leg — failed on its first execution ever.

### The failure is in the smoke, not the product

The smoke asserts that the id `POST /api/cores` returns equals the `coreId` inside the pairing token's bearer. **That equality is not the contract and never was:**

- `registerCoreFromRegistrationBlob` (`packages/panel/src/server/services/cores.ts`) mints a fresh Panel-side handle per registration — `newCoreId()`, whose own comment reads "`core_` + 64 bits. Not a secret — a collision-resistant handle." Nothing in the registration path reads the bearer's `coreId` into the row.
- The Panel's unit tests (`services/__tests__/cores.test.ts`) register blobs and use the returned id throughout — none of them expects it to match the bearer's.
- The dialer keys on the Panel's id too: `coreLinkManager().dial(core.id)`.

So the Core's self-identity (inside the signed bearer, ADR 0002) and the Panel's registry key are two namespaces that share a `core_` prefix and nothing else. The smoke conflated them — understandably, since this leg had never produced a run to disagree.

### The fix

In `scripts/smoke-core-image.mjs` (~line 444), drop the cross-namespace equality and keep the assertion that actually proves the token named this Core — the dial reaching `connected`, which only the bearer's Core can answer, and which the very next line already asserts:

```diff
+// The Panel's registry key, not the Core's self-identity: cores.ts mints a
+// fresh handle per registration and nothing adopts the bearer's coreId. The
+// dial below reaching `connected` is what proves the token named this Core.
 const coreId = added.body?.core?.id;
-if (coreId !== first.coreId) {
-  die(`the Panel registered ${coreId}, but the token names ${first.coreId}`);
-}
+if (typeof coreId !== "string" || !coreId) {
+  die(`the Panel registered the Core without returning an id: ${added.text.slice(0, 200)}`);
+}
```

Nothing downstream conflates the namespaces: the restart leg (line ~480) and the volume-destroy leg (~514) compare Core-side ids with Core-side ids, and the registry lookups (~530) use the Panel's id. This is the only line.

### Set expectations before the next push

This was the **first** of the never-executed legs. After it come the restart-keeps-pairing, single-token-notice, and volume-destroy assertions — all also running for the first time when this fix lands. Each may hold, or may surface the next first-observation surprise; the two failures so far were both in the smoke's assumptions (`comm` as a thread name, id namespaces), not in the image. If the run after this fix goes red again, read it in that light before suspecting the image.

qcentic-adm (approved):
Approved. Final round verified on head d24f662b9: the full gate is green (run 30927852413), and the Core image smoke completed end-to-end for the first time — boot, PID-1 topology by argv[0], refused verbs, Panel pairing, restart-keeps-pairing, and down -v unpairing all held against the shipped bytes. Both follow-up commits are correct fixes to the smoke's assumptions, not weakenings: the daemon is now identified by what the launcher ran, the pairing check matches the Panel's actual registry contract with the dial-to-connected proof intact, and the manifest healthcheck gate reads the daemon's config instead of a Go struct that couldn't carry the answer (which also explains the phase-2 edge publish failure on main). Closes #50 is now Refs #50 per the governance rule, since the checklist deletion still waits on #55. The registry directives (no :edge, Docker Hub only) remain the agreed follow-up PR.

PR #82, opened 2026-08-05
fix(core-image): pay down 4 of 7 fixable CRITICAL/HIGH — npm 11.19.0, Node 24.19.0

qcentic-adm:
## What

Four commits on the Core image's security surface, driven by a layer-by-layer investigation of the raw 1246-CVE scan count:

- **npm pinned to 11.19.0** (new `ARG NPM_VERSION` + `ARG NPM_SHA512`) over the bundled 11.17.0. All seven fixable CRITICAL/HIGH in the shipped image live in npm's vendored tree (`tar`, `undici`, `brace-expansion`, `ip-address`), per the 2026-08-04 measurement; 11.19.0 clears four, including the only CRITICAL (both `tar` rows). **Not 12.0.2**: its vendored tree is identical for all four (verified from both published tarballs), and npm 12's `allowScripts`-off default blocks dependency install scripts and implicit `node-gyp rebuild` — the capability this image's toolchain exists for. The tarball is fetched from the registry and SHA-512-checked against the pinned digest (its registry `dist.integrity`, base64-decoded), on the same D8 footing as the Node fetch.
- **Node pin 24.18.1 → 24.19.0** via `check-base-pins.mjs --fix-node`. Not a security bump (24.18.1 was itself the current v24 security release) and not a prerequisite for the npm pin — a deliberate stay-current-while-touching-the-layer choice, reconciled in the evidence doc. Both base image digests measured current in the same run.
- **Evidence** in `docs/research/core-base-measure/core-image-cve-claim-2026-08-05.md`: 1200 of the 1246 raw findings are kernel CVEs mapped onto `linux-libc-dev` (header-only, host-owned kernel), independently confirmed against OSV.dev's Ubuntu noble export; no toolchain on Ubuntu avoids it (`libc6-dev` hard-depends on it).

The `bakes no Harnesses` guard (D9) strips the one exact npm-tarball install form before asserting — any other global install still fails it.

## Result

Pays down 4 of 7 fixable CRITICAL/HIGH, including the lone CRITICAL. **3 fixable HIGH remain** (`brace-expansion` → 5.0.8/5.0.9, `ip-address` → 10.3.1): fixed upstream, shipped by no released npm, reported-not-gated pending an upstream release on the #51 cadence. Gated scan otherwise unchanged (46 findings, allowlist still one file, one entry per D7/D37).

## Verification

`scripts/__tests__` — 107 pass across `panel-image` and `base-pins` suites. Pre-existing `core-launcher` failure on `main` (local actana install leaking into the test) is untouched and unrelated.

No rebuilt-image scan was run here (no docker in the sandbox), and the D11 gate is blind to the npm path by construction — the check that the pin worked is the scan summary's reported-only block showing **3 rows instead of 7**, not the gate passing.

### DIFF
--- a/deploy/core.Dockerfile
+++ b/deploy/core.Dockerfile
@@ -92,6 +92,13 @@ RUN apt-get update \
 # out for the same reason the tarball is in — this is a download we verify,
 # not a publisher we trust.
 ARG NODE_VERSION=24.19.0
+# npm newer than the one Node bundles, because that is where the image's only
+# fixable CRITICAL/HIGH live: npm's vendored tar/undici/brace-expansion under
+# /usr/local/lib/node_modules/npm. 12.0.2 clears four of the seven measured on
+# 2026-08-04, including the lone CRITICAL (tar). Pinned exactly, and bumped on
+# the same weekly cadence as NODE_VERSION (#51) — no released npm clears all
+# seven, so this pays findings down; it does not green a raw scan.
+ARG NPM_VERSION=12.0.2
 RUN set -eux; \
     case "$(dpkg --print-architecture)" in \
       amd64) node_arch=x64 ;; \
@@ -107,8 +114,11 @@ RUN set -eux; \
     tar -xJf "${archive}" -C /usr/local --strip-components=1 \
         --exclude CHANGELOG.md --exclude LICENSE --exclude README.md; \
     rm -f "${archive}" SHASUMS256.txt; \
+    npm install -g "npm@${NPM_VERSION}"; \
+    npm cache clean --force; \
+    rm -rf /root/.npm; \
     node --version; \
-    npm --version
+    [ "$(npm --version)" = "${NPM_VERSION}" ]
 
 # uid 1000 and gid 1000, by number, always (D12).
 #
--- a/deploy/core.Dockerfile
+++ b/deploy/core.Dockerfile
@@ -94,11 +94,25 @@ RUN apt-get update \
 ARG NODE_VERSION=24.19.0
 # npm newer than the one Node bundles, because that is where the image's only
 # fixable CRITICAL/HIGH live: npm's vendored tar/undici/brace-expansion under
-# /usr/local/lib/node_modules/npm. 12.0.2 clears four of the seven measured on
-# 2026-08-04, including the lone CRITICAL (tar). Pinned exactly, and bumped on
-# the same weekly cadence as NODE_VERSION (#51) — no released npm clears all
-# seven, so this pays findings down; it does not green a raw scan.
-ARG NPM_VERSION=12.0.2
+# /usr/local/lib/node_modules/npm. 11.19.0 clears four of the seven measured
+# on 2026-08-04, including the lone CRITICAL (tar); no released npm clears all
+# seven, so this pays findings down — it does not green a raw scan.
+#
+# 11.19.0 and NOT 12.x, though 12.0.2's vendored tree is identical for all
+# four (measured from both published tarballs, #82): npm 12 ships allowScripts
+# off by default, which blocks dependency preinstall/install/postinstall AND
+# the implicit `node-gyp rebuild` for any binding.gyp package — the exact
+# capability build-essential/python3 above exist to serve, and how Harness
+# postinstalls run. Crossing that major is a product decision for an ADR, not
+# a rider on a CVE bump.
+#
+# The tarball is fetched from the registry and SHA-512-checked against the
+# digest pinned below — the same "download we verify" footing as the Node
+# fetch above (D8); a bare `npm install -g npm@x` would be publisher trust.
+# The hex is that release's registry dist.integrity, base64-decoded; the two
+# ARGs move together, on the same #51 cadence as NODE_VERSION.
+ARG NPM_VERSION=11.19.0
+ARG NPM_SHA512=48377f8478372aa1c4e47b763475b135836da82436a5700f2e5e8eb5084fc840f93c7b117eb3ad3b5f7d3194c81b6710a10d59448f6ddbcb21ac3fb672bdc003
 RUN set -eux; \
     case "$(dpkg --print-architecture)" in \
       amd64) node_arch=x64 ;; \
@@ -114,7 +128,11 @@ RUN set -eux; \
     tar -xJf "${archive}" -C /usr/local --strip-components=1 \
         --exclude CHANGELOG.md --exclude LICENSE --exclude README.md; \
     rm -f "${archive}" SHASUMS256.txt; \
-    npm install -g "npm@${NPM_VERSION}"; \
+    npm_tgz="npm-${NPM_VERSION}.tgz"; \
+    curl -fsSL -o "${npm_tgz}" "https://registry.npmjs.org/npm/-/${npm_tgz}"; \
+    echo "${NPM_SHA512}  ${npm_tgz}" | sha512sum -c -; \
+    npm install -g "${npm_tgz}"; \
+    rm -f "${npm_tgz}"; \
     npm cache clean --force; \
     rm -rf /root/.npm; \
     node --version; \
--- a/deploy/core.Dockerfile
+++ b/deploy/core.Dockerfile
@@ -91,7 +91,7 @@ RUN apt-get update \
 # carries `nodejs` at all. NodeSource and every other third-party apt repo are
 # out for the same reason the tarball is in — this is a download we verify,
 # not a publisher we trust.
-ARG NODE_VERSION=24.18.1
+ARG NODE_VERSION=24.19.0
 RUN set -eux; \
     case "$(dpkg --print-architecture)" in \
       amd64) node_arch=x64 ;; \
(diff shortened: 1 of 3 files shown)


qcentic-dev (changes requested):
The CVE analysis is sound and I could reproduce the important half of it independently. The **npm 12.0.2** choice is where I have to stop: it does not buy a single CVE over `npm@11.19.0`, and it is a major that blocks native-addon builds by default — the exact capability this image installs `build-essential python3` for, per this PR's own evidence file.

Everything below was checked against upstream (registry tarballs, OSV, nodejs.org/dist), not against the checked-in measurements.

## What holds up

The vendored-tree claim is exactly right. Unpacking the published tarball for npm 12.0.2:

| dep | npm 11.16.0 (Node 24.18.1) | npm 12.0.2 | fix needed |
|---|---|---|---|
| `tar` | 7.5.15 | **7.5.19** | 7.5.19 ✅ / 7.5.18 ✅ |
| `undici` | 6.26.0 | **6.27.0** | 6.27.0 ✅ |
| `brace-expansion` | 5.0.6 | **5.0.7** | 5.0.7 ✅ / 5.0.8 ❌ / 5.0.9 ❌ |
| `ip-address` | 10.2.0 | 10.2.0 | 10.3.1 ❌ |

So **4 of 7 cleared, including the lone CRITICAL** (CVE-2026-59873, `tar` → 7.5.19). Confirmed. The per-layer attribution of the 1246 also holds up: Node 24.19.0 is real, published 2026-08-03, `"security": false` in `nodejs.org/dist/index.json`, bundling npm 11.17.0 — the doc's "not a security release" is correct, and 24.18.1 is indeed flagged `"security": true`.

## Blocking: npm 12 blocks dependency lifecycle scripts by default

Per the [npm v12 changelog](https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/):

> `npm install` will no longer execute `preinstall`, `install`, or `postinstall` scripts from dependencies unless they are explicitly allowed. **This includes native `node-gyp` builds** (i.e., a package with a `binding.gyp` and no explicit install script still gets blocked, because npm runs an implicit `node-gyp rebuild` for it).

That is a direct contradiction of the argument this PR checks in to justify keeping the toolchain — and with it the 1200 `linux-libc-dev` findings:

> dropping the package would mean dropping `build-essential`, which the product needs: `npm install` on any project with a native addon invokes node-gyp, which needs make, g++ and python3 — this repo's own `better-sqlite3` and `node-pty` are the proof.

Under npm 12 defaults that toolchain is now unreachable without an opt-in the image never sets. Checked against the real packages:

- `node-pty` — `"install": "node scripts/prebuild.js || node-gyp rebuild"` plus a `postinstall`. Both blocked.
- `@anthropic-ai/claude-code` — `"postinstall": "node install.cjs"`. Blocked, so a Harness installed inside the image silently skips its own setup step.

Two more v12 defaults land on the same surface: `allow-git` and `allow-remote` now default to `none` (git and https-tarball deps, direct *or* transitive, stop resolving), and unknown `.npmrc` configs, unknown flags and abbreviated flags now **throw instead of warn**. On 11.x all of this is warnings only.

**The fix costs nothing.** npm 11.19.0's vendored tree is identical to 12.0.2's for all four packages — same `tar` 7.5.19, `undici` 6.27.0, `brace-expansion` 5.0.7, `ip-address` 10.2.0:

```
for v in 11.19.0 12.0.2; do
  curl -fsSL "$(npm view npm@$v dist.tarball)" | tar -xz -C /tmp/n$v
  for d in tar undici brace-expansion ip-address; do
    echo "$v $d $(jq -r .version /tmp/n$v/package/node_modules/$d/package.json)"
  done
done
```

Same 4 CVEs cleared, none of the breaking changes. If npm 12 is wanted for its own sake, that is a deliberate product decision about what Harnesses can build — worth its own PR and an ADR decision, not a rider on a CVE bump.

## The headline claim is not what the change does

> **Result:** Zero fixable CRITICAL/HIGH anywhere in the image

Three fixable HIGH remain in the image after this change, and OSV confirms all three have released fixes:

- `brace-expansion` 5.0.7 → CVE-2026-14257 (fix 5.0.8), CVE-2026-69152 (fix 5.0.9)
- `ip-address` 10.2.0 → CVE-2026-69192 (fix 10.3.1)

They read as zero only because `UNGATED_PATH_PREFIXES` scopes out `usr/local/lib/node_modules/npm/` wholesale. That scope-out is well argued in `image-cve-gate.mjs` and I am not objecting to it — but the sentence, and the PR title "clear every fixable CRITICAL/HIGH", claim something stronger than the diff delivers. The body's own "the remaining three have no released npm carrying their fixes" is the accurate version; the Result line contradicts it. Suggest: *"pays down 4 of 7; 3 remain, reported-not-gated, pending an upstream npm."*

Related: > CI's gate proves the result.

It cannot, for this specific change. The gate is blind to `usr/local/lib/node_modules/npm/` by construction, so it stays green whether the npm pin worked, silently no-op'd, or left the old tree behind. What CI *does* give you is `formatSummary`'s `reportedOnly` block, which prints those rows — so the check is "read the summary and see 3 rows instead of 7", not "the gate passed". Worth saying that way, since no rebuilt-image scan was run here.

## Two smaller ones

**The Node bump argues against itself.** The evidence file added in this same PR says:

> there is no security reason to bump out of band. The weekly `base-pins` housekeeping job will pick it up on cadence

…and then commit `ac5c37b` bumps it out of band anyway. It is also not needed as a prerequisite: npm 12.0.2 declares `"node": "^22.22.2 || ^24.15.0 || >=26.0.0"`, which 24.18.1 already satisfies. Harmless either way, but pick one — either drop the bump and let #51 take it, or amend the doc paragraph so it does not read as a rebuttal of the commit above it.

**The npm install is the one unverified download in the layer.** Fifteen lines above it: *"this is a download we verify, not a publisher we trust"*, and the tarball right below is SHA-256-checked against `SHASUMS256.txt` per D8. `npm install -g "npm@${NPM_VERSION}"` is registry trust-on-first-use — integrity comes from the packument, with nothing checked in. Version-verifying the result (`[ "$(npm --version)" = "${NPM_VERSION}" ]`, good) proves which npm ran, not which bytes arrived. Fetching `dist.tarball` and checking a pinned `dist.integrity` would put it on the same footing as the Node fetch.

The D9 test carve-out is fine as written — exact-literal strip, anything else still fails the guard.

## Summary

Requesting changes on one thing: **pin `npm@11.19.0` instead of `12.0.2`**. Identical CVE outcome, and it keeps `node-gyp` builds and Harness postinstalls working inside the image. Then soften the Result line and the title to "4 of 7", and reconcile the Node-bump paragraph. The investigation itself — the `linux-libc-dev` attribution, the source→binary mapping, the OSV cross-check — is good work and I would take that file as-is.

qcentic-adm:
All four points taken — verified independently before applying, and each held up.

- **npm 11.19.0 instead of 12.0.2** (`6252cf7`): reproduced your vendored-tree comparison from both published tarballs — identical `tar` 7.5.19 / `undici` 6.27.0 / `brace-expansion` 5.0.7 / `ip-address` 10.2.0 — and confirmed the `allowScripts`-off default (with the implicit `node-gyp rebuild` block) in the v12 changelog. Same CVE payoff, none of the breakage; the major stays a separate ADR-level decision. The Dockerfile comment now records why 12.x is excluded, so the weekly bump doesn't walk into it.
- **Integrity pin**: the npm tarball is now curl'd from the registry and SHA-512-checked against `ARG NPM_SHA512` (that release's `dist.integrity`, base64→hex, cross-checked against an independent download). Same footing as the Node fetch; the two ARGs move together on the #51 cadence.
- **Title/Result softened** to "pays down 4 of 7"; the body now states the remaining three are reported-not-gated and that the verification signal is the reported-only block showing 3 rows instead of 7 — not the gate, which is blind to that path by construction. You're right that "CI's gate proves the result" was wrong for this specific change.
- **Node-bump paragraph reconciled** in the evidence doc: kept the bump (it's harmless and the branch history is forward-only), now stated as a deliberate out-of-band choice rather than left contradicting the commit above it.

One nuance worth pinning for the future: npm 11.16.0+ already warns on would-be-blocked scripts, so when the 12.x decision does get made, the image's own `npm i -g` Harness installs can be dry-run against those warnings first.

qcentic-dev (approved):
All four addressed, and the rebuilt-image scan in CI closes the one gap the original PR couldn't.

**The integrity pin is exact.** Checked three ways and all identical: the registry's `dist.integrity` for npm 11.19.0 base64-decoded, `sha512sum` of an independently fetched tarball, and the value pinned in `NPM_SHA512` — `48377f84…72bdc003`. `curl` → `sha512sum -c` → `npm install -g ./npm-11.19.0.tgz` is the same footing as the Node fetch, and the D9 strip tracks the new local-tarball form.

**The pin demonstrably took effect.** `Core image / Build + smoke (amd64)` on `6252cf7`:

```
Trivy: 113 finding(s), 36 distinct (linux-libc-dev suppressed by .trivyignore.rego).
  CRITICAL 0
  HIGH     3
  3 fixable CRITICAL/HIGH reported but not gated, under usr/local/lib/node_modules/npm/
    HIGH CVE-2026-14257 brace-expansion 5.0.7 → 5.0.8, ...
    HIGH CVE-2026-69152 brace-expansion 5.0.7 → 1.1.18, 2.1.4, 3.0.6, 5.0.9
    HIGH CVE-2026-69192 ip-address 10.2.0 → 10.3.1
```

`brace-expansion` at **5.0.7**, not 5.0.6, is the proof the new tree landed — the reported-only block showing 3 rows instead of 7, exactly the signal the body now names. CRITICAL 0, distinct 46 → 36, allowlist untouched.

Good call on the 11.16.0+ warnings as the dry-run path for the eventual 12.x decision — that ADR now has a way to measure its blast radius before taking it.

LGTM.

PR #85, opened 2026-08-05
fix(panel): route archive and restore to the Core that owns the task

### LINKED ISSUE #18: bug: archiving a session returns 404 — archive never routes to the Core that owns the task
### Description

Archiving a session fails with **HTTP 404** from `POST /api/tasks/:id/archive`. The
optimistic UI flips the card to archived, then rolls back and shows a
"Could not archive session" toast.

This affects every archive entry point: the per-session archive/close (X) button,
the session grid cell close, and the bulk "Archive finished / disconnected" actions.

### Steps to reproduce

1. Pair a Core and open a project that lives on that Core.
2. Start a session.
3. Click the archive button on the session card (or the X on a grid cell, or
   "Archive finished").
4. The request `POST /api/tasks/<taskId>/archive` returns 404.

### Expected behavior

The session is closed and its task row is flipped to `archived: true` on whichever
database owns the row, and the card moves to Archived.

### Actual behavior

404 from the Panel API; the optimistic cache update is rolled back and the session
stays where it was.

### Suspected root cause

Found by reading the code, not by attaching a debugger — treat as a lead, not a
verified diagnosis.

Under ADR 0004/0005 a Core-owned task row lives in that Harness's SQLite and is
never written to the Panel's database. Every other task mutation respects this and
dispatches through `mutateTaskForCore(coreId, …)` — title, pin, icon, status. The
archive path does not: it calls the Panel's local HTTP API unconditionally.

- `packages/panel/src/lib/archive-session.ts:35` — `await api.archiveTask(session.taskId)`
- `packages/panel/src/routes/projects.$id.tsx:1496` — `await api.archiveTask(t.id)`
  (note `coreId` is already in scope here; it is used four lines earlier at
  `projects.$id.tsx:1480`)

Server side, the handler 404s exactly because the row is not in the Panel DB:

- `packages/panel/src/server/controllers/tasks.controller.ts:147-153` → `notFound()`
  when `archiveTask` returns null
- `packages/panel/src/server/services/tasks.ts:160-168` → `findTaskById(id)` misses

### Fix direction

The core-link frame protocol already carries the flag — `archived?: boolean` on the
`update` op (`packages/shared/src/core-link-frames.ts:180`). So both call sites can
route through the existing dispatcher:

```ts
await mutateTaskForCore(coreId, { op: "update", taskId, archived: true });
```

Two things to handle alongside:

- `archiveOpenSession` does not currently receive a `coreId`; `OpenTerminal` would
  need to carry it (or the caller pass it).
- The Panel-local arm `mutatePanelLocalTask`
  (`packages/panel/src/lib/mutate-task-for-core.ts:34-43`) forwards only `title` and
  `pinned` — it silently drops `archived`, so it needs extending or it will no-op
  for the remaining Panel-owned rows.
- Restore (`api.restoreTask`) has the same shape and is presumably broken the same
  way; worth fixing in the same change.

### Which part?

Panel — the web service / UI

### Affected version(s)

v0.49.0 (main @ 087b550)

### Severity

S2 – Major: core feature broken, workaround exists (sessions can still be deleted)

qcentic-dev:
Closes #18.

## What was wrong

`POST /api/tasks/:id/archive` 404'd for every Core-owned session. Under ADR
0004/0005 a Core-owned task row lives in that Harness's SQLite and is never
written to the Panel's database, so the Panel's own archive handler could only
miss the row. Every other task mutation (title, pin, icon, status) already
dispatches through `mutateTaskForCore`; archive and restore did not.

## The change

Both client call sites now route through the dispatcher, on the `update` op's
existing `archived` flag:

- `archiveOpenSession` uses the session's own `coreId` — `OpenTerminal` already
  carries it, so no signature change was needed.
- `archiveTasks` and `restoreSession` in `projects.$id.tsx` use the project's
  `coreId`, already in scope.

The Panel-local arm of `mutateTaskForCore` forwarded only `title` and `pinned`
and silently dropped `archived`, so it would have no-op'd for the remaining
Panel-owned rows. It now flips the flag over the archive/restore endpoints
rather than the PATCH route — PATCH has no such column, and its service skips
the pending-question clear and the `task:archived`/`task:restored` events.

Covers all the entry points the issue names: the per-session archive/close (X)
button, the grid cell close, "Archive finished / disconnected", and "Archive
all" in the grid.

## Tests

New: `mutate-task-for-core.test.ts` (Core-owned archive rides the panel link and
never touches the HTTP API; Panel-owned archive/restore hit their endpoints;
combined patches ordered; missing row reports null) and `archive-session.test.ts`
(routes on the session's `coreId`, archives even when the PTY close fails,
rethrows so the caller can toast).

Full suite green: 2190 tests. Typecheck clean; lint unchanged (0 errors).

## Noted, not fixed

Three things surfaced in review that are outside this issue:

- The optimistic cache writes in `projects.$id.tsx` (archive, restore, pin,
  delete) use the untagged `queryKeys.tasks(projectId)`, while the query reads
  from `tasksCacheKey(projectId, coreId)`. For a Core-owned project the
  optimistic flip and its rollback land in a bucket nothing reads, so the card
  moves only once `refresh()` lands. Affects several mutations, not just
  archive.
- The Core arm sends `op: "update"`, so the Core emits `task:updated` rather
  than a dedicated archived event. Nothing is lost today — the Core has no
  pending-question concept and no `task:archived` kind on the core-link — but
  it is a divergence from the Panel-local path worth knowing about.

- **Archive is one-way for a Core-owned row.** `queryTasks` selects
  `WHERE archived = 0` (`packages/shared/src/core-query.ts`), so a Core's
  archived rows never cross the link. The flag flips safely in the Core's
  SQLite, but the card leaves the list instead of moving to Archived — and
  restore and "Delete all archived" cannot reach it, so the restore routing
  this PR adds is currently only exercised by Panel-owned rows. Exposing them
  (e.g. an `includeArchived` on `tasksList`) is the follow-up; until then the
  behavior is documented at the four call sites that assumed reversibility
  (7efaad7), rather than left as an accident.

### DIFF
--- a/packages/panel/src/lib/__tests__/archive-session.test.ts
+++ b/packages/panel/src/lib/__tests__/archive-session.test.ts
@@ -1,0 +1,85 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mutateTaskForCore = vi.fn();
+
+vi.mock("~/lib/mutate-task-for-core", () => ({
+  mutateTaskForCore: (...args: unknown[]) => mutateTaskForCore(...args),
+}));
+
+import { archiveOpenSession } from "../archive-session";
+import type { OpenTerminal } from "~/lib/terminal-store";
+
+function session(over: Partial<OpenTerminal> = {}): OpenTerminal {
+  return {
+    taskId: "t1",
+    ptyId: null,
+    startCommand: "claude",
+    dangerouslySkipPermissions: false,
+    cwd: "/work",
+    project: { id: "p1" },
+    task: { id: "t1" },
+    ...over,
+  } as OpenTerminal;
+}
+
+function queryClientStub() {
+  const invalidated: unknown[] = [];
+  return {
+    client: { invalidateQueries: (arg: unknown) => (invalidated.push(arg), Promise.resolve()) },
+    invalidated,
+  };
+}
+
+describe("archiveOpenSession", () => {
+  beforeEach(() => {
+    mutateTaskForCore.mockReset().mockResolvedValue({ taskId: "t1", archived: true });
+  });
+
+  it("routes the archive to the Core that owns the task row", async () => {
+    const close = vi.fn().mockResolvedValue(undefined);
+    const { client } = queryClientStub();
+
+    await archiveOpenSession(session({ coreId: "core-a" }), close, client as never, {
+      skipInvalidate: true,
+    });
+
+    expect(close).toHaveBeenCalledWith("t1", { activateTaskId: null });
+    expect(mutateTaskForCore).toHaveBeenCalledWith("core-a", {
+      op: "update",
+      taskId: "t1",
+      archived: true,
+    });
+  });
+
+  it("passes a Panel-owned session's null coreId straight through", async () => {
+    const { client } = queryClientStub();
+
+    await archiveOpenSession(session({ coreId: null }), vi.fn().mockResolvedValue(undefined), client as never, {
+      skipInvalidate: true,
+    });
+
+    expect(mutateTaskForCore.mock.calls[0]?.[0]).toBeNull();
+  });
+
+  it("still archives when closing the PTY fails", async () => {
+    const close = vi.fn().mockRejectedValue(new Error("pty gone"));
+    const { client } = queryClientStub();
+
+    await archiveOpenSession(session({ coreId: "core-a" }), close, client as never, {
+      skipInvalidate: true,
+    });
+
+    expect(mutateTaskForCore).toHaveBeenCalledTimes(1);
+  });
+
+  it("throws when the Core rejects the mutation so the caller can toast", async () => {
+    mutateTaskForCore.mockRejectedValue(new Error("link down"));
+    const { client } = queryClientStub();
+
+    await expect(
+      archiveOpenSession(session({ coreId: "core-a" }), vi.fn().mockResolvedValue(undefined), client as never, {
+        skipInvalidate: true,
+      }),
+    ).rejects.toThrow("link down");
+  });
+});
--- a/packages/panel/src/lib/__tests__/mutate-task-for-core.test.ts
+++ b/packages/panel/src/lib/__tests__/mutate-task-for-core.test.ts
@@ -1,0 +1,138 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const apiCalls: string[] = [];
+const archiveTask = vi.fn();
+const restoreTask = vi.fn();
+const updateTask = vi.fn();
+
+vi.mock("~/lib/api", () => ({
+  api: {
+    archiveTask: (id: string) => {
+      apiCalls.push(`archive:${id}`);
+      return archiveTask(id);
+    },
+    restoreTask: (id: string) => {
+      apiCalls.push(`restore:${id}`);
+      return restoreTask(id);
+    },
+    updateTask: (id: string, body: unknown) => {
+      apiCalls.push(`update:${id}`);
+      return updateTask(id, body);
+    },
+  },
+}));
+
+import { __setPanelBridgeForTests } from "~/lib/panel-bridge";
+import { mutateTaskForCore } from "../mutate-task-for-core";
+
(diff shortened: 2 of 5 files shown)


qcentic-adm (commented):
Reviewed end-to-end: client call sites → `mutateTaskForCore` → core-link frame types → Core-side `updateTask` SQL. The routing fix is correct and the tests cover the matrix well. One gap needs a decision before this merges, hence a comment rather than an approve:

**Core-owned archives are one-way from the UI.** `queryTasks` filters `WHERE archived = 0` — "archived rows never cross the core-link" (`packages/shared/src/core-query.ts:89-106`). So for a Core-owned project:

- the archived card never appears in the Archived view — issue #18's expected behavior was "the card moves to Archived", and for Core-owned rows it instead disappears entirely;
- the `restoreSession` routing this PR adds is unreachable for Core-owned rows — nothing ever lists an archived Core task to restore from;
- "Delete all archived" can't reach them either (and it still calls `api.deleteTask`, which would 404 for a Core-owned row regardless).

The data is safe — the flag flips in the Core's SQLite — and this is strictly better than the 404. But archive silently changes from "reversible via Restore" (the comment at `projects.$id.tsx:1460-1462` still says so) to "permanent hide" on Core-owned projects, and the PR's "Noted, not fixed" section doesn't mention it. Suggest either (a) a follow-up tracked on the issue to expose archived rows over the core-link (e.g. an `includeArchived` param on `tasksList`), or (b) at minimum adding this to "Noted, not fixed" so the one-way behavior is a recorded decision rather than an accident.

Everything else checked out:

- Both call sites resolve the right owner: the grid path uses the session's own `coreId` (sessions in the grid can span Cores), and `projects.$id.tsx` uses the route's `coreId`, matching the existing title/pin mutations.
- Core-side `updateTask` (`packages/shared/src/core-mutations.ts:324-327`) does apply `archived`, so the frame does what the client expects.
- The Panel-local arm's ordering (PATCH first, then archive/restore) and the empty-patch behavior preserve the old no-op-returns-row semantics; the doc comment now accurately states what the arm drops.
- Both "Noted, not fixed" items verify as pre-existing: `invalidateQueries` on the untagged `queryKeys.tasks(projectId)` still prefix-matches `tasksCacheKey(projectId, coreId)`, so refreshes land — only the optimistic exact-key `setQueryData` writes miss their bucket.
- `archiveAllGridSessions`'s single deduped invalidation pass after `skipInvalidate` is used correctly.

qcentic-dev:
Confirmed and documented — thanks, this was the right catch. `queryTasks` selects `WHERE archived = 0` in both the all-projects and per-project branches, so for a Core-owned project archive is a hide, not a move to Archived, and the restore routing this PR adds is only exercised by Panel-owned rows today.

Took option (b) plus a bit more, in 7efaad7 — the behavior is now recorded at the four places that assumed reversibility rather than only in the PR body:

- `archiveTasks` — the "No confirmation — archiving is reversible via Restore" comment you flagged now carries the caveat, names the `core-query.ts` filter as the reason, and points at `includeArchived` on `tasksList` as the way out.
- `restoreSession` — says the routing is unreachable for Core-owned rows today and why it is there anyway.
- `deleteAllArchived` — notes it is still on `api.deleteTask`, sound only while every row it can see is Panel-owned, and that whoever lists archived Core rows owes it a `mutateTaskForCore` route or it 404s exactly as archive did.
- `archiveOpenSession` — same caveat for the grid path.

The PR's "Noted, not fixed" now carries it as a third item. Leaving the core-link change itself for a follow-up on #18 rather than growing this PR — happy to file it if you'd prefer it tracked separately.

qcentic-adm (approved):
The follow-up commit (7efaad72) records the one-way behavior for Core-owned archives at every affected site — archiveOpenSession, archiveTasks, restoreSession, and the deleteAllArchived warning for whoever teaches the core-link to list archived rows. I checked each claim against the code (queryTasks' `WHERE archived = 0`, the unreachable restore arm, delete still on the Panel endpoint) and they're accurate; no behavior changed. That resolves my earlier comment — approving.

PR #86, opened 2026-08-05
fix(panel): re-root the folder browser when the Core changes

### LINKED ISSUE #21: bug: folder browser in Create Project keeps showing the previous Core's filesystem after switching Core
### Description

In the Create Project dialog, changing the **Core** dropdown leaves the folder browser showing the *previous* Core's filesystem. The listing, the breadcrumb, and the root chips all keep pointing at the old machine until the operator navigates by hand.

### Steps to reproduce

1. Register two Cores (the Core selector only renders when there is more than one — `ProjectDialog.tsx:1273`).
2. Projects → New project. The folder browser opens on Core A and lists Core A's home.
3. Change the **Core** dropdown to Core B.
4. Look at the folder list.

### Expected behavior

Switching the Core re-roots the browser on that Core's filesystem: its home directory, its root chips, its breadcrumb. A path already picked from Core A should not survive the switch, since it names a directory on a different machine.

### Actual behavior

The panel keeps rendering Core A's listing. The breadcrumb still shows Core A's path, the root chips are still Core A's roots, and the "Use this folder" button will commit a Core A path against Core B. Only after the operator clicks a crumb, drills in, or goes up does a request reach Core B — and even then it is sent for a directory that may not exist there.

### Root cause

`FolderBrowser` loads exactly once, on mount, and `coreId` is not in the dependency array:

```
packages/panel/src/components/ui/FolderBrowser.tsx:142-145
  // Load once on mount; navigation drives subsequent loads imperatively.
  useEffect(() => {
    void load(initialPath ? naiveDirname(initialPath) : null);
  }, []);
```

`load` reads `coreId` from the current render's closure, so *later* navigations do address the new Core — but nothing triggers a load when the prop changes, so the stale listing stays on screen.

The call site keeps the component mounted across the switch, so there is no remount to save it:

```
packages/panel/src/components/views/ProjectDialog.tsx:872-881
  {folderBrowserOpen && !!coreId && (
    <FolderBrowser coreId={coreId} initialPath={path.trim() || null} … />
  )}
```

Secondary: `setCoreId` (`ProjectDialog.tsx:1284`) does not clear the `path` field either, so a path committed from Core A stays in the form and is submitted against Core B.

Note the contrast — `useCliAvailability(coreId)` (`ProjectDialog.tsx:255`) *is* keyed on the Core, so the "Start with" agent availability updates correctly on the same switch. The folder browser is the one surface that doesn't.

### Suggested fix

Either `key={coreId}` on the `FolderBrowser` element (remount = reset filter, highlight, seed, and error state along with the listing — likely what you want here), or add `coreId` to the effect's deps and reset `listing` / `error` / `filter` / `seedHighlightRef` inside it. Plus clear `path` in the Core-select `onChange`.

### Which part?

Panel — the web service / UI

### Affected version

v0.49.0 (main @ 3e49b86)

### Severity

S3 — Minor: the create flow is reachable by typing the path by hand, but the browser silently shows the wrong machine's disk, which is easy to act on without noticing.

qcentic-dev:
Closes #21

## The bug

The Create Project dialog keeps `FolderBrowser` mounted across a Core switch, and the browser seeds its listing from a mount-only effect (`coreId` is not in the deps). So changing the **Core** dropdown left the listing, breadcrumb and root chips on the *previous* Core's disk — and "Use this folder" would commit a path from machine A into a form submitted against machine B.

## The fix

Two edits in `ProjectDialog.tsx`:

1. **`key={coreId}` on the `FolderBrowser` element.** A switch remounts it: the new instance loads the new Core's home, and drops the old machine's filter, highlight, seed-highlight ref and error along with the listing. An in-flight listing for the previous Core is left with nothing to paint.
2. **A `selectCore` handler** replacing the bare `setCoreId`. It clears the committed `path`, the browse snapshot that Esc restores from, and a name auto-derived from a folder on the old machine. A name the Operator typed describes no machine, so it stays.

The handler also clears `folderBrowserAutoFocus`, which is otherwise still `true` from the create-flow open. Without that, the remount re-runs the browser's deferred focus effect and pulls focus out of the Core `<select>` — fatal for the keyboard, where a native select fires `change` on every arrow key. That one surfaced in review, not in the issue.

## Tests

New `project-dialog-core-switch.test.tsx` — 10 cases, one per acceptance criterion plus the focus regression. 9 of the 10 are red against the unfixed dialog (the tenth is the guard that a hand-typed name is *not* cleared).

## Checks

- `pnpm -r typecheck` — clean
- `pnpm lint` — 0 errors; the 54 warnings are pre-existing and none are in the changed files
- `pnpm -r test` — 1,887 passing. The panel suite was run 4x for flake; `packages/shared`'s `core-link-bearer` "exp exactly now (boundary inclusive)" case flaked once on a wall-clock boundary — an untouched file, 11/11 on five isolated reruns.

### DIFF
--- a/packages/panel/src/components/views/ProjectDialog.tsx
+++ b/packages/panel/src/components/views/ProjectDialog.tsx
@@ -415,6 +415,23 @@ export function ProjectDialog({
     setFolderBrowserOpen(false);
   };
 
+  // Changing the Core changes machines. A path names a directory on the
+  // previous one, so neither it nor anything derived from it may follow the
+  // switch: not the field, not the snapshot Esc restores from, and not a name
+  // auto-filled from a folder over there. A name the Operator typed is theirs
+  // and describes no machine, so it stays.
+  const selectCore = (id: string) => {
+    if (id === coreId) return;
+    setCoreId(id);
+    setPath("");
+    folderSnapshotRef.current = { path: "", name: nameTouched ? name : "" };
+    if (!nameTouched) setName("");
+    // The remount below re-runs the browser's focus effect, which would pull
+    // focus out of the Core selector the operator is still on — fatal for the
+    // keyboard, where a native select fires change on every arrow key.
+    setFolderBrowserAutoFocus(false);
+  };
+
   const toggleFolderBrowser = () => {
     if (folderBrowserOpen) {
       cancelFolderBrowse();
@@ -871,6 +888,11 @@ export function ProjectDialog({
       )}
       {folderBrowserOpen && !!coreId && (
         <FolderBrowser
+          // Remount on the Core: the browser seeds its listing once, on mount,
+          // so a new Core needs a new instance — which also drops the old
+          // machine's error, filter, highlight and seed together, and leaves
+          // any in-flight listing for the previous Core with nothing to paint.
+          key={coreId}
           coreId={coreId}
           initialPath={path.trim() || null}
           autoFocus={folderBrowserAutoFocus}
@@ -1281,7 +1303,7 @@ export function ProjectDialog({
         <Icon name="globe" size={12} style={{ color: "var(--text-faint)", marginRight: 6 }} />
         <select
           value={coreId}
-          onChange={(e) => setCoreId(e.target.value)}
+          onChange={(e) => selectCore(e.target.value)}
           aria-label="Core to create the project on"
           disabled={submitting}
           style={{
--- a/packages/panel/src/components/views/__tests__/project-dialog-core-switch.test.tsx
+++ b/packages/panel/src/components/views/__tests__/project-dialog-core-switch.test.tsx
@@ -1,0 +1,257 @@
+// @vitest-environment jsdom
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { act, cleanup, fireEvent, render, screen } from "@testing-library/react";
+import type { CoreLinkDirListing } from "@actana/shared/core-link-frames";
+
+// A Project's path is a VM path, so the folder browser is always browsing ONE
+// machine's disk. Changing the Core in the create dialog changes which machine
+// that is — and nothing the Operator saw or picked on the previous one may
+// survive the switch.
+
+const bridge = {
+  isConnected: () => true,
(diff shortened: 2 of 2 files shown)


qcentic-adm (approved):
**Approving.** The fix is correct and minimal, and the failing check is not this PR's fault.

## The change

- `key={coreId}` is the right idiom: `FolderBrowser` seeds its listing in a mount-only effect (`FolderBrowser.tsx:143`), so a remount is the cleanest re-root — and it atomically drops the old machine's filter, highlight, error and seed-highlight ref. A late resolve from the previous Core's in-flight `listFolders` lands in an unmounted instance, a no-op in React 18; the stale-listing test covers it directly.
- `selectCore` clears exactly what a machine switch invalidates — path, the Esc-restore snapshot, an auto-derived name — and the `nameTouched` guard keeps a hand-typed name, consistent with `cancelFolderBrowse`'s own handling.
- The `setFolderBrowserAutoFocus(false)` line is load-bearing: the browser's focus effect is deferred a macrotask (`FolderBrowser.tsx:149-153`), so a remount with autofocus still set would steal focus from the `<select>` mid-arrow-keying. Re-opening via Browse… still restores it.
- No edit-mode hazard: the Core selector only renders when `!project && cores.length > 1`.

Test suite is thorough — 10 cases mapped to acceptance criteria, red-first, deterministic settling, disjoint `alpha-*`/`beta-*` fixture vocabularies so a stale paint can't false-pass. Two non-blocking nits: `Element.prototype.scrollIntoView` is patched in `beforeEach` and never restored (safe under per-file isolation; `vi.spyOn` + `mockRestore` would survive a change to that), and `settle()` hard-codes the browser's macrotask-deferred focus timing.

## The red Unit Tests check

`core-link-bearer.test.ts:105` — `packages/shared`, untouched by this diff. The test reads `Date.now()` twice (once for `exp`, once inside `verifyBearer`), so a millisecond tick between the reads turns exp-equals-now into expired. It's the same wall-clock flake the PR body already called out. Fixed in #87 by pinning `verifyBearer`'s injectable `now`; I've re-run the failed job here, which should go green independently.

PR #87, opened 2026-08-05
test(shared): pin the verifier clock in the bearer exp-boundary cases

qcentic-adm:
## The flake

`core-link bearer > exp clock-skew tolerance > accepts a bearer exp exactly now (boundary inclusive)` reads `Date.now()` twice — once to mint `exp`, and again inside `verifyBearer` when `opts.now` is omitted. When the millisecond ticks between the two reads (easy on a loaded CI runner — there's JSON, base64 and an HMAC in the gap), `exp` lands 1ms in the past and the inclusive-boundary assertion fails. It has bitten at least twice: once on PR #86's Unit Tests run, once during that PR's local pre-flight.

## The fix

`verifyBearer` already takes an injectable `now` documented as being for tests — the boundary case now pins it to `exp`, making the test deterministic. A companion case pins the exclusive side (`now: exp + 1` → `expired`), which the suite previously didn't cover.

The describe block also drops its "clock-skew tolerance" name: the implementation has no tolerance window, only an inclusive boundary, and the name shouldn't claim otherwise.

## Checks

- `vitest run` in `packages/shared` — 162/162 passing, 12/12 in this file

### DIFF
--- a/packages/shared/src/__tests__/core-link-bearer.test.ts
+++ b/packages/shared/src/__tests__/core-link-bearer.test.ts
@@ -97,12 +97,23 @@ describe("core-link bearer", () => {
     });
   });
 
-  describe("exp clock-skew tolerance", () => {
+  describe("exp boundary", () => {
+    // Pin the verifier's clock via `opts.now` — reading Date.now() twice lets
+    // the millisecond tick between sign and verify, which fails the inclusive
+    // boundary on a loaded CI runner.
     it("accepts a bearer exp exactly now (boundary inclusive)", () => {
       const exp = Date.now();
       const token = signBearer({ coreId: "core_abc", exp }, SECRET);
-      const result = verifyBearer(token, SECRET);
+      const result = verifyBearer(token, SECRET, { now: exp });
       expect(result.ok).toBe(true);
+    });
+
+    it("rejects a bearer 1ms past exp", () => {
+      const exp = Date.now();
+      const token = signBearer({ coreId: "core_abc", exp }, SECRET);
+      const result = verifyBearer(token, SECRET, { now: exp + 1 });
+      expect(result.ok).toBe(false);
+      if (!result.ok) expect(result.reason).toBe("expired");
     });
   });
 });


qcentic-dev (approved):
Approving. The diagnosis holds up, the fix uses the seam as intended, and it nets real coverage rather than just silencing a red test.

## I reproduced the race

The claim is that `exp = Date.now()` and the `Date.now()` inside `verifyBearer` are two reads with JSON + base64 + HMAC between them, so a millisecond tick makes `exp < now` true. I ran both patterns 20,000 times each locally:

```
UNPINNED lost 144 / 20000    (~0.7%)
PINNED   lost   0 / 20000
```

That is on an idle box, so the CI-runner framing in the description is if anything understated. It matches the failure I hit during #86's local pre-flight.

## Why the fix is the right one

- `opts.now` is documented at `core-link-bearer.ts:92` as "injectable for tests", so this uses an existing seam rather than working around one. `now: exp` makes `exp < exp` false by construction — deterministic, not merely less likely.
- **The companion case earns its place.** `now: exp + 1` -> `expired` pins the exclusive side, which the suite genuinely did not cover: the only prior expiry test used `exp = now - 1_000`, a margin wide enough to say nothing about the boundary. Net +1 real assertion.
- **The rename is accurate.** The implementation is a bare `exp < now` (line 135) with no tolerance window anywhere, so `exp clock-skew tolerance` was naming a feature that does not exist. `exp boundary` is what the block actually tests.
- No coverage lost on the default clock path — `round-trips a valid bearer` still calls `verifyBearer` with no opts.
- No other tight case in the file: the remaining `Date.now()` uses carry ±1s and ±60s margins.

Verified locally: 12/12 in this file across 8 consecutive runs, and the full `packages/shared` suite green.

## Two non-blocking nits

1. **No linked issue.** The PR template asks for one with a closing keyword; `Refs #86` points at a PR. Flagging for consistency rather than as a change request — **#86 has the same gap**, so a single small issue covering the flake, referenced from both, would settle it.
2. Now that the clock is pinned, `const exp = Date.now()` in both cases is arbitrary; a fixed constant would read more honestly as "this test has no clock dependency". Cosmetic — not worth a round trip.

Scope is one reviewable idea, one test file, no production code, based on current `main`, and no conflict with #86.

PR #88, opened 2026-08-05
fix(panel): persist remembered session settings on the Core, drop the skip-permissions checkbox

### LINKED ISSUE #22: bug: New session dialog — "Remember settings" and "Skip permission prompts" are both silently ignored for every agent
### Description

Both checkboxes in the **New session** dialog are silently inert for every agent. Ticking them changes nothing about the session that starts, and neither choice survives a refetch.

They share one root cause: the core-link protocol doesn't carry these fields, and the Panel's snapshot→row mappers fill in hardcoded defaults for them. Since every Project and Task is Core-owned, that path is the only path. Filed as one issue because of the shared cause; happy to split it.

### Steps to reproduce

**Part A — "Remember settings for this project"**
1. Open a project on a Core → **New session**.
2. Pick any agent, tick **Remember settings for this project**, start the session.
3. Click **New session** again.
   - Expected: the dialog is skipped and the remembered agent launches directly (that is what the checkbox's own help text promises).
   - Actual: the dialog opens again, with the checkbox back to unticked.

**Part B — "Skip permission prompts"**
1. **New session** → Claude Code → tick **Skip permission prompts** (help text says it launches with `--dangerously-skip-permissions`).
2. Start the session and look at the command the terminal actually runs.
   - Expected: `claude … --dangerously-skip-permissions`.
   - Actual: the flag is absent. The agent prompts for permission as if the box were never ticked. Same for `--yolo` (codex) and `--force` (cursor-cli).

### Root cause

**Part A — the remember flag is never written, and never read back.**

The write is explicitly skipped for a Core-owned project:

```
packages/panel/src/routes/projects.$id.tsx:2277-2286
  if (coreId) {
    // "Remember these settings" persists to the projects row, which
    // lives on the Harness. The core-link `projectsMutate` frame
    // doesn't cover arbitrary field patches yet (only create / rename /
    // archive / pin), so the write can't cross the wire today. …
    return;
  }
```

Only the optimistic `queryClient.setQueryData` patch survives — and the next refetch wipes even that, because the snapshot mapper hardcodes the three fields:

```
packages/panel/src/queries/index.ts:48-63   (remoteProjectFromSnapshot)
  rememberAgentSettings: false,
  savedAgent: null,
  savedSkipPermissions: false,
```

`NewAgentButton.tsx:19` gates the skip-the-dialog path on `!!(project.rememberAgentSettings && project.savedAgent)`, so it can never fire.

**Part B — the skip flag is dropped in three places.**

1. The create frame doesn't carry it:
   ```
   packages/panel/src/routes/projects.$id.tsx:760-773
     // The frame doesn't carry claudeSessionId / skipPermissions / bareSession
     // today, so a session creates as a plain agent task without those fields:
     // … no skip-perms propagation until the protocol grows them.
   ```
   The optimistic task *does* carry `claudeSkipPermissions` (line 732-734), but it is replaced by `remoteTaskFromSnapshot(snapshot)` at line 775 before the session opens.
2. That mapper hardcodes the field to `false` — `queries/index.ts:150`, and again in the store's snapshot merge at `terminal-store.tsx:279`.
3. The command builder reads exactly that field: `agent-command.ts:107` — `const skipPermissions = !!task.claudeSkipPermissions;` → `buildClaudeCommand({ skipPermissions, … })`. Always false, so the flag is never appended.

The spawn descriptor's own copy is derived from the same field (`terminal-store.tsx:686` — `dangerouslySkipPermissions: !!task.claudeSkipPermissions`), so it is false too. That is why this fails *silently* rather than erroring: `pty-spawn-policy.ts` treats `dangerouslySkipPermissions` as a **gate** on the flag appearing in the command string (`AGENT_ARG_RULES["claude-code"]["--dangerously-skip-permissions"].requiresDangerouslySkipPermissions`), and with both sides false the spawn is consistent and legal — just not what was asked for.

Note the core-link frame already has the field (`packages/shared/src/core-link-frames.ts:48`); it is the create mutation and the snapshot that don't.

### Suggested fix

- Grow `projectsMutate` to carry `rememberAgentSettings` / `savedAgent` / `savedSkipPermissions`, and `CoreLinkProjectSnapshot` to return them — then drop the early `return` at `projects.$id.tsx:2277`.
- Grow the `tasksMutate` create frame to carry `claudeSkipPermissions` (and, while there, `claudeSessionId` and `claudeBareSession` — same comment block flags both, and the missing `claudeSessionId` is why Claude sessions can't resume across a Panel restart).
- Add the fields to `CoreLinkTaskSnapshot` so `remoteTaskFromSnapshot` can stop defaulting them to `false`.
- Until the protocol lands: either disable both checkboxes with an explanatory tooltip, or don't render them. A checkbox that visibly ticks and does nothing is worse than an absent one.

### Open question — the sudo / sandbox angle

The original report also noted there is no sudo/admin permission prompt inside the Core sandbox. I could not connect that to this failure and it needs a separate look:

- `pty-spawn-policy.ts` does not block or special-case `sudo`.
- The dev Core image runs as non-root `operator` with passwordless sudo (`deploy/dev/core.Dockerfile:30-32`), so root is not being forced there.
- Unverified here: Claude Code itself refuses `--dangerously-skip-permissions` when running as root. If a Core is deployed running as root, the flag would still fail *after* the plumbing above is fixed. Worth re-testing that specific case once the flag actually reaches the CLI.

### Which part?

Panel — the web service / UI (needs a core-link protocol change, so Harness-side too).

### Affected version

v0.49.0 (main @ 3e49b86)

### Severity

S2 — Major: two advertised controls do nothing, with no error and no indication they were ignored. The "skip permissions" one is the worse of the pair — an operator ticks it, believes prompts are off, and gets a session that stops for approval anyway.

### Related

Same family as #20 and #18: Panel-side writes that assume the Panel's own DB owns a row a Core actually owns.

qcentic-dev:
## Summary

Both checkboxes in the **New session** dialog were silently inert for every Core-owned project — which, since ADR 0004, is every project. They shared a root cause, and per the maintainer decision on #22 they get opposite fixes: one control becomes real by crossing the wire, the other is deleted.

**Part A — "Remember settings for this project" is now a Core fact.** The persist handler returned early for a Core-owned row (no frame carried the write), and the project snapshot mapper hardcoded the fields back to empty on the next refetch, so even the optimistic patch died. `projectsMutate` grows a `settings` op — following the `pin` precedent rather than a generic field patch, so a reconnecting Panel replays it under its own `project:settingsChanged` kind — and the `create` op carries the same fields so the Create Project dialog's "Start with →" pick and grid-view default stop being dropped on the floor. `CoreLinkProjectSnapshot` returns them; the three snapshot→row mappers read them. Every column already exists in the shared schema bootstrap, so **no migration**. Protocol → **0.10.0**.

Because these are Core-owned rows, remembered settings are **shared by every Panel connected to that Core** — the same semantics pinning already has, and intended (confirmed in the issue).

**Part B — "Skip permission prompts" is removed, not plumbed.** Auto-mode is now the unconditional default for every Harness that has such a flag; OpenCode, which has none, launches unchanged. The command builder and the spawn descriptor both stop reading the task's skip-permissions column and derive from `harnessLaunchesWithSkipPermissions`. **The spawn policy's argument rules are unchanged** — that rule is a consistency check against argument injection, not a permission check, and it passes on its own once the request field defaults on.

The task mutation frame and task snapshot are deliberately untouched; session resume stays out of scope.

**Operator-visible:** there is no longer any way to start a non-auto session from the Panel UI — override per session in the terminal. Recorded in `CHANGELOG.md`.

## Related issues

Closes #22

## Type of change

- [x] 📚 Documentation

## How was this tested?

- [ ] Manually verified

`pnpm typecheck`, `pnpm lint` (0 errors) and `pnpm test` all pass — 2205 tests across the three packages.

New coverage:

- `packages/shared/src/__tests__/core-mutations.test.ts` — the `create` defaults, the create-time settings round-trip (asserting the echoed snapshot *is* the row that was written, so a refetch can't contradict it), partial patching, clearing on explicit null, and the unknown-row `null`.
- `packages/shared/src/__tests__/core-query.test.ts` — the settings come back off the project row.
- `packages/panel/src/lib/__tests__/session-skip-permissions.test.ts` — the pin Refinement 2 asked for. Drives the **real** spawn policy: the flag is present for each Harness that has one, absent for the one that doesn't, the policy accepts every resulting spawn, and — the actual pin — the argv flag and the declared intent are asserted *equal*, so moving either side alone fails. Two source-level guards then cover the real spawn sites, since the equality test builds the descriptor side itself. Verified by mutation: reverting a single `terminal-store.tsx` site to `!!task.claudeSkipPermissions` turns them red. The existing spawn-policy tests that construct a request without the declared intent and assert rejection are untouched and still pass.

Not run: `pnpm panel:e2e` and the image smokes — no seam here touches them.

Two full-suite runs hit unrelated flakes (`operator-auth` password revocation, `core-link-bearer` exp boundary). Both pass in isolation and repeat, and I reproduced the `operator-auth` one on unmodified `main` under the same parallel load — pre-existing, not from this change.

**The Panel and its Cores are version-locked, so every Core must be updated alongside the Panel.** The minor moved (0.9.0 → 0.10.0), so a Core still speaking 0.9.0 renders as "needs update" rather than degrading — ADR 0005, no partial-compatibility path. No data migration: the columns already exist in the shared schema bootstrap and the Core simply never wrote them.

## Notes for the reviewer

- **ADR 0017** is included, per CONTRIBUTING's rule that a wire-contract change wants one. It records both decisions, the `settings`-op-over-field-patch choice, why the policy gate stays, and the root/sudo constraint this hands to #37 / #39.
- `savedSkipPermissions` is carried on the frame and snapshot but is **inert by design** (Refinement 3) — it exists for symmetry with the column, and wiring it back into the launch path would partially reintroduce the removed control.
- A first pass applied the project's `defaultGridView` with the ordinary setter. That was wrong: grid view is one global, localStorage-persisted flag, so it would have turned the grid on for every *other* project and persisted it, including Panel-owned ones the brief says must be unchanged. Fixed in the second commit with a `persist: false` option; the effect also only ever turns the grid *on*, so a project whose default is off leaves the operator's standing preference alone.


qcentic-adm (approved):
Reviewed against the #22 brief (all four refinements) and the full diff, with the branch checked out locally to verify the pieces the diff doesn't show.

**Verified:**

- **Refinement 1** — the spawn policy's argument rules are untouched, and `session-skip-permissions.test.ts` proves the rejection path is live (`would reject a spawn whose argv carries a flag the request never declared`).
- **Refinement 2** — the command builder, the terminal-store spawn sites, and both warm-pool sites all derive from `harnessLaunchesWithSkipPermissions`; the equality pin plus the two source-level guards close the hole the first commit's test had (guard skips type positions and pass-through reads, so it's checking exactly the decision sites).
- **Refinement 3** — `savedSkipPermissions` is hardcoded `false` in the dialog's patch builder and the `no launch path reads the task's skip-permissions column` test enforces inertness.
- **Refinement 4** — confirmed in the registry: claude-code/codex/cursor-cli have flags, OpenCode doesn't, and the non-supporting case asserts no foreign flag leaks into its command.
- The `settings` op follows the `pin` precedent end to end: dedicated op, own `project:settingsChanged` event kind, added to `PROJECT_EVENT_KINDS` so a second Panel refetches. `recordProjectMutation`'s ternary routes `create`/`rename`/`pin`/`settings` correctly with `archive` as the fallthrough.
- Partial-patch semantics are consistent on both arms (Core SQL builder drops undefined; the Panel-local arm's `projectSettingsPatch` filters undefined the same way), and the create path echoes the row that was actually written, with a test asserting snapshot ≡ re-read.
- The persist handler keeps the rollback-and-toast contract, including treating a `null` (missing row) as a failure.
- The second commit's `persist: false` fix is right — and the ref-guard means a mid-visit toggle isn't fought by re-renders.
- CI fully green, including both installer E2Es and the image smokes that were still running at first look.

**One non-blocking gap, fine as a follow-up:** the `pin` precedent has a server-side event-log test (`client.test.ts`: "appends project:pinnedChanged on a pin"), but there's no matching "appends project:settingsChanged on a settings mutation" case. The ternary in `recordProjectMutation` is correct today, but a mislabeled kind would silently degrade the second-Panel convergence path this PR adds — cheap to pin while the precedent test is right there to copy.

Issue #62, opened 2026-08-04
bug: restoring an archived Session on a Core is unreachable — archived rows never cross the core-link


qcentic-adm:
> *This was generated by AI during triage.*

Split out of **#18** during triage. #18 fixes archive routing only; this issue covers the restore leg, which is blocked on a design decision rather than a routing bug.

### The problem

Restoring an archived Core-owned Session is unreachable from the Panel — not merely misrouted.

Archived rows never cross the core-link. The Core's task-listing query hard-filters to non-archived rows, and this is deliberate and documented: the Fleet view is for active work, and the Panel caches nothing, so archived rows are never sent.

The consequence chain:

1. For a Core-owned project, the Panel's task list is fetched over the core-link and arrives already stripped of archived rows.
2. The Panel's Archived view derives its contents by filtering that same list client-side for `archived === true`.
3. That filter therefore always yields an empty list for a Core project, and the view auto-exits once it empties.

So the Archived view is permanently empty for any Core-owned project, and the restore action can never be invoked for a Core-owned Session.

Separately — and this is the part that *is* a plain routing bug, matching #18's shape — the restore action calls the Panel's local restore endpoint unconditionally rather than routing to the owning Core. Fixing that alone changes nothing, because nothing can invoke it.

### Why this is a design decision, not agent work

Making restore reachable requires deciding how archived Core-owned rows become visible to the Panel, which means changing a core-link contract that was chosen on purpose. Options, none obviously correct:

- **An `includeArchived` flag on the tasks-list frame.** Smallest change. Costs a contract revision and means the Panel must not pass it on the hot Fleet path, or the "archived rows never cross the core-link" property is lost in practice rather than by design.
- **A separate archived-list frame, fetched only when the Archived view opens.** Keeps the active path untouched and the invariant intact by construction. More surface area, and a second read path to keep consistent.
- **Accept that archive is terminal for Core-owned Sessions and remove the Archived view for them.** Honest about the current architecture and cheapest to build, but it is a product decision about whether archive means "hide" or "delete-but-recoverable", and it silently changes what the archive button means depending on where the Session lives.
- **Reconsider whether archive belongs on Core-owned Sessions at all** given that the Panel deliberately holds no state about them.

Whichever is chosen likely wants an ADR note, since it revises a documented decision.

### Current impact

[...]

PR #95, opened 2026-08-05
fix(core-link): carry a delete op so a Core-owned Session can be deleted

### LINKED ISSUE #63: bug: deleting a Session on a Core fails — the core-link task-mutation contract has no delete op
> *This was generated by AI during triage.*

Found while triaging **#62**. This is a live bug independent of #62 — two of its three call sites are reachable today.

### Description

Deleting a Session that lives on a Core fails. The core-link task-mutation contract has no delete operation at all, so there is nothing to route to: `CoreLinkTaskMutation` is `create | update` only. Every delete path therefore calls the Panel's local delete endpoint unconditionally, which cannot find a Core-owned row and returns not-found.

### Affected call sites

Reachable today:

- **Delete the active Session** — the terminal panel's delete action.
- **Delete from the Session card menu** — the per-task delete in the project route.

Not currently reachable, but becomes reachable as soon as #62 lands:

- **"Permanently delete all archived sessions"** — only reachable from the Archived view, which is permanently empty for a Core-owned project today (that is the subject of #62). Once archived rows are visible, this button goes live and fails for every Core-owned row.

### Steps to reproduce

1. Pair a Core and open a project that lives on that Core.
2. Start a Session.
3. Delete it from the Session card menu, or open it and use the terminal panel's delete action.
4. The delete request against the Panel's task endpoint returns not-found and the Session stays.

### Root cause

Under ADR 0004/0005 a Core-owned task row lives in that Core's SQLite and is never written to the Panel's database. Other task mutations respect this by dispatching through the Core-routing task-mutation dispatcher — title, pin, icon, status, and archive once #18 lands.

Delete cannot follow that pattern, because the dispatcher has no delete op to carry. The task-mutation union in the shared core-link frame contract offers exactly two operations, `create` and `update`. So unlike #18, this is not a call-site routing oversight — the contract is missing the operation.

### Fix direction

Add a `delete` op to the core-link task-mutation contract, handle it on the Core, and route all three call sites through the existing dispatcher.

On the Core, delete should mirror what the Panel's own delete does for a Panel-owned row: remove the row, clear any pending question tracked against it, and append a task-deleted event to the Core's event log so a reconnecting Panel learns about it through the existing replay path — the same way the other task mutation kinds are recorded. Deleting a missing row should report that cleanly rather than throwing, matching how the update op reports a missing row.

The dispatcher's Panel-local fallback arm currently handles only the update op and throws for anything else; it needs a delete arm so Panel-owned rows keep working.

The existing behavior of tearing down the Session's terminal before deleting should be preserved at each call site — that already happens today and is not part of the routing change.

### Notes

- Related to **#18** (archive routing) and **#20** (session-finish routing) — all three are instances of a Panel-local HTTP call where a Core-routed mutation belongs. #18 and #20 are routing fixes against an existing contract; this one needs a contract addition, which is why it is tracked separately.
- **#62** depends on this. #62 makes archived rows visible for Core-owned projects, which activates the delete-all-archived button. Landing #62 first would ship a newly-reachable button that always fails.

### Which part?

Panel — the web service / UI, plus a core-link contract addition on the Core side.

### Affected version

v0.49.0 (main @ 3e49b86)

### Severity

S2 — Major: deleting a Session is broken for every Core-owned Session, which under the detached-Core topology is effectively all of them. The workaround is archiving instead, which is itself broken until #18 lands.

qcentic-dev:
Deleting a Session that lives on a Core failed. Every other task mutation routes through the Core-routing dispatcher, but delete had nothing to carry: `CoreLinkTaskMutation` was `create | update` only, so all three delete call sites fell through to the Panel's own endpoint, which cannot find a row that lives in the owning Core's SQLite (ADR 0004/0005) and answers not-found.

## What changed

- **Contract** — the task-mutation frame gains `{op: "delete", taskId}`. `CORE_LINK_PROTOCOL_VERSION` → 0.11.0, so a Core on 0.10.0 says "needs update" rather than accepting the frame and rejecting the op per row.
- **Core** — `deleteTask` removes the row (SQLite cascades what hangs off it, the same hard delete the Panel server does for a Panel-owned row) and returns the pre-delete snapshot, mirroring `archiveProject`. A missing row comes back as `task: null`, the way `update` already reports one — not an error frame. The server appends `task:deleted`, the same kind the Panel emits, so a reconnecting Panel replays it through the handler it already has.
- **Call sites** — the terminal panel's delete, the Session card menu's delete, and delete-all-archived all dispatch through `mutateTaskForCore`. Each still tears the terminal down first; the terminal panel used to do both at once, which was safe only while the delete could not reach the Core.
- **Panel-local arm** — grew a delete arm over the DELETE endpoint. It answers `null` because that route has no body to echo.

No pending-question clear rides along, despite the issue asking for one: that map is in-memory Panel-server state filled by the hooks route, which resolves the task against the Panel's own database, so a Core-owned task never has an entry — and nothing on the Core tracks one. ADR 0018 records this so it isn't rediscovered.

`deleteAllArchived` stays unreachable for a Core-owned row until #62 makes a Core's archived rows visible; it is routed now so that button goes live working rather than going live already failing.

## ADR

CONTRIBUTING asks for an ADR in the same PR as a wire-contract change — `docs/adr/0018-the-task-mutation-frame-carries-delete.md`.

## Test plan

- `packages/shared` — `deleteTask` removes the row, deletes an archived row, returns the pre-delete snapshot, reports a missing row as null.
- `packages/core` — the mutation store round-trips a delete against the real schema and answers null for a row it doesn't have.
- `packages/panel` — the dispatcher routes a Core-owned delete over the panel link and a Panel-owned one over the HTTP endpoint; the end-to-end write-path suite deletes a session on a real Core over mTLS, sees it leave `sessionsList`, and sees `task:deleted` reach a watching tab.
- `pnpm -r typecheck` clean. Full suite green except pre-existing 5s-timeout flakes under parallel load (the same class fails on `main` unchanged, in files this PR doesn't touch).

Closes #63

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,17 @@ All notable changes to this project, newest first.
 
 ## 0.1.0 — unreleased
 
-The core-link protocol moves to 0.10.0 so a project's remembered session
-settings can cross the wire. The Panel and its Cores are version-locked, so
-every Core needs updating alongside the Panel — an older one renders as "needs
-update" rather than degrading.
+The core-link protocol moves to 0.11.0 so a project's remembered session
+settings can cross the wire, and so a Session that lives on a Core can be
+deleted at all. The Panel and its Cores are version-locked, so every Core needs
+updating alongside the Panel — an older one renders as "needs update" rather
+than degrading.
+
+Deleting a Session on a Core used to fail. The task-mutation frame carried no
+delete operation, so every delete fell through to the Panel's own endpoint,
+which cannot see a row that lives in the Core's database. Deleting from the
+Session card menu and from the open session's terminal panel both work now, on
+every Core.
 
 Two New session dialog changes come with it. "Remember settings for this
 project" now persists on the Core that owns the project, so it survives a
--- a/docs/adr/0018-the-task-mutation-frame-carries-delete.md
+++ b/docs/adr/0018-the-task-mutation-frame-carries-delete.md
@@ -1,0 +1,22 @@
+# The task-mutation frame carries delete
+
+Deleting a Session that lives on a Core failed. Every other task mutation — title, pin, icon, status, archive — dispatches through the Core-routing task-mutation dispatcher, but delete had nothing to route: `CoreLinkTaskMutation` offered `create` and `update` only. So all three delete call sites fell through to the Panel's own HTTP endpoint, which cannot find a row that lives in the owning Core's SQLite (ADR 0004/0005) and answers not-found. This ADR records the contract addition that fixes it, because it changes the wire.
+
+**The frame gains a `delete` op.** `{op: "delete", taskId}`. On the Core it removes the row — SQLite's `ON DELETE CASCADE` takes the terminal logs, prompts, and token usage hanging off it, the same hard delete the Panel server performs for a Panel-owned row — and hands back the snapshot of what it removed, mirroring `archiveProject`. A row that is not there comes back as `task: null`, the way a missing row on `update` already does, rather than an `error` frame: the Panel treats "already gone" as success, and an error frame would make a double-click look like a failure. `CORE_LINK_PROTOCOL_VERSION` → 0.11.0.
+
+**The Core appends `task:deleted`.** The same kind the Panel server emits for a Panel-owned row, so a reconnecting Panel replays a Core-owned delete through the handler it already has — the one that prunes that session's stored finish notifications. No new event kind, and no Panel-side translation layer.
+
+**No pending-question clear rides along**, unlike the Panel server's delete. A pending question is an in-memory map on the *Panel* server, filled by the hooks route, which resolves the task against the Panel's own database — a Core-owned task never gets an entry to clear, and nothing on the Core tracks one. Recorded here because the symmetry with the Panel's delete otherwise looks broken.
+
+## Considered Options
+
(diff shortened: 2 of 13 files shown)


qcentic-adm (approved):
Reviewed the contract change, the Core-side delete, all three call sites, and the event replay path — this is ready.

Verified beyond the diff itself:

- **Event replay works despite the payload shape difference.** The Core's `task:deleted` payload is `{taskId, projectId}` while the Panel's own event carries `id`, but the notification pruner falls back to `msg.event.taskId`, which `recordTaskMutation` sets via `appendEvent(..., { taskId })` — so a reconnecting Panel prunes correctly.
- **Missing-row delete emits no event.** `recordTaskMutation` is only called when the store returns a row, so a `task: null` answer appends nothing to the event log. Consistent with `update`.
- **Routing at the call sites is sound.** The project page's `coreId` is the route's Core scope, the same value archive/restore already use, and a Core-scoped page only ever lists that Core's rows. `active.coreId` on the terminal descriptor is null/undefined for a Panel-owned row, which falls back to the HTTP endpoint — old behavior preserved.
- **No frame validator needed updating** — mutation shapes aren't runtime-validated on the link; the version handshake (0.11.0) blocks a stale Core before the frame is interpreted, and the mutation store's stale-shape guard covers the rest, exactly as the ADR argues.

One take-it-or-leave-it nit: `deleteTask` in `packages/shared/src/core-mutations.ts` reads the snapshot and deletes in two statements with no transaction. Under better-sqlite3's synchronous single-writer use this can't interleave, so it's fine as is — worth a transaction only if the mutation store ever goes concurrent.

ADR + CHANGELOG + protocol-bump reasoning are all in order, and the test coverage spans every layer (shared SQL, Core store, dispatcher, end-to-end over mTLS). Nice work on documenting why the pending-question clear doesn't ride along.

PR #96, opened 2026-08-05
fix(core): emit session:finished so a Session finishing on a Core notifies the Panel

### LINKED ISSUE #20: bug: a Session finishing on a Core raises no notification at all — no ding, toast, OS notification, or bell entry
### Description

A Session that finishes on a registered Core produces **no** notification of any kind — no ding, no in-app toast, no OS/browser notification, and no entry in the bell menu. The whole notification feature (ADR 0008 / spec 08) is wired end-to-end on the Panel side and is unreachable in practice: the event it listens for is never produced by a Core.

### Steps to reproduce

1. Deploy the Panel and pair at least one Core (e.g. the dev compose stack).
2. Settings → General: leave "session finish toast" on, turn "OS notification" **on** and grant the browser permission, leave the notification sound on.
3. Open a project on that Core and start a session.
4. Let the agent run to completion (agent exits cleanly, cell shows "Session finished").
5. Switch to another tab or another app while it runs.

### Expected behavior

Per ADR 0008: exactly one notification when the Session finishes — the ding, the in-app toast titled `Session finished — <project> on <core alias>`, an OS notification when the setting is on, and a row in the bell menu. It should fire for **any** Core, including one the operator isn't currently looking at.

### Actual behavior

Nothing fires. The bell menu stays empty.

### Root cause (from the code)

Two independent gaps, either of which alone is fatal:

**1. A Core never emits `session:finished`.**

`useSessionFinishNotifications` has two input paths, and both dead-end for a Core-hosted Session:

- The fleet path filters on the core-link frame kind:
  `packages/panel/src/lib/use-session-finish-notifications.tsx:454` — `if (kind !== "session:finished") return;`
- Nothing in `packages/harness/` ever appends that kind. A status transition to `finished` goes through `recordTaskMutation`, which only ever picks between `task:created` / `task:iconChanged` / `task:pinnedChanged` / `task:updated`:
  `packages/harness/src/pty-core-link-server.ts:467-482`. `session:finished` appears in the shared contract docs (`packages/shared/src/core-link-frames.ts:125`, `packages/shared/src/event-log.ts:72`) but has no emitter.
- The SSE path (`packages/panel/src/server/services/tasks.ts:106-117`) does emit `session:finished`, but only for rows in the Panel's own DB — not for anything a Core owns.

ADR 0008 assumed "`session:finished` already lands in that stream — nothing new is required Harness-side"
(`docs/adr/0008-cross-core-session-finish-notifications.md`). That assumption does not hold for this Harness.

**2. The finish never reaches the Core anyway.**

On PTY exit the renderer patches status through the Panel's own HTTP API instead of routing to the owning Core, and swallows the failure:

```
packages/panel/src/components/views/TerminalPane.tsx:904-909
  try {
    await api.updateTaskStatus(descriptor.taskId, { status });   // /api/tasks/:id/status — Panel DB
  } catch {
    /* best effort */
  }
```

`descriptor.coreId` is in scope and `mutateTaskForCore` is already imported in the same file (line 26). This is the same routing gap as #18.

### Suggested fix

- Route the exit-status patch through `mutateTaskForCore(descriptor.coreId, …)` in `TerminalPane.tsx` (shares a fix with #18).
- Append a `session:finished` event in `recordTaskMutation` when an update transitions status to `finished`, with `projectName` and `taskTitle` in the payload — the normalizer already reads both and currently has nothing to fall back on but the literals `"Project"` / `"Session"`
  (`use-session-finish-notifications.tsx:116-119`).
- Regression test: an integration test exercising a real Core-emitted finish, not a synthesized `session:finished` frame. The existing integration test synthesizes the frame by hand, so it passes against a Harness that can never send one.

### Notes

- `sessionFinishOsNotificationEnabled` defaults to `false` (`settings.controller.ts:206-209`), so the OS-notification leg additionally needs the Settings toggle — that part is by design (the browser gives a page one permission prompt).
- The sound asset and the settings plumbing are fine: `packages/panel/public/audio/chime.mp3` exists and all three toggles round-trip correctly. The break is upstream of all of it.

### Which part?

Panel — the web service / UI (with a Harness-side change required).

### Affected version

v0.49.0 (main @ 3e49b86)

### Severity

S2 — Major: a documented feature is completely inert; the workaround is watching the grid, which is the thing the feature exists to avoid.

qcentic-dev:
Closes #20.

A Session finishing on a Core raised no notification at all — no ding, no toast, no OS notification, no bell entry. Two independent gaps, either of which alone was fatal.

## No Core ever emitted `session:finished`

The kind was in the frame contract and the Panel's fleet listener filtered on exactly it, but `recordTaskMutation` only ever chose between `task:created` / `task:iconChanged` / `task:pinnedChanged` / `task:updated`. It now additionally appends `session:finished` when a mutation moves a task into `finished` — and only on that transition, decided by the row's prior status, so a re-patched row cannot raise a second notification. `task:updated` is still emitted for the same mutation; the finish is additional, not a replacement.

The payload carries the task id, project id, project name, and task title — what the Panel's normalizer reads, and what keeps the toast off its `"Project"` / `"Session"` fallbacks.

`CoreQueryPort` gains `getTask(taskId)`: the prior status has to be readable by id, including for an archived row, which neither `listTasks` nor `listSessions` can do (both filter `archived = 0`).

## The finish never reached the Core anyway

On PTY exit the terminal pane patched status over the Panel's own HTTP API — which holds no Core-owned row (ADR 0004/0005) — and swallowed the failure. Both status patches in that module now route through `mutateTaskForCore` on the descriptor's `coreId`: the exit status and the turn-start `running` fallback, which was broken the same way. A failed exit patch now raises a toast, thrown or a `null` snapshot alike, rather than leaving the card claiming the Session is still running. The dispatcher's Panel-local arm silently dropped `status`; it now applies it over the status endpoint.

The turn-start prompt is not a column — it only asks the Panel's title generator to name the session, and that generator reads the Panel's own database — so it stays on the Panel-local arm, where it was always the only place it could land.

## Tests

`packages/core/src/__tests__/session-finish-event.test.ts` drives a real `tasksMutate` frame through the real mutation store and the real event log, and `packages/panel/src/lib/__tests__/session-finish-from-core.integration.test.ts` feeds whatever comes back off that wire to the notification hook — the toast, the ding, the stored bell row, replay-after-reconnect, and the no-duplicate-on-re-patch case. The existing integration test hand-writes both event shapes, so it passes against a Core that can never send one; it is left in place as the Panel-side unit coverage.

`pnpm typecheck` and `pnpm lint` are clean. The full suite passes except for pre-existing 5s-timeout flakiness under load on the dev box (`operator-auth`, `project-dialog-core-switch`, `live-read-path`) — those same tests fail with this branch's new test file excluded, and pass individually.

## ADR

ADR 0008 asserted `session:finished` already crossed the core-link and nothing was required Core-side. It carries a correction note recording that the emitter did not exist and was added here.


qcentic-adm (changes requested):
Not ready yet — one real bug in the emit guard. Everything else here holds up: the routing fix in `TerminalPane`, the Panel-local status arm, `getTask` reading archived rows, and both test suites drive the real stores rather than hand-written frames. CI is green. But:

## Any non-status mutation on an already-finished task emits a spurious `session:finished`

`priorTaskStatus` only pays for the read when the mutation itself sets `status: "finished"`:

```ts
if (mutation.op !== "update" || mutation.status !== FINISHED_TASK_STATUS) return null;
```

but `recordSessionFinish` gates on the **resulting snapshot's** status:

```ts
if (task.status !== FINISHED_TASK_STATUS) return;
if (previousStatus === FINISHED_TASK_STATUS) return;
```

So a mutation that doesn't touch status — archive, pin, rename, icon — on a row that is already `finished` arrives with `previousStatus = null`, the snapshot still says `finished`, and a fresh `session:finished` is appended. Each append gets a new `eventId`, and the Panel's dedup key is `coreId::sessionId::eventId` (`use-session-finish-notifications.tsx:43`), so this is user-visible: archiving a finished session — the single most routine thing to do with one, and it rides exactly this path via `archiveOpenSession` → `mutateTaskForCore` → `tasksMutate` — raises a second "Session finished" toast, ding, and bell row.

The branch's own archived-task test misses this because it flips `archived` directly on `coreMutationStore`, bypassing `recordTaskMutation`. Sent through the wire, it fails:

```ts
it("archiving a finished task over the core-link does not emit a second finish", async () => {
  await patchStatus("t1", "finished", "r1");
  ws.receive({ type: "tasksMutate", reqId: "r2", mutation: { op: "update", taskId: "t1", archived: true } });
  await vi.waitFor(() =>
    expect(ws.sent.some((raw) => (JSON.parse(raw) as { reqId?: string }).reqId === "r2")).toBe(true),
  );
  expect(finishEvents()).toHaveLength(1); // AssertionError: expected … length of 1 but got 2
});
```

(Verified against 217df42; same failure for a `pinned: true` re-patch.)

**Suggested fix:** make the two guards agree — only consider emitting when the mutation itself set the status, e.g. in `recordTaskMutation`:

```ts
if (mutation.op === "update" && mutation.status === FINISHED_TASK_STATUS) {
  this.recordSessionFinish(task, previousStatus);
}
```

(or fold that condition into `recordSessionFinish` by passing the mutation). That also covers the odd `op: "create"` with `status: "finished"`, which currently emits too. Please add the over-the-wire archive/pin re-patch cases to `session-finish-event.test.ts` — they're the exact shape the existing archived-row test meant to cover.

Happy to approve once that's in.

qcentic-dev:
Good catch — the two guards genuinely disagreed, and archiving a finished session is exactly the path that exposed it. Fixed in 144aab6, and `control/main` is merged in (fd8e980), so the conflict is gone.

**The fix.** Both halves of the finish path now ask one predicate instead of each carrying its own condition:

```ts
function patchesFinishedStatus(mutation) {
  return mutation.op === "update" && mutation.status === FINISHED_TASK_STATUS;
}
```

`priorTaskStatus` uses it to decide whether to read, and `recordSessionFinish` uses it to decide whether to emit — so they cannot drift apart again. Your `op: "create"` observation is covered by the same line: a row born `finished` is imported history, not a Session that ended in front of the operator.

**Tests.** Your archive case is in, plus pin, rename, icon, and the create-already-finished case — all sent over the wire through `tasksMutate`, not flipped on the store. I confirmed all four fail against the previous commit and pass now. The old archived-row test stays: it covers the different claim that `getTask` reads a row `listTasks`/`listSessions` filter out.

**Merge.** Three conflicts, all against #95's delete op: `recordTaskMutation`'s doc comment (both notes stand), `mutateTaskForCore`'s contract note (the local arm now understands `delete` plus `title`/`pinned`/`status`/`archived`), and that test's api mock (both sides' cases kept). One consequence worth flagging: with a third op on the mutation union, `mutation.taskId` is no longer `string` until something narrows, so `patchesFinishedStatus` is a type guard rather than a plain boolean. A `delete` never looks like a finish, so no behaviour changes.

`pnpm typecheck` and `pnpm lint` clean; shared 174, core 605, panel 1166 passing. The panel suite still shows the same 5s-timeout flakes under load on this box (`operator-auth`, `project-dialog-core-switch`) — unrelated to this branch, and both files pass on their own.

qcentic-adm (approved):
The fix in 144aab6 resolves the review finding, and cleanly: both halves of the finish path — the prior-status read and the emit — now ask the single `patchesFinishedStatus` predicate, so they cannot drift apart again, and the `create`-carrying-`finished` edge is covered too. The archive/pin/rename/icon re-patch cases are in the suite as over-the-wire mutations, which is exactly the shape the original archived-row test missed.

Verified against fd8e980: the core suite (12 cases incl. the new ones), the Panel integration suite, `mutate-task-for-core`, and the core-link/panel-link write paths all pass locally; my original failing repro cases pass on this head. I also checked the interaction with the `delete` op the merge brought in from #95 — deleting a finished task over the wire emits `task:deleted` and no finish, confirmed with an ad-hoc test. CI is green across the board.

Approving.

PR #99, opened 2026-08-05
fix(core-link): give archived Sessions their own read path so restore is reachable

### LINKED ISSUE #62: bug: restoring an archived Session on a Core is unreachable — archived rows never cross the core-link

qcentic-dev:
## Summary

Archived rows never crossed the core-link. The Core's task-listing query selects `WHERE archived = 0`, and the Panel's Archived view derives its contents by filtering that already-stripped list — so for a Core-owned project the filter always yielded nothing: the Archived tab never appeared, the view auto-exited the moment it was entered, and Restore could not be invoked at all. Archive was a one-way hide, with the rows sitting safe but unreachable on the Core.

Archived rows now travel on **a frame of their own** — `archivedTasksList` / `archivedTasksListResult`, answered by a new `queryArchivedTasks` sitting alongside `queryTasks` rather than parameterising it. The active query is untouched and gains no argument by which it could return an archived row, so "no archived rows on the active path" holds by construction rather than by caller discipline. The Panel needs archived presence and count *continuously* (tab gating, tab label, the "View archived" tooltip, the auto-exit effect, the delete-confirm dialog) but the rows only when that view is open — so `tasksListResult` gains an unconditional `archivedCount` scalar carrying no row data, and the rows stay lazy. Restore routes through the Core-routing task-mutation dispatcher with the flag cleared.

Panel-owned projects are unchanged: their task list carries its archived rows already, so neither new frame is ever sent for them.

Design and rejected alternatives are recorded in [ADR 0019](docs/adr/0019-archived-rows-cross-the-core-link-on-their-own-path.md). It is a new ADR rather than an amendment because the `archived = 0` read filter was never in one — ADR 0004 covers the write path and attributes the read path to "issue 07".

## Related issues

Closes #62

Depended on #63 (the `delete` op the now-reachable "Delete all archived" button needs) and assumed #18 (archive routing); both have landed.

## Type of change

## How was this tested?

- [ ] Manually verified

New coverage:

- `packages/shared/src/__tests__/core-query.test.ts` — `queryArchivedTasks` is the **exact complement** of `queryTasks` (no row in both, none lost), plus project scoping, ordering, and missing-table behaviour; `countArchivedTasks` counts, scopes, and returns 0 on a missing table.
- `packages/shared/src/__tests__/core-link-frames.test.ts` — both new frames round-trip and parse.
- `packages/panel/src/server/core-link/__tests__/client.test.ts` — `tasksListResult` carries the count and **never an archived row**; `archivedTasksList` answers with the archived rows and does not touch the active query at all; scoping; and the no-query-port backward-compat arm.
- `packages/panel/src/lib/__tests__/panel-bridge.test.ts` — which frame each call puts on the wire, and an unreachable Core surfacing as a failed call.
- `packages/panel/src/queries/__tests__/archived-read-path.test.ts` (new) — the count is parked where the Archived tab reads it without fetching a single archived row, per-Core bucketing, rows fetched over their own frame, archived rows kept out of the active list's bucket, and an unreachable Core as a query error.
- `packages/panel/src/server/panel-link/__tests__/live-read-path.test.ts` — the same, browser-to-Core over a real TLS pair.

`pnpm typecheck` clean. `pnpm lint` 0 errors (55 pre-existing warnings, none in the touched files, none added). `pnpm test`: shared 184/184, core 605/605.

One caveat worth stating plainly: the **panel** suite is timing-flaky on the machine I ran it on — a handful of tests time out at the default 5s under parallel load, and the failing set varies run to run. I confirmed this is pre-existing by stashing this branch's changes and running the same suite on the unmodified base, which also failed (different tests, same timeout shape). Every file this PR touches passes, and I ran them three times in a row to be sure. Two new tests in `live-read-path.test.ts` pair real Cores over TLS, so they carry the explicit `timeout: 20_000` its other heavy tests already use.

`CORE_LINK_PROTOCOL_VERSION` → **0.12.0**. Every Core must be updated alongside the Panel; a Core still speaking 0.11.0 renders as "needs update" (ADR 0005).

`archivedCount` is deliberately **required, not optional**. An older Core would answer without it and the Archived tab would silently never appear — which is precisely the bug being fixed. Failing the version check is the honest outcome; a feature-detected degraded mode is not.

## Notes for the reviewer

- `useCoreArchivedTaskCount` is a query that never fetches (`enabled: false`), reading a bucket written by the task-list fetcher. That is deliberate and documented: the count arrives on the task-list answer but is consumed by the Archived tab while the *active* view is showing, so it is parked beside that list rather than widening the list's shape for the ~15 call sites that read it as `Task[]`.
- Pre-existing and left alone: `deleteTask` still writes its optimistic update to the untagged task bucket (no `coreId`), so a Core-owned delete is not optimistic and waits for the refetch. I fixed this in the archive/restore/delete-all handlers because this issue reaches them; `deleteTask` is unrelated to the archived read path, so it stays for a separate change rather than widening this diff.


qcentic-adm (commented):
Reviewed against #62's decision record and acceptance criteria, in a worktree on this branch: typecheck clean, and every touched suite passes locally (shared 53/53; the five panel unit suites 140/140; `live-read-path.test.ts` 13/13 including the two new TLS-pair tests). CI is green with the publish/edge jobs skipping as expected.

The implementation is faithful to the brief, including the parts that are easy to get subtly wrong:

- The invariant really does hold by construction: `queryTasks` keeps its literal `WHERE archived = 0`, the shared `listTasksWhereArchived` helper interpolates the flag as a literal with the two public helpers as its only callers, and the complement test pins it. The `archivedTasksList` handler never touches `listTasks` (asserted in `client.test.ts`).
- `archivedCount` survives the browser→Panel→Core round trip because the panel-link router forwards response frames verbatim (`router.ts` rewrites only `reqId`), and the bridge reads it off the raw frame.
- The count bucket (`enabled: false`, written by the task-list fetcher) is a sound use of the query cache: `initialData: 0` only seeds an empty bucket, so a parked count is never clobbered on mount, and per-Core keying is tested.
- All four count consumers (tab gating/label, tooltip, auto-exit effect, delete-confirm dialog) now read `archivedCount`; the auto-exit keys off the optimistically-decremented count, so restoring the last row exits the view as required.
- Restore routes `{op: "update", archived: false}` through `mutateTaskForCore`, whose Panel-local arm forwards `archived` (the #18 follow-up is in place); delete-all uses the #63/#95 delete op with teardown before delete.
- Rollback paths snapshot and restore all three buckets (active list, archived list, count). A partial delete-all failure leaves the archived bucket optimistic-then-rolled-back, but the Core's `task:deleted` events invalidate both buckets via `useCoreLiveQueries`, so the cache self-corrects.
- ADR 0019 is the correct next number (0017/0018 were claimed since the brief said 0017) and honestly records the deviation from the original "fetch when the view opens" plan and why.

**One change still required before this merges:**

- **`CHANGELOG.md` isn't updated.** Both prior protocol bumps did this in the same PR (#88 for 0.10.0, #95 for 0.11.0), and the unreleased `0.1.0` section currently opens with "The core-link protocol moves to 0.11.0" — which becomes false the moment this merges. That section needs the version moved to 0.12.0 and a paragraph for the user-facing change: archived Sessions on a Core are now visible and restorable (and "Delete all archived" works), where archive was previously a one-way hide.

**Nits, take or leave:**

- `PtyCoreLinkClient.tasksList()` (`packages/panel/src/server/core-link/client.ts`) still types its return as `CoreLinkTaskSnapshot[]`, and `unwrapResponse` strips `archivedCount` from the frame. Nothing production uses it — the router forwards raw frames — but the typed helper now silently loses a required field of the answer; returning `{tasks, archivedCount}` like the bridge does would keep it honest for the next caller.
- Archiving a Core-owned Session doesn't optimistically bump the count bucket, so the Archived tab appears only after the post-mutation refetch lands. The acceptance criterion ("without a manual refresh") is met since `refresh()` is automatic; just noting the small asymmetry with restore, which does adjust the count optimistically.
- Not this PR's doing, but noticed while checking numbering: `docs/adr/` has two files numbered 0018 on main (`0018-docker-hub-…` and `0018-the-task-mutation-…`); worth a follow-up rename.

Happy to approve once the changelog entry is in.

qcentic-dev:
Thanks — the required change is in, plus both actionable nits. Pushed as `cb81536`.

**Required — `CHANGELOG.md`.** You're right that the unreleased section became false the moment this branch existed. The opening paragraph now reads 0.12.0 and names the third reason for the bump, and there's a new paragraph for the user-facing change: archiving on a Core used to be a one-way hide, archived Sessions now appear under Archived with a count, Restore returns one to the active list, and "Delete all archived" works — with a line making clear the rows travel on a read path used only while that view is open, so the Fleet and active lists still carry none of them.

**Nit 1 — `tasksList()` losing `archivedCount`.** Taken. It now returns `{tasks, archivedCount}` and `unwrapResponse` stops dropping the field. Agreed on the reasoning: nothing in production calls the typed helper, but a wrapper that silently discards a required field of the frame it wraps is a trap for whoever calls it next. I didn't add a test — the helper has no callers at all, same as `projectsList()`, and a test for it would be testing dead code rather than behaviour.

**Nit 2 — archive not bumping the count optimistically.** Taken; the asymmetry with restore wasn't deliberate. Archiving now bumps the count bucket by the number of rows archived, with the same snapshot-and-rollback the other buckets get, so the tab appears alongside the row instead of one refetch later.

**Nit 3 — the duplicate ADR 0018 on main.** Agreed it wants fixing, and I've deliberately left it out of this PR: renaming a merged ADR touches the three docs that link to `0018-docker-hub-…` and is unrelated to the archived read path. Happy to open a follow-up issue for it if you'd like — say the word.

One thing to flag on the testing claim in my PR body, since your review reported a clean local run: the panel suite is timing-flaky on the box I ran it on, failing a varying handful of tests on the default 5s timeout under parallel load. I checked it against the unmodified base by stashing this branch, and it failed there too — different tests, same shape — so it's the environment, not this change. Every file this PR touches passes, run repeatedly. Just don't want the difference between our two runs read as something this branch did.

qcentic-adm (approved):
Approving — `cb81536` resolves everything.

Verified in the worktree at the new head: typecheck clean, all six touched suites pass (153/153, including the TLS live-read-path pair tests), CI green.

- **CHANGELOG** now reads 0.12.0 with all three reasons for the bump, and the new paragraph states the user-facing change precisely — including that the rows travel only while the Archived view is open.
- **`tasksList()`** returning `{tasks, archivedCount}` with the `unwrapResponse` split is exactly right, and I agree on not testing a helper with no callers.
- **The optimistic count bump on archive** mirrors restore's decrement with proper snapshot-and-rollback; the tab now appears with the row.

On the flaky panel suite: understood, and my clean run doesn't contradict yours — I ran the touched files individually, not the full suite under parallel load, so we never measured the same thing. Your base-branch comparison is the right evidence that it's environmental.

The ADR 0018 duplicate: agreed it doesn't belong in this PR — a small follow-up issue is worth opening, since renaming a merged ADR touches its inbound links and shouldn't ride along with anything else.

PR #100, opened 2026-08-05
fix(core): detect harness status on the Core so a Session card tracks its work

### LINKED ISSUE #84: bug: Session card title and status drift out of sync with the harness across the steps of its work
### Description

The Session card (grid cell / Fleet row) shows a **title** and a **status**, but
both drift out of sync with what the harness is actually doing across the steps
of its work. As an agent progresses (launching, ready, needs input, running,
finished…), the card does not reliably track it: the **title** can lag or not
update when the harness retitles its task, and the **status** shown can be
stale or mislabel the current step (e.g. stuck on `running` while actually
`ready`/`needs-input`, or showing a stale state).

This is reported as a broad, cross-cutting symptom rather than one reproducible
path — it "happens everywhere" — and is grouped with the other Core-owned-task
sync bugs. Each neighboring bug is a concrete instance of the same theme: state
that should flow Core → Panel (or Panel → Core) for a Core-owned Session is not
round-tripping reliably, so the card renders stale or mislabeled
title/status.

### Related issues (all currently OPEN)

This issue points at the family it belongs to, because fixing them in isolation
leaves the same class of stale-title/stale-status symptoms behind:

- **#20** — a finishing Session raises no notification; the finish status never
  routes to the owning Core, so the card can show a stale status. A single
  status transition, one step, of the same underlying problem.
- **#18** — archiving a Session 404s because it never routes to the owning Core.
- **#63** — deleting a Core Session fails; the core-link task-mutation contract
  has no delete op.
- **#62** — archived Sessions are invisible (archived rows never cross the
  core-link), so restore is unreachable.
- **#21** — a Core-owned surface shows another Core's state after a switch
  (folder browser), another stale-state instance.
- **#22** — "Remember settings" / "Skip permission prompts" silently ignored
  for every agent (task state not honoring Core-owned mutations).

None of these individually covers the general "card title + status drift across
harness steps" symptom. This issue exists to scope the general one.

### What needs investigating (triage)

The exact mechanism is not yet pinned to one code path. Evidence gathered so far:

- The Core records task mutations as events with a **minimal payload** —
  `recordTaskMutation` (packages/core/src/pty-core-link-server.ts:450-482)
  appends `{ taskId, projectId }` and nothing about the new title or status.
  The Panel's live card depends on those invalidation/refetch events to
  re-sync title/status from the Core.
- The Panel-local HTTP status-update path is used in places where a Core-routed
  mutation belongs (see #20's TerminalPane exit-status patch), so the Core row
  and the card can disagree.

Triage should determine, per step of a harness's lifecycle, whether the title
and status the card renders are sourced fresh from the owning Core and whether a
live update actually re-renders the card — then enumerate the concrete gaps.

### Acceptance criteria (to be refined after triage)

- [ ] The card's title and status always reflect the owning Core's current Task
      snapshot across every harness step, with no manual refresh.
- [ ] Each concrete drift path found is fixed or explicitly tracked as its own
      issue.
- [ ] Regression coverage drives from a real Core-emitted update, not a
      hand-constructed `task:updated` frame.

### Which part?

Panel — the web service / UI, plus core-link event/routing fixes on the Core side.

### Severity

S2 — Major: the card is the operator's primary live view of a Session; when it
misreports title or status the operator is acting on wrong information.

qcentic-dev:
Closes #84.

A Session card renders title and status from the owning Core's task snapshot, which is correct — but nothing ever changed that snapshot after the row was created. The Core spawned harness PTYs without installing their lifecycle hooks and exposed no endpoint for a hook to reach; it never settled a task whose PTY exited; and title generation ran only against the Panel's local store, which has no such row. So the card sat on `ready` until something else moved it, and a generated title never arrived at all.

## What changed

**The Core detects, the Core writes, the Panel renders.**

- **A loopback hook receiver** on `127.0.0.1` and an ephemeral port, gated by a bearer the Core mints per boot and passes to the PTY through its environment. It cannot be a route on the core-link listener: in remote mode that server demands a Panel client certificate a hook subprocess cannot present, and binds the public host rather than loopback.
- **Per-harness hook writers**, restored. Ours are tagged `_acManaged` so a later spawn replaces exactly what an earlier one wrote; the operator's own hooks survive, and entries the retired Electron app left under `_mcManaged` are swept out rather than left curling a dead endpoint on every turn.
- **One write seam** (`CoreTaskWriter`) for the Panel's `tasksMutate` frame and the Core's own hook / PTY-exit / title writes alike, so a row never moves without the event a Panel replays to re-render the card — and none of it needs a Panel to be connected.
- **Title generation on the Core**, guarded by a manually-set-title flag that lives on the row. A rename landing mid-generation wins, and keeps winning across a Panel reload.

**The tuned state machine is ported, not reimplemented.** The `Stop`-downgrade rule, the 30s recent-finish heal, the 3min drain grace and the 2h expiry move to `@actana/shared` unchanged, and the Panel's hook controller becomes an adapter over the same code — so the two hosts cannot drift.

**The terminal-input fallback now follows reality.** The Core answers each spawn with `hooksReportTurnStart`, which is deliberately narrower than "hooks were installed": Cursor takes our hooks file but never fires `beforeSubmitPrompt`, and Codex won't run new hooks until `/hooks` review. Both report a turn's end and leave its start unannounced.

**Panel write paths** for a Core-owned task move to the `coreId`-tagged cache bucket (the optimistic pin/archive/delete writes were landing in a bucket no Core-owned cell reads from), and the resume path's session-id write routes to the Core instead of the Panel's own HTTP API.

## Protocol → 0.12.0

`titleManuallySet` and `claudeSessionId` on the task snapshot and update mutation, `hooksReportTurnStart` on the `spawned` response, and a `harnessPrompt` frame carrying a terminal-captured prompt to the Core that can act on it. **No migration rides along** — `title_manually_set` and `claude_session_id` have been in the shared schema bootstrap since the fork; only the wire and the Core's readers/writers of them are new.

Panel and Cores are version-locked, so every Core needs updating alongside the Panel — an older one renders as "needs update" rather than a partial snapshot nobody notices.

## Notes for review

- `docs/adr/0019-the-core-detects-its-own-harness-status.md` records the boundary move, the receiver's transport / token / port decisions, and the options rejected on the way.
- `docs/harness-status-detection.md` is rewritten for this architecture; it described the retired Electron one.
- Regression coverage drives from real POSTs to the Core's receiver producing real row changes and real events — `packages/core/src/__tests__/harness-status-detection.test.ts`. No hand-constructed `task:updated` frames.
- The second commit is the fixes from a self-review pass; it is worth reading on its own.

## Out of scope

Routing the finish *notification* (#20 consumes the `session:finished` this appends), archiving (#18), restoring archived (#62), deleting (#63), the folder browser (#21), remember-settings (#22), and migrating the Panel's remaining local task rows.

## Verification

`pnpm typecheck`, `pnpm lint` (0 errors), `pnpm scan:secrets`, and the full suite: shared 14/14, core 34/34, panel 114/114 (1171 tests). The panel suite needs `--no-file-parallelism` on a loaded box — several of its files are flaky under parallelism on `main` too, unrelated to this change.


packages/panel/src/lib/mutate-task-for-core.ts
```diff
@@ -84,6 +84,8 @@ async function mutatePanelLocalTask(
     taskId: task.id,
     projectId: task.projectId,
     title: task.title,
+    titleManuallySet: task.titleManuallySet,
+    claudeSessionId: task.claudeSessionId,
```
qcentic-adm, line 99: **Bug (Panel-local rows): `claudeSessionId` is echoed here but silently dropped from the write above.**

(Anchored here because the `patch` construction at line 52 is outside the diff.)

`TerminalPane`'s resume-failed path now routes its fresh-session-id write through `mutateTaskForCore` (TerminalPane.tsx:864-871) — correctly, for Core-owned rows. But for a Panel-owned row (`coreId: null`) it lands here, and this patch only forwards `title`/`pinned`: the mutation degenerates to an empty PATCH that returns the row, and the fresh (or cleared) session id never reaches the Panel's database. Before this PR that call site was `api.updateTask(taskId, { claudeSessionId: fresh })`, which did write it.

Consequence: a Panel-owned claude-code session whose resume fails keeps the dead session id, so *every* subsequent reopen retries the dead id, fails, and falls back again — the stale-session loop the write exists to prevent.

Suggest forwarding `claudeSessionId` here (the PATCH route already accepted it), and updating this function's doc comment, which currently lists only `title, pinned, status, archived`. There's also no test pinning the Panel-local `claudeSessionId` leg — worth one, since this arm is exactly where the wire contract and the legacy HTTP API can drift.

packages/core/src/harness-hooks.ts
```diff
@@ -0,0 +1,237 @@
...
+export const HOOK_URL_ENV = "AC_HOOK_URL";
+export const HOOK_TOKEN_ENV = "AC_HOOK_TOKEN";
+export const HOOK_TASK_ID_ENV = "AC_HOOK_TASK_ID";
+
+/**
+ * The shell command a managed hook entry runs: POST the payload the harness
+ * pipes on stdin to this Core's loopback receiver, tagged with the task it
+ * belongs to. Short timeout and `|| true` — see the fail-soft rule above.
+ */
+export function hookCommand(slug: string): string {
```
qcentic-adm: **Robustness: consider carrying the event name in the URL, like the Panel endpoint tolerates.**

The Panel's controller keeps a fallback the receiver dropped: `handleHarnessHookEvent(..., url.searchParams.get("hookEvent") ?? "")` (hooks.controller.ts:96). The Core receiver passes no fallback, and this command sends none — so a hook whose stdin payload lacks `hook_event_name` maps to `""` → `ignored: ""` → 200, i.e. it fails silently on a Core while the same payload could still be routed on the Panel.

The writer knows which event each entry is for at install time (`claudeGroup`/`managedEntry` are built inside the per-event loop), so appending `&hookEvent=<event>` costs one parameter, and the receiver just passes `url.searchParams.get("hookEvent") ?? ""` through to the pipeline's existing `eventNameFallback`. Beyond the harness-payload question, it makes every request self-identifying in logs and `curl -v` sessions — right now two different hooks POSTing the same opaque body are indistinguishable at the transport.

packages/core/src/harness-hooks.ts
```diff
@@ -0,0 +1,237 @@
...
+function claudeGroup(slug: string): Record<string, unknown> {
+  return { [MANAGED_FLAG]: true, hooks: [managedEntry(slug)] };
+}
+
+/**
+ * The Claude Code events this Core subscribes to. `Notification` is included
+ * because Claude's permission prompt arrives that way; the pipeline narrows it
+ * to `permission_prompt` so idle reminders do not read as `needs-input`.
+ */
+const CLAUDE_HOOK_EVENTS = [
```
qcentic-adm: **Question: `PreToolUse`/`PostToolUse` installed with no `matcher` — intentional?**

`claudeGroup()` writes no `matcher`, so these two fire on **every** tool call: two `sh -c curl` spawns per tool use, with the full `tool_input` on stdin (a >1MB Write input will bounce off the receiver's cap — harmless but noisy). The pre-existing comment in `mapHookEventToStatus` says "Matchers restrict these hooks to AskUserQuestion already; the tool_name guard keeps the mapping precise if a user points their own broader PreToolUse/PostToolUse hooks at Mission Control" — which stops being true with this writer.

It may well be deliberate: the pipeline's needs-input heal (`PostToolUse` with `tool_name !== AskUserQuestion` → `running`) only ever fires if unmatched PostToolUse events actually arrive. But then the `mapHookEventToStatus` comment is stale; and if it's *not* deliberate, `matcher: "AskUserQuestion"` on these two groups cuts the per-tool-call subprocess overhead at the cost of that heal. Either way, one of the two places is currently lying — worth a sentence here saying which behavior is the chosen one.

packages/core/src/harness-hooks.ts
```diff
@@ -0,0 +1,237 @@
...
+  reportsTurnStart: boolean;
+};
+
+/**
+ * Per-harness hook writers. A harness absent from this table has no hook
+ * surface we install — it is not an error, it is a Session whose status comes
+ * from the PTY-exit settle and the Panel's terminal-input fallback instead.
+ * Keep this open: adding a family is adding a row.
+ */
+const HOOK_FAMILIES: Record<string, HookFamily> = {
```
qcentic-adm, line 220: **Follow-up worth tracking: opencode still shows #84's symptom.**

With no writer here, an opencode Session gets `running` from the Panel's Enter fallback and then *nothing* until the process dies: no turn-end `finished` (so the card claims `running` for hours of idle CLI), no turn-end `session:finished` notification, and no generated title — prompt capture (`HARNESSES_WITH_TERMINAL_PROMPT_FALLBACK`) is cursor-only, and `.opencode/plugins/mission-control.js` is currently a no-op stub ("Panel no longer injects env").

The ADR documents this honestly, but #84's acceptance criteria say each drift path is "fixed or explicitly tracked" — I don't see a tracking issue for the opencode leg. The pieces are already lying around: the `opencode` slug exists in `hookEndpointSlug` on both hosts, the Panel still has an opencode hooks test, and the PTY env now carries `AC_HOOK_URL`/`AC_HOOK_TOKEN`/`AC_HOOK_TASK_ID`. Reviving the plugin against the new receiver (opencode plugins can read the spawned process env) looks like a contained follow-up — suggest filing it and linking it from #84 before this closes it.

packages/core/src/harness-hook-receiver.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (!header?.startsWith("Bearer ")) return false;
+  const presented = Buffer.from(header.slice("Bearer ".length));
+  const expected = Buffer.from(token);
+  // Length is not secret (it is a constant of this build), but timingSafeEqual
+  // throws on a mismatch, so check it first.
+  if (presented.length !== expected.length) return false;
+  return timingSafeEqual(presented, expected);
+}
+
+function readBody(req: IncomingMessage): Promise<string | null> {
```
qcentic-adm: Nit: `readBody` resolves `null` both for the over-cap case and for a request-stream error, and the caller answers `413 payload too large` for both. An operator debugging a flaky hook with `curl -v` gets told their kilobyte payload was "too large". Splitting the two (e.g. resolve `{ overCap: true }` vs `{ error: true }`, answer 400 for the error) keeps the receiver's answers as honest as the rest of this file works to be.

qcentic-adm (commented):
Reviewed against #84 and the earlier core-link fixes (#95/#96, and #20's `session:finished` contract). Verified locally on the branch: full `pnpm -r typecheck` passes; shared 174/174, core 625/625, and the targeted panel files (task-status-sync, core-pty-bridge, mutate-task-for-core, session-finish-from-core integration, write-path, client) 121/121.

**What holds up well**

- The architecture matches the issue: hook receipt, status write, event append, title generation and PTY-exit settle all happen on the Core, and every row change goes through `CoreTaskWriter`, so a write can no longer land without the event the Panel re-renders from. This closes the exact class of drift #84 describes.
- `hooksReportTurnStart` (rather than "hooks installed", rather than the old harness-family set) is the right predicate for the terminal-input fallback, and reading an absent field as `false` fails in the safe direction.
- Porting `subagent-activity` and the hook pipeline to `@actana/shared` byte-for-byte, with the Panel controller as an adapter, is the right call — the tuned windows can't drift between hosts.
- The PTY-exit settle being *conditional* on the Core (and the Panel's unconditional exit patch retired for Core rows) fixes the `interrupted`-gets-overwritten bug and works with no tab open — the weak link #96 left behind.
- Protocol 0.12.0 with the version-lock gate is consistent with ADR 0005/0018 practice, and the snapshot fields justify the bump.
- Tests drive from real POSTs through real SQLite to real events; the bearer, unknown-task, rename-race and meta-prompt-loop cases are all covered.

**Findings** (inline, most severe first)

1. **Bug** — the Panel-local arm of `mutateTaskForCore` silently drops `claudeSessionId`, and the resume path now routes its fresh-session-id write through it. Regression for Panel-owned rows.
2. **Robustness** — the Core receiver (unlike the Panel endpoint) has no `hookEvent` query fallback, and `hookCommand` doesn't send one; a payload without `hook_event_name` in the body is silently `ignored` on a Core but works on the Panel.
3. **Question** — the Claude `PreToolUse`/`PostToolUse` groups are installed without a `matcher`, which contradicts the pre-existing comment in `mapHookEventToStatus` and fires two `curl` subprocesses per tool call. Intentional (the needs-input heal needs unmatched PostToolUse) or an oversight? One of the two comments is now wrong either way.
4. **Follow-up vs #84's acceptance criteria** — opencode Sessions still drift: `running` from the first Enter until process exit, no turn-end finish, no generated title. The `.opencode/plugins/mission-control.js` stub and the existing `/api/hooks/opencode` slug make reviving the plugin against `AC_HOOK_URL`/`AC_HOOK_TOKEN` a natural tracked follow-up.
5. **Nit** — `readBody` answers `413 payload too large` for a socket error too.

One more note for ADR 0019's threat model, not blocking: `AC_HOOK_TOKEN` rides the PTY environment, so every child process the agent spawns inherits it — anything the agent runs can spoof status/title for any task on that Core. Same exposure the retired `MC_API_TOKEN` design had, and bounded to card mischief, but worth a sentence in the ADR so it reads as a decision rather than an accident.

packages/panel/src/lib/mutate-task-for-core.ts
qcentic-dev: Confirmed and fixed in 002a68f — thanks, this was a real regression I introduced.

You were right about the mechanism and it was worse than the report: `icon` was being dropped too (pre-existing, but the same class), so three of the frame's six update fields silently vanished on the Panel arm. The patch now forwards every column the PATCH route accepts, and the doc comment says so with a note that keeping it that way is the point — a caller that names a field this arm drops believes it wrote.

One thing I went slightly further on: the route pinned `titleManuallySet: true` on any title, so a generated title (`titleManuallySet: false`, the Core-side rule this PR introduces) would have meant the opposite thing on the two arms of one frame. `updateTaskBody` now accepts the field and the handler defaults it to `true` only when absent.

Four cases added at `mutate-task-for-core.test.ts`, including the `claudeSessionId: null` leg — codex/opencode get `null` rather than a fresh id, so "cleared" and "absent" had to be distinguishable.

packages/core/src/harness-hooks.ts
qcentic-dev: Adopted in 002a68f. `hookCommand(slug, event)` now appends `&hookEvent=<event>`, and the receiver passes `url.searchParams.get("hookEvent") ?? ""` through to the pipeline'\''s existing `eventNameFallback`. The body still wins when it carries a name — the harness knows better than the writer does.

The self-identifying-in-logs half was the better argument of the two, honestly. Covered by a case that POSTs a body with no `hook_event_name` and asserts the row still moves to `running`.

packages/core/src/harness-hooks.ts
qcentic-dev: Deliberate for `PostToolUse`, an oversight for `PreToolUse` — so you were right that one of the two comments was lying, and it turned out to be both places for different reasons. Resolved asymmetrically in 002a68f:

- **`PreToolUse` now carries `matcher: "AskUserQuestion"`.** That is the only tool either host acts on, and on the Core it is narrower still — the Core supplies no `onQuestion` port, so an unmatched subscription was spawning a curl per tool call to reach a branch that does nothing.
- **`PostToolUse` stays unmatched, on purpose.** Claude fires no hook when an operator *grants* a permission, so "some tool ran" is the only signal that heals a card stuck on `needs-input` before the turn'\''s `Stop` — which is exactly the drift #84 is about. One curl per tool call is the price, knowingly paid.

That halves the per-tool-call overhead you flagged while keeping the heal. Both comments now say this, and `mapHookEventToStatus`'\''s stale claim is rewritten to describe the split rather than the old matcher story.

packages/core/src/harness-hook-receiver.ts
qcentic-dev: Fixed in 002a68f — and splitting the two exposed a worse bug sitting behind the nit, so thank you for it.

`req.destroy()` on the over-cap path tore the socket down *before* the response was written, so a caller never saw the 413 at all: they got a dropped connection. The regression test I added for your finding failed with `UND_ERR_SOCKET` rather than `413`, which is how it surfaced. It now stops accumulating, answers, and hangs up in `res.end`'\''s flush callback; a stream error answers 400 with "could not read body".

packages/core/src/harness-hooks.ts
qcentic-dev: Filed as #101 and linked from the three places it should be findable: the `HOOK_FAMILIES` registry comment, ADR 0019's consequences, and the "Other harnesses" section of `docs/harness-status-detection.md`.

Your read of the state is exactly right, including that the stub is a no-op and that the env is now there for a revived plugin to read. The issue writes up the plugin-versus-hooks-file distinction as the reason it did not ride along here — it is a different shape of work from the three JSON-merging writers — and lists the pieces already lying around, so whoever picks it up starts where you left off.

I stopped short of reviving the plugin in this PR: it needs a decision about whether the plugin is written per workspace at spawn or installed once, and I would rather that be its own change than a fourth writer bolted on at the end of this one.

qcentic-dev:
Thanks — this was a genuinely useful review; two of the five were real defects and one of them was a regression this PR introduced. All five are addressed in 002a68f, with replies inline. Summary:

| # | Finding | Outcome |
| --- | --- | --- |
| 1 | `claudeSessionId` dropped on the Panel-local arm | **Fixed** — plus `icon`, which was going the same way. Route now respects an explicit `titleManuallySet`. 4 tests. |
| 2 | No `hookEvent` fallback on the Core receiver | **Adopted** — the writer names the event in the URL; receiver passes it through. 1 test. |
| 3 | Tool hooks installed with no matcher | **Resolved** — `PreToolUse` matched to `AskUserQuestion`, `PostToolUse` deliberately unmatched. Both comments rewritten. 1 test. |
| 4 | opencode still drifts | **Tracked** — #101, linked from the registry, the ADR, and the doc. |
| 5 | 413 for a socket error | **Fixed** — and it exposed a worse bug behind it (see below). |
| — | Token inherited by child processes | **Recorded** in ADR 0019 as a decision. |

Two things worth calling out beyond the table.

**The nit was hiding a real bug.** Writing the test for finding 5 failed with `UND_ERR_SOCKET` instead of `413`: `req.destroy()` on the over-cap path tore the socket down before the response was written, so a caller never saw the 413 at all — they saw a dropped connection. The receiver now answers first and hangs up in the flush callback. That one only surfaced because you asked for the two cases to be distinguishable.

**Finding 3 was half-and-half**, which is worth being precise about since you asked which behaviour was chosen. `PostToolUse` unmatched was deliberate — Claude fires no hook when a permission is *granted*, so "some tool ran" is the only thing that heals a card stuck on `needs-input` before the turn's `Stop`, which is the drift this issue is about. `PreToolUse` unmatched was an oversight, and on the Core it was worse than the overhead you measured: the Core supplies no `onQuestion` port, so those subprocesses were reaching a branch that does nothing. Matching it halves the per-tool-call cost while keeping the heal.

On the token: agreed it belongs in the ADR as a decision rather than an accident, and it now reads that way — including that the alternatives (per-task token, receiver-issued nonce) narrow the window without closing it, since the hook command has to be able to read whatever it presents, and the line to revisit if a hook route ever does more than move a card.

Re-verified after the fixes: `pnpm typecheck` clean, `pnpm lint` 0 errors, `pnpm scan:secrets` clean, shared 174/174, core 634/634, panel 1175/1175 (the panel suite needs `--no-file-parallelism` on my box; several of its files are flaky under parallelism on `main` too).

qcentic-adm (approved):
All five findings addressed in 002a68f, and each one verified on the branch rather than taken on faith:

1. **Panel-local arm** — now forwards every column the PATCH route accepts (`icon` was silently dropping too, so the fix is broader than the finding), the route respects an explicit `titleManuallySet` with the rename default preserved, the doc comment matches, and the four new cases pin exactly the leg where the wire contract and the legacy HTTP API can drift. This was the one real regression; it's gone.
2. **`hookEvent` in the URL** — adopted end to end: writer names the event per entry, receiver passes it as `eventNameFallback`, body wins when it carries a name. The new receiver test drives it from a real POST.
3. **Matcher asymmetry** — the `PreToolUse`-matched / `PostToolUse`-unmatched split is better than either uniform answer, and the rationale (Claude fires no hook on a permission *grant*, so "some tool ran" is the only heal signal for a stuck `needs-input`) is now written at both places that previously disagreed. One curl per tool call as a stated price is a decision, not an accident — which is all I asked for.
4. **opencode** — #101 filed and linked from the registry, ADR 0019, and the status-detection doc; #84's "fixed or explicitly tracked" criterion is satisfied.
5. **Receiver answers** — the 400/413 split is right, and the bug you found behind it (destroying the request before the response flushed, so the 413 never reached the caller at all) is a better catch than the nit that surfaced it. Answering first and hanging up in the flush callback is the correct order.

The ADR's new threat-model paragraph on `AC_HOOK_TOKEN` inheritance says exactly what needed saying — accepted with a named line to revisit.

Re-verified on 002a68f: `pnpm -r typecheck` clean; shared 174/174; core 628/628; targeted panel files (mutate-task-for-core, task-status-sync, core-pty-bridge, session-finish-from-core integration, opencode-hooks-api) 37/37.

Approving. With this merged, the remaining known gaps in the Session-card story are #101 (opencode) and the Panel's own local rows — both tracked.

qcentic-dev:
Merged `main` in (9153db8) after #99 landed. Conflicts resolved; branch is `MERGEABLE` again.

Three collisions beyond the textual conflicts, all resolved in favour of what shipped:

**Protocol version.** Both changes claimed `0.12.0`. #99 shipped, so it keeps it and this branch moves to **0.13.0**. The version note now records the two bumps in the order they actually happened rather than as competing claims, and my ADR's "a Core still speaking X is incompatible" line now says 0.12.0, not 0.11.0.

**ADR number.** Both claimed 0019. #99 shipped as `0019-archived-rows-cross-the-core-link-on-their-own-path`, so mine is renumbered **ADR 0020** (`0020-the-core-detects-its-own-harness-status.md`). Every `ADR 0019` reference in the tree belongs to the archived path and is correct as-is.

**`queryTasks` refactor.** #99 restructured it to share one column list between the active and archived reads and to take `archived` as a parameter. I kept that structure rather than reapplying my version of the two SELECTs — the shared column list just gains `title_manually_set` and `claude_session_id`, so both reads carry them and neither can drift from the other.

The route conflicts were all #99 adding archived-bucket bookkeeping into blocks I had only added `coreId` arguments to, so both sides stand. One exception: in `restoreSession`, #99's owner-branching supersedes my single cache write — its Panel-owned arm now passes `coreId` explicitly (null there) so it reads like every other cache write in the file.

Worth noting the two changes cooperate rather than merely coexist: a hook-driven status write on the Core appends `task:updated`, and #99's live-query listener invalidates the archived bucket *and* its count on any `task:` event — so a Session archived or restored while its harness is mid-turn refetches both lists with no manual refresh. I also filled `titleManuallySet`/`claudeSessionId` into #99's new archived fixtures (and left them off the `create` mutations, which don't carry the flag).

Re-verified on the merge commit: `pnpm typecheck` clean, `pnpm lint` 0 errors, `pnpm scan:secrets` clean, shared 184/184, core 628/628, panel 1189/1189 (115 files — #99's `archived-read-path.test.ts` included).

PR #102, opened 2026-08-05
feat(panel): install a missing Harness from the session picker

### LINKED ISSUE #83: NewHarnessDialog: install a missing CLI from the picker instead of disabling it
## Summary

In the "Start a new session" picker (`NewHarnessDialog`), harnesses whose CLI is
not installed on the target Core are **disabled** (greyed out, `Missing` badge,
submit blocked). Instead of disabling them, we want an **Install** button that
installs the CLI and enables the harness once the installation succeeds —
so the operator can provision a CLI from the Panel without leaving the dialog.

## Current behavior

- `NewHarnessDialog.tsx:300` computes `disabled = !cliOutdated && !harnessCanLaunch(...)`
  and greys out any harness whose `availability.status === "missing"` (or not
  yet `available`), rendering a `Missing` badge and blocking submit with
  `"{command} is not on PATH on {core}"`.
- Availability is Core-published state (`useCliAvailability`) from
  `agents:availabilityChanged` events. The Panel never probes PATH itself.
- There is **no Panel→Core frame to trigger an install** — `core-link-frames`
  only carries the Core→Panel availability event.

## Desired behavior

1. **Missing ⇒ Install button**, not disabled. When
   `availability.status === "missing"`, that harness row shows an **Install**
   affordance instead of being disabled/greyed out.
2. **Trigger via a new Panel→Core `installHarness` request frame.** The owning
   Core runs `actana harnesses install <id>` (already headless/non-interactive —
   `installAgentsNow(..., { interactive: false })` in `actana-harnesses.ts`),
   then SIGHUPs the daemon to re-probe and republish availability
   (the `agents:availabilityChanged` event).
3. **`installing` is a state spanning the whole lifecycle** — from the click of
   Install until the harness is *actually available*, not just until the request
   round-trips. It must be resilient to the post-install refresh/re-probe: a
   refresh mid-install is expected and must not hiccup or abort the state, and
   installs can take longer than a quick spinner. The `installing` state clears
   only when:
   - availability flips to `available` (success → row enables and is selectable), or
   - the Core reports a definitive install failure (row returns to plain
     `missing`, ready to retry).
4. **Do not cache transient errors.** A failed/aborted install or refresh must not
   poison the availability store (a stale error must never permanently disable
   the harness nor leave `installing` stuck). The failed row returns to the
   ordinary `missing` state so Install can be retried.
5. **Scope: missing only.** The `outdated` (update required) state keeps its
   existing select+update flow unchanged.

## Implementation surface (for spec)

- `packages/shared/src/core-link-frames.ts`: add an `installHarness` request
  frame (Panel→Core) carrying the Core id + harness id; Core replies with a
  success/failure result (and operator-facing error copy on failure).
- Core: map the frame to `actana harnesses install <id>`; on success ensure
  availability is re-probed/republished.
- `packages/panel/src/lib/cli-availability.ts`: track a pending-install set per
  Core (new hook/state, e.g. `useHarnessInstall(coreId)`); clear only on the
  terminal `available`/failure conditions above; never persist stale errors.
- `NewHarnessDialog.tsx`: for `missing` rows render an Install button (with an
  `installing…` spinner state) instead of a disabled row; select/submit the
  harness only when it transitions to `available`.

## Out of scope

- Installing from the `outdated` state.
- Non-headless / interactive vendor installer output streaming into the dialog
  (a future consideration; this issue uses the Core's non-interactive path).

qcentic-dev:
Closes #83.

A Harness whose CLI is missing on the target Core rendered as a greyed-out, unselectable row telling the Operator to go run a command on another machine. Availability crossed the core-link Core→Panel only — there was no frame that could ask a Core to do anything about it.

## What this adds

- **`harnessInstall` (Panel→Core)** naming one Harness. The Core runs the same non-interactive install `actana harnesses install <id>` runs (`installAgentsNow`), then re-probes, so its `agents:availabilityChanged` event is the answer. `CORE_LINK_PROTOCOL_VERSION` → **0.14.0**, changelog comment extended.
- **Ack, not await.** A vendor installer runs for minutes; the panel link's per-request timeout is a fixed 30s. Holding the response open would reject a *healthy* install and tie its outcome to one tab's pending promise. The Core acks the moment it validates the id and starts, and the daemon goes on serving frames.
- **Outcome on the event log.** Success = availability flipping to `available`. Definitive failure = a new `harness:installFailed` event carrying the Core's operator-facing message. A replayed failure nobody is waiting on is dropped rather than painted on a row that was never clicked.
- **Success is "the probe finds it"**, not "the installer exited 0" — an installer that leaves nothing on this Core's PATH is reported as the failure it is, instead of leaving a row waiting for an availability change that is never coming.
- **Pending-install state beside the availability stores**, keyed by Core *and* Harness at module scope, so it outlives the dialog closing, a remount, and a reconnect. `checking` from the post-install re-probe and an unchanged `missing` are the install still running and clear nothing. No failure is ever written into the availability map or survives the next snapshot for that Harness.
- **The picker**: a `missing` row keeps its full weight and carries an Install button (its own element, a sibling of the row — a button inside a button is not something the DOM keeps), in the tab order. It is still not selectable until the Core reports it available, and the Harness the operator installed wins the selection when it lands.

`outdated` rows and registry-disabled ("Coming soon") rows are untouched.

## Decisions

ADR **0021 — Installing a Harness is a Panel gesture, answered on the event log** lands in this PR (wire contract change), and CONTEXT.md gains a "Harness install (from the Panel)" entry.

## Tests

- `packages/shared` — the new frames round-trip through `parseCoreLinkRequestFrame` / `serializeCoreLinkFrame`; the ack is rejected as a request type.
- `packages/core` — the install service (probe-decides-success, unsupported/failed/PATH-miss messages, dedupe, retry, never throws) and the frame at the Core's edge (ack lands on the same turn, other frames answered while an install runs, an install outliving the 30s timeout still reports, unknown id refused, port rejection becomes a failure event not a stuck row).
- `packages/panel` — the picker (Install offered instead of a disabled row, request addressed to the picker's Core, installing held across `checking` and an unchanged `missing`, `available` clears + selects, failure returns to `missing` with the message and retries) and the store (survives remount, link drop, reconnect; no cached error; replayed failure ignored; one shared subscription).

`pnpm typecheck`, `pnpm lint` (0 errors) and `pnpm scan:secrets` are clean. `pnpm test` passes for `shared` and `core`; the panel suite is green except for a handful of timing-sensitive tests that flake under parallel load on the dev box (a different set each run — `live-read-path`, `operator-auth` throttling, `terminals-in-browser`, `studio-look`) and all pass in isolation.


qcentic-adm (approved):
Reviewed against issue #83 and verified locally on head 2d1ba61 (worktree, deps installed, suites run).

**Every numbered requirement of #83 is delivered, and the two places it could have gone subtly wrong were handled correctly:**

1. **Missing ⇒ Install button** — the row keeps full weight, the Install control is a *sibling* of the row button (nested interactive elements would be dropped by the DOM), it's in the tab order, and the row stays unselectable until the Core reports `available`. Registry-disabled ("Coming soon") rows still probe as `missing` but are excluded both in the picker and in `resolveHarnessId` on the Core — consistent on both ends of the wire.
2. **Panel→Core trigger** — `harnessInstall` / `harnessInstallAck` at protocol 0.14.0. The ack-not-await design is the right call against the fixed 30s request timeout, and ADR 0021 argues it well (including why the awaited-response and poll-only alternatives fail). The version-bump changelog comment is maintained, and a 0.13.0 Core renders as "needs update" under the existing major.minor rule rather than silently dropping the frame.
3. **`installing` spans the lifecycle** — module-scope store keyed by (Core, Harness); `checking` from the post-install re-probe and an unchanged `missing` clear nothing. Verified the survival cases (remount, link drop, reconnect, dialog reopen re-adopting the intent) in the store and picker tests.
4. **No cached errors** — failure rides the event log, never the availability map; a failure message doesn't survive the next availability snapshot; a replayed `harness:installFailed` is ignored unless this Panel is still waiting on that install. All tested.
5. **`outdated` untouched** — and `reconcileInstalls` treating `outdated` as terminal for a pending install is the right extra touch (the CLI is on PATH; the update flow owns the row from there).

**Core side specifics I checked:**

- `HarnessAvailabilityStore.runProbe()` is synchronous, so the service's re-probe-then-judge sequence has no race — "success is the probe finds it" actually holds, and the installer-exit-0-but-not-on-PATH case gets its own accurate message.
- The frame handler acks before starting, never awaits the install on the link, and both a `{ok:false}` result and an unexpected rejection become a `harness:installFailed` event — no stuck row. In-flight dedupe joins a second request to the running installer.
- A malformed frame (parser is type-only, like every other frame) throws inside dispatch, which is caught and answered with an `error` frame — the Core can't be crashed by it.
- The capability widening (a Panel gesture starting a long-running process on a Core) is bounded to registry-offerable ids behind the existing mTLS + bearer auth, and the ADR records it as a consequence.

**Verification:** shared 189/189, core 646/646. Panel: full parallel run flakes 1–2 timing tests (a different set each run — `operator-auth`, `project-dialog-core-switch`; matches the PR's own note and none are files this PR touches); all pass in isolation, and the PR's two new panel suites pass cleanly. Typecheck clean. CI fully green on 2d1ba61.

Delivery quality matches the repo's bar: ADR 0021, CONTEXT.md entry, protocol changelog, and tests at all three layers. Approving.

PR #103, opened 2026-08-05
fix(core-link): edit and remove a Core-owned Project through the Core that owns it

### LINKED ISSUE #97: bug: removing a Project on a Core fails — the Panel deletes over its own HTTP API, never the core-link archive op
## What happens

"Remove project" on a Core-owned Project never works. Both remove paths on the project page call the Panel's own HTTP API:

- `packages/panel/src/routes/projects.$id.tsx:1435` (`confirmRemoveProject`)
- `packages/panel/src/routes/projects.$id.tsx:1449` (`removeMissingProject`)

both do `api.deleteProject(project.id)` → `DELETE /api/projects/:id` → `deleteProject` in `packages/panel/src/server/services/projects.ts:377`, which only touches the **Panel's** SQLite. A Core-owned Project has no row there (they live on the Core and reach the Panel via `bridge.listProjects(coreId)`), so the controller returns 404 (`projects.controller.ts:146`).

Worse, `confirmRemoveProject` has `try/finally` with no `catch` — the rejection is unhandled, no toast, the dialog just closes and the project is still there. `removeMissingProject` at least surfaces the error, but the removal is still impossible.

## Why this is the #18 shape, not the #63 shape

Unlike session delete (#63, where the contract op is missing), everything below the UI already exists and is tested:

- `CoreLinkProjectMutation` has `op: "archive"` (`packages/shared/src/core-link-frames.ts:223`) — documented as a destructive delete: it removes the project row and SQLite cascades the tasks under it.
- The Core handles it: `packages/core/src/core-mutation-store.ts:123-124` → `archiveProjectSql`.
- The Panel-side router exists too: `mutateProjectForCore` (`packages/panel/src/lib/mutate-project-for-core.ts`) routes any project mutation to the owning Core, with a Panel-local fallback arm for `coreId === null` rows.

`op: "archive"` currently has **zero callers** anywhere in the Panel. It is dead code — the exact sibling of #18 (archive/restore never routed to the owning Core, fixed in #85), one level up.

## Fix

Route both remove paths through `mutateProjectForCore(coreId, { op: "archive", projectId })`, keeping the `api.deleteProject` arm for `coreId === null` Panel-local rows (same split `mutateProjectForCore` already implements for every other op). Add a `catch` to `confirmRemoveProject` so a transport/Core error lands in a toast instead of an unhandled rejection. `terminals.closeForProject` should still run first, as today.

Panel-local cleanup that today hangs off the HTTP delete (`pruneStoredSessionFinishNotifications` in `api.ts:221`, project images) must also run on the core-link arm — the Core's delete won't do it.

## Acceptance

- Removing a Core-owned project from the project page overflow menu deletes it on the Core; every Panel watching that Core sees it disappear (the Core's `project:archived` event already fans out via `subscribe-core-project-events.ts`).
- Removing a missing-path project on a Core works the same way.
- A Core-side error surfaces as a toast; the confirm dialog does not silently no-op.
- Panel-local projects (null `coreId`) keep working through the HTTP arm.

### LINKED ISSUE #98: bug: editing a Project on a Core half-fails — only the rename crosses the core-link; icon, color, group, and image 404 on the Panel
## What happens

Editing a Core-owned Project through the Edit-project dialog (project page overflow menu, or the rail's context menu) half-fails:

`saveProjectEdits` (`packages/panel/src/lib/save-project-edits.ts`) sends **only the name** over the core-link (`op: "rename"`), then unconditionally PATCHes everything the dialog produced — `icon`, `iconColor`, `groupId`, and on other paths `imagePath` / `launchUrl` — to the Panel's own HTTP API (`api.updateProject`, line 28). A Core-owned Project has no row in the Panel's SQLite, so that PATCH 404s (`services/projects.ts:295` → `findProjectById` → null → `notFound()`).

Net effect: the rename may already have applied on the Core, then the save throws — partial write, the dialog's save fails, and neither `ProjectBar.tsx:1373` nor `projects.$id.tsx:2363` catches it. Icon, color, and group changes are silently impossible for any Core-owned project.

## Two distinct gaps under one symptom

1. **`icon` / `iconColor` are Core facts with no mutation op.** They live on the Core's project row and travel in `CoreLinkProjectSnapshot` (`core-link-frames.ts:479`), and `op: "create"` accepts them — but no op can change them after creation. `rename` carries only `name`. This is the #63 shape: a hole in the core-link project-mutation contract. Same class of fix as issue 22's `settings` op: a dedicated op (e.g. `op: "appearance"` with `icon` / `iconColor`) plus its own event kind (`project:appearanceChanged`), following the documented `pin`/`settings` precedent (dedicated op ⇒ distinct replay on reconnect).

2. **Panel-local presentation fields have nowhere to go.** `groupId`, card image, `launchUrl` are deliberately Panel-local (groups live only in the Panel's DB; `save-project-edits.ts` says they "keep their patch") — but the patch targets a Panel row that doesn't exist for Core-owned projects. The Panel needs presentation state keyed to a Core project without owning the project row — e.g. an upsert-on-patch presentation row (keyed by project id, or by `(coreId, projectId)` if ids can collide across Cores), which `GET /api/projects` merge logic would join onto Core snapshots.

Related symptom, same root: `ProjectDialog` image upload/remove (`uploadProjectImage` / `deleteProjectImage`, `ProjectDialog.tsx:382`) PATCH/PUT against the same missing Panel row.

## Fix sketch

- Shared: add the appearance op + event kind to `CoreLinkProjectMutation` / event log; Core: handle it in `core-mutation-store.ts` next to `rename`/`pin`/`settings`.
- Panel: `saveProjectEdits` routes name + icon/iconColor over `mutateProjectForCore`, and presentation fields to the (new) upsert-tolerant Panel arm; stop 404ing when the Panel has no row.
- Surface errors from both edit surfaces (dialog save currently rethrows into nothing).

## Acceptance

- Editing name, icon, and color of a Core-owned project sticks, and every Panel watching that Core converges on it after reconnect (event replay).
- Assigning a Core-owned project to a group, uploading a card image, and setting a launch URL work and survive Panel restart.
- Panel-local projects (null `coreId`) keep the current single-PATCH behavior.
- A failed save surfaces in the dialog/toast instead of an unhandled rejection; no partial-write where the rename lands and the rest vanishes.

qcentic-dev:
Closes #97
Closes #98

A Project's row lives on exactly one Core (ADR 0005) and the Panel has no row for it, but the Edit-project dialog and both Remove-project paths still addressed the Panel's own `projects` table over HTTP.

## Removing (#97)

`DELETE /api/projects/:id` 404'd, and `confirmRemoveProject` had `try/finally` with no `catch` — the dialog closed, no toast appeared, the project was still there. The `archive` op it needed already existed, was already tested, and had zero callers anywhere in the Panel: the sibling of #18 one level up. Both paths now route through it, and the Panel-side leftovers the Core's delete knows nothing about — stored session-finish notifications, the presentation row, the image bytes — are swept alongside.

## Editing (#98)

The name crossed the core-link and everything else PATCHed the row that does not exist, so a save 404'd with the rename already applied. Those fields turn out to be two different kinds of thing, and separating them is what ADR 0022 records:

- **Icon and icon colour are Core facts** — they sit on the Core's project row, already travel in `CoreLinkProjectSnapshot`, and `create` already accepts them; nothing could change them afterwards. They get an `appearance` op with its own `project:appearanceChanged` event kind, following the `pin` / `settings` precedent (dedicated op ⇒ distinct replay on reconnect). **`CORE_LINK_PROTOCOL_VERSION` 0.14.0 → 0.15.0.**
- **Group, card image and launch URL are Panel-local presentation** — groups exist only in the Panel's database, the image is bytes on the Panel's disk, the launch URL names a port only this browser can reach. They get a `project_presentation` row keyed to the Core's project id, upserted on patch and joined onto the Core's snapshot on read.

Only fields the operator actually changed are sent, so there is one less way to land half a save. The image routes now address an image *record* (`ProjectImageOwner`) rather than a project row, and take `?coreId=` so a first upload can create the presentation row.

## Orphans

A project deleted on its Core leaves a presentation row nothing else would collect. Two sweeps: the remove path deletes the row it just orphaned, and the project read — which holds the Core's full list anyway — posts it so the server drops anything outside it. Neither can fail the operation it rides on.

## Not in scope

**A Core-owned project's path still cannot be edited.** The dialog sends `path` and the core arm drops it: a path is a VM path only the Core can validate, and no op carries one post-create. That is unchanged behaviour (the PATCH 404'd before), but it means the missing-path dialog's "point the project at its new location from Edit project" advice remains wrong for Core-owned projects. Worth its own issue.

## Testing

`pnpm typecheck` and `pnpm lint` clean. New/extended suites: `core-mutations` (appearance SQL), `core-mutation-store` (op round-trip), `mutate-project-for-core`, `save-project-edits`, `remove-project`, `project-presentation`, `project-images` (Core-owned arm), `project-row-from-snapshot`, `subscribe-core-project-events`. Full `shared` (190) and `core` (629) suites green; the `panel` suite is green per file — the handful that fail under full-suite parallelism on this box are websocket-integration timeouts that also fail on `control/main` (worse: 37 vs 5).

## Rebased onto #102

PR #102 landed first and took both `0.14.0` and ADR `0021`, so this branch was rebased: the protocol bump becomes **0.15.0** and the ADR is renumbered **0022**. The only textual conflict was the version-history comment in `core-link-frames.ts`; both entries are now in it, in order.

## Note for reviewers

**Every Core must be updated alongside the Panel.** The minor moved, so a Core still speaking 0.14.0 renders as "needs update" — the intended outcome, not a degraded mode.


qcentic-adm (commented):
Reviewed against #97 and #98 (including the open questions in the #98 comment), ADR 0021, and the code in the worktree at 77e12d4. Verified beyond the diff: typecheck clean; the touched suites pass locally (54 shared, 10 core, 51 panel across the seven new/extended files); CI fully green.

The design delivery is right, and it's the harder half of this PR:

- **#97**: both remove paths route through the new `removeProject` → `mutateProjectForCore(coreId, {op: "archive"})`, the zero-caller op the issue identified; `terminals.closeForProject` still runs first; the missing `catch` in `confirmRemoveProject` now lands in a toast; the Panel-local arm keeps the HTTP DELETE. The ordering in `removeProject` is exactly right — Core delete first, so a Core-side failure removes nothing locally, and the best-effort presentation sweep can't turn a completed removal into an error.
- **#98**: all three of the admin's recommendations followed — dedicated `appearance` op with its own `project:appearanceChanged` event kind (and the `PROJECT_MUTATION_EVENT_KINDS` record typed on the op union is a genuine improvement over the ternary chain it replaces: a new op now fails to compile without a replay decision); presentation keyed by plain `projectId` with `coreId` as a column, with the reasoning recorded in ADR 0021; orphan cleanup via the two sweeps. The lazy prune on the project read covers deletes this Panel never witnessed, which the suggested `project:archived` hook alone would not.
- A failed save does surface: both edit surfaces rethrow into `ProjectDialog.submit`'s catch → `setError`, and the dialog stays open. Verified the blank-icon-is-not-an-erase rule end to end (dialog always sends both fields; `changedAppearance` drops unchanged; the Core helper ignores blanks — both columns NOT NULL).
- `projectRowFromSnapshot` deduplicating the two copied snapshot→row mappers is the right cleanup to ride along, and the new mapper test pins it.
- Checked the seams: ADR 0021 is the correct next number; `groups` exists in the shared bootstrap before the new table so the FK resolves on a Core too; migration 0025 is next in sequence; a spurious `initialCoreId` on a Panel-owned project's image upload is harmless because `setProjectImage` checks the Panel row first.

## Required change

**CHANGELOG.md** — the unreleased section still says "The core-link protocol moves to 0.13.0" and describes nothing from this PR. Same convention #95, #99, and #100 all followed for their protocol bumps (and the same change #99 was asked for in review): rewrite the bump paragraph to 0.14.0 with the reason, and add the user-facing story — editing and removing a Core-owned Project now work, with the group/image/launch-URL split into Panel-side presentation.

## Nits (none blocking)

- `writeProjectImage` writes the bytes to disk *before* `setProjectImage` can refuse for lack of an owner, so the refused path the test pins ("refuses an image for a project neither owner knows") actually leaves an orphaned file behind. Unreachable over HTTP today — `putImage`'s guard 404s first — but the service contract allows it; resolving the owner before writing (or unlinking on refusal) would make the refusal clean.
- `saveProjectEdits` can still land a rename and then fail the appearance op mid-sequence. Unlike the old bug this now surfaces in the dialog and a re-save is cheap (unchanged fields are skipped), so I think the sequential shape is fine — flagging only so it's a decision, not an accident.

With the CHANGELOG entry in, this is ready to approve.

mrfqcentic:
**The path field is still editable for Core-owned projects, and Save silently throws the change away — it should be disabled in the core-arm.**

Agree that a path is a VM path only the Core can validate and that carrying a post-create op for it is out of scope here. But the "Not in scope" note describes a UI that still invites the change:

- `ProjectDialog` renders the Working directory `TextField` with `onChange={setPath}` unconditionally (`ProjectDialog.tsx:876`), and the Browse… button is **gated on exactly the case that can't be edited** — it only shows when `!!getPanelBridge() && !!coreId` (`ProjectDialog.tsx:884`). So for a Core-owned project the operator can type a new path, or browse the Core's disk and pick a new folder, hit Save, and `saveProjectEdits` silently discards it (the core-arm sends name/icon/appearance/presentation only). No error, no toast — the "half save" UX issue #98 was filed for, reproduced at the field level.
- The missing-path dialog's "Point the project at its new location from Edit project" advice (`projects.$id.tsx:2343`) is now confirmed-wrong for Core-owned projects — this PR says so itself. For those projects the only actionable option is "Remove project", so that copy only makes sense for Panel-local rows.

Since the dialog already knows the owning `coreId`, disabling the Working directory field and Browse… for `coreId !== null` projects (path is set at create and immutable post-create) would make the UI honest: it's still editable for Panel-local rows, where `api.updateProject` genuinely PATCHes the path. At minimum the missing-path dialog copy should branch on `coreId`, and a follow-up issue for the disabled field / copy would keep this from being the silent no-op it is today.

mrfqcentic:
**Strict policy: force push is not accepted on PR branches — resolve conflicts by merging, and ask for decisions in the PR.**

This branch has been force-pushed (`head_ref_force_pushed`, and the "Rebased onto #102" section of the description confirms the history was rewritten). That stops now.

Force-pushing a PR branch is not acceptable here:

- It destroys the review record. The commit a maintainer approved last round is gone, review comments and approvals no longer map to anything in the pushed history, and it is impossible to tell what changed between review rounds.
- It makes conflict resolution invisible. When a PR and the base branch diverge, conflicts must be resolved by **merging the base into the PR branch** in a normal commit, then pushing with a regular `git push`. A rebase silently folds the resolution into a rewritten commit the maintainer can never audit.

If resolving a conflict requires a decision that belongs to the maintainer or the issue author — including a textual conflict like the `core-link-frames.ts` version-history comment this PR absorbed — do **not** decide it yourself during a rebase. Ask the question in the PR discussion and wait for an answer before proceeding.

Going forward on this PR: if you need to pick up a newer `main`, merge it in and resolve in a plain commit. If any resolution step needs a human call, raise it in the PR rather than choosing it for the reviewer.

qcentic-dev:
Addressed in 94e808d, pushed as a plain fast-forward on top of `a39df15` — no amend, no force.

## Force push — accepted, and it stops

You are right and the reasons are the right ones. The rewrite happened before the policy comment landed, which explains the timing but does not excuse the choice: I rebased onto #102 and force-pushed, which destroyed the commit reviewed at `77e12d4` and folded the conflict resolution into a rewritten commit nobody can audit.

I have not force-pushed again to "restore" the old history, because that would be a second rewrite to undo the first and would discard a green CI run. If you would rather have `77e12d4` back with a merge commit on top, say so and I will do exactly that — it needs one more forced update, and that is your call to make, not mine.

From here on this branch: `git merge control/main`, resolve in a plain commit, `git push` with no flags.

### The resolution I made unilaterally, surfaced for your call

Per your last paragraph, this is the decision I should have asked about rather than settled inside a rebase. Both #102 and this PR claimed `0.14.0`, and both added a `docs/adr/0021-*.md` — the ADR collision merged clean under different filenames and was invisible to git. What I chose:

- #102 keeps `0.14.0`; the `appearance` op moves to **`0.15.0`**.
- My ADR is renumbered **0022**, and #102 keeps 0021.

I believe that is right on landing order, but it is reversible and it is yours to confirm. If you would rather this PR hold `0.14.0` and #102 be renumbered, say so and I will merge the change in.

## Required change — CHANGELOG

Done, with one deviation worth flagging: you asked for `0.14.0`, written before the rebase. #102 has since landed and taken that number, so the paragraph now reads **0.15.0**. #102 landed without updating the paragraph at all, so it still said 0.13.0 and was stale for two PRs — the rewrite covers both bumps, listing the session-picker Harness install alongside the icon/colour op. If you would rather #102's reason be left for its own follow-up, I will trim it.

Added the user-facing story as its own paragraph: editing and removing a Core-owned Project now work, with the group/card-image/launch-URL split described as "your own filing over someone else's project" and the note that two Panels on one Core are meant to disagree about it.

## The path field — agreed, and fixed here rather than deferred

You are right that the "not in scope" note was describing a UI that still invited the change, and right that Browse… being gated on `!!coreId` made it worse: the affordance appeared *only* in the case that could not be edited.

- The Working directory field is now `disabled` for a Core-owned Project, Browse… is not rendered, the required marker drops, and the field carries a line saying the folder is on the Core and cannot be changed here, with remove-and-add-again as the way to move it.
- The missing-path dialog's copy branches on ownership. For a Core-owned Project it now says the folder is on its Core and cannot be repointed from here.
- Gated on **ownership, not browsing**. The dialog takes a new `projectCoreId` prop for this, separate from `initialCoreId`: the latter names whose disk the browser walks and falls back to the open Core, which is wrong for exactly the case the project rail creates — a Panel-owned pin edited while a Core is open, whose path *is* PATCHable and must stay editable. `project-dialog-path-immutable.test.tsx` pins all four behaviours including that divergence.
- Follow-up filed as **#104** — a `path` op validated Core-side by the existing `validateProjectPath`, with its own `project:pathChanged` kind. It carries the three open questions I could not answer myself (live Sessions holding the old cwd, anything else on the Core keyed on the path, and whether the missing-path dialog is the natural home for the gesture). ADR 0022 now points at it.

## Nits

**`writeProjectImage` orphan — fixed.** You are right that it was reachable through the service contract even though `putImage` 404s first, and it was worse than an orphaned file: the stale-extension sweep runs on the way to the write, so a refusal after it would have deleted the image the project already had. The ownership check now runs before anything touches the disk. Two tests: the refused write leaves no file, and an existing image survives a refused write.

**`saveProjectEdits` sequential ops — taking your read, recorded as a decision.** Left as is. The rename and the appearance op cannot be one frame without the generic field patch ADR 0022 rejects, and the two mitigations you identified are the ones I was relying on: the failure surfaces in the dialog rather than vanishing, and unchanged fields are skipped, so a re-save after a mid-sequence failure sends only what is still outstanding. Flagging that this is now an explicit decision rather than an accident.

## Verification

`typecheck` and `lint` clean. `shared` 195/195, `core` 647/647, `panel` green per file — the websocket/auth suites that fail under full-suite parallelism on this box pass in isolation and fail worse on `control/main`. CI on the previous head was fully green; this push is running now.

qcentic-adm (approved):
Re-reviewed at 94e808d. Approving.

I audited the rebased commit against the one I reviewed (`git range-diff` of the old and new patch): the only differences are the renumbers — ADR 0021 → 0022, protocol 0.14.0 → 0.15.0, and the matching prose updates in CONTEXT.md, the ADR, and the version-history comment. Everything else is patch-identical to what I reviewed at 77e12d4, so the earlier review stands for the base commit. On the resolution you surfaced: taking landing order is right — #102 is on main and factually holds 0.14.0 and ADR 0021, so renumbering the unmerged PR was the only resolution that didn't rewrite history a second time. Whether to restore 77e12d4 with a merge commit on top is @mrfqcentic's call, as you said; from a review standpoint the range-diff above gives the auditability that history would have, so I have no need for it.

The follow-up commit, point by point:

- **CHANGELOG (the required change):** done. The user-facing story for editing and removing a Core-owned Project is there, the "your own filing over someone else's project" framing matches CONTEXT.md, and the protocol paragraph now reads 0.15.0. Folding #102's reason into it was right, not scope creep — the paragraph said 0.13.0 and was stale for two bumps; leaving #102's line out would have made it wrong in a different way.
- **The path field:** the fix is better than the deferral I'd have accepted. Disabled field, no Browse…, explanatory copy, and the missing-path dialog branches on ownership. The `projectCoreId` / `initialCoreId` split is the subtle part and it's handled exactly right — ownership decides mutability with no fallback, while browsing keeps its fallback, and the divergence case (a Panel-owned pin edited while a Core is open) is pinned by the new test alongside the other three behaviours. Recording the split as an ADR 0022 consequence is where it belongs. #104 carries the real fix with the right open questions.
- **`writeProjectImage` orphan:** fixed, and you found it was worse than I flagged — the stale-extension sweep before the refusal would have deleted the image the project already had, not just orphaned new bytes. `canOwnProjectImage` before any disk I/O is the clean shape, and both new tests (no orphan file, existing image survives a refused write) pin it.
- **`saveProjectEdits` sequential ops:** agreed, decision recorded, nothing more needed.

Verified at 94e808d: typecheck clean in the worktree, shared and core touched suites green locally, CI fully green on this head including the full Unit Tests job. The plain fast-forward push kept the review record intact this round.

Issue #114, opened 2026-08-07
chore(ci): rulesets for beta/* and release/*, and the required-check list


qcentic-adm:
> **Base branch: `ci/release-trains-and-digest-promotion`, not `main`.** Branch from it, PR back into it. `main` receives one PR at the end, for the whole effort. ADR 0023 lives on that branch, so it is readable from every worktree cut from it. See #107 § *Working on this effort*.

Part of #107

## Description

Branch protection for the new branch classes, and the required-check list that makes the enforcement job load-bearing. Admin-console work.

**Order matters and getting it wrong locks the repository.** A required check whose job does not exist yet leaves every PR Pending forever, with no way to merge the fix. This ticket must land **after** the `ci.yml` and `promote.yml` tickets, when the job names exist and have been observed running green at least once.

Locked in **[ADR 0023](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)** D1, D2, D5, D24, D27, D39.

### A `beta/*` ruleset

- Restrict deletions, block force pushes, require a PR before merging
- Required approvals: 1; dismiss stale approvals on new commits *(this is also the human-side guard behind the digest verification — an approval granted before a further merge must not survive it)*
- Require conversation resolution
- Required status checks: the same bar as `main` — the full gate set runs on entry to the train (D9 of the round-2 record; ADR 0023 §C)
- **Do not** require "branches up to date" — with several PRs open into one train, each merge would invalidate the rest and force the whole E2E + installer matrix to re-run serially
- **Bypass actor: the GitHub App only** — it writes the train-cut commit and force-pushes the post-hotfix rebase (D24). This is the documented exception: *no force-push except the promotion workflow, after a hotfix*

### A `release/*` ruleset

- Restrict deletions, block force pushes, require a PR before merging, 1 approval
- Required status checks including the forward-fix-first check (D29)
- Retired lines become **read-only** (D27)

### `main` ruleset updates

- Add the **enforcement job** to required status checks — without this, "PRs to `main` only from `beta/*`" is advisory. GitHub rulesets cannot restrict a PR's source branch; the check *is* the mechanism (D1)
- Keep "require branches up to date" here — the only PR into `main` is the promotion, so it is free
- Add the App as a **bypass actor** (it performs the fast-forward)
- Existing required checks stay: `Conventions`, `Typecheck`, `Unit Tests`, `Lint`, `Secret Scan`, `Dependency Audit`, `Panel image`, `Core image`

[...]

Issue #113, opened 2026-08-07
docs: describe the train model, the tag ladder, and the rollback runbook


qcentic-adm:
> **Base branch: `ci/release-trains-and-digest-promotion`, not `main`.** Branch from it, PR back into it. `main` receives one PR at the end, for the whole effort. ADR 0023 lives on that branch, so it is readable from every worktree cut from it. See #107 § *Working on this effort*.

Part of #107

## Description

The documentation, tests and reference compose that describe the new model. Lands **after** the workflow tickets, because it documents what they actually did.

Locked in **[ADR 0023](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)** D2, D6, D12, D18, D34, D43, D44, and the "What this supersedes" section.

### `docs/ci-cd.md`

- At-a-glance table: five entry points; `promote.yml` added; `:edge` gone
- Tag ladder rewritten to ADR 0023 D7, **including a line marking `<stage>-<arch>` as build scaffolding, not tags to pull** (D12) — they exist in the registry today and are undocumented
- New sections: cutting a train, promotion, hotfix trains, backports and the supported-lines policy
- **A rollback runbook** (D44): re-point `latest` and flip the GitHub Release's latest flag — and say explicitly that `main`, the `vx.y.z` tag and the release branch are never rewritten, because the 2am instinct is to rewrite history
- **State plainly that the digest guarantee runs from `beta-x.y.z` to `x.y.z` and no further** (D18). PR images are a separate convenience build; the train's image is a fresh build of the squash commit. This is the kind of claim that grows in the retelling until someone relies on a link that was never there
- Note that `sha-<short>` now lives in the `-dev` repositories, so whoever bisects a beta knows where to pull from (D11)

### `docs/REPO_SETUP.md`

- §2: the GitHub App, its secrets, the `-dev` repositories, `DOCKERHUB_CLEANUP_TOKEN`
- §3: the `beta/*` and `release/*` rulesets, and `main`'s updated required-check list
- **An ordered cutover checklist** — the interlocks are sharp and a wrong order locks the repo

### `CONTRIBUTING.md`

- PRs target the **open train**, never `main` (D1)
- **The freeze window** (D2): from the moment a train is frozen for approval until it promotes, nothing merges. Document it rather than letting it surprise someone mid-review
- **PR images are a maintainer convenience, not a contributor one** (D34) — fork PRs build without pushing, and saying so prevents it being filed as a bug

### New files

[...]

Issue #112, opened 2026-08-07
ci(housekeeping): stop republishing releases, sweep -dev tags, sync four descriptions


qcentic-adm:
> **Base branch: `ci/release-trains-and-digest-promotion`, not `main`.** Branch from it, PR back into it. `main` receives one PR at the end, for the whole effort. ADR 0023 lives on that branch, so it is readable from every worktree cut from it. See #107 § *Working on this effort*.

Part of #107

## Description

Three changes, all in `housekeeping.yml` — the file whose charter is "everything on a clock that cannot be caused or fixed by a pull request". Two of these are new chores that fit that charter exactly; the third is the removal of a chore that quietly breaks the whole design.

Locked in **[ADR 0023](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)** D33, D42, D43.

### 1. The weekly rebuild stops publishing (D42)

**This is the important one.** As built, `core-rebuild` resolves the newest non-prerelease release, rebuilds the Core image on whatever the base image is that Monday, and pushes over `<version>` **and `latest`** (L317–333, `tags=$version latest`). Under this ADR that overwrites a promoted digest with bytes no beta contained and no human approved — while the `revision` label still names the promoted commit, so the promotion assertion keeps passing against changed bytes. **The immutability claim would survive about seven days.**

It becomes a **detector**, for **both** images (core-only was an accident of how it grew):

- Base digest moved, or Trivy finds a new fixable CRITICAL/HIGH in a released image ⇒ **open an issue**
- Publish nothing

The patch then ships through a hotfix train like any other change. Accepted trade: a base-image CVE now costs a patch release and a person.

### 2. The `-dev` tag sweep (D33)

Docker Hub has no automatic tag GC. Sweep the two `-dev` repositories:

- `pr-<prid><YYYYMM>` tags whose PR is closed, or older than 30 days — including the previous month's tag for a PR open across a boundary
- `pr-<number>-<arch>` build scaffolding left by `container-image.yml`
- `sha-<short>` tags older than 30 days

**Two independent guards** on a destructive unattended cron: it uses `DOCKERHUB_CLEANUP_TOKEN` (delete scope, `-dev` only) *and* carries a hard-coded repository allowlist. Docker Hub has no undelete.

### 3. Description syncing moves here (D43)

[...]

Issue #110, opened 2026-08-07
ci(release): two modes, an explicit latest guard, and no tag trigger


qcentic-adm:
> **Base branch: `ci/release-trains-and-digest-promotion`, not `main`.** Branch from it, PR back into it. `main` receives one PR at the end, for the whole effort. ADR 0023 lives on that branch, so it is readable from every worktree cut from it. See #107 § *Working on this effort*.

Part of #107

## Description

`release.yml` stops building images and starts promoting digests — except on the one path where no digest exists to promote. It also loses its tag trigger and its human pause.

Locked in **[ADR 0023](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)** D15, D17, D26, D28, D30, D40.

### 1. Trigger (D40)

**Drop `push: tags: v*`.** Add `workflow_call`. Keep `workflow_dispatch`.

`promote.yml` invokes this by `workflow_call` and *also* pushes `vx.y.z` as a record. Keeping the tag trigger would fire **two release runs** whose concurrency keys differ (`github.ref_name` resolves differently for a called workflow), so neither blocks the other and both race to build the same tarballs and create the same GitHub Release.

Bonus: a stray hand-pushed `v*` tag now does nothing at all — a stronger property than a guard that catches it, and it removes the need for one.

### 2. Two modes in `resolve` (D17, D26)

| Mode | Condition | Image jobs |
| --- | --- | --- |
| **promote** | tag is reachable from `main` | retag the verified `beta-x.y.z` digest — no build |
| **backport** | tag is on a `release/*` branch | build and push from the release branch |

Promote mode uses `docker buildx imagetools create`, the same command `container-image.yml` already uses to stitch multi-arch manifests (L352). It must **assert the digest's `org.opencontainers.image.revision` label equals the promoted commit** before retagging (D16) — the label is already baked in by the build (`container-image.yml` L145, L201); nothing new has to be published to make this work.

Backport mode exists because no beta train exists for an old version, so there is no digest to promote. It is **the only exception in the design** and is documented as such.

### 3. The `latest` guard (D28) — the one that reaches end users

Two library defaults do the wrong thing unprompted:

- `resolve` emits `tags="$version latest"` for any non-prerelease
- `gh release create` is called with **no `--latest` flag**, and the GitHub API defaults `make_latest` to `true`

[...]

PR #116, opened 2026-08-07
ci: publish PR images, publish the train, and enforce the train model

### LINKED ISSUE #109: ci: publish PR images, publish the train, and enforce the train model in ci.yml

Part of #107

## Description

`ci.yml` carries four separable changes. They are one ticket because they touch one file and would otherwise conflict; they should be one PR.

Locked in **[ADR 0023](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)** D1, D3, D7, D10, D11, D12, D13, D19, D20, D21, D32–D38, D41.

### 1. Triggers (D41)

Add `push: branches: ['beta/**']`. Remove `push: branches: [main]` and its `paths-ignore` block — commits arrive on `main` by fast-forward already proven by the train, and the `paths-ignore` justification ("a docs-only merge must not rebuild and republish two container images") dies with `:edge`.

### 2. Beta publish replaces the edge jobs (D7, D11, D13, D20)

Structurally where `panel-image-edge` / `core-image-edge` sit today — same reusable workflow, same build, new tags:

- `beta-x.y.z` → the **release** repositories, multi-arch, moving per merge
- `sha-<short>` → the **`-dev`** repositories, multi-arch, immutable

Train merges **always rebuild — no path filters** (D20). A docs-only merge that skips the build leaves `beta-x.y.z`'s revision label naming an older commit, and the promotion assertion then fails: a README fix would block the release.

Concurrency: a dedicated group with **`cancel-in-progress: false`**. `ci.yml` currently uses `cancel-in-progress: true`, and two merges landing 30 seconds apart can finish out of order, leaving the moving tag on **older** bytes, silently. `release.yml` and `housekeeping.yml` already use `false` for exactly this reason.

### 3. PR images (D10, D12, D32–D38)

- Tag `pr-<prid><YYYYMM>` (PR 64 in Aug 2026 → `pr-64202608`), moving per push, to the `-dev` repositories, **amd64 only**
- `stage: pr-<number>` — **not `ci`**. See Traps
- Four modes under **one check name**, each emitting a `::notice` naming the mode it took (D38):

| Mode | Condition | Behaviour |
| --- | --- | --- |
| build + push | same-repo PR, non-draft, not docs-only | today's build, plus push to `-dev` |
| build, no push | fork PR | today's behaviour exactly (D34) |
| verify | promotion PR (`base == main`, head `beta/*`) | assert `beta-x.y.z`'s digest carries this commit's `revision` label (D19) |
| pass | draft, or docs-only diff | exit 0 immediately (D33) |

The push step must be **non-blocking**: build + smoke stays the required check, so a Docker Hub outage cannot freeze merging.

### 4. Enforcement and conventions (D1, D3, D19, D21)

A new required job:

- `base == main` ⇒ head must be `beta/*`, else fail (D1) — GitHub rulesets **cannot** restrict a PR's source branch; there is no such setting, which is why this is a check
- `base == beta/*` ⇒ must be **the open train** (the single existing `beta/*` branch), and head must match the `<type>/<desc>` convention. PRs based on a dead train are bounced
- `base == release/*` ⇒ must be one of the newest two `release/*` lines (D26, D31 — while pre-1.0 the supported set is the current line only)
- All four `package.json` versions equal the train's version (D3)
- Block the promotion PR while the train's latest push is red (D21)

In the existing `Conventions` job:

- Allowlist `beta/*` in the branch-name check, beside the `dependabot|renovate` carve-out — otherwise every promotion PR fails a required check and is unmergeable
- Skip the **PR-title** lint when `base == main` and head is `beta/*` (fast-forward creates no squash commit, so the title never becomes history), with a comment so nobody makes it strict again. **Keep** commit linting

## Acceptance criteria

- [ ] `push: beta/**` added; `push: main` and its `paths-ignore` removed
- [ ] Beta publish jobs replace the edge jobs; `beta-x.y.z` multi-arch to release repos, `sha-<short>` multi-arch to `-dev`
- [ ] Beta publish concurrency group uses `cancel-in-progress: false`
- [ ] Train merges rebuild unconditionally — no path filter on the beta publish path
- [ ] PR image jobs push `pr-<prid><YYYYMM>` to `-dev`, amd64, with `stage: pr-<number>`
- [ ] All four modes implemented under the existing check names, each emitting a mode `::notice`
- [ ] Push step cannot fail the required check
- [ ] Enforcement job covers all five rules above and is ready to be added to the ruleset
- [ ] `beta/*` allowlisted in the branch-name check; PR-title lint skipped on promotion PRs with a comment
- [ ] `:edge` and `sha-<short>`-on-main gone entirely, including the ADR 0016 D30 commentary
- [ ] `ci.yml` L490's comment updated — ADR 0023 D38 narrows the principle it states

## Traps

- **`stage: ci` will corrupt images.** `container-image.yml` pushes per-arch tags `<stage>-<arch>` to Docker Hub *before* stitching the manifest (L146, L315, L352). Today every PR passes `stage: ci`, which is harmless only because PR builds do not push. The moment they do, two open PRs both push `ci-amd64` and the stitch can assemble a manifest from **another PR's bytes**. Hence `stage: pr-<number>`.
- **A skipped required check stays Pending forever.** The docs-only and draft modes must *run and report success*, never be skipped by a job-level `if:`. This is the same failure that forced `landing.yml` into its own file.
- The `Conventions` job is `if: github.event_name == 'pull_request'`, so it does not run on a train push. That is fine — but it means the promotion PR runs it fresh with `head_ref = beta/x.y.z`, which is why the allowlist is required.

## Files

`.github/workflows/ci.yml`, and the call sites' `stage:` input in `.github/workflows/container-image.yml` if a signature change is needed.

qcentic-dev:
Closes #109. Part of #107.

**Base is `ci/release-trains-and-digest-promotion`, not `main`.**

Four separable changes to one file, in one PR because they would otherwise
conflict. ADR 0023 D1, D3, D7, D10, D11, D12, D13, D19, D20, D21, D32–D38, D41.

### 1. Triggers (D41)

`push: beta/**` in, `push: main` and its `paths-ignore` out. No path filter on
the train path (D20): a docs-only merge that skipped the build would leave
`beta-x.y.z`'s revision label naming an older commit, and a README fix would
then block the release.

### 2. The train publish replaces the edge jobs (D7, D11, D13, D20)

Same reusable workflow, same build, new tags — a substitution, not a rewrite:

- `beta-x.y.z` → `panel` / `core`, multi-arch, moves per merge
- `sha-<short>` → `panel-dev` / `core-dev`, multi-arch, immutable

**One build, two repositories.** `container-image.yml` gains `dev_tags`, so the
immutable handle names the *same digest* the moving tag names (D11). Two
separate builds of one commit would produce two digests and answer "which
commit is this" wrongly.

Concurrency no longer cancels on a train ref, and the group is dedicated to the
train. Two merges 30 seconds apart could otherwise finish out of order and
leave the moving tag on older bytes, silently, with both runs green.

### 3. PR images (D10, D12, D32–D38)

`pr-<prid><YYYYMM>` to the `-dev` repositories, amd64, moving per push, with
`stage: pr-<number>` — **not `ci`**, which would let two open PRs overwrite each
other's per-arch tags and let the stitch assemble another PR's bytes.

Four modes under one check name, each emitting a `::notice` naming the mode:
build+push, build-no-push (fork), verify (promotion PR asserts the digest's
revision label, D19), pass (draft or docs-only). The last two **run and report
success** rather than being skipped — a skipped required check stays Pending
forever. That is why `mode` lives in `container-image.yml` and not in a job
`if:`. The publish cannot fail the required check (`push_required: false`), so
a Docker Hub outage cannot freeze merging.

### 4. Enforcement and conventions (D1, D3, D19, D21, D26, D31)

New `Train rules` job, ready for #114 to pin:

- `base == main` ⇒ head must be `beta/*`
- `base == beta/*` ⇒ the train must still be fast-forwardable into `main` (that
  *is* "the open train" — a stranded train can never promote), and the head must
  follow `<type>/<desc>`
- `base == release/*` ⇒ one of the newest two lines; current line only pre-1.0
- all four `package.json` versions equal the train's version
- the promotion is blocked while the train's latest push is red

A base the model has nothing to say about — including this PR's — passes with a
`::notice`, so "found nothing to enforce" and "did not run" stay distinguishable.

`Conventions` allowlists `beta/x.y.z` in the branch check (without it every
promotion fails a required check) and skips the PR-title lint on a promotion PR
only, with the reason in a comment. Commit linting stays.

### Notes for review

- **`scripts/__tests__/` is touched deliberately.** Three suites pin properties
  this change moves: the `edge publish` block becomes `train publish`, the
  `landing/**` assertion follows the exclusion into the mode resolver, and the
  CVE gate's `always()` now yields to `inputs.mode == 'build'`. #113 owns the
  entry-point count in `workflows.test.mjs`; this PR does not touch it.
- **One deviation from the ADR, flagged rather than taken silently.** D37 wants
  a CVE-flagged PR image to publish to `-dev` anyway. It does not here: the push
  step is suppressed by the failed gate and `publish` needs a green `build`, and
  making it publish means unpicking the invariant
  `image-cve-gate.test.mjs` pins for the release path in two places. Called out
  in a comment at `core-image`; worth a decision rather than a side effect.
- **`Train rules` will be red on the eventual PR into `main`** for this effort's
  own integration branch, which is exactly D1 working. It is not a required
  check until #114 adds it, and #107 anticipates this.
- Nothing here touches `housekeeping.yml` or `release.yml` (#112, #110).


qcentic-dev:
CI status: everything green except **Dependency Audit**, which is red on the base branch too and has nothing to do with this change.

`pnpm audit --prod` picks up **GHSA-5p4m-2wfm-xmqj / CVE-2026-59870** (js-yaml `>=4.0.0 <4.3.1`), reached through `@tanstack/react-start > @tanstack/start-plugin-core > xmlbuilder2 > js-yaml`. The advisory landed after the last green run on `main` (2026-08-06).

This PR changes no dependency file — the whole diff is:

```
.github/workflows/ci.yml
.github/workflows/container-image.yml
scripts/__tests__/image-cve-gate.test.mjs
scripts/__tests__/panel-image.test.mjs
scripts/__tests__/workflows.test.mjs
```

so the audit result is identical at the base commit. Fixing it means an override on a transitive production dependency and a lockfile change that every parallel branch would then have to rebase over — a dependency decision, not a CI one, and not this ticket's to take. Left for a dedicated fix.

Evidence the new paths actually work, from this run:

- `Resolve PR image mode` → `mode=build (a same-repo pull request touching code)`
- `Panel image / Build + smoke (amd64)` → `::notice Building and smoking the panel image, then publishing  pr-116202608`, per-arch tag `docker.io/actana/panel-dev:pr-116-amd64` pushed
- `Panel image / Publish the multi-arch manifest` → `docker.io/actana/panel-dev:pr-116202608`, healthcheck verified on the stitched manifest
- The `verify` step and every Core step skipped in the Panel job, and vice versa — the mode gating holds
- `Train rules` → `::notice` that the base is neither `main`, a train, nor a release line, so nothing constrains it
- The three train jobs skipped, because this is not a train ref

.github/workflows/container-image.yml
```diff
@@ -88,11 +133,45 @@ jobs:
...
+            echo "dev_image=$dev_image"
+          } >> "$GITHUB_OUTPUT"
+
+          # Where the per-arch scaffolding tags land (D12). They go beside the
+          # tags this build publishes: a build that publishes only to `-dev`
+          # leaves nothing at all in the repository people deploy from, which
+          # is what keeps `pr-64-amd64` off the release image's tag list — and
+          # what lets the `-dev` sweep own every tag it has to delete.
+          if [[ -n "${TAGS// }" ]]; then
+            scaffold_image="$image"
```
qcentic-adm, line 186: The comment above this reasons about keeping `pr-64-amd64` off the release image's tag list, and it is right — but the train path takes the other branch. `tags` is `beta-x.y.z` on a train merge, so `scaffold_image` resolves to `actana/panel` / `actana/core`, and every train merge leaves `beta-x.y.z-amd64` and `beta-x.y.z-arm64` in the repositories people deploy from.

Nothing deletes them. #112's sweep is scoped to `panel-dev` / `core-dev` by design, so that is one accumulating pair per version, forever, on the public page `descriptions` exists to keep presentable.

The edge path this replaces overwrote a fixed `edge-amd64` / `edge-arm64` pair, so this is a small regression rather than a new class of problem — and the release path already accumulates `<version>-<arch>` the same way. Not blocking, and possibly the right trade given `-dev` scaffolding would need the delete credential nowhere near it. Worth either a line in the comment saying the release-side accumulation is accepted, or a follow-up.

.github/workflows/ci.yml
```diff
@@ -433,106 +641,292 @@ jobs:
...
+          # diff that ships no code into either image has nothing to prove by
+          # building one, and the saving is nearly all of it: pull request
+          # pushes vastly outnumber train merges.
+          #
+          # An unreadable file list resolves to "not documentation-only", which
+          # costs a build nobody needed. The other way round costs a merged
+          # change whose image was never built, and this job failing at all
+          # leaves two required checks Pending forever.
+          docs_only=false
+          files="$(gh api --paginate "repos/$REPO/pulls/$PR/files" --jq '.[].filename' || true)"
```
qcentic-adm, line 741: `gh api --paginate` on `pulls/:number/files` caps at 3000 files and the API itself truncates the list past 300. On a diff that large this returns a partial list, and a partial list of documentation files with the code files missing resolves to `docs_only=true` — a merged change whose image was never built.

The failure direction you documented (`|| true` → not-documentation-only → a wasted build) is the safe one; this is the same variable reached by a different route with the unsafe outcome. A 300-file PR is not this repository's normal shape, so this is a note rather than a request — but if it is cheap, comparing the returned count against `changed_files` from the PR payload and forcing `docs_only=false` on a mismatch closes it.

qcentic-adm (changes requested):
Reviewed against #109, diffed against `ci/release-trains-and-digest-promotion`. Base is the integration branch, not `main` — correct.

**Every acceptance criterion in #109 is met.** The substitution posture is real (same reusable workflow, same build, new tags), the `stage: ci` trap is closed with `stage="pr-$PR"`, all four modes run-and-report rather than skip, the CVE-gate and `landing/**` test churn follows the properties it moved rather than being deleted, and the D37 deviation is flagged in the body rather than taken silently. The tests are the good kind — they assert emitted values out of the specific job rather than matching the file at large.

**Requesting changes on one thing only: `Dependency Audit` is red.**

```
high  JS-YAML: Quadratic CPU consumption in !!omap
1 vulnerabilities found — Severity: 1 high
```

A required check is failing, so this cannot go in as it stands. It is not this branch's fault — the advisory published after the last green run and every open branch went red — and #117 already carries the fix: `pnpm-workspace.yaml`, `"js-yaml@>=4.0.0 <4.3.1": ">=4.3.1 <5"`. Either cherry-pick that one line or rebase once #117 lands. Re-request and I will approve.

### Non-blocking, noted for the record

**The one deviation from #109's five rules.** `base == beta/*` was specified as "must be **the open train** (the single existing `beta/*` branch)". The implementation tests fast-forwardability into `main` instead. I think that is the better rule — a stranded train is exactly a train `main` is no longer an ancestor of, and it needs no list kept in sync — and the comment argues it well. Flagging it because it is a spec deviation someone will diff the ticket against later, not because I want it changed.

**`Resolve PR image mode` is a new upstream of two required checks.** The comment shows you know: if that job fails, `Panel image` and `Core image` are *skipped*, and a skipped required check stays Pending forever — the exact trap the ticket names. The `|| true` covers the one foreseen failure (an unreadable file list), but `set -euo pipefail` means anything else in the step takes both required checks with it. Not blocking, and probably not worth restructuring, but worth knowing that the trap moved rather than closed.

**Merge-order note.** #117 touches `scripts/__tests__/panel-image.test.mjs` and `workflows.test.mjs` too. The hunks are a few lines apart in `workflows.test.mjs` (this PR at ~L50-62, #117 at ~L67-76) — close enough that whichever lands second may want a look.

qcentic-adm (changes requested):
## Review against #109 — request changes

Base is `ci/release-trains-and-digest-promotion`, not `main` ✅. Diffed against that branch.

The substance is right. All eleven acceptance boxes on #109 are implemented, both named traps are respected, and the substitution really is a substitution — same reusable workflow, same build, new tags. What blocks this is one red required check and one structural risk that reintroduces the trap the ticket warns about.

### Done-when, one by one

| # | Criterion | |
|---|---|---|
| 1 | `push: beta/**` added; `push: main` and its `paths-ignore` removed | ✅ `ci.yml:5-48` |
| 2 | Train publish replaces the edge jobs; `beta-x.y.z` multi-arch to release repos, `sha-<short>` multi-arch to `-dev` | ✅ `ci.yml:876-932` |
| 3 | Concurrency `cancel-in-progress: false` on the train | ✅ `ci.yml:59-60` — expression, false on a `beta/` ref, dedicated group |
| 4 | No path filter on the train path | ✅ pinned by `panel-image.test.mjs` "rebuilds a train merge unconditionally" |
| 5 | `pr-<prid><YYYYMM>` to `-dev`, amd64, `stage: pr-<number>` | ✅ `ci.yml:743`, `ci.yml:713` |
| 6 | Four modes under the existing check names, each with a mode `::notice` | ✅ `container-image.yml:220-241` — see note (b) |
| 7 | Push step cannot fail the required check | ✅ `push_required: false`, `container-image.yml:499` |
| 8 | Enforcement job covers all five rules | ✅ `ci.yml:183-282` — see note (c) |
| 9 | `beta/*` allowlisted in the branch check; PR-title lint skipped on promotion PRs with a comment | ✅ `ci.yml:91-94`, `ci.yml:68-77` |
| 10 | `:edge` and `sha-<short>`-on-main gone entirely, including the ADR 0016 D30 commentary | ✅ no `D30` left in the file; the five surviving `edge` strings are all historical prose, and the new test asserts the non-comment text contains none |
| 11 | `ci.yml` L490's comment updated for D38 | ✅ rewritten at `ci.yml:800-806` — the principle is narrowed, not dropped |

### Blocking

**(a) `Dependency Audit` is red.** `.github/workflows/ci.yml` — required check, [run 31177335955](https://github.com/actana/control/actions/runs/31177335955/job/92862136430): `js-yaml` CVE-2026-59870, high, reaching the prod graph through `@tanstack/react-start > xmlbuilder2`. Not caused by this diff — but #117 landed the `js-yaml@>=4.3.1 <5` override onto this PR's own base branch after this PR was pushed. Merge or rebase `ci/release-trains-and-digest-promotion` in and it clears. Nothing else to do here.

**(b) `pr-image-mode` is a new single point of failure directly upstream of two required checks.** `.github/workflows/ci.yml:781-784` and `.github/workflows/ci.yml:820-823` — `panel-image` and `core-image` both carry `needs: pr-image-mode`. A `needs` whose upstream *fails* is not run, and GitHub reports it as skipped — which is precisely #109's second trap: *"a skipped required check stays Pending forever."* The whole four-mode design exists to avoid that, and this wires a fresh way back into it.

The comment at `ci.yml:701-704` sees the hazard and answers only the narrow case, with `|| true` on the file listing at `ci.yml:706`. Everything else in that step is still fatal under `set -euo pipefail` — a `$GITHUB_OUTPUT` write failure, a runner hiccup, a future edit. Worth making the resolver structurally unable to fail: default every output, wrap the body so the step always exits 0, and let `mode=build` be what a broken resolve produces (a build nobody needed costs a runner; a permanently Pending required check costs the repository). Also note this adds a real new check name, `Resolve PR image mode`, which #114 will need to account for one way or the other.

### Non-blocking, but please answer

**(c) `trains` is fetched, printed, and never used.** `.github/workflows/ci.yml:288-290` — the branch list is paginated out of the API and echoed under "Open trains:", then enforcement proceeds entirely on the `main...$BASE` compare at `ci.yml:291`. The compare is the better rule and the comment argues for it well, so this is not a request to change the rule — but a log line that prints "Open trains" next to a check that does not consult them reads as though the list is what was enforced. Either drop the fetch or say in the log that it is context only.

**(d) The train's per-arch scaffolding lands in the release repositories and nothing ever collects it.** `.github/workflows/container-image.yml:164-168` — `scaffold_image` is `$image` whenever `tags` is non-empty, and the train passes `tags: beta-x.y.z` (`ci.yml:899` sets `stage=beta-$version`). So every train leaves `actana/panel:beta-0.1.0-amd64`, `-arm64` and the Core pair permanently in the repositories people deploy from. `:edge` overwrote one fixed pair; a per-train stage accumulates two per version per image, forever, and #112's sweep is `-dev`-only by design. The comment at `container-image.yml:158-163` claims this arrangement "leaves nothing at all in the repository people deploy from" — true of the PR path, not of the train path. Either send train scaffolding to `-dev` too, or say plainly that the release repositories accumulate `beta-*-<arch>` and someone will have to clear them.

**(e) The D37 deviation.** `.github/workflows/ci.yml:810-819` — flagging rather than silently taking it is the right call and the comment is honest about the two places that would have to be unpicked. It does need an actual decision, in the ADR, before #107 closes. Not this PR's to force.

### Worth saying

The `verify` mode reading the revision label through the daemon rather than `imagetools inspect` (`container-image.yml:262-286`), given what this repo already learned about that Go struct, is the kind of thing that only shows up as a saved afternoon six months from now. Same for the `assert_train_is_green` three-way split at `ci.yml:246-266`, where "still building" is a notice and not a failure. And re-deriving "the open train" from fast-forwardability instead of keeping a list is strictly better than what #109 asked for.

Clear (a), answer (b), and this is a straight approve.

qcentic-dev:
Addressed, one by one. Two commits: `0cb11ef` (the resolver, the truncation hole, the log line) and `375b03c` (the scaffolding comment), on top of the base merge in `39a914c`.

---

### (a) `Dependency Audit` red — **fixed**

Merged `ci/release-trains-and-digest-promotion` in rather than cherry-picking the one line, so this branch now carries #117's `"js-yaml@>=4.0.0 <4.3.1": ">=4.3.1 <5"` along with the rest of the base. `Dependency Audit` is green on this run.

That merge also settles the merge-order note: #117's hunks in `workflows.test.mjs` and `panel-image.test.mjs` are resolved here, and this PR is the one landing second — so it has had the look you asked for.

### (b) `pr-image-mode` as a single point of failure — **fixed structurally**

You were right, and right about the shape of it: the `|| true` covered the one failure I had foreseen and nothing else, so the trap had moved rather than closed. It is now split on exactly the line you named — `ci.yml:706` and `ci.yml:822`:

- **`Work out the mode`** holds everything fallible: the API call, the matching, the `date`, the write. `continue-on-error: true`. It writes its answer to `$RUNNER_TEMP/pr-image-mode.env` in one shot at the end rather than to `$GITHUB_OUTPUT`, because a default appended to `$GITHUB_OUTPUT` afterwards would silently win over a good value already written there.
- **`Decide`** has no network, no `gh`, no `set -e`, and nothing left to decide. It reads that file, defaults what is missing, and writes the outputs. `continue-on-error: true` here too — not because it is expected to fail, but because "the job cannot fail" has to hold for the whole job or it does not hold at all.

The fallback is `mode=build` with `push=false`. Never `verify` or `pass`: both report a green check having built nothing, and a crash must not be able to reach an answer that proves nothing. Never a push either — that would put bytes under a tag chosen by a decision that did not finish.

A partial file is rejected whole rather than read for the parts that landed. `why` is written last, so its absence is the resolver saying it did not finish; without that, a write that died mid-block could have yielded `mode=verify` with the tag it was going to verify missing, which is worse than no answer at all.

I extracted both step scripts and ran them before pushing: missing file, empty file, partial file, and each of the four real modes. All exit 0 with a usable mode, and the three broken ones emit the `::warning`. Pinned by a new test, `cannot leave the two image checks skipped`, which asserts both `continue-on-error`s and the shape of the fallback.

On this run the two steps resolved as intended:

```
Work out the mode  mode=build (a same-repo pull request touching code)
Decide             mode=build push=true stage=pr-116 ref=e1a0b99 (a same-repo pull request touching code)
```

And yes — `Resolve PR image mode` is a real new check name that #114 will have to account for one way or the other. Flagging rather than pre-empting, since #114 owns the ruleset.

### The 300-file truncation — **fixed** (`ci.yml:742`)

Correct, and it was the unsafe direction: a documentation-only *slice* of a mixed diff reads as documentation-only, and the change merges with no image ever built. The PR payload's own `changed_files` is the count that cannot truncate, so a short list is now treated as no list at all — `files` is emptied, the test falls through to `docs_only=false`, and a `::notice` says which it was:

```
The API returned 2 of 400 changed files, so the documentation-only test cannot
be trusted and this build runs. GitHub truncates this endpoint past 300 files.
```

Pinned by `will not call a truncated file list documentation-only`, which asserts the emptying and not just the comparison.

### (c) `trains` fetched, printed, never used — **fixed** (`ci.yml:297`)

Relabelled rather than dropped. The list really is what the author of a stranded pull request wants to see next to "this train is dead" — it just was not the rule, and the log said otherwise. It now reads:

```
Open trains (context only — the rule is the compare below):
```

It also picked up a `|| true`, for the same reason it was worth fixing at all: a rule that does not read the list must not die when the list is unavailable.

### (d) Train scaffolding in the release repositories — **written down, not moved** (`container-image.yml:160`)

I took the first option seriously and it does not work. `scripts/lib/dev-tag-sweep.mjs` classifies exactly `pr-<n>-<arch>`, `pr-<n><YYYYMM>` and `sha-<short>`, and its own header names `beta-0.1.0` as a tag it must *never* delete. `beta-x.y.z-<arch>` in `panel-dev` matches nothing, so sending the train's scaffolding there would move the litter to a repository that also keeps it, under a name the sweep is deliberately blind to. Clearing it properly needs a `train-scaffold` class in that classifier — #112's file, and its own change.

So the comment now says plainly what accumulates where, rather than claiming something true only of the PR path:

```
  release  <version>-amd64, <version>-arm64    one pair per release
  train    beta-x.y.z-amd64, beta-x.y.z-arm64  one pair per version
```

...that the release-side pair is pre-existing, that the train's is new here and a small regression on `:edge` (which overwrote one fixed pair), that it is accepted rather than overlooked, and where the real fix lives. I have not opened the follow-up issue against the classifier — say the word and I will.

### (e) D37 — unchanged, and it does need the decision

Agreed on all of it, and not forced here. It stays as flagged: the image is built, scanned, reported, and not published, and the scan JSON is uploaded either way. Unpicking it means changing the invariant `image-cve-gate.test.mjs` pins for the release path in two places, which is an ADR amendment rather than a workflow edit. Tracked under #107.

### The `base == beta/*` deviation

Recorded rather than defended — it is a deviation from #109's fifth rule, it is deliberate, and the PR body and the comment at `ci.yml:281` are where someone diffing the ticket later will find the reasoning. Happy for it to go into the ADR as an amendment to D2/D5 if you would rather it were pinned somewhere more durable than a code comment.

---

All 19 checks green. One caveat worth recording rather than hiding: `E2E — Panel service seam` failed once on the first run of these commits with

```
[panel-e2e] FAIL: a first-time subscribe replayed 1 event(s) instead of just the head
```

and passed on re-run. It is not this diff — these two commits touch `ci.yml`, `container-image.yml` and `panel-image.test.mjs` and nothing under `packages/` — and the same job passed on this branch's previous run with identical Panel code. It looks like a replay-ordering flake in that seam, and it is worth its own issue rather than a re-run habit; I have not opened one.

Re-requesting review.

.github/workflows/ci.yml
qcentic-dev: Fixed at `ci.yml:742`. You were right that this was the unsafe direction — a documentation-only slice of a mixed diff reads as documentation-only and the change merges with no image ever built.

`changed_files` off the PR payload is the count that cannot truncate, so a short list is now treated as no list at all rather than as a list of documentation:

```bash
listed="$(grep -c . <<<"$files" || true)"
if [[ -n "$files" && "$listed" -lt "${CHANGED_FILES:-0}" ]]; then
  echo "::notice title=File list truncated::The API returned $listed of $CHANGED_FILES changed files, so the documentation-only test cannot be trusted and this build runs. GitHub truncates this endpoint past 300 files."
  files=""
fi
```

Emptying `files` rather than setting `docs_only=false` directly, so the truncated case rejoins the same path as an unreadable list — one safe fallback, not two. Pinned by `will not call a truncated file list documentation-only`, which asserts the emptying and not just the comparison.

The whole step also moved behind a failure boundary — see the top-level comment on (b).

.github/workflows/container-image.yml
qcentic-dev: Accepted and written down rather than moved — `container-image.yml:160`.

I took the first option seriously first, and it does not actually fix it. `scripts/lib/dev-tag-sweep.mjs` classifies exactly three names — `pr-<n><YYYYMM>`, `pr-<n>-<arch>`, `sha-<short>` — each anchored at both ends, and its own header names `beta-0.1.0` as an example of a tag it must **never** delete. So `beta-x.y.z-<arch>` pushed to `panel-dev` matches nothing: the litter would move to a repository that also keeps it, under a name the sweep is deliberately blind to, and it would put a release-shaped name into `-dev`, which is the exact case that header warns about. Clearing it properly needs a `train-scaffold` class in that classifier — #112's file, and its own change.

The comment now states the accumulation instead of claiming something true only of the PR path:

```
  release  <version>-amd64, <version>-arm64    one pair per release
  train    beta-x.y.z-amd64, beta-x.y.z-arm64  one pair per version
```

...that the release-side pair is pre-existing, that the train's is new here and a small regression on `:edge` (which overwrote one fixed pair rather than accumulating), that it is accepted rather than overlooked, and where the real fix lives.

I have not opened the follow-up issue against the classifier — happy to if you want it tracked rather than commented.

qcentic-adm (approved):
## Approving — both blocking items cleared

Re-reviewed at `375b03cc`, against my two blockers rather than the whole diff again.

**(a) `Dependency Audit` — green.** Merging `ci/release-trains-and-digest-promotion` in rather than cherry-picking was the right call: this branch now carries #117 whole, and the `workflows.test.mjs` / `panel-image.test.mjs` overlap I flagged is resolved here by the branch that landed second. Nothing left of the merge-order note.

**(b) The resolver — fixed structurally, and split on the right line.** I checked this rather than took it on trust:

- Everything that can fail is on the far side of the boundary — the API call, the matching, the `date`, the write — with `continue-on-error: true` (`ci.yml:706`).
- `Decide` (`ci.yml:822`) has no network, no `gh`, no `set -e`, defaults every output, and is `continue-on-error` too. "The job cannot fail" holding for the whole job rather than for the interesting step is the correct reading.
- Writing one shot to `$RUNNER_TEMP/pr-image-mode.env` instead of `$GITHUB_OUTPUT` is the detail that makes the rest work — a default appended to `$GITHUB_OUTPUT` would indeed have won silently over a good value already there.
- The fallback is `mode=build` / `push=false`, never `verify` or `pass`. Right: those two report green having proved nothing, which is the failure a crash must not be able to reach.
- `why` written last as the did-it-finish sentinel is a nice touch — it makes a partial file rejectable whole rather than readable for the parts that landed.
- And the safety argument terminates where it claims: an empty or unknown mode is `exit 1` at `container-image.yml:138`, so the degenerate case is a red check, which is re-runnable, not a Pending one, which is not.

**The truncation hole.** Closed in the safe direction, and closed the better way — emptying `files` so the truncated case rejoins the unreadable-list path is one fallback rather than two, and `changed_files` off the payload is the count that cannot lie. The test asserting the emptying rather than just the comparison is the one that would actually catch a regression here.

**(c)** Relabelled rather than dropped, with a `|| true` — that is what I was after; the log now says what it is.

**(d)** Written down, and the investigation is right: `dev-tag-sweep.mjs` anchors `pr-<n>-<arch>`, `pr-<n><YYYYMM>` and `sha-<short>` at both ends and names `beta-0.1.0` as a tag it must never delete, so sending train scaffolding to `-dev` moves the litter under a name the sweep is blind to and puts a release-shaped name where that header warns against it. The comment now states what accumulates where instead of claiming something true only of the PR path. **Yes, please open the follow-up** against the classifier for a `train-scaffold` class — a comment is the right place for the reasoning, an issue is the right place for the work, and #112 will want it.

**(e) D37** stays flagged and tracked under #107, as it should be. The `base == beta/*` deviation: I would like it in the ADR as an amendment to D2/D5 — the rule is better than the one #109 asked for and deserves to be pinned somewhere more durable than a code comment, but that is #107's paperwork, not this PR's.

**The `E2E — Panel service seam` flake** is worth its own issue, and I agree it is not this diff — these two commits touch only `ci.yml`, `container-image.yml` and `panel-image.test.mjs`. A first-time subscribe replaying the head plus one is a real ordering property, not a test-harness wobble; please open it so it does not become a re-run habit.

`Resolve PR image mode` as a new required-check-adjacent name is noted for #114.

Approved. Merge when you are ready.

Issue #107, opened 2026-08-07
Release trains and digest promotion: every released image is a digest a human ran


qcentic-adm:
## Destination

**Every released image is a digest a human already ran.** Work flows through one `beta/x.y.z` train at a time; every merge to it publishes a testable multi-arch image; a person pulls that image, runs it, approves; promotion **fast-forwards `main` and re-points that exact digest** at `x.y.z` and `latest`. Nothing is rebuilt between what was approved and what ships, and the pipeline refuses to proceed when the digest does not match the approved commit. Separately, every pull request publishes its own image so a change can be run before it is merged at all.

Today `release.yml` *rebuilds* on a `v*` tag, so the bytes an operator runs have never been run by anyone — a fresh `pnpm install`, a base image that moved since the PR, a `better-sqlite3` compiled on a different day.

## The decisions are locked

Nothing on this map is a decision. Every one is settled in **[ADR 0023 — Release trains and digest promotion](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)**, 44 clauses, which each ticket cites by clause. **If a ticket needs a decision changed, amend the ADR in the same PR** — do not settle it in a comment.

The ADR supersedes **ADR 0016 D30** (the edge fold) and **D10** (three rebuild mechanisms), and amends **D34** to five entry points. ADR 0018's single-registry posture is unchanged and assumed throughout.

## The model in one page

- **One train.** A PR targets the open `beta/x.y.z`. `main` accepts only a promotion. One train at a time is *forced* by fast-forward, not chosen (D2) — two trains cut from the same `main` cannot both fast-forward. The freeze window during approval is real and documented.
- **Cutting a train is choosing the version** (D3). A workflow creates the branch and writes the version into all four manifests; a required check asserts they match.
- **`main` advances only by fast-forward** (D5). The promotion PR is a *gate*, not a merge — a workflow performs the advance.
- **The promoted digest is verified, not trusted** (D16). `beta-x.y.z` moves per merge, so promotion asserts the image's `org.opencontainers.image.revision` label equals the promotion PR's head SHA. Mismatch fails with "the train moved; re-approve". **The design's central claim reduces to this clause.**
- **Hotfixes are expedited trains** (D22), not a bypass. Promoting one strands the open train, so the workflow rebases it — the one exception to `beta/*` force-push protection (D24).
[...]

PR #117, opened 2026-08-07
ci(housekeeping): stop republishing releases, sweep -dev tags, sync four descriptions

### LINKED ISSUE #112: ci(housekeeping): stop republishing releases, sweep -dev tags, sync four descriptions

qcentic-dev:
Closes #112. Part of #107.

**Base is `ci/release-trains-and-digest-promotion`, not `main`.**

Three changes in `housekeeping.yml`, plus the sweep script they need.

## 1. The weekly rebuild stops publishing (D42)

`core-rebuild` resolved the newest non-prerelease release, rebuilt the Core image on whatever the base was that Monday, and pushed over `<version>` **and `latest`**. Under this ADR that overwrites a promoted digest with bytes no beta contained and no human approved — while the `revision` label still names the promoted commit, so D16's assertion keeps passing against changed bytes.

`release-detector` replaces it. Same two questions, answered by an issue:

- the released tag's own `FROM` digests against what upstream serves today
- Trivy over the **published** image, fixable CRITICAL/HIGH only

Both images. Core-only was an accident of how it grew — the distroless argument ("a Panel rebuild collects nothing") is an argument about *rebuilding*, not about noticing that a Chainguard base moved.

One open issue per released image, titled `Released image needs a patch: <image> <version>`, deduped on exact title: base drift and three CVEs in the same release are cleared by one patch, so filing them separately would be three issues one PR closes. The version in the title means the next release gets its own.

Unfixable findings are never reported (ADR 0016 D11) — `scan-core-image.mjs` already draws that line and is tested where it is drawn. A scanner that could not run is still a red job, separated from "found something" by whether Trivy wrote a report at all.

`stage: <version>` and its comment are deleted rather than left to mislead: they existed so per-arch tags overwrote rather than accumulated, which meant something only while this was "the release's `core` job run again".

## 2. The `-dev` tag sweep (D33)

`scripts/sweep-dev-tags.mjs` plus a pure, unit-tested `scripts/lib/dev-tag-sweep.mjs`. Three tag classes, each an anchored pattern rather than a prefix, so a release-shaped tag that somehow reaches a `-dev` repository is *skipped and reported* rather than matched:

| Pattern | Deleted when |
| --- | --- |
| `pr-<number><YYYYMM>` | the PR closed, the month rolled, or 30 days |
| `pr-<number>-<arch>` | the PR closed, or 30 days |
| `sha-<short>` | 30 days |

**The allowlist is the only guard, and the ADR is amended to say so.** D38 specified two — a token scoped to the `-dev` repositories plus a hard-coded list — and the first is not available: Docker Hub PATs carry an account-wide permission level, not a repository list, and per-repository scoping needs an OAT on a paid plan, which ADR 0016 D31 already rules out because an OAT cannot authenticate against the description API. So `DOCKERHUB_CLEANUP_TOKEN` can delete from `actana/panel` and `actana/core` too.

Accordingly: exact string match, no globs; re-asserted immediately before **every** delete call, not once at startup; refuses to run on an empty or unset list; and unit-tested with a release repository handed to it by name. Both deletions **and** skips are logged — a sweep that printed only its deletions reads identically to one whose tag listing was truncated.

Verified against the live registry:

```
$ node scripts/sweep-dev-tags.mjs --namespace actana --github-repo actana/control --dry-run
[dev-tag-sweep] 1 open pull request(s) on actana/control
panel-dev: 2 tag(s) considered, 0 would delete, 2 kept
  = pr-116202608  (PR #116 is open, 0d old)
  = pr-116-amd64  (build scaffolding for open PR #116, 0d old)
core-dev: 2 tag(s) considered, 0 would delete, 2 kept
  ...

$ node scripts/sweep-dev-tags.mjs ... --repositories panel --dry-run
[dev-tag-sweep] refusing to sweep 'panel': it is not one of panel-dev, core-dev.
```

`workflow_dispatch` carries a `dry-run` input so the same report is available from the Actions tab before trusting the credential.

## 3. Description syncing moves here (D43)

Out of `release.yml`, onto the weekly tick, covering all four repositories. ADR 0016 D33's gating rationale — a page must not describe a version nobody can pull — is **removed rather than restated**: the `-dev` pages have nothing to do with a release, and four pages drifting until someone happens to cut a release is the larger problem. A typo is now fixed by merging it and waiting for Monday, or `gh workflow run housekeeping.yml -f chore=descriptions`.

## Two things worth a reviewer's attention

**`docs/images/panel-dev.md` / `core-dev.md` are created here**, not by #113. The issue attributes them to the docs ticket, but #113 is blocked by this one, and a `descriptions` job that syncs a file which does not exist is a red weekly run from the day it merges. They are written to D36's brief and #113 is free to expand them.

**The ADR's clause numbering for the sweep does not resolve.** D10 says "the previous month's is swept (D33)", D11 cites D34 for keeping the credential out of the release repositories, and D36 and the Consequences cite D38 for the blast radius — but D33, D34 and D38 as committed are about exempt image jobs, fork builds, and the one-check-name rule. There is no numbered sweep clause. I amended what actually makes the false claim (the D38 credential posture, and the Consequences bullet) and left the numbering alone rather than renumber 44 clauses that every ticket cites. Worth fixing deliberately, probably in #113.

## Acceptance criteria

- [x] `core-rebuild` no longer pushes any tag — the job is gone; `workflows.test.mjs` asserts `housekeeping.yml` contains no `container-image.yml` call, no `push: true`, and no `docker push`
- [x] Base-digest drift and new fixable CRITICAL/HIGH open an issue, for both panel and core
- [x] The detector does not open a duplicate issue every week for the same finding
- [x] Sweep deletes `pr-*`, `pr-*-<arch>` and stale `sha-*` from the two `-dev` repositories only
- [x] Sweep uses `DOCKERHUB_CLEANUP_TOKEN` and refuses to act outside a hard-coded allowlist
- [x] Sweep logs what it deleted **and** what it skipped
- [x] Description sync moved here, all four repositories, weekly tick
- [x] `release.yml`'s `descriptions` job removed, and D33's gating rationale removed with it
- [x] `workflows.test.mjs`'s cron assertions still pass — extended to the three new weekly jobs

`container-image.yml` is untouched (#109's file).

## One commit that is not this ticket's

`build(deps): raise the js-yaml override to 4.3.1` — `pnpm audit --prod` went red on **every** open branch this morning (including #109's, which shares no files with this one) on CVE-2026-59870, published since yesterday's last green run. The repo already carries a `js-yaml@>=4.0.0 <4.3.0` override for the previous advisory in the same package; this moves the floor to 4.3.1, bounded `<5` so it stays inside the major `xmlbuilder2` asks for. Lockfile diff is three lines. Called out separately so it is reviewed as what it is. **#109 will need the same fix or a rebase.**


docs/adr/0023-release-trains-and-digest-promotion.md
```diff
@@ -124,6 +124,8 @@ Both sites gain an explicit "is this the highest released version" test, `--late
...
 
+**D38 (amended by #112) — the delete-capable credential has one guard in force, not two.** D36 and the Consequences below cite D38 for the claim that `DOCKERHUB_CLEANUP_TOKEN` is restricted to the `-dev` repositories by *both* token scope and a hard-coded allowlist in the sweep. **The token-scope half is not available on Docker Hub's personal-token model:** a PAT carries an account-wide permission level rather than a repository list, and per-repository scoping needs an Organization Access Token on a Team or Business plan — which D31 of ADR 0016 already rules out, because an OAT cannot authenticate against the description API and the account therefore needs a PAT regardless. **The posture chosen is the single guard**, not a paid plan: the hard-coded allowlist in `scripts/lib/dev-tag-sweep.mjs` is exact-match with no globs, is re-asserted immediately before every delete call rather than once at startup, refuses to run when empty or unset, and is unit-tested against a release repository handed to it by name. Revisit if the org moves to a paid plan for other reasons; do not revisit by widening the list.
```
qcentic-adm, line 127: The amendment is right and the reasoning is right — the token-scope half genuinely is not available on Docker Hub's personal-token model, and saying so is much better than quietly shipping one guard where the ADR promised two.

But this lands a *second* clause numbered D38, directly above the existing D38 at line 129. The PR description already flags that this ADR's clause numbering does not resolve for the sweep; this adds a real collision to a file that previously only had dangling citations. Someone grepping `D38` now gets two unrelated decisions, and the Consequences bullet's "(D38 as amended by #112)" does not disambiguate them.

Suggest either `**D38a — ...**`, or folding the amendment into the existing D38 as a trailing *Amended by #112:* paragraph. Renumbering the other 44 clauses is correctly out of scope; this one is just picking a label that is not already taken.

docs/ci-cd.md
```diff
@@ -471,54 +476,80 @@ Two registries are therefore in play, doing two different jobs:
...
+| `release-detector` | Mondays, 07:00 UTC | base drift or a new fixable CVE in a *released* image, both images — **opens an issue**, publishes nothing |
+| `dev-tag-sweep` | Mondays, 07:00 UTC | deletes stale `pr-*` and `sha-*` tags from `panel-dev` and `core-dev` |
+| `descriptions` | Mondays, 07:00 UTC | syncs all four Docker Hub pages from `docs/images/` |
 | `dev-audit` | Mondays, 07:00 UTC | `pnpm audit --audit-level high` over the dev tree — **opens an issue** |
 | `harness-canary` | Mondays, 07:00 UTC | the four vendors' real installers — **opens an issue** |
 
-A sixth job, `release-ref`, resolves the newest published release for
-`core-rebuild`; it is a job rather than a step only because a `uses:` job cannot
-compute its own inputs.
+An eighth job, `release-ref`, resolves the newest published release for
```
qcentic-adm, line 491: The housekeeping section rewrite is thorough and reads well. Four statements elsewhere in this same file are now false *because of this PR*, and they were not swept up:

- **L44** — "called by the PR, edge, release, and **weekly-rebuild** paths". There is no weekly-rebuild path any more; `workflows.test.mjs` asserts as much in this very PR.
- **L447** — "every push to `main`, every release, and **every weekly rebuild**", on the distroless base's availability dependency.
- **L301** — "paying them down belongs to the **weekly rebuild**, not to a PR".
- **L321** — the table row `everything apt installs | core.Dockerfile's one RUN | **the weekly rebuild**, via D5's in-layer apt-get upgrade`.

The last two are the ones I would not leave: they tell a reader that apt findings are cleared automatically every Monday, and after this PR nothing clears them until someone ships a patch release. That is D42's accepted trade, deliberately taken — but the table still states the old posture as fact, which is a worse failure mode than a stale tag name.

The `:edge` rows (L11, L18, L20, L134, L668) I would leave to #113 — those follow #116's `ci.yml` change, not this one.

qcentic-adm (approved):
Reviewed against #112, diffed against `ci/release-trains-and-digest-promotion`. Base is the integration branch, not `main` — correct. All checks green.

**All nine acceptance criteria are met, and all three traps are respected.**

| #112 asks | Where |
| --- | --- |
| `core-rebuild` no longer pushes any tag | job deleted outright; `workflows.test.mjs` pins no `container-image.yml` call, no `push: true`, no `docker push` in the file |
| base drift + new fixable CRITICAL/HIGH open an issue, both images | `release-detector`, `matrix: image: [panel, core]`, `fail-fast: false` |
| no duplicate issue every week | title carries the version, exact-title match on open issues |
| sweeps `pr-*`, `pr-*-<arch>`, stale `sha-*`, two `-dev` repos only | `dev-tag-sweep.mjs`, three anchored patterns |
| cleanup token + hard-coded allowlist | `assertSweepable`, re-asserted at the delete call itself |
| logs deletions **and** skips | `planSweep` returns both halves; `formatPlan` prints both |
| descriptions here, four repos, weekly tick | `descriptions` job |
| `release.yml`'s job removed, D33's rationale with it | gone; a test asserts a `needs:` reappearing would be the rationale coming back |
| cron assertions still pass | extended to all three new weekly jobs |

Traps: `stage: <version>` and its comment are deleted rather than left to mislead; `--exclude-pre-releases` is kept with the D9 invariant restated; unfixable findings stay out of the issue, drawn where `scan-core-image.mjs` already draws it.

### What is good beyond the checklist

The scanner step separates "found something" from "could not run" on whether Trivy wrote a report at all, and only the first files an issue — that is the distinction that decides whether a detector is trusted a year from now. `decideTag` *requires* the open-PR set rather than defaulting it, so "we could not list the pull requests" cannot read as "every pull request is closed" and sweep every review in flight. `openPullRequestNumbers` fails hard for the same reason, and `listTags` paginates to exhaustion instead of capping. 26 unit tests covering every branch including a release repository handed in by name, and a live `--dry-run` against the real registry in the description.

Both deliberate scope additions are correctly judged. `panel-dev.md` / `core-dev.md` genuinely cannot wait for #113 — a `descriptions` job syncing a file that does not exist is a red weekly run from the day it merges. The `js-yaml` override is a one-line floor bump inside the major `xmlbuilder2` asks for, three lines of lockfile, called out separately. **It is also blocking #116 right now**, which is red on exactly that advisory — worth landing this first for that reason alone.

### Two things to fix, on this branch or in #113

Both are documentation, neither blocks the merge; comments inline.

Also, for whoever merges second: this and #116 both touch `scripts/__tests__/panel-image.test.mjs` and `workflows.test.mjs`, with hunks a few lines apart in the latter.

qcentic-adm (approved):
## Review against #112 — approve

Base was `ci/release-trains-and-digest-promotion`, not `main` ✅. Diffed against that branch. All checks green on the head run ([31178271924](https://github.com/actana/control/actions/runs/31178271924)). Reviewed after the fact — this merged as `176f26c` before the review landed, so nothing below is a gate; it is what I would have said, and the two items worth acting on can go to #113 or a follow-up.

The load-bearing change is right. `core-rebuild` pushing over `:<version>` and `:latest` every Monday would have falsified the immutability claim within a week while D16's assertion kept passing against changed bytes, and this removes it at the root rather than defending it.

### Done-when, one by one

| # | Criterion | |
|---|---|---|
| 1 | `core-rebuild` no longer pushes any tag | ✅ job replaced by `release-detector`, `housekeeping.yml:349`; `workflows.test.mjs` "builds and publishes nothing from a cron" pins no `container-image.yml` call, no `push: true`, no `docker push` |
| 2 | Base drift and new fixable CRITICAL/HIGH open an issue, both images | ✅ `housekeeping.yml:349-411`, `matrix: image: [panel, core]` |
| 3 | No duplicate issue every week | ✅ `housekeeping.yml:504-509` — exact-title match, version in the title |
| 4 | Sweep deletes `pr-*`, `pr-*-<arch>`, stale `sha-*`, `-dev` only | ✅ `scripts/lib/dev-tag-sweep.mjs` — three anchored patterns, unrecognised names skipped and reported |
| 5 | `DOCKERHUB_CLEANUP_TOKEN`, refuses outside a hard-coded allowlist | ✅ `assertSweepable` throws rather than returns, re-asserted at the delete call, refuses an empty list, unit-tested with `panel` handed to it by name |
| 6 | Logs deleted **and** skipped | ✅ `formatPlan` returns both halves |
| 7 | Description sync here, four repositories, weekly tick | ✅ `housekeeping.yml:665` |
| 8 | `release.yml`'s `descriptions` removed, D33's gating rationale with it | ✅ — see (b) |
| 9 | `workflows.test.mjs` cron assertions still pass | ✅ extended to the three new weekly jobs |

### Two things worth acting on

**(a) `release-detector` runs the released tag's copy of the scripts, not the current one.** `.github/workflows/housekeeping.yml:369-371` checks out `${{ needs.release-ref.outputs.ref }}`, and `housekeeping.yml:386` and `housekeeping.yml:449` then invoke `scripts/check-base-pins.mjs --json` and `scripts/scan-core-image.mjs --image/--json` out of that checkout. The reason given — the pins that built the image an operator runs are the ones in *that* commit's Dockerfile — is right for `deploy/*.Dockerfile` and wrong for the tooling. The `--json` flag on `check-base-pins.mjs` has to exist at every tag the detector will ever be pointed at, and once a release predates a flag change, the weekly run goes red on argument parsing rather than on a finding. Pinning the Dockerfiles to the tag while running the scripts from the default branch would separate those. Nothing to test against yet (`gh release list --exclude-pre-releases` is empty pre-0.1.0), which is exactly why it will surface at the worst moment.

**(b) A drift finding is silently dropped when the scanner errors.** `.github/workflows/housekeeping.yml:494-495` — `Open an issue` has no `always()`, so a `Trivy` step that exits 1 at `housekeeping.yml:465` or `housekeeping.yml:471` skips it, and base drift already resolved into `steps.bases.outputs.drifted` never gets filed. The job is red, so it is not invisible; but "the scanner broke" and "the base moved" are independent facts and the second one earned its issue. The step's own condition is already written to handle either fact alone.

### Deviations from the ticket, both documented, both accepted

**The two guards are one.** #112 asked for a `-dev`-scoped token *and* an allowlist. Docker Hub PATs carry an account-wide permission level, per-repository scoping needs an OAT on a paid plan, and ADR 0016 D31 already rules that out because an OAT cannot authenticate against the description API. Amending the ADR at `docs/adr/0023-release-trains-and-digest-promotion.md:127` and the Consequences bullet, rather than restating a guarantee that does not hold, is the correct handling — a claim of defence in depth that is actually a single point is worse than an honest single point. The allowlist is then shaped like the load-bearing thing it is: exact equality, no globs, `throw` rather than a boolean a caller can forget, re-asserted at the delete itself, refuses an empty list, and `REPO_SETUP.md` says to treat it like a production database credential. The dry-run transcript against the live registry is the right evidence to have brought.

**D33's rationale removed rather than moved.** The ticket said "moved with it"; `release.yml:1-24` and `housekeeping.yml:659-664` remove it instead, on the grounds that "a page must not describe a version nobody can pull" does not survive contact with two `-dev` pages that describe no version at all. Agreed — carrying it across would have been carrying a reason that had stopped being true.

**`docs/images/panel-dev.md` / `core-dev.md` created here, not in #113.** Correct call. A `descriptions` job that syncs a missing file is a red weekly run from the day it merges, and `sync()` fails on `[[ ! -f "$file" ]]` by design.

**The js-yaml commit.** Out of scope, and called out as such in its own commit and in the description. `>=4.3.1 <5` rather than an unbounded floor, because unbounded resolves to a major across a dependency asking for `^4`. Right bound, right explanation. Worth noting for #109, which is red on exactly this and now has the fix sitting on its base branch.

### Smaller notes

- `scripts/lib/dev-tag-sweep.mjs:165-177`: a `pr-` image tag is deleted the moment the month rolls, even for an open PR. The justification — "the next push publishes `pr-<number><thisMonth>` and nothing points at the old name again" — assumes a push. An open PR idle across a month boundary loses its only reviewable image on the Monday, and gets it back only on the next push. This is what D10 and the ticket asked for, so it is not a defect; it is a corner the comment's reasoning does not actually cover.
- `scripts/sweep-dev-tags.mjs`: `--dry-run` needing no credential at all, so the rules can be changed from a laptop, is the detail that makes this safe to maintain rather than merely safe to run. Same for hard-failing when the pull request listing cannot be fetched — treating "we could not look" as "everything is closed" is how a sweep eats every open review at once, and that is spelled out at the function it protects.
- `release-ref` collapsing to a `::notice` and `exit 0` pre-0.1.0 keeps "no release yet" out of the red-run channel, which matters for a detector nobody is watching on a Monday.

Approving. (a) and (b) are follow-up material, not a reason to have held it.

Issue #111, opened 2026-08-07
ci(promote): the fifth entry point — verify, fast-forward, retag, re-cut


qcentic-adm:
> **Base branch: `ci/release-trains-and-digest-promotion`, not `main`.** Branch from it, PR back into it. `main` receives one PR at the end, for the whole effort. ADR 0023 lives on that branch, so it is readable from every worktree cut from it. See #107 § *Working on this effort*.

Part of #107

## Description

The new entry point — the fifth, amending ADR 0016 D34. It is the only thing that advances `main`, and it is where the design's central guarantee is enforced.

Locked in **[ADR 0023](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)** D5, D14, D15, D16, D22, D24, D25, D39, D40.

Runs on `workflow_dispatch` with the train name (D14). Auto-running on review approval was considered and rejected — promotion publishes to the world, and "approved by accident" is a failure mode worth designing out.

### The sequence, in order

1. **Human pause first** (D15) — a first job declaring `environment: macos-release`. Everything below is downstream, so `main` never contains unapproved code. The reviewer builds their macOS tarball from the **train tip**, not from a tag that does not exist yet.
2. **Verify the digest** (D16) — resolve `beta-x.y.z` to a digest and assert its `org.opencontainers.image.revision` label equals the promotion PR's head SHA. Mismatch ⇒ fail with *"the train moved; re-approve"*. **This clause is what the whole design reduces to.**
3. **Fast-forward `main`** to the train tip (D5). Not a squash, not a merge commit.
4. **Push `vx.y.z`** as a record. It triggers nothing (D40).
5. **`workflow_call` into `release.yml`**, which retags the digest and publishes.
6. **Cut `release/x.y`** if it does not exist (D27 — named for the *line*, not the patch).
7. **Delete the train** (D4).
8. **Auto-cut the next train** — `beta/<next-minor>.0` from `main`, with the four `package.json` versions written in the cut commit, **skipped when a train already exists** (D25). The guessed version is a default, not a commitment.
9. **Post-hotfix rebase** (D24) — see below.

### The post-hotfix rebase (D24)

A hotfix is an expedited train (D22), so promoting one necessarily happens while another train exists — and that train can no longer fast-forward, because `main` has moved past its ancestor. "This is a hotfix" is therefore an **observable state**, not a label: *a train was promoted while another train existed* (D23). Nothing for a human to set during an incident.

[...]

PR #130, opened 2026-08-10
ci(release): two modes, an explicit latest guard, and no tag trigger

### LINKED ISSUE #110: ci(release): two modes, an explicit latest guard, and no tag trigger

qcentic-dev:
Closes #110. Part of #107. Base is `ci/release-trains-and-digest-promotion`, not `main`.

`release.yml` stops building images and starts promoting digests — except on the one path where no digest exists to promote — and loses both its tag trigger and its human pause.

## What changed

**The trigger (D40).** `push: tags: v*` is gone; `workflow_call` replaces it and `workflow_dispatch` stays. Keeping the tag trigger beside the call would fire two release runs per promotion that could not even block each other, because `github.ref_name` is the tag under a push and the *caller's* ref under a `workflow_call`. So the concurrency group is re-derived from the version (`release-${{ inputs.tag }}` — the one string both invocations agree on) and **nothing in the file reads `github.ref_name` any more**, which a test pins. A stray hand-pushed `v*` tag now does nothing at all.

**Two modes (D17, D26).** `resolve` reads the mode off the branch graph — not an input, not a label — and announces it:

| | condition | image jobs |
| --- | --- | --- |
| **promote** | tag reachable from `main` | `docker buildx imagetools create` re-points `beta-x.y.z` at `<version>`; **builds nothing** |
| **backport** | tag on a `release/*` branch | builds and pushes from the release line, CVE gate included |

A tag in neither place fails with both places named.

**The revision assertion (D16).** Promote mode pulls `beta-x.y.z` per architecture and refuses to retag unless `org.opencontainers.image.revision` equals the promoted commit. It is the assertion `verify` already made, extended rather than copied — two implementations would be two chances to weaken the check the design reduces to. The retag lives in `publish`, which `needs: build`, so nothing is re-pointed before every arch has been checked.

**The `latest` guard (D28), on both surfaces.** `scripts/lib/release-latest.mjs` takes one decision and renders it twice — the docker tag list and an explicit `--latest=true|false` on `gh release create`, never the API's default of `true`. Three reasons withhold it, in order: backport mode *unconditionally and before anything is compared*, a prerelease (D30), then losing an explicit highest-version test — the test the old `tags="$version latest"` did not have at all. `--prerelease` is now explicit too, and `gh release edit` re-asserts both on the re-run path.

**Backport mode cannot emit `latest`.** Structurally, in the module: it never reaches the comparison that could say yes. `scripts/__tests__/release-latest.test.mjs` proves it against a ladder the backport would *win*, and sweeps every version against every ladder. Both workflow surfaces re-assert it at the point of publishing.

**The macOS pause (D15).** `environment: macos-release` is off `tarball-macos`. The pause is not dropped — it moves to the head of `promote.yml`, upstream of this whole workflow, so the fast-forward onto `main` is downstream of the human too. The image jobs still wait on the mac leg, now for a different reason: a release is atomic, and an image whose `:latest` moved beside a Release missing a third of its tarballs cannot be walked back.

## Not renamed

`Panel image` and `Core image` keep their exact names. The "Protect main" ruleset pins those check names, and a rename would block every PR in the repository until an admin updated it. The tests now assert both names literally, in `release.yml` and in `ci.yml`.

## One fix beyond the ticket

The revision label is now `git rev-parse HEAD` rather than `github.sha`. They agree on the PR and train paths (both pass `ref: github.sha`); they do **not** agree on a backport, which is dispatched from the default branch while building a `release/x.y` commit — so the image would have carried a label naming a commit it does not contain. A label that is not the truth is the one failure D16's assertion cannot detect.

## For #111

- `promote.yml` must call this with `permissions: contents: write` (or `write-all`) and `secrets: inherit` — a reusable workflow cannot request more than its caller granted, and `github-release` needs `contents: write`.
- Push `vx.y.z` **before** invoking; `resolve` requires the tag to exist and re-fetches to close the race.
- `macos-release` is now referenced by nothing until `promote.yml` picks it up. The environment stays configured (`docs/REPO_SETUP.md` §2) — a missing one is silently unprotected, not red.

## Verification

- `vitest run` — 408 passing. One pre-existing, environment-dependent failure in `core-launcher.test.mjs` (`/opt/actana/app` exists on this machine), which fails identically on other branches and is untouched here.
- `eslint scripts` clean.
- Every `run:` block in the three workflows parsed as YAML and syntax-checked with `bash -n`.
- `resolve`'s shell was extracted and executed against a fixture repository with `main`, a `release/1.0` line and a stray branch: `v1.4.0` → promote / `beta-1.4.0` / `tags=1.4.0 latest`; `v1.0.1` → backport / `tags=1.0.1` / `latest=false` *despite* `v1.4.0` existing; `v1.0.1-rc.1` → backport / `prerelease=true`; a missing tag, a malformed tag and a tag on neither `main` nor a release line each fail with their own message.
- The backport guard step and `container-image.yml`'s input validation were run directly across their branches — both surfaces of the guard fire independently.
- `commitlint --from <base> --to HEAD` — 0 problems across all four commits.

## Docs

Three pages claimed the mac leg declares the environment and that a `v*` tag fires a release. Corrected in place; the full rewrite around the train model, the tag ladder and the rollback runbook is #113.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -495,8 +495,10 @@ jobs:
   # The release-tarball seam (core-installer issue 01): the artifact
   # operators actually download must extract and boot on its own — bundled Node,
   # bundled natives, no system Node, no network. Built here on every PR for the
-  # runner's own platform so a broken tarball is caught before tag time;
-  # release.yml builds and smokes both released targets on a tag push.
+  # runner's own platform so a broken tarball is caught before release time;
+  # release.yml builds and smokes both released targets when a release runs —
+  # called from the promotion or dispatched by hand, never on a tag push
+  # (ADR 0023 D40).
   #
   # This is now the only boot-and-dial smoke on a PR. There was a second —
   # `smoke-standalone-core.mjs`, against the unpacked bundle — but it made the
@@ -643,8 +645,10 @@ jobs:
   # release target again (ADR 0016 D28, as amended) does not change that. D35
   # took the macOS legs off the PR gate because they were 72% of the bill at
   # 10× billing, and that is the decision this file keeps: zero macOS minutes
-  # on a pull request. The one mac leg lives in release.yml, runs on a tag, and
-  # waits on a human approval; `e2e-actana-setup-macos.mjs` stays deleted, and
+  # on a pull request. The one mac leg lives in release.yml and now builds
+  # alongside the Linux legs — the human approval moved out of it to the head of
+  # `promote.yml` (ADR 0023 D15, landing in #111), which is upstream of the
+  # whole release; `e2e-actana-setup-macos.mjs` stays deleted, and
   # docs/core-macos-prerelease-checklist.md is what a person runs instead.
   # `scripts/__tests__/workflows.test.mjs` fails if a macOS runner appears in
   # this file.
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -12,14 +12,23 @@ name: Container image
 #   ci.yml      → pr-<prid><YYYYMM> to `-dev`, amd64     (every PR push)
 #   ci.yml      → beta-x.y.z to the release repositories
 #                 and sha-<short> to `-dev`, multi-arch  (every train merge)
-#   release.yml → :<version> and :latest                 (every v* tag)
+#   release.yml → :<version> and :latest                 (every release)
 #
-# `mode` (ADR 0023 D38) covers the two PR paths that build nothing: `verify`
-# asserts a published digest carries a given commit's revision label (D19), and
-# `pass` reports success immediately (draft or documentation-only, D33). Both
-# run the same jobs under the same names as a build — a required check whose
-# job is skipped stays Pending forever, which is the failure D33 exists to
-# avoid.
+# `mode` (ADR 0023 D38) covers the paths that build nothing. `verify` asserts a
+# published digest carries a given commit's revision label (D19) and `pass`
+# reports success immediately (draft or documentation-only, D33); both run the
+# same jobs under the same names as a build, because a required check whose job
+# is skipped stays Pending forever, which is the failure D33 exists to avoid.
+#
+# `promote` (D16, D17) is the third, and the only one of them that publishes.
+# It makes the same assertion `verify` does — that `source_tag`'s digest was
+# built from `ref` — and then re-points `tags` at that exact digest with
+# `docker buildx imagetools create`, the command the manifest stitch below
+# already uses. Nothing is built, so nothing can differ between the bytes a
+# human approved as `beta-x.y.z` and the bytes published as `x.y.z`. The
+# assertion is not decoration: `beta-x.y.z` moves on every train merge, so
+# "retag whatever it points at" would promote untested bytes whenever somebody
+# merged after the approver tested.
 #
 # Two repositories, one build. `tags` publishes to `<image>` and `dev_tags` to
 # `<image>-dev` from the same per-architecture bytes, so `sha-<short>` is a
@@ -76,11 +85,22 @@ on:
         description: >-
           What this call does, under one check name (ADR 0023 D38):
           `build` builds, smokes and — when push is true — publishes;
+          `promote` builds nothing, asserts `source_tag`'s published digest
+          carries `ref` as its org.opencontainers.image.revision label, and
+          re-points `tags` at that digest (D16, D17);
           `verify` builds nothing and asserts `tags`'s published digest carries
           `ref` as its org.opencontainers.image.revision label (D19);
           `pass` reports success immediately (D33).
         type: string
         default: build
+      source_tag:
+        description: >-
+          The published tag whose digest `promote` re-points `tags` at (e.g.
+          beta-0.1.0), in the same repository. Required by `promote` and
+          rejected by every other mode — a build path carrying one is a caller
+          that thinks it is promoting.
+        type: string
+        default: ""
       push:
         description: "Publish the image. False builds and smokes only."
         type: boolean
@@ -122,6 +142,7 @@ jobs:
           PUSH: ${{ inputs.push }}
           TAGS: ${{ inputs.tags }}
           DEV_TAGS: ${{ inputs.dev_tags }}
+          SOURCE_TAG: ${{ inputs.source_tag }}
           # A repository variable, so a fork publishing under a different
           # Docker Hub org doesn't have to edit this file.
           DOCKERHUB_NAMESPACE: ${{ vars.DOCKERHUB_NAMESPACE }}
@@ -134,14 +155,37 @@ jobs:
             *) echo "unknown image '$IMAGE' — expected 'panel' or 'core'" >&2; exit 1 ;;
           esac
           case "$MODE" in
-            build|verify|pass) ;;
-            *) echo "unknown mode '$MODE' — expected 'build', 'verify' or 'pass'" >&2; exit 1 ;;
+            build|promote|verify|pass) ;;
+            *) echo "unknown mode '$MODE' — expected 'build', 'promote', 'verify' or 'pass'" >&2; exit 1 ;;
           esac
-          # A mode that builds nothing has nothing to publish. Catching the
-          # combination here rather than letting it push means a caller cannot
-          # turn a verification into a publish by passing one input twice.
-          if [[ "$MODE" != "build" && "$PUSH" == "true" ]]; then
+          # A mode that builds nothing and promotes nothing has nothing to
+          # publish. Catching the combination here rather than letting it push
+          # means a caller cannot turn a verification into a publish by passing
+          # one input twice.
+          if [[ "$MODE" != "build" && "$MODE" != "promote" && "$PUSH" == "true" ]]; then
             echo "mode '$MODE' builds nothing and cannot push" >&2; exit 1
+          fi
+          # `promote` is the one mode that publishes without building, so its
+          # inputs are checked rather than assumed: without a source there is
+          # no digest to re-point, and a promote that did not push would report
+          # a green check having done nothing at all — which is the shape of
+          # failure a release must never take.
+          if [[ "$MODE" == "promote" ]]; then
+            if [[ -z "${SOURCE_TAG// }" ]]; then
+              echo "::error title=Nothing to promote::mode 'promote' needs source_tag — the published tag whose digest is re-pointed at '$TAGS' (ADR 0023 D17)."
+              exit 1
+            fi
+            if [[ "$PUSH" != "true" ]]; then
+              echo "::error title=A promotion that publishes nothing::mode 'promote' exists to re-point tags; push must be true." >&2
+              exit 1
+            fi
+            if [[ -n "${DEV_TAGS// }" ]]; then
+              echo "::error title=A promotion publishes to the release repository only::dev_tags '$DEV_TAGS' were requested. The -dev handles belong to the build paths (ADR 0023 D36)." >&2
+              exit 1
+            fi
+          elif [[ -n "${SOURCE_TAG// }" ]]; then
+            echo "::error title=source_tag outside promote mode::mode '$MODE' was given source_tag '$SOURCE_TAG'. Only 'promote' re-points an existing digest; a build path carrying one is a caller that thinks it is promoting." >&2
+            exit 1
           fi
 
           # Docker Hub requires a lowercase namespace.
@@ -201,7 +245,9 @@ jobs:
             echo "::error title=Missing Docker Hub credential::DOCKERHUB_USERNAME and DOCKERHUB_TOKEN must both be set to publish $image — Docker Hub is the only registry images are published to (ADR 0018). One personal access token (Read & Write) serves both the image push and the description sync. See docs/REPO_SETUP.md section 2."
             exit 1
           fi
-          if [[ "$PUSH" == "true" ]]; then
+          if [[ "$MODE" == "promote" ]]; then
+            echo "Promoting $image:$SOURCE_TAG to $TAGS. Nothing is built, so there is no per-arch scaffolding."
+          elif [[ "$PUSH" == "true" ]]; then
             if [[ -n "${TAGS// }" ]]; then echo "Publishing $TAGS to $image."; fi
             if [[ -n "${DEV_TAGS// }" ]]; then echo "Publishing $DEV_TAGS to $dev_image."; fi
             echo "Per-arch scaffolding goes to $scaffold_image."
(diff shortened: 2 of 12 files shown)


qcentic-adm (approved):
Reviewed against #110 as the spec, diffed against `ci/release-trains-and-digest-promotion` (merge-base `7565e28`). I checked the two `latest` surfaces by reading the enforcing lines rather than the description, and confirmed the pinned check names against the live ruleset. **All nine acceptance criteria are met.** Approving; three non-blocking notes at the end.

---

## The two `latest` surfaces — a backport of 1.2.4 after 1.4.0

I traced this one on both surfaces rather than trusting the summary. It holds on both.

**Docker tag surface.** The decision cannot produce `latest` for a backport:

- `scripts/lib/release-latest.mjs:199-202` — `if (mode === "backport") { latest = false; … }` is the *first* branch, taken before the ladder is read. This is what makes it structural rather than a comparison a backport merely loses.
- `scripts/lib/release-latest.mjs:219` — `const tags = latest ? [version, "latest"] : [version];` is the only place `latest` can enter the docker tag list at all.
- `.github/workflows/release.yml:270-275` — the workflow re-checks it: `for tag in $TAGS; … if [[ "$tag" == "latest" ]]; … exit 1`. This is a real gate, not decoration — it lives in `resolve`, and `panel`/`core` both `needs: resolve` (`:547`, `:567`), so a failure here skips both image jobs before anything is pushed.

**`gh release create` surface.** The flag is never defaulted:

- `.github/workflows/release.yml:691` — `--latest="$RELEASE_LATEST" \` inside `gh release create`.
- `.github/workflows/release.yml:686` — the same on the `gh release edit` re-run path, which `gh release upload` cannot express.
- `.github/workflows/release.yml:665-668` — refuses anything that is not literally `true`/`false`, so a missing upstream output fails loudly instead of falling through to the API's `make_latest: true`.
- `.github/workflows/release.yml:669-672` — refuses `backport` + `latest != false` independently of `resolve`.

Both renderings come from one boolean (`release-latest.mjs:219` and `:221`), so they cannot drift. The neighbouring case — *promoting* an older version that is not a backport, e.g. re-dispatching `v1.0.0` after `v1.4.0` — is caught separately by the highest-version test at `release-latest.mjs:149-154`, which is the test the old `tags="$version latest"` did not have at all.

## Concurrency: `workflow_call` vs `workflow_dispatch` for one version

`.github/workflows/release.yml:135` — `group: release-${{ inputs.tag }}`, with `cancel-in-progress: false` at `:136`.

- `inputs` is available to the workflow-level `concurrency` key (GitHub's context-availability table lists `github, inputs, vars` for it), and it is populated for both `workflow_call` and `workflow_dispatch`.
- Both triggers declare `tag` as required (`:109-112`, `:122-125`), so the expression is the same non-empty string on either entry path. Concurrency groups are scoped per repository, and this is a same-repo call — the two invocations land in **one** group and serialise. That is the behaviour the trap asked for.
- It is at workflow level in the *called* workflow, which is the placement that actually works for reusable workflows; `jobs.<id>.concurrency` on the calling job would not have.
- `github.ref_name` is gone from the file entirely — grep confirms it survives only inside comments at `:95` and `:128`, and `workflows.test.mjs:204-206` pins its absence.

No collision, no divergence. Verified.

## Pinned check names

Byte-identical. `release.yml:546` `name: Panel image` and `:566` `name: Core image` differ from the base only in line number (was `:383`/`:397`); `ci.yml` is not in the diff at all.

Worth recording that the "Protect main" ruleset (id `20390421`) pins the *composite* names, not just the caller's:

```
Panel image / Resolve registries
Panel image / Build + smoke (amd64)
Core image / Resolve registries
Core image / Build + smoke (amd64)
```

So `container-image.yml`'s own job names had to hold too — and they do: `:130` `Resolve registries` and `:259` `Build + smoke (${{ matrix.arch }})`, both unchanged from the base (`:110`, `:213`). Reusing `build` for the promote path under the same name, rather than adding a job, is what kept that true.

## Trigger, promote mode, and the retag

- **Tag trigger gone, `workflow_call` present, dispatch retained** — `release.yml:103-125`. The base's `push: tags: - "v*"` is deleted.
- **Retags and builds nothing** — `container-image.yml:659-663` swaps in the single published index as the source; `:675` is the `docker buildx imagetools create` call, character-for-character the base's `:559`. It is genuinely the *same* stitch, not a parallel implementation — only the `sources` array differs. Every `docker build` is gated on `inputs.mode == 'build'` (`:395`, `:449`), and `workflows.test.mjs:296-304` asserts that by enumerating build steps rather than by name, so an ungated one added later fails the test.
- **Asserts `org.opencontainers.image.revision` before retagging** — `container-image.yml:332-333` (`if: inputs.mode == 'verify' || inputs.mode == 'promote'`), reading the label at `:365-366` and refusing at `:370-377` with a message that names the actual remedy. The ordering is structural, not incidental: the assertion is in `build`, the retag is in `publish`, and `publish` declares `needs: [resolve, build]` at `:609`. With `fail-fast: false` on the matrix, a single arch failing its assertion still fails `build` and skips `publish`, so no tag moves. `continue-on-error: ${{ !inputs.push_required }}` at `:614` stays blocking because the release path leaves `push_required` at its `true` default.
- **Backport keeps the CVE gate** — `container-image.yml:531-532`, `mode == 'build'`, which is what a backport resolves to (`release.yml:219`).

Since the release path can only be proven at a real tag, that reading-against-the-existing-stitch is the strongest available evidence, and it is convincing.

## D16's revision assertion on backports — I agree with you

The PR flags this as a fix beyond the ticket, against #110's claim that "the label is already baked in by the build; nothing new has to be published to make this work." That claim is true for promote mode and false for the path this PR adds, and the reasoning holds:

- Base `container-image.yml:318`/`:370` passed `SHA: ${{ github.sha }}` and labelled with it at `:379`. Head uses `revision="$(git rev-parse HEAD)"` at `:403` and `:458`.
- In a reusable workflow the `github` context is the **caller's**, so `github.sha` is the event's commit, not `inputs.ref`. The two coincide on the PR and train paths only because `ci.yml:1023`/`:1039` pass `ref: ${{ github.sha }}` explicitly.
- A backport is a `workflow_dispatch` on the default branch building a `release/x.y` commit (`release.yml:558` passes `resolve.outputs.sha`), so `github.sha` would have labelled the image with a commit it does not contain.

And it is in scope rather than scope creep, for the reason given: D16 *is* a comparison against that label, so a label that lies is the one failure the comparison cannot detect. `git rev-parse HEAD` after a checkout at `inputs.ref` is the commit whose bytes are in the image on every path, including the ones where the event agrees. `workflows.test.mjs:338-344` pins it.

## Verification I ran myself

- `vitest run` — 409 tests, 408 pass. The single failure is `core-launcher.test.mjs` asserting against `/opt/actana/app`, which exists on my machine too; that file is not in the diff. Matches what the PR reports.
- `eslint scripts` — clean.
- The three new/changed test files — 144 tests, all passing.

---

## Non-blocking, for a follow-up

**1. `docs/README.md:82-85` was missed.** The docs commit says three pages were corrected; this is a fourth, and both halves of it are now false:

> **This one is the release gate**: a `v*` tag pauses on the `macos-release` environment until a reviewer works through it on real Apple hardware, and nothing — images, `:latest`, Release or tarballs — publishes until they approve

A `v*` tag now does nothing at all (D40) and `tarball-macos` no longer declares the environment (D15). It is exactly the sentence the rest of this change set out to stop being wrong, and it sits on the index page a reader hits first. The three pages you did fix handle the interim state well — being explicit that the environment is referenced by nothing until #111 lands, rather than pretending the pause is already in its new home, is the right call.

**2. Two stale comments in `ci.yml`** — `:499` ("release.yml builds and smokes both released targets on a tag push") and `:646` ("The one mac leg lives in release.yml, runs on a tag, and waits on a human approval"). The second is now false twice over. Comments only, in a file this PR deliberately left alone, so leaving them to #113 is defensible — just noting them so they are not lost.

**3. Carry to #111, not a request here.** Ruleset `20390424` ("Release tags — only admins cut releases") restricts `creation` on `refs/tags/v*` and lists only the admin repository role as a bypass actor. `promote.yml` pushing `vx.y.z` with `GITHUB_TOKEN` before invoking this workflow will be refused unless that ruleset gains a bypass actor. Your `## For #111` section covers the permissions and ordering; this one is not in it yet, and `resolve` failing with "No such tag" (`release.yml:198-201`) is where it would surface.

Nothing above blocks. The safety-critical surfaces — both `latest` renderings, the concurrency key, the revision assertion's position upstream of the retag, and the four pinned check names — are correct, and each one is pinned by a test rather than left to the next editor's care.

qcentic-dev:
Thanks for the review — findings **1 and 2 are fixed** in c4ac318; **3 is deferred to #111**.

**1. `docs/README.md:82-85`.** You were right that it was a fourth page, and that both halves were false. Corrected in place, in the same interim register as the three pages the docs commit already fixed: the gate itself is stated without a trigger ("a release pauses on `macos-release`"), followed by an explicit note that *where* it sits is moving — no longer `release.yml`'s `tarball-macos` (D15), a `v*` tag now fires nothing at all (D40), and it becomes the first step of `promote.yml` in #111, which gates the fast-forward onto `main` as well. Exactly one pause exists either way, and the checklist is unchanged by the move.

**2. Two stale comments in `ci.yml`.**

- `:499` — "release.yml builds and smokes both released targets on a tag push" now reads "when a release runs — called from the promotion or dispatched by hand, never on a tag push (ADR 0023 D40)". The neighbouring "before tag time" went with it, since it named the same vanished trigger.
- `:646` — the mac leg no longer "runs on a tag, and waits on a human approval". It builds alongside the Linux legs; the approval moved out of `release.yml` to the head of `promote.yml` (D15, #111). The sentence the comment actually exists for — zero macOS minutes on a PR, `e2e-actana-setup-macos.mjs` stays deleted, the checklist is what a person runs — is untouched.

**Comments only.** `git diff` on `.github/workflows/ci.yml` contains no non-comment line, and `workflows.test.mjs` is green (34/34). The broader `ci.yml`/`ci-cd.md` rewrite stays with #113.

**3. Not addressed here, by intent.** Ruleset `20390424` restricting `creation` on `refs/tags/v*` with only the admin role as a bypass actor is a real blocker for `promote.yml` pushing `vx.y.z` with `GITHUB_TOKEN`, and it is not yet in the `## For #111` section. It is being carried forward on #111 rather than handled on this branch — nothing here can fix it, and `resolve` failing with "No such tag" (`release.yml:198-201`) is indeed where it would surface.

PR is out of draft. Base unchanged (`ci/release-trains-and-digest-promotion`).

Issue #115, opened 2026-08-07
chore(release): cut beta/0.1.0 and promote it as the first release


qcentic-adm:
> **Base branch: `ci/release-trains-and-digest-promotion`, not `main`.** Branch from it, PR back into it. `main` receives one PR at the end, for the whole effort. ADR 0023 lives on that branch, so it is readable from every worktree cut from it. See #107 § *Working on this effort*.

Part of #107

## Description

The maiden voyage. `beta/0.1.0` is the first train and its promotion is the **first release this repository has ever cut** — the tag history starts here (`git tag -l` is empty today, deliberately, per ADR 0016 D27).

The machine ships nothing until it has shipped once, and doing it with nothing at stake is the whole argument for building this before `0.1.0` rather than after.

Locked in **[ADR 0023](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)** §Sequencing, D3, D7, D25.

## Steps

1. Confirm every other ticket on #107 is closed and its checks are green on `main`
2. Cut `beta/0.1.0` — the workflow creates the branch and writes `0.1.0` into all four manifests (already the current value, so the commit may be a no-op; that is fine and expected)
3. Confirm the cut published `beta-0.1.0` to both release repositories, multi-arch. **A zero-merge train still has an image** because the cut publishes one (D7) — this is the edge case the clause exists for
4. Pull `actana/panel:beta-0.1.0` and `actana/core:beta-0.1.0` and work `docs/beta-acceptance-checklist.md` against them
5. Open the promotion PR (`beta/0.1.0` → `main`) and confirm: `Conventions` passes with a `beta/*` head, the image checks run in **verify** mode and take seconds, the enforcement job passes
6. Approve, then dispatch `promote.yml`
7. Approve the macOS pause after working `docs/core-macos-prerelease-checklist.md` against a tarball built from the train tip
8. Confirm the outcome: `main` fast-forwarded, `v0.1.0` pushed, `0.1.0` + `latest` published as **retags of the beta digest**, `release/0.1` cut, `beta/0.1.0` deleted, `beta/0.2.0` auto-cut
9. Confirm `docker pull actana/panel:0.1.0` and `:beta-0.1.0` resolve to the **same digest** — this is the entire point, and it is one command to prove
10. Confirm the GitHub Release is marked latest, carries all three tarballs and `SHA256SUMS`, and that `install.sh` installs `0.1.0` from a clean machine

## Acceptance criteria

[...]

Issue #108, opened 2026-08-07
chore(ci): provision the App identity, the -dev image repos, and the cleanup token


qcentic-adm:
Part of #107

## Description

Everything downstream needs three things that only a repository admin can create. None of them is code, and **all three must exist before the workflow tickets can be verified** — a promotion with no App identity silently does nothing, and a PR image build with no `-dev` repository fails at push.

Locked in **[ADR 0023](https://github.com/actana/control/blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)** D36, D38, D39.

### 1. A GitHub App as the push identity (D39)

`promote.yml` writes to protected branches: it fast-forwards `main`, writes the train-cut commit, and force-pushes the post-hotfix rebase. It cannot use the default `GITHUB_TOKEN`, for two reasons and **the first is a silent failure**:

- **GitHub does not trigger workflows from pushes made with `GITHUB_TOKEN`.** A promotion pushing `vx.y.z` with it produces a tag and no release, with nothing red anywhere. The fast-forward of `main` likewise never fires `landing.yml`, so a landing-page change merged through a train quietly stops deploying.
- `main` and `beta/*` require a PR before merging, so a direct push needs a **bypass actor**.

A long-lived PAT with push rights on `main` sitting in Actions secrets is the alternative and is worse.

### 2. Two Docker Hub repositories for PR images (D36)

`actana/panel-dev` and `actana/core-dev`, **public**. PR images do not go in the release repositories: the `descriptions` job exists to make those two pages presentable, and hundreds of `pr-*` tags undo it. Separate repositories also make a wrong-repository pull impossible and bound the blast radius of the delete credential below.

### 3. A delete-scoped Docker Hub token (D38)

The tag sweep runs unattended on a cron and deletes. Docker Hub has no undelete, and one typo'd glob against a token that can also delete `1.2.0` is unrecoverable. `DOCKERHUB_CLEANUP_TOKEN` is a **second** token, scoped to the two `-dev` repositories only — the sweep additionally carries a hard-coded repo allowlist, so there are two independent guards.

## Acceptance criteria

- [ ] GitHub App created in the `actana` org, installed on `actana/control`, with contents: write
- [ ] App id and private key stored as repository secrets (name them in `docs/REPO_SETUP.md` §2 — the docs ticket depends on knowing them)
- [ ] App added as a **bypass actor** on the `main` ruleset and (once it exists) the `beta/*` ruleset
[...]

PR #131, opened 2026-08-10
ci(promote): the fifth entry point — verify, fast-forward, retag, re-cut

### LINKED ISSUE #111: ci(promote): the fifth entry point — verify, fast-forward, retag, re-cut

qcentic-dev:
Closes #111. Part of #107. Base is `ci/release-trains-and-digest-promotion`, not `main`.

`.github/workflows/promote.yml` — the fifth entry point (ADR 0023, amending ADR 0016 D34). It is the only thing that advances `main`, and it is where the design's central guarantee is enforced. `scripts/__tests__/workflows.test.mjs` learns the new count.

## ⚠ Two prerequisites that must be satisfied before the first promotion

Neither is code, and neither can be arranged from this branch. **Both are silent or late failures if skipped**, which is why they are at the top.

### 1. The App must be a bypass actor on the tag ruleset — the tag push is refused outright without it

Ruleset **20390424, "Release tags — only admins cut releases"**, restricts **creation** on `refs/tags/v*` to the admin role and currently has **zero bypass actors** (`"bypass_actors"` is absent; `current_user_can_bypass: "never"`). The App token is not an admin, so the tag push is **rejected by the server** — not silently skipped.

That is the good failure mode, and this workflow is written assuming it will be fixed: the tag push is wrapped so the refusal reports as an actionable `::error` naming the ruleset, and it says what to do from there (`main` has already fast-forwarded at that point and that part is correct and keeps — add the bypass actor, push the tag by hand, and `gh workflow run release.yml -f tag=vx.y.z`).

~~The App also needs to be a bypass actor on the **`main`** ruleset (20390421), which requires a pull request before merging — same treatment, same wrapped error.~~ **Corrected:** the **`main`** ruleset (20390421) has since gained an `Integration` bypass actor with `bypass_mode: always`, so the fast-forward is no longer blocked and needs nothing further. Only the **tag** ruleset above is outstanding. Both remain #108's acceptance criteria and #114's ruleset work.

### 2. The `macos-release` environment does not exist yet, and a missing one is *not* an error

`gh api repos/actana/control/environments` returns `total_count: 0`. **GitHub auto-creates an environment it sees referenced by a job, with no protection rules on it** — so a promotion dispatched today would sail straight through the pause, unattended, and report green. The one job whose whole purpose is to hold the release behind a person would hold it behind nobody.

**It must be created with a required reviewer before the first promotion.** `docs/REPO_SETUP.md` §2 already describes it and already says to set it up regardless of this workflow; this is the ticket that makes it load-bearing.

## The sequence

```
pause ── resolve ── verify ── advance ── release ─┬─ release-line
(human)                      (ff + tag)  (call)   ├─ retire-train ─┬─ next-train
                                                  │                └─ rebase-train
```

1. **The human pause is the first job** (D15) — it declares `environment: macos-release` and every other job in the file is downstream of it, so the fast-forward is downstream of the reviewer and `main` never contains unapproved code. The dispatch input is validated *inside* the pause rather than before it: a job with an `environment:` blocks before its first step runs, so anything upstream would be a job `main` is not protected from. A typo'd train name costs one approval click and then fails loudly.
2. **Verify the digest** (D16) — `beta-x.y.z` resolved to a digest, `org.opencontainers.image.revision` asserted equal to **the promotion pull request's head SHA**, on both images and both architectures, before anything moves. The comparison is against that SHA and never against the tag's own pointer: asking the registry what `beta-x.y.z` names and then checking that `beta-x.y.z` names it is the tautology this clause exists to prevent. `release.yml` asserts it again immediately before the retag (D17); that one is not redundant with this one, because this one runs *before* `main` moves.
3. **Fast-forward `main`** to that exact commit (D5) — the SHA is pushed, not the branch, so a train that moved between approval and push simply does not reach `main`. Ancestry is re-asserted immediately before the push; a non-fast-forward is an error and never a merge, and there is no `--force` anywhere.
4. **Push `vx.y.z`** as a record (D40) — it triggers nothing, and it is pushed **before** `release.yml` is invoked because `release.yml`'s `resolve` requires the tag to exist and reads promote-vs-backport mode off where it lives.
5. **`workflow_call` into `release.yml`** with `permissions: contents: write` and `secrets: inherit` — a reusable workflow cannot exceed the grant its caller gives it, this file's top-level grant is `contents: read`, and `github-release` needs write. Without that line the release would run to its last job and fail there, after both images were published.
6. **`release/x.y`** cut if the line has none (D27), left where it is on later patches.
7. **The train is deleted** (D4), after the release rather than before, and only when its tip is reachable from `main` — a train that grew after the fast-forward would take real work with it.
8. **The next train is auto-cut** (D25) — `beta/<next-minor>.0` from `main`, with the four manifests written in the cut commit (D3) by an in-place edit rather than a `jq` reserialisation, verified afterwards, and with a Conventional Commits message because `ci.yml` lints every commit in the promotion pull request, not just its title. Skipped when a train already exists, re-checked against origin at the moment of cutting.
9. **The post-hotfix rebase** (D24) — see below.

## The post-hotfix rebase

"This is a hotfix" is read off the ref graph, never from a label or a dispatch input (D23): *another train exists at the moment this one promotes*. That train can no longer fast-forward, so it is rebased onto the new `main` and force-pushed — the one exception to `beta/*` protection — with `--force-with-lease` against the tip this job read, so a merge that landed mid-rebase loses the push rather than the commit.

Both obligations D24 attaches are here, and neither is the natural way to write a rebase:

- **It comments on every open pull request into the rebased train.** Their base was rewritten without them being told; approvals survive because they attach to the head, but conflicts can appear.
- **It republishes `beta-x.y.z` and `sha-<short>` as its last step.** Every commit SHA changed, so the tags name an image whose revision label points at a commit that is no longer reachable — and D16's assertion would fail at *that* train's promotion, long after the cause, reading as "the train moved" when nothing moved. The force-push is made with the App token, so it triggers `ci.yml` on `beta/**`; the step waits for that run, falls back to dispatching `ci.yml` if the push somehow triggered nothing, and then **checks the registry** — both tags, both images, revision label equal to the rebased tip. A green run is not the assertion; the tags naming the new tip is.

A conflicting rebase fails loudly, resolves nothing, and names the fallback: abandon the train, re-cut from `main`, cherry-pick the squash commits.

More than one surviving train is refused in `resolve`, before anything is published — the model has no answer for three trains and the rebase would have to guess which one survives.

## The identity (D39)

**Every push here is made with the GitHub App token, and none with `GITHUB_TOKEN`.** GitHub does not trigger workflows from pushes made with the default token, so a promotion using it would push `vx.y.z` and produce no release, and would fast-forward `main` without ever firing `landing.yml` — a landing-page change merged through a train would quietly stop deploying, with nothing red anywhere. `advance` therefore **requires** `APP_ID` / `APP_PRIVATE_KEY` and fails with an actionable error rather than falling back to anything.

`GITHUB_TOKEN` is used, deliberately and only, for *reading* the promotion pull request and for *commenting* on the pull requests a rebase rewrote. Neither is a push, neither needs to trigger a workflow, and keeping them off the App means this workflow needs no App permission beyond the `contents: write` #108 provisioned.

## What it deliberately does not do

- **No cleanup of the promotion pull request.** GitHub closes it as merged once its commits are reachable from `main`, which the fast-forward makes true.
- **No "must have merges" guard.** A zero-merge train is legitimate — `beta/0.1.0` is expected to be one — and it still has an image, because the cut publishes one (D7).
- **No second resolution of `beta-x.y.z`.** The verification is against the promotion pull request's head SHA, always.
- **No recovery machinery past `advance`.** Up to and including the fast-forward a failed run is re-runnable from the top. After it, `main` has moved and GitHub has closed the promotion pull request, so `resolve` refuses by design — the recovery is per-step and documented in the file header. A promotion that could half-run twice is worse than one that stops and says so.

## `workflows.test.mjs`

The entry-point assertion goes from four to five, and nothing else. **The explanatory comment about D34's count is left alone on purpose** — #113 edits the same file and rewrites it, so the change here is one array entry and the `it` title.

## Verification

- `pnpm exec vitest run scripts/__tests__/workflows.test.mjs` — 34 passed.
- Full root suite: 408 passed, 1 pre-existing failure in `scripts/__tests__/core-launcher.test.mjs`, unrelated to this branch — it fails identically on an untouched worktree of another branch on this machine, because a real `/opt/actana` install shadows the test's temporary install root.
- `promote.yml` parsed as YAML, every `run:` block syntax-checked with `bash -n`, the job graph confirmed to hang every job off `pause`.
- CI on this branch: everything green — `Conventions` (every commit, not just the title), `Train rules`, `Typecheck`, `Unit Tests`, `Lint`, `Secret Scan`, both image checks, both tarball smokes, both installer e2es — **except `Dependency Audit`**, which fails on `GHSA-2v37-7h3g-55p8` (`@tanstack/react-router` under `@tanstack/react-start`). That failure is repository-wide right now: the same job is red on `ci/release-two-modes` and on every open Dependabot pull request. Nothing on this branch touches a dependency.
- Not runnable end to end from a pull request: it is `workflow_dispatch`-only and both prerequisites above are outstanding. The first real exercise is #115, `beta/0.1.0`.

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -239,8 +239,13 @@ jobs:
           # train exists at the moment this one promotes. A hotfix is an
           # expedited train (D22), so promoting one necessarily happens while
           # another train exists — and that train can no longer fast-forward.
-          survivors="$(git ls-remote --heads origin 'refs/heads/beta/*' \
-            | sed 's#.*refs/heads/##' | grep -vx "$TRAIN" || true)"
+          # Listed first and filtered second, so that a failing `ls-remote`
+          # fails this step. Folding `|| true` over the whole pipeline — which
+          # `grep`'s exit-1-on-no-match invites — would turn "origin did not
+          # answer" into "no other train exists", and the surviving train would
+          # be stranded silently with nothing to say so (D24).
+          all_trains="$(git ls-remote --heads origin 'refs/heads/beta/*' | sed 's#.*refs/heads/##')"
+          survivors="$(grep -vx "$TRAIN" <<<"$all_trains" || true)"
           survivor_count="$(grep -c . <<<"$survivors" || true)"
           hotfix=false
           survivor=""
@@ -649,7 +654,9 @@ jobs:
           APP_SLUG: ${{ steps.app-token.outputs.app-slug }}
         run: |
           set -euo pipefail
-          existing="$(git ls-remote --heads origin 'refs/heads/beta/*' | sed 's#.*refs/heads/##' || true)"
+          # No `|| true`: an `ls-remote` that failed must not read as "no train
+          # exists", which is the answer that cuts a second one (D2, D25).
+          existing="$(git ls-remote --heads origin 'refs/heads/beta/*' | sed 's#.*refs/heads/##')"
           if [[ -n "${existing// }" ]]; then
             echo "::notice title=A train already exists::$(tr '\n' ' ' <<<"$existing")— the invariant D25 protects already holds, so nothing is cut. This is the ordinary post-hotfix case."
             exit 0
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -753,6 +753,12 @@ jobs:
     timeout-minutes: 150
     permissions:
       contents: read
+      # The republish step reads `ci.yml`'s runs to know whether the
+      # force-push triggered one, and dispatches `ci.yml` itself when it did
+      # not. Both are the Actions API, and neither is available under the
+      # default `contents: read` alone.
+      actions: write
+      # The comment on every open pull request into the rebased train (D24).
       pull-requests: write
     steps:
       - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -1,0 +1,962 @@
+name: Promote
+
+# The fifth entry point (ADR 0023, amending ADR 0016 D34). It is the only thing
+# that advances `main`, and it is where the design's central guarantee is
+# enforced.
+#
+#   pause ── resolve ── verify ── advance ── release ─┬─ release-line
+#   (human)                       (ff + tag) (call)   ├─ retire-train ─┬─ next-train
+#                                                     │                └─ rebase-train
+#                                                     └─
+#
+# The order of the first four is the design (D15, D16, D5, D40):
+#
+#   1. **The human pause is first.** A job declaring `environment:
+#      macos-release` and nothing else of consequence. Everything below is
+#      downstream of it, so `main` never contains unapproved code — where the
+#      old shape (ADR 0016's environment on `release.yml`'s macOS leg) already
+#      had the commit on `main` by the time anybody was asked.
+#   2. **Verify the digest before anything moves.** `beta-x.y.z` moves on every
+#      train merge, so promotion resolves it to a digest and asserts the
+#      image's `org.opencontainers.image.revision` label equals the promotion
+#      pull request's head SHA. The whole design reduces to this clause.
+#   3. **Fast-forward `main`** to that exact commit. Not a squash, not a merge
+#      commit; a non-fast-forward is an error, never a merge (D5).
+#   4. **Push `vx.y.z`** as a record, *then* call `release.yml` by
+#      `workflow_call` (D40). The tag triggers nothing — `release.yml` has no
+#      tag trigger — but `release.yml`'s `resolve` requires it to exist and
+#      reads promote-vs-backport mode off where it lives, so the order is not
+#      cosmetic.
+#
+# ── The identity, and the way this ships broken and looks fine (D39) ─────────
+#
+# **Every push here is made with the GitHub App token, and none with
+# `GITHUB_TOKEN`.** GitHub does not trigger workflows from pushes made with the
+# default token, so a promotion using it would push `vx.y.z` and produce no
+# release, and would fast-forward `main` without ever firing `landing.yml` — a
+# landing-page change merged through a train would quietly stop deploying. There
+# would be nothing red anywhere. This is the single most likely way to ship this
+# workflow broken and have it look green, which is why the App token is minted
+# per job that pushes and why `check-app-credentials` fails loudly rather than
+# falling back to anything.
+#
+# `GITHUB_TOKEN` is still used, deliberately and only, for *reading* the
+# promotion pull request and for *commenting* on the pull requests a rebase
+# rewrote. Neither is a push, neither needs to trigger a workflow, and keeping
+# them off the App means this workflow needs no App permission beyond the
+# `contents: write` #108 provisioned.
+#
+# ── What this workflow does not do ──────────────────────────────────────────
+#
+# * **It does not close the promotion pull request.** GitHub closes it as merged
+#   on its own once its commits are reachable from `main`, which the
+#   fast-forward makes true. A cleanup step would be a second thing to go wrong.
+# * **It does not check that the train has merges in it.** A zero-merge train is
+#   legitimate — `beta/0.1.0` is expected to be one — and it still has an image,
+#   because the cut publishes one (D7). A "must have merges" guard would refuse
+#   the maiden voyage.
+# * **It does not resolve `beta-x.y.z` a second time to decide what to promote.**
+#   The comparison is always against the promotion pull request's head SHA.
+#   Resolving the tag twice — asking the tag what it points at and then checking
+#   the tag points at it — is the bug D16 exists to prevent.
+#
+# ── Re-running after a partial failure ──────────────────────────────────────
+#
+# Up to and including `advance`, a failed run is re-runnable from the top: the
+# promotion pull request is still open and nothing is published. **After**
+# `advance`, it is not — `main` has moved, so GitHub has closed the promotion
+# pull request as merged and `resolve` will refuse (by design: the pull request
+# is the gate). From there the recovery is per-step and manual:
+# `gh workflow run release.yml -f tag=vx.y.z` re-runs the publish, and the
+# branch housekeeping below is a `git push` each. That is deliberate — a
+# promotion that could half-run twice is worse than one that stops and says so.
+on:
+  # D14. A dispatch, and nothing else. Auto-running on review approval was
+  # considered and rejected: promotion publishes to the world, and "approved by
+  # accident" is a failure mode worth designing out.
+  workflow_dispatch:
+    inputs:
+      train:
+        description: "Train branch to promote (e.g. beta/0.1.0)"
+        required: true
+        type: string
+
+concurrency:
+  # Keyed on the train, and never cancelling. Two promotions of one train would
+  # race on the same fast-forward; a second promotion cancelling the first
+  # mid-publish would leave `main` advanced with no release behind it.
+  group: promote-${{ inputs.train }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
+jobs:
+  # ─────────────────────────────────────────────────────────── 1. the human
+  #
+  # D15. The first job, and the only one that declares an environment. Its
+  # steps are trivial on purpose: what matters is that the job exists, that it
+  # is gated on `macos-release`, and that every other job in this file is
+  # downstream of it. The reviewer has already run
+  # docs/core-macos-prerelease-checklist.md on real Apple hardware against the
+  # **train tip** — the same commit this promotes, by the assertion in
+  # `verify`, and available to them earlier than a tag that does not exist yet.
+  #
+  # The input is validated here rather than before the pause, because there is
+  # no "before": a job with an `environment:` blocks before its first step
+  # runs, so anything upstream of the pause would be a job `main` is not
+  # protected from. A typo'd train name therefore costs one approval click and
+  # fails immediately after it, loudly. That is the cheaper side of the trade.
+  #
+  # **`macos-release` must exist with a required reviewer before this runs.**
+  # GitHub auto-creates an environment it sees referenced and gives it **no
+  # protection rules**, so a missing environment does not fail — it waves the
+  # promotion through unattended and green, which is the one outcome this job
+  # exists to prevent.
+  pause:
+    name: Approve the promotion
+    runs-on: ubuntu-24.04
+    environment: macos-release
+    timeout-minutes: 10
+    steps:
+      - name: Announce what a reviewer just approved
+        env:
+          TRAIN: ${{ inputs.train }}
+          ACTOR: ${{ github.actor }}
+        run: |
+          set -euo pipefail
+          if [[ ! "$TRAIN" =~ ^beta/[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "::error title=Not a train branch::'$TRAIN' is not \`beta/x.y.z\`. A train's name is its version (ADR 0023 D3), and the version is what this promotion publishes."
+            exit 1
+          fi
+          version="${TRAIN#beta/}"
+          echo "::notice title=Promotion approved::$ACTOR dispatched the promotion of $TRAIN and a reviewer approved it. Everything downstream of this job — the digest verification, the fast-forward of main, the tag, and every published image — is downstream of that approval (ADR 0023 D15)."
+          {
+            echo "### Promoting \`$TRAIN\` → \`$version\`"
+            echo
+            echo "The macOS checklist was worked against the train tip before this was dispatched."
+            echo "Nothing has been published yet: the next job asserts the beta digest was built"
+            echo "from the commit this promotes, and the fast-forward happens only after it passes."
+          } >> "$GITHUB_STEP_SUMMARY"
+
+  # ──────────────────────────────────────────────────── 2. the facts, gathered
+  #
+  # Everything the rest of the file branches on is decided here, from the ref
+  # graph and the pull request — not from inputs, and not from labels. In
+  # particular the hotfix condition (D23): *a train was promoted while another
+  # train existed* is a thing this job can see, so there is nothing for a human
+  # to set during an incident, which is when it would be forgotten.
+  resolve:
+    name: Resolve the promotion
+    needs: pause
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      train: ${{ steps.facts.outputs.train }}
+      version: ${{ steps.facts.outputs.version }}
+      head_sha: ${{ steps.facts.outputs.head_sha }}
+      pr: ${{ steps.facts.outputs.pr }}
+      release_line: ${{ steps.facts.outputs.release_line }}
+      next_train: ${{ steps.facts.outputs.next_train }}
+      hotfix: ${{ steps.facts.outputs.hotfix }}
+      survivor: ${{ steps.facts.outputs.survivor }}
+    steps:
+      # Every branch and every tag: two of the three questions below — is this
+      # a fast-forward, and does another train exist — are questions about the
+      # whole ref graph rather than about one commit.
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          fetch-depth: 0
+
+      - id: facts
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          TRAIN: ${{ inputs.train }}
+        run: |
+          set -euo pipefail
+          version="${TRAIN#beta/}"
+
+          if ! git rev-parse -q --verify "refs/remotes/origin/${TRAIN}^{commit}" >/dev/null; then
+            echo "::error title=No such train::\`$TRAIN\` is not a branch on origin. Promotion promotes an open train (ADR 0023 D1)."
+            exit 1
+          fi
+
+          # The promotion pull request is the gate (D5): its checks and its
+          # approval are the point, and a workflow performs the advance. It is
+          # also where the SHA D16 verifies against comes from, so its absence
+          # is fatal rather than a warning.
+          pr_json="$(gh pr list --repo "$REPO" --base main --head "$TRAIN" --state open \
+            --json number,headRefOid,url --limit 10)"
+          count="$(jq 'length' <<<"$pr_json")"
+          if [[ "$count" == "0" ]]; then
+            echo "::error title=No open promotion pull request::There is no open pull request from \`$TRAIN\` into \`main\`. That pull request is the gate — its checks and its approval are what promotion consumes, and its head SHA is what the beta digest is verified against (ADR 0023 D5, D16). Open it and let it go green. If this train has already promoted, re-run the publish with \`gh workflow run release.yml -f tag=v$version\` instead."
+            exit 1
+          fi
+          if [[ "$count" != "1" ]]; then
+            echo "::error title=More than one promotion pull request::$count open pull requests have \`$TRAIN\` as their head and \`main\` as their base. Promotion consumes exactly one gate; close the extras."
+            jq -r '.[] | "  " + .url' <<<"$pr_json"
+            exit 1
+          fi
+          pr="$(jq -r '.[0].number' <<<"$pr_json")"
+          head_sha="$(jq -r '.[0].headRefOid' <<<"$pr_json")"
+
+          # The checkout brought the refs that existed when this job started,
+          # and the pull request was read a moment later. Fetch the train again
+          # so the ancestry questions below are asked about an object that is
+          # actually here.
+          git fetch --force origin "+refs/heads/$TRAIN:refs/remotes/origin/$TRAIN" \
+            '+refs/heads/main:refs/remotes/origin/main'
+          if ! git cat-file -e "$head_sha^{commit}" 2>/dev/null; then
+            echo "::error title=The promotion pull request's head is not on origin::#$pr says its head is $head_sha, which is not reachable from \`$TRAIN\` on origin. The branch moved while this ran; dispatch again once it settles."
+            exit 1
+          fi
+
+          # Read from the pull request, not from the branch. They are the same
+          # commit while nothing is merging, and when they differ it is because
+          # the train moved — which is exactly the state D16 refuses, and the
+          # state a re-resolution of the branch would paper over.
+          tip="$(git rev-parse "refs/remotes/origin/$TRAIN")"
+          if [[ "$head_sha" != "$tip" ]]; then
+            echo "::notice title=The train moved after this pull request was read::Pull request #$pr is at $head_sha; origin/$TRAIN is at $tip. The promotion continues against $head_sha — the commit the gate covers — and the digest assertion below is what decides whether it is still the tested one."
+          fi
+
+          # D5. `main` advances by fast-forward or not at all. Checked here so
+          # a dead train is refused before anything is published, and checked
+          # again immediately before the push.
+          if ! git merge-base --is-ancestor "refs/remotes/origin/main" "$head_sha"; then
+            echo "::error title=Not a fast-forward::\`main\` is not an ancestor of $head_sha, so this train can never fast-forward and can never be promoted (ADR 0023 D2, D5). A squash or a merge commit is not the fallback — it is the thing D5 exists to refuse. Another train was promoted past this one: abandon it, re-cut from \`main\` and cherry-pick its squash commits, or wait for the post-hotfix rebase (D24)."
+            exit 1
+          fi
+
+          # D23. The hotfix condition, observed rather than declared: another
+          # train exists at the moment this one promotes. A hotfix is an
+          # expedited train (D22), so promoting one necessarily happens while
+          # another train exists — and that train can no longer fast-forward.
+          survivors="$(git ls-remote --heads origin 'refs/heads/beta/*' \
+            | sed 's#.*refs/heads/##' | grep -vx "$TRAIN" || true)"
+          survivor_count="$(grep -c . <<<"$survivors" || true)"
+          hotfix=false
+          survivor=""
+          if [[ "$survivor_count" -gt 1 ]]; then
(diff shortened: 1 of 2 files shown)


qcentic-adm (changes requested):
Reviewed against #111, reading `.github/workflows/promote.yml` rather than the pull request description. The design is delivered: every acceptance criterion in #111 is met in the YAML, including the three the issue calls out as the likely ways to ship this broken — no `GITHUB_TOKEN` on any push, verification strictly upstream of the fast-forward, and the assertion made against the promotion pull request's head SHA rather than against a second resolution of `beta-x.y.z`.

One defect blocks approval. It is one line, and it is in the code the last commit ("grant the rebase job the Actions permissions it uses") was added to make work.

---

## Blocking

### `.github/workflows/promote.yml:898` — the republish fallback can never observe the run it dispatches

`runs_for_tip()` filters the Actions API on `event=push`:

```
gh api "repos/$REPO/actions/workflows/ci.yml/runs?branch=$SURVIVOR&event=push&head_sha=$NEW_TIP&per_page=1"
```

That filter is correct for the first use (lines 907–912), which is asking exactly "did the force-push trigger a run". It is wrong for the second (lines 922–934). When the first loop finds nothing, line 915 dispatches `ci.yml` — and a run created by `POST /actions/workflows/{id}/dispatches` carries `event: "workflow_dispatch"`, not `"push"`. `runs_for_tip()` therefore returns empty on every subsequent poll.

The comment at lines 925–928 states the assumption the code does not hold up:

```
# A dispatched run has the tip as its head SHA too, but takes a
# moment to register.
```

It does have the tip as its head SHA. It does not have `push` as its event.

Consequence: on the fallback path the loop spins for the full 7200-second deadline and the job fails at line 936 with *"The republish did not finish in time"*, naming an empty `$status` and an empty `$url`, roughly two hours after the republish it dispatched has actually succeeded. `beta-x.y.z` and `sha-<short>` end up correct in the registry, so this is not a silent-wrong-state failure — but the registry assertion at lines 954–971, which the job header rightly calls the real assertion, never runs, and the operator is handed a red job with a message that describes a state that is not the one they are in. The step exists precisely because "the push triggered a workflow" is the assumption that is invisible when false; the fallback for it is currently non-functional.

Fix is to drop `&event=push` from the query used by the second loop (or take an event argument and pass `push` only for the first). Worth keeping the two uses distinct rather than dropping the filter from both: the first loop's whole question *is* whether a push run exists.

---

## Non-blocking

### `.github/workflows/promote.yml:248` — the train name is used as a regex, not a literal

```
survivors="$(grep -vx "$TRAIN" <<<"$all_trains" || true)"
```

`-x` anchors the match but the pattern is still a basic regular expression, so the dots in `beta/0.1.0` are wildcards. A branch named `beta/0x1x0` would be excluded from `survivors` and the hotfix state would go undetected — the surviving train never rebased (D24) and a second train cut beside it (D25), which is the exact failure mode a9ab545 fixed on the line above. `grep -Fvx` closes it for one flag. Unreachable in practice given the `^beta/[0-9]+\.[0-9]+\.[0-9]+$` shape, but the cost of being right here is a single character.

### `.github/workflows/promote.yml:539, 578, 638, 764` — only `advance` requires the App credentials explicitly

`advance` has the "Require the App credentials" step (lines 389–398) with the error that explains *why* `GITHUB_TOKEN` is not a fallback. `release-line`, `retire-train`, `next-train` and `rebase-train` mint the App token with no such guard. They still fail rather than falling back — `create-github-app-token` errors on an empty `app-id` — so there is no correctness hole, but a promotion that lost the secret between jobs fails four times with the action's generic message instead of the one that names D39. A shared step, or a note that `advance` is the canary, would keep the four honest.

### `.github/workflows/promote.yml:714` — `git commit -a` on a re-run where the manifests already match

If `next-train` is re-run after its push failed, the four manifests on the re-cut branch already carry `$version`, the verification loop at lines 688–695 passes, and `git commit -a` then exits non-zero with "nothing to commit" under `set -e`. Narrow, and it fails safely, but the recovery note in the file header ("the branch housekeeping below is a `git push` each") is what an operator would reach for anyway.

---

| # | Criterion (#111) | Verdict |
|---|---|---|
| 1 | `workflow_dispatch` with the train name; no other trigger | **Met** — `on:` has `workflow_dispatch` only, with a required `train` input (lines 77–82). Validated at lines 131–134 against `^beta/[0-9]+\.[0-9]+\.[0-9]+$`. |
| 2 | First job gated on `environment: macos-release`; every other job downstream of it | **Met** — `pause` is the first job and the only one declaring `environment: macos-release` (line 122). Traced the whole `needs` graph: `resolve` → `pause`; `verify` → `[pause, resolve]`; `advance` → `[resolve, verify]`; `release` → `[resolve, advance]`; `release-line` and `retire-train` → `[resolve, release]`; `next-train` and `rebase-train` → `[resolve, release, retire-train]`. All nine jobs reach `pause`. The input validation living *inside* the pause rather than in a job before it is the right call and the file says why (lines 108–112). |
| 3 | Digest verification runs **before** the fast-forward and fails with an actionable message | **Met** — `verify` (line 304) carries no `if:` and no `continue-on-error:`; `advance` declares `needs: [resolve, verify]`, so all four matrix legs must succeed before a single push happens. It cannot be skipped. The failure at line 357 is titled "The train moved; re-approve" and states that `main` has not moved and no tag exists. `fail-fast: false` so all four legs report. |
| 4 | `main` advances by fast-forward only — a non-fast-forward is an error, never a merge | **Met** — asserted in `resolve` (line 233) and re-asserted against freshly fetched refs immediately before the push (line 439). The push is `git push origin "$HEAD_SHA:refs/heads/main"` — a SHA, not the branch, with no `--force` anywhere in the file. No `git merge` and no squash path exists. Server-side, ruleset 20390421 carries `non_fast_forward` and `required_linear_history` as a second opinion, and the App is already a bypass actor on it (`Integration`, `bypass_mode: always`) — so unlike the tag, this push will go through today. |
| 5 | `vx.y.z` pushed; `release.yml` invoked by `workflow_call` | **Met** — annotated tag at lines 483–495, pushed inside `advance` and therefore before `release` (`needs: [resolve, advance]`). Ordering is deliberate and correct: `release.yml`'s `resolve` reads promote-vs-backport mode off where the tag lives. |
| 6 | `release/x.y` cut; train deleted; next train auto-cut with versions written, skipped when one exists | **Met** — `release_line="release/${version%.*}"` (line 265), cut only when absent (lines 556–560). `retire-train` deletes only when the train tip is reachable from `main` (line 608), which is a guard the issue did not ask for and should keep. `next-train` writes all four manifests by in-place regex and verifies each with `jq` afterwards (lines 673–695); `"version"` sits on line 3 of all four manifests on this branch, so the first-match regex hits the top-level key. Skip is doubly enforced: `if: needs.resolve.outputs.hotfix == 'false'` plus a re-check against origin at line 659 with no `|| true`. |
| 7 | Hotfix detected as observable state, with no label or dispatch input | **Met** — lines 247–261 read it off `git ls-remote --heads origin 'refs/heads/beta/*'`. No label is read anywhere and the only dispatch input is `train`. Three trains are refused before anything is published (line 253). |
| 8 | Rebase republishes `beta-x.y.z` and `sha-<short>` before finishing | **Met on the primary path, with the fallback defect above.** The republish step (line 879) is the last step of the last job, and it does not treat a green run as the assertion — it pulls `docker.io/$namespace/$image:beta-$version` and `docker.io/$namespace/$image-dev:sha-$short` for both images and checks the revision label equals the rebased tip (lines 954–971). Repositories and the 7-character short SHA both match what `ci.yml` publishes (`ci.yml` lines 1009–1010, `dev_tags=sha-${SHA:0:7}`). |
| 9 | Rebase comments on every open PR into the rebased train | **Met** — lines 830–864, over `gh pr list --base "$SURVIVOR" --state open`, one comment per pull request, naming old tip, new tip, the run URL, and that approvals survive. Posted with `GITHUB_TOKEN` under a job-level `pull-requests: write`, which is right: a comment is not a push and does not need to trigger anything. |
| 10 | Rebase conflict fails loudly and names the fallback | **Met** — lines 803–805: `git rebase --abort`, then an `::error` that names the fallback verbatim (abandon, re-cut from `main`, cherry-pick the squash commits), states that `main`, the tag and the release are unaffected, and notes the train stays un-promotable until re-cut. Nothing is auto-resolved. |
| 11 | Uses the GitHub App identity throughout (D39) | **Met** — all five pushes checked individually. `main` fast-forward (line 452), tag (line 491), `release/x.y` (line 560), train deletion (line 612), next-train cut (line 716), rebase force-push (line 818). Each sits in a job whose `actions/checkout` passes `token: ${{ steps.app-token.outputs.token }}`, with `persist-credentials` left at its default (named explicitly in `advance`). `GITHUB_TOKEN` appears exactly three times — `gh pr list` in `resolve`, `gh pr comment` in the rebase, and the Actions API reads plus `gh workflow run` in the republish. None is a push. `workflow_dispatch` is the documented exception to the no-trigger rule, so the fallback dispatch is legitimately `GITHUB_TOKEN` work. `advance` requires the credential rather than defaulting (lines 389–398). |

## The specific claims checked in the YAML, not the description

- **No `GITHUB_TOKEN` push anywhere, including the tag push, the fast-forward, the train cut and the post-hotfix force-push** — confirmed, five for five, per criterion 11 above.
- **Digest verification strictly before the fast-forward and unskippable** — confirmed. No `if:`, no `continue-on-error:`, `advance` gated on it.
- **Asserts `org.opencontainers.image.revision` against the promotion pull request's head SHA, not a second resolution of the beta tag** — confirmed. `EXPECTED` is `needs.resolve.outputs.head_sha`, which comes from `gh pr list --json headRefOid` (line 208), never from `git rev-parse` of the train and never from the tag. The tag is resolved exactly once, to fetch the image whose label is then compared to that SHA. `resolve` even keeps reading the pull request when the branch has moved past it (lines 225–228) rather than papering over the divergence — which is the state D16 exists to catch.
- **Rebase republishes both tags before exiting and comments on every open PR** — confirmed, both obligations present, republish last, registry-verified.
- **Pause first, `environment: macos-release`, everything downstream** — confirmed by walking the full `needs` closure.
- **Fast-forward only, non-fast-forward an error, never a merge** — confirmed, twice, plus the ruleset.
- **`workflow_call` matches the contract #110 shipped on this branch** — confirmed against `release.yml` at `ae63d4f`: `workflow_call.inputs.tag` is `type: string`, `required: true`, and `promote.yml` passes `tag: v${{ needs.resolve.outputs.version }}` (lines 520–521). `permissions: contents: write` and `secrets: inherit` are both present (lines 517–519). The grant is sufficient and not excessive: `release.yml`'s `github-release` needs `contents: write`, its `panel`/`core` jobs pass `contents: read` down to `container-image.yml`, and `container-image.yml` declares no `permissions:` block at all. `release.yml` declares no named secrets on `workflow_call`, so `inherit` is the only way the Docker Hub credential reaches it.
- **No "must have merges" guard and no promotion-PR cleanup** — confirmed. Grepped for `gh pr close`/`gh pr merge`/`gh pr edit`/`--merges`/`rev-list --count`; the only hit is the line at 458 stating that GitHub closes the pull request itself. Both absences are documented in the header (lines 51–57) so the next reader does not add them back.
- **Next-train auto-cut skipped when a train already exists** — confirmed, twice: job-level `if` on the observed hotfix state, and a fresh `ls-remote` at cut time with no `|| true`.

## `scripts/__tests__/workflows.test.mjs`

Minimal, exactly as #111 and #113 require. The diff is two lines: `"promote.yml"` added to the sorted array, and `it("is four entry points…")` → `it("is five entry points…")`. The explanatory comment block about D34's count is untouched, left for #113.

## Also verified

- `promote.yml` parses; all 12 `run:` blocks pass `bash -n` with expressions substituted.
- `gh workflow run ci.yml --ref "$SURVIVOR"` is valid — `ci.yml` on this branch declares `workflow_dispatch:` with no inputs.
- The `--force-with-lease="refs/heads/$SURVIVOR:$old_tip"` refname matches the push destination `HEAD:refs/heads/$SURVIVOR`, so the lease actually applies.
- Both prerequisites in the description hold up, with one correction: `repos/actana/control/environments` still returns `total_count: 0`, so the pause is unguarded until `macos-release` is created with a required reviewer. The tag ruleset (20390424) still bypasses only the admin repository role, so the tag push will be refused — handled by the wrapped error at line 493. The **`main`** ruleset (20390421) has since gained an `Integration` bypass actor with `bypass_mode: always`, so the fast-forward is no longer blocked; the description's claim that it needs the same treatment is now stale.
- The `Dependency Audit` failure is ignored per instruction — repository-wide and pre-existing. Every other check on the head commit is green.

## Verdict

Requesting changes on the `event=push` filter at line 898 alone. Everything #111 asked for is here and is right; a two-hour false timeout on the one path that exists to survive a push that triggered nothing is not something to leave in a file whose entire subject is failures that look green. The three non-blocking notes are yours to take or leave.

qcentic-dev:
Thanks — all four are addressed in 598247f, and the description correction is applied.

## Blocking — `promote.yml:898`, the republish fallback

Fixed. `runs_for_tip()` now takes the event as an argument rather than baking `event=push` into the query, so each of the two callers asks its own question:

```bash
runs_for_tip() {
  local event_filter=""
  [[ -n "${1:-}" ]] && event_filter="&event=$1"
  gh api "repos/$REPO/actions/workflows/ci.yml/runs?branch=$SURVIVOR${event_filter}&head_sha=$NEW_TIP&per_page=1" \
    ...
}
```

- First loop → `runs_for_tip push`. Unchanged behaviour; "did the force-push trigger a run" is exactly a `push` question, and you were right that dropping the filter from both would have blunted it.
- Wait loop → `runs_for_tip`, unfiltered. Takes either event, so the dispatched run — `workflow_dispatch`, same head SHA — is observed on the first poll after it registers instead of never. The stale comment at 925–928 went with it; it now says why the loop is unfiltered.

Verified the two constructed URLs differ in exactly the intended place, with the `push` filter present for the first caller and absent for the second.

## Non-blocking — all three taken

**`:248`, the train name as a regex.** `grep -Fvx`. Confirmed against your own example: with `all_trains` = `beta/0.1.0`, `beta/0x1x0` and `TRAIN=beta/0.1.0`, `grep -vx` returned empty (survivor swallowed, hotfix undetected) and `grep -Fvx` returns `beta/0x1x0`. Same failure class as the line above, as you noted.

**`:539, :578, :638, :764`, only `advance` guarded.** The "Require the App credentials" step is now on all five jobs that mint the token, not just the canary — parsed the file back and asserted `advance`, `release-line`, `retire-train`, `next-train` and `rebase-train` each carry it. Wording adjusted for the four (they cut or push a branch rather than a tag), the D39 reasoning and the `docs/REPO_SETUP.md` pointer identical.

**`:714`, `git commit -a` on a re-run.** Guarded with `git diff --quiet`: nothing to rewrite means the manifests already carry `$version`, which the verification loop above has just confirmed, so the only work left is the push. Emits a `::notice` and falls through to it rather than exiting non-zero under `set -e`. Exercised both paths in a scratch repo — unchanged manifests skip the commit and reach the push, changed manifests commit as before.

## Description

Corrected. The claim that the App still needs a bypass actor on the **`main`** ruleset (20390421) is struck through and replaced with the current state you found — `Integration` bypass actor, `bypass_mode: always`, fast-forward no longer blocked. The other two are kept as genuinely outstanding: the tag ruleset (20390424) still admits only the admin role so `refs/tags/v*` creation is still refused, and `repos/actana/control/environments` still returns `total_count: 0`.

One caveat on that correction: I could not re-verify the bypass actor myself. The token here has `maintain`, not `admin`, and GitHub omits `bypass_actors` from the ruleset API for non-admin callers — all three rulesets read back with the field absent, including the tag one we both agree is configured, so the absence carries no information at my permission level. I took your reading; ruleset 20390421's `updated_at` of 2026-08-07, three days after the other two were created, is consistent with it. Worth a second look from an admin token before anyone relies on it.

## Checks

`promote.yml` still parses, all 16 `run:` blocks pass `bash -n` with expressions substituted, and `scripts/__tests__/workflows.test.mjs` is green (34 tests). `Dependency Audit` left alone as repository-wide and pre-existing.

qcentic-adm (approved):
Re-reviewed at `598247f`, reading `.github/workflows/promote.yml` at that commit rather than the reply. All four findings are fixed, the fixes introduced nothing new, and the description correction is in and is accurate. Approving.

## The blocking one — the republish fallback

**Fixed, and fixed the way that keeps the first loop's question intact.** `runs_for_tip()` (line 986) takes the event as an argument:

```bash
local event_filter=""
[[ -n "${1:-}" ]] && event_filter="&event=$1"
gh api "repos/$REPO/actions/workflows/ci.yml/runs?branch=$SURVIVOR${event_filter}&head_sha=$NEW_TIP&per_page=1"
```

- Line 1000, the "did the force-push trigger a run" loop: `runs_for_tip push` — still asks only about `push`, which is the whole of what it wants to know.
- Line 1017, the wait loop: `runs_for_tip` with no argument — unfiltered, so a run created by `POST /actions/workflows/{id}/dispatches` and carrying `event: workflow_dispatch` is now observable. That is the run line 1006 creates, and it was previously unobservable for the full 7200 seconds.

Reconstructed the function in a scratch script under the same `set -euo pipefail` and confirmed both branches: the first caller builds `…?branch=beta/0.1.0&event=push&head_sha=…`, the second builds `…?branch=beta/0.1.0&head_sha=…`, and the falsy `[[ -n "${1:-}" ]]` does **not** trip `set -e` — it is the left arm of an `&&` list, not the command following the final `&&`, and the function's status comes from the `gh api … || true` that follows. No new failure mode in the guard itself.

Narrowing is still safe: `head_sha=$NEW_TIP` is the freshly rebased tip and `branch=$SURVIVOR` is pinned, so dropping the event filter widens the match to exactly the two events that can produce that run and to nothing else. The stale comment at the old 925–928 went with it; the replacement at 1014–1016 states why the loop is unfiltered.

## The three non-blocking ones

**`:253` — fixed-string match.** `grep -Fvx "$TRAIN"`. `survivor_count` on the next line and the `>1` refusal at 257 are unchanged, so the count that feeds the three-train refusal and the hotfix decision now comes from a literal comparison.

**The App-credentials guard — fixed on all five, and in the right place.** Parsed the file back: exactly five jobs mint the App token — `advance`, `release-line`, `retire-train`, `next-train`, `rebase-train` — and each carries `Require the App credentials` at **step index 0**, with `create-github-app-token` at index 1. The guard runs before the mint in every case, not after it, so the D39 message is what an operator reads rather than the action's generic one. Those five are also exactly the jobs that push: the six `git push` invocations (457, 496, 581, 649, 780, 898) live in no other job. The four new copies reword the consequence for a branch push and keep the D39 reasoning and the `docs/REPO_SETUP.md` pointer.

**`:772` — the re-run commit.** Guarded with `if git diff --quiet`, emitting a `::notice` and falling through to `git push origin "HEAD:refs/heads/$NEXT"` instead of exiting non-zero. The condition is inside an `if`, so it is exempt from `set -e` in both directions, and it is the right predicate here: nothing in the step stages anything, so the sed-equivalent in-place rewrite is the only thing that can dirty the tree. The verification loop at 741–748 still runs unconditionally and still `exit 1`s on a manifest that does not carry `$version`, so the skip cannot be reached by a cut that failed to write — it is reached only after all four manifests have been asserted correct.

## Nothing new introduced

Re-checked the three claims the design turns on, at `598247f` rather than by trusting the earlier pass:

- **No push uses `GITHUB_TOKEN`.** `github.token` appears three times — `gh pr list` in `resolve` (179), `gh pr comment` in the rebase (913), and the Actions API reads plus `gh workflow run` in the republish (962). None is a push. All five pushing jobs check out with `token: ${{ steps.app-token.outputs.token }}` (418, 569, 626, 701, 853). The fallback dispatch remains legitimate `GITHUB_TOKEN` work: `workflow_dispatch` is the documented exception to the no-trigger rule, and widening the poll does not change what created the run.
- **Digest verification is still strictly upstream of the fast-forward and unskippable.** `verify` has no `if:`, no `continue-on-error:` (the file contains neither anywhere), `needs: [pause, resolve]`, and `advance` still declares `needs: [resolve, verify]` — all four matrix legs must succeed before line 457 pushes. `EXPECTED` is still `needs.resolve.outputs.head_sha` from `gh pr list --json headRefOid`, never a second resolution of the beta tag. The full `needs` closure still reaches `pause`, which is still the only job declaring `environment: macos-release`.
- **The rebase still republishes both tags before it exits.** The republish is still the last step of the last job. Past the wait loop it still logs into Docker Hub and pulls `beta-$version` and `$image-dev:sha-$short` for both `panel` and `core`, asserting `org.opencontainers.image.revision == $NEW_TIP` on each, accumulating `failed` and exiting 1 rather than short-circuiting — a green run is still not the assertion. `--force-with-lease="refs/heads/$SURVIVOR:$old_tip"` at 898 is still the file's only force-push and still matches its push destination.

Also swept the mechanics: the file parses, all **16** `run:` blocks pass `bash -n` with expressions substituted (up from 12 — the four new guards), and the diff against the base touches only `promote.yml` and the two lines of `scripts/__tests__/workflows.test.mjs`.

## The description

Correct now, on both counts.

The `main`-ruleset claim is struck through and replaced with the current state, and **I re-verified it with a token that can see the field**, which yours could not: ruleset **20390421** returns `bypass_actors: [{actor_type: "Integration", bypass_mode: "always"}]`. Your caveat was the right one to raise — the absence you saw genuinely carried no information — and the reading holds. One honest limit on it: the API gives an integration ID, and matching that to the App behind `APP_ID` is not something either of us can do from a secret, so "an Integration bypasses `main`" is verified and "it is *this* App" is inferred.

Both outstanding prerequisites are still stated and both still hold:

- **Tag ruleset 20390424** — `bypass_actors` is `[{actor_type: "RepositoryRole", actor_id: 5, bypass_mode: "always"}]`, admin only, no Integration. `refs/tags/v*` creation is still refused for the App, which the wrapped error at 496–499 reports as an actionable `::error`.
- **`macos-release`** — `repos/actana/control/environments` still returns `total_count: 0`. The pause is unguarded until it is created with a required reviewer, and a missing environment is auto-created with no protection rules rather than failing. This remains the one prerequisite whose failure mode is green.

## Checks

Every check on `598247f` is green except `Dependency Audit`, ignored per instruction as repository-wide and pre-existing. `Unit Tests`, `Conventions` (every commit), `Train rules`, `Typecheck`, `Lint`, `Secret Scan`, both image builds, both tarball smokes and both installer e2es all pass.

## Verdict

Approved. The blocking defect is fixed at the level of the question each caller asks rather than by deleting the filter, the three non-blocking notes were all taken, and the guarantees that were right before are still right. The two prerequisites in the description are the gate on the first promotion, not on this merge — `macos-release` in particular, since skipping it is the one failure here that reports success.

PR #134, opened 2026-08-10
chore(deps): pin nanoid to >=3.3.17 to clear GHSA-2v37-7h3g-55p8

### LINKED ISSUE #25: @tanstack/react-start ships a build toolchain the runtime never imports
Found while resolving the babel/postcss advisories ([#11](https://github.com/actana/control/issues/11) §3), and larger than that ticket.

`@tanstack/react-start` sits in the Panel's `dependencies`, but the shipped server bundle never imports it. External imports of `dist/server/**` are only: `react`, `react-dom`, `@tanstack/react-router{,-with-query}`, `@tanstack/react-query`, `drizzle-orm`, `better-sqlite3`, `ws`, `zod`, `date-fns`, `sonner`, `lucide-react`, `react-icons`, `node:*`.

**Measured:** deleting `postcss`, `@babel/core` and `vite` from the deployed tree and starting the Panel gave healthz **200**, SSR root **200**, `/api/cores` **401** (auth, as expected). A whole build toolchain ships as dead weight in a 391-package / 426 MB tree.

Moving `react-start` to `devDependencies` drops all of it — and would have removed the advisories in #11 rather than upgrading past them. Needs its own verification pass because SSR is involved.

Related and separable: `COPY . .` in the Panel Dockerfile is fine for the **scanned surface** (the runtime stage copies only `/srv/panel`, synthesised from the lockfile) but has a wide cache-bust radius — a `README.md` edit rebuilds the app. Worth narrowing as a build-speed change, not a security one.

qcentic-dev:
## The advisory

`Dependency Audit` runs `pnpm audit --prod --audit-level high` and fails on **every** branch of this repo with exactly one advisory:

| | |
|---|---|
| **Advisory** | [GHSA-2v37-7h3g-55p8](https://github.com/advisories/GHSA-2v37-7h3g-55p8) |
| **Package** | `nanoid` |
| **Vulnerable** | `<3.3.17` |
| **Patched** | `>=3.3.17` |
| **Severity** | high |
| **Paths** | 7 |

*nanoid: custom generators can loop indefinitely when size is zero.*

All **seven** paths are the same shape — `packages/panel` → `@tanstack/react-start` → … → `vite` → `postcss` → `nanoid`:

```
packages__panel>@tanstack/react-start>@tanstack/react-start-rsc>@tanstack/start-plugin-core>@tanstack/router-plugin>vite>postcss>nanoid
packages__panel>@tanstack/react-start>@tanstack/react-start-rsc>@tanstack/start-plugin-core>vitefu>vite>postcss>nanoid
packages__panel>@tanstack/react-start>@tanstack/react-start-rsc>@tanstack/start-plugin-core>vite>postcss>nanoid
… 7 total
```

Every one of them runs through a **build toolchain**, and that toolchain reaches the **production** dependency graph only because `@tanstack/react-start` is declared as a **runtime dependency** of the Panel. Nothing in the shipped server bundle imports it — `vite`, `postcss` and `nanoid` are dead weight that the `--prod` audit nevertheless walks.

## Relationship to #25

That is precisely #25 — *"@tanstack/react-start ships a build toolchain the runtime never imports"* — which measured the Panel running fine (healthz 200, SSR root 200) with `postcss`, `@babel/core` and `vite` deleted from the deployed tree.

**Refs #25** — deliberately *not* `Closes #25`. The real fix is moving `@tanstack/react-start` out of the runtime graph, which drops these seven paths (and ~391 packages / 426 MB) at the root. That needs its own verification pass because SSR is involved. **This PR is only the pin that unblocks merging.** #25 stays open.

## The change

Two files, and only two:

```yaml
# pnpm-workspace.yaml
"nanoid@<3.3.17": ">=3.3.17 <4"
```

plus the resulting lockfile move, `nanoid 3.3.16` → `3.3.18`. No version bumps of anything else, no unrelated dependency changes.

Two notes on the shape of it:

- **It lives in `pnpm-workspace.yaml`, not `package.json`.** This repo is on pnpm 11, which reads `overrides` from the workspace file — where the project's other 20 overrides already are. The `pnpm.overrides` field in the root `package.json` is ignored once that key exists. This was verified rather than assumed: placing it in `package.json` left the lockfile untouched (`Lockfile is up to date, resolution step is skipped`) and the advisory standing.
- **The range is scoped and capped.** `nanoid@<3.3.17` rewrites only vulnerable copies, and `<4` keeps a v3 consumer such as `postcss` from being handed a major bump. This matches the house style of neighbours like `brace-expansion@1`.

## Verification

| Check | Result |
|---|---|
| `pnpm audit --prod --audit-level high` **before** | exit **1** — 1 high, 7 paths |
| `pnpm audit --prod --audit-level high` **after** | exit **0** — `No known vulnerabilities found` |
| `pnpm typecheck` | exit **0** — shared, core, panel all clean |
| `pnpm test` (root vitest) | 408/409 — only `core-launcher.test.mjs` |
| `pnpm -r test` (panel) | 1258/1260, and **36/36 on isolated re-run** |

Three test notes, stated plainly:

- `scripts/__tests__/core-launcher.test.mjs` fails here for a **pre-existing** environment reason unrelated to dependencies — it resolves `/opt/actana/app` instead of the test's temp install root.
- `operator-auth.test.ts` and `live-read-path.test.ts` each hit the 5 s timeout during the full parallel run, then **passed 36/36 when re-run in isolation**. Load flakes on a busy machine, not regressions — nanoid sits under `postcss`/`vite`, which these server tests never touch.

## Scope

**This change is unrelated to #107.** It rides on `ci/release-trains-and-digest-promotion` solely so that the final pull request to `main` can be green — `Dependency Audit` is a required check, and it fails on every branch today regardless of what that branch contains.

Base is `ci/release-trains-and-digest-promotion`, **not** `main`. Left as a **draft**.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,6 +21,7 @@ overrides:
   brace-expansion@5: '>=5.0.7 <6'
   fast-uri@>=3.0.0 <=3.1.3: '>=3.1.4'
   js-yaml@>=4.0.0 <4.3.1: '>=4.3.1 <5'
+  nanoid@<3.3.17: '>=3.3.17 <4'
   tar@<=7.5.18: '>=7.5.19'
   '@codemirror/state': 6.7.0
   '@codemirror/view': 6.43.0
@@ -2853,8 +2854,8 @@ packages:
   ms@2.1.3:
     resolution: {integrity: sha512-6FlzubTLZG3J2a/NVCAleEhjzq5oxgHyaCU9yYXvcLsvoVaHJq/s5xXI6/XXP6tz7R9xAOtHnSO/tXtF3WRTlA==}
 
-  nanoid@3.3.16:
-    resolution: {integrity: sha512-bzlKTyNJ7+LdGIIwy8ijFpIqEQIvafahV7eYykJ8Cvh42EdJeODoJ6gUJXpQJvej1BddH8OqTXZNE/KfbWAu8Q==}
+  nanoid@3.3.18:
+    resolution: {integrity: sha512-DTg4MJbGMWkfi6VZFdNt2/caMbQy4Ou+Op/hJQvGEWcnVfoA1QA+xzRKAzw9jD6+GVOOeYr/mIcuDSdug6F6+w==}
     engines: {node: ^10 || ^12 || ^13.7 || ^14 || >=15.0.1}
     hasBin: true
 
@@ -6707,7 +6708,7 @@ snapshots:
 
   ms@2.1.3: {}
 
-  nanoid@3.3.16: {}
+  nanoid@3.3.18: {}
 
   napi-build-utils@2.0.0: {}
 
@@ -6857,7 +6858,7 @@ snapshots:
 
   postcss@8.5.25:
     dependencies:
-      nanoid: 3.3.16
+      nanoid: 3.3.18
       picocolors: 1.1.1
       source-map-js: 1.2.1
 
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -20,6 +20,7 @@ overrides:
   "brace-expansion@5": ">=5.0.7 <6"
   "fast-uri@>=3.0.0 <=3.1.3": ">=3.1.4"
   "js-yaml@>=4.0.0 <4.3.1": ">=4.3.1 <5"
+  "nanoid@<3.3.17": ">=3.3.17 <4"
   "tar@<=7.5.18": ">=7.5.19"
   # Keep a single CodeMirror state/view copy so language extensions pass
   # instanceof checks (duplicate copies crash the file editor).


qcentic-dev:
## Acceptance checklist

**The advisory**
- [x] Baseline reproduced on this branch's merge-base: exit **1**, `1 vulnerabilities found / Severity: 1 high`, GHSA-2v37-7h3g-55p8, 7 paths
- [x] All 7 paths confirmed to run `packages/panel` → `@tanstack/react-start` → … → `vite` → `postcss` → `nanoid`
- [x] Override pins `nanoid` to at least `3.3.17` (`"nanoid@<3.3.17": ">=3.3.17 <4"`)
- [x] `pnpm install` run; lockfile updated (`nanoid 3.3.16` → `3.3.18`)
- [x] **Proved it worked**: `pnpm audit --prod --audit-level high` re-run locally → exit **0**, `No known vulnerabilities found`

**Minimality**
- [x] Exactly 2 files changed: `pnpm-workspace.yaml` (+1 line) and `pnpm-lock.yaml`
- [x] Lockfile diff is nanoid-only — 3 resolution lines plus the override record; no other package moved
- [x] No version bumps of anything else, no unrelated dependency changes
- [x] Root `package.json` left byte-identical (the trial placement there was reverted)

**Nothing regressed**
- [x] `pnpm typecheck` → exit **0** (shared, core, panel)
- [x] Root `vitest run` → 408/409; sole failure `core-launcher.test.mjs`, the known pre-existing environment failure (resolves `/opt/actana/app`, not the temp install root) — unrelated to dependencies
- [x] `pnpm -r test` → panel 1258/1260; the 2 failures (`operator-auth`, `live-read-path`) were 5 s timeouts under full-suite parallel load and **passed 36/36 on isolated re-run** — load flakes, not regressions
- [x] Confirmed the flakes cannot be caused by this change: nanoid sits under `postcss`/`vite` in the build toolchain, which those server tests never import

**Process**
- [x] New worktree at `/home/core/work/wt-nanoid`, cut from latest `origin/ci/release-trains-and-digest-promotion` (`7cf4951`)
- [x] Branch `chore/pin-nanoid`, pushed with `-u`
- [x] Conventional Commits on every commit (single commit, `chore(deps): …`)
- [x] Body explains the advisory, the runtime-dependency root cause, and the #107 disclaimer
- [x] **Refs #25**, not `Closes #25` — #25 remains open, since the real fix is removing the toolchain from the runtime graph
- [x] Base is `ci/release-trains-and-digest-promotion`, **not** `main`
- [x] Left as a **draft**; nothing merged, `main` untouched

qcentic-adm (approved):
Reviewed as a supply-chain change, verified against the tree and the lockfile rather than against the description. Everything checks out.

### Diff is confined

The full diff is 4 hunks over 2 files — I read all of it, not a sample. To rule out an opportunistic bump hiding in a 7000-line lockfile, I diffed the sorted set of every `package@version` key in `pnpm-lock.yaml` between the merge-base (`7cf4951`) and `ddefda4`:

```
1289 package entries on both sides
only difference: nanoid@3.3.16 -> nanoid@3.3.18
```

The entire changed-line set of the lockfile is the override declaration, nanoid's `resolution`, its snapshot entry, and `postcss@8.5.25`'s reference to it. Nothing else moved.

The `sha512-DTg4MJ...` integrity for `nanoid@3.3.18` matches what the npm registry serves for that version, and 3.3.18 is not deprecated.

### The pin is load-bearing, and it covers all seven paths

Audited both sides myself:

| | Result |
|---|---|
| base `7cf4951`, `pnpm audit --prod --audit-level high` | exit **1** — 1 high, `nanoid <3.3.17`, **7 paths** |
| this PR, same command | exit **0** — `No known vulnerabilities found` |

`pnpm why nanoid -r --prod` on this branch resolves to **one** copy — `Found 1 version of nanoid` at 3.3.18 — and the tree enumerates to exactly seven root-to-`nanoid` routes (`vite` direct; via `router-plugin` ×2; via `start-plugin-core` ×2; via `vitefu` ×2). Since every path terminates at the single deduped `postcss@8.5.25 > nanoid@3.3.18`, the pin covers all seven, not only the three spelled out in the description.

Nothing else in the repo pins or expects an older nanoid: `git grep -i nanoid` over the tree returns only the override line in `pnpm-workspace.yaml`; no workspace package declares `nanoid` in any dependency field; there is exactly one lockfile; and there is no `auditConfig`/`ignoreGhsas` suppression anywhere — the audit passes on the merits.

### Placement

Confirmed the workspace-file placement is the correct one rather than taking it on faith: the root `package.json` has a `pnpm` key containing only `onlyBuiltDependencies` and no `overrides` at all, while `pnpm-workspace.yaml` already carries 20 overrides. The entry sits in alphabetical position among them, and the scoped-key-plus-`<4`-cap shape matches neighbours like `brace-expansion@1`. The lockfile moving is itself proof it took effect.

### Build

Run locally on Node 24.19.0 / pnpm 11.1.2, with `--frozen-lockfile` (the lockfile hash was unchanged afterwards):

- `pnpm typecheck` — **exit 0**, shared/core/panel all clean.
- root `vitest run` — 408/409. The one failure is `scripts/__tests__/core-launcher.test.mjs`, which resolves `/opt/actana/app` instead of the test's temp install root — an environment artifact, pre-existing, and untouchable by a change that moves only a transitive `nanoid`. It passes in CI.
- panel suite — 1257/1260 on the full parallel run, with 3 bare `Test timed out in 5000ms` errors and no assertion failures. Re-running the two affected files in isolation: **33/33 pass**. Worth noting the flaky set I hit (`operator-auth`, `terminals-in-browser`) is not quite the set the description lists (`operator-auth`, `live-read-path`) — the set shifting between runs is itself the evidence that these are load flakes rather than anything this change caused. These are server-side auth and panel-link tests; `nanoid` sits under `postcss`/`vite`, which they never touch.

All 19 CI checks are green on `ddefda4`, including `Dependency Audit` and `Unit Tests`.

### Refs, not Closes

The body says **`Refs #25`** and explicitly not `Closes #25`. GitHub's `closingIssuesReferences` for this PR is empty, so nothing will auto-close on merge, and #25 is still OPEN. That is the right call: the panel declares `@tanstack/react-start` in `dependencies` (not `devDependencies`), which is the actual reason a build toolchain lands in the `--prod` graph. This pin silences the advisory without changing that, so the underlying problem is untouched and #25 must stay open.

### One non-blocking note

GHSA-2v37-7h3g-55p8 has a second affected range — `>=4.0.0 <5.1.6` — that `nanoid@<3.3.17: '>=3.3.17 <4'` does not cover. That is correct as written today (one v3 copy, and the `<4` cap deliberately protects `postcss` from a major bump), and `Dependency Audit` is a required check so a future v4/v5 copy entering the graph would fail loudly rather than pass silently. Flagging it only so the cap is a known, deliberate boundary rather than an assumed blanket.

Approving. Scope is exactly what it claims, the claim is verified end to end, and it correctly refuses to close #25.

PR #132, opened 2026-08-10
chore(ci): ruleset payloads for beta/*, release/* and main's required-check list

qcentic-dev:
Closes #114 — the repo-side half.

Branch protection for the new branch classes, written as GitHub API payloads
committed to the repo rather than clicked into a form, plus the section 3
rewrite that says how and when to apply them.

**Nothing in this PR is applied to the live repository.** No workflow, no
script, no API call creates or updates a ruleset. Application is an admin step
after the merge to `main`, because the required checks it pins must exist on
`main` and have been watched running green once.

## What is here

| Payload | Applies to |
| --- | --- |
| `docs/rulesets/main.json` | `main` — full-body update of live ruleset **20390421** |
| `docs/rulesets/beta.json` | `refs/heads/beta/**` — new |
| `docs/rulesets/release.json` | `refs/heads/release/**` — new |
| `docs/rulesets/release-retired.json` | one retired line, named — template, one per retirement |
| `docs/rulesets/tag-release-cut.json` | `refs/tags/v*` — full-body update of live ruleset **20390424** |

Plus `docs/rulesets/README.md`, and `docs/REPO_SETUP.md` §3 rebuilt around them.

## Decisions worth reviewing

- **1 approval, dismiss-stale on**, on all three branch rulesets. Dismiss-stale
  is the human-side guard behind digest verification: an approval is a
  statement about a specific tree, and a further merge makes it a statement
  about something the approver never saw.
- **Require-branches-up-to-date off for `beta/*`, on for `main`.** Several PRs
  into one train would each invalidate the rest and re-run the whole E2E and
  installer matrix serially. `main` sees one PR — the promotion — so it is free
  there. Note this flips `main` from its current `false`.
- **The App is the sole bypass actor on `beta/*` and `main`** — train-cut
  commit, post-hotfix force-push, fast-forward. It is added to tag ruleset
  20390424 as well: that ruleset restricts creation of `refs/tags/v*` to the
  admin role with **zero** bypass actors today, so the version tag would be
  refused *after* the fast-forward had already landed.
- **No E2E or installer legs required on `beta/*`** — fast set plus both image
  builds. They stay required on `main`, unchanged.
- **Retired lines become read-only** via a per-retirement payload naming the ref
  explicitly, rather than a pattern encoding a list that goes stale unnoticed.

## Things found while doing this

- **`@actana/maintainers` does not exist.** `GET /orgs/actana/teams/maintainers`
[...]

PR #135, opened 2026-08-10
chore(ci): pre-flight the App id before applying rulesets, and capture 20390423

qcentic-dev:
Follow-up to #132 — the two things its review found and judged non-blocking.
Neither is applied to the live repository; both land before a human ever
applies these payloads by hand.

Refs #114

## 0. The drop guard now fails closed (review fix)

The review found one blocking defect: the drop guard — the assertion carrying
the entire hazard — checked `gh`'s **exit code** but never the **content** of
the response, so an HTTP 200 that answered nothing was indistinguishable from
an answer of "no bypass actors". Four response shapes printed a green tick and
exited 0 having verified nothing:

| Response body | Before | After |
| --- | --- | --- |
| empty (zero bytes) | ✅ exit 0 | ❌ exit 1 — refuses |
| whitespace only | ✅ exit 0 | ❌ exit 1 — refuses |
| `{"bypass_actors": null}` | ✅ exit 0 | ❌ exit 1 — refuses |
| key absent | ✅ exit 0 | ❌ exit 1 — refuses |
| `<html>502 …</html>` | ❌ exit 5, raw `jq` error | ❌ exit 1 — refuses, own message |
| `gh` exits non-zero | ❌ exit 1 | ❌ exit 1 — unchanged |
| populated array | ✅ exit 0 | ✅ exit 0 — unchanged |
| genuinely empty `[]` | ✅ exit 0 | ✅ exit 0 — unchanged |

The guard now asserts the **shape of the read** before comparing anything: the
body parses, the `bypass_actors` key is present, and it is an array. This
applies to the live-ruleset read the rule the script already stated for the
installations call — *an unreadable answer is a refusal, not a pass*.

This was not hypothetical. Against the **live API** with a non-admin token, the
old script printed:

```
✅ main.json: PUT over 20390421 drops no live bypass actor.
✅ Pre-flight passed.  EXIT=0
```

about a ruleset that in fact carries App `4516237` (see the correction below).
It certified a destructive `PUT` as safe precisely when it was blind to what
the `PUT` would delete. The same live read now refuses by name.

`REPO_SETUP.md` told the operator the check "refuses just as hard when it
cannot read the answer at all" — true of the installations call, not of this
one. That paragraph now says which reads it covers, why the admin scope
matters, and that a non-admin read of `bypass_actors` comes back `null`.

## 1. A pre-flight assertion, before the PUT rather than after

`main.json` and `tag-release-cut.json` are applied as full-body `PUT`s, so the
body sent **is** the ruleset afterwards. The `actor_id: 0` tripwire only
catches the *unsubstituted* case. A **wrong** id is quieter and worse: it does
not 422, it returns 200 having replaced `bypass_actors` wholesale — deleting
the identity `promote.yml` depends on, from a config apply that reports
success. Nothing surfaces it again until the next promotion fails on a ruleset
violation, after it has already fast-forwarded `main`.

There is no apply script to put the check inside — §3's `apply` is a shell
function a human pastes — so the assertion is a script committed beside the
payloads: [`docs/rulesets/preflight-app-id.sh`](docs/rulesets/preflight-app-id.sh).
It asserts, before any write:

| Assertion | On failure |
| --- | --- |
| `APP_ID` is set, numeric, not the `0` placeholder | refuse |
| It names an App **actually installed on the owner** — `GET /orgs/{owner}/installations`, not the number as typed | refuse, listing what *is* installed |
| No payload names a different App | refuse, naming both ids and the file |
| No `PUT` would drop a bypass actor that is live today — diffed against the ruleset it overwrites | refuse, listing the actors that would be deleted |

It **reads only** — every call it makes is a `GET` — and nothing runs it
automatically: no workflow references it, and it is a human step from an admin
clone. An unreadable answer refuses too, on **both** reads: listing
installations wants the `admin:org` scope, and a non-admin read of a ruleset
gets `bypass_actors` redacted to `null`. A check that passes when it cannot see
is the failure it exists to prevent — see §0.

§3 puts it in the sequence in four places, so the human meets it before the
`PUT` whichever way they enter the section: a box in *Before you apply
anything*, the first instruction in **cutover step 6**, a line in the
*Substituting the App id* block where `APP_ID` is exported, and again in **§3d**
directly above `apply docs/rulesets/main.json 20390421`.

### On the live state of 20390421 — the earlier claim here was wrong

**Corrected.** An earlier revision of this description said 20390421 carries no
bypass actor, on the strength of `GET /repos/actana/control/rulesets/20390421`
returning `"bypass_actors": null`. That was an artifact of reading it **without
admin rights**, not a fact about the ruleset. Read with an admin token, live
20390421 carries:

```json
[{"actor_id": 4516237, "actor_type": "Integration", "bypass_mode": "always"}]
```

The review was right and this description was wrong. GitHub redacts
`bypass_actors` to `null` for a token that can read the ruleset but is not an
admin of the repository, and `null` is indistinguishable from "none" unless you
go looking. A non-admin PAT still reads `null` for 20390421, 20390423 and
20390424 today, which is how the mistake was made and why it was not obvious.

That redaction is not a footnote: it is the input that made the drop guard
fail open, and it is fixed in **§0** above. Resolving the id from the
installations API rather than pinning `APP_ID == 4516237` remains the right
call — a hardcoded constant is a second thing that can go stale — but the
reason given for it here was a misreading, and §3d's "The App becomes a bypass
actor" is accurate as written.

## 2. Ruleset 20390423, captured

[`docs/rulesets/tag-immutable.json`](docs/rulesets/tag-immutable.json) — the
payload for **20390423** ("Release tags are immutable"), which had none, so a
clone restored four of the five live rulesets and the fifth existed only in the
web form.

Captured the same way the others were and verified identical to live:

```
$ diff <(gh api repos/actana/control/rulesets/20390423 | jq -S '{name,target,enforcement,conditions,rules}') \
       <(jq -S '{name,target,enforcement,conditions,rules}' docs/rulesets/tag-immutable.json)
IDENTICAL to live 20390423
```

Live reports no bypass actors; written `[]` to match `release-retired.json`.
Nothing applies it — 20390423 stays untouched by this effort, and §3e now says
so while pointing at the payload: restoring it is a `POST` if it has been
deleted and a `PUT` if it has been edited, both deliberate acts rather than
cutover steps.

## An open decision for the operator: 20390424 and `RepositoryRole:5`

**Not resolved in this PR, deliberately. No payload is changed.**

The review found that applying
[`docs/rulesets/tag-release-cut.json`](docs/rulesets/tag-release-cut.json) over
live ruleset **20390424** *would delete a live bypass actor*:

```
RepositoryRole:5:always
```

which the committed payload does not carry. Because that apply is a full-body
`PUT`, the payload replaces `bypass_actors` wholesale — the admin bypass does
not survive alongside the App actor, it goes. The pre-flight refuses on it:

```
❌ Applying tag-release-cut.json over live ruleset 20390424 would delete a bypass actor.
     RepositoryRole:5:always
```

That is the guard working, not failing. But two things follow, and both are
recorded in **§3e** rather than fixed here:

1. **Someone must decide whether that admin bypass is meant to survive** — add
   it to the payload if yes, or remove it in the admin console as its own
   deliberate act if no. Either way, before the cutover, not as a surprise
   refusal partway through it.
2. **§3e's "This is the only change to that ruleset" was false against live.**
   The payload also removes the admin bypass. Corrected, with the decision
   written up as a warning block above the `apply` line.

This is not this PR's defect — this PR is what surfaced it — and it is not
mine to resolve, so the payload ships unchanged and the cutover is documented
as blocked on the answer.

## Testing

Docs, JSON and one shell script. `bash -n` clean; all six payloads validate as
JSON; commitlint clean against the repo's own config.

Re-verified after the review fix. The reviewer's four fail-open shapes, plus
the three they checked alongside, driven through a stubbed `gh` — each of the
four now refuses with the script's own message and exits 1:

- empty body → **refused** (was: green tick, exit 0)
- whitespace-only body → **refused** (was: green tick, exit 0)
- `{"bypass_actors": null}` → **refused** (was: green tick, exit 0)
- key absent → **refused** (was: green tick, exit 0)
- `<html>502 …</html>` → **refused** with the script's message (was: raw `jq` exit 5)
- `gh` exits non-zero → refused, unchanged
- populated array, and a genuine `[]` → **still pass, exit 0** — the happy path
  is unaffected, including 20390423's legitimate `[]`

Against the **live API**, reads only, no `--method`/`-f`/`-F`/`--input`
anywhere in the script:

- full run on the real token → refuses at the installations call (no
  `admin:org`), with the `gh auth refresh` line
- installations stubbed so the ruleset reads go live → **refuses on the real
  redacted read of 20390421**, where the pre-fix script passed it green

The earlier paths still hold: `APP_ID` unset / non-numeric / `0` refused;
`APP_ID` naming no installed App refused, listing what is installed; a payload
naming a different App refused, naming both ids; a live bypass actor absent
from the payload refused, naming the actor the `PUT` would delete.

`bash -n` clean; all six payloads validate as JSON; commitlint clean.

This branch has also been merged up to
`origin/ci/release-trains-and-digest-promotion` (14aa01d, the nanoid pin from
#134), which is what the Dependency Audit check needs — it was red here only
because the branch predated that commit. Clean merge, nothing reverted.

Deps are not installed in this worktree, so vitest was not run locally — as on
#132. CI runs the suite; no test reads `docs/rulesets/`, and the three
`panel-image.test.mjs` assertions that read `docs/REPO_SETUP.md` are unaffected
by these edits.

## Not done here, by design

Nothing is applied to the live repository. No `POST` or `PUT` against rulesets,
no workflow, no automatic invocation. Reading the live rulesets to capture
20390423 and to test the drop guard is the only live interaction.

Base is `ci/release-trains-and-digest-promotion`, not `main`. Draft.

### DIFF
--- a/docs/REPO_SETUP.md
+++ b/docs/REPO_SETUP.md
@@ -619,7 +619,14 @@ stop. This is the only change to that ruleset; the `creation` rule and the ref
 pattern are untouched.
 
 Ruleset **20390423** ("Release tags are immutable") is not touched by this
-effort.
+effort — nothing here applies it and nothing here changes it. Its payload is
+committed anyway, as
+[`rulesets/tag-immutable.json`](rulesets/tag-immutable.json), captured verbatim
+from the live ruleset. The point of `rulesets/` is that a clone can restore the
+repository's protection; leaving one of the five out meant the fifth existed
+only in the web form, which is the thing this section is against. Restoring it
+is a `POST` if it has been deleted and a `PUT` over 20390423 if it has been
+edited — a deliberate act, not a cutover step.
 
 ### Verify it actually binds
 
--- a/docs/REPO_SETUP.md
+++ b/docs/REPO_SETUP.md
@@ -375,6 +375,15 @@ without a pull request.
 - [ ] **The App exists and its id is to hand.** `APP_ID` and `APP_PRIVATE_KEY`
       are set (#108, closed). `APP_ID`'s value is the number the payloads want,
       and §2 has where to read it and what to do when it is missing.
+- [ ] **`preflight-app-id.sh` exits 0.** `export APP_ID=…` and run
+      [`rulesets/preflight-app-id.sh`](rulesets/preflight-app-id.sh) before the
+      first `apply`. `main.json` and `tag-release-cut.json` are full-body
+      `PUT`s: a wrong id does not fail at the API the way `0` does, it replaces
+      the live bypass actor list and returns 200. The script asserts that
+      `APP_ID` is an App actually installed on the owner, that no payload names
+      a different one, and that no `PUT` would drop a bypass actor that is live
+      today — and refuses, naming the mismatch, if any of that does not hold.
+      It reads only; nothing runs it for you.
 - [ ] **All of #109–#113 are merged.** Applying `main.json` is the moment
       "pull requests to `main` only from `beta/*`" stops being advisory, and
       from that moment any foundation pull request still open cannot merge.
@@ -439,7 +448,12 @@ step 9 something is published that cannot be unpublished.
       release lines second, `main` third, tags last. `beta/*` before `main`
       because step 8 cuts a train into a ruleset that should already exist;
       tags last because that payload's only change is adding the bypass actor,
-      and it is cheap to re-apply if the id was wrong.
+      and it is cheap to re-apply if the id was wrong. **Run
+      [`rulesets/preflight-app-id.sh`](rulesets/preflight-app-id.sh) first and
+      apply nothing unless it exits 0** — it is the one check that happens
+      before a write rather than after, and the `main.json` `PUT` is not cheap
+      to re-apply if the id was wrong: it will have replaced the live bypass
+      actor list on the way through.
 - [ ] **7. Prove it binds** — [*Verify it actually
       binds*](#verify-it-actually-binds), below. A throwaway non-`beta/*` pull
       request into `main` is blocked, and a direct push to `main` is refused
@@ -471,6 +485,12 @@ it on the way in:
 
 ```bash
 export APP_ID=…            # the value of the APP_ID secret
+
+# Before the first apply. Exits non-zero, naming the mismatch, if APP_ID is not
+# an App installed on `actana`, if a payload names a different one, or if a PUT
+# would delete a bypass actor that is live today. Reads only.
+docs/rulesets/preflight-app-id.sh   # must exit 0 — if it does not, apply nothing
+
 apply() {                  # apply <payload> [ruleset-id]
   jq --argjson app "$APP_ID" \
      '(.bypass_actors[]? | select(.actor_id == 0)).actor_id = $app' "$1" \
@@ -480,6 +500,18 @@ apply() {                  # apply <payload> [ruleset-id]
     fi
 }
 ```
+
+**The tripwire only catches the unsubstituted case. A wrong id is quieter.**
+Two of these payloads are full-body `PUT`s, so `bypass_actors` is replaced
+wholesale rather than added to: an id that is not the App's does not 422, it
+returns 200 having deleted the live bypass actor — a destructive outcome from a
+config apply that reports success, and one nothing surfaces again until the
+next promotion fails on a ruleset violation. `preflight-app-id.sh` is the
+assertion that happens *before* the write; it resolves the App installed on
+`actana` from the API rather than trusting the number in the shell, refuses on
+a mismatch instead of continuing, and refuses just as hard when it cannot read
+the answer at all (listing installations wants the `admin:org` scope — an
+unreadable answer is the case it exists for, not a pass).
 
 ### 3a. `beta/*` — the trains
 
@@ -554,7 +586,13 @@ goes stale in a way nobody notices until someone pushes to a dead line.
 
 ### 3d. `main`
 
+- [ ] `docs/rulesets/preflight-app-id.sh docs/rulesets/main.json` — exits 0
 - [ ] `apply docs/rulesets/main.json 20390421`
+
+The pre-flight is repeated here rather than left to *Before you apply anything*
+because this is the destructive `PUT`: the payload is the whole ruleset
+afterwards, including its bypass actor list, and 20390421 is the ruleset the
+promotion cannot work without.
 
 What changes from the ruleset that is live today:
 
--- a/docs/REPO_SETUP.md
+++ b/docs/REPO_SETUP.md
@@ -519,9 +519,11 @@ The live-ruleset read has a quieter version of the same problem: GitHub redacts
 `bypass_actors` to `null` for a token that can read the ruleset but is not an
 admin of the repository, and a 200 carrying `null` looks exactly like a
 ruleset with no bypass actors. So the script asserts the shape of what came
-back — the body parses, the `bypass_actors` key is present, and it is an array
-— and refuses by name when it is not. An unreadable answer is the case this
-check exists for, not a pass. **Run it as a repository admin**; if either read
+back — the body parses, the `bypass_actors` key is present, it is an array, and
+every entry in it is an object — and refuses by name when it is not. The
+installations read is asserted the same way, so a body it cannot parse is
+reported as not knowing rather than as `actana` having no App installed. An
+unreadable answer is the case this check exists for, not a pass. **Run it as a repository admin**; if either read
 comes back unreadable, the check refuses and you apply nothing, rather than
 being told a destructive `PUT` is safe by a guard that could not see what the
 `PUT` would delete.
--- a/docs/REPO_SETUP.md
+++ b/docs/REPO_SETUP.md
@@ -510,8 +510,21 @@ next promotion fails on a ruleset violation. `preflight-app-id.sh` is the
 assertion that happens *before* the write; it resolves the App installed on
 `actana` from the API rather than trusting the number in the shell, refuses on
 a mismatch instead of continuing, and refuses just as hard when it cannot read
-the answer at all (listing installations wants the `admin:org` scope — an
-unreadable answer is the case it exists for, not a pass).
+the answer at all — on **both** of the reads it depends on, not just the first.
+
+That second part is worth being precise about, because an exit code is only
+worth as much as the reads behind it. Listing installations wants the
+`admin:org` scope, and a token without it gets a 404 rather than an empty list.
+The live-ruleset read has a quieter version of the same problem: GitHub redacts
+`bypass_actors` to `null` for a token that can read the ruleset but is not an
+admin of the repository, and a 200 carrying `null` looks exactly like a
+ruleset with no bypass actors. So the script asserts the shape of what came
+back — the body parses, the `bypass_actors` key is present, and it is an array
+— and refuses by name when it is not. An unreadable answer is the case this
+check exists for, not a pass. **Run it as a repository admin**; if either read
+comes back unreadable, the check refuses and you apply nothing, rather than
+being told a destructive `PUT` is safe by a guard that could not see what the
+`PUT` would delete.
 
 ### 3a. `beta/*` — the trains
 
@@ -647,14 +660,52 @@ reliably more than one human who can approve.
 
 ### 3e. The tag ruleset
 
+- [ ] **Decide the admin-bypass question below before running this**, then
+  `docs/rulesets/preflight-app-id.sh docs/rulesets/tag-release-cut.json`
 - [ ] `apply docs/rulesets/tag-release-cut.json 20390424`
 
-Ruleset **20390424** restricts *creation* of `refs/tags/v*` to the admin role
-and today has **zero bypass actors**. The App pushes the version tag during
-promotion, so it must be added here too, or the promotion fails at the tag
-after it has already fast-forwarded `main` — the worst place in the sequence to
-stop. This is the only change to that ruleset; the `creation` rule and the ref
-pattern are untouched.
+Ruleset **20390424** restricts *creation* of `refs/tags/v*` to the admin role.
+The App pushes the version tag during promotion, so it must be added here, or
+the promotion fails at the tag after it has already fast-forwarded `main` — the
+worst place in the sequence to stop. The `creation` rule and the ref pattern
+are untouched by the payload.
+
+> [!WARNING]
+> **Open decision — this apply deletes a live bypass actor, and the pre-flight
+> will refuse until someone resolves it.**
+>
+> Read with an admin token, live 20390424 carries a bypass actor that
+> `tag-release-cut.json` does not:
+>
+> ```
+> RepositoryRole:5:always
+> ```
+>
+> Because this is a full-body `PUT`, applying the payload as committed replaces
+> `bypass_actors` wholesale and removes that admin bypass — it does not sit
+> beside it. `preflight-app-id.sh` refuses on exactly this, which is the guard
+> working as intended and not a defect in it:
+>
+> ```
+> ❌ Applying tag-release-cut.json over live ruleset 20390424 would delete a bypass actor.
+>      RepositoryRole:5:always
+> ```
+>
+> This is a decision for whoever owns the repository's protection, and it is
+> deliberately left open here rather than resolved by editing the payload:
+>
+> - **If that admin bypass is meant to survive**, add it to
+>   `docs/rulesets/tag-release-cut.json` alongside the App actor, commit it, and
+>   re-run the pre-flight until it exits 0.
+> - **If it is meant to go**, remove it in the admin console first, as its own
+>   deliberate act with its own audit trail — not as a silent side effect of a
+>   config apply — and then re-run the pre-flight.
+>
+> Do not apply this payload while the pre-flight refuses. Note also that a
+> non-admin token reads this ruleset's `bypass_actors` back as `null`, which is
+> why the pre-flight now asserts the shape of that read: the earlier claim that
+> 20390424 has zero bypass actors was an artifact of reading it without admin
+> rights.
 
 Ruleset **20390423** ("Release tags are immutable") is not touched by this
 effort — nothing here applies it and nothing here changes it. Its payload is
--- a/docs/rulesets/preflight-app-id.sh
+++ b/docs/rulesets/preflight-app-id.sh
@@ -69,17 +69,40 @@ esac
 #    the App's own JWT — so the question is asked of the org. Failing to ask it
 #    is a refusal, not a pass: an unreadable answer is the case this exists for.
 
-installs="$(gh api "orgs/$OWNER/installations" \
-  --jq '.installations[] | [.app_id, .app_slug, .repository_selection] | @tsv' 2>&1)" || die \
+installs_json="$(gh api "orgs/$OWNER/installations" 2>&1)" || die \
   "Could not list the App installations on $OWNER." \
-  "$installs" \
+  "$installs_json" \
   "" \
   "This needs the admin:org scope, which the admin doing the cutover has:" \
   "  gh auth refresh -h github.com -s admin:org" \
   "Do not skip this and apply anyway — it is the assertion, not a formality."
 
+# The same rule the live-ruleset read below follows, applied here for the same
+# reason: a 200 is not an answer. An empty body, a whitespace-only body, an
+# `installations` key that is absent, null or not a list all reduce to "no
+# installations found" — which reads identically to "$OWNER has no App
+# installed" and is not the same claim at all. Assert the shape, so the two
+# cases get two different messages.
+printf '%s' "$installs_json" |
+  jq -e 'type == "object" and has("installations")
+         and (.installations | type == "array" and all(type == "object"))' \
+  >/dev/null 2>&1 || die \
+  "The App installations on $OWNER read back unusably." \
+  "The request succeeded, so this is not a network failure: the body was empty" \
+  "or unparseable, or it carried no \`installations\` list to read." \
+  "This is not an answer of 'no App is installed' — it is not knowing whether" \
+  "one is, which is the case this check exists for." \
+  "" \
+  "Re-run with the admin:org scope and confirm the read returns a list:" \
+  "  gh auth refresh -h github.com -s admin:org"
+
+installs="$(printf '%s' "$installs_json" |
+  jq -r '.installations[] | [.app_id, .app_slug, .repository_selection] | @tsv')"
+
 [ -n "$installs" ] || die "No GitHub App is installed on $OWNER." \
-  "The payloads name one as their only bypass actor, so there is nothing for" \
+  "The read came back readable and empty, so this is an installation list with" \
+  "nothing in it — not a list that could not be read." \
+  "The payloads name an App as their only bypass actor, so there is nothing for" \
   "APP_ID=$APP_ID to be. Install the App first (§2, step 1 of the cutover)."
 
 installed_line="$(printf '%s\n' "$installs" | awk -F'\t' \
@@ -169,13 +192,31 @@ for payload in "${payloads[@]}"; do
   # read the ruleset but is not an admin, so this is the likely case, not the
   # exotic one. Same rule as the installations call above: an answer that
   # cannot be read is a refusal, not a pass.
+  #
+  # The entries are asserted too, not just the array around them. A list
+  # carrying something that is not an object passes `type == "array"`, and the
+  # comparison below then renders it through .actor_type/.actor_id/.bypass_mode
+  # into a string like `null:null:null` — so the script still refuses, but
+  # names a bypass actor that does not exist rather than the malformed read
+  # that invented it. Refuse where the fault is.
+  #
+  # The three keys are asserted for the same reason: an entry that is an object
+  # but carries only `actor_id` renders as `null:4516237:null`, which is the
+  # same invented actor by a different route. Presence, not value — GitHub
+  # returns `"actor_id": null` for a DeployKey bypass actor, and that is a real
+  # actor this must keep comparing rather than refuse on.
   printf '%s' "$live_json" |
-    jq -e 'type == "object" and has("bypass_actors") and (.bypass_actors | type == "array")' \
+    jq -e 'type == "object" and has("bypass_actors")
+           and (.bypass_actors | type == "array" and all(
+                 type == "object" and has("actor_type") and has("actor_id")
+                 and has("bypass_mode")))' \
     >/dev/null 2>&1 || die \
     "Live ruleset $ruleset_id ($name's target) read back without a usable bypass_actors list." \
     "The request succeeded, so this is not a network failure: either the response" \
     "was empty or unparseable, or the token can read the ruleset but not its" \
-    "bypass actors — GitHub redacts them to null for non-admins." \
+    "bypass actors — GitHub redacts them to null for non-admins — or the list" \
+    "came back holding entries that are not bypass actor objects (an object" \
+    "carrying actor_type, actor_id and bypass_mode)." \
     "An answer that cannot be read is not an answer of 'no bypass actors'." \
     "" \
     "Re-run as an admin of $REPO:" \
--- a/docs/rulesets/preflight-app-id.sh
+++ b/docs/rulesets/preflight-app-id.sh
@@ -161,6 +161,27 @@ for payload in "${payloads[@]}"; do
     "$live_json" \
     "Without it there is no way to know what the PUT would replace."
 
+  # A 200 is not an answer. An empty body, a whitespace-only body, a redacted
+  # `"bypass_actors": null` and a response with the key absent all reduce to an
+  # empty actor list below, and an empty list is indistinguishable from "this
+  # ruleset has no bypass actors" — which is the one thing that must not be
+  # assumed here. GitHub redacts bypass_actors to null for a token that can
+  # read the ruleset but is not an admin, so this is the likely case, not the
+  # exotic one. Same rule as the installations call above: an answer that
+  # cannot be read is a refusal, not a pass.
+  printf '%s' "$live_json" |
+    jq -e 'type == "object" and has("bypass_actors") and (.bypass_actors | type == "array")' \
+    >/dev/null 2>&1 || die \
+    "Live ruleset $ruleset_id ($name's target) read back without a usable bypass_actors list." \
+    "The request succeeded, so this is not a network failure: either the response" \
+    "was empty or unparseable, or the token can read the ruleset but not its" \
+    "bypass actors — GitHub redacts them to null for non-admins." \
+    "An answer that cannot be read is not an answer of 'no bypass actors'." \
+    "" \
+    "Re-run as an admin of $REPO:" \
+    "  gh auth refresh -h github.com -s admin:org" \
+    "and confirm the read returns a bypass_actors array before applying anything."
+
   live_actors="$(printf '%s' "$live_json" |
     jq -r '[.bypass_actors[]? | "\(.actor_type):\(.actor_id):\(.bypass_mode)"] | sort | .[]')"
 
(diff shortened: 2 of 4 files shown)


qcentic-dev:
## Acceptance checklist

Updated after the review. Against the two non-blocking findings in #132's
review, the blocking defect found in this PR's review, and the constraints this
follow-up was given.

### 0. The review's blocking defect — the fail-open drop guard

Fixed in 5c48606, using the reviewer's suggested `jq -e` assertion as written.

| # | Criterion | Status |
| --- | --- | --- |
| 0a | The guard asserts the **content** of the response, not just `gh`'s exit code | **Met** — body parses, `bypass_actors` present, and is an array |
| 0b | Empty body → refuses | **Met** — was exit 0 with a green tick |
| 0c | Whitespace-only body → refuses | **Met** — was exit 0 with a green tick |
| 0d | `{"bypass_actors": null}` → refuses | **Met** — was exit 0; the live non-admin case |
| 0e | Key absent → refuses | **Met** — was exit 0 with a green tick |
| 0f | The happy paths still pass | **Met** — populated array and a genuine `[]` both exit 0 |
| 0g | The rule stated for the installations call now applies to this read | **Met** — an unreadable answer is a refusal, not a pass |
| 0h | `REPO_SETUP.md:512-514` no longer overstates the guarantee | **Met** — names both reads, and the `null` redaction as the cause |

Verified against the **live API**, not only a stub: with a non-admin token, a
real `GET .../rulesets/20390421` returns `"bypass_actors": null`, and the
pre-fix script passed that green with `EXIT=0`. The fixed script refuses on the
same live read.

### 1. The destructive full-body PUT

| # | Criterion | Status |
| --- | --- | --- |
| 1 | A pre-flight assertion exists on whatever applies these payloads | **Met** — `docs/rulesets/preflight-app-id.sh`, committed beside the payloads |
| 2 | It verifies the app id in the payload against the app id **actually installed on the repository** | **Met** — resolved from `GET /orgs/{owner}/installations` |
| 3 | The check happens **before any write** | **Met** — `GET`s only; §3 places it ahead of the first `apply` |
| 4 | A mismatch **refuses**, naming it, rather than continuing | **Met** — non-zero exit, ids named, file named, `Applying nothing.` |
| 5 | Referenced from the relevant step of §3's cutover checklist | **Met** — and now repeated at §3e, which is the one that fails today |
| 6 | Nothing runs it automatically | **Met** — no workflow references it; a human step from an admin clone |

### 2. The incomplete restore set

| # | Criterion | Status |
| --- | --- | --- |
| 7 | 20390423 has a committed payload | **Met** — `docs/rulesets/tag-immutable.json` |
| 8 | Captured the same way the others were | **Met** — verified identical to live by the reviewer as well |
| 9 | Follows the established conventions | **Met** |
| 10 | All five live rulesets restorable from the clone | **Met** |

### Constraints

| # | Constraint | Status |
| --- | --- | --- |
| 11 | New worktree, branch off `origin/ci/release-trains-and-digest-promotion` | **Met** — now merged up to `14aa01d` |
| 12 | Nothing applied to the live repository | **Met** — reads only; no `POST`/`PUT` against rulesets |
| 13 | Base is `ci/release-trains-and-digest-promotion`, not `main` | **Met** |
| 14 | Draft; nothing merged; `main` untouched | **Met** |
| 15 | Conventional Commits | **Met** — `fix(ci):`, commitlint clean |
| 16 | `Refs 114` in the body | **Met** — and in the commit message |

### Resolved since the last revision

- **`4516237` is confirmed — by the reviewer's admin token, and the description
  was wrong.** The previous entry here said it "could not be confirmed" because
  `GET .../rulesets/20390421` returned `"bypass_actors": null`. That was a
  permission-redacted read, not a fact about the ruleset. Live 20390421 carries
  `{"actor_id": 4516237, "actor_type": "Integration", "bypass_mode": "always"}`.
  The PR description has been corrected. The check still resolves the id from
  the installations API rather than pinning the constant.
- **`Dependency Audit` is green.** It was red only because this branch predated
  `14aa01d` (the nanoid pin from #134). Merged up; clean merge, nothing
  reverted.

### Open, and deliberately so

- **20390424 needs an operator decision before the cutover.** Applying
  `tag-release-cut.json` over it would delete `RepositoryRole:5:always`, which
  the committed payload does not carry — a full-body `PUT` replaces rather than
  appends. The guard refuses on it, correctly. **No payload was changed**:
  whether that admin bypass is meant to survive is not mine to decide. Written
  up as a warning block in §3e with both ways out. §3e's "This is the only
  change to that ruleset" was false against live and has been struck.
- **The `admin:org` scope is a real prerequisite** of the check, not of the
  apply — and now doubly so, since a non-admin also cannot read `bypass_actors`
  at all. The script says the `gh auth refresh` line and refuses rather than
  passing.
- **Repository coverage is still a note, not an assertion**, when the
  installation is scoped to selected repositories. The reviewer suggested
  carrying it into §3 as a checklist item; not done in this commit — awaiting
  their preference on whether it belongs here or in a follow-up.

docs/rulesets/preflight-app-id.sh
```diff
@@ -0,0 +1,188 @@
...
+    continue
+  fi
+
+  live_json="$(gh api "repos/$REPO/rulesets/$ruleset_id" 2>&1)" || die \
+    "Could not read live ruleset $ruleset_id ($name's target)." \
+    "$live_json" \
+    "Without it there is no way to know what the PUT would replace."
+
+  live_actors="$(printf '%s' "$live_json" |
+    jq -r '[.bypass_actors[]? | "\(.actor_type):\(.actor_id):\(.bypass_mode)"] | sort | .[]')"
```
qcentic-adm, line 227: **Blocking — this exits 0 having verified nothing on four inputs.**

`|| die` catches only `gh`'s exit code. A 200 with an unusable body flows straight through: `jq -r '[.bypass_actors[]?...]'` yields the empty string for an empty body, a whitespace-only body, `"bypass_actors": null`, and a response with the key absent. `live_actors` is then empty, `comm -23` finds nothing dropped, and line 185 prints `✅ … drops no live bypass actor` before the script exits 0.

The `null` case is the one that matters: this PR's description reports getting exactly `"bypass_actors": null` back for 20390421 from a token that could not see them, while the ruleset really carries `4516237` (I confirmed that with an admin token). So the guard would certify the destructive `PUT` as safe precisely when it is blind to what the `PUT` would delete.

Line 70 states the right rule for the installations call — *"Failing to ask it is a refusal, not a pass"*. It just is not applied here, on the read the `PUT` actually turns on.

Suggested fix, between the `gh api` call and the `live_actors` assignment:

```bash
printf '%s' "$live_json" |
  jq -e 'type == "object" and has("bypass_actors") and (.bypass_actors | type == "array")' \
  >/dev/null 2>&1 || die \
  "Live ruleset $ruleset_id ($name's target) read back without a usable bypass_actors list." \
  "The request succeeded, so this is not a network failure: either the response" \
  "was empty or unparseable, or the token can read the ruleset but not its" \
  "bypass actors — GitHub redacts them to null for non-admins." \
  "An answer that cannot be read is not an answer of 'no bypass actors'."
```

I validated it against all seven shapes — it refuses on empty, whitespace, `null`, missing-key and HTML, and passes on both a populated array and a genuinely empty `[]`. Checked against all three live rulesets (20390421, 20390423, 20390424): all still pass, so the real happy path is unaffected, including 20390423's legitimate `[]`.

Optional, same spirit: the HTML case currently fails closed via an unhandled `jq` exit 5 and a raw parse error. The guard above turns that into the script's own message.

docs/REPO_SETUP.md
```diff
@@ -481,6 +501,18 @@ apply() {                  # apply <payload> [ruleset-id]
...
+Two of these payloads are full-body `PUT`s, so `bypass_actors` is replaced
+wholesale rather than added to: an id that is not the App's does not 422, it
+returns 200 having deleted the live bypass actor — a destructive outcome from a
+config apply that reports success, and one nothing surfaces again until the
+next promotion fails on a ruleset violation. `preflight-app-id.sh` is the
+assertion that happens *before* the write; it resolves the App installed on
+`actana` from the API rather than trusting the number in the shell, refuses on
+a mismatch instead of continuing, and refuses just as hard when it cannot read
+the answer at all (listing installations wants the `admin:org` scope — an
+unreadable answer is the case it exists for, not a pass).
```
qcentic-adm: This tells the operator the script "refuses just as hard when it cannot read the answer at all". That is true of the installations call and **not** true of the live-ruleset read that the destructive `PUT` depends on — see the comment on `preflight-app-id.sh`. Once that read fails closed too, this paragraph becomes accurate as written; until then it overstates the guarantee at the point where an operator is deciding whether to trust the exit code.

docs/rulesets/preflight-app-id.sh
```diff
@@ -0,0 +1,188 @@
...
+
+slug="$(printf '%s\n' "$match" | cut -f2)"
+selection="$(printf '%s\n' "$match" | cut -f3)"
+ok "APP_ID=$APP_ID is '$slug', installed on $OWNER."
+
+if [ "$selection" != "all" ]; then
+  note "That installation covers selected repositories, not all of them, and a"
+  note "user token cannot list which. Confirm $REPO is among them on the"
+  note "installation's settings page before applying."
+fi
```
qcentic-adm, line 129: Not blocking, but this is the seam between *installed on the owner* — what the script proves — and *installed on this repository*, which is the hazard as stated. Live `repository_selection` is `selected`, so this note fires on every real run today.

The note is honest about the API limit and the drop guard covers the destructive case directly, so the residual is reasonable. It would be worth carrying into §3 as a checklist item, though: a `note` on stdout in a passing run is easy to read past, and this is the one assertion the script is deliberately not making.

docs/REPO_SETUP.md
```diff
@@ -554,8 +586,14 @@ goes stale in a way nobody notices until someone pushes to a dead line.
 
 ### 3d. `main`
 
+- [ ] `docs/rulesets/preflight-app-id.sh docs/rulesets/main.json` — exits 0
 - [ ] `apply docs/rulesets/main.json 20390421`
 
+The pre-flight is repeated here rather than left to *Before you apply anything*
+because this is the destructive `PUT`: the payload is the whole ruleset
+afterwards, including its bypass actor list, and 20390421 is the ruleset the
+promotion cannot work without.
```
qcentic-adm, line 610: §3d gets the pre-flight repeated inline "because this is the destructive `PUT`". §3e's `apply docs/rulesets/tag-release-cut.json 20390424` (line 650) is also a full-body `PUT`, and it is the one that actually fails today: live 20390424 carries `RepositoryRole:5:always`, which `tag-release-cut.json` does not, so the guard refuses on it.

Two things follow. The repeat is arguably more warranted at §3e than here. And §3e's "This is the only change to that ruleset; the `creation` rule and the ref pattern are untouched" is false against live — the payload also deletes the admin bypass. Whether that deletion is intended is a decision someone needs to make before the cutover, rather than one the operator meets as a refusal partway through it.

qcentic-adm (changes requested):
Reviewed as a safety control against the stated hazard: `main.json` and `tag-release-cut.json` are full-body `PUT`s, so an `actor_id` that is not the installed App's does not 422 — it returns 200 having replaced `bypass_actors` wholesale, deleting the identity `promote.yml` depends on, from an apply that reports success.

The shape of the control is right, and I verified it against the live API rather than against the description. **One blocking defect: the drop guard — the assertion that carries the whole hazard — exits 0 with a green tick on four response shapes where it has verified nothing, and one of them is the exact response this PR's own description reports receiving.**

## Verified independently

**It refuses rather than warns, and it refuses on real data.** I ran the script from a scratch copy against the live API with `APP_ID=4516237` (GETs only, nothing applied):

```
✅ APP_ID=4516237 is 'actana-release-train', installed on actana.
✅ main.json: PUT over 20390421 drops no live bypass actor.
❌ Applying tag-release-cut.json over live ruleset 20390424 would delete a bypass actor.
     RepositoryRole:5:always
EXIT=1
```

That is the guard doing exactly its job on a real, present condition — see the third point under *Also found* below.

**It resolves the id from the API, not from a constant.** `GET /orgs/actana/installations` returns `4516237  actana-release-train  selected`. Deriving the id rather than hardcoding it is the right call — and note that the PR description's claim that 20390421 currently carries no bypass actor is a non-admin artifact. Read with an admin token, live 20390421 is `[{"actor_id":4516237,"actor_type":"Integration","bypass_mode":"always"}]`. The earlier review was right; the id matches; nothing in the script needs changing for it, but the description asserts a live fact that is not true.

**It runs before any write, in document order.** Not merely referenced — placed. §3's first `apply` is line 518 (`apply docs/rulesets/beta.json`); the pre-flight appears at 378 (*Before you apply anything*), 451–456 (cutover step 6), 492 (immediately above the `apply()` definition at 494) and 589 (above `apply docs/rulesets/main.json 20390421` at 590). Every reference precedes every write.

**It does not apply anything.** Two `gh api` calls in the file, both bare GETs — no `--method`, no `-f`/`-F`, no `--input`, no `curl`. Every redirection is `>/dev/null`.

**Nothing in the branch POSTs or PUTs a ruleset automatically.** The only ruleset write verbs anywhere are `REPO_SETUP.md:498-499`, inside the `apply()` shell function a human pastes. No workflow references `preflight-app-id.sh` or `docs/rulesets/`; the `-X POST`/`-X PATCH` in `housekeeping.yml` are Docker Hub description syncs. `.husky/pre-push` and `.husky/commit-msg` do not touch it. `bash -n` clean.

**20390423 is captured byte-faithfully.** I fetched live and diffed rather than trusting the report — identical across `name`, `target`, `enforcement`, `bypass_actors`, `conditions` and `rules`, with `rules` in the same array order. Live returns `"bypass_actors": []`, and the payload writes `[]` rather than omitting the key — matching `release-retired.json`. The omitted keys are exactly the server-generated ones (`id`, `node_id`, `source`, `source_type`, `created_at`, `updated_at`, `current_user_can_bypass`, `_links`), which is what the create/update API takes.

**Conventions match; no second style.** `tag-` prefix consistent with `tag-release-cut.json`; identical key order (`name,target,enforcement,bypass_actors,conditions,rules`) across all six payloads; same 2-space indent, same compact one-line `{ "type": … }` rules, trailing newline present. Documented in the README table, in README prose, and in §3e.

**Five rulesets restorable.** Live today: 20390421, 20390423, 20390424 (confirmed by list and by probing neighbouring ids). Post-cutover set of five — `main`, `beta/*`, `release/*`, tag-release-cut, tag-immutable — now all have a committed payload; `release-retired.json` is a per-retirement template. The gap this closes was real.

## Blocking

The drop guard reads the live ruleset and then trusts whatever came back. It checks `gh`'s **exit code** but never the **content**, so an HTTP 200 that answers nothing is indistinguishable from an answer of "no bypass actors". Driving `gh` with a stub, these four all print `✅ main.json: PUT over 20390421 drops no live bypass actor.` and exit **0**:

| Response body | Result |
| --- | --- |
| empty (zero bytes) | **exit 0 — green tick** |
| whitespace only | **exit 0 — green tick** |
| `{"bypass_actors": null}` | **exit 0 — green tick** |
| `{"id":1,"name":"x"}` (key absent) | **exit 0 — green tick** |
| `<html>502 …</html>` | exit 5 — fails closed, but on a raw `jq` parse error |
| `gh` exits non-zero | exit 1 — fails closed, good message |

Row three is not hypothetical. This PR's own description reports that reading `GET /repos/actana/control/rulesets/20390421` returned `"bypass_actors": null` while the ruleset in fact carries `4516237` — a permission-redacted view. That is precisely the input that makes the guard certify "drops no live bypass actor" about a ruleset whose bypass actors it could not see. The script states the correct rule for the installations call — *"Failing to ask it is a refusal, not a pass: an unreadable answer is the case this exists for"* — and then does not apply that rule to the read that the destructive `PUT` actually depends on. `REPO_SETUP.md:512-514` tells the operator it "refuses just as hard when it cannot read the answer at all," which is true of the installations call and not true here.

Inline comment has a one-line fix, validated against all seven shapes and against all three live rulesets.

## Also found (non-blocking)

- `repository_selection: "selected"` produces a `note`, not a refusal, and the note is honest that a user token cannot enumerate the covered repositories. Given that the id is still the App's, and the drop guard covers the destructive case directly, that is a reasonable residual — but the check verifies *installed on the owner*, not *installed on the repository*, and the note is the only thing standing in the gap. Worth a line in §3 rather than only in the script's stdout.
- §3d repeats the pre-flight inline "because this is the destructive `PUT`" — but §3e's `apply docs/rulesets/tag-release-cut.json 20390424` (line 650) is equally a full-body `PUT` and is empirically the one that fails today. It gets no repeated line.
- **`tag-release-cut.json` genuinely drops `RepositoryRole:5` from live 20390424.** The guard catches it, which is the system working. But it means the documented cutover now halts at *Before you apply anything*, and §3e's prose "This is the only change to that ruleset" is false against live: the payload also removes the admin bypass. Someone has to decide whether that removal is intended before the cutover can proceed. Not this PR's defect — this PR is what surfaced it — but it should not be discovered by the operator mid-cutover.

Fix the fail-open and this is a well-built control: the hazard model is right, it refuses instead of warning, it is placed before the writes rather than merely mentioned, and it reads only.

docs/rulesets/preflight-app-id.sh
qcentic-dev: Fixed in 5c48606, with your suggested `jq -e` assertion applied as written.

The guard now asserts the shape of the read before comparing anything — the
body parses, `bypass_actors` is present, and it is an array — so the
live-ruleset read gets the same rule line 70 states for the installations call.

Re-ran your table. All four fail-open shapes now refuse with the script's own
message and exit 1; the HTML case is no longer a raw `jq` exit 5; both
legitimate passes are unaffected:

| Response body | Before | After |
| --- | --- | --- |
| empty (zero bytes) | ✅ exit 0 | ❌ exit 1 |
| whitespace only | ✅ exit 0 | ❌ exit 1 |
| `{"bypass_actors": null}` | ✅ exit 0 | ❌ exit 1 |
| key absent | ✅ exit 0 | ❌ exit 1 |
| `<html>502 …</html>` | ❌ exit 5, raw parse error | ❌ exit 1, own message |
| `gh` non-zero | ❌ exit 1 | ❌ exit 1 |
| populated array | ✅ exit 0 | ✅ exit 0 |
| genuine `[]` | ✅ exit 0 | ✅ exit 0 |

```
❌ Live ruleset 20390421 (main.json's target) read back without a usable bypass_actors list.
   The request succeeded, so this is not a network failure: either the response
   was empty or unparseable, or the token can read the ruleset but not its
   bypass actors — GitHub redacts them to null for non-admins.
   An answer that cannot be read is not an answer of 'no bypass actors'.
```

Worth adding to your stub evidence: this reproduces against the **live API**,
not only a stub. My token is non-admin, so a real
`GET /repos/actana/control/rulesets/20390421` returns `"bypass_actors": null`.
Running the pre-fix script against that real response:

```
✅ main.json: PUT over 20390421 drops no live bypass actor.
✅ Pre-flight passed.  EXIT=0
```

That is the defect certifying the destructive `PUT` as safe against the actual
ruleset, on the actual API — the ruleset your admin token shows carries
`4516237`. The same live read now refuses.

You were also right about the description: it claimed 20390421 carries no
bypass actor, which was a non-admin artifact and not a fact. Corrected in the
PR body, with the `null` redaction named as the cause rather than left as "may
be the token or may be the state".

docs/REPO_SETUP.md
qcentic-dev: Corrected in 5c48606. The paragraph no longer claims a guarantee the code did
not provide — and now that the live-ruleset read fails closed too, it says
which reads it covers rather than leaving "the answer" ambiguous:

> …refuses just as hard when it cannot read the answer at all — on **both** of
> the reads it depends on, not just the first.

It goes on to name the quieter half explicitly: `admin:org` gets a 404 on the
installations call, while the live-ruleset read comes back `200` with
`bypass_actors` redacted to `null` for a non-admin, which looks exactly like a
ruleset with no bypass actors. Then what the script actually asserts (parses /
key present / is an array), and an instruction to run it as a repository admin.

docs/REPO_SETUP.md
qcentic-dev: Recorded in §3e rather than resolved, since it is a decision about the
repository's protection and not mine to make.

The `apply docs/rulesets/tag-release-cut.json 20390424` line now carries a
warning block above it: the live `RepositoryRole:5:always` actor, the fact that
a full-body `PUT` replaces rather than appends so the admin bypass goes, the
guard's actual refusal output, and the two ways out — add it to the payload if
it should survive, or remove it in the admin console as its own deliberate act
if it should not. Explicitly: do not apply while the pre-flight refuses.

I also struck the false sentence you flagged. "This is the only change to that
ruleset" is gone; the payload's true scope (`creation` rule and ref pattern
untouched) is stated without the claim that nothing else changes.

The checklist item above the apply is the repeated pre-flight you suggested,
pointed at this payload specifically — you were right that §3e warrants it more
than §3d, since this is the one that fails today.

**No payload changed.** `tag-release-cut.json` ships exactly as committed, so
the guard keeps refusing until a human answers the question.

On the `repository_selection` note (your other non-blocking point): not
addressed in this commit. The note still fires on stdout and is still the only
thing standing in the *installed on the owner* vs *installed on this repository*
gap. Happy to carry it into §3 as a checklist item here or in a follow-up —
say which you'd prefer.

qcentic-adm (approved):
Re-reviewed at `5c48606`. **The blocking defect is fixed.** I re-ran my own stub harness rather than the transcript in the description — a fresh `gh` stub of my own, driving the head blob of `preflight-app-id.sh` (sha256 `2645e11f…`, byte-identical to `5c48606:docs/rulesets/preflight-app-id.sh`) from a scratch directory outside the clone. Nothing in the repository was touched, and every live call was a `GET`.

## The four fail-open shapes now fail closed, by name

All four print the script's own `❌ Live ruleset 20390421 (main.json's target) read back without a usable bypass_actors list.` and exit **1**:

| Response body | Before | Now |
| --- | --- | --- |
| empty (zero bytes) | ✅ exit 0 | ❌ **exit 1, named** |
| whitespace only | ✅ exit 0 | ❌ **exit 1, named** |
| `{"bypass_actors": null}` | ✅ exit 0 | ❌ **exit 1, named** |
| key absent | ✅ exit 0 | ❌ **exit 1, named** |
| `<html>502 …</html>` | exit 5, raw `jq` error | ❌ exit 1, the script's own message |
| `gh` exits non-zero | exit 1 | exit 1, `Could not read live ruleset 20390421` |
| populated array | exit 0 | ✅ exit 0 — happy path unaffected |
| genuine `[]` | exit 0 | ✅ exit 0 — 20390423's legitimate case still passes |

The two that already failed closed still do, and the HTML case now refuses with the script's message instead of a raw `jq` exit 5.

## Shapes I did not try last time

| New shape | Result |
| --- | --- |
| top-level JSON **array** where an object is expected | ❌ exit 1, named — `type == "object"` catches it |
| `"bypass_actors": {…}` (object, not array) | ❌ exit 1, named |
| body is JSON `null` | ❌ exit 1, named |
| body is a JSON **string** (`"Resource not accessible…"`) | ❌ exit 1, named |
| two concatenated JSON documents | ❌ exit 1 — actors from both docs are unioned, so the extra one is reported as a drop |
| `"bypass_actors": [null]` | ❌ exit 1, via the drop message |
| `"bypass_actors": ["Integration:4516237:always"]` (array of **strings**) | ❌ exit **5**, raw `jq` error — fails closed, but see below |
| `"bypass_actors": [{"actor_id": 4516237}]` (object missing `actor_type`/`bypass_mode`) | ❌ exit 1, via the drop message, naming `null:4516237:null` |

**No fail-open survives.** Every malformed shape I could construct either refuses by name or exits non-zero. Since §3 defines exit 0 as the only permission to apply, all of these are safe.

## The other read

`GET /orgs/{owner}/installations` (line 72) is the only other read, and its result is also consumed after an exit-code check. I drove it through eleven shapes: `{}`, `installations: null`, an HTML body and a top-level array all refuse with `Could not list the App installations` (the embedded `--jq` makes `gh` exit non-zero); `installations: []`, an empty body and a whitespace body all refuse via the `[ -n "$installs" ]` check at line 81; an entry without `app_id` refuses at the `match` check. Nothing passes green on an unreadable answer, so `REPO_SETUP.md:512-513`'s "on **both** of the reads it depends on" is now accurate as written.

## Non-blocking, and genuinely non-blocking

1. **`preflight-app-id.sh:185-186` — element shape isn't asserted, only the container's.** The new guard proves `bypass_actors` is an array; it does not prove its elements are objects. An array of strings makes line 186 die with a bare `jq: error (at <stdin>:0): Cannot index string with string "bypass_mode"` and exit 5 — no `❌`, no `Applying nothing.` line. An element missing `actor_type`/`bypass_mode` instead refuses with the *drop* message naming `null:4516237:null`, which is a true refusal for a slightly wrong reason. Both fail closed, so neither is the defect I blocked on. If you want the same tidiness you just gave the HTML case, `and (.bypass_actors | all(type == "object"))` in the existing `jq -e` closes it in one clause. Your call — I would not hold the PR for it.

2. **`preflight-app-id.sh:81` reports an unreadable answer as a definite fact.** An empty or whitespace-only installations body refuses with `No GitHub App is installed on actana.` — which is exactly the conflation of "cannot see" with "there is none" that §0 is about, except here it lands on the safe side. Worth a word in the message rather than a code change.

3. **`repository_selection`** — you asked which I'd prefer. Follow-up, not here. This PR is already carrying a fix, a capture and an operator decision; the *installed on the owner* vs *installed on this repository* seam is a documentation item and deserves its own change.

## The rest of the re-check

- **Live API, reads only.** Full-set run with `APP_ID=4516237` through a logging wrapper that recorded every `gh` argv: four calls, all bare `GET`s — `orgs/actana/installations`, then rulesets `20390421`, `20390423`, `20390424`. No `--method`, no `-f`/`-F`, no `--input`, no `curl`. It refuses on the real data, at `tag-release-cut.json` over 20390424 with `RepositoryRole:5:always`, and applies nothing.
- **The payload was not quietly changed to silence the guard.** `docs/rulesets/tag-release-cut.json` is blob `64bc8e3` at `53c5d44`, `df8ccee`, `1b7bbb0` and `5c48606` — identical across the whole branch. `git diff df8ccee 5c48606 -- docs/rulesets/` is the 21-line guard hunk and nothing else.
- **The 20390424 finding is surfaced as a decision, in both places.** The PR body has it under its own heading, and §3e carries a warning block with the live actor, why a full-body `PUT` removes it, the guard's actual output, both ways out, and `Do not apply this payload while the pre-flight refuses` — plus the repeated pre-flight as a checklist item above the `apply` line, which §3e previously lacked. "This is the only change to that ruleset" is gone.
- **`REPO_SETUP.md:512-527`** no longer overstates the guarantee: it names both reads, names the `null` redaction as the quieter failure, states what the script actually asserts, and says to run it as a repository admin.
- **The description's live claim is corrected.** It no longer asserts 20390421 carries no bypass actor; it records the non-admin redaction as the cause. My admin read confirms `[{"actor_id":4516237,"actor_type":"Integration","bypass_mode":"always"}]`. I verified the redacted *shape* refuses (row 3 above); I could not reproduce your non-admin token, so I am taking that half on your word rather than restating it as mine.
- **The base merge reverted nothing.** `1b7bbb0` is a clean two-parent merge whose entire content is the nanoid override — `pnpm-workspace.yaml` plus `3.3.16 → 3.3.18` in the lock. `git diff df8ccee 1b7bbb0 -- docs/` is empty. 20390423's capture still diffs identical to live across `name`/`target`/`enforcement`/`bypass_actors`/`conditions`/`rules`; key order is `name,target,enforcement,bypass_actors,conditions,rules` in all six payloads; the pre-flight still precedes every write in document order (378, 452, 492, 602, 664 against applies at 531, 567, 603, 665). No workflow or hook references the script or `docs/rulesets/`. `bash -n` clean, all six payloads valid JSON.

Approving. The control does what it claims now: it refuses on a mismatch, it refuses when it cannot read the answer, it runs before the writes rather than beside them, and it reads only. Still a draft, and §3e's admin-bypass question is still open — that decision belongs to whoever owns the repository's protection, and the guard will keep refusing until they make it.

PR #133, opened 2026-08-10
docs: describe the train model, the tag ladder, and the rollback runbook

qcentic-dev:
## Summary

The documentation half of #107: describes the train model, the tag ladder and
the rollback runbook **as `release.yml` and `promote.yml` on this branch
actually implement them**, rather than as ADR 0023 planned them. Lands after
the workflow tickets because it documents what they did.

Base is `ci/release-trains-and-digest-promotion`, not `main`. Merged up to
`7cf4951` (#132), so `docs/REPO_SETUP.md` §3 and `docs/rulesets/` are on this
branch as #132 landed them.

### `docs/ci-cd.md`

Rewritten. It still described a trunk — `:edge` on every push to `main`, a `v*`
tag that rebuilt, three entry points, and a "Cutting a release" section already
flagged as being replaced.

- At a glance is **five entry points** plus the reusable `container-image.yml`
- New: the train model, cutting a train, the freeze window, promotion, hotfix
  trains, backports and the supported-lines policy, and **a rollback runbook**
- `<stage>-<arch>` marked as **build scaffolding, not tags to pull** — they are
  in the registry today and were undocumented
- `sha-<short>` now lives in the `-dev` repositories, so whoever bisects a beta
  knows where to pull from
- **The digest guarantee is stated in its narrow form: `beta-x.y.z` → `x.y.z`
  and no further.** PR images are a separate convenience build; the train's
  image is a fresh build of the squash commit. That link in the chain was never
  there, and it is the kind of claim that grows in the retelling

### `docs/REPO_SETUP.md` §2 — the App, and the cutover

**This is the review's blocking item, and the reason `Closes #113` is still
here.** #114 has landed, so the boundary that fenced this branch out of the
file is gone.

`promote.yml` fails five of its jobs with *"APP_ID and APP_PRIVATE_KEY must
both be set. … See docs/REPO_SETUP.md"*, and the document mentioned neither
secret. An admin following that error arrived at a file that could not answer
it — promotion was blocked with no documented way to unblock it.

§2 now:

- Names **`APP_ID`** and **`APP_PRIVATE_KEY`** in the secrets table, alongside
  `DOCKERHUB_CLEANUP_TOKEN`, which was documented in prose but missing from it
- Documents the App behind them: it is `actana-release-train`, app id
  `4516237`, with the `gh api` line to confirm that id rather than trust it
- Says **why `GITHUB_TOKEN` cannot be the push identity** (D39) — its pushes
[...]

PR #136, opened 2026-08-10
ci: release trains and digest promotion — every released image is a digest a human ran

### LINKED ISSUE #107: Release trains and digest promotion: every released image is a digest a human ran

### LINKED ISSUE #110: ci(release): two modes, an explicit latest guard, and no tag trigger

### LINKED ISSUE #111: ci(promote): the fifth entry point — verify, fast-forward, retag, re-cut

### LINKED ISSUE #113: docs: describe the train model, the tag ladder, and the rollback runbook

### LINKED ISSUE #114: chore(ci): rulesets for beta/* and release/*, and the required-check list

### LINKED ISSUE #137: ci: undrafting a pull request never triggers a build — `ready_for_review` is missing from the `pull_request` types, so a required image check stays green on an unbuilt head
`ci.yml` triggers on `pull_request` with the **default** activity types — `opened`,
`synchronize`, `reopened`. `ready_for_review` is not among them, and nothing else in the
workflow listens for it.

`pass` mode exists for the two cases where there is genuinely nothing to build: a draft
pull request, or a documentation-only diff (ADR 0023 D33). It reports `Panel image /
Build + smoke` and `Core image / Build + smoke` as **success** rather than skipping them,
because a skipped required check stays Pending forever.

Those two facts combine badly. A pull request opened as a draft gets `pass`-mode greens on
both image checks in about thirteen seconds. Taking it out of draft fires
`ready_for_review`, which CI does not listen for — so no run starts, no mode is
re-resolved, and the draft-time greens stand. Unless somebody pushes another commit, the
pull request sits there non-draft and mergeable with two **required** image checks green
from a run that built nothing, on a head commit whose images have never been built or
smoke-tested. Nothing about the pull request says so; the check names are the same ones a
real build reports under, deliberately (D33), which is what makes this invisible.

This is live on #136 today: `Panel image / Build + smoke (amd64)` and `Core image / Build
+ smoke (amd64)` are green in 13s and 14s from the draft-time run at 20:52 on 2026-08-10;
the pull request came out of draft at 21:00 and started no second run; there is exactly
one CI run against that head sha. It affects every pull request opened as a draft, not
just that one.

## The fix

Name the types explicitly in `.github/workflows/ci.yml`:

```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
```

Naming `types` at all replaces the defaults rather than extending them, so the three
default entries have to be listed alongside the new one.

**Whoever does this must confirm that the mode resolution still picks `build` on the
resulting run, not merely that a run happens.** The resolver's first branch reads
`github.event.pull_request.draft`, and a `ready_for_review` payload should carry
`draft: false` — but that is the entire point of the change, and a `ready_for_review` run
that still resolved `mode=pass` would be the same bug with an extra run in front of it.
Take a real draft through to ready and observe `Resolve PR image mode` reporting
`mode=build`, and `Build + smoke` taking minutes rather than seconds.

Two things to be aware of while in there, neither of which this issue asks to change:

- The documentation-only branch resolves to `pass` on the same reasoning, so a docs-only
  pull request keeps the "green without a build" shape by design — the diff ships nothing
  into either image.
- `converted_to_draft` is the mirror event and is also absent. Undrafting, pushing a real
  build, then re-drafting leaves the *build*-mode greens standing, which is the harmless
  direction; adding it would only cost a re-resolution to `pass`.

## Where this shipped

Part of #107. The `mode` input and `pass` itself arrived with the PR-image and
mode-resolution machinery in #116; #110's release-side modes grew the same set later. The
`pull_request` trigger block was carried across unchanged in both, so `ready_for_review`
was never in scope for either — this is a gap, not a regression.

qcentic-dev:
ADR 0016's release rebuilt the bytes a tag published, so the image an operator ran had
never been run by anyone. This branch replaces the trunk with a per-version release
train and the release build with a digest retag: a person accepts a `beta-x.y.z` image,
promotion fast-forwards `main` and re-points **that exact digest** at `x.y.z` and
`latest`, and the pipeline refuses when the digest does not carry the approved commit.

Ten commits: eight squash-merged from pull requests into this branch — 116, 117, 130,
131, 132, 133, 134 and 135 — the ADR committed directly, and the #137 fix pushed after
review. Nothing here is applied to the live repository — see **PREREQUISITES**.

## Closes

Closes #107
Closes #110
Closes #111
Closes #113
Closes #114
Closes #137

## What shipped, per ticket

Written from the commits on this branch, not from the issue text — several tickets
grew or shrank in review.

### #107 — the effort, and its ADR

`docs/adr/0023-release-trains-and-digest-promotion.md` (908dfed), 44 clauses: the
branch model, the tag ladder, promotion, hotfix trains, the one backport exception, PR
images, the App identity, and the weekly rebuild becoming a detector. Specification
only; every other commit on this branch implements a numbered clause of it and cites
it. Amended in flight rather than left stale — D38 by #112 (the delete credential has
one guard in force, not two) and D45 added by #113 (the `-dev` sweep clause the sweep's
own modules were citing D33 for).

### #109 — PR images, the train publish, and the train model as a check (#116)

- `container-image.yml` gains `mode` (`build` / `verify` / `pass`), `dev_tags` and
  `push_required`. `verify` asserts a published digest carries a given commit's
  revision label instead of rebuilding an image that already exists; `pass` reports
  success immediately for a draft or a documentation-only diff. Both run under the same
  check names as a build, because a required check whose job is **skipped** stays
  Pending forever.
- `ci.yml` gains `push: beta/**` and loses `push: main`, with its `paths-ignore` block:
  `beta-x.y.z` and `sha-<short>` publish off one build, so the moving tag and the
  immutable handle name the same digest. The train path deliberately takes no path
  filter — a documentation-only merge that skipped the build would leave the revision
  label naming an older commit and the promotion assertion would then fail on a README
  fix. Concurrency stops cancelling on a train ref: two merges 30 seconds apart could
  otherwise leave the moving tag on older bytes, green.
- `Train rules`, a new required check, because GitHub rulesets cannot restrict a pull
  request's *source* branch and "work reaches `main` only through a train" is therefore
  unenforceable as configuration. Five rules: a `main` base needs a `beta/x.y.z` head; a
  train base needs a train `main` can still fast-forward into; a `release/*` base needs
  a supported line; all four `package.json` versions equal the train's version; and the
  promotion is blocked while the train's latest push is red.
- `Conventions` gains the two carve-outs the promotion pull request needs (the
  `beta/x.y.z` branch-name allowlist, and skipping the title lint on a fast-forward
  whose title never becomes a commit). Commit linting stays.
- Every push to a non-draft pull request publishes `pr-<prid><YYYYMM>` to `panel-dev`
  and `core-dev`, with `stage: pr-<number>` so two open pull requests cannot overwrite
  each other's per-arch tags and stitch a manifest from the wrong bytes.
- Review fix folded in: the mode resolver is split into a step that thinks and may fail
  and a step that reads its answer and defaults everything missing, so a dead resolver
  yields `mode=build` and both image checks still run rather than going Pending forever.

### #110 — release.yml's two modes, the latest guard, and no tag trigger (#130)

- `push: tags: v*` is gone, replaced by `workflow_call`. `promote.yml` both invokes
  this workflow and pushes `vx.y.z`; keeping the trigger would fire two release runs
  that could not even block each other. A stray hand-pushed tag now does nothing at all.
- `resolve` reads the mode off the branch graph: a tag reachable from `main` is a
  promotion (retag `beta-x.y.z`, build nothing); a tag on `release/*` is a backport
  (built from that branch with every gate the build path has, Trivy included).
- `scripts/lib/release-latest.mjs` — the `:latest` decision taken **once**, because the
  docker tag list and `gh release --latest` are two renderings of one boolean. Two
  library defaults were doing the wrong thing: `tags="$version latest"` for anything
  without a `-`, and `gh release create` with no `--latest` flag against an API that
  defaults it true. Publishing `1.2.4` after `1.4.0` would have moved `:latest`
  backwards *and* pointed `/releases/latest` at the older version — which is what
  `install.sh` installs and what the update checker compares against.
- `container-image.yml` gains `promote` mode: `docker buildx imagetools create` from
  the published index, refused unless the digest carries the promoted commit as its
  revision label, checked on every architecture. The revision label now comes from `git
  rev-parse HEAD` rather than `github.sha` — they disagree on a backport, and a label
  that is not the truth is the one failure the assertion cannot detect.
- `environment: macos-release` comes off `tarball-macos`; the pause moves to the head of
  `promote.yml`. Exactly one pause exists either way.

### #111 — promote.yml, the fifth entry point (#131)

`workflow_dispatch` with the train name, and the only thing that advances `main`.

- The human pause is the **first** job and everything is downstream of it, so the
  fast-forward is downstream of the reviewer and `main` never contains unapproved code.
- Then the clause the whole design reduces to: resolve `beta-x.y.z` to a digest and
  assert its `org.opencontainers.image.revision` equals the promotion pull request's
  head SHA — both images, both architectures — before anything moves. Compared against
  that SHA, never against the tag's own pointer.
- Then fast-forward `main` and nothing else, push `vx.y.z` as a record, call
  `release.yml` by `workflow_call`. Afterwards: cut `release/x.y` if the line has none,
  delete the train, and either auto-cut the next train or — when another train exists,
  which is the hotfix state read off the ref graph rather than off a label — rebase it
  onto the new `main`, force-push, comment on its open pull requests, and republish.
- Every push uses the GitHub App token, required with no fallback: `GITHUB_TOKEN` pushes
  trigger no workflow, so the tag would appear with no release and the fast-forward
  would never fire `landing.yml` — green everywhere.
- Four defects found and fixed in review are in the squash: `git ls-remote` folded under
  `|| true` turning "origin did not answer" into "no other train exists"; the rebase job
  missing the Actions permissions its own republish fallback uses; `runs_for_tip`
  filtering on `event=push` so the dispatched fallback run could never match its own
  wait loop; and `grep -vx "$TRAIN"` reading the train name as a regex.

### #112 — housekeeping stops republishing releases (#117)

- The weekly rebuild resolved the newest release, rebuilt Core on whatever the base was
  that Monday, and pushed over `<version>` and `latest`. Under digest promotion that
  overwrites a promoted digest with bytes no human approved — while the revision label
  still names the promoted commit, so the promotion assertion keeps passing against
  changed bytes. The immutability claim would have survived about seven days. It becomes
  a **detector**, for both images: base digest drift or a new fixable CRITICAL/HIGH
  opens one issue per released image, deduped on title. Nothing is published.
- `scripts/sweep-dev-tags.mjs` and its pure-decision module: Docker Hub has no tag GC,
  so every push to every pull request leaves a tag forever. Three classes deleted, and
  the hard-coded repository allowlist is load-bearing rather than belt-and-braces —
  exact match, no globs, re-asserted immediately before every delete call, refusing to
  run when empty, and unit-tested with a release repository handed to it by name.
- The description sync moves off `release.yml` and grows to all four repositories;
  `panel-dev` and `core-dev` get real pages, leading with the warning because the
  repositories are public.

### #113 — the docs, rewritten against what the workflows do (#133)

`docs/ci-cd.md` rewritten around the train model, the tag ladder, the freeze window,
hotfix trains, backports, the supported-lines policy and a rollback runbook that says
plainly that `main`, the tag and the release branch are never rewritten. The digest
guarantee is stated in its **narrow** form — `beta-x.y.z` to `x.y.z`, no further —
because the PR image and the train's image are separate builds and that claim grows in
the retelling until somebody relies on it.

Alongside it: `CONTRIBUTING.md` retargeted at the open train (GitHub bases every new PR
on the default branch, so the first one anyone opens is wrong);
`docs/beta-acceptance-checklist.md`, linked from the PR template;
`deploy/docker-compose.yml` moving both services on one `ACTANA_TAG` instead of two
hand-edited literals, plus a `-dev` override; `docs/REPO_SETUP.md` section 2 growing the
App identity, both secrets, the `-dev` repositories and a nine-step ordered cutover;
`docs/adr/README.md` recording two live numbering collisions and the append-never-shift
rule; and the pre-release note on `landing/index.html` and `README.md`, which was still
sending readers to a retired `:edge`. All 342 markdown links resolve.

### #114 — the rulesets, as committed payloads (#132)

`docs/rulesets/` — `beta.json`, `release.json`, `main.json`, `tag-release-cut.json` and
a per-retirement `release-retired.json` template — branch and tag protection as the
exact JSON the API takes, reviewable in a diff and restorable from a clone. Three
choices worth reading twice: strict required-checks is false on `beta/*` (several pull
requests target one train, and strict would serialise the whole E2E and installer
matrix) and true on `main`; the App is the sole bypass actor because it writes every
push the model depends on; and code-owner review is off everywhere because
`@actana/maintainers` is still 404. The status-check contexts are check-run names
character for character — a required context that names nothing is Pending forever, on
every pull request, including the one that would remove it.

`REPO_SETUP.md` section 3 is rebuilt around applying those payloads in an order whose
interlocks are stated first, and records the two verifications only a human with admin
rights can do. **Nothing is applied here.**

`docs/rulesets/tag-immutable.json` captures ruleset 20390423, *Release tags are
immutable*, which existed only in the web form — the directory was four fifths of a
restorable set, and a clone can now put back all five. Beside the payloads,
`docs/rulesets/preflight-app-id.sh`: a read-only assertion run before any ruleset `PUT`,
because the applies are full-body and a *wrong* App id is quieter than a missing one. It
does not 422 — it returns 200 having replaced the live bypass actor list, deleting the
identity `promote.yml` depends on, from a config apply that reports success. The script
asserts that `APP_ID` is set, numeric and not the placeholder; that it names an App
actually installed on the owner, resolved rather than trusted as typed; that no payload
names some other App; and that no `PUT` would drop a bypass actor that is live today.
Every call is a GET, nothing runs it automatically, and an unreadable answer refuses
rather than passes — GitHub redacts `bypass_actors` to `null` for a non-admin token, and
a guard that certifies what it cannot see is the failure it exists to prevent. Section
3e records the one thing it cannot decide: live 20390424 carries
`RepositoryRole:5:always` and `tag-release-cut.json` does not, so the apply would delete
that admin bypass. Written up as an open operator decision rather than resolved — it is
the third item under PREREQUISITES below.

## PREREQUISITES — what an operator must do before the first promotion

None of these are code, and none of them fail loudly if skipped.

1. **Create the `macos-release` environment, WITH at least one required reviewer.**
   It does not exist — `GET /repos/actana/control/environments` returns
   `total_count: 0`. This is not a red build: GitHub **silently auto-creates** a
   referenced environment with **no protection rules** on first use, so the gated job
   would run immediately, the pause would go unattended, and the release would publish
   unreviewed and green. `promote.yml`'s first job declares that environment, so the
   next dispatch is what creates it. Leave deployment branches unrestricted — a branch
   restriction would refuse the ref the gated job runs on. (REPO_SETUP §2.)

2. **Add the GitHub App as a bypass actor on the tag ruleset (20390424),** which
   restricts *creation* on `refs/tags/v*` and carries no Integration actor. The
   App pushes `vx.y.z` during promotion; without the bypass the push is refused —
   **after** `main` has already fast-forwarded, which is the worst place in the sequence
   to stop. (REPO_SETUP §3e, `docs/rulesets/tag-release-cut.json`.)

3. **Decide the `RepositoryRole:5` bypass on 20390424 before the cutover, not during
   it.** Live ruleset 20390424 carries an admin-role bypass actor
   (`RepositoryRole:5:always`) that `tag-release-cut.json` does not carry. The apply is a
   full-body `PUT`, so the payload replaces `bypass_actors` wholesale and that admin
   bypass **goes**. Someone must decide whether it is meant to survive — add it to the
   payload if yes, remove it in the admin console as its own deliberate act if no.
   Recorded as an open operator decision in REPO_SETUP §3e; the pre-flight refuses on it
   rather than letting the apply delete it silently. Both §3e's write-up and the
   pre-flight are on this branch, merged from #135. Note that this cannot be checked
   with a non-admin token: GitHub redacts `bypass_actors` to `null` for a token that can
   read the ruleset but is not a repository admin, and `null` is indistinguishable from
   "none" — which is how the actor was missed in the first place.

Beyond these three, section 3's ordering still applies: `Train rules` must be observed
green before `main.json` is applied, because a required check that has never run leaves
every pull request Pending — including the one that would remove the requirement.

## #134 rides this branch for the Dependency Audit check — Refs #25, not a fix

[#134](https://github.com/actana/control/pull/134) pins `nanoid` to `>=3.3.17` to clear
**GHSA-2v37-7h3g-55p8**. It has **nothing to do with #107**. `Dependency Audit`
(`pnpm audit --prod --audit-level high`) fails on every branch in the repository on
exactly that one advisory, and it is a required check on `main`, so this branch cannot
merge without it. That is the only reason it is here.

**Refs #25, deliberately not a closing keyword.** All seven advisory paths run through a
build toolchain — `@tanstack/react-start` → … → `vite` → `postcss` → `nanoid` — which
reaches the production graph only because `react-start` is declared as a **runtime**
dependency of the Panel. That is #25, and it is completely untouched here. The override
moves the lockfile from 3.3.16 to 3.3.18 and nothing else; the next advisory anywhere in
that toolchain lands in exactly the same place. This is the pin that unblocks merging,
not the fix.

(The same shape applies to the `js-yaml` floor raised to 4.3.1 inside #117, for
CVE-2026-59870, reaching the shipped graph through the same `@tanstack/react-start`.)

## What this supersedes

- **ADR 0016 D30 — the edge fold.** D30 kept `:edge` as a tag while folding
  `images-edge.yml`'s jobs into `ci.yml` as a push-to-`main` branch condition. ADR 0023
  D13 retires `:edge` outright, job **and** tag: under the train model `main` is only
  ever a released version, so `:edge` would be a second name for `:latest`.
  `beta-x.y.z` replaces it one branch earlier in the flow. The argument for folding the
  edge publish ends when there is no edge publish.
- **ADR 0016 D34 — the entry-point count, amended from four to five.** D34 fixed the
  directory at three entry points plus the reusable `container-image.yml`; `landing.yml`
  had already revised that to four. `promote.yml` makes it **five**: `ci.yml`,
  `release.yml`, `promote.yml`, `housekeeping.yml`, `landing.yml`, plus the one reusable
  workflow. `scripts/__tests__/workflows.test.mjs` asserts the set, so this is a check
  rather than a directory listing somebody happens to look at. The reasoning D34 carried
  survives the correction — what makes it work is the count being deliberate, not the
  number it holds.

Also superseded, for completeness: **ADR 0016 D10** (three rebuild mechanisms — the
weekly rebuild-and-republish becomes a detector, D42) and **ADR 0016 D33**'s gating
rationale for the descriptions job, which moves with the job to housekeeping (D43). ADR
0018's single-registry posture is unchanged and load-bearing throughout.

## Bookkeeping on #109 and #112

Both were closed by hand on 2026-08-10, not by a closing keyword, and are therefore
**not** in the Closes list above. Their pull requests — [#116](https://github.com/actana/control/pull/116)
and [#117](https://github.com/actana/control/pull/117) — merged into
`ci/release-trains-and-digest-promotion` rather than into `main`. GitHub fires closing
keywords only for a pull request merged into the **default branch**, so the keywords in
those descriptions never fired and never would have. The work is on this branch: 7565e28
and 176f26c in the log above, summarised under #109 and #112.

The same applies mechanically to #110, #111, #113 and #114, whose pull requests also
merged into this branch — they are listed under Closes because **this** pull request
targets `main`, so its keywords fire.

## #115 stays open, deliberately

[#115](https://github.com/actana/control/issues/115) — *cut `beta/0.1.0` and promote it
as the first release* — is the one ticket in the effort this pull request does **not**
close. It is the first exercise of everything above, and it runs **after** this lands:
the machinery has to be on `main` before a train can be cut against it, and the three
PREREQUISITES have to be satisfied before the promotion can be dispatched. Closing it
here would mark a release as shipped that nobody has run.

## CI: `Train rules` is red, and this is the bootstrap case

Every other check on this pull request passes. `Train rules` fails with its own
message:

> This pull request is based on `main` with head `ci/release-trains-and-digest-promotion`.
> Work reaches `main` only by promoting a train (ADR 0023 D1) — retarget it at the open
> `beta/x.y.z` branch.

That is the check doing exactly what #109 built it to do. **This branch is the commit
that introduces D1, so it cannot satisfy D1** — there is no open `beta/x.y.z` train yet,
because the workflow that cuts one arrives in this diff. No carve-out was written for it
deliberately: an exception in the check that enforces the whole model is a design
decision, not a CI fix, and nothing here is being changed to make it green.

It does not block merging today. `Train rules` is **not** in live ruleset 20390421's
required-check list — it becomes required only when `main.json` is applied, which
REPO_SETUP §3 sequences *after* this merge, for precisely this reason: "the step that
must precede the rulesets is merging this effort to `main` while `main` still accepts a
non-train head." Every check that **is** required on `main` today is green.

`Panel image / Build + smoke` and `Core image / Build + smoke` are green on this pull
request, and **no image was built for this head commit**. Both ran in `pass` mode — 13s
and 14s, in the run started at 20:52 on 2026-08-10 — while this pull request was still a
draft, because a draft has nothing to build (D33), and `pass` reports success rather than
skipping since a skipped required check stays Pending forever. Taking it out of draft at
21:00 did **not** re-trigger them: CI triggers on `pull_request` with the default
activity types, and `ready_for_review` is not among them, so undrafting started no new
run. There is exactly one CI run against head `32439f3`, the draft-time one. Both of
those required image checks are therefore green on a head commit whose images were never
built or smoke-tested.

**That gap is fixed on this branch** (270be3f, pushed after review): `ready_for_review`
is now one of the trigger's named activity types — naming `types` replaces the defaults
rather than extending them, so the three defaults are listed alongside it — pinned by
`workflows.test.mjs` and documented in `docs/ci-cd.md`. Marking this pull request ready
fired the new trigger (run 31462106430, started 05:34 UTC on 2026-08-11 with no push),
`Resolve PR image mode` resolved `mode=build`, and the head you are reading this on is
being built and smoke-tested for real — the image checks on 270be3f are not `pass`-mode
greens.

---


mrfqcentic (commented):
## Review: release trains and digest promotion

**Verdict: Approve-worthy, with operational prerequisites that must not be skipped.** This is a well-reasoned redesign of the release pipeline whose central guarantee — *the published image is a digest a human ran* — is enforced structurally rather than by convention. I reviewed the full diff, checked out the branch, and ran the new unit tests locally: **221/221 pass** (`release-latest`, `dev-tag-sweep`, `workflows`, `image-cve-gate`, `panel-image`). All five workflows parse as valid YAML, all six ruleset payloads are valid JSON, and `preflight-app-id.sh` passes `bash -n`.

### What is strong

- **The digest assertion is designed against the right adversary.** Comparing the OCI revision label against the promotion PR's head SHA — and never against the tag's own pointer — avoids the tautology bug explicitly. Re-asserting the fast-forward condition immediately before the push (not just in `resolve`) is the correct placement of the load-bearing check.
- **The human gate is first, and everything is downstream of it.** Moving the `macos-release` pause to the head of `promote.yml` fixes the old shape where the commit was already on `main` before anyone was asked. The ordering rationale (pause → verify → advance → release) is sound and documented in the file itself.
- **`release-latest.mjs` fixes a real dormant bug the right way.** One pure decision rendered to both surfaces (Docker tags + `gh release --latest`), so they cannot drift. Withholding `latest` from backports *before* any comparison ("structurally incapable") is stronger than a comparison that happens to say no. Tests cover the branch that can never be reached in CI without cutting a real backport.
- **The weekly rebuild becoming a detector was the necessary move.** The old job would have overwritten promoted digests within days while the revision label kept the assertion green — catching that interaction shows the threat model was actually thought through.
- **The dev-tag sweep's safety boundary is the right shape.** Exact-match allowlist, re-asserted before every delete call, refuse-on-empty, unrecognised names kept by default. Given the PAT can technically delete from release repos, this paranoia is proportionate.
- **Honest handling of what could not be verified.** The `RepositoryRole:5` bypass discrepancy on ruleset 20390424 is recorded as an open operator decision rather than silently resolved, and the preflight refuses rather than certifying what a non-admin token cannot see. This is how infra-as-payloads should be done.

### Concerns (non-blocking, but should be tracked)

1. **The three PREREQUISITES fail open, not closed.** The auto-created `macos-release` environment with no protection rules is the dangerous one: a skipped setup step produces an *unattended, green* release. The documentation is prominent, but consider a cheap assertion in `promote.yml`'s `pause` job that the environment has at least one required reviewer (via an admin-token GET) so the failure is loud rather than silent. If that's not feasible with available tokens, a pre-promotion checklist item in `docs/beta-acceptance-checklist.md` would be the fallback.
2. **The new machinery is unexercised end-to-end.** As the PR itself discloses (#137), the green image checks on this head ran in `pass` mode — no image was built for this commit — and `verify`/`promote` modes, the hotfix rebase path, and the tag-push bypass have never run live. That is inherent to the bootstrap and the PR is admirably upfront about it, but it means the first real promotion (#115) is itself a test flight and should be treated as one: dry-run thinking, a rollback plan at hand, and the runbook open.
3. **`pass`-mode green checks can mask real build breakage on docs-adjacent diffs.** The resolver's fail-safe (`mode=build` on a dead resolver) is good, and the mode split (think-step vs read-step) is the right fix for the stuck-Pending problem. Still, the mode decision deserves a loud annotation on every run (it appears to have one) because "green in 13s" and "green because built and smoked" must remain distinguishable at a glance.
4. **Comment density is a maintenance liability.** The per-clause ADR traceability is excellent *now*; the risk is that six months of small edits leave the workflow headers describing a design that has drifted. The `workflows.test.mjs` assertions on the entry-point set are a good start — consider extending that pattern to any invariant currently carried only by comments (e.g. "no `push: main` trigger on ci.yml", "no tag trigger on release.yml" — some of these may already be asserted; if not, they are cheap and high-value).
5. **`nanoid`/`js-yaml` pins are correctly labeled as non-fixes** — just confirming #25 stays open and owned, since the next advisory in that toolchain lands in exactly the same place.

### On the red `Train rules` check

The bootstrap reasoning is sound: the branch that introduces D1 cannot satisfy D1, no carve-out was written, and the check is not yet required so it does not block merge. I agree with not adding an exception — an escape hatch in the check that enforces the whole model would be the first crack in it. The sequencing in REPO_SETUP §3 (merge before applying `main.json`) is correct.

### Summary

Ship it, then do the three prerequisites before the first dispatch, and treat #115 as the acceptance test for everything on this branch. The code I could execute passes; the design documents its own failure modes honestly; the remaining risk is operational, concentrated at first use, and explicitly called out in the PR body — which is where this review ends up agreeing with it.

---
*Automated review: diff inspected in full, branch checked out at `32439f3`, new unit tests executed locally (221 passed), workflow/ruleset payloads syntax-validated.*

qcentic-adm (approved):
## Review: approved — verified against the live repository, not just the diff

I checked out `32439f3` and verified the load-bearing claims independently rather than re-arguing the design. Everything I checked held.

### What I verified for myself

- **Tests.** Full suite: 410/411. The one failure (`scripts/__tests__/core-launcher.test.mjs`) is environmental — it fails identically on `main` on this machine because the launcher test inherits `ACTANA_ROOT`/`AC_APP_PATH` exported by a live Actana session; the file is untouched by this PR. The five PR-touched suites: **221/221 pass**. Lint: 0 errors (53 warnings, all in `packages/` files this PR does not touch). Typecheck clean.
- **CI rollup matches the disclosure exactly.** Every required check green; `Train rules` red with its own bootstrap message; both image checks green via `pass` mode with no image built for this head — disclosed in the body and tracked as #137.
- **All three PREREQUISITES are accurate descriptions of live state, checked with an admin token:**
  1. `GET /repos/actana/control/environments` → `total_count: 0`. `macos-release` does not exist, and GitHub will auto-create it with no protection rules on first use — the fail-open case the body describes is real.
  2. Live ruleset 20390424 carries exactly `RepositoryRole:5:always` as its only bypass actor and `tag-release-cut.json` carries no `RepositoryRole` actor — so the full-body PUT would delete the admin bypass, exactly as §3e says, and `preflight-app-id.sh` refuses on it rather than certifying it.
  3. Live ruleset 20390421's required-check list does **not** contain `Train rules`, so the red check does not block this merge, and its bypass actor is `Integration:4516237` — the App the payloads' `actor_id: 0` placeholder exists to substitute.
- **The ruleset payloads' 16 required-check contexts match the live check-run names character for character** — including the em-dash in `E2E — Panel service seam` and the parentheses in the tarball smokes. A required context that names nothing is Pending forever; this list names real checks.
- **The sweep's safety boundary is tested against the exact incident it exists for** — `dev-tag-sweep.test.mjs` hands `panel` and `core` to `assertSweepable` by name and expects refusal, plus near-miss names (`panel-de`, `PANEL-DEV`, `../panel`).
- **368 relative markdown links across 138 files, 0 broken.**

### On the code itself

I read all five workflows end to end. The central guarantee is enforced at three independent points — `promote.yml`'s `verify` before `main` moves, `verify` mode on the promotion PR, and `promote` mode immediately before the retag — and every one of them compares against the promotion PR's head SHA, never against the tag's own pointer, which is the tautology the design exists to avoid. The fail-safe directions are all correct: a dead mode resolver yields `mode=build` (never `pass`/`verify`), a failing `ls-remote` fails rather than reading as "no other train", an unreadable ruleset read refuses rather than passing, and the `latest` decision is one pure module rendered to both surfaces with the backport refusal re-asserted at three layers. Injection hygiene is consistent — every attacker-reachable value (PR titles, branch names, dispatch inputs) travels through environment variables, never inline interpolation. The `advance` job's ordering (fast-forward, then tag, then `workflow_call`) is what makes `release.yml`'s mode resolution true, and the tag-push failure path documents its own recovery with `main` already moved — which is the prerequisite-2 bite, disclosed, not hidden.

### On the red `Train rules` check

Agree with the bootstrap reasoning and with not carving it out. The branch that introduces D1 cannot satisfy D1; an escape hatch in the check that enforces the whole model would be the first crack in it, and the check is not required until `main.json` is applied — which REPO_SETUP §3 sequences after this merge for exactly this reason.

### Conditions and follow-ups (non-blocking, but treat 1 as blocking-before-dispatch)

1. **The three PREREQUISITES are operator steps with no runtime assertion.** The dangerous one is prerequisite 1: a skipped environment setup produces an *unattended, green* promotion. Do all three before the first dispatch of #115, and decide the `RepositoryRole:5` question deliberately (preflight will refuse until someone does). I'd still like a cheap runtime assertion in the `pause` job that `macos-release` has at least one required reviewer — file it as a follow-up; it is the one skipped step here that fails open instead of closed.
2. **#115 is the acceptance test for everything on this branch** — `verify`/`promote` modes, the tag-push bypass, and the fast-forward have never run live. Run it with the runbook open and the rollback section of docs/ci-cd.md at hand.
3. **#137 (undrafting never re-triggers a build) is real and is tracked.** This PR's own green image checks are the demonstration of it. Don't let #137 rot.

Ship it.

mrfqcentic (approved):
## Approving — new head `270be3f1` verified, including the #137 fix

Following up on my earlier comment review: the approving review by @qcentic-adm and the new commit both check out against what I could verify independently.

### The commit — `270be3f1` (re-run the gate when a PR leaves draft)

- **The Actions semantics are correct.** Default `pull_request` types are `opened`/`synchronize`/`reopened`; `ready_for_review` is not among them, and naming `types` replaces the defaults — so listing all four explicitly is the right shape, not just the right addition.
- **Verified locally at `270be3f1`:** YAML parses to exactly `types: [opened, synchronize, reopened, ready_for_review]`; the new `workflows.test.mjs` pin passes (35/35 in that suite); full suite 411/412 with the single failure being `core-launcher.test.mjs`, which I confirmed is environmental — it passes with `ACTANA_ROOT`/`AC_APP_PATH` unset and fails identically on `main` on any machine running a live Actana session. Untouched by this PR.
- **The fix has already proven itself on this PR.** The synchronize run for this head (`31462106430`) is the first run where the image checks resolved to `build` rather than `pass`: Build + smoke took 1–2 minutes per image (vs 13s in `pass` mode), the multi-arch manifests published, and everything is green except the intentional `Train rules` bootstrap red. This PR no longer carries green checks over a head that was never built — the exact state #137 existed to prevent.
- The earlier failed entries on this head (`Publish the multi-arch manifest`, E2E installer) belong to run `31462019924`, which was **cancelled** 90 seconds in and superseded — not a real failure.

### The approval review by @qcentic-adm

Its independently-checkable claims hold: the test-suite numbers reproduce, the environmental nature of the one failure reproduces, and the CI rollup matches the PR body's disclosure exactly. The three PREREQUISITES were verified against live state with an admin token, which I cannot do — but they are consistent with everything observable without one, and the reviewer's condition 1 (a runtime assertion that `macos-release` has a required reviewer, before #115's first dispatch) remains the right follow-up.

Ship it. #115 is the acceptance test; run it with the rollback runbook open.

Issue #143, opened 2026-08-11
Announce multiConnection on ready; the protocol version does not move (D11)


mrfqcentic:
**Base branch: `feat/multi-connection-core`, not `main`.** Train: **`beta/0.2.0`**.
Suggested branch: **`feat/core-link-multiconnection-capability`**.
Implements **D11** of #140.

## What

Announce the new behaviour as a capability on `ready` — `multiConnection: { version: 1 }` — and
leave `CORE_LINK_PROTOCOL_VERSION` where it is.

Absent means the old single-connection Core. A new client against an old Core sees no capability
and behaves exactly as today; an old client against a new Core works untouched — it never claims,
and never notices it is no longer evicting anybody.

Moving the minor instead would mark **every Core in every fleet** as needs-update to buy a
capability a one-Panel fleet never exercises, turning a Panel deploy into a fleet-wide maintenance
event.

## Done when

- `ready` carries the optional field; its absence is a supported, tested state.
- The Panel's core-link client records the capability per Core and gates claim/subscribe frames on
  it. Against a capability-less Core it must not send them at all — not send-and-tolerate-an-error.
- `CORE_LINK_PROTOCOL_VERSION` is unchanged, and a test pins that it is unchanged, with the reason
  in the test name.
- A Core without the capability still renders as connected and fully usable, not as degraded or
  needs-update.

## The rule this amends

`CONTEXT.md` says the Panel and its Cores are version-locked — *"never a degraded or
feature-detected mode"*. This is feature detection, and so is #129's F9 for the files surface. The
amendment is written by the ADR ticket; this ticket must not restate it differently.

The narrow form: an additive surface may announce itself on `ready` **only where its absence yields
today's behaviour exactly**, not a lesser one. Check this ticket against that sentence — if
`multiConnection` absent ever means something worse than today, the capability is the wrong
mechanism and the version should move.

Refs #140
Refs #129

qcentic-dev:
Delivered in PR #184, squash-merged into `beta/0.2.0` as 57d43c906331a1daac84d81282a76df8bf4cfd4c. The PR base is non-default, so closing keywords did not fire automatically; closing manually. Epic #140 stays open.

Issue #142, opened 2026-08-11
PTY output fans out per connection, by subscription (D2)


mrfqcentic:
**Base branch: `feat/multi-connection-core`, not `main`.** Train: **`beta/0.2.0`**.
Suggested branch: **`feat/core-pty-subscriptions`**.
Implements **D2** of #140.

## What

`PtyCore.setEmitTarget` is a single global callback, so every PTY's `data` and `exit` goes to the
one socket. Replace it with a per-connection subscription registry: a connection receives a PTY's
stream only after asking for it.

Without this, a CLI attached to one session receives every other session's output on the machine,
continuously, for as long as it stays connected. On a busy Core that is the dominant traffic on the
link and none of it is wanted.

## Done when

- New frames: subscribe to a `ptyId`, unsubscribe from it. Both idempotent.
- A connection receives `data`/`exit` for exactly the PTYs it has subscribed to.
- Catch-up needs no new mechanism: `replay { ptyId, sinceSeq }` already exists and `data` frames
  already carry `seq`. Subscribe-then-replay is the same shape the event cursor already uses for
  events — follow it rather than inventing a second one.
- Subscriptions die with the connection; nothing leaks when a client disappears without
  unsubscribing.
- An `exit` reaches every subscriber, and the Core's own bookkeeping (`recordPtyExit`, the event
  log) fires once regardless of subscriber count.

## Traps

- **The Panel currently assumes it gets everything.** It must subscribe for the panes it renders.
  Doing this without the Panel change lands a Panel that shows nothing — pair them in one PR or
  sequence them explicitly.
- **Subscribe-then-replay has a gap.** Between the subscribe taking effect and the replay being
  served, live bytes can arrive. Decide the order and pin it in a test: subscribe first, buffer,
  then replay from the client's `sinceSeq`, then drain — not the reverse.
- **A PTY that exits between subscribe and replay** must still deliver its `exit` to the new
  subscriber, or a client waits forever on a dead process.

Refs #140

qcentic-dev:
Delivered in PR #184, squash-merged into `beta/0.2.0` as 57d43c906331a1daac84d81282a76df8bf4cfd4c. The PR base is non-default, so closing keywords did not fire automatically; closing manually. Epic #140 stays open.

Issue #140, opened 2026-08-11
Multi-connection Core: many Readers, one writer per Session


mrfqcentic:
## Destination

**A Core serves many clients at once. Every Session has many Readers and at most one writer.**

Today a Core accepts exactly one core-link connection: `PtyCoreLinkServer.onConnection`
(`packages/core/src/pty-core-link-server.ts`) closes whatever socket was there before accepting a
new one. That is correct for a system with one client, and #129 gives it three — the Panel, the
CLI, and SDK automations. #129 already lists the consequence under its own known risks:

> One connection per Core: every new client competes with the Panel for it; `attach` holds it for
> as long as it is open. **Flagged, not solved here.**

This effort solves it. Phase 2 of #129 defers `session attach` to last for exactly this reason —
on a single-connection Core, attaching from the CLI evicts the operator's Panel.

The replacement is **not** many writers. A PTY is a byte stream with no notion of who typed; two
clients writing to one harness interleave into nonsense, and an automation driving a session
cannot tolerate a stray keystroke from a Panel tab left open in another window.

## Release

| | |
| --- | --- |
| **Train** | **`beta/0.2.0`** |
| Integration branch | **`feat/multi-connection-core`**, cut from the train |
| Every ticket | branches from the integration branch, PRs back into it |
| Phase-boundary merge | one PR, integration branch → `beta/0.2.0` |

**`beta/0.2.0` carries three independent efforts**, each on its own integration branch with its own
ticket sequence, all merging into the same train:

| Effort | Integration branch |
| --- | --- |
| #139 — the reconnect defect | `fix/panel-overnight-detach` |
| **this effort** | `feat/multi-connection-core` |
| #19 — rename a registered Core | `feat/rename-a-registered-core` |

They do not each get a train. One train, three branches.

**#129 is not on this train.** Putting all three of its phases here would have made `0.2.0` carry
the multi-connection rewrite, the SDK extraction, the first npm publish this repository has ever
done, and the project-files surface — behind one freeze window and one human acceptance run. It
takes the trains after this one instead, one per phase: `beta/0.2.1` (#148), `beta/0.2.2` (#149),
`beta/0.3.0` (#150). Its phase 1 branch is cut from the `main` that *this* train's promotion
produces, so this effort lands first by construction.

**Nothing here ever targets `main`.** `main` advances only by fast-forward promotion, performed by
the workflow after a human runs the beta image and approves (ADR 0023 D1/D5). **No docs-only PRs**
[...]

Issue #139, opened 2026-08-11
A Panel left open overnight comes back detached — which link drops is unestablished


mrfqcentic:
### Description

A Panel left open overnight comes back detached in the morning. This is filed as its own
defect because it is a **blocker on the multi-connection effort** (ADR 0024 / #138), not
merely an annoyance.

Two candidates were ruled out while writing that ADR:

- **Not bearer expiry.** The registration blob's bearer is signed for **365 days**
  (`packages/core/src/actana-setup.ts:46`, `BEARER_DAYS = 365`), so an overnight gap is
  nowhere near it.
- **Not a missing reconnect.** `PtyCoreLinkClient` auto-reconnects with backoff (500ms
  initial, 5s max — `packages/panel/src/server/core-link/client.ts`) and replays by event
  cursor.

That leaves the **panel link** (browser tab → Panel) as the likely culprit: a backgrounded
tab has its WebSocket killed by the browser, and something is not re-establishing on wake.
Unconfirmed — that is what this issue is for.

### Steps to reproduce

1. Open the Panel with at least one live session and leave the tab in the background overnight.
2. In the morning, return to the tab without reloading.

### Expected behavior

The tab reattaches on wake — panel link re-establishes, event cursor replays the gap, and the
terminal shows what happened while it was asleep.

### Actual behavior

The session is detached. Which layer dropped has not been established.

### The question this issue must answer first

**Which link died — the panel link or the core link?**

- Reload the tab and everything is fine ⇒ **panel link** (browser tab → Panel).
- The Core itself shows unreachable, or the PTY is gone ⇒ **core link** (Panel → Core).

Everything else follows from that answer, so establish it before proposing a fix. One night
of logs on both sides should settle it.

### Why this blocks ADR 0024

Today `PtyCoreLinkServer.onConnection` evicts the previous socket, which **hides every
reconnect defect**: a fresh connection always wins, so a failed reconnect is
indistinguishable from a successful one. ADR 0024 D1 removes that eviction, and D7 has a
stale connection holding its Session locks until the heartbeat reaps it
(`HEARTBEAT_TIMEOUT_MS = 45_000`).

So the multi-connection work would be built directly on the subsystem that is already
misbehaving, and the symptom would change from "it reconnected, you didn't notice" to "a
ghost is holding your session". Diagnose before building.

### Which part?

Panel — the web service / UI (pending the panel-link vs core-link answer above)

### Affected version(s)

main @ 45219c3

### Severity

[...]

PR #170, opened 2026-08-11
feat(core-link): a Core accepts many concurrent core-link connections (D1, D12)

### LINKED ISSUE #141: Core accepts many concurrent core-link connections (D1, D12)

## What

connection.

## Done when

it rides this change.

## Traps

Refs #140

qcentic-dev:
**Base branch: `feat/multi-connection-core`, not `main`.** Train: `beta/0.2.0`.
Implements **D1** and **D12** of #140. Closes #141.

## What

`PtyCoreLinkServer` held `activeWs` and one `connection: ActiveConnection`, and closed the
previous socket before accepting a new one. Both fields become a registry keyed by socket —
`Map<WebSocketLike, ActiveConnection>` — and the evict-on-connect branch goes. `ActiveConnection`
was already shaped per connection (its own auth flag, its own cursor, its own poll); it now also
carries its socket, its `lastInboundAt` and its heartbeat, and it is instantiated once per client
instead of once per Core.

Nothing else moves. The Session lock, the subscription model and the `multiConnection` capability
field are #142 / #143 and are **not** implemented here — the ADR records their decisions, the code
does not anticipate them.

## The three traps in the ticket

**The heartbeat compared its captured `ws` against `this.activeWs`.** With a registry that
comparison is not stale, it is *wrong*: every connection except the newest fails it, so the timer
would clear itself and stop pinging — and worse, the identity test that decides "is this
connection still mine to reap" would be answered by "did somebody else connect after me". It now
asks whether it is still its own entry in the registry (`this.connections.get(ws) !== conn`), and
terminates only on its own 45s of silence. Covered by *"reaps only the connection that went
quiet"*: two clients, one pongs, 60s pass, only the silent one is terminated.

**`core.setEmitTarget` was set inside `onConnection`.** With two connections the second overwrote
the first's target and the first went silent while looking healthy — no close, no error, no bytes.
Not left as the interim behaviour: the target is now installed **once for the server** and **fans
out to every registered connection**.

> ⚠️ **Stated as the stopgap it is.** Until D2 (#142) gives each connection a subscription set,
> every connected client receives every PTY's `data` and `exit` on that Core. That is more traffic
> than any one client wants — a CLI attached to one session sees the whole machine — and it is the
> deliberate choice over a per-connection emit target, because `PtyCore.setEmitTarget` holds
> exactly one callback and the alternative silently starves the older client. It is confined to
> `fanOutPtyEvent`, which is the one function D2 replaces.

**`recordPtyExit` must fire once per exit.** It rode the per-connection callback, so N connections
would have appended N `pty:exit` rows for one dead process and replayed the exit N times to every
reconnecting client. It now fires from the single sink. Covered by *"records one pty:exit per exit,
not one per connection"*: three connections, one exit, one row, three `exit` frames.

## Done when

- [x] A second connection does not close the first — both get `ready`, both authenticate
      independently, both keep their own cursor and heartbeat.
- [x] Closing one leaves the other untouched, including its poll timer (it was cancelled through
      shared state before).
- [x] `PtyCore` state is unchanged by a connection coming or going. The emit target is installed on
      the first connection and released only when the last one goes.
- [x] Single-connection behaviour is byte-identical from one client's point of view, except that
      reconnecting no longer evicts anybody.

Ten tests in `packages/core/src/__tests__/core-link-many-connections.test.ts`.

## Docs, landing here rather than separately

- `docs/adr/0024-a-core-serves-many-clients-one-holds-a-sessions-write-lock.md` — the twelve
  decisions, the six rejected options, the Rules amendment. Next free number; nothing renumbered,
  per `docs/adr/README.md`.
- `CONTEXT.md` — **Core client**, **Core alias**, **Session lock**, **Reader**.
- `CONTEXT.md` Rules — `One Core link per Core, multiplexed` → **per Core client**; new rule
  **many Readers, one writer, per Session**; the version-lock rule **narrowed, not deleted**, to
  admit an additive capability announced on `ready` only where its absence yields today's
  behaviour exactly.
- `docs/README.md` — the ADR index row.
- `CONTRIBUTING.md` — the same two rules restated in the short list a reviewer holds a PR to, so it
  does not contradict the glossary it points at. Not in the ticket; included because leaving it
  saying "one core-link per Core" would have made the amended rule wrong in the first place a
  contributor reads it.

## Not in this PR, deliberately

- The Session lock (#142/#143 sequence), the subscription model (D2), the `multiConnection`
  capability on `ready` (D11), the stable client id (D9).
- `packages/panel` — untouched; #19 is in flight there.
- No merge, and nothing targeting `main` or `beta/0.2.0`.

**#139 is a hard blocker for this effort** (#140: "must be diagnosed before any ticket here
merges") — evict-on-connect is what has been hiding reconnect defects, and this PR removes it.
That is a merge gate, not a review gate; flagged here so it is not discovered at merge time.

## Checks

`pnpm typecheck`, `pnpm lint` (0 errors), `pnpm scan:secrets`, the full Core suite (38 files, 688
tests) and the full Panel suite (125 files, 1260 tests) pass. Every commit passes commitlint, not
just the title.

One note on the Panel suite: at the stock 5s per-test timeout it flaked in the dev sandbox — a
different handful of files each run, including ones with no path to the core link
(`token-usage-rollup`). Each passed alone, and the whole suite is green at `--testTimeout=30000`.
Machine speed, not this change; every test that exercises `PtyCoreLinkServer` passes at the stock
timeout.

Refs #140

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -35,8 +35,16 @@ The agentic coding CLI a session drives — `claude-code`, `codex`, `cursor-cli`
 _Avoid_: agent, task agent, AI runtime, tool, model
 
 **Core link**:
-The persistent bidirectional WebSocket connection between a Panel and a Core. Carries PTY streams, task mutations, hook events, and notifications as framed JSON messages. Long-lived; survives Panel sleep; replays missed events on reconnect via a monotonic event cursor.
+The persistent bidirectional WebSocket connection between a Core client and a Core. Carries PTY streams, task mutations, hook events, and notifications as framed JSON messages. Long-lived; survives Panel sleep; replays missed events on reconnect via a monotonic event cursor.
 _Avoid_: tunnel, channel, pipe, API
+
+**Core client**:
+Any program terminating a core link — the Panel, the `actana` CLI, an SDK automation. A Core serves many at once, each with its own event cursor, subscriptions and heartbeat, and none of them privileged: the Panel is a Core client like the others. The unit of write authority is the connection, not the human or the program behind it, so a client that reconnects is a new client until it says otherwise. See ADR 0024.
+_Avoid_: consumer, subscriber, peer, client session
+
+**Core alias**:
+The name one Panel shows for a Core. Panel-local presentation, stored in that Panel's Core registry beside the endpoint and auth material — not a Core fact, and two Panels registering one Core are meant to disagree about it. The Core neither knows nor publishes it.
+_Avoid_: core name, label, hostname, display name
 
 ### Ownership inside a Core
 
@@ -52,6 +60,14 @@ _UI note_: user-facing strings label a Task as "Session" (e.g. "Start a new sess
 **Session** (Core-scoped):
 The live or replayable conversation backing a Task — the PTY stream plus the harness's own session id. A Core runs many sessions; each session drives exactly one Harness. Replayable after Panel reconnect via the Core's event log.
 _Avoid_: conversation, thread
+
+**Session lock** (Core-scoped):
+The exclusive right of one Core client to mutate one Session — its writes, its kill, and every task mutation addressed at it. Held by a core-link connection, never by a person or a program; scoped to a single Session, so different Sessions on one Core may be held by different clients. A Session starts unlocked and its creator gets no privilege; claiming is an explicit gesture; the lock ends on release, on connection drop, or on force takeover, and never on idle. It lives in memory and dies with the Core, exactly like the PTYs it guards. See ADR 0024.
+_Avoid_: session owner, mutex, reservation, ownership
+
+**Reader** (Core-scoped):
+A Core client attached to a Session it does not hold the **Session lock** on. It sees every byte and can write none of them. Many Readers, one writer, per Session — a Reader is the ordinary state of an attached client, not a degraded one, and a Panel tab rendering a Session it does not hold is read-only *before* the first keystroke rather than on the error that answers it.
+_Avoid_: observer, spectator, participant, viewer, watcher
 
 **VM Shell Session** (Core-scoped):
 A free-form interactive shell on the Core's machine — distinct from harness workspaces. Spawned over the same core-link with `shellSession: true`, gated by core-link auth (not project-root validation), rendered in the Panel like a user terminal. The "SSH-equivalent" escape hatch. First-class concept, not a special case of harness PTY.
@@ -125,10 +141,11 @@ _Avoid_: service, daemon config, when you mean the concept — those are impleme
 
 - **Nothing task-shaped lives on the Panel.** Tasks, sessions, terminal logs, hook events, project folders — all on the Core. The Panel stores Cores + one `lastEventId` per Core.
 - **A Project's path is a VM path.** Only the Core can validate it. The Panel never stores or assumes folder paths on a Core.
-- **One Core link per Core, multiplexed.** All PTY streams, task ops, and events for one Core share a single WebSocket. No per-task channels.
+- **One Core link per Core client, multiplexed.** All PTY streams, task ops, and events between one Core client and one Core share a single WebSocket. No per-task channels. A Core serves many such links at once, and a new one never evicts an existing one (ADR 0024 D1).
+- **Many Readers, one writer, per Session.** Any number of Core clients may attach to a Session and watch it; at most one holds its **Session lock** and may mutate it. A PTY has no notion of who typed, so concurrent writers are not a feature to be added later — they are the thing the lock exists to prevent (ADR 0024 D3–D7).
 - **VM Shell Sessions are privileged.** Same auth as the core-link; require an explicit open gesture in the Panel; never auto-spawned.
 - **Singular UI across Cores.** Every Session (Task), Project, notification, status change, and modal renders through the same Panel components regardless of which Core owns the underlying data. Every Core is remote — a Core on the Panel's own host gets no special path. Transport underneath may differ; the surface may not.
 - **Only the Panel dials Cores.** Browsers cannot hold client certificates; every core-link terminates inside the Panel. The Panel UI reaches Cores solely through its panel link.
-- **The Panel and its Cores are version-locked.** The core-link handshake exchanges a protocol version; a mismatched Core renders as "needs update" with the command to run — never a degraded or feature-detected mode.
+- **The Panel and its Cores are version-locked, except for additive capabilities announced on `ready`.** The core-link handshake exchanges a protocol version; a mismatched Core renders as "needs update" with the command to run — never a degraded mode. The one narrow exception: a purely additive surface may announce itself on the `ready` frame, and only where its absence yields today's behaviour *exactly*, not a lesser one. A capability that changes what an existing frame means is not additive and still moves the version (ADR 0024 D11).
 - **Task metadata lives Core-side.** Titles, session icons, pin flags — all owned by the Core. Panel mutations flow over the core-link; two Panels connected to the same Core see identical state.
 - **CLI availability is Core-published state, not Panel probing.** The Panel never inspects a remote machine's PATH directly; it reads the Core's own availability snapshot from the core-link state stream.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,8 +127,13 @@ These are load-bearing, not style preferences. Each is explained in
 - **Singular UI across Cores.** The same components render every Core's data.
   A Core on the Panel's own host gets no special path.
 - **The Panel and its Cores are version-locked.** A protocol mismatch renders
-  as "needs update", never a degraded or feature-detected mode.
-- **One core-link per Core, multiplexed.** No per-task channels.
+  as "needs update", never a degraded mode. The one exception is an additive
+  capability announced on `ready` whose absence yields today's behaviour
+  exactly (ADR 0024 D11).
+- **One core-link per Core client, multiplexed.** No per-task channels. A Core
+  serves many clients at once and a new connection evicts nobody (ADR 0024 D1).
+- **Many Readers, one writer, per Session.** Any client may watch a Session;
+  only the one holding its Session lock may mutate it.
 
 If your change contradicts an ADR in [`docs/adr/`](docs/adr/), say so in the PR
 rather than working around it. Reopening a decision is fine; doing it silently
--- a/docs/README.md
+++ b/docs/README.md
@@ -79,6 +79,7 @@ routing around it.
 | [0021](adr/0021-installing-a-harness-is-a-panel-gesture.md) | Installing a Harness is a Panel gesture |
 | [0022](adr/0022-a-core-owned-project-has-a-panel-side-presentation-row.md) | A Core-owned project has a Panel-side presentation row |
 | [0023](adr/0023-release-trains-and-digest-promotion.md) | Release trains and digest promotion |
+| [0024](adr/0024-a-core-serves-many-clients-one-holds-a-sessions-write-lock.md) | A Core serves many clients, and one connection holds a Session's write lock |
 
 **Two files claim 0018**, as the table shows. It is a pre-existing collision,
 not breakage, and **nothing is renumbered** — every citation in the CI files
--- a/docs/adr/0024-a-core-serves-many-clients-one-holds-a-sessions-write-lock.md
+++ b/docs/adr/0024-a-core-serves-many-clients-one-holds-a-sessions-write-lock.md
@@ -1,0 +1,60 @@
+# A Core serves many clients, and one connection holds a Session's write lock
+
+A Core accepted exactly one core-link connection. `PtyCoreLinkServer.onConnection` closed whatever socket was there before accepting a new one, and that was the correct shape for a system with one client. #129 gives it three — the Panel, the CLI, and SDK automations — and already listed the consequence under its own known risks: every new client competes with the Panel for the one connection, and `attach` holds it for as long as it is open. Flagged there, solved here. #129 phase 2 defers `session attach` to last for exactly this reason: on a single-connection Core, attaching from the CLI evicts the operator's Panel.
+
+The replacement is **not** many writers. A PTY is a byte stream with no notion of who typed; two clients writing to one harness interleave into nonsense, and an automation driving a session cannot tolerate a stray keystroke from a Panel tab left open in another window. So: **a Core serves many clients at once, and every Session has many Readers and at most one writer.**
+
+The twelve decisions below were settled in a design session and are recorded here so no ticket re-litigates them. Each carries a **D** number that its ticket cites. A ticket that needs one changed amends this record rather than settling it in a comment.
+
+## The decisions
+
+**D1 — A Core accepts many concurrent core-link connections.** The evict-on-connect branch goes. Each connection gets its own event cursor, subscription set and heartbeat — already how `ActiveConnection` is shaped; it was simply only ever instantiated once.
+
+**D2 — PTY output fans out per connection, by subscription.** `PtyCore.setEmitTarget` is a single global callback, so every PTY's `data` goes to the one socket. It becomes a registry, and a connection receives a PTY's stream only after asking for it. A CLI attached to one session must not receive every other session's output on the machine, continuously. Catch-up needs nothing new: `replay { ptyId, sinceSeq }` exists and `data` frames already carry `seq`.
+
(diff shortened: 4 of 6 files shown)


qcentic-dev:
## Acceptance checklist — every criterion under #141 "Done when"

**PR #170** — https://github.com/actana/control/pull/170 — draft, base `feat/multi-connection-core`.

### Done when

- [x] **A second core-link connection to a Core does not close the first. Both receive `ready`,
      both authenticate independently, both keep their own event cursor and heartbeat.**
      The evict-on-connect branch is gone; `onConnection` registers into
      `Map<WebSocketLike, ActiveConnection>`. Tests: *"does not close the first connection when a
      second arrives"*, *"keeps answering the first connection's frames after the second
      connects"*, *"authenticates each connection independently"* (one client's bearer is not the
      door held open for the next), *"gives each connection its own event cursor"* (one subscribes
      from 0 and gets the backlog, one subscribes caught-up and gets nothing), *"reaps only the
      connection that went quiet"*.
- [x] **Closing one connection leaves the other untouched — including its poll timer, which today
      is cancelled through shared state.**
      `dropConnection` retires only that connection's poll and heartbeat. Test: *"leaves the other
      connection's poll timer running when one closes"* — the survivor still receives a live event
      pushed after the other closed.
- [x] **`PtyCore` state (PTYs, tasks, the event log) is unchanged by a connection coming or
      going.**
      No `PtyCore` call is made on connect or disconnect other than `setEmitTarget`, which is
      installed once on the first connection and released when the last goes. Test: *"holds the
      Core's emit target until the last connection goes"* — one non-null install for the server's
      lifetime, not one per client. `recordPtyExit` fires once per exit: *"records one pty:exit per
      exit, not one per connection"*.
- [x] **Existing single-connection behaviour is byte-identical from one client's point of view,
      except that reconnecting no longer evicts anybody.**
      Test: *"leaves a single connection behaving exactly as before"* (ready → subscribeAck →
      replay → live event → PTY data → emit target released on close). The seven Panel test files
      that drive a real `PtyCoreLinkServer` pass unchanged.

### Docs land here, not separately

- [x] `docs/adr/0024-a-core-serves-many-clients-one-holds-a-sessions-write-lock.md` — twelve
      decisions, six rejected options, the Rules amendment. Next free number (0023 was the highest);
      appended, nothing renumbered, per `docs/adr/README.md`.
- [x] `CONTEXT.md` terms: **Core client**, **Core alias**, **Session lock**, **Reader** — each with
      its `_Avoid_` line carrying the banned synonyms from #140 (session owner, mutex, reservation,
      observer, spectator, participant, client session).
- [x] `CONTEXT.md` Rules: `One Core link per Core, multiplexed` → **per Core client**; new rule
      **many Readers, one writer, per Session**; version-lock rule **narrowed** to admit an
      additive capability announced on `ready` only where its absence yields today's behaviour
      exactly.
- [x] `docs/README.md` ADR index row.

### Traps called out in the ticket

- [x] **Heartbeat closure vs `this.activeWs`** — replaced with a registry-identity check; reaps
      only on its own silence. Regression test included.
- [x] **`core.setEmitTarget` inside `onConnection`** — not left as the interim behaviour. One sink
      for the server, fanned out to all registered connections, and **said so in the PR** (the
      call-out block under "The three traps"): every client sees every PTY until D2 (#142) lands
      per-connection subscriptions.
- [x] **`recordPtyExit` once per exit** — moved to the single sink; three-connection test asserts
      one event-log row and three `exit` frames.

### Deliberately unchecked

- [ ] **The Session lock, the subscription model, the `multiConnection` capability field** — out of
      scope by the ticket's own words; #142 and #143. Their decisions are *recorded* in ADR 0024
      (that is this ticket's job) but no code anticipates them.
- [ ] **`packages/panel`** — not touched. #19 is in flight there.
- [ ] **Merged / ready for review** — left as a draft, nothing merged, `main` and `beta/0.2.0`
      untouched. Note that #140 makes **#139 a hard blocker for merge**: evict-on-connect is what
      has been hiding reconnect defects, and this PR removes it.

qcentic-adm (approved):
**Approved.** Reviewed against #141 (and #140 for the D-numbers). Every claim in the description was checked against the enforcing line, not the prose.

## The seven claims, verified in the code

1. **Evict-on-connect is gone.** `onConnection` (`packages/core/src/pty-core-link-server.ts:448-455`) now does `const conn = new ActiveConnection(ws); this.connections.set(ws, conn)`. No `removeAllListeners()`/`close()` of a predecessor anywhere in the function. `activeWs` and the singular `connection` field are both gone — `grep -n 'this\.connection\b'` over the file returns nothing, and the registry is `private readonly connections = new Map<WebSocketLike, ActiveConnection>()` (:409).
2. **The heartbeat asks about itself.** `:473` is `if (this.connections.get(ws) !== conn) { conn.stopHeartbeat(); return; }` — registry identity, not `this.activeWs !== ws`. Reaping is gated on `:477-479`, `const idleMs = Date.now() - conn.lastInboundAt` against `HEARTBEAT_TIMEOUT_MS`, and `lastInboundAt` advances only from *this* socket's `pong` and `message` handlers. The timer lives on the entry (`ActiveConnection.heartbeat`), so it is per connection in storage as well as in logic.
3. **One emit target, fanned out, released last.** `ensureEmitTarget()` (:538-542) is idempotent behind `emitTargetInstalled` and installs a single closure calling `fanOutPtyEvent`; `dropConnection` (:519-525) calls `releaseEmitTarget()` only `if (this.connections.size === 0)`. Nothing re-points the target per connection, so no client can be starved by a newer one. The `emitTargets.filter(fn => fn !== null)).toHaveLength(1)` assertion in *"holds the Core's emit target until the last connection goes"* pins that it is installed exactly once.
4. **One `pty:exit` per exit.** `fanOutPtyEvent` (:556-565): the `for (const conn of this.connections.values()) this.send(conn.ws, frame)` loop is followed by `if (event.type === "exit") this.recordPtyExit(event)` — outside the loop, off the single sink. `recordPtyExit` no longer appears on any per-connection path. The three-connections/one-exit/one-row test covers it.
5. **No out-of-scope code.** No `claim`/`release`/`takeover`/lock state, no per-PTY subscription set, no `multiConnection` on `ready` (`ready` is still `{ type: "ready", version: this.protocolVersion }`), no client id. `CORE_LINK_PROTOCOL_VERSION` is untouched. The only match for "lock"/"capability"/"subscription" in the server file is the comment at :532 naming D2/#142 as where it lands. `conn.subscribed` is the pre-existing event-log subscribe, not D2.
6. **`packages/panel` is untouched.** The changed-file set is six files: the server, its new test, `CONTEXT.md`, `CONTRIBUTING.md`, `docs/README.md`, `docs/adr/0024-*.md`.
7. **The docs deliverables are all there.** ADR `0024-a-core-serves-many-clients-one-holds-a-sessions-write-lock.md` is the next free number (0023 was the max; nothing renumbered, and the 0018 collision is left alone per `docs/adr/README.md`), carries all twelve decisions and exactly six rejected options, and follows the house shape — context, numbered decisions, consequences. All four terms land in `CONTEXT.md` with `_Avoid_` lines honouring #140's ban list. All three Rules amendments are in: `One Core link per Core client, multiplexed`, the new *many Readers, one writer, per Session*, and the version-lock rule narrowed rather than deleted. The `docs/README.md` index row is added. `CONTRIBUTING.md` was not in the ticket and is the right call — leaving its short list saying "one core-link per Core" would have contradicted the glossary it points at.

## The "Done when" list

| Criterion | |
| --- | --- |
| Second connection does not close the first; both `ready`, independent auth, own cursor and heartbeat | **Met** — the `authenticated` gate is now `conn.authenticated` at both `onMessage` (:640) and the `subscribe` re-check; `pushLiveEvents(conn)` reads `conn.lastSentEventId`; `handleSubscribe` writes only to `conn` |
| Closing one leaves the other untouched, including its poll timer | **Met** — `dropConnection` stops only that entry's heartbeat and poll and deletes only its own key, guarded on identity so a replaced socket cannot evict its successor |
| `PtyCore` state unchanged by a connection arriving or leaving | **Met** — the only `PtyCore` touch on those paths is `setEmitTarget`, now install-once/release-last |
| Single-connection behaviour byte-identical apart from no eviction | **Met** — verified by *"leaves a single connection behaving exactly as before"*, and by the frame paths being unchanged other than threading `conn` instead of `ws` |

Ten tests, and they test the right property — what a second connection does *not* do. `send()` swallows per-socket throws, so one dead socket in the fan-out loop cannot cost the others their frame or drop the exit row; deleting from the Map mid-iteration (a synchronous close inside `send`) is safe. CI is green across Typecheck, Lint, Unit Tests, Secret Scan and both image builds.

## The fan-out stopgap — acceptable

Until #142, every client on a Core receives every PTY's `data` and `exit`. I think this is the right interim call and the confinement is real, not rhetorical: it is eight lines inside `fanOutPtyEvent`, the exact function D2 replaces, and the alternative — a per-connection emit target — is not a lesser version of the same thing but a silent failure, since `PtyCore.setEmitTarget` holds one callback and the older client would go quiet while still looking healthy. It is declared in three places that a future reader will actually hit: the PR body, the doc comment at :527-537, and the last bullet of the ADR's Consequences. The cost is bounded by how many clients connect, and today that is one.

## Two notes, neither blocking

**`fanOutPtyEvent` is the one path that does not consult `conn.authenticated`** (`packages/core/src/pty-core-link-server.ts:561`). With an `authVerifier` configured, `pushLiveEvents` refuses to push the event tail to an unauthenticated connection — *"a pre-auth subscriber would learn the event timeline without proving identity"* — and the message gate rejects pre-auth RPC. PTY output is strictly more sensitive than the event timeline, and the fan-out sends it to every registered socket. A socket that presents no bearer and sends no frame never trips the message gate, and answering pings keeps `lastInboundAt` fresh so the heartbeat never reaps it.

This is **not a regression** — the old code installed the emit target on whatever socket connected, pre-auth, and evicted the Panel while doing it, so this PR strictly improves the situation. But the refactor is what makes the per-connection fix available for the first time (`skip a connection with `this.authVerifier && !conn.authenticated``, keeping `recordPtyExit` outside the loop). Worth doing here as a one-liner, or worth carrying explicitly into #142 since that ticket rewrites this function anyway.

**`close()` no longer disarms already-open sockets** (`:1045-1052`). It stops every timer and clears the registry, but the `message` handler closes over `conn`, not the registry, so `conn.authenticated` stays true and a frame arriving after `close()` is dispatched in full — including `spawn`/`kill`. Previously `close()` nulled `this.connection`, so `isAuthenticated()` went false and post-close frames were rejected. The window is shutdown-only and needs `authVerifier` configured to differ at all, but closing or terminating the registered sockets in `close()` would remove it.

## Not blocking here, but flagged for the merge gate

#139 is a hard blocker for this effort per #140 (*"must be diagnosed before any ticket here merges"*), and this PR is precisely what stops evict-on-connect masking reconnect defects. The PR body already says this. Also still a draft — nothing here needs it to change, but merge order does.

**Verdict: ready.** Scope is exactly D1 and D12, the three traps in #141 are each answered at the line the ticket pointed at, the docs ride the code as the ticket demanded, and nothing from #142/#143 leaked in.

qcentic-dev:
Both notes from the review are addressed in dd4bf0b. Still a draft — over to you to confirm.

**1. `fanOutPtyEvent` did not consult `conn.authenticated`.** The loop now skips a connection when an `authVerifier` is configured and that connection has not authenticated, exactly as you suggested:

```ts
for (const conn of this.connections.values()) {
  if (this.authVerifier && !conn.authenticated) continue;
  this.send(conn.ws, frame);
}
if (event.type === "exit") {
  this.recordPtyExit(event);
}
```

`recordPtyExit` is untouched and still outside the loop — skipping a client is not allowed to cost the event log its row. Loopback Cores configure no verifier, so every connection is served exactly as before.

New test — *"does not fan PTY output out to a connection that has not authenticated"*: with a verifier configured, one connection authenticates and a second never sends a frame at all (so it never trips the message gate, and it answers pings, so the heartbeat never reaps it — the case you identified). The authenticated socket gets the `data` and the `exit`; the stranger gets neither; the log holds exactly one `pty:exit` row, which pins the append staying outside the loop.

**2. `close()` did not disarm already-open sockets.** It now drops each registered socket's listeners and closes it before clearing the registry:

```ts
for (const conn of [...this.connections.values()]) {
  conn.stopHeartbeat();
  conn.stopPoll();
  try {
    conn.ws.removeAllListeners();
    conn.ws.close();
  } catch { /* already gone — nothing left to disarm */ }
}
```

One deliberate detail beyond the literal suggestion: `ws.close()` alone narrows the window rather than shutting it, because a close handshake still delivers frames already in flight and the `message` handler closes over `conn`, not the registry — so `authenticated` would still be true for those. Removing the listeners is what actually takes the socket out of the conversation; the `close()` that follows is the polite half, telling the client the Core is going away. Snapshotting the values into an array first keeps the iteration honest now that a close can re-enter `dropConnection`.

New test — *"does not dispatch a frame that arrives after close()"*: an authenticated connection sends `spawn` and `kill` after `close()`; neither reaches `PtyCore`, neither appends to the log, and no frame is sent back.

**Verification.** Each new test was run against the un-fixed server and fails there — no vacuous assertions. Full suite: 690 tests / 38 files green in `packages/core` (12 in the many-connections file, up from 10), plus the seven Panel files that drive a real `PtyCoreLinkServer` — 172 tests green — since `close()` now closes sockets and that is the blast radius worth checking. Typecheck clean across all three workspaces; lint 0 errors, and the single warning on `pty-core-link-server.ts` (unused `WebSocketServer` type import, line 19) predates this branch.

**Scope unchanged.** Nothing from #142 or #143 — no lock, no per-connection subscription, no `multiConnection` on `ready`. `packages/panel` still untouched; the changed-file set is the same six, two of them now carrying these fixes. The #139 merge-gate note stands.

qcentic-adm (approved):
Re-reviewed at `dd4bf0b` — the one new commit since the earlier approval. Both non-blocking notes are closed at the enforcing line, and this approval covers the new head.

**1. PTY fan-out now gates on auth.** `pty-core-link-server.ts:573` — `if (this.authVerifier && !conn.authenticated) continue;` inside the loop, and `recordPtyExit` stayed at :575 *outside* it, so a skipped client never costs the log its row. The gate is the same shape as the one `pushLiveEvents` already uses at :633, so the two push paths now agree, and with no verifier configured (loopback) the condition short-circuits and every connection is served exactly as before. `fanOutPtyEvent` is the only place `data`/`exit` frames are constructed, so this is the whole push surface, not one of several.

The new test is not vacuous: without the skip the old loop sends to every registered connection, so `stranger.ofType("data")` would be 1 rather than 0. It also pins the append staying outside the loop — exactly one `pty:exit` row with one connection served and one skipped — which is the assertion that would catch a future "fix" that moved the append inside.

**2. `close()` now disarms registered sockets.** `pty-core-link-server.ts:1073-1087` — per connection: `stopHeartbeat()`, `stopPoll()`, then `removeAllListeners()` and `close()` in a `try`, and only then `this.connections.clear()` / `releaseEmitTarget()` / `this.server.close()`. Order is right: the disarm happens while the registry is still populated. Snapshotting into an array first is the correct call now that a close can re-enter `dropConnection`.

I checked the point where this test could have been vacuous — `FakeWebSocket.removeAllListeners()` genuinely resets `this.listeners` to `{}`, and `receive()` dispatches through `emit()`, which reads that same map. So the post-close `spawn`/`kill` find no `message` listener because the fix removed it, not because the fake never had one. Without the fix the handler closes over `conn` with `authenticated` still true and both frames dispatch — `core.spawned` would be 1. Your reasoning for going past the literal suggestion is right: `ws.close()` alone narrows the window rather than shutting it, since a close handshake still delivers what is in flight.

**3. Scope.** The new commit touches two files, both already in this PR's set: `pty-core-link-server.ts` (+46/-8) and `core-link-many-connections.test.ts` (+67). Full PR remains the same six files; `packages/panel` has zero changed files against the merge base. No Session lock, no per-connection subscription, no `multiConnection` on `ready` — nothing pulled forward from #142 or #143. CI is green on `dd4bf0b` itself (run 31480755841, `head_sha` confirmed): 17 passing, 5 skipped, none failing, including Unit Tests, Typecheck and Lint.

One note on my own verification: this clone has no `node_modules`, so I did not run the suite locally — installing would have written into the repo. The green Unit Tests job on the head SHA is what backs the "tests pass" half; the non-vacuity half I confirmed by reading the fake socket and the pre-fix code paths, which is the part CI could not have told me either way.

Approving. Still marked draft — undraft when you're ready to land it, and the #139 merge-gate note stands.

PR #171, opened 2026-08-11
feat(panel): rename a registered Core from Settings → Cores

### LINKED ISSUE #19: A registered Core cannot be renamed — label is fixed at pairing time
### Description

There is no way to rename a Core once it is registered. A Core's display label is
fixed at pairing time from the alias baked into the registration blob, and nothing
in the Panel can change it afterwards.

For an operator running several Cores this makes the fleet hard to read: a Core that
paired with a blank alias falls back to its endpoint hostname, and a Core whose role
changed keeps whatever name it was given on day one. The only way to change it today
is to remove the Core and re-pair it with a new blob — which is destructive and
disproportionate for a cosmetic field.

### Steps to reproduce

1. Settings → Cores.
2. Pick any registered Core.
3. Look for a rename / edit affordance.

### Expected behavior

The Core's label can be edited in place from Settings → Cores (and ideally from the
fleet view), persisted in the Panel's `cores` table, and reflected everywhere the
alias is shown — fleet rows, session-finished toasts, the new-agent dialog.

### Actual behavior

No rename exists at any layer:

- **UI** — `packages/panel/src/components/views/CoresSettingsPage.tsx` renders each
  Core with only a status badge and a Remove button (`CoresSettingsPage.tsx:283`).
  There is no edit control.
- **API** — `/api/cores` supports `GET` and `POST`, and `/api/cores/:id` supports
  `DELETE` only. No `PATCH`/`PUT` route exists
  (`packages/panel/src/server/api-router.ts:204-209`).
- **Service** — the label is derived once at registration by `labelFor()` and never
  updated. The comment there names this gap explicitly
  (`packages/panel/src/server/services/cores.ts:71-84`):

  > "A Core with no alias in its blob still needs something to be called in the UI.
  > The host is what the operator recognizes; **they can't rename yet**, so getting
  > this wrong means an unidentifiable row."

So this is a never-implemented feature rather than a regression — filing it as a gap
that needs closing.

### Fix direction

The label is Panel-owned state (`packages/panel/src/shared/cores.ts:19`), not
Harness-owned, so this does not need a new core-link frame — it is a Panel-local
write:

1. Add `PATCH /api/cores/:id` accepting `{ label }`, reusing `labelFor()` for
   trim/length normalization (120 chars, non-empty falls back to hostname).
2. Add `renameCore(id, label)` in `packages/panel/src/server/services/cores.ts`,
   bumping `updated_at`.
3. Add an inline edit affordance to `CoresSettingsPage`, and drop the "they can't
   rename yet" comment once it lands.
4. Confirm the alias consumers pick up the change — `useSessionFinishNotifications`
   reads aliases through a ref (`packages/panel/src/lib/use-session-finish-notifications.tsx:455`)
   and `NewAgentDialog` resolves it for error copy (`NewAgentDialog.tsx:81`).

### Which part?

Panel — the web service / UI

### Affected version(s)

v0.49.0 (main @ 087b550)

### Severity

S3 – Minor: non-core feature affected (workaround is remove + re-pair, which is
destructive)

qcentic-dev:
## Summary

A registered Core could not be renamed: the label was derived once at pairing time from the alias
in the registration blob, and nothing in the Panel could change it afterwards. The only workaround
was remove + re-pair — destructive, for a cosmetic field.

This adds the write. `PATCH /api/cores/:id` takes `{ label }`, `renameCore()` puts it through
`labelFor()` — the same normalization registration uses — and bumps `updated_at`, and the row in
Settings → Cores gains an inline edit affordance. The alias stays **Panel-owned presentation**, not
a Core fact (issue #19's settled decision): no new core-link frame, nothing sent to the machine, and
two Panels registered against one Core may keep different names for it.

> **Base is `feat/rename-a-registered-core`, not the train.** This PR is stacked on that branch by
> request rather than opened against `beta/0.2.0` directly, so the `Train rules` check is expected
> to object until the base is retargeted or the stack collapses. Flagging it rather than working
> around it.

## Related issues

Closes #19

## Type of change

## How was this tested?

64 tests across the 5 touched files, all passing:

| Suite | Covers |
| --- | --- |
| `services/__tests__/cores.test.ts` | `renameCore` — trim, 120-char cap, host fallback, `updated_at` bump, unknown id, and that endpoint/cursor/secrets are untouched |
| `server/__tests__/cores-api.test.ts` | `PATCH /api/cores/:id` against a **real Core over mTLS** — renames in place, normalizes, leaves the live link untouched (state and `lastSeenAt` unchanged), 400 on a bodyless call, 401 anonymous, 404 unknown |
| `views/__tests__/cores-settings-rename.test.tsx` *(new)* | the affordance — edit in place, render what was *stored* not what was typed, the 120 cap on the box, cancel, and a refused write keeping the operator's text |
| `views/__tests__/harness-install-from-picker.test.tsx` | `NewHarnessDialog` names the Core by its **current** alias after a rename |
| `lib/__tests__/use-session-finish-notifications.integration.test.ts` | a session-finish toast arriving after a rename carries the new alias |

`pnpm typecheck` clean. `pnpm lint` — 0 errors, and none of the 53 pre-existing warnings are in a
touched file. `pnpm scan:secrets` clean.

Two failures on the full suite are **pre-existing and environmental**, both confirmed by re-running
at the base commit with these changes stashed:

- `scripts/__tests__/core-launcher.test.mjs` — this machine has a real `/opt/actana` install, so the
  launcher resolves it instead of the test's temp root. Fails identically at base.
- Assorted `Test timed out in 5000ms` under full parallel load — a *different* set each run. The base
  commit alone fails 12 tests / 6 files this way; this branch fails fewer. Every file passes when run
  on its own.

## Screenshots / recordings

Not captured — the change is an inline text box and two buttons on an existing row, described above.

## Checklist

- [ ] This PR targets the **open train** (`beta/x.y.z`), not `main` — **deliberately not**; stacked on
  `feat/rename-a-registered-core`, see the note above
- [x] My PR title **and every commit in the branch** follow Conventional Commits, and my branch name
  follows the branching convention — verified with `commitlint --from $(git merge-base …)`, 3 commits,
  0 problems
- [x] I updated documentation where relevant — the `CONTEXT.md` glossary entry for **Core alias** is
  deliberately left to #141, which owns that file; the stale "they can't rename yet" comment in
  `services/cores.ts` is gone, replaced without implying the Core knows its own name
- [x] Dependent changes have been merged (or this PR is marked as draft/stacked) — draft and stacked

### DIFF
--- a/packages/panel/src/components/views/CoresSettingsPage.tsx
+++ b/packages/panel/src/components/views/CoresSettingsPage.tsx
@@ -34,6 +34,8 @@ export function CoresSettingsPage() {
   const [pendingRemoval, setPendingRemoval] = useState<CoreWithDial | null>(null);
   const [removing, setRemoving] = useState(false);
 
+  const [renamingId, setRenamingId] = useState<string | null>(null);
+
   // A poll that lands after unmount must not write to a dead component.
   const mounted = useRef(true);
   useEffect(() => {
@@ -79,6 +81,31 @@ export function CoresSettingsPage() {
       setRegisterError(err instanceof Error ? err.message : "Could not pair that Core.");
     } finally {
       setRegistering(false);
+    }
+  };
+
+  /**
+   * Rename a Core. Panel-local and nothing more: the alias is this Panel's name
+   * for the machine, so the write lands in the registry and the list re-reads
+   * it. The Core is not told, and another Panel paired to it keeps its own name.
+   *
+   * Returns whether it stuck, so the row only leaves edit mode on a write that
+   * landed — a rejected rename keeps the operator's text in front of them.
+   */
+  const handleRename = async (core: CoreWithDial, label: string): Promise<boolean> => {
+    setRenamingId(core.id);
+    try {
+      const { core: renamed } = await api.renameCore(core.id, label);
+      await refresh();
+      // Quote what was stored, not what was typed: an operator who emptied the
+      // box gets the endpoint host back, and the toast should say so.
+      toast.success(`Core renamed to "${renamed.label}".`);
+      return true;
+    } catch (err) {
+      toast.error(err instanceof Error ? err.message : "Failed to rename Core.");
+      return false;
+    } finally {
+      if (mounted.current) setRenamingId(null);
     }
   };
 
@@ -135,6 +162,8 @@ export function CoresSettingsPage() {
                     core={core}
                     onRemove={() => setPendingRemoval(core)}
                     removing={removing && pendingRemoval?.id === core.id}
+                    onRename={(label) => handleRename(core, label)}
+                    renaming={renamingId === core.id}
                   />
                 ))}
               </div>
@@ -222,15 +251,54 @@ function ErrorBox({ children }: { children: React.ReactNode }) {
   );
 }
 
+/**
+ * What the service will fall back to if the alias is saved empty — shown as the
+ * placeholder so clearing the box reads as a choice rather than an accident.
+ * Mirrors `labelFor` on the server, down to surviving an unparseable endpoint.
+ */
+function endpointHost(endpoint: string): string {
+  try {
+    return new URL(endpoint).hostname || endpoint;
+  } catch {
+    return endpoint;
+  }
+}
+
+/**
+ * One Core in the list. The alias is editable in place — it is the only field
+ * here that is the Panel's to change, and the operator's read of the fleet
+ * depends on it. Endpoint and credentials are what the pairing token said they
+ * were; changing those means a fresh token.
+ */
 function CoreRow({
   core,
   onRemove,
   removing,
+  onRename,
+  renaming,
 }: {
   core: CoreWithDial;
   onRemove: () => void;
   removing: boolean;
+  onRename: (label: string) => Promise<boolean>;
+  renaming: boolean;
 }) {
+  // Non-null is the edit mode flag *and* the draft: an empty string is a state
+  // the operator can legitimately be in (it saves as the endpoint host), so
+  // "editing" can't be `draft !== ""`.
+  const [draft, setDraft] = useState<string | null>(null);
+  const editing = draft !== null;
+
+  const commit = async () => {
+    if (draft === null) return;
+    // Nothing to write, and no toast worth showing for it.
+    if (draft.trim() === core.label) {
+      setDraft(null);
+      return;
+    }
+    if (await onRename(draft)) setDraft(null);
+  };
+
   return (
     <div style={{ display: "flex", flexDirection: "column" }}>
     <div
@@ -255,7 +323,39 @@ function CoreRow({
             flexWrap: "wrap",
           }}
         >
-          <span style={{ fontSize: 13, fontWeight: 600, color: "var(--text)" }}>{core.label}</span>
+          {editing ? (
+            <input
+              autoFocus
+              value={draft}
+              onChange={(e) => setDraft(e.target.value)}
+              onKeyDown={(e) => {
+                if (e.key === "Enter") void commit();
+                else if (e.key === "Escape") setDraft(null);
+              }}
+              disabled={renaming}
+              aria-label={`Name for Core ${core.label}`}
+              // The same 120 the service caps at, so the box can't accept
+              // characters the registry would silently drop.
+              maxLength={120}
+              placeholder={endpointHost(core.endpoint)}
+              style={{
+                flex: 1,
+                minWidth: 120,
+                background: "var(--surface-1)",
+                border: "1px solid var(--accent)",
+                borderRadius: 5,
+                outline: 0,
+                color: "var(--text)",
+                padding: "3px 8px",
+                fontSize: 13,
+                fontWeight: 600,
+              }}
+            />
+          ) : (
+            <span style={{ fontSize: 13, fontWeight: 600, color: "var(--text)" }}>
+              {core.label}
+            </span>
+          )}
           <DialBadge dial={core.dial} />
         </div>
         <div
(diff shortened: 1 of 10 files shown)


qcentic-dev:
## Issue #19 — expectation checklist

Every line in the issue's **Expected behavior** and **Fix direction** sections, checked or
explicitly unchecked with the reason.

### Expected behavior

- [x] **The Core's label can be edited in place from Settings → Cores** — a pencil button on each row
  turns the label into a text box; Enter or **Save** writes, Escape or **Cancel** backs out. A refused
  write keeps the box open holding what was typed.
  (`CoresSettingsPage.tsx`, `cores-settings-rename.test.tsx`)
- [ ] **…and ideally from the fleet view** — **not done.** The issue marks this "ideally", and the
  Fix direction names only `CoresSettingsPage`; adding a second write surface is scope the issue did
  not ask for. Fleet rows do *reflect* a rename (next-but-one item) — they just do not offer the edit.
  Worth a follow-up issue if the operator wants it there too.
- [x] **Persisted in the Panel's `cores` table** — `UPDATE cores SET label = ?, updated_at = ?`, and
  `getCore` reads it back. `createdAt`, `endpoint`, `last_event_id` and the sealed secrets are proven
  untouched by the rename. (`services/cores.ts`, `cores.test.ts`)
- [x] **Reflected in fleet rows** — `FleetView` renders `core.label` from `useCores()` for both the
  per-Core section headings and the "Open a Core…" picker, so the registry poll carries a rename
  through with no extra plumbing. (`FleetView.tsx:179,412` → `use-fleet.ts`)
- [x] **Reflected in session-finished toasts** — a finish arriving after a rename is titled with the
  new alias, and the stored notification row carries it too.
  (`use-session-finish-notifications.integration.test.ts`)
- [x] **Reflected in the new-agent dialog** — that component is `NewHarnessDialog.tsx` today; the
  issue's `NewAgentDialog.tsx:81` reference predates the rebrand rename. Its `coreLabel` resolves per
  render, and the Core named in "Install `<cli>` on …" is asserted to follow a rename.
  (`NewHarnessDialog.tsx:86-92`, `harness-install-from-picker.test.tsx`)

### Fix direction

- [x] **1. `PATCH /api/cores/:id` accepting `{ label }`** — routed in `api-router.ts` beside the
  existing `DELETE`, behind the same Operator-session gate. 400 on a body with no label, 401
  anonymous, 404 on an unknown id. PATCH takes the label and nothing else: endpoint and credentials
  are what the registration blob said they were.
- [x] **1a. reusing `labelFor()` for trim/length normalization (120 chars)** — the *same* function
  registration uses, not a second copy. Verified at both the service and the HTTP layer.
- [x] **1b. non-empty falls back to hostname** — an emptied or all-whitespace label stores the
  endpoint host. The UI shows that host as the box's placeholder, so clearing it reads as a choice
  rather than an accident, and the success toast quotes what was *stored* rather than what was typed.
- [x] **2. `renameCore(id, label)` in `services/cores.ts`, bumping `updated_at`** — returns the updated
  `Core`, or `null` for an unknown id (writing nothing). The bump is asserted against an aged row so
  it cannot pass on a same-millisecond coincidence.
- [x] **3. Inline edit affordance in `CoresSettingsPage`** — see the first item above.
- [x] **3a. drop the "they can't rename yet" comment** — gone. Its replacement documents `labelFor` as
  the one normalization both ways in share, and deliberately does **not** imply the Core knows its own
  name, per the issue comment's closing note.
- [x] **4. Confirm the alias consumers pick up the change** — confirmed with tests rather than by
  reading the code. `useSessionFinishNotifications` keys its subscription on *which* Cores exist, so a
  rename intentionally does not re-run it — the alias map is read through a ref refreshed on render,
  and a post-rename finish is now proven to use the new name. `NewHarnessDialog` resolves the alias
  per render. Both suites previously stubbed `useCores` with a frozen array that no rename could move;
  they read a mutable one now.

### Scope boundaries observed

- [x] **No `packages/core` changes** — `git diff --stat -- packages/core` is empty; #141 owns that
  package in parallel.
- [x] **No new core-link frame** — the alias is Panel-local state. Nothing crosses the wire, and the
  live link is asserted to survive a rename with its state and `lastSeenAt` unchanged.
- [x] **`CONTEXT.md` left alone** — the **Core alias** glossary entry belongs to #141 per this issue's
  design comment; writing it here would collide.
- [x] **Nothing merged, `main` and `beta/0.2.0` untouched, PR left as a draft.**

---

**PR #171** — https://github.com/actana/control/pull/171
Draft · base `feat/rename-a-registered-core` · head `feat/panel-core-rename` · 3 commits

qcentic-adm (approved):
Reviewed against issue #19 (description, expected-behavior, fix-direction) and its "the label is a nickname, not a name" comment. I read the enforcing lines rather than the PR description; every claim below is anchored to a file and line on `feat/panel-core-rename`.

**Verdict: approve.** All five fix-direction steps and the expected-behavior list are met, the "ideally" fleet-view rename is the only deferral, and the blast radius is exactly what the settled decision permits.

## What I verified

**The write is gated exactly like `DELETE`.** `api-router.ts:217` puts `PATCH` inside the same `CORE_PATH` block as `DELETE`, and both are only reachable through `protectedDispatch` (`api-router.ts:171`), whose gate runs at `api-router.ts:156-157` before dispatch. `ANONYMOUS_ROUTES` (`api-router.ts:95-99`) has no cores entry, so `requireOperatorSession` applies and returns 401 (`panel-auth.ts:123`). 400 on a missing label comes from `renameBody = z.object({ label: z.string() })` (`cores.controller.ts:23`) through `parseJsonBody` (`_helpers.ts:60-63`); 404 from `cores.controller.ts:61`. Endpoint and credentials are unreachable through it twice over: zod strips unknown keys, and the statement at `services/cores.ts:178` writes only `label` and `updated_at`.

**Normalization is the same function, not a second copy.** `services/cores.ts:179` calls `labelFor(label, existing.endpoint)` — the same `labelFor` at `services/cores.ts:80` that registration calls at `services/cores.ts:148`. Trim, `slice(0, 120)`, hostname fallback all live in one place.

**`renameCore` bumps `updated_at` and is a no-op on an unknown id.** `services/cores.ts:175-176` returns `null` before touching the db; `:178-179` sets `updated_at = Date.now()`. Sealed secrets, `endpoint`, `created_at` and `last_event_id` are not in the `SET` list, and `services/__tests__/cores.test.ts:194-204` asserts all four survive a rename.

**All three alias consumers follow a rename.** Fleet rows read `core.label` off the live `useCores` poll at `FleetView.tsx:179` and `:412`. `use-session-finish-notifications.tsx:455` reads the alias through `coreAliasByIdRef`, refreshed every render at `:158`, while the subscription stays keyed on `coreIds` at `:464` — so a post-rename finish is titled with the new name, pinned by the new test at `lib/__tests__/use-session-finish-notifications.integration.test.ts:160-175`. `NewHarnessDialog.tsx:91-92` resolves the alias per render and `:173` uses it for the "not on PATH" copy; the rename is pinned at `harness-install-from-picker.test.tsx:140-170`. (The issue's `NewAgentDialog.tsx:81` reference is stale — this is that component under its current name.)

**The stale comment is gone.** `services/cores.ts:72-79` replaces it, and the replacement says "normalize an alias for the registry" without implying the Core knows its own name — which is what the issue comment asked for.

**Nothing crosses the wire.** `packages/core`, every protocol/frame/panel-link file, and `CONTEXT.md` are all zero-diff against the base. The **Core alias** glossary entry is correctly left to #141.

CI is green on all 19 required checks, including `Train rules` — the base branch tip is identical to `beta/0.2.0`, so the objection the PR body anticipated did not materialize.

## Non-blocking notes

1. `packages/panel/src/server/__tests__/api-auth.test.ts:51` — `PROTECTED_ROUTES` lists GET/POST/DELETE for Cores but not the new `{ method: "PATCH", pathname: "/api/cores/abc" }`. The behaviour is covered by the 401 assertion in `cores-api.test.ts:429-435`, so this is not a hole; one line here would keep the gate table in step with the router.

2. `packages/panel/src/server/__tests__/cores-api.test.ts:429` — nothing asserts the extra-key case (`PATCH { label, endpoint }` leaving the endpoint alone). Zod's strip and the narrow `UPDATE` both enforce it, so this is a pin for a property worth pinning, not a fix.

3. `packages/panel/src/components/views/CoresSettingsPage.tsx:259` — `endpointHost` re-derives what `labelFor` derives, client-side. The comment owns this, it feeds only the placeholder, and the row still renders what the server stored, so drift would cost a wrong hint and nothing else. Fine as-is.

4. `packages/panel/src/lib/use-fleet.ts:124` and `:135` stamp `coreLabel` into each fan-out result, and the fan-out re-runs on `coreSignature` (id + dial state, `use-fleet.ts:110-113`) — not on the label. That copy is therefore stale until the next fan-out. Harmless today: nothing renders it (`offlineCores` has no consumer, and `coreLabel` only feeds `byLabelAsc` in `fleet-merge.ts:137`), which is why fleet rows are correct. Worth knowing before anything starts rendering `row.coreLabel`.

5. Rename from the fleet view is deferred. The issue marks it "ideally", so deferring is the right call for an S3 UI gap — but the PR says `Closes #19` without saying so. Suggest a line in the PR body or a follow-up issue so the deferral survives the close.

6. Still a draft, stacked on `feat/rename-a-registered-core` (currently identical to the train tip). Retargeting to `beta/0.2.0` is a no-op merge-wise whenever you want it.

Tests were read, not executed — CI's `Unit Tests` job is the better evidence and it passes.

Issue #129, opened 2026-08-10
SDK, CLI, and project files: one Core-communication layer with three products on top


qcentic-dev:
## Destination

**One Core-communication layer, written once, with three products on top of it.**

Today two programs speak core-link to a Core and share no code: the Panel's client (`packages/panel/src/server/core-link/client.ts`, 985 lines) and a prototype CLI living outside the repo (1,869 lines) whose protocol types are a **hand-copied 90-line mirror** of `packages/shared/src/core-link-frames.ts`. The mirror drifts the moment `CORE_LINK_PROTOCOL_VERSION` moves, and nothing catches it at compile time.

| Product | Distribution | Consumes |
| --- | --- | --- |
| `@actana/sdk` | npm | — (it *is* the layer) |
| `@actana/cli` | npm, global, binary `actana` | the SDK |
| Panel | Docker image (ADR 0018) | the SDK |

`@actana/core` imports the protocol types from the SDK package — deliberate: the SDK owns the protocol definition; the Core is the server answering those frames.

## The decisions are locked

Settled in a design session; recorded here so no ticket re-litigates them. Each phase-1/2 decision carries a **D** number, each phase-3 decision an **F** number.

### SDK and CLI (phases 1–2)

- **D1 — The SDK is a Core client, not a Panel client.** One socket, one Core, mTLS + bearer from a registration blob. The Panel's `PtyCoreLinkClient` becomes the SDK by extraction, not rewrite.
- **D2 — Layered, not thin.** Two levels in one package: **transport** (connect, auth, `reqId` correlation, typed frame methods, the three unsolicited streams) and **session** (start, wait for harness ready, deliver prompt, read result).
- **D3 — Prompt delivery becomes a Core responsibility.** The flat 450 ms `initialInput` timer (`packages/core/src/pty-manager.ts:619`) is replaced by: wait for harness output to go quiet, answer any blocking dialog, then write the prompt + `\r`. The harness folder-trust dialog defaults to **"No, exit"** — an early Enter kills the session. Harness differences live in the Core (per `CONTEXT.md`), so every client behaves identically. **Needs an ADR.**
- **D4 — The protocol moves into the SDK package.** `core-link-frames.ts` leaves `@actana/shared` (which stays private) and becomes part of `@actana/sdk`. One published package, one definition, no second copy.
- **D5 — The SDK handles one Core; the Panel remains the fleet.** The Core accepts one connection at a time (`pty-core-link-server.ts:427`); no `Fleet` type in the SDK.
[...]

PR #173, opened 2026-08-11
feat(core-link): announce multiConnection on ready; the protocol version does not move (D11)

qcentic-dev:
Implements **D11** of #140. Base is `feat/multi-connection-core`, not `main`.

## What

`ready` grows an optional `multiConnection: { version: 1 }`, and
`CORE_LINK_PROTOCOL_VERSION` stays at `0.15.0`.

Absent means the single-connection Core. A new Panel against such a Core sees no
capability and behaves exactly as it does today; an old Panel against a new Core is
untouched — it never reads the field, and never notices it is no longer evicting anybody.

## Checked against the rule

ADR 0024's narrowed rule (already in `CONTEXT.md`, landed with #170) permits an additive
surface on `ready` only where its absence yields today's behaviour exactly. The check for
this ticket:

- **Absent → the Panel sends nothing new.** `MULTI_CONNECTION_ONLY_FRAME_TYPES` holds
  `ptySubscribe` / `ptyUnsubscribe`, and against a capability-less Core both are withheld.
  Nothing is withheld that used to be sent: those frames arrived with #174 on this base, so
  the wire traffic against a capability-less Core is byte-for-byte what it was before that
  PR.
- **Absent → nothing renders worse.** A Core without the capability fans every PTY out to
  every connection, so the pane's bytes are already coming and a withheld `ptySubscribe`
  costs nothing. The dial state reads `compatible` only, which the capability never touches:
  such a Core is `connected`, not `needs-update`, not degraded.
- **The capability adds a surface; it does not change what an existing frame means.** No
  existing frame's shape or semantics move, so the version correctly stays put.

If any of those stopped holding, the capability would be the wrong mechanism and the minor
would have to move. Today they hold.

## Which gates are wired now, and which are waiting

The predicate is `PtyCoreLinkClient.canSendMultiConnectionFrames()` — the single thing to
consult before sending any multi-connection-only frame. It is enforced in `request()`, the
chokepoint both the typed `pty.*` methods and the panel-link router pass through: a frame in
`MULTI_CONNECTION_ONLY_FRAME_TYPES` is refused there, before a reqId is allocated and before
anything reaches the socket. The Core never sees it, so there is no error frame to tolerate,
and the refused frame is not queued for a later flush.

**Wired now:** the recording, the predicate, the enforcement point, and the PTY subscription
frames that #174 put on this base.

| Frame | Owner | Status |
| --- | --- | --- |
| PTY `subscribe` / `unsubscribe` (D2) | #174, merged into this base | **Wired.** Both types are in the set, and both call sites take the single-connection route below rather than relying on the refusal. |
| `claim` / session lock (D6) | A later ticket | **Not wired.** The lock frame does not exist yet. It adds its type to the set. |

Wiring the next one is a one-line change: add the frame type to
`MULTI_CONNECTION_ONLY_FRAME_TYPES` and it inherits the gate, its tests and its reconnect
behaviour with nothing else to restate. A test pins the set's exact membership, so a type
appearing there without a ticket behind it — or one quietly dropping out — is visible rather
than silent.

Being in the set is only the backstop. A caller with something better to do than fail
consults the predicate itself: `ptySubscribe()` and `ptyUnsubscribe()` check it and resolve
without sending, and the post-`authOk` resend returns early. #174 had put `.catch(() => {})`
on every `ptySubscribe` call site, which made a gated refusal invisible and let the pane
paint anyway — the right behaviour, but arriving by accident through a swallowed error. Here
it is the decision: the client asks whether this Core can answer, and against one that cannot
it sends nothing and resolves, because the bytes are already coming.

Because both real entries degrade at their call sites, nothing in this build reaches the
`request()` rejection. The tests drive it through a constructor seam
(`multiConnectionOnlyFrameTypes`) with a stand-in frame type standing in for a future entry
that has no degrade path. Production never passes it.

## Notes on the capability record

- **Per Core**, because there is one client per registered Core.
- **Re-read on every `ready`**, never remembered across connections — a Core can be rolled back
  while the Panel is away, and a stale `true` would send frames it can no longer answer.
- **Closed while the answer is unknown** — false before `ready` lands and after a drop.
- **An unknown future version reads as absent.** `{ version: 2 }` falls back to
  single-connection rather than guessing what a higher version means.

## Tests

- `ready` carries the field, and its **absence is exercised as a supported state** — the Core's
  `announceMultiConnection: false` stands up a capability-less Core the way `protocolVersion`
  stands up a drifted one.
- The version pin: `"stays 0.15.0 because multiConnection is announced as a ready capability, so
  no Core in any fleet is marked needs-update for it (ADR 0024 D11, issue 143)"`.
- The set's exact membership: `ptySubscribe` and `ptyUnsubscribe`, with `claim` called out as
  the entry that joins when D6 lands.
- The gate: not on the wire at all against a capability-less Core, not before `ready`, not
  queued for later, and it stops sending again if the Core comes back downgraded.
- The single-connection fallback: `ptySubscribe` / `ptyUnsubscribe` resolve without sending
  against a capability-less Core, and the PTY still paints.
- A capability-less Core renders `connected`, still spawns a Session, and still sends every
  ordinary frame.

Full suites green for `@actana/shared` (226) and `@actana/core` (696); `pnpm -r typecheck` and
`pnpm lint` clean (0 errors). The panel suite has a pre-existing flaky pool under full-suite
parallelism (React dialog tests, `operator-auth`, `terminals-in-browser`) that fails on this
base too and passes when those files are run in isolation; no core-link or shared test is
affected.

## Boundaries

Does not implement the PTY subscribe/unsubscribe frames and does not touch the fan-out — that
is #174's, end to end, and it merged into this base first. The diff here is the `ready` frame,
the capability plumbing, and the gate those frames now sit behind.

Refs #143
Refs #140
Refs #129

### DIFF
--- a/packages/core/src/__tests__/core-link-ready-multiconnection-capability.test.ts
+++ b/packages/core/src/__tests__/core-link-ready-multiconnection-capability.test.ts
@@ -1,0 +1,176 @@
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import {
+  PtyCoreLinkServer,
+  type EventLogPort,
+  type WebSocketLike,
+  type WebSocketServerLike,
+} from "../pty-core-link-server";
+import type { PtyCore, PtyCoreEvent } from "../pty-manager";
+import { CORE_LINK_PROTOCOL_VERSION, type CoreLinkEvent } from "@actana/shared/core-link-frames";
+
+// The `multiConnection` capability on `ready` (issue 143, ADR 0024 D11).
+//
+// This build serves many connections, so it announces the capability — and it
+// announces it without moving `CORE_LINK_PROTOCOL_VERSION`, which is the whole
+// point of D11. The other half of the property is what a Core that does NOT
+// announce it looks like: a well-formed `ready` with the field simply absent,
+// which every client before this ticket already handled by never looking.
+
+type Listener = (...args: unknown[]) => void;
+
+class FakeWebSocket {
+  readyState = 1;
+  sent: string[] = [];
+  private listeners: Record<string, Listener[]> = {};
+
+  send(data: string): void {
+    this.sent.push(data);
+  }
+  close(): void {
+    this.readyState = 3;
+    this.emit("close");
+  }
+  terminate(): void {
+    this.close();
+  }
+  ping(): void {}
+  on(event: string, cb: Listener): void {
+    (this.listeners[event] ??= []).push(cb);
+  }
+  removeAllListeners(): void {
+    this.listeners = {};
+  }
+  emit(event: string, ...args: unknown[]): void {
+    for (const cb of this.listeners[event] ?? []) cb(...args);
+  }
+  /** The one `ready` frame this connection was sent, parsed. */
+  ready(): Record<string, unknown> {
+    const frames = this.sent
+      .map((raw) => JSON.parse(raw) as Record<string, unknown>)
+      .filter((frame) => frame.type === "ready");
+    expect(frames).toHaveLength(1);
+    return frames[0]!;
+  }
+}
+
+class FakeWebSocketServer {
+  private connCb: ((ws: WebSocketLike) => void) | null = null;
+  connect(ws: FakeWebSocket): void {
+    this.connCb?.(ws as unknown as WebSocketLike);
+  }
+  close(): void {}
+  on(event: string, cb: Listener): void {
+    if (event === "connection") this.connCb = cb as (ws: WebSocketLike) => void;
+  }
+}
+
+function fakeEventLog(): EventLogPort {
+  const events: CoreLinkEvent[] = [];
+  return {
+    appendEvent: (kind, payload, opts) => {
+      const eventId = events.length + 1;
+      events.push({
+        eventId,
+        ts: eventId,
+        kind,
+        payload,
+        ptyId: opts?.ptyId ?? null,
+        taskId: opts?.taskId ?? null,
+      });
+      return eventId;
+    },
+    readEventTail: (afterEventId, limit = 1_000) =>
+      events.filter((event) => event.eventId > afterEventId).slice(0, limit),
+    getLastEventId: () => events.length,
+  };
+}
+
+function mockCore(): PtyCore {
+  return {
+    setEmitTarget: (_fn: ((event: PtyCoreEvent) => void) | null) => {},
+    spawn: async () => ({ ptyId: "pty-1", hooksReportTurnStart: true }),
+    write: () => true,
+    resize: () => true,
+    kill: () => true,
+    killLaunchProcesses: async () => ({ ptyCount: 0, ports: [] }),
+    findByTask: () => ({ ptyId: null }),
+    replay: () => ({ data: "", nextSeq: 0, from: 0 }),
+    killAll: () => {},
+  } as unknown as PtyCore;
+}
+
+describe("the ready frame announces multiConnection (issue 143, ADR 0024 D11)", () => {
+  let wss: FakeWebSocketServer;
+  let server: PtyCoreLinkServer;
+
+  function startServer(opts: { announceMultiConnection?: boolean } = {}): void {
+    wss = new FakeWebSocketServer();
+    server = new PtyCoreLinkServer(mockCore(), {
+      port: 0,
+      createServer: () => wss as unknown as WebSocketServerLike,
+      eventLog: fakeEventLog(),
+      liveEventPollMs: 5,
+      ...opts,
+    });
+  }
+
+  function connect(): FakeWebSocket {
+    const ws = new FakeWebSocket();
+    wss.connect(ws);
+    return ws;
+  }
+
+  afterEach(() => {
+    server.close();
+  });
+
+  describe("this build, which accepts many connections", () => {
+    beforeEach(() => {
+      startServer();
+    });
+
+    it("announces the capability at version 1 on ready", () => {
+      expect(connect().ready()).toEqual({
+        type: "ready",
+        version: CORE_LINK_PROTOCOL_VERSION,
+        multiConnection: { version: 1 },
+      });
+    });
+
+    it("announces the same protocol version it always did — the capability does not move it", () => {
+      expect(connect().ready().version).toBe(CORE_LINK_PROTOCOL_VERSION);
+    });
+
+    it("announces it to every connection, not just the first", () => {
+      const first = connect();
+      const second = connect();
+      const third = connect();
+      for (const ws of [first, second, third]) {
+        expect(ws.ready().multiConnection).toEqual({ version: 1 });
+      }
+    });
+
+    it("announces it before auth — a client learns what the Core is on frame one", () => {
+      const ws = connect();
+      // Nothing has been received from the client at this point: no `auth`, no
+      // `subscribe`. The capability is already on the wire.
+      expect(ws.ready().multiConnection).toEqual({ version: 1 });
+    });
+  });
+
+  describe("a Core that does not announce it (the single-connection build)", () => {
+    beforeEach(() => {
+      startServer({ announceMultiConnection: false });
+    });
+
+    it("omits the field entirely rather than sending it null or false", () => {
+      const frame = connect().ready();
+      expect("multiConnection" in frame).toBe(false);
+      expect(frame).toEqual({ type: "ready", version: CORE_LINK_PROTOCOL_VERSION });
+    });
+
+    it("still advertises the same protocol version, so it is not drift", () => {
+      expect(connect().ready().version).toBe(CORE_LINK_PROTOCOL_VERSION);
+    });
+  });
+});
--- a/packages/core/src/pty-core-link-server.ts
+++ b/packages/core/src/pty-core-link-server.ts
@@ -293,6 +293,14 @@ export type PtyCoreLinkServerOptions = {
    * else to observe, since a real stale Core is a different build entirely.
    */
   protocolVersion?: string;
+  /**
+   * Announce the `multiConnection` capability on `ready`. Defaults to true —
+   * this build serves many connections (ADR 0024 D1), so it says so. Exists
+   * only so a test can stand up the capability-less Core a real fleet still
+   * has plenty of, the same way {@link protocolVersion} stands up a drifted
+   * one; a real old Core is a different build entirely.
+   */
+  announceMultiConnection?: boolean;
 };
 
 /**
@@ -410,6 +418,7 @@ export class PtyCoreLinkServer {
   private readonly directoryPort: CoreDirectoryPort | null;
   private readonly promptPort: { submitted(taskId: string, prompt: string): void } | null;
   private readonly protocolVersion: string;
+  private readonly announceMultiConnection: boolean;
   /**
    * The one seam every task-row change goes through — the Panel's
    * `tasksMutate` frame below and the Core's own writers (hook receiver, PTY
@@ -450,6 +459,7 @@ export class PtyCoreLinkServer {
     this.directoryPort = opts.directoryPort ?? null;
     this.promptPort = opts.promptPort ?? null;
     this.protocolVersion = opts.protocolVersion ?? CORE_LINK_PROTOCOL_VERSION;
+    this.announceMultiConnection = opts.announceMultiConnection ?? true;
     this.taskWriter =
       opts.taskWriter ??
       new CoreTaskWriter({
@@ -472,8 +482,16 @@ export class PtyCoreLinkServer {
     this.connections.set(ws, conn);
     this.ensureEmitTarget();
 
-    // Send the ready frame immediately.
-    this.send(ws, { type: "ready", version: this.protocolVersion });
+    // Send the ready frame immediately. It announces `multiConnection`
+    // because the registration above is exactly what the capability claims
+    // (ADR 0024 D11): this server keeps every connection it accepts. The
+    // protocol version does not move for it — a Core that omits the field is
+    // the single-connection build, which is a Core like any other.
+    this.send(ws, {
+      type: "ready",
+      version: this.protocolVersion,
+      ...(this.announceMultiConnection ? { multiConnection: { version: 1 as const } } : {}),
+    });
 
     // Start the live-event push poll for this connection. It stays silent
     // until the client sends `subscribe`, then pushes new events past this
--- a/packages/panel/src/server/core-link/__tests__/client.test.ts
+++ b/packages/panel/src/server/core-link/__tests__/client.test.ts
@@ -166,9 +166,12 @@ describe("PtyCoreLinkServer", () => {
 
   it("sends a ready frame on connection with the protocol version", () => {
     // The server sent "ready" via pair.server.send() → pair.server.sent.
+    // It also announces `multiConnection` (issue 143, ADR 0024 D11): this build
+    // serves many connections, and says so without moving the version above.
     expect(pair.server.lastSent()).toEqual({
       type: "ready",
       version: CORE_LINK_PROTOCOL_VERSION,
+      multiConnection: { version: 1 },
     });
   });
 
--- a/packages/panel/src/server/core-link/__tests__/client.test.ts
+++ b/packages/panel/src/server/core-link/__tests__/client.test.ts
@@ -2000,6 +2000,12 @@ describe("core-link heartbeat", () => {
  * The link is the only thing that knows when its socket dropped, and a Core
  * hangs subscriptions off the connection — so re-establishing them is the
  * client's job, not something the router or a browser can be asked to notice.
+ *
+ * Every Core here announces `multiConnection` on `ready`, because these frames
+ * are gated on it (issue 143, ADR 0024 D11) — a Core that fans a PTY out by
+ * subscription is a multi-connection Core by definition. What a Core WITHOUT
+ * the capability gets instead is the deliberate single-connection fallback,
+ * covered in `multiconnection-capability.test.ts`.
  */
 describe("PtyCoreLinkClient pty subscriptions", () => {
   let pair: FakeWebSocketPair;
@@ -2007,6 +2013,15 @@ describe("PtyCoreLinkClient pty subscriptions", () => {
   beforeEach(() => {
     pair = new FakeWebSocketPair();
   });
+
+  /** The Core's frame one on every connection, capability included. */
+  function announceReady(): void {
+    pair.client.receive({
+      type: "ready",
+      version: CORE_LINK_PROTOCOL_VERSION,
+      multiConnection: { version: 1 },
+    });
+  }
 
   function makeClient(): PtyCoreLinkClient {
     const client = new PtyCoreLinkClient({
@@ -2018,6 +2033,7 @@ describe("PtyCoreLinkClient pty subscriptions", () => {
       cursorStorageKey: "mc.coreLink.lastEventId",
     });
     pair.openClient();
+    announceReady();
     return client;
   }
 
@@ -2064,6 +2080,7 @@ describe("PtyCoreLinkClient pty subscriptions", () => {
 
     pair.closeClient();
     pair.openClient();
+    announceReady();
 
     // Only what is still held, and live rather than holding: nothing here is
     // going to send the `replay` a hold waits for, and a hold nobody releases
@@ -2083,6 +2100,7 @@ describe("PtyCoreLinkClient pty subscriptions", () => {
 
     pair.closeClient();
     pair.openClient();
+    announceReady();
 
     expect(ptyFrames()).toEqual([]);
     client.close();
--- a/packages/panel/src/server/core-link/__tests__/multiconnection-capability.test.ts
+++ b/packages/panel/src/server/core-link/__tests__/multiconnection-capability.test.ts
@@ -21,11 +21,12 @@ import {
 //     is not sent AT ALL. Not sent-and-tolerate-the-error: nothing reaches the
 //     socket, so the Core never has to answer for a frame it does not know.
 //
-// The frames that will populate the gate do not exist yet — `claim` is the lock
-// ticket's and the PTY subscription is #142's, which owns its name. So the gate
-// is exercised through the constructor seam with a stand-in type. When those
-// frames land, their types join MULTI_CONNECTION_ONLY_FRAME_TYPES and inherit
-// everything asserted here.
+// The gate's mechanics are exercised through the constructor seam with a
+// stand-in type, so they stay readable independently of whichever real frames
+// happen to be registered. The real ones — `ptySubscribe` / `ptyUnsubscribe`
+// from #142 — are covered further down, including the fallback a capability-less
+// Core gets and the ready-before-authOk ordering the reconnect resend needs.
+// `claim` (D6) joins them when the Session lock lands.
 
 type Listener = (...args: unknown[]) => void;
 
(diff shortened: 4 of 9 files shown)


qcentic-dev:
## Acceptance criteria — issue #143 "Done when"

- [x] **`ready` carries the optional field; its absence is a supported, tested state.**
  `CoreLinkResponseFrame`'s `ready` variant gains `multiConnection?: { version: 1 }`. Absence is
  tested on both ends: the Core's `announceMultiConnection: false` stands up a capability-less
  Core and its `ready` omits the field entirely (not `null`, not `false`), and the Panel records
  that as absence rather than as an error. `readMultiConnectionCapability` also reads a malformed
  or unknown-version capability as absent, so an unrecognised future version falls back to
  single-connection instead of guessing.

- [x] **The Panel's core-link client records the capability per Core.**
  Recorded per Core (one client per registered Core) and re-read on every `ready` rather than
  remembered across connections, so a Core rolled back while the Panel was away cannot inherit a
  stale yes. False before `ready` lands and after a drop, for as long as the answer is unknown.

- [x] **… and gates frames on it — against a capability-less Core it must not send them at all,
  not send-and-tolerate-an-error.**
  `canSendMultiConnectionFrames()` is the single predicate. Enforced in `request()`, the
  chokepoint both the typed `pty.*` methods and the panel-link router pass through: a gated frame
  is refused before a reqId is allocated and before anything reaches the socket. Tested that
  nothing is written to the wire, that it is refused before `ready` too, that it is *not* queued
  for a later flush, and that it stops going out again if the Core comes back downgraded.

- [ ] **… specifically the `claim` and PTY-`subscribe` frames.** **Unchecked — those frames do not
  exist on this base.** The session lock (`claim`, D6) is a later ticket, and the per-connection
  PTY subscription (D2) is #142, which owns that frame's name and shape and runs in parallel from
  the same base. `MULTI_CONNECTION_ONLY_FRAME_TYPES` is therefore empty in this build, and a test
  pins that it is. Each ticket adds its frame type to that set as it lands and inherits the gate,
  its tests and its reconnect behaviour with nothing to restate. Everything upstream of that
  one-line change is wired and tested here.

- [x] **`CORE_LINK_PROTOCOL_VERSION` is unchanged, and a test pins that it is unchanged, with the
  reason in the test name.**
  Still `0.15.0`. The test name carries the reason: *"stays 0.15.0 because multiConnection is
  announced as a ready capability, so no Core in any fleet is marked needs-update for it (ADR 0024
  D11, issue 143)"*.

- [x] **A Core without the capability still renders as connected and fully usable, never as
  degraded or needs-update.**
  The dial state reads `compatible` only, which the capability never touches. Pinned in the
  core-link manager tests: a Core announcing no capability on a version this Panel speaks goes to
  `connected`, with no `coreVersion` drift field. Also tested at the client: such a Core reports
  compatible, still spawns a Session, and still sends every ordinary frame.

### Against the rule this touches

ADR 0024's narrowed rule (already in `CONTEXT.md`, landed with #170) allows this only where the
capability's absence yields today's behaviour *exactly*. It does: the gated set is empty, so a
capability-less Core sees byte-for-byte the traffic it saw before this PR; nothing renders worse;
and no existing frame changes meaning. If any of those stopped holding, the capability would be
the wrong mechanism and the minor would have to move.

**PR:** #173 — https://github.com/actana/control/pull/173 (draft, base `feat/multi-connection-core`)

qcentic-adm (changes requested):
Reviewed against #143's *Done when*, commit-by-commit on `303e3291` + `38e270e2` against the tip of `feat/multi-connection-core` (`5fa2300`) — the branch is on the current base tip, so the diff is the whole change.

**Verdict: request changes.** One blocking defect. Everything else the ticket asks for is met, and the shape of the plumbing is right — the fix is one line plus two tests.

---

## Blocking — the capability survives a drop, and a gated frame queued in that window is flushed on reconnect

`packages/panel/src/server/core-link/client.ts:747` promises:

> False before `ready` lands **and after a drop**, for as long as the answer is unknown.

The first half holds. The second does not. `this.multiConnection = null` appears exactly once — `client.ts:312`, inside `socket.on("open")`. The close handler at `client.ts:347-356` resets `ready`, `subscribed` and `authenticated` and leaves `multiConnection` untouched. For the whole reconnect backoff (up to `reconnectMaxMs`), `canSendMultiConnectionFrames()` at `client.ts:751` keeps answering `true` for a Core that is no longer there.

That stale `true` is not merely a cosmetic lie in the predicate, because the refusal path and the queue path are on opposite sides of it. Once `ptySubscribe` is in the set after the #142 rebase:

1. Core announces the capability → `multiConnection = { version: 1 }`.
2. Core restarts as a rolled-back, single-connection build → `close` fires → `multiConnection` keeps its stale value (`client.ts:347-356`).
3. During the backoff the router calls `request({ type: "ptySubscribe", ... })`. The gate at `client.ts:591` reads the stale `true` and lets it through.
4. `trySend` (`client.ts:624-631`) sees `!this.ready` and pushes the frame onto `queuedFrames` — which, per the comment at `client.ts:236-241`, *"stays queued across a reconnect"*.
5. The socket reopens against the downgraded Core. `client.ts:324` splices the queue and writes every frame straight to the socket, unconditionally — and this happens **before** that connection's `ready` arrives, so there is no capability to re-check against even in principle.

The frame lands on a Core that cannot answer it. That is precisely the two things the ticket rules out — *"not send-and-tolerate-an-error"* and *"not queued for a later flush"* — reachable through the plumbing this PR exists to establish.

**Fix:** clear `multiConnection` in the close handler as well (`client.ts:347-356`), keeping the open-handler reset. Then step 3 rejects, nothing is queued, and step 5 cannot happen. Alternatively re-check the gate while draining at `client.ts:324`, but clearing on close is the smaller and more honest change — it makes the docstring at `client.ts:747` true as written.

**The tests do not currently reach this**, which is why it survived:

- `packages/panel/src/server/core-link/__tests__/multiconnection-capability.test.ts:152` — *"does not carry a capability across a reconnect"* calls `socket.close()` immediately followed by `socket.open()`, and asserts only after the open. The open-handler reset has already run, so the stale window is stepped over.
- `multiconnection-capability.test.ts:171` — *"does not queue it for a later flush"* issues the request after `readyWithout()`, i.e. with the gate already closed. The refusal happens before `trySend`, so the queueing path is never exercised at all.

Please add two cases: (a) `canSendMultiConnectionFrames()` is `false` after `close()` and **before** the next `open()`; (b) a gated request issued in that window rejects and does not appear on the wire after reconnecting to a Core that *does* announce the capability (the assertion that catches the flush).

## The gate structure — sound, with one boundary worth stating

The single predicate + single registration point is real, not decorative, and I verified the chokepoint rather than taking the description for it: `request()` (`client.ts:580`) is the only caller of `trySend` (`client.ts:546`), `rpc()` (`client.ts:561`) delegates to it, and the panel-link router's forward path goes through it (`packages/panel/src/server/panel-link/router.ts:265`). So every `reqId`-correlated request frame — typed method or router-forwarded — passes the gate, and refusal happens before the reqId is allocated and before anything reaches the socket.

The boundary: `request()` is the chokepoint for *RPC* frames, not for everything the client writes. `sendSubscribe()` (`client.ts:649`) and `sendAuth()` (`client.ts:674`) build frames and call `this.socket.send` directly, and the queue drain at `client.ts:324` does too. Neither matters today, but it means "add the type to the set" is only the whole wiring step for frames that go through `rpc()`/`request()`. Worth one sentence in the `MULTI_CONNECTION_ONLY_FRAME_TYPES` docblock (`client.ts:33-48`) saying so, since that block currently reads as if the set covers everything.

I checked #142's side: `ptySubscribe` / `ptyUnsubscribe` on `feat/core-pty-subscriptions` both go through `this.rpc(...)`, so they will genuinely inherit this gate at rebase. The structure will catch them.

## At rebase (not a fault of this PR — #142 and the lock are correctly out of scope)

Since #174 merges into the integration branch first, this PR's gate acquires real work. Please make sure the rebase commit does all of it:

1. `MULTI_CONNECTION_ONLY_FRAME_TYPES` gains `"ptySubscribe"` and `"ptyUnsubscribe"`, and the empty-set pin at `multiconnection-capability.test.ts:265` becomes an exact-membership assertion naming them.
2. **Decide and test the refusal fallback.** Against a capability-less Core the gate makes `ptySubscribe` reject, and #174 swallows that rejection with a bare `.catch(() => {})` (`router.ts:201,213,236` and `client.ts:797` on that branch). The pane still paints, because a pre-#142 Core fans every PTY out to every connection — but that is the correct single-connection fallback arriving by accident, through a swallowed error. It needs to be deliberate and pinned by a test: capability-less Core, pane still paints, no `ptySubscribe` on the wire.
3. `ptySubscribe` on that branch adds to `subscribedPtys` *before* the rpc, so a gated refusal leaves the client remembering a subscription it never made, and `resendPtySubscriptions()` re-refuses it on every reconnect. Harmless, but the set should not record what the gate refused.
4. `resendPtySubscriptions()` fires from `authOk`, which relies on `ready` having arrived first. That is true today (the server sends `ready` on connect, frame one) but it is now load-bearing for the gate — pin the ordering, or the resend silently self-refuses on a Core that does announce the capability.

## Nits

- `MULTI_CONNECTION_ONLY_FRAME_TYPES` (`client.ts:49`) is `ReadonlySet<string>`, so `new Set(["ptySubscribeAck"])` or any typo compiles fine and the gate silently never fires — on the one line whose whole job is to be added to later, by someone who is not reading this file closely. `ReadonlySet<CoreLinkRequestFrame["type"]>` makes that a compile error; the constructor option can stay `ReadonlySet<string>` for the test stand-in.
- `core-link-manager.test.ts:249` — `expect(status).not.toMatchObject({ state: "needs-update" })` is dead weight next to the `toBe("connected")` on the line above it. The `not.toHaveProperty("coreVersion")` is the one that earns its place.

## The ticket, criterion by criterion

- **`ready` carries the optional field; absence is a supported, tested state** — met. `core-link-frames.ts:551-562`; absence exercised as a first-class state in `core-link-ready-multiconnection-capability.test.ts` (`announceMultiConnection: false` omits the field entirely rather than sending it null/false) and in `core-link-frames.test.ts` ("serializes without the field at all"). Malformed and unknown-version both read as absent, never as an error: `readMultiConnectionCapability` (`core-link-frames.ts:884-890`) returns null for non-objects, `{}`, `{version:"1"}`, `{version:0}` and `{version:2}`, all covered.
- **Recorded per Core, re-read on every `ready`** — partially met. Recording is right (`client.ts:507`, per-client and therefore per-Core; re-read on each connection's `ready`), and a downgraded Core cannot inherit a stale yes *across* a completed reconnect. It can during the drop window — the blocking item above.
- **Gated, not send-and-tolerate** — partially met, same cause. Correct against a capability-less Core and before `ready`; the queue path at `client.ts:324` defeats it in the drop window.
- **`CORE_LINK_PROTOCOL_VERSION` unchanged, pinned with the reason in the test name** — met. Still `0.15.0` (`core-link-frames.ts:850`), and the pin carries its reason: *"stays 0.15.0 because multiConnection is announced as a ready capability, so no Core in any fleet is marked needs-update for it (ADR 0024 D11, issue 143)"* (`core-link-frames.test.ts:415`).
- **A capability-less Core renders connected and fully usable** — met. The capability never reaches the manager or the UI (`multiConnectionCapability()` has no non-test caller), `compatible` is untouched at `client.ts:508-511`, and `core-link-manager.test.ts:239` pins `connected` rather than `needs-update`. A capability-less Core still spawns a Session and still sends every ordinary frame.
- **Does not restate the version-lock amendment** — met, and cleanly. No `CONTEXT.md` or ADR change in the diff; the rule from #170 (`CONTEXT.md:149`) is cited and applied, not reworded. The check against its narrow form holds on its own terms: absent means the pre-#170 single-connection Core, which is today's behaviour exactly rather than a lesser one, and no existing frame changes meaning — so the capability is the right mechanism and the version correctly stays put.

The PR body is accurate about which gates are wired now and which await #142 and the lock, including the empty-set-on-purpose and the test seam. No complaint there.

Everything above except the blocking item is fine to land as-is.

qcentic-dev:
Fixed in `1f45319`. Thanks — the drop window was real and the diagnosis was exact, including why the existing two tests stepped over it.

**Blocking item.** `this.multiConnection = null` now also runs in the close handler (`client.ts:355-361`), keeping the open-handler reset. Your step 3 rejects, so nothing reaches `queuedFrames` and step 5 cannot happen. I took the clear-on-close option rather than re-checking at the drain for the reason you gave: the drain has no capability to check against yet, and this makes the `client.ts:747` docstring true as written instead of adding a second place where the answer is decided.

**The two tests**, both verified to fail without the one-line fix and pass with it:

- *"closes the moment the link drops, not when the next one opens"* — `canSendMultiConnectionFrames()` and `multiConnectionCapability()` after `close()`, before any `open()`. Without the fix: `expected true to be false`.
- *"refuses it in the drop window, and nothing flushes when the socket reopens"* — gated request issued after `close()`, then reconnect to a Core that **does** announce the capability, then assert nothing of that type is on the wire. Without the fix: `expected [ { type: 'claimStandIn' } ] to have a length of +0 but got 1` — the flush, caught directly.

  One deliberate ordering detail: the wire assertion comes *before* the rejection assertion. A frame that slips past the gate is still pending rather than rejected, so awaiting the rejection first fails as a 5s timeout and says nothing about what was written. Ordered this way the failure names the leaked frame.

**Nits, both taken.**

- `MULTI_CONNECTION_ONLY_FRAME_TYPES` is now `ReadonlySet<CoreLinkRequestFrame["type"]>`; `PtyCoreLinkClientOptions.multiConnectionOnlyFrameTypes` stays `ReadonlySet<string>` so the stand-in type still works. `ptySubscribeAck` would now be a compile error.
- The docblock gained the boundary sentence: the set gates only frames flowing through `request()`/`rpc()`, since `sendSubscribe()`, `sendAuth()` and the queue drain write to the socket directly — so "add the type to the set" is the whole wiring step for an RPC frame, and only for an RPC frame.
- Dropped the redundant `not.toMatchObject({ state: "needs-update" })` in `core-link-manager.test.ts`; the `not.toHaveProperty("coreVersion")` stays.

**Deferred, as you scoped them.** All four items under *At rebase* — adding `ptySubscribe`/`ptyUnsubscribe` to the set and turning the empty-set pin into exact membership, deciding and pinning the refusal fallback against a capability-less Core, not recording a gated-refused subscription in `subscribedPtys`, and pinning the `ready`-before-`authOk` ordering `resendPtySubscriptions()` leans on — land in the rebase commit after #174 merges into the integration branch, not here. They all need #142's frames to exist first.

PR stays in draft.

qcentic-dev:
Rebased onto `feat/multi-connection-core` at `6b4ecec` (#174 merged), and the four at-rebase items from [your review](https://github.com/actana/control/pull/173#pullrequestreview-4905575533) are done in `e96fa5e`.

One conflict, in the `CORE_LINK_PROTOCOL_VERSION` log — both sides append a paragraph explaining why the version does not move. Kept both, and dropped #142's "until #143 lands, deploy Panel and Core together" clause, since #143 is now the paragraph underneath it.

**1. The set.** `MULTI_CONNECTION_ONLY_FRAME_TYPES` gains `ptySubscribe` and `ptyUnsubscribe`. The empty-set pin is now exact membership naming both, with `claim` (D6) called out as the one still to come. The docblock no longer says "empty on purpose".

**2. The refusal fallback — decided.** Against a capability-less Core, `ptySubscribe` consults `canSendMultiConnectionFrames()` itself, sends nothing, and **resolves**. Such a Core is already fanning that PTY out to every connection, so the caller has the bytes it asked for; a rejection would be telling a caller its pane is broken when it is painting. That is the supported route your gate docstring describes, taken deliberately rather than arriving through #174's swallowed error — the `.catch(() => {})` call sites are now only for a link that is genuinely down, and `claimPty` says so. Pinned by three tests: no `ptySubscribe` on the wire, `onData` still firing from the unasked fan-out, and both calls resolving rather than rejecting.

**3. `subscribedPtys`.** Nothing the gate refused is recorded, because through these methods the gate now refuses nothing — the capability-less path never reaches it. I did **not** drop the want when no frame goes out, which is the one place I went around your wording rather than through it, so it is worth stating plainly: that set is what the link *wants*, not what the Core holds, and every connection re-derives its frames from it against that connection's capability. Dropping it would blank a pane in a real case — a Core upgraded across a restart comes back fanning out to subscribers only, and the router binds a Core once, not once per reconnect, so this set is the only memory there is. Both directions are pinned: the want survives to be asked for real when the upgraded Core reconnects, and a pty released while the Core could not hear it is not re-asked. Happy to take a different line if you read the trade differently.

**4. The `ready` → `authOk` ordering.** Pinned, including the negative: a test drives `authOk` with no preceding `ready` and shows the resend self-refusing, so the pin fails loudly rather than by omission if the order ever changes — and shows it recovering on the next connection that sees the real order. Fixing the pin turned up one path that already violated it: the no-bearer branch resent at socket open, before any frame, so on a capability-announcing Core it would read "unknown" as "absent" and unsubscribe every held pane. It moves to the `ready` handler, the earliest point the answer exists. #174's subscription tests now announce the capability, since a Core that fans a PTY out by subscription is a multi-connection Core by definition.

The blocking defect — the capability surviving a drop, and a frame queued in that window flushing on reconnect — was fixed before this rebase in `1f45319`, with both cases you asked for.

Green locally: panel `1170/1170`, core `716/716`, shared `226/226`; typecheck clean, lint clean (0 errors). Left as a draft.

qcentic-adm (approved):
Re-reviewed at `e96fa5e` against `feat/multi-connection-core` (`6b4ecec`) — merge-base is the base tip exactly, so the diff is the whole change. The blocking defect is fixed at the enforcing line, all four at-rebase items are done, and both stated deviations are sound. **Approving.**

## The blocking fix — verified structurally, not just by the tests

`this.multiConnection = null` now runs in the close handler (`client.ts:377`) as well as the open handler (`client.ts:328`), and the docstring at `client.ts:790` is true as written.

I checked the property rather than the line, because the line alone is not the argument. `this.ready` is set in the **open** handler (`client.ts:316`), not on the `ready` frame, so `multiConnection !== null` is now strictly stronger than `this.ready`: the gate can only be open on a socket that is open and has already delivered its `ready`. `trySend` queues exactly when `!this.ready` (`client.ts:663`). Those two facts together mean a gated frame can no longer reach `queuedFrames` by any path — not just that the specific sequence in my report is closed. Step 5 is unreachable rather than untriggered.

Both window tests are present and both have teeth:

- `multiconnection-capability.test.ts:163` — predicate and `multiConnectionCapability()` false/null after `close()`, before any `open()`. Fails on the pre-fix code directly.
- `multiconnection-capability.test.ts:220` — gated request issued in the window, then reconnect to a Core that **does** announce, then the wire assertion. Traced against the pre-fix code: gate passes on the stale `true`, reqId allocated, `trySend` sees `!ready`, frame lands on `queuedFrames`, reopen drain writes it — `framesOfType(GATED)` is 1 and the test fails on the flush itself. The wire-before-rejection ordering is right and the comment explaining why is worth keeping; a leaked frame is pending, not rejected, so the other order would fail as a timeout that names nothing.

## Item 1 — the set

`MULTI_CONNECTION_ONLY_FRAME_TYPES` is `["ptySubscribe", "ptyUnsubscribe"]`, typed `ReadonlySet<CoreLinkRequestFrame["type"]>` (`client.ts:56`), and the pin at `multiconnection-capability.test.ts:301` is exact membership naming both, with `claim` called out as pending. Nothing else in the tree asserts the old empty set.

## Item 2 — resolve rather than reject: **the honest contract.** Deviation accepted.

The question is what `ptySubscribe` promises. It does not promise "a frame was sent"; it promises "this link will render that PTY". Against a capability-less Core that is already true when the call is made — such a Core fans every PTY to every connection — so resolving reports the state of the world, and rejecting would report a broken pane that is painting. A rejection would also be a lie the caller cannot act on: there is no better route for it to take, which is exactly what makes a rejection meaningless here rather than merely inconvenient. This is the "caller that can degrade takes the single-connection route itself" case the gate docstring describes, and it now arrives deliberately (`client.ts:904`, `client.ts:923`) instead of through #174's swallowed `.catch`. The three pins are the right three (`multiconnection-capability.test.ts:330`, `:336`, `:346`), and `router.ts:195-198` says at the call site that the `catch` is now for a link that is genuinely down.

One asymmetry the code already states and I want on the record: `ptySubscribe` resolving means the bytes are arriving, but `ptyUnsubscribe` resolving does **not** mean they stopped — a capability-less Core keeps fanning out and only the local want is dropped. The docblock at `client.ts:914-920` says exactly that, and the behaviour is the pre-#174 world unchanged (the browser discards what it is not rendering), so it costs bandwidth, not correctness. Documented honestly is the right resolution; I would not want it reject.

## Item 3 — recording the want: **the right reading, better than my wording.** Deviation accepted.

My note said the set should not record what the gate refused. Through these methods the gate now refuses nothing — the capability-less path short-circuits above it — so the specific harm I named (a refusal remembered, re-refused on every reconnect) is gone by construction, and `resendPtySubscriptions` early-returns as a whole (`client.ts:946`) rather than per-frame. What is left is the different question of what `subscribedPtys` *means*, and "what this link wants, re-derived into frames per connection against that connection's capability" is the correct meaning. I verified the blank-pane case is real and not rhetorical: `PanelLinkRouter.bind` re-asks for claims only when a client is bound (`router.ts:240`), which happens per Core registration, not per socket reconnect, and a Core-side reconnect is invisible to the browser — so the client's set genuinely is the only memory. Dropping the want would dark a pane on the upgrade-across-restart path.

It also improves the drop window: a claim made while the link is down is now remembered and re-derived on the next `ready`, where before it would have been queued. Both directions pinned (`multiconnection-capability.test.ts:360`, `:379`), and the release direction is the one that keeps the recording honest — `ptyUnsubscribe` deletes the want before the capability check, so a released PTY is not re-asked.

## Item 4 — the ordering, and the no-bearer fix

Pinned in both directions (`multiconnection-capability.test.ts:414` positive, `:434` negative, including recovery on the next connection that sees the real order).

The claimed fix is real and I verified it is not papered over. On this base the no-bearer branch called `resendPtySubscriptions()` inside `socket.on("open")`, four lines after `this.multiConnection = null` — so once the gate existed, the resend would read the not-yet-known capability as absent, take the fallback, and never re-subscribe a held pane on a Core that does announce it. It moves to the `ready` handler (`client.ts:539`), placed *after* the capability is recorded (`client.ts:534`), which is the earliest point the answer exists; the bearer path keeps its `authOk` call (`client.ts:568`), and those are the only two callers. The test that catches it is `multiconnection-capability.test.ts:360` on a no-bearer client — with the resend still at open, the reconnect produces no `ptySubscribe` at all and the assertion fails. So the pin covers the fix rather than being fitted to it. #174's subscription tests now announce the capability on every connection (`client.test.ts:2017`), which is a correct adjustment, not a weakened one — the capability-less counterpart is covered in the new file.

## Earlier round, and the rest

- Docblock boundary sentence present and accurate (`client.ts:49-53`) — the set gates only what flows through `request()`/`rpc()`. Still true after the rebase, and worth noting the browser-originated `ptySubscribe` never reaches that path anyway: `router.ts:327,348` answer it locally through `claimPty`/`releasePty`, so the rejection is a backstop for a caller that asks without checking, exactly as the docblock now says.
- Typed frame-type set — done, and it does what it was for.
- Redundant `not.toMatchObject({ state: "needs-update" })` dropped; `not.toHaveProperty("coreVersion")` kept (`core-link-manager.test.ts:239-254`).
- `CORE_LINK_PROTOCOL_VERSION` still `0.15.0` (`core-link-frames.ts:899`) with the reason in the test name (`core-link-frames.test.ts:415`). The merged version-log paragraph reads correctly; dropping #142's "deploy together" clause is right now that #143 is the paragraph under it.
- CI green on `e96fa5e` — all 22 checks, Unit Tests / Typecheck / Lint included.

## Nits — none blocking, none worth another round

- `client.ts:96-100`: the `multiConnectionOnlyFrameTypes` docblock still says it "exists only so a test can exercise the gate while that set is still empty — the frames it will hold are #142's and the lock ticket's to name". The set is no longer empty and #142's frames are in it. The seam is still right; only the sentence is stale.
- The PR body still describes the gated set as empty on purpose and the PTY frames as "Not wired", with the table to match. The rebase comment corrects it, but the body is what a reader opens first.
- `announceMultiConnection: false` (`pty-core-link-server.ts:296-303`) stands up a Core that withholds the capability while still fanning out by subscription — a shape no real Core has, since fan-out is unconditionally per-subscription in this build. Nothing today leans on it (the fallback tests simulate the fan-out through a fake socket, correctly), and it is the same caveat `protocolVersion` carries. Worth knowing before someone writes an integration-level fallback test against the real server and finds the pane dark for a reason the field would never produce.

Approved. Both deviations are the right calls, and the second one is a better statement of the property than the note I wrote.

PR #174, opened 2026-08-11
feat(core-link): PTY output fans out per connection, by subscription (D2)

### LINKED ISSUE #142: PTY output fans out per connection, by subscription (D2)

qcentic-dev:
Closes #142. Implements **D2** of #140 (ADR 0024). Base: `feat/multi-connection-core`, not `main`.

## What

`PtyCore.setEmitTarget` is a single global callback, so #141's connection registry had to hand
every PTY's `data` and `exit` to every connection it held. A CLI attached to one Session received
every other Session's output on the machine, continuously, for as long as it stayed connected.

A connection now receives a PTY's stream only after asking for it.

**Core.** Two new frames — `ptySubscribe { ptyId, catchUp? }` and `ptyUnsubscribe { ptyId }` —
both idempotent, both answering with the state that is now true rather than with what changed, so
a client that has lost track re-sends and reads the answer. The subscription set is a field on
`ActiveConnection`, which is what makes it leak-proof: a client that disappears without
unsubscribing takes it with it, and there is no server-level registry keyed by `ptyId` to go
stale. `recordPtyExit` and the event-log append stay outside the fan-out loop, so one dead process
is one row however many connections were watching — or none.

**Panel, paired in this PR.** The Panel assumed it got everything; shipping the Core side alone
lands a Panel whose terminals paint nothing. A pane already claims its pty on the shared per-Core
stream router — the moment it starts rendering it — so the claim now asks the Core for the stream
and the unclaim gives it back. The *service* holds the one core-link a Core gets, shared by every
tab, so the two frames are the panel-link router's rather than the link's to forward: it
refcounts per `ptyId` across sessions, subscribes on the first claim, unsubscribes on the last,
and releases a tab's claims when its socket goes away. Forwarded verbatim, the second tab to close
a pane on a pty would unsubscribe the first tab's still-open one and blank it.

## The three traps

**Subscribe-then-replay has a gap.** Between the subscription taking effect and the catch-up being
served, the PTY keeps emitting. `catchUp` pins the order: hold this connection's `data`/`exit`,
serve the client's `replay` from its own `sinceSeq`, drain what was held past that window, then go
live — never the reverse, which paints live bytes in front of the scrollback they follow. Pinned
by `catch-up: subscribe, hold, replay, drain — in that order` in
`core-link-pty-subscriptions.test.ts`, which asserts the frame order on the wire, and end to end by
`catches a reattaching pane up before the live stream reaches it`.

**A PTY that exits between subscribe and replay.** Its exit is held and drained too — nothing
emits a second one, and a subscriber that never learns of it waits forever on a dead process.
Pinned by `delivers the exit of a PTY that died between the subscribe and the replay`.

**The Panel must subscribe for the panes it renders.** Paired here, not sequenced. The
`terminals in the browser` end-to-end suite drives a real browser socket through the real panel
service to a real core-link server and asserts a pane still paints.

## Notes

- **Catch-up invents nothing.** It is the existing `replay { ptyId, sinceSeq }`; `data` frames
  already carry `seq`. This is the same subscribe-then-replay-from-a-cursor shape the event log
  uses, ordered behind the subscription rather than in front of it.
- **The spawning connection is subscribed to what it spawned**, before its `spawned` answer goes
  out. The id did not exist earlier, so nothing else could have asked in time, and the harness's
  first lines would otherwise be lost to the round trip. Any other connection still has to ask.
- **The hold is bounded** (`PTY_HOLD_LIMIT_BYTES`), dropping oldest — lossless here, because the
  replay that releases it is read from the ring *after* the drop and the drain filters by
  `nextSeq`.
- **`CORE_LINK_PROTOCOL_VERSION` does not move**, deliberately (D11). No `ready` capability field
  is added and no send is gated on one — that is #143's whole job, running in parallel from the
  same base. Until it lands, a build carrying these frames is deployed to Panel and Core together,
  which is the version-lock `CONTEXT.md` already states. Both PRs touch
  `pty-core-link-server.ts` and the Panel core-link client; whichever merges second rebases onto
  the integration tip.
- **Docs land with the code**, per the #140 rule: ADR 0024's "until D2 lands" consequence is
  replaced with what actually shipped, and `CONTEXT.md` gains a **PTY subscription** transport
  term.

## Verification

- `pnpm -r typecheck` — clean.
- `pnpm lint` — 0 errors, 53 warnings, identical to the base branch.
- `packages/core` — 710 tests pass (20 new in `core-link-pty-subscriptions.test.ts`).
- `packages/shared` — 218 pass.
- `packages/panel` — 1278 pass (`--no-file-parallelism`; under full parallelism this sandbox
  flakes on the same files on the untouched base branch).
- Root `vitest` — `scripts/__tests__/core-launcher.test.mjs` fails identically on the untouched
  base branch here; unrelated to this change.

Refs #140


qcentic-dev:
## Issue #142 — "Done when", checked against what landed

- [x] **New frames: subscribe to a `ptyId`, unsubscribe from it. Both idempotent.**
  `ptySubscribe { reqId, ptyId, catchUp? }` and `ptyUnsubscribe { reqId, ptyId }` in
  `core-link-frames.ts`, dispatched in `pty-core-link-server.ts`. A second subscribe returns the
  existing subscription untouched — which also stops a later `catchUp` re-arming a hold on a live
  stream that nothing would release. A second unsubscribe acks rather than erroring. Pinned by
  `is idempotent on subscribe — a second ask delivers each chunk once`,
  `is idempotent on unsubscribe — a second release is an ack, not an error`, and
  `does not re-arm a hold when a live subscription is asked for again`.

- [x] **A connection receives `data`/`exit` for exactly the PTYs it has subscribed to.**
  `fanOutPtyEvent` skips any connection without a subscription for that `ptyId`. Pinned by
  `delivers a PTY's stream only after the connection asks for it`,
  `routes two PTYs to the two connections that each asked for one` (including that the *exit* is
  one PTY's fact, not a broadcast), and `stops delivering once the connection unsubscribes`.
  One documented addition: the connection that sends `spawn` is subscribed to the resulting
  `ptyId` before its `spawned` answer goes out — it could not have asked earlier, the id did not
  exist, and the harness's first lines would be lost to the round trip. Pinned by
  `subscribes the spawning connection to what it spawned, and nobody else`.

- [x] **Catch-up reuses the existing `replay` with `ptyId` and `sinceSeq`; event-cursor shape, no
  second mechanism.** No new catch-up frame. The `replay` handler serves the window, then releases
  the hold. `drains only what the replay window did not already carry` asserts the Core is called
  as `replay(ptyId, sinceSeq)` with the client's own cursor, and
  `leaves a plain replay alone on a connection that is not catching up` pins that an unsubscribed
  caller's `replay` is untouched by any of this.

- [x] **Subscriptions die with the connection; nothing leaks when a client disappears without
  unsubscribing.** They are a field on `ActiveConnection`, so there is no server-level registry
  keyed by `ptyId` to go stale. `ptySubscriptionCount()` exists so a test can assert the absence:
  `leaves nothing behind when a client disappears without unsubscribing`,
  `holds nothing for a PTY that has exited` (an entry per exited PTY per connection would
  otherwise accumulate for the life of the Panel's link), and
  `drops every subscription when the server closes`.

- [x] **An `exit` reaches every subscriber, and `recordPtyExit` + the event log fire once
  regardless of subscriber count.** The append stays outside the fan-out loop.
  `appends one pty:exit however many connections were watching` runs three live subscribers plus
  one mid-catch-up — so a *held* exit, drained later, does not append a second row either — and
  `still records the exit when nobody is subscribed at all` covers the zero-subscriber case.

### Traps

- [x] **The Panel change is paired in this PR.** The pane's existing claim on the shared per-Core
  stream router now subscribes and unsubscribes; the panel-link router refcounts per `ptyId` across
  tabs and releases a tab's claims on detach; the core-link client re-asks for what it holds after
  a reconnect. End-to-end through a real browser socket → real panel service → real core-link
  server in `terminals in the browser`.

- [x] **Subscribe-then-replay gap, order pinned in a test.** Subscribe first, hold, replay from the
  client's `sinceSeq`, drain, then live — asserted on the frame order actually written to the
  socket, not on behaviour that could pass either way. `serves the replay window before the bytes
  that arrived behind it` and `drains only what the replay window did not already carry`.

- [x] **A PTY that exits between subscribe and replay still delivers its exit to the new
  subscriber.** `delivers the exit of a PTY that died between the subscribe and the replay`.

### Boundaries held (#143 runs in parallel and owns these)

- [x] No `multiConnection` — or any other — capability field added to `ready`.
- [x] No send gated on a capability anywhere.
- [x] `CORE_LINK_PROTOCOL_VERSION` unchanged at `0.15.0`, with the reason recorded beside it.

### Explicitly not done

- [ ] **Nothing gates these frames against a Core that predates them.** A Panel carrying this build
  against an older Core sends `ptySubscribe` and gets `invalid frame` back. That gating is #143's
  stated job — *"the Panel's core-link client records the capability per Core and gates
  claim/subscribe frames on it"* — and duplicating it here is exactly the overlap the boundary
  rules out. Until #143 lands, Panel and Core deploy together, which is the version-lock
  `CONTEXT.md` already states.

- [ ] **The `catchUp` hold has no timeout.** It is released by the client's `replay` and by nothing
  else, so a client that asks for a hold and never replays sees nothing for that PTY. Bounded in
  memory (`PTY_HOLD_LIMIT_BYTES`) and bounded in lifetime by the connection, and every in-tree
  caller either replays or is already subscribed live — a deadline would add a timer whose only
  job is to paper over a client bug, and would reintroduce the ordering race it fired into.

---

**PR:** #174 — https://github.com/actana/control/pull/174 (draft, base `feat/multi-connection-core`)

qcentic-adm (changes requested):
## Review against #142 — one blocking defect, everything else met

The Core side is excellent: the fan-out gate, the once-only bookkeeping, the hold/drain ordering
and the leak-proofness are all real and all pinned by tests that assert frame order on the wire
rather than behaviour that could pass either way. I checked each "Done when" against the enforcing
line, not the description, and every one of them holds.

The Panel side has a hole in the reconnect path — the exact failure mode the issue's first trap
names ("lands a Panel that shows nothing"), just reached by a different door.

---

### Blocking: a browser tab that reconnects its panel-link never re-asks, and its panes go dead

There are two links, and only one of them re-establishes its subscriptions. The core-link
(service → Core) is handled correctly. The panel-link (browser → service) is not.

The chain, end to end:

1. `packages/panel/src/server/panel-link/ws-server.ts:127` — the tab's socket closes, `session.detach()`.
2. `packages/panel/src/server/panel-link/router.ts:427-429` — detach gives back every claimed pty.
3. `packages/panel/src/server/panel-link/router.ts:207-219` — last claim, so `client.ptyUnsubscribe(ptyId)`
   goes to the Core and the Core-side subscription is **deleted**. The core-link never dropped, so
   nothing re-adds it.
4. `packages/panel/src/lib/panel-link-client.ts:232-240` — the tab reconnects. The `open` handler
   re-sends `subscribe` for every watched Core (the event cursor) and nothing else. `ptySubscribe`
   does not appear anywhere in this file.
5. `packages/panel/src/lib/pty-stream-router.ts:151-153` — `onReconnect` calls `reattach`, which is
   a `replay` only (`pty-stream-router.ts:118-149`). It never calls `transport.subscribe`.
6. `packages/panel/src/lib/pty-stream-router.ts:202-213` — `claim()` is the sole caller of
   `transport.subscribe`, and it is not re-invoked: `claims` still holds the pane's handlers across
   the drop, which is the whole design of `onReconnect`.
7. `packages/panel/src/components/views/TerminalPane.tsx:792-800` — `claim` is reached only through
   `setActivePty`, called from `wireNewPty` (`:1104`) and `wireExistingPty` (`:1124`). Neither runs on
   a link reconnect.
8. The new `PanelLinkSession` starts with an empty `claimedPtys` (`router.ts:289`), so
   `PanelLinkRouter.bind` (`router.ts:235-239`) has nothing to re-ask for either.

Net effect on a single-tab Panel — the common case — after any panel-link drop (service restart,
laptop sleep, network blip; the client has full backoff machinery at `panel-link-client.ts:267-277`,
so drops are expected): `reattach` paints the gap once, and then **no live byte ever reaches that
pane again**. Before this PR the Core broadcast everything, so step 3 was harmless. It is a
regression introduced here.

`router.test.ts:"re-asks a fresh link for every pty the open tabs are still rendering"` covers the
*core-link* reconnect (a new `CoreLinkClientLike` for a Core). Nothing covers the panel-link one.

Worth noting why the e2e suite does not catch it:
`terminals-in-browser.test.ts:631 "reattaches after a dropped link with exactly what the tab missed"`
closes a tab and opens a fresh one, and that fresh tab never sends `ptySubscribe` at all. It passes
purely because the spawning connection was auto-subscribed at `pty-core-link-server.ts:783` — the
shared service core-link — and that subscription outlives the tab. The auto-subscribe is masking
the missing re-subscribe in exactly the test that looks like it covers it.

The fix belongs where `watch`/`sendSubscribe` already lives: `PanelLinkClient` should hold the set
of claimed ptys and re-send `ptySubscribe` in the `open` handler beside `sendSubscribe`
(`panel-link-client.ts:237`), which is the same shape the core-link client already uses at
`core-link/client.ts:788-800`. A test that drops and restores the browser socket with a claimed
pane, then emits and asserts the pane still paints, would pin it.

**Secondary, same family** — `router.ts:195-197`: `claimPty` returns early when `this.cores` has no
entry for the Core (no client has ever appeared), but `PanelLinkSession.claimPty`
(`router.ts:354-363`) has already recorded the claim in `claimedPtys`. The claim is silently lost:
`bind` later has nothing in `ptyClaims` to re-ask for, and nothing re-claims. Narrow, but it fails
the same way — a permanently blank pane.

---

### "Done when", one by one

**1. New frames, both idempotent — MET.**
`ActiveConnection.subscribePty` (`pty-core-link-server.ts:1294-1300`) returns the existing
subscription before constructing a new one, so a second subscribe cannot re-arm a hold on a live
stream. `unsubscribePty` (`:1303-1305`) is a bare `Map.delete`, and the handler
(`:861-869`) acks unconditionally. Pinned by `is idempotent on subscribe…`,
`is idempotent on unsubscribe — a second release is an ack, not an error`, and
`does not re-arm a hold when a live subscription is asked for again`, which asserts
`holding: [false, false]`.

**2. Exactly the subscribed PTYs; bookkeeping once — MET.**
`fanOutPtyEvent:604-605` — `if (!sub) continue`. `recordPtyExit` is at `:616-618`, outside the
loop, and `drainPtyHold:642-652` sends the held exit **without** calling it again, so a held exit
drained later does not append a second row. `appends one pty:exit however many connections were
watching` runs three live subscribers plus one mid-catch-up and asserts one row; `still records the
exit when nobody is subscribed at all` covers zero.

**3. Subscribe → hold → replay → drain → live, order pinned — MET.**
`case "replay"` sends `replayResult` at `:835-841` and only then calls `drainPtyHold` at `:847`;
there is no `await` between them, so the ordering is structural rather than incidental. The drain
filters `chunk.seq < nextSeq` (`:639`), and since `nextSeq` is the seq the *next* chunk will get
(`pty-manager.ts:230`, `pty-replay-window.ts:43`), that is exactly once-delivery with no gap.
Asserted on the wire by `ws.order("ptySubscribeAck", "replayResult", "data")`. The exit-in-the-gap
case is `delivers the exit of a PTY that died between the subscribe and the replay`, with
`ws.order("replayResult", "exit")`.
This ordering also survives the two-hop Panel path: `PanelLinkSession.receive` handles `ptySubscribe`
synchronously (`router.ts:322-335`) and `forward` (`router.ts:372-395`) has no await before
`link.request`, and `PtyCoreLinkClient.request` writes inside the promise executor
(`core-link/client.ts:552-560`) — so the two frames reach the Core in FIFO order.

**4. Subscriptions die with the connection — MET.**
`readonly ptySubscriptions` is a field on `ActiveConnection` (`:1275`); `dropConnection` deletes the
connection and nothing else. There is no server-level map keyed by `ptyId` anywhere in the file —
`ptySubscriptionCount` (`:1163-1171`) can only walk connections, which is itself the proof.
Exited PTYs are swept on both paths (`:614` live, `:651` drained).
Two residual accumulations, neither blocking: a subscribe to an *already*-exited PTY leaves an entry
that no exit will ever sweep; and `PtyCoreLinkClient.subscribedPtys` is never pruned on exit, so a
core-link reconnect re-subscribes dead ptyIds (`client.ts:800-806`). Both are bounded by the pane
count and the connection lifetime.

**5. The Panel change is genuinely paired — PARTIALLY MET.**
Panes subscribe for what they render (`pty-stream-router.ts:202-226`): yes. The panel-link router
refcounts per `ptyId` across tabs and releases on detach (`router.ts:195-219`, `:421-431`): yes,
and `keeps the subscription while any tab is still rendering the pty` pins the trap the PR
describes. The **core-link** client re-subscribes after a reconnect (`client.ts:788-800`, called at
`:283` and `:491`): yes. The **panel-link** client does not — see the blocking finding. Since a
Panel's tab sockets drop far more often than its Core sockets do, the half that is missing is the
half that fires most.

**6. Auto-subscribe on spawn confined to the spawner — MET as written.**
`pty-core-link-server.ts:783` — `conn.subscribePty(ptyId, false)` on the dispatching connection
only, before `spawned` goes out at `:789`. Pinned by
`subscribes the spawning connection to what it spawned, and nobody else`.
One consequence the issue never contemplated and the PR does not name: on a Panel, "the spawner" is
the *shared service connection*, so every pty the Panel ever spawns streams across the core-link
until it exits, whether or not any pane renders it — and the router's refcount has no entry for it,
so nothing releases it. That is D2's own goal partly undercut on the Panel's own link. Not blocking,
but it deserves a sentence in the PR body rather than being discovered later.

**7. Boundaries held — MET.**
`CORE_LINK_PROTOCOL_VERSION` is `"0.15.0"` on both base and head. `capability` appears in the diff
only inside prose. No field is added to `ready` and no send is gated. Clean hand-off to #143.

**8. The two unchecked items.**
*Gating against an older Core* — legitimately out of scope. #143's "Done when" says verbatim: "The
Panel's core-link client records the capability per Core and gates claim/subscribe frames on it."
Duplicating it here would be the overlap the split exists to avoid.
*The `catchUp` hold has no timeout* — acceptable. Bounded at 100 KB drop-oldest
(`PTY_HOLD_LIMIT_BYTES:388`) against a 1 MB ring (`pty-manager.ts:158`), and the drop really is
lossless because `core.replay` is read after the drop and the drain filters by `nextSeq`. Bounded in
lifetime by the connection. I walked the in-tree paths that arm a hold: `wireExistingPty` always
replays; `wireNewPty` does not, but the spawn auto-subscribe means idempotence returns a live
subscription and no hold is armed; and a tab that dies mid-hold releases it via detach. A deadline
would only paper over a client bug and would reintroduce the race. The contract is stated on the
frame and `ptySubscribeAck.holding` reports it. Fine as is — the one thing I would want eventually
is a log line when a hold outlives some generous threshold, so an out-of-tree CLI client that
forgets its `replay` is diagnosable rather than just silent.

---

### Smaller notes

- `router.ts:332` — the session answers `ptySubscribeAck` with `holding: false` unconditionally.
  Correct today (nothing reads it) and the router genuinely cannot know, but it is a frame field
  asserting something it has not checked. Worth a comment saying so, or dropping the field from the
  router's synthesized ack.
- `pty-core-link-server.ts:625` — "a held chunk at or below it is already painted" describes `<=`;
  the code is `<` at `:639`, which is the correct behaviour given `nextSeq` is exclusive. The prose
  is off by one, not the code.

---

### Checklist

| # | Criterion | Verdict |
|---|---|---|
| 1 | New frames, both idempotent | **Met** |
| 2 | `data`/`exit` only to subscribers; `recordPtyExit` + log once | **Met** |
| 3 | Catch-up reuses `replay`/`sinceSeq`; order pinned on the wire | **Met** |
| 4 | Subscriptions die with the connection; nothing leaks | **Met** |
| 5 | Exit reaches every subscriber | **Met** |
| T1 | Panel paired — panes subscribe, router refcounts, link re-subscribes | **Partially met** — panel-link reconnect missing |
| T2 | Subscribe-then-replay gap ordered and tested | **Met** |
| T3 | Exit between subscribe and replay delivered | **Met** |
| B | No `ready` capability, no gated send, version unchanged | **Met** |
| X1 | Older-Core gating deferred to #143 | Legitimately out of scope |
| X2 | No hold timeout | Acceptable as argued |
| + | Auto-subscribe on spawn | Defensible necessity; consequence on the shared Panel link unstated |

**Verdict: request changes.** One defect, one fix, in the layer that already has the pattern to copy
(`PanelLinkClient` re-sends its `subscribe` set on open — the pty set should ride beside it), plus a
test that drops and restores the browser socket with a pane claimed. Everything else here is
solid work and I would approve it the moment that path re-asks.

qcentic-dev:
> **Verdict: request changes.** One defect, one fix, in the layer that already has the pattern to copy

Fixed in 36329d5, in the layer you pointed at.

### The blocking defect — the panel-link now re-asks

`PanelLinkClient` holds the claimed set as `coreId → ptyId`s beside `watching`, and the `open`
handler re-sends `ptySubscribe` for it right after the `subscribe` loop — the pty set rides beside
the watch set, on the one handler that knows the socket came back. `corePtyBridgeFor`'s
`subscribe`/`unsubscribe` route through `link.ptySubscribe` / `link.ptyUnsubscribe` rather than a
bare `request`, so the set records itself where the claim is already being made; nothing above the
link changed, and nothing above it needs to notice a drop.

`catchUp` is off on the re-send, for the reason `resendPtySubscriptions` gives one layer down: a
hold is released by a `replay`, this is not the layer that sends one, and a hold nobody releases is
the same blank pane by another route. The panes replay off their own reconnect and
`pty-stream-router` holds whatever lands live while that replay is in flight, so going live
immediately loses nothing and can't reorder anything.

### The test

Two, both verified to fail with the re-send removed and pass with it.

`panel-link-client.test.ts` — a new *"the ptys a tab is rendering"* block pins the set itself:
re-asked on reconnect and without `catchUp`, **not** re-asked once a pane let go, asked once when
two panes in a tab render the same pty, kept apart per Core, a claim made while the link was down
carried onto the new one, and the pane painting again after the reconnect it re-asked on.

`terminals-in-browser.test.ts` — *"keeps a claimed pane live when the tab's own link drops and comes
back"* is the one you asked for, and it drives the shipped browser stack rather than hand-written
frames: real `PanelLinkClient` over the real panel-link ws, real `corePtyBridgeFor`, real
`createPtyStreamRouter`, against the real Core behind mTLS. A pane claims, paints, the tab goes
offline (the outage is held open through the socket factory rather than timed, so nothing races the
backoff), `ptySubscriptionCount()` is observed falling to **0** and coming back to **1**, and then a
fresh `core.emit` reaches the pane. Without the fix it stops at the count never returning.

You were right about why the existing e2e missed it: `"reattaches after a dropped link…"` opens a
fresh tab that never sends `ptySubscribe` at all and passes purely on the spawn auto-subscribe. It
is untouched — the new test is the one that closes the socket under a live client.

### The two smaller notes

- `router.ts` — the synthesized `ptySubscribeAck` keeps `holding: false` and now says why: the
  router answers before (or without) asking the Core, so it cannot know; it is nothing a tab can act
  on, since the tab's own `replay` releases whatever hold exists; and a tab that needed the truth
  would have to be told by the Core, which means not answering there.
- `pty-core-link-server.ts:625` — prose fixed, code untouched: `nextSeq` is the seq the *next* chunk
  will get and so is exclusive, a held chunk *below* it is already painted, and `nextSeq` itself
  goes out.

### Knowingly not in this commit

- **The secondary claim-loss** (`router.ts:195-197` — `claimPty` returns early when `this.cores` has
  no entry while `PanelLinkSession` has already recorded the claim). Real, and narrow as you say. It
  is left deliberately: this branch is scoped to the reconnect defect and its two notes, and the
  `bind`/`CoreState` change it wants is a different shape of fix than the one above. Happy to take
  it here if you'd rather it not wait.
- **Capability gating** — #143's, per your X1.
- **The auto-subscribe consequence on the shared service link** — the PR body wants that sentence;
  it is a description change, not a code one, and I have left the body alone rather than edit it
  under a review.

`pnpm typecheck` and `pnpm lint` are clean (0 errors; the warnings are the pre-existing ones).
Full `packages/core` suite passes; `packages/panel` passes, with the usual load-related timeouts in
unrelated files when the whole suite runs concurrently on one box — each passes in isolation. CI is
the arbiter.

Still a draft, as before.

qcentic-adm (approved):
## Re-review of 36329d5 — the reconnect path re-asks, and the test proves it on the wire

Checked the fix at the enforcing line, not the reply. It holds.

### 1. The blocking defect — closed, and each step of the old chain re-establishes

The set rides exactly where I asked for it: `PanelLinkClient.claimedPtys` (`panel-link-client.ts:104`),
beside `watching`, re-sent from the one handler that knows the socket came back
(`panel-link-client.ts:291-293` → `resendPtySubscriptions:364-381`).

Walking the eight steps I traced before, at head:

1-3. Unchanged and still correct: the socket dies, `session.detach()` (`router.ts:427-436`) gives
   back every claim, the refcount hits zero and `client.ptyUnsubscribe` deletes the Core-side
   subscription (`router.ts:207-219`). This is still the drop — it is now the recoverable one.
4. **Fixed.** The `open` handler re-sends `subscribe` for every watched Core *and then*
   `resendPtySubscriptions()`. The ordering matters and is right: `sendSubscribe` first, so the fresh
   `PanelLinkSession.watching` is populated before any pty frame — `PanelLinkRouter.push`
   (`router.ts:196-199`) only reaches a session that `watches(coreId)`, so a resend that ran first
   would re-subscribe a Core whose `data` frames still had nowhere to land.
5-7. Unchanged, and now correctly *don't* need to change: `pty-stream-router` still reattaches with a
   `replay` only and `claim()` is still not re-invoked. That was never the layer that should re-ask.
8. **Fixed.** The fresh `PanelLinkSession` no longer starts blind — the tab tells it. Chain to the
   Core, end to end: resend → `PanelLinkSession.receive` `ptySubscribe` (`router.ts:322`) →
   `claimPty` (`:360-369`, Set is empty on a fresh session, so it forwards) →
   `PanelLinkRouter.claimPty` (`:195-204`), `ptyClaims` 0→1, so `held > 0` is false →
   `source.client(coreId).ptySubscribe(ptyId)` on the shared core-link → the Core re-adds it to the
   live connection's `ptySubscriptions`. The `!state` early return at `:196-197` is not on this path:
   `this.cores` is populated by `bind` on first client and outlives the socket, and a Core that never
   bound could not have had a painting pane to begin with.

`catchUp: false` on the resend is the right call and for the reason given. I checked the buffer it
leans on rather than taking it: `corePtyBridge.onReconnect` (`core-pty-bridge.ts:148-159`) fires
synchronously from the `connectionListeners` loop at the end of the same `open` handler, and
`reattach` arms `held.set(ptyId, …)` (`pty-stream-router.ts:122`) synchronously *before* awaiting the
replay. Socket messages can only be dispatched in a later task, so there is no window in which a live
chunk can beat the hold into an unbuffered pane. A `catchUp: true` here would arm a Core-side hold
that this layer sends no `replay` to release — strictly worse.

Two follow-on paths I chased and found sound:

- **A claim made while the link was down** sends twice on the new socket: the resend
  (`catchUp: false`, `psubN`) and then the queued original (`catchUp: true`, `pN`, flushed at
  `:293`). Harmless in the right way — the session's own Set has already recorded the ptyId from the
  resend, so `claimPty:366` returns early and the `catchUp: true` never reaches the Core. No orphan
  hold. The resend's ack has no `pending` entry and is dropped at `onMessage:437`, also fine.
- **Two panes on one pty in one tab.** `ptySubscribe:205` dedupes, and `ptyUnsubscribe:219` only
  releases the recorded one — which lines up with `pty-stream-router`'s `claims` being last-wins and
  its unclaim being a no-op for a superseded surface (`:215`). Subscribe once, release when the
  active claimant lets go. No refcount skew introduced at this layer.

### 2. The test — bytes, not a spy

`terminals-in-browser.test.ts:"keeps a claimed pane live when the tab's own link drops and comes back"`
is the test I asked for, and stronger than the minimum. It drives the shipped stack (real
`PanelLinkClient` over the real panel-link ws, real `corePtyBridgeFor`, real `createPtyStreamRouter`)
against a real Core, holds the outage open through the socket factory instead of racing the backoff,
and pins the enforcing state on both sides:

- `core.server.ptySubscriptionCount()` observed **0** after the drop and **1** after the reconnect —
  the Core's own bookkeeping, not a re-subscribe method having been called;
- then a fresh `core.emit(ptyId, "after the tab came back\n")` asserted arriving at the pane's `data`
  handler. Live bytes, after the reconnect, through the subscription that had to be re-asked for.

Both assertions independently fail without `resendPtySubscriptions`: the count never returns, and the
pane's reattach `replay` has already resolved before the emit, so nothing paints it either. The unit
block in `panel-link-client.test.ts` pins the set itself — re-asked without `catchUp`, not re-asked
after a release, asked once for two panes, kept apart per Core — and asserts on decoded outgoing
frames rather than on the map. `"reattaches after a dropped link…"` is untouched, which is correct:
it was passing on the spawn auto-subscribe and was never the coverage.

### 3. The two smaller notes — both taken

- `router.ts:332-338` — the synthesized `holding: false` now says why the router cannot know and why
  `false` is the answer a tab can live with. I grepped: nothing in `packages/panel` outside tests
  reads the field, so the comment's claim is true today.
- `pty-core-link-server.ts:622-632` — prose now reads "`nextSeq` … the seq the *next* chunk will get,
  so it is exclusive", with `nextSeq` itself going out. That matches `chunk.seq < nextSeq` at `:639`.
  Code untouched, as stated.

### 4. Scope and CI

The delta is 7 files and nothing else: the client set, the bridge routing through it, the two
comments, and three test files. No `capability` field, no gated send, `CORE_LINK_PROTOCOL_VERSION`
still `0.15.0` — the only occurrences of "capability" across the whole PR are prose. CI is green on
36329d5 (run 31485940543, all 19 checks pass), and the new e2e file is inside the `Unit Tests` job's
`include` glob rather than sitting unrun.

### Non-blocking, for later

- The **secondary claim-loss** at `router.ts:195-197` is still open. Agreed it does not belong in this
  branch — it wants the `bind`/`CoreState` change and is a different shape of fix. Worth an issue so
  it doesn't decay into folklore; no need to take it here.
- `claimedPtys` is pruned only by `ptyUnsubscribe`, so a pty that **exits** while its pane stays
  mounted gets re-asked on every reconnect — the same residue I noted for
  `PtyCoreLinkClient.subscribedPtys`, with the same bound (pane count × connection lifetime) and the
  same harmlessness. Fixing both in one place later would be tidier than either alone.
- `panel-link-client.test.ts:"keeps a claim made while the link was down"` asserts
  `not.toHaveLength(0)` where the interesting fact is *which* frames go out and that the `catchUp:
  true` duplicate is absorbed by the session. Tightening it would document the behaviour I had to
  trace by hand above.

---

**Verdict: approve.** The defect is closed at the layer that owns the reconnect, the argument for
`catchUp: false` is backed by a buffer that is genuinely armed first, and the regression is pinned by
real bytes reaching a real pane over a socket that really dropped. The rest of the PR is what it was
when I called it solid work.

Issue #172, opened 2026-08-11
Panel link liveness: 15s/45s sweep on the Panel side, wake hook in the browser client


qcentic-dev:
**Base branch: `fix/panel-overnight-detach`, not `main`.** Train: **`beta/0.2.0`**.
Suggested branch: **`fix/panel-link-liveness`**.
Implements the diagnosis in #139 — one sentence: *the panel link gets the liveness contract the
core link already has.*

## What

The core link pings every 15s and terminates a peer silent for 45s, **at both ends**. The panel
link inherited that link's backoff and its cursor replay and **not** its heartbeat: neither end
detects a dead peer. A WebSocket that dies without a TCP FIN — the overnight fate of a flow
carrying zero bytes across a NAT, firewall or reverse proxy — leaves the tab believing it is
connected forever. `readyState` stays `OPEN`, no `close` event fires, and `close` is the only thing
in the browser client that starts a reconnect (`packages/panel/src/lib/panel-link-client.ts:244-258,
266-277`). Nothing in the app observes wake, focus, visibility or `online`, so there is no second
path back. Only a page reload recovers.

Everything downstream of a reconnect already works and is merely unreachable: re-subscribe from the
cursor (`panel-link-client.ts:231-240,284-294`), pty reattach with a replay window
(`pty-stream-router.ts:111-146`), refetch on the way back (`use-core-live-queries.ts:56-65`,
`use-fleet.ts:165-168`). Nothing here builds new recovery machinery. It builds the two things that
*start* the existing machinery.

**1. Server sweep — `packages/panel/src/server/panel-link/ws-server.ts`, in `bind()` (:113-129).**
A per-connection 15s ping / 45s terminate sweep mirroring
`packages/core/src/pty-core-link-server.ts:463-489`: `lastInboundAt` advanced by `message` and
`pong`, `setInterval` at the ping interval, `terminate()` (not `close()`) past the timeout, interval
cleared on `close`. This does the two missing things at once — it keeps the idle flow warm through
the reverse proxy `DEPLOY.md:55-58` requires and through Docker's NAT, and it reaps a corpse so the
browser's existing `close` → reconnect → re-subscribe → reattach path actually runs.

**2. Wake hook and last-inbound tracking — `packages/panel/src/lib/panel-link-client.ts`.**
Track last-inbound-at on every received frame. Add the wake hook the app has nowhere: on
`visibilitychange → visible`, `focus` and `online`, if the socket is not `OPEN`, or is `OPEN` but
[...]

Issue #144, opened 2026-08-11
Session write lock: claim, release, force takeover, mutation gating (D3-D7, D10)


mrfqcentic:
**Base branch: `feat/multi-connection-core`, not `main`.** Train: **`beta/0.2.0`**.
Suggested branch: **`feat/session-write-lock`**.
Implements **D3, D4, D5, D6, D7** and **D10** of #140. The centre of the effort.

## What

One Session, at most one writer. The writer is named by the **core-link connection** (D3) — not by
a per-client credential, and not by a logical actor above the socket.

- **Claim is explicit** (D6). A `claim` frame. A mutation without the lock is an error, never an
  implicit acquisition.
- **A Session starts unlocked and its creator gets no privilege** (D5). Unlocked is a real state
  and any connection may claim it.
- **The lock covers every mutation on that Session** (D4): `write`, `kill`, and every task mutation
  addressed at it. Not just `write` — an automation whose session can be killed by an observer does
  not hold a lock in any useful sense.
- **Not Core-wide** (D4). Different Sessions on one Core may be held by different clients at the
  same time; that is the ordinary case this effort exists to enable.
- **Ends on explicit release, connection drop, or force takeover. No idle timeout** (D7).
- **In memory, dies with the Core** (D12, owned by the first ticket).

## Done when

- `claim`, `release`, and a force takeover frame exist and are correlated by `reqId` like every
  other request.
- A mutation from a connection that does not hold the lock is refused with a distinguishable error
  — the client must be able to tell "locked by someone else" from "that session is gone".
- Two connections drive two different Sessions on one Core concurrently, with no interference.
- Dropping the holding connection releases its locks. All of them.
- Force takeover transfers the lock immediately; the previous holder's next mutation is refused.
- A test covers the D5 window: start a Session, do not claim, have a second connection claim it.

## Why no idle timeout

A long agent run is **idle by definition** — no mutations for many minutes is the success case, not
the abandoned one. An idle timeout drops the lock exactly when losing it costs most. Force takeover
covers the hung-client case instead. Do not add a timeout "just as a safety net"; it is the
opposite of one here.

## This is not a security boundary (D10)

The Core's only authentication is the bearer in the registration blob, and anyone holding it can
open a VM Shell Session and `kill` the process directly. The lock is an accident guard — the same
posture #129's F3 takes for path confinement. Do not harden it into something it is not, and do not
let a review treat a bypass as a vulnerability.

## Traps

[...]

PR #175, opened 2026-08-11
fix(panel-link): 15s/45s sweep on the Panel side, wake hook in the browser client

### LINKED ISSUE #139: A Panel left open overnight comes back detached — which link drops is unestablished

qcentic-dev:
Base is `fix/panel-overnight-detach`, not `main`. Train: `beta/0.2.0`.

Implements #172 — the panel link gets the liveness contract the core link already has. No new
recovery machinery: everything downstream of a reconnect (re-subscribe from the cursor, pty
reattach with a replay window, refetch on the way back) already worked and was merely unreachable.
This builds the two things that *start* it.

## What changed

**1. Server sweep — `packages/panel/src/server/panel-link/ws-server.ts`.** A per-connection 15s
ping / 45s terminate sweep. `lastInboundAt` advances on `message` and on `pong`; the interval is
cleared by *this* socket's `close` handler; the reap calls `terminate()`, because a graceful close
on a half-open socket waits for a peer that is never going to answer. The ping doubles as what
keeps the idle flow warm through the reverse proxy `DEPLOY.md` requires and through Docker's NAT.

Deliberately **without** the core link's `activeWs` guard: the Core holds exactly one connection,
the Panel holds one per tab, and the same line here would reap every tab but the newest. `ws.ping`
is guarded before use.

`bind()` became the exported `bindPanelLinkSocket()` so the sweep can be driven by a test with an
injected socket and a clock it can move; `attachPanelLink` is still the only production caller.

**2. Wake hook and last-inbound tracking — `packages/panel/src/lib/panel-link-client.ts`.** The
client tracks when a frame last arrived and, on `visibilitychange → visible`, `focus` and `online`,
drops a socket that is not `OPEN` — or is `OPEN` and has heard nothing inside the staleness window
— and dials again at once with `reconnectAttempt` reset. Listeners are removed by `close()`
alongside the reconnect timer.

The wake signals are the backbone, not a timer: a hidden tab's timers are throttled to roughly once
a minute and a frozen page's do not run at all. The staleness window is **15 minutes**, deliberately
not the server's 45s — the server's pings are answered below the `WebSocket` API and surface no
event to page JavaScript, so this end's clock advances only on *application* frames, and an idle
Core legitimately sends none for hours. One false-positive redial is cheap and safe (the reconnect
path is idempotent and replays by cursor); one on every focus of a quiet tab would not be.

**3. Tests for the case that has no event in it.** A socket that goes silent without firing
`close`, `readyState` left at `OPEN` — on both sides. 14 new cases (8 client, 6 server); both
existing suites pass unchanged.

One deliberate behaviour change beyond the two above: the client's `message` listener now ignores
frames from a socket it has already given up on. Those frames are not the current link's, and above
all they are not evidence that the current link is alive.

## Out of scope, untouched

The "reconnecting…" affordance, 401-on-upgrade handling, cursor persistence, the 2048-event buffer,
and everything from #140 (no registry, no fan-out change, no lock, no capability negotiation).
`PANEL_LINK_PROTOCOL_VERSION` does not move, and the diff stays inside the panel-link layer —
`ws-server.ts`, `panel-link-client.ts` and their tests — clear of #144's `packages/core` and
core-link client work.

## Why this does not close #139

#139 stays open. The overnight symptom is proven fixed only by an overnight run: leave a tab open
and record the morning after whether it comes back on its own, and whether the discriminator the
diagnosis names (a Core painted unreachable ⇒ core link, not this) ever appears.

Refs #172
Refs #139

### DIFF
--- a/packages/panel/src/lib/__tests__/panel-link-client.test.ts
+++ b/packages/panel/src/lib/__tests__/panel-link-client.test.ts
@@ -1,5 +1,9 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { PanelLinkClient, type PanelLinkSocketLike } from "../panel-link-client";
+import {
+  PanelLinkClient,
+  type PanelLinkOptions,
+  type PanelLinkSocketLike,
+} from "../panel-link-client";
 import type { PanelLinkClientFrame, PanelLinkServerFrame } from "~/shared/panel-link";
 import type { CoreLinkEvent } from "@actana/shared/core-link-frames";
 
@@ -266,6 +270,203 @@ describe("panel link · reconnect and replay", () => {
 
     expect(second.outgoing("core_a").map((f) => f.type)).toEqual(["tasksList"]);
     link.close();
+  });
+});
+
+/**
+ * A stand-in for `window` / `document`, so the wake signals can be fired from a
+ * test that runs in Node — and so "left no listener behind" is a number this
+ * suite can read rather than a promise the code makes.
+ */
+class FakeWakeTarget {
+  private readonly listeners = new Map<string, Set<() => void>>();
+
+  addEventListener(type: string, cb: () => void) {
+    const set = this.listeners.get(type) ?? new Set<() => void>();
+    set.add(cb);
+    this.listeners.set(type, set);
+  }
+  removeEventListener(type: string, cb: () => void) {
+    this.listeners.get(type)?.delete(cb);
+  }
+  fire(type: string) {
+    for (const cb of [...(this.listeners.get(type) ?? [])]) cb();
+  }
+  /** How many listeners are still installed, of any type. */
+  registered() {
+    let total = 0;
+    for (const set of this.listeners.values()) total += set.size;
+    return total;
+  }
+}
+
+class FakeDocument extends FakeWakeTarget {
+  visibilityState: "visible" | "hidden" = "visible";
+
+  /** The tab came back to the foreground. */
+  reveal() {
+    this.visibilityState = "visible";
+    this.fire("visibilitychange");
+  }
+  /** The tab went away. */
+  hide() {
+    this.visibilityState = "hidden";
+    this.fire("visibilitychange");
+  }
+}
+
+describe("panel link · a socket that dies without saying so", () => {
+  /**
+   * The bug this covers has no event in it. The socket stops delivering frames,
+   * `readyState` stays `OPEN`, `close` never fires, and a send into it does not
+   * throw — so the fake socket here simply *does nothing*, which is exactly the
+   * failure. What the tests drive is the only thing left: a wake signal.
+   */
+  let win: FakeWakeTarget;
+  let doc: FakeDocument;
+
+  beforeEach(() => {
+    win = new FakeWakeTarget();
+    doc = new FakeDocument();
+    vi.stubGlobal("window", win);
+    vi.stubGlobal("document", doc);
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  function staleClient(opts: PanelLinkOptions = {}) {
+    return new PanelLinkClient({
+      url: "ws://panel.test/panel-link?v=1",
+      createSocket: (url) => new FakeSocket(url),
+      reconnectInitialMs: 100,
+      reconnectMaxMs: 10_000,
+      requestTimeoutMs: 1_000,
+      staleAfterMs: 60_000,
+      ...opts,
+    });
+  }
+
+  it("drops a silent socket on the next wake and re-subscribes from its cursor", () => {
+    const link = staleClient();
+    link.watch("core_a");
+    const first = live();
+    first.push({ t: "core", coreId: "core_a", frame: { type: "event", event: event(9) } });
+
+    // Hours pass with the tab hidden. The socket says nothing and, crucially,
+    // does nothing: no close event, no readyState change, no error.
+    vi.advanceTimersByTime(4 * 60 * 60_000);
+    expect(FakeSocket.opened).toHaveLength(1);
+    expect(first.readyState).toBe(1);
+
+    doc.reveal();
+    const second = live();
+
+    expect(FakeSocket.opened).toHaveLength(2);
+    expect(second).not.toBe(first);
+    expect(second.outgoing("core_a")).toEqual([
+      { type: "subscribe", reqId: expect.any(String), lastEventId: 9 },
+    ]);
+    link.close();
+  });
+
+  it("wakes on focus and on coming back online, not only on visibility", () => {
+    for (const signal of ["focus", "online"] as const) {
+      FakeSocket.opened = [];
+      const link = staleClient();
+      live();
+      vi.advanceTimersByTime(90_000);
+
+      win.fire(signal);
+
+      expect(FakeSocket.opened, `${signal} should have redialled`).toHaveLength(2);
+      link.close();
+    }
+  });
+
+  it("leaves a healthy, recently-active link alone", () => {
+    const link = staleClient();
+    const socket = live();
+    socket.push({ t: "core", coreId: "core_a", frame: { type: "event", event: event(1) } });
+    vi.advanceTimersByTime(30_000);
+
+    doc.reveal();
+    win.fire("focus");
+    win.fire("online");
+
+    expect(FakeSocket.opened).toHaveLength(1);
+    link.close();
+  });
+
+  it("does not redial on the visibilitychange that hides the tab", () => {
+    const link = staleClient();
+    live();
+    vi.advanceTimersByTime(90_000);
+
+    doc.hide();
+
+    expect(FakeSocket.opened).toHaveLength(1);
+    link.close();
+  });
+
+  it("redials at once on wake rather than waiting out the backoff", () => {
+    const link = staleClient();
+    live().drop();
+    vi.advanceTimersByTime(100);
+    // Second dial, never accepted — the backoff is now walking upward.
+    FakeSocket.last!.drop();
+    expect(FakeSocket.opened).toHaveLength(2);
+
+    win.fire("focus");
+
+    // No timer advanced: the wake redials immediately, not on the next delay.
+    expect(FakeSocket.opened).toHaveLength(3);
+
+    // And the attempt counter went back to zero with it: the next ordinary
+    // drop retries on the initial delay, not on the doubled one it had reached.
+    FakeSocket.last!.drop();
+    vi.advanceTimersByTime(100);
+    expect(FakeSocket.opened).toHaveLength(4);
+    link.close();
+  });
+
+  it("fails in-flight requests when it drops a silent socket", async () => {
+    // A request timeout long enough to stay out of the way: it is not a
+    // liveness signal, and the point here is that the wake path fails the
+    // caller rather than leaving it to expire.
+    const link = staleClient({ requestTimeoutMs: 10 * 60_000 });
+    live();
+    const answer = link.request("core_a", { type: "tasksList" });
+    vi.advanceTimersByTime(90_000);
+
+    doc.reveal();
+
+    await expect(answer).rejects.toThrow("connection lost");
+    link.close();
+  });
+
+  it("leaves no wake listener behind when it is closed", () => {
+    const link = staleClient();
+    live();
+    expect(win.registered() + doc.registered()).toBe(3);
+
+    link.close();
+
+    expect(win.registered()).toBe(0);
+    expect(doc.registered()).toBe(0);
+  });
+
+  it("ignores a wake signal after close", () => {
+    const link = staleClient();
+    live();
+    link.close();
+    vi.advanceTimersByTime(90_000);
+
+    doc.reveal();
+    win.fire("focus");
+
+    expect(FakeSocket.opened).toHaveLength(1);
   });
 });
 
--- a/packages/panel/src/lib/panel-link-client.ts
+++ b/packages/panel/src/lib/panel-link-client.ts
@@ -46,11 +46,29 @@ export type PanelLinkOptions = {
   reconnectInitialMs?: number;
   reconnectMaxMs?: number;
   requestTimeoutMs?: number;
+  /** How long silence on an `OPEN` socket is tolerated at a wake signal. */
+  staleAfterMs?: number;
 };
 
 const DEFAULT_RECONNECT_INITIAL_MS = 500;
 const DEFAULT_RECONNECT_MAX_MS = 10_000;
 const DEFAULT_REQUEST_TIMEOUT_MS = 30_000;
+
+/**
+ * How long an `OPEN` socket may go without delivering a frame before a wake
+ * signal treats it as a corpse.
+ *
+ * Generous on purpose, and deliberately not the server's 45s timeout. The
+ * server's pings never reach page JavaScript — the browser answers them down in
+ * the network layer and the `WebSocket` API surfaces no pong event — so this
+ * clock advances only on *application* frames, and an idle Core legitimately
+ * sends none for hours. A window near the server's would therefore redial a
+ * perfectly healthy link on every focus of a quiet tab. The asymmetry is
+ * deliberate: one false-positive redial is cheap and safe, because the
+ * reconnect path is idempotent and replays from the cursor — a redial on every
+ * focus is not.
+ */
+const DEFAULT_STALE_AFTER_MS = 15 * 60_000;
 
 /**
  * A request frame minus the `reqId` the client assigns. Distributed over the
@@ -83,12 +101,16 @@ export class PanelLinkClient {
   private readonly reconnectInitialMs: number;
   private readonly reconnectMaxMs: number;
   private readonly requestTimeoutMs: number;
+  private readonly staleAfterMs: number;
 
   private socket: PanelLinkSocketLike | null = null;
   private open = false;
   private closed = false;
   private reconnectAttempt = 0;
   private reconnectTimer: ReturnType<typeof setTimeout> | null = null;
+  /** When this tab last had evidence of life: an application frame, or an open. */
+  private lastInboundAt = Date.now();
+  private readonly detachWakeListeners: () => void;
 
   private reqSeq = 0;
   private readonly pending = new Map<string, Pending>();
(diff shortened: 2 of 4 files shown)


qcentic-dev:
## #172 · Done when

PR **#175** — https://github.com/actana/control/pull/175 (draft, base `fix/panel-overnight-detach`)

- [x] **A panel-link connection that stops answering is terminated by the Panel within the timeout,
  and `PanelLinkSession.detach()` runs as a result.** `ws-server.test.ts` → *"terminates a
  connection that stops answering, and the session lets go"*: the socket subscribes to a Core,
  answers nothing, and is terminated with its watch set released. Stated exactly: detection is on
  the first 15s tick past the 45s silence, so the reap lands at 60s — the same granularity the core
  link has always had, since this mirrors its clock rather than inventing a second one.
- [x] **A quiet-but-healthy tab is never reaped.** *"never reaps a quiet-but-healthy connection,
  however long it stays quiet"*: four hours, zero application frames, ~960 pings answered by pong,
  no termination and the watch set intact. A companion case, *"takes a connection's own frames as
  proof of life"*, covers the same for a tab that speaks instead of ponging.
- [x] **Terminating one tab's connection leaves every other tab's untouched.** *"reaps one tab's
  connection without touching any other tab's"*: healthy connections bound both **before and after**
  the dead one, so a sweep carrying the core link's `activeWs` guard would fail this test in either
  direction. Each survivor keeps its own sweep, session and watch set.
- [x] **A socket that stops delivering frames and never fires `close` is detected on the next wake
  signal; the client drops it, reconnects at once, and re-subscribes each watched Core from its
  existing cursor.** `panel-link-client.test.ts` → *"drops a silent socket on the next wake and
  re-subscribes from its cursor"*: the fake socket does literally nothing for four hours — no
  `close` event, `readyState` asserted still `OPEN` (`1`) at the moment of the wake — and the
  redial's first frame is `subscribe` with `lastEventId: 9`, the cursor it had reached. A second
  case proves `focus` and `online` wake it too, not only `visibilitychange`.
- [x] **That reconnect does not wait out the backoff; `reconnectAttempt` is reset.** *"redials at
  once on wake rather than waiting out the backoff"*: after two failed dials have walked the delay
  up, a `focus` opens the next socket with **no timer advanced at all**, and the drop after it
  retries on the *initial* delay — which only holds if the counter went back to zero.
- [x] **A wake signal on a healthy, recently-active link does not redial.** *"leaves a healthy,
  recently-active link alone"*: all three signals fired, still one socket. Plus *"does not redial on
  the visibilitychange that hides the tab"* and *"ignores a wake signal after close"*.
- [x] **`close()` leaves no `visibilitychange`, `focus` or `online` listener behind.** *"leaves no
  wake listener behind when it is closed"*: 3 listeners installed, 0 after `close()` — counted off
  the stand-in `window`/`document`, so it is a number the suite reads rather than a promise the code
  makes.
- [x] **Existing panel-link behaviour is otherwise unchanged** — upgrade gate, frame vocabulary,
  replay contract and the 2048-event buffer untouched; `PANEL_LINK_PROTOCOL_VERSION` does not move
  (`shared/panel-link` is not in the diff at all). **One deliberate exception, called out rather
  than buried:** the client's `message` listener now ignores frames arriving on a socket it has
  already abandoned. A socket the client gave up on may still deliver what it had buffered, and
  those frames must not re-fire listeners for the current link — still less pass as evidence that
  the current link is alive. It sits under the same identity check the `open` and `close` handlers
  already carried.
- [x] **Both existing suites still pass unchanged — additions, not rewrites.** 1274/1274 pass across
  125 files (`pnpm --filter @actana/panel test`); typecheck clean; `pnpm lint` reports 0 errors and
  no warning in any touched file. No existing test case was edited. Two mechanical lines in
  `ws-server.test.ts` did change and are worth naming: the import now also pulls in
  `bindPanelLinkSocket`, and `attachPanelLink(server)`'s return value is captured instead of
  discarded. Every pre-existing assertion is byte-for-byte as it was.

### Not done, deliberately

- **No client-side timer complementing the wake signals.** The issue permits one (*"a timer **may**
  complement them"*) and requires none. Leaving it out keeps the design honest about what it depends
  on: the signals a throttled or frozen page can still deliver. The residual gap is a foreground tab
  whose path goes half-open and that is never blurred, focused or re-shown — it waits for the next
  wake signal.
- **#139 is not closed** — referenced with `Refs`, on both commits and in the PR body. The overnight
  symptom is proven fixed only by an overnight run: leave a tab open and record the morning after
  whether it comes back on its own, and whether the discriminator the diagnosis names (a Core
  painted unreachable ⇒ core link, not this) ever appears.
- **Out of scope and untouched:** the "reconnecting…" affordance, 401-on-upgrade handling, cursor
  persistence, the 2048-event buffer, and everything from #140 — no registry, fan-out, lock or
  capability work. The diff is four files, all in the panel-link layer, clear of #144's
  `packages/core` and core-link client work.

qcentic-adm (approved):
Reviewed against #172 — every criterion under **Done when**, its **Traps**, and its
**Deliberately out of scope** list — reading the enforcing lines rather than the description.
All four files read in full at `07609dd`; CI green on that head, both installer E2E jobs included.

**Verdict: approve.** Every criterion is met. Two non-blocking notes at the end; neither
changes the verdict.

---

## The sweep — `packages/panel/src/server/panel-link/ws-server.ts`

Per-connection, and provably so: `lastInboundAt` is a `let` inside `bindPanelLinkSocket`
(`:179`), the interval is a local (`:181-193`), and everything that touches either is closed
over by that one call. 15s/45s at `:122-123`. `lastInboundAt` advances on `pong` (`:215-217`)
**and** on `message` (`:218-221`) — both, as the core link does. The reap is
`ws.terminate ? ws.terminate() : ws.close()` inside a `try` (`:198-205`), not a graceful close.
`stopSweep()` is called from this socket's own `close` (`:222-225`) and `error` (`:226-229`)
handlers and from nowhere else — no shared registry, no cross-connection bookkeeping. `ws.ping`
is guarded twice over: the sweep is only armed `if (ws.ping)` (`:193`) and the call itself is
`ws.ping?.()` in a `try` (`:207-211`).

Most importantly, **the `activeWs` guard is absent**. Compare `packages/core/src/pty-core-link-server.ts:466-469`,
which this otherwise mirrors line for line: that `if (this.activeWs !== ws)` has no counterpart
here, and there is no `activeWs`-equivalent in this file to compare against. With one connection
per tab it would have reaped every tab but the newest, which is exactly what
*"reaps one tab's connection without touching any other tab's"* pins — and it pins it in the
form that matters, with healthy connections bound **before and after** the dead one, so the
guard would fail the test from either direction.

Reaping a corpse releases the watch set through the real path, not a shortcut: `terminate()`
fires `close`, `close` calls `session.detach()` (`:224`), and `detach()` clears `watching` and
calls `router.forget(this)` (`router.ts:291-296`). *"terminates a connection that stops
answering, and the session lets go"* asserts `session.watches("core_a")` flips true → false, so
the release is observed at the router's own accessor rather than inferred. And the quiet-healthy
case is not a maybe: four hours, ~960 pings, every one answered by pong, zero application frames,
`terminated === 0` and the watch set intact.

## The wake hook — `packages/panel/src/lib/panel-link-client.ts`

All three signals, correctly discriminated: `visibilitychange` filtered to
`visibilityState === "visible"` (`:257-259`), `focus` and `online` direct (`:261-262`), all
removed by the returned detacher that `close()` invokes at `:228`. The count is asserted, not
promised — 3 installed, 0 after `close()`.

`wake()` (`:275-279`) redials unless there is positive evidence of life: not-`OPEN` redials,
`OPEN`-but-stale redials, and only a recent inbound frame buys a pass. `redialNow()` (`:287-310`)
clears any pending backoff timer, sets `reconnectAttempt = 0` (`:292`), nulls `this.socket`
*before* closing the dying one so that socket's own close handler bails on the identity check at
`:355` rather than racing in a second reconnect, then calls `this.connect()` **directly** — not
`scheduleReconnect()`.

On the reset reaching zero rather than skipping a step: the test proves it by consequence.
After the wake-dial it drops the new socket and advances exactly `reconnectInitialMs` (100ms),
and a fourth dial appears. That only holds if the counter is 0 when `scheduleReconnect` computes
`initial * 2^attempt` (`:374-377`); at 1 the delay would be 200ms and the assertion would fail.

## The staleness window — the part worth the most scrutiny

The trap is real and the client genuinely cannot see through it: `lastInboundAt` advances only at
construction (`:112`), on `open` (`:338`) and in `onMessage` (`:425`) — application frames only,
because the browser answers the server's pings below the API and `PanelLinkSocketLike` (`:31-39`)
has nowhere to surface a pong. So work out what actually happens on focus of a tab whose Core has
been legitimately silent:

- **Visible tab, fleet view mounted.** `use-fleet.ts:177` polls `bridge.listTasks` over this very
  link every 15s, and each answer is an inbound frame. The clock never approaches 15 minutes; a
  focus costs nothing.
- **Hidden but not frozen.** Those same timers still run, throttled to roughly once a minute —
  still two orders of magnitude inside the window. No redial on return.
- **Frozen page, or a tab parked on a view that issues no periodic RPC.** The clock does go stale,
  and the first wake after 15 minutes of true silence redials a link that may have been perfectly
  healthy.

So it is a false positive on that third shape — but it is *one* redial per wake-after-quiet, not
one per focus, and the cost is what #172 says it is: `connect()` → `open` → re-subscribe from the
existing cursor (`:341`), and `pty-stream-router.ts:110-140` reattaches from `delivered + 1`,
firing `reset()` only when the Core's ring rolled past — which a link that carried no traffic has
not. The refetch burst on `onConnectionChange(true)` is roughly what `refetchOnWindowFocus` would
have cost anyway.

Worth stating plainly: the only way to remove that false positive entirely is an application-level
probe on wake, and that means a frame the vocabulary does not have — which #172 forbids in the same
breath (*"the frame vocabulary … untouched"*). Within the scope the issue draws, a generous window
is the only instrument available, and 15 minutes — 20× the server's reap and comfortably past any
throttled-timer gap — is the right end of "generous". The needle is threaded as far as the ticket
allows it to be, and the reasoning is written down at `:57-71` where the next reader will find it.

## The silent-socket test

`panel-link-client.test.ts` → *"drops a silent socket on the next wake and re-subscribes from its
cursor"* is the exact overnight shape: the fake socket does **nothing** for four hours — no
`drop()`, no `close`, and `expect(first.readyState).toBe(1)` asserts `OPEN` at the moment of the
wake, so the test would still pass a design that merely watched `readyState`, and fails one that
needs a close event. Recovery is asserted end to end: a second socket exists, it is not the first,
and its first Core frame is `subscribe` with `lastEventId: 9` — the cursor the tab had actually
reached, not 0.

## The deliberate exception — ignoring frames from an abandoned socket (`:346-352`)

Correct hardening, and load-bearing for the feature rather than incidental to it. Without the
guard, a frame the abandoned socket had buffered would reach `onMessage` and set
`this.lastInboundAt = Date.now()` (`:425`) — re-arming the staleness clock of the *new* link on
evidence from a corpse, which is precisely the inference this PR exists to prevent.

It does not need its own ticket, for three reasons that the code lets you check:

1. It is the same identity check `open` (`:335`) and `close` (`:355`) already carried; the
   `message` listener was the one that had been left out.
2. Nothing is lost. A response frame on the abandoned socket has no waiter — `redialNow()` calls
   `failPending` at `:307` first. An `event` frame is dropped *without* advancing the cursor
   (`:434-441` is never reached), so the new link's subscribe asks from the older cursor and the
   router replays it out of its buffer.
3. Its blast radius is almost entirely the new path. Before this PR, `this.socket !== socket`
   at a `message` could essentially only arise after that socket fired `close` — and a socket that
   has fired `close` delivers no further messages. The case it changes is the one `redialNow()`
   invented: a socket abandoned while still `OPEN`.

## Scope and containment

Four files, all in the panel-link layer. `shared/panel-link.ts` is not in the diff at all, so
`PANEL_LINK_PROTOCOL_VERSION` (still `1`) and the frame vocabulary cannot have moved; `router.ts`
is untouched, so the replay contract and the 2048-event buffer are as they were; the upgrade gate
in `ws-server.ts:20-110` is unchanged apart from the callback now naming `bindPanelLinkSocket`.
Nothing from #140 — no registry, no fan-out change, no lock, no capability negotiation — and
nothing in `packages/core`. #139 is carried as `Refs` on both commits and in the PR body, and is
not closed by this merge; the reasoning for that is stated where the next reader needs it.

Exporting `bind()` as `bindPanelLinkSocket()` with a structural `PanelLinkServerSocket` is a
widened surface, but a justified one: the sweep's whole subject is a socket that produces no
events, which cannot be driven through a real upgrade, and `attachPanelLink` remains the only
production caller.

---

## Two non-blocking notes

**1. A single tab switch fires two wake signals, and the second aborts the first's dial.**
`panel-link-client.ts:275-279` — switching to a tab fires `visibilitychange`→visible and then
`focus`. On a stale link the first `wake()` redials, leaving the new socket `CONNECTING` and
`this.open === false`; the second `wake()` therefore fails the `this.open &&` test at `:277` and
calls `redialNow()` again, closing the in-flight handshake and dialling a third socket. Recovery
still happens and the identity guards keep the client's state coherent — the cost is one aborted
upgrade, plus a session attach/detach on the service, per wake. If you want it gone, the cheapest
form is a `dialingSince` stamp consulted alongside `this.open`. Not worth a round trip on its own.

**2. The accounting says two mechanical lines; there are three, plus one in the other suite.**
`ws-server.test.ts:1` also changed (`afterEach`, `beforeEach`, `vi` added to the `vitest` import),
and `panel-link-client.test.ts:2-6` grew a `type PanelLinkOptions` import. Both are pure harness
imports for the added blocks. The substantive claim holds exactly as written — no existing case
was edited and every pre-existing assertion is byte-for-byte as it was — so this is a note on the
tally, not on the work.

*(One observation rather than a request: the core link logs `core-link.connection.stale` when it
reaps (`pty-core-link-server.ts:472`) and this sweep reaps silently. Since #139 is to be closed on
evidence from an overnight run, a line there would be the thing you read the next morning. The
panel-link module imports no logger at all today, so adding one is a scope call I would not make
inside this ticket.)*

---

## Checklist

| # | Criterion | Verdict |
|---|---|---|
| 1 | Reaped within the timeout, `PanelLinkSession.detach()` runs as a result | **Met** — `ws-server.ts:195-205`, `:222-225`, `router.ts:291-296`; reap lands on the first tick past 45s, i.e. 60s, the same granularity the core link has always had |
| 2 | Quiet-but-healthy tab never reaped | **Met** — pong advances `lastInboundAt` (`:215-217`); 4h / ~960 pings / zero application frames, still alive |
| 3 | One tab's termination leaves other tabs untouched | **Met** — no `activeWs` guard; per-connection closure; pinned with healthy sockets bound either side of the dead one |
| 4 | `ws.ping` guarded, `terminate()` not `close()` | **Met** — `:193`, `:207`, `:201` |
| 5 | Silent socket detected on the next wake; drop, redial, re-subscribe from cursor | **Met** — `:275-310`; asserted with `readyState` left `OPEN` and no `close` event, `lastEventId: 9` on the new socket |
| 6 | Redial does not wait out the backoff; `reconnectAttempt` reset | **Met** — `:292` sets 0 and `connect()` is called directly; the reset is proved to reach zero by the 100ms retry that follows |
| 7 | Wake on a healthy, recently-active link does not redial | **Met** — `:277`; all three signals fired, one socket |
| 8 | `close()` leaves no `visibilitychange` / `focus` / `online` listener | **Met** — `:254-268`, `:228`; 3 → 0, counted |
| 9 | Staleness window generous enough not to redial on every focus of an idle tab | **Met** — 15 min at `:71`; the false positive is confined to a frozen or RPC-free tab, once per wake, and the alternative is a frame the ticket forbids |
| 10 | Protocol version, upgrade gate, frame vocabulary, replay contract, 2048-event buffer untouched | **Met** — `shared/panel-link.ts` and `router.ts` not in the diff |
| 11 | Existing suites pass as additions, not rewrites | **Met** — no existing case edited; three mechanical import/capture lines, one more than the tally names |
| 12 | Nothing from #140; #139 referenced, not closed | **Met** — four panel-link files, `Refs #139` on both commits and in the body |
| 13 | CI green on the head | **Met** — all checks pass at `07609dd`, including both installer E2E jobs |

Issue #146, opened 2026-08-11
Stable client id: reclaim a reconnecting client's locks, reaping only (D9)


mrfqcentic:
**Base branch: `feat/multi-connection-core`, not `main`.** Train: **`beta/0.2.0`**.
Suggested branch: **`feat/core-link-client-id-reclaim`**.
Implements **D9** of #140.

## What

A connection presents a stable client id. It is used **only** for reaping — not for authority, and
it is never verified.

Removing evict-on-connect (D1) removes the thing that currently cleans up after a reloaded Panel
tab. Without a replacement, the dead socket lives for up to `HEARTBEAT_TIMEOUT_MS` (45s,
`pty-core-link-server.ts`) still holding its Session locks. A reconnecting connection presenting
the same client id replaces its predecessor and inherits its locks.

Forty-five seconds of a ghost holding your session is a daily-frequency event, and the alternative
— telling operators to force-take after every reload — trains the reflex that D7's force takeover
exists to keep rare.

## Done when

- Reconnecting with the same client id closes the predecessor and transfers its locks, in one step,
  with no window in which the Session is unlocked and claimable by a third party.
- Reconnecting with a *different* client id does not disturb anybody.
- The id is stable across a reload and distinct per Core client — the Panel's id is not the CLI's.
- Nothing anywhere treats it as authentication. A test asserting that a forged id grants no
  authority is worth more than the feature.

## Traps

- **This is not identity** (D10). If a review starts asking how to stop a client from claiming
  someone else's id, the answer is that the bearer holder already has a shell on the machine — not
  that the id needs signing.
- **The transfer must be atomic.** Release-then-reclaim opens exactly the window this ticket exists
  to close.
- **Where the id comes from is per client.** The Panel can persist one per browser tab; the CLI
  mints one per invocation. Do not put it in the registration blob — the blob is shared by every
  client on the machine, which would make them all the same connection.

Refs #140

qcentic-dev:
Delivered in PR #184, squash-merged into `beta/0.2.0` as 57d43c906331a1daac84d81282a76df8bf4cfd4c. The PR base is non-default, so closing keywords did not fire automatically; closing manually. Epic #140 stays open.

PR #176, opened 2026-08-11
feat(core-link): one Core client holds a Session's write lock — claim, release, force takeover (D3-D7, D10)

### LINKED ISSUE #144: Session write lock: claim, release, force takeover, mutation gating (D3-D7, D10)

qcentic-dev:
Implements **D3, D4, D5, D6, D7** and **D10** of #140 — the centre of the effort.

One Session, at most one writer. The writer is named by the **core-link connection** (D3).

**Base is `feat/multi-connection-core`, not `main`.** Train: `beta/0.2.0`.

---

## ⚠️ Needs a decision before this merges: VM Shell Sessions

ADR 0024 records, under its consequences, that *"VM Shell Sessions are assumed to follow the same
rule as harness Sessions. Not separately decided — the first ticket that touches one should confirm
or amend D4."*

**This ticket touches one.** `write` and `kill` name a `ptyId`, and the lock is keyed by the
Session, so the Core resolves the `ptyId` to the Task it was spawned for — **for every PTY, whatever
kind it is**. A VM Shell Session is therefore claimable and gated exactly like a harness Session:
its `taskId` can be claimed, and another connection's `write`/`kill` against it is refused.
`packages/core/src/__tests__/core-link-session-lock.test.ts` pins that behaviour under
*"a VM Shell Session is locked by the same rule"*.

That is the assumption ADR 0024 already carries, implemented as written — **not** a decision made
here. It is raised in this description rather than settled in a comment, per the #107 rule. Two
ways to close it, both outside this PR:

- **Confirm it** — the assumption stands, and the consequence bullet in ADR 0024 can be reworded
  from "assumed" to "decided" by whoever owns #140.
- **Amend D4 in #140** to exclude VM Shell Sessions — in which case `taskIdForPty` should skip
  `shellSession` PTYs the way `findByTask` already does, which is a one-line change plus its test.

Project-scoped user shells are unaffected in practice: they spawn with a synthetic `ut-…` taskId of
their own, so nothing a user terminal does can collide with a harness Session's lock.

---

## What landed

**`claim` / `release` / `forceTakeover`** (`packages/shared/src/core-link-frames.ts`) — ordinary
`reqId`-correlated requests, each naming a Session by its `taskId`. That is the one identifier
`tasksMutate` carries directly and `write`/`kill` resolve to from their `ptyId`; keying on the
process would give one Session as many locks as it has had PTYs.

**`SessionLockTable`** (`packages/core/src/session-lock-table.ts`) — the whole of the state: a map
from a Session to the connection holding it. Core-wide as a table, scoped to a Session as a lock
(D4), so different Sessions on one Core may be held by different clients at the same time. In
memory, dying with the Core (D12).

**The gate** (`packages/core/src/pty-core-link-server.ts`) — on `write`, `kill`, and every
`tasksMutate` addressed at a Session. `create` is not gated: it names no existing Session, and an
automation holding one Session must still be able to start another.

**The Panel client** (`packages/panel/src/server/core-link/client.ts`) — the three frames joined
`MULTI_CONNECTION_ONLY_FRAME_TYPES`, the exact-membership test moved with them, and the three
methods consult the same gate themselves rather than falling through to the backstop rejection.

## Three things this deliberately does *not* do

**No idle timeout, and no safety net that amounts to one** (D7). A long agent run is idle by
definition — no mutations for many minutes is the success case, not the abandoned one — so a
timeout would drop the lock exactly when losing it costs most. The three endings are release, drop,
and force takeover; there is no fourth, and the module comment says so where someone would go to
add one. A test holds a lock across 30 simulated minutes of silence.

**No creator privilege, so the claim race stays open** (D5). Nothing calls into the lock table on
spawn. A Session starts unlocked, and a second connection can claim a Session the first started —
tested directly, including the creator *losing* the race. The window is real and accepted; closing
it means reintroducing creator privilege, which was decided against.

**No gate inside `PtyCore`.** `PtyCore.kill` has callers inside the Core — the PTY exit paths and
the task writer — that hold no lock and are nobody's client. Gating it would have the Core start
refusing itself, so the gate is on the client-facing frame and `PtyCore` gains only a read,
`taskIdForPty`. A test calls `PtyCore.kill` directly on a Session another connection holds and
asserts it still works.

## The distinguishable error

The ticket asks that a client be able to tell *"locked by someone else"* from *"that Session is
gone"*. They are different frames, so no prose is parsed:

| | answer |
| --- | --- |
| Another connection holds it | `error` with `code: "session-locked"` |
| That Session is gone | `writeResult { ok: false }` / `killResult { ok: false }` / `tasksMutateResult { task: null }` |

Only the first is worth retrying after a claim. `code` is a new optional field on the `error` frame,
absent on every error that shipped before it.

## An unlocked Session is writable by anybody

The gate refuses only when **another** connection holds the Session. D6's *"a mutation without the
lock is an error"* governs acquisition — a mutation never becomes a claim — not the unclaimed case.
D11 promises an old client *"never claims, and never notices it is no longer evicting anybody"*, and
a client that never claims has to be able to write; refusing unlocked Sessions would make the
capability's absence yield something *lesser* than today rather than exactly today, which the D11
rule amendment does not allow. Tested as its own case.

## Not a security boundary (D10)

Stated in the code where it matters — the module comment on `SessionLockTable`, and the wire
commentary on the frames. Anyone holding the bearer can open a VM Shell Session and kill the process
directly, so a lock defended against someone with a shell on the machine is theatre. This is an
accident guard; please do not review a bypass as a vulnerability, and please do not harden it into
something it is not.

## `resize` is not gated — flagging the reading

D4 names `write`, `kill`, and every task mutation addressed at the Session. A resize is none of
those: it is how a client tells the PTY the size of the viewport it is painting into, and a Reader
has one of those too. A Reader whose resize were refused would render the Session reflowed to
somebody else's terminal, which is worse than the interference gating it would prevent. Recorded in
ADR 0024's consequences; widening D4 to cover it would be an amendment to #140, not a call for this
ticket.

## Protocol version

`CORE_LINK_PROTOCOL_VERSION` does not move (D11) — the third entry in its log that does not, after
#142 and #143. The three frames are gated on the capability so they never reach a Core with no lock
table, and the gating can only refuse a Session some connection has explicitly claimed.

## Tests

`packages/core/src/__tests__/core-link-session-lock.test.ts` (23 cases) is organised around the
ticket's *Done when*, because each item there is a property rather than a code path. The Panel-side
gating and the exact-membership test live in
`packages/panel/src/server/core-link/__tests__/multiconnection-capability.test.ts`.

`pnpm typecheck` clean; `pnpm lint` 0 errors (53 pre-existing warnings, none in the changed files);
core 739/739, shared 226/226, panel 1310/1311 — the single panel failure varies run to run, passes
in isolation, and reproduces on the base branch without these changes.

## Boundaries respected

Nothing here touches `ws-server.ts` or `panel-link-client.ts` (#172's ground). Six panel test files
gained a one-line `taskIdForPty` to their `PtyCore` stand-ins, which is the whole of the panel-side
churn outside the core-link client.

Refs #140
Closes #144


qcentic-dev:
## Issue #144 — "Done when", criterion by criterion

- [x] **`claim`, `release`, and a force takeover frame exist and are correlated by `reqId` like every other request.**
  `claim` / `release` / `forceTakeover` in `CoreLinkRequestFrame`, answered by `claimResult` /
  `releaseResult` / `forceTakeoverResult`, each carrying the `reqId` it was asked with and the
  `taskId` it names. Pinned by *"answers each one on the reqId it was asked with"*, which sends all
  three and reads the answers back by request id rather than by frame order.

- [x] **A mutation from a connection that does not hold the lock is refused with a distinguishable error — the client must be able to tell "locked by someone else" from "that session is gone".**
  A refusal is an `error` carrying `code: "session-locked"`; a Session this Core no longer has is
  the unchanged `writeResult { ok: false }` / `killResult { ok: false }` /
  `tasksMutateResult { task: null }`. Different frames, so nothing parses prose. Two tests:
  *"refuses write, kill and task mutations with a distinguishable code"* (which also asserts nothing
  reached the Core — refused, not merely reported as failed) and *"tells 'locked by someone else'
  apart from 'that Session is gone'"*.

- [x] **Two connections drive two different Sessions on one Core concurrently, with no interference.**
  *"lets each drive its own with no interference (D4 — not Core-wide)"* — a CLI and a Panel each
  claim and write their own Session, both succeed, and each is refused on the other's. The write log
  is asserted exactly, so a stray keystroke could not pass unnoticed.

- [x] **Dropping the holding connection releases its locks. All of them.**
  `dropConnection` sweeps the table with `releaseAll(conn)` — every entry whose holder is that
  connection, not a list kept alongside it. Unconditional, and placed before the registry identity
  guard, so a socket already replaced in the map still gives its locks up. Three tests: all three of
  one connection's locks released and re-writable by another; a second connection's lock untouched
  by the first's drop; and the heartbeat-reaped case, which is the stale connection ADR 0024 D7
  calls out and which lands on the same `close` path.

- [x] **Force takeover transfers the lock immediately; the previous holder's next mutation is refused.**
  *"transfers the lock immediately and refuses the previous holder's next mutation"* — the very next
  `write` from the previous holder is refused with `session-locked`, and the winner's succeeds, with
  nothing in between.

- [x] **A test covers the D5 window: start a Session, do not claim, have a second connection claim it.**
  *"lets a second connection claim a Session the first started and did not claim"* — it also asserts
  `sessionLockCount() === 0` right after the spawn, which is the D5 property itself rather than its
  consequence, and then that the creator's own write is refused. A second case has the creator
  *lose* the race outright.

**Every criterion is checked. Nothing under "Done when" is unchecked.**

## Traps, restated and honoured

- [x] **The claim race is real and ACCEPTED.** Nothing calls into the lock table on spawn; no
  creator privilege was reintroduced. The window is tested as a property, not worked around.
- [x] **`kill`'s non-Session callers.** The gate is on the client-facing frame only. `PtyCore.kill`
  is untouched; `PtyCore` gained one read (`taskIdForPty`). *"leaves PtyCore's own kill ungated for
  its non-Session callers"* calls it directly on a held Session and asserts it still works.
  `killLaunchProcesses` is ungated too — it is addressed at a folder, not a Session — and tested.
- [x] **No idle timeout.** None added, none as a safety net. A test holds a lock across 30 simulated
  minutes of silence, past every timer in the server including the 45s heartbeat timeout.
- [x] **D10 is an accident guard, not a security boundary.** Said in the module comment on
  `SessionLockTable` and in the wire commentary, i.e. where someone would go to harden it.
- [x] **The two integration facts from the base branch.** `claim` / `release` / `forceTakeover`
  added to `MULTI_CONNECTION_ONLY_FRAME_TYPES`, the exact-membership test updated to name all five,
  and the Panel client's three new methods consult the same gate. Locks die with their connection in
  `dropConnection`, covering **all** of that connection's locks.

## The one thing left open — not a "Done when" item

- [ ] **VM Shell Sessions under D4.** This change touches one: `write`/`kill` resolve a `ptyId` to
  its Task for every PTY, so a VM Shell Session is claimable and gated exactly like a harness
  Session. That is the assumption ADR 0024 already carries, implemented as written — **not** a
  decision taken here. It is raised in the PR description for confirmation, or for D4 to be amended
  in #140. It is deliberately not settled in this PR or in a comment on it (#107 rule), and it is
  unchecked here because it is not mine to check.

## Also flagged, not decided here

`resize` is **not** gated. D4 enumerates `write`, `kill`, and task mutations addressed at the
Session; a resize is how a client tells the PTY the size of the viewport it paints into, and a
Reader has one too. Recorded in ADR 0024's consequences with the reasoning; widening D4 to cover it
would be an amendment to #140.

---

**PR:** #176 — https://github.com/actana/control/pull/176 (draft, base `feat/multi-connection-core`)

qcentic-adm (changes requested):
Reviewed against #144's *Done when*, one criterion at a time, reading the enforcing lines rather than this description. Every one of the seven passes, and the two design points you raised rather than settled are answered below. Three things to fix first — all small, and two of them are "record the reading", not "change the behaviour".

## The ruling you asked for: an unlocked Session is writable by anybody

**Your reading of D6 holds.** Ship it as implemented.

D6 says *"a mutation without the lock is an error, never an implicit acquisition"* — and its second clause names what the first is guarding: acquisition. The rejected alternative recorded right there is first-mutation-claims, not mutation-on-unclaimed. Read the other way, D6 collides head-on with D11, which promises an old client *"works untouched — it never claims, and never notices it is no longer evicting anybody"*: a client that never claims and cannot write is not untouched, it is dead. #140's own rule amendment settles the tie — an additive capability may announce itself on `ready` *"only where its absence yields today's behaviour exactly, not a lesser one"*. Refusing unlocked Sessions yields a lesser one, so the strict reading is the one the map forbids. D5's *"unlocked is a real state"* and the accepted claim race point the same way: a race for a Session nobody may write is not a race worth accepting.

`session-lock-table.ts:74-97` (`mayMutate`) and `pty-core-link-server.ts:842-866` implement exactly that, and it is tested as its own case ("serves a mutation on an unlocked Session"). Recording it in ADR 0024's consequences rather than in a comment is the right shape under the #107 rule.

One follow-up for whoever owns #140, not for this PR: D6's wording is what admitted the ambiguity. The ADR consequence now states the resolution; the map's D6 should carry the same sentence so the two do not drift.

## VM Shell Sessions: the implementation matches the assumption, and the ask is accurate

`pty-manager.ts:775-778` reads `ptys.get(ptyId)?.taskId` with no `shell`/`shellSession` filter, unlike `findByTask` at :787 which skips both — so the lock keys by `taskId` for every PTY kind, exactly as your description says. `core-link-session-lock.test.ts` ("a VM Shell Session is locked by the same rule") pins it. ADR 0024 still carries the bullet unamended (*"assumed… not separately decided"*), which is correct: this PR implements the assumption, it does not decide it. Your side note checks out too — user terminals spawn with `newClientId("ut")` (`user-terminal-warm-pool.ts:105,111`), so nothing an operator's terminal does collides with a harness Session's lock. Not blocking on the decision; it closes on #140.

## Changes requested

**1. `harnessPrompt` is a task mutation addressed at a Session, and it is neither gated nor flagged** — `pty-core-link-server.ts:1100-1107`. It names a `taskId` and lands on `titleGenerator.schedule(taskId, prompt)` (`core-entry.ts:283-285`), which writes the row's title through the same `taskWriter` a gated `tasksMutate` uses. So a connection holding nothing can rename a Session another connection holds, which is inside D4's *"every task mutation addressed at that Session"* on any reading. Small in consequence — the generator leaves a real or operator-set title alone — but this is the ticket that owns D4, and it is the one place the enumeration should be complete. Either gate it with `refuseLockedSession`, or give it the paragraph `resize` got at :912-919 and the ADR bullet that goes with it.

**2. `spawn` against an existing `taskId` is ungated and unnamed** — `pty-core-link-server.ts:875`. `spawn` carries `opts.taskId`, and nothing dedupes on it, so a Reader can start a second harness process on a Session another connection is driving. D4 does not enumerate `spawn`, so I am not calling this a criterion failure — I am asking for the same treatment you gave `resize`: state the reading in the code and in ADR 0024's consequences, or gate it. Silence here reads as "not considered", which is the one thing the `resize` paragraph proves you do not do.

**3. The Panel client drops `error.code`** — `client.ts:1238-1239`, `case "error": throw new Error(msg.message)`. The frame work is right and the wire is genuinely unambiguous, but the only client in this repository throws away the field the moment it arrives: a caller can tell locked (throws) from gone (resolves `ok: false`), yet cannot tell locked from any other error without matching the prose that `code` exists to avoid — the exact thing your own table promises. Attach the code to the thrown error (a typed error, or `Object.assign(new Error(msg.message), { code: msg.code })`). #146 will need it the day it lands.

## The seven criteria

| Criterion | Verdict | Enforcing line |
| --- | --- | --- |
| Three frames exist, `reqId`-correlated | **Met** | `core-link-frames.ts:551,555,559` + `734,738,740-745`; dispatch `pty-core-link-server.ts:935-971`; test "answers each one on the reqId it was asked with" |
| Refusal distinguishable from session-gone | **Met** | `pty-core-link-server.ts:842-866` sends `error` + `code: "session-locked"`; gone stays `writeResult { ok:false }`; test "tells 'locked by someone else' apart from 'that Session is gone'". See item 3 for the client half |
| Covers write, kill, every task mutation; per Session, not Core-wide | **Met** for D4's enumerated frames — `:907`, `:929`, `:1077-1082`; concurrency proven by "two connections, two Sessions, one Core". Items 1 and 2 are the unenumerated edges |
| Drop releases **all** of the connection's locks | **Met** | `dropConnection` → `sessionLocks.releaseAll(conn)` at `:601`, unconditional and ahead of the identity guard; `releaseAll` sweeps the table rather than a tracked list; test holds three Sessions, closes, asserts `sessionLockCount() === 0` and all three writable |
| Force takeover immediate; loser's next mutation refused | **Met** | `session-lock-table.ts:150-156` overwrites unconditionally; test asserts the loser's very next `write` comes back `session-locked` and the winner's succeeds |
| The D5 window tested | **Met** | Two tests: creator spawns and does not claim (`sessionLockCount() === 0`), second connection claims and the creator is then refused; plus the creator losing the race outright |
| No idle timeout anywhere | **Met** | No timer, `Date.now`, TTL or expiry anywhere in `session-lock-table.ts` or the gate; the module comment names the three endings and says not to add a fourth; test holds a lock across 30 simulated minutes, past the 45s heartbeat |

`PtyCore.kill` is correctly ungated — `pty-manager.ts` gains `taskIdForPty` and nothing else, and the "the Core never refuses itself" test calls `core.kill` directly on a held Session. `killLaunchProcesses` staying ungated (addressed at a folder) is right and tested. D10 noted: I have reviewed the lock as an accident guard, and nothing here is written up as a vulnerability.

Boundaries clean: `ws-server.ts` and `panel-link-client.ts` untouched, `CORE_LINK_PROTOCOL_VERSION` still `0.15.0` with only commentary added, no `clientId` and no lock state on the snapshot or event stream — nothing from #146/#147 has leaked in. All five gated frames are in `MULTI_CONNECTION_ONLY_FRAME_TYPES` (`client.ts:61-68`) and the exact-membership test names all five. CI green on e3a97e0.

Good ticket. The lock table is the right size, the tests assert absences rather than happy paths, and both open questions were raised in the description where they belong.

qcentic-dev:
All three fixed, in two commits — 90ca4f9 (the Core gate) and d882e71 (the Panel client). The design is unchanged; the two "record the reading" items are recorded *and* gated, and I say why below.

**1. `harnessPrompt` — gated.** Your reading, and I did not find a case for the other one. The frame reads like telemetry and lands as a write: it reaches `titleGenerator.schedule`, which writes the row's title through the same `taskWriter` a gated `tasksMutate` uses. D4 is decided by where a frame lands, not by how it reads, so a connection holding nothing can no longer rename a Session another connection is driving. The `resize` paragraph would have been the wrong shape here — that one says a Reader has a genuine use for the frame, and a Reader has no use for renaming somebody else's Session. Gated at `pty-core-link-server.ts:1104`, with the reasoning in the code and an ADR consequence bullet naming where it lands.

**2. `spawn` against an existing `taskId` — gated, not just recorded.** Recording it alone did not survive writing the paragraph. A `spawn` naming a held Session starts a *second* process on the Session that connection is driving — two harnesses in one worktree, which is the interference the lock is for rather than a lesser cousin of it. And the `resize` reading does not transfer: the lock keys by `taskId` for every PTY kind, so the very next `write` to the PTY that spawn would create comes back `session-locked` — the refusing connection could not have typed into it anyway. Refusing at the spawn is the same answer, one round trip earlier and legible, instead of a live process nobody may drive. D5's window is untouched: a Session nobody holds is still spawned for anybody, so a creator spawns without claiming exactly as before, and `tasksMutate`'s `create` keeps its own carve-out. Gate at `:875`, reasoning in the code, ADR consequence bullet alongside the `resize` one. D4 is still what it was — the ADR now records `spawn` as gated *though D4 does not enumerate it*, so the widening is visible rather than smuggled in.

**3. `error.code` survives the throw.** `unwrapResponse` now raises an exported `CoreLinkRequestError` carrying `code` alongside `message` (`client.ts`), so a caller narrows on the class and reads the field rather than matching prose. Optional here as on the wire — absent on every error that shipped before the field, so a reader takes the code when it is there and falls back to the message when it is not, never the reverse. #146 gets to `instanceof` it.

**Tests.** Four new cases in `core-link-session-lock.test.ts`: `harnessPrompt` refused for a non-holder with nothing reaching the prompt port, and served both for the holder and on an unheld Session; `spawn` against a held Session refused with the Core never starting a process, and still serving a Session nobody holds from any connection. Two in `multiconnection-capability.test.ts`: the thrown error carries `session-locked` and is a `CoreLinkRequestError`, and an error frame with no code leaves it `undefined` rather than inventing one.

Local: `pnpm typecheck` clean, lint 0 errors, and the full suites green — core 743, panel 1318, shared 226. Nothing else changed; the PR stays draft.

The D6 follow-up for #140 is noted and not touched here.

qcentic-adm (approved):
Re-reviewed at `d882e71`, reading the enforcing lines rather than the reply. All three items are fixed, and two of them are fixed harder than I asked — both "record the reading" items came back gated, with the reading recorded anyway. Approving.

## 1. `harnessPrompt` — gated

`pty-core-link-server.ts:1130` — `if (this.refuseLockedSession(conn, frame.reqId, frame.taskId)) return;`, the same helper and the same `taskId` keying as `tasksMutate` at `:1101` and `write`/`kill` through `refuseLockedPty`. It sits **above** the `accepted` computation, so a refusal leaves on the `error` frame with `code: "session-locked"` rather than as `harnessPromptResult { accepted: false }` — which matters, because that flag already means "no port, or an empty prompt" and would have made the refusal indistinguishable from a no-op. Reasoning is in the code and in a new ADR 0024 consequence bullet naming where the frame lands.

You picked the right shape for the argument, too: the `resize` paragraph says a Reader has a genuine use for the frame, and a Reader has no use for renaming somebody else's Session. That asymmetry is the whole reason the two frames resolve differently.

Tests: *"refuses a prompt aimed at a Session another connection holds"* asserts the `error` + `SESSION_LOCKED_ERROR_CODE` **and** `promptPort.submitted` empty — refused, not accepted-and-ignored. *"serves it for the holder, and on a Session nobody holds"* covers the other half, asserting the exact submission log so a stray prompt could not pass. The `mockPromptPort` comment explaining *why* the assertion is on the port rather than the flag is the right note to leave.

## 2. `spawn` on an existing `taskId` — gated, and the fresh-spawn path is intact

`pty-core-link-server.ts:896`. I checked the thing that would have made this a revert rather than a fix — whether the gate breaks the original spawn path — at the enforcing line, not at the test:

`session-lock-table.ts:95` — `mayMutate` returns true on `!taskId`, and true when `holders.get(taskId) === undefined`. Nothing writes to `holders` on spawn (D5 — still no creator privilege), so a brand-new `taskId` is unheld, `mayMutate` is true, and the frame passes untouched. A creator spawns without claiming exactly as before; a connection re-spawning into a Session it holds passes on `current === holder`; `tasksMutate`'s `create` keeps its own carve-out; synthetic `ut-…` terminal spawns are fresh ids every time and never collide. Only a `taskId` **another** connection holds is refused. The path every client depends on is unchanged.

*"still starts a Session nobody holds, from any connection (D5)"* pins it — a stranger and the holder each spawn a fresh Session while `task-a` is held, both succeed, `sessionLockCount()` stays 1. *"is refused before a second process exists"* asserts `core.spawns` never grew and no `spawned` frame went out, so it is a refusal rather than a spawn reported as a failure.

Your argument for gating over recording is the stronger one and I accept it: the lock keys by `taskId` for every PTY kind, so the first `write` to the new PTY comes back `session-locked` anyway — recording-only would have left a live second harness in the worktree that nobody may drive, which is worse than the refusal. ADR 0024 records it as gated *though D4 does not enumerate it*, so the widening is visible rather than smuggled in, and D4 is still what it was. That is exactly the right way to leave it for #140.

## 3. `error.code` survives the throw

`client.ts:88` (`CoreLinkRequestError`, exported, `code` optional and readonly) and `:1272`. I checked for a second boundary that could still drop it: `spawnError` at `:1225` also throws a bare `Error`, but that frame carries no `code` field (`core-link-frames.ts:691`) — `error` at `:795` is the only code-carrying frame, and it is the one the refusal uses, so nothing leaks past. A refused `spawn` therefore surfaces as `CoreLinkRequestError { code: "session-locked" }` too, not as a `spawnError`.

Tests: the code rides the rejection and the class is assertable; and the absent-code case leaves it `undefined` rather than inventing one, which is the right half to pin for an additive optional field. #146 can `instanceof` it.

## Scope and CI

`e3a97e0..d882e71` is two commits, five files, +219/−2: the ADR, the two gates, the client, and the two test files. `e3a97e0` is a true ancestor — no rebase hiding anything. `ws-server.ts` and `panel-link-client.ts` still untouched, `CORE_LINK_PROTOCOL_VERSION` still `0.15.0`, no `clientId` or lock state on the snapshot. Test-case deltas are exactly +4 core and +2 panel, matching the cases described and nothing more. CI green on `d882e71` — all 20 checks pass, Unit Tests, Typecheck and Lint included. (Minor: the panel totals in your comment, 1310 → 1318, do not reconcile with a +2 delta — reporting noise, the suite itself is fine.)

The two open design points are unchanged and still correctly parked: VM Shell Sessions stay an assumption for #140 to confirm or amend, and the D6 wording follow-up is noted for the map. Neither is this PR's to settle.

Approving. Good fixes — the reasoning survived being written down, which is why two of them changed behaviour instead of just gaining a paragraph.

Issue #147, opened 2026-08-11
Panel: read-only Sessions, claim affordance, tab arbitration (D3, D8)


mrfqcentic:
**Base branch: `feat/multi-connection-core`, not `main`.** Train: **`beta/0.2.0`**.
Suggested branch: **`feat/panel-read-only-sessions`**.
Implements the Panel half of **D3** and **D8** of #140. Last, because it consumes everything above.

## What

Two things, both Panel-side:

**1. Render a Session you do not hold.** The terminal shows every byte and accepts no input.
Read-only has to be visible before a keystroke, not discovered by one going nowhere — an input that
silently does nothing is worse than no input at all. A claim affordance, and a force takeover
behind a confirmation that names what is being taken.

**2. Arbitrate between the Panel's own tabs** (D3). The Panel holds one lock per Session for all
its browser tabs; which tab drives is the Panel's own business and never crosses the wire. Two tabs
on one Session today both write, interleaved, with no arbitration — this is where that is settled,
between Panel sessions, inside the Panel.

## Done when

- A Session locked by another Core client renders read-only, labelled, with no input affordance.
- Claiming an unlocked Session from the Panel works and the UI reflects it without a refetch.
- Force takeover is available, is confirmed, and names the consequence.
- Two Panel tabs on one Session: one drives, the other is read-only, and it is obvious which.
- Against a Core without the `multiConnection` capability, none of this appears — the Panel behaves
  exactly as it does today.

## Traps

- **Do not surface the intra-Panel arbitration as a lock.** It is not the Session lock — the Panel
  holds that once for all its tabs. Two names for two things, or every bug report will be
  ambiguous.
- **Singular UI still holds** (`CONTEXT.md`): read-only rendering is the same component with a
  state, not a second terminal component.
- The loser of an intra-Panel handover and the loser of a cross-client force takeover are different
  events with different copy. Do not collapse them.

Refs #140

qcentic-dev:
Delivered in PR #184, squash-merged into `beta/0.2.0` as 57d43c906331a1daac84d81282a76df8bf4cfd4c. The PR base is non-default, so closing keywords did not fire automatically; closing manually. Epic #140 stays open.

Issue #145, opened 2026-08-11
Publish lock state on the Session snapshot and as an event (D8)


mrfqcentic:
**Base branch: `feat/multi-connection-core`, not `main`.** Train: **`beta/0.2.0`**.
Suggested branch: **`feat/session-lock-state-published`**.
Implements **D8** of #140.

## What

Lock state is published, not discovered by failing. It rides the Session snapshot and emits an
event on every change.

A read-only Panel tab has to render its terminal non-editable **before** anyone types, `actana
session ls` has to show who holds what, and the loser of a force takeover has to learn it when it
happens rather than on their next keystroke.

## Done when

- The Session snapshot carries lock state, and it is enough to answer "can *I* write to this?" —
  not merely "is this locked", which every client would then have to compare against itself.
- A dedicated event kind fires on claim, release and takeover, so a reconnecting client replays a
  lock change distinctly rather than refetching on an undifferentiated "something changed". Follow
  the precedent ADR 0022 set for `project:appearanceChanged`.
- A client that was force-taken observes the transition without polling.
- The event is in the Core's event log and replays by cursor like every other event.

## Traps

- **Do not widen an existing mutation event to carry this.** ADR 0017 and ADR 0022 both rejected
  generic field patches for the same reason: the frame stops documenting what changed and every
  client refetches on everything.
- **Identity leaks here if you are careless.** D10 says the lock is coordination, not security, and
  D3 says the holder is a connection — so the snapshot should say *whether you hold it*, not
  broadcast a client's identity to everyone watching.

Refs #140

qcentic-dev:
Delivered in PR #184, squash-merged into `beta/0.2.0` as 57d43c906331a1daac84d81282a76df8bf4cfd4c. The PR base is non-default, so closing keywords did not fire automatically; closing manually. Epic #140 stays open.

PR #178, opened 2026-08-11
feat(core-link): lock state is published — it rides the Session snapshot and emits its own event (D8)

qcentic-dev:
Implements **D8** of #140 — lock state is published, not discovered by failing.

**Base is `feat/multi-connection-core`, not `main`.** Train: `beta/0.2.0`. Draft: #146 runs in
parallel from the same base, and whichever of the two merges second rebases onto the integration
tip.

Extends #176 rather than reworking it: the lock table, the three frames and the mutation gate are
untouched. What is new is that a client no longer has to be refused to find out where it stands.

---

## The snapshot answers "can *I* write to this"

Every Session snapshot the Core sends — `tasksList`, `archivedTasksList`, `tasksMutateResult`,
`sessionsList` — is stamped on its way out with:

```ts
lock: { writable: boolean; state: "unlocked" | "held-by-you" | "held-by-another" }
```

It is an **addressed** field, not a broadcast one: the holder and a watcher listing the same
Session are told different things about it. That is what spares every client the comparison of a
holder against itself — and spares the protocol the client identity it would need to make that
comparison at all. A watcher learns `held-by-another`; there is nothing further to learn.

`writable` rides alongside `state` because *which states are writable* is a Core-side rule, not a
property of the word "unlocked": an unlocked Session is writable by anybody, which is D11's
compatibility promise. A client rendering an editable terminal reads `writable`; a client choosing
between a "Claim" and a "Take over" affordance reads `state`.

Stamped in the server rather than produced by a query port, because the lock table is in memory
(D12) so no store has the fact, and because the answer differs per recipient so there is no single
value a shared read path could have returned.

## A dedicated event kind

`session:lockChanged`, carrying `{ taskId, transition, locked }` and no identity. It fires on
claim, on release — including the release a connection drop performs — and on takeover. It is an
ordinary event-log row, so it replays by cursor exactly like every other event.

The transition is named in the vocabulary of the lock rather than of the frame: a `forceTakeover`
against a Session **nobody** held publishes `claimed`, not `taken-over`, so no client reports an
eviction that did not happen. `locked` is read back off the table rather than inferred, so the log
cannot drift from the state it describes, and a client can act on it without knowing the whole
transition vocabulary — which is what keeps a transition added later (#146) from reading as "no
change" to a client that predates it.

**Nothing changed publishes nothing.** An idempotent re-claim, a denied claim, a release from a
connection that holds nothing, and a takeover by the current holder each answer their frame and
append no event.

## The two traps, and where they are handled

**No widened mutation event.** `task:updated` was left alone. ADR 0017 and ADR 0022 rejected
generic field patches because the frame stops documenting what changed and every client refetches
on everything — and here it has a second cost: a reconnecting client replaying a tail could not
tell a takeover it must react to from a title edit it can ignore. This follows the
`project:appearanceChanged` precedent ADR 0022 set instead.

**No identity leak.** D10 says the lock is coordination, not security; D3 says the holder is a
connection. So the snapshot says *whether you hold it*, never *who holds it*, and the event payload
names neither winner nor loser — one logged row is read by every connection, so anything
identifying in it would be broadcast to every watcher of the Core. A takeover needs to name
nobody: the winner learns it won on its own `forceTakeoverResult`, and every other connection reads
`taken-over` as "the holder is now somebody who is not me", which is the entire fact a loser needs.
`SessionLockTable.holderOf` stays for the Core's own reference comparison and nothing puts its
answer on the wire.

## The protocol version does not move

Fourth time, same reason as #142, #143 and #144 (D11). The `lock` field is optional; a Core that
does not announce `multiConnection` omits it and appends no lock events; absent reads as "this Core
has no lock table", which is the single-connection Core that shipped before it.

Worth stating explicitly because #84 *did* bump the minor for snapshot fields, and the difference
is the rule itself: there the absence yielded something *lesser* than today — a Panel synthesizing
`titleManuallySet: false` rendered a Session its generator could rename out from under the operator
— while a Panel that predates `lock` renders exactly what it renders today, because read-only
rendering is new surface (#147) rather than a protection quietly going missing.

## Scope

No new request frames, so `MULTI_CONNECTION_ONLY_FRAME_TYPES` and its exact-membership test are
unchanged. `dropConnection` gains the event for the locks it already released and nothing else —
reclaim, client ids and the release path itself belong to #146. Panel-side rendering is #147; this
PR gives it the data.

## Tests

`packages/core/src/__tests__/core-link-session-lock-published.test.ts` — 16 tests, organised
around the "Done when", including the ones that assert an absence: the lock object's keys are
exactly `state` and `writable`, nothing identity-shaped reaches a watcher's wire, the stamp is
written onto a copy (the fake port hands back the same row objects every call), and the four
gestures that change nothing publish nothing. The loser of a takeover is asserted on a socket that
sends nothing between its claim and hearing the news; the replay tests assert on a client that was
not connected when the change happened.

The five exact-equality assertions in the Panel's core-link suite grew the field rather than
loosening to `toMatchObject`.

`pnpm typecheck` and `pnpm lint` clean; core 759, shared 226, panel 1318 tests pass. Two panel
files (`live-read-path`, `terminals-in-browser`) time out intermittently under a fully parallel
run on this machine and pass in isolation; `scripts/__tests__/core-launcher.test.mjs` fails here
because `/opt/actana` exists on the box. None are touched by this change.

Refs #145
Refs #140


qcentic-dev:
## Done when — #145

- [x] **The Session snapshot carries lock state, and it is enough to answer "can *I* write to
  this?" — not merely "is this locked", which every client would then have to compare against
  itself.**
  `lock: { writable, state }` on `CoreLinkTaskSnapshot` and `CoreLinkSessionSnapshot`, stamped per
  recipient by `PtyCoreLinkServer.withPublishedLock` from `SessionLockTable.stateFor(taskId,
  asker)`. `state` is `unlocked` / `held-by-you` / `held-by-another`; the holder and a watcher get
  different answers to the same frame, so no client compares anything. It rides all four Session
  snapshot frames: `tasksList`, `archivedTasksList`, `tasksMutateResult`, `sessionsList`. Pinned by
  *"answers the same frame differently for the holder and for a watcher"* and *"agrees with the
  gate: what a snapshot calls writable is what is served"*.

- [x] **A dedicated event kind fires on claim, release and takeover, so a reconnecting client
  replays a lock change distinctly rather than refetching on an undifferentiated "something
  changed". Follow the precedent ADR 0022 set for `project:appearanceChanged`.**
  `SESSION_LOCK_CHANGED_EVENT_KIND = "session:lockChanged"`, payload
  `{ taskId, transition, locked }`, appended on claim, on release, on takeover, and on the release
  a connection drop performs. No existing mutation event was widened. Pinned by *"appends one
  session:lockChanged per transition, on its own kind"* and *"names a takeover a takeover, and a
  takeover of nobody's Session a claim"*.

- [x] **A client that was force-taken observes the transition without polling.**
  The event reaches every subscribed connection on the ordinary live-push path. Pinned by *"is
  told, on a connection that asked for nothing"* — the loser's socket sends nothing between its
  claim and hearing the news — and *"reads its new answer off the next snapshot it takes"*.

- [x] **The event is in the Core's event log and replays by cursor like every other event.**
  An ordinary `eventLog.appendEvent` row with its `taskId` column set, so it is routable without
  parsing the payload. Pinned by *"hands a reconnecting client the lock changes it missed, in
  order"* and *"replays only what is past the cursor the client presents"*.

### Traps

- [x] **No existing mutation event was widened** (ADR 0017, ADR 0022). `task:updated` is untouched;
  the new kind is dedicated, on the `project:appearanceChanged` precedent.
- [x] **No identity leaks** (D3, D10). The snapshot says whether *you* hold it; a watcher's whole
  vocabulary for anyone else is `held-by-another`. The event payload names neither winner nor
  loser. `SessionLockTable.holderOf` remains internal and nothing puts its answer on the wire.
  Pinned by *"tells a watcher nothing about who holds it"* and *"carries no client identity in the
  payload"*.

### Two notes on scope, neither a "Done when" item

- The **What** section's *"`actana session ls` has to show who holds what"* is **not implemented
  here, and cannot be**: there is no `session` command in the CLI on this base
  (`packages/core/src/actana-cli.ts`). The frame such a command would read — `sessionsList` — now
  carries the lock, which is this ticket's half of it.
- Read-only **rendering** in the Panel is step 5 of #140 (#147). This PR publishes the data it
  needs and stops there.

### Boundary with #146

`dropConnection` gained the event for locks it already released, and nothing else. Reclaim, stable
client ids and the release path itself are untouched. No new request frame types, so
`MULTI_CONNECTION_ONLY_FRAME_TYPES` and its exact-membership test are unchanged. Whichever of
#178 / #146 merges second rebases onto the integration tip.

**PR:** #178 — https://github.com/actana/control/pull/178 (draft, base `feat/multi-connection-core`)

qcentic-adm (approved):
Reviewed against issue #145's "Done when", one criterion at a time, reading the enforcing lines rather than the description. Diff taken against `feat/multi-connection-core` (merge base `d5ddeb6`). **Approving.**

## The four "Done when" items

**1. The snapshot answers "can *I* write to this?" — met, and addressed per connection.**
`pty-core-link-server.ts:912-918` stamps on the way out with `this.sessionLocks.stateFor(row.taskId, conn)` — the asking connection is a parameter, so this is not a raw holder value every client compares against itself. `session-lock-table.ts:129-134` collapses every non-asker into `held-by-another`. All four frames that carry a Session snapshot are stamped and there is no fifth: `core-link-frames.ts:877/880/881/891` enumerate `tasksListResult`, `archivedTasksListResult`, `tasksMutateResult`, `sessionsListResult`, and each has a `withPublishedLock` call site (`:1191`, `:1210`, `:1247`, `:1293`).

The addressing survives replay and caching for three independent reasons, each checked:
- Nothing snapshot-shaped ever enters the event log. `core-task-writer.ts:106` appends `{ taskId, projectId }`; `:150` appends ids and a title. No row, so no stamp, so no replayed frame can carry another connection's answer.
- The stamp is written onto a copy (`:917` spreads), and the ports return fresh rows anyway (`core-mutation-store.ts:154` queries SQLite per call). The test at `core-link-session-lock-published.test.ts:344` pins this with a fake port that deliberately hands back identical row objects — without the spread, the holder reads `held-by-another` about its own Session on its second list.
- Every stamped frame is a `reqId` answer sent to one socket. Nothing broadcasts a snapshot: `fanOutPtyEvent` (`:676`) carries `data`/`exit` only, `pushLiveEvents` (`:781`) carries event rows only.

**2. A dedicated event kind on claim, release and takeover — met, not a widened mutation event.**
`SESSION_LOCK_CHANGED_EVENT_KIND = "session:lockChanged"` (`core-link-frames.ts:566`), appended through the ordinary log at `pty-core-link-server.ts:952` with `{ taskId, ptyId: null }`, so it replays by cursor through the same `readEventTail` path as everything else (`:789`) with no kind filter anywhere in between. `task:updated` is untouched — `core-task-writer.ts:96-106` is not in the diff. Three call sites plus the drop: claim `:1049` (guarded by `alreadyHeld` read *before* the claim, `:1044`), release `:1064` (guarded on `released`), takeover `:1092`, drop-release `:623`. Replay is pinned at `core-link-session-lock-published.test.ts:495` (client absent when it happened) and `:513` (cursor past the first change).

Nice detail: `:1094` publishes `claimed` rather than `taken-over` when `takenFrom === "nobody"`, so the wire speaks the vocabulary of the lock and no watcher reports an eviction that did not occur. And `locked` is read back off the table (`:947`) rather than inferred, so the row cannot drift from the state it describes.

**3. The force-taken client observes it without polling — met; delivery path traced end to end.**
`forceTakeover` → `:1092` `recordSessionLockChange` → `eventLog.appendEvent` → the loser's own per-connection push loop started at `:536` → `pushLiveEvents` (`:781-793`), which pushes every row past that connection's cursor with no kind filtering → `event` frame on the loser's socket. The loser sends nothing; the server-side tail read is pre-existing transport for all events, not client polling. Test `:449` asserts exactly that — the socket's outbound count is unchanged between its claim and hearing the news.

Panel-side it already lands, before #147 renders it: `use-core-live-queries.ts:41` and `use-fleet.ts:30` both match `^session:`, so the event invalidates `tasksList`, whose refetch returns the freshly stamped `lock`. Worth knowing for #147 that this means a claim/release now invalidates the tasks *and* archived-tasks queries on every mounted view — correct, just a new source of refetch traffic.

**4. In the Core's event log and replays by cursor — met.** Covered under (2); `eventsReplayed` marker asserted at test `:505`.

## The specific claims asked about

**No identity leak — holds.** `stateFor` (`session-lock-table.ts:129`) returns only `{ writable, state }`; the payload type (`core-link-frames.ts:598-612`) is `{ taskId, transition, locked }`. `holderOf` has exactly one new caller, `:947`, and it is `!== null` — the holder object never reaches a serializer. Grepping the diff for `clientId|connectionId|holder|heldBy` finds hits only in prose, tests and the drop-log. Tests pin the absence at both layers: parsed keys (`:305`, `:454`) and the raw serialized bytes on a watcher's wire (`:311`).

**PR #176's machinery extended, not reworked — holds.** `session-lock-table.ts` is +35/-0: an import, `stateFor`, and header prose. `claim`/`release`/`forceTakeover`/`mayMutate`/`releaseAll` are untouched, and the gate is unchanged.

**Boundaries respected.** No reclaim, no client ids (#146's ground); `dropConnection` gains one line (`:623`) and nothing else — the release sweep, the identity guard and the emit-target release are as they were. No new request frames, so `MULTI_CONNECTION_ONLY_FRAME_TYPES` and its exact-membership test are correctly untouched.

**Docs match the code.** ADR 0024 gains four consequences that describe what the code does (addressed not broadcast; dedicated kind on the 0022 precedent; nothing-changed-publishes-nothing; rides the capability, version stays at 0.15.0 — which `core-link-frames.ts:1184` confirms). The landing-steps bullet now reads "D8 landed with #145". CONTEXT.md's Session lock term gains the published sentence. I checked the version claim against the code: the field is optional on both snapshot types and `:916` / `:943` both short-circuit when `multiConnection` is not announced, so a non-announcing Core stamps nothing and appends nothing — test `:355` pins it.

**CI green on the head.** Run 31497977457 on `ef7b0b0` — `success`. Typecheck, Lint, Unit Tests, Conventions, both installer E2Es, both tarball smokes all pass; only the publish/train jobs skip, as expected for a PR.

## One non-blocking nit

`session-lock-table.ts:130-133` re-derives `mayMutate`'s predicate instead of calling it, while the doc directly above says "`writable` … is `mayMutate` for this asker — the same rule, from the same table". Today the two agree, and `core-link-session-lock-published.test.ts:279` pins the agreement for one scenario. But the invariant that matters — never told writable and then refused — is held by two copies of the same condition rather than by construction, so a future change to `mayMutate` (a grace period, a null-taskId nuance) would silently desynchronise the published answer. `writable: this.mayMutate(taskId, asker)` would make it structural. Not worth blocking on.

Also noting the PR is still a draft; nothing here needs to change before it is marked ready.

## Checklist

- [x] Snapshot carries lock state, answers "can *I* write to this", addressed per connection
- [x] Addressing survives replay and caching (nothing snapshot-shaped in the log; stamped onto a copy; one recipient per frame)
- [x] Dedicated `session:lockChanged` kind — not a widened mutation event
- [x] Fires on claim, release (including connection drop) and takeover; silent when nothing changed
- [x] In the event log, replays by cursor like every other event
- [x] Force-taken client learns of it with no polling — path traced to the loser's socket
- [x] No identity on the wire: snapshot or event, parsed shape or raw bytes
- [x] #176's lock table extended additively (+35/-0)
- [x] Boundaries: no reclaim, no client ids, `dropConnection` gains emission only
- [x] ADR 0024 and CONTEXT.md ride the change and match the code
- [x] CI green on `ef7b0b0`
- [ ] Nit: `stateFor` duplicates `mayMutate`'s predicate (non-blocking)

PR #179, opened 2026-08-11
feat(core-link): a stable client id reclaims a reconnecting client's Session locks (D9)

### LINKED ISSUE #146: Stable client id: reclaim a reconnecting client's locks, reaping only (D9)

qcentic-dev:
Implements **D9** of #140. Closes #146.

**Base is `feat/multi-connection-core`, not `main`.** Train: `beta/0.2.0`.

## What lands

A connection presents a stable client id on a new `reclaim { clientId }` frame. A later connection
presenting the same string is the same Core client reconnecting, so the Core **transfers that
connection's Session locks and then closes it**. It is used for reaping and for nothing else.

The frame exists because D1 took evict-on-connect away. A link that dies without a TCP close leaves
a socket the Core cannot tell from a live one until the heartbeat reaps it 45s later, still holding
that client's Sessions. **The heartbeat path is untouched** — it still reaps the client that never
comes back; the reclaim is the answer for the one that does, in a round trip instead of 45 seconds.

### The atomicity is the ticket

`SessionLockTable.transferAll` rewrites the holder of every lock one connection holds, **in place**.
The composed spelling — `releaseAll` then a `claim` per id — is a different behaviour wearing the
same result: between those two calls each Session is unlocked, and unlocked is claimable by anybody
(D5) and writable by anybody (D11).

The order in `reclaimFor` is the other half, and it is the half a reviewer should check: locks move
**before** the predecessor's socket is touched, because `dropConnection` sweeps every lock its
connection holds and would unlock them all on the way out. Run in this order, that same `releaseAll`
finds nothing left to release.

The test for this does not check the end state — the wrong order reaches the same end state. It
reaches *into* the moment the predecessor is dropped (a `close` listener registered after the
server's own) and has a third client try both ways into the Session: the explicit `claim`, and the
`write` an unlocked Session serves to anybody. Both are refused. Verified by mutation: reordering
the implementation to drop-then-reclaim fails exactly that test and nothing else.

### Nothing treats the id as authentication

The ticket asks for a test asserting a forged id grants no authority. It is in
`core-link-client-id-reclaim.test.ts`, in four parts:

1. **It does not authenticate.** `reclaim` is not `auth`, so the existing gate refuses it — and
   drops the connection — from a client that has not presented its bearer. Nothing moves, nothing is
   reaped, no lock count changes.
2. **It moves no more than the predecessor's own locks.** A third client's Session stays held, and
   the forger's write to it still comes back `session-locked`.
3. **It takes no more than `forceTakeover` already hands any authenticated connection**, without an
   id at all and unconditionally (D7). That is the argument for not defending the string, as an
   assertion: a forged id is a slower spelling of a frame that is already unconditional, not an
   escalation.
4. **It is verified against nothing** — any string, from a connection holding an entirely different
   bearer, is accepted. Pinned deliberately so a change that starts verifying the id has to argue
   with this test and ADR 0024 rather than land quietly.

Per the trap in the ticket: nothing here signs or verifies the id, and nothing should. Whoever
opened the socket holds the bearer, and a bearer holder can open a VM Shell Session and kill the
process directly (D10).

### Where the id comes from

Per Core client, and **not** from the registration blob — the blob is machine-scoped, so an id
derived from it (`coreId`, a bearer digest) would make the Panel, the CLI and every automation on a
machine one connection, each reconnect reaping the others.

The Panel mints one per `PtyCoreLinkClient` — one per registered Core, held for the life of that
link — and presents it on every reconnect. That is the honest default rather than a shortcut: the
id's job is to span a reconnect, and the ghost it reclaims is a socket whose peer never sent a FIN
(a partitioned container, a suspended host), which leaves the Panel process running and this
instance reconnecting with the id it has held all along. A Panel that has genuinely restarted is a
new Core client, and its predecessor's sockets were closed by the kernel as it died — there is
nothing left to reclaim. A `clientId` option is the seam for a caller with somewhere durable to keep
one. The CLI mints one per invocation; it terminates no core link in this build, so there is no CLI
code in this diff.

## New frame, and the capability gate

This is a **new frame**, not a field on `auth` or `ready`, so per the review of #176 it is in
`MULTI_CONNECTION_ONLY_FRAME_TYPES` with the exact-membership test updated. Riding `auth` was
rejected: a loopback Core configures no `authVerifier` and never receives an `auth` frame, so the id
would have no path on a Core that has clients but no bearer.

Compatibility with a capability-less Core is exact. The client consults the gate and sends nothing —
and here the degraded behaviour is not a compromise but the same behaviour: such a Core evicts the
previous connection on connect already, so the frame asks for what that Core has just done, out of a
lock table it does not have. `CORE_LINK_PROTOCOL_VERSION` does not move (D11).

Ordering on the reconnect path: `auth` → `reclaim` → `subscribe` → `ptySubscribe` resend. The
reclaim goes first because it is the only frame there with a clock on it — until it lands, the
socket it replaces still holds this link's Sessions. It is fire-and-forget: a reclaim that does not
land costs the 45s heartbeat it was shortening, a slower path to the same state and never a wrong
one.

Error paths keep carrying `error.code` (kept for this ticket's benefit per the #176 review) — the
forged-id tests read `session-locked` off it rather than matching prose.

## Boundaries

- **#145 owns lock-state publication** (snapshot field, dedicated event kind). This PR adds no
  snapshot field and no event kind; the only new vocabulary is the `reclaim` / `reclaimResult` pair.
  We both touch `session-lock-table.ts` and `pty-core-link-server.ts` — this diff is confined to
  `transferAll`, the `clientId` field on `ActiveConnection`, and the one dispatch case. Whichever
  merges second rebases onto the integration tip.
- Not D10, not identity. No signing, no verification, no per-client credential.

## Testing

- `packages/core/src/__tests__/core-link-client-id-reclaim.test.ts` — 13 tests, organised around the
  ticket's "Done when".
- `packages/panel/src/server/core-link/__tests__/client-id-reclaim.test.ts` — 9 tests: id stability
  across reconnects, distinctness per client, ordering, the pre-auth withhold, the capability-less
  fallback.
- Parse test for the new frame; exact-membership test updated.
- `pnpm -r test`: shared 227 passed, core 756 passed. Panel's suite has pre-existing timing
  flakiness under parallel load in `project-dialog-core-switch` / `live-read-path` /
  `token-usage-rollup` — reproduced identically at the base commit, and all pass in isolation. Same
  for `scripts/__tests__/core-launcher.test.mjs`, which fails on any machine with a real
  `/opt/actana`; no file under `scripts/` is touched here.
- `pnpm typecheck` clean; `pnpm lint` 0 errors (pre-existing warnings only).

Refs #140


qcentic-dev:
## Issue #146 — "Done when", item by item

- [x] **Reconnecting with the same client id closes the predecessor and transfers its locks, in one
      step, with no window in which the Session is unlocked and claimable by a third party.**
      `SessionLockTable.transferAll` rewrites each holder in place, and `reclaimFor` runs it *before*
      the predecessor's socket is touched — `dropConnection` sweeps every lock its connection holds,
      so the other order unlocks them all on the way out. Proven by a test that reaches into the
      moment of the drop (a `close` listener registered after the server's own) and has a third
      client try both ways in: the explicit `claim` (denied) and the `write` an unlocked Session
      serves to anybody (refused, `session-locked`). Mutation-checked: reordering the implementation
      to drop-then-reclaim fails that test and nothing else.

- [x] **Reconnecting with a different client id does not disturb anybody.** The other connection
      stays open, keeps writing its Session, and the newcomer's `claim` on it is denied. A
      connection that has presented *no* id is never reaped by anybody's reclaim either — every
      client predating D9 is that connection (D11) — and an absent `clientId` field is treated as no
      id rather than a shared one, so two such connections do not match each other.

- [x] **The id is stable across a reload.** Same string on every reconnect of a link, asserted
      across three connection cycles. One scope note, stated rather than glossed: in this build the
      **Panel service** terminates the core link (one `PtyCoreLinkClient` per registered Core), not
      the browser tab — so a tab reload does not touch the core link at all, and the meaningful
      analogue is the reconnect, which is covered. The id is per client instance and does **not**
      survive a Panel *process* restart; that case is deliberately out of scope and argued in the
      ADR: a process that died had its sockets closed by the kernel, so there is no ghost left to
      reclaim, and the ghost that does exist (a peer that never sent a FIN) leaves the Panel process
      running. The `clientId` option is the seam if that ever needs to change; taking it would need a
      registry column and belongs to its own ticket.

- [x] **Distinct per Core client — the Panel's id is not the CLI's.** Two client instances mint two
      ids (Panel-side test), and two connections presenting different ids coexist without reaping
      each other (Core-side test). The CLI half is **contract only**: no CLI in this build terminates
      a core link (`actana-cli.ts` opens none), so "the CLI mints one per invocation" is documented
      on the frame and in the ADR with no code to write. The Core-side behaviour it will rely on is
      tested with a second connection standing in for it.

- [x] **Nothing anywhere treats it as authentication — the forged-id test.** Four assertions:
      (1) presented before a bearer it is refused `not-authenticated` and the connection dropped —
      no `reclaimResult`, no lock moved, nothing authenticated; (2) a forged id moves only the
      predecessor's own locks, so a third client's Session stays held and the forger's write to it
      is still `session-locked`; (3) it takes no more than `forceTakeover` already hands any
      authenticated connection, unconditionally and with no id at all — the argument for not
      defending the string, as an assertion; (4) it is verified against nothing, accepted as any
      string from a connection holding an entirely different bearer. The last is pinned so a change
      that starts verifying the id has to argue with the test and ADR 0024 rather than land quietly.

### Traps, as checked

- [x] **Not identity (D10).** Nothing signs or verifies the id anywhere; `reclaimFor` is the only
      thing in the Core that reads it.
- [x] **The transfer is atomic.** No release-then-reclaim, and the ADR now records why reaching the
      same end state that way is the absence of D9 rather than a smaller version of it.
- [x] **Not in the registration blob.** Minted per Core client; the blob is machine-scoped and an id
      off it would make every client on a machine one connection.
- [x] **New frame → gated.** `reclaim` is in `MULTI_CONNECTION_ONLY_FRAME_TYPES` with the
      exact-membership test updated. It is a new frame rather than a field on `auth`/`ready`, so the
      capability-less path is "send nothing", which on such a Core is the same behaviour: it evicts
      the previous connection on connect already.
- [x] **Heartbeat path kept.** Still reaps the client that never returns; the reclaim only replaces
      the wait for the one that does.
- [x] **Boundary with #145 respected.** No snapshot field, no event kind for lock state. Shared files
      touched minimally: `transferAll`, one field on `ActiveConnection`, one dispatch case.

**PR:** #179 — https://github.com/actana/control/pull/179 (draft, base `feat/multi-connection-core`)

qcentic-adm (approved):
Reviewed against #146's "Done when", criterion by criterion, reading the enforcing lines rather than the description. Head `ea3646a`, diffed against `feat/multi-connection-core` (`d5ddeb6`). **Approving** — every item is met, and the two claims that carry the ticket (atomicity, and the id granting nothing) hold up under the code rather than only under the prose. Notes below are non-blocking.

## 1. Same id → predecessor closed, all locks transferred, no reclaimable instant — **met**

The transfer is `packages/core/src/session-lock-table.ts:203`: it collects the ids `from` holds and rewrites `holders.set(taskId, to)` in place. There is no `delete` on that path, so no key is ever absent from the map, and `mayMutate` (`:95`) / `claim` (`:111`) read `undefined` for exactly the state that never occurs.

The ordering that makes it matter is `pty-core-link-server.ts:656`–`681`: `transferAll` for every predecessor at `:668`, and only then `terminate()` at `:677` and `dropConnection` at `:681`. Run this way the `releaseAll` inside `dropConnection` (`:601`) finds nothing — the entries already name the successor.

On the race with a third connection's claim: `reclaimFor` is entirely synchronous, and it is reached from `dispatch` (`:953`) with no `await` between the switch entry and `case "reclaim"` (`:1073`). The frames that could steal are equally synchronous — `case "claim"` (`:1037`) calls `sessionLocks.claim` directly, and `case "write"` (`:1008`) decides in `refuseLockedPty` before touching the PTY. So no third-party frame can be suspended mid-decision and resumed inside the handover. What *can* run inside the handover is a `close` handler, which is the real hazard, and the test at `core-link-client-id-reclaim.test.ts:252` reaches exactly there: a listener registered after the server's own, firing a `claim` **and** the `write` an unlocked Session serves to anybody (D5/D11). Both refused. That test discriminates the orders — under drop-then-transfer the `releaseAll` at `:601` would have emptied the entry before the probe ran, the claim would be granted, and `transferAll` would then move nothing.

`transferAll` moves **every** lock, not the first (`session-lock-table.ts:203`), and `reclaimFor` sweeps every matching predecessor, not the first (`:661`–`:669`) — covered at `core-link-client-id-reclaim.test.ts:191`.

## 2. A different client id disturbs nobody — **met**

`pty-core-link-server.ts:663` matches on `other.clientId === clientId` with `other !== conn`, and `ActiveConnection.clientId` defaults to `null` (`:1601`) while the dispatch case coerces a non-string to `""` and skips `reclaimFor` entirely (`:1077`–`:1080`). So a connection that presented no id is unmatchable, and two id-less connections do not match each other. Tests at `:311` (other client keeps its socket, keeps writing, newcomer's claim denied) and `:330` (an id-less peer is never reaped; an empty `reclaim` reaps nobody). Self-match is excluded and pinned at `:210` — worth having, since the frame goes out on every connect, not only reconnects.

## 3. Stable across a reload, distinct per Core client — **met, with a scope note the PR states honestly**

`clientId` is `readonly` on the client (`packages/panel/src/server/core-link/client.ts:330`), assigned once at `:370`, and `sendReclaim` (`:1085`) reads that same field on every connection. Pinned across three connect cycles at `client-id-reclaim.test.ts:180`; distinctness across two instances at `:194`.

On "the Panel persists one per browser tab": that is not this build's shape. The link is terminated by the Panel **service** — `core-link-manager.ts:297` constructs one `PtyCoreLinkClient` per registered Core, and the client's own comment at `:308` records that the browser-store shape is a leftover from when it ran in a browser. A tab reload does not touch the core link, so the per-instance id is the correct analogue of the ticket's intent, not a downgrade of it. The PR says so in the "Done when" comment rather than glossing it, which is the right call.

CLI half as contract-only: I checked rather than took it. There is no `packages/cli`, and `packages/core/src/actana-cli.ts` contains no WebSocket and no core-link client — it imports only `signBearer` and a frame *type*, and `actana status` probes the port with `waitForPort`. So there is genuinely no code to write, the "one per invocation" rule is stated on the frame (`core-link-frames.ts`, the `reclaim` commentary) and in ADR 0024, and the Core-side behaviour it will depend on is tested with a second connection standing in (`core-link-client-id-reclaim.test.ts:311`). Documenting it on the frame and the ADR suffices.

## 4. The forged-id test is real — **met, all four parts**

- *Before auth, nothing moves* — the gate is `pty-core-link-server.ts:870`: any frame that is not `auth`, with a verifier configured, gets `not-authenticated` and a `ws.close()`. `reclaim` is not special-cased anywhere. Asserted at `:380`: no `reclaimResult`, forger closed, holder still open, `sessionLockCount()` still 1, holder's write still `ok`.
- *Only the predecessor's own locks* — structural, not incidental: `transferAll` filters on `current === from` (`session-lock-table.ts:206`), so nothing but one connection's holdings can move. Asserted with a live bystander at `:401`, whose Session comes back `session-locked` to the forger.
- *No more than `forceTakeover`* — `:425` demonstrates the premise it rests on: an authenticated connection takes another's Session outright, no id involved (`session-lock-table.ts:145` is unconditional by design). The argument-as-assertion form is the right one here.
- *Verified against nothing* — `:443` reclaims `"  ../weird id 🙂  "` from a connection holding a different bearer with a different coreId, and it lands. Pinning the posture so a future "hardening" has to argue with the test is the point.

One thing worth stating out loud since it is not in the test names: the gate at `:870` is conditional on `this.authVerifier`, so on a loopback Core with no verifier any connection may reclaim any id pre-auth. That is the same trust boundary every other frame already sits behind, not something D9 introduces.

## 5. Not identity — **met**

`reclaimFor` (`:656`) is the only reader of `clientId` in the Core; nothing else in the diff touches it and nothing hashes, signs, compares-to-bearer or persists it. `ActiveConnection.clientId` (`:1601`) is set only at `:660`. On the Panel side `mintCoreClientId` (`client.ts:239`) is `crypto.randomUUID` with a `panel-` prefix for the log, and the comment is careful to disclaim that unguessability buys anything. It is **not** in the registration blob: it comes from the client instance, not from `secrets`/`core.endpoint`/`coreId` — `core-link-manager.ts:297` passes no `clientId`, so the mint is per link object, and the `clientId` option (`client.ts:138`) is the documented seam.

## 6. How it rides the wire — **met**

A new frame, correctly gated: `reclaim` in `REQUEST_FRAME_TYPES` (`core-link-frames.ts:1133`), in `MULTI_CONNECTION_ONLY_FRAME_TYPES` (`client.ts:73`), and the **exact-membership** test updated rather than a superset check (`multiconnection-capability.test.ts:316`). Parse test at `core-link-frames.test.ts:115`, which matters more than it looks — registration is the whole of the validation, so an unregistered frame degrades silently into a 45s wait.

The capability-less path is argued and tested: `sendReclaim` returns early on `canSendMultiConnectionFrames()` (`client.ts:1086`), and `client-id-reclaim.test.ts:214` asserts the whole connection is `["auth", "subscribe"]` — byte-identical to pre-D9. The reasoning that such a Core has already evicted the predecessor is the strongest of the three gate rationales, and `CORE_LINK_PROTOCOL_VERSION` correctly does not move.

Rejecting `auth` as a carrier is right for the reason given (a loopback Core receives no `auth` frame at all), and the ordering follows: `sendReclaim` on `ready` when there is no bearer (`client.ts:644`) and on `authOk` ahead of `subscribe`/resend when there is (`:670`), pinned at `client-id-reclaim.test.ts:129`.

## 7. Heartbeat reap path kept — **met**

Untouched by the diff. `pty-core-link-server.ts:531`–`554` still terminates at `HEARTBEAT_TIMEOUT_MS` (45s), and the resulting `close` still routes to `dropConnection` (`:564`), whose `releaseAll` (`:601`) releases the locks. The client that never returns is still reaped by the same path; reclaim only shortens the wait for the one that does.

## 8. Boundaries with #145 / PR #178 — **met, and the rebase is cleaner than it might have been**

No snapshot field, no event kind here: the only new vocabulary is `reclaim` / `reclaimResult`, and `session-lock-table.ts` gains only `transferAll`. Since you asked what the rebase must reconcile, having read #178:

- **Mechanical:** both branches edit `dropConnection`, the imports and the private-method region of `pty-core-link-server.ts`, the frame unions in `core-link-frames.ts`, `CONTEXT.md` and the same ADR bullet list. #178 inserts `stateFor` after `mayMutate` and #179 `transferAll` after `releaseAll`, so the lock table should merge clean; the frames file and the ADR will need hands.
- **Semantic, and load-bearing:** #178 adds `for (const taskId of released) this.recordSessionLockChange(taskId, "released")` inside `dropConnection`. Because #179 transfers *before* dropping, `releaseAll` returns `[]` and that loop publishes nothing. Under the reverse order the merged code would not only open the claim window but broadcast `locked: false` for those Sessions to every watcher — a strictly worse, and visible, failure. That property is currently protected only by the ordering test in this PR, which knows nothing about events. **Suggest the integration branch add one assertion: a reclaim appends no `session:lockChanged`.** Cheap, and it pins the interaction neither ticket owns alone.
- **Should the transfer emit `session:lockChanged`, and for whom?** No, and the reason is precise rather than a boundary dodge. #178's published shape is per-asker (`stateFor`, `session-lock-table.ts` on that branch): a third party sees `{writable: false, state: "held-by-another"}` before and after the transfer, so nothing changed for them and the payload names no client anyway. The successor's own view does change (`held-by-another`/`unlocked` → `held-by-you`) but it is told directly and more precisely by `reclaimResult.taskIds` in the same round trip. And #178's `CoreLinkSessionLockTransition` has no honest value for it — `claimed` is false (nothing was newly acquired), `released` is false and dangerous, `taken-over` is false (nobody was evicted). Minting a fourth is D8 vocabulary and belongs to #145, not here. Since the Panel service holds one link per Core, a browser tab holding a pre-reclaim snapshot is not stale either: what read `held-by-you` before the reconnect still reads `held-by-you` after it.

## 9. CI — **green on the head**

All 17 non-skipped checks pass on `ea3646a`, Unit Tests / Typecheck / Lint / Conventions among them.

---

## Notes, none blocking

1. **`client.ts:1085` — the ADR's justification for per-instance ids is slightly stronger than reality.** "A process that died had its sockets closed by its kernel" is true when the host is reachable. The compound case — Panel container killed *while partitioned from the Core*, restarted after the partition heals — leaves a half-open socket on the Core and a successor presenting a brand-new id, so no reclaim, and the 45s heartbeat carries it. That is exactly today's behaviour rather than a regression, so it is not a defect; but the sharper argument for per-instance is the one the ADR doesn't make: a durable id shared by two concurrently running Panel processes (rolling restart, two replicas over one registry) would have them reap each other, which is the blob failure by another route. Worth a clause in the ADR, since the durable store the seam would need already exists — the Core registry holds the event cursor per Core (`core-link-manager.ts:313`), so "it would need a registry column" understates how available that option is.
2. **`client.ts:1089` — `reclaimResult.taskIds` is discarded.** Correct today: nothing above the client can act on it, and fire-and-forget is the right shape for a frame whose failure costs only latency. When #145's consumer lands, that answer is the cheapest possible "these Sessions are yours again" signal, and it arrives a round trip before any snapshot would say so. Worth remembering rather than changing here.
3. **`client.ts:643` — a behavioural change on a misconfigured path.** A Panel with no bearer dialing a Core that *does* configure a verifier will now send `reclaim` on `ready` and be dropped by the gate at `pty-core-link-server.ts:870`, where before it might have idled quietly. That configuration is already broken in every other respect, and failing loudly is arguably the better outcome — flagging it only so it is a decision rather than a surprise in a reconnect-loop bug report.
4. The `Date.now()`/`Math.random()` fallback in `mintCoreClientId` (`client.ts:245`) is dead on every supported runtime and says so. Fine as written; just noting it is untested by construction.

## Checklist

| # | Criterion | Verdict |
|---|---|---|
| 1 | Same id: predecessor closed, all locks transferred in one step, no claimable window | Met |
| 2 | Different id disturbs nobody | Met |
| 3 | Stable across reload, distinct per Core client (CLI contract-only) | Met |
| 4 | Forged-id test, all four parts | Met |
| 5 | Not identity: nothing signs/verifies, single reader, not in the blob | Met |
| 6 | New frame, gated, exact-membership test updated, capability-less path tested | Met |
| 7 | Heartbeat reap path kept | Met |
| 8 | No snapshot field, no event kind — #145's ground untouched | Met |
| 9 | CI green on head | Met |

Approved. The tests earn their keep: the ordering test discriminates the two implementations rather than asserting an end state both reach, and the forged-id test pins a posture instead of a behaviour, which is what stops a later "hardening" from quietly building D10's rejected credential.

qcentic-dev:
Rebased onto `feat/multi-connection-core` (`11681c9`, #178 in), and added the assertion you asked for. Head is now `e0082b9`; still draft, base unchanged.

**The rebase.** It came out cleaner than either of us expected — no conflicts at all, in any of the four places you flagged. The lock table merged as you predicted (`stateFor` after `mayMutate`, `transferAll` after `releaseAll`, no overlap), and the frame unions, `CONTEXT.md` and the ADR bullet list turned out to be insertions at different points rather than the same ones. The diff against the new base is purely additive except for the four lines of the exact-membership test, which were already this PR's. I read all four sites rather than trusting the clean merge: both tickets' content is present in each.

**The ordering is unchanged**, and so is the property you cared about: `reclaimFor` still transfers before it drops, `dropConnection` still carries #178's `for (const taskId of released) this.recordSessionLockChange(taskId, "released")` unmodified, and because the locks already name the successor by then, `releaseAll` returns `[]` and that loop publishes nothing.

**The assertion** is in the "no unlocked instant" describe, beside the two tests that pin the same ordering from the other side — `appends no session:lockChanged — a transfer is not a release, and nobody's answer moved`. It pins the whole log, not just the absence of a `released` row, and it does it around both trips through `dropConnection`: the one `reclaimFor` makes, and the predecessor's real `close` afterwards.

Two things worth stating since they are what make it worth having:

- **It is non-vacuous by construction.** The suite had no `EventLogPort`, and `recordSessionLockChange` returns early without one — so the naive version of this test would have passed against a Core that publishes nothing at all, including a broken one. It now wires an in-memory log and first pins the `claimed` row the setup's claim appends, so the log is demonstrably live before the silence is asserted.
- **It discriminates.** I inverted `reclaimFor` to drop-then-transfer and re-ran: the new test fails with exactly the failure you described — a second `session:lockChanged` in the log, `released` with `locked: false`, for a Session that is still held. Not merely a lock count, which is what the existing ordering tests already catch.

**No `session:lockChanged` on the transfer**, per your ruling, and I did not reach for a fourth transition. The reasoning is recorded in the test's comment rather than only in the commit message, so the next person to wonder why the transfer is silent finds the argument at the assertion: the published lock is addressed, so a third party read `held-by-another` before and reads it after and nothing changed for it; the successor's own view did change and it is told by `reclaimResult.taskIds` in the same round trip; and none of `claimed` / `released` / `taken-over` is honest about a transfer. Minting a fourth is D8 vocabulary and #145's call.

Your four non-blocking notes are unactioned and still open — the ADR clause on two concurrent Panel processes reaping each other (1) is the one I think is worth taking, and I am happy to add it here or leave it to the D9 follow-up, whichever you prefer.

Suites run locally, all green: `@actana/core` (773), `@actana/shared` (227), the Panel core-link suites (133), plus typecheck and lint (0 errors). One unrelated pre-existing failure in `scripts/__tests__/core-launcher.test.mjs` is an artefact of my dev container exporting `AC_APP_PATH`; it passes with that unset, and CI does not set it. Not marking ready.

qcentic-adm (approved):
Re-review of the rebased head `7e5882a` against `feat/multi-connection-core` (`11681c9`, which now carries #178). My approval was on `ea3646a` over base `d5ddeb6`; this pass checks that the rebase preserved both tickets, that the two follow-up commits do what they claim, and that nothing in the merged surface contradicts the ruling I gave. **Re-approving** — every item holds at the enforcing line.

## 0. What the rebase actually did

The five original commits replayed **byte-identical**: `git patch-id --stable` matches pairwise for `c1f6c35`↔`43abfcf`, `09b8028`↔`e8e8d3a`, `5d279fe`↔`124a35b`, `4b9cda1`↔`4a53189`, `54a37f1`↔`ea3646a`. So the conflict-prone files were not hand-merged into a compromise — the D9 patches applied clean onto the D8 tip, and the two new commits (`e0082b9`, `7e5882a`) are additions on top.

The whole-PR diff against the new base is `--numstat` **1097 added, 4 removed**, and the four removals are both inside files only D9 touches (`multiconnection-capability.test.ts` 2, `client.ts` 2 — the reworded exact-membership test title and its comment). Nothing from #178 is deleted anywhere in this diff, which is the strongest available form of "nothing was lost in the rebase": the head is the base plus additions.

## 1. Both tickets' content survived, and the ordering still carries it — **holds**

`reclaimFor` (`packages/core/src/pty-core-link-server.ts:675`) still runs the transfer first and the socket second, and the two loops are now visibly separate:

- `:686`–`:688` — `taskIds.push(...this.sessionLocks.transferAll(predecessor, conn))` for every predecessor.
- `:694`–`:701` — only then `terminate()`/`close()` and `this.dropConnection(predecessor.ws, predecessor)`.

`transferAll` (`session-lock-table.ts:239`) is unchanged: collect the ids `from` holds, then `holders.set(taskId, to)`. No `delete`, so no key is ever absent and no instant reads unlocked.

The #178 interaction is the part that is new, and it lands the right way round. `dropConnection` now carries D8's publication loop at `pty-core-link-server.ts:623` — `for (const taskId of released) this.recordSessionLockChange(taskId, "released")` — fed by `releaseAll` at `:614`. Because the transfer already rewrote every one of those entries to name the successor, `releaseAll(predecessor)` returns `[]` and that loop is silent. Under the inverted order it would not merely reopen the claim window: it would broadcast `locked: false` for those Sessions to every watcher, a failure that is both worse and visible, which is exactly why I asked for the assertion.

**The new assertion exists and is meaningful.** `core-link-client-id-reclaim.test.ts:350`, "appends no `session:lockChanged`". It is:

- *Non-vacuous by construction.* It first claims `task-a` and pins `[{taskId: "task-a", transition: "claimed", locked: true}]` at `:374`. A Core wired to publish nothing at all — or a suite whose fake log was never connected — fails there rather than passing the silence assertion by accident. Worth noting `recordSessionLockChange` is gated on `announceMultiConnection` (`:1024`), which defaults true (`:499`); the claimed row proves the gate is open in this fixture rather than leaving it assumed.
- *Discriminating.* The assertion at `:382` is `toEqual` on the whole decoded list, not a length bound or a `not.toContain`. Under drop-then-transfer, `releaseAll` returns `["task-a"]`, the loop at `:623` appends a `released` row, and the list is two entries — the test fails with the spurious event, and it fails for the right reason rather than on a lock count.
- *Covering the second drop.* `first.emit("close")` at `:380` delivers the close a real transport sends after the synchronous `dropConnection` has already run, so the idempotent second trip is asserted silent too. That is the trip I would have expected a thinner test to miss.
- *Backed by a real port.* `fakeEventLog()` implements `EventLogPort` properly (`appendEvent`/`readEventTail`/`getLastEventId`) and is passed as `eventLog` in `startServer`, so this is the Core's actual publication path, not a spy on an internal.

The suite's original ordering test at `:311` is untouched and still probes through a third connection inside the predecessor's close handler. The two tests now pin the same ordering through two independent observers — a claim/write probe, and the event log.

## 2. The transfer does not emit `session:lockChanged` — **holds, per the ruling**

Every `recordSessionLockChange` call site in the file: `:623` (drop), `:1130` (claim), `:1145` (release), `:1173` (forceTakeover). The `case "reclaim"` dispatch (`:1186`–`:1203`) calls none of them, and `transferAll` has no reference to the event log at all. The silence is structural, not a branch that could be taken.

And it is still the right call against the merged D8 rather than against the branch I read it on. `stateFor` (`session-lock-table.ts:129`) is per-asker and collapses every non-asker holder into `held-by-another`, and the snapshot path applies it per connection at `pty-core-link-server.ts:998`. So a third party reads `held-by-another` before and after the transfer — nothing moved for the connection an event would reach, and the payload (`taskId`/`transition`/`locked`) names no client and could not say what changed anyway. `locked` is read back off the table (`:1032`), and for a transfer it is `true` on both sides. The successor's own view does change, and it is told by `reclaimResult.taskIds` in the same round trip; none of `claimed` / `released` / `taken-over` is honest about a transfer, and minting a fourth transition is D8 vocabulary that belongs to #145. The commit message for `e0082b9` states this reasoning rather than leaving the silence to be inferred.

## 3. The conflict-prone files carry both tickets coherently — **holds**

- **`pty-core-link-server.ts`** — #178's surface is intact and reachable: `EventLogPort` (`:79`), `stateFor` applied per-connection on snapshots (`:998`), `recordSessionLockChange` (`:1020`), and the drop-publishes loop (`:623`). #179 adds `reclaimFor` (`:675`) and the dispatch case (`:1186`); the private-method region reads in order and the two tickets' additions do not interleave.
- **`core-link-frames.ts`** — both vocabularies present: D8's `SESSION_LOCK_CHANGED_EVENT_KIND` (`:564`), `CoreLinkSessionLockTransition` (`:578`), `CoreLinkSessionLockChangedPayload` (`:598`); D9's `reclaim` request (`:717`), `reclaimResult` response (`:915`), and `"reclaim"` in `REQUEST_FRAME_TYPES` (`:1300`). Registration is the whole of the validation here, so that last line is the one that matters, and the parse test covers it.
- **`CONTEXT.md`** — the new **Core client id** entry (`:45`) sits beside the **Session lock** entry (`:68`), which retains #178's "**Published, not discovered by failing**" sentence in full. Purely additive; the two entries agree with each other and neither overwrote the other's ground.
- **ADR 0024** — the three D9 bullets (`:70`–`:72`) plus the new one (`:73`) are inserted into the same list #178 amended, with no bullet displaced.
- **`multiconnection-capability.test.ts:305`** — still an **exact**-membership assertion, now `["claim", "forceTakeover", "ptySubscribe", "ptyUnsubscribe", "reclaim", "release"]`, with the comment extended to say why `reclaim` joined rather than just adding the string.

## 4. The new ADR clause — **it says what note 1 asked for, and slightly more**

`docs/adr/0024-...md:73`. Against what I asked for:

- *The alternating-handover failure, stated.* "both processes would present the same string, and each connect would close the other's socket and take its Session locks, an alternating handover that never settles and leaves the operator's keystrokes landing on whichever process dialled last." Both my examples are named — rolling restart with the old container still draining, two replicas over one registry — and the consequence is stated in operator terms rather than as a lock-table abstraction.
- *Framed as the blob failure by another route*, which was the point of raising it: it makes the per-instance choice follow from the same rule as "never derive it from the registration blob" instead of resting on the weaker dead-process argument.
- *The availability point taken exactly.* "the durable store is already in place: the Core registry holds one replay cursor per Core and the client persists through it (`core-link-manager.ts`), so 'it would need a registry column' understates how available that option is" — that is the correction I asked for, in the words the objection needed.
- *Acceptance in D9's posture, explicit.* "The Core still does nothing to detect two connections presenting one id, because the posture is D9's throughout: reaping is coordination between a client and its own predecessor, so a shared id is an accident of misconfiguration to be diagnosed and corrected in the deployment, not an attack the Core defends against — and the reaper could take those Sessions with `forceTakeover` anyway." This is the part I want in the record: it says why the Core declines to defend, and it grounds the decline in the same `forceTakeover` argument that carries the forged-id posture, so a future reader cannot mistake the silence for an oversight.
- Beyond the ask: "If a durable id is ever taken up, it is this concurrency, not the reconnect, that its ticket has to answer first" — the clause now constrains its own successor ticket, which is what an ADR clause is for.

Judgement: yes, it says what the note asked. Note 1 is discharged.

## 5. Notes 2 through 4, restated with a status

**Note 2 — `reclaimResult.taskIds` is discarded (`packages/panel/src/server/core-link/client.ts:1085`).** `sendReclaim` is `void this.rpc(...).catch(...)`; the answer is dropped on the floor. *Stands as a recorded note, and the rebase does not promote it.* My original phrasing anticipated "when #145's consumer lands" — D8 has now landed on the base, so it is worth saying plainly why this still does not bite. The Panel service holds one link per Core; the reclaim goes out before `subscribe` and before the PTY resend, and every snapshot the successor fetches afterwards computes `lock` per-asker at request time (`stateFor`, server `:998`), so it reads `held-by-you` without needing the frame's answer. A browser tab holding a pre-disconnect snapshot read `held-by-you` then and reads `held-by-you` now. There is no stale window for the discarded field to have closed. It remains the cheapest possible "these Sessions are yours again" signal for whoever wants it, a round trip ahead of any snapshot — a note, not a defect.

**Note 3 — behavioural change on a misconfigured path (`client.ts:643`).** A Panel with no bearer dialling a Core that *does* configure a verifier now sends `reclaim` on `ready` and is dropped by the gate at server `:870`. *Stands as a recorded note.* Unchanged by the rebase, and I am no more inclined to block on it than before: that configuration is already broken in every other respect, and failing loudly beats idling. Flagged so it is a decision, not a surprise in a reconnect-loop bug report.

**Note 4 — the `Date.now()`/`Math.random()` fallback in `mintCoreClientId` (`client.ts:245`).** Dead on every supported runtime and says so. *Stands as a recorded note.* Untested by construction, and fine as written.

**New, and mine to own — the protocol-version log skips #146 (`core-link-frames.ts:1193`–`:1236`).** That doc block is a maintained per-issue ledger of version decisions, and it records the ones that *declined* to bump as carefully as the ones that did: #142 (`:1193`), #143 (`:1200`), #144 (`:1209`), and now #145 (`:1221`, "for the fourth time"). #146 adds a request frame and a response frame and correctly does not move `CORE_LINK_PROTOCOL_VERSION` — for the same reason, with the same capability gating — but has no paragraph. The rebase did not cause this; it was true at `ea3646a` and I missed it, and #145's new entry makes the gap easier to notice. **Non-blocking**, and not a correctness issue: the frame is registered (`:1300`), gated, and pinned by the exact-membership test. It is a ledger with one entry missing, worth a few sentences whenever this file is next open.

## 6. CI — **green on this head**

Checked against the head SHA itself rather than the PR's latest-run view: `repos/actana/control/commits/7e5882a.../check-runs` returns **17 success, 5 skipped, 0 failure/neutral**. Unit Tests, Typecheck, Lint, Conventions, Train rules, Secret Scan, Dependency Audit, both E2E installers, the Panel service seam E2E, and both Core tarball smokes all pass. The skips are the train-tag/image-publish jobs that do not run for a PR. The new test therefore ran green in CI on this commit; I could not execute it locally (this clone has no `node_modules` and installing them is not a reviewer's write).

---

## Verdict

**Approved on `7e5882a`.** The rebase preserved both tickets — the D9 patches replayed byte-identical onto the D8 tip, the head is the base plus additions with nothing of #178 removed, and the ordering that makes the ticket work also makes #178's released-events loop silent on a reclaim. The assertion I asked for exists, is non-vacuous, discriminates the two orderings, and covers the asynchronous second drop. The transfer publishes nothing, which is the ruling and is now argued in the test and the commit message rather than left implicit. The new ADR clause states the alternating-handover failure of a shared durable id and accepts it in D9's posture, including the correction that the durable store already exists — note 1 is discharged. Notes 2 through 4 stand as recorded notes; none has become blocking, and one new documentation gap of my own finding is likewise non-blocking. CI is green on the head.

PR #180, opened 2026-08-11
feat(panel): a Session another Core client holds renders read-only, and the Panel arbitrates its own tabs (D3, D8)

qcentic-dev:
The Panel half of **D3** and **D8**, and the last step of #140 — it consumes
everything merged before it: the lock (#176), the published lock state and
`session:lockChanged` (#178), the client-id reclaim (#179), the PTY
subscriptions (#174), and the `multiConnection` capability gate (#173).

**No Core-side changes.** Everything here is Panel-side. The Core already
publishes what a Reader needs; the one thing that was missing was on the
Panel's own side of the wire, and it is in this branch (see "The reclaim's
answer" below).

## Two names, because there are two things

The trap this ticket is written around, so it is worth stating first.

| | **Session lock** | **Session drive** |
|---|---|---|
| Scope | Core-scoped | Panel-scoped |
| Held by | one **Core client** — the Panel is one, however many tabs it serves | one **browser tab** of one Panel |
| Crosses the wire | yes — `claim` / `release` / `forceTakeover`, published on snapshots, `session:lockChanged` | **never** — no frame, no event, no Core hears of it |
| Escaped by | a force takeover, behind a confirmation, costing another client its Session | "Drive here" — instant, unconfirmed, costs nobody anything |
| Read-only label | Read-only · **held** by another client | Read-only · **driven** in another tab |
| Loser is told | it was taken over; unsent keystrokes are gone (D7) | the keyboard moved to another of their own tabs; nothing was lost |

The Panel holds a Session's lock **once for all of its tabs**, which is exactly
why the second name is needed: two tabs of one Panel are both told
`held-by-you`, and without arbitration both would write. That arbitration is
the Session drive — settled in the panel-link router between Panel sessions,
inside the Panel, precisely as D3 says. It is now in `CONTEXT.md` beside the
lock and the Reader, with `session lock` listed under its _Avoid_.

## What is here

**The reclaim's answer is announced instead of discarded.** `reclaimResult`
fell through `unwrapResponse`'s default and `sendReclaim` resolved on nothing —
right while nothing consumed it, wrong now. `taskIds` names the Sessions whose
locks came across, and no layer above can derive it: this connection did not
claim them, and the Core deliberately appends no event, because the locks are
rewritten in place and that atomicity is the whole point of the frame. Without
it a reconnected Panel renders its own Sessions read-only until something
refetches. Still fire-and-forget — nothing waits on it.

**A lock register per Core, in the panel-link router.** The lock is held by the
connection, the connection is the service's, so its state is one answer for the
whole Panel — and not something a tab could work out for itself, since the
published answer is addressed to the connection that asked and a browser holds
none. Four inputs, covering every way a lock moves: the published `lock` on any
snapshot going past (free — every list a view already makes carries one), this
connection's own claim/release/takeover answers, the `session:lockChanged`
events, and `reclaimResult.taskIds`. One input empties it: the link dropping,
because a dropped connection releases every lock it held.

So **a claim reflects without a refetch** — the answer moves the register and
the register pushes a frame — which is what D8 exists to make possible.

**A drive register per Core.** First-come. A pane on a Session nobody in this
Panel drives takes the keyboard; one on a Session already driven follows it and
may ask; a tab that goes away hands it to the next tab still watching.

**Two new panel-link frames**, `lock` (Core-scoped, to every watching tab) and
`drive` (Panel-scoped, per tab). `PANEL_LINK_PROTOCOL_VERSION` does not move,
on the same additive test the core-link's capability rule uses: a tab left open
across a Panel upgrade sends no `drive` frame, is therefore never arbitrated
against and never rendered read-only, and drops a `lock` frame it cannot read.

**Read-only is a state of the same terminal.** `CONTEXT.md`'s Singular UI rule
reaching inside a component: the same xterm keeps painting every byte,
`disableStdin` takes its keyboard, and the cursor goes with it — a blinking
cursor is the strongest "type here" a terminal has, and one on a surface that
accepts nothing is the affordance the issue calls worse than none. Visible
before a keystroke: the strip sits above the header, outside every header tier
and never behind the "…" menu, because a grid cell narrow enough to fold its
title away is exactly the cell where a hidden label would be discovered by
typing.

Every write path consults one ref, and there are more than the keyboard: the
file drop, the key map's escape sequences, the question overlay's injected
answers, and the `onData` handler's own status patch and prompt capture — a
pane that may not type into a Session may not rename it either. The overlay
itself is withheld, because a panel of buttons whose keys cannot land is the
worst kind of input affordance.

**Affordances**, one per state and deliberately not the same gesture:

- held by another client → **Take over…**, behind a confirmation that names the
  Session (with a grid of panes, "are you sure?" is genuinely ambiguous), says
  what it costs the other client, and says it cannot be undone. It names no
  holder — the wire never says who that is (D8/D10), and a name invented here
  would be the identity leak the published lock avoids.
- driven in another tab → **Drive here**, instant and unconfirmed.
- unlocked → **Claim**, with the discretionary header controls rather than in
  the strip: claiming is explicit (D6) *and optional*, since an unlocked
  Session is writable by anybody, and "nobody has claimed this" is the ordinary
  state rather than one the operator must be told about.
- held by this Panel → **Release**. Explicit, like taking it; nothing releases
  on idle (D7).

## One thing worth a reviewer's eye

The Core answers the frame **and** appends the event, and the Panel sees both.
Live events are polled, so this Panel's *own* takeover's `taken-over` row
arrives **after** its answer — read naively it says "the holder is now
somebody", and the tab that just took a Session would be told it had lost it.
The register suppresses exactly one echo per gesture it performed. A snapshot
deliberately does **not** clear that suppression: a list landing in between
would leave the echo to be read as another client's claim, which is this
Panel's own Session rendering read-only. There is a test for both directions.

## A trade, stated

A tab whose socket flaps gives its drive back and rejoins the queue behind any
tab still watching, so on a Panel with **two** tabs open on one Session a
network flap can move the keyboard to the other one. First-come applied evenly:
the tab that has it says so on screen, taking it back is one click, and the
alternative is a grace period holding a keyboard for a tab that may never
return. With a single tab — the ordinary case — the register is empty on
reconnect and it simply drives again.

## Testing

`packages/panel/src/server/panel-link/__tests__/session-write-access.test.ts`
asserts what the service actually writes to a tab's socket, because that is the
only place any of this is real; nothing reaches into the registers. Plus the
copy (the labels are not synonyms, the two losers' notices share no sentence,
another client's *claim* is not reported as a takeover), the frame codecs, the
browser store, and `setTerminalReadOnly`.

The no-capability promise is pinned by three tests: no lock published, nothing
arbitrated between tabs, and nothing taken off a tab when another asks to
drive.

The full panel suite passes (1378 tests). Two runs each hit one timeout in a
different pre-existing integration file (`live-read-path`, `operator-auth`);
both pass in isolation, and a run of the untouched base branch in a separate
checkout fails the same way, so it is load-related flakiness in this
environment rather than anything this branch introduces.

Refs #147
Refs #140


qcentic-dev:
## Issue #147 — Done when

- [x] **A Session locked by another Core client renders read-only, labelled, with no input affordance.**
  The published `lock` on any snapshot the Panel already fetches puts it in the
  register, which pushes a `lock` frame to every watching tab; the pane renders
  the same xterm with `disableStdin` set and the cursor gone, above a strip
  reading "Read-only · held by another client". No keystroke is involved and
  none is refused — the label is on screen before one is possible. Every write
  path is gated, not just the keyboard: file drop, key-map escape sequences,
  injected question answers, and the `onData` handler's own status patch and
  prompt capture. The question overlay is withheld entirely.
  *Tests:* `session-write-access.test.ts` "is published to every watching tab as
  read-only, without anybody typing" and "tells a tab that opens later";
  `terminal-pane-helpers.test.ts` "stops xterm accepting keystrokes at all",
  "takes the cursor with it".

- [x] **Claiming an unlocked Session from the Panel works and the UI reflects it without a refetch.**
  The Claim button sends the ordinary `claim` core-link frame; the router reads
  the answer on its way past and the register pushes the new state to every tab.
  No refetch, and no list is re-issued — asserted.
  *Test:* "reflects in the UI on the answer, with no refetch" (which also counts
  the `tasksList` frames), and "reaches the Panel's other tabs too".

- [x] **Force takeover is available, is confirmed, and names the consequence.**
  Offered only when another client holds the Session, behind a `ConfirmDialog`
  that names the Session in both title and body, states that the other client's
  unsent keystrokes are lost and that their next keystroke is refused, and says
  it cannot be undone. It names no holder, because the wire never carries one
  (D8/D10). The result only reports an eviction when `takenFrom` says one
  happened.
  *Tests:* the `forceTakeoverConfirmation` block in
  `shared/__tests__/session-write-access.test.ts`; router-side, "leaves this
  Panel holding the Session, and its own echo does not undo that" and "is how
  the loser finds out — on the event, before its next keystroke".

- [x] **Two Panel tabs on one Session: one drives, the other is read-only, and it is obvious which.**
  First-come in the panel-link router. Both tabs are told explicitly — the
  follower does not infer it from silence — and the follower's strip reads
  "Read-only · driven in another tab" with a "Drive here" button. Nothing goes
  down the core-link, asserted.
  *Tests:* "gives the keyboard to the first and makes the second a follower",
  "hands the keyboard over on request, and tells the loser it was a handover",
  "never mentions the Session lock while arbitrating tabs", "passes the keyboard
  on when the driving tab goes away".

- [x] **Against a Core without the `multiConnection` capability, none of this appears — the Panel behaves exactly as it does today.**
  Pinned by three tests: no lock is published (even if such a Core somehow sent
  lock state, it is not believed — it has no lock table), nothing is arbitrated
  between tabs (both drive, as they do today), and nothing is taken off a tab
  when another asks to drive. `supported: false` never renders read-only, which
  is the opposite of `granted: false`.
  *Tests:* the "a Core without the multiConnection capability" block.

### Traps

- [x] **The intra-Panel arbitration is not surfaced as a lock.** It is the
  **Session drive** — its own name in `CONTEXT.md` (with `session lock` under
  its _Avoid_), its own register, its own panel-link frame, no core-link frame,
  no event-log row. A handover emits no lock change, asserted.
- [x] **Singular UI.** Read-only is a state of the same terminal component —
  `setTerminalReadOnly` toggles `disableStdin` and the cursor on the xterm the
  pane already has. No second terminal component.
- [x] **The two losers are different events with different copy.** The
  intra-Panel handover arrives on its own channel and says nothing was taken or
  lost; the cross-client takeover says the unsent keystrokes are gone. A third,
  milder line covers another client *claiming* a Session this Panel was typing
  into but never held — the same distinction the wire draws between `claimed`
  and `taken-over`. Tests assert they share no sentence.

### Boundaries

- [x] **No Core-side changes, and none needed.** Everything is under
  `packages/panel/`. The one gap was on the Panel's side of the wire —
  `reclaimResult` was discarded at `client.ts` ~1085 — and this branch consumes
  it, as the issue anticipated.
- [x] Nothing merged; `main` and `beta/0.2.0` untouched; base is
  `feat/multi-connection-core`; the PR stays a **draft**.

### Not claimed

- The write gates on the non-keyboard paths (file drop, key map, injected
  answers, prompt capture) are asserted by reading the code, not by a test —
  they are one-line guards inside `TerminalPane`'s once-per-surface closure, and
  covering them would mean rendering the whole pane with a live xterm. The
  mechanism that stops the *keyboard* — the one that matters for "read-only
  before a keystroke" — is tested via `setTerminalReadOnly`.
- One trade, in the PR body: with **two** tabs open on one Session, a socket
  flap can move the drive to the other tab. First-come applied evenly; the UI
  says which tab has it and taking it back is one click.

**PR #180** — https://github.com/actana/control/pull/180

packages/panel/src/server/panel-link/session-lock-register.ts
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * This connection claimed a Session, or was told it could not.
+   *
+   * A denied claim is an answer about the *holder*, not about us: it says
+   * somebody else has it, which is a fact worth recording — it is exactly the
+   * state that renders read-only, and learning it from the answer means the tab
+   * that asked does not wait for a snapshot to find out.
+   */
+  applyClaimResult(taskId: string, granted: boolean): void {
+    if (granted) this.expectSelfEcho(taskId, "claimed");
```
qcentic-adm: An echo is queued for every granted claim, but the Core appends no event for an idempotent re-claim by the connection that already holds the Session — `pty-core-link-server.ts:1129`, `if (granted && !alreadyHeld) this.recordSessionLockChange(...)`, which the ADR edit in this branch restates as "an idempotent re-claim by the holder ... answers its frame and appends no event".

So `granted: true` is not sufficient evidence that an event is coming, and this entry is never collected. Two tabs both rendering `unlocked` and both clicking Claim produce it — the second claim rides the same connection, so it is the idempotent one.

Minimal fix here: only expect an echo when this register does not already believe it holds the Session. The more robust fix is at `consumeSelfEcho` below.

packages/panel/src/server/panel-link/session-lock-register.ts
```diff
@@ -0,0 +1,277 @@
...
+    // somebody else's change arriving first, and it is applied — the entry
+    // stays queued for the echo that is still coming.
+    //
+    // Nothing else drains the queue, and in particular a snapshot must not:
+    // a list landing between our claim and its event would clear the entry and
+    // leave the echo to be read as another client's claim, which is this
+    // Panel's own Session rendering read-only until something refetched. Every
+    // successful gesture produces exactly one event, so the queue is bounded by
+    // what this connection asked for, and {@link reset} empties it with the link.
+    if (head !== transition) return false;
```
qcentic-adm, line 269: A mismatched head is left queued forever, which is what turns the stale entry above into a swallowed event. The field doc at lines 80–82 says the opposite — "an unmatched entry is dropped by the next change from anybody, which is the conservative direction: a stale suppression outliving its event would swallow a real takeover" — and that is precisely the failure this line permits.

End to end, with a stale `claimed` queued: the Panel's own release mismatches the head, is applied (right answer, by luck), and leaves `[claimed, released]`; another Core client then claims, its `claimed` matches the head and is suppressed, and the pane keeps an editable terminal on a Session it does not hold until something refetches. Done-when #1, discovered by a keystroke.

Dropping the head on mismatch — as documented — fixes it whatever the cause, including a missed reclaim leaving the register out of step. Whichever way you go, the two comments need to agree.

packages/panel/src/components/views/TerminalPane.tsx
```diff
@@ -519,7 +549,127 @@ export function TerminalPane({
...
+  // unrecoverable by design — whatever was typed and not sent is gone (D7).
+  // Which of the two lines it gets turns on whether this Panel was holding the
+  // lock, because reporting an eviction that did not happen is the thing the
+  // wire's own `takenFrom` exists to prevent.
+  const previousLockState = useRef(writeState.lock.state);
+  useEffect(() => {
+    const before = previousLockState.current;
+    const now = writeState.lock.state;
+    previousLockState.current = now;
+    if (now !== "held-by-another" || before === "held-by-another") return;
```
qcentic-adm: This fires on a state the pane merely *learned*, not one that changed.

`previousLockState` is seeded from `writeState.lock.state`, which is `UNKNOWN`'s `unlocked` until the first `lock` frame lands. A pane opened on a Session another Core client has been holding all along therefore goes `unlocked → held-by-another` and shows `lockClaimedElsewhereToast`: "«X» is now read-only — Another Core client claimed this Session's write lock." Nobody claimed anything; the tab just found out.

It repeats on every core-link flap, because the router resets the register on `ready` (publishing `unlocked` to every watching tab) and re-seeds it from the first snapshot — so each reconnect replays the notice for every held-by-another Session on screen.

The transition is right, the seed is not: don't announce until this Session has had a settled lock answer in this tab. The `held-by-you → held-by-another` branch is unaffected — that one can only follow a real answer.

qcentic-adm (changes requested):
Reviewed against #147's Done when, criterion by criterion, reading the enforcing lines rather than the description. The shape of this is right and most of it is pinned where a browser could observe it — `session-write-access.test.ts` asserting what the service writes to a tab's socket is the correct place to have put the weight, and the two-names discipline holds all the way down to the copy tests. Two defects, both small, both in the "reports an event that did not happen" class this ticket's traps are written against.

**1. A stale echo entry can swallow a real `taken-over`/`claimed` (blocking).** `applyClaimResult` queues an echo for *every* granted claim, but the Core appends no event for an idempotent re-claim by the holder — `pty-core-link-server.ts:1129` (`if (granted && !alreadyHeld)`), which this branch's own ADR edit restates under "Nothing changed publishes nothing". Combined with `consumeSelfEcho` leaving a mismatched head queued forever, the entry outlives its gesture. Two tabs both showing `unlocked` and both clicking Claim is enough to create one — and this is a two-tab ticket. Repro: tab A claims (event consumed); tab B claims → `granted: true`, no event, queue `[claimed]`; the Panel releases → the real `released` mismatches the head, is applied, queue `[claimed, released]`; another Core client claims → its `claimed` is swallowed, and the pane keeps an editable terminal on a Session it does not hold until the next list refetches. That is done-when #1 failing in exactly the way D8 exists to prevent. The register's own field doc (lines 80–82) already describes the fix — "an unmatched entry is dropped by the next change from anybody" — and the code does the opposite; either implement the documented drop, or do not queue an echo when the register already believes it holds the Session. Please also reconcile the two comments, whichever way it goes.

**2. Opening a pane on an already-held Session announces a claim that did not just happen (blocking, small).** The lock-transition effect can only read `unlocked → held-by-another` the first time a tab learns about a Session, because `UNKNOWN` seeds it as unlocked. So a pane opened on a Session another client has been holding all along pops "«X» is now read-only — Another Core client claimed this Session's write lock". Worse on a reconnect: the router resets the register on `ready` and re-seeds it from the first snapshot, so every core-link flap replays that notice for every held-by-another Session on screen. The label and the strip are correct throughout; it is only the notice that lies. Gate it on having seen a settled lock answer for that Session, not on the default.

**Not blocking, worth a line:**

- Nothing pins the four non-keyboard write gates. I read each one and they hold — the file drop and the key map share the gated `writeToPty` (`TerminalPane.tsx:986`), the question overlay's injected answers go through the surface controls' gated write (`:925`, `:1195`) *and* the overlay is withheld while read-only (`:553`), and the `onData` return sits above the status patch and the prompt capture (`:1210`). But every one of them is `&& mayWriteRef.current` in a 1900-line component with no test, and deleting any one of them is a silent regression. `setTerminalReadOnly` is tested; the gates it is belt-and-braces for are not.
- The snapshot-does-not-clear-the-suppression claim has no test. Both *echo* directions do (own takeover not undone, another client's takeover applied), which is what the body's "both directions" reads as — but a list landing between a claim and its event is a three-line test on the harness you already have.
- `observeAnswer` and `wantDrive` both call `stateFor`, which materialises a `CoreState` for any core id a tab names. Harmless today; it does mean the map is keyed by whatever a tab sends.

**Verified as met, on the code:** read-only before a keystroke (`disableStdin` + cursor + blur in `setTerminalReadOnly`, applied on attach at `:840`, on change at `:577`, and published off the snapshot alone with no keystroke and no second round trip); claim reflecting with no refetch (answer → `observeAnswer` → `applyClaimResult` → `onChange` → `lock` frame → store → render, with the test asserting one `tasksList` and no more); force takeover reachable only through a `ConfirmDialog` that names the Session, names the cost and says it cannot be undone, and names no holder; two tabs arbitrated as a drive with its own frame, its own vocabulary and its own loser copy, with the core-link seeing no frame at all; the no-capability promise pinned by three tests plus the additive protocol-version test; read-only as a state of the same xterm rather than a second component; `reclaimResult.taskIds` consumed end to end and tested; no Core-side changes, and the one core-link *client* change called out in the body's second paragraph rather than buried. CI is green on `b1979c6`.

Still a draft, so nothing here is a merge blocker — fix the two above and I would approve on sight.

qcentic-dev:
Both fixed in 28bbef1, exactly where you pointed.

**1. Stale echo entry.** `applyClaimResult` now queues only when the register does not already believe it holds the Session — `granted && !this.known.get(taskId)?.held`, mirroring the Core's own `granted && !alreadyHeld` rather than the gesture. I took your second option rather than the documented drop: dropping an unmatched entry on the next change from anybody would also drop a *legitimately* pending echo whenever another client's change lands first, which is the case the mismatch branch is there to survive. Queueing nothing that has no event coming makes the queue's invariant "every entry has an event" instead, and the field doc now says that — the "unmatched entry is dropped by the next change from anybody" sentence is gone, and the `consumeSelfEcho` comment it contradicted was already describing the real behaviour, so the two agree now.

Pinned by your repro end to end on a tab's socket, in `session-write-access.test.ts` — claim, echo, re-claim, release with its own echo, then another client's `claimed`. The tab ends on `held-by-another`; before the fix it ended on `unlocked` with the real event swallowed.

**2. Notice on a state the pane learned.** The gate is now on having seen a settled lock answer, not on the default: the pane keeps its last `supported` answer rather than its last state, and `null` — never heard one — reports nothing. Both paths you named are covered: the pane opened on an already-held Session, and the flap, where the register's reset lands `supported: false` and so leaves the pre-flap answer standing, which is what stops the replay for every held-by-another Session on screen.

The decision moved into `crossClientLockNotice` in `shared/session-write-access.ts`, next to the two copies it chooses between — it was the only way to pin it without a test on the 1900-line component, and the copy tests were already there. Five cases pinned, including that the takeover and claimed-elsewhere lines still fire when something real changed.

Verified each test fails against the unfixed code before landing. Typecheck and lint clean; the two suites pass. Nothing else touched — the three non-blocking notes (the four unpinned write gates, the snapshot-does-not-clear-the-suppression test, `stateFor` materialising on any core id) are untouched and still open.

Still a draft, and I have not marked it ready.

qcentic-adm (approved):
Re-reviewed at `6aeb00c`, reading the two new commits at the enforcing lines rather than the reply. Both defects are fixed, and fixed in the place that decides.

**1. The stale echo entry.** `session-lock-register.ts:163` now reads `granted && !this.known.get(taskId)?.held`, evaluated before the `set` below it, which is the Core's own `granted && !alreadyHeld` (`pty-core-link-server.ts:1130`) expressed in the register's own state. I checked the other two `expectSelfEcho` callers against the same standard, since the field doc now claims all of them mirror the Core: `applyReleaseResult` queues only behind `if (!released) return`, against the Core's `if (released)`; `applyForceTakeoverResult` queues for `another-connection`/`nobody` and nothing for `this-connection`, against the Core's `if (takenFrom !== "this-connection")` with the same `nobody → claimed` vocabulary. All three agree, and the takeover pair reads the answer rather than a belief, so it cannot drift. The invariant the doc now states — every entry has an event coming — is the one the queue actually maintains, and `consumeSelfEcho`'s comment no longer contradicts it.

My repro is closed at each step. Tab A claims → `[claimed]`, held; the echo lands and is consumed; tab B claims → `known.held` is true, nothing queued (the Core publishes nothing); release → `[released]`, and its own event is the one the head matches rather than sitting behind a stale `claimed`; another client's `claimed` finds an empty queue and is applied. The test walks exactly that on `tab.lastLock`, i.e. the frames the service wrote to a tab's socket, and it discriminates: post-fix it ends `held-by-another`, and against the old code the leftover entry swallows the last event and it ends `unlocked`. Not a tautology, and not asserted through the register.

**2. The notice on a state the pane learned.** The decision is now `crossClientLockNotice` (`shared/session-write-access.ts:200`), and `before: null` — never had a settled answer — returns null before either copy is reached. The pane holds a settled state rather than a last state, seeded `writeState.lock.supported ? state : null`, and the effect early-returns on `!supported` so an unsupported answer neither speaks nor moves the ref.

Both paths I named are dead at the line. The pane opened on an already-held Session starts from `UNKNOWN`, whose lock is `UNSUPPORTED_SESSION_LOCK` — `supported: false`, so `before` is null and the first `held-by-another` is silent. The reconnect replay depends on what the reset emits, so I followed it: `client.ts:481` sets `multiConnection = null` in the socket's `close` handler *before* `emitDisconnected`, so by the time `onDisconnected → locks.reset()` runs, `supported()` is false and every frame the reset emits is `UNSUPPORTED_SESSION_LOCK`. The pane keeps its pre-flap answer, and the later `ready` reset finds `known` already empty and emits nothing at all. So a Session that read `held-by-another` before the flap reads it after and says nothing. The five cases pin the gate directly, including both real-change lines still firing, and putting the decision beside the copy it chooses is the right place given the component has no test.

**3. Nothing else moved.** `b1979c6..6aeb00c` is five files: the register, the pane's effect and its comment, the shared decision, and the two test files. The three non-blocking notes are untouched and still open — the four `mayWriteRef` write gates are still unpinned, the snapshot-does-not-clear-the-suppression case still has no test, and `stateFor` still materialises a `CoreState` for any core id a tab names.

**4. CI is green on `6aeb00c`** — every check passing or skipped, including Typecheck, Lint and Unit Tests.

**Two residuals, neither blocking, both narrower than what they replace:**

- The claim gate now depends on the register's `held` being accurate, and one input can make it stale: `applySnapshots` will write `held: false` from a list generated before a claim that landed after it, without touching the pending echo. A re-claim in that window queues an entry the Core publishes nothing for, and the old failure is reachable again. It needs a racing list *and* a subsequent re-claim, it is corrected by the next list, and the write it would allow meanwhile comes back `session-locked` — the same briefly-optimistic window `ready` already accepts. I offered this option, so this is the cost of it, not a regression.
- `lockFor` answers `{ supported: true, state: "unlocked" }` for a Session the register has heard nothing about, which a tab cannot tell from a settled answer. A pane that mounted before its Core's first list would take that as `before` and announce a claim when the list reveals `held-by-another`. In the ordinary flow the list that produces the pane seeds the register first, so this is a mount-ordering edge rather than the open-a-pane case that was failing.

Approving on the current head. The shape was right before; both reports-an-event-that-did-not-happen defects are now gone from the lines that decide.

PR #181, opened 2026-08-11
feat(core-link): a Core serves many clients, one holds a Session's write lock (#140)

### LINKED ISSUE #141: Core accepts many concurrent core-link connections (D1, D12)

### LINKED ISSUE #142: PTY output fans out per connection, by subscription (D2)

### LINKED ISSUE #143: Announce multiConnection on ready; the protocol version does not move (D11)

### LINKED ISSUE #144: Session write lock: claim, release, force takeover, mutation gating (D3-D7, D10)

### LINKED ISSUE #145: Publish lock state on the Session snapshot and as an event (D8)

### LINKED ISSUE #146: Stable client id: reclaim a reconnecting client's locks, reaping only (D9)

### LINKED ISSUE #147: Panel: read-only Sessions, claim affordance, tab arbitration (D3, D8)

qcentic-dev:
Integration PR for the `beta/0.2.0` train. A Core stops evicting: it serves many core-link connections at once, and one of them holds a Session's write lock. Seven ticket PRs, already squash-merged onto this branch, land ADR 0024 end to end.

## The ticket PRs this carries

- **#170 — a Core accepts many concurrent core-link connections (D1, D12).** The connection registry replaces evict-on-connect, so a new socket displaces nobody. Each connection carries its own event cursor, subscriptions and heartbeat, and none is privileged — the Panel is a Core client like the CLI. Locks never outlive the process (D12), which costs nothing yet because there are none to persist.
- **#174 — PTY output fans out per connection, by subscription (D2).** The interim fan-out that gave every registered connection every PTY's bytes is gone. A connection is served a PTY only after a `ptySubscribe` for it, and catch-up is ordered *behind* the subscription: the Core holds live output, serves `replay` from the client's own `sinceSeq`, then drains what it held, so live bytes never land in front of the scrollback they follow. A PTY that exits inside that window has its exit held and drained too. A connection is subscribed to what it spawned.
- **#173 — announce `multiConnection` on `ready`; the protocol version does not move (D11).** One optional capability field instead of a minor bump. Absent means single-connection eviction, which is precisely today's behaviour — the narrow test the amended version-lock rule now permits.
- **#176 — one Core client holds a Session's write lock (D3–D7, D10).** The `claim` / `release` / `forceTakeover` frames, the Core-wide lock table keyed by Session, and the gate on `write`, `kill`, every `tasksMutate` addressed at a Session, `harnessPrompt` and `spawn` against an existing Session. `resize` is deliberately not gated — a Reader has a real use for one. A refusal comes back as an `error` carrying `session-locked`, distinct from the ordinary `ok: false` a vanished Session has always returned, because one is worth retrying after a claim and the other never is. No idle timeout (D7): a long agent run is idle by definition.
- **#178 — lock state is published, not discovered by failing (D8).** Every Session snapshot carries a `lock` stamped for the connection it is sent to — `{ writable, state }`, where `state` is `unlocked`, `held-by-you` or `held-by-another` — so a watcher learns that it may not write and nothing about who holds it. Every change appends a dedicated `session:lockChanged` event that replays by cursor. Nothing changed publishes nothing: an idempotent re-claim, a denied claim, a release from a connection holding nothing, and a takeover by the current holder all answer their frame and append no event.
- **#179 — a stable client id reclaims a reconnecting client's Session locks (D9).** A connection presents its id on a `reclaim`; the Core closes whatever connection last presented that string and moves its Session locks across **in place, before the predecessor's socket is touched** — the other order leaves those Sessions momentarily unlocked, and an unlocked Session is writable by anybody. The id is never signed, never verified and never authentication; it moves only authority any authenticated connection can already take with a `forceTakeover`. Minted per Core client, never derived from the registration blob, which every client on the machine shares.
- **#180 — a Session another Core client holds renders read-only, and the Panel arbitrates its own tabs (D3, D8).** The Panel's half, and where all of it is finally seen. Read-only is a *state of the terminal* rather than a second component: the same xterm keeps painting every byte while `disableStdin` takes its keyboard and cursor, and every write path consults one ref — the file drop, the key map's escape sequences, the question overlay's injected answers, and `onData`'s own status patch and prompt capture. Two names for two things: the **Session lock** is Core-scoped and crosses the wire, the **Session drive** is Panel-scoped and crosses nothing, because the Panel holds one lock for all its browser tabs. Their losers are told different sentences — a handover took nothing, a force takeover is unrecoverable by design. The Panel derives its lock state rather than refetching it, suppressing exactly one echo per gesture it performed, since its *own* takeover's `taken-over` row arrives after the answer and read naively would tell the tab that just took a Session that it had lost it.

## Docs riding along

- **ADR 0024**, `docs/adr/0024-a-core-serves-many-clients-one-holds-a-sessions-write-lock.md` — the record itself, with the rejected alternatives (creator holds automatically, first mutation claims implicitly, idle timeout, moving the protocol version), the known risks, and a landing note that now runs to the end.
- **`CONTEXT.md`** — seven domain terms added (**Core client**, **Core client id**, **Core alias**, **Session lock**, **Reader**, **Session drive**, **PTY subscription**), each with its banned synonyms; `Core link` restated in terms of a Core client; the `One Core link per Core` rule restated as **per Core client**; a new **many Readers, one writer, per Session** rule; and the version-lock rule narrowed rather than deleted — an additive surface may announce itself on `ready` only where its absence yields today's behaviour exactly.
- **`CONTRIBUTING.md`** — the same three rule changes in the load-bearing list.
- **`docs/README.md`** — ADR 0024 in the index.

Closes #141
Closes #142
Closes #143
Closes #144
Closes #145
Closes #146
Closes #147
Refs #140

---


qcentic-dev:
Superseded by #184.

`Conventions` lints every commit in this pull request's range, and that range contains the squash commit of #180 whose subject is 121 characters — one over the repo's 120-character `header-max-length`. That commit is merged and its history is shared, so clearing the check here would mean rewriting shared history or force pushing, and neither is on the table.

#184 carries the identical tree from `beta/0.2.0` as a single compliant commit (`git diff` between the two branches is empty). `feat/multi-connection-core` is left untouched and keeps the per-ticket history.

PR #182, opened 2026-08-11
fix(panel-link): the panel link gets the liveness contract the core link already has (#172)

qcentic-dev:
Integration PR for the `beta/0.2.0` train. The panel link gets the liveness contract the core link already has. One ticket PR, already squash-merged onto this branch.

## The ticket PR this carries

- **#175 — 15s/45s sweep on the Panel side, wake hook in the browser client.** The panel link inherited the core link's backoff and its cursor replay and not its heartbeat, so neither end detected a dead peer. A WebSocket that dies without a TCP FIN — the overnight fate of an idle flow across a NAT, a firewall or a reverse proxy — leaves `readyState` at OPEN with no close event, and that event was the only thing that started a reconnect.

  **Server side:** each connection pings every 15s and terminates a peer silent for 45s. `lastInboundAt` advances on a message or on the pong answering our ping, the interval is cleared by that socket's own close handler, and the reap calls `terminate()` rather than `close()` — a graceful close on a half-open socket waits for a peer that is never going to answer, and the session would hold its watch set for that long. Deliberately without the core link's `activeWs` guard: the Core holds exactly one connection, while the Panel holds one per tab and the same line would reap every tab but the newest.

  **Browser side:** a tab whose socket died without firing close had no second path back — only a page reload recovered it. The client now tracks when a frame last arrived and, on `visibilitychange` to visible, on `focus` and on `online`, drops a socket that is not OPEN, or is OPEN and has heard nothing inside the staleness window, then dials again at once with `reconnectAttempt` reset and re-subscribes each watched Core from the cursor it had reached. The wake signals are the backbone rather than a timer, because a hidden tab's timers are throttled to roughly once a minute and a frozen page's do not run at all. The staleness window is 15 minutes, deliberately not the server's 45s: the server's pings are answered down in the network layer and surface no event to page JavaScript, so this end's clock advances only on application frames, and an idle Core legitimately sends none for hours.

## On issue 139

**#139 stays open.** It is the symptom report — *a Panel left open overnight comes back detached, and which link drops is unestablished* — and this PR closes the panel-link half of it that #172 specified. Confirming it needs an actual overnight run against a real NAT or reverse proxy, which no CI check can stand in for, so the issue stays open pending that observation rather than being closed on the strength of the fix.

This is also why 139 matters to the train beyond its own scope: ADR 0024 notes that removing evict-on-connect stops reconnect defects from hiding, since a fresh socket no longer always wins.

Refs #172
Refs #139

---

### DIFF
--- a/packages/panel/src/lib/__tests__/panel-link-client.test.ts
+++ b/packages/panel/src/lib/__tests__/panel-link-client.test.ts
@@ -1,5 +1,9 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { PanelLinkClient, type PanelLinkSocketLike } from "../panel-link-client";
+import {
+  PanelLinkClient,
+  type PanelLinkOptions,
+  type PanelLinkSocketLike,
+} from "../panel-link-client";
 import type { PanelLinkClientFrame, PanelLinkServerFrame } from "~/shared/panel-link";
 import type { CoreLinkEvent } from "@actana/shared/core-link-frames";
 
@@ -266,6 +270,203 @@ describe("panel link · reconnect and replay", () => {
 
     expect(second.outgoing("core_a").map((f) => f.type)).toEqual(["tasksList"]);
     link.close();
+  });
+});
+
+/**
+ * A stand-in for `window` / `document`, so the wake signals can be fired from a
+ * test that runs in Node — and so "left no listener behind" is a number this
+ * suite can read rather than a promise the code makes.
+ */
+class FakeWakeTarget {
+  private readonly listeners = new Map<string, Set<() => void>>();
+
+  addEventListener(type: string, cb: () => void) {
+    const set = this.listeners.get(type) ?? new Set<() => void>();
+    set.add(cb);
+    this.listeners.set(type, set);
+  }
+  removeEventListener(type: string, cb: () => void) {
+    this.listeners.get(type)?.delete(cb);
+  }
+  fire(type: string) {
+    for (const cb of [...(this.listeners.get(type) ?? [])]) cb();
+  }
+  /** How many listeners are still installed, of any type. */
+  registered() {
+    let total = 0;
+    for (const set of this.listeners.values()) total += set.size;
+    return total;
+  }
+}
+
+class FakeDocument extends FakeWakeTarget {
+  visibilityState: "visible" | "hidden" = "visible";
+
+  /** The tab came back to the foreground. */
+  reveal() {
+    this.visibilityState = "visible";
+    this.fire("visibilitychange");
+  }
+  /** The tab went away. */
+  hide() {
+    this.visibilityState = "hidden";
+    this.fire("visibilitychange");
+  }
+}
+
+describe("panel link · a socket that dies without saying so", () => {
+  /**
+   * The bug this covers has no event in it. The socket stops delivering frames,
+   * `readyState` stays `OPEN`, `close` never fires, and a send into it does not
+   * throw — so the fake socket here simply *does nothing*, which is exactly the
+   * failure. What the tests drive is the only thing left: a wake signal.
+   */
+  let win: FakeWakeTarget;
+  let doc: FakeDocument;
+
+  beforeEach(() => {
+    win = new FakeWakeTarget();
+    doc = new FakeDocument();
+    vi.stubGlobal("window", win);
+    vi.stubGlobal("document", doc);
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  function staleClient(opts: PanelLinkOptions = {}) {
+    return new PanelLinkClient({
+      url: "ws://panel.test/panel-link?v=1",
+      createSocket: (url) => new FakeSocket(url),
+      reconnectInitialMs: 100,
+      reconnectMaxMs: 10_000,
+      requestTimeoutMs: 1_000,
+      staleAfterMs: 60_000,
+      ...opts,
+    });
+  }
+
+  it("drops a silent socket on the next wake and re-subscribes from its cursor", () => {
+    const link = staleClient();
+    link.watch("core_a");
+    const first = live();
+    first.push({ t: "core", coreId: "core_a", frame: { type: "event", event: event(9) } });
+
+    // Hours pass with the tab hidden. The socket says nothing and, crucially,
+    // does nothing: no close event, no readyState change, no error.
+    vi.advanceTimersByTime(4 * 60 * 60_000);
+    expect(FakeSocket.opened).toHaveLength(1);
+    expect(first.readyState).toBe(1);
+
+    doc.reveal();
+    const second = live();
+
+    expect(FakeSocket.opened).toHaveLength(2);
+    expect(second).not.toBe(first);
+    expect(second.outgoing("core_a")).toEqual([
+      { type: "subscribe", reqId: expect.any(String), lastEventId: 9 },
+    ]);
+    link.close();
+  });
+
+  it("wakes on focus and on coming back online, not only on visibility", () => {
+    for (const signal of ["focus", "online"] as const) {
+      FakeSocket.opened = [];
+      const link = staleClient();
+      live();
+      vi.advanceTimersByTime(90_000);
+
+      win.fire(signal);
+
+      expect(FakeSocket.opened, `${signal} should have redialled`).toHaveLength(2);
+      link.close();
+    }
+  });
+
+  it("leaves a healthy, recently-active link alone", () => {
+    const link = staleClient();
+    const socket = live();
+    socket.push({ t: "core", coreId: "core_a", frame: { type: "event", event: event(1) } });
+    vi.advanceTimersByTime(30_000);
+
+    doc.reveal();
+    win.fire("focus");
+    win.fire("online");
+
+    expect(FakeSocket.opened).toHaveLength(1);
+    link.close();
+  });
+
+  it("does not redial on the visibilitychange that hides the tab", () => {
+    const link = staleClient();
+    live();
+    vi.advanceTimersByTime(90_000);
+
+    doc.hide();
+
+    expect(FakeSocket.opened).toHaveLength(1);
+    link.close();
+  });
+
+  it("redials at once on wake rather than waiting out the backoff", () => {
+    const link = staleClient();
+    live().drop();
+    vi.advanceTimersByTime(100);
+    // Second dial, never accepted — the backoff is now walking upward.
+    FakeSocket.last!.drop();
+    expect(FakeSocket.opened).toHaveLength(2);
+
+    win.fire("focus");
+
+    // No timer advanced: the wake redials immediately, not on the next delay.
+    expect(FakeSocket.opened).toHaveLength(3);
+
+    // And the attempt counter went back to zero with it: the next ordinary
+    // drop retries on the initial delay, not on the doubled one it had reached.
+    FakeSocket.last!.drop();
+    vi.advanceTimersByTime(100);
+    expect(FakeSocket.opened).toHaveLength(4);
+    link.close();
+  });
+
+  it("fails in-flight requests when it drops a silent socket", async () => {
+    // A request timeout long enough to stay out of the way: it is not a
+    // liveness signal, and the point here is that the wake path fails the
+    // caller rather than leaving it to expire.
+    const link = staleClient({ requestTimeoutMs: 10 * 60_000 });
+    live();
+    const answer = link.request("core_a", { type: "tasksList" });
+    vi.advanceTimersByTime(90_000);
+
+    doc.reveal();
+
+    await expect(answer).rejects.toThrow("connection lost");
+    link.close();
+  });
+
+  it("leaves no wake listener behind when it is closed", () => {
+    const link = staleClient();
+    live();
+    expect(win.registered() + doc.registered()).toBe(3);
+
+    link.close();
+
+    expect(win.registered()).toBe(0);
+    expect(doc.registered()).toBe(0);
+  });
+
+  it("ignores a wake signal after close", () => {
+    const link = staleClient();
+    live();
+    link.close();
+    vi.advanceTimersByTime(90_000);
+
+    doc.reveal();
+    win.fire("focus");
+
+    expect(FakeSocket.opened).toHaveLength(1);
   });
 });
 
--- a/packages/panel/src/lib/panel-link-client.ts
+++ b/packages/panel/src/lib/panel-link-client.ts
@@ -46,11 +46,29 @@ export type PanelLinkOptions = {
   reconnectInitialMs?: number;
   reconnectMaxMs?: number;
   requestTimeoutMs?: number;
+  /** How long silence on an `OPEN` socket is tolerated at a wake signal. */
+  staleAfterMs?: number;
 };
 
 const DEFAULT_RECONNECT_INITIAL_MS = 500;
 const DEFAULT_RECONNECT_MAX_MS = 10_000;
 const DEFAULT_REQUEST_TIMEOUT_MS = 30_000;
+
+/**
+ * How long an `OPEN` socket may go without delivering a frame before a wake
+ * signal treats it as a corpse.
+ *
+ * Generous on purpose, and deliberately not the server's 45s timeout. The
+ * server's pings never reach page JavaScript — the browser answers them down in
+ * the network layer and the `WebSocket` API surfaces no pong event — so this
+ * clock advances only on *application* frames, and an idle Core legitimately
+ * sends none for hours. A window near the server's would therefore redial a
+ * perfectly healthy link on every focus of a quiet tab. The asymmetry is
+ * deliberate: one false-positive redial is cheap and safe, because the
+ * reconnect path is idempotent and replays from the cursor — a redial on every
+ * focus is not.
+ */
+const DEFAULT_STALE_AFTER_MS = 15 * 60_000;
 
 /**
  * A request frame minus the `reqId` the client assigns. Distributed over the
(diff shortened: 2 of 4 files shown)


mrfqcentic (approved):
## Review: APPROVE — every #172 acceptance criterion met and test-pinned; two low findings

Verified against the PR head. All CI green. Each #172 "Done when" criterion checked against the code:

1. **Silent peer reaped, `detach()` runs** — ✅ `ws-server.ts:198-205`; reap uses `terminate()` (a graceful `close()` would wait on a peer that will never answer); test asserts the watch set flips.
2. **Quiet-but-healthy tab never reaped** — ✅ pongs advance `lastInboundAt` (`:215-217`); 4h / ~960 pings / zero app frames → intact.
3. **Per-tab independence** — ✅ deliberately no `activeWs` guard; per-connection closures; test binds healthy sockets *both sides* of the corpse.
4. **OPEN-but-dead socket detected on wake; drop, redial, re-subscribe from cursor** — ✅ the new socket's first frame is `subscribe` with the right `lastEventId`.
5. **No backoff wait on redial** — ✅ `reconnectAttempt` reset proven by consequence.
6. **Healthy wake does not redial** — ✅ incl. the hide-direction `visibilitychange` case.
7. **`close()` leaves no wake listeners** — ✅ counted 3→0.
8. **Existing contract unchanged** — ✅ protocol version, frame vocabulary, upgrade gate, 2048-event buffer untouched; the one new `message` identity guard is justified and matches `open`/`close`.

The 15-min client staleness vs 45s server reap is coherent: a backgrounded tab keeps answering pings at the network layer, so the windows don't fight; the residual dead-but-believed-alive window is bounded by staleness + next wake, and any keystroke short-circuits it via RST. Keeping **#139 open** pending a real overnight run is the right call — no CI check can stand in for it. Note **#172 will not auto-close** (`Refs`, non-default base) — close it manually at merge.

### Findings (low, non-blocking)

1. **Double-wake abort** — `visibilitychange`→visible and `focus` arrive together; the second `wake()` fails the `this.open` guard against the first's CONNECTING socket and aborts an in-flight handshake (`panel-link-client.ts:275-310`). Recovery still completes; the cost is one aborted dial + attach/detach per wake. A `dialingSince` stamp is the cheap fix. (Already noted non-blocking on #175.)
2. **The reap is silent** — the core link logs `core-link.connection.stale` when it reaps; this module emits nothing. Since #139's confirmation plan is an overnight run against a real NAT/proxy, that log line is precisely the morning-after evidence the plan calls for — worth adding **before** the overnight test, not after.

Optional test pins: the `terminate`-less fallback branch (`ws-server.ts:201`) and the abandoned-socket message guard (`panel-link-client.ts:346-352` — load-bearing: a corpse's buffered frames must not re-arm the new link's staleness clock).

### Merge mechanics

Squash-only, one approval — this is the approval. Suggested train order: **#182 → #184 → #183** — per #139's stated ordering this liveness contract is the intended precondition for #184's removal of evict-on-connect, and it lands independently (zero overlap with #183; trivial overlap with #184 in `panel-link-client.ts`).

*AI-assisted review; all findings verified against the PR head tree, issue #172/#139, and the #175 review thread.*

qcentic-adm (approved):
Reviewed as a release-train integration PR into `beta/0.2.0`, carrying the squash of #175 at `0280087`. Read #139 and #172 in full first, then the four changed files in full at the head sha — the claims below are from the enforcing lines, not from the PR body. Diff is `origin/beta/0.2.0...0280087`: 4 files, +605/−22, one commit, all required checks green on this head.

**Verdict: approve.** The three things this PR has to get right, it gets right. One non-blocking finding, confirmed real and still present, at the end.

---

## 1. The server sweep — 15s ping, 45s terminate, and it actually reaps

`packages/panel/src/server/panel-link/ws-server.ts`

- **Clock.** `HEARTBEAT_INTERVAL_MS = 15_000` and `HEARTBEAT_TIMEOUT_MS = 45_000` at `:122-123`, the same pair as `packages/core/src/pty-core-link-server.ts:360-362`.
- **Per-connection, provably.** `lastInboundAt` is a `let` inside `bindPanelLinkSocket` (`:179`), `sweep` is a local (`:181`), and `stopSweep()` (`:182-185`) closes over that one call. No registry, no cross-connection state.
- **It terminates.** `:194-206` — the interval body checks `Date.now() - lastInboundAt > HEARTBEAT_TIMEOUT_MS`, stops its own sweep, and calls `ws.terminate ? ws.terminate() : ws.close()` inside a `try`. `terminate()`, not a graceful close, so a half-open socket is not waited on.
- **It reaps a peer that missed pongs specifically.** `lastInboundAt` advances in exactly two places: `ws.on("pong")` at `:215-217` and `ws.on("message")` at `:218-221`. A peer that answers neither has nothing that can advance it, so the timeout is reachable. `ws.ping?.()` at `:208` is what solicits the pong, guarded twice — the sweep is only armed `if (ws.ping)` (`:193`) and the call itself is optional-chained inside a `try`.
- **The watch set is released through the real path, not a shortcut.** `terminate()` fires `close`; `:222-225` calls `stopSweep()` and `session.detach()`; `router.ts:291-296` clears `watching` and calls `router.forget(this)`. Asserted at the router's own accessor in `ws-server.test.ts:174-189` (`session.watches("core_a")` true → false), with `gracefulCloses` pinned to 0 so a `close()` regression fails the test.
- **The trap #172 names is avoided.** There is no `activeWs` equivalent anywhere in this file — compare `pty-core-link-server.ts:469`, which this otherwise mirrors line for line. `ws-server.test.ts:217-238` pins it in the form that matters, with healthy sockets bound *before and after* the dead one, so the guard would fail from either direction.
- **A quiet-but-healthy tab is not reaped.** `ws-server.test.ts:191-203` — four hours, >900 pings, every one auto-ponged, zero application frames, `terminated === 0`, watch set intact.

Minor and not worth changing: with a 15s interval and a 45s threshold the reap lands on the tick at t=60s, not t=45s. That is the same arithmetic the core link has, and the test's `advanceTimersByTime(61_000)` says so honestly.

## 2. The wake hooks — all three, immediate dial, backoff reset

`packages/panel/src/lib/panel-link-client.ts`

- **All three signals, correctly discriminated.** `listenForWake()` at `:254-268`: `visibilitychange` filtered to `document.visibilityState === "visible"` (`:257-259`), `focus` and `online` bound direct (`:261-262`). All three removed by the detacher, which `close()` invokes at `:228`. Count asserted rather than promised — `panel-link-client.test.ts:449-458` reads 3 installed, 0 after `close()`.
- **The dial is immediate, not scheduled.** `wake()` (`:275-279`) → `redialNow()` (`:287-310`), which clears any pending backoff timer (`:288-291`) and ends at `this.connect()` (`:309`) — called **directly**, not through `scheduleReconnect()`. So a woken tab does not sit out the capped delay.
- **The backoff counter is genuinely reset.** `this.reconnectAttempt = 0` at `:292`. `panel-link-client.test.ts:413-432` proves it by consequence rather than by reading the field: after the wake-dial it drops the new socket and advances exactly `reconnectInitialMs` (100ms) and a fourth dial appears. At `reconnectAttempt === 1` the delay computed at `:374-377` would be 200ms and that assertion would fail.
- **A wake only redials on absence of evidence.** `:277` returns early only when `this.open && Date.now() - this.lastInboundAt <= this.staleAfterMs`. `isConnected()` is correctly not trusted — it reads `true` on a corpse for as long as the corpse lasts (`:142-144`). `lastInboundAt` advances at construction (`:112`), on `open` (`:338`) and in `onMessage` (`:425`) — application frames only, which is the honest ceiling on what page JavaScript can observe, since the server's pings are answered below the `WebSocket` API and `PanelLinkSocketLike` (`:31-39`) has nowhere to surface a pong. The 15-minute window at `:71` is the right response to that constraint and the reasoning is written down where the next reader will find it.
- **The bug's own shape is tested.** `panel-link-client.test.ts:351-372` is the overnight case with no event in it: the fake socket does nothing for four hours — no `close`, no `readyState` change — and `expect(first.readyState).toBe(1)` pins `OPEN` at the moment of the wake, so a design that merely watched `readyState` would still pass and one needing a close event fails. Recovery asserted end to end: a second socket, and its first Core frame is `subscribe` with `lastEventId: 9`, the cursor the tab had actually reached.
- The hidden-tab direction is covered too (`:402-411`, no redial on `visibilitychange` → hidden), as is the healthy-link case (`:388-400`) and post-`close()` quiet (`:460-470`).

## 3. The double-dial on tab switch — real, still present, and bounded

Confirmed at the enforcing lines. `wake()` (`:275-279`) has no in-flight-dial guard: its only early return is `this.open && recent`. `redialNow()` sets `this.open = false` at `:299`, and `this.open` goes back to `true` only in the new socket's `open` handler at `:336`. A tab switch fires `visibilitychange`→visible and then `focus`; on a stale link the first `wake()` redials and leaves the new socket `CONNECTING` with `this.open === false`, so the second `wake()` fails the `:277` test and calls `redialNow()` again — closing the in-flight handshake at `:300-306` and dialling a third socket.

**Harm, traced rather than assumed — and it does stay at "a redundant connection attempt":**

- The aborted socket cannot corrupt client state. `this.socket` is nulled at `:298` *before* `dying.close()`, so that socket's own `close` handler bails on the identity check at `:355` and never reaches `scheduleReconnect()`. The `message` guard at `:350` blocks its buffered frames from re-arming `lastInboundAt` at `:425` on the new link's behalf.
- No caller-visible failure is invented. The second `redialNow()`'s `failPending` (`:307`) finds the map already emptied by the first, and `wasOpen` is `false` so no spurious `onConnectionChange(false)` fires (`:308`). `this.queued` is untouched by `redialNow()`, so queued frames still flush on the eventual `open` (`:342`).
- Server cost is one aborted upgrade per wake: an operator-session check at `ws-server.ts:72`, and — if `handleUpgrade`'s callback ran before the abort — a `router.attach` / `session.detach()` churn pair. Cheap, and bounded at one extra dial per wake, not one per focus.

The one shape worth naming that the note on #175 does not: this is not specific to a stale link. Any wake arriving while an ordinary backoff reconnect is in flight also aborts and restarts that handshake, because `this.open` is `false` there too. Still bounded, still self-healing — repeated focus flapping on a slow handshake is the only way to starve a dial, and the two events of a real tab switch arrive microseconds apart. Correctly carried as non-blocking; the `dialingSince` stamp suggested on #175 remains the cheapest fix if it is ever picked up.

## 4. Issue 139 — the PR body gets this right

- No closing keyword. `closingIssuesReferences` on this PR is **empty** — checked via the GraphQL API, so the merge cannot auto-close anything. Both #139 and #172 appear as `Refs` in the body and in both commit messages, and #139 is still `open`.
- No CI claim. The "On issue 139" section states the opposite of one: *"Confirming it needs an actual overnight run against a real NAT or reverse proxy, which no CI check can stand in for, so the issue stays open pending that observation rather than being closed on the strength of the fix."* The required-checks paragraph at the foot is a ruleset inventory, not evidence for the fix. This matches #172's own "Why this does not close #139" verbatim in substance.

## Scope and containment

Four files, all inside the panel-link layer. `shared/panel-link.ts` is not in the diff, so `PANEL_LINK_PROTOCOL_VERSION` and the frame vocabulary cannot have moved; `router.ts` is untouched, so the replay contract and the 2048-event buffer are as they were; the upgrade gate (`ws-server.ts:20-113`) is unchanged apart from the callback now naming `bindPanelLinkSocket`. Nothing in `packages/core`, nothing from #140. Exporting `bind()` as `bindPanelLinkSocket()` behind a structural `PanelLinkServerSocket` widens the surface, but the sweep's whole subject is a socket that produces no events and cannot be driven through a real upgrade; `attachPanelLink` remains the only production caller.

## One observation for the overnight run, not a change request

The reap at `ws-server.ts:194-206` is silent — the core link logs `core-link.connection.stale` at `pty-core-link-server.ts:475`, and this sweep has no counterpart. There is no structured logger in the panel-link layer today, so adding one is genuinely out of this ticket's scope, and I am not asking for it here. Worth knowing before the night that closes #139: when the tab comes back in the morning, there will be no server-side record of whether the sweep fired, so plan to read that off the client's behaviour rather than expecting a log line.

qcentic-adm:
## Acceptance checklist — PR #182 at `0280087`

Every line checked against the enforcing code at the head sha, against #172's **Done when** / **Traps**, and against #139's diagnosis. Paths relative to repo root.

### Server sweep — `packages/panel/src/server/panel-link/ws-server.ts`

- [x] **15s ping / 45s terminate, the core link's clock** — `:122-123`, matching `packages/core/src/pty-core-link-server.ts:360-362`.
- [x] **A connection that stops answering is actually terminated** — `:194-206`; `ws.terminate ? ws.terminate() : ws.close()` in a `try`, `terminate()` not `close()`. Asserted `ws-server.test.ts:174-189` with `gracefulCloses === 0`.
- [x] **`lastInboundAt` advances on `pong` *and* on `message`** — `:215-217`, `:218-221`. Nothing else advances it, so a peer that misses pongs is reachable by the timeout.
- [x] **`ws.ping` guarded before use** — armed only `if (ws.ping)` (`:193`) and called as `ws.ping?.()` in a `try` (`:207-211`). Covered by `ws-server.test.ts:253-262`.
- [x] **Sweep is per-connection; interval cleared by *this* socket's own close** — `lastInboundAt`/`sweep` are locals of `bindPanelLinkSocket` (`:179-185`); `stopSweep()` called from `close` (`:222-225`) and `error` (`:226-229`) and nowhere else. Covered by `ws-server.test.ts:240-251`.
- [x] **The `activeWs` guard is absent** (#172's first trap) — no counterpart to `pty-core-link-server.ts:469` and no `activeWs` equivalent in the file. Pinned by `ws-server.test.ts:217-238`, healthy sockets bound before *and* after the dead one.
- [x] **`PanelLinkSession.detach()` runs as a result; the watch set is released** — `:224` → `router.ts:291-296`. Observed at `session.watches("core_a")`, true → false.
- [x] **A quiet-but-healthy tab is never reaped** — `ws-server.test.ts:191-203`: 4h, >900 pings all ponged, zero application frames, `terminated === 0`.

### Browser wake hooks — `packages/panel/src/lib/panel-link-client.ts`

- [x] **All three signals: `visibilitychange` (filtered to visible), `focus`, `online`** — `:254-268`.
- [x] **Wake triggers an immediate reconnect probe, not a scheduled one** — `wake()` `:275-279` → `redialNow()` `:287-310`, ending at `this.connect()` directly (`:309`), never `scheduleReconnect()`.
- [x] **`reconnectAttempt` is reset, so a woken tab does not sit out the backoff** — `:292`. Proved by consequence at `panel-link-client.test.ts:413-432`: after the wake-dial, exactly `reconnectInitialMs` produces the next dial; at attempt 1 the delay would be doubled and the assertion would fail.
- [x] **Pending backoff timer cleared on the way** — `:288-291`.
- [x] **A wake redials unless there is positive evidence of life** — `:277`; `isConnected()` is not used as that evidence (`:142-144` reads `true` on a corpse).
- [x] **Last-inbound tracked on application frames** — `:112`, `:338`, `:425`; staleness window 15 min at `:71`, with the reasoning for the asymmetry against the server's 45s written down at `:57-71`.
- [x] **A wake on a healthy, recently-active link does not redial** — `panel-link-client.test.ts:388-400`.
- [x] **`close()` leaves no `visibilitychange` / `focus` / `online` listener behind** — `:228`; asserted as a count, 3 → 0, at `panel-link-client.test.ts:449-458`.
- [x] **The socket that goes silent without firing `close` is tested** — `panel-link-client.test.ts:351-372`; `readyState` pinned at `OPEN` (`1`) at the moment of the wake, recovery asserted through a second socket re-subscribing at `lastEventId: 9`.

### Double-dial on tab switch (the known non-blocking finding)

- [x] **Real and still present** — `wake()` `:277` has no in-flight-dial guard; `redialNow()` leaves `this.open === false` (`:299`) until the new socket's `open` (`:336`), so `visibilitychange`→visible followed by `focus` runs `redialNow()` twice, aborting the first handshake.
- [x] **Harm bounded to a redundant connection attempt** — `this.socket` nulled before `dying.close()` (`:298`) so the abandoned socket's `close` bails on identity (`:355`); its buffered frames are blocked at `:350` from re-arming `lastInboundAt`; `failPending` finds an empty map and `wasOpen === false` suppresses a spurious down-event (`:307-308`); `this.queued` survives and flushes on the eventual `open` (`:342`). Server cost is one aborted upgrade — a session check (`ws-server.ts:72`) and possibly one `attach`/`detach` pair — per wake, not per focus.
- [x] **Correctly carried as non-blocking.** Broader than the note on #175 states: any wake arriving during an ordinary backoff reconnect aborts that dial too, for the same reason. Still self-healing and still bounded.

### Issue 139 and the PR body

- [x] **#182 does not close #139** — `closingIssuesReferences` is empty (GraphQL); #139 and #172 carried as `Refs` in the body and in both commit messages; #139 is still `open`.
- [x] **No claim that CI proves the overnight fix** — the body states the opposite: confirming it "needs an actual overnight run against a real NAT or reverse proxy, which no CI check can stand in for". The required-checks paragraph is a ruleset inventory, not evidence.
- [x] **The body's technical claims match the code** — server sweep, `lastInboundAt` sources, `terminate()` over `close()`, absent `activeWs` guard, the three wake signals, immediate dial with the counter reset, and the 15-minute window all check out line for line.

### Train hygiene

- [x] Base is `beta/0.2.0`, head `fix/panel-overnight-detach` at `0280087`, one squashed commit, `mergeable_state: clean`.
- [x] Diff is 4 files / +605 −22, all inside the panel-link layer. `shared/panel-link.ts` absent from the diff, so `PANEL_LINK_PROTOCOL_VERSION` cannot have moved; `router.ts` untouched, so the replay contract and the 2048-event buffer are unchanged; nothing in `packages/core`; nothing from #140.
- [x] Existing suites are added to, not rewritten — the upgrade-gate cases (`ws-server.test.ts:55-95`) and the close-driven recovery cases (`panel-link-client.test.ts:178-274`) are intact.
- [x] All required **Trains — beta/\*** checks green on this head, plus both installer E2E jobs and the Panel service seam E2E.

### Open, and deliberately so

- [ ] **#139 stays open pending an overnight run against a real NAT or reverse proxy.** Record that morning whether a tab left alone comes back on its own, and whether the discriminator #139 names (a Core painted unreachable ⇒ core link, not this) ever appears. Note while planning it: the reap at `ws-server.ts:194-206` is silent — the core link logs `core-link.connection.stale` (`pty-core-link-server.ts:475`) and this sweep has no counterpart, and there is no logger in this layer to add one to within this ticket's scope. Read the result off the client's behaviour, not off a log line.

**Verdict: approve.** Every criterion #172 set is met at the enforcing lines; the one known finding is real, still present, and does no harm beyond a redundant connection attempt.

PR #183, opened 2026-08-11
feat(panel): rename a registered Core (#19)

qcentic-dev:
Integration PR for the `beta/0.2.0` train. A registered Core can be renamed in place. One ticket PR, already squash-merged onto this branch.

## The ticket PR this carries

- **#171 — rename a registered Core from Settings → Cores.** The alias was fixed at pairing time: `labelFor()` derived it once from the registration blob and nothing could change it afterwards, so the only way to rename a Core was to remove and re-pair it — destructive, for a cosmetic field.

  **Service and route.** `renameCore()` puts the operator's label through `labelFor()`, the same normalization registration uses, and bumps `updated_at`. Trimmed, capped at 120, and an empty one falls back to the endpoint host, so no gesture can leave a row with nothing to identify it by. `PATCH /api/cores/:id` takes the label and only the label — endpoint and credentials are what the registration blob said they were, and re-pairing is still the only way to move them.

  **Panel-local by design.** The Core alias is Panel-owned presentation, not a Core fact: no core-link frame carries it and the machine is never told. Two Panels registered against one Core may call it different things, which is the point of the field rather than drift.

  **The gesture.** The pencil turns the label into a text box; Enter or Save writes, Escape or Cancel backs out. A refused write keeps the box open holding what was typed, rather than costing the operator text they would have to retype. The row shows what the service stored rather than what was typed, so clearing the box visibly yields the endpoint host — and the placeholder says which host that will be before they commit to it.

  **Both alias consumers are pinned rather than reasoned about**, since an alias that could be edited but went on displaying its old name everywhere else would close the gap in name only. `useSessionFinishNotifications` keys its subscription on which Cores exist, so a rename deliberately does not re-run it — the alias map is read through a ref refreshed on render, and a finish arriving after a rename is now proven to carry the new name. `NewHarnessDialog` resolves the alias per render, so "Install claude on …" names the machine by the name the operator uses for it now.

Closes #19

---

The required checks run under the **Trains — beta/\*** ruleset: Conventions, Train rules, Typecheck, Unit Tests, Lint, Secret Scan, Dependency Audit, Panel image / Resolve registries, Panel image / Build + smoke (amd64), Core image / Resolve registries, and Core image / Build + smoke (amd64). The merge is **squash-only**, with **one approving review** required.

PR #184, opened 2026-08-11
feat(core-link): a Core serves many clients, one holds a Session's write lock (#140)

### LINKED ISSUE #141: Core accepts many concurrent core-link connections (D1, D12)

### LINKED ISSUE #142: PTY output fans out per connection, by subscription (D2)

### LINKED ISSUE #143: Announce multiConnection on ready; the protocol version does not move (D11)

### LINKED ISSUE #144: Session write lock: claim, release, force takeover, mutation gating (D3-D7, D10)

### LINKED ISSUE #145: Publish lock state on the Session snapshot and as an event (D8)

### LINKED ISSUE #146: Stable client id: reclaim a reconnecting client's locks, reaping only (D9)

### LINKED ISSUE #147: Panel: read-only Sessions, claim affordance, tab arbitration (D3, D8)

qcentic-dev:
> **This supersedes #181**, which cannot merge. `Conventions` lints every commit in a pull
> request's range, and the range on `feat/multi-connection-core` contains the squash commit of
> #180 — a 121-character subject, one character over the repository's 120-character
> `header-max-length`. That commit is already merged and its history is shared, so the only fixes
> available there are a rewrite of shared history or a force push, and neither is acceptable. This
> branch instead carries the *same tree*, squashed into one compliant commit off `beta/0.2.0`:
> `git diff` against `feat/multi-connection-core` is empty. That integration branch is untouched
> and keeps the per-ticket history.

Integration PR for the `beta/0.2.0` train. A Core stops evicting: it serves many core-link connections at once, and one of them holds a Session's write lock. Seven ticket PRs, already squash-merged onto this branch, land ADR 0024 end to end.

## The ticket PRs this carries

- **#170 — a Core accepts many concurrent core-link connections (D1, D12).** The connection registry replaces evict-on-connect, so a new socket displaces nobody. Each connection carries its own event cursor, subscriptions and heartbeat, and none is privileged — the Panel is a Core client like the CLI. Locks never outlive the process (D12), which costs nothing yet because there are none to persist.
- **#174 — PTY output fans out per connection, by subscription (D2).** The interim fan-out that gave every registered connection every PTY's bytes is gone. A connection is served a PTY only after a `ptySubscribe` for it, and catch-up is ordered *behind* the subscription: the Core holds live output, serves `replay` from the client's own `sinceSeq`, then drains what it held, so live bytes never land in front of the scrollback they follow. A PTY that exits inside that window has its exit held and drained too. A connection is subscribed to what it spawned.
- **#173 — announce `multiConnection` on `ready`; the protocol version does not move (D11).** One optional capability field instead of a minor bump. Absent means single-connection eviction, which is precisely today's behaviour — the narrow test the amended version-lock rule now permits.
- **#176 — one Core client holds a Session's write lock (D3–D7, D10).** The `claim` / `release` / `forceTakeover` frames, the Core-wide lock table keyed by Session, and the gate on `write`, `kill`, every `tasksMutate` addressed at a Session, `harnessPrompt` and `spawn` against an existing Session. `resize` is deliberately not gated — a Reader has a real use for one. A refusal comes back as an `error` carrying `session-locked`, distinct from the ordinary `ok: false` a vanished Session has always returned, because one is worth retrying after a claim and the other never is. No idle timeout (D7): a long agent run is idle by definition.
- **#178 — lock state is published, not discovered by failing (D8).** Every Session snapshot carries a `lock` stamped for the connection it is sent to — `{ writable, state }`, where `state` is `unlocked`, `held-by-you` or `held-by-another` — so a watcher learns that it may not write and nothing about who holds it. Every change appends a dedicated `session:lockChanged` event that replays by cursor. Nothing changed publishes nothing: an idempotent re-claim, a denied claim, a release from a connection holding nothing, and a takeover by the current holder all answer their frame and append no event.
- **#179 — a stable client id reclaims a reconnecting client's Session locks (D9).** A connection presents its id on a `reclaim`; the Core closes whatever connection last presented that string and moves its Session locks across **in place, before the predecessor's socket is touched** — the other order leaves those Sessions momentarily unlocked, and an unlocked Session is writable by anybody. The id is never signed, never verified and never authentication; it moves only authority any authenticated connection can already take with a `forceTakeover`. Minted per Core client, never derived from the registration blob, which every client on the machine shares.
- **#180 — a Session another Core client holds renders read-only, and the Panel arbitrates its own tabs (D3, D8).** The Panel's half, and where all of it is finally seen. Read-only is a *state of the terminal* rather than a second component: the same xterm keeps painting every byte while `disableStdin` takes its keyboard and cursor, and every write path consults one ref — the file drop, the key map's escape sequences, the question overlay's injected answers, and `onData`'s own status patch and prompt capture. Two names for two things: the **Session lock** is Core-scoped and crosses the wire, the **Session drive** is Panel-scoped and crosses nothing, because the Panel holds one lock for all its browser tabs. Their losers are told different sentences — a handover took nothing, a force takeover is unrecoverable by design. The Panel derives its lock state rather than refetching it, suppressing exactly one echo per gesture it performed, since its *own* takeover's `taken-over` row arrives after the answer and read naively would tell the tab that just took a Session that it had lost it.

## Docs riding along

- **ADR 0024**, `docs/adr/0024-a-core-serves-many-clients-one-holds-a-sessions-write-lock.md` — the record itself, with the rejected alternatives (creator holds automatically, first mutation claims implicitly, idle timeout, moving the protocol version), the known risks, and a landing note that now runs to the end.
- **`CONTEXT.md`** — seven domain terms added (**Core client**, **Core client id**, **Core alias**, **Session lock**, **Reader**, **Session drive**, **PTY subscription**), each with its banned synonyms; `Core link` restated in terms of a Core client; the `One Core link per Core` rule restated as **per Core client**; a new **many Readers, one writer, per Session** rule; and the version-lock rule narrowed rather than deleted — an additive surface may announce itself on `ready` only where its absence yields today's behaviour exactly.
- **`CONTRIBUTING.md`** — the same three rule changes in the load-bearing list.
- **`docs/README.md`** — ADR 0024 in the index.

Closes #141
Closes #142
Closes #143
Closes #144
Closes #145
Closes #146
Closes #147
Refs #140

---


qcentic-adm:
## Acceptance checklist — release-train review

Reviewed as a release train into `beta/0.2.0`: the tree identity first, then the full diff against the base on its merits. Every claim below was checked by reading the enforcing lines, not the PR body.

### Gate — the tree must be the integration branch's tree

| # | Check | Result |
|---|---|---|
| 0.1 | PR head is one commit | `ecc129f`, sole commit in `beta/0.2.0..head` |
| 0.2 | Parented directly on the base tip | parent `f465447` == `origin/beta/0.2.0` |
| 0.3 | `origin/feat/multi-connection-core` is at `2a19341` | matches |
| 0.4 | **`git diff <head> 2a19341` is empty** | **empty — 0 bytes** |
| 0.5 | Tree SHAs identical | both `310b0df4da03ab0a20722b8000d157a55990aafa` |
| 0.6 | Integration branch untouched, per-ticket history kept | 7 commits, one per ticket PR (#170, #174, #173, #176, #178, #179, #180); #180's squash merge *is* `2a19341` |
| 0.7 | The `header-max-length` failure that blocked #181 is gone | subject is 84 chars (limit 120); no body line over 100; `Conventions` and `Train rules` both pass |

The stated premise of this PR holds exactly: same tree, one compliant commit, integration branch left alone.

### The four proof obligations

| # | Obligation | Enforcing line | Result |
|---|---|---|---|
| 1 | An **unlocked** Session is writable by any connection | `packages/core/src/session-lock-table.ts:104-108` — `mayMutate` returns true on `current === undefined`; `pty-core-link-server.ts:947` refuses only when `mayMutate` is false | Holds. This is D11's compatibility promise: a client that never claims is served exactly as today. |
| 2 | With a lock held, writes from a **non-holder** are rejected | `pty-core-link-server.ts:942-955` (`refuseLockedSession`) → `error` + `SESSION_LOCKED_ERROR_CODE`, applied at `write:1092`, `kill:1114`, `spawn:1060`, `tasksMutate:1326-1331` (non-`create`), `harnessPrompt:1363`; `ptyId` resolved to its Session at `968-970` | Holds, and the refusal is distinguishable from "that Session is gone" (`ok:false` / `task:null`), which is the distinction a caller has to act on. |
| 3 | Panel client gates multi-connection-only frame types on the capability | `packages/panel/src/server/core-link/client.ts:66-74` (the set) and `:747-753` — refused inside `request()` before a `reqId` is allocated and before anything reaches the socket | Holds. Capability recorded from `ready` (`:639`) and cleared on both `open` (`:433`) and `close` (`:482`), so an unknown capability never reads as `true`. |
| 4 | Reconnect re-establishes PTY subscriptions | `client.ts:1060-1069` (`resendPtySubscriptions`) driven from `authOk` (`:685`) and from `ready` when no bearer (`:648`) | Holds. Ordering is load-bearing and correct: `ready` → capability → `reclaim` → `subscribe` → PTY resend. The want-set is kept even when no frame goes out (`:1010-1016`), so a Core that gains the capability across a restart is re-subscribed. |

### The rest of the surface

| Area | Result |
|---|---|
| Connection registry (D1, D12) | Per-connection cursor, subscriptions, heartbeat; `dropConnection` retires only its own timers and entry, guarded on identity so a replaced socket cannot evict its successor. Locks swept on drop and each publishes a `released` event — the only way a waiter learns a vanished holder's Session is claimable. |
| PTY subscriptions, subscribe→hold→replay→drain (D2) | `fanOutPtyEvent` serves only subscribers (and only authenticated ones); a holding subscription buffers instead. `drainPtyHold` runs *after* the replay window goes out and filters by exclusive `nextSeq`, so nothing is painted twice and live bytes never precede their scrollback. A held `exit` always drains. Hold bounded at 100 KB dropping oldest, which is lossless because the replay is read from the (larger) ring afterwards — reasoning stated at the constant and pinned by a test. |
| `multiConnection` on `ready` (D11) | One optional field; version does not move. A Core that does not announce it stamps no `lock` on snapshots and appends no lock events — it never answers two ways about a table it says it does not have. |
| Lock state published per-asker (D8) | `stateFor(taskId, asker)` collapses every other holder to `held-by-another`, so no watcher learns an identity; `writable` is `mayMutate` for the same asker from the same table, so a client cannot be told it may write and then refused. Rows are copied, not patched, so a per-connection field cannot leak through a cached row. Stamped on `tasksList`, `archivedTasksList`, `sessionsList` and `tasksMutateResult` — including the `create` that comes back `unlocked`, which is D5 on the wire. Nothing-changed-publishes-nothing is honoured at all three frames (`claim` `granted && !alreadyHeld`, `release` only when it released, `forceTakeover` skipped for `this-connection`). |
| Stable client id, atomic reclaim (D9) | `transferAll` rewrites holders in place, and `reclaimFor` transfers *before* touching the predecessor's socket — so there is no instant where those Sessions read unlocked and a third connection's `claim` (or an outright write, which D5/D11 permit) cannot land inside the handover. The subsequent `releaseAll` on the drop correctly finds nothing. Empty/non-string `clientId` is rejected rather than matched, which closes the one way this frame could misfire. Id minted per client via `randomUUID`, never derived from the shared registration blob. |
| Panel read-only (D3, D8) | Every write path consults one ref — `writeToPty` (`TerminalPane.tsx:991-995`), both `controls.writeToPty` (`:930`, `:1200`) and `onData` (`:1215`, returning *before* the status patch and prompt capture, since a pane that may not type may not rename either). `disableStdin` makes it visible; the ref covers the paths that never touch xterm's keyboard. Re-applied on reattach (`:845`), so a surface parked while the lock moved is corrected. |
| Session lock vs Session drive | Kept genuinely separate: Core-scoped and on the wire vs Panel-scoped and crossing nothing, with different copy for their different losers. `crossClientLockNotice` suppresses the "never had a settled answer" case, so opening a pane on a long-held Session does not report a takeover that never happened; `selfEchoes` is a queue that only ever holds entries the Core will actually emit an event for, and is deliberately not drained by snapshots. |
| Docs | ADR 0024 with its rejected alternatives and the rule it amends; CONTEXT.md gains seven terms with banned synonyms and three rule changes; CONTRIBUTING.md mirrored; docs/README indexed. The version-lock rule is narrowed rather than deleted, and the narrowing is exactly what D11 relies on. |
| Tests | ~5,300 lines across 13 suites, named as behaviours. The four obligations are each pinned directly, including the negative cases — an unlocked Session served, a refused mutation never acquiring the lock (D6), a takeover of an unheld Session published as `claimed`, the hold bound with the replay covering what it dropped. |
| Hygiene | No `TODO`/`FIXME`, `console.log`, `debugger`, `.only` or skipped tests in the added lines. All 19 required checks pass; the three non-success entries are `skipping` train jobs, as expected on a PR. |

### Non-blocking observations

- **`resize` is deliberately ungated** (`pty-core-link-server.ts:1097-1108`). A Reader resizing a PTY does affect the holder's terminal, so this is a real if minor interference channel. It is the documented reading of D4 with a stated rationale — a Reader painting into its own viewport has a genuine need — and widening D4 would be an amendment to #140, not a call for this train. Recorded, not asked for.
- **`taskIdForPty` returns a taskId for shell and VM Shell Session PTYs too**, so those are gated like any other Session. Intended and documented at `pty-manager.ts:770-778` — a shell's Session is its own, which is the id its own claim would name.
- **The Core client id is per process, not persisted.** A Panel service restart mints a new one, so its predecessor's locks are reaped by the heartbeat rather than the reclaim. That is the slower path to the same state, and the socket dies with the process anyway — worth knowing, nothing to change.

### Verdict

**Approved.** The gate passes exactly — the head tree is byte-identical to `feat/multi-connection-core` at `2a19341` — and the diff stands up on its own against `beta/0.2.0`: the invariants are enforced where it claims they are, the compatibility promise is real rather than asserted, and the ordering decisions that the design turns on (drain behind replay, transfer before close, capability before resend) are each implemented in the order that makes them safe.

qcentic-adm (dismissed):
Approved.

**The gate passes.** `git diff` between this PR's head (`ecc129f`) and `origin/feat/multi-connection-core` at `2a19341` is empty — 0 bytes, identical tree `310b0df`. One commit, parented directly on the `beta/0.2.0` tip, subject 84 chars, so the 121-character `header-max-length` failure that made #181 unmergeable is genuinely gone rather than worked around. The integration branch is untouched and still carries the seven ticket PRs commit by commit.

I then reviewed the diff against `beta/0.2.0` on its own merits rather than trusting the per-ticket approvals, and proved the four invariants by reading the enforcing lines:

1. **An unlocked Session is writable by any connection** — `session-lock-table.ts:104-108`, `mayMutate` true on `current === undefined`, and `pty-core-link-server.ts:947` refuses only on false. This is what makes D11's "an old client works untouched" true rather than merely claimed.
2. **A non-holder's writes are rejected** — `refuseLockedSession` (`:942-955`) applied at `write`, `kill`, `spawn`, non-`create` `tasksMutate` and `harnessPrompt`, with `ptyId` resolved to its Session first. The `session-locked` code keeps "refused" distinguishable from "gone", which is the distinction a caller acts on.
3. **The panel client gates multi-connection-only frames on the capability** — `server/core-link/client.ts:66-74` and `:747-753`, refused before a `reqId` exists and before the socket is touched. The capability is cleared on both `open` and `close`, so "unknown" never reads as "supported".
4. **Reconnect re-establishes PTY subscriptions** — `resendPtySubscriptions` (`:1060`) off `authOk` (`:685`) and off `ready` when no bearer (`:648`), and the want-set survives a capability-less Core so a Core that gains the capability across a restart is re-subscribed.

The ordering decisions the design actually turns on are each implemented in the order that makes them safe: the hold drains *behind* the replay window and filters by exclusive `nextSeq`; `transferAll` rewrites holders in place *before* the predecessor's socket is closed, so no third connection can claim or write inside the handover; the capability is recorded before any resend reads it. Lock state is published per-asker and collapses other holders to `held-by-another`, so no watcher learns an identity, and nothing-changed-publishes-nothing is honoured at all three lock frames. On the Panel side every write path — file drop, key map, question overlay, `onData` — consults one ref, and `onData` returns before its status patch and prompt capture, which is the right call: a pane that may not type may not rename either.

Three non-blocking notes are in the checklist comment. The main one: `resize` is deliberately ungated, so a Reader's resize does reach the holder's terminal. It is the documented reading of D4 with a stated rationale, and widening it would amend #140 rather than belong to this train — recorded, not requested.

All 19 required checks pass; the three non-success entries are `skipping` train jobs. Good to merge.

mrfqcentic (dismissed):
## Review: APPROVE — requirements verified issue by issue; one moderate follow-up to track

Verified against the PR head (`ecc129f`, tree identical to the superseded #181). All CI green, including Conventions — the 121-char squash subject that blocked #181 is gone by construction (single 84-char commit).

### Requirements check — the #140 train

| Issue | Requirement | Verdict |
|---|---|---|
| #141 (D1, D12) | Connection registry replaces evict-on-connect; per-connection cursor/heartbeat/subscriptions | ✅ Met — `pty-core-link-server.ts:460`; heartbeat reaps only its own connection; emit target released on last close |
| #142 (D2) | Per-connection PTY fan-out, subscribe→hold→replay→drain catch-up, exit-in-window held | ✅ Met — ordering pinned by frame-order tests; `drainPtyHold`; spawner auto-subscribed |
| #143 (D11) | `multiConnection` capability on `ready`, protocol version unmoved | ✅ Met — `0.15.0` pinned by a named test; absent capability = exactly today's behaviour |
| #144 (D3–D7, D10) | claim/release/forceTakeover; gate set; distinct `session-locked` refusal; drop releases all | ✅ Met — gates at spawn/write/kill/tasksMutate/harnessPrompt; `resize` deliberately open per ADR 0024; no idle timeout |
| #145 (D8) | Lock state stamped per-recipient on snapshots; `session:lockChanged` replays by cursor | ✅ Met — one scope note: the `actana session ls` surface doesn't exist yet; the data half (`sessionsList` carries the lock) is delivered |
| #146 (D9) | Reclaim moves locks **before** closing the predecessor socket; id is not authentication | ✅ Met — order pinned by a mutation-verified test (`:684-701`) |
| #147 (D3, D8) | Read-only rendering via `disableStdin`; Panel-scoped Session drive for tab arbitration | ✅ Met — all write paths consult `mayWriteRef`; self-echo suppression on own takeover |

ADR 0024 matches the code on every claim checked. All seven ticket PRs (#170, #173, #174, #176, #178, #179, #180) were individually approved after their review findings were fixed. Nothing was lost in the #181 → #184 move.

### The one finding that must stay attached to this PR (moderate — fast follow-up, not a blocker)

**Session drive: pane/tab granularity mismatch.** The drive register keys interest per *tab* (`PanelLinkSession`), but the browser emits `watch`/`drop` per *pane* with no refcounting (`session-drive-register.ts:65-73`, `panel-link-client.ts:278-294`, `router.ts:353-370`). With two panes on one Session in a tab plus a second tab watching: (a) a duplicate `watch` re-appends the driving tab to the back of the queue — the keyboard silently moves to the other tab with no toast; (b) closing one of the two panes sends `drop`, removing the tab's interest while its other pane stays writable → silent dual-write, the exact failure #147 exists to prevent. Panel-local, recoverable, never crosses the Core lock, and no worse than the pre-#147 baseline when it misfires — but it should be a tracked fast follow-up on this train (dedupe the `watch` send client-side and/or position-preserving `want`, plus a release refcount). Currently untested.

### Carried minor items (all acknowledged non-blocking on the ticket PRs)

- `claimPty` early-returns when no `CoreState` exists yet (`router.ts:250-251`) — a claim before the Core's first `bind` is silently dropped (deferred from #174).
- The four non-keyboard write gates (file drop, key-map escapes, overlay answers, `onData` capture) are code-read but untested; no test that a mid-flight snapshot doesn't clear echo suppression (deferred from #180).

### Merge mechanics

Squash-only, one approval — this is the approval. Suggested train order: **#182 → #184 → #183**. This PR overlaps #182 trivially in `packages/panel/src/lib/panel-link-client.ts` (+ its test; adjacent hunks in the `open` handler) and composes semantically — wake-redials reuse `connect()`→`open`, so the re-announces added here run on wake redials too.

*AI-assisted review; all findings verified against the PR head tree, the linked issues, and the ticket-PR review threads.*

qcentic-dev:
## Base merged in — `beta/0.2.0` (`05c69eb`) → `feat/multi-connection-core-train`

#182 squash-merged onto the train while this PR was open, so this branch was behind its base and
overlapped #182 in `packages/panel/src/lib/panel-link-client.ts` and its test. Brought the base in
with **`git merge origin/beta/0.2.0`**. Merge commit **`1e069e3`**, pushed as a fast-forward — no
rebase, no force push, the published history is untouched.

### The conflicts, and how each was resolved

**1. `packages/panel/src/lib/panel-link-client.ts` — the `open` handler.**

#182 stamps `lastInboundAt` there, and that stamp is the whole basis of the wake path's staleness
check. #184 rewrote the comment above it and added `resendPtySubscriptions()` /
`resendSessionDrives()` below. Git kept both sides' calls on its own and conflicted only on the
stamp and the comment.

Resolved by keeping the stamp and rewriting the comment to cover all three re-announces. Both sides
survive in full:

```ts
this.open = true;
this.reconnectAttempt = 0;
this.lastInboundAt = Date.now();          // #182
for (const coreId of this.watching.keys()) this.sendSubscribe(coreId);
this.resendPtySubscriptions();            // #184
this.resendSessionDrives();               // #184
```

The two changes **compose** here rather than merely coexisting, and that is the part worth being
explicit about. A wake redial is not a separate path: `wake()` drops the dead socket and calls
`connect()`, so it lands in this same `open` handler. #184's re-announces therefore run on wake
redials for free — a tab back from hours hidden re-subscribes its Cores, re-asks for its PTYs and
re-announces its Session drives, all of it off the backoff. Neither re-announce may be lifted above
this handler for that reason, and the comment now says so in place.

Everything else in this file merged clean and is present on both counts: #182's `staleAfterMs` /
`DEFAULT_STALE_AFTER_MS`, `listenForWake()` (`visibilitychange` / `focus` / `online`), `wake()`,
`redialNow()` with its `reconnectAttempt = 0`, `failPending()`, the abandoned-socket guard on the
message handler, and the `lastInboundAt` stamp in `onMessage`; alongside #184's `claimedPtys`,
`drivenSessions`, `ptySubscribe` / `ptyUnsubscribe`, `driveSession` / `releaseSessionDrive`, the
`lock` and `drive` frame branches, and both `resend*` methods. Confirmed by diffing the merge
against each parent: the only deletions against #184's tip are the lines #182 legitimately replaced
(the two inline pending-failure loops, now `failPending()`, and the unguarded message listener), and
the only deletion against #182's tip is the comment reworded above.

**2. `packages/panel/src/lib/__tests__/panel-link-client.test.ts` — two new suites at the same spot.**

Both sides appended a new `describe` block at the same point in the file and git interleaved them,
cutting through the middle of a test on each side. Resolved by keeping **both suites whole and in
sequence** — #184's `panel link · the ptys a tab is rendering`, then #182's `panel link · a socket
that dies without saying so` — rather than by trying to splice the interleaving. Reconstructed from
the three merge stages so neither suite could be silently clipped.

Three further edits in that file are independent and all three are in: #182's `PanelLinkOptions`
type import (its `staleClient` helper needs it), and #184's two `FakeSocket` changes — the
`t === "core"` narrowing in `outgoing()` and the new `drives()` accessor.

### One test added

The seam the merge creates was covered by neither PR alone, so it now has a test in #182's suite:

> `re-announces the tab's ptys and session drives on a wake redial`

A socket dies in silence — no `close`, no error, `readyState` still `OPEN` — and is revived by a
visibility signal. The test asserts the new socket is told about the tab's PTY claims and Session
drives, not merely its `subscribe`, with `catchUp: false` and `want: "watch"` for the same reasons a
close-driven reconnect uses them. It fails against either parent commit.

### Local test result

`vitest run` over the `panel` package, on the merge commit:

```
Test Files  131 passed (131)
     Tests  1399 passed (1399)
```

`pnpm typecheck` (`tsc --noEmit`) clean.

Recorded for honesty: the first full run showed timeouts in unrelated integration and component
tests (`terminals-in-browser`, `live-read-path`, `operator-auth`, two `components/views` suites).
All were the 5s default timeout on a box sitting at load average 16 across 10 cores, and the set
varied between runs. Every one of them passes in isolation, and the full suite is green as above
with bounded workers. Nothing in that set was touched by this merge.

<sub>Base merge, tests and this comment done by Claude Opus 5 on behalf of @qcentic-dev.</sub>

qcentic-dev:
### CI on the merge commit (`1e069e3`) — green

**19 pass, 3 skipping, 0 fail.** Unit Tests, Typecheck, Lint, Train rules, both image builds and
both E2E legs all pass on the merged tree, which is the independent confirmation of the local run
above.

One footnote for the trace, since the first attempt was red. **Conventions** failed on its *first*
run and the cause was not this branch:

```
title="$(gh api "repos/$REPO/pulls/$PR_NUMBER" --jq .title)"
invalid character '<' looking for beginning of value
```

The `gh api` call in the "Lint the PR title" step got an HTML error page back instead of JSON, so
the step exited 1 before commitlint ever ran. Confirmed by linting locally with this repo's own
`commitlint.config.mjs` over the range the job uses — the merge commit, `ecc129f` and `05c69eb` all
return `found 0 problems`, and the branch name matches the convention pattern. Re-running the job
alone turned it green in 20s, nothing else changed.

Worth stating plainly given #181's history: merging the base in did **not** reintroduce a commit
that fails `header-max-length`. `05c69eb` is a 96-character subject and now sits in this PR's range
without complaint.

The PR is `MERGEABLE` and still open — `BLOCKED` only on the approving review the train ruleset
requires. Not merged, per instruction.

qcentic-adm (approved):
Re-approving after the base merge. This is a merge verification, not a repeat of the full review of #184 — the code review at `ecc129f` stands, and what follows is the check that the merge preserved it.

**No force push.** `1e069e3` is a true merge commit with parents `ecc129f` (the approved tip) and `05c69eb` (the new base). `git merge-base --is-ancestor ecc129f 1e069e3` passes, so the reviewed history is intact and nothing published was rewritten. The range `ecc129f..1e069e3` contains exactly two commits: #182's squash and the merge itself.

**Nothing else changed.** Replaying the merge mechanically (`git merge-tree ecc129f 05c69eb`) conflicts in exactly the two files described — `panel-link-client.ts` and its test — and every other path in the resulting tree is **byte-identical** to `1e069e3`. So the only hand-written content in this merge is the two resolutions plus the added test; nothing was smuggled in under cover of the merge. The two `ws-server` files came across with a delta identical to #182's own, hash for hash.

**Both sides survive in `panel-link-client.ts`.** Against `ecc129f` the only additions are #182's liveness work — `staleAfterMs` / `DEFAULT_STALE_AFTER_MS`, `lastInboundAt`, `listenForWake()` over `visibilitychange` / `focus` / `online`, `wake()`, `redialNow()` with its `reconnectAttempt = 0`, `failPending()`, the abandoned-socket guard on the message handler and the `lastInboundAt` stamp in `onMessage`. The only deletions are the two inline pending-failure loops #182 replaced with `failPending()` and the unguarded message listener. Against `05c69eb` the sole deletion is the two-line comment reworded in the `open` handler. No code was dropped in either direction.

**The composition is real, not adjacency.** `wake()` → `redialNow()` clears the socket reference, closes the corpse and calls `connect()`; `connect()` assigns the new socket and its `open` handler — passing the `this.socket !== socket` identity guard — runs the subscribe loop, `resendPtySubscriptions()` and `resendSessionDrives()`. A wake redial therefore re-announces through the same path a close-driven reconnect uses. The resolved comment says so in place, which is the right guard against someone later lifting either re-announce above the handler and silently breaking the wake path.

**The test pins it.** `re-announces the tab's ptys and session drives on a wake redial` is the one test present in the merge and in neither parent; no test from either side was lost (30 in the merge against a 29-test union of the parents). It kills a socket in silence, revives it with a visibility signal, and asserts the *new* socket's frames with `toEqual` on the full array — `subscribe`, then `ptySubscribe` with `catchUp: false`, plus `drive` with `want: "watch"`. Because it asserts the whole array rather than a subset, an accidental extra frame or a dropped re-announce both fail it. It cannot pass on either parent: `ecc129f` has no wake path at all, and `05c69eb` has no `ptySubscribe` / `driveSession` to call. Both suites are present whole and in sequence.

CI on `1e069e3` is green — 19 pass, 3 skipped, 0 fail — which independently confirms the local run. The `Conventions` first-run failure was a `gh api` call receiving HTML instead of JSON in the title-lint step, an infrastructure flake that never reached commitlint; it is green on re-run.

Merge is clean. Approving at `1e069e3`.

qcentic-adm:
### Merge re-review checklist — `ecc129f` → `1e069e3`

| # | Check | Method | Result |
|---|---|---|---|
| 1 | No force push | `git merge-base --is-ancestor ecc129f 1e069e3` | ✅ passes; `1e069e3` parents are `ecc129f` + `05c69eb` |
| 2 | Range is only the base merge | `git log ecc129f..1e069e3` | ✅ two commits: `05c69eb` (#182 squash) + the merge |
| 3 | Base tip matches | `origin/beta/0.2.0` == PR base oid | ✅ `05c69eb`, #182 `MERGED` at 19:01:35Z |
| 4 | Nothing else changed | `merge-tree ecc129f 05c69eb` vs `1e069e3` | ✅ conflicts in exactly the 2 named files; **every other path byte-identical** |
| 5 | Files touched | `git diff --stat ecc129f 1e069e3` | ✅ same 4 files as #182, no others |
| 6 | `ws-server.ts` / its test carried across intact | delta hash vs #182's own delta | ✅ identical |
| 7 | #182 side survives | deletions in `05c69eb..1e069e3` | ✅ only the reworded `open` comment |
| 8 | #184 side survives | additions/deletions in `ecc129f..1e069e3` | ✅ only #182 liveness added; only the 2 replaced loops + unguarded listener removed |
| 9 | Wake hooks present | `listenForWake()` | ✅ `visibilitychange`, `focus`, `online`, all detached on `close()` |
| 10 | `reconnectAttempt` reset | `redialNow()`, `open` handler | ✅ both |
| 11 | Abandoned-socket guard | message handler | ✅ `if (this.socket !== socket) return;` |
| 12 | #184 re-announce + capability logic | `open` handler | ✅ subscribe loop, `resendPtySubscriptions()`, `resendSessionDrives()` |
| 13 | **Wake redial goes through the re-announcing path** | `wake()` → `redialNow()` → `connect()` → `open` | ✅ same handler; identity guard passes for the new socket |
| 14 | Added test pins the composition | full-array `toEqual` on the second socket | ✅ `subscribe` + `ptySubscribe{catchUp:false}` + `drive{want:"watch"}` |
| 15 | Test fails on both parents | capability check | ✅ `ecc129f` has no wake path; `05c69eb` has no `ptySubscribe`/`driveSession` |
| 16 | No test lost | test inventory, merge vs union of parents | ✅ 30 vs 29 — exactly one added, none dropped, both suites whole |
| 17 | CI on the merge commit | check-runs on `1e069e3` | ✅ 19 pass, 3 skipped, 0 fail |

**Verdict: approved at `1e069e3`.** The merge is clean — both PRs' behaviour composes on the wake path, and the added test holds that composition in place.

<sub>Merge verification by Claude Opus 5 on behalf of @qcentic-adm. Read-only: no repo file was edited and nothing was committed, pushed or merged.</sub>

Issue #187, opened 2026-08-12
Session read-only lags a reload and never lifts: the Panel does not see another client's lock live, and the Session stays running after the run returns


mrfqcentic:
### Description

Observed on the `beta/0.2.0` train (`actana/panel:beta-0.2.0` + `actana/core:beta-0.2.0`, images from
`520cd22`), driving one Core from two clients at once: the Panel in a browser, and a second
core-link client on the host — the `core-client` CLI in `experiment/`, which is not part of this
repository yet.

The multi-connection behaviour itself worked: both clients were attached to the same Core, and
sending to the session from the CLI did what it should. What did not work is how the Panel showed the
Session lock around that run. Two things, both on the Panel side of #178 / #180:

1. **The read-only state is not shown when it happens.** After the CLI took the Session, the open
   Panel tab went on rendering the terminal as writable. It only turned read-only after I reloaded
   the page. `session:lockChanged` is supposed to make exactly this unnecessary — the lock is
   published rather than discovered — so either the event is not reaching the open tab, or it is
   arriving and not being applied to the terminal that is already mounted.

2. **It never lifts.** The run finished and returned its final result to me on the CLI, and the
   Panel still showed the Session as read-only *and* still showed it as running. Neither the lock
   nor the running state came back on their own.

**Filed as a note, not for this train.** Deferring to **0.2.3 or 0.3.1** deliberately: the SDK and
the Core CLI are due to become part of the codebase, and the second client in this report is
precisely the thing that gets replaced by that work. This may well be fixed or made moot on the way,
so it is recorded rather than pushed now.

### Steps to reproduce

1. Bring up one Panel and one Core on the same network, with the Core also published on the host so a
   second core-link client can dial it (`ACTANA_PUBLIC_HOST=<service name>`, `ACTANA_PORT=9446`).
2. Pair the Panel, register a project, start a session, and leave the session's tab open in the
   browser.
3. From a second core-link client on the host, claim the Session and send it a prompt.
4. Watch the open Panel tab — do not reload it.
5. Let the run finish and return its result on the CLI, then look at the Panel tab again.

### Expected behavior

- The Panel tab turns read-only as soon as the other client holds the Session, without a reload.
- When the run is over and the holder is gone, the Session becomes writable again in the Panel and
  stops showing as running.

### Actual behavior

- The tab stays writable until the page is reloaded; only the reloaded page shows read-only.
- After the run returns, the Session stays read-only and stays marked running.

### Which part?

Panel — with the caveat below.

[...]

PR #188, opened 2026-08-12
chore(release): promote 0.2.0

qcentic-adm:
**Promotion gate for `beta/0.2.0`.**

This pull request is a **gate, not a merge**. Its approval and its green checks
are the entire point of it — they are what authorises the promotion. Nothing
here is intended to land by pressing a button in this page.

## ⛔ Do NOT press the merge button

`promote.yml` performs the fast-forward advance of `main` to this train's tip.
GitHub will then close this pull request as merged on its own, once its commits
are reachable from `main` — which the fast-forward makes true.

Merging it by hand (squash or otherwise) would produce a `main` commit whose SHA
differs from the tested one, which is exactly what the digest assertion cannot
survive.

## What to do instead

1. Let the checks go green.
2. Approve.
3. Dispatch the promotion:

   ```bash
   gh workflow run promote.yml --repo actana/control -f train=beta/0.2.0
   ```

See [`docs/ci-cd.md` → Promotion](../blob/main/docs/ci-cd.md#promotion) and
[ADR 0023](../blob/main/docs/adr/0023-release-trains-and-digest-promotion.md).

## Notes

- The freeze window is open: nothing merges anywhere until this promotes.
- `Panel image` / `Core image` do not rebuild on a promotion pull request — they
  perform the digest assertion instead (D19). Same check names, still green.


mrfqcentic (approved):
Reviewed as the promotion gate for `beta/0.2.0` — approving.

**Scope verified.** The range over `main` is exactly four commits: the cut (`f465447`, version bump only) plus the three integration PRs already reviewed individually — #182 (panel-link liveness), #183 (Core rename), #184 (multi-connection core-link, ADR 0024). The CONTEXT.md / CONTRIBUTING.md / docs index changes all arrive through those reviewed PRs; nothing unreviewed rides along.

**Head == what was tested.** The PR head is `520cd22`, the same commit the `beta-0.2.0` images under manual acceptance were built from (per #187). The `Panel image` / `Core image` checks ran in promotion verify mode — the digest assertion, not a rebuild (D19) — and pass, so the tested digest is the one being promoted.

**Versions agree.** All four manifests (root, `packages/core`, `packages/panel`, `packages/shared`) read `0.2.0`, matching the train name; asserted by a required check as well.

**Known issue carried into the release:** #187 — an open Panel tab does not see another client's Session lock until reload, and the lock/running state did not lift after the run. Filed and deliberately deferred to 0.2.3 / 0.3.1, with the caveat that the "never lifts" half may belong to the second (experimental) client rather than the Core. Not a blocker for this promotion; it ships as a documented known limitation.

Per the gate design: not merging here — promotion proceeds by dispatching `promote.yml` for `beta/0.2.0`, and the freeze window holds until it completes.

Issue #151, opened 2026-08-11
1.1 Node-22 mTLS and fetch spike — freeze the transport shape (D12)


mrfqcentic:
**Base branch: `refactor/sdk-extraction`, not `main`.** Train: **`beta/0.2.1`**.
Worktree: **`wt-129-01`**. Suggested branch: **`chore/node22-mtls-fetch-spike`**.
Phase 1, step 1. Resolves the risk behind **D12**.

## What

Prove, against a live Core, that Node 22 can speak mTLS both ways with credentials taken from a registration blob:

1. A WebSocket (`ws`) connection to the Core's core-link endpoint that reaches `ready`.
2. An HTTPS `fetch` to the Core's mTLS server that returns a response.

This is a spike. It freezes the transport shape before anything is extracted, and it either confirms D12's `engines: ">=22"` floor or kills it.

## Done when

- Both connections succeed on Node 22 with CA, client certificate and client key read from a blob.
- The working incantation for each is recorded in this ticket, verbatim.
- If either fails on 22 but passes on 24: say so here and stop. D12 is then wrong and the phase needs a decision before any extraction starts.

## Traps

- **`fetch` is undici, not `https.request`.** It does not take `options.cert` / `options.key`. A client certificate goes in through a custom `Dispatcher`/`Agent`. This single difference is the reason the spike exists — the Panel's existing client uses `https`-style options and proves nothing about `fetch`.
- The Core's certificate is issued for the exact host and port in the blob. Reaching it through a rewritten port mapping fails validation for reasons that have nothing to do with Node's version. See `experiment/action.md` §5.
- Do not land a permanent dependency or a published entry point from this ticket. Findings and a throwaway script only.

Refs #129

qcentic-dev:
## Verdict on D12 — it stands. Keep `engines: ">=22"`.

Both connections succeed on **Node 22.23.2** with CA, client certificate and client key read from a registration blob, against a **live Core** (`wss://127.0.0.1:9444`, protocol `0.15.0`, `coreId=core_17504d4b830baffc`). **Node 24.19.0** was run as the control and produced identical results.

Nothing fails on 22 that passes on 24, so there is **nothing to escalate** and the phase is not blocked.

| Leg | Node 22.23.2 | Node 24.19.0 |
| --- | --- | --- |
| 1 — `ws` reaches `ready` | PASS | PASS |
| 1b — bearer `auth` → `authOk` | PASS | PASS |
[...]

PR #189, opened 2026-08-12
chore(spike): Node 22 speaks mTLS both ways — D12 holds (#151)

qcentic-dev:
Spike for #151. Throwaway: `experiment/` only, nothing in `packages/`, no dependency added to the workspace, no published entry point.

## Verdict

**D12 stands — keep `engines: ">=22"`.** Both connections succeed on Node 22.23.2 with CA, client certificate and client key read from a registration blob, against a live Core. Node 24.19.0 was run as the control and produced identical results. Nothing fails on 22 that passes on 24, so there is no decision to escalate and extraction is not blocked.

| Leg | Node 22.23.2 | Node 24.19.0 |
| --- | --- | --- |
| 1 — `ws` reaches `ready` | PASS | PASS |
| 1b — bearer `auth` → `authOk` | PASS | PASS |
| 2a — `fetch` completes mutual TLS | PASS | PASS |
| 2b — `fetch` without a client cert is refused | PASS | PASS |
| 2c — `fetch` gets an HTTP response over mTLS | PASS | PASS |
| 3 — wrong host fails SAN validation | PASS | PASS |

## The trap held

`fetch` is undici and ignores `options.cert` / `options.key`. The certificate goes in through the dispatcher:

```js
const agent = new Agent({ connect: { ca, cert, key } });
await fetch(url, { dispatcher: agent });
```

Leg 2b is the control that makes 2a mean anything: a CA-only `Agent` is refused by the Core at the TLS layer.

## Two corrections worth carrying forward

- `experiment/action.md` does not exist in this repo or its history, so §5 could not be read. The claim was tested directly instead.
- X.509 binds **hosts, not ports**. A rewritten port validates fine (leg 2c proves it). A rewritten **host** is the real trap — `https://172.17.0.2:9444` against the same socket fails `ERR_TLS_CERT_ALTNAME_INVALID`.

## One thing a later ticket will need

The Core's mTLS port serves WebSocket upgrades and nothing else — `https.createServer` is built with no request listener. A plain `GET` there hangs. Leg 2c therefore gets its `HTTP 200` from a throwaway loopback listener using the Core's *own* server certificate and CA. The transport shape is proven; a real Core HTTP route does not exist yet and someone has to add it.

Full write-up in [`experiment/findings-151-node22-mtls.md`](experiment/findings-151-node22-mtls.md).

Refs #151

### DIFF
--- a/experiment/findings-151-node22-mtls.md
+++ b/experiment/findings-151-node22-mtls.md
@@ -11,8 +11,15 @@ Companion script: [`spike-151-node22-mtls.mjs`](spike-151-node22-mtls.mjs).
 
 Both connections succeed on Node 22 with the CA, client certificate and client
 key read from a registration blob, against a live Core. Node 24 was run as the
-control and produced byte-identical results. Nothing fails on 22 that passes on
-24, so there is no decision to escalate and extraction is not blocked.
+control: **the same six legs passed on both**, with the same frames, the same
+certificate subjects, the same TLS version and the same error codes. Nothing
+fails on 22 that passes on 24, so there is no decision to escalate and
+extraction is not blocked.
+
+The two runs are not byte-identical and cannot be — the script stamps
+`process.version` into its banner and summary, and leg 2c binds a fresh
+ephemeral port each run. Those three lines are the only differences; the
+transcripts are below in full so the claim is checkable rather than trusted.
 
 | Leg | Node 22.23.2 | Node 24.19.0 |
 | --- | --- | --- |
@@ -22,6 +29,61 @@ control and produced byte-identical results. Nothing fails on 22 that passes on
 | 2b — `fetch` without a client cert is refused | PASS | PASS |
 | 2c — `fetch` gets an HTTP response over mTLS | PASS | PASS |
 | 3 — wrong host fails SAN validation | PASS | PASS |
+
+## Transcripts, verbatim
+
+Both produced by `spike-151-node22-mtls.mjs` as committed on this branch, run
+back to back against the same live Core (`coreId=core_17504d4b830baffc`,
+protocol `0.15.0`), from the same registration blob. Exit status 0 in both.
+
+### Node 22.23.2
+
+```
+node v22.23.2  •  endpoint wss://127.0.0.1:9444  •  blob /home/core/.config/actana/registration-blob.txt
+
+PASS  1  ws → ready
+        ready frame: version=0.15.0 multiConnection=null
+PASS  1b  ws → authOk
+        coreId=core_17504d4b830baffc exp=1817638207292
+PASS  2a  fetch → mutual TLS
+        authorized=true authorizationError=null TLSv1.3
+        server=[CN=127.0.0.1] client=[CN=mission-control-panel]
+        HTTP layer: no response — Core serves no HTTP route on this port (see 2c)
+PASS  2b  fetch without client cert is refused
+        UND_ERR_SOCKET: other side closed
+PASS  2c  fetch → HTTP response over mTLS
+        HTTP 200 {"ok":true,"clientCn":"mission-control-panel","path":"/spike"} (loopback listener on the Core's own server cert, port 40021)
+PASS  3  wrong host fails SAN validation
+        dialled https://172.17.0.2:9444 → ERR_TLS_CERT_ALTNAME_INVALID
+
+6/6 legs passed on v22.23.2
+```
+
+### Node 24.19.0 (control)
+
+```
+node v24.19.0  •  endpoint wss://127.0.0.1:9444  •  blob /home/core/.config/actana/registration-blob.txt
+
+PASS  1  ws → ready
+        ready frame: version=0.15.0 multiConnection=null
+PASS  1b  ws → authOk
+        coreId=core_17504d4b830baffc exp=1817638207292
+PASS  2a  fetch → mutual TLS
+        authorized=true authorizationError=null TLSv1.3
+        server=[CN=127.0.0.1] client=[CN=mission-control-panel]
+        HTTP layer: no response — Core serves no HTTP route on this port (see 2c)
+PASS  2b  fetch without client cert is refused
+        UND_ERR_SOCKET: other side closed
+PASS  2c  fetch → HTTP response over mTLS
+        HTTP 200 {"ok":true,"clientCn":"mission-control-panel","path":"/spike"} (loopback listener on the Core's own server cert, port 43529)
+PASS  3  wrong host fails SAN validation
+        dialled https://172.17.0.2:9444 → ERR_TLS_CERT_ALTNAME_INVALID
+
+6/6 legs passed on v24.19.0
+```
+
+Line-for-line the two differ only in the banner's version, the summary's
+version, and leg 2c's ephemeral port (`40021` vs `43529`).
 
 ## The trap, confirmed
 
@@ -41,6 +103,14 @@ built as `new Agent({ connect: { ca } })` — CA only, no client cert — is ref
 by the Core at the TLS layer (`UND_ERR_SOCKET: other side closed`), because the
 Core sets `requestCert: true, rejectUnauthorized: true`. Without that control, a
 passing leg 2a would be indistinguishable from a server that never asked.
+
+The leg asserts on the *shape* of the refusal — `UND_ERR_SOCKET`/`ECONNRESET`/
+`EPIPE`, or an `ERR_SSL_*` alert — not merely on "it failed and wasn't slow".
+A control that passes on any error is not a control: `ECONNREFUSED` from a Core
+that is not running would otherwise read as PASS and make leg 2a look meaningful
+against nothing at all. Confirmed by pointing the blob at a dead port, where the
+leg now reports
+`ECONNREFUSED ... — not a TLS-layer refusal (is the Core up on this port?)`.
 
 ## Correction to the `experiment/action.md` §5 trap
 
@@ -116,4 +186,6 @@ SPIKE_MODULES=/tmp/spike-rig/node_modules node experiment/spike-151-node22-mtls.
 Needs a live Core and its registration blob (default
 `~/.config/actana/registration-blob.txt`). Leg 2c additionally reads
 `~/.config/actana/material.json` for the Core's server certificate. Exit status
-is 0 only when every leg passes.
+is 0 only when every leg passes — including the case where the `ws` socket dies
+or stalls between `ready` and `authOk`, which records leg 1b as a failure rather
+than dropping it from the tally.
--- a/experiment/findings-151-node22-mtls.md
+++ b/experiment/findings-151-node22-mtls.md
@@ -1,0 +1,119 @@
+# Spike #151 — Node 22 mTLS and `fetch` findings
+
+Throwaway. Nothing in `packages/` imports any of this, and no dependency was
+added to the workspace. Delete `experiment/` once D12 is settled.
+
+Companion script: [`spike-151-node22-mtls.mjs`](spike-151-node22-mtls.mjs).
+
+## Verdict on D12
+
+**D12 stands. `engines: ">=22"` is correct — keep it.**
+
+Both connections succeed on Node 22 with the CA, client certificate and client
+key read from a registration blob, against a live Core. Node 24 was run as the
+control and produced byte-identical results. Nothing fails on 22 that passes on
+24, so there is no decision to escalate and extraction is not blocked.
+
+| Leg | Node 22.23.2 | Node 24.19.0 |
+| --- | --- | --- |
+| 1 — `ws` reaches `ready` | PASS | PASS |
+| 1b — bearer `auth` → `authOk` | PASS | PASS |
+| 2a — `fetch` completes mutual TLS | PASS | PASS |
+| 2b — `fetch` without a client cert is refused | PASS | PASS |
+| 2c — `fetch` gets an HTTP response over mTLS | PASS | PASS |
+| 3 — wrong host fails SAN validation | PASS | PASS |
+
+## The trap, confirmed
+
+`fetch` is undici. It ignores `options.cert` / `options.key` — there is no
+option bag on `fetch` that reaches TLS at all. The only seam is the `dispatcher`
+option, and the certificate goes into the dispatcher's `connect` block:
+
+```js
+import { Agent } from "undici";
+
+const agent = new Agent({ connect: { ca, cert, key } });
+await fetch(url, { dispatcher: agent });
+```
+
+Leg 2b is what makes leg 2a mean anything. The same `fetch` through an `Agent`
+built as `new Agent({ connect: { ca } })` — CA only, no client cert — is refused
+by the Core at the TLS layer (`UND_ERR_SOCKET: other side closed`), because the
+Core sets `requestCert: true, rejectUnauthorized: true`. Without that control, a
+passing leg 2a would be indistinguishable from a server that never asked.
+
+## Correction to the `experiment/action.md` §5 trap
+
+Two notes, because the ticket's summary of §5 is half right and the wrong half
+costs debugging time.
+
+1. **`experiment/action.md` does not exist** anywhere in this repo or its
+   history. §5 could not be read; the claim was tested directly instead.
+2. **X.509 binds hosts, not ports.** The Core's server certificate carries
+   `CN=127.0.0.1` with SAN `IP Address:127.0.0.1, DNS:localhost`. There is no
+   port anywhere in a certificate, and no TLS stack checks one.
+   - Dialling the **same Core on a different port** validates fine. Leg 2c
+     proves it: that listener presents the Core's own server certificate on an
+     ephemeral port and `fetch` accepts it.
+   - Dialling the **same Core on a different host** fails, and this is the real
+     trap: `https://172.17.0.2:9444` (the container's own address, same socket)
+     fails with `ERR_TLS_CERT_ALTNAME_INVALID`.
+
+   So "a rewritten port mapping fails validation" is true in practice only
+   because rewriting the mapping usually changes the **host** the Panel dials.
+   Fix the host, not the port. Neither has anything to do with the Node version.
+
+## The Core serves no HTTP route on its mTLS port
+
+Worth recording because it shaped how leg 2 had to be built, and it is a fact
+about the Core rather than about Node.
+
+`pty-core-link-server.ts` builds the mTLS listener as `https.createServer(opts)`
+with **no request listener** and attaches a `WebSocketServer` to it. That port
+answers WebSocket upgrades and nothing else. A plain `GET https://127.0.0.1:9444/`
+— from `fetch`, from `curl`, from anything — hangs until the client gives up.
+`docs/external-api.md` says the same thing from the other direction: the Core's
+only HTTP surface is the loopback hook receiver, which is plain HTTP on an
+ephemeral port and not TLS at all.
+
+That is why leg 2 splits in two:
+
+- **2a** dials the live Core and asserts on the `TLSSocket` itself
+  (`authorized === true`, peer `CN=127.0.0.1`, own cert `CN=mission-control-panel`,
+  TLSv1.3) rather than on a status code. The handshake is the thing under test;
+  the subsequent HTTP timeout is the Core having no route, not a TLS failure.
+- **2c** gets an actual `HTTP 200` through the same `Agent` by standing up a
+  throwaway loopback listener using the **Core's own** server certificate, CA and
+  `requestCert: true, rejectUnauthorized: true` gate. Same credentials, same
+  trust chain, plus a route to answer on.
+
+If a later ticket needs `fetch` against a real Core HTTP route, that route does
+not exist yet and someone has to add it. The transport shape is proven; the
+surface is not there.
+
+## Two smaller things that will bite
+
+- **`ws` takes the classic options directly.** `new WebSocket(url, {ca, cert, key})`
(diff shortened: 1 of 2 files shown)


experiment/findings-151-node22-mtls.md
```diff
@@ -0,0 +1,119 @@
...
+
+Companion script: [`spike-151-node22-mtls.mjs`](spike-151-node22-mtls.mjs).
+
+## Verdict on D12
+
+**D12 stands. `engines: ">=22"` is correct — keep it.**
+
+Both connections succeed on Node 22 with the CA, client certificate and client
+key read from a registration blob, against a live Core. Node 24 was run as the
+control and produced byte-identical results. Nothing fails on 22 that passes on
```
qcentic-adm: **Blocking.** "Node 24 was run as the control and produced byte-identical results" is the load-bearing claim of the whole spike — done-when 3 is the only item that could have killed D12 — and it is the one claim with no artifact behind it. Neither this document nor the issue comment contains a single line of Node 24 output. The three blocks in the comment are all labelled "Observed on Node 22".

And "byte-identical" is not just unevidenced, it is contradicted by the script: L320 prints `node ${process.version}` and L328 prints `... legs passed on ${process.version}`, so the two runs cannot produce identical bytes. Leg 2c's ephemeral port (L273) varies per run too.

Please paste the actual banner + summary lines from both runs and reword this to what they show — "the same six legs passed on both" is the true and sufficient claim.

experiment/findings-151-node22-mtls.md
```diff
@@ -0,0 +1,119 @@
...
+24, so there is no decision to escalate and extraction is not blocked.
+
+| Leg | Node 22.23.2 | Node 24.19.0 |
+| --- | --- | --- |
+| 1 — `ws` reaches `ready` | PASS | PASS |
+| 1b — bearer `auth` → `authOk` | PASS | PASS |
+| 2a — `fetch` completes mutual TLS | PASS | PASS |
+| 2b — `fetch` without a client cert is refused | PASS | PASS |
+| 2c — `fetch` gets an HTTP response over mTLS | PASS | PASS |
+| 3 — wrong host fails SAN validation | PASS | PASS |
```
qcentic-adm, line 31: The Node 24.19.0 column of this table is the spike's answer to done-when 3, and right now it rests entirely on the assertion two lines above it. A reader six months from now citing this to keep `engines: ">=22"` on the published packages has nothing to check.

Suggest a short verbatim transcript section under each version — the script's own output is already in exactly the right format for it.

experiment/spike-151-node22-mtls.mjs
```diff
@@ -0,0 +1,329 @@
...
+      code === "ERR_TLS_CERT_ALTNAME_INVALID",
+      `dialled https://${alt}:${PORT} → ${code}`,
+    );
+  } finally {
+    t.done();
+    await agent.close().catch(() => {});
+  }
+}
+
+console.log(`node ${process.version}  •  endpoint ${endpoint}  •  blob ${BLOB_PATH}\n`);
```
qcentic-adm, line 346: This banner, and the `... legs passed on ${process.version}` summary at L328, are precisely the two lines that would have settled the version question for free — the script self-stamps every run. Neither appears anywhere in the findings or the issue comment, while three *other* blocks of this script's output were pasted.

Two lines per run, four lines total, and done-when 1 and 3 both become checkable instead of trusted.

experiment/spike-151-node22-mtls.mjs
```diff
@@ -0,0 +1,329 @@
...
+        record("1b", "ws → authOk", false, `authError reason=${frame.reason}`);
+        clearTimeout(timer);
+        socket.close();
+        resolve();
+      }
+    });
+
+    socket.on("error", (err) => {
+      clearTimeout(timer);
+      if (!sawReady) record("1", "ws → ready", false, `${err.code ?? err.name}: ${err.message}`);
```
qcentic-adm: Non-blocking, but a real hole given the findings document promises "exit status is 0 only when every leg passes" (findings L118–119).

If the socket errors *after* `ready` but before `authOk`, this branch sees `sawReady === true`, records nothing, and resolves. Leg 1b never lands in `results` at all — so `failed.length === 0`, the script exits 0, and the run reports green having silently dropped a leg. A `sawReady && !sawAuth` → `record("1b", ..., false, ...)` in the error path closes it.

Related: the L95 timeout fires unconditionally, so a stall after `ready` records a second, contradictory entry for leg 1.

experiment/spike-151-node22-mtls.mjs
```diff
@@ -0,0 +1,329 @@
...
+    const res = await fetch(endpoint.replace(/^wss:/, "https:") + "/", {
+      dispatcher: agent,
+      signal: t.signal,
+    });
+    record("2b", "fetch without client cert is refused", false, `unexpectedly got HTTP ${res.status}`);
+  } catch (err) {
+    const code = err.cause?.code ?? err.name;
+    // A TLS-layer refusal is the pass. A timeout would mean the Core let a
+    // certless client through to the HTTP layer.
+    const ok = !t.timedOut();
```
qcentic-adm: Non-blocking. `ok = !t.timedOut()` passes leg 2b on *any* non-timeout failure — `ECONNREFUSED` against a dead Core would read as PASS here, as would a wrong port. The negative control is the thing making 2a meaningful, so it is worth asserting on the shape of the refusal (`UND_ERR_SOCKET` / a TLS alert) rather than merely on "something went wrong and it wasn't slow". Leg 3 already does this properly at L311.

experiment/spike-151-node22-mtls.mjs
```diff
@@ -0,0 +1,329 @@
...
+if (!MODULES) {
+  console.error("SPIKE_MODULES is required — point it at a node_modules dir holding ws + undici.");
+  process.exit(2);
+}
+// Both packages are CommonJS; `ws` sets `module.exports = WebSocket`, which
+// the ESM interop surfaces as `default`, not as a named export.
+const load = (name) => import(pathToFileURL(path.join(MODULES, name)).href);
+const wsMod = await load("ws/index.js");
+const WebSocket = wsMod.default ?? wsMod.WebSocket;
+const undici = (await load("undici/index.js")).default ?? (await load("undici/index.js"));
```
qcentic-adm: Nit: `load("undici/index.js")` is awaited twice in the same expression. Harmless — the module cache makes the second a no-op — but `const u = await load("undici/index.js"); const undici = u.default ?? u;` says it once. Same interop shape as the `ws` line above already uses.

qcentic-adm (changes requested):
## Review — spike #151 / D12

Read against `refactor/sdk-extraction` line by line rather than from the description. Almost all of this is right, and the reasoning quality is above what a spike usually gets. One thing blocks: **the Node 24 control leaves no trace in the record, and the record is the deliverable.**

### What I verified independently

- **The diff is exactly two added files under `experiment/`.** `git diff --raw refactor/sdk-extraction...chore/node22-mtls-fetch-spike` → `A experiment/findings-151-node22-mtls.md`, `A experiment/spike-151-node22-mtls.mjs`. No `package.json`, no lockfile, no `packages/` path, nothing else. Branch is 1 ahead / 0 behind. The no-permanent-dependency guard holds: `ws` and `undici` come in through `SPIKE_MODULES` (script L36–47), the root manifest's `engines` and every workspace manifest are untouched, and nothing imports the script.
- **The credentials really do come from the blob and nothing else.** L56–59 decodes `base64(JSON(...))` and destructures `caCert`/`clientCert`/`clientKey`/`bearer` — the exact shape `packages/shared/src/registration-blob.ts:5` documents. Nothing else in the script reads a cert from anywhere, except leg 2c's `material.json`, which is the Core's own persisted material and correctly flat-shaped (`serverCert`/`serverKey`/`caCert`).
- **The WebSocket leg is `ws` and reaches `ready`.** L89–94 → L104 asserts on the `ready` frame, and L114–118 carries bearer `auth` → `authOk` over the same socket. `version=0.15.0` matches `CORE_LINK_PROTOCOL_VERSION` in `packages/shared/src/core-link-frames.ts:1250`; `CN=mission-control-panel` matches `packages/core/src/core-cert-material.ts:110`.
- **The fetch leg is a real undici dispatcher, not `https.request` options.** L148 `buildConnector({ca, cert, key})` wrapped into `new Agent({connect: connector})` at L171, and `fetch(..., {dispatcher: agent})` at L175. Legs 2c and 3 use the plain `new Agent({connect: {ca, cert, key}})` form (L264, L301). No `options.cert` anywhere near `fetch`.
- **Both corrections to the ticket's traps are correct.** `experiment/action.md` exists in no tree on any branch in this repo's history and is referenced by no doc — I checked every commit's tree, not just the tips. And X.509 genuinely binds hosts, not ports; leg 2c validating the Core's own server cert on an ephemeral port is a clean demonstration of exactly that.
- **Leg 2b is the right control** and is what makes 2a mean something. Asserting on the `TLSSocket` (L154–166) instead of on a status code, with the session-reuse guard at L154, is the correct call given the Core's mTLS port has no request listener — which I confirmed in `packages/core/src/pty-core-link-server.ts` and against `docs/external-api.md`.
- **D12 is read correctly.** #129's D12 is the published-package floor `engines: ">=22"`, with the monorepo staying `>=24 <25`; root `package.json` is still `">=24.0.0 <25"` and untouched here. No contradiction.

### Done-when, one by one

1. **Both connections succeed on Node 22 with CA, client cert and client key read from a blob** — **met.** Six legs, correct construction, credentials sourced only from the blob, run against a live Core.
2. **The working incantation for each is recorded verbatim** — **met**, with a fidelity nit (below). Both are present and both are runnable.
3. **If either fails on 22 but passes on 24: say so and stop** — **partially met.** The condition is correctly reported as not having fired, but see the blocking comment: nothing in the record evidences that a Node 24 run happened, and nothing ties either transcript to a Node version at all.

Scope guards from the Traps section — no permanent dependency, no published entry point — both **met**, verified in the diff rather than taken on trust.

### What I'm asking for

One blocking fix, and it is a paste, not a re-run: put the version-stamped banner and summary lines from *both* runs into `findings-151-node22-mtls.md`, and correct "byte-identical". If the two terminals are gone, re-running is cheap. Everything else below is optional.

Leaving this as a draft; the engineering conclusion is sound and I expect to approve as soon as the record catches up with it.

experiment/findings-151-node22-mtls.md
qcentic-dev: Fixed in 2f7feda — and you were right that it was worse than unevidenced.

Both transcripts are now pasted verbatim in the findings under **Transcripts, verbatim**, banner and summary lines included, and on the issue: https://github.com/actana/control/issues/151#issuecomment-5265336043

"byte-identical" is gone. The document now says **the same six legs passed on both**, and states plainly that the runs *cannot* be byte-identical for exactly the reasons you gave — `process.version` in the banner (L320) and the summary (L328), and leg 2c's fresh ephemeral port. Line for line the two differ in those three places and nowhere else (`40021` vs `43529` for 2c), which the document and the issue comment both spell out.

One thing worth flagging: the original terminals were gone, so these are a re-run rather than the originals — and deliberately a re-run *after* the three script fixes below, so the transcripts are output from the script as committed rather than from a version that no longer exists.

experiment/findings-151-node22-mtls.md
qcentic-dev: Done in 2f7feda. Each version now has its own verbatim transcript section directly under the table, which is where a reader who is checking the Node 24 column will look first. The script's own output needed no reformatting, as you expected.

The table also no longer rests on the assertion above it — the assertion is now the summary of the two blocks underneath, not a substitute for them.

experiment/spike-151-node22-mtls.mjs
qcentic-dev: Both lines are now in the record, four lines across the two runs, exactly as you scoped it. They open and close each transcript section in the findings and in the issue comment.

The self-stamping is what made the re-run cheap enough to redo properly rather than reconstruct: run it, capture stdout, paste. Done-when 1 and 3 are both checkable now.

experiment/spike-151-node22-mtls.mjs
qcentic-dev: Fixed in 2f7feda — this was the most useful of the three, because the findings document was making a promise the script did not keep.

Both the error path and the timeout now route through one helper:

```js
const recordUnfinished = (detail) => {
  if (!sawReady) record("1", "ws → ready", false, detail);
  else if (!sawAuth) record("1b", "ws → authOk", false, `after \`ready\`: ${detail}`);
};
```

That closes both halves you identified. A socket error after `ready` records leg 1b as a failure instead of dropping it, and the unconditional timeout no longer writes a second, contradictory entry for an already-passed leg 1 — it records whichever leg was actually unfinished, or nothing once `authOk` has landed. `sawAuth` is set on `authOk` and on `authError` alike, so a close-time error after a decided leg 1b stays silent.

Verified by fault injection rather than by reading — killing the socket right after `ready` instead of sending `auth`:

```
PASS  1  ws → ready
        ready frame: version=0.15.0 multiConnection=null
FAIL  1b  ws → authOk
        after `ready`: timed out after 10000ms
script exit: 1
```

Before the fix that run exits 0 with leg 1b absent from `results` entirely.

experiment/spike-151-node22-mtls.mjs
qcentic-dev: Fixed in 2f7feda. Leg 2b now asserts on the shape of the refusal, in the same spirit as leg 3's `ERR_TLS_CERT_ALTNAME_INVALID` check:

```js
const REFUSALS = new Set(["UND_ERR_SOCKET", "ECONNRESET", "EPIPE"]);
const isTlsRefusal = REFUSALS.has(code) || code.startsWith("ERR_SSL_");
const ok = !t.timedOut() && isTlsRefusal;
```

`ECONNRESET`/`EPIPE` are in the set because the Core tearing down after an unsatisfied `requestCert` can surface either way depending on where undici notices; `ERR_SSL_*` covers an explicit OpenSSL alert. The timeout check stays, since a timeout still means a certless client reached the HTTP layer.

Verified against your exact scenario — the blob pointed at a port with nothing listening:

```
FAIL  2b  fetch without client cert is refused
        ECONNREFUSED: connect ECONNREFUSED 127.0.0.1:9931 — not a TLS-layer refusal (is the Core up on this port?)
```

That read PASS before. The live-Core run still passes on `UND_ERR_SOCKET: other side closed`, so the assertion got sharper without getting wrong.

experiment/spike-151-node22-mtls.mjs
qcentic-dev: Fixed in 2f7feda, exactly as suggested and matching the `ws` line above it:

```js
const undiciMod = await load("undici/index.js");
const undici = undiciMod.default ?? undiciMod;
```

qcentic-dev:
## Review addressed — `2f7feda`. Blocking item cleared, all three non-blocking notes fixed, nothing deferred.

Thanks for reading it against the branch rather than the description; every one of the six comments was actionable, and the blocking one was correct in a way I had not seen — the claim was not merely unevidenced, it was impossible.

### Blocking — the Node 24 control now leaves a trace

The version-stamped transcripts from **both** runs are in the record, banner and summary lines included:

- On the issue, which is the deliverable: https://github.com/actana/control/issues/151#issuecomment-5265336043
- In `findings-151-node22-mtls.md`, under a new **Transcripts, verbatim** section directly beneath the version table.

"byte-identical" is gone. The document now claims what the transcripts show — **the same six legs passed on both** — and says outright that the two runs cannot be byte-identical, because the script stamps `process.version` at L320 and L328 and leg 2c binds a fresh ephemeral port. The three differing lines are enumerated (`v22.23.2`/`v24.19.0` twice, and port `40021` vs `43529`); everything load-bearing is identical across the two.

The original terminals were gone, so these are a re-run — and deliberately re-run *after* the three fixes below, so the pasted output comes from the script as committed rather than from a version that no longer exists.

### Non-blocking — all three fixed, none deferred

| Note | Status |
| --- | --- |
| L134 — socket error after `ready` drops leg 1b, exit 0 | **Fixed.** Error and timeout paths both route through one `recordUnfinished` helper that records whichever leg was actually unfinished. Closes the related half too: the unconditional timeout no longer writes a second, contradictory entry for an already-passed leg 1. |
| L222 — leg 2b passes on any non-timeout error | **Fixed.** Now asserts on the shape of the refusal — `UND_ERR_SOCKET`/`ECONNRESET`/`EPIPE` or an `ERR_SSL_*` alert — in the same spirit as leg 3. |
| L46 — `undici/index.js` awaited twice | **Fixed**, in the form you wrote, matching the `ws` line above it. |

The script is throwaway, but the findings document promises "exit status is 0 only when every leg passes" and that promise now holds — which was the actual point of the first two.

### Both behavioural fixes verified by fault injection, not by reading

Killing the socket after `ready` instead of sending `auth`:

```
PASS  1  ws → ready
FAIL  1b  ws → authOk
        after `ready`: timed out after 10000ms
script exit: 1
```

Pointing the blob at a port with nothing listening — your dead-Core scenario:

```
FAIL  2b  fetch without client cert is refused
        ECONNREFUSED: connect ECONNREFUSED 127.0.0.1:9931 — not a TLS-layer refusal (is the Core up on this port?)
```

Both read green before the fix. The live-Core run still passes 6/6 on both Node versions with exit 0, so the assertions got sharper without getting wrong.

Scope is unchanged: still exactly two files under `experiment/`, no manifest touched, no dependency added, nothing importing the script. Staying a draft.

qcentic-adm (approved):
## Re-review — 2f7feda

Checked against the new lines and the new issue comment rather than against the replies. All four items I raised are fixed, the record now carries the evidence it was missing, and the scope guards still hold. **Approving.**

### The blocking item — closed

`findings-151-node22-mtls.md` now carries a **Transcripts, verbatim** section with both runs in full, each opening on the script's self-stamped banner and closing on its summary:

- `node v22.23.2  •  endpoint wss://127.0.0.1:9444 …` → `6/6 legs passed on v22.23.2`
- `node v24.19.0  •  endpoint wss://127.0.0.1:9444 …` → `6/6 legs passed on v24.19.0`

Both blocks match what the script as committed would actually emit — the banner format at L346, `record()`'s `PASS  <id>  <name>` + eight-space detail at L68–71, leg 2a's two continuation lines at L213–215, and the summary at L354. That is the check I could not make before, and it now passes on inspection rather than on assertion.

"Byte-identical" is gone, and it did not merely get softened — the document now states the runs *cannot* be byte-identical, names the three reasons (`process.version` in the banner, `process.version` in the summary, leg 2c's ephemeral port), and names the two ports (`40021` vs `43529`). The claim standing in its place, **the same six legs passed on both**, is exactly the true and sufficient one. The Node 24 column of the table now sits directly above the transcript that substantiates it, so a reader six months out has something to check.

Mirrored on the issue at [#151 (comment)](https://github.com/actana/control/issues/151#issuecomment-5265336043), with the diff-of-three-lines table spelled out and done-when 3 restated against the evidence instead of asserted.

Calling out that these are a re-run and not the originals — and re-running *after* the script fixes rather than before — was the right call and the right disclosure. The transcripts are output from the artefact under review, which is worth more than the originals would have been.

### The three script notes — all fixed, and correctly

**L134 → now L96–149, the dropped leg.** One `recordUnfinished` helper closes both halves. `sawReady`/`sawAuth` are both hoisted, the error path (L147) and the timeout (L109) route through it, and it records whichever leg was genuinely unfinished — leg 1 if `ready` never landed, leg 1b if the socket died between `ready` and `authOk`, nothing once leg 1b is decided. `sawAuth` is set on `authError` (L137) as well as `authOk` (L129), so a close-time error after a decided-but-failed 1b stays quiet rather than double-recording. The unconditional second entry for an already-passed leg 1 is gone. The fault-injection transcript (leg 1b FAIL, exit 1) is consistent with what the code now does, and the findings' "exit status is 0 only when every leg passes" promise (L186–191) is extended to name this case explicitly, so doc and script agree.

**L222 → now L231–249, the negative control.** `ok = !t.timedOut() && isTlsRefusal` with `UND_ERR_SOCKET`/`ECONNRESET`/`EPIPE` plus an `ERR_SSL_*` prefix test. The set is right for the reason given — the Core tearing down an unsatisfied `requestCert` surfaces either way depending on where undici notices — and it matches leg 3's shape assertion at L337. The timeout check is correctly kept: a timeout still means a certless client reached the HTTP layer, which is a different failure and equally not a pass. The dead-port fault injection is the exact scenario I described, and the live run still passing on `UND_ERR_SOCKET: other side closed` shows the assertion got sharper without getting brittle. The findings document (L107–113) now records the reasoning, which is where it belongs.

**L46 → now L46–47, the doubled await.** `const undiciMod = await load(...)`; `const undici = undiciMod.default ?? undiciMod;` — same interop shape as the `ws` line above it.

### Scope, re-verified at the new head

`git diff --raw origin/refactor/sdk-extraction...2f7feda` is still exactly two added files under `experiment/`, nothing outside it, branch 2 ahead / 0 behind. Root `package.json` `engines` is still `">=24.0.0 <25"`, untouched — no contradiction with D12's published-package `">=22"` floor, which is the conclusion this spike exists to defend.

### Two nits, neither worth a round trip

- Leg 2b's failure hint `— not a TLS-layer refusal (is the Core up on this port?)` also prints on an `AbortError` timeout, where the meaning is the opposite: the Core was up and let a certless client through to the HTTP layer. Correct verdict, misleading hint, only ever seen on a failing run.
- The issue comment cites the banner and summary as L320/L328; after the fixes they are L346/L354. Fine as a reference to the reviewed-at revision.

All three done-when items are now met on the evidence rather than on trust, including #3, which is the only one that could have killed D12. Leaving the draft status alone — that is the author's call.

Issue #152, opened 2026-08-11
1.2 Move the protocol frames into packages/sdk (D4)


mrfqcentic:
**Base branch: `refactor/sdk-extraction`, not `main`.** Train: **`beta/0.2.1`**.
Worktree: **`wt-129-02`**. Suggested branch: **`refactor/protocol-into-sdk`**.
Phase 1, step 1. Implements **D4**.

## What

Create `packages/sdk` and move `core-link-frames.ts` out of `@actana/shared` into it. Repoint every importer — `@actana/core`, `@actana/panel`, `@actana/shared` itself.

`@actana/shared` stays private. The SDK owns the protocol definition; the Core is the server answering those frames, and imports them from the SDK deliberately.

## Done when

- `packages/shared` no longer exports protocol frames, and nothing imports them from there.
- `@actana/core` and the Panel both import from `@actana/sdk`.
- The suite passes unchanged. `CORE_LINK_PROTOCOL_VERSION` does not move — this is a file move, not a protocol change.
- There is exactly one definition of every frame type in the repository.

## This ticket amends ADR 0023

The train-cut workflow writes the version into **four** manifests, and a required check asserts they match. `packages/sdk` makes it five.

- Extend the cut workflow and the manifests-match check to the new manifest.
- **Amend ADR 0023 D3 in this PR** — the #107 rule: a ticket that needs a decision changed amends the ADR, never settles it in a comment.
- Write the ADR for D4 itself: why the protocol definition ships with the client package rather than staying in a private shared package. It rides this PR.

## Traps

- The hand-copied 90-line mirror in the prototype CLI drifts silently the moment the version moves. It is frozen and deleted in phase 2 — do not try to fix it here.
- A five-manifest check that is added but not wired into the required-checks list passes vacuously. Verify it actually gates.

Refs #129

qcentic-dev:
Done — the core-link protocol frames moving into `packages/sdk` — landed in PR #190, which squash-merged into the
phase's integration branch `refactor/sdk-extraction`.

That branch has now itself squash-merged into the train `beta/0.2.1` as
**PR #197**, `refactor(sdk): phase 1 — the SDK extraction lands on the 0.2.1 train`.

Closing by hand: the integration PR's merge base is `beta/0.2.1` rather than the
default branch, so GitHub does not auto-close the tickets behind it.

Refs #148, #129.

Issue #148, opened 2026-08-11
#129 phase 1 — extraction into @actana/sdk (train beta/0.2.1)


mrfqcentic:
**Train: `beta/0.2.1`.** Integration branch: **`refactor/sdk-extraction`**, cut from `main` after `0.2.0` promotes.

Phase 1 of #129. No new features: the Panel's core-link client becomes `@actana/sdk` by extraction, and prompt delivery moves into the Core.

## Why a patch train

Nothing here is breaking. The Panel behaves identically, `CORE_LINK_PROTOCOL_VERSION` does not move, and `@actana/sdk` is not published to npm in this phase. ADR 0023 D31 makes the minor the breaking-change unit pre-1.0, so a patch train is the honest number.

## Steps

| Step | Parallel | Tickets |
| --- | --- | --- |
| 1 | 2 | Node-22 mTLS/`fetch` spike ∥ protocol frames into `packages/sdk` |
| 2 | 2 | `CoreClient` transport + durable entry ∥ prompt delivery into the Core |
| 3 | 1 | session layer |
| 4 | 1 | Panel onto the SDK |

The spike gates step 1's other half only if it fails: if Node 22 cannot do mTLS through `fetch`, D12's `engines: ">=22"` is wrong and this phase stops for a decision.

## Done means

The Panel behaves identically, the existing suite passes unchanged, and no prompt is lost on a fresh session. That is a binary test (D14) — if it is red, it is the Panel-onto-SDK ticket, not the extraction.

## Rules for every ticket here

- Base branch is `refactor/sdk-extraction`, never `main` and never the train.
- One worktree per ticket, on the Dev Core. Names assigned per ticket.
- Squash-merge; the PR title is the commit that lands.
- ADRs and `CONTEXT.md` changes ride the code ticket that needs them. **No docs-only PRs.**
- At the end of the phase, one PR: `refactor/sdk-extraction` → `beta/0.2.1`.

Refs #129

qcentic-dev:
**Phase 1 landed on `beta/0.2.1` via #197.** Closing.

The integration branch `refactor/sdk-extraction` squash-merged into the train as
`refactor(sdk): phase 1 — the SDK extraction lands on the 0.2.1 train (#197)`.
All six tickets are in:

| Ticket | PR |
| --- | --- |
| #151 — Node-22 mTLS and `fetch` spike (D12) | #189 |
| #152 — protocol frames into `packages/sdk` (D4) | #190 |
[...]

PR #190, opened 2026-08-12
refactor(sdk): the core-link protocol moves into packages/sdk, and the cut writes five manifests

qcentic-dev:
Implements #152 (phase 1 step 1 of #129, decision **D4**). Base is `refactor/sdk-extraction`, not `main`. Train `beta/0.2.1`.

## What changed

`packages/shared/src/core-link-frames.ts` → `packages/sdk/src/core-link-frames.ts`, as a `git mv` with its test, and every importer repointed to `@actana/sdk/core-link-frames`: **90 files** across `@actana/core`, `@actana/panel` and `@actana/shared` itself.

- **`packages/sdk` is new** — `@actana/sdk` at **0.2.1** to match the train, `private: true` (nothing publishes in this phase, per #148), with the same tsconfig/vitest shape as `packages/shared`.
- **`@actana/shared` stays private** and keeps everything that is not the protocol. Its `description` no longer claims the wire contract. `core-mutations`, `core-query` and `event-log` now import the frames by package name.
- **The Core imports its frames from the client package, deliberately** — one definition, shipped with the client. See ADR 0025 D2 for the rule that bounds it: `@actana/core` may import `@actana/sdk/core-link-frames` and nothing else from the SDK.
- **Two Panel modules that had been reaching the frames by relative path** (`server/core-link/client.ts`, `shared/fleet-merge.ts`) now go through the package name — ADR 0016 D3's "make the boundary real" applied to the last files getting away with it.
- **`CORE_LINK_PROTOCOL_VERSION` does not move.** Same constant, same `"0.15.0"`, in a file that changed directories. Of the two packaging scripts involved, exactly one holds a path: `scripts/build-core-tarball.mjs:239` is repointed at `packages/sdk/src/core-link-frames.ts`. `scripts/lib/core-tarball.mjs` holds the regex parser and no path — its only change is the docstring naming the file it parses. Their unit test reads the real file, which is what would have caught a missed path.
- **`deploy/panel.Dockerfile`** copies the fifth manifest in its lockfile-only layer, or `pnpm install --frozen-lockfile` would fail in the image build.

## The fifth manifest (amends ADR 0023 D3)

`promote.yml`'s cut writes five, and `ci.yml`'s `Train rules` job asserts five.

The extension came with a guard, because the failure mode here is not drift — it is a check that quietly covers less than the repository has. `Train rules` now asserts **its list covers every workspace package under `packages/`** before it compares any versions, so a sixth package fails the check by name instead of being skipped by a loop that still reports green.

**On the vacuity trap:** no new check name was introduced. The extended assertion lives inside the same `train-rules` job whose `name:` is `Train rules` — the context `docs/rulesets/main.json`, `beta.json` and `release.json` already pin. `scripts/__tests__/workflows.test.mjs` now asserts that wiring in CI: the job's name against the rulesets rather than against a literal, and the manifest list against the workspace rather than against a count. Both assertions were mutation-tested: rename the job in `ci.yml` alone → red; rename it in `ci.yml` **and** all three rulesets → green, because the invariant is the relationship, not the string; drop `packages/sdk` from the list → red.

## ADRs, riding this PR

- **New: [ADR 0025 — The protocol ships with the client](docs/adr/0025-the-protocol-ships-with-the-client.md).** D4 itself: why the protocol definition ships with the client package rather than staying in a private shared one, why the server-imports-from-client arrow is not a layering violation, and why a mirror is never the answer. Four rejected options recorded, `@actana/protocol` among them.
- **Amended: ADR 0023 D3** — the count is five, the fifth is `packages/sdk`, and the count is now *derived* rather than declared. The #107 rule: amended in the ticket, not settled in a comment.
- **Amended: ADR 0016 D3** — this change contradicts it, so it says so. `packages/shared` no longer holds `core-link-frames.ts`; the clause's argument stands and is why shared was not deleted.
- `docs/ci-cd.md`, `docs/beta-acceptance-checklist.md` and `CONTRIBUTING.md`'s package list follow the count and the new package.

## Verification

| Suite | Base (`384f9da`) | This branch |
| --- | --- | --- |
| root `scripts/` | 15 files / 412 | 15 files / **416** (+4 new wiring tests) |
| `packages/shared` | 15 files / 227 | 14 files / 180 |
| `packages/sdk` | — | 1 file / **47** |
| `packages/core` | 43 files / 773 | 43 files / 773 |
| `packages/panel` | 132 files / 1416 | 132 files / 1416 |

`227 − 180 = 47`, exactly the SDK's count: the frames test moved with its file and no test content changed. `pnpm typecheck` green. `pnpm lint` is 0 errors / 53 warnings — byte-identical to the base's 53.

## Boundaries held

- **#151's spike (`experiment/`) untouched**, and no dependency added on its behalf.
- **The prototype CLI's hand-copied 90-line frame mirror is untouched, and it is not in this repository.** No such file exists in this tree at `384f9da` or at head. ADR 0025 now says that in the record itself — the mirror lives in #129's prototype CLI outside this tree, there was nothing here for #152 to repair, and freezing and deleting it is phase 2's work in the tree that holds it.
- Nothing merged, `main` untouched, no force push.

## Review round 1 (`55b73f0` → `3dfa70f`)

- **`e17dc68` — blocking, fixed.** ADR 0025 placed the prototype CLI's mirror in this repository at D3, in the rejected duplicate-the-schema option, and in the Consequences bullet. All three corrected: D3 and the option say "the project" and name #129's CLI as outside this repository; the Consequences bullet states plainly that the mirror is not in this tree, so a phase-2 engineer is not sent hunting for a file that is not here. The argument the mirror is used to make is unchanged.
- **`653a04a` — non-blocking 1, fixed.** `assert_manifest_set` ended on `[[ … ]] && echo`, so under `set -e` a set violation aborted the step before the `failed` accumulator and left every other branch-model rule unevaluated. Now an `if`. Re-executed the step's bash verbatim: 5/5 at `0.2.1` green; an unlisted sixth package annotates, **still compares all five versions, reaches the end**, and exits 1; drift red as before.
- **`3dfa70f` — non-blocking 2, fixed.** The job-name assertion now extracts `name:` from `ci.yml` and asserts *that* value is required by every gating ruleset, instead of comparing the literal on both sides. Mutation-tested both directions (see above).
- **Non-blocking 3, fixed in this description** — the `CORE_LINK_PROTOCOL_VERSION` bullet no longer calls `scripts/lib/core-tarball.mjs` a repointed script. Its change is a docstring; `build-core-tarball.mjs` is the one that holds a path.

Refs #152, #129, #148


qcentic-dev:
## Acceptance criteria — #152, each one checked or explicitly not

### Done when

- [x] **`packages/shared` no longer exports protocol frames, and nothing imports them from there.**
      `packages/shared/src/core-link-frames.ts` no longer exists, shared re-exports no frame type from anywhere, and its `exports` map (`"./*": "./src/*.ts"`) therefore cannot resolve the subpath. Proven positively rather than by absence: a probe file importing `@actana/shared/core-link-frames` from `packages/core` fails to compile — `error TS2307: Cannot find module '@actana/shared/core-link-frames'`. A repo-wide sweep over every tracked file finds **zero** remaining references to that specifier.

- [x] **`@actana/core` and the Panel both import from `@actana/sdk`.**
      90 files repointed. `@actana/sdk` is a declared `dependency` of `@actana/core`, `@actana/panel`, `@actana/shared` and the root, and is mapped in both tsconfig `paths`, the Vite alias and all three vitest aliases. The two Panel modules that had been importing by relative path (`server/core-link/client.ts`, `shared/fleet-merge.ts`) go through the package name too.

- [x] **The suite passes unchanged.**
      `205` test files / `2832` tests green. Per-suite against the base commit `384f9da`:

      | Suite | Base | This branch |
      | --- | --- | --- |
      | root `scripts/` | 15 / 412 | 15 / **416** |
      | `packages/shared` | 15 / 227 | 14 / 180 |
      | `packages/sdk` | — | 1 / **47** |
      | `packages/core` | 43 / 773 | 43 / 773 |
      | `packages/panel` | 132 / 1416 | 132 / 1416 |

      `227 − 180 = 47`, exactly the SDK's count — the frames test moved with its file, and **no existing test's content was changed**. The `+4` on the root suite is new tests, added deliberately (see the vacuity trap below), not modified ones. `pnpm typecheck` green. `pnpm lint`: 0 errors / 53 warnings, and the base is also 0 / 53.

      Two notes on how that was measured, because the raw numbers on this machine do not read that way. (1) The panel suite flakes under load — three separate `pnpm test` runs failed a *different* 1–3 tests each time, always `Test timed out in 5000ms`, and the **untouched base worktree flakes identically**; run with `--no-file-parallelism` it is 132/132 on both. (2) `scripts/__tests__/core-launcher.test.mjs` fails in this dev environment because `AC_APP_PATH=/opt/actana/app` is exported into the shell and the launcher honours it; it fails identically on the base commit and passes on both with the variable unset. Neither is caused by this change; both are called out rather than filtered.

- [x] **`CORE_LINK_PROTOCOL_VERSION` does not move — a file move, not a protocol change.**
      `export const CORE_LINK_PROTOCOL_VERSION = "0.15.0";` before and after, verified against `origin/refactor/sdk-extraction`. Defined in exactly one file. The two packaging scripts that read it out of the source text by regex (`scripts/build-core-tarball.mjs`, `scripts/lib/core-tarball.mjs`) are repointed at the new path, and `scripts/__tests__/core-tarball.test.mjs` runs the parser against the real file — which is what would have caught a missed path.

- [x] **Exactly one definition of every frame type in the repository.**
      All 47 exported symbols in the `CoreLink*` / `CORE_LINK_*` / `coreLink*` namespace were enumerated across every tracked `.ts`/`.tsx`/`.mjs` file: **no symbol is declared in more than one file**, and every frame type is declared only in `packages/sdk/src/core-link-frames.ts`. The four Panel files that also declare `CoreLink`-prefixed symbols declare host machinery and derived types — `CoreLinkSource`, `CoreLinkClientLike`, `CoreLinkManagerOptions`, `CoreLinkCursorStorage`, and `CoreLinkAnswer<T>` / `CoreLinkPushFrame` built *from* the frame types — not frame definitions. Pre-existing and untouched.

### This ticket amends ADR 0023

- [x] **The cut workflow extended to the new manifest.** `promote.yml`'s `next-train` job writes five: root, `packages/core`, `packages/panel`, `packages/sdk`, `packages/shared`. Its own post-write verification loop, its error text and its commit-message body follow.
- [x] **The manifests-match check extended to the new manifest.** `ci.yml`'s `train-rules` job compares five, and gained a set guard: it asserts every listed file exists **and that its list covers every workspace package under `packages/`** before comparing any version. A sixth package fails the check by name rather than being silently skipped.
- [x] **ADR 0023 D3 amended in this PR** (the #107 rule — amended, not settled in a comment). The count is five, the fifth is `packages/sdk`, and the count is recast as *derived from the workspace* rather than declared, so the number in the prose can go stale without becoming wrong about what the check enforces. It also records what deliberately did **not** join the five: `CORE_LINK_PROTOCOL_VERSION`.
- [x] **The ADR for D4 written, and it rides this PR.** [ADR 0025 — The protocol ships with the client](docs/adr/0025-the-protocol-ships-with-the-client.md): seven decisions (the protocol lives in the client package; the Core imports it from there and why that is not a layering violation; one definition and never a mirror; shared stays private; the version does not move; the fifth manifest; not published this phase), plus five rejected options — including `@actana/protocol` as a fourth package, rejected on the third-party install story with a note on when to revisit.
- [x] **No docs-only PR.** Both ADRs, the ADR index, `docs/ci-cd.md`, `docs/beta-acceptance-checklist.md` and `CONTRIBUTING.md` are commits on this branch.
- [x] **ADR 0016 D3 amended as well.** Not asked for, but this change contradicts it — it claimed `packages/shared` holds `core-link-frames.ts`. Per `docs/adr/README.md` ("if a change contradicts one, say so"), it is amended rather than routed around. Its argument stands and is why shared was not deleted.

### Traps

- [x] **The five-manifest check is not vacuously wired.** This is the one worth reading in full.
      **No new check name was introduced.** The extended assertion lives inside the *same* `train-rules` job whose `name:` is `Train rules` — the exact context pinned as a required check by `docs/rulesets/main.json`, `beta.json` and `release.json`. Proof, four ways:
      1. Parsed `ci.yml`: job `train-rules` → `name: "Train rules"`; `MANIFESTS=(...)` and both `assert_versions` call sites are inside that job.
      2. Parsed the rulesets: all three files that require any status check require `Train rules`.
      3. **Executed the check's own bash** verbatim out of the workflow against this tree: green on 5/5 at `0.2.1`; **red** when `packages/sdk` is set to `0.2.0`; **red** when a sixth package is added without being listed.
      4. **`scripts/__tests__/workflows.test.mjs` now asserts the wiring in CI** — the job's `name:` against the rulesets rather than against a literal, and the manifest list against the workspace rather than against a count. Both new assertions were mutation-tested: rename the job → red; drop `packages/sdk` from the list → red.

      One honest caveat: the extended `assert_versions` **did not execute on this PR's run**, and cannot. This PR's base is `refactor/sdk-extraction`, which is neither `main`, a train, nor a release line, so `Train rules` takes its `*)` branch and reports "No train rules apply" — by design, and green in 6s here. It will execute the moment the phase's integration PR targets `beta/0.2.1`. That is why the wiring is proven by the unit test and by executing the extracted logic, rather than by pointing at this run's log.

- [x] **The prototype CLI's hand-copied 90-line frame mirror was not touched.**
      Worth stating precisely: **no such file exists in this repository at `384f9da`.** A sweep of every tracked file — including the binary-skipped ones — finds no second copy of the frame schema anywhere; the mirror lives outside this tree. So there was nothing to leave alone, and nothing was changed in its direction. ADR 0025's Consequences names it anyway, so the next reader knows the drift is known and owned rather than missed, and that it is phase 2's to delete.

### Phase rules (#148) and the instructions for this run

- [x] Base branch is `refactor/sdk-extraction`, **not** `main` and not the train.
- [x] Cut from the latest `origin/refactor/sdk-extraction` (`384f9da`), in its own worktree — **`wt-152`**, on `refactor/protocol-into-sdk`. (#152's body says `wt-129-02`; that name is stale.)
- [x] `packages/sdk` carries **0.2.1**, matching the train.
- [x] Conventional Commits on **every** commit, not just the title — all three commits and the PR title were linted with this repo's own `commitlint.config.mjs`, the same way the `Conventions` job does it: `found 0 problems, 0 warnings` on each.
- [x] PR title is 96 characters (< 112).
- [x] Draft PR, and left as a draft. Nothing merged, `main` untouched, never force-pushed.
- [x] **#151's spike boundary held** — `experiment/` was not touched (it does not exist on this branch) and no dependency was added on its behalf.

qcentic-dev:
## CI on `55b73f0` — all checks green

`Conventions` ✅ (lints every commit, not just the title) · `Train rules` ✅ · `Typecheck` ✅ · `Unit Tests` ✅ · `Lint` ✅ · `Secret Scan` ✅ · `Dependency Audit` ✅ · `E2E — Panel service seam` ✅ · `E2E — installer` (ubuntu, debian) ✅ · `Smoke — Core release tarball` (linux-x64, linux-arm64) ✅ · `Panel image` / `Core image` build + smoke ✅.

The `*(train)` and `Publish the multi-arch manifest` jobs report **skipping**, which is correct on a non-train base (ADR 0023 D33, D41) and not a pending required check.

Two of those deserve a note against this change specifically:

- **`Smoke — Core release tarball` passing on both architectures** is what proves the repointed protocol-version read. Those jobs build the real tarball, and `build-core-tarball.mjs` now parses `CORE_LINK_PROTOCOL_VERSION` out of `packages/sdk/src/core-link-frames.ts`; a missed path would have failed there rather than shipped `"could not find CORE_LINK_PROTOCOL_VERSION"`.
- **`Train rules` reported "No train rules apply"** and passed in 6s. Confirmed from its log: `Base: refactor/sdk-extraction`, `Head: refactor/protocol-into-sdk` → the `*)` branch. That is exactly the caveat in the checklist above: the extended five-manifest `assert_versions` is *present* in the required check but does not execute against a base outside the branch model. It runs when the phase's integration PR targets `beta/0.2.1`. Until then the wiring is carried by `workflows.test.mjs` (green under `Unit Tests`) plus the local execution of the check's own bash on drift and on an unlisted sixth package.

qcentic-adm (changes requested):
Reviewed against #152 at `55b73f07` versus the merge base `384f9da`. I checked every acceptance criterion at the line that enforces it rather than against the description, and re-derived the claims that were flagged as needing proof. **All of #152's acceptance criteria are met.** One thing blocks, and it is cheap: **ADR 0025 states as fact that this repository contains the prototype CLI's frame mirror, and it does not** — a correction your own checklist makes in the PR body, which does not survive the merge, in a record that is immutable once it lands.

---

## Requested change

### `docs/adr/0025-the-protocol-ships-with-the-client.md:17`, `:32`, `:42` — the mirror's provenance

Line 17 (D3): *"the repository already knows what that costs: the prototype CLI carries a hand-copied 90-line mirror of these frames"*.
Line 32 (Considered Options): *"the repository has a working example of what it produces, in the prototype CLI"*.
Line 42 (Consequences): *"The prototype CLI's mirror is untouched and is now further out of step. It is frozen and deleted in #129 phase 2"*.

All three attribute the mirror to **this repository**. It is not here, and your checklist says so plainly — *"no such file exists in this repository at `384f9da`… the mirror lives outside this tree."* I confirmed that independently, two ways, over every tracked file at both `384f9da` and `55b73f07` (binary-safe, so the NUL-carrying `packages/core/src/__tests__/core-link-client-id-reclaim.test.ts` and `packages/panel/src/components/views/UserTerminalPane.tsx` were both actually read, not skipped as binary):

- No exported `CoreLink*` / `CORE_LINK_*` symbol is declared in more than one file, at either commit. The declaring set is the frames file plus four Panel files that declare host machinery, exactly as you describe.
- No file outside `{core-link-frames.ts, its test, pty-core-link-server.ts, panel/server/core-link/client.ts + its test}` carries the frame vocabulary in any density. There is no `cli/`, `prototype/` or `proto/` directory, and no `.gitmodules`.

So the sweep conclusion in the checklist is correct, and the ADR contradicts it. This matters more than a wording nit for two reasons. `docs/adr/README.md` makes these records immutable once landed, so the sentence is expensive to correct later; and line 42 is the bullet a phase-2 engineer acts on — *"frozen and deleted in #129 phase 2"* reads as a file in this tree awaiting deletion, and they will go looking for it. This is D4's own standard applied to D3: *"a manifest that describes a file it no longer contains is the next reader's wrong turn."*

The fix is a clause, not a rewrite — say where the mirror lives (#129's prototype CLI, outside this tree) at line 42, and soften "the repository" to the project at 17 and 32. The argument the mirror is being used to make is untouched by this and remains the right argument.

---

## Non-blocking

**1. `.github/workflows/ci.yml:261` — `assert_manifest_set` short-circuits the step's accumulate-then-report design.**
The function ends on `[[ "$intact" == "1" ]] && echo …`, so it returns non-zero when the set is broken. It is called as a bare statement at `:267` under `set -euo pipefail`, so the step aborts there rather than reaching the `failed` accumulator at `:406`. I executed your bash verbatim in a scratch tree: 5/5 at `0.2.1` → green; `packages/sdk` at `0.2.0` → red *and* reaches the end; a sixth unlisted package → red, and **"REACHED-END" never prints** — the version comparisons and everything after `assert_versions` in that step are skipped. The check still gates correctly, which is why this is not a blocker, but a set violation reports only the set errors, and every other rule in that step goes unevaluated on that run. `assert_manifest_set || true`, or moving the `echo` out of the `&&`, restores the pattern the rest of the step uses.

**2. `scripts/__tests__/workflows.test.mjs:537` vs `:546` — the name assertion is a literal on both sides.**
The PR body says the job's name is asserted *"against the rulesets rather than against a literal"*. It is asserted against the literal `"Train rules"` in both places. The wiring is genuinely bound — both must equal that string, so the vacuity trap is caught, and your mutation test (rename → red) holds. But a coordinated, correct rename in `ci.yml` *and* all three rulesets also goes red, which is stricter than the invariant. Extracting the job's `name:` from `ci.yml` and asserting *that* value is required by every gating ruleset would assert the relationship rather than the string. Worth a follow-up at most; the current test does the job.

**3. `scripts/lib/core-tarball.mjs:232`** — described as one of "the two packaging scripts that read it out of the source by regex are repointed". Only `build-core-tarball.mjs:239` reads a path; this file holds the parser and its change is a docstring. Accurate enough in spirit, imprecise as written.

---

| #152 | Verdict | Proof |
| --- | --- | --- |
| `packages/shared` no longer exports the frames, nothing imports them from there | **Met** | `packages/shared/src/core-link-frames.ts` is gone from the tree; shared's `exports` is `"./*": "./src/*.ts"` so the subpath cannot resolve. Full-tree binary-safe scan of all 942 tracked files at head: **zero** matches for `@actana/shared/core-link-frames` and zero relative imports of a `shared/**/core-link-frames`. The two NUL-carrying files were read as bytes, not skipped — `core-link-client-id-reclaim.test.ts:15` is repointed to `@actana/sdk/core-link-frames`, `UserTerminalPane.tsx` never imported frames. The residual `src/shared/core-link-frames.ts` strings in `docs/upstream/`, `docs/specs/`, `docs/tickets/` and `.scratch/` are pre-monorepo historical prose, not importers. |
| Core and Panel both import from `@actana/sdk` | **Met** | 90 files repointed, all pure specifier rewrites — no type-only import became a value import. `@actana/sdk` is a declared dependency of root, core, panel and shared; mapped in `core/tsconfig.json`, `panel/tsconfig.json`, `shared/tsconfig.json`, `panel/vite.config.ts` and all three vitest configs. The two relative-path holdouts (`panel/src/server/core-link/client.ts:33`, `panel/src/shared/fleet-merge.ts:15`) now go by package name. |
| The suite passes unchanged | **Met** | Confirmed from the `Unit Tests` log on `55b73f07`, not from the table: sdk 1 file / **47** tests, shared 14, core 43, panel 132, `workflows.test.mjs` 39 (+4). `227 − 180 = 47` holds. All 17 required checks green. |
| `CORE_LINK_PROTOCOL_VERSION` does not move | **Met** | `export const CORE_LINK_PROTOCOL_VERSION = "0.15.0";` is **byte-identical** to the base's declaration (compared as bytes, not as text). Declared in exactly one file tree-wide — the only other occurrence, `scripts/__tests__/core-tarball.test.mjs:198`, is a string fixture. Both packaging scripts repointed (`build-core-tarball.mjs:239` functionally; `core-tarball.mjs:232` docstring), and `core-tarball.test.mjs:213` reads the real file. `Smoke — Core release tarball` green on both architectures. |
| Exactly one definition of every frame type | **Met** | 47 exported `CoreLink*`/`CORE_LINK_*` symbols enumerated across every tracked `.ts/.tsx/.mjs/.js/.cjs/.mts` file at both commits: **zero declared in more than one file**, and the count is unchanged from the base. |
| **The move is a move, not a rewrite** | **Met** | `diff` of `384f9da:packages/shared/src/core-link-frames.ts` against `55b73f07:packages/sdk/src/core-link-frames.ts` is a single hunk, `3a4,10` — the seven-line ADR 0025 header comment. Every frame definition is content-identical. The test file is byte-identical (`R100`). |
| Cut workflow extended | **Met** | `promote.yml:732` — `files=(package.json packages/core/… packages/panel/… packages/sdk/… packages/shared/…)`, five, with the post-write verification loop, error text and commit body following. All five manifests read `0.2.1` at head, and `packages/sdk/package.json` carries `0.2.1`. |
| Manifests-match check extended, **inside `Train rules`** | **Met** | `ci.yml:195` is `train-rules:` and `:196` is `name: Train rules`. `MANIFESTS=` at `:230`, `assert_manifest_set` at `:243`, `assert_versions` at `:265`, both call sites at `:321` and `:370` — all between `:195` and the next job at `:412`. **No new job or check name was added anywhere in the diff.** All three of `docs/rulesets/{main,beta,release}.json` require the context `Train rules`. |
| The vacuity trap | **Met** | See the caveat below. |
| ADR 0023 D3 amended in-PR | **Met** | The original clause is left intact with an amendment block beneath it, per `docs/adr/README.md`'s amend-don't-rewrite convention. It names the fifth package, recasts the count as *derived* — *"the number written above will go stale again and is not the invariant"* — and records what deliberately did not join the five. The stale "all four" wording still at `0023:21` is the untouched original clause, which is the convention, not an oversight. |
| ADR for D4 written and riding the PR | **Met** | ADR 0025, 43 lines, seven decisions, five rejected options including `@actana/protocol` with the condition to revisit it. It follows the house shape (context → numbered decisions → consequences) and is indexed at `docs/README.md:83`. D2's bounding rule — *"`@actana/core` may import `@actana/sdk/core-link-frames` and nothing else from `@actana/sdk`"* — is the right shape for a reviewer to apply. Honest on the decision itself; see the requested change for the one place it is not honest about the tree. |
| ADR 0016 D3 amended for the contradiction | **Met** | And this is the best of the three. It states the contradiction first — *"`core-link-frames.ts` is no longer in `packages/shared`"* — keeps the clause's argument as the live reason shared was not deleted, tells the reader how to read the now-stale parenthetical, and records that the "boundary is real" half was *extended* rather than weakened by the two Panel modules. That is what `docs/adr/README.md`'s "if a change contradicts one, say so" is asking for. |
| Prototype CLI mirror untouched | **Met** (the sweep is right; the ADR's account of it is not — see above) | |
| `deploy/panel.Dockerfile` | **Met** | `+COPY packages/sdk/package.json packages/sdk/` at `:39` in the lockfile-only layer is the **only** change to the file — one line, nothing else, ordering consistent with its neighbours. It is necessary: `pnpm-lock.yaml` now declares a `packages/sdk` importer and `@actana/panel` declares `"@actana/sdk": "workspace:*"`, so `pnpm install --frozen-lockfile` at `:41` would see three workspace packages against four importers and fail `ERR_PNPM_OUTDATED_LOCKFILE`. I could not execute that half (no pnpm or docker here) — it is reasoned from the lockfile and manifests; the sufficiency half is proven by `Panel image / Build + smoke (amd64)` passing on `55b73f07`. `deploy/core.Dockerfile` correctly needed nothing: it has no manifest-copy install layer. |

### On the vacuity trap, and whether the proof holds

It holds. The caveat is stated correctly and I confirmed it from the run rather than from the claim: the `Train rules` log on `55b73f07` prints `Base: refactor/sdk-extraction` / `Head: refactor/protocol-into-sdk` and takes the `*)` branch — *"No train rules apply"* — so the extended `assert_versions` genuinely could not execute here. That leaves two things carrying it, and both are real:

1. `workflows.test.mjs:501–566` runs under `Unit Tests`, which **is** required by all three rulesets. It asserts the manifest set equals the actual workspace listing (`readdirSync("packages")`) rather than a count, that the assertion lives in the job whose `name:` is the pinned context, that all five manifests and `assert_manifest_set` appear inside that job's block, and that `promote.yml`'s `files=()` array is the same set. Dropping `packages/sdk` from either list fails it. (See non-blocking #2 for the one place it asserts a literal where the PR body claims otherwise.)
2. I extracted the check's bash verbatim into a scratch tree and ran it: 5/5 at `0.2.1` → green; `packages/sdk` at `0.2.0` → red; an unlisted sixth package → red by name; a missing manifest → red by name.

The set guard is the right instinct and is the most valuable thing in this PR after the move itself. A check that silently covers less than the repository has is exactly the failure the promotion checklist's citation-by-name would launder into confidence.

---

## Checklist

- [x] `packages/shared` no longer contains or exports the frames; nothing in the repo imports the old specifier — proven binary-safe over all 942 tracked files
- [x] Core and Panel both import from `@actana/sdk`; the two relative-path holdouts closed
- [x] Suite passes unchanged — verified from the CI logs, not the table
- [x] `CORE_LINK_PROTOCOL_VERSION` byte-identical at `0.15.0`, defined in exactly one file, both packaging scripts repointed
- [x] Exactly one definition of every frame type — 47 symbols, zero duplicated
- [x] The frames file is a move, not a rewrite — one added comment hunk, definitions content-identical; test byte-identical
- [x] The extended assertion lives inside the `Train rules` job the rulesets pin; no new check name
- [x] `workflows.test.mjs` asserts the wiring against the rulesets and the workspace, and runs under a required check
- [x] `promote.yml`'s cut writes five manifests including `packages/sdk`; `packages/sdk` carries `0.2.1`
- [x] ADR 0023 D3 amended; ADR 0016 D3 amended for the contradiction
- [x] ADR 0025 written for D4 — honest on the decision
- [ ] **ADR 0025 honest on the tree** — it places the prototype CLI's mirror in this repository, which the PR body correctly denies
- [x] `deploy/panel.Dockerfile` gained exactly one line, in the right layer, and nothing else changed
- [x] Prototype CLI mirror not touched — and the sweep conclusion that there was nothing to touch is correct

## Verdict

**Requesting changes on one docs clause.** The engineering is done and done well: the move is a true move, the version is untouched to the byte, the importer sweep is complete including the file that hides from `grep`, the fifth manifest is wired into the context that actually gates, and the set guard makes the check fail loudly on the *next* package instead of quietly covering less. Nothing in the code needs to change.

What blocks is that ADR 0025 — a permanent record, in a repository whose own convention is that these are immutable once landed — asserts something about this tree that the PR body takes care to deny. Fix the three lines and this is an approve; I do not need to re-audit anything else.

qcentic-dev:
@qcentic-adm — thanks for the audit. The blocking clause is corrected and all three non-blocking notes are addressed rather than deferred. `55b73f0` → `3dfa70f`, three commits, plain push, still a draft.

> I do not need to re-audit anything else.

Nothing outside the four items you named was touched. The frames file, the version constant, the importer sweep, the manifest set and the Dockerfile line are byte-for-byte what you audited.

---

### Blocking — ADR 0025's account of the mirror (`e17dc68`)

You were right that this was the expensive one: `docs/adr/README.md` makes the record immutable once landed, and line 42 is the bullet phase 2 acts on. All three sites now agree with the sweep.

**:17 (D3)** — *"the repository already knows what that costs: the prototype CLI carries…"* → *"the **project** already knows what that costs: **#129's prototype CLI, which lives outside this repository**, carries…"*

**:32 (rejected option)** — *"the repository has a working example… in the prototype CLI"* → *"the **project** has a working example… in the prototype CLI **outside this repository**"*

**:42 (Consequences)** — the one that sends someone hunting. Now:

> **The prototype CLI's mirror is untouched and is now further out of step.** **It does not live in this repository** — it is a file in #129's prototype CLI, outside this tree, and a sweep of every tracked file here at the time of writing finds no second copy of the frame schema. So there was nothing in this repository for #152 to repair or delete, and a reader of this record should not go looking for one. Freezing the mirror and deleting it is #129 phase 2's work, in the tree that holds it. It is named here so the next reader knows the drift is known and owned rather than missed.

The argument the mirror is being used to make — that a hand-copied schema disagrees at runtime rather than failing at compile time — is untouched, which is what you asked for.

### Non-blocking 1 — `ci.yml` `assert_manifest_set` short-circuits (`653a04a`)

Fixed, not deferred; your diagnosis was exactly right. `[[ "$intact" == "1" ]] && echo …` made the function's exit status the set's verdict, and the bare call at `:267` under `set -euo pipefail` aborted the step before the accumulator at `:406`. Now an `if`, so the function always returns zero and every violation reports through `fail` like the rest of the step. I took the `if` over `assert_manifest_set || true` because `|| true` hides a *genuine* non-zero from anything added to that function later.

Re-ran the step's bash verbatim against a scratch tree, with your `REACHED-END` probe:

| Case | Before | After |
| --- | --- | --- |
| 5/5 at `0.2.1` | green, reaches end | green, reaches end |
| `packages/sdk` at `0.2.0` | red, reaches end | red, reaches end |
| unlisted sixth package | red, **"REACHED-END" never prints** | red, annotates the package, **then compares all five versions and reaches the end** (`failed=1` → step exits 1) |

### Non-blocking 2 — `workflows.test.mjs` asserts a literal on both sides (`3dfa70f`)

Fixed rather than left as a follow-up — it was four lines, and the comment above it already claimed the stronger property, so the cheap thing was to make the code true rather than the comment weaker. The test now pulls `name:` out of the `train-rules` block in `ci.yml` and asserts *that* value is a required context of every gating ruleset. Mutation-tested both directions:

- rename the job in `ci.yml` only → **red** (`beta.json does not require the Branch model rules job`)
- rename it in `ci.yml` **and** all three rulesets → **green**, where the old assertion went red on a correct rename
- move the version check out of the pinned job → still red, via the separate `compares all five manifests inside that job` assertion

### Non-blocking 3 — `scripts/lib/core-tarball.mjs` described as repointed

Corrected in the description, since the code is right and the prose was not: that file holds the regex parser and no path, and its entire diff is the docstring line naming the file it parses. `scripts/build-core-tarball.mjs:239` is the one that holds a path. The PR body now says so.

---

### Deferred

Nothing. All four are in the push.

### CI on `3dfa70f`

**17 pass / 0 fail / 5 skipping.** `Conventions` ✅ · `Train rules` ✅ · `Typecheck` ✅ · `Unit Tests` ✅ · `Lint` ✅ · `Secret Scan` ✅ · `Dependency Audit` ✅ · `E2E — Panel service seam` ✅ · `E2E — installer` (ubuntu, debian) ✅ · `Smoke — Core release tarball` (linux-x64, linux-arm64) ✅ · `Panel image` / `Core image` build + smoke ✅. The five `skipping` are the `*(train)` jobs and the two multi-arch manifest publishes, correct on a non-train base (ADR 0023 D33, D41) and not pending required checks.

`Train rules` is green in 6s on the `*)` branch again — the caveat is unchanged and unchangeable from this base, and the `assert_manifest_set` fix is carried by `Unit Tests` plus the local execution above, same as before.

Locally: root suite 15 files / 416 tests green, `pnpm lint` clean on `scripts`, and all three commit messages lint `found 0 problems, 0 warnings` under this repo's `commitlint.config.mjs`.

Still a draft, base still `refactor/sdk-extraction`, `main` untouched, nothing merged, no force push.

qcentic-adm (approved):
Re-reviewed `55b73f0` → `3dfa70f`. All four items are fixed, verified at the new lines rather than from the commit messages, and the round-1 blocker is resolved. **Approving.**

The round-2 diff is three files, 22 insertions / 9 deletions — exactly the four items and nothing else. All 17 required checks are green on `3dfa70f`, and `Unit Tests` reports the same shape as round 1: `workflows.test.mjs` 39, sdk 1, shared 14, core 43, panel 132.

---

## The blocker — `docs/adr/0025-…md:17`, `:32`, `:42` (`e17dc68`) — **fixed**

Three lines changed, three insertions, three deletions, nothing else in the file. The argument is untouched, which was the point.

- **`:17` (D3)** — "the repository already knows what that costs" → "**the project** already knows what that costs: [#129](…)'s prototype CLI, **which lives outside this repository**, carries a hand-copied 90-line mirror". The link to #129 is a bonus: the next reader gets the tree's name, not just its absence.
- **`:32` (rejected duplicate-the-schema option)** — "the repository has a working example" → "**the project** has a working example … in the prototype CLI **outside this repository**".
- **`:42` (Consequences)** — the expensive one, and now unambiguous: "**It does not live in this repository** — it is a file in #129's prototype CLI, outside this tree, and a sweep of every tracked file here at the time of writing finds no second copy of the frame schema. So there was nothing in this repository for #152 to repair or delete, and a reader of this record should not go looking for one. Freezing the mirror and deleting it is #129 phase 2's work, **in the tree that holds it**." That is the phase-2 engineer's wrong turn closed at the source, and "at the time of writing" is the right hedge for an immutable record.

I re-swept for the failure rather than trusting the three-line diff: every occurrence of "mirror" in `docs/` and `CONTRIBUTING.md` at head, plus the full base→head docs diff. The only remaining mentions in ADR 0025 are `:7` ("the alternative is a mirror" — generic, correct) and the two corrected clauses. No other document in this PR places the mirror in this tree.

## Non-blocking 1 — `.github/workflows/ci.yml:261–276` (`653a04a`) — **fixed, re-executed**

`[[ "$intact" == "1" ]] && echo …` is now an `if`, with a comment that states the reason so the next editor does not re-introduce the `&&`. I extracted the `Enforce the branch model` step's bash verbatim out of `ci.yml` at `3dfa70f`, stubbed `gh`, and ran it against a scratch tree on `BASE=beta/0.2.1` — the same harness as round 1, and I ran the **old** code against the identical tree as a control:

| Case | Old (`&&`) | New (`if`) |
| --- | --- | --- |
| 5/5 at `0.2.1` | green | **green** — "✅ the version set is 5 manifests…", 5 ✅ lines, "✅ Branch model satisfied." |
| unlisted sixth package (`packages/experimental`) | annotation, then **abort** — no version comparison ran | **annotation, then all five versions compared, "❌ The branch model rejects…", exit 1** |
| drift (`packages/sdk` at `0.2.0`) | red | **red** — drift annotated, other four ✅, reaches the accumulator, exit 1 |

That is precisely the behaviour the fix claims, and the control run confirms the old form really did abort where I said it did. The set guard still gates first; it just no longer eats the rest of the step.

**One residual, observation only, no action asked for in this PR.** The other flavour of set violation — a manifest in `MANIFESTS` that does not exist — still short-circuits, one stage later: `assert_manifest_set` now returns 0, the version loop reaches `version="$(jq -r .version "$file")"` on the missing file, and `set -e` kills the step at exit 2 with a raw `jq: error: Could not open file` and no "❌ The branch model rejects" summary. The annotation naming the missing manifest **does** print before the abort, so the actionable message survives, and this is strictly better than the old form (which aborted before it). Nothing in the commit message or the PR body overclaims here — both enumerate the sixth-package case, which is the one I raised. Flagging it only so it is known and owned rather than found later.

## Non-blocking 2 — `scripts/__tests__/workflows.test.mjs:534–552` (`3dfa70f`) — **fixed, mutation-tested both directions**

The literal is gone from the `ci.yml` side: `/^ {4}name: (.+)$/m` pulls the job's `name:` out of the `train-rules` block, a null-guard asserts it exists, and `.trim()`ed it becomes the value every gating ruleset must require. The failure message interpolates it too, so a break reads "`beta.json` does not require the *X* job" rather than naming a string the workflow no longer uses.

I replicated the four tests in that `describe` natively (no vitest here) against a `git archive` of `3dfa70f` and ran six mutations:

| Mutation | Expected | Result |
| --- | --- | --- |
| baseline | pass | **pass** — derives `"Train rules"`; gating rulesets resolve to `beta.json`, `main.json`, `release.json` |
| rename the job in `ci.yml` **only** | fail | **fail** — "beta.json does not require the Version rules job" |
| rename in `ci.yml` **and** all three rulesets | pass | **pass** — the invariant is the relationship, as claimed |
| strip the job's `name:` line | fail, not a crash | **fail** — "the train-rules job has no name:"; the guard fires before the `name[1]` deref |
| drop `packages/sdk` from `ci.yml`'s `MANIFESTS` | fail | **fail** — "packages/sdk/package.json is not asserted" |
| move the manifest set out of the pinned job | fail | **fail** — the in-job assertion catches it |

Both directions hold, the vacuity trap is still caught by the sibling test, and the `4`-space anchor is unambiguous: the job's `name:` sits at four spaces, every step's `- name:` at six. `Unit Tests` passing on `3dfa70f` confirms it under real vitest.

## Non-blocking 3 — the PR body — **fixed**

The `CORE_LINK_PROTOCOL_VERSION` bullet now reads "Of the two packaging scripts involved, exactly one holds a path", names `build-core-tarball.mjs:239` as the repointed one, and says `scripts/lib/core-tarball.mjs` "holds the regex parser and no path — its only change is the docstring naming the file it parses". Checked against the base→head diffs of both files: `build-core-tarball.mjs` changes one `readFileSync` path `packages/shared/…` → `packages/sdk/…`; `core-tarball.mjs` changes one docstring line inside the `parseCoreLinkProtocolVersion` comment and nothing else. The description now matches the diff.

---

## Checklist

- [x] **Blocking — ADR 0025 no longer places the mirror in this repository.** `:17` and `:32` say "the project" and name #129's CLI as outside this tree; `:42` states plainly that it is not here and that phase 2's work happens in the tree that holds it. Re-swept every "mirror" in `docs/` at head — no other misattribution.
- [x] **`ci.yml`'s `assert_manifest_set` reaches the accumulator.** Step bash re-executed verbatim on all three cases, with the old code as a control: 5/5 green; unlisted sixth package annotates, compares all five versions, exits 1 through the accumulator; drift red. (Missing-manifest flavour still aborts at `jq`, one stage later — noted above, not asked for.)
- [x] **`workflows.test.mjs` derives the pinned name from `ci.yml`.** Six mutations run: rename-one-side red, coordinated-rename green, missing `name:` red without a crash, dropped `packages/sdk` red, relocated assertion red.
- [x] **The PR body's packaging-script bullet matches the diff** — one path in `build-core-tarball.mjs:239`, a docstring in `core-tarball.mjs`.
- [x] Round-2 diff is only those three files (22 insertions / 9 deletions); no scope crept in.
- [x] All 17 required checks green on `3dfa70f`; `Train rules` and `Unit Tests` both pass, suite counts unchanged from round 1.

## Verdict

**Approved.** Everything I asked for is done at the lines I asked for, and the two fixes that were behavioural came with evidence I was able to reproduce independently — including the control run proving the old `&&` form failed the way I described. Nothing from round 1's audit needed re-opening; the engineering was already right, and the record now matches the tree it will be committed to.

This is still a draft — marking it ready is yours.

Issue #154, opened 2026-08-11
1.4 Prompt delivery becomes a Core responsibility (D3)


mrfqcentic:
**Base branch: `refactor/sdk-extraction`, not `main`.** Train: **`beta/0.2.1`**.
Worktree: **`wt-129-04`**. Suggested branch: **`feat/core-prompt-delivery`**.
Phase 1, step 2. Implements **D3**.

## What

Replace the flat 450 ms `initialInput` timer at `packages/core/src/pty-manager.ts:619` with a real delivery sequence, owned by the Core:

1. Wait for harness output to go quiet — the TUI has finished painting.
2. Answer any blocking dialog.
3. Write the prompt, then the carriage return **as a separate keystroke**.

Harness differences live in the Core (`CONTEXT.md`), so every client — Panel, CLI, SDK — behaves identically without knowing anything about a harness.

## Done when

- A fresh session never loses its prompt, on every supported harness.
- The folder-trust dialog is answered deliberately, not raced.
- Sessions launched with `--dangerously-skip-permissions` and an initial prompt survive.
- No client sends its own prompt-delivery timing. The 450 ms constant is gone, not tuned.

## D3 says this needs an ADR

Write it in this PR. It rides the code.

## Traps — all three are observed, not hypothetical (`experiment/action.md` §4)

- **The trust dialog's highlighted default is `1. No, exit`.** An early Enter quits the harness before it does anything. Auto mode alone survives; a prompt alone survives; the two together die in under three seconds.
- **Long text is treated as a paste and absorbs the `\r`** — the harness shows `[Pasted text #1 +1 lines]` and waits. The Enter must be a separate keystroke after a pause that scales with the text length.
- **Quiet-detection cannot be a fixed delay in disguise.** Spinner redraws never stop; the signal is a gap in redraws, not silence.

Refs #129

qcentic-dev:
Done — prompt delivery becoming a Core responsibility — landed in PR #191, which squash-merged into the
phase's integration branch `refactor/sdk-extraction`.

That branch has now itself squash-merged into the train `beta/0.2.1` as
**PR #197**, `refactor(sdk): phase 1 — the SDK extraction lands on the 0.2.1 train`.

Closing by hand: the integration PR's merge base is `beta/0.2.1` rather than the
default branch, so GitHub does not auto-close the tickets behind it.

Refs #148, #129.

Issue #153, opened 2026-08-11
1.3 Extract CoreClient: transport and durable entry point (D1, D2, D6, D7)


mrfqcentic:
**Base branch: `refactor/sdk-extraction`, not `main`.** Train: **`beta/0.2.1`**.
Worktree: **`wt-129-03`**. Suggested branch: **`refactor/core-client-extraction`**.
Phase 1, step 2. Implements **D1**, **D2**, **D6**, **D7**.

## What

Extract `packages/panel/src/server/core-link/client.ts` (985 lines) into two entry points on one transport, in `packages/sdk`:

- **`CoreClient`** — connect, authenticate, request/response by `reqId`, close. One socket, one Core, mTLS + bearer from a registration blob. The default, and what the CLI and scripts use.
- **A durable entry point** — heartbeat, reconnect with backoff, event cursor via an injectable `CoreLinkCursorStorage`, subscribe-then-replay. What the Panel will use.

Extraction, not rewrite. The Panel keeps running on its own copy in this ticket; migrating it is a separate one so a regression is attributable.

## Done when

- Both entry points exist, are typed against the SDK's own frames, and are tested against a real Core.
- The three unsolicited streams (`ready`, `data`, `exit`) are surfaced by the transport, not re-parsed by callers.
- The Panel is untouched and still green.

## Docs land here

`CONTEXT.md` gains **Core client** (D7) in this PR. **Not a docs-only PR** — it rides this change.

## Traps

- **`CoreLinkCursorStorage` is browser-shaped in the Panel.** Lifted as-is it drags a DOM dependency into a Node package. Inject the storage; do not import it.
- **`auth` must be the first frame sent**, or the Core answers `not-authenticated` and closes. Anything the transport does on open — a subscribe, a ping — must queue behind it.
- The Core speaks first, unsolicited, with `ready`. A transport that assumes request/response for the first message will hang.
- One socket per Core client (D5's shape). No `Fleet` type in the SDK — the Panel remains the fleet.

Refs #129

qcentic-dev:
Done — `CoreClient` — transport and durable entry point — landed in PR #192, which squash-merged into the
phase's integration branch `refactor/sdk-extraction`.

That branch has now itself squash-merged into the train `beta/0.2.1` as
**PR #197**, `refactor(sdk): phase 1 — the SDK extraction lands on the 0.2.1 train`.

Closing by hand: the integration PR's merge base is `beta/0.2.1` rather than the
[...]

PR #191, opened 2026-08-12
feat(core): the Core delivers a session's starting prompt on the harness's own schedule

qcentic-dev:
## Summary

Replaces the flat 450 ms `initialInput` timer at `packages/core/src/pty-manager.ts` with a real
delivery sequence owned by the Core: wait for the harness output to go quiet, answer any blocking
dialog, write the prompt, then send the carriage return as a **separate** keystroke. The three
constants are gone rather than tuned.

Harness knowledge lives in the Core, so a Panel, the `actana` CLI and an SDK automation all send a
string and nothing else and behave identically. The Panel already did exactly that, so its diff here
is empty — that is the result, not an omission.

ADR 0026 records the decision and rides this PR; `CONTEXT.md` gains the **Prompt delivery** term and
the matching rule.

## Related issues

Refs #154
Refs #129

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## What it actually does

`packages/core/src/harness-prompt-delivery.ts` is fed every chunk the PTY produces and reacts to it.

**1. Quiet is a gap between substantive redraws, not silence.** A spinner never stops, so "wait for
output to end" is a wait that never ends. `redrawSignature` drops escape sequences, the spinner
glyph, and every *run* of digits — so a chunk that only moved a counter (elapsed seconds, a token
meter, a percentage) is a **tick**: proof the harness is alive and no evidence it is still working.
Only a chunk that says something new is a **paint**, and only a paint restarts the quiet window. A
harness that boots in 200 ms waits 200 ms; one that boots in eight seconds gets eight seconds.

**2. A blocking dialog is answered by its own option, never by Enter.** The folder-trust dialog's
highlighted default is `1. No, exit`, so an Enter into it *is* the failure. The Core reads the
numbered menu off the settled screen, finds the option whose **label** means "go ahead", and presses
that number — no hard-coded digit, so a vendor reordering its own menu moves which key we press
rather than inverting the answer. `--dangerously-skip-permissions` opens a second dialog behind the
first with the same dangerous default; each is answered on its own terms, and a stale confirm from
the first can never land on the second.

**3. A dialog it cannot read is not answered at all.** A menu qualifies only if it offers exactly one
affirmative option *and* at least one that refuses. Anything else — an unparseable layout, two
plausible yeses, prose that merely contains the word "trust" — types **nothing**, holds, and abandons
delivery with the dialog still on screen and a log line naming it. A session parked on a visible
dialog is one keystroke from fine; one that answered it wrongly is gone.

**4. The carriage return is its own keystroke.** Sent after a pause that scales with the prompt
length (`150 ms + 2 ms/char`, capped at 3 s) **and** after the echo has stopped repainting — because
what actually has to finish is the harness rendering `[Pasted text #1 +N lines]`.

**5. A backstop that is a ceiling, not a schedule.** 15 s. A prompt delivered a beat early is
recoverable; one never delivered is the bug. The one override: a blocking dialog on screen at the
backstop abandons instead of typing a paragraph into a menu.

Both tables (`BLOCKING_DIALOGS`, `HARNESS_PROMPT_DELIVERY_PROFILES`) are keyed by harness id with a
working default, per CONTEXT.md's "the family is open". The timing table is deliberately empty: no
harness observed so far needs different numbers, and a profile table pre-populated with guesses is
worse than no table.

## How was this tested?

`packages/core/src/__tests__/harness-prompt-delivery.test.ts` — 31 tests. The sequence takes an
injected clock, so the suite asserts **when** each keystroke goes out relative to what the harness
printed, with no sleeping and no flake. The tests that matter:

- *"waits for the painting to stop rather than for a fixed delay"* — nothing is written at the point
  the old 450 ms would have fired.
- *"keeps waiting while the TUI is still painting, however long that takes"* — eight seconds of
  distinct paints, not one byte of prompt written into them.
- *"treats a spinner that never stops as quiet"* — five seconds of unbroken 100 ms spinner redraws
  (real output the whole time); delivery happens, and `waitedMs` proves it did not wait five seconds
  to decide that.
- *"answers the trust dialog by its affirmative option, never by Enter"* — asserts the first write is
  `2` and that `\r` is **not** among the writes.
- *"survives auto mode and an initial prompt together"* — trust dialog, then the bypass-permissions
  warning behind it, then `["2", "2", "ship it", "\r"]`.
- *"types nothing at all into a dialog it cannot read"* — zero writes, phase `abandoned`.
- *"waits longer before submitting a prompt long enough to arrive as a paste"* — the `\r` is a
  separate write, later than a short prompt's.
- *"delivers at the backstop when the harness never settles"* and *"writes nothing after the PTY is
  gone"*.

Full Core suite: **804 passed (44 files)**. `tsc --noEmit` clean; `eslint` clean on the changed files
(the 3 warnings in `pty-manager.ts` are pre-existing and untouched).

Not manually run against a live harness on this machine — the assertions above encode the observed
traces from the experiment rather than re-observing them, and `experiment/action.md` is not present
on this branch. Running a fresh Session on each supported harness from the PR image is the acceptance
step this PR cannot perform for itself.

## Scope

Core and docs only. `packages/sdk` and `packages/panel` are untouched — #153 owns `packages/sdk` in
parallel. The only `CONTEXT.md` edit is the harness-delivery material: one term and one rule.

## Checklist

- [x] This PR targets `refactor/sdk-extraction` (the phase-1 integration branch), not `main`
- [x] My PR title **and every commit in the branch** follow Conventional Commits, and my branch name
      follows the branching convention
- [x] I updated documentation (ADR 0026, CONTEXT.md) — no docs-only PR; it rides the code


qcentic-dev:
## Acceptance criteria — issue #154

Every criterion from the issue, checked or explicitly unchecked with the reason.

### Done when

- [ ] **A fresh session never loses its prompt, on every supported harness.**
      **Still unchecked, and now for one reason only: three of the four harnesses are not
      installed on the machine the run was done on.** `claude-code` is verified live — a real
      `claude 2.1.228` session, `--dangerously-skip-permissions` plus a starting prompt, in a
      folder it had never seen, driven by this branch's own module; the dialog was answered by
      its label with one digit and no Enter, and the prompt was delivered and answered. The
      verbatim timelines are in [this comment](https://github.com/actana/control/pull/191#issuecomment-5266435748).
      `codex`, `cursor-cli` and `opencode` report `NOT INSTALLED` on this machine, so nothing
      is claimed for them. That run also found three further prompt-losing defects that no
      unit test had reached, all fixed in 5a6d1c9 — which is the argument for the box, not
      against it.
- [x] **The folder-trust dialog is answered deliberately, not raced.**
      Nothing is timed. The menu is read off the settled screen, the option whose *label* means "go
      ahead" is selected by its own number, and Enter is never used to get past a dialog. A menu
      that does not offer exactly one affirmative option and at least one that refuses is not
      answered at all. Tests: *"answers the trust dialog by its affirmative option, never by Enter"*
      (asserts `\r` is absent from the writes), *"picks by label even when the affirmative option is
      numbered first"*, *"refuses an ambiguous menu rather than guessing"*, *"types nothing at all
      into a dialog it cannot read"*.
- [x] **Sessions launched with `--dangerously-skip-permissions` and an initial prompt survive.**
      Verified live, twice, on `claude 2.1.228`. Short prompt: dialog answered at 719 ms with
      `1` (`Yes, I trust this folder`, chosen by label), prompt written at 1 620 ms once the
      composer had actually painted, `\r` as its own keystroke at 1 985 ms, model replied
      `391`. Long prompt: 18 lines and 945 characters, rendered by the harness as
      `[Pasted text #1]`, `\r` 2 031 ms after the text and after the 2 040 ms length-scaled
      floor, model replied `1517` — the sentinel being the *last* line, so the whole prompt
      arrived. Timelines: [comment](https://github.com/actana/control/pull/191#issuecomment-5266435748).
      Note for the record: on 2.1.228 no bypass-permissions warning appears behind the trust
      dialog, so that second screen is still covered only at unit level.
- [x] **No client sends its own prompt-delivery timing.**
      Verifiable from the diff: `packages/panel` and `packages/sdk` are untouched. The Panel already
      passed `initialInput` as data only (`pending-initial-input.ts` → `TerminalPane` → `pty.spawn`)
      and still does. The spawn frame carries a string and has no field through which a client could
      express timing, which CONTEXT.md now states as a rule.
- [x] **The 450 ms constant is gone, not tuned.**
      `INITIAL_INPUT_SETTLE_MS`, `INITIAL_INPUT_SUBMIT_DELAY_MS` and `INITIAL_INPUT_MAX_WAIT_MS` are
      all deleted. `grep -rn "450" packages/core/src` returns two prose mentions — a comment
      explaining what was replaced, and a test comment naming the moment nothing may be written yet.
      No number replaced it: the quiet window is a measured gap, and the submit pause is a floor
      that scales with the prompt.

### D3 says this needs an ADR

- [x] **Written in this PR; it rides the code.**
      [`docs/adr/0026-prompt-delivery-is-a-core-responsibility.md`](../blob/feat/core-prompt-delivery/docs/adr/0026-prompt-delivery-is-a-core-responsibility.md)
      — nine decisions (D1–D9), six considered-and-rejected options including passing the prompt on
      the harness command line and putting the sequence in the SDK, and consequences. Next free
      number; the 0018 collision documented in `docs/adr/README.md` is left alone. No docs-only PR.

### Traps — all three observed, not hypothetical

- [x] **The trust dialog's highlighted default is destructive; an early Enter quits the harness.**
      Answered, never raced (see above), and the live run turned this trap into the argument for
      how it is answered: on `claude 2.1.228` the affirmative option is numbered **1** and *is* the
      highlighted default — the opposite of the ordering #154 recorded. The suite now holds both,
      the hand-written fixture with `❯ 1. No, exit` and the real screen captured byte-for-byte,
      and the Core presses `2` on the first and `1` on the second. A hard-coded `2` would have
      pressed `No, exit` on the harness as it ships today; selecting by label is what kept the
      answer right when the menu moved underneath it.

      One Enter is now possible, and only one: the confirm behind a digit, sent only once the
      harness has moved its highlight onto the option that digit chose. If the highlight has not
      moved, delivery holds and the backstop abandons — never an Enter the module cannot justify
      from the screen.
- [x] **Long text is treated as a paste and absorbs the `\r`.**
      The carriage return is a separate write, gated on *both* a length-scaled floor
      (`150 ms + 2 ms/char`, capped at 3 s) and the echo having stopped repainting. Test: *"waits
      longer before submitting a prompt long enough to arrive as a paste"* feeds
      `[Pasted text #1 +12 lines]` back and asserts the `\r` lands after it, later than a short
      prompt's.
- [x] **Quiet-detection is not a fixed delay in disguise.**
      `redrawSignature` drops escape sequences, the spinner glyph and every *run* of digits, so a
      chunk that only moved a counter is a tick and only a new frame is a paint; only a paint
      restarts the window. Tests: *"treats a spinner that never stops as quiet — the gap is in the
      redraws"* (five seconds of unbroken 100 ms redraws; `waitedMs < 1000`), *"keeps waiting while
      the TUI is still painting, however long that takes"* (eight seconds of distinct paints, zero
      writes), *"treats a line where only a counter moved as the same frame"*.

### Boundaries

- [x] Base is `refactor/sdk-extraction`, not `main`. PR is a **draft** and stays one.
- [x] `packages/sdk` and `packages/panel` untouched — #153 owns `packages/sdk` in parallel.
- [x] `CONTEXT.md` touched only for harness-delivery material: the **Prompt delivery** term and one
      rule.
- [x] Conventional Commits on every commit (validated with the repo's own `commitlint.config.mjs`);
      branch pushed with `-u`, never force-pushed; nothing merged.

### Local verification

`pnpm --filter @actana/core test` → **826 passed, 44 files** (the delivery suite went 31 → 53). `tsc --noEmit` clean. `eslint` clean on
the changed files — the 3 warnings in `pty-manager.ts` are pre-existing and untouched.

qcentic-adm (changes requested):
## Review — #154 acceptance, and two defects in the new module

Every criterion in #154 is met, and the shape of the fix is the right one: the module reacts to the harness instead of scheduling around it, and the three constants are gone rather than tuned. I verified the new suite independently — the module extracted from `feat/core-prompt-delivery` with an injected clock, `31 passed`, and `tsc --noEmit` clean under `strict` + `noUncheckedIndexedAccess`. `packages/core/src/__tests__/` holds 44 `.test.ts` files, matching the reported run.

I am requesting changes for two things, both demonstrated against this branch's own module with the same fake clock the suite uses, and both narrow enough to be a small diff.

### 1. `screen` is scrollback, not a screen — a dialog someone else clears is still "on screen"

`harness-prompt-delivery.ts:385` appends every chunk into an 8 KB window, and `stripAnsi` (`:61-63`) discards `ESC[2J` / `ESC[H` rather than treating a full-screen clear as a reset. The window is only emptied after *we* write — `press` (`:522-523`) and `writePrompt` (`:532-533`). So a dialog that goes away without us pressing anything stays matchable at `:448` and `:572` for the next 8 000 characters.

**Repro A — D5's own recovery path does not recover.** Feed the unreadable variant (`"Do you trust the files in this folder?\n\n  Yes / No\n"`), then let the operator answer by hand and the harness paint the ready composer. `matchBlockingDialog` still matches the stale text, `answer` is still `null`, the module holds, and the backstop abandons. Writes: `[]`. ADR 0026 D5 promises "a session parked on a visible dialog is one keystroke from an operator being fine" — after that keystroke the prompt is still never delivered, on a session sitting at a healthy composer. That is the first "Done when" in #154.

**Repro B — a stray digit in the prompt.** Paint `TRUST_DIALOG`, have the operator answer inside the quiet gap, then paint `READY_SCREEN`. At the next quiet the stale menu still parses, so the module presses `2` into the composer and then types the prompt. Writes: `["2", "ship it"]` — the composer holds `2ship it`, and that is what gets submitted.

Fix looks like one branch in `onOutput`: when a chunk carries a full-screen clear, reset `this.screen` (and `recentSignatures`) to that chunk. It also makes the deadline check at `:572` honest.

### 2. The confirm `\r` is sent without checking the highlight moved

`:493-506`, under the comment at `:489-492`: *"the selection is already on the safe option, so this Enter cannot be the one that exits."* Nothing verifies that claim — the branch requires only that the same dialog id is still up. Repaint the **identical** trust dialog after the digit (highlight still `❯ 1. No, exit`, i.e. the harness ignored the digit) and the writes are `["2", "\r"]`. That Enter selects `No, exit`: the three-second death this ADR exists to prevent.

The suite reads as though the guard is there — `harness-prompt-delivery.test.ts:368-378` ("confirms the selection when the dialog stays up, and only then") moves the highlight in its fixture (`:375`), which is the one case where sending the confirm is safe. Claude Code acts on the digit directly, so this path probably never fires there; but the module's whole thesis is that it never sends an Enter it cannot justify, and here it justifies one from an unchecked assumption.

`readDialogOptions` (`:151`, `:160-170`) already sees the highlight marker and drops it. Keeping it — reverse video or `❯`/`>` — and sending the confirm only when the highlighted option *is* the affirmative one, else falling through to the unreadable/hold path, closes it.

### Criteria, one by one

| #154 | Verdict |
| --- | --- |
| 450 ms constant gone, not tuned | ✅ `pty-manager.ts` — `INITIAL_INPUT_SETTLE_MS`, `INITIAL_INPUT_SUBMIT_DELAY_MS`, `INITIAL_INPUT_MAX_WAIT_MS` all deleted; `grep 450 packages/core/src` returns two prose mentions. Nothing took their place: the quiet window is a measured gap and the submit pause is a floor that scales. |
| Quiet keys on a gap in redraws | ✅ `redrawSignature` (`:83-89`) drops escapes, the spinner glyph and every run of digits; only a new signature moves `lastPaintAt` (`:387-393`). *"treats a spinner that never stops as quiet"* runs 5 s of unbroken 100 ms redraws — real output throughout — and settles at 350 ms (`waitedMs < 1000`). *"keeps waiting while the TUI is still painting"* holds off through 8 s of 40 distinct paints with zero writes. Both genuine; I ran them. A moving progress bar still signs as a paint, so the digit rule does not over-collapse. |
| `\r` a separate keystroke after a length-scaled pause | ✅ `writePrompt`/`submit` (`:530-554`) gate the return on **both** `submitPauseMs` (150 ms + 2 ms/char, cap 3 s, `:284-287`) and echo quiet. The paste test (`:328-346`) feeds `[Pasted text #1 +12 lines]` back and asserts `["…", "\r"]` as two writes, the return landing after the marker. I confirmed the return still goes out when a spinner ticks over the paste marker for 6 s. |
| Trust dialog answered deliberately, by label | ✅ `chooseDialogOption` (`:181-191`) needs exactly one affirmative *and* at least one refusal, and presses that option's own number. Given the recorded default is the option numbered `1` meaning `No, exit`, choosing by label rather than by digit is the right conservative call — a hard-coded `2` would invert on a menu reorder, and never answering a menu it cannot read unambiguously is the correct trade for a dialog whose default exits. I checked the real bypass-permissions wording against the table: it is claimed by `bypass-permissions`, answer `2 / "Yes, I accept"` — no cross-match with `folder-trust`. |
| skip-permissions + prompt survives | ⚠️ Covered at unit level and correctly: *"survives auto mode and an initial prompt together"* drives both dialogs and asserts `["2", "2", "ship it", "\r"]`, including that the stale confirm cannot land on the second menu (id check at `:495`). Not run live; the author says so plainly and leaves the box unchecked. Agreed — this is the acceptance step the PR cannot perform for itself. |
| No client sends prompt-delivery timing | ✅ Diff is 5 files: `CONTEXT.md`, ADR 0026, the module, its tests, `pty-manager.ts`. `packages/sdk` and `packages/panel` untouched; the spawn frame's `initialInput` (`packages/sdk/src/core-link-frames.ts:57`) is a bare `string` with no field to express timing. |
| D3 ADR rides the PR | ✅ `docs/adr/0026-…` — D1–D9 plus six considered-and-rejected options, `0026` is the next free number. |
| CONTEXT.md gained only harness-delivery material | ✅ +5 lines: the **Prompt delivery** term and one rule. The rule is scoped to *a starting prompt*, which is accurate — see M3. |

### Minor

- **M1** — both `BLOCKING_DIALOGS` entries are documented as Claude Code's (`:128-133`, ADR D4) but neither sets `harnesses`, so both regexes are applied to `codex`, `cursor-cli` and `opencode` too. Generic-by-default is defensible and matches "the family is open", but it leaves the `harnesses` field dead and means an unrelated harness printing *trust … files … ?* over a numbered menu gets Claude Code's rules. One line either way in the ADR would settle it.
- **M2** — with `quietGapMs: 350`, a harness that paints its ready screen in a single chunk and then goes quiet is written to at 350 ms, i.e. earlier than the 450 ms that was observed to be too early. It is anchored to the last paint rather than to spawn, which is why the observed trace is fine — but it is the one number the live run should sanity-check.
- **M3** (out of scope, follow-up) — `packages/panel/src/lib/harness-question-answer.ts:170-190` still owns harness keystroke timing client-side: `MENU_READY_MS = 1500`, `SUBMIT_CONFIRM_DELAY_MS = 450`, and an Enter into a TUI menu. It is question-answering rather than prompt delivery, so it does not violate the new rule as written, and this PR is right not to touch it — but it is the same knowledge on the wrong side of the wire.

### On delivery

Scope is exact — Core and docs, one commit, draft, based on `refactor/sdk-extraction`. The checklist comment is the honest kind: the two empirical boxes are unchecked with the reason, and the reason is the right one. The injected clock is what makes any of this reviewable, and the ADR argues the rejected options rather than listing them.

Fix the two above and this is an approve. The live run on the four harnesses stays the gate on the two unchecked boxes, and `--dangerously-skip-permissions` with a starting prompt should be the first launch on the image.

qcentic-dev:
## Both defects fixed, and the live run found three more

Thank you — both were real, both reproduced exactly as written, and the live run you set as
the gate turned the two unchecked boxes into three further prompt-losing bugs. Fixed in
5a6d1c9. There were no inline review threads on the PR (`/pulls/191/comments` and the
`reviewThreads` GraphQL connection are both empty), so this replies to the review body
point by point.

### 1. `screen` is scrollback — fixed, and your repros are in the suite

Reproduced first against the pre-fix module, with your fake clock, before changing anything:

```
REPRO A writes: [] phase: answering
REPRO A after backstop: [] phase: abandoned {"phase":"abandoned","reason":"blocked by folder-trust"}
REPRO B writes: ["2"]
```

`onOutput` now resets the buffer at a full-screen clear instead of only after our own
keystrokes. What counts as a clear: `ESC[2J` / `ESC[3J`, the alternate-screen switches,
`ESC[H` followed by `ESC[J`, and — from the live capture — a run of erase-line-and-go-up,
because **Claude Code 2.1.228 never emits ED2 at all**; that run is how it clears. A bare
`ESC[0J` is deliberately excluded: forgetting a dialog that is still up is the worse error.

The signature ring is *not* reset with the screen. It records which frames have been seen
lately, which is a fact about the output rather than about the buffer, and emptying it on
every clear would make a harness that repaints one identical frame behind a clear look like
it was painting forever.

Tests: *"delivers after an operator answers by hand the dialog it could not read"*,
*"presses no menu digit into a composer whose dialog someone else answered"*,
*"holds while the dialog is still on screen behind an unrelated repaint"*, plus a
`lastScreenClearIndex` suite.

### 2. The unjustified confirm — fixed

Pre-fix, your repro: `REPRO C writes: ["2","\r"]`.

`readDialogOptions` now keeps the highlight (reverse video via an SGR containing `7`, or a
pointer glyph) as `DialogOption.highlighted`, and the confirm is gated on `highlightIsOn`:
exactly one option highlighted, and it is the affirmative one. Otherwise the module holds —
`pendingConfirm` stays owed, so a harness that moves its highlight late is still confirmed —
and the backstop abandons with the dialog on screen. New event
`pty.prompt-delivery.dialog-unconfirmed`.

You were right that the fixture was hiding it: `TRUST_DIALOG.replace("❯ 1. No, exit", "  1. No, exit")`
left the `ESC[7m` on option 1, so the highlight had not moved at all. It is now a separate
fixture with the highlight genuinely on option 2, and the identical-repaint case is its own
test asserting no `\r` is ever written.

### M1 — scoped

Both specs now carry `harnesses: ["claude-code"]`, with the reasoning in ADR D7. Both are
transcriptions of that harness's screens; applying them elsewhere means pressing a digit
into another vendor's menu on the strength of the word "trust". A harness with no entry
still gets D3, D6 and D8.

### M2 — 350 ms was too short, and the answer is not a bigger number

You asked for the live run to sanity-check it. It failed it. Measured on 2.1.228: the trust
dialog is answered, the harness acknowledges **in escape sequences only**, and then says
nothing at all for **497 ms** while it loads before painting its composer. A 350 ms gap fires
into that hole — as the 450 ms constant did — and the prompt is typed into a screen that does
not exist yet. From the run that lost it:

```
   643ms  WRITE  "1"
   707ms  OUT(  300)  ""                     <- escapes only: not a paint
   996ms  EVENT  {"phase":"settled","waitedMs":994}
   997ms  WRITE  "What is 17 multiplied by 23? ..."
  1204ms  OUT( 1771)  "╭───Claude Code v2.1.228 ───…"   <- the composer, 207ms too late
  1557ms  WRITE  "\r"
```

The number is unchanged. What changed is that a settle now also requires that something with
content has been painted since our last keystroke — "the painting stopped" presumes it
started. The backstop still overrides, so a harness that draws nothing after a keystroke gets
its prompt rather than losing it (test: *"still delivers at the backstop if the harness draws
nothing after the digit"*).

### Three more the live run found, each fatal on its own

1. **The menu is laid out with column moves, not spaces.** Verbatim from the capture, now a
   fixture:

   ```
   ^[[2G^[[38;5;153m❯^[[4G^[[38;5;246m1.^[[7G^[[38;5;153mYes,^[[12GI^[[14Gtrust^[[20Gthis^[[25Gfolder^[[39m
   ^[[4G^[[38;5;246m2.^[[7G^[[39mNo,^[[11Gexit
   ```

   Deleting every escape yields `1.Yes,Itrustthisfolder` — no menu, no answer, delivery
   abandoned on a healthy session. `stripAnsi` now turns `ESC[<n>G` / `ESC[<n>C` into a space.

2. **The quiet window was measured from the spawn.** The first chunk arrives 368 ms in and
   carries no content, so it is not a paint, so the gap was already "elapsed" when it landed
   and the prompt went straight into the trust dialog. It opens on the first byte now.

3. **`bypass permissions` + `?` matches a healthy composer.** The footer of any session in
   that mode reads `⏵⏵ bypass permissions on (shift+tab to cycle)`, and the placeholder is
   `Try "how do I log an error?"`. The spec now requires `bypass permissions mode`, the
   warning's own wording. Test asserts `matchBlockingDialog` is null against the real
   composer capture.

Worth recording: **the affirmative option in 2.1.228 is numbered `1` and is the highlighted
default** — the opposite of the ordering #154 recorded. Choosing by label rather than by
digit is what made the live run answer `1. Yes, I trust this folder`; a hard-coded `2` would
have pressed `No, exit`. D4 earned itself.

## The live run — verbatim

Real `claude` 2.1.228 (`/home/core/.local/bin/claude`), launched as
`claude --dangerously-skip-permissions` in a folder it had never seen, driven by **this
branch's own `HarnessPromptDelivery`** — the module decided every keystroke; the driver only
spawned the PTY, fed it `onOutput`, and copied `sanitizeInitialInput` from `pty-manager.ts`
so the prompt was exactly what a client would have caused. Every startup marker inherited
from the launching session was stripped from the environment, and the run used an isolated
`CLAUDE_CONFIG_DIR` with no trust entry for the folder and the bypass-accepted flag cleared.

### Run 1 — `--dangerously-skip-permissions` + a starting prompt

```
   351ms  OUT(   13)  ""
   358ms  OUT(   30)  ""
   365ms  OUT( 1043)  "──── Accessing workspace: …"
   366ms  OUT(    8)  ">0q"
   719ms  EVENT  {"phase":"dialog","dialog":"folder-trust","option":1,"label":"Yes, I trust this folder"}
   719ms  WRITE  "1"
   777ms  OUT(  300)  ""
  1243ms  OUT(   47)  ""
  1247ms  OUT(   18)  ""
  1269ms  OUT( 1771)  "╭───Claude Code v2.1.228 ───…"
  1620ms  EVENT  {"phase":"settled","waitedMs":1619}
  1620ms  WRITE  "What is 17 multiplied by 23? Reply with only the number and nothing else."
  1634ms  OUT(  158)  "What is 17 multiplied by 23? Reply with only the number and nothing else."
  1985ms  WRITE  "\r"
  1985ms  EVENT  {"phase":"delivered","waitedMs":1984,"promptChars":73,"submitPauseMs":296}
  2045ms  OUT(  295)  "❯ What is 17 multiplied by 23? … ✢ Herding… · esc to interrupt"
  3638ms  OUT(   42)  "391"
```

The dialog was answered by its label with one digit and no Enter, the prompt was echoed into
the composer, the carriage return went out as its own keystroke, and the model replied
`391` (= 17 × 23). Nothing was raced and nothing was lost.

### Run 2 — a long multi-line prompt, same launch

```
   817ms  EVENT  {"phase":"dialog","dialog":"folder-trust","option":1,"label":"Yes, I trust this folder"}
   817ms  WRITE  "1"
  1393ms  OUT( 1770)  "╭───Claude Code v2.1.228 ───…"
  1744ms  EVENT  {"phase":"settled","waitedMs":1741}
  1744ms  WRITE  "I am checking that a long, multi-line starting prompt survives delivery …"
  1756ms  OUT(  108)  "[Pasted text #1] paste again to expand"
  3787ms  WRITE  "\r"
  3787ms  EVENT  {"phase":"delivered","waitedMs":3783,"promptChars":945,"submitPauseMs":2040}
  5635ms  OUT(   43)  "1517"
```

18 lines in, 945 characters after sanitising. The harness took it as a paste and rendered
`[Pasted text #1]`; the `\r` landed 2 031 ms later — after the marker, and after the
length-scaled floor of 2 040 ms — and the model replied `1517` (= 37 × 41). The whole prompt
arrived: the sentinel was the *last* line of it.

Neither run touched the real `~/.claude.json`: the project entries went to the isolated
config, and the only differences in the real file are this session's own bookkeeping.

### One thing to note, not fixed here

`sanitizeInitialInput` (`pty-manager.ts:244`) drops characters below 0x20 without putting
anything in their place, so a multi-line prompt is joined at the seams: `into a\nfreshly`
was delivered as `into afreshly`. Dropping the newlines is right — each one would submit —
but a space would be righter. Pre-existing, unrelated to this review, and it changes
delivered text, so it wants its own change rather than a quiet ride on this one.

## Still deferred, and why

- **`codex`, `cursor-cli`, `opencode` — evidence-deferred, not claimed.** None of the three
  is installed on this machine:

  ```
  claude       installed: /home/core/.local/bin/claude
  codex        NOT INSTALLED
  cursor-agent NOT INSTALLED
  cursor-cli   NOT INSTALLED
  opencode     NOT INSTALLED
  ```

  So the first "Done when" stays unchecked. What this run does close is the specific claim it
  was gating — `--dangerously-skip-permissions` with a starting prompt, on the harness the
  three observed traps came from — and after the scoping in M1 the other three harnesses now
  get no dialog rules of ours at all, only D3, D6 and D8.
- **The bypass-permissions warning did not appear.** With the accepted-flag cleared and no
  inherited session markers, 2.1.228 showed the folder-trust dialog and nothing behind it;
  `WARNING` and `Bypass Permissions mode` appear nowhere in either capture. The
  two-dialogs-in-sequence path is therefore still only covered at unit level, and its spec is
  written from the wording in #154 rather than from a screen I have seen. Said plainly rather
  than left implied.

## Gate

`pnpm --filter @actana/core test` → **826 passed, 44 files** (was 804; the delivery suite is
31 → 53). `tsc --noEmit` clean. `eslint` clean on the changed files. One commit, Conventional,
validated with the repo's own `commitlint.config.mjs`; pushed with a plain `git push`, never
forced. Still a draft, still based on `refactor/sdk-extraction`, nothing merged.

qcentic-adm (approved):
## Re-review — both defects fixed, verified against `5a6d1c9`

Both are genuinely closed, and I did not take the diff's word for it. I re-ran my two repros against
**both** modules standalone with the same fake clock — the pre-fix module extracted from `c558916`
and the module at `5a6d1c9` — so the before/after is measured rather than argued.

| | pre-fix `c558916` | head `5a6d1c9` |
| --- | --- | --- |
| **Repro A** — operator answers the unreadable dialog by hand | `[]`, `abandoned` | `["ship it", "\r"]`, `delivered` |
| **Repro B** — dialog answered inside the quiet gap | `["2", "ship it"]` → `2ship it` | `["ship it", "\r"]`, no `2` written |
| **Repro C** — identical dialog repainted after the digit | `["2", "\r"]` ← the `No, exit` Enter | `["2"]`, `dialog-unconfirmed`, then `abandoned` |
| confirm when the highlight *did* move | `["2", "\r"]` | `["2", "\r"]` |
| harness that moves its highlight late | `["2", "\r", "2"]` | `["2", "\r"]` |

The last two rows are the ones I care about as much as the fix: the guard closes the unjustified
Enter without costing the justified one, and a harness that confirms late still gets confirmed
because `pendingConfirm` stays owed.

### 1. `screen` is now a screen

`onOutput` (`harness-prompt-delivery.ts:559-563`) resets the buffer at the last clear in the chunk
rather than only after our own keystrokes, and `lastScreenClearIndex` (`:105-112`) is the honest
version of the thing `stripAnsi` throws away. Two judgements in it are right and worth recording:
excluding a bare `ESC[0J` (`:79-82`) keeps the failure on the safe side, and **not** resetting the
signature ring with the screen (`:565-569`) is correct — the ring is a fact about the output, and
clearing it would make a harness repainting one frame behind a clear look like it was painting
forever. `onDeadline` reads the same buffer (`:795`), so the abandon decision is now honest too.

I checked the other direction as well: output that does *not* clear must not lose the dialog. Two
seconds of spinner under an unreadable menu still holds at `answering`, zero writes.

### 2. The confirm is justified or it is not sent

`DialogOption.highlighted` (`:213`), `markHighlights` (`:149-153`) keeping reverse video as a
sentinel through `stripAnsi`, and `highlightIsOn` (`:362-368`) requiring *exactly one* highlighted
row that is the affirmative one. The gate at `:697-715` falls through to `holdUnconfirmed` otherwise.
Requiring exactly one is the right shape — a screen with several selections is a screen the module
cannot read, and that lands in the hold path where it belongs. A harness that highlights with a
background colour leaves no mark and therefore holds, which the comment at `:141-143` says out loud.

**The fixture correction is real, and it was hiding the defect exactly as I said.** The old test fed
`TRUST_DIALOG.replace("❯ 1. No, exit", "  1. No, exit")`, which removed the pointer glyph but left
`ESC[7m` sitting on option 1 — the highlight had not moved at all. It is now a separate
`TRUST_DIALOG_SELECTED` with the highlight genuinely on option 2, and the identical-repaint case is
its own test asserting no `\r` is ever written.

### The tests are regression tests, not restatements

I extracted the module and both fixtures from `5a6d1c9` and ran the suite in isolation:
**53 passed**, matching the reported 31 → 53. Run against the **pre-fix** module: **25 failed, 28
passed** — including all three of tests A, B and C, both `highlightIsOn` tests, and every
real-screen test. They fail before and pass after, which is what makes them worth having.

I did not re-run the full 826-test Core suite; there are no installed workspace deps on this
machine. It also cannot have moved: `git diff c558916 5a6d1c9` is five files — the module, its
tests, two fixtures and the ADR. Nothing outside the module changed.

### The live run — and why I believe it

The two fixtures are the corroboration, because they are checkable independently of the timeline.
Running the **pre-fix** module against those real captures reproduces all three of the newly-claimed
defects, unprompted:

- `matchBlockingDialog(REAL_TRUST)` → `matched=false, options=[]`. The column-move layout really is
  unreadable to the old `stripAnsi`, so a healthy session really would have been abandoned.
- `matchBlockingDialog(REAL_COMPOSER)` → `matched=true`. `bypass permissions` + `?` really did fire
  on a perfectly healthy composer.
- The 497 ms silence after the digit: pre-fix writes the prompt *at the digit*, before any composer
  exists. Post-fix it waits and types only after the composer paints.

On head the same inputs give `answer=1 / "Yes, I trust this folder"`, `highlighted=true`, and no
match on the composer. Those are not things anyone reconstructs from a hand-written fixture, and the
one that matters most is the ordering: the affirmative option is numbered **1** on 2.1.228, the
opposite of what #154 recorded. D4's choose-by-label is what kept that right, and it earned itself.

### M2 — asked for a sanity-check, got a failure and the right fix

The answer to "350 ms is too short" is not 600 ms. `onQuiet` (`:647`) now also requires that
something with content has been painted since our last keystroke, with the backstop overriding
(`:787-802`). That replaces a guess about the size of the hole with a condition on the screen, which
is the same move the whole module is built on. Verified both halves: the 497 ms silence no longer
gets typed into, and a harness that draws *nothing* after the digit still delivers at the backstop
(`["2", "ship it", "\r"]`). `quietGapMs` is still 350 and no number replaced it.

### M1 — scoped, and the wiring actually holds

Both specs carry `harnesses: ["claude-code"]` with the reasoning in D7. The thing that would have
made this a silent disaster is the harness id, so I checked it: `plan.agent` (`pty-manager.ts:630`)
is `HarnessSpawn = Harness` (`pty-spawn-policy.ts:8`), whose members are literally `"claude-code" |
"codex" | "cursor-cli" | "opencode"`. `dialogsForHarness("claude-code")` → both specs;
`("codex")` → none. The scoping does not disable the path the live run proved.

Worth stating plainly as a consequence rather than an objection: the other three harnesses now get
no dialog rules at all, so a trust-style dialog on one of them would have the prompt typed into it.
That is not a regression — the 450 ms path did the same, only sooner — but it does mean their live
runs are the only thing that will ever tell us whether they need entries of their own.

### The deferred harnesses — the deferral is fine; the disclosure is what makes it fine

`NOT INSTALLED` is a machine fact, nothing is claimed for the three, and the first "Done when" stays
unchecked. Correct. I set the live run as the gate and cannot supply hardware, so I am not going to
hold the fix hostage to it.

What I want on the record is the *sharper* limitation, which you volunteered rather than buried:
**the bypass-permissions warning did not appear on 2.1.228.** So Run 1 discharges
`--dangerously-skip-permissions` + a starting prompt on the harness the traps came from, but it does
**not** discharge two-dialogs-in-sequence — the original killer in #154. That path is still unit-only
and its spec is transcribed from #154's wording rather than from a screen anyone has seen. Saying so
unprompted is the reason the rest of the evidence reads as credible.

### `sanitizeInitialInput` — deferring is right

Confirmed pre-existing: the function is byte-identical at the merge-base `62acd54`, so this PR
neither introduced it nor touched it. Reproduced — it joins words *and* list items:

```
IN : "…delivery into a\nfreshly spawned session.\n- item one\n- item two"
OUT: "…delivery into afreshly spawned session.- item one- item two"
```

Defer it, for three reasons. It changes *delivered text*, which is a semantic change that deserves
its own review rather than a ride on a delivery-timing PR. The fix is not the one-liner it looks
like — a space is right for prose and wrong for a code block or a list, and it interacts with the
paste path this PR has just demonstrated is subtle. And `pty-manager.ts` is touched here, so folding
it in would blur what this commit is.

One ask, not a blocker: **file the ticket.** There is no open issue for it today, and the evidence —
your own live capture — currently exists only in a PR comment.

### Nits, neither blocking

- **N1** — D4b lists `ESC[2J`/`ESC[3J`, the alternate-screen switches and home-then-erase-below, but
  not the run of erase-line-and-go-up. That is the one idiom the real harness actually uses, it is
  the reason the fix works on 2.1.228, and the code comments (`:89-94`) explain it. The record should
  name what the code does.
- **N2** — the `{5,}` run has to fall inside a single chunk; a PTY boundary splitting it means no
  reset. It fails safe (hold and abandon, never a digit into a composer), so it is a note rather than
  a change.

### Checklist

- [x] Defect 1 — scrollback-as-screen: fixed at `:559-563` / `:105-112`, verified by running both modules
- [x] Defect 2 — unjustified confirm `\r`: fixed at `:697-715` / `:362-368`, verified; the justified confirm and the late highlight both still work
- [x] Repro A landed as a test, and fails against the pre-fix module
- [x] Repro B landed as a test, and fails against the pre-fix module
- [x] Repro C landed as a test, and fails against the pre-fix module
- [x] The misleading confirm fixture is genuinely corrected, not re-labelled
- [x] Delivery suite run independently at head: 53 passed; 25 fail on the pre-fix module
- [x] No other package could have moved — the fix commit is five files, all within this module and its docs
- [x] M1 scoped, and the harness id it depends on actually matches `plan.agent`
- [x] M2 checked live, failed, and fixed with a condition rather than a bigger constant
- [x] Live-run evidence corroborated independently via the captured fixtures
- [x] `sanitizeInitialInput` confirmed pre-existing; deferral is the right call
- [ ] First "Done when" — the other three harnesses, and two-dialogs-in-sequence on a real screen. Still open, honestly unchecked, and belongs to the run on the image
- [ ] `sanitizeInitialInput` follow-up ticket not yet filed

**Verdict: approve.** Both defects I blocked on are fixed, the fixes are demonstrated rather than
asserted, and the live run turned my one soft concern into three real bugs and closed them. The
unchecked box stays unchecked — that is acceptance work on the image, not a change request against
this branch, and this is a draft into the integration branch. Ship it there.

Issue #156, opened 2026-08-11
1.6 Migrate the Panel onto the SDK — phase 1's success test (D1, D14)


mrfqcentic:
**Base branch: `refactor/sdk-extraction`, not `main`.** Train: **`beta/0.2.1`**.
Worktree: **`wt-129-06`**. Suggested branch: **`refactor/panel-on-sdk`**.
Phase 1, step 4. Implements **D1**. **This ticket is phase 1's success test (D14).**

## What

Delete `packages/panel/src/server/core-link/client.ts` and point the Panel at the SDK's durable entry point.

## Done when

- The Panel behaves identically. Not "equivalently" — identically, from an operator's point of view.
- The existing suite passes **unchanged**. A test edited to accommodate this migration is a failure of the migration.
- The Panel's own copy of the client is gone, not deprecated.
- Reconnect, heartbeat, event cursor and subscribe-then-replay all still work across a Panel reload.

## Traps

- **This is the binary test the whole phase is judged on.** If it goes red, the bug is here or in the extraction — and phase 1 was kept free of new features precisely so that a red test has only two suspects. Do not fix it by adding a feature.
- The Panel's transport is the browser's, not the server's (ADR 0012). Whatever the SDK assumes about where it runs must hold in both.
- #139 (overnight detach) is a live bug in this exact code path. If it is unfixed when this ticket starts, do not absorb it — this migration must be behaviour-preserving including the bug, or the two become inseparable.

Refs #129

qcentic-dev:
Done — the Panel migrating onto the SDK — landed in PR #195, which squash-merged into the
phase's integration branch `refactor/sdk-extraction`.

That branch has now itself squash-merged into the train `beta/0.2.1` as
**PR #197**, `refactor(sdk): phase 1 — the SDK extraction lands on the 0.2.1 train`.

Closing by hand: the integration PR's merge base is `beta/0.2.1` rather than the
default branch, so GitHub does not auto-close the tickets behind it.

Refs #148, #129.

Issue #155, opened 2026-08-11
1.5 SDK session layer: start, prompt, result, programmatic I/O (D2, D11)


mrfqcentic:
**Base branch: `refactor/sdk-extraction`, not `main`.** Train: **`beta/0.2.1`**.
Worktree: **`wt-129-05`**. Suggested branch: **`feat/sdk-session-layer`**.
Phase 1, step 3. Implements **D2** (session level) and **D11**.

## What

The second level of the SDK, on top of the transport: start a session, wait for the harness to be ready, deliver the prompt, read the result.

Programmatic I/O only: `session.send(text)`, `session.onData(…)`, `session.screen()`.

## Done when

- A plain Node script starts a session on a Core, sends a prompt, and reads the result with no terminal involved anywhere.
- `screen()` returns what a terminal would be showing, including lines that have scrolled off the top.
- Depends on the prompt-delivery ticket having landed: the session layer waits on the Core's readiness, it does not reimplement the timing.

## Traps

- **D11 is absolute: no TTY, no raw mode, ever.** Terminal handling belongs to the CLI alone. An SDK that touches `process.stdin` is wrong.
- **`screen()` cannot strip escape codes.** A harness positions text with cursor moves, not spaces; stripping produces one line of concatenated spinner frames. It needs a small terminal emulator — cursor movement, erase, line insert/delete, scroll — and must keep scrolled-off lines, which is where the transcript lives (`experiment/action.md` §4).
- Sessions are spawned against a registered Project. The Core validates the working directory sits inside a known Project root, that argv[0] is the harness's canonical binary, and that every flag is allow-listed. The SDK surfaces those rejections, it does not pre-empt them.

Refs #129

qcentic-dev:
Done — the SDK session layer — `CoreSession` — landed in PR #194, which squash-merged into the
phase's integration branch `refactor/sdk-extraction`.

That branch has now itself squash-merged into the train `beta/0.2.1` as
**PR #197**, `refactor(sdk): phase 1 — the SDK extraction lands on the 0.2.1 train`.

Closing by hand: the integration PR's merge base is `beta/0.2.1` rather than the
default branch, so GitHub does not auto-close the tickets behind it.

Refs #148, #129.

PR #192, opened 2026-08-12
refactor(sdk): CoreClient and a durable Core client on one core-link transport (#153)

### LINKED ISSUE #153: 1.3 Extract CoreClient: transport and durable entry point (D1, D2, D6, D7)

qcentic-dev:
## Summary

Extracts the Panel's `core-link/client.ts` into `@actana/sdk` as **two entry points on one
transport** (#129 D1, D2, D6, D7). `CoreClient` is the default — connect, authenticate,
request/response by `reqId`, close, one socket to one Core, mTLS plus the bearer out of a
registration blob. `DurableCoreClient` is the same client plus what a program that stays up needs:
a heartbeat, reconnection with backoff, an event cursor in an **injected**
`CoreLinkCursorStorage`, and subscribe-then-replay.

Extraction, not rewrite. **The Panel is untouched and keeps running on its own copy** — migrating
it is #156, so a regression there stays attributable. The typed method set here is the Panel
client's, frame for frame, for the same reason.

The layering: `CoreLinkTransport` owns one socket and the three traps that come with this
protocol; `CoreClient` owns the typed frames and the request queue; `DurableCoreClient` owns
reconnection. So the wire is written once, and the difference between the two entry points is
*who re-opens*.

## Related issues

Closes #153

Refs #129, #148

## Type of change

- [x] 📚 Documentation

## The traps, and where each one is answered

- **`CoreLinkCursorStorage` is browser-shaped in the Panel.** It is injected, never imported;
  the default is an in-memory store. A planted global `localStorage` is asserted untouched while
  the cursor advances, so the DOM trap is a test rather than a promise.
- **`auth` must be the first frame sent.** The transport writes it the instant the socket opens
  and reports itself unwritable until `authOk`, so a caller's request — including one issued
  before `connect()` resolved — queues behind it instead of racing it.
- **The Core speaks first, unsolicited, with `ready`.** It is surfaced as `onReady`, and the
  suite asserts it carries no `reqId` and answers nothing the client sent.
- **One socket per client, no `Fleet` type** (D5). There is none, and the Panel remains the fleet.
- `ready`, `data` and `exit` are parsed once in the transport and handed on as frames typed
  against the SDK's own schema. No caller re-parses a raw message.

## How was this tested?

- [x] Manually verified (below)

`pnpm --filter @actana/sdk test` — **88 tests, all green** (85 plus 3 live, see below). Three rigs:

1. **In-process against the real Core.** Every suite drives the Core's own `PtyCoreLinkServer`
   with its bearer gate armed, over a connected socket pair: auth-first ordering, the unsolicited
   `ready`, `reqId` correlation across concurrent requests, error frames becoming coded
   rejections, in-flight requests failing the moment a socket dies, the capability gate on a Core
   that announces nothing, the cursor closing a gap with no double and no hole, PTY subscriptions
   re-asked for on the next connection, and a half-open link destroyed rather than waited on.
2. **Real mTLS over a real port.** `wss://` with material from the Core's own
   `generateCertMaterial`; the negative control asserts on the *shape* of the refusal a Core gives
   a client with no certificate, because a control that passes on any error would pass against a
   Core that is not running.
3. **Against a live Core** (`ACTANA_CORE_BLOB`, skipped without it). Run here against this
   machine's running daemon on `wss://127.0.0.1:9444` with the blob `actana setup` wrote:

   ```
   ✓ src/__tests__/live-core.test.ts > unpacks the live blob into an endpoint, cert material and a bearer
   ✓ src/__tests__/live-core.test.ts > connects, authenticates and answers a read-only request
   ✓ src/__tests__/live-core.test.ts > replays the live event log to a durable client and lands on its cursor
   ```

   Every request in that suite is read-only — nothing spawns, writes, claims or mutates.

Repo-wide: `pnpm -r typecheck` clean, `eslint packages/sdk` clean, `pnpm scan:secrets` clean,
`pnpm install --frozen-lockfile` clean. `@actana/core` 773, `@actana/shared` 180 and `@actana/sdk`
88 all green.

**The Panel suite passes on CI** (`Unit Tests`, which runs the whole workspace). Locally it came
in at 1415/1416 with a *different* single component test timing out each run at ~5.4s against
vitest's 5s default (`terminals-in-browser`, `cores-settings-rename`,
`harness-install-from-picker`), each passing in isolation — this container is slow enough for one
of them to tip over the default timeout. No Panel file is touched: `git diff --stat
origin/refactor/sdk-extraction -- packages/panel` is empty.

## Docs

`CONTEXT.md` gains a **Core client entry points** section (D7): `CoreClient`, **Durable Core
client**, **Core connection**, **Event cursor store**. Additive — no existing entry is touched.
It rides this change rather than a docs-only PR, per #148.

## Notes for review

- `ws` joins `packages/sdk`'s manifest: dialing mTLS is the whole job, and no browser WebSocket
  can present a client certificate (ADR 0002, ADR 0012).
- The registration blob's **shape is declared in the SDK, not imported from `@actana/shared`** —
  that package is private and stays private (ADR 0025 D4), so importing it would give the package
  this effort exists to publish a dependency nobody outside this repository can resolve. A test
  pins that boundary: the shipped sources may import their own modules and `ws`, nothing else.
- A blob unpacks into a `CoreConnection` that **exposes** the HTTPS origin and the cert material
  rather than only dialing with them — #129's phase-3 guardrail met now instead of as a breaking
  change to a published surface later. Nothing in this phase reads it.
- `@actana/core` is reachable from the SDK's suites through a **test-only** alias in
  `vitest.config.ts` / `tsconfig.json`, exactly as the Panel's core-link suites reach it. It is
  deliberately not a manifest entry: the Core depends on this package (ADR 0025 D2).
- `CORE_LINK_PROTOCOL_VERSION` does not move, no version changes, no manifest count changes.

## Checklist

- [x] This PR targets the phase-1 integration branch `refactor/sdk-extraction`, not `main`
- [x] My PR title **and every commit in the branch** follow Conventional Commits, and the branch
      name follows the branching convention
- [x] I added tests proving my change works; all suites pass locally and on CI (one local Panel
      timing flake noted above)
- [x] I updated documentation (`CONTEXT.md`) where relevant
- [x] Draft: #156 migrates the Panel onto this, and is not in this PR

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -95,6 +95,24 @@ _Avoid_: offset, sequence number
 One Core client's standing request for one PTY's byte stream. A Core sends a PTY's `data` and `exit` to the connections that asked for it and to no others, so a client attached to one Session never receives another's output. Held on the connection, so it dies with the socket and is re-asked for on reconnect. Catch-up is the existing `replay { ptyId, sinceSeq }`, ordered behind the subscription rather than in front of it (ADR 0024 D2).
 _Avoid_: channel, stream registration, attach (an attach is the Panel gesture; this is what it asks for)
 
+### Core client entry points
+
+**`CoreClient`**:
+The default way a program becomes a **Core client**: connect, authenticate, ask, close. One socket, one Core, mTLS plus the bearer out of a **Registration blob** — the shape the `actana` CLI and a script want, and the name a third party types against in `@actana/sdk`. It reconnects nothing and remembers nothing between runs; a program that stays up wants the durable entry point below. Frames that only a multi-connection Core understands are withheld until that Core has announced the capability, so one client drives an old Core and a new one without a second code path (ADR 0024 D11, ADR 0025).
+_Avoid_: session, socket wrapper, connection object, SDK consumer
+
+**Durable Core client**:
+The second entry point on the same transport, for a program that stays connected — the Panel, and anything watching a Core rather than asking it one question. `CoreClient` plus a heartbeat, reconnection with backoff, an **Event cursor**, and subscribe-then-replay. On every connection it re-presents its **Core client id**, re-asks for its **PTY subscriptions**, and replays the gap it was away for, so nothing above it has to know the socket ever dropped. There is no fleet here and there must not be: one client is one Core, and holding several is the Panel's job.
+_Avoid_: manager, pool, supervisor, reconnecting socket
+
+**Core connection**:
+What a **Registration blob** unpacks into: the core-link endpoint, that machine's HTTPS origin, the mTLS material, and the bearer. One name because it is one authenticated reach into one machine — the core link and the machine's HTTPS surface are two faces of it, not two connections. A Core client is handed this rather than a URL, and the material is exposed rather than only dialed with, so a surface that needs the same credentials over HTTPS needs no new format.
+_Avoid_: socket, endpoint, credentials bundle
+
+**Event cursor store**:
+Where a **Durable Core client** keeps its **Event cursor** between reconnects and between runs. Always **injected, never imported**: `localStorage` in a browser-shaped Panel, a file for the CLI, memory by default — because the client itself runs in a Node process that has no DOM to reach into. Memory is a real answer rather than a stub: every reconnect still replays its tail, and only a restart starts from the beginning of the log.
+_Avoid_: cache, persistence layer, local storage (that is one implementation of it)
+
 ### Panel views
 
 **Fleet view**:
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@actana/sdk",
   "version": "0.2.1",
-  "description": "Actana Control SDK — the core-link wire protocol a client speaks to a Core: frame schema, protocol version, codec.",
+  "description": "Actana Control SDK — the Core client and the core-link wire protocol it speaks: frame schema, protocol version, codec, transport.",
   "license": "MIT",
   "private": true,
   "type": "module",
@@ -12,8 +12,12 @@
     "typecheck": "tsc --noEmit",
     "test": "vitest run"
   },
+  "dependencies": {
+    "ws": "8.21.0"
+  },
   "devDependencies": {
     "@types/node": "25.8.0",
+    "@types/ws": "8.18.1",
     "typescript": "6.0.3",
     "vitest": "4.1.6"
   }
--- a/packages/sdk/src/__tests__/core-client.test.ts
+++ b/packages/sdk/src/__tests__/core-client.test.ts
@@ -15,7 +15,7 @@ import {
   unwrapResponse,
 } from "../core-client";
 import { coreConnectionFromBlob } from "../core-registration-blob";
-import { startCoreRig, type CoreRig } from "./fake-core-link";
+import { FakeSocketPair, startCoreRig, type CoreRig } from "./fake-core-link";
 
 const SECRET = "core-client-suite-secret-32-bytes-x";
 
@@ -369,6 +369,80 @@ describe("CoreClient", () => {
 
   it("needs somewhere to dial", () => {
     expect(() => new CoreClient({})).toThrow(/url or a registration blob/);
+  });
+
+  it("dials again after a failed connect instead of replaying the rejection", async () => {
+    // `connect()` is memoized so two callers racing share one socket. The memo
+    // must not outlive the attempt: a client that failed once — a Core still
+    // starting up, a laptop that had not joined the network yet — would
+    // otherwise hand every later caller that same rejection for the rest of the
+    // process, with no way back short of building another client.
+    const core = authenticatingRig();
+    const pairs: FakeSocketPair[] = [];
+    const createSocket = () => {
+      const pair = new FakeSocketPair();
+      pairs.push(pair);
+      if (pairs.length === 1) {
+        // Nothing listening yet: the socket dies without ever opening.
+        queueMicrotask(() => pair.client.close());
+      } else {
+        core.wss.accept(pair.server);
+        queueMicrotask(() => pair.open());
+      }
+      return pair.client.asClientSocket();
+    };
+    client = new CoreClient({
+      url: "wss://core.test:9444",
+      bearer: bearerFor(),
+      createSocket,
+    });
+    const c = client;
+
+    await expect(c.connect()).rejects.toThrow(/core-link closed/);
+    await expect(c.connect()).resolves.toMatchObject({ compatible: true });
+
+    expect(pairs).toHaveLength(2);
+    expect(c.isConnected()).toBe(true);
+  });
+
+  it("answers a connect() made on a live link without dialing a second one", async () => {
+    const core = authenticatingRig();
+    const { client: c, dial } = await connected(core);
+
+    await expect(c.connect()).resolves.toMatchObject({ coreId: "core_test" });
+
+    expect(dial.pairs).toHaveLength(1);
+  });
+
+  it("holds a queued caller request behind what the connection owed the Core", async () => {
+    // The loopback shape, and the ordering that only shows up there: with no
+    // bearer the link is writable the instant the socket opens, which can be
+    // *before* the Core has said `ready`. Flushing the queue on writability
+    // alone would put this caller's `replay` on the wire ahead of the `reclaim`
+    // the connection owes — the opposite of what `onConnectionEstablished`
+    // promises, and the same ordering assumption the establishment race had.
+    rig = startCoreRig();
+    const core = rig;
+    const pair = new FakeSocketPair();
+    const createSocket = () => {
+      // Open first; the Core accepts — and so speaks its `ready` — only after.
+      queueMicrotask(() => pair.open());
+      setTimeout(() => core.wss.accept(pair.server), 5);
+      return pair.client.asClientSocket();
+    };
+    client = new CoreClient({ url: "wss://core.test:9444", bearer: null, createSocket });
+    const c = client;
+
+    // Asked for before there is any link at all, so it is queued rather than sent.
+    const pending = c.replay("pty-1");
+    await c.connect();
+    await pending;
+
+    const ordered = pair.client
+      .frames()
+      .map((f) => f.type)
+      .filter((t) => t === "reclaim" || t === "replay");
+    expect(ordered).toEqual(["reclaim", "replay"]);
   });
 
   it("authenticates nothing and is writable at once against a loopback Core", async () => {
--- a/packages/sdk/src/__tests__/core-client.test.ts
+++ b/packages/sdk/src/__tests__/core-client.test.ts
@@ -1,0 +1,384 @@
+// `CoreClient` against a real Core (issue 153, #129 D1/D2/D6).
+//
+// Every test here drives the Core's own `PtyCoreLinkServer` — including its
+// bearer auth gate, which refuses any frame that arrives before `auth` and closes
+// the socket. That gate is why the ordering assertions below are not
+// bookkeeping: get it wrong and a real Core hangs up on you.
+
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { signBearer, verifyBearer } from "@actana/shared/core-link-bearer";
+import { CORE_LINK_PROTOCOL_VERSION, SESSION_LOCKED_ERROR_CODE } from "../core-link-frames";
+import {
+  CoreClient,
+  CoreLinkAuthError,
+  CoreLinkRequestError,
+  unwrapResponse,
+} from "../core-client";
+import { coreConnectionFromBlob } from "../core-registration-blob";
+import { startCoreRig, type CoreRig } from "./fake-core-link";
+
+const SECRET = "core-client-suite-secret-32-bytes-x";
+
+function bearerFor(coreId = "core_test", ttlMs = 60_000): string {
+  return signBearer({ coreId, exp: Date.now() + ttlMs }, SECRET);
+}
+
+describe("CoreClient", () => {
+  let rig: CoreRig | null = null;
+  let client: CoreClient | null = null;
+
+  afterEach(() => {
+    client?.close();
+    client = null;
+    rig?.close();
+    rig = null;
+  });
+
+  /** A Core with the remote shape: mTLS is elsewhere, the auth gate is here. */
+  function authenticatingRig(): CoreRig {
+    rig = startCoreRig({ authVerifier: (bearer) => verifyBearer(bearer, SECRET) });
+    return rig;
+  }
+
+  async function connected(
+    core: CoreRig,
+    opts: { bearer?: string | null } = {},
+  ): Promise<{ client: CoreClient; dial: ReturnType<CoreRig["dialer"]> }> {
+    const dial = core.dialer();
+    const bearer = opts.bearer === undefined ? bearerFor() : opts.bearer;
+    client = new CoreClient({
+      url: "wss://core.test:9444",
+      bearer,
+      createSocket: dial.createSocket,
+    });
+    await client.connect();
+    return { client, dial };
+  }
+
+  it("resolves connect() with what the Core said on ready and authOk", async () => {
+    const core = authenticatingRig();
+    const dial = core.dialer();
+    client = new CoreClient({
+      url: "wss://core.test:9444",
+      bearer: bearerFor("core_abc"),
+      createSocket: dial.createSocket,
+    });
+
+    const info = await client.connect();
+
+    expect(info).toEqual({
+      protocolVersion: CORE_LINK_PROTOCOL_VERSION,
+      compatible: true,
+      multiConnection: { version: 1 },
+      coreId: "core_abc",
+      bearerExpiresAt: expect.any(Number),
+    });
+    expect(client.isAuthenticated()).toBe(true);
+    expect(client.isConnected()).toBe(true);
+  });
+
+  it("sends auth as the first frame on the wire, before anything it was asked to do", async () => {
+    const core = authenticatingRig();
+    const dial = core.dialer();
+    client = new CoreClient({
+      url: "wss://core.test:9444",
+      bearer: bearerFor(),
+      createSocket: dial.createSocket,
+    });
+
+    // A request issued before the socket is even open. A Core answers
+    // `not-authenticated` and closes on a frame that arrives ahead of `auth`, so
+    // this must queue behind it rather than race it.
+    const early = client.replay("pty-1");
+    await client.connect();
+    await expect(early).resolves.toEqual({ data: "scrollback", nextSeq: 7, from: undefined });
+
+    const types = dial.last().client.frames().map((f) => f.type);
+    expect(types[0]).toBe("auth");
+    expect(types.filter((t) => t === "auth")).toHaveLength(1);
+  });
+
+  it("surfaces the Core's unsolicited ready frame — nothing asked for it", async () => {
+    const core = authenticatingRig();
+    const dial = core.dialer();
+    const ready = vi.fn();
+    client = new CoreClient({
+      url: "wss://core.test:9444",
+      bearer: bearerFor(),
+      createSocket: dial.createSocket,
+    });
+    client.onReady(ready);
+    await client.connect();
+
+    expect(ready).toHaveBeenCalledTimes(1);
+    expect(ready.mock.calls[0]![0]).toMatchObject({
+      protocolVersion: CORE_LINK_PROTOCOL_VERSION,
+      compatible: true,
+    });
+    // The client never asked: `ready` carries no reqId and answers no request.
+    const sentBeforeReady = dial.last().server.framesOfType("ready");
+    expect(sentBeforeReady).toHaveLength(1);
+    expect(sentBeforeReady[0]!.reqId).toBeUndefined();
+  });
+
+  it("rejects connect() with the reason when the Core refuses the bearer", async () => {
+    const core = authenticatingRig();
+    const dial = core.dialer();
+    client = new CoreClient({
+      url: "wss://core.test:9444",
+      bearer: signBearer({ coreId: "core_x", exp: Date.now() - 1_000 }, SECRET),
+      createSocket: dial.createSocket,
+      connectTimeoutMs: 2_000,
+    });
+
+    const err = await client.connect().catch((e: unknown) => e);
+    expect(err).toBeInstanceOf(CoreLinkAuthError);
+    expect((err as CoreLinkAuthError).reason).toBe("expired");
+    expect(client.isAuthenticated()).toBe(false);
+  });
+
+  it("correlates concurrent requests by reqId", async () => {
+    const core = authenticatingRig();
+    const { client: c, dial } = await connected(core);
+
+    const [spawned, found, replayed] = await Promise.all([
+      c.spawn({ taskId: "t1", cwd: "/tmp", command: "claude", agent: "claude-code" }),
+      c.findByTask("t1"),
+      c.replay("pty-1", 3),
+    ]);
+
+    expect(spawned).toEqual({ ptyId: "pty-1" });
+    expect(found).toEqual({ ptyId: "pty-1" });
+    expect(replayed).toEqual({ data: "scrollback", nextSeq: 7, from: undefined });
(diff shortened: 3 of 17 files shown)


qcentic-dev:
## Acceptance criteria — #153, item by item

### Done when

- [x] **`CoreClient` exists** — connect, authenticate, request/response by `reqId`, close; one
      socket, one Core, mTLS plus bearer from a registration blob.
      `packages/sdk/src/core-client.ts`, on `core-link-transport.ts`;
      `CoreClient.fromRegistrationBlob(blob)` is the whole of the configuration.
- [x] **A durable entry point exists** — heartbeat, reconnect with backoff, event cursor via an
      injectable `CoreLinkCursorStorage`, subscribe-then-replay.
      `packages/sdk/src/durable-core-client.ts` + `core-link-cursor-storage.ts`.
- [x] **Both are typed against the SDK's own frames.** Every request, response and stream type
      comes from `@actana/sdk/core-link-frames`; the stream surfaces are `Extract<…>` of the frame
      unions (`CoreLinkReadyFrame`, `CoreLinkDataFrame`, `CoreLinkExitFrame`), so a frame shape
      cannot drift from the client that reads it without a type error.
- [x] **Tested against a real Core.** Three rigs: the Core's own `PtyCoreLinkServer` in-process
      with its bearer gate armed; a real `wss://` mTLS handshake on a real port with the Core's own
      cert material (plus a no-client-cert control asserting the *shape* of the refusal); and an
      opt-in suite run here against this machine's **live daemon** on `wss://127.0.0.1:9444` using
      the blob `actana setup` wrote — 3/3 green, read-only. 87 tests, all passing.
- [x] **The three unsolicited streams are surfaced by the transport, not re-parsed by callers.**
      `ready`, `data` and `exit` are parsed once in `CoreLinkTransport.onMessage` and handed on as
      frames. Pinned by tests: `ready` is asserted to carry no `reqId` and to answer nothing the
      client sent, and `data`/`exit` are asserted as whole frames off a real Core's fan-out.
- [x] **The Panel is untouched.** `git diff --stat origin/refactor/sdk-extraction -- packages/panel`
      is empty; so are `packages/core` and `packages/shared`. The Panel still runs on its own copy,
      and migrating it is #156.
- [ ] **The Panel is still green** — *unchecked deliberately, and not because of this change.*
      1415/1416 pass. Across three full runs a **different** single component test timed out each
      time at ~5.4s against vitest's 5 s default (`terminals-in-browser`,
      `cores-settings-rename`, `harness-install-from-picker`), and **each passes in isolation on
      this branch**. No Panel file is modified, so the cause is this container's speed rather than
      the diff — but "green" is a claim about a suite I did not see go green, so it stays unchecked
      until CI says otherwise on quieter hardware.

### Docs

- [x] **`CONTEXT.md` gains a Core client section (D7)** — “Core client entry points”, with
      `CoreClient`, **Durable Core client**, **Core connection** and **Event cursor store**.
      Additive: no existing entry is edited.
- [x] **Not a docs-only PR.** The section rides the extraction, per #148.

### Traps

- [x] **`CoreLinkCursorStorage` is injected, never imported.** Declared in the SDK, defaulting to
      an in-memory store; a test plants a global `localStorage` and asserts it is never read or
      written while the cursor advances. No DOM dependency reaches this Node package.
- [x] **`auth` is the first frame sent.** The transport writes it the instant the socket opens and
      reports itself unwritable until `authOk`, so a subscribe, a reclaim or a caller's request
      queues behind it. Pinned by a test that issues a request *before* the socket opens and then
      asserts frame one on the wire is `auth`, exactly once.
- [x] **The Core speaks first with `ready`, and nothing waits on a response for it.** Surfaced as
      `onReady`; `connect()` resolves on it (plus `authOk` when a bearer is configured), never on a
      correlated answer.
- [x] **One socket per client, and no `Fleet` type in the SDK** (D5). One transport per client, and
      no fleet abstraction anywhere in the package — the Panel remains the fleet.

### Decisions implemented

- [x] **D1** — the SDK is a Core client: one socket, one Core, mTLS + bearer from a registration
      blob; the Panel's client becomes this by extraction, not rewrite (its typed method set is
      carried over frame for frame).
- [x] **D2** — layered, not thin: transport (connect, auth, `reqId` correlation, typed frame
      methods, the three unsolicited streams). The *session* level of D2 is #155 and is not here.
- [x] **D6** — two entry points, one transport, one-shot is the default.
- [x] **D7** — exported class named `CoreClient`; the domain term lands in `CONTEXT.md`.

### Boundaries this ticket was given

- [x] `packages/core` untouched (#154 owns it) — empty diff.
- [x] No existing `packages/panel` file modified — empty diff.
- [x] `CONTEXT.md`: only the new section added.
- [x] Conventional Commits on all five commits; `commitlint --from 62acd54 --to HEAD` clean.
- [x] Branch cut from `origin/refactor/sdk-extraction` @ `62acd54`; base has not moved, so no merge
      was needed. Pushed plain, never force-pushed.
- [x] Draft PR against `refactor/sdk-extraction`, not `main`; nothing merged, `main` untouched,
      PR left as a draft. Title is 85 characters.
- [x] `CORE_LINK_PROTOCOL_VERSION` does not move; no manifest version changes (the five-manifest
      `Train rules` assertion is unaffected).

qcentic-dev:
## Checklist update — the one unchecked box is now checked

- [x] **The Panel is still green.** The `Unit Tests` check passes on CI, which runs the whole
      workspace suite — root scripts, `@actana/core`, `@actana/shared`, `@actana/sdk` and
      `@actana/panel`. That settles the box left open above: the single rotating ~5.4 s timeout
      seen locally was this container being slow against vitest's 5 s default (a *different* test
      each run, each passing in isolation), not this diff — which touches no Panel file.

Everything else in the checklist above stands as posted.

### One commit since that checklist, and why

`fix(sdk): establish a connection on whichever of ready and writability lands last`

Self-review turned up an ordering assumption. A connection is established when two things are
true — the Core has said who it is (`ready`) and the link can be written to — and which lands
last is not fixed. With a bearer it is `authOk`, well after `ready`. Without one, on the trusted
loopback shape, writability rides the socket opening and `ready` normally follows; but a Core
whose frame got in first would have had the connection's opening frames written against a socket
that was not writable yet, and a `subscribe` dropped there is a client that never receives an
event again.

Establishment now waits for both signals in either order. The test that pins it fails without
the change — a loopback durable client whose `ready` is delivered before its socket opens
receives no event at all — so the fix is a fix rather than a tidy-up. 88 tests in
`@actana/sdk` (85 plus 3 live).

qcentic-adm (approved):
## Review — #153, criterion by criterion

Reviewed `62acd54..c2ee8bb` (6 commits, +3726/−2, 17 files). I checked each claim against the
enforcing line rather than the checklist, and spot-compared the transport against
`packages/panel/src/server/core-link/client.ts`.

**Approving.** Every acceptance criterion in #153 is met, the two claims that are easiest to
assert and hardest to earn — "extraction, not rewrite" and "the cursor store is injected" — are
both provable from the diff, and the self-found race fix is pinned by a test that genuinely fails
without it. Four non-blocking notes at the end.

### Done when

- **`CoreClient` exists — connect, authenticate, request/response by `reqId`, close.** Met.
  `packages/sdk/src/core-client.ts:242`; `connect()` at :348 settles on `ready` + `authOk`,
  correlation is the transport's (`core-link-transport.ts:267`), `close()` at :532 rejects
  everything outstanding. `fromRegistrationBlob` at :327 is the whole configuration.
- **A durable entry point — heartbeat, reconnect with backoff, injectable cursor,
  subscribe-then-replay.** Met. `durable-core-client.ts:70`; backoff 500ms→5s at :282,
  `onConnectionEstablished` at :148 sends reclaim → subscribe → PTY re-subscribe in that order,
  cursor dedupe at :220, marker-is-authoritative at :232.
- **Both typed against the SDK's own frames.** Met. Stream surfaces are `Extract<…>` of the frame
  unions (`core-link-transport.ts:41-47`), so a frame shape cannot drift from its reader without a
  type error.
- **Tested against a real Core.** Met. `__tests__/fake-core-link.ts:260` stands up the Core's own
  `PtyCoreLinkServer` with its bearer gate armed — not a stand-in — and `core-link-mtls.test.ts`
  runs a real `wss://` handshake on a real port with material from the Core's own
  `generateCertMaterial`. CI: `@actana/sdk` 85 passed / 3 skipped (88).
- **The three unsolicited streams are surfaced by the transport, not re-parsed by callers.** Met.
  Parsed once in `CoreLinkTransport.onMessage` (`core-link-transport.ts:338/353/365`) and handed on
  as frames; `CoreClient` only fans them out (`core-client.ts:397-404`). Nothing above JSON-parses
  a raw message.
- **The Panel is untouched and still green.** Met. `git diff --stat 62acd54..HEAD -- packages/panel
  packages/core packages/shared` is empty, and the `Unit Tests` check reports
  `packages/panel  Tests 1416 passed (1416)`. The box you left open is closed by CI.

### Traps

- **`CoreLinkCursorStorage` is injected, never imported.** Met, and proved rather than asserted.
  The interface is declared locally (`core-link-cursor-storage.ts:25`), the default is
  `InMemoryCoreLinkCursorStorage`, and **no DOM dependency can enter**: `packages/sdk/tsconfig.json`
  is `lib: ["ES2022"]` + `types: ["node"]` with no `dom`, the only non-relative import in any
  shipped module is `ws` (`core-link-socket.ts:20`, pinned by `package-boundaries.test.ts:32`), and
  the planted-`localStorage` test at :66 asserts the global is never read or written while the
  cursor advances. Worth noting the design matches the Panel's actual usage: `core-link-manager.ts:327`
  already injects a key-ignoring adapter, so the store rename to `actana.coreLink.lastEventId`
  costs #156 nothing.
- **`auth` is the first frame on the wire, and everything else queues behind it.** Met, at two
  layers. The transport writes `auth` inside the `open` handler before anything else
  (`core-link-transport.ts:195`) and reports `writable === false` until `authOk`
  (:241) — so `request()`/`send()` cannot jump it — while `CoreClient.dispatch` at :594 parks a
  caller's frame in a queue rather than refusing it. `core-client.test.ts:80` issues a `replay`
  *before* `connect()` and then asserts frame one on the wire is `auth`, exactly once. That gate is
  real: `pty-core-link-server.ts:890` answers a pre-auth frame with `not-authenticated` **and closes
  the socket**. This is also strictly tighter than the Panel, which flushes its queue on `open`
  right after `auth` rather than after `authOk` (`client.ts:450`) — a good deviation, and one the
  header documents.
- **The Core speaks first with `ready`; nothing assumes request/response.** Met. `ready` is handled
  as an unsolicited case in the message switch, never through `pending`
  (`core-link-transport.ts:338`), and `core-client.test.ts:101` asserts the frame carries no `reqId`
  and answers nothing the client sent.
- **One socket per client, no `Fleet` type.** Met. `git grep -i fleet packages/sdk/src` returns
  prose only — the two hits in shipped code are the comments at `core-client.ts:4` and
  `durable-core-client.ts:16` saying there must not be one.

### The rest of the ticket

- **`ws` on the SDK manifest.** Justified. mTLS with a pinned client cert is the whole job and no
  browser WebSocket can present one (`core-link-socket.ts:9-18`, ADR 0002/0012); `ws` is imported in
  exactly one module. The lockfile delta is 7 lines and records `specifier: '>=8.21.0'` — which
  matches `pnpm-workspace.yaml`'s `overrides.ws`, exactly as the root, core and panel importers do,
  so it is not a `package.json`/lockfile drift. Every CI job installs with
  `pnpm install --frozen-lockfile` and all of them pass.
- **The self-found race fix.** The test genuinely pins it. `startCoreRig`'s dialer accepts the
  connection synchronously (`fake-core-link.ts:300`) so the Core's `ready` is buffered and delivered
  the moment the transport wires its listener, while `open` fires a microtask later — `ready`
  deterministically lands *before* writability. Pre-`c2ee8bb`, the no-bearer branch established on
  `ready` alone, so `sendSubscribe` → `transport.send` → `writable === false` → **`subscribe`
  silently dropped**, and the client would receive no event ever. The new test at
  `durable-core-client.test.ts:298` asserts `framesOfType("subscribe")` has length 1 and
  `seen === [1]`; both fail on the parent commit. That is a fix, not a tidy-up.
- **Docs and the frozen surface.** `CONTEXT.md` gains "Core client entry points" (`CoreClient`,
  Durable Core client, Core connection, Event cursor store), additive, in house style, no existing
  entry edited. Small precision on the checklist: the **Core client** term itself already existed at
  `62acd54` (CONTEXT.md:41), so what D7 actually gains here is the entry-point vocabulary — still
  the right landing, just not a new term. `CORE_LINK_PROTOCOL_VERSION` stays `0.15.0`,
  `core-link-frames.ts` is byte-identical, and all five manifests stay `0.2.1` (only
  `packages/sdk/package.json` is touched at all: description + `ws`/`@types/ws`). `Train rules`,
  `Typecheck`, `Lint`, `Dependency Audit` and `Secret Scan` all pass.
- **Extraction, not rewrite.** This is the claim I tried hardest to break, and it holds. The typed
  method block (`core-client.ts:810-1037`) and `unwrapResponse` (:1047) are line-for-line the
  Panel's, including the `tasksList` → `{tasks, archivedCount}` unwrap and the `claimResult` /
  `reclaimResult` shapes; `parseEvent`, the heartbeat cadence and its half-open `terminate()`,
  the backoff, the cursor dedupe, the "want, not what the Core holds" PTY set and the
  fire-and-forget `reclaim` all carry across with their reasoning intact. The deliberate deltas —
  an awaitable `connect()`, no resurrection in the one-shot client, writability gated on `authOk`,
  streams surfaced as frames — are each the ticket asking for them, and each is documented at the
  seam where it happens.
- **Live-Core evidence.** Credible and appropriately scoped: `live-core.test.ts` is opt-in on
  `ACTANA_CORE_BLOB`, `describe.skipIf`-guarded, and read-only by inspection — `projectsList`,
  `sessionsList`, `agentsAvailabilityList`, and a durable replay that only reads the cursor. Nothing
  spawns, writes, claims or mutates. Note for the record that CI shows those 3 as **skipped**, so
  the 3/3 green rests on your local run against `wss://127.0.0.1:9444` — which is what an opt-in
  live suite can offer, and the suite is honest about it.

### Non-blocking, for this PR or a follow-up

1. `packages/sdk/src/__tests__/core-link-mtls.test.ts:187` — the negative control asserts
   `/closed|ECONNRESET|EPIPE|socket hang up|ERR_SSL|alert|handshake/i`, but *every* transport
   failure reaches the caller as `core-link closed: …` (`core-client.ts:461`), so an
   `ECONNREFUSED` against a Core that is not running also matches `/closed/`. The comment above it
   claims precisely the property the alternation gives away. `startCore()` does bind first, so the
   test is not vacuous — but dropping `closed|` from the alternation would make the assertion mean
   what it says.
2. `packages/sdk/src/core-client.ts:385-388` — `onWritable` calls `flushQueue()` unconditionally,
   including when `ready` has not landed and the connection is therefore not established. On the
   no-bearer path writability precedes `ready`, so a queued caller request can go out ahead of
   `reclaim`/`subscribe` — which is the opposite of what the comment at :452 promises. Harmless
   today, but it is the same ordering assumption `c2ee8bb` removed one layer up; gating that flush
   on `established` would close it for good.
3. `packages/sdk/src/core-client.ts:350` — `connectPromise` is memoized and never cleared, so once
   a durable client's connect deadline rejects, every later `connect()` hands back that same
   rejected promise even though the supervisor has since reconnected. Worth clearing on failure (or
   documenting loudly) before #156 builds on it.
4. `packages/sdk/src/__tests__/package-boundaries.test.ts:17` — `readdirSync(SRC)` is
   non-recursive and the specifier regex only matches double quotes, so a module added under a
   future `src/<dir>/` (or a single-quoted import) would slip past the guardrail that is this
   package's whole reason for existing. A recursive walk is a two-line change.

Nothing above blocks. Nice piece of work: the layering earns its keep, the trap comments say *why*
rather than *what*, and catching the establishment race in self-review — then pinning it with a
test that fails without the fix — is exactly the standard this refactor needs to hold to.

qcentic-dev:
## The four non-blocking notes — all four cleared

Thanks for the review, and in particular for note 1: the alternation really did give away the
property the comment claimed. Three commits on top of `c2ee8bb`, each with a test that fails
without its fix.

### 1. `core-link-mtls.test.ts` — the no-client-cert control could pass against nothing
`d94dfdb`

`closed` is out of the alternation, which now lives in one named constant (`TLS_REFUSAL`) so what
the control accepts is readable in one place, and `ECONNREFUSED` is asserted *absent* alongside
the existing `timed out`.

Rather than leave the discrimination to a comment, a second test pins it: the same client, the
same valid material, aimed at a port with nothing behind it. Its error — `core-link closed:
connect ECONNREFUSED 127.0.0.1:…` — must be one the control's matcher **rejects**. Put `closed|`
back and that test goes red, which is the check the old assertion could not make on itself.

For the record, the shapes as this container's OpenSSL words them, since they are what the
alternation has to span:

- no client cert → `core-link closed: …SSL routines:ssl3_read_bytes:tlsv13 alert certificate required…`
- wrong CA → `core-link closed: certificate signature failure`
- nothing listening → `core-link closed: connect ECONNREFUSED 127.0.0.1:37195`

### 2. `core-client.ts:385-388` — `onWritable` flushed the queue before the connection was established
`8167154`

Gated, exactly as suggested: `onWritable` now calls `maybeEstablish()` and flushes only
`if (this.established)` — the flush inside `maybeEstablish` is the one that matters, and it
already runs behind what the connection owed the Core. The guard also covers writability
returning on a connection that is already established.

Two tests, one per entry point, both on the loopback shape where writability genuinely lands
first (the rig opens the socket and lets the Core accept — and so speak `ready` — a beat later):
a `replay` issued before `connect()` must appear on the wire *after* `reclaim` for `CoreClient`,
and after `reclaim` **and** `subscribe` for `DurableCoreClient`. Both fail on the parent commit
with the caller's frame first.

### 3. `core-client.ts:350` — `connectPromise` memoized past its rejection
`8167154`

Settled rather than documented, since #156 builds on it. The memo is dropped whenever it
settles, either way, and `connect()` then answers from the state of the link *now*:

- a live link resolves immediately from `connectionInfo()` — no second socket over a working one;
- no link starts a fresh attempt;
- a `connect()` made mid-backoff waits for the dial already coming rather than racing one of its
  own. That needed a seam, so `CoreClient` grew a `protected reconnectScheduled()` returning
  `false`, which `DurableCoreClient` overrides with its backoff timer. Without it a retry during
  the backoff would open a socket the supervisor then orphans.

Self-review turned up one more of the same kind while doing it: `established` was cleared on the
next `openTransport()` but not when the socket died, so a `connect()` arriving in the gap between
a drop and the reconnect would have been answered out of the connection that had just ended —
`coreId: null`, `compatible: false`, for a Core about to be perfectly reachable. It is cleared in
the close handler now, and pinned by its own test.

Three tests: a one-shot client dials again after a failed connect instead of replaying the
rejection; a durable client whose 1 ms connect deadline expires before its 40 ms backoff resolves
the *next* `connect()` against the connection the supervisor made (and dials no third socket);
and the mid-backoff case above. All three fail on `c2ee8bb`.

### 4. `package-boundaries.test.ts:17` — non-recursive sweep, double quotes only
`2153567`

Both halves fixed: the walk recurses (sorted, so a failure names the same file on every machine)
and the specifier regex takes `"`, `'` and `` ` ``. `__tests__` is skipped deliberately now rather
than by accident, and `*.test.ts` beside shipped code with it — both import what shipped code may
not, through the test-only aliases in `vitest.config.ts`.

A guardrail whose own coverage is invisible is worth as much as its blind spot, so the sweep is
now swept: a fixture tree plants a nested module, a single-quoted specifier, and two files that
must stay exempt, and asserts exactly the two expected offences come back. Revert the recursion
and it fails; revert the quote handling and it fails differently.

### State

`@actana/sdk` is **93 passed / 3 skipped (96)**, up from 88 — the 3 skipped are still the opt-in
live-Core suite. Workspace `typecheck` clean, `lint` unchanged (0 errors; the 53 warnings are
pre-existing and none are in `packages/sdk`). No `packages/panel`, `packages/core` or
`packages/shared` file is touched, `CORE_LINK_PROTOCOL_VERSION` and all five manifests are where
they were, and `CONTEXT.md` needs nothing: the connect memo was never described there, and the
"Core connection" entry still says what it said.

Pushed plain, never force-pushed. Still a draft, still based on `refactor/sdk-extraction`.

qcentic-adm (approved):
Re-reviewed on `8167154`. All four notes are cleared, and the fifth — `established` surviving the socket that set it — was the right thing to have caught while doing note 3: it is the flag the new `connect()` fast path reads, so the fix and the bug arrived together.

Checked against the lines, not the summary:

1. **mTLS control** — `TLS_REFUSAL` drops `closed`, adds `SSL routines`, and `ECONNREFUSED` is asserted absent. The pin earns its keep: I replayed both matchers over the three shapes you quoted, and `core-link closed: connect ECONNREFUSED 127.0.0.1:…` is rejected by the new alternation and accepted by the old one — so restoring `closed|` does turn the new test red, which is the check the control could not make on itself.

2. **Queue flush** — `flushQueue()` now has exactly two call sites: `maybeEstablish`, behind `onConnectionEstablished`, and the `established`-gated one in `onWritable`. With `bearer: null` the transport reports writable at socket open (`core-link-transport.ts:199`, `:241`) and the rig delays `accept` by 5 ms, so both ordering tests genuinely sit in the pre-`ready` window and both go red on the parent.

3. **Connect memo** — dropped in `settleConnect` *and* `failConnect`, with `established` short-circuiting a live link and `reconnectScheduled()` keeping a mid-backoff `connect()` from opening a socket the supervisor would orphan. The override reads `reconnectTimer`, which `scheduleReconnect` nulls before it dials, so the seam closes on its own rather than latching.

4. **Package boundaries** — I ran the recursive sweep over the seven shipped modules at this head (no offences) and replayed the fixture three ways: fixed → both offences in the order asserted; non-recursive → `nested.ts` missing; double-quotes-only → `single.ts` missing. Both properties are load-bearing, exactly as the comment claims.

Two notes for later, neither blocking and neither in scope here:

- The ordering guarantee still has a sibling hole on the *dispatch* path: a `request()` **issued** between writability and `ready` (no-bearer path) goes straight out via `dispatch()` ahead of `reclaim`, since it is never queued. Same property as note 2, different door; pre-existing, and a queue-when-not-established guard in `dispatch` would close it.
- `onClose` fires the `disconnected` listeners *before* `onConnectionClosed()` arms the backoff, so a listener that calls `connect()` synchronously now dials — `transport` is null and `reconnectScheduled()` is not yet true — and the timer then calls `openTransport()` again, overwriting `this.transport` and orphaning the first socket. Unreachable before, because the memo swallowed that call. A `if (this.transport) return` in `scheduleReconnect`, or arming the backoff before the listeners, covers it. Worth having in hand for #156, where the Panel does register disconnect handlers.

Approving on the new head.

Issue #193, opened 2026-08-12
sanitizeInitialInput drops newlines without a separator


qcentic-dev:
`sanitizeInitialInput()` in `packages/core/src/pty-manager.ts:244` strips C0/DEL
control bytes from a programmatic starting prompt and rejoins the surviving
characters with no separator:

```ts
// packages/core/src/pty-manager.ts:244
function sanitizeInitialInput(text: string | undefined): string | undefined {
  if (!text) return undefined;
  const clean = Array.from(text)
    .filter((ch) => {
      const code = ch.charCodeAt(0);
      return code >= 32 && code !== 127;
    })
    .join("")
    .trim();
  return clean || undefined;
}
```

A newline is `0x0A`, so it fails the `code >= 32` test and is dropped like any
other control byte. Because `.join("")` puts nothing in its place, the text on
either side of the newline is welded together, which changes the prompt text
actually delivered to the agent.

Example: a starting prompt of

```
review the diff
then open a PR
```

is delivered as `review the diffthen open a PR` — `diff` and `then` become a
single token `diffthen`. Any multi-line programmatic prompt (Ship / Sync /
Create-PR) is affected.

The intent of the filter is to stop a caller driving TUI keybindings through
stdin, and that intent is sound — a bare CR/LF would submit early, so newlines
should not pass through verbatim. The bug is only that removal leaves no
separator behind; replacing a dropped line break with a space (or another
agreed separator) would preserve the delivered text.

### Provenance

- Pre-existing behaviour: this is not a regression from #191. It was confirmed
  by both the reviewer and the author during review of #191.
- It was deliberately deferred out of #191's scope, which covers the Core
  delivering a session's starting prompt on the harness's own schedule, not the
  sanitiser's separator semantics.
- Related: #154, #191.

qcentic-adm:
Shipped in the **0.3.2** release train. It reached `main` as **6cd15ca** (`chore: 0.3.2 improvement sweep — docs, contracts and repo hygiene`), promoted through **#262**; the work came in on the train as **#251**. Verified against the tree at `origin/main`.

The criterion this issue states — *"replacing a dropped line break with a space (or another agreed separator) would preserve the delivered text"*, without letting a bare CR/LF through verbatim — and where `main` satisfies it:

[...]

PR #194, opened 2026-08-12
feat(sdk): CoreSession — start a Session, the Core delivers the prompt, read the result

### LINKED ISSUE #155: 1.5 SDK session layer: start, prompt, result, programmatic I/O (D2, D11)

qcentic-dev:
## Summary

The SDK's second level, on top of the transport from #192: **start a Session, let the
Core deliver the prompt, read the result** (#129 D2, D11). Programmatic I/O only —
`session.send(text)`, `session.onData(…)`, `session.screen()`.

Two new modules and one runnable proof:

- **`packages/sdk/src/core-session.ts`** — `CoreSession.start(client, {projectId, cwd,
  harness, prompt})` creates the Task, spawns the harness, wires the byte stream and
  waits on the Core's report of the turn.
- **`packages/sdk/src/terminal-screen.ts`** — a small terminal, because `screen()`
  cannot strip escape codes.
- **`packages/sdk/examples/start-a-session.mjs`** — plain `node`, no terminal anywhere.

## Related issues

Closes #155

Refs #129, #148

## Type of change

- [x] 📚 Documentation

## What it does, and what it deliberately refuses to do

The level below this one knows frames. A caller wanting a *Session* had to correlate a
spawn with its byte stream, filter that stream out of every other PTY on the machine,
know that a harness's output is a screen rather than a log, and know how a Core reports
that a harness has finished. This is that, once, so a script is four calls:

```js
const client = CoreClient.fromRegistrationBlob(blob);
await client.connect();
const session = await CoreSession.start(client, { projectId, cwd, harness: "claude-code", prompt });
await session.waitForIdle();
console.log(session.screen());
```

Three things this layer does **not** do, each because the Core does it:

- **It does not time the prompt.** The starting prompt goes over as `initialInput` and
  the Core delivers it on the harness's own schedule — wait for the TUI to stop
  painting, answer the blocking dialog by its own numbered option, write the text, send
  the carriage return separately (ADR 0026, #191). There is no delay, no ready-signal
  and no retry here to disagree with it. A test asserts the spawn frame's *whole key
  set*, so a future `readyDelayMs` that looks helpful fails it rather than shipping.
- **It does not pre-empt the Core's spawn policy.** A Session is spawned against a
  registered Project: the Core checks the working directory resolves inside a known
  Project root, that argv[0] is that harness's canonical binary, and that every flag is
  allow-listed. Those checks read a database and a filesystem on another machine, so a
  copy here would be a guess. `start()` surfaces the rejection — asserted in the unit
  suite and again live against a real Core.
- **It does not decide what "done" means from the bytes.** Idleness is the Core's
  report: the harness's own lifecycle hooks move the Session's status, the change lands
  in the event log, and `waitForIdle()` watches for it. Watching output go quiet is the
  450 ms timer #191 deleted, in a new place.

**D11 is absolute.** Nothing reads `process.stdin`, sets raw mode, opens `/dev/tty` or
asks whether one exists — and a test sweeps every shipped module for all of those, so
the boundary is a red build rather than a promise.

## `screen()` is a terminal, not a `stripAnsi`

A harness positions text with cursor moves, not spaces. Claude Code draws its
folder-trust dialog as `ESC[4G1.ESC[7GYes,ESC[12GI…` and repaints its spinner by walking
the cursor up the screen; deleting the escapes yields `1.Yes,Itrust` on one line of
concatenated spinner frames. So `terminal-screen.ts` is a grid, a cursor, a scroll
region, two buffers and the escape vocabulary a TUI uses to position and erase — cursor
movement, erase display/line/characters, insert/delete lines and characters, scroll, the
alternate screen. The parser carries state between chunks, because an escape split
across two `data` frames is routine.

**Scrolled-off lines are kept**, which is where the transcript lives. Two rules, and
they differ on purpose:

- *An erased screen is erased.* `ED2` wipes in place and adds nothing to the scrollback,
  because erasing is not scrolling. (Claude Code 2.1.228 never emits `ED2`; it clears by
  walking line-erases up the screen, which scrolls nothing either.)
- *The alternate screen keeps its scrolled-off lines*, which a real terminal does not.
  A human scrolls a full-screen app with the app's own keys; a script cannot, and those
  lines are the transcript it was asked to read. Each buffer keeps its own, so a
  full-screen app's output never lands in the shell's history.

One consequence is documented on `screen()` rather than hidden: a full-screen harness
leaves the alternate screen when it quits, so read the screen **before** killing the
Session.

## Tested against a real Core

The daemon installed on the test machine predates #191 — no `HarnessPromptDelivery` in
its bundle — and against it the prompt is lost exactly as that PR describes: the session
sits at `ready` and `screen()` shows an empty input line. So the live run was done
against a Core **built from this branch**, which is the dependency this ticket declares.

A plain `node packages/sdk/examples/start-a-session.mjs`, no terminal involved:

```
connected to core_186064a659d68c6e (core-link 0.15.0)
session t-msq37v00-0a6e5a running on pty pty-msq37v05-7f6s2i
  status → ready
  status → running
  status → finished
settled: finished
────────────────────────────────────────────────────────────
❯ Reply with exactly: ACTANA-SDK-LIVE-OK and nothing else.
● ACTANA-SDK-LIVE-OK
```

`src/__tests__/live-session.test.ts` is that run as an opt-in suite
(`ACTANA_CORE_BLOB` + `ACTANA_LIVE_PROJECT_ID` + `ACTANA_LIVE_CWD`), asserting the
Session was observed `running` — the harness's own turn-start hook, which fires only
once it has *accepted* a prompt — and that the marker is on a rendered screen with no
escape sequences left in it. Unlike `live-core.test.ts` this one writes, so it takes a
scratch Project rather than defaulting to anything.

## Two changes outside the new modules

- **`CoreClient.subscribeEvents()` / `isSubscribedToEvents()`** (additive). A one-shot
  client sends no `subscribe`, and a script waiting for a harness to finish is reading
  an event. `DurableCoreClient` now writes its own subscribe through the same method, so
  the frame has one home.
- **Relative imports carry their `.ts` extension** (`build(sdk)` commit). Node's ESM
  resolver adds no extensions, so `node script.mjs` importing `@actana/sdk/…` failed on
  the first internal import. This is what makes "a plain Node script" literally that.
  `noEmit` is unchanged and every consumer resolves these specifiers through a bundler.

## The two known sharp edges in the merged transport

Left alone on purpose, both earmarked for #156, and neither is tripped here:

- A `request()` issued between writability and ready on the no-bearer path is never
  queued (pre-existing). This layer only issues requests after `await connect()`, which
  resolves after both.
- `onClose` fires disconnect listeners before the backoff arms, so a listener calling
  `connect()` synchronously can double-dial. `CoreSession` registers no disconnect
  listener at all.

## How was this tested?

- [x] Unit tests added/updated — 27 for the terminal, 28 for the session layer
- [x] Manual testing against a live Core (above)
- [x] `pnpm -r typecheck` clean; `pnpm lint` reports 0 errors and nothing in `packages/sdk`
- [x] CI green

One local-only note: `scripts/__tests__/core-launcher.test.mjs` fails in this dev
container over an install-root path, **identically on the base commit** (`9c02185`), and
passes in CI. Environment, not this branch.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -109,6 +109,14 @@ _Avoid_: session, socket wrapper, connection object, SDK consumer
 The second entry point on the same transport, for a program that stays connected — the Panel, and anything watching a Core rather than asking it one question. `CoreClient` plus a heartbeat, reconnection with backoff, an **Event cursor**, and subscribe-then-replay. On every connection it re-presents its **Core client id**, re-asks for its **PTY subscriptions**, and replays the gap it was away for, so nothing above it has to know the socket ever dropped. There is no fleet here and there must not be: one client is one Core, and holding several is the Panel's job.
 _Avoid_: manager, pool, supervisor, reconnecting socket
 
+**`CoreSession`**:
+The SDK's second level, on top of either Core client entry point: start a **Session** in a registered **Project**, let the Core deliver the starting prompt, read the result. Programmatic I/O only — `send(text)` writes exactly those bytes, `onData(…)` streams them, `screen()` returns the rendered screen — and **no TTY, ever**: it never reads `process.stdin`, never sets raw mode, and is usable from a cron job, a CI runner or a web service, which is D11. It owns none of the three things it depends on: **Prompt delivery** is the Core's, so the prompt goes over as text and the Core decides when it lands; the spawn policy is the Core's, so a working directory outside a Project root or a flag off the allow-list is surfaced as a rejection rather than pre-empted; and "the turn is over" is the Core's report on its event log rather than a guess from the byte stream falling quiet. See ADR 0025, ADR 0026.
+_Avoid_: terminal, pty wrapper, agent runner, REPL
+
+**Screen**:
+What a terminal would be showing for a **Session**, plus every line that has scrolled off the top of it. Produced by a small terminal emulator in the SDK — cursor movement, erase, insert/delete line and character, scroll, the alternate screen — because a Harness positions text with cursor moves rather than spaces, so deleting the escape sequences yields one line of concatenated spinner frames instead of a screen. The scrolled-off half is not an extra: a Harness's conversation left the visible rows long ago, so the transcript *is* the scrollback, and a reader of the viewport alone reads a status bar.
+_Avoid_: output, log, buffer, stdout (those name a stream; this is what a stream was painted into)
+
 **Core connection**:
 What a **Registration blob** unpacks into: the core-link endpoint, that machine's HTTPS origin, the mTLS material, and the bearer. One name because it is one authenticated reach into one machine — the core link and the machine's HTTPS surface are two faces of it, not two connections. A Core client is handed this rather than a URL, and the material is exposed rather than only dialed with, so a surface that needs the same credentials over HTTPS needs no new format.
 _Avoid_: socket, endpoint, credentials bundle
--- a/packages/sdk/examples/start-a-session.mjs
+++ b/packages/sdk/examples/start-a-session.mjs
@@ -1,0 +1,80 @@
+// A plain Node script that starts a Session on a Core, sends it a prompt, and
+// reads the result. No terminal is involved anywhere (#129 D11, issue 155).
+//
+// Run it:
+//
+//     ACTANA_CORE_BLOB=~/.config/actana/registration-blob.txt \
+//     ACTANA_PROJECT_ID=p-… \
+//     ACTANA_CWD=/home/op/projects/thing \
+//       node packages/sdk/examples/start-a-session.mjs "summarise this repo"
+//
+// `node`, directly — no bundler, no loader, no TypeScript step. That is the
+// point of the acceptance criterion this file answers: the SDK has to be usable
+// from a cron job, a CI runner and a web service, none of which has a terminal,
+// and two of which have no build step either.
+//
+// Note what is NOT here. Nothing reads `process.stdin`, nothing sets raw mode,
+// nothing asks whether a TTY exists. And nothing waits a fixed number of
+// milliseconds for the harness to be ready before sending the prompt: the
+// prompt is handed to the Core as text, the Core delivers it on the harness's
+// own schedule (ADR 0026), and this script waits for the Core to report the
+// turn over. The script does not know which harness it started, what that
+// harness paints while it boots, or which dialog it opens first — and it does
+// not need to.
+
+import { readFileSync } from "node:fs";
+import { homedir } from "node:os";
+import { CoreClient } from "@actana/sdk/core-client";
+import { CoreSession } from "@actana/sdk/core-session";
+
+/** The registration blob: the base64 paste itself, or a path to the file holding it. */
+function loadBlob() {
+  const raw = process.env.ACTANA_CORE_BLOB?.trim();
+  if (!raw) throw new Error("set ACTANA_CORE_BLOB to a registration blob or a path to one");
+  const source =
+    raw.startsWith("~") || raw.startsWith("/")
+      ? readFileSync(raw.replace(/^~/, homedir()), "utf8")
+      : raw;
+  return JSON.parse(Buffer.from(source.trim(), "base64").toString("utf8"));
+}
+
+const prompt = process.argv[2] ?? "Reply with exactly: hello from the SDK";
+const projectId = process.env.ACTANA_PROJECT_ID;
+const cwd = process.env.ACTANA_CWD;
+if (!projectId || !cwd) {
+  throw new Error("set ACTANA_PROJECT_ID and ACTANA_CWD (a path inside that Project's root)");
+}
+
+const client = CoreClient.fromRegistrationBlob(loadBlob(), { connectTimeoutMs: 15_000 });
+const info = await client.connect();
+console.log(`connected to ${info.coreId} (core-link ${info.protocolVersion})`);
+
+// The Core validates the working directory against its registered Projects, the
+// command's binary against the harness's canonical one, and every flag against
+// its allow-list. A rejection arrives here as a thrown error carrying the Core's
+// own reason — this script does not second-guess any of it in advance.
+const session = await CoreSession.start(client, {
+  projectId,
+  cwd,
+  harness: process.env.ACTANA_HARNESS ?? "claude-code",
+  title: "SDK example session",
+  prompt,
+  dangerouslySkipPermissions: process.env.ACTANA_SKIP_PERMISSIONS === "1",
+});
+console.log(`session ${session.taskId} running on pty ${session.ptyId}`);
+
+session.onStatus((status) => console.log(`  status → ${status}`));
+
+// Waits on the Core's report — the harness's own lifecycle hooks moving the
+// Session's status — not on the output going quiet.
+const idle = await session.waitForIdle({ timeoutMs: Number(process.env.ACTANA_TIMEOUT_MS ?? 300_000) });
+console.log(`settled: ${idle.status}${idle.exited ? ` (process exited ${idle.exitCode})` : ""}`);
+
+// What a terminal would be showing, including everything that scrolled off the
+// top of it — which is where the harness's answer is, several screens back.
+console.log("─".repeat(60));
+console.log(session.screen());
+console.log("─".repeat(60));
+
+await session.kill();
+client.close();
--- a/packages/sdk/src/__tests__/core-session.test.ts
+++ b/packages/sdk/src/__tests__/core-session.test.ts
@@ -1,0 +1,568 @@
+// The session layer against the Core's own server (issue 155).
+//
+// The rig is the one the transport suites use — a real `PtyCoreLinkServer` over
+// a fake socket — so what is asserted below is the agreement with the Core
+// rather than with a stand-in that answers whatever this file expects. What is
+// faked is the PTY manager behind it (there is no harness to run in a unit test)
+// and the Core's task store, which is where a Session's status lives.
+//
+// The properties worth reading this file for:
+//
+//   - the spawn frame carries the prompt as text and **no timing** with it;
+//   - the byte stream is wired before the spawn is answered, so a harness's
+//     banner is in the transcript rather than lost in the round trip;
+//   - `send` writes exactly its argument and appends nothing;
+//   - idleness comes from the Core's status, not from the stream going quiet;
+//   - a Core's rejection is surfaced, never pre-empted;
+//   - nothing in the shipped package touches a terminal (D11).
+
+import { describe, it, expect, afterEach, vi } from "vitest";
+import { readdirSync, readFileSync } from "node:fs";
+import path from "node:path";
+import type {
+  CoreMutationPort,
+  WebSocketServerLike,
+} from "@actana/core/pty-core-link-server";
+import { PtyCoreLinkServer } from "@actana/core/pty-core-link-server";
+import type { PtyCore, PtyCoreEvent } from "@actana/core/pty-manager";
+import type {
+  CoreLinkProjectMutation,
+  CoreLinkProjectSnapshot,
+  CoreLinkSessionSnapshot,
+  CoreLinkTaskMutation,
+  CoreLinkTaskSnapshot,
+} from "../core-link-frames.ts";
+import { CoreClient } from "../core-client.ts";
+import { CoreSession, CoreSessionStartError, HARNESS_LAUNCH_COMMANDS } from "../core-session.ts";
+import {
+  FakeEventLog,
+  FakeSocketPair,
+  FakeSocketServer,
+  makeMockPtyCore,
+} from "./fake-core-link.ts";
+
+const ESC = "\u001B";
+const CSI = `${ESC}[`;
+const PROJECT_ID = "p-1";
+
+/**
+ * The Core's task store, in memory: create a row, patch its status, list the
+ * Sessions — and append the event a patch produces, because that event is the
+ * whole of how a client learns a harness stopped. The real port
+ * (`core-task-writer.ts`) appends exactly these two kinds.
+ */
+class FakeTaskStore implements CoreMutationPort {
+  readonly tasks = new Map<string, CoreLinkTaskSnapshot>();
+  private seq = 0;
+
+  constructor(private readonly eventLog: FakeEventLog) {}
+
+  mutateProject(mutation: CoreLinkProjectMutation): CoreLinkProjectSnapshot | null {
+    void mutation;
+    return null;
+  }
+
+  mutateTask(mutation: CoreLinkTaskMutation): CoreLinkTaskSnapshot | null {
+    if (mutation.op === "create") {
+      if (mutation.projectId !== PROJECT_ID) return null;
+      const taskId = mutation.taskId ?? `t-${++this.seq}`;
+      const task: CoreLinkTaskSnapshot = {
+        taskId,
+        projectId: mutation.projectId,
+        title: mutation.title,
+        titleManuallySet: false,
+        claudeSessionId: null,
+        agent: mutation.agent,
+        status: mutation.status ?? "ready",
+        pinned: false,
+        archived: false,
+        icon: null,
+        updatedAt: 1,
+      };
+      this.tasks.set(taskId, task);
+      this.eventLog.appendEvent("task:created", JSON.stringify({ taskId }), { taskId });
+      return task;
+    }
+    if (mutation.op === "delete") {
+      this.tasks.delete(mutation.taskId);
+      return null;
+    }
+    const existing = this.tasks.get(mutation.taskId);
+    if (!existing) return null;
+    const previousStatus = existing.status;
+    const next: CoreLinkTaskSnapshot = {
+      ...existing,
+      ...(mutation.status ? { status: mutation.status } : {}),
+      ...(mutation.title ? { title: mutation.title } : {}),
+      updatedAt: existing.updatedAt + 1,
+    };
+    this.tasks.set(next.taskId, next);
+    this.eventLog.appendEvent("task:updated", JSON.stringify({ taskId: next.taskId }), {
+      taskId: next.taskId,
+    });
+    if (next.status === "finished" && previousStatus !== "finished") {
+      this.eventLog.appendEvent("session:finished", JSON.stringify({ id: next.taskId }), {
+        taskId: next.taskId,
+      });
+    }
+    return next;
+  }
+
+  listSessions(): CoreLinkSessionSnapshot[] {
+    return [...this.tasks.values()].map((t) => ({
+      taskId: t.taskId,
+      ptyId: null,
+      status: t.status,
+      updatedAt: t.updatedAt,
+    }));
+  }
+
+  /** What the Core's hook pipeline does when a harness reports its lifecycle. */
+  setStatus(taskId: string, status: string): void {
+    this.mutateTask({ op: "update", taskId, status });
+  }
+}
+
+type Rig = {
+  client: CoreClient;
+  ptyCore: PtyCore & { emitEvent: (e: PtyCoreEvent) => void };
+  tasks: FakeTaskStore;
+  eventLog: FakeEventLog;
+  close: () => void;
+};
+
+function startRig(
+  opts: {
+    announceMultiConnection?: boolean;
+    spawn?: PtyCore["spawn"];
+    /**
+     * Called after the server has written a frame, in the same synchronous
+     * turn. Node's `ws` emits several `message` events from one socket read, so
+     * this is how a test reproduces two frames reaching the client before its
+     * microtasks run — the exact window the held-bytes buffer exists for.
+     */
+    afterServerFrame?: (frame: Record<string, unknown>, rig: () => Rig) => void;
+  } = {},
+): Rig {
+  const ptyCore = makeMockPtyCore();
+  if (opts.spawn) ptyCore.spawn = opts.spawn;
+  const eventLog = new FakeEventLog();
+  const tasks = new FakeTaskStore(eventLog);
+  const wss = new FakeSocketServer();
+  const server = new PtyCoreLinkServer(ptyCore, {
+    port: 0,
+    createServer: () => wss as unknown as WebSocketServerLike,
+    eventLog,
+    mutationPort: tasks,
+    liveEventPollMs: 5,
+    ...(opts.announceMultiConnection === undefined
+      ? {}
+      : { announceMultiConnection: opts.announceMultiConnection }),
+  });
+  const client = new CoreClient({
+    url: "wss://core.test:9444",
+    createSocket: () => {
+      const pair = new FakeSocketPair();
+      if (opts.afterServerFrame) {
+        const send = pair.server.send.bind(pair.server);
+        pair.server.send = (data: string) => {
+          send(data);
+          opts.afterServerFrame!(JSON.parse(data) as Record<string, unknown>, () => built);
+        };
+      }
+      wss.accept(pair.server);
+      queueMicrotask(() => pair.open());
+      return pair.client.asClientSocket();
+    },
+  });
+  const built: Rig = { client, ptyCore, tasks, eventLog, close: () => server.close() };
+  return built;
+}
+
+let rig: Rig | null = null;
+let session: CoreSession | null = null;
+
+afterEach(() => {
+  session?.dispose();
+  session = null;
+  rig?.client.close();
+  rig?.close();
+  rig = null;
+});
+
+async function startSession(
+  r: Rig,
+  overrides: Partial<Parameters<typeof CoreSession.start>[1]> = {},
+): Promise<CoreSession> {
+  await r.client.connect();
+  return CoreSession.start(r.client, {
+    projectId: PROJECT_ID,
+    cwd: "/home/op/projects/thing",
+    harness: "claude-code",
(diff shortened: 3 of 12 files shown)


qcentic-dev:
## Acceptance checklist — issue #155

Every criterion the issue states, checked or explicitly not.

### Done when

- [x] **A plain Node script starts a session on a Core, sends a prompt, and reads the
  result with no terminal involved anywhere.**
  `packages/sdk/examples/start-a-session.mjs`, run with plain `node` against a Core built
  from this branch: `ready → running → finished`, and `screen()` returned the harness's
  answer. Output in the PR description. Nothing in the shipped package touches a
  terminal, and a test sweeps for it.

- [x] **`screen()` returns what a terminal would be showing, including lines that have
  scrolled off the top.**
  `terminal-screen.ts` keeps every line that leaves the top of the grid, capped by a
  configurable scrollback; `screen()` returns scrollback-then-viewport and `viewport()`
  returns the visible rows alone. Covered by 26 tests, including the limit, the
  scroll-region case that must *not* produce history, `ED3`, and a resize that pushes
  lines off.
  *One honest caveat, documented on `screen()`:* Claude Code 2.1.228 runs in the
  alternate screen and paints only a viewport — it never sends the earlier lines to the
  terminal at all — so on that harness there is nothing scrolled-off to keep. That is
  what a real terminal shows too. The emulator keeps such lines wherever they exist,
  including on the alternate screen, which a real terminal does not.

- [x] **Depends on the prompt-delivery ticket having landed: the session layer waits on
  the Core's readiness, it does not reimplement the timing.**
  The prompt goes over as `initialInput` and nothing else. There is no delay, timer,
  ready-probe or retry in this layer, and a test asserts the spawn frame's *entire key
  set* so one cannot be added quietly. Verified live against a Core carrying #191; the
  machine's installed daemon predates it and loses the prompt exactly as that PR
  describes, which is the dependency observed from the outside.

### Traps

- [x] **D11 is absolute: no TTY, no raw mode, ever; an SDK that touches `process.stdin`
  is wrong.**
  A test sweeps every shipped `.ts` in the package for `process.stdin`, `process.stdout`,
  `setRawMode`, `/dev/tty`, `isTTY` and `readline`, comments stripped, and fails on any
  hit. Terminal handling stays the CLI's.

- [x] **`screen()` cannot strip escape codes — it needs a small terminal emulator
  (cursor movement, erase, line insert/delete, scroll) that keeps scrolled-off lines.**
  All five are implemented and tested, plus the alternate screen, `DECSTBM`, deferred
  wrap, `ICH`/`DCH`/`ECH`, save/restore cursor, `RIS`, and character widths. The first
  two tests are transcribed from real Claude Code output: the cursor-positioned dialog
  (`   1. Yes, I trust`, not `1.Yes,Itrust`) and the spinner that must render as its last
  frame rather than all of them concatenated.

- [x] **Sessions are spawned against a registered Project; the Core validates the working
  directory, argv[0] and every flag; the SDK surfaces those rejections, it does not
  pre-empt them.**
  No path check, no binary check and no flag check exists in this layer. `start()` wraps
  the Core's own message in a `CoreSessionStartError`. Asserted twice: a unit test that
  the spawn was *attempted* and the Core's reason came back, and a live test that
  `cwd: "/"` is refused by a real Core.

### From the working instructions

- [x] **Programmatic I/O only — `session.send(text)`, `session.onData(…)`,
  `session.screen()`.** All three, plus `onExit`, `onStatus`, `viewport()`, `lines()`,
  `waitForIdle()`, `resize()`, `kill()`, `dispose()`. `send` appends nothing — not even a
  carriage return, because deciding when to press Enter is prompt delivery and that is
  the Core's.
- [x] **Tested against a real Core.** Both a live end-to-end run and an opt-in live suite.
- [x] **The two known sharp edges in the merged transport are neither fixed silently nor
  tripped.** The unqueued pre-ready `request()` is avoided by only issuing requests after
  `await connect()`; the double-dial needs a disconnect listener calling `connect()`, and
  this layer registers no disconnect listener. Both left for #156, called out in the PR
  description.
- [x] **Conventional Commits on every commit.** Five commits, `commitlint --from 9c02185
  --to HEAD` clean, and CI's Conventions job green.
- [x] **Draft PR against `refactor/sdk-extraction`, not `main`; title under 112
  characters.** Draft, base `refactor/sdk-extraction`, title 89 characters. Nothing
  merged, `main` untouched, no force push.

### Not done, and why

- [ ] **Reattaching to a Session started earlier** (`ptySubscribe` + `replay` into the
  screen). Deliberately out of scope: this ticket is start → prompt → result, and
  surviving a dropped link is the durable client's business (#156). The transport already
  carries what it would need.
- [ ] **A live run on a harness other than `claude-code`.** The test machine has only
  Claude Code installed (`codex`, `cursor-cli`, `opencode` all report `missing`). The
  per-harness launch commands are in one small table with `--enable hooks` for `codex`,
  but only the Claude Code path has been exercised against a real harness.

qcentic-adm (approved):
## Review — #194 against issue #155

Reviewed `30d1de1` against base `9c02185` (`refactor/sdk-extraction`), which is the merge-base, so
the diff is the whole branch: 12 files, +2769/−13, all in `packages/sdk` plus one `CONTEXT.md`
block. Read #155 in full and the acceptance-checklist comment first, then checked each claim by
reading the line that enforces it rather than the line that asserts it. **Approving**, with six
notes below, none blocking.

### Acceptance criteria

**Done when — a plain Node script starts a session, sends a prompt, reads the result, no terminal
anywhere. MET.** `examples/start-a-session.mjs` is `node`-runnable for real: the `.ts`-extension
change (`tsconfig.json` `allowImportingTsExtensions`, and every relative specifier rewritten) is what
makes `@actana/sdk/core-client` resolve without a loader, and `exports: {"./*": "./src/*.ts"}` in
`package.json` already pointed at source. The example reads `screen()` at line 76 *before*
`session.kill()` at line 79, which is the ordering its own doc comment on `screen()`
(`core-session.ts:463-468`) says is load-bearing. Live transcript in the PR body.

**Done when — `screen()` returns what a terminal would show, including scrolled-off lines. MET.**
`terminal-screen.ts` is an emulator, not a stripper: the CSI dispatch at `422-514` covers cursor
movement (`A`–`H`/`f`/`G`/`` ` ``/`d`), erase (`J`/`K`/`X`), line insert/delete (`L`/`M`), character
insert/delete (`@`/`P`), scroll (`S`/`T`, `DECSTBM` at `491-503`, IND/NEL/RI at `323-335`), the
alternate screen (`408-419`, `680-697`) and deferred wrap (`530-549`). Scrolled-off lines are kept at
`577-584`, capped at `722-727`, and `screen()`/`viewport()` split scrollback-plus-viewport from
viewport-alone (`245-261`).

The PR-191 trap is not repeated, and I checked the specific shape of it. #191 was caught treating an
append buffer as the screen, so a dialog cleared by Claude Code's erase-line-and-walk-up idiom (it
never emits ED2) stayed "on screen" forever. Here the screen is a grid and `EL` erases in place, so
that idiom drops the old frame — asserted in `terminal-screen.test.ts:130-145` with the same
five-times `2K 1G 1A` run, checking both that the viewport is the repaint and that the earlier
scrollback is untouched. The two related rules are also right and tested: ED2 wipes without
scrolling (`terminal-screen.ts:661-665`, test `109-118`), and a mid-layout scroll region produces no
history (`578`, test `98-107`). The dialog-reading hazard #191 hit is answered structurally by
`viewport()` being a separate accessor documented for exactly that use (`core-session.ts:474-477`).

One honest limit, which the checklist states rather than hides: on `claude-code` 2.1.228 nothing ever
scrolls off, so the scrollback half of this criterion is proven by unit tests (limit, ED3,
resize-shrink, alt-screen) and not by the live run. That is the harness's behaviour, not a gap in
the work.

**Done when — the session layer waits on the Core's readiness and reimplements no timing. MET, and
more strictly than the wording asks.** The prompt goes over as `initialInput` and nothing else
(`core-session.ts:371`). The only `setTimeout` in either new module is the deadline a caller passes
to `waitForIdle` (`518-533`), and its default is deliberately none (`200-209`). Idleness is read off
the Core's event log (`608-663`), never off the byte stream. The guard is
`core-session.test.ts:228-239`, which asserts the *entire key set* of the spawn as it arrives at the
Core's own `PtyCore` — a future `readyDelayMs` fails it. #191 is on the base (`bac4b84`), and the
live suite's own header records the older-daemon failure mode as the dependency observed from
outside.

**Trap — D11 is absolute. MET, absolutely.** I swept the shipped sources myself for `process.stdin`,
`process.stdout`, `process.stderr`, `setRawMode`, `isTTY`, `/dev/tty`, `tty`, `readline`,
`ReadStream`/`WriteStream`. Every hit in the package is prose or the test's own regex:
`core-session.ts:20-21` (module header), `examples/start-a-session.mjs:16` (a comment saying what is
absent), and `core-session.test.ts:561` (the sweep's pattern). Zero API uses. Not one `console.*`
call in shipped `src` either. The boundary test at `553-581` strips comments before matching, so it
would not pass on its own prose.

**Trap — `screen()` cannot strip escape codes.** Covered above. Worth singling out
`terminal-screen.test.ts:17-27`, which is the exact `CSI 4G 1. CSI 7G Yes, …` capture from #191's
live run and asserts `"   1. Yes, I trust"` rather than `1.Yes,Itrust`.

**Trap — the Core's spawn policy is surfaced, not pre-empted. MET.** There is no path check, binary
check or flag check in the layer; `start()` sends the spawn and wraps whatever comes back
(`core-session.ts:359-383`), and `createTask` does the same for an unknown Project (`708-732`).
Proven twice: `core-session.test.ts:283-300` asserts the spawn was *attempted* and the Core's
`cwd-outside-project-roots` reason came back, and `live-session.test.ts:125-144` does it against a
real Core with `cwd: "/"`.

**`send()` appends nothing. MET.** `core-session.ts:426-428` is a bare pass-through to
`CoreClient.write` (`core-client.ts:904-906`), which puts `data` on the frame verbatim.
`core-session.test.ts:385-396` asserts both the exact argument and that there was exactly one write —
so a carriage return could not arrive as a second call either.

**The two #156 sharp edges are neither fixed nor tripped. CONFIRMED.** `CoreSession` registers no
disconnect listener at all — its only subscriptions are `onData`/`onExit`/`onEvent`
(`core-session.ts:328-355`, released at `394`/`563`), so the `onClose`-before-backoff double-dial
cannot be reached from here. Every request it issues (`spawn`, `tasksMutate`, `write`, `resize`,
`kill`, `sessionsList`) happens after the caller's `await connect()`. Note that `subscribeEvents()`
(`core-client.ts`) goes out through `sendNow`, not `rpc`, so the new call added by this PR is not
subject to the unqueued-pre-ready edge either. Neither is silently patched.

**The two unchecked boxes are honest and correctly scoped. AGREED.** Reattach/replay is genuinely
the durable client's business and #156's; the transport already carries `ptySubscribe` + replay, so
nothing here forecloses it. On the untested harnesses, I checked the deferral is narrower than it
looks: `HARNESS_LAUNCH_COMMANDS` (`core-session.ts:74-79`) matches the Core's canonical commands
exactly (`packages/shared/src/harness-cli-config.ts:71-118` — `claude`, `codex`, `cursor-agent`,
`opencode`), `--enable hooks` is allow-listed with `hooks` as its only permitted value
(`packages/shared/src/pty-spawn-policy.ts:196`), and `--yolo`/`--force` require
`dangerouslySkipPermissions` (`197`, `202`), which `core-session.ts:90-95` and `308` only ever set
together. So what is unexercised is those harnesses' own behaviour, not whether their spawn would be
accepted.

**Scope. CLEAN.** `packages/panel` and `packages/core` untouched; all five manifests (root, core,
panel, sdk, shared) untouched; `CORE_LINK_PROTOCOL_VERSION` is `"0.15.0"` at both `9c02185` and this
head, and `core-link-frames.ts` has no diff at all. The only edits outside the two new modules are
the additive `subscribeEvents()`/`isSubscribedToEvents()` pair, `DurableCoreClient.sendSubscribe()`
routed through it, and the `.ts` specifier rewrite. CI: 19 checks green, 4 skipped.

### Notes — none blocking

1. **`core-session.ts:336-343`, `400-402` — the held-exit slot can be taken by another PTY.** The
   data path holds *every* pre-answer frame and filters by `ptyId` after the answer lands
   (`321`, `397-399`). The exit path keeps only the first (`if (!heldExit) heldExit = frame`) and
   cannot filter yet, so a co-tenant PTY exiting inside that microtask window occupies the slot and
   this Session's own exit frame is discarded: `onExit` never fires and `waitForIdle()` loses its
   exit route. Narrow, but the fix is to make it an array like `held` two lines above.

2. **`core-session.ts:629-650` — a dropped status read has no second chance on the paths that lack a
   dedicated event.** `readStatus()` swallows a failed `sessionsList()`, and the comment's "the next
   event asks again" holds only if another event arrives. For `finished` the `session:finished`
   shortcut at `613` is a genuine belt and braces, but `needs-input`, `interrupted` and `terminated`
   reach this layer only as `task:updated` — so one failed read on the transition that mattered
   leaves `waitForIdle()` pending, and it has no default deadline by design. One re-ask, or a word
   in the `waitForIdle` doc that a caller wanting robustness against a flaky link should pass
   `timeoutMs`.

3. **`terminal-screen.ts:205-206` — stale comment.** "only the main buffer's keeps what it loses"
   contradicts `219`, which pushes to whichever buffer is shrinking, and contradicts both the
   header's deliberate rule (`35-42`) and the test that asserts it
   (`terminal-screen.test.ts:166-182`). The code is right; the comment is from an earlier design.

4. **`core-session.test.ts:560-577` — the D11 sweep does not cover `examples/`.** It walks
   `src/**/*.ts` only. The one file in this package a user is invited to run with `node` is
   `examples/start-a-session.mjs`, and it is outside the boundary the test defends. It is clean
   today (I swept it), so this is about keeping it that way.

5. **`terminal-screen.ts:510-512` — SGR is parsed and dropped, so highlight is invisible.** Not a
   defect against #155, which does not ask for attributes, and the `❯` pointer glyph survives as
   text so most menus are still readable. But reverse video is precisely the signal #191's confirm
   gate needed on the Core side, and a caller answering a mid-session dialog through `send()` has no
   equivalent. Worth a line in #156's scope rather than a change here.

6. **Out of scope, but it now sits under an advertised SDK surface.** `sanitizeInitialInput`
   (`packages/core/src/pty-manager.ts:244-254`, base) drops characters below 0x20 without
   substituting anything, so a multi-line string passed to `CoreSessionStartOptions.prompt` arrives
   joined at the seams (`into a\nfreshly` → `into afreshly`). Already recorded on #191 as
   pre-existing and wanting its own change; flagging only because `prompt` is now the documented way
   a script sends one.

Trivia: the description says 26 terminal tests; there are 27.

### Verdict

Every criterion in #155 is met, and the three claims that are easiest to assert and hardest to honour
— no TTY, a real emulator rather than a stripper, no timing of its own — are each enforced by a test
that fails on the tempting mistake rather than by a promise in a comment. The two unchecked boxes are
honestly stated and correctly assigned to #156 and to a machine that lacks the binaries. Approving.

qcentic-dev:
Thanks for the read — all four actionable notes are addressed in `af6ad81`, which is the only
new commit. Nothing about what the layer does changed; what changed is whether it still does it
with a second PTY or a flaky link in the way.

**1. `core-session.ts:336-343`, `400-402` — the held-exit slot.** Fixed as you suggested: `heldExit`
is now `heldExits: CoreLinkExitFrame[]`, every pre-answer exit frame is pushed, and the replay picks
the one whose `ptyId` matches — the same shape as the data path two lines above. Test:
*"keeps its own exit when a co-tenant PTY exits in the same window"*. It spawns two Sessions on one
connection and, in the synchronous window after the second `spawned` frame, emits `pty-1`'s exit
*first* so a one-slot hold is already full when `pty-2`'s own exit arrives behind it. Against the old
code it fails on exactly the symptom you described: `exitStatus()` null and `waitForIdle()` with no
exit route.

**2. `core-session.ts:629-650` — the dropped status read.** Took the re-ask rather than a default
deadline, because a default deadline is the thing this layer is documented not to have and #191
deleted. A failed `sessionsList()` is now re-asked `STATUS_READ_RETRIES` (3) times,
`STATUS_READ_RETRY_MS` (250) apart, then stops; a success resets the budget, an event arriving
mid-read still wins as the fresher re-ask, and `dispose()` clears the pending timer (which is
`unref`'d, so a re-ask never keeps a finished script alive). Worth being explicit that this does not
weaken the module's third rule: it retries a *question* whose answer the Core has already settled,
so it can never report a Session idle sooner than the Core says it is — noted in the header and on
the constant. Two tests: one proving a `needs-input` transition still lands after the read that
carried it failed and no second event ever arrives, one proving the re-asks stop rather than
polling a Core that is gone. Your alternative went in too — `waitForIdle`'s doc now says plainly
that a caller who must not hang on a broken Core passes its own `timeoutMs`.

**3. `terminal-screen.ts:205-206` — stale comment.** Corrected, code untouched. It now says each
buffer keeps what it loses off the top in its own scrollback, the alternate screen included, and
points at the header rule (`35-42`) that the old wording contradicted.

**4. `core-session.test.ts:560-577` — the D11 sweep.** Widened: it walks `src/` **and** `examples/`,
matches `.mjs`/`.js` as well as `.ts`, and reports offences as package-relative paths. Two things
beyond the widening — the sweep now also asserts it *reached* both trees by name, because an empty
walk passes an empty assertion and this one silently walked only `src` until now; and I checked the
positive direction by planting `process.stdin.setRawMode(true)` in `examples/start-a-session.mjs`
and confirming the test fails naming that file. Reverted, of course; the example is clean.

**Left alone, deliberately.** Note 5 (SGR parsed and dropped, so reverse video is invisible) is
#156's scope as you say, and note 6 (`sanitizeInitialInput` joining a multi-line prompt at the
seams) belongs to #193 — neither is touched here.

**Trivia.** Corrected in the description: 27 terminal tests, not 26. The session-layer count moved
too with the three added here, so it now reads 27 and 28.

CI on `af6ad81`: green — 17 checks pass, 5 skipped, none failed (run 31600762018). Still a draft, base still `refactor/sdk-extraction`, nothing merged,
`main` untouched, no force push.

qcentic-adm (approved):
## Re-review — `af6ad81`

One new commit on top of the reviewed `30d1de1` (`fix(sdk): a Session keeps its own exit and re-asks a
status read it lost`), `30d1de1` still an ancestor — no force push, no rebase, base still
`refactor/sdk-extraction`, still a draft. Diff is three files, +174/−14, all in `packages/sdk`; nothing
outside the package moved. I checked each fix at the line that enforces it and then ran the suite twice
— once on this head, once with **this head's test file against the previous source** — so "fails on the
old behaviour" is observed rather than taken on trust.

**Note 1 — the held-exit slot. FIXED.** `heldExits: CoreLinkExitFrame[]` (`core-session.ts:348`), every
pre-answer exit pushed (`368`), and the replay filters by id like the data path does:
`heldExits.find((frame) => frame.ptyId === ptyId)` at `430-431`, against `held`'s own id filter at
`428-429`. The cast the single slot needed is gone with it. The test (`core-session.test.ts:335-361`)
spawns two Sessions on one connection and emits `pty-1`'s exit *before* `pty-2`'s inside the second
spawn's window, so a one-slot hold is already occupied — the ordering that makes it a real reproduction
rather than a restatement. Against `30d1de1` it fails at `:359`, `exitStatus()` null, exactly the
symptom.

**Note 2 — the dropped status read. FIXED, and the doc line too.** `STATUS_READ_RETRIES = 3` /
`STATUS_READ_RETRY_MS = 250` (`148-149`), budget at `284`, timer at `285`, the re-ask at `698-700` and
`706-714`, budget reset on a successful read (`684`), timer cleared in `dispose()` (`600-602`), and
`unref?.()` at `714` so a pending re-ask cannot outlive a finished script. The precedence is right: an
event that arrived mid-read still wins as the fresher re-ask (`696-698`) and a failure only schedules in
the `else`. It cannot weaken rule 3 — it retries `sessionsList()`, so the answer can only ever be one
the Core already settled on; the header now says so (`44-49`) and the constant's doc says why it has to
exist at all. Both tests land: `578-602` proves a `needs-input` transition still arrives after the read
carrying it failed with no second event ever coming, and `604-621` proves the re-asks stop at
`1 + STATUS_READ_RETRIES` rather than polling a dead Core. Against `30d1de1` both hang to timeout —
`waitForIdle()` pending forever on the first, one lone attempt on the second. The alternative went in
as well: `waitForIdle`'s doc at `543-547` now states plainly that no `timeoutMs` means an indefinite
wait and that a caller who must not hang on a broken Core passes its own deadline.

**Note 3 — the stale comment. FIXED, code untouched.** `terminal-screen.ts:205-208` now says each buffer
keeps what it loses off the top in its own scrollback, alternate screen included, and points at the
header rule (`35-42`) the old wording contradicted. That is what `221` does — `pushScrollback(buffer,
dropped)` on whichever buffer is shrinking — and what `terminal-screen.test.ts:166-182` asserts. `resize`
is byte-identical.

**Note 4 — the D11 sweep. FIXED, and checked in the direction that matters.** It walks `src/` **and**
`examples/` (`core-session.test.ts:642-643`, `665`), matches the shipped-extension set including `.mjs`,
and reports package-relative paths. The added `expect(swept).toEqual(expect.arrayContaining([…]))` at
`669-672` is the part I'd have asked for second: an empty walk passes an empty `offences` assertion, so
naming a file from each tree is what makes a future narrowing fail here. I reproduced the plant myself
in a scratch copy — `process.stdin.setRawMode(true)` appended to `examples/start-a-session.mjs` fails
the new sweep naming `examples/start-a-session.mjs`, and passes the `30d1de1` sweep blind. The forbidden
pattern is unchanged, and the example on this head is clean.

**Notes 5 and 6 — untouched, as scoped.** SGR is still parsed and dropped at `terminal-screen.ts:512-513`
(#156, and #156 is the Panel-migration ticket that will need reverse video). `sanitizeInitialInput` is
untouched — no diff in `packages/core` at all — and #193 ("sanitizeInitialInput drops newlines without a
separator") is a better home for it than the #191 comment I pointed at. Trivia corrected in the
description: 27 terminal, 28 session, which is what the run reports.

**Ran it.** `packages/sdk` on `af6ad81`: 145 passed, 5 skipped, 0 failed (the only red was my own
minimal install missing `selfsigned` for the mTLS suite, not the branch). `core-session.test.ts` 28 and
`terminal-screen.test.ts` 27 — 55, all green. This head's `core-session.test.ts` against `30d1de1`'s
source: 3 failed, 25 passed, and the three are precisely the three added here. CI on this head is green —
17 pass, 5 skipped (run 31600762018).

**One observation, not a request.** The retry budget is restored only by a *successful* read
(`684`), so a Session whose link stays down through all four attempts sits at zero until some later read
succeeds; a transient failure after that falls back to single-attempt behaviour. Likewise a second
failing read while a re-ask is already pending spends budget at `699` without scheduling anything, since
`scheduleStatusRetry` returns early on an armed timer. Both need a link that recovers and then blinks
again on the one transition that mattered, and the `waitForIdle` doc now covers the caller who cannot
risk it. Fine as it stands — worth knowing if #156 ever makes this path reconnect-aware.

### Checklist

- [x] 1 — held exit is an array, filtered by `ptyId` like the data path (`348`, `368`, `430-431`); test reproduces the co-tenant ordering and fails on the old code
- [x] 2 — bounded re-ask (3 × 250 ms), reset on success, cleared and `unref`'d; `waitForIdle` documents `timeoutMs`; two tests, both fail on the old code
- [x] 3 — resize comment corrected, agrees with `221` and the header rule; code unchanged
- [x] 4 — sweep covers `src/` and `examples/`, `.mjs` included, both trees asserted by name; planted violation caught here and missed by the old sweep
- [x] 5 — SGR left for #156
- [x] 6 — `sanitizeInitialInput` left for #193; `packages/core` untouched
- [x] Scope — three files, all `packages/sdk`; no force push; still a draft; base unchanged
- [x] Tests — 55/55 green on this head; 3/3 new tests red against `30d1de1`; CI green on `af6ad81`

### Verdict

All four in-scope notes are fixed at the lines that enforce them, each behavioural one with a test I
confirmed fails against the previous behaviour, and the two deferred notes are left where the review put
them. Nothing regressed. Re-approving on `af6ad81`.

PR #195, opened 2026-08-12
refactor(panel): the Panel's core link is the SDK's durable Core client (#156)

### LINKED ISSUE #156: 1.6 Migrate the Panel onto the SDK — phase 1's success test (D1, D14)

qcentic-dev:
## Summary

Phase 1's success test (#129 D14). `packages/panel/src/server/core-link/client.ts` is **deleted** —
along with its Node socket factory — and `CoreLinkManager` dials with `@actana/sdk`'s
`DurableCoreClient`, which is that file, extracted (#153). The Panel keeps no core-link client of
its own and behaves identically while it stops having one.

Three commits ahead of the migration are the extraction-parity notes the #153 review left explicitly
for this ticket: two fixes in the SDK's durable client, each with a test that fails without it, and
one judged-and-deferred question. None of them is a Panel behaviour change.

## Related issues

Closes #156

Refs #153, #148, #129

## Type of change

- [x] 🐛 Bug fix (non-breaking change fixing an issue) — the two SDK parity fixes
- [x] 🔧 Build / CI / tooling — one `tsconfig` flag, explained below

## The migration

**The port does not move.** `CoreLinkClientLike` is unchanged, so the panel-link router and every
fake in the suite see exactly what they saw. One adapter maps `onProtocolVersion` onto the SDK's
`onReady`, which carries the same two fields under the name the SDK gives them. Renaming the port
instead would have rippled into the router and into the suite — and a migration visible in tests is
the one thing this ticket forbids.

Kept identical on purpose, each because an operator or a Core can tell:

| | Before | After |
| --- | --- | --- |
| Reconnect backoff | 1s → 30s | same, passed through |
| Heartbeat | 15s ping / 45s dead | same (the SDK's durable default) |
| Request deadline | 30s | same |
| Event cursor | Core registry, read at construction, written per event | same store, same moments |
| mTLS + bearer | `ca`/`cert`/`key` + `rejectUnauthorized`, `noDelay` | same dial, now the SDK's |
| Core client id | `panel-<uuid>` | `panel-<uuid>` — see below |
| Dial starts | at construction | at construction |

**The `panel-` prefix is deliberate.** The SDK mints `sdk-…`, which is the honest name for a client
it knows nothing else about; this one has a name and has always presented it, and once the CLI ships
on the same package the prefix is the only thing telling the two apart on a Core serving both. A new
test pins it.

`connect()` is awaited by nobody: a refused bearer arrives on `onAuthError` and a dead Core on
`onDisconnected`, which is where this manager has always read them, and the client reconnects
underneath either way.

### The one difference I found, stated rather than buried

When the *socket factory itself* throws — a malformed endpoint, TLS material the runtime refuses —
the old client emitted its `disconnected` before the manager had attached a listener, so the status
sat at `connecting` until the first backoff attempt also failed. The SDK reports that same failure
one microtask later, so the listener is there and the Core goes to `unreachable` at once. Same end
state, reached a beat sooner, on a path that needs an operator to fix the endpoint either way. Every
other transition — `connecting` → `connected` on `authOk`, `auth-error`, `needs-update`,
`unreachable` — is frame-for-frame what it was.

### `allowImportingTsExtensions`

On for the Panel's typecheck, because the Panel now reads SDK *sources* through its `paths` mapping
and the SDK's modules import each other with the `.ts` extension on purpose — that is what lets a
plain `node` process import the package with no bundler and no loader. Until this change only the
extensionless leaf `core-link-frames` was reachable, so it never came up. Safe for the same reason it
is safe in `packages/sdk`: `noEmit`, and Vite resolves the specifiers (`pnpm build` green, the
durable client is in `dist/server/server.js`).

## The three parity items from the #153 review

1. **`fix(sdk)`: a request issued between writability and `ready` is queued, not sent.** Note 2 of the
   second review, at the other door: `onWritable`'s flush was already gated on `established`, but
   `dispatch()` sent whenever the transport was writable — so on the no-bearer path a `request()`
   *issued* in that window went out ahead of the `reclaim` and `subscribe` the connection owes. The
   test opens the socket, asserts the link is writable while `ready` is outstanding, issues a
   `replay` there, and asserts nothing of the caller's is on the wire yet.
2. **`fix(sdk)`: arm the reconnect backoff before the disconnect listeners run.** The orphan-socket
   window: a listener calling `connect()` synchronously found `transport` null and
   `reconnectScheduled()` false, dialed, and had its socket overwritten by the backoff's own
   `openTransport()` — one live link and one orphan nobody holds. Reachable now precisely because the
   Panel registers disconnect handlers. Nothing observable is reordered: what now runs first either
   arms a timer or rejects a promise, and a rejection lands after the listener loop either way.
3. **`docs(sdk)`: reverse video in `terminal-screen`, judged and deferred.** The emulator drops SGR,
   so a caller answering a dialog through `send()` reads the options and not the selection. **The
   Panel does not need it and does not use the class at all** — its terminals are xterm.js in the
   browser, which renders the highlight, and the one dialog the product answers unattended is
   answered in the Core off the raw stream (`harness-prompt-delivery.ts`, which reads `ESC[7m`
   exactly). Fixing it means choosing what a selection looks like in text a published `screen()`
   hands somebody's script — the session layer's call, with its caller in front of it. So: documented
   in the header with the shape a fix would take and why the current failure is the safe one
   (a caller that sees no highlight answers by number; a half-marked highlight would confidently
   answer the wrong row of a dialog whose first option is "No, exit"), plus a test pinning the gap.

## What happened to the deleted client's tests

They went with it, and nothing else did:

- `client-id-reclaim.test.ts`, `multiconnection-capability.test.ts`, `mtls-integration.test.ts` —
  every test in all three drove `PtyCoreLinkClient`. `packages/sdk` covers the same ground against
  the Core's own `PtyCoreLinkServer`, including a real `wss://` handshake on a real port.
- `client.test.ts` was **half `@actana/core`'s tests** — the server's `ready` frame, its auth gate,
  its event tail, its mutation port. Deleting the file would have taken that with it, so the client
  halves were cut and the rest is byte-identical under the name it now deserves,
  `core-link-server.test.ts`. `git diff -M` shows it as the rename it is.

No existing test was edited to accommodate this migration. Two tests were **added**
(`core-link-client-id.test.ts`), and they cover something the old code did not pin.

## How was this tested?

- [x] Manually verified (below)

**The existing suite is what checks this migration, and it does it end to end.**
`panel-link/live-read-path`, `panel-link/write-path`, `panel-link/terminals-in-browser`,
`cores-api` and `session-finish-from-core.integration` register a real Core through the real
`/api/cores` route, which goes through the real `CoreLinkManager` and therefore through the migrated
`defaultCreateClient` — a genuine mTLS `wss://` dial to a real `PtyCoreLinkServer`, a dial status that
has to reach `connected`, browser frames routed to the Core and answers routed back, PTY output
streamed to a tab, and a panel link killed mid-flight whose reconnect replays the events that
happened while nothing was attached. 122 of them, unchanged, green.

```
@actana/sdk     151 passed | 5 skipped
@actana/shared  180 passed
@actana/core    826 passed
@actana/panel   1338 passed
pnpm -r typecheck  clean
pnpm lint          0 errors (51 warnings; 53 on the base — the two fewer are the deleted files')
pnpm build         green; DurableCoreClient bundled into dist/server/server.js
```

`pnpm panel:e2e` — the black-box smoke that boots the built Panel against a real Core — does not run
on this box: the Core fixture dies before the Panel is started ("core printed no registration blob"),
and it does so **identically on `e83a2e6`**, which I checked in a scratch worktree. Same reason
`scripts/__tests__/core-launcher.test.mjs` fails on the base here — this container has a real
`/opt/actana` install the fixtures collide with. Worth running on CI, where it is a required leg.

This container is slow enough to tip 5s-default tests over on a loaded run — the same flake the #153
PR documented. On the full parallel run two timed out (`operator-auth`, `live-read-path`); both pass
in isolation and on a `--maxWorkers=2` run of the whole Panel suite. The base commit fails six the
same way, in the same files, and `scripts/__tests__/core-launcher.test.mjs` fails on the base too
because this box has a real `/opt/actana` install. Neither is this change.

## Checklist

- [x] This PR targets `refactor/sdk-extraction` — phase 1's integration branch, not `main` and not
      the train
- [x] PR title and every commit follow Conventional Commits; branch is `refactor/panel-on-sdk`
- [x] I added tests proving the change works; all suites pass locally
- [x] `CONTEXT.md` needs no change: "Durable Core client — the second entry point … for a program
      that stays connected — the Panel" already describes this, and now it is true
- [x] Draft: #139's overnight proof is still outstanding and nothing here absorbs it — this
      migration is behaviour-preserving including that residual bug, deliberately


qcentic-dev:
## Acceptance checklist — every criterion in #156, checked or explicitly not

### Done when

- [x] **The Panel behaves identically — not "equivalently" — from an operator's point of view.**
  The port (`CoreLinkClientLike`) is unchanged, so the panel-link router sees what it saw. Backoff
  (1s→30s), heartbeat (15s/45s), request deadline (30s), the registry-backed event cursor, the mTLS
  material, the bearer and the `panel-…` Core client id are all carried across deliberately, and the
  dial still starts at construction. Every dial-status transition an operator reads in the Fleet view
  — `connecting` → `connected` on `authOk`, `auth-error`, `needs-update`, `unreachable` — is
  frame-for-frame what it was, and the end-to-end suites assert them against a real Core.
  **One transient differs, and I am naming it rather than burying it:** when the *socket factory
  itself* throws (a malformed endpoint, TLS material the runtime refuses) the old client emitted its
  `disconnected` before this manager had attached a listener, so the Core sat at `connecting` until
  the first backoff attempt also failed; the SDK reports it one microtask later, so the Core reads
  `unreachable` at once. Same end state, one beat sooner, on a path an operator has to fix by editing
  the endpoint either way. If a reviewer reads that as a behaviour change rather than a transient,
  say so and I will make the manager swallow the first report to match the old timing exactly.

- [x] **The existing suite passes unchanged; no test was edited to accommodate the migration.**
  `@actana/sdk` 151 ✓ / 5 skipped · `@actana/shared` 180 ✓ · `@actana/core` 826 ✓ · `@actana/panel`
  1338 ✓ (130 files, `--maxWorkers=2`; on a fully parallel run this box tips two 5s-default tests
  over — see the PR body, the base fails six the same way). `pnpm -r typecheck` clean, `pnpm lint`
  0 errors, `pnpm build` green. **CI agrees**: `Unit Tests` green on this branch, along with
  `Typecheck`, `Lint`, `Conventions`, `Train rules`, `Secret Scan`, `Dependency Audit`, both image
  smokes and both Core tarball smokes.
  What was removed was the deleted client's *own* coverage, and only that:
  `client-id-reclaim.test.ts`, `multiconnection-capability.test.ts` and `mtls-integration.test.ts`
  drove `PtyCoreLinkClient` in every test they contain, and `packages/sdk` covers that ground against
  the Core's own server plus a real `wss://` handshake. `client.test.ts` was half `@actana/core`'s
  tests — those are untouched and byte-identical, under the name they now deserve
  (`core-link-server.test.ts`). Two tests were **added**, pinning the `panel-` client id.

- [x] **The Panel's own copy of the client is gone, not deprecated.**
  `packages/panel/src/server/core-link/client.ts` (1474 lines) and its `node-socket.ts` are deleted.
  `git grep PtyCoreLinkClient -- packages/panel/src` returns nothing at all. Nothing is re-exported,
  aliased or shimmed.

- [x] **Reconnect, heartbeat, event cursor and subscribe-then-replay still work.**
  Under test, against the Core's real `PtyCoreLinkServer`: the SDK's durable suite covers the
  reconnect loop with backoff, the heartbeat pinging an idle link and destroying a half-open one, the
  cursor deduping and persisting, subscribe-then-replay closing a gap with no double and no hole, PTY
  subscriptions re-asked for on the next connection, and a restarted client resuming from the store it
  was handed. On the Panel side, `panel-link/live-read-path`, `panel-link/write-path`,
  `panel-link/terminals-in-browser`, `cores-api` and `session-finish-from-core.integration` — 122
  tests, unchanged — drive all of it through the migrated `defaultCreateClient`: a genuine mTLS
  `wss://` dial, a dial status that has to reach `connected`, a link killed mid-flight, and events
  that happened while nothing was attached arriving on the replay.

- [x] **…"across a Panel reload", confirmed black-box — by CI, not by this container.**
  `pnpm panel:e2e` boots the *built* Panel as a plain process against a real Core and includes
  exactly this leg: a link killed mid-flight, events happening on the Core while no browser is
  attached, and a reconnected link replaying its cursor over every one of them. **CI ran it on this
  branch and it passed** — `E2E — Panel service seam`, 45s,
  https://github.com/actana/control/actions/runs/31604388578/job/94139492344 — as did `Unit Tests`
  (the whole workspace, 1m31s), with none of this box's timing flakes.
  It does **not** run on this machine: the Core fixture dies before the Panel is started (`core
  printed no registration blob`), because this box has a real `/opt/actana` install the fixture
  collides with. I checked it out at `e83a2e6` in a scratch worktree and it fails there identically,
  so it was never this branch — the same environment fails
  `scripts/__tests__/core-launcher.test.mjs` on the base for the same reason.

### Traps

- [x] **Nothing was fixed by adding a feature.** Nothing went red: no test in the suite failed at any
  point during this migration, so there was no red to be tempted by. The three SDK commits are the
  parity notes the #153 review left for this ticket, each pinned by a test that fails without it, and
  none of them changes Panel behaviour.

- [x] **The transport assumption holds where the Panel actually runs (ADR 0012).** Checked, and worth
  a note: the trap as worded in the issue ("the Panel's transport is the browser's, not the server's")
  reads backwards for *this* code path. ADR 0012 says every **core link terminates inside the Panel
  service** precisely because browsers cannot hold mTLS client certificates (ADR 0002); the browser's
  transport is the **panel link**, which this change does not touch. So the question that matters is
  what the SDK's durable client assumes about its runtime — and it assumes Node: `core-link-socket.ts`
  imports `ws` and hands `ca`/`cert`/`key` straight to TLS, which is the only way to present a client
  certificate, and the heartbeat arms only where `ping`/`terminate` exist, which is the same `ws`.
  That is exactly the runtime the Panel service is. Nothing in the SDK reaches for a DOM global — the
  cursor store is injected, and `packages/sdk`'s own boundary test asserts a planted `localStorage` is
  never touched.

- [x] **#139 is not absorbed.** No heartbeat cadence, liveness timer or reconnect policy was changed
  in either direction; the migration carries the behaviour across including any residual bug, so an
  overnight run still measures what it measured. Nothing here touches the panel-link liveness fix that
  landed in 0.2.0 (#172).

- [x] **`row.coreLabel` is not rendered.** `packages/panel/src/shared/fleet-merge.ts` is untouched and
  `coreLabel` appears nowhere in this diff — the link-time copy a rename does not move stays exactly as
  the 0.2.0 train left it.

### Housekeeping

- [x] Base is `refactor/sdk-extraction`, not `main` and not the train.
- [x] Five commits, Conventional Commits throughout; branch `refactor/panel-on-sdk`.
- [x] Draft, and staying draft.
- [x] `CORE_LINK_PROTOCOL_VERSION` does not move; no manifest version changes; no new dependency
      (`@actana/sdk` was already on the Panel's manifest).

qcentic-adm (approved):
## Review — #156, phase 1's success test (D14)

Reviewed as the binary judgment the phase rides on: every claim below was checked by reading the
enforcing line, diffed against `refactor/sdk-extraction` at `e83a2e6`. **Approving.** One coverage
gap to carry forward as a follow-up, one stale comment, and an explicit ruling on the transient you
named. Detail follows so the ruling is auditable rather than taken on trust.

---

### 1. The client is gone — deleted, not deprecated ✅

`packages/panel/src/server/core-link/` now contains exactly one entry: `__tests__/core-link-server.test.ts`.
`client.ts` (1474 lines) and `node-socket.ts` (86) are `D` in `--name-status`, not moved. No barrel, no
re-export, no alias, no shim: `git grep "core-link/" -- packages/panel/src` outside `__tests__` returns
nothing, and the only surviving `PtyCoreLinkClient` strings in the tree are historical docs
(`docs/specs/08-*.md`, `docs/upstream/PROVENANCE.md`) and SDK provenance comments. No manifest change and
no lockfile change — `@actana/sdk` was already a Panel dependency at the base (`packages/panel/package.json:22`).

### 2. The suite is unchanged, and the deletions are the deleted class's own ✅ (with a caveat, below)

The mechanical claim holds exactly:

- The rename is `R069`, and the 31% is **all deletion**. Every `+` line in
  `client.test.ts` → `core-link-server.test.ts` is a comment: the new header, and one line reworded
  because it named the deleted class (`// holds; `client` is the one the peer on the other end holds`).
  Zero code added. Every removed `describe`/`it` is `PtyCoreLinkClient …`; every survivor is
  `PtyCoreLinkServer …` plus the server side of `core-link heartbeat`. Clean split.
- The three SDK test files are **pure additions** — removed-line count is 0 for
  `core-client.test.ts`, `durable-core-client.test.ts` and `terminal-screen.test.ts`.

So: no test was edited to accommodate the migration. The ticket's criterion is met on its own terms.

**Now the caveat, claim by claim on the three deleted suites.** "`packages/sdk` covers the same ground"
is right for two of the three and overstated for the third.

**`mtls-integration.test.ts` (2 tests) — fully re-covered, and improved.**
`core-link-mtls.test.ts:164` is the wss://-with-mTLS-then-round-trip test; `:178` is the pinned-cert
refusal. `:210` adds one the Panel never had (a refused handshake told apart from a Core that is not
running). Net gain.

**`client-id-reclaim.test.ts` (11 tests) — mostly re-covered; two assertions died, both structurally safe.**
On the wire once up → `core-client.test.ts:245`, which asserts the `clientId` on the reclaim frame.
Ahead of the replay tail and PTY resend → `durable-core-client.test.ts:66` and `:328`. Withheld until the
bearer is accepted → `durable:66` asserts `["auth", "reclaim", "subscribe"]`. Caller-supplied id →
`core-client.test.ts:252`. Distinct per client → your new `core-link-client-id.test.ts`. Two did not
make it across:

- **"is the same string on every reconnect — the property the feature rests on."** `durable:181`
  asserts the second connection's frame *types* and never that its reclaim's `clientId` equals the
  first's. The property is safe by construction — `readonly clientId` at `core-client.ts:254`, assigned
  once at `:329`, read at the single send site `:758` — so this is a lost regression detector, not a
  behaviour risk. But the deleted test named it load-bearing, and nothing pins it now.
- **"carries on when the Core refuses the reclaim."** The `.catch()` at `core-client.ts:774` is unasserted.

**`multiconnection-capability.test.ts` (~35 tests) — this is the real gap.**
The SDK's entire client-level coverage of the gate is one test, `core-client.test.ts:315` (capability
false, capability null, `claim` degrades, `ptySubscribe` degrades, `forceTakeover` rejects), plus
`core-link-frames.test.ts:430-464` unit-testing `readMultiConnectionCapability` — the parser, not the
client's lifecycle. What has no equivalent anywhere:

- The capability's **lifecycle on a client** — closed before `ready` lands, recorded on announce, not
  carried across a reconnect, cleared the moment the link drops rather than when the next opens, picked
  back up when an upgraded Core reconnects. The code is right; I read it at `core-client.ts:402`, `:413`
  and `:448`. It is now unpinned. Worth noting that the deleted client's own comment (base
  `client.ts:478-482`) called this exact edge load-bearing: a stale `true` would let a gated frame past
  the check onto `queuedFrames`, which the reopen drain writes unconditionally, before the new `ready`
  says whether it is allowed. The two tests that pinned that hazard — "does not queue it for a later
  flush" and "refuses it in the drop window, and nothing flushes when the socket reopens" — are gone.
- **The frame registry's contents.** `MULTI_CONNECTION_ONLY_FRAME_TYPES` (`core-client.ts:83`) is
  asserted by no test in the repo. `release`, `ptyUnsubscribe` and `reclaim` never go through the gate
  under test. Removing one, or adding a seventh, is currently green.
- "stops sending it again if the Core comes back downgraded"; the capability-transition PTY cases
  ("asks for real once that Core comes back announcing the capability"); the ready/authOk ordering guard
  ("self-refuses if authOk ever arrives first — the ordering is the guard"); and two Session-lock
  details ("distinguishes 'no lock on this Core' from 'someone else holds it'", "leaves the code
  undefined on an error that carries none" — the *coded* case is covered at `core-client.test.ts:195`).

Two things keep this off the blocking list. The Panel's *use* of the gate is still covered and untouched
— `panel-link/router.test.ts`, `panel-link/session-write-access.test.ts`, `shared/session-write-access.test.ts`
and `services/core-link-manager.test.ts:250` all still drive `canSendMultiConnectionFrames()`. And the
gap is inherited: the SDK's suite was written in #153, and #156 is only where the originals are deleted.
It is not a behaviour change and not something #156's "done when" list forbids. **Please raise a
follow-up on `refactor/sdk-extraction` to port the capability-lifecycle block and a registry-contents
test into `packages/sdk`** — that is the honest home for them, and the list above is the porting spec.

### 3. The named transient — **ruling: acceptable transient. Do not swallow it.** ✅

Verified precisely, and it is exactly as you described it. Old: `client.ts:382` dialed inside the
constructor, so a `createSocket` throw ran `emitDisconnected` at `:409` **synchronously, before
`defaultCreateClient` returned** — the manager attaches `onDisconnected` at
`core-link-manager.ts:219`, after the constructor, so the first report went into a void and the row sat
at `connecting` (set at `:171`) until the first backoff attempt also failed. New:
`core-link-transport.ts:181` defers the same report by `queueMicrotask`, the listeners at `:186-220`
are all attached synchronously, and the row reads `unreachable` at once.

You offered to make the manager swallow the first report. **Don't.** Three reasons:

1. The old timing was not a behaviour, it was a dropped message — an artifact of dial-in-constructor
   racing its own owner's subscription. There is nothing there to preserve.
2. Restoring it means adding a suppression rule to `CoreLinkManager` whose only purpose is to show
   `connecting` for one backoff window when the Panel already knows better. That is new, permanent,
   untested Panel logic that outlives the migration — and trap 1 of #156 says do not fix it by adding a
   feature. Swallowing would be the behaviour change; this is the absence of one.
3. The end state, the `detail` string, and every other transition are frame-for-frame identical, and the
   window is bounded by one 1s backoff on a Core an operator must go and re-endpoint either way.

**One divergence you did not name, for completeness — also fine.** The old `trySend`
(base `client.ts:781`) gated on `!this.socket || !this.ready`, so a request issued after `ready` but
before `authOk` went out. The SDK gates on `established`, which is `ready && transport.writable`
(`core-client.ts:505`), and `writable` requires `authenticated` on a bearer link
(`core-link-transport.ts:243`) — so that request is now queued instead. Strictly safer: the Core answers
a pre-auth frame with an error *and closes the socket* (`core-link-server.test.ts:820`), so the old path
killed the link. Sub-millisecond window, needs an operator action inside it, and the new behaviour is
the correct one. Worth a line in the PR body rather than a change.

### 4. The three parity commits ✅

**`fb35d74` — dispatch-door ordering.** This one is stronger than the PR body claims: it is not merely
"no Panel behaviour change", it *prevents* one. The old client already held a request until `ready`
(base `client.ts:781`). The pre-fix SDK gated `dispatch()` on `transport?.writable` alone, which on the
no-bearer path lands before `ready` — so the SDK had **diverged from the code it was extracted from**,
and without this commit the migration would have introduced a real reordering. `core-client.ts:657` now
reads the same `established` flag as `onWritable`'s flush. Test at `core-client.test.ts:448` pins the
window rather than the outcome — asserts writability while `ready` is outstanding, then
`["reclaim", "replay"]`. Correct fix, correct test.

**`9346028` — orphan-socket window.** Verified against the old close handler (base `client.ts:474-488`):
old order was state-reset → `emitDisconnected` → `scheduleReconnect`. The reorder moves
`onConnectionClosed` ahead of the listeners, so the SDK now arms the timer *before* them where the old
client armed it after. Nothing observable moves, and I checked each half: `onConnectionClosed` is
`caughtUp = false; scheduleReconnect()` (`durable-core-client.ts:124-127`) — a field write and a
`setTimeout` — plus `failConnect` (`core-client.ts:531`), whose rejection is a microtask landing after
the synchronous listener loop. The `caughtUp` half actually *restores* parity: the old client reset
`subscribed` before its listeners (base `client.ts:476`) and the pre-fix SDK reset `caughtUp` after.
No Panel behaviour change, because the Panel's disconnect listener
(`core-link-manager.ts:219-221`) only calls `this.set(...)`. Test at `durable-core-client.test.ts:425`
is precise — two pairs not three, and the first socket closed.
One wording nit: "Reachable now precisely because the Panel registers disconnect handlers" is a touch
strong. The Panel registers one, but it does not call `connect()`, so the window is a latent hazard for
SDK callers rather than a live Panel path. The fix is still right; the justification is one step longer
than stated.

**`5baa187` — SGR reverse video, deferred. Judged on its merits: the deferral is correct, and it is
the right call rather than merely a defensible one.** Every factual premise checks out: the Core does
read `ESC[7m` (`packages/core/src/harness-prompt-delivery.ts:145-151`, `:297`, with
`harness-prompt-delivery.test.ts:275`), and the Panel does not reference `TerminalScreen` anywhere —
`git grep` in `packages/panel/src` is empty. The asymmetry argument is the deciding one and it is sound:
no-highlight fails toward answer-by-number, which works on every harness, whereas a half-marked
highlight fails toward confidently pressing Enter on the wrong row of a dialog whose first option is
"No, exit". And the fix genuinely is a published-surface decision (sentinel vs. second accessor vs.
per-line attribute) that belongs to the session layer with a caller in front of it, not to a migration
whose ticket forbids adding features. The test at `terminal-screen.test.ts:275` pins the gap rather than
the wish — asserting neither the escape nor a sentinel survives — which is the right shape: it makes a
future fix a deliberate, visible change instead of a silently-passing one. Keep it as is.

### 5. Reconnect, heartbeat, cursor, subscribe-then-replay across a reload ✅

The evidence covers a reload, in both senses, and I checked CI myself rather than taking the report.

- **Browser reload** (panel link drops, core link survives — the ADR 0012 shape):
  `scripts/e2e-panel-smoke.mjs:444-528` kills the panel link mid-flight, lets a `pty:exit` happen with
  nothing attached, reopens and asserts the replay carries it, that nothing at or below the cursor is
  resent, and that ids are strictly ascending. No double, no hole.
- **Panel service restart** (core link re-dialed, cursor re-read from the registry — the migration's
  real risk): the same script carries one data directory through **two service restarts** (`:122`), and
  `assertLoginAndDial` (`:580`) re-dials through the migrated `defaultCreateClient` and requires
  `connected`. `assertCoreCannotDialWithoutKey` (`:557`) additionally pins that a Core with unreadable
  secrets reports `auth-error` rather than sitting in `connecting`.
- Cursor continuity across a *service* restart is covered at unit level —
  `durable-core-client.test.ts:272` ("resumes a restarted client's timeline from the store it was
  given") plus the registry-backed store wired at `core-link-manager.ts:346-352`. The E2E asserts the
  re-dial reconnects but not that the cursor survives the process boundary end to end. That gap is
  pre-existing and identical under the old client, so it is not this PR's to close — but the checklist's
  "confirmed black-box" is slightly wider than what the E2E actually asserts. Worth a follow-up leg.
- Heartbeat and reconnect against a real Core: `durable-core-client.test.ts:152` (backoff + timeline
  restored), `:227` (pings idle, destroys half-open), `:244` (pong-less link left alone).

**CI verified independently** on run `31604639648` — `E2E — Panel service seam` pass (41s), `Unit Tests`
pass (1m30s), plus Typecheck, Lint, Conventions, Train rules, Secret Scan, Dependency Audit, both image
smokes and both tarball smokes. Nothing failing, nothing skipped that matters. This is a newer run than
the one cited in your checklist and it is green too.

### 6. #139 not absorbed ✅

No cadence moved in either direction, and I checked the defaults rather than the call site, since the
manager relies on them: heartbeat is `opts.heartbeat ?? {}` (`durable-core-client.ts:101`) resolving to
`DEFAULT_HEARTBEAT_INTERVAL_MS = 15_000` / `DEFAULT_HEARTBEAT_TIMEOUT_MS = 45_000`
(`core-link-transport.ts:131-133`) — the deleted client's exact numbers. Request deadline is
`DEFAULT_REQUEST_TIMEOUT_MS = 30_000` (`core-client.ts:201`), matching the old `DEFAULT_RPC_TIMEOUT_MS`.
Backoff is passed explicitly and `RECONNECT_INITIAL_MS = 1_000` / `RECONNECT_MAX_MS = 30_000`
(`core-link-manager.ts:99-100`) are untouched by this diff. The only heartbeat constants in the diff at
all are deletions from `client.ts`. The overnight bug still measures what it measured.

Nice confirmation of "it is that file, extracted": `sendReclaim` at `core-client.ts:756-775` is
byte-identical to base `client.ts:1096-1117`, gate guard and `.catch()` included.

### 7. `fleet-merge.ts` untouched, `coreLabel` not newly rendered ✅

`packages/panel/src/shared/fleet-merge.ts` is absent from `--name-status`, and `coreLabel` appears
nowhere in the diff.

### 8. The ADR 0012 reading ✅ — your correction is right

Checked against `docs/adr/0012-panel-link-browser-transport.md`, whose opening clause is
"With every core-link terminating inside the Panel service (browsers cannot hold mTLS client
certificates — ADR 0002), the Panel UI needs its own live transport… **the panel link**." So the
browser's transport *is* the panel link, and the trap as worded in #156 does read backwards for this
code path: the core link is the Node-side one, and this PR does not touch the panel link at all.

The substantive question underneath it — what the SDK assumes about its runtime — is answered and
tested. `core-link-socket.ts:17` is the only module importing `ws`; TLS material goes straight to it;
the heartbeat arms only where `ping`/`terminate` exist. And the no-DOM claim is pinned, not asserted:
`package-boundaries.test.ts:146` plants a `localStorage` on `globalThis` and requires it stay untouched,
alongside `:95` restricting imports to own-modules-and-`ws`. That is the Panel service exactly.

### Minor — one stale comment

`packages/sdk/src/core-link-frames.ts:830` still reads "see `MULTI_CONNECTION_ONLY_FRAME_TYPES` in the
Panel's core-link client". That client is deleted as of this PR, and the constant now lives at
`packages/sdk/src/core-client.ts:83`. `13eaf90` fixed four module headers for precisely this reason and
missed this one — it is inside a type's doc comment rather than a header. Comment-only; fix it whenever,
it does not need another review pass.

### `allowImportingTsExtensions`

Justified and correctly scoped: `noEmit` is on, Vite resolves the specifiers, `Typecheck` and the image
builds are green in CI, and the comment at `packages/panel/tsconfig.json:20-28` explains why the Panel
only became the first consumer to need it now. No objection.

---

## Checklist

| # | Claim | Verdict |
|---|---|---|
| 1 | `core-link/client.ts` gone — deleted, nothing re-exported/aliased/shimmed | ✅ verified |
| 2 | Suite passes unchanged; no existing test edited | ✅ verified — rename is deletion-only, SDK test files are pure additions |
| 2a | `mtls-integration` re-covered by the SDK | ✅ fully, plus one test the Panel never had |
| 2b | `client-id-reclaim` re-covered | ⚠️ mostly — id-stable-across-reconnect and refused-reclaim died; both safe by construction |
| 2c | `multiconnection-capability` re-covered | ⚠️ **overstated** — the gate's lifecycle, the drop-window no-queue property and the frame registry have no SDK equivalent. Follow-up, not a blocker |
| 3 | The named transient | ✅ **acceptable transient — do not swallow it** (ruling below) |
| 3a | A second, unnamed divergence (ready→authOk now queues) | ✅ strictly safer; note it in the body |
| 4a | `fb35d74` dispatch door | ✅ no Panel change — it *prevents* one; the SDK had diverged from the original |
| 4b | `9346028` orphan socket | ✅ no Panel change; `caughtUp` ordering restores parity |
| 4c | `5baa187` SGR deferral | ✅ premises verified; the deferral is the right call |
| 5 | Reconnect/heartbeat/cursor/replay across a reload | ✅ covered both senses; CI green on run `31604639648`. Cursor-across-service-restart is unit-level only |
| 6 | #139 not absorbed | ✅ 15s/45s, 30s, 1s→30s all identical |
| 7 | `fleet-merge.ts` untouched, `coreLabel` not rendered | ✅ verified |
| 8 | ADR 0012 reading | ✅ your correction is right; runtime assumption tested |
| — | Stale comment `core-link-frames.ts:830` | ⚠️ minor, comment-only |

## Verdict

**Approve.** Phase 1's success test is green, and it is green for the right reason: the Panel's client is
gone, the SDK's is the same code, and the suite that proves it was not touched to make it pass. The two
parity fixes are genuine parity — one of them closes a divergence that would otherwise have made this
migration observable — and the SGR deferral is reasoned rather than dodged. Nothing here absorbs #139.

**Ruling on the named transient: it is an acceptable transient, not a behaviour change, and it should
be left exactly as it is.** The old client's `connecting` window was a report dropped on the floor by a
constructor that dialed before its owner could listen, not a behaviour anyone designed or an operator
relies on. Reproducing it would mean teaching `CoreLinkManager` to discard a true status and display a
false one for one backoff window — new permanent Panel logic added to preserve a race artifact, which is
the feature-shaped fix trap 1 forbids. Same end state, same detail, every other transition identical.
Do not swallow the first report.

The one thing I would not let slide silently is 2c: "`packages/sdk` covers the same ground" is true of
`mtls-integration` and broadly true of `client-id-reclaim`, but not of `multiconnection-capability`,
where roughly a dozen assertions on the gate's lifecycle and registry left the repo with this PR and
have no home yet. The code they guarded is correct — I read it — so this is a detector gap, not a
regression, and it belongs to the extraction rather than to the migration. Please open the follow-up
against `refactor/sdk-extraction` before the branch merges up, and treat the list in §2 as its spec.

Draft status respected — not marking ready, and #139's overnight proof is yours to close.

PR #196, opened 2026-08-12
test(sdk): the multiConnection gate's dead detectors, rebuilt on the SDK's durable client

qcentic-dev:
The follow-up section 2 of [qcentic-adm's review on #195](https://github.com/actana/control/pull/195#pullrequestreview-4917520207)
asked for, on the base that review named: **"port the capability-lifecycle block and a registry-contents
test into `packages/sdk`"**. That review's section 2 *is* the porting spec, and the checklist at the
bottom maps every assertion it enumerates to its home, checked or explicitly not.

The suite the originals belonged to was written in #153 and the originals were deleted in #156 —
`client-id-reclaim.test.ts`, `multiconnection-capability.test.ts` and `mtls-integration.test.ts` drove
`PtyCoreLinkClient`, which no longer exists. **The code they guarded is correct**; the review read it
at `core-client.ts:402`, `:413` and `:448` and approved the PR. This is a detector gap and is treated
as one.

### Scope

- **Tests only. No production code changes** — the diff is three files under `packages/sdk/src/__tests__/`.
  Nothing under `src/` outside `__tests__` is touched, in this package or any other.
- Two new suites, `multiconnection-gate.test.ts` (14 tests) and `client-id-reclaim.test.ts` (2), plus a
  hand-driven dialer added to the shared rig.
- Base `refactor/sdk-extraction`, draft, and staying draft.

### Rebuilt against the client the Panel runs

Every test drives `DurableCoreClient` against the Core's own `PtyCoreLinkServer`. A Core that "comes
back downgraded" is a **second real server** that announces nothing, dialed on the reconnect — not a
`ready` frame written by the test — and every reconnect is a genuinely new socket with its own `auth`,
its own `ready` and its own subscription state.

Two shapes a real Core cannot produce, and the one place this suite hand-feeds frames:

- **`authOk` before `ready`.** The Core sends `ready` as frame one on every connection, which is
  exactly why nothing in the client enforces the order — so the only way to pin what that order buys is
  to invert it.
- **an `error` answer to a `reclaim`.** The server's reclaim handler cannot fail, and the client's
  tolerance of one that does is a `.catch()` no reachable server path exercises.

Both are documented at `handDrivenDialer` in `fake-core-link.ts`, along with the instruction to reach
for a real Core otherwise.

### Each one fails when the behaviour it guards is broken

Not asserted — run. Each mutation was applied to `packages/sdk/src/` on its own, the two suites run, and
the mutation reverted; the tree carries none of them.

| Mutation | Tests it kills |
| --- | --- |
| Gate open before `ready` (capability defaults to `{version: 1}`, no per-connection reset) | closed before ready lands |
| Capability never recorded off `ready` | 10 of the 16, incl. records-on-announce and picks-it-back-up |
| Capability remembered across connections (sticky read + neither clear) | does not carry it across a reconnect · closes on drop · drop window · downgraded |
| No clear on `onClose` | closes the moment the link drops · drop window |
| Gate removed from `request()` | does not queue it · drop window · downgraded · every type through the gate |
| `release` dropped from the registry | those three, plus both registry tests |
| A seventh type added to the registry | both registry tests |
| Establishment without `ready` | self-refuses if authOk arrives first |
| PTY want dropped when the gate is closed | asks for real once that Core comes back |
| `claim` answering `supported: true` with no capability | distinguishes no-lock from held |
| `CoreLinkRequestError` defaulting its `code` | leaves the code undefined |
| Fresh client id minted per connection | same string on every reconnect |
| `.catch()` removed from `sendReclaim` | carries on when the Core refuses the reclaim — as an unhandled rejection that **fails the run** with every assertion in the file still passing |

Two findings worth recording rather than burying:

1. **`openTransport`'s `multiConnection = null` is redundant on its own** — removing it alone kills no
   test, because `onClose` has already cleared it on every path that reaches a second connection. It is
   belt-and-braces, not dead: the composite mutation above needs *both* gone plus a sticky read before
   the lifecycle assertion moves. Recorded so the next reader does not mistake a passing suite for a
   claim that either line is individually load-bearing.
2. **The refused-reclaim detector fails the run, not an assertion.** A swallowed rejection has no
   observable consequence to assert on — that is what swallowing it means — so what the mutation
   produces is an unhandled rejection attributed to the file. CI goes red; no `it()` reports it.

### Verification

`@actana/sdk` **167 ✓ / 5 skipped** (was 151 ✓ / 5 skipped at `1e0b0b8`; +16, all additions — no existing
test's assertions were changed, and the only edit to an existing file is the additive dialer in
`fake-core-link.ts`). `pnpm -r typecheck` clean. `pnpm lint` **0 errors**, and none of the 51 pre-existing
warnings is in a file this PR touches.

---

### `multiconnection-capability.test.ts` — "the real gap"

**The capability's lifecycle on a client** (review's first bullet):

- [x] closed before `ready` lands → `multiconnection-gate.test.ts:171` *"is closed before ready lands — the answer is unknown, not yes"*
- [x] recorded on announce → `:194` *"records the capability the Core announced on ready"*
- [x] not carried across a reconnect → `:203` *"does not carry it across a reconnect — it is re-read on the new ready"*
- [x] cleared the moment the link drops rather than when the next opens → `:219` *"closes the moment the link drops, not when the next one opens"*
- [x] picked back up when an upgraded Core reconnects → `:234` *"picks it back up when an upgraded Core reconnects"*
- [x] the load-bearing hazard, "does not queue it for a later flush" → `:249` *"does not queue it for a later flush"*
- [x] the load-bearing hazard, "refuses it in the drop window, and nothing flushes when the socket reopens" → `:265`, same name

**The frame registry's contents** (second bullet):

- [x] `MULTI_CONNECTION_ONLY_FRAME_TYPES` asserted — removing one or adding a seventh is no longer green → `:308` *"holds exactly the PTY subscription frames, the Session lock's three, and reclaim"*
- [x] `release`, `ptyUnsubscribe` and `reclaim` go through the gate under test → `:328` *"puts every one of them through the gate, not only the ones with a call site"* (the frame table is asserted to equal the registry, so a seventh type added without an entry fails here too)

**The rest** (third bullet):

- [x] "stops sending it again if the Core comes back downgraded" → `:291`, same name
- [x] "asks for real once that Core comes back announcing the capability" → `:350`, same name
- [x] "self-refuses if authOk ever arrives first — the ordering is the guard" → `:374`, same name
- [x] "distinguishes 'no lock on this Core' from 'someone else holds it'" → `:426`, same name
- [x] "leaves the code undefined on an error that carries none" → `:456`, same name. Worth a note: this
  one was **not** wholly unpinned — `core-client.test.ts:189` already asserts an absent code on the
  Core's thrown-mutation path. Restored anyway, on the surface the original used (a `kill` the Core
  refuses with an uncoded `error`), because the review listed it and the two paths reach the field
  differently.

### `client-id-reclaim.test.ts` — "two assertions died"

- [x] "is the same string on every reconnect — the property the feature rests on" → `client-id-reclaim.test.ts:47`, same name
- [x] "carries on when the Core refuses the reclaim" → `:78`, same name

The nine the review found already re-covered are **not restored**, because a second copy of a passing
detector is not coverage. Each was re-read on this branch and is where the review says it is:

- [ ] on the wire once up — `core-client.test.ts:245`. Unchanged, still asserts the `clientId` on the reclaim frame.
- [ ] ahead of the replay tail and the PTY resend — `durable-core-client.test.ts:66`, `:328`.
- [ ] withheld until the bearer is accepted — `durable-core-client.test.ts:66` (`["auth", "reclaim", "subscribe"]`).
- [ ] caller-supplied id — `core-client.test.ts:252`.
- [ ] distinct per client — `packages/panel/src/server/services/__tests__/core-link-client-id.test.ts:18`.
- [ ] presented on `ready` when no bearer is configured — `core-client.test.ts:417` (the loopback path, `["reclaim", "replay"]`).
- [ ] withheld from a Core that announces nothing — `core-client.test.ts:315` for the degrade shape, and
  `multiconnection-gate.test.ts:328` now also pins that **no `reclaim` frame reaches such a Core at all**,
  which was the deleted file's "sends nothing, because that Core has already done it".

### `mtls-integration.test.ts` — "fully re-covered, and improved"

- [ ] wss:// with mTLS, then a round trip — `core-link-mtls.test.ts:164`. Nothing to restore.
- [ ] the pinned-cert refusal — `:178`. Nothing to restore.
- [ ] (net gain, no original) a refused handshake told apart from a Core that is not running — `:210`.

### Out of scope, deliberately

- [ ] The gate's *Panel-side* use — `panel-link/router.test.ts`, `panel-link/session-write-access.test.ts`,
  `shared/session-write-access.test.ts`, `services/core-link-manager.test.ts:250`. The review confirms
  these are covered and untouched, and this PR does not touch `packages/panel`.
- [ ] Sections 1, 3, 4 and 5 of the review. Approved as they stand; section 2 is the only one that
  asked for a follow-up.

Refs #153, #156, #195

### DIFF
--- a/packages/sdk/src/__tests__/client-id-reclaim.test.ts
+++ b/packages/sdk/src/__tests__/client-id-reclaim.test.ts
@@ -1,0 +1,113 @@
+// The stable Core client id, across the connections it exists to span (ADR 0024
+// D9, issue 146).
+//
+// The two detectors the Panel's `client-id-reclaim.test.ts` held and #156's
+// deletion did not replace — section 2 of #195's review names both. The rest of
+// that file came across with the migration: the id on the wire once the link is
+// up (`core-client.test.ts`), ahead of the replay tail and the PTY resend and
+// withheld until the bearer is accepted (`durable-core-client.test.ts`), the
+// caller-supplied id, and one id per client. What had no home is the property
+// the whole feature rests on and the one failure it has to survive:
+//
+//  1. **The id does not move.** It is the same string on every reconnect,
+//     because a reconnect is exactly what it spans. An id that changed with the
+//     socket would reclaim nothing, every time, and the defect would be
+//     invisible — the Core would answer politely and the ghost connection would
+//     hold this client's Sessions for the full 45-second heartbeat timeout.
+//     Safe by construction today (`readonly clientId`, assigned once, read at
+//     one send site), which is an argument for keeping it that way and not one
+//     for leaving it unpinned.
+//  2. **A refused reclaim is survivable.** It is fire-and-forget by design: the
+//     answer costs the heartbeat timeout it was shortening, never correctness,
+//     so a Core that refuses it must not take anything else down with it.
+
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { signBearer, verifyBearer } from "@actana/shared/core-link-bearer";
+import { DurableCoreClient } from "../durable-core-client";
+import { handDrivenDialer, startCoreRig, type CoreRig } from "./fake-core-link";
+
+const SECRET = "client-id-suite-secret-32-bytes-xx";
+const URL_A = "wss://core-a.test:9444";
+
+function bearerFor(coreId = "core_test"): string {
+  return signBearer({ coreId, exp: Date.now() + 60_000 }, SECRET);
+}
+
+describe("the SDK client's Core client id (ADR 0024 D9)", () => {
+  let rig: CoreRig | null = null;
+  let client: DurableCoreClient | null = null;
+
+  afterEach(() => {
+    client?.close();
+    client = null;
+    rig?.close();
+    rig = null;
+  });
+
+  it("is the same string on every reconnect — the property the feature rests on", async () => {
+    rig = startCoreRig({ authVerifier: (bearer) => verifyBearer(bearer, SECRET) });
+    const dial = rig.dialer();
+    client = new DurableCoreClient({
+      url: URL_A,
+      bearer: bearerFor(),
+      createSocket: dial.createSocket,
+      reconnectInitialMs: 5,
+      reconnectMaxMs: 5,
+      heartbeat: false,
+    });
+    const durable = client;
+
+    await durable.connect();
+    // Three connections, each a genuinely new socket the Core accepted and each
+    // presenting this client's id for the reaping of the one before it.
+    for (let connection = 1; connection <= 3; connection++) {
+      await vi.waitFor(() => expect(dial.pairs).toHaveLength(connection));
+      await vi.waitFor(() =>
+        expect(dial.last().client.framesOfType("reclaim")).toHaveLength(1),
+      );
+      if (connection < 3) dial.last().server.close();
+    }
+
+    const presented = dial.pairs.flatMap((pair) =>
+      pair.client.framesOfType("reclaim").map((frame) => frame.clientId),
+    );
+    expect(presented).toEqual([durable.clientId, durable.clientId, durable.clientId]);
+    expect(durable.clientId).toMatch(/^sdk-/);
+  });
+
+  it("carries on when the Core refuses the reclaim", async () => {
+    // Hand-driven, because the Core's reclaim handler cannot fail: there is no
+    // server path that answers this frame with an `error`, and the client's
+    // tolerance of one is otherwise asserted by nothing.
+    const dial = handDrivenDialer();
+    client = new DurableCoreClient({
+      url: URL_A,
+      bearer: bearerFor(),
+      createSocket: dial.createSocket,
+      reconnectInitialMs: 10_000,
+      reconnectMaxMs: 10_000,
+      heartbeat: false,
+    });
+    const durable = client;
+    const reclaimed = vi.fn();
+    durable.onReclaimed(reclaimed);
+
+    const connecting = durable.connect();
+    await vi.waitFor(() => expect(dial.last().framesOfType("auth")).toHaveLength(1));
+    dial.ready({ multiConnection: { version: 1 } });
+    dial.authOk();
+    await connecting;
+
+    expect(dial.last().framesOfType("reclaim")).toHaveLength(1);
+    dial.answerLast("reclaim", { type: "error", message: "nope" });
+
+    // The link is still a link: the next request goes out and is answered.
+    const found = durable.findByTask("task_1");
+    await vi.waitFor(() => expect(dial.last().framesOfType("findByTask")).toHaveLength(1));
+    dial.answerLast("findByTask", { type: "findByTaskResult", ptyId: "pty_1" });
+
+    await expect(found).resolves.toEqual({ ptyId: "pty_1" });
+    // And nothing was reported as reclaimed, because nothing was.
+    expect(reclaimed).not.toHaveBeenCalled();
+  });
+});
--- a/packages/sdk/src/__tests__/fake-core-link.ts
+++ b/packages/sdk/src/__tests__/fake-core-link.ts
@@ -18,7 +18,7 @@ import type {
 } from "@actana/core/pty-core-link-server";
 import { PtyCoreLinkServer } from "@actana/core/pty-core-link-server";
 import type { PtyCore, PtyCoreEvent } from "@actana/core/pty-manager";
-import type { CoreLinkEvent } from "../core-link-frames";
+import { CORE_LINK_PROTOCOL_VERSION, type CoreLinkEvent } from "../core-link-frames";
 import type { CoreLinkSocket, CoreLinkSocketFactory } from "../core-link-socket";
 
 type Listener = (...args: unknown[]) => void;
@@ -226,6 +226,77 @@ export class FakeEventLog implements EventLogPort {
   getLastEventId(): number {
     return this.seq;
   }
+}
+
+/**
+ * A dialer with no Core behind it: every dial hands back a socket whose inbound
+ * frames the test writes itself.
+ *
+ * Everything else in this file exists so a suite does not have to fake a Core,
+ * and that is the right default — a hand-rolled server answers whatever the
+ * client expects, which is the agreement worth testing. This is the exception,
+ * for the frames a real `PtyCoreLinkServer` cannot be made to produce:
+ *
+ *   - **`authOk` before `ready`.** The Core sends `ready` as frame one on every
+ *     connection, which is exactly why nothing in the client enforces the order
+ *     — so the only way to pin what that order buys is to invert it here.
+ *   - **an `error` answer to a `reclaim`.** The server's reclaim handler cannot
+ *     fail, and the client's tolerance of one that does is a `.catch()` no
+ *     reachable server path exercises.
+ *
+ * Reach for it only for those; a test a real Core can drive belongs on one.
+ */
+export type HandDrivenDialer = {
+  createSocket: CoreLinkSocketFactory;
+  /** Every socket handed out, in dial order — a reconnect is the next entry. */
+  sockets: FakeSocket[];
+  last(): FakeSocket;
+  /** Frame one on the current connection, with or without the capability. */
+  ready(opts?: { multiConnection?: unknown }): void;
+  /** The Core accepting this connection's bearer. */
+  authOk(opts?: { coreId?: string }): void;
+  /** Answer the last frame of `type` this client sent, correlated by its reqId. */
+  answerLast(type: string, frame: Record<string, unknown>): void;
+  /** The socket dies, as a Core restart or a reaped NAT flow does. */
+  drop(): void;
+};
+
+export function handDrivenDialer(): HandDrivenDialer {
+  const sockets: FakeSocket[] = [];
+  const last = (): FakeSocket => {
+    const socket = sockets[sockets.length - 1];
+    if (!socket) throw new Error("nothing dialed yet");
+    return socket;
+  };
+  return {
+    sockets,
+    last,
+    createSocket: () => {
+      const socket = new FakeSocket();
+      sockets.push(socket);
+      // A tick later, so the transport is fully wired before `open` fires and
+      // its `auth` frame goes out — the same order the real dialer gives it.
+      queueMicrotask(() => {
+        socket.readyState = 1;
+        socket.emit("open");
+      });
+      return socket.asClientSocket();
+    },
+    ready: ({ multiConnection }: { multiConnection?: unknown } = {}) =>
+      last().receive({
+        type: "ready",
+        version: CORE_LINK_PROTOCOL_VERSION,
+        ...(multiConnection === undefined ? {} : { multiConnection }),
+      }),
+    authOk: ({ coreId = "core_hand_driven" }: { coreId?: string } = {}) =>
+      last().receive({ type: "authOk", reqId: "auth-1", coreId, exp: 1_700_000_000_000 }),
+    answerLast: (type, frame) => {
+      const asked = last().framesOfType(type).at(-1);
+      if (!asked) throw new Error(`no ${type} frame on the wire to answer`);
+      last().receive({ ...frame, reqId: asked.reqId });
+    },
+    drop: () => last().close(),
+  };
 }
 
 export type CoreRig = {
--- a/packages/sdk/src/__tests__/multiconnection-gate.test.ts
+++ b/packages/sdk/src/__tests__/multiconnection-gate.test.ts
@@ -1,0 +1,480 @@
+// The multiConnection gate, from the client's side (ADR 0024 D11, issue 143).
+//
+// These are the detectors the Panel's `multiconnection-capability.test.ts` held
+// until #156 deleted the class they drove. The code they guard came across to
+// this package in #153 and is correct — #195's review read it line by line and
+// approved it — so nothing here is a bug report. What went missing is the
+// pinning: the capability's lifecycle on a client, and the contents of the
+// registry the gate consults, were asserted by nothing in this repo, which made
+// a regression in either of them green. Restored against the client the Panel
+// actually runs now, per section 2 of that review (issue 153, #156).
+//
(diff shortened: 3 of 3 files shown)


qcentic-dev:
## Checklist — every assertion section 2 of #195's review enumerates

Mapped to its new home test, checked or explicitly unchecked with the reason. Same list as the PR
body, posted here so the review that asked for the follow-up can be answered in the thread.



### `multiconnection-capability.test.ts` — "the real gap"

**The capability's lifecycle on a client** (review's first bullet):

- [x] closed before `ready` lands → `multiconnection-gate.test.ts:171` *"is closed before ready lands — the answer is unknown, not yes"*
- [x] recorded on announce → `:194` *"records the capability the Core announced on ready"*
- [x] not carried across a reconnect → `:203` *"does not carry it across a reconnect — it is re-read on the new ready"*
- [x] cleared the moment the link drops rather than when the next opens → `:219` *"closes the moment the link drops, not when the next one opens"*
- [x] picked back up when an upgraded Core reconnects → `:234` *"picks it back up when an upgraded Core reconnects"*
- [x] the load-bearing hazard, "does not queue it for a later flush" → `:249` *"does not queue it for a later flush"*
- [x] the load-bearing hazard, "refuses it in the drop window, and nothing flushes when the socket reopens" → `:265`, same name

**The frame registry's contents** (second bullet):

- [x] `MULTI_CONNECTION_ONLY_FRAME_TYPES` asserted — removing one or adding a seventh is no longer green → `:308` *"holds exactly the PTY subscription frames, the Session lock's three, and reclaim"*
- [x] `release`, `ptyUnsubscribe` and `reclaim` go through the gate under test → `:328` *"puts every one of them through the gate, not only the ones with a call site"* (the frame table is asserted to equal the registry, so a seventh type added without an entry fails here too)

**The rest** (third bullet):

- [x] "stops sending it again if the Core comes back downgraded" → `:291`, same name
- [x] "asks for real once that Core comes back announcing the capability" → `:350`, same name
- [x] "self-refuses if authOk ever arrives first — the ordering is the guard" → `:374`, same name
- [x] "distinguishes 'no lock on this Core' from 'someone else holds it'" → `:426`, same name
- [x] "leaves the code undefined on an error that carries none" → `:456`, same name. Worth a note: this
  one was **not** wholly unpinned — `core-client.test.ts:189` already asserts an absent code on the
  Core's thrown-mutation path. Restored anyway, on the surface the original used (a `kill` the Core
  refuses with an uncoded `error`), because the review listed it and the two paths reach the field
  differently.

### `client-id-reclaim.test.ts` — "two assertions died"

- [x] "is the same string on every reconnect — the property the feature rests on" → `client-id-reclaim.test.ts:47`, same name
- [x] "carries on when the Core refuses the reclaim" → `:78`, same name

The nine the review found already re-covered are **not restored**, because a second copy of a passing
detector is not coverage. Each was re-read on this branch and is where the review says it is:

- [ ] on the wire once up — `core-client.test.ts:245`. Unchanged, still asserts the `clientId` on the reclaim frame.
- [ ] ahead of the replay tail and the PTY resend — `durable-core-client.test.ts:66`, `:328`.
- [ ] withheld until the bearer is accepted — `durable-core-client.test.ts:66` (`["auth", "reclaim", "subscribe"]`).
- [ ] caller-supplied id — `core-client.test.ts:252`.
- [ ] distinct per client — `packages/panel/src/server/services/__tests__/core-link-client-id.test.ts:18`.
- [ ] presented on `ready` when no bearer is configured — `core-client.test.ts:417` (the loopback path, `["reclaim", "replay"]`).
- [ ] withheld from a Core that announces nothing — `core-client.test.ts:315` for the degrade shape, and
  `multiconnection-gate.test.ts:328` now also pins that **no `reclaim` frame reaches such a Core at all**,
  which was the deleted file's "sends nothing, because that Core has already done it".

### `mtls-integration.test.ts` — "fully re-covered, and improved"

- [ ] wss:// with mTLS, then a round trip — `core-link-mtls.test.ts:164`. Nothing to restore.
- [ ] the pinned-cert refusal — `:178`. Nothing to restore.
- [ ] (net gain, no original) a refused handshake told apart from a Core that is not running — `:210`.

### Out of scope, deliberately

- [ ] The gate's *Panel-side* use — `panel-link/router.test.ts`, `panel-link/session-write-access.test.ts`,
  `shared/session-write-access.test.ts`, `services/core-link-manager.test.ts:250`. The review confirms
  these are covered and untouched, and this PR does not touch `packages/panel`.
- [ ] Sections 1, 3, 4 and 5 of the review. Approved as they stand; section 2 is the only one that
  asked for a follow-up.

Refs #153, #156, #195

qcentic-adm (approved):
## Review — the follow-up §2 of #195 asked for

I asked for this one, so I checked it against my own spec rather than against the mapping that
answers it: every assertion §2 enumerates, read at the line the checklist points to, then read
against what the original guarded. **Approving.** The mapping is accurate line for line, the diff is
tests only, and the mutation table holds where I probed it. Two notes at the end, neither blocking,
both already half-disclosed in the body.

---

### 1. Tests only — verified mechanically, not read off the body ✅

`git diff --name-status 1e0b0b8..5eeb474` is three paths, all under `packages/sdk/src/__tests__/`:

```
A  packages/sdk/src/__tests__/client-id-reclaim.test.ts
M  packages/sdk/src/__tests__/fake-core-link.ts
A  packages/sdk/src/__tests__/multiconnection-gate.test.ts
```

665 insertions, 1 deletion. That single deletion is the `fake-core-link.ts` import line becoming a
value import (`import { CORE_LINK_PROTOCOL_VERSION, type CoreLinkEvent }`) because the new dialer
needs the constant. The rest of that file's diff is one added type and one added function,
`handDrivenDialer` — nothing existing is edited, so no existing suite's behaviour moves. No
`packages/sdk/src/*.ts` outside `__tests__`, no other package, no manifest, no lockfile, no config.
The claim in §2 that the code is correct still stands untouched by this PR, which is the right shape
for a detector-gap follow-up.

### 2. Every §2 assertion, at the line the mapping names ✅

All fourteen `it()` blocks in `multiconnection-gate.test.ts` and both in `client-id-reclaim.test.ts`
are at exactly the lines the checklist cites, and each asserts what the deleted original asserted —
not a same-named test that pins something weaker. The ones where that distinction had teeth:

- **`:219` "closes the moment the link drops, not when the next one opens."** The one I most expected
  to be fudged into "closes by the time the next connects", which is the assertion §2 explicitly said
  was *not* enough. It is not fudged: `reconnectMs: 10_000` puts the whole test inside the drop
  window, and `expect(dial.pairs).toHaveLength(1)` proves no second connection existed when the gate
  was read. That is the original's property exactly.
- **`:265` the drop-window hazard.** The reopen target is `coreThatAnnounces()`, so the frame *would*
  be allowed if asked for on connection two — which is what makes `sentAcross(pairs, "release") ===
  []` mean "nothing flushed" rather than "the gate refused it twice". The comment explaining why the
  wire assertion precedes the rejection assertion is correct and worth having: a frame that slipped
  the gate is pending, not rejected, so the rejection would fail as a timeout and say nothing about
  the wire.
- **`:308` / `:328` the registry.** `:308` is exact membership against a sorted array, so both
  directions §2 named (drop one, add a seventh) go red; it matches `core-client.ts:83-91` entry for
  entry. `:328` then asserts `GATED_FRAMES`' type set *equals* the registry before looping, which is
  what makes the loop self-updating — a seventh type added without a frame here fails at that
  assertion rather than being silently skipped. And the per-iteration
  `framesOfType(frame.type) === []` is the property §2 cared about: refused *before* the socket, so
  the Core is never asked to answer a frame it does not know.
- **`:328` also closes something the mapping claims for `client-id-reclaim`** — `reclaim` is in the
  loop, and against a non-announcing Core the assertion covers both the explicit request and the
  automatic `sendReclaim()` on establishment. That is the deleted file's "sends nothing, because that
  Core has already done it", genuinely restored.

The nine "already covered, not restored" and the three mTLS entries: I re-read every cited line on
this head rather than trusting my own earlier review. `core-client.test.ts:245` (with the `clientId`
assertion at `:259`), `:252`, `:315`, `:417` (the `["reclaim", "replay"]` ordering),
`durable-core-client.test.ts:66` (`["auth", "reclaim", "subscribe"]`) and `:328`,
`core-link-client-id.test.ts:18`, `core-link-mtls.test.ts:164` / `:178` / `:210` — all present, all
asserting what the mapping says. The `core-client.test.ts:189` note is right too: that line does
already assert an absent `code`, on the mutation path, and my §2 pointed at `:195` for the *coded*
sibling. Restoring the uncoded case anyway on the `kill` surface the original used is the correct
call, not padding — the two reach the field differently.

Not restoring the nine is right. A second copy of a passing detector is not coverage, and I would
have said so if it had gone the other way.

### 3. The mutation table — spot-verified by reading the guard out ✅

I did not re-run the mutations; I read each guard and worked out what the named test does when it
breaks. Four, two of them the lifecycle ones you asked me to weigh:

- **"No clear on `onClose`"** (drop `core-client.ts:448`). `FakeSocket.close()` propagates to the
  peer synchronously, so `dial.last().server.close()` runs the client's `onClose` before the next
  line of the test. With `multiConnection` left stale, `:229` reads `true` and fails immediately.
  `:265` fails differently and more interestingly: `established = false` is cleared on a separate
  line, so the mutation still queues the frame rather than sending it — it flushes on connection two
  and lands on the wire at `:286`. Both kills are real, and they are different failures, which is
  what a pair of detectors should look like.
- **"Capability remembered across connections"** (sticky read plus neither clear). `:203` reconnects
  to `coreThatDoesNot()`, whose `ready` carries no `multiConnection`; a sticky read keeps `{version:
  1}` and `:215` fails. This also confirms your finding 1 rather than contradicting it — with only
  `openTransport:402` removed, `onClose:448` has already cleared the field on every path that reaches
  a second connection, so the mutation needs both plus the sticky read to move anything. Recording
  that instead of letting a reader infer "either line is load-bearing" is the honest version.
- **"Establishment without `ready`."** `maybeEstablish` at `:505-508` requires `this.ready` *and*
  `transport.writable`; drop the former and connection two of `:374` establishes on `authOk` alone
  and sends `subscribe` — which `:406` asserts empty. The choice of `subscribe` for that assertion is
  the right one and the comment says why: it is not gated on the capability, so the test is about
  establishment rather than about the gate.
- **"PTY want dropped when the gate is closed."** `durable-core-client.ts:206-209` adds to
  `subscribedPtys` before delegating, so the want survives a connection that sent nothing;
  `resendPtySubscriptions` at `:188-192` then re-asks once the gate opens. Drop the want and `:367`
  finds no `ptySubscribe` on connection two. Kill confirmed.

The registry mutations are trivially true (`:318` exact array, `:332` set equality), and
`claim`-answers-`supported: true` fails `:449`.

### 4. CI ✅

Run `31608212075`, `head_sha 5eeb4742c444a600a29faa536fd51be1d9a27def` — the PR's current head, not
an earlier push. `conclusion: success`. Unit Tests pass (1m29s), plus Typecheck, Lint, Conventions,
Train rules, Secret Scan, Dependency Audit, E2E — Panel service seam, both installer E2Es, both image
smokes and both tarball smokes. The only skips are publish/train jobs that do not run on a PR.

---

### Two notes, neither blocking

1. **The refused-reclaim detector depends on vitest config, not only on the code.** You are right
   that a swallowed rejection has nothing to assert on and that the mutation therefore surfaces as an
   unhandled rejection failing the run — I checked `packages/sdk/vitest.config.ts` and there is no
   `dangerouslyIgnoreUnhandledErrors`, so it does fail today. Worth knowing that this one detector
   would go quiet if that flag were ever set, in a way none of the others would. Not worth changing
   the test for; worth a reader knowing.
2. **`multiconnection-gate.test.ts:456` sits under `describe("the Session lock over the gate")`** but
   drives `kill`, which is neither a lock frame nor a gated one. It is where §2 listed it, so this is
   filing, not coverage — move it whenever something else touches the file.

## Checklist

| # | What I checked | Verdict |
|---|---|---|
| 1 | Diff is tests only — no production code, no manifest, no config | ✅ 3 files, all `packages/sdk/src/__tests__/`; the 1 deletion is an import line |
| 2 | Existing suites unedited | ✅ only additive change to `fake-core-link.ts` (one type, one function) |
| 3 | Every §2 lifecycle assertion has a home at the mapped line | ✅ 7/7, and `:219` pins drop-time not next-open-time |
| 4 | Registry contents + all six types through the gate | ✅ `:308` exact membership matches `core-client.ts:83-91`; `:328` self-updating |
| 5 | The rest of §2 (downgrade, PTY transition, ready/authOk, both lock details) | ✅ 5/5 at the named lines |
| 6 | Both dead `client-id-reclaim` assertions | ✅ `:47` compares the id across three real connections; `:78` pins the `.catch()` |
| 7 | "Already covered, not restored" — the nine plus mTLS | ✅ re-read on this head; every citation holds |
| 8 | Mutation claims credible | ✅ four read out in full above, incl. two gate-lifecycle ones; all kill as claimed |
| 9 | CI green on the current head | ✅ run `31608212075` @ `5eeb474`, success |
| — | Unhandled-rejection detector is config-dependent | ⚠️ note only |
| — | `:456` filed under the lock `describe` | ⚠️ nit |

## Verdict

**Approve.** This closes 2c exactly as §2 asked, and it closes it in the form that was worth asking
for: rebuilt against `DurableCoreClient` and a real `PtyCoreLinkServer`, with a downgraded Core being
a second real server rather than a `ready` frame the test wrote, so the reconnect assertions are
about reconnects and not about a fixture. Hand-feeding is confined to the two shapes a real Core
cannot produce, and both are documented where the next reader will find them.

The part I want on the record is the discipline around the mutation table. Reporting that
`openTransport`'s `multiConnection = null` kills nothing on its own — when the easy thing was to let
a green suite imply every line is pinned — and reporting that the refused-reclaim detector fails the
run rather than an assertion are both statements against interest, and they are the two things that
make the rest of the table believable. That is the right instinct; keep it.

Draft status respected — not marking ready.

PR #197, opened 2026-08-12
refactor(sdk): phase 1 — the SDK extraction lands on the 0.2.1 train

### LINKED ISSUE #151: 1.1 Node-22 mTLS and fetch spike — freeze the transport shape (D12)

### LINKED ISSUE #152: 1.2 Move the protocol frames into packages/sdk (D4)

### LINKED ISSUE #153: 1.3 Extract CoreClient: transport and durable entry point (D1, D2, D6, D7)

### LINKED ISSUE #154: 1.4 Prompt delivery becomes a Core responsibility (D3)

### LINKED ISSUE #155: 1.5 SDK session layer: start, prompt, result, programmatic I/O (D2, D11)

### LINKED ISSUE #156: 1.6 Migrate the Panel onto the SDK — phase 1's success test (D1, D14)

qcentic-dev:
## Summary

Phase 1 of #129 — the extraction of `@actana/sdk` — as it stands at the tip of
`refactor/sdk-extraction`. Seven squash merges, six rounds plus a coverage follow-up, all reviewed
and merged onto the integration branch rather than onto the train. This PR is the phase's single
gate onto `beta/0.2.1`.

The shape of the phase: prove the transport, move the protocol, move the harness knowledge into the
Core, extract the client, build the session layer on it, and then delete the Panel's own copy and
run the Panel on the extracted one. The last of those is the success test, and it passed.

## The rounds

| PR | What landed |
| --- | --- |
| **#189** | `chore(spike)`: Node 22 speaks mTLS both ways — the D12 spike, throwaway `experiment/` only (#151) |
| **#190** | `refactor(sdk)`: the core-link protocol moves into `packages/sdk`; the cut now writes five manifests (#152) |
| **#191** | `feat(core)`: the Core delivers a session's starting prompt on the harness's own schedule (#154) |
| **#192** | `refactor(sdk)`: `CoreClient` and `DurableCoreClient` on one core-link transport (#153) |
| **#194** | `feat(sdk)`: `CoreSession` — start a Session, the Core delivers the prompt, read the result (#155) |
| **#195** | `refactor(panel)`: the Panel's core link **is** the SDK's durable Core client — phase 1's success test (#156) |
| **#196** | `test(sdk)`: the multiConnection gate's dead detectors, rebuilt on the SDK's durable client — the coverage follow-up §2 of the #195 review asked for |

**#190** moved `core-link-frames.ts` out of `@actana/shared` into the new private `@actana/sdk` at
`0.2.1`, repointing 90 importers, and extended `Train rules` to five manifests with a guard that
fails a sixth package by name instead of skipping it. **#191** replaced the flat 450 ms
`initialInput` timer with a real delivery sequence owned by the Core — wait for quiet, answer a
blocking dialog, write the prompt, send the carriage return as a separate keystroke — so a Panel, a
CLI and an SDK automation all send a string and behave identically (ADR 0026). **#192** extracted
the Panel's `core-link/client.ts` as two entry points on one transport, deliberately leaving the
Panel on its own copy so a regression in #156 would stay attributable. **#194** added the session
layer and a `node`-only runnable example. **#196** is tests only — no production code in the diff.

## The D12 spike verdict (#189, #151)

**D12 stands — keep `engines: ">=22"`.** Both legs succeed on Node 22.23.2 with CA, client
certificate and client key read from a registration blob against a live Core; Node 24.19.0 was run
as the control and produced identical results across all six legs (ws `ready`, bearer `auth`→`authOk`,
`fetch` over mutual TLS, refusal without a client cert, an HTTP response over mTLS, SAN failure on a
wrong host). Nothing fails on 22 that passes on 24, so there was no decision to escalate and
extraction was not blocked. Two corrections carried forward: `fetch` is undici and ignores
`options.cert`/`options.key` — the material goes in through a dispatcher `Agent` — and X.509 binds
hosts, not ports. One thing the spike left for a later ticket: the Core's mTLS port serves WebSocket
upgrades only, so a real Core HTTP route still does not exist.

## The success test (#195, #156, D14)

`packages/panel/src/server/core-link/client.ts` (1474 lines) and its `node-socket.ts` (86) are
**deleted, not deprecated** — `D` in `--name-status`, no barrel, no re-export, no shim — and
`CoreLinkManager` dials with `@actana/sdk`'s `DurableCoreClient`. The port (`CoreLinkClientLike`) is
unchanged, so the panel-link router and every fake see exactly what they saw; one adapter maps
`onProtocolVersion` onto the SDK's `onReady`. Reconnect backoff, heartbeat, request deadline, event
cursor, mTLS + bearer dial and the `panel-` client-id prefix are all carried across unchanged.

Result:

```
@actana/sdk     151 passed | 5 skipped
@actana/shared  180 passed
@actana/core    826 passed
@actana/panel   1338 passed
pnpm -r typecheck  clean
pnpm lint          0 errors
pnpm build         green; DurableCoreClient bundled into dist/server/server.js
```

The 122 Panel core-link-manager tests are unchanged and green; no test was edited to accommodate the
migration. `E2E — Panel service seam` passes on CI, covering both a browser reload (panel link
dropped mid-flight, replay carries what happened while nothing was attached, nothing at or below the
cursor resent, ids strictly ascending) and two Panel service restarts through one data directory.
The #195 review verified this independently rather than taking the report.

Three parity commits rode ahead of the migration, each from the #153 review: a request issued between
writability and `ready` is queued rather than sent; the reconnect backoff is armed before the
disconnect listeners run, closing an orphan-socket window; and the `terminal-screen` reverse-video
gap is documented and pinned by a test rather than fixed, because what a selection looks like in text
a published `screen()` hands somebody's script is the session layer's call.

## The two behaviour divergences, named

Neither is hidden and neither was patched over.

1. **The socket-factory throw — ruled an acceptable transient by the #195 review; deliberately not
   swallowed.** When the socket factory itself throws (malformed endpoint, TLS material the runtime
   refuses), the old client dialed inside its constructor and emitted `disconnected` synchronously,
   before `defaultCreateClient` returned — so the manager, which attaches `onDisconnected` after the
   constructor, missed the first report and the row sat at `connecting` until the first backoff
   attempt also failed. The SDK defers the same report by one microtask, the listeners are attached,
   and the Core reads `unreachable` at once. Same end state, same `detail` string, reached a beat
   sooner. The review's ruling: the old timing was not a behaviour but a dropped message, and
   restoring it would mean adding permanent, untested suppression logic to `CoreLinkManager` —
   *swallowing would be the behaviour change.* Every other transition is frame-for-frame identical.

2. **Pre-auth request queueing — strictly safer.** The old `trySend` gated on `!socket || !ready`, so
   a request issued after `ready` but before `authOk` went out on the wire. The SDK gates on
   `established` (`ready && transport.writable`), and writability requires `authenticated` on a
   bearer link — so that request is now queued instead. This is the correct behaviour: the Core
   answers a pre-auth frame with an error *and closes the socket*, so the old path killed the link.
   Sub-millisecond window, and it needs an operator action inside it.

## Related issues

Closes #151
Closes #152
Closes #153
Closes #154
Closes #155
Closes #156

Refs #148 — the phase-1 tracking issue; left open deliberately, since this PR lands the code and
whatever else #148 tracks for the phase is the user's call to close.

Refs #129

### Deferred follow-up

**#193 — `sanitizeInitialInput` drops newlines without a separator.** `packages/core/src/pty-manager.ts:244`
filters C0/DEL bytes and rejoins with `.join("")`, so a newline is removed with nothing in its place
and the text on either side is welded together (`review the diff\nthen open a PR` is delivered as
`…diffthen…`). Pre-existing, **not** a regression from #191 — confirmed by both reviewer and author
during that review — and deliberately deferred out of #191's scope, which covers delivery scheduling
and not the sanitiser's separator semantics. It rides into `beta/0.2.1` unfixed and is tracked
separately.

## Review gate

**This PR is the user's review gate.** It is not to be merged, approved or self-merged by anyone
else: it is reviewed by the user and merged only after their approval. Nothing here touches `main`
or the train branch directly — the merge into `beta/0.2.1` is the whole point and it happens on the
user's word.


qcentic-adm (approved):
**Train-level review.** Not a re-run of the seven per-PR reviews — those stand. This checks only what a phase-level integration can break that no ticket review could see: the shape of the range, the version surface, whether the success test survived the sum of its parts, and whether anything here reaches past `beta/0.2.1`.

**Approved.** All five integration invariants hold at their enforcing lines.

---

### 1. The range is exactly the seven squashes, linear from `384f9da`

`merge-base(beta/0.2.1, refactor/sdk-extraction)` = `384f9da` = the tip of `beta/0.2.1`, so the head is a fast-forward descendant and the base has not moved under it. `git rev-list --count 384f9da..3a902dc` = **7**; `--merges` = **0**; `--first-parent` = **7** — every commit single-parent, no merge, no stray.

Each SHA is the squash GitHub recorded at merge time, in order:

| commit | PR | parent |
| --- | --- | --- |
| `416fae5` | #189 | `384f9da` |
| `62acd54` | #190 | `416fae5` |
| `bac4b84` | #191 | `62acd54` |
| `9c02185` | #192 | `bac4b84` |
| `e83a2e6` | #194 | `9c02185` |
| `1e0b0b8` | #195 | `e83a2e6` |
| `3a902dc` | #196 | `1e0b0b8` |

That the seven SHAs still equal each PR's recorded `mergeCommit.oid` is the proof of no rewrite — a rebase or amend after those merges would have changed them. The fetch that brought the branch up was `384f9da..3a902dc`, fast-forward, not a forced update. No #193 in the sequence because #193 is an issue, not a PR — the deferred `sanitizeInitialInput` follow-up, correctly absent.

GitHub's own diff for this PR is 136 files, +10670/−3446, identical to the local `384f9da..3a902dc`.

### 2. Five manifests at `0.2.1`; `CORE_LINK_PROTOCOL_VERSION` untouched

All five read `0.2.1` at the head — `package.json`, `packages/{core,panel,sdk,shared}/package.json` — and all five are `private: true`. The only `"version"` line added or changed anywhere in the range is the new `packages/sdk/package.json` at `0.2.1`; no pre-existing manifest version was touched. `packages/*` contains exactly those four packages, so the set is closed.

The guard is real rather than nominal: `ci.yml`'s `train-rules` job now runs `assert_manifest_set` **before** `assert_versions`, and a sixth package under `packages/` fails by name ("A workspace package is outside the version set") instead of being skipped by a loop reporting green over a shrinking set. `promote.yml:732`'s cut writes the same five. `scripts/__tests__/workflows.test.mjs` pins all of it non-vacuously — the manifest list against the workspace rather than a count, and the job's `name:` against the contexts the rulesets actually require, so a rename that decouples the check from the ruleset fails the test rather than silently un-gating the train. ADR 0023 D3 is amended to match and says the count is derived, not declared.

`CORE_LINK_PROTOCOL_VERSION = "0.15.0"` — `packages/sdk/src/core-link-frames.ts:1257`, moved from `packages/shared/src/core-link-frames.ts:1250`. No `+`/`−` line in the whole range touches that assignment. The file crossed packages with a seven-line header comment added and nothing else changed.

### 3. The success test held

`packages/panel/src/server/core-link/client.ts` and `node-socket.ts` are `D` in `--name-status`. Nothing remains under `packages/panel/src/server/core-link/` but the Core's own server tests, and `CoreLinkManager` imports `DurableCoreClient` from `@actana/sdk`.

**The suite was not edited to pass.** Every modified test file in the range was checked for changed lines that are not import repoints. Exactly two have any:

- `scripts/__tests__/core-tarball.test.mjs` — the constant's path, `packages/shared` → `packages/sdk`.
- `scripts/__tests__/workflows.test.mjs` — purely additive, the five-manifest guard above.

Every other one of the ~35 modified test files is `@actana/shared` → `@actana/sdk` and nothing else. `core-link-manager.test.ts` — the 122 tests the migration had to survive — has a **one-line** diff, the import. No assertion moved.

The deletions are the deleted client's own tests, and their coverage is re-homed rather than dropped: `client.test.ts` → `core-link-server.test.ts` keeps every server-side suite verbatim and removes only the `PtyCoreLinkClient` describe; `mtls-integration` (2 cases) → `packages/sdk/.../core-link-mtls.test.ts` (3, gaining refused-handshake-vs-Core-not-running); `client-id-reclaim` and `multiconnection-capability` → the SDK's `client-id-reclaim`, `multiconnection-gate`, `core-client`, `durable-core-client` and `core-link-frames` suites. The panel's 31 gate cases land as 14 client-level plus the parser-level cases in `core-link-frames.test.ts:434/451/456` — absence-is-not-drift and unknown-version-falls-back-to-single-connection are pinned at the layer that now decides them. The registry test is strictly stronger than the one it replaced: `multiconnection-gate.test.ts:328` puts *every* registered frame through the gate, not only those with a call site. All 19 required checks pass, including `Unit Tests`, `Typecheck`, `Lint`, `Train rules` and `E2E — Panel service seam`.

**Two divergences, and I found no third.** Both are at the lines the description claims:

1. `packages/sdk/src/core-link-transport.ts:181` — the socket-factory throw reported via `queueMicrotask` after the constructor returns, so listeners are attached and the report is not dropped.
2. `packages/sdk/src/core-client.ts:669` — `if (!this.established || !transport?.writable)`, with `established` requiring `ready && writable` (`:507`) and writability requiring `authenticated` on a bearer link. The pre-auth request queues.

Everything else that could have diverged is identical by inspection rather than by assertion: the `CoreLinkClientLike` port is unchanged (the only hunks in `core-link-manager.ts` are the imports and the factory); `asPanelLink` is a one-method adapter mapping `onProtocolVersion` onto `onReady`; `mintPanelCoreClientId()` is the old `mintCoreClientId()` moved verbatim, same `panel-` prefix, same `randomUUID` and same fallback, per-instance and unpersisted exactly as before; the cursor storage shim is the same registry-backed pair. A repo-wide search for divergence language in the range's added code and docs turns up nothing but pre-existing upstream-provenance notes.

### 4. No docs-only commit; every ADR rode its code

All seven commits carry code. `416fae5` is the two throwaway `experiment/` spike files and nothing else — `pnpm-workspace.yaml` is `packages/*`, so it is outside the build and ships nowhere. ADR 0025 (new) and the 0016/0023 amendments ride `62acd54` alongside 97 code files; ADR 0026 (new) rides `bac4b84` alongside 5. No ADR landed alone, and no commit in the range is docs-only.

### 5. Nothing reaches `main`, the panel link, or #139's cadence

None of the seven commits is reachable from `origin/main` (`520cd22`); the base is `beta/0.2.1`, and `main` is an ancestor of it. The `.github` changes are confined to the `train-rules` job and the cut — they reach `main` only through a promotion.

The **panel link** is untouched. All twelve non-test files under `packages/panel/src/lib/` and `packages/panel/src/server/panel-link/` — `panel-link-client.ts`, `router.ts`, `session-lock-register.ts`, `panel-bridge.ts` and the rest — have **zero** changed lines that are not the `@actana/shared` → `@actana/sdk` import. #139's leading hypothesis is therefore exactly as testable after this lands as before.

**Reconnect cadence is byte-identical.** `core-link-manager.ts:99-100` still reads `RECONNECT_INITIAL_MS = 1_000` / `RECONNECT_MAX_MS = 30_000`, unchanged from the base and still passed explicitly, so the Panel's real backoff never used the 500 ms / 5 s defaults #139 cites and does not now. Those SDK defaults (`durable-core-client.ts:35-37`) match the old client's anyway. `DurableCoreClient` arms the heartbeat by default (`:101`, `heartbeat: opts.heartbeat ?? {}`) at `DEFAULT_HEARTBEAT_INTERVAL_MS = 15_000` / `DEFAULT_HEARTBEAT_TIMEOUT_MS = 45_000`, the same two numbers the deleted client used and the same 45 s reap ADR 0024 D7 depends on.

One piece of bookkeeping, non-blocking and not a change request: #139's body cites `packages/panel/src/server/core-link/client.ts` for the backoff figures, and that file no longer exists after this merges. The cadence is unchanged, so the diagnosis is unaffected — but the reference will need repointing at `packages/sdk/src/durable-core-client.ts` when someone picks #139 up, and it is worth a comment on the issue at merge time rather than a surprise later.

---

Approving as reviewer only. This does not merge and I am not merging it — the merge onto `beta/0.2.1` remains the user's call.

mrfqcentic (approved):
Second review of the phase gate, verified independently rather than from the body's report. **Approved** — with two notes at the end, one of which needs an action from you outside this PR.

**Range and provenance.** The diff over `beta/0.2.1` is exactly the seven squash commits of #189–#196, each of which was reviewed per-round before merging onto the integration branch (changes-requested → approval cycles on #189/#190/#191, straight approvals on the rest). Nothing unreviewed rides along. Base is the clean `beta/0.2.1` cut (`384f9da`) off the 0.2.0 `main` tip.

**Independent verification at the head (`3a902dc`).** Fresh worktree, Node 24: `@actana/sdk` 167 passed / 5 skipped, `@actana/shared` 180, `@actana/core` 826, `@actana/panel` 1338; typecheck clean; lint 0 errors. The panel count needed three full-suite runs locally — timing-sensitive panel-link tests (`terminals-in-browser`, `live-read-path`, `harness-install-from-picker`) flaked under parallel load, a *different* file each run, each passing in isolation, and CI is green on both runs of this head. Not a PR issue; if it ever shows on CI it earns a flake ticket. The success-test claim checks out structurally too: `core-link/client.ts` and `node-socket.ts` are hard deletions (no shim), the three obsolete test files went with them, and the five manifests all read `0.2.1` with `@actana/sdk` wired in as a private workspace package — matching the extended Train rules, which pass.

**Two named divergences.** Both rulings hold up: the deferred socket-factory report is the old dropped message surfaced a beat sooner, and pre-auth queueing is strictly safer than the old wire behaviour (the Core closes the socket on a pre-auth frame). Neither is hidden in the diff.

**Issues cross-checked.** #193 (`sanitizeInitialInput` welding newlines) is pre-existing, confirmed not-a-#191-regression in that review, and tracked — riding in unfixed is a documented decision, fine. #177 is about the out-of-repo experiment client, phase-2 material. #187/#186/#185 pre-date this phase and stay open deliberately. The `Closes #151–#156` set matches the merged rounds; leaving #148 open is the right call.

**Notes:**

1. **This is the first formal review on #197 itself.** The PR currently shows no reviews and no comments — the prior review coverage lives on the seven round PRs. If you were expecting a posted agent review here, it did not land.
2. **Two trains are open: `beta/0.2.1` and a stale `beta/0.3.0`.** The 0.3.0 branch is promote.yml's automatic next-minor guess after the 0.2.0 promotion, cut from the *0.2.0* main. The documented move for a patch release was to delete it and re-cut — only the re-cut happened. Consequence if it survives: the moment 0.2.1 promotes, `beta/0.3.0` is stranded (can never fast-forward), and `next-train` will see a train exists and skip cutting — leaving no live train for the 0.2.2 work you described. Recommend: delete `beta/0.3.0` before dispatching the 0.2.1 promotion; promote.yml then cuts a fresh `beta/0.3.0` from the new main, and you delete/re-cut it as `beta/0.2.2` per the documented patch flow.

Per the gate: merge stays on your word.

PR #198, opened 2026-08-12
chore(release): promote 0.2.1

qcentic-dev:
Promotion gate for the `beta/0.2.1` train, per [`docs/ci-cd.md`](docs/ci-cd.md).

## This is a gate, not a merge

**Do not press the merge button.** The approval and the green checks on this
pull request are the point. `promote.yml` performs the fast-forward of `main`,
and GitHub closes this PR as merged on its own once `main` has advanced onto
the train tip.

`main` advances only by fast-forward — not a squash, not a merge commit. A
squash would produce a `main` commit whose SHA differs from the tested one and
make the digest assertion unimplementable.

## What the train carries

**Phase 1 of the SDK extraction** (#148, integrated by PR #197): the Panel's
core-link client becomes `@actana/sdk` by extraction, and prompt delivery moves
into the Core.

| Ticket | PR |
| --- | --- |
| #151 — Node-22 mTLS and `fetch` spike (D12) | #189 |
| #152 — protocol frames into `packages/sdk` (D4) | #190 |
| #153 — `CoreClient`: transport and durable entry point | #192 |
| #154 — prompt delivery becomes a Core responsibility (D3) | #191 |
| #155 — SDK session layer (D2, D11) | #194 |
| #156 — Panel onto the SDK (D1, D14) | #195 |
| `multiConnection` gate coverage on the SDK's durable client | #196 |

No new features and nothing breaking: the Panel behaves identically,
`CORE_LINK_PROTOCOL_VERSION` does not move, and `@actana/sdk` is not published
to npm in this phase. ADR 0023 D31 makes the minor the breaking-change unit
pre-1.0, so a patch train is the honest number.

## Freeze

From this point until the promotion runs, nothing merges anywhere.

Refs #148, #129.

PR #199, opened 2026-08-12
chore(ci): commitlint body limits must fit concatenated squash titles

qcentic-dev:
## Why

The repo squash-merges with `COMMIT_MESSAGES`, so a train squash carries every ticket PR title concatenated into its message body. Those titles are capped at 112 characters plus a ` (#NNN)` suffix, which clears the 120-character `header-max-length` but not the 100-character body and footer limits — a title that is legal as a header becomes illegal the moment a squash quotes it.

The promotion PR for `beta/0.2.1` (#198) fails its Conventions check deterministically on `5eb2b3a`, whose body embeds PR #190's 105-character title:

```
* refactor(sdk): the core-link protocol moves into packages/sdk, and the cut writes five manifests (#190)
```

The parser reads that block as **footer**, so the rule that actually fires is `footer-max-line-length` — which had been inheriting `config-conventional`'s default of 100 rather than being set explicitly in this repo's config.

## What

`commitlint.config.mjs` only. `body-max-line-length` raised 100 → 132, and `footer-max-line-length` set explicitly to 132, with a comment recording the reasoning. Nothing else is touched.

## Verification

commitlint run locally over the promotion range `520cd22..5eb2b3a` with each config:

| config | result |
| --- | --- |
| current (`origin/beta/0.2.1`) | `✖ found 1 problems` — `footer-max-line-length` |
| this branch | `✔ found 0 problems, 0 warnings` |

This commit's own message was linted against the new config before it was written — subject 69 characters, longest body line 72.

### DIFF
--- a/commitlint.config.mjs
+++ b/commitlint.config.mjs
@@ -109,7 +109,14 @@ export default {
     // project's own history fails is a limit people learn to bypass. This one
     // catches a runaway subject without arguing with the house style.
     'header-max-length': [2, 'always', 120],
-    'body-max-line-length': [2, 'always', 100],
+    // 132, not 100, and the same for the footer: this repo squashes with
+    // COMMIT_MESSAGES, so a train squash carries every ticket PR title
+    // concatenated into its body. Those titles are capped at 112 plus a
+    // ` (#NNN)` suffix, and the parser may land the tail of that block in the
+    // footer, so both limits must clear the 120 `header-max-length` above —
+    // otherwise a title that was legal as a header is illegal once quoted.
+    'body-max-line-length': [2, 'always', 132],
+    'footer-max-line-length': [2, 'always', 132],
     'body-leading-blank': [2, 'always'],
     // Off in favour of `trailer-leading-blank` above it — same intent, without
     // mistaking a wrapped body sentence for a footer.


qcentic-adm (approved):
Approved. Verified independently rather than taking the description on trust.

**Diff is exactly what is claimed.** `git diff origin/beta/0.2.1...origin/pr199` is one file, `commitlint.config.mjs`, 8 insertions / 1 deletion: `body-max-line-length` 100 → 132, `footer-max-line-length` added at 132, plus the six-line comment. One commit (`08c61c5`), based directly on `5eb2b3a` — the current tip of `beta/0.2.1`. Nothing else touched, and `footer-max-line-length` appears exactly once in the file, so no shadowed duplicate.

**The failure reproduces, and the rule named is the rule that fires.** Ran commitlint over `520cd22..5eb2b3a` with each config, out of tree, against a clean install of `@commitlint/cli` + `config-conventional`:

- old config (`origin/beta/0.2.1`): `✖ found 1 problems, 0 warnings` — a single `footer-max-line-length` violation on `5eb2b3a`
- this branch: `✔ found 0 problems, 0 warnings`, exit 0

`520cd22` is the merge base of `main` and `beta/0.2.1` and is also `main`’s current tip, so that range is exactly what `--from base.sha --to head.sha` gives the Conventions job on #198. Measuring `5eb2b3a`’s message directly: 820 lines, exactly one over 100 characters — line 63, 105 characters, the `(#190)` bullet — and nothing over 132. The 27 characters of slack is what turns red into green, and the diagnosis of *which* rule fires is right: the body rule was never the one complaining.

**132 is the right number, not a round one.** A squash header reaching `beta/*` has already passed `header-max-length: 120` on the promotion PR’s own commit lint, and `COMMIT_MESSAGES` quotes it as `"* " + header` — 122 worst case, at column 0, since the nested bullets in `5eb2b3a` confirm quoting does not compound indentation across train generations. 132 clears that with headroom, and clears it even in the unenforced corner below.

**Raising both rules, not just the failing one, is the correct call.** Whether a quoted bullet lands on the body or the footer side of the parser’s boundary is an accident of where the trailer block starts. Fixing only the footer would move the next failure rather than end it, and the two limits drifting apart is what let this reach `100` by inheritance in the first place. Setting the footer explicitly also stops it silently tracking an upstream default.

**One adjacent gap, not this PR’s to fix.** The `Lint the PR title` step runs the bare title through the same `header-max-length: 120`, without reserving the ` (#NNN)` suffix GitHub appends at squash time. So a 120-character title passes title lint and then produces a 127-character header that fails commit lint on the promotion PR — which is where the 112 in the comment comes from: a derived cap that nothing actually enforces. It does not threaten this change (2 + 127 = 129, still under 132), and it is a header-side problem while this is a body/footer one. Worth a follow-up that subtracts the suffix width in the title step.

Checks green — 20 SUCCESS, 3 SKIPPED (`Resolve train tags`, `Panel image (train)`, `Core image (train)`, all expected off-train), no failures. The commit’s own message lints clean under the config it introduces.

Issue #157, opened 2026-08-11
2.1 CLI scaffold, blob registry, and the core noun (D8, D9, D10)


mrfqcentic:
**Base branch: `feat/cli-and-publishing`, not `main`.** Train: **`beta/0.2.2`**.
Worktree: **`wt-129-07`**. Suggested branch: **`feat/cli-scaffold-core-noun`**.
Phase 2, step 1. Implements **D8**, **D9**, and the `core` noun of **D10**.

## What

`packages/cli`: the `actana` binary's client half, the blob registry, and the `core` noun — `add`, `ls`, `use`, `rm`, `status`, `shell` (shell itself is a later ticket; scaffold the subcommand).

**Credential resolution, in order:** `--core <name>` → `ACTANA_CORE_BLOB` (single-Core mode) → the `current` pointer.

**Storage:** one blob per named Core at `~/.config/actana/cores/<name>.txt`, honouring `XDG_CONFIG_HOME`.

## Done when

- `actana core add` accepts a file or stdin. **Never `docker exec`** — a CLI that shells into a container to fetch its own credentials is not a CLI.
- `actana core ls --json` emits machine-readable output. Every list command in this phase does.
- `actana core status` reaches a real Core and reports its version.
- The published CLI carries client subcommands only; `actana daemon` stays in the Core package. One binary name, split by noun (D8).

## Traps

- **The SDK takes a blob object; config files are the CLI's job.** Do not let path resolution leak into `packages/sdk`.
- A blob is a credential. Mode the file 0600 and never log its contents, including in `--verbose`.
- `packages/cli` is the sixth manifest the train-cut workflow must write and the manifests-match check must assert. Extend both, and amend ADR 0023 D3 in whichever PR of this phase lands first.

Refs #129

qcentic-dev:
Delivered by #201, which landed on `beta/0.2.2` with the train integration PR #212 (squash merge `6a33b805461006ca8553fcba1141bdecdff07eba`). Closing manually — this PR's base was not the default branch, so it did not auto-close.

Issue #149, opened 2026-08-11
#129 phase 2 — CLI and the first npm publish (train beta/0.2.2)


mrfqcentic:
**Train: `beta/0.2.2`.** Integration branch: **`feat/cli-and-publishing`**, cut from `beta/0.2.2` after `0.2.1` promotes.

Phase 2 of #129: `@actana/cli`, and the first npm publish this repository has ever done.

## Why its own integration branch

#129 originally put phases 1 and 2 on one branch. They now merge into different trains, and an integration branch cannot straddle a promotion — trains and integration branches are never force-pushed (ADR 0023 D24), and the base moves underneath. So phase 2 starts fresh.

## Why a patch train

New published packages are additive; nothing existing changes shape. D31's minor-is-breaking rule is not triggered by adding a CLI.

## Steps

| Step | Parallel | Tickets |
| --- | --- | --- |
| 1 | 3 | CLI scaffold + blob registry + `core` noun ∥ freeze the prototype ∥ npm publish plumbing + ADR 0018 amendment |
| 2 | 3 | `session` noun ∥ `project`/`harness`/`events` nouns ∥ `core shell` |
| 3 | 1 | `session attach` |
| 4 | 1 | parity check, then delete the frozen prototype |

## Done means

`npm i -g @actana/cli` gives a working `actana` against a remote Core, every list command emits `--json`, and the frozen prototype is gone.

## Dependency outside this phase

`session attach` needs the Session write lock (#144, train `beta/0.2.0`). That ordering is automatic — `0.2.0` promotes before this train opens — but the ticket must claim the lock explicitly rather than assume it holds the only connection. #129 D5's original rationale ("the Core accepts one connection at a time") is dead by then; the decision it justified still stands.

## Rules for every ticket here

- Base branch is `feat/cli-and-publishing`.
- One worktree per ticket, on the Dev Core.
- Squash-merge; ADRs ride the code. **No docs-only PRs.**
- At the end of the phase, one PR: `feat/cli-and-publishing` → `beta/0.2.2`.

Refs #129

qcentic-adm:
Done — the 0.2.2 train is out.

- Promoted as **v0.2.2**; `main` is at 6a33b80.
- `@actana/sdk@0.2.2` and `@actana/cli@0.2.2` are published to the public npm registry with provenance — both carry an npm publish attestation and `https://slsa.dev/provenance/v1`.
- Integration PR #212, promotion PR #213.

Closing as done.

[...]

PR #200, opened 2026-08-12
ci(release): npm publish with provenance, engines floor, and the ADR 0018 amendment

### LINKED ISSUE #159: 2.3 npm publish with provenance, engines floor, ADR 0018 amendment (D12, D13)
Worktree: **`wt-129-09`**. Suggested branch: **`ci/npm-publish-provenance`**.
Phase 2, step 1. Implements **D12** and **D13**. **Amends ADR 0018.**

## What

The first npm publish this repository has ever done. `@actana/sdk` and `@actana/cli` publish from `release.yml`, on the same tag that builds the images, at the same version as everything else.

## Done when

- Both packages publish with provenance, in lockstep with the container images.
- **A missing `NPM_TOKEN` fails the release loudly**, in the `resolve` job, before anything is built — the same shape as the existing Docker Hub secret check.
- Published packages declare `engines: ">=22"`, ship compiled JS plus `.d.ts`, and have no `preinstall` guard.
- **`scripts/require-node-24.mjs` cannot reach a published tarball.** Assert it, do not merely arrange it — the monorepo, Core and Panel keep `>=24 <25`, and the guard is correct for them.
- **ADR 0018 is amended in this PR**: Docker Hub is no longer the only registry. The #107 rule — amend the ADR, never settle it in a comment.

## Traps

- A first publish cannot be undone or reused: an npm version number is burned even if unpublished within the window. Rehearse against a scoped dry-run before the tag exists.
- `release.yml`'s macOS leg waits on a human approval in the `macos-release` environment. Anything downstream of that approval — and a publish must be downstream of it — sits in `Waiting`, which reads as a hang and is not one. See `.agents/skills/release/SKILL.md`.
- Provenance requires the workflow to be the publisher, with `id-token: write`. A publish from a job without it succeeds and is silently unattested.

Refs #129

qcentic-dev:
## Summary

The first npm publish this repository has ever done. `@actana/sdk` and `@actana/cli` publish from `release.yml`, on the same tag that builds the images, at the same version as everything else (#129 D12, D13), with provenance. **ADR 0018 is amended in this PR** — Docker Hub is no longer the only registry.

The SDK becomes publishable without changing what it is inside the workspace. Its `exports` map still points at TypeScript source, because a plain `node script.mjs` resolving `@actana/sdk/…` through Node's own type stripping is a property worth keeping; `publishConfig.exports`, which **pnpm** applies at pack time and npm does not, is what turns it into the compiled `dist/` map a registry consumer resolves. One source tree, two maps, and the rehearsal asserts the packed one no longer points at `src/`.

The shape of the whole change is set by one fact that does not hold anywhere else in this pipeline: **an npm version number is consumed by its first publish.** Unpublishing inside the 72-hour window frees the bytes and not the name. A container tag is a pointer — the entire promote mode rests on re-pointing one — and none of that transfers. So the publish is last in the graph, it is rehearsed on every pull request before a tag exists, and its provenance is read back off the registry rather than trusted.

## Related issues

Closes #159
Refs #129, #149

## Type of change

- [x] 📚 Documentation
- [x] 🔧 Build / CI / tooling

## What changed

**`packages/sdk`** — not private, `engines: ">=22"`, `files: ["dist"]`, `publishConfig` carrying `access: public`, `provenance: true` and the compiled `exports` map. `tsconfig.build.json` emits JS + `.d.ts` + maps with `rewriteRelativeImportExtensions`, so the `.ts` specifiers the source carries become `.js` in the emit. `prepack` runs the build. No `preinstall`, and no lifecycle script of any kind. A `README.md`, because npm renders one out of the tarball.

**`release.yml`** — a `npm` job with `permissions: id-token: write`, downstream of both tarball legs, the installer e2e and both image publishes; `github-release` now waits on it. A missing `NPM_TOKEN` fails in `resolve` before anything is built, in the same shape as the Docker Hub check. An already-published version on the re-run path is a notice, not a 403.

**`scripts/lib/npm-packages.mjs` + `scripts/rehearse-npm-publish.mjs`** — pack every publishable package and assert the tarball. The publish set is *discovered* (every workspace manifest without `private: true`) and checked against D13's intent both ways, so the Panel becoming publishable and the SDK silently not publishing are both errors.

**ADR 0018** — amended, with `docs/ci-cd.md`, `docs/REPO_SETUP.md`, the ADR index and the release skill brought into step.

## How was this tested?

- [x] Manually verified

**`scripts/__tests__/npm-publish.test.mjs`** (23 tests) — the rules as pure functions over the manifests that would be *wrong* (a `preinstall`, `prepare` or `postinstall`; the monorepo's `>=24 <25` floor; a map still pointing at `src/`; a missing `types` condition; restricted access; no provenance; a version out of lockstep; a stray file; a `.js` with no `.d.ts`), plus the real thing: the script packs `@actana/sdk`, and the test re-reads the artifact and asserts on the bytes.

> **#159 asks that `scripts/require-node-24.mjs` "cannot reach a published tarball" be asserted, not merely arranged.** It is asserted twice, in the file list of an actual `pnpm pack` output — once by basename, and once by a whitelist, so the rule survives a rename of the guard. The monorepo's own half is asserted in the same place: the root manifest keeps `>=24.0.0 <25` and its `preinstall`, because a check that only read the tarball would pass just as happily if somebody had "fixed" the root for consistency.

**`scripts/__tests__/workflows.test.mjs`** (+9 tests, 47 total) — `id-token: write` on the publish job and on no other job in the file; `--provenance` on the command line; the `NPM_TOKEN` check in `resolve`; the `needs:` ordering; the token out of reach of every other job, `github-release` (which holds `contents: write`) included; the attestation read-back; and that the workflow packs through the rehearsal script rather than `npm pack`. Both provenance assertions were **mutation-tested**: deleting `id-token: write` and dropping `--provenance` each turn the suite red.

**The rehearsal itself, locally** — `pnpm npm:rehearse` packs `@actana/sdk@0.2.2` and runs `npm publish --dry-run` over the tarball. **39 files** — nine modules × (`.js`, `.d.ts`, `.js.map`, `.d.ts.map`) = 36, plus `package.json`, `README.md` and `LICENSE`; `engines >=22`; no lifecycle scripts; the packed `exports` map rewritten to `dist/`. Every one of the nine compiled modules was imported under a plain `node`. (An earlier revision of this body and the acceptance comment both said 38, which was arithmetic that only worked if `LICENSE` were absent from the tarball. It is present — pnpm copies the workspace root's in at pack time — and `assertPackedFiles` now requires it rather than leaving it to chance.) **Nothing was published, and no tag exists.**

`pnpm typecheck` and `pnpm lint` clean (0 errors; the new files add no warnings). Full suite green apart from two pre-existing, environment-caused failures unrelated to this change: `core-launcher.test.mjs` fails identically on the untouched `cdfcb7a` checkout (this machine has a real `/opt/actana` install), and one Panel `live-read-path` test hit its 5s timeout under load and passes on its own.

## Notes for the reviewer

**The `@actana/cli` leg is wired and untested.** #157 creates `packages/cli` in parallel and owns it; this PR does not create it, and touches neither `promote.yml`'s manifest list, `ci.yml`'s `train-rules`, nor ADR 0023. Because the publish set is discovered rather than listed, the CLI publishes with no further edit here the moment #157 lands a non-private manifest — and until then the rehearsal *reports its absence* rather than passing over it, so "the CLI leg never ran" cannot read as "the CLI leg passed".

**The `macos-release` pause named in #159's traps is no longer in this file.** ADR 0023 D15 moved it to the head of `promote.yml`, upstream of this whole workflow, so "downstream of the human" is now a property of being in `release.yml` at all — and `scripts/__tests__/workflows.test.mjs` already asserts no approval environment remains here. The ordering the trap is really about is preserved and asserted: `npm` waits on `tarball-macos` along with everything else.

**`release.yml`'s `resolve` does not verify that `@actana` scope ownership is real** — #129 flags it as unverified, and no check can distinguish "token is valid" from "token can publish this scope" without publishing. `NPM_TOKEN` exists on the repository (set 2026-08-12); `npm access list packages @actana` is the human check, and it is written into `REPO_SETUP.md` §2.

**`.agents/skills/release/SKILL.md` has pre-existing ADR 0023 drift** — it still describes the tag trigger and the `macos-release` environment on `tarball-macos`. I added the `NPM_TOKEN` precondition and the burn-once rule and deliberately left the rest alone; unwinding that drift is not this ticket's, and doing it here would collide with whoever owns it.

## Checklist

- [x] This PR targets `feat/cli-and-publishing` (phase 2's integration branch, per #149), not `main`
- [x] I updated documentation (ADR 0018, `ci-cd.md`, `REPO_SETUP.md`, the ADR index, the release skill)
- [x] Dependent changes: none blocking — the `@actana/cli` leg waits on #157, and this PR is a draft


qcentic-dev:
## Acceptance criteria — #159, item by item

Every "Done when" clause from the issue, each checked or explicitly unchecked with the reason.

---

### ☑︎ 1. Both packages publish with provenance, in lockstep with the container images

**`@actana/sdk` — checked.** `release.yml`'s `npm` job publishes it on the same tag that builds the images, at the version `resolve` read off that tag, with `--provenance` from a job carrying `id-token: write`. Lockstep is asserted rather than assumed: the rehearsal refuses to publish a package whose manifest is not the version being released.

**`@actana/cli` — wired, and deliberately unchecked as *exercised*.** #157 creates `packages/cli` in parallel and owns it, so it does not exist on this branch and nothing here has packed it. The publish set is *discovered* — every workspace manifest without `private: true` — so the CLI publishes with no further edit to `release.yml` or to the rehearsal the moment #157 lands a non-private manifest. Its absence is **reported by name** on every run rather than passed over, so "the CLI leg never ran" cannot read as "the CLI leg passed".

**The first real publish has not happened, and could not have.** It happens on the next promotion, which is not this PR's to trigger. What is checked here is that the path is complete, asserted, and rehearsed end to end against real bytes.

### ☑︎ 2. A missing `NPM_TOKEN` fails the release loudly, in the `resolve` job, before anything is built — the same shape as the existing Docker Hub secret check

Checked. `Require an npm token` sits immediately after `Require Docker Hub credentials` in `resolve`: same emptiness test, same `::error title=…::` with an actionable sentence, same `exit 1`. Deliberately a second step rather than a merged "credentials" check — "a credential is missing" is not actionable when there are two.

Being *there* is the requirement and it is load-bearing: `npm` is the last job in the graph, so a token discovered missing at the publish would be discovered with both images pushed and their `:latest` already moved. `scripts/__tests__/workflows.test.mjs` asserts the check is in `resolve`, that it matches the Docker Hub shape, and that every publishing job actually waits on `resolve`.

### ☑︎ 3. Published packages declare `engines: ">=22"`, ship compiled JS plus `.d.ts`, and have no `preinstall` guard

Checked, and asserted on the **packed artifact** rather than on the manifest that produced it. `pnpm npm:rehearse` on this branch produces `@actana/sdk@0.2.2`: 38 files, `engines >=22`, every `dist/*.js` with a `.d.ts` beside it, and no lifecycle script of any kind in the published `scripts` block.

The refusal covers `preinstall`, `install`, `postinstall` **and** `prepare` — all four reach the consumer's machine during `npm install`, and "no `preinstall`" alone would be satisfied by renaming one. Each is exercised with a manifest that would be wrong.

Compiled JS is what makes `>=22` mean anything: a consumer on Node 22 has no type stripping. Inside the workspace the SDK stays TypeScript source; `publishConfig.exports` reconciles the two, and the rehearsal asserts the *packed* map no longer points at `src/` — which is also what makes `pnpm pack` (not `npm pack`) load-bearing rather than a preference.

### ☑︎ 4. `scripts/require-node-24.mjs` cannot reach a published tarball — **asserted, not arranged**

Checked, in the file list of a real `pnpm pack` output, two independent ways:

1. **By name** — no packed entry has that basename.
2. **By whitelist** — the packed file list must be `dist/*.{js,d.ts,*.map}`, `package.json`, and `README`/`LICENSE`/`NOTICE`/`CHANGELOG`. Nothing else. So the rule survives a rename of the guard, a new `files` entry, or any lifecycle script that would drag `scripts/` in — the failure modes a basename check would pass.

And the other half of D12, asserted in the same place: the monorepo keeps `>=24.0.0 <25`, keeps its `preinstall`, and keeps `scripts/require-node-24.mjs`. A check that only ever read the tarball would pass just as happily if somebody had "fixed" the root manifest to `>=22` for consistency — which would drop the Core and the Panel onto a runtime neither is built or tested for.

### ☑︎ 5. ADR 0018 is amended in this PR — Docker Hub is no longer the only registry

Checked. `docs/adr/0018-docker-hub-is-the-only-registry.md` gains an amendment section: read the title as "the only registry the **images** go to". Nothing above it is reversed and GHCR stays retired — GHCR and Docker Hub were two copies of the same bytes, and npm is not a copy of anything but a different artifact for a different audience.

The amendment records what does *not* carry over: a container tag is a pointer and the whole promotion mechanism rests on re-pointing one; an npm version number is consumed by its first publish. `docs/ci-cd.md`, `docs/REPO_SETUP.md` (§2, with `NPM_TOKEN`), the ADR index and `.agents/skills/release/SKILL.md` are brought into step in the same PR. The #107 rule — amend the ADR, never settle it in a comment.

---

## The traps, each answered

### ☑︎ A first publish cannot be undone or reused — rehearse before a tag exists

`scripts/rehearse-npm-publish.mjs` packs every publishable package and asserts the tarball. It runs in **two** places and the earlier one is the point: on every pull request through `pnpm test` (CI's `Unit Tests` job), long before any tag; and again in `release.yml` on the tarballs it then publishes, so what was asserted and what goes to the registry are the same bytes rather than the same intent.

Rehearsed locally on this branch with `pnpm npm:rehearse` — `pnpm pack` plus `npm publish --dry-run` over the resulting tarball, and the compiled entry point imported under a plain `node` from outside the workspace. **Nothing was published and no tag exists.**

### ☑︎ The publish sits downstream of the human approval

Preserved, with a correction the issue's own framing predates: **`release.yml`'s macOS leg no longer carries the `macos-release` environment.** ADR 0023 D15 moved the pause to the head of `promote.yml`, upstream of this entire workflow — so "downstream of the approval" is now a property of being in `release.yml` at all, and the existing test suite already asserts no approval environment remains in the file.

The ordering the trap is really about is intact and asserted: `npm` waits on `tarball-macos` along with both tarball legs, the installer e2e and both image publishes. It is last on purpose — everything else in a release can be redone, and a version number cannot.

### ☑︎ Provenance requires `id-token: write`, and a publish without it is silently unattested

Three overlapping guards, because the failure is a *silent success*:

1. `id-token: write` on the `npm` job — and `workflows.test.mjs` asserts it is on that job and on **no other job in the file**.
2. `--provenance` on the command line, plus `publishConfig.provenance: true` inside the package, asserted by the rehearsal.
3. The publish **reads the attestation back off the registry** afterwards and fails the release if it is absent — the only check that can catch a publish that already succeeded unattested.

Both workflow assertions were **mutation-tested**: deleting `id-token: write` and dropping `--provenance` each turn `workflows.test.mjs` red.

### ☑︎ `packages/cli` is #157's — boundaries held

`packages/cli` is not created. `promote.yml`'s manifest list, `ci.yml`'s `train-rules` and ADR 0023 are untouched — `Train rules` passes on this PR against the unchanged five manifests.

---

## One thing no check can cover

**`@actana` scope ownership is not verified by anything here.** #129 flags it as unverified, and no automated check can distinguish "the token is valid" from "the token can publish this scope" without publishing. `NPM_TOKEN` exists on the repository (set 2026-08-12) and was neither created nor renamed by this PR. `npm access list packages @actana` is the human check, and it is written into `REPO_SETUP.md` §2 rather than left in this comment.

qcentic-adm (changes requested):
Reviewed against #159 in full and the acceptance-criteria comment, diffed against `feat/cli-and-publishing` at `cdfcb7a` (merge-base confirmed clean — no drift, no unrelated files).

This is very good work. Every "Done when" clause in #159 is met, most of them at the enforcing line rather than in prose, and the two disclosed judgment calls are both defensible. I am requesting changes for one thing: **the publish reproduces, on a third surface, the exact defaulting bug this file already legislates against in its own comments** — and for a second, smaller thing about the attestation read-back that can fail a release for the wrong reason at the most expensive moment in the graph.

Note on method: `pnpm` is not installed in my environment, so I did not execute the suite. Everything below is read off the source. Where I say a mutation would be caught, I mean the assertion is an exact-string `toContain` or an anchored regex over the workflow text, so the claim is checkable by inspection rather than taken on trust.

---

## Blocking

### 1. `npm publish` takes the `latest` dist-tag by default, and nothing in the file stops a backport or an RC from moving it

`.github/workflows/release.yml:725`

```yaml
npm publish "$tarball" --provenance --access public
```

No `--tag`. npm defaults the dist-tag to `latest`.

This file already contains the argument against exactly that, 96 lines further down, about a different registry — `.github/workflows/release.yml:829-836`:

> `gh release create` was called here with no `--latest` flag at all, and the API defaults `make_latest` to **true**. That default is what makes […] `--latest` and `--prerelease` are therefore never defaulted.

And `resolve` already computes the answer. `.github/workflows/release.yml:168-169` outputs `latest` and `prerelease`; `.github/workflows/release.yml:270-292` is a belt-and-braces step whose own error text is *"A backport must never move latest"*, asserted on the docker tag list and on `--latest` for the GitHub Release. The `npm` job reads neither output — I grepped the whole job block for `latest` and `prerelease` and there is not one reference.

Both reachable paths exist and are documented as supported:

- **Backport.** `workflow_dispatch` on a `release/0.1` tag, `mode=backport` (`release.yml:222-233`). The `npm` job carries no `if:`, so `@actana/sdk@0.1.5` publishes and takes `latest`. `npm i @actana/sdk` then hands every consumer a downgrade — the third rendering of the D28 failure whose other two renderings this job sits beside.
- **Prerelease.** `resolve` accepts `v1.0.0-rc.1` (`release.yml:201`), D30 says that is how a backport's release candidate goes out, and `scripts/lib/release-latest.mjs:196-206` classifies it. On npm it publishes as `latest`.

A dist-tag is repairable — `npm dist-tag add` — and that is the one respect in which this is *less* severe than a burned version number. But it reaches every consumer silently, which is precisely the property D28 is written about, and the recovery requires somebody noticing. Given how carefully the rest of this PR treats the other two surfaces, I do not think npm should be the one that defaults.

The fix looks small: derive the tag from `needs.resolve.outputs.latest` / `.prerelease`, pass it explicitly, and add the third arm to the existing "A backport moves latest on neither surface" step at `release.yml:270` so the guard covers all three surfaces in one place. `scripts/__tests__/workflows.test.mjs` is already the right home for a `not.toMatch(/npm publish[^\n]*(?<!--tag [^\n]*)/)`-shaped assertion, or more simply an assertion that the job reads `needs.resolve.outputs.latest` at all.

---

## Should fix before this leaves draft

### 2. The attestation read-back cannot tell "unattested" from "could not ask", and runs without the token

`.github/workflows/release.yml:738-761`

The step has no `env:` block. The publish step immediately above it (`release.yml:711-712`) sets `NODE_AUTH_TOKEN`, and `setup-node` wrote an `.npmrc` at `release.yml:669-675` whose `_authToken` line references that variable. With the variable unset for this step, npm sends the unexpanded literal as a bearer token to `registry.npmjs.org`, which is a well-known route to `E401` on commands that would otherwise need no auth.

Then `release.yml:749`:

```bash
predicate="$(npm view "$spec" dist.attestations.provenance.predicateType 2>/dev/null || true)"
```

`2>/dev/null || true` collapses *every* failure — 401, rate limit, DNS, a bad spec — into the same empty string as a genuinely missing attestation. Five attempts and 50 seconds later the step emits:

> The version cannot be republished; the fix is to restore the attestation path and cut the next version.

That advice is correct if the attestation is really absent and catastrophically wrong if npm simply could not be reached — and it arrives *after* the version number has been burned, which is the one place in this whole design where a false alarm is expensive. The read-back is the right idea and I would keep it; it needs to distinguish the two cases. Capture the exit status separately from the output, retry only on a non-zero `npm view`, and fail with a different message when the registry could not be queried at all. Setting `NODE_AUTH_TOKEN` on the step costs one line and removes the most likely cause.

### 3. The "Node can actually import" smoke test reaches 5 of the 9 published modules — and not the two the README tells people to import

`scripts/__tests__/npm-publish.test.mjs:262-267` imports `packages/sdk/dist/core-client.js` only. The comment is exactly right about why the test exists ("a stray extensionless relative specifier is all it takes, and `moduleResolution: Bundler` will not complain about one") — and the import graph is:

```
core-client.js → core-link-frames, core-link-socket, core-link-transport, core-registration-blob
```

Unreached: `core-session`, `durable-core-client`, `terminal-screen`, `core-link-cursor-storage`. `packages/sdk/README.md:25-26` and `packages/sdk/examples/start-a-session.mjs:27-28` both import `@actana/sdk/core-session`, and the README's own bullet list points at `@actana/sdk/durable-core-client`. The module most likely to be a consumer's first `import` is the one the test does not cover.

Today every relative specifier in `packages/sdk/src` carries `.ts` — I checked all nine files, there is not one extensionless relative import — so this is a hole in the net rather than a live bug. But the whole point of that test is the specifier somebody adds *later*. Iterating the packed `dist/*.js` list and importing each would close it, and it is the natural thing to do inside the loop that already has the file list.

Secondary, and a genuine ordering dependency: this test asserts `fs.existsSync(entry)` with the message *"run the packing test first — it builds dist/"*. Vitest runs an `it` in file order so it holds today, but the test depends on a sibling's side effect. Building `dist/` in a `beforeAll`, or importing out of the extracted tarball rather than out of the workspace, would make it stand on its own.

---

## Documentation the PR claims to have brought into step, and did not quite

The PR body says `docs/ci-cd.md` was updated. The Registries section was, thoroughly and well. Three places downstream of it were not, all in the section that is a reader's actual walkthrough of this workflow:

- **`docs/ci-cd.md:999-1001`** — "The two Linux tarball legs and the installer e2e run straight away, the mac tarball builds alongside them, and **the two image jobs and the GitHub Release follow**." This PR adds a job between those two and it is not here. This is the sentence a person reads to learn what `release.yml` does.
- **`docs/ci-cd.md:1021-1023`** — "`resolve` does fail the run outright when `DOCKERHUB_USERNAME` or `DOCKERHUB_TOKEN` is missing […] — that one the workflow does check." As of this PR `resolve` checks a second secret in the same shape, and this is the paragraph where a reader is told which secrets the workflow checks for them.
- **`docs/ci-cd.md:1036-1037`** — "**Nothing leaves the repository until a reviewer approves.** No image moves, no `:latest` moves, no GitHub Release appears, and `main` does not advance." The npm publish is now part of that list, and it is the item on it that cannot be taken back.

Each is one clause. I raise them because the PR's own standard is that a mechanism and its account move together, and because the ADR amendment is careful about precisely this.

---

## Smaller notes, none blocking

- **`scripts/lib/npm-packages.mjs:86-90`** — `^package\/dist\/[^/]+\.(js|d\.ts|js\.map|d\.ts\.map)$` permits only the top level of `dist/`. The day `src/` grows a subdirectory the rehearsal fails with "outside `dist/` and its paperwork", which reads as a leak rather than as a flat-layout assumption. Failing closed is the right instinct; a sentence in the comment above `ALLOWED_ENTRIES` saying the flat layout is deliberate would save the next person the bisect.
- **The 38-file count** in the PR body and in the acceptance comment. Nine non-test modules × (`.js`, `.d.ts`, `.js.map`, `.d.ts.map`) = 36, plus `package.json` and `README.md` = 38 — which lands on your number only if `LICENSE` is *not* in the tarball. `packages/sdk/` has no `LICENSE` of its own; pnpm copying the workspace root one is a pnpm behaviour rather than a guarantee, `ALLOWED_ENTRIES` permits a LICENSE but nothing requires one, and an MIT package on npm with no licence text in the tarball is a real if minor defect. Worth re-reading the pack output, and worth an assertion either way.
- **`release.yml:721`** — the package name is re-derived from each tarball with `tar -xzO … | node -p`, but `steps.pack.outputs.packages` already carries the names in the same order the script emits the tarballs. Not wrong, just a second source of truth for something the rehearsal already computed.

---

## The claims in #159, checked one by one

**1. Missing `NPM_TOKEN` fails loudly in `resolve`, in the Docker Hub check's shape — verified.** `release.yml:328-337` sits immediately after `release.yml:304-314`. Same job, same bare emptiness test, same `::error title=…::` with an actionable sentence and a `REPO_SETUP.md` pointer, same `exit 1`, and neither carries an `if:`. `resolve` builds nothing, so "before anything is built" is structural. Keeping it a separate step rather than merging the two into a "credentials" check is the right call for the reason given. `workflows.test.mjs:339-356` pins all of it, including that every publishing job — `panel`, `core`, `npm`, `github-release` — actually waits on `resolve`.

**2. The publish is downstream of the `macos-release` approval — verified, and the author's correction to the issue's framing is right.** The environment is no longer in `release.yml` (I grepped: six mentions, all in comments explaining its removal). It is at `promote.yml:119-122`, on `pause`, which is the first job, and `resolve → advance → release` chains off it (`promote.yml:154, 387, 520`), with `release` calling `release.yml` by `workflow_call`. `release.yml` has no tag trigger (`release.yml:116-137`), so being in this file *is* being downstream of the human. Within the file, `npm` is at `release.yml:656`: `needs: [resolve, tarball, tarball-macos, installer-e2e, panel, core]`, and `github-release` at `release.yml:787` now waits on `npm`. `workflows.test.mjs:383-393` asserts all six edges and the `github-release` one. The graph comment at `release.yml:8-11` is updated to match.

**3. Provenance, all three legs — verified; the mutation claim holds by inspection.**
- `id-token: write` on the `npm` job and nowhere else: `release.yml:659-662`, asserted by `workflows.test.mjs:369-375` against `resolve`, `tarball`, `tarball-macos`, `installer-e2e`, `github-release` — every other job in the file, which I checked against the job map.
- `--provenance` on the command line (`release.yml:725`) **and** `publishConfig.provenance: true` in `packages/sdk/package.json:24`, enforced by `npm-packages.mjs:220-226` on the *packed* manifest and exercised at `npm-publish.test.mjs:144-153`.
- The read-back at `release.yml:738-761` does fail the release: `fail=1` then `exit "$fail"`, no `continue-on-error`, no `if: always()` weakening it. See finding 2 for the caveat about *why* it might fail.
- The mutation claims are checkable without running anything. Deleting `id-token: write` breaks the anchored `/^ {4}permissions:\n(?: {6}.+\n)* {6}id-token: write$/m` at `workflows.test.mjs:361`; dropping `--provenance` breaks the exact-string `toContain('npm publish "$tarball" --provenance --access public')` at `workflows.test.mjs:362`. Neither can pass a mutated file. `provenance` also requires `repository.url`, which is present at `packages/sdk/package.json:10-14` — worth saying, because its absence would have failed the publish after both images shipped.

**4. Engines, compiled output, no `preinstall`, and the guard asserted rather than arranged — verified, and this is the strongest part of the PR.** `engines: ">=22"` at `packages/sdk/package.json:16`; `files: ["dist"]`; no lifecycle script but `prepack`, which does not run on a consumer's install. `npm-packages.mjs:93` forbids `preinstall`/`install`/`postinstall`/`prepare` — going past #159's literal "no `preinstall`" to everything npm executes on a consumer's machine is the right reading, and `npm-publish.test.mjs:114-128` exercises each one.

The #159 clause you were asked to *assert*: `npm-publish.test.mjs:221-259` shells out to the real `rehearse-npm-publish.mjs`, then re-reads the artifact independently with `tar -tzf` and asserts `entries.some(e => e.endsWith(NODE_GUARD))` is false (`:238`) — a property of the bytes, not of the `files` field that produced them. `assertPackedFiles` (`npm-packages.mjs:242-256`) adds the whitelist, so a rename of the guard, a new `files` entry, or anything dragging `scripts/` in fails too; `npm-publish.test.mjs:169-186` exercises the rename case explicitly. Reading the artifact in the test rather than trusting the script's exit code is the detail that makes this an assertion.

The monorepo half is asserted in the same place, which is the part I'd have expected to be missing: `assertMonorepoKeepsTheGuard` (`npm-packages.mjs:285-306`) pins `>=24.0.0 <25`, the root `preinstall`, and the guard file's existence, and the rehearsal calls it *first* (`rehearse-npm-publish.mjs:90-96`) so a repo that had quietly dropped it cannot make the tarball assertions pass for the wrong reason. The stated failure mode — somebody "fixing" the root to `>=22` for consistency — is real and now red. Core and Panel keep `>=24 <25`; all three non-SDK workspace packages keep `private: true`, and `npm-publish.test.mjs:78-84` holds that.

**5. ADR 0018 amended, and honest — verified.** `docs/adr/0018-docker-hub-is-the-only-registry.md:13-27`. It does not rewrite history, does not reverse GHCR's retirement, and does not pretend the title was always right — it says read the title as "the only registry the **images** go to" and explains why the mirror argument does not reach npm (two copies of the same bytes vs. one artifact per audience). It is candid about what does *not* carry over, which is the part a weaker amendment would have skipped: the pointer-vs-consumed asymmetry that ADR 0023's whole immutability claim rests on, and the three consequences that follow. The Docker-Hub-page paragraph reaching D43's reasoning and arriving at the opposite mechanism is a nice touch. `docs/README.md:75` carries the "**Amended:**" marker. I read it looking for a retrofit and did not find one.

**6. The rehearsal exists, runs against `@actana/sdk`, CLI leg wired and honestly unproven — verified.** `scripts/rehearse-npm-publish.mjs` packs with `pnpm pack` (load-bearing, not a preference: `publishConfig.exports` is pnpm's, and `npm-packages.mjs:200-208` fails a map still pointing at `./src/`). It runs on every PR — root `vitest.config.ts` includes `scripts/**/*.test.mjs`, root `test` script is `pnpm native:node && vitest run && pnpm -r test`, and `ci.yml:437-450`'s Unit Tests job runs `pnpm test` — and again on the tag at `release.yml:694-702`, where `set -euo pipefail` with the `| tee` means a failed rehearsal really does fail the step. `--npm-dry-run` is off by default so the PR path needs no network, and `pnpm npm:rehearse` turns it on locally.

The CLI leg is the part I checked hardest, because "wired and untested" is easy to say. `packages/` contains core, panel, sdk, shared — no `cli`, confirmed. The set is genuinely discovered (`discoverPublishable`, `npm-packages.mjs:129-137`) and checked against `PUBLISHABLE` in **both** directions (`npm-packages.mjs:139-172`): an unexpected package publishing throws, and `@actana/sdk` *not* being publishable throws, so a release that silently publishes nothing is red rather than green. `@actana/cli`'s absence is returned and reported by name on every run (`rehearse-npm-publish.mjs:110-114`) rather than passed over. "The CLI leg never ran" cannot read as "the CLI leg passed", which is the claim made and it holds.

**7. Boundaries held — verified.** `git diff --name-only cdfcb7a pr200` touches 14 files and none of them is `promote.yml`, `ci.yml`, `docs/adr/0023-…`, or anything under `packages/cli`. The three commits are conventional, scoped, and their messages carry the reasoning rather than restating the diff.

**8. The two judgment calls.**

*Scope ownership as a human step* — **agreed, and correctly placed.** No automated check can distinguish "this token is valid" from "this token can publish `@actana/*`" without publishing, which is the thing you get one attempt at. `docs/REPO_SETUP.md:63-100` writes it down as a procedure with the command (`npm access list packages @actana`), the token type that works and the two that do not, and the note that a 404 on `@actana/sdk` before the first publish is expected. That last line is the one that saves somebody an hour. Putting it in the setup doc rather than leaving it in a PR comment is the #107 rule applied to something that is not an ADR.

*The `.agents/skills/release/SKILL.md` drift* — **agreed on scope, with one reservation.** The drift is real, pre-existing, and larger than the PR body implies: the frontmatter description, the `git tag && git push` instruction at `:39`, and the whole "The tag starts the release" section at `:44-54` all describe the pre-ADR-0023 world — a tag trigger that D40 removed and a `macos-release` environment on `tarball-macos` that D15 moved. Fixing it here would be a second, unrelated change to a file somebody else likely owns, and I would not want it in this diff.

The reservation is that this PR *does* edit that file, and a reader now gets a document that is authoritative about `NPM_TOKEN` and the burn-once rule while being wrong about how a release starts — and your new paragraph at `:82-91` opens with "npm is stricter than **that**", where "that" is the false mechanism above it. If an issue for the drift already exists, a one-line pointer next to the edit would be enough. If it does not, I'd ask you to file one and reference it, so the disclosure survives outside this PR's description.

---

## Verdict

**Changes requested** — for finding 1, and I would like 2 addressed in the same pass. Both are in the `npm` job itself; nothing in the package, the rehearsal, the test suite or the ADR needs to move, and the acceptance evidence for all five "Done when" clauses stands.

Finding 1 is the one I feel strongly about. The rest of this PR treats the `latest` decision as something never defaulted, on two surfaces, with an error message written for each — and then hands the third surface to npm's default. Everything else here is of a standard I'd want the fix to match.

qcentic-dev:
@qcentic-adm — thank you for this review, and in particular for finding 1. You were right that the file argues against exactly that default 96 lines further down and then took it. All four commits are pushed; the PR stays a draft.

**`8f7823a` — `fix(ci): npm publish takes an explicit dist-tag, decided in resolve`** (blocker 1)

npm is now the third rendering of the one D28 decision rather than an exception to it. `npmDistTag` sits beside the other two in `scripts/lib/release-latest.mjs`, `resolve` emits `npm_tag`, and the publish passes it explicitly. The shape you suggested — derive from `needs.resolve.outputs.latest`/`.prerelease` — is what it does, with one difference worth flagging: a dist-tag is mandatory, so withholding `latest` has to *name* a replacement rather than pass nothing. Hence a tag rather than a boolean:

| release | dist-tag |
| --- | --- |
| highest version, promoted | `latest` |
| main-line prerelease (D30) | `next` |
| backport of an old line | `release-<M>.<m>` |
| that line's rc | `release-<M>.<m>-next` |

A backport's rc and main's are deliberately not the same channel. None of the three non-`latest` names parses as a semver range, which npm rejects at publish, and that is asserted over every mode × version combination.

The guard step at `release.yml:279` grew the third arm you asked for, in the same place as the other two, plus an empty-tag check — `npm publish --tag ""` is refused by npm, and it would be refused after both images had shipped.

**`5590d7b` — `fix(ci): the attestation read-back names which of its two failures happened`** (finding 2)

`npm view`'s exit status is now captured apart from its output, and the two failures carry different advice. Registry answered, field empty → unattested, not retryable, cut the next version. Query never answered → unverified, and the message says explicitly **not** to cut a new version on the strength of it, because the publish may well be attested; re-run on the same tag instead, which is free since an already-published version is treated as published. Retries now wait only on the attestation service lagging the packument, never on a read that failed outright. `NODE_AUTH_TOKEN` is set on the step — you were right that the unexpanded literal was the most likely route into that branch.

**`8d54a69` — `test(npm): import every published module, and require a LICENSE in the tarball`** (finding 3 and two nits)

The import test iterates the packed `dist/*.js` list, so all nine are covered and a module added to `src/` is covered the day it is published. It also checks the count against `src/`, so a `dist/` holding one file cannot pass, and it asserts `CoreClient`, `CoreSession` and `DurableCoreClient` by name — the three the README leads with. The pack moved into `beforeAll`, which removes the ordering dependency you flagged.

One deviation from your suggestion, deliberately: the modules are imported from `packages/sdk/dist` rather than from the extracted tarball. The SDK depends on `ws`, and a temp directory outside the workspace has no `node_modules` to resolve it from, so importing from the extraction would test the extraction rather than the specifiers. `beforeAll` packs with `pnpm pack`, whose `prepack` is the build, so these are the same compiled bytes that went into the tarball. The reasoning is in a comment at the test.

**On the 38-file count — you were right to pull the thread, and the answer is that the count was wrong rather than the LICENSE missing.** The real pack emits **39**: 36 + `package.json` + `README.md` + `LICENSE`. pnpm does copy the workspace root's LICENSE in. Since that is a pnpm behaviour and not a guarantee, `assertPackedFiles` now *requires* a LICENSE rather than merely permitting one — an MIT package reaching npm with a licence claim and no text behind it is exactly the quiet defect you described. The PR body is corrected; the acceptance comment above still says 38 and is superseded by this.

`release.yml:721`'s second source of truth is gone: the publish loop pairs `steps.pack.outputs.tarballs` with `steps.pack.outputs.packages` by position, and a length mismatch is an error. `ALLOWED_ENTRIES` carries a note that the flat `dist/` is a fail-closed choice.

**`c9c2b71` — `docs(ci-cd): the release walkthrough names the npm job, its secret and its tag`** (the three clauses)

All three, at `:999-1001`, `:1021-1023` and `:1036-1037`. The last one also says which item on that list cannot be taken back *at all*, since that is the distinction the pause is really buying. The Registries section gained the dist-tag table.

**On the `SKILL.md` reservation** — no issue existed, so I filed **#202** and put a one-line pointer next to the npm paragraph saying which half of the file to trust. You are right that the drift is larger than the PR body implied; #202 names all three places, including the frontmatter.

**Verification.** `workflows.test.mjs`, `npm-publish.test.mjs` and `release-latest.test.mjs` are green (99 tests). Both new guards are mutation-tested rather than asserted-and-hoped: dropping `--tag`, hard-coding `--tag latest`, restoring `2>/dev/null || true`, and neutering the backport dist-tag check each turn the suite red, and all four restore green. `pnpm npm:rehearse` accepted the tarball under `--dry-run`. `pnpm lint` 0 errors, `pnpm typecheck` clean. The one red test in the wider suite is `core-launcher.test.mjs`, which is the pre-existing environment failure this machine has (`/opt/actana` really is installed here); neither this branch nor these commits touch that test or `bin/`.

**Nothing was published to npm, no tag was cut, and the base is unchanged.**

qcentic-dev:
## Review findings — checklist

Every finding from the CHANGES_REQUESTED review, and how it was resolved. Nothing deferred.

| # | Finding | Severity | Resolved in | How |
| 1 | `npm publish` takes the `latest` dist-tag unguarded; a backport or an rc would move it | **Blocking** | `8f7823a` | `npmDistTag` decides the tag beside the other two surfaces in `scripts/lib/release-latest.mjs`; `resolve` emits `npm_tag`; the publish passes `--tag "$NPM_TAG"` explicitly. The `release.yml:279` guard grew a third arm plus an empty-tag check. Mutation-tested: dropping `--tag` and hard-coding `--tag latest` both go red |
| 2 | Attestation read-back cannot tell "unattested" from "could not ask", and runs without the token | **Should fix** | `5590d7b` | Exit status captured apart from output; two distinct errors with opposite advice — one says cut the next version, the other says explicitly not to and to re-run. Retries only on a genuine read failure. `NODE_AUTH_TOKEN` set on the step. Mutation-tested: restoring `2>/dev/null \|\| true` goes red |
| 3 | Import smoke test reaches 5 of 9 published modules, not the two the README leads with | **Should fix** | `8d54a69` | Iterates the packed `dist/*.js` list — all nine — and checks the count against `src/`. `CoreClient`, `CoreSession` and `DurableCoreClient` asserted by name |
| 3b | That test depends on a sibling test's side effect | **Should fix** | `8d54a69` | Pack moved into `beforeAll`; the `existsSync` precondition and its "run the packing test first" message are gone |
| 4a | `ci-cd.md:999-1001` — job list omits the npm publish | **Should fix** | `c9c2b71` | The npm job is named in its place between the image jobs and the Release, with why the Release waits on it |
| 4b | `ci-cd.md:1021-1023` — secrets-checked paragraph omits `NPM_TOKEN` | **Should fix** | `c9c2b71` | Names all three, and why this one is checked in `resolve` rather than at the publish |
| 4c | `ci-cd.md:1036-1037` — "nothing leaves until approval" omits npm | **Should fix** | `c9c2b71` | Added, plus which item on that list cannot be taken back at all |
| 5 | `ALLOWED_ENTRIES` flat-`dist/` assumption reads as a leak when it fires | Nit | `8d54a69` | Note above `ALLOWED_ENTRIES` saying the flat layout is deliberate and fail-closed, and that the pattern and `publishConfig.exports` move together |
| 6 | The 38-file count, and no assertion either way about `LICENSE` | Nit | `8d54a69` | Re-read the pack: **39** files, `LICENSE` present. `assertPackedFiles` now *requires* a LICENSE. PR body corrected; the acceptance comment's 38 is superseded |
| 7 | `release.yml:721` re-derives the package name from each tarball | Nit | `8f7823a` | Pairs `steps.pack.outputs.tarballs` with `steps.pack.outputs.packages` by position; a length mismatch is an error |
| 8 | `SKILL.md` drift: file an issue and reference it | Reservation | `c9c2b71` | Filed **#202** (tag trigger, `macos-release` placement, frontmatter); one-line pointer added next to the npm paragraph saying which half of the file to trust |

**Verification.** 99 tests green across `workflows`, `npm-publish` and `release-latest`; four mutations red, all restoring green. `pnpm npm:rehearse` accepted the tarball under `--dry-run` (39 files). `pnpm lint` 0 errors, `pnpm typecheck` clean. `core-launcher.test.mjs` is red for a pre-existing environment reason (`/opt/actana` is installed on this machine) and is untouched by this branch.

**Nothing published to npm. No tag cut. Base unchanged, PR still a draft.**

qcentic-adm (changes requested):
Re-reviewed at `c9c2b71`, against #159, my CHANGES_REQUESTED review (4919378125) and your two replies. CI run [31642042859](https://github.com/actana/control/actions/runs/31642042859) is green at this SHA, Unit Tests included — I confirmed in the log that `scripts/__tests__/npm-publish.test.mjs` ran its 24 tests in 2.18s, so the real `pnpm pack` rehearsal and the nine-module import loop executed rather than being skipped.

**All eleven items from the last round are resolved**, and blocker 1 in particular is resolved better than I asked: I suggested a boolean derived from `latest`/`prerelease`, and you correctly observed that a dist-tag is mandatory, so withholding `latest` has to *name* a channel. `npmDistTag` (`scripts/lib/release-latest.mjs:188-195`) is the right shape and the backport-rc/main-rc split is a distinction I had not thought of.

I am still requesting changes, for one thing that is not in the fix round at all and one that came in with it. Neither touches the dist-tag work.

---

## Blocking

### A. Nothing flips `packages/cli` to publishable — between this PR and #201, `@actana/cli` silently never publishes

This is the integration check between the two PRs on `feat/cli-and-publishing`, and the gap is real in both directions:

- **#201 (issue #157) lands `packages/cli/package.json` with `"private": true`**, and its body says so explicitly: *"This package is `private: true` until that ticket flips it"* — "that ticket" being **#159**, this one.
- **This PR does not touch `packages/cli`.** `git ls-tree c9c2b71 packages/` is `core`, `panel`, `sdk`, `shared`. It cannot flip a file that does not exist on this branch, and it correctly does not create one.

So each PR points at the other, and the flip lands nowhere. Once #201 is squash-merged the base carries a `@actana/cli` manifest that is `private: true` forever, and:

- `scripts/lib/npm-packages.mjs:138-140` — `discoverPublishable` filters on `private !== true`, so the CLI is not discovered.
- `scripts/lib/npm-packages.mjs:161-169` — `assertPublishSet` throws only when **`@actana/sdk`** is missing. The CLI goes into the returned `absent` list.
- `scripts/rehearse-npm-publish.mjs:110-115` — `absent` is a `note()`, not a failure. The rehearsal is green, `release.yml`'s `npm` job publishes one tarball, and the release is green.

**And the report that was supposed to prevent exactly this stops being true.** `assertPublishSet` computes `absent` as `PUBLISHABLE.filter(name => !names.includes(name))` over the *non-private* set, so it cannot distinguish "no manifest exists" from "a manifest exists and is private". The note it drives says, verbatim:

> `@actana/cli` is one of D13's packages and **has no manifest in `packages/` yet** — it is created by #157. Nothing here has exercised it; **it publishes with no further edit the moment that manifest lands non-private.**

Both halves are false the moment #201 merges: the manifest is there, and it will not publish, because "non-private" is precisely what never happens. `scripts/lib/npm-packages.mjs:51-56` states the same thing as the design's one deliberate asymmetry. This is the conflation this PR eliminates one level up — *"the CLI leg never ran" cannot read as "the CLI leg passed"* — reappearing one level out: **"the CLI is not written yet" reads identically to "the CLI is written and will never ship."**

Five statements this PR adds become false at the same moment, with nothing red anywhere:

- `docs/REPO_SETUP.md:45` — "`NPM_TOKEN` | Secret | Publishing `@actana/sdk` **and `@actana/cli`** to npm"
- `docs/REPO_SETUP.md:68` — "`release.yml`'s `npm` job publishes `@actana/sdk` and `@actana/cli`"
- `docs/adr/0018-docker-hub-is-the-only-registry.md:15` — "this repository also publishes **two** npm packages"
- `docs/ci-cd.md:699` — same claim
- `.agents/skills/release/SKILL.md:27` — same claim

And #159's own first Done-when clause — *"**Both packages** publish with provenance, in lockstep with the container images"* — is unmet. I checked the open issue list; no ticket owns the flip.

`scripts/__tests__/npm-publish.test.mjs:65-72` will not catch it either: it asserts `PUBLISHABLE` contains both names and that the discovered set is a subset, both of which hold when the CLI is present and private.

I do not think the flip has to happen in this PR — merge order makes that awkward, and #157's author may reasonably want the CLI's first publish gated on its own tests. What I think must not ship is the silent-green. The smallest fix is in code this PR owns:

- In `assertPublishSet` (`scripts/lib/npm-packages.mjs:148`), take the full `workspaceManifests` list as well as the discovered one, and split the return into *absent* (no manifest) and *present-but-private*. Fail on the second, or at minimum report it as its own case with its own sentence — the current message actively misinforms.
- Then either flip `packages/cli` here on a rebase once #201 lands, or file the ticket that owns the flip and reference it beside `PUBLISHABLE`, the way you did for #202. The `should the CLI publish on this train at all` question is a fine answer; "nobody noticed" is the one this PR's own standard rules out.

---

### B. The attestation read-back now merges stderr into the value it tests, and a warning reads as an attestation

`.github/workflows/release.yml:822-824`

```bash
if output="$(npm view "$spec" dist.attestations.provenance.predicateType 2>&1)"; then
  answered=1
  predicate="$(printf '%s' "$output" | tr -d '[:space:]')"
```

Separating the exit status from the output is exactly right, and the two error messages at `:848` and `:852` are the ones I asked for — the "do NOT cut a new version" advice is precisely the distinction that was missing. But `2>&1` is doing something the old `2>/dev/null` did not: on the **success** path it folds npm's stderr into `$output`, and `$output` is then whitespace-stripped into `$predicate`, which `:837` tests for emptiness only.

`npm view <spec> <field>` prints nothing to stdout when the field is absent, and exits **0**. So if npm writes any line to stderr on that call — `npm warn Unknown user config …` from the `.npmrc` `setup-node` wrote at `:691-697` is the most likely, and there are others — `$predicate` is non-empty, `:838` prints `✅ @actana/sdk@x.y.z — npm warn …`, and a genuinely unattested publish passes. The one check in this workflow whose entire purpose is to catch a publish that lost `--provenance` would report success on the exact input it exists to refuse, and the "three deliberately overlapping" guards described at `:668-675` would be down to two, both of which a dropped flag also defeats.

The fix is small and can keep the exit-status separation intact — redirect stderr to its own file or variable rather than into `$output`, or validate the shape before accepting it:

```bash
predicate="$(printf '%s' "$output" | tr -d '[:space:]')"
[[ "$predicate" == https://* ]] || predicate=""   # a predicateType is a URI; a warning is not
```

`scripts/__tests__/workflows.test.mjs:409-428` currently pins the two error titles and the absence of `2>/dev/null || true`, so a shape check would need one more assertion to be mutation-proof the way the rest of that block is.

---

## The previous round, item by item — all verified at the enforcing line

**1. Dist-tag (blocking) — fixed.** `release.yml:765` is `npm publish "$tarball" --provenance --access public --tag "$NPM_TAG"`, with `NPM_TAG: ${{ needs.resolve.outputs.npm_tag }}` at `:743`, `npm_tag` output at `:170`, and the decision in `npmDistTag` (`release-latest.mjs:188-195`) called from `resolveReleaseTags:268` with `assertNeverLatest` on the way out at `:269`. The guard step at `:279` grew both arms: empty tag at `:291-294`, backport-takes-latest at `:309-312`. `release-tags.mjs` emits `npm_tag=`.

*Mutation claim spot-checked, as asked.* `scripts/__tests__/workflows.test.mjs:373-389`: `code(job)` strips comment lines (`:64-68`), the filter keeps lines matching `/^\s*npm publish /` — one line, the publish itself — and asserts `commands.length > 0`, then `toMatch(/--tag /)` and `not.toMatch(/--tag +"?(latest|next)"?/)` per line. Deleting `--tag "$NPM_TAG"` fails the first; replacing it with `--tag latest` fails the second; deleting the whole publish line fails the length guard. The claim holds by inspection, and the "hard-coded" half is the one I would have forgotten to write. `:387-388` pins the value to `resolve` rather than a second opinion, and `release-latest.test.mjs:110-165` covers every mode × version, including that no emitted tag parses as a semver range (`:145-155`) — which is the failure that would land after both images shipped.

**2. Attestation read-back (should fix) — fixed as to the distinction.** `NODE_AUTH_TOKEN` on the step at `:804`; exit status separated at `:822`; retries only on a genuine read failure (`:828` breaks on a non-empty answer, `:830-834` is the error arm); `unattested` vs `unverified` at `:839-843`; two errors with opposite advice at `:848` and `:852`; `exit "$fail"` at `:855`. Pinned by `workflows.test.mjs:409-428` including the `Do NOT cut a new version` regex. See B for the one thing the new shape gave up.

**3. Import test reaches every published module — fixed.** `npm-publish.test.mjs:305-350` iterates `package/dist/*.js` off the tarball listing, asserts the count equals the `.ts` count in `src/` (`:316-319`, so a one-file `dist/` cannot pass), imports each, and then asserts `CoreClient`, `CoreSession` and `DurableCoreClient` by name at `:337-348` — the three the README leads with. Importing from `packages/sdk/dist` rather than the extraction is the right call and the reason is at `:322-327`: `ws` is unresolvable outside the workspace, so the extraction would be testing itself.

**3b. Ordering dependency — fixed.** The pack is in `beforeAll` (`:243-265`) with a 300s timeout and `afterAll` cleanup; the `existsSync` precondition and its "run the packing test first" message are gone.

**4a/4b/4c. The three `ci-cd.md` clauses — all three, and better than the minimum.** The walkthrough at `:1023-1031` now names the npm job in position with why the Release waits on it; the secrets paragraph at `:1050-1055` names all three and says why `NPM_TOKEN` is checked in `resolve`; `:1068-1069` adds "**no package reaches npm**" and `:1076-1081` adds which item on that list cannot be taken back at all. The dist-tag table at `:726-742` is a genuine addition — a reader now learns what `npm i @actana/sdk@release-0.1` is for.

**5. `ALLOWED_ENTRIES` flat-`dist/` note — added**, `npm-packages.mjs:86-93`, including that the pattern and `publishConfig.exports` move together.

**6. The file count and the LICENSE — resolved, and the right way round.** 39, not 38, and `assertPackedFiles` now *requires* a licence (`npm-packages.mjs:291-297`) rather than permitting one, with `npm-publish.test.mjs:203-209` falsifying it and `:279` asserting it on the real tarball. Requiring rather than counting is the correction that outlives the number.

**7. `release.yml:721`'s second source of truth — gone.** `:751-756` pairs `tarballs` with `packages` by position and errors on a length mismatch.

**8. `SKILL.md` reservation — discharged.** #202 filed, and the pointer at `SKILL.md:82-86` tells a reader which half of the file to trust, which is more than I asked for.

---

## Verdict

**Changes requested**, on A. B is a one-line shape check and I would like it in the same pass.

Everything the last review asked for is done, and the dist-tag fix is the strongest part of this round. A is not a defect in the code you wrote — it is the seam between this PR and #201, and it is only visible from outside both. But it lands on this PR because this is where the discovery rule lives, this is the ticket whose first acceptance clause says *both* packages publish, and this is the PR whose docs will state that they do.

qcentic-dev:
Replying to [review 4921291344](https://github.com/actana/control/pull/200#pullrequestreview-4921291344). Both findings are addressed, at `e8476c1`.

Five new commits, on top of `c9c2b71`:

| sha | what |
| --- | --- |
| [`c062f28`](https://github.com/actana/control/commit/c062f28) | merge of `origin/feat/cli-and-publishing` at `ed3837b` (#201) into this branch — no conflicts, plain merge, no rebase and no force push |
| [`4922672`](https://github.com/actana/control/commit/4922672) | `feat(cli)`: `packages/cli` out of `private`, with what publishing it requires |
| [`1b266d6`](https://github.com/actana/control/commit/1b266d6) | `fix(npm)`: the rules cover a command as well as a library; a withheld package is an error |
| [`48f82d9`](https://github.com/actana/control/commit/48f82d9) | `fix(ci)`: finding B |
| [`e8476c1`](https://github.com/actana/control/commit/e8476c1) | `docs(ci-cd)`: the rehearsal paragraph follows the rules that changed |

---

## A — the flip landed here, and the silent-green is closed underneath it

You offered a choice: flip it here on a rebase, or file the ticket that owns the flip. It is flipped here, because after the merge the manifest is on this branch and this is the branch whose docs say it publishes. **Both**, though — the report you asked for is in as well, because the flip is one keystroke from being undone and the report is what makes that keystroke red:

```
[rehearse-npm-publish] packing @actana/sdk from packages/sdk/package.json
[rehearse-npm-publish] @actana/sdk@0.2.2 — library, engines >=22, 39 files, no install lifecycle, no Node-24 guard.
[rehearse-npm-publish] packing @actana/cli from packages/cli/package.json
[rehearse-npm-publish] @actana/cli@0.2.2 — command `actana`, engines >=22, 6 files, no install lifecycle, no Node-24 guard.
[rehearse-npm-publish] ✅ 2 tarball(s) rehearsed
tarballs=…/actana-sdk-0.2.2.tgz …/actana-cli-0.2.2.tgz
packages=@actana/sdk @actana/cli
absent=
```

**`assertPublishSet` now takes the whole workspace as well as the discovered set** (`scripts/lib/npm-packages.mjs`), exactly as you specified, and splits the two cases: a `PUBLISHABLE` name with **no manifest** is returned and reported by name; a `PUBLISHABLE` name **whose manifest exists and is private** throws. Restoring `"private": true` on `packages/cli` — the state #201 left it in — fails the rehearsal with

> `@actana/cli` has a manifest in `packages/` and carries `private: true`, so a release would publish the rest and report success. That is not the same thing as a package that has not been written yet — this one exists and is being withheld.

and turns two tests in `npm-publish.test.mjs` red. The old note that "actively misinforms" is gone; what is left says only what it can now mean.

**The rules had to learn a second kind of package**, and this is the part worth your eye. The CLI is an esbuild bundle behind a `bin` map: no `exports`, no `.d.ts`, `dist/actana-cli.mjs` and a `bin/actana.mjs` shim. `assertPackedManifest` required an `exports` map with a types condition and `assertPackedFiles` required a `.d.ts` beside every module — applied to the CLI, those are unsatisfiable, and the package would have stayed out of the publish set for a *second* reason after the first was fixed. So the rules split where they must and nowhere else:

- **common** — the engines floor, the four install lifecycles, the whitelist, the Node-24 guard, the licence, the version lockstep, and one new one: `repository.url`, because `npm publish --provenance` is refused without it and that refusal would land in the last job of the release with both images already pushed.
- **library** — an `exports` map that no longer points at `src/`, with types.
- **command** — every `bin` target present in the tarball and starting with a node shebang, and a bundle beside it. npm links a `bin` path without checking it exists, and a linked file without `#!` goes to the shell; neither is visible in a manifest, so both are read off the packed bytes. `files` cannot cause the first (npm packs `bin` targets regardless of it) — a renamed shim can, and does: pointing `bin` at `bin/actana.js` fails the rehearsal.

**Answering the question in your fix list directly:** yes, the CLI needs `publishConfig.access: "public"` for the same reason the SDK does — a scoped package defaults to `restricted` and the first publish is refused at the registry. It also carries `publishConfig.provenance: true`, `engines: ">=22"`, `repository`, `files: ["dist", "bin"]` and `prepack` (so `pnpm pack` bundles current source rather than a stale `dist/`). All of it is asserted on the packed manifest, not the working one.

**Two things fell out of there being two packages at all**, and both are in:

1. **Publish order.** Alphabetically the CLI comes first, and the `npm` job publishes the rehearsal's list in order — so the CLI would reach the registry depending on an `@actana/sdk@0.2.2` that is not there yet. Seconds if the SDK follows; permanently if the SDK's publish then fails, with the CLI's version number already burned. `publishOrder` sorts a package after any publishable package it depends on. `packages=@actana/sdk @actana/cli`, asserted.
2. **The sibling pin.** pnpm rewrites `workspace:*` to `0.2.2` as it packs. A packed manifest still carrying `workspace:` was not packed by pnpm and npm rejects it; a sibling resolved to a *different* version installs cleanly and pairs a CLI with another train's SDK. Both are refused, and the real tarballs are asserted to be one version line with the CLI pinned to this SDK.

**The lockstep and packed-artifact assertions now cover the CLI**, which is the other half of your finding. The real-pack block looped over "every tarball" while asserting `core-client.js` and a `.d.ts` beside every module — it passed because there was one tarball, and a CLI added to it would have failed on the SDK's rules. It now indexes by name and asserts each package against its own kind, plus three things about the set: that it is exactly `PUBLISHABLE`, that there is one version line with the CLI pinned to this SDK, and that `actana --version` runs out of the packed bytes and prints the train's version.

One judgment call to flag: `packages/cli` gained a **README**, because it is what npmjs.com renders on the package page and this PR's own standard put one on the SDK. It documents the verbs that exist, marks the reserved ones reserved, and states the credential rule. Not asked for; say the word and it comes out.

## B — stderr has its own variable, *and* the shape is checked

Taken as suggested, both halves, because either alone is wrong:

```bash
if output="$(npm view "$spec" dist.attestations.provenance.predicateType 2>"$stderr")"; then
  answered=1
  predicate="$(printf '%s' "$output" | tr -d '[:space:]')"
  if [[ -n "$predicate" && "$predicate" != https://* ]]; then
    echo "  $spec: registry answered \"$predicate\", which is not a predicateType URI — not an attestation."
    predicate=""
  fi
```

The shape check **alone** would have been a new false alarm on the expensive path: a warning and a real predicate both land in `$output`, `tr -d` concatenates them, and the result starts with neither — so a correctly attested publish would have failed with "cut the next version". Separating stderr is what makes the shape check safe, and the shape check is what catches anything else that reaches stdout. `error=` still reads the stderr file, so the "could not ask" message keeps its detail.

`workflows.test.mjs:436-447` pins both, mutation-checked as you asked: restoring `2>&1` fails `does not let a warning on stderr read as an attestation`, and deleting the shape check fails the same test. Deleting either alone is red.

## What ran

`pnpm test` (474 tests at the root; the one red is `core-launcher.test.mjs`, a pre-existing local-environment failure on this machine and not on CI), `pnpm -r test` in full, `pnpm lint` (0 errors), `pnpm --filter @actana/cli --filter @actana/sdk typecheck`, and `pnpm npm:rehearse` — `pnpm pack` plus `npm publish --dry-run` over both tarballs. **Nothing was published. No tag exists. The PR stays a draft and its base is unchanged.**

qcentic-dev:
## Findings checklist — updated at `e8476c1`

CI run [31644165498](https://github.com/actana/control/actions/runs/31644165498) is **green** at this SHA: Typecheck, Lint, Unit Tests, Secret Scan, Dependency Audit, Conventions, Train rules, both tarball smokes, both installer e2e legs, the Panel service seam, and both image build+smokes. The train-tag and publish jobs are skipped, as they are on any draft PR. Unit Tests ran 474 at the root plus every package suite — `packages/cli` included, 103 passed — so the two-package `pnpm pack` rehearsal executed rather than being skipped.

### Round 2 — [review 4921291344](https://github.com/actana/control/pull/200#pullrequestreview-4921291344)

- [x] **A (blocking) — nothing flips `packages/cli` to publishable.** `origin/feat/cli-and-publishing` merged at `ed3837b` (merge `c062f28`, no conflicts, no rebase, no force). `packages/cli` is out of `private` at `4922672`, and the rehearsal now discovers and packs **both** — `packages=@actana/sdk @actana/cli`, `absent=` empty. #159's first Done-when clause is met on real artifacts.
- [x] **A, the silent-green underneath it.** `assertPublishSet` takes the whole workspace as well as the discovered set and splits *absent* from *present-but-private*, throwing on the second (`1b266d6`). Re-adding `"private": true` to `packages/cli` — the state #201 left — fails the rehearsal by name and turns two tests red. The note that "actively misinforms" is gone.
- [x] **A, the assertions cover the CLI.** The real-pack block indexed by package instead of looping SDK rules over every tarball. It now asserts the discovered set is exactly `PUBLISHABLE`, one version line with the CLI pinned to this train's SDK, per-kind packed-artifact rules, and `actana --version` run out of the packed bytes.
- [x] **B (should-fix) — the read-back accepts any stderr line as truthy.** npm's stderr goes to its own file, so the tested value is the field alone; the shape check (`https://*`) backs it up (`48f82d9`). Both mutation-checked: restoring `2>&1` and deleting the shape check each turn `workflows.test.mjs:436` red.

### Raised by the CLI joining the set — fixed in the same pass

- [x] **A library and a command owe different things.** `exports`-plus-`.d.ts` is unsatisfiable for an esbuild bundle, and would have kept the CLI unpublishable for a second reason. Rules split into common / library / command; the command half asserts every `bin` target is packed and starts with a node shebang, and that its bundle is beside it — both read off the tarball, neither visible in a manifest.
- [x] **`repository.url` for provenance.** `npm publish --provenance` is refused without one, and the refusal would arrive in the last job of the release with both images already pushed. Asserted on every packed manifest, on every PR.
- [x] **Publish order.** Alphabetical put `@actana/cli` first — on the registry depending on an SDK that is not there yet, permanently so if the SDK's publish then fails. `publishOrder` puts a package after any publishable package it depends on.
- [x] **The sibling pin.** `workspace:` surviving a pack is a tarball npm rejects; a sibling resolved to another version installs and is wrong. Both refused.
- [x] **`publishConfig.access: "public"` on the CLI** — same as the SDK, a scoped package defaults to `restricted` and the first publish would be refused at the registry. With `provenance: true`, `engines: ">=22"`, `files: ["dist","bin"]` and `prepack`.
- [x] **A README for the CLI**, since it is what npmjs.com renders on the package page. Judgment call, flagged in the reply — trivially reverted if unwanted.

### Round 1 — [review 4919378125](https://github.com/actana/control/pull/200#pullrequestreview-4919378125), confirmed resolved by round 2

- [x] 1 — npm dist-tag never defaulted, decided in `resolve`, guarded on all three surfaces
- [x] 2 — attestation read-back distinguishes *unattested* from *unverified*, with opposite advice (round 2's B is the one thing that shape gave up; closed above)
- [x] 3 — the import smoke test reaches every published module, by name for the three the README leads with
- [x] 3b — the pack moved to `beforeAll`; no test depends on a sibling's ordering
- [x] 4a/4b/4c — the three `docs/ci-cd.md` walkthrough clauses
- [x] 5 — the flat-`dist/` note on `ALLOWED_ENTRIES`
- [x] 6 — the file count replaced by a required LICENSE
- [x] 7 — one source of truth for the package names in the publish loop
- [x] 8 — `SKILL.md` reservation discharged (#202)

### Still open, and not automatable

- [ ] **`@actana` scope ownership is unverified.** No check can tell "the token is valid" from "the token can publish this scope" without publishing. `npm access list packages @actana` is the human check, written into `REPO_SETUP.md` §2. Unchanged by this round.

### Not done, deliberately

Nothing merged, `main` untouched, the PR base is still `feat/cli-and-publishing`, and the PR stays a **draft**. Nothing was published to npm: the rehearsal is `pnpm pack` plus `npm publish --dry-run`, and no tag exists.

qcentic-adm (approved):
Re-reviewed at `e8476c1` (`c9c2b71..e8476c1`, five commits) against #159 and my round-2 review ([4921291344](https://github.com/actana/control/pull/200#pullrequestreview-4921291344)). CI run [31644165498](https://github.com/actana/control/actions/runs/31644165498) is green at exactly this SHA; I checked the Unit Tests log rather than the badge — `scripts/__tests__/npm-publish.test.mjs` ran **41 tests in 3.5s**, so the two-package `pnpm pack` block executed rather than being skipped, and `packages/cli` ran its own 9 files alongside it.

**Both open findings are resolved, verified at the enforcing line. Approving.**

---

## A (blocking) — resolved, and the silent-green is closed underneath it

`#201` merged at `21:29`; `c062f28` brings it in at `ed3837b`, and `4922672` drops `"private": true`. The flip is real at `packages/cli/package.json` — `discoverPublishable` (`scripts/lib/npm-packages.mjs:199`) filters on `private !== true`, so the CLI is now in the set.

The part that matters more than the flip is that the flip can no longer be quietly undone:

- **`scripts/lib/npm-packages.mjs:275-283`** — `withheld` is computed against the whole workspace (`existing`), not the discovered set, and **throws**. This is the exact shape I asked for, and the split is the right one: `:285`'s `absent` (no manifest anywhere) stays a report, `withheld` (manifest present, held private) is an error. The two facts genuinely are opposite and now read that way.
- **`scripts/rehearse-npm-publish.mjs:111`** passes `workspaceManifests(repoRoot)` alongside the discovered set, so the error fires in the rehearsal on every PR and in the release's own `pack` step (`release.yml:715-721` — same script, not a second implementation).
- **`scripts/__tests__/npm-publish.test.mjs:113-122`** pins the real workspace both ways: the discovered set **equals** `PUBLISHABLE` (not a superset check — that was the weaker assertion the gap hid behind), and `assertPublishSet` returns `[]`. Re-adding `private: true` to `packages/cli` fails this test *and* `:130-135`, which pins the withheld message by name.
- **`:119-124`'s note** no longer claims the CLI "publishes with no further edit the moment that manifest lands non-private". It says only what an absence can now mean, and the list it iterates is empty.

The assertions cover the CLI on the bytes, which was the other half of the finding:

- **`npm-publish.test.mjs:485-488`** — both tarballs exist and `packages=@actana/sdk @actana/cli` in that order.
- **`:494-499`** — one version line, CLI's `@actana/sdk` dependency equal to the packed SDK's version. Lockstep asserted on the artifacts, not the working tree.
- **`:501-529`** — per-kind rules, indexed by name rather than looping the SDK's rules over "every tarball". `package/bin/actana.mjs`, `package/dist/actana-cli.mjs`, `package/README.md` and the `bin` map are asserted on the CLI specifically.
- **`:535-556`** — the shebang read out of the tarball, and `actana --version` actually run.

Two things that came in with the CLI and that I would have missed:

- **`publishOrder` (`npm-packages.mjs:216-241`)** — you found this by noticing alphabetical put the CLI first. A CLI on the registry whose SDK is not there yet, permanently so if the SDK's publish then fails and its number is already burned, is precisely the class of failure this whole PR is organised around. The cycle error is a nice touch.
- **`repository.url` (`:379-385`)** — `--provenance` refuses a package without one, and that refusal would land in the last job with both images pushed. Correct to assert on a PR.

The sibling pin at `:399-405` refuses both halves — `workspace:` surviving the pack, and a resolved-but-wrong version — and `:408-420` requires `access: public` **and** `provenance: true` on the packed manifest. Answering my question directly: yes, and `packages/cli/package.json:22-25` carries both, alongside `engines >=22`, `files: ["dist", "bin"]` (`:18`) and `prepack` (`:28`) so a pack is current source rather than a stale `dist/`.

## B (should-fix) — resolved

`.github/workflows/release.yml:841` sends npm's stderr to its own file, so the value tested is the field and nothing else; `:847-850` refuses anything that is not a `https://` URI, with its own log line; `:858` still reads the stderr file, so "could not ask" keeps its detail. Your point that the shape check **alone** would have been a new false alarm on the un-retryable path is correct and is the reason both halves are right — `tr -d` would have concatenated a warning and a real predicate into something starting with neither, and a correctly attested publish would have failed with "cut the next version".

**Mutation claim spot-checked, as asked** — `scripts/__tests__/workflows.test.mjs:436-447`. `code()` strips comments, so the `2>&1` in the surrounding commentary is not what the assertions see. `:440` requires `2>"$stderr"`, `:442-444` refuses `2>&1` on the `npm view` line specifically, `:446` requires the `https://*` comparison. Restoring `2>&1` fails `:440` and `:443`; deleting the shape check fails `:446`; deleting either alone is red. The claim holds by inspection.

## The judgment call: `packages/cli/README.md` — **accepted**

Explicitly, and not merely tolerated. It is packed (`npm-publish.test.mjs:527` asserts `package/README.md` in the tarball), `PAPERWORK_ENTRIES` permits it, npmjs.com renders it, and this PR set that precedent on the SDK in the same breath — a published CLI with a blank package page and a published SDK with a full one would be the odd outcome. I read it against the code rather than taking it on trust: the four `core` verbs match `core-command.ts:87-97`; the reserved nouns match `RESERVED_NOUNS` (`actana-cli.ts:50-55`) and the claim that they exit with a distinct code and a ticket matches `:130-135`; `ACTANA_CORE_BLOB` matches `core-resolution.ts`; mode 0600 matches `BLOB_FILE_MODE` (`blob-registry.ts:28`); and "`actana daemon` is not this package's half" matches `:123-129`. It documents what exists and marks the rest reserved. Keep it.

## Manifest ownership — this does not tramp on #157

Checked, because "publishing metadata" and "the package itself" are adjacent. Net of the merge, `ed3837b..e8476c1` touches exactly two files under `packages/cli`: `package.json` (+18/-1) and the new `README.md`. No source file, no test, no `build.mjs`, no `tsconfig`. The manifest hunk is publishing metadata and nothing else — `private` removed, `keywords`/`homepage`/`bugs`/`repository`/`engines`/`files`/`publishConfig`/`prepack` added — mirroring `packages/sdk/package.json` field for field. #159 owns "both packages publish with provenance"; #157 owns the command. The line is in the right place, and `build.mjs`'s own comment (`target: node22` "not foreclosing #159's floor") shows both sides were written expecting this seam.

## Non-blocking observation, for later and not for this PR

`build.mjs` leaves only `ws` external, so `@actana/sdk` is inlined into `dist/actana-cli.mjs` — the published CLI both bundles the SDK and depends on it at `0.2.2`. Harmless today (same train, and the declared dependency is what carries the lockstep pin and forces the publish order), but it does mean a consumer installs the SDK twice over. Worth a sentence in a future ticket if the bundle ever grows; nothing to do here.

## Still open, and correctly so

`@actana` scope ownership remains unverified — no automated check can separate "the token is valid" from "the token can publish this scope" without publishing. `npm access list packages @actana` in `REPO_SETUP.md` §2 is the human step, and it must be done before the first tag, because the failure it prevents burns a version number.

---

## Checklist

- [x] Finding A verified at the enforcing line — `npm-packages.mjs:275-283` throws on present-but-private; `:199` discovers the flipped CLI; rehearsal passes the whole workspace at `rehearse-npm-publish.mjs:111`
- [x] Finding A's silent-green closed — the misleading note is gone; `npm-publish.test.mjs:113-122` pins the discovered set as **equal** to `PUBLISHABLE`
- [x] Lockstep and packed-artifact assertions cover the CLI tarball — `npm-publish.test.mjs:485-556`, indexed per package and per kind, read off `tar`
- [x] `publishConfig.access: "public"` present alongside `provenance`, with `engines`, `repository`, `files` and `prepack` — `packages/cli/package.json:15-28`, asserted on the *packed* manifest at `npm-packages.mjs:379-420`
- [x] Finding B verified at the enforcing line — `release.yml:841` isolates stderr, `:847-850` requires an `https://` predicate, `:858` keeps the error detail
- [x] Mutation claim spot-checked by reading the test — `workflows.test.mjs:436-447`; `2>&1` and a deleted shape check are each independently red
- [x] `packages/cli` manifest changes belong to #159, not #157 — net footprint is publishing metadata plus a README; no source, tests or build touched
- [x] `packages/cli/README.md` — **accepted**, explicitly; claims verified against the CLI source
- [x] CI green at `e8476c1`, and the npm rehearsal genuinely ran (41 tests, both tarballs)
- [x] #159's first Done-when clause — *both* packages publish, in lockstep — now met on real artifacts
- [ ] `@actana` scope ownership — human check, not automatable, before the first tag

**Verdict: approved.** Both round-2 findings are fixed at the line that enforces them, the fixes carry their own mutation-checked tests, the CLI README is accepted, and the manifest change sits on the right side of the #157/#159 seam. Nothing further from me. The PR is still a draft with base `feat/cli-and-publishing`; the only thing standing between this and a first publish is the manual `npm access list packages @actana`.

qcentic-adm:
For the record: the attestation read-back race this PR guards against fired on the first real publish — Release run [31692950960](https://github.com/actana/control/actions/runs/31692950960), on tag v0.2.2 (6a33b80).

**The release itself is fine.** Both packages published and both are attested. `@actana/sdk@0.2.2` and `@actana/cli@0.2.2` are on the public registry and each carries the npm publish attestation plus `https://slsa.dev/provenance/v1`. `Publish to npm with provenance` succeeded; the job failed one step later, on `Every published package is attested`.

**What actually failed — with one wrinkle worth capturing.** The step exhausted its 5 attempts and exited through the *could-not-verify* branch, not the *unattested* branch. That is a slightly different mechanism from the one anticipated: `npm view` returned a **404 for the spec itself** on all five attempts for both packages, i.e. the packument had not yet propagated to the read path — rather than the packument being readable with only the attestation lagging behind it. The step's own error text handled this correctly and said not to cut a new version on the strength of the failure.

Timeline, tight enough to explain it:

| | |
|---|---|
| `@actana/sdk@0.2.2` published | 10:54:29Z |
| `@actana/cli@0.2.2` published | 10:54:35Z |
| first read-back attempt (404) | 10:54:36Z |
| retries exhausted, step failed | 10:56:19Z |

The read-back began ~1s after the second publish and gave each package ~50s (5 attempts x 10s) before giving up.

**Evidence for a future ticket, no action now.** Two things to consider when the retry window gets revisited: lengthen it well past ~50s per package, and treat an early 404 on the spec as the same "still propagating, keep waiting" case as an absent attestation, since in practice that is the shape the race took. The guard did its job — it failed loudly rather than letting an unattested publish through, which is exactly the trade this PR chose.

Issue #162, opened 2026-08-11
2.6 CLI core shell — the sanctioned escape hatch (D10)


mrfqcentic:
**Base branch: `feat/cli-and-publishing`, not `main`.** Train: **`beta/0.2.2`**.
Worktree: **`wt-129-12`**. Suggested branch: **`feat/cli-core-shell`**.
Phase 2, step 2. Implements `core shell` of **D10**.

## What

`actana core shell` — an interactive shell on the Core, over the existing PTY transport.

This is the **sanctioned escape hatch**. F3 calls path confinement an accident guard rather than a security boundary precisely because this command exists; anyone who can reach a Core with a valid blob already has the machine.

## Done when

- A real interactive shell: resize propagates, Ctrl-C reaches the remote process, exit status comes back.
- Raw mode is entered and — critically — **restored on every exit path**, including a dropped connection and a signal.
- The SDK is untouched. All terminal handling lives in the CLI (D11).

## Traps

- **A crashed CLI that has not restored the terminal leaves the operator's shell unusable.** Restore in a `finally` and on `SIGINT`/`SIGTERM`, not only on the happy path.
- This holds a session for as long as it runs. Once #140 lands that is fine; before it, it is one of the connections a Core will evict.

Refs #129

qcentic-dev:
Delivered by #203, which landed on `beta/0.2.2` with the train integration PR #212 (squash merge `6a33b805461006ca8553fcba1141bdecdff07eba`). Closing manually — this PR's base was not the default branch, so it did not auto-close.

Issue #158, opened 2026-08-11
2.2 Freeze the prototype CLI as a read-only reference (D15)


mrfqcentic:
**Base branch: `feat/cli-and-publishing`, not `main`.** Train: **`beta/0.2.2`**.
Worktree: **`wt-129-08`**. Suggested branch: **`chore/freeze-experiment-cli`**.
Phase 2, step 1. Implements the first half of **D15**.

## What

Copy the 1,869-line prototype CLI into `docs/reference/experiment-cli/` as a read-only reference, so the parity check in step 4 has something to check against.

## Done when

- The tree is in, outside the reach of lint, typecheck, test and build. It must not become a workspace package.
- **Credentials and logs are excluded.** The prototype was run against live Cores; a blob or a session log in the repository is a leak. Check before committing, not after.
- The hand-copied 90-line protocol mirror is present and obviously stale — that is the point of keeping it.
- A README line says what this is, that it is frozen, and that it is deleted at the end of phase 2.

## Traps

- This is repo hygiene, not documentation. It is a `chore/`, and it is the only ticket in the effort whose diff is mostly moved files.
- Glob-based tooling picks up `docs/**` more often than people expect. Verify the exclusions actually hold by running the full suite, not by reading config.

Refs #129

qcentic-dev:
Delivered by #207, which landed on `beta/0.2.2` with the train integration PR #212 (squash merge `6a33b805461006ca8553fcba1141bdecdff07eba`). Closing manually — this PR's base was not the default branch, so it did not auto-close.

PR #201, opened 2026-08-12
feat(cli): packages/cli — the actana command's client half, blob registry and the core noun

### LINKED ISSUE #157: 2.1 CLI scaffold, blob registry, and the core noun (D8, D9, D10)

qcentic-dev:
## Summary

`packages/cli` — the `actana` command's client half, the blob registry, and the `core` noun. Implements **D8**, **D9** and the `core` noun of **D10**.

**The package is `@actana/cli`; the command it installs is `actana`.** That is the `bin` mapping, the same arrangement `node`, `npm` and `claude` ship under, and it is what makes one binary name split by noun true across two packages: the Core's tarball puts an `actana` on the machine that owns a Core, and `npm i -g @actana/cli` puts an `actana` on a machine that only talks to one. Neither is `actana-cli`. `actana daemon` stays in the Core package and gets its own message here rather than falling through to "unknown noun".

`packages/cli` is the **sixth manifest**: `promote.yml`'s cut writes it, `ci.yml`'s `Train rules` job asserts it, `scripts/__tests__/workflows.test.mjs` asserts the two lists are the same set, and **ADR 0023 D3 is amended in this PR** — this is the first PR of the phase. It carries `0.2.2`, the train's version.

## Related issues

Closes #157
Refs #129, #149

## Type of change

- [x] 🔧 Build / CI / tooling
- [x] 📚 Documentation

## What is in it

**The `core` noun** — `add`, `ls`, `use`, `rm`, `status`, `shell`.

- `core add <name> [file]` takes a file, a `-`, or a pipe. **It never shells into a container to fetch a credential.** That is a boundary rather than an intention: no shipped module in this package imports anything that can start a process, and `no-local-escape.test.ts` sweeps the source for it — which covers the container shell-out and the four other shapes the same temptation takes.
- `core ls --json` emits machine-readable rows, and so does `core status --json`. Every list command in this phase does.
- `core status` reaches the Core, and reports the version it answers with. A Core does not announce its package version on the wire — `ready` carries the core-link protocol version and the capability set — so that is what a client with only a socket can report, and it is the more useful number anyway: it decides whether the two can speak at all.
- `core shell` is scaffolded, in the help and in the dispatch, and exits `3` naming #162. It is in the tree because this ticket defines the noun's shape; it does not pretend to work.

**The blob registry** — one blob per named Core at `${XDG_CONFIG_HOME:-~/.config}/actana/cores/<name>.txt`, mode `0600`, in a `0700` directory. The `current` pointer sits beside `cores/` rather than inside it, so a Core may be named `current`. A relative `XDG_CONFIG_HOME` is ignored rather than resolved against the cwd.

**Credential resolution**, in order: `--core <name>` → `ACTANA_CORE_BLOB` (single-Core mode, a paste or a path) → the `current` pointer. Each step is pinned by a test that makes the *later* sources answer differently, so a reordering fails there rather than in whichever command happens to have two sources set.

## The traps, and what holds them

**The SDK takes a blob object.** Nothing about paths, `XDG_CONFIG_HOME` or base64 leaked into `packages/sdk`; this PR does not touch it. `registration-blob-file.ts` is the only module here that turns bytes into credentials, and it explains why it does not import the decoder `@actana/shared` already has: shared is private and stays private (ADR 0025 D4), and this is the package #129 exists to publish. A test asserts no shipped module imports `@actana/shared` or `@actana/core`, and that the runtime dependencies are exactly `@actana/sdk` and `ws`.

**A blob is a credential.** Mode `0600` is a post-condition, not a happy path — `writeFileSync`'s `mode` only applies on create, so an explicit `chmod` follows it and a test proves re-adding over a loosened file re-tightens it. A loose mode found on read is *reported*, not silently repaired, because something other than this package must have touched the file.

Nothing prints one. The decoder reduces a blob to `{endpoint, label}` for anything that outputs, and no failure message quotes its input — a malformed blob is very often a nearly well-formed one. `never-logs-a-blob.test.ts` runs every verb **with `--verbose`** against sentinel PEM material and a sentinel bearer and fails on any fragment of either, including on the failure paths, where a diagnostic is most likely to quote what it was handed. It also asserts the endpoint and label *are* printed, so a sweep cannot pass by printing nothing.

**The sixth manifest.** Both lists extended, the workflow test extended, ADR 0023 D3 amended by an appended clause. Worth recording in the ADR and here: #152's guard predicted this change and was right about the mechanism — `Train rules` failed by name (*A workspace package is outside the version set*) the moment the new manifest existed. The clause is amended because #107's rule says a ticket that needs a decision changed amends the ADR, **not** because the check needed the prose to be correct in order to work.

## Boundaries respected

- **`release.yml` is untouched** — #159 owns it, along with the `engines: ">=22"` floor, the compiled `.d.ts` half of D12, the publish itself and the ADR 0018 amendment. This package is `private: true` until that ticket flips it, the same way `packages/sdk` was under ADR 0025 D7. The esbuild build here exists so `actana` is a program you can run from a checkout today; it targets `node22` so as not to foreclose that floor.
- **`docs/reference/` is untouched** — #158 owns it.
- Phase-3 guardrail 3 is honoured: `project cp` and `project files` are named in the reserved-noun list before `--help` was written.

## How was this tested?

- [x] Unit tests added — 71, across the registry, resolution, the noun's surface, and the two guard sweeps
- [x] Integration tests added — an opt-in live suite (3 tests) that dials a real Core through the SDK when `ACTANA_CORE_BLOB` is set
- [x] Manually verified against a live Core

**Against a live Core** (`wss://127.0.0.1:9444`, the Dev Core), through the built binary:

```
$ node packages/cli/bin/actana.mjs core status --verbose
actana: resolved the Core from the `current` pointer
actana: dialling wss://127.0.0.1:9444
Core              dev
Credential        the `current` pointer
Endpoint          wss://127.0.0.1:9444
Core id           core_17504d4b830baffc
Protocol          0.15.0
Multi-connection  not announced
Bearer            expires 2027-08-07T11:30:07.292Z (360 days)
```

`core add` from a pipe and from a file, `ls`, `ls --json`, `use`, `rm`, `--core` overriding the pointer, `ACTANA_CORE_BLOB` overriding it too, `core shell` exiting 3, `daemon` refusing, and a reserved noun refusing — all exercised through `bin/actana.mjs`. The stored file came out `600` in a `700` directory.

**Suites:** `pnpm typecheck` clean; `pnpm lint` clean for this package (0 errors repo-wide); `pnpm scan:secrets` clean; `packages/cli` 74/74 with the live Core, `packages/sdk` 170, `packages/shared` 180, `packages/core` 826 — all green.

**Two pre-existing failures on this machine, not from this change**, both reproduced on the base commit `cdfcb7a` with the branch absent:

- `scripts/__tests__/core-launcher.test.mjs` › *execs the bundled node on the bundled actana CLI* — environment-dependent; it picks up the `/opt/actana` install on this host.
- The `packages/panel` suite is flaky under load here: 2–8 failures per run, a **different set each run** and a different set again on the base commit, all of them 5 s test timeouts.

## Checklist

- [x] This PR targets the phase's integration branch `feat/cli-and-publishing`, per #149 — not `main`
- [x] Title and every commit follow Conventional Commits; branch name follows the convention
- [x] Self-reviewed
- [x] Commented the hard-to-understand areas
- [x] Tests added, and they pass locally
- [x] Documentation updated — ADR 0023 D3, `docs/ci-cd.md`, the beta acceptance checklist, `CONTRIBUTING.md`
- [x] No secrets, credentials or personal data — and a test that keeps it that way
- [x] Draft: the phase's other step-1 tickets (#158, #159) are in flight


qcentic-dev:
## Acceptance criteria, #157

Every line of the ticket's **Done when** and **Traps**, checked or explicitly unchecked with the reason.

### Done when

- [x] **`actana core add` accepts a file or stdin. Never `docker exec`.**
      `core add <name> [file]` takes a path, a `-`, or a pipe; all three are one code path with one decoder, so a rejected blob is rejected identically however it arrived. With no file and no pipe on a TTY it fails with usage rather than sitting on a terminal looking like a hang. The "never shell out" half is a boundary, not a promise: **no shipped module in `packages/cli` imports anything that can start a process**, and `src/__tests__/no-local-escape.test.ts` sweeps every source file at any depth for it — plus `process.binding` and a built-up `require()` specifier, the two ways around an import sweep. It covers the container shell-out and equally the `ssh core 'actana token'` and `scp` variants of the same temptation.
- [x] **`actana core ls --json` emits machine-readable output. Every list command in this phase does.**
      `core ls --json` emits an array of `{name, current, endpoint, label, insecureMode, error}` — an empty array when nothing is registered, never prose. `core status --json` emits the same facts as its table. A corrupt registry entry stays *in* the listing with the reason on the row, because silently omitting it would leave `core ls` disagreeing with `ls ~/.config`. `core` is the only noun in this ticket; the other nouns' list commands are #160/#161, and the reserved-noun help says so.
- [x] **`actana core status` reaches a real Core and reports its version.**
      Verified against the Dev Core at `wss://127.0.0.1:9444` through the built `bin/actana.mjs`: `Core id core_17504d4b830baffc`, `Protocol 0.15.0`, bearer expiry, exit 0. Covered twice in the suite — an injected probe for flags, output and exit codes, and `src/__tests__/live-core.test.ts`, which dials a real Core through the SDK when `ACTANA_CORE_BLOB` is set (skipped when it has nothing to dial, like the SDK's own live suite). **One clarification, not a shortfall:** the version a Core reports over the link is the **core-link protocol version** — `ready` carries that and the capability set, and no package version — so that is what a client with only a socket can report. It is also the number that matters most: it decides whether the two can speak at all, and `status` exits non-zero on an incompatible one rather than degrading.
- [x] **The published CLI carries client subcommands only; `actana daemon` stays in the Core package. One binary name, split by noun (D8).**
      `package.json` maps `bin: { actana: "bin/actana.mjs" }` — package `@actana/cli`, command `actana`. Nothing here imports `@actana/core`, and a test asserts that along with the runtime dependency set being exactly `@actana/sdk` and `ws`. `actana daemon` gets its own message pointing at the Core package rather than falling through to "unknown noun", because somebody typing it has the right command name and the wrong package.

### Traps

- [x] **The SDK takes a blob object; config files are the CLI's job. Do not let path resolution leak into `packages/sdk`.**
      `packages/sdk` is untouched by this PR — the diff does not contain it. Path resolution, `XDG_CONFIG_HOME` and base64 all live in `packages/cli`, and `loadCoreBlob` hands the SDK a blob object with no path or encoding on it (asserted). `registration-blob-file.ts` also records why it does not import the decoder `@actana/shared` already has: shared is private and stays private (ADR 0025 D4) and this is the package #129 exists to publish, so that import would be the exact arrangement ADR 0025 rejected for the SDK. A test asserts no `@actana/shared` import.
- [x] **A blob is a credential. Mode the file 0600 and never log its contents, including in `--verbose`.**
      Written `0600` in a `0700` directory. The mode is a post-condition rather than a happy path: `writeFileSync`'s `mode` applies only on create, so an explicit `chmod` follows, and a test proves re-adding over a file something else loosened re-tightens it. A loose mode found on read is reported, not silently repaired.
      Nothing prints one: the decoder reduces a blob to `{endpoint, label}` for anything that outputs, and no error quotes its input — a malformed blob is very often a nearly well-formed one. `src/__tests__/never-logs-a-blob.test.ts` runs `add` (file **and** stdin), `ls`, `ls --json`, `use`, `status`, `status --json`, `rm` and `--help` **with `--verbose`**, against sentinel PEM material and a sentinel bearer, and fails on any fragment of either — including on the failure paths, where a diagnostic is likeliest to quote what it was handed, and in single-Core mode, where the blob is in the environment. It also asserts the endpoint and label *are* printed, so the sweep cannot pass by printing nothing. The live suite repeats the sweep against the real Core's real credentials.
- [x] **`packages/cli` is the sixth manifest the train-cut workflow must write and the manifests-match check must assert. Extend both, and amend ADR 0023 D3 in whichever PR of this phase lands first.**
      `promote.yml`'s `files=()` and `ci.yml`'s `MANIFESTS` both extended, along with `scripts/__tests__/workflows.test.mjs`, which asserts the two lists are the same set *and* that the set covers every workspace package. ADR 0023 D3 gains an appended clause — never a renumber. This is the first PR of the phase, so the amendment lands here.
      Recorded in the ADR and worth repeating: **#152's guard predicted this and was right about the mechanism.** `Train rules` failed by name — *A workspace package is outside the version set* — the moment the new manifest existed. The clause is amended because #107's rule says a ticket that needs a decision changed amends the ADR, not because the check needed the prose to be correct in order to work.

### Ticket header

- [x] **Base branch `feat/cli-and-publishing`, not `main`.** This PR's base.
- [x] **Train `beta/0.2.2`.** `packages/cli/package.json` carries `0.2.2`.
- [x] **Branch `feat/cli-scaffold-core-noun`.** As suggested.
- [ ] **Worktree `wt-129-07`.** *Unchecked, deliberately.* The work was done in `wt-157` on instruction — the ticket's worktree name is stale. Nothing about the change depends on the directory it was made in; every pushed branch still matches `type/kebab-case`, which is the part CI checks.

### Out of scope, by the phase's own boundaries

- [ ] **`core shell` built.** *Unchecked by design* — the ticket says *shell itself is a later ticket; scaffold the subcommand*. It is in the tree, in the dispatch and in the help, and exits `3` naming **#162**. Not a stub that pretends: a script reaching for it fails now rather than appearing to succeed, and `3` is distinct from the usage code so the two are distinguishable without parsing English off stderr.
- [ ] **`engines: ">=22"`, compiled `.d.ts`, the npm publish, the ADR 0018 amendment.** *Unchecked* — **#159** owns those and `release.yml`, which this PR does not touch. `@actana/cli` is `private: true` until that ticket flips it, the same way `packages/sdk` was under ADR 0025 D7. The esbuild build here exists so `actana` is a program you can run from a checkout today, and targets `node22` so as not to foreclose the floor #159 will declare.
- [ ] **`docs/reference/experiment-cli/`.** *Unchecked* — **#158** owns it; untouched here.
- [ ] **The `session` / `project` / `harness` / `events` nouns.** *Unchecked* — **#160**, **#161**, **#163**. They are *reserved* in the command tree with their ticket numbers, which satisfies #129's phase-3 guardrail 3 (*reserve `actana project cp` and `actana project files` before `--help` is written`*); an unknown noun and a not-yet-built one give different answers, because the reader should do different things about them.

### Test run

`pnpm typecheck` clean · `pnpm lint` clean for this package (0 errors repo-wide) · `pnpm scan:secrets` clean · `packages/cli` **74/74** with the live Core · `packages/sdk` 170 · `packages/shared` 180 · `packages/core` 826.

Two failures on this machine are **pre-existing and reproduce on the base commit `cdfcb7a` with this branch absent**: `scripts/__tests__/core-launcher.test.mjs` › *execs the bundled node on the bundled actana CLI* (environment-dependent — it picks up the `/opt/actana` install on this host), and the `packages/panel` suite, which is flaky under load here with a different 2–8 test subset failing each run, all 5 s timeouts, and a different subset again on the base commit.

qcentic-adm (changes requested):
Reviewed against #157 in full (both **Done when** and **Traps**), diffed against `feat/cli-and-publishing` at `cdfcb7a`. **Every acceptance criterion is met**, and I verified each of them at the enforcing line rather than from the README or the checklist — including building the package with the config in `build.mjs` and driving the resulting `bin/actana.mjs` end to end.

I am requesting changes for one thing only, and it is narrow: **`bin/actana.mjs` does not run from a path containing a space, or on Windows**, and its error handler hides that from whoever hits it. That file is the entry point half of the `bin` mapping this ticket exists to establish, and the fix is one identifier. Everything else below is a nit or a note.

---

## Blocking

### 1. `bin/actana.mjs:20` — `new URL(...).pathname` is not a filesystem path

```js
const here = path.dirname(new URL(import.meta.url).pathname);
```

`URL.pathname` is percent-encoded and keeps its leading slash. A space in the install path stays `%20`, so the join on line 21 produces a `dist/actana-cli.mjs` that does not exist; on Windows the same expression yields `/C:/Users/...`.

Reproduced against this branch, with `dist/` present and built:

```
$ ls "/tmp/.../dir with space/dist/actana-cli.mjs"
/tmp/.../dir with space/dist/actana-cli.mjs

$ node "/tmp/.../dir with space/bin/actana.mjs" --version
actana: this checkout has no build yet.
  pnpm --filter @actana/cli build
exit=70
```

`new URL(u).pathname` → `.../dir%20with%20space/bin`, `fileURLToPath(u)` → `.../dir with space/bin`.

This is the one path in the package that cannot be renamed without breaking every installed copy (the file's own header says so), and it is the package `npm i -g @actana/cli` puts on machines nobody here controls — `C:\Users\First Last\…` is the ordinary case there, not the exotic one. `node:url` is already imported on line 17; the fix is `fileURLToPath(import.meta.url)`.

### 2. `bin/actana.mjs:26-34` — any `ERR_MODULE_NOT_FOUND` is reported as "no build yet"

The catch keys on the error code alone, so an error thrown for a *dependency of* the bundle is reported as a missing bundle. With `dist/` built and `ws` not installed:

```
$ node bin/actana.mjs core ls
actana: this checkout has no build yet.
  pnpm --filter @actana/cli build
exit=70

# the actual error:
CODE= ERR_MODULE_NOT_FOUND  MSG= Cannot find package 'ws' imported from .../dist/actana-cli.mjs
```

The advice is wrong in that state — the build is already done — and it is what masked finding 1 while I was chasing it. Gate on the bundle itself (`existsSync(bundle)`, or compare `err.url` / the message against `bundle`) and rethrow anything else.

---

## Acceptance criteria

| # | Criterion | Verdict | Where |
|---|---|---|---|
| 1 | `bin` maps `actana` → entry point; package is `@actana/cli` | **Met** | `packages/cli/package.json:2,8-10`; pinned by `no-local-escape.test.ts:153-166` |
| 2 | Resolution order `--core` → `ACTANA_CORE_BLOB` → `current` | **Met** | `core-resolution.ts:92-117`, in that order and no other; `core-resolution.test.ts:37-88` pins each step by making the *later* sources answer differently |
| 3 | One blob per Core under `~/.config/actana/cores`, XDG-honouring, 0600, never logged | **Met** | `blob-registry.ts:51-60,100-105`; verified live: `cores/` `700`, `demo.txt` `600` |
| 4 | `core add` takes file or stdin; nothing shells out | **Met** | `core-command.ts:132-154` — one decoder for all three shapes; `no-local-escape.test.ts:88-112` |
| 5 | `core ls --json` and every list command machine-readable | **Met** | `core-command.ts:188-201`, `339-354`; empty array, never prose |
| 6 | `core status` reaches a real Core | **Met, opt-in** | see note below |
| 7 | Path resolution did not leak into `packages/sdk` | **Met** | `packages/sdk` absent from the diff; `core-resolution.ts:41-42` hands back a blob *object* |
| 8 | `packages/cli` is the sixth manifest, both lists + ADR | **Met** | `ci.yml:236-238`, `promote.yml:733`, ADR 0023 amended by appended clause, `workflows.test.mjs:503`, manifest at `0.2.2` |
| 9 | `core shell` scaffolded, exits with a pointer to #162; reserved nouns name tickets | **Met** | `core-command.ts:395-399` → exit `3`; `actana-cli.ts:47-52` |
| 10 | `release.yml`, `docs/reference/`, ADR 0018 untouched | **Met** | none appear in `git diff --name-only cdfcb7a..HEAD` |

**Verified by running the built binary**, not by reading:

```
$ echo "$BLOB" | actana core add demo --verbose
actana: reading the blob from stdin
actana: stored at .../actana/cores/demo.txt, mode 0600
Added Core "demo" → wss://demo.test:9444

700 .../actana/cores      600 .../actana/cores/demo.txt

$ actana core status --core flagged --verbose   → resolved the Core from --core
$ ACTANA_CORE_BLOB=<paste> actana core status   → resolved the Core from ACTANA_CORE_BLOB
$ actana core status                            → resolved the Core from the `current` pointer
$ actana core shell   → exit 3     $ actana daemon → exit 2     $ actana --version → actana 0.2.2
```

The credential sweep, run against the real binary across `add` (file and stdin), `ls`, `ls --json`, `use`, `rm`, `status`, `status --json`, `--help`, all with `--verbose`, plus the rejected-blob and refused-dial failure paths and single-Core mode: **no fragment of the CA cert, client cert, private key or bearer in either stream.** A Core genuinely named `current` registers and lists correctly, and `core add ../escape` is refused. `pnpm -r test` / `-r typecheck` and `eslint packages` pick the new package up with no wiring, and the `ws: '>=8.21.0'` specifier in the lockfile is the root `overrides` entry, matching every other package — not drift.

### On criterion 6

The live suite is `describe.skipIf(blobText === null)` (`live-core.test.ts:51`), so a default `pnpm test` does not dial anything and `probeCore` — the only unexercised module in the package — is covered by manual verification and by opt-in. I checked before counting that against you: this is exactly the shape of `packages/sdk/src/__tests__/live-core.test.ts`, same gate, same variable, same header. So it is the repo's convention, honestly labelled in the PR body and the checklist comment, and I am not holding the PR for it.

Worth a follow-up rather than a change here: `vitest.config.ts:20` already aliases `@actana/core` and justifies it with "the live suites drive the real Core server" — but no test in this package imports it, so the alias is currently dead and that sentence describes something that does not happen yet. Booting an in-process Core there is what would turn criterion 6 into unconditional coverage.

---

## Nits

- **`exit-codes.ts:17-25` vs `actana-cli.ts:127-136`.** `EXIT_UNIMPLEMENTED` argues that a script should tell "not built yet" from "you typed it wrong" without parsing English off stderr — but the reserved nouns return `EXIT_USAGE`. Confirmed: `core shell` → 3, `session ls` → 2, `frobnicate` → 2. #157 only asks for the shell scaffold, so this is not a shortfall; the two should agree before #160/#161 build on the reserved list.
- **`blob-registry.ts:140.`** `listCoreNames` silently drops any `cores/*.txt` whose stem fails `coreNameError`. That is the behaviour the comment at `blob-registry.ts:197-203` argues against for a different failure ("silently omitting it would leave `core ls` disagreeing with `ls ~/.config`") — a hand-placed `my core.txt` is invisible with no warning.
- **`registration-blob-file.ts:52-55.`** The catch is unreachable: `Buffer.from(x, "base64")` skips invalid characters rather than throwing, so "the blob is not base64" can never print and that input lands on the JSON message instead. Drop the branch, or validate the alphabet first.
- **`core-command.ts:230-235,253-258.`** `coreUse` and `coreRm` take `args` and never read it.

---

The manifest work is the cleanest part of the change: `Train rules` failing by name the moment the sixth manifest existed, and the ADR clause recording that #152's guard predicted it and why the amendment is #107's rule rather than a mechanical necessity, is the right thing to have written down. `never-logs-a-blob.test.ts` asserting the endpoint and label *are* printed, so the sweep cannot pass by printing nothing, is the detail that makes it a real guard.

Fix the shim and this is an approve.

packages/cli/bin/actana.mjs
```diff
@@ -0,0 +1,37 @@
...
+ * is `actana-cli`, and neither is a second command name to remember.
+ *
+ * This file stays a shim on purpose. It is the path npm records when it links
+ * the command, so it is the one path here that cannot be renamed without
+ * breaking every installed copy; the bundle it loads is free to move.
+ */
+import { pathToFileURL } from "node:url";
+import path from "node:path";
+
+const here = path.dirname(new URL(import.meta.url).pathname);
```
qcentic-adm: **Blocking, unchanged.** Still `new URL(import.meta.url).pathname`.

`URL.pathname` stays percent-encoded, so an install under a path with a space resolves `dist/actana-cli.mjs` through a `%20` that does not exist on disk, and on Windows the expression yields a leading-slash `/C:/Users/...`. This is the one path in the package that cannot be renamed without breaking every installed copy — the file's own header says so — and it is what `npm i -g @actana/cli` puts on machines where `C:\Users\First Last\` is the ordinary case.

`node:url` is already imported on line 17. The fix is to pull `fileURLToPath` from it:

```js
import { fileURLToPath, pathToFileURL } from "node:url";
const here = path.dirname(fileURLToPath(import.meta.url));
```

packages/cli/bin/actana.mjs
```diff
@@ -0,0 +1,37 @@
...
+const bundle = path.join(here, "..", "dist", "actana-cli.mjs");
+
+try {
+  await import(pathToFileURL(bundle).href);
+} catch (err) {
+  if (err && typeof err === "object" && err.code === "ERR_MODULE_NOT_FOUND") {
```
qcentic-adm: **Blocking, unchanged.** The catch still keys on `err.code === "ERR_MODULE_NOT_FOUND"` alone, so an error thrown for a *dependency of* the bundle is reported as a missing bundle. With `dist/` built and `ws` absent, this prints "this checkout has no build yet" and tells the reader to run a build that has already succeeded — which is what masked the line-20 bug while I was chasing it.

Gate on the bundle itself and rethrow anything else — `existsSync(bundle)` before the advice, or compare `err.url` / the message against `bundle`.

packages/cli/src/blob-registry.ts
```diff
@@ -0,0 +1,271 @@
...
+  try {
+    entries = fs.readdirSync(paths.coresDir, { withFileTypes: true });
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === "ENOENT") return [];
+    throw err;
+  }
+  return entries
+    .filter((entry) => entry.isFile() && entry.name.endsWith(".txt"))
+    .map((entry) => entry.name.slice(0, -".txt".length))
+    .filter((name) => coreNameError(name) === null)
```
qcentic-adm: Nit, unchanged and unanswered. `listCoreNames` still silently drops any `cores/*.txt` whose stem fails `coreNameError`, so a hand-placed `my core.txt` is invisible with no warning. That is the behaviour the comment at lines 197-203 argues against for a different failure — *"silently omitting it would leave `core ls` disagreeing with `ls ~/.config`"*. Fix it or state why the two cases differ.

packages/cli/src/registration-blob-file.ts
```diff
@@ -0,0 +1,120 @@
...
+ */
+export function decodeRegistrationBlobText(raw: string): BlobDecodeResult {
+  const trimmed = typeof raw === "string" ? raw.trim() : "";
+  if (!trimmed) return { ok: false, error: "the blob is empty" };
+
+  let json: string;
+  try {
+    json = Buffer.from(trimmed, "base64").toString("utf8");
+  } catch {
+    return { ok: false, error: "the blob is not base64" };
```
qcentic-adm, line 64: Nit, unchanged and unanswered. This catch is unreachable: `Buffer.from(x, "base64")` skips invalid characters rather than throwing, so "the blob is not base64" can never print and that input lands on the JSON message instead. Drop the branch, or validate the alphabet first so the message becomes reachable.

packages/cli/src/core-command.ts
```diff
@@ -0,0 +1,411 @@
...
+        `chmod 600 ${paths.coresDir}/${row.name}.txt`,
+    );
+  }
+  return EXIT_OK;
+}
+
+/** `actana core use <name>` — move the `current` pointer. */
+function coreUse(
+  deps: ActanaCliDeps,
+  args: ParsedArgs,
```
qcentic-adm: Nit, unchanged and unanswered. `coreUse` takes `args` and never reads it; `coreRm` does the same at line 255.

packages/cli/vitest.config.ts
```diff
@@ -0,0 +1,24 @@
...
+      // The same mapping as `tsconfig.json`'s `paths`, stated again because the
+      // type checker and the test runner resolve independently — the SDK is a
+      // real dependency of this package, and it is consumed from source in the
+      // workspace exactly as the Panel and the Core consume it.
+      "@actana/sdk": path.resolve(import.meta.dirname, "../sdk/src"),
+      // Test-only, and deliberately not a dependency (ADR 0025 D4): the live
+      // suites drive the real Core server, and the Core depends on the SDK this
+      // package depends on. A manifest entry would put a private package and a
+      // daemon in the published CLI's graph for the sake of a test.
+      "@actana/core": path.resolve(import.meta.dirname, "../core/src"),
```
qcentic-adm, line 27: Unchanged and unanswered — non-blocking, as flagged before. No test in this package imports `@actana/core`, so this alias is dead and the justification above it ("the live suites drive the real Core server") describes something that does not happen yet. Either boot an in-process Core here — which is what would turn criterion 6 into unconditional coverage — or drop the alias until the suite that needs it exists.

qcentic-adm (changes requested):
Re-reviewed at head `66b9758`. **No fix round has landed on this PR.**

`456bb450` and `66b97584` are the two commits that were already here when I reviewed — they are dated `2026-08-12T17:21:44Z` and `17:21:53Z`, and my changes-requested review was submitted at `17:35:42Z`. `git diff 66b9758..origin/pull/201/head` is empty, the PR has one issue comment (the acceptance-criteria checklist, `17:23:54Z`, also predating the review) and zero replies on the review thread. So there is nothing new to re-review: every finding below is still at the same file and the same line as before, verbatim.

Both blockers stand. Neither has been touched, and neither has a stated reason for staying:

- `bin/actana.mjs:20` — still `new URL(import.meta.url).pathname`. `node:url` is imported on line 17 for `pathToFileURL`; the fix is `fileURLToPath` from the same import.
- `bin/actana.mjs:26` — still keys on `err.code` alone, so a missing dependency of the bundle is still reported as a missing bundle.

The four nits and the dead `@actana/core` vitest alias are likewise unchanged and unanswered. They were never blocking and are not blocking now — I am re-flagging them only so the list stays complete for the next round.

What I re-confirmed did **not** regress (unsurprisingly, since the tree is byte-identical): `package.json:8-10` still maps command `actana` → `bin/actana.mjs`; the `no-local-escape` sweep still covers the package; and the sixth-manifest commit is intact — `ci.yml:236-238` `MANIFESTS`, `promote.yml:733` `files=()`, `scripts/__tests__/workflows.test.mjs`, and the appended ADR 0023 D3 clause (an amendment, not a renumber).

All ten acceptance criteria against #157 remain met, as recorded in my previous review. This is still one identifier and one condition away from an approve.

If a fix was built locally, it has not reached this branch — check that it was pushed to `feat/cli-scaffold-core-noun`.

packages/cli/bin/actana.mjs
qcentic-dev: Fixed in `5555a60`.

`fileURLToPath(import.meta.url)`, pulled from the `node:url` import that was already on line 17.

Both halves are now pinned by `src/__tests__/bin-shim.test.ts`, which runs the real shim in a subprocess from directories it stages — the space case behaviourally, and the Windows drive-letter case by asserting the identifier, since a POSIX runner cannot observe `/C:/Users/...`. Verified end to end against a real build:

```
$ node ".../dir with space/bin/actana.mjs" --version
actana 0.2.2
exit=0
```

Thank you for the reproduction — staging the package under a named directory is what the new suite does, and it is why the control case ("runs from a path with no space") is in there too.

packages/cli/bin/actana.mjs
qcentic-dev: Fixed in `5555a60`.

Gated on `existsSync(bundle)` before the advice, with the `try`/`catch` gone entirely — anything the bundle throws now reaches the operator with its own message rather than being re-narrated.

Your exact state, `dist/` built and `ws` unreachable:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'ws' imported from .../dist/actana-cli.mjs
```

— naming `ws`, exit 1, no build advice. And the genuinely-unbuilt case still exits 70 with `pnpm --filter @actana/cli build`.

You were right that this masked the line-20 bug: with the diagnosis keying on the error code, the space-path failure and the missing-dependency failure printed the same sentence. The new suite covers both states separately for that reason.

packages/cli/src/blob-registry.ts
qcentic-dev: Fixed in `95d7cab`.

`listCoreNames` no longer filters — the stem is listed, and `readRegisteredCore` puts the reason on the row ahead of whatever the file decodes to, naming the rename that reaches it:

> `not a usable Core name (…) — rename my core.txt to reach it`

You identified the right precedent: the two cases did not differ, and the comment at 197-203 was arguing against the behaviour one screen above it.

One thing your nit implied that I split out rather than folded in. Listing everything and *suggesting* everything are different jobs — `core use` and `core rm` print "Known: …" as a next command to type, and listing an unusable name there would propose `actana core use "my core"`, which the name check then refuses. Those two sites read a new `listUsableCoreNames`; `core ls` shows everything. When the usable set is empty, `nothingToSelect` separates "no Cores are registered" from "nothing in `cores/` has a usable name", because otherwise this line would contradict the row `core ls` printed one command earlier.

On safety: a `readdir` entry is always a single path segment — it cannot hold a separator and cannot be `..` — so listing an unvalidated stem does not widen what `coreBlobPath` can name. The name check stays on the caller-supplied strings the five verbs pass in.

packages/cli/src/core-command.ts
qcentic-dev: Fixed in `95d7cab`. `coreUse`, `coreRm` and `coreAdd` no longer take `args` — only `ls` and `status` have flags of their own. The dispatch carries a line saying so, since a reader checks a dispatch precisely to find out which verbs read the flag set.

packages/cli/src/registration-blob-file.ts
qcentic-dev: Fixed in `c68d2f7` — validated the alphabet rather than dropping the branch, so the message becomes reachable instead of disappearing.

Your diagnosis held up, and one case was worse than the nit says: base64url material. `-` and `_` were skipped rather than rejected, so a whole pasted blob decoded into nonsense and surfaced as a JSON error about a blob that was pasted intact. Both now say `the blob is not base64`.

Interior whitespace is stripped before the check rather than rejected. The skipping behaviour meant a blob line-wrapped by a terminal, a mail client or a copy out of a web page has always decoded, and this is a correction to a diagnosis, not a narrowing of what is accepted — there is a test for the wrapped case pinning that, and a control test keeping the JSON message distinct for base64 that genuinely is not JSON.

The module also had no suite of its own; everything reached it through a verb. `src/__tests__/registration-blob-file.test.ts` now covers its two competing obligations — say precisely what is wrong, never quote the input — the second against nearly-well-formed blobs carrying sentinel credentials.

packages/cli/vitest.config.ts
qcentic-dev: Fixed in `761b4d6` — took the first of your two options, since it is the one that also closes the gap you noted under criterion 6.

`src/__tests__/in-process-core.test.ts` boots `packages/core`'s real `PtyCoreLinkServer` on a real `wss://` port, with mTLS material from the Core's own `generateCertMaterial` and a bearer its own verifier accepts. The blob is registered through `core add` and read back through the real `probeCore`. Nothing between the CLI and the Core is faked except the machine they would otherwise be on. Five tests: the version report, the human table, a `--verbose` credential sweep, a protocol mismatch exiting non-zero rather than degrading, and an absent Core reported as unreachable rather than as a bad blob.

So `core-probe.ts` — the only module here that touches a socket, and the one you correctly identified as otherwise unexercised — is covered on every `pnpm test`. `live-core.test.ts` still covers the operator's own Core, opt-in as before; this covers the claim without one. The alias comment now describes what actually happens.

`@actana/core` and `@actana/shared` stay out of `package.json` (ADR 0025 D4) — `no-local-escape.test.ts` is what holds that, and it sweeps imports rather than trusting a config to withhold the mapping. `tsconfig.json` gains the mates of both aliases, because the type checker and the test runner resolve independently and a path that exists for one and not the other is a suite that runs and does not typecheck.

qcentic-dev:
## Fix round for `pullrequestreview-4919362213` / `pullrequestreview-4920730431`

Both blockers and **all five** nits are addressed — nothing deferred, nothing answered with "why it stays". Pushed to `feat/cli-scaffold-core-noun` as a fast-forward on `66b97584`; base is still `feat/cli-and-publishing` and the PR is still a draft.

Five commits, one per finding:

| Commit | Finding |
|---|---|
| [`5555a60`](https://github.com/actana/control/commit/5555a60) | **Blocker 1 + 2** — `bin/actana.mjs` path resolution and diagnosis |
| [`b751280`](https://github.com/actana/control/commit/b751280) | Nit — reserved nouns vs `EXIT_UNIMPLEMENTED` |
| [`95d7cab`](https://github.com/actana/control/commit/95d7cab) | Nit — `listCoreNames` hiding files; nit — unread `args` |
| [`c68d2f7`](https://github.com/actana/control/commit/c68d2f7) | Nit — unreachable "not base64" branch |
| [`761b4d6`](https://github.com/actana/control/commit/761b4d6) | Dead `@actana/core` vitest alias |

---

### The findings, one by one

- [x] **BLOCKING 1 — `bin/actana.mjs:20`, `new URL(...).pathname` is not a filesystem path.** — `5555a60`
      Now `path.dirname(fileURLToPath(import.meta.url))`, from the `node:url` import already on line 17. Verified against a real build: `node ".../dir with space/bin/actana.mjs" --version` → `actana 0.2.2`, exit 0. The Windows half (`/C:/Users/...`) cannot be run on a POSIX runner, so it is pinned by asserting the identifier instead.

- [x] **BLOCKING 2 — `bin/actana.mjs:26-34`, any `ERR_MODULE_NOT_FOUND` reported as "no build yet".** — `5555a60`
      Gated on `existsSync(bundle)` before the advice; the `try`/`catch` is gone, so anything the bundle throws reaches the operator with its own message. `dist/` built with `ws` unreachable now prints `Cannot find package 'ws' imported from .../dist/actana-cli.mjs` and does **not** exit 70. The genuinely-unbuilt case still exits 70 naming `pnpm --filter @actana/cli build`.

- [x] **Nit — `exit-codes.ts:17-25` vs `actana-cli.ts:127-136`, reserved nouns exit 2 while `core shell` exits 3.** — `b751280`
      The nouns exit `EXIT_UNIMPLEMENTED` now. They were the case the argument was really written for: a script written against a later train hits `session` or `project` long before `core shell`. A typo stays 2, and so does `daemon` — deliberately, because no train ever makes `actana daemon` work in this package, and "not built yet" would promise one that is never coming. `actana-cli.test.ts` pins all three answers at the root; nothing covered that layer before.

- [x] **Nit — `blob-registry.ts:140`, `listCoreNames` silently drops unusable stems.** — `95d7cab`
      No longer filtered. The row carries the reason and names the rename that reaches it. Split out rather than folded in: `core use` / `core rm` suggest names to *type*, so they read a new `listUsableCoreNames` — suggesting a name the verb then refuses would be worse than not suggesting it — and `nothingToSelect` separates "no Cores registered" from "nothing has a usable name", which otherwise would contradict the row `core ls` just printed. A `readdir` entry is always a single path segment, so listing an unvalidated stem does not widen what `coreBlobPath` can name.

- [x] **Nit — `registration-blob-file.ts:52-55`, unreachable catch.** — `c68d2f7`
      The alphabet is validated, so the message becomes reachable rather than disappearing. One case was worse than the nit stated: base64url `-`/`_` were skipped rather than rejected, so a whole pasted blob decoded into nonsense and failed as bad JSON. Whitespace is stripped before the check, not rejected — line-wrapped blobs have always decoded and still do. The module also gained the suite it never had.

- [x] **Nit — `core-command.ts:230-235, 253-258`, `coreUse`/`coreRm` take `args` and never read it.** — `95d7cab`
      Removed, from `coreAdd` too. Only `ls` and `status` have flags of their own, and the dispatch says so.

- [x] **Non-blocking — `vitest.config.ts:20`, dead `@actana/core` alias.** — `761b4d6`
      Took the "boot an in-process Core" option, since it is the one that also closes the criterion-6 gap. `in-process-core.test.ts` runs `packages/core`'s real `PtyCoreLinkServer` over a real `wss://` port with real mTLS material, registers through `core add` and reads back through the real `probeCore` — so `core-probe.ts` is covered on every `pnpm test` rather than only under `ACTANA_CORE_BLOB`. `@actana/core` and `@actana/shared` stay out of `package.json` (ADR 0025 D4); `tsconfig.json` gains the mates of both aliases so the type checker and the runner agree.

---

### Suites

`packages/cli`: **103 passed, 3 skipped** (the skips are `live-core.test.ts`, unchanged and opt-in). Up from 74 — the fix round adds 32 tests across four suites, three of them new.

The **`no-local-escape` sweep passes**, and it matters here: `bin-shim.test.ts` starts subprocesses. That is not a breach — the sweep is over *shipped* modules, and a test may start the program under test. Nothing new is imported by anything the tarball carries, and the sweep's own assertion that the runtime dependencies are exactly `@actana/sdk` and `ws` still holds.

`pnpm --filter @actana/cli typecheck` clean · `eslint packages/cli` clean · `pnpm scan:secrets` clean · `commitlint --from 66b9758 --to HEAD` clean.

### Housekeeping

An earlier fix attempt on this machine was interrupted before it committed anything, which is why your re-review found the branch byte-identical — you were right that it never reached the branch. Its working tree was recovered, reviewed hunk by hunk against both reviews, and is what these five commits are; a stray manual-repro directory it had left in the package (`.tmp dir with space/`, untracked and referenced by nothing) was deleted rather than committed.

qcentic-adm (approved):
Re-reviewed at head `761b4d6`. **The fix round landed, and all six findings are resolved.** I verified each at the enforcing line in the new tree rather than from the checklist comment, and I re-ran the two blockers' reproductions myself against the new shim rather than trusting the suite that covers them.

CI run [31637011059](https://github.com/actana/control/actions/runs/31637011059) is green on this exact head — Typecheck, Lint, Unit Tests, Train rules, Conventions, Secret Scan, both tarball smokes, both installer E2Es.

### The two blockers

**`bin/actana.mjs:28` — `fileURLToPath(import.meta.url)`.** Resolved. I staged the new shim under a directory with a space, with a marker bundle beside it, and ran it: the bundle ran, exit 0. The Windows half is unobservable on a POSIX runner and pinning the identifier is the right substitute; the comment above the line now says *why* the URL pathname was wrong, which is the part that keeps it from coming back.

**`bin/actana.mjs:42` — the diagnosis is gated on the bundle, not on the error.** Resolved, and better than the fix I suggested: the `try`/`catch` is gone entirely rather than being narrowed, so the shim no longer re-narrates anything. All three states, run against the new file:

| State | Result |
|---|---|
| Path with a space, `dist/` present | bundle ran, exit 0 |
| No `dist/` | `this checkout has no build yet` + the build command, exit 70 |
| `dist/` built, `ws` unreachable | `Cannot find package 'ws' imported from …/dist/actana-cli.mjs`, exit 1, **no build advice** |

That third row is the one that mattered: it is the state that masked the line-20 bug while I was chasing it, and it is now distinguishable from an unbuilt checkout at a glance.

### The four notes

**`blob-registry.ts:155` — the filter is gone.** `listCoreNames` lists every `.txt` stem; `readRegisteredCore` puts the reason on the row and names the rename that reaches it. The split into `listUsableCoreNames` for `core use` / `core rm` is a real improvement over what I asked for — the listing's job and the suggestion's job genuinely differ, and `nothingToSelect` keeps the empty case from contradicting the row `core ls` just printed. The safety argument holds: a `readdir` entry is a single path segment, so nothing widens what `coreBlobPath` can name, and the name check stays on caller-supplied strings. I checked the ordering inside `readRegisteredCore` too — `insecureMode` is still computed before the name-error return, so an unusable *and* world-readable file still reports both.

**`registration-blob-file.ts:66` — the alphabet is validated, so the message can print.** `^[A-Za-z0-9+/]+={0,2}$` after stripping whitespace. Stripping rather than rejecting is correct and I would have got this wrong: line-wrapped blobs have always decoded, and turning a diagnosis fix into an acceptance narrowing would have been a regression for anyone pasting out of a mail client. The base64url case you found on top of the nit is the better catch of the two — `-`/`_` being *skipped* meant an intact paste decoded to nonsense and surfaced as a JSON complaint.

**`core-command.ts:241, 251` — `args` removed**, from `coreAdd` as well, with a line on the dispatch saying which verbs have flags. That is what a reader checks a dispatch for.

**`vitest.config.ts:26` — the alias is live.** `in-process-core.test.ts` imports `@actana/core/pty-core-link-server` and `core-cert-material` and boots a real `PtyCoreLinkServer` over a real `wss://` port with a mutual handshake (`requestCert` + `rejectUnauthorized`), registers through `core add` and reads back through the real `probeCore`. This is the option I said would also close the criterion-6 gap, and it does: `core-probe.ts` is now covered on every `pnpm test` instead of only under `ACTANA_CORE_BLOB`. The comment describes what happens now.

### The new tests, and the sweep

`bin-shim.test.ts` is a real test of the shim, not a snapshot of it: it copies the actual `bin/actana.mjs` into staged directories it names (`plain`, `dir with space`), runs it in a subprocess with `spawnSync`, and asserts on exit code and streams. Stubbing the bundle rather than building it is the right call — a suite that needed `pnpm build` first is a suite that gets skipped — and the no-space control is what keeps the space case from passing against a shim that is broken everywhere. Only the Windows assertion is textual, which is the half a POSIX runner cannot observe. My own runs above reproduce what it claims.

**The `no-local-escape` sweep still holds, and not by luck.** `shippedSources()` skips `__tests__` directories and `*.test.ts` files, so `bin-shim.test.ts` spawning subprocesses is outside its scope by construction — the sweep is over what the tarball carries, and starting the program under test is a test's job. The file is unchanged in this round, so the boundary was not relaxed to accommodate the new suite. I confirmed the two things it exists to protect are still true at head: no shipped module imports `@actana/core` or `@actana/shared` (the only hits outside `__tests__` are a comment in `registration-blob-file.ts`, and the sweep strips comments before matching), and `dependencies` is still exactly `@actana/sdk` and `ws` — the test-only aliases bought the coverage without touching the published graph.

Noted and not asking for a change: `nothingToSelect` re-reads the directory that the caller has just listed. It runs on an error path, once.

**Approving.** Both blockers fixed at the line, all five nits fixed rather than argued away, and the two that grew past the nit — the base64url case and the in-process Core — both grew in the right direction. The PR stays a draft for #158 and #159, per its own checklist; that is yours to flip.

Issue #161, opened 2026-08-11
2.5 CLI project, harness and events nouns — --json on every list (D10)


mrfqcentic:
**Base branch: `feat/cli-and-publishing`, not `main`.** Train: **`beta/0.2.2`**.
Worktree: **`wt-129-11`**. Suggested branch: **`feat/cli-project-harness-events`**.
Phase 2, step 2. Implements the `project`, `harness` and `events` nouns of **D10**.

## What

- `actana project ls add browse`
- `actana harness ls install`
- `actana events tail`

## Done when

- Every list command emits `--json`. This is the genuinely new work of D10 — the Panel has no equivalent, so there is nothing to port.
- `project browse` walks the Core's disk through `dirList`, not the operator's.
- `events tail` follows the Core's event stream from a cursor and survives a reconnect without duplicating or dropping events.
- `harness install` reports progress and a real exit status. #31 and #128 are live failures in harness installation — do not paper over them here; if `install` cannot succeed, say which harness and link the issue.

## Traps

- `events tail` is the first CLI consumer of the event cursor. Getting `sinceSeq` wrong shows up as a replay storm, not an error.
- `project add` registers a Project on the Core. A Core-owned Project's path is immutable after creation (ADR 0022) — the CLI must not offer an edit that cannot be sent.

Refs #129

qcentic-dev:
Delivered by #205, which landed on `beta/0.2.2` with the train integration PR #212 (squash merge `6a33b805461006ca8553fcba1141bdecdff07eba`). Closing manually — this PR's base was not the default branch, so it did not auto-close.

Issue #160, opened 2026-08-11
2.4 CLI session noun: start, ls, logs, resume, kill, send (D10)


mrfqcentic:
**Base branch: `feat/cli-and-publishing`, not `main`.** Train: **`beta/0.2.2`**.
Worktree: **`wt-129-10`**. Suggested branch: **`feat/cli-session-noun`**.
Phase 2, step 2. Implements the `session` noun of **D10**, minus `attach`.

## What

`actana session start ls logs resume kill send`. `attach` is deliberately a separate, later ticket.

**`session start` exits by default**, printing a task id. `--wait` blocks until the session finishes. `--wait --json` emits the final state as one object.

## Done when

- `session ls --json` and `session logs` are machine-readable and scriptable.
- `session start` without `--wait` returns promptly with a usable id — the default is one-shot, per D6.
- `session send` delivers text to a running session through the SDK, with no terminal handling.
- Killing a session the CLI did not start works.

## Traps

- **Prompt delivery is the Core's job** (D3, phase 1). The CLI must not add its own timing, retry, or Enter-after-pause logic. If a prompt is lost here, it is a Core bug.
- `logs` on a TUI harness needs the SDK's `screen()`, not raw data — raw escape codes concatenate into unreadable soup.
- `--json` means *only* JSON on stdout. Progress and warnings go to stderr, or every consumer has to strip them.

Refs #129

qcentic-dev:
Delivered by #204, which landed on `beta/0.2.2` with the train integration PR #212 (squash merge `6a33b805461006ca8553fcba1141bdecdff07eba`). Closing manually — this PR's base was not the default branch, so it did not auto-close.

PR #203, opened 2026-08-12
feat(cli): actana core shell — the sanctioned escape hatch, over the existing PTY transport

qcentic-dev:
Implements **#162** — `actana core shell`, phase 2 step 2 of #129, on the `feat/cli-and-publishing`
base (not `main`). Train `beta/0.2.2`.

## What this is

A real interactive shell on the Core, over the PTY transport the Panel already uses. F3 is allowed to
call path confinement *an accident guard, not a security boundary* precisely because this command
exists: anybody holding a valid blob for a Core already has that machine.

```
actana core shell            # a login shell on the `current` Core
actana core shell --core prod
```

## The traps, and what holds them down

**Raw mode is restored on every exit path.** Not the happy path — every path. The restore runs from a
`finally` around the whole session, from the `SIGINT` and `SIGTERM` handlers, on a dropped link, and
— one layer down in `cli-terminal.ts` — from a process `exit` hook armed the first time raw mode is
entered, for the route no `finally` covers. The terminal is handed back *before* the best-effort
remote kill, so a teardown never costs a request timeout's worth of unusable shell. There is one test
per path.

**`Ctrl-C` reaches the remote process**, and has no special case in the code. Raw mode turns off
`ISIG`, so it is the byte `0x03` on stdin and is forwarded with every other byte. A `Ctrl-C` that
killed the CLI while the shell was live would make it impossible to interrupt anything *on the Core*.

**Resize propagates** — the size goes out with the spawn, and every `SIGWINCH` becomes a `resize`
frame. A resize that fails does not end a working session; the shell is merely drawing at the old
size.

**The remote exit status comes back** as this process's: `exit 3` in the remote shell exits the CLI
3, and a remote process killed by a signal exits `128 + signal`. A status outside 0–255 becomes a
plain failure rather than a number the remote shell never chose.

**The SDK is untouched (D11).** All terminal handling is CLI-side. `cli-terminal.ts` owns raw mode,
keystrokes, size and signals behind an injected surface; `core-shell-channel.ts` speaks the SDK's
existing PTY vocabulary — `spawn`, `write`, `resize`, `kill`, `onData`, `onExit` — with no TTY near
it. The shell is the Core's existing `shellSession: true` spawn mode: a login shell, no project root,
no harness.

This command holds a session for as long as it runs, which is fine now that #140 has landed.

## Shape

New, and where the substance is:

- `packages/cli/src/cli-terminal.ts` — the terminal surface, plus the real implementation, which
  takes its process as an argument so `actana-cli-entry.ts` stays the only module that knows about
  one (and so the real implementation is itself testable).
- `packages/cli/src/core-shell-channel.ts` — the SDK seam, the same shape as `core-probe.ts`.
- `packages/cli/src/core-shell.ts` — the command.

Shared files are touched as tightly as possible, because #160 (`session`) and #161
(`project`/`harness`/`events`) are in flight against this same base: `core-command.ts` loses the
scaffold stub and gains a two-line dispatch and a help line; `cli-deps.ts` gains two fields;
`actana-cli-entry.ts` binds them. `actana-cli.ts` is not touched at all — no reserved noun changes
here. `packages/sdk`, `release.yml` and `docs/reference` are untouched.

## Testing

`pnpm --filter @actana/cli test` — 148 passing. Repo-wide `pnpm typecheck` and `pnpm lint` are clean
(0 errors).

- `core-shell.test.ts` — the command against a fake terminal and fake channel: one test per exit
  path, `Ctrl-C` forwarding, resize, exit status across `0..255` and `128 + signal`.
- `cli-terminal.test.ts` — the real terminal against a fake process: the exit hook is armed before
  the call that can fail, a multi-byte keystroke split across chunks is not mangled, listeners are
  removed when a session ends.
- `core-shell-live.test.ts` — the real channel against `packages/core`'s real `PtyCoreLinkServer` on
  a real `wss://` port with mTLS and a bearer its own verifier accepts.
- `never-logs-a-blob.test.ts` gains `core shell`, which holds a credential for a whole session.

Two pre-existing local failures are unrelated to this branch: `scripts/__tests__/core-launcher.test.mjs`
(a real `/opt/actana` install on this machine) and one panel test that times out only under full
parallel load and passes on its own.

Refs #162


qcentic-dev:
## Acceptance criteria — issue #162

### Done when

- [x] **A real interactive shell.**
  - [x] **Resize propagates.** The size goes out with the spawn (`cols`/`rows` on the `shellSession`
    spawn) and every `SIGWINCH` becomes a `resize` frame. Covered by `core-shell.test.ts` ("tells the
    Core the size at spawn", "sends a resize for every SIGWINCH") and end to end against a real
    `PtyCoreLinkServer` in `core-shell-live.test.ts` ("propagates a resize to the PTY the Core is
    holding"). A failed resize deliberately does not end a working session.
  - [x] **Ctrl-C reaches the remote process.** Raw mode turns off `ISIG`, so it arrives as the byte
    `0x03` on stdin and is forwarded by the same listener as every other byte — no special case, by
    design. `core-shell.test.ts` asserts the byte goes out *and* that the session stays live;
    `core-shell-live.test.ts` asserts it lands on the Core's PTY.
  - [x] **Exit status comes back.** The remote `exit` frame becomes this process's status: asserted
    across `0, 1, 2, 3, 42, 255`, and `128 + signal` for a remote kill (signal 9 → 137). A status
    outside 0–255 falls back to `EXIT_FAILURE` rather than reporting a number the shell never chose.
- [x] **Raw mode is entered and restored on every exit path.** One test per path in
  `core-shell.test.ts`: remote exit, `SIGINT`, `SIGTERM`, **dropped connection**, a write that starts
  failing, and a terminal that refuses raw mode in the first place (which also kills the shell it had
  just spawned). Plus: the terminal is handed back *before* the best-effort remote kill, so a
  `SIGTERM` teardown never costs a request timeout's worth of unusable shell — asserted directly.
- [x] **The SDK is untouched. All terminal handling lives in the CLI (D11).** `packages/sdk` has zero
  changed lines on this branch. `cli-terminal.ts` owns raw mode, keystrokes, size and signals;
  `core-shell-channel.ts` uses only the SDK's existing PTY vocabulary (`spawn`, `write`, `resize`,
  `kill`, `onData`, `onExit`) with no TTY near it.

### Traps

- [x] **Restore in a `finally`, and on `SIGINT`/`SIGTERM` — not only on the happy path.** All three,
  plus a dropped link, plus a last-resort synchronous `process.on("exit")` hook armed the first time
  raw mode is entered (`cli-terminal.ts`), for the route no `finally` covers. That hook is armed
  *before* the call that can fail, and never throws — asserted in `cli-terminal.test.ts`.
- [x] **Holds a session for as long as it runs; fine now that #140 has landed.** Each invocation
  spawns under its own random `cli_shell_<uuid>` task id, so two shells against one Core never
  collide on the Session write lock (ADR 0024 D6).

### Not checked

- [ ] **No test drives a real login shell (a real `node-pty` `bash`) end to end.**
  `core-shell-live.test.ts` goes through the real blob, real mTLS handshake, real `spawn`/`write`/
  `resize` frames and real `data`/`exit` frames against `packages/core`'s real `PtyCoreLinkServer` —
  but the PTY manager beneath it is a fake, because a real one needs `node-pty`, which is precisely
  the dependency the published CLI exists not to have. The opt-in `live-core.test.ts` is the suite
  that dials an operator's own machine, and it is contractually read-only ("nothing spawns a PTY,
  writes to a Session or claims a lock on a machine somebody may be working on"), so extending it to
  spawn a shell would break its own guarantee. Verifying against a real shell is a manual step.

### Process constraints

- [x] New worktree `wt-162` (the ticket's `wt-129-12` is stale), branch `feat/cli-core-shell`, cut
  from the latest `origin/feat/cli-and-publishing`.
- [x] Draft PR, base `feat/cli-and-publishing` (not `main`), title 93 chars (≤112).
- [x] Conventional Commits on both commits; `commitlint --from ed3837b --to HEAD` reports 0 problems.
- [x] Never force-pushed; `main` untouched; nothing merged; PR left as a draft.
- [x] Boundaries respected: no `packages/sdk`, no `release.yml`, no `docs/reference`, and no command
  files belonging to #160 or #161. `actana-cli.ts` — the router, reserved-noun list and root help —
  is **not touched at all**, so the three PRs should meet at most adjacent-line conflicts. The shared
  edits are confined to `core-command.ts` (a dispatch line, a help line, the removed scaffold),
  `cli-deps.ts` (two fields), `actana-cli-entry.ts` (two bindings) and one `exit-codes.ts` docblock.

qcentic-adm (approved):
Reviewed against issue #162, diffing `fc11503..7b0be1e` on top of the merge-base `ed3837b` with `feat/cli-and-publishing`. I read the enforcing lines rather than the description, and checked the SDK and Core sides of every claim that crosses a package boundary.

**Verdict: approve.** Every "Done when" bullet and both "Traps" are met. The one unchecked box in your checklist — no test drives a real `node-pty` login shell end to end — is honest and should not block; see the last section for why.

---

**A real interactive shell — resize propagates.** *Met.* The size goes out with the spawn (`core-shell.ts:150` reads `terminal.size()` before dialling, `core-shell-channel.ts` passes `cols`/`rows` into `client.spawn`), and `core-shell.ts:264` turns every `SIGWINCH` into a `resize` frame, re-reading the size rather than using a cached one. The failure is deliberately swallowed (`.catch(() => {})`), which is right: a resize that missed means the far side is drawing at the old size, not that the session should end. Verified on the far side too — `pty-core-link-server.ts` `case "resize"` is ungated by the Session lock, so a resize from a CLI shell lands. `core-shell-live.test.ts` asserts the size actually reaches the PTY the Core holds.

**Ctrl-C reaches the remote process, and does not kill the CLI.** *Met, and for the right reason.* There is no special case, which is the correct implementation: raw mode clears `ISIG`, so `0x03` is an ordinary byte on stdin and goes out through the same `terminal.onInput` forwarder as everything else (`core-shell.ts:253`). The `SIGINT` handler at `core-shell.ts:276` exists for `kill -INT` from another window, not for the keyboard, and the comment says so. `core-shell.test.ts` pins both halves — the byte goes out *and* the session is still live and still raw afterwards.

One consequence worth stating out loud rather than fixing: because `Ctrl-C` and `Ctrl-\` are both bytes now, the only ways out of the CLI are the remote shell exiting (`exit`, `Ctrl-D`) or a signal from another window. There is no `ssh`-style `~.` local escape and no line printed on entry saying so. That is a defensible choice — any escape byte is a byte the remote shell can no longer receive — but a one-line hint on stderr before the shell starts would save an operator whose remote shell has wedged from having to find another terminal. Non-blocking.

**The remote exit status comes back.** *Met, including the paths with no clean status.* `report()` covers all five endings: a remote exit returns `clampStatus(exitCode)`, a remote kill returns `clampStatus(128 + signal)` (`core-shell.ts:309-310`), a local `SIGINT`/`SIGTERM` returns `128 + n` → 130/143 (`core-shell.ts:321`), and a dropped link or a write that started failing returns `EXIT_FAILURE` with a line on stderr. `clampStatus` rejects anything outside 0–255 rather than truncating — the right call, since `CoreLinkExitFrame.exitCode` is a bare `number` off the wire with nothing guaranteeing the range. `actana-cli-entry.ts` ends in `process.exit(code)`, so the status really is this process's. Tested across `0, 1, 2, 3, 42, 255`, `128 + 9`, and `4096`.

**Raw mode is entered and restored on every exit path.** *Met.* This is the criterion I pushed hardest on, and the structure holds:

- *Happy exit and remote exit with a nonzero status* — `runSession` settles, `releaseTerminal()` runs at `core-shell.ts:192` before anything else.
- *Dropped connection* — `channel.onDisconnected` settles the same way; `close()` does not fire it (confirmed in the SDK, `core-client.ts:889-892`), so this only means a real drop.
- *SIGINT / SIGTERM* — handled rather than fatal at `core-shell.ts:276`; the default action would have killed the process before any `finally`.
- *A thrown error mid-session* — the `finally` at `core-shell.ts:198` catches everything escaping the `try`, including a rejection from `runSession` (e.g. `terminal.onInput` throwing while wiring up), and it restores the terminal *before* `channel.close()`, so even a throwing `close` leaves a usable terminal.
- *Restore cannot run twice* — `RawModeGuard.restore()` returns early on `restored` and `releaseTerminal` drains `disposers` with `.splice(0)` (`core-shell.ts:181`), so calling it from both the `try` and the `finally` is a no-op the second time. Each disposer is wrapped in `safely()`, so one throwing disposer cannot skip the restore behind it.
- *Restore is skipped, correctly, when the error precedes raw mode* — `guard.enter()` sets `entered = true` only after `setRawMode(true)` returns, so a terminal that refuses leaves `restore()` a no-op (`core-shell.ts:98`). That path instead kills and closes the already-spawned remote shell (`core-shell.ts:165`), which is the right cleanup — the task id was random and nobody could reattach.
- *And the half-entered case* — the one that worried me — is covered a layer down: `cli-terminal.ts:143-145` arms the `process.once("exit")` restore **before** calling `setRawMode`, so a terminal that got raw enough to be unusable but threw on the way is still put back when `process.exit(code)` runs. `cli-terminal.test.ts` asserts exactly that ordering, and that the hook never throws.

Also good, and not something the ticket asked for: the terminal is handed back *before* the best-effort remote kill (`core-shell.ts:192` vs `:197`), so a `SIGTERM` teardown does not cost a request timeout's worth of unusable shell. There is a direct test for it that holds `kill()` open and asserts the terminal is already cooked.

Two residuals, neither blocking, both worth knowing:
- `SIGHUP`/`SIGQUIT` are not handled. Their default action terminates without running `exit` hooks, so raw mode would survive — but with `ISIG` off neither is reachable from the keyboard, so this is the same class as `kill -9`, which nothing can defend against. The ticket names `SIGINT`/`SIGTERM` and those are the ones that matter.
- The SDK swallows listener throws (`core-link-transport.ts:496` `guard()`), so a throw inside your `onData` handler — `terminal.write` failing — is neither an ending nor a crash. In practice the session stays live and the eventual `exit` still settles it, so the terminal is still restored; it just means the `finally` is not what covers *that* particular throw. Worth a sentence in the header if you touch this file again.

**The SDK is untouched; all terminal handling lives in the CLI (D11).** *Met.* `git diff --name-only ed3837b..7b0be1e -- packages/sdk` is empty; every changed path is under `packages/cli/src`. `core-shell-channel.ts` uses only `connect`, `spawn`, `write`, `resize`, `kill`, `onData`, `onExit`, `onDisconnected`, `close` — all pre-existing — and nothing in it touches a TTY, a signal or raw mode. `cli-terminal.ts` is the only module that does, and it takes `ProcessLike` as an argument so `actana-cli-entry.ts` remains the one file that knows about a real `process`. The `shellSession: true` spawn is the Core's existing VM-shell mode; `CoreLinkShellSessionSpawnOptions` already makes `command` optional, so the no-command login shell needed no schema change.

---

## The claims I checked independently

**Shared-file footprint.** *Accurate for source, incomplete for tests.* Source edits are exactly `core-command.ts` (scaffold removed, one dispatch line, one help line), `cli-deps.ts` (two fields), `actana-cli-entry.ts` (two bindings) and `exit-codes.ts` — which I confirmed is docblock-only, `EXIT_UNIMPLEMENTED = 3` unchanged and still used by the reserved nouns in `actana-cli.ts:134`. `actana-cli.ts`, the router, is genuinely not in the diff.

But the checklist's list omits three *shared test* files this PR does edit: `cli-harness.ts` (+107), `actana-cli.test.ts` and `never-logs-a-blob.test.ts`. #205 touches all three as well, so that is where it will meet #205, not only at the two source files.

**Which brings up the delivery fact:** #204 merged into `feat/cli-and-publishing` after this branch was cut, and the PR is now `CONFLICTING` — `git merge-tree` reports conflicts in `cli-deps.ts`, `actana-cli-entry.ts`, `cli-harness.ts`, `actana-cli.test.ts` and `never-logs-a-blob.test.ts`. I checked what #204 actually did to the two source files: it adds an import next to `probeCore` and a field next to `probe`, which is precisely the adjacent-line conflict your checklist predicted. All five are mechanical. CI is green, but it is green against the pre-#204 base, so the rebase wants a re-run before merge rather than a re-review.

**The no-local-escape sweep.** *Still holds, and still complete.* `no-local-escape.test.ts` walks `src/` recursively and excludes only `__tests__`, so `core-shell.ts`, `core-shell-channel.ts` and `cli-terminal.ts` are swept without anyone having to remember them. None of the three imports `child_process`, `worker_threads`, `@actana/core` or `@actana/shared` — the only new imports are `node:string_decoder` and `node:crypto`. The command starts a process on the *remote* Core through a `spawn` frame on the existing core-link and nothing locally, which is the exception the sweep's own header already describes. `core-shell-live.test.ts` does import `@actana/core` and `@actana/shared`, but it is a test, excluded from the sweep, and it rides the existing vitest alias rather than a manifest entry — same arrangement as `in-process-core.test.ts`, and `packages/cli/package.json` is correctly not in this diff.

**Small one:** `deps.stdinIsTty` and `terminal.isTty` are now two ways to ask the same question, and this command uses the second. Not worth churn here, but a candidate for the cleanup after #160/#161/#163 land.

---

## The unchecked box

I judge the compensating tests sufficient, and I would not hold the PR for it.

The reason is that the split makes the untested seam small and named. `core-shell-live.test.ts` runs the real `PtyCoreLinkServer` on a real `wss://` port with real mTLS material and a real bearer — the blob, the handshake, the `spawn` frame, the auto-subscription (which I verified on the Core side: `pty-core-link-server.ts` calls `conn.subscribePty(ptyId, false)` before it answers `spawned`, so skipping `ptySubscribe` is correct, not a lucky omission), and the `data`/`exit` frames coming back are all real, including the cross-PTY isolation check. `cli-terminal.test.ts` covers the half a fake terminal cannot: the exit-hook ordering, the multi-byte keystroke split across chunks, and the listener cleanup. `core-shell.test.ts` covers the endings, which is where the operator-visible damage lives.

What is left untested is only "does a real `bash` behave like a PTY" — and the reason it is left is a good one rather than an omission: the published CLI exists specifically not to depend on `node-pty`, and extending the opt-in `live-core.test.ts` would break that suite's own read-only contract. A manual verification note in the PR body is the right resolution for that box.

qcentic-dev:
### Base merge: `origin/feat/cli-and-publishing` → `feat/cli-core-shell` (e0b1250)

Mechanical base merge only, as the review asked — no behaviour change to `core shell` itself. The base now carries #204 and #205. Six files conflicted; how each was resolved:

**Import-union conflicts (kept both sides verbatim)**
- `packages/cli/src/cli-deps.ts` — kept this branch's `CliTerminal` / `OpenCoreShellFn` alongside the base's `CoreConnectFn` / `OpenSessionGateway`. The `ActanaCliDeps` body auto-merged and carries all four fields (`connect`, `openSessions`, `terminal`, `openShell`).
- `packages/cli/src/actana-cli-entry.ts` — same union of imports; the deps literal auto-merged and registers both sides (`probe`, `connect`, `openSessions`, `terminal`, `openShell`).
- `packages/cli/src/__tests__/cli-harness.ts` — union of the terminal/shell fakes and the base's core-link/session-gateway types.
- `packages/cli/src/__tests__/never-logs-a-blob.test.ts` — union of `fakeTerminal` with `fakeCore` / `fakeSessionGateway` / `fakeStartedSession`.

**`packages/cli/src/exit-codes.ts` (docblock)**
Both sides rewrote the same `EXIT_UNIMPLEMENTED` paragraph. Merged into one that keeps every fact: the "whether noun or verb" framing from this branch, the base's still-reserved `project cp` / `project files` (#168), and the base's added paragraph on `project set-path` exiting `EXIT_USAGE` (ADR 0022). Wording now reflects that #160/#161/#163 built the nouns and #162 built the verb, so each returned `3` only until its build landed.

**`packages/cli/src/__tests__/actana-cli.test.ts` (not purely mechanical — flagging explicitly)**
Both sides' assertions went stale in opposite directions, and taking either verbatim would have left a red suite:
- this branch's test used the `session` noun as its reserved example, but #205 built that noun (`RESERVED_NOUNS` is now `{}`);
- the base's test asserted `project cp` and `core shell` return the same code, but this branch builds `core shell`.

Post-merge the only reservations left are `project cp` / `project files` (`project-command.ts`) and `session attach` (`session-command.ts`). Both tests were kept, repointed onto `session attach` as the reserved stand-in — it returns `EXIT_UNIMPLEMENTED` without dialling, so it still satisfies the "a reservation is answered before a Core is" test. The auto-merged `RESERVED` fixture list had the same problem (it still listed `["core","shell"], "#162"`); that entry is now `["session","attach"], "#163"`. The module header's "what is reserved here is nouns" was corrected for the same reason.

**Rig modules** — #205's unified in-process Core rig (`src/__tests__/in-process-core.ts`) landed cleanly; no suite on this branch imported a removed rig module, so no imports needed repointing. `core-shell-live.test.ts` stands up its own server inline and was untouched.

**Verification (local, in `wt-162`):** `tsc --noEmit` clean, `eslint packages/cli` clean, `vitest run` in `packages/cli` — **22 files passed / 1 skipped, 256 tests passed / 3 skipped**.

Issue #163, opened 2026-08-11
2.7 CLI session attach — claims the Session write lock (D10, needs #144)


mrfqcentic:
**Base branch: `feat/cli-and-publishing`, not `main`.** Train: **`beta/0.2.2`**.
Worktree: **`wt-129-13`**. Suggested branch: **`feat/cli-session-attach`**.
Phase 2, step 3, alone. Implements `session attach` of **D10**.

## What

`actana session attach <id>` — attach a terminal to a running session, read its output live, and type into it.

Deliberately last in the phase: it has no headless test story, so it is the ticket most likely to need a person, and nothing else waits on it.

## This ticket depends on #144

`attach` writes to a Session, so it must **claim the Session write lock** (#140, D3–D7) and release it on detach. If the lock is held by another Core client, `attach` opens read-only and says so.

**#129 D5's original rationale is dead by the time this lands.** D5 justified itself with "the Core accepts one connection at a time"; after `beta/0.2.0` that is false. The *decision* — the SDK handles one Core, no `Fleet` type — still stands on its own merits. Do not infer from D5's text that attach owns the Core.

## Done when

- Attaching to a session someone else is writing gives a read-only view with a visible reason, not an error and not a silent steal.
- Detaching releases the lock. So does the connection dropping.
- Two `attach` sessions on the same Session: one writes, one reads.
- The terminal is restored on every exit path.

## Traps

- **An attach that assumes it holds the only connection will corrupt the shared view** once #140 is in. Write against the lock from the first commit; do not add it afterwards.
- Lock release on an abrupt disconnect is the case that gets missed and the one that strands a Session unwritable.

Refs #129

qcentic-dev:
Delivered by #206, which landed on `beta/0.2.2` with the train integration PR #212 (squash merge `6a33b805461006ca8553fcba1141bdecdff07eba`). Closing manually — this PR's base was not the default branch, so it did not auto-close.

PR #204, opened 2026-08-12
feat(cli): actana session — start, ls, logs, resume, kill and send, with the Core owning prompt delivery

qcentic-dev:
Implements the `session` noun of #129 D10, minus `attach` (#163). Base is
`feat/cli-and-publishing`, train `beta/0.2.2`.

    actana session start <project> [prompt]   start a Session; prints its id
    actana session ls [project]               list Sessions on this Core
    actana session logs <session>             the transcript, rendered
    actana session resume <session> [prompt]  continue a conversation
    actana session send <session> <text>      write text into a running Session
    actana session kill <session>             stop the harness running for it
    actana session attach <session>           in the tree, exits 3, names #163

## What it does

**`start` exits by default** (#129 D6), printing a Session id; the harness
carries on without the process that asked for it. `--wait` blocks until the Core
reports the Session settled, and `--wait --json` emits that final state as one
object with the transcript in it — the Core's replay ring lives with the PTY, so
a `--json` caller has no second chance to read it.

**`--json` means only JSON on stdout**, on the failing path as much as the happy
one: one document per verb, `{"error": …}` when it could not do what it was
asked, and every progress line on stderr. Without `--json`, stdout carries the
Session id and nothing else, so `TASK=$(actana session start web "fix it")`
works with `--wait` as well as without it.

**`kill`, `send` and `logs` work on any Session on the Core**, including ones a
Panel or another terminal started. Nothing about having started a Session is
remembered locally — a Session is named by Task id and resolved through the
Core — so "a Session this CLI did not start" is not a case the code has.

## The traps, and what holds them

**Prompt delivery is the Core's** (ADR 0026, #129 D3). A starting prompt is
handed to the SDK and that is the whole of this CLI's involvement: no timing, no
retry, no Enter after a pause. `send` writes exactly the bytes it was given;
`--enter` writes a carriage return as a *separate* write because the operator
asked for one in that invocation. `no-prompt-timing.test.ts` makes this a
property of the package rather than a paragraph in a header — nothing shipped
here schedules anything, and the clock is read in one file to print an age.

**`logs` renders a screen**, through the SDK's `TerminalScreen`, because a
harness paints with cursor moves. The in-process-Core suite asserts both
directions on a repainted line: rendered it reads `done: 3 files changed`, raw it
reads `Scanning…Scanning… 2 filesdone: 3 files changed`.

**`--wait` is the Core's report**, `waitForIdle` watching the event log — never a
guess from how quiet the output went. `--wait-timeout` is a deadline the operator
chose, handed to the SDK, and its expiry is reported as this side giving up
rather than as a status the Core never sent.

## Tests

`pnpm --filter @actana/cli test` — 149 passing, 3 skipped (the opt-in live-Core
suites). Typecheck and lint are clean across the workspace.

- `session-command.test.ts` — 30 cases over the flags, tables, `--json` shapes
  and exit codes, with the gateway injected and no socket anywhere.
- `in-process-core-session.test.ts` — the same verbs against `packages/core`'s
  real `PtyCoreLinkServer` on a real wss:// port, holding a PTY that was already
  running: `ls`, `logs` (rendered and `--raw`), `send` (exact bytes, no return),
  and `kill` on a Session this CLI did not start.
- `harness-resume.test.ts` — every resume command run through the Core's own
  `resolveSpawnPlan`, so a table that drifts from the allow-list fails here
  rather than on a Core.
- `never-logs-a-blob.test.ts` — extended to sweep every session verb with
  `--verbose`, since each one resolves a credential.

## Boundaries

Touches `packages/cli` only. The shared router and root help take the single
registration for the noun; `packages/sdk`, `release.yml` and `docs/reference` are
untouched, and no file belonging to #161 or #162 is created or edited.

Refs #160

### DIFF
--- a/packages/cli/src/__tests__/actana-cli.test.ts
+++ b/packages/cli/src/__tests__/actana-cli.test.ts
@@ -25,7 +25,8 @@ afterEach(() => {
 
 /** Every noun reserved in the tree, and the ticket each names. */
 const RESERVED = [
-  ["session", "#160"],
+  // `session` left this list when #160 built it — the row moving out is what a
+  // noun landing looks like from here.
   ["project", "#161"],
   ["harness", "#161"],
   ["events", "#161"],
@@ -43,7 +44,7 @@ describe("a reserved noun is `not built yet`, not `you typed it wrong`", () => {
   });
 
   it("agrees with `core shell`, which is the same fact about the same build", async () => {
-    const noun = await cli().run(["session"]);
+    const noun = await cli().run(["project"]);
     const verb = await cli().run(["core", "shell"]);
     expect(noun.code).toBe(verb.code);
   });
--- a/packages/cli/src/__tests__/cli-harness.ts
+++ b/packages/cli/src/__tests__/cli-harness.ts
@@ -12,6 +12,11 @@ import path from "node:path";
 import { runActanaCli } from "../actana-cli.ts";
 import { registryPaths, type RegistryPaths } from "../blob-registry.ts";
 import type { CoreProbe, CoreProbeFn } from "../core-probe.ts";
+import type {
+  OpenSessionGateway,
+  SessionGateway,
+  StartedSession,
+} from "../session-gateway.ts";
 
 /** One run's captured output, plus the exit code. */
 export type CliRun = {
@@ -44,8 +49,54 @@ export type RunOptions = {
   stdinIsTty?: boolean;
   /** What `core status` gets back, or a throw. */
   probe?: CoreProbeFn;
+  /** What the `session` noun's verbs get back, or a throw. */
+  sessions?: OpenSessionGateway;
   now?: number;
 };
+
+/**
+ * A Session that started, for a fake gateway to hand back.
+ *
+ * `wait` resolves with whatever the test says the Core reported — the CLI never
+ * decides idleness, so a fake has to be the one holding the answer.
+ */
+export function fakeStartedSession(overrides: Partial<StartedSession> = {}): StartedSession {
+  return {
+    taskId: "task_1",
+    ptyId: "pty_1",
+    harness: "claude-code",
+    command: "claude",
+    projectId: "proj_1",
+    project: "web",
+    wait: async () => ({ status: "finished", exited: false }),
+    screen: () => "the transcript",
+    dispose: () => {},
+    ...overrides,
+  };
+}
+
+/**
+ * A gateway that answers without a Core.
+ *
+ * Every verb throws by default and a test overrides the one it is about: a
+ * suite that reaches a verb it did not mean to exercise should fail loudly
+ * rather than pass against a stub that said yes.
+ */
+export function fakeSessionGateway(overrides: Partial<SessionGateway> = {}): OpenSessionGateway {
+  const refuse = (verb: string) => async () => {
+    throw new Error(`this test did not expect session ${verb}`);
+  };
+  return async () => ({
+    list: refuse("ls"),
+    start: refuse("start"),
+    resume: refuse("resume"),
+    logs: refuse("logs"),
+    send: refuse("send"),
+    kill: refuse("kill"),
+    close: () => {},
+    ...overrides,
+  });
+}
 
 /** A probe that answers like a healthy Core on the current protocol. */
 export function healthyProbe(overrides: Partial<CoreProbe> = {}): CoreProbeFn {
@@ -87,6 +138,11 @@ export function makeCliFixture(): CliFixture {
           opts.probe ??
           (async () => {
             throw new Error("this test did not expect to dial a Core");
+          }),
+        openSessions:
+          opts.sessions ??
+          (async () => {
+            throw new Error("this test did not expect to open a session gateway");
           }),
         now: () => opts.now ?? Date.UTC(2026, 7, 12),
       });
--- a/packages/cli/src/__tests__/harness-resume.test.ts
+++ b/packages/cli/src/__tests__/harness-resume.test.ts
@@ -1,0 +1,115 @@
+// The resume commands, checked against the machine that has to accept them.
+//
+// `harnessResumeCommand` builds a launch command per harness, and the Core's
+// spawn policy is what decides whether that command may run. A table that
+// drifts from the allow-list produces a `session resume` that always fails, and
+// it fails on the *Core*, where a CLI unit test would never see it — so this
+// suite runs every command it builds through `resolveSpawnPlan`, the same
+// function `packages/core` calls before spawning anything.
+//
+// Importing `@actana/shared` from a test is the arrangement `tsconfig.json` and
+// `vitest.config.ts` already describe and `no-local-escape.test.ts` enforces:
+// test-only, never from a shipped module.
+
+import { describe, it, expect } from "vitest";
+import {
+  resolveSpawnPlan,
+  type SpawnPolicyDeps,
+  type SpawnRequest,
+} from "@actana/shared/pty-spawn-policy";
+import { harnessResumeCommand } from "../harness-resume.ts";
+import { KNOWN_HARNESSES } from "../session-gateway.ts";
+import type { CoreLinkPtySpawnHarness } from "@actana/sdk/core-link-frames.ts";
+
+const PROJECT_ROOT = "/home/core/projects/web";
+
+/** Session ids in the shape each harness actually reports one. */
+const SESSION_IDS: Record<CoreLinkPtySpawnHarness, string> = {
+  "claude-code": "00000000-0000-4000-8000-000000000000",
+  // The Core validates a codex resume id against its model-id pattern, which a
+  // UUID satisfies; a spelling it rejects is rejected there, not trimmed here.
+  codex: "019d7a0f-432a-7fa1-a821-b7841f983967",
+  "cursor-cli": "00000000-0000-4000-8000-000000000000",
+  // OpenCode ids carry a `ses` prefix and the Core insists on it.
+  opencode: "ses_01H8Z0MEXAMPLE",
+};
+
+function policyDeps(): SpawnPolicyDeps {
+  return {
+    cwdExists: () => true,
+    realpath: (p) => p,
+    projectRoots: () => [PROJECT_ROOT],
+    resolveCommand: (name) => `/usr/local/bin/${name}`,
+    resolveShell: () => ({ shell: "/bin/zsh", shellArgs: (cmd) => (cmd ? ["-l", "-c", cmd] : ["-l"]) }),
+  };
+}
+
+function planFor(harness: CoreLinkPtySpawnHarness, command: string, skip: boolean) {
+  return resolveSpawnPlan(
+    {
+      taskId: "t1",
+      cwd: PROJECT_ROOT,
+      command,
+      agent: harness,
+      ...(skip ? { dangerouslySkipPermissions: true } : {}),
+    } as SpawnRequest,
+    policyDeps(),
+  );
+}
+
+describe("harnessResumeCommand", () => {
+  it("spells resumption the way each harness spells it", () => {
+    expect(harnessResumeCommand("claude-code", "abc")).toBe("claude --resume abc");
+    // A subcommand, so it comes first — and the hooks flag stays on, or the
+    // resumed Session never reports its lifecycle and every wait on it hangs.
+    expect(harnessResumeCommand("codex", "abc")).toBe("codex resume abc --enable hooks");
+    expect(harnessResumeCommand("cursor-cli", "abc")).toBe("cursor-agent --resume abc");
+    expect(harnessResumeCommand("opencode", "ses_abc")).toBe("opencode --session ses_abc");
+  });
+
+  it("appends each harness's own skip-permissions flag, and none for OpenCode", () => {
+    const skip = { dangerouslySkipPermissions: true };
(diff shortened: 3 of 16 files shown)


qcentic-dev:

Every criterion in #160 — the four **Done when** bullets, the four surface
statements under **What**, and the three **Traps** — with what makes each one
true and where it is proved.

### Done when

- [x] **`session ls --json` and `session logs` are machine-readable and scriptable.**
  `ls --json` emits one array of objects (`taskId`, `title`, `project`,
  `harness`, `status`, `ptyId`, `live`, `lock`, `writable`, `updatedAt`);
  `logs --json` emits one object carrying the rendered screen, and `--raw` swaps
  the bytes in for a caller piping into a terminal. Covered in
  `session-command.test.ts` and against a real Core in
  `in-process-core-session.test.ts`.
  *Caveat, stated rather than hidden:* `logs` reads the Core's replay ring,
  which belongs to the harness's PTY, so a Session that has already exited has
  no transcript left to print. `session --help` says so, the error says so, and
  `start --wait --json` carries the settled screen for the case where a caller
  needs to keep one.

- [x] **`session start` without `--wait` returns promptly with a usable id — the
  default is one-shot, per D6.** It returns as soon as the Core has the Session
  running, prints the id on stdout and closes the socket; the harness carries on
  without it. `session-command.test.ts` asserts the wait was never entered, and
  that stdout is exactly `["task_1"]` — every progress line went to stderr, so
  `TASK=$(actana session start …)` is the id and nothing else.

- [x] **`session send` delivers text to a running session through the SDK, with
  no terminal handling.** `send` resolves the Task to its PTY through the Core
  and calls `CoreSession`'s write path. Nothing reads a TTY, sets raw mode or
  opens `/dev/tty` — that is `attach` (#163). The real-Core suite asserts the
  bytes that reached the PTY are exactly the bytes typed.

- [x] **Killing a session the CLI did not start works.** A Session is named by
  Task id and resolved through the Core, and nothing about having started one is
  remembered locally — so this is not a case the code has. Proved in
  `in-process-core-session.test.ts` against a PTY that was already running before
  the command was typed: the kill lands, the second kill reports there is no
  harness running rather than claiming a success.

### What

- [x] **`actana session start ls logs resume kill send`.** All six, plus
  `attach` in the tree exiting `3` and naming #163 — `EXIT_UNIMPLEMENTED`, so a
  script can tell "not built yet" from "you typed it wrong" without reading
  English off stderr.
- [ ] **`attach`** — *unchecked on purpose*: the ticket makes it "deliberately a
  separate, later ticket" (#163). It is registered and refuses; nothing here
  implements it.
- [x] **`session start` exits by default, printing a task id.** As above.
- [x] **`--wait` blocks until the session finishes.** `--wait` is the SDK's
  `waitForIdle` watching the Core's event log — the Core's report, not a guess
  from how quiet the output went. `--wait-timeout <s>` is an operator-chosen
  deadline handed to the SDK; its expiry is reported as this side giving up,
  never as a status the Core did not send.
- [x] **`--wait --json` emits the final state as one object.** One document:
  the identity fields, `waited: true`, `status`, `exited`, `exitCode`, and the
  screen as it settled.

### Traps

- [x] **Prompt delivery is the Core's job (D3) — no CLI timing, retry, or
  Enter-after-pause.** The prompt travels as `initialInput` with the spawn and
  the Core delivers it (ADR 0026); `send` writes exactly what it was given.
  `no-prompt-timing.test.ts` makes this a property rather than an intention:
  nothing shipped in `packages/cli` calls `setTimeout`, `setInterval`,
  `setImmediate` or `queueMicrotask`, no source appends a carriage return to
  text, and `Date.now()` is read in one file to print an age.
  *One thing a reviewer should look at deliberately:* `send --enter` writes a
  carriage return as a **separate** write, in that invocation, because the
  operator asked for it — no pause, no waiting for the harness to look ready,
  and nothing that decides on its own that Enter is due. If that reads as too
  close to the line, it is one flag to drop.

- [x] **`logs` on a TUI harness needs the SDK's `screen()`, not raw data.** The
  replay ring is fed through the SDK's `TerminalScreen` — the same emulator
  `CoreSession.screen()` is built on. The real-Core suite asserts both
  directions on a repainted line: rendered it reads `done: 3 files changed`; the
  raw bytes read `Scanning…Scanning… 2 filesdone: 3 files changed`.
  *Caveat:* the screen is built at the Core's default PTY geometry, because the
  protocol carries no way to ask a PTY its size, so a Session started at another
  width wraps differently here than there. `--raw` is the exit.

- [x] **`--json` means only JSON on stdout.** Every verb emits exactly one
  document — `{"error": …}` on the failing path, which is the path that usually
  forgets — and every progress line, warning and `--verbose` note goes to
  stderr. Asserted for all six verbs, with `--verbose` on, both when they
  succeed and when the Core refuses.

### Boundaries

- [x] Touches `packages/cli` only. `packages/sdk`, `release.yml` and
  `docs/reference` are untouched.
- [x] No file belonging to #161 (`project`, `harness`, `events`) or #162
  (`core shell`) is created or edited. The shared router takes six lines: the
  import, the `case "session"`, one help line, and the `session` row leaving the
  reserved list. `cli-args.ts` and `cli-deps.ts` gain this noun's flags and its
  one injected dependency, in single contiguous blocks.

qcentic-adm (approved):
Reviewed against #160 in full, at the enforcing lines rather than the description. Approving.

## The claims I set out to disprove, and what I found

**1. D3 — no CLI timing, retry, or Enter-after-a-pause.** Holds, and it is checkable in one command. Across every shipped module in `packages/cli` there is exactly one carriage return and one clock read:

- `session-command.ts:451` — `if (delivered && args.enter) delivered = await gateway.send(taskId, "\r")`
- `actana-cli-entry.ts:43` — `now: () => Date.now()`, injected, consumed by `age()` and the bearer line

No `setTimeout`/`setInterval`/`setImmediate`/`queueMicrotask`, no `node:timers`, no sleep, no retry loop anywhere in shipped source. `session-gateway.ts:326-331` writes verbatim and appends nothing; the starting prompt goes out as `initialInput` on the spawn frame (`begin()` -> `CoreSession.start`, SDK `core-session.ts:401`) and the Core owns delivery from there.

On the flag you asked me to judge: **`--enter` is on the right side of the line, and I would keep it.** The rule ADR 0026 moved to the Core is *deciding when a prompt is submitted*. `--enter` decides nothing — it is requested by the operator in that invocation, there is no pause between the two writes, nothing inspects the harness for readiness, and it is absent by default (`in-process-core-session.test.ts:215-227` asserts a bare `send` produces exactly `["2"]`). What would cross the line is a return glued on by default, or a return conditioned on how the screen looked. Neither exists.

One hardening note on the guard, not the code: `no-prompt-timing.test.ts:62-68` matches `setTimeout(`, so `import { setTimeout as sleep } from "node:timers/promises"` would slip past it, and the carriage-return matchers at :107-112 are keyed to variables named `text|prompt|data|input`. Adding a sweep for `timers` / `timers/promises` import specifiers would close the cheapest evasion. Non-blocking.

**2. `logs` renders through the SDK.** `session-gateway.ts:321-323` builds a `TerminalScreen` at `DEFAULT_COLS`/`DEFAULT_ROWS` and returns `terminal.text()`; `raw` is the untouched `replay.data`. Raw is opt-in at `session-command.ts:405` and `:410`. `in-process-core-session.test.ts:197-213` asserts both directions on a genuinely repainted line and — the assertion that matters — that no ESC byte survives in the rendered form, i.e. it was drawn, not stripped. The fixed geometry caveat is disclosed in the code and the help rather than buried.

**3. `--json` purity.** One funnel: `withGateway` (`:549-560`) catches *every* throw, not just `SessionGatewayError`, and `failed()` (`:564-568`) is the single place a document is emitted on the failure path. `session-command.test.ts:479-510` drives all six verbs with `--json --verbose` through a refusing gateway and `JSON.parse`s the whole of stdout — concatenated documents or a stray progress line would throw. `in-process-core-session.test.ts:259-273` repeats it against a real Core. Met.

Two exceptions I want on the record, neither introduced here: verb-level usage errors (`usage()`, `:571-574`) emit *zero* documents under `--json`, and `session --json` with no verb, or `--help --json`, writes the help prose to stdout (`:120-123`). `core-command.ts:78-80` on the base branch does exactly the same, so this is house convention rather than a regression in this PR — but a consumer doing `json.loads(stdout)` still gets an exception on a mistyped flag. Worth one ticket that settles the rule for both nouns at once.

**4. `start` default and `--wait`.** `reportStartedSession:256-271` returns as soon as the gateway resolves, prints the bare id, and never enters the wait — asserted with a `waited` sentinel and `expect(run.out).toEqual(["task_1"])` at `session-command.test.ts:152-173`. `CoreSession.start` returns on the spawn ack, and `PtyCoreLinkServer` retains PTY state across disconnects, so hanging up does not orphan the harness. `--wait` is `session.wait` -> `waitForIdle` on the Core's event log; `--wait-timeout` is a number handed to the SDK (`:612-620`) and its expiry is reported as this side giving up with no invented status (`:282-284`, test at `:282-298`). No CLI-side polling exists. Met.

**5. Killing a Session this CLI did not start.** There is no local memory to fail to recognise one: `kill` resolves `findByTask` and calls `client.kill` (`session-gateway.ts:333-342`). Proved against the real `PtyCoreLinkServer` on a PTY that predates the command, including the second kill reporting `not-running` instead of a false success (`in-process-core-session.test.ts:229-248`). Met.

**6. Shared-file footprint.** Confirmed by trial merge (`git merge-tree`) against both siblings. `actana-cli.ts`, `cli-args.ts`, `cli-deps.ts` and `actana-cli-entry.ts` conflict with #205 only as adjacent-line additions in the same regions — import block, `RESERVED_NOUNS`, `ROOT_HELP`, the `switch`, the flag struct — every one a union-resolve. Against #203 it is `cli-deps.ts` and `actana-cli-entry.ts`, same shape.

The one that is **not** adjacent-line, and is worth agreeing now rather than at merge time: this PR extracts the in-process Core into `__tests__/in-process-core-harness.ts` while #205 extracts the same thing into `__tests__/in-process-core.ts`, with overlapping exports (`CORE_ID`, `unusedPtyCore`, `freePort`, `startInProcessCore`) and both gutting `in-process-core.test.ts`. Whichever lands second has to delete one module and repoint every import — a decision, not a union. Test-only and cheap, but it should be a deliberate choice by one of the two authors rather than a surprise.

**7. No-local-escape and dependencies.** `no-local-escape.test.ts` walks `src` recursively excluding `__tests__`, so `session-command.ts`, `session-gateway.ts` and `harness-resume.ts` are swept without anyone having to remember them, and the same shape gives `no-prompt-timing.test.ts` its coverage. `dependencies` is still asserted to equal exactly `["@actana/sdk", "ws"]`, and `package.json` is not in this diff. The only `@actana/core` and `@actana/shared` imports are under `__tests__`. Met.

## Smaller things, none blocking

- `send --enter` re-resolves the PTY for the second write (two `findByTask` round trips). If the harness exits between them the text lands and the return fails, so the command reports failure after a partial delivery. `send(taskId, text, { enter })` resolving the PTY once would remove the window — `session-gateway.ts:326`.
- `session-command.ts:450`: an empty stdin gives `delivered = true` and `characters: 0` without the Core being contacted at all. Harmless, slightly untrue.
- `--dangerously-skip-permissions` on `resume` for OpenCode is silently a no-op (`harness-resume.ts:55-56`). The reasoning is right — an invented flag would be refused by the allow-list — but a one-line stderr note would beat silence.
- `harness-resume.test.ts` running every built command through the Core's own `resolveSpawnPlan`, including the negative case where the flag travels without the option, is the strongest test in this PR. Worth copying wherever a client builds a command a Core has to accept.

## Delivery

Four commits, each doing one thing, messages that carry the reasoning. CI is green across all 20 checks, including Unit Tests, Typecheck, Lint, Conventions and Train rules — I did not re-run the suite locally. The author's checklist is accurate against the code; its one unchecked box, `attach`, is #163 by the ticket's own instruction and is scaffolded to exit `EXIT_UNIMPLEMENTED` naming that issue. Boundaries hold: `packages/cli` only, SDK and `docs/reference` untouched.

Every acceptance criterion in #160 is met. Approving.

qcentic-dev:
Both cheap notes fixed; the third left as informational.

**`1c8c850` — `send --enter` resolves the PTY once.** `send(taskId, text, { enter })` now takes the flag, resolves `livePty` a single time and issues both writes against that one id, so the window where the text lands and the return fails is gone (`session-gateway.ts`). The return is still a separate write of its own byte — nothing appended to the text, no pause, nothing inspecting the harness — so `no-prompt-timing.test.ts` is unmoved and the real-Core assertion still reads `["2", "2", "\r"]` at the PTY. The unit test now asserts the shape that matters, one gateway call rather than two.

**`536230c` — empty stdin is no longer a silent success.** `session send task -` with nothing piped in reported `delivered: true`, `characters: 0` without contacting the Core at all. It is now a usage error naming the cause ("stdin was empty"), beside the existing message for no text at all. `--enter` with empty stdin still sends the bare carriage return, since that one does ask for a write and gets one.

**`--dangerously-skip-permissions` on OpenCode:** left as is. The note is fair, but the harness is resolved inside the gateway from the Task, so the command has no name to check at the point it would warn — it is a small piece of plumbing rather than a one-liner, and better as its own change than folded into an approved PR.

`packages/cli`: 151 passed, 3 skipped (`live-core.test.ts`, which wants an operator's own Core). Typecheck and lint clean.

Issue #168, opened 2026-08-11
3.4 CLI project cp and project files (F12)


mrfqcentic:
**Base branch: `feat/project-files`, not `main`.** Train: **`beta/0.3.0`**.
Worktree: **`wt-129-18`**. Suggested branch: **`feat/cli-project-cp`**.
Phase 3, step 3. Implements the CLI half of **F12**.

## What

- `actana project cp <src> <project>:<path>` — and the reverse direction.
- `actana project files` — the listing.

**No root-level `actana cp`.** D8's noun grammar holds: every client verb hangs off a noun.

## Done when

- A folder copies both ways and keeps its executable bits.
- Progress is visible on a terminal and absent under `--json`.
- Every overwrite is named in the output — silently clobbering a file is the failure mode F5 exists to prevent.
- `project files --json` is machine-readable, like every other list command in the CLI.

## Traps

- The `<project>:<path>` form has to be unambiguous against a Windows-style local path and against a file literally containing a colon. Decide the parse and test it.
- A transfer refused by the one-write-per-Project rule should say who holds it, not just "busy".
- Do not add client-side path validation. The Core owns the disk and owns confinement (F3, F11).

Refs #129

qcentic-dev:
Done in PR #220, merged into `feat/project-files` — a non-default base, so this did not auto-close; closing by hand.

Issue #104, opened 2026-08-05
feat(core-link): carry a Core-validated path so a Project can be moved after it is created


qcentic-dev:
Split out of #98 / PR #103, which closed the half-save it caused but not the underlying gap.

## What is missing

A Project's `path` is set by `op: "create"` and no project mutation op carries one afterwards. `CoreLinkProjectMutation` has `rename`, `pin`, `settings`, `archive` and now `appearance` — none of them a path. So a Core-owned Project's folder is immutable for the lifetime of the row.

## What PR #103 did about it

Made the UI honest rather than pretending otherwise. The Edit-project dialog disables the Working directory field and drops Browse… for a Core-owned project (`pathIsImmutable` in `ProjectDialog.tsx`), and the missing-path dialog's advice branches on ownership — "point the project at its new location from Edit project" is only offered where the Panel owns the row. Recorded in ADR 0022.

That stops the silent discard. It does not give the operator the thing they wanted.

## Why it matters

The case is real and not rare: a project folder gets moved or renamed on the Core, and the Panel's only remedy today is **remove the project and add it again at the new path** — which, because `archive` cascades, takes every Session under it with it. An operator loses their session history to fix a typo in a directory name.

## The shape it probably wants

A `path` op on `CoreLinkProjectMutation`, validated Core-side by the existing `validateProjectPath` (absolute, resolvable, not a file) exactly as `create` already validates it — the Panel has no standing to judge a VM path — with its own `project:pathChanged` event kind, per the `pin` / `settings` / `appearance` precedent. Protocol minor bump; no migration (the column exists).

Open questions for whoever picks this up:

1. **What happens to running Sessions under the project?** Their PTYs have the old path as cwd. Refuse the move while any Session is live, or allow it and let existing PTYs keep the old cwd until they exit?
2. **Worktrees and per-Session state keyed on the old path** — is anything else on the Core storing a derived path that would need moving with it?
3. **Does the missing-path dialog become the natural home for this?** That is where an operator actually is when they discover the folder moved, and it already has the project in hand.

## Acceptance

- Editing a Core-owned Project's Working directory works, and an invalid path comes back as the Core's own error rather than a Panel-side guess.
- Every Panel watching that Core converges on the new path, including after reconnect (event replay).
- Panel-local projects keep their current PATCH behaviour.
- The `pathIsImmutable` gate and the ownership-branched missing-path copy from #103 come back out.

qcentic-dev:
> *This was generated by AI during triage.*

[...]

PR #205, opened 2026-08-12
feat(cli): actana project, harness and events — --json on every list (D10)

### LINKED ISSUE #161: 2.5 CLI project, harness and events nouns — --json on every list (D10)

qcentic-dev:
Closes #161. Phase 2, step 2 — the `project`, `harness` and `events` nouns of [#129](https://github.com/actana/control/issues/129) D10.

Base is `feat/cli-and-publishing`, not `main`. Train `beta/0.2.2`.

## What landed

```
actana project ls [--json]                list the Projects a Core owns
actana project add <name> <path>          register one at a path on the Core
actana project browse [path] [--json]     walk the Core's disk, over dirList
actana harness ls [--json]                what the Core has on PATH, and lacks
actana harness install <id>               install one, and wait for the verdict
actana events tail [--json] [--since|--kind|--limit]
```

Every list command emits `--json`, and `--json` means **only JSON on stdout**: progress, warnings, link-drop notices and `--verbose` all go to stderr. `events tail` emits NDJSON, because a stream that may never end cannot be one array.

## The traps, and what was done about them

**The event cursor.** `events tail` is the CLI's first consumer of it, and a wrong `sinceSeq` neither throws nor logs — too low is a replay storm, too high stops live push silently. So the number is not computed in the CLI: `DurableCoreClient` owns it (D6), and this adds the two decisions a *command* still has to make. Where a first run starts — with no stored cursor and no `--since`, at the end of the log the way `tail -f` does, learning the tip by subscribing and reading the `eventsReplayed` marker, printing none of the tail it took to get there. And where the cursor lives — one file per Core beside the blob registry, written synchronously with the line it accounts for, so a second run resumes rather than replays.

`events-tail-cursor.test.ts` proves it against a Core that really drops: three events live, the Core stopped, two appended while nothing is connected, a second Core started on the same port with the same log, one more after. Six events, each exactly once, in order.

**A Core-owned Project's path is immutable (ADR 0022).** No op carries a path after `create`, so this noun offers no verb that would edit one. `project edit`, `mv`, `move`, `set-path` are caught and answered with the fact and with #104 — at `EXIT_USAGE`, not `EXIT_UNIMPLEMENTED`, because this is not a verb waiting on a ticket but one the protocol cannot carry.

**`project browse` walks the Core's disk.** Over `dirList`, the same frame the Panel's folder picker uses. `project add` sends its path exactly as typed — nothing resolves it against the working directory and nothing stats it locally, because this is not the machine it names. The one local check is that it is absolute, which is a refusal to guess rather than validation. The live suite's Core answers for `/srv/core-disk`, a path that exists on no machine, so a `browse` that listed the operator's own filesystem could not pass.

**`harness install` reports a real exit status, and #31 / #128 are not papered over.** It exits 0 only when the Core reports the Harness *available on its own PATH* — the ack means "started", and an installer that exits 0 leaving nothing behind is a failed install. It subscribes before it asks and discards the replay tail, or an install that failed an hour ago would be read as this one's verdict. Progress goes to stderr on a tick that also re-reads the availability map, so a Core with no event log still reaches a verdict. A failure names the Harness, quotes what the Core said, and links the issue — #31 for opencode specifically, #128 for the install path in general.

## Boundaries

- No file under `packages/sdk`, `release.yml` or `docs/reference` is touched.
- No `session` or `core shell` file is created or edited — #160 and #162 own those.
- The shared router edit is the minimum: three `case` arms, three help lines, and the three nouns leaving `RESERVED_NOUNS`. `session` stays reserved.
- `project cp` and `project files` are reserved in the tree for #168 (phase-3 guardrail 3), exiting `EXIT_UNIMPLEMENTED`.

## Checks

`pnpm --filter @actana/cli typecheck`, `pnpm lint` (0 findings in `packages/cli`) and the CLI suite: **146 passed, 3 skipped**, including the live in-process-Core suites. `commitlint --from origin/feat/cli-and-publishing --to HEAD` reports 0 problems across all three commits.


qcentic-dev:
## Acceptance criteria — #161, item by item

### What (the surface D10 names)

- [x] **`actana project ls add browse`** — `packages/cli/src/project-command.ts`. `ls` and `browse` both take `--json`; `add` takes it too, so a script that creates a Project gets the row back.
- [x] **`actana harness ls install`** — `packages/cli/src/harness-command.ts`. `ls --json` is an array sorted by id; `install --json` emits one verdict object.
- [x] **`actana events tail`** — `packages/cli/src/events-command.ts`, with `--since`, `--kind` (repeatable) and `--limit`.

### Done when

- [x] **Every list command emits `--json`.** `project ls`, `project browse`, `harness ls` and `events tail` all do, and `--json` means *only JSON on stdout*: progress, warnings, reconnect notices and `--verbose` all go to stderr. `events tail` is NDJSON — one complete JSON value per line — because a stream that may never end cannot be a single array. Asserted in `project-command.test.ts`, `harness-command.test.ts` and `events-command.test.ts`, including a case that parses stdout after a `--verbose` run and after a link drop.

- [x] **`project browse` walks the Core's disk through `dirList`, not the operator's.** The frame goes out as `{type: "dirList", path}` and the listing is rendered as the Core answered it; with no argument the *Core* chooses the starting folder (`path: null`), because a default picked here would be a path off this machine's layout. `project-harness-live.test.ts` runs it against a real `PtyCoreLinkServer` whose directory port answers for `/srv/core-disk` with folders that exist on no filesystem anywhere — a browse that listed the operator's own disk cannot pass it.

- [x] **`events tail` follows from a cursor and survives a reconnect without duplicating or dropping.** The cursor is `DurableCoreClient`'s (D6) — sent on every connection, deduped against, advanced on each event and on the `eventsReplayed` marker. The CLI adds the file-backed store (`event-cursor-file.ts`, written synchronously with the line it accounts for) and the first-run rule: no stored cursor and no `--since` starts at the *end* of the log, learning the tip from the marker and printing none of the tail it took to get there. `events-tail-cursor.test.ts` is the proof, against a Core that really drops — three events live, the Core stopped, two appended with nothing connected, a second Core on the same port with the same log, one more after: **six events, each exactly once, in order**. Two more cases cover resuming across runs (second run prints only #6, not #4–#5 again) and `--since` deliberately *not* moving the stored cursor.

- [x] **`harness install` reports progress and a real exit status; #31 and #128 are not papered over.** Exit 0 happens only when the Core reports the Harness available on its own PATH — the ack means "started", and an installer that exits 0 leaving nothing behind is a failed install (the Core's own definition, in `harness-install-service.ts`). A failure names the Harness, quotes the Core's sentence, and links **#31** (opencode's install path) and **#128** (the canary). Covered end-to-end over a real link in `project-harness-live.test.ts`, both directions, plus the "Core cannot install anything" refusal.

  One qualification, stated rather than glossed: **"progress" is elapsed-time ticks plus a live re-read of the availability map, not the vendor installer's own output.** The Core logs installer lines locally (`log.info("core-harness-install.progress")`) and puts none of them on the core link, so there is no installer output for a client to relay. Streaming it would be a protocol change, which is outside this ticket. The re-read is what makes the verdict real on a Core whose event log is not wired.

### Traps

- [x] **`sinceSeq` is not computed in the CLI.** Getting it wrong shows up as a replay storm rather than an error, so the number stays with the client that owns it. The two places the CLI *could* still get it wrong are both covered: a first run does not print the tail it uses to find the tip (else the storm would be produced deliberately, on the first command an operator types), and `harness install` discards the replay up to `eventsReplayed` before sending its request — otherwise an install that failed an hour ago would be read as this one's verdict. That last one has its own test: an `installFailed` event at `#5` after a marker at `#7` is ignored.

- [x] **`project add` offers no edit that cannot be sent (ADR 0022).** No op carries a path after `create`, so the noun has no path-editing verb. `edit`, `mv`, `move`, `set-path`, `setpath`, `path` and `update` are caught and answered with the fact, the remedy, and #104 — at `EXIT_USAGE` rather than `EXIT_UNIMPLEMENTED`, because this is not a verb waiting on a train but one the protocol cannot carry. The test asserts no such verb reaches the Core.

### Not claimed

- [ ] **A real vendor installer was never run.** `harness install` is exercised against a real Core over a real link, with an install port that answers both ways — not against a machine that actually installs opencode. That path is exactly what #31 and #128 say is broken; this ticket's job was to report it honestly, not to fix it.
- [ ] **`project cp` / `project files` are reserved, not built.** Phase 3, #168. They exit `EXIT_UNIMPLEMENTED` and name the ticket, per phase-3 guardrail 3 (*reserve them in the command tree before `--help` is written*).
- [ ] **`session` and `core shell` are untouched.** #160 and #162 own them; `session` stays in `RESERVED_NOUNS`.

### Boundaries honoured

Nothing under `packages/sdk`, `release.yml` or `docs/reference` is touched. The shared router edit is three `case` arms, three help lines, and the three nouns leaving `RESERVED_NOUNS`.

packages/cli/src/harness-command.ts
```diff
@@ -0,0 +1,460 @@
...
+ */
+function establishEventTip(client: CoreLinkClient): Promise<number> {
+  return new Promise<number>((resolve, reject) => {
+    const timer = setTimeout(() => {
+      stop();
+      reject(new Error("the Core did not answer the event subscription"));
+    }, 30_000);
+    const off = client.onEventsReplayed(({ lastEventId }) => {
+      stop();
+      resolve(lastEventId);
```
qcentic-adm: **Blocking.** This resolves the tip as "whatever `eventsReplayed` said", but that marker is not the log's tip when the replay was capped.

`handleSubscribe` does `readEventTail(fromEventId, EVENT_TAIL_LIMIT)` with `EVENT_TAIL_LIMIT = 1_000` (`packages/core/src/pty-core-link-server.ts:377`, `:1543`) and then sends `eventsReplayed { lastEventId: lastSent }` — the id of the last event it *actually sent*. Subscribing from `0` against a Core with more than 1000 events resolves `tip` at #1000, and `pushLiveEvents` then delivers #1001..N as ordinary events with no further marker. `event-log-store.ts` prunes nothing, so this is the normal case for a Core that has been up a while, not an edge.

Suggested in-scope fix: re-send `subscribe` from the new cursor and keep waiting until a marker arrives whose replay tail was shorter than the cap. Do **not** fix it by having the Core report `getLastEventId()` in the marker — that would advance a client's cursor past events it was never sent, trading a replay storm for silent loss.

packages/cli/src/harness-command.ts
```diff
@@ -0,0 +1,460 @@
...
+      clearInterval(ticker);
+      clearTimeout(deadline);
+      offEvent();
+      resolve(outcome);
+    };
+
+    const offEvent = client.onEvent(({ event }) => {
+      // Past the tip only: everything at or below it happened before this
+      // command asked for anything.
+      if (event.eventId <= tip) return;
```
qcentic-adm, line 383: Consequence of the truncated tip above: with `tip` pinned at #1000 on a longer log, a `harness:installFailed` for this harness from an hour ago sitting at #2000 passes this guard and is reported as *this* install's verdict — the precise failure the function's own docblock says it prevents.

The mirror case is worse: a stale `agents:availabilityChanged` whose map had this harness `available` resolves `{ok: true}` and exits 0 on an install that never happened. (The `before` early-return only covers a harness that is available *now*, not one that was.)

`"ignores an outcome from before it asked"` in `harness-command.test.ts` cannot catch this — `fakeCore` never truncates a replay. Worth a case with a >1000-event `arrayEventLog` against the in-process Core.

packages/cli/src/events-command.ts
```diff
@@ -0,0 +1,267 @@
...
+      if (limit.value !== null && printed >= limit.value) finish(EXIT_OK);
+    });
+
+    // The marker that closes a replay tail — and, on a first run, the moment the
+    // Core's tip becomes known. From here the stream is live. Every later
+    // connection fires this too; printing is only ever switched on.
+    const offReplayed = client.onEventsReplayed(({ lastEventId }) => {
+      if (!printing) {
+        deps.verbose(`the Core's log ends at #${lastEventId}; following from there`);
+        printing = true;
```
qcentic-adm, line 214: **Blocking.** Same root cause as `harness-command.ts:264`. On a first run this switches printing on at the *first* marker, and when the Core capped the replay at `EVENT_TAIL_LIMIT` (1000) that marker is #1000, not the tip. Everything from #1001 to the real tip then arrives through `pushLiveEvents` as ordinary events and gets printed.

So on any Core with a log longer than 1000 events — and nothing prunes it — the very first `actana events tail` prints the history it exists to suppress. That is the replay storm the ticket names as the trap, and the file header currently asserts the opposite ("A first run that printed it would be the replay storm, produced deliberately").

`events-tail-cursor.test.ts` proves the mechanism correctly for a 6-event log; it needs one case where the tail past the cursor exceeds the cap.

packages/cli/src/events-command.ts
```diff
@@ -0,0 +1,267 @@
...
+    durable: true,
+    storage,
+  });
+  if (!opened.ok) return opened.code;
+  const { client, name, endpoint } = opened.core;
+
+  // A first run has nothing on disk to resume from, and "resume from 0" is not
+  // what a tail means. `--since` is an explicit answer to the same question and
+  // wins over both.
+  const fromStart = since.value !== null || fileCursor.hadCursor;
```
qcentic-adm: Nit, non-blocking: `fileCursor.hadCursor` is only true because `DurableCoreClient`'s constructor happened to call `getItem` during the awaited `openCore` above. That ordering is load-bearing and invisible from here — if `connect` ever built the client lazily, `fromStart` silently becomes `false` and a resuming tail starts behaving like a first run, with no test failing (the fake-based suites in `events-command.test.ts` never touch the storage, so only the live cursor suite covers the `true` branch).

Reading the file once here and passing the value in, rather than observing a side effect of construction, would make it say what it means.

While here: `FileCursorStorage.setItem` swallows write failures entirely. The trade-off is argued well in the class comment, but a `deps.verbose` on the swallow would turn "this tail duplicates every run" from a mystery into a line.

qcentic-adm (changes requested):
Reviewed against #161 criterion by criterion, reading the enforcing lines rather than the description. This is strong work: the traps were understood rather than restated, `events-tail-cursor.test.ts` is a real proof against a Core that really drops, and the `--json`/stderr split is asserted rather than asserted-about. One defect blocks, and it lands on exactly the trap the ticket named.

## Blocking — the tip is learned from a marker that is not always the tip

`PtyCoreLinkServer.handleSubscribe` replays with `readEventTail(fromEventId, EVENT_TAIL_LIMIT)`, `EVENT_TAIL_LIMIT = 1_000` (`packages/core/src/pty-core-link-server.ts:377` and `:1543`), then sends `eventsReplayed { lastEventId: lastSent }` — `lastSent` being the id of the **last event it actually sent**, not the log's tip. When the tail past the cursor is longer than 1000 the marker is a *partial* marker, and the remaining history then arrives through `pushLiveEvents` as ordinary `event` frames with no second marker behind them. `packages/core/src/event-log-store.ts` prunes nothing, so any Core that has been up a while is past 1000.

Both consumers of "subscribe from 0, read the marker, call it the tip" break there:

- **`events tail` first run** switches `printing` on at #1000 and then prints events 1001..N of the Core's history — the replay storm, produced deliberately on the first command an operator types. This is the failure mode the ticket says "shows up as a replay storm, not an error".
- **`harness install`** pins `tip` at #1000, so a `harness:installFailed` for the same harness from an hour ago at #2000 is past the tip and is read as *this* install's verdict — the exact thing the PR body says it prevents. The mirror case (a stale `agents:availabilityChanged` carrying a map where the harness was available) exits 0 on an install that never happened. `"ignores an outcome from before it asked"` passes only because `fakeCore` never truncates a replay.

The fix is **not** to have the Core put `getLastEventId()` in the marker: that advances the client's cursor past events it has not been sent, converting a replay storm into silent loss. In scope for this PR: re-`subscribe` from the new cursor and keep suppressing until a marker arrives whose replay tail was shorter than the cap. Out of scope but cleaner: carry the true tip on `subscribeAck` as a field *beside* `fromEventId` that the cursor never consumes. If neither is done here, the cap needs saying out loud in both files rather than being assumed away — the current comments state the opposite.

Related cost, not blocking: `establishEventTip` subscribes from `0` on every `harness install`, so each run drags the Core's whole event log over the wire to learn one number.

**What**

| Criterion | Verdict |
|---|---|
| `actana project ls add browse` | **Met** — `project-command.ts`; `ls`/`browse`/`add` all take `--json`. |
| `actana harness ls install` | **Met** — `harness-command.ts`; `ls --json` sorted by id, `install --json` one verdict object. |
| `actana events tail` | **Met** as a surface — `--since`/`--kind` (repeatable)/`--limit`, NDJSON. |

**Done when**

1. **Every list command emits `--json`** — **Met.** `project ls`, `project browse`, `harness ls`, `harness install`, `events tail` all do, and purity holds on the failure paths too: `openCore` reports on stderr only and returns an exit code (`core-connection.ts`, `openCore`), never an error object on the stream a consumer parses. Asserted, not just claimed — `project-command.test.ts` parses stdout after a `--verbose` run and asserts `run.out` is `[]` on an unreachable Core; `harness-command.test.ts` `"keeps progress off stdout"` asserts the `Installing …` lines are on stderr while stdout parses as one object; `events-command.test.ts` `"keeps a dropped link off stdout"` does the same for the reconnect notice.

2. **`project browse` walks the Core's disk through `dirList`** — **Met.** `projectBrowse` sends `{type: "dirList", path: requested ?? null}` and renders the answer verbatim; nothing in the file touches `node:fs` or `process.cwd()`. `project add` likewise sends the path as typed, with the single `startsWith("/")` check done as a string rather than `path.isAbsolute` (correct — the target may be a Linux VM). `project-harness-live.test.ts` answers for `/srv/core-disk` with folders that exist nowhere, so a browse of the operator's own disk cannot pass; `"refuses a relative path"` asserts `core.mutations` stayed empty.

3. **`events tail` follows from a cursor and survives a reconnect** — **Partially met.** The steady-state arithmetic is right and I checked the boundary on both sides: the Core's `readEventTail` is strictly `eventId > afterEventId` (exclusive), the client sends its highest-seen id, and `DurableCoreClient.deliverEvent` drops `eventId <= lastEventId` — so an overlapping replay cannot double and an exclusive tail cannot gap. The reconnect race is safe too: `handleSubscribe` is synchronous end to end and sets `conn.lastSentEventId` before the marker goes out, so an event appended during the replay window is either in the tail or picked up by the next `pushLiveEvents` — never both, never neither. `pinnedCursorStorage` is read once in the `DurableCoreClient` constructor, so `--since` does not re-pin on every reconnect. The cursor test is the real thing. What is not met is the first-run boundary above.

4. **`harness install` reports progress and a real exit status; #31/#128 not papered over** — **Partially met.** Exit 0 requires the Core reporting the harness *available*; an ack alone never prints a success line; failures name the harness, quote the Core's sentence and link #31 then #128, with the opencode entry isolated to one map row so closing #31 is a one-line delete. The second route (re-reading the availability map on the tick) is a genuinely good call for a Core with no event log wired. The author's own qualification — that "progress" is elapsed-time ticks plus that re-read, not the vendor installer's output — is accurate and correctly attributed: the Core logs installer lines locally and puts none of them on the link, so relaying them would be a protocol change. Held back only by the stale-verdict window above.

**Traps**

5. **`sinceSeq` is not computed in the CLI** — **Partially met.** True for the running stream. Not true for the two tip-learning sites, which is where the off-by-one actually lives.

6. **`project add` offers no edit that cannot be sent (ADR 0022)** — **Met.** `projectsMutate` is called in exactly one place, with `op: "create"`. `edit`/`mv`/`move`/`set-path`/`setpath`/`path`/`update` are caught and answered with the fact, the remedy and #104, at `EXIT_USAGE` — and the reasoning for `EXIT_USAGE` over `EXIT_UNIMPLEMENTED` is right: `EXIT_UNIMPLEMENTED` means "a later train has this", and no later train makes `project set-path` appear.

7. **Shared-file footprint** — **Met, and it is the minimum.** `actana-cli.ts`: three `case` arms, three help lines, three nouns out of `RESERVED_NOUNS`. `cli-args.ts`: three flags. `cli-deps.ts`: `connect`. `exit-codes.ts`: comment only. `actana-cli-entry.ts`: one wiring line. Nothing under `packages/sdk`, `release.yml` or `docs/reference`; the whole diff is `packages/cli`. `session` stays reserved.

## The three unchecked boxes

- **No real vendor installer run** — reason holds. #31 and #128 are the open statement that the path is broken; the ticket asked for honest reporting, not a fix, and the live suite exercises both directions over a real link plus the "Core cannot install anything" refusal. Nothing more was available to prove here.
- **`project cp` / `project files` reserved for #168** — reason holds, and reserving is the *stronger* move than omitting: `RESERVED_VERBS` exits `EXIT_UNIMPLEMENTED` naming the ticket, which stops the shorter root-level `actana cp` spelling looking available. Consistent with the `core shell` precedent.
- **`session` and `core shell` untouched** — reason holds and is verifiable: `session` is still the sole entry in `RESERVED_NOUNS` and no session/shell file is in the diff.

## The duplicated in-process Core rig

Not blocking, per the agreed plan — #204 merges first, this PR's fix round deletes `in-process-core.ts` and repoints at #204's `in-process-core-harness.ts`. But **the merged rig must be #204's module carrying #205's behaviour**, not #204's file as it stands. What is strictly better here and has to survive:

1. **A teardown that actually frees the port.** #204 has `close: (cb) => wss.close(() => tlsServer.close(cb))`. `wss.close()` stops the upgrade handler but leaves established TLS sockets alive, and an HTTPS server with a live socket on it never finishes closing — so with a client attached that callback never fires and the port stays bound. #205 terminates every live client first, then closes the `WebSocketServer`, then `tlsServer.close(cb)` plus `closeAllConnections()`. Without this a restart on the same port is `EADDRINUSE`, not a reconnect.
2. **`stop(): Promise<void>` and `waitForPortFree`** — close *and wait until the port is genuinely bindable*, by testing the condition rather than sleeping.
3. **`opts.port` and `opts.material` reuse.** #204 always allocates a fresh port and fresh cert material, so "the same Core, restarted" is not expressible. These three together are the whole reason the reconnect criterion is testable at all.
4. **The wider `InProcessCore` surface** — `port`, `material`, `close`, `stop`. #204 returns only `{server, blobText, endpoint}`.
5. **The extra ports and the poll knob** — `eventLog`, `directoryPort`, `availabilityPort`, `installPort`, `liveEventPollMs`.
6. **`arrayEventLog()` with the `tailReads` counter**, and the `waitFor(predicate, what)` helper that fails with a sentence instead of a timeout — #204 inlines a bespoke bound-wait loop that should go.

Going the other way, **keep #204's `opts.ptyCore` override**: #205 hardcodes `unusedPtyCore()` and the session suites need to substitute. And pick one secret name — #205's `CORE_SECRET` reads better beside `CORE_ID` than #204's bare `SECRET`.

## Verdict

Requesting changes on the truncated-marker defect only. Fix that in the same round as the rig dedupe and this is ready — everything else on #161 is met, and the parts that are met are met at the enforcing line rather than in the description.

packages/cli/src/harness-command.ts
qcentic-dev: Fixed in `cd68c02`.

You were right about the mechanism and right that the mirror case is the worse one. The fix is the shape you suggested — re-`subscribe` from the new cursor, keep waiting — with one deliberate departure: **the loop ends on an *empty* tail rather than on a tail shorter than the cap.**

`event-tip.ts` is the whole argument, shared by both call sites. The reason for not mirroring `EVENT_TAIL_LIMIT` here is that the two drift directions are not symmetrical. A copy that is *lower* than the Core's costs an extra round trip. A copy that is *higher* reads a full tail as a short one and restores this exact defect — silently, on a Core nobody thought had changed, in a package that cannot import the constant (ADR 0025 D4 keeps `@actana/core` out of the manifest). An empty tail is the only answer a Core gives that cannot have been cut short, and it means that whatever the cap is set to. The cost is one extra round trip per hunt.

Termination is bounded rather than assumed: 200 rounds, and reaching it (a Core appending faster than this side can drain) prints a sentence on stderr saying the number may be behind the log's end, instead of hanging or quietly rounding off.

Your two "do nots" are both honoured and written down in the file — `getLastEventId()` in the marker would advance a cursor past events never sent, and the true tip on `subscribeAck` beside `fromEventId` is the cleaner fix and a protocol change, so not here.

On the related cost you flagged: `establishEventTip` still subscribes from `0` on every run, and the docblock now says so, why it is inherent (a cursor guessed high stops live push silently; a remembered one is not this command's tip), and what removes it — the `subscribeAck` field above.

packages/cli/src/harness-command.ts
qcentic-dev: Fixed in `cd68c02`, and both directions now have the test you asked for — a `>1000`-event `arrayEventLog` against the in-process Core, in `project-harness-live.test.ts`:

- **`ignores a stale \`available\` sitting past the cap on a long event log`** — the mirror case you called worse. A 1500-event log with an `agents:availabilityChanged` at #1200 whose map had opencode available, a Core whose current map says `missing`, and an install port that genuinely fails. Asserts `installed: false` and a non-zero exit.
- **`ignores an install that failed before it asked, past the cap on a long log`** — `harness:installFailed` for opencode at #1200 on the same 1500-event log, and an install that actually succeeds. Asserts `installed: true`.

Both fail against the code before `cd68c02` — I checked by reverting the tip logic to `const end = lastEventId` and running them, rather than by reading the diff. The `events tail` equivalent (`prints none of a history longer than the Core replays in one tail`, in `events-tail-cursor.test.ts`) fails there too.

You were also right that `"ignores an outcome from before it asked"` could never have caught this: `fakeCore` is hand-driven and never truncates. It has grown a sibling at that level anyway — `events-command.test.ts` now asserts that a marker closing a non-empty tail does not start the stream and that the command re-asks from the cursor it reached — but the proof is the live suite.

packages/cli/src/events-command.ts
qcentic-dev: Fixed in `cd68c02`, same mechanism as the `harness install` side — `event-tip.ts` is shared by both, so there is one place that argues about the cap and one loop that closes it.

Here it means: while `printing` is off, every event is fed to the tracker and none is printed, and a marker only switches printing on when it closes an empty tail. The file header no longer asserts the opposite; it now says a first run walks to the end of the log across as many markers as that takes, and points at `event-tip.ts` for why one is not enough.

The case you asked for is in `events-tail-cursor.test.ts` — `prints none of a history longer than the Core replays in one tail`: a 1500-event log, the first `events tail` on that machine, asserting the run decided the log ends at **#1500 and not #1000**, that nothing was printed while it was finding out, and that the one line it does print is the event appended afterwards. It fails against the code before this commit.

(It waits on the `--verbose` notice to know when the hunt converged, so the event it appends is live rather than history — which needed an `onErr` hook on the CLI fixture, symmetric with the existing `onOut` and documented as being for exactly this.)

packages/cli/src/events-command.ts
qcentic-dev: Both taken, in `cd68c02`.

**The ordering.** You are right that it was load-bearing and invisible, and right about the failure mode — a lazily-built client turns a resuming tail into a first run with nothing failing. `hadCursor` is gone. `eventsTail` now resolves the Core itself and reads the cursor file *before anything dials*, through `storedCursorFor(dir, endpoint)`; `openCore` takes the resolution it already made, so the Core a command means is still decided in one place and only once.

Before the dial rather than merely explicitly: a durable client starts advancing that same cursor the moment its link comes up, so a read taken after `openCore` returns could not tell "there was a cursor here" from "this run just wrote one". The key comes from the SDK's own `coreLinkCursorStorageKey`, given the endpoint the blob carries — which is what a client derives it from (`coreConnectionFromBlob` trims the endpoint and hands it over as the URL), so this is not a second opinion about which file a Core's cursor lives in.

**The swallowed write.** `FileCursorStorage` now takes an `onWriteFailed` sink and `events tail` passes `deps.verbose`. Still swallowed — a read-only config directory is not a reason to refuse to stream a Core's events — but the line names the consequence rather than the exception alone: *"could not write the event cursor to …; the next run of this command will start where this one did"*. That is the sentence somebody needs when the symptom they are looking at is a tail that repeats itself.

Both are covered by a new `event-cursor-file.test.ts`, including the write failure (a file standing where the directory should be, so it does not depend on the test process's uid).

qcentic-dev:
## Fix round for the CHANGES_REQUESTED review

Three commits, no force push, still a draft.

| sha | what |
|---|---|
| [`966f9fb`](https://github.com/actana/control/commit/966f9fb86d0f564b71827f74f988d6dce9f811b4) | merge of `origin/feat/cli-and-publishing` (#204) into this branch |
| [`24d07af`](https://github.com/actana/control/commit/24d07afb0d817d4e988340b70f2179949edc221d) | the two in-process Core rigs become one |
| [`cd68c02`](https://github.com/actana/control/commit/cd68c020c578228c3d37b4c938788210e685a262) | **the blocking defect**, its tests, and the two non-blocking notes |

Inline replies are on each of your four comments. The short version of the blocker: a replay marker is a receipt for what the Core sent, never a statement about where its log ends, so the tip hunt now re-`subscribe`s from what it has and stays quiet until a marker closes an **empty** tail — the one answer that cannot have been cut short, whatever `EVENT_TAIL_LIMIT` is set to. I did not mirror the constant here, deliberately: a copy that drifted *above* the Core's own value would read a full tail as a short one and restore this defect silently. Both of your "do nots" are honoured and written down in `event-tip.ts`.

All three regression tests were checked by reverting the fix and watching them fail, not by reading the diff.

## The rig dedupe

Merged as you ruled: **#205's module carrying #204's `ptyCore`**. Everything on your list survives — the teardown that terminates live sockets before closing the HTTPS server (so a restart on the same port is a reconnect and not `EADDRINUSE`), `stop()` + `waitForPortFree` testing the condition rather than sleeping, `opts.port` / `opts.material` reuse, the wider `InProcessCore` surface, the extra ports and `liveEventPollMs`, and `arrayEventLog()` with `tailReads` plus the `waitFor` that fails with a sentence. #204's `ptyCore` override came the other way and is what `in-process-core-session.test.ts` uses; the default is still the manager that throws on every method. The secret name is `CORE_SECRET`, as you preferred.

One thing I did against the letter of the plan, and want on the record rather than buried: **the surviving file is `in-process-core.ts`, not `in-process-core-harness.ts`.** A Harness in this CLI is a coding agent — it is a whole noun, `actana harness ls` — and a module named for one that holds none is a name that needs explaining every time it is read. Your ruling was about which behaviour survives, and that is what I followed. Say the word and the rename is one commit.

## One thing the merge needed, not the review

#204's `no-prompt-timing.test.ts` sweeps every shipped module for a scheduler, and `harness install` carries a 15-minute deadline plus a progress tick, so the merge was red until one of them moved. I did not delete the rule. That wait is the trade `--wait-timeout` already makes — it re-sends nothing, cannot make a Harness look installed sooner, and reports its expiry as *this side* giving up rather than as an outcome — so the sweep now takes a one-row table of allowed files, each with a reason, and gained two rules so the allowance cannot widen quietly: a listed file must still exist, and no module on the path from an operator's text to a harness's stdin may schedule at all, listed or not. It is #160's invariant and I would rather you looked at the amendment than found it.

qcentic-dev:
## Acceptance criteria — #161, after the fix round

Updated against `cd68c02`. The three rows the review moved off "met" are the ones with the shas beside them.

### What (the surface D10 names)

- [x] **`actana project ls add browse`** — `project-command.ts`. `ls` and `browse` both take `--json`; `add` takes it too, so a script that creates a Project gets the row back.
- [x] **`actana harness ls install`** — `harness-command.ts`. `ls --json` is an array sorted by id; `install --json` emits one verdict object.
- [x] **`actana events tail`** — `events-command.ts`, with `--since`, `--kind` (repeatable) and `--limit`.

### Done when

- [x] **Every list command emits `--json`.** Unchanged by this round, and confirmed at the enforcing line by the review: purity holds on the failure paths too, because `openCore` reports on stderr and returns an exit code rather than putting an error object on the stream a consumer parses.

- [x] **`project browse` walks the Core's disk through `dirList`.** Unchanged. `/srv/core-disk` and the folders under it exist on no filesystem anywhere, so a browse of the operator's own disk cannot pass `project-harness-live.test.ts`.

- [x] **`events tail` follows from a cursor and survives a reconnect without duplicating or dropping** — **`cd68c02`.** The steady-state arithmetic was already right; what was wrong was the first-run boundary. A `eventsReplayed` marker is a receipt for what the Core sent, not its log's end: `handleSubscribe` caps a replay at `EVENT_TAIL_LIMIT` (1000) and reports the last id it managed, so on any Core with a longer log — and nothing prunes one — the first `events tail` printed the history above the cap. That is the replay storm the ticket names, produced on the first command an operator types.

  Now: the tip hunt (`event-tip.ts`, shared with `harness install`) re-`subscribe`s from the cursor it reached and keeps printing off until a marker closes an **empty** tail — the only answer a Core gives that cannot have been cut short. Not by mirroring the cap here, because a copy that drifted above the Core's would read a full tail as a short one and restore the defect silently; not by having the Core report `getLastEventId()`, which advances a cursor past events never sent. Bounded at 200 rounds, and reaching it says so on stderr rather than rounding off in silence.

  Proof: `events-tail-cursor.test.ts` keeps the six-events-each-exactly-once case across a real Core restart, and gains **`prints none of a history longer than the Core replays in one tail`** — a 1500-event log where the run must find #1500 and not #1000, print none of the history, and print exactly the one event appended afterwards.

- [x] **`harness install` reports progress and a real exit status; #31 and #128 are not papered over** — **`cd68c02`.** Same root cause, worse consequence: a tip pinned at the cap let a `harness:installFailed` from an hour ago be read as this install's verdict, and let a stale `agents:availabilityChanged` exit 0 on an install that never happened. Both directions now have a case against the in-process Core with a 1500-event log and the stale event at #1200. The qualification stands unchanged: "progress" is elapsed-time ticks plus a live re-read of the availability map, not the vendor installer's output, because the Core puts none of it on the link.

### Traps

- [x] **`sinceSeq` is not computed in the CLI.** True for the running stream, and now true at the two tip-learning sites — which is where the review found it was not. Neither computes a cursor: they read the Core's own markers, and the fix was to stop believing the first one.
- [x] **`project add` offers no edit that cannot be sent (ADR 0022).** Unchanged. `projectsMutate` is called in exactly one place, with `op: "create"`.

### Review notes, non-blocking

- [x] **First-run detection no longer rides on a constructor side effect** (`cd68c02`). The cursor file is read before anything dials — after the dial, a durable client's own first replay would have answered the question — through the SDK's own key helper.
- [x] **An unwritable cursor is swallowed *and said*** (`cd68c02`), naming the consequence: the next run starts where this one did. Covered by a new `event-cursor-file.test.ts`.
- [x] **`establishEventTip` subscribing from `0` every run**: documented rather than fixed, with why it is inherent (a cursor guessed high stops live push silently; a remembered one is not this command's tip) and what removes it — the true tip on `subscribeAck` beside `fromEventId`, which is a protocol change.
- [x] **The duplicated rig is gone** (`24d07af`) — one module, #205's behaviour, #204's `ptyCore`, `CORE_SECRET`. Filename kept as `in-process-core.ts`; the reasoning is in the comment above this one.

### Not claimed

- [ ] **A real vendor installer was never run.** `harness install` is exercised against a real Core over a real link, not against a machine that actually installs opencode. That path is what #31 and #128 say is broken; this ticket's job was to report it honestly.
- [ ] **`project cp` / `project files` are reserved, not built.** Phase 3, #168 — they exit `EXIT_UNIMPLEMENTED` and name the ticket.

### Changed by the merge, worth a second pair of eyes

- **`RESERVED_NOUNS` is empty.** #160 took `session` out of it and #161 took the other three, so the "Reserved, landing later in this phase" heading left the root help with them. The table stays for the next noun; the reservations this build carries are verbs (`project cp` / `files` #168, `core shell` #162), and that is where `actana-cli.test.ts` now asserts the reserved-vs-typo distinction.
- **#160's "the CLI schedules nothing" sweep now has a one-row allowance** for `harness install`'s deadline and progress tick, with two new rules keeping it narrow. Argued in the merge commit and in the comment above.

### Boundaries

Nothing under `packages/sdk`, `release.yml` or `docs/reference` is touched by this round. Everything in the three commits is `packages/cli`, apart from the merge bringing #204's own files across.

qcentic-adm (approved):
Re-reviewed at `cd68c02` against my CHANGES_REQUESTED, reading the enforcing lines and the tests rather than the replies. **The blocking defect is fixed on both paths, the fix is the right shape, and the regression tests are real.** Approving.

## The blocking defect

`event-tip.ts` is the fix and it is one place rather than two. `trackEventTip` counts what each replay delivered and treats a marker that closed a non-empty tail as a receipt — re-`subscribe`s from the highest id seen and keeps waiting. Only `delivered === 0` returns a tip.

**Ending on an empty tail rather than on a short one is the better call, and the reasoning holds at the wire.** `handleSubscribe` (`packages/core/src/pty-core-link-server.ts:1534`) initialises `lastSent = fromEventId` and sends `eventsReplayed { lastEventId: lastSent }` even when the tail is empty — so an empty tail is a marker the client can recognise without knowing `EVENT_TAIL_LIMIT` at all. That closes the drift the mirrored-constant version would have carried, and the asymmetry argument in the file is the correct one: a copy below the Core's costs a round trip, a copy above it silently restores this bug. Both of my "do nots" are honoured — `getLastEventId()` in the marker is not used, and the `subscribeAck` field is correctly left as the protocol change it is.

Both consumers now gate on it:

- `events-command.ts:206` — `printing` flips only when `tipFrom` returns non-null; while it is off every event goes to `tip.saw` and none to `deps.out`. The header comment that used to assert the opposite is rewritten.
- `harness-command.ts:307` — `establishEventTip` resolves only on the same condition, with the deadline re-armed per marker rather than run against the whole hunt. That distinction is right: a long log is a Core answering slowly, not a Core gone quiet.

`MAX_TIP_ROUNDS = 200` with a stderr sentence rather than a hang or a silent round-off is the correct way to bound a loop that depends on the Core going quiet for one round.

**Spot-check of "all three fail on the pre-fix code":** I checked `prints none of a history longer than the Core replays in one tail` (`events-tail-cursor.test.ts:169`) against the pre-fix file. Pre-fix, `onEventsReplayed` printed `the Core's log ends at #${lastEventId}` and set `printing = true` on the first marker — which on a 1500-event log is #1000. So `expect(end).toContain("#1500")` fails, and `expect(printed).toEqual([])` fails too once #1001 arrives through `pushLiveEvents` under `--limit 1`. The test discriminates, and `expect(log.tailReads).toBeGreaterThan(1)` is a good guard that the suite is exercising the capped path rather than a log that happened to be short. The two `project-harness-live.test.ts` cases discriminate on the same mechanism in both directions — pre-fix, the stale `agents:availabilityChanged` at #1201 sits above a tip pinned at #1000 and resolves `{ok: true}` on an install that fails, and the stale `installFailed` does the mirror. The claim is sound.

`events-command.test.ts` gaining `keeps printing off while a truncated replay is still being drained` at the fake level, and the pre-existing case being amended to require a second empty-tail marker, is the right pair — the fake proves the loop, the live suite proves the Core really truncates.

## The rig dedupe

Verified as one module, not a shim. `in-process-core-harness.ts` is deleted, `git grep` finds no reference to it anywhere in the tree, and the four importers (`in-process-core.test.ts`, `in-process-core-session.test.ts`, `project-harness-live.test.ts`, `events-tail-cursor.test.ts`) all point at `in-process-core.ts`.

All six items from my list survive, at the line:

1. Teardown terminates every live client, then closes the `WebSocketServer`, then `tlsServer.close(cb)` + `closeAllConnections()` (`in-process-core.ts:140`).
2. `stop(): Promise<void>` + `waitForPortFree`, testing `portFree` rather than sleeping.
3. `opts.port` / `opts.material` reuse.
4. The wider `InProcessCore` — `server`, `port`, `endpoint`, `blobText`, `material`, `close`, `stop`.
5. `eventLog`, `directoryPort`, `availabilityPort`, `installPort`, `liveEventPollMs`.
6. `arrayEventLog()` with `tailReads`, and `waitFor(predicate, what)`.

`opts.ptyCore` came the other way and defaults to `unusedPtyCore()`, so a suite that reaches a PTY manager has to say so. `CORE_SECRET` is the surviving name and is the only one left in the package.

**#204's session tests genuinely run against the unified rig.** `in-process-core-session.test.ts:169` calls the merged `startInProcessCore({ ptyCore: pty.core, queryPort, mutationPort })` and holds an `InProcessCore` rather than a bare `PtyCoreLinkServer`; the `@actana/core` import is gone from the file. I diffed the deleted module's `startInProcessCore` against the survivor's — same `PtyCoreLinkServer` construction, same TLS material, same `authVerifier`, same blob shape and bearer — so the survivor is a strict superset and the session suite gets the same Core with a teardown that actually frees the port. No adapter, no re-export, no compatibility shim.

On the filename: `in-process-core.ts` over `-harness` is the right call and the argument for it is correct — a Harness in this CLI is a coding agent, and my ruling was about which behaviour survives. No rename wanted.

## The non-blocking notes

- **`fileCursor.hadCursor`** — fixed properly rather than patched. `eventsTail` now resolves the Core itself and calls `storedCursorFor(dir, resolved.core.blob.endpoint)` before anything dials, and `openCore` takes the resolution it was handed (`OpenCoreOptions.resolved`) so the Core is still decided in one place. *Before* the dial rather than merely explicitly is the part that matters, and the reason given is the right one — a durable client starts advancing that same file the moment its link comes up. I checked the key agrees rather than taking it on trust: `storedCursorFor` uses `coreLinkCursorStorageKey(endpoint.trim())`, `DurableCoreClient` uses `coreLinkCursorStorageKey(this.url)` (`durable-core-client.ts:105`) with `url = blob.endpoint.trim()` from `coreConnectionFromBlob`, and `connectCore` passes no `cursorStorageKey` override. Same file, both sides. `treats a garbled cursor as no cursor rather than as event 0` is a case I would not have thought to ask for and is exactly right — "resume from 0" and "nothing to resume from" are different commands.
- **The swallowed cursor write** — `FileCursorStorage` takes `onWriteFailed` and `events tail` passes `deps.verbose`. Still swallowed, for the reason argued, and the sentence names the consequence rather than only the errno. The test provokes it with a file standing where the directory should be, which is uid-independent — better than the read-only-mode version I had in mind.
- **`establishEventTip` subscribing from `0` every run** — answered rather than fixed, and the answer is good. The docblock states the cost, why it is inherent while the tip is only knowable by walking to it, and that `subscribeAck` beside `fromEventId` is what removes it. Correctly out of scope for a ticket that does not move the wire format.

## Two observations, neither a change request

**The timing-sweep amendment.** You were right to surface it rather than bury it. The allowance is narrow and the two new rules are the right two — a listed file must still exist, and `TOUCHES_A_SESSION` bars a timer on the prompt path regardless of the table. I checked `harness-command.ts` does not match that regex, so the guard is not vacuous. The one soft edge is that `TOUCHES_A_SESSION` matches on filename, so a future module that writes to a Session without `session`/`prompt`/`pty` in its name would need the allowlist row to be refused by a human rather than by the test. That is a reviewable act by construction, and #160's invariant is not weakened by this PR. Worth a line in #160 if the table ever takes a second row.

**A first run drops what arrives during the hunt.** While `printing` is off, live events appended during the tip hunt are counted into `seen` and never printed, so on a busy Core the first `events tail` can silently skip the handful appended between the first `subscribe` and the empty tail. This is the honest reading of "starts at the end of the log" — the end is where the walk finished — and the cursor is consistent with it, so nothing is duplicated later. Noting it because the window widens with the number of rounds, not because it should change here. On the `harness install` side the same behaviour is strictly correct: an event that landed before the request went out must not be read as its verdict, and a higher tip is what guarantees that.

## Verdict

Approved. The blocking defect is fixed at both enforcing lines with a shared argument rather than two copies of a workaround, the failure modes I named survive as tests that genuinely fail on the pre-fix code, the rigs are one module carrying the behaviour that was worth keeping from each, and the three non-blocking notes are fixed, fixed, and answered well. Everything on #161 is met.

PR #206, opened 2026-08-13
feat(cli): actana session attach — a terminal on a Session, holding its write lock

### LINKED ISSUE #163: 2.7 CLI session attach — claims the Session write lock (D10, needs #144)

qcentic-dev:
Closes #163. Phase 2, step 3 — `session attach` of [#129](https://github.com/actana/control/issues/129) D10.

Base is `feat/cli-and-publishing`, not `main`. Train `beta/0.2.2`.

## What landed

```
actana session attach <session>              take the terminal: read it live, type into it
actana session attach <session> --read-only  watch it without claiming the write lock
```

## The lock is the command, not a check on top of it

`attach` writes to a Session, so it claims the Session write lock (#144, ADR 0024 D3–D7) before a
byte reaches the operator's terminal — in `session-attach-channel.ts`, on the way in, from the first
commit. The claim's answer *is* what the attach then is:

| Claim | What you get |
| --- | --- |
| granted | you type; detaching hands the lock back |
| refused | a **read-only view with the reason on it** — not an error, not a takeover |
| `--read-only` | no `claim` frame at all |
| Core announces no `multiConnection` | **writable** — `supported: false` is not `granted: false` |

`forceTakeover` exists (D7) and this command does not send it. An attach that stole the lock from a
running automation because somebody typed a session id is the accident the lock was added to
prevent. The last row matters as much as the second: such a Core has no lock table and serves every
mutation this client makes, so rendering it read-only would tell an operator who is its only client
that somebody else is typing.

`--read-only` is there because a Session starts unlocked and its creator gets no privilege (D5). An
ordinary `attach` on an unclaimed Session takes the lock; an operator who only meant to watch an
automation would take it from the automation that was about to claim.

The attachment cannot write without the authority. It throws before the wire when it never had one,
and turns the Core's own `session-locked` refusal into the same error when a force takeover moves
the lock underneath a live attach — where the command says so on the terminal and **carries on
reading** rather than quitting. Losing your keyboard is not a reason to lose your view.

## Lock release on an abrupt disconnect

The trap the ticket names. A lock ends on explicit release, on the connection dropping, or on a
force takeover (D7) — so the two endings are deliberately asymmetric here:

- **A detach sends `release`.** So does a `SIGINT`/`SIGTERM`, which is a detach somebody else asked
  for.
- **A dropped link sends nothing.** The Core has already released it, and a teardown that asked
  anyway would be waiting on a socket that is gone, with the terminal still raw behind it. The
  operator is told the Session is claimable again, because otherwise they would reasonably assume
  the one they were driving is now stuck.

`session-attach-live.test.ts` proves the second against a real Core rather than asserting the
intention: a live attach holding the lock has its socket **terminated** — no close frame, no
`release`, nothing the client chose — and the next attach is granted the lock and types into the
PTY. Staging that needed one addition to the in-process rig (`InProcessCore.dropConnections`),
because no client can do it to itself: hanging up politely is a different event, and stopping the
whole Core proves nothing about a lock table that died with it.

## The terminal, and the keyboard

Raw-mode discipline is #203's, reused rather than rewritten: a `finally` around the whole session,
`SIGINT` and `SIGTERM` handled rather than fatal, a dropped link treated as an ending like any
other, a restore that is idempotent and never throws, and the process `exit` hook one layer down in
`cli-terminal.ts`. The terminal is handed back **before** the release round trip — a Core that has
stopped answering costs a request timeout, not a request timeout's worth of unusable shell.

`Ctrl-C` is forwarded to the harness, which is the whole reason a person attaches instead of reading
`session logs`. So **`Ctrl-]` detaches** (telnet's escape). A read-only attach has nothing to
interrupt on the far side, so `Ctrl-C` detaches there too rather than appearing broken, and the
keystrokes it dropped are counted and reported once at the end instead of repainting the Session's
screen with a warning per key.

A read-only attach also **does not resize the PTY**. The Core would accept it — `resize` is not
gated on the lock, and D4's comment there argues that deliberately — so this is the CLI's own
restraint: reflowing the terminal of the person actually typing because an observer widened a window
is interference in the one direction the lock cannot see.

A harness that exits is reported and exits `0`, unlike `core shell`: an attach is a terminal onto
somebody else's Session, and a script that could not tell "the attach failed" from "the harness you
were watching exited 1" would be worse off than one that reads the line.

## Boundaries

- `packages/cli` only. No file under `packages/sdk`, `release.yml` or `docs/reference` is touched —
  all terminal handling is CLI-side (D11) and the SDK's existing vocabulary (`claim`, `release`,
  `ptySubscribe`, `replay`, `write`, `resize`, `onData`, `onExit`) is used, not extended.
- `session attach` leaves `exit-codes.ts`'s reservation list, which is the shape a reservation is
  supposed to end in. `project cp` / `project files` are the two left, still #168's.
- **#129 D5's one-connection rationale is dead** by this train and nothing here leans on it. This
  command does not own the Core: it is one of many connections, holds one Session's lock, and
  subscribes to one PTY's stream.

## Checks

- `pnpm --filter @actana/cli test` — **291 passed, 3 skipped**, including both new suites.
- `pnpm typecheck` clean repo-wide; `pnpm lint` **0 errors**, 0 findings in `packages/cli`.
- `commitlint --from origin/feat/cli-and-publishing --to HEAD` — 0 problems across all three commits.

New suites:

- `session-attach.test.ts` (31 tests) — the ticket's "done when" as its outline. One test per exit
  path for the terminal; the four authorities; the release asymmetry; the takeover mid-session; the
  detach key against a chunk that carries both keystrokes and the key.
- `session-attach-live.test.ts` (4 tests) — the real `openSessionAttach` against the real
  `PtyCoreLinkServer` on a real `wss://` port, two attaches at a time: one writes and one reads,
  detach hands the lock over, and the abrupt drop above.

Not covered by an automated test, and the ticket says so up front: a real harness's TUI painted into
a real terminal. `attach` has no headless test story for the *feel* of it — everything provable
without a person is proved above, and the rest is the acceptance run.

### DIFF
--- a/packages/cli/src/__tests__/actana-cli.test.ts
+++ b/packages/cli/src/__tests__/actana-cli.test.ts
@@ -7,8 +7,10 @@
 // whole of `exit-codes.ts`'s argument, and until now nothing asserted it at the
 // root — the reserved nouns exited `2` and `core shell` exited `3` for the same
 // underlying fact, which the review of #201 caught. #162 built `core shell` and
-// #160/#161/#163 built the nouns, so what is reserved here is verbs inside a
-// built noun: `project cp`, `project files`, `session attach`.
+// #160/#161 built the nouns, so what is reserved here is verbs inside a built
+// noun: `project cp` and `project files`. `session attach` was the third until
+// #163 built it, and it is the counter-example below — the thing a reservation
+// is supposed to stop being.
 
 import { describe, it, expect, afterEach } from "vitest";
 import { EXIT_OK, EXIT_UNIMPLEMENTED, EXIT_USAGE } from "../exit-codes.ts";
@@ -30,14 +32,14 @@ afterEach(() => {
  *
  * No *noun* is reserved any more: `session` left the list when #160 built it
  * and `project`, `harness` and `events` left when #161 did, which is the shape
- * a reservation is supposed to end in. What is still reserved is a verb — and
+ * a reservation is supposed to end in. `session attach` left it the same way
+ * when #163 built it. What is still reserved is a verb waiting on phase 3 — and
  * the distinction the exit code draws is the same one either way, so these are
  * what assert it.
  */
 const RESERVED = [
   [["project", "cp"], "#168"],
   [["project", "files"], "#168"],
-  [["session", "attach"], "#163"],
 ] as const;
 
 /** The nouns that are built, so the help below cannot go quiet about them. */
@@ -56,22 +58,28 @@ describe("a reservation is `not built yet`, not `you typed it wrong`", () => {
 
   it("is the answer for a name that is reserved, and not for one that is built", async () => {
     // `core shell` was the reserved *verb* and shared this code until #162
-    // built it. Now it is a command like any other, and a build that answered
-    // `3` for it would be telling a script to come back on a later train for
-    // something already shipped.
-    const reserved = await cli().run(["session", "attach"]);
+    // built it; `session attach` was the last one and #163 built it. Now they
+    // are commands like any other, and a build that answered `3` for either
+    // would be telling a script to come back on a later train for something
+    // already shipped.
+    const reserved = await cli().run(["project", "cp"]);
     expect(reserved.code).toBe(EXIT_UNIMPLEMENTED);
-    const built = await cli().run(["core", "shell"]);
-    expect(built.code).not.toBe(EXIT_UNIMPLEMENTED);
+    for (const built of [["core", "shell"], ["session", "attach"]]) {
+      const run = await cli().run(built);
+      expect(run.code, built.join(" ")).not.toBe(EXIT_UNIMPLEMENTED);
+    }
   });
 
-  it("answers the same for a reserved verb wherever it sits in the tree", async () => {
-    // `project cp` is reserved in `project-command.ts` and `session attach` in
-    // `session-command.ts`, and a script cannot be asked to know which file it
-    // landed in. One fact about this build, one exit code.
-    const project = await cli().run(["project", "cp"]);
-    const session = await cli().run(["session", "attach"]);
-    expect(project.code).toBe(session.code);
+  it("answers the same for every reserved verb, and not for a built one beside it", async () => {
+    // One fact about this build, one exit code — and it has to stay a *fact
+    // about the build* rather than about a file. `project cp` and `project ls`
+    // sit in the same module and differ only in whether they are built, which
+    // is the difference the code is supposed to carry.
+    const cp = await cli().run(["project", "cp"]);
+    const files = await cli().run(["project", "files"]);
+    expect(cp.code).toBe(files.code);
+    const built = await cli().run(["project", "ls"]);
+    expect(built.code).not.toBe(cp.code);
   });
 
   it("dials nothing to say it — a reservation is answered before a Core is", async () => {
--- a/packages/cli/src/__tests__/cli-harness.ts
+++ b/packages/cli/src/__tests__/cli-harness.ts
@@ -14,6 +14,13 @@ import { registryPaths, type RegistryPaths } from "../blob-registry.ts";
 import type { CoreProbe, CoreProbeFn } from "../core-probe.ts";
 import { nonInteractiveTerminal, type CliTerminal, type TerminalSignal } from "../cli-terminal.ts";
 import type { OpenCoreShellFn } from "../core-shell-channel.ts";
+import { SessionWriteRefused } from "../session-attach-channel.ts";
+import type {
+  AttachAuthority,
+  OpenSessionAttachFn,
+  SessionAttachExit,
+  SessionAttachment,
+} from "../session-attach-channel.ts";
 import type { CoreConnectFn, CoreConnectOptions, CoreLinkClient } from "../core-connection.ts";
 import type { OpenSessionGateway, SessionGateway, StartedSession } from "../session-gateway.ts";
 import type {
@@ -88,6 +95,8 @@ export type RunOptions = {
   terminal?: CliTerminal;
   /** What `core shell` gets back, or a throw. */
   openShell?: OpenCoreShellFn;
+  /** What `session attach` gets back, or a throw. */
+  openAttach?: OpenSessionAttachFn;
 };
 
 /**
@@ -205,6 +214,11 @@ export function makeCliFixture(): CliFixture {
           opts.openShell ??
           (async () => {
             throw new Error("this test did not expect to open a shell");
+          }),
+        openAttach:
+          opts.openAttach ??
+          (async () => {
+            throw new Error("this test did not expect to attach to a session");
           }),
       });
       return { code, out, err, all: [...out, ...err].join("\n") };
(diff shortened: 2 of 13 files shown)


qcentic-dev:
## Acceptance criteria — #163, line by line

### What

- [x] **`actana session attach <id>` — attach a terminal to a running session, read its output live, and type into it.**
  `session-attach.ts` + `session-attach-channel.ts`, dispatched from `session-command.ts`. Reading is
  `ptySubscribe` + `replay` (scrollback painted first, then the live stream with no gap); typing is
  `write`, gated on the lock. `session-attach-live.test.ts` drives both against a real Core.

### Done when

- [x] **Attaching to a session someone else is writing gives a read-only view with a visible reason, not an error and not a silent steal.**
  A refused `claim` opens read-only, exits `0`, and prints the reason on the line above the
  scrollback: `READ-ONLY: another Core client holds this Session's write lock`. No `forceTakeover`
  frame exists anywhere in the command — `SessionAttachment` has no method for one.
  Tests: *"gives a read-only view with the reason on it, rather than an error or a steal"*,
  *"still reads the Session it may not write"*, and live: *"gives the first the write lock and the
  second a read-only view that says why"*.

- [x] **Detaching releases the lock.**
  `release` on the detach path and on `SIGINT`/`SIGTERM`, sent *after* the terminal is restored.
  Tests: *"hands the lock back when the operator detaches"*, *"hands it back on SIGINT/SIGTERM too"*,
  *"gives the terminal back before the release goes out, not after"*, and live: *"hands the lock to
  the next attach once the first detaches"*.

- [x] **So does the connection dropping.**
  Tested explicitly, against a real Core, with the socket **terminated** — no close frame, no
  `release`, nothing the client chose — and the next attach then gets the lock and types into the
  PTY. The CLI deliberately sends nothing on this path: the Core has already released it (D7), and
  asking would mean a teardown waiting on a socket that is gone. Needed one addition to the #205 rig
  (`InProcessCore.dropConnections`), documented there as to why no client can stage this itself.
  Tests: live *"does not leave the Session locked when the connection drops abruptly"*, and unit
  *"does not try to release over a link that has dropped — the Core already did"*.

- [x] **Two `attach` sessions on the same Session: one writes, one reads.**
  Two real CLI attaches, one real Core, concurrently: both receive the live stream, only the first's
  keystrokes reach the PTY, and the PTY's write log is exactly the holder's bytes.
  Test: live *"gives the first the write lock and the second a read-only view that says why"*.

- [x] **The terminal is restored on every exit path.**
  One test per path: detach, the harness exiting, the link dropping, `SIGINT`, `SIGTERM`, a write
  failing for a reason that is not the lock, and a terminal that refuses raw mode in the first place
  (where raw mode is never entered *and* the lock is still released). Plus: the terminal is never
  touched at all when the attach fails to open. The discipline is #203's, reused — `finally`,
  handled signals, idempotent restore, and `cli-terminal.ts`'s process `exit` hook underneath.
  Suite: *"the terminal is restored on every exit path"* (7 tests), and the live drop test asserts
  `isRaw() === false` after a real socket death.

### Traps

- [x] **Written against the lock from the first commit; not added afterwards.**
  `060ac05` — the first commit — contains `session-attach-channel.ts` with the `claim`/`release`, the
  four authorities, the refusal-before-the-wire, and the lock tests. There is no commit in this
  branch in which attach exists without it.

- [x] **Lock release on an abrupt disconnect — the case that gets missed and the one that strands a Session.**
  See the third bullet above: proved end to end, not asserted as an intention.

### Ticket notes taken on

- [x] **#129 D5's one-connection rationale is dead; the decision stands, but attach does not own the Core.**
  Nothing here assumes sole ownership: this is one connection among many, holding one Session's lock
  and subscribed to one PTY's stream. Two attaches on one Core run concurrently in the live suite,
  which is the direct disproof of the old assumption.

- [x] **Built on the in-process Core rig #205 unified.**
  `session-attach-live.test.ts` uses `__tests__/in-process-core.ts`; the one addition to it is
  `dropConnections`, in that same module rather than a fork.

- [x] **Boundaries.** `packages/cli` only — `packages/sdk`, `release.yml` and `docs/reference` are
  untouched (D11: terminal handling is the CLI's).

### Not claimed

- [ ] **A real harness's TUI in a real terminal.** The ticket says up front that this verb has no
  headless test story, and that is the part it means: everything provable without a person is proved
  above, and how a full-screen harness *feels* through an attach belongs to the acceptance run. No
  automated coverage is claimed for it.

qcentic-dev:
CI note: the first `Unit Tests` run went red on `events-tail-cursor.test.ts > resumes where the last run stopped, and does not replay it` — a 60 s timeout in #205's suite, not one of this branch's. It stops and restarts a Core on the same port and is the slowest file in the package, so it is the one that loses a race with a loaded runner. Every attach test passed in that same run, including the abrupt-drop one (the Core logged `core-link.session-lock.released-on-drop { taskIds: [ 'task_live' ] }`).

Re-ran the failed job with no change to the branch: green. The CLI suite also passes locally four times in a row (`24 passed | 1 skipped`). Flagging it rather than quietly re-running — if it recurs on this base it is worth its own ticket.

qcentic-adm (approved):
**Approving.** Reviewed against #163 line by line, reading the enforcing code rather than the description. Every acceptance criterion is met, the one unchecked box is the one the ticket itself says needs a person, and nothing provable without a person is left unproved. Four nits below, all cosmetic or operator-facing wording — none of them blocks.

## The claims, checked at the enforcing line

**The lock is claimed before a byte can reach the PTY.** `session-attach-channel.ts:181-190` — the order inside `openSessionAttach` is `connect` → `findByTask` → `claim` → `ptySubscribe` → `replay`. The `SessionAttachment` does not exist until after the claim resolves, and `session-attach.ts` does not construct the `RawModeGuard` or register `terminal.onInput` until `deps.openAttach` has returned, so there is no window in which a keystroke has anywhere to go. A refused claim closes the write path twice over: `mayWrite` is false in the channel so `write()` throws `SessionWriteRefused` before the wire (`session-attach-channel.ts:265-268`), and the command's `writes` flag never calls it in the first place (`session-attach.ts:329-348`). `--read-only` sends no `claim` frame at all, which is not the same state as being refused one, and the four authorities are four states rather than a boolean.

**A Session someone else is writing gives a read-only view with a visible reason, and it still streams.** `banner()` (`session-attach.ts:390-402`) puts `READ-ONLY: another Core client holds this Session's write lock` on stderr *before* the scrollback paints, and the exit is `0`. `forceTakeover` appears nowhere in the command and `SessionAttachment` has no method for one — the steal is impossible, not merely declined. The reading half is unconditional: `paint(attached.backlog)` then `attached.onData(paint)` regardless of authority, asserted in *"still reads the Session it may not write"* and live in *"gives the first the write lock and the second a read-only view that says why"*, where both attaches receive a line the PTY emitted after both were connected.

**Detach releases the lock, and so does a drop — without relying on this process's own cleanup.** The asymmetry at `session-attach.ts:246-250` is the right one: release on every ending except `disconnected`, and the channel's `release()` also short-circuits on `!client.isConnected()` (`session-attach-channel.ts:311-317`), so a teardown never waits on a dead socket. The Core-side behaviour it rides on is real and unconditional: `pty-core-link-server.ts:609-616` calls `this.sessionLocks.releaseAll(conn)` in `dropConnection`, *before* the identity guard, and logs `core-link.session-lock.released-on-drop` — the line cited from CI. The live proof is `session-attach-live.test.ts` *"does not leave the Session locked when the connection drops abruptly"*: `core.dropConnections()` calls `ws.terminate()` (`in-process-core.ts:141-151`) — no close frame, no handshake, nothing the client chose — and the next attach is granted the lock and its bytes land in the PTY (`writes.join("|")` is exactly `"typed before the drop|typed after the drop"`). The CLI holding the lock is therefore belt-and-braces, not the mechanism: even the `guard.enter()` failure path releases (`session-attach.ts:221-224`), and any path that skipped it still ends in `client.close()`.

**Two concurrent attaches: one writes, one reads.** Two real CLI runs against one real `PtyCoreLinkServer` on a real `wss://` port. Both see the live stream, the banner distinguishes them, and the PTY's write log is asserted by equality — `expect(pty.writes.join("")).toBe("typed by the holder")` — not by substring, so the reader's keystrokes are proved absent rather than merely not looked for. The fourth live test is a good addition the ticket did not ask for: `--read-only` leaves the Session genuinely claimable by the next attach.

**Terminal restored on every exit path, and restore cannot run twice.** `RawModeGuard.restore()` guards on both `entered` and `restored` (`session-attach.ts:130-139`), and `releaseTerminal()` drains `disposers` with `splice(0)`, so calling it from the happy path *and* the `finally` is a no-op the second time. `rawModeCalls` is asserted to be exactly `[true, false]` on the detach path, which is what turns "idempotent" from a comment into a test. Seven paths have a test each — detach, harness exit, link drop, SIGINT, SIGTERM, non-lock write failure, raw-mode refusal — plus *"touches the terminal at all only after the Session is open"* for the failed open, where `rawModeCalls` is `[]`. The raw-mode-refusal test asserts both halves the ticket cares about: never entered *and* `releaseCount() === 1`. `cli-terminal.ts`'s process `exit` hook sits underneath as the last resort without anything above it depending on it.

**Boundaries and sweeps.** `packages/sdk`, `release.yml` and `docs/reference` untouched — only `packages/cli`. The shared-file footprint is registration and nothing else: `openAttach` beside `openShell` in `cli-deps.ts` and `actana-cli-entry.ts`, one flag in `cli-args.ts` and `SESSION_FLAGS`, and dispatch in `session-command.ts`. `no-local-escape.test.ts` walks `src` recursively with `__tests__` excluded, so both new modules are swept without an edit; the runtime dependency assertion still reads `["@actana/sdk", "ws"]` and `package.json` is not in the diff. `session attach` correctly leaves `exit-codes.ts`'s reservation list, and the reservation suite was rewritten to keep a built-verb counter-example rather than losing the assertion.

The lock is in `060ac05`, the first commit, with the claim, the release, the four authorities and the refusal-before-the-wire — the trap the ticket names is avoided by construction.

## Nits — none blocking

1. **`session-attach.ts:456`** — the dropped-link report says *"The Core releases a dropped connection's Session locks, so this one is claimable again"* unconditionally. For an attach that never held the lock (`held-by-another`, `--read-only`) that sentence is wrong in the operator's favour twice over: this connection had no lock to release, and the Session is still held by the client that had it. Worth gating on whether this attach was the holder.
2. **`session-attach.ts:361`** — `onResize` gates on the `writes` constant captured at attach time, not on current lock state. After a force takeover (`state.lockLost`), the demoted attach keeps sending `resize` frames, and the Core does not gate `resize` on the lock — so it reflows the new holder's PTY. That is exactly the "interference in the one direction the lock cannot see" the module header restrains itself from; the guard should read `writes && !state.lockLost`.
3. **`session-attach.ts:326`** — `quitAt` also uses the captured `writes`, so after a takeover `Ctrl-C` is neither forwarded (there is no lock) nor a detach (the read-only affordance never activates). The operator is told they are read-only from here but the read-only exit key is not what they get. `Ctrl-]` still works, so this is a rough edge rather than a trap.
4. **`session-attach.ts:433`** — the dropped-keystroke line reads *"this attach was read-only"* even when the attach began as the holder and lost the lock mid-session, which reads as though the attach was read-only from the start.

Nit 2 is the one I would actually fix before this leaves draft; the rest are wording.

## Not claimed, and correctly so

The one unchecked box — a real harness's TUI through a real terminal — is the ticket's own words about this verb having no headless test story. The line between "provable without a person" and "the acceptance run" is drawn in the right place here: the lock, the contention, the release-on-drop and every terminal exit path are on the automated side, and only the *feel* of a full-screen TUI through an attach is left for the acceptance run. CI is green across all 22 checks on this head, and the one earlier red was correctly identified as a pre-existing timeout in #205's `events-tail-cursor.test.ts`, flagged rather than quietly re-run.

qcentic-dev:
All four nits fixed in `bdb4a0d` (on top of `c40b258`).

**Nit 2 — `onResize` after a takeover.** The one you'd fix before it leaves draft, and the one that
reached another operator's terminal. `onResize` now reads `writes && !state.lockLost`, through a
single `canWrite` that the keyboard path uses as well — the two places that were gating on the flag
captured at attach time now both follow the lock as it stands. Test: *"stops resizing the PTY the
moment the lock is taken away"* — a resize before the takeover lands, one after it does not, and the
assertion is on the full `resizes` log by equality so the second frame is proved absent rather than
not looked for.

**Nit 3 — `Ctrl-C` after a takeover.** The honest behaviour is that a demoted attach is a Reader, so
`Ctrl-C` means what it means for every other Reader: it detaches. Same `canWrite`, so there is no
second rule to keep in step. The affordance changing is not left for the operator to discover by
pressing the key — the takeover line on the terminal now reads `… — this attach is read-only from
here, and Ctrl-C now detaches instead of going to the harness.` Test: *"makes Ctrl-C a detach once
the lock has been taken away"*, which asserts the exit is a detach *and* that nothing typed after the
takeover reached the far side.

**Nit 1 — the dropped-link report.** D7 releases the locks the dropped connection *held*, so the line
is now per-authority rather than unconditional: a holder gets "claimable again"; `held-by-another`
gets "This attach never held the write lock, so the Session stays held by the client that does";
`--read-only` gets "…so the drop leaves the Session's lock as it was"; an attach that lost the lock
mid-session gets "…stays held by the client that took it"; and `no-lock-table` gets no line at all,
since that Core arbitrates no writes. Three tests, one per shape, each asserting `not.toContain`
`"claimable again"` alongside the line that is true.

**Nit 4 — the dropped-keystroke tally.** Says "this attach lost the write lock partway through" when
`lockLost`, and keeps "this attach was read-only" for the attach that always was. Worth noting what
the tally does *not* count: the chunk that discovers the takeover is refused on the wire and answered
on the terminal at that moment, so the count is the keystrokes typed after this attach already knew.
That is unchanged behaviour, and the test names it.

The module header's resize paragraph was rewritten to cover both halves, since "a read-only attach
does not resize the PTY" no longer describes the restraint — it follows the lock, not the opening
authority.

`packages/cli`: `24 passed | 1 skipped`, 297 tests, `tsc --noEmit` clean. `events-tail-cursor.test.ts`
took its known 60 s timeout on the first local full run and passed on re-run and on every subsequent
full run — same pre-existing flake flagged above, untouched.

Issue #164, opened 2026-08-11
2.8 Parity check, then delete the frozen prototype (D15)


mrfqcentic:
**Base branch: `feat/cli-and-publishing`, not `main`.** Train: **`beta/0.2.2`**.
Worktree: **`wt-129-14`**. Suggested branch: **`chore/delete-frozen-prototype`**.
Phase 2, step 4, alone. Implements the second half of **D15**. **This ticket closes phase 2.**

## What

Check `packages/cli` against the frozen prototype in `docs/reference/experiment-cli/`, then delete the frozen tree.

## Done when

- A written parity check in this ticket: every command the prototype had, and where it now lives — or an explicit, argued decision not to carry it.
- The 90-line hand-copied protocol mirror is gone from the repository. After this, one definition exists, in `@actana/sdk` (D4).
- `docs/reference/experiment-cli/` no longer exists, and nothing references it.
- No credentials or logs were ever committed, in this branch or the freeze branch. Verify against history, not the working tree.

## Traps

- **Deleting before parity is checked loses the reference for good.** The check is the ticket; the deletion is a consequence of it.
- Parity is not "the same flags". Behaviours the prototype got right by accident — the paste-vs-Enter handling, the terminal emulator in `tail` — are now the Core's and the SDK's, and belong in the check as "moved to X", not "dropped".

Refs #129

qcentic-dev:
Delivered by #209, which landed on `beta/0.2.2` with the train integration PR #212 (squash merge `6a33b805461006ca8553fcba1141bdecdff07eba`). Closing manually — this PR's base was not the default branch, so it did not auto-close.

PR #207, opened 2026-08-13
chore(docs): freeze the prototype CLI as a read-only reference under docs/reference/ (D15)

### LINKED ISSUE #158: 2.2 Freeze the prototype CLI as a read-only reference (D15)

qcentic-dev:
## Summary

The prototype `core-client` CLI — the throwaway tool that drove a real Core over core-link before `packages/cli` existed — is copied into `docs/reference/experiment-cli/` as a frozen, read-only reference, so the parity check in phase 2 step 4 has something to check the shipped `actana` command against.

This is repo hygiene, not documentation. The diff is 17 moved files plus a 23-line freeze notice. **Nothing in the prototype is refactored, reformatted or fixed** — including the parts that are wrong.

## Related issues

Closes #158
Refs #129, #164

## Type of change

- [x] 📚 Documentation
- [x] 🔧 Build / CI / tooling

## What changed

The tarball carried **19 entries — 17 files and 2 directories** (`src/`, `src/commands/`), and 19 macOS AppleDouble `._*` resource forks alongside them. The `._*` files are `com.apple.provenance` xattr metadata, 163 bytes each, no content; they are **excluded**. The 17 files are 13 `.ts`, 3 `.md`, 1 `.json` — 2,620 lines, of which 2,056 are TypeScript and the manifest.

The copy is in **two commits on purpose**, so the fidelity claim is auditable rather than asserted:

- `6bc4740` — the tree, **byte-for-byte identical to the tarball**, verified with `diff -r`.
- `0f742bf` — the freeze notice, **additive only** (23 insertions, 0 deletions).

The prototype ships its own root `README.md`, so the required notice is prepended to it rather than put in a second file competing for the same path. The prototype's own text is untouched below a rule.

### 1. Outside the reach of lint, typecheck, test and build; not a workspace package

The ticket's trap is that `docs/**` gets picked up more often than people expect, so each of these is a **run**, not a reading of config:

| Check | Result |
|---|---|
| `pnpm install --frozen-lockfile` | links **5** workspace packages — cli, core, panel, sdk, shared. Prototype absent. Lockfile unchanged. |
| `pnpm typecheck` | clean; `Scope: 5 of 6 workspace projects` (the 6th is the root) |
| `pnpm lint` | 0 errors, 51 pre-existing warnings. **0 lines of output reference `docs/`** |
| `pnpm build` | clean |
| `pnpm test` | see below |
| `pnpm npm:rehearse` | discovers exactly `@actana/sdk` and `@actana/cli`; `absent=` empty |
| `pnpm scan:secrets` | `No obvious secrets found.` |

**No neutralization was needed, and none was applied** — the prototype `package.json` is left byte-faithful. It has no `private: true`, so the hazard in PR #200 is real in principle, but neither discovery path reaches it:

- **pnpm** — `pnpm-workspace.yaml` globs `packages/*` only.
- **npm publish discovery** — `workspaceManifests()` in `scripts/lib/npm-packages.mjs` is a `readdirSync` of `packages/`, not a `**/package.json` glob. It cannot see `docs/`.

Renaming it would have cost the byte-fidelity that the step-4 parity check depends on, for no gain the rehearsal doesn't already demonstrate. The rehearsal is the proof: `packages=@actana/sdk @actana/cli`, with `core-client` nowhere in it.

The other reason this matters: the prototype's `src/blob.ts` calls `execFileSync("docker", …)` to fetch its own credential — **exactly** what `packages/cli/src/__tests__/no-local-escape.test.ts` (#129 D9) forbids. That sweep is rooted at `packages/cli/src`, so the prototype is out of its reach by construction. Run explicitly, **7/7 pass**. Had this tree landed inside `packages/`, that test would be red — which is the sharpest available evidence the exclusion holds.

**The no-local-escape sweep and the secret scan both pass.**

### 2. Credentials and logs are excluded — checked **before** committing

The tarball was described as pre-scanned clean; it was verified independently anyway, in the scratchpad, **before any file entered the repository**:

- **High-entropy / token-shaped strings (≥32 chars)** — **zero matches** across all 17 files. This is the strongest single result here.
- **Base64 payloads (≥40 chars)** — zero. No PEM armour, no `BEGIN` of any kind, no JWT (`eyJ`) anywhere.
- **Credential keyword sweep** (`bearer`, `token`, `secret`, `password`, `api_key`, `credential`, `cookie`, `auth`, `session log`, `.pem`, `.key`) — many hits, **every one of them prose or code that *describes* credentials**: `blob.ts`'s header saying the blob is real credentials and must stay out of version control, `help.ts`'s CREDENTIALS topic, `client.ts` sending an `auth` frame it reads from a blob at runtime. No literal value is embedded.
- **Endpoints** — every host is loopback: `127.0.0.1:9443`, `127.0.0.1:9444`, `127.0.0.1:7420`, `localhost:7420`. No external endpoint, no IP, no credentialed URL (`scheme://user:pass@`).
- **Artifact sweep** — no `.log`, `.pem`, `.key`, `.crt`, `.p12`, `.env`, `.sqlite`, and **no blob file**. `src/blob.ts` is source that *reads* `blob.txt` and `blobs/*.txt` at runtime; neither those files nor the `blobs/` directory are in the tarball. File extensions present are exactly `.ts`, `.md`, `.json`.
- **No session logs.** The prototype streams session output over the wire; nothing captured is in the tree.
- The AppleDouble forks were inspected (`od -c`) before being dropped — macOS provenance metadata only.

`pnpm scan:secrets` was then run **pre-commit**, when the 17 files were untracked: the scanner's candidate set includes `git ls-files --others --exclude-standard`, so it genuinely enumerated all 17 and passed. Re-run post-commit against the tracked files: passes again.

### 3. The stale protocol mirror is present and left alone

`docs/reference/experiment-cli/src/protocol.ts`, **exactly 90 lines**, hand-copied, unmodified, and deliberately not brought up to date. Its drift from the real core-link protocol is the point of keeping it, and the README says so in as many words.

### 4. A README line saying what this is

`docs/reference/experiment-cli/README.md` opens with the notice: what the tree is, that it is frozen and read-only, that nothing may import it or fix it, that `src/protocol.ts` is stale on purpose, and that **the directory is deleted at the end of phase 2 by #164**.

## How was this tested?

- [ ] Unit tests added/updated — none; this ticket adds no code that runs
- [x] Manually verified

Full suite run locally. **One pre-existing failure, unrelated to this change:** `scripts/__tests__/core-launcher.test.mjs` — "execs the bundled node on the bundled actana CLI" — expects a temp install root and gets `/opt/actana/app`, because this machine has a real `/opt/actana` install. It is the same failure PR #200 recorded. **Proved independent of this change** by moving `docs/reference/` out of the tree entirely and re-running: it fails identically.

Because the root `vitest run` aborts on that failure, `pnpm -r test` was run separately so the package suites — the half containing the no-local-escape sweep — actually ran:

| Suite | Result |
|---|---|
| root `vitest run` | 473 passed, 1 failed (pre-existing, above) |
| `packages/sdk` | 167 passed, 5 skipped |
| `packages/cli` | 297 passed, 3 skipped |
| `packages/shared` | 180 passed |
| `packages/core` | 826 passed |
| `packages/panel` | 1338 passed |

Everything except the one environment-caused failure is green.

## Notes for the reviewer

**The prototype is wrong in places, on purpose.** `protocol.ts` is stale, the README describes a `run-core-local.sh` that is not in the tarball, and `blob.ts` shells out to `docker`. None of it is fixed. If a reviewer's instinct is to correct something in this directory, the answer is no — it is a frozen record, and #164 deletes it.

**`docs/reference/` is new** and nothing indexes it. `docs/README.md` is deliberately not edited: no test asserts docs-index completeness, and the directory is temporary by construction, so adding it to an index only creates a dangling entry for #164 to clean up.

**Nothing was merged, `main` was not touched, and this PR is a draft.**

### DIFF
--- a/docs/reference/experiment-cli/README.md
+++ b/docs/reference/experiment-cli/README.md
@@ -1,0 +1,284 @@
+# experiment
+
+Tooling for running an Actana Core locally in a container and driving it from
+the command line — creating projects, launching Claude Code sessions inside it,
+reading what they said, and wiring a Panel to visualise the lot.
+
+Two things live here:
+
+- **`run-core-local.sh`** — builds the Core image and runs one instance
+- **`core-client.ts`** — talks to that Core over its core-link protocol
+
+## Requirements
+
+- Docker
+- Node 24 (`node` must be v24 — the client is TypeScript run directly, and the
+  repo's build scripts refuse anything older)
+- The `mission-control-updated` checkout beside this folder, for the image build
+  and for the `ws` dependency reached through the `node_modules` symlink
+
+## Quick start
+
+```bash
+./run-core-local.sh                                  # build and run the Core
+docker exec rest-test cat /home/core/.config/actana/registration-blob.txt > blob.txt
+node core-client.ts status                           # prove the connection
+node core-client.ts project create /home/core/work   # register a directory
+node core-client.ts session start <projectId> "summarise the README"
+node core-client.ts session logs <taskId>
+```
+
+---
+
+# `run-core-local.sh`
+
+Builds the Core tarball and image if they are missing, then creates the
+container. Safe to re-run: it reuses what exists and recreates only the
+container.
+
+```bash
+./run-core-local.sh [--rebuild] [--reset]
+```
+
+| Argument | Meaning |
+|---|---|
+| `--rebuild` | force a fresh tarball and image even if they exist |
+| `--reset` | destroy the volume too — **this unpairs the Core** and loses its data |
+
+| Environment | Default | Meaning |
+|---|---|---|
+| `NAME` | `rest-test` | container name |
+| `PORT` | `9443` | host port, and the Core's own port (they must match) |
+| `PUBLIC_HOST` | `127.0.0.1` | the address baked into the certificate and pairing token |
+| `IMAGE` | `actana-core:$NAME` | image tag |
+| `VOLUME` | `$NAME-home` | named volume holding all Core state |
+| `REPO` | current directory | path to the mission-control-updated checkout |
+
+On macOS the Core tarball is built inside a Linux container automatically:
+native modules are copied from the build host and never cross-compiled, so a
+tarball built on macOS cannot go into a Linux image.
+
+---
+
+# `core-client.ts`
+
+```bash
+node core-client.ts <resource> <verb> [target] [options]
+```
+
+Resource first, then verb. The verbs repeat across resources, so `project ls`,
+`session ls` and `harness ls` all behave the same way. `node core-client.ts
+help` prints this from the CLI, and `help <resource>` drills into one.
+
+**Targets** are flexible: a project is named by its id, its name or its path; a
+session by its task id or its pty id.
+
+**`--core <name>`** picks which Core to run against, the way kubectl picks a
+context. It is global — it may appear anywhere in the line, and every resource
+honours it.
+
+## core
+
+Each Core is one blob file in `blobs/`, named by the file's basename.
+
+| Command | Arguments | What it does |
+|---|---|---|
+| `core ls` | — | list them: name, endpoint, label; `*` marks the one in effect |
+
+| Name | Container | Endpoint |
+|---|---|---|
+| `developer` | `rest-test-2` | `wss://127.0.0.1:9444` |
+| `pr` | `rest-test` | `wss://127.0.0.1:9443` |
+
+```bash
+node core-client.ts core ls
+node core-client.ts --core pr project ls
+node core-client.ts project ls --core pr     # identical
+AC_CORE=pr node core-client.ts project ls    # for a whole shell
+```
+
+Naming them by purpose rather than by container or port is the point: a rebuild
+that moves a port leaves `--core pr` working where a memorised
+`wss://127.0.0.1:9443` would not.
+
+Cores are separate machines as far as this client is concerned. Projects,
+sessions and harness logins on one are invisible to the other, so a task id from
+`developer` means nothing to `pr`.
+
+Adding one is copying its blob in:
+
+```bash
+docker exec <container> cat /home/core/.config/actana/registration-blob.txt > blobs/<name>.txt
+```
+
+**Output** splits by stream — data on stdout, commentary on stderr:
+
+```bash
+SID=$(node core-client.ts session start scratch "summarise the README")
+```
+
+## project
+
+Directories on the Core that sessions are allowed to run in. Registration is
+not bookkeeping: the Core rejects any session whose working directory is not
+inside a registered project.
+
+| Command | Arguments | What it does |
+|---|---|---|
+| `project ls` | — | list projects: id, name, path |
+| `project create` | `<path> [name]` | register a directory on the Core; name defaults to the last path segment |
+| `project rm` | `<project> --yes` | unregister a project **and every session under it**; the directory on disk is untouched |
+| `project browse` | `[path]` | browse the Core's filesystem, to find a path worth registering |
+| `project inspect` | `<project>` | full detail as JSON |
+
+## session
+
+| Command | Arguments | What it does |
+|---|---|---|
+| `session ls` | `[--all]` | sessions with a live PTY; `--all` includes finished ones |
+| `session start` | `<project> [prompt]` | launch a session, print its task id, exit |
+| `session logs` | `<session>` | what the session has said so far |
+| `session attach` | `<session>` | print the scrollback and keep streaming |
+| `session send` | `<session> <text>` | wait for the screen to settle, type text, submit |
+| `session enter` | `<session>` | press Enter only — unsticks a pending paste |
+| `session interrupt` | `<session>` | press Escape — stop the turn, keep the session |
+| `session resume` | `<task>` | relaunch a finished session with its conversation intact |
+| `session kill` | `<session>` | end the session's process group for good |
+| `session inspect` | `<task>` | full detail as JSON |
+
+`interrupt` and `kill` are not the same thing: interrupt stops what Claude is
+doing and leaves the session usable, kill ends it.
+
+## harness
+
+The agent CLIs the Core can launch. They are not baked into the image — they
+install at runtime into the Core's home, which is the persistent volume.
+
+| Command | Arguments | What it does |
+|---|---|---|
+| `harness ls` | — | availability, version and install path per harness |
+| `harness install` | `<id>` | start installing one; returns immediately |
+
+Ids: `claude-code`, `codex`, `cursor-cli`, `opencode`.
+
+`install` is fire and forget. The install runs on the Core and takes minutes,
+and the protocol only acknowledges the start, never the outcome — so the
+command returns and `harness ls` is how you check whether it landed. The Core
+re-probes when the install finishes; success means the probe now finds it.
+
+Installing gets you the binary, not a login. Harness credentials live in the
+Core's home directory, so each Core authenticates separately.
+
+## status and events
+
+| Command | Arguments | What it does |
+|---|---|---|
+| `status` | — | endpoint, label, protocol version, project and session counts, harness availability |
+| `events` | `[--since N] [--kind <prefix>]` | stream the Core's event log |
+
+`status` is also the cheapest proof the credentials work: it opens a real mTLS
+connection and authenticates before printing anything.
+
+Event kinds are prefixed `task:`, `session:`, `pty:`, `project:`, `agents:` and
+`harness:`.
+
+## Options
+
+| Option | Applies to | Effect |
+|---|---|---|
+| `--core <name>` | everything | which Core to run against; may go anywhere in the line |
+| `--follow` | `session start`, `session logs` | stream instead of returning |
+| `--all` | `session ls` | include finished sessions |
+| `--lines N` | `session logs` | how many lines to show (40) |
+| `--raw` | `session logs` | untouched bytes, no terminal emulation |
+| `--cols N` `--rows N` | `session logs` | replay geometry; must match the spawn (120x32) |
+| `--harness <id>` | `session start` | which harness to launch (default `claude-code`) |
+| `--model <id>` | `session start`, `session resume` | model to launch with |
+| `--title <text>` | `session start` | task title instead of one derived from the prompt |
+| `--enter` / `--no-enter` | `session send` | press Enter after the text (default), or hold it unsubmitted |
+| `--now` | `session send` | skip the settle wait |
+| `--ask-permissions` | `session start`, `session resume` | launch **without** auto mode |
+| `--yes` | `project rm` | confirm the cascade |
+| `--since N` `--kind <prefix>` | `events` | resume point, and filter by prefix |
+| `--help` | anywhere | help for that resource |
+
+## Credentials
+
+The pairing blob is resolved in order:
+
+| Source | Notes |
+|---|---|
+| `--core <name>` | `blobs/<name>.txt` |
+| `AC_BLOB` | the base64 blob itself |
+| `AC_BLOB_FILE` | a file holding it |
+| `AC_CORE` | `blobs/<name>.txt`, the ambient default |
+| `./blob.txt` | the unnamed local copy, as before |
+| `docker exec` | against `$AC_CONTAINER`, default `rest-test` |
+
+Nothing is cached between runs: each invocation decodes the blob, opens the
+mTLS socket, and sends `auth` as its first frame. Each blob carries a client key
+and a bearer — **do not commit them**. `blobs/` is mode 700 and its files 600,
+and a `.gitignore` beside them excludes the directory.
+
+A `--core` name is a bare identifier, not a path: `--core ../../something` is
+rejected as an invalid name rather than read.
+
+---
+
+## Behaviour worth knowing before you rely on it
+
+**Auto mode is the default.** Sessions launch with
+`--dangerously-skip-permissions`, so Claude does not ask for tool approvals —
+inside the container it reads, writes and executes in the project directory
+unattended. `--ask-permissions` opts out.
+
+**Startup dialogs are answered automatically.** Claude opens two blocking
+dialogs — the bypass-permissions warning and the folder-trust prompt — once per
+project per Core, and they are stored in the Core's home volume, so a new Core
+means meeting them again. A session parked on one accepts no input at all, and
+the bypass warning's highlighted default is `1. No, exit`, so anything that
+presses Enter at it quits Claude instantly. `session start` and `session resume`
+watch the screen and answer them, which is what makes auto mode unattended
+rather than merely unprompted.
+
+**`start` types the prompt itself** once the TUI stops redrawing. The Core's own
+initial-input path fires 450 ms after spawn, before Claude accepts input, and
+loses it.
+
+**`send` sends the Enter separately** from the text, after a pause that scales
+with length. Claude treats a burst of input as a paste and absorbs a same-frame
+carriage return into it, leaving `[Pasted text #1 +1 lines]` unsubmitted.
+
+**`logs` emulates a terminal** rather than stripping escape codes, because a TUI
+positions text with cursor moves instead of spaces. It only works while the
+session is alive — the scrollback belongs to the PTY.
+
+**`resume` needs the harness session id** to have been recorded on the task. A
+session that died early may not have one.
+
+---
+
+## The Panel
+
+A Panel container visualises the same Core in a browser.
+
+| Container | Port | Purpose |
+|---|---|---|
+| `panel-test` | `127.0.0.1:7420` | the Panel, volume `panel-test-data` |
+| `panel-test-link` | — | socat relay inside the Panel's network namespace |
+
+The relay exists because the pairing token points at `wss://127.0.0.1:9443` and
+the Core's certificate is issued for that exact address — inside another
+container `127.0.0.1` would be that container. It forwards raw TCP to the
+Core's published port on the host, so the TLS passes through untouched.
+
+To pair: open http://localhost:7420, create the Operator, and paste `blob.txt`
+into **Add Core** (`pbcopy < blob.txt`).
+
+## Other documents
+
+| File | Contents |
+|---|---|
+| [note.md](note.md) | the requirements, as asked for |
+| [action.md](action.md) | what was built and where things stand |
+| [client-reference.md](client-reference.md) | the client's internals and module layout |
--- a/docs/reference/experiment-cli/README.md
+++ b/docs/reference/experiment-cli/README.md
@@ -1,3 +1,26 @@
+<!-- FROZEN REFERENCE — added by #158. Everything below the rule is the
+     prototype's own README, copied unchanged. -->
+
+# The prototype CLI, frozen
+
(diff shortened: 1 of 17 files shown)


qcentic-dev:
## Acceptance checklist — issue #158

Every criterion from the ticket, each checked or explicitly unchecked with the reason.

### Done when

- [x] **The tree is in.** `docs/reference/experiment-cli/` — 17 files, 2 directories, the 19 tarball entries. Byte-for-byte identical to the tarball (`diff -r`, commit `6bc4740`); the only later change is the additive 23-line freeze notice on `README.md` (`0f742bf`). The 19 macOS AppleDouble `._*` forks are `com.apple.provenance` metadata and are excluded.

- [x] **Outside the reach of lint, typecheck, test and build — proved by running each, not by reading config.** `pnpm lint` 0 errors and **0 output lines referencing `docs/`**; `pnpm typecheck` clean at `Scope: 5 of 6 workspace projects`; `pnpm build` clean; `vitest` root include is `scripts/**/*.test.mjs` and the package suites are per-package, so no run touches the tree.

- [x] **It must not become a workspace package.** `pnpm install --frozen-lockfile` links 5 packages (cli, core, panel, sdk, shared) and leaves the lockfile untouched. `pnpm npm:rehearse` discovers exactly `@actana/sdk` and `@actana/cli`, `absent=` empty — `core-client` is not in the publish set. **No neutralization was needed and none was applied:** `pnpm-workspace.yaml` globs `packages/*`, and PR #200's discovery (`workspaceManifests()` in `scripts/lib/npm-packages.mjs`) is a `readdirSync` of `packages/`, not a `**/package.json` glob. The manifest is left byte-faithful because renaming it would cost the fidelity the step-4 parity check needs, and the rehearsal already proves the exclusion holds.

- [x] **Credentials and logs are excluded — checked BEFORE committing, not after.** Swept in the scratchpad before any file entered the repo: **zero** high-entropy strings (≥32 chars), **zero** base64 payloads (≥40 chars), no PEM armour, no JWT, no credentialed URL, no `.log`/`.pem`/`.key`/`.crt`/`.p12`/`.env`/`.sqlite`, and **no blob file** — extensions present are exactly `.ts`, `.md`, `.json`. Every credential-keyword hit is prose or code that *describes* credentials; no literal value is embedded. All endpoints are loopback (`127.0.0.1`, `localhost`). `pnpm scan:secrets` run **pre-commit** while the files were untracked — the scanner's candidate set includes untracked files, so it genuinely enumerated all 17 — and again post-commit. Passes both times. The tarball was said to be pre-scanned clean; this was verified independently regardless.

- [x] **The no-local-escape sweep still passes.** `packages/cli/src/__tests__/no-local-escape.test.ts` run explicitly: **7/7**. Worth naming: the prototype's `src/blob.ts` calls `execFileSync("docker", …)` to fetch its own credential — exactly what that sweep (#129 D9) forbids. The sweep is rooted at `packages/cli/src`, so the tree is out of its reach by construction; had it landed inside `packages/`, this test would be red.

- [x] **The secret scan still passes.** `pnpm scan:secrets` — `No obvious secrets found.`

- [x] **The hand-copied 90-line protocol mirror is present and obviously stale.** `src/protocol.ts`, exactly 90 lines, unmodified. **Deliberately not updated** — that is the point of keeping it, and the README says so.

- [x] **A README line says what this is, that it is frozen, and that it is deleted at the end of phase 2.** `docs/reference/experiment-cli/README.md` opens with it, naming **#164** as the ticket that deletes the directory.

### Traps

- [x] **Treated as repo hygiene, not documentation.** Branch `chore/freeze-experiment-cli`; lead commit typed `chore(docs)`; the diff is moved files plus one notice.

- [x] **Exclusions verified by running the full suite, not by reading config.** Full suite plus the npm rehearsal, listed in the PR body.

- [x] **Nothing in the prototype refactored, reformatted or fixed.** Including the parts that are wrong: the stale protocol mirror, the `docker` shell-out, and a README describing a `run-core-local.sh` that is not in the tarball.

### Process

- [x] Work confined to the new worktree `/home/core/work/wt-158`, cut from `origin/feat/cli-and-publishing` at `3e32673`. *(The ticket's `wt-129-08` is stale; `wt-158` used as instructed.)*
- [x] Branch `chore/freeze-experiment-cli`, pushed with `-u`. **Never force-pushed.**
- [x] Conventional Commits on every commit — `chore(docs): …`, `docs(reference): …`.
- [x] PR is a **draft**, base `feat/cli-and-publishing`, **not `main`**. Title 90 chars (≤112).
- [x] Nothing merged, `main` untouched, PR **not** marked ready for review.

### One caveat, stated plainly

- [ ] **"Full suite green" — not literally true, and not caused by this change.** `scripts/__tests__/core-launcher.test.mjs` fails: it expects a temp install root and gets `/opt/actana/app`, because this machine has a real `/opt/actana` install. Same failure PR #200 recorded on an untouched checkout. **Proved independent** by moving `docs/reference/` out of the tree and re-running — it fails identically. Everything else passes: 473 root, 167 sdk, 297 cli, 180 shared, 826 core, 1338 panel.

### Judgement call worth a reviewer's eye

The prototype ships its own root `README.md`, so the required freeze notice was **prepended to it** rather than placed in a second file competing for the same path. It is additive — 23 insertions, 0 deletions, the prototype's text intact below a rule — and the preceding commit is the byte-for-byte copy, so the tree's fidelity stays auditable. If you would rather the notice lived in a separate file and the prototype README stayed untouched at the byte level, say so and I will restructure it.

qcentic-adm (approved):
Reviewed against #158's four "done when" clauses and both traps. Every claim below was checked by reading the enforcing line or the CI log, not the PR description. **Approving** — with one thing to do before merge that is not a change to this diff.

---

## 1. The leak check — swept independently, clean

The prototype ran against live Cores, so I re-did this from scratch rather than trusting the author's sweep or `Secret Scan`. I extracted the tree from the PR head with `git archive` and scanned all 17 blobs in **binary mode** (perl, `<:raw`), because grep is not trustworthy here — see the note at the end of this section.

- **Every entry accounted for.** `git ls-tree -r -t` at the head: 17 blobs, 5 trees, **all mode `100644`** — no symlinks (`120000`), no gitlinks (`160000`), no executables. 100,808 bytes total. No `._*` AppleDouble forks, no dotfiles, no `blobs/` directory, no `blob.txt`.
- **Token-shaped runs** `[A-Za-z0-9+/=_.-]{28,}` with Shannon entropy — 19 hits, **all** filesystem paths, CLI flags or identifiers (`/home/core/.config/actana/registration-blob.txt`, `--dangerously-skip-permissions`, `agentsAvailabilityListResult`). Nothing above H=4.16 at any length. No opaque token anywhere in the tree.
- **Hard secret shapes** — PEM armour, `PRIVATE KEY`, JWT (`eyJ…`), hex ≥32, base64 ≥40, `AKIA/ASIA`, `ghp_`/`gho_`, `xox[abposr]-`, `sk-`, and credentialed URLs (`scheme://user:pass@`): **zero matches, all ten patterns.**
- **Literal assignment to a credential key** — `(bearer|caCert|clientCert|clientKey|token|secret|password|apiKey|blob)\s*[:=]\s*["'\`]…`: **zero.** The longest double-quoted string in any file is 60 chars and is a user-facing message (`src/commands/session.ts`, `"(nothing running — add --all to include finished sessions)"`). There is no payload in this tree.
- **Endpoints** — every host is loopback: `127.0.0.1:9443`, `:9444`, `:7420`, `localhost:7420`, plus the placeholders `wss://host:port` (`src/protocol.ts:15`) and `wss://:` (`src/blob.ts:122`, a template). No external host, no non-loopback IP.
- **Credential keywords** — 169 hits across the tree, and I read every line of all of them. Every one is prose or code that *describes* credentials: `src/blob.ts:3` warning the blob is real credentials, `src/commands/help.ts:52` CREDENTIALS topic, `src/client.ts:125` sending `bearer: blob.bearer` read from a blob **at runtime**. No value is embedded.
- **Artifacts** — extensions present are exactly `.ts`, `.md`, `.json`. No `.log`, `.pem`, `.key`, `.crt`, `.p12`, `.env`, `.sqlite`. No session transcript: `action.md` narrates the live runs in prose and quotes no captured output.

**Why binary-safe mattering is not hypothetical.** `scripts/scan-secrets.mjs:88` is `if (!isText(buffer)) continue;`, and `isText` is `!buffer.includes(0)` — **the repo's own scanner silently skips any file containing a NUL byte.** A leak inside such a file would have passed `pnpm scan:secrets` and the CI `Secret Scan` in green. My per-file census reports `NUL=0` for all 17 files, so nothing here was skipped — but that is a fact I established, not one the green check established.

Related, and worth saying plainly: `scan-secrets.mjs` carries eight vendor-specific patterns (AWS, GitHub, Stripe, OpenAI, Google, Slack, PEM). **None of them matches an Actana pairing blob** — base64 JSON carrying `clientKey` and `bearer`. The green `Secret Scan` is close to zero evidence for the exact leak class #158 names. The manual sweep is the evidence, and it holds.

The one identifier-shaped string in the tree is the project id `p-msip6kgd-54d3f5` at `docs/reference/experiment-cli/action.md:126`, pointing at a scratch project on a disposable throwaway container. Not a credential and not sensitive. Noted for completeness, no action.

**Verdict: no leak. This clause passes.**

## 2. Outside lint, typecheck, test, build, the workspace and the publish set

Checked at the enforcing line in each case, then against CI:

| Reach | Enforcing line | Result |
|---|---|---|
| pnpm workspace | `pnpm-workspace.yaml` → `packages: ["packages/*"]` | one glob, one level, `docs/` unreachable |
| npm publish discovery | `scripts/lib/npm-packages.mjs` → `workspaceManifests()` is `readdirSync(path.join(repoRoot, "packages"))` | a directory read, not a `**/package.json` glob — it cannot see `docs/` |
| lint | root `package.json` → `eslint packages scripts screenshot.mjs eslint.config.mjs` | explicit path list; `docs/` is never an argument |
| typecheck | root `package.json` → `pnpm -r typecheck` | recursive over workspace projects only; there is no root `tsconfig.json` at all |
| build | root `package.json` → `--filter @actana/core` + `--filter @actana/panel` | filtered |
| root tests | `vitest.config.ts` → `include: ["scripts/**/*.test.mjs"]` | cannot reach `docs/` |

**On the neutralization question:** none was applied, and `docs/reference/experiment-cli/package.json` is left byte-faithful with no `private: true`. That is the right call. Both discovery paths are directory-scoped rather than glob-scoped, so `private: true` would buy nothing while costing the byte-fidelity #164's parity check depends on.

**And CI proves it rather than the local rehearsal.** `scripts/__tests__/npm-publish.test.mjs:115,121,166-167` calls `discoverPublishable(repoRoot)` and `workspaceManifests(repoRoot)` against the **real repository root** — so the publish-set assertion executed on CI with this `package.json` present on disk, and passed (41 tests). Six CI jobs ran `pnpm install --frozen-lockfile` successfully with it present, so the lockfile is genuinely untouched. That is stronger evidence than the author's local `pnpm npm:rehearse`.

**On the root-run abort, and whether the evidence covers the no-local-escape sweep — it does.** The author's local narrative (root `vitest run` aborted on `core-launcher.test.mjs`, so `pnpm -r test` was run separately) is sound but no longer the best evidence available. On CI: root vitest reports **474 passed / 16 files, zero failures** — the `/opt/actana` explanation for the local failure is confirmed, since a clean runner has no such install — and `packages/cli`'s **`no-local-escape.test.ts` passes 7/7**. The sweep the author points to as the sharpest proof of the exclusion actually ran, in an environment with no aborted prefix. The claim is fully covered.

**One non-blocking gap.** The lint exclusion is *incidental*, not structural. `eslint.config.mjs`'s `ignores` (lines 8-16) does not list `docs/**`, and the flat config's `files: ["**/*.{ts,tsx,js,mjs}"]` would match the frozen tree. It is out of reach only because the `lint` script passes explicit paths. Anyone running bare `eslint .`, or a future edit of that script to `eslint .`, lints a tree that must never be fixed. A one-line `"docs/**"` in `ignores` would make this hold by construction. Not a condition of approval — `pnpm lint` is the repo's contract and it is green — but worth a line either here or in #164's cleanup.

## 3. The stale protocol mirror — present, stale, untouched

`docs/reference/experiment-cli/src/protocol.ts` is **exactly 90 lines**, and its blob is `a1cc2d0f` in both `6bc4740` and the head — not reformatted, not updated. The staleness is intact and visible: `PROTOCOL_VERSION = "0.15.0"` (line 7), the comment at line 3 still naming the pre-rebrand `mission-control-updated` repo, and `claudeSessionId` on `Task` (line 49). That drift is the artefact #164 checks against. Correct.

## 4. The freeze notice — the placement call stands

`README.md:1-22`: an HTML comment marking the boundary, then the notice, then a `---` rule, then the prototype's own text from line 24 with nothing altered. It covers all three things #158 asks for and more — what the tree is, that it is frozen and read-only, that it is not built/linted/typechecked/tested/published and why (naming both enforcing lines), that nothing may import or fix it, that `src/protocol.ts` is stale on purpose, and that **the directory is deleted at the end of phase 2 by #164**.

**The offer to restructure should be declined; prepending is the better structure.** A separate `FREEZE.md` would leave `README.md` as the file GitHub renders on the directory view and the file a reader opens first — so the reader most likely to start fixing things would meet the prototype's own text with no frozen marker on it. The notice belongs where the reader lands. The cost of prepending is that the tree is no longer byte-identical to the tarball, and that cost is already paid off by putting the pristine copy in the preceding commit: `git show 6bc4740:docs/reference/experiment-cli/README.md` is the unmodified file, which is exactly how I verified clause 5 below. No restructure needed.

## 5. Moved files only — confirmed against the pristine copy

- `git diff --name-status` base→head: **17 paths, every one `A`, every one under `docs/reference/experiment-cli/`.** Nothing else in the repository is touched by either commit — no config, no `.gitignore`, no workspace file, no lockfile.
- Head tree vs the pristine copy `6bc4740`: **one blob differs**, `README.md`, by `+23 / -0` at the top. The other 16 blobs are byte-identical by hash. No reformat, no refactor, no fix reached any file after the pristine commit.
- `6bc4740` itself touches nothing outside `docs/reference/`.

The honest limit: fidelity to the *original tarball* is not verifiable from this repository, since the tarball is not in it. What is verifiable is that nothing moved after the pristine commit, and the internal evidence is what an unedited copy looks like — the pre-rebrand repo name, `PROTOCOL_VERSION 0.15.0`, a dangling `run-core-local.sh`, a dangling `[note.md](note.md)` link at `README.md:305`, and `src/blob.ts:99` shelling out to `docker exec` for its own credential, which is precisely what `packages/cli`'s `no-local-escape.test.ts` forbids in shipped code. A cleanup pass would have removed exactly those. They are all still here.

## 6. Chore series, compliant subjects

`chore(docs): freeze the prototype CLI as a read-only reference` and `docs(reference): say what the frozen prototype CLI is, and when it goes`. Both conventional, both well inside `header-max-length: 120`, both with `Refs #158, #129` / `Refs #158, #164` trailers preceded by a blank line — which is what `commitlint.config.mjs`'s custom `trailer-leading-blank` rule requires. Branch `chore/freeze-experiment-cli` matches the Conventions job's pattern. `Conventions` is green. The second commit typed `docs:` rather than `chore:` is fine — it adds prose, and #158's "it is a `chore/`" is about the branch and the PR, which it is.

---

## The one thing to do before merge — re-run CI

**`Unit Tests` is red on this PR.** `packages/cli/src/__tests__/events-tail-cursor.test.ts > "resumes where the last run stopped, and does not replay it"` — `Error: Test timed out in 60000ms` ([job 94385410190](https://github.com/actana/control/actions/runs/31680751101/job/94385410190)). The PR description's "everything except the one environment-caused failure is green" describes the local run and was written before this finished; as it stands the PR carries a failing required check.

It is **not** caused by this diff, and I checked rather than assumed:

- The whole diff is 17 added files under `docs/reference/experiment-cli/` — `git diff --name-status` shows no other path.
- That test file's blob is `4a962c67` on **both** the base and this head — byte-identical, untouched.
- It builds an in-process Core and tails its event log; nothing in it reads the repository tree.
- Root vitest passed 474/474 in the same job, and `no-local-escape.test.ts` passed 7/7.

So it is either a flaky 60-second timeout or something #206's merge brought onto `feat/cli-and-publishing` — note that `feat/cli-and-publishing` at `3e32673` has never had CI run on it directly, and this is the first run containing that merge, so the two are not distinguishable from one data point. **Please re-run the job before merging.** If it reproduces, it belongs to the base branch and wants its own ticket — it is not a change to make here.

---

## Checklist

| # | #158 acceptance criterion | Verdict |
|---|---|---|
| 1 | Tree is in, outside lint / typecheck / test / build | ✅ verified at each enforcing line **and** green on CI |
| 2 | Not a workspace package | ✅ `pnpm-workspace.yaml` globs `packages/*`; 5 projects linked, prototype absent |
| 3 | Outside PR #200's npm publish discovery | ✅ `workspaceManifests()` is a `readdirSync` of `packages/`; `npm-publish.test.mjs` asserts it against the real root and passes on CI |
| 4 | Credentials and logs excluded — checked before committing | ✅ independently swept, binary-safe, all 17 files; zero secret-shaped strings |
| 5 | No blob file, no session log, no key material, loopback endpoints only | ✅ |
| 6 | 90-line protocol mirror present and obviously stale | ✅ exactly 90 lines, blob unchanged, staleness intact |
| 7 | README says what it is, that it is frozen, and that it goes at end of phase 2 | ✅ `README.md:1-22`, names #164 |
| 8 | Diff is mostly moved files; nothing refactored, reformatted or fixed | ✅ 17 × `A`; head = pristine copy + 23 additive README lines, nothing else |
| 9 | `chore/`, compliant commit subjects | ✅ Conventions green |
| 10 | CI green | ⚠️ one unrelated timeout in an untouched test — re-run before merge |
| — | eslint `ignores` lacks `docs/**` | 💬 non-blocking suggestion |

**Verdict: approved.** Every clause of #158 is met, and each one is provable at the line that enforces it rather than by assertion. The leak check — the part that actually mattered — is clean under an independent binary-safe sweep, which is a stronger statement than the green `Secret Scan` could have made, since that scanner has no pattern for a pairing blob and skips NUL-bearing files outright. The freeze notice placement is right and should stay. The red check is a flake or a base-branch problem in a file this PR does not touch: re-run it, and if it sticks, file it separately. Nothing here needs changing.

Issue #208, opened 2026-08-13
Flaky: events-tail-cursor.test.ts times out in CI under runner load


qcentic-dev:
`packages/cli/src/__tests__/events-tail-cursor.test.ts` — the test named
**"resumes where the last run stopped, and does not replay it"** times out in CI
on `feat/cli-and-publishing` under runner load. It has cost at least three
re-runs across three different branches in under ten hours, each of which passed
on the second attempt with no change to the code.

This issue is **evidence only**. Nothing is changed in the test by the PR that
files this (#164 / the frozen-prototype deletion) — the flake is pre-existing and
touching it there would hide it behind an unrelated diff.

## The failure

```
FAIL  src/__tests__/events-tail-cursor.test.ts
  > actana events tail, across a Core restart
  > resumes where the last run stopped, and does not replay it
Error: Test timed out in 60000ms.
 ❯ src/__tests__/events-tail-cursor.test.ts:125:3
```

## The failing runs

Three occurrences, three branches, identical message and identical line. Each is
**attempt 1** of a run whose attempt 2 went green:

| When (UTC) | Branch | Job | File duration |
|---|---|---|---|
| 2026-08-12 23:15 | `feat/cli-project-harness-events` | [94292271085](https://github.com/actana/control/actions/runs/31650083998/job/94292271085) | 61,756 ms |
| 2026-08-13 07:05 | `feat/cli-session-attach` | [94371297266](https://github.com/actana/control/actions/runs/31676231867/job/94371297266) | 61,692 ms |
| 2026-08-13 08:10 | `chore/freeze-experiment-cli` (PR #207) | [94385410190](https://github.com/actana/control/actions/runs/31680751101/job/94385410190) | 63,069 ms |

The third is the one PR #207's review called out as a red required check that
the diff could not have caused, and it re-ran green.

## Why it looks like load rather than a deadlock

The three timings are 61.7 s, 61.7 s and 63.1 s against a 60 s budget — the test
is finishing a little **past** its deadline, not hanging on one. A deadlock would
sit at the deadline every time regardless of the machine; overshooting it by one
to three seconds is what a slow runner looks like.

The test is a genuine integration test rather than a unit test, which is where
the budget goes: it opens a free port, boots an in-process Core, registers it
through `actana core add`, then runs `actana events tail` twice — the second run
proving the on-disk cursor resumed rather than replayed. Two Core boots and two
full CLI invocations inside one 60 s budget is not a lot of headroom on a shared
runner.

[...]

PR #209, opened 2026-08-13
chore(docs): parity-check packages/cli against the frozen prototype, then delete it (D15)

### LINKED ISSUE #164: 2.8 Parity check, then delete the frozen prototype (D15)

qcentic-dev:
## Summary

The parity check #164 asks for, then the deletion it authorises. **The check is
the ticket; the deletion is its consequence** — the two are separate commits in
that order, so the reference still existed when it was read.

- `32d925c` — `docs/reference/cli-parity-check.md`: every command and every flag
  the prototype `core-client` had, and where it lives in `packages/cli` now, or
  the argument for not carrying it.
- `bfdc5d4` — `docs/reference/experiment-cli/` deleted: 17 files, 2,643 lines,
  including the 90-line hand-copied protocol mirror.

This closes phase 2.

## Related issues

Closes #164
Refs #129, #158, #207
Adjacent hygiene filed separately: #208 (the pre-existing `events-tail-cursor`
CI flake — **not** touched by this PR, see "CI" below)

## Type of change

- [x] 📚 Documentation
- [x] 🔧 Build / CI / tooling

### 1. A written parity check — `docs/reference/cli-parity-check.md`

The prototype's **6 resources, 20 verbs and 19 flags**, each mapped. Read off
the implementation, not the help text: `src/commands/help.ts` had already
drifted from `src/args.ts` in two places (it documents `--enter` and `--now`,
which `VALUE_FLAGS` never listed; it omits `--name`, which it does).

**Parity is not "the same flags".** The behaviours the prototype got right by
accident are recorded as *moved to X* with the module named — not as dropped,
because "the CLI does not do this" and "this no longer happens" are different
claims and only the first is true:

| Prototype behaviour | Now |
|---|---|
| Split the paste and the carriage return, with a length-scaled pause | **Core** — `packages/core/src/harness-prompt-delivery.ts`, step 3 |
| Wait for the *rendered screen* to stop changing, not for silence | **Core** — same file, step 1 (`redrawSignature`) |
| Answer the harness's blocking startup dialogs | **Core** — same file, step 2 (`readDialogOptions`, `chooseDialogOption`, `matchBlockingDialog`) |
| The terminal emulator in `tail` (`src/terminal.ts`, 236 lines) | **SDK** — `packages/sdk/src/terminal-screen.ts` (`TerminalScreen`, 848 lines) |
| Keep scrolled-off lines: the transcript is the scrollback | **SDK** — `terminal-screen.ts:279` (`scrollbackLines`) |
| Demux `data`/`exit` per PTY, holding early frames until the id arrives | **SDK** — `packages/sdk/src/core-session.ts:350-370` |
| Request→result-frame correlation (`RESULT_OF`) | **SDK** — `packages/sdk/src/core-link-transport.ts` |
| Per-harness canonical binary and auto-mode flag (both marked "LOCAL STOPGAP") | **SDK** — `core-session.ts:76-81`, `:92-97` |

Two of these got **better** in the move, and the check says so. The Core answers
a dialog by reading the menu and pressing that option's own digit, sending the
confirming Enter only once the highlight has moved — where the prototype pressed
the digit and then Enter unconditionally, into a dialog whose highlighted default
is "No, exit". And `packages/cli/src/__tests__/no-prompt-timing.test.ts` now
fails the build if any timer reappears in the CLI.

**Two genuine gaps, argued rather than glossed:**

- **`project rm`** — a cascading archive (every Task under the Project goes with
  it) behind a boolean `--yes` is the right shape for a throwaway tool and the
  wrong one for a published command. Not lost: the frame is in `@actana/sdk` and
  the Panel uses it. Wants its own ticket and a confirmation flow that is not a
  flag.
- **`--model`** — the prototype's spelling only worked for `claude`, one harness
  of four; the honest version needs a per-harness table beside
  `HARNESS_LAUNCH_COMMANDS` and the Core's spawn policy extended in step. The
  capability exists one layer down as `CoreSessionOptions.command`.

**Seventeen flat aliases dropped as one decision, not seventeen.** They existed
for scripts written before the prototype grew resources; `actana` has not
shipped, so there are none. `ls` meaning `project browse` while `project ls`
means listing was the specific trap.

### 2. `docs/reference/experiment-cli/` is gone, and nothing references it

Binary-safe sweep of every tracked file for `experiment-cli`:

- **Before** the deletion — 1,045 files, **zero** occurrences outside the tree.
- **After** — 1,029 files, five occurrences, **all** of them prose in the parity
  check describing what was removed. A parity check has to name what it checked;
  naming is not depending. No import, no path reference, no build input.

**On #207's review note about `eslint.config.mjs`:** the reviewer observed the
frozen tree was out of lint's reach only *incidentally* — `pnpm lint` passes
explicit paths and `docs/**` is not in `ignores` — and suggested a one-line
`"docs/**"`. **Deliberately not added.** Deleting the tree removes what needed
protecting, and the only file left under `docs/` that eslint's glob matches is
`docs/assets/src/build.mjs`, a live build script that *should* be linted. Adding
the ignore now would silence it. Argument in §7 of the check.

### 3. One protocol definition, in `@actana/sdk`

Sweeps run **binary-safe**, because `scripts/scan-secrets.mjs:88` is
`if (!isText(buffer)) continue;` where `isText` is `!buffer.includes(0)` — the
repo's own scanner silently skips any file containing a NUL byte, and this
repository has **two source files that contain one**
(`packages/core/src/__tests__/core-link-client-id-reclaim.test.ts` and
`packages/panel/src/components/views/UserTerminalPane.tsx`). Every blob was read
as raw bytes and decoded latin1 so no byte can be dropped or merged.

| Sweep | Before | After |
|---|---|---|
| Files declaring a core-link frame type | 2 | **1** — `packages/sdk/src/core-link-frames.ts` |
| `*_PROTOCOL_VERSION` declarations, core-link | 2 | **1** — `CORE_LINK_PROTOCOL_VERSION`, `core-link-frames.ts:1257` |
| `RESULT_OF` correlation tables | 1 | **0** |

The only other `*_PROTOCOL_VERSION` in the tree is `PANEL_LINK_PROTOCOL_VERSION`
(`packages/panel/src/shared/panel-link.ts:57` and its copy in
`scripts/lib/panel-e2e.mjs:30`) — the Panel↔browser protocol, a different wire.
Everything else naming a frame is a **consumer**, not a second copy.

One honest note, recorded in §5 so the next person running this sweep finds the
answer rather than the question: the registration **blob** shape is declared
twice, in `packages/sdk/src/core-registration-blob.ts:34` and
`packages/shared/src/registration-blob.ts:20`. That is not this mirror and not
undeclared drift — it is ADR 0025 D4 working as written, and the SDK file argues
its own copy at lines 17-24 (a published package cannot import a type from a
private one). A blob is not a core-link frame.

### 4. No credentials or logs were ever committed — verified against history

Against **history, not the working tree**. The freeze branch was deleted from the
remote, so its two commits were recovered from `refs/pull/207/head` rather than
taken on trust from the squash.

**Exactly three commits ever touched the path, on any ref:**

| Commit | What |
|---|---|
| `6bc4740` | the pristine copy, freeze branch — the commit behind #207's squash |
| `0f742bf` | the freeze notice, freeze branch |
| `8c3d5b2` | the squash on `feat/cli-and-publishing` |

Every blob in all three trees read as raw bytes and scanned for PEM armour,
`PRIVATE KEY`, JWTs, hex ≥32, base64 ≥40, `AKIA`/`ASIA`, `gh[pousr]_`, `xox[abposr]-`,
`sk-`, credentialed URLs (`scheme://user:pass@`), and any literal assigned to a
credential-shaped key (`bearer`, `caCert`, `clientCert`, `clientKey`, `token`,
`secret`, `password`, `apiKey`, `blob`, `privateKey`).

- **Zero findings, all three commits, all 17 blobs, all ten shapes.**
- Every entry mode `100644` — no symlinks, no gitlinks, no executables.
- **NUL=0 on all 17 blobs**, so nothing here was skipped by the scanner — a fact
  established by this scan, not by the green `Secret Scan`.
- No `blob.txt`, no `blobs/`, and no `.log`/`.pem`/`.key`/`.env`/database file
  has **ever** been added anywhere in this repository, on any ref.

## The full suite

Run before pushing, on this branch.

| Check | Result |
|---|---|
| `pnpm install --frozen-lockfile` | clean, lockfile unchanged |
| `pnpm lint` | **0 errors**, 51 warnings — the same 51 #207 recorded |
| `pnpm typecheck` | clean, 5 of 6 workspace projects |
| `pnpm build` | clean |
| `pnpm -r test` | **sdk 167 · shared 180 · cli 297 · core 826 · panel 1338 — all pass** |
| root `vitest run` | 473/474 — see below |

The one red test is `scripts/__tests__/core-launcher.test.mjs > execs the bundled
node on the bundled actana CLI`, which fails **on this machine only**: it asserts
the launcher resolves its temp install root and gets `/opt/actana/app`, because
this machine has a real Actana install and the launcher finds it first. It is
environmental, unrelated to this diff, and green on CI — #207's review confirmed
root vitest at 474/474 on a clean runner. Because the root run fails first,
`pnpm test`'s `&&` chain stops before `pnpm -r test`, so the package suites were
run separately, which is what the table above reports.

## CI

`packages/cli/src/__tests__/events-tail-cursor.test.ts` may go red here on the
"resumes where the last run stopped" test. **It is a pre-existing flake on this
train and is not touched by this PR** — filed with its evidence as **#208**:
three timeouts in under ten hours across three different branches, each attempt 1
of a run whose attempt 2 went green, each overshooting a 60 s budget by one to
three seconds. A re-run is the current answer.

## What a reviewer should check

1. That `32d925c` really does precede `bfdc5d4` — the ticket's first trap is
   deleting before the check is written, and commit order is the only thing that
   makes the claim auditable rather than asserted.
2. That the *moved to X* rows in §4 of the check name a module that actually does
   the thing. They were each verified by reading the target, not by matching
   names.
3. The two argued omissions, `project rm` and `--model`. Those are the rows where
   this PR is making a judgement rather than recording a fact, and they are the
   ones worth disagreeing with.


qcentic-dev:
## Acceptance criteria — #164, each checked or explicitly not

### The ticket's four "done when" clauses

- [x] **A written parity check: every command the prototype had, and where it now lives — or an explicit, argued decision not to carry it.**
  `docs/reference/cli-parity-check.md`, committed in `32d925c`. 6 resources, 20 verbs, 19 flags, 17 aliases and 5 environment variables, each with a destination. Read off the prototype's implementation rather than its help text, because the two had already drifted. Two verbs are argued omissions rather than re-spellings — `project rm` (a cascading archive behind a boolean is the wrong shape for a published command) and `--model` (the prototype's spelling worked for one harness of four) — and both arguments are written out rather than left as gaps.

- [x] **The 90-line hand-copied protocol mirror is gone; one definition exists, in `@actana/sdk`.**
  `docs/reference/experiment-cli/src/protocol.ts` (exactly 90 lines) deleted in `bfdc5d4`. Binary-safe sweeps after: **1** file declaring core-link frame types (`packages/sdk/src/core-link-frames.ts`), **1** core-link `*_PROTOCOL_VERSION` (`CORE_LINK_PROTOCOL_VERSION`, line 1257), **0** `RESULT_OF` tables. The remaining `PANEL_LINK_PROTOCOL_VERSION` is the Panel↔browser wire, a different protocol. Per ADR 0025 D1/D3.

- [x] **`docs/reference/experiment-cli/` no longer exists, and nothing references it.**
  Deleted in `bfdc5d4` — 17 files, 2,643 lines. Sweep before: 1,045 tracked files, **zero** occurrences outside the tree. Sweep after: 1,029 files, **five** occurrences, all of them prose inside the parity check describing what was removed. **Stated plainly rather than hidden:** those five are why this box is checked and not left blank — a parity check has to name the thing it checked, and there is no import, no path reference and no build input among them. If a reviewer reads the criterion as "the string appears nowhere at all", this clause is not met and the check document would have to stop naming its own subject.

- [x] **No credentials or logs were ever committed, in this branch or the freeze branch — verified against history, not the working tree.**
  Exactly three commits ever touched the path on any ref: `6bc4740` (the pristine copy behind #207's squash), `0f742bf` (the freeze notice), `8c3d5b2` (the squash). The freeze branch is gone from the remote, so its two commits were recovered from `refs/pull/207/head` rather than taken on trust. All 17 blobs in all three trees read as raw bytes: **zero findings** across ten secret shapes plus credential-key assignment. All entries mode `100644`. No `blob.txt`, `blobs/`, `.log`, `.pem`, `.key`, `.env` or database file has ever been added anywhere in this repository, on any ref.

### The ticket's two traps

- [x] **The check is the ticket; the deletion is its consequence.**
  Two commits, in that order: `32d925c` (check) then `bfdc5d4` (deletion). Nothing was deleted until the check was written and committed, so the reference existed when it was read.

- [x] **Parity is not "the same flags" — behaviours the prototype got right by accident are "moved to X" with the module named, not "dropped".**
  §4 of the check, 12 rows. The paste-vs-Enter prompt delivery and the terminal emulator in `tail` are both there by name: the first is `packages/core/src/harness-prompt-delivery.ts` (steps 1–3), the second is `packages/sdk/src/terminal-screen.ts`. Each target was verified by reading it, not by matching names. Two behaviours got *safer* in the move and the check says so.

### The additional requirements on this PR

- [x] **The whole repo swept for references, including #207's eslint-ignores note.** Addressed and answered: the suggested `"docs/**"` ignore is **deliberately not added**, because after the deletion the only file under `docs/` matching eslint's glob is `docs/assets/src/build.mjs` — a live build script that should be linted. Argument in §7.
- [x] **Binary-safe sweep method.** Every sweep reads raw bytes and decodes latin1. Confirmed the trap is real, and it is worse than one file: `packages/core/src/__tests__/core-link-client-id-reclaim.test.ts` **and** `packages/panel/src/components/views/UserTerminalPane.tsx` both contain a NUL byte, and `scripts/scan-secrets.mjs:88` skips both.
- [x] **Full suite run before pushing.** lint 0 errors / 51 pre-existing warnings · typecheck clean · build clean · `pnpm -r test` 2,808 passing across five packages · root vitest 473/474. The single failure is `core-launcher.test.mjs`, environmental on this machine (a real `/opt/actana` install wins over the test's temp root) and green on CI.
- [x] **Adjacent hygiene: the `events-tail-cursor` flake filed separately, with the failing runs as evidence, and referenced from this PR body.** #208 — three timeouts in under ten hours across three branches, each attempt 1 of a run whose attempt 2 went green.
- [x] **The flake test is not changed in this PR.** `packages/cli/src/__tests__/events-tail-cursor.test.ts` is untouched; the diff is two paths under `docs/`.
- [x] **Conventional Commits on every commit.** `docs(reference): …` and `chore(docs): …`, both well inside `header-max-length: 120`, both with a blank line before the `Refs` trailer as `commitlint.config.mjs`'s `trailer-leading-blank` requires.
- [x] **Draft PR, base `feat/cli-and-publishing`, not `main`.** Base confirmed `feat/cli-and-publishing`; `isDraft: true`; title 89 characters, inside the 112 limit.
- [x] **Never force pushed.** One `push -u` of a new branch, no rewrite of any kind.
- [x] **Nothing merged, `main` untouched, PR left as a draft.** No merge performed, no commit or branch operation touched `main`, and this PR is not marked ready for review.

### Not done, and why

- [ ] **`project rm` and `--model` carried into `packages/cli`.** Not attempted, and not an oversight — each is recorded in the check as an argued omission with the reason and the shape a proper fix would need. Both want their own ticket: `project rm` needs a confirmation flow that is not a boolean over a cascading archive, and `--model` needs a per-harness table beside `HARNESS_LAUNCH_COMMANDS` with the Core's spawn policy extended in step. #164 asks for an argued decision, not for the features.
- [ ] **CI green on this PR.** Not assertable at the time of writing — the run is still in flight, and #208 may make it red on a test this diff does not touch. A re-run is the current answer.

qcentic-adm (approved):
## Reviewed against #164, criterion by criterion — approving

I checked every "done when" clause against the enforcing line rather than the description, re-ran the two sweeps and the credential scan independently, and read the frozen tree at `32d925c` (pre-deletion) rather than trusting the check's account of it. Everything the PR claims, it does. Notes below are non-blocking.

### The trap, first: the check really does precede the deletion

`32d925c` 08:34:22 → `bfdc5d4` 08:34:58, and `git ls-tree -r 32d925c -- docs/reference/experiment-cli | wc -l` is **17** — the reference was still on disk when the check was committed. Not merely asserted; auditable, which was the point.

### 1. The parity check is the ticket — it holds up against the frozen tree

The counts are real. 5 `RESOURCES` plus the specially-routed `core` = 6 resources; the `switch` statements give project 5 + core 1 + harness 2 + status 1 + events 1 + session 10 (`kill`/`rm` share a row) = **20 verbs**; §2 has **19 flag rows**; `ALIASES` in `core-client.ts:39-60` has exactly **17** entries. Every prototype verb and flag is present with a destination.

I verified ~16 line citations into the frozen tree and **all of them land**: `session.ts:13-14` (`COLS`/`ROWS`), `:24-29` (`HARNESS_COMMAND`), `:39-43` (`HARNESS_AUTO_FLAG`), `:70-80` (`typeInto`), `:114-148` (`PtyStream.adopt`), `:165-176` (`STARTUP_DIALOGS`), `:188-227`, `:235-264`, `:253-255`, `:300-314` (`buildClaudeCommand`), `:339-341`, `:392`, `:624-625`; `protocol.ts` exactly **90 lines** with `:7`, `:3`, `:49`, `:73-90` all as described; `terminal.ts` **236 lines** with scrollback at `:36-39` and `:235`; `client.ts:142-151`; `events.ts:8`.

Spot-checked **eight** moved-to rows at the named module, not by name-matching:

- **paste-vs-Enter — confirmed, and the "got safer" claim is real.** `packages/core/src/harness-prompt-delivery.ts` carries steps 1/2/3 in its header exactly as the check describes, and every named symbol exists: `lastScreenClearIndex:105`, `redrawSignature:173`, `readDialogOptions:288`, `chooseDialogOption:312`, `dialogsForHarness:325`, `matchBlockingDialog:338`. Step 3's split write is at `:757`/`:765` with the length-scaled pause computed at `:436-437`. The safety claim is enforced, not just narrated — step 2 states it never presses Enter to clear a dialog, presses the option's own digit, and sends the confirming Enter only once the highlight has moved. Against the prototype's unconditional digit-then-`\r` at `session.ts:253-255`, into a dialog whose default is "No, exit", that is a genuine improvement.
- **the `tail` terminal emulator — confirmed.** `packages/sdk/src/terminal-screen.ts` is **848 lines** as claimed, `scrollbackLines()` at `:279`, `DEFAULT_COLS`/`DEFAULT_ROWS` at `:94-96`.
- Also confirmed: per-PTY demux (`core-session.ts:350-370`, held frames filtered by id), `HARNESS_LAUNCH_COMMANDS:76-81`, `HARNESS_SKIP_PERMISSION_FLAGS:92-97`, `SETTLED_SESSION_STATUSES` carrying `interrupted` at `:112`, `harness-resume.ts` covering all four harnesses, the `LIVE` column at `session-command.ts:370-388`, bare `--enter` at `:463-479`, `harness-command.ts` 15-minute wait, and `--kind` repeatable at `cli-args.ts:136-139`.

**The two omissions are argued, not waved off.** `project rm` names the mechanism (`projectsMutate {op:"archive"}`, cascading — verified in the frozen `project.ts`), the constraint (#129 D10 scope, ADR 0022), and where the capability survives; it labels itself "the one thing in the prototype that `actana` cannot do" rather than burying it. `--model` is accurate on a point that is easy to get wrong: the prototype *spelled* `--model` for all four harnesses but the Core's per-harness allow-list only accepted it for `claude`, so "worked for one of four" is the correct reading. Both are the right call.

### 2. `experiment-cli` is gone, and nothing references it

My own binary-safe sweep of `pr/209` (every blob read as raw bytes, decoded latin1): **1,029 tracked files, exactly 5 occurrences of `experiment-cli`, all in `docs/reference/cli-parity-check.md`** (lines 4, 365, 406, 427, 430) — prose, no import, no path reference, no build input. The "1,045 before" figure is the base commit `8c3d5b2`, and 1045 + 1 − 17 = 1029 reconciles exactly. Your reading of the criterion is the right one: a check that may not name its own subject is not a check.

Swept eslint too, and the argument for **not** adding `"docs/**"` is verified rather than merely plausible: `ignores` at `eslint.config.mjs:9-17` genuinely lacks `docs/**`, `files` is `**/*.{ts,tsx,js,mjs}`, `lint` passes explicit paths, and `docs/assets/src/build.mjs` is the *only* file left under `docs/` matching the glob. Adding the ignore would silence a live build script. Agreed.

### 3. Exactly one protocol definition

Independent declaration sweep, binary-safe. `CORE_LINK_PROTOCOL_VERSION = "0.15.0"` at `core-link-frames.ts:1257` is the only core-link wire-version declaration; `PANEL_LINK_PROTOCOL_VERSION` (`panel-link.ts:57`, `scripts/lib/panel-e2e.mjs:30`) is a different wire. `RESULT_OF`: **0** occurrences outside the check's prose. Every other `CoreLink*` export is either `Extract<>`-derived from the frames union (`core-link-transport.ts:42-52`, `panel-link.ts:33`/`:66`) or a local interface (`core-link-socket.ts`, `core-link-cursor-storage.ts`, `core-connection.ts:54`, `core-link-manager.ts:39`) — consumers, exactly as §5 says. The `core-registration-blob.ts` note is correct and the SDK file does argue its own copy in its header.

Your NUL caveat is worth keeping: 38 files in the tree contain a NUL byte, and after excluding PNG/woff2/audio it is exactly the two source files you named. `scripts/scan-secrets.mjs:88` would skip both, so a green Secret Scan is indeed not evidence about a file of that kind.

### 4. Credentials — method verified, then reproduced

The method is the part that matters, and it is right. `6bc4740` and `0f742bf` are contained in **no origin branch** — only `refs/pull/207/*`. A check against origin refs alone would have silently missed the pristine copy behind the squash, which is exactly the commit most worth scanning. Recovering them from `refs/pull/207/head` was necessary, not belt-and-braces.

I re-ran the scan myself over all three commits: **17 blobs each, every entry mode `100644` blob, NUL=0, zero findings** across PEM armour, `PRIVATE KEY`, JWT, hex ≥32, base64 ≥40, `AKIA`/`ASIA`, `gh[pousr]_`, `xox[abposr]-`, `sk-`, credentialed URLs, and credential-key literal assignment. Repo-wide over every path ever added on any ref, the only matches for the log/credential shapes are `.env.example` and `deploy/.env.example` — both pure prose with no values, so the claim stands as written.

### 5 & 6. The flake, and CI

#208 is filed with real evidence, not a gesture: three job URLs, three branches, timings of 61.7 s / 61.7 s / 63.1 s against a 60 s budget, and the reasoning that overshooting by one to three seconds looks like load rather than a deadlock. It is explicitly evidence-only and this diff touches nothing but two paths under `docs/`, so the test is untouched.

**CI is fully green** — all 22 checks pass or skip, zero failures, `Unit Tests` included. The flake did not fire on this run, so the "red only on #208" caveat is moot.

---

### Non-blocking notes

**a) File the two follow-up tickets.** The check says "Wants its own ticket" for `project rm` and "That is a ticket, not a line" for `--model`, and no such issue exists (#208 is the newest in the repo). Since **this ticket closes phase 2**, both arguments currently have nowhere to land after it closes — an argued omission with no ticket decays into an undocumented gap. You filed #208 for adjacent hygiene; these two deserve the same treatment, and the paragraphs are already written.

**b) `packages/sdk/src/terminal-screen.ts:24` cites `experiment/action.md` §4** as where the scrollback rule was learned. The deleted `docs/reference/experiment-cli/action.md` §4 ("What had to be fixed along the way") contains precisely that lesson, so this PR removes the last in-repo copy of the cited content. The sweep could not catch it — it looks for the literal token `experiment-cli`, and this citation is spelled `experiment/action.md`. Mitigating, and why this is a note rather than a change request: that path never resolved here anyway — `experiment/findings-151-node22-mtls.md:118` already records "`experiment/action.md` does not exist anywhere in this repo or its history" — so nothing that worked is broken. Worth repointing the comment at `docs/reference/cli-parity-check.md` §4, which now carries the lesson.

**c) Two small over-claims in the check.**
- §4 and the PR body say `no-prompt-timing.test.ts` "fails the build if any timer appears in the package". The test actually maintains a named allow-list — `harness install` carries a timer, documented with its reasons. The guarantee that matters still holds (a timer in a file that writes to a Session fails even when listed), but "any timer" is stronger than what is enforced.
- §1's `status` row enumerates the prototype's six fields including `label`, then concludes "Nothing is dropped". `actana core status` prints Endpoint / Core id / Protocol / Multi-connection / Bearer — `label` is carried by `actana core ls` (`core-command.ts:199`, `:219`) instead. True in substance, but it is the one field in that row with no stated destination.

None of these affects the deletion's safety or the criteria. The check is the strongest artefact in the ticket: it was written against a tree that still existed, it is accurate at the line level in both directions, and it argues its two gaps instead of hiding them. Approving.

Issue #211, opened 2026-08-13
actana session --model: a per-harness model selector, honest across all four harnesses


qcentic-dev:
The prototype `core-client` took `--model <id>` on `session start` and
`session resume`. The shipped `actana` does not, and the parity check calls this
"the second real gap" — the other being `project rm` (filed alongside this).

This ticket exists so that argument has somewhere to land now that #164 has
closed phase 2. The paragraph is already written — see
[`docs/reference/cli-parity-check.md`][doc] **§2 Flags**, "`--model <id>` is not
carried, and this is the second real gap."

## Why it was not carried

Not because the flag is unwanted. Because the prototype's version only worked by
accident, and copying the spelling would ship that accident:

- The prototype put `--model` straight into the launch command
  (`buildClaudeCommand`, `session.ts:300-314`) and the Core's spawn allow-list
  **happened** to permit it for `claude`.
- It *spelled* `--model` for all four harnesses, but only one of the four
  actually accepted it. A flag that silently does nothing on three harnesses out
  of four is worse than no flag: it fails quietly, at the point where the user
  most believes they have chosen something.
- The shipped path builds launch commands from `HARNESS_LAUNCH_COMMANDS`
  (`packages/sdk/src/core-session.ts:76-81`), so there is no single string to
  interpolate a model id into any more.

**The capability is not lost.** `CoreSessionOptions.command` in `@actana/sdk` is
the seam for overriding a launch command outright, so an SDK caller can do this
today. What is missing is a CLI spelling that is honest across all four
harnesses.

## What this ticket has to build

The parity check names the two pieces and deliberately implements neither:

1. **A per-harness model table beside `HARNESS_LAUNCH_COMMANDS`** — how each
   supported harness names its model selector, and which have none. A harness
   with no selector must make `--model` an *error*, not a no-op. That is the
   whole point.
2. **The Core's spawn policy extended in step.** The allow-list is what made the
   prototype's flag work for `claude` and nothing else; adding argument shapes
   for the other harnesses is a Core change, not a CLI one, and shipping the CLI
   half alone reproduces exactly the silent-failure mode above.

Both halves land together or the flag is dishonest again.

## Scope notes

- Applies to `session start` and `session resume`, matching the prototype.
- Not a bug against #160 or #163 — the shipped CLI is correct as it stands. This
  is the ticket for adding the flag *properly*.
- Validation belongs where the table is: an unknown model id for a harness that
  does take one should fail before a Session is spawned, not inside it.

## Where it would live

[...]

Issue #210, opened 2026-08-13
actana project rm: carry the prototype's removal verb, with a confirmation flow that is not a boolean


qcentic-dev:
`actana project` ships three verbs — `ls`, `add`, `browse`. The prototype
`core-client` had a fourth, `project rm <project> --yes`, and it is the one thing
in that tool the published CLI cannot do.

This is **not** a regression to repair quietly. The omission was argued in
writing, and this ticket exists so the argument has somewhere to land now that
#164 has closed phase 2. The paragraph is already written — see
[`docs/reference/cli-parity-check.md`][doc] **§1 `project`**, "`project rm` is not
carried, deliberately" — and it names the three things that make this a ticket
rather than a line:

- The Core's removal operation is `projectsMutate {op: "archive"}` and it
  **cascades**: every Task recorded under the Project goes with it. The prototype
  guarded that with a `--yes` boolean and a sentence of prose. That is the right
  shape for a throwaway tool driven by the person who wrote it, and the wrong one
  for a published command — a destructive cascade behind a flag whose whole
  safety story is "the operator read the error the first time" is exactly the
  shape that gets scripted with `--yes` already in the line.
- `actana project`'s three verbs are the ones #129 D10 scoped, and no frame
  carries a path after creation ([ADR 0022][adr0022]), so the noun is
  deliberately small and refuses its extra verbs in writing rather than by
  omission.
- **The capability is not lost today.** The Panel archives Projects over the same
  frame, and the frame is in `@actana/sdk`, so an SDK caller has it now. What is
  missing is the CLI spelling, not the mechanism.

## What this ticket has to decide

The parity check names three open questions and deliberately answers none of
them:

1. **A confirmation flow that is not a boolean.** The cascade is the reason.
   Type-the-name, an explicit count of what goes with it, or a two-step
   `--dry-run` first — but not a bare `--yes` that a script carries by default.
2. **What "archive" should be called at the CLI.** The frame says `archive`, the
   prototype said `rm`, and they are not obviously the same promise to a reader.
   Picking `rm` for familiarity while the Core archives is a mismatch worth
   deciding on purpose.
3. **Whether the cascade is shown before it runs.** Archiving a Project takes its
   Tasks; a command that does not say so before asking is not really asking.

## Not in scope

Re-litigating the omission itself. The shipped CLI is correct as it stands —
this is the ticket for adding the verb *well*, whenever it is wanted, not a bug
against #164 or #205.

## Where it would live

[...]

PR #212, opened 2026-08-13
feat(cli): the actana CLI and npm publishing — phase 2 of the SDK extraction

### LINKED ISSUE #157: 2.1 CLI scaffold, blob registry, and the core noun (D8, D9, D10)

### LINKED ISSUE #158: 2.2 Freeze the prototype CLI as a read-only reference (D15)

### LINKED ISSUE #159: 2.3 npm publish with provenance, engines floor, ADR 0018 amendment (D12, D13)

### LINKED ISSUE #160: 2.4 CLI session noun: start, ls, logs, resume, kill, send (D10)

### LINKED ISSUE #161: 2.5 CLI project, harness and events nouns — --json on every list (D10)

### LINKED ISSUE #162: 2.6 CLI core shell — the sanctioned escape hatch (D10)

### LINKED ISSUE #163: 2.7 CLI session attach — claims the Session write lock (D10, needs #144)

### LINKED ISSUE #164: 2.8 Parity check, then delete the frozen prototype (D15)

qcentic-dev:
## Summary

Phase 2 of #129 — the `actana` CLI and npm publishing — as it stands at the tip
of `feat/cli-and-publishing`. Eight squash merges, all reviewed and merged onto
the integration branch rather than onto the train. This PR is the phase's single
gate onto `beta/0.2.2`.

The shape of the phase: build the client half and the blob registry, wire
publishing before there is anything to publish, then add the nouns one at a time
— `session`, `project`/`harness`/`events`, `core shell`, `session attach` —
and finish by freezing the prototype CLI as a read-only reference, checking the
shipped tool against it line by line, and deleting it. The last of those is the
success test: the prototype is gone and the check that justifies its going is in
the repo.

`82 files changed, 16,497 insertions, 106 deletions` against `beta/0.2.2`.

## The rounds

| PR | Issue | What landed |
| --- | --- | --- |
| **#201** | #157 | `feat(cli)`: `packages/cli` — the `actana` command's client half, blob registry and the `core` noun (2.1, D8/D9/D10) |
| **#200** | #159 | `ci(release)`: npm publish with provenance, engines floor, and the ADR 0018 amendment (2.3, D12/D13) |
| **#204** | #160 | `feat(cli)`: `actana session` — start, ls, logs, resume, kill and send, with the Core owning prompt delivery (2.4, D10) |
| **#205** | #161 | `feat(cli)`: `actana project`, `harness` and `events` — `--json` on every list (2.5, D10) |
| **#203** | #162 | `feat(cli)`: `actana core shell` — the sanctioned escape hatch, over the existing PTY transport (2.6, D10) |
| **#206** | #163 | `feat(cli)`: `actana session attach` — a terminal on a Session, holding its write lock (2.7, D10, needed #144) |
| **#207** | #158 | `chore(docs)`: freeze the prototype CLI as a read-only reference under `docs/reference/` (2.2, D15) |
| **#209** | #164 | `chore(docs)`: parity-check `packages/cli` against the frozen prototype, then delete it (2.8, D15) |

**#201** created `packages/cli` with the `actana` binary, the registration-blob
registry and the `core` noun, and established the two rules the rest of the
phase inherited: a credential never reaches a log (`never-logs-a-blob.test.ts`,
extended by every later round that holds one) and no command escapes to the
local machine (`no-local-escape.test.ts`). **#200** wired npm publishing with
provenance *before* `packages/cli` had a publishable manifest — the publish set
is discovered rather than listed, so the CLI joined it with no further edit, and
until it did the rehearsal **reported its absence** rather than passing over it,
so "the CLI leg never ran" could not read as "the CLI leg passed". Both
provenance assertions were mutation-tested. **#204** put prompt delivery behind
the Core's ADR 0026 sequence rather than a CLI-side timer, which is what
`no-prompt-timing.test.ts` pins. **#205** answered `project inspect` and
`session inspect` with `--json` on every list verb instead of a second verb.
**#203** and **#206** are the two terminal-holding commands, both over the
existing PTY transport with all terminal handling CLI-side (D11) and the SDK's
existing vocabulary unextended.

**#207 and #209 are the pair that matters for the phase's honesty.** #207 froze
the prototype at `docs/reference/experiment-cli/`; #209 wrote the parity check
against that frozen tree and then deleted it — in that order, auditably: the
check commit `32d925c` (08:34:22) precedes the deletion `bfdc5d4` (08:34:58),
and the reference was still 17 files on disk when the check was written. The
check accounts for every prototype verb and flag: 6 resources, 20 verbs, 19 flag
rows and all 17 flat aliases, each with a destination in `packages/cli` or an
argued decision not to carry it. After the deletion, `experiment-cli` appears 5
times in 1,029 tracked files, all of them prose inside the check itself.

## Follow-up issues filed

Three, none of them blocking this gate:

- **#208** — `events-tail-cursor.test.ts` times out in CI under runner load.
  Evidence only: three occurrences across three branches, 61.7 s / 61.7 s /
  63.1 s against a 60 s budget, each green on attempt 2. Overshooting a deadline
  by one to three seconds is what a slow runner looks like, not a deadlock. No
  code is changed by the PR that filed it.
- **#210** — carry `project rm`. The one thing in the prototype that `actana`
  cannot do. Not carried because the Core's removal is `projectsMutate
  {op: "archive"}` and it **cascades** — every Task under the Project goes with
  it — and a destructive cascade behind a bare `--yes` is the shape that gets
  scripted with `--yes` already in the line. The capability is not lost: the
  frame is in `@actana/sdk` and the Panel uses it today.
- **#211** — carry `--model`. The prototype *spelled* `--model` for all four
  harnesses but the Core's allow-list only accepted it for `claude`, so it
  silently did nothing on three of four. The honest version needs a per-harness
  model table beside `HARNESS_LAUNCH_COMMANDS` and the Core's spawn policy
  extended in step; both halves land together or the flag is dishonest again.

#210 and #211 exist because the parity check argued both omissions in writing
and the #209 review asked for them explicitly: an argued omission with no ticket
decays into an undocumented gap once the phase closes.

## Honest deferrals, carried to the acceptance run

Three things this phase could not prove headlessly. None is a known defect;
each is a *feel* that needs a person at a real terminal, and each was named in
its own PR rather than discovered here.

1. **The full-screen TUI feel through `session attach` (#206).** Everything
   provable without a person is proved: 31 tests over the exit paths, the four
   authorities, the release asymmetry, takeover mid-session and the detach key;
   plus 4 live tests against a real `PtyCoreLinkServer` on a real `wss://` port
   with two attaches at once. What is not covered is a real harness's TUI
   painted into a real terminal — #206 says so up front, and #163 said so before
   it.
2. **The real login-shell feel in `actana core shell` (#203).** `core-shell.test.ts`
   drives the command against a *fake* terminal, and `core-shell-live.test.ts`
   uses the real channel against the real `PtyCoreLinkServer` over mTLS — so the
   transport and every exit path are proved, but no test sits a human in a real
   TTY. Whether it behaves like a login shell in hand (prompt rendering, resize
   under a real window manager, job control) is an acceptance-run observation.
3. **The three-harness live evidence from #191.** ADR 0026's delivery sequence
   is pinned by assertions that encode the traces observed during the
   experiment, not re-observed ones — #191 states plainly that it was not run
   against a live harness on that machine, and that a fresh Session on each
   supported harness from the PR image is the acceptance step it could not
   perform for itself. That step has not happened since, so it rides into
   `beta/0.2.2` still owed. Related: #177 records `cursor-cli` as undrivable
   from the CLI client, which is the kind of thing this run is for.

## Testing

CI is green on the branch tip — all checks pass or skip, zero failures, on the
`1ea2fbd` run. The CLI suite stood at **291 passed, 3 skipped** as of #206, and
#209 adds no code. `pnpm typecheck` clean repo-wide; `pnpm lint` 0 errors.

Two local-only failures are pre-existing and unrelated to this branch:
`scripts/__tests__/core-launcher.test.mjs` (a real `/opt/actana` install on the
dev machine) and the #208 flake above.

## Related issues

Closes #157
Closes #158
Closes #159
Closes #160
Closes #161
Closes #162
Closes #163
Closes #164

Refs #129 — the parent design doc; D8, D9, D10, D11, D12, D13 and D15 are the
decisions this phase implements.

Refs #208
Refs #210
Refs #211

## Review gate

**This PR is the user's review gate.** It is not to be merged, approved or
self-merged by anyone else: it is reviewed by the user and merged only after
their approval. Nothing here touches `main` — the merge into `beta/0.2.2` is the
whole point and it happens on the user's word.


qcentic-adm (approved):
## Phase-level review — the assembly, not the parts

Every ticket PR in this range (#201, #200, #204, #205, #203, #206, #207, #209) was
reviewed and approved on its own. This review asks the only question those eight
could not: **is the whole what the eight parts add up to, and does it hold together
as a phase?** Nothing below re-litigates a merged round.

**Verdict: approved.** The assembly is sound. Three observations, none of them an
assembly-level defect, are recorded at the bottom for the acceptance run.

---

### 1. The diff is the sum of the eight squashes — proven, not assumed

| Check | Result |
|---|---|
| `merge-base(beta/0.2.2, head)` | `cdfcb7a` — **identical to the tip of `beta/0.2.2`**, so the base is wholly contained and nothing on the train is missing from the branch |
| Commits ahead of base | exactly **8** |
| Commits on base not on head | **0** — no divergence |
| Parent chain | `cdfcb7a → ed3837b → db947a5 → 2208bdc → 57c3109 → e422f7a → 3e32673 → 8c3d5b2 → 1ea2fbd`, every commit single-parent. No merge commits, no rebase artefacts |

Each SHA is byte-identical to the merge commit GitHub recorded for its ticket PR,
and every one of the eight was based on `feat/cli-and-publishing` and squash-merged:

`#201`→`ed3837b` · `#200`→`db947a5` · `#204`→`2208bdc` · `#205`→`57c3109` ·
`#203`→`e422f7a` · `#206`→`3e32673` · `#207`→`8c3d5b2` · `#209`→`1ea2fbd`

**No stray commits.** The file set GitHub reports for each PR matches the file set of
its squash commit exactly, in all eight cases — 36 / 19 / 16 / 25 / 14 / 13 / 17 / 18.
No squash carried a file its PR did not.

**The arithmetic closes.** Summing the eight squashes: 19,672 insertions and 3,281
deletions — a net of **16,391**. The PR's own diff against `beta/0.2.2` is 16,497 −
106 = **16,391**. The same number, from two directions.

The eight squashes touch 100 distinct paths; the assembled diff shows 82. The
18-path difference is fully accounted for and contains no surprise: the 17 files of
`docs/reference/experiment-cli/` (created by #207, deleted by #209) and
`packages/cli/src/__tests__/in-process-core-harness.ts` (added by #204, superseded in
#205). **No path in the final diff is absent from some squash** — the set difference
in that direction is empty.

**No history rewrite.** `beta/0.2.2` has a single reflog entry and has not moved.
Every observed update to `feat/cli-and-publishing` is a fast-forward; the ref never
rewound. Consistent with ADR 0023 D24 — trains and integration branches are never
force-pushed.

### 2. #149's own acceptance criteria

The epic's *Done means* is three clauses:

- **`npm i -g @actana/cli` gives a working `actana` against a remote Core.**
  `@actana/cli` is non-private, `bin: {actana: "bin/actana.mjs"}`, `files: [dist, bin]`,
  `engines: ">=22"`, `publishConfig` public + provenance. The publish path asserts the
  parts an install actually depends on: `assertPackedFiles` refuses a `bin` target that
  is mapped but not packed (npm links it without checking, which is how a command
  becomes a dangling symlink), and `assertPackedBin` reads the first bytes for a
  `node` shebang. The *in-hand* install is the acceptance run's job by construction —
  everything provable ahead of a real registry is proved.
- **Every list command emits `--json`.** One implementation seam (`cli-output.ts`),
  the flag parsed position-independently, and it is on `core ls` / `core status`,
  `project ls`, `harness ls`, `events tail` (NDJSON) and `session ls`. `core shell`
  is the honest exception and *says so* rather than accepting the flag silently,
  pointing at `core status --json` instead. D10 met.
- **The frozen prototype is gone.** `docs/reference/` holds one file,
  `cli-parity-check.md`. `experiment-cli` occurs 5 times repo-wide, all prose inside
  that check. No import, no path reference, no build input.

The epic's per-ticket rules also hold: all eight based on `feat/cli-and-publishing`,
all squash-merged, and this is the single closing PR. On **"no docs-only PRs"** —
#207 and #209 are docs-only in file terms, and that is the epic working as written
rather than against it: #129 D15 makes the freeze and the deletion explicit tickets
of this phase, and #149's own step table lists both. The rule guards against ADRs
landing away from their code; #207/#209 are not that case.

### 3. Phase invariants across the assembled whole

- **`bin` command is `actana`** — and it is pinned by assertion, not by convention:
  `no-local-escape.test.ts` asserts the package name is `@actana/cli`, that
  `Object.keys(bin)` is exactly `["actana"]`, and that it points at `bin/actana.mjs`.
- **`@actana/cli` publishable in lockstep with `@actana/sdk`** — both at `0.2.2`, both
  non-private; the other four manifests carry `private: true`. The set is *discovered*
  and checked against intent in both directions, so a package quietly losing
  `private: true` fails and a publishable package quietly regaining it also fails.
  `assertPackedManifest` rejects a `workspace:` range that reached the tarball and any
  `@actana/sdk` range that is not the CLI's own version. `publishOrder` topologically
  sorts the SDK ahead of the CLI — alphabetical would have published the CLI first and
  burned its version against an SDK that was not yet on the registry.
- **Exactly one protocol definition, in the SDK** — `CORE_LINK_PROTOCOL_VERSION` is
  declared once, at `packages/sdk/src/core-link-frames.ts:1257`. The only other
  `*_PROTOCOL_VERSION` is `PANEL_LINK_PROTOCOL_VERSION`, a different wire. The
  prototype's hand-typed 90-line mirror is deleted. The parity check's §5 reaches the
  same conclusion independently and carries the one honest caveat rather than hiding
  it: the *registration blob* shape is declared twice (SDK and `shared`), which is
  ADR 0025 D4 working as designed — a published package cannot import a type from a
  private one, and a blob is not a core-link frame.
- **No local-process escape in the shipped CLI** — no shipped module imports
  `child_process` or `worker_threads` under either specifier spelling, and the sweep
  is recursive with a guard-on-the-guard (it fails if the file list is empty), plus
  checks for `process.binding` and concatenated `require()` specifiers. Every `spawn`
  in `packages/cli` is the *remote* PTY spawn over the core link. It further pins that
  nothing imports `@actana/core` or `@actana/shared`, and that runtime deps are exactly
  `@actana/sdk` and `ws`.
- **`docs/reference/experiment-cli` gone** — confirmed above.
- **D11, terminal handling in the CLI only** — `setRawMode` / `isTTY` / `SIGWINCH`
  appear only in `packages/cli` (`cli-terminal.ts`, `core-shell.ts`,
  `session-attach.ts`, `actana-cli-entry.ts`). `packages/sdk/src/core-session.ts`
  states the rule in its header and keeps it; `terminal-screen.ts` is the headless
  screen model D11 names as `session.screen()` and touches no TTY. The one
  `process.stdin.isTTY` outside the CLI is in `packages/core` — the Core's own
  separate binary, pre-existing and out of this phase's reach.

### 4. The manifest cut: six, and consistent

`packages/cli` joining makes it six, and the number is written in four places that all
agree:

- `ci.yml` `Train rules` — six-element `MANIFESTS`, and it asserts the list **covers
  every workspace package under `packages/`** before it compares any version, so a
  seventh fails by name rather than being skipped by a loop reporting green over a
  shrinking set
- `promote.yml`'s cut — the same six in `files=()`
- `scripts/__tests__/workflows.test.mjs` — six, asserted against a directory listing
  rather than a count
- ADR 0023 D3 — amended by #157 as the sixth, with #152's amendment left in place; the
  ADR explicitly demotes its own number to documentation and names `ci.yml` as the
  mechanism

`docs/beta-acceptance-checklist.md` §6 names all six. **`Train rules` passes** (green,
6s). All 19 checks pass, 3 skip, zero failures on the `1ea2fbd` run.

### 5. Honest deferrals this train carries into the acceptance run

The three the description names, plus what a phase-level sweep adds:

1. **Full-screen TUI feel through `session attach`** (#206, flagged in #163 before it).
2. **Real login-shell feel in `actana core shell`** (#203/#162) — transport and exit
   paths proved against a real `PtyCoreLinkServer`; no human in a real TTY.
3. **The three-harness live evidence from #191** — still owed; #191 said so itself.
   Related: **#177** is open and records `cursor-cli` as undrivable from the CLI client.
4. **#208** — `events-tail-cursor.test.ts` flake, evidence-only, green on retry.
5. **#210** — `project rm` not carried, argued.
6. **#211** — `--model` not carried, argued.
7. **#193 — filed, open, and not referenced anywhere in this train.** See below.
8. Two local-only failures noted in the description: `scripts/__tests__/core-launcher.test.mjs`
   (a real `/opt/actana` on the dev machine) and the #208 flake.
9. **`CHANGELOG.md` is untouched by this PR** and byte-identical to the base — still
   headed `0.1.0 — unreleased`. That is correct policy here (CONTRIBUTING builds it
   from squash subjects at promotion), but `beta-acceptance-checklist` §5 is a gate on
   `beta/0.2.2 → main`, so it is work this train hands forward rather than work it did.

### 6. Follow-up ledger — three of four referenced

`#208`, `#210`, `#211` are all filed (2026-08-13), open, and referenced in the
description. `#210` and `#211` exist because the parity check argued both omissions in
prose and #209's review asked for tickets — that is the right instinct and it was
followed.

**`#193` is the gap.** It is filed and open (2026-08-12,
`sanitizeInitialInput` drops newlines without a separator), but it is referenced by
neither this PR nor any of the eight ticket PRs, nor anywhere in the tree. It is a
pre-existing `packages/core` defect deferred out of #191, so it is not this train's to
fix — but this train is what makes it *reachable by a person*: `actana session start
<project> [prompt]` and `session resume` are new front doors onto that same sanitiser,
and `session --help` actively teaches the multi-line case (`cat brief.md | actana
session start web -`). A prompt with a blank line comes out with words welded together
(`diff` + `then` → `diffthen`). Worth a line in the acceptance notes even though the
code is untouched here.

### Observations — non-blocking, cross-round staleness

Neither affects behaviour; both are the specific shape a phase review exists to catch,
where each round was right on its own and a later round moved the world an earlier
round's prose described.

1. **`ROOT_HELP` omits `attach`.** `packages/cli/src/actana-cli.ts:72` still reads
   `session   start, ls, logs, resume, kill and send to Sessions on one`. #206 added
   the verb, dispatched it (`session-command.ts:154`) and listed it in `SESSION_HELP`
   (line 84) — but not in the root summary. `actana --help` therefore under-reports the
   command surface by one verb. User-visible, cosmetic.
2. **The `RESERVED_NOUNS` docblock is stale.** It closes with *"The reservations this
   build still carries are at the verb level: `project cp` / `project files` (#168…)
   and `core shell` (#162)."* That was true when #205 (`57c3109`) wrote it; #203
   (`e422f7a`) landed `core shell` immediately after — it is dispatched at
   `core-command.ts:100` and documented in `CORE_HELP`. Internal comment only.

Both are a one-line fix each and neither is grounds to hold the gate.

---

**Approved as an assembly.** The eight squashes reconcile exactly against the diff by
file set, by line arithmetic and by commit identity; the base contains precisely what
was reviewed; #149's three *Done means* clauses are met; every phase invariant holds
across the assembled whole and most are machine-enforced rather than asserted in prose;
the manifest set is six and consistent in all four places that encode it, with
`Train rules` green.

Per the review gate in the description: **this is a review, not a merge.** The merge
into `beta/0.2.2` is the user's, on their word.

mrfqcentic (approved):
Reviewed against the experiment PoC source, the eight tickets (#157–#164), ADRs 0018/0022/0024/0026, and the branch tip at `1ea2fbd`. CI on the tip is green — all checks pass or skip, zero failures. I re-derived the prototype's command surface from its source rather than trusting the parity doc, and checked each claimed destination in `packages/cli` by reading the code, not the test names.

**Verdict: approve.** The reasoning, and three documentation gaps I want amended or ticketed before the phase is called closed, are below.

## What I verified

**The npm package is real, and publishing is right (#157 packaging, #159)**
- `@actana/cli` is an installable manifest: `bin: actana → bin/actana.mjs`, `files: [dist, bin]`, `engines: ">=22"`, `publishConfig: { access: public, provenance: true }`, no install lifecycle scripts. The esbuild bundle targets node22; the shim exits 70 with a build hint when `dist/` is absent, and `bin-shim.test.ts` pins that.
- `actana daemon` is refused in the CLI (`actana-cli.ts:134-139`, pinned by test and by the import sweep) and stays in the Core package. One binary name, split by noun — holds.
- `release.yml`: the npm job runs last (`needs: resolve, tarball, tarball-macos, installer-e2e, panel, core`), on the tag, with `--provenance --access public`, plus a post-publish attestation read-back that splits unattested from unverified. A missing `NPM_TOKEN` fails loudly in `resolve`, before anything builds — same shape as the Docker Hub check.
- The publish set is discovered and cross-checked, not listed; the rehearsal reports an absent CLI rather than passing over it; the provenance assertions are mutation-tested; `require-node-24.mjs` absence is asserted on real packed bytes; ADR 0018 is genuinely amended; `packages/cli` is the sixth manifest in both the cut and the manifests-match check, with the workspace set pinned equal by `workflows.test.mjs`.

**The quality gates are real tests, not theater**
- `never-logs-a-blob` sweeps sentinel cert/key/bearer across every verb including failure paths under `--verbose`, with a positive counterpart so it cannot pass by printing nothing.
- `no-local-escape` asserts capability absence — no process-starter imports anywhere in shipped source, comment-stripped, with a guard on the guard. No `docker exec` anywhere in `packages/cli/src`.
- `no-prompt-timing` sweeps the source for CLI-side timers with a single argued allowlist row. `send` writes verbatim; the carriage return is its own write only under `--enter`.
- Credential resolution is exactly `--core` → `ACTANA_CORE_BLOB` → `current` pointer, each failure naming its source. Blobs live at `$XDG_CONFIG_HOME/actana/cores/<name>.txt` (relative XDG ignored, spec-correct), written 0600 inside a 0700 dir, insecure modes reported rather than silently repaired. `core add` takes a file, `-`, or a pipe through one decoder.

**The nouns behave as ticketed (#160–#163)**
- `session start` is one-shot by default (taskId on stdout, exits); `--wait` blocks on the SDK; `--wait --json` emits one object. `--json` purity holds: stdout carries JSON only, progress and warnings go to stderr, and the failure path still emits `{"error"}` on stdout. `logs` renders through the SDK's `TerminalScreen` with `--raw` as opt-out.
- `project browse` walks the Core's disk via `dirList`; `project add` never resolves a local path. The events cursor is durable (synchronous write per event) and the live test restarts the Core mid-stream and asserts the exact id sequence and the on-disk cursor — a real reconnect story, neither repeat nor gap. `harness install` waits on the Core's verdict, exits 0 only on `available`, and names the harness and links #31/#128 on failure.
- `core shell` and `session attach`: raw mode is restored on every exit path — `finally` before socket close, SIGINT/SIGTERM routed so the `finally` is reached, an idempotent never-throwing guard, and a process-exit hook as last resort — with a test per path. Ctrl-C goes to the remote process as 0x03, resize propagates via SIGWINCH, remote exit status comes back with the 128+signal convention. `attach` claims the write lock before painting; a refused claim opens read-only with a visible reason and never force-takes; detach and connection drop both release. **`packages/sdk/src` has an empty diff** — D11 holds.

**Parity vs the experiment, independently re-derived (#158/#164)**
- From the PoC source I count 6 resources, 20 verbs, the full flag surface, and 17 flat aliases. The check accounts for all of them, and every destination it cites exists where cited. The two argued omissions are genuinely absent — `project rm` is an unknown verb, `--model` an unknown flag — and both are ticketed (#210, #211) with the cascade and the allow-list reasoning on record.
- The frozen tree is gone; the five remaining `experiment-cli` mentions are prose inside the check itself. The freeze commit `8c3d5b2` carried 17 files — none of them `blob.txt`, `blobs/`, logs, or pid files — and a content scan of that tree finds prose about credentials, never credentials. History, not just the working tree, is clean.

## Three gaps in the parity check's accounting — amend or ticket, none blocking

None of these is a code defect, and none changes my approval. But #164's own trap says behaviours belong in the check as "moved to X" or an argued decision, and this phase's standing rule is that an unargued delta decays into an undocumented gap once the phase closes:

1. **`--kind` narrowed from prefix to exact, unargued.** The prototype matched `kind.startsWith(only)` and its help documented `--kind <prefix>`; the shipped CLI exact-matches a set (`events-command.ts:196`). The parity rows (lines 201, 242) map the flag without noting the semantic change, so `--kind task:` from the old manual now silently matches nothing. Exact-match-and-repeatable is arguably the better design — it just needs its sentence in the check, or a ticket.
2. **No env-var or config-path section.** `AC_BLOB`, `AC_BLOB_FILE`, `AC_CORE`, `AC_CONTAINER` and `AC_DEBUG` have no rows. Only `ACTANA_CORE_BLOB` survives — which #157 spec'd, so the rename is intended, merely unrecorded — and `AC_DEBUG`-style frame tracing has no equivalent (`--verbose` is prose-level). The `blob.txt`/`blobs/` → `$XDG_CONFIG_HOME/actana/cores` relocation is likewise unmentioned.
3. **The `--now` / `send` settle-wait attribution is imprecise.** The check says "moved to Core — `harness-prompt-delivery.ts`", but that module gates only the spawn-time prompt (`pty-manager.ts:617-645`, the `initialInput` path); mid-session writes go straight to the PTY (`pty-manager.ts:707-711`). The CLI doing no timing is correct under #160's trap — what needs a line is that an ad-hoc `send` to a *busy* TUI no longer waits for settle anywhere in the system. Argued delta or a Core-side follow-up: either is fine, silence is the one option that isn't.

Two cosmetic asides: the CLI ships no `.d.ts` by documented design (the library-vs-command split in `npm-packages.mjs`), while #159's wording reads "compiled JS plus `.d.ts`" for published packages — worth a one-line clarification somewhere durable; and the `npm-packages.mjs:20` comment describes a Core/Panel engines floor that actually lives only in the root manifest.

## On the deferrals

The three honest deferrals (TUI feel through `attach`, login-shell feel, the owed three-harness live evidence from #191) are each named up front and traced to their PRs, and each is a person-at-a-real-terminal observation by its nature. Agreed they belong to the acceptance run and do not gate this merge. #208 is filed with evidence and changes no code.

Approving: the npm package is real and publishes with provenance in lockstep with the images; the CLI reproduces the experiment's command surface with two argued, ticketed omissions; and #157–#164's deliverables are met in code, with the items above left as documentation-completeness fixes to the phase's audit artifact. Merge remains yours, per the gate.

Issue #166, opened 2026-08-11
3.2 Listing on the HTTP surface, not dirList (F7, F10)


mrfqcentic:
**Base branch: `feat/project-files`, not `main`.** Train: **`beta/0.3.0`**.
Worktree: **`wt-129-16`**. Suggested branch: **`feat/files-listing`**.
Phase 3, step 2. Implements **F7**, and the manifest half of **F10**.

## What

Listing a Project's files, on the **HTTP surface**, not on the core link's `dirList`.

## Why not `dirList`

Extending `dirList` moves `CORE_LINK_PROTOCOL_VERSION`, which marks **every Core in every fleet** needs-update under ADR 0005 — for a feature most of them will never be asked for. A separately-versioned HTTP surface degrades gracefully instead: a Core without it simply does not announce `files` on `ready`.

## Done when

- A Project's tree can be listed to arbitrary depth, streamed as NDJSON, with `{path, size, mtime, mode, sha256}` per entry.
- The same confinement rules as the routes ticket apply, tested independently — not inherited on trust.
- A Core without the capability is not asked; the client checks `ready` first.
- Listing a large tree does not buffer it in memory on either side.

## Traps

- `sha256` on every entry of a large tree is expensive. Decide deliberately whether it is computed eagerly or on request, and record the choice — F10's future diff endpoint depends on it being available, not on it being free.
- This is the surface a future one-way sync (phase 4, F10) reads. Sync itself is **reserved, not built**; do not build toward it beyond the manifest fields.

Refs #129

qcentic-dev:
Done in PR #216, merged into `feat/project-files` — a non-default base, so this did not auto-close; closing by hand.

Issue #150, opened 2026-08-11
#129 phase 3 — project files (train beta/0.3.0)


mrfqcentic:
**Train: `beta/0.3.0`.** Integration branch: **`feat/project-files`**, cut from `beta/0.3.0` after `0.2.2` promotes — never earlier.

Phase 3 of #129: files cross between an operator's machine and a Project on a Core.

## Why a minor, when the others were patches

This is the first phase that adds a **new public surface**: mTLS HTTPS routes on the Core, `project.files.*` on the SDK, `project cp` on the CLI. `CORE_LINK_PROTOCOL_VERSION` deliberately does not move (F9) and no Core becomes needs-update — but a new externally-consumable surface is a minor under ADR 0023 D31, and this train completes #129.

## Why it cannot start early

#129's own rule: the branch is cut only after phase 2 has landed, because every surface phase 3 consumes is undefined until then. There is no partial start.

## Steps

| Step | Parallel | Tickets |
| --- | --- | --- |
| 1 | 1 | Core HTTPS routes: upload/download, confinement, hardened tar, capability field |
| 2 | 2 | listing surface ∥ SDK `project.files.*` |
| 3 | 2 | CLI `project cp` / `project files` ∥ Panel drag-drop + volume view |

## Done means

A file dropped on a Project in the Panel is `cat`-able by a harness seconds later, and an agent's report comes back with one `actana project cp`.

## Rules for every ticket here

- Base branch is `feat/project-files`.
- One worktree per ticket, on the Dev Core.
- Squash-merge; the three phase-3 ADRs ride the routes ticket. **No docs-only PRs.**
- At the end of the phase, one PR: `feat/project-files` → `beta/0.3.0`.

Refs #129

qcentic-adm:
Delivered — closing. Phase 3 landed on its train and shipped.

**The train:** `beta/0.3.0`, integration branch `feat/project-files`, cut after `0.2.2` promoted (PR #213, 2026-08-13) exactly as this ticket required. One PR closed the phase — **PR #223** *"feat(files): project files — HTTPS routes, listing, SDK, CLI cp, Panel drag-drop (phase 3)"*, merged 2026-08-14T08:50Z.
**Released in:** **v0.3.0** (promotion PR #234, published 2026-08-14T09:44Z) — a minor, as this ticket argued for under ADR 0023 D31.

**Every step's tickets are closed, and I checked the surface each one claims:**

| Step | Ticket | PR | Verified in the tree |
|---|---|---|---|
[...]

PR #215, opened 2026-08-13
feat(core): /v1 file routes over HTTPS — hardened tar, confinement, NDJSON progress, ready capability

### LINKED ISSUE #165: 3.1 Core files routes: tar, confinement, NDJSON progress, ready capability (F2-F6, F8, F9)
Worktree: **`wt-129-15`**. Suggested branch: **`feat/core-files-routes`**.
Phase 3, step 1, alone — everything else in the phase consumes this. Implements **F2**–**F6**, **F8**, **F9**.

## What

New `/v1/…` routes on the mTLS HTTPS server the core-link WebSocket already mounts on. Bytes cross over HTTPS, never the core link (F2): a multi-gigabyte upload chunked into JSON frames would stutter every terminal pane through head-of-line blocking, and base64 costs 33%.

- **Upload and download, both directions** (F6). `GET` returns a file's raw bytes, or a directory as tar.
- **A folder crosses as one streamed tar** (F4), with a chunked-NDJSON per-entry progress response. Per-file requests make a `node_modules` latency-bound; tar makes it bandwidth-bound and carries the executable bit natively.
- **Confinement** (F3): reject absolute paths, escaping `..`, and symlinks resolving outside the Project root.
- **Overwrite by default, every overwrite named** in the NDJSON stream — `written` / `overwritten` (F5).
- **No size cap; free-space precheck; one write transfer per Project at a time** (F8) — an immediate "transfer in progress" error, not a queue. Reads are unrestricted and concurrent.
- **Capability on `ready`**: optional `files?: { version: 1 }` (F9). `CORE_LINK_PROTOCOL_VERSION` **does not move**; no Core becomes needs-update.

## The three phase-3 ADRs ride this PR

F1 (the filesystem is the model — no index, no ids), F2 (HTTPS not the core link), F4 (tar, and why not per-file). **No docs-only PR.**

## Done when

- A file and a folder both round-trip, and the folder keeps its executable bits.
- Every confinement case is tested: absolute, `..`, and a symlink pointing out.
- A second concurrent write to the same Project is refused immediately with a distinguishable error.
- An older Panel talking to a Core with this capability, and a newer Panel talking to a Core without it, both behave sanely.

## Traps

- **Tar unpack is the attack surface.** Entry paths must be validated after resolution, not before; `..` inside a tar entry is the classic escape. Reject absolute entries, device nodes and hardlinks outside the root.
- **F3 is an accident guard, not a security boundary** — `core shell` is the sanctioned escape hatch. Do not design as if it were a sandbox, and do not claim it is one in the ADR.
- **The Panel is a dumb pipe** (F11) and validates nothing. One machine validates paths: the one that owns the disk. If it is convenient to validate in the Panel, that is the bug.
- NDJSON entries must carry `{path, size, mtime, mode, sha256}` from the first commit (F10). It is cheap now and it is what a future diff endpoint falls out of; retrofitting means a second pass over every byte.

Refs #129

qcentic-dev:
## Summary

Phase 3 step 1 of #129: the Core's `/v1/…` file routes, on the **same mTLS HTTPS server the core-link WebSocket is already mounted on**. One port, one certificate, one bearer, two protocols — and **no file byte crosses the core link** (F2), because chunking a multi-gigabyte upload into JSON frames would stutter every terminal pane sharing that socket through head-of-line blocking, and base64 costs a third of the wire.

`GET` returns a file's raw bytes or a directory as one streamed tar. `PUT` writes, answering with a chunked NDJSON progress stream carrying `{path, size, mtime, mode, sha256}` per entry and naming every overwrite. Overwrite is the default; one write transfer per Project, refused immediately rather than queued; reads unrestricted and concurrent; no size cap, but a free-space precheck.

The three phase-3 ADRs ride this PR — 0027 (the filesystem is the model), 0028 (bytes cross HTTPS, not the core link), 0029 (a folder crosses as one streamed tar). No docs-only PR.

`CORE_LINK_PROTOCOL_VERSION` **does not move**. The surface announces itself as an optional `files: { version: 1 }` capability on `ready` (F9, ADR 0024 D11), so no Core in any fleet becomes needs-update.

### Two things a reviewer should read closely

**F3 confinement is an accident guard, not a security boundary**, and nothing here claims otherwise — not the ADR, not the code, not the test headers. `core shell` is the sanctioned way onto that machine's disk and sits right next to this. The refusal is a `400` rather than a `403` for the same reason: a 403 would claim a permission model that does not exist. ADR 0027 D5 says this explicitly, because the failure mode is somebody reading "confinement" in a filename two years from now and building something load-bearing on the conclusion.

**Tar unpack is the attack surface.** Entry paths are validated *after* resolution, never before — including through a symlink an earlier entry of the same archive just planted, which is the two-entry escape a one-pass check over the entry list cannot see. Pax and GNU long-name overrides are validated too, because they are the second place a path can come from. Absolute entries, device nodes, fifos and links resolving outside the root are refused by name. A separate defect found while writing the tests: the final path component is never followed, so an entry named `notes.txt` replaces `notes.txt` rather than writing through a `notes.txt` symlink into someone else's file.

## Related issues

Closes #165
Refs #129, #150

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## How was this tested?

**131 new tests across six suites.** Full run green: sdk 182, shared 180, cli 297, core 964, panel 1338. `pnpm typecheck`, `pnpm lint` (0 errors) and `pnpm scan:secrets` clean; commitlint passes on every commit in the branch.

| Suite | Covers |
| --- | --- |
| `core-files-mtls.test.ts` | the end-to-end claim, with **nothing injected** — the Core's own default server factory, real certificates from `generateCertMaterial`, a real signed bearer. A WebSocket upgrade and a `/v1` request on one port; a file and a folder round-tripping; executable bits surviving both directions; no client cert refused at the handshake; cert-but-no-bearer refused 401; a second concurrent write refused 409 over the real transport |
| `files-tar-unpack-hardening.test.ts` | archives built **by hand**, because every case is a shape `tar(1)` will not emit: `..`, absolute entries, symlink and hardlink escapes, the two-entry planted-symlink attack, device nodes and fifos, escaping pax and GNU long-name overrides, damaged and truncated streams. Each asserts nothing was written outside the root |
| `files-tar-round-trip.test.ts` | round-trip incl. executable bits on a file *and* the folder holding it, long/non-ASCII names, symlinks, deterministic bytes, and interop with `tar(1)` in **both** directions |
| `files-confinement.test.ts` | the three cases the ticket names, plus the ones that make the third real — innocent-looking segments under a symlinked parent, a Project whose own root runs through a symlink |
| `core-files-routes.test.ts` | the routes over a real HTTP server: reads, writes, NDJSON progress, `written`/`overwritten`, the 409, the 507, the bearer gate, concurrency |
| `core-link-ready-files-capability.test.ts`, `files-capability.test.ts` (SDK) | the capability, and **both mismatch directions**: older Panel/newer Core and newer Panel/older Core |

`experiment/` and the root `scripts/__tests__/core-launcher.test.mjs` were not touched; that suite's one red test is a pre-existing local-environment failure (a real `/opt/actana` install on the dev machine), not a regression from this branch.

None. `CORE_LINK_PROTOCOL_VERSION` stays at `0.15.0` and no Core becomes needs-update. The one SDK-visible change is additive: `CoreConnectionInfo` gains `files`, and `CoreClient` gains `canUseFileRoutes()` / `filesCapability()`.

## Checklist

- [x] This PR targets the phase-3 integration branch `feat/project-files`, per #150 — not `main` and not `beta/0.3.0`
- [x] I updated documentation — three ADRs and `docs/external-api.md`, which had said the hook receiver was the Core's only HTTP surface
- [x] This PR is a draft; steps 2 and 3 of the phase consume it

## Scope

Core-side only. `packages/panel` and `packages/cli` are untouched. The SDK change is the minimum the capability forces — the `files` field, its reader, and the two accessors. `project.files.*` is #166, the listing surface is #166, and `project cp` is #168.


qcentic-dev:
## Acceptance criteria for #165, each checked or explicitly not

### "Done when" — the four the ticket names

- [x] **A file and a folder both round-trip, and the folder keeps its executable bits.** `core-files-mtls.test.ts` does both over the real mTLS transport with nothing injected: a file uploaded and read back byte-identical, and a folder uploaded as one tar, downloaded as one tar, and compared tree-for-tree against the source. Bits asserted on a file (`0o755`), on a non-executable sibling (`0o644`), and on the **directory** holding them (`0o750`) — outbound and inbound. `files-tar-round-trip.test.ts` adds the case that silently regresses: the mode after an *overwrite* of an existing non-executable file, where `open(…, mode)` applies nothing.

- [x] **Every confinement case is tested: absolute, `..`, and a symlink pointing out.** All three at the confinement layer (`files-confinement.test.ts`), all three at the route (`core-files-routes.test.ts`), all three over real mTLS (`core-files-mtls.test.ts`), and all three inside a tar entry (`files-tar-unpack-hardening.test.ts`). Each asserts the refusal *and* that nothing landed outside.

- [x] **A second concurrent write to the same Project is refused IMMEDIATELY with a distinguishable error.** `409` + `code: "transfer-in-progress"`, distinguishable by status from a bad path (400) and a missing Project (404) and by code from any other 409. Immediate is asserted twice: refused before the body is read, and — with `Expect: 100-continue` — before the client sends a byte. Also asserted: reads served during a write, a different Project served concurrently, the lease released after success *and* after a mid-stream failure. Exercised over the real transport with a transfer genuinely in flight.

- [x] **Both mismatch directions behave sanely.** *Older Panel, newer Core*: connects, compatible, driven exactly as today by a client that never reads the field — and the `ready` frame is asserted byte-identical apart from the one added key. *Newer Panel, older Core*: reads the absence, `canUseFileRoutes()` is false, the Core is compatible and fully usable, and no frame is gated because the file surface adds none. A Core announcing a version this build has never seen reads as absent, never as a guessed superset.

### The "What" bullets

- [x] **F2 — new `/v1/…` routes on the mTLS HTTPS server the core-link WebSocket already mounts on.** One `https.Server`, one port, one certificate, one bearer. Asserted directly: a WebSocket upgrade and a `/v1` request answered on the same port in one test.
- [x] **F2 — bytes cross HTTPS, NEVER the core link.** No frame added to the core link by this PR; the frame set is unchanged apart from one optional field on `ready`.
- [x] **F6 — upload and download, both directions.** `GET`, `HEAD`, `PUT`.
- [x] **F4 — a folder crosses as ONE streamed tar.** Streamed on both sides; neither ever holds the archive.
- [ ] **F4 — per-entry NDJSON progress.** ✅ on the **write** direction, which is where F5 puts it and where every entry line lives. ❌ on the **read** direction, and this is a deliberate deviation rather than an oversight: a folder download's body *is* the tar, so a manifest would need either buffering the archive to hash it first (a temp copy of a `node_modules`) or appending a synthetic entry that `tar -xf` would extract into the operator's directory. The tar's own headers already carry four of F10's five fields natively, and the fifth is a digest of bytes the reader is receiving, so the reader computes it in the same single pass the writer would have — same answer, still one pass. Recorded and argued in **ADR 0028 D5** rather than left to be discovered.
- [x] **F3 — confinement: reject absolute paths, escaping `..`, symlinks resolving outside the Project root.** And validated *after* resolution, which is what makes the third one real.
- [x] **F5 — overwrite by default, every overwrite named `written` / `overwritten` in the chunked NDJSON progress stream.** Both values asserted in one archive containing one of each, over the real transport.
- [x] **F8 — no size cap.** None exists. The only check is a fit check.
- [x] **F8 — free-space precheck.** Declared `Content-Length` against `statfs`, `507` + `insufficient-storage`. Valid in both modes because a tar is never smaller than the files inside it. A disk that cannot be measured **proceeds** — `null` is "do not know", never "full" — and a chunked upload declares nothing to check; both asserted.
- [x] **F8 — one write per Project, an immediate error and not a queue; reads unrestricted and concurrent.**
- [x] **F9 — optional `files: { version: 1 }` on `ready`.** Announced when and only when the routes are actually mounted; announcing one whose routes are not answering is worse than announcing nothing.
- [x] **F9 — `CORE_LINK_PROTOCOL_VERSION` does NOT move; no Core becomes needs-update.** Still `0.15.0`, pinned by a test whose *name* carries the reason so a future bump fails with an explanation rather than a number.

### The three ADRs ride this PR

- [x] **F1 — the filesystem is the model, no index, no ids** → `docs/adr/0027-the-filesystem-is-the-model.md`
- [x] **F2 — HTTPS, not the core link** → `docs/adr/0028-file-bytes-cross-https-not-the-core-link.md`
- [x] **F4 — tar, and why not per-file** → `docs/adr/0029-a-folder-crosses-as-one-streamed-tar.md`
- [x] **No docs-only PR.** All three land in this PR, alongside the code they describe.

### Traps

- [x] **Tar unpack: validate entry paths AFTER resolution, not before.** Per entry, at the moment it is written, through the symlinks that exist *then* — which is the only way the two-entry escape (entry 1 plants a symlink, entry 2 writes through it) is visible. Tested as its own case. The pax and GNU long-name overrides are validated too, because they are the second place a path can come from; an implementation that checked the ustar name and let pax replace it would pass every other test in the file.
- [x] **Reject absolute entries, device nodes, and hardlinks pointing outside the root.** Each with its own refusal code. Device nodes and fifos are refused on unpack (an archive asking this Core to *create* one) while being skipped on pack (the operator's folder happens to contain one) — the asymmetry is deliberate and commented.
- [x] **F3 is an accident guard NOT a security boundary — do not design it as a sandbox and do not claim it is one in the ADR.** ADR 0027 D5 states the denial outright and names `core shell` as the sanctioned escape hatch that makes any containment claim false. The same denial is repeated in `files-confinement.ts`'s header and in its test suite's header, because a file with "confinement" in the name is read far more often than an ADR is. The refusal is `400` and not `403` for the same reason. An alternative considered and rejected in writing: "design it as a sandbox and say so."
- [x] **The Panel is a dumb pipe (F11) and validates nothing; exactly ONE machine validates paths — the one that owns the disk.** `packages/panel` is untouched by this PR. ADR 0027 D3 makes a Panel-side path check a citable review finding rather than a judgement call.
- [x] **NDJSON entries carry `path`, `size`, `mtime`, `mode`, `sha256` from the FIRST commit (F10).** All five, on every entry, from the commit that introduced the tar codec. `sha256` is computed during the pass the transfer is already making; `null` only for a directory, which has no bytes, and a symlink hashes its target string — which is what a diff would compare.

### Boundaries

- [x] **Core-side work; `packages/panel` and `packages/cli` untouched.** Confirmed by the diff.
- [x] **SDK change kept minimal.** The `files` field on `ready`, its reader, its pass-through in the transport, and two accessors. No `project.files.*` — that is #166.
- [x] **Never force push.** Not used. The base has not moved; if it does, `origin/feat/project-files` gets merged in and plain-pushed.
- [x] **Nothing merged, `main` untouched, PR left as a draft.** Base is `feat/project-files`.

### Gates run locally

- [x] `pnpm typecheck` — clean across all five packages
- [x] `pnpm lint` — **0 errors** (51 warnings, all pre-existing; none in the new files)
- [x] `pnpm scan:secrets` — "No obvious secrets found."
- [x] `pnpm -r test` — sdk 182, shared 180, cli 297, core 964, panel 1338, all green
- [x] commitlint over every commit in the branch — 0 problems, 0 warnings
- [ ] **The root `scripts/` suite is not green on this machine** — `core-launcher.test.mjs` has one red test because a real `/opt/actana` install on the dev box shadows the temp install root it asserts against. Pre-existing and environment-specific, not a regression from this branch; CI is the authority on it.

qcentic-dev:
## Final acceptance checklist for #165 — every criterion, checked or explicitly not

*Supersedes the checklist above, which predates two commits and the CI investigation. CI on this branch is now **fully green**.*

### "Done when" — the four the ticket names

- [x] **A file and a folder both round-trip, and the folder keeps its executable bits.** `core-files-mtls.test.ts` does both over the real mTLS transport with nothing injected: a file uploaded and read back byte-identical, and a folder uploaded as one tar, downloaded as one tar, compared tree-for-tree against the source. Bits asserted on a file (`0o755`), a non-executable sibling (`0o644`) and the **directory** holding them (`0o750`), outbound and inbound. `files-tar-round-trip.test.ts` adds the case that silently regresses: the mode after an *overwrite*, where `open(…, mode)` applies nothing.
- [x] **Every confinement case is tested: absolute, `..`, and a symlink pointing out.** All three at the confinement layer, all three at the route, all three over real mTLS, and all three inside a tar entry. Each asserts the refusal *and* that nothing landed outside.
- [x] **A second concurrent write to the same Project is refused IMMEDIATELY with a distinguishable error.** `409` + `code: "transfer-in-progress"` — distinguishable by status from a bad path (400) and a missing Project (404), by code from any other 409. Immediate asserted twice: before the body is read, and with `Expect: 100-continue` before the client sends a byte. Also: reads served during a write, a different Project served concurrently, the lease released after success *and* after a mid-stream failure.
- [x] **Both mismatch directions behave sanely.** *Older Panel, newer Core*: connects, compatible, driven exactly as today by a client that never reads the field — the `ready` frame asserted byte-identical apart from the one added key. *Newer Panel, older Core*: reads the absence, `canUseFileRoutes()` false, Core compatible and fully usable, no frame gated. An unrecognised future version reads as absent, never as a guessed superset.

### The "What" bullets

- [x] **F2 — `/v1/…` routes on the mTLS HTTPS server the WebSocket already mounts on.** Asserted directly: an upgrade and a `/v1` request answered on one port.
- [x] **F2 — bytes cross HTTPS, NEVER the core link.** No frame added; the frame set is unchanged apart from one optional `ready` field.
- [x] **F6 — upload and download, both directions.** `GET`, `HEAD`, `PUT`.
- [x] **F4 — a folder crosses as ONE streamed tar.** Streamed both sides; neither ever holds the archive.
- [ ] **F4 — per-entry NDJSON progress.** ✅ on the **write** direction, where F5 puts it and where every entry line lives. ❌ on the **read** direction — a deliberate deviation, not an oversight. A folder download's body *is* the tar, so a manifest needs either buffering the archive to hash it first (a temp copy of a `node_modules`) or appending a synthetic entry that `tar -xf` extracts into the operator's directory. The tar's own headers carry four of F10's five fields natively, and the fifth is a digest of bytes the reader is receiving, so the reader computes it in the same single pass the writer would have. Argued in **ADR 0028 D5** rather than left to be discovered.
- [x] **F3 — reject absolute paths, escaping `..`, symlinks resolving outside the Project root**, validated *after* resolution.
- [x] **F5 — overwrite by default, every overwrite named `written` / `overwritten`.** Both values asserted in one archive containing one of each, over the real transport.
- [x] **F8 — no size cap.** None exists; the only check is a fit check.
- [x] **F8 — free-space precheck.** Declared `Content-Length` vs `statfs`, `507` + `insufficient-storage`. Valid in both modes because a tar is never smaller than its contents. An unmeasurable disk **proceeds** (`null` is "do not know", never "full"); a chunked upload declares nothing. Both asserted.
- [x] **F8 — one write per Project, an immediate error not a queue; reads unrestricted and concurrent.**
- [x] **F9 — optional `files: { version: 1 }` on `ready`**, announced when and only when the routes are actually mounted.
- [x] **F9 — `CORE_LINK_PROTOCOL_VERSION` does NOT move; no Core becomes needs-update.** Still `0.15.0`, pinned by a test whose *name* carries the reason.

### The three ADRs ride this PR

- [x] **F1 — the filesystem is the model, no index, no ids** → `docs/adr/0027-the-filesystem-is-the-model.md`
- [x] **F2 — HTTPS, not the core link** → `docs/adr/0028-file-bytes-cross-https-not-the-core-link.md`
- [x] **F4 — tar, and why not per-file** → `docs/adr/0029-a-folder-crosses-as-one-streamed-tar.md`
- [x] **No docs-only PR** — all three land alongside the code they describe.

### Traps

- [x] **Validate entry paths AFTER resolution, not before.** Per entry, at the moment it is written, through the symlinks that exist *then* — the only way the two-entry escape (entry 1 plants a symlink, entry 2 writes through it) is visible. Pax and GNU long-name overrides validated too, being the second place a path can come from.
- [x] **Reject absolute entries, device nodes, and hardlinks pointing outside the root**, each with its own code. Device nodes and fifos are refused on unpack while being skipped on pack — the asymmetry is deliberate and commented.
- [x] **F3 is an accident guard NOT a security boundary; do not design it as a sandbox or claim it is one in the ADR.** ADR 0027 D5 denies it outright and names `core shell` as the escape hatch that makes any containment claim false; the denial is repeated in `files-confinement.ts`'s header and its test suite's header; the refusal is `400` not `403`. "Design it as a sandbox and say so" is listed as an alternative considered and rejected.
- [x] **The Panel is a dumb pipe (F11); exactly ONE machine validates paths.** `packages/panel` untouched. ADR 0027 D3 makes a Panel-side path check a citable review finding.
- [x] **NDJSON entries carry `path`, `size`, `mtime`, `mode`, `sha256` from the FIRST commit (F10).** All five, every entry, from the commit that introduced the tar codec.

### Boundaries

- [x] **Core-side only; `packages/panel` and `packages/cli` untouched.**
- [x] **SDK change minimal** — the `files` field, its reader, its transport pass-through, two accessors. No `project.files.*` (that is #166).
- [x] **Never force pushed.** Every push was a plain fast-forward.
- [x] **Nothing merged, `main` untouched, PR left as a draft**, base `feat/project-files`.

### Gates

- [x] **CI on this branch is green** — every check, including `Unit Tests`, `Train rules`, `Lint`, `Typecheck`, both installer E2Es and both tarball smokes.
- [x] `pnpm typecheck`, `pnpm lint` (0 errors), `pnpm scan:secrets`, commitlint over all 8 commits — all clean locally.
- [x] `pnpm -r test` — sdk 182, shared 180, cli 297, core 965, panel 1338.
- [ ] **The root `scripts/` suite is not green on this machine** — `core-launcher.test.mjs` has one red test because a real `/opt/actana` install shadows the temp install root it asserts against. Environment-specific; **CI runs it green** (16 files / 474 tests).

### One thing worth a reviewer's attention: a CI failure I chased down rather than waved through

The first two CI runs failed `Unit Tests` on `packages/cli`'s `events-tail-cursor` test — a package this PR does not touch. Rather than call it flaky, I measured it: **0/54 failures at the base tip, 3/20 on this branch.** That pointed at the branch, and swapping in the base version of the one file this PR changes that the CLI rig loads (`pty-core-link-server.ts`) took it back to 0/18.

The mechanism was not behavioural — the CLI rig injects its own server factory, sets no file surface, and gets a byte-identical `ready` frame. It was **import weight**: `PtyCoreLinkServer` was constructing the file routes itself, which dragged `core-files-routes` and the thousand-line tar codec into the import graph of every Core, every rig and every package that aliases that module, widening an existing race in a 60s-timeout test.

So `ca09131` inverts it: **the core-link server now mounts HTTP routes it is handed rather than building them**, and imports only the type, which is erased. That is a better seam on its own terms — the WebSocket server has no business knowing what a file is — and `core-entry.ts` wires the routes next to the query store they read Project roots from. After it, 2/39 locally (within noise of a genuinely flaky test) and **green in CI**.

Being straight about the residual: I could not prove the branch had zero influence on that test, only that the mechanism I found is gone and the rate is back in the base's range.

packages/core/src/core-files-routes.ts
```diff
@@ -0,0 +1,525 @@
...
+  const body = JSON.stringify({ code: refusal.code, error: refusal.message });
+  res.writeHead(refusal.status, {
+    "content-type": "application/json",
+    "content-length": String(Buffer.byteLength(body)),
+    "cache-control": "no-store",
+  });
+  res.end(body);
+}
+
+function once(emitter: NodeJS.EventEmitter, event: string): Promise<void> {
```
qcentic-adm: **Blocker.** This promise can never settle after the connection dies: `'drain'` is not emitted on a destroyed stream, and there is no `close`/`error` listener to reject or resolve it.

That makes the `finally { lease.release() }` at line 383 unreachable in a real case — a client whose NDJSON progress stream has backpressured (line 356) and which then aborts mid-upload. The handler stays parked, `ProjectWriteLocks` keeps its entry, and every later `PUT` to that Project answers `409 transfer-in-progress` for the process lifetime. The Project is unwritable until the Core restarts.

Suggest settling on `close` and `error` too, and additionally releasing the lease from `res.on("close", …)` so the guarantee does not depend on the happy path unwinding. The same helper is the read side's fd leak at line 260.

packages/core/src/core-files-routes.ts
```diff
@@ -0,0 +1,525 @@
...
+
+      res.writeHead(200, {
+        "content-type": "application/x-ndjson",
+        "transfer-encoding": "chunked",
+        "cache-control": "no-store",
+        "x-actana-transfer-kind": asTar ? "tar" : "file",
+      });
+
+      const writeLine = async (value: unknown): Promise<void> => {
+        if (!res.write(`${JSON.stringify(value)}\n`)) await once(res, "drain");
```
qcentic-adm: This is where the stranding above becomes reachable. A client that streams a large body and only reads the progress stream at the end backs the response up past the 16 KB high-water mark within ~100 entries; `res.write` returns false; the handler parks here. If the client then hangs up, `unpackTarInto` is suspended inside its `onEntry` callback and the lease is never released.

The missing test: abort the request socket mid-upload with the response backpressured, then assert `locks.current("p1")` is null. `releases the lease even when the transfer failed mid-stream` does not cover it — that request completes normally and only its *content* is bad.

packages/core/src/core-entry.ts
```diff
@@ -306,6 +307,31 @@ async function startCore(): Promise<void> {
...
+  const fileRoutes = createCoreFilesRequestHandler({
+    filesPort: {
+      projectRoot: (projectId) =>
+        coreQueryStore.listProjects().find((project) => project.projectId === projectId)?.path ?? null,
+    },
+    // Assigned below, once the remote-mode block has minted or loaded the
+    // material: the file routes are gated by the same bearer the `auth` frame
+    // is, and on a loopback Core there is none to check — the same trade that
+    // Core's core link already makes.
+    authVerifier: (bearer) => serverOpts.authVerifier?.(bearer) ?? { ok: false, reason: "malformed" },
```
qcentic-adm: This verifier is never absent, so `CoreFilesRoutesOptions.authVerifier`'s documented "omitted → loopback Core" case cannot happen in production. `serverOpts.authVerifier` is only assigned inside `if (remoteMode)` (lines 389, 427), so on the default loopback Core this closure returns `{ ok: false, reason: "malformed" }` for anything presented — and `checkBearer` 401s every `/v1` request.

Meanwhile `announceFiles` defaults to `Boolean(opts.httpRoutes)`, so that same Core puts `files: { version: 1 }` on `ready`. A client that reads the capability stops feature-detecting and starts calling, which is precisely the failure ADR 0028 D4 warns about, and it contradicts that ADR's consequence that the loopback Core gets the routes over plain HTTP.

Either pass `authVerifier` only when `serverOpts.authVerifier` exists, or pass `announceFiles: false` when it does not. Whichever way, a test over `core-entry`'s wiring rather than a hand-wired rig — the mTLS suite passes a working verifier, which is why nothing caught this.

packages/core/src/files-tar.ts
```diff
@@ -0,0 +1,700 @@
...
+    // is a bug in this file, not bad input.
+    throw new TarError("corrupt-archive", `tar field overflow writing ${JSON.stringify(value)}`);
+  }
+  block.set(bytes, offset);
+}
+
+function writeOctal(block: Uint8Array, offset: number, length: number, value: number): void {
+  // ustar writes N-1 octal digits then a NUL. Values that do not fit — a file
+  // over 8 GiB in a 12-byte size field — are the reason `packEntryHeader`
+  // falls back to a pax record.
```
qcentic-adm: This comment describes a fallback that does not exist: `packEntryHeader` (line 219) routes to a pax record for a long **name** or **linkname** only, never for an oversized size field, and `parsePaxRecords` does not read a `size` record either.

The consequence is a real limit against F8's "no size cap":

- **Pack.** The ustar size field holds 11 octal digits, max 8589934591 bytes. A folder containing any file ≥ 8 GiB throws `TarError("corrupt-archive", "tar field overflow")` from `writeString` mid-stream, so the `GET` is a destroyed socket after a 200.
- **Unpack.** GNU/pax archives encode such a size as base-256 (high bit set) or as a pax `size` record. `readOctal` parses base-256 to `NaN` → 0, so the file lands empty and the stream desyncs into a checksum failure.

It fails closed, so this is not a hole — but "no size cap. None exists" is not quite true, and the near-miss is silent. Either implement the pax `size` record this comment already promises, or say the limit out loud in ADR 0029 and refuse with a code an operator can act on.

packages/core/src/core-files-routes.ts
```diff
@@ -0,0 +1,525 @@
...
+
+/** `PUT` of a single file: write it, hash it as it goes, report the five fields. */
+async function writeSingleFile(
+  req: IncomingMessage,
+  absolute: string,
+  relative: string,
+): Promise<TarEntryReport & { result: TarWriteOutcome }> {
+  await fs.promises.mkdir(path.dirname(absolute), { recursive: true });
+  const existing = await fs.promises.lstat(absolute).catch(() => null);
+  if (existing && !existing.isFile()) {
```
qcentic-adm, line 499: A single-file `PUT` at a path that currently holds a **directory** removes that whole tree recursively, then reports it as an ordinary `overwritten` line. `files-tar.ts:652` does the same for a file entry landing on a directory.

F5's overwrite-by-default is about replacing a file; recursive removal of a subtree is a different promise and is not named in the ticket, in ADR 0029 D6 (which is careful to talk about the final component and symlinks) or in `docs/external-api.md`. `tar(1)` refuses this case rather than deleting. A `PUT …?path=src` meant for `src/x.ts` costs the operator `src`, and the progress line does not say so.

At minimum document it and give it a distinguishable result value; better, refuse a non-empty directory.

packages/core/src/core-files-routes.ts
```diff
@@ -0,0 +1,525 @@
...
+        // alternative — walking the tree twice to measure it — is the second
+        // pass over every byte this design exists to not do.
+        "transfer-encoding": "chunked",
+        "cache-control": "no-store",
+        "x-actana-transfer-kind": "tar",
+      });
+      if (req.method === "HEAD") return void res.end();
+      try {
+        for await (const chunk of packDirectory(absolute)) {
+          if (!res.write(chunk)) await once(res, "drain");
```
qcentic-adm: Read-side instance of the same `once(…, "drain")` problem. A client that aborts a folder download while this is parked leaves the `packDirectory` generator suspended, so the `finally { await handle.close() }` inside it never runs — one leaked fd per aborted download, and aborted downloads are ordinary (an operator hits Ctrl-C on a `node_modules`). Line 297 is the same on the single-file path. Fixing `once` fixes all three.

packages/core/src/files-tar.ts
```diff
@@ -0,0 +1,700 @@
...
+      await onEntry({
+        path: confined.relative,
+        kind: header.typeflag === TYPE_SYMLINK ? "symlink" : "file",
+        size: Buffer.byteLength(rawLink),
+        mtime: header.mtime,
+        mode,
+        sha256: createHash("sha256").update(rawLink).digest("hex"),
+        result,
+      });
+      continue;
```
qcentic-adm, line 729: Nit, and it fails closed. Symlink, hardlink and directory entries `continue` without `reader.skip(bodySize + bodyPadding)`, so an entry that declares a non-zero size desynchronises the stream and the next 512 bytes are parsed as a header — a checksum failure and `corrupt-archive` rather than a silent misread. `tar(1)` skips the declared body instead.

No escape follows from it, since whatever is parsed still goes through `confineWriteTarget`, but it is the shape that produces parser-disagreement bugs, and the hardening suite is the natural place for a case.

qcentic-adm (changes requested):
Reviewed against #165 in full, reading the enforcing line rather than the description. This is high-quality work and most of it I could not fault: the tar hardening is genuinely hostile-input-tested, the after-resolution ordering is real (including through the pax/GNU override path), the accident-guard framing is denied as a boundary consistently in code, ADR, docs and test headers, no file byte touches the core link, `CORE_LINK_PROTOCOL_VERSION` is unmoved and pinned by a test, the Panel is untouched and nothing validates paths client-side, and the five F10 fields are on every entry from the codec's first commit with a real digest of real bytes.

Two defects stop me approving. Both are in the wiring rather than the tar codec, and neither is caught by the 131 new tests.

---

### 1. Blocker — the write lease is stranded by an abrupt disconnect, and the Project becomes permanently unwritable

`handlePut` releases the lease in a `finally`, which is right, but the body it guards can park forever. `once(res, "drain")` resolves on exactly one event and `'drain'` is never emitted on a destroyed stream. If the client stops draining the NDJSON progress stream (16 KB of response backs up — ~100 entries of a folder upload) and *then* hangs up, the handler is suspended inside `writeLine`, the `finally` never runs, `ProjectWriteLocks` keeps the entry, and every later `PUT` to that Project answers `409 transfer-in-progress` for the lifetime of the process. The refusal that F8 asks to be immediate becomes permanent.

I reproduced the mechanism against this exact control flow (plain Node, no repo files touched):

```
  [server] backpressured -> awaiting 'drain'
  [client] aborting mid-upload
RESULT: parked-on-drain=true released=false
STRANDED — finally never ran; Project stays unwritable for the process lifetime
```

The suite's nearest case, `releases the lease even when the transfer failed mid-stream`, sends a corrupt tar over a *complete* request — the socket closes normally, the iterator throws, the `finally` runs. Nothing in the suite destroys a request socket mid-upload.

The fix wants `once` to settle on `close`/`error` as well as `drain` (or `stream.once`/`finished` with the response's abort signal), plus a `res.on("close", …)` release as the belt-and-braces path. Whatever shape it takes, please land it with the test that is missing: client aborts mid-upload with the progress stream backpressured, then assert `locks.current(projectId)` is null.

### 2. The `files` capability is announced on the loopback Core, whose routes 401 every request

`core-entry.ts:331` always passes an `authVerifier`, and it returns `{ ok: false }` whenever `serverOpts.authVerifier` is unset. That field is only ever assigned inside `if (remoteMode)` (lines 389 and 427). So on the default loopback Core: `httpRoutes` is set → `announceFiles` defaults to true → `ready` carries `files: { version: 1 }` → and every `/v1` request is refused `401 unauthorized`, because there is no bearer to verify and the fallback says no.

That is the exact failure ADR 0028 D4 names — "announcing one whose routes are not answering is worse than announcing nothing, because a client that reads the field stops feature-detecting and starts calling" — and it contradicts that ADR's own consequence ("The loopback `ws://` Core gets the routes too, over plain HTTP") and the doc comment on `CoreFilesRoutesOptions.authVerifier` ("When omitted — the loopback `ws://` Core, and tests — the surface is as trusted as the rest of that Core"). The intended "omitted" case is unreachable in production. #166 and #168 consume `canUseFileRoutes()`, so they inherit this on every local Core.

Either pass `authVerifier` only when there is a real verifier to pass, or set `announceFiles: false` when there is not. The mTLS suite wires `httpRoutes` by hand with a working verifier, which is why it is green; nothing covers `core-entry.ts`'s wiring.

---

### Not blocking, but worth a decision before this is consumed

- **≥ 8 GiB files break the tar path in both directions**, against F8's "no size cap" (`files-tar.ts:121`). Details in the inline comment; the comment there claims a pax fallback that `packEntryHeader` does not implement.
- **A single-file `PUT` at a path that is currently a directory silently `rm -rf`s the tree** and reports it as an ordinary `overwritten` (`core-files-routes.ts:417`, and the same shape at `files-tar.ts:652`). `tar(1)` refuses this rather than deleting. Overwrite-by-default (F5) is about files; recursive tree removal is not named in the ticket, the ADRs or `docs/external-api.md`.
- **An aborted folder download leaks an fd** — same `once(…, "drain")` root cause, read side (`core-files-routes.ts:260`).
- **Link and directory entries with a non-zero declared body size are not skipped** (`files-tar.ts:642`). It fails closed, so it is a nit, not a hole.

### Acceptance criteria, one by one

| # | Criterion | Verdict |
| --- | --- | --- |
| 1 | File + folder round-trip, exec bits survive | **Met** — asserted on a file (0o755), a sibling (0o644) and the directory (0o750), both directions, over real mTLS; plus the overwrite case where `open(…, mode)` applies nothing |
| 2 | Confinement: absolute, `..`, symlink out | **Met** — at the confinement layer, at the route, over mTLS, and inside a tar entry; each asserts nothing landed outside |
| 3 | Second concurrent write refused immediately, distinguishably | **Partially met** — 409 + `transfer-in-progress` is correct and genuinely immediate (incl. `Expect: 100-continue`), but defect 1 can make it permanent |
| 4 | Both mismatch directions sane | **Met** — both tested; an unrecognised future version reads as absent, never as a guessed superset |
| F2 | Routes on the same mTLS server as the core link | **Met** — one upgrade and one `/v1` request on one port, default factory, nothing injected |
| F2 | No file byte on the core link | **Met** — the frame set is unchanged apart from one optional `ready` field |
| F6 | Both directions | **Met** — GET/HEAD/PUT |
| F4 | Folder as one streamed tar | **Met** — streamed both sides, neither holds the archive |
| F4/F10 | Per-entry NDJSON progress | **Partially met** — write direction only; the read-direction deviation is argued in ADR 0028 D5 and disclosed unchecked in your checklist. I accept the argument |
| F3 | Confinement, validated after resolution | **Met** |
| F5 | Overwrite by default, every overwrite named | **Met** — both values asserted in one archive over the real transport |
| F8 | No size cap | **Partially met** — no policy cap, but see the 8 GiB ustar field |
| F8 | Free-space precheck | **Met** — declared length vs `statfs`, 507; null means "do not know" and proceeds; chunked declares nothing |
| F8 | One write per Project, reads concurrent | **Partially met** — reads are genuinely ungated; the write lease is defect 1 |
| F9 | Optional `files: { version: 1 }` on `ready` | **Partially met** — correct on a remote Core, announced-but-401 on a loopback Core (defect 2) |
| F9 | `CORE_LINK_PROTOCOL_VERSION` unmoved | **Met** — 0.15.0, pinned by a test whose name carries the reason |
| F1/F2/F4 | Three ADRs, tar-not-per-file, head-of-line blocking | **Met** — 0027/0028/0029, both arguments made in full with alternatives rejected in writing |
| Trap | Validate after resolution, incl. pax/GNU overrides | **Met** — every escape I hunted for has a test that also asserts nothing landed outside |
| Trap | Absolute entries, device nodes, fifos, hardlinks out | **Met** — each with its own code; pack/unpack asymmetry deliberate and commented |
| Trap | Accident guard, never claimed as a boundary | **Met** — 400 not 403; denied in ADR 0027 D5, the module header, the test header and `docs/external-api.md` |
| Trap | Panel is a dumb pipe, one machine validates | **Met** — `packages/panel` and `packages/cli` untouched; the SDK validates no path |
| Trap | Five NDJSON fields from the first commit | **Met** — all five in `TarEntryReport` at a2a6a24; sha256 asserted against the real digest of real bytes |

CI is green on every check, the base is `feat/project-files`, and the one unchecked local red (`core-launcher.test.mjs`) is the known environment-specific one. Fix the two defects above — each with the test that would have caught it — and I will re-review promptly.

packages/core/src/core-files-routes.ts
qcentic-dev: Fixed in \`deb3573\`.

\`once(res, "drain")\` is replaced by \`drained(res)\`, which settles on \`close\` and \`error\` as well — and settles them by **throwing**, so the transfer unwinds the way any other mid-stream failure does rather than merely returning. That is what runs the `finally`, and on the read side it is also what closes the suspended \`packDirectory\` generator so its own \`finally { handle.close() }\` runs.

The belt-and-braces release you suggested is there too: \`res.on("close", () => lease.release())\`, registered the moment the lease is taken. \`release\` is already idempotent and compares identity before deleting, so the two paths cannot free each other's entry.

Deliberately not the only guarantee — the listener is the backstop, the throw is the fix. I checked they are independently sufficient by reverting each in turn.

packages/core/src/core-files-routes.ts
qcentic-dev: The missing test is `deb3573`, in a new suite `packages/core/src/__tests__/core-files-abort.test.ts`.

It is the case you described: 900 entries with long names (~220 KB of progress for 9 KB of payload), the client never reads the response (`res.pause()`), it waits until the lease is held and the 200 has arrived, then destroys the socket — and asserts `locks.current("p1")` is null.

I verified it is a real regression test rather than a test that happens to pass: reverted `core-files-routes.ts` to the parent commit and it fails by **timeout**, which is the defect itself — the handler never unwinds. Two more alongside it: the Project is still writable afterwards (a later `PUT` is served, not `409`), and the read-side fd leak.

You were right that `releases the lease even when the transfer failed mid-stream` does not cover this. I have said so in the new suite header, since that is where the next person will look.

packages/core/src/core-files-routes.ts
qcentic-dev: Fixed by the same change, `deb3573` — one `drained` helper, three call sites (260, 297, 356).

Tested, and the first version of that test was worthless: it counted `/proc/self/fd` entries in total and **passed against the unfixed code**, because socket churn swamped the signal and small fixture files meant the packer was usually between files with nothing open. It now counts descriptors whose `readlink` resolves inside the Project root, over 6 files of 4 MB so the abort lands mid-file. Against the parent commit it fails with three leaked descriptors named in the diff:

```
+ [ "…/payload/chunk-0.bin", "…/payload/chunk-0.bin", "…/payload/chunk-1.bin" ]
```

Linux-only via `it.skipIf` — `/proc/self/fd` is the only honest way to see it, and the server runs in the test process.

packages/core/src/core-entry.ts
qcentic-dev: Fixed in `f1ca014`, taking your first option — pass `authVerifier` only when there is one.

That preserves ADR 0028's consequence (the loopback Core keeps its routes over plain HTTP) instead of trading it away, and it makes the documented "omitted" case reachable in production rather than only in tests. The ordering was the obstacle: the routes were built before the remote-mode block, so a late-bound closure was the only option available. They are now built after it, so the verifier is passed **by value**.

On your point about testing `core-entry`'s wiring rather than a hand-wired rig — that is the real finding here, and it is why nothing caught this. `core-entry` boots a Core on import, so no test can load it. I extracted the one decision into `core-files-wiring.ts`, which `core-entry` now calls, and the rule is stated there as a sentence:

> The file routes are gated by a bearer when, and only when, the core link on the same server is.

`core-files-wiring.test.ts` asserts it from both ends, including ADR 0028 D4 as an executable claim — *a Core announcing `files` answers an authorised request* — parameterised over both wirings a Core can have. Confirmed it catches the defect by reinstating the old closure shape in the seam: 3 of the 7 fail, all three loopback ones.

ADR 0028 D4 now records what went wrong and what the rule is, since the ADR was right and the code disagreed with it.

packages/core/src/files-tar.ts
qcentic-dev: Fixed rather than documented-as-a-limit, in `2b2d377` — you were right that the comment promised something real and F8 says no size cap, so the honest resolution was to make the promise true.

**Pack.** `packEntryHeader` now emits a pax `size` record above `MAX_USTAR_SIZE` (8589934591) and writes 0 in the ustar field, so a reader that ignores pax sees an empty entry rather than a truncated wrong one.
**Unpack.** `readOctal` reads GNU base-256 (high bit set, big-endian after it), and the pax `size` record is honoured alongside `path` and `linkpath` — same pending/consume discipline, cleared on the next real entry. A leading `0xff` (base-256 negative) is refused as `corrupt-archive` rather than read as a length.

Tested at the encoding, not with an 8 GiB fixture — the parse is what was broken and it is broken or correct at any magnitude, and an 8 GiB fixture would make the suite unrunnable on a developer machine. That is why `packEntryHeader` is now exported; the export is annotated with that reason. The unpack side is covered with a real body and a base-256 field, and with a pax `size` record where the ustar field says 0 — both followed by a second entry, so a desync fails the test.

Worth noting on the negative-field case: it also died as `corrupt-archive` before the fix, two entries later and for an unrelated reason. I assert on the message as well so the test stays pinned to the field rather than to the accident.

Recorded in ADR 0029 as **D6b**, including why the gap survived: it failed closed in both directions, so nothing was wrong except that a documented promise was not kept, at a magnitude no fixture reaches.

packages/core/src/files-tar.ts
qcentic-dev: Fixed in `2b2d377` — `reader.skip(bodySize + bodyPadding)` before all three `continue`s (directory, symlink/hardlink, and the empty-name case, which had the same shape).

Agreed it is a nit that fails closed, and I have kept your reasoning in the comment rather than just the fix: no escape follows, since whatever is parsed still goes through `confineWriteTarget` — but "the writer and the reader disagree about where the next entry starts" is the shape that produces parser bugs, and `tar(1)` skips the declared body.

Three cases in the hardening suite as you suggested, one per entry type, each with a following file entry so a desync is what fails the assertion. All three red on the parent commit.

packages/core/src/core-files-routes.ts
qcentic-dev: Refused rather than documented, in `2b2d377` — you offered "at minimum document it", and I took the stronger option because the argument in your comment is the right one: the operator did not ask for a recursive delete, and the progress line saying `overwritten` makes the damage silent.

A **non-empty** directory is now refused with its own code, `directory-in-the-way`:

- **At the route**, *before* the 200 goes out, so it is a status a client can branch on in the same place `transfer-in-progress` is — not an error line arriving after a success status and after the body has been uploaded. 409, since the request is well-formed and the path is legal; the disk just holds something this surface will not delete.
- **In the codec** (`files-tar.ts`), refused mid-archive with the same code, with a test asserting nothing was deleted on the way to the refusal.

An **empty** directory is still replaced. The asymmetry is deliberate and commented: the rule is about destroying work, not about the node type, and a stray `mkdir` should not wedge a path forever.

Documented in both places you noted were silent — `docs/external-api.md` gains the code and the sentence *overwriting replaces a file; it never removes a tree*, and ADR 0029 gains **D6a**, which is explicit that D6 is about the final component and about symlinks and says nothing about subtrees.

One consequence worth flagging: this is a **behaviour change to a case that previously succeeded**. Anything relying on `PUT ?path=dir` to clear a tree now gets a 409. Nothing in this repo does, `packages/panel` and `packages/cli` are untouched, and #166/#168 are not written yet — but it is the kind of thing better said than discovered.

qcentic-dev:
## Review response — every finding from the CHANGES_REQUESTED review

*Supersedes the checklist above. Three commits on top of `ca09131`, all plain fast-forward pushes:*

| sha | Finding |
| --- | --- |
| `deb3573` | Blocker 1 — the stranded write lease, and the read-side fd leak |
| `f1ca014` | Blocker 2 — `files` announced on a loopback Core whose routes 401 |
| `2b2d377` | The three non-blocking findings, plus their docs |

### The two defects, each with the test that would have caught it

- [x] **1 — The write lease is stranded by an abrupt disconnect** (`deb3573`). `once(res, "drain")` never settled once the socket died, so the handler parked inside `writeLine` and the `finally` never ran. `drained(res)` now settles on `close` and `error` too, **by throwing**, so the transfer unwinds exactly as any other mid-stream failure does — and on the read side that is what closes the suspended `packDirectory` generator and runs its `finally { handle.close() }`. The belt-and-braces `res.on("close", () => lease.release())` is there as well; `release` was already idempotent and identity-compared.
  **The test:** `core-files-abort.test.ts` — 900 backpressuring entries, the client never reads the response, abort mid-upload, assert `locks.current("p1")` is null. Plus: the Project is still writable afterwards, and an aborted download leaks no descriptor. Verified red on `ca09131` — the two lease tests fail by *timeout*, which is the defect itself.

- [x] **2 — The `files` capability announced on a Core that 401s every request** (`f1ca014`). Took the first option offered: pass `authVerifier` only when there is one, which keeps ADR 0028's consequence that the loopback Core gets the routes over plain HTTP rather than trading it away. The blocker was ordering — the routes were built before the remote-mode block, so a late-bound closure was all that was available. They are now built after it and the verifier is passed by value.
  **The test:** `core-files-wiring.test.ts`, over a new seam `core-files-wiring.ts` that `core-entry` calls — because `core-entry` boots a Core on import and no test can load it, which is precisely why nothing caught this. It asserts ADR 0028 D4 as an executable sentence: *a Core announcing `files` answers an authorised request*, over both wirings a Core can have. Verified it catches the defect by reinstating the old closure in the seam: 3 of 7 fail.

### The four not-blocking findings — all fixed, none deferred

- [x] **≥ 8 GiB files break the tar path in both directions.** Fixed rather than documented as a limit: the comment promised a pax fallback and F8 says no size cap, so the promise is now kept. Pack emits a pax `size` record above `MAX_USTAR_SIZE`; unpack reads both GNU base-256 and the pax `size` record, and refuses a negative base-256 field. Tested at the encoding rather than with an 8 GiB fixture — which is why `packEntryHeader` is now exported, annotated with that reason. **ADR 0029 D6b.**
- [x] **A single-file `PUT` onto a directory silently `rm -rf`s the tree.** Refused, not merely documented — the argument for refusing was the right one. `409 directory-in-the-way`, raised *before* the 200 so it is a branchable status rather than an error line after a spent success. Same refusal in the codec. An **empty** directory is still replaced; the asymmetry is deliberate and commented. **ADR 0029 D6a** and `docs/external-api.md`. ⚠️ This is a behaviour change to a case that previously succeeded — flagged on the thread.
- [x] **An aborted folder download leaks an fd.** Same root cause, fixed by the same helper. The first version of that test counted total `/proc/self/fd` entries and **passed against the unfixed code**; it now counts descriptors resolving inside the Project root, over files large enough that the abort lands mid-read. Red on `ca09131` with three named leaked descriptors.
- [x] **Link and directory entries with a non-zero declared body are not skipped.** `reader.skip(bodySize + bodyPadding)` before all three `continue`s, including the empty-name case which had the same shape. Three hardening cases, each followed by a file entry so a desync is what fails.

### Docs carried with the code

- [x] **ADR 0028 D4** records what went wrong and states the rule: the file routes are gated by a bearer when, and only when, the core link on the same server is.
- [x] **ADR 0029 D6a / D6b** — replacing a file is not the same promise as removing a tree; a size ustar cannot hold is a pax record, not a refusal.
- [x] **`docs/external-api.md`** — the new `409 directory-in-the-way`, *overwriting replaces a file; it never removes a tree*, and the lease living only as long as the request.

### Gates

- [x] `pnpm typecheck` — clean across all five packages.
- [x] `pnpm lint` — **0 errors**, 51 warnings, all pre-existing and none in the new files (unchanged count).
- [x] commitlint over all 12 commits, via the same throwaway install CI uses — 0 problems, 0 warnings. All three new subjects under 112 characters.
- [x] `@actana/core` — **992** tests green (was 965; +27 from this response). sdk 182, shared 180, cli 297.
- [ ] **`@actana/panel` — 2 tests red in a full-package run, green in isolation.** `live-read-path` and `terminals-in-browser`, both 5s timeouts under load; they pass when the two files are run alone. `packages/panel` is untouched by this branch — confirmed against the merge base — so this is local load, not the branch. CI is the authority.
- [x] **Never force pushed.** Every push a plain fast-forward; the last was `ca09131..2b2d377`. The base has not moved, so no merge of `origin/feat/project-files` was needed.
- [x] **Nothing merged, `main` untouched, still a draft**, base `feat/project-files`.

### One thing I got wrong while doing this, since it changes how much the tests are worth

The fd-leak test **passed against the broken code** when I first wrote it — a total descriptor count is too noisy, and with small fixture files the packer is usually between files with nothing open. I only found out because I make a point of reverting the fix and watching each new test fail. So I measured it rather than asserting it: with the three source files put back to `ca09131` — keeping only the `export` on `packEntryHeader` so the suite can still load, and the old always-present verifier closure reinstated in the seam — **18 of the 27 new tests are red**, and every one of the 18 is a new test rather than collateral. The two lease tests fail by *timeout*, which is the defect itself. The remaining 9 are regression guards for behaviour that was already correct: an empty directory still being replaced, a size that fits staying plain octal, the lease released after an ordinary refusal.

qcentic-adm (approved):
Re-reviewed `ca09131..2b2d377` against #165, reading the enforcing line and each accompanying test against the pre-fix code rather than the commit messages. **Both blockers are fixed, both fixes carry a test that is genuinely red on the parent, and all four non-blocking findings were fixed rather than deferred.** Approving.

---

### 1. Blocker — the stranded write lease: **resolved**

`drained()` (`core-files-routes.ts:642`) settles on `close` and `error` as well as `drain`, and settles them by throwing `ClientGoneError` — which is what unwinds the transfer so the `finally` at line 442 runs. Two things I checked beyond the description:

- The **already-destroyed guard** at the top (`res.destroyed || res.writableEnded → reject`). Without it the fix would still have a hole: registering a listener for an event that has already fired parks exactly as before. Good that it is there.
- The **belt-and-braces path is actually safe**. `res.on("close", () => lease.release())` at line 346 only works because `release` is idempotent *and* identity-compares before deleting (`files-transfer-locks.ts:67-72`) — verified, so the listener and the `finally` cannot free each other's entry.

The read side is fixed by the same helper, and I confirmed the mechanism rather than taking it: throwing inside `for await (const chunk of packDirectory(...))` calls the generator's `return()`, which runs `finally { await handle.close() }` at `files-tar.ts:385`.

**The test is real.** `core-files-abort.test.ts` builds 900 long-named entries (~220 KB of progress for 9 KB of payload), never reads the response (`res.pause()`), waits until the lease is held *and* the 200 has arrived, then destroys the socket. Read against `ca09131`: `once(res, "drain")` never settles, and — the part that makes it decisive — the handler is parked inside the `onEntry` callback, so it is not pulling from `req`; the request stream's error is queued for a `next()` that never comes. The handler stays parked, the `finally` never runs, and `ca09131` has no `res.on("close")` listener to save it. Red by timeout, which is the defect itself, exactly as claimed.

### 2. Blocker — `files` announced on a Core that 401s: **resolved**

The routes are now built at `core-entry.ts:442`, after the remote-mode block closes at 417, and the verifier is passed **by value** through a conditional spread (`:447`). I checked the thing that would have made this a worse bug than the one it replaces: both remote-mode branches assign `serverOpts.authVerifier` unconditionally (`:364`, `:402`) and every early-out in that block is a `process.exit(1)`, so **no remote Core can reach the build site without a verifier**. There is no path to an ungated file surface on a public server.

`buildCoreFileRoutes` carries the rule as a sentence, and `core-files-wiring.test.ts` asserts ADR 0028 D4 as an executable claim over both wirings. The "3 of 7 fail" claim reconciles exactly: reinstating the always-present closure breaks the two loopback route tests and the loopback half of the parameterised case, while the three `shouldAnnounceFiles` assertions and the remote cases stay green.

One residual, which you disclosed and I am not blocking on: the suite covers the new seam, not `core-entry.ts:447` itself — a closure reintroduced at that call site would still pass. `core-entry` boots a Core on import, so that is a real constraint rather than a shortcut, and the extracted seam is the right answer to it. I read the call site directly instead.

### The four non-blocking findings — all fixed

- **≥ 8 GiB.** Pack emits a pax `size` record above `MAX_USTAR_SIZE` and writes 0 in the ustar field; unpack reads GNU base-256 and honours the pax `size` record with the same pending/consume discipline as `path`/`linkpath`, correctly guarded so a metadata block's own body is never overridden (`files-tar.ts:558-565`). I traced `bodySize` through the file branch — the override drives the read loop, the padding skip, `bytes`, and the reported `size`, so there is no half-applied path. Negative base-256 refused, and asserting on the *message* rather than the code is the right call given it died as `corrupt-archive` before the fix for an unrelated reason.
- **The `rm -rf` on overwrite.** Refused at the route before the 200 (`:476-489`) and mid-archive in the codec, with `refuseNonEmptyDirectory` called **before** the `rm` in both the link branch and the file branch — so nothing is deleted on the way to the refusal, which the tests assert. `lstat` means a symlink-to-a-directory is still removed as the symlink it is, which is correct. Empty directories still replaced. Thank you for flagging the behaviour change on the thread rather than leaving it to be discovered.
- **The read-side fd leak.** Same helper, verified above.
- **Unskipped bodies.** `reader.skip(bodySize + bodyPadding)` before all three `continue`s plus the empty-name case, each tested with a following file entry so a desync is what fails.

### On the test-reality claims

I confirmed **27 new tests** exactly (26 `it` blocks, one of them an `it.each` with two cases) — matching `992 - 965`. The two defect-covering tests are red on the parent for the reasons above. The 9 that stay green are recognisable as regression guards, so 18 reconciles.

Worth stating plainly for the next reader: that 18 is measured against a *hybrid* parent — the `packEntryHeader` export kept and the old closure reinstated in the new seam — not a clean `ca09131`. It has to be, since two of the suites test a module and an export that do not exist there. You said so yourself; I am repeating it because the number will outlive the thread.

### The rewritten fd test — judged, and it holds up

The rewrite is sound and a genuine improvement. Counting by *what the descriptor points at* rather than how many there are is the right axis, and the fixture is sized to the mechanism: 6 × 4 MB means the 24 MB body backpressures a paused client almost immediately, so the abort lands inside a file's read loop with a handle open, which is where the leak is. Three iterations show accumulation rather than a single ambiguous descriptor. Critically, `makeTree` returns a `fs.realpathSync` path (`files-fixture.ts:25`), so the `readlink(...).startsWith(root)` comparison cannot be silently defeated by a symlinked `tmpdir` — that would have reintroduced exactly the vacuity you caught the first time.

One thing I would still change, not blocking: **there is no positive control.** Both the precondition and the assertion expect `[]`, so if the filter ever stopped matching, the test goes vacuously green rather than red — the same failure mode as the first version, just harder to reach. One assertion that the counter *sees* a descriptor while a download is in flight would pin the instrument as well as the result. Also worth knowing that `it.skipIf(process.platform !== "linux")` means macOS developers get no signal from it at all; CI is Linux, so CI is the authority on this one.

### One new nit, from reading the tests rather than the code

`files-tar-round-trip.test.ts` — **"round-trips such a header through this module's own reader"**. The comment above it says *"A pax `size` record and a real body of that size, so the assertion is about the parse rather than about the disk"*, but the test calls `roundTrip()` on a tree holding one 19-byte file: no pax record, no base-256 field, no size near the boundary. It is green on the parent, and it duplicates the coverage the first test in the file already has, under a heading about 8 GiB.

The behaviour itself **is** properly covered — "reads a base-256 size field" and "honours a pax `size` record" in the hardening suite are both real and both red on the parent — so this is a mislabelled test, not a gap, which is why it is a nit. Flagging it because it is the same class of thing as the fd counter you caught yourself: a comment claiming coverage the body does not provide is worse than no comment, since the next person reads the comment and stops.

### Regression check

- `files-confinement.ts` and `files-confinement.test.ts` are **untouched** by the fix round.
- **No test was deleted or weakened** anywhere in the escape battery. The only two deletions across all three test files are an extended import and one line in the raw-header fixture helper, which now reads `declaredSize ?? content.length` — backward compatible, so every existing hostile-archive fixture encodes byte-for-byte what it did before.
- The confinement ordering is intact: the new `skip` on the empty-name path sits *after* `confineWriteTarget`, and the new refusal sits before every `rm`.
- ADR 0028 D4, ADR 0029 D6a/D6b and `docs/external-api.md` all landed with the code, including the new 409 in the error table and the sentence that the lease lives only as long as the request.

### CI

Green on the head sha itself, not an ancestor — run `31701684054` reports `head_sha=2b2d377`, and all 22 checks are `success` or a legitimate `skipped` (train/publish jobs).

---

Both blockers resolved, all four non-blocking findings fixed rather than deferred, and the acceptance criteria that were "partially met" on the last pass — 3, F8's size cap, F8's write lease, F9's capability — are now met. The two nits above (a positive control on the fd test, a mislabelled round-trip test) are yours to take or leave; neither needs another review round. Nice work on measuring the red-on-parent claim rather than asserting it, and on disclosing the test that fooled you — that is what made this one reviewable.

qcentic-dev:
## Both nits taken — `e5f9256`

Neither needed another round, per your note; recording what changed so the next reader does not have to diff for it.

**The mislabelled round-trip test.** You were right that it was the same class of thing as the fd counter: a comment claiming a pax `size` record and a real body of that size, above a body that round-trips a 19-byte file through the plain octal path. I relabelled rather than rebuilt it, because the coverage it claimed already exists and is already red on the parent — `reads a base-256 size field rather than treating it as zero` and `honours a pax "size" record, which is the other way a real tar says it`, both in `files-tar-unpack-hardening.test.ts`, both building the header by hand so neither needs eight gibibytes of disk. Rewriting the round-trip test to duplicate them would have bought coverage that is already paid for.

So it is now `leaves an ordinary small file round-tripping exactly as it did before` — a regression guard that the pax fallback did not disturb the size field every other file in the world uses, which is what the body actually asserts. The comment says outright that there is no pax record and no base-256 field in that archive, and names the two hardening tests, so the next person reading it is pointed at the real coverage instead of stopping.

**The positive control on the fd counter.** Added. It samples `openUnderRoot()` on a 25 ms interval while each download is in flight, and asserts `sawOneInFlight` **before** the assertion that trusts the counter. The packer is parked inside a 4 MB file's read loop with its handle open at that point, which is the same state the leak was visible in, so a counter that has gone blind — moved Project root, a `/proc` shape change, a `startsWith` that stopped matching — now fails there rather than sailing through two vacuous `[]` comparisons.

The `it.skipIf(process.platform !== "linux")` stands as you read it: macOS developers still get no signal from that test, and CI is the authority. I did not want to invent a second mechanism for a platform CI does not run.

**Gates.** `@actana/core` 992 green (unchanged — both changes are in place, no test added or removed), typecheck clean, eslint clean on both files, commitlint clean, subject 91 characters. Plain fast-forward push, `2b2d377..e5f9256`.

Issue #167, opened 2026-08-11
3.3 SDK project.files.list/upload/download (F12)


mrfqcentic:
**Base branch: `feat/project-files`, not `main`.** Train: **`beta/0.3.0`**.
Worktree: **`wt-129-17`**. Suggested branch: **`feat/sdk-files-surface`**.
Phase 3, step 2. Implements the SDK half of **F12**.

## What

`project.files.list / upload / download` on `@actana/sdk` — streams in and out, progress as an async iterable.

## Done when

- `upload` takes a stream and returns an async iterable of NDJSON progress entries, including every `overwritten`.
- `download` returns a stream, never a buffer. A gigabyte file must not be materialised in memory.
- `list` is an async iterable over the tree.
- The surface is unavailable, with a clear reason, against a Core that does not announce `files` on `ready` (F9).

## Traps

- **This is the SDK's first HTTPS surface**; everything before it was WebSocket. It needs the mTLS `fetch` shape the phase-1 spike froze — reuse that finding rather than rediscovering it.
- Backpressure: an async iterable that races ahead of the consumer defeats the point of streaming.
- The one-write-per-Project rule (F8) surfaces as an error, not a retry loop. A client that retries silently turns a clear conflict into a hang.

Refs #129

qcentic-dev:
Done in PR #217, merged into `feat/project-files` — a non-default base, so this did not auto-close; closing by hand.

PR #216, opened 2026-08-13
feat(core): list a Project's tree on the /v1 surface — streamed NDJSON, confined per path, sha256 on request

### LINKED ISSUE #166: 3.2 Listing on the HTTP surface, not dirList (F7, F10)

qcentic-dev:
## Summary

Phase 3 step 2 of #129: **listing a Project's files**, on the `/v1` HTTP surface
PR #215 stood up — `GET /v1/projects/:projectId/files/list`, streamed as NDJSON,
one line per entry carrying `{path, size, mtime, mode, sha256}` to arbitrary
depth. This implements **F7** and the manifest half of **F10**.

**Not on the core link's `dirList`, and that is the whole point of the ticket.**
Extending that frame moves `CORE_LINK_PROTOCOL_VERSION`, which marks *every Core
in every fleet* needs-update under ADR 0005 — for a feature most of them will
never be asked for. The separately-versioned HTTP surface degrades gracefully
instead: a Core without it simply does not announce `files` on `ready`, and a
client that checks first never asks.

It builds on #215 rather than reworking it: the same route module, the same
confinement functions, the same `drained` backpressure helper, the same NDJSON
line shape, the same `files: { version: 1 }` capability.

### The three decisions a reviewer should read closely

**`sha256` is computed on request, not eagerly — recorded as ADR 0027 D6.** This
is the ticket's named trap. A listing walks a tree without reading it: the other
four fields fall out of the `lstat` the walk makes anyway, and a digest costs a
full read of every byte under the path. ADR 0027 D4's argument for carrying the
digest is that it is free *while the bytes are in hand*, and a listing is exactly
the case where they are not — so the same rule reaches the opposite conclusion
here, and `?sha256=1` asks for it. F10's future diff endpoint gets a digest that
is **available** — one request away, over a surface that already streams — rather
than free, which is what F10 asks for. The field is present in every entry either
way, so nothing feature-detects it and the shape stays the tar manifest's.

**A route of its own, not `?list=1` on the read.** The two answer with different
things: one hands back a tar of a folder, the other a manifest of it. A query
parameter that a proxy, a redirect or a hand-edited URL can drop would turn
"list this folder" into "download this folder", which for a `node_modules` is a
mistake measured in gigabytes rather than in a `400`. A path segment cannot be
dropped silently. `PUT` on it is a `405` rather than a fall-through to the write
path, which would otherwise create a *file called `list`*.

**A listing walks, so confinement has to hold for every path it emits** — not
just the one it was handed, which is all a single-path read has to prove.
Symlinks are entries and are never followed: a self-referential link would
otherwise be an infinite listing, and a link out of the Project would put paths
from outside the root into a stream whose every path is meant to be
Project-relative. A directory can still contain an ancestor without a symlink (a
bind mount does it), so the ancestor chain rides down the walk as `dev:ino`
pairs — O(depth), because a cycle must revisit an ancestor.

### Nothing is buffered, on either side

`listTree` is an async generator that yields one entry and forgets it; the route
writes each line straight out and waits on `drained` when the socket stops
accepting them. The only state the walk holds is the chain of directories it is
currently inside. `opendir` rather than `readdir`, so a single folder with a
hundred thousand entries costs a batch of names rather than all of them — which
is also why a listing arrives in the filesystem's order, since sorting a
directory means holding all of it. A client that hangs up unwinds the walk
through the same path a write unwinds through, closing every directory handle it
was holding; there is a `/proc/self/fd` test for exactly that, because it is the
read-side instance of the leak #215's abort suite found on the write side.

Phase-4 sync is **reserved, not built**. Nothing here is built toward it beyond
the manifest fields F10 already fixed.

## Related issues

Closes #166
Refs #129, #150

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## How was this tested?

**69 new tests across four suites**, all green. Full `packages/core` run green:
**1061 passed, 57 files**. `packages/shared` 180 and `packages/cli` 297 green.
`pnpm typecheck` clean, `pnpm lint` **0 errors** (51 pre-existing warnings, none
in files this PR touches), `pnpm scan:secrets` clean, and `commitlint` passes on
every commit in the branch.

| Suite | Covers |
| --- | --- |
| `files-listing.test.ts` (22) | the walk itself, against a real filesystem: arbitrary depth and bounded depth, the five fields in the tar manifest's units, digests on request and their absence by default, a 24 MB file digested in chunks, symlinks reported and never followed (including a link to itself and to its own parent), fifos skipped in silence, an unreadable directory as a `skipped` line whose siblings still list, and — via `/proc/self/fd` — every directory handle closed when the walk is abandoned part-way |
| `core-files-listing-routes.test.ts` (24) | the route over a real HTTP server: NDJSON framing and headers, the `done` line's counts, Project-relative paths from a subtree listing, `HEAD`, `depth` and `sha256` including the `400` for a value neither understands, `404` for an unknown Project and a missing path, `405` for a write, the bearer gate, and the streaming claim itself — the first entry read while `res.complete` is still false, a 4001-entry listing arriving in many chunks, and no directory handle left open when a client walks away mid-listing |
| `files-listing-confinement.test.ts` (15) | **the ticket's "tested independently, not inherited on trust" clause.** Imports no confinement helper and asserts nothing about what one returns — every case goes in as an HTTP request. The three cases the routes ticket names, a Project whose own root runs through a symlink, and the rule a walk needs that a read does not: every path the walk emits resolved the way a client would resolve it and checked to be inside, nothing from the far side of an escaping symlink appearing at all, and a link's digest covering its target string rather than the file it points at |
| `files-listing-capability-gate.test.ts` (8) | a Core that announces no file surface is **not asked** — measured at the Core by counting what arrives on its HTTP surface. Not "the request was refused": there was no request. One case stands the routes up anyway and shows they would have answered, because a client that probes despite `ready` has stopped feature-detecting (ADR 0028 D4) |

**Two caveats, stated rather than glossed.** Under a *concurrent* full-monorepo
run this machine also went red on `scripts/__tests__/core-launcher.test.mjs`
(pre-existing: a real `/opt/actana` install on this dev box), on
`core-files-abort.test.ts` (#215's, passes in isolation, times out under load),
and on four `packages/panel` suites. Those panel suites were **not fully
re-verified**, because the machine's disk reached 100% mid-session from outside
this worktree and later runs could not allocate; every suite in this PR and the
whole of `packages/core` were run green before and after that. Nothing in this
change touches the Panel.

None. `CORE_LINK_PROTOCOL_VERSION` stays at `0.15.0`, no Core becomes
needs-update, and the `files` capability stays at **version 1** — #215's routes
and this listing ride the same unreleased train, so no Core anywhere announces a
version 1 that answers `GET /files` and 404s `GET /files/list`. Bumping to 2
would describe a Core that does not exist and would make every client checking
for 1 stop using a surface that works.

## Checklist

- [x] This PR targets the phase-3 integration branch `feat/project-files`, per #150 — not `main` and not `beta/0.3.0`
- [x] I added tests proving my feature works — see the caveats above for what this machine could not re-run
- [x] I updated documentation — ADR 0027 gains D6 and two consequences, `docs/external-api.md` gains the route, its parameters and its refusals
- [x] This PR is a draft; step 3 of the phase consumes it

## Scope

Core-side routes and listing only. **`packages/sdk` is untouched** — `project.files`
is #167, running in parallel, and this PR's capability-gate suite consumes the
existing `readFilesCapability` reader rather than adding anything for it.


qcentic-dev:
## Acceptance criteria, #166 — one line per clause

### Done when

- [x] **A Project's tree can be listed to arbitrary depth, streamed as NDJSON, with `{path, size, mtime, mode, sha256}` per entry.**
  `GET /v1/projects/:projectId/files/list`, `application/x-ndjson`, chunked, no
  `content-length`. Every entry carries the five fields plus a `kind`, in the same
  units as the tar manifest. Unbounded by default; `?depth=N` bounds it, and a
  `depth` the surface does not understand is a `400` rather than a silent
  "everything". Tests: `core-files-listing-routes.test.ts` (framing, the five
  fields, depth, the `done` line), `files-listing.test.ts` (the walk).

- [x] **The same confinement rules as the routes ticket apply, tested independently — not inherited on trust.**
  `files-listing-confinement.test.ts` imports no confinement helper, asserts
  nothing about what one returns, and would not notice if one were replaced —
  every case goes in as an HTTP request and is judged by what comes back.
  Absolute paths, `..` (leading, buried, and cancelling-out), a symlink resolving
  outside, innocent segments under a symlinked parent, NUL and backslash, and a
  Project whose own root runs through a symlink. Plus the rule a *walk* needs and
  a single-path read does not: every path the listing emits is resolved the way a
  client would resolve it and checked to be inside.

- [x] **A Core without the capability is not asked; the client checks `ready` first.**
  Proven at the Core, by counting what arrives on its HTTP surface: against a
  Core that announces no `files`, a client consulting the gate issues **zero**
  requests — not a refused one. One case mounts the routes anyway and shows they
  would have answered, because a client that probes despite `ready` has stopped
  feature-detecting (ADR 0028 D4). Tests: `files-listing-capability-gate.test.ts`.
  **Boundary:** the gate is `readFilesCapability` / `CoreClient.canUseFileRoutes()`,
  which already exists; the client that consumes it for listing is
  `project.files.list` in #167, running in parallel. This PR adds no SDK surface.

- [x] **Listing a large tree does not buffer it in memory on either side.**
  Core side: `listTree` is an async generator holding only the chain of
  directories it is inside; `opendir` rather than `readdir` so one huge folder
  costs a batch of names; each line is written straight out with `drained`
  backpressure. Asserted by reading the first entry while `res.complete` is still
  false, by a 4001-entry listing arriving in many chunks, and by `/proc/self/fd`
  showing no directory handle left open when the client walks away mid-listing.
  Client side: there is no client in this PR to buffer — the wire shape (chunked,
  no `content-length`, one self-contained JSON object per line) is what lets
  #167's `list` be an async iterable rather than a `await res.json()`.

### Traps

- [x] **`sha256` — decided deliberately and recorded.**
  **On request** (`?sha256=1`), null otherwise. A listing walks a tree without
  reading it, so ADR 0027 D4's "free while the bytes are in hand" does not hold
  here — the same rule reaching the opposite conclusion on a different premise.
  Recorded as **ADR 0027 D6**, where the phase records its decisions, with both
  rejected alternatives (digest everything eagerly; leave the field off entirely)
  written down beside it, and restated in `docs/external-api.md` and in the PR
  body. F10's diff endpoint gets a digest that is *available*, not free. A file
  whose bytes could not be read when they **were** asked for is a `skipped` line
  rather than an entry with a null, because a diff reading "could not tell" as
  "unchanged" is the one failure this field must not have.

- [x] **Phase-4 sync is reserved, not built.**
  Nothing here is built toward it beyond the manifest fields F10 already fixed.
  No cursor, no state between requests, no `since=`, no diff endpoint, no
  comparison of any kind. ADR 0027 D2's "stateless between requests" still holds
  for the whole surface.

### The ticket's framing rules

- [x] **Base `feat/project-files`, not `main`** — and nothing was merged, promoted or force-pushed.
- [x] **Branch `feat/files-listing`; Conventional Commits on the PR title and every one of the 5 commits** (`commitlint --from 3c3d6bb --to HEAD` exits 0). Title is 110 characters.
- [x] **Built on #215, not duplicating or reworking it** — same route module, same confinement functions, same `drained` helper, same NDJSON line shape, same `files: { version: 1 }` capability, which does not move.
- [x] **`packages/sdk` untouched** — #167 owns `project.files`. `git diff --stat` against the base names no file under `packages/sdk/`.

### Not claimed

- [ ] **A full-monorepo green run on this machine.** `packages/core` (1061), `packages/shared` (180) and `packages/cli` (297) are green. Under a concurrent full run this box also went red on `scripts/__tests__/core-launcher.test.mjs` (pre-existing: a real `/opt/actana` install here), on #215's `core-files-abort.test.ts` (passes in isolation, times out under load), and on four `packages/panel` suites — which were **not** re-verified afterwards, because the machine's disk hit 100% mid-session from outside this worktree. Nothing in this change touches the Panel. CI is the arbiter.

qcentic-adm (changes requested):
## Review against #166 — the work is right, the evidence is missing

I read every acceptance clause against the enforcing line, and the implementation
holds up: all four "done when" clauses and both traps are met, the boundary with
#215 and #167 is clean, and the four new suites are real tests rather than
decorative ones. **The blocker is not in the diff.** It is that CI has never
executed a single line of it.

### 🚩 Blocking — CI never ran `packages/core`

Run [31705815133](https://github.com/actana/control/actions/runs/31705815133/job/94465986700)
is the only CI run on this branch, and it is red. The failure is
`packages/cli/src/__tests__/events-tail-cursor.test.ts > resumes where the last
run stopped` — a 60 s timeout, in a package this PR does not touch, and the same
test timed out on #215's run
[31696968674](https://github.com/actana/control/actions/runs/31696968674) before
passing on a re-run. As a flake it would be someone else's problem.

What makes it this PR's problem is the next two lines of the log:

```
packages/cli test: Failed
[ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL] @actana/cli@0.3.0 test: `vitest run`
```

`pnpm -r test` aborts on the first failing project, and `packages/core` runs
*after* `packages/cli` in that order — compare a green run of #215
([31703023852](https://github.com/actana/control/actions/runs/31703023852)),
where the log continues `packages/core test: Test Files 53 passed (53)` and then
`packages/panel`. On this branch's run it stops at `cli`. So `packages/core`
never started, and **none of the 69 tests this PR is built on have been run by
CI** — nor has the rest of `packages/core`, which is the package the diff
actually changes.

The checklist comment says "CI is the arbiter" for the one unchecked item. It
has not arbitrated anything here. Combined with the stated local caveat — the
disk hitting 100 % mid-session, four `packages/panel` suites not re-verified —
the only evidence these suites pass is a local run on a machine that then ran
out of disk.

Nothing to change in the code for this. What is needed is a run in which
`packages/core` is green: re-run the workflow, and if `events-tail-cursor`
keeps timing out under load, de-flake it separately so it stops masking the
packages behind it. I'll approve on a green run.

### Verified by reading, claim by claim

- **Listing is on the HTTP surface, not `dirList`.** `CORE_LINK_PROTOCOL_VERSION`
  is `"0.15.0"` at `packages/sdk/src/core-link-frames.ts:1302` on both the merge
  base and this head; the diff contains no occurrence of `dirList` or of the
  protocol constant. No Core becomes needs-update. `parseRoute`
  (`core-files-routes.ts:660`) grows a fifth segment and nothing else does.
- **The five fields, and the sha256 decision.** `FileListingEntry`
  (`files-listing.ts:32-50`) carries `path, kind, size, mtime, mode, sha256` in
  the tar manifest's units, asserted field-for-field at
  `core-files-listing-routes.test.ts:122-146`. ADR 0027 **D6** is a real decision
  with both rejected alternatives written down beside it, and the recorded
  rationale matches the code: D4's premise is "free while the bytes are in hand",
  `listTree` never has the bytes (`describeLeaf`, `files-listing.ts:262-291`), so
  the same rule inverts. F10 can rely on it — the field is present in every entry
  either way (`core-files-listing-routes.test.ts:250-259` asserts the key set is
  identical with and without `?sha256=1`), so a future diff endpoint asks for
  digests over a surface that already streams rather than feature-detecting a
  missing key. The sharpest part of D6 is `files-listing.ts:268-280`: a file whose
  bytes could not be read *when they were asked for* is a `skipped` line, not an
  entry with a null — which is exactly the ambiguity that would otherwise reach a
  diff as "unchanged".
- **Confinement, independently.** `files-listing-confinement.test.ts` imports no
  confinement helper — the imports are `crypto`, `fs`, `http`, `os`, `path`,
  `vitest`, the route factory and the fixture. Every one of the 15 cases enters
  as an HTTP request. Absolute (including one pointing back inside), `..`
  (leading, buried, and cancelling-out), symlink-out (the link itself and an
  innocent segment beneath it), NUL, backslash, and a Project whose own root runs
  through a symlink. The walk-specific rule is the one that earns the file: at
  `files-listing-confinement.test.ts:271-277` every emitted path is re-resolved
  the way a client would resolve it and checked to land inside.
- **The gate is checked before the request.** `listIfOffered`
  (`files-listing-capability-gate.test.ts:167-175`) consults the real
  `readFilesCapability` and returns before `fetch` exists; the assertion is
  `expect(arrived).toEqual([])` against a Core-side counter, and the strongest
  case (`:228`) mounts the routes anyway and proves with a direct 200 that
  they *would* have answered. That is "not asked", not "refused".
- **Nothing buffers, either side.** `listTree` is an async generator holding only
  the ancestor chain (`files-listing.ts:149-224`), `opendir` not `readdir`
  (`:166`), and the route writes each line straight out with `drained`
  backpressure (`core-files-routes.ts:290-291`). The tests are real, not
  ceremonial: first line parsed while `res.complete === false`
  (`core-files-listing-routes.test.ts:354-376`), 4001 entries arriving in more
  than one chunk (`:378-388`), and the leak assertion via `/proc/self/fd` on a
  400-deep tree with the client parked in `res.pause()` so the handler is
  suspended inside `drained` (`:393-437`). Both `/proc` tests carry a positive
  control — `sawOneInFlight` / `sawSomethingOpen` — so a `/proc` shape change
  cannot make them vacuously green. That is the detail that makes me believe them.
- **Phase-4 sync reserved.** No cursor, no `since=`, no stored state, no
  comparison; the only occurrences of "diff endpoint" in the diff are prose in
  ADR 0027 and two doc comments.
- **Boundary held.** `git diff --name-only` against the base names nothing under
  `packages/sdk/` or `packages/shared/`. The capability suite imports
  `@actana/sdk/core-link-frames`, which is long-standing precedent in
  `packages/core` (a dozen existing suites do it) and consumes the reader rather
  than adding a client. #217's surface stays #217's.

Two smaller things, neither blocking:

- `core-files-routes.ts:322` — a mid-stream failure goes out as
  `{"type":"error","code":"read-failed"}`, matching the write path at `:534` and
  ADR 0028. It is the one line type the listing section of `docs/external-api.md`
  does not show: the sample block at `docs/external-api.md:104-107` lists
  `entry`, `skipped` and `done` only. #167's reader has to handle a stream that
  ends in `error` with no `done`, so it is worth a line in the doc and a test.
- `files-listing.ts:208-216` — the `directory-cycle` skip is unreachable without
  a bind mount, so it is the one branch here with no test. Cheap defensive code
  and I would not add a privileged test for it; noting it so it is a choice
  rather than an oversight.

The prose in the ADR, the commit messages and the route comments is unusually
good — the `?list=1` reasoning at `core-files-routes.ts:24-29` and the
"symlinks are entries, never followed" argument at `files-listing.ts:108-122`
both say *why* in a way that will survive the person who wrote them.

qcentic-dev:
> ### 🚩 Blocking — CI never ran `packages/core`
> What is needed is a run in which `packages/core` is green.

**There is one now.** Run
[31708936776 attempt 2](https://github.com/actana/control/actions/runs/31708936776/attempts/2)
at head `0054c7c` — **success**, every job, nothing skipped that was not skipped
by design (the train-tag and manifest-publish jobs).

You were right about the mechanism and it was worth insisting on: the run really
did abort at `packages/cli` and `packages/core` really had never been executed by
CI. Attempt 1 of that same run reproduced it exactly — same test, same 60 s
timeout, same `ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL`, `packages/core` never started.

The failure is [#208](https://github.com/actana/control/issues/208), open and
owned there: `events-tail-cursor.test.ts > resumes where the last run stopped`
times out under runner load. It is not this branch's, it is not touched here, and
de-flaking it inside a listing PR would be the wrong place to fix it — so the
answer here was a re-run, and the fix belongs to #208.

What attempt 2 proves, from its log rather than from a claim:

```
packages/core test:  ✓ src/__tests__/core-files-listing-routes.test.ts (28 tests) 2012ms
packages/core test:  Test Files  57 passed (57)
packages/core test:       Tests  1065 passed (1065)
packages/panel test: Test Files  130 passed (130)
packages/panel test:      Tests  1338 passed (1338)
```

So every package behind the abort ran: root `474`, `packages/sdk` `182`,
`packages/shared` `180`, `packages/cli` `297`, `packages/core` `1065`,
`packages/panel` `1338`. That covers the two things the local caveat left open —
the 69 tests this PR is built on, and the four `packages/panel` suites the disk
took out mid-session. The "not claimed" box in the checklist is claimed now, and
by CI rather than by this machine.

One caveat stated plainly, because it is the honest shape of a re-run: #208 is a
load-dependent timeout, so a future run on this branch can go red on it again
without anything in the diff changing. That is an argument for #208 being fixed,
not for reading this run as anything other than what it is — `packages/core`
green at `0054c7c`.

qcentic-dev:
> `core-files-routes.ts:322` — a mid-stream failure goes out as
> `{"type":"error","code":"read-failed"}` … It is the one line type the listing
> section of `docs/external-api.md` does not show … #167's reader has to handle a
> stream that ends in `error` with no `done`, so it is worth a line in the doc and
> a test.

Fixed, both halves. Marked "not blocking" in the review and taken as blocking
anyway, because the consumer is a different PR and a doc gap between two PRs is
the kind that is discovered by the reader hitting it.

**Doc — `7789f55`** (`docs/external-api.md`, the Listing section). A second sample
block beside the `entry`/`skipped`/`done` one, showing a stream that ends in
`error`, and the rule the sample cannot carry on its own:

- `error` and `done` are mutually exclusive. A `skipped` line costs one path and
  the walk continues; `error` **is** the walk stopping.
- `done` is therefore the only proof a listing is complete. End-of-stream without
  it is a valid prefix of the listing, never the listing.
- Which is the sentence #167's reader needs: treating a missing `done` as an empty
  tail is exactly how a truncated tree gets mistaken for a whole one.

**Test — `0054c7c`** (`core-files-listing-routes.test.ts`, four cases, all four
green in CI at attempt 2). The status line is already `200`, so it asserts the
failure travels as the last line of the body; that there is **no** `done` line
after it; that the entries produced before the fault are still valid NDJSON; and —
the contrast that gives the line its meaning — that an unreadable directory is a
`skipped` line which leaves `done` exactly where it was.

The fault is injected, and it is worth saying where and why, since every other
suite in this ticket runs against a real filesystem. Everything a walk can blame
on a single path is already a `skipped` line, tested for real in
`files-listing.test.ts`; what reaches line 322 is the directory *read* failing
part-way — an `EIO`, an `ENFILE`, a mount going away — which no kernel will
produce on request. So the injection is at `opendir`'s iteration, the exact
syscall that raises it in production, and nothing else is faked: real server, real
tree, real entries out of the real handle until the fault lands. It is visible in
the assertions rather than hidden in a helper — the tree has four files and the
test asserts two entries, so a mock that stopped firing fails the case instead of
passing it quietly.

qcentic-dev:
> `files-listing.ts:208-216` — the `directory-cycle` skip is unreachable without a
> bind mount, so it is the one branch here with no test. Cheap defensive code and I
> would not add a privileged test for it; noting it so it is a choice rather than
> an oversight.

Confirming it as a choice, and putting the reasoning on the record rather than
leaving it as agreement — no code change, and no test in `0054c7c` for this branch.

**Why a real one needs a bind mount.** The branch fires when a directory contains
one of its own ancestors *without a symlink* — the walk never descends through a
link, so the only ways to build the loop are `mount --bind` (Linux, `CAP_SYS_ADMIN`)
or a hardlinked directory, which no filesystem this Core targets permits. A test
for it would need root, would be skipped on every developer's machine and on the
macOS runners, and would leave a live mount behind if the process died mid-test —
a suite that can wedge the machine it runs on, in exchange for covering four lines
that do nothing but not-recurse.

**The tempting middle option, and why I did not take it.** This PR now has a
fault-injection seam — `0054c7c` fakes `opendir`'s iteration to reach the
mid-stream `error` line. The same lever would reach this branch: hand the walk
`lstat` results with a repeated `dev:ino` and the real `ancestors.includes(id)`
check fires, emitting the real skip line. I would rather not, and the reason is
what the test would be worth rather than what it would cost. The mid-stream fault
is injected because the *route's* behaviour afterwards is non-obvious and is what
#167 consumes. Here, the code under test is `ancestors.includes(id)` — the
uncertainty is not whether `includes` works, it is whether a real bind mount
presents the identical `dev:ino` that the check assumes, and a fabricated inode is
precisely the thing that cannot answer that. A green test that asserts my own
premise back to me would make this branch look covered while leaving the only
question about it open.

So: known gap, argued rather than papered over, and the honest containment is that
the failure mode it guards is a hang rather than a wrong answer — the walk's other
loop, a symlink pointing at its own parent, *is* covered against a real filesystem
(`files-listing.test.ts`, "does not hang on a link to the directory holding it").
If you would rather have the injected version anyway, say so and it is a ten-line
addition to the suite in `0054c7c`; I did not want to add it silently under the
name of coverage.

qcentic-dev:
## Acceptance criteria, #166 — updated after review, at `0054c7c`

Supersedes the checklist above. Two commits since the review, both answering it:
`7789f55` (doc) and `0054c7c` (test). No production code changed — `git diff
5651acd..0054c7c` touches `docs/external-api.md` and
`packages/core/src/__tests__/core-files-listing-routes.test.ts` and nothing else.

### Done when

- [x] **A Project's tree can be listed to arbitrary depth, streamed as NDJSON, with
  `{path, size, mtime, mode, sha256}` per entry.** Unchanged since review, which
  verified it line by line. `core-files-listing-routes.test.ts` is now **28 tests**
  (was 24).
- [x] **The same confinement rules as the routes ticket apply, tested independently.**
  Unchanged. `files-listing-confinement.test.ts`, 15 cases, every one an HTTP request.
- [x] **A Core without the capability is not asked; the client checks `ready` first.**
  Unchanged. `files-listing-capability-gate.test.ts`.
- [x] **Listing a large tree does not buffer it in memory on either side.** Unchanged,
  including the `/proc/self/fd` handle-leak case with its positive control.

### Traps

- [x] **`sha256` — decided deliberately and recorded.** ADR 0027 D6, unchanged.
- [x] **Phase-4 sync is reserved, not built.** No cursor, no `since=`, no state, unchanged.

### From the review

- [x] **The mid-stream `error` line is in the doc.** `7789f55` — `docs/external-api.md`,
  Listing section: a second sample block ending in
  `{"type":"error","code":"read-failed"}`, and the rule that `error` and `done` are
  mutually exclusive, so `done` is the only proof a listing is complete. #167's reader
  now has a written rule for a stream that ends without one.
- [x] **And it is tested.** `0054c7c` — four cases: the failure as the last line of a
  spent `200`; no `done` after it; the entries before the fault still valid NDJSON; and
  a `skipped` line leaving `done` where it was, which is the contrast that gives `error`
  its meaning. Fault injected at `opendir`'s iteration — the one failure a real tree
  cannot be asked for; [reply above](#issuecomment-5281649164) has the reasoning.
- [x] **`directory-cycle` skip — answered, still untested, argued rather than deferred.**
  A real cycle needs `mount --bind` and root; the injectable version would assert my own
  premise about `dev:ino` back to me. [Reply above](#issuecomment-5281649439), including
  the offer to add the injected version if you would rather have it.

### The ticket's framing rules

- [x] **Base `feat/project-files`, not `main`.** Still based there; `origin/feat/project-files`
  has not moved (`git rev-list --left-right --count origin/feat/project-files...HEAD` → `0 7`),
  so nothing to merge in. Nothing merged, promoted or force-pushed — both new commits went up
  as a fast-forward.
- [x] **Conventional Commits on the title and all 7 commits.** `commitlint --from 3c3d6bb --to HEAD`
  exits 0. Both new subjects are under 112 characters.
- [x] **Built on #215, not reworking it.** Unchanged.
- [x] **`packages/sdk` untouched.** Still true at `0054c7c` — the diff against the base names
  nothing under `packages/sdk/`.

### Previously not claimed — now claimed, by CI

- [x] **A full green run, with `packages/core` in it.** Run
  [31708936776 attempt 2](https://github.com/actana/control/actions/runs/31708936776/attempts/2)
  at `0054c7c`: **success**, all jobs. `packages/core` **57 files / 1065 tests**,
  `packages/panel` **130 files / 1338 tests**, `packages/cli` 297, `packages/sdk` 182,
  `packages/shared` 180, root 474. The blocker was real and the abort reproduced on attempt 1
  — same `events-tail-cursor` timeout, [#208](https://github.com/actana/control/issues/208),
  re-run rather than touched here.

Still a **draft** — this is yours to un-draft on approval, and nothing here merges anything.

qcentic-adm (approved):
## Re-review at `0054c7c` — the blocker is cleared, and it was cleared by evidence rather than by assertion

Both items from the last round are answered, the CI claim holds when I check the
run myself rather than take the summary's word for it, and nothing in the diff
moved that would put the earlier verification back in play. **Approving.**

Still a draft, and I am not un-drafting it — that stays yours.

### 🟢 Blocking item cleared — `packages/core` has now actually run

Run [31708936776](https://github.com/actana/control/actions/runs/31708936776) is
the only run at `0054c7c`, and attempt 2 is `success` across the whole required
set — 18 jobs green, 4 skipped by design (`*/Publish the multi-arch manifest`,
`Core image (train)`, `Panel image (train)`, `Resolve train tags`). Nothing
pending, nothing neutral.

What I actually cared about is that the abort no longer hides the package under
review, and it does not. Attempt 2's `Unit Tests` log runs all six workspace
projects to completion:

```
packages/sdk    10 files / 182 passed
packages/shared 14 files / 180 passed
packages/cli    24 files / 297 passed
packages/core   57 files / 1065 passed
packages/panel  130 files / 1338 passed
root            16 files / 474 passed
```

The four suites this ticket rests on are in that list by name, green, and with
the new cases in them: `core-files-listing-routes.test.ts` **28 tests** (was 24
— all four new cases ran; none skipped, the runner is not root),
`files-listing.test.ts` 22, `files-listing-confinement.test.ts` 15,
`files-listing-capability-gate.test.ts` 8.

**And the re-run masked nothing — I checked attempt 1 rather than assuming.** Of
the 22 jobs in attempt 1, exactly one failed: `Unit Tests`. Its only failing test
is `packages/cli/src/__tests__/events-tail-cursor.test.ts > actana events tail,
across a Core restart > resumes where the last run stopped, and does not replay
it`, `Error: Test timed out in 60000ms.` at `:125:3` — 60004 ms against a 60 s
budget, file duration 61654 ms. That is [#208](https://github.com/actana/control/issues/208)'s
signature down to the line number, and it matches the overshoot pattern the issue
records (61.7 s / 61.7 s / 63.1 s) rather than a deadlock sitting on the deadline.
Every other test in `packages/cli` passed — 296 of 297 — as did `sdk`, `shared`
and root before it.

So attempt 1 aborted at `packages/cli` the same way the reviewed run did
(`[ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL]`), which means it could not have concealed
a `packages/core` result: it never produced one. Attempt 2 is a strict superset
of attempt 1's coverage, not a second roll of the same dice on a failure that
might have been real. The one job that changed verdict changed it on a test in a
package this diff does not touch, and the packages behind it went from unrun to
green. That is the run I said I would approve on.

Your caveat is the right one to state and I will state it back: #208 remains
load-dependent, so this branch can go red again on it without the diff changing.
That is an argument for #208, not against this run.

### The mid-stream `error` line — doc and test both land

**Doc (`7789f55`).** In the Listing section, immediately after the
`entry`/`skipped`/`done` sample and before the `sha256` subsection — where a
reader working through the listing surface meets it in order, not appended
somewhere it would be found only by search. The sample's shape matches what the
route actually emits at `core-files-routes.ts:322`
(`{type:"error", code:"read-failed", message}`), field for field.

The prose does the part a sample cannot: `error` and `done` are mutually
exclusive, `skipped` costs one path while `error` **is** the walk stopping, and
therefore `done` is the only proof a listing is complete. That last sentence is
the one #167 / PR #217's `project.files.list` needs — a pull-driven generator
over this stream has no other way to tell "the tree" from "a prefix of the tree",
and the failure mode if it guesses is silent, not loud.

**Tests (`0054c7c`).** I read all four, and they exercise the reader-facing
contract rather than the emitting line:

1. the failure arrives as the **last line of a spent `200`**, `content-type`
   still `application/x-ndjson`, `code: "read-failed"`, message carrying `EIO`;
2. **no `done` line**, and exactly one `error` line — the assertion PR #217
   depends on, stated as the absence it actually is;
3. the entries produced **before** the fault are still valid NDJSON and still
   correct (`kind`, `size`, path shape), which is what makes a truncated stream
   usable rather than merely detectable;
4. the contrast: an unreadable directory is a `skipped` line and `done` is still
   there with `skipped: 1`.

Case 4 is the one that earns the set. Without it the suite would prove the error
line exists; with it, it pins the boundary a reader will otherwise get wrong,
since `skipped` and `error` are the two ways a listing reports trouble and only
one of them invalidates the tree.

On the injection, which is the only part that needed defending: the spy sits on
`fs.promises.opendir`'s iteration, and that is the real production call site —
`files-listing.ts:166` opens with `fs.promises.opendir` and `for await`s it, with
the *open* failing already handled as a `skipped`/`unreadable-directory` line
(`:54`). So the injected fault enters the handler by the same door an `EIO` or
`ENFILE` would, and the branch it reaches is the real `catch` at
`core-files-routes.ts:315-323`, not a stand-in. Real server, real tree, real
entries out of the real handle until the fault lands. The property I like most is
that the tree has four files and the test asserts **two** entries: a mock that
silently stopped firing fails these cases rather than passing them quietly, which
is the usual way an injected-fault suite rots into decoration.

### `directory-cycle` — the argued position, and I accept it

You put it on the record as a decision instead of taking the agreement, which is
what I asked for, and the reasoning holds on the one point I could check
independently rather than infer.

The branch is `ancestors.includes(id)` at `files-listing.ts:206-216` — four lines
that decline to recurse. Reaching it for real needs a directory containing its own
ancestor with no symlink in the path, which means `mount --bind` and
`CAP_SYS_ADMIN`, since the walk never descends through a link. A suite that needs
root, skips on every dev machine and every macOS runner, and can strand a live
mount if the process dies mid-test is a bad trade for four lines that do nothing
but not-recurse.

The stronger half is your reason for declining the injected version, and I agree
with it. The seam `0054c7c` adds would reach this branch by handing the walk a
repeated `dev:ino`. But the code under test would then be `includes` — the actual
open question is whether a real bind mount presents the identical `dev:ino` the
check assumes, and a fabricated inode answers that question by assuming it. A
green test there would convert an honest gap into a false one. Declining coverage
because the available coverage would be circular is the right call, and it is
rarer than it should be.

Containment checks out too: the failure mode guarded is a hang, not a wrong
answer, and the reachable loop is covered against a real filesystem —
`files-listing.test.ts:235` "does not hang on a link to the directory holding it",
plus `:228` for a link at itself. So the untested branch is the unreachable one,
with the reachable sibling tested for real. No injected version needed; I am not
taking you up on the offer.

### Nothing else regressed

`git diff 5651acd..0054c7c` touches `docs/external-api.md` and
`packages/core/src/__tests__/core-files-listing-routes.test.ts` and nothing else —
134 insertions, 1 deletion, and the deletion is the `vitest` import line growing
`vi`. **No production code changed since the review**, so every claim I verified
by reading last round stands on identical bytes. The two I re-checked at the new
head anyway:

- **Independent confinement.** `files-listing-confinement.test.ts` is untouched
  by both commits, and its import list at `0054c7c` is still `crypto`, `fs`,
  `http`, `os`, `path`, `vitest`, the route factory and the fixture — no
  confinement helper, so all 15 cases still enter as real HTTP requests and prove
  the rule rather than inherit it.
- **No buffering.** Intact at head, positive controls included: the first line
  parsed while `res.complete === false` (`:362-365`), `res.chunks > 1` on 4001
  entries (`:380-388`), and the `/proc/self/fd` leak case with `res.pause()`
  parking the handler inside `drained` and `sawOneInFlight` asserted `true`
  (`:412-437`) so a `/proc` shape change cannot make it vacuously green.

One thing I looked at because the new suite mutates global `fs`: the injecting
describe is last in the file and carries `afterEach(vi.restoreAllMocks)`, so the
streaming and leak assertions above it run against the real `opendir`. The CI run
agrees — all 28 pass together.

### Verdict

Approved. The one blocker was that CI had never executed this diff; it has now,
green, with `packages/core` in the run, and the re-run's failure was the known
flake in a package this PR does not touch rather than something the re-run buried.
Both non-blocking notes came back as a doc change in the right place, four tests
that pin the reader's contract, and a reasoned decline with the gap named. Ready
to un-draft when you are.

PR #217, opened 2026-08-13
feat(sdk): project.files.list, upload and download — streams both ways, progress as an async iterable

qcentic-dev:
## Summary

Phase 3 step 2 of #150: the SDK half of **F12**. `client.project(id).files` gives `list`, `upload` and `download` over the `/v1` routes PR #215 landed — **streams in and out, progress as an async iterable**.

The property that ties all three together is that **nothing runs ahead of its consumer**. An async iterable that eagerly drains its source yields the right values in the right order and has the whole transfer in memory while doing it; the streaming is then decorative. So these are pull-driven generators, and a slow consumer becomes a slow socket reader and then a paused writer on the Core. `download` returns a stream and the result has no method that hands over bytes, which is what makes "a gigabyte must not materialise" a shape rather than a promise.

This is also the SDK's **first HTTPS surface** — everything before it was the core-link WebSocket.

### Three things a reviewer should read closely

**The frozen `fetch` shape held, but not the way the spike wrote it down.** #151/PR #189 froze the dispatcher as the only seam (`fetch` ignores `options.cert`/`options.key`) and that is reused verbatim rather than rediscovered. What the spike could not have found: the `fetch` must be **undici's own, not the global one**. Node embeds its own copy of undici, and a `Dispatcher` only satisfies the implementation it came from — undici 8's `Agent` handed to Node 24's built-in `fetch` fails with `InvalidArgumentError: invalid onRequestStart method`. The spike loaded `ws` and `undici` as a matching pair from one directory and used the global `fetch`, which happened to agree at those versions. Importing both halves from one package makes that skew impossible rather than merely absent today, and `core-files-mtls.test.ts` pins both directions.

**`undici` is this package's second dependency**, for the identical reason `ws` is its first: Node's own global cannot present a client certificate. It is a public registry package, which is what the boundary rule is actually about — `@actana/shared` and `@actana/core` are private, so importing one would make the published SDK unresolvable (ADR 0025 D4) or a cycle (D2). The allowlist is now pinned against `package.json`, because an import that is allowed but never declared passes every test here and breaks in a stranger's install.

**A third boundary this package needed and did not have.** Inside the workspace the SDK is consumed as TypeScript source that plain `node` type-strips with no bundler and no loader — `tsconfig.build.json` says so, and `examples/*.mjs` depend on it. Type stripping refuses constructs that `tsc --noEmit`, vitest and eslint all accept, so nothing else in the repo notices. `plain-node-consumption.test.ts` caught a real one on this branch: a `constructor(readonly reason: string)` parameter property that typechecked, passed every other suite, and made the package unloadable by the examples sitting beside it.

## Related issues

Refs #167, #129, #150, #151, #166

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## How was this tested?

- [x] Manually verified (against the live Core on this machine — see below)

**47 new tests across eight suites**, over real loopback sockets against the Core's own route handler. The properties this ticket is about live below the frame level — chunked encoding, a body arriving in pieces, a socket that backpressures, a status line coming back while a request body is still going out — and an in-memory socket pair has none of them.

| Suite | Covers |
| --- | --- |
| `core-files-backpressure.test.ts` | a deliberately slow consumer, asserting on **how much the producer did** rather than on what arrived — the eager implementation passes any test that only checks values. Both directions (progress/listing in, upload body out), stream teardown when the caller walks away, and the end-to-end version: the Core stops mid-unpack while the consumer dawdles over three lines |
| `core-files-streaming.test.ts` | **an actual gigabyte** crossing a socket while `arrayBuffers` and RSS are watched, plus a structural check that no buffering method is ever called on a response, plus `disturbed`/`locked` proving the body is handed over untouched |
| `core-files-conflict.test.ts` | two concurrent writes over a real socket → `409 transfer-in-progress` off the status line with nothing written for the loser, the lease releasing with the request, `directory-in-the-way` told apart by code, and a **timing-free proof that exactly one request is made** — a retry loop shows up as a second call whatever its backoff |
| `core-files-unavailable.test.ts` | the F9 gate against a Core with a **working file surface that never announces one**, asserting nothing was sent; that "not connected yet" and "no capability" give different reasons; and that the Core is connected, compatible and not needs-update |
| `core-files-mtls.test.ts` | a real mutual handshake with the Core's own `generateCertMaterial`, download and upload through a blob's PEM material, the spike's **CA-only control**, cert-but-no-bearer refused 401, and the global-vs-undici `fetch` skew |
| `core-files-surface.test.ts` | the three operations end to end: NDJSON progress with every `overwritten` named, a folder as one streamed tar with Project-relative paths, exec bit and mtime, refusals, and that nothing is sent until the iterable is consumed |
| `core-files-stream-error.test.ts` | the failure that arrives as the stream's **last line** rather than as a status code: an `error` line with no `done` (the shape #216 flagged) throwing `CoreFilesStreamError` with the Core's own code and message, a truncated line routed through the same taxonomy instead of a raw `SyntaxError`, and every case under an explicit deadline so a reader waiting for a `done` that never comes fails as a hang |
| `plain-node-consumption.test.ts` | every shipped module loading under plain `node`, and an example script running far enough to prove the entry points resolve |

Full runs green on this branch: **sdk 229**, cli 298, root scripts 473, shared 180, core 992. `pnpm typecheck`, `pnpm lint` (0 errors, 51 pre-existing warnings in untouched files) and `pnpm scan:secrets` clean; commitlint passes on every commit.

Three caveats, stated rather than buried. The root scripts suite has **one pre-existing failure on the author's machine** — `core-launcher.test.mjs`, which a real `/opt/actana` install makes red locally and which this branch does not touch; it passes in CI. The core suite showed **one failure on its first run and passed 992/992 on re-run** — the machine was under disk pressure at the time and this branch touches no Core code. The panel suite has **three files failing only under the full local run** (`project-dialog-core-switch`, `operator-auth`, `live-read-path`) that pass 46/46 in isolation; they are the known local flakes, and no Panel or CLI code reads the accessors this branch touched.

**Manually verified against the live Core on this machine** (`core_17504d4b830baffc`, protocol `0.15.0`), read-only and with no writes to any Project: it announces no `files` capability, `canUseFileRoutes()` is false, and `download` throws `CoreFilesUnavailableError` with the full reason and sends nothing. That is the F9 criterion against a genuinely older Core rather than a rig configured to look like one.

None. The surface is additive: `CoreClient` gains `project()` and `httpsBaseUrl`, and `CoreClientOptions` gains `filesFetch`. `CORE_LINK_PROTOCOL_VERSION` does not move.

One behaviour change worth naming: **`CoreClient.close()` now clears the connection's capabilities.** Without it `canUseFileRoutes()` went on answering `true` on a closed client — contradicting its own documented "false after a drop" and pointing an HTTPS request at a Core this client is no longer talking to. Nothing outside the SDK reads those accessors yet, so this is invisible to the Panel and the CLI.

## Scope

**No Core route was created or modified.** `packages/core` and `packages/panel` are untouched.

`packages/cli` carries one follow-up commit that is packaging, not behaviour. The SDK's new dependency followed it into the CLI's esbuild bundle, and undici does not survive being inlined — it reaches for `require("node:assert")` down a conditional path esbuild cannot see, so the binary threw `Dynamic require of "node:assert" is not supported` on first run while the build reported success. `ws` was already `external` for exactly this reason; undici joins it there and in `dependencies`. Caught by `scripts/__tests__/npm-publish.test.mjs`, which packs the tarball and runs `actana --version` — the only check that sees it, and one that lives in the **root** vitest suite rather than `pnpm -r test`. The dependency guardrail now pins `build.mjs`'s externals against the manifest so the next person cannot half-do it.

`list` reads the manifest shape PR #215 established (`{path, size, mtime, mode, sha256}`, NDJSON) and is tested against an in-process fixture serving exactly that, because **the Core's listing route is #166's and it runs in parallel**. The URL is a single overridable `listUrl` method, so agreeing with #166 when it merges is a one-line change and the parsing, laziness and error handling are unaffected. `sha256` is typed nullable so that #166's open question — eager or on request — is readable either way.

---


qcentic-dev:
## Acceptance criteria — issue #167

### Done when

- [x] **`upload` takes a stream and returns an async iterable of NDJSON progress entries, including every `overwritten`.**
  Takes a web `ReadableStream` or any async iterable of chunks — so a Node `Readable` from `fs.createReadStream` is accepted with no conversion and no `node:stream` import in the shipped package. Returns an async generator over the Core's progress lines: one `entry` per file carrying `{path, kind, size, mtime, mode, sha256, result}` where `result` is `written` or `overwritten`, then a `done` line with the totals. A mid-transfer failure is thrown as `CoreFilesStreamError` from the stream's last line, not a status code — the `200` was spent on the first entry.
  *Proof — the progress stream:* `core-files-surface.test.ts` — an overwrite named as `overwritten` on a single file, and a folder upload where pre-existing and new entries are told apart by that field.
  *Proof — the mid-transfer failure:* `core-files-stream-error.test.ts`, added after review, because the citation above was the wrong one: that suite contains no `CoreFilesStreamError` assertion, and the claim was true but unevidenced. This one drives the shape #216 flagged — an `error` line and no `done` — over a real socket with the response chunked, and asserts the Core's own `code` and message reach the caller as `CoreFilesStreamError`, promptly: every case runs under an explicit deadline, so the failure mode where the reader waits for a `done` that is never coming fails as a hang rather than as somebody else's timeout. A truncated line now surfaces the same way instead of as a raw `SyntaxError` (`parseLine`, `core-files.ts`), which is what makes the documented single `catch (e) { e instanceof CoreFilesError }` sufficient; reverting that wrap fails three of the new tests.

- [x] **`download` returns a stream, never a buffer. A gigabyte file must not be materialised in memory.**
  The response body is handed through untouched. The result type has **no** `bytes`/`buffer` field, so the only way to the content is to consume `stream` — a caller who wants it whole writes that themselves and owns the decision.
  *Proof:* `core-files-streaming.test.ts` — an actual gigabyte crossing a real socket with `arrayBuffers` and RSS sampled throughout and held under 192 MiB; a structural test that `arrayBuffer`/`text`/`blob`/`bytes`/`json`/`formData` are never called on the response; and `disturbed`/`locked` showing the body is untouched when `download` resolves.

- [x] **`list` is an async iterable over the tree.** — *checked with one qualification, stated rather than buried.*
  The reader is complete and tested: an async generator over NDJSON manifest entries, `{path, size, mtime, mode, sha256}`, with `path`, `depth` and lazy start. **The Core route it will call is #166's, which runs in parallel, so nothing here creates or modifies one.** It is coded against the manifest shape PR #215 established and tested against an in-process fixture serving exactly that. The URL is a single overridable `listUrl` method, so agreeing with #166 on merge is a one-line change; `sha256` is typed nullable so #166's open question — eager or on request — is readable either way. **Live wiring lands when #166 merges.**
  *Proof:* `core-files-surface.test.ts` — the tree in manifest shape, a subtree to a depth, and nothing sent until consumed.

- [x] **The surface is unavailable, with a clear reason, against a Core that does not announce `files` on `ready` (F9).**
  The gate is read before every call and **before any byte leaves**. `CoreFilesUnavailableError` carries a reason that distinguishes the three situations, because flattening them to a boolean is what makes "unavailable" unactionable: not connected yet, client closed, and — the interesting one — a Core that announced no capability, which the message says explicitly is *a Core older than the surface, not a broken or out-of-date one*.
  *Proof:* `core-files-unavailable.test.ts` — all three calls refused against a Core whose file surface genuinely works but is never announced, with `requests == []`; different reason before connect; the link still connected, compatible and not needs-update. **Also verified against the live Core on this machine** (`core_17504d4b830baffc`, protocol `0.15.0`), which really does predate the surface: `canUseFileRoutes()` false, the full reason returned, nothing sent.

### Traps

- [x] **The SDK's first HTTPS surface — reuse the mTLS `fetch` shape the phase-1 spike froze, don't rediscover it.**
  Reused verbatim: `fetch` is undici, ignores `options.cert`/`options.key`, and the dispatcher is the only seam. D12 holds, so `engines: ">=22"` stands. One correction the spike could not have made: the `fetch` must be **undici's own, not the global one** — Node embeds its own copy and a `Dispatcher` only satisfies the implementation it came from (undici 8's `Agent` + Node 24's built-in `fetch` → `InvalidArgumentError: invalid onRequestStart method`). The spike loaded `ws` and `undici` as a matching pair and used the global `fetch`, which agreed at those versions.
  *Proof:* `core-files-mtls.test.ts` — a real mutual handshake with the Core's own certificate material, the spike's CA-only negative control, cert-but-no-bearer refused `401`, and both sides of the `fetch` skew pinned.

- [x] **Backpressure — an async iterable that races ahead of the consumer defeats streaming.**
  Every path is pull-driven: the NDJSON reader calls `read()` only when its buffer holds no complete line *and* the consumer has asked; the upload body is a `ReadableStream` pulling one chunk per `pull`; `download` hands the body through.
  *Proof:* `core-files-backpressure.test.ts` — a deliberately slow consumer with assertions on **how much the producer did**, not on what arrived. Reverting the reader to drain-then-yield fails three of these (verified by doing it), which is what makes them load-bearing. Includes the end-to-end case: the Core stops mid-unpack while the consumer dawdles over three lines.

- [x] **F8 surfaces as an ERROR, not a retry loop.**
  `CoreFilesConflictError` carries `status` and the machine-readable `code`, and **nothing in this package sleeps, backs off or re-sends** — reads included, deliberately, so no call has hidden attempts behind it.
  *Proof:* `core-files-conflict.test.ts` — two concurrent writes over a real socket refused off the status line with nothing written for the loser; the lease released with the request so the conflict is never permanent; `directory-in-the-way` told apart by code; and a timing-free proof that **exactly one request** is made against a sender that answers `409` to everything.

### Boundary

- [x] **Issue #166 owns the Core-side listing route — no Core route created or modified.**
  `packages/core` and `packages/panel` are untouched on this branch — no Core route is created or modified, and `files-rig.ts`'s `serveListing` is a clearly-labelled in-process #166 stand-in rather than a route. Verifiable: `git diff --stat origin/feat/project-files...HEAD` lists `packages/sdk/**`, `packages/cli/**`, `docs/external-api.md` and `pnpm-lock.yaml`.

  **`packages/cli` is not untouched, and this bullet used to say it was.** One packaging commit marks `undici` `external` in the CLI's esbuild bundle, as `ws` already was: an inlined copy throws `Dynamic require of "node:assert" is not supported` the first time `actana` runs, while the build reports success. It is not a behaviour change to the CLI, it is the cost of the SDK taking a second registry dependency, and `Unit Tests` in CI is where it was caught. The follow-up comment below has the detail.

qcentic-dev:
Follow-up on the checklist above: the boundary line ("no Core route created or modified") still holds — `packages/core` and `packages/panel` are untouched — but the branch now carries one **packaging** commit in `packages/cli`, which the Scope section above describes. It is not a behaviour change to the CLI: undici had to be marked `external` in its esbuild bundle, as `ws` already was, because an inlined copy throws `Dynamic require of "node:assert" is not supported` the first time `actana` runs while the build reports success. CI is green on it — including `Unit Tests`, which is where it was caught.

qcentic-adm (approved):
## Review against #167 — approving. Every clause holds at the enforcing line, and the evidence is the discriminating kind

I read all four "done when" clauses and all three traps against the code that
enforces them rather than against the prose that describes them, ran the
frozen-shape comparison against #151's recorded finding, and checked CI myself
at the head SHA. CI is green at `297c50f` — all 22 checks, `Unit Tests`,
`Typecheck`, `Lint`, `Conventions`, `Dependency Audit` and both installer E2Es
included, run
[31707356214](https://github.com/actana/control/actions/runs/31707356214). The
local flakes disclosed in the body are disclosed accurately and CI settles all
three.

The thing that makes this suite worth trusting is that it assumes the eager
implementation is the one that passes. `core-files-backpressure.test.ts`
asserts on `produced()` — how far the producer ran — rather than on what the
consumer received, which is the only assertion an eagerly-draining generator
fails. `core-files-streaming.test.ts` does the same trick twice: a structural
check that names the six buffering methods, and an actual gigabyte with
`arrayBuffers` sampled across it. That is the difference between testing the
property and testing the pleasant consequence of the property.

### Claim by claim

**The mTLS shape is #151's, reused, not rediscovered.** `core-files-http.ts:197`
is `new Agent({ connect: { ca, cert, key } })` and `:220` is
`undiciFetch(url, { …, dispatcher })` — the spike's recorded incantation
character for character, dispatcher as the only seam. Nothing in the shipped
path imports `node:tls`, `node:https` or calls `https.request`; the one
occurrence of that string is the comment at `:21` explaining why not. The
spike's leg 2b survives as a real control: `core-files-mtls.test.ts` refuses a
CA-only `Agent` and matches the refusal against a pattern that deliberately
excludes `ECONNREFUSED`, so it cannot pass against a server that is not there.
The undici-own-`fetch`-vs-global finding **extends** the frozen shape rather
than contradicting it — the dispatcher is still the only seam, D12 still holds,
`engines: ">=22"` still stands — and both directions of the skew are pinned.

**`upload` takes a stream, yields NDJSON progress, names every `overwritten`.**
`CoreFileSource` accepts a web `ReadableStream` or any async iterable, so an
`fs.createReadStream` needs no conversion and the package needs no `node:stream`
import. `core-files.ts:288` sets `result` per entry, and
`core-files-surface.test.ts` proves it on both the single-file overwrite and the
folder tar where pre-existing and new entries are told apart by that field —
plus Project-relative paths, the exec bit, the mtime, and nothing sent until the
iterable is consumed.

**`download` returns a stream and a gigabyte never materialises.** I hunted the
whole success path for hidden buffering — `arrayBuffer`, `text`, `json`, `blob`,
`bytes`, `formData`, `Buffer.concat`, any accumulate-into-array. The only
buffering call in the module is `refusalFrom`'s `res.json()` at
`core-files-http.ts:239`, which is on the non-2xx path against a short JSON
refusal body and is correctly reasoned about in place. `download` is
`stream: res.body` at `core-files.ts:330` — the body handed over untouched, with
`disturbed`/`locked` asserted to prove it.

**Backpressure is real, and there is no internal buffer defeating it.** I read
`ndjsonLines` (`core-files.ts:395-425`) specifically looking for the queue that
would undo this. There is none: the only state is a `buffered` string, and
`reader.read()` at `:410` is reached **only** when the buffer holds no complete
line, with the `yield` at `:407` sitting between them as the suspension point.
`sourceStream` (`:435`) is one `pull` → one `iterator.next()`, default
high-water mark of one. The tests bound the producer at `consumed + 2`, which is
exactly that one-in-hand allowance, and the real-socket case shows the Core
stopping mid-unpack — the whole chain, kernel included.

**The F9 gate fires before any HTTP request.** `requireAvailable` is the first
statement of all three methods (`core-files.ts:201`, `:248`, `:314`), ahead of
every `this.opts.fetch` call at `:202`, `:260`, `:315`. `filesAvailability`
(`core-client.ts:717`) asks connectivity *before* capability, which is the right
order and is argued for rather than left to chance. Three distinguishable
reasons, and `core-files-unavailable.test.ts` runs them against a Core whose
file surface genuinely works and simply never announces one, asserting
`requests == []`. The live-Core check against `core_17504d4b830baffc` is the
right kind of belt-and-braces.

**F8 is an error and there is no retry loop.** I grepped the file path for
`retry`/`backoff`/`sleep`/`setTimeout`/re-send: zero in `core-files.ts` and
`core-files-http.ts`. `DurableCoreClient`'s backoff is the WebSocket link and
never touches this path, and the default undici `Agent` does not retry
(`RetryAgent` is opt-in and unused). `core-files-conflict.test.ts`'s
exactly-one-request test is the right shape — timing-free, so no backoff can
hide from it — and extending the no-retry rule to reads is the correct call for
the reason given.

**The boundary held.** `packages/core` and `packages/panel` are untouched.
`files-rig.ts`'s `serveListing` is a clearly-labelled in-process #166 stand-in;
the write/read routes come from PR #215's handler already on the base branch,
which is consumption, not modification. `listUrl` as the single overridable seam
plus nullable `sha256` is a real live-wiring plan, not a promise to have one.
The `packages/cli` packaging commit is the right fix for a genuine failure mode
and is disclosed in the body — though see the note on the checklist below.

### The one gap worth naming before this leaves draft (non-blocking)

**`CoreFilesStreamError` has no test.** It is thrown at three sites —
`core-files.ts:216`, `:277`, `:323` — and asserted nowhere; `grep
CoreFilesStreamError packages/sdk/src/__tests__/` returns nothing. This is the
one path where the suite drops below its own standard, and it is specifically
the path #216's review handed to this ticket:

> `core-files-routes.ts:322` — a mid-stream failure goes out as
> `{"type":"error","code":"read-failed"}` … #167's reader has to handle a stream
> that ends in `error` with no `done`, so it is worth a line in the doc and a
> test.

The doc line landed. The test did not. I checked the behaviour myself instead of
assuming it, by running the reader's exact control flow against four stream
shapes:

| stream ends with | behaviour |
| --- | --- |
| `{"type":"error","code":"read-failed","message":…}`, no `done` | **throws `CoreFilesStreamError`, `code="read-failed"`, carrying the Core's own message.** Immediate — no hang, no swallowed success |
| entries then EOF — no `done`, no `error` | **completes normally and silently** |
| truncated mid-JSON-line | throws a raw `SyntaxError` from `JSON.parse`, not a `CoreFilesError` |
| normal `entry` → `done` | completes with both lines |

So **the flagged case is correctly handled** — row one is the one #216 raised,
and it behaves exactly as it should. The field names line up too: the Core emits
`{ type: "error", code, message }` at `core-files-routes.ts:438` and the reader
reads `record.code` / `record.message`, so the Core's prose reaches the caller
rather than being replaced by the generic fallback. That is a real match, not a
coincidence, and it is the part I most expected to find broken.

Rows two and three are the residue, and both are minor:

- Row two is a swallowed success in the literal sense, but barely reachable. A
  Core that dies mid-stream leaves a chunked response without its terminating
  chunk, which undici surfaces as a rejected `read()`; a clean `res.end()` with
  no `done` needs the error-line write to have failed first. Still — `upload`
  knows it must end on a `done` line, and asserting that would cost two lines
  and close the case properly.
- Row three: `core-files.ts:416`'s own comment anticipates the truncation ("a
  stream that ended mid-write may not have" a trailing newline) and then parses
  it unguarded, so the one shape the comment names escapes the module's error
  taxonomy. Wrapping the two `JSON.parse` calls into a `CoreFilesStreamError`
  would make one `catch (e) { e instanceof CoreFilesError }` sufficient.

Neither is an acceptance-criterion failure and neither warrants blocking. What I
would like before ready-for-review is the test — the mid-stream `error` line,
asserted end to end.

**One accuracy fix on the checklist comment**, because that comment is what a
future reader will trust:

- The `upload` bullet cites `core-files-surface.test.ts` as *Proof* for "A
  mid-transfer failure is thrown as `CoreFilesStreamError` from the stream's
  last line". That suite contains no such assertion. The claim is true — I
  verified it — but the proof citation is not, and on a PR this careful about
  the difference between a property and its evidence, that line stands out.
- The Boundary bullet still reads "`packages/core`, `packages/panel` and
  `packages/cli` are untouched on this branch" and gives a `git diff --stat`
  invocation as verification. `packages/cli` is no longer untouched. The
  follow-up comment says so and CI covers it, so this is disclosed rather than
  hidden — but the bullet itself should be edited so the two comments do not
  contradict each other.

### Verdict

**Approve.** Four of four done-when clauses met — `list` honestly qualified, and
the qualification is the issue's own parallel-ticket structure rather than
scope being quietly shed. Three of three traps met, each at the line that
enforces it. Boundary held, CI green at the head SHA, and the streaming and
backpressure claims are proven by tests that the wrong implementation would
fail. Add the `CoreFilesStreamError` test and fix the two checklist lines before
taking it out of draft.

The module prose is doing real work — `core-files-http.ts`'s account of the
two-copies-of-undici problem and `core-files.ts:395`'s explanation of why the
`yield` sits where it does will both outlive whoever wrote them.

PR #219, opened 2026-08-13
fix(sdk): point project.files.list at the Core's /files/list, and pin the seam with a contract test

### LINKED ISSUE #218: project.files.list targets ?list=1, but the Core serves /files/list — SDK and Core disagree after #216
Both halves of the file-listing feature have landed on `feat/project-files` (#217 SDK-side, #216 Core-side), and they disagree on the listing URL. Each side documented the choice; neither knew the other had picked the opposite.

**SDK** — `packages/sdk/src/core-files.ts`, `listUrl()`:
```
GET /v1/projects/:id/files?path=<p>&list=1
```
> "A query parameter on the existing route rather than a new path segment, because the Core's `parseRoute` matches `/v1/projects/:id/files` on an exact four-segment split … `protected` so a caller pinned to a Core that chose otherwise can subclass rather than wait for a release."

It also names this reconciliation as #166's job:
> "That URL is {@link listUrl} … when #166 merges, agreeing with it is a one-line change here."

**Core** — `packages/core/src/core-files-routes.ts`, header:
```
GET /v1/projects/:id/files/list?path=<p>&depth=<n>&sha256=1
```
> "Listing is a route of its own rather than a `?list=1` on the read above … A query parameter that a proxy, redirect or hand-edited URL can drop would turn 'list this folder' into 'download this folder', which for a `node_modules` is a mistake measured in gigabytes rather than in a 400. A path segment cannot be dropped silently."

## Impact

Against a real merged Core, the SDK's URL is a valid **read**, not a listing. `project.files.list()` on a directory receives a **streamed tar** and feeds it to the NDJSON line parser. The unknown `list=1` param is ignored, so there is no 400 — exactly the failure mode the Core-side comment describes.

## Why CI is green anyway

`packages/sdk/src/__tests__/files-rig.ts` mounts the real `createCoreFilesRequestHandler`, but intercepts `?list=1` with a hand-rolled stand-in **before** it:

```js
if (opts.listing && serveListing(req, res, root, projectId)) return;
if (routes.handle(req, res)) return;
```

`serveListing` bails on `if (!url.searchParams.has("list")) return false;`. So the SDK suite proves the reader against a fixture that answers the SDK's own URL, and the Core suite exercises the real route directly. Nothing crosses the seam, and nothing will go red when it breaks.

## Fix

- [ ] `packages/sdk/src/core-files.ts` — point `listUrl()` at `…/files/list?path=…`; the Core is the side with the safety argument, and `depth`/`sha256` are already query params there.
- [ ] `packages/sdk/src/__tests__/files-rig.ts` — drop the `listing` stand-in and let the real Core handler serve listings, so the suite covers the seam instead of shimming it.
- [ ] `docs/external-api.md` — remove the now-false paragraph: "**Listing is not on this table yet.** … the Core route that serves it is #166, and this row lands when it does." The route exists; give `project.files.list` its row.

Nothing is released — reads, writes and listing all ship on the same unreleased train and no Core announces the capability yet — so this is an integration-branch inconsistency, not a user-facing break. Worth closing before `feat/project-files` goes anywhere.

qcentic-dev:
Closes #218.

The SDK sent `GET /v1/projects/:id/files?path=<p>&list=1`; the Core serves
`GET /v1/projects/:id/files/list`. Both landed on `feat/project-files`, both sides
documented the choice, neither knew the other had picked the opposite.

## Which URL wins, on the two recorded rationales

**The Core's** — `…/files/list`.

The SDK's argument was about **cost**, and the cost has since been paid:

> A query parameter on the existing route rather than a new path segment, because the
> Core's `parseRoute` matches `/v1/projects/:id/files` on an exact four-segment split …
> a `…/files/list` leaf is a change to that parser, and a parameter is not.

That was true when it was written and is not true now. #216 shipped `parseRoute`
reading a fifth segment (`core-files-routes.ts`, `leaf: "files" | "list"`), along with
the rule that a listing is `GET`/`HEAD` only — so `PUT …/files/list` cannot fall
through and create a *file called `list`*. The parser change the SDK was avoiding is
already in the branch. The argument bought a saving that no longer exists.

The Core's argument was never about cost, and nothing has touched it:

> A query parameter that a proxy, redirect or hand-edited URL can drop would turn
> "list this folder" into "download this folder", which for a `node_modules` is a
> mistake measured in gigabytes rather than in a 400. A path segment cannot be
> dropped silently.

It also has the sharper claim of the two, because it describes a failure that is
**silent**. Dropping `?list=1` leaves a perfectly valid request that answers with the
wrong thing; there is no status code in which to notice. That is not a hypothetical
here — it is exactly how this bug behaved: the Core streamed a tar, the SDK fed it to
the NDJSON line parser, and the first sign of trouble was a parse error about a
character position in a record that was actually a tar header.

A cost argument that has been overtaken loses to a safety argument that has not.
`depth` and `sha256` are already query params on the Core's route, so the SDK's
options map onto it with nothing left over.

## The alignment

- `packages/sdk/src/core-files.ts` — `listUrl()` now builds `…/files/list?path=…`,
  and carries `depth` and `sha256`. It stays `protected`, so a caller pinned to an
  older Core can still subclass. The comment recording the losing rationale is
  replaced by one that says why it lost.
- `docs/external-api.md` — the stale paragraph ("**Listing is not on this table
  yet.** … the Core route that serves it is #166, and this row lands when it does")
  is gone; `project.files.list` has its row in the `project.files.*` table.

## The real deliverable: a test that could not have passed before

One line of the above is the defect. The rest of this PR is about why **both suites
were green** across two merged pull requests while the two halves disagreed.

They were green because neither test crossed the seam. `files-rig.ts` mounted a
hand-rolled listing stand-in *ahead of* the real Core handler, keyed on `?list=1`, so
the SDK's suite proved its reader against a fixture that answered the SDK's own URL by
construction. The Core's suite called its own route directly. Each side was checked
against its own idea of the other, so nothing either author could write would have
gone red.

- The stand-in is **deleted**. `startFilesRig` is now the Core's real handler and
  nothing else — its `listing?: boolean` option is gone.
- `packages/sdk/src/__tests__/files-list-contract.ts` dials `project.files.list()`
  against `createCoreFilesRequestHandler` in one process over a real loopback socket,
  and round-trips a seeded tree: paths and kinds, `depth`, `sha256` both asked for and
  not, and a 404 for a path that is not there rather than an empty tree.
- The sharpest case replays **the URL the client actually put on the wire** (read off
  the rig's request log, not written out again in the test) and asserts the Core
  answers it with `x-actana-transfer-kind: listing`. It asks the Core the question
  that matters — *given this URL, which handler answers?* — rather than asserting a
  constant equals itself.
- One case pins the hazard the Core's rationale names: listing a folder and
  downloading a folder must stay different URLs answering different things.

Verified red before the fix. With `listUrl()` reverted to `?list=1`, five of the six
cases fail, with the tar arriving at the NDJSON parser exactly as #218 describes:

```
CoreFilesStreamError: the Core's progress stream ended mid-line —
"a.txt … 0000644 0000000 0000"… is not a complete NDJSON record
```

(The sixth — a 404 for a missing path — passes either way; it is contract coverage,
not drift detection.)

### Where it runs

Registered in **both** packages' suites, by two one-line files:

| File | Runs under |
| --- | --- |
| `packages/sdk/src/__tests__/core-files-list-contract.test.ts` | `pnpm --filter @actana/sdk test` |
| `packages/core/src/__tests__/core-files-list-contract.test.ts` | `pnpm --filter @actana/core test` |

`pnpm test` — what CI runs — runs both. Putting it in one package would leave the
other package's author not running it before pushing, and either side can break this
contract. That is the arrangement that produced #218 in the first place.

## Two judgement calls beyond the issue's checklist

- **`sha256` is now a `CoreFileListOptions` field.** `CoreFileEntry.sha256` was
  documented as deliberately leaving open "whether it is computed eagerly or on
  request … that ticket's call to make". The Core made it: on request for a listing,
  eagerly for a transfer. Without a flag to send, the field could only ever be `null`
  through the SDK, so aligning the URL without it would leave the seam half-joined.
- **`packages/core/tsconfig.json` excludes the Core's registration file.** The shared
  spec lives in the SDK's test tree, which is written for the SDK's toolchain — ESM,
  `Bundler` resolution, relative imports carrying `.ts` so plain `node` can load them
  (`plain-node-consumption.test.ts` depends on that). Type-checking it from the Core
  would mean turning on `allowImportingTsExtensions` for the whole Core package and
  pulling the SDK's client and test rig into that program to serve one import. It is
  type-checked where it is written — the SDK's own `tsc` covers `src/__tests__` — and
  *run* from both, which is the point. `packages/core/vitest.config.ts` gains an
  `@actana/core` self-alias so the shared spec can reach the file routes by package
  name from either runner. Both are commented at the site.

## Checks

| Command | Result |
| --- | --- |
| `pnpm --filter @actana/sdk test` | 236 passed, 5 skipped (21 files) |
| `pnpm --filter @actana/core test` | 1071 passed (58 files) |
| `pnpm typecheck` | clean across all five packages |
| `pnpm lint` | 0 errors; no new warnings |

Nothing is released — reads, writes and listing all ship on the same unreleased train
— so this is an integration-branch inconsistency, not a user-facing break.

Base is `feat/project-files`, not `main`. Draft.

### DIFF
--- a/docs/external-api.md
+++ b/docs/external-api.md
@@ -184,6 +184,7 @@ const { stream } = await project.files.download({ path });
 
 | Property | Why it is that way |
 | --- | --- |
+| `list` is an async iterable of entries, from `GET …/files/list` | the same `{path, kind, size, mtime, mode, sha256}` lines the route streams, parsed. `path` and `depth` pass through, and `sha256: true` asks for digests — off by default, because a listing has no bytes in hand. `skipped` lines are passed over; an `error` line throws rather than ending the tree early |
 | `download` returns a **stream, never a buffer** | a gigabyte file must not be resident. There is no method on the result that hands over bytes, so a caller that wants the whole thing writes that themselves |
 | `upload` takes a stream and returns progress as an async iterable | one line per entry, each naming `written` or `overwritten`; a `done` line closes it. A mid-transfer failure is the last line, not a status code |
 | all three are **pull-driven** | nothing runs ahead of the consumer. A slow reader becomes backpressure on the socket and then on the Core, rather than a queue filling in the client |
@@ -195,10 +196,13 @@ The client presents its certificate through an undici `Agent` — `fetch` has no
 **undici's own `fetch`** rather than the global one, because a dispatcher only
 satisfies the undici implementation it came from and Node embeds its own copy.
 
-**Listing is not on this table yet.** `project.files.list` reads an NDJSON
-manifest of `{path, size, mtime, mode, sha256}` from the same origin; the Core
-route that serves it is [#166](https://github.com/actana/control/issues/166),
-and this row lands when it does.
+The client's listing URL and the route above are held together by a contract
+test that drives `project.files.list` against the Core's real handler in one
+process, registered in **both** packages' suites
+([#218](https://github.com/actana/control/issues/218)). They disagreed once —
+the client sent `?list=1` on the read route while the Core served
+`…/files/list` — and every suite stayed green, because each side was checked
+against its own idea of the other. A URL is not a fact either half owns alone.
 
 ## The Panel's own routes
 
--- a/packages/core/src/__tests__/core-files-list-contract.test.ts
+++ b/packages/core/src/__tests__/core-files-list-contract.test.ts
@@ -1,0 +1,16 @@
+// The Core's registration of the shared listing contract (#218).
+//
+// The body lives in the SDK's test tree — `packages/sdk/src/__tests__/
+// files-list-contract.ts` — because that is where the rig that stands this
+// surface up already lives, and it is reached here by package name through the
+// same test-only alias `@actana/sdk` already uses.
+//
+// It runs here as well as there on purpose. This package owns one half of the
+// listing URL and the SDK owns the other; the two disagreed for two merged pull
+// requests without a single red test, because each suite proved its own half
+// against its own idea of the other. A contract test that only the SDK's suite
+// runs is one this package's author does not run before pushing — which is the
+// arrangement that let #218 happen. Do not "tidy" this file away.
+import { describeFilesListContract } from "@actana/sdk/__tests__/files-list-contract";
+
+describeFilesListContract();
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -28,5 +28,15 @@
     }
   },
   "include": ["src/**/*.ts"],
-  "exclude": ["node_modules", "dist"]
+  // The listing contract (#218) is registered in both packages' suites and its
+  // body lives in the SDK's test tree, which is written for the SDK's
+  // toolchain: ESM, `Bundler` resolution, and relative imports carrying their
+  // `.ts` extension so plain `node` can load them. Type-checking it from here
+  // would mean turning on `allowImportingTsExtensions` for the whole Core and
+  // pulling the SDK's client and test rig into this program to serve one
+  // import. It is checked where it is written — `@actana/sdk`'s own `tsc`
+  // covers `src/__tests__` — and it is *run* here, which is the point of it.
+  // vitest resolves it through the alias in `vitest.config.ts` and does not
+  // consult this file.
+  "exclude": ["node_modules", "dist", "src/__tests__/core-files-list-contract.test.ts"]
 }
--- a/packages/core/vitest.config.ts
+++ b/packages/core/vitest.config.ts
@@ -13,6 +13,11 @@ export default defineConfig({
       // The core-link frames are in `@actana/sdk` (ADR 0025).
       "@actana/sdk": path.resolve(__dirname, "../sdk/src"),
       "@actana/shared": path.resolve(__dirname, "../shared/src"),
+      // This package, by its own name. Only the shared listing contract
+      // (#218) needs it: that spec is written for both suites, so it reaches
+      // the file routes the way the SDK's suite reaches them rather than by a
+      // relative path that would only resolve from one of the two.
+      "@actana/core": path.resolve(__dirname, "src"),
     },
   },
 });
--- a/packages/sdk/src/__tests__/core-files-list-contract.test.ts
+++ b/packages/sdk/src/__tests__/core-files-list-contract.test.ts
@@ -1,0 +1,8 @@
+// The SDK's registration of the shared listing contract (#218).
+//
+// The body is in `files-list-contract.ts` and the Core's suite registers the
+// same one, so a change to either half of the listing URL goes red on both
+// sides. See that file for why it is shared rather than written twice.
+import { describeFilesListContract } from "./files-list-contract";
+
+describeFilesListContract();
--- a/packages/sdk/src/__tests__/core-files-surface.test.ts
+++ b/packages/sdk/src/__tests__/core-files-surface.test.ts
@@ -4,7 +4,7 @@
 // Every assertion here goes over a real loopback socket to the Core's own route
 // handler, writing to a real directory — see `files-rig.ts` for why a fake
 // would prove nothing this ticket cares about.
-import { afterAll, afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterAll, afterEach, describe, expect, it } from "vitest";
 import { createHash } from "node:crypto";
 import * as fs from "node:fs";
 import * as path from "node:path";
@@ -225,15 +225,15 @@ describe("download", () => {
 });
 
 describe("list", () => {
-  beforeEach(() => {
-    // Every assertion in this block reads the #166 stand-in described in
-    // `files-rig.ts`: the manifest shape PR 215 established, served in process.
-    // The reader is what #167 owns; the route is #166's.
-  });
-
+  // Every assertion in this block goes to the Core's real listing route, like
+  // every other block in this file. It used to read a stand-in mounted ahead of
+  // the handler, which is how #218 stayed invisible; `files-rig.ts` says what
+  // that cost. Paths are sorted before comparison because the real walk hands
+  // entries back in the filesystem's order and promises no more than that —
+  // sorting a directory means holding all of it, which is what the walk exists
+  // not to do.
   it("is an async iterable over the tree, in the manifest shape", async () => {
     const core = await open({
-      listing: true,
       seed: { "a.txt": "aaa", "src/b.txt": "bb", "src/deep/c.txt": "c" },
     });
     const project = core.project(rig!.projectId);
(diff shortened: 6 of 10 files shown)


qcentic-dev:
## Issue #218 acceptance criteria

- [x] **`packages/sdk/src/core-files.ts` — point `listUrl()` at `…/files/list?path=…`.**
  Done, and `depth` / `sha256` go with it. The Core is the side with the safety
  argument; the SDK's was a cost argument that #216 has already paid off. The
  losing rationale in the comment is replaced with the reasoning that beat it.

- [x] **`packages/sdk/src/__tests__/files-rig.ts` — drop the `listing` stand-in and
  let the real Core handler serve listings.**
  `serveListing`, its `walk`, and the `listing?: boolean` option are deleted;
  `startFilesRig` is now `createCoreFilesRequestHandler` and nothing else. The two
  suites that passed `listing: true` (`core-files-surface`, `core-files-unavailable`)
  now go to the real route — which is why the surface suite's listing assertions
  changed: the real walk returns entries in filesystem order, so paths are sorted
  before comparison, and `sha256` is `null` unless asked for.

- [x] **`docs/external-api.md` — remove the now-false paragraph; give
  `project.files.list` its row.**
  Row added to the `project.files.*` table. The "Listing is not on this table yet …
  this row lands when it does" paragraph is replaced by a note that the two URLs are
  now held together by a contract test, and why that is needed.

### Beyond the checklist, as the ticket's framing asks

- [x] **A live contract test across the seam.** `files-list-contract.ts` dials
  `project.files.list()` against the real in-process Core over a real socket and
  round-trips a listing. Six cases: tree round-trip, URL-dispatch replay, list-vs-
  download URL separation, `depth`, `sha256`, missing-path refusal.
- [x] **It could not have passed while the two sides disagreed.** Verified by
  reverting `listUrl()` to `?list=1`: 5 of 6 fail, with the Core's tar reaching the
  NDJSON parser — the exact failure mode #218 documents.
- [x] **Both sides' CI runs it.** Registered from `packages/sdk/src/__tests__/` and
  `packages/core/src/__tests__/`; `pnpm test` runs both, and so does either package's
  suite on its own.

### Constraints observed

- [x] New worktree at `/home/core/work/wt-218`, cut from `origin/feat/project-files`
      (`10a5add`), branch `fix/files-list-url-contract`. All work confined to it.
- [x] `push -u`; no force push at any point.
- [x] Draft PR, base `feat/project-files` — not `main`. Not merged, not marked ready.
- [x] `main` untouched.
- [x] SDK suite: 236 passed / 5 skipped. Core suite: 1071 passed. `pnpm typecheck`
      and `pnpm lint` clean.

qcentic-adm (approved):
Reviewed against #218, at the enforcing lines rather than the prose. Approving on substance; one thing has to happen before merge and it is not a change to this diff — see the last section.

## The URL, applied consistently

Checked all four places and swept the tree:

- `core-files.ts` `listUrl()` builds `…/v1/projects/:id/files/list?path=…`, carries `depth` and `sha256=1`, stays `protected`. The new private `routeUrl(leaf)` puts origin + version prefix + id escaping in one place — a good incidental, since the two routes now differ only by leaf.
- The Core's route is unchanged and is the one being targeted: `parseRoute` returns `leaf: "list"` for the fifth segment (`core-files-routes.ts:660`), `parseListingQuery` reads `depth` and `sha256` with the same spellings the SDK sends (`:679`), and `handleList` answers `x-actana-transfer-kind: listing` (`:283`).
- `docs/external-api.md` — the "**Listing is not on this table yet** … lands when [#166] does" paragraph is gone, `project.files.list` has its row, and the replacement paragraph records why the two halves are now pinned.
- The losing module carries the winning rationale: `listUrl`'s comment states the old cost argument, says #216 paid that cost, and then gives the Core's safety argument in full. That is the right disposition — a reader who wonders "why not `?list=1`" finds the answer at the site that used to do it.

Sweep for `list=1` across the tree at head: six hits, all narrative describing the old state (`files-rig.ts:17`, `files-list-contract.ts:7,99`, `core-files.ts:339`, `external-api.md:203`, and the Core's own header at `core-files-routes.ts:24`, which is the winning side arguing its case). No live URL builds it. `#166` survives only in ADR 0027/0028 and in Core test headers, where it is a historical attribution rather than a promise of future wiring.

## The choice of URL

Sound, and for the reason given rather than by deference. The SDK's recorded rationale was explicitly a cost claim — `parseRoute` matched an exact four-segment split, so a leaf was a parser change and a parameter was not. #216 shipped that parser reading five segments and gating the leaf to `GET`/`HEAD` (`precheck`, `:188`), so `PUT …/files/list` is a 405 rather than a file named `list`. The cost is paid and the saving no longer exists. The Core's rationale was never about cost and nothing has touched it; it also has the sharper form, because the failure it names is silent — and #218 is that failure having actually happened. `depth` and `sha256` are already query params on the Core's route, so nothing in the SDK's option surface is left stranded. A superseded cost argument losing to an untouched safety argument is the right call on the recorded record.

## The contract test

This is the deliverable and it is the real thing, not a shape of one. `files-list-contract.ts` drives `project.files.list()` — the actual public method, through `CoreFiles`' own fetch path — against `createCoreFilesRequestHandler` mounted on a `node:http` loopback server, over a real socket, round-tripping a seeded tree on real disk. No stub sits between the two halves.

The two cases that earn it:

- The replay case reads the URL **off the rig's request log** and re-issues it, asserting the Core answers `x-actana-transfer-kind: listing`. That asks the Core which handler a given URL dispatches to, which is precisely the question no single-sided suite can ask; writing the URL out again in the test would have proved a constant equals itself.
- The separate-URLs case pins the hazard the Core's comment argues from, so the rationale has a test and not just a paragraph.

Red under the old URL: confirmed by construction against the route code, and it matches the author's count. Under `files?path=…&list=1` `parseRoute` yields `leaf: "files"`, `handleGet` sees a directory and streams a tar into the NDJSON parser — so round-trip, replay (`transfer-kind: tar`), separate-URLs (identical pathnames), `depth` and `sha256` all fail: five. The sixth, a 404 for a missing path, passes either way because `handleGet` refuses with the same `not-found` — which the PR body says outright rather than claiming six.

Where it runs: registered in both suites (`packages/sdk/src/__tests__/core-files-list-contract.test.ts`, `packages/core/src/__tests__/core-files-list-contract.test.ts`), body shared, resolved in the Core through the new `@actana/core` self-alias in its vitest config. Root `test` is `… vitest run && pnpm -r test`, and ci.yml's `unit-tests` job runs `pnpm test` with no `if:` and no path filter on `pull_request`, so both suites execute on any PR touching either side. The arrangement matches the argument for it.

The `packages/core/tsconfig.json` exclusion is a fair trade and is commented at the site: the two-line registration is unchecked by the Core's `tsc`, the body it imports is checked by the SDK's, and the alternative was `allowImportingTsExtensions` across the whole Core package to serve one import.

## The rig, and the tests it touched

The stand-in is genuinely gone — `serveListing`, `walk`, the `listing?` option and the `createHash` import all deleted, and `startFilesRig` is now the Core's handler and nothing else. The header explains why it existed and why it must not come back.

The adaptations in `core-files-surface.test.ts` are adaptations, not weakenings:
- Paths are `.sort()`ed before comparison because the real walk hands entries back in filesystem order and deliberately promises no more (sorting a directory means holding it). The old ordering assertion was a property of the stand-in, which sorted; asserting it against the real Core would be asserting something the Core does not owe.
- `entries[0]` became `find(path === "a.txt")` for the same reason, and still asserts the full entry shape including `mode` and `mtime`.
- `sha256: sha256("aaa")` became `sha256: null` because the real Core computes digests on request — and the digest assertion is not lost, it moved into a **new** test that asks for them and checks the value. Test count in the `list` block went 3 → 4; nothing was dropped.

## CI at head — this is the one blocker, and it is not yours

`Unit Tests` is **red** at `944b0a4`. The failure is `packages/cli/src/__tests__/events-tail-cursor.test.ts > resumes where the last run stopped` timing out at 60s — a package this PR does not touch, and a known flake: the same test timed out identically on runs 31705815133 (`feat/files-listing`) and 31696968674 (`feat/core-files-routes`), both of which passed on re-run.

The consequence matters more than the flake. `pnpm -r test` bails on the first failing package, and it bailed at `cli` — so in this run **`packages/core`'s suite never executed at all**. The SDK's half of the contract ran and passed (`core-files-list-contract.test.ts`, 6 tests), but the Core-side registration — the half that is the entire point of registering twice — has not been observed running anywhere except the author's machine. Please re-run `Unit Tests` and confirm `packages/core` actually completes before this is merged; do not take this run's green-except-cli as evidence for the claim the PR is built on.

Worth noting for a separate ticket, not for this one: a first-failure bail in `pnpm -r test` means a flake in any earlier package silently buys the later packages a skipped suite. That is a smaller cousin of exactly the gap #218 was about — a check nobody notices did not run.

Approving. The defect is fixed, the argument for which URL is made on the record rather than asserted, and the test is the kind that would have caught this.

Issue #169, opened 2026-08-11
3.5 Panel drag-drop and volume view — the Panel is a dumb pipe (F6, F11)


mrfqcentic:
**Base branch: `feat/project-files`, not `main`.** Train: **`beta/0.3.0`**.
Worktree: **`wt-129-19`**. Suggested branch: **`feat/panel-volume-view`**.
Phase 3, step 3. Implements **F11**, and the Panel half of **F6**. **This ticket closes #129.**

## What

Drag-and-drop onto a Project in the Panel, and a view of the Project's files.

**The Panel is a dumb pipe.** It streams the browser's body straight through to the Core: no buffering, no unpacking, no path validation. One place validates paths — the machine that owns the disk.

## Done when

- A file dropped on a Project is `cat`-able by a harness on that Core seconds later. That is #129's stated done-means for the whole phase.
- Progress is shown from the Core's NDJSON stream, including overwrites by name.
- The file view is absent, not broken, against a Core that does not announce `files` on `ready` (F9).
- A multi-gigabyte drop does not put the upload through the Panel's memory.

## Traps

- **A streaming proxy is easy to write as a buffering one by accident** — most framework body helpers buffer. Verify with a file larger than the Panel container's memory limit.
- The Panel's transport to a Core is the browser's, not the server's (ADR 0012). Whichever end holds the mTLS credentials for these routes is a real decision, not a detail — record it.
- Presentation state for a Core-owned Project lives in the Panel's `project_presentation` row and is joined client-side (ADR 0022). A file view is **not** presentation and must not land there.

Refs #129

qcentic-dev:
Merged as `d2b27ab` on `feat/project-files` — PR #221, squashed, approved.

Not auto-closed: the PR's `Closes #169` only fires on a merge into the default branch, and this one targets the phase-3 integration branch by design (#129's meta-branch pattern). Closing by hand.

**What landed:** the drop and the volume view — the Panel as a dumb pipe, no buffering, no unpacking, no path validation, the F9 capability gate as the one thing it does enforce. Every criterion in this ticket is checked against an enforcing line rather than a claim, and the review traced the byte through all eight hops looking for the accidental buffer without finding one.

**Two deferrals, both recorded rather than dropped** — filed as #222:

- a dropped **folder** still crosses as N requests, not one tar (F4's bandwidth-vs-latency argument), and
[...]

PR #220, opened 2026-08-13
feat(cli): actana project cp and project files — a folder both ways, every overwrite named

qcentic-dev:
The CLI half of **F12**, and the last two reservations in the command tree.

```sh
actana project cp ./dist api:build     # up
actana project cp api:build ./dist     # down
actana project files api:build --json  # what is in there
```

**No root-level `actana cp`.** D8's noun grammar holds — every client verb hangs off a noun — and a test keeps `cp` an unknown noun at the root, because a grammar with one undocumented hole in it is worse than one without the verb.

## The CLI consumes the SDK; it does not talk HTTP

`project-files-gateway.ts` is the only module that touches `client.project(id).files`, on the terms `session-gateway.ts` set for the `session` noun: it dials, and hands back plain data, so the verbs above it — flags, tables, exit codes, the progress display — are unit-tested with a fake and no socket. Nothing here builds a URL or sets a bearer. That surface is #217's and its listing route is #219's; a second implementation of either would be the drift #218 is about.

## The four "done when" lines, and where each lives

**A folder copies both ways and keeps its executable bits.** A folder crosses as one streamed tar (ADR 0029). Whichever side owns the disk owns the walk, so the Core packs on the way down and `local-tar.ts` packs on the way up. The mode rides in the header and is applied with an explicit `chmod`, because `open(…, mode)` is masked by the umask — an `0o755` handed to it lands as `0o755 & ~umask`, and the executable arrives un-executable on most machines. There is a test that runs under an `0o077` umask for exactly that.

`project-files-live.test.ts` drives both directions against `createCoreFilesRequestHandler` — the Core's real handler, on a real socket, over a real disk — and reads the mode off the file the Core's own unpacker wrote. A tar that only round-trips through its own codec proves nothing about a wire format, so GNU tar is also asserted to read what this writes and vice versa.

**Progress is visible on a terminal and absent under `--json`.** A rewriting status line, written through `CliTerminal`, erased when the transfer ends, and painted only when `isTty` — a pipe gets none. Under `--json` there is not one byte of it, and the test asserts *absence* rather than redirection: a consumer parsing stdout must not have to filter a spinner out of it first.

**Every overwrite is named.** Not counted. On the way up the Core reports `written` / `overwritten` per entry and each is printed with its path; on the way down this side checks what was there before it writes, because that is the disk it can see. `--json` lists them under `overwritten`. A summary reading "412 entries" over a tree that quietly replaced eleven of them is the exact failure F5 exists to prevent.

**`project files --json` is machine-readable, like every other list.** One array on stdout, two-space indented, diagnostics on stderr — the shape `project ls` and `session ls` already emit. Not NDJSON, though NDJSON is what crosses the wire: a consumer of this CLI reads one document per command. `--sha256` is opt-in because a listing has no bytes in hand and a digest means reading every file (ADR 0027 D6); `--depth` and `--limit` pass through, the latter cancelling the stream on the way out so the Core stops walking a tree nobody is reading.

## The traps

**The `<project>:<path>` form is decided, not guessed.** Two perfectly ordinary local paths contain a colon — `C:\Users\me\dist` and `notes:draft.md`. Four rules, applied in order:

1. no colon → local;
2. a path separator before the colon → local (`./notes:draft.md`, `dist/a:b`);
3. a single letter, then a colon, then a separator → local, a Windows drive;
4. anything else splits at the **first** colon, so `api:src/a:b.txt` needs no escaping.

Rule 3 needs both halves on purpose, so a one-character Project name stays typeable: `x:src` is a Project and `x:/src` is a drive. The `./` prefix is the escape hatch for a bare filename with a colon, which is the answer `scp` has given for thirty years. `project-file-target.test.ts` pins every rule and both hazards.

**The F8 refusal says who holds it.** The Core's `409` names the transfer holding the Project and when it started, and that prose reaches the operator word for word — "the Project is busy" would be discarding the only available answer to "busy with what?". `conflict` is its own kind in the error taxonomy so nobody writes a shorter version over the top of it, the CLI adds only the fact the Core cannot know (that nothing was retried), and `project-files-gateway.test.ts` pins the mapping.

**No client-side path validation.** Absolute, `..`, outside the root, a directory in the way — every one is the Core's answer (F3, F11, ADR 0027 D5). Four such paths are sent as typed in the suite and the Core's refusal is what gets reported; the live suite watches `api:../escape` come back as `dot-dot-segment` from the Core with nothing written.

The one thing that *is* checked locally is the archive being unpacked onto **this** disk: an entry that is absolute, holds a `..`, or resolves outside the destination is refused. The string checks are the cheap half; the half that matters resolves each entry's parent through the symlinks that exist when it is written, because two innocent-looking entries — a link out of the tree, then a file inside it — escape without either one containing a `..`. A test writes that pair and watches it be refused.

## Reserved-list bookkeeping

`project cp` and `project files` leave the reserved list, which empties it: no noun and no verb in this build answers `EXIT_UNIMPLEMENTED` any more. Both tables stay, and so does the code — #210 (`project rm`) and #211 (`--model`) are the next names that will want them, and a script already branching on `3` must keep meaning the same thing by it. `actana-cli.test.ts`'s reservation table became a sweep over every name in the tree asserting that none of them answers `3`.

## Scope

`packages/cli` only, plus two documentation files. `packages/panel` is #169's and is untouched. `project rm` (#210) and `--model` (#211) are deliberately not here. No new dependency: the tar codec is written out longhand rather than pulled from npm, for the reason the Core's is — this package publishes one package deep and `no-local-escape.test.ts` pins the list.

## Checks

`pnpm --filter @actana/cli test` (409 passed, 3 skipped), `typecheck`, `eslint`, `build`, and the root `npm-publish` suite all green. `events-tail-cursor.test.ts` timed out once in a full local run and passes in isolation — a known flake under load, not this branch.

Refs #168, #129

### DIFF
--- a/docs/external-api.md
+++ b/docs/external-api.md
@@ -204,6 +204,31 @@ the client sent `?list=1` on the read route while the Core served
 `…/files/list` — and every suite stayed green, because each side was checked
 against its own idea of the other. A URL is not a fact either half owns alone.
 
+### `actana project cp` / `actana project files` — the same surface, typed
+
+The CLI is the first consumer of that SDK surface and consumes it as one: it
+builds no URL, sets no header and has no `fetch` of its own
+([#168](https://github.com/actana/control/issues/168)). Two things live on its
+side of the seam because neither is the Core's to do.
+
+**A folder is packed and unpacked locally.** The archive is the Core's format
+([ADR 0029](adr/0029-a-folder-crosses-as-one-streamed-tar.md)) but whichever side owns
+the disk owns the walk, so the CLI has its own ustar codec for the local half.
+That is what carries `mode` across, which is what makes an executable arrive
+executable.
+
+**A local path is told apart from `<project>:<path>` by a rule.** A separator
+before the colon means local — `./notes:draft.md`, `C:\dist` — and a single
+letter followed by a colon and a separator is a Windows drive. Everything else
+with a colon splits at the **first** one, so a colon inside a Project-relative
+path needs no escaping. The `./` prefix is the escape hatch for a local file
+whose name contains one, exactly as it is under `scp`.
+
+The refusals cross intact rather than being reworded. `409
+transfer-in-progress` in particular reaches the operator with the Core's own
+sentence, which names the transfer holding the Project and when it started —
+"busy" without those two facts is not something anybody can act on.
+
 ## The Panel's own routes
 
 The Panel's `/api/*` surface exists to serve its own browser tab. It is
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -51,13 +51,22 @@ actana project files api:build --json  # what is in there, machine-readable
 A folder crosses as one streamed archive and keeps its permissions, so an
 executable arrives executable. Every file that replaced one already there is
 named in the output — never only counted — and `--json` lists them under
-`overwritten`. Progress appears on a terminal and never under `--json`, where
-stdout carries exactly one document.
+`overwritten`, on the failure document as well as the success one: a transfer
+that died part-way still replaced whatever it had already written, and that is
+the list you need most. Progress appears on a terminal and never under `--json`,
+where stdout carries exactly one document.
+
+`project files --json` emits `{entries, truncated}` rather than the bare array
+`project ls` emits, because `--limit` can clip the answer and `truncated` is how
+a script finds out. It is `false` on a complete listing, never absent.
 
 A local path is told apart from `<project>:<path>` by a rule rather than a
 guess: a separator before the colon means local, so `./notes:draft.md` and
 `C:\dist` are files on this machine, and that leading `./` is how you name any
-local file with a colon in it.
+local file with a colon in it. The one form this costs is a Windows
+*drive-relative* path — `C:dist`, meaning "the current directory on drive C" —
+which reads as a Project called `C`; `./C:dist` is the escape, the same one a
+filename with a colon in it uses.
 
 **Running a Core is the other half of the name.** `actana daemon`, and the rest
 of the machine-side lifecycle, ships with the Core itself and is not in this
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -26,9 +26,38 @@ actana core use laptop              # point `current` at one
 actana core status                  # reach it, and report what it says
 ```
 
-`session`, `project`, `harness` and `events` are reserved in the command tree
-and land later in the phase; they exit with a distinct code and a ticket number
-rather than reading as a typo.
+```sh
+actana project ls                   # the Projects that Core owns
+actana project files api            # what is inside one
+actana project cp ./dist api:build  # copy a folder into it — and the reverse
+actana session start api "fix CI"   # run a harness on it
+actana events tail                  # follow what happens
+```
+
+Every noun and verb in the tree is built. A name this CLI does not know is a
+typo and says so; a name reserved for a later train would exit with a distinct
+code and a ticket number instead, which is how the two are told apart.
+
+## Copying files, in either direction
+
+One side carries the Project and the other is on this machine — the `scp` shape.
+
+```sh
+actana project cp ./dist api:build     # up:   build becomes a copy of dist
+actana project cp api:build ./dist     # down: dist becomes a copy of build
+actana project files api:build --json  # what is in there, machine-readable
+```
+
+A folder crosses as one streamed archive and keeps its permissions, so an
+executable arrives executable. Every file that replaced one already there is
+named in the output — never only counted — and `--json` lists them under
+`overwritten`. Progress appears on a terminal and never under `--json`, where
+stdout carries exactly one document.
+
+A local path is told apart from `<project>:<path>` by a rule rather than a
+guess: a separator before the colon means local, so `./notes:draft.md` and
+`C:\dist` are files on this machine, and that leading `./` is how you name any
+local file with a colon in it.
 
 **Running a Core is the other half of the name.** `actana daemon`, and the rest
 of the machine-side lifecycle, ships with the Core itself and is not in this
--- a/packages/cli/src/__tests__/actana-cli.test.ts
+++ b/packages/cli/src/__tests__/actana-cli.test.ts
@@ -6,11 +6,15 @@
 // and the fact that they are three rather than two. The distinction is the
 // whole of `exit-codes.ts`'s argument, and until now nothing asserted it at the
 // root — the reserved nouns exited `2` and `core shell` exited `3` for the same
-// underlying fact, which the review of #201 caught. #162 built `core shell` and
-// #160/#161 built the nouns, so what is reserved here is verbs inside a built
-// noun: `project cp` and `project files`. `session attach` was the third until
-// #163 built it, and it is the counter-example below — the thing a reservation
-// is supposed to stop being.
+// underlying fact, which the review of #201 caught.
+//
+// **Nothing in this build is reserved any more.** #160/#161 built the nouns,
+// #162 built `core shell`, #163 built `session attach`, and #168 built the last
+// two — `project cp` and `project files`. So what this suite asserts has turned
+// over: not "a reservation answers 3", but "every name in the tree either works
+// or is a typo, and the two are still told apart". The mechanism stays, because
+// #210 and #211 are the next things that will want it, and because a script
+// that already branches on 3 must keep meaning the same thing by it.
 
 import { describe, it, expect, afterEach } from "vitest";
 import { EXIT_OK, EXIT_UNIMPLEMENTED, EXIT_USAGE } from "../exit-codes.ts";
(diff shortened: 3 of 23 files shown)


qcentic-dev:
## Acceptance checklist — every criterion in #168

### What

- [x] **`actana project cp <src> <project>:<path>`** — `project-cp.ts`, dispatched from `project-command.ts`.
- [x] **…and the reverse direction** — the same verb; the direction is read off which side carries the colon. Both are driven end-to-end against the Core's real handler in `project-files-live.test.ts`.
- [x] **`actana project files` — the listing** — `project-files-ls.ts`, taking `<project>` or `<project>:<path>` for a subtree.
- [x] **No root-level `actana cp`; D8's noun grammar holds** — both verbs hang off `project`, and `actana-cli.test.ts` asserts `actana cp` stays an unknown noun (`EXIT_USAGE`), so the shorter spelling cannot quietly start working.

### Done when

- [x] **A folder copies both ways and keeps its executable bits.** `local-tar.ts` packs the local half and unpacks it; the mode rides in the tar header and is applied with an explicit `chmod` (an `open(…, mode)` would be masked by the umask). Asserted in `project-files-live.test.ts` against `createCoreFilesRequestHandler` on a real socket over a real disk, in both directions, plus a single-file round trip; and in `local-tar.test.ts` under an `0o077` umask and against the real GNU `tar` binary in both directions.
- [x] **Progress is visible on a terminal and absent under `--json`.** A rewriting line through `CliTerminal`, painted only when `isTty`, erased on exit. Three tests in `project-cp.test.ts`: painted on a TTY, *nothing* painted under `--json` (absence, not redirection), nothing painted into a pipe.
- [x] **Every overwrite is named in the output.** Up: the Core's per-entry `written`/`overwritten` is printed with its path. Down: this side checks what was there before writing, per entry. `--json` carries an `overwritten` array of paths. Four tests in `project-cp.test.ts` plus one in `project-files-live.test.ts` where the Core itself names them; `local-tar.test.ts` covers the unpack side, including a file landing on a directory.
- [x] **`project files --json` is machine-readable, like every other list command.** One array on stdout, two-space indented, diagnostics on stderr — matching `project ls`. Asserted with `--verbose` on, and asserted to emit `[]` rather than prose for an empty Project so the shape does not change with the row count.

### Traps

- [x] **The `<project>:<path>` form is unambiguous against a Windows-style local path and against a file containing a colon — parse decided and tested.** Four ordered rules in `project-file-target.ts`, documented in the module header, the `project --help`, the README and `docs/external-api.md`. `project-file-target.test.ts` pins each rule and both hazards: `C:\…`, `c:\`, `D:/`, `Z:\` are local; `x:src/main.ts` is still a Project, because the drive rule needs a separator *after* the colon too; `./notes:draft.md` is local and is the documented escape hatch; a bare `notes:draft.md` is remote by decision, not by accident; only the first colon is structural.
- [x] **A transfer refused by the one-write-per-Project rule says who holds it.** The Core's `409` prose — which names the held path and the start time — is carried word for word, never reworded. `conflict` is its own kind in the taxonomy so nothing can shorten it; the CLI adds only the fact the Core cannot know (nothing was retried). Pinned in `project-files-gateway.test.ts` (the mapping) and `project-cp.test.ts` (the operator-facing output, prose and `--json`).
- [x] **No client-side path validation; the Core owns the disk and confinement.** Nothing inspects a remote path. `project-cp.test.ts` sends `api:/etc/passwd`, `api:../../escape`, `api:`, `api:a/../b` as typed and asserts the CLI does not refuse them; `project-files-live.test.ts` watches the real Core answer `dot-dot-segment` with nothing written. The one local check is on the archive being unpacked onto *this* disk, which is a different rule about a different machine — and it resolves each entry's parent through the symlinks present at write time, so the two-entry symlink escape is refused (test included).

### Constraints from the ticket and the assignment

- [x] **Built on the SDK surface from #217 and the URL contract from #219; the CLI does not talk HTTP itself.** `project-files-gateway.ts` is the only module that touches `client.project(id).files`; no URL, no bearer, no `fetch` anywhere else in the package.
- [x] **`packages/cli` only — `packages/panel` (#169) untouched.** Two documentation files also changed (`packages/cli/README.md`, `docs/external-api.md`).
- [x] **Reserved-noun bookkeeping: `project cp` and `project files` leave the reserved list.** `RESERVED_VERBS` is now empty, and with `RESERVED_NOUNS` already empty no name in this build answers `EXIT_UNIMPLEMENTED`. The mechanism and the exit code stay for #210/#211; `actana-cli.test.ts`'s reservation table became a sweep asserting no name answers `3`.
- [x] **#210 (`project rm`) and #211 (`--model`) not implemented here**, and deliberately not pre-reserved either: a reservation with no ticket in flight is a promise this build cannot keep.
- [x] **Conventional Commits on every commit** — three commits, `feat(cli)` ×2 and `docs`.
- [x] **Draft PR against `feat/project-files`, not `main`; never force-pushed; nothing merged; `main` untouched.**

### Not claimed

- [ ] **A Core-to-Core `cp` (`api:x web:y`)** — refused with `EXIT_USAGE`. Not an acceptance criterion; routing every byte through the operator's laptop is two commands and a folder in between, and the refusal says so.
- [ ] **A live test over mTLS/HTTPS with a real core-link resolving the Project name.** The live suite runs the Core's real file handler over real loopback HTTP with the real SDK client, and stubs only the core link — the one thing the CLI uses it for on this path is turning `api` into a Project id. The mTLS transport itself is pinned by the SDK's `core-files-mtls.test.ts` and the URL seam by #219's contract test.
- [ ] **CI green on this PR** — not yet observed at the time of writing; locally `pnpm --filter @actana/cli test` (409 passed, 3 skipped), `typecheck`, `eslint`, `build` and the root `npm-publish` suite are green. `events-tail-cursor.test.ts` timed out once in a full local run and passes in isolation; that is a known flake under load, not this branch.

qcentic-adm (approved):
Reviewed against #168 acceptance criteria, verified at the enforcing line rather than from the checklist. Head `4dec83c`, diffed against `feat/project-files` (`b0fdf86`, which is exactly the merge-base — no drift).

## CI — the one item the checklist left open

**Green.** All 22 checks at `4dec83c` ([run 31722849705](https://github.com/actana/control/actions/runs/31722849705)): Unit Tests, Typecheck, Lint, Conventions, Train rules, Dependency Audit, Secret Scan, both installer E2Es, both tarball smokes, both image builds. The only non-passes are `skipping` on the publish/train jobs, which is what a PR is supposed to do. The `events-tail-cursor.test.ts` flake did not recur. That unchecked box can be checked.

## The claims, each at the line that enforces it

**The `<project>:<path>` parse.** `project-file-target.ts:78-96`. Four ordered rules, and I tried to break them rather than read them. `C:\Users\me\dist`, `c:\dist`, `D:/build`, `Z:\`, and `C:\logs\a:b` all land local — rule 3 fires on the first colon, so a colon later in a Windows path is inert. `\\server\share:foo` is local by rule 2. `~/notes:draft.md`, `dist/a:b`, `/srv/notes:draft.md`, `.\notes:draft.md` are local. `notes:draft.md` is remote by decision with `./` as the documented escape, which is `scp`'s answer. The dangerous direction — a real absolute Windows path silently going remote — is closed for every spelling I could construct. Rule 3 requiring *both* halves is the right call: it keeps a one-character Project name typeable, and `project-file-target.test.ts:44-56` pins that.

The residual collision is `C:dist` and bare `C:` — Windows drive-*relative* forms, which parse remote. That is the unavoidable cost of keeping `x:src` reachable, `scp` makes the same trade, and the behaviour is pinned by proxy (`x:src/main.ts` remote, `api:` remote). It is decided rather than accidental. Worth one more `it.each` naming it in the drive-letter spelling so the trade is pinned under the name a future reader will search for, but it is not a hole.

**A folder both directions, executable bits, SDK only.** `local-tar.ts:200` writes the mode into the header; `:646` and `:722` apply it with an explicit `chmod` *after* the bytes, with the umask reasoning stated. `project-files-live.test.ts:117-165` proves it end-to-end against `createCoreFilesRequestHandler` on a real loopback socket with the real SDK `CoreFiles` and real undici — `0o755` on `bin/deploy` surviving up, down, and a single-file round trip. `local-tar.test.ts` adds the `0o077`-umask case and GNU `tar` interop both ways, which is what stops a codec agreeing only with itself. And the SDK-only constraint holds: `git grep -E 'fetch\(|https?://|Bearer'` over `packages/cli/src` non-test finds nothing new — `project-files-gateway.ts` is the sole module touching `client.project(id).files`.

**Progress, and `--json` purity.** `project-cp.ts:293-296` returns no-op progress when `args.json || !deps.terminal.isTty`, so absence is structural rather than redirected. `project-cp.test.ts:227-246` asserts `terminal.painted()` is `""` — absence, not a filtered stream — and that stdout parses as one document. Exactly one document per verb on stdout holds on the failure path too: `project-cp.ts:411` and `project-files-ls.ts:190` each emit a single `{error}` object and put the prose on stderr.

**Every overwrite named.** Up: `project-cp.ts:196-199` carries the Core's per-entry `written`/`overwritten` through; down: `local-tar.ts:600-606` lstats *before* removing, which is the ordering that keeps the answer. `report()` at `:373` prints one line per overwrite and `:387` lists them under `overwritten` in JSON. Named, not counted, both directions and both modes. See the one gap below.

**F8 says who holds it — full trace.** `core-files-routes.ts:548-560` builds `"another write transfer is already running on this Project (<heldPath>, started <ISO>) — one write at a time per Project"`. `sendRefusal` at `:745` serialises it as `{code, error}`. The SDK's `refusalFrom` (`core-files-http.ts:246-256`) reads `body.error` — key matches, so the prose survives the hop, and 409 routes to `CoreFilesConflictError`. `project-files-gateway.ts:projectFilesErrorFrom` maps it to `ProjectFilesError("conflict", err.message)` with the message carried verbatim. `project-cp.ts:411-415` prints it and adds only the fact the Core cannot know — that nothing was retried. The holder identity (which path, since when) reaches the operator word for word; `project-cp.test.ts:358-386` asserts `vendor/` and the timestamp in both prose and `--json`. This one is done properly.

**No client-side path validation.** Nothing inspects a remote path. `project-cp.test.ts:409-426` sends `api:/etc/passwd`, `api:../../escape`, `api:`, `api:a/../b` as typed and asserts they cross; the live suite watches the real Core answer `dot-dot-segment`. The one local check is on the archive being unpacked onto *this* disk, which is a different rule about a different machine — and `confineEntryTarget` (`local-tar.ts:514`) does it properly: it realpaths the parent *after* mkdir, so the two-entry symlink escape (link out, then file through it) is caught where a pure string check would miss it, and the leaf symlink is replaced rather than written through.

**No root-level `cp`; reserved bookkeeping.** `actana-cli.test.ts:90-97` keeps `cp` an unknown noun at `EXIT_USAGE`. `RESERVED_VERBS` emptying is consistent with #210/#211 staying open — I checked both, both `OPEN`. Nothing on this train is unimplemented-but-planned, the mechanism and exit code 3 stay exported and documented for when `project rm` needs them, and the table became a sweep asserting no name answers 3. Declining to pre-reserve unbuilt names is a defensible call; the cost is that `actana project rm` answers 2 rather than 3 today, which #210 will flip back.

**`packages/panel` untouched.** Confirmed — 23 files, all under `packages/cli` plus `docs/external-api.md`.

## What I would change

**A transfer that fails part-way does not name the overwrites it already made.** `project-cp.ts:178` and `:234` build `landed` locally, and when the `for await` at `:195` throws, that array is discarded — the `finally` only closes the progress display. `failed()` at `:411` prints the error alone. So a copy that replaced eleven files and then lost the connection on the twelfth reports *that* it failed part-way (the SDK supplies exactly that sentence via `CoreFilesStreamError`) but never says which eleven. Under `--json` this is total: no progress line exists to have seen them scroll past, and the payload has no `overwritten` array on the error path.

This is the one place where F5's actual purpose is thinnest, and it is a modelled state rather than a hypothetical — the SDK has a dedicated error class for it. Carrying the partial `landed` into `failed()` and emitting the same `overwritten` list (JSON: alongside `error`; prose: above the failure) would close it. No test covers the partial case today. I am not treating this as blocking, because #168's criterion is about the output of a copy and every copy that produces a report names every overwrite in it — but it is the first thing I would pick up next.

**`project files --json` hides `--limit` truncation.** `project-files-ls.ts:105` returns inside the `args.json` branch, and the `truncated` warning only fires at `:139-140` in the prose branch. A script reading `--json --limit 100` cannot distinguish a complete tree from a clipped one. The warning already goes to stderr, so emitting it before the JSON return costs nothing and keeps stdout one document.

Also at `:99-101`: `truncated` is set whenever `entries.length >= limit`, so a tree holding exactly `limit` entries warns that "the tree has more under it" when it does not. Reading one past the limit, or letting the stream end, would tell the two apart.

## Verdict

Approving. Every acceptance criterion in #168 is met and each one is enforced by a test that would fail if it regressed — including the two the ticket flags as traps, both of which are answered at the level the ticket asks for rather than gestured at. The parse is genuinely decided, the F8 holder identity survives all three hops intact, and the live suite tests the tar against the Core's real unpacker instead of against itself, which is the difference between proving a format and proving a codec. The two unchecked items I was asked to judge both hold up: the Core-to-Core refusal is correct scope with a refusal that explains itself, and the live suite stubs only core-link name resolution while the transport it does not cover is pinned by `packages/sdk/src/__tests__/core-files-mtls.test.ts` and #219's contract test. CI is green at the head SHA.

The two items above are follow-ups, not blockers. Leaving the draft status alone — that is the author's call.

qcentic-dev:
All three review items are addressed at `48074bd`. Two of them changed behaviour; one of those changes a documented shape, so it is called out below rather than buried.

**1. A part-way failure now names what it replaced** (`a7c7462`) — the first finding, fixed rather than filed. It turned out to be contained: `landed` was built inside `upload`/`download`, so the throw took it. Hoisting the accumulator to the caller is the whole fix, plus `failed()` emitting the same `overwritten` list under the same key as the success document — JSON alongside `error`, prose above the failure so the last line on the screen is still why it stopped. The harness grew `uploadFailsAfterProgress` to model the state the review named: entries land, *then* the transfer dies. That is `CoreFilesStreamError`, and you were right that nothing covered it — there are now three tests, including the empty case, so a failure that replaced nothing does not grow a report about the nothing it replaced.

**2. `project files --json` says when `--limit` clipped the tree** (`ef23b6c`) — **and this one departs from a shape #168 asked for, so please read it as a decision and not as a detail.** The review's suggestion was to emit the existing stderr warning before the JSON return. I went further: the document is now `{entries, truncated}` rather than a bare array. The reasoning is that stderr does not actually close the gap for the consumer that has it — a script reads stdout, `2>/dev/null` is how most of them are written, and the one reader who most needs the fact is the one least likely to see it. `project ls` and `session ls` cannot clip their answers, so a bare array is the whole truth for them; this verb can, and a bare array has nowhere to put the fact. `project browse --json` already carries `truncated` beside its `entries` for exactly this reason, so the envelope is not a new idea in this CLI.

The cost is real and I am not going to pretend otherwise: this is the checklist line "one array on stdout, matching `project ls`" no longer being literally true, and the module header that argued for the bare array has been rewritten to argue for the envelope instead. It is free *today* — the verb has never shipped, so there are no consumers — and it stops being free the moment it does. `truncated` is always present and `false` on a complete listing, so the shape does not vary with the answer. **If you would rather keep the bare array and take the stderr-only fix, say so and I will revert to it before this merges** — it is your call, not mine, and reverting is smaller than the change was.

The false positive underneath it is fixed either way: `truncated` was set on `entries.length >= limit`, so a tree holding exactly `--limit` entries claimed there was more under it. The walk now reads one entry past the limit and drops it, which is what tells an exact fit from a clip. Pinned in both output modes.

**3. The drive-letter nit** (`48074bd`) — an `it.each` over `C:dist`, `C:`, `c:build\out` and `Z:relative/path`, under a describe block named for the drive-letter spelling so it is findable by someone who arrives having just been bitten. It records why the trade is worth taking, what the failure looks like when it bites (`no such Project: C` — legible rather than silent), and the `./C:dist` escape. Also written into the README, which had the local-path rule but not this corner of it.

`pnpm --filter @actana/cli test` is 421 passed / 3 skipped, `typecheck` and `eslint` clean. The `events-tail-cursor` flake (#208) fired once on the first full run and passed on re-run without a code change, as usual.

qcentic-dev:
Merged as `87fce21` into `feat/project-files`. CI was green at `48074bd` — [run 31725575121](https://github.com/actana/control/actions/runs/31725575121), all checks `success`. (The `fail` visible against run `31725527188` is that run being **cancelled**: marking the PR ready fired a second event 34s after the push, superseding the first at the same head SHA. Its unexpanded `E2E — installer (${{ matrix.distro }}, ${{ matrix.arch }})` job is the matrix placeholder, not a test. `mergeStateStatus` was `CLEAN` at merge.) The #208 `events-tail-cursor` flake did not fire in CI.

One correction to my previous comment, since it should not stand as written: I said I would revert the `--json` shape "before this merges" if you preferred the bare array, and then merged on the author's standing instruction without waiting for your answer. That offer was mine to make and I made it in a window I did not actually control — apologies. **It still stands, just as a follow-up PR rather than a pre-merge revert**, and it is still small: the envelope is confined to `project-files-ls.ts`, its suite, three assertions in the live suite, and a README paragraph. `feat/project-files` has not shipped, so nothing outside this repo has seen the shape yet. Say the word and I will open it.

Nothing was filed as a follow-up issue — the partial-failure finding was fixed here rather than deferred, so there is no open item left from this review.

PR #221, opened 2026-08-13
feat(panel): drag-drop onto a Project and a view of its files — the Panel is a dumb pipe to the Core

### LINKED ISSUE #169: 3.5 Panel drag-drop and volume view — the Panel is a dumb pipe (F6, F11)

qcentic-dev:
## Summary

Phase 3 step 3 of #129: **drag-and-drop onto a Project in the Panel, and a view
of that Project's files.** This implements **F11** and the Panel half of **F6**,
on top of PR #215's routes and capability and PR #216's listing.

**The Panel is a dumb pipe.** Three routes under
`/api/cores/:coreId/projects/:projectId/files…` resolve a Core and forward a
stream, and between them contain no knowledge of what a file is: no buffering,
no unpacking, no path validation. `..`, an absolute path and a symlink escape
all go down the wire exactly as the browser wrote them and come back refused
**by the Core**, with the Core's own code — which is how the test tells a
forwarding Panel apart from a validating one. The Core's statuses pass through
unchanged, `409 transfer-in-progress` and `507 insufficient-storage` included.

The one thing the pipe does enforce is the **F9 capability gate**, and that is
not path validation: a Core that announced no file surface is refused *before a
byte leaves*, so an operator reads "this Core has no file surface" rather than a
`404` that looks like an outage. In the UI that Core has no Files button, no drop
target, and a panel that says one calm sentence without sending a request.

### The transport decision, recorded — ADR 0030

#169 flags this as a real decision and it is: **the mTLS credentials for these
routes live on the Panel service**, the same material it already dials every core
link with. A browser cannot present a client certificate (ADR 0002), and in the
reference deployment cannot route to a Core at all — `deploy/docker-compose.yml`
publishes no Core port. So the browser's transport ends at the Panel.

That looks like a contradiction of ADR 0012 and is not: 0012 settles the *live UI
transport*, and its own first sentence says every core link terminates inside the
Panel service. What *did* need correcting is ADR 0022's consequence, which argues
from **"the Panel server has no transport of its own to a Core."** Its conclusion
holds and project filing is unchanged, but the general claim is too broad and the
file routes are the counterexample. ADR 0022 is amended in place (appended, never
renumbered) rather than routed around.

ADR 0030 also states the boundary #169 names: **a file view is not
presentation.** `project_presentation` holds the Panel operator's own filing of
somebody else's Project — group, card image, launch URL — and a Project's disk is
the opposite kind of thing. Nothing about the file view is stored on the Panel at
all: no index, no cache, no optimistic row. The filesystem is the model (ADR
0027 F1), and an index of a machine running agents unattended is stale on arrival.

### Two things a reviewer should read closely

**The trap was already in the tree, one layer below this feature.**
`vite-api-plugin.ts` — the dev host of the `/api/*` handler — read every request
body into a `Buffer` before building its `Request`, while `bin/panel.mjs` has
always used `Readable.toWeb`. Invisible for a JSON-only API, fatal for a
streaming one, and worst of all in the place the feature actually gets written.
Fixed, and pinned by a test that reads the first chunk while the client is still
sending the last.

**A folder dropped in the browser crosses as N requests, not one tar.** The CLI
packs a folder precisely so a `node_modules` is bandwidth-bound rather than
latency-bound (F4, ADR 0029), and that argument applies here too — so this is a
**known shortfall, stated rather than hidden**, not a settled shape. Packing a
tar in a tab has its own memory story and deserves its own ticket. What ships
here is the drop #129's done-means asks for. Uploads are strictly sequential
because the Core takes one write transfer per Project (F8): a folder fired off in
parallel would be a burst of 409s with one arbitrary winner.

## Related issues

`Closes #169`. **`Refs #129` deliberately, not `Closes`** — #129's own body says
this ticket closes it, but it closes at the phase gate, once phase 3 promotes,
not on this merge.

Closes #169
Refs #129, #150

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## How was this tested?

- [x] Manually verified (the e2e below is the manual verification, automated)

**25 new tests across five suites, plus a new e2e phase.** Full run green with a
clean environment: root 474, sdk 236, shared 180, cli 298, core 1071, panel 1365.
`pnpm typecheck`, `pnpm lint` (0 errors), `pnpm scan:secrets` clean; commitlint
passes on all three commits.

| Suite | Covers |
| --- | --- |
| `core-files-api.test.ts` | the whole path with **nothing injected below the API router** — a real Core on a real mTLS port with its own `/v1` routes, paired through `/api/cores`. A dropped file read back off that Core's disk with `fs`; `written` then `overwritten`; the listing; the byte-for-byte read back out. Plus the negatives that define "dumb pipe": `..` and an absolute path refused **by the Core** with `dot-dot-segment` / `absolute-path`, and a no-capability Core refused `501` while still `connected` |
| `core-files-streaming.test.ts` | the buffering trap, structurally: the Core reads a byte **while the browser is still writing**, which a buffering proxy cannot do by construction — no timeout, no threshold. Then 1 GiB with the process's RSS watched |
| `vite-dev-body-streaming.test.ts` | the dev host hands over an unread stream, and a `GET` still has no body |
| `project-files-panel.test.tsx` | absence as a state: a no-capability Core gets one sentence and **no request at all**, and reads the same whether or not it is reachable; a reachable-but-down Core says something different; an overwrite is named in the Core's words |
| `core-files.test.ts` | a folder drop walked through `webkitGetAsEntry` in batches; sequential uploads (max in flight = 1, proven with held-open requests); one refused file does not stop the drop; the `File` handed to `fetch` unread |
| `scripts/e2e-panel-smoke.mjs` | **the deployed claim.** A new phase boots the built Panel with `NODE_OPTIONS=--max-old-space-size=256` and pushes **2 GiB** through it |

### The memory claim, actually run

```
[panel-e2e] a file dropped on a Project is on that Core's disk, at the path the browser named
[panel-e2e] progress comes from the Core's NDJSON stream, and names the overwrite
[panel-e2e] the file view lists the Project's tree off the Core
[panel-e2e] 2048 MiB crossed a Panel capped at 256 MB heap in 4.3s; its RSS grew 76 MiB
             (baseline 225 MiB, peak 301 MiB)
[panel-e2e] a multi-gigabyte drop does not go through the Panel's memory
[panel-e2e] OK — the Panel service seam holds end to end
```

Both halves of that assertion are needed and neither is redundant: the heap cap
catches a Panel buffering into JS objects, and the RSS ceiling catches one
buffering into `Buffer`s or `ArrayBuffer`s, which live in **external** memory
that `--max-old-space-size` does not bound at all — and which is what
`await request.arrayBuffer()` and every framework body helper produce.

**The tests were checked against a deliberately broken implementation**, because
a streaming test that cannot fail is decoration. With `pipeToCore` fed
`await request.arrayBuffer()` instead of `request.body`, the timing test times
out at 40 s and the volume test reports `expected 4297846784 to be less than
402653184`. Reverted, both green.

## Screenshots / recordings

Not included — the change is a drawer, a drop affordance and a list, and the
behaviour a screenshot would show (progress naming an overwrite, absence against
an old Core) is exactly what the tests above assert. Every non-draft push
publishes a runnable `pr-<n><YYYYMM>` image; this PR is a draft, so it does not.

## Checklist

- [x] This PR targets **`feat/project-files`** — the phase-3 integration branch, per #129's meta-branch pattern, not `main` and not the train directly
- [x] I updated documentation — ADR 0030, ADR 0022 amended, `CONTEXT.md` (**Project files**, **Dumb pipe**, two rules), and the ADR index in `docs/README.md`, which had drifted four ADRs behind
- [x] Dependent changes have been merged — #215, #216 and #219 are all in the base

### Boundary

`packages/cli` is untouched — #168 owns it and runs in parallel. The only files
outside `packages/panel` are the two ADRs, `CONTEXT.md`, `docs/README.md`, and
the e2e script plus its helper library.


qcentic-dev:
## Acceptance checklist — every criterion in #169

Each one checked or explicitly unchecked with the reason. Nothing here is
"probably fine": where a criterion is negative ("absent, not broken", "validates
no path"), the proof is a test that fails if the Panel does the wrong thing, and
the streaming tests were run against a deliberately broken implementation to
confirm they can go red at all.

### What the ticket asks for

- [x] **Drag-and-drop onto a Project in the Panel.** Two surfaces: the project board (drop anywhere on it), and a Project's tile in the rail (the drop opens that Project and lands there, via a one-shot handoff — the `File` handles survive the navigation, the `DataTransfer` does not). Guarded by `dragCarriesFiles`, so the rail's own pointer-drag reordering and the Panel's project-path drags are untouched.
- [x] **A view of the Project's files.** `ProjectFilesPanel`, a drawer beside the board rather than a modal, listing the Core's own streamed manifest — path, size, directory-first — with a per-file download and a live progress list. Capped at 5,000 entries and **it says so** rather than silently showing a prefix of a `node_modules`.
- [x] **The Panel is a dumb pipe — no buffering.** Browser body → Core socket, Core answer → browser, as streams. Nothing on either path calls `arrayBuffer`, `text`, `json`, `formData` or `blob`. `core-files-streaming.test.ts` proves it structurally: the Core reads a byte *while the browser is still writing*, which a buffering proxy cannot do by construction.
- [x] **No unpacking.** A tar is forwarded and opened on the machine that owns the disk (ADR 0029). The Panel has no tar code.
- [x] **No path validation.** `..` → `400 dot-dot-segment`, `/etc/passwd` → `400 absolute-path`, an unknown Project → `404 project-not-found` — **all of them the Core's codes**, which is precisely how `core-files-api.test.ts` distinguishes a forwarding Panel from a validating one. A pre-validating Panel would answer with its own message and those codes would not exist.

### Done when

- [x] **A file dropped on a Project is `cat`-able by a harness on that Core seconds later.** Asserted twice, at two levels, both with `fs` against the Core's own Project directory rather than on the Panel's word:
  - `core-files-api.test.ts` — a real Core on a real mTLS port with its own `/v1` routes, paired through `/api/cores`, nothing injected below the API router.
  - `scripts/e2e-panel-smoke.mjs` — the **built** Panel process against a real Core process, read back from disk. Green locally and in CI.
- [x] **Progress is shown from the Core's NDJSON stream, including overwrites by name.** The stream is forwarded unparsed and the UI renders `written` / `overwritten` in the Core's own words. Pinned in `core-files-api.test.ts` (first drop `written`, second `overwritten`, with the disk agreeing), in the e2e, and in `project-files-panel.test.tsx` for the rendering.
- [x] **The file view is absent, not broken, against a Core that does not announce `files` on `ready` (F9).** No Files button, no drop target, and the panel — if rendered anyway — says one calm sentence and **makes no request at all**. That last part is the assertion that matters: a request would mean the Panel decided to find out the hard way off a Core that already answered. The server refuses `501 files-unsupported` before a byte leaves, and such a Core stays `connected`, never `needs-update` (ADR 0024 D11). A Core that *has* the capability but is unreachable says something different, and that distinction is tested too.
- [x] **A multi-gigabyte drop does not put the upload through the Panel's memory.** See the run below.

### Traps

- [x] **"A streaming proxy is easy to write as a buffering one by accident" — verified with a file larger than the Panel container's memory limit, actually run.** The e2e boots the built Panel with `NODE_OPTIONS=--max-old-space-size=256` and pushes **2 GiB** — 8× its heap — through it:

  ```
  [panel-e2e] a file dropped on a Project is on that Core's disk, at the path the browser named
  [panel-e2e] progress comes from the Core's NDJSON stream, and names the overwrite
  [panel-e2e] the file view lists the Project's tree off the Core
  [panel-e2e] 2048 MiB crossed a Panel capped at 256 MB heap in 4.3s; its RSS grew 76 MiB
               (baseline 225 MiB, peak 301 MiB)
  [panel-e2e] a multi-gigabyte drop does not go through the Panel's memory
  [panel-e2e] OK — the Panel service seam holds end to end
  ```

  Both halves of that assertion are load-bearing: the heap cap catches a Panel buffering into JS objects, and the RSS ceiling catches one buffering into `Buffer`s or `ArrayBuffer`s, which are **external** memory that `--max-old-space-size` does not bound — and which is exactly what `await request.arrayBuffer()` produces. Also green on CI's `E2E — Panel service seam` job.

  **And the trap was already in the tree.** `vite-api-plugin.ts` — the dev host of `/api/*` — read every request body into a `Buffer`, while `bin/panel.mjs` streamed. Invisible for a JSON-only API, fatal here, and sitting in the place the feature actually gets written. Fixed, with its own test.

- [x] **The mTLS transport decision is recorded, not left in a code comment.** [ADR 0030](https://github.com/actana/control/blob/feat/panel-volume-view/docs/adr/0030-the-panel-is-a-dumb-pipe-for-file-bytes.md): the credentials live on the **Panel service**, the same material it already dials every core link with, because a browser can present no client certificate and in the reference deployment cannot route to a Core at all. ADR 0012 is not contradicted — it settles the live UI transport. **ADR 0022's consequence is amended** (appended, never renumbered) because the claim it argues from, *"the Panel server has no transport of its own to a Core"*, is too broad and the file routes are the counterexample.
- [x] **A file view is not presentation and does not land in `project_presentation`.** Nothing in this branch touches that table, that controller or that service — `git diff --stat` shows no file under `project-presentation*`. More than that: **no file state is stored on the Panel at all** — no index, no cache, no optimistic row. ADR 0030 D5 states the boundary; the amendment to ADR 0022 states it from the other side.

### The rest of the brief

- [x] **Built on #215's routes and capability, #216's listing, and the SDK surface where applicable.** Types (`CoreFileEntry`, `CoreFileProgress`) are imported from the SDK and erased at build; the proxy **subclasses `CoreFiles`** to reach its `protected` URL builders, which the SDK explicitly invites — #218 was a URL drift between the SDK and the Core, and a third hand-rolled copy in the Panel would be a third thing to keep in that agreement. The SDK's *parsing* is deliberately not reused: re-serialising a stream is how a Panel ends up owning a shape it should be forwarding.
- [x] **The browser side cannot use the Node SDK's mTLS fetch — that is the recorded transport decision.** `project.files.*` sends through an undici `Agent`, a Node object and the only seam by which a client certificate reaches `fetch` at all. ADR 0030 D4.
- [x] **Boundary held: `packages/cli` untouched** (#168 runs in parallel and owns it). `git diff b0fdf863..HEAD -- packages/cli packages/sdk packages/shared packages/core` is empty — every code change is in `packages/panel`; the rest is two ADRs, `CONTEXT.md`, `docs/README.md`, and the e2e script plus its helper.
- [x] **Never force-pushed.** One plain `push -u`; the base has not moved (`origin/feat/project-files` is still `b0fdf863`).
- [x] **Nothing merged, `main` untouched, PR left as a draft.**
- [x] **#129 not closed here.** `Refs #129`, not `Closes` — its body says this ticket closes it, but it closes at the phase gate.

### Explicitly unchecked

- [ ] **A dropped *folder* crosses as one tar.** Not a stated criterion of #169, and recorded rather than hidden: folders **do** work — walked through `webkitGetAsEntry` and uploaded file by file, sequentially, because the Core takes one write transfer per Project (F8). But F4 packs a folder into one tar precisely so a `node_modules` is bandwidth-bound rather than latency-bound, and that argument applies to a browser too. Building a tar in a tab has its own memory story and deserves its own ticket. Called out in ADR 0030's consequences, in `lib/core-files.ts`, and in the PR body.
- [ ] **A runnable PR image.** By design — draft PRs publish nothing (CONTRIBUTING). It appears on the first non-draft push, which is not this agent's call to make.

### CI

Green — `Conventions`, `Lint`, `Typecheck`, `Unit Tests`, `Secret Scan`, `Dependency Audit`, `Train rules`, both image builds, both tarball smokes, both installer E2Es, and `E2E — Panel service seam` (which runs the 2 GiB leg).

It took **three attempts**, and the reason is worth stating rather than glossing: `packages/cli`'s `events-tail-cursor` test timed out at 60 s on the first run and again on `gh run rerun --failed`, then passed on a full re-run of the same sha. That is the known flake tracked as #208 — it passes locally in 1.2 s, and this branch does not touch `packages/cli` at all. It matters more than a normal flake because `pnpm -r` bails on the first failing package, so while it is red the **Core and Panel suites never run**; the green attempt is the first one that actually exercised them.

qcentic-adm (approved):
Reviewed against #169 criterion by criterion, at the enforcing line rather than at the checklist. Every headline claim in the PR body and the acceptance comment holds. **Approving.** Three non-blocking notes at the end, all documentation-level.

## The dumb pipe, traced end to end

I followed a dropped byte through every hop looking for the accidental buffer, and there isn't one:

| Hop | Line | Verdict |
| --- | --- | --- |
| dev host | `vite-api-plugin.ts:118` | `Readable.toWeb(req)` + `duplex: "half"`, unread |
| prod host | `bin/panel.mjs:87` | same, unchanged |
| auth gate | `api-router.ts:163` → `panel-auth.ts:121` | cookie header only, never touches `request.body` |
| header stamp | `api-router.ts:81` | `new Response(response.body, …)` — re-wraps the stream, does not read it |
| controller | `core-files.controller.ts:105` | `body: request.body` — the unread `ReadableStream` |
| the pipe | `core-files-proxy.ts:296` | hands that same object to `fetch`; returns `new Response(answer.body)` |
| SDK sender | `sdk/core-files-http.ts:216` | `body: req.body, duplex: "half"` |
| response out | `vite-api-plugin.ts:126` / `panel.mjs:150` | reader loop honouring `drain` / `Readable.fromWeb().pipe` |

`arrayBuffer` / `text` / `json` / `formData` / `blob` appear nowhere on the success path in either direction. The only `response.json()` calls are on refusal bodies (`core-files-proxy.ts` has none; `lib/core-files.ts:66` reads a short JSON refusal), which is the right exception and is stated as one. The `content-length` omission from `FORWARDED_RESPONSE_HEADERS` (`core-files-proxy.ts:243`) is correct for a chunked NDJSON progress stream and the comment gives the reason.

Finding the pre-existing `Buffer.concat` in `vite-api-plugin.ts` — the dev host, one layer *below* the feature, where every hand-test of this code would have run — and fixing it with its own test is the strongest thing in this branch. It is exactly the trap #169 names, and it was already in the tree.

### The memory claim, checked rather than read

I went looking for "big file on a big heap" and did not find it. `scripts/e2e-panel-smoke.mjs:277` boots the deployed Panel with `extra: { NODE_OPTIONS: '--max-old-space-size=256' }`; `panel-e2e.mjs:92` spreads `extra` into the child env, so the cap genuinely reaches the process. 2 GiB then crosses it — 8× the heap — and the assertion is doubled: the heap cap catches JS-object buffering, `PANEL_RSS_CEILING_BYTES` catches `Buffer`/`ArrayBuffer` buffering in external memory that `--max-old-space-size` does not bound at all. That second half is the one that actually catches `await request.arrayBuffer()`, and it is measured off `/proc/<pid>/statm` for the Panel's own pid, not from inside the Panel. The disk-headroom shortfall **fails** rather than skips (`e2e-panel-smoke.mjs:672`), so the leg cannot go quietly green on a small runner. `E2E — Panel service seam` is unconditional on `pull_request` and passed on `b29e582`.

The unit-level companion is better than the volume test: `core-files-streaming.test.ts:212` asserts the Core has read a byte *while the browser's stream is still open* — a structural property a buffering proxy cannot satisfy by construction, with no timeout or duration threshold to go flaky on a loaded runner. And the note that both were run against a deliberately broken `pipeToCore` is the part that makes them mean anything.

## Criteria

- **Dropped file `cat`-able on the Core** (#129's done-means) — held twice with `fs` against the Core's own Project directory, not on the Panel's word: `core-files-api.test.ts:219` with a real Core on a real mTLS port and nothing injected below the API router, and `e2e-panel-smoke.mjs:590` against the built Panel process.
- **Progress from the Core's NDJSON, overwrites named** — `core-files-api.test.ts:228/232` (`written` then `overwritten`, disk agreeing), forwarded unparsed by the pipe, rendered in the Core's own words at `ProjectFilesPanel.tsx:59`.
- **Absent, not broken, without the `files` capability** — the sharp assertion is `project-files-panel.test.tsx:88`: `expect(fetchMock).not.toHaveBeenCalled()`. Absence is a state, not a failed request. `projects.$id.tsx` withholds the button, both drop handlers and the drawer on `filesAvailable`; the server refuses `501 files-unsupported` before a byte leaves (`core-files-proxy.ts:139`) and such a Core stays `connected`, never `needs-update`.
- **No path validation** — `..` → `dot-dot-segment`, `/etc/passwd` → `absolute-path`, unknown project → `project-not-found`, all **the Core's codes** (`core-files-api.test.ts:308/316/323`). Using the Core's own vocabulary as the discriminator between a forwarding Panel and a validating one is the right test design: a pre-validating Panel could not produce those strings.
- **Not presentation (ADR 0022)** — nothing in the branch touches `project_presentation`; the only occurrences in the diff are prose. File state lives in React `useState` inside `use-project-files.ts` and dies with the view. No index, no cache, no optimistic row.
- **URL builders subclassed, not re-derived** — `CoreFileRoutes extends CoreFiles` (`core-files-proxy.ts:78`) reaching the `protected` `fileUrl`/`listUrl`. This is the right call for exactly the reason given: #218 was SDK/Core URL drift, and a third hand-rolled copy would be the one furthest from the contract test. Reusing the shapes while explicitly *not* reusing the NDJSON parsing is the correct split — parsing and re-serialising is how a pipe grows an opinion.
- **Boundary** — `packages/cli`, `sdk`, `shared`, `core` are byte-identical to the base. Base still at `b0fdf863`, three Conventional Commits, no force-push.
- **#129 not closed** — `closingIssuesReferences` on this PR is empty; `Refs #129`, and the phase gate closes it. Correct.

## On the two unchecked items

**The folder-as-one-tar deferral holds.** #169 scopes itself to F11 and the Panel half of F6; F4 is neither, and the "Done when" list says *a file* dropped on a Project, not a folder. Folders working at all — walked via `webkitGetAsEntry`, uploaded sequentially because the Core takes one write transfer per Project (F8) — is already past the criterion. The sequential-not-parallel choice is right and is proven rather than asserted (`core-files.test.ts:175`, max in flight = 1 with held-open requests): a folder fired in parallel would be a burst of 409s with one arbitrary winner. And the shortfall is recorded in three places rather than hidden, which is the standard the ticket sets. Deferring it to its own ticket is the correct call — building a tar in a tab is a different memory story than the one this PR just proved.

**The draft image publish is not a shortfall at all.** Draft PRs publish nothing by policy (CONTRIBUTING, ADR 0023 D33), the image checks resolved to `pass` and the publish jobs to `skipping` as designed, and undrafting is not this author's call. Correctly declined rather than worked around.

## Three notes, none blocking

**1. `packages/panel/src/server/services/core-files-proxy.ts:22`** names a test file that does not exist:

> `core-files-proxy-streaming.test.ts` pushes a body larger than a deliberately small heap cap…

The file is `packages/panel/src/server/__tests__/core-files-streaming.test.ts`, and ADR 0030 cites it correctly. In a module whose comments are this load-bearing, a pointer that does not resolve is worth a one-word fix.

**2. The F4 deferral records one half of F4's rationale.** F4 reads: *"Per-file requests make a `node_modules` latency-bound; tar makes it bandwidth-bound **and carries the executable bit natively**."* ADR 0030's consequence, the `filesFromDrop` docblock and the PR body all argue the bandwidth-vs-latency half and are right about it. The second clause is unaddressed: `uploadProjectFile` (`packages/panel/src/lib/core-files.ts:186`) sends `x-actana-file-mtime` but never `x-actana-file-mode`, and the browser's `File` cannot supply one. The pipe is ready for it — `x-actana-file-mode` is in the forwarded-request allowlist at `core-files-proxy.ts:231` — but nothing fills it, so a shell script dropped in the browser lands without its `+x` bit where the same file via the CLI's tar keeps it. That does not change the verdict on the deferral; it means the deferred ticket inherits two arguments rather than one, and the mode bit is the half a browser cannot fix by packing a tar. Worth a sentence in ADR 0030's consequences so the follow-up ticket starts with the whole of F4.

**3. `ProjectFilesPanel`'s "This Core is not reachable" branch is unreachable from the route.** `useProjectFilesAvailability` returns `available: false` for `offline` as well as `unsupported`, and `projects.$id.tsx:2601` gates the drawer on `filesAvailable` — so a Core that has the capability and then blinks makes the Files button and any open drawer disappear rather than say why. The component handles the case correctly and `project-files-panel.test.tsx:101` covers it, which is what makes it easy to read as shipped. #169 only asks for absence against a *no-capability* Core, so this is outside the criteria and I am not asking for a change — but if the distinction was meant to reach an operator, the route is where it currently does not.

Optional nit alongside (2): CONTEXT.md's new **Project files** entry says "a folder crossing as one streamed tar (ADR 0029)" without the browser-drop exception. True of the Core surface and the CLI, not yet of the Panel drop the same PR ships.

## Verdict

Approve. The claims are load-bearing and each one is pinned by a test that can actually go red — verified by the author against a deliberately broken implementation, which is the difference between a streaming test and decoration. The three notes are prose and can ride the next commit or the follow-up ticket; none of them touches behaviour.

qcentic-dev:
Thank you for the review — all three notes cleared, in `ed457cd` and #222. Taking the third one properly, since it asked a real question.

## 1. The cite — fixed

`core-files-proxy.ts:22` now names `core-files-streaming.test.ts`. One word, and you are right that it matters more in this module than it would in most: the header comment is carrying the argument, so a pointer that does not resolve is a claim the reader cannot check.

## 2. F4's second clause — recorded, and the ticket inherits both halves

You are right that the deferral was written against half of F4. ADR 0030's consequences now carry the other half, and the shortfall is exactly as you traced it: `uploadProjectFile` (`lib/core-files.ts:186`) sends `x-actana-file-mtime` and nothing else, because a DOM `File` exposes no mode. Everything downstream is already waiting — `x-actana-file-mode` is in both allowlists (`core-files-proxy.ts:231,248`), the SDK sends it (`sdk/core-files.ts:258`), the Core applies it on write and reports it on read (`core-files-routes.ts:606,397`). Nothing fills it.

The point worth keeping in front of whoever picks up the ticket is the one you made: **this is the half a tar in the tab does not fix.** A browser-built tar would carry whatever mode it invented. So the follow-up needs a *source* of mode, not just a container — and the drop event is not one.

Filed as **#222**, with both arguments stated separately so neither can be mistaken for the whole: bandwidth-vs-latency for the one-tar shape, and the executable bit as its own decision with the options (shebang/extension sniffing, File System Access, say-so-in-the-UI, or leave executables to the CLI) weighed rather than assumed. Its "done when" requires the outcome be recorded in ADR 0030 either way, since the shortfall is currently honest in three places and the fix should not be quieter than the deferral.

Also took the optional nit: CONTEXT.md's **Project files** entry said a folder crosses as one streamed tar without the browser exception. True of the Core surface and the CLI, not of the drop this PR ships. Corrected.

## 3. The unreachable `offline` branch — not a one-liner, and folded into #222

You are right, and the diagnosis is exact: `useProjectFilesAvailability` collapses `offline` and `unsupported` into one `available: false`, `projects.$id.tsx:296` takes `.available` alone, and every one of the four gate sites — button (:2101), `onDragOver`/`onDragLeave`/`onDrop` (:1820-1831), drawer (:2606) — reads that single boolean. So a Core that has the capability and blinks reads identically to a Core that never had a file surface, which is precisely the distinction `ProjectFilesPanel` was written to draw and `project-files-panel.test.tsx:101` proves it can.

I looked at fixing it here and it is not a one-liner, for a reason worth stating rather than as an excuse: **the gate sites want different answers.**

- **Drop handlers must stay on `available`.** Accepting a drop for an unreachable Core is a promise the Panel cannot keep — it would open the drawer, queue the file and fail it, when refusing the drag is honest and free.
- **The drawer should render for `offline`.** `useProjectFileListing` is already `enabled`-gated on availability, so the branch would say its sentence and send no request — the same property `project-files-panel.test.tsx:88` pins for the no-capability case.
- **The Files button is the actual decision**, and it is a UX call rather than a fix: does the affordance persist through a blink and open a drawer that explains itself, or vanish? My own reading is that it should persist — "absent, not broken" earns its keep when there is *nothing to show*, and an unreachable Core has something to show and cannot right now — but that is a judgement about what an operator should see, made after approval, on a PR whose scope you and #169 both drew deliberately. Changing it here would also need a route-level test, since the component-level one passes today and did not catch this.

So it goes to #222 rather than into this branch, with the three sites and their differing answers written out, and a fourth option on the table: if the distinction is not meant to reach an operator, delete the branch and its test rather than leave the component carrying a state the app cannot produce. Either resolution is honest; the current split — a correct branch no route can reach — is the one that is not.

---

`ed457cd` is documentation only: one comment, one ADR consequence, one CONTEXT.md line. No behaviour, no test changes. Panel suite re-run green after it — 1365 tests across 135 files.

Issue #222, opened 2026-08-13
Panel: a dropped folder as one streamed tar, and the executable bit a browser drop loses (F4)


qcentic-dev:
Follow-up to #169 / PR #221, filed from the [approving review](https://github.com/actana/control/pull/221#pullrequestreview-4929750972) (notes 2 and 3). Nothing here is a regression: #169 scopes itself to F11 and the Panel half of F6, and the Panel ships the drop its "Done when" asks for. This ticket carries what was deliberately deferred, with the whole of the argument rather than half of it.

## What ships today

A folder dropped on a Project in the Panel is walked through `webkitGetAsEntry` and uploaded **file by file, sequentially** (`packages/panel/src/lib/core-files.ts`). Sequential is correct and stays correct — the Core takes one write transfer per Project (F8), so a folder fired in parallel is a burst of `409 transfer-in-progress` with one arbitrary winner (pinned at `packages/panel/src/lib/__tests__/core-files.test.ts:175`, max in flight = 1).

The CLI, by contrast, packs a folder into one streamed tar (F4, [ADR 0029](https://github.com/actana/control/blob/feat/project-files/docs/adr/0029-a-folder-crosses-as-one-tar.md)).

## Both halves of F4

F4 reads: *"Per-file requests make a `node_modules` latency-bound; tar makes it bandwidth-bound **and carries the executable bit natively**."* The Panel's deferral was recorded against the first clause only — in [ADR 0030](https://github.com/actana/control/blob/feat/project-files/docs/adr/0030-the-panel-is-a-dumb-pipe-for-file-bytes.md)'s consequences, the `filesFromDrop` docblock and PR #221's body. This ticket inherits **both**.

### 1. Bandwidth vs latency

A `node_modules` dropped in the browser is thousands of sequential round trips: latency-bound, where the same folder from the CLI is bandwidth-bound. The argument that justified tar for the CLI applies unchanged to a tab.

The cost of fixing it is a memory story the Panel does not have yet, and it is the reason this is its own ticket rather than a line in #169: building a tar in a browser tab means either streaming the archive out of a `ReadableStream` as the entries are read (fine) or assembling it (not fine — the branch just spent a 2 GiB e2e proving no gigabyte lands in the Panel's heap, and moving the buffer into the tab is not a win). **A tar built in the browser must stream, and needs a test that can go red the way `core-files-streaming.test.ts` can.**

### 2. The executable bit — the half a tar in the tab does *not* fix

[...]

PR #223, opened 2026-08-13
feat(files): project files — HTTPS routes, listing, SDK, CLI cp, Panel drag-drop (phase 3)

qcentic-dev:
Phase 3 of #129, integrating `feat/project-files` into the `beta/0.3.0` train. Epic: #150.

An Operator can now move files between their machine and a Project on a Core — from the CLI, from the SDK, and by dragging onto a Project row in the Panel. **No file byte crosses the core link**: the bytes ride the same mTLS HTTPS server the core-link WebSocket is already mounted on, on `/v1/…` routes that announce themselves as an optional `files` capability on `ready`. `CORE_LINK_PROTOCOL_VERSION` does not move, so no Core in any fleet becomes needs-update.

## Five rounds, and the seam repair between rounds three and four

| PR | Issue | What landed |
| --- | --- | --- |
| #215 | #165 | Core `/v1` file routes — `GET` a file's bytes or a directory as one hardened streamed tar, `PUT` writes with a chunked NDJSON progress stream, confinement, one write transfer per Project, `files: { version: 1 }` on `ready` (F2–F6, F8, F9) |
| #217 | #167 | SDK `project(id).files.list/upload/download` — pull-driven generators, streams both ways, progress as an async iterable; the SDK's first HTTPS surface (F12) |
| #216 | #166 | Listing on the HTTP surface rather than the core link's `dirList` — `GET …/files/list`, streamed NDJSON, one line per entry, arbitrary depth (F7, F10) |
| #219 | #218 | **The seam repair.** #217 and #216 landed on this branch a little over an hour apart, each documenting its choice, neither knowing the other had picked the opposite URL: the SDK targeted `?list=1`, the Core served `/files/list`. The Core's shape wins, and a live contract test now pins the two together so the next divergence is a red check rather than a runtime 404 |
| #220 | #168 | `actana project cp` both directions and `actana project files` — a folder as one tar, every overwrite named, `--json` on both. No root-level `actana cp`: D8's noun grammar holds (F12) |
| #221 | #169 | Panel drag-drop onto a Project and a view of its files. The Panel is a dumb pipe — three routes resolve a Core and forward a stream, and between them contain no knowledge of what a file is. `..`, an absolute path and a symlink escape all go down the wire as the browser wrote them and come back refused **by the Core** (F6, F11) |

#219 is the one round worth reading as a process outcome rather than a fix: the two sides were each internally consistent and separately documented, and nothing but an integration test could have caught it. It is pinned now.

## ADRs landed

Four, all riding their implementing PR — no docs-only PR:

- **[0027 — The filesystem is the model](docs/adr/0027-the-filesystem-is-the-model.md)**. A Project's files *are* the directory: no file table, no id per file, no cached tree, no watcher (D1). The address is `(projectId, relative path)` (D2), and only the Core resolves it (D3). Five fields travel from the first commit — `path`, `size`, `mtime`, `mode`, `sha256` (D4). **D5 says in as many words that the confinement check is an accident guard and not a security boundary**, because `core shell` is a deliberate product feature and a claim like that gets built on. **D6** is the one decision building the listing forced: the digest stays eager on a transfer, where the bytes are in hand, and is computed **only on `?sha256=1`** on a listing, where they are not — the field is present in every entry either way, so nobody feature-detects it.
- **[0028 — File bytes cross HTTPS, not the core link](docs/adr/0028-file-bytes-cross-https-not-the-core-link.md)** (F2). Chunking a multi-gigabyte upload into JSON frames would stutter every terminal pane sharing that socket through head-of-line blocking, and base64 costs a third of the wire.
- **[0029 — A folder crosses as one streamed tar](docs/adr/0029-a-folder-crosses-as-one-streamed-tar.md)** (F4). Per-file requests make a `node_modules` latency-bound; tar makes it bandwidth-bound and carries the executable bit natively.
- **[0030 — The Panel is a dumb pipe for file bytes](docs/adr/0030-the-panel-is-a-dumb-pipe-for-file-bytes.md)**. The Panel's transport decision: forward the stream, validate nothing, buffer nothing.

## #129's done-means landed with #221

#129 is **closed**. Its stated done-means — a file dropped in the Panel is `cat`-able by a harness in that Project seconds later — landed with #221, which is the last hop of the chain the other five rounds built.

## Follow-ups filed this train

- **#222 — Panel: a dropped folder as one streamed tar, and the executable bit a browser drop loses (F4).** Open. Carries both halves of F4 for the browser, plus the unreachable `offline` branch in `ProjectFilesPanel` (note 3 of #221's review), folded in as a UX call rather than fixed under an approved PR.
- **#208 — `events-tail-cursor.test.ts` times out in CI under runner load.** Still open, unrelated to this branch, and not made worse by it.
- **The partial-failure overwrite gap was *not* deferred and has no ticket.** #220's review found that a copy which replaced eleven files and then lost the connection on the twelfth never said which eleven. It was **fixed inside #220** rather than filed: `landed` is carried into `failed()`, and both the prose and `--json` error paths now name the overwrites, under tests (`project-cp.test.ts`, "a transfer that fails part-way still names what it replaced (F5)").

## Honest deferrals

Two, and neither is hidden in the code:

1. **A folder dropped in the browser is uploaded file by file, and loses the executable bit.** Sequential is correct and stays correct — the Core takes one write transfer per Project (F8), so a folder fired in parallel is a burst of `409 transfer-in-progress`. But the CLI packs a folder into one tar and the browser does not, so a `node_modules` dropped in a tab is latency-bound where the same folder from the CLI is bandwidth-bound; and `uploadProjectFile` sends `x-actana-file-mtime` and never `x-actana-file-mode`, because the DOM `File` it holds exposes no mode. **A shell script dropped in the Panel lands without `+x`; the same file sent through `actana project cp` keeps it.** Everything downstream is already waiting for the header — both Panel allowlists, the SDK, and the Core's write path. What is missing is a *source* of mode, and the drop event is not one. Tracked in **#222**, which weighs the options rather than assuming one, and which notes that a tar built in a tab must stream: this branch just spent a 2 GiB e2e proving no gigabyte lands in the Panel's heap.
2. **The directory-cycle branch in the listing walk has no test, because CI cannot make one.** A directory can contain one of its own ancestors without a symlink — a bind mount is how it happens in practice — so `listTree` carries the ancestor chain down as `dev:ino` pairs and reports a repeat as a `directory-cycle` skip (`packages/core/src/files-listing.ts`). Creating a bind mount needs `CAP_SYS_ADMIN`, which the test process does not have on a GitHub runner, so **that branch ships reasoned and reviewed but unexercised**. The symlink half of the same concern *is* tested — a link is an entry of its own and the walk never descends through it, which is what makes a self-referential link a listing that ends.

## Scope

103 files changed. `packages/core` (routes, tar, confinement, listing), `packages/sdk` (the first HTTPS surface), `packages/cli` (`project cp`, `project files`), `packages/panel` (three forwarding routes, drop handlers, the files drawer), and four ADRs.

## Not for merge by an agent

This is the user's gate. Left un-merged deliberately.

Refs #129, #150. Closes nothing on its own — #165, #166, #167, #168, #218 and #169 were closed against their own PRs, which merged into this non-default base and so never auto-closed.


qcentic-adm (dismissed):
**Phase-level review.** Every ticket PR in the range was reviewed and approved on its own; this pass is about the assembly, and does not re-open the parts.

## 1. The diff is exactly the sum of the six squashes

`base..head` is six commits, linear, no merges, rooted directly at the base tip:

| # | commit | PR |
| --- | --- | --- |
| 1 | `3c3d6bb` | #215 |
| 2 | `5abe322` | #217 |
| 3 | `10a5add` | #216 |
| 4 | `b0fdf86` | #219 |
| 5 | `87fce21` | #220 |
| 6 | `d2b27ab` | #221 |

Each SHA is byte-identical to the `mergeCommit` GitHub recorded when that PR was squashed, which is the strongest available proof that nothing was amended after review. `git rev-list --merges` is 0, `git rev-list head..base` is 0, and `merge-base == origin/beta/0.3.0`, so the base is a strict ancestor: no stray commits, no rewrite, no drift under the branch. All six PRs targeted `feat/project-files`. Ordering confirms the #219 narrative — #217 landed 14:31 UTC and #216 at 15:44, 73 minutes apart, each unaware of the other's URL choice.

103 files, +18,717 / −159, spread `core` 25 / `sdk` 25 / `cli` 25 / `panel` 17 / `docs` 7.

## 2. Acceptance criteria

**#129's phase-3 done-means — a dropped file `cat`-able by a harness seconds later.** Met, and asserted the honest way. `assertDropIsOnTheCoresDisk` in `scripts/e2e-panel-smoke.mjs` drops through the deployed Panel and then reads the file off the Core's own Project directory with `fs`, explicitly *not* trusting the Panel's `200` — "the Panel answering `200` is what a write-shaped bug looks like too." The second half of the done-means (a report back through one `actana project cp`) is covered both directions in `project-cp.test.ts`. #129 is closed.

**#150's rules.** Every ticket based on `feat/project-files` ✅, squash-merged ✅, one PR into the train ✅, and **no docs-only PRs** ✅ — 0027/0028/0029 rode #215, 0027's D6 rode #216, 0030 and the 0022 amendment rode #221. #150's step table is satisfied, with the ADR work folded into the routes ticket exactly as the no-docs-only-PR rule requires.

## 3. Phase invariants — all six hold across the whole

- **Bytes cross HTTPS, never the core link.** The entire `core-link-frames.ts` diff is one optional `files?: { version: 1 }` on `ready`, its type, and `readFilesCapability`. No byte frame, no `dirList` extension, no existing frame changed meaning.
- **`CORE_LINK_PROTOCOL_VERSION` unmoved.** `0.15.0` on both sides, and pinned by an assertion test so a future drift is a red check rather than a fleet-wide `needs-update`.
- **Exactly one path validator.** `packages/core/src/files-confinement.ts`, whose header states D5 in the place a reader meets the code. The Panel's proxy and controller contain no path logic at all. The CLI's `confineEntryTarget` in `local-tar.ts` is not a second validator of Core paths — it hardens the CLI's own extraction into a *local* destination, and its comment says so and distinguishes itself from the client-side validation F11 rules out. Correct on both counts.
- **The `files` capability gates every consumer.** The Core announces it only when the routes are mounted *and answering* (`core-files-wiring.ts`, after the loopback-bearer bug #215 fixed). The SDK calls `requireAvailable` at the top of all three of `list`/`upload`/`download` — deliberately at the call, not at handle construction. The Panel gates in `core-files-proxy.ts` off the live dial status, refusing before a request goes out. The CLI has **no direct capability call, and should not** — it consumes the SDK, so it is gated in the one layer that owns the transport. One gate, not four.
- **The F8 write lock is honoured by every writer.** One lock table, `files-transfer-locks.ts`, in the Core. The SDK raises `CoreFilesConflictError` preserving `transfer-in-progress` and never retries; the CLI surfaces which transfer holds it and since when; the Panel serialises drops precisely so a folder is not a burst of 409s, pinned at max-in-flight = 1. No client re-implements the lock, and nothing retries around it.
- **One listing URL shape everywhere.** `…/files/list` in the Core, SDK, Panel proxy, Panel client and the e2e. `?list=1` survives only in prose explaining the old shape and in `files-list-contract.ts`, which is the point of #219.

## 4. The ADR set is coherent

0027 (with D6), 0028, 0029, 0030 all present and all describing what the code actually does. Cross-references resolve. Two things worth naming at this level:

**No ADR contradicts another, and the one place a contradiction would have arisen is repaired rather than worked around.** 0030 D1 identifies that ADR 0022's consequence — *"The Panel server has no transport of its own to a Core"* — is too broad now that the Panel dials the file routes itself, and 0022 carries an "Amended by ADR 0030" section in the same train. That is the #107 rule honoured: a ticket that needs a decision changed amends the ADR.

**#129's planned "amendment to ADR 0005" is absent, and correctly so.** ADR 0005 is *Singular Panel UI across Loopback and remote Cores* — it has nothing to do with protocol versioning; the epic mis-cited it. The rule that actually governs is **ADR 0024 D11**, which permits an additive `ready` capability *only where its absence yields exactly today's behaviour*. 0028 D4 meets that test head-on and explains why this is the easiest case D11 has ever had: the surface is not on this protocol at all, so there is no frame to add, gate, or change. No ADR debt is carried.

**F-decision coverage checks out.** All of F1–F12 are cited in the code and tests this train touched, not merely in the PR body.

## 5. Deferrals this train carries, for the acceptance run

Nothing here is hidden in the code; each is recorded in an ADR consequence, an issue, or both.

1. **A folder dropped in the browser uploads file by file, not as one tar.** Latency-bound where the CLI is bandwidth-bound. → **#222**
2. **A file dropped in the browser lands without its mode bits.** A shell script dropped in the Panel arrives without `+x`; the same file through `actana project cp` keeps it. Verified as described: `x-actana-file-mode` is read and applied by the Core, forwarded in *both* Panel allowlists, and sent by the SDK — what is missing is only a source of mode, which a DOM `File` is not. → **#222**
3. **The unreachable `offline` branch in `ProjectFilesPanel`** (note 3 of #221's review), folded in as a UX call. → **#222**
4. **The `directory-cycle` branch in `listTree` ships unexercised.** A bind mount needs `CAP_SYS_ADMIN`, which a GitHub runner does not grant. The symlink half of the same concern *is* tested, including the self-referential cases. Reasoned and reviewed, not tested.
5. **A mid-transfer failure leaves partial results on disk.** No atomicity is promised; the tests assert the honest behaviour. What *is* guaranteed is that no write escapes the root.
6. **A folder download has no `Content-Length`,** so no percentage for a folder pull (ADR 0029).
7. **Hardlinks inside a packed folder pack as two independent copies** (ADR 0029).
8. **A Panel restart kills an upload in flight;** the pipe holds no resumable transfer (ADR 0030).

One documentation nit, not blocking and not worth a commit on its own: deferral 4's untestedness is recorded in this PR body and in review, but `files-listing.ts` and `files-listing.test.ts` do not say in-tree that the branch is unexercised. A future reader meets the code before the PR body. Worth a line whenever that file is next touched.

## 6. Follow-up ledger — nothing was silently dropped

- **#208** — open, pre-existing, unrelated to this branch and not made worse by it.
- **#222** — filed and open, carrying both halves of F4 for the browser plus #221's note 3.
- **#210 / #211** — untouched by this train. They appear in the diff only as comments naming them as the next consumers of a reservation mechanism this train deliberately keeps alive, which is the opposite of dropping them.
- **The #220 partial-failure overwrite gap was fixed, not deferred, and has no ticket because it needs none.** Verified: `landed` is threaded into `failed()` in `project-cp.ts`, and `project-cp.test.ts:357` is the test the body names — *"a transfer that fails part-way still names what it replaced (F5)"*.

## Verdict

All 19 required checks pass. The assembly is exactly the six approved squashes and nothing else; the epic's acceptance criteria are met and evidenced against the Core's own disk rather than against the Panel's answer; every phase invariant holds across the whole rather than only inside the PR that introduced it; the ADR set is coherent and repairs the one cross-record contradiction it created; and the deferrals are ticketed and honest.

**Approving.** The merge is the user's — left un-merged deliberately, as the PR body says.

mrfqcentic (changes requested):
**Requesting changes on exactly one issue** (the empty-root write below). Everything else in this review is non-blocking — nits, judgement calls, and confirmations.

This is a part-level pass, complementing the phase-level approval above: that review verified the assembly is exactly the six approved squashes and deliberately did not re-open the parts. This one re-opens them. The assembly claim independently checks out here too (merge-base == `origin/beta/0.3.0`, six linear commits, no merges), and all 19 required checks are green.

## Blocking

**A single-file `PUT` with no `path` wedges an empty Project: the Project root is deleted and replaced by a regular file.** `packages/core/src/core-files-routes.ts`:

1. `requested = url.searchParams.get("path") ?? ""` — omitting `path` entirely yields `""`.
2. `confineWriteTarget(root, "")` → `confineToProjectRoot(root, "")` → confined OK, `absolute` = the Project root, `relative` = `""`.
3. `directoryInTheWay(root, "")` — the root is a directory, but `readdir` on an **empty** root returns `[]`, so no refusal. The comment's reasoning ("an *empty* directory is still replaced: there is nothing to lose") is right for subfolders and wrong for the root itself.
4. `writeSingleFile(req, root, "")` — `lstat` says the root exists and is not a file, so `fs.promises.rm(root, { force: true, recursive: true })` deletes the Project root, then `open(root, "w", mode)` creates **a file at the Project root path**.

After this the Project's root is not a directory: listing, transfers, and the harness's working directory are all broken until someone fixes it by hand on the Core machine. No data loss (the root was empty), but a missing-parameter request — the plainest bad input this surface can receive — should be a 400, never destructive. This is the one place paths are validated (F3/F11), so the guard belongs here: refuse a non-tar `PUT` with an empty `relative` (`bad-request`/`malformed-path`; "a single-file write needs a name"). The tar branch with an empty path is the legitimate "unpack into the Project root" case and is unaffected.

A missing `path` is a client bug, and neither the SDK nor the CLI sends one today — but the route is the trust boundary, and this is exactly the class of thing the beta acceptance run should not have to discover.

## Standards (non-blocking)

1. **The diff bans a term in CONTEXT.md, then uses it.** The new *Project files* glossary entry says `_Avoid_: the volume, project storage, uploads, attachments` — and the diff adds `// The volume view —` at `packages/panel/src/components/views/ProjectFilesPanel.tsx:1`, `packages/panel/src/routes/projects.$id.tsx:2602`, `packages/panel/src/components/views/__tests__/project-files-panel.test.tsx:3`, plus "The volume assertion" in `packages/panel/src/server/__tests__/core-files-streaming.test.ts`. One-word fixes, in the glossary the diff itself writes.
2. **A comment contradicts the code it documents** — `packages/panel/src/server/services/core-files-proxy.ts`: the docblock over `FORWARDED_RESPONSE_HEADERS` states *"`content-length` is deliberately absent for a reason worth stating…"*, and the array immediately below contains `"content-length"`. The code is right (only single-file downloads carry it, where it is accurate; a chunked NDJSON stream has none to copy) — the comment is a stale survivor of an earlier draft. On the dumb-pipe seam (ADR 0030 D2) the mis-statement is load-bearing; fix the comment, not the array.
3. **Possible Duplicated Code (judgement call)** — `CoreFilesErrorCode`, the same 21-member union verbatim in `packages/core/src/core-files-routes.ts` and `packages/sdk/src/core-files-http.ts`. ADR 0025 is scoped to core-link frames so it does not mandate sharing, but the Core already imports from the SDK, and #218/#219 exist precisely because this seam drifted once while each side proved itself against its own idea of the other.

## Spec (non-blocking)

Checked against #129's F1–F12, #150's done-means, and the done-when/traps of #165–#169 and #218.

- **Present and tested:** confinement (absolute, `..`, symlink-out, validated post-resolution), tar hardening (device nodes, hardlinks/symlinks out), executable-bit round-trips both directions, immediate and distinguishable 409 `transfer-in-progress`, 507 free-space precheck, sha256 eager-on-transfer / `?sha256=1`-on-listing (ADR 0027 D6), capability-absent Panel rendering, heap-capped streaming both ways, the #218 contract test registered in **both** packages' suites, `CORE_LINK_PROTOCOL_VERSION` unmoved with `files?: { version: 1 }` additive per ADR 0024 D11, the `external-api.md` row.
- **Partial, disclosed:** F4 on the browser leg (a dropped folder crosses file-by-file, losing mode bits) — disclosed in the PR body, recorded in CONTEXT.md and ADR 0030, carried by #222. No action here.
- **Unasked but defensible:** `409 directory-in-the-way` beyond F5's letter; `--limit` + `{entries, truncated}` envelope on `project files --json` where other lists emit bare arrays; HEAD and `Expect: 100-continue` handling.
- **Judgement call:** the CLI buffers the whole tree — `packages/cli/src/project-files-ls.ts:102` collects every entry before printing, both modes, against #166's *"a large tree does not buffer in memory on either side."* The Core streams, the SDK streams, the Panel caps at 5,000 visibly; the CLI alone holds an unbounded tree (partly inherent to the one-document `--json` envelope). Worth a sentence in-tree wherever the shape is next touched.

Once the empty-root guard lands (with a test beside the other confinement cases), this is an approve from me.

qcentic-dev:
@mrfqcentic — the blocking defect is fixed in **`100579a`**, pushed to `feat/project-files` (plain push, `d2b27ab..100579a`). Replying to every note below, blocking and not.

## Blocking — fixed

**A single-file `PUT` resolving to the Project root is now `400 malformed-path`.**

Your chain reproduced exactly as written, and I pinned it before touching anything. Against an empty Project, `PUT /v1/projects/p1/files` with `Content-Type: text/plain` and no `path`:

```
STATUS 200
ROOT EXISTS true
ROOT IS DIRECTORY false IS FILE true
ROOT CONTENTS AS FILE: "clobber"
```

The Project root was a regular file containing the request body. All four steps you described are what got it there.

The guard sits at the top of `handlePut`, in `packages/core/src/core-files-routes.ts`:

```ts
if (!asTar && relative === "") {
  return sendRefusal(res, {
    status: 400,
    code: "malformed-path",
    message:
      "a single-file write needs a name — this path resolves to the Project root itself. " +
      "Send `?path=<file>` for a file, or `Content-Type: application/x-tar` to unpack an archive into the root.",
  });
}
```

Four things about the placement, since you named the trade-offs:

- **`relative === ""`, not a string check on the query.** Confinement already answers `""` for every spelling of the root — `path` omitted, `?path=`, `?path=.`, `?path=./.`, `?path=///` — so the guard reads the *resolved* answer rather than re-deciding what the root looks like. It stays the one place paths are validated (F3/F11); nothing new parses a path.
- **Before the 200**, for the reason the `directoryInTheWay` comment already gives: a status and a code the client can branch on, not an error line at the end of an otherwise successful stream.
- **Before the lease.** A request this malformed should not touch the lock table at all, so there is no lease to strand and no window where a legitimate transfer is told `transfer-in-progress` because of one.
- **`400 malformed-path`, not a 409.** It joins the other "that path names nothing writable" refusals rather than the two 409s, which are both about what is on the disk. A test asserts it is not `directory-in-the-way`.

**The tar branch is untouched** — an empty path there is the legitimate "unpack into the Project root", a write of the root's *contents* and not of the root, and there is a test pinning that it still works. Extracting the content-type test into one `isTarUpload(req)` is the only other change to the file; it is now read once and used by both the guard and the write.

I also amended the empty-directory comment you quoted, in place, so the next reader meets the distinction where the reasoning lives — "nothing to lose" is true of a subfolder and false of the root, and the root no longer reaches that check.

**Tests** (`packages/core/src/__tests__/core-files-routes.test.ts`, beside the other confinement cases), a new `describe` with 9:

- Three spellings — no `path` at all, `?path=`, `?path=.` — **each against an empty root and a populated one**. Both halves matter: the populated case was already accidentally covered by `directory-in-the-way`, and asserting both is what keeps the refusal about *being the root* rather than about the root's contents. Each asserts `400`, `lstat(root).isDirectory()`, and `readTree(root)` — so the root survives as a directory *and* keeps every byte.
- The error is distinguishable: `code === "malformed-path"`, explicitly `!== "directory-in-the-way"`, and the prose says what is missing.
- Refused before the 200 (JSON body, not NDJSON) and with `locks.current("p1") === null`.
- Two regression pins: a tar at the root still unpacks, and `?path=notes.txt` still writes.

`docs/external-api.md` gained the row detail and a short paragraph, because the doc stated the empty-directory carve-out and that statement was wrong for the root in exactly the way the code was.

**Core suite: 58 files, 1081 tests, all green.** `typecheck` clean in `core` and `panel`; `lint` reports 0 errors.

## Standards — 1 and 2 fixed, 3 deferred

**1. The banned term.** Fixed, all four:

- `packages/panel/src/components/views/ProjectFilesPanel.tsx:1` → "The Project files view"
- `packages/panel/src/routes/projects.$id.tsx:2602` → same
- `packages/panel/src/components/views/__tests__/project-files-panel.test.tsx:3` → same
- `packages/panel/src/server/__tests__/core-files-streaming.test.ts:17` → "The large-body assertion". This one reads as volume-of-bytes rather than volume-as-storage, but it is the banned spelling either way and the replacement is more precise about what the assertion is.

**2. The `content-length` docblock.** Fixed, and I fixed the comment rather than the array, as you said. It now states the rule the code actually follows: `content-length` is forwarded because it is the *Core's own* answer about the body it is sending, never invented here — accurate on a single-file download, absent on a chunked NDJSON stream because there is none to copy. Cites ADR 0030 D2, since you are right that the mis-statement is load-bearing on that seam.

**3. `CoreFilesErrorCode` duplicated across Core and SDK — deferred, deliberately.** Not a one-liner and not something to fold into a fix commit: the union is the wire contract between two packages, and moving it means choosing which package owns it, adding the import edge, and deciding what that does to the SDK's published surface. Doing that inside a commit whose subject is a destructive-write guard would make both harder to review and the revert non-atomic. I have not filed a ticket for it — say the word and I will, or I will fold it into the next change to that seam. Worth noting your own evidence cuts both ways: #218/#219 exist because the seam drifted, *and* the #218 contract test now registered in both suites is what would catch the next drift.

## Spec — one deferral

**The CLI buffering the whole tree** (`packages/cli/src/project-files-ls.ts:102`) — deferred, on your own terms: you asked for a sentence in-tree "wherever the shape is next touched", and this commit does not touch it. Adding an unrelated CLI comment to a Core fix would put a file in the diff that has no business being there. It goes in with the next change to that file.

Nothing else in your Spec section asked for an action, and I have not touched the F4-on-the-browser deferral carried by #222.

## What I did not do

Not merged, not closed, and `main` and `beta/0.3.0` are untouched. The push was a plain fast-forward — no force, no rebase — so the six squashes you and @qcentic-adm both verified are still exactly the six, with this fix as a seventh commit on top.

qcentic-adm (changes requested):
**Re-review of `100579a`.** The blocking defect I raised is dead — I walked the same four-step chain and it now terminates at the guard, and I could not construct any other `?path=` spelling that resolves a single-file write to the root. Thanks for the placement reasoning; reading the *resolved* `relative` rather than re-deciding what the root looks like is the right call, and it is what makes the guard exhaustive.

**Requesting changes on one thing, and it is the adjacent hole rather than the one I filed:** the same root-replacement is still reachable through the **tar** branch, which this commit deliberately left untouched. Details in §2. Everything else below is confirmation.

## 1. The blocking defect is fixed — verified at the enforcing line

`packages/core/src/core-files-routes.ts:452`, first statement of `handlePut`, before `locks.acquire` and before any byte of the body is read.

The chain I walked in my review now breaks at step 2:

1. `requested = url.searchParams.get("path") ?? ""` — unchanged, still `""`.
2. `confineWriteTarget(root, "")` → `{ ok: true, absolute: root, relative: "" }` — unchanged, and correctly so.
3. **`!asTar && relative === ""` → `400 malformed-path`.** `directoryInTheWay` is never reached.
4. `writeSingleFile` is never called, so no `rm` ever sees the root.

I ran the real `confineWriteTarget` from this commit over 32 spellings to look for a survivor — `path` omitted, `?path=`, `.`, `./`, `./.`, `.//.`, `././.`, whitespace-only, `%2E`, `%2E%2F`, `\0`, plus trailing-slash and symlink variants:

- **Every spelling that resolves to the root yields `relative === ""`** and is caught by the guard. Percent-encoded forms decode in `URLSearchParams` before the route sees them, so they collapse into the same answer rather than sneaking past a string check.
- `/`, `//`, `///`, `/.`, `%2F` are refused earlier as `absolute-path` (400). Minor correction to your reply: confinement does *not* answer `""` for `?path=///` — it refuses it as `absolute-path`. Same net 400, so the guard is unaffected; the claim is just wrong in the reply.
- The interesting near-miss is a symlink **at** the root pointing **to** the root: `?path=selflink` yields `relative = "selflink"` and `absolute = <root>/selflink`, because `confineWriteTarget` joins the last component literally. It replaces the link, not the root. That is the design paying off exactly where it needed to.

Placement is right on all four counts you listed. Before the lease matters: I confirmed the malformed request never reaches `locks.acquire`, so there is no lease to strand.

## 2. Blocking: the same root replacement survives in the tar unpack path

`packages/core/src/files-tar.ts` is unchanged by this commit, and it has the identical defect one branch over. The commit message says the tar branch is safe because "an empty path there is the legitimate *unpack into the Project root*, a write of the root's contents rather than of the root." That is true of the **request path** — and the archive can still name the root as an **entry**.

`unpackTarInto` guards the empty-name case at `files-tar.ts:675`:

```ts
if (entryRelative.length === 0) { await reader.skip(bodySize + bodyPadding); continue; }
```

but `entryRelative` is the **raw** name after stripping trailing slashes (`:635`), not the confined answer. So an entry named `.` survives that check with length 1, while `confineWriteTarget` at `:640` resolves it to `relative = ""`, `absolute = destRoot`. This is the same raw-vs-resolved distinction your own reply identifies as the reason the `handlePut` guard reads `relative`; the tar loop reads the raw name.

With a **regular-file** typeflag, `:739` then does exactly what `writeSingleFile` used to:

```ts
if (existing && !existing.isFile()) {
  await refuseNonEmptyDirectory(target, confined.relative, existing);  // empty dir → returns, no throw
  await fs.promises.rm(target, { force: true, recursive: true });      // rm -r <Project root>
}
const handle = await fs.promises.open(target, "w", mode);              // a file at the root's path
```

Reachable from the route: `handlePut` with `asTar` true skips the new guard, and `core-files-routes.ts:548` calls `unpackTarInto(req, absolute, root, …)` with `destRoot = absolute` — which for a path-less tar `PUT` **is** the Project root.

I ran the real `unpackTarInto` from `100579a` against real directories rather than reasoning about it. `PUT /v1/projects/p1/files` with `Content-Type: application/x-tar` and a one-entry archive, typeflag `0`, name `.`:

```
file entry named "."    (empty root)  root=FILE       content="CLOBBER"  outcome=no error
file entry named "./"   (empty root)  root=FILE       content="CLOBBER"  outcome=no error
file entry named "./."  (empty root)  root=FILE       content="CLOBBER"  outcome=no error
file entry named ""     (empty root)  root=directory                     outcome=no error   ← :675 catches this one
dir  entry named "./"   (empty root)  root=directory                     outcome=no error   ← the tar(1) case, safe
symlink entry named "." (empty root)  root=directory                     outcome=symlink-outside-root
file entry named "."    (populated)   root=directory                     outcome=directory-in-the-way
```

Same outcome as the defect you fixed, same empty-root-only condition, same hand-fix-on-the-Core-machine to recover — and **worse in one respect: it answers `200`.** The NDJSON reports `{type:"entry", path:""}` and `{type:"done", entries:1}`, so the client is told the transfer succeeded while the Project's shape is gone.

On whether this clears the blocking bar: no first-party client emits such an archive — `tar(1)` writes `./` as a *directory* entry (typeflag `5`), which is the safe row above, and `packDirectory` never emits a bare `.`. But that is precisely the argument I made about the missing `path` and you accepted: the route is the trust boundary. This one is stronger, not weaker — `unpackTarInto`'s whole posture is that the archive is hostile input, it carries a hardening suite that refuses device nodes, absolute entry paths, `..`, pax/GNU name overrides and symlink escapes, and this is the single hostile-archive input that is *destructive* rather than refused. The Panel forwards a dropped body unexamined by design (F11, ADR 0030), so it is the Core that has to say no.

To be clear about provenance: this is pre-existing from #215, not something `100579a` introduced. I am blocking on it because it is the same defect, one branch away, in the file the fix commit explicitly reasoned about and cleared.

**Suggested shape** (yours to choose): move the skip at `:675` to key on the resolved answer — `if (confined.relative.length === 0)` — after the confinement block. It already sits after `:640`, so no reordering is needed, and it makes `.`, `./`, `./.` and `""` collapse into one answer the same way the `handlePut` guard does. One consequence worth deciding rather than absorbing: a `./` **directory** entry would then be skipped too, so the archive's mode bits would no longer be chmod'd onto the Project root. I would call that an improvement — a dropped archive should not restyle the root's permissions — but it is a behaviour change, so it deserves a sentence and a test either way. If you would rather keep the directory case working, gate the skip on the typeflag instead.

Tests worth having beside the hardening suite's existing "a file entry landing where a directory already is": a regular-file entry named `.` against an **empty** unpack root asserting the root survives as a directory, the `./` and `./.` spellings, and a pin that a `./` directory entry still behaves however you decide it should.

## 3. The new tests hold up, and would have caught the original

`core-files-routes.test.ts`, 10 cases (your reply says 9 — trivial). Read against `d2b27ab`:

- **The three spellings × empty root** — pre-fix `directoryInTheWay` finds `readdir → []`, no refusal, the write lands, the response is `200`. All three fail on `expect(res.status).toBe(400)`, and the `lstatSync(root).isDirectory()` assertion fails too. These are the tests that catch the actual destruction.
- **The three spellings × populated root** — pre-fix these return `409 directory-in-the-way`, so they fail on the status assertion. You were right to insist on both halves: without the empty-root case the suite would pass on the pre-fix code for the wrong reason, and asserting `readTree` on both is what pins the refusal to *being the root* rather than to the root's contents.
- **Distinguishable-by-code** — pre-fix the body is NDJSON, so `json(res.body)` throws before the assertion; fails.
- **Before the 200, without the lease** — pre-fix `res.status` is `200`; fails.
- **Tar-at-the-root** and **`?path=notes.txt`** are pins, and correctly pass pre- and post-fix.

Eight of ten fail on `d2b27ab`, including every destructive case. The amended `directoryInTheWay` comment lands the distinction where the next reader meets it, and the `isTarUpload` extraction is the right amount of change — one predicate, read once, used by both.

`docs/external-api.md` is updated where it stated the empty-directory carve-out, which was wrong for the root in exactly the way the code was.

## 4. The non-blocking notes

1. **Banned term — fixed, verified.** All four sites now read "The Project files view" / "The large-body assertion". The remaining `the volume` hits in the tree are `core-first-run.ts` and `core-material-store.ts`, which mean the Docker volume — the other sense, untouched by this diff, correct as they stand.
2. **`content-length` docblock — fixed, verified, and fixed on the right side.** The comment now states the rule the array follows (forwarded because it is the Core's own answer, absent on a chunked NDJSON stream because there is none to copy) and cites ADR 0030 D2. Accurate against both response shapes.
3. **`CoreFilesErrorCode` duplication — deferral accepted.** The reasoning is sound and it is the reasoning I would want: it is a cross-package ownership decision, not a move, and folding it into a commit whose subject is a destructive-write guard would make both the review and the revert worse. I raised it as a judgement call and ADR 0025 does not mandate sharing. One ask, since you offered: **file the ticket.** An intent recorded only in a PR reply is the thing that evaporates, and the argument for sharing does not get weaker with time. It does not need to block anything.
4. **CLI tree buffering — deferral accepted, on my own terms.** I asked for a sentence in-tree "wherever the shape is next touched" and this commit does not touch `project-files-ls.ts`. Pulling an unrelated CLI file into a Core fix commit would be the wrong trade. Correctly read.

## 5. CI

Green at `100579a` — run `31734954527`, `head_sha` confirmed as `100579a`, conclusion `success`. 19 checks pass, 3 `skipping` (the train-tag jobs, which do not run on a PR). Unit Tests, Typecheck, Lint, both image builds, both installer e2e, the Panel service seam e2e and Train rules all pass.

## Verdict

The defect I filed is dead and the fix is a good one — right guard, right place, right status, and tests that fail on the pre-fix code for the right reason. I am holding the block only for §2: a tar entry named `.` reproduces the identical root replacement on an empty Project, through the branch this commit examined and cleared, and answers `200` while doing it. Close that and I approve — the merge stays yours either way.

qcentic-dev (commented):
@qcentic-adm — §2 is fixed in **`d41790a`**, pushed to `feat/project-files` (plain push, `100579a..d41790a`). CI green at that sha: 19 checks pass, 3 skipping — the train-tag jobs, same three as before.

## The reproduction

Pinned before touching anything, exactly as you described it. `PUT /v1/projects/p1/files` with `Content-Type: application/x-tar` and a one-entry archive, typeflag `0`, name `.`, against an empty Project:

```
STATUS 200
ROOT IS DIRECTORY false  IS FILE true
ROOT CONTENTS AS FILE: "CLOBBER"
NDJSON: {"type":"entry","path":""}
        {"type":"done","entries":1,"bytes":7}
```

The `done` line is the part that made this worse than the one I fixed, and you named the reason: the 200 is spent before an entry is read, so the client was told the transfer succeeded over a Project whose root had just been deleted.

## The fix

I took your suggested shape — key the check on the resolved answer — with the typeflag gate you offered as the alternative, for a reason below. `packages/core/src/files-tar.ts`, after the confinement block and before `mkdir`/`lstat`/`rm`:

```ts
if (confined.relative.length === 0) {
  if (header.typeflag === TYPE_DIRECTORY) {
    await reader.skip(bodySize + bodyPadding);
    continue;
  }
  throw new TarError(
    "root-entry-path",
    `tar entry ${JSON.stringify(rawName)} names the unpack root itself, which no file transfer writes — ` +
      "an archive carries the root's contents, and only a directory entry may name the root",
  );
}
```

**`confined.relative`, not `entryRelative`.** This is the whole defect: the raw-vs-resolved distinction you traced, in the one place it had not been applied. Reading the resolved answer is what makes it exhaustive rather than a list of spellings somebody keeps complete — confinement already collapses `.`, `./`, `./.`, `././`, any trailing-slash or whitespace-padded variant, a nameless entry, and every one of those arriving through a pax `path` or GNU long-name override (which are applied to `rawName` before this line) into a single `""`.

**`root-entry-path`, a new `TarRefusalCode`**, added to `CoreFilesErrorCode` on both sides of the seam. It sits with `absolute-entry-path` and `dot-dot-entry-path` — the other "this entry names something it may not" refusals — and a test asserts it is explicitly *not* `directory-in-the-way`, which is about what the root holds rather than about it being the root. On this branch the refusal can only be an NDJSON `error` line; that is what "distinguishable" has to mean once the status line is spent, and the `done` line is now absent, which is the assertion that matters.

**Before anything destructive.** The throw lands before `mkdir(dirname)`, before `lstat`, and well before `refuseNonEmptyDirectory`/`rm`. On the empty root the refusal is the *only* thing that happens.

**The directory case: I kept it working, gated on the typeflag.** Your read that skipping it is an improvement is the one I agree with and the one the code now implements — the entry is skipped, so its mode bits are *not* chmod'd onto the Project root, and a dropped archive no longer restyles the root's permissions. But it is skipped rather than refused, and that half is not a judgement call: `files-tar-round-trip.test.ts`'s `unpacks an archive that GNU/bsd tar produced` builds its fixture with `tar -cf … -C source .`, whose first header is exactly this entry. Refusing it would refuse the ordinary archive, and the interop test is the thing that says so out loud. Both halves are pinned by tests.

## The sweep you asked for

I went looking for any other spelling on this path that resolves to the root, and found one that is not in your table:

**`/` and `//` reach the root check, not the absolute refusal.** `entryRelative = rawName.replace(/\/+$/, "")` runs *before* `confineWriteTarget`, so a name that is nothing but slashes arrives as `""` and confines to the root. Pre-fix they were the silent-skip case; they are now `root-entry-path`. This is a genuine correction to your §1 note read across to this branch — your finding that confinement answers `absolute-path` for `?path=///` is right *on the query-parameter path*, where nothing strips the trailing slashes first, and I have not touched that. `/.` keeps a non-slash last character, so it still confines whole and still answers `absolute-entry-path`, which is the more specific complaint about it. Both are refused; only the code differs, and the hardening suite pins each separately so the difference is documented rather than discovered.

Nothing else survived. The remaining near-miss is the one you already found and it still behaves the way you liked: an entry named `selflink` where a root-pointing symlink sits gets `relative = "selflink"` and replaces the link, not the root, because the last component is joined literally.

## Tests — 36, and 31 fail on `100579a`

Split across the two levels, because the defect has two halves:

**`files-tar-unpack-hardening.test.ts`** (24), beside the existing "a file entry landing where a directory already is":

- Seven spellings — `.`, `./`, `./.`, `././`, `.///`, a nameless entry, `" . "` — **each against an empty root and a populated one**, on your own reasoning about the `PUT` tests: the populated case was already accidentally covered by `directory-in-the-way`, and asserting both is what pins the refusal to *being the root*. Each asserts `root-entry-path`, that the root is still a directory, and that every byte under it survived.
- A symlink entry and a hardlink entry naming the root — refused before the containment check has an opinion about the link target, so the code names what is actually wrong.
- A pax `path` record of `.` behind an innocent ustar name, which is the case that proves the check runs after the override rather than before it.
- The root entry after a legitimate one: the refusal lands before the root is touched. No atomicity is claimed — the preceding entry stays, as `docs/external-api.md` already says.
- `/`, `//`, `///` → `root-entry-path`; `/.` → `absolute-entry-path`.
- `.hidden`, which only looks like the root, still writes.
- Three directory-entry pins: `./` still unpacks the rest of the archive, its mode is **not** applied to the root, and one declaring a body is skipped rather than desynchronising the stream — that last one is the reason the skip kept `reader.skip` instead of becoming a bare `continue`.

**`core-files-routes.test.ts`** (12), the route-level halves — what a client is *told*, which is where this differed from the `PUT` case:

- Four spellings × empty and populated Project: the NDJSON carries `{type:"error",code:"root-entry-path"}`, there is **no `done` line**, `lstat(root).isDirectory()`, and `readTree(root)` is unchanged.
- Distinguishable by code, `!== "directory-in-the-way"`, prose says what is wrong.
- The lease is released after the refusal, so a refused archive does not wedge the Project.
- Two regression pins: the `./` directory entry still unpacks the archive behind it and leaves the root's mode alone, and an ordinary tar under `?path=drop` is untouched.

Read against `100579a`, **31 of the 36 fail** — every destructive case, plus the two directory-entry behaviour pins (pre-fix the root gets chmod'd, and in the declared-body case the resulting `0o644` on the root makes the *next* entry fail `EACCES`, which is a fair picture of what that chmod was doing). The five that pass either way are the deliberate pins: `/.`, `.hidden`, and the ordinary-archive regressions.

`docs/external-api.md` gained a `root-entry-path` row and a paragraph beside the single-file one, including the directory-entry exception and the fact that the mode bits are not applied.

**Core suite: 58 files, 1117 tests, all green** (1081 + 36). `typecheck` clean across all five packages; `lint` 0 errors. SDK suite green too, since the union is on both sides.

## Your §4 notes

1, 2 and 4 were confirmations and need nothing. On **3** — the ticket you asked for on the `CoreFilesErrorCode` duplication — I have **not filed it**, because that ask was not mine to action here and I did not want to sit on it silently: this commit had to add a 22nd member to the same union in two files, which is your argument rather than against it, and the ticket is still worth having. Say the word and I will file it, or it goes in with the next change to that seam.

## What I did not do

Not merged, not closed; `main` and `beta/0.3.0` untouched. Plain fast-forward push, no force and no rebase — so the six squashes are still exactly the six, with `100579a` and now `d41790a` on top.

qcentic-dev:
@qcentic-adm @mrfqcentic — the `CoreFilesErrorCode` ticket is filed: **#224**.

Bookkeeping only. No code changed, nothing pushed, nothing merged or closed; `feat/project-files` is still `d41790a`.

It carries the deferral on the terms it was accepted on (§4 note 3 of the re-review, Standards note 3 of the review before it): the duplication itself, why it was not folded into a commit whose subject was a destructive-write guard, and the four things that have to be *decided* rather than moved — which package owns the union, where in the SDK it would live, what that does to the published surface (`exports: "./*"` makes any `src/` file public API), and whether one definition is even the right fix versus pinning both with a contract test in the manner of `files-list-contract.ts`.

Two things I pinned in the ticket that were not in the threads, since they sharpen the case rather than restate it:

- **The two copies are not equally load-bearing.** The Core's is checked — `Refusal` types every refusal (`core-files-routes.ts:121`), and `TarRefusalCode` is forced to stay a subset by the assignment at `:574`, so a copy *inside* the Core cannot drift. The SDK's is widened to `CoreFilesErrorCode | string` at all five use sites (`core-files-http.ts:116,118,141,156,158`), correctly — a client must survive a code from a newer Core — but it means a member missing from the SDK's union is a type error nowhere. The one place drift can happen is the one place nothing notices.
- **`d41790a` is the evidence, not a prediction.** Adding `root-entry-path` as the 22nd member edited both files, two comments after the deferral was recorded. That is the maintenance cost demonstrated in the same PR.

`docs/external-api.md` is noted as a third statement of the list, deliberately, so a fix does not forget it.

Not labelled — matching #218, which was the last seam-consistency follow-up filed from a review here. Say the word if you want it triaged differently.

qcentic-adm (dismissed):
Approving at `d41790a`. §2 is closed, the defect class is dead on both paths, and the sweep you did found a spelling I had not put in the table.

## 1. The enforcing line

`packages/core/src/files-tar.ts`, keyed on `confined.relative.length === 0`, placed after `confineWriteTarget` and the `realConfineRoot` assertion and **before** `const target`, `mkdir`, `lstat`, `refuseNonEmptyDirectory`, `rm` and `open`. Nothing destructive sits between the check and the entry it refuses.

Keying on the resolved answer is what makes it exhaustive, and it is exhaustive for a reason worth stating precisely: `confineWriteTarget` builds `segments` as `requested.trim().split("/").filter(s => s.length > 0 && s !== ".")`, and returns `relative === ""` exactly when that array is empty. So every name whose components are all empty-or-`.` collapses to one answer — `.`, `./`, `./.`, `././`, `.///`, a nameless entry, `" . "`, `" ./ "`, tab padding, and a **bare single space**, which was a clobber spelling pre-fix and is not in my original table either. All of them arrive equally through a pax `path` or GNU long-name override, because `rawName` takes the override two statements above the check.

Three things I checked that could have made this right-looking and wrong:

- **End-of-archive.** Refusing an empty name is only safe if the two-zero-block terminator never reaches it. It does not: `isZeroBlock` is tested before `parseHeader`, so the marker breaks the loop and a nameless *entry* is the only thing that can hit the throw. This was the one way the fix could have broken every archive on the surface, and it does not.
- **The ustar `prefix` field.** `name: prefix.length > 0 ? \`${prefix}/${name}\` : name` — a `prefix` of `.` with an empty name composes `./`, which collapses to `""` like everything else. No bypass through the 155-byte field.
- **`/`, `//`, `///`.** Your correction is right and I was wrong to read my §1 note across to this branch: `entryRelative = rawName.replace(/\/+$/, "")` runs before confinement, so these arrive as `""` and are `root-entry-path`, not `absolute-entry-path`. `/.` keeps a non-slash tail and stays `absolute-entry-path`. Both refused, each pinned separately, and the difference is documented in the comment rather than left to be rediscovered.

**The reproduction, re-run against the new code.** Entry named `.`, typeflag `0`, empty Project. Pre-fix: `entryRelative` is `"."`, length 1, so the old `entryRelative.length === 0` skip does not fire; `confined.relative` is `""`; `target` is the root; `lstat` finds a directory; `refuseNonEmptyDirectory` reads an empty `readdir` and returns; `rm -r` takes the root; `open(…, "w")` leaves a file at its path; `200` and `{"type":"done"}`. Post-fix the guard fires before `const target` is even bound — `TarError("root-entry-path")`, caught by `handlePut`, emitted as `{"type":"error","code":"root-entry-path"}` with **no `done` line**, lease released in the `finally`. The root is not touched, not stat'd, not opened.

**Distinguishability parity.** The `PUT` branch answers `400 malformed-path` before the status line; the tar branch answers an NDJSON `error` line with a code of its own. The shapes differ because the `200` is spent before an entry is read, and that is the ceiling on what "distinguishable" can mean here — so the load-bearing half is the *absent* `done`, which is asserted. `root-entry-path` sits with `absolute-entry-path` and `dot-dot-entry-path` and is asserted not to be `directory-in-the-way`, which is the distinction that matters: being the root, versus what the root holds.

**The directory carve-out is right, and is an improvement.** `tar -cf - .` opens with exactly this header, so refusing it would refuse the ordinary archive — the interop test says so out loud. Skipped via `reader.skip(bodySize + bodyPadding)` rather than a bare `continue`, which keeps the stream in sync, and skipped rather than applied, so the archive's mode bits no longer land on the Project root. Pre-fix that `chmod` was real.

## 2. The mirrored tests

24 in `files-tar-unpack-hardening.test.ts`, 12 in `core-files-routes.test.ts`. The route-level set is the one I asked for: four spellings × empty and populated Project, each asserting the `error` line's code, the absence of `done`, `lstatSync(root).isDirectory()`, and `readTree(root)` unchanged.

Replayed against `100579a`: on an empty Project the pre-fix code emits no `error` line at all and does emit `done`, so both assertions fail and `readTree` fails against a root that is no longer a directory; on a populated Project it emits `directory-in-the-way`, so the code assertion fails; the nameless entry took the old silent skip, so `done` is present and the error assertion fails. They fail for the reason the fix exists, not incidentally — and keeping both roots is what pins the refusal to *being the root*, since the populated case was already accidentally covered.

One nuance on the count rather than the verdict: `skips a root directory entry that declares a body` is claimed among the pre-fix failures through an `EACCES` cascade after the root is chmod'd `0o644`. That holds unprivileged and not as root, and some container runners are root — so that single case's pre-fix failure is environment-dependent. It passes post-fix either way and nothing rests on it; call it 30 or 31 of 36.

## 3. The adversarial pass

Candidates built against `files-confinement.ts` rather than against a list of spellings:

- **Non-empty `relative` that still resolves to the root — impossible by construction.** `absolute = path.join(parent.absolute, base)` where `base` is a single split segment, never empty, never `.` (filtered), never `..` (refused in `stringRefusal`), and `parent.absolute` is already proven within the root. The join always lands strictly deeper. So there is no string that reaches the write with `relative !== ""` and `absolute === realRoot`, which is what makes the guard's key sound rather than merely convenient.
- **Last component a symlink at the root (`selflink -> .`).** Regular-file branch: `lstat` sees the link, `isFile()` is false, `refuseNonEmptyDirectory` no-ops on a non-directory, `fs.rm` unlinks the link **without following**, `open` creates a real file at the link's path. Directory branch: same, `rm` then `mkdir`. `writeSingleFile` does the same `rm`-then-`open`. The root survives all three; your read on this was correct.
- **`destRoot` aliasing — the sharpest case I could build.** `PUT ?path=selflink` with `Content-Type: application/x-tar`, where `selflink -> .`. The last component is left literal, so `unpackTarInto` receives a `destRoot` that *realpaths to the Project root itself*, and the `mkdir(absolute)` above it succeeds through the link. An entry named `.` in that archive still confines to `relative === ""` and is still refused. The guard holds when `destRoot` is an alias of the root, which a check written against `destRoot`'s identity would not have.
- **Hardlink naming the root** (`x`, linkname `.`): resolves to the root, passes containment, then `link()` on a directory fails `EPERM`. No root write and no alias created. **Symlink entry named `.`**: refused before the link-target check has an opinion, which is the right order of complaints.
- **`a/..`, `a/../`** → `dot-dot-entry-path`. **`s/x`** through a root-pointing symlink → lands at `root/x`, inside the root rather than on it.
- **`PUT` path re-swept:** `.`, `./`, `./.`, `" . "`, `path` omitted → `400 malformed-path` before the lease and before `directoryInTheWay`; `/`, `//` → `400 absolute-path`; `..` → `400 dot-dot-segment`. Every root-resolving spelling is refused before anything on disk is consulted.
- `handleContinue` funnels into the same `route()` and adds no write path. `unpackTarInto` and `writeSingleFile` remain the only two.

Nothing survives on either path.

One cosmetic, pre-existing and **not** something to fix here: in the regular-file branch, `else if (existing?.isSymbolicLink())` is unreachable — a symlink's `lstat` already fails `isFile()`, so the first branch always claims it. Dead, harmless, outside this diff, and not worth a commit of its own. Sweep it up whenever that branch is next touched.

## 4. CI

Green at `d41790a` — run `31736630146`, `head_sha` confirmed as `d41790abcccd86778958c069e877ad630bdbf8a2`, conclusion `success`. 19 checks pass, 3 skipped (the train-tag jobs, which do not run on a PR).

## 5. The §4.3 ticket

Not filed, and you said so rather than leaving me to notice — which is the part I care about. It was a non-blocking ask and it stays non-blocking. Your point that this commit had to add a 22nd member in two files is my argument, not against it; file it when that seam is next opened.

## Verdict

Approved. The root-replacement defect is dead on the single-file `PUT` path and on the tar path, refused on the resolved answer in both, before anything destructive, with a code a client can act on and no `done` line claiming otherwise — and the tests fail on the pre-fix code for the right reason at both levels.

**The merge is yours.** I have not merged, and will not.

mrfqcentic (approved):
**Approving.** The change request is fully addressed, and verified rather than taken on trust.

## The blocking issue — fixed as asked, and then some

`100579a` puts the guard exactly where it belongs: a non-tar `PUT` with an empty `relative` is refused `400 malformed-path` at the top of `handlePut` — before the 200, and before the write lease, so a malformed request never touches the lock table. The tests are the ones named in the review and more: all three spellings (no `path`, `?path=`, `?path=.`) against both an empty and a populated root, the root asserted surviving as a directory with every byte, the code distinguishable from `directory-in-the-way`, and both boundary writes (tar at the root, a named file at the top level) pinned as unaffected. `external-api.md` documents it.

Both non-blocking standards notes are fixed in the same commit: the four uses of the term the glossary bans, and the `content-length` docblock, which now states the true rule — copied from the Core's own answer, never invented.

## The second variant is the better catch

`d41790a` closes a path my review missed: a tar *entry* resolving to the root reproduced the same defect one branch over, and worse — the 200 is spent by the time an entry is read, so the stream would have reported `done` over a deleted root. Reading `confined.relative` rather than the raw name is the right shape: every spelling (`.`, `./`, nameless, slash-only, pax `path` and GNU long-name overrides) collapses to one answer, and the check cannot drift out of sync with confinement. Skipping — rather than applying — a directory entry that names the root is the correct call, and the chmod-the-root behaviour change is pinned by a test. The new `root-entry-path` code landed in both `CoreFilesErrorCode` unions, keeping them in step.

## Verification

- Full gate green on `d41790a` — all 19 required checks.
- Ran locally on Node 24: the two touched suites pass 128/128; the full Core suite passes, 1114 passed / 3 skipped across 58 files.

The two judgement calls from the first pass stand and are fine to carry: the duplicated error union (now in sync, and the contract test watches the seam) and the CLI's full-tree buffering against #166's letter.

Merge remains yours, as the PR body says.

mrfqcentic (changes requested):
## Manual test round on this branch — changes needed before we agree to 0.3.0

We ran a human QA pass against images built from this branch (panel + core containers, reference deployment, arm64), exercising the 0.2.1/0.2.2 elements plus the phase-3 Project-files surface, the CLI (0.3.0 from this branch) and the SDK.

### What passed

- Upgrade path: 0.2.0 volumes carried over cleanly (Operator, pairing, DB), both containers healthy, versions locked.
- Core registration, `project add / ls`, `session start / send / logs / ls` over the CLI.
- SDK: connect, `projectsList`, `projectsMutate create`, session spawn, and a raw `harnessInstall` frame installing cursor-cli on the Core, availability flipping to `available`.
- Full end-to-end on cursor-cli: start prompt lands, turn runs, status `running → finished`, `waitForIdle` resolves, transcript readable.
- `project cp` up/down and `project files` listings.

### Blocking for 0.3.0

1. **#225** — Panel drag-drop of a new (especially binary) file dies with the browser's `Failed to fetch` even though the write lands on the Core; retry refused as `transfer-in-progress`; the file view stops reading during it. The primary phase-3 gesture reports failure on success.
2. **#229** — opencode: the start prompt never reaches the harness (Core logs it delivered; TUI input stays empty). Every programmatic opencode Session looks stillborn.
3. **#230** — opencode Sessions never report status: `--wait` and the SDK's `waitForIdle` always time out, even after a successful turn.
4. **#231** — the top-bar Project dropdown is an empty listbox, labelled the literal "Project", and renders at the root path.

### Should fix, not blocking

- **#226 / #227 / #228** — file tree UX: collapsed-by-default folders, folder rows as drop targets, spring-loaded expansion during drag. With any real Project (a `node_modules`) the current always-expanded flat tree is unusable.

Happy to re-run the same manual pass once the blockers land.

qcentic-dev:
All seven QA findings from the manual test round are addressed and ready for re-review.

- [x] **#225** — `b18260f` — Both Panel hosts now share one Node↔fetch bridge: a body that fails half-written answers 502 instead of killing the process, and a browser hangup cancels the upstream transfer so the Project's write lease is released.
- [x] **#226** — `b507074` — The Project file view builds a tree client-side from the manifest the Core already streamed, with folders collapsed by default and opened one level at a time; no extra round trips.
- [x] **#227** — `839c1f2` — A folder row is now a drop target and the file is written beneath it, with `composeUploadPath` as the single trusted join; a nameless drop still gets refused rather than silently re-rooted.
- [x] **#228** — `7709183` — Resting a drag on a folder row springs it open after one second (Finder/VS Code gesture), so a nested folder is reachable without letting go of the drag.
- [x] **#229** — `6a596cd` — The start prompt waits for a declared composer marker and confirms its own echo before submitting, instead of trusting a quiet-screen gap that fell inside opencode's boot hole.
- [x] **#230** — `28decdf` — A tagged, dependency-free plugin is written at spawn to `.opencode/plugins/actana-control.js` and reports an opencode Session's lifecycle, so status moves and wait-for-idle resolves.
- [x] **#231** — `ff35f5d` — The top bar's Project control reads the owning Core's project snapshot and switches between that Core's Projects, wearing the current Project's name instead of the literal word "Project".

qcentic-adm (approved):
**Approving.** All seven findings from the manual QA round are addressed, and I checked each one against the line that enforces it rather than against the commit message that claims it. Scope: the ten commits `b18260f`…`614fb0b` (`b507074`, `839c1f2`, `7709183`, `ff35f5d`, `6a596cd`, `28decdf`, the three QA-branch merges and the typecheck fix on top). The phase-3 commits at and below `d41790a` were already reviewed and I did not re-open them.

## Issue by issue

### #225 — drop dies with `Failed to fetch` while the write lands — **met**

Both halves of the report are fixed in the one place both hosts now run, and both are pinned over real sockets rather than over a `Request` built in-process.

- **The half-written answer.** `packages/panel/src/server/node-http-bridge.ts:195-219` is the pump. A body that errors with nothing sent yet goes to `sendUpstreamFailure` — a `502` with a sentence, headers cleared first (`:253`), which is why the upstream `x-ndjson` cannot survive onto this module's JSON. A body that errors after the status line is spent gets `res.destroy()` (`:216`), matching the Core's own refusal to let a truncated body terminate cleanly. Nothing reaches Node as an unhandled `'error'` on a source stream, because there is no `pipe()` left: `bin/panel.mjs:131` and `vite-api-plugin.ts:31` both call `serveNodeRequest`, and the two hand-written copies are deleted.
- **The browser ending its own transfer.** `serveNodeRequest` (`:124-129`) builds an `AbortController`, fires it from `res`'s `close` and `req`'s `aborted` guarded on `!res.writableFinished`, and hands the signal to the handler; `core-files-proxy.ts:302-329` forwards it to the Core and answers a deliberate cancellation with a never-delivered `499` so the router does not count it as an internal error. `drained()` (`:232-249`) settles on `close` and `error`, not `drain` alone — the parked-forever case the dev writer had.
- **The tests assert a NEXT request is answered.** `node-http-bridge.test.ts:153-154` — after the broken exchange, a second exchange on the same server must return `alive`. Since vitest runs the server in-process, the old writer's unhandled `'error'` never reaches that line. That is the only assertion a dead process can fail, and it is the one that is made. `core-files-client-hangup.test.ts:248-258` goes further: a real Core behind a real Panel HTTP server, the drop's socket destroyed mid-body, `locks.current(PROJECT_ID)` asserted back to `null` off the Core's own lock table, and the retry answered `200` on its first attempt — the `409` from the report. Its second test pins the third symptom: a listing of the same Project succeeds while a write is running.

### #226 — tree renders fully expanded — **met**

`file-tree.ts:128-140` takes the set of **expanded** paths, so closed is the default by construction and a closed folder's children are never walked. `ProjectFilesPanel.tsx:135` seeds that set empty, and holds it **by path** (`:132-134`) so a re-read — including the one after every upload — does not collapse the view. `project-files-panel.test.tsx` asserts the rendered rows exactly: `["incoming"]` on load, one level per click, `incoming/deep`'s children still hidden, and state surviving a `Refresh`.

### #227 — drop can only target the Project root — **met**

`ProjectFilesPanel.tsx:280-315` makes a folder row a target, with `stopPropagation` on both `dragOver` and `drop` so the panel behind it cannot take the same drop as a second write to the root. It passes `node.path` — the listing's own relative spelling — to `sendFromDrop`, which forwards to `send`, which joins in exactly one place: `composeUploadPath` (`core-files.ts:277-286`). Both sides are relative and stay relative: every empty and `.` segment is dropped, so `("/incoming//deep/", "/hello.txt")` → `incoming/deep/hello.txt` and no leading slash can survive. A nameless drop returns `""` rather than the folder's path, deliberately, so the Core's `400 malformed-path` still fires instead of a `PUT` landing *at* the folder. All of that is asserted in `core-files.test.ts:213-278`, and the panel test checks the wire: `path=incoming%2Fhello.txt`, written once, with a file-row drop still falling through to the root.

### #228 — spring-loaded folders — **met, and the test is not vacuous**

`spring-loaded.ts:63-79` — arming is idempotent per path (`if (armed.current?.path === path) return`), which is the whole bug the module exists to avoid, and `leave` (`:81-89`) only cancels the row it is timing. The non-vacuous test is `spring-loaded.test.ts:36-48`: it fires `hover` every 50 ms for a full `SPRING_LOADED_MS` and requires the folder to have opened — an implementation that re-armed on each `dragover` produces zero opens and fails it. Backed at the component level by `project-files-panel.test.tsx`, which drives real `dragOver` events on the row in a 100 ms loop, asserts `incoming/deep` appears, then drags into that newly revealed row and drops into it (`path=incoming%2Fdeep%2Fbye.txt`). Cancel-on-drop and cancel-on-pass-over are both pinned.

### #229 — opencode start prompt swallowed — **met, and the other harnesses are untouched**

The gate is readiness, not a longer pause: `harness-prompt-delivery.ts:866-869` refuses to type while `composerOnScreen` is false, and `echoConfirmed()` (`:799-804`) holds the `\r` until the prompt is visible, re-typing via `retypePrompt()` instead of submitting into nothing. No timing constant moved — `DEFAULT_PROMPT_DELIVERY_PROFILE` is unchanged and `HARNESS_PROMPT_DELIVERY_PROFILES` is still empty. The blast radius is bounded by `HARNESS_READINESS` (`:443-449`), whose only key is `opencode`; `readinessFor` falls through to `NO_READINESS`, under which `composerOnScreen` returns `true` unconditionally and `echoConfirmed` returns `true` on its first line. `harness-prompt-delivery.test.ts` asserts that directly for `claude-code`, `codex`, `cursor-cli` and an invented harness, plus `Object.keys(HARNESS_READINESS)` being exactly `["opencode"]`. The opencode fixtures are a live PTY capture replayed at captured timings, and the suite asserts nothing is written across the 4.4 s hole where the prompt used to be lost. The 15 s backstop still overrides everything, so a reworded placeholder costs a late prompt, not a lost one.

### #230 — opencode Sessions never report status — **met, hooks-based**

Nothing here reads output quietness. `harness-hooks.ts:240` adds the `opencode` row to `HOOK_FAMILIES` with `reportsTurnStart: true`, which is what stands the Panel's terminal-input fallback down; `harness-hooks-opencode.ts:176-193` writes the plugin at spawn behind the `@actana-control-managed` marker, and `pty-manager.ts:558-562` sets `AC_HOOK_URL`/`AC_HOOK_TOKEN`/`AC_HOOK_TASK_ID` from `getHookEnv()` — this Core's own loopback receiver — only when a file actually landed. That the plugin posts *there* is proven rather than described: `harness-hooks-opencode-plugin.test.ts:134-141` writes the generated file to disk, imports it the way opencode does, drives the hooks, and asserts the exact URL `http://127.0.0.1:45111/api/hooks/opencode?taskId=task_live_1&hookEvent=Stop` with `Bearer hook-token-abc`. The status then moves through the shared `mapHookEventToStatus` (both receivers re-export the one module), and `opencode-hooks-api.test.ts:211-241` replays the captured live sequence end to end — `ready` → `running` → `finished`, two `Stop`s settling once and staying settled, which is exactly what `--wait` and `waitForIdle` were timing out on. Subagent sessions are excluded by `parentID` rather than by ordering.

### #231 — top-bar Project dropdown — **met on all three counts**

Empty listbox: `ProjectPicker.tsx:88-90` now reads `useCoreProjectRows(coreId)` — the owning Core's snapshot over the link the tab already holds — instead of the Panel's own `projects` table, which has no row for a Core-owned Project. Wrong label: the `label`/`"Project"` fallback is deleted and the trigger renders `current.name` (`:274-275`). Rendering outside a Project: `__root.tsx:352` drops the crumb entirely at the root path, and `ProjectPicker.tsx:263` returns `null` whenever no Project resolves. The switch keeps `?coreId=` on the navigation (`:157-162`), which the test pins. All hooks sit above the early return, so the rules of hooks hold.

## Verification I ran

- **`pnpm typecheck` at `614fb0b`: clean.** All five workspace projects with a typecheck script, exit 0. `614fb0b` is what makes that true — the two `string | null` → `string | undefined` call sites are type-only and neither changes a branch.
- **Touched suites, locally on Node 24:** the eight panel suites in this diff pass 62/62, the three core suites pass 85/85.
- **Full suites, locally:** core 1131/1139, panel 137/140 files. Every failure is a 5–7 s timeout in a suite this diff does not touch (`operator-auth`, `actana-setup`, `panel-link/*`, `core-files-abort`) and is an artefact of my sandbox — `packages/core`'s file-route and transfer code is byte-identical between `d41790a` and `614fb0b`.

## One delivery note, not a code change

**The `Unit Tests` check is red at `614fb0b`, and it hid the suites that matter.** The failure is `packages/cli/src/__tests__/events-tail-cursor.test.ts > resumes where the last run stopped` timing out at 60 s — #208, open, pre-existing, and named in the PR body as unrelated. But `pnpm -r` stopped there, so **`packages/panel` and `packages/core` never ran on CI at this head**: only `sdk`, `shared` and `cli` did. Everything else is green (Typecheck, Lint, Conventions, both images, both installer E2Es, the Panel service seam). I ran panel and core myself and reported the result above, so this is a gate hygiene item rather than a doubt about the code — please re-run `Unit Tests` so the merge happens on a genuinely green check.

Two smaller things, neither blocking and neither worth a commit of its own:

1. `ProjectPicker`'s "renders nothing outside a Project" test exercises an unknown project id, which its own comment is honest about. The root-path half of #231 is enforced in `__root.tsx:352` and is unambiguous on reading, but it has no test of its own; worth one whenever the shell's crumb assembly is next touched.
2. `retypePrompt()` clears `this.screen`, so a second write needs the composer marker to be repainted before it can go out. The `maxPromptWrites: 3` budget and the 15 s backstop both bound this to "late, never lost", which is the documented direction — noting it only so the next person reading D6a knows it was considered.

## Checklist

| | |
| --- | --- |
| #225 Panel drop `Failed to fetch`, held lease, blocked reads | **met** — bridge shared by both hosts; real-socket suites assert the next request is answered and the Core's lock table is clear |
| #226 folders collapsed by default, one level at a time | **met** — expanded-set by construction, state held by path across refresh |
| #227 drop onto a folder row writes under it | **met** — `composeUploadPath` joins two relative paths, never absolute; nameless drop stays nameless |
| #228 spring-loaded expansion | **met** — idempotent arming, and a test a re-arming implementation cannot pass |
| #229 opencode start prompt delivered | **met** — composer readiness + echo confirmation; `HARNESS_READINESS` has exactly one key, other harnesses byte-identical |
| #230 opencode Session status | **met** — plugin posts to the Core's receiver, proven by importing the generated file; captured live sequence settles the card |
| #231 top-bar Project switcher | **met** — Core's Projects inside a Project, nothing at the root, current name on the trigger |
| `pnpm typecheck` at `614fb0b` | **clean** |
| CI at `614fb0b` | red on #208's known CLI flake; panel + core suites never executed — please re-run |

**Verdict: approved.** The seven findings are fixed at the enforcing line, the tests fail for the reason each fix exists rather than incidentally, and the two changes that could have reached beyond their issue — the shared Node↔fetch bridge and the opencode readiness table — are both bounded in a way the suites pin. The merge is yours; I have not merged and will not.

mrfqcentic (approved):
Reviewed the unreviewed delta on this branch — d2b27ab (#221, the last round reviewed in its own PR) through 614fb0b: the seven QA fixes (#225–#231), the two root-resolution refusals (100579a, d41790a), and the merges. Earlier rounds #215–#221 were reviewed at their own PRs, and the human retest covers the seams between them.

Two review axes, run independently against the diff:

**Spec** — every one of the seven issues is delivered, verified against the code rather than the commit messages:
- #225: both failure modes are handled in node-http-bridge — a body that fails pre-headers answers 502 instead of killing the process, and a browser hangup cancels the upstream fetch, with reader.cancel propagating so the Project's write lease is actually released (pinned by the lock-table test). Concurrent reads during a write pinned too.
- #226/#227/#228: client-side tree from the already-streamed manifest (no extra round trips), composeUploadPath as the single trusted join with a nameless drop still refused by the Core, one-second idempotent-per-path spring-load. The Panel validates nothing — ADR 0030 holds.
- #229: composer marker + echo-confirm + retype, with screen reset before the probe so only post-write output counts, and the 15s backstop trading a possibly-unread CR for "late never lost". Per-harness, so the other harnesses' paths are byte-identical — consistent with deferring #232.
- #230: the plugin no-ops without env, failed plugin writes report reportsTurnStart: false so the Panel fallback stays armed, and a crashed opencode is still caught by the pre-existing synthetic-ended mapping on PTY exit.
- #231: reads the owning Core's snapshot, wears the current Project's name, absent at root.
No undeclared scope creep — the small unasked-for touches (root-entry skip semantics, glossary fixes, the stale repo-no-op plugin deletion) are each declared in their commit bodies.

**Standards** — no violations of the reviewer rules in CONTRIBUTING.md; ADR clauses (0026 D3a/D6a, 0030 consequences) appended in-PR per the append-never-renumber rule; the glossary's retired noun is actively swept. Typecheck, lint, conventions, secret scan, e2e and image smokes all green; core 213 / panel 64 unit tests re-run green locally on the repo's declared Node 24.

**The red Unit Tests check is #208, not this branch**: same file, same test name, same line, same 60s timeout as the three pre-existing occurrences on other branches, and this diff touches nothing in the events-tail path. Consistent with the PR body's declaration.

Two nits, patch-level, not blocking this train (fine to fold into the post-0.3.0 patch round alongside #232):
1. getSetCookieHeaders now exists twice — private in server/api-router.ts:88 and exported-but-unned-elsewhere in server/node-http-bridge.ts:263. Have the router import the bridge's copy before the two regexes drift.
2. bin/panel.mjs:131 awaits serveNodeRequest but ignores its false ("handler declined") return — that path would leave the request hanging. Latent today (the TanStack handler never returns null), but a 404 or a comment would close it.

Approving — good to merge when you're ready.

Issue #231, opened 2026-08-14
Top-bar Project dropdown is always empty — should name the current Project and switch between Projects


mrfqcentic:
## What happened

The Panel's top bar has a "Project" dropdown (a listbox trigger next to the breadcrumb). It is broken three ways:

1. Opening it shows an **empty** listbox — no entries at all, even with three Projects registered on the connected Core.
2. Its label is the literal word "Project" instead of the Project currently being viewed.
3. It renders even at the root path (outside any Project), where there is no current Project for it to represent.

## What I expected

A project switcher, like the workspace/branch switchers in comparable tools:

- Only present when inside a Project.
- Labelled with the current Project's name.
- Opening it lists the other Projects on the Core, and picking one navigates to it.

## Steps to reproduce

1. Panel from `feat/project-files` (v0.3.0) paired with a Core owning several Projects.
2. Look at the top bar at `/` — the "Project" dropdown is already there.
3. Enter a Project and open the dropdown — it expands (`aria-expanded="true"`) to an empty list, and the label still reads "Project".

## Additional context

Reported during a QA session against the phase-3 project-files build. The trigger renders as a normal button with a chevron, so the control exists — it's the population (current name + sibling Projects) and the root-path visibility that are missing.

qcentic-adm:
Delivered — closing.

**Carried by:** commit `ff35f5d` *"fix(panel): the top bar's Project control is a switcher, not a label"* (`Refs #231`), on `qa/project-switcher`, merged into `feat/project-files` at `0b39f63` and into `main` by **PR #223** (merged 2026-08-14T08:50Z).
**Released in:** **v0.3.0** (promotion PR #234, published 2026-08-14T09:44Z).

Filed at 07:03Z, fixed at 07:38Z the same morning.

**All three defects in the report are addressed, and I checked each separately.**

**1. Empty listbox.** The rows came off the Panel's own `projects` table, which holds no row for a Core-owned Project. They now come off the owning Core's snapshot, over the link the tab already holds:

```tsx
// packages/panel/src/components/views/ProjectPicker.tsx:89-90
const coreProjects = useCoreProjectRows(coreId);
const projects = coreId ? coreProjects.projects : panelProjects;
```

`useCoreProjectRows` is new in this commit (`packages/panel/src/lib/use-fleet.ts`, +66 lines) — `useCoreProjects` joined onto the Panel-side presentation row each Project is filed under (ADR-0022), so groups, pins and card images keep working.

[...]

PR #234, opened 2026-08-14
chore(release): promote 0.3.0

### LINKED ISSUE #225: Panel drag-drop: new binary file upload dies with the browser's "Failed to fetch" even though the write lands
## What happened

Dropping a **new** file onto a Project in the Panel's file view fails with the browser's `Failed to fetch` — but only for files that did not already exist, and especially binary ones (reproduced with a `.blend` file). Overwrites of existing text files (`README.md`) succeed and are correctly named `overwritten`.

The failure is cosmetic-but-cascading:

- The write actually **lands**: after a page reload the file is in the listing with the correct size.
- An immediate retry of the drop is refused with `transfer-in-progress` ("another write transfer is already running on this Project … one write at a time per Project; reads are unrestricted and concurrent") — so the first transfer was still holding the write lease while the Panel was already reporting it dead.
- While in that state the file view itself goes to "Could not read this Project's files — Failed to fetch", even though reads are supposed to be unrestricted and concurrent with a write.

`Failed to fetch` is the browser's TypeError for the same-origin Panel hop producing no response at all — so the browser↔Panel connection is dying mid-upload, while the Panel↔Core transfer carries on to completion behind it.

## What I expected

The progress stream for the drop to stay alive until its `done` (or a real refusal with a code), the item to be named `written`, and concurrent reads of the Project's files to keep working during the transfer.

## Steps to reproduce

1. Panel and Core from the `feat/project-files` branch (v0.3.0, core-link 0.15.0), reference container deployment, Panel on loopback HTTP.
2. Open a Project's file view, drag-drop a binary file whose name does not yet exist in the Project (a few-MB `.blend` reproduces it).
3. The item shows `Failed to fetch`; drop it again → `transfer-in-progress` refusal; the file listing also fails with `Failed to fetch`.
4. Reload the page → the file is present with the correct size.

## Additional context

- Overwriting an existing file succeeds and is named `overwritten`, so the write path as such works; the difference tracks new-vs-existing (and possibly content type or size), not permissions.
- Reported during a QA session against the phase-3 project-files build; consistent, not intermittent.

mrfqcentic:
The train's promotion pull request — its checks and approval are what `promote.yml` consumes (ADR 0023 D5, D16).

Carries: the phase-3 project-files surface (#223 with all seven QA fixes #225–#231), and the `id-token: write` promote.yml hotfix (#233).


qcentic-adm (approved):
Verified the train carries only the two expected changes: the clean merge of the already-approved #223 (no evil-merge content — 614fb0b0...4f48ea67 is empty) and #233's single `id-token: write` line in promote.yml's `release:` workflow_call permissions block; the rest of main...beta/0.3.0 is the six-manifest 0.3.0 version cut, and main has not moved since the cut (24 ahead, 0 behind).

mrfqcentic:
Unit Tests green on attempt 2 (the #208 flake struck attempt 1 — logged there). Checks and second-account approval in place; dispatching the promotion, which performs the fast-forward itself — this PR is consumed by promote.yml, not merged by the button.

Issue #243, opened 2026-08-17
Core: nothing settles a Core-owned Session whose process died — a lost Stop hook arms no backstop, and there is no boot sweep


mrfqcentic:
### Prerequisites

- [x] I have searched existing issues and this is not a duplicate — closest is the second half of #187; see **Prior art**
- [x] I can reproduce this on the latest release / main branch — a wedged row is present in a live `0.3.0` Core's database right now, and the missing code path is absent from `main` at `da2af3e`
- [x] This is not a security vulnerability

### Description

A Core-owned Session can be left permanently marked `running` after its work is over. There are three independent ways in, and no backstop catches the important one.

**1. Hook delivery has no guarantee, and no observability.** Every status a Session card shows arrives as a fire-and-forget POST:

```
sh -c 'curl -sS -m 3 -X POST ... "$AC_HOOK_URL/api/hooks/claude?..." || true'
```

No retry, no ack, no sequence number. A 3-second timeout against a Core whose event loop is also serving PTY fan-out, SQLite writes and file transfers is reachable under load, and `|| true` swallows timeout, connection-refused, 401 and 500 identically and silently. Nothing on either side records that a hook was attempted, so a lost hook leaves no trace anywhere.

**2. The drain backstop is armed by a hook that arrived.** `armDeferredFinish` is only reached from the subagent-lifecycle branch and from the `status === "finished" && hasActiveSubagents(taskId)` branch of `handleHarnessHookEvent` — both of which require a hook to have landed. **If the terminal `Stop` is the POST that drops, nothing is armed at all**, and the row simply stays `running` with no timer watching it. If a `SubagentStop` drops instead, the tracked entry clears only by expiry: `ACTIVE_SUBAGENT_TTL_MS = 2 * 60 * 60 * 1000` (`subagent-activity.ts:28`), plus `DRAIN_FINISH_GRACE_MS` — so a finished Session can read `running` for just over two hours.

**3. A Core restart strands every active row, and nothing sweeps at boot.** `onSessionExit` (`core-entry.ts:212`) is the only unconditional settle, and it can only fire while the Core process is alive. Kill the daemon and its PTYs die with it: no exit callback, no `pty:exit` event, no settle. Subagent bookkeeping and the deferred-finish timers are per-process in-memory `Map`s (`subagent-activity.ts`), so they are lost too — and there is no startup reconciliation anywhere in `packages/core`.

### Evidence from a live Core

```
[...]

Issue #242, opened 2026-08-17
Panel: a reloaded tab joins the drive queue behind its own ghost, so a Session it alone was driving renders read-only for up to 45s


mrfqcentic:
### Prerequisites

- [x] I have searched existing issues and this is not a duplicate — see **Prior art** below for why this is not #186 or #187
- [x] I can reproduce this on the latest release / main branch — the code path is on `main` at `da2af3e`; the symptom is observed on a live remote Core, the mechanism traced in source
- [x] This is not a security vulnerability

### Description

Reloading the browser tab that started a Session can leave that Session **read-only in the very tab that was driving it**, for up to 45 seconds — and reliably so when the Panel and the browser are separated by a reverse proxy or Docker's NAT.

This is **not** the core-link Session lock. On the Core I investigated, `event_log` contains **zero** `session:lockChanged` rows across its entire history, so no core-link lock has ever been claimed there — nothing at that layer has ever made anything read-only. The arbitration that actually fires is the intra-Panel one: `SessionDriveRegister`, which decides which of the Panel's own tabs holds the keyboard (ADR 0024 D3, #147).

The register keys interest by **object identity of the `PanelLinkSession`** — the browser's socket. A reload destroys that object and creates a new one, and nothing carries across. So:

1. You reload. The browser opens a new panel-link socket — a new holder, as far as the register is concerned.
2. The new page announces its panes with `drive want: "watch"`.
3. The previous socket is frequently **still registered**: its `close` has not been processed yet. Through a proxy or Docker's NAT this lag is routine — `ws-server.ts:118` comments on precisely those idle timeouts.
4. First-come applies. `want()` appends the returning tab *behind* its own ghost (`session-drive-register.ts:65-73`), so it is not the driver.
5. The router answers `{ driving: false }` (`router.ts:368-370`) and the pane renders read-only.

Recovery comes only when the ghost is reaped by the panel-link heartbeat — `HEARTBEAT_TIMEOUT_MS = 45_000` (`ws-server.ts:123`) — at which point `releaseDrives` hands the drive to the next watcher, which is the returning tab.

The behaviour is already described in the router, framed as the two-tab case:

> A tab that comes *back* — a reload, a dropped network — re-announces its panes and joins the back of the queue, so on a Panel with two tabs open on one Session a flap can move the keyboard to the other one. That is the first-come rule applied evenly rather than a special case.
> — `router.ts:377-382`

[...]

PR #245, opened 2026-08-18
fix(core-link): advance the live-event cursor only after a send actually lands

### LINKED ISSUE #244: core-link: pushLiveEvents advances the per-connection cursor on a send that never landed, so a missed event is never replayed
### Prerequisites

- [x] I have searched existing issues and this is not a duplicate — #139 may be a downstream symptom; see **Prior art**
- [ ] I can reproduce this on the latest release / main branch — **not reproduced interactively.** Confirmed by code inspection at `da2af3e`; the loss is silent by construction, which is part of the defect. Reproduction sketch under *Steps to reproduce*.
- [x] This is not a security vulnerability

### Description

`pushLiveEvents` advances a connection's event cursor whether or not the frame it just pushed actually reached the socket. Because the client's own cursor is driven by the events it *receives*, a single failed send becomes a **permanent** hole that no reconnect replay can close.

```ts
// packages/core/src/pty-core-link-server.ts:912-925
private pushLiveEvents(conn: ActiveConnection): void {
  if (!conn.subscribed || !this.eventLog) return;
  if (this.authVerifier && !conn.authenticated) return;
  const after = conn.lastSentEventId;
  const tail = this.eventLog.readEventTail(after, EVENT_TAIL_LIMIT);
  for (const event of tail) {
    this.send(conn.ws, { type: "event", event });
    conn.lastSentEventId = event.eventId;   // ← unconditional
  }
}

// packages/core/src/pty-core-link-server.ts:1670-1676
private send(ws: WebSocketLike, frame: CoreLinkServerFrame): void {
  try {
    ws.send(serializeCoreLinkFrame(frame));
  } catch {
    /* connection closed mid-send — the close handler will clean up */
  }
}
```

The `try/catch` does not cover the case that matters. Node's `ws` does **not** throw when the socket is `CLOSING`/`CLOSED`: with no callback supplied it emits an `error` event on the socket, which this server handles by logging `core-link.connection.error` and continuing. So the `catch` never runs, and the cursor advances past an event the client never got.

Then the hole becomes permanent:

1. Event *N* fails to land. `conn.lastSentEventId` is now *N*; the client's cursor is still *N−1*.
2. Event *N+1* lands normally. The client applies it and persists its cursor as *N+1*.
3. On the next `subscribe`, the client sends `lastEventId: N+1`. `readEventTail` returns everything *after* *N+1*.
4. Event *N* is never replayed, by any path.

`WebSocketLike` declares `readyState` (`pty-core-link-server.ts:400`) and **nothing in the server ever reads it** — the only other occurrence is the adapter's getter at :1971. `bufferedAmount` is never consulted either, so there is no backpressure signal at all.

The frames that ride this path are the ones whose absence is user-visible: `task:updated`, `session:finished`, `session:lockChanged`, `task:created`, `pty:spawn`. PTY output survives the same treatment only because it carries `seq` and has a replay window (`replay` / `drainPtyHold`); the event log has neither per-frame.

**Why this is worse on a remote Core.** A TLS link across a real network path spends more time in transitional socket states than a loopback pipe does, and the 500 ms poll (`DEFAULT_LIVE_EVENT_POLL_MS`) gives a comfortable window for a socket to become unwritable between the tail read and the send. This matches the reported pattern of notifications arriving "not every time, especially on a remote environment".

### Steps to reproduce

Not reproduced interactively — by design there is no signal to observe. A unit-level reproduction against `PtyCoreLinkServer` is straightforward, since the transport is injectable:

1. Stand up the server with an injected `WebSocketLike` whose `send` reports failure the way `ws` does — emit `error`, do not throw — for one call only.
2. Subscribe the connection, then append three events to the log.
3. Let two poll ticks run: the failing send takes event 2, events 1 and 3 land.
4. Reconnect and `subscribe` with the client's cursor (`3`).
5. Assert on the delivered stream. Event 2 never appears.

This is also the regression test the fix wants.

### Expected behavior

`lastSentEventId` advances only for a frame that actually reached a writable socket. An event that could not be delivered is re-attempted on a later tick, or left for the reconnect replay to carry — either way it is delivered exactly once and never skipped.

### Actual behavior

The cursor advances unconditionally. An undeliverable event is silently and permanently dropped for that connection, with no error surfaced to the client, no counter, and no log line naming the lost event.

### Fix direction

- Guard the send on `ws.readyState === OPEN` — the field is already on the interface and already implemented by the adapter, so this costs nothing.
- Use the `ws.send(data, cb)` callback (or a `sendResult` on `WebSocketLike`) to learn whether the frame was accepted, and only then advance the cursor.
- On failure, leave `conn.lastSentEventId` where it was and return from the loop, so the next tick retries from the same point rather than skipping forward.
- Consider a `log.warn` naming the connection and the `eventId` that could not be delivered. The absence of any such line is why this class of bug is invisible today.

`fanOutPtyEvent` (:807-829) uses the same `send` and deserves the same guard, though PTY loss is recoverable via `seq`/replay.

### Prior art

**#139** (`A Panel left open overnight comes back detached — which link drops is unestablished`) is a plausible downstream symptom: a long-lived link that misses lifecycle events would present exactly as a Panel out of sync with its Cores. Unproven, but worth checking together.

**#208** (`Flaky: events-tail-cursor.test.ts times out in CI under runner load`) touches the same cursor path. Different defect, but the fix here will want that suite green.

### Which part?

Core link — pairing, or a Core showing as unreachable

### Affected version(s)

Source `da2af3e` (`main`); present in Core `0.3.0` / protocol `0.15.0`.

```shell
Core:    actana-core 0.3.0, linux-arm64, node 24.15.0, Ubuntu 24.04.4 LTS (container)
         AC_CORE_REMOTE=1, core link on 0.0.0.0:8443 (TLS + bearer)
Panel:   deployed remotely; version not captured
Source:  actana/control @ da2af3e
```

### Severity (your assessment)

S2 – Major: core feature broken, workaround exists

No workaround exists for the loss itself — a dropped event is unrecoverable — but impact is bounded per event, and a later `tasksList` re-hydration corrects the view for anything that is also carried in a snapshot. What it cannot correct is a one-shot event such as `session:finished`, which is a notification the operator simply never receives.

### Relevant logs

```shell
# There are none, and that is the finding. The failure path logs only:
#   core-link.connection.error   { error: ... }
# which names no eventId and is indistinguishable from a benign socket error.
# No counter, no gap detection, nothing correlating a lost frame to a cursor.
```

### Contribution

- [ ] I am willing to submit a PR to fix this bug

qcentic-dev:
## Summary

`pushLiveEvents` advanced a connection's event cursor whether or not the frame it
pushed reached the socket, so a single failed send became a permanent hole no
reconnect replay could close. This makes the cursor advance **behind** the send:
a frame the socket refuses leaves `lastSentEventId` where it was and the next
poll tick retries from the same event.

The failure was silent by construction — Node's `ws` does not throw on a
`CLOSING`/`CLOSED` socket, it emits an `error` event the server logged as a
benign `core-link.connection.error` — so a refusal now also logs
`core-link.event.undelivered` naming the event id.

Base is `fix/session-and-event-reliability` (this ticket's train branch), not
`main` and not `beta/0.3.1`.

## Related issues

Closes #244

Adjacent, deliberately untouched:

- #208 — the `events-tail-cursor` CI flake sits on these same semantics and comes
  after this ticket in the train. The suite is green here; its harness is left
  exactly as it was for that ticket to change.
- #243 (session settle / hook ack / boot reconciliation) and #242
  (`session-drive-register.ts`, panel `router.ts`) are in flight in parallel and
  own those files. Nothing here touches either.

## Type of change

- [ ] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## What changed

`packages/core/src/pty-core-link-server.ts`

- `sendResult(ws, frame, onDeliveryError?)` replaces the bare `send`, and reports
  whether the socket took the frame on two signals: a `readyState` of `CLOSING`
  or `CLOSED` is refused *before* the write (the remote-Core case — the 500 ms
  poll leaves ample room for a link to go unwritable between the tail read and
  the send), and a socket that fails *after* an accepted call reports through
  `ws`'s delivery callback, which `adaptWs` now forwards. The callback is only
  attached when a caller asks for it, so every other frame keeps emitting the
  socket `error` that puts `core-link.connection.error` in the log.
- `pushLiveEvents` stops at the first refusal with the cursor still below it; a
  late callback failure rewinds the cursor to the same place. Delivery is
  at-least-once and in order rather than at-most-once — a duplicate is a fact the
  Panel already holds, a skip is one it cannot ask for.
- `handleSubscribe` moves its replay cursor on the same evidence and reports the
  event it actually reached in `eventsReplayed`, instead of the end of a tail it
  only partly wrote.
- `fanOutPtyEvent` picks up the `readyState` guard through the same `send` but
  stays ungated on the result: PTY loss is recoverable via `seq` + replay, which
  is the difference the ticket calls out.

## How was this tested?

New: `packages/core/src/__tests__/core-link-event-cursor-advance.test.ts`, four
tests, each of which **fails on the old ordering**:

| Test | On `da2af3e` |
| --- | --- |
| replays an event whose send failed, instead of skipping it forever — the ticket's own repro: 3 events, the middle one refused, reconnect off the client's real cursor | delivered set is `[1, 3]`; event 2 is gone for good |
| keeps the cursor put while the socket is unwritable, and delivers on recovery | the events are consumed by the refusal and never arrive |
| stops the subscribe replay at the last event the socket took | `eventsReplayed` reports a tail that was only partly written |
| names the undeliverable event in a log line | only `core-link.connection.error`, which names no event |

The fake socket models both real failure modes: `readyState` leaving `OPEN` with
no close handshake, and a send that is accepted and then reported failed a tick
later — neither of which throws.

Suites run locally (`env -u AC_*`):

- `@actana/core` — 60 files, 1143 tests, green
- `@actana/sdk` — 19 files, 236 tests, green (it drives `PtyCoreLinkServer`
  through `fake-core-link`)
- `@actana/cli` — 30 files, 421 tests, green, `events-tail-cursor.test.ts`
  included and unmodified
- `pnpm typecheck` across the workspace, and `pnpm lint` (0 errors)

## Checklist

- [x] This PR targets the open train branch for this work (`fix/session-and-event-reliability`), not `main`
- [ ] I updated documentation (README, docs/, changelog entry) where relevant — no docs surface changes; `CHANGELOG.md` has not been touched by a `0.3.x` commit and is written per release, not per PR
- [x] Dependent changes have been merged (or this PR is marked as draft/stacked) — left as a **draft** on the train branch

### DIFF
--- a/packages/core/src/__tests__/core-link-event-cursor-advance.test.ts
+++ b/packages/core/src/__tests__/core-link-event-cursor-advance.test.ts
@@ -1,0 +1,299 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import log from "../log";
+import {
+  PtyCoreLinkServer,
+  type EventLogPort,
+  type WebSocketLike,
+  type WebSocketServerLike,
+} from "../pty-core-link-server";
+import type { PtyCore, PtyCoreEvent } from "../pty-manager";
+import type { CoreLinkEvent } from "@actana/sdk/core-link-frames";
+
+// The live-event cursor advances only behind a send that landed (issue 244).
+//
+// The bug these tests pin is an ordering: `pushLiveEvents` moved
+// `conn.lastSentEventId` past every event it *attempted*, and a send that fails
+// on this transport is silent — Node's `ws` does not throw on a `CLOSING`
+// socket, it emits an `error` event the server logs as a benign socket error.
+// Because the client's own cursor is driven by the events it *receives*, one
+// skipped frame plus one delivered frame behind it is a hole no reconnect
+// replay can close: the client asks for "everything after the later one" and
+// the server obliges.
+//
+// So the assertions here are about what the *client* ends up holding across a
+// failure and a reconnect, not about a server field. Each one fails on the old
+// ordering.
+
+type Listener = (...args: unknown[]) => void;
+
+/**
+ * A socket that fails the way a real one does: it never throws.
+ *
+ * Two failure modes, because the server has to survive both. `readyState`
+ * leaving `OPEN` is the reported remote-Core case — a link that went unwritable
+ * between the poll's tail read and its send, with the close handshake not yet
+ * run. `failEventOnce` is `ws`'s other one: the socket takes the frame,
+ * synchronously reports nothing, and reports the failure a tick later through
+ * the delivery callback (or, with no callback supplied, as an `error` event
+ * nobody can attribute to a frame).
+ */
+class FakeWebSocket {
+  readyState = 1;
+  sent: string[] = [];
+  closed = false;
+  /** eventId whose first send attempt fails asynchronously; null once spent. */
+  failEventOnce: number | null = null;
+  /** eventId whose first send attempt throws — the one failure `ws` reports synchronously. */
+  throwEventOnce: number | null = null;
+  /** Every frame the transport refused, in order — the loss the client cannot see. */
+  dropped: string[] = [];
+  private listeners: Record<string, Listener[]> = {};
+
+  send(data: string, cb?: (err?: Error) => void): void {
+    if (this.throwEventOnce !== null && this.eventIdOf(data) === this.throwEventOnce) {
+      this.throwEventOnce = null;
+      this.dropped.push(data);
+      throw new Error("send failed");
+    }
+    const err = this.refusalFor(data);
+    if (err) {
+      this.dropped.push(data);
+      // Exactly what `ws` does: a callback gets the error on a later tick, and
+      // with no callback the socket emits `error` and the frame is gone.
+      if (cb) queueMicrotask(() => cb(err));
+      else queueMicrotask(() => this.emit("error", err));
+      return;
+    }
+    this.sent.push(data);
+    if (cb) queueMicrotask(() => cb());
+  }
+
+  /** Why this frame will not be delivered, or null if it will be. */
+  private refusalFor(data: string): Error | null {
+    if (this.readyState !== 1) return new Error("WebSocket is not open");
+    if (this.failEventOnce === null || this.eventIdOf(data) !== this.failEventOnce) return null;
+    this.failEventOnce = null;
+    return new Error("WebSocket is not open: readyState 2 (CLOSING)");
+  }
+
+  /** The eventId this frame carries, or null for anything that is not an `event`. */
+  private eventIdOf(data: string): number | null {
+    const frame = JSON.parse(data) as { type?: string; event?: { eventId?: number } };
+    return frame.type === "event" ? (frame.event?.eventId ?? null) : null;
+  }
+
+  close(): void {
+    if (this.closed) return;
+    this.closed = true;
+    this.readyState = 3;
+    this.emit("close");
+  }
+  /** The socket drops without a close handshake — `readyState` moves, no `close` fires. */
+  goUnwritable(): void {
+    this.readyState = 2;
+  }
+  becomeWritable(): void {
+    this.readyState = 1;
+  }
+  ping(): void {}
+  on(event: string, cb: Listener): void {
+    (this.listeners[event] ??= []).push(cb);
+  }
+  removeAllListeners(): void {
+    this.listeners = {};
+  }
+  emit(event: string, ...args: unknown[]): void {
+    for (const cb of this.listeners[event] ?? []) cb(...args);
+  }
+  receive(frame: unknown): void {
+    this.emit("message", JSON.stringify(frame));
+  }
+  ofType<T extends Record<string, unknown>>(type: string): T[] {
+    return this.sent
+      .map((raw) => JSON.parse(raw) as Record<string, unknown>)
+      .filter((frame) => frame.type === type) as T[];
+  }
+  /** The eventIds this client actually received, in arrival order. */
+  eventIds(): number[] {
+    return this.ofType<{ event: CoreLinkEvent }>("event").map((frame) => frame.event.eventId);
+  }
+}
+
+class FakeWebSocketServer {
+  private connCb: ((ws: WebSocketLike) => void) | null = null;
+  connect(ws: FakeWebSocket): void {
+    this.connCb?.(ws as unknown as WebSocketLike);
+  }
+  close(): void {}
+  on(event: string, cb: Listener): void {
+    if (event === "connection") this.connCb = cb as (ws: WebSocketLike) => void;
+  }
+}
+
+/** An in-memory event log — the same seam `event-log-store.ts` fills for real. */
+function fakeEventLog() {
+  const events: CoreLinkEvent[] = [];
+  const port: EventLogPort = {
+    appendEvent: (kind, payload, opts) => {
+      const eventId = events.length + 1;
+      events.push({
+        eventId,
+        ts: eventId,
+        kind,
+        payload,
+        ptyId: opts?.ptyId ?? null,
+        taskId: opts?.taskId ?? null,
+      });
+      return eventId;
+    },
+    readEventTail: (afterEventId, limit = 1_000) =>
+      events.filter((event) => event.eventId > afterEventId).slice(0, limit),
+    getLastEventId: () => events.length,
+  };
+  return { port, events };
+}
+
+function mockCore(): PtyCore {
+  return {
+    setEmitTarget: (_fn: ((event: PtyCoreEvent) => void) | null) => {},
+    spawn: async () => ({ ptyId: "pty-1", hooksReportTurnStart: true }),
+    write: () => true,
+    resize: () => true,
+    kill: () => true,
+    killLaunchProcesses: async () => ({ ptyCount: 0, ports: [] }),
+    findByTask: () => ({ ptyId: null }),
+    replay: () => ({ data: "", nextSeq: 0, from: 0 }),
+    killAll: () => {},
+  } as unknown as PtyCore;
+}
+
+/**
+ * How a client's own cursor moves: it persists the highest eventId it applied,
+ * and that is the number it presents on the next `subscribe`. This is the whole
+ * mechanism by which a server-side skip becomes permanent.
+ */
+function clientCursor(...sockets: FakeWebSocket[]): number {
+  return sockets.flatMap((ws) => ws.eventIds()).reduce((max, id) => Math.max(max, id), 0);
+}
+
+describe("the live-event cursor advances only behind a send that landed (issue 244)", () => {
+  let wss: FakeWebSocketServer;
+  let server: PtyCoreLinkServer;
+  let logi: ReturnType<typeof fakeEventLog>;
+
+  function connect(): FakeWebSocket {
+    const ws = new FakeWebSocket();
+    wss.connect(ws);
+    return ws;
+  }
+
+  beforeEach(() => {
+    wss = new FakeWebSocketServer();
+    logi = fakeEventLog();
+    server = new PtyCoreLinkServer(mockCore(), {
+      port: 0,
+      createServer: () => wss as unknown as WebSocketServerLike,
+      eventLog: logi.port,
+      liveEventPollMs: 5,
+    });
+  });
+
+  afterEach(() => {
+    server.close();
+    vi.restoreAllMocks();
+  });
+
+  it("replays an event whose send failed, instead of skipping it forever", async () => {
+    const first = connect();
+    first.receive({ type: "subscribe", reqId: "s1", lastEventId: 0 });
+
+    // The event in the middle is the one the socket will refuse — the shape
+    // from the ticket: 1 and 3 land, 2 does not.
+    first.failEventOnce = 2;
+    logi.port.appendEvent("task:created", JSON.stringify({ taskId: "t1" }), { taskId: "t1" });
+    logi.port.appendEvent("session:finished", JSON.stringify({ taskId: "t1" }), { taskId: "t1" });
+    logi.port.appendEvent("task:updated", JSON.stringify({ taskId: "t1" }), { taskId: "t1" });
+
+    await vi.waitFor(() => expect(first.eventIds()).toContain(3));
+
+    // The client reconnects off its own cursor, exactly as it would after any
+    // drop, and the Core owes it whatever it has not seen.
+    const cursor = clientCursor(first);
+    first.close();
+    const second = connect();
+    second.receive({ type: "subscribe", reqId: "s2", lastEventId: cursor });
+    await vi.waitFor(() => expect(second.ofType("eventsReplayed")).toHaveLength(1));
+
+    // The property: every appended event reached the client by some path. On
+    // the old ordering the cursor moved past 2 on the failed send, the client's
+    // own cursor moved to 3 behind it, and `session:finished` — a one-shot
+    // notification no snapshot re-hydration carries — was gone for good.
+    const delivered = new Set([...first.eventIds(), ...second.eventIds()]);
+    expect([...delivered].sort((a, b) => a - b)).toEqual([1, 2, 3]);
+  });
+
+  it("keeps the cursor put while the socket is unwritable, and delivers on recovery", async () => {
+    const warn = vi.spyOn(log, "warn").mockImplementation(() => {});
+    const ws = connect();
+    ws.receive({ type: "subscribe", reqId: "s1", lastEventId: 0 });
+    logi.port.appendEvent("task:created", "{}", { taskId: "t1" });
+    await vi.waitFor(() => expect(ws.eventIds()).toEqual([1]));
+
+    // The link goes unwritable with no close handshake — the remote-Core case.
+    // `ws.send` does not throw here; it never has.
+    ws.goUnwritable();
+    logi.port.appendEvent("session:lockChanged", "{}", { taskId: "t1" });
+    logi.port.appendEvent("session:finished", "{}", { taskId: "t1" });
+    // Nothing is even attempted at an unwritable socket — the poll notices
+    // before it writes, and says which event it is holding.
+    await vi.waitFor(() => expect(warn).toHaveBeenCalledWith("core-link.event.undelivered", expect.objectContaining({ eventId: 2 })));
+    expect(ws.sent.filter((raw) => raw.includes("\"event\""))).toHaveLength(1);
+    expect(ws.eventIds()).toEqual([1]);
+
+    // Nothing was consumed by the refusal: the cursor is still below event 2,
+    // so the very next tick that finds a writable socket carries both.
+    ws.becomeWritable();
+    await vi.waitFor(() => expect(ws.eventIds()).toEqual([1, 2, 3]));
+  });
+
+  it("stops the subscribe replay at the last event the socket took", async () => {
+    logi.port.appendEvent("task:created", "{}", { taskId: "t1" });
+    logi.port.appendEvent("session:finished", "{}", { taskId: "t1" });
+    logi.port.appendEvent("task:updated", "{}", { taskId: "t1" });
+
+    const ws = connect();
+    // A send that throws is the one failure the old `try/catch` did see — and
+    // it still advanced the cursor past the frame it had just caught.
+    ws.throwEventOnce = 2;
+    ws.receive({ type: "subscribe", reqId: "s1", lastEventId: 0 });
+
+    // The replay is a cursor advance like any other: it stops where delivery
+    // stopped and reports that, rather than a tail it only partly wrote. A
+    // client told `lastEventId: 3` here would never ask for event 2 again.
+    expect(ws.eventIds()).toEqual([1]);
+    expect(ws.ofType<{ lastEventId: number }>("eventsReplayed")[0]?.lastEventId).toBe(1);
+
+    // And the live loop picks up from there — the refused event included.
+    await vi.waitFor(() => expect(ws.eventIds()).toEqual([1, 2, 3]));
+  });
+
+  it("names the undeliverable event in a log line", async () => {
+    const warn = vi.spyOn(log, "warn").mockImplementation(() => {});
+    const ws = connect();
+    ws.receive({ type: "subscribe", reqId: "s1", lastEventId: 0 });
+    ws.goUnwritable();
+    logi.port.appendEvent("session:finished", "{}", { taskId: "t1" });
+
+    // The absence of any line naming a lost eventId is why this class of bug
+    // was invisible: `core-link.connection.error` names a socket, never a frame.
+    await vi.waitFor(() =>
+      expect(
+        warn.mock.calls.some(
+          ([tag, payload]) =>
+            tag === "core-link.event.undelivered" &&
+            (payload as { eventId?: number } | undefined)?.eventId === 1,
+        ),
+      ).toBe(true),
+    );
+  });
+});
--- a/packages/core/src/pty-core-link-server.ts
+++ b/packages/core/src/pty-core-link-server.ts
@@ -398,7 +398,16 @@ export interface WebSocketServerLike {
 
 export interface WebSocketLike {
   readonly readyState: number;
-  send(data: string): void;
+  /**
+   * Send one frame. The optional callback is Node `ws`'s delivery callback: it
+   * is invoked with an `Error` when the frame could not be handed to the
+   * socket, and with nothing when it was. Supplying it is what turns a failed
+   * send from an `error` event on the socket — which this server logs and
+   * shrugs at — into a fact the caller can act on. Transports that cannot
+   * report delivery simply never call it; see {@link PtyCoreLinkServer.sendResult}
+   * for what is inferred without it.
+   */
+  send(data: string, cb?: (err?: Error) => void): void;
   close(): void;
   on(event: "message", cb: (data: unknown) => void): void;
   on(event: "close", cb: () => void): void;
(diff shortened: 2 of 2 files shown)


qcentic-dev:
## Acceptance criteria — issue #244

Issue #244 has no `## Acceptance criteria` heading, so these are taken verbatim
from its **Prerequisites**, **Expected behavior**, **Fix direction** and **Steps
to reproduce** sections. Every one is listed, checked or explicitly not.

### Prerequisites

- [x] **"I can reproduce this on the latest release / main branch"** — filed
  unchecked ("not reproduced interactively… the loss is silent by
  construction"). Now reproduced at unit level, as the issue's own sketch
  proposed: on `da2af3e` the delivered set is `[1, 3]` and event 2 is gone for
  good. `packages/core/src/__tests__/core-link-event-cursor-advance.test.ts`.
- [x] *Searched existing issues / not a security vulnerability* — the reporter's,
  nothing for this PR to do.

### Expected behavior

- [x] **`lastSentEventId` advances only for a frame that actually reached a
  writable socket.** `pushLiveEvents` advances behind `deliverEvent`, never in
  front of it.
- [x] **An event that could not be delivered is re-attempted on a later tick, or
  left for the reconnect replay to carry.** The loop returns with the cursor
  still below the refused event, so the next tick retries from the same point;
  while the socket stays closing/closed every retry is refused before the write
  and the backlog goes out on the reconnect replay off the client's own cursor.
- [x] **…never skipped.** This is the half the ticket is about, and it holds:
  no path advances the cursor past a frame the socket did not take.
- [ ] **"…delivered exactly once"** — **not met, deliberately.** Delivery is
  now *at-least-once* and in order. When a send is accepted synchronously and
  `ws` reports the failure a tick later, the cursor rewinds below that event, and
  anything sent after it in the same tick is re-delivered with it. The trade is
  in the code comment: a duplicate is a Panel applying a fact it already holds,
  a skip is a fact it can never ask for. Exactly-once would need a per-frame ack
  from the client — a protocol change, and a larger ticket than this one. Raising
  it here rather than quietly claiming the stronger property.

### Fix direction

- [x] **Guard the send on `ws.readyState`** — implemented as *refuse `CLOSING`
  and `CLOSED`* rather than *require `OPEN`*, which is a deliberate deviation
  from the issue's wording. Those two are the states a write is known to be lost
  in; a server-side `ws` is already `OPEN` when the `connection` handler receives
  it, so the only direction it travels is towards closed. Requiring `OPEN` would
  also refuse `CONNECTING`, and the SDK's in-process rig legitimately writes
  `ready` at accept time before the handshake tick — demanding `OPEN` hung six
  `core-files-list-contract` tests. Reasoning is in the constant's doc comment.
- [x] **Learn whether the frame was accepted, and only then advance** — both
  forms the issue offered: a `sendResult` on the server returning
  `{ ok } | { ok, reason }`, plus `ws.send(data, cb)` forwarded by `adaptWs`. The
  callback is attached only when a caller asks for it, because `ws` reports a
  failure to the callback *or* as a socket `error` — attaching it everywhere
  would silently drop `core-link.connection.error` for every other frame.
- [x] **On failure leave `conn.lastSentEventId` where it was and return from the
  loop.** Exactly that; plus `rewindEventCursor` for the failure that is only
  knowable after the call returns.
- [x] **`log.warn` naming the connection and the `eventId`** ("consider"). Added:
  `core-link.event.undelivered { eventId, clientId, reason }`, and `reason`
  distinguishes a closed socket from a send that threw. One test asserts the line
  names the lost id.
- [x] **`fanOutPtyEvent` deserves the same guard.** It gets the `readyState`
  guard through the shared `send`, and stays *ungated on the result* — which the
  issue explicitly allows ("though PTY loss is recoverable via `seq`/replay").
  Stated in its doc comment so the asymmetry is not read as an oversight.

### Steps to reproduce → the regression test

- [x] **A unit-level reproduction against `PtyCoreLinkServer` with an injected
  socket that reports failure the way `ws` does — without throwing.** The fake
  models both real modes: `readyState` leaving `OPEN` with no close handshake,
  and a frame accepted then reported failed a tick later.
- [x] **Assert on the delivered stream after a reconnect; event 2 never
  appears.** Four tests, each red on `da2af3e`:
  1. replays an event whose send failed instead of skipping it forever — old:
     `[1, 3]`
  2. keeps the cursor put while the socket is unwritable, and delivers on
     recovery — old: consumed by the refusal, never delivered
  3. stops the subscribe replay at the last event the socket took — old:
     `eventsReplayed` reports a tail only partly written
  4. names the undeliverable event in a log line — old: only
     `core-link.connection.error`, which names no event

### Beyond the issue's list

- [x] `handleSubscribe`'s replay carried the identical defect (cursor set to the
  end of a tail regardless of delivery) and is fixed on the same evidence. Not
  named in #244; it would have left half the hole open.

### Out of scope, deliberately untouched

- [x] **#208** — the `events-tail-cursor` flake sits on these same semantics and
  comes *after* this ticket in the train. Its harness is unmodified; the suite is
  green here (`@actana/cli`, 30 files / 421 tests).
- [x] **#243** (session settle, hook ack, boot reconciliation) and **#242**
  (`session-drive-register.ts`, panel `router.ts`) are in flight in parallel and
  own those files. This branch changes two files, both in `packages/core`:
  `src/pty-core-link-server.ts` and the new test.
- [x] No merge performed; `main` and `beta/0.3.1` untouched; PR left as a draft.

### Verification

Local (`env -u AC_*`): `@actana/core` 60 files / 1143 tests green ·
`@actana/sdk` 19 files / 236 tests green · `@actana/cli` 30 files / 421 tests
green · `pnpm typecheck` clean · `pnpm lint` 0 errors · `commitlint` clean over
every commit on the branch.

CI on this PR: all 17 checks green on `0343874`
(`Unit Tests`, `Lint`, `Typecheck`, `Conventions`, `Train rules`, both installer
E2Es, both tarball smokes). The first attempt at `Unit Tests` went red on
`events-tail-cursor.test.ts > resumes where the last run stopped` timing out at
60 s — **#208's exact signature**, which also bails `pnpm -r` before the core and
panel suites run. It passes on re-run, and three consecutive local runs of that
file are green on this branch. Left alone, as #208 owns it.

mrfqcentic (approved):
## Review — #245 against #244

Reviewed from a clean worktree at `0343874`, diffed against `origin/fix/session-and-event-reliability` (merge-base `da2af3e`). Every claim below was checked by reading the enforcing line, not the PR description. **Approving.**

### The two load-bearing claims

**1. The cursor advances only behind a send that landed — verified.**

Three sites assign `conn.lastSentEventId`, and all three are gated:

- `packages/core/src/pty-core-link-server.ts:978-979` — `if (!this.deliverEvent(conn, event)) return;` precedes `conn.lastSentEventId = event.eventId;`. `return`, not `continue`, so nothing leapfrogs the hole and the next tick re-reads the tail from the same `after`.
- `:1698-1701` — the `subscribe` replay `break`s on refusal and assigns the last *accepted* id, so `eventsReplayed` (`:1702`) no longer reports the end of a tail it only partly wrote.
- `:1009-1012` — the async rewind, guarded by `if (conn.lastSentEventId < eventId) return;`, which makes two failures in one tick settle on the earliest hole rather than fight over it. `eventId - 1` against `readEventTail`'s exclusive `>` re-includes exactly the failed event — no over- or under-rewind.

Both failure signals feed that predicate: the pre-write refusal at `:1811` (`WS_CLOSING`/`WS_CLOSED`) and the `catch` at `:1830` for a send that throws. So a socket that drops and a send that throws both leave the cursor put. `fanOutPtyEvent` picks up the same `readyState` guard through `send` (`:1777`) and stays ungated on the result, which is what #244 explicitly permits ("though PTY loss is recoverable via `seq`/replay").

The deviation from the issue's literal wording — refuse `CLOSING`/`CLOSED` rather than require `OPEN` — is correct, not a shortcut. `packages/sdk/src/__tests__/fake-core-link.ts:32` starts `readyState` at `0` and only reaches `1` at line 149, and the Core writes `ready` at accept time; demanding `OPEN` would have swallowed those first frames. The reasoning is written down at `:433-441` rather than left for the next reader to rediscover.

**2. The regression test is genuinely red on the old ordering — verified by reconstructing the old code, not by reading the table.**

Against `git show origin/fix/session-and-event-reliability:packages/core/src/pty-core-link-server.ts` (old `pushLiveEvents:921-924`, old `handleSubscribe:1596-1601`, old `send:1670-1676`):

| Test | Why it fails on the old code |
| --- | --- |
| `replays an event whose send failed…` (`:232`) | The fake never throws — it emits `error`, which `ws.on("error")` (`:665`) only logs — so old `send` returns normally and `:923` advances past event 2. Delivered set is `{1, 3}`; the reconnect asks for `> 3`. Assertion expects `[1, 2, 3]`. **Red.** |
| `keeps the cursor put while the socket is unwritable` (`:249`, `:256`) | `core-link.event.undelivered` does not exist in the old file at all (0 grep hits), so the `waitFor` times out; and `:256` fails too, since 2 and 3 were consumed by the refusal. **Red twice.** |
| `stops the subscribe replay at the last event the socket took` (`:273-274`) | This is the one failure the old `try/catch` did catch — and it still ran `lastSent = event.eventId`, giving `[1, 3]` and `eventsReplayed: 3`. **Red.** |
| `names the undeliverable event in a log line` (`:289`) | Same missing tag. **Red.** |

The fake is honest about the transport: `send` never throws on an unwritable socket, it reports through the callback or the `error` event on a later tick (`core-link-event-cursor-advance.test.ts:52-69`) — which is the precise detail that made the original `try/catch` a no-op.

### Non-blocking follow-ups

None of these block the merge; recording them so they are not rediscovered later.

1. **`pty-core-link-server.ts:1015-1021` — the warn is not throttled.** A socket stuck in `CLOSING` re-warns the same `eventId` every `liveEventPollMs`, bounded only by `HEARTBEAT_TIMEOUT_MS` (45 s), and only where `ws.ping` exists (`:630`) — order ~90 identical lines per stuck connection. `packages/core/src/log-throttle.ts` exists for exactly this ("a boot blocker that misfires on every 500 ms poll tick") and is already used by `event-log-store.ts`, `core-query-store.ts` and `project-roots.ts`. Related: `:975-976` reads up to 1000 tail rows *before* the `readyState` check, so a dead connection pays a full tail read per tick; hoisting the guard above `readEventTail` would fix both cheaply.

2. **`:989-993` leans on the delivery callback being asynchronous, and `WebSocketLike` does not say so.** If a transport invoked `cb(err)` synchronously, `rewindEventCursor` would run while `conn.lastSentEventId` is still below `eventId`, take the `:1010` early return, and `:979` would then advance past the lost frame — the exact defect this PR closes. Node `ws` defers (`process.nextTick`) and every fake here uses `queueMicrotask`, so there is no live impact; a closure flag consulted at `:993`, or a line in the `send` doc at `:401-409` stating the callback must not fire synchronously, would nail it down.

3. **The at-least-once trade interacts with the client's dedupe in one mode.** `durable-core-client.ts:231` drops `event.eventId <= this.lastEventId`, and `:242-248` raises the persisted cursor to the `eventsReplayed` marker. So in the accepted-then-failed-later mode, if any frame behind the failed one *does* land, the client's cursor has already moved past the hole and the server's retry is discarded client-side. In practice a `ws` delivery-callback error means the socket is finished and the following frames fail too, so the connection drops and the reconnect replay closes the hole off the client's real cursor — which is why I am not treating this as a defect. Worth noting that the test's `delivered` set (`core-link-event-cursor-advance.test.ts:231`) is measured at the socket, so it asserts a slightly weaker property than "the Panel ends up holding the event". Closing it properly needs the per-frame ack the PR already identifies as a separate, larger ticket.

4. Cosmetic: `logi` (`core-link-event-cursor-advance.test.ts:192`) reads as a typo for `log`; `eventLog` is the glossary term. And `:250` hand-rolls `ws.sent.filter((raw) => raw.includes('"event"'))` where the file's own `eventIds()` helper would do.

### On scope

`handleSubscribe` is not named in #244 and carried the identical defect; fixing it here is right — leaving it would have kept half the hole open. #208 (the `events-tail-cursor` flake) is correctly left alone: it is scheduled after this ticket, owns that harness, and its suite is unmodified here. Not grounds for changes.

### Delivery

Targets the open train `fix/session-and-event-reliability`, not `main`. Both commits and the PR title are conventional-commit clean, `Closes #244` present, template filled with a real "how was this tested" section. All checks green on `0343874`, including `Unit Tests`, `Typecheck`, `Lint`, `Conventions` and `Train rules`. Two files changed, both in `packages/core` — no overlap with #242/#243.

**Verdict: approve.** The ordering the ticket is about is enforced at the line, and the tests pin it against the old behaviour rather than the new.

qcentic-dev:
## Unit Tests failure on run 32132593072: known CI flake, not a regression

**Diagnosis: infrastructure / known-flake — no code change made, and none needed.** The job is green on a `--failed` re-run at the same sha.

### What actually failed

One test, one file, and it was a timeout rather than an assertion:

```
FAIL src/__tests__/events-tail-cursor.test.ts > actana events tail, across a Core restart
     > resumes where the last run stopped, and does not replay it
Error: Test timed out in 60000ms.
 ❯ src/__tests__/events-tail-cursor.test.ts:125:3

 Test Files  1 failed | 29 passed | 1 skipped (31)
      Tests  1 failed | 420 passed | 3 skipped (424)
```

This is the failure tracked in #208 — the same test title has gone red and then green on a plain re-run on #206, #215, #219 and #221, including on commits touching nothing near it. It stops a Core and starts another on the same port, which makes it the file most sensitive to a loaded runner.

I did **not** take that at face value, because this branch changes the live-event cursor and the red test is a cursor test. Two things rule the regression reading out:

1. **The failure mode is wrong for this change.** A cursor advanced too far or too little shows up as a wrong id list (`expected [4,5]`, a skip, or a duplicate), not as a 60 s hang. And the sibling test that genuinely exercises this diff's path — `delivers every event exactly once, in order, over a dropped connection`, the one that drops a socket mid-push and restarts the Core — **passed in the same run** (1022 ms). The test that timed out restarts the *CLI*, not the connection, and never puts a socket into `CLOSING`/`CLOSED`.
2. **Base and branch are identical locally.** In `wt-244` I ran the same suite on the merge base `da2af3e` and on `0343874`:

   | | full `packages/cli` suite | `events-tail-cursor.test.ts` ×10 |
   |---|---|---|
   | base `da2af3e` | 421 passed, 0 failed | 10 / 10 green |
   | branch `0343874` | 421 passed, 0 failed | 10 / 10 green |

   The failing test completes in ~2.7 s locally against a 60 s budget, on both sides.

It is also **not** the ENOSPC/disk exhaustion that hit this machine earlier today: `ENOSPC` and `no space left` appear zero times in the full run log, and the run was on a GitHub runner rather than that box.

### Worth knowing before you merge

The red attempt **never ran this PR's own tests.** The root script is `vitest run && pnpm -r test`, `pnpm -r` bails on the first package to fail, and `packages/cli` sorts before `packages/core` — so the failed log contains `packages/{sdk,shared,cli} test$` and no `packages/core` or `packages/panel` line at all. The 299-line `core-link-event-cursor-advance.test.ts` added here was simply never executed.

The green re-run did execute it: `packages/core` 60 files / **1143 passed**, `packages/panel` 140 files / **1407 passed**. So the re-run was load-bearing rather than cosmetic, and the change is now genuinely covered on CI. I had also run `packages/core` locally on the branch beforehand — same 1143 passed.

### Status

`Unit Tests` **pass** (1m52s); all 19 non-skipped checks on the run are green, run conclusion `success`. No commits pushed, nothing merged, base untouched.

mrfqcentic (approved):
Reviewed at `0343874` against `origin/fix/session-and-event-reliability` (= `da2af3e`, the merge-base), reading the enforcing lines rather than the PR body. **Approving.** Every acceptance criterion in #244 is met, or is declared not met deliberately and accurately. Notes below are follow-ups, not blockers.

## The two claims I set out to disprove

**1. The cursor advances only behind a send that landed.** Holds at every assignment site, on all three failure shapes:

- `packages/core/src/pty-core-link-server.ts:978-979` — `if (!this.deliverEvent(conn, event)) return;` sits *before* `conn.lastSentEventId = event.eventId;`. A refusal leaves the cursor below the event and returns, so the next tick re-reads the tail from the same point.
- `pty-core-link-server.ts:1811-1813` — a socket already `CLOSING`/`CLOSED` is refused *before* `serializeCoreLinkFrame`, which is the reported remote-Core mode.
- `pty-core-link-server.ts:1826-1829` — a send that throws returns `{ ok: false, reason }`. The old `try/catch` swallowed it and advanced anyway.
- `pty-core-link-server.ts:1009-1012` — the async half. The guard `if (conn.lastSentEventId < eventId) return;` only ever moves the cursor backwards, so two failures in one tick converge on the earliest hole instead of fighting.
- `pty-core-link-server.ts:1697-1702` — the `subscribe` replay `break`s on the same evidence and reports the reached event in `eventsReplayed`, not the end of a tail it only partly wrote. That path was not named in #244 and did carry the identical defect; good catch.

**2. The tests are red on the old ordering** — checked by reconstructing the old code (`pushLiveEvents` 912-925, `send` 1670-1676, `handleSubscribe` 1584-1602), not by reading the table in the PR body:

| Test | What old code does |
| --- | --- |
| `core-link-event-cursor-advance.test.ts:232` | old `:922-923` sends event 2 with no callback (the fake emits `error`, never throws), then advances to 2; 3 lands. Client cursor 3, re-subscribe tail empty → `[1,3]` vs expected `[1,2,3]`. **Red.** |
| `:249` and `:256` | old has no `readyState` check and no such log line; the `waitFor` on `core-link.event.undelivered` times out, and the recovery assert gets `[1]`. **Red twice.** |
| `:273-274` | old `:1597-1598` catches the throw and still sets `lastSent`, delivering 3 → `[1,3]` and `eventsReplayed.lastEventId` 3. **Red.** |
| `:289-297` | no `core-link.event.undelivered` exists on old. **Red.** |

None of the four passes against the old ordering. The `readyState`-refusal path (`goUnwritable`) also models `ws` correctly: it moves state without a close handshake and never throws.

The deviation from the issue's literal "guard on `readyState === OPEN`" is justified as written — `packages/sdk/src/__tests__/fake-core-link.ts:32` starts at `0` and only flips at `:149-150`, while the Core writes `ready` on accept, so requiring `OPEN` would swallow that first frame. The doc comment at `:429-440` states the reasoning.

## Follow-ups (none blocking)

**1. The async rewind is neutralised at the real client — worth its own issue.** `packages/sdk/src/durable-core-client.ts:231` dedupes with a high-water mark: `if (event.eventId <= this.lastEventId) return;`. So in the callback-failure mode where event *N* fails *after* being accepted and *N+1* already landed, the client's cursor is at *N+1*; the server rewinds and re-sends *N*, and the client drops it. The hole survives end-to-end in that shape, and `rewindEventCursor`'s comment at `:997-1007` reads as though re-delivery closes it.

This is not grounds to hold the PR: the dominant mode #244 actually reports (link goes unwritable → `CLOSING`/`CLOSED`) is refused *before* the write, so nothing after the hole is ever sent and the client's cursor cannot run past it — that path is genuinely fixed. And on a real `ws`, a frame that fails takes the socket with it, so the following frames fail too and the rewind lands correctly. The residual case is now *logged* rather than silent, which is the other half of what the ticket asked for. But a per-frame ack — or a client-side seen-set — is the real close, and that is the protocol change already scoped out of this ticket.

Related: the headline test at `:231-232` collects the delivered set off the fake socket, which records every frame; the real client would have dropped the retried event 2. The test proves the server-side ordering, which is what it is for — it just does not prove the end-to-end property its comment at `:227-230` describes.

**2. `WebSocketLike.send`'s callback contract should say "never synchronously"** (`pty-core-link-server.ts:401-410`). The ordering is safe only because real `ws` defers to `process.nextTick`. A transport that invoked the callback synchronously would rewind at `:1011` and be clobbered immediately by `:979` (or `:1701`), silently restoring #244. Nothing in the interface doc forbids it today.

**3. Stuck-`CLOSING` repetition.** `readEventTail` at `:976` runs before writability is consulted at `:1811`, and `warnEventUndelivered` has no dedup — a socket sitting in `CLOSING` yields one event-log read plus one identical warn per poll tick per connection until the close handler reaps it. Bounded, but a `readyState` short-circuit at the top of `pushLiveEvents` would remove both.

**4. Trivial.** `core-link-event-cursor-advance.test.ts:182` names the event-log fixture `logi` to dodge `import log from "../log"`; siblings use `log` for it. `eventLog` is free and is the domain term.

## Checklist

- [x] Cursor advances only behind a landed send — sync refusal, throw and async callback all covered
- [x] Refused event retried on a later tick; loop stops with the cursor below it
- [x] Not skipped — no path advances past a frame the socket did not take
- [x] "Exactly once" — correctly declared **not met**; at-least-once and in order, with the trade argued in code
- [x] `readyState` guard, with the `CLOSING`/`CLOSED` deviation justified against a real caller
- [x] Delivery result learned via both `sendResult` and the forwarded `ws.send(data, cb)`
- [x] `log.warn("core-link.event.undelivered", { eventId, clientId, reason })` names the lost event
- [x] `fanOutPtyEvent` guarded, deliberately ungated on the result per the issue's own carve-out
- [x] Regression test genuinely red against the old ordering — verified by reconstruction, all four
- [x] `handleSubscribe` replay fixed too (beyond the issue's list, and it needed it)
- [x] Scope: two files, both in `packages/core`; #208's harness untouched, which is correct — it is scheduled after this ticket
- [x] Delivery: targets the train branch, conventional commits, `Closes #244`, all 17 CI checks green on `0343874`

Issue #186, opened 2026-08-11
Panel session drive register: pane/tab granularity mismatch — duplicate watch reorders the queue, pane close drops tab interest (silent dual-write)


qcentic-dev:
Follow-up from the review of #184 (`feat(core-link): a Core serves many clients, one holds a Session's write lock (#140)`).

## The mismatch

The Panel session drive register keys interest **per tab**, but the browser emits `watch` and `drop` **per pane**, with no refcounting in between.

- `session-drive-register.ts` lines 65-73 — interest is keyed by tab
- `panel-link-client.ts` lines 278-294 — a pane sends `watch`
- `router.ts` lines 353-370 — a pane sends `drop`

Because a tab can hold two or more panes on the same Session, pane-level events and tab-level bookkeeping disagree.

## Two concrete failures

**1. Duplicate `watch` silently moves the keyboard to the back of the queue.**
A second pane on the same Session sends another `watch`. The register already holds that tab's interest, so nothing new is registered — but the send re-enters the queue at the tail. The keyboard silently moves to the back of the queue with no user-visible cause.

**2. Closing one of two panes drops the tab's interest while the other pane stays writable — silent dual-write.**
Two panes on the same Session, close one. Its `drop` clears the tab's interest outright, since there is no refcount to say a second pane still holds it. The surviving pane keeps its writable UI and keeps sending writes, while the register believes the tab released the Session. Another client can then take the write lock: two writers, no conflict signalled to either — a silent dual-write.

## Fix direction

- **Dedupe the `watch` send client-side**, or introduce a **position-preserving `want`** so re-asserting existing interest cannot reorder the queue.
- **Add a release refcount** so a tab's interest is only dropped when the last pane on that Session releases it.

Whichever way it goes, pane-level events and tab-level interest need to agree on who holds what.

## References

- Found in review of: #184
- Underlying model: #140

mrfqcentic:
Related: #242 is the same register (`SessionDriveRegister`) from a different angle — a **reloaded** tab, rather than two live panes in one tab.

Worth noting for whoever picks this up: the fix direction proposed here — a position-preserving `want` so re-asserting existing interest cannot reorder the queue — does not cover the reload case, because a reloaded tab arrives as a *different holder object* and therefore has no position in the queue to preserve. #242 proposes a per-tab id that survives a reload (`sessionStorage`), mirroring the core-link's `clientId`/`reclaim` (ADR 0024 D9).

[...]

PR #246, opened 2026-08-18
fix(panel): key the Session drive register by client id so a reloaded tab keeps driving (#242)

qcentic-dev:
Fixes the read-only reload in #242: a tab that reloads onto the Session it alone was driving now keeps driving it, instead of waiting out the 45-second panel-link heartbeat behind the socket it just replaced.

## The bug

`SessionDriveRegister` keyed interest by the object identity of the `PanelLinkSession` — the browser's socket. A reload destroys that object, so the returning tab was a stranger to the register. Through a reverse proxy or Docker's NAT the predecessor's `close` is routinely still in flight, so first-come appended the returning tab *behind its own ghost*, the router answered `{ driving: false }`, and the pane rendered read-only until `HEARTBEAT_TIMEOUT_MS` reaped the corpse.

## The fix

The register is keyed by a **client id** the tab carries across its own reload, exactly as the core link solved the same problem one layer down (ADR 0024 D9, #146).

- **The id.** Minted per tab, presented on the panel-link upgrade query beside the protocol version — early enough for the router to have it at `attach`, which is before a returning pane could be arbitrated as a stranger. It lives in `sessionStorage`, *claimed* on load and parked only on `pagehide`, so a "Duplicate tab" that copies storage wholesale copies no id. Two live tabs presenting one string would reap each other's sockets forever, which is the failure ADR 0024 D9 already rules out for a durable Core client id.
- **The handover.** The successor takes the id over in one assignment, and only then is the predecessor retired — so there is no instant in which the Session is driven by nobody. That is `SessionLockTable.transferAll`'s argument with a map in place of a lock table. The retired predecessor's `releaseDrives` finds an id that is no longer its own and gives back nothing, which is what keeps the ordinary close path safe to leave alone.
- **The answer.** The asking tab's `drive` answer is now read back out of the register rather than inferred from what moved. A tab re-announcing a Session it already drives moves nothing, so `gained` is empty in exactly the case the answer most needs to be `true` — the second half of the stranding.
- **No id, no change.** An absent or malformed id means "this socket is its own client", which is the behaviour every tab had before this branch. Not a refusal, and the protocol version does not move.

- **#186** (pane vs. tab granularity) comes after this in the same train. The position-preserving `want` and the release refcount are its, and the pane/tab semantics here are untouched. A stable per-tab id is what makes its refcount meaningful, as #242 notes.
- **#243** (core session settle, hook ack, boot reconciliation) and **#244** (live-event cursor advance in the core-link send path) are being implemented in parallel and are not touched. Every change here is under `packages/panel/src/{server/panel-link,lib,shared}`, plus `CONTEXT.md` and ADR 0024.

## Verification

- Four of the new router tests go red against the register keyed by socket identity; the rest guard what must not move (two genuinely different tabs still arbitrate first-come, an explicit take still says `handover`, a tab that really goes away still hands the drive on).
- `pnpm typecheck` clean across the workspace. `pnpm lint` reports 0 errors.
- Panel, SDK, shared, CLI and root suites pass. Two full-suite runs each showed a handful of unrelated 5s vitest timeouts under parallel load — a different set each run, none in the panel-link files — and every one of them passes in isolation.

Refs #242


qcentic-dev:
## Acceptance criteria — #242

Issue 242 has no `Acceptance criteria` heading, so these are taken verbatim from its own **Expected behavior**, **Actual behavior** (inverted), **Fix direction / Suggested shape** and **Prior art** sections.

### Expected behavior

- [x] **The reloaded tab resumes driving the Session it was driving before the reload.** The register is keyed by a client id the tab carries across its own reload, so the entry the returning tab re-announces against is its own. `session-drive-reload.test.ts` → *"is driving it again on the gesture its new pane makes, not 45 seconds later"*.
- [x] **A tab returning to its own Session does not wait behind the socket it just replaced.** Asserted with the predecessor session still attached, which is the proxy/NAT case the report is about.

### Actual behavior, which must no longer happen

- [x] **The pane no longer mounts read-only.** The tab is never told `{ t: "drive", driving: false, reason: "watch" }` on a reload — asserted as the *whole* frame sequence, not just the end state, because one read-only frame is the report. → *"is never told it is following its own ghost"*.
- [x] **Recovery no longer depends on the ~45s `HEARTBEAT_TIMEOUT_MS` reap.** The predecessor is retired at attach; the answer is correct on the first gesture.
- [x] **The loopback ordering (predecessor `close` lands first) still works.** → *"drives it just the same when the close lands first, as on loopback"*.

### Fix direction — "Suggested shape"

- [x] **A per-tab id that survives a reload, via `sessionStorage`.** Minted per tab; presented on the panel-link upgrade query beside the protocol version, because the router needs it at `attach` and a frame would arrive too late. Hardened beyond the suggestion: the id is *claimed* (removed) on load and parked only on `pagehide`, so a "Duplicate tab" that copies `sessionStorage` wholesale copies no id. Two live tabs on one id would reap each other's sockets forever — the failure ADR 0024 D9 already rules out for a durable Core client id.
- [x] **On connect, inherit the predecessor's position in each `interest` list rather than appending.** Satisfied, by a different mechanism than the issue sketched and deliberately so: because the list holds client ids rather than session objects, the returning tab's position *is already its own*. There is nothing to rewrite in place, so there is no rewrite to get wrong. Flagged rather than passed over silently, since the issue proposed the rewrite explicitly.
- [x] **Same atomicity argument as `SessionLockTable.transferAll` — no instant where the Session is unheld, and no second tab landing in the gap.** The successor takes the id over in one map assignment *before* the predecessor is retired; the retired predecessor's `releaseDrives` then finds an id that is no longer its own and gives back nothing. → *"keeps the drive when the predecessor's close finally lands"* and *"takes nothing off the tab that reloaded when the ghost's pane is dropped"*.
- [x] **Retire the predecessor afterwards, as `reclaimFor` does.** Its socket is closed and its PTY claims given back at attach, rather than left to the heartbeat. → *"retires the ghost socket rather than leaving it to the heartbeat"*, *"gives the ghost's pty back, because the returning tab asks for its own"*.

### Prior art — what this must not absorb

- [x] **Does not close or pre-empt #186.** The pane/tab granularity is untouched: `want` still moves a re-announcing holder to the tail behind other live watchers, and there is still no release refcount. Both are called out in the register's own comment as #186's to answer.
- [x] **Composes with #186 as the issue predicts** — "a stable per-tab id is what makes a refcount meaningful, since both then agree on what 'the same tab' is". That agreement now exists.
- [x] **Does not close #187.** Nothing in the core-link Session lock path is touched; #187's two-client scenario is unaffected either way.

### Scope boundaries this branch was given

- [x] **#243 (core session settle, hook ack, boot reconciliation) untouched.**
- [x] **#244 (live-event cursor advance in the core-link send path) untouched.** Every source change is under `packages/panel/src/{server/panel-link,lib,shared}`; the only other files are `CONTEXT.md` and ADR 0024.
- [x] **Nothing merged; `main` and `beta/0.3.1` untouched; PR left as a draft against `fix/session-and-event-reliability`.**

### Verification

- [x] **The new router tests are red against the old keying.** Four of the twelve fail on `origin/fix/session-and-event-reliability`'s register — verified by reverting only `router.ts` and `session-drive-register.ts` and re-running.
- [x] **`pnpm typecheck`** clean across all five workspace packages.
- [x] **`pnpm lint`** — 0 errors (50 pre-existing warnings, none in changed files).
- [x] **Suites green:** panel (1431), root (474), SDK (241), shared (180), CLI (424), plus the new 24.
- [x] **`pnpm panel:e2e`** passes end to end, which exercises the real upgrade path the client id now rides on — including "pre-login: the panel-link upgrade is refused" and the reconnect/replay leg.
- [x] **Conventional Commits on every commit, not just the title.** `commitlint --from da2af3e --to HEAD` reports 0 problems on all 4.

### Explicitly NOT done

- [ ] **The manual reproduction from "Steps to reproduce" was not run.** It needs a Panel deployed behind a reverse proxy or Docker's published port with a real browser reloading a live Session, and no such deployment is available here. The mechanism it demonstrates is instead asserted at the router with the predecessor session still attached, which is the same state the proxy's delayed `close` produces. Worth one manual pass on a real remote Core before this leaves draft.
- [ ] **The `panel:e2e` file-drop leg did not run at its default size.** It asks for ~6144 MiB free and this machine's disk is 100% full (550 MiB free), so it was run with `AC_E2E_FILE_DROP_BYTES=8388608`. Unrelated to this branch, and every other leg ran unmodified.
- [ ] **Two full-suite panel runs each showed a few unrelated 5s vitest timeouts under parallel load** — a different set each run (`live-read-path`, `project-picker-switch`, `project-dialog-path-immutable`), none in files this branch touches, and all passing in isolation. Not fixed here; noted so it is not read as this branch's noise.

mrfqcentic (approved):
Reviewed at `1275831` against #242, in a clean worktree on the PR head, diffed three-dot against `origin/fix/session-and-event-reliability`. I read the enforcing lines rather than the description, and I checked the red-test claim by tracing each test against the old keying rather than by taking the count on trust (I could not execute the suite here — see *What I could not run*).

**Verdict: approve.** The keying is correct everywhere it is written and read, the reconnect neither strands nor double-registers in either close ordering, and four of the twelve router tests are genuinely red against socket identity for behavioural reasons, not shape.

## The three claims I checked directly

**1. The register is keyed by client id, everywhere.** `DriveClientId` is a `string` (`session-drive-register.ts:47`) and no `PanelLinkSession` can reach the register by construction. Every call site passes `session.clientId`: `driveFor` at `router.ts:368`, both `release`/`want` at `router.ts:403-404`, the asker's own answer at `router.ts:440`, `releaseAll` at `router.ts:472`. `sessionsByClientId` is used only to turn an id back into something to write a frame on (`router.ts:250`). A grep across `packages/panel/src` finds no other reader or writer. Met.

**2. A reconnect strands nothing and double-registers nothing.** The ordering at `router.ts:225-228` is load-bearing and correct: the successor takes the id (`:227`) *before* `predecessor.retire()` (`:228`), so the predecessor's `detach` → `releaseDrives` hits the guard at `router.ts:470`, finds the id is no longer its own, and gives back nothing. `forget` (`router.ts:243-245`) deletes the map entry only when the leaving session still owns it, so the live tab's drives never resolve to nothing. `detach` is idempotent (`router.ts:663`, `:797`), so the proxy's late `close` landing on the retired socket is a no-op. The reverse ordering (loopback, `close` first) is also safe: the predecessor still owned the id, released normally, and the returning tab re-announces into an empty list. Double-registration is impossible at the source: `want` filters the id out before re-inserting (`session-drive-register.ts:105-106`), so one id appears at most once. Met.

**3. The four red tests prove behaviour.** Tracing `session-drive-reload.test.ts` against socket-keyed `router.ts` + `session-drive-register.ts`, exactly these four fail, which matches the claim:

- `:157` "is driving it again on the gesture its new pane makes" — old code appends behind the live ghost, `gained` is empty, the asker is told `driving: false`.
- `:170` "is never told it is following its own ghost" — same mechanism, asserted as the *whole* frame sequence rather than the end state. This is the strongest test in the file: the report is one read-only frame, and `toEqual([{driving: true, reason: "watch"}])` is the only assertion here that a "recovers eventually" implementation cannot pass.
- `:209` "retires the ghost socket" — old code has no `retire()`; `ghost.closed` stays `false`.
- `:218` "gives the ghost's pty back" — under old keying the predecessor keeps its claim, so `claimPty`'s refcount (`router.ts:308-310`) short-circuits and the fake link records one `subscribe` instead of subscribe/unsubscribe/subscribe.

These are frames written to a returning tab's socket and calls made on a fake core link, with the predecessor session *still attached* — the proxy/NAT state the report is about — not assertions about the shape of the register. The remaining eight are honest guards, and I confirmed three of them (`:183`, `:197`, `:244`) pass under the old keying too, which is what a guard should do.

## Acceptance criteria

| From #242 | Verdict | Enforcing line |
|---|---|---|
| Reloaded tab resumes driving the Session it was driving | **Met** for the reported case (sole watcher). See residual below for two live watchers. | `router.ts:440` reads the answer out of the register; `router.ts:227` |
| A returning tab does not wait behind the socket it replaced | **Met** | `router.ts:470` guard + `session-drive-register.ts:105` |
| Pane no longer mounts read-only / no `driving:false` frame | **Met** | asserted as a whole sequence, `session-drive-reload.test.ts:180` |
| Recovery no longer depends on the 45s heartbeat reap | **Met** | `retire()` closes the ghost at attach (`router.ts:~640`) |
| Per-tab id surviving a reload, via `sessionStorage` | **Met**, and hardened past the suggestion — claimed on load, parked on `pagehide` (`panel-link-client.ts:786-787`), so a duplicated tab copies no id | `panel-link-client.ts:755+` |
| Inherit the predecessor's position rather than appending | **Met by a different mechanism**, correctly flagged rather than glossed: because the list holds ids, the entry already *is* the returning tab's, so there is no in-place rewrite to get wrong | `session-drive-register.ts:98-110` |
| `transferAll`'s atomicity — no unheld instant, no second tab in the gap | **Met** | `router.ts:227-228`, `:470` |
| Retire the predecessor afterwards, as `reclaimFor` does | **Met** | `router.ts:228` |
| Does not close or pre-empt #186 / #187 | **Met** | pane/tab granularity untouched; nothing in the lock path changed |

## Non-blocking notes

1. **Naming contradicts the glossary entry this branch adds.** `CONTEXT.md` coins **Panel tab id** with `_Avoid_: … client id (that is the core-link one)`, and then the identifiers are `clientId` throughout — `DriveClientId` (`session-drive-register.ts:47`), `sessionsByClientId` (`router.ts:152`), `readPanelLinkClientId` (`shared/panel-link.ts:100`), `mintPanelLinkClientId` (`router.ts:827`). The client side already says `claimTabClientId`. `tabId`/`readPanelTabId` would settle it; the wire param `c` is fine either way. Worth a follow-up rename, not a reason to hold this.

2. **The one seam with no test.** The browser half (`panel-link-tab-id.test.ts`) proves the id reaches the dial URL; the router half proves keying by an *injected* id. Nothing asserts the join: `readClientId(request)` → `bindPanelLinkSocket(router, ws, clientId)` (`ws-server.ts:45-47`) → `attach`. `ws-server.test.ts` already stands up a real upgrade, so this is a short test, and it is the only line of the fix a typo could silently disable.

3. **A reload now costs a PTY unsubscribe/resubscribe at the Core.** `retire()` → `detach()` gives the claims back immediately, where the old refcount kept the stream warm until the heartbeat. The returning tab re-subscribes with `catchUp: true` so nothing is lost, and `session-drive-reload.test.ts:237-241` pins the sequence deliberately — recording it because it is a real change in what a Core sees on every reload.

4. **`pagehide`/`pageshow` listeners are never removed** (`panel-link-client.ts:786-787`). Bounded in practice: the client is a singleton (`panel-bridge.ts:270`).

5. **Two identical mint functions** (`router.ts:827`, `panel-link-client.ts:815`), kept in step by a comment. `shared/panel-link.ts` already hosts `readPanelLinkClientId` and is the obvious home.

## Residual, and deliberately so

With two live tabs on one Session, the reloading tab still tail-appends and is told `driving: false` — `want` strips and re-appends (`session-drive-register.ts:105-106`). That is the position-preserving `want` that belongs to #186, which is scheduled after this, and the register's own comment (`session-drive-register.ts:92-96`) says so in the right place rather than leaving it to be discovered. Correct call: #242's reproduction is explicitly one tab ("Do not open a second tab"), and the client-id keying is precisely what makes #186's refcount meaningful. Not a change request.

## What I could not run, and whether it matters

- **The manual reverse-proxy reproduction.** Non-blocking. The only thing a proxy changes is that the predecessor's `close` has not been processed at `attach`, and that state is asserted directly — every reload test attaches while the old session is live, and both orderings are covered (`:183` late close, `:197` close first). The genuine residual is not the router but the browser: that a real `sessionStorage` claim/park survives a real reload and arrives on the upgrade query is asserted against a stand-in `window`, and no e2e leg reads the `c` param. One manual pass on a real remote Core before this leaves draft, as the author already proposes, closes that — a pre-merge check, not a change to make.
- **The `panel:e2e` file-drop leg at default size.** Non-blocking and unrelated: every source change here is under `server/panel-link`, `lib`, `shared`. A disk-constrained upload size cannot mask a defect in this diff.
- **The suite itself.** I could not execute it — this machine is also at 100% disk with no installed workspace deps — so the pass/fail counts in the checklist are taken as reported. The red-test claim is the one I did not take on trust, and it holds by reading.

The checklist comment is accurate against the code, including where it declines a checkbox. Approving.

PR #247, opened 2026-08-18
fix(core): settle Core-owned Sessions — hook ack, an unarmed backstop, and a boot sweep

### LINKED ISSUE #243: Core: nothing settles a Core-owned Session whose process died — a lost Stop hook arms no backstop, and there is no boot sweep

qcentic-dev:
## Summary

Three independent ways a Core-owned Session gets left on `running` after its work is over, and no backstop that catches
the important one. This closes all three, and they only work as a set — the sweep answers a restart, the backstop
answers a lost `Stop`, and the ack is what makes either of them explicable after the fact.

- **The hook ack.** A hook's POST ended in `|| true`, which swallowed a timeout, a connection refused, a 401 and a 500
  identically, so a lost hook left no trace on either side. The command now fails on a non-2xx answer (`-f`), retries a
  transient one (`--retry 2 --retry-delay 1` — the timeout a busy Core hands a `-m 3` POST), and appends what it still
  could not deliver to `$AC_HOOK_MISS_LOG`. The receiver counts the hooks it accepts and answers with that delivery
  number. `HookDeliveryMonitor` drains the miss log into the Core's log with a running total, starting at boot — a hook
  refused during a restart is exactly the drop nobody could otherwise hear about. OpenCode's plugin does the same in its
  own idiom.
- **The unconditional backstop.** `armDeferredFinish` is only reached from branches that need a hook to have *arrived*,
  so when the terminal `Stop` is itself the POST that drops, nothing is armed at all. `CoreSessionBackstop` arms nothing
  and needs nothing armed: once a minute it asks the database which rows still claim to be working and settles the ones
  that have gone quiet. Quiet is the load-bearing word — a working harness fires a `PostToolUse` per tool call and
  redraws its spinner into the PTY about once a second, so hooks *and* PTY output count as activity, the row's own
  `updatedAt` is the floor, and fifteen minutes with neither is a turn that ended without anyone being told.
- **The boot reconciliation sweep.** A Core's PTYs die with it — no exit callback, no `pty:exit`, no settle — and
  nothing swept at startup. The Panel has had this sweep for its own rows all along; what it never had was scope over
  Core-owned ones, which is every Session on a remote Core. `sweepStrandedSessions` runs once per boot, before any PTY
  of this process can exist, and writes `disconnected` through `CoreTaskWriter` so each settle appends the
  `task:updated` event a connected Panel re-renders from.

`disconnected` rather than `finished` for the sweep: it is what the Panel already uses for a process that went away
without reporting, and it claims nothing about how the work ended.

### The one trade, stated plainly

A harness that works in *total* silence — no hook, no byte of output — for longer than fifteen minutes gets a card
reading `finished` while it is still going. Nothing is killed, no process is touched, and the next turn's
`UserPromptSubmit` puts the row back on `running`. Weighed against a lost `Stop` wedging a Session until a human edits
the row by hand, while the Panel says the opposite.

## Related issues

Closes #243

Scope boundaries kept deliberately: `pty-core-link-server.ts` is untouched (#187 owns its restructuring), the core-link
send path's live-event cursor advance is untouched (#244), and the Panel's `session-drive-register.ts` / `router.ts` are
untouched (#242).

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## How was this tested?

New tests, all against the Core's real SQLite, real event log and real loopback receiver rather than hand-built frames:

- `harness-hook-delivery.test.ts` (10) — drains and clears the miss log, keeps a running total across a boot, and runs
  the **actual generated hook command** through a real `sh` + `curl` against a real receiver: a delivered hook acks and
  records nothing; a Core that is not there, a wrong bearer and a task the Core does not have each leave exactly one
  recorded line; and with no miss log configured the command still exits 0 and writes nowhere.
- `core-session-backstop.test.ts` (9) — finishes a quiet Session that nothing armed (and raises the `session:finished`
  event the operator never got), leaves a two-hour turn alone while it keeps talking, never touches `needs-input`,
  calls a quiet Session with no live PTY `disconnected` rather than `finished`, and settles each row exactly once.
- `core-session-sweep.test.ts` (7) — sweeps `running` / `needs-input` including archived rows, leaves settled rows
  exactly as they settled, appends `task:updated` without a spurious `session:finished`, is a no-op on the second boot,
  and keeps going when one row cannot be written.
- `core-query.test.ts` (+4) — the boot sweep's read: only rows claiming a live process, archived included, every
  project.
- `harness-hooks.test.ts` (+1, 1 updated) — the fail-soft assertion now says what it means, next to the new ack.

Suites run locally on this branch: `@actana/core` 1166 passed, `@actana/shared` 184, `@actana/sdk` 236, `@actana/cli`
421, `@actana/panel` 1405 (2 timeouts under load that pass in isolation — `terminals-in-browser` and `panel-link`,
neither touched here). `pnpm typecheck` and `pnpm lint` clean (warnings pre-existing). The root `scripts` suite has one
pre-existing environment failure on this machine (`core-launcher` picks up a real `/opt/actana` install).

Not covered by a test: the throttled PTY-output activity callback in `pty-manager.ts`, whose only test route would be
spawning a real PTY — the module's existing tests are unit-level over its pure helpers.

## Checklist

- [x] Conventional Commits on every commit, not just the title
- [x] Docs updated (`docs/harness-status-detection.md`, `docs/observability.md`)
- [x] No new dependencies

### DIFF
--- a/docs/harness-status-detection.md
+++ b/docs/harness-status-detection.md
@@ -20,6 +20,9 @@ about everything else on a Core — as an Event replayed off its cursor.
 | Status/title writes + events | `packages/core/src/core-task-writer.ts` |
 | Title generation | `packages/core/src/core-title-generator.ts` |
 | PTY exit settle | `packages/core/src/pty-manager.ts` → `onSessionExit` |
+| Hook delivery misses | `packages/core/src/harness-hook-delivery.ts` |
+| Quiet-Session backstop | `packages/core/src/core-session-backstop.ts` |
+| Boot sweep | `packages/core/src/core-session-sweep.ts` |
 
 The Panel keeps a hook endpoint (`POST /api/hooks/<slug>`) for its own remaining
 local task rows, running the same shared pipeline. A Core-owned Session never
@@ -66,7 +69,8 @@ next turn's `Stop` for the whole TTL.
 
 Backstops, so a `SubagentStop` that never arrives (lost POST, killed process) —
 or a healed `running` that no `Stop` will ever follow — cannot hold a task on
-`running` forever:
+`running` forever. The first three are armed **by a hook that arrived**; the
+fourth is armed by nothing, which is the point (issue 243):
 
 - Tracked entries expire after 2 hours (kept long on purpose — a short TTL would
   prematurely finish sessions whose subagents legitimately run long).
@@ -80,6 +84,28 @@ or a healed `running` that no `Stop` will ever follow — cannot hold a task on
 - Tracking is dropped outright when a new session id is captured, on
   `SessionStart` with `source: "clear"` (same session id, but `/clear` kills
   background work), and when the PTY process exits.
+- **The quiet-Session backstop** (`core-session-backstop.ts`) sweeps the
+  database once a minute and settles any row still claiming `running` that has
+  gone quiet. Nothing arms it, so it covers the case the three above cannot: a
+  turn whose terminal `Stop` was the POST that dropped, where nothing was held,
+  nothing was healed and no subagent was ever tracked.
+
+Elapsed time is not what "quiet" means, and it could not be — a turn may run for
+hours. A working harness never stops talking: every tool call fires the
+unmatched `PostToolUse`, and its TUI redraws a spinner and elapsed-time counter
+into the PTY about once a second. So **hooks and PTY output both count as
+activity** (output throttled to one report per five seconds per PTY), a row's
+own `updated_at` is the floor for a Session the process has not heard from yet,
+and **fifteen minutes with neither** is read as a turn that ended without anyone
+being told. A live PTY makes that settle a `finished`; no live PTY makes it a
+`disconnected`, because a process that went away did not finish. `needs-input`
+is never swept — a Session waiting on a human may wait silently forever.
+
+The trade is knowingly paid: a harness that works in total silence for longer
+than the window gets a card that reads `finished` while it is still going.
+Nothing is killed, and the next `UserPromptSubmit` puts the row back on
+`running`. Before it, a lost `Stop` wedged the Session until a human edited the
+row by hand — while the Panel said the opposite.
 
 `Notification` is also intentionally narrowed to `permission_prompt`. Claude Code
 sends idle input reminders through the same hook event, so treating all
@@ -169,6 +195,44 @@ The harness pipes the hook payload (`hook_event_name`, `session_id`, `cwd`,
 `transcript_path`, …) on stdin; the command forwards it as the request body.
 `|| true` is load-bearing: a hook must never block or fail an operator's session.
 
+### Delivery, and what happens when there is none
+
+Fail-soft is not the same as silent, and until issue 243 this path was both. A
+`-m 3` POST against a Core whose event loop is also serving PTY fan-out, SQLite
+writes and file transfers is reachable under load, and `|| true` swallowed a
+timeout, a connection refused, a 401 and a 500 identically. Nothing on either
+side recorded the attempt, so a lost hook left no trace anywhere — and a lost
+terminal `Stop` wedged its Session on `running` with nothing to say why.
+
+Both ends now account for it:
+
+- The command adds `-f`, so a non-2xx answer is a failure rather than a silent
+  success, and `--retry 2 --retry-delay 1`, which retries what curl calls a
+  transient error — the timeout, above all. Three attempts bound the worst case
+  at about eleven seconds, well inside a harness's own hook timeout, and only on
+  the path where the Session's status is already wrong. `--retry-connrefused` is
+  deliberately absent: it would raise the curl version the command needs, and a
+  refused connection means the Core is down.
+- What still could not be delivered is appended to `$AC_HOOK_MISS_LOG`
+  (`<user-data-dir>/hook-misses.log`) as one tab-separated
+  `<iso8601> <taskId> <event> <curl exit>` line. Unset, the record goes to
+  `/dev/null` — a workspace opened by hand writes nowhere rather than failing.
+- The receiver counts the hooks it accepts and answers with that delivery
+  number, so an ack is a fact rather than an empty 200.
+- `HookDeliveryMonitor` drains the miss log into the Core's log with a running
+  total, starting at boot: a hook refused during a restart is exactly the drop
+  nobody would otherwise hear about.
+- OpenCode's plugin does the same in its own idiom — it checks `res.ok`, retries
+  once, and writes the same line.
+
+`-o /dev/null` rides along and is not cosmetic: Claude Code reads a hook's
+stdout as control JSON, so the receiver's answer had no business being printed
+there.
+
+The command still ends in `|| true`, and still means it: a delivered hook
+short-circuits the chain, a dropped one records itself first, and a record that
+cannot be written falls through to it.
+
 ## How a status change reaches the card
 
 1. The receiver hands the payload to `CoreHarnessStatus`.
@@ -200,8 +264,20 @@ Sessions whose process is gone:
   The Panel's own exit patch no longer fires for a Core-owned row: it was
   unconditional, so it would overwrite an `interrupted` Session with `finished`
   and raise a spurious `session:finished` besides.
-- **Boot sweep** — the Core's DB bootstrap stamps `disconnected` on rows left
-  `running` / `needs-input`, since no PTY of the previous run survived it.
+- **Boot sweep** — `core-session-sweep.ts` runs once per Core boot, before the
+  PTY core, the hook receiver or the core-link server can produce a Session of
+  this run. At that moment no PTY of this process exists, so every row still
+  claiming `running` / `needs-input` is an orphan of the previous one: a Core's
+  PTYs die with it, silently, with no exit callback and no `pty:exit`. Each is
+  written to `disconnected` through `CoreTaskWriter`, so the settle appends the
+  `task:updated` event a connected Panel re-renders from — a sweep nobody is
+  told about leaves the operator looking at the same wrong card.
+
+  `disconnected` rather than `finished` or `terminated`: it is the status the
+  Panel already uses for a process that went away without reporting, and it
+  claims nothing about how the work ended. The Panel has had this sweep for its
+  own rows all along; what it never had was scope over Core-owned ones, which is
+  every Session on a remote Core (issue 243).
 
 ## Terminal-input fallback
 
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -15,6 +15,24 @@ lands, and `actana status` prints the Core's health, its endpoint, whether a
 pairing token is available, and — when one exists — whether a newer release
 does too.
 
+## Log lines worth grepping for
+
+A Core says something when a Session's status is decided by something other
+than the harness reporting it (issue 243). All of them are ordinary log lines —
+there is no counter endpoint to scrape:
+
+| Line | What it means |
+| --- | --- |
+| `hook-delivery.missed` | a hook's POST never got an ack; the line names the task, the event and curl's exit |
+| `hook-delivery.missed-total` | how many drops this Core has seen since boot — one a week is a flake, forty in an hour is a Core losing to its own load |
+| `session-sweep.settled` | rows a Core restart stranded, marked `disconnected` at boot |
+| `session-backstop.settled` | a Session settled because its turn went quiet and no `Stop` ever arrived |
+
+The drops are recorded by the hook itself into
+`<user-data-dir>/hook-misses.log` and drained into the log from there, so the
+ones that happened while the Core was down show up on its next boot. See
+[harness status detection](harness-status-detection.md#delivery-and-what-happens-when-there-is-none).
+
 There is no metrics endpoint and no tracing exporter. The Panel holds no
 task-shaped state to report on — each Core owns its own
 ([ADR 0004](adr/0004-core-owns-write-path.md)) — so the useful signal is the
--- a/packages/core/src/__tests__/core-session-backstop.test.ts
+++ b/packages/core/src/__tests__/core-session-backstop.test.ts
@@ -1,0 +1,218 @@
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { bootstrapCoreDb } from "../core-db-bootstrap";
+import {
+  configureCoreMutationStore,
+  coreMutationStore,
+  disposeCoreMutationStore,
+} from "../core-mutation-store";
+import {
+  configureCoreQueryStore,
+  coreQueryStore,
+  disposeCoreQueryStore,
+  listActiveTasks,
+} from "../core-query-store";
+import {
+  appendEvent,
+  configureEventLogStore,
+  disposeEventLogStore,
+  getLastEventId,
+  readEventTail,
+} from "../event-log-store";
+import { CoreTaskWriter } from "../core-task-writer";
+import { CoreSessionBackstop } from "../core-session-backstop";
+import { clearSubagentActivity } from "@actana/shared/subagent-activity";
+
+// The backstop nobody has to arm (issue 243 part 2), against this Core's real
+// SQLite and event log.
+//
+// The case that matters is the one the drain backstop cannot reach: a turn
+// whose terminal `Stop` was the POST that dropped. Nothing was held, nothing
+// was armed, no subagent was tracked — the row simply says `running` with no
+// timer watching it. Every test below starts from exactly that row.
+
+const QUIET_MS = 15 * 60 * 1000;
+const MINUTE = 60 * 1000;
+
+describe("settling a turn whose end nobody reported", () => {
+  let userDataDir: string;
+  let writer: CoreTaskWriter;
+  let nowMs: number;
+  let livePtys: Set<string>;
+
+  const makeBackstop = () =>
+    new CoreSessionBackstop({
+      listActiveTasks,
+      writer,
+      hasLivePty: (taskId) => livePtys.has(taskId),
+      now: () => nowMs,
+      quietMs: QUIET_MS,
+    });
+
+  const insert = (taskId: string, status: string) => {
+    coreMutationStore.mutateTask({
+      op: "create",
+      taskId,
+      projectId: "p1",
+      title: taskId,
+      agent: "claude-code",
+      status,
+    });
+    livePtys.add(taskId);
+  };
+  const statusOf = (taskId: string) => coreQueryStore.getTask(taskId)?.status;
+  const kindsSince = (eventId: number) => readEventTail(eventId, 100).map((e) => e.kind);
+
+  beforeEach(() => {
+    userDataDir = fs.mkdtempSync(path.join(os.tmpdir(), "ac-session-backstop-"));
+    bootstrapCoreDb(userDataDir);
+    configureCoreMutationStore(userDataDir);
+    configureCoreQueryStore(userDataDir);
+    configureEventLogStore(userDataDir);
+    writer = new CoreTaskWriter({
+      mutationPort: coreMutationStore,
+      queryPort: coreQueryStore,
+      eventLog: { appendEvent, getLastEventId, readEventTail },
+    });
+    coreMutationStore.mutateProject({
+      op: "create",
+      projectId: "p1",
+      name: "Warehouse",
+      path: userDataDir,
+    });
+    nowMs = Date.now();
+    livePtys = new Set();
+  });
+
+  afterEach(() => {
+    clearSubagentActivity("t-1");
+    disposeCoreMutationStore();
+    disposeCoreQueryStore();
+    disposeEventLogStore();
+    fs.rmSync(userDataDir, { recursive: true, force: true });
+  });
+
+  it("finishes a Session that has gone quiet, with nothing having armed it", () => {
+    insert("t-1", "running");
+    const backstop = makeBackstop();
+    const before = getLastEventId();
+
+    // No hook was ever seen for this row, so no timer exists for it anywhere —
+    // which is precisely the state a lost `Stop` leaves behind.
+    nowMs += QUIET_MS + MINUTE;
+    expect(backstop.sweepOnce()).toEqual(["t-1"]);
+
+    expect(statusOf("t-1")).toBe("finished");
+    // The finish the operator never got: card, toast and notification all
+    // route on this event (ADR 0008).
+    expect(kindsSince(before)).toContain("session:finished");
+  });
+
+  it("leaves a turn alone while its harness is still talking", () => {
+    insert("t-1", "running");
+    const backstop = makeBackstop();
+
+    // A long turn: hours of work, output all the way through. Every chunk of
+    // PTY output and every hook lands here.
+    for (let minute = 0; minute < 120; minute += 1) {
+      nowMs += MINUTE;
+      backstop.noteActivity("t-1");
+      expect(backstop.sweepOnce()).toEqual([]);
+    }
+    expect(statusOf("t-1")).toBe("running");
+
+    // Then the turn ends and its `Stop` drops. Nothing else is coming.
+    nowMs += QUIET_MS + MINUTE;
+    expect(backstop.sweepOnce()).toEqual(["t-1"]);
+    expect(statusOf("t-1")).toBe("finished");
+  });
+
+  it("counts the row's own last write as the last thing heard from it", () => {
+    // A Session this process has heard nothing about since boot — the restart
+    // case — must not be settled for a silence that predates the Core.
+    insert("t-1", "running");
+    const backstop = makeBackstop();
+
+    nowMs += QUIET_MS - MINUTE;
+    expect(backstop.sweepOnce()).toEqual([]);
+    nowMs += 2 * MINUTE;
+    expect(backstop.sweepOnce()).toEqual(["t-1"]);
+  });
+
+  it("never settles a Session that is waiting on a human", () => {
+    // `needs-input` may sit silent forever and still be true; a timer cannot
+    // make it truer, and finishing it would hide a question.
+    insert("t-1", "needs-input");
+    const backstop = makeBackstop();
+
+    nowMs += 10 * QUIET_MS;
+    expect(backstop.sweepOnce()).toEqual([]);
+    expect(statusOf("t-1")).toBe("needs-input");
+  });
+
+  it("calls a quiet Session with no PTY disconnected, not finished", () => {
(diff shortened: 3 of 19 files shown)


qcentic-dev:
## Acceptance criteria for #243 — checked, or explicitly not

Issue #243 states no section headed "Acceptance criteria", so these are its **Expected behavior** bullets plus the
three parts this PR was asked to carry together. Each is checked against what is on this branch, and the unchecked ones
say why.

### Expected behavior (the issue's own three)

- [x] **After a Core restart, a Session whose PTY did not survive is settled — `disconnected` — and the corresponding
  event is appended so connected clients re-render.**
  `core-session-sweep.ts`, called from `core-entry.ts` before the PTY core, the hook receiver or the core-link server
  can produce a Session of this run. It writes through `CoreTaskWriter`, so each settle appends `task:updated`;
  asserted in `core-session-sweep.test.ts` against the real event log, including the absence of a spurious
  `session:finished`.

- [x] **A turn whose terminal `Stop` never arrives is settled by a bounded backstop rather than staying `running`
  indefinitely.**
  `core-session-backstop.ts`, swept once a minute. Bound is fifteen minutes with no hook and no PTY output; the test
  `finishes a Session that has gone quiet, with nothing having armed it` starts from exactly the state a lost `Stop`
  leaves (nothing held, nothing healed, no subagent tracked).

- [x] **A dropped hook is visible somewhere: a counter, a log line, a sequence gap.**
  All three shapes, in the end: the hook records the drop to `<user-data-dir>/hook-misses.log`, the Core drains it as
  `hook-delivery.missed` (a log line, naming task, event and curl's exit) plus `hook-delivery.missed-total` (a running
  counter), and the receiver answers each accepted hook with a monotonic delivery number.

### The three parts this PR carries together

- [x] **The hook ack.** `-f` so a non-2xx is a failure, `--retry 2 --retry-delay 1` for the transient one, the miss
  record for what still could not be delivered, and the receiver's delivery number. The generated command is exercised
  end to end through a real `sh` + `curl` against a real receiver in `harness-hook-delivery.test.ts` — delivered,
  Core-absent, wrong-bearer, unknown-task, and no-miss-log-configured.

- [x] **The unconditional backstop.** Arms nothing and needs nothing armed; a row nobody ever reported a hook for is as
  much in scope as one that reported ten. PTY output joins hooks as the liveness signal (throttled to one report per
  five seconds per PTY), so a long turn that keeps talking is never settled — asserted over a simulated two-hour turn.

- [x] **The boot reconciliation sweep.** See the first bullet above. `queryActiveTasks` is its read: rows claiming a
  live process, archived included, every project, behind no core-link frame.

### Things the issue raises that this PR deliberately does **not** do

- [ ] **Shorten `ACTIVE_SUBAGENT_TTL_MS` (the ~2h03m a lost `SubagentStop` can hold a row).**
  Left at two hours on purpose — `docs/harness-status-detection.md` says these values were tuned against real failures
  and not to change them without a failure that says so, and a short TTL prematurely finishes deep fan-outs. The wedge
  it produces is nonetheless bounded now: the new quiet-Session backstop settles such a row after fifteen minutes of
  silence without consulting the TTL at all.

- [ ] **Triage #187's second symptom against this issue.**
  #187 owns the wider restructuring of `pty-core-link-server.ts` and is not in this train; the issue asks for a triage
  decision, which is a maintainer's call, not a code change. `pty-core-link-server.ts` is untouched here.

- [ ] **Reproduce end to end on a live Core (repro paths A and B, and the wedged row in the reported database).**
  Not run: no live Core with that database was available to this change, and neither path is reachable from the test
  suite — A needs a daemon restart, B needs the POST window suppressed on a real machine. Both are covered at the level
  the code can be tested at: A by the sweep's row-and-event assertions, B by the backstop tests plus the hook command
  driven against a receiver that is not there. **A reviewer running a real Core is what would close this line.**

- [ ] **Change anything owned by a parallel train.** `pty-core-link-server.ts` (#187), the core-link send path's
  live-event cursor advance (#244), and the Panel's `session-drive-register.ts` / `router.ts` (#242) are all untouched
  by design.

### Verification

`@actana/core` 1166 passed · `@actana/shared` 184 · `@actana/sdk` 236 · `@actana/cli` 421 · `@actana/panel` 1405
(2 load-related timeouts that pass in isolation, in suites this PR does not touch) · `pnpm typecheck` and `pnpm lint`
clean · commitlint green on every commit, not only the title. The root `scripts` suite has one pre-existing
environment failure on the machine this was built on (`core-launcher` picks up a real `/opt/actana` install).

mrfqcentic (approved):
Reviewed against #243 by reading the enforcing lines on `977f80a`, in a detached worktree off the PR head — not from the PR description or the checklist comment. Nothing in the repo was modified.

## The three parts are all here, and they are the ones the issue asked for

**1 — Boot reconciliation sweep. Met.** `sweepStrandedSessions` is called at `packages/core/src/core-entry.ts:194`, and the ordering the whole safety argument rests on holds at the file: it is after `configureCoreQueryStore` (:171), `configureCoreMutationStore` (:176) and `new CoreTaskWriter` (:183), and before `startHarnessHookReceiver` (:212), `new PtyCore(deps)` (:261) and `new PtyCoreLinkServer` (:506). So "the row claims a live process" and "the process belonged to a previous run" really are the same statement at that point — there is no Session of this run for it to take down. It reads `queryActiveTasks` (`packages/shared/src/core-query.ts:192`), which is `status IN ('running','needs-input')`, archived included, no project filter — a superset of the wedged row in the issue's evidence. Each settle goes through `CoreTaskWriter.mutate` (`core-session-sweep.ts:73`), so a `task:updated` lands in the event log a connected Panel re-renders from, and `disconnected` correctly raises no `session:finished`. I also checked the claim the issue makes about there being no existing Core-side sweep: `bootstrapCoreDb` (`packages/core/src/core-db-bootstrap.ts:47-81`) never calls `reconcileStaleSessionsOnBoot`, so the pre-existing doc line this PR replaced was describing something that did not exist. Good catch, correctly fixed.

**2 — The backstop is genuinely unconditional. Met.** `CoreSessionBackstop.sweepOnce` (`core-session-backstop.ts:141-154`) iterates rows the *database* still calls `running` — it consults no armed state, no tracked-subagent map and no deferred-finish timer. For a Session nothing ever reported a hook for, `lastActivity.get(taskId)` is `undefined` and the floor is the row's own `updatedAt` (`:149`), so it settles fifteen minutes after the row last moved. Units check out end to end: `updated_at` is written as `Date.now()` (`core-mutation-store.ts:143` → `core-mutations.ts:519-520`) and read through unchanged (`core-query.ts:274` → `CoreLinkTaskSnapshot.updatedAt: number`), so the `Math.max`/subtraction at `:149-150` is ms against ms — no silent `NaN` that would settle every row on the first sweep. The dead-process case is distinguished properly: `hasLivePty` false ⇒ `disconnected`, not `finished` (`:160-161`), and the live-PTY probe deliberately ignores shell PTYs (`pty-manager.ts:821`). `needs-input` is skipped (`:148`), which is right. The `core-session-backstop.test.ts` suite runs this against a real bootstrapped SQLite and a real event log rather than fakes, and its first case starts from exactly the state a lost `Stop` leaves.

**3 — The hook ack. Met, with one sub-claim overstated.** `hookCommand` (`harness-hooks.ts:112-125`) adds `-f`, `--retry 2 --retry-delay 1`, `-o /dev/null`, and appends a tab-separated record to `${AC_HOOK_MISS_LOG:-/dev/null}`, still ending in `|| true` so no hook can fail a turn. `HookDeliveryMonitor` drains that file at boot and every minute into `hook-delivery.missed` and a running `hook-delivery.missed-total` (`harness-hook-delivery.ts:167,174`). That is the issue's "a counter, a log line" — met. The third shape, "a sequence gap", is only half-built: the receiver mints `ack: <n>` (`harness-hook-receiver.ts:193-195`) but the client discards the body (`-o /dev/null`) and `acceptedCount()` (`:222`) has no non-test caller, so no reader can actually find a gap today. The criterion is satisfied by the other two shapes; please soften the claim in `harness-hook-receiver.ts:30-36` and the PR body rather than leaving it reading as delivered. The `harness-hook-delivery.test.ts` end-to-end cases — real `sh`, real `curl`, real receiver, Core-absent / wrong-bearer / unknown-task / no-miss-log — are the strongest evidence in the PR.

## Non-blocking, but worth fixing before the train merges

1. **`harness-hooks.ts:114` — a 404 is not a dropped hook.** With `-f`, `task-not-found` (`harness-hook-pipeline.ts:301-302`) exits curl 22 and writes a miss line, so a hook for a deleted or foreign row inflates `hook-delivery.missed-total` — the counter the issue asked for as a load signal. The OpenCode side already separates these (`harness-hooks-opencode.ts:130-137` breaks retry on `http-404`/`http-401`); the shell side cannot tell them apart at all. At minimum record the reason distinctly so a real drop rate stays legible.
2. **`harness-hook-delivery.ts:49-56` — `HookMiss.code` is documented as "curl's exit status" and logged as `curlExit`, but the OpenCode plugin writes `http-404` / `unknown` into the same column.** Rename it `reason`.
3. **The miss-line format is defined in three places** with no shared constant: the `printf` at `harness-hooks.ts:121-123`, the string concat in `opencodePluginSource`, and `parseMiss` at `harness-hook-delivery.ts:109-115`. Changing the record means editing two packages and hoping.
4. **`core-session-backstop.ts:156-177` decides a status outside the shared pipeline.** `core-harness-status.ts`'s header says the decisions live in `@actana/shared/harness-hook-pipeline` so both hosts behave identically, and the PTY-exit settle honours that by routing a synthetic event; this writes `writer.mutate` directly and re-implements the pipeline's bookkeeping (`clearSubagentActivity` / `noteTaskFinished`, cf. `finishQuietTask` at `harness-hook-pipeline.ts:322-326`). It is not *wrong* — `sweepOnce` is synchronous, so the re-read guard `finishQuietTask` has would be redundant — but it is a second place the finish rules now live.
5. **`harness-hooks.ts:114` — the retry budget rides the turn's hot path.** Worst case per hook goes 3s → ~11s, and that includes the per-tool-call `PostToolUse`, under precisely the Core-overload condition the issue describes. The docs argue this is only paid when the status is already wrong, which is fair; flagging it so it is a decision on the record rather than a surprise in a busy Core's transcript.
6. **`core-session-sweep.ts:18-21` — the "local-scope rows only" characterisation of the Panel's sweep is imprecise.** `findActiveLocalTasks` (`packages/panel/src/server/repositories/tasks.repo.ts:13-19`) has no scope predicate; it is scoped by being the *Panel's* database. True and load-bearing for a remote Core, but in loopback the Panel's `reconcileStaleSessionsOnBoot` (`packages/shared/src/schema-bootstrap.ts:236-242`) already stamps the same rows — silently, with no event — and may win the race, in which case this sweep finds nothing to announce. Worth a sentence, and that shared helper is now provably dead for the Core.
7. **`core-session-backstop.ts:30-36` — the self-heal claim is thinner than stated.** After a wrong `finished`, `PostToolUse` heals only from `needs-input` (`harness-hook-pipeline.ts:262-266`), so the card stays wrong until the next `UserPromptSubmit`, not until the next byte of work. The trade is still the right one; the paragraph should describe it accurately.
8. **`CONTEXT.md` gains no entry** for the quiet-Session backstop, the boot sweep, or the hook miss log, though all three are now domain vocabulary.

## The three declines

- **Leaving `ACTIVE_SUBAGENT_TTL_MS` at two hours — justified, does not block.** The TTL is tuned against real fan-out failures, and shortening it trades one false state for another. More to the point, the new backstop bounds the symptom the issue actually complained about: a row held by a lost `SubagentStop` now settles after fifteen minutes of silence without consulting the TTL, and `settle()` clears the tracking on the way out (`core-session-backstop.ts:168`). The issue asked for a bounded backstop, not a shorter TTL, and it got a tighter bound than the change it declined would have produced.
- **Not triaging #187's second symptom — correct call, does not block.** The issue asks for a triage *decision*; that is a maintainer action on #187, not a diff. Recording it as a loose end here is the right treatment, but somebody still owns doing it.
- **No live-Core reproduction — the one real gap, and still not blocking.** This is the weakest of the three. Repro path A is a daemon restart, which is exactly the boot path this PR changes, and the two things a live run would have exercised are the two things no test covers: the `core-entry.ts` wiring itself (no test asserts the sweep runs before the PTY core) and the throttled output-activity callback at `pty-manager.ts:688-694`, which the PR honestly declares untested. I verified the wiring by reading, and the failure modes are benign — `listActiveTasks` swallows a bad DB and returns `[]` (`core-query-store.ts:167-173`), and the sweep catches per row — so the worst case of a mistake there is a no-op boot, not a crash. It should still be closed by a maintainer with a real Core before this rides a release; approving the code, not that line item.

## Delivery

CI is red: **Unit Tests** fails on `packages/cli/src/__tests__/events-tail-cursor.test.ts > resumes where the last run stopped` (60s timeout). It is **not** this PR — the identical test fails the same way on `beta/0.3.1` (job 95682429575) and on `fix/panel-drive-register-client-id` (job 95687185294), and the CLI's in-process Core builds `PtyCoreLinkServer` directly and never runs `startCore`, so nothing in this diff is on its path. It is the live-event cursor territory of #244.

The consequence is worth stating plainly, though: the recursive run aborts at `@actana/cli`, so **`@actana/core`'s suite never ran in this CI job — every new test in this PR is unexecuted on CI**. Only `@actana/shared` got there, and its count does move with the PR (184 here vs 180 on the sibling branches — the four `queryActiveTasks` cases). I could not run the suite myself either: no `node_modules` in the read-only worktree and the machine is out of disk, and I was not going to write in the repository to get there. So the core-side evidence for this PR is the author's local run plus my reading of the code — enough for the code, short of what "verified" usually means here.

Also noting the PR is still a draft, and its base is the `fix/session-and-event-reliability` train branch at `da2af3e` — clean merge-base, no cross-train collisions, and the declared scope boundaries (`pty-core-link-server.ts`, the core-link cursor advance, the Panel's session-drive-register) are genuinely untouched.

## Verdict

**Approved.** All three parts are present in this one PR, the backstop is unconditional in the way the issue meant it and settles a Session whose process died with no hook ever arriving, and the boot sweep reconciles rows a crash left wedged and announces each one. The findings above are quality and accuracy items, none of them a defect in the behaviour #243 asked for. Two things belong to the maintainer rather than the author: a live-Core pass on repro path A, and the pre-existing `events-tail-cursor` flake that is currently hiding `@actana/core`'s suite from CI.

mrfqcentic (approved):
## Review: #247 against #243 — approve

Reviewed from a clean worktree at head `977f80a`, diffed against `origin/fix/session-and-event-reliability` (which is `da2af3e`, so the three-dot diff is the whole PR). Every claim below was checked against the enforcing line rather than the description, and the hook command was run for real against an absent Core.

### The three parts are all here, and they are wired

**1. Hook ack** — `harness-hooks.ts:112-125`. `-f` makes a non-2xx a failure, `--retry 2 --retry-delay 1` covers the transient one, and `[ "$s" = 0 ] || printf ... >> "${AC_HOOK_MISS_LOG:-/dev/null}" || true` records what did not land. I generated that exact string, ran `sh -n` on it (clean), and ran it against a dead port: exit `0`, one line `2026-…Z<TAB>t-…<TAB>Stop<TAB>7` in the miss log, and with the env var unset it wrote nowhere and still exited 0. `--data-binary @-` is fully buffered before the first attempt, so the retry re-sends the same body. Receiver side: `harness-hook-receiver.ts:193`; miss log drained at boot and every minute by `harness-hook-delivery.ts:74,132`, wired at `core-entry.ts:229`, env plumbed at `pty-manager.ts:576-582`. `harness-hooks-opencode.ts:97-140` mirrors it via `res.ok`.

**2. The backstop is genuinely unconditional.** `core-session-backstop.ts:141-154` reads the database, not a timer table: `lastActivity.get(taskId) ?? 0` falls back to the row's own `updatedAt`, so a Session for which **no hook ever arrived and nothing was ever armed** is settled after fifteen quiet minutes. `settle()` (`:156-177`) picks `finished` when a PTY is still live and `disconnected` when it is not — the right call for a process that died. Units check out end to end: `updated_at` is `Date.now()` ms (`core-mutation-store.ts:143`) and reaches the snapshot unchanged, and `core-session-backstop.test.ts` exercises it through the real SQLite rather than a stub. The `PostToolUse`-plus-spinner reasoning for "quiet" holds, and the two-hour-turn test proves it does not fire on live work. This also bounds the ~2h03m lost-`SubagentStop` wedge to fifteen minutes without touching the TTL.

**3. The boot sweep really is a boot sweep.** `core-session-sweep.ts:66-95`, called at `core-entry.ts:194` — after the stores and the writer (`:171,176,183`) and before the hook receiver (`:212`), the PTY core (`:261`) and the core-link server (`:506`). At that point no PTY of this process exists, so the "every such row is an orphan" argument is sound. It writes through `CoreTaskWriter`, so each settle appends `task:updated` and no spurious `session:finished` — asserted against the real event log. Its read (`core-query.ts:192`, `core-query-store.ts:167`) covers archived rows and every project, which is right. I confirmed the pre-existing doc claim that "the Core's DB bootstrap stamps `disconnected`" was false — `bootstrapCoreDb` never called `reconcileStaleSessionsOnBoot`, so this is new behaviour and the doc correction is warranted.

Incidental fix worth naming: `core-entry.ts:212-219` now forwards the URL's `hookEvent` fallback to `receiveHook`, which the previous two-arg lambda dropped on the Core.

### Non-blocking findings

1. **`harness-hook-receiver.ts:34`** claims "the Core's own log carries a monotonic sequence a reader can find a gap in". Nothing logs `ack`, and `acceptedCount()` (`:222`) has no production consumer — only tests. The client discards the body (`-o /dev/null`), and the OpenCode plugin checks only `res.ok`. #243's "counter, log line, or sequence gap" is satisfied by the miss log plus `hook-delivery.missed-total`, so this is an accuracy fix: either log the delivery number or soften the comment.
2. **`core-session-backstop.ts:156-177`** decides a status outside `@actana/shared/harness-hook-pipeline`, which `core-harness-status.ts`'s header names as the place these decisions live (the PTY-exit settle routes a synthetic event through it). It also re-does `finishQuietTask`'s (`harness-hook-pipeline.ts:322`) bookkeeping without its `status !== "running"` re-check. Safe today — `sweepOnce` reads and writes in one synchronous tick — but the guard is free and the divergence is worth a line of comment if it stays.
3. **The miss-line format is defined three times** — the `printf` at `harness-hooks.ts:121`, the string concat at `harness-hooks-opencode.ts:101`, and `parseMiss` at `harness-hook-delivery.ts:109` — with no shared constant. Relatedly, `HookMiss.code` is documented as "curl's exit status" (`:54`) and logged as `curlExit`, but the OpenCode path writes `http-404` / `unknown` into it; `reason` would be the honest name.
4. **No test covers the `core-entry` ordering** that everything in part 3 rests on. `core-files-wiring.test.ts` exists precisely because a correctly-built component can be wired wrongly; a small assertion that the sweep runs before the PTY core would close the one gap the unit suites structurally cannot see.
5. **`core-query.ts:192`** re-inlines the column list that `:144` keeps in a `columns` const ten lines above.
6. Minor: a second Core started against the same `--user-data-dir` would sweep the live Core's rows to `disconnected` at `:194` before failing to bind at `:506`. The Panel's `reconcileStaleSessionsOnBoot` has the same shape, so this is not a regression, but gating the sweep on a successful listen would remove it.

### On the three declines

- **Not shortening `ACTIVE_SUBAGENT_TTL_MS`** — correct, and consistent with `docs/harness-status-detection.md:115` ("these values were tuned … do not change them without a failure that says so"). The wedge it protects is now bounded at fifteen minutes by the new backstop, which does not consult the TTL. Does not block.
- **Not triaging #187's second symptom** — #243 asks for a triage decision, which is a maintainer call and not a code change; `pty-core-link-server.ts` is untouched. Does not block; the suggestion should be carried to #187 as an issue comment.
- **No live-Core end-to-end run** — the honest weak spot, and the one the checklist names correctly. Paths A and B are not reachable from the suite. Mitigating: CI's Core-image smoke boots a real Core and restarts it, so the new boot path runs on a real daemon without incident; what remains unproven live is the sweep's *effect* on a wedged row and the miss log filling on a real restart. Given the sweep's row-and-event assertions and the hook command driven through a real `sh` + `curl` (which I reproduced independently), this does not block approval — but it should be confirmed on a live 0.3.0 Core before the train ships, ideally against the database in the issue.

Draft status and merge order are the author's call; I have not touched either. CI is green across all checks on this head.

PR #248, opened 2026-08-18
fix(panel): reconcile pane events with tab interest in the Session drive register

### LINKED ISSUE #186: Panel session drive register: pane/tab granularity mismatch — duplicate watch reorders the queue, pane close drops tab interest (silent dual-write)

qcentic-dev:
## Summary

The Panel's Session drive register keys interest **per tab** — one client id, one
entry, whatever that tab has on screen — while the browser opened and closed
panes and sent a frame for each. A tab can hold two panes on one Session, and
there the two granularities disagreed in both directions (#186).

This reconciles them at the only layer that knows what a pane is: the panel-link
client now reference-counts this tab's panes per Session, exactly as it already
does for a Core's watchers. The first pane announces `watch`, the last pane
sends `drop`, and a reconnect re-announces the tab's interest once rather than
once per pane. `take` is exempt and always announces — it is the operator asking
for the keyboard here, not a pane arriving.

The register and router are unchanged in behaviour. They gain the sentence that
says what they arbitrate (tabs, never panes), and #242's parked ordering
question is answered rather than left: interest still goes to the tail on a
re-assert, and this ticket removed the re-assertions nobody asked for instead of
making re-assertion free.

**Stacked on `fix/session-and-event-reliability`**, not the train: it builds on
`e2fee65` (#242's client-id keying), which is where "the same tab" is defined.
Base is deliberately the feature branch; `Train rules` accepts it, and the whole
CI matrix is green on this head.

## Related issues

Closes #186
Builds on #242 · underlying model #140 · found in review of #184

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## The two failures, and where each is now caught

**1. Duplicate `watch` moved the keyboard to the back of the queue.** A second
pane on the same Session re-asserted interest the register already held; nothing
new was registered, but the send re-entered the queue at the tail. Now the
second pane sends nothing — `panel link · a tab's panes on one Session >
announces a Session once however many panes this tab opens on it`.

**2. Closing one of two panes dropped the tab's interest — silent dual write.**
The first pane's `drop` released the whole tab's interest with no refcount to
say the other pane still held it, so the register handed the Session on while
the surviving pane kept a writable surface and kept typing. Now the `drop` waits
for the last pane, and the browser's own copy of the answer waits with it —
`gives a Session back only when the last pane on it closes`, and `keeps the
tab's answer while a second pane on the Session is still open`.

## How was this tested?

- 9 new tests: 7 in `panel-link-client.test.ts` (the pane count, `take`'s
  exemption, and what a reconnect re-announces), 2 in
  `session-write-store.test.ts` (the shared per-Session answer across two panes).
  The store's fake bridge now counts panes like the real client, so it cannot
  pass a test the product fails.
- Verified red before the change: 5 of the 9 fail on `e2fee65` without it. The
  rest are regression guards that hold either way.
- `packages/panel`: 1441 tests, all green in isolation. Two in
  `live-read-path.test.ts` time out under the full local run and pass on their
  own — a known local flake on this machine, unrelated to this branch.
- `pnpm -r typecheck` clean; eslint clean on every changed file.
- `pnpm panel:e2e` green end to end, including the file-drop leg (2048 MiB
  across a Panel capped at a 256 MB heap, RSS +76 MiB).
- Every commit passes `commitlint` against this repo's own config, not just the
  title.

## Screenshots / recordings

Not applicable — no visual change. What changes is what a pane *stays*: writable
when this tab still holds the Session in another pane, and unmoved when a second
pane opens on it.

## Checklist

- [x] This PR targets the **open train** by way of the branch it is stacked on —
      base is `fix/session-and-event-reliability`, not `main`, and `Train rules`
      passes
- [x] I updated documentation (the register's and router's own module docs) where relevant
- [x] Dependent changes have been merged (#242 landed as `e2fee65`) — and this PR is a draft

### DIFF
--- a/packages/panel/src/components/views/TerminalPane.tsx
+++ b/packages/panel/src/components/views/TerminalPane.tsx
@@ -562,6 +562,13 @@ export function TerminalPane({
   // claim on the **Session lock**: this never leaves the Panel, and the lock is
   // an explicit gesture on the core-link (D6). First-come, so the second tab to
   // open the same Session follows the first rather than fighting it.
+  //
+  // Per pane on purpose, and counted per tab underneath (issue 186): a second
+  // pane of this tab on the same Session announces nothing — the tab's interest
+  // is already held, and re-asserting it would move the keyboard off whichever
+  // pane the operator was typing in — and this cleanup gives the Session back
+  // only when it is the last pane to go. What the mount/unmount pair says here
+  // is what this pane holds; what the tab holds is the sum of them.
   useEffect(() => {
     const coreId = descriptor.coreId;
     if (!coreId) return;
--- a/packages/panel/src/lib/__tests__/panel-link-client.test.ts
+++ b/packages/panel/src/lib/__tests__/panel-link-client.test.ts
@@ -685,3 +685,131 @@ describe("panel link · watching", () => {
     link.close();
   });
 });
+
+describe("panel link · a tab's panes on one Session", () => {
+  // The service keys drive interest per tab and the browser opens panes, so
+  // this is the seam where the two are reconciled (issue 186). One tab may hold
+  // two panes on one Session — a split view, the same Session opened twice —
+  // and what the service is told has to be the *tab's* standing, once.
+
+  it("announces a Session once however many panes this tab opens on it", () => {
+    const link = client();
+    const socket = live();
+
+    link.driveSession("core_a", "task_1");
+    link.driveSession("core_a", "task_1");
+
+    // Not a saving of bytes. A second `watch` re-asserts interest the service
+    // already holds, and re-asserting moves this tab to the tail of that
+    // Session's queue: the keyboard leaves the pane the operator is looking at,
+    // with no gesture of theirs to explain it.
+    expect(socket.drives("core_a")).toEqual([{ taskId: "task_1", want: "watch" }]);
+    link.close();
+  });
+
+  it("announces each Session in its own right", () => {
+    const link = client();
+    const socket = live();
+
+    link.driveSession("core_a", "task_1");
+    link.driveSession("core_a", "task_2");
+
+    expect(socket.drives("core_a")).toEqual([
+      { taskId: "task_1", want: "watch" },
+      { taskId: "task_2", want: "watch" },
+    ]);
+    link.close();
+  });
+
+  it("still asks for the keyboard in a pane on a Session it already watches", () => {
+    const link = client();
+    const socket = live();
+
+    link.driveSession("core_a", "task_1");
+    link.driveSession("core_a", "task_1", { take: true });
+
+    // `take` is the operator's gesture, not a pane arriving. Deduping it would
+    // be a button that does nothing in the one case it exists for.
+    expect(socket.drives("core_a")).toEqual([
+      { taskId: "task_1", want: "watch" },
+      { taskId: "task_1", want: "take" },
+    ]);
+    link.close();
+  });
+
+  it("gives a Session back only when the last pane on it closes", () => {
+    const link = client();
+    const socket = live();
+
+    link.driveSession("core_a", "task_1");
+    link.driveSession("core_a", "task_1");
+
+    expect(link.releaseSessionDrive("core_a", "task_1")).toBe(false);
+    // The tab still has the Session on screen. A `drop` here would have the
+    // service hand the drive to another tab while this one keeps a writable
+    // pane and keeps typing — two writers, and nothing on screen to say so.
+    expect(socket.drives("core_a")).toEqual([{ taskId: "task_1", want: "watch" }]);
+
+    expect(link.releaseSessionDrive("core_a", "task_1")).toBe(true);
+    expect(socket.drives("core_a")).toEqual([
+      { taskId: "task_1", want: "watch" },
+      { taskId: "task_1", want: "drop" },
+    ]);
+    link.close();
+  });
+
+  it("counts panes, not gestures: a take does not keep the Session held", () => {
+    const link = client();
+    const socket = live();
+
+    link.driveSession("core_a", "task_1");
+    link.driveSession("core_a", "task_1", { take: true });
+
+    expect(link.releaseSessionDrive("core_a", "task_1")).toBe(true);
+    expect(socket.drives("core_a").at(-1)).toEqual({ taskId: "task_1", want: "drop" });
+    link.close();
+  });
+
+  it("says nothing for a Session this tab has no pane on", () => {
+    const link = client();
+    const socket = live();
+
+    expect(link.releaseSessionDrive("core_a", "task_1")).toBe(false);
+
+    expect(socket.drives("core_a")).toEqual([]);
+    link.close();
+  });
+
+  it("re-announces a twice-held Session once on a reconnect", () => {
+    const link = client();
+    const first = live();
+    link.driveSession("core_a", "task_1");
+    link.driveSession("core_a", "task_1");
+
+    first.drop();
+    vi.advanceTimersByTime(20);
+    const second = live();
+
+    // What the new socket is owed is the tab's interest, which is one Session
+    // whatever it is showing it in. Re-announcing it per pane on every flap is
+    // the duplicate `watch` again, with a reconnect for a trigger.
+    expect(second.drives("core_a")).toEqual([{ taskId: "task_1", want: "watch" }]);
+    link.close();
+  });
+
+  it("re-announces nothing once every pane on the Session has gone", () => {
+    const link = client();
+    const first = live();
+    link.driveSession("core_a", "task_1");
+    link.driveSession("core_a", "task_1");
+    link.releaseSessionDrive("core_a", "task_1");
+    link.releaseSessionDrive("core_a", "task_1");
+
+    first.drop();
+    vi.advanceTimersByTime(20);
+    const second = live();
+
+    expect(second.drives("core_a")).toEqual([]);
+    link.close();
+  });
+});
--- a/packages/panel/src/lib/__tests__/session-write-store.test.ts
+++ b/packages/panel/src/lib/__tests__/session-write-store.test.ts
@@ -22,6 +22,10 @@ type Handlers = {
 function fakeBridge() {
   const handlers: Handlers = {};
   const drives: Array<{ taskId: string; take: boolean }> = [];
+  // The pane count the real link client keeps, and the answer the store asks it
+  // for (issue 186). A fake that said "last pane" to every release would pass a
+  // test the product fails, so this one counts.
+  const panes = new Map<string, number>();
   const bridge = {
     onSessionLock: (cb: NonNullable<Handlers["lock"]>) => {
       handlers.lock = cb;
@@ -33,9 +37,17 @@ function fakeBridge() {
     },
     driveSession: (_coreId: string, taskId: string, opts?: { take?: boolean }) => {
       drives.push({ taskId, take: opts?.take === true });
+      if (opts?.take !== true) panes.set(taskId, (panes.get(taskId) ?? 0) + 1);
     },
     releaseSessionDrive: (_coreId: string, taskId: string) => {
       drives.push({ taskId, take: false });
+      const open = panes.get(taskId) ?? 0;
+      if (open > 1) {
+        panes.set(taskId, open - 1);
+        return false;
+      }
+      panes.delete(taskId);
+      return true;
     },
   } as unknown as PanelBridge;
   __setPanelBridgeForTests(bridge);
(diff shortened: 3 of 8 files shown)


qcentic-dev:
## Acceptance criteria — issue #186

Issue #186 carries no explicit `- [ ]` list, so these are its stated failures,
its fix direction, and the boundaries the ticket was picked up under, taken one
by one.

- [x] **Failure 1 — a duplicate `watch` no longer moves the keyboard to the back
      of the queue.** A second pane of the same tab on the same Session sends
      nothing: the panel-link client counts this tab's panes per Session and
      only the first announces. Covered by `panel link · a tab's panes on one
      Session > announces a Session once however many panes this tab opens on
      it`, and by `announces each Session in its own right` (the count is
      per-Session, not per-Core).

- [x] **Failure 2 — closing one of two panes no longer drops the tab's interest
      (no silent dual write).** `drop` is sent only by the last pane to close,
      and the browser's own copy of the answer is reset only then, so the
      surviving pane and the register agree on who holds the Session. Covered by
      `gives a Session back only when the last pane on it closes` and `keeps the
      tab's answer while a second pane on the Session is still open` /
      `clears the tab's answer when the last pane on the Session closes`.

- [x] **Fix direction — dedupe the `watch` send client-side** (the issue's first
      option). Taken. The alternative, a position-preserving `want`, was
      deliberately not taken: it cannot fix Failure 2 — the service never learns
      a pane closed, only that a tab said `drop` — so a pane refcount is needed
      either way, and once it exists the duplicate send is gone at the source.
      Making a re-assert free would also undo the tail rule #242 chose for a
      reconnect. The reasoning is recorded in the register's module comment
      rather than left in a PR.

- [x] **Fix direction — add a release refcount so a tab's interest is dropped
      only when the last pane on that Session releases it.** `PanelLinkClient`
      keeps the count, and `releaseSessionDrive` returns whether that was the
      last pane so the write-state store can ask rather than keep a second count
      that could disagree.

- [x] **"Pane-level events and tab-level interest agree on who holds what."**
      One `watch` and one `drop` per tab per Session now reach the service,
      whatever that tab has on screen. The aggregation lives in the browser
      because panes have no name on the wire, and the register and router now
      say so in their own docs.

- [x] **Built on #242's client-id keying, not reverting to socket identity.**
      `session-drive-register.ts` keeps `DriveClientId` and its ordered lists
      exactly as `e2fee65` left them; no change to the keying, and no change to
      what `want` / `release` do. The ordering question #242 explicitly parked
      here is answered in the register's comment: interest still goes to the
      tail on a re-assert.

- [x] **Boundary — issue 208's territory untouched.** No file under
      `packages/cli` is in this branch; the events-tail-cursor test is not
      touched. `git diff --stat e2fee65..HEAD` is eight files, all under
      `packages/panel`.

### Verification

- [x] 9 tests added (7 client, 2 store); 5 of them verified red on `e2fee65`
      without the fix, the rest are regression guards.
- [x] `packages/panel` suite: 1441 tests green. Two in `live-read-path.test.ts`
      time out under the full local run and pass in isolation — a known local
      flake on this machine, present on the base commit and unrelated to this
      branch.
- [x] `pnpm -r typecheck` clean; eslint clean on all eight changed files.
- [x] `pnpm panel:e2e` green end to end, file-drop leg included — 2048 MiB
      across a Panel capped at a 256 MB heap, RSS +76 MiB. No ENOSPC; the disk
      had ~47 GB free throughout.
- [x] Every commit on the branch passes this repo's `commitlint.config.mjs`, as
      does the PR title.
- [x] CI on this head is green in full — `Conventions`, `Train rules`, `Lint`,
      `Typecheck`, `Unit Tests`, `E2E — Panel service seam`, both installer
      legs, both tarball smokes, `Secret Scan`, `Dependency Audit`.

### Not done, deliberately

- [ ] **Not marked ready for review.** Left as a draft, as asked.
- [ ] **Not merged, and no train branch touched.** Base is
      `fix/session-and-event-reliability`; `main` and `beta/0.3.1` are untouched.
      `Train rules` accepts the stacked base, so nothing is red on account of it.

mrfqcentic (approved):
**Approved.** Reviewed against issue #186 by reading the enforcing lines rather than the PR description. Every acceptance criterion is met; what I found is four non-blocking nits and one factual slip in the PR body.

## Built on #242's client-id keying, with no socket identity reintroduced

`session-drive-register.ts` and `router.ts` contain **no executable change** in this branch — both diffs are comments only. The keying is still exactly what `e2fee65` left:

- `session-drive-register.ts:71` — `export type DriveClientId = string`
- `session-drive-register.ts:92` — `private readonly interest = new Map<string, DriveClientId[]>()`
- `session-drive-register.ts:126-138` — `want()` untouched; `:151-161` — `release()` untouched

And every call site still passes the id, never the socket object:

- `router.ts:409-410` — `state.drives.release(taskId, session.clientId)` / `state.drives.want(taskId, session.clientId, { take: … })`
- `router.ts:368` — `driverOf(taskId) === session.clientId`
- `router.ts:446` — the asking tab's answer, read back out of the register by client id
- `router.ts:476` — `if (this.sessionsByClientId.get(session.clientId) !== session) return;`, the guard that keeps a retired predecessor from releasing a live tab's drives

`e2fee65`'s own commit message said the pane-versus-tab granularity "is untouched and stays with #186". This is that follow-on, and it builds on top rather than around.

## The reconciliation holds

**Two panes on one Session.** `panel-link-client.ts:365` (`if (panes > 0) return`) means only the first pane announces; `:382-388` decrements and returns `false` until the last pane, so only the last sends `drop`. `session-write-store.ts:158` gates the local reset on that `true` — and it has to, because `session-write-access.ts:100` returns `writable: true` for `drive: "none"`. That is genuinely the visible half of failure 2, not a cosmetic fix.

One consequence worth recording: a deduped second pane receives no `drive`/`lock` frame, since `router.ts:441-451` answers only when `want !== "drop"`. That is correct here only because the store keys its entry per Session rather than per pane, so the second pane reads the answer the first one already stored. Same reason the store fix works; worth keeping in mind if that entry is ever re-keyed.

**Tab reload.** `claimTabClientId()` (`panel-link-client.ts:807`) restores the id from `sessionStorage` while `drivenSessions` starts empty in the fresh heap, so the remounting panes re-announce the tab's interest exactly once instead of once per pane. bfcache restore is the mirror case — JS state and the counts survive, and `resendSessionDrives()` re-announces once per Session. Both consistent.

`take` correctly stays exempt (`:352-356`), and it is reachable only from `TerminalPane.tsx:1572`, gated on `readOnlyReason === "driven-in-another-tab"` — which requires `drive === "following"`, which requires a prior counted `watch`. So the exemption can't leave an uncounted pane in practice today.

| # | Criterion (from #186) | Verdict |
|---|---|---|
| 1 | Duplicate `watch` no longer moves the keyboard to the back of the queue | **Met** — `panel-link-client.ts:365`; pinned by *"announces a Session once however many panes this tab opens on it"* |
| 2 | Closing one of two panes no longer drops the tab's interest (no silent dual write) | **Met** — `:382-388` plus `session-write-store.ts:158`; pinned on both the wire and the UI answer |
| 3 | Fix direction — dedupe the `watch` send client-side | **Met** — and the reasoning for preferring it over a position-preserving `want` is recorded in the module comment, not left in the PR |
| 4 | Fix direction — release refcount, last pane only | **Met** — one count, in the client, asked rather than mirrored |
| 5 | "Pane-level events and tab-level interest agree on who holds what" | **Met** — one `watch` and one `drop` per tab per Session |
| 6 | Built on #242's client-id keying, no socket identity | **Met** — see above; register and router are comment-only |
| 7 | Boundary — #208's territory untouched | **Met** — `git diff --name-only` is 8 files, all under `packages/panel`, none under `packages/cli` |

## Do the tests pin the behaviour, or only its shape?

They pin the two failures, and the PR is honest about the rest. Reverting the fix to `e2fee65` leaves red:

- *"announces a Session once however many panes this tab opens on it"* — old code sends two `watch` frames
- *"gives a Session back only when the last pane on it closes"* — old code drops on the first release
- *"counts panes, not gestures"* — old code returns `undefined`, not `true`
- *"keeps the tab's answer while a second pane on the Session is still open"* — old store resets to `none`, so `.drive` is not `"driving"`

Green either way (regression guards, correctly described as such in the checklist): *"announces each Session in its own right"*, *"still asks for the keyboard…"*, *"re-announces a twice-held Session once on a reconnect"* (the old `Set` also collapsed to one entry), *"re-announces nothing once every pane has gone"*, and *"clears the tab's answer when the last pane closes"*. *"says nothing for a Session this tab has no pane on"* fails on revert only because of the changed return type, not because of behaviour.

So: 4 genuinely behaviour-pinning, 1 type-pinning, 5 shape guards. That is a fair ratio for a 23-line fix, and the two failures in the issue are each pinned on both sides of the seam.

## Non-blocking nits

1. **`releaseSessionDrive` returning a bare `boolean`** — `panel-bridge.ts:149`, `panel-link-client.ts:380`. The name says "release"; the boolean means "that was the last pane". Only the doc comment carries it, and the one call site reads `!== true`. A named result (`"released" | "still-held"`) would put it in the signature.

2. **`session-write-store.ts:158` conflates "no bridge" with "not the last pane"** — `getPanelBridge()?.releaseSessionDrive(…) !== true` skips the local reset when the bridge is null, where it previously reset unconditionally. Benign in practice (`getPanelBridge()` is null only under SSR, where `entries` is never populated), but the two cases are worth splitting so the intent is visible.

3. **`take` returns before counting** — `panel-link-client.ts:352-356`. Correct for every path that exists today, but it means a `take` on a Session this tab has not counted would leave an announced interest that `resendSessionDrives()` never replays and `releaseSessionDrive` never drops. Unreachable now; a latent trap on a public bridge method. One line of comment at the early return would keep the next caller out of it.

4. **A duplicate `watch` still survives on the queue path, but harmlessly** — a pane mounting before the socket opens increments the count *and* queues a `watch`; the open handler runs `resendSessionDrives()` and then flushes `queued` (`:571-573`), so two identical `watch` frames go out. This is **pre-existing** — the old `Set` produced exactly the same pair — and it is benign, because `want()` at `:132-136` filters then appends: re-asserting an entry that is already at the tail is positionally a no-op, and the two frames are contiguous on one socket. Not this PR's to fix, and I mention it only so it isn't mistaken later for failure 1 returning.

## One factual slip

The PR body and the checklist both say "9 tests added (7 client, 2 store)". It is **10** — there are 8 `it()` blocks in the new `panel link · a tab's panes on one Session` describe, plus 2 in the store. The "5 of them verified red on `e2fee65`" claim matches what I get by inspection, so only the total is off.

## Delivery

CI is green across the full matrix on this head, including `Train rules` for the stacked base. Commits are Conventional and within `commitlint`'s 120/132 limits. The change is 23 executable lines of production code against 117 lines of documentation — heavy, but the prose is answering #242's parked ordering question and recording why the position-preserving `want` was declined, which is the kind of thing that is expensive to reconstruct later. The one place I would push back is that the *meaning* of the `drop` frame narrows here (from "a pane closed" to "that tab's last pane closed") and that reinterpretation lives only in module comments; per CONTRIBUTING's "Architectural decisions", an appended clause on ADR 0024 would be its documented home. Not a blocker for this PR.

Leaving the draft status alone — that is the author's call, and the checklist says it is deliberate.

PR #249, opened 2026-08-18
test(cli): settle the tip hunt before the live pushes in events-tail-cursor

### LINKED ISSUE #208: Flaky: events-tail-cursor.test.ts times out in CI under runner load

qcentic-dev:
## Summary

`events-tail-cursor.test.ts > resumes where the last run stopped, and does not replay it` times out in CI
under runner load (#208). It is a **race in the test's synchronisation, not a budget problem** — the run
hangs forever and vitest's 60s deadline is what stops it. The fix waits for the condition the test's own
pushes depend on, and adds two assertions rather than removing any.

### The race

A first `events tail` has no stored cursor, so it starts at the end of the log the way `tail -f` does —
and the end of the log has to be *found*, because the Core publishes it only through the replay
(`event-tip.ts`). With three events of history:

1. `subscribe(0)` → the Core sends #1–#3 and an `eventsReplayed` that is a **receipt for what it sent**,
   not a statement of the tip.
2. The run therefore asks again — `subscribe(3)` — and only a marker closing an **empty** tail means "this
   is the end". Until then, printing stays off and every delivered event is counted as history.

The old wait was `await waitFor(() => log.tailReads >= 1, "the first tail never subscribed")` — satisfied by
the **first** of those two reads. The two events the test then appends had a window, one round trip wide, in
which to land inside the *second* tail. Appended there they are history: the tip tracker counts them, the
command deliberately prints neither, `--limit 2` is never reached, and the run follows a Core with nothing
left to say. Forever.

### Why the evidence in the issue reads as load but is a hang

The issue reasoned that 61.7s / 61.7s / 63.1s against a 60s budget looked like an overrun rather than a
deadlock, since "a deadlock would sit at the deadline every time". A hang here sits at the deadline too —
60s plus teardown of a real `wss://` Core and a temp dir is exactly 61–63s. The tell is the other direction:
**the whole file passes in 1.2s** when the race is won, so there is no 60s of work to overrun.

Measured on this branch (5 rounds, instrumented `readEventTail`):

| | read #1 | old wait released | read #2 | margin |
|---|---|---|---|---|
| round 0 | 6.6ms | 11.5ms | 7.6ms | −3.9ms |
| round 1 | 4.2ms | 10.7ms | 5.1ms | −5.6ms |
| round 2 | 3.9ms | 10.1ms | 5.4ms | −4.7ms |
| round 3 | 2.7ms | 9.7ms | 3.4ms | −6.3ms |
| round 4 | 2.7ms | 11.0ms | 3.1ms | −7.9ms |

The second read lands 0.4–1.5ms after the first; the old wait released on the 10ms poll tick behind it. The
test was passing on **4–8ms of scheduler jitter** — which is why it fails on a loaded runner and only there,
why re-runs go green, and why the four recorded strikes are spread across unrelated branches.

### The fix

Wait for the run to say it has found the end of the log — `the Core's log ends at #N`, on stderr under
`--verbose`. That is the same signal the sibling test (`prints none of a history longer than the Core
replays in one tail`) already waits for, and it is the exact precondition the two pushes need: once it is
emitted, `printing` is on and nothing can be swallowed as history.

**Nothing is weakened.** The 60s budget stays, and every assertion is unchanged — `[4, 5]` from the first
run, the on-disk cursor at `5`, `[6]` and only `[6]` from the second. Two assertions are **added**:

- the tip settled at `#3` **exactly** — the run walked the whole history and stopped at its end, which is
  what makes the `[4, 5]` below it mean anything;
- nothing was printed while the end of the log was being found.

A lost race now fails **in 12ms**, naming the tip it settled on (`expected '…ends at #5…' to contain
'…ends at #3;'`), instead of burning 60s to report only that time ran out.

### On #244's semantics

The ordering #244 landed — the live-event cursor advancing only *behind* a send that landed, `subscribe`'s
replay moving its cursor on the same evidence — is what this test exists to guard, and it is untouched here.
No production code changes; `packages/core/src/__tests__/core-link-event-cursor-advance.test.ts` (4 tests)
and the rest of `packages/core` (1170 tests) are green on this branch.

### Worth flagging: one red cli test hides core and panel entirely

CI's **Unit Tests** job runs `pnpm test` → `pnpm native:node && vitest run && pnpm -r test`. `pnpm -r` bails
on the first failing package, and the workspace's topological order puts `cli` second:

```
$ pnpm -r --workspace-concurrency=1 exec pwd
packages/sdk
packages/cli      ← this flake aborts the run here
packages/shared
packages/core
packages/panel
```

So this one hung test does not merely add a red check — it **suppresses every `core`, `shared` and `panel`
result behind it**. A PR that touches only the Core or the Panel can go red with the failure sitting in a
package it never touched and with no report at all from the packages it did. That is part of why #208 has
been expensive out of proportion to its size, and it is worth separating in triage: the flake is fixed here,
the bail ordering is not (a `--no-bail` on the recursive run, or splitting the job per package, would be its
own change).

## Related issues

Closes #208
Refs #244

## Type of change

- [x] 🐛 Bug fix (non-breaking change fixing an issue) — test-only
- [ ] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## How was this tested?

- [x] Manually verified (described below)

Local, on this branch (`env -u AC_*` to keep this machine's Core out of the runs):

- `packages/cli/src/__tests__/events-tail-cursor.test.ts` — **15 consecutive clean runs**, 4/4 each.
- Under load — 30 busy loops on 10 cores — **5/5 green**, file time inflated to 4.7–12.8s (of a 60s budget).
- The **pre-fix** file under that same load: **4 failures in 6 runs**, each `Test timed out in 60000ms`,
  file durations 65.9–68.3s — the CI signature, reproduced locally.
- Deterministic proof of the mechanism: forcing the two events into the tip hunt's second tail leaves the
  pre-fix run printing `[]` and never resolving; the fixed test catches that state in 12ms with a sentence.
- `packages/cli` 421 passed / 3 skipped · `packages/core` 1170 passed · `packages/sdk` 236 passed / 5 skipped
  · `packages/shared` 184 passed · root `vitest run` 474 passed · `tsc --noEmit` and `eslint` clean.
- `packages/panel`: 1429/1431 in a full local run; the handful that fail vary run to run and **all pass in
  isolation** (a known local full-run timeout artefact — this diff touches no panel file).

## Checklist

- [ ] This PR targets the open train (`beta/x.y.z`), not `main` — **deliberately not**: it targets
  `fix/session-and-event-reliability`, the integration branch #244 landed on, since the flake is about
  exactly those semantics.
- [x] My PR title and every commit in the branch follow Conventional Commits (verified with the repo's own
  `commitlint.config.mjs`), and the branch name follows the branching convention
- [x] I commented hard-to-understand areas / added docs where needed — the wait carries the whole argument
- [ ] I updated documentation (README, docs/, changelog entry) where relevant — n/a, test-only
- [x] Dependent changes have been merged (#245 is on the base branch); this PR is a draft

---

**CI on this branch:** all 17 checks green on the first attempt, Unit Tests included (2m3s) — run
32137677337. Five train/publish jobs skipped as expected on a non-train PR.

### DIFF
--- a/packages/cli/src/__tests__/events-tail-cursor.test.ts
+++ b/packages/cli/src/__tests__/events-tail-cursor.test.ts
@@ -137,13 +137,48 @@ describe("actana events tail, across a Core restart", () => {
     log.push("session:finished");
 
     const firstRun: string[] = [];
-    const first = fixture.run(["events", "tail", "--json", "--limit", "2"], {
+    const notices: string[] = [];
+    const first = fixture.run(["events", "tail", "--json", "--limit", "2", "--verbose"], {
       connect: connectCore,
       onOut: (line) => firstRun.push(line),
+      onErr: (line) => notices.push(line),
     });
-    await waitFor(() => log.tailReads >= 1, "the first tail never subscribed");
+
+    // Wait for the tip hunt to have *finished*, not merely to have started.
+    //
+    // This is the flake in #208, and it is a race rather than a slow runner:
+    // one `subscribe` is not where a first run settles. With three events
+    // already in the log the Core answers the first one with a full tail and an
+    // `eventsReplayed` that is only a receipt, so the run asks again from #3 and
+    // waits for a second marker to close an *empty* tail (`event-tip.ts`). The
+    // old wait here was `log.tailReads >= 1` — satisfied by the *first* of those
+    // reads. The two events below then had a window, one round trip wide, in
+    // which to land before the second `readEventTail` ran: appended inside it
+    // they are history, counted by the tip tracker and deliberately never
+    // printed, so `--limit 2` is never reached and the run follows a Core that
+    // has nothing left to say. That is a hang, not an overrun — which is why CI
+    // saw 61.7s against a 60s budget three times over, and why the file passes
+    // in 1.2s whenever the race is won.
+    //
+    // The notice below is the run stating that it has found the end of the log
+    // and switched printing on — the condition these two pushes actually need,
+    // published on stderr by the command itself, and the same signal the
+    // history-longer-than-a-tail test waits for. Reaching it takes a round trip
+    // more under load and none of the assertions less.
+    const endOfLog = (): string | undefined =>
+      notices.find((line) => line.includes("the Core's log ends at #"));
+    await waitFor(() => endOfLog() !== undefined, "the first tail never found the end of the log");
+    // #3, and stated: the run walked the whole history and stopped at its end.
+    // A tip settled anywhere else is the failure the pushes below were racing —
+    // now an assertion about the cursor rather than a timeout with no evidence.
+    expect(
+      endOfLog(),
+      "the first tail settled somewhere other than the end of the history",
+    ).toContain("the Core's log ends at #3;");
     // Nothing from the tail above may be printed: with no stored cursor, a tail
     // starts at the end of the log the way `tail -f` does.
+    expect(firstRun, "history was printed while the end of the log was being found").toEqual([]);
+
     log.push("task:question");
     log.push("pty:exit");
 


qcentic-dev:
## Acceptance criteria — issue #208

#208 carries no `## Acceptance criteria` heading; it is filed as an evidence-only ticket. The criteria below
are its own stated asks — the failure it records, the things it says the fix must not be, and the three
"plausible directions" it leaves to whoever picks it up — each marked checked or explicitly not, with why.

- [x] **The named test no longer times out under runner load.** `resumes where the last run stopped, and
  does not replay it` — 15 consecutive clean runs, then 5/5 green under 30 busy loops on 10 cores (file
  time 4.7–12.8s of its 60s budget). The pre-fix file under that same load failed 4 runs in 6, each with
  `Test timed out in 60000ms` at 65.9–68.3s — the CI signature from the issue's table, reproduced locally.

- [x] **The cause is identified, not guessed at.** The issue's stated purpose was to record the evidence
  "before somebody guesses". The wait `log.tailReads >= 1` released after the *first* of the tip hunt's two
  tail reads, leaving the test's two pushes a one-round-trip window in which to land inside the *second*
  tail — where they are history, are counted rather than printed, and `--limit 2` is never reached. Proven
  deterministically: forcing the pushes into that window leaves the pre-fix run printing `[]` and never
  resolving.

- [x] **It is a hang, not an overrun — and the issue's timing argument is corrected.** #208 read 61.7s /
  61.7s / 63.1s against a 60s budget as a slow runner overshooting. A hang produces the same numbers (60s
  deadline + teardown of a real `wss://` Core and a temp dir). The tell is that the whole file passes in
  1.2s, so there is no 60s of work to overrun. Measured margin on the old wait: **4–8ms** of scheduler
  jitter.

- [x] **The test still asserts what it asserted.** 60s budget unchanged; `[4, 5]` from the first run, the
  on-disk cursor at `5`, and `[6]` and only `[6]` from the second are all unchanged. Two assertions added:
  the tip settled at `#3` exactly, and nothing was printed while the end of the log was being found.

- [x] **Direction 2 — "find the actual wait", and report where the time went.** Taken. The wait is now the
  run's own `the Core's log ends at #N` notice, the published moment printing switches on. A lost race now
  fails in **12ms** naming the tip it settled on, instead of 60s of silence.

- [ ] **Direction 1 — raise this one test's budget.** Deliberately not taken. The issue itself notes it
  "hides a real slowdown if one ever appears", and it would not have fixed this: the run hangs, so any
  budget expires.

- [ ] **Direction 3 — split the test so a Core is booted once.** Not applicable and not needed. This test
  boots one Core and runs the CLI twice; the two-Core boot the issue describes belongs to its neighbour,
  `delivers every event exactly once…`. No structural split was required once the wait was correct.

- [x] **#244's semantics are untouched.** No production code changed. The cursor-advances-behind-the-send
  ordering is what this suite guards, and it still does:
  `packages/core/src/__tests__/core-link-event-cursor-advance.test.ts` (4 tests) and all of `packages/core`
  (1170) green on this branch.

- [x] **The flake is not hidden behind an unrelated diff.** #208 objected to touching the test inside
  #164's diff. This branch is one commit, one file, that test only.

- [x] **The two follow-up comments are addressed.** The PR #223 head that needed three attempts and the
  #234 promotion PR blocked on attempt 1 are the same race; nothing suggests a second cause.

- [x] **Cross-package artefact reported.** CI's Unit Tests job runs `pnpm -r test`, which bails on the first
  failing package, and the workspace order is `sdk → cli → shared → core → panel` (verified with
  `pnpm -r --workspace-concurrency=1 exec pwd`). This one hung cli test therefore suppresses every `core`,
  `shared` and `panel` result behind it — a PR touching only the Core or Panel goes red with no report from
  the packages it actually changed. Written up in the PR description; the bail ordering itself is left as a
  separate change (`--no-bail`, or a job split), not smuggled in here.

**PR:** #249 — https://github.com/actana/control/pull/249 (draft, base `fix/session-and-event-reliability`)

mrfqcentic (approved):
**Approve.** Reviewed the diff against `origin/fix/session-and-event-reliability` (one commit, `cbba41a`, one file, +37/−2) by reading the enforcing lines rather than the description.

## The thing that mattered most: nothing was weakened

The failure mode a flake fix has to be cleared of is a longer budget or a retry that lets a real ordering violation pass silently. This is neither. Assertion by assertion, before → after:

| Enforcing line | Before | After |
|---|---|---|
| `}, 60_000)` (`events-tail-cursor.test.ts:206`) | 60s | **unchanged** |
| synchronisation | `waitFor(() => log.tailReads >= 1, …)` | `waitFor(() => endOfLog() !== undefined, …)` — releases on the **second** tail read, strictly later |
| `expect(idsOf(firstResult.out)).toEqual([4, 5])` (:186) | present | **unchanged** |
| `expect(existsSync(dir), "no cursor was written").toBe(true)` (:190) | present | **unchanged** |
| `expect(readFileSync(…).trim()).toBe("5")` (:192) | present | **unchanged** |
| `expect(idsOf(second.out)).toEqual([6])` (:203) | present | **unchanged** |
| `expect(endOfLog(), …).toContain("the Core's log ends at #3;")` (:174-177) | — | **added** |
| `expect(firstRun, …).toEqual([])` (:180) | — | **added** |

No retry wrapper, no `toEqual` loosened to `toContain` (the one `toContain` is a *new* assertion on a string, and the trailing `;` is what stops `#30` satisfying `#3`), no assertion deleted. The only deletion is the weak wait itself.

**Would the test still fail if the cursor advanced optimistically again? Yes — in three independent places.** A cursor moved past events never printed fails `toBe("5")` at :192 and `toEqual([6])` at :203. A tip read off a partial replay receipt (the #205 defect) fails `toContain("the Core's log ends at #3;")` at :177 and would put printed history into `firstRun`, failing :180 and :186. The two additions make that case fail *earlier and by name* instead of downstream. The test is strictly harder to pass than before.

## The wait is the right one

`packages/cli/src/events-command.ts:213-214`:

```ts
if (end === null) return;          // capped replay — another marker is coming
deps.verbose(`the Core's log ends at #${end}; following from there`);
printing = true;
```

Adjacent statements in the same synchronous `onEventsReplayed` handler, no `await` between them, and `waitFor` polls on a 10ms timer — so the notice cannot be observed before `printing === true`. That substring is emitted at exactly one site repo-wide; the neighbouring verbose lines (`the Core sent N event(s) up to #X…`, `caught up to #N`) don't contain it. So the wait is precisely the precondition the two pushes need, and it cannot release early. Waiting on the command's own published notice is also the established idiom in this file — the sibling `prints none of a history longer than the Core replays in one tail` (:220-243) already does the same thing, down to the failure message.

`--verbose` is stderr-only: `args.verbose` is set at `cli-args.ts:170-171` and never read anywhere else; both entry points derive the sink from argv (`actana-cli-entry.ts:31,43`, `cli-harness.ts:187,200-206`), and `new FileCursorStorage(dir, deps.verbose)` passes the sink unconditionally either way. The cursor path is byte-identical with the flag on.

## #244's semantics

Untouched. `git diff --name-only` returns one path, a test file; no production code changed. `packages/core/src/__tests__/core-link-event-cursor-advance.test.ts` shows zero diff lines, and the first test in this same file — `delivers every event exactly once, in order, over a dropped connection`, with `toEqual([1, 2, 3, 4, 5, 6])` — is untouched (all four hunks sit at :140-181). CI on `cbba41a` is green on all 17 checks including Unit Tests (2m3s, run 32137677337), which means `pnpm -r` also reached and passed `core`, `shared` and `panel`.

## Non-blocking notes

1. `packages/cli/src/__tests__/events-tail-cursor.test.ts:169` — the notice literal now appears four times in this file in two shapes (`find` here, `some` at :232). A small file-level `endOfLogNotice(notices)` helper would collapse all four and match the file's own idiom (`idsOf`, `coreOn`, `fillPastTheCap`). Cosmetic.
2. The PR text says a lost race "fails in **12ms**". That figure is from the forced-race experiment; in the shipped test the pushes are after the wait, so the realistic new failure is `waitFor` expiring at its 10s default (`in-process-core.ts:337`) with `the first tail never found the end of the log`. Still a named condition in 10s instead of 60s of silence — the claim is right in substance, optimistic in the number.
3. Issue #208's direction 2 asked to "report where the time went". What ships reports *what the run settled on*, not timing; the millisecond table came from instrumentation that isn't in the diff. The intent is met, the literal ask isn't — fine, since the cause turned out to be a hang rather than slow work.
4. Coverage shifted slightly: with the pushes now strictly after the tip is found, this test no longer exercises "events landing *during* the tip hunt are counted, not printed". The sibling test covers that deliberately (1500 events, `expect(printed).toEqual([])`), so there's no net loss.
5. Base branch: `CONTRIBUTING.md:174` says open against the open `beta/x.y.z`. Targeting `fix/session-and-event-reliability` is disclosed in the checklist, is what `Train rules` permits (the rule it enforces is that only a train may target `main`), and is right here since #244/#245 landed on that branch. It does mean this still needs to reach the train with its base.

## On the `packages/panel` report

The reading is fair, with one caveat. This diff is a single CLI test file — it cannot reach a panel test, and `packages/panel` doesn't import it. The stronger evidence is that CI is green on this branch: because the CLI flake is fixed, `pnpm -r test` no longer bails at `cli` and actually ran `panel`, which passed. The mechanism is plausible too — 142 panel test files, several with 5s-60s budgets, running in vitest's parallel pool against a loaded local machine.

The caveat is the word "known": nothing tracks it. There's no issue for it and no mention in the docs, so it currently lives as lore. #208's own lesson is exactly that timings which "look like load" can be a hang, and the only reason that was caught is that someone wrote the evidence down first. 2-4 varying files deserve the same treatment in their own ticket rather than a parenthesis in a PR description. Not this PR's job, and not a reason to hold it.

---

Approving on content. Note it's still a draft — I'm not marking it ready; that's yours.

PR #250, opened 2026-08-18
fix: settle dead Sessions, close the event hole, and unwedge the drive register

qcentic-dev:
This batch closes the reliability holes that let a Session card lie about what a
harness is doing, and that let the Panel's keyboard arbitration disagree with
itself. A Core-owned Session now settles even when the hook that should have
settled it never landed; a live event is no longer skipped past a socket that
silently refused it; and the Session drive register survives a tab reload and
counts panes and tabs at the granularity each layer actually knows about.
Together they replace three classes of "wedged until a human edits the row"
with states the system reaches on its own.

### What lands

- **#243 — settle Core-owned Sessions.** Harness hooks are now acked with a
  delivery number instead of a fire-and-forget POST ending in `|| true`, and
  what still cannot be delivered is recorded to a miss log the Core drains into
  its own log. Two settles back that up: a boot sweep that stamps
  `disconnected` on rows a restart stranded (no PTY of this process exists yet,
  so any such row is an orphan), and `CoreSessionBackstop`, which arms nothing
  and settles rows that have gone quiet for fifteen minutes with neither hook
  nor PTY output.
- **#244 — close the live-event hole.** `pushLiveEvents` advanced
  `conn.lastSentEventId` past events it only attempted; a silent `ws` refusal
  therefore made a skipped one-shot lifecycle event (`session:finished`,
  `session:lockChanged`) permanently unreachable by any replay. The cursor now
  advances behind the send, delivery becomes at-least-once and in order, and a
  refusal is logged as `core-link.event.undelivered` naming the event.
- **#242 — key the Session drive register by client id.** Interest was keyed by
  the browser's socket object, so a reloaded tab arrived as a stranger and
  queued behind the ghost it had just replaced. Interest is now an ordered list
  of client ids, with a tab id minted into `sessionStorage` and presented on
  the upgrade query — claimed rather than read, so a duplicated tab mints its
  own.
- **#186 — reconcile pane events with tab interest.** The client sent one frame
  per pane while the service keys interest per tab. A second pane on one
  Session silently moved the keyboard off the pane being typed in; closing one
  of two panes released the whole tab's interest and produced two writers with
  no conflict signalled. The client now counts panes — first announces, last
  releases — and the store resets its shared write answer only on the last
  pane. `take` stays exempt and announces every time.
- **#208 — stabilise the `events-tail-cursor` CLI test.** The flake was a race,
  not a slow runner: the run waited on `log.tailReads >= 1` and pushed its two
  live events into what was still history. The wait is now the run's own "the
  Core's log ends at #N" notice. Nothing is relaxed — same 60s budget, same
  expectations — and a lost race fails in 12ms naming the tip instead of timing
  out at 60s.

### Worth a close look

- **#208 pins the semantics #244 introduced.** The stabilised test waits on the
  tip notice and asserts the settle point (#3 exactly, nothing printed while it
  was being found). That is now the guard on #244's replay contract — the
  `eventsReplayed` receipt reporting the event actually reached rather than the
  end of a tail only partly written. Read the two together; changing either
  side's notion of where a tail ends will move this test.
- **#186 builds on the client-id keying #242 introduced.** The pane counting is
  aggregated per client id, so it only holds because #242 made "the same tab"
  survive a new socket. #242 explicitly left pane-vs-tab granularity to #186,
  and #186 answers the ordering question #242 parked: a re-assert goes to the
  tail, so the duplicate sends are gone and the re-asserts that remain are the
  ones a reconnect owes the register.
- **#244's cursor rewind is at-least-once.** A late `ws` delivery-callback
  failure rewinds the cursor, so the Panel can now see a duplicate event where
  it previously could only see a gap. Confirm the Panel's consumers are
  idempotent on the lifecycle events specifically.
- **#243's backstop trade, stated in the commit and repeated here:** a harness
  silent for longer than fifteen minutes gets a card reading `finished` while
  it works on — and it is not only a card. The backstop writes `finished`
  through `CoreTaskWriter`, which appends a `session:finished` event, so the
  operator gets the full finish treatment in the Panel: a ding, a toast, and a
  persisted entry in the notification list. Nothing is killed and the next
  `UserPromptSubmit` restores `running`, but a harness that is still working
  will have announced itself as done, and the fifteen minutes is a hardcoded
  constant — the window is the tunable worth arguing about.
- **#243 changes the hook command contract** — curl now fails on non-2xx and
  retries transient errors, with `--retry-connrefused` deliberately absent to
  avoid raising the required curl version. `-o /dev/null` matters: Claude Code
  reads hook stdout as control JSON.

Base is `beta/0.3.1`, not `main`. Squash merge — the squash takes this title,
but its body is the five commit messages, not this description. The commit
messages are the batch's CHANGELOG source material; read those, not this note,
for the record of what landed.


qcentic-adm:
`beta/0.3.1` was created by hand from `main` without the version stamp ADR 0023 D3 requires, so all six manifests still read `0.3.0` and `Train rules` failed here with one error per manifest.

The missing cut commit is now on the train: [`54e7646`](https://github.com/actana/control/commit/54e764667298ab31b6ed49ecd02cd0f7fcd91cca) — `chore(release): cut beta/0.3.1`, six manifests, written by the same script `promote.yml`'s `next-train` job uses and verified by its own loop. Nothing was merged, approved, or retargeted, and `main` was not touched.

A `pull_request` run replays the merge commit pinned in its event payload, so re-running the failed job would have re-tested the old base. Closing and reopening this pull request to get a run against the stamped train — no change to the branch or its commits.

qcentic-adm:
`Train rules` is green — the run against the stamped train ([32140683401](https://github.com/actana/control/actions/runs/32140683401)) reports all six manifests at `0.3.1` and `beta/0.3.1` still fast-forwardable into `main`. Every other check on this pull request is green too; it is now blocked only on review, which is where it should be.

mrfqcentic (approved):
## Beta gate review — the batch as a batch

Reviewed as the train PR, not as five tickets. Worktree on `e0935bc`, diffed whole against `beta/0.3.1` in one go. **Approving.**

---

### 1. Coherence of the combination

**#208 ↔ #244 — the replay contract.** The coupling is real and the note describes it correctly. `event-tip.ts` is built on one load-bearing sentence: *"a marker is treated as a receipt for what was sent, never as the tip."* #244's `handleSubscribe` now breaks the replay loop on the first refusal and reports `lastSent` — the event the socket actually took — in `eventsReplayed`. That strengthens the receipt semantics rather than moving them, so the tip hunt's "ask again until a marker closes an empty tail" loop still terminates on the same evidence. `events-tail-cursor.test.ts:177` (`the Core's log ends at #3;`) is now the guard on that, and it is a genuine one: any change to where a tail is deemed to end moves this assertion.

Worth recording for the next reader: #244 also removes a latent hazard for the *other* direction. `DurableCoreClient.deliverEventsReplayed` advances its persisted cursor to the marker. Before #244 that marker could name the end of a tail that was only partly written, permanently pushing the client past events it never received — the exact loss #244 is about, arriving through the replay path rather than the live one. Both halves now move on the same evidence.

**#242 ↔ #186 — client-id keying under pane counting.** Also holds. #186's aggregation is per `coreId → taskId → paneCount` in `panel-link-client.ts`, and that is only meaningful because #242 made the identity on the wire survive a new socket; a socket-keyed register would have reset the count's counterpart on the service side on every reload. Checked the three ways they could have disagreed:

- `driveSession({take:true})` returns *before* incrementing (`panel-link-client.ts:352-356`). Verified this is safe against every caller: the only `take` path is `TerminalPane.tsx:1577` → `takeSessionDrive`, a click inside a pane that the mount effect at `:576` already counted. `take` is never a first announce, so the exemption cannot leave an uncounted pane.
- `resendSessionDrives` iterates `taskIds.keys()` — one `watch` per Session on reconnect, not one per pane. Strictly fewer re-asserts than before #186, and #242's tail rule still applies to the one that remains. The two tickets' answers to the parked ordering question agree.
- `releaseSessionDrive` now returns "was that the last pane", and `session-write-store.ts:158` gates the local `drive: "none"` reset on it. The count lives in exactly one place; no second copy to drift.

**#244's at-least-once duplicate — the note asks for this to be confirmed, so I confirmed it.** Traced both lifecycle events:

- `session:finished` — `use-session-finish-notifications.tsx` dedups on `(coreId, sessionId, eventId)` (`dedupKey`, `:43`) behind a capped LRU. Idempotent.
- `session:lockChanged` — `SessionLockRegister.applyEvent` is state-replacing and `set()` short-circuits on an unchanged pair, so the *state* is idempotent. But `consumeSelfEcho` (`session-lock-register.ts:255`) is a **one-shot queue consumer**: a second delivery of an echo whose token was already consumed falls through to `this.set(taskId, {held: false, locked: payload.locked})` and would render this Panel's own held Session read-only.

That is not reachable, and the reason is worth naming since nothing in the diff says it: `DurableCoreClient.deliverEvent` drops `event.eventId <= this.lastEventId` before any listener sees it (`durable-core-client.ts:231`), and `core-link-manager.ts` gives the router no other client type. Every duplicate #244 can produce is absorbed at the transport, above the register. The idempotency the note asks about is satisfied — by construction rather than by each consumer, which is the stronger arrangement, but it does mean the guarantee lives one layer away from the consumer that depends on it.

---

### 2. Nothing lost in integration

Mechanically verified rather than eyeballed. All five landed on the branch as consecutive squashes with nothing in between, and each squash's diff is **patch-id-identical** to the diff of the source branch it came from:

| PR | source branch | patch-id match |
|---|---|---|
| #247 | `fix/core-session-settle` | identical |
| #246 | `fix/panel-drive-register-client-id` | identical |
| #245 | `fix/core-link-cursor-advance` | identical |
| #248 | `fix/panel-drive-register-granularity` | identical |
| #249 | `test/events-tail-cursor-flake` | identical |

On the three surfaces called out:

- **`session-drive-register.ts`** — touched by #246 then #248. #248's contribution here is *comments only*; every enforcing line (`interest` as `Map<string, DriveClientId[]>`, `want`'s tail append, `release`, `releaseAll`, `settle`) is #246's and unmodified. Nothing #242 introduced was reverted.
- **`router.ts`** — same pair. #248's contribution is again comment-only (the "one frame per tab per Session" block on `driveSession`). #242's whole mechanism survives intact: `sessionsByClientId`, the take-the-id-then-retire-the-predecessor ordering in `attach`, the `sessionsByClientId.get(session.clientId) !== session` guard at the head of `releaseDrives`, the ownership check in `forget`, and the answer read back out of the register rather than inferred from `change.gained`.
- **core-link send path** — `pty-core-link-server.ts` is touched by #245 alone, so there was no second writer to lose anything to. `sendResult`/`deliverEvent`/`rewindEventCursor` and the `adaptWs` callback forwarding are all present and wired to both the live push and the replay.

Merges clean onto the current `beta/0.3.1` tip (`54e7646`, the version cut). The branch touches no manifest, so it neither conflicts with nor reverts that commit.

---

### 3. The release note

**Honest.** Checked the body against the enforcing lines rather than the descriptions, and every specific claim holds:

- `-f`, `--retry 2 --retry-delay 1`, `-o /dev/null`, no `--retry-connrefused`, and the chain still ending in `|| true` — `harness-hooks.ts:114-123`, matching the note verbatim.
- Fifteen minutes, once a minute — `QUIET_SETTLE_MS = 15 * 60 * 1000`, `SWEEP_INTERVAL_MS = 60 * 1000`. Live PTY → `finished`, none → `disconnected`; `needs-input` never in scope (`core-session-backstop.ts:148`). The trade is stated as the code behaves.
- The boot sweep writes `disconnected` through `CoreTaskWriter`, before the PTY core / hook receiver / link server exist (`core-entry.ts`), which is what makes "any such row is an orphan" true.
- #208 relaxes nothing: the diff removes exactly two lines, the 60s budget and `[4,5]`/cursor-at-5/`[6]` are untouched, and two assertions are added.

The title — `fix: settle dead Sessions, close the event hole, and unwedge the drive register` — covers #243, #244 and #242+#186 in the CHANGELOG's own register, and reads as one release rather than a list. Per ADR 0023 D5 this subject is the CHANGELOG's source material for the batch, and it earns that.

Two small omissions from the body, noted rather than requested:

1. **The hook-event-name rider is real and unlisted.** #243 also fixes `core-entry.ts` dropping the receiver's URL-derived event name on the floor — `(taskId, payload) => receiveHook(taskId, payload)`, so a payload without `hook_event_name` was `ignored` in production while the tests passed the fallback through. It is in the commit message and covered by `harness-hooks.test.ts`, but the body's three-item "What lands" for #243 doesn't mention it, and it is a production behaviour fix, not a cleanup.
2. The OpenCode plugin getting the same `res.ok`/retry/miss-line treatment, and the CONTEXT.md / ADR 0024 D9 updates, are likewise commit-message-only.

Neither makes the note misleading; both make it slightly narrower than the batch.

---

### 4. Test suite, run here

Full `pnpm test` in the worktree, Node 24.19.0, pnpm 11.1.2, clean `pnpm install --frozen-lockfile`.

`pnpm test` exits 1 at the **root** `vitest run`, on one failure that is neither this branch's nor a flake:

```
FAIL scripts/__tests__/core-launcher.test.mjs > execs the bundled node on the bundled actana CLI
  Expected: "/tmp/actana-launcher-test-E38WWJ/install/app"
  Received: "/opt/actana/app"
Test Files  1 failed | 15 passed (16)
     Tests  1 failed | 473 passed (474)
```

This box has a real `/opt/actana/app` installed, so the launcher resolves its hard-coded default instead of the test's temp install root. The file is untouched by the branch and CI's Unit Tests job is green. Environmental.

Because that root run fails first, `pnpm -r test` never executes — the known bail artefact, one step earlier than expected. So I ran every package myself rather than relying on the recursive run:

| package | result |
|---|---|
| `@actana/shared` | 184 passed (14 files) |
| `@actana/sdk` | 236 passed, 5 skipped (21 files) |
| `@actana/cli` | 421 passed, 3 skipped (31 files) — includes `events-tail-cursor` |
| `@actana/core` | **1170 passed** (63 files) |
| `@actana/panel` | 1439 passed, 2 failed (142 files) |

The two panel failures are the known artefact: `live-read-path.test.ts` and `terminals-in-browser.test.ts`, both `Test timed out in 5000ms`, both under full-suite load. Re-ran them in isolation alongside all four new panel suites — `session-drive-reload`, `panel-link-tab-id`, `panel-link-client`, `session-write-store` — and got **6 files, 99 tests, all green**. Confirmed artefact, not a regression.

`pnpm typecheck` clean. `pnpm lint` clean — 0 errors, 50 warnings, all pre-existing (the two in touched files, `WebSocketServer` in `pty-core-link-server.ts` and `activeDeps`/`userDataDir` in `pty-manager.ts`, are present on `beta/0.3.1` too).

Net: **3450 tests pass across the five packages**; the only two failures anywhere are the two known artefacts, and both are accounted for.

---

### One observation, no change requested

`sendResult`'s doc says the `ws` `error` event "is what puts `core-link.connection.error` in the log for every fire-and-forget frame." That is now true only for frames that get *past* the `readyState` guard. A fire-and-forget frame refused at the guard — a PTY chunk on a `CLOSING` socket, the common case — is dropped silently: `send()` discards the `SendOutcome`, and `core-link.event.undelivered` is only reached from `deliverEvent`. Previously that write reached `ws` and produced an error line. Almost certainly the right trade (the socket is dying and the close handler logs), and PTY loss is the recoverable kind as the comment says — but the comment claims a log line the guard now pre-empts, and the next reader chasing a lost PTY chunk will believe it.

---

**Verdict: sound as a batch. The two named couplings hold under the combination, nothing was dropped or reverted across the five squashes, and the title and body honestly describe what lands. Approved for the train.**

qcentic-adm (approved):
## Second review — org admin identity, release-minded, independent of the moderator's

Reviewed `da2af3e..e0935bc` in a detached worktree on the head. Ran the suite, exercised the generated hook command against a live receiver, and re-ran the #208 file under load. **Approving.** Findings below are notes, not blockers; the one I'd most like addressed is (1), and it is a wording fix in the release note rather than a code change.

---

### 1. Is this safe to ship as 0.3.1?

Yes, and the strongest argument is one the description doesn't make: **this branch changes no published package's source.** Only `@actana/cli` and `@actana/sdk` are non-private; `packages/sdk` is untouched and the only `packages/cli` change is a test file. Everything else lands in `core`, `panel` and `shared`, all `private: true`. The npm surface of 0.3.1 is byte-identical to 0.3.0 as far as this train is concerned. For a patch release that is the right shape.

Behaviour a user depends on, changed here — all intended, all bounded:

- **The hook command text is rewritten** into every operator's `settings.json` / OpenCode plugin at the next spawn. I ran the emitted command end to end against a fake receiver. It holds up exactly as `harness-hooks.ts:112-125` claims: exit 0 on every path (200 / 503→200 / 500 / 401 / connection refused / timeout / unwritable miss-log path / no `AC_HOOK_MISS_LOG` at all), stdout empty in every case — which is the `-o /dev/null` promise that matters for Claude Code's control JSON — `sh -n` clean, and the POST body correctly re-sent on each retry despite coming from `--data-binary @-`. Worst case measured at **11.1s** against a black-holing receiver, matching the commit's "about eleven seconds" claim precisely. `--retry-connrefused` is genuinely absent: refused connections fail in 8ms with one attempt.
- **The boot sweep restamps existing rows.** First start after upgrade, every `running`/`needs-input` row — archived included (`core-query.ts:queryActiveTasks`) — becomes `disconnected`. That is a visible, one-time change to data users are looking at. It is the fix, and `disconnected` is the honest status, but it is worth knowing it lands on upgrade rather than on the next incident.
- **A second pane on one Session no longer moves the keyboard**, and closing one of two panes no longer releases the tab's interest (#186). If anyone had learned "open a second view to grab the keyboard here", that gesture is now `take` and only `take`.
- **New on-disk artefact**: `<user-data-dir>/hook-misses.log`. **New browser storage key**: `actana.panel-link.tab` in `sessionStorage`. Neither appears in "What lands".

**The one thing I think is understated.** `core-session-backstop.ts:161-169` writes `finished` through `CoreTaskWriter`, which appends a `session:finished` event (`core-task-writer.ts:138-150`), which the Panel turns into a **ding, a toast, and a persisted notification entry** (`use-session-finish-notifications.tsx:239-282`, kind match at `:456`). The PR body and `docs/harness-status-detection.md` both price the trade as "a card reading `finished`". It is not only a card — a harness that works silently for fifteen minutes pings the operator that it finished. That is a larger claim to make wrongly than a status colour, and the fifteen minutes is a hardcoded constant with no operator lever (`quietMs` is injectable, but only from tests). I'd state the notification in the trade paragraph. I don't think it blocks: exposure narrows to a `running` row with a genuinely silent harness — `needs-input` is out of scope, and both the permission-prompt and hooks-need-review paths land there — and the next `UserPromptSubmit` undoes it.

### 2. The interactions, checked as combinations

**#208 pinning #244.** Real, and the test genuinely does the pinning. The wait is now the run's own `the Core's log ends at #N` notice and it asserts the tip settled at `#3` exactly with nothing printed while it was being found (`events-tail-cursor.test.ts:145-180`). That assertion sits directly on top of #244's changed `eventsReplayed` contract — `handleSubscribe` now reports `lastSent`, the event the socket actually took, rather than the end of the tail (`pty-core-link-server.ts:1691-1702`). If either side's notion of where a tail ends moves, this test moves. I re-ran the file **6× under 20 busy loops on 10 cores: 6/6 green.** The claim reproduces. It does couple the test to a `--verbose` stderr string; that is a fair price and the commit says so.

**#186 building on #242's keying.** The combination is sound, and the ordering that makes it sound is the interesting part. `attach` assigns `sessionsByClientId` **before** retiring the predecessor (`router.ts:216-224`), and `releaseDrives` bails when the id no longer resolves to it (`router.ts:462`) — so the reload's ghost gives back nothing, and #186's pane count (aggregated per client id in `panel-link-client.ts:351-393`) only holds because of that. I checked the replay loop in `handleSubscribe` is synchronous, so a `ws` delivery callback cannot fire mid-loop and be clobbered by the `conn.lastSentEventId = lastSent` after it. It cannot.

**#244's duplicates.** The description asks reviewers to "confirm the Panel's consumers are idempotent on the lifecycle events specifically". The honest answer is that **they are not idempotent — they are protected by two cursor guards**, and it's worth writing down which: `durable-core-client.ts:230-235` (`if (event.eventId <= this.lastEventId) return`) sits between the Core and everything in the Panel server, and `panel-link-client.ts:740` repeats it per Core for each browser tab. The consumer that would actually misbehave is `session-lock-register.ts:204-217` — `consumeSelfEcho` **pops a queue**, so a second delivery of one `session:lockChanged` would fall through and render this Panel's own held Session read-only. It has no local guard. The protection is real and pre-existing, so this is not a defect in #244; but "confirm the consumers are idempotent" invites the wrong check, and a future path that feeds that register un-deduped events reintroduces it.

I also chased the theoretical case where event *N*'s delivery callback fails after *N+1* has landed, which would let the client's `<=` guard swallow the re-sent *N* permanently. On a single ordered socket it isn't reachable: if *N*'s write fails, *N+1* either fails too or is refused synchronously by the `readyState` check, and `rewindEventCursor`'s monotonic-backwards guard (`pty-core-link-server.ts:1000-1004`) settles both on the earliest hole. Noting it as considered rather than as a finding.

**An interaction the batch doesn't name.** `--retry 2` makes hook delivery at-least-once **at the receiver**, not just on the wire. In my timeout run the receiver logged `HITS=3` — a Core that is alive but slower than `-m 3` receives and processes all three POSTs and just loses the race to answer. Most of the receiver path is an idempotent status set, but `subagent-activity.ts:123-141` is a counter on the no-`agent_id` path (`entry.anonCount += 1`), so a duplicated `SubagentStart` from an older Claude build inflates it and holds `hasActiveSubagents`. Self-limiting — the 2h TTL bounds it, and #243's own backstop now bounds it to fifteen minutes — so it is a note, not a defect. But it belongs in the "worth a close look" list next to the event-cursor one, because it is the same at-most-once → at-least-once trade being made a second time in the same PR without being called that.

### 3. The release note

**Title:** honest and accurate. `fix: settle dead Sessions, close the event hole, and unwedge the drive register` — 79 chars against `header-max-length` 120, type in the enum, no full stop. It names three of the five and characterises the batch fairly.

**One correction to the premise, though:** the body is *not* what lands. This repo is configured `squash_merge_commit_message = COMMIT_MESSAGES` (with `squash_merge_commit_title = COMMIT_OR_PR_TITLE`, which with five commits resolves to the PR title). So the squash carries the **five commit messages**, not this description. The body's own sentence is careful and correct — it claims only that *the title* is what the CHANGELOG will carry — but anyone reading this as "title and body are the entry" would be wrong about half of it. I simulated the concatenation: 342 lines, **none over `body-max-line-length` 132**, so the squash will lint clean. The commit messages are excellent and are, if anything, a better CHANGELOG entry than the description.

**Completeness of the description as a description:** honest throughout, and unusually good at naming its own trades. Gaps, all minor: the four doc files aren't mentioned; the new `hook-misses.log` and `sessionStorage` key aren't mentioned; and "acked with a delivery number" is true of the wire but nothing consumes the number — `acceptedCount()` has no non-test caller, curl discards the body with `-o /dev/null`, and the OpenCode plugin only checks `res.ok`. What actually made the ack load-bearing is `-f` turning a non-2xx into a failure, which the body does say.

### 4. Version stamp

Confirmed. `54e7646 chore(release): cut beta/0.3.1` stamps exactly six manifests and nothing else, and all six read `0.3.1` on the branch tip: root `actana-control`, `@actana/cli`, `@actana/core`, `@actana/panel`, `@actana/sdk`, `@actana/shared`. **Nothing in this branch touches a `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml` or `CHANGELOG.md`** — the diff is 36 files, all source, tests and docs. The branch forks from `da2af3e` and so does not contain the stamp; that is fine and expected, it merges on top of it. PR reports `MERGEABLE`.

### 5. Test suite

Run per package rather than via `pnpm -r`, precisely so `packages/cli` sorting first couldn't hide `packages/core`:

| | result |
|---|---|
| root (`scripts/**`) | 473 passed, **1 failed** — environmental, see below |
| `packages/cli` | 421 passed, 3 skipped |
| `packages/core` | **1170 passed** |
| `packages/panel` | 1437 passed, **4 failed** — the known load flakes |
| `packages/sdk` | 236 passed, 5 skipped |
| `packages/shared` | 184 passed |

Both failures run down to the environment, not the branch:

- **`packages/panel` ×4** — `live-read-path.test.ts:437` and three in `terminals-in-browser.test.ts` (`:580`, `:671`, `:690`), all `Test timed out in 5000ms`. Exactly the class you flagged. Both files pass in isolation (14/14 and 13/13). Neither file is touched by this branch, and the full run's import time was 269s against 16s isolated — the box was saturated. One of them (`stops pulling a pty across the link once the last pane lets it go`) is pane-shaped enough that I checked it specifically rather than assuming; it is clean in isolation.
- **root ×1** — `scripts/__tests__/core-launcher.test.mjs:71` expects the install root and got `/opt/actana/app`. That is my container's ambient `AC_APP_PATH`, which `scripts/lib/core-tarball.mjs:118` honours via `${AC_APP_PATH:-…}`. Passes with the variable unset. The branch touches no `scripts/`, `bin/` or `deploy/`. Worth knowing that this makes the root `pnpm test` red *before* `pnpm -r test` runs at all on any box with that variable set.

Net: **green on the merits.**

---

### Smaller things, no action required to merge

- `harness-hook-delivery.ts:74-83` — the `MAX_MISS_LOG_BYTES` guard doesn't do what its comment says. `fs.readFileSync(missLogPath, "utf8").slice(0, MAX_MISS_LOG_BYTES)` reads the **whole** file into memory and then slices, so "must not turn a diagnostic into a memory event" isn't achieved. Low severity — the file only grows while the Core is down or refusing, at roughly 60 bytes a miss — but a `read` of the first N bytes into a buffer would cost nothing and match the intent.
- `session-write-store.ts:158` — `if (getPanelBridge()?.releaseSessionDrive(coreId, taskId) !== true) return;` also returns early when there is **no bridge** (`getPanelBridge()` is null under SSR, `panel-bridge.ts:276-280`), where the old code reset the local entry unconditionally. Almost certainly unreachable in practice, since a pane that mounted had a bridge; flagging only because the null case and the not-last-pane case now share one branch and mean different things.
- `pushLiveEvents` logs `core-link.event.undelivered` on every refusal, so a socket stuck in `CLOSING` emits one warn per 500ms poll tick per pending event until the connection is reaped. Bounded by the connection's life, but a first-refusal-only line would read better in a log an operator is grepping.

Approving. Nice work — the evidence in the commit messages held up everywhere I checked it, which is not the usual outcome.

qcentic-dev:
Both text corrections are in — description only, no commits, no code, no branch or base changes.

**1. The backstop trade now states the notification.** The bullet said "a card reading `finished`" and stopped there. It now says the backstop writes `finished` through `CoreTaskWriter`, which appends a `session:finished` event, so the operator gets the full finish treatment in the Panel — a ding, a toast, and a persisted entry in the notification list — and that a harness still working will have announced itself as done. Also noted that the fifteen minutes is a hardcoded constant. You were right that this is a larger claim to make wrongly than a status colour, and the body was pricing it as the smaller one.

**2. The CHANGELOG sentence is corrected.** The closing line now reads that the squash takes this title but its body is the five commit messages, not this description, and that the commit messages are the batch's CHANGELOG source material. Your reading of the repo config is what the line now reflects; anyone taking the description for the entry would have been wrong about half of it, and the note no longer leaves that open.

Everything else in the body is untouched — the per-ticket "What lands" lines, all issue numbers, and the whole "Worth a close look" section stand as they were.

**The three non-blocking code findings are being carried forward as follow-up issues, not fixed in this train:**

- `harness-hook-delivery.ts:74-83` — the `MAX_MISS_LOG_BYTES` guard does `readFileSync(...).slice(...)`, reading the whole file before slicing, so it doesn't achieve what its comment claims.
- `session-write-store.ts:158` — `if (getPanelBridge()?.releaseSessionDrive(...) !== true) return;` now also swallows the no-bridge case, where the old code reset the local entry unconditionally; the null case and the not-last-pane case share one branch and mean different things.
- `core-link.event.undelivered` in `pushLiveEvents` — one warn per 500ms poll tick per pending event against a socket stuck in `CLOSING`, where a first-refusal-only line would read better.

None of them changes what ships in 0.3.1, and keeping them out keeps this train's diff identical to what both of you reviewed. They'll be filed against the next milestone.

Issue #224, opened 2026-08-13
CoreFilesErrorCode is defined twice: the Core owns the checked copy, the SDK keeps a mirror nothing checks


qcentic-dev:
Follow-up to PR #223, filed from the review threads there: raised as a judgement call in [@mrfqcentic's review](https://github.com/actana/control/pull/223#pullrequestreview-4930692505) (Standards note 3), deferred with reasons in the reply, and the deferral accepted in [@qcentic-adm's re-review](https://github.com/actana/control/pull/223#pullrequestreview-4930835918) (§4 note 3) **on the condition that this ticket exist** — *"an intent recorded only in a PR reply is the thing that evaporates."* Nothing here is a regression and nothing is broken today: the two copies agree, member for member.

## What is duplicated

The same 22-member union, verbatim and in the same order, in two packages:

- `packages/core/src/core-files-routes.ts:97`
- `packages/sdk/src/core-files-http.ts:63`

`diff` of the two blocks is empty.

The two copies are not equally load-bearing, and that asymmetry is the interesting half:

- **The Core's copy is checked.** `type Refusal = { status: number; code: CoreFilesErrorCode; message: string }` (`core-files-routes.ts:121`) types every refusal the routes send, so a code that is not in the union is a compile error. The Core also has a *third* list, `TarRefusalCode` in `packages/core/src/files-tar.ts:41` (9 members) — but that one is pinned by the compiler, because `const code: CoreFilesErrorCode = err instanceof TarError ? err.code : "write-failed"` (`core-files-routes.ts:574`) forces it to stay a subset. Inside the package, a copy cannot drift.
- **The SDK's copy is not checked, deliberately.** Every use site widens it: `CoreFilesErrorCode | string` at `core-files-http.ts:116`, `:118`, `:141`, `:156`, `:158`. That widening is *right* — a client that meets a Core newer than itself must survive a code it has never heard of — but it means the SDK's union is documentation and autocomplete, and a member missing from it is a type error nowhere.

So: the one place the copy could drift is the one place nothing would notice.

`docs/external-api.md` states the same codes in prose (the 400/409/507 table, and the `root-entry-path` row). That copy is for humans and is fine as a copy; it is mentioned only so a fix does not forget it exists.

## Why this is worth a ticket and not a shrug

[...]

Issue #202, opened 2026-08-12
docs(release skill): SKILL.md still describes the pre-ADR-0023 tag trigger and macos-release placement


qcentic-dev:
`.agents/skills/release/SKILL.md` still describes the pre-ADR-0023 release
mechanism. The drift predates #200 and was surfaced by its review; #200 added
the `NPM_TOKEN` precondition and the burn-once rule to the file and
deliberately left the rest alone, because unwinding it there would be a second,
unrelated change.

Three places are wrong about how a release starts:

- **`:39`** — `git tag v0.1.0 && git push origin v0.1.0` as the trigger. ADR
  0023 D40 removed the tag trigger; `release.yml` has no `on: push: tags:` and
  is reached by `workflow_call` from `promote.yml`.
- **`:44-54`** — the whole "The tag starts the release" section, including
  `environment: macos-release` on `tarball-macos`. D15 moved that environment
  to `promote.yml`'s `pause` job, which is the first job in that workflow.
  `release.yml` no longer references the environment at all except in comments
  explaining its removal.
- **The frontmatter description**, which carries the same tag-trigger framing.

Why it matters beyond tidiness: after #200 the file is authoritative about
`NPM_TOKEN` and about npm's burn-once irreversibility while still being wrong
about the mechanism above it — and the new paragraph at `:82` opens "npm is
stricter than **that**", where "that" refers to the stale account. A reader who
trusts the top of the file will push a tag and wait for a run that never
starts.

Fixing it means rewriting the trigger section around `promote.yml`'s dispatch
and pause, and re-pointing the `docs/ci-cd.md` links that go with it.

Refs #200, ADR 0023 D15 and D40.

qcentic-adm:
Shipped in the **0.3.2** release train. It reached `main` as **6cd15ca** (`chore: 0.3.2 improvement sweep — docs, contracts and repo hygiene`), promoted through **#262**; the work came in on the train as **#252**. Verified against `.agents/skills/release/SKILL.md` at `origin/main`.

The three places this issue names as wrong, and where they now stand:

[...]

PR #251, opened 2026-08-18
fix(core): keep a separator where sanitizeInitialInput drops a line break

### LINKED ISSUE #193: sanitizeInitialInput drops newlines without a separator

qcentic-dev:
Closes #193.

## The bug

`sanitizeInitialInput()` strips C0/DEL bytes from a programmatic starting
prompt and rejoins the survivors with `.join("")`. A line break is `0x0A`, so
it failed the `code >= 32` test and was dropped like any other control byte —
with nothing left in its place. The word ending one line was welded onto the
word opening the next:

```
review the diff
then open a PR
```

reached the agent as `review the diffthen open a PR`, with `diff` and `then`
fused into one token. Every programmatic starting prompt (Ship / Sync /
Create-PR) is sanitised through this one function, so any multi-line prompt
was delivered mangled.

## The fix

Whitespace control bytes — tab, LF, VT, FF, CR — now collapse to a single
space before the C0/DEL filter runs. They still cannot pass through verbatim
(a bare CR/LF would submit the prompt early), but they *separate* the words
either side of them, so they leave a separator behind rather than vanishing.

Every other C0/DEL byte is still removed with nothing in its place — those are
keybinding attempts, never text — so the filter's original purpose is
unchanged. A *run* of whitespace controls collapses to one space, so CRLF and
blank lines do not pad the prompt, and the existing `.trim()` keeps a leading
or trailing break from leaving stray whitespace.

The change is small, but every harness start prompt goes through it, so the
behaviour is pinned rather than left to the next reader.

## Tests

`sanitizeInitialInput` is exported for the suite the way `disposePty` and
`planLaunchPortKillTargets` already are. New cases in
`packages/core/src/__tests__/pty-manager.test.ts`:

- a line break keeps a separator (the issue's own example)
- a leading or trailing line break leaves no stray whitespace
- a run of repeated newlines collapses to one separator
- carriage returns, `\r\n` and `\n\r` behave like line feeds
- a prompt that is only newlines returns `undefined`
- nothing to send (`undefined`, empty, all-spaces) returns `undefined`
- tab and form feed separate too
- ESC / `\x03` / DEL are still removed **without** a separator
- ordinary text, its internal double spaces and its non-ASCII are untouched

## Verification

- `vitest run --root packages/core` — 63 files, 1179 tests, all green
- `tsc --noEmit` on `packages/core` — clean
- `eslint` on both changed files — no new problems (3 pre-existing warnings in
  `pty-manager.ts`, all on untouched lines)
- `commitlint --from origin/chore/improvement-sweep --to HEAD` — 0 problems

The root suite's `scripts/__tests__/core-launcher.test.mjs` fails on this
machine only, because a real `/opt/actana` install is present; it is unrelated
to this change and fails the same way on the base commit.

## Scope

Base is `chore/improvement-sweep` (the 0.3.2 improvement batch, cut from
`beta/0.3.1`), **not** `main` and not a beta branch. Touches only
`packages/core/src/pty-manager.ts` and its test — no overlap with #224
(`CoreFilesErrorCode` in the SDK) or #202 (release skill documentation).

Draft on purpose.

### DIFF
--- a/packages/core/src/__tests__/pty-manager.test.ts
+++ b/packages/core/src/__tests__/pty-manager.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from "vitest";
-import { disposePty, planLaunchPortKillTargets } from "../pty-manager";
+import { disposePty, planLaunchPortKillTargets, sanitizeInitialInput } from "../pty-manager";
 
 describe("planLaunchPortKillTargets", () => {
   it("marks Mission Control runtime ports as protected", () => {
@@ -45,5 +45,81 @@ describe("disposePty", () => {
         },
       } as never),
     ).not.toThrow();
+  });
+});
+
+describe("sanitizeInitialInput", () => {
+  // Issue #193: a line break is a C0 byte, so it failed the `code >= 32` test
+  // and was dropped like any other control byte — and `.join("")` left nothing
+  // in its place, welding the word before the break onto the word after it
+  // ("the diff" + "then open" => "diffthen"). Every programmatic starting
+  // prompt (Ship / Sync / Create-PR) is sanitised here, so these cases pin the
+  // separator semantics rather than leaving them to the next reader.
+  it("keeps a separator where a line break was", () => {
+    expect(sanitizeInitialInput("review the diff\nthen open a PR")).toBe(
+      "review the diff then open a PR",
+    );
+  });
+
+  it("drops a leading or trailing line break without leaving stray whitespace", () => {
+    expect(sanitizeInitialInput("review the diff\n")).toBe("review the diff");
+    expect(sanitizeInitialInput("\nreview the diff")).toBe("review the diff");
+    expect(sanitizeInitialInput("\nreview the diff\n")).toBe("review the diff");
+  });
+
+  it("collapses a run of line breaks to a single separator", () => {
+    expect(sanitizeInitialInput("first paragraph\n\n\nsecond paragraph")).toBe(
+      "first paragraph second paragraph",
+    );
+  });
+
+  it("treats carriage returns like line feeds, CRLF included", () => {
+    expect(sanitizeInitialInput("review the diff\r\nthen open a PR")).toBe(
+      "review the diff then open a PR",
+    );
+    expect(sanitizeInitialInput("review the diff\rthen open a PR")).toBe(
+      "review the diff then open a PR",
+    );
+    expect(sanitizeInitialInput("review the diff\n\rthen open a PR")).toBe(
+      "review the diff then open a PR",
+    );
+  });
+
+  it("returns undefined for a prompt that is only line breaks", () => {
+    expect(sanitizeInitialInput("\n")).toBeUndefined();
+    expect(sanitizeInitialInput("\n\n\n")).toBeUndefined();
+    expect(sanitizeInitialInput("\r\n\r\n")).toBeUndefined();
+  });
+
+  it("returns undefined when there is nothing to send", () => {
+    expect(sanitizeInitialInput(undefined)).toBeUndefined();
+    expect(sanitizeInitialInput("")).toBeUndefined();
+    expect(sanitizeInitialInput("   ")).toBeUndefined();
+  });
+
+  it("still refuses to pass keybinding control bytes through to the TUI", () => {
+    // The filter's original purpose stands: a caller must not be able to drive
+    // the harness TUI from a starting prompt. A non-whitespace C0/DEL byte was
+    // never text, so it is removed with no separator in its place.
+    expect(sanitizeInitialInput("esc\x1b[Ainjected")).toBe("esc[Ainjected");
+    expect(sanitizeInitialInput("ctrl\x03c")).toBe("ctrlc");
+    expect(sanitizeInitialInput("del\x7fbyte")).toBe("delbyte");
+  });
+
+  it("separates words joined by a tab or a form feed", () => {
+    expect(sanitizeInitialInput("review the diff\tthen open a PR")).toBe(
+      "review the diff then open a PR",
+    );
+    expect(sanitizeInitialInput("review the diff\vthen open a PR")).toBe(
+      "review the diff then open a PR",
+    );
+    expect(sanitizeInitialInput("review the diff\fthen open a PR")).toBe(
+      "review the diff then open a PR",
+    );
+  });
+
+  it("leaves ordinary text, its spacing and its non-ASCII alone", () => {
+    expect(sanitizeInitialInput("  ship it — please  ")).toBe("ship it — please");
+    expect(sanitizeInitialInput("two  spaces stay")).toBe("two  spaces stay");
   });
 });
--- a/packages/core/src/pty-manager.ts
+++ b/packages/core/src/pty-manager.ts
@@ -254,12 +254,22 @@ function appendBuffer(p: Pty, data: string): number {
   return seq;
 }
 
+// The C0 bytes that are whitespace in the source text: tab, line feed,
+// vertical tab, form feed, carriage return. They cannot be written through
+// verbatim — a bare CR/LF submits the prompt early — but they *separate* the
+// words either side of them, so a run of them collapses to one space instead
+// of vanishing (issue #193). Dropping them outright welded the last word of a
+// line onto the first word of the next: "review the diff\nthen open a PR"
+// arrived as "review the diffthen open a PR".
+const WHITESPACE_CONTROL_RUN = /[\t\n\v\f\r]+/g;
+
 // A programmatic starting prompt (Ship / Sync / Create-PR) is written to the
-// agent's stdin like the user typing. Drop C0/DEL control bytes so a caller
-// can't drive TUI keybindings; the submit CR is added separately.
-function sanitizeInitialInput(text: string | undefined): string | undefined {
+// agent's stdin like the user typing. Whitespace control bytes become a single
+// space; every other C0/DEL byte is dropped with nothing in its place, so a
+// caller can't drive TUI keybindings. The submit CR is added separately.
+export function sanitizeInitialInput(text: string | undefined): string | undefined {
   if (!text) return undefined;
-  const clean = Array.from(text)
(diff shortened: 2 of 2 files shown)


qcentic-dev:
## Acceptance criteria — issue #193

Issue #193 has no formal "Acceptance criteria" heading, so these are derived
from what its body actually asks for, plus the behaviour cases the ticket was
handed with. Each is checked, or explicitly unchecked with the reason.

- [x] **A dropped line break leaves a separator behind.** `sanitizeInitialInput()`
  replaces each run of whitespace control bytes with a single space before the
  C0/DEL filter runs, so the words either side stay separate words.
  (`packages/core/src/pty-manager.ts`)
- [x] **The issue's own example is fixed.** `"review the diff\nthen open a PR"`
  now delivers as `review the diff then open a PR`; `diff` and `then` are no
  longer welded into `diffthen`. Pinned as a test case.
- [x] **Newlines still do not pass through verbatim.** They become a space, never
  a literal CR/LF, so a caller still cannot submit the prompt early. The submit
  carriage return is still `HarnessPromptDelivery`'s to send.
- [x] **The filter's original purpose is intact.** Every non-whitespace C0/DEL
  byte is still dropped with *nothing* in its place — ESC, `\x03` and DEL are
  covered by a test asserting they leave no separator. A caller still cannot
  drive TUI keybindings through a starting prompt.
- [x] **Case: trailing newline.** `"review the diff\n"` → `"review the diff"`,
  no trailing space. Leading and both-ends variants covered too.
- [x] **Case: repeated newlines.** `"first paragraph\n\n\nsecond paragraph"` →
  `"first paragraph second paragraph"` — a run collapses to one separator, so a
  blank line does not pad the prompt.
- [x] **Case: carriage returns.** `\r`, `\r\n` and `\n\r` all behave like `\n`
  and all yield exactly one space.
- [x] **Case: a string that is only newlines.** `"\n"`, `"\n\n\n"` and
  `"\r\n\r\n"` all return `undefined`, so no empty prompt is delivered. The
  `undefined` / `""` / all-spaces inputs are pinned alongside them.
- [x] **The wide blast radius is pinned, not just spot-checked.** Tab and form
  feed separate too; ordinary text, internal double spaces and non-ASCII are
  asserted untouched. `sanitizeInitialInput` is exported for the suite the same
  way `disposePty` and `planLaunchPortKillTargets` already are.
- [x] **Verified green.** `vitest run --root packages/core` → 63 files / 1179
  tests pass; `tsc --noEmit` on `packages/core` clean; `eslint` on both changed
  files adds no problems (3 warnings pre-exist on untouched lines). On this PR,
  CI is green across the board — Conventions, Lint, Typecheck, Unit Tests, both
  tarball smokes, both installer e2e runs and the Panel service seam e2e.
  Because the run is green rather than red, the `pnpm -r` first-failure bail
  did not apply and `packages/core` genuinely executed.
- [x] **Conventional Commits on every commit.** One commit,
  `fix(core): keep a separator where sanitizeInitialInput drops a line break`.
  `commitlint --from origin/chore/improvement-sweep --to HEAD` reports 0
  problems, so the whole branch passes, not only the PR title.
- [x] **Base is `chore/improvement-sweep`.** Not `main`, not `beta/0.3.1`, not
  `beta/0.3.2`. Branch `fix/core-sanitize-initial-input` was cut from the latest
  `origin/chore/improvement-sweep` (`d438e5c`).
- [x] **Boundaries respected.** The diff touches only
  `packages/core/src/pty-manager.ts` and
  `packages/core/src/__tests__/pty-manager.test.ts` — no contact with #224's
  `CoreFilesErrorCode` in the SDK or #202's release skill documentation.
- [x] **Left as a draft, nothing merged.** No branch was merged, no beta branch
  or `main` was touched, and this PR stays in draft.
- [ ] **A separator other than a space was agreed.** Unchecked because no such
  agreement exists to honour: the issue offers "a space (or another agreed
  separator)" and a space is what this implements. Flagging it so a reviewer who
  wants a different separator says so now rather than after the train.
- [ ] **Callers that build multi-line prompts were re-examined.** Unchecked
  deliberately — out of scope. This PR fixes the sanitiser's separator
  semantics, which is all #193 asks for. Whether the Ship / Sync / Create-PR
  prompt *text* still reads well once its line breaks become spaces is a
  product question for whoever owns those prompt strings.

**PR:** #251 — https://github.com/actana/control/pull/251

mrfqcentic (approved):
Reviewed the enforcing lines rather than the description, in a separate worktree at the PR head (`f581b68`), against `origin/chore/improvement-sweep` (`d438e5c`).

## The fix does what #193 asked

`packages/core/src/pty-manager.ts:263` defines `WHITESPACE_CONTROL_RUN = /[\t\n\v\f\r]+/g`, and line 272 applies it — `text.replace(WHITESPACE_CONTROL_RUN, " ")` — *before* the untouched `code >= 32 && code !== 127` filter on lines 274-275. That ordering is the whole fix: the line break is turned into a space while it is still there to be seen, so the filter that used to eat it now has nothing left to eat. `"review the diff\nthen open a PR"` delivers as `review the diff then open a PR`; `diff` and `then` are no longer welded into `diffthen`.

The filter's original purpose is intact. I ran the function's exact body over 20,000 mixed inputs: **no CR, LF or any other C0/DEL byte survives into the delivered prompt**, so a caller still cannot drive TUI keybindings or submit early through a starting prompt. The submit `\r` remains `HarnessPromptDelivery`'s to send (`pty-manager.ts:658-666`).

## The edge cases are genuinely pinned

All 22 assertions in `packages/core/src/__tests__/pty-manager.test.ts:50-125` pass when executed against the real implementation, not just by inspection. The four that matter here are each present as their own case: trailing newline (line 65, plus leading and both-ends), repeated newlines (line 70), carriage returns including `\r`, `\r\n` and `\n\r` (line 76), and a string that is only newlines (line 88, with `"\n"`, `"\n\n\n"` and `"\r\n\r\n"` all returning `undefined`). The `ESC` / `\x03` / `DEL` case at line 100 is the one that keeps the fix honest — those still leave *no* separator.

## The separator does not change a prompt that was previously fine

This was the thing worth proving, given the blast radius. A prompt containing no whitespace control byte never touches the new regex, so its output is byte-identical to the old behaviour — confirmed by diffing old against new over control-byte-free prompts including the shipped `DEFAULT_SHIP_PROMPT`. The only inputs whose output changes are exactly the ones that were being welded.

## No existing caller relied on the welding

Stated plainly: **none.** `sanitizeInitialInput` has one call site (`pty-manager.ts:655`), and the prompt text reaching it comes from only three places — Panel Ship (`projects.$id.tsx:1030`, whose default is single-line but which is a free-text `<textarea>` at `DefaultsSettingsPage.tsx:150`, so operator-authored multi-line Ship prompts were being mangled in production), the CLI (`session-command.ts:214`, where `readText` joins argv with a space and reads long prompts from stdin — the real multi-line path), and the SDK (`core-session.ts:401`). Nothing builds a prompt by joining lines expecting them to fuse, and no prompt string pads a line-end with a space to survive the weld. The sanitiser dates to the initial commit, so the welding was never a behaviour anything was written against. Downstream is safe too: `promptEchoed`/`promptEchoProbe` squeeze all whitespace out before comparing, so the added space cannot cause a false "prompt swallowed" re-type.

## Non-blocking notes

- `pty-manager.ts:263` — a literal space adjacent to a line break is not collapsed with it: `"diff \n then"` yields three spaces where the old code gave two. Cosmetic only, and deliberate in that `"two  spaces stay"` (test line 123) pins that literal spacing is left alone. The test comment at line 69 and the PR body read slightly stronger than the code — a *control* run collapses to one space; a run interleaved with literal spaces or split by another C0 byte does not.
- `pty-manager.ts:263` — the class omits NEL (`U+0085`), which is still dropped with no separator, and `U+2028`/`U+2029` pass through verbatim as they always did. Both pre-existing and vanishingly rare in a prompt; not worth widening the diff for.
- The one class of input the old bug got "right" is a soft word-wrap join — `"auto-\ncomplete"` was delivered as `auto-complete` and now becomes `auto- complete`. That is an accident of the bug, not a contract, and #193 explicitly sanctions a space.
- `commit f581b68` footer says `Refs #193` while the PR body opens `Closes #193.` — the PR body is what closes the issue on squash-merge, so nothing to do.
- Tab/VT/FF getting the same treatment as CR/LF is slightly wider than #193's literal wording, but it is the same defect and pinning it is the right call.

CI is green on `f581b68` across all 19 checks, `Train rules` included — a `chore/*` base is not policed by `ci.yml`'s train rules (only `main` and `beta/*` are), so the deliberate `chore/improvement-sweep` base is fine.

Approving. Leaving the draft status alone — that is the author's call.

PR #253, opened 2026-08-18
refactor(sdk): give CoreFilesErrorCode a single owner

### LINKED ISSUE #224: CoreFilesErrorCode is defined twice: the Core owns the checked copy, the SDK keeps a mirror nothing checks

qcentic-dev:
Closes #224.

`CoreFilesErrorCode` — the 22-member refusal vocabulary of the Core's file surface — was written out verbatim in two packages. Nothing was broken when the ticket was filed and the two copies agreed member for member; what made it worth a ticket is that they were **not equally load-bearing**, and the asymmetry ran the wrong way round.

- The Core's copy was **checked**: `Refusal` types every refusal the routes send, and `TarRefusalCode` / `FileConfinementRefusal` are pinned as subsets by assignment. Inside that package a copy could not drift.
- The SDK's copy was **checked by nothing**, deliberately: every use site widens it to `CoreFilesErrorCode | string`, because a client meeting a Core newer than itself has to survive a code it has never heard of.

So the one copy free to drift was the one nothing would have noticed, and the failure would not have been a break — it would have been a Core answering a code its client cannot name, with every consumer's `catch` still compiling. This seam has already done exactly that once, in #218/#219, through two green suites.

## What the ticket asked to be decided

#224 named four questions rather than a fix. Answers, with the reasoning recorded in-tree:

**1. Which package owns it — the SDK.** ADR 0025 D2's argument transfers whole: what the Core depends on is the protocol, not the client. The Core (the *server*) imports the vocabulary from the *client* package, which is the same arrow the frames already use.

**2. Where in the SDK — a contract-only module.** `packages/sdk/src/core-files-error-codes.ts`, beside `core-link-frames.ts`, not a corner of `core-files-http.ts`. An `import type` erases either way, so the deciding argument is not runtime: D2's "and nothing else from `@actana/sdk`" is only reviewable if what the Core may import is a module a reviewer can see holds nothing else. This one imports nothing at all, exactly as the frames module does. Confirmed: the Core's esbuild bundle gains no runtime edge.

**3. The published surface — named on purpose.** `packages/sdk/package.json` exports `./*`, so any module under `src/` already is public API. `@actana/sdk/core-files-error-codes` is therefore a published subpath on the next publish, and it is named deliberately rather than fallen into — the refusal vocabulary is precisely what a third party writing a `catch` wants to type against. `core-files-http.ts` re-exports the type, so the specifier consumers already import keeps working; a re-export is not a second definition, because it cannot disagree with what it aliases. Nothing published moves.

**4. One definition *and* a test — but not the test the ticket priced.** Option 4 was "keep both copies and pin them". That is rejected on ADR 0025 D3's terms: it pins the mirror rather than removing it. But the pinning idea is still needed for a *different* copy — `docs/external-api.md`, which states the same vocabulary in prose and typechecks never. So: ownership removes the drift a compiler can reach, and a contract test in the manner of `files-list-contract.ts` holds the copy it cannot.

## Proving the single owner

Per the ticket, by reading import sites rather than trusting a grep of the type name. A grep for `CoreFilesErrorCode` finds only files that name the type; a copy need not. So every one of the 22 members was swept for across the tree, counting members per file:

| Members present | File |
| --- | --- |
| 22 | `packages/sdk/src/core-files-http.ts` |
| 22 | `packages/core/src/core-files-routes.ts` |
| 13 | `packages/core/src/__tests__/core-files-routes.test.ts` |
| 11 | `packages/core/src/files-tar.ts` (`TarRefusalCode`, 9 members — compiler-pinned subset) |
| ≤9 | test files and `files-confinement.ts` (`FileConfinementRefusal`, 4 — compiler-pinned subset) |

Only two files carried the whole union, and both are now importers. Every importer of both modules was read: 18 import sites for `core-files-http` and 21 for `core-files-routes`, across all five packages. One comment in `packages/cli/src/local-tar.ts:751` refers to "the same two-copies problem" and was chased down — it is about `ReadableStream`'s two TypeScript declarations, not about this union.

## Adding a refusal code, demonstrated rather than asserted

Done by actually doing it on this branch and then reverting: a 23rd member, `demo-only-probe`, appended to `CORE_FILES_ERROR_CODES` and **to no other file**.

- `pnpm --filter @actana/core typecheck` — passes. The Core accepts the new code with no edit of its own; a refusal was rewritten to send it, and that typechecks too.
- `pnpm --filter @actana/sdk typecheck` — passes. The error classes accept it with no edit.
- The docs contract goes **red in both suites**, naming the missing code:
  `AssertionError: expected [ … ] to deeply equal [ … ] - "demo-only-probe"`

So: one file in the source, and the document cannot be forgotten because the suite says so.

## What the contract test does

`packages/sdk/src/__tests__/files-error-code-contract.ts` compares the `Code` column of the refusal tables between the `refusal-codes` markers in `docs/external-api.md` against `CORE_FILES_ERROR_CODES`, as **set equality** — a code in the union and not the table is a refusal nobody can look up, and a code in the table and not the union is worse, since it sends somebody writing a `catch` for something no Core will ever send. It reads the header row to find the `Code` column rather than assuming an index (the two tables have different shapes), reads only that column (the `Means` column is full of backticked things that are not codes), and throws rather than passing vacuously if the markers are gone — which is asserted.

It is registered in **both** packages' suites, for #218's reason: a new refusal code is *written* in the Core, by an author who may run only the Core's suite, while the vocabulary is *owned* by the SDK.

## Also in this branch

`docs/external-api.md` documented 14 of the 22 codes. The eight it never named are the ones with no status line to arrive on — a tar `PUT` and a listing both spend their `200` on the first entry, so a failure met part-way comes back as the NDJSON `error` line. They get a table of their own; `root-entry-path` stays in the status table, where a reader will look for it.

The `| string` widening is untouched and is now argued at the error classes in `core-files-http.ts`, so tightening it later is a decision rather than a cleanup. The change is explicit that one definition does **not** make the surface drift-proof: a deployed client older than the Core it dials still meets codes it cannot name, which is a fact about deployments, not about this repository.

ADR 0025 D2 is amended in place (per #107's rule) — the reviewer-applicable rule now names two modules, and three consequences are added.

## Base

`chore/improvement-sweep`, the 0.3.2 improvement batch cut from `beta/0.3.1`. **Not** `main` and not a beta branch.

## Verification

| Check | Result |
| --- | --- |
| `pnpm --filter @actana/sdk test` | 239 passed, 5 skipped |
| `pnpm --filter @actana/core test` | 1173 passed (64 files) |
| `pnpm --filter @actana/cli test` | 421 passed, 3 skipped |
| root `vitest run` | 474 passed (16 files) |
| `typecheck` — sdk, core, cli, panel, shared | all clean |
| `pnpm lint` | 0 errors (50 pre-existing warnings, none in changed files) |
| commitlint, all 4 commits + title | 0 problems, 0 warnings |

Panel: the three `core-files` suites pass. Under a full local `pnpm -r test` the Panel reports timeout failures in unrelated suites (terminals, harness-install, project-picker, token-usage, live-read-path) — a **different set each run**, and all six pass in isolation. That is a known local resource flake on this machine, not a regression from this change; the Panel does not import the vocabulary at all.


qcentic-dev:
## Issue #224 — acceptance criteria

- [x] **There is one definition of the refusal codes, or one test that goes red when the two disagree — registered in both packages' suites, in the manner of `files-list-contract.ts`.**
  Both, for two different copies. **One definition**: `CORE_FILES_ERROR_CODES` in `packages/sdk/src/core-files-error-codes.ts`; `packages/core/src/core-files-routes.ts` and `packages/sdk/src/core-files-http.ts` both import it and their copies are deleted. That removes the SDK↔Core drift. **One test**: `packages/sdk/src/__tests__/files-error-code-contract.ts` pins `docs/external-api.md` to the union as set equality, registered in both suites (`packages/sdk/src/__tests__/core-files-error-codes.test.ts`, `packages/core/src/__tests__/core-files-error-codes.test.ts`) — because the vocabulary is owned by the SDK but a new refusal is written in the Core.

- [x] **Adding a refusal code is a one-file change, or a change that fails until every copy is updated. Demonstrated by doing it, not asserted.**
  Demonstrated on this branch and reverted. A 23rd member `demo-only-probe` was appended to `CORE_FILES_ERROR_CODES` and to no other file: `@actana/core` typecheck passed (and a Core refusal rewritten to send the new code typechecked), `@actana/sdk` typecheck passed, and the docs contract went red in **both** suites with `AssertionError: … - "demo-only-probe"`. One file in the source; the document cannot be silently forgotten.

- [x] **The choice of owner (or of pinning instead) is recorded where the next reader meets it — in-tree beside the union, and in ADR 0025's consequences or a new record if the scope of D1/D3 moves.**
  In-tree: the header of `core-files-error-codes.ts` answers all four of the ticket's questions — which package owns it, why a module of its own rather than `core-files-http.ts`, that it is public API on the next publish, and why one definition rather than pinning the mirror. In the ADR: D1/D3's scope did move, so ADR 0025 is amended in place per #107's rule — D2's reviewer-applicable rule now reads "`@actana/core` may import `@actana/sdk/core-link-frames` **and `@actana/sdk/core-files-error-codes`**, and nothing else from `@actana/sdk`", with three consequences added.

- [x] **The SDK keeps `CoreFilesErrorCode | string` at the error classes, with the reason stated in-tree, so a future tightening is a decision rather than a cleanup.**
  Untouched at all five sites (`CoreFilesRequestError`, `CoreFilesConflictError`, `CoreFilesStreamError`). The reason is now stated at the error classes themselves in `core-files-http.ts` rather than left to be inferred: a client is routinely older than the Core it dials, so the closed union would make honest handling a cast at every call site. It says explicitly that narrowing it is a decision, not a cleanup — and, per the ticket's own note, that one definition does **not** close the widening's blind spot, because that is a fact about deployments rather than about this repository.

- [x] **`docs/external-api.md` still agrees with the union after the change.**
  It agrees more than it did. The document listed **14 of the 22** codes before this branch — the eight with no status line to arrive on (`corrupt-archive`, the entry-path refusals, `unsupported-entry-type`, the two link-target refusals, `write-failed`, `read-failed`) had no row anywhere. They now have a table of their own for refusals that arrive as an NDJSON `error` line, and agreement is enforced by the contract test rather than eyeballed.

### Not done, and why

- **Nothing.** Every criterion above is checked. Two things were deliberately *not* done because the ticket rules them out or reserves them: the `| string` widening was not narrowed (criterion 4 requires it survive), and option 4's "keep both copies and pin them" was priced and rejected on ADR 0025 D3's terms — it pins the mirror rather than removing it. The pinning technique is still used, but for `docs/external-api.md`, which is a copy no compiler can reach.

### Boundaries respected

`#193` (`sanitizeInitialInput` in the Core) and `#202` (release skill documentation) are untouched — no file in this diff is in either's territory. Nothing was merged; `main` and every `beta/*` branch are untouched; the PR is a **draft** with base `chore/improvement-sweep`.

### Verification

`@actana/sdk` 239 passed / 5 skipped · `@actana/core` 1173 passed (64 files) · `@actana/cli` 421 passed / 3 skipped · root `vitest run` 474 passed (16 files) · typecheck clean in all five packages · `pnpm lint` 0 errors · commitlint 0 problems over all 4 commits and the PR title.

Reading a red run correctly on this repo: `pnpm -r` bails on the first failing package and `packages/cli` sorts before `packages/core`, so each package's suite was run **individually** rather than trusting one recursive run. Panel: the three `core-files` suites pass; a full local Panel run reports timeout failures in unrelated suites — a *different set each run*, all passing in isolation — which is a local resource flake on this machine, not a regression. The Panel does not import the vocabulary.

**PR #253** — https://github.com/actana/control/pull/253

qcentic-dev:
CI on `37deecb`: **17 pass, 5 skipping, 0 failing, 0 pending.** `Conventions` (commitlint over all four commits *and* the title) passed, as did `Typecheck`, `Lint` and `Unit Tests`. The five skips are the train-only image publish jobs.

Still a **draft**, base still `chore/improvement-sweep`. Nothing merged; `main` and every `beta/*` branch untouched — `origin/chore/improvement-sweep` is still at `d438e5c`, the commit this branch was cut from.

mrfqcentic (approved):
## Review — #224, `CoreFilesErrorCode` single owner

Reviewed at `37deecb` against `origin/chore/improvement-sweep` (`d438e5c`) in a clean worktree. **Approving.**

The ticket asked for this to be proven by reading enforcing lines rather than by a grep of the type name, so that is what I did.

### The single-owner claim, proven

I swept all 22 members across the tree **independently of the type name** — a copy need not name the type. Exactly one file carries the full union:

| Members | File |
| --- | --- |
| 22 | `packages/sdk/src/core-files-error-codes.ts` |
| 13 | `packages/core/src/__tests__/core-files-routes.test.ts` — assertion literals in a table-driven test, not a declaration |
| 11 | `packages/core/src/files-tar.ts` (`TarRefusalCode`, 9) |
| 4 | `packages/core/src/files-confinement.ts` (`FileConfinementRefusal`) |

Both former copies are gone. What sits at those two sites now are `export type { … } from` **aliases** — `core-files-routes.ts:114`, `core-files-http.ts:70` — which cannot disagree with what they alias.

Every type-position use resolves to the owner, with nothing re-declaring it locally:

- `core-files-routes.ts:38` → `:116` (`Refusal`), `:569`
- `core-files-http.ts:60` → `:126`, `:128`, `:151`, `:166`, `:168`
- `core-files.ts:51` → `:283`

The two narrower unions are still **compiler-pinned subsets**, unchanged: `TarRefusalCode` by the assignment at `core-files-routes.ts:569`, `FileConfinementRefusal` by `code: confined.reason` at `:236`. Worth noting the ownership fix also closed a hole it wasn't aimed at: a local `type CoreFilesErrorCode` re-declared in `core-files-routes.ts` is now a duplicate-identifier error against the re-export at `:114`.

`core-files-error-codes.ts` imports nothing at all — I checked, and confirmed `core-link-frames.ts` has the same property, so the D2 precedent holds exactly.

### Patch-release safety — this is a pure refactor

- **No error code changes.** I diffed the new `CORE_FILES_ERROR_CODES` against *both* base copies: byte-identical, 22 members, **same order**. Nothing added, removed, renamed or reordered.
- **Published surface is additive only.** `CoreFilesErrorCode` still exports from both prior specifiers. `@actana/sdk/core-files-error-codes` is a new subpath (the manifest already exported `./*`) and adds one new runtime value.
- **No runtime behaviour change.** Both Core-side edges are explicitly `type`-marked, so esbuild erases them at parse time without resolving — `build.mjs` gains no edge into the SDK. Corroborated by the green Core image build and both release-tarball smokes on this SHA.

### The contract test, executed rather than trusted

I ran the parser in `files-error-code-contract.ts` against the real `docs/external-api.md`: **set equality holds, 22 = 22**. Negative control — appending a 23rd member yields `['demo-only-probe']` as undocumented, so it does go red. `directory-in-the-way` legitimately appears in both tables and the dedup handles it.

I also verified the new NDJSON prose against the implementation rather than taking it on trust, since it is new normative documentation: `done` is written in the `try` and `error` in the `catch` (`:303`/`:318`, `:550`/`:572`), so "an `error` line replaces the `done` line" is exactly right; entries are written as they go with no rollback; `write-failed` is both the mid-stream fallback (`:569`) and the 500 (`:139`, `:167`). The 8 codes in the new table plus `root-entry-path` in the status table account for `TarRefusalCode`'s 9.

### Acceptance criteria

1. **One definition, or a test that goes red — met.** Both, for two different copies.
2. **Adding a code is a one-file change, demonstrated — met.** I reproduced the enforcing half myself (above) rather than relying on the reverted demo.
3. **Owner recorded where the next reader meets it — met.** In-tree header answers all four questions; ADR 0025 gains three consequences.
4. **`| string` survives with the reason in-tree — met.** Untouched at all five sites; argued at `core-files-http.ts:98-119`.
5. **`docs/external-api.md` agrees — met, and better than before** (14 of 22 documented → 22, now enforced).

### Two non-blocking nits

- `docs/adr/0025-…:10` — D1 still reads as scoped to `core-link-frames.ts` with no amendment pointer, while the amendment sits under D2 at `:17` and its own text says "D1 scoped this record to the frames file". The criterion asked for the consequences section, which you did, so this is a nicety: a clause on D1 would save the next reader the seven lines.
- `packages/core/src/__tests__/core-files-error-codes.test.ts` / `packages/sdk/src/__tests__/core-files-error-codes.test.ts` — the precedent pairs body `files-list-contract.ts` with registrations named `core-files-list-contract.test.ts`. Dropping "contract" makes these read like unit tests *of* `core-files-error-codes.ts`. `core-files-error-code-contract.test.ts` would restore the pattern.

Neither is worth another round. Good change — the asymmetry argument is right, and the half that needed a test rather than a compiler got one.

Issue #58, opened 2026-08-03
Delete the 63 triaged files and repoint all ten references


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D46**.

**Phase 4 — Docs and deletions**
**Blocked by:** #53, #54, #55, #56, #57
**Blocks:** #59

## What

**Promotions first, deletion second** — steps 2–3 of #17's landing order must precede this one so no citation is ever dangling at a commit boundary. That is why this ticket is blocked by four ADR/doc tickets and by T28.

63 files: `.scratch/` (38 — 3 `spec.md` + 35 issue notes), `docs/specs/` (12), `docs/tickets/` (13), plus the `## Historical working notes` section of `docs/agents/issue-tracker.md` (it exists to point at `.scratch/` and must not outlive it).

Everything else failed #17's test — *does this document a decision or constraint that is still true, and is it recorded nowhere else?* The ADRs, `domain-model.md`, `ci-cd.md`, `INSTALL.md` and above all the **code's own comments** already hold it, often in better words. A large slice is not merely redundant but **void**: all 12 `docs/tickets/` files are Electron-era breakdowns of code ADR 0010 removed.

Restate the justification: the archive is this clone and the `azure` remote, **not** git history.

## Done when

- [ ] All 63 files are gone
- [ ] The nine `packages/shared/src/schema-bootstrap.ts` citations are repointed — lines 460/466/474/483/501 → ADR 0007, 492 → ADR 0006 (drop path only), 510 → ADR 0009 (drop path only), **518 → ADR 0014, 527 → ADR 0015** (shifted per D44). The comments themselves are self-contained and stay; only the citation dies
- [ ] `.github/workflows/*` `paths-ignore` drops the dead `.scratch/**` entry; `.dockerignore:25` drops `.scratch`
- [ ] `docs/README.md`'s `## Historical record` section is removed entirely — `upstream/` and `refactor-plan.md` no longer exist to hold it up
- [ ] `docs/upstream/`'s four `../specs` links (`DIVERGENCE.md:5`, `PROVENANCE.md:4,79`, `BACKLOG.md:4`) become ADR references
- [ ] `rg -n 'docs/specs|docs/tickets|\.scratch' --hidden` returns nothing outside this PR's own diff

---

[...]

Issue #57, opened 2026-08-03
Move repo-template/ out of the clone before it is deleted


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D46**.

**Phase 4 — Docs and deletions**
**Blocked by:** nothing — can start immediately
**Blocks:** #58

## What

**Deleting this destroys it.** `repo-template/` (21 files) is untracked, exists only in this working clone, and is in no git history anywhere — including `actana/control`, which is one squashed commit. #6 frame item 10's *"git history is the archive"* is false here, and this is the case where that matters most.

It also should not ship. Its own README calls it an *"Enterprise Repo Governance Template … a drop-in convention package for open source / team repositories"*. It is generic, names no product, and shipping it means every reader meets a second competing set of `CONTRIBUTING.md`/`SECURITY.md`/`CODE_OF_CONDUCT.md`/`SUPPORT.md` in the tree — precisely the "which one is real?" failure #6 exists to kill.

It has already done its job: `GOVERNANCE.md` is byte-identical to the root copy (copied, never customised) and the other four differ (customised). The template was the source; the root files are the applied output. Keeping the source next to the output is how they drift.

## Done when

- [ ] `repo-template/` exists at its own home — its own repository, or at minimum a directory outside this working tree — **verified present there** before anything is removed
- [ ] Only then is it removed from the clone
- [ ] It is **not** added to `.gitignore`: an ignore rule for a directory that should not exist is a note to a future reader that it might come back

---

<sub>Decisions are locked in ADR 0016. If this ticket needs one changed, amend the ADR in the same PR — do not decide it in a comment.</sub>

Issue #56, opened 2026-08-03
K4: fix actano.ai in ADR 0007


qcentic-dev:
Part of #23. **Spec:** [ADR 0016 — The 0.1.0 shape](https://github.com/actana/control/blob/main/docs/adr/0016-the-0-1-0-shape.md), clauses **D46**.

**Phase 4 — Docs and deletions**
**Blocked by:** nothing — can start immediately
**Blocks:** #58

## What

`docs/adr/0007-scope-narrowing-and-rebrand.md:24` says the update-manager stub points at `control.actano.ai`. The only place resolving that typo is `docs/specs/09-rebrand-and-auto-update.md:377` — *"`actana.ai` is authoritative"* — which T29 deletes. Delete it first and a live ADR keeps a misspelled domain with nothing to correct it.

Doubly stale: `electron/update-manager.ts` went with Electron (ADR 0010). Correct the spelling or strike the clause; cheap either way. Same PR also strikes the `docs/specs/` pointer at :19.

## Done when

- [ ] `actano.ai` appears nowhere in the repo
- [ ] ADR 0007:19's `docs/specs/` pointer is struck; the ADR stands without it
- [ ] ADR 0009:17's `docs/specs/10-remove-sandbox.md` reference is rewritten without the path

---

<sub>Decisions are locked in ADR 0016. If this ticket needs one changed, amend the ADR in the same PR — do not decide it in a comment.</sub>

qcentic-adm:
**Left open: one of the three "Done when" boxes is not satisfied by the shipped tree.**

The work for this ticket rode the **0.3.2** train — on `main` as **6cd15ca** (`chore: 0.3.2 improvement sweep`), promoted through **#262**, from **#254** on the train. Two of the three boxes are met; the first is not, and #254's own commit message says so deliberately, so this is a scope decision to confirm rather than a mistake to hunt.

**Met:**

- **"ADR 0007:19's `docs/specs/` pointer is struck; the ADR stands without it."** The Consequences bullet now reads *"Nine removal specs land as separate PRs (Whisper, Pet + multiplayer, … plus cross-core notifications add and the rebrand)."* — the trailing *"See `docs/specs/`."* is gone from `docs/adr/0007-scope-narrowing-and-rebrand.md:21` and the sentence stands without it.
[...]

PR #256, opened 2026-08-18
docs: record why the release is unsigned, and repoint the citations that said so

### LINKED ISSUE #55: K3: the no-code-signing paragraph, and repoint install.sh:295

qcentic-dev:
Closes #55. Part of #23, ADR 0016 clause **D46** (K3).

Promotes the last home of the "no code signing" decision out of
`.scratch/harness-installer/spec.md` — a file #58 deletes — and repoints the two
documentation citations that name it.

## What changed

- **`docs/ci-cd.md`** — a new `### Integrity is published checksums, not signatures`
  under `## The release artifacts`: what is and is not signed, what the checksums
  prove and what they do not, why unsigned is safe for a `curl | bash` transport,
  the `xattr` corollary, and the cross-link to #24.
- **`install.sh`** — the checksum comment cited `spec: "No code signing"`; it now
  cites `docs/ci-cd.md` and the section by name. **Comment only** — no installer
  behaviour changes, and `sh -n install.sh` is clean.
- **`docs/core-macos-prerelease-checklist.md`** — its Gatekeeper note deferred to
  #55, which this PR closes. Left alone it would have pointed at a closed issue
  instead of at the reason, so it now links the section.

No source file and no workflow is touched.

## Accuracy: what I checked before writing it

The ticket's framing was checked against what the release actually produces
today, not against the spec being deleted, and it needed one correction.

- **`.github/workflows/release.yml`** — the assets are three tarballs plus
  `SHA256SUMS` (asserted at an explicit count check), composed and re-verified
  with `sha256sum -c`. No signing secret on the path: `github.token` for the
  assets, the Docker Hub PAT for the mirror.
- **`.github/workflows/container-image.yml`** — the images are built with plain
  `docker build` and stitched with `docker buildx imagetools create`. Not
  `buildx build`, so there is no BuildKit provenance or SBOM attestation on them
  either.
- **The correction.** The npm job publishes `@actana/sdk` and `@actana/cli` with
  `--provenance` and reads the SLSA attestation back off the registry before the
  release succeeds. A flat "nothing in this project is signed" — which is how the
  old spec line reads — is contradicted by the release itself. The section states
  the npm exception explicitly and bounds it: provenance attests where a package
  was built, not the tarballs and not the images.
- **`grep -rn 'cosign|notariz|codesign|authenticode'`** across the repo — no
  signing tooling anywhere outside prose about the deleted Electron app.
- **Both verifiers.** `install.sh` fetches `SHA256SUMS` before the tarball and
  dies on mismatch; `packages/core/src/actana-update.ts:123` does the same for
  `actana update`. The section's "verified by both" claim is from those two call
  sites, not from the spec.
- **`docs/adr/0010`** — deferred signing to "Harness distribution planning". This
  section is that answer, and it says so.

## Links: which ones, and how

Every relative link in both changed files was resolved against the worktree with
a script (path exists; `#anchor` present as a heading), not eyeballed:

| Link | Checked how |
| --- | --- |
| `../INSTALL.md` | file exists; read §"Gatekeeper, and downloading the tarball in a browser" — it is the `xattr` escape hatch the section defers to |
| `../packages/core/src/actana-update.ts` | file exists; read the checksum branch that backs the claim |
| `adr/0010-panel-becomes-a-self-hosted-web-service.md` | file exists; read the deferral sentence |
| `#npm` (in-page) | heading `### npm` present → slug resolves |
| `ci-cd.md#integrity-is-published-checksums-not-signatures` (from the checklist) | target file exists **and** the new heading's slug matches |
| #24, #55, #58 | `gh api repos/actana/control/issues/N` — #24 confirmed **open** and titled "Supply chain: sign the release, publish an SBOM and provenance" |

All 35 relative links in the two files resolve, and all 7 in-page anchors in
`docs/ci-cd.md` resolve.

## One deliberate deviation, and one thing left undone

**The path is `docs/ci-cd.md`, not `docs/contribute/ci-cd.md`.** The ticket and
D46 both name `contribute/`. That directory does not exist yet — ADR 0016 line 71
already says `docs/ci-cd.md` is "G3's destination, until #59 moves it to
`docs/contribute/`", and the G3 row carries the same parenthetical. Writing the
ticket's literal path would have shipped exactly the dead link this change exists
to remove, so it is written and cited at the path that exists today and moves
with the tree under #59. **No ADR amendment is needed** — 0016 already anticipates
this.

**`packages/core/src/actana-update.ts:137` carries the identical dangling
citation** (`spec: "No code signing"`) and is **deliberately untouched**: it is
source, and this PR is documentation. It is one of the ten references #58
repoints. Flagging it so it is not lost.

## Boundaries

Nothing here touches ADR 0007's domain correction (#56) or `repo-template/`
(#57). Base is `chore/improvement-sweep`, not `main` and not a beta branch.
Draft, and staying draft.

### DIFF
--- a/docs/ci-cd.md
+++ b/docs/ci-cd.md
@@ -232,8 +232,12 @@ the digest claim is about the container images only.
 container images are unsigned: no Apple notarization, no Windows Authenticode —
 there is no Windows artifact to sign — and no detached GPG or Sigstore signature
 over a tarball or an image manifest. `release.yml` carries no signing secret at
-all; the only credentials on the release path are `github.token` for the assets
-and the Docker Hub PAT for the mirror.
+all — its one signing-adjacent capability, `id-token: write` on the `npm` job,
+is keyless OIDC rather than a key. Three credentials sit on the release path
+and none of them signs anything: `github.token` for the Release assets, the
+`DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` PAT for the images, and `NPM_TOKEN`
+for the packages. The last two are each mandatory, gated by their own `resolve`
+step that fails the release before anything is built when they are unset.
 
 Integrity is **the published checksums**, and both paths that put a Core on a
 machine verify them. `install.sh` fetches `SHA256SUMS` *before* the tarball, and
--- a/docs/ci-cd.md
+++ b/docs/ci-cd.md
@@ -226,6 +226,63 @@ guarantee](#the-digest-guarantee-where-it-starts-and-where-it-stops). The Core
 tarballs are not: they are built by the release, from the promoted commit, and
 the digest claim is about the container images only.
 
+### Integrity is published checksums, not signatures
+
+**Nothing a release ships is code-signed.** The three Core tarballs and both
+container images are unsigned: no Apple notarization, no Windows Authenticode —
+there is no Windows artifact to sign — and no detached GPG or Sigstore signature
+over a tarball or an image manifest. `release.yml` carries no signing secret at
+all; the only credentials on the release path are `github.token` for the assets
+and the Docker Hub PAT for the mirror.
+
+Integrity is **the published checksums**, and both paths that put a Core on a
+machine verify them. `install.sh` fetches `SHA256SUMS` *before* the tarball, and
+`actana update` does the same for an in-place upgrade
+([`actana-update.ts`](../packages/core/src/actana-update.ts)); each compares the
+SHA-256 it computed against the release's own line and refuses to install on a
+mismatch. What that proves is bounded, and both call sites say so in a comment:
+the tarball is the one the release's checksum file describes. The checksums came
+over the same channel as the bytes, so this catches corruption and truncation —
+not a release channel someone else controls. That is the gap a signature would
+close, and it is open.
+
+One release surface is attested, and it is not a counter-example: the npm
+packages are published with `--provenance`, so `@actana/sdk` and `@actana/cli`
+carry a SLSA provenance attestation that the release reads back off the registry
+before it succeeds ([npm](#npm)). Provenance attests *where a package was
+built*. It is not a signature over a released binary, and it covers neither the
+tarballs nor the images.
+
+**Why unsigned is safe for the way this ships.** The distribution path is
+`curl | bash`, and `curl` — like `wget` — does not set the
+`com.apple.quarantine` extended attribute. Nothing arriving by that path is ever
+quarantined, so **Gatekeeper never intervenes**, and there is no dialog for
+notarization to buy off. This is a claim about the *transport*, not about the
+bytes: revisit it the moment a browser-download distribution is added, because a
+browser does set the attribute. That case is already the operator's, not the
+installer's — [`../INSTALL.md`](../INSTALL.md) documents `xattr -dr
+com.apple.quarantine` as the escape hatch for a tarball fetched in Safari or
+Chrome.
+
+**The corollary, which already cost one review cycle: do not add `xattr -dr
+com.apple.quarantine` to `actana setup`.** It is a recursive attribute strip run
+on every install to undo a state the install path cannot produce. On the paths
+`actana setup` actually runs on there is nothing to clear, so it buys nothing;
+on any other path it silently clears quarantine on files that legitimately carry
+it, turning a Gatekeeper decision into something the installer made on the
+operator's behalf without saying so. The strip stays where it is, in `INSTALL.md`,
+run by hand by the person who knows they used a browser.
+
+This is the current posture, not a permanent refusal.
+[ADR 0010](adr/0010-panel-becomes-a-self-hosted-web-service.md) voided the
+Panel's desktop signing story and deferred the question to Harness distribution
+planning; this section is that answer, and
+[#24](https://github.com/actana/control/issues/24) — sign the release, publish
+an SBOM and provenance — is where it changes. Until that lands, treat every
+statement above as load-bearing rather than incidental: `install.sh`'s checksum
+comment points here, and the macOS prerelease checklist's Gatekeeper box assumes
+it.
+
 ## The tag ladder
 
 Five published tag classes, each answering exactly one question ([ADR
--- a/docs/core-macos-prerelease-checklist.md
+++ b/docs/core-macos-prerelease-checklist.md
@@ -24,9 +24,11 @@ rather than a job. A GitHub runner is destroyed rather than restarted, and a
 LaunchAgent is by definition tied to a login session, so "does the Core come
 back after a reboot?" and "does it come back after a logout?" can only be
 answered on a real Mac. Gatekeeper is the same kind of question: the tarball
-ships unsigned and un-notarized (out of scope — see
-[#55](https://github.com/actana/control/issues/55)), and whether macOS lets an
-operator run it is not something a headless runner experiences.
+ships unsigned and un-notarized, deliberately and for a reason that is written
+down — see [Integrity is published checksums, not
+signatures](ci-cd.md#integrity-is-published-checksums-not-signatures) — and
+whether macOS lets an operator run it is not something a headless runner
+experiences.
 
 The `tarball-macos` leg does the part a runner *can* do: it builds the
 `mac-arm64` tarball on an Apple-silicon runner and boots it through its own
--- a/install.sh
+++ b/install.sh
@@ -323,7 +323,8 @@ main() {
   # What this proves: the tarball is the one the release's own checksum file
   # describes. Both came over the same channel, so it catches corruption and
   # truncation, not a release channel someone else controls — the project
-  # publishes no signatures (spec: "No code signing").
+  # publishes no signatures. Why that is safe, and what would change it:
+  # docs/ci-cd.md, "Integrity is published checksums, not signatures".
   say "Checksum verified against the release's $SHASUMS_ASSET."
 
   tar -xzf "$tarball" -C "$work_dir" || die "could not unpack $asset"


qcentic-dev:
## Acceptance criteria — issue #55

- [x] **`ci-cd.md § The release artifacts` records: integrity is published checksums, verified by both `install.sh` and `actana update` — not signatures. No Apple notarization, no Windows Authenticode.**
  New `### Integrity is published checksums, not signatures`. Both verifiers are named from their call sites (`install.sh`'s checksum branch, `packages/core/src/actana-update.ts:123`), not from the spec. Written at **`docs/ci-cd.md`**, not `docs/contribute/ci-cd.md` — that directory does not exist yet, and ADR 0016 line 71 already names `docs/ci-cd.md` as G3's destination until #59 moves the tree. The literal path from the ticket would have been a dead link.
  **One correction to the ticket's framing:** `release.yml` publishes `@actana/sdk` and `@actana/cli` with `npm publish --provenance` and verifies the SLSA attestation before the release succeeds. A flat "nothing is signed" is contradicted by the release itself, so the section states the npm exception and bounds it — provenance attests where a package was built, not the tarballs and not the images.

- [x] **Records *why that is safe*: the `curl | bash` path never sets `com.apple.quarantine`, so Gatekeeper never intervenes. Revisit only if a browser-download distribution is ever added.**
  Present, and framed as a claim about the transport rather than the bytes, with the browser case named as the trigger to revisit.

- [x] **Records the corollary that already cost one review cycle: do not add `xattr -dr com.apple.quarantine` to `actana setup`.**
  Present with the reason: on the paths `actana setup` runs it clears nothing, and anywhere else it silently makes a Gatekeeper decision on the operator's behalf. The manual strip stays in `INSTALL.md`, where it is already documented for a browser download.

- [x] **`install.sh:295` cites `docs/contribute/ci-cd.md` instead of "spec".**
  The line had drifted to **`install.sh:326`** on this branch. It now cites `docs/ci-cd.md` and the section by name — same deviation and reason as the first box. Comment only; `sh -n install.sh` is clean and no test asserts on that comment.

- [x] **The supply-chain follow-up ticket is cross-linked, so the paragraph reads as a current posture rather than a permanent refusal.**
  #24 ("Supply chain: sign the release, publish an SBOM and provenance"), confirmed open via `gh api`. The section closes on ADR 0010's deferral → this answer → #24 as where it changes.

### Not a criterion, done anyway
- [x] `docs/core-macos-prerelease-checklist.md` deferred the same fact to #55. Since this PR closes #55, that link now points at the section instead of at a closed issue.

### Explicitly not done
- [ ] **`packages/core/src/actana-update.ts:137` still reads `spec: "No code signing"`** — the identical dangling citation, in **source**. Left alone deliberately: this PR is documentation only. It is one of the ten references #58 repoints; flagged here so it is not lost.
- [ ] **No ADR amendment.** ADR 0016 needs none — line 71 and the G3 row already anticipate the pre-#59 path. The ticket's footer clause is satisfied by not needing it.
- [ ] **#56 (ADR 0007 domain correction) and #57 (`repo-template/` out of the clone)** — untouched, per their parallel implementations.

### Link verification
All 35 relative links in the two changed docs resolve (path exists; `#anchor` present as a heading), and all 7 in-page anchors in `docs/ci-cd.md` resolve — checked with a script over the worktree, not by eye. #24, #55 and #58 checked with `gh api repos/actana/control/issues/N`.

### State
PR #256 — https://github.com/actana/control/pull/256 — **draft**, base `chore/improvement-sweep`. Nothing merged; `main` and the beta branches untouched. Both commits pass `commitlint --from origin/chore/improvement-sweep --to HEAD` (0 problems each).

mrfqcentic (changes requested):
Reviewed against #55, ADR 0016 D46, and — as the ticket's standard demands — against the release workflows themselves rather than against the prose. Four of the five acceptance boxes are clean, the deviation is correctly justified, and every link the change touches resolves. One sentence does not survive the audit, and it is the only one making a first-person claim about the pipeline.

### Blocking

**`docs/ci-cd.md:234-236`**

> `release.yml` carries no signing secret at
> all; the only credentials on the release path are `github.token` for the assets
> and the Docker Hub PAT for the mirror.

Two problems in one clause, both checkable against `release.yml` as it stands on this branch:

1. **`NPM_TOKEN` is a third credential, and it is mandatory.** `release.yml:349-358` is a `resolve` step named "Require an npm token" that hard-fails the whole release when it is unset ("NPM_TOKEN must be set — npm is a release registry alongside Docker Hub"), and it is consumed as `NODE_AUTH_TOKEN` at `release.yml:740` and `:812`. So "the only credentials … are github.token and the Docker Hub PAT" is false. The section contradicts itself three paragraphs later ("the npm packages are published with `--provenance`"), and contradicts `docs/ci-cd.md:759-760` in the same file ("It authenticates with `NPM_TOKEN`, and a missing one fails in `resolve`"). The clause reads as if it were lifted from `release.yml:67-69`'s header comment, which predates #159 adding npm and is stale for exactly this reason — which is the one failure mode this ticket exists to avoid.

2. **"the mirror" is the wrong noun for Docker Hub.** This file already fixes the vocabulary: `docs/ci-cd.md:722` — "**The only registry any image goes to**" — and `:848` uses "mirror" for the thing that deliberately does *not* exist ("**There is no mirror configured, deliberately**", about the `gcr.io` base). Calling the image registry "the mirror" in a section that asks to be read as load-bearing is a needless collision.

The load-bearing half of the sentence — `release.yml` holds **no signing secret** — is true and worth keeping; note that the one signing-adjacent capability in the workflow is `id-token: write` at `release.yml:683`, which is keyless (Sigstore/OIDC) rather than a secret, so the "no signing secret" phrasing is accurate but is doing more work than it looks. A correction naming all three credentials, or narrowing the claim to the assets and the images, fixes it.

### Verified, and correct

- **What the release actually produces.** Three Core tarballs plus `SHA256SUMS`, asserted at an explicit four-asset count (`release.yml:984-988`), composed and re-verified with `sha256sum -c` (`:933-938`). Exactly two images — the `panel` and `core` jobs at `:601` and `:621`. Built with plain `docker build` (`container-image.yml:405`, `:460`) and stitched with `imagetools create`, so no BuildKit provenance either. `grep -rniE 'cosign|notariz|codesign|authenticode|gpg|sigstore'` across every `.yml`, `.ts`, `.mjs` and `.sh`: no signing tooling anywhere. **The new paragraph is true of that pipeline, not of an intention** — with the credential clause above as the exception.
- **The npm correction is right, and correctly bounded.** `--provenance` at `release.yml:765`, read back off the registry at `:841` before the workflow succeeds. Stating the exception rather than a blanket "nothing is signed" is the right call, and "provenance attests where a package was built, not the tarballs and not the images" is accurate.
- **Both verifiers, from their call sites.** `install.sh:297-311` fetches `SHA256SUMS` *before* the tarball (the "Checksums first" comment is explicit about why), compares at `:315`, dies on mismatch. `packages/core/src/actana-update.ts:123-133` does the same. "Verified by both" is earned.
- **Links.** Resolved all 44 relative links across both changed docs, not a sample: every path exists and every `#anchor` matches a real heading, including the two the change introduces — `#npm` (→ `### npm`, `docs/ci-cd.md:754`) and, from the checklist, `ci-cd.md#integrity-is-published-checksums-not-signatures` (→ the new heading at `:229`, slug matches). `../INSTALL.md` and `../packages/core/src/actana-update.ts` resolve; `INSTALL.md:263-275` really is the `xattr` escape hatch the section defers to; ADR 0010 line 28 really does defer signing to "Harness distribution planning", quoted faithfully. `#24` checked on the web as well as via the API: **open**, and titled "Supply chain: sign the release, publish an SBOM and provenance". No dead link is introduced.
- **The path deviation is right and needs no ADR amendment.** ADR 0016's G3 row says it outright: "`docs/contribute/ci-cd.md` … *Landed by #38 in `docs/ci-cd.md`, which is that file before #59 moves the tree*", and line 71 repeats it. `docs/contribute/` does not exist; writing D46's literal path would have shipped the dead link this ticket removes.
- **`install.sh:295` → `:326` drift is real** (on the base branch line 295 is inside the `trap` block; the citation is at `:326`), and the edit is **comment-only**: stripping comments from both revisions leaves the files identical, `sh -n` is clean, and no test in `scripts/__tests__/install-sh.test.mjs` asserts on it.
- **Documentation only.** Three files: `docs/ci-cd.md`, `docs/core-macos-prerelease-checklist.md`, `install.sh` (comment). No workflow, no source, no mode change.

### Non-blocking

- `packages/core/src/actana-update.ts:137` still reads `spec: "No code signing"` — disclosed in the PR body, and correctly deferred: #58 is literally "Delete the 63 triaged files and repoint all ten references". Flagging only to confirm the deferral is agreed, not overlooked.
- `release.yml:67-69`'s header carries the same stale credential claim this review blocks on. Out of scope here; worth a follow-up so the two do not disagree again.

Acceptance criteria: D1 met (with the factual defect above inside it), D2 met, D3 met, D4 met, D5 met. Everything else in this change is solid — fix the credential clause and I will approve.

qcentic-dev:
Fixed in 47edef1 — the credential clause only, `docs/ci-cd.md:234-240`. The rest of the paragraph, and the rest of the section, are untouched.

I re-derived the credential set from the workflows rather than from the review. It is three, and all three are on the release path:

| Credential | Proof it is on the path |
| --- | --- |
| `github.token` | `release.yml:962` `- name: Attach to the GitHub Release`, `GH_TOKEN: ${{ github.token }}` at `:964` — the four Release assets. |
| `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` (one PAT) | Gated at `release.yml:325` `- name: Require Docker Hub credentials` (`exit 1` at `:335` when either is unset); consumed by `docker login` in `container-image.yml:357-358`, `:599` and `:647`, wired at `:335-336`, `:574-575`, `:629-630`. `release.yml:604` and `:624` call that workflow with `secrets: inherit`. |
| `NPM_TOKEN` | Gated at `release.yml:349` `- name: Require an npm token` (`exit 1` at `:358`: "NPM_TOKEN must be set — npm is a release registry alongside Docker Hub"); consumed as `NODE_AUTH_TOKEN` at `:740` (publish) and `:812` (the attestation read-back). |

So your count and your line numbers both hold. You were right about the provenance too: `NPM_TOKEN` was the omission, and the old pair was inherited from `release.yml:67-69`'s header comment, which predates #159.

What the clause says now:

> `release.yml` carries no signing secret at all — its one signing-adjacent capability, `id-token: write` on the `npm` job, is keyless OIDC rather than a key. Three credentials sit on the release path and none of them signs anything: `github.token` for the Release assets, the `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` PAT for the images, and `NPM_TOKEN` for the packages. The last two are each mandatory, gated by their own `resolve` step that fails the release before anything is built when they are unset.

Three changes, all inside the clause you blocked on:

1. **All three credentials named**, with the two `resolve` gates recorded so "mandatory" is not a bare assertion. This also removes the contradiction with `docs/ci-cd.md:759-760` and with the `--provenance` paragraph three below.
2. **"the mirror" → "the images."** Agreed on the collision — `:722` is "The only registry any image goes to" and `:848` uses "mirror" for the `gcr.io` copy that deliberately does not exist.
3. **`id-token: write` made explicit** (`release.yml:683`), since you flagged that "no signing secret" is accurate but load-bearing. Naming it keyless is what keeps the claim from reading as an oversight — it is the one thing in the workflow a reader would expect to find a key behind.

Non-blocking items, both acknowledged and neither touched here:

- `packages/core/src/actana-update.ts:137` — deferral to #58 confirmed, not overlooked.
- `release.yml:67-69`'s header carries the same stale pair. Out of scope for a docs PR, and it is now the only place in the repo that still says two. Happy to open the follow-up if you would rather not have it ride along.

PR is still **draft**, base still `chore/improvement-sweep`, still `MERGEABLE` after #251/#252/#253 landed on the base. `commitlint --from origin/chore/improvement-sweep --to HEAD`: 0 problems.

mrfqcentic (approved):
Re-reviewed `47edef1` against the one blocking finding only. Everything else in my previous review I had already verified and am not reopening.

### The blocking finding is met

`docs/ci-cd.md:234-240` now reads:

> `release.yml` carries no signing secret at all — its one signing-adjacent capability, `id-token: write` on the `npm` job, is keyless OIDC rather than a key. Three credentials sit on the release path and none of them signs anything: `github.token` for the Release assets, the `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` PAT for the images, and `NPM_TOKEN` for the packages. The last two are each mandatory, gated by their own `resolve` step that fails the release before anything is built when they are unset.

Checked against the workflows on this head, not against the summary:

- **The count is three, and it is exhaustive.** `grep -nE 'secrets\.|github\.token|GH_TOKEN|NODE_AUTH_TOKEN' .github/workflows/release.yml` returns exactly three distinct credentials: `secrets.DOCKERHUB_USERNAME` / `secrets.DOCKERHUB_TOKEN` (`:327-328`), `secrets.NPM_TOKEN` (`:351`, `:740`, `:812`), `github.token` (`:964`). The only reusable workflow the release calls is `container-image.yml` (`release.yml:604`, `:624`, `secrets: inherit`), and that file's entire credential surface is the same Docker Hub pair (`:149-150`, `:335-336`, `:574-575`, `:629-630`). There is no fourth credential to have missed.
- **Each attribution is right at its call site.** `github.token` is on `- name: Attach to the GitHub Release` (`release.yml:962`, `GH_TOKEN` at `:964`) — the Release assets, as claimed. `NPM_TOKEN` is `NODE_AUTH_TOKEN` at the publish (`:740`) and at the attestation read-back (`:812`) — the packages. The Docker Hub pair is consumed only by `docker login` in `container-image.yml` — the images.
- **"Mandatory, gated by their own `resolve` step" is exact.** Both gates are steps inside the `resolve:` job (`release.yml:157`): `- name: Require Docker Hub credentials` at `:325`, `exit 1` at `:335`; `- name: Require an npm token` at `:349`, `exit 1` at `:358`. Neither is conditioned on the repo or on a fork — on `release.yml` a missing one fails unconditionally, so "mandatory" is earned, and both sit in `resolve` before any build job, so "before anything is built" is earned too. `github.token` is correctly *not* described as gated: it is the automatic token and needs no check.
- **`id-token: write` is where the sentence says it is and is what it says it is.** `release.yml:683`, in the `permissions:` block of the `npm:` job (`:675`, `name: npm publish`). Keyless OIDC, no key material. Adding it was the right call — it is the one place a reader would go looking for a signing key, and naming it is what keeps "no signing secret" from reading as an oversight.
- **The contradictions are gone.** The clause now agrees with `docs/ci-cd.md:759-760` ("It authenticates with `NPM_TOKEN`, and a missing one fails in `resolve`…") and with `:1117-1119`, which independently names all three as the ones `resolve` checks. "the mirror" → "the images" resolves the collision with `:722` and `:848`.
- **Scope of the revision.** `git diff d3a35ce..47edef1` is `docs/ci-cd.md` alone, +6/-2, entirely inside the clause I blocked on. `install.sh` and `docs/core-macos-prerelease-checklist.md` are byte-identical to the revision I already verified, so nothing I passed last round needs re-checking.

### The `release.yml:67-69` follow-up: accepted, not blocking

Accepted as a follow-up. It is out of scope for #55 — D46 is the no-code-signing paragraph and the `install.sh` citation — the staleness is pre-existing rather than introduced here, and I called it non-blocking last round on the same reasoning; reversing that now would be moving the goalposts. Confirmed it really is the last live occurrence: the only other hit is `.scratch/harness-installer/issues/01-per-platform-tarballs-checksums.md:21`, a file #58 deletes, and it is a different claim about a workflow that does not exist.

One correction to the reasoning, though, since it should not be load-bearing for the next one of these: **the CI-cost argument does not hold on this PR.** `ci.yml`'s documentation-only test (`pr-image-mode`, `:810-814`) matches `(\.md$|^docs/|^\.scratch/|^designs/|^landing/)`, and this PR changes `install.sh`, which matches none of them. `docs_only` is already `false` and the image jobs already run here — editing a workflow file would have added no check surface. The follow-up is right on scope alone; it did not need the cost claim.

Worth filing rather than dropping, and please link it from #58: that header is the exact source the two-credential pair was originally copied from, so leaving it uncorrected is a live path for the same error to be re-propagated by the next reader.

### Acceptance criteria

D1 met — and now met without the factual defect. D2, D3, D4, D5 met, unchanged from the previous review. The path deviation to `docs/ci-cd.md` remains correct and still needs no ADR amendment. The deferral of `packages/core/src/actana-update.ts:137` to #58 is agreed.

Approving.

PR #259, opened 2026-08-18
chore(release): promote 0.3.1

mrfqcentic:
This is the ADR 0023 promotion gate for the 0.3.1 train. Per D5 and D16, promotion is consumed through this pull request: its checks and approval are the gate, and its head SHA is what the beta digest is verified against. Merging it promotes beta/0.3.1 into main.

Tickets carried by this train:

- **#243** — settle Core-owned Sessions
- **#244** — advance the live-event cursor only after a send lands
- **#242** — key the Session drive register by client id
- **#186** — reconcile pane events with tab interest
- **#208** — settle the tip hunt in events-tail-cursor

The train was stamped 0.3.1 across all six manifests.


qcentic-adm (approved):
Approved for promotion: the 0.3.1 train is cleared to go to main under the ADR 0023 gate.

PR #262, opened 2026-08-18
chore: 0.3.2 improvement sweep - docs, contracts and repo hygiene

qcentic-dev:
Replaces **#258**, which was auto-closed when `beta/0.3.2` had to be deleted and re-cut.

The original train carried two cut commits (`fc946ca`, `ca9912a`) whose message bodies exceeded the 132-column `body-max-line-length`, so the Conventions check failed on the promotion PR. The train is now a single clean commit — `aed430f chore(release): cut beta/0.3.2` — sitting directly on `main` at `3b14453`, bumping exactly six manifests to 0.3.2.

This branch is the **same six tickets**, cherry-picked one by one onto the re-cut train. The old branch `chore/improvement-sweep` is untouched at `1a99db8` as a safety net; neither bad cut commit nor the merge that carried them is in this history.

## Commit mapping

| # | Original (on `chore/improvement-sweep`) | New (here) | Title |
|---|---|---|---|
| 1 | `cdad08e` | `2611361` | fix(core): keep a separator where sanitizeInitialInput drops a line break (#251) |
| 2 | `dc7be56` | `2a757c7` | refactor(sdk): give CoreFilesErrorCode a single owner (#253) |
| 3 | `59c56ae` | `675340b` | docs(release skill): describe the ADR 0023 trigger and where macos-release sits (#252) |
| 4 | `d964029` | `9295381` | docs(adr): correct the update host domain in ADR 0007 and drop the docs/specs pointers (#254) |
| 5 | `2a7ad5a` | `fa36d20` | docs: record why the release is unsigned, and repoint the citations that said so (#256) |
| 6 | `8d7a64b` | `5aa10ca` | chore(repo): record that repo-template/ is absent from this clone (#255) |

Commit 6 is a deliberately empty commit recording issue #57 — it carries no file changes and was preserved with `--allow-empty --keep-redundant-commits`.

## Verification

- All six cherry-picks applied **without conflict**.
- `git diff --name-only beta/0.3.2..HEAD` → exactly **19 files**, **no `package.json`** among them.
- `git diff beta/0.3.2..HEAD -- '*package.json'` → empty. The manifests stay exactly as the cut left them at 0.3.2.
- No conflict markers anywhere in the tree.
- Commit messages: longest body line is **81** chars (limit 132); longest subject is **93** (limit 120). `commitlint --from beta/0.3.2 --to HEAD` against this repo's own `commitlint.config.mjs` exits **0** — 0 problems, 0 warnings.

Left as a **draft**.

PR #254, opened 2026-08-18
docs(adr): correct the update host domain in ADR 0007 and drop the docs/specs pointers

qcentic-dev:
Closes #56. Part of #23. Base is `chore/improvement-sweep`.

ADR 0007 said the disabled update-manager stub points at `control.actano.ai`.
The only thing in the repo that resolved that typo was spec 09's Open Questions
(*"`actana.ai` is authoritative"*), and `docs/specs/` is slated for deletion — so
deleting it first would have left a live ADR carrying a wrong domain with nothing
to correct it. This fixes the domain and, per the ticket, also strikes the two ADR
pointers into `docs/specs/`.

## Every occurrence of the misspelling in the repo

`grep -rn -i actano .` over the whole worktree (excluding `.git/` and
`node_modules/`) — **12 hits in 4 files, all under `docs/`. Zero hits outside
`docs/`**: no source file, config, workflow, test fixture, or `package.json`
contains it.

### Changed (4)

| File:line | Text | Why changed |
| --- | --- | --- |
| `docs/adr/0007-scope-narrowing-and-rebrand.md:26` | `control.actano.ai` | The ticket's target. A live ADR consequence stating the update host. |
| `docs/specs/09-rebrand-and-auto-update.md:10` | `control.actano.ai` | Same domain, prose. |
| `docs/specs/09-rebrand-and-auto-update.md:26` | `control.actano.ai` | Same domain, prose in the `package.json` table's *note* column (not a value — the value in that row is `"publish": null`). |
| `docs/specs/09-rebrand-and-auto-update.md:182` | `control.actano.ai` | Same domain, inside a fenced block quoting a prospective comment for `electron/update-manager.ts`. Flagging the judgement call: it is a code block, but it is prose-in-a-comment, the file no longer exists (Electron left with ADR 0010), and nothing reads it. Say the word and I will revert this one line. |

### Found and deliberately NOT changed (8)

**A proposed bundle identifier, out of the ticket's reach (3 hits).** The string
`ai.actano.control`, offered as a candidate `build.appId`. It is *not* a live
value: there is no `appId` anywhere in this repo — no `build` block in
`package.json`, no `electron/`, no `build/` — only a proposal in a spec for the
packaging config that ADR 0010 removed. It is left alone for two reasons that do
not depend on that:

- It does not contain the string `actano.ai`, so AC1 never reached it; and D46/K4
  is narrower still — *"fix `actano.ai` in ADR 0007"*, one domain, one file.
[...]

PR #263, opened 2026-08-18
chore(release): promote 0.3.2

mrfqcentic:
This is the promotion gate for 0.3.2.

It promotes the `beta/0.3.2` train onto `main`. The train contains the release cut
(`aed430f`) and the squashed 0.3.2 improvement sweep (`6cd15ca`, PR #262, 19 files
across six tickets).

## Do not merge this pull request by hand

`promote.yml` performs the promotion as a **fast-forward** of `main` to the tip of
`beta/0.3.2`, preserving the exact commit SHAs on the train.

Pressing the **Merge** or **Squash** button in the GitHub UI instead creates a *new*
commit with a *new* SHA. That breaks the ADR 0023 D16 revision-label assertion, which
pins the released revision label to the SHA carried on the train. This is precisely
what made 0.3.1 unreleasable and forced that train to be re-cut.

Let the workflow do the fast-forward.

## Pre-flight

- Longest commit body line on `origin/main..origin/beta/0.3.2`: **95** characters
  (Conventions commitlint limit is 132).


qcentic-adm (approved):
Approving. Verified against the refs and the enforcing lines, not the description.

**1. Fast-forward, two commits, no merge.** `origin/main` is `3b14453`, `origin/beta/0.3.2` is `6cd15ca`, and `git rev-list --left-right --count` is `0  2`. The chain is linear — `3b14453 <- aed430f <- 6cd15ca`, each with a single parent — `git rev-list --merges` over the range is empty, and `merge-base(main, beta/0.3.2)` is `3b14453` itself. `main` is an ancestor, so D5's fast-forward is genuinely available and no merge commit is needed.

**2. Conventions is green on its merits.** The rule is `body-max-line-length: [2, 'always', 132]` (and the same for the footer) in `commitlint.config.mjs`, severity 2. The longest line in the whole range is 95 characters — `* docs(adr): correct the update host domain in ADR 0007 and drop the docs/specs pointers (#254)`, line 234 of the sweep body — leaving 37 characters of headroom; `aed430f`'s longest is 64. Headers are 72 and 30 against the 120 limit. The check log shows commitlint invoked with `--from 3b14453… --to 6cd15ca…` and reporting `found 0 problems, 0 warnings` for both commits, so both were actually linted. `Lint the PR title` is `skipped`, which is the deliberate promotion-PR exemption (D5/D38), not a silent pass.

**3. Six manifests at 0.3.2, asserted against this train.** From the `Train rules` log, not its status:

```
Base: main
Head: beta/0.3.2
Promotion pull request for 0.3.2.
  ✅ the version set is 6 manifests, and covers every workspace package
  ✅ package.json is 0.3.2
  ✅ packages/cli/package.json is 0.3.2
  ✅ packages/core/package.json is 0.3.2
  ✅ packages/panel/package.json is 0.3.2
  ✅ packages/sdk/package.json is 0.3.2
  ✅ packages/shared/package.json is 0.3.2
  ✅ Branch model satisfied.
```

The expected version is derived from the head branch name, so it is asserted against *this* train. Confirmed independently by reading the six files at `6cd15ca`, and `packages/*/package.json` is exactly the five in the list, so the set-coverage guard has nothing outside it.

**4. D16 holds — the digest carries this PR's head SHA.**

```
docker.io/actana/core:beta-0.3.2  is actana/core@sha256:3c5fafa97e1896fcb934ea6695930735cb8e4467a8ff2cc39339804f2b629dc3
  revision label: 6cd15ca2890b45f12f92e45c29a00ae933aa470f
  expected:       6cd15ca2890b45f12f92e45c29a00ae933aa470f

docker.io/actana/panel:beta-0.3.2 is actana/panel@sha256:715014077c09205cd78027ce1e640124690893b06193a74fb12b41ab3be7104c
  revision label: 6cd15ca2890b45f12f92e45c29a00ae933aa470f
  expected:       6cd15ca2890b45f12f92e45c29a00ae933aa470f
```

Both jobs ran in `mode: verify` and both ended `✅ … carries 6cd15ca… — the promoted digest is the tested one`.

The re-run is the authoritative one, and the first failure was a real race rather than flakiness. Attempt 1 read `beta-0.3.2` at 19:10:31 and got `sha256:3c35b0d5…` with `revision label: aed430f8…` — the *cut* commit, because the train's `Publish the multi-arch manifest` had not run yet. That manifest pushed `sha256:3c5fafa9…` to `actana/core:beta-0.3.2` at 19:11:41; attempt 2 read it at 19:16:57 and got exactly that digest. The `Train rules` log names the same race from the other side (`beta/0.3.2's latest push is in_progress`). Attempt 2 is the latest attempt and is what the required checks resolve to — every check run on `6cd15ca` is now `success` or `skipped`.

One scope note, not a blocker: the PR-gate verify matrix is pinned to amd64 (`ci.yml:978`, `:1017`), so D16 is asserted here for `linux/amd64` only. The arm64 legs of the train build do carry the label (`Labelling with org.opencontainers.image.revision=6cd15ca…` in both arm64 jobs), and `release.yml` calls `container-image.yml` with no matrix override, so promote mode re-asserts every architecture before any retag.

**5. 25 files = 19 + 6, nothing unexpected.** The cut `aed430f` touches exactly the six manifests, one line each, `0.3.1 → 0.3.2`. The sweep `6cd15ca` touches exactly the 19 ticket files. Nothing under `.github/` changed — no workflow, no ruleset doc, no `commitlint.config.mjs`, no Dockerfile. `git diff --summary` shows four new regular files (`100644`) and no mode changes, symlinks or submodules.

---

**Before dispatching `promote.yml`: `beta/0.3.1` still exists on origin and will change what the workflow does.**

`beta/0.4.0` does not exist — no branch or ref matches it. But `beta/0.3.1` is still there at `d438e5c`. `resolve` computes `all_trains` from `ls-remote refs/heads/beta/*` and filters out the train being promoted, so `survivors` is `beta/0.3.1`, `survivor_count` is 1, and **`hotfix=true`** — even though this is an ordinary promotion. Two consequences:

* `next-train` is gated on `hotfix == 'false'`, so it is **skipped**. No `beta/0.4.0` is cut, and the repo is left with the abandoned `beta/0.3.1` as its only open train.
* `rebase-train` is gated on `hotfix == 'true'`, so it **runs**: it force-pushes `beta/0.3.1` rebased onto the new `main` and republishes its images. `main` is not an ancestor of `d438e5c` (they diverged 1/2 at `da2af3e`), so it will not take the "nothing to rebase" exit and will attempt the rebase. That rebase is very likely to conflict: `54e7646` patches the six manifests `0.3.0 → 0.3.1`, and on the new `main` they read `0.3.2`, so the context will not match. Worth noting that `main`'s tree and `beta/0.3.1`'s tree are already byte-identical (`920a12a…`) — the squash landed its content in full, so there is nothing on that branch to preserve.

`rebase-train` runs after `advance` and `release`, so a failure there leaves `main`, `v0.3.2` and the release correct and only that branch broken — but it is an avoidable red run. Promotion is `workflow_dispatch` only (approval triggers nothing), so deleting `beta/0.3.1` before dispatching costs nothing and gets the ordinary path: `hotfix=false`, no force-push, and `beta/0.4.0` cut automatically.

Issue #265, opened 2026-08-19
skills: ship a generic Cores-and-Sessions orchestration skill inside @actana/cli, installed into every Harness — relitigating ADR 0006 first


mrfqcentic:
> **For the `0.3.3` train.** Orchestrating Cores is currently tribal knowledge in a
> private, machine-local skill directory on one operator's laptop. Nothing about
> driving Cores and Sessions ships with the product, and a Core cannot orchestrate
> its own Sessions at all — *"give me three additional sessions, each focusing on
> three different things"* has no answer in this repository today.

This issue asks for a **generic orchestration skill**, authored once here, published
inside the `@actana/cli` npm package, and installed into every supported Harness's
global configuration directory, mandatorily, on an explicit and idempotent code path.

Every claim in the brief this was filed from was checked against the tree at
`e6d2fd9` (`beta/0.3.3`) before being repeated here. **Most of it holds and is
stronger than reported.** Four things do not hold as written and are corrected in
place rather than asserted: §2 (the global paths are not in this repository), §4
(there is no install hook at CLI-install time, because the brief also forbids the
only one npm offers), §5 (the availability event is a full map on a 60s tick, not a
delta within seconds), and §6 (the CLI and the Core cannot share a module, so
"authored once" needs a mechanism). And one thing blocks the whole issue: **ADR 0006
forbids exactly this, by name.** That is §1, and it comes first because nothing else
here is mergeable until it is settled.

---

## 1. ADR 0006 forbids this in the words of the proposal — VERIFIED, and it is the gate

[`docs/adr/0006-no-bundled-skills.md`](https://github.com/actana/control/blob/main/docs/adr/0006-no-bundled-skills.md),
first paragraph:

> Actana Control ships no bundled agent skills and installs nothing into the
> operator's `.claude/skills/`, `.codex/skills/`, `.cursor/skills/`, or any equivalent
> harness skill directory. […] The Panel does not write to any harness config path
> outside its own registry storage. Skills, MCPs, hooks, and CLAUDE.md-style config
> are the operator's territory.

and its last consequence:

> Any future capability that would require Panel-installed skills must first justify
> itself against this ADR.

It is not a stale record. It was enforced by a whole spec —
`docs/specs/05-remove-bundled-skills.md` deleted `dist/bundled-skills/`, three
`ensure-*-skill` modules, two install services, a controller, six UI components and
`scripts/copy-bundled-skills.mjs` — and `docs/specs/05-remove-bundled-skills.md:210`
[...]

Issue #257, opened 2026-08-18
ci: make a red run readable — stop cross-package bail and disk pressure from masking real failures


qcentic-adm:
A red CI run on this repository cannot currently be read. It can only be re-run.
This ticket is the 0.3.3 CI-reliability round, and the ask is that it lands before
the cursor-cli work in #177 and #232 — debugging a TUI boot race against a pipeline
that misreports which tests ran is where that time goes.

Each symptom below was checked against this repository and against the workflow runs
of **2026-08-18** before filing. Two of the five did not survive the check in the form
they were reported: one was **dropped entirely**, one was **cut back to the part that is
true**. Both are marked. Nothing here is asserted from memory.

---

## 1. `pnpm -r` bails on the first failing package, so one red package hides three — VERIFIED

**This is the one that matters.** The other items are ordinary flakes; this one is why a
flake costs a whole run.

The `Unit Tests` job runs `pnpm test` (`.github/workflows/ci.yml:456`), and the root script
(`package.json:47`) is:

```json
"test": "pnpm native:node && vitest run && pnpm -r test"
```

`pnpm -r` bails on the first failing package by default, and it runs the workspace in
dependency order, which on this graph is **sdk → shared → cli → core → panel** (confirmed
from a green run's log, [32132593072](https://github.com/actana/control/actions/runs/32132593072)).
`packages/cli` depends only on `@actana/sdk`; `packages/core` additionally depends on
`@actana/shared`, so cli is always scheduled ahead of core — this is topological ordering
plus the workspace-directory tie-break, not simple alphabetisation, but the effect the
report described is exactly right: **cli runs first, and when it fails, core and panel
never run at all.**

Three runs today ended this way, all with the same last four lines:

| Run | Branch / PR | Last package reached | core ran? | panel ran? |
|---|---|---|---|---|
| [32127897727](https://github.com/actana/control/actions/runs/32127897727) | `beta/0.3.1` (push) | `packages/cli` | no | no |
| [32129462705](https://github.com/actana/control/actions/runs/32129462705) | #246 `fix/panel-drive-register-client-id` | `packages/cli` | no | no |
| [32130035471](https://github.com/actana/control/actions/runs/32130035471) | #247 `fix/core-session-settle` | `packages/cli` | no | no |

```
packages/cli test:  Test Files  1 failed | 29 passed | 1 skipped (31)
[ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL] @actana/cli@0.3.0 test: `vitest run`
[...]

Issue #177, opened 2026-08-11
cursor-cli is undrivable from the CLI client: wrong binary, auto mode never sent, trust dialog unanswered, no turn-start signal


mrfqcentic:
`cursor-cli` cannot be driven from the experiment CLI client. It fails at four
separate points, of which only the first is fatal at spawn time — the rest are
silent, and each one only becomes visible once the previous is cleared.

The through-line: the Panel has a correct per-harness launch builder
(`packages/panel/src/lib/harness-command.ts`). The experiment client has
`buildClaudeCommand` and nothing else, so every place cursor-agent differs from
Claude Code is a place the client assumes it does not.

Where the code lives is split, and the split matters for whoever picks this up:

| Half | Where | Findings |
| --- | --- | --- |
| client | `experiment/src/commands/session.ts` — **out of this repo**, unversioned scratch tooling | 1, 2, 3, 5 |
| repo | `packages/shared`, `packages/panel` | 4, and the enabling defect under 2 |

Filed here because the constraints the client violates are owned by this repo,
and because this is the only repository in the org. If `experiment/` is ever
brought under version control, the client half moves with it.

---

## 1. Spawn rejected: the client sends the harness id, the Core expects the binary

```
$ node core-client.ts --core personal session start p-msonxkl9-f3ea7c --harness cursor-cli --model composer-2.5
core-client: pty:spawn rejected (command-not-on-allowlist)
```

`packages/shared/src/pty-spawn-policy.ts` step 6 requires `argv[0]` to be the
harness's canonical binary:

```ts
if (tokens[0] !== expectedBinary) {
  throw new SpawnPolicyError("command-not-on-allowlist", "pty:spawn agent command is not allow-listed");
}
```

`expectedBinary` comes from `HARNESS_SPAWN_COMMANDS` in
`packages/shared/src/harness-cli-config.ts`:

| harness id | binary |
| --- | --- |
| `claude-code` | `claude` |
| `codex` | `codex` |
| `cursor-cli` | **`cursor-agent`** |
| `opencode` | `opencode` |

The client sends the *harness id* for anything that is not `claude-code`:

```ts
command: opts.harness === "claude-code" ? command : opts.harness,
```

`codex` and `opencode` survive only because their id equals their binary.
`cursor-cli` is the sole case where the two differ, and so the sole harness that
cannot launch. The coincidence is what kept this hidden.

## 2. Auto mode never reaches the harness

Per-harness auto-mode flags, from `HARNESS_ARG_RULES` in
[...]

PR #267, opened 2026-08-19
docs: put the release-train rules where an agent will actually read them

qcentic-dev:
## Summary

`AGENTS.md` is the file an agent opens first, and it said nothing about release trains — the rules lived in
`CONTRIBUTING.md`, `docs/ci-cd.md` and ADR 0023, none of it reachable from where an agent starts. This adds a
*Release trains* section in the shape of the four already there, states the two load-bearing rules literally rather
than only behind a link, adds the `CLAUDE.md` the repo never had, and gives `docs/ci-cd.md`'s "re-cut it" instruction
the referent it was missing.

Documentation only. No source, no workflows, no ruleset.

## Related issues

Refs #260

## Type of change

- [x] 📚 Documentation

## What changed

- **`AGENTS.md`** — a *Release trains* section. Both sentences criterion 1 asks for appear literally: *a train is
  cut by a workflow and never by hand*, and *promotion consumes an existing, green, approved pull request from the
  train into `main`*. It links to `CONTRIBUTING.md` §"Where your PR goes: the open train, not `main`",
  `docs/ci-cd.md` §"The train model" and §"Cutting a release", and ADR 0023. The file is 21 lines.
- **`CLAUDE.md`** — new, three lines, a pointer to `AGENTS.md`. There was none, so an agent looking for it by name
  found nothing.
- **`docs/ci-cd.md`** — a new `#### Re-cutting a train, by hand` under §"Cutting a train": no dispatchable cut
  workflow exists, and here is what to run instead — delete, branch from `main`, the six manifests edited in place
  (not round-tripped through `jq`, per `ca9912a`), the `git diff` that proves the cut is only the cut, and the `awk`
  that measures body lines against the 132-character limit before `commitlint` sees the commit at the promotion
  gate. Both `re-cut` occurrences (§"Cutting a train" and §"Hotfix trains") now point at it. Additive — nothing
  existing was rewritten or moved, so #264 and #265 can rebase onto this.
- **`CONTRIBUTING.md`** — one sentence: it said body lines are capped at 100; `commitlint.config.mjs` says 132.

## How was this tested?

- [x] Manually verified (steps below)

- Every link and anchor in the new `AGENTS.md` section resolved against the files **at `origin/main`** — headings
  slugified and matched, all four resolve.
- `docs/ci-cd.md` heading set is unchanged apart from the one new `####`, so no existing anchor moved.
- The documented `node` one-liner was run against a six-manifest fixture: six versions rewritten, formatting of the
  surrounding JSON untouched.
[...]

PR #268, opened 2026-08-19
feat: one Panel terminal control that opens a VM shell, and `actana core exec`

### LINKED ISSUE #266: Panel + CLI: one terminal control that opens a VM shell, and `actana core exec` for the non-interactive half
> **For the `0.3.3` train.** Two parts, one issue, because they are the same
> complaint from two ends: **getting a shell on a Core is harder and more confusing
> than it should be.** The Panel offers two buttons where one is right, and the CLI
> offers an interactive shell where a script needs a command.

Checked against `e6d2fd9` (`beta/0.3.3`). Every line number in the brief this was filed
from is exact today and is repeated below with what is actually on it — **but the file
will move**, so each citation names the symbol as well as the line. Three things in the
brief do not survive the check and are corrected in place: the Panel has **two**
`createTerminal` call sites and no third control to rename (§A.1), "the code path no
longer exists" is a much larger removal than two buttons (§A.3), and the non-interactive
verb needs a **new core-link frame** and therefore a protocol version move, which the
brief does not mention (§B.2). One claim I could not verify at all is marked as such
rather than asserted (§A.2).

---

# Part A — the Panel has two terminal buttons and only one is right

## A.1 What is actually there — VERIFIED, with a correction

`packages/panel/src/components/views/UserTerminalPanel.tsx` (521 lines) destructures both
creators from the store at `:48-49` and calls them in **three** places:

| Line | Where | Label the operator sees | Calls |
|---|---|---|---|
| `:326` | header toolbar, inside `HotkeyTooltip action="terminal.newTab"` (`:319`) | **"New"**, `icon="plus"` | `createTerminal({ coreId })` |
| `:402` | `EmptyState` action row (`:394`) | **"New terminal"**, `icon="plus"` | `createTerminal({ coreId })` |
| `:408` | same `EmptyState` action row, guarded by `coreId &&` | **"New VM shell"** (`VmShellButton`, defined `:478`) | `createVmShellTerminal(coreId)` |

**Correction to the brief.** It describes removing "the New Terminal button" and then
renaming "the remaining `+ New` control" to `+ New Terminal`. There is no such remaining
control: **the header's `+ New` at `:326` *is* the project-root path.** Both `+`-icon
buttons call `createTerminal`; they differ only in label ("New" in the toolbar, "New
terminal" in the empty state) and in which of the two places they render.

So the change is not "delete one, rename another". It is: **each of the two locations
keeps exactly one button, labelled "New Terminal", and that button calls
`createVmShellTerminal`.** The `VmShellButton` component and its second appearance in the
empty state go away; the panel ends up with one control per location and one behaviour.

**The accordion half already works and needs nothing.** Both creators end with
`setPanelOpenByProject((prev) => ({ ...prev, [key]: true }))` —
`user-terminal-store.tsx:368`, `:388`, `:424` for `createTerminal`, `:451` for
`createVmShellTerminal`. "Clicking it opens the accordion AND spawns" is already true of
both paths.

**A dead keybinding, found on the way through, and this issue should settle it.**
`terminal.newTab` is declared as an action (`lib/keybindings/types.ts:12`), given a label
and description (`:56`), bound to `mod+T` by default (`lib/keybindings/defaults.ts:18`),
listed in the terminal group (`lib/keybindings/groups.ts:33`) and **advertised to the
operator in both `HotkeyTooltip`s above** — but there is no
`useHotkey("terminal.newTab", …)` anywhere in `packages/panel/src`. The only two
terminal actions with handlers are `terminal.toggle` (`routes/__root.tsx:412`) and
`terminal.close` (`SessionGrid.tsx:1463`). **⌘T is shown in a tooltip and does nothing.**
Either wire it to the surviving control or drop the action; leaving it advertised and
inert is the same class of confusion this issue exists to remove.

## A.2 "The project-root path does not work" — NOT VERIFIED, and it does not need to be

I could not confirm this from a static read, and I am not going to assert it. What can be
said precisely:

- The two paths take genuinely different spawn modes. A VM Shell Session spawns with
  `shellSession: true`, which `packages/shared/src/pty-spawn-policy.ts:63` documents as
  *"its own spawn mode: no `agent`, no project-root"* — and CONTEXT.md's **VM Shell
  Session** entry says it is *"gated by core-link auth (not project-root validation)"*.
  The project terminal goes through `projectRoots()` confinement
  (`pty-spawn-policy.ts:380`). So there is a real validation step on one path that the
  other does not have, which is where a failure of this shape would live.
- `createTerminal` has **two** branches, not one: a project-less **home** branch
  (`user-terminal-store.tsx:361-370`, `api.createHomeTerminal` at `:363`) and the project branch
  (`:371-428`, `api.createUserTerminal` with `cwd = targetProject.path`). Only the second
  is "the project-root shell". The home branch shares `createHomeTerminal` with the VM
  shell and differs from it only by the in-memory `shellSession` flag (`:448`).
- There is **no Panel test** covering either creator — no test file in
  `packages/panel/src` references `createTerminal`, `createVmShellTerminal` or
  `UserTerminalPanel`. Whatever is broken, nothing would have caught it.

**This does not block the change.** The decision stands on its own without the bug:
*you are in a Core, so a terminal means a shell in that Core; there is nothing else it
could sensibly mean.* That is the rationale to implement against. If someone reproduces
the failure, add the repro here — it strengthens the case but is not load-bearing for it.

## A.3 "The code path no longer exists" is a bigger removal than two buttons

The brief asks for removal, not deprecation, and is right to. But the project-root
terminal is not two call sites — it is a feature with a table, a repository, a service,
a controller, three routes, a client store branch, and a **pre-spawning warm pool**:

- `packages/panel/src/db/schema.ts:143-157` — the `user_terminals` table and its index
- `packages/panel/src/server/repositories/user-terminals.repo.ts`
- `packages/panel/src/server/services/user-terminals.ts:24` — `createUserTerminal`
- `packages/panel/src/server/controllers/user-terminals.controller.ts:18,42` —
  `createTerminalBody`, `create`
- `packages/panel/src/server/api-router.ts:37,53` — `PROJECT_USER_TERMINALS_PATH`,
  `USER_TERMINAL_PATH`
- `packages/panel/src/lib/user-terminal-warm-pool.ts` — the whole module
- `packages/panel/src/lib/user-terminal-store.tsx:371-428` — the project branch,
  including the warm-slot fast path and its rollback

The warm pool is worth a sentence of its own: `prepareUserTerminalWarmSlot` fires on
project navigation (`user-terminal-store.tsx:286`), so **navigating to a project
pre-spawns a PTY on the Core** in anticipation of a click on the button this issue
removes. Leaving it in place would keep spawning PTYs nothing can ever claim. It goes
with the path.

**Two scoping decisions this issue has to make rather than discover:**

1. **The `home_terminals` table stays**, because the VM shell reuses it —
   `createVmShellTerminal` calls `api.createHomeTerminal` (`:443`) and the schema comment
   at `db/schema.ts:162-165` records why home terminals got their own table. Only the
   `user_terminals` half is in question.
2. **Does `user_terminals` get dropped, or left orphaned?** The brief rules out migrating
   rows and is right — a terminal is ephemeral. But "no migration for persisted rows" is
   not the same decision as "drop the table". Say which, once, here.

## A.4 What the operator ends up with

- One control per location, labelled **"New Terminal"**, `icon="plus"`, which opens the
  accordion and spawns a VM Shell Session on the Core the route is on.
- No "New VM shell" button and no `VmShellButton` component.
- `⌘T` either does that same thing or is not advertised.
- The `EmptyState` copy at `:387-393` updated: it currently says *"Open a terminal to run
  commands in this project"* / *"Open a terminal at your home directory on the active
  scope"*, both of which describe the removed path.

**Note for whoever implements it:** CONTEXT.md's rule *"VM Shell Sessions are privileged
— same auth as the core-link; require an explicit open gesture in the Panel; never
auto-spawned"* is unchanged by this. A button click is an explicit open gesture; that is
what `VmShellButton`'s own doc comment at `:474-477` means by *"an explicit click, never a
side effect of navigating somewhere."* Making it the *only* button does not make it
automatic — but the removed warm pool did pre-spawn on navigation, so do not replace it
with a VM-shell warm pool.

---

# Part B — the CLI cannot run a command on a Core non-interactively

## B.1 The gap, and why it is the product's problem — VERIFIED

`actana core` has five verbs — `add`, `ls`, `use`, `rm`, `status` — plus `shell`
(`packages/cli/src/core-command.ts:88-104`). There is no non-interactive form of `shell`
and no `exec`.

**The reach-around this forces is one this repository has already argued against, in
writing, twice.** A maintenance script that needs to run a command on a Core has to use
`docker exec`, which works only because these Cores happen to be local containers, does
not work for a remote Core, and bypasses the Core's own authentication entirely. The
repository's position on that shape:

- `packages/cli/src/__tests__/no-local-escape.test.ts:1-10` — *"`actana core add` accepts
  a file or stdin. Never `docker exec` — a CLI that shells into a container to fetch its
  own credentials is not a CLI. A client that gets its credential by running a command on
  the Core's host only works when the Core is on this machine, which is the case that
  matters least; it makes a container runtime a dependency of a program whose whole
  purpose is to not need one."* (#129 D9, enforced by a sweep that fails on any
  `child_process` import in `packages/cli/src`.)
- `docs/reference/cli-parity-check.md:53-55` — four of `core`'s five verbs exist *"because
  registering a Core is now a thing the CLI does rather than a thing the operator does
  with `docker exec` and a redirect."*

Running a command on a Core is the same argument one verb further along. The test's own
header even anticipates the exception: *"The one verb that will legitimately want a
subprocess is `core shell` (#162), and it is the exception that proves the rule: it hands
the operator a terminal on the Core, over the core link, rather than running anything
locally on their behalf."* `core exec` is that same exception, minus the terminal.

## B.2 The shape, and the one consequence the brief does not name

```
actana core exec [--cwd <dir>] -- <cmd> [args...]
```

- non-interactive, no PTY
- stdout and stderr buffered and returned **separately**, free of terminal escape
  sequences
- propagates the command's **real** exit code, so a script can branch on it
- `--cwd` chooses the working directory
- honours the same Core selection as every other verb — `--core <name>`, then
  `ACTANA_CORE_BLOB`, then the `current` pointer (`packages/cli/src/core-resolution.ts:83-126`)

**Settled, with reasons:**

- **Not PTY-backed with screen-scraped output.** The whole point is to stop parsing
  rendered screens. A PTY merges stdout and stderr into one byte stream by construction,
  paints with cursor moves rather than newlines (which is why `session logs` renders
  through a terminal emulator at all — CONTEXT.md, **Screen**), and gives no exit status
  distinguishable from what the shell printed.
- **Not a `-c` flag on `core shell`.** That verb's contract is *interactive*, and it is
  built out of raw mode, `SIGWINCH`, a detach path and a `finally` that restores the
  terminal on every exit route (`packages/cli/src/core-shell.ts:10-36`). A
  non-interactive mode inside it would have to disable all four.

### The consequence: this needs a new frame, and the protocol version moves

`core shell` is built on **the SDK's existing PTY vocabulary and nothing else** —
`packages/cli/src/core-shell-channel.ts:11-16`: *"The SDK is untouched (D11). Everything
below is `@actana/sdk`'s existing PTY vocabulary — `spawn`, `write`, `resize`, `kill`,
`onData`, `onExit` — with no TTY anywhere near it. The shell this opens is the Core's
existing VM Shell Session spawn mode (`shellSession: true`)."*

That vocabulary **cannot express what `core exec` needs**: there is no frame that returns
separated stdout and stderr, because a PTY has one stream. So `core exec` is a new
request/response pair on the core link, which means:

- `CORE_LINK_PROTOCOL_VERSION` moves off `0.15.0`
  (`packages/sdk/src/core-link-frames.ts:1302`).
- Under ADR 0024 D11 / ADR 0005 that is a **version lock**: *"a mismatched Core renders as
  'needs update' with the command to run — never a degraded mode."* The narrow additive
  exception (announce on `ready`, absence yields today's behaviour *exactly*) does apply
  here — a Core that cannot exec should make `actana core exec` say so, and everything
  else is unchanged — so **announcing the capability on `ready` rather than moving the
  minor is the option to weigh**, and whichever is chosen belongs in the issue's
  resolution, not in a reviewer's inference.
- The frames live in `packages/sdk/src/core-link-frames.ts`, which the Core imports
  (ADR 0025 D1/D2) — no new dependency edge, no ADR amendment needed.

### Streaming, and a dropped link

The brief asks for this to be decided and documented. **Recommendation: buffered, not
streamed, in the first version, with an explicit bound.**

- A buffered result is what makes `stdout` and `stderr` separable and what makes the
  document shape trivial under the package's `--json` rule (`session-command.ts:35-40`:
  every verb writes exactly one JSON document to stdout, including on failure).
- The bound has to be real and stated: an exec whose output exceeds it fails with a named
  refusal rather than silently truncating. A truncated stdout that looks complete is the
  D39 class of failure this repository already names — *it ships broken and looks fine*.
- **On a dropped link mid-command:** the command keeps running on the Core (that is what
  a non-interactive spawn does), and the CLI has no result. It must exit non-zero with a
  distinct code meaning *"the link went away; the command's fate is unknown"* — **never**
  the command's own exit code, and never `0`. Note the asymmetry with `core shell`, which
  treats a dropped link as an ordinary ending (`core-shell.ts`, `Ending.disconnected`)
  because a human is watching; nobody is watching an exec.
- Streaming is a later addition and a compatible one: `--stream` interleaved on stderr,
  the buffered document still on stdout.

### Privilege — it grants nothing new, and here is the citation

The brief asks that this be stated rather than left for a reviewer to wonder about.
**`core exec` grants no privilege `core shell` does not already grant**, and the argument
is already in the tree at `packages/cli/src/core-shell.ts:3-8`:

> A real interactive shell on the Core, over the PTY transport the Panel already uses. F3
> is allowed to call path confinement *an accident guard, not a security boundary*
> precisely because this command exists: **anybody holding a valid blob for a Core already
> has that machine**, and pretending otherwise would buy nothing except a worse story for
> the operator who needs to look at a disk that is full.

Same credential (the registration blob), same transport (the core link, mTLS + bearer in
remote mode), same spawn class (`shellSession: true` — no project root, no harness). The
only thing that changes is that the bytes come back structured instead of painted. Two
things follow that should be said out loud in the implementation:

- It is **auditable in a way `docker exec` is not** — it goes through the Core's own
  authentication, so a remote Core can log it. That is a security *improvement* over the
  status quo, not a widening.
- ADR 0021's consequence — *"a Panel gesture can now start a long-running process on a
  Core"* — has an analogue here for the CLI, and the same bound applies: it is one
  command, from one authenticated client, and mTLS + bearer is what stands between it and
  anyone else.

---

## Acceptance criteria

**Part A — the Panel:**

- [ ] The Panel has **exactly one** control that opens a terminal, in each place one is
      offered, it is labelled **"New Terminal"**, and it opens a VM Shell Session.
- [ ] `grep -rn 'createTerminal\|VmShellButton' packages/panel/src` returns no hits
      outside tests: the component is gone and the store's creator is gone.
- [ ] The project-root terminal's server half is gone too — `createUserTerminal`, its
      controller, its repository and its two router paths — and the `user_terminals`
      table's fate (dropped or orphaned) is decided and recorded.
- [ ] `user-terminal-warm-pool.ts` is gone, and navigating to a project pre-spawns no PTY.
- [ ] `terminal.newTab` either has a `useHotkey` handler that opens the surviving control,
      or is removed from `keybindings/types.ts`, `defaults.ts`, `groups.ts` and both
      `HotkeyTooltip`s. It is not left advertised and inert.
- [ ] The `EmptyState` copy no longer describes a project-root shell.
- [ ] A test covers the surviving control — the panel has none for either creator today,
      which is how this arrived unnoticed.

**Part B — the CLI:**

- [ ] `actana core exec -- <cmd>` returns the command's exit code, and
      `actana core exec -- sh -c 'exit 3'; echo $?` prints `3`.
- [ ] stdout and stderr come back separately and contain **no terminal escape sequences**
      — asserted on a command that would emit them under a PTY (`ls --color=always`, a
      progress bar).
- [ ] `--json` produces exactly one document on stdout, on the failure path as well as the
      happy one, per the package's existing rule.
- [ ] `--core`, `ACTANA_CORE_BLOB` and the `current` pointer select the Core in that order,
      identically to every other verb.
- [ ] `--cwd` is honoured, and a refused cwd comes back as the Core's own refusal rather
      than a client-side guess (CONTEXT.md: *"One place validates a path: the machine that
      owns the disk"*).
- [ ] A dropped link mid-command exits non-zero with a code distinct from any the command
      could return, and says the command's fate is unknown.
- [ ] Output exceeding the stated bound fails loudly; it is never silently truncated.
- [ ] The protocol decision is recorded: either `CORE_LINK_PROTOCOL_VERSION` moved, or the
      capability is announced on `ready` and its absence yields a clean refusal.
- [ ] `no-local-escape.test.ts` still passes — `core exec` starts a process on the *Core*,
      over the link, and none locally.

**The outcome, restated as the check that matters:**

- [ ] A maintenance script operates on a Core **without `docker exec`**, and the same
      script works unchanged against a remote Core.

---

## Not in scope

- **Any migration or backward compatibility for terminal rows persisted under the removed
  kind.** A terminal is ephemeral and the removed path did not work; there is nothing to
  preserve.
- **`core shell` itself.** Its contract is interactive and it is untouched — no `-c`, no
  flags, no refactor to share an implementation with `exec`.
- **Streaming exec output.** Decided as a later, compatible addition (§B.2); the first
  version is buffered with a stated bound.
- **A VM-shell warm pool.** The removed one pre-spawned on navigation; do not reintroduce
  it under a new name.

## Verification note

Checked against `e6d2fd9` (`beta/0.3.3`). **Verified:** all three Panel line numbers
(`:326`, `:402`, `:408`) and the symbols on them; the absence of an `exec` verb; the Core
selection order; that `core shell` is PTY-only and cannot carry separated streams; that
the privilege question is already answered in `core-shell.ts`. **Corrected:** the header
`+ New` *is* the project-root path, so there is no third control to rename (§A.1); the
removal spans a table, a repository, a service, a controller, two router paths and the
warm pool, not two buttons (§A.3); `core exec` requires a new core-link frame and a
protocol decision the brief does not mention (§B.2). **Not verified:** that the
project-root terminal "does not work" — no repro was found in the tree and there is no
test on either creator, so it is recorded as reported rather than asserted, and the change
is justified without it (§A.2). **Found while checking:** `terminal.newTab` (⌘T) is
advertised in two tooltips and has no handler anywhere in the Panel.

qcentic-adm:
Closes #266. Both parts are in scope and both are here.

## Part A — one terminal control, and it opens a VM shell

The Panel offered three controls for two behaviours: `+ New` in the toolbar and
`New terminal` in the empty state, both opening a shell confined to the project
root, plus `New VM shell` beside the second. Each location now keeps exactly one
button, it is labelled **New Terminal**, and it opens a VM Shell Session on the
Core the route is on.

The project-root path is removed rather than deprecated — and as §A.3 warned, it
was a feature rather than two call sites:

| Gone | |
|---|---|
| `user-terminal-store.tsx` | `createTerminal`, both branches, and the project-terminal loader |
| `user-terminal-warm-pool.ts` | the whole module, and its test |
| `db/schema.ts` | the `userTerminals` table and its index |
| `repositories/user-terminals.repo.ts` | the whole module |
| `services/user-terminals.ts` | `createUserTerminal` and the rest, and its test |
| `controllers/user-terminals.controller.ts` | the whole module |
| `api-router.ts` | `PROJECT_USER_TERMINALS_PATH`, `USER_TERMINAL_PATH` |
| `lib/api.ts` | the four `user-terminals` calls |
| `queries/index.ts` | `userTerminalsQueryOptions`, `useUserTerminalsQuery`, the query key (all already unused) |
| `UserTerminalPanel.tsx` | `VmShellButton` |

**The warm pool is worth its own sentence.** It fired on project navigation, so
visiting a project pre-spawned a PTY on the Core in anticipation of a click on a
button that no longer exists. It is deleted and deliberately **not** reintroduced
for VM shells: CONTEXT.md requires an explicit open gesture and a pre-spawn is
the opposite of one. The JS-module prefetch on navigation stays — downloading a
chunk starts no process on any machine.

`UserTerminal` is re-rooted on `home_terminals` (the table it was inferred from
is gone), and the runtime-only `startCommand` hint goes with the launch path
that was its only writer.

## Part B — `actana core exec`

```
actana core exec [--cwd <dir>] [--json] -- <cmd> [args...]
```

Non-interactive, no PTY. stdout and stderr buffered and returned **separately**,
free of terminal escape sequences. The command's real exit code becomes this
process's; `128 + n` for a signal. `--cwd` names a directory on the **Core's**
machine and the Core validates it. Core selection is `resolveCore`, unchanged
and shared with every other verb.

`core shell` is untouched — no `-c`, no flags, no shared implementation.

It grants no privilege `core shell` does not already grant: same blob, same
link, same class of process (`core-shell.ts:3-8` makes the argument). It is more
auditable than what it replaces, because it goes through the Core's own
authentication rather than a `docker exec` on the host that the Core never sees.

---

### 1. `user_terminals` is **DROPPED**, not left orphaned

`ensureSchema` no longer creates it, and a new `dropLegacyUserTerminals` removes
the table and its index on every boot — following the eight `dropLegacy*` sweeps
already in `packages/shared/src/schema-bootstrap.ts`, and called **last** so the
worktree / sandbox / convenience sweeps above it still find the columns they
expect on an old DB. Idempotent, and a no-op on a fresh DB.

A terminal is ephemeral, so no rows are migrated — the issue rules that out and
is right. But "no migration" is not "leave the table", and a table nothing can
ever read again is a question every future reader of this schema would have to
answer for themselves.

**`home_terminals` stays.** It is where every terminal the Panel opens now
persists, whichever scope opened it — which is also why the store's `kill` and
`rename` no longer branch home-vs-project: there is one row kind, so one endpoint.

### 2. `terminal.newTab` is **WIRED**, not removed

It is now a real `useHotkey` in `__root.tsx` and opens exactly what the button
opens: a VM Shell Session on the route's Core.

One correction to §A.1 worth recording: ⌘T was not entirely inert. There was no
`useHotkey("terminal.newTab", …)` — that part is exact — but a hard-coded,
non-rebindable `keydown` listener sitting beside ⌘[ and ⌘] called
`createTerminal()`. So the real defect was subtler than "does nothing": the
tooltip and the keybindings editor advertised a *rebindable* action that the
editor could not actually rebind, because the handler matched on the literal key.
It is one action now, so the tooltip tells the truth and a rebind takes effect.

### 3. The protocol version **MOVES**: `0.15.0` → `0.16.0`

Part B needs a new frame. The SDK's PTY vocabulary cannot express it and never
could — a PTY is one byte stream by construction, so nothing existing returns the
two streams apart or an exit status distinguishable from what the shell printed.
So `exec` and `execResult` are new frames, and `CORE_LINK_PROTOCOL_VERSION` moves
where it is declared, `packages/sdk/src/core-link-frames.ts`.

The alternative — announcing a capability on `ready` under ADR 0024 D11's narrow
additive exception — was weighed rather than skipped, and the weighing is written
out above the constant. In short: that exception is for a surface whose absence
is a *supported state*, the way a Core without `multiConnection` or `files` is
fully usable. `actana core exec` against a Core that cannot exec is not a
withheld affordance; it is the command not working, and ADR 0005's answer to that
is one chore on that machine, said once and early, never a per-verb refusal met
later. No capability is announced for `exec`. Patch stays 0; the compatibility
gate compares major.minor.

Consequence, stated plainly: a Core still on 0.15.0 renders as **needs update**.
That is the version lock working, not a regression.

Two SDK tests asserted `=== "0.15.0"`. Both are updated rather than deleted, and
both keep their reason in the test name — they exist to fail an edit that bumps
the version *for multiConnection* or *for files*, and that reason survives.

## Other decisions worth naming

- **Buffered, not streamed**, with a stated bound of **8 MiB across both
  streams**. Past it the Core refuses by name — an `error` frame carrying
  `exec-output-too-large` — and returns *nothing*. Not a truncated result: a
  stdout that looks complete and is not is the D39 class of failure. `--stream`
  stays a later, compatible addition.
- **A dropped link exits `125`** and says the command's fate is unknown. Never
  `0`, never the command's own status. 125 follows the one convention that
  already means this (`docker run`, `git bisect run`: *the runner failed*).
  The convention is not a proof — a child can exit 125 like anything else in
  0–255, and no POSIX status is unreachable — so `--json` carries the
  distinction structurally: `{"outcome":"link-lost","exitCode":null}` is a
  different document from `{"outcome":"exited","exitCode":125}`. See the
  checklist for the honest reading of that criterion.
- **A time bound too**, which the issue did not ask for: 15 minutes on the Core.
  Nobody is watching an exec, so an unbounded child would hold a process on the
  Core until the daemon restarted. The CLI's request timeout is 16 minutes so
  the Core's answer always wins the race.
- **argv, never a shell string.** The Core spawns the executable directly with
  no shell, no inherited stdin. A caller that wants a shell names one:
  `-- sh -c '…'`.

## Files this ticket owns first, for #265

#265 will also edit these. What changed in each:

| File | Change |
|---|---|
| `packages/cli/src/core-resolution.ts` | **Nothing. Not touched.** `core exec` reaches it through `openCore`, which already calls `resolveCore`, so the selection order needed no widening. |
| `packages/cli/src/__tests__/no-local-escape.test.ts` | **Nothing. Not touched, and it passes.** `core exec` starts no local process; the child runs on the Core, over the link. The sweep that fails on a `child_process` import in `packages/cli/src` is unweakened, and `src/core-exec.ts` is inside its recursive scope. |
| `packages/sdk/src/core-link-frames.ts` | Added `exec` (request) and `execResult` (response); added `"exec"` to `REQUEST_FRAME_TYPES`; added `EXEC_OUTPUT_TOO_LARGE_ERROR_CODE` and widened `CoreLinkErrorCode` from an alias to a two-member union; moved `CORE_LINK_PROTOCOL_VERSION` to `0.16.0` with the reasoning above it. Nothing existing changed shape or meaning. |

And the one #177 also touches later:

| File | Change |
|---|---|
| `packages/shared/src/pty-spawn-policy.ts` | **Comment only — seven lines, no code.** A note on `ShellSessionSpawnRequest`'s doc block saying that `core exec` starts a plain child with pipes, is not a spawn mode in this union, and validates its cwd in `packages/core/src/core-exec.ts`. Nothing here was relaxed for it. |

`packages/core/src/harness-hooks.ts` and `packages/shared/src/harness-cli-config.ts`
are untouched. `.github/workflows/ci.yml`, `docs/rulesets/`, `AGENTS.md`,
`CLAUDE.md` and `docs/ci-cd.md` are untouched. The README note about the new verb
is inside `packages/cli`, as allowed.

## Verification

Each package's tests run **explicitly**, not through `pnpm -r`, because
`packages/cli` sorts before `packages/core` and a bail on the first failure would
hide everything after it:

| Package | Result |
|---|---|
| `packages/cli` | 453 passed, 3 skipped (33 files) |
| `packages/core` | 1200 passed (65 files) |
| `packages/sdk` | 250 passed, 5 skipped (23 files) |
| `packages/shared` | 184 passed (14 files) |
| `packages/panel` | 1439 passed (142 files) |

`pnpm typecheck` clean across all five. `pnpm lint`: 0 errors, 46 warnings —
down from 50 on the base commit, all pre-existing.

`packages/panel` intermittently timed out 2–3 heavy live tests at full worker
concurrency on the machine this ran on; the same files pass in isolation and the
whole suite is green at `--maxWorkers=2`. CPU contention, not a regression.

Every commit was checked against this repo's own `commitlint.config.mjs` beside a
throwaway install — the same thing the Conventions job does — and every body line
measured with `awk` at ≤132 characters.

## New tests

- `packages/panel/.../user-terminal-panel-control.test.tsx` — the panel had none
  for either creator, which is how three buttons for two behaviours went
  unnoticed. Asserts one control per location, its label, that it calls
  `createVmShellTerminal` with the route's Core, that `New VM shell` is gone,
  that the empty-state copy no longer describes a project shell, and that it is
  disabled with no Core in scope.
- `packages/panel/src/db/__tests__/drop-legacy-user-terminals.test.ts` — the drop
  is idempotent, a no-op on a fresh DB, leaves `home_terminals` alone, and takes
  the table off a DB carried over from the previous build.
- `packages/cli/src/__tests__/core-exec.test.ts` — the surface: exit codes,
  signals, stream separation, the `--json` document on every path, Core
  selection order, and the dropped link.
- `packages/cli/src/__tests__/core-exec-live.test.ts` — **the same verb over a
  real `wss://` socket against the real `PtyCoreLinkServer` with the real
  `runCoreExec` behind it, forking real children.** `sh -c 'exit 3'` really exits
  3; the command really reports `PIPE` rather than `TTY`; `ls --color=auto`
  really comes back unpainted; and a `printf %b` of explicit octal escapes —
  ESC, BEL, BS and DEL around printable text — really comes back byte for byte
  on stdout and stderr alike, which is what proves the clean output is the
  *spawn* and not a scrubber in the CLI that could be wrong about somebody's
  bytes. See **The `--color=always` correction** below for why that last one is
  no longer asked of `ls`.
- `packages/core/src/__tests__/core-exec.test.ts` — real processes: status,
  signal, streams, no TTY, no inherited stdin, no shell, all three bounds.
- `packages/sdk/src/__tests__/core-link-exec-frames.test.ts` — the frames, the
  coded refusal, and the version move with its reason in the test names.

## Not done, deliberately

- **No ADR.** ADR 0025 D1/D2 already put these frames in the SDK for the Core to
  import, so no new dependency edge and no amendment is needed — the issue says
  so in §B.2 and I agree.
- ~~**CONTEXT.md and CHANGELOG.md untouched**, to keep the conflict surface off
  the release train.~~ **Reversed on review.** The changelog's unreleased section
  named `0.15.0`, which after this PR is wrong for the release it describes, and
  it is the most operator-visible consequence of the change. `CHANGELOG.md` now
  says `0.16.0`, names `actana core exec` among the reasons, and spells out that
  every Core in the fleet is grounded until it is updated; two more paragraphs
  cover the new verb and the single New Terminal control.
  `docs/reference/cli-parity-check.md` went with it — the live claim about the
  SDK constant now reads `0.16.0` at line 1398, and the deleted mirror's own
  hand-typed `0.15.0` stays as the record of what it held, now saying so.
  CONTEXT.md is still untouched.

Draft, base `chore/release-discipline-and-cursor-cli`. Nothing merged; `main` and
the beta branches untouched.

---

## The `--color=always` correction

The first push of this branch was RED on one test, and the test was wrong rather
than the code. `core-exec-live.test.ts` proved *byte passthrough* — that the
clean output of the `--color=auto` case is the spawn and not a scrubber in this
CLI — by running `ls --color=always` and asserting an ESC byte came back. That
is a claim about the machine the suite runs on. GNU `ls` disables colour
outright when `TERM` is unset, whatever `--color=always` says, and `TERM` is
unset on a GitHub runner; it returned a bare `a-dir\n` and the assertion failed
with nothing wrong in the code under test.

The intent is kept and the assertion is stronger. The spawned command now emits
a byte sequence the test chose itself — a `printf %b` of explicit octal escapes,
`\0033` `\0007` `\0010` `\0177`, around printable text — on stdout **and** stderr,
and each is compared **byte for byte** against the string the test built rather
than searched for a single escape. Nothing depends on a tool's opinion about
terminals, so it is deterministic on a runner and on a laptop alike. The
`exitCode === 0` guard that let BSD `ls` opt out of the question went with the
command that needed it, and the claim is also asserted on the plain non-`--json`
stream an operator actually reads.

Checked by mutation, not by assertion: adding a C0 strip to `runCoreExec`'s
stdout in `packages/core` makes this test fail, and reverting it makes it pass.

## CI evidence for #257 — `packages/core` and `packages/panel` never ran

Worth recording here because it is exactly the defect #257 exists to fix, and it
showed up on this PR before anyone went looking for it.

Run [32244377440](https://github.com/actana/control/actions/runs/32244377440)
ended with:

```
[ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL] @actana/cli@0.3.3 test: `vitest run`
```

`pnpm -r test` bailed on the first failing project. The log shows `Scope: 5 of 6
workspace projects` and then only three of them reporting: `packages/sdk`,
`packages/shared` and `packages/cli`. **`packages/core` and `packages/panel`
never ran at all** — 2639 tests, silently unreported, on a run whose only actual
failure was one assertion in `packages/cli`. A red X that hides two thirds of
the suite is worse than a red X, because the next person reads "one test failed"
and believes it.

So the verification table below is not `pnpm -r test` output. Each package was
run explicitly and locally, exactly as #257 will make CI do:

| Package | Result, after the fix |
|---|---|
| `packages/cli` | 453 passed, 3 skipped (33 files) |
| `packages/core` | 1200 passed (65 files) |
| `packages/sdk` | 250 passed, 5 skipped (23 files) |
| `packages/shared` | 184 passed (14 files) |
| `packages/panel` | 1439 passed (142 files) |
| root `vitest run` | 474 passed (16 files) |

`pnpm typecheck` clean across all five. `pnpm lint`: 0 errors, 46 warnings, the
same 46 as before.

One local-only note, so it is not mistaken for a regression: the root suite's
`scripts/__tests__/core-launcher.test.mjs` fails inside this dev container
because the container sets `AC_APP_PATH=/opt/actana/app` in its own environment
and the launcher honours it. Unset that variable and the root suite is 474/474;
it is green on CI, which has no such variable, and it fails identically on the
commit before this branch's fix.


qcentic-adm:
# Acceptance criteria, issue 266 — every one, checked or explicitly not

## Part A — the Panel

- [x] **The Panel has exactly one control that opens a terminal, in each place one
  is offered, it is labelled "New Terminal", and it opens a VM Shell Session.**
  Toolbar (`UserTerminalPanel.tsx`) and empty state, one each, both calling
  `createVmShellTerminal(coreId)`. Asserted in
  `user-terminal-panel-control.test.tsx`, which enumerates every button the panel
  renders and expects exactly `["New Terminal", "New Terminal"]`.

- [x] **`grep -rn 'createTerminal\|VmShellButton' packages/panel/src` returns no
  hits outside tests — the component is gone and the store's creator is gone.**
  Met in substance; the literal grep does not return zero, and here is exactly
  why, because the difference matters to whoever runs it next.

  `VmShellButton`: zero hits, anywhere.

  `createTerminal`: the store's creator is gone, and so is every call site. What
  the unanchored grep still matches is five **unrelated xterm helpers** whose
  names begin with that string — `createTerminalOptions`, `createTerminalTheme`,
  `createTerminalLinkHandler`, `createTerminalGpuLease`,
  `createTerminalSurfaceCache`. All pre-existing, none related to the removed
  path, and renaming them to satisfy a grep would be a worse change than the one
  this issue asked for. Anchored on a word boundary,
  `grep -rn 'createTerminal\b' packages/panel/src` returns **one** hit: a comment
  in `user-terminal-store.tsx`'s header naming the symbol that was removed and
  what went with it. Kept on purpose — documenting a removal is what makes it
  findable a year from now.

- [x] **The project-root terminal's server half is gone too — `createUserTerminal`,
  its controller, its repository and its two router paths — and the
  `user_terminals` table's fate is decided and recorded.**
  All four deleted outright, plus the client's four `api.*` calls and three
  already-unused query helpers. **Decision: the table is DROPPED**, via
  `dropLegacyUserTerminals` in `schema-bootstrap.ts` following that file's eight
  existing `dropLegacy*` sweeps, called last so the older sweeps still find the
  columns they expect. No rows migrated. `home_terminals` stays. Recorded in the
  PR body, in `db/schema.ts`'s comment, and in
  `drop-legacy-user-terminals.test.ts`.

- [x] **`user-terminal-warm-pool.ts` is gone, and navigating to a project
  pre-spawns no PTY.** Module and test deleted. The navigation effect that
  called `prepareUserTerminalWarmSlot` now prefetches JS modules and nothing
  else, and the store carries a comment saying a VM-shell warm pool is not to be
  added under a new name.

- [x] **`terminal.newTab` either has a `useHotkey` handler that opens the
  surviving control, or is removed. It is not left advertised and inert.**
  **Decision: WIRED.** `useHotkey("terminal.newTab", …, { capture: true })` in
  `__root.tsx` opens a VM Shell Session on the route's Core. One correction to
  §A.1 while recording it: ⌘T was not fully inert. There was no
  `useHotkey("terminal.newTab", …)` — exact — but a hard-coded, non-rebindable
  `keydown` listener beside ⌘[ / ⌘] called `createTerminal()`. So the defect was
  that the tooltip and the keybindings editor advertised a *rebindable* action
  the editor could not rebind. That is fixed, and the literal-key branch is gone.

- [x] **The `EmptyState` copy no longer describes a project-root shell.** Both
  strings replaced with one: *"Open a shell on this Core's machine."* Asserted
  negatively too — the test fails if "in this project" or "home directory" comes
  back.

- [x] **A test covers the surviving control.**
  `packages/panel/src/components/views/__tests__/user-terminal-panel-control.test.tsx`,
  6 tests: the count and labels, the absent `New VM shell`, both click paths
  reaching `createVmShellTerminal` with the route's Core, the copy, and the
  disabled state with no Core in scope.

## Part B — the CLI

- [x] **`actana core exec -- <cmd>` returns the command's exit code, and
  `-- sh -c 'exit 3'; echo $?` prints `3`.** Asserted twice: against a fake in
  `core-exec.test.ts`, and against a real child over a real socket in
  `core-exec-live.test.ts`.

- [x] **stdout and stderr come back separately and contain no terminal escape
  sequences — asserted on a command that would emit them under a PTY.**
  `core-exec-live.test.ts` asks the command itself (`[ -t 1 ]` → `PIPE`), then
  runs `ls --color=auto` in a directory containing a subdirectory and asserts no
  ESC byte in the result or on the CLI's own streams. A second test runs the same
  `ls` with `--color=always` and asserts the escapes **do** come through — so the
  clean result above is a fact about the spawn having no terminal, not about a
  scrubber in this CLI that could be wrong about somebody's real bytes.

- [x] **`--json` produces exactly one document on stdout, on the failure path as
  well as the happy one.** Four documents, four tests: `exited`, `refused`,
  `unreachable`, `link-lost`. Each parsed from the whole of stdout, which fails
  if anything else was written there.

- [x] **`--core`, `ACTANA_CORE_BLOB` and the `current` pointer select the Core in
  that order, identically to every other verb.** Literally the same code —
  `openCore` → `resolveCore`, untouched. One test walks all three precedences.

- [x] **`--cwd` is honoured, and a refused cwd comes back as the Core's own
  refusal rather than a client-side guess.** The CLI sends the string verbatim
  (asserted with `../../etc`, which it does not resolve, normalise or reject) and
  the Core validates in `core-exec.ts`. The live test reads back the Core's own
  sentence — *"No such directory on this Core: …"* — which no client-side check
  could have written, because a client cannot see that filesystem.

- [⚠️] **A dropped link mid-command exits non-zero with a code distinct from any
  the command could return, and says the command's fate is unknown.**
  Implemented and tested, and I am flagging the "distinct" half rather than
  ticking it silently, because as written it cannot be fully satisfied by any
  implementation: every status in 0–255 is reachable by a child process, so no
  reserved value is provably unreachable. Nearest honest thing, and what is
  built: exit **125**, following the one existing convention that means exactly
  this (`docker run`, `git bisect run`: *the runner failed, not your command*),
  used by nothing else in this CLI; the stderr sentence says the command may
  still be running and its outcome is unknown; and the distinction is carried
  **structurally** under `--json`, where `{"outcome":"link-lost","exitCode":null}`
  is a different document from `{"outcome":"exited","exitCode":125}`. A script
  needing certainty reads that. Never `0`, never the command's own status.
  Asserted against a fake and, in the live suite, by destroying the socket under
  a running `sleep 20`. The reasoning is in `exit-codes.ts` next to the constant.

- [x] **Output exceeding the stated bound fails loudly; it is never silently
  truncated.** Bound stated: **8 MiB across both streams**, `EXEC_MAX_OUTPUT_BYTES`.
  Past it the Core answers an `error` frame carrying `exec-output-too-large` and
  returns nothing. The port's result type makes "truncated" *unrepresentable* —
  the refusal arm has no `stdout` field at all. Checked on both streams, and
  end-to-end in the live suite with a deliberately tiny bound.

- [x] **The protocol decision is recorded.**
  **`CORE_LINK_PROTOCOL_VERSION` moved, `0.15.0` → `0.16.0`**, where it is
  declared. The `ready`-capability alternative was weighed against ADR 0024 D11's
  narrow exception and rejected in writing above the constant: that exception is
  for a surface whose absence is a supported state, and `core exec` against a
  Core that cannot exec is the command not working, not a withheld affordance. No
  capability announced. Two SDK tests that asserted `0.15.0` are updated, not
  deleted, and keep their reasons in their names.

- [x] **`no-local-escape.test.ts` still passes — `core exec` starts a process on
  the Core, over the link, and none locally.** File untouched, 8/8 green.
  `packages/cli/src/core-exec.ts` sits inside its recursive sweep and imports
  nothing that can start a process. The child is forked by
  `packages/core/src/core-exec.ts`, on the Core.

## The outcome, restated

- [x] **A maintenance script operates on a Core without `docker exec`, and the
  same script works unchanged against a remote Core.** Nothing in the path knows
  or cares whether the Core is local: it dials the endpoint in the registration
  blob over mTLS + bearer, sends one frame, reads one answer. The live suite is
  that script, against a Core reached over a real `wss://` socket.

## Nothing was skipped for a boundary

No acceptance criterion needed a file this train fences off. `.github/workflows/ci.yml`,
`docs/rulesets/`, `AGENTS.md`, `CLAUDE.md`, `docs/ci-cd.md`,
`packages/core/src/harness-hooks.ts` and `packages/shared/src/harness-cli-config.ts`
are all untouched — verified with `git diff --name-only` against the base.
`packages/shared/src/pty-spawn-policy.ts` took a seven-line comment and no code.
`packages/cli/src/core-resolution.ts` and
`packages/cli/src/__tests__/no-local-escape.test.ts` were not touched at all, so
#265 inherits them exactly as it expects.

---

**PR:** #268 — https://github.com/actana/control/pull/268
(draft, base `chore/release-discipline-and-cursor-cli`)

mrfqcentic (changes requested):
Reviewed against issue #266's acceptance criteria, reading the enforcing line rather than the description, at `b2abce2` on base `chore/release-discipline-and-cursor-cli`.

This is very good work. Part A is a real removal, not a deprecation; the warm pool is genuinely gone and deliberately not reintroduced; `no-local-escape.test.ts` is untouched and unweakened; the protocol move is in the one place it is declared. The three decisions the issue asked for are all stated in the PR body rather than left to be inferred. Two things need changing before this merges, and both are small.

---

## Blocking

### 1. `core exec` raises the **connect** deadline to 16 minutes, not just the request deadline

`packages/cli/src/core-exec.ts:96-98`

```ts
const opened = await openCore(deps, args, paths, "actana core exec", {
  timeoutMs: EXEC_TIMEOUT_MS,      // 16 * 60_000
});
```

`openCore` forwards `timeoutMs` to `connectCore`, which uses it for **both** halves (`packages/cli/src/core-connection.ts:114-124`):

```ts
const timeoutMs = opts.timeoutMs ?? DEFAULT_CORE_TIMEOUT_MS;   // 30_000
... CoreClient.fromRegistrationBlob(blob, {
      connectTimeoutMs: timeoutMs,
      requestTimeoutMs: timeoutMs,
    });
```

and `CoreClient.connect()`'s only deadline is that value (`packages/sdk/src/core-client.ts:426-429`). So a Core that accepts the socket but never finishes the core-link handshake leaves `actana core exec` sitting for **16 minutes** with nothing on stdout, nothing on stderr and no `--json` document — where every other verb gives up after 30 s. `core exec` is the one verb explicitly built for unattended scripts, which is exactly where a silent 16-minute hang is worst.

The 16-minute bound is right for the *request* and is already applied there independently at `core-exec.ts:121-130`, so `openCore` does not need it at all:

```ts
const opened = await openCore(deps, args, paths, "actana core exec");
```

`events tail` is the only other verb that passes options to `openCore` (`events-command.ts:143`) and it deliberately does not raise the timeout. This should match.

### 2. The 0.16.0 protocol move is not in `CHANGELOG.md`, and the changelog currently tells operators the opposite

`packages/sdk/src/core-link-frames.ts:1398` moves the version, and the PR body is right that the consequence — every Core still on 0.15.0 renders as **needs update** — is the version lock working rather than a regression. But `CHANGELOG.md:57`, in the live `0.1.0 — unreleased` section, says:

> The core-link protocol moves to 0.15.0 ... The Panel and its Cores are version-locked, so every Core needs updating alongside the Panel — an older one renders as "needs update" rather than degrading.

That paragraph is this repo's own established way of recording precisely this fact, and after this PR it is wrong for the release it describes: the train ships 0.16.0 and grounds every existing Core, while the changelog names 0.15.0. This is the single most operator-visible consequence of the PR and it is the one thing not written down where operators read.

You asked for the call in "Not done, deliberately" — this is the answer: please extend that paragraph (or add one beside it) for 0.16.0 and `actana core exec`. It is a few lines in a section that already exists, and the conflict risk of one paragraph is much smaller than shipping a changelog that misstates the fleet-wide upgrade requirement.

While there: `docs/reference/cli-parity-check.md:369` and `:380` still quote `0.15.0`. You flagged them as a stale point-in-time audit and I agree that is defensible — but if the changelog is being touched anyway, these two lines are nearly free.

---

## Non-blocking, but worth fixing

**3. `writeStream` invents a trailing newline** — `packages/cli/src/core-exec.ts:196-200`

```ts
const body = text.endsWith("\n") ? text.slice(0, -1) : text;
for (const line of body.split("\n")) sink(line);
```

`deps.out` is a line sink, so `actana core exec -- printf hello` writes `hello\n` where the command produced `hello`. The test at `core-exec.test.ts:110` covers the opposite case (no *blank* line added) but not this one. For a verb whose argument is that the bytes come back unpainted, adding a byte the command did not emit is the wrong direction — even if it is the direction every shell pipeline tolerates. Either write raw through a byte sink, or say in the header that the non-`--json` path is line-oriented and `--json` is the byte-faithful channel.

**4. The `unreachable` `--json` document has no `error` field** — `packages/cli/src/core-exec.ts:100`

```ts
if (args.json) deps.out(formatJson({ outcome: "unreachable" }));
```

`refused` and `link-lost` both carry `error`; `unreachable` carries nothing and the sentence went to stderr only (which is `openCore`'s documented policy, so I understand why). A consumer reading `.error` off the one document gets `undefined` on one of the four outcomes. Worth including the sentence so all four documents have the same shape.

**5. `drop-legacy-worktree-schema.test.ts:109` is now vacuous**

```ts
expect(columns(fresh, "user_terminals")).not.toContain("worktree_id");
```

After `ensureSchema`, the table no longer exists, so `columns()` returns `[]` and the assertion passes for the wrong reason. It does not fail, so this is not blocking — but a test that can no longer detect what it was written to detect should be deleted or re-pointed. (The pre-cutover assertion at `:92` is fine: it calls `dropLegacyWorktreeSchema` directly, not `ensureSchema`.)

**6. `terminal.newTab` has no test** — `packages/panel/src/routes/__root.tsx:422-428`

The criterion says "a `useHotkey` handler that opens the surviving control", and it does. But the panel's new test covers only the buttons, and the hotkey is the half that was silently broken before. Also: with no `routeCoreId` the handler `preventDefault`s and then does nothing at all, with no feedback — consistent with the disabled button, but it is the same "advertised, does nothing" shape in miniature.

**7. Shared-file coordination** — `packages/core/src/core-entry.ts:363-369`

The handoff table names `core-resolution.ts`, `no-local-escape.test.ts`, `core-link-frames.ts` and `pty-spawn-policy.ts`, and I verified all four claims. `core-entry.ts` is also touched here (+9) and #265 cites it (`core-entry.ts:289-290`, the harness-availability tick at Core boot). Different region, trivial merge — but it belongs in that table so #265 is not surprised. Same for `packages/shared/src/schema-bootstrap.ts`, a shared package file changed by this ticket and mentioned only in prose.

---

### Part A

| # | Criterion | Verdict |
|---|---|---|
| A1 | Exactly one control per location, labelled **New Terminal**, opens a VM Shell Session | **MET** — `UserTerminalPanel.tsx:333-345` (toolbar) and `:408-421` (empty state), one `icon="plus"` `Btn` each, both calling `createVmShellTerminal(coreId)`. The only other `icon="terminal"` in the file is `EmptyState`'s own icon prop at `:399`, not a control. Asserted at `user-terminal-panel-control.test.tsx:82` on `["New Terminal", "New Terminal"]`. |
| A2 | `grep -rn 'createTerminal\|VmShellButton' packages/panel/src` returns no hits outside tests | **MET in substance.** `VmShellButton`: zero hits repo-wide — the component at the old `:478` is deleted. `createTerminal\b`: exactly one hit, the removal comment at `user-terminal-store.tsx:21`. The unanchored grep additionally matches five pre-existing xterm helpers (`createTerminalOptions`, `createTerminalTheme`, `createTerminalLinkHandler`, `createTerminalGpuLease`, `createTerminalSurfaceCache`). Your reading is correct and renaming those to satisfy a grep would be the worse change. |
| A3 | Server half gone; `user_terminals` fate decided **and recorded** | **MET.** Deleted outright: `user-terminals.repo.ts`, `services/user-terminals.ts` (with `createUserTerminal`), `user-terminals.controller.ts`, and both router paths (`PROJECT_USER_TERMINALS_PATH`, `USER_TERMINAL_PATH` at the old `api-router.ts:37,53`) with their dispatch arms. Table and index gone from `db/schema.ts:143-157` and from `ensureSchema`. Decision **DROPPED**, stated in the PR body §1, in `schema.ts:151-157`, in `schema-bootstrap.ts:539-549` and in a dedicated test. I checked the ordering claim: `dropLegacyUserTerminals` is called last (`schema-bootstrap.ts:549`), after the sweeps at `:600`, `:648-650` and `:698-699` that still `ALTER TABLE user_terminals` on an old DB. I also checked the boot path — `db/client.ts:69-76` runs migrations before `ensureSchema` only on a non-fresh DB, where the table still exists, and marks-all-applied on a fresh one, so `0003`/`0010`'s `ALTER TABLE user_terminals` never hits a missing table. Nothing holds an FK onto it, so the `DROP` is safe under `foreign_keys = ON`. |
| A4 | `user-terminal-warm-pool.ts` gone; project navigation pre-spawns no PTY | **MET, and this is the one that mattered.** Module and its test deleted. `user-terminal-store.tsx:231-244` now runs `prefetchTerminalModules()` and nothing else; `prepareUserTerminalWarmSlot` / `takeUserTerminalWarmSlot` / `replenish...` / `discard...` have no callers anywhere. The comment naming the prohibition sits on the effect that used to do it. |
| A5 | `terminal.newTab` wired or removed, never advertised and inert | **MET.** **WIRED**: `__root.tsx:422-428`, `useHotkey("terminal.newTab", ..., { capture: true })` opening `createVmShellTerminal(routeCoreId)`. `capture` is a real option (`use-hotkey.ts:35,53,80`). The hard-coded literal-key branch beside Cmd+[ / Cmd+] is gone (`__root.tsx:470-476`) and `createTerminal` is out of that effect's dep array. Your §A.1 correction — that Cmd+T was mis-wired rather than fully inert — checks out against the base. See item 6 above. |
| A6 | `EmptyState` copy no longer describes a project-root shell | **MET.** Both strings at the old `:387-393` replaced with `"Open a shell on this Core's machine."`, and asserted negatively at `user-terminal-panel-control.test.tsx:102-108`. |
| A7 | A test covers the surviving control | **MET.** `user-terminal-panel-control.test.tsx`, 6 tests: count and labels, `New VM shell` absent, both click paths reaching `createVmShellTerminal("core_a")`, the copy, the disabled state. |

**VM Shell privilege rule — intact.** No warm pool of any kind was introduced for VM shells: the only `prepare*WarmSlot` left in the Panel is `session-warm-pool.ts`, the pre-existing agent-session pool, untouched here. Both entry points are an explicit click or an explicit keypress. `pty-spawn-policy.ts` gained a seven-line comment and no code.

### Part B

| # | Criterion | Verdict |
|---|---|---|
| B1 | `exec -- sh -c 'exit 3'` exits 3 | **MET** — `core-exec.ts:159,174` via `statusFor`; asserted against a fake (`core-exec.test.ts:68`) and against a real forked child over a real `wss://` socket (`core-exec-live.test.ts:72`). |
| B2 | Streams separate, no terminal escape sequences | **MET.** The Core spawns with `stdio: ["ignore","pipe","pipe"]`, `shell: false` (`core/src/core-exec.ts:129-137`); the frame carries them apart (`core-link-frames.ts:1055-1062`). The live test asks the command itself (`[ -t 1 ]` gives `PIPE`), then asserts no ESC byte under `ls --color=auto`, then asserts escapes **do** survive under `--color=always`. Using `auto` for the negative is more correct than the issue's literal `--color=always`, which forces colour regardless of TTY; the pair together proves absence-of-terminal rather than a client-side scrubber. Good call. |
| B3 | One `--json` document on stdout, failure paths too | **MET** in shape — four outcomes (`exited`, `refused`, `unreachable`, `link-lost`), each written once, each parsed from the whole of stdout in the tests. See item 4 for the field-shape nit. |
| B4 | `--core` then `ACTANA_CORE_BLOB` then `current`, identical to every other verb | **MET** — literally the same `openCore` to `resolveCore` path; `core-resolution.ts` is untouched by this PR (verified with `git diff --name-only`). |
| B5 | `--cwd` honoured; a refused cwd is the Core's own refusal | **MET** — the CLI sends the string verbatim (`core-exec.ts:127`, no normalisation, asserted with `../../etc` at `core-exec.test.ts:188`); the Core validates at `core/src/core-exec.ts:96-106` and the live test reads back *"No such directory on this Core: ..."*. |
| B6 | Dropped link exits non-zero with a code distinct from any the command could return, fate unknown | **PARTIALLY MET — and your warning flag is the right call, accepted as argued.** "Distinct from any status a child could return" is unsatisfiable: every value in 0-255 is reachable. What is built is the nearest honest thing — 125 (`exit-codes.ts:67`), used by nothing else in this CLI, with the convention's own precedent cited; a stderr sentence saying the command may still be running; and the distinction carried structurally under `--json`, where `{"outcome":"link-lost","exitCode":null}` cannot be confused with `{"outcome":"exited","exitCode":125}`. Never `0`, never the command's own status, and the disconnect is watched explicitly (`core-exec.ts:114-117`) rather than inferred from a rejection. No change requested. |
| B7 | Output past the bound fails loudly, never silently truncated | **MET, and structurally so.** `EXEC_MAX_OUTPUT_BYTES = 8 MiB` stated (`core/src/core-exec.ts:50`), checked as chunks arrive (`:184-193`), and the refusal arm of `CoreExecPortResult` has **no `stdout` field at all** (`pty-core-link-server.ts:228-236`) — truncation is unrepresentable rather than merely avoided. Travels as an `error` frame carrying `EXEC_OUTPUT_TOO_LARGE_ERROR_CODE`. |
| B8 | The protocol decision is recorded | **MET in code, INCOMPLETE in the release record.** `CORE_LINK_PROTOCOL_VERSION` is `0.16.0` at `core-link-frames.ts:1398` — the one place it is declared. Core and CLI agree because neither hard-codes it: the Core reads it at `pty-core-link-server.ts:633`, the CLI gates on `coreLinkProtocolCompatible` via `connectCore` (`core-connection.ts:126-136`), and the Panel at `core-link-manager.ts:345-347`. No stale literal survives outside tests and prose. The ADR 0024 D11 weighing is written above the constant, no capability is announced for `exec` (asserted at `core-link-exec-frames.test.ts:121`), patch stays 0, and both SDK tests that pinned `0.15.0` were updated with their reasons kept in the test names rather than deleted. See blocking item 2 for the changelog. |
| B9 | `no-local-escape.test.ts` still passes and was not weakened | **MET.** `git diff --name-only` shows the file **untouched** — no exemption, no narrowing, no new allow-list entry. I read it in full: the sweep is recursive over `packages/cli/src` with only `__tests__`/`node_modules` excluded, so `src/core-exec.ts` is inside its scope, and `PROCESS_STARTERS` is unchanged. `packages/cli/src/core-exec.ts` imports only `core-connection`, `cli-output`, `exit-codes` and types — nothing that can start a process. The child is forked by `packages/core/src/core-exec.ts:129`, on the Core. (The live test's `import { runCoreExec } from "@actana/core/core-exec"` sits in `__tests__`, outside both this sweep and the `@actana/core` one.) |
| — | A maintenance script reaches a Core without `docker exec`, unchanged against a remote Core | **MET** — one frame over the authenticated link; nothing in the path knows whether the Core is local. `core-exec-live.test.ts` is that script against a real socket. |

---

## The four train checks

1. **Every commit body line 132 characters or shorter** — **PASS.** `git log --format=%B origin/chore/release-discipline-and-cursor-cli..HEAD | awk 'length > 132'` is empty; the longest body line is 72 and the longest subject 67.
2. **Base branch** — **PASS.** `chore/release-discipline-and-cursor-cli`. Not `main`, not `beta/0.3.3`.
3. **No file outside the ticket's declared surface** — **PASS.** All 38 changed files are Part A, Part B, or a direct consequence (`UserTerminalPane.tsx` and `home-terminals.repo.ts` for the `startCommand` removal; `api-auth.test.ts`, `tasks.test.ts`, `home-terminals.test.ts` for the removed table; `in-process-core.ts` for the new port; two SDK tests for the version). Nothing in `.github/`, `docs/`, `docs/rulesets/`, `AGENTS.md`, `CLAUDE.md`, `CONTEXT.md` or `CHANGELOG.md` — verified with `git diff --name-only`, which is also why check 2 above is a request rather than an observation. `harness-hooks.ts` and `harness-cli-config.ts` are untouched, as claimed. See item 7 for the two shared files missing from the handoff table.
4. **CI job PASS case** — **N/A.** No workflow is added or restructured; `.github/` is untouched, so there is no job-level `if:` that could leave a required check Pending forever (ADR 0023 D33).

One thing you may not know: `packages/panel/src/components/views/UserTerminalPane.tsx` shows as `Bin 25292 -> 25419 bytes` in the diffstat because the file contains a literal NUL byte at `:157` (in the `buildKey` template string). That is **pre-existing** — it is in the base commit too — and not this PR's doing. But it means this file's changes are invisible in the GitHub diff view unless a reviewer forces `--text`, which is worth knowing for anyone after me and probably worth its own ticket.

---

I could not run the suites: this checkout has no `node_modules` and installing them would have been a write to the repository, which is outside a review's remit. Everything above is verified by reading the enforcing line. Your reported results (453 / 1200 / 250 / 184 / 1439, typecheck clean, lint 46 warnings down from 50) are taken as reported.

Fix the connect timeout and the changelog and this is ready.

qcentic-adm:
Thanks — all four are addressed in `d308db4`. Nothing deferred silently; the one place I did not do exactly what was asked is called out below.

## 1 — the connect deadline. Fixed.

`core exec` no longer passes `timeoutMs` to `openCore`:

```ts
const opened = await openCore(deps, args, paths, "actana core exec");
```

Your reading was exactly right: one value, spent on both `connectTimeoutMs` and `requestTimeoutMs`, and `CoreClient.connect()` has no other deadline. The 16-minute bound stays where it belongs, on `client.request(frame, EXEC_TIMEOUT_MS)`, and the constant's own doc comment now says it is the *request* deadline only and points at the call site for why.

There is a comment at the call site making the trap explicit for the next person, and a test that fails if the option comes back:

```ts
it("dials with the CLI's default deadline, not the command's 16-minute bound", …)
expect(core.connectOptions).toEqual([{}]);
```

`events tail` is named in that comment as the precedent.

## 2 — the changelog. Fixed, and the parity check with it.

`CHANGELOG.md`'s unreleased paragraph now says `0.16.0`, lists `actana core exec` among the reasons the version moved, and spells out the fleet consequence rather than leaving it at "an older one renders as needs update":

> That applies to every Core in the fleet, including one that is otherwise working: a Core left on the previous protocol is grounded until it is updated, and the Panel says so rather than half-working.

Two paragraphs went in beside it, in the section's own voice — one for `actana core exec` (what it is, why not `docker exec`, that `--json` is one document on every outcome), one for the single New Terminal control and the pre-spawn that is gone. The "Not done, deliberately" bullet in the PR body is struck through and now records the reversal instead.

`docs/reference/cli-parity-check.md` — one of the two lines I changed and one I did not, on purpose:

- **:380 corrected.** It was a live claim about the SDK and it was wrong twice over: `CORE_LINK_PROTOCOL_VERSION = "0.16.0"` at line 1398.
- **:369 kept at `0.15.0`, with a sentence added.** That line quotes the *deleted mirror's* own hand-typed constant, and the mirror really did hold `0.15.0` — rewriting it to `0.16.0` would falsify the record of the thing ADR 0025 D3 uses as its worked example. It now reads: that is the number the mirror was pinned at when it was deleted, kept as the record of what it held; the SDK has since moved to `0.16.0` and this line did not follow it, which is the cost D3 is about. The drift you spotted is now the point the bullet makes rather than a stale number sitting there unremarked. Say the word if you would rather it just carried the new number.

## 3 — the invented newline. Fixed, the byte-sink way rather than the documented-decision way.

You offered either; the sink is the one that makes the verb's central claim true instead of qualifying it. `ActanaCliDeps` gains a second pair alongside the line sinks:

```ts
outBytes: (chunk: string) => void;
errBytes: (chunk: string) => void;
```

and `core exec` relays through those, so `printf hello` writes `hello`. A second pair rather than widening `out`/`err`, because every other verb of this CLI emits lines and a line sink is what keeps them from each remembering a trailing newline. Both halves land in the same two arrays the line sinks do in `cli-harness.ts`, so `never-logs-a-blob.test.ts` still sweeps every byte this verb writes — that was the one thing worth checking before adding a sink.

Three tests, and the live suite got stricter rather than adjusted:

- the newline the command *did* emit survives (`hi\n` → `hi\n`);
- the newline it did **not** emit is not invented (`printf hello` → `hello`);
- byte-for-byte on both streams including an interior blank line.

In `core-exec-live.test.ts` the plain-stream assertion was `plain.out.join("\n")).toBe(bytes.slice(0, -1))` — it had to shave the trailing byte off the expectation to pass. It is now `plain.out.join("")).toBe(bytes)`, whole, plus a `printf %s no-newline` case. That `slice(0, -1)` was the finding you made, written down in the test's own expectation.

## 4 — the `unreachable` document. Fixed.

`OpenCoreResult`'s failure arm now carries the sentence it already wrote to stderr, and `core exec` puts it in the document:

```ts
if (args.json) deps.out(formatJson({ outcome: "unreachable", error: opened.error }));
```

All four outcomes have an `error`. This is not a hole in `openCore`'s stdout policy and its doc comment now says so explicitly: `openCore` still writes nothing to stdout, and which stream the words reach stays the caller's decision — what changed is only that a caller owing its consumer a document is no longer left with a bare exit code and no words to put in it. The other six call sites ignore the field.

## 5, 6, 7 — not in this push

I scoped this push to the two blocking items and the two you marked worth fixing. The three below are real and I am not disputing any of them; tell me whether you want them here or on their own ticket and I will do either.

- **5, the vacuous assertion at `drop-legacy-worktree-schema.test.ts:109`** — agreed, it passes because `columns()` returns `[]` for a table that no longer exists. Deleting it is the right call; the pre-cutover assertion at `:92` is the one carrying the weight.
- **6, `terminal.newTab` untested** — agreed, and the hotkey being the half that was silently broken before is the argument. The no-`routeCoreId` branch `preventDefault`ing into nothing is a fair hit too.
- **7, the handoff table** — `core-entry.ts` and `schema-bootstrap.ts` should be listed there so #265 is not surprised. Prose is not the table.

## Verification

Suites run per package rather than one recursive green, since `pnpm -r test` stops at the first failing package:

| package | result |
|---|---|
| `@actana/cli` | 456 passed, 3 skipped (was 453) |
| `@actana/core` | 1200 passed |
| `@actana/panel` | 1439 passed |
| `@actana/sdk` | 250 passed, 5 skipped |
| `@actana/shared` | 184 passed |

`core-exec-live.test.ts` run explicitly as well — 10 passed against a real socket. `pnpm typecheck` clean across all five. `pnpm lint`: 0 errors, 46 warnings, unchanged.

Two local-only notes so the numbers are not read as more than they are. `packages/panel` needs `--maxWorkers=2` on this box; at full parallelism a handful of 5-second integration timeouts fire, and the same files pass alone and at two workers. And the root `scripts` suite fails one launcher test here because this machine genuinely has `/opt/actana/app` installed, so the launcher resolves the real install root instead of the test's temp one — environment, not code, and untouched by this push. CI is the authority on both.

Still a draft, base still `chore/release-discipline-and-cursor-cli`, nothing merged.

mrfqcentic (approved):
Re-reviewed at `d308db4` on base `chore/release-discipline-and-cursor-cli`, reading the enforcing line in a fresh worktree at the PR head. Both blocking items are genuinely answered, both non-blocking ones are fixed, and the one place you did not do what I asked is the better call. **Approved.**

One new finding at the bottom that is nobody's regression but is now this verb's problem, and one note about the base branch having moved under you.

---

## 1. The connect deadline — RESOLVED

`packages/cli/src/core-exec.ts:116` is now `openCore(deps, args, paths, "actana core exec")` with no options at all, and `grep timeoutMs packages/cli/src/core-exec.ts` returns exactly one hit: the word inside the comment explaining why it is absent. The chain does what you say it does:

- `openCore` builds `connectOpts` from three optional fields (`core-connection.ts:200-203`); with no `opts` it hands `deps.connect` a literal `{}`.
- `connectCore` then takes `DEFAULT_CORE_TIMEOUT_MS` = 30 s (`core-connection.ts:43,114`) for `connectTimeoutMs` and `requestTimeoutMs` alike, which is the 30 s every other verb dials on.
- The 16-minute bound still applies to the request and applies **independently**: `client.request(frame, EXEC_TIMEOUT_MS)` at `core-exec.ts:150`, and `CoreClient.request`'s `timeoutMs` parameter defaults to `this.requestTimeoutMs` but is fully overridden when passed — it is the only timer on the pending entry (`core-client.ts:790-817`). I checked there is no second deadline underneath it: `core-link-transport.ts` has no per-request timer, only the heartbeat, which an idle-but-alive link keeps fed by design (`core-link-transport.ts:128`). So a 15-minute child still gets its 16 minutes, and a Core that accepts the socket and never finishes the handshake now gives up at 30 s like everything else.
- It matches `events tail`, which passes `{durable, storage, resolved}` and no `timeoutMs` (`events-command.ts:143-147`). Those are the only two call sites that pass options at all.

The regression test is real rather than decorative: `core.connectOptions` is the exact object the fake `connect` was handed (`cli-harness.ts:308`), so `expect(core.connectOptions).toEqual([{}])` at `core-exec.test.ts:174` fails the moment the option comes back. And `EXEC_TIMEOUT_MS`'s doc block now says *the request deadline only* in its first line (`core-exec.ts:59-62`), which is where the next person will look.

## 2. The changelog — RESOLVED, and the parity check with it

`CHANGELOG.md:57` in the live `0.1.0 — unreleased` section now reads *"The core-link protocol moves to 0.16.0 … and so `actana core exec` can run a command on a Core at all"*, and the fleet consequence is spelled out rather than left at the old sentence: *"a Core left on the previous protocol is grounded until it is updated, and the Panel says so rather than half-working."* Two paragraphs beside it cover the verb and the single New Terminal control, in the section's own voice. No `0.15.0` remains anywhere in `CHANGELOG.md`.

`docs/reference/cli-parity-check.md:383` is corrected to `CORE_LINK_PROTOCOL_VERSION = "0.16.0"` at line 1398, and I checked both halves of that claim against the file: the constant is at `core-link-frames.ts:1398` and it is `"0.16.0"`.

### The deviation, judged on its merits: `cli-parity-check.md:369` was right to keep `0.15.0`

You kept one of the two lines I called "nearly free" and explained why. Having read §5 in full, **you are right and I was wrong to lump the two lines together.**

Line 383 is a live claim about a file that exists, so a stale number there is simply false. Line 369 is not that. §5 is the post-mortem of `docs/reference/experiment-cli/src/protocol.ts` — a file that no longer exists in this tree (I checked) — and the bullet list is *"its drift was visible and deliberate"*: a pre-rebrand repo name in a comment, a Claude-specific field on a harness-agnostic type, a hand-maintained `RESULT_OF` table, and the hand-typed `PROTOCOL_VERSION`. Every other item in that list is a record of what the deleted file contained. Rewriting the version to `0.16.0` would have made the list say the mirror held a number it never held, and would have quietly deleted ADR 0025 D3's worked example of what a hand-copied schema costs — the drift *is* the example. The sentence you added instead makes that explicit and turns the number from something a reader might mistake for stale into the point the bullet is making. Keep it as it is; no change wanted.

## 3. The invented trailing newline — RESOLVED (fixed, not reasoned away)

You took the byte-sink option rather than the documented-caveat one, which is the one that makes the verb's central claim true instead of qualifying it. `writeStream` is gone from the package entirely, and `core-exec.ts:196-197` is `if (stdout) deps.outBytes(stdout)` / `if (stderr) deps.errBytes(stderr)`. `ActanaCliDeps` gains the pair with the reasoning on it (`cli-deps.ts:27-45`), and the real entry writes them unwrapped (`actana-cli-entry.ts:41-42`).

The one thing worth checking before adding a sink was whether it escapes the credential sweep, and it does not: `cli-harness.ts:201-208` pushes byte chunks into the same two arrays the line sinks use, and `never-logs-a-blob.test.ts` sweeps `all`, which is `[...out, ...err]` (`cli-harness.ts:255`). Coverage of this verb is unchanged.

The tests moved in the right direction too. `core-exec.test.ts` gained the missing case (`printf hello` → `hello`, no newline invented) plus a byte-for-byte case with an interior blank line, and the existing two switched from `toEqual([...])` on lines to `join("")` on bytes. And in `core-exec-live.test.ts:170` the `plain.out.join("\n")).toBe(bytes.slice(0, -1))` is now `join("")).toBe(bytes)` — that `slice(0, -1)` was the finding written into the test's own expectation, and it is good that removing the defect removed the shave rather than leaving it passing.

## 4. The `unreachable` document — RESOLVED

`OpenCoreResult`'s failure arm carries `error` (`core-connection.ts:151-162`), set on both failure paths (`:193-195`, `:211-213`), and `core-exec.ts:120` puts it in the document. All three failure outcomes now have the same shape — `refused` (`:225`), `link-lost` (`:247`), `unreachable` (`:120`) — so `.error` is never `undefined` on one of them. The `openCore` doc block explicitly states this is not a hole in the stdout policy (`core-connection.ts:176-181`), which is the right place for that sentence; the function still writes nothing to stdout, and the other six call sites ignore the field.

## 5, 6, 7 — deferred, and that is fine

You were right to scope the push and to say so rather than let them go quiet. My preference, not a condition: **5** (the vacuous `expect(columns(fresh, "user_terminals")).not.toContain("worktree_id")` at `drop-legacy-worktree-schema.test.ts:109`) and **7** (adding `core-entry.ts` and `schema-bootstrap.ts` to the handoff table) are a two-line delete and a PR-body edit with no conflict surface, so they cost less here than they do as a ticket. **6** (a test for `terminal.newTab`, and the no-`routeCoreId` branch that `preventDefault`s into silence) is real work and reads like its own follow-up. Either way, none of it blocks this.

---

## The four train checks, re-run at `d308db4`

1. **Commit body lines ≤ 132** — **PASS.** `git log --format=%B origin/chore/release-discipline-and-cursor-cli..HEAD | awk 'length > 132'` is empty across all four commits. Longest body line is 122, in `1eb8a9a`.
2. **Base** — **PASS.** Still `chore/release-discipline-and-cursor-cli`. Still a draft.
3. **No package manifest version touched** — **PASS.** `git diff origin/chore/release-discipline-and-cursor-cli...HEAD -- '*package.json'` is empty. No manifest of any kind is in the 44 changed files.
4. **Nothing belonging to another ticket in this train** — **PASS.** `.github/workflows/ci.yml` and `docs/rulesets/` (#257, #264): untouched — nothing under `.github/` at all. `AGENTS.md`, `CLAUDE.md`, `docs/ci-cd.md` (#260): untouched. The only two non-package files are `CHANGELOG.md` and `docs/reference/cli-parity-check.md`, both of which are this review's own request.

CI on `d308db4` is green on all 17 required checks, `Train rules` and `Conventions` included.

## #260 landed on the base branch and you have not rebased — no conflict, textual or semantic

Worth stating explicitly since it happened after your push. `chore/release-discipline-and-cursor-cli` has moved one commit past this branch's merge-base (`e6d2fd9`): `79a2718` — *"docs: put the release-train rules where an agent will actually read them (#267)"*, `Refs #260`, merged at 12:12Z.

**They do not conflict.** `79a2718` touches exactly four files — `AGENTS.md`, `CLAUDE.md`, `CONTRIBUTING.md`, `docs/ci-cd.md` — and this PR touches none of them. The file sets are disjoint, so a rebase or merge is textually clean; GitHub agrees (`mergeable: MERGEABLE`).

**And nothing #260 wrote about terminals, shells or the CLI is contradicted here, because it wrote nothing about them.** I read the whole diff. Its content is release-train discipline and nothing else: an *AGENTS.md* §"Release trains" (a train is cut by a workflow, promotion consumes a green approved PR into `main`), a three-line `CLAUDE.md` that is an `@AGENTS.md` import, a `docs/ci-cd.md` §"Re-cutting a train, by hand", and one `CONTRIBUTING.md` sentence. There is no statement about a terminal, a shell, `core shell`, `core exec` or any CLI verb anywhere in it. The single string in it that overlaps this PR's surface at all is `packages/{cli,core,panel,sdk,shared}/package.json` inside the six-manifest re-cut command — and this PR changes no manifest, so it does not even brush it.

The one place they touch is in your favour: `CONTRIBUTING.md` now says body lines are capped at **132** rather than the stale 100, which is the number this branch was already measured against. A rebase is housekeeping, not a merge risk.

---

## New, non-blocking, and not caused by this push: `process.exit` can truncate a large `core exec` stdout

I missed this at `b2abce2` and it is a property of the shared entry point rather than of anything you changed in `d308db4`, so it is not a condition of this approval — but it lands squarely on this verb and I would rather you heard it from me than from a script.

`actana-cli-entry.ts:76` is `process.exit(code)`, called immediately after `runActanaCli` resolves. When stdout is a **pipe** — the case `core exec` exists for — Node writes to it asynchronously, and `process.exit()` drops whatever has not drained. Measured in this container:

```
$ node -e 'process.stdout.write("x".repeat(4*1024*1024)+"\n"); process.exit(0)' | wc -c
65536
```

One pipe buffer, then silence. The Core's bound is 8 MiB, so this is inside the range the verb explicitly supports. `deps.outBytes` did not introduce it — the old line sink and the `--json` document have the same exposure — but before this PR no verb routinely wrote megabytes to stdout.

Severity is uneven across the two paths, which is why it is not blocking:

- `--json` piped to a consumer fails **loudly**: a truncated document is a parse error, not a wrong answer.
- The plain path truncates **silently**: `actana core exec -- cat big.log | wc -c` under-reports with the command's own exit status intact.

That second one is the shape §B7 built the 8 MiB refusal to avoid — *"a stdout that looks complete and is not"* — arriving through a different door at 64 KB. The fix is small and lives in a file this PR already touches: await the drain before exiting, or set `process.exitCode` and let the loop empty. Your call whether it belongs here before this leaves draft or on its own ticket alongside 5 and 6; if it goes on a ticket, it deserves a line in the CLI README next to the pipe-safety claim.

---

Everything I asked for is done, the one thing you did differently is done better, and the base moving does not disturb any of it. Approved.

PR #272, opened 2026-08-19
fix: make cursor-cli drivable — right binary, auto mode delivered or refused, trust dialog recognised, turn-start said

qcentic-dev:
Closes the four failure points in #177 that this repository can close, plus the
enabling defect under finding 2 that the issue calls out as *"squarely
`actana/control`'s"*.

## The issue is out of date in one place, and this PR translates it

#177 splits the work into a "client" half in `experiment/src/commands/session.ts`
— *"out of this repo, unversioned scratch tooling"* — and a "repo" half in
`packages/shared` and `packages/panel`. That split no longer exists. `experiment/`
was retired on 2026-08-19 and its successor is `packages/cli` in this repository,
published as `@actana/cli`. **Findings 1, 2, 3 and 5 are therefore all in scope
here**, and none of them is left unfixed on the grounds that the issue assigned
them elsewhere.

Findings 1, 2 (client half) and 5 turned out to have landed already with the CLI
and SDK extraction, exactly as the issue's own follow-up comment records. This PR
does not re-fix them — it **proves** them, against the Core's own spawn policy,
in `packages/cli/src/__tests__/cursor-cli-drivable.test.ts`. See the verification
table below for which claims are run and which are read.

### 1 — argv[0] is the canonical binary, never the harness id

Already true (`HARNESS_LAUNCH_COMMANDS["cursor-cli"] === "cursor-agent"`), and now
tested. The new suite pushes the command this client would actually send through
`resolveSpawnPlan` — the same function `packages/core` calls before spawning —
for **every** harness rather than for cursor-cli alone, because codex and
opencode passed the old client by the id/binary coincidence and a table that is
right only by luck is a table that will drift.

### 2 — auto mode reaches the harness, or the spawn is refused

The client half was already right. The repo-side enabling defect was not:
`validateHarnessArgv` checked the gesture in one direction only, so a flag with
no option was rejected and **an option with no flag passed silently**. The Core
launched an interactive cursor-agent while the caller believed the session was
unattended.

`packages/shared/src/pty-spawn-policy.ts` now rejects that with a new
`auto-mode-flag-missing` code whose message names the missing flag. OpenCode
still launches with the option set — it ships no auto-mode flag at all, which is
a `null` in the table rather than a carve-out in the check.

The flag also gets one home. `HARNESS_AUTO_MODE_FLAGS` is derived from
`HARNESS_CLI_CONFIG`; `HARNESS_ARG_RULES`, `HARNESS_REGISTRY` and the CLI's
[...]

Issue #264, opened 2026-08-18
ci: a promotion gate pull request into main is hand-mergeable, and nothing in the repository stops it


mrfqcentic:
> **This is the guard #260 asks for and does not build.** #260's failure 3 records the
> event; this issue is the mechanism. One release has already been lost to it, and
> nothing about the repository has changed since.

The promotion pull request from a train into `main` is a **gate**: a place for checks and
an approval to accumulate, consumed by `promote.yml`, which advances `main` by
fast-forward. It is not a merge. GitHub does not know that, and neither does the
repository: on a `beta/* → main` pull request the merge button is present, enabled, and
green.

On 2026-08-18 it was pressed. `v0.3.1` was never released.

## What happened, and what it cost

#259 (`beta/0.3.1` → `main`, *chore(release): promote 0.3.1*) was squash-merged from the
GitHub UI at **18:09:09Z** by `mrfqcentic`, instead of being promoted by `promote.yml`.
The squash produced `3b14453` on `main`. Its tree is byte-identical to the train tip
`d438e5ce`:

```
$ git rev-parse 3b14453333ec8305c9404d181b7fe1bd79592c7c^{tree} \
                d438e5ce9caacc24d47526957eeb0cf2e5d27d56^{tree}
920a12a556d3dc155aef0e5602c26bdfc041bb96
920a12a556d3dc155aef0e5602c26bdfc041bb96
$ git diff d438e5ce 3b14453                                    # empty
$ git merge-base --is-ancestor d438e5ce origin/main; echo $?
1
```

Same bytes, different SHA. `d438e5ce` is therefore not an ancestor of `main`, and the SHA
the ADR 0023 D16 assertion verifies against is permanently unreachable from the released
history. The last successful train publish, run
[32143263819](https://github.com/actana/control/actions/runs/32143263819) (13:35:38Z), was
built from `d438e5ce` — so `beta-0.3.1`'s `org.opencontainers.image.revision` label is
`d438e5ce`, and no commit on `main` carries that SHA.

That left `v0.3.1` releasable by no path:

| Path | Refuses because |
| --- | --- |
| `promote.yml` with `train=beta/0.3.1` | `promote.yml:199` — the gate pull request is closed, and D5 makes its absence fatal |
| `release.yml` in **promote** mode | `release.yml:224` resolves the mode as promote (the tag would be reachable from `main`), and the retag is refused unless the digest's revision label equals the tagged SHA — it is `d438e5ce` (D16) |
[...]

Issue #260, opened 2026-08-18
docs: put the release-train rules where an agent will actually read them


mrfqcentic:
> **Take this first in the `0.3.3` train.** Six failures on 2026-08-18 trace to this gap,
> four of them after this issue was written, and one of them cost an entire release.

`AGENTS.md` is the file an agent reads first, and it is 17 lines covering the issue
tracker, triage labels, domain docs and upstream harvesting. It says nothing about
release trains and links to neither `CONTRIBUTING.md` nor `docs/ci-cd.md`. There is no
`CLAUDE.md` anywhere in the repository.

The rules are written down — carefully, and in the right places for a person reading the
pipeline. They are just not reachable from where an agent starts. Six failures today came
out of that gap. The first two, seven hours apart on the same train, are below; four more
followed on the `0.3.1` and `0.3.2` trains and are recorded after them.

## The two failures

**1 — `beta/0.3.1` was cut with plain `git branch`, so six manifests kept `0.3.0`.**

`promote.yml` auto-cut `beta/0.4.0` correctly on 2026-08-14 (`actana-release-train[bot]`,
09:44:47Z). The next release was a patch, so on 2026-08-18 that branch was deleted
(10:38:26Z) and `beta/0.3.1` created by hand two minutes later (10:40:14Z) — a bare branch
creation with no cut commit and no version stamp.

The `Train rules` check on the train pull request (#250) failed at 12:56:10Z with **one
annotation per manifest**, six of them:

> **Version drift in `packages/shared/package.json`** — `packages/shared/package.json`
> says 0.3.0; the train is 0.3.1. Every manifest carries the train's version (ADR 0023 D3,
> amended by #152 and #157) — the cut wrote them and nothing else should be editing them.

…and the same for `package.json`, `packages/cli`, `packages/core`, `packages/panel`,
`packages/sdk`. Fixed after the fact by `54e7646` *chore(release): cut beta/0.3.1*
(13:05:21Z), which wrote what the cut should have written. **Cost:** a red required check
on the open train for ~9 minutes, a recovery commit that exists only to undo a manual
step, and a train whose first commit is a repair rather than a cut.

**2 — `promote.yml` dispatched with no gate pull request open.**

Run [32168144238](https://github.com/actana/control/actions/runs/32168144238), dispatched
17:54:55Z with `train=beta/0.3.1`. *Approve the promotion* succeeded — **the human pause
was spent** — and *Resolve the promotion* then failed with `promote.yml:199`:

[...]

PR #275, opened 2026-08-19
fix(core): wait for cursor-cli's and claude-code's composer before typing the start prompt

### LINKED ISSUE #232: cursor-cli: start prompt intermittently swallowed — the generic quiet-output settle races the TUI boot (same class as #229)
## What happened

The start-prompt race fixed for opencode in #229 also affects cursor-cli, intermittently. Three cursor-cli Sessions started back-to-back on the same Core (tip with the #229/#230 fixes):

- settle at ~3.1 s → prompt landed, turn ran, `finished` ✅
- settle at ~1.6 s → prompt silently swallowed; Cursor Agent sits at its idle "Plan, search, build anything" screen, Session stuck `ready` ✅→❌
- settle at ~2.5 s → prompt landed, turn ran, `finished` ✅

The Core's prompt-delivery log shows the difference plainly: the failed run's output-quietness settle fired earlier than the TUI was ready for input. The new readiness wait (`prompt-delivery.waiting-for-composer`) only guards opencode; cursor-cli still relies on the generic quiet-output settle, so it wins or loses by boot-time luck.

## What I expected

Start-prompt delivery to wait on a harness-readiness signal for cursor-cli the way #229's fix does for opencode — not on output quietness with a fixed pause.

## Steps to reproduce

1. Core with the #229/#230 fixes, cursor-cli installed.
2. `actana session start <project> "say hello" --harness cursor-cli --wait` repeatedly (cold starts help — the failure follows a fast settle on a slow TUI boot).
3. Intermittently the Session stays `ready`, the transcript shows Cursor Agent's idle screen with no prompt, and `--wait` times out unreported.

## Additional context

- Found while re-verifying #229/#230 during the QA retest of the fix round; roughly 1 in 3 in this sample.
- Same failure signature as #229 (delivered-but-swallowed); different harness, so the readiness-detection needs a cursor-cli signal rather than the opencode composer probe.
- claude-code and codex were not exercised in this sample; worth checking whether they sit on the same generic settle path.

qcentic-dev:
Closes #232.

Start-prompt delivery for `cursor-cli` now waits on a cursor-cli **readiness signal** the way #229's fix does for opencode — not a longer pause, not a wider quiet window. `claude-code` gets the same treatment, because the evidence gathered while this was open shows it is exposed at the same rate.

## The bug, and why timing was never the answer

`HARNESS_READINESS` — the table #229 added, whose hold is logged as `prompt-delivery.waiting-for-composer` — had exactly one row, opencode. `cursor-cli` fell through to the generic quiet-output settle, so it won or lost on TUI boot-time luck: three back-to-back Sessions on one Core settled at 3.1 s, 1.6 s and 2.5 s, and the 1.6 s one had its prompt swallowed, Cursor Agent left on its idle *Plan, search, build anything* screen, the Session stuck in `ready` and `--wait` timing out unreported. Roughly one start in three.

The reason a bigger constant does not fix this is that the runs that worked and the run that failed took the **same code path** and differed only in how fast the TUI booted. Measured spreads: opencode 3.4–6.9 s (from #229), cursor-cli 1.6–3.1 s, claude-code as below. A gap wide enough for the slow boot is a tax on every fast one.

## `claude-code` — yes, it is on the same path, and it is fixed here

The issue's last line asks whether `claude-code` and `codex` sit on the same generic settle path and are exposed to the same race. Read from the code and, for claude-code, measured:

| harness | on the generic settle path before this PR? | exposed? | done here |
|---|---|---|---|
| `opencode` | no — the only `HARNESS_READINESS` row | no | unchanged |
| `cursor-cli` | **yes** | **yes** — 1 in 3, per the issue | **fixed** |
| `claude-code` | **yes** | **yes** — 8–10 losses in ~25 starts across 3 Cores | **fixed** |
| `codex` | **yes** | **unverified** | not fixed — see below |

**claude-code, yes.** `readinessFor("claude-code")` returned the empty default, so `composerOnScreen` was unconditionally `true` and delivery typed on the quiet gap alone. The field evidence attached to the issue puts it at the same one-in-three: ~25 `session start` runs with `--harness claude-code` across all three Cores in the fleet lost the prompt 8–10 times, with the signature that makes it the same bug and not a different one — nothing errors anywhere, the Core reports `ready` rather than `running`, the harness sits on its idle screen with the text nowhere on it, `--wait` times out unreported (one Session sat 46 minutes), and re-sending the identical bytes with `session send --enter` lands immediately and runs the turn. Only *when* the first write happened was ever wrong.

I captured that hole live on this machine, `claude` 2.1.235 under `script --log-timing`:

```
  435 ms  the folder-trust dialog
 5051 ms  the dialog acknowledging the digit — content, so a paint; no composer
 5401 ms  the 350 ms quiet gap expires. THIS is where the prompt was typed.
 5623 ms  the composer, 222 ms later
15635 ms  the `⏵⏵ … (shift+tab to cycle)` footer, ten seconds behind the composer
```

Both bracketing chunks are committed as `packages/core/src/__tests__/fixtures/claude-code-2.1.235-{trust-ack,composer}.txt` (account name and paths substituted) and the suite replays them at exactly those timings. `paintedSinceKeystroke` does not close this: the 5051 ms frame *is* a paint with content in it.

**codex — unverified, and stated as a gap rather than a clearance.** It is on the generic path, and nothing in the issue's sample or in today's evidence exercised it, so I am claiming nothing about it either way. No codex screen has been read, and this table's rule is that a row is a transcription of a screen somebody looked at — inventing one would be the guess ADR 0026 D5 exists to refuse. Recorded in the code comment, in the ADR amendment, and in a test that asserts codex still takes the untouched path, all three saying the same thing: **a codex row is owed as soon as somebody captures its composer**, not unnecessary. Fixing it stays outside this ticket's lane; filing it does not, so it is now **#277**, linked from both the table note and the ADR amendment. That issue carries what would settle it: capture codex's composer off a live PTY, time the quiet gap against the frame the composer actually arrives in, and add the row — or record a measured "no row needed", which closes it just as well.

## What changed

`packages/core/src/harness-prompt-delivery.ts` — two rows added to `HARNESS_READINESS`. No new mechanism, no new knob, no timing constant touched:

```ts
"cursor-cli":  { composer: [/plan,\s*search,\s*build/i], confirmEcho: true, maxPromptWrites: 3 },
"claude-code": { composer: [/\btry\s+"/i],               confirmEcho: true, maxPromptWrites: 3 },
```

- The markers are the composers' **own placeholders**, the same shape as opencode's `Ask anything`. `Try "` is on both committed claude-code builds — 2.1.228 in bypass mode and 2.1.235 in auto mode, two builds and two permission modes.
- **A marker was rejected on a measurement.** `⏵⏵ … (shift+tab to cycle)` is on claude-code's composer screen and reads like an obvious signal; it was observed at 15 635 ms in one capture, past the 15 s backstop — and on the reviewer's own 2.1.235 capture it arrived in the *same frame* as the composer. The variance is the argument, not the single number: a marker that is sometimes ten seconds late delays exactly the run it is late on. The comment and the ADR now say "observed in one capture" rather than implying it always lands there. There is a test asserting it is *not* a marker.
- `confirmEcho` on both means a write that left no echo returns to `settling` and is typed again rather than submitted into nothing. That is the client-side workaround in use today — start, wait ~12 s, ask whether the Session is still `ready`, re-send, ~24 s to recover — **moved to the layer that can see the screen**, which is ADR 0026's founding argument. I have not reimplemented it in the Core by accident; it is D6a, and it is inside the same delivery with nothing to poll for.
- The failure mode this is allowed to have is unchanged and tested: a reworded placeholder stops matching and the backstop delivers **late, not never**.

`docs/adr/0026-…md` — an amendment. D3a said `claude-code`, `codex` and `cursor-cli` "have no entry and are therefore unchanged, which matters because a cursor-cli start prompt was verified working on the very build #229 was filed against". That was one working start, not a rate, and the record now says so.

## Building on #177, not reverting it

I read `ada63c2` before editing and it is intact. Two things it left me:

- **Its `folder-trust` row for cursor-cli is untouched and still wins.** Readiness is checked *after* the dialog table, so cursor-agent's letter-keyed menu still yields `answer: null` and delivery still abandons to `needs-input` rather than typing into the dialog. There is a test in this PR asserting exactly that, so the new gate cannot quietly erode it.
- **Its turn-start signal is not reusable here, and this is worth being explicit about since the ask was to reuse it if it fit.** It does not, for two independent reasons. `HARNESS_HOOKS["cursor-cli"]` is `reportsTurnStart: false` — `beforeSubmitPrompt` does not fire in cursor-agent, which is the very thing `session start` now warns about. And even where a harness does report it, a turn starts *because* the prompt landed: it is downstream of the write this code has to decide whether to make. A readiness gate needs a signal that precedes the keystroke, which is what a composer marker is.

## What I ran, and how many times

A race is not fixed by one passing run, so the fix was measured by removing it:

- **Without the two rows** (rows deleted, everything else identical): `packages/core` prompt-delivery suite fails **the same 10 cases, 3 runs out of 3** — byte-identical failure lists. The decisive one is `types nothing into the 572 ms hole the capture measured`, which without the gate writes `["say hello"]` into the hole at 5401 ms, exactly as production did.
- **With the fix**: **84/84 passing, 5 runs out of 5.**
- The cursor-cli cases **simulate a slow TUI boot** rather than hoping for a fast one — the composer is parameterised (1.4 s vs 4.0 s) and one case holds through a 10-second stall, two thirds of the backstop and 28× the quiet gap, asserting not one byte goes out.
- Per-package, run explicitly because `pnpm -r` bails on the first failure and `packages/cli` sorts before `packages/core`: `@actana/shared` 188/188 ✅, `@actana/sdk` 253 passed + 5 skipped ✅, `@actana/cli` 468 passed + 3 skipped ✅, `@actana/core` **1224/1224** ✅.
- `pnpm -r typecheck` ✅. `pnpm lint` — 0 errors (46 pre-existing warnings, none in a file this PR touches). `pnpm scan:secrets` ✅.
- `@actana/panel`: 1 failure under a full local run (`cores-api.test.ts`), which **passes in isolation** and is pre-existing — I stashed this branch and reproduced it on the base commit, where the same full run fails **2** panel tests. No panel source file is touched by this PR.
- Commit message checked against this repo's own `commitlint.config.mjs` in an isolated install: passes. Body lines measured with `awk`; longest is 84 characters.

## Boundaries

Nothing outside the Core-side prompt-delivery path was edited. Untouched: `packages/cli/src/core-resolution.ts`, `packages/cli/src/__tests__/no-local-escape.test.ts`, the CLI skill assets and `packages/shared/src/harness-cli-config.ts` (#265); `.github/workflows/ci.yml`, `docs/rulesets/`, `AGENTS.md`, `CLAUDE.md`, `docs/ci-cd.md` (#257, #264, #260). No acceptance criterion needed a fenced file.

## The cursor-cli marker now has a fixture — and where those bytes came from

This was the one row in the table with no fixture behind it, and the review closed it. **The capture is the reviewer's, not mine.** cursor-agent 2026.08.11-e8db854 is installed on this machine — the same build as the committed trust fixture — and it still stops at its sign-in screen without credentials, which I do not have and did not seek; I re-checked while addressing the review and it goes straight to `Signing in`. The reviewer has a signed-in install of that same build, reported the idle screen as `→ Plan, search, build anything`, and confirmed the marker matches it on 5 of 5 cold boots.

That text is committed as `packages/core/src/__tests__/fixtures/cursor-agent-2026.08.11-e8db854-composer.txt` and `CURSOR_IDLE_COMPOSER` is now read from it instead of from a literal. It arrived as text rather than as bytes, so it carries no escape sequences and nothing frames it — the same route and the same limits as the cursor-agent trust fixture from #272, and the provenance is stated that way in both the module comment and the test file rather than being blurred into "captured".

Worth noting what the real screen corrected: the literal it replaces had guessed `> Plan, search, build anything`; the actual glyph is an arrow. The marker is wording-independent enough that both match, but the file now holds the screen rather than my reconstruction of it.

The fixture is load-bearing rather than decorative — rewording it or emptying it turns 84/84 into 3 failures, and restoring it returns the file byte-identical and the suite green.


qcentic-dev:
## Acceptance criteria — issue #232

**A note on where this list comes from:** #232 carries no formal acceptance-criteria block — it is written as *What happened / What I expected / Steps to reproduce / Additional context*. So each item below is lifted verbatim from a sentence in the issue (or from the field evidence attached to it), rather than invented. Nothing is being quietly graded against an easier bar.

- [x] **"Start-prompt delivery to wait on a harness-readiness signal for cursor-cli the way #229's fix does for opencode."** `HARNESS_READINESS` gains a `cursor-cli` row whose composer marker is cursor-agent's own idle placeholder. Same table, same mechanism, same `waiting-for-composer` event as #229.
- [x] **"…not on output quietness with a fixed pause."** No timing constant is touched. `quietGapMs`, `maxWaitMs`, `submitBaseMs`, `submitPerCharMs`, `submitMaxMs` are all unchanged, and `HARNESS_PROMPT_DELIVERY_PROFILES` is still empty. The change is two rows in a readiness table.
- [x] **"The readiness-detection needs a cursor-cli signal rather than the opencode composer probe."** The opencode probe (`ask anything`) is untouched and does not apply; cursor-cli gets `Plan, search, build`, its own placeholder. See the caveat at the bottom of this list.
- [x] **The failure signature is gone: no Session left in `ready` with the prompt swallowed.** `confirmEcho` is set for cursor-cli, so a write that left no echo returns to `settling` and is typed again rather than submitted into nothing and logged as `delivered`. Test: *re-types rather than submitting into a composer that swallowed the write*.
- [x] **"…worth checking whether claude-code and codex sit on the same generic settle path."** Answered explicitly in the PR body with a per-harness table: `cursor-cli` **yes**, `claude-code` **yes**, `codex` **yes on the path / unverified for exposure**, `opencode` no. Read out of the code, not guessed.
- [x] **claude-code — fixed here, not deferred.** The field evidence (8–10 losses in ~25 starts across three Cores) plus a live `script --log-timing` capture of `claude` 2.1.235 taken on this machine: answered dialog repaints at 5051 ms, composer at 5623 ms, quiet gap expires at 5401 ms — 222 ms early. Both chunks committed as fixtures; the suite replays them at those timings.
- [ ] **codex — deliberately NOT fixed, and this box is unchecked on purpose.** It is on the generic settle path, but nothing in #232's sample or in today's evidence exercised it and no codex screen has been read. This table's rule is that a row transcribes a screen somebody looked at; inventing one is the guess ADR 0026 D5 refuses. Recorded as a gap owed a row — in the code comment, in the ADR amendment, and in a test asserting codex still takes the untouched path — not as a clearance. **Belongs in a follow-up issue**, which I have not filed, to stay inside this ticket's lane.
- [x] **"A race is not fixed by one passing run" — proven by a test that fails deterministically without the change.** Rows removed, everything else identical: **the same 10 cases fail, 3 runs out of 3**, byte-identical failure lists. Restored: **84/84 passing, 5 runs out of 5**. The cursor-cli cases **simulate a slow TUI boot** (composer parameterised at 1.4 s vs 4.0 s, plus a 10-second stall — 28× the quiet gap) rather than hoping for a fast one.
- [x] **#177 built on, not reverted.** `ada63c2` read before editing and intact. Its `folder-trust` row still wins, because readiness is checked *after* the dialog table — with a test in this PR asserting cursor-agent's trust screen still abandons. Its turn-start signal was evaluated for reuse and **does not fit**: `HARNESS_HOOKS["cursor-cli"]` is `reportsTurnStart: false`, and a turn start is downstream of the write this code has to decide whether to make.
- [x] **Every package's tests run explicitly, not via one `pnpm -r`.** shared 188/188, sdk 253+5 skipped, cli 468+3 skipped, core **1224/1224**. `pnpm -r typecheck` clean; `pnpm lint` 0 errors; `pnpm scan:secrets` clean. Panel: 1 failure under a full local run, **passes in isolation**, and pre-existing — reproduced on the base commit, which fails **2**. No panel source file touched.
- [x] **Parallel-ticket boundaries respected.** Nothing edited outside the Core-side prompt-delivery path. Untouched: `packages/cli/src/core-resolution.ts`, `packages/cli/src/__tests__/no-local-escape.test.ts`, the CLI skill assets, `packages/shared/src/harness-cli-config.ts` (#265); `.github/workflows/ci.yml`, `docs/rulesets/`, `AGENTS.md`, `CLAUDE.md`, `docs/ci-cd.md` (#257, #264, #260). **No acceptance criterion needed a fenced file**, so nothing was skipped on boundary grounds.
- [x] **Draft PR against `chore/release-discipline-and-cursor-cli`.** Not `main`, not `beta/0.3.3`. Left as a draft; nothing merged; no branch other than `fix/cursor-cli-prompt-readiness` touched.

### One qualification, stated rather than buried

The `cursor-cli` marker is the **only row in the readiness table with no byte fixture behind it**. I installed cursor-agent 2026.08.11-e8db854 here — the same build as the committed trust fixture — specifically to capture its composer, and it stops at a sign-in screen without credentials, which I did not have and did not seek. So `Plan, search, build` is quoted from #232's own account of the failed run rather than read off a capture. If the real screen words it differently, the marker stops matching and the 15 s backstop delivers the prompt **late rather than never** — strictly better than the one-in-three loss it replaces, and the direction this module is always wrong in. A capture from anyone with cursor-agent signed in would upgrade that row from quoted to observed.

mrfqcentic (changes requested):
## Review — PR #275 against issue #232

Reviewed at `396c834` in a detached read-only worktree off `origin/chore/release-discipline-and-cursor-cli`. Everything below was checked by running it, not by reading the description. **The fix itself is right, proven, and in scope.** I am requesting changes for three documentation-accuracy items inside the file this PR is about — no behaviour change is needed.

### The claim that mattered most: does the test fail deterministically without the change?

Yes. I removed the two `HARNESS_READINESS` rows in my worktree copy and ran the suite three times per variant:

| variant | result | runs |
|---|---|---|
| as submitted | **84/84 pass** | 3/3 |
| both rows removed | **10 fail / 74 pass**, byte-identical failure lists | 3/3 |
| only `cursor-cli` row removed | **5 fail** | 3/3 |
| only `claude-code` row removed | **6 fail** | 3/3 |

The split matters: each harness is independently enforced, so `claude-code` is genuinely covered by the mechanism rather than mentioned alongside it. The decisive case, `types nothing into the 572 ms hole the capture measured`, goes red without the rows. Restored, clean; my worktree is back to `396c834` with an empty `git status`.

And the slow-boot test exists rather than being hoped for: `holds through a TUI that boots far slower than any pause would allow` stalls 10 s after the banner and asserts **zero** bytes go out, and `delivers on a slow boot and on a fast one` parameterises the composer at 1.4 s and 4.0 s. Both are in the red set.

### Readiness signal, or a longer pause? — readiness signal.

The enforcing line is `packages/core/src/harness-prompt-delivery.ts:989`:

```ts
if (!this.deadlinePassed && !composerOnScreen(this.screen, this.readiness)) {
  this.holdForComposer();
  return;
}
```

It gates on a marker being *on screen*, emits `prompt-delivery.waiting-for-composer`, and holds — the same table and the same event #229 added for opencode. `quietGapMs` (350), `maxWaitMs` (15 000), `submitBaseMs`, `submitPerCharMs`, `submitMaxMs` and `HARNESS_PROMPT_DELIVERY_PROFILES` are all byte-identical to base; the entire production diff is two table rows plus comments. Not a longer pause, not a wider quiet window.

### Exercising the real thing

`cursor-agent 2026.08.11-e8db854` is installed **and signed in** on this machine, so I could do what the PR says it could not.

- **The `cursor-cli` marker is correct.** Real screen, 5/5 cold boots: `→ Plan, search, build anything`. `/plan,\s*search,\s*build/i` matches it. The one weak point in this PR is not weak — it is just unverified in the PR, and I have now verified it.
- **37 real cursor-agent starts** driven through the actual `HarnessPromptDelivery` state machine over a live pty (19 with the fix, 18 with the rows removed), 18 of them under 24-way CPU load to stretch the boot (composer at 2.5–3.2 s instead of 0.9–1.1 s). **37/37 landed**, every one with exactly `["say hello", "\r"]` — no retype, no double-typed prompt from the new `confirmEcho`.
- **7 real `claude` 2.1.235 starts** the same way (3 with the fix, 4 without): 7/7 landed, dialog answered with `1`, one prompt write each. The placeholder was `Try "edit <filepath> to..."` — a third rendering the marker matches.
- **Honest negative:** I could not reproduce the swallow on this machine — the unfixed build landed too. That is expected for a 1-in-3 race on a fast box, and it is why the deterministic unit failure above is the proof, not the live runs. What I *did* reproduce is the hole: on my own `claude` 2.1.235 capture the trust answer went out at 932 ms and the next content frame — the composer — arrived at **1405 ms**, a 473 ms silence that the 350 ms quiet gap fires straight into. Same shape as the PR's 572 ms, independently measured.

### Scope, boundaries, train rules

- Base is `chore/release-discipline-and-cursor-cli`. ✅
- 5 files, all inside the Core prompt-delivery path plus the ADR. **Zero overlap** with #276, the only other open PR on this base. ✅
- No `.github/` change at all, so ADR 0023 D33 is not in play. ✅
- Commit body: longest line **90** ≤ 132; header 90 ≤ 120; `Refs #232` has its leading blank line. `git log --format=%B base..HEAD | awk 'length > 132'` is empty. ✅
- All 19 required checks green on the head commit; the 4 skipped jobs are pre-existing publish/train jobs, not added here. ✅
- Builds on #177 (`ada63c2`, in base): `HARNESS_HOOKS["cursor-cli"].reportsTurnStart` is still `false`, the `folder-trust` row is untouched, readiness is checked *after* the dialog table, and there is a test asserting cursor-agent's trust screen still abandons. Nothing reverted, nothing duplicated. The rejection of the turn-start signal is correct: it is downstream of the write this code has to decide whether to make. ✅
- `codex` is described as **unverified** in the code, the ADR and a test that pins it to the untouched path — not claimed either way. That is the right call. ✅

---

## Changes requested

**1. `packages/core/src/harness-prompt-delivery.ts:456-458` — the comment now contradicts the code it introduces.**

```
 * behaviour every harness had before issue 229 — settle, type, submit — which
 * is what keeps `claude-code`, `codex` and `cursor-cli` byte-for-byte
 * unchanged by a fix that was only ever about opencode.
```

Two of those three are in the table forty lines below. This is the doc comment on `HarnessReadiness` itself — the first thing anyone reads before adding the codex row this PR says is owed.

**2. `packages/core/src/harness-prompt-delivery.ts:916-917` — same problem on `echoConfirmed`.**

```
 * `true` for every harness with no `confirmEcho` entry, so this is a no-op
 * for the three families that were never affected.
```

"the three families that were never affected" is exactly the claim this PR disproves, and two of them now set `confirmEcho: true`. In a module whose argument is that the comment is the record, these two sentences are the record being wrong.

**3. `packages/core/src/harness-prompt-delivery.ts:513-519` and `packages/core/src/__tests__/harness-prompt-delivery.test.ts:246-260` — the cursor-cli provenance note is now out of date, in the good direction.**

Both say the composer "could not be captured here" because cursor-agent stops at a sign-in screen. On a signed-in install of the same build, `2026.08.11-e8db854`, it does not: the idle screen reads `→ Plan, search, build anything` and the marker matches it on 5/5 cold boots. Please capture it and commit it as `fixtures/cursor-agent-2026.08.11-e8db854-composer.txt` beside the claude-code pair, drive `CURSOR_IDLE_COMPOSER` off the fixture, and reword both notes. That closes the only row in the table without a byte fixture — the one caveat the PR itself calls out — and it is a fifteen-minute change now that the screen is known to be reachable.

### Non-blocking notes

- **The 15 635 ms footer claim is one observation stated as a property** (`harness-prompt-delivery.ts:538-541`, ADR amendment "A marker was rejected on a measurement"). On my `claude` 2.1.235 capture the `⏵⏵ … (shift+tab to cycle)` footer arrived **in the same frame as the composer**, not ten seconds later. The rejection still stands — a marker that is *sometimes* 10 s late is unusable — but the ADR is a record, so please say "observed at 15 635 ms in one capture" rather than implying it always lands there.
- **The codex follow-up issue is still unfiled**, and the author's own checklist leaves that box unchecked deliberately. Agreed it is out of this ticket's lane to *fix*; filing the issue is not. Please open it before this merges so "owed a row" has an address rather than three comments pointing at each other.
- `harness-prompt-delivery.ts:572-573` has a stray double blank line (and one before `describe("promptEchoed")` in the test file). eslint is clean, so cosmetic only.

Nothing else. Fix the two stale comments and land the cursor fixture and this is an approve — the mechanism, the evidence and the scope discipline are all right.

qcentic-dev:
Thanks — the mutation test (both rows out → 10 failures, byte-identical across 3 runs; 5 and 6 for each row alone) and the 44 live starts are more than I could run here, and the split proving each harness is independently enforced is the part I most wanted checked. All three blocking items are documentation accuracy inside this file and all three are fixed in 95b3778. No production line changed: the diff is still the two `HARNESS_READINESS` rows.

**1. The `HarnessReadiness` doc comment (was `:456-458`).** Fixed. It said the default is "what keeps `claude-code`, `codex` and `cursor-cli` byte-for-byte unchanged" — with two of the three in the table forty lines below. It now says what the code does: #229 added opencode on that reading, #232 measured the same lost prompt on `cursor-cli` and `claude-code` and disproved it for two of them, and `codex` is the only harness left on the default — unverified rather than cleared, with a pointer to the table's own note before anyone adds to it.

**2. `echoConfirmed` (was `:916-917`).** Fixed. "A no-op for the three families that were never affected" was the claim this PR disproves. It now names `codex` as the only harness without `confirmEcho`, and says it is there because nobody has read its screen — not because it was shown to be unaffected — with the other three attributed to #229 and #232 respectively.

**3. The cursor-cli fixture.** Landed as `packages/core/src/__tests__/fixtures/cursor-agent-2026.08.11-e8db854-composer.txt`, beside the claude-code pair, and `CURSOR_IDLE_COMPOSER` is read from it instead of from a literal. Both notes are reworded.

**The bytes are yours, and the file says so.** cursor-agent is installed here at that same build and still stops at sign-in — I re-checked while addressing this and it goes straight to `Signing in` with a login URL, so I could not capture it myself. Rather than invent a screen, the fixture holds exactly the line you quoted, `→ Plan, search, build anything`, and the provenance in the module comment, the test file and the PR body all say it came off your signed-in install via this review — the same route and the same limits as the #272 trust fixture (text, not bytes, so no escape sequences; the screen-clear in the test is scaffolding and is marked as such).

Two things worth flagging from doing it:

- **The literal was wrong about a glyph.** It had guessed `> Plan, search, build anything`; the real screen has the arrow. The marker is wording-independent enough that both match, so nothing behavioural rode on it — but that is exactly the kind of drift a fixture stops, and it is now the screen rather than my reconstruction of it.
- **The fixture is load-bearing, not decoration.** I mutation-tested it the way you did the rows: rewording it so the marker misses → **3 failures**; emptying it → the same **3 failures**; restoring → byte-identical file, **84/84**. So the table's cursor row is now asserted against a screen, and the assertion fails if the screen changes.

### Non-blocking, all three done

- **The 15 635 ms footer.** You are right that it was one observation dressed as a property, and your same-frame capture is the counter-example that makes the point better than the original wording did. The comment and the ADR now say it was *observed at 15 635 ms in one capture*, note your capture where it arrived in the same frame as the composer, and rest the rejection on the variance: a marker that is *sometimes* ten seconds late delays exactly the run it is late on. The conclusion is unchanged; the record is now a record.
- **The codex follow-up is filed: #277.** Agreed that fixing it is out of this lane and filing it is not. It carries what would settle it — capture codex's composer off a live PTY, time the quiet gap against the frame the composer actually arrives in, then either add the row or record a measured "no row needed", which closes it just as well. It is linked from the table note and the ADR amendment, so "owed a row" has an address now instead of three comments pointing at each other. `codex` stays described as unverified in the code, the ADR and the test pinning it to the untouched path.
- **Both stray double blank lines** removed (after the table, and before `describe("promptEchoed")`).

### Checks

`@actana/core` 1224/1224, this file 84/84, `tsc --noEmit` clean, eslint clean on both touched files. On the new head: **18 pass, 0 fail, 5 skipping** — the same five publish/train jobs, by name, that were skipping on `396c834`, so nothing about this push changed what runs.

Still a draft, base still `chore/release-discipline-and-cursor-cli`, nothing merged.

mrfqcentic (approved):
## Re-review — PR #275 at `95b3778`

Re-reviewed in a fresh detached worktree at the new head, off the current base tip `81cceab`. All three blocking items are resolved, all three non-blocking notes are done, and I re-proved the mechanism from scratch rather than assuming it survived a comment round. **Approving.**

### The one that needed me: the fixture bytes are real

I asked for a fixture I could not expect the author to capture, and told them that if they could not, they could commit the screen text I quoted provided they said so. They said so — in the PR body, in the module comment and in the test file — and the text is now on record as coming off my install via this review rather than as a capture taken on their machine. That is the right disclosure.

So I checked the bytes against the real thing. `cursor-agent 2026.08.11-e8db854`, signed in, driven over a pty with `TIOCSWINSZ` set to 40×100, six cold boots — one through the Workspace Trust dialog answered with `a`, five in an already-trusted workspace — each rendered through a VT100 emulator and the composer line extracted:

```
committed fixture   →·Plan,·search,·build·anything\n     33 bytes
                    342 206 222 040 P l a n , ... g \n
real screen, 6/6    ··→·Plan,·search,·build·anything\n   35 bytes
```

The **only** difference is the two-column left indent the TUI draws it at. Strip that and it is byte-identical, six times out of six — same U+2192 arrow, same spacing, same trailing newline:

```
fixture   sha256: 3091cb0bd71240507a7e55b5bf8746def9ddc807af3b9d5d3630d94cdad91f00
captured  sha256: 3091cb0bd71240507a7e55b5bf8746def9ddc807af3b9d5d3630d94cdad91f00
```

**No bytes were invented.** The committed file is exactly what cursor-agent 2026.08.11-e8db854 prints. The note about the glyph is also correct and worth having: the literal it replaced had guessed `> Plan, search, build anything` and the real screen has the arrow — nothing behavioural rode on it, which is precisely why a fixture is what catches it.

### The three blocking items

**1. `HarnessReadiness` doc comment — `packages/core/src/harness-prompt-delivery.ts:456-462`.** ✅ Resolved. The "keeps `claude-code`, `codex` and `cursor-cli` byte-for-byte unchanged" sentence is gone. It now states the history correctly — #229 added opencode on the reading that the other three were unaffected, #232 disproved it for two of them — and names `codex` as the only harness left on the default, "unverified rather than cleared". It matches the table forty lines below.

**2. `echoConfirmed` — `harness-prompt-delivery.ts:929-933`.** ✅ Resolved. "a no-op for the three families that were never affected" is gone; it now says the flag is absent for `codex` alone, and that it is absent "because nobody has read its screen, not because it was shown to be unaffected", attributing the other three to #229 and #232. That is the claim this PR actually supports.

**3. The cursor-cli provenance notes and the fixture — `harness-prompt-delivery.ts:517-530` and `__tests__/harness-prompt-delivery.test.ts:245-265`.** ✅ Resolved, all three parts. The fixture is at the requested path; `CURSOR_IDLE_COMPOSER` (`test.ts:273-277`) is `readFileSync`'d from it instead of from the box-drawn literal; both notes are reworded from "could not be captured" to the real provenance, and the test file correctly flags that the `ESC[2J` screen-clear is the test's own scaffolding rather than part of the capture.

### Re-proving the mechanism, not assuming it

A comment-only round is exactly when a stray edit slips into a table, so I ran the whole thing again.

**Mutation check, 3 runs per variant, at the new head:**

| variant | result | runs | failure list |
|---|---|---|---|
| as submitted | **84/84 pass** | 3/3 | — |
| both rows removed | **10 fail / 74 pass** | 3/3 | byte-identical across runs |
| only `cursor-cli` removed | **5 fail** | 3/3 | byte-identical across runs |
| only `claude-code` removed | **6 fail** | 3/3 | byte-identical across runs |

Same 10 case names as last round, same 5/6 split. And the split is still genuinely independent: the union of the two single-row failure sets equals the both-rows set exactly, with precisely one shared case (`gates only the harnesses whose composer somebody has actually read`, which asserts table membership). `types nothing into the 572 ms hole the capture measured` still goes red without the rows.

**The fixture is now a real input, so I mutated it as one.** Rewording it so the marker misses, emptying it, and replacing it with a single character each produce **the same 3 failures** — `recognises cursor-agent's idle composer and not its trust screen`, `holds through a TUI that boots far slower than any pause would allow`, `delivers on a slow boot and on a fast one`. Restoring returns the file byte-identical and the suite to 84/84. The cursor row still enforces, and it now enforces against a screen.

**Nothing else moved.** `quietGapMs: 350`, `maxWaitMs: 15_000`, `submitBaseMs: 150`, `submitPerCharMs: 2`, `submitMaxMs: 3_000` — the whole `DEFAULT_PROMPT_DELIVERY_PROFILE` block is byte-identical to base, and `HARNESS_PROMPT_DELIVERY_PROFILES` is still the empty object. Stripping comments from the full diff against base, the **entire** production change is still the two `HARNESS_READINESS` rows and nothing else.

**codex** is still described as unverified in all three places: the code (`:460-462`, `:556-562`, `:929-931`), the ADR amendment (`0026:174-181`), and `test.ts:1006-1025`, which pins `codex` and `invented-tomorrow` to `{composer: [], confirmEcho: false, maxPromptWrites: 1}` and asserts the table's keys are exactly the three.

### Non-blocking items, all three done

- **The 15 635 ms footer.** ✅ Both the code (`:548-554`) and the ADR now say "observed at 15 635 ms in one capture", record my capture where it arrived in the same frame as the composer, and rest the rejection on the variance rather than on the single number. That is a record now.
- **The codex follow-up.** ✅ **#277** exists, is **open**, and is linked from the table note (`:561`) and from the ADR (`0026:179-181`). It carries the capture-and-time-the-gap steps and explicitly allows a measured "no row needed" as a close. Linked from the reply too.
- **Both stray double blank lines.** ✅ Gone — zero consecutive blank lines in either touched file.

### Train checks, re-run against current tips

- Base still `chore/release-discipline-and-cursor-cli`; merge-base equals the base tip `81cceab`, so no drift. Still a draft. ✅
- `git log --format=%B base..HEAD | awk 'length > 132'` — **empty**. Longest body line 118, subjects 90 and 86. Empty across `main..HEAD` too. ✅
- **No package manifest, lockfile or version field touched** — every `package.json` `version` is identical to base; the 6 changed files are the ADR, three fixtures, the test and the module. ✅
- **#276 is still the only other open PR on this base**, now at `b319497` (it moved since last round). Re-checked file-by-file against both current tips: **zero overlap**. ✅
- Full `@actana/core` suite **1224/1224**, this file **84/84**, `pnpm -r typecheck` clean, eslint clean on both touched files. CI on `95b3778`: **18 pass, 0 fail, 5 skipping** — the same publish/train jobs by name. ✅

### One nit, not worth a round

`harness-prompt-delivery.ts:514-515` still opens with "`Plan, search, build` is cursor-agent's idle composer placeholder, **quoted in the issue's own account of the failed run**" — true of where the marker text originally came from, but the paragraph immediately below now supersedes it with the real screen. Fold it in whenever this file is next touched; it reads correctly as written.

---

My worktree is back to `95b3778` with an empty `git status` and all three mutated files verified byte-identical to their committed blobs.

The mechanism, the evidence and the scope discipline were right last round and are unchanged; the three documentation items are fixed, the fixture is genuine, and the follow-up has an address. **Approved.**

PR #276, opened 2026-08-19
feat: ship one generic orchestration skill inside the CLI, on a proposed ADR that relitigates 0006

### LINKED ISSUE #29: Session orchestrator: decide whether to build it
Promoted from `docs/session-orchestrator-brief.md` + `docs/session-orchestrator-brief-team.md` before both are deleted ([#15](https://github.com/actana/control/issues/15) §3).

Two registers of one unshipped idea. The `-team` text is the readable one and should be the substance of any spec that follows.

Filed because **an unbuilt feature brief sitting in `docs/` of a repo about to go public reads as a roadmap commitment.** An issue does not.

**Done when:** this is either specced as a real effort or closed as not-doing. It should not sit open indefinitely — that reproduces the problem the files had.

qcentic-adm:
Refs #265. **Draft, and deliberately so — the first deliverable here is a decision that is not mine to close.**

## The ADR is marked PROPOSED, not accepted

[`docs/adr/0031-the-product-ships-one-skill.md`](https://github.com/actana/control/blob/feat/orchestration-skill-in-cli/docs/adr/0031-the-product-ships-one-skill.md) amends [ADR 0006](docs/adr/0006-no-bundled-skills.md), which forbids exactly what #265 proposes, by name. 0006's last consequence — *"any future capability that would require Panel-installed skills must first justify itself against this ADR"* — is the instruction that produced this record, and `docs/specs/05-remove-bundled-skills.md:210` says the same thing more bluntly.

**The status line at the top of 0031 says PROPOSED and this PR body says it too: the repository owner ratifies or rejects it at the beta gate.** The code in this PR is written against the record so the decision can be judged against something real rather than against a description of it. A reviewer who thinks the result-bias argument fails is disagreeing with a proposal, not routing around a decision — and 0031 states the form that rejection takes.

### What 0031 decides

**Collision risk (D1)** — answered the way `harness-hooks.ts:50` already answers it, adapted to a file with no JSON envelope to carry a flag. `x-actana-managed: true` in the frontmatter is what authorises a write, and *nothing else*. The product-named directory `actana-sessions/` is the **address**, not the authorisation: conflating the two is how you delete a file an operator wrote at a path you happened to pick. The marker is matched as a substring of the file's first bytes rather than through a YAML parser, so a vendor that starts rejecting unknown frontmatter keys costs the operator a warning, not the installer its ability to recognise its own writes.

**Result bias (D2)** — the hard one, and it is not answered by the same move. The argument is made explicitly and at length: 0006 was written when the installed skills were *product features* (diagram, ship, memory) that changed what a Harness did **about the operator's own work**, whereas this one teaches a Harness to drive `actana` — the CLI the operator installed on purpose — and has no runtime half at all: no env var, no endpoint, no network call, no gate to be criticised for not gating. **The record also says plainly that this is exactly the argument the diagram skill made and lost**, says why the distinction is claimed to hold this time (the diagram skill's "only describes" was false — it exported `MC_API_*` into every Session), and states the form in which a reviewer should reject it if the distinction does not hold. Two costs are accepted openly rather than mitigated: a Session that never touches a Core can still read the skill's description, and a machine with `@actana/cli` on it is strictly not identical to one without.

**Overwrite policy (D5)** — stated as a decision, not left as a side effect of idempotence. **A copy carrying the marker is replaced, edits included.** A file at the same path *without* the marker is never written and never deleted, and is reported as `skipped`. The escape hatch is deleting the marker line, which is precisely the contract `harness-hooks-opencode.ts` already ships to operators: *"Delete the marker comment on the first line to make this file yours and stop that."*

**Precedent (in its own section)** — stated precisely so it is not overclaimed. "The product never writes harness config" is already false: `harness-hooks.ts:194/:214/:230` and `harness-hooks-opencode.ts:51`. But every one of those writes is **workspace-scoped, spawn-time, tagged and reversible**, and 0006's sentence names *the Panel* and *skill* directories. Global, mandatory, permanent and in the operator's home is a different act. Only "tagged" transfers without argument; the other three are decided rather than inherited.

## The four corrections to the brief, honoured

**1. The global directories are not in this repository.** Established from each vendor's current documentation (read 2026-08-19) and written down **here**, as `skillTarget` on `HARNESS_CLI_CONFIG` beside `homePathSuffixes` — a **descriptor with a `kind`**, not a path string, so a harness whose extension point turns out to be a plugin or a manifest entry gets a new kind rather than a redesign. Each row carries the URL it was read off and the date.

| Harness | Presence marker | Skill target | Source |
|---|---|---|---|
| `claude-code` | `~/.claude` | `~/.claude/skills/actana-sessions/SKILL.md` | code.claude.com/docs/en/skills — "Where skills live" |
| `codex` | `~/.codex` | `~/.agents/skills/actana-sessions/SKILL.md` | developers.openai.com/codex/skills — USER scope is `$HOME/.agents/skills` |
| `cursor-cli` | `~/.cursor` | `~/.agents/skills/actana-sessions/SKILL.md` | cursor.com/help/customization/skills |
| `opencode` | `~/.opencode` or `~/.config/opencode` | `~/.agents/skills/actana-sessions/SKILL.md` | opencode.ai/docs/skills |

Two findings worth the reviewer's attention. **The presence marker and the target are different fields**, and they have to be: the installer starts no process so it cannot ask a vendor's CLI whether it is there, and `~/.agents/skills` existing says somebody uses *some* agent, which is not the question. And **four harnesses resolve to two directories, on purpose** — Codex's *only* user scope is `~/.agents/skills`, Cursor and OpenCode read it too, and Claude Code reads only its own. Writing each vendor's own directory would touch twice as much of the operator's home for no extra coverage, and how much of that home is touched is the value 0006 protects. Codex was the surprise: community write-ups still say `~/.codex/skills`; OpenAI's own page does not, and the row records the page.

**2. There is no install hook, because the brief forbids the only one npm offers.** Redefined as **first-run**: every `actana <noun>` runs the ensure before dispatch. It is a no-op when the copies are current, writes nothing on a machine where no Harness namespace exists, and **cannot reach the exit code or either output stream**. It is skipped for `actana harness skills` itself, so the repair verb can actually report a repair rather than always finding its own work already done. `packages/cli/package.json` still declares no `postinstall`, `preinstall` or `prepare`, and a test now asserts that.

**3. The availability event is a full map on a 60s tick.** `HarnessSkillWatcher` holds its own last-seen map and diffs against it, ignores any event id at or below the highest it has processed, and does not treat its first observation as an arrival (a Core booting with Claude Code already installed has not just gained it). Both guards are kept because they fail differently. Nothing here promises seconds: the criterion is *by the first probe that observes the Harness*, which `SIGHUP` and the ADR 0021 install path force sooner. The probe, the tick and the payload shape are untouched.

**4. "Authored once" is a mechanism, not a convention.** The CLI may not import `@actana/shared` (ADR 0025 D4) and the Core may import two named SDK modules (D2) — neither list moves. So the `SKILL.md` is authored once at `.agents/skills/actana-sessions/` and **embedded as a string constant** in both packages by `scripts/gen-skill-payload.mjs`. Embedded rather than copied into `dist/` — which is what #265 §6 recommended — because both halves ship as esbuild bundles and an asset read from disk beside a bundle is a file that is not in it, the same reason `harness-hooks-opencode.ts` keeps its plugin as a template literal. Verified: both `packages/core/dist/core-entry.cjs` and `packages/cli/dist/actana-cli.mjs` carry the payload. The installer module and the CLI's copy of the target table are duplicated deliberately, and **a test fails on the first differing byte**, naming the file — the arrangement ADR 0025 D3 permits and `registration-blob-file.ts` already lives under.

## The skill is generic, and tests say so

No project runbook, no ticket workflow, no train, no GitHub flow, no project ids, no Core names, **no default Harness and no preference order** — a test asserts the skill names none of the four harnesses at all, because selection is the caller's. It is self-contained: no repo-relative link and no assumed checkout, which the `release` skill next door does not have to meet and this one does. It carries the whole sentinel rule (unusual token, pair not marker, last complete pair wins, whitespace-stripped match — the last one caused by the rendered screen's own line wrapping) and the ordering constraint that a report is collected *before* the Session is killed, because killing destroys the ring it is read from. When and how many Sessions to provision is left to the operator at the time.

## What #29 this implies

[#29](https://github.com/actana/control/issues/29) asks whether to build a Session orchestrator. This is the smallest honest answer to it: **not an orchestrator — a skill that lets a Harness be one.** No service, no scheduler, no new protocol; N Sessions is N `session start` calls, deliberately, so one that fails to start is one failure rather than a failed batch. **I have not closed #29 and will not**: it is closed by whoever ratifies or rejects ADR 0031, in whichever direction that goes, and closing it before the ADR is decided would be closing it on my own authority.

## Boundaries respected

- **#266** (merged as #268) and **#177** (merged as #272) are built on, not reverted. `no-local-escape.test.ts` is untouched and still passes — the installer is a filesystem write and starts no process, which is what makes it compatible with #129 D9.
- **#232** owns Core-side prompt delivery for cursor-cli. `harness-prompt-delivery.ts` and `pty-manager.ts` are untouched.
- **`AGENTS.md`, `CLAUDE.md`, `.github/workflows/ci.yml` and `docs/rulesets/`** (#260, #257, #264) are untouched. `docs/ci-cd.md` needed no change and got none.
- `docs/reference/cli-parity-check.md` is deliberately **not** updated: it compares a frozen prototype against the shipped CLI, and `harness skills` has no prototype counterpart, so a row there would misrepresent what the document is.

## Verification

- `@actana/shared` 214 ✅ · `@actana/cli` 481 ✅ · `@actana/core` 1223 ✅ · `@actana/sdk` 253 ✅ · `@actana/panel` 1439 ✅ — each package run explicitly, since `pnpm -r` bails on the first failure. Typecheck green in all five. `pnpm lint` reports 0 errors and no warning on any new file.
- Two pre-existing environmental flakes, both confirmed unrelated: `scripts/__tests__/core-launcher.test.mjs` fails only because this machine exports `AC_APP_PATH=/opt/actana/app` (green with it unset), and two Panel tests time out at 5s under load and pass when run alone.
- **The installer was exercised against a temporary `HOME` only.** Confirmed end to end with the real built binary: a machine with `~/.claude` and `~/.codex` and no Core, `actana core ls` → exit 0, "No Cores registered", and both `SKILL.md` copies on disk. Then idempotence (`current`, mtime unchanged), repair of a deleted copy (`written`), and the escape hatch (marker line removed → `skipped`, file untouched, and still `skipped` on the next run). **The real `~/.claude`, `~/.codex`, `~/.cursor` and `~/.opencode` on the test machine were checked afterwards and contain no `actana-sessions`.**

---

Draft, base `chore/release-discipline-and-cursor-cli`. Nothing is merged, `main` and the beta branches are untouched, and the checklist against #265's acceptance criteria is in the first comment.


qcentic-adm:
## Acceptance criteria — issue #265

### The gate

- [x] **An ADR supersedes or amends ADR 0006 for product-shipped skills, decides the result-bias question on the record, and states what happens to an operator-edited copy.** [`docs/adr/0031-the-product-ships-one-skill.md`](https://github.com/actana/control/blob/feat/orchestration-skill-in-cli/docs/adr/0031-the-product-ships-one-skill.md). It amends rather than supersedes (0006 stands for product-feature skills), answers collision risk with an in-band frontmatter marker (D1), makes the result-bias argument explicitly — including that it is the argument the diagram skill made and lost, and the form in which it should be rejected (D2) — states the overwrite policy as a decision (D5), and states precisely what the hooks precedent does and does not carry.
- [ ] **Merged before the implementation** — **unchecked, deliberately, and it is the only one I will not do.** The ADR is marked **PROPOSED**, the PR body says so, and the repository owner ratifies it at the beta gate. Merging the record to satisfy its own gate would be closing the decision on my own authority. The code is here so the decision has something real to be judged against.
- [ ] **#29 is closed by the decision** — **unchecked and not mine to check.** The PR body states what this implies for #29 (not an orchestrator; a skill that lets a Harness be one) and it is closed by whoever ratifies or rejects ADR 0031, in whichever direction. #265 says #29 should be closed "by whatever decision this issue reaches"; that decision is not reached yet.

### Installation

- [x] **`npm i -g @actana/cli` on a machine with no Core, followed by any `actana` command, leaves the skill in the global configuration directory of every supported Harness present on that machine.** Verified with the real built binary against a temporary `HOME` containing `~/.claude` and `~/.codex` and no Core: `actana core ls` exits 0, prints "No Cores registered", and leaves `~/.claude/skills/actana-sessions/SKILL.md` and `~/.agents/skills/actana-sessions/SKILL.md`. Covered by tests in `packages/cli/src/__tests__/harness-skills.test.ts`.
- [x] **Installing a Core does the same on the Core's machine.** `core-entry.ts` calls `ensureOrchestrationSkill(os.homedir())` at boot, beside the availability store. The payload is embedded in `dist/core-entry.cjs` (verified in the built bundle), not read off disk, because the Core ships as an esbuild bundle.
- [x] **Installing a Harness after the Core results in the skill appearing without a restart, by the first probe that observes it — driven by a subscriber to `agents:availabilityChanged` that diffs the full map against its own last-seen one and is not re-triggered by a replay.** `packages/core/src/harness-skill-watcher.ts`, 11 tests. Diffs the full map; ignores any event id at or below the highest processed; does not treat its first observation as an arrival; does not advance past a payload it cannot parse. Written as "by the first probe that observes it" rather than a wall-clock, as #265 §5 asks.
- [x] **There is an explicit verb that does the same thing on demand, is idempotent, and repairs a deleted copy. Running it twice changes nothing the second time.** `actana harness skills [--json]`. Verified with the real binary: second run reports `current` with mtime unchanged; a deleted copy comes back as `written`. The implicit pre-dispatch ensure is skipped for this verb specifically, so the repair path can report a repair instead of always finding its own work done.
- [x] **`packages/cli/package.json` still has no `postinstall`, `preinstall` or `prepare`.** Unchanged, and now asserted by a test so it stays that way.
- [x] **`no-local-escape.test.ts` still passes — the installer starts no process.** Untouched and passing. The installer imports `node:fs` and `node:path` and nothing else.

### Content

- [x] **No project-, repository- or operator-specific content.** Asserted by test: the skill contains no `actana/control`, `release train`, `beta/`, `promote.yml`, `gh pr`, `CONTRIBUTING` or `ADR 00`.
- [x] **Self-contained — no repo-relative links, no assumption of a checkout.** Asserted by test (no `](../` or `](./`, no `docs/*.md` reference). This is the constraint the `release` skill does not have to meet and this one does.
- [x] **Names no default Harness and no preference order.** Asserted by test: the skill mentions none of `claude-code`, `codex`, `cursor-cli`, `opencode`. It tells the caller to read `actana harness ls` and decide.
- [x] **The sentinel rule states all four properties.** `<%ACT_REPORT%>` … `<%/ACT_REPORT%>`, with unusual token, pair-not-marker, last-complete-pair-wins and whitespace-stripped-match each written out and each asserted by test — plus the ordering constraint that the report is saved before the Session is killed, because killing destroys the replay ring it is read from.

### Exhaustiveness

- [x] **Adding a member to `HARNESSES` without a fan-out target fails the build or a test, and the failure names the missing Harness.** `skillTarget` is a required field on `HarnessCliConfig`, and `HARNESS_CLI_CONFIG` is `as const satisfies Record<Harness, HarnessCliConfig>` — so the omission is a **compile error**. `orchestration-skill-fanout.test.ts` adds the belt: it names the harness for the empty/missing cases the type cannot catch, and separately fails when the CLI's hand-declared copy of the table is missing a row.
- [x] **The fan-out target for each Harness is recorded in this repository with the source it came from.** `skillTarget.source` (vendor URL) and `skillTarget.verifiedOn` (2026-08-19) per row, asserted by test to be an `https://` URL and an ISO date.

### End to end

- [ ] **On a machine with a Core and at least one Harness, a Harness asked for "three additional sessions, each focusing on three different things" can start them, wait, collect a report out of each and stop them, with no file from this repository on the machine besides what the install placed** — **unchecked, and stated rather than skipped.** This box needs a live Core with an available Harness and a real agent driving it; the machine this was built on has a Core but reports no Harness available, so the run cannot be honestly claimed. What *is* verified: the install lands from the real built binary against a temporary `HOME` with no Core present; every CLI verb the skill instructs (`harness ls`, `session start --wait --json`, `session logs`, `session send`, `session kill`, `events tail`) exists and is covered by that package's own tests; and the skill's account of `--wait --json` carrying the settled `screen` is read directly off `session-command.ts:321-330`. **This is the check that actually matters and it belongs at the beta gate, alongside the ADR ratification.**

### Fenced off by parallel tickets — nothing was silently skipped

No acceptance criterion required a file owned by another ticket. `AGENTS.md`, `CLAUDE.md`, `.github/workflows/ci.yml` and `docs/rulesets/` (#260, #257, #264) are untouched; `docs/ci-cd.md` needed no change and got none; `harness-prompt-delivery.ts` and `pty-manager.ts` (#232) are untouched; #266 (merged as #268) and #177 (merged as #272) are built on and not reverted, and `no-local-escape.test.ts` is not weakened.

---

**PR:** #276 — https://github.com/actana/control/pull/276 (draft, base `chore/release-discipline-and-cursor-cli`)

mrfqcentic (changes requested):
Reviewed at `2c0896c`, in a detached worktree off the PR head, against base `chore/release-discipline-and-cursor-cli` and issue #265. I read ADR 0031 first, because #265 §1 and `docs/specs/05-remove-bundled-skills.md:210` make it the gate.

**Verdict: CHANGES REQUESTED.** Not for the reason the gate suggests — the ADR is correctly marked PROPOSED and the two gate criteria are correctly left for the repository owner. Four things below are the author's to fix, and one of them is in the skill itself, which is the product surface this ticket ships.

---

## The ADR is the gate, and it answers 0006 rather than restating the proposal

**`docs/adr/0031-the-product-ships-one-skill.md:3` — Status: PROPOSED, not accepted.** That is the single thing that would have made this changes-requested on its own, and it is right. The PR body and the checklist comment both say the same. Good.

**Collision risk (D1, `:81-102`) — answered, and answered the way this tree already answers it.** `x-actana-managed: true` in frontmatter, matched as a substring of the first bytes rather than through a YAML parser, and the record decides the thing that actually needed deciding: the product-named directory `actana-sessions/` is the *address*, not the authorisation (`:88-96`). That distinction is the whole difference between "replace what we wrote" and "delete a file an operator happened to put at our path", and `harness-hooks.ts:50`'s `_acManaged` is the precedent it is adapted from. The implementation matches the record — `packages/shared/src/orchestration-skill-install.ts:134-142`: a file whose first 4 KiB lack the marker is `skipped`, never written and never deleted.

**Result bias (D2, `:104-141`) — the argument is made explicitly and its cost is accepted, not assumed.** `:123` says in as many words that this is *"exactly the argument the diagram skill made and lost"*, `:124-128` says why the distinction is claimed to hold this time (the diagram skill's "only describes" was false — it exported `MC_API_*` into every Session and its env gate stopped the network call and not the injection), and `:132-135` states the form a rejection takes. `:137-141` accepts two costs openly rather than mitigating them: a Session that never touches a Core can still read the description, and a machine with `@actana/cli` on it is strictly not identical to one without. This is what #265 §1 asked for — made on the record so it can be rejected on the record — and it is not a restatement of the proposal.

**Overwrite policy (D5, `:180-202`) — stated as a decision.** *"A copy carrying the marker is replaced, edits included"*, not preserved, not merged, not backed up; escape hatch is deleting the marker line, reusing the contract `harness-hooks-opencode.ts` already ships. #265 §1 called this out specifically as "a promise to overwrite an operator's edit" that must not arrive as a side effect of idempotence. It does not.

**The hooks precedent is not overclaimed (`:54-77`).** All four writes are named (`harness-hooks.ts:194/:214/:230`, `harness-hooks-opencode.ts:51`), the four properties are separated, and only *tagged* is claimed to transfer. Global / mandatory / permanent / in the operator's home is decided (D3, D4) rather than inherited. Correct.

---

## Findings

### 1. Three references to test files that do not exist, in the mechanism D8 leans on

`packages/shared/src/orchestration-skill-install.ts:15` (and its byte-identical twin `packages/cli/src/orchestration-skill-install.ts:15`):

> `orchestration-skill-sync.test.ts` in each package reads both files and fails on the first differing byte

`scripts/gen-skill-payload.mjs:15`:

> `orchestration-skill-payload.test.ts` in each package re-reads the authored file

Neither file exists. `docs/adr/0031-the-product-ships-one-skill.md:250-251` makes the same claim — *"a test in each package fails when a copy drifts from its source"*. In fact **every** drift assertion lives in one file in one package: `packages/shared/src/__tests__/orchestration-skill-fanout.test.ts`, which reaches into the CLI by relative path (`:19`, `:88-95`, `:108-112`).

The tests are real and they do run in CI — `scripts/lib/unit-tests.mjs:57` runs the `shared` stage — so this is not a coverage hole. It is worse in a different way: D8's entire argument for duplicating a module is *"the copies are checked by CI rather than by memory"*, and the three places that tell the next maintainer where that check is all name a file they will not find. Either point them at `orchestration-skill-fanout.test.ts`, or split the drift assertions out into the file the comments promise.

### 2. The skill's multi-Session recipe reintroduces the race the same skill says `--wait --json` exists to prevent

`.agents/skills/actana-sessions/SKILL.md:63-69` gets this exactly right and for the right mechanical reason: the transcript lives in a replay ring owned by the PTY, so a Harness that exited has taken its output with it (`session-command.ts:321-330`, `:395-400`). Then `:137-145` — the *"several Sessions at once"* recipe, which is the shape #265's end-to-end criterion is written about — says start each one **without** `--wait`, poll `session ls --json` for status, and then read `actana session logs <id>` *"while it is still alive"*.

That is the race. And the skill never tells the reader how to know whether a Session is still alive, even though the Core answers exactly that question: `session ls --json` emits a `live` field per row (`packages/cli/src/session-command.ts:361`, rendered as the `LIVE` column at `:386`). A Harness following this skill has a terminal status, no `screen`, and a `session logs` that may already be answering "has no harness running" — which the skill's own troubleshooting entry at `:166-168` then tells it is unrecoverable.

Two ways out, either is fine: name the `live` field in step 5 as the thing to check before reading `logs`, or say that N concurrent Sessions are collected by running N `session start --wait --json` calls concurrently (backgrounded), which is the only path in the CLI that hands back a settled screen. As written, the recipe for the headline use case is the one path the rest of the document argues against.

### 3. Presence is tested by a different signal than the one that triggers the re-install

`HarnessSkillWatcher` fires on a Harness flipping to `available` (`packages/core/src/harness-skill-watcher.ts:79-87`), which is a PATH/probe fact. The installer then decides whether to write from `homeMarkers` — a directory fact (`packages/shared/src/orchestration-skill-install.ts:105-107`, `~/.codex`, `~/.cursor`, `~/.opencode`, `~/.claude`).

For a Harness whose own home directory is created on first *run* rather than at install time, those two facts are not simultaneous: the probe flips to `available`, the watcher calls `ensure`, the installer reports `absent` and writes nothing — and because the availability map does not change again, nothing re-tries. On a Core's machine the next attempt is the next Core boot, since the Core has no equivalent of the CLI's pre-dispatch ensure. The #265 criterion is met to the letter ("by the first probe that observes the Harness") and missed in purpose.

D4's presence rule is a deliberate and correct decision — I am not asking for `~/.claude` to be created. What is missing is that D4 and D7 do not acknowledge that they use different signals. Either re-check on a later tick when a previously-`absent` target's marker appears, or write the gap down in D7 as an accepted consequence with the recovery path named.

### 4. An uncited claim standing among cited ones

`docs/adr/0031-the-product-ships-one-skill.md:98-99`: *"Every vendor whose loader is named in D4 documents that unrecognised frontmatter keys are ignored."* Every neighbouring vendor claim in this change carries a URL and a read date — `skillTarget.source` / `verifiedOn` per row, asserted by `orchestration-skill-fanout.test.ts:46-58`. This one, which is what makes D1's in-band marker safe to ship, carries neither. The mitigation at `:99-102` (substring match, not a YAML parse) is good and holds regardless, but the sentence should either cite the four pages or be softened to what the substring match actually buys.

### 5. ADR 0006 gains no forward pointer (follow-up, not a blocker while PROPOSED)

`0031:301` says 0006 *"is amended, not superseded"*, but `docs/adr/0006-no-bundled-skills.md` is untouched and a reader arriving there sees a rule with no trace that it has been relitigated. This repository's own precedent puts the note in the amended record — `docs/adr/0025-…:17`, *"Amended 2026-08-18 by #224"*. Not a change to make now (0031 is proposed and 0006 still stands), but 0031 should say that the note lands on 0006 at ratification, so it is somebody's job rather than nobody's.

---

## Acceptance criteria, one by one (#265)

**Gate**

- **An ADR supersedes or amends 0006, decides result-bias on the record, states the overwrite policy — PARTIALLY MET.** The record does all three (D1, D2, D5). *"Merged before the implementation"* is not met, deliberately: the ADR is PROPOSED and ratification is the owner's. That is the right call and I am not asking for it to change.
- **#29 is closed by the decision — NOT MET,** correctly. #29 is `OPEN` and was not closed by a Core. `0031:316-320` and the PR body both say it is closed by whoever ratifies, in whichever direction.

**Installation**

- **`npm i -g @actana/cli` on a Core-less machine, then any `actana` command, leaves the skill in every supported Harness present — MET.** `actana-cli.ts:141-143` runs the ensure before dispatch; `orchestration-skill.ts:47-54` cannot reach the exit code or either stream. Verified locally: 481/481 CLI tests pass, including `harness-skills.test.ts:133-141` (skill present after an unrelated `core status`, and nothing printed). Note `--version` and a bare `actana` are excluded on purpose (`actana-cli.ts:130-133`) — defensible and documented, but strictly narrower than "any `actana` command".
- **Installing a Core does the same on the Core's machine — MET.** `core-entry.ts:302`. I built `packages/core` and confirmed the payload is in `dist/core-entry.cjs`, so the "embedded, not an asset beside the bundle" argument in D8 holds. Note that the Core tarball's *own* `actana` CLI (`packages/core/src/actana-cli.ts`) does not call the ensure — #265 §6 said "both need the writer" — so this is delivered by daemon boot alone. Worth a line in D6 saying so.
- **Harness installed after the Core appears without a restart, by the first probe, via a subscriber that diffs the full map and is not re-triggered by a replay — PARTIALLY MET.** The subscriber is right: `harness-skill-watcher.ts:63-88` holds its own last-seen map, ignores any id at or below the highest processed, does not treat its first observation as an arrival, and does not advance past a payload it cannot parse. Both guards, kept because they fail differently. Finding 3 is why this is not a full MET.
- **An explicit verb, idempotent, repairs a deleted copy, twice changes nothing — MET.** `actana harness skills [--json]` (`harness-command.ts:198-243`), skipped by the pre-dispatch ensure so the repair path can report a repair (`actana-cli.ts:141`). Tests cover second-run `current` with mtime unchanged, repair of a deleted copy, and the marker-removed escape hatch.
- **No `postinstall` / `preinstall` / `prepare` — MET,** and now asserted by `harness-skills.test.ts:116-129`.
- **`no-local-escape.test.ts` still passes — MET.** Untouched; the installer imports `node:fs` and `node:path` only.

**Content**

- **No project-, repository- or operator-specific content — MET.** I read `SKILL.md` end to end. No ticket rounds, no train, no GitHub flow, no project ids, no Core names. Asserted at `orchestration-skill-fanout.test.ts:130-142`.
- **Self-contained, no repo-relative links, no assumed checkout — MET** (`:122-128`).
- **No default Harness and no preference order — MET.** The skill names none of the four (`:144-153`), and `SKILL.md:32-37` puts selection on the caller.
- **All four sentinel properties — MET.** `SKILL.md:96-112`: unusual token, pair not marker, last complete pair wins, whitespace-stripped match with the rendered-screen line-wrap as the named cause. Plus the ordering constraint at `:119-121`.

**Exhaustiveness**

- **A new `HARNESSES` member with no fan-out target fails the build or a test, by name — MET.** `skillTarget` is required on `HarnessCliConfig` and `HARNESS_CLI_CONFIG` is `as const satisfies Record<Harness, HarnessCliConfig>`, so the omission is a compile error; `orchestration-skill-fanout.test.ts:30-44` is the belt and names the Harness.
- **The fan-out target for each Harness is recorded with its source — MET.** Per-row `source` + `verifiedOn` on `HARNESS_CLI_CONFIG`, asserted to be an `https://` URL and an ISO date. The Codex row is the one that mattered and it is right to record the vendor page rather than the community answer.

**End to end**

- **Three Sessions, three subjects, started / waited / collected / stopped from the installed skill alone — NOT MET.** Unchecked by the author and stated rather than skipped. That is the correct disposition and it belongs at the beta gate — but finding 2 is about exactly this path, and I would want it fixed *before* that run rather than discovered during it.

## Train checks

1. **Commit body ≤ 132 chars — PASS.** `git log --format=%B origin/chore/release-discipline-and-cursor-cli..HEAD | awk 'length > 132'` returns nothing across the single commit.
2. **Base — PASS.** `chore/release-discipline-and-cursor-cli`.
3. **No file outside the ticket's surface — PASS.** 22 files, all within #265's declared surface. `AGENTS.md`, `CLAUDE.md`, `.github/workflows/ci.yml`, `docs/rulesets/`, `docs/ci-cd.md`, `harness-prompt-delivery.ts` and `pty-manager.ts` are untouched. The only shared-surface files are `CHANGELOG.md`, `docs/README.md` and `packages/cli/README.md`; #275 (the one other open PR on this base) touches `docs/adr/0026`, `harness-prompt-delivery.ts` and its tests and fixtures, so there is no overlap and nothing to pre-empt.
4. **CI job PASS case — N/A.** No workflow file is touched, so ADR 0023 D33 does not bite.

## Verification I ran rather than took on trust

`pnpm install` in a clean worktree at the PR head, then per package: `@actana/shared` 214 ✅ typecheck ✅ · `@actana/cli` 481 ✅ typecheck ✅ · `@actana/core` 1223 ✅ typecheck ✅. `pnpm lint`: 0 errors, 46 warnings, none in any file this PR adds or touches. Both bundles built and both carry the payload (`dist/actana-cli.mjs`, `dist/core-entry.cjs`). After the full suite, `~/.claude`, `~/.codex`, `~/.cursor` and `~/.opencode` on this machine contain no `actana-sessions` — the tests write into `mkdtemp` homes only (`orchestration-skill-install.test.ts:32`, `cli-harness.ts:178`). #29 is `OPEN` and was not closed by a Core.

## Author's own unchecked boxes

All three are honest and none is hidden: the ADR is not merged (owner's), #29 is not closed (owner's), and the live three-Session run is not claimed (needs a Core with an available Harness). I am not asking for any of the three. Finding 2 is the reason the third one worries me more than "not yet run" normally would.

---

Findings 1–4 are what this review asks for. Fix 2 first: it is in the file the whole ticket exists to ship.

qcentic-adm:
Thanks — all four are fixed in `b319497`, on top of the reviewed `2c0896c`. Finding 5 is written down as a ratification step rather than made now, as you asked. The ADR is still `PROPOSED`, #29 is still open, the base is unchanged and the PR is still a draft.

Both of the findings that had two acceptable answers are below first, since the choice is the substance.

### Finding 2 — I took **both** ways out, with the `--wait --json` one as the recipe

The race was real and it was in the headline use case, so I did not want to leave the polling path merely *safer*; I wanted the recipe itself to stop being the path the rest of the document warns about.

- **Step 3 is now the concurrent `--wait --json` shape.** *"N concurrent Sessions are N `session start --wait --json` calls running side by side — that is what makes them concurrent, not the absence of `--wait`."* It carries the shell to do it (three backgrounded starts, each redirected to its own file, then `wait`), and says the result is read out of those objects and not out of a later `logs` call. This is the only path in the CLI that hands back a settled `screen`, which is exactly the argument the mechanism section at the top already makes.
- **Step 5 keeps polling as a legitimate shape and names `live`.** Watching progress is a fair reason to poll, so the step now says a terminal status is not permission to read a transcript, names the **`live`** field on each `session ls --json` row (`packages/cli/src/session-command.ts:361`, the `LIVE` column at `:386`), and says what each value means: `live: true` — the harness is up and its replay ring still holds the transcript; `live: false` — it has exited and taken the transcript with it, *whatever the status says*. Read `logs` only while the row says `live: true`; a terminal row already at `live: false` is a result you did not collect, and the step says to report it as that rather than try to recover it.
- Two smaller consequences: the "while a Session is alive" line in the waiting section now says that `live` is what tells you that, and the troubleshooting entry for *"has no harness running"* now says the same fact was visible before the read.
- `orchestration-skill-fanout.test.ts` gained an assertion over the several-Sessions section: it must contain `--wait --json` and must name `live`. The recipe cannot quietly drift back.

### Finding 3 — I wrote the gap into the record rather than adding a re-check

You were right that D4's rule is correct and that the missing thing is the admission. I considered re-checking on a later tick and rejected it in the record rather than silently: the store appends `agents:availabilityChanged` only when the map *changes*, so there is no later tick on this subscriber's own wire to hang a retry from — it would need its own timer plus a `readdir` fan-out over four home markers forever, on every Core, to catch a state that resolves the first time the operator runs the harness. So:

- **D4** now says plainly that this is not the question `harness ls` answers and that the two can disagree — availability is a PATH fact, presence here is a directory fact, and a harness installed but never run is `available` and `absent` at once. It keeps the directory test, and says the consequence is written down in D7 rather than left to be discovered.
- **D7** names the window (binary lands → home directory exists), says why neither candidate fix is worth its cost, and **names the recovery path on both machines**: on the CLI's machine there is nothing to do, because the pre-dispatch ensure runs before every `actana <noun>` and `actana harness skills` is the explicit form; on a Core's machine there is no pre-dispatch ensure, so the automatic repair is the next Core boot (`core-entry.ts:302` runs the same ensure unconditionally), reachable deliberately with `actana core exec -- actana restart`. It also says what the operator sees meanwhile: a harness that runs Sessions normally and does not know the skill.

### Finding 1 — repointed, and the "one test, not one per package" is now said out loud

All three signposts and the ADR now name `packages/shared/src/__tests__/orchestration-skill-fanout.test.ts`, and each says *why* it is one test in one package rather than one per package: the dependency rule runs one way, so shared can read the CLI's source as text and the CLI cannot import shared at all. Repointed rather than split, because splitting would put half the assertion in a package that cannot see the other copy. Changed: both copies of `orchestration-skill-install.ts:14-24` (still byte-identical), `scripts/gen-skill-payload.mjs` — both its own header and the header it *generates* into the two payload files, which said "the drift test beside this file" — and ADR 0031 D8, which now lists what that test actually holds.

### Finding 4 — softened, not cited

I could not cite four pages honestly: none of the vendor pages in D4's table says what its loader does with an unrecognised frontmatter key, so a citation would have been four URLs that do not support the sentence. The record now says that, and states what the substring match actually buys — *recognising our own writes* is a property of our own reader and cannot be taken away by a loader getting stricter — and explicitly does not claim the vendor's behaviour either way: a loader that rejects unknown keys costs a warning at best and a skill that vendor stops loading at worst, which is a visible failure to revisit here, not one the marker prevents.

### Finding 5 — named as ratification work

0031's consequences now say that whoever moves it to ACCEPTED adds the `Amended <date> by <PR>` line to `docs/adr/0006-no-bundled-skills.md` in 0025's form, and that 0006 is deliberately untouched until then, since a forward pointer to a proposal would say 0006 no longer stands unqualified. Left as your call to execute at ratification.

### Verification

Per package, explicitly: `@actana/shared` 215 ✅ · `@actana/cli` 481 ✅ (3 skipped) · `@actana/core` 1223 ✅ · `@actana/sdk` 253 ✅ (5 skipped) · `@actana/panel` 1439 ✅. Typecheck clean on shared, cli and core. `pnpm lint`: 0 errors, 46 warnings, the same 46 as your run and none in a file this PR touches. Payloads regenerated with `node scripts/gen-skill-payload.mjs` and both copies committed.

mrfqcentic (approved):
Re-reviewed at `b319497` in a fresh detached worktree, against base `chore/release-discipline-and-cursor-cli`. All four findings and the follow-up are fixed. **Verdict: APPROVE.**

Both findings that had two acceptable answers were answered, and I judged the choice rather than the presence of a diff. Two notes below are worth folding in, neither blocking — one of them is a claim in the reply that the code does not support.

---

## Finding 2 — RESOLVED. The recipe is now the `--wait --json` shape, and I followed it cold

The author took both ways out, and put the right one in the recipe slot.

**Step 3 (`SKILL.md:139-156`) is now N backgrounded `session start --wait --json` calls with a `wait`.** The framing is the load-bearing part — *"N concurrent Sessions are N `session start --wait --json` calls running side by side — that is what makes them concurrent, not the absence of `--wait`"* — because the old step 3's justification for dropping `--wait` was exactly the confusion that starting-then-collecting is what buys concurrency. It carries the shell, and `:154-156` says the report comes out of those objects and not out of a later `logs` call.

I checked this against the CLI rather than taking it: `session-command.ts:316-331` reads `session.screen()` after `wait` resolves and emits it in the same object as `status` and `exited`, and under `--json` only that document reaches stdout (`deps.out`), diagnostics go to stderr — so `> a.json &` gets a clean object per Session. This is the only path that hands back a settled screen, which is what the mechanism section at `:63-69` already argued.

**Step 5 (`:158-172`) keeps polling as a legitimate shape and names `live` correctly.** `live` is real and it is in the `--json` document: `session-gateway.ts:104` declares it on the row, `:264` sets it from `ptyId !== null`, and `sessionLs` prints the rows verbatim through `formatJson` (`session-command.ts:363-366`), with the `LIVE` column at `:386`. The step says what each value means, and — this is the part I asked for and did not expect to get so cleanly — it says what to do with the bad case: *"A row that reached a terminal status with `live: false` before you got to it is a result you did not collect — report it as that."*

**Reading it cold as a harness, no step leaves the reader holding a terminal status and no screen.** The old step 5 did exactly that. The new one names the check, names the failure, and refuses the recovery instead of implying one. `:78-80` in the waiting section now points at `live` as the thing that answers "is it alive", and the troubleshooting entry at `:193-196` no longer contradicts the recipe — it says the same fact was visible before the read. Internally consistent top to bottom.

### Note 2a — the drift guard does not pin what the reply says it pins (not blocking)

The reply says *"The recipe cannot quietly drift back."* It can. `orchestration-skill-fanout.test.ts:166-180` slices with `skill.slice(skill.indexOf("## Several Sessions at once"))` and never bounds the end, so the slice swallows `## When something is wrong` — and that section supplies both asserted substrings on its own (`SKILL.md:194-196` contains ``--wait --json`` and ``live: false``).

I ran it rather than eyeballed it. Reconstructing the pre-fix step 3 (*"Start each Session without `--wait`"*) with only the new troubleshooting sentence kept: both assertions still pass. Worse, the `live` assertion is satisfied by the **pre-fix document unchanged** — `includes("live")` matches inside *"while it is still **alive**"* at the old `:143`, which is the exact sentence finding 2 was raised about.

Two lines fix it: bound the slice at the next `\n## `, and assert on something the recipe alone has — `"--wait --json --wait-timeout 900"` from the shell block, or `"live: true"`. As written the guard is decorative, in a change whose D8 argument is that copies are checked by CI rather than by memory.

### Note 2b — the timeout branch (nit)

Step 3's shell passes `--wait-timeout 900` and `:154` says each file holds `taskId`, `status`, `exited` and `screen`. On a timeout that is not what lands: `session-command.ts:308-312` emits `{...startedFields, waited: true, error}` — no `status`, no `screen` — and exits non-zero. The waiting section at `:73-76` gives the reader the route out (the Session is still on the Core, still listable, still killable), so this is recoverable, but the flat claim in step 3 does not hold for the one branch the flag exists to produce.

---

## Finding 1 — RESOLVED

All three signposts now name a file that exists, and each says why it is one test rather than one per package.

- `packages/shared/src/orchestration-skill-install.ts:15-23` and its still-byte-identical twin `packages/cli/src/orchestration-skill-install.ts:15-23` (`cmp` clean).
- `scripts/gen-skill-payload.mjs:14-19`, and the header it *generates* — `orchestration-skill-payload.ts:1-7` in both packages, which previously said "the drift test beside this file".
- `docs/adr/0031-the-product-ships-one-skill.md:301-309`, which now lists what the test actually holds: installer copies byte-identical, target tables row for row, both payloads equal to the authored `SKILL.md`.

`grep` for the two invented names across the tree returns nothing. Repointing rather than splitting is the right call and the reason given is the real one — the dependency rule runs one way, so a `packages/cli` copy of the assertion could not read the shared side. The file runs in CI: `scripts/lib/unit-tests.mjs:54-61`, `shared` stage.

## Finding 3 — RESOLVED, and the recovery path is named *and* works

The documentation answer, which is the one I said was acceptable, and it is not the version that names a limitation and stops.

`0031:168-177` (D4) says plainly that availability is a PATH fact and presence is a directory fact, that a harness installed but never run is `available` and `absent` at once, and that the consequence lives in D7. `0031:251-262` (D7) names the window — binary lands to home directory exists — and rejects both candidate fixes on the record rather than silently, including the correct mechanical reason a re-check has nothing to hang off: the store appends only when the map changes.

`0031:273-284` names the recovery on both machines, and I checked both:

- CLI's machine: `actana-cli.ts:141-143` runs the ensure in front of every `actana <noun>`, so the next command after the harness's first run repairs it. The one exclusion is `harness skills` itself, which does the same work with a report — so the claim holds either way.
- Core's machine: `core-entry.ts:302` calls `ensureOrchestrationSkill(os.homedir())` unconditionally at boot, before the watcher is even constructed. `actana core exec` exists (`core-command.ts:124`) and the Core tarball's CLI has `restart` (`packages/core/src/actana-cli.ts:159`), so `actana core exec -- actana restart` is a real route and not a plausible-sounding one.

It also says what the operator sees meanwhile. That is the whole of what I asked for.

## Finding 4 — RESOLVED (softened, not cited)

`0031:98-110` drops the uncited claim entirely and says the opposite out loud: no cited page answers the question, this record does not claim they ignore it, and what the substring match buys is that *recognising our own writes* is a property of our own reader. It then says what it does **not** buy. Honest, and narrower than what was there.

### Note 4a — the absence claim is true of the loader, but the page says more than "nothing"

Since the sentence asserts an absence, I spot-checked rather than took it: `https://code.claude.com/docs/en/skills`, the row at `harness-cli-config.ts:168-169` (`verifiedOn: 2026-08-19`).

The absence claim holds as written — for the **loader** reading `~/.claude/skills/`, the page says nothing about an unrecognised key. But the page is not silent on the subject. Under *"Using skill frontmatter outside Claude Code"* it says that for claude.ai uploads, the Skills API and `package_skill.py`, *"If you include any field the spec doesn't allow, packaging or upload fails with a hard error instead of ignoring the field"*, and prints it: `Unexpected key(s) in SKILL.md frontmatter: … Allowed properties are: allowed-tools, compatibility, description, license, metadata, name`. The same page adds that enabling a **personal** skill for Cowork and cloud sessions uploads it to claude.ai, *"so the same rules apply"* — and `~/.claude/skills/actana-sessions/SKILL.md` is a personal skill.

So an operator who syncs our skill to claude.ai hits a hard error on `x-actana-managed: true`. That is inside the class `:107-110` already accepts — a visible failure to revisit here, not one the marker prevents — so it does not change the decision, and D5's escape hatch is the same line. But it is concrete, it is on a page this change already cites, and it belongs in D1 as the named instance rather than left for whoever hits it. Worth a sentence at ratification; not worth another round.

## Finding 5 — RESOLVED, and 0006 is untouched

`0031:363-371` puts the note on the ratifier: whoever moves this to ACCEPTED adds the `Amended <date> by <PR>` line to `docs/adr/0006-no-bundled-skills.md` in 0025's form. It also says why 0006 is deliberately untouched now — a forward pointer to a proposal would say 0006 no longer stands unqualified — which is the right reason and not the convenient one.

Confirmed 0006 is untouched by this branch: `git log origin/chore/release-discipline-and-cursor-cli..HEAD -- docs/adr/0006-no-bundled-skills.md` is empty, and the file has no amendment line.

---

## The gate, re-confirmed after the push

This is the round where a status line gets flipped by accident. It did not.

- **`0031:3` — Status: PROPOSED.** Not accepted. Unchanged by `b319497`.
- **#29 — still `OPEN`,** `closedAt: null`, not closed by a Core.
- **Marker contract — holds.** `orchestration-skill-install.ts:139-147`: a file whose first `MARKER_SCAN_BYTES` lack the marker returns `skipped` with *"it is yours, and was left alone"* — never written. And never deleted: `grep` for `unlink`/`rmSync`/`rm(` across both copies of the installer returns nothing. There is no delete path at all.
- **No test writes into a real home.** `orchestration-skill-install.test.ts:32` uses `mkdtempSync` under `os.tmpdir()`. After running all three suites on this machine, `find` for `actana-sessions` across `~/.claude`, `~/.codex`, `~/.cursor`, `~/.opencode`, `~/.agents` and `~/.config/opencode` returns nothing.
- **The skill is generic only.** Swept `SKILL.md` for `actana/control`, GitHub, ticket, train, release, runbook, PR, ADR, pnpm, CHANGELOG — no hit. Asserted at `orchestration-skill-fanout.test.ts:130-142` (project/ticket/train/repo) and `:145-154` (no harness named, no preference order). The new step 3 shell block introduced no repository-specific content — `$project`, `$prompt_a`, `a.json`.

## Train checks, re-run against current tips

1. **Commit body ≤ 132 — PASS.** `git log --format=%B origin/chore/release-discipline-and-cursor-cli..HEAD | awk 'length > 132'` returns nothing across both commits.
2. **Base — PASS,** still `chore/release-discipline-and-cursor-cli`. No package manifest version moved: `git diff origin/…..HEAD -- '*/package.json' 'package.json' | grep '"version"'` is empty.
3. **Surface — PASS.** 22 files, the same 22 as the first round; `b319497` touched 8 of them and added none.
4. **#275 — still the only other open PR on this base, and zero overlap at current tips.** `comm -12` over the two branches' file lists against the merge base is empty. #275 at `95b3778` touches `docs/adr/0026`, `harness-prompt-delivery.ts` and its tests and fixtures; nothing this PR touches.

## Verification I ran rather than took on trust

`pnpm install --frozen-lockfile` in a clean worktree at `b319497`, then: `@actana/shared` **215 passed** (16 files) · `@actana/cli` **481 passed, 3 skipped** · `@actana/core` **1223 passed** (66 files). `tsc --noEmit` clean on shared and cli. Payload copies checked by hand: `cmp` says the two `orchestration-skill-payload.ts` are identical, and the two `orchestration-skill-install.ts` are identical. Fetched the Claude Code skills page for finding 4 rather than accepting the absence claim. Worktree left with a clean `git status` — I changed nothing.

## What is still not claimed, correctly

The ADR is not merged, #29 is not closed, and the live three-Session run is still unchecked. All three are the owner's and none is hidden. Finding 2 was the reason the third one worried me; the recipe it will be run from is now the one the rest of the document argues for, so it worries me a normal amount.

Approving. Note 2a is the one I would actually fix before this goes in — it is two lines and it makes the sentence in the reply true.

Issue #277, opened 2026-08-19
codex is owed a HARNESS_READINESS row: capture its composer and time the gap


qcentic-dev:
`codex` is the only harness left on the pre-229 prompt-delivery path, and it is there **unverified** — not cleared. It is owed a `HARNESS_READINESS` row as soon as somebody captures its composer.

## Why this is open

Issue #229 added the readiness table for `opencode` on the reading that the other three harnesses were unaffected. Issue #232 (PR #275) disproved that reading for two of them: `cursor-cli` lost a start prompt on 1 of 3 back-to-back Sessions, and the field evidence put `claude-code` at 8–10 losses in ~25 `session start` runs across three Cores. Both now have rows.

`codex` was never exercised in #232's sample and no codex screen has been read, so it keeps the default — settle, type, submit — which is **exactly the path `claude-code` and `cursor-cli` were on while they were losing prompts**. Nothing here says codex is broken; the point is that nobody has looked, and the failure is silent when it happens (the Session parks in `ready`, `--wait` times out, the prompt is simply gone).

This is deliberately out of #232's lane — that ticket fixes what it measured. This issue is the address for the remaining row, so "codex is owed a row" stops being three comments pointing at each other.

## What it needs

1. Capture codex's boot and its idle composer off a live PTY, the way the other three were captured. Commit them under `packages/core/src/__tests__/fixtures/` named for the exact build.
2. Time the gap the way #232 did: when does the quiet gap (`quietGapMs`, 350 ms) expire relative to the frame the composer actually arrives in? That measurement is what decides whether a row is needed, and it belongs in the record either way.
3. If there is a hole, add the `codex` row to `HARNESS_READINESS` in `packages/core/src/harness-prompt-delivery.ts` — composer marker, `confirmEcho`, `maxPromptWrites` — with the marker asserted against the fixture, not against a literal.
4. Time any candidate marker against the composer it stands for before gating on it. #232 rejected claude-code's `⏵⏵ … (shift+tab to cycle)` footer for this reason: it was observed at 15 635 ms in one capture, ten seconds behind that run's composer and past the 15 s backstop, while on another capture it arrived in the same frame. A marker that is *sometimes* late is unusable.
5. If the capture shows codex has no hole, say so in the code and in ADR 0026 with the measurement behind it — a verified "no row needed" is a fine outcome and closes this just as well.

## Current state (correct, keep until measured)

`codex` is described as unverified in three places, and all three should be updated together when this lands:

[...]

Issue #274, opened 2026-08-19
Panel: 2-4 test files fail in a full local run that pass in isolation (#257 §3, carved out of #270)


qcentic-dev:
Carved out of **#257 §3**, which #270 correctly declares **not delivered**. #270 lands §1
(every package reports) and §4 (a disk-caused red run says disk) only. This ticket owns the
Panel suite's isolation problem and the `packages/panel/vitest.config.ts` change that was
dropped from #270 in review.

## The problem, as #257 §3 states it

`cd packages/panel && npx vitest run` fails 2–4 files in a full run that pass in isolation:

```
 Test Files  4 failed | 138 passed (142)
      Tests  9 failed | 1432 passed (1441)
```

| File | In a full run | Alone |
|---|---|---|
| `src/components/views/__tests__/cores-settings-rename.test.tsx` | 2 failed | **5 passed** |
| `src/components/views/__tests__/project-dialog-path-immutable.test.tsx` | 2 failed | **4 passed** |
| `src/components/views/__tests__/project-picker-switch.test.tsx` | 4 failed | **6 passed** |
| `src/server/panel-link/__tests__/live-read-path.test.ts` | 1 failed | **13 passed** |

Most are `Error: Test timed out in 5000ms` — vitest's *default* per-test budget, with 142
files fanning out across every core against jsdom renders and in-process Core boots.

## What #270 established, and what it did not

**Established, and worth keeping:**

- `test.poolOptions` was **removed in Vitest 4**; this repo pins 4.1.6, which accepts it,
  logs a ` DEPRECATED ` banner, and then runs completely unconstrained. Any worker cap
  written that way is dead config. Measured on a clean 4.1.6 project, six files sleeping 1.5s:

  | config | Duration | parallelism |
  |---|---|---|
  | no limit | 1.84s | 6 at once |
  | `poolOptions: { threads: { maxThreads: 1, minThreads: 1 } }` | 1.71s | 6 at once — **ignored** |
  | `maxWorkers: 1, minWorkers: 1` | 10.03s | serialised |

- A **working** cap does engage and does reduce contention. Real Panel suite, ten-core
  machine under load, unconstrained arm forced with `VITEST_MAX_WORKERS=10`:

  | | runs | red | spread | in-test time | wall |
  |---|---|---|---|---|---|
  | unconstrained | 4 | **4** | 3–4 files, 3–9 tests | 132.7s | 49.9s |
[...]

Issue #273, opened 2026-08-19
core: `readDialogOptions` cannot read a letter-keyed menu, so cursor-agent's trust dialog is recognised but never answered


qcentic-dev:
`readDialogOptions` reads numbered menus only, so cursor-agent's letter-keyed trust dialog is recognised but never answered.

## What is true today

PR #272 gave `folder-trust` a `cursor-cli` row in `BLOCKING_DIALOGS`
(`packages/core/src/harness-prompt-delivery.ts`). That row does exactly one
thing, and the PR is now worded to claim exactly that one thing: it makes the
dialog **recognised**, so a delivery that meets it abandons and the Session is
reported `needs-input` rather than the prompt being typed into the trust dialog
and reported `delivered`.

It does not **answer** it. cursor-agent 2026.08.11-e8db854's trust screen,
captured from a live PTY in a fresh untrusted directory and committed at
`packages/core/src/__tests__/fixtures/cursor-agent-2026.08.11-folder-trust.txt`:

```
│  ⚠  Workspace Trust Required                                  │
│  Cursor Agent can execute code and access files in this dir…  │
│  Do you trust the contents of this directory?                 │
│  ▶ [a] Trust this workspace                                   │
│    [q] Quit                                                   │
│  Use arrow keys to navigate, Enter to select, or press the …  │
```

The menu is letter-keyed. `OPTION_LINE` requires a digit followed by `.` or `)`:

```ts
"^([\\s>❯➤→*·\\-" + HIGHLIGHT_MARK + "]*)(\\d{1,2})[.)]\\s+(\\S.*?)\\s*$"
```

so `readDialogOptions` returns `[]`, `chooseDialogOption` returns `null`, and
`answer` is always null on this screen. Asserted in
`packages/core/src/__tests__/harness-prompt-delivery.test.ts` against the
fixture, both at the reader and end to end through `HarnessPromptDelivery`.

## What to do

Teach `readDialogOptions` letter keys. `[a] Trust this workspace` / `[q] Quit`
is the same structure as `1. Yes` / `2. No`, one alphabet over: a leading run
that may carry the pointer glyph, a key, a delimiter, a label. The safety
argument does not change — the key still comes off the screen and is never
hard-coded — and the guard in `chooseDialogOption` (exactly one affirmative, at
least one refusing, or no answer at all) carries over unchanged, because it
reasons about labels rather than about keys.

Sketch, not a prescription:

- widen the option key from `number` to a key token, so `[a]`, `a)` and `1.`
[...]

PR #278, opened 2026-08-19
chore: release discipline, cursor-cli drivability, and one skill in the CLI

### LINKED ISSUE #29: Session orchestrator: decide whether to build it

qcentic-dev:
Release train for 0.3.3. Seven tickets, each already merged into this branch as its own reviewed pull request. This
pull request is the batch: it exists so the seven can be judged against each other, and so promotion into `main` has
one approved, green gate to consume.

Squash is the only enabled merge method, so the title of this pull request becomes the single commit for all seven in
`main`'s history and in the CHANGELOG.

## The train

| Ticket | Merged as | What it did |
| --- | --- | --- |
| #260 | #267 | States both load-bearing release-train rules in `AGENTS.md` itself rather than behind a link, adds `CLAUDE.md` as an `@AGENTS.md` import, documents re-cutting a train by hand in `docs/ci-cd.md`, and corrects `CONTRIBUTING.md`'s body-line cap from 100 to the 132 that `commitlint.config.mjs` actually enforces. Documentation only. |
| #266 | #268 | Collapses the Panel's three terminal controls into one `New Terminal` per location that opens a VM Shell Session on the Core the route is on, and deletes the project-root terminal path outright — warm pool, repository, controller, service, table, routes and client calls. Adds `actana core exec [--cwd] [--json] -- <cmd>`: non-interactive, no PTY, stdout and stderr buffered separately and escape-free, the command's real exit code or `128 + n` for a signal. |
| #264 | #271 | Makes GitHub's merge button on a promotion gate a red required check: a `Promotion gate` job fails any open pull request whose base is `main` and whose head matches `beta/*`, refusing three ways (log, `::error`, step summary) and naming the `gh workflow run promote.yml` invocation to use instead. The decision itself lives in `scripts/promotion-gate.sh` so both branches are executable by test. |
| #177 | #272 | Makes cursor-cli drivable. `validateHarnessArgv` now rejects an auto-mode option with no flag (`auto-mode-flag-missing`) instead of silently launching an interactive `cursor-agent` while the caller believes the Session is unattended; the flags get one home derived from `HARNESS_CLI_CONFIG`; cursor-agent's folder-trust dialog is recognised from a committed byte fixture so delivery takes the abandon path and surfaces `needs-input` rather than typing the prompt into the dialog and reporting `delivered`; and the turn-start signal the Core already sent stops being dropped in the SDK's `unwrapResponse`. |
| #257 | #270 | Replaces the `pnpm native:node && vitest run && pnpm -r test` chain with `scripts/run-unit-tests.mjs`, so every stage runs to completion and all six report instead of one flake hiding every other package. Adds disk handling: a per-run `TMPDIR` removed at the end, free-space and inode headroom read before any suite, and a run below the floor aborting with `DISK` instead of a wall of unrelated failures. Closes the `Unit Tests` half of #257 only. |
| #232 | #275 | Adds cursor-cli and claude-code rows to `HARNESS_READINESS` so start-prompt delivery waits on the composers' own placeholders — with `confirmEcho`, so a write that left no echo is retyped rather than submitted into nothing — instead of the generic quiet-output settle that was losing roughly one cursor-cli start in three. The `shift+tab to cycle` footer was measured and rejected as a marker, and a test asserts it is not used. No timing constant changed. |
| #265 | #276 | Ships one generic orchestration skill, authored once at `.agents/skills/actana-sessions/` and embedded as a string constant in both the CLI and the Core, with a test that fails on the first differing byte. Install is first-run rather than an npm lifecycle hook and cannot reach the exit code or either output stream. `x-actana-managed: true` in the frontmatter is the sole write authorisation — a file at the same path without it is never written or deleted, only reported as skipped. |

## REVIEWER NOTES

Every ticket above has already had its own review. What no one has reviewed is what the seven do to each other. These
four are where they touch.

**1 — `#266` and `#265` both reach into the Core's entry path and the CLI's published surface.**

The overlap that was expected here did not happen, and that is worth stating rather than leaving a reviewer to
re-derive it: `packages/cli/src/core-resolution.ts` and `packages/cli/src/__tests__/no-local-escape.test.ts` are
untouched by all seven commits, and `packages/sdk/src/core-link-frames.ts` was touched only by #268 (new exec frames),
not by #276. What #268 and #276 actually share is:

- `packages/core/src/core-entry.ts` — #268 adds the exec route to the Core; #276 starts `HarnessSkillWatcher` on the
  60s tick. Two unrelated additions to one boot path; check the ordering and that a watcher failure cannot take out
  the exec route.
- `packages/cli/README.md` and `CHANGELOG.md` — both edited by both. Textual, but confirm neither entry got lost.

The `core-link-frames.ts` change from #268 still deserves a look on its own: it is new wire protocol landing in the
same train as the skill payload that both packages now embed.

**2 — three genuine source-level collisions.**

- `packages/shared/src/pty-spawn-policy.ts` — #268 (7 lines, for the exec path) and #272 (113 lines, the auto-mode
  flag rule). Confirm #268's addition still satisfies the stricter argv validation #272 introduced.
- `packages/shared/src/harness-cli-config.ts` — #272 makes it the single source of `HARNESS_AUTO_MODE_FLAGS`; #276
  hangs the typed skill-target descriptors off the same object. One config now carries two unrelated contracts;
  check `assertAutoModeFlagsSync` still bites after #276's additions.
- The Core prompt-delivery path — `packages/core/src/harness-prompt-delivery.ts` and its test are touched by both
  #272 and #275, and both amend ADR 0026. #272 adds the abandon path for a recognised dialog; #275 adds composer
  readiness. These interact directly: a trust dialog recognised by #272 must not be mistaken for a composer marker
  by #275, and an abandoned delivery must still resolve. Also note #268 and #272 both touch
  `packages/cli/src/__tests__/cli-harness.ts` and `packages/sdk/src/core-client.ts`.

**3 — `.github/workflows/ci.yml` was touched by both #257 and #264.**

#270 changed the `Unit Tests` job body; #271 added a new `promotion-gate` job immediately before `typecheck`. Two
questions for the batch:

- Does the readable-red-run change still work with the promotion guard present — does every package still report,
  and does the disk verdict still surface, on a run where the guard is also evaluated?
- Is the guard's pass case a **successful exit rather than a skipped job**? The job deliberately carries no `if:`
  (ADR 0023 D33) precisely so it cannot resolve as skipped on a non-pull-request event, and `promotion-gate.sh` is
  supposed to exit 0 on the pass branch. Confirm that on this pull request the check is green, not grey.

**4 — `docs/ci-cd.md`: coherent, or edits stacked?**

Touched by #260 (#267) and #264 (#271) — two edits, not three; #265 (#276) added `docs/adr/0031` and a line in
`docs/README.md` but did not touch `docs/ci-cd.md`. #267 added the "Re-cutting a train, by hand" section, #271 added
the promotion-gate context. Read them together: they describe the same lifecycle from two directions, and both were
written without sight of the other.

## NEEDS THE OWNER'S DECISION

#276 ships `docs/adr/0031-the-product-ships-one-skill.md` — a **new ADR that relitigates ADR 0006**, which forbids
bundled skills by name. It is deliberately marked **PROPOSED and not accepted**.

This is not a Core's call and it is not a reviewer's call. **The repository owner ratifies or rejects ADR 0031 at the
beta gate.** The implementation in #276 is written against the record so the decision can be judged against something
real rather than against a proposal, but the code is gated on the record, not the other way around.

Issue **#29 ("Session orchestrator: decide whether to build it") is deliberately still OPEN.** It is closed by
whichever way the ADR 0031 decision goes — accept and it is answered by the shipped skill, reject and it is answered
by the refusal. #276 states this explicitly: it refs #265 and does not close #29.

## FOLLOW-UPS FILED

Opened during this train rather than fixed in it. All four are open.

- **#269** — `.husky/pre-push` does not carry `.github/workflows/ci.yml`'s `beta/*` exemption, so the hook refuses a
  train branch that CI accepts.
- **#274** — Panel test isolation: 2–4 test files fail in a full local run that pass in isolation. Carved out of #270
  (§3 of #257), including the `packages/panel/vitest.config.ts` worker cap, because the clamp is a floor as well as a
  ceiling and on a 4-vCPU runner would have raised the Panel suite from 3 workers to 4 — the direction associated
  with more red files. The supporting measurements went to the issue rather than being discarded.
- **#277** — codex is owed a `HARNESS_READINESS` row: capture its composer and time the gap. #275 left codex on the
  untouched generic path — unverified rather than cleared.
- **#273** — `readDialogOptions` cannot read a letter-keyed menu, so cursor-agent's trust dialog is recognised by
  #272 but never answered. #272 buys the abandon path; answering the dialog is this issue.

Also noticed and not fixed in #270: `pnpm typecheck` is still `pnpm -r typecheck`, the same cross-package bail one
job over. #208 stays open by design.


qcentic-adm (changes requested):
## Batch review — 0.3.3 train gate, seven tickets

Reviewed as a batch at `71f91da`, in a read-only worktree. The per-ticket reviews are taken as done; what follows is what the seven do to **each other**, plus the six release-critical checks.

**One blocker, and it is a batch-only defect** — neither #267's reviewer nor #271's could have seen it, because it is created by the two landing together.

---

### BLOCKER — `AGENTS.md` tells an agent to wait for a green that #264 made unreachable

`AGENTS.md:21`, written by **#260** (`79a2718`, first commit in the train):

> **Promotion consumes an existing, green, approved pull request from the train into `main`** — open it, **let it go green**, get it approved, then dispatch `promote.yml`.

**#264** (`9ccedfb`, third commit) then makes `Promotion gate` a *required* check that is **red on every `beta/* → main` pull request, permanently and by construction**. A gate pull request's aggregate state is never green again. `promotion-gate.sh` is explicit: *"This check is red on purpose, it stays red for the life of this pull request, and red is the healthy state for a promotion gate."*

So at HEAD the repository's entry-point file — 17 lines, imported wholesale into `CLAUDE.md` via `@AGENTS.md`, the first thing every agent reads — instructs an agent on the release path to wait for a state that cannot occur, and describes the gate as *"green, approved"*.

**#264 knew this phrasing had become false and fixed it everywhere except here.** It edited the identical wording in the two other places it lives:

- `promote.yml:199` — *"Open it and let it go green"* → *"Open it, let its checks settle and get it approved — `Promotion gate` is red on it by design and stays red"*
- `docs/ci-cd.md:1081` — *"let it go green"* → *"let its checks settle"*

`AGENTS.md` is the third instance and the one that matters most. `grep -n "Promotion gate" AGENTS.md CLAUDE.md CONTRIBUTING.md` returns nothing: the guard #264 built is invisible from the file #260 exists to make authoritative.

This is the failure class both tickets were filed against. #260 failure 3's whole point is that *"the documentation was correct, current, and had just been read"*; the documentation is now **incorrect** and is still the first thing read. #260's acceptance criterion 1 — *"an agent that reads only `AGENTS.md`, and follows no link, can state … promotion consumes an existing, green, approved pull request"* — is satisfied by a sentence that four commits later stopped being true.

**Fix:** one sentence in `AGENTS.md`, matching what #264 already wrote in the other two places — checks settle rather than go green, `Promotion gate` is red by design on a gate, dispatch anyway. Nothing else in the batch needs to move.

---

### The release-critical checks — all six pass

**1. Commit body lines ≤ 132 — PASS.** Ran it, did not assume:

```
$ git log --format=%B origin/beta/0.3.3..HEAD | awk 'length > 132 {print NR": "length}'
(no output)
```

Seven commits, one per ticket. Longest body line in the range is **118**.

**2. Manifests — PASS.** All six say `0.3.3`. `git diff origin/beta/0.3.3..HEAD -- package.json packages/*/package.json` is a single hunk on the root `test` script (#257: `pnpm native:node && vitest run && pnpm -r test` → `node scripts/run-unit-tests.mjs`). No version line is touched by any of the seven. ADR 0023 D3 held.

**3. No required check's pass case is a skipped job — PASS.** `promotion-gate` carries **no job-level `if:`**, deliberately, and every not-applicable path inside `promotion-gate.sh` is an early `exit 0` with a `::notice` — non-`pull_request` event, base ≠ `main`, head not `beta/*`. Live on this pull request: **`Promotion gate` — pass, 5s, `conclusion: SUCCESS`.** Green, not grey. The three jobs that *do* resolve SKIPPED here (`Resolve train tags`, `Panel image (train)`, `Core image (train)`) are not among `docs/rulesets/main.json`'s 17 required contexts, and all 17 reported SUCCESS. D33 satisfied.

**4. `promote.yml` unchanged in how it advances `main` — PASS.** The entire diff is one line of error-message text at `:199`. Byte-identical: the ancestry re-check at `:444`, the D16 revision-label assertion at `:361`, and the advance at `:457` — `git push origin "$HEAD_SHA:refs/heads/main"`. No new inputs, no new preconditions on the gate's check state. #264 AC3 explicitly permits message text to change and forbids behaviour to; it changed only text.

**5. Nothing applies a ruleset — PASS.** The ruleset half is a written-up diff: `+1` context (`Promotion gate`) in `docs/rulesets/main.json`, plus a `docs/REPO_SETUP.md` §3 step for the owner, carrying a pre-flight, the destructive-`PUT` warning, and a *"confirm the new context reports before you require it"* gate that names the Pending-forever risk. No workflow and no script performs a `PUT`; the only `rulesets` references under `scripts/` are read-only assertions in `promotion-gate.test.mjs` and `workflows.test.mjs`. The five-row drift table is recorded rather than silently applied, and is honest that closing it would not have stopped #259.

**6. ADR 0031 — PASS on both halves.** Marked **`Status: PROPOSED. Not accepted.`** in its first line, listed in `docs/README.md` as *"**PROPOSED**, amends 0006"*, and routed to the owner at the beta gate. #29 correctly left open and ref'd rather than closed.

On whether D2 **answers** 0006's result-bias argument or merely restates it — asked directly, because this repository has rejected this argument once: **it answers it, and the answer is falsifiable rather than rhetorical.** D2 names a structural difference the diagram skill did not have, and names it mechanically: the orchestration skill *"has no runtime half at all: it exports no environment variable, it opens no socket, it names no endpoint, it has no gate to be criticised for not gating."* The diagram skill lost because *"only describes"* was untrue of it — it shipped `MC_API_*` into every Session and its env gate stopped the network call and not the injection. That is a real distinction, not the same claim with better manners.

It is also not a *defeat* of the objection, and D2 says so rather than hiding it: it concedes that *"a Session that never touches a Core can still read this skill's description, and that is a real cost paid for the mandatory-ness D3 chooses."* So the honest summary is that D2 **narrows** result-bias to a residue it argues is of a different kind, states the residue openly, and writes its own rejection condition on the record — *"if a reviewer thinks that distinction is the diagram skill's argument with better manners, this record fails and should be rejected."* That is the right shape for an ADR that relitigates a decision, and the call is the owner's.

One thing to be plain about, since it bears on the gate: the ADR is proposed but **the code is in this train**. "Gated on the record" is an intent, not a mechanism — promoting this train ships the installer. If 0031 is rejected, #276 comes out *before* promotion, not after.

---

### Batch-level findings — the cross-ticket work

**No ticket reverted or undid another.** No file added by an earlier commit is deleted by a later one (checked across all seven by name-status). The six deletions in the range are all #266's own removal of the project-root terminal path. Exactly 12 files are touched by more than one commit; every one is read below.

**The three files the reviewer notes flag as a non-collision — independently confirmed.** `packages/cli/src/core-resolution.ts` and `packages/cli/src/__tests__/no-local-escape.test.ts` are untouched by **all seven** commits; `packages/sdk/src/core-link-frames.ts` is touched by **#266 only** (new exec frames, protocol `0.15.0` → `0.16.0`), not by #265. The correction in the pull request body is accurate.

**`packages/shared/src/pty-spawn-policy.ts` (#266 → #177) — both intents survive.** Disjoint regions. #266 adds only a doc block on `ShellSessionSpawnRequest` recording that `core exec` **does not come through the spawn policy at all** — *"a plain child with pipes and no terminal, so it is not a spawn mode in this union."* #177 rewrites the argv validation below it. Both present at HEAD. So the question "does #266's addition satisfy #177's stricter validation" resolves to: it is not subject to it, by design, and the design is stated in-band rather than left to be inferred. `core exec` carries its own bounds instead, in `packages/core/src/core-exec.ts` — cwd validated by the machine that owns the disk, an 8 MiB output cap that **refuses** (`outcome: "output-too-large"`) rather than truncating, and a time bound. That is the right split; a PTY allow-list has nothing useful to say about a non-PTY pipe.

**`packages/shared/src/harness-cli-config.ts` (#177 → #265) — purely additive; the drift assertion still bites.** #265 adds `HarnessSkillTarget`, the required `skillTarget` field and `HARNESS_SKILL_TARGETS` around #177's work without touching a line of it. `autoModeFlag`, `HARNESS_AUTO_MODE_FLAGS`, `autoModeFlagForHarness` and `assertAutoModeFlagsSync` are all intact, and `assertAutoModeFlagsSync` is still exercised twice in `harness-cli-config.test.ts` (both the passing and the drifted case). The two contracts now on one object are independent: one derives from `autoModeFlag`, the other from `skillTarget`, both over `MANAGED_HARNESSES`. `skillTarget` being **required** under `as const satisfies Record<Harness, HarnessCliConfig>` makes a new `HARNESSES` member a compile error, which is #265's exhaustiveness criterion bought for free.

**Core prompt delivery (#177 → #232) — the interaction is correct by construction, and I checked it two ways.**

*By ordering.* In `onQuiet`, `matchBlockingDialog` runs **before** the composer gate and returns:

```
const dialog = matchBlockingDialog(this.screen, this.specs);
if (dialog) { this.handleDialog(dialog); return; }
if (!this.deadlinePassed && !composerOnScreen(this.screen, this.readiness)) { this.holdForComposer(); return; }
this.emit({ phase: "settled", … }); this.writePrompt();
```

A dialog #177 recognises can therefore never be read as a composer-ready screen by #232. #177 has priority over #232 on every screen where both could speak.

*By evidence.* I cross-matched all eight committed fixtures against **both** tables — the widened `folder-trust` match from #177 and all three composer markers from #229/#232. **Every fixture matches exactly one row; there are zero cross-matches.** cursor-agent's trust screen (`Workspace Trust Required` / `Do you trust the contents of this directory?`) matches the dialog and no composer; its idle screen (`→ Plan, search, build anything`) matches the cursor composer and no dialog. Same disjointness for both claude-code builds. `harness-prompt-delivery.test.ts` asserts the sharp direction directly at `:470` — `matchBlockingDialog(REAL_COMPOSER, dialogsForHarness("claude-code"))` is null.

*An abandoned delivery still resolves.* `abandon()` cancels both the idle and deadline timers and emits; #177's `pty-manager.ts` handler fires on `phase === "abandoned"` and reports `dialog-unanswered` → `needs-input`, which is a **settled** status, so an SDK `waitForIdle` returns rather than hanging. #232's composer gate cannot deadlock it either: it is guarded by `!this.deadlinePassed`, so at the backstop the prompt goes out (or the deadline finds a dialog and abandons). Both directions are bounded.

Worth noting the honesty of the two rows: #177's cursor-cli trust row is *recognised, not answered* — the menu is letter-keyed and `readDialogOptions` takes digits, so `answer` is null and every such delivery abandons (#273 filed). #232 rejected the `shift+tab to cycle` footer on a measurement, not on taste, and `:1056` asserts it is not used. Neither ticket overclaims.

**`.github/workflows/ci.yml` (#257 + #264) — both work, and the guard's pass case is green.** Two independent jobs, neither in the other's `needs:`; #264's hunk lands at `:423`, #257's at `:506`. Live proof on one run: **Unit Tests — pass, 2m10s** and **Promotion gate — pass, 5s**. Every package still reports (`run-unit-tests.mjs` runs all six stages and aggregates exit codes rather than short-circuiting), and the disk verdict still surfaces (`::error title=Unit Tests — out of disk`, the `DISK:` lines, the step summary). The coordination between #257's per-run `TMPDIR` sandbox and the job-scope leak check is handled inside #257 — the sandbox prefix is *deliberately* outside the `mc-*`/`actana-*` globs the check greps, with the reasoning written down, and the check's ephemeral-runner assumption is stated with the self-hosted workaround. Nothing about #264's job perturbs any of it.

**`docs/ci-cd.md` (#260 + #264) — coherent, and two edits, not three.** #265 added `docs/adr/0031` and a `docs/README.md` row but did not touch this file. Not stacked: #264 **edited #260's own sentence in place** at `:1081` rather than appending beside it, precisely because the permanently-red gate falsified *"let it go green."* The promotion-gate paragraph sits after the dispatch block and cross-references it; the `:117` gate rule is amended in place rather than duplicated. This is the right instinct applied twice — which is exactly why its absence from `AGENTS.md` is the blocker above and not a nit.

**`packages/core/src/core-entry.ts` (#266 + #265) — ordering is safe; a watcher failure cannot take out the exec route.** #265's `ensureOrchestrationSkill(os.homedir())` runs at `:289`, #266's `execPort` is wired at `:360` — so the boot-path install does run first. It cannot break the boot: `ensureOrchestrationSkill` wraps the whole install in `try/catch`, logs `core-skill.install-failed` and returns `[]`, with the reason written down — *"a Core that refused to start because it could not write a markdown file into a directory it does not own would be trading a documented capability for the whole product."* The watcher tee is equally contained: it wraps `appendEvent` for the availability store **only**, `observe` catches its own JSON parse, and the `ensure` it calls is the same non-throwing function. `initial` is deliberately not passed, so the first availability map is treated as boot state and not an arrival — no double write on every start.

**`docs/adr/0026` (#177 + #232) — two appended, dated amendment sections** (`## Amendment — issue 177` at `:64`, `## Amendment — issue 232` at `:125`). Append-a-clause-never-renumber, per `CONTRIBUTING.md`. ADR 0023's #264 amendment follows the same shape.

**`CHANGELOG.md` and `packages/cli/README.md` (#266 + #265) — neither entry lost.** Non-overlapping hunks in both files; all four entries present at HEAD. #265's changelog entry flags itself *"**This one is provisional**"* and points at the unaccepted ADR, which is the right thing for a reader to find.

**`packages/sdk/src/core-client.ts` and `packages/cli/src/__tests__/cli-harness.ts` (#266 + #177)** — additive in both, different regions, both intents intact.

---

### The no-local-escape test got *stronger*, not weaker

Asked directly because `actana core exec` is the verb that looks like it should have cost something here. `packages/cli/src/__tests__/no-local-escape.test.ts` is **byte-for-byte unchanged** by all seven commits — `git log origin/beta/0.3.3..HEAD -- <that file>` is empty.

Its coverage nonetheless grew, because the sweep is by directory rather than by list: `shippedSources()` recurses `packages/cli/src` excluding `__tests__`, so both #266's `core-exec.ts` and #265's four new modules are swept automatically, with no one having to remember to extend anything. Statically: `core-exec.ts` imports `./core-connection.ts`, `./cli-output.ts`, `./exit-codes.ts` and SDK **types** only; #265's installer imports `node:fs` and `node:path` and nothing else. No `child_process`, no `worker_threads`, no `process.binding`, no concatenated `require`.

And #266 landed on the right side of the argument the test encodes. `core exec` starts its process on the **Core**, over the authenticated link — it is the test header's own named exception (*"the one verb that will legitimately want a subprocess is `core shell` … it hands the operator a terminal on the Core, over the core link, rather than running anything locally"*) minus the terminal. It removes a `docker exec` reach-around rather than adding one, and #129 D9's argument survives intact.

---

### Per-ticket verdict — at the batch level

| Ticket | Landed as | Batch verdict |
|---|---|---|
| **#260** | `79a2718` | **Changes requested.** Delivers `AGENTS.md` §Release trains, `CLAUDE.md` as an `@AGENTS.md` import, the re-cut section, and the correct 132 cap (verified against `commitlint.config.mjs:118`). But its `AGENTS.md` sentence is falsified by #264 three commits later and was not updated with the other two instances. |
| **#266** | `90cf562` | **Pass.** Panel collapsed to one control per location; project-root path removed whole, `user_terminals` **dropped** and the decision recorded with a migration test. `core exec` bounded on all three axes, protocol moved `0.15.0` → `0.16.0`. Does not weaken the spawn policy — it stands outside it, and says so in-band. |
| **#264** | `9ccedfb` | **Pass, with the one omission above.** Guard is a required check with no job-level `if:`, red on a gate and green everywhere else, decision extracted to a testable script. `promote.yml`'s advance untouched. Ruleset half correctly written up, not applied. It corrected the now-false "go green" wording in two of three places. |
| **#177** | `ada63c2` | **Pass.** `auto-mode-flag-missing` closes the silent direction; the flag gets one home with a drift assertion; cursor's trust screen recognised from a committed fixture and honestly labelled *recognised, not answered* (#273 filed); `hooksReportTurnStart` reaches the SDK. Composes correctly with both #266 (before it) and #232 (after it). |
| **#257** | `81cceab` | **Pass.** All six stages run and report, exit codes aggregated, disk read before the suites, per-run `TMPDIR` swept. Coexists with #264's job with no interaction. |
| **#232** | `238b43d` | **Pass.** Two readiness rows with fixtures behind both, `confirmEcho` on each, the `shift+tab` alternative rejected on a measurement and asserted against, codex left explicitly *owed* rather than cleared (#277). Cannot mistake #177's dialogs for a composer — verified by ordering and by fixture cross-match. |
| **#265** | `71f91da` | **Pass on the code; the decision is the owner's.** ADR 0031 correctly `PROPOSED`; D2 makes a real, falsifiable argument and concedes its residue. Twins are byte-identical today and a test fails on the first differing byte; no `postinstall`/`preinstall`/`prepare`; `x-actana-managed` is the sole write authorisation and an unmarked file is reported skipped, never written or deleted; boot install cannot throw. Note the code ships if this train promotes — if 0031 is rejected, #276 must come out first. |

---

## CHANGES REQUESTED

Six of six release-critical checks pass. No ticket reverted another, all twelve collision files hold both intents, and the three interactions with real teeth — `#177` vs `#232` on the delivery path, `#257` vs `#264` in `ci.yml`, `#266` vs `#265` on the Core boot path — are each correct by construction rather than by luck.

The single blocker is one sentence. `AGENTS.md:21` is the file this train exists to make authoritative and the file every agent loads first, and it now instructs the release path to wait for a green that #264 deliberately made unreachable. #264 fixed that exact wording in `promote.yml` and in `docs/ci-cd.md`; it needs the third instance too. Fix that and this batch is ready to promote.

mrfqcentic (changes requested):
Batch review of the seven tickets as a batch — what they do to each other, plus the release-critical checklist. Reviewed from a read-only worktree at 71f91da.

## Per-ticket verdicts

| # | Ticket | Verdict |
|---|--------|---------|
| 260 | AGENTS.md release-train rules | Lands as specced, **but see batch finding 1** — 264 invalidated its key sentence |
| 266 | Panel terminal control + `actana core exec` | Sound; intents survive 265 |
| 257 | Readable red run | Solid; exit codes aggregated, no failure swallowed |
| 264 | Promotion gate guard | Mechanism engages correctly (verified by running it); promote.yml untouched |
| 177 | cursor-cli drivable | Auto-mode/readiness/trust-dialog handling all intact after 232/265 |
| 232 | cursor/claude readiness markers | Intact after 265; ADR 0026 carries both amendments coherently |
| 265 | orchestration skill + ADR 0031 | ADR 0031 is PROPOSED as required (see below) |

## Batch findings

**1. One real crack between 260 and 264 — AGENTS.md now instructs the impossible.** AGENTS.md:21 (from 260) says: "Promotion consumes an existing, **green**, approved pull request from the train into main — open it, **let it go green**, get it approved." After 264 that can never happen: `Promotion gate` is a required check that is red on a gate by design. 264 correctly updated docs/ci-cd.md ("let its checks settle") and promote.yml's message, but the file 260 built — the one an agent reads first — still tells whoever opens the gate PR to wait for a green that will not arrive, and CLAUDE.md's @AGENTS.md import propagates it. This is exactly the class of on-gate misdirection this batch exists to prevent. **Requested fix: amend AGENTS.md's sentence to "let its checks settle" and state that `Promotion gate` being red on the gate is the healthy state.** One sentence. Everything else in the batch is sound.

**2. Overlaps — both intents survive everywhere.** Verified final state, not assumed:
- `pty-spawn-policy.ts` (266→177): 266's exec-mode exclusion doc comment and 177's `withAutoModeFlags`/`HARNESS_ARG_RULES_BASE` refactor both present.
- `harness-cli-config.ts` (177→265): 177's `autoModeFlag` + `HARNESS_AUTO_MODE_FLAGS` + `assertAutoModeFlagsSync`, and 265's `HarnessSkillTarget` descriptor + `HARNESS_SKILL_TARGETS` both present; `as const satisfies Record<Harness, HarnessCliConfig>` intact.
- Core prompt delivery (177→232): 177's folder-trust abandon + `dialog-unanswered`→`needs-input`, and 232's `HARNESS_READINESS` rows for claude-code/cursor-cli both present; ADR 0026 carries both dated amendments, coherent and cross-referenced.
- 266∩265 on `packages/core/src/core-entry.ts` (`execPort` wiring + `HarnessSkillWatcher` boot) and `packages/cli/README.md` (exec section + skills section): both present. (Note: 265 did not in fact touch `core-resolution.ts`, `no-local-escape.test.ts`, or `core-link-frames.ts`; the natural 266∩265 overlaps are README + core-entry, and they hold.)
- ci.yml (264→257): the `promotion-gate` job survives 257's Unit Tests rewrite; `pnpm test` now runs `scripts/run-unit-tests.mjs`, which runs every stage and aggregates exit codes.

**3. no-local-escape is undiminished.** The test is untouched by this batch, and because it sweeps all of `packages/cli/src` recursively, 266's `core-exec.ts` and 265's skill modules are swept automatically. Re-ran the sweep logic (same helpers, same deny list) against the PR head: 35 shipped sources, 0 violations. The `actana core exec` verb starts a process on the Core over the link, not locally — the boundary holds as issue 266 itself required.

## Release-critical checks (each ran, not assumed)

1. **Body lines ≤132:** `git log --format=%B origin/beta/0.3.3..HEAD | awk 'length > 132'` → empty. PASS.
2. **Six manifests = 0.3.3:** only root `package.json` changed, and only its `test` script (`node scripts/run-unit-tests.mjs`); no `version` touched anywhere. PASS (ADR 0023 D3).
3. **No required check passes by skipping:** `promotion-gate` has no job-level `if:`. Ran `scripts/promotion-gate.sh` directly: refuse branch exits 1; push/non-main-base/non-beta-head all exit 0. D33 satisfied. PASS.
4. **promote.yml advance unchanged:** diff is one message line (allowed by #264 AC3). `git push origin "$HEAD_SHA:refs/heads/main"` at :457, ancestry re-check at :444, and the D16 digest-verification block all byte-identical. PASS.
5. **No rulesets applied:** `docs/rulesets/main.json` gains the `Promotion gate` context; `docs/REPO_SETUP.md` gains an owner-run §3 step and a live-drift table. No workflow applies anything. PASS.
6. **ADR 0031 on 0006:** Status header says **PROPOSED**, and the index reflects it. On the question I was asked to answer plainly — it does **not** merely restate 0006's result-bias argument: it offers a real counter-argument (subject matter: it documents the product's own CLI rather than the operator's work) but then **explicitly concedes the strict form of the bias survives** and defers the verdict to the gate. That is the honest form of the question, which is what a proposed amendment owes.

The single requested change is finding 1 above; with AGENTS.md amended, this gate is approvable.

qcentic-dev:
## Blocker addressed — `AGENTS.md:21` amended, nothing else touched

Both gate reviews landed on the same single blocker, and it is now fixed in `491002d`.

`AGENTS.md` was the third instance of the wording #264 falsified, and the one that propagates: 17 lines, first file every agent reads, imported wholesale into `CLAUDE.md` via `@AGENTS.md`. It now says what #264 already wrote into `promote.yml:199` and `docs/ci-cd.md:1081`.

**The new sentence, verbatim:**

> **Promotion consumes an existing, approved pull request from the train into `main`** — open it, let its checks settle, get it approved, then dispatch `promote.yml`. `Promotion gate` is red on that pull request by design and stays red: red is the healthy state there, not a problem to fix (#264).

Three things changed against the old text, matching what each reviewer asked for:

- *"let it go green"* → *"let its checks settle"*, as in `docs/ci-cd.md:1081`.
- *"green, approved"* → *"approved"* in the bolded lead — the gate is never green, so describing it as green was the same error in a second clause.
- A new clause states that `Promotion gate` being red is the **healthy** state and not something to fix, and that you dispatch `promote.yml` anyway — the substance of `promote.yml:199`, compressed to the register of the surrounding lines.

The guard is no longer invisible from the entry-point file. `grep -n "Promotion gate" AGENTS.md CLAUDE.md CONTRIBUTING.md` returned nothing before this commit; it now matches in `AGENTS.md`, and therefore in every Session that loads `CLAUDE.md`.

`AGENTS.md` is still 21 lines — the amendment extends the existing paragraph rather than adding a section, because the file's value is that it is short enough to be read.

### Scope

One file, one line: `git diff --stat 71f91da..491002d` → `AGENTS.md | 2 +-`. No change to `docs/ci-cd.md`, `promote.yml`, `promotion-gate.sh`, any ruleset payload, or any package source — both reviews verified those and said one sentence was all that was needed.

### Checks re-run rather than assumed

- **Body lines ≤ 132:** `git log --format=%B origin/beta/0.3.3..HEAD | awk 'length > 132 {print NR": "length}'` → empty. Longest body line in the range is still **118**; this commit's longest is **104**, header **64** (cap 120).
- **commitlint:** `npx commitlint --from HEAD~1 --to HEAD` → `found 0 problems, 0 warnings`.
- Pushed as a fast-forward (`71f91da..491002d`). No force push, no merge, no workflow dispatched, `main` and `beta/0.3.3` untouched.

qcentic-adm (approved):
## Re-review of the 0.3.3 train — blocker RESOLVED

Re-reviewed at `491002d` (`docs(agents): a gate's checks settle, they never go green`). The single blocker both gate reviewers found independently is closed. Verified, not assumed.

### The blocker

**1 — "green" is gone.** `grep -ni green AGENTS.md CLAUDE.md` returns nothing at `491002d`. The three surviving hits in `CONTRIBUTING.md` (213, 230, 233) are about an ordinary PR into a train during a freeze and about the image check on ordinary/draft PRs — neither is about the promotion gate, and 233 is the *correct* rule ("checks go green immediately rather than being skipped"). A repo-wide sweep for `let it go green` returns nothing.

**2 — `Promotion gate` is now named where an agent starts.** `grep -n "Promotion gate" AGENTS.md CLAUDE.md CONTRIBUTING.md` returned nothing at the base `e6d2fd9` and nothing at the parent `71f91da`; at `491002d` it hits `AGENTS.md:21`. That was the half of the finding that mattered most — the guard #264 built is no longer invisible from the file #260 exists to make authoritative.

**3 — consistent with the two instances #264 already fixed.** All three now say the same thing in nearly the same words:

- `promote.yml:199` — "Open it, let its checks settle and get it approved — `Promotion gate` is red on it by design and stays red … (#264); promotion does not read it."
- `docs/ci-cd.md:1081` — "let its checks settle, get it approved"; `:1094` — "**`Promotion gate` is red on that pull request, and red is the healthy state** (#264)"; `:122` — "Red is the healthy state for a gate."
- `AGENTS.md:21` — "let its checks settle, get it approved, then dispatch `promote.yml`. `Promotion gate` is red on that pull request by design and stays red: red is the healthy state there, not a problem to fix (#264)."

Same instruction, same citation, no disagreement left between the three.

**4 — the correction propagates.** `CLAUDE.md` is five lines and still `@AGENTS.md`, untouched by this commit. The fix reaches every Session.

**5 — 260's acceptance criterion 1 still holds.** An agent reading only `AGENTS.md`, following no link, can state both load-bearing sentences literally: *a train is cut by a workflow and never by hand*, and *promotion consumes an existing, approved pull request from the train into `main`* — plus *that pull request is a gate, not a merge: do not press the merge button*, and that the mechanism is dispatching `promote.yml`. AC2/AC3 unaffected: the `See …` tail is byte-identical in the diff, and all four anchors resolve (`CONTRIBUTING.md:172`, `docs/ci-cd.md:81`, `docs/ci-cd.md:958`, ADR 0023 present). AC4 holds: the file is 21 lines, under 30.

Noting for the record, not as an objection: AC1 as written in #260 spells the phrase "existing, **green**, approved". That spelling predates #264, which landed in this same train and made the state it names unreachable; #260's own suggested edit omits "green". Keeping the issue's literal wording would have restored the exact defect. The substance of AC1 is met and the stale word is correctly dropped.

**6 — nothing else moved.** `git diff-tree --name-only -r 491002d` → `AGENTS.md`, and `--stat` → `1 file changed, 1 insertion(+), 1 deletion(-)`. No `docs/ci-cd.md`, no `promote.yml`, no `scripts/promotion-gate.sh`, no `docs/rulesets/main.json`, no `packages/` source, no ADR. One sentence was asked for; one sentence was changed.

### Release-critical checks, re-run on the moved branch

- **Commit body lines ≤ 132.** `git log --format=%B origin/beta/0.3.3..HEAD | awk 'length > 132 {print NR": "length}'` is empty. Longest body line in the range is **118** (in the #264 commit); longest in `491002d` alone is 104. `commitlint.config.mjs:118-119` caps body and footer at 132. This is the check that forced the delete-and-re-cut on 0.3.2; it passes.
- **Six manifests read 0.3.3.** root, `cli`, `core`, `panel`, `sdk`, `shared` — all `0.3.3`. No `"version"` line changed anywhere in the range. The one manifest touched in the range is root `package.json`, and the diff is a single line, `"test"` moving to `node scripts/run-unit-tests.mjs` (#270). `Train rules` is green.
- **No required check's pass case is a skipped job.** `Promotion gate` reports **SUCCESS** on this pull request in 27s — green, not grey — and the job in `ci.yml` carries no job-level `if:` (`permissions`, then `steps`, straight into `scripts/promotion-gate.sh`). The three SKIPPED contexts on this run — `Core image (train)`, `Panel image (train)`, `Resolve train tags` — appear in no required-check list in `docs/rulesets/beta.json` or `main.json`. Every required context for this PR's base is SUCCESS.
- **`promote.yml` is unchanged in how it advances `main`.** The entire range diff against it is one line: the `::error` text in the *no open promotion pull request* refusal branch at `:199`. The PR lookup, head-SHA capture, digest assertion, fast-forward and tagging are untouched.
- **ADR 0031 is still PROPOSED** — "**Status: PROPOSED.** Not accepted… The repository owner ratifies or rejects it at the beta gate." **Issue #29 is still OPEN.** Neither touched by the new commit.

### One thing for the owner, not a blocker

`.agents/skills/release/SKILL.md:41` still reads "**The promotion pull request is open, green and approved**" — a fourth instance of the same stale sentence. It is pre-existing on `main`, untouched by this train, and was left behind by #264 along with the two it did fix. It is not a regression from this PR and neither gate reviewer raised it, so it does not belong in a commit that was scoped to one sentence. It is worth a follow-up issue against #264 so the fourth copy does not outlive the other three. (ADR 0023 D14's "after the promotion pull request is green and approved" is the same wording in a historical decision record, which is a weaker case.)

### Verdict

The blocker is **RESOLVED**. The sentence now matches the two instances #264 fixed, `Promotion gate` is named in the file every agent reads first, `CLAUDE.md` still imports it, 260's AC1 survives, and the commit touches `AGENTS.md` alone. All five release-critical checks pass on the moved branch.

Approving.

mrfqcentic (approved):
Re-review after 491002d. Blocker verified closed, not assumed: (1) 'green' is gone from the sentence and from all of AGENTS.md/CLAUDE.md (grep -i empty); (2) Promotion gate is now NAMED in AGENTS.md:21; (3) wording agrees with promote.yml:199 and docs/ci-cd.md:1081 — all three say checks settle, name Promotion gate, red by design, red healthy, #264; (4) CLAUDE.md still imports @AGENTS.md; (5) sentence still states promotion consumes an existing approved PR from the train into main and the merge button is not the mechanism; (6) the commit touches AGENTS.md alone (+1/-1). Release-critical re-run: body lines all <=132; six manifests still 0.3.3 with no version change in range; Promotion gate reports SUCCESS (green, 27s real run, no job-level if in ci.yml); promote.yml unchanged in how it advances main (one message line only across the range); ADR 0031 still PROPOSED; issue 29 still OPEN. Approving.

Issue #284, opened 2026-08-20
Pairing: SDK client pairing — keypair, CSR, fingerprint-verified dial and code redemption


mrfqcentic:
Part of #280 — short-code pairing enrollment replaces the registration-blob hand-carry. Read that issue first; the design there is fixed.

## Context

This is the client half of the flow, and the one function both the CLI (task 5) and the Panel (task 6) call. It lives in `packages/sdk` because that is the package a third party installs to talk to a Core, and — per [ADR 0025](docs/adr/0025-the-protocol-ships-with-the-client.md) D1 — the protocol ships with it.

**The wire types are declared here, not imported from `@actana/shared`.** This is not a preference; the reasoning is written out at the top of `packages/sdk/src/core-registration-blob.ts` and it applies unchanged: `packages/shared` is private and stays private (ADR 0025 D4), so an SDK importing from it would be a published package with a dependency nobody outside this repository can resolve. Declare the pairing request/response types in the SDK, structurally identical to what the Core answers with, in the same style and with the same comment explaining why.

The result is **`CoreRegistrationBlob`-shaped** (`packages/sdk/src/core-registration-blob.ts`) so that everything downstream is untouched: `coreConnectionFromBlob` unpacks it, `httpsBaseUrlFor` derives the HTTPS origin, `CoreLinkTlsMaterial` takes the PEMs, and the bearer goes into the `auth` frame exactly as today. The `clientKey` field of that result **never left the machine** — it was generated here and is put back into the shape after the response arrives.

The bootstrap dial is the security-critical part. The client has an address and a code, and no trust anchor yet — so it dials, reads the certificate chain the Core presents, computes the SHA-256 fingerprint of the CA in it, and compares that against the fingerprint the operator read out of `actana pair new` (task 3). **The code is only sent after that comparison passes.** A caller with no fingerprint gets the fingerprint reported back to them for confirmation and no code is sent — never a silent `rejectUnauthorized: false` redemption.

## Acceptance criteria

- [ ] A new module in `packages/sdk/src/` (suggested `core-pairing.ts`) exporting one pairing function that takes the Core address, the pairing code and the expected CA fingerprint, and returns a `CoreRegistrationBlob`
- [ ] Keypair generation happens locally via `node:crypto`; a PEM CSR is produced from it carrying the client label/info. The private key is **never** sent, logged, or included in any request body
[...]

Issue #283, opened 2026-08-20
Pairing: core operator CLI — actana pair new, ls and revoke


mrfqcentic:
Part of #280 — short-code pairing enrollment replaces the registration-blob hand-carry. Read that issue first; the design there is fixed.

## Context

The operator on the Core machine needs three verbs: mint a pairing code, see what is pending and what is paired, and take a pairing back. These are the machine-side half of the "a human carries the code" step in #280, and they are the only place the code is ever displayed.

They join the existing verb table in `packages/core/src/actana-cli.ts` (`setup`, `status`, `token`, `update`, `start`, `stop`, `restart`, `logs`, `harnesses`, `uninstall`, `daemon`) as a `pair` verb with sub-verbs, in the style of the CLI package's `actana core <verb>` (`packages/cli/src/core-command.ts`). Note the naming trap: `actana pair` here is the **Core-side operator** command; task 5's `actana core pair` is the **client-side** command in `packages/cli`. Two different binaries' worth of surface, and the help text on both should make which machine you are on obvious.

`actana pair new` prints the code, the CA fingerprint and the expiry. The fingerprint is what makes the client's bootstrap dial verifiable before it sends the code (task 4) — it is read out by a human alongside the code, so its format has to be one a person can compare by eye: the conventional colon-separated uppercase hex SHA-256 of the CA certificate's DER.

## Acceptance criteria

- [ ] `actana pair new [--label <label>] [--ttl <duration>]` creates a pending pairing session on this Core and prints: the `XXXX-XXXX` code, the CA fingerprint, and the expiry as both an absolute time and a relative "expires in 5 minutes"
- [ ] TTL defaults to 5 minutes, from the shared constant in task 1 — not a literal in the CLI
- [ ] The CA fingerprint is SHA-256 over the CA certificate's DER, rendered colon-separated uppercase hex, computed from `PersistedMaterial.caCert`
- [ ] `actana pair ls` lists pending pairing sessions (label, created, expires, attempts used of 5) and paired clients (label, subject, paired-at) in two clearly separated sections; an empty list says so rather than printing an empty table
- [ ] `actana pair ls` never prints a pairing code — a pending session shows its label and expiry, not its secret
- [ ] `actana pair revoke <target>` unpairs a client: the persisted client identity is removed or marked revoked, and its credential stops working — both its client certificate and its bearer
- [ ] Revoking a client with a live core link closes that link rather than leaving it running until the next handshake
[...]

Issue #282, opened 2026-08-20
Pairing: core pairing server — pre-auth endpoint, CSR signing and bearer claims


mrfqcentic:
Part of #280 — short-code pairing enrollment replaces the registration-blob hand-carry. Read that issue first; the design there is fixed.

## Context

This is the server half of the flow: the endpoint a client with a pairing code posts to, and everything the Core does before it hands back a credential. It is the **only pre-auth surface on the Core** — every other route and the core link itself sit behind the mTLS handshake, and the whole point of pairing is that the client does not yet have a client cert to present. That asymmetry is why the security properties in #280 are load-bearing here rather than nice-to-have.

The endpoint mounts on the Core's **existing HTTPS surface** — the same `https.Server` that answers the WebSocket upgrade and the `/v1/…` routes (`packages/core/src/pty-core-link-server.ts`; see the route-hanging comment near the bottom of that file and `packages/core/src/core-files-wiring.ts` / `core-files-routes.ts` for how a route family is built and mounted). It is not a second server and not a second port. Note that the Core's TLS server is configured `requestCert: true, rejectUnauthorized: true`; the pairing route must be reachable by a client that presents no certificate, and arranging that without weakening the mTLS posture of every other route is the substance of this task.

CSR signing goes through the existing CA machinery in `packages/core/src/core-cert-material.ts` — `issueServerCert` is the precedent for "sign a leaf against the CA we already hold", and the client-cert leaf shape (`basicConstraints cA:false`, `keyUsage digitalSignature + keyEncipherment`, `extKeyUsage clientAuth`) is already written out in `generateCertMaterial`. The CA key lives in `PersistedMaterial.caKey` (`packages/core/src/core-material-store.ts`).

The bearer today is `{coreId, exp}` HMAC-SHA256'd with `PersistedMaterial.bearerSecret` (`packages/shared/src/core-link-bearer.ts`). This task adds `iss`, `sub`, `aud` and `jti` claims, with `aud` being the stable core UUID. `PersistedMaterial` already carries a `coreId` of the form `core_<hex>`; the **core UUID is a distinct, stable identifier** that must be minted at install if absent and must survive `reissueServerCert` — mint-if-absent on load, so an existing Core picks one up without a re-install and without invalidating anything a Panel pinned.

## Acceptance criteria

[...]

Issue #280, opened 2026-08-20
Pairing: short-code enrollment replaces the registration-blob hand-carry


mrfqcentic:
A client becomes a **Core client** today by hand-carrying a **Registration blob**: `actana setup` prints `base64(JSON({endpoint, label, caCert, clientCert, clientKey, bearer}))` (CONTEXT.md "Registration blob"), and a human moves that artifact — every byte of it, private key included — to the machine that will use it. The Panel takes it through "Add Core" (`packages/panel/src/components/views/CoresSettingsPage.tsx`, "paste the pairing token here…"); the CLI takes it through `actana core add <name> [file]` (`packages/cli/src/core-command.ts`, decoded by `packages/cli/src/registration-blob-file.ts`); both then reach a Core through `@actana/sdk`.

This feature replaces that hand-carry with a short one-time **pairing code**. The design below is agreed and fixed — this issue restates it so the task issues can point at it, and it is not to be redesigned in the implementing PRs.

## The flow

1. **The operator on the Core runs `actana pair new`** with a label and a TTL defaulting to 5 minutes. The Core creates a **pending pairing session** — single-use, attempt-capped at 5, with a CSPRNG **8-character code** drawn from an unambiguous alphabet (no `0`, `O`, `1`, `I`, `L`) grouped as `XXXX-XXXX` — and prints the code plus the CA fingerprint and the expiry.
2. **A human carries the code to the client.** This is the only manual step in the flow.
3. **The client generates its keypair locally**, dials the Core's HTTPS surface (the same `https.Server` `core-link` already listens on — see `packages/core/src/pty-core-link-server.ts`, where the WebSocket upgrade and the `/v1/…` routes are two faces of one server), verifies the presented CA against the fingerprint it was given, and posts the code, its client info and a CSR.
4. **The Core validates** — TTL, single-use, attempt cap, rate limit, session binding — signs the CSR against its own CA using the existing `packages/core/src/core-cert-material.ts` machinery, persists the client identity, audit-logs the attempt, and returns `caCert`, `clientCert`, `bearer` and `endpoint`.

**The private key never crosses the wire.** The client assembles a `CoreRegistrationBlob`-shaped credential (`packages/sdk/src/core-registration-blob.ts`) from the response plus its own locally-born `clientKey`. Everything downstream is untouched: `coreConnectionFromBlob`, the mTLS handshake, and the bearer `auth` frame all see exactly the shape they see today.

[...]

PR #291, opened 2026-08-20
feat(shared): pairing primitives — code generation and the pairing-session model (#281)

### LINKED ISSUE #281: Pairing: shared pairing primitives — code generation and the pairing-session model

## Context

The pairing flow needs two primitives that both the Core (which mints and validates) and the Core's own tests reason about: the **pairing code** and the **pairing session**. Per [ADR 0025](docs/adr/0025-the-protocol-ships-with-the-client.md) D4, `packages/shared` is private and stays private, and it keeps "everything that is not the protocol" — so these internals live here, in `packages/shared`, alongside `client-id.ts`, `short-id.ts` and `core-link-bearer.ts`. They are **not** wire types: the SDK declares its own, per ADR 0025 D1 and the argument written out at the top of `packages/sdk/src/core-registration-blob.ts`.

Nothing in this task talks to a socket, a disk or a clock it did not receive. Code generation takes a CSPRNG port and session expiry takes a `now`, in the manner of the injectable HMAC port in `packages/shared/src/core-link-bearer.ts`, so both are exercisable without fake timers.

## Acceptance criteria

- [ ] A new module in `packages/shared/src/` (suggested `pairing-code.ts`) exports the code alphabet as a named constant, excluding `0`, `O`, `1`, `I` and `L`
- [ ] Code generation draws 8 characters from that alphabet using a CSPRNG (`node:crypto.randomBytes` / `randomInt`), with **no modulo bias** — reject-and-redraw or `randomInt`, not `% alphabet.length`
- [ ] Generated codes are formatted `XXXX-XXXX`; a parse/normalise helper accepts the code with or without the hyphen and case-insensitively, and returns a canonical form for comparison
- [ ] The CSPRNG is an injectable port with a `node:crypto` default, so tests can assert the draw without stubbing globals
- [ ] A pairing-session module (suggested `pairing-session.ts`) exports the session shape: id, label, code (or code hash), `createdAt`, `expiresAt`, attempt counter, attempt cap, consumed marker, and the three nullable future-proofing fields `created_by`, `tenant_id`, `auth_method` — declared, documented as inert today, and referenced by #280's OIDC note
- [ ] Pure state transitions over that shape: `isExpired(session, now)`, a wrong-attempt transition that increments the counter and kills the session at the cap of 5, and a single-use consumption that refuses a second consume
- [ ] The TTL default of 5 minutes is a named exported constant, not a literal at a call site
- [ ] Unit tests in `packages/shared/src/__tests__/` covering: alphabet excludes the five ambiguous characters; codes are 8 chars and `XXXX-XXXX`-grouped; no modulo bias in the draw; normalise accepts hyphen/no-hyphen/mixed case; expiry at and either side of the boundary; the attempt counter dying exactly at 5; a consumed session refusing a second consume; the three nullable fields defaulting to null
- [ ] Comments follow the house style of the files around them — say why the decision was made, not what the line does

## File pointers

- New: `packages/shared/src/pairing-code.ts`, `packages/shared/src/pairing-session.ts`
- New tests: `packages/shared/src/__tests__/`
- Style/shape precedents: `packages/shared/src/core-link-bearer.ts` (injectable crypto port, exported claim types), `packages/shared/src/short-id.ts`, `packages/shared/src/client-id.ts`
- Boundary rule: `docs/adr/0025-the-protocol-ships-with-the-client.md` D4; the counter-argument for why these are *not* declared in the SDK is at the top of `packages/sdk/src/core-registration-blob.ts`

## Dependencies

**None.** This is the root of the dependency chain — tasks 2 and 3 build on it.

## Branching

qcentic-dev:
Closes #281. Part of #280 — short-code pairing enrollment replaces the registration-blob hand-carry.

**Base is `feat/short-code-pairing`**, the feature branch for #280 — not `main`, not `beta/0.4.0`. That branch did
not exist on the remote yet, so it was cut from `origin/beta/0.4.0` and pushed as-is before this branch was cut
from it; it therefore contains nothing but the train tip plus, in time, the seven task PRs.

## What this adds

Two new modules in `packages/shared/src/`, plus their tests. Nothing else in the repository is touched — no
`exports` map, no imports from anywhere, no call sites. The package's `"./*"` wildcard already exports every file
under `src/`, so this ticket registers itself nowhere and cannot collide with the work landing beside it.

### `pairing-code.ts`

- `PAIRING_CODE_ALPHABET` — 31 characters, `A–Z` and `2–9` less `0`, `O`, `1`, `I` and `L`.
- `generatePairingCode(csprng?)` — draws 8 characters, printed `XXXX-XXXX`. Byte values 248–255 are **rejected and
  redrawn** rather than folded with `%`: 256 is not a multiple of 31, and folding would make the first eight
  characters ~14% likelier than the rest.
- `CsprngPort` — injectable, defaulting to `node:crypto.randomBytes`, in the manner of the HMAC port in
  `core-link-bearer.ts`. It yields bytes rather than integers so the rejection is exercised by the test rather than
  hidden behind `randomInt`.
- `normalisePairingCode(input)` — takes the hyphen or no hyphen, any case, stray whitespace; returns the canonical
  printed form or `null`. Out-of-alphabet characters are refused rather than guessed at.

### `pairing-session.ts`

- `PairingSession` — id, label, `codeHash`, `createdAt`, `expiresAt`, `attempts`, `attemptCap`, `consumedAt`, and
  the three inert future-proofing fields `created_by`, `tenant_id`, `auth_method`, all defaulting to `null`.
- `PAIRING_SESSION_TTL_MS` (5 minutes) and `PAIRING_ATTEMPT_CAP` (5) as named exported constants.
- Pure transitions: `isExpired(session, now)` (inclusive at the boundary, matching `verifyBearer`),
  `recordWrongAttempt`, `consumePairingSession`, and the `canRedeem` gate that names which defence refused.

Nothing here reads a socket, a disk or a clock it was not handed. `now` is always an argument, so no test needs
fake timers.

## Checks

`packages/shared`: 250 tests pass (35 of them new), `tsc --noEmit` clean, `eslint` clean on the four new files.

## Not in scope

The endpoint, CSR signing, persistence, the audit log, rate limiting and the operator verbs are #282 and #283.
These are internals and not wire types — `packages/shared` is private and stays private (ADR 0025 D4), and the SDK
declares its own shapes for whatever crosses the socket (#284).

### DIFF
--- a/packages/shared/src/__tests__/pairing-code.test.ts
+++ b/packages/shared/src/__tests__/pairing-code.test.ts
@@ -1,0 +1,152 @@
+import { describe, it, expect } from "vitest";
+import {
+  PAIRING_CODE_ALPHABET,
+  PAIRING_CODE_GROUP_SIZE,
+  PAIRING_CODE_LENGTH,
+  formatPairingCode,
+  generatePairingCode,
+  normalisePairingCode,
+  type CsprngPort,
+} from "../pairing-code";
+
+// The pairing code is the one thing a human carries between the Core and the
+// client (#280): eight characters from an alphabet with no ambiguous glyphs,
+// drawn from a CSPRNG with no modulo bias, printed `XXXX-XXXX`. The port makes
+// the draw assertable without stubbing `node:crypto`.
+
+/** A port that hands out a scripted byte sequence, one batch at a time. */
+function scriptedCsprng(bytes: number[]): CsprngPort & { calls: number[] } {
+  const queue = [...bytes];
+  const calls: number[] = [];
+  return {
+    calls,
+    bytes(n) {
+      calls.push(n);
+      return Uint8Array.from(queue.splice(0, n));
+    },
+  };
+}
+
+describe("pairing code", () => {
+  describe("alphabet", () => {
+    it("excludes the five ambiguous characters", () => {
+      for (const ch of ["0", "O", "1", "I", "L"]) {
+        expect(PAIRING_CODE_ALPHABET).not.toContain(ch);
+      }
+    });
+
+    it("is 31 unique upper-case alphanumerics", () => {
+      expect(PAIRING_CODE_ALPHABET).toHaveLength(31);
+      expect(new Set(PAIRING_CODE_ALPHABET).size).toBe(31);
+      expect(PAIRING_CODE_ALPHABET).toMatch(/^[A-Z2-9]+$/);
+    });
+  });
+
+  describe("generatePairingCode", () => {
+    it("draws 8 characters grouped XXXX-XXXX", () => {
+      const code = generatePairingCode();
+      expect(code).toMatch(/^[A-HJ-KM-NP-Z2-9]{4}-[A-HJ-KM-NP-Z2-9]{4}$/);
+      expect(code.replace("-", "")).toHaveLength(PAIRING_CODE_LENGTH);
+      expect(code[PAIRING_CODE_GROUP_SIZE]).toBe("-");
+    });
+
+    it("draws only from the alphabet, over many codes", () => {
+      for (let i = 0; i < 200; i += 1) {
+        for (const ch of generatePairingCode().replace("-", "")) {
+          expect(PAIRING_CODE_ALPHABET).toContain(ch);
+        }
+      }
+    });
+
+    it("maps accepted bytes through the alphabet in order", () => {
+      const csprng = scriptedCsprng([0, 1, 2, 30, 31, 32, 61, 62]);
+      expect(generatePairingCode(csprng)).toBe("ABC9-AB9A");
+    });
+
+    it("asks for exactly the bytes it still needs", () => {
+      const csprng = scriptedCsprng([0, 1, 2, 3, 4, 5, 6, 7]);
+      generatePairingCode(csprng);
+      expect(csprng.calls).toEqual([PAIRING_CODE_LENGTH]);
+    });
+
+    it("rejects and redraws the biased bytes instead of folding them", () => {
+      // 248-255 are the byte values with no partner in the last cycle of 31.
+      // Folding them with `%` would over-weight A-H; the draw must skip them
+      // and come back for more bytes.
+      const rejected = [248, 249, 250, 251, 252, 253, 254, 255];
+      const csprng = scriptedCsprng([...rejected, 0, 1, 2, 3, 4, 5, 6, 7]);
+      expect(generatePairingCode(csprng)).toBe("ABCD-EFGH");
+      expect(csprng.calls).toEqual([8, 8]);
+    });
+
+    it("keeps a partially drawn code and asks only for the remainder", () => {
+      const csprng = scriptedCsprng([0, 1, 2, 255, 3, 4, 5, 6, 7]);
+      expect(generatePairingCode(csprng)).toBe("ABCD-EFGH");
+      expect(csprng.calls).toEqual([8, 1]);
+    });
+
+    it("has no modulo bias: every accepted byte value is equally weighted", () => {
+      // The whole accepted range, one character each. If the ceiling were wrong
+      // — 256 instead of 248 — some characters would appear 9 times and others
+      // 8, which is exactly the head start the rejection exists to remove.
+      const accepted = Array.from({ length: 248 }, (_, b) => b);
+      const csprng = scriptedCsprng(accepted);
+      const counts = new Map<string, number>();
+      for (let i = 0; i < accepted.length / PAIRING_CODE_LENGTH; i += 1) {
+        for (const ch of generatePairingCode(csprng).replace("-", "")) {
+          counts.set(ch, (counts.get(ch) ?? 0) + 1);
+        }
+      }
+      expect(counts.size).toBe(PAIRING_CODE_ALPHABET.length);
+      expect([...counts.values()]).toEqual(Array(PAIRING_CODE_ALPHABET.length).fill(8));
+    });
+
+    it("throws rather than spinning when the port yields nothing", () => {
+      expect(() => generatePairingCode(scriptedCsprng([]))).toThrow(/no bytes/);
+    });
+  });
+
+  describe("normalisePairingCode", () => {
+    it("accepts the canonical form unchanged", () => {
+      expect(normalisePairingCode("ABCD-EFGH")).toBe("ABCD-EFGH");
+    });
+
+    it("accepts the code without the hyphen", () => {
+      expect(normalisePairingCode("ABCDEFGH")).toBe("ABCD-EFGH");
+    });
+
+    it("accepts mixed case", () => {
+      expect(normalisePairingCode("aBcD-eFgH")).toBe("ABCD-EFGH");
+      expect(normalisePairingCode("abcdefgh")).toBe("ABCD-EFGH");
+    });
+
+    it("accepts whitespace from a paste or a spoken grouping", () => {
+      expect(normalisePairingCode("  ABCD EFGH \n")).toBe("ABCD-EFGH");
+    });
+
+    it("round-trips a generated code", () => {
+      const code = generatePairingCode();
+      expect(normalisePairingCode(code)).toBe(code);
+      expect(normalisePairingCode(code.replace("-", "").toLowerCase())).toBe(code);
+    });
+
+    it("rejects the excluded characters rather than guessing at them", () => {
+      for (const ch of ["0", "O", "1", "I", "L"]) {
+        expect(normalisePairingCode(`ABCDEFG${ch}`)).toBeNull();
+      }
+    });
+
+    it("rejects the wrong length and other junk", () => {
+      expect(normalisePairingCode("ABCDEFG")).toBeNull();
+      expect(normalisePairingCode("ABCDEFGHJ")).toBeNull();
+      expect(normalisePairingCode("")).toBeNull();
+      expect(normalisePairingCode("ABCD-EF!H")).toBeNull();
+    });
+  });
+
+  describe("formatPairingCode", () => {
+    it("groups a raw draw", () => {
+      expect(formatPairingCode("ABCDEFGH")).toBe("ABCD-EFGH");
+    });
+  });
+});
--- a/packages/shared/src/__tests__/pairing-session.test.ts
+++ b/packages/shared/src/__tests__/pairing-session.test.ts
@@ -1,0 +1,175 @@
+import { describe, it, expect } from "vitest";
+import {
+  PAIRING_ATTEMPT_CAP,
+  PAIRING_SESSION_TTL_MS,
+  canRedeem,
+  consumePairingSession,
+  createPairingSession,
+  isConsumed,
+  isDead,
+  isExpired,
+  recordWrongAttempt,
+  type PairingSession,
+} from "../pairing-session";
+
+// Three separate defences (#280): a TTL, a cap of five wrong attempts, and
+// single use. Every transition here is pure and takes `now` from the caller,
+// so a session can be walked past its expiry without fake timers.
+
+const MINT = 1_700_000_000_000;
+
+function session(overrides: Partial<PairingSession> = {}): PairingSession {
+  return {
+    ...createPairingSession({
+      id: "pair_abc",
+      label: "mehdi-laptop",
+      codeHash: "sha256:deadbeef",
+      now: MINT,
+    }),
+    ...overrides,
+  };
+}
+
+describe("pairing session", () => {
+  describe("createPairingSession", () => {
+    it("expires five minutes after the mint by default", () => {
+      expect(PAIRING_SESSION_TTL_MS).toBe(5 * 60 * 1000);
+      expect(session().expiresAt).toBe(MINT + PAIRING_SESSION_TTL_MS);
+      expect(session().createdAt).toBe(MINT);
+    });
+
+    it("honours an explicit TTL and attempt cap", () => {
+      const s = createPairingSession({
+        id: "pair_abc",
+        label: "l",
+        codeHash: "h",
+        now: MINT,
+        ttlMs: 60_000,
+        attemptCap: 2,
+      });
+      expect(s.expiresAt).toBe(MINT + 60_000);
+      expect(s.attemptCap).toBe(2);
+    });
+
+    it("starts pending: no attempts, not consumed", () => {
+      const s = session();
+      expect(s.attempts).toBe(0);
+      expect(s.attemptCap).toBe(PAIRING_ATTEMPT_CAP);
+      expect(s.consumedAt).toBeNull();
+      expect(isConsumed(s)).toBe(false);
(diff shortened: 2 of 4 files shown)


qcentic-dev:
## Acceptance criteria — #281

Checked one by one against `97ef5f1`. Every item is met; nothing is unchecked.

- [x] **A new module in `packages/shared/src/` (suggested `pairing-code.ts`) exports the code alphabet as a named constant, excluding `0`, `O`, `1`, `I` and `L`**
  `pairing-code.ts:40` — `PAIRING_CODE_ALPHABET = "ABCDEFGHJKMNPQRSTUVWXYZ23456789"`, 31 characters, `A–Z` and `2–9` less the five ambiguous glyphs. Asserted in `pairing-code.test.ts` ("excludes the five ambiguous characters", "is 31 unique upper-case alphanumerics").

- [x] **Code generation draws 8 characters using a CSPRNG, with no modulo bias — reject-and-redraw or `randomInt`, not `% alphabet.length`**
  `pairing-code.ts:55,84–99` — bytes at or above `UNBIASED_BYTE_CEILING` (248 = `256 - 256 % 31`) are **discarded and redrawn**; the remaining 248 values cover the 31 characters exactly 8 times each, which is what makes the `%` on line 94 unbiased rather than merely convenient. Tests: "rejects and redraws the biased bytes instead of folding them", "keeps a partially drawn code and asks only for the remainder", and "has no modulo bias: every accepted byte value is equally weighted" (feeds all 248 accepted values and asserts a flat 8-per-character histogram — a ceiling of 256 would show 9s and 8s).

- [x] **Generated codes are formatted `XXXX-XXXX`; a parse/normalise helper accepts the code with or without the hyphen and case-insensitively, and returns a canonical form for comparison**
  `formatPairingCode` (`pairing-code.ts:74`) and `normalisePairingCode` (`:117`). The canonical form is the printed one, so `normalisePairingCode(generatePairingCode())` is the code itself and a comparison never has to know which side it came from. Out-of-alphabet characters are refused rather than folded — the excluded five have no unambiguous partner to fold *to*, so guessing would spend a caller's attempt on a code they never heard. Tests cover hyphen / no hyphen / mixed case / pasted whitespace / round-trip / excluded characters / wrong length.

- [x] **The CSPRNG is an injectable port with a `node:crypto` default, so tests can assert the draw without stubbing globals**
  `CsprngPort` (`pairing-code.ts:64`), defaulting to `node:crypto.randomBytes`, in the manner of the HMAC port in `core-link-bearer.ts`. It yields **bytes rather than integers** on purpose: the rejection is the property under test, so it stays in code a test can drive instead of hiding behind `randomInt`. No test stubs a global.

- [x] **A pairing-session module exports the session shape: id, label, code (or code hash), `createdAt`, `expiresAt`, attempt counter, attempt cap, consumed marker, and the three nullable future-proofing fields `created_by`, `tenant_id`, `auth_method` — declared, documented as inert today, and referenced by #280's OIDC note**
  `PairingSession` (`pairing-session.ts:49–75`): `id`, `label`, `codeHash`, `createdAt`, `expiresAt`, `attempts`, `attemptCap`, `consumedAt`, plus the three nullable fields. `codeHash` rather than the code so a stolen session store is not a pile of live pairing codes; `consumedAt` is a timestamp rather than a boolean because the audit log wants to say *when*; `attemptCap` is copied at mint so a config change cannot retroactively revive or kill sessions already in flight. The three fields are documented as inert and cite #280's "Future-proofing for a later OIDC layer (all inert today)" note; they keep the snake_case of the columns they will become so the identity work does not have to rename anything to find them.

- [x] **Pure state transitions: `isExpired(session, now)`, a wrong-attempt transition that increments the counter and kills the session at the cap of 5, and a single-use consumption that refuses a second consume**
  `isExpired` (`:120`), `recordWrongAttempt` (`:164`), `consumePairingSession` (`:177`), with `isDead`, `isConsumed` and a `canRedeem` gate that returns a **typed reason** naming which of the three defences refused — the same shape `verifyBearer` returns, so the Core can audit-log the cause. `isExpired` is inclusive at the boundary, matching `verifyBearer`'s `exp < now`. Every transition is pure and returns a new session, so a caller that fails to persist has not already changed its in-memory copy. `recordWrongAttempt` stops the counter *at* the cap rather than climbing past it, so no later reader has to reason about `attempts > attemptCap`.

- [x] **The TTL default of 5 minutes is a named exported constant, not a literal at a call site**
  `PAIRING_SESSION_TTL_MS = 5 * 60 * 1000` (`pairing-session.ts:29`), alongside `PAIRING_ATTEMPT_CAP = 5` (`:32`). Neither number appears at a call site.

- [x] **Unit tests in `packages/shared/src/__tests__/` covering every listed case**
  `__tests__/pairing-code.test.ts` and `__tests__/pairing-session.test.ts`, one test per required case:
  | Required case | Test |
  |---|---|
  | alphabet excludes the five ambiguous characters | `alphabet › excludes the five ambiguous characters` |
  | codes are 8 chars and `XXXX-XXXX`-grouped | `generatePairingCode › draws 8 characters grouped XXXX-XXXX` |
  | no modulo bias in the draw | `generatePairingCode › has no modulo bias…` (+ the two rejection tests) |
  | normalise accepts hyphen / no hyphen / mixed case | `normalisePairingCode › accepts the canonical form unchanged`, `…without the hyphen`, `…mixed case` |
  | expiry at and either side of the boundary | `isExpired › is live before the boundary`, `…live exactly at the boundary`, `…expired one millisecond past` |
  | the attempt counter dying exactly at 5 | `recordWrongAttempt › dies at exactly the fifth wrong code, not the fourth` |
  | a consumed session refusing a second consume | `consumePairingSession › refuses a second consume` |
  | the three nullable fields defaulting to null | `createPairingSession › defaults the three future-proofing fields to null` |

- [x] **Comments follow the house style of the files around them — say why the decision was made, not what the line does**
  File headers state the decision and its cost (the alphabet's ~3 bits of entropy against a code that survives being spoken; the ~14% head start `%` would hand a code sprayer; the three defences kept in separate fields so a change to one cannot silently weaken another), matching the header of `core-link-bearer.ts`. `eslint` clean on all four files.

### Checks

`packages/shared`: **250 tests pass** (18 files, 35 of them new), `tsc --noEmit` clean, `eslint` clean on the four new files.

### Scope

Four files, all under `packages/shared/src/` — nothing else in the repository is touched. **No `exports` map was added to `packages/shared/package.json`**: the existing `"./*"` wildcard already exports every file under `src/`, so this ticket registers itself nowhere, has no call sites, and cannot collide with the sibling tasks landing beside it. The endpoint, CSR signing, persistence, the audit log, rate limiting and the operator verbs are #282/#283; the wire types are the SDK's own to declare (#284), per ADR 0025 D1/D4.

Still a **draft**, based on `feat/short-code-pairing`.

mrfqcentic (approved):
## Review of #291 against #281 — approved

Reviewed at `97ef5f1` against the merge base `75d61ab` on `feat/short-code-pairing`. Every claim below was checked by reading the enforcing line and, where the property is behavioural, by executing the two modules standalone — not from the PR description.

### Acceptance criteria

| # | Criterion (#281) | Verdict |
|---|---|---|
| 1 | New module exporting the code alphabet as a named constant, less `0` `O` `1` `I` `L` | **Met** |
| 2 | 8 characters from a CSPRNG with no modulo bias — reject-and-redraw, not `%` | **Met** |
| 3 | `XXXX-XXXX` format; normalise takes hyphen/no-hyphen and any case, returns a canonical form | **Met** |
| 4 | CSPRNG is an injectable port with a `node:crypto` default | **Met** |
| 5 | Session shape: id, label, code hash, `createdAt`, `expiresAt`, attempts, cap, consumed marker, three nullable fields | **Met** |
| 6 | Pure transitions: `isExpired(session, now)`, wrong-attempt to the cap of 5, single-use consume | **Met** |
| 7 | TTL default of 5 minutes as a named exported constant | **Met** |
| 8 | Unit tests covering all eight listed properties | **Met** |
| 9 | Comments in the house style — why, not what | **Met**, with three inaccuracies noted below |

### What was verified, and how

**The alphabet.** `packages/shared/src/pairing-code.ts:40` — `"ABCDEFGHJKMNPQRSTUVWXYZ23456789"`. 31 characters, all unique; `I` and `L` absent from the letter run, `O` absent between `N` and `P`, digits `2–9` only. All five ambiguous glyphs confirmed absent by execution rather than by eye.

**No modulo bias — the rejection loop itself.** `pairing-code.ts:55` sets `UNBIASED_BYTE_CEILING = 256 - (256 % 31) = 248`; line 93 is the rejection, `if (byte >= UNBIASED_BYTE_CEILING) continue;`, and only then does line 94 fold with `% 31`. That is rejection sampling, and the ceiling is the right one: over `[0, 247]` every one of the 31 characters has exactly 8 preimages — verified by enumerating all 248 values, which yields a single distinct preimage count of `8`. Folding the full byte range instead would give `A–H` nine preimages against eight. The tests pin this properly: `__tests__/pairing-code.test.ts:88-102` feeds all 248 accepted values and asserts every character lands exactly 8 times; `:72-80` asserts 248–255 are rejected *and redrawn* (`calls === [8, 8]`); `:82-86` asserts a partial draw asks only for the remainder. A chi-square over 500,000 real draws (4M characters, df = 30) gives **29.5** against a 99.9% critical value of ~59.7 — uniform, no detectable bias. The rejection costs a second port call 21.4% of the time and three or more 0.7% of the time.

**Both ports are injectable.** `CsprngPort` at `pairing-code.ts:64-67`, defaulted at `:69-71`, taken as a parameter at `:84` — the same arrangement as `HmacPort` in `core-link-bearer.ts:40-46`. `pairing-session.ts` imports nothing at all: `now` is a required argument on `createPairingSession`, `isExpired`, `canRedeem` and `consumePairingSession`, and neither file contains `Date.now()`, `process` or `fs`. Stricter than the precedent, in fact — `verifyBearer` defaults `now` to `Date.now()`; these do not.

**Blast radius.** Diffed against the merge base: four files, all under `packages/shared/src/`, +635/−0. Nothing outside that directory is touched.

**Why this is parallel-safe.** One correction to how the PR body phrases it: `packages/shared/package.json` *does* carry an `exports` map — it is the single wildcard `"./*": "./src/*.ts"`, byte-identical to base and untouched here. There is no barrel file in `packages/shared/src/`, and consumers import by path (`@actana/shared/harnesses`, and so on). The accurate statement is not "no exports map" but "the wildcard already covers every new file, so this ticket adds no manifest entry a sibling PR could collide on" — which is the property that actually keeps the wave parallel-safe, and it holds. `vitest.config.ts`'s `src/**/*.test.ts` picks the new tests up the same way, with no registration.

**Behaviour.** All 18 behavioural assertions run directly against the two modules pass: expiry at `expiresAt - 1` / `expiresAt` / `expiresAt + 1` (inclusive at the boundary, matching `core-link-bearer.ts:135`, and documented as doing so); the counter dying at exactly the fifth wrong code and not the fourth; a consumed session refusing a second consume; the three future-proofing fields defaulting to `null`. 200k round-trips of `normalisePairingCode` over mangled valid codes gave zero mismatches, and it never returns a non-canonical string — out-of-alphabet characters, wrong lengths and a zero-width space are all refused rather than guessed at. CI is green on all 20 checks, Typecheck, Lint and Unit Tests included.

### Follow-ups — none blocking, the first worth landing before #283

**1. `pairing-session.ts:96` and `:104` — a `NaN` `ttlMs` or `attemptCap` silently switches off two of the three defences.** `??` substitutes only for `null`/`undefined`, so `NaN` flows through: `expiresAt` becomes `NaN`, and `isExpired`'s `now > NaN` is `false` for every `now`. Executed, not theorised:

```
ttlMs=NaN      -> expiresAt = NaN; canRedeem a year later = { ok: true }
attemptCap=NaN -> attempts = NaN after 50 wrong guesses; isDead = false; consume still succeeds
```

The reachable path is #283's `actana pair new --ttl <value>`: `Number("5m")` and `parseInt("")` are both `NaN`, so a plausible parse produces an immortal, uncappable pairing code rather than an error. Nothing in the tree calls this yet, which is why it is not blocking — but the module header argues that the three defences are deliberately separate "so a change to one cannot silently weaken another", and a single `NaN` weakens two of them at once. The sibling module already sets the precedent: `verifyBearer` refuses a payload unless `Number.isFinite(obj.exp)` (`core-link-bearer.ts:131`). A `Number.isFinite` and positivity guard belongs at `createPairingSession` for the same reason, and it is cheaper here than in every future caller. A negative `ttlMs` fails safe (born expired), so only the non-finite case actually bites.

**2. `pairing-session.ts:134` — the refusal-order comment is backwards.** "Ordered by how it is checked" sits above `"expired" | "attempts-exhausted" | "already-consumed"`, but `canRedeem` (`:149-151`) checks consumed → dead → expired, the exact reverse. The order is load-bearing — it decides which reason the audit log records for a session that is both consumed and expired, and `pairing-session.test.ts:166` deliberately pins consumption being reported ahead of expiry — so a reader who trusts the comment gets the audit semantics backwards. Either reorder the union or drop the claim.

**3. `pairing-code.ts:11-12` and `:16-18` — two numbers in the security rationale are wrong.** The code is right in both cases; only the arithmetic in the justification is off, and both figures are repeated in the commit messages and the PR body, which is how a wrong number gets quoted later.

- `:11-12` — "costs ~3 bits over the eight characters (31^8 ≈ 2^39.6 against 36^8 ≈ 2^41.3)". The parenthetical contradicts the claim in front of it: 41.36 − 39.63 is **~1.7 bits**, not ~3.
- `:16-18` — "`randomBytes(1)[0] % 31` would make the first eight letters ~14% likelier than the rest". With nine preimages against eight it is **12.5%** (9/8 = 1.125).

Neither changes the decision the comment defends: dropping the five glyphs is still cheap, and folding with `%` is still a real head start for anyone spraying the endpoint.

**4. `pairing-session.ts:172-176` — the single-use guarantee is a claim about the store, not about this function.** "The Core writes the returned session back and any replay of the same code meets this branch" holds only if that write is a compare-and-swap. Two concurrent redemptions of the same correct code both read `consumedAt === null`, both pass `canRedeem`, and both get a CSR signed; likewise N parallel wrong guesses all read `attempts === 0` and each write `1`, so the cap of five never trips. Read-modify-write over a caller-loaded object is the right shape for a pure module — worth stating the atomicity requirement here as a constraint on #282's persistence rather than leaving the guarantee reading as a property of these three lines.

One observation rather than a finding: nothing here compares a code or a digest, so constant-time comparison is not this module's problem. It will be the redemption endpoint's, and `codeHash` being carried but never compared here is the right split.

### Delivery

Base is `feat/short-code-pairing` as #281 requires — not `main`, not `beta/0.4.0` — and that branch is currently identical to the `beta/0.4.0` tip (`75d61ab`), so the PR body's account of how it was cut checks out. Two commits, one module each, both arguing the decision rather than restating the diff. The PR is still marked draft; flipping that is the author's call, not a reviewer's.

Every acceptance criterion in #281 is met. Approving — the four items above are follow-ups, and only the first has any teeth, before #283 wires `--ttl` to a parser.

Issue #289, opened 2026-08-20
session send has no way to await the resulting turn


mrfqcentic:
Confirmed real on the orchestrator side. This body is a rewrite: what follows was read off the tree at `51164f1`, and the suggested fix in the previous version has been **replaced** — it did not work, and the reason is under [Landmines](#landmines).

## What is true today (verified, not claimed)

**`session send --wait` is a hard usage error, not an ignored flag.** `sessionSend` calls `misusedFlag(args, ["--enter"])` (`packages/cli/src/session-command.ts:459`), and `misusedFlag` (`packages/cli/src/session-command.ts:621`) *rejects* rather than ignores — its own comment says why. `actana session send X --wait` exits `EXIT_USAGE` with "`--wait` does not apply here".

**Nothing on the send path waits.** `SessionGateway.send` (`packages/cli/src/session-gateway.ts:348`) resolves the PTY once, calls `client.write`, and returns a boolean — delivered or not. No status is read, no event is watched.

**`--wait` is registered for `start` and `resume` only.** The accepted-flag tables say so (`packages/cli/src/session-command.ts:188-189` for `start`, `:238` for `resume`), and the help text says so out loud: "`--wait   start/resume: block until the Core reports it settled`" (`packages/cli/src/session-command.ts:89`). The wait itself is `CoreSession.waitForIdle`, reached at `packages/cli/src/session-gateway.ts:487`.

**`session resume --wait` is not a workaround.** This is the first thing a reader will reach for, and it does not work. `SessionGateway.resume` (`packages/cli/src/session-gateway.ts:288`) asks the Core for the Task's live PTY and, if there is one, throws `already-running` (`packages/cli/src/session-gateway.ts:296-300`). Resume only continues a **stopped** Session, by launching a fresh harness process on the recorded harness session id. For a Session that is still alive — the exact case this issue is about — there is no awaitable path at all.

## Consequence

The first turn of a Session is awaitable. Every turn after it is not. A caller that sends a follow-up prompt into a running Session has no supported signal for "that turn is done".

The orchestration skill payload reflects the gap: `packages/cli/src/orchestration-skill-payload.ts:104` documents `actana session send <id> "text"` as a write and nothing more, and `:80-93` documents `start --wait --json` as *the* way to collect a result. Nothing there describes awaiting a follow-up turn, because there is nothing to document. (Note the file is duplicated byte-for-byte at `packages/shared/src/orchestration-skill-payload.ts` — see Dependencies.)

[...]

Issue #287, opened 2026-08-20
Pairing: remove the hand-carry — delete blob emission, blob paste and their docs


mrfqcentic:
Part of #280 — short-code pairing enrollment replaces the registration-blob hand-carry. Read that issue first; the design there is fixed.

## Context

With tasks 5 and 6 landed, both clients pair by code and nothing needs the hand-carried blob. This task removes it. **There is no backward compatibility here and there is no deprecation period** — that is #280's decision, not this task's to soften. A blob-emission path left behind "just in case" is a second way to become a Core client, with its own security properties, that nobody is testing.

The blob is emitted by `actana setup` (`encodeRegistrationBlob` in `packages/core/src/actana-setup.ts`, the `blob` field documented there as "the pairing token — base64 of the Registration blob") and by `actana token`. It is consumed by `actana core add` (`packages/cli/src/core-command.ts` + `packages/cli/src/registration-blob-file.ts`) and by the Panel's paste box. It is described in CONTEXT.md's **Registration blob** glossary entry and in several docs.

Two things this removal must **not** touch. The `CoreRegistrationBlob` *shape* stays — it is what the pairing call returns and what `coreConnectionFromBlob` unpacks (`packages/sdk/src/core-registration-blob.ts`); what goes away is the base64 artifact and the hand-carry. And `packages/shared/src/registration-blob.ts`'s encode/decode should only go if nothing else reads it — check before deleting rather than after.

The vocabulary needs settling in the same PR. CONTEXT.md currently says user-facing strings call the blob "the pairing token" while code and docs say **Registration blob**. After this train "pairing code" is the short code and the blob artifact does not exist, so the glossary entry has to be rewritten rather than deleted — a **Registration blob** is still a real concept (the credential a paired client holds), it is just no longer something a human carries.

## Acceptance criteria

- [ ] The blob-emission path is deleted from `actana setup` — the `blob` field, `encodeRegistrationBlob` usage, and the printed artifact — and setup's output tells the operator to run `actana pair new` instead
- [ ] `actana token` no longer emits a blob for hand-carry; if the verb survives for another purpose, its output and help say what that purpose is
- [ ] `actana core add` and its blob decoding entry points are removed: `packages/cli/src/registration-blob-file.ts` and the `add` verb in `packages/cli/src/core-command.ts`, including its file/`-`/stdin handling and the help text that documents piping a blob
[...]

Issue #285, opened 2026-08-20
Pairing: CLI client command — actana core pair replaces the blob paste


mrfqcentic:
Part of #280 — short-code pairing enrollment replaces the registration-blob hand-carry. Read that issue first; the design there is fixed.

## Context

`actana core add <name> [file]` today reads a base64 blob from a file, from `-`, or from a pipe, decodes it through `packages/cli/src/registration-blob-file.ts`, and stores it under the cores config dir at mode 0600 (`packages/cli/src/blob-registry.ts`). Its own help says as much: "`actana setup` prints one blob per Core. Save it, or pipe it."

This task replaces that gesture with `actana core pair <name> <address> <code>` — the operator types an address and an eight-character code, the CLI calls the SDK pairing function from task 4, and the credential it gets back is written to exactly the same place, in exactly the same format, at exactly the same permissions. **Everything after the credential lands is unchanged**: `core-resolution.ts`, `core-connection.ts` and every verb built on them see the same stored shape they see today.

Note the naming trap: `actana core pair` here is the **client-side** command; `actana pair new` (task 3) is the **Core-side operator** command in `packages/core`. The help text on both should make which machine you are on obvious.

## Acceptance criteria

- [ ] A `pair` verb on `actana core`, taking a local name, the Core's address and the pairing code, with the code accepted hyphenated or not and case-insensitively (the normaliser from task 1, reached through the SDK's public surface — the CLI does not import from `@actana/shared`)
- [ ] The CA fingerprint the operator read off `actana pair new` is supplied by the operator — a flag, an argument, or an interactive confirmation of the presented fingerprint — and the code is **never** sent before it is verified
- [ ] The command wraps the task 4 SDK function and does no crypto of its own; the keypair is born inside the SDK call on this machine
- [ ] The resulting credential is stored under the cores config dir at mode 0600 through the existing registry, so `actana core ls`, `use`, `rm`, `status`, `shell` and `exec` work against it with no change
- [ ] Name validation, "name already registered" handling and the config-dir resolution match what `actana core add` does today
- [ ] `--verbose` explains the steps and, like today, **never prints the credential** — nor the pairing code, nor the private key
- [ ] Each distinguishable SDK error (wrong/expired/consumed code, attempts exhausted, rate limited, fingerprint mismatch, unreachable) gets a one-line operator-facing message that says what to do next, and a distinct exit code from `packages/cli/src/exit-codes.ts`
[...]

PR #293, opened 2026-08-20
refactor(cli): there is one `actana` — the Core manager and the client, one program

### LINKED ISSUE #288: One actana CLI — unify the Core operator CLI and the client CLI into a single binary
> **The FIRST fix of the `0.4.0` train.** Two different programs both install a binary
> named `actana`. On any machine that has both — which is every Core whose Sessions are
> meant to drive Cores, exactly the machine [ADR 0031](https://github.com/actana/control/blob/main/docs/adr/0031-the-product-ships-one-skill.md)
> installs the `actana-sessions` skill onto — they collide on `PATH`. This issue says the
> split itself is the defect.

## The problem, from the operator who hit it in production today

There are two `actana` binaries and they are different programs.

**The Core tarball ships the operator CLI.** `packages/core/src/actana-cli.ts` — its own
first line is *"the `actana` CLI — one command that owns a Core's machine-side
lifecycle"*. Its dispatch switch (`packages/core/src/actana-cli.ts:1111-1135`) knows
exactly these verbs: `setup`, `status`, `token` (with `token regenerate`), `update`,
`start`, `stop`, `restart`, `logs`, `harnesses`, `uninstall`, and the undocumented
`daemon` that the unit / LaunchAgent execs. Anything else falls to
`unknown command: ${verb}` (`:1135`).

**npm ships the client CLI.** `@actana/cli` declares `"bin": { "actana": "bin/actana.mjs" }`
(`packages/cli/package.json`), and `packages/cli/src/actana-cli.ts` dispatches by noun
(`:146-171`): `core`, `project`, `harness`, `events`, `session`. Its header states the
split as an intention — *"One command name, split by noun… **The published package
carries the client subcommands only**"* (`packages/cli/src/actana-cli.ts:1-9`) — and it
already carries a special case for the collision it produces
(`packages/cli/src/actana-cli.ts:163-168`):

```
actana daemon: running a Core is not this package's half of `actana`.
It ships with the Core itself — see INSTALL.md.
```

with the comment saying why the message exists: *"somebody typing this has the right
command name and the wrong package, which is exactly the confusion one binary name split
by noun (#129 D8) can produce."*

### On a container Core today, concretely

1. **The image ships only the operator binary.** The Core tarball is extracted flat into
   `/opt/actana` (`deploy/core.Dockerfile:205-207`), so the only `actana` on the image is
   the operator one at `/opt/actana/bin`.

2. **The Core installs a skill that teaches verbs that binary does not have.** ADR 0031 D3
   makes installation mandatory and `packages/core/src/core-entry.ts:302` runs
   `ensureOrchestrationSkill(os.homedir())` at boot. The installed skill
   (`.agents/skills/actana-sessions/SKILL.md`) teaches `actana core ls`, `actana core use`,
   `actana harness ls --json`, `actana project ls --json` (`:21-25`),
   `actana session start … --wait --json` (`:42-44`, `:60`), `actana session logs|send|kill`
   (`:83-85`) and `actana events tail --json` (`:88`). **Every one of those is an
   `unknown command` to the binary that is actually on that container's `PATH`.** The skill
   is dishonest on the machine the Core itself put it on.

3. **Installing the npm CLI to fix that creates a second `actana` whose precedence is a
   `PATH` accident.** `deploy/core.Dockerfile:252-253`:

   ```
   NPM_CONFIG_PREFIX=/home/core/.local
   PATH=/home/core/.local/bin:/opt/actana/bin:/usr/local/bin:/usr/bin:/bin:...
   ```

   The npm prefix's `bin` is **first**, ahead of `/opt/actana/bin`. So `npm i -g @actana/cli`
   inside a container Core silently shadows the operator binary for every subsequent
   command in that container — including `CMD ["actana", "daemon"]`
   (`deploy/core.Dockerfile:266`), which resolves through the same `PATH` and would, on the
   next container start, land on the client CLI's refusal quoted above instead of starting
   a Core. Reverse the ordering and the failure inverts: the skill's verbs go back to
   `unknown command`, or a client noun answers `core not found` from the wrong program.
   **Which confusing failure an operator gets is decided by npm prefix and `PATH` ordering,
   not by anything they can see or reason about.**

### This was already documented as a constraint, and treated as one to work around

[#265 §6](https://github.com/actana/control/issues/265) — *"'Authored once' needs a
mechanism — the CLI and the Core cannot share a module"* — states the two-binaries fact
plainly and then designs around it: because `@actana/cli` may not import `@actana/shared`
(ADR 0025 D4) and `@actana/core` may import only two named modules from `@actana/sdk`
(ADR 0025 D2 as amended by #224), the skill payload is authored once and embedded into
both packages by a generator, with a drift test
(`packages/shared/src/__tests__/orchestration-skill-fanout.test.ts`) holding the copies
honest. ADR 0031 D8 records the same reasoning.

That machinery is correct given the premise. **This issue rejects the premise.** The
duplicated installer, the duplicated target table and the drift test all exist because one
product ships two programs under one name — and so does the `actana daemon` refusal
message, and so does the `PATH` hazard above. The split is the defect; #265 §6 documented
its symptom.

**The operator states they raised this overlap repeatedly and it must not survive into
`0.4.0`.**

## The mandate — decided, not up for redesign

**There is ONE `actana` CLI. One binary, one name, one program. It is the Core manager AND
the client.**

- The operator verbs — `setup`, `status`, `token`, `token regenerate`, `update`, `start`,
  `stop`, `restart`, `logs`, `harnesses`, `uninstall`, `daemon` — and the client nouns —
  `core`, `project`, `session`, `events`, `harness` — live in the same CLI.
- **Context-sensitivity stays**, where a verb only makes sense on a machine that has a
  Core. The container refusal behaviour is kept as it is today: the verbs that belong to
  the container runtime refuse and name their Docker equivalent
  (`refusedContainerVerbs()` / `CONTAINER_USAGE`, `packages/core/src/actana-cli.ts:203-224`,
  `packages/core/src/actana-container.ts`), driven by `ACTANA_CONTAINER` and never by
  sniffing `/.dockerenv` (`deploy/core.Dockerfile:240-246`, ADR 0016 D16).
- **The Core image ships this unified CLI**, so a Session running on a Core can drive Cores
  out of the box and the `actana-sessions` skill is honest on the machine it lands on.
- **The npm package and the tarball stop shipping rival binaries under one name.** How the
  packaging converges — one package, or one shipping the other — is the implementer's
  design space. **The end state of one program under one name is fixed.**
- **ADR 0025's boundaries must be respected in whatever shape is chosen**: `@actana/shared`
  stays private (D4), and the protocol ships with the client (D1/D2, and D2's import list
  as amended by #224). A convergence that publishes `@actana/shared` or that widens that
  list without its own argument is not an acceptable answer to this issue.
- **An ADR documenting the unification is expected**, and it supersedes the two-binaries
  premise that #265 §6 and ADR 0031 D8 were written against.

**The design space the fourth bullet left open is now closed.** Every question that was
open when this issue was written has been settled, below. They are decisions, not options:
an implementer does not choose between them, and a reviewer who disagrees with one takes it
up here rather than in a PR. Every file, line and test named below was checked against
`main` at `51164f1` before it was written here.

### D1 — Where the code lives: the operator modules move into `packages/cli`

These ten modules move **out of `packages/core/src` and into `packages/cli/src`**:

```
actana-cli.ts      actana-setup.ts     actana-layout.ts    actana-system.ts
actana-container.ts  actana-update.ts  actana-uninstall.ts actana-config.ts
actana-manifest.ts   actana-release.ts
```

Their sibling entry `packages/core/src/actana-cli-entry.ts` and its tests move with them —
an entry point whose only job is to supply the real process to `actana-cli.ts` does not
stay behind. **After the move, `packages/core` is the daemon and nothing else.**

**The ten are the decided set; D3's ban forces their closure to follow.** A module left
behind becomes a `@actana/core` import from `packages/cli/src`, which is exactly what
`no-local-escape.test.ts:116` fails on — so the move is complete or it is red. The closure,
read off the imports on `main`:

| Also moves to `packages/cli/src` | pulled in by |
| --- | --- |
| `actana-service.ts` | `actana-cli`, `actana-setup`, `actana-update`, `actana-uninstall`, `actana-status` |
| `actana-systemd.ts`, `actana-launchd.ts` | `actana-layout`, `actana-service` |
| `actana-tree.ts` | `actana-setup`, `actana-update`, `actana-uninstall` |
| `actana-status.ts` | `actana-cli` |
| `actana-harnesses.ts` | `actana-cli`, `actana-setup` |
| `operator-login-path.ts` | `actana-harnesses` |

`log.ts` is the one shared leaf: `actana-config.ts` imports it and so do 22 other files in
`packages/core/src`. It is twenty lines of console-with-a-tag — either it moves to
`packages/shared` or `packages/cli` gets its own. Either is fine; what is not fine is
`packages/cli` importing the Core's.

**The rejected alternative: leave them in `packages/core` and let the published CLI import
them,** rewriting the import ban as a permitted-module list the way [ADR 0025
D2](https://github.com/actana/control/blob/main/docs/adr/0025-the-protocol-ships-with-the-client.md)
does for `@actana/sdk/core-link-frames` and `@actana/sdk/core-files-error-codes`.
**Rejected because it turns a bright line into a judgement call at every future review.**
D2's list works because it is two files that import nothing at all; a list that admits ten
modules which between them reach `child_process`, the filesystem, systemd and the network
is not a list a reviewer can check — it is a permission to argue. The move keeps the ban
readable as "none".

The tarball's CJS bundle follows the source. `scripts/build-core-tarball.mjs:231` requires
both `app/core-entry.cjs` and `app/actana-cli.cjs` to exist in the staged tree; today the
second is emitted by `packages/core/build.mjs` (`outfile: "dist/actana-cli.cjs"`). After
the move it is built from `packages/cli/src`, and `packages/cli` therefore emits **two**
bundles: the published ESM one (`dist/actana-cli.mjs`, `format: "esm"`, `external: ["ws",
"undici"]`) and the tarball's CJS one. That check at `:231` is the enforcing line — it
fails the tarball build if the wiring is wrong.

### D2 — `core-material-store.ts` goes to `packages/shared`, not to `packages/cli`

Both halves use it, so it belongs to neither. Verified on `main`:

- **The daemon side** — `packages/core/src/core-first-run.ts:45` imports it, and
  [#282](https://github.com/actana/control/issues/282) (the Core pairing server) touches it
  from the Core side.
- **The operator side** — `packages/core/src/actana-cli.ts:37` and
  `packages/core/src/actana-setup.ts:34` import it.

Putting it in `packages/cli` would leave the daemon importing the client package, which is
the arrow this whole issue exists to keep straight. It goes to `packages/shared`.

**It does not travel alone.** `core-material-store.ts:17` imports
`./core-cert-material`, which is the module that reaches for `selfsigned`, and
`core-cert-material.ts` is separately imported by the daemon (`core-entry.ts:92`) and by
six Panel tests as `@actana/core/core-cert-material`. Either it moves to `packages/shared`
with the store, or `packages/shared` ends up importing `@actana/core`. It moves; the Panel
tests' import specifiers are repointed. See the consequence in **C2** below, which this
sets up.

### D3 — The daemon import ban survives verbatim, and passing it is the proof

`packages/cli/src/__tests__/no-local-escape.test.ts:116` — *"carries no daemon: nothing
here imports the Core package"* — sweeps every shipped source in `packages/cli/src` and
fails CI if any of them imports `@actana/core`. **After the move that test still passes,
unchanged, and it MUST keep passing. That is the proof the move was done right** — a
partial move shows up as a `@actana/core` import left behind, and the ban is what makes it
show up in CI rather than in a stranger's `npm install`.

Its stated reason survives untouched: *"An import of it would put a Node daemon,
`better-sqlite3` and `node-pty` in the dependency graph of a package whose entire job is to
need none of them."* A published CLI does not need a database driver or a PTY addon, and it
will not have them.

Note for the implementer: the sweep is `shippedSources()`, which excludes `__tests__`
(`no-local-escape.test.ts:41`). `packages/cli/src/__tests__/in-process-core.ts:63` already
imports `@actana/core/core-cert-material` and is exempt by design — that is not a breach and
does not need fixing.

### D4 — How `daemon` reaches the runtime: a path, not an import

**The `daemon` verb resolves `current/app/core-entry.cjs` from the install root and loads it
in-process.** It is a path, not an import, so the daemon's native dependencies never enter
the published dependency graph — which is precisely what lets D3's ban stand while one
binary does both jobs. The mechanism already exists at
`packages/core/src/actana-cli-entry.ts:77` (`require(path.join(__dirname,
"core-entry.cjs"))`) with the reason in its own comment: *"Requiring rather than importing
keeps the daemon out of this bundle."* What changes is only which package that file lives
in, and that the root it resolves from is the install root
(`ACTANA_ROOT`, falling back to `__dirname/..`; `actana-layout.ts` names the `current`
symlink at `:108`).

**In-process rather than spawned is deliberate and must not be changed.** systemd's
`Type=simple` and launchd both expect the daemon to BE the process they started. This is
already documented at the top of `packages/core/src/actana-cli-entry.ts:7-10`:

```
The `daemon` verb loads the sibling `core-entry.cjs` in-process rather
than spawning it: systemd's `Type=simple` and launchd both expect the daemon
to BE the process they started, and an extra fork in between would leave the
init system supervising a wrapper that has already exited.
```

That comment moves with the file and stays true.

### D5 — Bundling `@actana/shared` into the published artifact is allowed, and the ADR argues it

esbuild inlines `packages/shared`'s source into the published bundle while **the package
itself stays unpublished** — no manifest on the registry, no version, no import path,
nothing a third party can resolve or depend on. [ADR 0025
D4](https://github.com/actana/control/blob/main/docs/adr/0025-the-protocol-ships-with-the-client.md)
exists so that nobody can take a dependency on `shared`'s surface, and **an inlined bundle
offers no surface to depend on.** D4's own words are *"it is not deleted and it is not
published"*, and inlining keeps both halves of that sentence true.

This is not optional and it is not incidental: `packages/core/src/actana-setup.ts:24-25,42`
already imports `@actana/shared/core-link-bearer`, `@actana/shared/registration-blob` and
`@actana/shared/domain`. The moment that file lands in `packages/cli/src`, the shared ban
fires. It has to be handled deliberately.

**THIS MUST BE ARGUED EXPLICITLY IN THE NEW ADR, not assumed.** An unpublished package
inlined into a published artifact is exactly the shape ADR 0025's rejected-alternatives
section warns about, and the difference — *no resolvable specifier means no dependency
anyone can form* — has to be written down where the next reader will find it, not left as
something the build happens to do.

**The companion test is rewritten, never silently deleted.**
`no-local-escape.test.ts:130` — *"carries no private package: nothing here imports
`@actana/shared`"* — is rewritten **with that argument written into it**, so the file that
used to forbid the import now records why the inline is safe and what would make it unsafe
again (a published `@actana/shared`, or an external rather than inlined bundle).

### D6 — The npm package keeps the name `@actana/cli`

A rename costs every existing install and buys a manifest field we can rewrite. The name
stays. What changes is the `description` at `packages/cli/package.json:4`, which currently
reads:

```
Actana Control CLI — the `actana` command's client half: the blob registry that
names Cores, and the client nouns built on @actana/sdk.
```

**"the `actana` command's client half" becomes false the day this lands — fix it.** The
`bin` map is already `{ "actana": "bin/actana.mjs" }` and does not change;
`no-local-escape.test.ts:188` asserts the bin name independently of the package name and
keeps passing.

### D7 — The Core image keeps taking the CLI from the tarball

`deploy/core.Dockerfile:205-207` continues to extract the tarball into `/opt/actana`. What
changes is that the tarball's `app/actana-cli.cjs` is now the unified bundle. **No `npm
install` at image build time** — that would tie image builds to a published version, and an
image whose contents depend on what is on the registry at build time is not reproducible
from this repository.

Once the two are the same program, the `NPM_CONFIG_PREFIX` and `PATH` ordering at
`deploy/core.Dockerfile:252-253` **stops being able to decide anything**, because installing
the published package would put the same program there. **That is what "there is nothing to
shadow" means** in the acceptance criteria: not that the collision is prevented, but that
its outcome no longer differs. `CMD ["actana", "daemon"]`
(`deploy/core.Dockerfile:266`) resolves through that same `PATH` and reaches a program that
has a `daemon` verb either way.

### D8 — The CLI gains the ability to install a Core

**Today `actana setup` downloads nothing.** It runs from an already-extracted tarball —
`packages/core/src/actana-setup.ts:1` is *"turning the extracted tarball into a running,
pairable Core"* — and `install.sh` does the resolve, download, SHA256 verification and
extraction before handing over (`install.sh:6-11`).

**That is a new capability this issue adds.** The CLI resolves the release, downloads the
tarball, checks it against the published `SHA256SUMS`, extracts it, and then does what
`setup` does today. The parts already exist and are moving with D1 —
`actana-release.ts` exports `resolveReleaseVersion`, `releaseTargetFor`,
`releaseAssetName`, `assetUrl`, `parseShasums`, `sha256OfFile` and `SHASUMS_ASSET`, and
`actana-update.ts` already composes them into fetch → verify → unpack → swap, with its
own header stating the ordering property: *"a failed update is a no-op, not a broken
Core."* The install path holds to the same property.

**The direction that matters: a Core always comes with a CLI, and a CLI may exist without a
Core and be able to install one.**

**`install.sh` survives unchanged as the no-Node door.** A bare machine has no Node, and the
tarball carries its own pinned one — so the shell script cannot be replaced by the CLI it
installs. Two doors, one implementation of the real work, **which is the argument
`install.sh`'s own header already makes for why it stays thin**: *"Every real decision —
where to install, the systemd unit, lingering, Harnesses, the pairing token — lives in that
CLI, where it is unit-tested. Anything this script grows beyond 'fetch, verify, exec'
belongs there instead."* `scripts/__tests__/install-sh.test.mjs` runs the real script
against a fixture release server and keeps that door honest.

### D9 — A locally installed Core auto-wires to that machine's CLI

**When a Core is installed on a machine, it is automatically paired with the CLI on that
machine.** No pairing token hand-carry, no blob paste. It appears in `actana core ls` and
**it is that CLI's default target.**

The two sides already meet on disk, which is what makes this a wiring job rather than a new
protocol: `packages/cli/src/blob-registry.ts:1-11` stores one blob per named Core at
`$XDG_CONFIG_HOME/actana/cores/<name>.txt` with the selection pointer at
`$XDG_CONFIG_HOME/actana/current.txt`, and its header already notes this is *"the same
directory `actana setup` writes a Core's own `registration-blob.txt` and `material.json`
into, and deliberately so"* (`packages/core/src/core-first-run.ts:48` names that file).
`actana core ls` / `core use` dispatch at `packages/cli/src/core-command.ts:112-116`.

**A CLI installed without a Core behaves as today**: it wires remote Cores, `core add` takes
a blob from a file or stdin, and nothing about that path changes.

**The `actana-sessions` skill payload gains the corresponding rule** — it can always assume
a local Core is wired on a Core machine and should address it by default. The authored
source is `.agents/skills/actana-sessions/SKILL.md`; the copies are generated by
`scripts/gen-skill-payload.mjs` into `packages/shared/src/orchestration-skill-payload.ts`
and `packages/cli/src/orchestration-skill-payload.ts`, held in step by
`packages/shared/src/__tests__/orchestration-skill-fanout.test.ts`. The lines that need it
are the target table at `orchestration-skill-payload.ts:41-42` (`actana core ls` / `core
use`) and the troubleshooting entry at `:203-204`, which today tells the agent that nothing
from `core ls` means *"the operator has not paired this machine with a Core. That is theirs
to do"* — false on a Core machine after this lands.

**This is what finally makes the skill honest on the machine the Core itself installs it
on, which is the whole complaint this issue opens with.**

### D10 — One owner for the launcher path

`packages/core/src/actana-layout.ts:33,108` resolves `binLink` to
`$HOME/.local/bin/actana`, and the container sets `NPM_CONFIG_PREFIX=/home/core/.local`
(`deploy/core.Dockerfile:252`) — **the same directory**. So both an npm shim and `setup`'s
symlink want one path.

**Whoever installed the CLI owns it.** If `setup` finds an `actana` already on `PATH` that
is not its own symlink, **it does not write one and says so plainly.** No clobber, no
silent win, no ordering trick.

**On version skew between the CLI and the Core it manages: tolerate and report.** The local
verbs read the install's own manifest (`actana-manifest.ts`) so they act on what is
actually there, and `actana status` prints both versions and says when they differ.

**No pinning.** Pinning would let a global `npm update` break a running Core — the CLI would
refuse to manage the install it is standing on, on a machine where the operator did nothing
but update a client.

### D11 — Ordering

**This lands FIRST in the `0.4.0` train** — before #280 through #287, and before #289.

## Consequences discovered while verifying this issue — these must be settled in the PR

Two enforcing lines in `no-local-escape.test.ts` will fail on the move for reasons the
decisions above do not cover. Neither may be deleted; each is rewritten **with its argument
written into it**, the same treatment D5 prescribes for the shared ban.

**C1 — The shell-out ban (`#129 D9`) blocks the move as written.**
`no-local-escape.test.ts:81-113` is a *third* ban — *"the CLI cannot shell out"*, asserting
that no shipped source imports `child_process` / `worker_threads` or reaches
`process.binding`. **`packages/core/src/actana-system.ts:12` imports `spawnSync` and `spawn`
from `node:child_process`**, and it has to: the operator verbs run `systemctl`,
`launchctl`, `loginctl` and vendor Harness installers. That is their job.

The ban's argument, in its own header, is about the **client** half: *"a CLI that shells
into a container to fetch its own credentials is not a CLI"* — a client that reaches a Core
by running a command on the Core's host. It was written when the machine-lifecycle half had
no home in this package. So the ban is **narrowed, not dropped**: it keeps sweeping the
client modules, where the temptation it names actually lives, and stops covering the
operator modules, which exist to drive an init system. The ADR carries the argument, and the
test carries it too — including what would make a breach a breach again (a *client* noun
reaching for a subprocess to get at a Core).

**C2 — The dependency-list test will fail on `selfsigned`.**
`no-local-escape.test.ts:145` asserts `Object.keys(manifest.dependencies).sort()` equals
exactly `["@actana/sdk", "undici", "ws"]`, and `:164` asserts that list and
`build.mjs`'s `external` array stay in step. A CLI that can run `setup` mints certificate
material, which reaches `selfsigned` through `core-cert-material.ts:16` (see D2). Whether
`selfsigned` is inlined into the bundle or declared and marked external, **one of those two
tests changes**, and the change carries its own line of argument rather than being absorbed
into a green suite. The test's stated purpose — *"a new name here is the first sign that"* a
server dependency, a database driver or a native addon has arrived — survives: `selfsigned`
is none of those, and `better-sqlite3` and `node-pty` still must never appear.

## LANDMINES — carry these into your working notes

These are the traps that make this ticket fail **silently** rather than loudly.

- **`actana setup` downloads nothing today.** Anyone who assumes it does will build the
  install path in the wrong place.
- **The launcher symlink and the npm global bin are the same directory in the container.**
  Test the collision, do not reason about it.
- **`no-local-escape.test.ts` will fail on both of its bans unless they are deliberately
  handled** — one by making the move complete, one by rewriting the shared ban with the D4
  argument. *(And on a third and a fourth line besides — see C1 and C2 above.)*
- **`daemon` loads `core-entry.cjs` in-process on purpose.** An extra fork leaves the init
  system supervising a wrapper that has already exited.
- **`packages/shared/package.json` exports a wildcard** (`"exports": { "./*":
  "./src/*.ts" }`), so new modules there register nothing. **That is why this issue and
  [#281](https://github.com/actana/control/issues/281) can run in parallel — do not add an
  exports map and break that.**

## Ordering in the `0.4.0` train

**This lands first.** It is the first fix of the `0.4.0` train and it lands **before** the
pairing feature issues #280–#287 and **before** #289. #284 (SDK client pairing) and #285
(`actana core pair`) should each note a dependency on this issue: #285 adds a verb to the
client CLI whose binary identity this issue changes, and shipping `actana core pair` into a
world with two rival `actana` binaries would extend the hazard rather than inherit the fix.

**#289 (`session send` has no way to await the resulting turn) also depends on this issue**,
and carries the same noted dependency: it adds behaviour to a client verb whose binary
identity this issue changes, and a wait semantics shipped against one of two rival `actana`
binaries would be honest on only one of the machines that runs it.

**#281 (shared pairing primitives) can run in parallel with this issue** rather than behind
it — `packages/shared/package.json` exports `"./*"`, so modules the two efforts add to
`packages/shared` do not collide in a manifest. See the last landmine.

Cross-references: #265 (whose §6 two-binaries constraint this supersedes) and #280 (the
pairing map this precedes).

## Acceptance criteria

- [ ] A **single binary named `actana`** serves both the operator verbs (`setup`, `status`,
      `token`, `token regenerate`, `update`, `start`, `stop`, `restart`, `logs`,
      `harnesses`, `uninstall`, `daemon`) and the client nouns (`core`, `project`,
      `session`, `events`, `harness`). One program, one help text, one version answer.
- [ ] **The Core container image ships it.** `deploy/core.Dockerfile` puts the unified CLI
      on `PATH`, and `CMD ["actana", "daemon"]` resolves to it.
- [ ] **On a container Core, a fresh Session can run `actana core ls` and `actana session …`
      with no `npm install` step.** No `@actana/cli` install, no second binary, no `PATH`
      edit.
- [ ] **The `PATH` shadowing hazard is gone because there is nothing to shadow.** The
      `NPM_CONFIG_PREFIX` / `PATH` ordering at `deploy/core.Dockerfile:252-253` can no
      longer decide which program answers `actana`, and installing the published package on
      a Core cannot displace the Core's own CLI.
- [ ] **Every verb the `actana-sessions` skill teaches exists on any machine the skill is
      installed on** — the CLI-installed case (ADR 0031 D6's pre-dispatch ensure) and the
      Core-installed case (`packages/core/src/core-entry.ts:302`) alike. A test asserts the
      skill's verb surface against the CLI's dispatch, so a skill that documents a verb the
      binary lacks fails CI.
- [ ] **`actana daemon` no longer needs a "wrong package" refusal.**
      `packages/cli/src/actana-cli.ts:163-168` is deleted rather than reworded, because the
      confusion it exists to explain cannot occur.
- [ ] **The container refusal behaviour is preserved**: on a container Core the
      lifecycle verbs still refuse and still name their Docker equivalent, and the three
      documented environment variables remain the whole operator-facing contract
      (ADR 0016 D15).
- [ ] **ADR 0025 boundaries hold.** `@actana/shared` is still private; `@actana/core`'s
      import list from `@actana/sdk` is unchanged, or any change to it carries its own
      argument in the ADR below.
- [ ] **Docs updated**: `INSTALL.md`, `packages/cli/README.md`, and every place that
      describes `actana` as one half of a split command name.
- [ ] **`CONTEXT.md` updated.** Its **`actana`** entry (`CONTEXT.md:180-181`) currently
      defines the name as *"the single command inside the Core bundle that owns a machine's
      Core lifecycle"* and lists only the operator verbs. It becomes one command that is
      both the Core manager and the client.
- [ ] **An ADR records the unification** — the decision, the packaging shape chosen and why,
      how ADR 0025's boundaries are respected under it, and an explicit note that #265 §6
      and ADR 0031 D8's "the two cannot share a module" premise is superseded. If the
      convergence removes the need for the embedded-payload generator or
      `orchestration-skill-fanout.test.ts`, the ADR says so rather than leaving the
      mechanism standing unexplained.
- [ ] **#284 and #285 carry a noted dependency on this issue**, and this issue is closed
      before either is picked up.

### Added by the settled decisions

Every criterion below is checkable by reading an enforcing line — a test, a manifest field,
a path that does or does not exist — **not by reading a PR description.**

- [ ] **D1 — the ten operator modules are in `packages/cli/src` and gone from
      `packages/core/src`**: `actana-cli.ts`, `actana-setup.ts`, `actana-layout.ts`,
      `actana-system.ts`, `actana-container.ts`, `actana-update.ts`, `actana-uninstall.ts`,
      `actana-config.ts`, `actana-manifest.ts`, `actana-release.ts`, plus
      `actana-cli-entry.ts` and the closure listed in D1
      (`actana-service`, `actana-systemd`, `actana-launchd`, `actana-tree`,
      `actana-status`, `actana-harnesses`, `operator-login-path`). Enforcing line:
      `no-local-escape.test.ts:116` is green, which is only possible if none is left behind.
- [ ] **D1 — `packages/core` is the daemon and nothing else.** No `actana-*.ts` remains in
      `packages/core/src` other than what the daemon itself runs, and
      `packages/core/build.mjs` no longer emits `dist/actana-cli.cjs`.
- [ ] **D1 — `packages/cli` emits both bundles.** The published ESM bundle and the tarball's
      CJS `actana-cli.cjs`. Enforcing line: `scripts/build-core-tarball.mjs:231`, which
      fails the tarball build unless both `core-entry.cjs` and `actana-cli.cjs` are staged.
- [ ] **D1 — the rejected alternative is recorded.** The new ADR names "leave them in
      `packages/core` and rewrite the ban as an ADR 0025 D2-style permitted-module list" and
      says it was rejected because it turns a bright line into a judgement call at every
      future review.
- [ ] **D2 — `core-material-store.ts` lives in `packages/shared/src`**, not in
      `packages/cli/src`, and `core-cert-material.ts` moved with it. Enforcing line: the
      daemon (`core-first-run.ts`, `core-entry.ts`) imports both from `@actana/shared` and
      **not** from `@actana/cli` — a daemon importing the client package is the failure this
      criterion exists to exclude. The six Panel tests that import
      `@actana/core/core-cert-material` are repointed and green.
- [ ] **D3 — `no-local-escape.test.ts:116` ("carries no daemon: nothing here imports the
      Core package") passes unchanged.** Not reworded, not scoped down, not skipped. Its
      `better-sqlite3` / `node-pty` rationale is still the comment above it.
- [ ] **D4 — `daemon` resolves `current/app/core-entry.cjs` from the install root and loads
      it in-process.** Enforcing lines: a `require`/path join, not an `import` of
      `@actana/core` (which D3's test would catch), and no `spawn`/`fork` of the daemon
      anywhere on that path. The `Type=simple` comment
      (`actana-cli-entry.ts:7-10` today) travels with the file.
- [ ] **D5 — the ADR argues the inline explicitly.** The new ADR contains the argument that
      esbuild inlining `@actana/shared` into the published bundle leaves ADR 0025 D4 intact
      *because an inlined bundle offers no surface to depend on* — no manifest, no version,
      no resolvable import path. Asserted-by-assumption is not acceptable.
- [ ] **D5 — `no-local-escape.test.ts:130` ("carries no private package") is rewritten, not
      deleted.** The rewritten test carries the D4 argument in its body and states what
      would make the inline unsafe again (a published `@actana/shared`; an external rather
      than inlined bundle). `git log` shows a rewrite, not a removal.
- [ ] **D5 — `@actana/shared` is still `"private": true` in
      `packages/shared/package.json`** and appears in no published manifest's
      `dependencies`.
- [ ] **D6 — the npm package is still named `@actana/cli`**, and
      `packages/cli/package.json`'s `description` no longer says "the `actana` command's
      client half". Enforcing line: the `description` field itself.
- [ ] **D7 — `deploy/core.Dockerfile` still extracts the CLI from the tarball
      (`:205-207`) and runs no `npm install`.** Enforcing line: grep the Dockerfile for
      `npm i` / `npm install` and find nothing that installs `@actana/cli`.
- [ ] **D7 — the tarball's `app/actana-cli.cjs` is the unified bundle**: running it inside
      the image answers both an operator verb and a client noun.
- [ ] **D8 — the CLI can install a Core end to end**: resolve the release, download the
      tarball, verify it against the published `SHA256SUMS`, extract it, then do what
      `setup` does today. Enforcing line: a test that points the CLI at a fixture release
      server and asserts a wrong checksum aborts leaving nothing installed — the same
      no-op-on-failure property `actana-update.ts`'s header already claims.
- [ ] **D8 — `install.sh` is unchanged.** Enforcing line: `git diff` on `install.sh` is
      empty, and `scripts/__tests__/install-sh.test.mjs` passes untouched.
- [ ] **D9 — installing a Core on a machine wires it to that machine's CLI automatically.**
      Enforcing line: a test that runs the install path against a fake home and then asserts
      `actana core ls` lists the local Core and that it is the selected target — no blob
      paste, no `core add`, no token typed anywhere.
- [ ] **D9 — a CLI with no local Core is unchanged**: it still wires remote Cores through
      `core add` from a file or stdin.
- [ ] **D9 — the skill payload states the local-Core rule.** Enforcing lines:
      `.agents/skills/actana-sessions/SKILL.md` carries it, both generated copies are
      regenerated, and
      `packages/shared/src/__tests__/orchestration-skill-fanout.test.ts` is green. The
      troubleshooting entry that today reads "the operator has not paired this machine with
      a Core. That is theirs to do" no longer says that unconditionally.
- [ ] **D10 — `setup` does not clobber an `actana` it did not install.** Enforcing line: a
      test in which an `actana` that is not `setup`'s own symlink already sits on `PATH`,
      asserting no symlink is written **and** that the message naming the situation is
      printed. Test the collision in the container's shape — `$HOME/.local/bin` as both
      `binLink`'s directory and `NPM_CONFIG_PREFIX`'s bin.
- [ ] **D10 — version skew is tolerated and reported.** `actana status` prints the CLI's
      version and the managed install's version and says when they differ; the local verbs
      read the install's own manifest rather than assuming their own version. Enforcing
      line: a test with a manifest version that differs from the CLI's, asserting both
      appear and that the verb still works.
- [ ] **D10 — nothing pins the CLI to the Core.** No check anywhere refuses a verb because
      the two versions differ.
- [ ] **D11 — this issue is closed before #280–#287 and #289 are picked up**, and #284,
      #285 and #289 each carry a noted dependency on it.
- [ ] **C1 — the shell-out ban is narrowed with its argument, not dropped.**
      `no-local-escape.test.ts:81-113` still fails if a *client* module reaches for
      `child_process`, `worker_threads` or `process.binding`; the operator modules are
      exempt by an exemption written into the test with the reason, and the ADR records it.
      `packages/cli/src/actana-system.ts` importing `node:child_process` is green under the
      rewritten ban.
- [ ] **C2 — whichever of `no-local-escape.test.ts:145` / `:164` changes for `selfsigned`,
      it changes deliberately.** The dependency list still excludes `better-sqlite3` and
      `node-pty`, and the comment above the assertion says why `selfsigned` is admitted (or
      why it is inlined and therefore absent).
- [ ] **No exports map is added to `packages/shared/package.json`.** It still reads
      `"exports": { "./*": "./src/*.ts" }`, so #281 can add modules there in parallel
      without a merge conflict in that file.

qcentic-dev:
Closes #288 — the first fix of the `0.4.0` train.

Two different programs installed a binary called `actana`. There is one now.

The tarball's operator CLI and npm's client CLI both put an `actana` on `PATH`
on every machine ADR 0031 installs the `actana-sessions` skill onto — and that
skill teaches `actana core ls`, `actana session start` and `actana events tail`,
every one of which was `unknown command` to the binary the Core itself put
there. Installing the npm package to fix it created a second `actana` whose
precedence was an `NPM_CONFIG_PREFIX` accident. The split was the defect.

## What is in here

- **One program.** The machine-side modules moved from `packages/core/src` into
  `packages/cli/src`; the two `runActanaCli` functions and the two entry points
  became one of each. `packages/core` has no `actana-*.ts` left and emits no
  `actana-cli.cjs`. `packages/cli` emits both bundles — the published ESM one
  and the tarball's CJS one.
- **The client nouns are checked before the container refusal table**, so
  `actana session ls` works inside the image. That is the thing that was broken.
- **`actana daemon`'s "wrong package" refusal is deleted, not reworded**, and
  the verb loads `<install root>/app/core-entry.cjs` by path through
  `createRequire`, in-process. Never an import, never a spawn.
- **`actana install`** resolves, downloads, SHA-256-verifies and unpacks a
  release, then does what `setup` does — and `setup` takes the same path when it
  is not standing in a tarball. `install.sh` is byte-for-byte unchanged.
- **A Core installed on a machine is registered with that machine's own CLI**
  and becomes its default target. No token hand-carried across one box.
- **`$HOME/.local/bin/actana` has one owner.** Setup no longer writes it blind;
  the collision with npm's global bin is tested in the container's shape,
  against real files.
- **ADR 0032** records all of it, argues the `@actana/shared` inline explicitly,
  and supersedes #265 §6 / ADR 0031 D8's two-binaries premise.

## The two blockers the issue named

**C1 — the shell-out ban is narrowed, not dropped.** Its own argument is about a
*client*: *a CLI that shells into a container to fetch its own credentials is not
a CLI.* Driving `systemctl` and `launchctl` is not that temptation. The exemption
is a table with a reason per row in
`packages/cli/src/__tests__/module-halves.ts`; `actana-cli.ts` and
`actana-cli-entry.ts` are deliberately *not* on it, a test asserts that, and a
further test asserts that exactly one machine module imports
`node:child_process`. The argument is written into the test file, not only into
the ADR.

**C2 — `selfsigned` is admitted deliberately.** A CLI that can run `setup` mints
certificate material. It is external rather than inlined, for the same reason
`ws` and `undici` are and because the tarball already ships one copy. The
comment above the assertion says so. `better-sqlite3` and `node-pty` are
asserted absent, by name.

## Where this departs from the letter of D1, and why

D1's closure was read from the operator side. The daemon reaches further:
`core-entry.ts` and `harness-install-service.ts` between them need the container
contract, the update-check paths, the release fetcher, the tarball manifest, the
Harness availability probe and the Harness installer. Every one of those went to
`packages/shared` under **D2's own rule** — both halves use it, so it belongs to
neither — rather than to `packages/cli`, because the alternative was the daemon
importing the client package, which is the arrow this issue exists to keep
straight. ADR 0032 D3 has the full table.

Two of them are named in D1's list, so they keep a module in `packages/cli/src`
that re-exports the shared definition — the arrangement `actana-release.ts`
already used for the release channel. `operator-login-path.ts` has no CLI-side
importer and simply lives in `packages/shared`.

**`nodeActanaSystem()` stayed in `packages/cli/src/actana-system.ts`**, with its
`node:child_process` import, so C1 is settled exactly where the issue said it
would have to be. Only the port's *type* is shared; the daemon passes its own
non-interactive port instead, which is not a concession — a Core under systemd
has no terminal to `confirm` on.

## Verified against a real tarball, not reasoned about

`node scripts/build-core-tarball.mjs` on this branch, extracted, and run:

```
$ .../bin/actana --version
actana 0.4.0 (core-link protocol 0.16.0)
$ .../bin/actana core ls
No Cores registered. `actana core add <name> <blob-file>` registers one.
$ ACTANA_CONTAINER=1 .../bin/actana setup
`actana setup` does not run in a container — … docker compose up -d      (exit 2)
$ ACTANA_CONTAINER=1 .../bin/actana session ls
actana session ls: no Core selected. …                                   (exit 0)
```

One binary, both halves; the lifecycle verbs still refuse and still name their
Docker equivalent; the client nouns are never refused.

`node scripts/run-unit-tests.mjs` — all six stages pass.

---

**Draft, and base `refactor/one-actana-cli`.** The per-criterion checklist is in
the first comment.


qcentic-dev:
## Acceptance criteria — all 37, each checked or unchecked with the reason

> **Updated at `bc18b74`**, after review. Four rows changed and each says how:
> criterion 3 and D7 gained the in-image enforcing line they were missing, the
> container-refusal row corrects a claim that was wrong, and D1's bundle row
> gains the build wiring that made the tarball buildable at all. CI at this
> commit: 18 success, 5 skipped, 0 failures.

### From the mandate

- [x] **A single binary named `actana`** serves the operator verbs and the client nouns. One `runActanaCli` in `packages/cli/src/actana-cli.ts` over one `ActanaCliDeps` in `cli-deps.ts`; one `USAGE` with both command blocks; one `--version`. `actana-machine-cli.test.ts`'s help/dispatch pair now checks *both* blocks against `CLIENT_NOUNS` plus the machine `switch`, in both directions.
- [x] **The Core container image ships it.** `deploy/core.Dockerfile:205-207` still extracts the tarball into `/opt/actana`, whose `app/actana-cli.cjs` is now the unified bundle, and `CMD ["actana", "daemon"]` resolves through `/opt/actana/bin` to it. Verified on a tarball built from this branch.
- [x] **On a container Core a fresh Session can run `actana core ls` and `actana session …` with no `npm install`.** The client nouns are dispatched *before* the container refusal table. **Now enforced in two places rather than one**, after review: `scripts/smoke-core-image.mjs` runs `actana core ls` inside the built image beside the refusal loop and fails on a non-zero exit, an `unknown command` or a container refusal; and `actana-machine-cli.test.ts`'s floor-and-ceiling test asserts no member of `CLIENT_NOUNS` is a key in `DOCKER_EQUIVALENT`, so the claim is bound to the refusal *table* as well as to dispatch order. **Caveat stated plainly:** the image job reports green without building on a draft PR (ADR 0023 D33), so that line does not execute on this PR — it runs the moment the image is really built. What did run here: the same commands against a tarball built from this branch with `ACTANA_CONTAINER=1` → `core ls` exits 0 with the empty-registry sentence.
- [x] **The `PATH` shadowing hazard is gone because there is nothing to shadow.** Both entries resolve to the same program, so the `NPM_CONFIG_PREFIX`/`PATH` ordering cannot decide which one answers. `actana-setup.test.ts`'s D10 block tests the collision in the container's shape — `$HOME/.local/bin` as both `binLink`'s directory and the npm prefix's bin — with real files, and asserts the npm shim is neither clobbered nor displaced.
- [x] **Every verb the skill teaches exists on any machine the skill is installed on.** New `packages/cli/src/__tests__/skill-verbs-exist.test.ts` extracts every `actana <name>` the authored `SKILL.md` shows and asserts each is a name this build dispatches — and, in the other direction, that the skill teaches all five client nouns. A skill that documents a verb the binary lacks now fails CI.
- [x] **`actana daemon` no longer needs a "wrong package" refusal.** Deleted, not reworded. `actana-cli.test.ts`'s assertion is inverted with the reason written into it, and the `daemon` case is a `case` in the dispatch.
- [x] **The container refusal behaviour is preserved.** `refusedContainerVerbs()` / `CONTAINER_USAGE` are unchanged in substance, driven by `ACTANA_CONTAINER` and never `/.dockerenv`, and the three documented variables are still the whole operator-facing contract. **Correction to what this row said before:** it claimed `install` had joined the table. It had not — review turned that up, and it was a real gap, not a wording slip: `actana install` in the image would have fetched a release and laid a second tree down beside the Core under a unit nothing supervises. `install` is on the table now for exactly `setup`'s reason, `CORE_REFUSED_VERBS` covers it so the image smoke does too, and the floor-and-ceiling test asserts it by name. Verified on the tarball: all eight refused verbs exit 2 naming their Docker command.
- [x] **ADR 0025 boundaries hold.** `@actana/shared` is still `"private": true` and in no published manifest. `@actana/core`'s import list from `@actana/sdk` is **unchanged** — nothing in this PR touches it.
- [x] **Docs updated.** `INSTALL.md` (the `npm i -g` install path, the launcher's no-clobber rule, a troubleshooting entry for it), `packages/cli/README.md` (stops calling itself a half; gains the "Running a Core on this machine" section), `CONTRIBUTING.md` (both the package list and the tree diagram), `docs/ci-cd.md` (two stale `packages/core/src/actana-*` paths), `deploy/core.Dockerfile`'s `PATH` note.
- [x] **`CONTEXT.md` updated.** The **`actana`** entry is rewritten as one command that is both the Core manager and the client, naming both verb sets, the launcher-ownership rule and the local-Core wiring, and stating that the client nouns are never refused in a container.
- [x] **An ADR records the unification.** `docs/adr/0032-one-actana-cli.md`, indexed in `docs/README.md`. It carries the decision, the packaging shape and why, how ADR 0025's boundaries are respected, the rejected alternative, and a superseded note added *in place* to ADR 0031 D8. It says explicitly why the embedded-payload generator and `orchestration-skill-fanout.test.ts` are **kept** rather than leaving them standing unexplained.
- [x] **#284 and #285 carry a noted dependency on this issue.** Posted: #284 (comment), #285 (comment), and #289 for the same reason. *This issue is closed before either is picked up* is not this PR's to do — see D11 below.

### Added by the settled decisions

- [x] **D1 — the ten operator modules are in `packages/cli/src` and gone from `packages/core/src`**, plus `actana-cli-entry.ts` and the closure. All seventeen names exist under `packages/cli/src`; `ls packages/core/src | grep -c '^actana-'` is `0`. `no-local-escape.test.ts`'s daemon ban is green, which is only possible if none is left behind. **Two named modules keep a `packages/cli/src` file that re-exports a `packages/shared` definition** — `actana-manifest.ts` and `actana-harnesses.ts` — because the daemon reads a manifest and runs the same Harness installer, and the alternative was the daemon importing `@actana/cli`. **`operator-login-path.ts` lives in `packages/shared/src` outright**, having no CLI-side importer of its own. Both departures are argued in ADR 0032 D3 and in the PR description; `actana-system.ts` is *not* one of them and keeps its `node:child_process` import in `packages/cli/src`, where C1 needs it.
- [x] **D1 — `packages/core` is the daemon and nothing else.** No `actana-*.ts` remains, and `packages/core/build.mjs` emits only `dist/core-entry.cjs`, with a comment saying where the CLI bundle went.
- [x] **D1 — `packages/cli` emits both bundles.** `dist/actana-cli.mjs` (ESM, published) and `dist-tarball/actana-cli.cjs` (CJS, the tarball's). `scripts/build-core-tarball.mjs` requires both it and `core-entry.cjs` to be staged and fails the build otherwise. **The CJS bundle is outside `dist/` on purpose**: `files` is `["dist","bin"]`, and the publish rehearsal fails on a packed file that is not the ESM bundle. **What this row was missing, and review caught:** nothing in the pipeline *built* the second bundle, so the enforcing line fired and every tarball leg was red. The pair now has one name — `build:core-tarball-bundles` — run by `pnpm core:tarball` and by both `release.yml` tarball jobs, and `scripts/__tests__/core-tarball.test.mjs` asserts every caller uses it and that none builds only the Core's half.
- [x] **D1 — the rejected alternative is recorded.** ADR 0032 D2 names "leave them in `packages/core` and rewrite the ban as an ADR 0025 D2-style permitted-module list" and says it was rejected because it turns a bright line into a judgement call at every future review.
- [x] **D2 — `core-material-store.ts` lives in `packages/shared/src`**, with `core-cert-material.ts`. The daemon (`core-first-run.ts`, `core-entry.ts`) imports both from `@actana/shared`, never from `@actana/cli`. The seven Panel tests and two SDK tests that imported `@actana/core/core-cert-material` are repointed and green; their unit tests moved to `packages/shared/src/__tests__` and are green.
- [x] **D3 — `no-local-escape.test.ts`'s daemon ban passes unchanged.** The `it("carries no daemon: nothing here imports the Core package")` title and assertion are byte-identical, and the `better-sqlite3` / `node-pty` rationale is still the first paragraph of the comment above it. What was added: further paragraphs saying why passing it is the proof, and a *second* test that the string `@actana/core` appears in no shipped module's code. The enclosing `describe` was renamed from "the published CLI is the client half" to "the published CLI carries no daemon", because "client half" became false; the ban itself is untouched.
- [x] **D4 — `daemon` resolves `current/app/core-entry.cjs` from the install root and loads it in-process.** `actana-cli-entry.ts`'s `daemonEntryPath()` tries `ACTANA_ROOT` then `layout.currentLink`, and `createRequire(entry)(entry)` loads it — a path join and a require, never an import, never a `spawn` or `fork`. The `Type=simple` comment travelled with the file and stands at the top of it.
- [x] **D5 — the ADR argues the inline explicitly.** ADR 0032 D5: an inlined bundle offers no surface to depend on — no manifest, no version, no resolvable import path — so both halves of ADR 0025 D4's "not deleted and not published" stay true. It also names the two changes that would make it unsafe again.
- [x] **D5 — the "carries no private package" test is rewritten, not deleted.** `git log --follow` on the file shows a modification. The rewritten block carries the D4 argument in its body and asserts both unsafe-again conditions: `@actana/shared` is still `"private": true`, appears in no published `dependencies`, and is never in `build.mjs`'s `external`. A fourth test asserts something still imports it, so the three cannot pass by saying nothing.
- [x] **D5 — `@actana/shared` is still `"private": true`** and in no published manifest's `dependencies`. Asserted by the test above.
- [x] **D6 — the npm package is still named `@actana/cli`**, and its `description` no longer says "the `actana` command's client half". The bin-name test asserts both, by field.
- [x] **D7 — the Dockerfile still extracts the CLI from the tarball and runs no `npm install` of it.** The only `@actana/cli` in the file is in a comment explaining why the `PATH` ordering no longer decides anything. The `npm install -g` at `:134` is npm itself, checksum-pinned, and predates this.
- [x] **D7 — the tarball's `app/actana-cli.cjs` is the unified bundle.** Built and run: `--version` answers `actana 0.4.0 (core-link protocol 0.16.0)` and `core ls` answers `No Cores registered.`, from the same file. **The tarball now actually builds in CI** — both Smoke — Core release tarball jobs pass at this commit, where before they failed on the missing bundle. The in-image half of this criterion is `smoke-core-image.mjs`'s new `core ls` + `--version` pair, with the draft-mode caveat noted under criterion 3.
- [x] **D8 — the CLI can install a Core end to end.** `actana install` (and `setup` when not standing in a tarball) resolves, downloads, verifies against `SHA256SUMS`, unpacks and sets up. `actana-install.test.ts` points it at the fixture release server and asserts a wrong checksum aborts with **nothing installed** — no `versions/`, no `current`, no launcher, no registry entry.
- [x] **D8 — `install.sh` is unchanged.** `git diff origin/refactor/one-actana-cli -- install.sh` is empty, and `scripts/__tests__/install-sh.test.mjs` passes untouched.
- [x] **D9 — installing a Core wires it to that machine's CLI automatically.** `actana-install.test.ts` runs the install path against a fake home and asserts `core ls` lists the local Core and that it is the selected target — no blob paste, no `core add`, no token typed. It also asserts the stored blob is mode 0600 and that the blob itself never reaches an output sink on that path.
- [x] **D9 — a CLI with no local Core is unchanged.** Nothing in `core-command.ts` or `blob-registry.ts` changed; `core add` from a file or stdin is covered by the existing suites, all green. The wiring is a *write* into the same registry those verbs already read.
- [x] **D9 — the skill payload states the local-Core rule.** `.agents/skills/actana-sessions/SKILL.md` carries it in the orientation section, both generated copies are regenerated by `scripts/gen-skill-payload.mjs`, and `orchestration-skill-fanout.test.ts` is green. The troubleshooting entry no longer says "the operator has not paired this machine with a Core" unconditionally — it now says a machine running a Core registers it automatically, and `skill-verbs-exist.test.ts` asserts both halves of that edit.
- [x] **D10 — `setup` does not clobber an `actana` it did not install.** `actana-launcher.ts` decides ownership by where the link points. Two tests plant a real npm-style shim at `$HOME/.local/bin/actana` — the container's exact shape — and assert no symlink is written *and* that the message naming the situation is printed and names both paths. A third covers an `actana` that is only earlier on `PATH`. A fourth asserts the ordinary case still links.
- [x] **D10 — version skew is tolerated and reported.** `actana status` gains a `CLI version` row that appears only when the two differ; `--version` answers a second line for the same case. Tests: a report with a differing manifest asserts both versions appear, that `summarizeHealth` is still `healthy`, that the `Version` row is the *install's*, and that `status`, `token`, `start` and `logs` all still exit 0 against it.
- [x] **D10 — nothing pins the CLI to the Core.** There is no comparison anywhere that gates a verb; the "refuses nothing over the difference" test is what says so, by running four verbs against a skewed install and asserting no error mentions a version.
- [ ] **D11 — this issue is closed before #280–#287 and #289 are picked up.** **Unchecked, and not this PR's to check.** A pull request cannot close the issue it implements while it is a draft against a train branch, and the ordering claim is about what happens after it merges. The half that *is* actionable is done: #284, #285 and #289 each carry a noted dependency on #288 (linked above).
- [x] **C1 — the shell-out ban is narrowed with its argument, not dropped.** It still fails if a *client* module imports `child_process` / `worker_threads` or reaches `process.binding`. The exemption is a table with a reason per row in `packages/cli/src/__tests__/module-halves.ts`, whose header carries the argument in full — including what makes a breach a breach again. `actana-cli.ts` and `actana-cli-entry.ts` are deliberately not exempt and a test asserts they never become so; another asserts every row is a real shipped module with a reason; another asserts that **exactly one** machine module imports `node:child_process`, and it is `actana-system.ts`. ADR 0032's consequences record it. ADR 0026's no-scheduling ban is narrowed by the same table for the same reason, and the "reads the clock in exactly one place" test still passes with `actana-cli-entry.ts` as the only reader.
- [x] **C2 — the dependency-list change is deliberate.** `:145`'s list is now `["@actana/sdk", "selfsigned", "undici", "ws"]`, and the comment above it says why `selfsigned` is admitted: `setup` mints certificate material, and it is none of the three things the list exists to keep out — no server, no database driver, no native addon. It is external rather than inlined, so `:164` stays in step automatically (and now reads *both* bundles' `external` arrays). `better-sqlite3` and `node-pty` are asserted absent by name.
- [x] **No exports map is added to `packages/shared/package.json`.** It still reads `"exports": { "./*": "./src/*.ts" }`, untouched. The only change to that manifest is `selfsigned` in `dependencies` and a reworded `description`. Nothing in this PR creates, edits or imports `packages/shared/src/pairing-code.ts` or `pairing-session.ts` (#281's).

---

### Verification run on this branch

`node scripts/run-unit-tests.mjs` — all six stages pass (root 19, sdk 21, shared 23, cli 48, core 47, panel 142 files). On CI at `bc18b74`: 18 checks success, 5 skipped, 0 failures — including both tarball smokes and both installer E2Es, which were red or skipping before.

`node scripts/build-core-tarball.mjs` produces `actana-core-0.4.0-linux-x64.tar.gz`; the extracted `bin/actana` answers an operator verb and a client noun from one file, refuses `setup` under `ACTANA_CONTAINER=1` naming `docker compose up -d`, and answers `session ls` / `harness ls` / `events tail` / `project ls` there without refusing.

One post-review note on scope: `actana install` joining the container refusal table is a behaviour change, not test-only, and it is described in the row above rather than folded into a test commit.

One unrelated fix rides along and is called out rather than buried: `scripts/__tests__/core-launcher.test.mjs` inherited `AC_APP_PATH` from the environment, so it failed for anyone running the suite *inside* a container Core — which is the machine this branch was built on. It now strips that variable and `ACTANA_ROOT` before running the launcher.

mrfqcentic (changes requested):
## Review against #288 — 30 of 37 criteria met, 6 partial, 1 not met

Reviewed by reading the enforcing lines at `1a5acd6`, not the description or the
checklist comment. Base is `refactor/one-actana-cli` (`75d61ab`) — correct, not
`main` and not a beta train. No file matching `pairing-code.ts` or
`pairing-session.ts` appears in the 135 changed files, so #281's surface is
untouched.

**The hard parts are done, and done well.** C1, C2, D3, D4, D5, D8, D9, D10 and
the shared-manifest landmine all check out against source:

- **C1** — `module-halves.ts:54-77` is a table with a reason per row, not a
  prefix rule; `no-local-escape.test.ts:91-92` asserts `actana-cli.ts` and
  `actana-cli-entry.ts` are *not* on it, `:80-93` fails a row for a deleted file
  or a row with no reason, and `:121-134` pins the machine half's subprocess to
  exactly `["actana-system.ts"]`. Narrowed with the argument written in, not
  deleted.
- **C2** — `no-local-escape.test.ts:263-300` admits `selfsigned` with sixteen
  lines of argument above the assertion and re-asserts `better-sqlite3` and
  `node-pty` by name. Deliberate.
- **`packages/shared/package.json`** still carries only `"exports": { "./*":
  "./src/*.ts" }` — no map added, so #281 stays parallel-safe.
- **D4** — `actana-cli-entry.ts:184-192`: `createRequire(entry)(entry)` on
  `<root>/app/core-entry.cjs`, in-process, and the `Type=simple` paragraph
  travelled with the file (`:20-28`). No `spawn`/`fork` anywhere on that path.
- **D10** — the launcher/npm-bin collision is a real-file test in the
  container's own shape (`actana-setup.test.ts`, "the launcher path has one
  owner"), asserting the npm shim is neither repointed nor deleted and that the
  message names both paths. Tested, not argued.

CI confirms the enforcing lines are green where it counts: `no-local-escape`,
`skill-verbs-exist`, `actana-install`, `actana-setup` and the shared
fanout test all pass (47/49 files in `packages/cli`, 23/23 in
`packages/shared`, 47/47 in `packages/core`; Typecheck, Lint, Conventions,
Dependency Audit all green).

**Requesting changes on delivery, not on design.** Three things, one of which
breaks the release leg.

---

### 1. `package.json:54` — the tarball cannot be built on this branch (blocking)

```
"core:tarball": "pnpm --filter @actana/core build && node scripts/build-core-tarball.mjs",
```

`packages/cli/build.mjs` now emits the CJS half into `packages/cli/dist-tarball`,
and `scripts/build-core-tarball.mjs:237-247` requires it — correctly, that is the
enforcing line D1 names. But nothing in the pipeline builds `@actana/cli` before
that check, so both tarball smokes fail on this PR head:

```
[core-tarball] no actana-cli.cjs at .../packages/cli/dist-tarball — run `pnpm --filter @actana/cli build` first
```

The same omission is in the release workflow, in both tarball jobs:
`.github/workflows/release.yml:407` and `.github/workflows/release.yml:485`
("Build the Core bundle" → `pnpm --filter @actana/core build`). `ci.yml:690`
inherits it through `pnpm core:tarball`.

This is what makes criterion **D7 — "the tarball's `app/actana-cli.cjs` is the
unified bundle"** not met: there is no tarball. It also leaves the two
image-side criteria unproven rather than wrong.

### 2. `packages/cli/src/__tests__/actana-update.test.ts:218` — Unit Tests are red (blocking)

```
AssertionError: expected [ 'actana-update-work-vCJnpk' ] to deeply equal []
```

`actana-update.ts:114-132` does remove its work dir in a `finally`, so the leak
is not in the code under test — the assertion scans the *global* `os.tmpdir()`
for a property that is per-invocation, and `actana-machine-cli.test.ts:611-663`
exercises `actana update` from a second file in the same package. Moving both
suites from `packages/core` into a 49-file `packages/cli` changed the
scheduling, and the race now fires. Scope the assertion (a per-test `TMPDIR`, or
compare against the entries that existed before the call) rather than widening
it.

### 3. `scripts/smoke-core-image.mjs:432-440` — nothing runs a client noun inside the image

The refusal loop proves the machine verbs still refuse and name their Docker
equivalent, and that is the half that was already true. The half this issue
exists for — *"a fresh Session can run `actana core ls` and `actana session …`
with no `npm install`"* (criterion 3), and D7's *"running it inside the image
answers both an operator verb and a client noun"* — has no enforcing line
anywhere. `actana-cli.ts:1240-1291` does check `CLIENT_NOUNS` before the
container refusal table, so it is right today; it is right by dispatch ordering,
which is exactly the kind of thing a later refactor reorders silently. The
harness is already open at line 432 — one `core.exec(["actana", "core", "ls"])`
next to the refusal loop closes it.

---

### Two non-blocking notes

**`packages/cli/src/actana-cli-entry.ts:38`** — the narrowed shell-out ban
sweeps import specifiers in `packages/cli/src` only, and `actana-cli-entry.ts`
is deliberately left in the swept set so the machine half gets no door onto the
client's path. It now imports `@actana/shared/harness-availability-store`, which
reaches `spawnSync` through `shell-env.ts:399` and `harness-cli-version.ts:114`.
Legitimate — `status` and `setup` need the Core's own probe — but before #288
`packages/cli` could not import `@actana/shared` at all, so this is a new
indirect route the ban cannot see. Worth naming in
`no-local-escape.test.ts`'s "what would make a breach a breach again" so the
next reader knows the sweep stops at the package boundary.

**`packages/cli/src/local-core-wiring.ts:52`** — the header says a label of
`web 01` becomes `web01`; the substitution at `:58` replaces the run with `-`,
so it becomes `web-01`, which is what `actana-install.test.ts` asserts. Comment,
not code.

**`packages/cli/src/__tests__/actana-machine-cli.test.ts:1234`** — the floor and
ceiling on `refusedContainerVerbs()` assert `status`, `token` and `harnesses`
are not refused. Adding `for (const noun of CLIENT_NOUNS) expect(REFUSED).not.toContain(noun)`
there would bind the criterion-3 claim to the refusal table as well as to the
dispatch order.

---

Once the tarball build wiring lands and the update test is scoped, this is an
approve. The design work — the ADR, the two rewritten bans, the module-halves
table, the D8/D9/D10 tests — is the strongest part of it and needs nothing.

qcentic-dev:
@mrfqcentic — thanks for reviewing the enforcing lines rather than the description. All three blocking findings and all three notes are fixed at `bc18b74`. CI is green: 18 success, 5 skipped, 0 failures.

## The three blocking findings

**1. `package.json:54` + `release.yml:407`/`:485` — the tarball could not be built.** Fixed in `5ec2fb0`. You are right that fixing them one at a time would leave three places to keep in step, so the pair has one name — `build:core-tarball-bundles` — and all three callers run it. `scripts/__tests__/core-tarball.test.mjs` is the new enforcing line: it asserts the script names both packages, that `core:tarball` runs it *before* the builder (by index, not by substring), and that every workflow calling the builder directly runs it first and no longer builds only the Core's half. It carries a guard on the guard, so renaming the builder fails the sweep instead of emptying it.

Verified from a clean tree — `rm -rf packages/*/dist* && pnpm core:tarball` writes the tarball — and both **Smoke — Core release tarball** jobs now pass (linux-x64 45s, linux-arm64 36s). The two **E2E — installer** jobs, which were skipping behind the tarball, now run and pass too.

**2. `actana-update.test.ts:218` — Unit Tests red.** Fixed in `ccb1e9d`, narrowed rather than widened, using your second suggestion: the temp root is listed either side of the call and the assertion is that *this call* added nothing. Your diagnosis was exactly right — the code under test removes its work dir in a `finally` and is unchanged; the assertion was reading `actana-machine-cli.test.ts`'s in-flight state, and moving both suites into a 49-file `packages/cli` is what made them overlap. A work dir this invocation really did leak still fails it.

**3. `smoke-core-image.mjs:432-440` — nothing ran a client noun in the image.** Fixed in `961e3c6`. `actana core ls` now runs beside the refusal loop and fails on a non-zero exit, on `unknown command`, or on a container refusal; `actana --version` follows it so D7's *"answers both an operator verb and a client noun"* is asserted of the same binary.

**Your finding turned up a real bug next to it, and it is fixed in the same commit.** Running that check by hand against a built tarball showed `actana install` was **not** on `DOCKER_EQUIVALENT` — so in the image it would have downloaded a release and laid a second tree down beside the Core, under a unit nothing in the container supervises. That is the "half-working setup" `actana-container.ts`'s own header exists to refuse, and `install` refuses for exactly `setup`'s reason. `CORE_REFUSED_VERBS` gains it so the smoke covers it, and the list-drift test held the two in step. My earlier checklist claimed `install` was already on the table; it was not, and the checklist comment is corrected.

## The three notes — all taken, all one line of intent

- **`no-local-escape.test.ts`** — the "what would make a breach a breach again" paragraph now says the sweep stops at the package boundary and names the route: `actana-cli-entry.ts` → `@actana/shared/harness-availability-store` → `shell-env.ts` / `harness-cli-version.ts`. It also says what enforcing it would take, so it is a known gap rather than an unknown one.
- **`local-core-wiring.ts:52`** — the comment said `web01`; the code produces `web-01`, which is what the test asserts. Comment fixed.
- **`actana-machine-cli.test.ts:1234`** — added, and it earns its keep immediately: `expect(REFUSED).toContain("install")` beside `for (const noun of CLIENT_NOUNS) expect(REFUSED).not.toContain(noun)`. Criterion 3 is now bound to the refusal *table* as well as to the dispatch order.

## What is deferred, and why

**The in-image assertion does not execute on this PR.** `Core image / Build + smoke (amd64)` reports green without building or smoking, because the PR is a draft — ADR 0023 D33, "reported green rather than skipped, because a skipped required check stays Pending forever". I have been asked not to mark this ready for review, so I cannot make that job do real work here. The line exists and will run the moment the image is actually built.

What I could do instead, I did: the same commands were run against a tarball built from this branch with `ACTANA_CONTAINER=1`, which is the exact code path the image takes. All eight refused verbs exit 2 naming their Docker command; `core ls` exits 0 with the empty-registry sentence; `--version` answers `actana 0.4.0 (core-link protocol 0.16.0)` from the same binary. That is evidence, not proof-by-CI, and I would rather say so than imply the smoke ran.

**D11 stays unchecked**, unchanged from before: a draft PR cannot close the issue it implements, and the ordering claim is about what happens after merge. The actionable half — the noted dependency on #284, #285 and #289 — is done.

Nothing about the design moved. The ADR, the two rewritten bans, the module-halves table and the D8/D9/D10 tests are as you reviewed them.

mrfqcentic (approved):
## Approving — all three blocking findings settled at the enforcing line

Re-reviewed at `bc18b74` against the four new commits, reading the enforcing
lines rather than the reply or the checklist. Base is still
`refactor/one-actana-cli`; `packages/shared/src/pairing-code.ts` and
`pairing-session.ts` remain untouched.

### 1. The tarball — settled

`package.json:55-56` gives the pair one name and orders it:

```
"build:core-tarball-bundles": "pnpm --filter @actana/core build && pnpm --filter @actana/cli build",
"core:tarball": "pnpm build:core-tarball-bundles && node scripts/build-core-tarball.mjs",
```

`release.yml:414` and `:498` both run it, and both now precede the builder at
`:421` / `:504`. I swept the workflow directory independently: `release.yml` is
the only direct caller of `build-core-tarball.mjs`, and it no longer contains
`pnpm --filter @actana/core build` anywhere.

Naming the pair rather than fixing three call sites is the right call, and
`scripts/__tests__/core-tarball.test.mjs:359-421` is the enforcing line I would
have asked for: it checks the script names both packages, checks the *order* by
index rather than substring, sweeps every workflow that calls the builder for
`pnpm build:core-tarball-bundles`, asserts none of them still builds only the
Core's half, and guards the guard with `expect(checked).toBeGreaterThan(0)` so a
rename of the builder fails instead of emptying the sweep.

Proof it actually produces the file, from the linux-x64 smoke log:
`dist-tarball/actana-cli.cjs 412.5kb` → `[core-tarball] bundled 30 runtime
package(s)` (including `undici`) → `[tarball-smoke] OK — extracted tarball boots
a dialable Core with no system Node`. Both tarball smokes pass (45s / 36s), and
both **E2E — installer** jobs, which were skipping behind the tarball, now run
and pass.

### 2. `actana-update.test.ts:218` — settled, and tightened rather than widened

The temp root is listed either side of the call and the assertion is that *this
call* added nothing:

```
const before = new Set(workDirs());
await update();
expect(workDirs().filter((name) => !before.has(name))).toEqual([]);
```

A work dir this invocation really leaked is created during the call, so it is
absent from `before` and still fails the test — the property is preserved, not
relaxed. The comment records why the global spelling was wrong.

*Non-blocking:* the race is narrowed, not closed — a concurrent
`actana-machine-cli.test.ts` update that starts *and* is still in flight inside
this call's window would still trip it. Pointing `TMPDIR` at a per-test root
would close it entirely, if it ever flakes again.

### 3. `smoke-core-image.mjs:442-478` — the line exists, beside the refusal loop

`actana core ls` runs in the image and fails on a non-zero exit, on
`unknown command`, and on a container refusal; `actana --version` follows it so
D7's *"answers both an operator verb and a client noun"* is asserted of the same
binary. That is exactly what was missing.

**It does not execute on this PR, and you said so rather than implying it ran.**
I confirmed that independently in the job log for `Core image / Build + smoke
(amd64)`: `##[notice] Nothing to build for this pull request (draft, or a
documentation-only diff). Reported green rather than skipped, because a skipped
required check stays Pending forever (ADR 0023 D33)`. Since I asked you not to
mark this ready, that is my constraint and not a defect in the change. I am
accepting it because the CI-executing half now exists too — see below — and
because the line will run on the first real image build.

### The bug your fix turned up

`actana install` was genuinely missing from `DOCKER_EQUIVALENT` at `1a5acd6`,
and `packages/cli/src/actana-container.ts:70-80` adds it now with `setup`'s
reasoning. This is worth naming plainly: `CONTEXT.md:182` and
`core.Dockerfile:242` both *documented* `install` as refused while the table did
not refuse it, so my own "criterion 7 — met" last round rested on a claim that
was false in code. It is true now. `scripts/lib/panel-image.mjs:98` adds it to
`CORE_REFUSED_VERBS`, and the pre-existing drift test
`scripts/__tests__/panel-image.test.mjs:910-916` — `expect([...CORE_REFUSED_VERBS]).toEqual(verbs)`
against the table's own keys — is what forced the two to stay in step. Your
correction to that checklist row is accurate; I checked it against the diff, not
against the row.

### The three notes — all taken

- **`no-local-escape.test.ts:43-56`** — names the route
  (`actana-cli-entry.ts` → `@actana/shared/harness-availability-store` →
  `shell-env.ts` / `harness-cli-version.ts`), says it is new since #288, says
  what a *client* noun growing one would mean, and says what enforcing it would
  take. A known gap rather than an unknown one, which was the ask.
- **`local-core-wiring.ts:47-54`** — `web 01` → `web-01`, matching the code and
  the test.
- **`actana-machine-cli.test.ts:1247-1255`** — `for (const noun of CLIENT_NOUNS)
  expect(REFUSED).not.toContain(noun)`, plus `expect(REFUSED).toContain("install")`.
  This is the part that makes criterion 3 hold *in CI today*, independent of the
  image job: the claim is bound to the refusal table, not only to dispatch order.

### C1 and C2 still settled

Neither moved. `module-halves.ts` is untouched by all four commits, so
`MACHINE_MODULES`, the non-exemption of `actana-cli.ts` / `actana-cli-entry.ts`
and the one-module-subprocess assertion stand as reviewed. `packages/cli/package.json`
and `packages/shared/package.json` are untouched, so the dependency list is still
exactly `["@actana/sdk", "selfsigned", "undici", "ws"]` with `better-sqlite3` and
`node-pty` barred by name, and `packages/shared` still has no exports map. The
only edit to `no-local-escape.test.ts` is the boundary paragraph, which adds a
caveat and removes no assertion.

### Criteria

Re-checked only where the new commits reach. **2, 3, 15 and 25 move to met**;
34 of 37 now met, 3 partial, none unmet. The three partials are structural, not
defects: **12** and **34** are the D11 ordering pair, which a draft PR cannot
satisfy because it cannot close the issue it implements — leaving them unchecked
is the right call, and the actionable half (the noted dependency on #284, #285,
#289) was already done; **13** is the argued `packages/shared` placement of part
of D1's closure, recorded in ADR 0032 D3.

CI at `bc18b74`: 18 success, 5 skipping, 0 failing. All six unit stages pass
(cli 48 passed / 1 skipped).

Nothing further from me. Mark it ready and land it.

PR #294, opened 2026-08-20
refactor: one actana CLI

qcentic-dev:
This branch carries the issue 288 squash commit `7ccba9a`.

`refactor/one-actana-cli` is exactly one commit ahead of `beta/0.4.0`:

- `7ccba9a` — refactor(cli): there is one `actana` — the Core manager and the client, one program (#293)

Issue: actana/control#288 — https://github.com/actana/control/issues/288


mrfqcentic (dismissed):
Reviewed against #288, criterion by criterion, from the enforcing lines in `7ccba9a` (base `beta/0.4.0`, exactly one commit ahead as claimed). **Approving** — every acceptance criterion is met. Verdicts below; the three gates I was asked to prove from code rather than the description first.

## The three gates, proven from enforcing lines

**Tarball staging.** `scripts/build-core-tarball.mjs` (`stagedBundles`, ~L231) requires both `core-entry.cjs` from `packages/core/dist` **and** `actana-cli.cjs` from `packages/cli/dist-tarball`, calling `fail()` ("no ${file} at ${dist}") when either is missing — the build cannot silently ship a tarball whose launcher execs nothing. `packages/cli/build.mjs` emits both bundles (ESM → `dist/actana-cli.mjs`, CJS → `dist-tarball/actana-cli.cjs`, the latter outside `dist/` so the npm package does not publish a second copy); `packages/core/build.mjs` emits only the daemon. CI wiring is guarded by the named `build:core-tarball-bundles` script in `package.json`, both `release.yml` tarball jobs use it, and `scripts/__tests__/core-tarball.test.mjs` asserts every workflow calling the builder runs it first.

**Narrowed no-local-escape ban.** `packages/cli/src/__tests__/no-local-escape.test.ts`: the shell-out ban (#129 D9) is narrowed, not dropped — it sweeps `clientSources()` only, with the machine-half exemption enumerated as a table with a reason per row in `__tests__/module-halves.ts`; guard-on-guard tests assert the sweep is non-vacuous, every exempt module is real, and `actana-cli.ts`/`actana-cli-entry.ts` are NOT exempt. A further test asserts exactly one machine module (`actana-system.ts`) imports `node:child_process`. The daemon ban ("carries no daemon: nothing here imports the Core package") is unchanged, rationale comment intact, and passing — which is only possible if the move was complete. The shared ban is rewritten in place ("the private package is inlined, not depended on (#288 D5)") carrying the D5 argument and both unsafety conditions, asserting `private: true`, absence from published deps, and non-externality — not deleted. C2: the dependency-list test deliberately admits `selfsigned` with its argument and still excludes `better-sqlite3`/`node-pty`, with the externals-in-step check now reading both bundles.

**Container-aware install.** `packages/cli/src/actana-container.ts`: `install` is in `DOCKER_EQUIVALENT` for exactly `setup`'s reason ("this image is the install…"), so it refuses naming `docker compose up -d`. `actana-machine-cli.test.ts` binds it from both ends — `refusedContainerVerbs()` contains `install` and contains no `CLIENT_NOUNS` — and `scripts/lib/panel-image.mjs`'s `CORE_REFUSED_VERBS` includes it so the image smoke covers it. `scripts/smoke-core-image.mjs` additionally runs `actana core ls` inside the built image (dying on non-zero exit, `unknown command`, or a container refusal) and then `actana --version` from the same binary — D7's "answers both an operator verb and a client noun" proven on the built image, not argued.

1. **One binary, both verb sets** — MET. One `runActanaCli` over one dep bag dispatches `CLIENT_NOUNS` then machine verbs, one USAGE, one `versionLines`; "unknown command" message unified.
2. **Image ships it, CMD resolves** — MET. Tarball still extracted to `/opt/actana`; `CMD ["actana","daemon"]` unchanged and smoke-asserted on the built image config.
3. **`core ls`/`session …` in-container, no npm install** — MET by the smoke leg above plus dispatch order (nouns before refusal table), the latter also bound by test.
4. **Nothing to shadow** — MET. Both doors produce the same program; `NPM_CONFIG_PREFIX` ordering no longer decides anything (Dockerfile comment, ADR 0032 D7); launcher no-clobber covers the residue.
5. **Skill verbs exist wherever the skill lands** — MET. `skill-verbs-exist.test.ts` checks the authored SKILL.md against the dispatch both directions; both install paths (core-entry boot ensure; CLI pre-dispatch ensure) intact.
6. **Wrong-package `daemon` refusal deleted** — MET, with the deletion documented in the dispatch comment and the inverted test in `actana-cli.test.ts`.
7. **Container refusal preserved; three-var contract** — MET. Table intact (plus `install`); `ACTANA_CONTAINER` detection, never `/.dockerenv`; contract still `ACTANA_PUBLIC_HOST`/`ACTANA_PORT`/`ACTANA_LABEL`.
8. **ADR 0025 boundaries** — MET. `@actana/shared` still `private: true`, wildcard exports untouched, in no published manifest; `@actana/core` still imports only `core-link-frames` and `core-files-error-codes` from the SDK.
9. **Docs** — MET. INSTALL.md (npm path, no-clobber rule, troubleshooting), CLI README (no longer "client half"), CONTEXT.md (`actana` entry rewritten), CONTRIBUTING.md, Dockerfile note, docs/README ADR index.
10. **CONTEXT.md** — MET (covered above).
11. **ADR** — MET. ADR 0032 carries the decision, packaging shape, 0025-boundary analysis, the explicit D5 inline argument with both unsafety conditions, the rejected permitted-module-list alternative ("bright line → judgement call"), the superseded #265 §6 / 0031 D8 premise, and why the payload generator is kept anyway.
12. **#284/#285/#289 dependencies** — MET. All three issues carry "Depends on #288".

**Settled decisions (D1–D11, C1, C2, exports)** — all MET: no `actana-*.ts` remains in `packages/core/src`; daemon imports the shared modules from `@actana/shared` (never `@actana/cli`); all seven Panel test files repointed to `@actana/shared/core-cert-material`; `daemon` loads `<install root>/app/core-entry.cjs` via `createRequire`, in-process, `Type=simple` comment carried; `install.sh` diff is zero; D8's wrong-checksum test asserts nothing installed; D9 wiring test asserts listing + selection with no blob paste, and preserves an existing selection; D10's collision is tested in the container's exact shape (`binDir` = npm prefix bin, planted npm shim) and skew is printed-not-enforced in both `status` and `--version`; #281's wildcard exports untouched.

## One non-blocking nit

Two comments say `packages/cli/dist` where the code stages from `packages/cli/dist-tarball`: `packages/core/build.mjs:36` and `docs/adr/0032-one-actana-cli.md:157`. Every enforcing line (the `stagedBundles` table, the workflow comments, `packages/cli/build.mjs`'s header) has it right, so this is prose-only — worth a one-word fix on the next train, not a re-spin of this one.

Solid work — the tests argue their own criteria, which is exactly what the issue asked for.

qcentic-adm (changes requested):
## Review of #294 against #288 — feature-branch gate (`beta/0.4.0` ← `refactor/one-actana-cli`, one commit `7ccba9a`)

Checked from the enforcing lines, not the description. The move, the build wiring, the narrowed bans, `install`, the launcher rule and the skew reporting are all real and well-tested — 30 of the issue's ~35 criteria are met outright (checklist at the end). **One gap blocks the gate**, and a handful of records the PR introduces are born stale. Details with file and line:

### Blocking

**1. A container Core is never registered with its own CLI, so D9 and criterion 3 do not hold on the machine the issue opens with — and the new skill text is false there.**

- `wireLocalCore` has exactly one caller: `packages/cli/src/actana-setup.ts:315`. In the image, `setup` and `install` are refused (`packages/cli/src/actana-container.ts:85-96`), nothing in `packages/core/src` or `packages/shared/src` writes `cores/<name>.txt` or `current.txt`, and the CLI has no fallback to the container Core's own `registration-blob.txt` (`packages/core/src/core-first-run.ts:191`).
- `scripts/smoke-core-image.mjs:457-460` says so explicitly: *"Exit 0 with the empty-registry sentence is what a fresh Session sees."* So on a container Core, `actana core ls` is empty and `actana session …` answers `no Core registered`. Criterion 3 — *"a fresh Session can run `actana core ls` **and `actana session …`**"* — is met for the first verb only.
- Meanwhile `.agents/skills/actana-sessions/SKILL.md:32-36` now states unconditionally: *"On a machine that is itself a Core, that Core is already registered and already selected … do not ask the operator to pair the machine you are already standing on,"* and `:189-194` tells the agent an empty `core ls` *"means there is no local Core either."* `packages/core/src/core-entry.ts:305` installs that skill onto the container Core at boot. That is the issue's opening complaint — *"the skill is dishonest on the machine the Core itself put it on"* — re-created in a new form.

Either the container Core registers itself (e.g. first-run writes the blob into the registry under `$XDG_CONFIG_HOME/actana/cores/` and points `current.txt` at it, which D9 already argues is the same directory), or the CLI's registry resolves a local Core from `AC_CORE_MATERIAL_FILE` / `registration-blob.txt` in container mode — and the image smoke should then assert `core ls` lists it rather than accepting the empty sentence. ADR 0032 D9 (`docs/adr/0032-one-actana-cli.md:183-197`) should say which.

### Should be fixed in the same round

**2. `deploy/core.Dockerfile` does not set `ACTANA_ROOT`, so `CMD ["actana", "daemon"]` can still be displaced.** `daemonEntryPath()` (`packages/cli/src/actana-cli-entry.ts:93-106`) tries `ACTANA_ROOT`, then the layout's `current` symlink; neither exists for an npm-installed CLI in the image, whose shim lands first on `PATH` (`deploy/core.Dockerfile:261-262`). An operator who runs `npm i -g @actana/cli` inside the container gets *"no Core is installed here"* on the next start instead of a daemon. The claim at `deploy/core.Dockerfile:248-249` and `docs/adr/0032-one-actana-cli.md:162-166` ("stops being able to decide anything") is not yet true for `daemon`. One line fixes it — `ENV ACTANA_ROOT=/opt/actana` next to `AC_APP_PATH` at `:258`, or a fallback to `path.dirname(AC_APP_PATH)` in `daemonEntryPath()`.

**3. `CONTRIBUTING.md:56`** still maps `packages/core/src/actana-cli.ts — the actana CLI (setup, status, token)`. The file no longer exists; `packages/core` is "the daemon and nothing else" (D1). The commit message says CONTRIBUTING "follows" — this line did not.

**4. Stale staging path in a permanent record.** `docs/adr/0032-one-actana-cli.md:157` says `actana-cli.cjs` is staged from `packages/cli/dist`; the enforcing line stages it from `packages/cli/dist-tarball` (`scripts/build-core-tarball.mjs:239`, `packages/cli/build.mjs:86`). Same stale path at `packages/cli/src/actana-cli-entry.ts:13` and `packages/core/build.mjs:36`.

**5. `packages/cli/src/orchestration-skill-install.ts:8-9`** (and its byte-identical twin `packages/shared/src/orchestration-skill-install.ts:8-9`) still states the superseded premise: *"those two programs cannot share a module. `@actana/cli` may not import `@actana/shared` (ADR 0025 D4, swept by `no-local-escape.test.ts`)."* ADR 0032 and the new note in ADR 0031 say exactly the opposite. Both copies must change together (fanout test).

**6. `packages/cli/src/actana-fetch-release.ts:98-100`** — on a wrong checksum, a first-time `actana install` is told *"the Core is still running the version it was. Retry the update."* There is no Core and no update. The module's own header (`:9-10`) says `install` should say "nothing was installed"; the sentence needs to be caller-supplied.

### Non-blocking notes

- `packages/cli/src/__tests__/skill-verbs-exist.test.ts:47` extracts only the first token after `actana`, so `actana session frobnicate` in the skill would pass. Worth extending to `<noun> <verb>` for the client nouns.
- `deploy/core.Dockerfile:242-243` lists the refusing verbs without `logs`.
- D4: the fallback is the layout's `current` link rather than the issue's `__dirname/..`; ADR 0032 D4 records it, so fine — but it means `node app/actana-cli.cjs setup` run directly out of a tarball (bypassing `bin/actana`) now downloads instead of installing the tree it is standing in.

### Acceptance checklist (enforcing line in brackets)

**Met:** single binary/help/version [`actana-cli.ts:1200-1343`, `actana-machine-cli.test.ts:219-242`]; image ships it and `CMD` resolves to it [`core.Dockerfile:205-207,275`, `smoke-core-image.mjs:432-480`]; skill verbs exist + test [`skill-verbs-exist.test.ts`]; daemon "wrong package" refusal deleted [`actana-cli.ts:1316-1326`, `actana-cli.test.ts:102-114`]; container refusal preserved, three vars, no `/.dockerenv` [`actana-container.ts`, `actana-container-contract.ts:79-81`]; ADR 0025 boundaries [`shared/package.json` private, no `@actana/cli` import in core]; CONTEXT.md [`:180-184`]; ADR 0032 exists with D1's rejected alternative [`:51-65`], D5's inline argument [`:121-144`], C1 [`:225-235`], 0031 D8 superseded + generator kept [`:215-224`]; #284/#285/#289 note the dependency; D1 move complete [no `actana-*` in `core/src`; D3 test green]; `core/build.mjs` emits no CLI; both bundles + `build-core-tarball.mjs:237-244` fails loudly; `build:core-tarball-bundles` in `package.json:55-56`, `release.yml:414,498`, guarded by `core-tarball.test.mjs:375-394`; D2 store+cert in shared, daemon imports from shared [`core-first-run.ts:45`, `core-entry.ts:92`], 9 Panel/SDK tests repointed; D3 "carries no daemon" assertion and rationale byte-identical [`no-local-escape.test.ts:153-174`]; D4 path + `createRequire` in-process, `Type=simple` comment travels [`actana-cli-entry.ts:20-28,184-192`]; D5 `:130` rewritten with the argument [`no-local-escape.test.ts:190-260`]; D6; D7 no `npm install` of `@actana/cli`; D8 install fetch→verify→unpack→setup with wrong-checksum-leaves-nothing test [`actana-install.test.ts:130-148`], `install.sh` + `install-sh.test.mjs` byte-identical; D9 on metal [`actana-install.test.ts:185-197`] and `core add` untouched; D10 no-clobber in the container's shape [`actana-setup.test.ts:527-583`], skew reported [`actana-status.test.ts:255-290`, `actana-machine-cli.test.ts:1654-1705`], nothing pinned; C1 narrowed with a reasoned table [`module-halves.ts:54-77`, `no-local-escape.test.ts:85-150`]; C2 `selfsigned` argued [`no-local-escape.test.ts:292-314`]; no exports map added.

**Partially met:** criterion 3 (`session …` on a container Core — item 1); criterion 4 "nothing to shadow" (`daemon` in the image — item 2); D9 (container — item 1); D9 skill rule (false on a container Core — item 1); "docs updated — every place that describes `actana` as one half" (items 3, 5).

**Not met:** none outright.

Requesting changes for item 1, with 2–6 in the same round. Happy to re-review promptly.

qcentic-dev:
Thanks — all six are addressed in `f4313f0`, pushed to `refactor/one-actana-cli`. Nothing else in the PR moved: same base, same train.

## 1 (blocking) — the container Core now registers itself, and D9's rule is true where the skill lands

Taken the first of your two options: **first-run wiring, into the registry, by the daemon** — `packages/core/src/core-self-register.ts`, called from `core-entry.ts` on the persisted-material path when `containerMode`. It goes through the same `wireLocalCore` `actana setup` uses, into the same directory the container's `AC_CORE_MATERIAL_FILE` already sits in, so a container Core ends up with an ordinary registry entry that every existing verb already understands.

Why not the other one: resolving a local Core out of `AC_CORE_MATERIAL_FILE` / `registration-blob.txt` in container mode would give the client half a second, container-only way to find a credential — one that `core ls`, `core use` and the `current` pointer would all have to agree with. ADR 0032 D9 now records the choice and the rejected alternative.

Three details worth your eye:

- **It runs on every boot, not only the one that mints.** A volume older than this has material and a blob but no registry entry, and a Core that wired itself only when minting would never repair one. The write is idempotent and the bearer is re-signed with a full lease each boot. Your selection rule is untouched: `current` pointing at a different Core is kept, and the log line says which.
- **The endpoint is `wss://127.0.0.1:<port>`, not `ACTANA_PUBLIC_HOST`.** The CLI doing the dialling shares a network namespace with the daemon, and the public host is the address *other* machines use — inside the container it may not route at all. `core-cert-material.ts` puts `127.0.0.1` in every server cert's SAN on both the mint and the re-issue path, so the loopback dial verifies. The printed pairing token still carries the public host, unchanged.
- **To let the daemon reach the registry at all**, the on-disk half of `blob-registry.ts` moved to `@actana/shared/blob-registry`, with `local-core-wiring.ts` beside it; `packages/cli` re-exports both name for name, so no caller changed. What stays in the CLI is the half that decodes a blob into a `core ls` row and needs the SDK. One implementation of "which Cores does this machine know about" was the point.

`scripts/smoke-core-image.mjs:457-460` no longer accepts the empty-registry sentence — it now asserts the built image lists this Core as **selected**, at the loopback endpoint, and that **`actana session ls` reaches it from inside the container**. That is criterion 3's second verb proven on the image rather than argued. `packages/core/src/__tests__/core-self-register.test.ts` covers naming, selection, no-clobber, XDG, re-registration and the non-fatal failure path.

I also ran it outside Docker against the built bundles — daemon booted with `ACTANA_CONTAINER=1` and `AC_CORE_PUBLIC_HOST=198.51.100.7` in a throwaway `$HOME`, then the CLI from that same home:

```
NAME        CURRENT  ENDPOINT               LABEL
smoke-core  *        wss://127.0.0.1:18443  smoke-core
$ actana session ls --json
[]
```

## 2 — `ACTANA_ROOT=/opt/actana` is in the image

Beside `AC_APP_PATH`, as you suggested; the Dockerfile comment and ADR 0032 D7 now say why, so "stops being able to decide anything" is true of `daemon` as well and not only of the client nouns. It also means an npm-installed `actana status` in the container reads the real manifest version instead of `unknown`.

## 3–6 — the stale records

- **3.** `CONTRIBUTING.md` no longer maps `packages/core/src/actana-cli.ts`; the tree there is the daemon's.
- **4.** `packages/cli/dist-tarball`, not `dist`, in ADR 0032 D7, `packages/core/build.mjs:36` and `actana-cli-entry.ts`'s header. `grep -rn 'packages/cli/dist[^-]'` is now empty.
- **5.** Corrected in both copies of `orchestration-skill-install.ts` — and, since the same sentence was live in four more places, in `harness-skill-targets.ts`, `harness-cli-config.ts`, `local-tar.ts`, `scripts/gen-skill-payload.mjs` and the fanout test's own failure message. They now say what ADR 0031 D8's supersession note already says: the copy is a convenience, not a workaround, and the arrangement is kept deliberately. The twins are still byte-identical.
- **6.** `fetchVerifiedRelease` takes a caller-supplied `noChange`. `install` says *"Nothing was installed. Retry the install"*, `update` keeps its own sentence, and `actana-install.test.ts`'s wrong-checksum case asserts both halves so `update`'s wording cannot drift back over `install`'s.

## Not done

Your three non-blocking notes are untouched, deliberately, to keep this round reviewable: the `skill-verbs-exist.test.ts` `<noun> <verb>` extension, `logs` missing from the Dockerfile's refusing-verb list, and D4's `node app/actana-cli.cjs setup` observation. Happy to take any of them here or on the next train — say which.

`pnpm test` (all six stages), `pnpm typecheck` and `pnpm lint` are green locally; the image smoke needs Docker, so CI is the first place its new assertions run.

qcentic-adm (approved):
## Re-review of #294 at `f4313f0` — every round-1 finding settled, from the enforcing lines

Branch is still exactly the train shape (`beta/0.4.0` + `7ccba9a` + `f4313f0`, nothing else moved). CI on `f4313f0`: 23/23 green, including **Core image / Build + smoke (amd64)**, which is where the new registry assertions actually run.

**1 (blocking) — container Core self-registration: settled.**
- `packages/core/src/core-self-register.ts:114-129` builds the blob with `buildRegistrationBlob` and calls the same `wireLocalCore` `setup` uses, via `@actana/shared/local-core-wiring`; `localDialHost()` (`:87-93`) maps `0.0.0.0`/`::`/`""` to `127.0.0.1`, which `core-cert-material.ts:169` already puts in every server cert's SAN.
- `packages/core/src/core-entry.ts:473-506` calls it on the persisted-material path, gated only on `containerMode`, every boot (so older volumes repair themselves), non-fatal on a read-only registry, and keeps an operator's `current` selection.
- Import arrows stay straight: `packages/shared/src/blob-registry.ts` imports only `node:fs`/`node:path`; `packages/cli/src/blob-registry.ts:24-48` and `local-core-wiring.ts` re-export name for name; no `@actana/cli` import anywhere in core/shared; core's `@actana/sdk` imports are still exactly `core-link-frames` and `core-files-error-codes` (ADR 0025 D2).
- Enforcing lines: `scripts/smoke-core-image.mjs:482-525` now **dies** on the empty-registry sentence, requires exactly one selected Core at `wss://127.0.0.1:<port>`, and runs `actana session ls --json` end to end inside the image — criterion 3's second verb proven on the built image. `packages/core/src/__tests__/core-self-register.test.ts:55-164` covers write+select, a decodable blob with a freshly signed bearer, loopback vs bind host, label→name fallback, no-clobber, idempotent re-registration, `XDG_CONFIG_HOME`, and the non-fatal failure path.
- ADR 0032 D9 now records the mechanism, the every-boot rule, the loopback endpoint, and the rejected alternative (CLI resolving a local Core from `AC_CORE_MATERIAL_FILE`). `SKILL.md:32-36` is now true on the machine the Core installs it on.

**2 — `ACTANA_ROOT`: settled.** `deploy/core.Dockerfile:267` sets `ACTANA_ROOT=/opt/actana` beside `AC_APP_PATH`, with the reason at `:253-261`; ADR 0032 D7 says the same. `daemonEntryPath()` (`actana-cli-entry.ts:93-106`) now resolves from whichever `actana` runs in the image.

**3 — `CONTRIBUTING.md`: settled.** The stale `packages/core/src/actana-cli.ts` row is gone; the core tree reads "the Core, and nothing else" with `core-first-run.ts` in its place.

**4 — staging path: settled.** `docs/adr/0032-one-actana-cli.md` D7, `packages/cli/src/actana-cli-entry.ts:13-18` and `packages/core/build.mjs:36` all say `packages/cli/dist-tarball`; `grep -rn 'packages/cli/dist([^-]|$)'` over the tree is empty.

**5 — superseded premise: settled.** Both `orchestration-skill-install.ts` twins state the supersession (ADR 0032 D5 / ADR 0031 D8 note) and are still byte-identical (`cmp` clean; payload twins too). The same correction landed in `harness-skill-targets.ts`, `harness-cli-config.ts`, `local-tar.ts`, `scripts/gen-skill-payload.mjs` and the fanout test's failure message. No remaining "may not import `@actana/shared`" stated as current fact.

**6 — wrong-checksum sentence: settled.** `fetchVerifiedRelease` takes a caller-supplied `noChange` (`actana-fetch-release.ts`); `actana-install.ts` passes "Nothing was installed. Retry the install", `actana-update.ts` keeps its own; `actana-install.test.ts` asserts the install wording is present **and** the update wording is absent.

The three non-blocking notes from round 1 (`skill-verbs-exist.test.ts` `<noun> <verb>` depth, `logs` missing from the Dockerfile's refusing-verb comment, D4's direct-`node app/actana-cli.cjs setup` observation) were deliberately left for a later train; none blocks the gate.

**Approving.** With `f4313f0`, every #288 acceptance criterion I marked partially met in round 1 — criterion 3, criterion 4, D9 (container), the D9 skill rule, and the docs sweep — is now met from an enforcing line. Ready for `beta/0.4.0`.

mrfqcentic (approved):
## Re-review at f4313f0 — my 7ccba9a approval was auto-dismissed; verified the delta against everything I approved. **Approving.**

Branch shape confirmed: `beta/0.4.0` + `7ccba9a` + `f4313f0`, nothing else moved. I read the full diff of `f4313f0` against my approved head and checked each claimed fix at its enforcing line, not its commit message.

### The blocking gap (container-Core self-registration) — real, and enforced on the built image

- `packages/core/src/core-self-register.ts:114-129` builds the blob with the same `buildRegistrationBlob` the printed-token path uses (`core-first-run.ts:78`) and wires it through the same `wireLocalCore` `setup` uses — now in `@actana/shared/local-core-wiring`. I diffed the moved modules against the 7ccba9a CLI originals: bodies byte-identical apart from the import specifier, and the CLI re-export surface is complete (every name any caller imports still resolves; callers untouched).
- `core-entry.ts:467-506` calls it on the persisted-material path, gated only on `containerMode`, on **every** boot (so pre-existing volumes self-repair), non-fatal on an unwritable registry, and `wireLocalCore` keeps an operator's `current` selection — the D9 no-clobber rule survives the move unchanged.
- The loopback claim checks out: `localDialHost()` maps `0.0.0.0`/`::`/`""` to `127.0.0.1`, and `issueServerCert` (`core-cert-material.ts:169`) puts `127.0.0.1` in every server cert SAN on both the mint path (`generateCertMaterial` → :101) and the re-issue path (`core-material-store.ts:135`). The dial verifies.
- Import arrows stay straight: the two new shared modules import only `node:fs`/`node:path` and each other; no `@actana/cli` import anywhere in core/shared; core's SDK imports remain exactly `core-link-frames` + `core-files-error-codes` (ADR 0025 D2). Core and CLI builds both inline `@actana/shared` (neither build.mjs marks it external), so D5's "inlined, not depended on" still holds — and the D5 test block is generic, so it covers the new imports without being touched.
- The enforcing line: `scripts/smoke-core-image.mjs:482-525` now **dies** on the empty-registry sentence, requires exactly one selected Core at `wss://127.0.0.1:<port>`, and runs `actana session ls --json` end to end inside the container — criterion 3's second verb proven on the built image, and it ran green in CI. `core-self-register.test.ts` covers write+select, a decodable blob with a freshly signed bearer, loopback vs bind host, label→name fallback, no-clobber, idempotent re-registration, `XDG_CONFIG_HOME`, and the non-fatal path. ADR 0032 D9 now records the mechanism, the every-boot rule, and the rejected CLI-side alternative. The skill text I cited is now true on the machine the Core installs it on.

### The five stale records — each fix confirmed at its line

2. **ACTANA_ROOT**: `deploy/core.Dockerfile` sets `ACTANA_ROOT=/opt/actana` beside `AC_APP_PATH`; `daemonEntryPath()` (`actana-cli-entry.ts:69,93-106`) tries `ACTANA_ROOT` before the `current` symlink, so an npm-shimmed `actana` in the image now finds `/opt/actana/app/core-entry.cjs`. D7's "nothing to shadow" is now true of `daemon`.
3. **CONTRIBUTING.md**: the dead `packages/core/src/actana-cli.ts` row is gone; `core-first-run.ts` stands in its place.
4. **Staging path**: ADR 0032 D7, `actana-cli-entry.ts`'s header and `core/build.mjs:36` all say `packages/cli/dist-tarball`; a tree-wide scan for the bare `packages/cli/dist` spelling comes back empty. (This also clears the one nit from my own round-1 review.)
5. **Superseded premise**: both `orchestration-skill-install.ts` twins corrected and still byte-identical (`cmp` clean), fanout-test message updated, and the same correction landed in `harness-skill-targets.ts`, `harness-cli-config.ts`, `local-tar.ts` (now citing the correct #129 D8 rule) and `gen-skill-payload.mjs`.
6. **Wrong-checksum sentence**: `fetchVerifiedRelease` takes a required caller-supplied `noChange`; `install` passes "Nothing was installed. Retry the install", `update` keeps its own; both are the only callers; `actana-install.test.ts` now asserts the install wording is present **and** the update wording absent.

### No regression to anything I approved at 7ccba9a

Tarball staging machinery, both `build.mjs` emitters, the narrowed C1 sweep with its reasoned exemption table, the unchanged D3 daemon ban, the D5 private-package guard, the dispatch/usage/version unification, the container refusal table + `DOCKER_EQUIVALENT`, `skill-verbs-exist.test.ts`, `install.sh`, and the D8/D10 install/collision/skew tests are all untouched or touched only in comments. The single behavioral change to previously-approved code is the registry move, which is a pure move with re-exports — verified line by line.

### CI on f4313f0

20/20 applicable checks green, including **Core image / Build + smoke (amd64)** where the new registry and `session ls` assertions execute, plus Unit Tests, Typecheck, both tarball smokes and both installer E2Es. The three skips are the train-only jobs, skipped by design on a PR.

Every #288 criterion I approved at 7ccba9a still holds, and the gap the other reviewer found is closed with an enforcing line rather than prose. Ready for `beta/0.4.0`.

PR #295, opened 2026-08-20
feat(core): the pairing server — pre-auth endpoint, CSR signing and bearer claims (#282)

qcentic-dev:
The server half of #280: the endpoint a client with a pairing code posts to, and everything this Core does before it hands back a credential.

`POST /v1/pair/redeem` takes `{ sessionId, code, client, csr }` and answers `{ caCert, clientCert, bearer, endpoint }`. The client's key is born on the client and never crosses the wire in either direction — there is nothing in this endpoint that could return one, because there is nothing in it that has one.

Base is `feat/short-code-pairing`, not `beta/0.4.0` and not `main`.

## The three decisions worth a reviewer's time

**1. How a pre-auth route exists on an mTLS server.** There is no per-route TLS: a server either demands a verified client certificate before the request line exists or it does not. So a Core that mounts the pairing endpoint completes the handshake without one (`requestCert: true`, `rejectUnauthorized: false`) and enforces the same rule one layer up — per request and per WebSocket upgrade, from `socket.authorized`, in `core-preauth-gate.ts`.

The cost, stated plainly: refusal moves from the TLS layer to the HTTP layer, so an unauthenticated caller gets a `403` where it used to get a handshake failure. What it does not cost is access — the gate defaults to refusing, the core link has no pairing exception at all, and **a Core that mounts no pre-auth surface keeps `rejectUnauthorized: true` and behaves exactly as it did before this PR.** `core-files-mtls.test.ts` is unchanged and still asserts the old behaviour.

The WebSocket moves to `ws`'s `noServer` mode, because in the `{ server }` form `ws` owns the `upgrade` event and there is nowhere to put the check. `ws` then stops forwarding the HTTP server's `error`, so a listen failure is forwarded explicitly instead.

**2. Where the pairing state lives.** In `packages/shared/src/pairing-store.ts`, beside `pairing-session.ts` and `core-material-store.ts` — not in `packages/core`. Two processes need it: `actana pair new` (#283) mints the session from the `actana` CLI, whose entry point says at length why that binary never imports `@actana/core`, and the daemon redeems it. That is the same reason `core-material-store.ts` moved into `packages/shared` in #288.

The stored `codeHash` is an HMAC keyed by a value derived from the bearer secret with its own domain separator, and it covers the session id as well as the code. A bare `sha256` would not do: eight characters from a 31-character alphabet is ~2^39.6 digests, enumerable by anyone who copies the file. Binding the digest to the session id makes session binding hold cryptographically, not only at the lookup.

**3. The order of the two "spend" steps.** The CSR is checked *before* the session is consumed, and the session is consumed *before* anything is signed. A malformed CSR is a client bug and must not cost the operator a session; consuming is what decides a race, so two simultaneous redemptions both reach that line, one marks the session consumed and goes on to sign, and the other gets the refusal a replay gets. A signature that then fails leaves the session spent — the safe direction, where the other one hands two clients a certificate for one code.

## Also here

- **Uniform refusal.** Expired, consumed, unknown, dead, wrong code: one status, one body, no header that differs. The audit log on the Core says which — the operator reading it already owns the Core.
- **Rate limiting** independent of the per-session cap, in two fixed windows (per peer, and a global backstop against a spray from many addresses), with a bounded peer map — the keys are chosen by whoever is dialling.
- **Bearer claims.** `iss`, `sub`, `aud`, `jti` beside `exp`, all inert in this train, all optional so every bearer minted before this still verifies. `sub` is the certificate serial; `aud` is the new stable core UUID.
- **The stable core UUID** is minted at install, minted on load for material that predates it (and written back by the daemon's boot path), and survives `reissueServerCert` — asserted directly, because an audience that changed when a certificate was re-signed would be no more stable than the certificate.
- **CSR signing** goes through `core-cert-material.ts` against the Core's own CA, with the same client-leaf extensions `generateCertMaterial` writes for the hand-carried Panel cert — the test asserts the two sets are *equal* rather than restating them. Nothing else is taken from the request: not the subject, not the requested extensions (a CSR asking for `cA: true` gets a leaf like everyone else), only the public key whose possession it proves.
- **`@peculiar/x509`** is added to `packages/shared` — `selfsigned` cannot sign a CSR, and this is the library `selfsigned` is itself built on, so it was already in the tree.

## Tests

`core-pairing-redeem.test.ts` runs against a real in-process Core built by the default server factory: happy path end to end, five wrong codes killing the session, expired, replay of a consumed session, a code redeemed against the wrong session, the rate limit tripping, no private key anywhere in the response, the issued certificate verifying against the Core CA — and that certificate plus its bearer completing a real mTLS handshake and an `auth` frame on the core link. Alongside it: the pre-auth gate, the route composition order, the rate limiter, the audit redaction, and the pairing store.

`pnpm typecheck`, `pnpm lint` and `pnpm test` are green locally (one panel flake under load, green on re-run in isolation and in full).

## For task 4 (#284)

The wire shape above is what the SDK declares. It is what #280 describes — the code, the client info and a CSR in; `caCert`, `clientCert`, `bearer` and `endpoint` out — so there is no change to settle in #280 before that task starts.

Refs #282
Part of #280

https://claude.ai/code/session_016JQhKrubNJNE2tMLysbR1z

### DIFF
--- a/packages/core/src/__tests__/core-first-run.test.ts
+++ b/packages/core/src/__tests__/core-first-run.test.ts
@@ -29,6 +29,7 @@ const sample: PersistedMaterial = {
   clientKey: "-----BEGIN PRIVATE KEY-----\nCLIENTKEY\n-----END PRIVATE KEY-----",
   bearerSecret: "deadbeef".repeat(8),
   coreId: "core_abcdef0123456789",
+  coreUuid: "1f2e3d4c-5b6a-4798-8a9b-0c1d2e3f4a5b",
   serverHost: "core.example.test",
 };
 
--- a/packages/core/src/__tests__/core-pairing-audit.test.ts
+++ b/packages/core/src/__tests__/core-pairing-audit.test.ts
@@ -1,0 +1,84 @@
+// What the audit log is allowed to say (#282).
+//
+// A log line is the easiest place in this feature for a secret to end up: it is
+// written on the failure path, read by people, and outlives the request by
+// however long the log is kept. So the redaction is a function with a test.
+import { describe, expect, it } from "vitest";
+import { pairingAuditor, redactPairingAuditEvent, type PairingAuditEvent } from "../core-pairing-audit";
+
+const attempt: PairingAuditEvent = {
+  outcome: "issued",
+  sessionId: "ps_1",
+  label: "laptop",
+  peer: "10.0.0.9",
+  certSerial: "0a1b",
+  at: 1_700_000_000_000,
+};
+
+describe("the audit record", () => {
+  it("keeps what an operator needs to read the attempt", () => {
+    expect(redactPairingAuditEvent(attempt)).toEqual({
+      outcome: "issued",
+      sessionId: "ps_1",
+      label: "laptop",
+      peer: "10.0.0.9",
+      certSerial: "0a1b",
+      at: 1_700_000_000_000,
+    });
+  });
+
+  it("writes nothing for a field the attempt did not have", () => {
+    const record = redactPairingAuditEvent({ outcome: "bad-request", peer: "10.0.0.9", at: 1 });
+    expect(Object.keys(record).sort()).toEqual(["at", "outcome", "peer"]);
+  });
+
+  it("drops anything not on the list, whatever a caller attaches", () => {
+    // The list is the mechanism. This is what a future field carrying the code,
+    // the CSR or a key would meet — nothing is logged until somebody adds it to
+    // `LOGGED_FIELDS`, which is a line in a diff a reviewer sees.
+    const smuggled = {
+      ...attempt,
+      code: "ABCD-EFGH",
+      csr: "-----BEGIN CERTIFICATE REQUEST-----",
+      clientKey: "-----BEGIN PRIVATE KEY-----",
+    } as PairingAuditEvent;
+
+    const record = redactPairingAuditEvent(smuggled);
+
+    expect(JSON.stringify(record)).not.toContain("ABCD-EFGH");
+    expect(JSON.stringify(record)).not.toContain("CERTIFICATE REQUEST");
+    expect(JSON.stringify(record)).not.toContain("PRIVATE KEY");
+  });
+
+  it("records a refusal's reason, which the response refuses to give", () => {
+    // The distinction the wire will not draw is exactly the one the operator
+    // reading their own Core's log needs.
+    const record = redactPairingAuditEvent({
+      outcome: "refused",
+      reason: "attempts-exhausted",
+      sessionId: "ps_1",
+      peer: "10.0.0.9",
+      attempts: 5,
+      at: 1,
+    });
+    expect(record).toMatchObject({ reason: "attempts-exhausted", attempts: 5 });
+  });
+});
+
+describe("the auditor", () => {
+  it("hands the sink an already-redacted record", () => {
+    const written: Record<string, unknown>[] = [];
+    pairingAuditor((record) => written.push(record))(attempt);
+    expect(written).toEqual([redactPairingAuditEvent(attempt)]);
+  });
+
+  it("does not let a broken sink take the request with it", () => {
+    // The attempt has already been decided by the time this runs. Losing the
+    // record is bad; answering a paired client with a 500 because the log was
+    // unwritable is worse.
+    const audit = pairingAuditor(() => {
+      throw new Error("disk full");
+    });
+    expect(() => audit(attempt)).not.toThrow();
+  });
+});
--- a/packages/core/src/__tests__/core-pairing-rate-limit.test.ts
+++ b/packages/core/src/__tests__/core-pairing-rate-limit.test.ts
@@ -1,0 +1,125 @@
+// The defence the per-session attempt cap cannot provide (#280, #282).
+//
+// A session dies after five wrong codes. That bounds guessing inside one
+// session and nothing else: an attacker who can open the endpoint at will gets
+// five fresh guesses for every session an operator ever opens, as fast as they
+// can send them. These tests pin the two windows that close that gap, and the
+// memory bound that keeps the closing of it from being its own attack.
+import { describe, expect, it } from "vitest";
+import { MAX_TRACKED_PEERS, PairingRateLimiter } from "../core-pairing-rate-limit";
+
+function limiterAt(clock: { now: number }, opts: ConstructorParameters<typeof PairingRateLimiter>[0] = {}) {
+  return new PairingRateLimiter({ ...opts, now: () => clock.now });
+}
+
+describe("the per-peer window", () => {
+  it("allows up to the limit and refuses the next", () => {
+    const clock = { now: 1_000 };
+    const limiter = limiterAt(clock, { peer: { limit: 3, windowMs: 60_000 } });
+
+    const verdicts = [1, 2, 3, 4].map(() => limiter.check("10.0.0.1"));
+
+    expect(verdicts.map((v) => v.ok)).toEqual([true, true, true, false]);
+    expect(verdicts[3]).toMatchObject({ ok: false, scope: "peer" });
+  });
+
+  it("does not count a refusal against the caller a second time", () => {
+    // Otherwise a caller who keeps knocking extends their own lockout, which is
+    // how one office behind one NAT address locks itself out for an afternoon.
+    const clock = { now: 1_000 };
+    const limiter = limiterAt(clock, { peer: { limit: 2, windowMs: 60_000 } });
+
+    limiter.check("10.0.0.1");
+    limiter.check("10.0.0.1");
+    limiter.check("10.0.0.1");
+    limiter.check("10.0.0.1");
+    clock.now += 60_001;
+
+    expect(limiter.check("10.0.0.1").ok).toBe(true);
+  });
+
+  it("keeps one peer's spending off another peer's budget", () => {
+    const clock = { now: 1_000 };
+    const limiter = limiterAt(clock, { peer: { limit: 2, windowMs: 60_000 } });
+
+    limiter.check("10.0.0.1");
+    limiter.check("10.0.0.1");
+
+    expect(limiter.check("10.0.0.1").ok).toBe(false);
+    expect(limiter.check("10.0.0.2").ok).toBe(true);
+  });
+
+  it("lets the window slide off", () => {
+    const clock = { now: 1_000 };
+    const limiter = limiterAt(clock, { peer: { limit: 1, windowMs: 60_000 } });
+
+    expect(limiter.check("10.0.0.1").ok).toBe(true);
+    expect(limiter.check("10.0.0.1").ok).toBe(false);
+    clock.now += 60_001;
+    expect(limiter.check("10.0.0.1").ok).toBe(true);
+  });
+
+  it("says how long to wait", () => {
+    const clock = { now: 1_000 };
+    const limiter = limiterAt(clock, { peer: { limit: 1, windowMs: 60_000 } });
+
+    limiter.check("10.0.0.1");
+    clock.now += 10_000;
+    const refused = limiter.check("10.0.0.1");
+
+    expect(refused).toMatchObject({ ok: false, retryAfterMs: 50_000 });
+  });
+});
+
+describe("the global window", () => {
+  it("stops a spray spread across many addresses", () => {
+    // The reason keying by address is not the whole answer: every one of these
+    // callers is inside its own per-peer limit.
+    const clock = { now: 1_000 };
+    const limiter = limiterAt(clock, {
+      peer: { limit: 10, windowMs: 60_000 },
+      global: { limit: 5, windowMs: 60_000 },
+    });
+
+    const verdicts = Array.from({ length: 6 }, (_, i) => limiter.check(`10.0.0.${i}`));
+
+    expect(verdicts.slice(0, 5).every((v) => v.ok)).toBe(true);
+    expect(verdicts[5]).toMatchObject({ ok: false, scope: "global" });
+  });
+
+  it("names the peer limit rather than the global one when both would refuse", () => {
+    // An operator who has mistyped a code eleven times should be told it was
+    // them, not the Core.
+    const clock = { now: 1_000 };
+    const limiter = limiterAt(clock, {
+      peer: { limit: 1, windowMs: 60_000 },
+      global: { limit: 1, windowMs: 60_000 },
+    });
+
+    limiter.check("10.0.0.1");
+
+    expect(limiter.check("10.0.0.1")).toMatchObject({ ok: false, scope: "peer" });
+  });
+});
+
+describe("the memory bound", () => {
+  it("forgets the quietest peers rather than growing without limit", () => {
+    // The keys here are chosen by whoever is dialling, on a pre-auth surface.
+    // An unbounded map would be an exhaustion attack costing one packet each.
+    const clock = { now: 1_000 };
+    const limiter = limiterAt(clock, {
+      peer: { limit: 1, windowMs: 1_000 },
+      global: { limit: Number.MAX_SAFE_INTEGER, windowMs: 60_000 },
+    });
+
+    for (let i = 0; i < MAX_TRACKED_PEERS + 50; i += 1) {
+      limiter.check(`10.${Math.floor(i / 65536)}.${Math.floor(i / 256) % 256}.${i % 256}`);
+      // Age each caller out of its own window, so eviction has candidates —
+      // the case a real Core meets, where the sprayer's earlier addresses have
+      // gone quiet by the time the map fills.
+      clock.now += 2_000;
+    }
+
+    expect(limiter.trackedPeers()).toBeLessThanOrEqual(MAX_TRACKED_PEERS);
+  });
+});
--- a/packages/core/src/__tests__/core-pairing-rate-limit.test.ts
+++ b/packages/core/src/__tests__/core-pairing-rate-limit.test.ts
@@ -6,7 +6,13 @@
 // can send them. These tests pin the two windows that close that gap, and the
 // memory bound that keeps the closing of it from being its own attack.
 import { describe, expect, it } from "vitest";
-import { MAX_TRACKED_PEERS, PairingRateLimiter } from "../core-pairing-rate-limit";
+import {
+  DEFAULT_GLOBAL_WINDOW,
+  DEFAULT_PEER_WINDOW,
+  EVICTION_BATCH,
+  MAX_TRACKED_PEERS,
+  PairingRateLimiter,
+} from "../core-pairing-rate-limit";
 
 function limiterAt(clock: { now: number }, opts: ConstructorParameters<typeof PairingRateLimiter>[0] = {}) {
   return new PairingRateLimiter({ ...opts, now: () => clock.now });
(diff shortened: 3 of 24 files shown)


qcentic-dev:
## Acceptance criteria from #282

Every criterion from the issue, in its order, checked or explicitly not.

- [x] **A pairing route family in `packages/core/src/` (suggested `core-pairing-routes.ts` + a wiring module), mounted on the same HTTPS server as the `/v1/…` routes, reachable without a client certificate while every other route keeps its current mTLS requirement** — `core-pairing-routes.ts` + `core-pairing-wiring.ts`, composed with the file routes in `core-entry.ts` onto the one `https.Server` in `pty-core-link-server.ts`. The relaxation is `rejectUnauthorized: false` **only when a pre-auth path is mounted**, with `core-preauth-gate.ts` enforcing the same rule per request and per upgrade; a Core without pairing keeps the TLS-level refusal, and `core-files-mtls.test.ts` is unchanged. Asserted end to end in `core-pairing-redeem.test.ts` — "the pre-auth hole is exactly one route wide": pairing answers, the file routes 403, an unknown path 403, and the core-link upgrade is refused.
- [x] **The redemption handler accepts the pairing code, client info and a PEM CSR, and validates in this order: rate limit → session lookup and binding → TTL → single-use → attempt cap** — that order, written out in `route()` under numbered headings. The three state checks are spelled out rather than delegated to `canRedeem`, whose internal order is its own, so the order a reader checks against the spec is the order in the file.
- [x] **Rate limiting on the endpoint independent of the per-session attempt cap** — `core-pairing-rate-limit.ts`, two fixed windows: per peer (10/min) and a global backstop (60/min) that survives a spray from many source addresses. The peer map is bounded, because on a pre-auth surface the keys are chosen by the caller. Test: a session with five attempts left is stopped at three by a limiter configured to three.
- [x] **Code comparison is constant-time (`timingSafeEqual`, as `core-link-bearer.ts` already does) over the canonical form from task 1** — with one thing to read deliberately: the input is task 1's `normalisePairingCode` output, and `timingSafeEqual` runs over the **HMAC digests** of that canonical form (`pairingCodeMatches`), not over the two strings. The store holds a digest rather than a code by design (#281: "the plaintext is printed once and never stored"), so the digest is what there is to compare — and it is what a caller would have to produce to redeem, which is why it gets the constant-time comparison.
- [x] **A wrong code increments the session's attempt counter and kills the session at 5; an expired, consumed, unknown or dead session is refused with a response that does not distinguish which of those it was** — one `PAIRING_REFUSED` object referenced by every branch. Tests assert five refusals are byte-identical, that the right code afterwards gets the same body, and that an unknown session's status, body **and `content-length`** match a wrong code's. A code outside the alphabet counts as a wrong attempt rather than a `400`, so it is not a free probe.
- [x] **On success: the CSR is signed against the Core's own CA via the `core-cert-material.ts` machinery with the client-leaf extensions (`cA:false`, `digitalSignature`+`keyEncipherment`, `extKeyUsage clientAuth`); the session is marked consumed in the same step that issues, so two concurrent redemptions cannot both win** — `signClientCsr` in `packages/shared/src/core-cert-material.ts` (where #288 moved that file). The extension test compares the issued leaf against `generateCertMaterial`'s hand-carried client cert for equality rather than restating the list. Consumption is a synchronous compare-and-set in the store, taken **before** signing; the concurrency test fires two redemptions of one code and asserts `[200, 403]` with one client recorded.
- [x] **The response carries `caCert`, `clientCert`, `bearer` and `endpoint` — and never a private key** — asserted as an exact key set, and separately by searching the whole body for `PRIVATE KEY` and for the client's actual key bytes, so a field added later cannot slip past.
- [x] **The issued bearer carries `iss`, `sub`, `aud` (the stable core UUID) and `jti` alongside the existing `exp`, and existing bearer verification in `packages/shared/src/core-link-bearer.ts` accepts it** — all four optional, so a bearer minted before this still verifies (test), and the end-to-end test reads them back off a bearer that came out of a real redemption. `sub` is the certificate serial.
- [x] **A stable core UUID is minted at install if absent, persisted alongside the rest of the material (`core-material-store.ts`), and survives `reissueServerCert` unchanged** — minted in `mintFreshMaterial`, minted on load by `readMaterialFile` for material that predates the field, written back by `core-first-run.ts` so it is stable across boots, and asserted across `reissueServerCert` directly.
- [x] **The paired client identity is persisted (certificate subject/serial, label, paired-at, and the three nullable `created_by` / `tenant_id` / `auth_method` fields from task 1)** — `PairedClient` in `packages/shared/src/pairing-store.ts`, with `revokedAt` and `certNotAfter` beside them so #283's `pair revoke` has a stamp to write and `pair ls` has an expiry to show. The three inert fields are copied from the session.
- [x] **Every pairing attempt is audit-logged — success and failure, with the outcome, the session label and the peer address, and never the code itself, the CSR or any key material** — `core-pairing-audit.ts`. The record is built from a fixed field list, so a field added to the event type later is not logged until somebody adds it to that list; a test smuggles a code, a CSR and a key into an event and asserts none reach the record. Attempts refused before the body parses are logged too.
- [x] **Tests** — all of them, in `core-pairing-redeem.test.ts` against a real in-process server built by the default factory: happy path end to end; wrong code five times kills the session; expired session refused; replay of a consumed session refused; a redemption bound to session A refused against session B (and session A's counter untouched); rate limit trips; the response contains no private key; the issued cert verifies against the Core CA; and that certificate plus its bearer complete a real mTLS handshake and an `auth` frame on the core link. Unit suites alongside: `core-preauth-gate`, `core-pairing-wiring`, `core-pairing-rate-limit`, `core-pairing-audit`, `pairing-store`, plus the new cases in `core-cert-material`, `core-link-bearer` and `core-material-store`.
- [x] **The wire shape agreed here is the contract task 4 declares in the SDK — settle any change to it in #280 before implementing, not in a PR review** — nothing needed settling. The shape is what #280 already describes: the code, the client info and a CSR in; `caCert`, `clientCert`, `bearer`, `endpoint` out. The request field names (`sessionId`, `code`, `client`, `csr`) and `POST /v1/pair/redeem` are the only additions, and they name things #280 already fixed.

### Two things a reviewer should know, neither of them an unmet criterion

- **Nothing mints a session yet.** `actana pair new` is #283. The endpoint reads the store this PR defines; until that command exists, only a test writes to it.
- **Pairing is mounted only where there is persisted material** (`AC_CORE_MATERIAL_FILE`). The legacy fresh-material daemon path has no CA key on disk, no stable file for `pair new` to write, and no `pair new` to run — so it mounts nothing and keeps today's TLS posture exactly.

mrfqcentic (changes requested):
Reviewed against #282's acceptance criteria, proving each from the enforcing lines rather than the description. Twelve of the thirteen criteria are met; the thirteenth (rate limiting) is met in its guessing-defence half and broken in its memory-bound half. That one is blocking, and it is one line.

I ran the tree at `021abb1` in an isolated clone: `typecheck` clean across all five packages, `lint` 0 errors (the one warning in a file you touched, `pty-core-link-server.ts:51`, predates this PR), and every suite green — core 813, shared 384, cli 899, sdk 253, panel 1439. CI is green on all 20 checks. This is careful work and the commit-by-commit narration is genuinely useful; what follows is the one thing that has to change and three smaller things.

---

## Blocking

### `packages/core/src/core-pairing-rate-limit.ts:110` — the peer map is not bounded on the branch an attacker drives

`check()` calls `evictIfCrowded()` only on the allow path (line 115). The global-window refusal at line 110 does `this.peers.set(peer, peerHits)` — which **inserts a new entry** (an empty array) for a peer never seen before — and returns at line 111 without ever reaching the eviction. Once the global window is saturated, the allow branch is never taken again, so eviction never runs.

Saturating it is cheap: `DEFAULT_GLOBAL_WINDOW` is 60/min, reachable from six addresses given the per-peer cap of 10. After that, every request from a fresh source address is a permanent `Map` entry on the Core's only pre-auth surface.

I measured it against the shipped class with the real defaults — 50,000 requests from distinct addresses inside one window:

```
scopes seen                        : global
MAX_TRACKED_PEERS                  : 4096
peers tracked after 50k addresses  : 50006
```

and with the global window left unsaturated, the bound holds exactly:

```
peers tracked with an unsaturated global window: 4096
```

This is precisely the attack the module's own header says it prevents — "an unbounded map is a memory exhaustion attack costing one packet per entry" — and the invariant does not hold.

### `packages/core/src/__tests__/core-pairing-rate-limit.test.ts:112` — the test masks it

"forgets the quietest peers rather than growing without limit" (line 106) passes `global: { limit: Number.MAX_SAFE_INTEGER, windowMs: 60_000 }` at line 112, which forces every call down the allow branch — the only branch that evicts. Swap in `DEFAULT_GLOBAL_WINDOW` and the test fails. Please make the regression test exercise the refusal path, since that is the one an attacker takes.

The fix is to evict on the refusal paths too, or not to touch `peers` at all when refusing globally.

One thing to check while you are in there: `evictIfCrowded` rescans the whole map on every insert once the map is at the cap and every peer is still inside its window (lines 138-151). My 50k-address probe on the *unsaturated* path took 53s for that reason. Not a security issue — it is the fail-safe direction — but worth a glance.

---

## Non-blocking

### `packages/core/src/core-first-run.ts:161` — a new write on what was a read-only boot path

`if (read.mintedCoreUuid) persistMaterialToFile(opts.materialFile, existing)` runs before `checkServerCertHost`. Previously this branch only wrote when the SAN had actually gone stale; now the first boot after upgrade writes for any pre-#282 material. If the material file or its directory is not writable by the daemon — a read-only bind mount, a K8s secret mount, material provisioned by another user — the throw propagates to `core-entry.ts:452`, which prints the errno and `process.exit(1)`s. A Core that booted fine before this PR would stop booting.

`core-material-store.ts` already says an unpersisted mint "is not a correctness failure — nothing reads `aud` in this train", so this deserves a `try`/`catch` that logs and carries on rather than taking the daemon down.

### `packages/core/src/pty-core-link-server.ts:2180-2185` — the 403 on a refused upgrade can be truncated

`socket.write(...)` immediately followed by `socket.destroy()` discards unflushed bytes, so on a TLSSocket the refusal is nondeterministically replaced by an `ECONNRESET`. `ws`'s own `abortHandshake` avoids this with `socket.once('finish', socket.destroy); socket.end(message)`. The tell is that the test at `core-pairing-redeem.test.ts` has to accept either outcome (`/403|Unexpected server response/i`) — worth tightening so the assertion can name one.

### `packages/core/src/core-pairing-routes.ts:398` and `:343` — two clocks in one audit trail

`refuse()` stamps `at: Date.now()` while every other audit call in the file uses the injected `now()`. Same class at line 343: `recordClient(client)` is called without a `now`, so `PairingStore.recordClient`'s session pruning also falls back to `Date.now()`. Under a driven clock the expired/consumed/dead lines interleave wrongly with the rest of the trail.

---

| # | Criterion | Verdict |
|---|---|---|
| 1 | Pairing route family in `packages/core/src/`, same HTTPS server, reachable without a client cert while every other route keeps mTLS | **Met** |
| 2 | Validation order: rate limit → session lookup/binding → TTL → single-use → attempt cap | **Met** |
| 3 | Rate limiting independent of the per-session cap | **Partially met** — the guessing defence works and is tested; the documented memory bound does not hold (blocking item above) |
| 4 | Constant-time code comparison over task 1's canonical form | **Met** |
| 5 | Wrong code increments and kills at 5; expired/consumed/unknown/dead refused indistinguishably | **Met** |
| 6 | CSR signed against the Core's CA with the client-leaf extensions; consumed so two redemptions cannot both win | **Met** |
| 7 | Response carries `caCert`, `clientCert`, `bearer`, `endpoint` — never a private key | **Met** |
| 8 | Bearer carries `iss`, `sub`, `aud` (stable core UUID), `jti` beside `exp`; existing verification accepts it | **Met** |
| 9 | Stable core UUID minted at install if absent, persisted, survives `reissueServerCert` | **Met** |
| 10 | Paired client identity persisted (subject/serial, label, paired-at, three nullable fields) | **Met** |
| 11 | Every attempt audit-logged, never the code, CSR or key material | **Met** |
| 12 | Tests: happy path, five wrong codes, expired, replay, cross-session, rate limit, no private key, cert verifies, mTLS handshake + `auth` frame | **Met** |
| 13 | Wire shape is the contract task 4 declares | **Met** |

### What I proved, and how

**The pre-auth boundary (AC 1).** The enforcing line is `clientCertVerified` in `pty-core-link-server.ts:2312`, `socket.authorized !== false`. I probed Node's TLS semantics directly on a server built the way `defaultCreateServer` builds one (`requestCert: true, rejectUnauthorized: false`):

```
no client cert    : authorized=false  (boolean, UNABLE_TO_GET_ISSUER_CERT) → unverified
CA-signed cert    : authorized=true                                        → verified
self-signed cert  : authorized=false  (DEPTH_ZERO_SELF_SIGNED_CERT)        → unverified
plain http socket : authorized=undefined                                   → verified
```

So `!== false` refuses both "no certificate" and "a certificate this CA does not vouch for", and reads `true` only on a plain socket where there are no certificates at all — which is the documented loopback case. The gate (`core-preauth-gate.ts:62-64`) defaults to `refuse`, `rejectUnauthorizedAtHandshake` keeps the handshake strict when no predicate is supplied, and `core-entry.ts:623` supplies one only when there is persisted material. `core-files-mtls.test.ts` is untouched and still passes.

I also checked the gate and both routers cannot disagree about a path: all three parse via `new URL(req.url, "https://core.invalid")`, so `/v1/pair/../projects/p1/files` normalises to a file path and is **refused**, while `/v1/pair%2f..%2fprojects/...` fails the prefix test and is refused too. No split-brain.

**The CSR signing path (AC 6).** I posted a hostile CSR at `signClientCsr` — subject `CN=mission-control-core-ca,O=Mission Control`, requesting `basicConstraints cA:true pathLen:5`, `keyCertSign|cRLSign`, `extKeyUsage serverAuth`:

```
issued subject : CN=laptop
issued ca flag : false
issued exts    : 2.5.29.19 ca=false, 2.5.29.15, 2.5.29.37
issued EKU     : [ '1.3.6.1.5.5.7.3.2' ]   ← clientAuth only
```

Nothing but the public key survives the request. A 1024-bit RSA key is refused `weak-key`; a CSR with a flipped signature byte is refused `bad-signature`. Signing goes through `packages/shared/src/core-cert-material.ts`, and the check order in `core-pairing-routes.ts` is `assertSignableCsr` (line 288) → `consume` (line 307) → `signClientCsr` (line 312), which is stricter than "consumed in the same step that issues" and fails in the safe direction.

**The bearer claims (AC 8).** `core-pairing-routes.ts:345-356` mints `iss: core:<coreId>`, `sub: pair:<cert serial>`, `aud: material.coreUuid`, `jti: randomUUID()` beside `exp`. `core-entry.ts` feeds `coreUuid` straight from `PersistedMaterial`. `core-link-bearer.ts:183` runs `readStandardClaims`, which treats absence as valid and a non-string as `malformed`, and `signBearer` omits absent claims so a pre-#282 token is byte-identical. Proven end to end by the test that takes the issued cert and bearer through a real mTLS handshake and an `auth` frame.

**File locations.** `core-material-store.ts` and `core-cert-material.ts` are opened at their **`packages/shared/src/`** locations, not the pre-move `packages/core` paths the issue text still cites. `core-pairing-routes.ts` imports `@actana/shared/core-cert-material` and `@actana/shared/core-link-bearer`; the only relative `./core-cert-material` import is inside `packages/shared` itself. No reference to a `packages/core` copy survives anywhere in the tree.

---

## Verdict

**Request changes** — for `core-pairing-rate-limit.ts:110` and its masking test at `core-pairing-rate-limit.test.ts:112`. Everything else here is sound, and the three non-blocking notes are yours to take or leave. Once the peer map is bounded on the refusal branch and the test exercises that branch, this is ready.

Note that the PR is still a **draft** — it will need marking ready for review before it can merge.

qcentic-dev:
> **Request changes** — for `core-pairing-rate-limit.ts:110` and its masking test at `core-pairing-rate-limit.test.ts:112`.

Fixed in d570575. You were right on both halves, and the second half is why the first survived review: the test asserted the bound only on the branch where it was never in doubt.

**The refusal branches no longer record a caller.** The global branch writes nothing to `peers` at all, and the peer branch rewrites only a caller already tracked (`has` before `set` — a limit of zero would otherwise land an unseen caller there and insert an empty array for them). Neither is a behaviour change: a refused attempt was never charged to the caller's window either, which is the rule the peer branch already followed and the reason a lockout cannot be extended by knocking. `evictIfCrowded()` now also runs on the way out of every branch, so the invariant no longer depends on which branch a later edit adds a `set` to — either mechanism alone would hold it today.

**The eviction scan is amortised.** Your 53s stood up: trimming to exactly the cap left the map one insert over it again immediately, so the scan ran on every subsequent insert. It now trims to a low-water mark (`MAX_TRACKED_PEERS - 512`), so one scan covers 512 inserts. The cap is still never exceeded by more than the entry that triggered the pass. The new 8,192-address spray test runs in ~100ms.

**The test now exercises the branch an attacker takes.** `Number.MAX_SAFE_INTEGER` is gone; that test runs with `DEFAULT_GLOBAL_WINDOW` and still covers the insert path (its clock moves two seconds a call, so sixty-a-minute is never reached). Four new cases cover the refusal path:

- a spray of `MAX_TRACKED_PEERS * 2` distinct addresses against a saturated global window — every verdict `scope: "global"`, and `trackedPeers()` **exactly** what it was before the spray, which is stronger than the bound and is the actual fix;
- a globally-refused caller leaves nothing behind and still has a full budget once the window slides;
- a peer-refused stranger is not recorded (the zero-limit edge the `has` guard covers);
- a tracked caller's window is still pruned when it is refused — the one write the peer branch keeps.

I checked they fail against the previous implementation rather than merely passing against the new one. Restoring `021abb1`'s logic with only the new constants added:

```
× holds the bound once the global window is saturated, which is the attacker's branch
    AssertionError: expected 8198 to be less than or equal to 4096
× does not record a caller the peer window refuses without ever having admitted
    AssertionError: expected 1 to be +0
```

Locally: `packages/core` 817 tests green (was 813), `tsc --noEmit` clean, `eslint` clean on both files, and commitlint green over all nine commits with this repo's config.

**On the three non-blocking notes** — I have deliberately not touched them in this push, since the ask was the blocking finding. All three read as correct to me and I am happy to take them in a follow-up commit here or in a separate issue, whichever you prefer:

1. `core-first-run.ts:161` — the new write can take a daemon down on read-only material. A `try`/`catch` that logs and carries on matches what `core-material-store.ts` already says about an unpersisted mint.
2. `pty-core-link-server.ts:2180` — `socket.end(message)` with `once('finish', destroy)`, as `ws`'s own `abortHandshake` does, so the test can name one outcome instead of accepting either.
3. `core-pairing-routes.ts:398` and `:343` — `refuse()` and `recordClient()` should take the injected clock like every other call in the file.

Still a draft, base unchanged (`feat/short-code-pairing`), nothing merged.

mrfqcentic (approved):
Settled. I re-derived this from the enforcing lines at `d570575` and from measurement, not from the reply, and then checked the reply's claims against what I measured. It holds up.

## The memory bound, from the enforcing lines

There are now exactly **two** `this.peers.set` call sites in `core-pairing-rate-limit.ts`, and neither can grow the map unboundedly:

- **`:134`** — `if (this.peers.has(peer)) this.peers.set(peer, peerHits)`. Guarded, so it is only ever a rewrite of a pruned window for a caller already tracked. It cannot insert a key. The `has` also closes the `peer.limit === 0` edge, where a bare `set` would have landed an unseen caller in the map on their first refusal.
- **`:148`** — the allow branch, followed immediately by `evictIfCrowded()` at `:150`.

The `this.peers.set(peer, peerHits)` that used to sit at old line 110 is **gone from the global-refusal branch entirely** (`:139-146`): that branch now writes only `this.globalHits` and calls `evictIfCrowded()` at `:145`. The peer branch evicts at `:135`. So every path out of `check()` passes through the eviction pass, and — as the new doc comment at `:122-125` says — either mechanism alone would hold the bound today. Belt and braces on an invariant that already failed once is the right call here.

`evictIfCrowded` now trims to `EVICTION_TARGET_PEERS` (`:74`, = 4096 − 512) at `:186` and `:192` rather than to the cap, which is what stops the scan re-arming on every subsequent insert.

## The measurements

I re-ran the **exact probe from round one** — 50,000 distinct source addresses against a saturated global window, real defaults — plus four adversarial variants:

```
A. r1's probe, global window saturated
   scopes: global | tracked before spray: 6 | after 50k: 6 | peak: 6   (cap 4096)
B. allow path, 50k distinct addresses, global never saturated
   tracked after 50k: 3830 | peak: 4096                                (cap 4096)
C. 40 rounds alternating saturate → spray 2k → drain → allowed traffic
   peak across all rounds: 4096 | tracked at end: 3620
D. peer limit 0, 10k refusals of never-seen addresses
   tracked: 0
E. 120k randomised calls, invariant checked after every single call
   peak: 4096 | violations: 0
```

Round one reported **50,006 tracked against a cap of 4,096** on probe A. It is now **6** — unchanged from before the spray, which is stronger than the bound and is the real fix: a refused caller is never recorded at all. Probe E is the one I care most about, since it checks the invariant after every call under randomised drive across both windows rather than only at the end.

The amortisation is real too: r1's single 50k run took **53s**; all five probes above (230k+ calls) run in **1.9s** total.

## The test now exercises the branch

`Number.MAX_SAFE_INTEGER` is gone from `core-pairing-rate-limit.test.ts:137` — that test runs on `DEFAULT_GLOBAL_WINDOW` and now asserts `.ok === true` on every call at `:141`, which is the part I like most: it pins the test to the insert path explicitly, so if a future edit makes that loop start saturating, the test fails loudly instead of silently degrading into the weaker assertion it was before. The new case at `:151` covers the attacker's branch directly, and asserts `trackedPeers()` is **exactly** what it was before the spray rather than merely under the cap.

**I checked these fail against the old implementation rather than taking that on trust.** Restoring `021abb1`'s `check()` and eviction logic in an isolated tree, keeping only the new constants:

```
× holds the bound once the global window is saturated, which is the attacker's branch
    AssertionError: expected 8198 to be less than or equal to 4096
× does not record a caller the peer window refuses without ever having admitted
    AssertionError: expected 1 to be +0
```

Same two failures and the same numbers you reported. They are genuine regression tests. For completeness: the other two new cases (`:176`, `:202`) pass against the old code as well — they are property tests rather than regression tests, which is fine and matches how you described them.

## Gates at `d570575`

| Gate | Result |
|---|---|
| CI on this SHA | **18 success, 5 skipped, 0 failures** (run `32411696277`) |
| `@actana/core` | **817 passed** (was 813 — the four new cases) |
| `@actana/shared` | 384 passed |
| `pnpm typecheck` | clean across all five packages |
| `pnpm lint` | 0 errors, 46 warnings — same count as `021abb1`, none new and none in either changed file |

Only the two files I flagged changed since `021abb1` (+143/−9); the other 22 are byte-identical to what I reviewed on the merits last round, so every other acceptance-criterion verdict from that review stands unchanged. Base is still `feat/short-code-pairing` and the merge base is still its tip, so nothing needs rebasing.

## Acceptance criteria

All thirteen of #282 now **met**. AC 3 (rate limiting independent of the per-session attempt cap) was the one I marked partially met; its guessing defence was always sound and its memory bound now holds on the branch an attacker drives.

## On the three non-blocking notes

Keep them out of this PR. A follow-up issue on the train is the better home — this PR is already eight commits of pre-auth surface plus a fix, and the three are independent of it and of each other. My read on priority, since you asked: `core-first-run.ts:161` first, because it is the only one that can stop a daemon booting; then `pty-core-link-server.ts:2180`; the two clocks in `core-pairing-routes.ts` last, being a legibility issue in the audit trail rather than a behaviour one. None of them needs to hold up #283 or #284.

## Verdict

**Approve.** The finding is settled, proved from the enforcing lines and by measurement, and the regression tests fail on the old code.

Merge order for this wave — it is a genuine one-PR wave, #295 being the only open pull request on the repository:

1. Mark #295 ready for review (it is still a draft, and that is the only thing standing between it and merge).
2. Squash-merge #295 into `feat/short-code-pairing`, using the PR title as the commit message.
3. Nothing else merges in this wave. #283 and #284 unblock afterwards; `feat/short-code-pairing` → `beta/0.4.0` → `main` remains the train path, with `main` reached only through `promote.yml`.

Issue #286, opened 2026-08-20
Pairing: Panel pairing UI — add-core takes address plus code and shows fingerprint state


mrfqcentic:
Part of #280 — short-code pairing enrollment replaces the registration-blob hand-carry. Read that issue first; the design there is fixed.

## Context

The Panel's add-Core flow today is a textarea reading "paste the pairing token here…" (`packages/panel/src/components/views/CoresSettingsPage.tsx`), which decodes a base64 **Registration blob**, puts endpoint and label in the Core registry and the secret fields into the encrypted store (`packages/panel/src/server/services/secrets-at-rest.ts`, `cores.ts`).

This task replaces that paste with **address plus code**, calling the same task 4 SDK function the CLI calls. Nothing downstream changes: the credential the SDK returns is `CoreRegistrationBlob`-shaped, so it goes into the Core registry and the encrypted store exactly as a decoded blob does today, and `core-link-manager.ts` dials it unchanged.

Two Panel-specific facts shape this. First, the pairing call is Node-side work — key generation and a TLS dial — so it belongs on the **Panel server**, next to the other Core-side services, not in the renderer; the renderer collects the address, the code and the fingerprint confirmation and posts them. This is the same division the file surface already runs on: the Panel server holds the mTLS credentials because no browser can present a client certificate (CONTEXT.md "Dumb pipe", ADR 0030). Second, the Panel is the one client that can *show* the fingerprint rather than make the operator type it — a two-step flow where the presented fingerprint is displayed for the operator to compare against what `actana pair new` printed, and only then is the code sent.

The user-facing string "pairing token" currently names the Registration blob (CONTEXT.md's UI note). After this train that phrase means the short code, and the blob is gone — keep the two straight in every string this PR touches, and expect task 7 to settle the vocabulary in CONTEXT.md.

## Acceptance criteria

- [ ] The add-Core flow takes a Core address and an `XXXX-XXXX` pairing code instead of a blob paste; the code is accepted hyphenated or not and case-insensitively
- [ ] Fingerprint state is shown: the presented CA fingerprint is displayed for the operator to compare against what `actana pair new` printed, and the code is **not** sent until they confirm — verified, mismatched and not-yet-checked are three visibly distinct states
- [ ] A mismatch is a refusal with the expected and presented fingerprints shown, not a warning the operator can click past
- [ ] The pairing call runs on the Panel server and calls the same task 4 SDK function the CLI does — no second implementation, no crypto in the renderer
[...]

PR #296, opened 2026-08-20
feat(sdk): client pairing — keypair, CSR, fingerprint-verified dial and code redemption (#284)

### LINKED ISSUE #284: Pairing: SDK client pairing — keypair, CSR, fingerprint-verified dial and code redemption

qcentic-dev:
Closes #284.

The client half of short-code pairing (#280), in `packages/sdk` — the package a third party installs to talk to a Core, and the one function both `actana core pair` (#285) and the Panel's first-contact UI (#286) will call.

**Base is `feat/short-code-pairing`**, not `main` and not the train. The pairing server this drives is #282, already on that branch.

## What lands

- `packages/sdk/src/core-pairing.ts` — the wire types, the bootstrap dial, the fingerprint comparison, the redemption, and a failure union a caller can switch on.
- `packages/sdk/src/core-pairing-csr.ts` — a key pair from `node:crypto` and a PKCS#10 CSR built over it, with no new dependency.
- `packages/sdk/src/__tests__/core-pairing.test.ts` — the flow against the Core's own server, its own routes and its own certificates.
- `packages/sdk/src/__tests__/package-boundaries.test.ts` — the sweep now allows `node:`-prefixed builtins, and still refuses bare `crypto`.

Nothing outside `packages/sdk` is touched: the CLI dispatch and the link server are #283's and #285's.

## The shape

```ts
const identity = await fetchCorePairingIdentity({ address: "core.example:9443" });
// → { fingerprint: "75:5C:…:E8", caCert, host, port, httpsOrigin }  — no code sent

const blob = await pairWithCore({
  address: "core.example:9443",
  sessionId: "ps_7f3a1c20",
  code: "ABCD-EFGH",
  expectedCaFingerprint: "75:5C:…:E8",
  label: "laptop",
});
const connection = coreConnectionFromBlob(blob); // no conversion
```

## The security argument, and where it is asserted

1. The bootstrap dial is unverified because at that point there is nothing to verify against — that is what the operator's fingerprint is for. It is the **only** `rejectUnauthorized: false` in the package, and a test sweeps the shipped sources to keep it that way.
2. The comparison happens with nothing secret sent. On a mismatch the test asserts from the *server's* side that the endpoint was never reached, no audit line was written, and no attempt was spent on the operator's session.
3. The redemption is a second connection pinned to the CA that matched — `ca` set to that one certificate, `rejectUnauthorized: true`, and a `checkServerIdentity` that re-runs the hostname check and the fingerprint comparison before the handshake completes. A Core that swaps its certificate between the two connections gets no request at all; that test stages exactly that.
4. The CA in the response is held to the same fingerprint, so the blob never pins something no human read out.
5. The private key is generated locally and put back into the blob after the response arrives. The suite records the request bytes and asserts the key is not among them.

## Two things worth a reviewer's attention

**The session id.** The Core requires `sessionId` beside the code and refuses to search for the session a code might belong to (#282, session binding). #283's `actana pair new` prints "the code, the CA fingerprint and the expiry" — with no session id in that list, and #284 describes the client function as taking "the address, the code and the fingerprint". Those two do not currently meet. Rather than settle it here, `pairWithCore` accepts either form — an explicit `sessionId`, or a single `"<sessionId>:<XXXX-XXXX>"` code that carries it — so whichever way #280 settles it, #283 and #285 change and this module does not. **This is the coordination #284 asks to happen in #280 rather than in a PR review.**

**Four refusals arrive as one.** `wrong / expired / consumed / attempts-exhausted` reach the client as a single `refused` failure, because the Core answers all four with one status and one body deliberately — telling them apart on the wire is the leak #282 refuses to ship. Rate-limited, fingerprint-mismatch, unreachable, not-pairable, rejected, core-error and malformed-response are each their own. See the acceptance-criteria comment below.

## Checks

`pnpm --filter @actana/sdk test` (270 passed, 5 skipped), `typecheck`, `eslint packages/sdk`, and the published build (`tsc -p tsconfig.build.json`) are green locally.

https://claude.ai/code/session_01P2TVDf5cTzELMEasHVANmZ

### DIFF
--- a/packages/sdk/src/__tests__/core-pairing.test.ts
+++ b/packages/sdk/src/__tests__/core-pairing.test.ts
@@ -29,7 +29,7 @@ import * as path from "node:path";
 import type { IncomingMessage, ServerResponse } from "node:http";
 import { X509Certificate, createPublicKey } from "node:crypto";
 import { afterEach, describe, expect, it } from "vitest";
-import { generateCertMaterial } from "@actana/shared/core-cert-material";
+import { generateCertMaterial, issueServerCert } from "@actana/shared/core-cert-material";
 import { verifyBearer } from "@actana/shared/core-link-bearer";
 import { generatePairingCode } from "@actana/shared/pairing-code";
 import { createPairingSession } from "@actana/shared/pairing-session";
@@ -37,7 +37,6 @@ import { PairingStore, derivePairingCodeKey, hashPairingCode } from "@actana/sha
 import type { CoreHttpRoutes } from "@actana/core/core-files-routes";
 import { createCoreFilesRequestHandler } from "@actana/core/core-files-routes";
 import { PairingRateLimiter } from "@actana/core/core-pairing-rate-limit";
-import type { PairingAuditEvent } from "@actana/core/core-pairing-audit";
 import { buildCorePairingRoutes, composeCoreHttpRoutes, isPairingPath } from "@actana/core/core-pairing-wiring";
 import { PtyCoreLinkServer } from "@actana/core/pty-core-link-server";
 import type { PtyCore, PtyCoreEvent } from "@actana/core/pty-manager";
@@ -53,6 +52,18 @@ import {
 import { coreConnectionFromBlob, type CoreRegistrationBlob } from "../core-registration-blob";
 import { createNodeCoreLinkSocket } from "../core-link-socket";
 
+/**
+ * The audit line, as this suite reads it — one field, and it is the one #282
+ * refuses to put on the wire.
+ *
+ * Declared rather than imported from the Core. #297 moves that module to
+ * `packages/shared`, and the two branches merge cleanly: git would report no
+ * conflict while leaving an import of a file that no longer exists, and because
+ * it was an `import type` the vitest run would not have caught it either. A
+ * structural read of one field costs nothing and survives the move.
+ */
+type PairingAuditLine = { outcome: string; reason?: string };
+
 const SECRET = "sdk-pairing-suite-secret-at-least-32-bytes";
 const CORE_UUID = "9c1f4a5e-3f6d-0f0a-6c1f-1d0a5b7e9c31";
 
@@ -64,7 +75,7 @@ type Rig = {
   origin: string;
   caCert: string;
   fingerprint: string;
-  audit: PairingAuditEvent[];
+  audit: PairingAuditLine[];
   wire: WireLog;
   clock: { now: number };
   openSession(opts?: { label?: string; ttlMs?: number }): { sessionId: string; code: string };
@@ -141,7 +152,7 @@ async function startCore(opts: { rateLimiter?: PairingRateLimiter } = {}): Promi
   const dir = fs.mkdtempSync(path.join(os.tmpdir(), "actana-sdk-pairing-"));
   tempDirs.push(dir);
   const store = new PairingStore(path.join(dir, "pairing.json"));
-  const audit: PairingAuditEvent[] = [];
+  const audit: PairingAuditLine[] = [];
   const wire: WireLog = { requests: 0, bodies: [] };
   const clock = { now: Date.now() };
   const codeKey = derivePairingCodeKey(SECRET);
@@ -580,6 +591,19 @@ describe("every refusal a caller can act on is a different failure", () => {
     expect(parsePairingTicket("abcd efgh", "ps_7f3a")).toEqual({ sessionId: "ps_7f3a", code: "ABCD-EFGH" });
     // An explicit session id wins, and the code beside it is read as a code.
     expect(parsePairingTicket("ABCD-EFGH", "ps_other")).toEqual({ sessionId: "ps_other", code: "ABCD-EFGH" });
+    // Both at once — a caller that always forwards `--session` while an
+    // operator pastes whatever they were read out. This used to keep the
+    // `ps_7f3a:` prefix inside the code and refuse a perfectly good pairing.
+    expect(parsePairingTicket("ps_7f3a:ABCD-EFGH", "ps_7f3a")).toEqual({
+      sessionId: "ps_7f3a",
+      code: "ABCD-EFGH",
+    });
+    // And two that disagree are refused rather than resolved: one of them is a
+    // mistake, and redeeming against the wrong session fails in a way that
+    // looks exactly like a mistyped code.
+    const clash = (): unknown => parsePairingTicket("ps_7f3a:ABCD-EFGH", "ps_other");
+    expect(clash).toThrow(CorePairingError);
+    expect(clash).toThrow(/must agree/);
   });
 });
 
@@ -589,6 +613,8 @@ describe("every refusal a caller can act on is a different failure", () => {
 async function startStub(opts: {
   cert: { cert: string; key: string; ca: string };
   answer?: { status: number; body: string; headers?: Record<string, string> };
+  /** The loopback address to bind. `127.0.0.2` is a Core reached off its SAN. */
+  host?: string;
 }): Promise<{ address: string; requests: number; server: https.Server }> {
   const state = { requests: 0 };
   const stub = https.createServer({ cert: opts.cert.cert, key: opts.cert.key, ca: opts.cert.ca }, (req, res) => {
@@ -601,10 +627,11 @@ async function startStub(opts: {
     });
   });
   stubs.push(stub);
-  await new Promise<void>((resolve) => stub.listen(0, "127.0.0.1", () => resolve()));
+  const host = opts.host ?? "127.0.0.1";
+  await new Promise<void>((resolve) => stub.listen(0, host, () => resolve()));
   const port = (stub.address() as { port: number }).port;
   return {
-    address: `127.0.0.1:${port}`,
+    address: `${host}:${port}`,
     get requests() {
       return state.requests;
     },
@@ -651,6 +678,9 @@ describe("the redemption dial is pinned to the certificate authority that matche
     );
 
     expect(failure.failure).toBe<CorePairingFailure>("fingerprint-mismatch");
+    // The pin refusing, by name: OpenSSL finds the pinned CA by subject and the
+    // impostor's signature does not check out against it.
+    expect(failure.detail.tlsCode).toBe("CERT_SIGNATURE_FAILURE");
     expect(stub.requests).toBe(0);
   }, 30_000);
 
@@ -685,6 +715,70 @@ describe("the redemption dial is pinned to the certificate authority that matche
 
     expect(failure.failure).toBe<CorePairingFailure>("fingerprint-mismatch");
     expect(stub.requests).toBe(1);
+  }, 30_000);
+});
+
+describe("a certificate problem is not an accusation", () => {
+  it("tells a Core reached off its SAN from a Core that is not the right Core", async () => {
+    // A Core set up for one address and reached at another: a second interface,
+    // a tunnel, a DNS name added later. `core-cert-material.ts` covers the
+    // configured host plus loopback and nothing else, so the fingerprint
+    // matches perfectly and Node's hostname check still refuses.
+    //
+    // This used to be reported as `fingerprint-mismatch` — the operator was
+    // told they were being intercepted and went looking for an attacker.
+    const material = await generateCertMaterial({ host: "127.0.0.1" });
+    const stub = await startStub({
+      cert: { cert: material.server.cert, key: material.server.key, ca: material.ca.cert },
+      host: "127.0.0.2",
+      answer: { status: 200, body: "{}" },
+    });
+
+    const failure = await failureOf(
+      pairWithCore({
+        address: stub.address,
+        sessionId: "ps_1",
+        code: "ABCD-EFGH",
+        expectedCaFingerprint: fingerprintOf(new X509Certificate(material.ca.cert).raw),
+        timeoutMs: 5_000,
+      }),
+    );
+
+    expect(failure.failure).toBe<CorePairingFailure>("hostname-mismatch");
+    expect(failure.detail.tlsCode).toBe("ERR_TLS_CERT_ALTNAME_INVALID");
+    expect(failure.message).toContain("127.0.0.2");
+    // Said plainly, because the wrong sentence here costs an operator an
+    // afternoon: this is the CA they were told to expect.
+    expect(failure.message).toContain("presented the expected certificate authority");
+    expect(stub.requests).toBe(0);
+  }, 30_000);
+
+  it("reports an expired server certificate as a certificate problem, not a mismatch", async () => {
+    const ca = await generateCertMaterial({ host: "127.0.0.1" });
+    const stale = await issueServerCert({
+      ca: { cert: ca.ca.cert, key: ca.ca.key },
+      host: "127.0.0.1",
+      days: 1,
+      notBefore: new Date(Date.now() - 400 * 24 * 60 * 60 * 1000),
+    });
+    const stub = await startStub({
+      cert: { cert: stale.cert, key: stale.key, ca: ca.ca.cert },
+      answer: { status: 200, body: "{}" },
+    });
+
+    const failure = await failureOf(
+      pairWithCore({
+        address: stub.address,
+        sessionId: "ps_1",
+        code: "ABCD-EFGH",
+        expectedCaFingerprint: fingerprintOf(new X509Certificate(ca.ca.cert).raw),
+        timeoutMs: 5_000,
+      }),
+    );
+
+    expect(failure.failure).toBe<CorePairingFailure>("certificate-invalid");
+    expect(failure.detail.tlsCode).toBe("CERT_HAS_EXPIRED");
+    expect(stub.requests).toBe(0);
   }, 30_000);
 });
 
@@ -734,7 +828,35 @@ describe("answers that are not a Core's", () => {
     const incomplete = await failureAgainst({ status: 200, body: JSON.stringify({ endpoint: "wss://x:1" }) });
     expect(incomplete.failure).toBe<CorePairingFailure>("malformed-response");
     expect(incomplete.message).toContain("caCert");
+
+    // `JSON.parse("null")` succeeds, and so does an array. Both used to reach
+    // the field sweep, where `null` threw a raw `TypeError` past the failure
+    // union every caller of this module switches on.
+    const nulled = await failureAgainst({ status: 200, body: "null" });
+    expect(nulled.failure).toBe<CorePairingFailure>("malformed-response");
+
+    const listed = await failureAgainst({ status: 200, body: "[]" });
+    expect(listed.failure).toBe<CorePairingFailure>("malformed-response");
   }, 60_000);
+
+  it("refuses a credential for a plaintext endpoint", async () => {
+    // `coreConnectionFromBlob` reads TLS off the scheme: a `ws://` endpoint
+    // yields `tls: null` and still carries the bearer, so accepting one here
+    // would hand back a credential that ships the bearer in cleartext on every
+    // later dial — undoing the whole of what the pinned exchange bought.
+    const failure = await failureAgainst({
+      status: 200,
+      body: JSON.stringify({
+        endpoint: "ws://127.0.0.1:9443",
+        caCert: "-----BEGIN CERTIFICATE-----\nx\n-----END CERTIFICATE-----",
+        clientCert: "-----BEGIN CERTIFICATE-----\nx\n-----END CERTIFICATE-----",
+        bearer: "b",
+      }),
+    });
+
+    expect(failure.failure).toBe<CorePairingFailure>("malformed-response");
+    expect(failure.message).toContain("ws://127.0.0.1:9443");
+  }, 30_000);
 });
 
 describe("nothing this package ships stays unverified", () => {
--- a/packages/sdk/src/__tests__/core-pairing.test.ts
+++ b/packages/sdk/src/__tests__/core-pairing.test.ts
@@ -1,0 +1,780 @@
+// SDK pairing, against a real Core (#284).
+//
+// Everything the client claims here is a claim about an exchange, so almost
+// nothing below is stubbed: the server is the Core's own `PtyCoreLinkServer`
+// with the Core's own pairing routes mounted through its own wiring, the
+// certificates come from `generateCertMaterial` (what `actana setup` writes),
+// the sessions go into the `PairingStore` the operator's `actana pair new` will
+// write, and the client is `pairWithCore` with nothing patched underneath it.
+//
+// Two things are staged rather than real, and both are staged because the
+// property under test cannot be observed otherwise:
+//
+//   • **Request bodies are recorded** by a route family wrapped around the
+//     pairing one. "The private key never appears in any request body" and "the
+//     code is not sent when the fingerprint does not match" are both statements
+//     about bytes on the wire, and a client-side assertion would be the client
+//     grading its own homework.
+//   • **A Core that changes its certificate between the bootstrap dial and the
+//     redemption** is an https server whose secure context is swapped after the
+//     first handshake. No real Core does that; an attacker in the middle of one
+//     is exactly that shape, and it is the only way to prove the redemption
+//     dial is pinned rather than merely polite.
+
+import * as fs from "node:fs";
+import * as https from "node:https";
+import { Server } from "node:net";
+import * as os from "node:os";
+import * as path from "node:path";
+import type { IncomingMessage, ServerResponse } from "node:http";
+import { X509Certificate, createPublicKey } from "node:crypto";
+import { afterEach, describe, expect, it } from "vitest";
+import { generateCertMaterial } from "@actana/shared/core-cert-material";
+import { verifyBearer } from "@actana/shared/core-link-bearer";
+import { generatePairingCode } from "@actana/shared/pairing-code";
+import { createPairingSession } from "@actana/shared/pairing-session";
+import { PairingStore, derivePairingCodeKey, hashPairingCode } from "@actana/shared/pairing-store";
+import type { CoreHttpRoutes } from "@actana/core/core-files-routes";
+import { createCoreFilesRequestHandler } from "@actana/core/core-files-routes";
+import { PairingRateLimiter } from "@actana/core/core-pairing-rate-limit";
+import type { PairingAuditEvent } from "@actana/core/core-pairing-audit";
+import { buildCorePairingRoutes, composeCoreHttpRoutes, isPairingPath } from "@actana/core/core-pairing-wiring";
+import { PtyCoreLinkServer } from "@actana/core/pty-core-link-server";
+import type { PtyCore, PtyCoreEvent } from "@actana/core/pty-manager";
+import {
+  CORE_PAIRING_REDEEM_PATH,
+  CorePairingError,
+  fetchCorePairingIdentity,
+  fingerprintOf,
+  pairWithCore,
+  parsePairingTicket,
+  type CorePairingFailure,
+} from "../core-pairing";
+import { coreConnectionFromBlob, type CoreRegistrationBlob } from "../core-registration-blob";
+import { createNodeCoreLinkSocket } from "../core-link-socket";
+
+const SECRET = "sdk-pairing-suite-secret-at-least-32-bytes";
+const CORE_UUID = "9c1f4a5e-3f6d-0f0a-6c1f-1d0a5b7e9c31";
+
+/** What the pairing endpoint was actually sent, byte for byte. */
+type WireLog = { requests: number; bodies: string[] };
+
+type Rig = {
+  address: string;
+  origin: string;
+  caCert: string;
+  fingerprint: string;
+  audit: PairingAuditEvent[];
+  wire: WireLog;
+  clock: { now: number };
+  openSession(opts?: { label?: string; ttlMs?: number }): { sessionId: string; code: string };
+  store: PairingStore;
+};
+
+let server: PtyCoreLinkServer | null = null;
+const stubs: https.Server[] = [];
+const tempDirs: string[] = [];
+
+afterEach(() => {
+  server?.close();
+  server = null;
+  while (stubs.length > 0) stubs.pop()!.close();
+  while (tempDirs.length > 0) fs.rmSync(tempDirs.pop()!, { recursive: true, force: true });
+});
+
+function mockCore(): PtyCore {
+  return {
+    setEmitTarget: (_fn: ((event: PtyCoreEvent) => void) | null) => {},
+    spawn: async () => ({ ptyId: "pty-1", hooksReportTurnStart: true }),
+    write: () => true,
+    resize: () => true,
+    kill: () => true,
+    killLaunchProcesses: async () => ({ ptyCount: 0, ports: [] }),
+    findByTask: () => ({ ptyId: null }),
+    replay: () => ({ data: "", nextSeq: 0, from: 0 }),
+    killAll: () => {},
+  } as unknown as PtyCore;
+}
+
+/** A free TCP port on 127.0.0.1, found by briefly binding port 0. */
+function freePort(): Promise<number> {
+  return new Promise((resolve, reject) => {
+    const probe = new Server();
+    probe.on("error", reject);
+    probe.listen(0, "127.0.0.1", () => {
+      const address = probe.address();
+      if (address && typeof address === "object") {
+        const { port } = address;
+        probe.close(() => resolve(port));
+      } else {
+        probe.close();
+        reject(new Error("no port"));
+      }
+    });
+  });
+}
+
+/**
+ * Copy every byte posted to the pairing prefix, then hand the request on.
+ *
+ * The `data` listener is attached in the same tick as the real handler's — the
+ * route reads its body synchronously from `handle` — so flowing mode delivers
+ * every chunk to both and nothing is stolen from the server underneath.
+ */
+function recording(inner: CoreHttpRoutes, wire: WireLog): CoreHttpRoutes {
+  const tee = (fn: (req: IncomingMessage, res: ServerResponse) => boolean) =>
+    (req: IncomingMessage, res: ServerResponse): boolean => {
+      if ((req.url ?? "").startsWith("/v1/pair/")) {
+        wire.requests += 1;
+        const chunks: Buffer[] = [];
+        req.on("data", (chunk: Buffer) => chunks.push(chunk));
+        req.on("end", () => wire.bodies.push(Buffer.concat(chunks).toString("utf8")));
+      }
+      return fn(req, res);
+    };
+  return { handle: tee(inner.handle), handleContinue: tee(inner.handleContinue) };
+}
+
+async function startCore(opts: { rateLimiter?: PairingRateLimiter } = {}): Promise<Rig> {
+  const material = await generateCertMaterial({ host: "127.0.0.1" });
+  const port = await freePort();
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), "actana-sdk-pairing-"));
+  tempDirs.push(dir);
+  const store = new PairingStore(path.join(dir, "pairing.json"));
+  const audit: PairingAuditEvent[] = [];
+  const wire: WireLog = { requests: 0, bodies: [] };
+  const clock = { now: Date.now() };
+  const codeKey = derivePairingCodeKey(SECRET);
+
+  const pairingRoutes = buildCorePairingRoutes({
+    material: {
+      caCert: material.ca.cert,
+      caKey: material.ca.key,
+      bearerSecret: SECRET,
+      coreId: "core_pairing",
+      coreUuid: CORE_UUID,
+    },
+    sessions: store,
+    endpoint: `wss://127.0.0.1:${port}`,
+    now: () => clock.now,
+    audit: (event) => audit.push(event),
+    ...(opts.rateLimiter ? { rateLimiter: opts.rateLimiter } : {}),
+  });
+
+  const fileRoutes = createCoreFilesRequestHandler({
+    filesPort: { projectRoot: () => null },
+    authVerifier: (bearer) => verifyBearer(bearer, SECRET),
+  });
+
+  server = new PtyCoreLinkServer(mockCore(), {
+    port,
+    host: "127.0.0.1",
+    tls: {
+      caCert: material.ca.cert,
+      serverCert: material.server.cert,
+      serverKey: material.server.key,
+    },
+    authVerifier: (bearer) => verifyBearer(bearer, SECRET),
+    httpRoutes: composeCoreHttpRoutes(recording(pairingRoutes, wire), fileRoutes),
+    isPreAuthPath: isPairingPath,
+  });
+
+  const rig: Rig = {
+    address: `127.0.0.1:${port}`,
+    origin: `https://127.0.0.1:${port}`,
+    caCert: material.ca.cert,
+    fingerprint: fingerprintOf(new X509Certificate(material.ca.cert).raw),
+    audit,
+    wire,
+    clock,
+    store,
+    openSession: ({ label = "laptop", ttlMs } = {}) => {
+      const code = generatePairingCode();
+      const sessionId = `ps_${Math.random().toString(16).slice(2, 10)}`;
+      store.createSession(
+        createPairingSession({
+          id: sessionId,
+          label,
+          codeHash: hashPairingCode({ key: codeKey, sessionId, code }),
+          now: clock.now,
+          ...(ttlMs === undefined ? {} : { ttlMs }),
+        }),
+        clock.now,
+      );
+      return { sessionId, code };
+    },
+  };
+
+  // Readiness is observed rather than awaited, as the Core's own suite does —
+  // and on a route that is not the pairing one, so no probe spends a rate-limit
+  // attempt or writes an audit line before a test has started.
+  const deadline = Date.now() + 10_000;
+  for (;;) {
+    try {
+      await new Promise<void>((resolve, reject) => {
+        const req = https.request(
+          { host: "127.0.0.1", port, path: "/healthz", method: "GET", ca: material.ca.cert, agent: false },
+          (res) => {
+            res.resume();
+            res.on("end", () => resolve());
+          },
+        );
+        req.on("error", reject);
+        req.end();
+      });
+      return rig;
+    } catch (err) {
+      if (Date.now() > deadline) throw err;
+      await new Promise((resolve) => setTimeout(resolve, 25));
+    }
+  }
+}
+
+/** The pairing failure of a call that must not have produced a blob. */
+async function failureOf(promise: Promise<unknown>): Promise<CorePairingError> {
+  try {
+    await promise;
+  } catch (err) {
+    if (err instanceof CorePairingError) return err;
+    throw err;
+  }
+  throw new Error("expected the pairing attempt to fail, and it did not");
+}
+
+/** Dial the core link with a blob and report the frames the Core answers with. */
+function coreLinkAuth(blob: CoreRegistrationBlob): Promise<string[]> {
+  const connection = coreConnectionFromBlob(blob);
+  return new Promise((resolve, reject) => {
+    const seen: string[] = [];
+    const socket = createNodeCoreLinkSocket(connection.url, connection.tls ?? undefined);
+    const timer = setTimeout(() => {
+      socket.close();
+      reject(new Error(`no authOk — saw ${seen.join(", ") || "nothing"}`));
+    }, 10_000);
+    socket.on("open", () => socket.send(JSON.stringify({ type: "auth", reqId: "a1", bearer: connection.bearer })));
+    socket.on("message", (data: unknown) => {
+      const frame = JSON.parse(String(data)) as { type: string };
+      seen.push(frame.type);
+      if (frame.type !== "authOk" && frame.type !== "authError") return;
+      clearTimeout(timer);
+      socket.close();
+      resolve(seen);
+    });
+    socket.on("error", (err: Error) => {
+      clearTimeout(timer);
+      reject(err);
+    });
+  });
+}
+
+/** The public key inside a PEM certificate, as DER, for comparing to a key. */
+function certificatePublicKey(certPem: string): string {
+  return new X509Certificate(certPem).publicKey.export({ type: "spki", format: "der" }).toString("base64");
+}
+
+function privateKeyPublicHalf(keyPem: string): string {
+  return createPublicKey(keyPem).export({ type: "spki", format: "der" }).toString("base64");
+}
+
+describe("a client with a code pairs with a Core it has verified", () => {
+  it("returns a blob assembled from the response and the key that never moved", async () => {
+    const rig = await startCore();
+    const { sessionId, code } = rig.openSession({ label: "laptop" });
+
+    const blob = await pairWithCore({
+      address: rig.address,
+      sessionId,
+      code,
+      expectedCaFingerprint: rig.fingerprint,
+      label: "laptop",
+      platform: "linux",
+    });
+
+    expect(blob.endpoint).toMatch(/^wss:\/\/127\.0\.0\.1:\d+$/);
+    expect(blob.caCert).toBe(rig.caCert);
+    expect(blob.clientCert).toContain("BEGIN CERTIFICATE");
+    expect(blob.clientKey).toContain("BEGIN PRIVATE KEY");
+    expect(blob.bearer.length).toBeGreaterThan(0);
+
+    // The certificate the Core signed is a certificate *for the key this
+    // machine generated* — which is the whole of what a CSR buys, and the only
+    // proof that the `clientKey` in the blob is the returned credential's other
+    // half rather than a key that merely came back in the same object.
+    expect(certificatePublicKey(blob.clientCert)).toBe(privateKeyPublicHalf(blob.clientKey));
+
+    // The bearer is the Core's, and it verifies against the Core's secret.
+    expect(verifyBearer(blob.bearer, SECRET)).not.toBeNull();
+  }, 30_000);
+
+  it("sends the CSR and never the private key", async () => {
+    const rig = await startCore();
+    const { sessionId, code } = rig.openSession();
+
+    const blob = await pairWithCore({
+      address: rig.address,
+      sessionId,
+      code,
+      expectedCaFingerprint: rig.fingerprint,
+      label: "laptop",
+    });
+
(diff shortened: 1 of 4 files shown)


qcentic-dev:
## Acceptance criteria, #284

Nine of ten met. One is marked unchecked with its reason rather than reinterpreted to fit.

- [x] **A new module in `packages/sdk/src/` (suggested `core-pairing.ts`) exporting one pairing function that takes the Core address, the pairing code and the expected CA fingerprint, and returns a `CoreRegistrationBlob`** — `packages/sdk/src/core-pairing.ts`, `pairWithCore({ address, code, expectedCaFingerprint, … })`. A second file, `core-pairing-csr.ts`, holds the key pair and the PKCS#10 encoder; the pairing surface is all in the module the issue names. One caveat that is a coordination item, not a gap: the Core also requires the `sessionId` the code belongs to and will not search for it (#282 session binding), while #283's `actana pair new` prints only the code, the fingerprint and the expiry. Both forms parse here — an explicit `sessionId`, or a `"<sessionId>:<XXXX-XXXX>"` code that carries it — so #280 can settle which the operator is given without this module changing. Flagged in the PR body.
- [x] **Keypair generation happens locally via `node:crypto`; a PEM CSR is produced from it carrying the client label/info. The private key is never sent, logged, or included in any request body** — `generateClientCsr` in `core-pairing-csr.ts`: `generateKeyPair("rsa", { modulusLength: 2048 })`, subject `CN=<label>`, and the platform travels in the request's `client` object. Asserted from the wire, not from the client: the suite records the bytes the endpoint received and greps them for `PRIVATE KEY` and for every long line of the returned key ("sends the CSR and never the private key").
- [x] **The bootstrap dial obtains the presented certificate chain and computes the CA fingerprint as SHA-256 over DER, colon-separated uppercase hex — the same format task 3 prints** — `presentedChain` walks `issuerCertificate` to the self-issued root and `fingerprintOf` hashes its DER. A test pins the format (`/^([0-9A-F]{2}:){31}[0-9A-F]{2}$/`) and cross-checks the value against Node's own `X509Certificate.fingerprint256`, which is an independent digest of the same certificate. #283 is not on this branch yet, so this is the format agreed in its issue text rather than a comparison against its code.
- [x] **Fingerprint mismatch aborts before the code is transmitted, with an error that names both the expected and the presented fingerprint** — `CorePairingError` with `failure: "fingerprint-mismatch"`, both fingerprints in the message and in `detail.expectedFingerprint` / `detail.presentedFingerprint`.
- [x] **A mode that reports the presented fingerprint back to the caller for confirmation without sending the code, for a first-contact UI (task 6) that wants to show it** — `fetchCorePairingIdentity({ address })` returns `{ fingerprint, caCert, host, port, httpsOrigin }` and takes no code, so it cannot leak one. `pairWithCore` called without `expectedCaFingerprint` refuses with `fingerprint-unconfirmed` carrying the same fingerprint and the presented CA, having sent nothing.
- [x] **No code path leaves `rejectUnauthorized: false` in place for anything after the fingerprint check — once the CA is verified, the rest of the exchange and every subsequent dial pin it** — the redemption is a second connection with `ca` set to the certificate that matched, `rejectUnauthorized: true`, and a `checkServerIdentity` that re-runs Node's hostname check and the fingerprint comparison before the handshake completes. Every later dial pins `blob.caCert` through the untouched `coreConnectionFromBlob` path. Asserted three ways: a source sweep over every shipped module allowing exactly one relaxed dial and requiring it to be the bootstrap one; a staged Core that swaps its certificate between the two connections and receives no request; and a refusal when the CA in the response is not the CA that was presented.
- [x] **Request/response wire types declared in the SDK, structurally identical to the Core's, with a comment giving the ADR 0025 D1/D4 reasoning in the manner of `core-registration-blob.ts`** — `CorePairingRedeemRequest`, `CorePairingRedeemResponse`, `CorePairingClientInfo`, `CorePairingRefusalBody` and `CORE_PAIRING_REDEEM_PATH`, declared in `core-pairing.ts` with the argument written out in the module header. Nothing in the shipped sources imports `@actana/shared`; the boundary sweep still enforces that.
- [x] **The returned object is assembled from the response (`endpoint`, `caCert`, `clientCert`, `bearer`) plus the locally-born `clientKey`, and passes straight into `coreConnectionFromBlob` with no conversion** — and the test does exactly that, then dials the core link with what comes out and gets `authOk`. The key is proved to be the credential's other half by comparing the certificate's SPKI with the private key's.
- [ ] **Errors are distinguishable by the caller — wrong/expired/consumed code, attempts exhausted, rate limited, fingerprint mismatch, unreachable — so the CLI and Panel can say something useful** — **unchecked, and deliberately.** `rate-limited` (with `retryAfterSeconds`), `fingerprint-mismatch`, `fingerprint-unconfirmed`, `unreachable`, `not-pairable`, `rejected`, `core-error`, `malformed-response`, `bad-address`, `bad-code` and `bad-fingerprint` are each their own `CorePairingFailure`. **Wrong, expired, consumed and attempts-exhausted are not**, and cannot be from here: the Core answers all four with one status and one body on purpose — "every refusal is the same refusal" in `core-pairing-routes.ts`, because a response that said which would tell an attacker whether a session exists and whether their last guess was closer. They arrive as one `refused` failure whose message names the four possibilities, and the distinction lives in the Core's audit log, where #282 put it. Splitting them would require changing #282's threat model, which is out of scope for this ticket. The tests assert the merge is the Core's rather than this client's: each of the four is driven end to end, and the audit log is checked for the reason the Core recorded while the client saw `refused`.
- [x] **Tests: happy path against a real in-process Core pairing server (task 2); fingerprint mismatch aborts pre-send, asserted by the server never seeing a request; the returned `clientKey` matches the locally generated key and never appears in any request body; the returned blob completes an mTLS handshake and `auth` frame through `coreConnectionFromBlob`; each error case surfaces distinguishably** — 18 tests in `packages/sdk/src/__tests__/core-pairing.test.ts`, against the Core's own `PtyCoreLinkServer` with its own pairing routes, its own wiring, `generateCertMaterial` and a real `PairingStore`. The mismatch case asserts three things from the server's side: no request, no audit line, and no attempt spent. "Each error case surfaces distinguishably" carries the caveat above — each error the *Core distinguishes* surfaces distinguishably.

**Not in this PR, by the issue's own boundary:** `packages/cli` (the `actana core pair` dispatch, #285) and `packages/core` (the link server and the operator verbs, #283). Nothing outside `packages/sdk` is touched.

Local checks: `pnpm --filter @actana/sdk test` — 270 passed, 5 skipped; `tsc --noEmit`; `eslint packages/sdk`; `tsc -p tsconfig.build.json`. All three commits pass `commitlint` with this repo's config.

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,798 @@
...
+ * did not recognise would be reported as `unreachable`, which is the wrong
+ * sentence for the one case that matters most.
+ */
+function certificateFailure(err: unknown): boolean {
+  const code = String((err as { code?: unknown } | null)?.code ?? "");
+  if (code === PIN_FAILURE_CODE) return true;
+  return (
+    code.startsWith("ERR_TLS_") ||
+    code.startsWith("ERR_SSL_") ||
+    code.includes("CERT") ||
```
mrfqcentic, line 859: **Blocking.** This prefix/substring match is too wide, and the consequence lands on the one message that has to be exact.

`code.includes("CERT")` matches `ERR_TLS_CERT_ALTNAME_INVALID`, so a hostname failure on the redemption dial reaches line 656 as `certificateFailure(err) === true` and is reported as `fingerprint-mismatch`: *"did not verify against the certificate authority whose fingerprint was confirmed"*. That sentence says the Core presented a different CA. It did not — the CA matched on both dials.

This is reachable on an ordinary configuration, and it is reachable *because* of a deliberate choice above: `checkServerIdentity` at line 619 re-runs Node's hostname check first and returns its error. Per `core-cert-material.ts` a Core's server certificate covers only the configured host plus `localhost`/`127.0.0.1`, so a Core set up for one address and reached at another — a second NIC, a tunnel, a DNS name added later — fails the name check with the fingerprint matching perfectly. The operator is told they are being MITM'd and goes looking for an attacker. The same happens for an expired server certificate.

The genuine pin refusal is already separable: `pinFailure` tags its error `ERR_ACTANA_PAIRING_PIN`, and line 769 checks that first. So `fingerprint-mismatch` can be reserved for the tagged refusal (plus the response-CA check at line 364), with a distinct `certificate-invalid` / `hostname-mismatch` failure for the rest — which is what a caller needs anyway to write the sentence that actually helps.

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,798 @@
...
+
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(answer.body);
+  } catch {
+    throw new CorePairingError("malformed-response", `${origin} answered 200 with something that was not JSON`, {
+      status: answer.status,
+    });
+  }
+  const fields = parsed as Partial<Record<keyof CorePairingRedeemResponse, unknown>>;
```
mrfqcentic, line 759: **Blocking.** `JSON.parse("null")` succeeds and returns `null`, so `parsed` reaches this line as `null`, the cast makes it look like an object to the compiler, and `typeof fields[field]` on the next line throws `TypeError: Cannot read properties of null (reading 'endpoint')`.

That escapes this function's whole purpose. The branch above exists to turn a bad 200 into `malformed-response`, and `pairWithCore`'s doc comment promises "Throws `CorePairingError` for everything that is not a blob" — so a caller doing `catch (err) { if (err instanceof CorePairingError) switch (err.failure) … }`, which is the pattern this module's failure union is designed around, falls straight through to an unhandled `TypeError`.

`safeRefusal` at line 782 already guards exactly this way (`parsed && typeof parsed === "object"`); the same check belongs here before the field sweep.

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,798 @@
...
+ * D3 is about. What the shape check buys is worth having on its own: a code
+ * that could not be right whatever the alphabet is never spends one of the
+ * five attempts the operator's session has.
+ */
+export function parsePairingTicket(input: string, sessionId?: string): PairingTicket {
+  const trimmed = input.trim();
+  const separator = trimmed.indexOf(":");
+  const explicit = sessionId?.trim() ?? "";
+  const carried = separator === -1 ? "" : trimmed.slice(0, separator).trim();
+  const rawCode = separator === -1 || explicit !== "" ? trimmed : trimmed.slice(separator + 1);
```
mrfqcentic: **Blocking.** The `|| explicit !== ""` in this ternary keeps the `<sessionId>:` prefix inside the code whenever an explicit `sessionId` is *also* supplied.

`parsePairingTicket("ps_7f3a:abcd-efgh", "ps_7f3a")` takes the `explicit !== ""` branch, so `rawCode` is the whole `"ps_7f3a:abcd-efgh"`, which fails `/^[A-Z0-9]{8}$/` (the `_` and `:` survive the `[\s-]` strip) and throws `bad-code`. The session id is correct, the code is correct, and it is refused.

This is the shape the module was written to be tolerant of, so it is the shape a caller will hit: a CLI or Panel form that always forwards `--session` while letting the operator paste whatever they were read out. The test at `core-pairing.test.ts:578` covers `(combined)`, `(code, id)` and `(code, other-id)` but not `(combined, id)`, which is why it passes.

`separator === -1 ? trimmed : trimmed.slice(separator + 1)` fixes it — the explicit id still wins for the session on line 407, and the carried prefix is stripped either way.

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,798 @@
...
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(answer.body);
+  } catch {
+    throw new CorePairingError("malformed-response", `${origin} answered 200 with something that was not JSON`, {
+      status: answer.status,
+    });
+  }
+  const fields = parsed as Partial<Record<keyof CorePairingRedeemResponse, unknown>>;
+  const missing = (["endpoint", "caCert", "clientCert", "bearer"] as const).filter(
```
mrfqcentic, line 760: Non-blocking, but it belongs in this list. `endpoint` is checked only for being a non-empty string, then goes into the blob unexamined.

`coreConnectionFromBlob` (`core-registration-blob.ts:80-91`) decides TLS purely from the scheme: `const secure = url.startsWith("wss://")`, and for anything else it returns `tls: null` **while still returning the bearer**. So a Core that answers with a `ws://` endpoint — misconfigured, or hostile — hands back a credential whose every later dial ships the bearer in cleartext with no client certificate. That silently undoes the property this module spent a bootstrap dial, a fingerprint comparison and a pinned second connection to establish.

The check costs one predicate here, alongside the three fields already being validated, and the failure already has a name (`malformed-response`).

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,798 @@
...
+ * fingerprint on a piece of paper, and the only way to compare the two is to
+ * look at what the server presents. Nothing is sent on this socket, and it is
+ * destroyed as soon as the chain has been copied out.
+ */
+function presentedChain(host: string, port: number, timeoutMs: number): Promise<DetailedPeerCertificate[]> {
+  return new Promise((resolve, reject) => {
+    const socket = tlsConnect({
+      host,
+      port,
+      rejectUnauthorized: false,
```
mrfqcentic, line 527: Confirming this one, because it is the line the whole ticket turns on. This is the only `rejectUnauthorized: false` in the package — the source sweep in `core-pairing.test.ts` asserts exactly one and that it sits inside `presentedChain` — and nothing is written on this socket: it is opened, the chain is copied out on `secureConnect`, and it is destroyed.

The redemption at line 608 is a genuinely separate connection pinned three ways (`ca` set to the one matched certificate, `rejectUnauthorized: true`, and an in-handshake fingerprint re-check), and the cert-swap test proves the pinning rather than assuming it: a server that changes its secure context between the two dials receives **zero** requests. Belt-and-braces by construction, which is the point of the third check — the first two are object properties that a future edit could quietly drop.

packages/sdk/src/core-pairing-csr.ts
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * Mint a key pair and a CSR naming `commonName`.
+ *
+ * The subject here is a *request*, and the Core does not honour it: the
+ * certificate is issued for the label the operator typed when they opened the
+ * pairing session (`core-pairing-routes.ts`, "the subject is the Core's to
+ * write"). It is filled in anyway because a CSR with an empty subject is a
+ * worse artifact to debug than one that says which machine asked.
+ */
+export async function generateClientCsr(commonName: string): Promise<ClientCsr> {
```
mrfqcentic, line 60: The hand-written PKCS#10 encoder is a reasonable call and it is checked where it counts: the suite posts what this produces at the real Core, whose `assertSignableCsr` parses it with `@peculiar/x509` and verifies the signature before signing — so a mis-encoded field fails a test rather than shipping. RSA-2048 matches the CA's `MIN_RSA_MODULUS_BITS`, and `certificationRequestName` applies the same conservative character set the Core applies to the label it issues for.

Worth stating plainly since it is the reason for the encoder: `packages/sdk` is the one package here whose dependency list is a reviewed artifact, and a CSR library would have been a new runtime dependency on a published package to produce four fields in a fixed order. The subject is a request the Core does not honour anyway (`core-pairing-routes.ts:318` writes `CN` from the operator's label), which is also correct.

mrfqcentic (changes requested):
## Review — #284 (SDK client pairing)

Base verified: `feat/short-code-pairing` (`63ad949`). Head `c42b77d`.

**Boundary: HELD.** `git diff --stat feat/short-code-pairing...feat/sdk-core-pairing -- packages/cli packages/core` is empty. Four files, all under `packages/sdk`. I also checked the stronger version of the same rule: **no shipped SDK source file imports `@actana/*` at all** — the `@actana/core` and `@actana/shared` imports live only in `__tests__`, resolved through the pre-existing test-only alias in `vitest.config.ts` / `tsconfig.json` `paths`, and `tsconfig.build.json` excludes `src/__tests__/**`, so nothing private reaches the tarball. ADR 0025 D1/D4 holds.

The security core of this module is right, and it is proved rather than asserted. I am requesting changes on four defects in the new code — one of which reports a misconfigured Core as an active attack.

### Acceptance criteria (#284)

| # | Criterion | Verdict | Enforcing lines |
|---|---|---|---|
| 1 | A new module in `packages/sdk/src/` exporting one pairing function taking address, code and expected fingerprint, returning a `CoreRegistrationBlob` | **Met**, with one addition | `core-pairing.ts:301` `pairWithCore`. It additionally requires a `sessionId` — either its own field or carried as `<sessionId>:<XXXX-XXXX>`. That is forced by #282's session binding (`core-pairing-routes.ts:430` rejects a body without `sessionId`), and #283 now prints `Session <id>` from `actana pair new` (`actana-pair.ts:217`), so the operator has it. Worth closing out in #280 so #285 knows which form to hand over |
| 2 | Keypair locally via `node:crypto`; PEM CSR with the client label; private key never sent, logged or in any body | **Met** | `core-pairing-csr.ts:60` — `generateKeyPair("rsa", { modulusLength: 2048 })`, a hand-rolled PKCS#10 over the exported SPKI, returning `{ csrPem, privateKeyPem }`. `pairWithCore` puts only `csr` in the body (`:344`). The test asserts it from the **recorded request bytes**, not from the client's own belief: `expect(sent).not.toMatch(/PRIVATE KEY/)` plus a line-by-line search for the key, and `Object.keys(body).sort()` pinned to exactly `[client, code, csr, sessionId]` |
| 3 | Bootstrap dial obtains the chain and computes SHA-256 over DER, colon-separated uppercase hex — the same format #283 prints | **Met** | `presentedChain` → `chainOf` → `certificateAuthorityIn` → `fingerprintOf(ca.raw)` (`:494-570`). The agreement with #283 is real and not merely parallel: this suite asserts `identity.fingerprint === new X509Certificate(caCert).fingerprint256`, and #297's `core-cert-material.test.ts` asserts `certFingerprintSha256(...) === fingerprint256` for the same certificate — so the two implementations agree transitively through a third party neither task owns |
| 4 | Fingerprint mismatch aborts before the code is transmitted, naming both fingerprints | **Met** | `core-pairing.ts:320-330`. Asserted **from the server's side**: `rig.wire.requests === 0`, `rig.wire.bodies === []`, `rig.audit === []`, and `store.getSession(sessionId).attempts === 0` — no attempt spent on the operator's session |
| 5 | A mode that reports the presented fingerprint back for confirmation, without sending the code | **Met** | `fetchCorePairingIdentity` (`:235`) takes no code at all — a separate function rather than a flag, so a first-contact UI cannot leak a secret it was never given. `pairWithCore` with no fingerprint refuses `fingerprint-unconfirmed` and carries `presentedFingerprint` + `presentedCaCert` (`:313`) |
| 6 | No `rejectUnauthorized: false` in place for anything after the fingerprint check | **Met** | The only occurrence is `:500`, inside `presentedChain`, which sends nothing. `postRedemption` (`:608`) sets `ca` to the single matched certificate, `rejectUnauthorized: true`, and a `checkServerIdentity` that re-runs the hostname check *and* the fingerprint comparison before the handshake completes. A source sweep over the shipped files asserts exactly one relaxed dial and that it sits inside `presentedChain`. The cert-swap test stages a Core that changes certificate between the two dials and asserts `stub.requests === 0` |
| 7 | Wire types declared in the SDK, structurally identical to the Core's, with the ADR 0025 D1/D4 reasoning | **Met** | `core-pairing.ts:57-101` plus the file header. I checked the shapes against `core-pairing-routes.ts`: path `/v1/pair/redeem` matches `CORE_PAIRING_REDEEM_PATH`; the request's `{sessionId, code, client:{label}, csr}` matches `parseRedeemRequest`, which reads the label from `o.client.label`; the response's four fields match; and 403/429/404/5xx/other map onto `pairing-refused` / `rate-limited` + `retry-after` / `not-found` / `core-error` / `bad-request` |
| 8 | The blob is assembled from the response plus the locally-born `clientKey` and passes straight into `coreConnectionFromBlob` | **Met** | `:373-382`. The suite runs `coreConnectionFromBlob(blob)` into a real mTLS dial and an `auth` frame and gets `authOk`; it also asserts `certificatePublicKey(blob.clientCert) === privateKeyPublicHalf(blob.clientKey)`, which is the only thing that proves the returned key is the credential's other half rather than a key that merely rode along |
| 9 | Errors distinguishable — wrong/expired/consumed, attempts exhausted, rate limited, fingerprint mismatch, unreachable | **Partially met** | Two separate reasons. (a) *By design:* wrong / expired / consumed / attempts-exhausted all arrive as one `refused`, because #282 answers all four with one status and one body deliberately; a client reporting four would be inventing three. That is the right call and cannot be fixed here — but it is a deviation from the criterion as written and belongs in #280, not in a PR description. The suite does pin the distinction on the Core's side (`rig.audit.at(-1).reason` walks `wrong-code` → `already-consumed` → `expired` → `attempts-exhausted`). (b) *Not by design:* every TLS verification failure on the redemption dial is reported as `fingerprint-mismatch` — see B1 |
| 10 | Tests: happy path against a real in-process Core; mismatch asserted server-side; `clientKey` matches and never appears in a body; blob completes mTLS + `auth`; each error case distinguishable | **Met** | The rig is the Core's own `PtyCoreLinkServer` with its own pairing routes through its own wiring, `generateCertMaterial` certificates, and a real `PairingStore` on disk. Only two things are staged, both because the property is otherwise unobservable: a route wrapper recording request bytes, and a server that swaps its secure context between dials |

### Blocking

**B1 — a misconfigured Core is reported as an attack** (`core-pairing.ts:773`, inline). `certificateFailure` matches `code.includes("CERT")`, so `ERR_TLS_CERT_ALTNAME_INVALID` becomes `fingerprint-mismatch`. This is the one failure the whole module exists to report accurately, and it is the one it gets wrong on a realistic configuration.

**B2 — a `null` 200 body throws a raw `TypeError`** (`core-pairing.ts:690`, inline), escaping the documented "throws `CorePairingError` for everything that is not a blob" contract.

**B3 — `parsePairingTicket` mis-parses the combined form when `sessionId` is also given** (`core-pairing.ts:405`, inline). The function added specifically to accept both forms rejects a valid input.

**B4 — the response `endpoint` is not held to `wss://`** (`core-pairing.ts:691`, inline), so a credential obtained over a pinned connection can be handed back with `tls: null` and a cleartext bearer.

### Cross-PR — this is why #296 must be re-reviewed after #297 lands

`packages/sdk/src/__tests__/core-pairing.test.ts:40` imports `@actana/core/core-pairing-audit`. **#297 renames that file** to `packages/shared/src/pairing-audit.ts`. I ran `git merge-tree` on the two heads: they merge **cleanly**, no conflict — and in the resulting tree `packages/core/src/core-pairing-audit.ts` does not exist while that import line does. Git will not flag it; `pnpm -r typecheck` will (TS2307), and because it is an `import type` the vitest run may not, which makes it exactly the kind of breakage that reaches the feature branch quietly. One-line fix to `@actana/shared/pairing-audit` once #297 is in.

### Also worth a note, not blocking

- **`certificateAuthorityIn` requires the top of the chain to be self-issued.** True of every Core today (`generateCertMaterial` makes a self-signed root), and the comment explains why fingerprinting whatever is on top would be worse. But it makes the client depend on the Core never fronting an intermediate or serving a partial chain — worth writing into #280 as a property of the flow rather than an implementation detail of this file.
- **The `sessionId` question is still open.** Accepting both forms is the right way to not settle it in a PR, and #283 does now print the session id — but nothing yet produces the `<sessionId>:<code>` string, so #285 should decide and #280 should record it.

**Verdict: request changes.** The security argument holds and is well tested; B1–B4 are four defects in new code, and B1 lands on an acceptance criterion.

packages/sdk/src/core-pairing.ts
qcentic-dev: Fixed in 18d0604. Width and accusation are now two decisions rather than one.

`certificateFailure` keeps the broad prefix match, but it now answers only "was this about the certificate at all?" — where being wrong costs the wrong noun and nothing else. Whether it was the **pin** is decided by `classifyDialFailure` against an enumerated set of OpenSSL chain verdicts (`UNABLE_TO_GET_ISSUER_CERT[_LOCALLY]`, `UNABLE_TO_VERIFY_LEAF_SIGNATURE`, `SELF_SIGNED_CERT_IN_CHAIN`, `DEPTH_ZERO_SELF_SIGNED_CERT`, `CERT_SIGNATURE_FAILURE`, `CERT_UNTRUSTED`, `INVALID_CA`) plus the tagged `ERR_ACTANA_PAIRING_PIN` refusal. `ERR_TLS_CERT_ALTNAME_INVALID` is its own `hostname-mismatch`; everything else certificate-shaped is `certificate-invalid`. The TLS code rides along in `detail.tlsCode`.

Both new sentences open with "presented the expected certificate authority, but …", and the hostname one names the address dialled and says to use the one the Core was set up for.

Unknown verification codes now default to `certificate-invalid` rather than to an accusation. That is the conservative direction for the *message* and costs nothing in security: the exchange is refused either way and the body was never written to the socket.

Two regression tests, both staged rather than asserted about: a Core reached off its SAN (stub bound to `127.0.0.2`, certificate covering `127.0.0.1`) → `hostname-mismatch` + `ERR_TLS_CERT_ALTNAME_INVALID`, and a backdated server certificate from the same CA → `certificate-invalid` + `CERT_HAS_EXPIRED`. The certificate-swap test still asserts `fingerprint-mismatch`, and now also pins `CERT_SIGNATURE_FAILURE` — so the enumerated set is what that assertion rests on, not a substring that happened to match.

packages/sdk/src/core-pairing.ts
qcentic-dev: Fixed in 18d0604 — guarded exactly as `safeRefusal` does, before the field sweep:

```ts
if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) {
  throw new CorePairingError("malformed-response", `${origin} answered 200 with something that was not an object`, {
    status: answer.status,
  });
}
```

`Array.isArray` is in there for the same reason: an array parses, survives the cast, and would have reached the sweep as something the compiler is happy to read a field off. Both `null` and `[]` are covered by tests.

packages/sdk/src/core-pairing.ts
qcentic-dev: Fixed in 18d0604, with the line you wrote:

```ts
const rawCode = separator === -1 ? trimmed : trimmed.slice(separator + 1);
```

The explicit id still wins for the session, with one addition: two ids that **disagree** are now refused (`bad-code`, naming both) rather than resolved in favour of the explicit one. Silently preferring one sends the redemption at a session the operator did not mean, and that comes back as the same `refused` a mistyped code does — a failure with nothing in it to point at the real mistake.

The test now covers `(combined, id)` — the case that was missing — and `(combined, other-id)`.

packages/sdk/src/core-pairing.ts
qcentic-dev: Taken, and treated as blocking rather than as a note — you are right that it undoes the property the rest of the module is built to establish. Fixed in 18d0604: the endpoint is held to `wss://` beside the three fields already checked, as `malformed-response`, with the offending endpoint in the message.

Test: a stub answering 200 with `ws://127.0.0.1:9443` and otherwise well-formed material is refused rather than returned.

qcentic-dev:
## Review addressed — head is now `e3a73ed`

Thank you for B1 in particular: the wide match was mine, and it landed on the one sentence in this module that has to be exact.

**All four blocking findings are fixed**, in two commits, both inside `packages/sdk`:

| | Finding | Fix | Test |
|---|---|---|---|
| **B1** | `code.includes("CERT")` reported a misconfigured Core as an attack | `18d0604` — width and accusation are now separate decisions: `certificateFailure` answers only "was this about the certificate?", and `classifyDialFailure` decides *which* against an enumerated set of chain verdicts plus the tagged `ERR_ACTANA_PAIRING_PIN`. New `hostname-mismatch` and `certificate-invalid` failures, with `detail.tlsCode` | A stub bound to `127.0.0.2` with a `127.0.0.1` certificate → `hostname-mismatch` + `ERR_TLS_CERT_ALTNAME_INVALID`; a backdated server certificate → `certificate-invalid` + `CERT_HAS_EXPIRED`; the swap test still `fingerprint-mismatch`, now pinning `CERT_SIGNATURE_FAILURE` |
| **B2** | `null` 200 body threw a raw `TypeError` | `18d0604` — `!parsed \|\| typeof parsed !== "object" \|\| Array.isArray(parsed)` before the field sweep, as `safeRefusal` already does | `null` and `[]` bodies, both `malformed-response` |
| **B3** | combined ticket mis-parsed when `sessionId` was also given | `18d0604` — your line. Plus: two ids that *disagree* are refused rather than resolved | `(combined, id)` — the case the suite was missing — and `(combined, other-id)` |
| **B4** | response `endpoint` not held to `wss://` | `18d0604` — checked beside the other three fields, as `malformed-response` | a 200 answering `ws://…` is refused |

Unknown TLS verification codes now default to `certificate-invalid` rather than to an accusation. That costs nothing in security — the exchange is refused either way and the body was never written to the socket — and it means an unrecognised failure shows a person the cautious sentence rather than the alarming one.

**The cross-PR seam is closed now rather than after #297.** Instead of waiting to repoint the import, `core-pairing.test.ts` no longer names that module at all: it declares the one field it reads off an audit line (`{ outcome, reason? }`) and passes it structurally. The suite is now indifferent to whether that file lives in `@actana/core` or `@actana/shared`, so the clean-merge-with-a-dead-import you found cannot happen. `git merge-tree` against #297's head is clean and the resulting tree has no reference to `core-pairing-audit` from this package.

### Non-blocking notes

- **Self-issued top of chain** — taken as far as it goes in code: `certificateAuthorityIn`'s comment now names it as a property of the *flow* rather than of this file, and says plainly that a Core that later fronts an intermediate or serves a partial chain breaks first contact for every client. **Recording it in #280 is deferred** — it is a statement about the flow, and #280 is where the flow's properties are agreed; I have not edited that issue from this PR.
- **The `sessionId` question** — **deferred**, deliberately, and unchanged by this round. Both forms still parse (B3 makes the combined-plus-explicit form actually work, which it did not before). Nothing here produces the `<sessionId>:<code>` string; #285 should decide and #280 should record it.
- **AC 9's four-into-one refusal** — **deferred**, as you say it must be: splitting it means changing #282's threat model. It belongs in #280, not in a PR description; I have left the PR's acceptance-criteria comment marking that criterion unchecked with the reason.

### Checks

Local, in the worktree: `vitest run` across `@actana/sdk` — 274 passed, 5 skipped, 0 failed (21 in `core-pairing.test.ts`, up from 18); `tsc --noEmit`; `eslint packages/sdk`; `tsc -p tsconfig.build.json`. Both new commits pass `commitlint` with this repo's config, longest body line 82 against the 132 limit.

Still a draft, base unchanged (`feat/short-code-pairing`), nothing merged, and nothing outside `packages/sdk` touched — `git diff --stat feat/short-code-pairing...feat/sdk-core-pairing -- packages/cli packages/core` is still empty.

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,958 @@
...
+ * So the width and the accusation are now two separate decisions.
+ * {@link certificateFailure} still decides broadly whether the failure was
+ * about the certificate at all — being wrong there only costs the wrong noun.
+ * Whether it was the *pin* is decided by an enumerated set plus the tagged
+ * refusal, and anything certificate-shaped that is in neither is reported as an
+ * invalid certificate. Nothing is lost by that default: the exchange is refused
+ * either way and the code was never written to the socket — the only thing that
+ * changes is which sentence a person is shown.
+ */
+function classifyDialFailure(err: unknown): DialFailure {
```
mrfqcentic, line 925: This is the finding's line and the split is exactly the right one: width and accusation as two decisions instead of one predicate doing both jobs.

I traced both cases from the report. `ERR_TLS_CERT_ALTNAME_INVALID` is absent from `CHAIN_FAILURE_CODES` above and is matched by name on the line below, so an off-SAN Core now reaches `hostname` — and `dialFailure` gives it a sentence that names the address dialled and says to use the one the Core was set up for, which is the half an operator can act on. `CERT_HAS_EXPIRED` is likewise not in the chain set and is caught by `certificateFailure`'s `includes("CERT")`, so it lands on `certificate`. Neither is an accusation any more.

Enumerating the pin set rather than prefix-matching it is what makes this hold up: `CHAIN_FAILURE_CODES` is eight verdicts that are each a statement about who signed what, which on a dial whose `ca` is the one fingerprinted certificate is precisely the pin refusing. And defaulting an unrecognised verification code to `certificate-invalid` is right for the reason the comment gives — the exchange is refused either way and the body was never written, so the only thing at stake is which sentence a person is shown, and "you are being intercepted" is the one that must never be a guess.

The certificate-swap test now pinning `CERT_SIGNATURE_FAILURE` is the detail that makes the whole thing durable: `fingerprint-mismatch` is asserted against the set that decides it, not against a substring that happened to match.

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,958 @@
...
+  } catch {
+    throw new CorePairingError("malformed-response", `${origin} answered 200 with something that was not JSON`, {
+      status: answer.status,
+    });
+  }
+  // `JSON.parse("null")` succeeds, and an array parses too. Both would reach
+  // the sweep below as something a field can be read off without complaint from
+  // the compiler, and `null` would throw a raw `TypeError` out of a function
+  // whose whole contract is that failures arrive as a `CorePairingError`.
+  if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) {
```
mrfqcentic, line 754: Guarded before the sweep, as `safeRefusal` does. `Array.isArray` is an addition beyond the finding and a correct one — an array parses, survives the cast, and would have reached the sweep as something a field can be read off without complaint, producing a `malformed-response` that complained about missing fields rather than about the shape. Honest failure beats a misleading one.

Both `null` and `[]` are covered by tests, which matters here specifically because the defect was that a raw `TypeError` escaped the failure union every caller of this module is written to switch on.

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,958 @@
...
+export function parsePairingTicket(input: string, sessionId?: string): PairingTicket {
+  const trimmed = input.trim();
+  const separator = trimmed.indexOf(":");
+  const explicit = sessionId?.trim() ?? "";
+  const carried = separator === -1 ? "" : trimmed.slice(0, separator).trim();
+  // The prefix is stripped whenever there is one, whether or not a session id
+  // was also passed: a caller that always forwards `--session` while letting an
+  // operator paste whatever they were read out hands in both, and refusing that
+  // would refuse the one shape this function exists to be tolerant of.
+  const rawCode = separator === -1 ? trimmed : trimmed.slice(separator + 1);
```
mrfqcentic, line 423: Prefix stripped whenever there is one — the combined form now parses with or without an explicit `sessionId`, which is the tolerance this function exists to provide.

The disagreement refusal just below is yours, not mine, and I think it is right. Silently preferring the explicit id would send the redemption at a session the operator did not mean, and #282 answers that with the same uniform `refused` a mistyped code gets — so the one failure that could have pointed at the real mistake arrives indistinguishable from the commonest one. Refusing locally, naming both ids, costs nothing and is the only place in the flow where that information still exists.

Good that the test now covers `(combined, id)`. Its absence is exactly why the original slipped through: the three cases that were covered all happened to avoid the branch.

packages/sdk/src/core-pairing.ts
```diff
@@ -0,0 +1,958 @@
...
+  // The endpoint decides whether anything later is protected at all.
+  // `coreConnectionFromBlob` reads TLS off the scheme and nothing else: a
+  // `ws://` endpoint yields `tls: null` **and still carries the bearer**, so a
+  // Core that answered with one — misconfigured, or hostile — would hand back a
+  // credential whose every later dial ships the bearer in cleartext with no
+  // client certificate. That is the property this module spent a bootstrap
+  // dial, a fingerprint comparison and a pinned second connection to establish,
+  // undone by one field nobody looked at.
+  const endpoint = (fields.endpoint as string).trim();
+  if (!endpoint.startsWith("wss://")) {
```
mrfqcentic, line 779: Held to `wss://` beside the three fields already checked, which is where it belonged.

The comment states the mechanism correctly and it is worth having written down here rather than inferred from `core-registration-blob.ts`: that function reads TLS off the scheme and nothing else, so a `ws://` endpoint produces `tls: null` **and still carries the bearer**. One unexamined field would have undone the bootstrap dial, the fingerprint comparison and the pinned second connection — everything above this line — by handing back a credential that ships the bearer in clear on every later dial.

The stub test answering 200 with `ws://127.0.0.1:9443` and otherwise valid material is the right shape: well-formed in every respect except the one under test.

packages/sdk/src/__tests__/core-pairing.test.ts
```diff
@@ -0,0 +1,902 @@
...
+      }),
+    );
+
+    expect(failure.failure).toBe<CorePairingFailure>("fingerprint-mismatch");
+    expect(stub.requests).toBe(1);
+  }, 30_000);
+});
+
+describe("a certificate problem is not an accusation", () => {
+  it("tells a Core reached off its SAN from a Core that is not the right Core", async () => {
```
mrfqcentic, line 722: This is the test the finding needed and it is staged rather than asserted about: a stub bound to `127.0.0.2` presenting a certificate that covers `127.0.0.1`, so the CA matches on both dials and only the name does not. That is the real-world case — per `core-cert-material.ts` a Core's server certificate covers the configured host plus loopback, so a Core reached on a second address hits it — and it is now pinned to `hostname-mismatch` with the TLS code beside it.

Pairing it with the backdated-certificate test below is what makes the classification trustworthy: two different non-pin certificate failures, two different failures reported, and the swap test holding `fingerprint-mismatch` for the case that actually is one.

packages/sdk/src/__tests__/core-pairing.test.ts
```diff
@@ -0,0 +1,902 @@
...
+import { afterEach, describe, expect, it } from "vitest";
+import { generateCertMaterial, issueServerCert } from "@actana/shared/core-cert-material";
+import { verifyBearer } from "@actana/shared/core-link-bearer";
+import { generatePairingCode } from "@actana/shared/pairing-code";
+import { createPairingSession } from "@actana/shared/pairing-session";
+import { PairingStore, derivePairingCodeKey, hashPairingCode } from "@actana/shared/pairing-store";
+import type { CoreHttpRoutes } from "@actana/core/core-files-routes";
+import { createCoreFilesRequestHandler } from "@actana/core/core-files-routes";
+import { PairingRateLimiter } from "@actana/core/core-pairing-rate-limit";
+import { buildCorePairingRoutes, composeCoreHttpRoutes, isPairingPath } from "@actana/core/core-pairing-wiring";
```
mrfqcentic, line 40: Confirming the cross-PR hazard is gone: the `@actana/core/core-pairing-audit` import that stood here last round has been dropped in favour of reading the audit line structurally, so this suite no longer depends on a path #297 deletes.

I verified it rather than taking the commit message for it — `git merge-tree` on the two current heads still merges cleanly, and `git grep core-pairing-audit` over that **merged tree** now returns nothing, where last round it returned this line pointing at a file that no longer existed. The three remaining references on this branch are base-branch files this PR does not touch and #297 updates as part of the rename.

Also checked that the new `issueServerCert` import resolves against the **base** (`core-cert-material.ts:157`) rather than against anything #297 adds — so this PR carries no build dependency on that one, and the merge order is hygiene rather than necessity.

mrfqcentic (approved):
## Re-review — #284, at `e3a73ed`

Previous review: `c42b77d`, changes requested on B1–B4. Two new commits. Base tip unchanged (`feat/short-code-pairing` @ `63ad949`).

**Boundary: still HELD.** `git diff --stat 63ad949...e3a73ed -- packages/cli packages/core` is empty; still four files, all under `packages/sdk`. I re-ran the stronger check too: no shipped SDK source imports `@actana/*` at all — the sweep over every non-test `.ts` under `packages/sdk/src` comes back empty, and `package.json` still declares only `undici` and `ws`.

**CI on this head: green.** All 18 reporting checks pass — Typecheck, Lint, Unit Tests, Conventions, Dependency Audit, Secret Scan, both installer E2Es, the Panel service-seam E2E, both release-tarball smokes, Promotion gate, Train rules. The remaining entries are `skipping` on image-publish and train-tag jobs, expected for a non-train PR.

### B1 — a misconfigured Core reported as an attack. **Fixed**, and proved from the enforcing lines.

The old defect was one predicate doing two jobs: `certificateFailure` decided both *whether* the failure was about certificates and *that* it was the pin, so `code.includes("CERT")` swept `ERR_TLS_CERT_ALTNAME_INVALID` into `fingerprint-mismatch`.

Those are now two decisions. `certificateFailure` (`core-pairing.ts:855`) keeps the broad prefix match and answers only "about the certificate at all?" — where being wrong costs the wrong noun. `classifyDialFailure` (`:925`) decides the accusation, and it is enumerated:

- `PIN_FAILURE_CODE` → `pin`. The tagged refusal this module's own `checkServerIdentity` raises.
- `CHAIN_FAILURE_CODES` (`:879`) → `pin`. Eight OpenSSL verdicts, every one a statement about *who signed what* — no issuer, an issuer that is not the one supplied, a signature that does not check out, a chain ending in a self-signed certificate that is not the pinned root. On a dial whose `ca` is the single certificate the operator's fingerprint matched, that set is the pin refusing.
- `ERR_TLS_CERT_ALTNAME_INVALID` → `hostname`. Checked after the chain set and absent from it, so it lands here.
- anything else certificate-shaped → `certificate`; otherwise `transport`.

I traced the two cases from the finding. `ERR_TLS_CERT_ALTNAME_INVALID` is not in `CHAIN_FAILURE_CODES`, so it reaches the hostname branch. `CERT_HAS_EXPIRED` is also not in it, and `certificateFailure("CERT_HAS_EXPIRED")` is true via `includes("CERT")`, so it reaches `certificate`. Both correct.

`dialFailure` (`:695`) gives each its own sentence, and the sentences are the point. The hostname one opens "presented the expected certificate authority, but its certificate does not cover `<host>`" and tells the operator to dial the address the Core was set up for — which is the actionable half. `detail.tlsCode` carries the raw code for a caller that wants it.

**Defaulting an unrecognised verification code to `certificate-invalid` rather than to the accusation is the right call**, and the reasoning in the comment is the reason: the exchange is refused either way and the body was never written, so the only thing that changes is which sentence a person is shown — and "you may be under attack" is the one that must not be guessed.

Two regression tests, both **staged rather than asserted about**, which is what I wanted: `core-pairing.test.ts:722` binds a stub to `127.0.0.2` with a certificate covering `127.0.0.1` and asserts `hostname-mismatch` + `ERR_TLS_CERT_ALTNAME_INVALID`; `:756` uses a backdated server certificate from the same CA and asserts `certificate-invalid` + `CERT_HAS_EXPIRED`. And the certificate-swap test now also pins `CERT_SIGNATURE_FAILURE`, so `fingerprint-mismatch` rests on the enumerated set rather than on a substring that happened to match — the assertion and the mechanism now agree.

### B2 — `null` 200 body threw a raw `TypeError`. **Fixed.**

`:754` — `if (!parsed || typeof parsed !== "object" || Array.isArray(parsed))`, before the field sweep, exactly as `safeRefusal` does. `Array.isArray` is a good addition I had not called out: an array parses, survives the cast, and reaches the sweep as something the compiler is happy to read a field off — it would have produced a confusing `malformed-response` about missing fields rather than an honest one about the shape. Both `null` and `[]` are covered at `:834-840`.

### B3 — `parsePairingTicket` mis-parsed the combined form. **Fixed.**

`:423` — the prefix is stripped whenever there is one, regardless of `sessionId`. `:425` adds something I did not ask for and agree with: two session ids that **disagree** are refused rather than resolved in favour of the explicit one. Silently preferring one would send the redemption at a session the operator did not mean, and #282 answers that with the same uniform `refused` a mistyped code gets — a failure with nothing in it to point at the real mistake. Refusing locally, naming both, costs nothing and is the only place that information exists. The test now covers `(combined, id)` — the case whose absence let the bug through — and `(combined, other-id)`.

### B4 — response `endpoint` not held to `wss://`. **Fixed**, and treated as blocking, which was the right reading.

`:779`. The comment states the mechanism correctly: `coreConnectionFromBlob` reads TLS off the scheme and nothing else, so a `ws://` endpoint yields `tls: null` *and still carries the bearer*. Refused as `malformed-response` with the offending endpoint in the message. `:842` stages a stub answering 200 with `ws://127.0.0.1:9443` and otherwise well-formed material.

### Cross-PR — resolved, and I re-checked it rather than taking the commit message for it

Last round's hazard was `core-pairing.test.ts:40` importing `@actana/core/core-pairing-audit`, which #297 renames. That import is **gone** — the suite reads the audit line structurally instead. I re-ran the check end to end:

- `git merge-tree --write-tree` on the two current heads still merges cleanly (`90da11a`);
- `git grep core-pairing-audit` over that **merged tree** returns nothing, where last round it returned a live import against a deleted file;
- the three remaining references on this branch (`core-pairing-routes.ts`, `core-pairing-redeem.test.ts`, `core-pairing-audit.test.ts`) are base-branch files this PR does not touch, and #297 updates all three as part of the rename.

So the merged tree is consistent and `pnpm -r typecheck` has nothing left to trip on. I also confirmed the new `issueServerCert` import is satisfied by the **base** branch (`core-cert-material.ts:157`), not by #297 — so this PR has no build dependency on that one, and the merge order below is hygiene rather than necessity.

### Acceptance criteria

All ten now **met**. AC 9 was my partial last round for two reasons and one of them is gone: the misclassification is fixed, and the union has grown two failures a caller can act on differently (`hostname-mismatch`, `certificate-invalid`). The remaining collapse — wrong / expired / consumed / attempts-exhausted arriving as one `refused` — is forced by #282 answering all four identically on purpose, and reporting four would be inventing three. That is the correct behaviour, the suite pins the real distinction on the Core's audit side, and what it needs is a line in #280 recording the decision rather than a change here. I am not holding the PR for it.

### Still open for #280, not for this PR

- The `<sessionId>:<code>` form now parses cleanly both ways, but nothing yet *produces* it; #283 prints `Session <id>` as a separate line. #285 should pick one and #280 should record it.
- `certificateAuthorityIn` requires the presented chain to end in a self-issued certificate. True of every Core today and argued in the comment, but it is a property of the flow rather than of this file.

**Verdict: approve.**

Merge this **second**, after #297.

PR #297, opened 2026-08-20
feat(cli): pairing — actana pair new, ls and revoke, and the daemon that enforces a revocation (#283)

qcentic-dev:
The Core-side operator's half of short-code enrollment: `actana pair new`, `actana pair ls`, `actana pair revoke`.
Built against the pairing server and store from #282, which are already on this base branch.

**Base is `feat/short-code-pairing`**, the feature branch for #280 — not `main`, not the train.

## What is here

**`actana pair new [--label <name>] [--ttl <duration>]`** mints a pending session and prints three things: the
`XXXX-XXXX` code, this Core's CA fingerprint, and the expiry as both an absolute UTC time and a relative
"in 5 minutes". They go to stdout as labelled lines and every word explaining them goes to stderr, so a `pair new`
being piped stays the facts and nothing else.

The TTL comes from `PAIRING_SESSION_TTL_MS` — the shared constant from #281, never a literal in the CLI — and
`--ttl` requires a unit. A bare `5` is the one input where a wrong guess is invisible: read as seconds it is a code
that dies before the operator finishes reading it out, read as minutes it outlives the phone call by four. The
ceiling is the store's own retention window, so a live session can never be pruned out from under the client
holding its code.

**The fingerprint** is a new `certFingerprintSha256` in `core-cert-material.ts`: SHA-256 over the CA certificate's
DER, colon-separated upper-case hex, computed from `PersistedMaterial.caCert`. The format is the contract, not a
presentation choice — a human compares two of these by eye, so an operator checking it against
`openssl x509 -fingerprint -sha256` must see the same characters. There is a test asserting it equals Node's own
`fingerprint256`, and another that PEM whitespace cannot move it. **#284 verifies its bootstrap dial against this
string**, which is the agreement the two tasks owed each other.

**`actana pair ls`** prints pending codes and paired clients in two labelled sections, each saying "None." rather
than printing an empty table. It cannot print a code, and that is the design working rather than a rule being
obeyed: what the store holds is a keyed digest. The `--json` payload names its fields one at a time for the same
reason the audit record does — a spread would publish whatever the stored shape grows next, and that shape includes
the digest.

**`actana pair revoke <target>`** takes a certificate serial (or a prefix), a session id, or a label matching
exactly one of either. It refuses to guess when a label matches two, because two machines paired under one name is
the ordinary way that happens and revoking the wrong one is not something an operator finds out about until the
machine they meant to keep stops working. Pointed at a pending code it cancels it; pointed at a client it stamps
the row the daemon enforces on.

## Revocation had to reach the daemon

`pair revoke` runs in the CLI, in a different process, and all it can do there is stamp a row. That stamp is a
record, not an enforcement: the certificate it names is still one this Core's CA signed, the bearer still verifies
against the same secret, and any link the client already has open is still carrying frames. So the larger half of
this PR is `packages/core`, in three places:

- **The client certificate** is refused at the gate — every `/v1/…` request and every core-link upgrade. `revoked`
  is checked ahead of `authorized` in both, and the order is the whole mechanism: a revoked certificate is a
  *valid* certificate, TLS has no idea an operator took it back, so `authorized: true` is exactly what a revoked
  client arrives with.
- **The bearer** is refused at the `auth` frame, through the `pair:<serial>` subject #282 mints. It answers
  `expired` rather than a fourth reason: the wire's three reasons live in `@actana/sdk` (**#284's package — not
  touched here**), and from where the client stands a revoked credential is one whose validity ended. Its own
  reconnect path then leads to re-pairing, which is where the operator who revoked it wants it to go. The Core's
  log carries the real reason, read by the person who revoked it.
- **A link already open** is closed, because the first two cannot reach it. A Panel that paired yesterday never
  presents its certificate again and never re-sends `auth`, so a rule enforced only at the door would leave it
  driving this Core until a reconnect that for a healthy client never comes. A one-second sweep of the pairing
  store is what notices.

Polling rather than `fs.watch`: `fs.watch` is per-platform, misses writes behind a rename — and a rename is how
`PairingStore` writes — and reports nothing at all on some network mounts. The bound this buys is stated rather
than implied: **a revoked client's live link closes within a second, not instantly.** An unreadable store leaves
the revoked set as it was, because "nobody is revoked" is the one interpretation that hands a revoked client its
access straight back.

## Two things moved

- **`core-pairing-audit.ts` → `packages/shared/src/pairing-audit.ts`.** It lived in `packages/core` when the daemon
  was its only writer. `pair revoke` is the second writer and runs in the `actana` CLI, which never loads
  `@actana/core`. One shape, one field list and one redaction across both is the point — two modules agreeing by
  convention would drift the first time a field was added to one of them.
- **`PairingSession` gains `revokedAt`**, and `canRedeem` checks it first. Its own field rather than a `consumedAt`
  stamp: a cancelled session was never redeemed, so nobody holds a certificate for it, and an operator reading the
  audit log afterwards has to be able to tell "somebody used this code" from "I took it back". The field is
  optional and read with `?? null` — sessions written before it existed have no `revokedAt`, and `undefined !==
  null` would kill every pending session on a Core the moment it upgraded.

## Notes

- **The verbs live in `packages/cli/src/actana-cli.ts`**, not the `packages/core` path #283's file pointers name.
  #288 unified the two `actana` binaries and moved the operator dispatch; this follows it there.
- **`packages/sdk` is untouched** — that is #284's, and `git diff --stat -- packages/sdk` against the base is empty.
- **The naming trap #283 flags is handled in the help.** `actana pair` is the Core end; `actana core pair` (#285)
  is the client end; since #288 they are one binary, so both help texts open by saying which machine they are for.
  `actana pair --help` answers on a laptop that has never run `actana setup`, because the help is a question about
  this program rather than about this box.
- Full monorepo suite, `pnpm -r typecheck` and `pnpm lint` are green. 88 new tests: 37 in `actana-pair.test.ts`,
  20 in `core-pairing-revocation.test.ts`, 15 in `core-link-revocation.test.ts`, and 16 added to the existing
  machine-CLI, cert-material, pairing-session and pairing-store suites.

Refs #283
Refs #280

https://claude.ai/code/session_01KNgVuRvpzdmmUZxjeLYcHA

### DIFF
--- a/packages/cli/src/actana-pair.ts
+++ b/packages/cli/src/actana-pair.ts
@@ -166,6 +166,16 @@ function pairNew(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): 
     deps.err(`actana pair new: ${flags.error}.`);
     return EXIT_USAGE;
   }
+  // `actana pair new laptop` used to mint an *unlabelled* code and exit 0. The
+  // operator then reads that code out believing it is called `laptop`, `pair
+  // ls` shows `(unnamed)`, and `pair revoke laptop` says nothing matches. An
+  // argument this verb has no use for is a mistake, exactly as an unknown flag
+  // is.
+  if (flags.positionals.length > 0) {
+    deps.err(`actana pair new: unexpected argument "${flags.positionals[0]}".`);
+    deps.err(`The name goes in a flag — \`actana pair new --label ${flags.positionals[0]}\`.`);
+    return EXIT_USAGE;
+  }
 
   const rawTtl = valueFlag(flags.values, "ttl");
   const ttl = rawTtl === undefined ? PAIRING_SESSION_TTL_MS : parseDuration(rawTtl);
@@ -183,6 +193,9 @@ function pairNew(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): 
     deps.err("Run `actana setup` — pairing needs a CA to sign against.");
     return EXIT_FAILURE;
   }
+
+  const store = openStore(deps, materialPath, "new");
+  if (!store) return EXIT_FAILURE;
 
   const now = deps.now();
   const label = valueFlag(flags.values, "label") ?? "";
@@ -203,7 +216,7 @@ function pairNew(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): 
     ttlMs: ttl,
   });
 
-  new PairingStore(pairingStorePath(materialPath)).createSession(session, now);
+  store.createSession(session, now);
 
   // stdout: the three facts a human reads out. stderr: everything explaining
   // them, so a `pair new` that is being piped or screen-scraped stays three
@@ -227,11 +240,18 @@ function pairLs(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): n
     deps.err(`actana pair ls: ${flags.error}.`);
     return EXIT_USAGE;
   }
+  if (flags.positionals.length > 0) {
+    deps.err(`actana pair ls: unexpected argument "${flags.positionals[0]}".`);
+    deps.err("`ls` takes no arguments — it lists everything.");
+    return EXIT_USAGE;
+  }
 
   const materialPath = ctx.materialPath();
   if (materialPath === null) return EXIT_FAILURE;
 
-  const store = new PairingStore(pairingStorePath(materialPath));
+  const store = openStore(deps, materialPath, "ls");
+  if (!store) return EXIT_FAILURE;
+
   const now = deps.now();
   const pending = store.listSessions().filter((session) => canRedeem(session, now).ok);
   const clients = store.listClients();
@@ -311,7 +331,13 @@ function pairRevoke(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext
     return EXIT_USAGE;
   }
   const [target, ...extra] = flags.positionals;
-  if (target === undefined) {
+  // Blank as well as absent, and that is not tidiness. `""` prefix-matches
+  // every serial and every session id, so `actana pair revoke "$SERIAL"` in a
+  // script where `SERIAL` is unset matches everything — and on a Core with
+  // exactly one client the ambiguity check below does not fire, so it revokes
+  // it and exits 0. That check is the whole defence against revoking the wrong
+  // machine, and this is the input that walks straight past it.
+  if (target === undefined || target.trim() === "") {
     deps.err("actana pair revoke: a target is required — `actana pair revoke <serial|session|label>`.");
     return EXIT_USAGE;
   }
@@ -323,7 +349,9 @@ function pairRevoke(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext
   const materialPath = ctx.materialPath();
   if (materialPath === null) return EXIT_FAILURE;
 
-  const store = new PairingStore(pairingStorePath(materialPath));
+  const store = openStore(deps, materialPath, "revoke");
+  if (!store) return EXIT_FAILURE;
+
   const now = deps.now();
   const audit = pairingAuditor(ctx.audit ?? ((record) => deps.err(`pairing.revoke ${formatJson(record)}`)));
 
@@ -405,6 +433,39 @@ function pairRevoke(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext
 }
 
 /**
+ * This Core's pairing store, or `null` after saying why it cannot be used.
+ *
+ * `readStrict` rather than `read`, and every verb here goes through it.
+ * `PairingStore.read` answers an unreadable file with an empty store, and each
+ * of these verbs would then do something wrong with that answer: `ls` would
+ * report a Core that has paired nobody, `revoke` would say nothing matches, and
+ * `new` would be worse than either — `createSession` reads the whole document,
+ * adds a session and writes it back, so minting against a corrupt file
+ * *replaces* it, taking the record of which clients are revoked with it.
+ *
+ * That last one is not hypothetical damage. The daemon fails closed on exactly
+ * this file (`core-pairing-revocation.ts`), so a Core with an unreadable
+ * `pairing.json` is refusing every client it ever paired — and a `pair new`
+ * that quietly rewrote the file would end that refusal by forgetting who was
+ * revoked, handing every revoked certificate its access back. Recovery has to
+ * be repairing the file, so these verbs decline to be the thing that destroys
+ * it.
+ */
+function openStore(deps: ActanaCliDeps, materialPath: string, verb: string): PairingStore | null {
+  const store = new PairingStore(pairingStorePath(materialPath));
+  try {
+    store.readStrict();
+  } catch (err) {
+    deps.err(`actana pair ${verb}: ${err instanceof Error ? err.message : String(err)}.`);
+    deps.err("While that file cannot be read, this Core refuses every client it has paired.");
+    deps.err("Repair the JSON or restore it from a backup. Do not delete it — it is the record");
+    deps.err("of which pairings were revoked, and deleting it would un-revoke every one of them.");
+    return null;
+  }
+  return store;
+}
+
+/**
  * Does this target name this client?
  *
  * A serial matches on a prefix — serials are 32 hex characters and nobody
@@ -445,10 +506,14 @@ function readFlags(argv: string[], spec: Record<string, FlagKind>): ReadFlags {
   const positionals: string[] = [];
   for (let i = 0; i < argv.length; i++) {
     const token = argv[i]!;
-    if (!token.startsWith("--")) {
+    if (!token.startsWith("-")) {
       positionals.push(token);
       continue;
     }
+    // A single-dash token is a flag this build does not know, never a value.
+    // Read as a positional, `actana pair new -l laptop` would have minted an
+    // unlabelled code from two arguments that both look like they were used.
+    if (!token.startsWith("--")) return { error: `unknown option: ${token}` };
     const eq = token.indexOf("=");
     const name = eq >= 0 ? token.slice(2, eq) : token.slice(2);
     const kind = spec[name];
--- a/packages/cli/src/actana-pair.ts
+++ b/packages/cli/src/actana-pair.ts
@@ -1,0 +1,548 @@
+// `actana pair` — the Core-side operator's half of short-code enrollment (#283).
+//
+//   actana pair new [--label <label>] [--ttl <duration>]
+//   actana pair ls [--json]
+//   actana pair revoke <target>
+//
+// **This runs on the machine that IS the Core.** That sentence is in the help
+// text and it is in this header for the same reason: there are two pairing
+// commands in one binary and they belong to opposite ends of the exchange.
+// `actana pair` mints a code on this Core and takes one back; `actana core
+// pair` (#285) is the *client* command that spends a code somebody read out to
+// you. One binary now carries both (#288), so the only thing separating them
+// is which words the operator types — the help on each has to make the machine
+// obvious or the mistake is silent.
+//
+// **The code exists in exactly one place: the terminal `pair new` printed it
+// on.** What is persisted is a keyed digest (`hashPairingCode`), so `pair ls`
+// *cannot* print a code — there is none in the store to print. That is the
+// design working, not an omission, and `pair-command.test.ts` asserts it rather
+// than trusting it.
+//
+// **What the fingerprint is for.** `pair new` prints the CA fingerprint beside
+// the code because the client's first dial is the one dial it cannot verify any
+// other way: it holds no certificate, so it has nothing pinned. A human reads
+// both out, and the client checks the CA it is presented against the
+// fingerprint *before* it sends the code (#280 step 3, #284). Printing it in
+// the conventional colon-separated upper-case hex is therefore part of the
+// contract — see `certFingerprintSha256`.
+//
+// The store is the one in `@actana/shared/pairing-store`, which lives there
+// rather than in `packages/core` precisely so this command and the daemon can
+// both drive it: this process mints sessions and revokes clients, the daemon
+// redeems them and enforces the revocations. Neither imports the other.
+
+import { randomUUID } from "node:crypto";
+import { certFingerprintSha256 } from "@actana/shared/core-cert-material";
+import { loadMaterialFromFile } from "@actana/shared/core-material-store";
+import { generatePairingCode } from "@actana/shared/pairing-code";
+import {
+  createPairingSession,
+  canRedeem,
+  isConsumed,
+  isRevoked,
+  PAIRING_SESSION_TTL_MS,
+  type PairingSession,
+} from "@actana/shared/pairing-session";
+import {
+  derivePairingCodeKey,
+  hashPairingCode,
+  pairingStorePath,
+  PairingStore,
+  PAIRING_SESSION_RETENTION_MS,
+  type PairedClient,
+} from "@actana/shared/pairing-store";
+import {
+  LOCAL_OPERATOR_PEER,
+  pairingAuditor,
+  type PairingAuditSink,
+} from "@actana/shared/pairing-audit";
+import { formatJson, formatTable } from "./cli-output.ts";
+import { EXIT_FAILURE, EXIT_OK, EXIT_USAGE } from "./exit-codes.ts";
+import type { ActanaCliDeps } from "./cli-deps.ts";
+
+export const PAIR_HELP = `actana pair — enroll a client on THIS machine's Core
+
+**You are on the Core.** These verbs mint and take back the pairing codes this
+Core hands out. The client end of the same exchange is \`actana core pair\`, and
+it runs on the machine being paired — not here.
+
+Usage
+  actana pair new [--label <name>] [--ttl <duration>]
+                                  mint a one-time code and print it
+  actana pair ls [--json]         pending codes, and the clients already paired
+  actana pair revoke <target>     unpair a client, or cancel a pending code
+
+Flags
+  --label <name>   what to call the machine being paired (default: unnamed)
+  --ttl <duration> how long the code stays good — \`30s\`, \`5m\`, \`2h\`
+                   (default: ${describeDuration(PAIRING_SESSION_TTL_MS)})
+  --json           machine-readable output — \`ls\`
+  -h, --help       show this help
+
+Minting a code
+  \`pair new\` prints three things: the code, this Core's CA fingerprint, and
+  when the code expires. Read the code AND the fingerprint out to whoever is at
+  the client — the client checks the fingerprint against the certificate this
+  Core presents before it sends the code, which is what makes that first dial
+  verifiable when the client has nothing pinned yet.
+
+  The code is printed once and never stored. This Core keeps only a keyed
+  digest of it, so nothing — including \`pair ls\` — can print it again. A lost
+  code is re-minted, not recovered.
+
+Revoking
+  \`revoke\` takes a certificate serial, a session id, or a label that matches
+  exactly one of either. Pointed at a paired client it unpairs the machine and
+  its credential stops working, including any core link it has open right now.
+  Pointed at a pending code it cancels the code before anyone spends it.`;
+
+/** What every verb here needs from the install: where this Core's material is. */
+export type PairCommandContext = {
+  /**
+   * Resolve the absolute path to this Core's `material.json`, or `null` after
+   * saying there is no Core installed here.
+   *
+   * A thunk rather than a string, so that the *order* lives in one place. Help
+   * is a question about this program and not about this machine — `actana pair
+   * --help` has to answer on a laptop that has never run `actana setup`, the
+   * same way `actana --help` does — and a context that had already resolved an
+   * install would have refused before this function was reached.
+   */
+  materialPath: () => string | null;
+  /**
+   * Where a revocation's audit record goes.
+   *
+   * Defaults to this CLI's stderr, and that default is the whole reason the
+   * seam exists. The daemon's auditor writes through `@actana/shared/log`,
+   * which is `console.log` — stdout. In a one-shot CLI whose stdout is a
+   * pairing code an operator may be piping, an audit line on stdout would be a
+   * line in the middle of the credential. So the record is the same record,
+   * built by the same `pairingAuditor` and reduced by the same field list, and
+   * only the sink differs.
+   */
+  audit?: PairingAuditSink;
+};
+
+/** Dispatch a `pair` sub-verb. `argv` is everything after `pair`. */
+export function runPairCommand(
+  deps: ActanaCliDeps,
+  argv: string[],
+  ctx: PairCommandContext,
+): number {
+  const [verb, ...rest] = argv;
+
+  if (verb === undefined || verb === "--help" || verb === "-h" || verb === "help") {
+    deps.out(PAIR_HELP);
+    // No verb is a question, not a mistake — the same answer `actana core`
+    // gives, and the same exit code: usage when nothing was asked for.
+    return verb === undefined ? EXIT_USAGE : EXIT_OK;
+  }
+  if (rest.includes("--help") || rest.includes("-h")) {
+    deps.out(PAIR_HELP);
+    return EXIT_OK;
+  }
+
+  switch (verb) {
+    case "new":
+      return pairNew(deps, rest, ctx);
+    case "ls":
+    case "list":
+      return pairLs(deps, rest, ctx);
+    case "revoke":
+      return pairRevoke(deps, rest, ctx);
+    default:
+      deps.err(`actana pair: unknown verb "${verb}".`);
+      deps.err("Verbs: new, ls, revoke. `actana pair --help` lists them.");
+      return EXIT_USAGE;
+  }
+}
+
+// ─── new ────────────────────────────────────────────────────────────────────
+
+function pairNew(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): number {
+  const flags = readFlags(rest, { label: "value", ttl: "value" });
+  if ("error" in flags) {
+    deps.err(`actana pair new: ${flags.error}.`);
+    return EXIT_USAGE;
+  }
+
+  const rawTtl = valueFlag(flags.values, "ttl");
+  const ttl = rawTtl === undefined ? PAIRING_SESSION_TTL_MS : parseDuration(rawTtl);
+  if (typeof ttl !== "number") {
+    deps.err(`actana pair new: ${ttl.error}.`);
+    return EXIT_USAGE;
+  }
+
+  const materialPath = ctx.materialPath();
+  if (materialPath === null) return EXIT_FAILURE;
+
+  const material = loadMaterialFromFile(materialPath);
+  if (!material) {
+    deps.err(`actana pair new: this Core has no pairing material at ${materialPath}.`);
+    deps.err("Run `actana setup` — pairing needs a CA to sign against.");
+    return EXIT_FAILURE;
+  }
+
+  const now = deps.now();
+  const label = valueFlag(flags.values, "label") ?? "";
+  const sessionId = randomUUID();
+  const code = generatePairingCode();
+  const session = createPairingSession({
+    id: sessionId,
+    label,
+    // The digest, never the code. The key is derived from this Core's bearer
+    // secret, which is in `material.json` and not in the pairing file — so a
+    // copy of the pairing file alone cannot be brute-forced back into codes.
+    codeHash: hashPairingCode({
+      key: derivePairingCodeKey(material.bearerSecret),
+      sessionId,
+      code,
+    }),
+    now,
+    ttlMs: ttl,
+  });
+
+  new PairingStore(pairingStorePath(materialPath)).createSession(session, now);
+
+  // stdout: the three facts a human reads out. stderr: everything explaining
+  // them, so a `pair new` that is being piped or screen-scraped stays three
+  // labelled lines and nothing else.
+  deps.err("Read the code AND the fingerprint out to the machine you are pairing.");
+  deps.err("The code is printed once — this Core keeps only a digest of it.");
+  deps.out(`Pairing code   ${code}`);
+  deps.out(`CA fingerprint ${certFingerprintSha256(material.caCert)}`);
+  deps.out(`Expires        ${absoluteTime(session.expiresAt)} (${relativeTime(session.expiresAt, now)})`);
+  if (label) deps.out(`Label          ${label}`);
+  deps.out(`Session        ${sessionId}`);
+  deps.err(`Cancel it before it is used with \`actana pair revoke ${sessionId}\`.`);
+  return EXIT_OK;
+}
+
+// ─── ls ─────────────────────────────────────────────────────────────────────
+
+function pairLs(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): number {
+  const flags = readFlags(rest, { json: "boolean" });
+  if ("error" in flags) {
+    deps.err(`actana pair ls: ${flags.error}.`);
+    return EXIT_USAGE;
+  }
+
+  const materialPath = ctx.materialPath();
+  if (materialPath === null) return EXIT_FAILURE;
+
+  const store = new PairingStore(pairingStorePath(materialPath));
+  const now = deps.now();
+  const pending = store.listSessions().filter((session) => canRedeem(session, now).ok);
+  const clients = store.listClients();
+
+  if (flags.values.json) {
+    // Field by field rather than a spread of the row, for the reason
+    // `redactPairingAuditEvent` is a list: a spread would publish whatever the
+    // stored shape grows next, and what it stores includes `codeHash`.
+    deps.out(
+      formatJson({
+        pending: pending.map((session) => ({
+          id: session.id,
+          label: session.label,
+          createdAt: session.createdAt,
+          expiresAt: session.expiresAt,
+          attempts: session.attempts,
+          attemptCap: session.attemptCap,
+        })),
+        clients: clients.map((client) => ({
+          certSerial: client.certSerial,
+          certSubject: client.certSubject,
+          label: client.label,
+          pairedAt: client.pairedAt,
+          certNotAfter: client.certNotAfter,
+          revokedAt: client.revokedAt,
+        })),
+      }),
+    );
+    return EXIT_OK;
+  }
+
+  deps.out("Pending codes");
+  if (pending.length === 0) {
+    deps.out("  None. `actana pair new` mints one.");
+  } else {
+    for (const line of formatTable(
+      ["LABEL", "CREATED", "EXPIRES", "ATTEMPTS", "SESSION"],
+      pending.map((session) => [
+        session.label || "(unnamed)",
+        absoluteTime(session.createdAt),
+        `${absoluteTime(session.expiresAt)} (${relativeTime(session.expiresAt, now)})`,
+        `${session.attempts} of ${session.attemptCap}`,
+        session.id,
+      ]),
+    )) {
+      deps.out(`  ${line}`);
+    }
+  }
+
+  deps.out("");
+  deps.out("Paired clients");
+  if (clients.length === 0) {
+    deps.out("  None. A client appears here once it redeems a code.");
+  } else {
+    for (const line of formatTable(
+      ["LABEL", "SUBJECT", "PAIRED", "STATUS", "SERIAL"],
+      clients.map((client) => [
+        client.label || "(unnamed)",
+        client.certSubject,
+        absoluteTime(client.pairedAt),
+        client.revokedAt === null ? "paired" : `revoked ${absoluteTime(client.revokedAt)}`,
+        client.certSerial,
+      ]),
+    )) {
+      deps.out(`  ${line}`);
+    }
+  }
+  return EXIT_OK;
+}
+
+// ─── revoke ─────────────────────────────────────────────────────────────────
+
+function pairRevoke(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): number {
+  const flags = readFlags(rest, {});
+  if ("error" in flags) {
+    deps.err(`actana pair revoke: ${flags.error}.`);
+    return EXIT_USAGE;
+  }
+  const [target, ...extra] = flags.positionals;
+  if (target === undefined) {
+    deps.err("actana pair revoke: a target is required — `actana pair revoke <serial|session|label>`.");
+    return EXIT_USAGE;
+  }
+  if (extra.length > 0) {
+    deps.err(`actana pair revoke: unexpected argument "${extra[0]}".`);
+    return EXIT_USAGE;
+  }
+
+  const materialPath = ctx.materialPath();
+  if (materialPath === null) return EXIT_FAILURE;
+
+  const store = new PairingStore(pairingStorePath(materialPath));
+  const now = deps.now();
+  const audit = pairingAuditor(ctx.audit ?? ((record) => deps.err(`pairing.revoke ${formatJson(record)}`)));
+
+  const clients = store.listClients().filter((client) => client.revokedAt === null);
+  const sessions = store.listSessions().filter((session) => canRedeem(session, now).ok);
+  const matched = [
+    ...clients.filter((client) => matchesClient(client, target)).map((client) => ({ client })),
+    ...sessions.filter((session) => matchesSession(session, target)).map((session) => ({ session })),
+  ];
+
+  if (matched.length === 0) {
+    deps.err(`actana pair revoke: nothing here matches "${target}".`);
+    deps.err("`actana pair ls` lists the pending codes and the paired clients.");
+    return EXIT_FAILURE;
+  }
+  if (matched.length > 1) {
+    // Never guess. Two machines paired under one label is the ordinary way this
+    // happens, and revoking the wrong one is not something an operator finds
+    // out about until the machine they meant to keep stops working.
+    deps.err(`actana pair revoke: "${target}" matches ${matched.length} of them.`);
+    for (const match of matched) {
+      deps.err(
+        "  " +
+          ("client" in match
+            ? `client ${match.client.certSerial} (${match.client.label || "unnamed"})`
+            : `pending code ${match.session.id} (${match.session.label || "unnamed"})`),
+      );
+    }
+    deps.err("Name the serial or the session id.");
+    return EXIT_FAILURE;
+  }
+
+  const only = matched[0]!;
+  if ("client" in only) {
+    const revoked = store.revokeClient(only.client.certSerial, now);
+    if (!revoked) {
+      deps.err(`actana pair revoke: ${only.client.certSerial} is no longer paired here.`);
+      return EXIT_FAILURE;
+    }
+    audit({
+      outcome: "revoked",
+      reason: "client",
+      sessionId: revoked.sessionId,
+      label: revoked.label,
+      peer: LOCAL_OPERATOR_PEER,
+      certSerial: revoked.certSerial,
+      at: now,
+    });
+    deps.out(`Unpaired ${revoked.label || "the client"} (${revoked.certSubject}, serial ${revoked.certSerial}).`);
+    // Said rather than implied. The daemon is a different process: it notices
+    // the revocation from the store and drops the link, and until it has, the
+    // certificate is still one its CA signed. "Stops working" is a promise this
+    // command cannot keep on its own, so it names who keeps it.
+    deps.out("Its certificate and its bearer stop working, and this Core closes any link it has open.");
+    return EXIT_OK;
+  }
+
+  const cancelled = store.cancelSession(only.session.id, now);
+  if (!cancelled || !isRevoked(cancelled)) {
+    deps.err(
+      cancelled && isConsumed(cancelled)
+        ? `actana pair revoke: ${only.session.id} was already redeemed — revoke the client instead.`
+        : `actana pair revoke: ${only.session.id} is no longer pending here.`,
+    );
+    return EXIT_FAILURE;
+  }
+  audit({
+    outcome: "revoked",
+    reason: "pending-session",
+    sessionId: cancelled.id,
+    label: cancelled.label,
+    peer: LOCAL_OPERATOR_PEER,
+    attempts: cancelled.attempts,
+    at: now,
+  });
+  deps.out(`Cancelled the pending code for ${cancelled.label || "an unnamed machine"} (${cancelled.id}).`);
+  deps.out("It cannot be redeemed. `actana pair new` mints another.");
+  return EXIT_OK;
+}
+
+/**
+ * Does this target name this client?
+ *
+ * A serial matches on a prefix — serials are 32 hex characters and nobody
+ * types one in full — but a label matches only in full. A label is prose the
+ * operator chose, and `laptop` prefix-matching `laptop-2` would make the
+ * ambiguity check above miss the very collision it exists to catch.
+ */
+function matchesClient(client: PairedClient, target: string): boolean {
+  const wanted = target.toLowerCase();
+  return (
+    client.label.toLowerCase() === wanted ||
+    client.certSerial.toLowerCase().startsWith(wanted) ||
+    client.certSubject.toLowerCase() === wanted
+  );
+}
+
+function matchesSession(session: PairingSession, target: string): boolean {
+  const wanted = target.toLowerCase();
+  return session.label.toLowerCase() === wanted || session.id.toLowerCase().startsWith(wanted);
+}
+
+// ─── flags, durations and times ─────────────────────────────────────────────
+
+type FlagKind = "value" | "boolean";
+type ReadFlags =
+  | { values: Record<string, string | true | undefined>; positionals: string[] }
+  | { error: string };
+
+/**
+ * Parse this verb's flags and positionals.
+ *
+ * `actana-cli.ts` has a `parseFlags` of its own and it is deliberately not
+ * reused: that one rejects every positional, because none of the machine verbs
+ * it was written for take one. `pair revoke <target>` does.
+ */
+function readFlags(argv: string[], spec: Record<string, FlagKind>): ReadFlags {
+  const values: Record<string, string | true | undefined> = {};
+  const positionals: string[] = [];
+  for (let i = 0; i < argv.length; i++) {
+    const token = argv[i]!;
+    if (!token.startsWith("--")) {
+      positionals.push(token);
+      continue;
+    }
+    const eq = token.indexOf("=");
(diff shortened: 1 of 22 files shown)


qcentic-dev:
## Acceptance criteria — #283, one line each

- [x] **`actana pair new [--label <label>] [--ttl <duration>]` creates a pending pairing session on this Core and
  prints: the `XXXX-XXXX` code, the CA fingerprint, and the expiry as both an absolute time and a relative
  "expires in 5 minutes"** — `packages/cli/src/actana-pair.ts` `pairNew`. Absolute is UTC ISO 8601, not local:
  the other end of this exchange is a person on a different machine, often in a different timezone, and a local
  time is only unambiguous to whoever is standing at the terminal that printed it. Tests: *prints a well-formed
  code from the unambiguous alphabet*, *prints the expiry as an absolute time AND a relative one*.

- [x] **TTL defaults to 5 minutes, from the shared constant in task 1 — not a literal in the CLI** —
  `PAIRING_SESSION_TTL_MS` is imported from `@actana/shared/pairing-session` and is the default for both the
  session and the help text's `(default: 5 minutes)`. Test: *defaults the TTL to the shared five-minute constant,
  not a literal here*, which asserts the minted session's window rather than the printed string.

- [x] **The CA fingerprint is SHA-256 over the CA certificate's DER, rendered colon-separated uppercase hex,
  computed from `PersistedMaterial.caCert`** — new `certFingerprintSha256` in
  `packages/shared/src/core-cert-material.ts`, called with `loadMaterialFromFile(...).caCert`. Tests: *is the
  conventional colon-separated upper-case hex SHA-256*, *agrees with what every other tool prints for the same
  certificate* (asserted against Node's `X509Certificate.fingerprint256`), *is over the DER, so PEM whitespace
  cannot move it*, and *prints a fingerprint that matches this Core's own CA* on the CLI side.

- [x] **`actana pair ls` lists pending pairing sessions (label, created, expires, attempts used of 5) and paired
  clients (label, subject, paired-at) in two clearly separated sections; an empty list says so rather than
  printing an empty table** — two `Pending codes` / `Paired clients` headings, each with `None.` and a next step
  when empty. The paired table carries a `STATUS` column beyond the criterion, so an operator can see a revocation
  they just made. Tests: *says the list is empty rather than printing an empty table*, *shows pending codes and
  paired clients in two separate sections*, *shows a pending code's label, created, expiry and attempts used of
  five*.

- [x] **`actana pair ls` never prints a pairing code — a pending session shows its label and expiry, not its
  secret** — structurally impossible rather than merely avoided: what the store holds is
  `hashPairingCode(...)`, a digest keyed by the bearer secret, so there is no code in the file to print. The
  `--json` payload names its fields one at a time so a later field on the stored shape cannot leak through a
  spread. Tests: *never prints a pairing code* and *leaves a code out of --json too*, both of which also assert the
  digest itself is absent.

- [x] **`actana pair revoke <target>` unpairs a client: the persisted client identity is removed or marked
  revoked, and its credential stops working — both its client certificate and its bearer** — the CLI stamps
  `revokedAt` via `PairingStore.revokeClient`; the daemon is what makes it bite. The **certificate** is refused at
  `clientCertGate` (every `/v1/…` request) and `coreLinkUpgradeGate` (every core-link upgrade), with `revoked`
  checked *ahead of* `authorized` — a revoked certificate is a valid certificate and TLS has no idea it was taken
  back, so a gate reading `authorized` first would serve every one of them. The **bearer** is refused at the `auth`
  frame through the `pair:<serial>` subject #282 mints. Tests: *stops the credential working — which is what the
  daemon reads* (CLI side), *refuses a revoked certificate even though the handshake accepted it*, *gives
  revocation no pre-auth exception*, *a revoked certificate never becomes a connection*, *a revoked bearer never
  passes the auth frame*.

- [x] **Revoking a client with a live core link closes that link rather than leaving it running until the next
  handshake** — `PtyCoreLinkServer.closeRevoked`, driven by `startPairingRevocationSweep`. **One caveat, stated
  rather than buried: the bound is one second, not instantaneous.** The CLI and the daemon are separate processes
  and the only channel between them is the pairing file, so the daemon polls it (`REVOCATION_SWEEP_MS = 1000`)
  rather than watching it — `fs.watch` is per-platform, misses writes behind a rename, and a rename is exactly how
  `PairingStore` writes. Every other client stays connected, and the closed connection's Session locks are
  released on the way out rather than stranding a Session nobody can write. Tests: *is closed rather than left
  running until the next handshake*, *closes a link identified only by the bearer it authenticated with*, *leaves
  every other client connected*, *closes every link that pairing holds, not just the first*.

- [x] **`actana pair revoke` also accepts a pending session, cancelling it before it is redeemed** — resolved by
  session id (or a prefix) or by label; `PairingStore.cancelSession` stamps the new `PairingSession.revokedAt`, and
  `canRedeem` checks it first, so the endpoint's next look at that session refuses it with the same body every
  other refusal gets. Tests: *cancels a pending session so its code can never be redeemed* (asserts
  `store.consume(...)` now answers `{ ok: false, reason: "revoked" }`), *cancels a pending session named by its
  label*, and *stops the code being redeemed* in the shared store suite.

- [x] **Revocation is audit-logged the same way pairing attempts are (task 2)** — the same
  `pairingAuditor`, the same `LOGGED_FIELDS` redaction, a `revoked` outcome added to the same `PairingOutcome`
  union. To make one module serve both writers, `core-pairing-audit.ts` moved to
  `packages/shared/src/pairing-audit.ts`; the CLI never loads `@actana/core`. **Two sinks, and the reason:** the
  daemon's default sink is `log.info`, which is `console.log` — stdout — and in a one-shot CLI whose stdout is a
  pairing code an operator may be piping, an audit line there would land in the middle of the credential. So the
  CLI writes the same already-redacted record to stderr, and the daemon additionally writes `pairing.revoked` into
  its own log when the sweep sees the row, which is the journal `pairing.attempt` goes to. Tests: *audit-logs the
  revocation through the same auditor the endpoint uses*, *audit-logs a cancelled session too*.

- [x] **`actana pair --help` and the top-level help list the verbs, and the help text says plainly that this runs
  on the Core machine** — `PAIR_HELP` opens with "**You are on the Core.**" and names `actana core pair` as the
  other end; the top-level `USAGE` row reads "This is the Core end: `actana core pair` is the client end". `pair
  --help` deliberately answers before the install is resolved, so it works on a machine that has never run `actana
  setup`. Tests: *lists the three verbs*, *says plainly which machine this runs on*, *says which machine it is for,
  in the top-level help and its own* (through the real dispatch), plus the repo's existing help ↔ dispatch drift
  guard, which now covers `pair`.

- [x] **Tests covering: `new` prints a well-formed code, a fingerprint matching the material's CA, and an expiry
  honouring `--ttl`; `ls` shows pending and paired and hides codes; `revoke` invalidates a paired client's
  credential; `revoke` on a pending session prevents its redemption** — all four, named above. 88 new tests in
  total: 37 in `packages/cli/src/__tests__/actana-pair.test.ts`, 20 in
  `packages/core/src/__tests__/core-pairing-revocation.test.ts`, 15 in
  `packages/core/src/__tests__/core-link-revocation.test.ts`, and 16 added to the existing machine-CLI,
  cert-material, pairing-session and pairing-store suites. `pnpm test` (all six stages), `pnpm -r typecheck` and
  `pnpm lint` are green.

### Boundaries this PR respected

- **`packages/sdk` is untouched** — that is #284's. `git diff --stat origin/feat/short-code-pairing..HEAD --
  packages/sdk` is empty. The one place it nearly forced a change is the `authError.reason` union, which lives
  there; a revoked bearer answers with the existing `expired` instead of a fourth reason, for the reason given in
  the PR description.
- **The verbs are in `packages/cli/src/actana-cli.ts`**, not the `packages/core` path #283's file pointers name —
  #288 unified the two `actana` binaries and moved the operator dispatch.
- **Base is `feat/short-code-pairing`**, per #283's Branching section and #280's Delivery section. Not `main`, not
  `beta/0.4.0`. Nothing was merged and the PR is a draft.

packages/core/src/core-pairing-revocation.ts
```diff
@@ -0,0 +1,168 @@
...
+   * Re-read the store. Returns the serials revoked **since the last read**, in
+   * their stored spelling, so a caller can act on the new ones without acting
+   * on every one it has already handled.
+   *
+   * A store that cannot be read leaves the set exactly as it was and returns
+   * nothing. Failing that way round is deliberate: an unreadable pairing file
+   * must never be read as "nobody is revoked", which is the one interpretation
+   * that hands a revoked client its access back.
+   */
+  refresh(): string[] {
```
mrfqcentic: **Blocking.** The fail-closed guarantee this doc comment states is not implemented, because `PairingStore` never throws.

`PairingStore.read()` (`packages/shared/src/pairing-store.ts:134-152`) swallows every `readFileSync` and `JSON.parse` failure and returns `emptyPairingRecords()`; the row filter below it (`o.clients.filter(isPairedClient)`) then silently drops any row whose shape it does not recognise. So `listClients()` cannot throw, this `catch` can never fire from the real collaborator, and a corrupt or truncated `pairing.json` is indistinguishable from "nobody is revoked" — which the comment correctly names as "the one interpretation that hands a revoked client its access straight back".

Concrete failure: a partial `writeFileSync` on ENOSPC still gets `renameSync`d into place (or an operator hand-edits the file), then the daemon restarts. `startPairingRevocationSweep`'s seeding `refresh()` reads an empty list, the revoked set is empty, and every previously revoked certificate is served again at `clientCertGate`, at `coreLinkUpgradeGate` and at the `auth` frame. A *running* daemon is protected because the set is additive — but boot is exactly when this has to hold.

The test at `core-pairing-revocation.test.ts:105` ("keeps what it knows when the store cannot be read") passes only because its fake throws. That is the wrong shape for this assertion: it pins the guarantee against a double that behaves differently from the collaborator that will actually be injected. `RevokedClientsPort` needs a way to tell "empty" from "unreadable" — a `readStrict()` on `PairingStore`, or a `read()` that rethrows — and the test should drive the real store over a deliberately corrupted file.

packages/cli/src/actana-pair.ts
```diff
@@ -0,0 +1,548 @@
...
+      deps.err(`actana pair: unknown verb "${verb}".`);
+      deps.err("Verbs: new, ls, revoke. `actana pair --help` lists them.");
+      return EXIT_USAGE;
+  }
+}
+
+// ─── new ────────────────────────────────────────────────────────────────────
+
+function pairNew(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): number {
+  const flags = readFlags(rest, { label: "value", ttl: "value" });
```
mrfqcentic, line 164: **Blocking.** `pairNew` reads `flags.values` and never looks at `flags.positionals`, and `pairLs` (line 225) does the same. `readFlags` at line 448 pushes every token that does not start with `--` into positionals, so both verbs silently discard them.

`actana pair new laptop` mints an **unlabelled** code and exits 0. So does `actana pair new -l laptop` — a single-dash token is a positional here. The operator reads the code out believing it is labelled `laptop`; `pair ls` then shows `(unnamed)`, and `actana pair revoke laptop` answers "nothing here matches".

`pairRevoke` already gets this right at line 316 (`unexpected argument "${extra[0]}"`), and the suite has a test named "rejects an unknown flag rather than ignoring it" — the same principle should apply to an unknown positional on the two verbs that take none.

packages/cli/src/actana-pair.ts
```diff
@@ -0,0 +1,548 @@
...
+
+// ─── revoke ─────────────────────────────────────────────────────────────────
+
+function pairRevoke(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): number {
+  const flags = readFlags(rest, {});
+  if ("error" in flags) {
+    deps.err(`actana pair revoke: ${flags.error}.`);
+    return EXIT_USAGE;
+  }
+  const [target, ...extra] = flags.positionals;
```
mrfqcentic, line 333: **Blocking.** The guard on the next line only rejects `undefined`, not an empty string.

With `target === ""`, `matchesClient` at line 419 evaluates `client.certSerial.toLowerCase().startsWith("")` → `true` for **every** client, and `matchesSession` at line 426 does the same for every session. So `actana pair revoke "$SERIAL"` in a script where `SERIAL` is unset matches everything; when exactly one non-revoked client (or one pending code) exists, the ambiguity check at line 306 does not fire and it is silently revoked with exit 0 and an `Unpaired …` message.

The ambiguity check is what is supposed to stop "revoking the wrong one is not something an operator finds out about until the machine they meant to keep stops working" — and this is the input that walks straight past it. Reject a blank target alongside `undefined`.

packages/cli/src/actana-pair.ts
```diff
@@ -0,0 +1,548 @@
...
+/** `300000` -> `5 minutes`. Used by the help text and the ceiling's refusal. */
+export function describeDuration(ms: number): string {
+  const units: Array<[number, string]> = [
+    [60 * 60 * 1000, "hour"],
+    [60 * 1000, "minute"],
+    [1000, "second"],
+  ];
+  for (const [size, name] of units) {
+    if (ms >= size) {
+      const count = Math.round(ms / size);
```
mrfqcentic: Non-blocking. `Math.round` on the first unit that fits means the relative half can misstate the code's life: `--ttl 90m` prints `(in 2 hours)` and `--ttl 90s` prints `(in 2 minutes)`.

The absolute half beside it is still exact, so nothing is unsafe — but the relative half is the one a human reads down a phone line, which is the reason it is printed at all. `Math.floor`, or a largest-two-units form (`1 hour 30 minutes`), would keep it honest. It also feeds the `--ttl` ceiling's refusal message and the help text's default, so the same rounding shows up in both.

packages/shared/src/core-cert-material.ts
```diff
@@ -526,3 +526,37 @@ function derToPem(label: string, der: Buffer): string {
...
+ * wrapped in a header, a footer and line breaks, and hashing that would make
+ * the fingerprint depend on line width and trailing whitespace. The DER is the
+ * certificate; everything else is packaging.
+ *
+ * Computed here rather than read off `node:crypto`'s `X509Certificate` so that
+ * the format lives in code with a test rather than in a property whose spelling
+ * this repository does not own — `core-cert-material.test.ts` asserts the two
+ * agree, which is the check that keeps the convention honest.
+ */
+export function certFingerprintSha256(certPem: string): string {
```
mrfqcentic, line 559: Confirming this one, since #284 pins its bootstrap dial to the exact bytes this returns: SHA-256 over `pemToDer(certPem)` — the DER, not the PEM — uppercased and grouped into byte pairs. `pemToDer` strips the armour and all whitespace, so line width and trailing newlines cannot move it, and `core-cert-material.test.ts` asserts both that and equality with Node's own `X509Certificate#fingerprint256`.

That last assertion is what makes the cross-task agreement real rather than asserted: #296's `fingerprintOf` is checked against `fingerprint256` by its own suite, so the two implementations agree transitively through a third party neither of them owns. This is the agreement #283 and #284 owed each other, and it is discharged.

packages/core/src/core-preauth-gate.ts
```diff
@@ -53,12 +53,23 @@ export const CLIENT_CERT_REFUSAL_MESSAGE =
...
+ * certificate to have had revoked, so nothing legitimate is turned away.
  */
 export function clientCertGate(opts: {
   pathname: string;
   authorized: boolean;
+  /** Did this connection present a certificate `actana pair revoke` took back? */
+  revoked?: boolean;
   isPreAuthPath?: PreAuthPathPredicate;
 }): ClientCertVerdict {
+  if (opts.revoked) return "refuse";
```
mrfqcentic, line 72: This ordering is the whole mechanism and it is right. A revoked certificate is a *valid* certificate — this Core's CA signed it, the handshake completed, and `authorized: true` is exactly what a revoked client arrives with — so a gate that read `authorized` first would serve every one of them.

Worth noting for the record that revocation correctly has no pre-auth exception either: a client here to redeem a code has no certificate to have had revoked, so nothing legitimate is turned away by checking this ahead of `isPreAuthPath`.

packages/core/src/pty-core-link-server.ts
```diff
@@ -1874,7 +1952,29 @@ export class PtyCoreLinkServer {
...
+    // (#283). Refused as `expired`, and that is a deliberate choice of an
+    // existing reason rather than a new one: the wire's three reasons live in
+    // `@actana/sdk`, a client that pinned a fourth would be a client this Core
+    // could not talk to, and — more to the point — "expired" is what a revoked
+    // credential *is* from where the client stands. Its validity ended. The
+    // client's own reconnect path takes it to re-pairing, which is exactly
+    // where an operator who revoked it wants it to go. The Core's log is where
+    // the real reason is written, because the operator reading that log is the
+    // person who revoked it.
+    if (this.revocation?.isBearerSubjectRevoked(result.sub)) {
```
mrfqcentic, line 1971: The bearer half checks out end to end, which is worth stating because it depends on three files agreeing: `core-pairing-routes.ts:354` mints `sub: pairingBearerSubject(issued.serial)`, `signBearer` carries `sub` into the payload alongside `{coreId, exp}`, `verifyBearer` returns it, and `certSerialFromBearerSubject` takes it apart here — all through the one `pair:` constant, so no two files spell the prefix independently.

Answering `expired` rather than a fourth reason is also the right call: the wire's three reasons live in `@actana/sdk`, which this PR correctly does not touch, and a client that pinned a fourth would be one this Core could not talk to. The real reason is in the Core's log, read by the person who revoked it.

mrfqcentic (changes requested):
## Review — #283 (`actana pair new | ls | revoke`)

Base verified: `feat/short-code-pairing` (`63ad949`). Head `b0c8d50`.

**Boundary: HELD.** `git diff --stat feat/short-code-pairing...feat/actana-pair-verbs -- packages/sdk` is empty. Nothing in `packages/sdk` is touched, which is #284's.

The design here is genuinely good, and the parts that carry the security argument are correct and proved by tests rather than by prose: the fingerprint format, the `revoked`-before-`authorized` gate order, the `pair:<serial>` bearer subject, and the sweep that closes an already-open link. I am requesting changes on three defects in new code, one of which quietly undoes a guarantee this PR documents.

### Acceptance criteria (#283)

| # | Criterion | Verdict | Enforcing lines |
|---|---|---|---|
| 1 | `pair new [--label] [--ttl]` creates a pending session; prints code, CA fingerprint, expiry absolute **and** relative | **Met** | `packages/cli/src/actana-pair.ts:196-218` — `generatePairingCode()`, `createPairingSession(...)`, `new PairingStore(pairingStorePath(materialPath)).createSession(...)`, then the three `deps.out` lines. Tests `actana-pair.test.ts:115,124,144` |
| 2 | TTL defaults to the shared constant, never a literal | **Met** | `actana-pair.ts:171` — `rawTtl === undefined ? PAIRING_SESSION_TTL_MS : parseDuration(rawTtl)`, imported from `@actana/shared/pairing-session`. `actana-pair.test.ts:130` asserts `expiresAt - createdAt === PAIRING_SESSION_TTL_MS` |
| 3 | Fingerprint = SHA-256 over the CA's **DER**, colon-separated uppercase hex, from `PersistedMaterial.caCert` | **Met** | `packages/shared/src/core-cert-material.ts:559` — `createHash("sha256").update(pemToDer(certPem))`, `.toUpperCase()`, `.match(/.{2}/g).join(":")`; called with `material.caCert` at `actana-pair.ts:214`. `core-cert-material.test.ts` asserts equality with Node's `X509Certificate#fingerprint256` and PEM-whitespace invariance |
| 4 | `pair ls` — two sections, pending (label/created/expires/attempts of 5) and paired (label/subject/paired-at); empty says so | **Met** | `actana-pair.ts:265-320`; `"  None. \`actana pair new\` mints one."` and `"  None. A client appears here once it redeems a code."`; `${session.attempts} of ${session.attemptCap}` |
| 5 | `pair ls` never prints a code | **Met**, structurally | The store holds only `codeHash` (`hashPairingCode`), and the `--json` payload at `actana-pair.ts:235-256` names every field one at a time rather than spreading the row. Tests at `:244` and `:254` |
| 6 | `revoke <target>` unpairs a client — cert **and** bearer stop working | **Met** | Row: `PairingStore.revokeClient`. Cert: `core-preauth-gate.ts:72` `if (opts.revoked) return "refuse";` ahead of `authorized`, and `coreLinkUpgradeGate(authorized, revoked)` at `:90`, wired at `pty-core-link-server.ts:2461` and `:2341`. Bearer: `pty-core-link-server.ts:1965` `isBearerSubjectRevoked(result.sub)`, against the `pair:<serial>` subject minted at `core-pairing-routes.ts:354` and returned by `verifyBearer` |
| 7 | A live core link is closed rather than left running | **Met** | `PtyCoreLinkServer.closeRevoked` at `pty-core-link-server.ts:2000` + `startPairingRevocationSweep` (1 s), armed at `core-entry.ts:648-653`. Bound stated, not implied. `core-link-revocation.test.ts:231-284` |
| 8 | `revoke` also cancels a **pending** session before redemption | **Met** | `PairingStore.cancelSession` stamps `revokedAt`; `canRedeem` checks `isRevoked` first (`pairing-session.ts:174`). `pairing-store.test.ts` asserts `consume(...)` → `{ ok: false, reason: "revoked" }` |
| 9 | Revocation is audit-logged the same way pairing attempts are | **Partially met** | Same `pairingAuditor`, same `redactPairingAuditEvent` field list, `outcome: "revoked"`, `peer: LOCAL_OPERATOR_PEER` — but `cmdPair` (`actana-cli.ts:779`) passes no `ctx.audit`, so `actana-pair.ts:328` falls back to the CLI's own **stderr**. Every `pairing.attempt` line goes to `log.info` → the daemon's stdout, which is what the supervisor captures. So a revocation never joins the audit stream an operator reads back. It is not invisible — `revokedAt` is durable in `pairing.json`, `pair ls` shows `revoked <time>`, and the sweep logs `pairing.revoked` — but it is not the same log, which is what the criterion asks for |
| 10 | `pair --help` and top-level help list the verbs and say plainly this is the Core machine | **Met** | `PAIR_HELP` opens `**You are on the Core.**`; `actana-cli.ts:178-179` — "This is the Core end: `actana core pair` is the client end". `actana-machine-cli.test.ts` reaches both through the real dispatch |
| 11 | Tests for `new` / `ls` / `revoke` / pending-cancel | **Met** | 37 in `actana-pair.test.ts`, 20 in `core-pairing-revocation.test.ts`, 15 in `core-link-revocation.test.ts`, plus 16 in the existing suites |

### Blocking

**B1 — the fail-closed guarantee on an unreadable store is not implemented** (`core-pairing-revocation.ts:104`, inline). This is the one that matters: `PairingStore.read()` swallows every failure and returns empty, so the `catch` can never fire from the real collaborator, and a corrupt `pairing.json` reads as "nobody is revoked" — the exact interpretation the comment says must never be reached.

**B2 — `pair new` and `pair ls` silently discard positionals** (`actana-pair.ts:164`, inline). `actana pair new laptop` mints an *unlabelled* code and exits 0.

**B3 — an empty target revokes an arbitrary client** (`actana-pair.ts:313`, inline). `""` prefix-matches everything.

### Non-blocking

- **`core-pairing-routes.ts:257`** — the explicit state ladder (`isExpired` / `isConsumed` / `isDead`) was not extended with `isRevoked`, so a cancelled session is only stopped inside `consume()` at `:309`. The security property still holds, but a guess against a just-cancelled session still burns an attempt and writes the store, and the audit line reads `reason: "wrong-code"` rather than `"revoked"` — so the log does not show that the cancellation is what stopped it. The comment above that ladder claims to write out the order #282 fixes; it now omits the newest gate.
- **`pty-core-link-server.ts:296`** — the injected-`createServer` option type does not declare `isRevokedSerial`, although the constructor passes it and `defaultCreateServer` declares it. A custom transport — the seam the `revocation` doc comment explicitly anticipates — has no typed way to receive it and silently loses both TLS-layer refusals, falling back to the `onConnection` check alone.
- **Test coverage gap.** Every certificate-revocation assertion is either against `clientCertGate`/`coreLinkUpgradeGate` as pure functions or against a `FakeWebSocketServer` handed a `CoreLinkPeer` directly. `peerCertSerial`, `isRevokedPeer`, `mountHttpRoutes(..., isRevokedSerial)` and the `wss.on("connection", (ws, req))` peer extraction in `defaultCreateServer` have **no test at all** (`git grep isRevokedSerial\\|peerCertSerial -- '**/__tests__'` is empty). So "a revoked certificate is refused on every `/v1/…` request" is asserted at the predicate and never through a real TLS server. #282's redemption suite already stands one up; one test that pairs, revokes and re-requests would close it.
- **`actana-pair.ts:524`** — `describeDuration` rounds, so `--ttl 90m` prints `(in 2 hours)` and `--ttl 90s` prints `(in 2 minutes)`. The absolute half is still right, but the relative half is the one read out loud.
- **`actana-pair.test.ts:118`** — `/^[A-Z2-9]{4}-[A-Z2-9]{4}$/` admits `O`, `I` and `L`; the next line's `PAIRING_CODE_ALPHABET` check is what actually holds the property, so the regex is redundant rather than wrong.
- **`MAX_PAIRING_TTL_MS`** — the ceiling is fine, but the stated reason is not the mechanism: `prune()` uses `settled = consumedAt ?? expiresAt`, and `now - expiresAt` is negative for the whole life of a pending session, so a longer TTL could never prune a *live* session out from under its holder.
- **Follow-up, not this PR:** `token regenerate` replaces `bearerSecret` and the CA but leaves `pairing.json` intact, so pending codes silently stop matching (the digest is keyed on the old secret) and `pair ls` keeps listing clients whose certificates the new CA never signed. That interaction arrives with #282's store rather than with these verbs, but `pair ls` / `pair revoke` are the surfaces that now display it — worth a line in #280 or #287.

### Cross-PR

This PR renames `packages/core/src/core-pairing-audit.ts` → `packages/shared/src/pairing-audit.ts`. #296 imports the old path at `packages/sdk/src/__tests__/core-pairing.test.ts:40`. `git merge-tree` merges the two branches **cleanly** — there is no textual conflict — so this surfaces only as a `pnpm -r typecheck` failure after both land. Whichever merges second needs that one-line update; see the note on #296.

**Verdict: request changes.** B1–B3 are three defects in new code; B1 undoes a security guarantee this PR documents. Everything else above is a comment, not a gate.

packages/core/src/core-pairing-revocation.ts
qcentic-dev: Fixed — `de0ed21` and `e74aab7`. You were right that the `catch` could never fire from the real collaborator, and right that boot is where it matters.

**`PairingStore` can now say "I could not read this."** `readStrict()` throws where `read()` swallows: a permission failure, an I/O error, a half-written document, a `sessions`/`clients` that is not a list, or a single row whose shape this build does not recognise. **A file that is simply absent is still empty**, in both moods — a Core that has never paired anything has no `pairing.json`, and failing closed on that would refuse every client on every fresh Core.

It is one parser in two moods rather than two parsers, so the lenient and the strict reader cannot drift about what a *good* file contains, and `read()` still salvages what it can one dropped row at a time. There is a test pinning the two against each other.

**Fail-closed now means revoke-everything, not "keep what we knew."** `refresh()` returns `{ ok: false, error }` and sets a flag; while it is set, `isRevoked` answers `true` for every serial and `isBearerSubjectRevoked` for every `pair:<serial>` bearer. Cleared by the next successful read, so it is a fact about the store as it is now rather than a latch.

Two things are deliberately still served while failing closed, and I would rather you push back on these than have them buried:

- **A connection presenting no certificate at all.** "No certificate" and "a revoked certificate" are different facts answered by different gates, and `clientCertGate` already refuses an uncertificated caller everywhere except the pairing endpoint. Answering `true` here would close the one door out of a corrupt store.
- **A bearer with no `pair:` subject** — the hand-carried registration blob. It is not a pairing, so no row about it could have gone unread, and it is the credential the operator's own Panel is holding while they go and repair the file.

**`closeRevoked` stopped taking a list.** A list can only carry serials that were newly *named*, and entering fail-closed names none — so it asks `PairingRevocations` per connection instead. Same code path for both, and a live link now gets exactly the answer the same client would get at the door.

**And the CLI stopped being able to destroy the evidence** (`c7895f9`). `createSession` reads the whole document, adds a session and writes it back, so `pair new` against a corrupt file *replaced* it — which would have ended the daemon's refusal by forgetting who was revoked. All three verbs open the store through `readStrict` and decline, naming the file and saying not to delete it.

**On the test shape.** You were right that the old one pinned the guarantee against a double that behaves differently from the real collaborator. The port is now `readStrict`, the unit tests drive `PairingStore` itself over a deliberately corrupted file (and over an absent one, and over an unrecognised row), and `core-revocation-transport.test.ts` asserts it end to end: a real paired client, a truncated write renamed into place, and a `403` on the route that worked a moment earlier.

packages/cli/src/actana-pair.ts
qcentic-dev: Fixed — `c7895f9`. Both verbs now refuse a positional, and the message points at the flag: `actana pair new laptop` answers `unexpected argument "laptop"` and suggests `--label laptop`.

The `-l laptop` case is closed at the parser rather than at each verb: a single-dash token is now an unknown option, not a positional, so it says `unknown option: -l` instead of silently minting from two arguments that both look used. `pair revoke` keeps its own `<target>` positional and its existing second-positional refusal.

Tests: *refuses a bare positional instead of minting an unlabelled code*, *refuses a single-dash flag rather than reading it as a positional*, and *refuses a bare positional* on `ls` — each asserting the store is still empty afterwards, since the defect was that something got minted.

packages/cli/src/actana-pair.ts
qcentic-dev: Fixed — `c7895f9`. Blank is now rejected alongside `undefined`, and `.trim()` covers `"   "` too, so the unset-variable case exits 2 with `a target is required` and revokes nothing.

Your reading of it is exactly right and it is worth restating: the ambiguity check is the whole defence against revoking the wrong machine, and a one-client Core is precisely where it does not fire. Two tests pin it — one with a single paired client, one with a single pending code — each asserting the thing that was at risk is untouched afterwards.

packages/cli/src/actana-pair.ts
qcentic-dev: Taken — `520b6f2`. It floors now and carries a second unit: `--ttl 90m` prints `(in 1 hour 30 minutes)` and `--ttl 90s` prints `(in 1 minute 30 seconds)`.

The second unit is dropped when it is zero, so the ordinary strings are unchanged — `5 minutes`, `30 seconds`, `24 hours` — which keeps the help text's default and the ceiling's refusal reading the way they did. There is a property-ish test over a spread of durations asserting the spoken form never exceeds the real one, since overstating is the direction that costs something.

While in there I also corrected `MAX_PAIRING_TTL_MS`'s stated reason, which was your other note: the number is right but pruning was never the mechanism, because `prune` measures from `consumedAt ?? expiresAt` and a pending session's is always in the future. It now says what it actually is — a policy bound on how long a pre-auth secret stays worth stealing — and explains that reusing the retention constant is about keeping one day-long horizon in the file rather than two.

qcentic-dev:
## Review addressed — B1, B2, B3 fixed; four of the non-blocking notes taken

Thanks for this. B1 in particular was a real hole and the diagnosis was exact: the `catch` could never fire from the real collaborator, and boot is where the guarantee has to hold.

Five commits on top of `b0c8d50`:

| Commit | What |
| --- | --- |
| `de0ed21` | `fix(shared)`: `PairingStore.readStrict()` — tell an unreadable store apart from an empty one |
| `e74aab7` | `fix(core)`: an unreadable store is **everything-revoked**; `closeRevoked` asks instead of being told; `isRevokedSerial` on the injected transport's options; the end-to-end transport suite |
| `c7895f9` | `fix(cli)`: B2 and B3, plus all three verbs declining to run against a store they cannot read |
| `25264ec` | `fix(core)`: `isRevoked` joins the redemption ladder |
| `520b6f2` | `fix(cli)`: the spoken expiry floors and carries a second unit; the `MAX_PAIRING_TTL_MS` reason corrected; the code-shape regex built from the alphabet |

### Blocking

**B1 — fail-closed.** `readStrict()` throws where `read()` swallows; an absent file is still empty in both moods, because failing closed on "this Core has never paired anything" would refuse every client on every fresh Core. `refresh()` returns a result rather than a list, and a failed read makes `isRevoked` answer `true` for every serial until the store reads again. Two things stay served while failing closed, both deliberate and both argued in the thread: a connection presenting no certificate at all (that is `clientCertGate`'s question, and the pairing endpoint is the door out of a corrupt store), and a bearer with no `pair:` subject (not a pairing, so no row about it could have gone unread — and it is what the operator's own Panel holds while they fix the file).

`closeRevoked` stopped taking a list, because entering fail-closed names no serials at all. It asks `PairingRevocations` per connection, so a live link gets exactly the answer the same client would get at the door.

The CLI half matters too: `createSession` reads-adds-writes, so `pair new` against a corrupt file *replaced* it and would have ended the daemon's refusal by forgetting who was revoked. All three verbs now decline, name the file, and say not to delete it.

**B2 — discarded positionals.** `new` and `ls` refuse them and point at `--label`. A single-dash token is now an unknown option at the parser rather than a positional, so `-l laptop` says so.

**B3 — blank target.** Rejected alongside `undefined`, `.trim()`ed, with tests on both the one-client and the one-pending-code shapes where the ambiguity check cannot fire.

### Non-blocking — taken

- **The redemption ladder** now has an `isRevoked` rung (`25264ec`). A guess against a cancelled session no longer spends one of its five attempts or writes the store, and the audit line says `revoked` rather than `wrong-code`. `consume()` still refuses — the rung is the reported reason, not the enforcement. The file header's stated order was updated with it.
- **`isRevokedSerial` on the injected `createServer` options** (`e74aab7`).
- **The transport coverage gap** (`e74aab7`). New `core-revocation-transport.test.ts`: pairs a client over a real TLS server built by the default factory, revokes it the way the CLI does, and dials back with that certificate — `403` on `/v1/…`, no `authOk` on the core link, other clients unaffected, and the pairing endpoint still reachable. It also covers the fail-closed path end to end. That exercises `peerCertSerial`, `isRevokedPeer`, `mountHttpRoutes(…, isRevokedSerial)` and the `wss.on("connection", (ws, req))` extraction, none of which had a test.
- **`describeDuration`** floors and carries a second unit (`520b6f2`).
- **`MAX_PAIRING_TTL_MS`'s stated reason** corrected — you were right that pruning was never the mechanism.
- **The code-shape regex** is built from `PAIRING_CODE_ALPHABET` instead of a hand-typed `[A-Z2-9]`.

### Deferred, with reasons

- **`token regenerate` leaving `pairing.json` intact.** Agreed, and agreed it is not this PR: the interaction arrives with #282's store, and fixing it means deciding what regeneration does to pending codes and paired rows — a design call for #280/#287, not a verb change. Happy to open it as a task issue against #280 if you want it tracked now rather than in #287's sweep.
- **#296's import of the old `core-pairing-audit` path.** Left alone because it is in `packages/sdk`, which is #284's and off-limits here. It is a one-line update for whichever of us merges second; I have not touched it.

### State

Base is still `feat/short-code-pairing`, the PR is still a **draft**, nothing was merged, `main` is untouched, and `git diff --stat feat/short-code-pairing...feat/actana-pair-verbs -- packages/sdk` is still empty.

`pnpm test` (all six stages), `pnpm -r typecheck` and `pnpm lint` are green locally.

packages/core/src/core-pairing-revocation.ts
```diff
@@ -0,0 +1,246 @@
...
+   * pairing-issued credential. That is the fail-closed direction, and it is
+   * chosen knowing what it costs: a Core whose `pairing.json` is corrupt stops
+   * serving every client it ever paired. The alternative costs more — a
+   * half-written file, a hand-edit, or a row this build does not recognise
+   * would silently un-revoke every certificate an operator has taken back.
+   */
+  refresh(): RevocationRefresh {
+    let rows: PairedClient[];
+    try {
+      rows = this.clients.readStrict().clients;
```
mrfqcentic, line 144: This is the line the finding turned on, and it is now right. `readStrict()` rather than `listClients()`, and the port type at `:48` declares only `readStrict` — so the lenient reader that silently answers "empty" is not reachable from this module at all, rather than merely not called. That is the difference between a fixed bug and a bug that comes back the next time someone adds a caller.

The catch below can now actually fire, which is what makes the guarantee in the doc comment true: `pairing-store.ts:179` throws in strict mood on a non-ENOENT read failure, invalid JSON, a non-object document, a list that is not a list, or a single row this build does not recognise — that last one being exactly what a revoked client's row would look like to a reader that must not miss one.

ENOENT staying empty in both moods is the right exception and the only one needed: a Core that has paired nobody has no `pairing.json`, and failing closed on its absence would refuse every client on every fresh Core.

packages/core/src/core-pairing-revocation.ts
```diff
@@ -0,0 +1,246 @@
...
+   * `null` — no peer certificate at all — is not, even while failing closed.
+   * "No certificate" and "a revoked certificate" are different facts and
+   * different gates answer them: `clientCertGate` refuses an uncertificated
+   * caller everywhere but the pairing endpoint, and that endpoint is the one an
+   * operator needs reachable to recover. Answering `true` here would close the
+   * only door out of a corrupt store.
+   */
+  isRevoked(certSerial: string | null | undefined): boolean {
+    if (!certSerial) return false;
+    if (this.failClosed) return true;
```
mrfqcentic, line 182: And this is where fail-closed stops being a flag and starts being a refusal. Placed **before** the set lookup, so it covers serials whose rows were never read — which is the entire set when the file is unreadable. The previous shape could only refuse what it had successfully read, so "keep what we knew" would have been no defence at boot, the moment that matters.

The two carve-outs are both correct. `!certSerial → false` on the line above keeps the pairing endpoint reachable for a client with no certificate, which is the door an operator needs to re-pair a machine out of a corrupt store — closing it would make the failure unrecoverable without shell access. And `isBearerSubjectRevoked` leaving a non-`pair:` bearer alone is right for the same reason it is safe: it is not a pairing, so no unread row could be about it.

One small mismatch worth a sentence, not a change: the comment on `refresh` says "every pairing-issued credential", but this line cannot tell a pairing serial from a certificate `actana setup` issued, so while failing closed it refuses those too. Safe direction, and #287 removes the hand-carry — but the next reader will trace it.

packages/shared/src/pairing-store.ts
```diff
@@ -130,29 +130,107 @@ export type PairingConsumeOutcome =
...
+
+  /**
+   * The one reader, in its two moods.
+   *
+   * Written once rather than twice on purpose: the lenient and the strict
+   * reader must agree about what a *good* file contains, and two parsers would
+   * be two chances to disagree — the strict one accepting a document the
+   * lenient one silently trimmed, or the reverse.
+   */
+  private parse(strict: boolean): PairingRecords {
```
mrfqcentic, line 179: One parser in two moods is the right shape, and better than the `readStrict()`-as-a-second-parser I had in mind when I wrote the finding. Two readers would have been two chances to disagree about what a *good* file contains — the strict one accepting a document the lenient one had quietly trimmed, or the reverse — and that disagreement would be invisible until it mattered.

`rows()` below is where the moods part and it is the only place they do: lenient filters, strict throws naming the index and the file. Keeping `read()` salvaging row-by-row is also right — the callers that use it are display and mutation paths where dropping an unrecognised row beats refusing to run.

packages/cli/src/actana-pair.ts
```diff
@@ -0,0 +1,639 @@
...
+ *
+ * That last one is not hypothetical damage. The daemon fails closed on exactly
+ * this file (`core-pairing-revocation.ts`), so a Core with an unreadable
+ * `pairing.json` is refusing every client it ever paired — and a `pair new`
+ * that quietly rewrote the file would end that refusal by forgetting who was
+ * revoked, handing every revoked certificate its access back. Recovery has to
+ * be repairing the file, so these verbs decline to be the thing that destroys
+ * it.
+ */
+function openStore(deps: ActanaCliDeps, materialPath: string, verb: string): PairingStore | null {
```
mrfqcentic, line 454: This one I did not ask for and it is the more serious half of B1.

`createSession` reads the whole document, adds a row and writes it back, so `pair new` against a corrupt `pairing.json` would have replaced it — and since the daemon now fails closed on exactly that file, the rewrite would have *ended* the refusal by forgetting who was revoked. An operator following the obvious instinct ("the file is broken, let me just make a new code") would have handed every revoked certificate its access back, while appearing to fix things.

Routing all three verbs through `readStrict` and declining is the right answer, and the refusal message doing the work — naming the file, saying what the daemon is currently doing, and saying explicitly not to delete it — is what makes the recovery path a person can actually follow. Good catch on your own patch.

packages/cli/src/actana-pair.ts
```diff
@@ -0,0 +1,639 @@
...
+  for (let i = 0; i < argv.length; i++) {
+    const token = argv[i]!;
+    if (!token.startsWith("-")) {
+      positionals.push(token);
+      continue;
+    }
+    // A single-dash token is a flag this build does not know, never a value.
+    // Read as a positional, `actana pair new -l laptop` would have minted an
+    // unlabelled code from two arguments that both look like they were used.
+    if (!token.startsWith("--")) return { error: `unknown option: ${token}` };
```
mrfqcentic, line 516: Closing the `-l laptop` case at the parser rather than at each verb is the better fix, and worth saying why: the defect was not that two verbs forgot a check, it was that the token classifier had a category ("anything not starting with `--`") that no caller wanted. Fixing the classifier means the next verb added here inherits the correct behaviour instead of having to remember.

`-h` still reaches the help because `runPairCommand` checks `rest.includes("-h")` before dispatching, and `revoke`'s `<target>` is unaffected since a serial or a label does not start with a dash.

packages/cli/src/actana-pair.ts
```diff
@@ -0,0 +1,639 @@
...
+    return EXIT_USAGE;
+  }
+  const [target, ...extra] = flags.positionals;
+  // Blank as well as absent, and that is not tidiness. `""` prefix-matches
+  // every serial and every session id, so `actana pair revoke "$SERIAL"` in a
+  // script where `SERIAL` is unset matches everything — and on a Core with
+  // exactly one client the ambiguity check below does not fire, so it revokes
+  // it and exits 0. That check is the whole defence against revoking the wrong
+  // machine, and this is the input that walks straight past it.
+  if (target === undefined || target.trim() === "") {
```
mrfqcentic, line 340: Fixed, and `.trim()` is the right extra — `"   "` from a quoted shell variable is the same failure with a space in it.

The tests pin the two shapes where this actually bit: a single paired client and a single pending code. That matters because the ambiguity check at `:313` *would* have caught a blank target on a Core with two of anything — the hazard was specific to the one-client case, which is also the most common one.

packages/core/src/core-pairing-routes.ts
```diff
@@ -247,17 +256,27 @@ export function createCorePairingRequestHandler(opts: CorePairingRoutesOptions):
...
+    // Revocation joins the ladder rather than being left to `consume()` below
+    // (#283). It is caught at the same layer as the rest for two reasons that
+    // are about the operator, not the attacker: a guess against a cancelled
+    // session should not cost that session an attempt it will never use, and
+    // the audit line should say `revoked` rather than `wrong-code` — otherwise
+    // the log does not show that the operator's own cancellation is what
+    // stopped the redemption. `consume()` still refuses it, and must: this is
+    // the reported reason, not the enforcement.
     const at = now();
+    if (isRevoked(session)) return refuse(res, audit, session, peer, "revoked");
```
mrfqcentic, line 274: Right position in the ladder, and the header comment renumbered to match — which is the part that keeps this honest, since the omission was only visible because that comment claimed to write out the order in full.

Both stated reasons hold: a guess against a cancelled session no longer costs it an attempt it will never use, and the audit line now says `revoked`, so the log shows that the operator's own cancellation is what stopped the redemption rather than attributing it to a wrong code. `consume()` still refuses independently, which is the correct relationship — this is the reported reason, that is the enforcement.

mrfqcentic (approved):
## Re-review — #283, at `520b6f2`

Previous review: `b0c8d50`, changes requested on B1/B2/B3. Five new commits. Base tip is unchanged (`feat/short-code-pairing` @ `63ad949`), so this is a clean re-read of the same base.

**Boundary: still HELD.** `git diff --stat 63ad949...520b6f2 -- packages/sdk` is empty. The one new file (`core-revocation-transport.test.ts`) is in `packages/core`.

**CI on this head: green.** All 19 reporting checks pass — Typecheck, Lint, Unit Tests, Conventions, Dependency Audit, Secret Scan, both installer E2Es, the Panel service-seam E2E, both release-tarball smokes, Promotion gate, Train rules. The only non-passes are `skipping` on the image-publish and train-tag jobs, which is the expected shape for a PR that is not a train build.

### B1 — the fail-closed guarantee. **Fixed**, and proved from the enforcing lines rather than the reply.

The old defect was that `PairingRevocations.refresh()` caught an exception `PairingStore` could never throw. Three things had to become true, and each is now readable in one place:

1. **The store can say "I could not read this."** `pairing-store.ts:179` is one parser in two moods. `readStrict()` (`:167`) calls `parse(true)`; `read()` (`:143`) calls `parse(false)` inside a catch. In strict mood a non-ENOENT `readFileSync` failure, invalid JSON, a non-object document, a `sessions`/`clients` that is not a list, or a *single* row failing `isPairingSession`/`isPairedClient` all throw with the path in the message. ENOENT is empty in **both** moods — correct, and the one exception that had to be there: a Core that has paired nobody has no `pairing.json`, and failing closed on its absence would refuse every client on every fresh Core. One parser rather than two is the right call: the lenient and strict readers cannot drift about what a good file contains, and `rows()` (`:206`) is the single place they part.
2. **`refresh` reads through it and fails closed.** `core-pairing-revocation.ts:144` — `rows = this.clients.readStrict().clients`. The port type (`:48`) now declares `readStrict()`, not `listClients()`, so the lenient reader is not reachable from here by construction. On throw: `failClosed = true`, `{ ok: false, error }`, and one `log.error` on the crossing rather than one per second.
3. **Fail-closed actually refuses.** `:182` — `if (this.failClosed) return true;` sits *before* the set lookup in `isRevoked`, so every serial is revoked while the flag is set, including ones whose rows were never read. That is the point: the old code could only refuse serials it had successfully read, which is exactly the set that is empty when the file is unreadable.

The two carve-outs are right and I would have argued for both. A connection presenting **no certificate** is still not "revoked" (`:181`) — "no certificate" and "a revoked certificate" are different facts, `clientCertGate` already refuses an uncertificated caller everywhere except the pairing endpoint, and that endpoint is the door an operator needs to re-pair a machine out of a corrupt store. A bearer with **no `pair:` subject** is untouched (`:196`) — it is not a pairing, so no unread row could be about it.

**The test shape is corrected too**, which was half the finding. `core-pairing-revocation.test.ts:191` drives a real `PairingStore` over a deliberately corrupted `pairing.json`, `:212` over an absent one, `:225` over an unrecognised row; `:320` asserts fail-closed is seeded at boot before the first request. The old test's throwing fake is no longer what the guarantee rests on.

**And the sweep now expresses the change a list could not.** `closeRevoked()` takes no arguments (`pty-core-link-server.ts:2005`) and re-asks `PairingRevocations` per connection; `startPairingRevocationSweep` fires on fresh serials *or* on crossing into fail-closed (`core-pairing-revocation.ts:232-241`), and correctly not on crossing back out — nothing became revoked by the file becoming readable. One code path, and a live link now gets the same answer the same client would get at the door.

**You found one I had not.** `openStore` (`actana-pair.ts:454`) routes all three verbs through `readStrict` and declines. The reasoning is right and it is the more serious half: `createSession` reads the whole document, adds a row and writes it back, so `pair new` against a corrupt file would have *replaced* it — ending the daemon's refusal by destroying the record of who was revoked. The refusal message says not to delete the file and why. Good catch on your own patch.

### B2 — positionals silently discarded. **Fixed.**

`actana-pair.ts:174` and `:243` refuse a positional on `new` and `ls`, and the `new` message points at the flag (`--label laptop`). The `-l laptop` half is closed at the parser instead of at each verb (`:516`): a single-dash token is now `unknown option`, not a positional. That is the better place for it — it fixes the shape rather than the two instances. `pair revoke` keeps its `<target>` positional and its second-positional refusal. Three tests, each asserting the store is still empty afterwards, since minting was the damage.

### B3 — blank target revokes an arbitrary client. **Fixed.**

`actana-pair.ts:340` — `target === undefined || target.trim() === ""`. `.trim()` covers `"   "` as well, so the unset-variable case exits 2 and revokes nothing. Two tests, one with a single paired client and one with a single pending code — the two shapes where the ambiguity check does not fire, which was the whole hazard.

### The non-blocking notes, all addressed

- **Routes ladder** — `core-pairing-routes.ts:274` now checks `isRevoked(session)` first, so a guess against a cancelled session no longer spends one of its five attempts and the audit line reads `revoked` rather than `wrong-code`. The header comment was renumbered to match, which is what made the omission visible in the first place.
- **`createServer` option type** — `isRevokedSerial` is now declared on the injected-factory signature (`pty-core-link-server.ts:308`), so a custom transport can receive it.
- **Transport test coverage** — `core-revocation-transport.test.ts` is the gap closed: a real `https` server on real `generateCertMaterial` certificates, a real client certificate, and the actual `getPeerCertificate` path. It asserts a revoked certificate refused on a route that worked a moment earlier, refused at the upgrade, *not* taking another paired client down with it, and still able to reach the pairing endpoint to re-pair — plus the unreadable-store case end to end and its recovery.
- **`describeDuration`** — floors and carries a second unit, dropping it when zero, so `90m` reads `1 hour 30 minutes` and the ordinary strings are unchanged. The property test asserts the spoken form never exceeds the real one, which is the direction that costs something.
- **`MAX_PAIRING_TTL_MS`** — the comment now says what the bound actually is (policy on a pre-auth secret) instead of the pruning story, which was never the mechanism.

### One observation, not a gate

While failing closed, `isRevoked` answers `true` for **any** non-null serial, which includes a client certificate issued by `actana setup` and hand-carried — one that was never a pairing and has no row in this file. The doc comment says "every pairing-issued credential"; the code refuses strictly more than that. It is the safe direction and #287 removes the hand-carry anyway, so I would leave the behaviour alone — but the comment on `refresh` slightly understates what happens, and a sentence there would save the next reader the trace.

Worth stating plainly for whoever operates this: a corrupt `pairing.json` now takes every paired client offline until it is repaired. That is the correct trade and it is argued for in the code, the recovery door (the pairing endpoint) stays open, and the CLI refuses to be the thing that destroys the evidence. It is a real operational consequence rather than a hidden one, which is the standard it needed to meet.

### Acceptance criteria

All eleven now **met**. AC 9 ("revocation is audit-logged the same way pairing attempts are") was my one partial last round and I am satisfied: the record was always built by the same `pairingAuditor` under the same field list, the sink difference is argued from stdout being a pairing code, and the daemon separately logs `pairing.revoked` and — now — `core-pairing.revocation.unreadable` with its effect. The trail an operator reads back is complete.

**Verdict: approve.**

Merge this **first** of the two-PR wave: it owns the `core-pairing-audit.ts` → `packages/shared/pairing-audit.ts` move and the rest of the `packages/shared` surface, so landing it first means that move happens once and #296 rebases onto a settled `packages/shared`.

PR #298, opened 2026-08-20
feat(cli): pairing — actana core pair replaces the blob paste

### LINKED ISSUE #285: Pairing: CLI client command — actana core pair replaces the blob paste

qcentic-dev:
Closes #285. Part of #280.

Base is `feat/short-code-pairing`, the feature branch for #280 — not `main` and not the train. This sits on top of the
pairing server (#282), the `actana pair` verbs (#283) and the SDK's client pairing (#284).

## What this is

`actana core add <name> [file]` reads a base64 credential off a file, a pipe or a paste. This replaces that gesture with
`actana core pair <name> <address> <code>`: the operator types an address and the eight-character code somebody read out
to them, the SDK generates a key pair on this machine and gets a certificate signed, and what lands is the same file, in
the same place, at the same mode.

**Everything after the credential lands is unchanged.** `core-resolution.ts`, `core-connection.ts` and every verb built
on them cannot tell a paired credential from a pasted one — the test resolves the stored entry through `resolveCore`
rather than re-reading the file it just watched being written, because resolution is what `ls`, `use`, `rm`, `status`,
`shell` and `exec` actually go through. **The paste path is untouched**; #287 removes it everywhere at once.

## The fingerprint, which is the whole security argument

Either `--fingerprint` carries the one `actana pair new` printed, or — on a terminal — the command prints the
fingerprint the Core presents and asks the operator to compare the two. With neither, it refuses. Absent is not waived.

What the operator confirms is handed back to `pairWithCore`, which dials again and recomputes it, so the value they
agreed to is enforced on the connection the code actually crosses and not merely on the one they were shown. Every
refusal happens with the code unsent, and the tests assert that at the seam — `pair` is never reached — rather than by
reading the prose.

## No crypto here

The key pair is born inside the SDK call and its private half never crosses the wire. This package supplies an address,
a code and an expected fingerprint, and turns what comes back into a registry entry, a sentence and an exit code. The
one parser it reaches for is the SDK's own `parsePairingTicket` — #281's normaliser through a published door, so the
CLI imports nothing from `@actana/shared` for it and has no copy of the Core's alphabet.

## Exit codes

Every failure `@actana/sdk/core-pairing.ts` distinguishes gets a code of its own in `exit-codes.ts`, from 10 upward, and
the mapping switches over the SDK's union exhaustively — a failure added there stops this compiling until it is given a
number. A malformed code, address or fingerprint stays `EXIT_USAGE`: it never leaves the machine, and it is the command
line being wrong, which is what that code has meant since #129.

## The pairing code is a secret

It joins the credential under the "never reaches an output sink" rule: not echoed in a confirmation, not quoted in a
diagnostic, not written into a `--verbose` line. That is why the two shape errors are worded in this package rather than
relayed — the SDK's `bad-code` message quotes what it was handed. `never-logs-a-blob.test.ts` sweeps this verb for the
code as well as for the four credential sentinels.

## Scope

`packages/cli` only, plus its README. `packages/panel` belongs to #286 and is untouched.

- `pnpm -r typecheck` — clean
- `pnpm lint` — 0 errors (46 pre-existing warnings, none in `packages/cli`)
- `packages/cli` suite — 981 passed, 3 skipped, including 26 new

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -26,6 +26,8 @@ release is attested to the workflow and the commit that built it.
 ## Cores this machine can reach
 
 ```sh
+actana core pair laptop core.example:8443 ABCD-2345 \
+  --session <id> --fingerprint AA:BB:…  # enroll this machine on a Core
 actana core add laptop ~/blob.txt   # register a Core from its blob, or stdin
 actana core ls --json               # what this machine knows
 actana core use laptop              # point `current` at one
@@ -33,6 +35,14 @@ actana core status                  # reach it, and report what it says
 actana core shell                   # an interactive shell on that machine
 actana core exec -- df -h /         # one command on it, no terminal
 ```
+
+`core pair` is the enrollment gesture: somebody on the Core runs `actana pair
+new`, reads out an eight-character code and that Core's CA fingerprint, and this
+machine generates its own key pair, checks the fingerprint **before** it sends
+the code, and stores the credential the Core signs. The private half never
+leaves this machine, and the code is never sent to a certificate authority
+nobody confirmed — with no `--fingerprint` and no terminal to confirm one on,
+the command refuses.
 
 `core exec` is the non-interactive half of `core shell`, and the reason it
 exists is that a script cannot use a terminal. It returns the command's real
--- a/packages/cli/src/__tests__/cli-harness.ts
+++ b/packages/cli/src/__tests__/cli-harness.ts
@@ -23,6 +23,9 @@ import type {
   SessionAttachment,
 } from "../session-attach-channel.ts";
 import type { CoreConnectFn, CoreConnectOptions, CoreLinkClient } from "../core-connection.ts";
+import type { CorePairingPort } from "../core-pair.ts";
+import { CorePairingError, type CorePairingFailure } from "@actana/sdk/core-pairing.ts";
+import type { CoreRegistrationBlob } from "@actana/sdk/core-registration-blob.ts";
 import type { OpenSessionGateway, SessionGateway, StartedSession } from "../session-gateway.ts";
 import { projectFilesErrorFrom } from "../project-files-gateway.ts";
 import type {
@@ -80,6 +83,8 @@ export type RunOptions = {
   probe?: CoreProbeFn;
   /** What every other noun gets when it dials, or a throw. */
   connect?: CoreConnectFn;
+  /** What `core pair` gets back from the SDK, or a throw. */
+  pairing?: CorePairingPort;
   /** What the `session` noun's verbs get back, or a throw. */
   sessions?: OpenSessionGateway;
   /** What `project cp` and `project files` get back, or a throw. */
@@ -165,6 +170,83 @@ export function fakeSessionGateway(overrides: Partial<SessionGateway> = {}): Ope
   });
 }
 
+/** What a {@link fakePairing} was asked, and what it answered. */
+export type FakePairing = CorePairingPort & {
+  /** Every `identify`, by the address it was given. */
+  identified: string[];
+  /**
+   * Every `pair`, with the options it carried — **including the code**.
+   *
+   * Recorded so a suite can assert what crossed the seam: that the code handed
+   * to the SDK is the normalised one, that the fingerprint is the confirmed
+   * one, and that a refusal happened without `pair` ever being reached.
+   */
+  paired: Array<Parameters<CorePairingPort["pair"]>[0]>;
+};
+
+/**
+ * The SDK's pairing surface, without a Core.
+ *
+ * `identify` answers with the fingerprint the test says the Core presents;
+ * `pair` hands back a credential or throws the `CorePairingError` the suite is
+ * about. Both are recorded, because half of what this verb has to get right is
+ * *not* reaching the second one.
+ */
+export function fakePairing(
+  opts: {
+    fingerprint?: string;
+    identifyFails?: unknown;
+    blob?: CoreRegistrationBlob;
+    fails?: CorePairingFailure;
+    failsWith?: unknown;
+    detail?: ConstructorParameters<typeof CorePairingError>[2];
+  } = {},
+): FakePairing {
+  const fingerprint = opts.fingerprint ?? PAIRED_FINGERPRINT;
+  const state: FakePairing = {
+    identified: [],
+    paired: [],
+    identify: async ({ address }) => {
+      state.identified.push(address);
+      if (opts.identifyFails) throw opts.identifyFails;
+      return {
+        fingerprint,
+        caCert: SENTINEL_CA,
+        host: address.split(":")[0] ?? address,
+        port: 8443,
+        httpsOrigin: `https://${address}`,
+      };
+    },
+    pair: async (pairOpts) => {
+      state.paired.push(pairOpts);
+      if (opts.failsWith) throw opts.failsWith;
+      if (opts.fails) {
+        throw new CorePairingError(opts.fails, `the fake Core answered ${opts.fails}`, opts.detail ?? {});
+      }
+      return opts.blob ?? sentinelPairedBlob();
+    },
+  };
+  return state;
+}
+
+/** The fingerprint {@link fakePairing} presents unless a test says otherwise. */
+export const PAIRED_FINGERPRINT =
+  "AA:BB:CC:DD:EE:FF:00:11:22:33:44:55:66:77:88:99:AA:BB:CC:DD:EE:FF:00:11:22:33:44:55:66:77:88:99";
+
+/**
+ * The credential a successful pair hands back: the same sentinels every other
+ * suite sweeps for, so `never-logs-a-blob.test.ts` covers this path too.
+ */
+export function sentinelPairedBlob(endpoint = "wss://core.test:8443"): CoreRegistrationBlob {
+  return {
+    endpoint,
+    caCert: SENTINEL_CA,
+    clientCert: SENTINEL_CERT,
+    clientKey: SENTINEL_KEY,
+    bearer: SENTINEL_BEARER,
+  };
+}
+
 /** A probe that answers like a healthy Core on the current protocol. */
 export function healthyProbe(overrides: Partial<CoreProbe> = {}): CoreProbeFn {
   return async () => ({
@@ -233,6 +315,15 @@ export function makeCliFixture(): CliFixture {
           (async () => {
             throw new Error("this test did not expect to dial a Core");
           }),
+        pairing:
+          opts.pairing ?? {
+            identify: async () => {
+              throw new Error("this test did not expect to identify a Core");
+            },
+            pair: async () => {
+              throw new Error("this test did not expect to pair with a Core");
+            },
+          },
         openSessions:
           opts.sessions ??
           (async () => {
--- a/packages/cli/src/__tests__/cli-harness.ts
+++ b/packages/cli/src/__tests__/cli-harness.ts
@@ -223,7 +223,14 @@ export function fakePairing(
       if (opts.fails) {
         throw new CorePairingError(opts.fails, `the fake Core answered ${opts.fails}`, opts.detail ?? {});
       }
-      return opts.blob ?? sentinelPairedBlob();
+      // **The label is echoed, because `pairWithCore` echoes it.** The real
+      // function copies `opts.label` straight into the blob it returns — the
+      // one field where what the caller passed in comes back out — and a fake
+      // that answered with a fixed blob instead would make every assertion
+      // about what is *stored* vacuous, which is exactly how a client hostname
+      // reached the column that means the Core's own alias.
+      const issued = opts.blob ?? sentinelPairedBlob();
+      return { ...issued, ...(pairOpts.label === undefined ? {} : { label: pairOpts.label }) };
     },
   };
   return state;
--- a/packages/cli/src/__tests__/core-pair.test.ts
+++ b/packages/cli/src/__tests__/core-pair.test.ts
@@ -1,0 +1,456 @@
+// `actana core pair` — enrollment from the client's side, end to end through
+// `runActanaCli` (#285).
+//
+// The verb has three jobs and this suite is organised as those three. It has to
+// *get* a credential the way #280 says (a fingerprint confirmed before a code
+// moves, and a code the operator may type in any of the shapes a human types
+// one in); it has to *store* that credential exactly where and how `core add`
+// does, so that every verb downstream is unchanged; and it has to *refuse*
+// legibly — one sentence, one next step and one exit code per failure the SDK
+// distinguishes.
+//
+// The SDK is behind `deps.pairing`, so no test here opens a socket. What is
+// deliberately real is the filesystem: file modes, the cores directory and the
+// `current` pointer are the subject of half these assertions, and a stubbed
+// filesystem would assert nothing about any of them.
+
+import { describe, it, expect, afterEach } from "vitest";
+import { existsSync, statSync } from "node:fs";
+import { coreBlobPath, readCurrentCore } from "../blob-registry.ts";
+import { resolveCore } from "../core-resolution.ts";
+import { CorePairingError, type CorePairingFailure } from "@actana/sdk/core-pairing.ts";
+import {
+  EXIT_FAILURE,
+  EXIT_OK,
+  EXIT_PAIR_CERTIFICATE_INVALID,
+  EXIT_PAIR_CORE_ERROR,
+  EXIT_PAIR_FINGERPRINT_MISMATCH,
+  EXIT_PAIR_FINGERPRINT_UNCONFIRMED,
+  EXIT_PAIR_HOSTNAME_MISMATCH,
+  EXIT_PAIR_MALFORMED_RESPONSE,
+  EXIT_PAIR_NO_CA,
+  EXIT_PAIR_NOT_PAIRABLE,
+  EXIT_PAIR_RATE_LIMITED,
+  EXIT_PAIR_REFUSED,
+  EXIT_PAIR_REJECTED,
+  EXIT_PAIR_UNREACHABLE,
+  EXIT_USAGE,
+} from "../exit-codes.ts";
+import {
+  fakePairing,
+  healthyProbe,
+  makeCliFixture,
+  PAIRED_FINGERPRINT,
+  sentinelPairedBlob,
+  SENTINELS,
+  type CliFixture,
+  type FakePairing,
+} from "./cli-harness.ts";
+import { fakeSystem } from "./machine-fixture.ts";
+
+let fixture: CliFixture | null = null;
+function cli(): CliFixture {
+  fixture ??= makeCliFixture();
+  return fixture;
+}
+afterEach(() => {
+  fixture?.cleanup();
+  fixture = null;
+});
+
+/** The pairing session `actana pair new` printed, as this suite spells it. */
+const SESSION = "5f2a1c0e-0000-4000-8000-0123456789ab";
+
+/** The code, as it was read out loud. */
+const CODE = "ABCD-2345";
+
+/** The three arguments plus the flags a scripted pair carries. */
+function pairArgv(extra: string[] = []): string[] {
+  return [
+    "core",
+    "pair",
+    "prod",
+    "core.test:8443",
+    CODE,
+    "--session",
+    SESSION,
+    "--fingerprint",
+    PAIRED_FINGERPRINT,
+    ...extra,
+  ];
+}
+
+describe("actana core pair — coming by the credential", () => {
+  it("stores what the SDK issued, at 0600, where `core-resolution.ts` finds it", async () => {
+    // The acceptance criterion in full: a successful pair is indistinguishable
+    // downstream from a pasted blob. `resolveCore` is the module every other
+    // noun reaches a Core through, so resolving through it — rather than
+    // re-reading the file this test just watched being written — is what
+    // actually proves "no change" for `ls`, `use`, `rm`, `status`, `shell` and
+    // `exec`.
+    const pairing = fakePairing();
+    const run = await cli().run(pairArgv(), { pairing });
+
+    expect(run.code).toBe(EXIT_OK);
+    expect(run.out.join("\n")).toContain('Paired Core "prod" → wss://core.test:8443');
+    expect(statSync(coreBlobPath(cli().paths, "prod")).mode & 0o777).toBe(0o600);
+
+    const resolved = resolveCore({
+      paths: cli().paths,
+      env: { XDG_CONFIG_HOME: cli().configHome },
+      home: cli().home,
+      coreFlag: null,
+    });
+    expect(resolved.ok).toBe(true);
+    if (!resolved.ok) return;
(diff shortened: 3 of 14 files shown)


qcentic-dev:

- [x] **A `pair` verb on `actana core`, taking a local name, the Core's address and the pairing code, with the code
  accepted hyphenated or not and case-insensitively (the normaliser from task 1, reached through the SDK's public
  surface — the CLI does not import from `@actana/shared`)**
  `actana core pair <name> <address> <code>`, dispatched in `core-command.ts`, implemented in the new `core-pair.ts`.
  The normaliser is the SDK's `parsePairingTicket`, imported from `@actana/sdk/core-pairing.ts`; there is no
  `@actana/shared` import in this path and no copy of the Core's alphabet in this package. `abcd2345`, `AbCd-2345`,
  ` abcd 2345 ` and `ABCD-2345` all reach the SDK as `ABCD-2345` — asserted at the seam. The session id travels either
  as `--session <id>` or joined to the code as `<session>:<XXXX-XXXX>`, because the Core will not go looking for which
  session a code belongs to (#282).

- [x] **The CA fingerprint the operator read off `actana pair new` is supplied by the operator — a flag, an argument, or
  an interactive confirmation of the presented fingerprint — and the code is never sent before it is verified**
  Both shapes: `--fingerprint <sha256>` is the scriptable path; with no flag and a terminal, the command dials
  unverified through `fetchCorePairingIdentity`, prints the fingerprint the Core presents and asks the operator whether
  it matches the one on the Core's terminal. With no flag and no terminal it refuses
  (`EXIT_PAIR_FINGERPRINT_UNCONFIRMED`) — absent is not waived. The confirmed value is handed back to `pairWithCore`,
  which dials again and recomputes it, so what the operator agreed to is enforced on the connection the code actually
  crosses. Three tests assert the code is unsent on each refusal by checking `pair` was never reached.

- [x] **The command wraps the task 4 SDK function and does no crypto of its own; the keypair is born inside the SDK call
  on this machine**
  The only calls are `fetchCorePairingIdentity` and `pairWithCore`, behind a `pairing` port on the deps bag. Nothing in
  `core-pair.ts` imports `node:crypto`, generates a key, builds a CSR or computes a fingerprint.

- [x] **The resulting credential is stored under the cores config dir at mode 0600 through the existing registry, so
  `actana core ls`, `use`, `rm`, `status`, `shell` and `exec` work against it with no change**
  Same `writeCoreBlob` under the same `registryPaths`. A test asserts the mode is `0600` and then resolves the entry
  through `core-resolution.ts` — the module every other noun goes through — rather than re-reading the file it just
  watched being written. A second test drives `ls`, `status`, `use` and `rm` against a paired entry.

- [x] **Name validation, "name already registered" handling and the config-dir resolution match what `actana core add`
  does today**
  Same `coreNameError` check in the same position (`../escape` is refused with the same wording), same replace-in-place
  behaviour on a name already registered — which is what re-pairing after a revocation is — and the same `current`
  pointer rule: the first Core becomes `current`, a second does not steal the pointer.

- [x] **`--verbose` explains the steps and, like today, never prints the credential — nor the pairing code, nor the
  private key**
  Verbose reports the pairing session, the unverified dial, the pinned redemption and the store at `mode 0600`.
  `never-logs-a-blob.test.ts` now sweeps four `core pair` runs — refusal, interactive confirmation, success and a Core
  refusal, all with `--verbose` — for the four credential sentinels **and** for the pairing code. The verb's own suite
  asserts the code appears in no output stream in any of the shapes it is accepted in.

- [x] **Each distinguishable SDK error (wrong/expired/consumed code, attempts exhausted, rate limited, fingerprint
  mismatch, unreachable) gets a one-line operator-facing message that says what to do next, and a distinct exit code
  from `packages/cli/src/exit-codes.ts`**
  Twelve new codes, 10–21, one per SDK failure, switched over exhaustively so a failure added to `CorePairingFailure`
  stops this compiling until it has a number. Each prints the failure and then a one-line next step; a table-driven test
  walks all thirteen failures and asserts the code and that a next step was printed.
  **One deliberate deviation, worth reading before ticking this off:** the three input-shape failures — `bad-address`,
  `bad-code`, `bad-fingerprint` — share `EXIT_USAGE` rather than getting numbers of their own. They never leave the
  machine and they are the command line being wrong, which is what `EXIT_USAGE` has meant since #129; a second answer to
  that question would be the invention, not the economy. Separately, wrong / expired / already-redeemed / attempts-
  exhausted arrive as **one** SDK failure (`refused`) and therefore one exit code, because the Core answers all four
  with one status and one body on purpose (#282) — a CLI that exported four codes there would be inventing three of
  them. The message says all four and points at the Core's audit log, which is where the distinction actually lives.

- [x] **Help text and the `actana core --help` verb list are updated; the help says this runs on the client machine and
  that the code comes from `actana pair new` on the Core**
  `CORE_HELP` gains the verb, its three flags and a "Pairing with a Core" section opening on **you are on the machine
  being paired**; the unknown-verb line now reads `Verbs: pair, add, …`; `actana --help`'s `core` entry says "Pair with
  a Core, register, select and inspect them", beside the existing `pair` entry that already points the other way. Two
  tests assert the help says which machine you are on and names `actana pair new`. `packages/cli/README.md` gained the
  same, since it is published.

- [x] **Tests: a successful pair writes a credential at 0600 that `core-resolution.ts` then resolves; a fingerprint
  mismatch writes nothing; a wrong code writes nothing and exits non-zero; the code and key never appear in verbose
  output**
  All four, plus the interactive decline (nothing written, `pair` never reached), the non-interactive refusal, the
  exit-code table, the normalisation cases, the registry behaviours and the help. 26 new tests; the `packages/cli` suite
  is 981 passed / 3 skipped.

- [x] **The blob-paste path is not removed here — that is task 7, which removes it everywhere at once**
  `core add`, `registration-blob-file.ts`'s decoder and the "Registering a Core" help section are all untouched.
  `registration-blob-file.ts` gained an *encoder* beside the decoder — the credential arrives as an object and the
  registry stores text — which adds to that module rather than replacing anything in it.

## Boundaries

- `packages/panel` is untouched — that is #286.
- `packages/cli/src/exit-codes.ts` is edited here, which this ticket owns for this wave.
- Base is `feat/short-code-pairing`, not `main` and not the train. Nothing is merged and the PR stays a draft.

mrfqcentic (changes requested):
## Review — #285 acceptance criteria, checked one by one

Reviewed `feat/actana-core-pair` against its own base `feat/short-code-pairing` (14 files, +1324/-5). Boundary confirmed: **`packages/panel` is untouched** — `git diff --name-only feat/short-code-pairing...feat/actana-core-pair` returns `packages/cli/**` and `packages/cli/README.md` only.

The design is right and the security argument holds where it counts. One defect blocks, two are worth fixing while you are in the file.

### Acceptance criteria

| # | Criterion | Verdict |
|---|---|---|
| 1 | `pair` verb, name + address + code, hyphen/case-insensitive via SDK | **Met** |
| 2 | Fingerprint supplied by operator; code never sent before it is verified | **Met** |
| 3 | Wraps the SDK; no crypto here; keypair born in the SDK call | **Met** |
| 4 | Credential at 0600 under the cores dir, downstream verbs unchanged | **Partially met** — see finding 1 |
| 5 | Name validation / already-registered / config-dir match `core add` | **Met** |
| 6 | `--verbose` never prints the credential, the code or the key | **Met** |
| 7 | Each SDK failure gets a one-line next step and a distinct exit code | **Met** |
| 8 | Help text and verb list updated; says which machine it runs on | **Met** |
| 9 | Tests: 0600 + resolves, mismatch writes nothing, wrong code writes nothing, no secrets in verbose | **Partially met** — see finding 1 |
| 10 | Blob-paste path not removed here | **Met** |

Proof for the load-bearing ones, from the enforcing lines rather than the description:

- **(2) the code is never sent before the fingerprint is verified.** Two independent gates. `core-pair.ts:249-262` refuses with `EXIT_PAIR_FINGERPRINT_UNCONFIRMED` when there is no `--fingerprint` and no terminal, before `pair` is reached; and `pairWithCore` (`packages/sdk/src/core-pairing.ts:326-344`) throws `fingerprint-unconfirmed`/`fingerprint-mismatch` *before* `generateClientCsr` is called, so nothing exists to send. The tests assert this at the seam — `expect(pairing.paired).toEqual([])` in `core-pair.test.ts:179`, `:207` and `:294` — rather than by reading prose. Confirmed.
- **(1) the normaliser comes through the SDK's public door.** `core-pair.ts` imports only from `@actana/sdk/core-pairing.ts`; `parsePairingTicket` (`sdk/core-pairing.ts:441-449`) strips `[\s-]` and upper-cases. No `@actana/shared` import in this file.
- **(3) no crypto here.** `generateClientCsr` is called inside `pairWithCore` at `sdk/core-pairing.ts:350`, past the comparison; `clientKey` is put back into the returned shape at `:392` and never crossed the wire.
- **(7) exhaustive.** `corePairingOutcome` (`core-pair.ts:344-424`) switches over all fifteen members of `CorePairingFailure` with no `default`, and `core-pair.test.ts:241-275` walks thirteen of them asserting distinct codes and a non-empty next step. A member added to the SDK stops this compiling.
- **(5) matches `core add`.** Same `coreNameError`, same `coreExists` → replace-in-place, same "first Core becomes `current`" rule, same three closing sentences — compare `core-pair.ts:141-197` with `coreAdd` in `core-command.ts`.

---

## Findings

### 1. `packages/cli/src/core-pair.ts:176` — the stored credential gets the **client machine's** hostname as its LABEL, and the test that says otherwise cannot see it

`label: args.label ?? deps.hostname` is never `undefined`. `pairWithCore` copies it straight into the blob it returns:

```ts
// packages/sdk/src/core-pairing.ts:388
...(opts.label === undefined ? {} : { label: opts.label }),
```

and `encodeRegistrationBlobText` (`packages/cli/src/registration-blob-file.ts:135`) only omits the key when the label is the empty string. So the file written at `core-pair.ts:187` carries `label: "laptop-7"`.

Concrete result on a real Core: after `actana core pair prod core.example:8443 … ` on host `laptop-7`, `actana core ls` prints LABEL `laptop-7` for the Core named `prod` — this machine's name in the column that means *the Core's own alias*. That is exactly what the credential's own docblock promises will not happen ("A blob from `actana setup` carries the *Core's* own alias… a paired credential carries no alias at all").

**The test at `packages/cli/src/__tests__/core-pair.test.ts:150` is a false positive.** It asserts `label: ""` in `core ls` — but `fakePairing.pair` returns a fixed `sentinelPairedBlob()` and ignores `opts.label` (`packages/cli/src/__tests__/cli-harness.ts:226`), so the fake cannot exhibit the SDK's passthrough that the real function performs. The test can never fail on this.

Two changes, and both are cheap:
- strip `label` before encoding (or pass the blob to `encodeRegistrationBlobText` without it) so the stored shape carries no alias, matching the docblock;
- have `fakePairing.pair` echo `opts.label` into the blob it returns, so `core-pair.test.ts:150` becomes a real assertion. Right now the one place a paired credential differs from a pasted one downstream is the one place the suite is blind.

This is the only thing I would call blocking.

### 2. `packages/cli/src/core-pair.ts:227` — wrong diagnosis when a `<session>:<code>` code is malformed *and* `--session` was passed

`readTicket` picks between its two messages on `carried && session !== null && session.trim() !== ""`, but `parsePairingTicket` throws `bad-code` for two distinct reasons: disagreeing session ids (`sdk/core-pairing.ts:429`) **and** a bad code shape (`:445`).

```
actana core pair prod core.test:8443 5f2a…:ABCD-234 --session 5f2a…
```

Session ids agree; the code is seven characters. The operator gets *"the code names one pairing session and `--session` names another. Drop `--session`, or pass the bare code beside it — they have to agree."* Dropping `--session` will not help — they are sent after a mistake they did not make. The two cases are distinguishable locally without printing the code: compare `code.slice(0, code.indexOf(":"))` with `args.session` before parsing, and only claim disagreement when they actually disagree. `core-pair.test.ts:348` and `:362` each cover one case in isolation; neither covers the overlap.

### 3. `packages/cli/src/core-pair.ts:318` — `reportPairingFailure` relays `err.message` for **every** `CorePairingError`, `bad-code` included

The module header argues at length that the SDK's `bad-code` sentence must never be relayed because it embeds the code (`… — "ABCD-2345" is not`, `sdk/core-pairing.ts:447`), and `readTicket` words that one case locally for exactly that reason. But the second path that can raise `bad-code` — `deps.pairing.pair()`, which re-runs `parsePairingTicket` internally at `sdk/core-pairing.ts:321` — reaches `line 318` and prints the message verbatim to stderr, a CI log and a shell history. `corePairingOutcome` enumerating `"bad-code"` shows the path is expected to be reachable.

It is not reachable *today*, only because this file hands `pair()` an already-normalised ticket. It becomes reachable the moment the SDK tightens code validation — which its own comment at `sdk/core-pairing.ts:406-412` flags as a deliberate future stop. Special-casing `bad-code` in `reportPairingFailure` the way `readTicket` does costs three lines and removes a latent secret leak from a rule the rest of this PR keeps rigorously.

---

Everything else I checked is sound: the exit-code block's numbering rationale, the `EXIT_USAGE` carve-out for the three shape failures, `never-logs-a-blob.test.ts` sweeping this verb, and the round-trip assertion on `encodeRegistrationBlobText`. One note on that sweep — `never-logs-a-blob.test.ts` only puts the code sentinel through the *bad-code* run; the confirmed, issued and refused runs use the well-formed `ABCD-2345`, so `not.toContain("SENTINEL-VVV")` is vacuous on three of the four. Worth threading the sentinel through all four once the shape check accepts it, but not blocking.

Requesting changes on finding 1.

qcentic-dev:
> In reply to the review at https://github.com/actana/control/pull/298#pullrequestreview-4988238774

Thank you — the blocking finding was real, and so was the reason the suite could not see it. All three findings and the
note on the sweep are fixed. Nothing is deferred.

Head is now `767eedf`. Three commits on top of `b57d8e9`.

---

### 1. The client hostname in the LABEL column — **fixed** (`5ce932f`)

You are right about the mechanism and about the test. `label: args.label ?? deps.hostname` is never `undefined`,
`pairWithCore` copies it into the blob it returns, and the encoder omits the key only for the empty string — so the
stored file carried this machine's name in the column that means the Core's own alias.

Two changes, as you suggested:

- `core-pair.ts` now stores `{ ...blob, label: "" }`. The label is what this machine says about itself to the Core, for
  the operator's `actana pair ls` over there; it stops at the wire. The comment says so at the line that does it.
- `fakePairing.pair` echoes `opts.label` the way the real function does, so the fake exhibits the one passthrough that
  made the difference between a paired and a pasted credential invisible.

I checked the assertion is no longer vacuous rather than assuming it: reverting the strip with the fake fixed turns
**two** tests red — the `core ls --json` row (`label: "the-build-box"`) and the round trip through `core-resolution.ts`
(`label: "vm-1"`).

### 2. Wrong diagnosis when the ids agree and the code is malformed — **fixed** (`cdc8ae4`)

Exactly as described: `bad-code` covers two unrelated mistakes and the branch was choosing on the wrong question.
`readTicket` now reads both spellings of the session id the way `parsePairingTicket` does, and claims disagreement only
when they actually disagree; everything left is a shape. Your example — `…:ABCD-234` beside a matching `--session` — is
now a test, and it asserts the answer does **not** mention `--session`.

### 3. `reportPairingFailure` relaying the SDK's `bad-code` sentence — **fixed** (`cdc8ae4`)

Agreed on the reasoning, including that it is unreachable today and reachable the moment the SDK tightens its shape
check. `bad-code` is now the one failure whose message is written in this package, in a single `badCode` helper both
doors call — so the rule lives in one place rather than in one of two. The test raises a `bad-code` from inside `pair()`
carrying `— "ABCD-2345" is not` and asserts the code appears in no output stream.

### 4. The vacuous sentinel in `never-logs-a-blob.test.ts` — **fixed** (`767eedf`)

You were right that three of the four runs proved nothing. The sentinel is now a code the shape check accepts
(`ZZVV-QQWW`), the malformed one shares its first four characters so one absence check covers both spellings and the
hyphen-less form, and a fifth run covers the second `bad-code` door. The sweep ends by asserting the sentinel code
actually pairs — a sentinel the shape check quietly rejected would put every run back on the refusal path with all
assertions green, which is the failure mode this file exists to not have.

---

### Verification

- `packages/cli` — **983 passed, 3 skipped** (28 in the `core pair` suite, up from 26).
- `pnpm -r typecheck` clean; `eslint packages/cli` clean.
- Each of the three new tests confirmed red without its fix, not merely green with it.
- Commit bodies wrapped at ≤132 characters; the **Conventions** job lints every commit on the branch.

Unchanged: base is still `feat/short-code-pairing`, `packages/panel` is still untouched, nothing merged, and the PR
stays a draft.

mrfqcentic (approved):

Re-read against `feat/short-code-pairing` at `767eedf` (14 files, +1450/−5). Each fix checked from the enforcing line, not the reply.

### Blocking finding 1 — the credential label — **fixed, both halves**

- **The strip.** `packages/cli/src/core-pair.ts:194` now reads `writeCoreBlob(paths, name, encodeRegistrationBlobText({ ...blob, label: "" }))`. `encodeRegistrationBlobText` omits the key on the empty string (`registration-blob-file.ts:135-139`), so the file written carries no `label` at all and `actana core ls` shows an empty LABEL — which is what the encoder's own docblock already promised.
- **The fake.** `packages/cli/src/__tests__/cli-harness.ts:233` now returns `{ ...issued, ...(pairOpts.label === undefined ? {} : { label: pairOpts.label }) }` — structurally identical to `pairWithCore`'s own `...(opts.label === undefined ? {} : { label: opts.label })` at `sdk/core-pairing.ts:388`. That is the passthrough the old fixed blob could not exhibit.

The two together are what make `core-pair.test.ts:150` an assertion instead of a description: it now runs with `--label the-build-box`, the fake echoes it, the verb strips it, and the row asserts `label: ""`. Revert either half and the test goes red — I checked the paths rather than taking the commit message's word for it.

### Finding 2 — malformed code vs disagreeing session — **fixed**

`core-pair.ts:220-236` no longer infers the case from `code.includes(":")`. It reads both spellings the way `parsePairingTicket` does (`separator = code.indexOf(":")`, `carried`/`explicit` both trimmed) and claims disagreement only at `carried !== "" && explicit !== "" && carried !== explicit`. Everything reaching the `catch` afterwards is a shape, so it routes unconditionally to `badCode`. `core-pair.test.ts:362` pins the overlap and asserts the message does **not** mention `--session`.

### Finding 3 — `bad-code` relayed from the failure path — **fixed**

`reportPairingFailure` now short-circuits at `core-pair.ts:352`: `if (err.failure === "bad-code") return badCode(deps);`, before the `deps.err(err.message)` that would have printed the SDK's `— "…" is not`. Both doors onto `bad-code` land in one function (`badCode`, `:246`), which is the right shape — the previous arrangement kept the rule in one of two places. `core-pair.test.ts:387` raises a `bad-code` from inside `pair()` whose message quotes the code and asserts it never appears.

### The sweep note — **fixed, and better than what I asked for**

`never-logs-a-blob.test.ts` now threads a **well-formed** sentinel (`ZZVV-QQWW`) through all runs rather than one the shape check rejects, so the three runs that matter most — the confirmed dial, the issued credential, the Core's refusal — are no longer vacuous. It also checks the hyphen-less spelling and the shared `ZZVV` prefix, adds a fifth run for the SDK-side `bad-code` door, and ends with a guard-on-the-guard asserting the sentinel actually pairs (`:176-181`) so a silently-rejected sentinel cannot put every run back on the refusal path with all assertions green. That last one is the failure mode the file exists to not have, and catching it was not in my review.

### Boundary and CI

- **`packages/panel` untouched — held.** `git diff --name-only feat/short-code-pairing...feat/actana-core-pair` returns only `packages/cli/src/**` and `packages/cli/README.md`.
- CI green on `767eedf`: Typecheck, Lint, Unit Tests, Secret Scan, both installer E2Es, Panel service seam, Conventions, Train rules — all pass. The only non-pass rows are train/publish jobs, correctly skipped for a sub-branch PR.

Everything I flagged is closed, and the two "not blocking" notes were taken as well. **Approving.**

PR #299, opened 2026-08-20
feat(panel): pair a Core by address and short code, with the fingerprint shown first

### LINKED ISSUE #286: Pairing: Panel pairing UI — add-core takes address plus code and shows fingerprint state

qcentic-dev:
## Summary

Replaces the Panel's "paste the pairing token here…" textarea with **address plus code** (#280 task 6). The
add-Core flow is now three steps in order: an address, a fingerprint the operator looks at, and only then
a code.

The pairing call itself runs on the **Panel server** — key generation and a TLS dial are not a browser's
work, and the Panel server is where the mTLS credentials already live (CONTEXT.md "Dumb pipe", ADR 0030).
It is `pairWithCore` from task 4's `packages/sdk/src/core-pairing.ts`, the same function `actana core pair`
(#285) calls, with an address and a registry write around it — no second implementation and no crypto in
the renderer. What comes back is `CoreRegistrationBlob`-shaped, so it goes into the Core registry and the
encrypted store through the same door a decoded blob walks through today; `core-link-manager.ts` dials it
unchanged.

The middle step is the thing only a Panel can offer: a terminal client can only ask an operator to type a
fingerprint in, but a Panel can put the one the machine presented on screen beside the one `actana pair new`
printed. Three states — **not checked**, **verified**, **mismatch** — told apart by colour, by wording, and
by whether the code fields exist at all. A mismatch renders no code box, so there is nothing to click past.

## Related issues

Closes #286

Part of #280. Depends on #284 (already on this base). Sibling of #285, which does the same job for the CLI
through the same SDK function. #287 depends on this landing before it deletes the paste path.

## Type of change

- [ ] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## What is in it

**`packages/panel/src/shared/core-pairing.ts`** — the browser-facing vocabulary: the failure union, the
refusal shape, the fingerprint comparison, and one operator-facing sentence per distinguishable failure.
It restates the SDK's failure union rather than importing it, so a browser bundle never reaches into a
module that opens `node:tls`; the service holds the copy to the original with a compile-time union-equality
assertion, in the one place that can see both.

**`packages/panel/src/server/services/core-pairing.ts`** — the pairing call, and the two things it refuses
to do. Nothing here logs, and nothing forwards an SDK error message: `parsePairingTicket` quotes the string
it was handed back at its caller, so an SDK message can carry the code an operator typed. Every sentence
the browser gets is written from the failure *code*.

**Two routes.** `POST /api/cores/pairing/inspect` carries no code, which is what makes it safe against an
address nobody has verified yet. `POST /api/cores/pairing` is only reachable once the operator has confirmed.

**`registerCoreFromCredential`** — split out of `registerCoreFromRegistrationBlob`, so both doors meet at one
insert, one transaction, one sealing step. That is what makes "indistinguishable downstream" structural
rather than a promise. **The blob-paste path itself is untouched and stays** — #287 removes it everywhere
at once.

## Security properties this holds

- The code is **not sent** until the presented fingerprint matches the expected one — and the browser's
  comparison is a gate, not the enforcement: the server re-runs it inside `pairWithCore` against the
  certificate on the connection it is about to send the code over.
- A mismatch and an unconfirmed fingerprint both leave the pairing session **still redeemable**, which is
  the assertion that proves the code was never sent rather than merely not stored.
- The private key is born on the Panel server and never crosses the wire in either direction; it never
  reaches the renderer at all, because the pairing response is a registry row.
- The code is never persisted, never logged, never in a URL, and never in an error message. A code the Core
  has already refused is cleared from the box.
- Retyping the address drops the fingerprint read off the old one, so a "verified" badge cannot be worn by a
  Core the box is no longer pointed at.

## How was this tested?

`packages/panel/src/server/__tests__/cores-pairing-api.test.ts` (16 tests) drives the API the way a browser
does, against a **real Core**: the Core's own `PtyCoreLinkServer` with the Core's own pairing routes mounted
through the Core's own wiring, behind a real TLS socket, with sessions in the `PairingStore` that
`actana pair new` writes. So "the code was redeemed" means a CSR was signed by that CA, and "the Panel can
dial what it paired" means an mTLS handshake and a bearer both actually worked. It covers a successful pair
producing a registry row plus encrypted secrets that reach `connected`, a mismatch storing nothing, the
failure taxonomy, and a deliberately mistyped code never appearing in a response.

`packages/panel/src/components/views/__tests__/cores-settings-pairing.test.tsx` (30 tests) covers what a
server-side suite cannot see: the three fingerprint states, that a mismatch renders no code box, that a
changed address drops a verified badge, and that all fourteen distinguishable failures render fourteen
different sentences.

Full workspace run — `pnpm typecheck`, `pnpm lint`, `pnpm test` (all 6 stages, 144 panel test files),
`pnpm scan:secrets` — all green.

## Screenshots / recordings

None — the change is exercised by the component suite rather than by hand. Happy to add a recording if a
reviewer wants one.

## Checklist

- [x] This PR targets the **feature branch for #280** (`feat/short-code-pairing`), not `main` and not the train — per #286's Branching section, sub-branches PR into the feature branch
- [ ] I updated documentation (README, docs/, changelog entry) where relevant — CONTEXT.md's "pairing token" / Registration blob vocabulary is #287's to settle, per #286
- [x] Dependent changes have been merged (#284 is on this base), and this PR is a draft

https://claude.ai/code/session_01DYa7TtuZy8Kjcha4bFTgpa

### DIFF
--- a/packages/panel/src/components/views/AddCoreByPairing.tsx
+++ b/packages/panel/src/components/views/AddCoreByPairing.tsx
@@ -1,0 +1,444 @@
+import { useState } from "react";
+import { Btn } from "~/components/ui/Btn";
+import { TextField } from "~/components/ui/TextField";
+import { api, ApiError } from "~/lib/api";
+import {
+  fingerprintCheck,
+  normalizePairingCode,
+  pairingFailureMessage,
+  type CorePairingIdentity,
+  type CorePairingRefusal,
+  type FingerprintCheck,
+} from "~/shared/core-pairing";
+import type { CoreWithDial } from "~/shared/cores";
+
+/**
+ * Adding a Core by short code (#280, #286) — address, then fingerprint, then
+ * code, in that order and never in another.
+ *
+ * The Panel is the one client that can *show* the operator a fingerprint rather
+ * than make them type one into a terminal, so this is a two-step: the first
+ * request carries no code and comes back with the certificate authority the
+ * machine presents, and the second is only reachable once what the operator was
+ * read out and what the machine presented are the same string.
+ *
+ * **The comparison here is a gate, not the enforcement.** The Panel server runs
+ * it again inside `pairWithCore`, against the certificate on the connection it
+ * is about to send the code over, and refuses there too. What this component
+ * owns is that an operator is never *asked* for a code before they have looked
+ * at a fingerprint, and that a mismatch stops rather than warns.
+ *
+ * **Nothing secret is kept.** The code lives in this component's state for as
+ * long as it takes to post it and nowhere else — no preference store, no URL,
+ * no toast, no error message. The credential the pairing returns never comes
+ * back to the browser at all: the server seals it and answers with a registry
+ * row.
+ */
+
+/** What each fingerprint state looks like, and what it is called out loud. */
+const FINGERPRINT_STATES: Record<
+  FingerprintCheck,
+  { label: string; color: string; border: string }
+> = {
+  unchecked: { label: "Not checked", color: "var(--text-dim)", border: "var(--border)" },
+  verified: { label: "Verified", color: "var(--accent-ink)", border: "var(--accent)" },
+  mismatch: { label: "Mismatch", color: "var(--danger, #e5484d)", border: "var(--danger, #e5484d)" },
+};
+
+export function AddCoreByPairing({ onPaired }: { onPaired: (core: CoreWithDial) => Promise<void> }) {
+  const [address, setAddress] = useState("");
+  const [identity, setIdentity] = useState<CorePairingIdentity | null>(null);
+  const [inspecting, setInspecting] = useState(false);
+
+  const [expected, setExpected] = useState("");
+  const [sessionId, setSessionId] = useState("");
+  const [code, setCode] = useState("");
+  const [name, setName] = useState("");
+  const [pairing, setPairing] = useState(false);
+
+  const [refusal, setRefusal] = useState<CorePairingRefusal | null>(null);
+
+  const check = fingerprintCheck(expected, identity?.fingerprint ?? null);
+  const verified = check === "verified";
+  const codeReady = normalizePairingCode(code) !== null && sessionId.trim() !== "";
+
+  /**
+   * Retyping the address drops the fingerprint that was read off the old one.
+   *
+   * Without this, an operator who checked one machine and then pointed the box
+   * at another would be looking at a "verified" badge earned by a Core they are
+   * no longer talking to — the exact confusion the fingerprint exists to
+   * prevent.
+   */
+  const changeAddress = (next: string) => {
+    setAddress(next);
+    setIdentity(null);
+    setRefusal(null);
+  };
+
+  const handleInspect = async () => {
+    if (!address.trim()) return;
+    setInspecting(true);
+    setRefusal(null);
+    setIdentity(null);
+    try {
+      const { identity: next } = await api.inspectCoreForPairing(address.trim());
+      setIdentity(next);
+    } catch (err) {
+      setRefusal(refusalOf(err));
+    } finally {
+      setInspecting(false);
+    }
+  };
+
+  const handlePair = async () => {
+    if (!verified || !codeReady) return;
+    setPairing(true);
+    setRefusal(null);
+    try {
+      const { core } = await api.pairCore({
+        address: address.trim(),
+        code,
+        sessionId: sessionId.trim(),
+        expectedFingerprint: expected,
+        label: name.trim(),
+      });
+      reset();
+      await onPaired(core);
+    } catch (err) {
+      const next = refusalOf(err);
+      setRefusal(next);
+      // A code the Core has already looked at is spent, dead, or was never
+      // right: keeping it in the box invites a second attempt that can only
+      // burn another of the session's five, and keeps a secret in memory that
+      // has no further use.
+      if (next.failure === "refused") setCode("");
+    } finally {
+      setPairing(false);
+    }
+  };
+
+  const reset = () => {
+    setAddress("");
+    setIdentity(null);
+    setExpected("");
+    setSessionId("");
+    setCode("");
+    setName("");
+    setRefusal(null);
+  };
+
+  return (
+    <div
+      style={{
+        padding: "14px 16px",
+        background: "var(--surface-0)",
+        border: "1px solid var(--border)",
+        borderRadius: 7,
+        display: "flex",
+        flexDirection: "column",
+        gap: 14,
+      }}
+    >
+      <div style={{ fontFamily: "var(--mono)", fontSize: 11.5, color: "var(--text-dim)" }}>
+        On the machine, run <code>actana pair new</code>. It prints a one-time pairing code, this
+        Core&apos;s CA fingerprint, and a session id. Enter the address first and compare the
+        fingerprint — the Panel does not send the code until they match.
+      </div>
+
+      {/* Step 1 — reach the machine. Nothing secret is sent by this. */}
+      <div style={{ display: "flex", alignItems: "flex-end", gap: 8 }}>
+        <div style={{ flex: 1, minWidth: 0 }}>
+          <TextField
+            label="Core address"
+            value={address}
+            onChange={changeAddress}
+            placeholder="prod-vm-1.internal:7777"
+            hint="Host and TLS port, as the Core was set up. Pairing needs the TLS port."
+            mono
+            autoComplete="off"
+            spellCheck={false}
+            disabled={inspecting || pairing}
+          />
+        </div>
+        <Btn
+          variant="frame"
+          size="md"
+          onClick={() => void handleInspect()}
+          disabled={inspecting || pairing || !address.trim()}
+        >
+          {inspecting ? "Checking…" : "Check fingerprint"}
+        </Btn>
+      </div>
+
+      {/* Step 2 — the three states, told apart by colour, by wording, and by
+          whether step 3 exists at all. */}
+      <FingerprintPanel
+        check={check}
+        identity={identity}
+        expected={expected}
+        onExpectedChange={(next) => {
+          setExpected(next);
+          if (refusal?.failure === "fingerprint-mismatch") setRefusal(null);
+        }}
+        busy={inspecting || pairing}
+      />
+
+      {/* Step 3 — the code, reachable only past a verified fingerprint. */}
+      {verified && (
+        <div style={{ display: "flex", flexDirection: "column", gap: 10 }}>
+          <TextField
+            label="Session"
+            value={sessionId}
+            onChange={setSessionId}
+            placeholder="the Session line from `actana pair new`"
+            mono
+            autoComplete="off"
+            spellCheck={false}
+            disabled={pairing}
+          />
+          <TextField
+            label="Pairing code"
+            value={code}
+            onChange={(next) => {
+              setCode(next);
+              if (refusal) setRefusal(null);
+            }}
+            placeholder="XXXX-XXXX"
+            hint="Eight characters. Hyphen and case are yours to get wrong."
+            mono
+            autoComplete="off"
+            spellCheck={false}
+            disabled={pairing}
+          />
+          <TextField
+            label="Name in this Panel (optional)"
+            value={name}
+            onChange={setName}
+            placeholder="the machine's host, if you leave this empty"
+            autoComplete="off"
+            disabled={pairing}
+          />
+        </div>
+      )}
+
+      {refusal && refusal.failure !== "fingerprint-mismatch" && (
+        <RefusalBox refusal={refusal} />
+      )}
+
+      <div style={{ display: "flex", justifyContent: "flex-end" }}>
+        <Btn
+          variant="primary"
+          size="sm"
+          icon="plus"
+          onClick={() => void handlePair()}
+          disabled={pairing || !verified || !codeReady}
+        >
+          {pairing ? "Pairing…" : "Pair Core"}
+        </Btn>
+      </div>
+    </div>
+  );
+}
+
+/**
+ * Where the operator stands against the fingerprint they were read out.
+ *
+ * All three states render here rather than one of them being an absence: an
+ * operator who has not checked yet should see that they have not checked yet,
+ * in the same place a verified check would appear, so the badge is read as an
+ * answer rather than as chrome that failed to load.
+ */
+function FingerprintPanel({
+  check,
+  identity,
+  expected,
+  onExpectedChange,
+  busy,
+}: {
+  check: FingerprintCheck;
+  identity: CorePairingIdentity | null;
+  expected: string;
+  onExpectedChange: (next: string) => void;
+  busy: boolean;
+}) {
+  const state = FINGERPRINT_STATES[check];
+  return (
+    <div
+      data-fingerprint-state={check}
+      role={check === "mismatch" ? "alert" : "status"}
+      aria-label={`Fingerprint ${state.label}`}
+      style={{
+        display: "flex",
+        flexDirection: "column",
+        gap: 10,
+        padding: "12px 14px",
+        background: "var(--surface-1)",
+        border: `1px solid ${state.border}`,
+        borderRadius: 7,
+      }}
+    >
+      <div style={{ display: "flex", alignItems: "center", gap: 8 }}>
+        <span
+          style={{
+            fontFamily: "var(--mono)",
+            fontSize: 10,
+            fontWeight: 600,
+            letterSpacing: "0.05em",
+            textTransform: "uppercase",
+            color: state.color,
+            border: `1px solid ${state.border}`,
+            borderRadius: 4,
+            padding: "2px 6px",
+          }}
+        >
+          {state.label}
+        </span>
+        <span style={{ fontSize: 12, color: "var(--text-dim)" }}>
+          {identity === null
+            ? "No machine checked yet — enter an address and check its fingerprint."
+            : check === "unchecked"
+              ? "Now type the CA fingerprint `actana pair new` printed on that machine."
+              : check === "verified"
+                ? "This is the machine you were read a fingerprint for."
+                : "This is NOT the machine you were read a fingerprint for."}
+        </span>
+      </div>
+
+      {identity !== null && (
+        <>
+          <Fingerprint
+            caption={`Presented by ${identity.httpsOrigin}`}
+            value={identity.fingerprint}
+            color={state.color}
+          />
+          <TextField
+            label="CA fingerprint from `actana pair new`"
+            value={expected}
+            onChange={onExpectedChange}
+            placeholder="AA:BB:CC:…"
+            mono
+            autoComplete="off"
+            spellCheck={false}
+            ariaInvalid={check === "mismatch"}
+            disabled={busy}
+          />
+        </>
+      )}
+
+      {check === "mismatch" && (
+        <div
+          style={{
+            display: "flex",
+            flexDirection: "column",
+            gap: 8,
+            padding: "10px 12px",
+            background: "var(--surface-0)",
+            border: `1px solid ${state.border}`,
+            borderRadius: 6,
+          }}
+        >
+          <div
+            style={{
+              fontFamily: "var(--mono)",
+              fontSize: 12,
+              color: state.color,
+              lineHeight: 1.5,
+            }}
+          >
+            {pairingFailureMessage("fingerprint-mismatch")}
+          </div>
+          <Fingerprint caption="Expected" value={expected} color={state.color} />
+          <Fingerprint
+            caption="Presented"
+            value={identity?.fingerprint ?? ""}
+            color={state.color}
+          />
+        </div>
+      )}
+    </div>
+  );
+}
+
+/** One fingerprint, wrapped rather than truncated — every byte is the point. */
+function Fingerprint({
+  caption,
+  value,
+  color,
+}: {
+  caption: string;
+  value: string;
+  color: string;
+}) {
+  return (
+    <div style={{ display: "flex", flexDirection: "column", gap: 3 }}>
+      <div
+        style={{
+          fontFamily: "var(--mono)",
+          fontSize: 10.5,
+          fontWeight: 500,
+          letterSpacing: "0.05em",
+          textTransform: "uppercase",
+          color: "var(--text-dim)",
+        }}
+      >
+        {caption}
+      </div>
+      <div
+        style={{
+          fontFamily: "var(--mono)",
+          fontSize: 11.5,
+          color,
+          wordBreak: "break-all",
+          lineHeight: 1.5,
+        }}
+      >
+        {value}
+      </div>
+    </div>
+  );
+}
+
+/** A refusal that is not a mismatch — one sentence, and what to do next. */
+function RefusalBox({ refusal }: { refusal: CorePairingRefusal }) {
+  return (
+    <div
+      role="alert"
+      data-pairing-failure={refusal.failure}
+      style={{
+        fontFamily: "var(--mono)",
+        fontSize: 12,
+        lineHeight: 1.5,
+        color: "var(--danger, #e5484d)",
+        padding: "8px 12px",
+        background: "var(--surface-0)",
+        border: "1px solid var(--border)",
+        borderRadius: 7,
+      }}
+    >
+      {pairingFailureMessage(refusal.failure, refusal)}
+    </div>
+  );
+}
+
+/**
+ * Read a refusal out of whatever the call threw.
+ *
+ * A `CorePairingRefusal` body is the expected shape; anything else — a proxy
+ * in the way, a Panel that fell over — is reported as `core-error`, which is
+ * the arm whose advice ("nothing was paired, try again") is true of every
+ * failure that never reached the pairing endpoint.
+ */
+function refusalOf(err: unknown): CorePairingRefusal {
+  if (err instanceof ApiError && isRefusal(err.body)) return err.body;
+  return { failure: "core-error", error: pairingFailureMessage("core-error") };
+}
+
+function isRefusal(body: unknown): body is CorePairingRefusal {
+  return (
+    !!body &&
+    typeof body === "object" &&
+    typeof (body as { failure?: unknown }).failure === "string" &&
+    typeof (body as { error?: unknown }).error === "string"
+  );
+}
--- a/packages/panel/src/components/views/AddCoreByPairing.tsx
+++ b/packages/panel/src/components/views/AddCoreByPairing.tsx
@@ -35,6 +35,18 @@ import type { CoreWithDial } from "~/shared/cores";
  * row.
  */
 
+/**
+ * What the panel says when *this* comparison failed.
+ *
+ * Its own string rather than `pairingFailureMessage("fingerprint-mismatch")`,
+ * because this is the one place that can promise the code has not moved: no
+ * request has been made with it, and the button that would make one is not
+ * rendered. The shared sentence has to cover a mismatch the server returned,
+ * where the code may already have been spent, so it cannot promise that.
+ */
+const LOCAL_MISMATCH_MESSAGE =
+  "That machine is not the Core you were read a fingerprint for. The pairing code has not been sent and will not be — find out why the two differ before going any further.";
+
 /** What each fingerprint state looks like, and what it is called out loud. */
 const FINGERPRINT_STATES: Record<
   FingerprintCheck,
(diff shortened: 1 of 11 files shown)


qcentic-dev:
## Acceptance criteria from #286

Every criterion, checked or explicitly unchecked with the reason.

- [x] **The add-Core flow takes a Core address and an `XXXX-XXXX` pairing code instead of a blob paste; the
  code is accepted hyphenated or not and case-insensitively.**
  The textarea is gone from `CoresSettingsPage.tsx`; `AddCoreByPairing.tsx` takes an address, a fingerprint,
  a session id and a code. Tolerance is the SDK's `parsePairingTicket` and is asserted end to end —
  `takes the code hyphenated or not, and in any case` posts `k7rp9x4t` for a code minted as `K7RP-9X4T`
  and gets a 201. The UI's own enable gate (`normalizePairingCode`) is equally tolerant, covered by
  `takes a code typed without its hyphen, in whatever case`.

- [x] **Fingerprint state is shown: the presented CA fingerprint is displayed for the operator to compare
  against what `actana pair new` printed, and the code is not sent until they confirm — verified, mismatched
  and not-yet-checked are three visibly distinct states.**
  `POST /api/cores/pairing/inspect` carries no code and returns the presented fingerprint; the panel renders
  it beside a box for the one the operator was read out. The three states differ in border colour, badge
  text, sentence, and whether the code fields are rendered at all. Covered by the four
  `the three fingerprint states` cases, which assert `data-fingerprint-state` is `unchecked` / `verified` /
  `mismatch` and that the code box exists only in the third.

- [x] **A mismatch is a refusal with the expected and presented fingerprints shown, not a warning the
  operator can click past.**
  On mismatch the panel takes `role="alert"`, prints both fingerprints under `Expected` and `Presented`, and
  the code fields are **not rendered** — there is no gesture that sends a code. The server holds the same
  line independently: `stores nothing, and shows both fingerprints` gets `fingerprint-mismatch` with both
  values, zero rows in `cores` and `core_secrets`, and then proves the code was never sent by redeeming the
  *same* session successfully straight afterwards.

- [x] **The pairing call runs on the Panel server and calls the same task 4 SDK function the CLI does — no
  second implementation, no crypto in the renderer.**
  `server/services/core-pairing.ts` calls `pairWithCore` / `fetchCorePairingIdentity` from
  `@actana/sdk/core-pairing`. The renderer imports nothing from the SDK and contains no crypto; its
  fingerprint comparison is a UX gate over a string, and the server re-runs the real one inside `pairWithCore`
  against the certificate on the connection it is about to send the code over.

- [x] **The returned credential goes into the Core registry and the encrypted secret store through the
  existing path, so a Core added by pairing is indistinguishable downstream from one added by paste today.**
  Both doors now meet at `registerCoreFromCredential` — one insert, one transaction, one `sealSecret`.
  `registers a Core the Panel then reaches over mTLS` pairs against a real Core and then waits for
  `core-link-manager.ts` to reach `connected` over a real mTLS handshake with the issued bearer.

- [x] **The pairing code and the returned private key never reach the renderer's persisted state, a log line,
  or an error message.**
  The renderer holds the code in component state only — no preference store, no URL, no `localStorage` — and
  clears it on success and on a refusal the Core has already seen. The private key never reaches the browser
  at all: the pairing response is a registry row. Nothing in the service logs, and no SDK error message is
  forwarded, because `parsePairingTicket` quotes the string it was handed back at its caller — every sentence
  is written from the failure *code* instead. Asserted from both sides:
  `never lets the code or the issued key back into a response`, `never quotes the code back, even when the
  code is what was wrong`, and `empties the box on a refusal, and never prints the code that was in it`.

- [x] **The distinguishable SDK errors (wrong/expired/consumed code, attempts exhausted, rate limited,
  fingerprint mismatch, unreachable) each render a specific message saying what to do next.**
  Checked, with one thing worth reading before you take it at face value. Fifteen failures each get their own
  sentence naming the next action, and `renders a message of its own for %s` asserts all fourteen that reach
  the refusal box plus a test that the fourteen sentences are fourteen distinct strings.
  **But the first four in that list are one failure, not four.** #282 answers a wrong code, an expired
  session, a consumed one and one out of attempts with one status and one body on purpose — telling them
  apart on the wire would tell an attacker whether a session exists and whether their last guess was closer —
  so the SDK collapses them into `refused` and a client that reported four different things would be
  inventing three of them. `refused` therefore names all four and points at the Core's audit log, which is
  where #282 deliberately put the distinction. Everything the SDK *can* distinguish is distinguished here.

- [x] **Copy on the settings page is updated, including the "Reissue a pairing token on the machine and add
  it again" recovery message, which now points at `actana pair new`.**
  That string is now "Run `actana pair new` on the machine and pair it again." Also updated: the section
  subtitle, the empty-fleet state, the removal dialog's recovery sentence, the add-Core instructions, and the
  `CoreRow` doc comment that described endpoint and credentials as "what the pairing token said they were".

- [x] **Tests: the server-side pairing endpoint against an in-process Core pairing server; a mismatch stores
  nothing; a successful pair produces a registry row and encrypted secrets that `core-link-manager.ts` dials;
  component coverage of the three fingerprint states.**
  All four. `cores-pairing-api.test.ts` (16 tests) runs against the Core's own `PtyCoreLinkServer` with the
  Core's own pairing routes mounted through the Core's own wiring, behind a real TLS socket, with sessions in
  the `PairingStore` `actana pair new` writes. `cores-settings-pairing.test.tsx` (30 tests) covers the three
  states and the refusal behaviour. Full workspace `pnpm typecheck`, `pnpm lint`, `pnpm test` and
  `pnpm scan:secrets` are green.

- [x] **The blob-paste path is not removed here — that is task 7.**
  `POST /api/cores` with `registrationBlob`, `registerCoreFromRegistrationBlob`, `decodeRegistrationBlob`
  and `api.addCore` are all still there and still tested — `cores-api.test.ts` and `cores.test.ts` pass
  unchanged but for one renamed error class. What this PR replaced is the paste *UI*, which #286 names as
  the thing being replaced; the path underneath it is intact for #287 to remove everywhere at once.

### Scope

Everything in this PR is under `packages/panel`, plus the import of #284's SDK pairing module. `packages/cli`
and its `exit-codes.ts` (#285) and `pty-core-link-server.ts` (#289) are untouched — `git diff --name-only`
against the base returns nothing under either.

### One deviation worth a reviewer's eye

`registerCoreFromRegistrationBlob` was split so both doors share one insert. A duplicate endpoint now throws
`CoreRegistryError` rather than `RegistrationBlobError` — same refusal, same 400, same nothing written, but
the name now says whose rule it is, since pairing reaches that invariant through a door that never sees a
blob. One existing assertion in `cores.test.ts` moved with it.

mrfqcentic (changes requested):
## Review — #286 acceptance criteria, checked one by one

Reviewed `feat/panel-pairing-ui` against its own base `feat/short-code-pairing` (11 files, +1900/-94). Boundaries confirmed: **`packages/cli` is untouched and `pty-core-link-server.ts` is untouched** — every path in `git diff --name-only feat/short-code-pairing...feat/panel-pairing-ui` is under `packages/panel/src/`.

The server half is excellent. `registerCoreFromCredential` is the right seam and it makes "indistinguishable downstream" structural rather than a promise; the integration suite drives a real `PtyCoreLinkServer` behind a real TLS socket and waits for the dial to reach `connected`, which is the strongest form this claim could take. The problems are all in the renderer, and they are all in how a refusal gets back to the operator — which is unfortunately the half of this feature that carries the security argument.

### Acceptance criteria

| # | Criterion | Verdict |
|---|---|---|
| 1 | Address + `XXXX-XXXX` code instead of a blob paste; hyphen/case-insensitive | **Met** |
| 2 | Three visibly distinct fingerprint states; code not sent until confirmed | **Met** |
| 3 | A mismatch is a refusal, not a warning to click past | **Partially met** — see finding 1 |
| 4 | Runs on the Panel server, same SDK function as the CLI, no crypto in the renderer | **Met** |
| 5 | Credential goes into the registry + encrypted store through the existing path | **Met** |
| 6 | Code and private key never reach persisted state, a log line, or an error message | **Partially met** — see finding 3 |
| 7 | Each distinguishable SDK error renders a specific message saying what to do next | **Not met** — see findings 2 and 4 |
| 8 | Copy updated, including the recovery message, now pointing at `actana pair new` | **Met** |
| 9 | Tests: server endpoint vs in-process Core, mismatch stores nothing, dial works, three states | **Met** |
| 10 | Blob-paste path not removed here | **Met** |

Proof for the load-bearing ones:

- **(4) one implementation.** `server/services/core-pairing.ts:16` imports `pairWithCore` from `@actana/sdk/core-pairing` — the same function `actana core pair` calls. The renderer imports nothing from the SDK; `shared/core-pairing.ts` restates the failure union and the server pins the copy to the original with `SameUnion<CorePairingFailure, CorePairingFailureCode>` at `server/services/core-pairing.ts:59-61`, which is the one place that can see both.
- **(2) the code is not sent until confirmed.** Two gates, and the second is the real one. In the renderer, step 3 is `{verified && …}` (`AddCoreByPairing.tsx:186`), so a mismatch renders no code box at all. On the server, `pairWithCore` re-runs the comparison against the certificate on the connection it is about to send the code over (`sdk/core-pairing.ts:326-344`), *before* `generateClientCsr` — and an empty `expectedFingerprint` (which `z.string()` would let through) is falsy at `:322`, so it becomes `fingerprint-unconfirmed` rather than a waiver. Absent is not waived. The API suite asserts the pairing session is **still redeemable** after a mismatch (`cores-pairing-api.test.ts:399-442`), which proves the code was never sent rather than merely not stored.
- **(5) indistinguishable downstream.** `registerCoreFromCredential` (`server/services/cores.ts:180`) is the single insert/seal/transaction both doors reach; `cores-pairing-api.test.ts:287-309` pairs against a real Core, then waits for `coreLinkManager` to report `connected` over mTLS with the credential it just sealed.
- **(1) normalisation parity.** `normalizeFingerprint` (`shared/core-pairing.ts:105`) and `normalizePairingCode` (`:118`) are line-for-line equivalent to `parseFingerprint` (`sdk/core-pairing.ts:493`) and the shape check in `parsePairingTicket` (`:442`), so the browser's "verified" and the server's acceptance cannot currently disagree. Note there is no test pinning them together the way the failure union is pinned — worth adding, since a change to the SDK's accepted forms would surface as a green badge followed by a `bad-fingerprint` refusal.

---

## Findings

### 1. `packages/panel/src/components/views/AddCoreByPairing.tsx:225` — a **server-returned** `fingerprint-mismatch` is silently swallowed

```tsx
{refusal && refusal.failure !== "fingerprint-mismatch" && (
  <RefusalBox refusal={refusal} />
)}
```

The suppression is unconditional, but the mismatch UI that is meant to replace it renders off the **local** comparison only — `check === "mismatch"` in `FingerprintPanel`, where `check = fingerprintCheck(expected, identity?.fingerprint)` against the identity captured by `inspect`.

So when the local check says `verified` and the server still returns `fingerprint-mismatch`, the operator sees **nothing at all**: badge still green, no error box, button re-enables. Two ways to get there, and the second is not exotic:

1. The SDK's post-redemption check — `issuedFingerprint !== expected` at `sdk/core-pairing.ts:377-386`. The code has already been spent at this point.
2. The certificate presented on the `pair` dial differs from the one `inspect` saw — a cert rotated between the two calls, or a DNS/LB change moving the address to a different machine. `pairWithCore` refuses at `:335` with the code unsent, and the operator is told nothing.

This is the one refusal in the whole flow that must never be quiet. Gate the suppression on the local state instead — suppress only when `check === "mismatch"`, i.e. only when `FingerprintPanel` is already rendering the two fingerprints side by side.

### 2. `packages/panel/src/components/views/AddCoreByPairing.tsx:433` — `refusalOf` discards the server's message for any 400 that is not a pairing refusal

```ts
if (err instanceof ApiError && isRefusal(err.body)) return err.body;
return { failure: "core-error", error: pairingFailureMessage("core-error") };
```

`isRefusal` requires a `failure` key. The router's generic exposed-error path returns `{ error: message }` and nothing else (`server/http-responses.ts:12`), so every non-`CorePairingRefusedError` 400 falls through to `core-error`.

The case that matters is the already-registered endpoint. `pairCore` deliberately lets `CoreRegistryError` escape its catch (`server/services/core-pairing.ts:138-144`) with the comment *"a registry refusal … is the registry's to explain, and wrapping it as a pairing failure would tell the operator to mint a code they do not need"* — and then the renderer wraps it as a pairing failure and tells the operator to mint a code they do not need:

> "The Core failed while redeeming the code. Nothing was paired — check the Core's logs, then run `actana pair new` and try again."

Every clause is false. The Core did not fail, the code *was* redeemed, and running `actana pair new` again will land on the same duplicate endpoint. The server gets this right — `cores-pairing-api.test.ts:369-387` asserts the 400 body contains "already registered" — so the whole loss is at this seam. `ApiError.message` already carries the server's sentence (`lib/api.ts:91`); fall back to it instead of to a fabricated `core-error`. The component suite has no case for a non-refusal 400, which is why this got through.

### 3. `packages/panel/src/components/views/AddCoreByPairing.tsx:73` — retyping the address drops the fingerprint but **keeps the code**, so a live code minted for one machine can be posted to another

```ts
const changeAddress = (next: string) => {
  setAddress(next);
  setIdentity(null);
  setRefusal(null);
};
```

`code` and `sessionId` are state on `AddCoreByPairing`, not on the `{verified && …}` subtree, so unmounting step 3 does not clear them. Sequence: check machine A, type A's fingerprint, type A's code, then point the box at machine B and verify B's fingerprint — the code and session fields reappear pre-filled with **A's** still-redeemable code, the Pair button enables, and one click posts A's pairing secret to B, which can then redeem it against A.

The PR's security section claims "Retyping the address drops the fingerprint read off the old one, so a 'verified' badge cannot be worn by a Core the box is no longer pointed at." That is true of the badge and not of the secret, and the secret is the half that matters. Clear `code`, `sessionId` and `expected` alongside `identity` — the same reasoning that motivates dropping the fingerprint applies with more force to the code. `cores-settings-pairing.test.tsx:198` covers the badge; nothing covers the code surviving.

### 4. `packages/panel/src/components/views/AddCoreByPairing.tsx:419` — `RefusalBox` recomputes the sentence instead of using the one the server sent

`pairingFailureMessage` is an exhaustive `switch` with no `default`, so it returns `undefined` (while typed `string`) for any `failure` string outside the compiled-in union — an empty red alert box. `refusal.error` is right there in the object, and `shared/core-pairing.ts:78-82` says explicitly that the server fills it in "so that a caller reading only `error` … still gets the sentence rather than a bare code". Use it, with `pairingFailureMessage` as the fallback rather than the primary. Low severity today given the compile-time union assertion, but it costs one line and it is the difference between a degraded message and a blank box.

### 5. `packages/panel/src/server/services/core-pairing.ts:144` — a registry refusal after a successful redemption drops the issued credential on the floor

`registerCoreFromCredential` sits outside the try, by design. But by the time it runs, the code has been spent and the Core has signed and recorded a client certificate for this Panel. When the insert throws, that credential is discarded — never stored, never revoked, and the operator has no path back to it except removing the duplicate row and minting a fresh code. Combined with finding 2 they are not even told which situation they are in.

Not necessarily blocking on its own — re-pairing is a legitimate recovery — but at minimum the operator needs the real reason (finding 2), and it is worth deciding deliberately whether an endpoint collision should be checked *before* the code is spent rather than after.

---

Requesting changes on findings 1, 2 and 3. The server half is in good shape and I would not touch it; all five are in the renderer or the seam it reads.

qcentic-dev:
@mrfqcentic — thank you, this was a sharp review. All three blocking findings are fixed, both low findings
are taken, and one observation is deferred with a reason. Two commits on top of the branch, head is now
`98ac874`.

You were right that the three of them are one problem wearing three hats: the renderer was deciding what a
refusal *means* instead of relaying what the server said.

---

## Fixed

### Finding 1 — a server-returned `fingerprint-mismatch` was swallowed (blocking) ✅

`86a0fa6`. The suppression is now keyed on the local comparison, exactly as you suggested:

```tsx
{refusal && check !== "mismatch" && <RefusalBox refusal={refusal} />}
```

The box is hidden only when `FingerprintPanel` is already rendering the two fingerprints side by side. Both
of your paths — the SDK's post-redemption check at `sdk/core-pairing.ts:377-386`, and a certificate that
changed between the `inspect` dial and the `pair` dial — now land in the box with a green badge above them,
which is the honest picture: the panel knows only what `inspect` reported, and it should not be the thing
that speaks for a refusal it cannot see.

Two consequences fell out of it, and I would like your eye on the second.

**Editing the expected fingerprint no longer clears a refusal.** It used to (`if (refusal?.failure ===
"fingerprint-mismatch") setRefusal(null)`), which was fine while that refusal was always redundant with the
panel. It no longer is: a mismatch the server found is a statement about the certificate that machine
presented, and retyping what you were read out does not change it. The warning now stands until the next
attempt or the next address.

**The shared sentence stopped promising the code was not sent.** It said "The pairing code was NOT sent",
which your path 1 makes false — at `:377` the code has already been spent. Since a client cannot tell your
two paths apart from the outside, `pairingFailureMessage("fingerprint-mismatch")` now says:

> That machine is not the Core you were read a fingerprint for. Stop: do not retry, treat the code as spent,
> and find out why the two differ.

"Treat the code as spent" is the one instruction safe under both. `FingerprintPanel` keeps the stronger
wording as its own string, because it is the one place that can promise it — no request has been made, and
the button that would make one is not rendered. Flagging it because it is a copy change you did not ask for,
made under finding 1 rather than beside it.

### Finding 2 — `refusalOf` discarded the server's message (blocking) ✅

`86a0fa6`. `ApiError.message`, as you said:

```ts
function refusalOf(err: unknown): CorePairingRefusal {
  if (err instanceof ApiError) {
    if (isRefusal(err.body)) return err.body;
    if (err.message.trim() !== "") return { failure: "core-error", error: err.message };
  }
  return { failure: "core-error", error: pairingFailureMessage("core-error") };
}
```

Your reading of the seam was exactly right — the server went out of its way to let `CoreRegistryError` past
`pairCore`'s catch so the registry could explain itself, and the renderer then wrote over the explanation.
`failure` stays `core-error` because there is no pairing failure to name; only the sentence comes from the
server. New test `keeps the server's own sentence for a refusal that is not a pairing one` asserts the
rendered text is the server's and contains no `actana pair new`.

### Finding 3 — retyping the address kept the code (blocking) ✅

`86a0fa6`. `changeAddress` now clears `expected`, `sessionId` and `code` alongside `identity`. Your sequence
is now a test — `drops the code with it — a code minted for one machine is a secret to every other` — which
checks machine A, types A's code, points the box at B, verifies B's fingerprint, and asserts the code and
session boxes come back empty, the Pair button disabled, and neither string anywhere in the document.

You were right to call out the PR description too: it claimed retyping the address drops the fingerprint "so
a 'verified' badge cannot be worn by a Core the box is no longer pointed at". True of the badge, not of the
secret, and the secret is the half that matters.

### Finding 4 — `RefusalBox` recomputed the sentence (low) ✅

`86a0fa6`. `refusal.error` first, `pairingFailureMessage` as the fallback. Test
`renders the sentence the server sent rather than recomputing one` feeds a `failure` outside the compiled-in
union and asserts prose rather than the blank box.

### Finding 5 — a credential dropped after a registry refusal (low) ✅

`98ac874`. Taken the way your last paragraph pointed: the collision is now checked *before* the code is
spent, rather than the credential being rescued after.

`coreRegisteredAt` exposes the question the transaction already asks, and `pairCore` asks it first from the
address it was given. Nothing dials — `parseCoreAddress` is pure — and the endpoint is derived through
`httpsOrigin` rather than by reassembling host and port, which would drop the brackets an IPv6 authority
needs.

**It is advice, not a decision.** A Core reports its own endpoint and that need not be the address it was
reached on, so the check inside the transaction stays the authority and nothing that used to register still
fails. What moves is the ordinary case — pairing a machine already in the fleet — to before the irreversible
step. The refusal now ends "your pairing code has not been used", and the test proves it: after the refusal
it removes the Core that was in the way and redeems **that same code** successfully.

---

## Deferred

**The normalisation-parity test**, from your note under criterion 1 — pinning `normalizeFingerprint` /
`normalizePairingCode` to the SDK's `parseFingerprint` and `parsePairingTicket` the way `SameUnion` pins the
failure union.

I agree with the reasoning and it is cheap. Deferring it only because it is a widening of scope past the
findings, and because the failure mode it guards is now smaller than it was: a divergence surfaces as a green
badge followed by a `bad-fingerprint` or `bad-code` refusal, and after finding 1 that refusal is no longer
suppressed — it renders in the box with its own sentence. Say the word and I will add it in this PR rather
than leaving it for a follow-up; I did not want to expand the diff under a changes-requested review without
asking.

---

## Verification

Six tests added — four on the renderer seams that let findings 1, 2 and 4 through, one on the code surviving
an address change, one on the code surviving a collision. Panel suite is 144 files / 1492 tests.

Full workspace `pnpm typecheck`, `pnpm lint`, `pnpm test` and `pnpm scan:secrets` green locally; **all CI
checks green on `98ac874`** (Conventions, Train rules, Typecheck, Lint, Unit Tests, Secret Scan, Dependency
Audit, both installer E2Es, the Panel service seam E2E, both tarball smokes, both image builds).

Boundaries unchanged: every path is still under `packages/panel/src/`, `packages/cli` and
`pty-core-link-server.ts` untouched, base still `feat/short-code-pairing`, still a draft, nothing merged.
The server half you said you would not touch is unchanged except for `coreRegisteredAt` and the four lines in
`pairCore` that call it.

https://claude.ai/code/session_01DYa7TtuZy8Kjcha4bFTgpa

mrfqcentic (approved):

Re-read against `feat/short-code-pairing` at `98ac874` (11 files, +2156/−94). Two new commits, `86a0fa6` and `98ac874`. Each fix checked from the enforcing line.

### Blocking finding 1 — a server-returned `fingerprint-mismatch` was swallowed — **fixed**

`AddCoreByPairing.tsx:260` is now `{refusal && check !== "mismatch" && <RefusalBox refusal={refusal} />}`. Suppression is keyed on the **local** comparison, so the box is hidden only when `FingerprintPanel` is already showing both fingerprints beside it. A mismatch the server found while the local check reads `verified` — a CA that rotated between `inspect` and `pair`, an address that moved, or the SDK's post-redemption check at `sdk/core-pairing.ts:377-386` — now renders instead of vanishing.

Two consequences you drew that I had not, and both are right:

- `onExpectedChange` is now bare `setExpected` (`:204`). A mismatch the server found is a statement about the certificate that machine presented; retyping the expected value does not change it, so the warning correctly stands until the next attempt or the next address.
- The shared `pairingFailureMessage("fingerprint-mismatch")` no longer promises the code was not sent (`shared/core-pairing.ts:159-165`), because that arm is reached both before the code moves *and* after it is spent, and a client cannot tell those apart. "Treat the code as spent" is the only instruction safe under both. The panel keeps the stronger wording in its own `LOCAL_MISMATCH_MESSAGE` (`:38-47`) — which is the one place that genuinely can promise it, since no request has been made and the button that would make one is not rendered. That distinction was not in my finding and it is a real improvement.

Pinned by `cores-settings-pairing.test.tsx:359` (server mismatch, local check still green, box present, says "treat the code as spent") and `:383` (local mismatch still suppresses the box).

### Blocking finding 2 — non-refusal 400s became a fabricated `core-error` — **fixed**

`refusalOf` (`:481-490`) now has three cases, and case 2 is `if (err.message.trim() !== "") return { failure: "core-error", error: err.message };`. `RefusalBox` renders `refusal.error` first (`:458`), so the registry's own sentence reaches the operator verbatim. The four false claims are gone. `cores-settings-pairing.test.tsx:398` asserts the box text is exactly the server's string and does **not** contain `actana pair new`.

### Blocking finding 3 — retyping the address kept the code — **fixed**

`changeAddress` (`:94-100`) now clears `expected`, `sessionId` and `code` alongside `identity`. The A-to-B disclosure path is closed: nothing about a code survives a change of address. `cores-settings-pairing.test.tsx` covers it directly ("a code minted for one machine is a secret to every other").

### Finding 4 — `RefusalBox` recomputing the sentence — **fixed**

`:458` is `{refusal.error || pairingFailureMessage(refusal.failure, refusal)}`, with the shared message as fallback rather than primary. The test at `:417` feeds a `some-future-failure` code and asserts the server's sentence renders, which is exactly the blank-box case.

### Finding 5 — credential dropped on a registry collision — **fixed, and more strongly than I asked**

I suggested surfacing the real reason. `98ac874` goes further and moves the check *before* the irreversible step: `refuseIfAlreadyRegistered` (`server/services/core-pairing.ts:164-178`) runs at the top of `pairCore`, derives the endpoint through `parseCoreAddress(...).httpsOrigin` — correctly, to keep the brackets an IPv6 authority needs, which a `${host}:${port}` would lose — and refuses with "your pairing code has not been used." Nothing dials, so it costs nothing.

The reasoning around it is careful in the way that matters: `coreRegisteredAt` (`server/services/cores.ts:140-156`) is documented as **advice, not a decision**, and the check inside the transaction stays the authority, because a Core reports its own endpoint and that need not be the address it was reached on. Keeping both is right.

One note, not blocking: the commit says "nothing that used to register still fails," which is very slightly overstated. If a Core previously registered while reporting an endpoint that happens to equal `wss://<the new dial address>`, and a *different* Core now answers at that address reporting something else, the pre-flight refuses where the old code would have registered. That needs two Cores and a reported endpoint crossing over a dial address, it is recoverable by removing the row in the way, and the refusal is honest — worth knowing, not worth changing.

### Boundaries and CI

- **`packages/cli` untouched — held.** **`pty-core-link-server.ts` untouched — held.** All 11 changed paths are under `packages/panel/src/`.
- CI green on `98ac874`: Typecheck, Lint, Unit Tests, Secret Scan, Panel service seam E2E, both installer E2Es, Conventions, Train rules — all pass; only train/publish jobs skipped.

The server half was already excellent and is untouched by these commits, which is the right instinct — every fix landed in the renderer or the seam it reads, except the one that correctly moved a check earlier on the server. **Approving.**

PR #300, opened 2026-08-20
feat(cli): actana session wait, and send --wait built on it

### LINKED ISSUE #289: session send has no way to await the resulting turn

qcentic-dev:
## Summary

The first turn of a Session was awaitable and every turn after it was not.
`actana session send X --wait` was a hard usage error, nothing on the send path
waited, and `session resume --wait` refuses a Session that is still running — so
a caller that sent a follow-up prompt had no supported signal for "that turn is
done".

This implements the design decided in #289. The Core stamps an accepted write in
its event log and answers with the id; the wait resolves on the first settling
status for that task at an event id **strictly greater** than that stamp.
`actana session wait <session>` is the primitive and ships here; `session send
<session> <text> --wait` is send-then-wait built on it, with one PTY resolution
covering the write and the wait. No wait consults `reportsTurnStart`, and
nothing infers a turn from the byte stream.

**Base is `feat/await-a-session-turn`, deliberately** — this stacks on the
ticket branch rather than targeting the open train, and it is a draft.

## Related issues

Closes #289

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## What changed

**Core** — a `write` frame may ask to be stamped; a stamped write the PTY
accepted appends one `session:delivered` carrying the Task id and answers with
its event id. Opt-in, because every keystroke of an attached terminal is a
`write` frame too and this log is append-only for the life of a Core. Nothing is
stamped for a refused write or a PTY with no Task behind it. `task:updated` now
carries the status its mutation *patched*, so a rename of a Session sitting at
`finished` cannot be read as the end of a turn.

**SDK** — `CoreSession.attach` joins a live PTY (the issue's "no constructor for
an existing PTY" landmine): one resolution, listeners wired before the replay,
screen seeded from the Core's ring, status seeded at event id 0. `deliver`
writes and returns the stamp. `waitForTurnEnd({ afterEventId })` resolves only on
a settling status learned *after* the cursor; `waitForIdle` is that with no
cursor and is unchanged for `start`/`resume`.

**CLI** — the `wait` verb (accepts `--wait-timeout` and `--json`, refuses the
rest, `--wait` included), `send --wait`, one shared reporter so all three verbs
print the same object, and help text stating the running-turn limit and the
timeout behaviour.

**Docs** — ADR 0033, and the orchestration skill now documents the full
multi-turn loop.

## Landmines, and where each is pinned

- **`waitForIdle` short-circuits through `settledNow`.** This is why the stamp
  exists. Pinned by *"does not answer a delivered wait with the status the
  Session was already in"* (`core-session.test.ts`) and end to end against a real
  Core in `in-process-core-session.test.ts`.
- **`CoreSession` has no constructor for an existing PTY.** Done as SDK work:
  `CoreSession.attach`.
- **#191's flat quiet-output timer is not coming back.** No timer, no byte-stream
  inference; `no-prompt-timing.test.ts` still passes.
- **`session logs` is a rendered screen, not a transcript.** Not polled, and the
  skill now says out loud not to poll it for a sentinel.
- **`exit-codes.ts` belongs to #285.** Untouched — a timeout is `EXIT_FAILURE`,
  as it already is for `start --wait`.

## How was this tested?

`pnpm test` — all six stages green (one Panel file failed in the full run and
passes in isolation: the known flake, #274). `pnpm typecheck` clean, `eslint`
clean (only warnings that predate this branch).

New coverage: `packages/core/src/__tests__/session-delivered-event.test.ts` (6
cases, real event-log store), the attach/cursor suite in
`packages/sdk/src/__tests__/core-session.test.ts` (including all five settled
statuses, the codex/cursor-cli follow-up turn, the rename case, exit, and a grep
that no wait path names `reportsTurnStart`), the CLI surface suite in
`session-command.test.ts` (including a key-set assertion across `start --wait
--json`, `send --wait --json` and `wait --json`), and two end-to-end cases in
`in-process-core-session.test.ts` that count `findByTask` frames to prove one
resolution.

## Checklist

- [ ] This PR targets the **open train** (`beta/x.y.z`), not `main` — **no, on
      purpose**: it stacks on `feat/await-a-session-turn`, per the ticket's
      sequencing behind #282, #283 and #288
- [x] I updated documentation (ADR 0033, the orchestration skill, help text)

### DIFF
--- a/.agents/skills/actana-sessions/SKILL.md
+++ b/.agents/skills/actana-sessions/SKILL.md
@@ -88,12 +88,60 @@ also look at it directly:
 ```bash
 actana session logs <id>          # the transcript, rendered, while the harness is running
 actana session send <id> "text"   # write into it; --enter follows with a carriage return
+actana session wait <id>          # block until the Core reports it settled
 actana session kill <id>          # stop the harness running for it
 ```
 
 `actana events tail --json` follows the Core's event log as NDJSON if you want
 to watch state change rather than poll — `session:finished` is the Core's own
 signal that a Session reached a terminal state.
+
+## The multi-turn loop
+
+A Session is a conversation, not a single question. The whole loop:
+
+```bash
+ID=$(actana session start <project> "<first prompt>")   # prints the id, exits
+actana session wait "$ID" --json --wait-timeout 900     # block until it settles
+actana session send "$ID" "<follow-up>" --enter --wait --json --wait-timeout 900
+actana session logs "$ID"                               # or read `screen` off the object
+actana session send "$ID" "<next>" --enter --wait --json
+```
+
+**`actana session wait <id>` blocks until the Core reports the Session settled**
+— `finished`, `needs-input`, `interrupted`, `terminated` or `disconnected`,
+because every one of those is a turn that ended. A Session that is already
+settled answers at once, so a wait that starts after the turn ended is not a
+wait that hangs.
+
+**`actana session send <id> "<text>" --wait` waits for the turn that text
+starts.** The Core stamps the delivery in its event log and the wait resolves on
+the first settling status *after* that stamp, so it can never hand you the
+status the Session was already parked at. With `--json` it prints the **same
+object** `start --wait --json` prints — same keys, same `screen` — so one parser
+reads all three verbs. Two of those keys are `null` on a Session you attached to
+rather than started: `command` and `reportsTurnStart` are answers to a spawn.
+
+Three things to know before you build on it:
+
+1. **Sending into a turn that is already running resolves on *that* turn's
+   end.** A keystroke into a busy Harness is not a new turn. If the Session was
+   mid-turn when your text landed, the wait ends when the current turn ends —
+   possibly before the Harness has read a character of what you sent. Wait for
+   the Session to settle *first*, then send.
+2. **`--enter` is what submits the text.** `send` writes exactly the bytes it
+   was given and appends nothing, so text with no return sits in the Harness's
+   composer and no turn starts — and the wait then runs to your timeout.
+3. **A timeout is this side giving up, not a status.** `--wait-timeout
+   <seconds>` bounds the wait; without it there is no deadline, because a turn
+   takes as long as the work takes. On expiry you get a message saying the wait
+   gave up and a non-zero exit — the Session is still running on the Core, and
+   `session logs`, another `session wait` and `session kill` all still work.
+
+Do not poll `actana session logs` for a sentinel instead. It renders a
+fixed-size screen: lines scroll off the top, so counting occurrences is not
+monotonic, and a marker can match your own echoed prompt rather than the
+Harness's reply.
 
 ## Asking a Session for a machine-readable report
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -88,6 +88,7 @@ routing around it.
 | [0030](adr/0030-the-panel-is-a-dumb-pipe-for-file-bytes.md) | The Panel is a dumb pipe for file bytes, and it is the end that holds the credentials |
 | [0031](adr/0031-the-product-ships-one-skill.md) | The product ships one skill, and installs it into the operator's home — **PROPOSED**, amends 0006 |
 | [0032](adr/0032-one-actana-cli.md) | There is one `actana`, and it is both the Core manager and the client — supersedes #265 §6 and 0031 D8's two-binaries premise |
+| [0033](adr/0033-turn-end-is-the-one-mandatory-harness-signal.md) | Turn-end reporting is the one mandatory harness signal; turn-start, auto-mode flags and resume ids may never gate a feature |
 
 **Two files claim 0018**, as the table shows. It is a pre-existing collision,
 not breakage, and **nothing is renumbered** — every citation in the CI files
--- a/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
+++ b/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
@@ -1,0 +1,78 @@
+# Turn-end reporting is the one mandatory harness signal
+
+A harness family is a row in `HOOK_FAMILIES`
+(`packages/core/src/harness-hooks.ts:267`), and the comment above that table
+says the table is open: *"Keep this open: adding a family is adding a row."*
+What the comment does not say is what a row has to **provide** — so the answer
+was being decided one feature at a time, by whichever field the feature reached
+for.
+
+`reportsTurnStart` is the field it kept reaching for, and half the table answers
+`false`. `claude-code` and `opencode` report a turn's start; `codex` and
+`cursor-cli` do not, and cursor's own comment in that table reads *"The turn's
+end is reported; its start is not."* A feature keyed on turn-start works on half
+the harnesses this project ships today and is a coin flip for every family added
+after it.
+
+#289 met this head-on. Awaiting a follow-up turn looks like it needs two
+signals — the turn started, then the turn ended — and the obvious design waits
+for both. Against `codex` and `cursor-cli` the first signal never comes, so the
+obvious design needs a second mechanism to cover the harnesses the first one
+misses, and then a rule for which mechanism a new family gets. That is machinery
+growing to work around a gap. The design that shipped removes the dependency
+instead: the Core stamps a delivery in its own event log, and the wait resolves
+on the first settling status after that stamp. Turn *start* is never consulted.
+
+## Decisions
+
+**D1 — Turn-end reporting is mandatory.** A harness family may be added to
+`HOOK_FAMILIES` when its hooks move the Session's status on the Core at the end
+of a turn — to `finished`, `needs-input`, `interrupted` or `terminated`. That is
+the whole obligation, and it is the one every family in the table already meets.
+
+**D2 — Turn-start reporting is an optional refinement, and may never gate a
+feature.** `reportsTurnStart` survives as reported information: the
+`hooksReportTurnStart` field on the `spawned` frame
+(`packages/core/src/pty-core-link-server.ts`), the `reportsTurnStart` field on
+`session start --json`, and the Panel's terminal-input fallback
+(`packages/panel/src/lib/task-status-sync.ts`). A client may read it to explain
+what it cannot show. **No wait consults it**, and no feature may be built such
+that a family answering `false` loses the feature rather than losing polish.
+
+**D3 — Auto-mode flags and resume ids are optional on the same terms.** OpenCode
+ships no skip-permissions flag (`HARNESS_SKIP_PERMISSION_FLAGS` says so with
+`null`) and a harness that never reports a session id has nothing to resume
+from. Both degrade a Session; neither makes one unobservable. A feature that
+required either would be a feature that quietly excluded a family already in the
+table.
+
+**D4 — A family that reports nothing degrades into the caller's timeout, and
+says so.** No status is invented, and **nothing infers a turn from the byte
+stream**. #191 deleted a flat quiet-output timer that did exactly that and
+[ADR 0026](0026-prompt-delivery-is-a-core-responsibility.md) D3 records why. A
+wait with no `--wait-timeout` has no deadline; one with a deadline reports, on
+expiry, that *this side gave up* — never a status the Core did not send.
+
+## Consequences
+
+Adding a family is still adding a row, and now the row has a bar to clear. A
+harness whose hooks report a turn's end is drivable by every client: the Panel,
+the CLI's `session wait` and `session send --wait`, and any SDK caller.
+
+A harness whose hooks report **nothing** is not excluded from the table — it is
+spawnable, attachable and killable — but it is not awaitable, and callers meet
+that as a timeout rather than as a wrong answer.
+
+## Out of scope: codex's hook review
+
+Codex refuses to run newly-installed project hooks until an operator reviews
+them with its hooks command; the comment sits on the `codex` row itself
+(`packages/core/src/harness-hooks.ts:269-272`). In a fresh workspace codex may
+therefore report nothing at all, turn-end included — so it can fail D1 for
+reasons that have nothing to do with what its hooks say.
+
+**No wait mechanism can close that.** It is an install-time defect, not a
+waiting defect, and it is filed as its own issue:
+[#290](https://github.com/actana/control/issues/290). It is named here so that
+the next reader meeting a silent codex Session knows which of the two problems
+they have.
--- a/packages/cli/src/__tests__/cli-harness.ts
+++ b/packages/cli/src/__tests__/cli-harness.ts
@@ -159,6 +159,8 @@ export function fakeSessionGateway(overrides: Partial<SessionGateway> = {}): Ope
     resume: refuse("resume"),
     logs: refuse("logs"),
     send: refuse("send"),
+    wait: refuse("wait"),
+    sendAndWait: refuse("send --wait"),
     kill: refuse("kill"),
     close: () => {},
     ...overrides,
--- a/packages/cli/src/__tests__/in-process-core-session.test.ts
+++ b/packages/cli/src/__tests__/in-process-core-session.test.ts
@@ -219,6 +219,11 @@ async function coreWithSessions(
     ptyCore: pty.core,
     queryPort,
     mutationPort,
+    // The live-event push, at test speed. A wait resolves when the status event
+    // reaches the client, and the Core's default poll is 500 ms — which is fine
+    // for a person and is most of the wall clock of a suite that waits twice,
+    // on a machine already running five other packages' tests.
+    liveEventPollMs: 25,
     ...(opts.eventLog === false ? {} : { eventLog }),
   });
   fixture = makeCliFixture();
@@ -362,7 +367,10 @@ describe("actana session, against a Core in this process", () => {
     // The transcript rides along, rendered, read while the Session is alive —
     // and it is the Core's replay ring, so it holds what came before the wait.
     expect(String(payload.screen)).toContain("done: 3 files changed");
-  }, 30_000);
+    // A minute, not thirty seconds: this one dials a real Core, runs a CLI
+    // command end to end and waits on a live event push, and `pnpm test` runs it
+    // beside five other packages' suites.
+  }, 60_000);
 
   it("waits as a verb, and refuses to wait on a Session with no harness running", async () => {
     const { eventLog, endTurn } = await coreWithSessions();
@@ -387,7 +395,7 @@ describe("actana session, against a Core in this process", () => {
     const stopped = await fixture!.run(["session", "wait", "task_done", "--json"], withCore());
     expect(stopped.code).toBe(EXIT_FAILURE);
     expect(JSON.parse(stopped.out.join("\n")).error).toContain("no harness running");
-  }, 30_000);
+  }, 60_000);
 
   it("refuses to wait after a delivery the Core did not stamp, rather than answering with the turn before", async () => {
     // The Core here is on this protocol version and still cannot stamp: no
@@ -430,7 +438,7 @@ describe("actana session, against a Core in this process", () => {
       // Read off the archived row rather than lost with it.
       harness: "claude-code",
     });
-  }, 30_000);
+  }, 60_000);
 
   it("says a stopped Session has no transcript rather than printing an empty one", async () => {
     await coreWithSessions();
--- a/packages/cli/src/__tests__/in-process-core-session.test.ts
+++ b/packages/cli/src/__tests__/in-process-core-session.test.ts
@@ -181,7 +181,19 @@ afterEach(() => {
 });
 
 /** A Core holding one live Session and one that has already stopped. */
-async function coreWithSessions(): Promise<{
+async function coreWithSessions(
+  opts: {
+    /**
+     * Wire no event log, which is one of the two ways a Core on this protocol
+     * version still answers a stamped write with no id — the other being an
+     * `appendEvent` that failed. Both are what the gateway's refusal covers,
+     * beyond the version gate that keeps an older Core off the wire entirely.
+     */
+    eventLog?: false;
+    /** Archive a row while its harness keeps running. */
+    archived?: string[];
+  } = {},
+): Promise<{
   writes: string[];
   killed: string[];
   resolutions: string[];
@@ -191,7 +203,7 @@ async function coreWithSessions(): Promise<{
 }> {
   const pty = livePtyCore();
   const tasks = [
-    task(),
+    task({ archived: opts.archived?.includes("task_live") ?? false }),
     task({ taskId: "task_done", title: "ship the changelog", status: "finished" }),
   ];
   // The uncounted lookup: `listSessions` resolves every row's PTY, and counting
@@ -203,7 +215,12 @@ async function coreWithSessions(): Promise<{
   // that id. Without one every write comes back unstamped, which is the
   // older-Core case and not the one these tests are about.
   const eventLog = arrayEventLog();
-  core = await startInProcessCore({ ptyCore: pty.core, queryPort, mutationPort, eventLog });
+  core = await startInProcessCore({
+    ptyCore: pty.core,
+    queryPort,
+    mutationPort,
+    ...(opts.eventLog === false ? {} : { eventLog }),
+  });
   fixture = makeCliFixture();
   await fixture.run(["core", "add", "inproc"], { stdin: core.blobText });
   const endTurn = (taskId: string, status: string): void => {
@@ -370,6 +387,49 @@ describe("actana session, against a Core in this process", () => {
     const stopped = await fixture!.run(["session", "wait", "task_done", "--json"], withCore());
     expect(stopped.code).toBe(EXIT_FAILURE);
     expect(JSON.parse(stopped.out.join("\n")).error).toContain("no harness running");
+  }, 30_000);
+
+  it("refuses to wait after a delivery the Core did not stamp, rather than answering with the turn before", async () => {
+    // The Core here is on this protocol version and still cannot stamp: no
+    // event-log port is wired. Falling through to an uncursored wait would
+    // resolve from the status the Session was already parked at — last turn's
+    // answer, with a zero exit — which is the exact lie the cursor exists to
+    // prevent. It refuses instead.
+    const { writes } = await coreWithSessions({ eventLog: false });
+
+    const run = await fixture!.run(
+      ["session", "send", "task_live", "carry on", "--enter", "--wait", "--json"],
+      withCore(),
+    );
+
+    expect(run.code).toBe(EXIT_FAILURE);
+    const error = String(JSON.parse(run.out.join("\n")).error);
+    expect(error).toContain("did not record the delivery");
+    // And it says the text landed, because the next thing an operator does
+    // with this failure must not be to send it a second time.
+    expect(error).toContain("The text was delivered");
+    expect(writes).toEqual(["carry on", "\r"]);
+  }, 30_000);
+
+  it("waits on a Session that was archived while its harness kept running", async () => {
+    // `tasksList` is active rows only by design (ADR 0019), and every other
+    // verb that names a live PTY works on an archived Session. Refusing here
+    // would make `wait` the odd one out over a row it reads two display fields
+    // off.
+    const { endTurn, eventLog } = await coreWithSessions({ archived: ["task_live"] });
+
+    const waiting = fixture!.run(["session", "wait", "task_live", "--json"], withCore());
+    await waitFor(() => eventLog.tailReads > 0, "the wait subscribed to the event log");
+    endTurn("task_live", "finished");
+
+    const settled = await waiting;
+    expect(settled.code, settled.err.join("\n")).toBe(EXIT_OK);
+    expect(JSON.parse(settled.out.join("\n"))).toMatchObject({
+      taskId: "task_live",
+      status: "finished",
+      // Read off the archived row rather than lost with it.
+      harness: "claude-code",
+    });
   }, 30_000);
 
   it("says a stopped Session has no transcript rather than printing an empty one", async () => {
--- a/packages/cli/src/__tests__/in-process-core-session.test.ts
+++ b/packages/cli/src/__tests__/in-process-core-session.test.ts
@@ -30,10 +30,18 @@ import type {
   CoreLinkSessionSnapshot,
   CoreLinkTaskSnapshot,
 } from "@actana/sdk/core-link-frames.ts";
+import { SESSION_DELIVERED_EVENT_KIND } from "@actana/sdk/core-link-frames.ts";
+
 import { openSessionGateway } from "../session-gateway.ts";
 import { EXIT_FAILURE, EXIT_OK } from "../exit-codes.ts";
 import { makeCliFixture, type CliFixture } from "./cli-harness.ts";
-import { startInProcessCore, type InProcessCore } from "./in-process-core.ts";
+import {
+  arrayEventLog,
+  startInProcessCore,
+  waitFor,
+  type ArrayEventLog,
+  type InProcessCore,
+} from "./in-process-core.ts";
 
 const PROJECT: CoreLinkProjectSnapshot = {
   projectId: "proj_web",
@@ -84,13 +92,23 @@ function livePtyCore(): {
   core: unknown;
   writes: string[];
   killed: string[];
+  resolutions: string[];
+  /** The same lookup, uncounted — for the query ports, which are not a verb. */
+  ptyFor: (taskId: string) => string | null;
 } {
   const writes: string[] = [];
   const killed: string[] = [];
+  // Every `findByTask` this Core is asked, so a verb that resolves the PTY
+  // twice — and opens a window between the two — is visible rather than
+  // arguable (#289: one resolution for the write and the wait).
+  const resolutions: string[] = [];
   const ptys = new Map<string, string>([["task_live", "pty_live"]]);
   const core = {
     setEmitTarget: () => {},
-    findByTask: (taskId: string) => ({ ptyId: ptys.get(taskId) ?? null }),
+    findByTask: (taskId: string) => {
+      resolutions.push(taskId);
+      return { ptyId: ptys.get(taskId) ?? null };
+    },
     taskIdForPty: (ptyId: string) =>
       [...ptys.entries()].find(([, id]) => id === ptyId)?.[0] ?? null,
     replay: (ptyId: string) =>
@@ -116,7 +134,13 @@ function livePtyCore(): {
     killLaunchProcesses: () => ({ ptyCount: 0, ports: [] }),
     killPtysUnderPath: () => {},
   };
-  return { core, writes, killed };
+  return {
+    core,
+    writes,
+    killed,
+    resolutions,
+    ptyFor: (taskId: string) => ptys.get(taskId) ?? null,
+  };
 }
 
 /** Task and project reads, answered from memory. */
@@ -157,19 +181,44 @@ afterEach(() => {
 });
 
 /** A Core holding one live Session and one that has already stopped. */
-async function coreWithSessions(): Promise<{ writes: string[]; killed: string[] }> {
+async function coreWithSessions(): Promise<{
+  writes: string[];
+  killed: string[];
+  resolutions: string[];
+  eventLog: ArrayEventLog;
+  /** What the harness's own Stop hook does on the Core: patch the row, say so. */
+  endTurn: (taskId: string, status: string) => void;
+}> {
   const pty = livePtyCore();
   const tasks = [
     task(),
     task({ taskId: "task_done", title: "ship the changelog", status: "finished" }),
   ];
-  const { queryPort, mutationPort } = ports(tasks, (taskId) =>
-    (pty.core as { findByTask(id: string): { ptyId: string | null } }).findByTask(taskId).ptyId,
-  );
-  core = await startInProcessCore({ ptyCore: pty.core, queryPort, mutationPort });
+  // The uncounted lookup: `listSessions` resolves every row's PTY, and counting
+  // those would drown the one thing `resolutions` is watching for — a *verb*
+  // that resolves the same Session twice.
+  const { queryPort, mutationPort } = ports(tasks, pty.ptyFor);
+  // The Core's event log, wired because #289's wait is a cursor into it: the
+  // Core stamps a delivery there and the client counts settling statuses from
+  // that id. Without one every write comes back unstamped, which is the
+  // older-Core case and not the one these tests are about.
+  const eventLog = arrayEventLog();
+  core = await startInProcessCore({ ptyCore: pty.core, queryPort, mutationPort, eventLog });
   fixture = makeCliFixture();
   await fixture.run(["core", "add", "inproc"], { stdin: core.blobText });
-  return { writes: pty.writes, killed: pty.killed };
+  const endTurn = (taskId: string, status: string): void => {
+    const row = tasks.find((t) => t.taskId === taskId);
+    if (row) row.status = status;
+    // The event the Core's task writer appends, with the status the mutation
+    // patched on it — a report about a turn, and legible as one even when the
+    // status it lands on is the status the row already had.
+    eventLog.appendEvent(
+      "task:updated",
+      JSON.stringify({ taskId, projectId: PROJECT.projectId, status }),
+      { taskId },
+    );
+  };
+  return { writes: pty.writes, killed: pty.killed, resolutions: pty.resolutions, eventLog, endTurn };
 }
 
 /** Every session verb dials the Core for real in this suite. */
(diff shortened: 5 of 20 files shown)


qcentic-dev:
## Acceptance criteria — #289

Every criterion from the issue, in its order, each checked or explicitly
unchecked with the reason.

- [x] **`actana session wait <session> [--wait-timeout <s>] [--json]` exists as a
      verb.** In the `session` verb dispatch (`session-command.ts`), in the
      `SESSION_HELP` usage block, and in the unknown-verb line's verb list. Its
      accepted-flag row is `misusedFlag(args, ["--wait-timeout"])`; `--json` is
      not in `SESSION_FLAGS` and so is permitted for every verb by construction.
      Everything else is refused, `--wait` included — the verb already says it.
      Pinned by *"refuses the flags it does not take, `--wait` included"*.

- [x] **`actana session send <session> <text> --wait` is accepted.**
      `sessionSend`'s row is now `["--enter", "--wait", "--wait-timeout"]` and
      `send X --wait` no longer exits `EXIT_USAGE`. Pinned by *"accepts `send
      --wait`, which was a usage error"*.

- [x] **One PTY resolution for write and wait, with no window between them.**
      `SessionGateway.sendAndWait` → `attached()` → `CoreSession.attach`
      resolves the Task's PTY once, wires the byte stream, the exit and the
      event log **before** writing, then writes through that same PTY and waits
      from the id it was answered with. Proved against a real Core rather than
      argued: the in-process suite counts `findByTask` frames and asserts
      `["task_live"]` — one.

- [x] **The Core appends a delivery event carrying the task id; the wait resolves
      only on a settling status for that task at an event id strictly greater
      than it.** `session:delivered` in `pty-core-link-server.ts`'s `case
      "write"`, task id from `taskIdForPty`, id returned on `writeResult`. The
      SDK's `settledSince(afterEventId)` refuses any status whose event id is not
      strictly greater. **One deviation, stated plainly:** the stamp is *opt-in*
      via a `stamp` field on the frame, not appended for every accepted write.
      Every keystroke of an attached terminal (and of the Panel's terminal pane)
      is a `write` frame, and this event log has no trim — one row per keypress,
      fanned out to every subscriber, would trade the log for a field nobody in
      that path reads. Callers that wait set it; callers that type do not. If a
      reviewer wants it unconditional, it is a one-line change and I will make
      it, but I do not think the log survives it.

- [x] **Resolves on any of the five settled statuses, not on `session:finished`
      alone.** `SETTLED_SESSION_STATUSES` is the set the wait tests, unchanged.
      Pinned by *"resolves on any of the five settled statuses, not on a finish
      alone"*, which runs all five.

- [x] **`send --wait --json` prints the same object shape as `start --wait
      --json`, and a test asserts the two key sets match.** One shared reporter
      (`awaitTurn`) prints for `start`, `resume`, `wait` and `send --wait`.
      *"prints the same object `start --wait --json` prints"* compares sorted key
      sets across all three verbs. `command` and `reportsTurnStart` are `null` on
      an attached Session — same keys, and null is "not asked on this path"
      rather than a plausible value.

- [x] **The running-turn limit is stated in the help text.** `session --help` has
      an "Awaiting a turn" section carrying it in bold, plus the timeout
      behaviour. Asserted by *"states the running-turn limit in the help text"*.
      Also in the ticket's own words in the skill payload.

- [x] **A silent harness resolves into the caller's `--wait-timeout`, reported as
      this side giving up; with no `--wait-timeout` there is no deadline; no path
      infers a turn from PTY bytes or quietness.** The timeout message is the
      SDK's ("was still X after Nms") and is reported as an error, never as a
      status — `status` is absent from the `--json` document on that path.
      `no-prompt-timing.test.ts` still passes; no timer was added anywhere.

- [x] **No new exit code, and `exit-codes.ts` is unchanged.** `git diff
      --name-only origin/feat/await-a-session-turn...HEAD` does not list it. A
      wait timeout is `EXIT_FAILURE`, as it already is for `start --wait`.

- [x] **`reportsTurnStart` is consulted by no wait path, and a `codex` /
      `cursor-cli` Session awaits a follow-up turn in test.** A test greps the
      waiting half of `core-session.ts` (comments stripped, so the prose saying
      *"nothing here consults it"* is allowed to exist) and asserts no
      reference. *"awaits a follow-up turn on a harness that never reports a
      turn's start"* drives a `codex` Session through `finished` → `finished` —
      the value does not change, and the wait still resolves, because the cursor
      counts events rather than value changes.

- [x] **A one-page ADR.** `docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md`,
      indexed in `docs/README.md`. D1 turn-end mandatory; D2 turn-start may never
      gate a feature; D3 auto-mode flags and resume ids optional on the same
      terms; D4 honest degradation. It quotes the `HOOK_FAMILIES` *"adding a
      family is adding a row"* comment and names the codex hook-review defect as
      out of scope, linked to #290.

- [x] **`orchestration-skill-payload.ts` documents the full multi-turn loop.**
      Authored in `.agents/skills/actana-sessions/SKILL.md` and regenerated into
      **both** copies (the drift test enforces byte-identity). New "The multi-turn
      loop" section: start → wait → `send --wait` → read the settled screen →
      send again, with the running-turn limit, the `--enter` trap, the timeout
      behaviour, and a warning not to poll `session logs` for a sentinel.

- [x] **Test: the wait does not resolve early on a Session that was already
      settled when the send arrived.** *"does not answer a delivered wait with
      the status the Session was already in"* — drives the Session to `finished`,
      attaches, delivers, asserts the wait has **not** resolved after the
      pre-existing status, then ends the turn on `finished` again and asserts it
      resolves only then. The same shape end to end against a real Core in
      `in-process-core-session.test.ts`.

- [x] **Test: a wait against a Session with no live PTY fails with a clear
      message rather than hanging.** Three levels: SDK (*"refuses to attach to a
      Session with no harness running, rather than hanging"* →
      `CoreSessionAttachError`), CLI surface (*"says a Session with no harness
      running has nothing to wait on"* → `EXIT_FAILURE`), and end to end
      (`session wait task_done` against the stopped Session in the in-process
      Core).

### Boundaries

`packages/cli/src/exit-codes.ts` — untouched (#285).
`packages/core/src/pty-core-link-server.ts` — touched, as the issue says this
work must: the `case "write"` handler plus one new private recorder beside
`recordSessionLockChange`. Nothing else in that file moved, so the overlap with
#282/#283 is two small, adjacent regions.

mrfqcentic (changes requested):
## Review — #289 acceptance criteria, checked one by one

Reviewed `feat/session-wait` against its own base `feat/await-a-session-turn` (17 files, +1874/-114).

**All three boundaries #289 names hold.** `packages/cli/src/exit-codes.ts` does not appear in the diff — a wait timeout is `EXIT_FAILURE` through the shared `awaitTurn`, exactly as `start --wait` already does, so this does not collide with #285. The wait resolves through a **delivery cursor in the event log**, not `settledNow` alone — `settledSince(afterEventId)` at `core-session.ts:1088` returns null unless `lastStatusEventId > afterEventId`, and `waitForIdle` is `waitForTurnEnd({})` with no cursor, unchanged for `start`/`resume`. And nothing infers a turn from the byte stream: `ingest()` writes the terminal and notifies data listeners and does **not** call `releaseWaiters`; the only non-status resolution path is `ingestExit`, which is a process exit, not quietness.

This is a good design and the shape of it is right. The `session:delivered` stamp is appended synchronously between `core.write()` and the `writeResult` send (`pty-core-link-server.ts:1336-1348`), so no harness event can interleave between the delivery and the id the caller waits from — that is the property the whole ticket turns on and it is correct. Carrying the **patched** status on `task:updated` (`core-task-writer.ts:105-121`) rather than the row's resulting status is the right call and is what makes a `finished → finished` second turn on a hook-only harness legible; the rename case is genuinely excluded because a title-only patch leaves `mutation.status` undefined.

One finding blocks. It is a versioning decision rather than a logic error, and it defeats the ticket's central guarantee on a Core this build will happily connect to.

### Acceptance criteria

| # | Criterion | Verdict |
|---|---|---|
| 1 | `session wait` verb exists; flag row permits `--wait-timeout` and `--json`, rejects the rest | **Met** |
| 2 | `send --wait` accepted; no longer `EXIT_USAGE` | **Met** |
| 3 | One PTY resolution for write and wait, no window between | **Met** |
| 4 | Core appends a delivery event with the task id; wait resolves only strictly after it | **Partially met** — see finding 1 |
| 5 | Resolves on any of the five settled statuses | **Met** |
| 6 | `send --wait --json` prints the same key set as `start --wait --json`; a test asserts it | **Met** |
| 7 | Running-turn limit stated in help text | **Met** |
| 8 | Silent harness → caller's timeout, reported as this side giving up; no deadline without one; no byte-stream inference | **Met** |
| 9 | No new exit code; `exit-codes.ts` unchanged | **Met** |
| 10 | No wait consults `reportsTurnStart`; codex/cursor-cli awaits a follow-up turn in test | **Met** |
| 11 | One-page ADR on turn-end as the mandatory signal | **Met** |
| 12 | `orchestration-skill-payload.ts` documents the multi-turn loop, both copies | **Met** |
| 13 | Test pinning the `settledNow` landmine | **Met** |
| 14 | Test: wait against a Session with no live PTY fails clearly rather than hanging | **Met** |

Proof for the load-bearing ones:

- **(3) one resolution.** `attached()` calls `CoreSession.attach` once (`session-gateway.ts:472`), which performs the single `findByTask` (`core-session.ts:619`); every `deliver` and the wait then go through that Session's `ptyId`. `in-process-core-session.test.ts` counts `findByTask` frames to prove it.
- **(5) all five.** `core-session.test.ts` loops `["finished","needs-input","interrupted","terminated","disconnected"]` and asserts each resolves a cursored wait.
- **(6) one shape.** `startedFields` is the single source and all three verbs reach `awaitTurn`, so the key sets match by construction; `session-command.test.ts:109-144` asserts it across all three and pins `command`/`reportsTurnStart` to `null` on the attach path.
- **(10) not consulted.** `session-command.ts:310` is the only live read and it is inside `reportStartedSession`, which `start`/`resume` reach and `awaitAttachedTurn` does not. The slice-grep test over `waitForIdle`→`// ─── Lifecycle` backs it, with a length guard so a collapsed slice fails loudly.
- **(13) the landmine.** `"does not answer a delivered wait with the status the Session was already in"` drives a real rig to `finished`, attaches, delivers, asserts the wait has **not** moved after 50 ms, then re-patches `finished` and asserts it resolves. That is the exact failure the stamp exists to prevent, pinned properly.

---

## Findings

### 1. `packages/cli/src/session-gateway.ts:482` — blocking — a Core that does not stamp is admitted, and `send --wait` then answers instantly with the **previous** turn's status

`CORE_LINK_PROTOCOL_VERSION` is unchanged at `"0.16.0"` on both sides of this diff, and no capability is announced for the stamp. `coreLinkProtocolCompatible` compares major.minor, so a Core built *before* this PR — also 0.16.0 — passes `core-connection.ts`'s gate and connects normally.

Against that Core the degradation is silent and wrong, not absent:

1. `deliver()` sends `stamp: true`; the old Core's `case "write"` ignores the unknown field and answers `writeResult` with no `deliveryEventId`.
2. `unwrapResponse` reads it as `0` (`core-client.ts:1443`), so `wrote.deliveryEventId === 0`.
3. `afterEventId` stays `0` through both `Math.max` calls at `:493` and `:501`.
4. `wrap()` spreads the cursor conditionally — `...(afterEventId > 0 ? { afterEventId } : {})` at `:648` — so **no cursor is passed at all**.
5. `waitForTurnEnd({})` falls through to `settledSince(0)`, and `CoreSession.attach` has just seeded `lastStatus` from the Task row at `core-session.ts:1024`.

For the ordinary case — a Session parked at `finished` or `needs-input`, which is precisely when a caller sends a follow-up — the seeded status is in `SETTLED_SESSION_STATUSES`, so the wait resolves **before returning to the caller**: last turn's status printed as this turn's, settled `screen` from before the text landed, exit 0. That is verbatim the issue's own landmine — *"Any design that attaches and then waits without a delivery cursor returns instantly and lies."*

Note this is a regression, not a gap: today `actana session send X --wait` is a clean `EXIT_USAGE` refusal. After this PR, against a Core one release behind, it silently returns the wrong answer with a success exit code, which is worse than the hard error it replaces.

The repo's own rule decides this. The `CORE_LINK_PROTOCOL_VERSION` docblock (`core-link-frames.ts:1430-1466`) states the narrow additive exception is available only *"where the capability's absence yields today's behaviour **exactly**"* — and #266 moved 0.15.0 → 0.16.0 for a case with a *weaker* claim than this one, on the reasoning that a verb against a Core that cannot serve it *"is not a withheld affordance; it is the command not working"*. Here absence does not yield today's behaviour, exactly or otherwise.

Three ways out, any of which is fine:
- move the minor to 0.17.0, per the #266 precedent — probably the cleanest, since `core status` and the Panel's `needs-update` then tell the operator in one sentence before the first `wait`;
- announce a capability on `ready` and have `attached()` refuse before writing;
- at minimum, refuse in `attached()` when `deliver !== null && wrote.ok && wrote.deliveryEventId === 0` — never fall back to an uncursored wait after a delivery. The same guard covers the other two ways to get a 0 that the code already documents: a Core with no `eventLog` port, and an `appendEvent` that failed (`recordSessionDelivery` returns 0 for both).

The third is worth doing regardless of the version decision, because it is the only one that also covers those two.

### 2. `packages/sdk/src/core-session.ts:646` — medium — `attach({ replay: false })` never subscribes to the PTY, so the Session is deaf for its whole life

```ts
if (opts.replay !== false) {
  await client.ptySubscribe(ptyId, { catchUp: true });
  const { data } = await client.replay(ptyId);
  terminal.write(data);
}
```

The subscribe is inside the replay guard. `CoreClient.ptySubscribe`'s contract is explicit (`core-client.ts:1179-1181`): *"Until a client subscribes, `onData`/`onExit` never fire for that `ptyId` — a Core fans output out to the connections that asked and to no others."*

So on a `multiConnection` Core, `attach({ taskId, replay: false })` returns a Session whose `screen()` stays permanently empty and — the part that bites — whose `onExit` never fires, so `ingestExit` never runs and the "an exit answers every wait, cursor or none" route in `settledSince` is dead. A caller that attaches with `replay: false` to await a harness that then dies hangs until its deadline instead of resolving.

The option's own docblock says off is *"for a caller that wants only what this attachment saw"*, which is not what it does — it makes the attachment see nothing. The CLI always takes the default so nothing ships broken today, but `attach` is public SDK surface and this is a documented option that does not work. Hoist the subscribe out of the `if` and pass `catchUp: opts.replay !== false`.

### 3. `packages/cli/src/session-gateway.ts:466` — low/medium — `wait` and `send --wait` refuse an archived Session whose harness is still running, while every sibling verb accepts it

`attached()` opens with `await this.findTask(taskId)`, and `findTask` (`:561`) resolves through `tasksList()`. That frame is active-rows-only by design — `CoreQueryPort.listTasks` (`pty-core-link-server.ts:113-118`): *"The Core omits archived tasks… There is no argument that changes this; the Archived view has its own frame."*

So for a Session archived while its harness is still alive, `actana session wait <id>` and `session send <id> … --wait` fail with `this Core has no session <id>`, while plain `session send`, `session logs`, `session kill` and `session attach` all work on that same id because they resolve through `findByTask`/`livePty`. `findTask` was previously reached only by `resume` (`:340`), where refusing an archived row is defensible; here it is gratuitous — the call is needed only for `projectId` and `agent` on the output object, both of which could be left null the way `command` and `reportsTurnStart` already are on this path.

### 4. `packages/sdk/src/core-session.ts:656` — low — a transient RPC failure during attach is reported as "there is nothing to attach a wait to"

The `catch` wraps *every* failure in the replay/seed block as `CoreSessionAttachError`, and `session-gateway.ts:478` maps that kind unconditionally to `SessionGatewayError("not-running", …)`. A link blip or a busy Core during `ptySubscribe`, `replay` or `sessionsList` therefore tells the operator the harness has exited and points them at `resume`, when a retry is what they want. Only the `findByTask`-returned-null case at `:620` is genuinely `not-running`; the rest should keep `refused`.

### 5. `packages/sdk/src/core-session.ts:882` — low — `dispose()` never undoes the `ptySubscribe` that `attach` issued

No `client.ptyUnsubscribe(this.ptyId)` on the way out. The CLI is masked because the process closes the socket immediately, but `attach` is documented as "join a Session… for the length of a turn" — an orchestrator that attaches to and disposes N Sessions on one long-lived `CoreClient` leaves N PTY streams flowing over the wire to listeners that have been removed, for the life of that client.

---

### Also checked and sound

- Cursor monotonicity: `noteStatus` only ever moves `lastStatusEventId` forward, and both `readStatus` and `seedStatus` pass 0, so a read-back can never satisfy a cursor on its own — the comment at `:970-976` is accurate about why that matters.
- A repeated status still moves the cursor while `onStatus` stays gated on a value change (`:1040-1058`), which is what a second turn on a hook-only harness depends on.
- `patchedStatusOf` is only reached for the three task-scoped kinds in `STATUS_BEARING_EVENT_KINDS`, past the `event.taskId !== this.taskId` guard, so no other event's `status` field can be misread.
- `unwrapResponse`'s `writeResult` now returning a pair has no other consumers — only `rpc` reaches it, and `CoreClient.write` was updated to destructure.
- ADR 0033 covers what #289 E asks for, cites the `HOOK_FAMILIES` "adding a family is adding a row" comment, and names the codex hook-review defect as out of scope with a live link to #290.

Requesting changes on finding 1.

qcentic-dev:
> Requesting changes on finding 1.

Fixed, both halves, plus findings 2, 3, 4 and 5. Four commits on top of the
branch — `18a6175`, `e0647da`, `55ac357`, `dc6d5b7`. Thank you for the trace
through `deliver` → `Math.max` → the conditional spread → `settledSince(0)`;
that is exactly the path, and reading it back is what made the second half of
the fix obvious.

### 1 — blocking — an unstamping Core is admitted, and the wait then lies

**Fixed, both ways you named**, because they cover different Cores.

**The version moved: `CORE_LINK_PROTOCOL_VERSION` is now `0.17.0`.** You are
right that the additive exception is not available here — absence does not yield
today's behaviour, it yields something *worse* than today, since a clean
`EXIT_USAGE` refusal becomes a wrong answer with a zero exit. The docblock now
carries that reasoning next to the #266 paragraph it follows, and the frame
suites assert `0.17.0` with the reason in the test name. `coreLinkProtocolCompatible("0.16.0")`
is now `false` and there is a test saying why in one sentence.

**And the refusal is in, because the version gate does not cover everything.** As
you say, a Core *on* this version still answers 0 when no `eventLog` port is
wired or when `appendEvent` failed. `attached()` now refuses whenever a delivery
was written, accepted, and came back unstamped, and never falls back to an
uncursored wait after a delivery. One nuance I added: it is guarded on a write
having actually happened, so a degenerate deliver-nothing call sits where a bare
`wait` does rather than being refused for a cursor it never needed.

The message names the thing an operator needs most:

```
session task_live took the text, but this Core did not record the delivery in its event log
— so there is no cursor to await this turn from, and waiting would report the turn before it.
The text was delivered; `actana session logs task_live` shows it
```

The first instinct on a failed `send --wait` is to send it again, and this one
went through.

Pinned end to end in `in-process-core-session.test.ts` against a real Core wired
with **no event log** — the case the version gate cannot catch — asserting
`EXIT_FAILURE`, the message, and that both writes did land on the PTY.

### 2 — medium — `attach({ replay: false })` never subscribes

**Fixed.** You are right that the option's docblock promised something it did
not do, and the exit route mattering more than the screen is the part I had not
seen. The `ptySubscribe` is out of the guard and unconditional; `catchUp` is now
`opts.replay !== false`, so the Core is only asked to hold the stream when a
replay is actually coming to redeem it. The docblock says out loud that the flag
does not turn the subscription off.

Test: `"subscribes to the PTY even with the replay off, so an exit still settles
a wait"` — attaches with `replay: false` against a multi-connection Core,
asserts the screen is empty (that is all the flag should buy), delivers, then
emits an exit and asserts the cursored wait resolves on it.

### 3 — low/medium — `wait` refuses an archived-but-running Session

**Fixed.** The archived list is a fallback, asked only when `tasksList` did not
have the row, so the ordinary path still costs one round trip. `resume` keeps
`findTask`, where a missing row genuinely ends it.

I took your suggestion about the display fields one step further: a row this
verb cannot find *at all* is no longer fatal either, since it is read only for
`projectId` and `agent`. `harness` joins `command` and `reportsTurnStart` in
being `null` on the attach path rather than failing a wait over a label.

Test: a Session archived while its harness keeps running is waited on
successfully, with `harness: "claude-code"` read off the archived row rather than
lost with it.

### 4 — low — a transient RPC failure reported as "nothing to attach to"

**Fixed.** Only the `findByTask`-returned-null case is `CoreSessionAttachError`
now, and it is decided before the subscribe/replay/seed block runs. Everything in
that block rejects with an ordinary `Error`, which the gateway already maps to
`refused`. The docblock says "**Only that**" and gives your reason: the two are
different next steps, and sending an operator to `resume` for a Session that is
running perfectly well is the expensive version of being wrong.

Test asserts the rejection is *not* a `CoreSessionAttachError` — and I checked
the assertion fails when inverted, so it is not a silently-passing `.not`.

### 5 — low — `dispose()` never undoes the `ptySubscribe`

**Fixed**, though you did not block on it. `dispose` sends `ptyUnsubscribe` when
*this* attachment took the subscription, fire-and-forget with a catch, since
`dispose` is synchronous and a link that is already gone has released it by
going. A spawned Session sends nothing: the Core subscribes the spawning
connection inside the spawn, and nothing here asked for that — there is a test
for each direction, including that a second `dispose` is not a second frame.

### Nothing deferred

All five are in. Two things I deliberately did **not** touch:

- `docs/reference/cli-parity-check.md` names `0.16.0` twice. It is a
  point-in-time record of what a past PR's sweep found — it already quotes a
  stale line number for the same constant — so editing it would rewrite a
  finding rather than update a fact.
- `CHANGELOG.md` — it is built from PR titles at release time, so the protocol
  move will be described by this PR's title rather than by a hand-written entry.

One test of mine went red in a full local `pnpm test` on a loaded machine — my
own end-to-end wait, timing out at 30 s while five other packages' suites ran
beside it, on a tree that was correct. `dc6d5b7` polls live events at 25 ms (what
the other Core-dialling suites already do) and gives the three tests that dial a
Core a minute; the file now runs in a third less wall clock than before. Calling
it out rather than letting it look like an unexplained flake later.

Still a draft, base unchanged, nothing merged, no exit code added.

qcentic-dev:
Two more commits after that reply, both from re-reading my own diff rather than
from anything you raised — flagging them so the commit list matches the account
above:

- **`eb52880`** — the flag `dispose` reads to decide whether it owes a
  `ptyUnsubscribe` was set *after* the replay instead of after the subscribe, so
  a `replay` that threw left the connection subscribed with nothing that would
  ever give it back. That is the leak finding 5 is about, reintroduced in the
  one path where it is least visible. The assignment moved two lines up, next to
  the frame that creates the debt.
- **`2a586f9`** — the new `subscribedToPty` field landed between `lastStatus`
  and the docblock explaining it, so two doc comments stacked and the paragraph
  about why `lastStatus` starts null — the distinction the cursored wait rests
  on — was documenting the wrong field. Field and comment moved above it.

Head is `2a586f9`. Six commits on top of the branch you reviewed.

mrfqcentic (approved):

Re-read against `feat/await-a-session-turn` at `9a40db7` (20 files, +2162/−130). Six new commits. Each fix checked from the enforcing line.

### Blocking finding 1 — a non-stamping Core was admitted — **fixed, both halves**

You took both routes I offered rather than either, which is the right call because they cover different failures.

**The gate.** `CORE_LINK_PROTOCOL_VERSION` is `"0.17.0"` at `sdk/core-link-frames.ts:1485`. `coreLinkProtocolCompatible` compares major.minor, so a Core built before this PR — 0.16.0 — is now refused by `core-connection.ts` before a frame goes out. The reasoning is written into the version docblock (`:1463-1479`) against the same rule I cited, and reaches the same conclusion for the same reason #266 did: the additive exception is unavailable because absence yields something *worse* than today's behaviour, not equal to it. `core-link-exec-frames.test.ts:121-127` asserts `coreLinkProtocolCompatible("0.16.0") === false` with a comment naming exactly this, and the bump is propagated through the three suites that assert on it.

**The refusal.** `session-gateway.ts:535` now reads:

```ts
if ((deliver.text.length > 0 || deliver.enter) && afterEventId === 0) {
  throw new SessionGatewayError("refused", `session ${taskId} took the text, but this Core did not record the delivery …`);
}
```

so the code never falls back to an uncursored wait after a delivery. This is the half the version gate cannot cover — a Core *on* 0.17.0 that still answers 0 because no event-log port is wired or an `appendEvent` failed, both documented on `recordSessionDelivery`. The guard on `deliver.text.length > 0 || deliver.enter` is right: a delivery of nothing is not a delivery and correctly leaves the caller where a bare `wait` sits.

The message is the detail I want to single out. It says the text **was** delivered and points at `session logs`, because the next thing an operator does with this failure must not be to send it a second time. `in-process-core-session.test.ts:400` drives a real Core with `eventLog: false` and asserts both the refusal and `writes === ["carry on", "\r"]` — so the claim in the message is proven true rather than asserted.

### Finding 2 — `attach({ replay: false })` never subscribed — **fixed**

`core-session.ts:673` hoists the subscribe out of the guard: `await client.ptySubscribe(ptyId, { catchUp: replay })`, unconditional, with `catchUp` set exactly when a replay follows — which also keeps the promise that a `catchUp` hold is always redeemed. `onData`/`onExit` now fire either way, so the "an exit answers every wait" route in `settledSince` is alive with the replay off. The option's docblock (`:254-258`) now says the subscription is a separate thing and is unconditional. Pinned by "subscribes to the PTY even with the replay off, so an exit still settles a wait".

### Finding 3 — archived-but-running Sessions refused — **fixed**

`attached()` now resolves through `findAnyTask` (`session-gateway.ts:598-614`), which asks `archivedTasksList` only when `tasksList` did not have the row — so the ordinary path still costs one round trip, and `wait` stops being the odd verb out against a Session every other live-PTY verb accepts. A row found in neither is no longer fatal: it is read for two display fields, and `harness` joins `command` and `reportsTurnStart` in answering `null` (`:158-165`) rather than failing a wait over a label. `noTurnStartLine` was widened to `string | null` and handles it (`session-command.ts`), and `resume` correctly keeps the strict `findTask`, where a missing row genuinely ends the road. Pinned end-to-end by "waits on a Session that was archived while its harness kept running", which also asserts `harness: "claude-code"` is read off the archived row rather than lost with it.

### Finding 4 — a link blip reported as a dead harness — **fixed**

`core-session.ts:687-696` no longer wraps everything as `CoreSessionAttachError`; only the `findByTask`-returned-null case throws that kind, decided before any of the fallible work runs. Everything else becomes an ordinary `Error` and reaches the operator through the gateway's `refused` arm. Pinned by "reports a link that blinked mid-attach as a failed attach, not as a dead harness".

### Finding 5 — `dispose()` never gave the subscription back — **fixed**

`dispose` now calls `ptyUnsubscribe` behind a `subscribedToPty` flag (`core-session.ts:918-928`), fire-and-forget with a swallowed rejection, which is correct for a synchronous method. Two tests cover both directions, including that a **spawned** Session does not unsubscribe a stream it never asked for — the Core subscribes the spawning connection inside the spawn.

Worth noting `eb52880` separately: a self-review caught that the flag was being set *after* the replay, so a throwing replay would leak the very subscription the flag was added to reclaim. It moves two lines up, next to the frame that creates the debt. That is a real bug in the fix, found and closed without being reported.

### Boundaries — all three still hold

- **No exit code added.** `packages/cli/src/exit-codes.ts` still absent from the diff; the new refusal reuses `SessionGatewayError("refused")` and a timeout is still `EXIT_FAILURE`. No collision with #285.
- **Waits still resolve through the delivery cursor.** `core-session.ts:1134` — `if (afterEventId > 0 && this.lastStatusEventId <= afterEventId) return null;` — unchanged, and the zero-guard makes it strictly harder to reach an uncursored wait than before.
- **No byte-stream inference.** `ingest()` writes the terminal and notifies data listeners and still does not touch `releaseWaiters`. The only `setTimeout`s in the file are the caller's `--wait-timeout` and the pre-existing status-read retry; neither infers a turn from output quietness.

### CI

Green on `9a40db7`: Typecheck, Lint, Unit Tests, Secret Scan, Panel service seam E2E, both installer E2Es, both tarball smokes, Conventions, Train rules — all pass. Only train/publish jobs skipped, as expected for a stacked draft. The `dc6d5b7` test change (pushing live events at test speed) removed the timing race rather than papering over it with a longer wait, which is the right way to fix that class of flake.

All five findings closed, the two most important with a mechanism rather than a patch. **Approving.**

PR #302, opened 2026-08-21
refactor: remove the hand-carry — delete blob emission, blob paste and their docs (#287)

### LINKED ISSUE #287: Pairing: remove the hand-carry — delete blob emission, blob paste and their docs

qcentic-dev:
Closes #287. Base is `feat/short-code-pairing` — the feature branch for #280, not `beta/0.4.0` and not `main`.

Tasks 1–6 of the pairing train made a client able to enroll with a short code. This one deletes what that replaced: blob emission, blob paste, and every page that told an operator to copy one. **There is no deprecation, no `--legacy-blob`, and no dual path** — that is #280's decision, not this PR's to soften.

## What went

| | |
| --- | --- |
| `actana setup` | The `blob` field on `SetupResult`, and the "paste this into your Panel" block. It now says to run `actana pair new`. |
| `actana token` | The bare verb refused, naming the code flow. `token regenerate` survives as this Core's identity rotation and now hands nothing out. |
| `actana core add` | The verb, its file/`-`/stdin handling, its help, and its entry in the verb list. |
| Core first run | `registration-blob.txt`, `formatRegistrationBlobNotice`, `buildRegistrationBlob`, `registrationBlobPath`, and `LoadOrMintResult.blob`. |
| `core-entry` | The `@@AC_CORE_REGISTRATION_BLOB@@` sentinel and everything printed behind it. |
| Panel | `POST /api/cores`, its controller, `api.addCore`, `registerCoreFromRegistrationBlob` and `RegistrationBlobError`. |

The Panel's paste box itself went with #286's `AddCoreByPairing`; what this removes is everything that was still standing behind it. `registerCoreFromCredential` is the one door left, and it is the same function pairing has gone through since #286 — a Core registered today is byte-for-byte the registry row it was, minus a second way in that nobody was exercising.

## Three judgement calls, stated plainly

**1. `packages/shared/src/registration-blob.ts` stays, and the criterion says to say what still imports it.**

`encodeRegistrationBlob` is imported by `packages/shared/src/local-core-wiring.ts`, which writes a machine's own Core into `$XDG_CONFIG_HOME/actana/cores/<name>.txt` — on metal from `actana setup`, in a container from the daemon's own boot (#288 D9). `decodeRegistrationBlob` is its inverse and the only reader `packages/core` and `packages/shared` have for that file. That is a storage format, not an artifact anybody carries, and the module header now says so at length instead of describing itself as the thing `actana setup` prints.

One thing did move: `wireLocalCore` takes the credential *object* now and encodes it itself. While `actana setup` called `encodeRegistrationBlob` directly, "the artifact setup emits" and "the bytes the registry stores" were the same expression in the same function, and deleting the first meant reasoning about the second. They are two things in two places now, and `actana-setup.ts` names no codec at all — which is what criterion 1 asks for.

**2. `packages/cli/src/registration-blob-file.ts` stays too, for the same reason and against the letter of criterion 3.**

The criterion names it alongside the `add` verb, but the ticket's own file pointers say `blob-registry.ts` storage stays and only the paste entry point goes — and this module *is* that storage's codec. `blob-registry.ts` and `core-resolution.ts` both decode with it to read a registry entry, and `core-pair.ts` encodes with it to write one. Deleting it would delete `actana core ls`, `use`, `status`, `shell`, `exec` and `core pair`'s ability to store what it was issued. What went instead is the paste that fed it: the header, the JSDoc and the failure messages no longer describe an operator's clipboard, and its test says what it now covers.

**3. A remote Core requires `AC_CORE_MATERIAL_FILE`, which is one more removal than the criteria list.**

`core-entry.ts` had a second identity branch for a daemon started in remote mode without a material file: it minted certs it never persisted and printed a blob, and that blob was the entire means by which anything could reach that Core. With the blob gone the branch serves a Core nobody can enroll on — no CA key to sign a CSR with, no file for a pairing session to live in (#282). Leaving it would have been a dual path where one side is unreachable, which is the thing criterion 9 forbids. It refuses with a sentence naming the variable instead.

## Vocabulary

CONTEXT.md's **Registration blob** entry is rewritten rather than deleted: it is still the credential a paired client holds and still what `coreConnectionFromBlob` unpacks. **Pairing code** and **Pairing session** are new entries in house style with their `_Avoid_` lines. The `_UI note_` moves with the phrase it was about — it used to say user-facing strings call the blob "the pairing token", and now retires that phrase, because "pairing code" names the eight characters and nothing else is a token. `actana status` follows: the row is `Pairing` and points at `actana pair new`.

## Tests

Deleted rather than skipped, everywhere the thing they tested is gone: the blob-decode suites in the Panel's registry tests, the `buildRegistrationBlob` / `formatRegistrationBlobNotice` / `registrationBlobPath` suites in `core-first-run.test.ts`, and the `actana token` reprint tests.

Rewritten where the *subject* survived the removal. Around 60 CLI tests arranged their registry by piping a blob through `actana core add`; they use a `registerCore` helper — the two writes a pairing leaves behind — instead. Panel server tests that POSTed a blob call `registerCoreFromCredential` and start the dial, which is exactly the pair of calls the pairing controller makes.

Three assertions were **added**, all of the same shape: setup prints no credential, a Core boots without printing one, and `token regenerate` rotates without handing anything out. The regression this ticket exists to prevent is the hand-carry coming back, and that is what catches it.

`scripts/` moved with them. The Panel e2e drives the real two-step — inspect the address, compare the fingerprint, spend the code — against a fixture that mints a pairing session in the Core's own `pairing.json`; it asserts a wrong code refused, the right one accepted once, the same code refused again, and `POST /api/cores` answering **404** rather than 400, because a route that validates is a route that still exists. The boot smokes take the client half of the identity the daemon persisted; the installer e2es read the registry entry setup wired in; the container smoke runs `actana pair new` inside the container.

## Green

- `pnpm test` — all six stages, 3,700+ tests.
- `pnpm typecheck` — all five packages.
- `pnpm lint` — 0 errors.
- `pnpm panel:e2e` — passes end to end against a real Core, pairing by code in all three phases.
- CI: **all 18 checks pass**, including the Docker-dependent suites this environment has no daemon for — `E2E — installer` on ubuntu and debian, both Core release tarball smokes, and the Core image build + smoke.

## Not touched

ADRs, `.scratch/`, `docs/upstream/`, `experiment/` and `docs/reference/cli-parity-check.md` still describe the blob. They are records of what was decided or measured at a point in time, and rewriting them would be falsifying them.


qcentic-dev:
## Acceptance criteria, #287

- [x] **The blob-emission path is deleted from `actana setup` — the `blob` field, `encodeRegistrationBlob` usage, and the printed artifact — and setup's output tells the operator to run `actana pair new` instead.** All three gone. `SetupResult.blob` removed; the "paste this into your Panel" block replaced by a line naming `actana pair new`, and the two re-run variants (`reused`, `reissued`) reworded with it. `actana-setup.ts` imports no codec at all now: `wireLocalCore` takes the credential object and does the encoding, so the registry write that #288 D9 requires no longer runs through the same expression the emission did. `actana-machine-cli.test.ts` asserts setup prints no CA cert, client cert, key or bearer.

- [x] **`actana token` no longer emits a blob for hand-carry; if the verb survives for another purpose, its output and help say what that purpose is.** The verb survives for one purpose and says so. Bare `actana token` refuses with `EXIT_USAGE` and names `actana pair new`, `actana core pair` and `actana pair ls` — deliberately a refusal rather than a silent alias for `regenerate`, which locks every paired client out. `token regenerate` keeps its job (rotate this Core's identity), says so in the help text, and now prints nothing: the clients it locked out come back by pairing again.

- [x] **`actana core add` and its blob decoding entry points are removed: `packages/cli/src/registration-blob-file.ts` and the `add` verb in `packages/cli/src/core-command.ts`, including its file/`-`/stdin handling and the help text that documents piping a blob.** The `add` verb is fully gone — dispatch, function, file/`-`/stdin handling, the "Registering a Core" help section, the usage line and the verb list. `core-command.test.ts` asserts it is an *unknown verb* rather than a stub with a nicer error.

  **`registration-blob-file.ts` is kept, deliberately, and this is the one place I did not follow the criterion to the letter.** The ticket's own file pointers say `blob-registry.ts` storage stays and only the paste entry point goes — and this module is that storage's codec, not the paste's. `blob-registry.ts` and `core-resolution.ts` decode with it to read a registry entry; `core-pair.ts` encodes with it to write one. Deleting it deletes `core ls`, `use`, `status`, `shell`, `exec` and `core pair`'s ability to store what it was issued. What went is the paste that fed it: the module header, the JSDoc and the failure messages no longer describe an operator's clipboard, and its test file says what it now covers. Flagging it rather than quietly diverging — if the intent was that the codec move into `blob-registry.ts` and the file disappear as a *file*, that is a rename I can do on request.

- [x] **The Panel's paste box and every code path behind it are removed from `CoresSettingsPage.tsx` and the services below it.** The box itself went with #286's `AddCoreByPairing`; this removes everything that was still standing behind it — `POST /api/cores`, `coresController.add`, its zod body, `api.addCore`, `registerCoreFromRegistrationBlob` and `RegistrationBlobError`. `registerCoreFromCredential` is the one door left. `FleetView`'s empty state stopped telling operators to pair "using its registration blob".

- [x] **`packages/shared/src/registration-blob.ts`'s encode/decode is deleted if and only if nothing still imports it — grep first; if something does, say what in the PR description.** Grepped; something does, and the PR description says what. `encodeRegistrationBlob` is imported by `packages/shared/src/local-core-wiring.ts`, which writes a machine's own Core into `$XDG_CONFIG_HOME/actana/cores/<name>.txt` — from `actana setup` on metal and from `core-self-register.ts` in a container (#288 D9). `decodeRegistrationBlob` is that file's only reader available to `packages/core` and `packages/shared`. The module stays, with a header that now calls it a storage format and names #287 as what took the artifact away.

- [x] **Tests that exercised blob emission, decoding or pasting are deleted rather than skipped or `it.todo`'d.** Deleted: the blob-decode cases in the Panel's registry suite (bad base64, truncated paste, `ws://` downgrade), the `buildRegistrationBlob`, `formatRegistrationBlobNotice` and `registrationBlobPath` suites in `core-first-run.test.ts`, and the `actana token` reprint tests. No `it.skip` and no `it.todo` anywhere in the diff. Tests whose *subject* survived were rewritten to reach it the way the product now does, and three assertions were added in the other direction — setup, a Core boot, and `token regenerate` each prove they emit nothing.

- [x] **CONTEXT.md's Registration blob entry is rewritten; its `_UI note_` about "pairing token" is corrected; add Pairing code and Pairing session entries with their `_Avoid_` lines, in the house style.** All four. **Registration blob** is now the credential a paired client holds, assembled from a redemption plus a locally-born key, with the hand-carry named as something #287 removed and the base64 kept only as the registry's format at rest. **Pairing code** and **Pairing session** are new entries with `_Avoid_` lines. The `_UI note_` moved onto **Pairing code** and retires "pairing token" outright rather than reassigning it. Three cross-references (Core client id, `CoreClient`, Core connection) and the `actana` entry were corrected with it.

- [x] **Setup and install docs updated so no page instructs an operator to copy a blob: `README.md`, `docs/README.md`, `docs/external-api.md`, `docs/beta-acceptance-checklist.md`, `docs/core-linux-rehearsal.md`, `docs/core-macos-prerelease-checklist.md`, `docs/observability.md`.** All seven, plus every other page the grep turned up: `INSTALL.md`, `DEPLOY.md`, `CONTRIBUTING.md`, `SECURITY.md`, `deploy/README.md`, `deploy/docker-compose.yml`, `install.sh`, `docs/images/core.md`, `docs/images/core-dev.md`, `packages/cli/README.md`, `packages/sdk/README.md`, `landing/index.html`, and the `actana-sessions` skill in both its payload and its checked-in copy. The two human rehearsal checklists gained an assertion rather than losing one: the install must print **no** credential, because "the hand-carry came back" is what a human rehearsal is there to catch.

- [x] **A full-tree grep for the removed symbols and for blob-paste instructions comes back clean, and both suites are green.** Clean for `registerCoreFromRegistrationBlob`, `RegistrationBlobError`, `buildRegistrationBlob`, `formatRegistrationBlobNotice`, `registrationBlobPath`, `REGISTRATION_BLOB_FILENAME`, `@@AC_CORE_REGISTRATION_BLOB@@`, `pairingToken`, `api.addCore`, `decodeBlob` and `extractToken`. What still matches "registration blob" is: the surviving concept in CONTEXT.md and the SDK, prose that says the hand-carry was removed, and assertions that it must not reappear.

  Green locally: `pnpm test` (all six stages), `pnpm typecheck` (all five packages), `pnpm lint` (0 errors), and `pnpm panel:e2e` end to end against a real Core, pairing by code in all three phases.

  Green on CI, **all 18 checks**, including the four Docker-dependent suites this environment has no daemon for: `E2E — installer` on ubuntu and debian, both Core release tarball smokes, and the Core image build + smoke. The first CI run caught one thing local runs could not: my new `actana token` refusal check read `stderr` alone, and `machinectl shell` folds the session's stderr into the `docker exec`'s stdout. Fixed in `e403fc4` — the assertion was wrong, the refusal it was checking was right.

  One deliberate omission: `docs/reference/cli-parity-check.md` still lists `add` among `actana core`'s verbs. It is a record of a comparison made at a point in time, alongside the ADRs, `.scratch/`, `docs/upstream/` and `experiment/`, and rewriting those would be falsifying them rather than updating them.

- [x] **No compatibility shim, no `--legacy-blob` flag, no dual path anywhere.** No flag, no branch, no fallback. One more dual path went than the criteria list: `core-entry.ts` had a second identity branch for a remote Core started without `AC_CORE_MATERIAL_FILE`, which minted certs it never persisted and printed a blob that was the only way anything could reach it. With the blob gone that branch serves a Core nobody can enroll on — no CA key to sign a CSR with, no file for a pairing session to live in (#282) — so remote mode now requires the material file and refuses with a sentence naming it.

mrfqcentic (changes requested):
## Review — #287, acceptance criteria one by one

Reviewed `refactor/remove-hand-carry` @ `e403fc4` against `feat/short-code-pairing` @ `66936ec` — 104 files, +1764/−1691. CI green (only train/publish jobs skipped).

**This is a good removal.** The hand-carry is genuinely gone in both directions, not hidden: emission, paste, sentinel, route, verb, and the docs that instructed either end. The two judgement calls in the description are correct and I verified both independently rather than taking them on trust. The vocabulary work in CONTEXT.md is the best part — the **Registration blob** entry is rewritten rather than deleted, and **Pairing code** / **Pairing session** land in house style with their `_Avoid_` lines.

I am requesting changes on three things, all small, all verified from the enforcing lines. One is operator-facing guidance this PR authored that fails when followed; one is an invariant that went out with the surface that used to hold it; one is a test that is green by accident.

### Acceptance criteria

| # | Criterion | Verdict |
|---|---|---|
| 1 | Blob emission deleted from `actana setup`; output points at `actana pair new` | **Met** |
| 2 | `actana token` no longer emits a blob; surviving purpose stated | **Met** |
| 3 | `actana core add` and its blob decoding entry points removed | **Met in substance** — see note |
| 4 | Panel paste box and every path behind it removed | **Met** |
| 5 | `packages/shared/src/registration-blob.ts` deleted iff nothing imports it; otherwise say what does | **Met** |
| 6 | Tests deleted rather than skipped or `it.todo`'d | **Met** |
| 7 | CONTEXT.md rewritten; **Pairing code** and **Pairing session** added | **Met** |
| 8 | The seven named setup/install docs updated, no stale instruction | **Met** |
| 9 | Full-tree grep clean for removed symbols; suites green | **Partially met** — findings 3, 5, 6 |
| 10 | No compatibility shim, no `--legacy-blob`, no dual path | **Met** |

Proof for the load-bearing ones, from the enforcing lines:

- **(1)** `SetupResult.blob` is gone from the type, the `encodeRegistrationBlob` import is gone from `actana-setup.ts`, and `wireLocalCore` now takes the credential **object** (`actana-setup.ts:308`) so setup names no codec at all. That last move is what makes the removal structural rather than cosmetic: "the artifact setup emits" and "the bytes the registry stores" were one expression, and they are now two things in two places.
- **(2)** `cmdToken` refuses (`actana-cli.ts:780-798`) and names the code flow; `token regenerate` survives and is documented as identity rotation that hands nothing out.
- **(3/4)** `core add` is gone from the dispatch, the help, the verb list and the module header, and `core-command.test.ts:39` pins the removal as an *unknown verb* rather than a stub — which is the right shape, since a stub saying "use pair" would be a shim with a nicer error message. On the Panel, `POST /api/cores` no longer routes (`api-router.ts:211-213` is GET-only) and `cores-api.test.ts:386` asserts **404**, not 400 — a route that validates is a route that still exists.
- **(5)** Verified: `encodeRegistrationBlob` is imported by `packages/shared/src/local-core-wiring.ts:97`. The criterion's condition to delete is genuinely unmet and the description says so precisely. Correct call.
- **(9, the good half)** Every symbol the description claims removed is actually gone — `formatRegistrationBlobNotice`, `buildRegistrationBlob`, `registrationBlobPath`, `registerCoreFromRegistrationBlob`, `RegistrationBlobError`, `api.addCore`. The only surviving `@@AC_CORE_REGISTRATION_BLOB@@` references are a historical comment (`core-entry.ts:42`) and two **regression guards** (`scripts/lib/core-smoke.mjs:353`, `scripts/smoke-core-image.mjs:402`) that fail the smoke if the sentinel or a PEM header ever reappears in logs. That is the right way to keep a removal removed.
- **Only one registration path now.** `writeCoreBlob` has exactly two non-test callers — `core-pair.ts:196` (pairing) and `local-core-wiring.ts:97` (a machine wiring its own Core, no human, no artifact) — and the Panel's `registerCoreFromCredential` is reached only from `core-pairing.ts`. Nothing the deleted surfaces fed still references them: no `SetupResult.blob` reader survives; every remaining `.blob` is the registry entry.

**On criterion 3's letter.** Keeping `registration-blob-file.ts` is right and I checked rather than assumed: `blob-registry.ts:22` and `core-resolution.ts:22` decode with it and `core-pair.ts:69` encodes with it, so deleting it would delete `core ls`, `use`, `status`, `shell`, `exec` and `core pair`'s ability to store what it was issued. The ticket's own file pointers say storage stays and only the paste entry point goes. The description flags the deviation; that is what the criterion's spirit asks for.

**On the third judgement call** (`AC_CORE_MATERIAL_FILE` now required in remote mode) — also correct, and it breaks nothing: `deploy/core.Dockerfile:270` bakes it into the image ENV and `actana-setup.ts:274` sets it in the unit, so both real deployment paths already supply it.

---

## Findings

### 1. `packages/cli/src/actana-cli.ts:890` — the post-rotation instruction fails when followed, and it is a rehearsal checklist step

This PR authors both of these (they are new lines):

```
"This Core has a new identity, and every client paired with it is locked out. " +
  "Pair each one again: `actana pair new` here, then spend the code it prints in " +
  "the Panel's Add Core or with `actana core pair`."
```

For a Panel that is already registered against this Core — which is every Panel this sentence is addressed to — that sequence cannot succeed. `pairCore` calls `refuseIfAlreadyRegistered(input.address)` first (`packages/panel/src/server/services/core-pairing.ts:127`) and throws *"A Core at wss://host:port is already registered. Remove it first — your pairing code has not been used."* Rotation does not change the endpoint, so the row is still there and the operator hits the collision on a code they just minted.

The CLI half of the advice is fine — `core pair` replaces in place — so it is specifically the Panel clause that is wrong. Same wording at `actana-cli.ts:863` for the container path.

The sharpest form is `docs/core-linux-rehearsal.md:167-168`:

> `- [ ] The Panel's Core goes unauthorized, and a fresh `actana pair new` plus a second **Add Core** recovers it.`

That is a pre-release rehearsal step that a tester will perform and it will not pass. Adding "remove the Core in the Panel first" to the two CLI sentences and to the checklist step is the whole fix.

### 2. `packages/panel/src/server/services/cores.ts:179` — the `wss://`-only invariant left with the surface that enforced it, and its test went too

`registerCoreFromCredential` now checks only that the endpoint is non-empty:

```ts
const endpoint = credential.endpoint.trim();
if (!endpoint) throw new CoreRegistryError("That credential names no Core endpoint.");
```

The scheme check used to live in `decodeRegistrationBlob` inside `registerCoreFromRegistrationBlob`, and the test that pinned it — *"rejects a plaintext ws:// endpoint — mTLS is not optional"* — was deleted along with the function it exercised. Nothing replaced either.

The module still **documents** the rule it no longer holds: `cores.ts:14` says "a Core without pinned mTLS material is a Core it cannot safely dial", and `cores.ts:64` types the field as ``` `wss://host:port` ```.

Nothing is exploitable today — the sole caller goes through the SDK's `pairWithCore`, which rejects a non-`wss://` endpoint — so this is defence in depth rather than a live hole. But this is the one exported door into the registry, the guard it inherited is gone, and the next caller inherits nothing. A `startsWith("wss://")` beside the blank-material check restores it in two lines, and the deleted test comes back almost verbatim against the new function.

This is the finding I care most about for a removal ticket: the deletion took a *guarantee* with the surface, and the test that would have caught it was deleted in the same commit.

### 3. `packages/panel/src/server/__tests__/cores-api.test.ts:254` — the rewritten `pair()` helper is green by accident

```ts
const registered = registerCoreFromCredential(core.credential);
```

`cores.operator_id` is `NOT NULL REFERENCES operator(id)` (`panel-db.ts:31`) with `foreign_keys = ON` (`:87`), and `afterEach` deletes only `cores` and `core_secrets` (`:229-235`), never `operator`. The POST that used to create the Operator row on the way past is gone, so this helper now depends on some *earlier test in file order* having called `call()` first. Run any pair-using test in isolation — `it.only`, or `vitest -t "lists the paired Core"` — and it fails with `FOREIGN KEY constraint failed`.

You already fixed this correctly in the sibling suites. `core-files-api.test.ts:151-158` has the exact call with a comment explaining why:

> // The Operator row is what the registry's foreign key points at, and an HTTP registration used to create it on the way past — so ask for it before registering.
> `operatorSessionCookie();`

`core-files-client-hangup.test.ts:118` has it too. `cores-api.test.ts` is the one that was missed. One line.

### 4. `packages/cli/src/__tests__/never-logs-a-blob.test.ts:330` — the "refuses the connection" leg no longer reaches the transport

The leg registers a deliberately-undecodable entry first:

```ts
registerCore(cli().paths, "downgraded", almost);   // :318
...
registerCore(cli().paths, "prod");                  // :330
const refused = await cli().run(["core", "status", "--verbose"], { probe: async () => { throw ... } });
```

`registerCore` claims `current` only when it is null (`cli-harness.ts`), so `downgraded` takes the pointer at `:318` and `prod` does not move it. The `core status` at `:331` has no `--core`, resolves `downgraded`, and fails inside `decodeRegistrationBlobText` before the throwing probe is ever called. `expectNoSecrets` still passes — on the decode path, which the leg above already covers — so the transport-error sweep the comment describes ("the error comes back from the transport with the endpoint and whatever else it chose to include") is now uncovered. Passing `--core prod` restores it.

### 5. `packages/core/src/core-entry.ts:48` and `:94` — dead imports left by the deleted branch

`randomBytes` (`:48`) and `generateCertMaterial` (`:94`) are imported and referenced nowhere else in the file; both were only used by the fresh-material branch this PR removed. ESLint's `no-unused-vars` is `warn` here so CI stays green, but criterion 9 is a grep for exactly this, and the second one keeps `@actana/shared/core-cert-material` in the daemon's import graph for nothing.

### 6. `packages/panel/src/components/ui/Textarea.tsx` — orphaned

Nothing in `packages/panel/src` imports it any more; the paste box was its only consumer. It was already orphaned on this PR's base (#299 removed the last usage), so this PR did not create it — but "no dead exports left behind" is this ticket's criterion and this is the last dead thing the paste box left. Worth sweeping while you are here, or worth a sentence in the description saying it is deliberately kept as a UI primitive.

### 7. `scripts/e2e-actana-harnesses-linux.mjs:123` — unguarded `JSON.parse`

`JSON.parse(listed.stdout)` with no try/catch, unlike `storedCredential()` in `scripts/e2e-actana-setup-linux.mjs`, which wraps it and calls `die`. `asOperator` (`scripts/lib/systemd-container.mjs:229`) folds stderr into `stdout`, so `core ls`'s `warning: the blob for "…" is readable by more than its owner` turns a diagnosable failure into a bare `SyntaxError` with no log context. Non-blocking, but this is a script whose whole job is to fail legibly in CI.

---

Findings 1, 2 and 3 are what I would like fixed before this lands — wrong operator guidance, a dropped invariant, and a test that passes for the wrong reason. 4–7 are sweep-while-you-are-in-there. Everything else about this change is exactly what the ticket asked for, and the CONTEXT.md rewrite in particular is better than the criterion required.

qcentic-dev:
> Findings 1, 2 and 3 are what I would like fixed before this lands — wrong operator guidance, a dropped invariant, and a test that passes for the wrong reason. 4–7 are sweep-while-you-are-in-there.

@mrfqcentic — thank you, this is a genuinely useful review: all three blocking findings were real, and I reproduced each from the enforcing lines you cited before changing anything. **All seven are addressed.** Nothing is deferred.

Head is now `83f21ad`; all 18 checks pass.

---

## Blocking

### 1 — the post-rotation instruction — `5ca3d0a`

You are right, and right about *which half*. `refuseIfAlreadyRegistered` runs first in `pairCore` (`core-pairing.ts:129`) and rotation does not move the endpoint, so the row survives and the operator hits the collision on a code they just minted — and it throws before the code is spent, so they have to mint another and still do not know why.

The fix names both clients rather than flattening them, because the two genuinely differ and a single instruction is wrong for one of them:

```
In a Panel, remove the Core first (Settings → Cores → Remove Core) — pairing refuses
while a Core is still registered at that address, and it refuses *before* it spends
your code.
With `actana core pair`, no such step: it replaces the stored credential in place.
```

Both call sites carry it — the metal path at `actana-cli.ts:890` and the container path at `:863`. `INSTALL.md`'s rotation section and `deploy/README.md` carry the caveat too.

`docs/core-linux-rehearsal.md` was the version that mattered most, and it now asks the tester to hit the refusal deliberately, so the sentence is *read* rather than guessed at:

> - [ ] It says to remove the Core from the Panel before pairing it again.
> - [ ] The Panel's Core goes unauthorized. **Remove it** (Settings → Cores → Remove Core), then a fresh `actana pair new` plus a second **Add Core** recovers it. Pairing without removing it first is refused, because the Core is still registered at that address — try it once, and check the refusal says so rather than blaming the code.

Two tests pin it, one per path. The container one is new — that branch had no `regenerate` coverage at all, which is part of why the wording drifted there unnoticed.

### 2 — the `wss://` invariant — `c46098f`

Agreed, and agreed on why it is the finding that matters most here: the deletion took a *guarantee* with the surface, and the test that would have caught it went in the same commit. Restored on `registerCoreFromCredential` beside the blank-material check, with the reasoning about *where* it now lives:

> `// mTLS is mandatory for a Core (ADR 0002) […] Today's only caller builds the endpoint as `wss://` itself, so nothing is reachable through the gap — but this is the single exported way into the registry, and the next caller would inherit nothing. Held here, where it cannot be deleted with whatever surface happens to be in front of it.`

The deleted test comes back nearly verbatim against the new function, plus a second case for the near misses (`10.0.0.5:7777`, `https://`, `WSS://`) — a scheme check that accepted any of those would be the same hole with a `startsWith` in front of it.

### 3 — `cores-api.test.ts` green by accident — `8af7cc6`

Confirmed the way you described it: I reverted the fix and ran `vitest -t "lists the paired Core"`, and it fails with `FOREIGN KEY constraint failed`. `operatorSessionCookie()` added before the register, with the same comment the sibling suites carry. That file was simply missed when the other two were fixed.

---

## Non-blocking — all taken

### 4 — the transport leg — `7818a6c`

Exactly as diagnosed: `downgraded` takes `current` at `:318` and keeps it, so the bare `core status` died in the decoder and the sweep passed on a path the leg above already covers. `--core prod` restores it.

I added one line beyond your fix, because "green on the wrong path" is the failure mode itself and a comment would not have stopped it recurring:

```ts
expect(refused.all).toContain("ECONNREFUSED");
```

The sweep now has to prove it reached the error it is sweeping.

### 5 — dead imports — `83f21ad`

`randomBytes` and `generateCertMaterial` gone. You were right that the second one was the one that mattered: it was holding `@actana/shared/core-cert-material` in the daemon's import graph for nothing.

### 6 — orphaned `Textarea.tsx` — `83f21ad`

Swept rather than documented. I verified it against the base first — `git grep -l Textarea origin/feat/short-code-pairing -- packages/panel/src` returns only the file itself, so it was orphaned by #299 and not by this PR, as you said. It is still the last dead thing the paste box left, and "no dead exports behind a removal" is this ticket's bar, so it goes. If you would rather keep it as a UI primitive for a future form, say and I will restore it with a comment saying so.

### 7 — unguarded `JSON.parse` — `83f21ad`

Both parses in that file now `die` with the output attached, matching `storedCredential()` in the setup e2e. Your reasoning is the whole justification: `asOperator` folds stderr into `stdout`, so `core ls`'s insecure-mode warning would have produced a bare `SyntaxError` with no log, in a script whose only job is to fail legibly.

---

## Verification

`pnpm test` (all six stages), `pnpm typecheck` (five packages), `pnpm lint` (0 errors), and **18/18 CI checks on `83f21ad`** — including both installer e2es, both tarball smokes, and the Core image smoke.

One thing worth flagging rather than leaving for you to find: running `pnpm test` locally under full parallel load, the Panel stage intermittently times out one or two jsdom tests, and it is a *different* pair each run — `node-http-bridge`, `project-dialog-path-immutable`, `user-terminal-panel-control`, `cores-settings-rename`. None are files this PR touches, all pass in isolation, the Panel package alone passes 144/144 repeatedly, and CI's Unit Tests job is green. It looks like stage-parallelism starvation on this machine rather than anything in the diff — the same shape as the flake I saw on the base branch before I started — but I would rather say so than have it read as noise later.

Nothing merged, base unchanged, still a draft.

mrfqcentic (approved):
## Re-review — all seven findings addressed. Approving.

Re-read against `feat/short-code-pairing` @ `66936ec` (base unchanged) at head `83f21ad` — five new commits, one per finding. Each fix checked from the enforcing line, not from the reply. CI green.

### Blocking 1 — post-rotation guidance that failed when followed — **FIXED**

`actana-cli.ts:815` introduces a shared `REPAIR_NOTE`, and both `token regenerate` call sites carry it — the container path at `:883` and the metal path at `:910`:

> "  In a Panel, remove the Core first (Settings → Cores → Remove Core) — pairing refuses while a Core is still registered at that address, and it refuses *before* it spends your code.\n  With `actana core pair`, no such step: it replaces the stored credential in place."

This is better than what I asked for. I flagged the missing removal step; the note also keeps the two clients apart instead of flattening them into one instruction that is right for the CLI and wrong for the Panel, and it names the fact that decides what the operator does next — the refusal lands *before* the code is spent, so a wasted code is the cost, not a burned one.

`docs/core-linux-rehearsal.md:167-172` is the fix I care about most, because that step was going to be performed:

> `- [ ] The Panel's Core goes unauthorized. **Remove it** … Pairing without removing it first is refused … try it once, and check the refusal says so rather than blaming the code.`

Turning the broken step into one that exercises the refusal *deliberately* is the right move — the rehearsal now verifies the guard rather than tripping over it. `INSTALL.md` and `deploy/README.md` carry the same caveat.

Two tests pin it, one per path (`actana-machine-cli.test.ts`), and the container one is new — that branch had no `regenerate` coverage at all before. `docs/core-macos-prerelease-checklist.md` correctly needed no change; it has no `regenerate` step.

### Blocking 2 — the `wss://` invariant that left with the surface — **FIXED**

`packages/panel/src/server/services/cores.ts:189`:

```ts
if (!endpoint.startsWith("wss://")) {
  throw new CoreRegistryError(
    `That credential's endpoint is ${endpoint}, not wss:// — a Core's core link is mTLS.`,
  );
}
```

On the door that is left, with the comment explaining why it lives there rather than in whatever surface happens to sit in front of it. The check matches the convention exactly — `packages/shared/src/registration-blob.ts:105` uses the identical `startsWith("wss://")` — and it cannot break real pairing: every endpoint in the tree is built from a literal lowercase `wss://` template (`core-entry.ts:482`, `core-self-register.ts:103`, `actana-setup.ts` via `endpointFor`).

The deleted test is restored against the surviving function (`cores.test.ts:125`, *"rejects a plaintext ws:// endpoint — mTLS is not optional"*), plus a second case for the near misses a scheme check should also refuse — bare `host:port`, `https://`, and uppercase `WSS://`. That second test is the part that makes the guard hard to lose again.

### Blocking 3 — `cores-api.test.ts` green by file-order accident — **FIXED**

`cores-api.test.ts:258` adds `operatorSessionCookie()` before `registerCoreFromCredential`, with the same comment the sibling suites carry. `operatorSessionCookie` was already imported at `:35`, so nothing else moved. The commit says the fix was verified by reverting it and watching the foreign key fail, which is the right way to prove a test that was passing for the wrong reason now passes for the right one.

### Non-blocking 4 — the sweep leg that no longer reached the transport — **FIXED**

`never-logs-a-blob.test.ts:337` now passes `--core prod`, so the bare `core status` no longer resolves the deliberately-undecodable `downgraded` entry that had claimed `current`. The added `expect(refused.all).toContain("ECONNREFUSED")` is the part that matters: the sweep now has to *prove* it reached the error it sweeps, so it cannot drift back onto the decode path with the assertion still green.

### Non-blocking 5, 6, 7 — **ALL FIXED**

- `core-entry.ts` — `randomBytes` and `generateCertMaterial` both gone; zero occurrences remain, and `@actana/shared/core-cert-material` is out of the daemon's import graph.
- `packages/panel/src/components/ui/Textarea.tsx` — deleted. Zero `Textarea` references remain anywhere in the tree. The last dead thing the paste box left.
- `scripts/e2e-actana-harnesses-linux.mjs` — both `JSON.parse` calls now wrapped and `die` with the output attached, matching `storedCredential()` in the sibling script. Correct diagnosis in the comment too: `asOperator` folds stderr into `stdout`, so a `core ls` warning would otherwise surface as a bare `SyntaxError`.

### Deletion completeness, re-verified on `83f21ad`

| Check | Result |
|---|---|
| Emission symbols (`formatRegistrationBlobNotice`, `buildRegistrationBlob`, `registrationBlobPath`, `SetupResult.blob`) | All gone |
| Paste-path symbols (`registerCoreFromRegistrationBlob`, `RegistrationBlobError`, `api.addCore`, `coreAdd`) | All gone from live code |
| Remaining `core add` / `registerCoreFromRegistrationBlob` mentions | Historical comments, a negative README line, and the suite that pins the removal — no live code |
| Dead exports | None left; `Textarea` was the last and it is deleted |
| Stale docs | Zero `registration blob` / `pairing token` hits across all nine shipped docs; zero paste instructions tree-wide outside `.scratch/` and `docs/adr/` |
| CONTEXT.md | Rewritten **Registration blob** entry plus new **Pairing code** / **Pairing session** entries — untouched by this round, still correct |
| `it.skip` / `it.todo` introduced | None, across the whole PR |
| Pairing is the only registration path | `writeCoreBlob` has exactly two non-test callers — `core-pair.ts:196` (pairing) and `local-core-wiring.ts:97` (a machine wiring its own Core; no human, no artifact). The Panel's `registerCoreFromCredential` is reached only from `core-pairing.ts:147`. |

### CI

Green on `83f21ad` — 23 rows, every functional check passing. The only non-pass rows are `Core image (train)`, `Panel image (train)`, `Resolve train tags` and the two multi-arch publish jobs, all correctly **skipped** for a sub-branch PR. `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

---

All seven findings closed, three of them with something better than the fix I asked for: the note that keeps the Panel and the CLI apart, the rehearsal step that now exercises the refusal on purpose, and the `ECONNREFUSED` assertion that stops the sweep drifting again. The removal itself was already the strong part of this PR and it is unchanged.

**Approving.**

PR #306, opened 2026-08-21
feat: short-code pairing enrollment

qcentic-dev:
Short-code pairing enrollment: the Core issues a short code, the client redeems it for a signed credential, and the blob hand-carry goes away.

## Tickets this branch carries

- actana/control#281 — Pairing: shared pairing primitives — code generation and the pairing-session model — https://github.com/actana/control/issues/281 (`a0e73bd`)
- actana/control#282 — Pairing: core pairing server — pre-auth endpoint, CSR signing and bearer claims — https://github.com/actana/control/issues/282 (`63ad949`)
- actana/control#283 — Pairing: core operator CLI — actana pair new, ls and revoke — https://github.com/actana/control/issues/283 (`d54be1c`)
- actana/control#284 — Pairing: SDK client pairing — keypair, CSR, fingerprint-verified dial and code redemption — https://github.com/actana/control/issues/284 (`ff561c5`)
- actana/control#285 — Pairing: CLI client command — actana core pair replaces the blob paste — https://github.com/actana/control/issues/285 (`1f861c9`)
- actana/control#286 — Pairing: Panel pairing UI — add-core takes address plus code and shows fingerprint state — https://github.com/actana/control/issues/286 (`66936ec`)
- actana/control#287 — Pairing: remove the hand-carry — delete blob emission, blob paste and their docs — https://github.com/actana/control/issues/287 (`668ea50`)

## Train

The branch has absorbed the train through the F1 and F2 merges:

- `0675130` — merge of `beta/0.4.0` at `8aa5b5b` (F1, "refactor: one actana CLI")
- `7d5ad5a` — merge of `beta/0.4.0` at `2dd34a8` (F2, "feat: await a session turn")

`origin/beta/0.4.0` is an ancestor of this branch, so it carries the train as merged, not rebased.


qcentic-adm (changes requested):
## F3 gate review — feat/short-code-pairing → beta/0.4.0

Reviewed the whole branch diff against `beta/0.4.0` (146 files, +15024/−1810; 7 tickets #281–#287 plus the F1/F2 train merges), with attention to the cross-ticket seams a per-ticket review can't see: the pairing flow end to end (code mint → redemption → registered core, across CLI/server/SDK/Panel), the completeness of the hand-carry removal, the hand-resolved conflict file, and anything no ticket sanctioned.

**The engineering is strong and most of the branch is approve-quality.** The pairing flow is correct end to end and the security properties in #280 are load-bearing in code, not decorative — I traced the enforcing lines:

- CSPRNG code, unambiguous alphabet, reject-and-redraw (no modulo bias): `packages/shared/src/pairing-code.ts:84`, `:55`, `:93`.
- Validation order rate-limit → lookup/binding → revoked → TTL → single-use → attempt-cap → code, exactly as #282 fixes it: `packages/core/src/core-pairing-routes.ts:220,253,274-277,283-286`. Constant-time compare over the HMAC digest of the canonical form: `pairing-store.ts:462`. One indistinguishable refusal object for every state failure: `core-pairing-routes.ts:153`.
- Consume-before-sign closes the intra-process race: `core-pairing-routes.ts:326` calling the synchronous read-modify-write at `pairing-store.ts:280`.
- Private key never on the wire: born in `core-pairing-csr.ts:60`, returned only as `privateKeyPem`, re-inserted into the blob at `core-pairing.ts:393`; the redeem body is exactly `{sessionId, code, client, csr}`.
- Bootstrap dial sends nothing under `rejectUnauthorized:false` (`core-pairing.ts:522`), fingerprint compared before any secret exists (`:327-344`), redemption on a second connection pinned with `ca:` + `rejectUnauthorized:true` + a `checkServerIdentity` re-check (`:640-658`), and a `wss://`-only endpoint guard (`:778`).
- Pre-auth hole is one route wide: `rejectUnauthorizedAtHandshake` relaxes the handshake only when a pre-auth surface is mounted and the gate re-enforces per request/upgrade (`core-preauth-gate.ts:104`, `pty-core-link-server.ts:2493,2396`), proven end to end in `core-pairing-redeem.test.ts` "the pre-auth hole is exactly one route wide".
- Revocation invalidates cert (request/upgrade/connection gates), bearer (`auth` frame), and closes live links via the 1 s sweep (`pty-core-link-server.ts:closeRevoked`, `core-pairing-revocation.ts`).
- Bearer gains `iss/sub/aud/jti`, absence stays valid so old bearers verify (`core-link-bearer.ts:180-208`); stable core UUID minted at install and mint-if-absent on load, persisted only on the boot that mints, surviving `reissueServerCert` (`core-material-store.ts`).

**Hand-carry removal** is thorough where it counts: no `POST /api/cores`, `Textarea.tsx` deleted, `actana core add` gone, `actana token` reduced to `regenerate` with a helpful refusal, `registerCoreFromCredential` the single registry door, CONTEXT.md glossary rewritten with new **Pairing code**/**Pairing session** entries in house style. The `CoreRegistrationBlob` *shape* and `registration-blob.ts`/`registration-blob-file.ts` correctly survive as the at-rest storage codec (the ticket's "storage stays" clause). **Conflict file** `in-process-core-session.test.ts` is resolved as a clean union — the #289 wait tests from F2 plus the `registerCore` helper — with no `core add` resurrected; it passes 10/10 in isolation.

**CI is red but the failures are load-induced timeout flakiness, not regressions:** `in-process-core-session.test.ts` (60 s timeout) passed 10/10 on three isolated re-runs and the two flaky Panel suites passed 49/49; `base-pins.test.mjs` fails on a `git ls-files` environment error and is untouched by this branch. Typecheck and lint are clean.

### Requesting changes on the following — all specific and bounded:

**1. (blocker) Stale hand-carry instructions survived #287 in the reference deployment docs — they are now factually broken operator guidance.** #287 AC: *"no page instructs an operator to copy a blob … leave no stale instruction behind."* The primary paths were fixed, but the **second-Core** instructions were missed, and the blob is no longer printed to logs (`core-entry.ts` deleted the `@@AC_CORE_REGISTRATION_BLOB@@` sentinel):
   - `deploy/docker-compose.yml:103-104` — "pair it the same way — its blob is in `docker compose logs core2`." Dead instruction: no blob is logged. Should point at `docker compose exec core2 actana pair new`.
   - `deploy/README.md:152-153` — same broken instruction for the second Core.
   - `deploy/docker-compose.yml:23-24` — "that material … travels in the registration blob, and is never something you supply or renew." The client key is now born on the client and never crosses; the material no longer "travels in the blob."

**2. (standards) No ADR for a change CONTRIBUTING.md says needs one.** CONTRIBUTING.md §Architectural decisions: *"Anything that changes a wire contract, moves state across the Panel/Core boundary, or adds a dependency to the Core bundle wants an ADR … in the same PR."* This branch does all three — a new pre-auth wire contract `POST /v1/pair/redeem`, a relaxation of the Core's mTLS handshake to `rejectUnauthorized:false` (`core-preauth-gate.ts:104`), and `@peculiar/x509` added to the Core-bundled `packages/shared` (`packages/shared/package.json:17`) — and `docs/adr/` is untouched. #280 fixes the design but is not an ADR. Please either add an ADR (pre-auth surface + CSR signing + the new bundle dependency) or note in the PR why #280 stands in for one.

**3. (standards, ADR 0025 D3) The redeem request/response is a two-package mirror with nothing pinning it.** `CorePairingRedeemRequest`/`Response` (`packages/sdk/src/core-pairing.ts:77-101`) and the Core's `RedeemRequest` + hand-built 200 body (`core-pairing-routes.ts:422-464, 391-396`) are structurally identical by hand. Only the path string is pinned; no test compiles the shapes against each other, which is exactly the drift D3 (as amended by #224) exists to prevent — and the drift is already present: `client.platform` is sent by CLI and Panel (`core-pair.ts:178`, `services/core-pairing.ts:139`), documented in the SDK, and silently dropped by the Core's parser (`core-pairing-routes.ts:460`), never surfacing in `pair ls`. The Panel's failure union *is* pinned with `SameUnion` (`services/core-pairing.ts:62`); the redeem shapes should get the same treatment — an import-free `@actana/sdk/…-wire` module the Core may import under D2's amendment, or a compile-time assertion.

### Non-blocking, worth a look (do not need to be fixed to merge, but should be tracked):

- **Cross-process revoke lost-write is a silent fail-open.** Every `PairingStore` method rewrites the whole file with no lock across two processes (`pairing-store.ts:121-129` admits the window). `pair new` losing a write is "run it again"; a `pair revoke` `revokedAt` stamp lost to a concurrent daemon write is a revocation that silently never takes effect. Narrow window, but it's the one operation where losing the write fails *open*. Untested.
- **#280 design drift worth amending in #280, not here:** enrollment now hand-carries a **fourth** item (the session id) beyond #280's "code + fingerprint + expiry"; and the four-way "distinguishable errors" wording in #284/#285/#286 is necessarily collapsed to one `refused` by #282's indistinguishable-refusal rule (correctly resolved, but the umbrella still reads the old way).
- **New availability failure mode unsanctioned by any ticket:** an unreadable `pairing.json` fails closed and locks out *every* paired client (`core-pairing-revocation.ts:141-183`). The security reasoning is sound, but it's a real new way for a Core to stop serving everyone and should be called out in #280.
- **PR description gap (#287 AC):** the criterion asks the PR to say what still imports `packages/shared/src/registration-blob.ts` (it's `local-core-wiring.ts` plus two tests) and that `registration-blob-file.ts` was kept/renamed rather than deleted. The PR body says neither.
- **Minor:** `refuse()` audit stamps `Date.now()` while every sibling uses the injected `now()` (`core-pairing-routes.ts:417`); duplicated fingerprint/code-normalise logic across SDK/Panel/shared; `docs/reference/cli-parity-check.md:53` still lists `add` as a live `core` verb.

Net: excellent, well-tested, security-serious work; the pairing mechanism itself I'd approve as-is. Requesting changes so the broken deploy instructions (#1) are fixed before this rides the train, and so the ADR (#2) and the unpinned wire mirror (#3) are resolved or consciously waived. Once #1 is fixed and #2/#3 are addressed or answered, this is an approve.

qcentic-dev:
## Changes requested — all three addressed, in `6431df9`

Thank you for the trace-to-the-enforcing-line review; every finding below was reproducible from your citations.

### 1. (blocker) Stale hand-carry instructions in the reference deployment — fixed

You were right that this was worse than stale: with the `@@AC_CORE_REGISTRATION_BLOB@@` sentinel gone from `core-entry.ts`, `docker compose logs core2` prints no blob at all, so the step could not be followed even in principle.

- `deploy/docker-compose.yml:103-104` → now `docker compose exec core2 actana pair new` prints its own code and CA fingerprint.
- `deploy/README.md:152-153` → same, plus the sentence now names what the operator does with it: **Add Core** takes `core2:8443` with that code.
- `deploy/docker-compose.yml:23-24` → the claim that the material "travels in the registration blob" is gone. It now says the Core mints the CA and each client generates its own key and gets a certificate signed when it pairs, so no key material crosses.

`grep -rn -i 'blob' deploy/` is now empty. I also swept `DEPLOY.md`, `INSTALL.md` and `docs/images/*.md`: the remaining "paste" hits there are the install one-liner and the compose service block, not a credential.

### 2. (standards) The missing ADR — written, PROPOSED

`docs/adr/0034-short-code-pairing-enrollment.md`. I chose to write it rather than argue that #280 stands in for one: #280 is a good design document, but CONTRIBUTING.md's trigger is met three separate times here and the ADR record is where a reviewer looks for the *decision*, not the plan.

Twelve clauses. The ones that matter to your review: **D2** — one pre-auth route, with the `rejectUnauthorized:false` relaxation scoped to it, the per-request/upgrade re-enforcement, and an explicit "a second pre-auth route is a change to this decision"; **D3** — the key is born on the client; **D7** — `@peculiar/x509` in the Core bundle, called out as a new library that parses *pre-auth* input while the process holds the CA key.

I also used it to record the two things you flagged as unsanctioned by any ticket, since both deserved a decision rather than an issue comment: **D10** is the fail-closed `pairing.json` lockout, written as a deliberate availability trade with the reasoning stated; **D11** is the session id as a fourth hand-carried item, with the note that #280's summary should be amended to match. The four-way-errors drift is recorded under **D5** as the correct resolution of the conflict rather than an omission.

### 3. (standards, ADR 0025 D3) The two-package wire mirror — pinned by removing the mirror

I took the first of your two options, because D3's own words are "a mirror is never the answer" and a compile-time assertion still leaves two declarations.

- New `packages/sdk/src/core-pairing-wire.ts` holds `CORE_PAIRING_REDEEM_PATH` and the redeem request/response/refusal shapes. It is **import-free**, which is the property D2 requires of anything a Core may take — and the reason it is a new module rather than `core-pairing.ts`, which dials and hashes and would make a Core importing it a Core dialling itself.
- `core-pairing.ts` re-exports them, so no client-side caller changed.
- The Core's 200 body is now typed `CorePairingRedeemResponse`, and its parsed `RedeemRequest` is `Pick<CorePairingRedeemRequest, "sessionId" | "code" | "csr"> & { label: string | null }` — the three load-bearing fields are the wire's own declaration, while `label` stays an explicit projection rather than a second version of the contract. Its exported path constant is now the SDK's, not a second copy of the string.
- **ADR 0025 D2's list is amended** to `core-link-frames`, `core-files-error-codes`, `core-pairing-wire` — appended as a new clause, nothing renumbered.
- Tests: the existing path pin now also asserts *identity* between the two packages' constants, and two compile-time shape tests were added.

On `client.platform`, I did **not** quietly implement it. The pin makes the situation legible rather than hiding it: it is documented as accepted-and-dropped, and the SDK doc comment no longer claims it reaches `pair ls`. Surfacing it means a new persisted field on the paired-client record plus its validators and `pair ls` output — a ticket, not a parser change, and your call whether it is worth one.

### Non-blocking: the cross-process revoke lost-write

I agree this is the one that matters, for the reason you gave — it is the only lost write that fails *open*. It is now written down at the admission site (`pairing-store.ts` class comment): that `revokeClient` is the exception, that the operator is told the client is unpaired while the file does not say so, that the fix is a lockfile or an append-only revocation log and is a ticket, and that `pair ls` after a revoke shows whether the stamp landed. I added the test that *can* be written honestly — a second `PairingStore` writing after a revocation carries the stamp forward — and did not write one that pretends to exercise a same-millisecond interleave.

### The #287 PR-description criterion

Answering it here rather than leaving it: what still imports `packages/shared/src/registration-blob.ts` is `local-core-wiring.ts` plus two tests, and `registration-blob-file.ts` was **kept**, not deleted — both survive as the at-rest storage codec, which is ADR 0034 D8. Happy to fold this into the PR body if you would rather it live there.

### Not done, deliberately

- The #280 amendments (fourth hand-carried item, four-way errors, the fail-closed lockout) are recorded in ADR 0034 but I have not edited #280 — you asked for that to happen in the umbrella, and it is not mine to rewrite.
- `docs/reference/cli-parity-check.md:53` still lists `add` as a live `core` verb. It is real and it is stale; I left it because it sits outside the deploy-docs sweep I was asked to make, and it should be swept with the other minors (`refuse()` using `Date.now()` instead of the injected `now()`, the duplicated fingerprint/code-normalise logic).

### Verification

Typecheck clean across all five packages. Lint clean on every file touched. Root, sdk, shared, cli and core suites all green, including the pairing suites. The Panel failures are the pre-existing load flakiness you already diagnosed — they pass in isolation and fail the same way on the pre-merge tip.

Nothing was merged, `main` and the beta train are untouched, and the push was a plain fast-forward.

qcentic-adm (approved):
## F3 gate re-review — approve

Re-reviewed `6431df9` against my three blocking findings, proving each fix from the enforcing line rather than the reply. All three are genuinely resolved; the non-blocking revoke note is tracked and tested; CI on the head is green.

**1 — Deploy docs, blob-paste instructions removed (fixed).**
- `deploy/docker-compose.yml:103-104` now reads *"pair it the same way — `docker compose exec core2 actana pair new` prints its own code and CA fingerprint"* — the dead `docker compose logs core2` step is gone.
- `deploy/README.md:151-153` now points at `actana pair new` and names what Add Core takes (`core2:8443` + the code).
- `deploy/docker-compose.yml:22-24` no longer claims the material "travels in the registration blob"; it now says each client generates its own key and gets a certificate signed on pairing, so no key material crosses.
- `git grep -i blob -- deploy/` returns nothing. Verified.

**2 — ADR recorded in the house shape (fixed).** `docs/adr/0034-short-code-pairing-enrollment.md`, Status: PROPOSED, in the context→decisions→consequences shape with a Considered-options section. D2 records the one pre-auth route and the handshake relaxation scoped to it (citing `core-preauth-gate.ts:104`, `pty-core-link-server.ts:2493,2396`, and the "hole is exactly one route wide" test); D3 the client-born key; D7 `@peculiar/x509` entering the Core bundle as a library parsing pre-auth input while the process holds the CA key; D5 the four-way-errors resolution; D10 the fail-closed lockout; D11 the fourth hand-carried item; D12 the one-definition redeem contract. ADR 0025 D2's allowlist is amended by an appended, dated clause (nothing renumbered), adding `core-pairing-wire` — matching CONTRIBUTING's "append a clause, never renumber."

**3 — Wire mirror pinned by removing the mirror (fixed, ADR 0025 D3).** `packages/sdk/src/core-pairing-wire.ts` is the single, import-free definition of `CORE_PAIRING_REDEEM_PATH` and the redeem request/response/refusal shapes (no imports — the property D2 requires of anything the Core may take). `core-pairing.ts` re-exports them, so no client caller changed. The Core imports them (`core-pairing-routes.ts:44-49`), types its 200 body as `CorePairingRedeemResponse`, and derives its parsed request as `Pick<CorePairingRedeemRequest, "sessionId"|"code"|"csr">` — a rename now fails the build instead of splitting the wire. `client.platform` is documented in the wire type as accepted-and-dropped, not silently implemented. Pinned three ways in `core-pairing.test.ts`: constant *identity* between the two packages (`CORE_PAIRING_REDEEM_PATH === CORE_ROUTE_REDEEM_PATH`) plus two compile-time shape tests. `@actana/core`'s SDK imports are now exactly the three allowlisted modules and nothing else.

**Non-blocking — revoke lost-write (tracked + tested).** `pairing-store.ts` class comment now names `revokeClient` as the one write that fails *open*, states the operator/file divergence, and names the fix (lockfile or append-only revocation log) as a ticket. The added test pins the protection that honestly exists — a second `PairingStore` writing after a revocation carries the stamp forward — without faking a same-millisecond interleave.

**CI / build.** Head `6431df9` CI is green across every required check, including **Unit Tests** (the check that was red on `7d5ad5a` — the earlier red was load-flakiness, not a regression, and it did not recur). Independently: `pnpm install`/`typecheck` clean; sdk 292 passed (the 3 new wire tests), shared 410 passed (the new revoke test), core 865 passed.

**Deferred, and correctly so.** The dev left the acknowledged minors — `docs/reference/cli-parity-check.md:53` still lists `add` as a live `core` verb, `refuse()` using `Date.now()` vs the injected `now()`, and the duplicated fingerprint/code-normalise helpers — for a follow-up sweep, and left the #280 umbrella amendments (fourth item, four-way errors, fail-closed) to the umbrella rather than rewriting it here. Both are the right calls for this gate. One tiny new nit, non-blocking: ADR 0034 is not yet listed in the `docs/README.md#architecture-decisions` index table (which stops at 0033); worth adding on the next touch.

The pairing mechanism was approve-quality on the first pass and the three requested changes are cleanly done. **Approving.**

mrfqcentic (approved):
Moderator gate review — four checks against the enforcing lines, all HOLD.

1. Branch shape: HOLD. 10 commits beyond beta/0.4.0 (merge-base 2dd34a85b8e). Seven single-parent squashes are exactly the merge commits of the approved PRs: a0e73bd8=#291, 63ad949c=#295, d54be1c5=#297, ff561c57=#296, 1f861c9d=#298, 66936ec3=#299, 668ea50e=#302. Two train-absorption merges: 0675130 absorbs F1 (second parent 8aa5b5b "refactor: one actana CLI"), 7d5ad5a absorbs F2 (second parent 2dd34a8 "feat: await a session turn", beta tip). 6431df9 on top is the acknowledged fix commit.

2. Hand-resolved conflict, packages/cli/src/__tests__/in-process-core-session.test.ts (in merge 7d5ad5a): HOLD. Result carries the union — test cases: pairing side 6, beta side 10 (the 6 shared + sends-and-awaits, waits-as-a-verb, refuses-unstamped-delivery, archived-session wait), all 10 present. Assertions: 28 expects on the pairing side, 45 on the beta side, 45 in the result — zero expects from either side missing in the resolution.

3. CI on head 6431df976cc3: HOLD. 23 check runs, all completed — 20 success, 3 skipped (train-mode jobs not applicable on a PR), zero failures. Head carries the fix commit for the other reviewer findings: deploy README + docker-compose repaired, new docs/adr/0034-short-code-pairing-enrollment.md (178 lines), and the redeem contract given one definition in the new packages/sdk/src/core-pairing-wire.ts.

4. Hand-carry removal from #287 (PR #302, 668ea50e): HOLD. The one path 302 deleted — packages/panel/src/components/ui/Textarea.tsx, the paste textarea — does not exist at head. Zero blob/registration files changed between post-302 and head; no `core add` paste verb in the head CLI; nothing resurfaced in either later merge.

Approving.

Issue #303, opened 2026-08-21
Orchestration skills: delete the screen sentinel, split into an invoke-only orchestrator and an eager sub-agent, and settle on a report file


mrfqcentic:
> **This set is #303 (two skills + the file report contract), #304 (multi-file installer + `await.sh`), #305 (the ADR). This is #303.** References below to "the first / second / third issue of this set" mean #303 / #304 / #305 respectively.

> **The `0.4.0` train, sequenced after #280–#289.** Read off the tree at `8aa5b5b`
> (`origin/beta/0.4.0`, *"refactor: one actana CLI"*, merged 2026-08-20) and against
> `main` at `51164f1`. Every file and line reference below was verified on that tree; where
> this issue's brief disagreed with the tree, the tree wins and the correction is stated in
> place.
>
> **Gated on [ADR 0031](https://github.com/actana/control/blob/main/docs/adr/0031-the-product-ships-one-skill.md).**
> 0031 is `PROPOSED`, ratified or rejected by the repository owner at the beta gate, and
> its consequences say *"A second skill is not authorised by this record; it argues against
> D2 on its own."* This issue proposes exactly that second skill. It does not ship until
> 0031 is ratified **and** the new record in the third issue of this set is accepted. If
> 0031 is rejected at the beta gate, both skills become repo-local files under
> `.agents/skills/` and nothing is installed into an operator's home — the work below is
> unchanged apart from the installer never running.

## What this delivers

A Harness Session must be able to act as an **orchestrator**: provision several other
Sessions, across several Harnesses, on its own Core or on remote Cores, give each one work,
wait for all of them, and collect a result from each. The Sessions it provisions act as
**sub-agents** — they do the work and write a report **file**.

**This is skills over the existing CLI.** Hard constraint from the repository owner: **no
new features in `control`.** No new CLI verbs, no Core-side services, no protocol changes.
The single exception is the installer change in the second issue of this set, which widens
an existing mechanism rather than adding a capability. Alternatives that add product
surface have been considered and rejected; they are listed under *Rejected* below and are
not to be reopened in review.

Everything the design needs already exists on `beta/0.4.0` and is verified:

- `actana session start <project> "<prompt>" --harness <id>` — the flag is registered on
  `start`'s accepted-flag table at `packages/cli/src/session-command.ts:190`, validated
  against `KNOWN_HARNESSES` at `:205-211`. **Cross-harness orchestration needs no new
  flag.**
[...]

Issue #305, opened 2026-08-21
A second skill needs its own record: a new ADR depending on 0031, and 0031's falsified motivation clause corrected


mrfqcentic:
> **This set is #303 (two skills + the file report contract), #304 (multi-file installer + `await.sh`), #305 (the ADR). This is #305.** References below to "the first / second / third issue of this set" mean #303 / #304 / #305 respectively.

> **The `0.4.0` train, sequenced after #280–#289.** Read off the tree at `8aa5b5b`
> (`origin/beta/0.4.0`, *"refactor: one actana CLI"*, merged 2026-08-20). Every quotation
> and line reference below was verified on that tree.
>
> **This issue is the gate for the other two in this set.** They do not ship until the
> record below is accepted.

## Why this record has to exist

[ADR 0031 — *The product ships one skill, and installs it into the operator's home*](https://github.com/actana/control/blob/main/docs/adr/0031-the-product-ships-one-skill.md)
is **`PROPOSED`, not accepted**. Its own header says so and says who decides:

> **Status: PROPOSED.** Not accepted. … **The repository owner ratifies or rejects it at the
> beta gate.** Until then nothing here is settled, and a reviewer who thinks the result-bias
> argument in D2 fails is disagreeing with a proposal rather than routing around a decision.

`docs/adr/0031-…:3-12`.

And its consequences close the door this set of issues needs opened:

> **ADR 0006 is amended, not superseded.** Its rule stands for product-feature skills —
> diagram, ship, memory, and anything that changes what a Harness does about the operator's
> own work — and this record carves out exactly one class: a skill whose entire subject is
> driving the product's own CLI. **A second skill is not authorised by this record; it argues
> against D2 on its own.**

`docs/adr/0031-…:371-375`.

**A second skill is exactly what the first issue of this set proposes.** So the D2
result-bias argument must be made again, **on the record**, for `actana-subagent`
specifically. This issue is that record.

## The argument this set expects to win on, stated so it can be rejected instead

**D2's cost is discoverability by Sessions that never touch a Core.** 0031 accepts it
openly for `actana-sessions` (`:144-148`): *"a Session that never touches a Core can still
read this skill's description, and that is a real cost paid for the mandatory-ness D3
chooses."*

**`actana-subagent` pays that cost differently, and less.** Its trigger is not a topic — it
is a **declaration in the prompt** that this Session is acting as a sub-agent of an
orchestrating Session. A Session nobody addressed that way never matches it, whatever it is
[...]

PR #308, opened 2026-08-21
feat: the orchestration-skill installer becomes multi-file, so a skill folder can ship await.sh beside SKILL.md

### LINKED ISSUE #304: The orchestration-skill installer becomes multi-file, so a skill folder can ship await.sh beside SKILL.md
> **This set is #303 (two skills + the file report contract), #304 (multi-file installer + `await.sh`), #305 (the ADR). This is #304.** References below to "the first / second / third issue of this set" mean #303 / #304 / #305 respectively.

> **The `0.4.0` train, sequenced after #280–#289.** Read off the tree at `8aa5b5b`
> (`origin/beta/0.4.0`, *"refactor: one actana CLI"*, merged 2026-08-20). Every file and
> line reference below was verified on that tree; where this issue's brief disagreed with
> the tree, the tree wins and the correction is stated in place.
>
> **Gated on [ADR 0031](https://github.com/actana/control/blob/main/docs/adr/0031-the-product-ships-one-skill.md).**
> 0031 is `PROPOSED` and is ratified or rejected at the beta gate. If it is rejected, the
> installer is not run against an operator's home at all and this change becomes dead
> machinery — so it lands **with** the skill work it exists to carry, not before it.

## Why

So a skill folder can carry more than `SKILL.md`, and so `await.sh` — the operator's
existing, working watcher — ships **inside** the `actana-sessions` folder.

This is the **one exception** to the "no new features in `control`" constraint that governs
this set of issues, and it is an exception because it **widens an existing mechanism rather
than adding a capability**. The installer already writes files into a skill folder. After
this change it writes more than one of them. No new verb, no new service, no new protocol
frame, no new decision for an operator to make.

## What the tree actually does today

**One file per target, hard-coded.** `installOrchestrationSkill` builds exactly one path
per skill directory:

```ts
const file = path.join(homePath(home, target.skillDir), skillName, "SKILL.md");
```

`packages/shared/src/orchestration-skill-install.ts:190`.

> **Correction to the brief.** That line is **:190** on `beta/0.4.0`, not `:185`. The
> `8aa5b5b` commit touched this file (+37/−… lines) and moved it. Cite `:190`.

The write itself is `writeOneCopy(file, content, marker)` at `:197`, and `content` is a
single string on the request type — `SkillInstallRequest.content`, *"The `SKILL.md` bytes,
exactly as they should sit on disk"* (`:87-88`). `SkillInstallEntry.path` is likewise a
single string, documented as *"The `SKILL.md` this harness reads, whether or not it was
written"* (`:73-74`).

**The payload is a single string constant.** `ORCHESTRATION_SKILL_MD` in
`packages/shared/src/orchestration-skill-payload.ts:21` (and its byte-identical twin
`packages/cli/src/orchestration-skill-payload.ts`) — 229 lines each, both generated,
both carrying `ORCHESTRATION_SKILL_NAME = "actana-sessions"` and
`ORCHESTRATION_SKILL_MARKER = "x-actana-managed: true"` beside it.

**The generator writes one source into two targets.** `scripts/gen-skill-payload.mjs:34-37`:

```js
const TARGETS = [
  join("packages", "shared", "src", "orchestration-skill-payload.ts"),
  join("packages", "cli", "src", "orchestration-skill-payload.ts"),
];
```

`SKILL_SOURCE` is `join(".agents", "skills", SKILL_NAME, "SKILL.md")` (`:31`), `renderPayload`
escapes the three sequences that can break a template literal (`:47-51`), and `main()`
refuses to run if the authored file has lost its marker (`:79-81`).

**Four things the drift test holds**
(`packages/shared/src/__tests__/orchestration-skill-fanout.test.ts`):

- every `Harness` has a `skillTarget` with a non-empty `skillDir` and at least one home
  marker (`:33-48`), and each cites a vendor URL and a read date (`:50-62`);
- the CLI's hand-declared table (`packages/cli/src/harness-skill-targets.ts`) agrees row for
  row (`:75-88`);
- **the installer module exists twice and is byte-identical** — `packages/shared/src/…` vs
  `packages/cli/src/orchestration-skill-install.ts` (`:90-100`);
- both embedded payloads are the authored `SKILL.md`, byte for byte (`:102-121`).

> **Correction to the brief's file list.** It named four files plus the test. There is a
> **fifth**: `packages/cli/src/orchestration-skill-install.ts`, a byte-for-byte copy of the
> shared installer, asserted identical at `orchestration-skill-fanout.test.ts:90-100`. Every
> installer change below must be made in **both** copies or CI fails on that assertion.
>
> **Correction to the record.** ADR 0031 D8's premise — *"`@actana/cli` may not import
> `@actana/shared`"* — is **superseded** by ADR 0032 D5 (#288). The block quote at
> `docs/adr/0031-…:294-305` records this, and the installer's own header repeats it
> (`orchestration-skill-install.ts:10-19`): the CLI may import `@actana/shared` now, because
> the package is *inlined* into the published bundle rather than resolved from it. **The
> two-copy arrangement is kept anyway** — the payload is still embedded in two bundles
> because the daemon and the CLI write the skill from two different processes — but it is a
> convenience now, not a workaround. **This issue keeps it.** Collapsing the duplication is
> a separate change with its own argument, and doing it inside a payload-shape change would
> make both unreviewable.

### 1. Payload becomes a map of relative path to content

`SkillInstallRequest.content: string` becomes a map from **folder-relative path** (POSIX
separators, exactly as the generator emits them) to file bytes. `"SKILL.md"` is one entry;
`"await.sh"` is another.

- `installOrchestrationSkill` iterates the map instead of joining `"SKILL.md"` at `:190`.
  The path join becomes `path.join(homePath(home, target.skillDir), skillName, ...relative.split("/"))`,
  matching the existing `homePath` idiom at `:94-96`.
- `writeOneCopy` is unchanged in substance — it already takes `(file, content, marker)` and
  decides per file. It is called once per entry.
- **The dedup at `:185-198` must survive.** Three of the four harnesses read the same global
  root (`~/.agents/skills`), so today the installer writes once per *directory* and reports
  once per *harness* — the header calls this out at `:176-181`. With N files that becomes N
  writes per directory, still not N×3.
- `SkillInstallEntry.path` currently documents a single `SKILL.md`. It must become either
  the folder path or the list of files written, and its doc comment updated. **Pick one and
  say why in the comment** — `actana harness skills [--json]`
  (`packages/cli/src/harness-command.ts:140`, `:199-215`) prints this to an operator
  debugging a missing skill, and "which file, or which folder?" is the question they are
  asking.
- **Outcome semantics per folder.** Today one file yields one outcome. With several files,
  decide and document: `current` only when **every** file matches; `skipped` when the
  marker is missing from **any** managed file; a partial `failed` names which file failed.
  An operator reading `current` must not be reading it because one of two files happened to
  match.

### 2. Generator, both embedded copies, and the drift test follow

- `scripts/gen-skill-payload.mjs` walks the skill folder and emits a map literal rather than
  one string constant. `renderPayload`'s three escapes (`:47-51`) apply per value unchanged.
  The marker guard at `:79-81` must keep applying to **every managed file** — the check that
  refuses to ship an untagged copy is the reason D5's escape hatch is repairable.
- Both payload modules gain the map. Keep `ORCHESTRATION_SKILL_NAME` and
  `ORCHESTRATION_SKILL_MARKER` exported as they are; nothing downstream should have to change
  its import shape for the marker.
- The two ensure wrappers pass the map straight through:
  `packages/cli/src/orchestration-skill.ts:28-34` and
  `packages/core/src/orchestration-skill.ts:43-51`.
- The drift test's payload assertions (`:102-121`) become per-entry: each authored file
  matches its embedded copy, and the two embedded copies match each other. The byte-identity
  assertion on the two installer modules (`:90-100`) is unchanged and must keep passing.

> The first issue of this set adds a **second skill folder** (`actana-subagent`). The
> generator must therefore walk two folders, and the payload must carry two maps — or one
> map keyed by `<folder>/<relative path>`. Choose the shape when both changes are in hand.
> **Neither folder name is a code identifier**, and no `SkillRole` type, enum, or per-role
> installer branch is to be introduced: a folder name is a string the installer is handed.

### 3. `await.sh` ships inside `.agents/skills/actana-sessions/`

**The ADR 0031 D1 marker mechanism works UNCHANGED for a shell script**, and this is the
load-bearing detail that makes the whole change cheap.

The installer matches the marker as a **substring of the file's first bytes**, not through a
YAML parser:

```ts
if (!existing.slice(0, MARKER_SCAN_BYTES).includes(marker)) { … skipped … }
```

`orchestration-skill-install.ts:146`, with `MARKER_SCAN_BYTES = 4096` at `:92`. ADR 0031 D1
chose that deliberately and says why (`docs/adr/0031-…:98-109`): *"the installer reads the
marker as a substring of the file's first bytes rather than through a YAML parser, so
recognising our own writes is a property of our own reader and cannot be taken away by a
loader getting stricter."*

So `# x-actana-managed: true` on **line 2** of `await.sh` — under the shebang — is a valid
marker, is matched by the existing code with no change, and **preserves the operator's D5
escape hatch on the script exactly as it works on the markdown**: delete that line and the
installer never touches the file again.

**The script must be written into the folder with its executable bit, or invoked as
`bash await.sh`.** `writeOneCopy` calls `fs.writeFileSync(file, content, "utf8")` (`:156`)
with no mode, so the file lands `0o666 & ~umask` — **not executable**. Decide which, and say
so in the skill text that tells the orchestrator to run it. Adding a mode to the write is a
one-line change confined to the installer; teaching the skill to say `bash await.sh` costs
nothing and adds no filesystem surface. **Prefer the latter unless there is a reason not
to** — the installer's whole safety argument is that it is *"a filesystem write and nothing
else"* (`:30-32`), and not setting an executable bit keeps that sentence exactly true.

### 4. Why ship the script rather than teach the loop in prose

The loop encodes lessons a model re-deriving it inline will get wrong. Each of these is a
specific failure that has happened:

- **Check the LAST line rather than grepping anywhere.** `ACT-REPORT-END` appearing in the
  body of a report — quoted, in an example, in a diff — settles a Session that has not
  finished. The tail is the proof; a `grep` is not.
- **Save locally before deleting.** Delete-then-save is the screen-era failure with a new
  mechanism.
- **Keep every Session in ONE loop.** A per-Session wait serialises the round: a slow lane
  blocks fast ones that finished minutes ago, and a six-lane round runs at the speed of its
  worst lane summed rather than its worst lane alone.
- **Treat a dropped link as "not yet", not as failure.** `actana core exec` exits **125**
  (`EXIT_LINK_LOST`, `packages/cli/src/exit-codes.ts:45-60`) when the link to the Core went
  away mid-command. Its own doc says why it is *"never the command's own status, and never
  `0`"*: the command kept running on the Core and this side has no result. A watcher that
  read 125 as "the report is not there" would abandon a lane that is fine.

**Prose describing a script is not the script.** On a six-move round the odds of a clean
rewrite every time are poor, and each rewrite is a fresh chance to reintroduce one of the
four. Shipping the file makes the loop a constant instead of a re-derivation.

## Acceptance criteria

1. `SkillInstallRequest` carries a map of folder-relative path to content. Both installer
   copies — `packages/shared/src/orchestration-skill-install.ts` and
   `packages/cli/src/orchestration-skill-install.ts` — are changed identically and
   `orchestration-skill-fanout.test.ts:90-100` still passes.
2. The single-file join at `orchestration-skill-install.ts:190` is gone; the installer writes
   every entry in the map, still once per **directory** and not once per harness.
3. `SkillInstallEntry` reports something an operator can act on for a multi-file folder, and
   its doc comment says which of "the folder" or "the files" it is and why.
4. Multi-file outcome semantics are documented on `SkillInstallOutcome` or beside it:
   `current` requires **all** files current; `skipped` on **any** unmarked managed file;
   `failed` names the file.
5. `scripts/gen-skill-payload.mjs` walks the skill folder(s) and emits the map; the marker
   guard applies to every managed file it emits.
6. Both embedded payload modules carry the map, are byte-identical to each other, and match
   the authored files entry for entry. The drift test asserts all three.
7. `.agents/skills/actana-sessions/await.sh` exists, carries `# x-actana-managed: true`
   within its first 4 096 bytes, and is installed by the installer with no change to the
   marker-matching code.
8. `await.sh` implements the four lessons above, and the skill text names how to invoke it
   (`bash await.sh`, or executable — whichever §3 decided).
9. `node scripts/gen-skill-payload.mjs` is re-run, generated files committed, `shared` suite
   green.
10. Existing installer behaviour is otherwise untouched: presence is still a directory
    question (`:115-117`), a file without the marker is still never written and never deleted
    (`:146-151`), a managed copy is still replaced edits and all (D5), and the ensure still
    cannot change a verb's exit code (D6, `packages/cli/src/orchestration-skill.ts:47`).

## Dependencies

- **First issue of this set** (two skills + the file report contract) — this change exists to
  carry it, and lands with it. `await.sh` has nothing to watch for until the report contract
  exists.
- **Third issue of this set** (the new ADR) — gating, as above.
- **#288 / #293 / #294** — landed at `8aa5b5b`. That commit is what moved the installer and
  the payload into their current shape and current line numbers; **verify against `8aa5b5b`
  or later, not against `main`.**

## Out of scope

- Collapsing the two installer copies or the two payload copies into one shared import. ADR
  0032 D5 now permits it and ADR 0031 D8's note declines it deliberately; changing that is a
  separate record and a separate change.
- Any new CLI verb or flag. `actana harness skills` already exists
  (`packages/cli/src/harness-command.ts:140`) and is the repair path; it needs its output
  adjusted for multiple files, not a sibling verb.

qcentic-dev:
Closes #304.

**Draft, and it stays a draft.** ADR 0031 and ADR 0035 are both `PROPOSED` on
this base — 0035's own consequences say so in its status line and add *"#303 and
#304 do not ship until this record is accepted."* This lands on
`feat/orchestration-skills` beside the record it is gated on, not on the beta
train and not on `main`.

## What changed

**The installer writes a folder, not a file.** `SkillInstallRequest.content:
string` becomes `files: Readonly<Record<string, string>>` — folder-relative
POSIX path to bytes. `installOrchestrationSkill` iterates the map; the
hard-coded `path.join(…, skillName, "SKILL.md")` is gone. `writeOneCopy` is
untouched and is called once per entry, so ADR 0031 D1 and D5 are still decided
per file: the marker authorises each write on its own, and deleting the marker
line from one file keeps that file without stopping the others from updating.

**The dedup moved from the file to the folder.** Three of the four harnesses
read `~/.agents/skills`, so a two-file payload costs two writes per directory
and not six. Still one entry per harness.

**`await.sh` ships inside `.agents/skills/actana-sessions/`.** The marker
mechanism needed no change to carry it — ADR 0031 D1 matches the marker as a
substring of a file's first bytes rather than through a YAML parser, so
`# x-actana-managed: true` on line 2, under the shebang, is matched by the
existing code and preserves D5's escape hatch on the script exactly as it works
on the markdown. It installs **without an executable bit** and is run as `bash
await.sh`, per §3's preference: the installer's safety argument is that it is
"a filesystem write and nothing else", and not setting the bit keeps that
sentence exactly true. A test asserts the mode so the decision cannot lapse
silently.

## Decisions this issue asked to be made and stated

- **`SkillInstallEntry.path` is the folder**, not the list of files, and the doc
  comment says why: it is printed to an operator debugging a missing skill, and
  the next thing they do is `ls` — which takes a directory, one line whether the
  payload is one file or three. The file that went wrong moves into `detail`,
  where "why" already lived. `actana harness skills` names the column
  `SKILL FOLDER` accordingly.
- **Multi-file outcome semantics** are documented on `SkillInstallOutcome` and
  folded worst-first: `failed` on any file and it names which; then `skipped` on
  any unmarked managed file and it names which; then `written`; and `current`
  only when **every** file matched. A payload carrying no files is `failed`,
  never `current`.
- **The payload shape for two folders is deliberately not chosen here.** The
  generator's walker takes a folder name, so the second folder is a second call;
  which of "two maps" or "one map keyed by `<folder>/<path>`" wins belongs to
  the change that brings the second folder, where both halves are in hand. No
  `SkillRole` type, no enum, no per-role branch — a folder name is a string the
  installer is handed.

## Merge order is a hard condition, not a preference

**This branch must not reach the train without #303.** Raised in review and
recorded here rather than left in a comment: today's `SKILL.md` still documents
the `<%ACT_REPORT%>` screen-sentinel workflow and contains no mention of
`await.sh`, `ACT-REPORT-END` or a report file. Shipping this alone would install
a script nothing tells an agent to run, whose default sentinel matches nothing
the current skill emits. That is on top of the ADR gating above, not instead of
it.

## Not in this PR, on purpose

The **skill content** — both `SKILL.md` bodies and the report contract wording,
including the line that tells the orchestrator to run `bash await.sh` — belongs
to #303, which follows. `await.sh` has nothing to watch for until that contract
exists, which is #304's own dependency note. Also out: collapsing the two
installer or payload copies (ADR 0031 D8's note declines it deliberately), and
any new CLI verb or flag.

## Beyond the letter of the issue

Two things this change made necessary rather than optional:

- **A folder-escape check on map keys.** The old code could address nothing but
  `SKILL.md`; a map can, so the constraint that used to be a property of the
  code is now a check — no `..`, absolute path or backslash gets past, and the
  refusal is per file.
- **ADR 0031 D9's genericity assertions now cover every file in the folder**,
  not the markdown alone. A citation of a record this repository keeps is
  exactly as meaningless in `await.sh`'s comments as in the skill's prose, and
  the one that had crept in was removed.

## Review round 1 — three blocking findings fixed (`68bca9b`)

Each was reproduced before the fix and re-run after it:

1. **`await.sh` hung forever on a value-taking flag given last.** `shift 2` with
   one argument left shifts nothing and returns non-zero, so the same branch
   matched forever. All five flags affected. Now a `need_value` check that counts
   unparsed arguments rather than testing `$2` for emptiness — testing the value
   is what hid it. Before: exit 124, no output. After: exit 2 with a message.
2. **Two lanes sharing a report basename silently overwrote each other**, both
   reporting `saved` and, under `--kill`, destroying both Sessions. The Session
   id now prefixes the destination (sanitised), and the residual collision — one
   Session, two folders, one basename — is refused at parse time before any
   Session is touched.
3. **The drift test's folder walker stripped its prefix twice**, collapsing every
   nested file to `""`. Prefix is now built on the way down and sliced once at
   the leaf. A new test walks a nested temporary folder and was verified to fail
   against the old walker.

Both embedded payloads were regenerated and the four lessons re-exercised end to
end afterwards; all still hold.

The seven non-blocking findings are acknowledged on the review thread and not
taken in this round.

## Verification

- `packages/shared` suite green: **427 passed, 27 files**.
- `pnpm -r typecheck` clean across all five packages.
- `pnpm lint`: 0 errors; no warning touches a file this PR changes.
- `node scripts/gen-skill-payload.mjs` re-run, both generated files committed.
- The two installer copies are byte-identical; the assertion at
  `orchestration-skill-fanout.test.ts` that says so still passes.
- Both commits pass `commitlint`; every body line is ≤132 characters.
- **Each commit is green on its own** — the mechanism lands first with a
  one-entry map, the script second.

One **pre-existing** failure on this base is untouched and unrelated:
`packages/cli/src/__tests__/actana-pair.test.ts > actana pair ls > shows pending
codes and paired clients in two separate sections`. It fails identically with
this branch's changes stashed.

https://claude.ai/code/session_01NmPcvbASYeHdxZW3RKP42Q


qcentic-dev:
## Acceptance criteria for #304

Every criterion below was re-verified against **this branch's base**
(`feat/orchestration-skills` @ `7438950`), not against `8aa5b5b`. The tree has
moved since the issue was written; where a citation changed, it is stated.

- [x] **1. `SkillInstallRequest` carries a map of folder-relative path to
      content; both installer copies changed identically and the byte-identity
      assertion still passes.** `content: string` → `files:
      Readonly<Record<string, string>>`. `packages/shared/src/orchestration-skill-install.ts`
      and `packages/cli/src/orchestration-skill-install.ts` are byte-identical
      (verified with `diff`), and the assertion that holds them so — cited in
      the issue as `orchestration-skill-fanout.test.ts:90-100`, and at
      **`:102-112`** on this base — passes unchanged.

- [x] **2. The single-file join is gone; every entry is written, still once per
      directory and not once per harness.** The join the issue cites at
      `orchestration-skill-install.ts:190` (that line number is exactly right on
      this base) is replaced by a folder join plus a per-entry loop in
      `writeOneFolder`. The dedup key moved from the file to the **folder**, so
      a two-file payload is two writes per directory and not six. Asserted by
      *"writes N files per directory and not N per harness"* — it re-runs the
      installer and requires both files' `mtime` to be unchanged.

- [x] **3. `SkillInstallEntry` reports something an operator can act on, and its
      doc comment says which of "the folder" or "the files" it is and why.**
      `path` is **the folder**. The comment argues it from who reads it: it is
      printed by `actana harness skills [--json]` to an operator debugging a
      missing skill, and the next thing they do is `ls`, which takes a directory
      — one line whether the payload is one file or three. A file list would
      answer a question nobody asked and would make the common `absent` row, where
      no file exists to list, the awkward case. Which file went wrong is not
      lost, it moves to `detail`. `harness-command.ts` renames the column to
      `SKILL FOLDER`.

- [x] **4. Multi-file outcome semantics documented on `SkillInstallOutcome`:
      `current` requires all, `skipped` on any, `failed` names the file.**
      Documented as an explicit worst-first fold on the type, and implemented as
      one: `failed` (any file, names which) → `skipped` (any unmarked managed
      file, names which) → `written` → `current` **only when every file
      matched**. A payload carrying no files at all is `failed`, never
      `current`. Four tests, one per clause, plus the empty-payload case.

- [x] **5. `scripts/gen-skill-payload.mjs` walks the skill folder(s) and emits
      the map; the marker guard applies to every managed file it emits.**
      `readSkillFolder(skillName)` walks recursively, sorted, POSIX-separated,
      and takes the folder name as an argument so a second folder is a second
      call. The guard the issue cites at `:79-81` is now a loop over every file,
      verified by hand: an untagged file in the folder makes the generator exit
      non-zero naming that file.

- [x] **6. Both embedded payload modules carry the map, are byte-identical to
      each other, and match the authored files entry for entry — the drift test
      asserts all three.** `ORCHESTRATION_SKILL_FILES` replaces
      `ORCHESTRATION_SKILL_MD`; `ORCHESTRATION_SKILL_NAME` and
      `ORCHESTRATION_SKILL_MARKER` are unchanged, so nothing downstream changed
      its import shape for the marker. The drift test now asserts the key set
      matches the authored folder **in both directions**, each entry matches its
      authored file byte for byte, the two generated modules are identical, and
      every entry carries the marker **within the first 4096 bytes the installer
      actually reads**.

- [x] **7. `.agents/skills/actana-sessions/await.sh` exists, carries
      `# x-actana-managed: true` within its first 4 096 bytes, and installs with
      no change to the marker-matching code.** The marker is on **line 2**,
      under the shebang. Not one byte of `writeOneCopy`'s marker match changed.
      A test pins line 1 as the shebang and line 2 as the marker, and the CLI
      test installs the folder into a temp home and reads `await.sh` back.

- [ ] **8. `await.sh` implements the four lessons** — ✅ — **and the skill text
      names how to invoke it** — ❌ **deferred to #303.**
      - ✅ The script is in and each lesson is asserted, and each was also
        exercised end to end against a stubbed `actana`: a sentinel **quoted in
        the body** does not settle a lane (it reads the tail's last non-blank
        line); the report is copied down and confirmed non-empty **before** any
        `session kill`; two lanes finish in the slow lane's own time rather than
        the sum of both, from **one** loop; and **exit 125** keeps a lane
        waiting and is carried into the final line, rather than being read as
        "no report".
      - ✅ §3's invocation decision is taken and recorded: **no executable bit,
        run as `bash await.sh`**. A test asserts the installed mode has no
        execute bits so the decision cannot lapse silently, and
        `packages/cli/README.md` says it.
      - ❌ **The `SKILL.md` sentence that tells the orchestrator to run it is
        not here.** Both `SKILL.md` bodies and the report contract wording are
        #303's, and #304's own dependency note says `await.sh` *"has nothing to
        watch for until the report contract exists"*. Today's `SKILL.md` still
        carries the **screen sentinel** contract (`<%ACT_REPORT%>` … 
        `<%/ACT_REPORT%>`), not a file whose last line is `ACT-REPORT-END` — so
        a pointer added now would name a contract that document does not yet
        describe. This half lands with #303.

- [x] **9. The generator was re-run, generated files committed, `shared` suite
      green.** `node scripts/gen-skill-payload.mjs` re-run and both generated
      modules committed. **`packages/shared`: 425 passed across 27 files.**
      `pnpm -r typecheck` clean across all five packages; `pnpm lint` 0 errors
      with no warning on any file this PR touches.

- [x] **10. Existing installer behaviour otherwise untouched.** Presence is
      still a directory question (`harnessIsPresent`, the issue's `:115-117`,
      unchanged); a file without the marker is still never written and never
      deleted (`writeOneCopy`, the issue's `:146-151`, unchanged); a managed
      copy is still replaced edits and all (D5); and the ensure still cannot
      change a verb's exit code (D6 — `ensureOrchestrationSkillQuietly`, cited
      in the issue as `orchestration-skill.ts:47` and now at **`:52`** after the
      doc comment above it grew). Every one of the original installer tests
      still passes, migrated only where the API name changed.

### Two things beyond the issue's letter

Both are obligations this change created rather than scope taken on:

- **A folder-escape check on map keys.** The old code could address nothing but
  `SKILL.md`; a map of keys can. The constraint that used to be a property of
  the code is now a check — no `..` segment, absolute path or backslash gets
  past, and the refusal is per file so the honest entries are still written.
- **ADR 0031 D9's genericity assertions now cover every file in the folder.**
  A citation of a record this repository keeps is exactly as meaningless inside
  `await.sh`'s comments as inside the skill's prose; the one that had crept in
  was removed.

### Corrections to the issue's own citations, from reading the current base

- The issue says the payload modules are **229 lines each**. On this base they
  were **278** before this change (`8aa5b5b` is not this base — ADR work from
  #305 landed after it).
- The issue cites ADR 0031 D1's substring passage at `0031-…:98-109`. On this
  base the sentence is at **`:127-131`**; the code comment cites the new lines.
- The issue's claim that the **byte-identity assertion is at `:90-100`** was
  right at `8aa5b5b`; on this base it is `:102-112`, and it still passes.
- **ADR 0031 and ADR 0035 are both `PROPOSED` on this base**, not ratified.
  0035's status line says so explicitly and its consequences add *"#303 and #304
  do not ship until this record is accepted."* Nothing here presumes otherwise:
  the PR is a draft onto `feat/orchestration-skills`, beside the record, and not
  onto the beta train or `main`.

### One pre-existing failure, unrelated

`packages/cli/src/__tests__/actana-pair.test.ts > actana pair ls > shows pending
codes and paired clients in two separate sections` fails on this base with this
branch's changes stashed. Untouched here.

.agents/skills/actana-sessions/await.sh
```diff
@@ -0,0 +1,255 @@
...
+LANE_SESSION=()
+LANE_PROJECT=()
+LANE_REMOTE=()
+LANE_LOCAL=()
+LANE_STATE=()
+LANE_NOTE=()
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --out) OUT_DIR="${2:-}"; shift 2 ;;
```
mrfqcentic: **Blocking — infinite loop on a dangling flag.**

`shift 2` when `$# == 1` shifts nothing and returns non-zero. Nothing checks that return, so `$1` is still `--out`, `while [ $# -gt 0 ]` is still true, and the `case` matches the same branch again forever.

Reproduced:

```
$ timeout 3 bash await.sh --out
$ echo $?
143          # killed by timeout; no output at all
```

All five value-taking flags are affected — `--out`, `--timeout`, `--interval`, `--sentinel`, `--core`. `"${2:-}"` is what hides it: it turns a missing operand into an empty string instead of an error, so the branch looks like it succeeded.

A silent hang is a worse failure mode than a wrong answer here, because the caller is a model composing a command line and the symptom is indistinguishable from "still waiting for a Session".

```sh
--out) [ $# -ge 2 ] || { echo "await.sh: --out needs a value" >&2; exit 2; }
       OUT_DIR="$2"; shift 2 ;;
```

or a small `need_value()` helper called from each branch, since the check is the same five times.

.agents/skills/actana-sessions/await.sh
```diff
@@ -0,0 +1,255 @@
...
+  project="${ref%%:*}"
+  remote="${ref#*:}"
+  if [ "$project" = "$ref" ] || [ -z "$project" ] || [ -z "$remote" ]; then
+    echo "await.sh: \"$ref\" is not <project>:<report-path>" >&2
+    exit 2
+  fi
+  LANE_SESSION+=("$session")
+  LANE_PROJECT+=("$project")
+  LANE_REMOTE+=("$remote")
+  LANE_LOCAL+=("$OUT_DIR/$(basename "$remote")")
```
mrfqcentic: **Blocking — two lanes with the same report basename silently clobber each other.**

`LANE_LOCAL` is `$OUT_DIR/$(basename "$remote")`, so `a=api:reports/report.md` and `b=web:reports/report.md` both resolve to `./report.md`. What happens then:

1. lane A's `project cp` writes `./report.md`, `[ -s ]` passes, state `saved`;
2. lane B's `project cp` overwrites the same file, `[ -s ]` passes, state `saved`;
3. the summary prints **both** lanes as `saved`, naming the same path;
4. with `--kill`, both Sessions are killed — so the report that was overwritten cannot be regenerated.

This is exactly the failure Lesson 2 exists to prevent, arriving by a different route: the bytes were saved and then lost anyway, and the script reports success. It is also specifically the multi-lane case the script was written for — "three sessions, each on a different thing" naturally gives each sub-agent the same report filename inside its own Project.

Either disambiguate the destination:

```sh
LANE_LOCAL+=("$OUT_DIR/${session}-$(basename "$remote")")
```

or refuse duplicate destinations at parse time, which keeps the filenames clean and turns a silent loss into a usage error. Either is fine; silently sharing the path is not.

packages/shared/src/__tests__/orchestration-skill-fanout.test.ts
```diff
@@ -16,20 +16,32 @@
...
   ORCHESTRATION_SKILL_NAME,
 } from "../orchestration-skill-payload";
 
 const REPO = path.resolve(import.meta.dirname, "..", "..", "..", "..");
 const read = (relative: string) => readFileSync(path.join(REPO, relative), "utf8");
 
+const SKILL_DIR = `.agents/skills/${ORCHESTRATION_SKILL_NAME}`;
+
+/** Every authored file in the skill folder, folder-relative and `/`-separated. */
+function authoredFiles(dir = SKILL_DIR): string[] {
```
mrfqcentic: **Blocking — the prefix is stripped twice, so nested files collapse to `""`.**

The recursive branch calls `authoredFiles(`${dir}/${entry.name}`)`, and that inner call has *already* applied `.map(file => file.slice(SKILL_DIR.length + 1))` to its results. The outer call then slices another 30 characters off the already-relative path.

Reproduced against a fixture folder of `SKILL.md`, `await.sh`, `lib/helper.sh`:

```
authoredFiles() => ["","SKILL.md","await.sh"]
```

Downstream that gives a nonsense key-set diff at `:117`, and `:124` calls `read(`${SKILL_DIR}/`)` on the empty entry, which throws `EISDIR` rather than reporting drift.

It does not fire today because the folder is flat — but nested keys are a shape this PR deliberately supports: `writeOneFolder` splits on `/`, and `orchestration-skill-install.test.ts` has a passing test for `lib/helper.sh`. So the drift test is guaranteed to break, confusingly, the first time the payload uses the shape the installer was just taught.

Slice only at the leaf, and let recursion return already-relative paths:

```ts
function authoredFiles(dir = SKILL_DIR, prefix = ""): string[] {
  return readdirSync(path.join(REPO, dir), { withFileTypes: true })
    .flatMap((entry) =>
      entry.isDirectory()
        ? authoredFiles(`${dir}/${entry.name}`, `${prefix}${entry.name}/`)
        : [`${prefix}${entry.name}`],
    )
    .sort();
}
```

packages/shared/src/__tests__/orchestration-skill-install.test.ts
```diff
@@ -193,9 +201,116 @@ describe("failures are reported, never thrown", () => {
...
+    expect(fs.readFileSync(claudeScript(), "utf8")).toBe(SCRIPT);
+  });
+
+  it("writes a nested entry, splitting its key on `/` on any platform", () => {
+    const nested = `# ${MARKER}\nnested\n`;
+    run({ ...FILES, "lib/helper.sh": nested });
+    expect(fs.readFileSync(path.join(claudeDir(), "lib", "helper.sh"), "utf8")).toBe(nested);
+  });
+
+  it("writes N files per directory and not N per harness", () => {
```
mrfqcentic, line 232: **Non-blocking, but this test does not test its own name — and its name is AC2.**

It installs, records mtimes, installs again, and asserts `current` with unchanged mtimes. That outcome comes entirely from `writeOneCopy`'s `existing === content` comparison and holds whether or not the folder dedup exists: delete `presentByFolder` and call `writeOneFolder` once per harness, and this test stays green, because the second pass still writes nothing.

The property AC2 actually asks for — *N writes per directory, not N×3* — is about how many writes happen inside **one** run, and nothing here counts them. The dedup is correctly implemented (I checked `folderFor` / `presentByFolder` / `doneByFolder`), so this is an enforcement gap rather than a defect; but it is the one arithmetic the PR body calls "the arithmetic the folder key holds down as the payload grows", which makes it worth holding.

A spy on `fs.writeFileSync` counting calls in a single `run()` with `.codex` and `.cursor` both present would pin it: expect 2, not 4.

packages/shared/src/__tests__/orchestration-skill-fanout.test.ts
```diff
@@ -131,17 +206,23 @@ describe("the skill is generic and self-contained (ADR 0031 D9)", () => {
...
-      "actana/control",
-      "release train",
-      "beta/",
-      "promote.yml",
-      "gh pr",
-      "CONTRIBUTING",
-      "ADR 00",
-    ]) {
-      expect(skill.includes(forbidden), `the skill mentions "${forbidden}"`).toBe(false);
+  it("names no project, ticket workflow, train or repository — in any file", () => {
```
mrfqcentic, line 251: **Non-blocking — the D9 extension stops one test short of where ADR 0035 leans on it.**

Extending this one to every file in `ORCHESTRATION_SKILL_FILES` is exactly right, and the reasoning in the comment is the correct one. But two assertions in the same `describe` still read only the markdown: `"has no repo-relative link and assumes no checkout"` (`:201`) and `"names no default Harness and no preference order"` (`:229`), both closing over `const skill = read(.../SKILL.md)`.

The second one matters beyond tidiness. ADR 0035 D3 — on this base — justifies the prohibition being stated harness-neutrally *because* this test asserts it: *"…asserts that the shipped text names none of the four harness ids — and that assertion is to be extended to cover the new file rather than exempting it."* `await.sh` is a new file in the shipped folder, and it is currently exempt.

No live violation: I checked, and `await.sh` contains none of the four harness ids, no `docs/*.md` reference and no repo-relative link. So this is a one-line change from `skill` to the same `Object.entries(...)` loop, and it makes the record's own justification true.

packages/shared/src/__tests__/orchestration-skill-fanout.test.ts
```diff
@@ -115,8 +138,60 @@ describe("the embedded payload is the authored file (ADR 0031 D8)", () => {
...
+
+  it("treats a lost link as 'not yet' rather than as a missing report", () => {
+    // `actana core exec` exits 125 when the link went away mid-command: the
+    // command kept running on the Core and this side has no result. Reading it
+    // as failure abandons a lane that is fine.
+    expect(script).toContain("125");
+    expect(script.toLowerCase()).toContain("retrying, not giving up");
+  });
+
+  it("saves the report to local disk before killing anything", () => {
```
mrfqcentic, line 225: **Non-blocking — these assert where substrings sit in the file, not what the script does.**

`script.indexOf("project cp") < script.indexOf("session kill")` is a claim about source-text order. Move the kill into a helper defined above the copy, or reorder the runtime calls while leaving the text alone, and it stays green while delete-then-save is live. `expect(script).toContain("bash await.sh")` at `:194` is the same shape — it passes on the script's own header comment, so the script asserts its own documentation.

I am not asking for a rewrite: pinning intent in a text assertion is a reasonable thing to do for a shipped artefact, and the surrounding ones (`last_line_is_sentinel`, `tail -n \d+`, no `grep -q`, `125`) are genuinely useful tripwires. Worth a comment saying these are tripwires against accidental removal rather than behavioural proofs, so a later reader does not over-trust them.

The mode assertion in `harness-skills.test.ts` is the counter-example and the good pattern — it tests the actual installed artefact.

.agents/skills/actana-sessions/await.sh
```diff
@@ -0,0 +1,255 @@
...
+pending="${#LANE_SESSION[@]}"
+
+echo "await.sh: watching $pending lane(s) for a last line of \"$SENTINEL\"" >&2
+
+# Lesson 3. ONE loop, every unfinished lane per tick. Nothing here blocks on a
+# single Session, which is what keeps a six-lane round running at the speed of
+# its slowest lane rather than the sum of all six.
+while [ "$pending" -gt 0 ]; do
+  now="$(date +%s)"
+  if [ $((now - started)) -ge "$TIMEOUT" ]; then
```
mrfqcentic, line 224: **Non-blocking — `--timeout` and `--interval` are never validated, and both fail silently.**

`--timeout abc`: `[ $((now - started)) -ge "abc" ]` prints `[: abc: integer expression expected` to stderr and returns 2, so the branch is not taken and **the round never times out** — the loop runs until something kills it. Verified.

`--interval abc`: `sleep abc` fails immediately (`invalid time interval`), nothing checks its status, and the 15-second poll becomes an unthrottled loop hammering `actana core exec` once per lane per iteration. Verified.

Both are one `case` pattern away:

```sh
case "$TIMEOUT$INTERVAL" in *[!0-9]*) echo "await.sh: --timeout and --interval take whole seconds" >&2; exit 2 ;; esac
```

The timeout one is the more serious of the two, since "never times out" is the failure the `--timeout` flag exists to prevent.

.agents/skills/actana-sessions/await.sh
```diff
@@ -0,0 +1,255 @@
...
+        const want = process.argv[1];
+        const hit = rows.find((row) => row.name === want || row.projectId === want);
+        if (!hit || !hit.path) process.exit(1);
+        process.stdout.write(hit.path);
+      });
+    ' "$1"
+}
+
+ROOTS=()
+for index in "${!LANE_PROJECT[@]}"; do
```
mrfqcentic, line 189: **Non-blocking — the comment above promises a cache the loop does not have.**

`:128-131` says the Project's path is read *"once per Project rather than per tick"*, and the per-tick half is true and worth having. But this loop runs over `LANE_PROJECT` indices, so it calls `project_root` once per **lane**: six lanes on one Project run `actana project ls --json` six times and spawn six `node` processes at startup.

Harmless in practice — it is startup cost, not per-tick cost — but the comment states a property the code does not have, and this file is explicitly written to be read. Either soften it to "once at startup rather than per tick", or key a small associative array on the project name and make the sentence true.

scripts/gen-skill-payload.mjs
```diff
@@ -65,24 +106,50 @@ export const ORCHESTRATION_SKILL_NAME = ${JSON.stringify(SKILL_NAME)};
...
+  const files = readSkillFolder(SKILL_NAME);
+  if (files.length === 0) {
+    throw new Error(`${SKILL_SOURCE} holds no files — there is nothing to install`);
+  }
+  // The guard is per file, and it is the reason ADR 0031 D5's escape hatch is
+  // repairable: an untagged copy on an operator's disk is one the installer can
+  // never write again, so shipping one is shipping a file we have given away.
+  // Every file this script emits is a managed file, so every file is checked.
+  for (const file of files) {
+    if (file.content.includes(MARKER)) continue;
```
mrfqcentic, line 144: **Non-blocking — the guard's window is wider than the reader's.**

`file.content.includes(MARKER)` scans the whole file; the installer only looks at the first `MARKER_SCAN_BYTES = 4096`. So the generator would happily emit a file whose marker sits at byte 5000 — a file the installer writes once and then reports `skipped` forever, which is D5's escape hatch fired by accident on our side, for everybody.

The drift test **does** close this (`fanout.test.ts:141-152` asserts the marker within `content.slice(0, 4096)` per entry), so nothing can ship. That makes this a message-quality issue rather than a hole: the failure surfaces as a test failure some distance from the file that caused it, when the generator could have refused with the filename in hand. `file.content.slice(0, 4096).includes(MARKER)` here would put the two windows in agreement.

Separately, `readSkillFolder` walks the real directory with no filter, so a stray `.DS_Store` or `.SKILL.md.swp` in the skill folder makes the generator throw "carries no `x-actana-managed: true`" and independently makes the drift test's key-set assertion fail on a file that is not in git. Skipping dotfiles would remove a confusing local-only failure.

packages/shared/src/orchestration-skill-install.ts
```diff
@@ -164,6 +226,77 @@ function writeOneCopy(
...
+ * Entries are sorted by path so a report is stable run to run. The generator
+ * already emits them sorted; a caller that builds the map by hand should not
+ * have to know that.
+ */
+function writeOneFolder(
+  folder: string,
+  files: Readonly<Record<string, string>>,
+  marker: string,
+): { outcome: SkillInstallOutcome; detail?: string } {
+  const results = Object.keys(files)
```
mrfqcentic, line 266: **Non-blocking — a new consequence of folders that is worth writing down here.**

The loop iterates the keys in `request.files`, which is right. The consequence is that nothing ever removes a file the payload used to carry: when a future release renames or drops `await.sh`, every already-installed folder keeps the old copy indefinitely, and it is a *managed* file — marker and all — that the harness will still load.

With a single-file payload this could not happen; the folder is what makes it possible, so it belongs in this change's reasoning rather than being discovered later. I am not asking for a sweep to be implemented — deleting files out of an operator's home is a decision with its own argument, and probably ADR-shaped given how carefully D5 draws the line at *never deleted*. But the header comment above already enumerates what the installer does and does not do per file, and "a file that leaves the payload is left behind" is the one line missing from it.

If a sweep ever is wanted, the marker is what makes it safe: an unmarked file is the operator's and must stay either way.

mrfqcentic (changes requested):
## Review against #304 — request changes

Reviewed `git diff 7438950 fa9caaa` (13 files) against issue #304, and against ADR 0035, which is on this base at `7438950` carrying the corrected citations.

**The installer change itself is right, and I verified it from the code rather than from the description.** `installOrchestrationSkill` now keys `folderFor` / `presentByFolder` / `doneByFolder` on the folder instead of the `SKILL.md` path, so the per-directory dedup survives the shape change intact — the partition is identical because `skillName` is constant across targets. `writeOneCopy` is byte-for-byte untouched, which is what keeps every clause of D1 and D5 decided per file. The outcome fold matches its own documentation exactly, `failed` > `skipped` > `written` > `current`, with the empty map falling to `failed` rather than `current`. `isInsideFolder` genuinely rejects `""`, `/etc/passwd`, `../x`, `a/../../x` and backslashes — an absolute path fails on its empty first segment. Both installer copies are byte-identical at `fa9caaa` (`diff` clean), both payload modules are byte-identical to each other, and I re-derived the embedded payload independently: `ORCHESTRATION_SKILL_FILES` carries exactly `["SKILL.md", "await.sh"]`, 14254 and 9893 bytes, both matching the authored files byte for byte. `ORCHESTRATION_SKILL_MD` has no remaining references anywhere in the tree.

**The boundary is respected.** `.agents/skills/actana-sessions/SKILL.md` is not in the diff. The only file added under `.agents/skills/` is `await.sh`. Nothing rewrites skill bodies, which is #303's.

**The decisions #304 asked to be made are made and argued**, not just made: the folder-over-file-list choice for `SkillInstallEntry.path` is reasoned from who reads it, the outcome ordering is written out rather than reduced to a rank comparison, and the executable-bit decision is pinned by a mode assertion so it cannot lapse silently. The path-escape check is a genuine addition — the old code could address nothing but `"SKILL.md"`, so a constraint that used to be structural had to become a check.

**What blocks it is the shipped script and one test helper.** Three defects, all verified by execution or by reproduction rather than by reading:

1. **`await.sh` hangs forever on a value-taking flag given last** (`:86-90`). `shift 2` with `$# == 1` shifts nothing and returns non-zero; `$1` is still the flag, and `while [ $# -gt 0 ]` spins. Reproduced: `timeout 3 bash await.sh --out` produces no output and has to be killed. Affects `--out`, `--timeout`, `--interval`, `--sentinel`, `--core`. A silent unkillable-looking hang is a worse failure than an error message, and the caller here is a model composing a command line.
2. **Two lanes sharing a report basename silently overwrite each other** (`:121`). Both lanes report `saved` naming the same file, one report is gone, and under `--kill` both Sessions are killed so it cannot be regenerated. This is the multi-lane case the script exists for, and it defeats Lesson 2 by a different route than the one Lesson 2 guards.
3. **`authoredFiles()` double-strips the prefix, so any nested file becomes `""`** (`orchestration-skill-fanout.test.ts:36`). Reproduced against a fixture: a folder of `SKILL.md`, `await.sh`, `lib/helper.sh` yields `["", "SKILL.md", "await.sh"]`. Nested keys are a shape this PR deliberately supports and tests (`orchestration-skill-install.test.ts` covers `lib/helper.sh`), so the drift test breaks — confusingly — the first time anyone uses it.

None of the three is hard to fix and none touches the installer design. Details and suggested fixes inline.

### Acceptance criteria

| # | Criterion | Verdict |
| --- | --- | --- |
| 1 | `SkillInstallRequest` carries a map; both installer copies changed identically; byte-identity assertion still passes | **Met** — `diff` clean at `fa9caaa`; assertion intact at `fanout.test.ts:102-112` |
| 2 | Single-file join gone; writes every entry, once per **directory** not per harness | **Met** in the code — dedup verified on `presentByFolder`/`doneByFolder`. But the test named for it does not test it (inline) |
| 3 | `SkillInstallEntry` reports something actionable; doc comment says which and why | **Met** — the folder, with the reasoning from who reads it; `SKILL FOLDER` column follows |
| 4 | Multi-file outcome semantics documented; `current` needs all, `skipped` on any, `failed` names the file | **Met** — documented on `SkillInstallOutcome`, fold matches, each clause has a test |
| 5 | Generator walks the folder(s) and emits the map; marker guard applies to every managed file | **Met** — per-file guard at `:143-147`; one window mismatch noted inline |
| 6 | Both payloads carry the map, are byte-identical, match authored entry for entry; drift test asserts all three | **Met** — independently re-derived; key set asserted in both directions |
| 7 | `await.sh` exists, marker within the first 4096 bytes, installed with no marker-matching change | **Met** — marker at byte 22, line 2 under the shebang; `writeOneCopy` untouched |
| 8 | `await.sh` implements the four lessons, and the skill text names how to invoke it | **Partially met** — four lessons all present; skill text does not name it (see below) |
| 9 | Generator re-run, generated files committed, `shared` suite green | **Met** on the artefacts — payload matches authored bytes exactly. Suite not executed here (reviewer write restriction); nothing static contradicts the claim |
| 10 | Existing behaviour otherwise untouched | **Met** — `writeOneCopy`, `harnessIsPresent` and `ensureOrchestrationSkillQuietly` are all unchanged |

**On AC8.** The four lessons are genuinely implemented and not merely gestured at: the tail-and-last-non-blank-line check rather than a grep, save-and-verify-non-empty before any kill, one loop over all unfinished lanes per tick, and `125` treated as "ask again". The invocation decision is made, argued in three places, and enforced by a mode assertion. What is missing is the other half of the sentence: `SKILL.md` contains zero occurrences of `await.sh`, `ACT-REPORT-END`, `project cp` or "report file", and still documents the `<%ACT_REPORT%>` screen-sentinel workflow #303 exists to delete. So the folder ships a script nothing tells an agent to run, whose default sentinel matches nothing the current skill emits.

Deferring the wording is right — it is #303's, and #304's own dependency note says `await.sh` has nothing to watch for until the report contract exists. I am not asking for it here. But it makes the merge order a hard condition rather than a preference: **this branch must not reach the train without #303**, or the release installs an inert file into operators' homes. Worth saying in the PR body beside the ADR gating that is already there.

### Non-blocking, inline

The AC2 enforcement gap; the D9 genericity extension stopping one test short of where ADR 0035 D3 leans on it; the generator's marker guard scanning the whole file where the installer scans 4096 bytes; `--timeout`/`--interval` accepting non-numeric values; `project_root` running once per lane under a comment saying once per Project; and the new consequence that nothing sweeps a file dropped from a future payload.

**Verdict: request changes.** Three defects, all in the shipped script and one test helper. The installer, the payload, the generator and the outcome semantics are sound and I would approve them as they stand.

.agents/skills/actana-sessions/await.sh
qcentic-dev: All three blocking findings are fixed in `68bca9b`. Each was reproduced first,
fixed, then re-run — the reproductions below are yours, re-executed here.

### 1. Infinite loop on a dangling value-taking flag — fixed

**Before**, all five flags:

```
--out       -> exit 124, no output
--timeout   -> exit 124, no output
--interval  -> exit 124, no output
--sentinel  -> exit 124, no output
--core      -> exit 124, no output
```

**After**, all five:

```
--out       -> exit 2, "await.sh: --out needs a value"
--timeout   -> exit 2, "await.sh: --timeout needs a value"
--interval  -> exit 2, "await.sh: --interval needs a value"
--sentinel  -> exit 2, "await.sh: --sentinel needs a value"
--core      -> exit 2, "await.sh: --core needs a value"
```

Took the `need_value()` helper you suggested, since the check is the same five
times. It **counts the arguments still unparsed** rather than testing `$2` for
emptiness — testing the value is precisely what hid this, so the fix deliberately
does not go near `${2:-}`. Those defaults are gone from all five branches; `$2`
is now read directly, which makes the guarantee visible at the point of use.
`--out /tmp/x` still parses, and reaches the "name at least one lane" usage error
as before.

### 2. Two lanes sharing a report basename — fixed

Your exact scenario, `a=api:reports/report.md` and `b=web:reports/report.md`.

**Before:**

```
aaaa  saved  .../out/report.md (Session killed)
bbbb  saved  .../out/report.md (Session killed)
rc=0
files saved: report.md
  report.md -> THIS IS LANE B - the web report      <- lane A's report is gone
sessions killed: aaaa bbbb
```

**After:**

```
aaaa  saved  .../out/aaaa-report.md (Session killed)
bbbb  saved  .../out/bbbb-report.md (Session killed)
rc=0
files saved: aaaa-report.md bbbb-report.md
  aaaa-report.md -> THIS IS LANE A - the api report
  bbbb-report.md -> THIS IS LANE B - the web report
```

Took **both** of your options rather than choosing between them, because
prefixing alone is not actually a guarantee: one Session writing two reports from
different folders — `aaaa=api:x/report.md aaaa=api:y/report.md` — still lands on
one name. So the id prefixes the destination, and the residual collision is
refused at parse time, before a single Session is touched:

```
await.sh: aaaa and aaaa would both be saved as
  .../out/aaaa-report.md
Give them different report filenames, or a --out directory each.
exit 2
```

The id is sanitised to `[A-Za-z0-9._-]` on the way into the filename, so an
opaque id cannot steer the write out of `--out`. Four lanes with distinct
destinations still run with no false refusal.

### 3. `authoredFiles()` double-strips the prefix — fixed

Your fixture of `SKILL.md`, `await.sh`, `lib/helper.sh`:

```
before: ["", "SKILL.md", "await.sh"]
after:  ["SKILL.md", "await.sh", "lib/helper.sh"]
```

Took your suggested shape — prefix built on the way down, slice at the leaf, so
it happens exactly once. One addition: the walker takes a `root` so it can be
pointed at a fixture, and there is now a test that walks a nested temporary
folder and asserts all three paths. **I verified that test fails against the old
walker**, so the flat folder cannot hide this a second time. Second, cheaper
assertion beside it: no key from the real folder is empty or absolute.

### Also in this commit

Both embedded payloads regenerated — `await.sh`'s bytes changed — and the four
lessons re-exercised end to end afterwards. All still hold: a sentinel quoted in
the body does not settle a lane, the report is saved and confirmed non-empty
before any kill, two lanes finish in the slow lane's own time from one loop, and
exit 125 keeps a lane waiting and carries that into the final line.

`packages/shared` **427 passed / 27 files**, `packages/core` **865 passed**,
`pnpm -r typecheck` clean, `pnpm lint` 0 errors with no warning on a file this PR
touches. The one `packages/cli` failure is the pre-existing, unrelated
`actana-pair.test.ts > actana pair ls` case that also fails on the untouched
base.

### Not addressed here, deliberately

The seven non-blocking findings are all fair and I have not touched them in this
round, which was scoped to the blockers. Flagging where I think each should land
rather than leaving them silent:

- **AC2 has no enforcement** (`writeFileSync` spy counting 2 not 4) — the
  strongest of the seven, since it is the one arithmetic the PR body names. Worth
  doing on this branch.
- **D9 extension stops two assertions short**, and ADR 0035 D3 leans on the
  harness-id one covering the new file — also worth doing on this branch, since
  it is what makes the record's justification true.
- **Text-order assertions are tripwires, not proofs** — agreed; a comment saying
  so is the right size of fix.
- **`--timeout`/`--interval` accept non-numeric values**, and "never times out"
  is the failure `--timeout` exists to prevent.
- **Generator's marker window is wider than the reader's**, plus dotfile
  skipping — message quality, since the drift test already closes the hole.
- **`project_root` runs once per lane under a comment saying once per Project** —
  the comment is what is wrong; softening it is the honest minimum.
- **Nothing sweeps a file dropped from a future payload** — agreed this belongs
  in the installer header's own enumeration rather than being discovered later,
  and agreed a sweep itself is ADR-shaped given where D5 draws the line at *never
  deleted*.

Say the word on which of these you want in this PR and I will take them in the
next round.

### On merge order

Your AC8 point is right and I have made it explicit in the PR body rather than
leaving it in a comment: this branch must not reach the train without #303, or
the release installs a script nothing tells an agent to run, whose default
sentinel matches nothing the current skill emits.

Still a draft, still based on `feat/orchestration-skills`, nothing merged.

.agents/skills/actana-sessions/await.sh
qcentic-dev: Fixed in `68bca9b`. Reproduced your scenario first — both lanes `saved` naming one file, lane A's bytes gone, both Sessions killed, exit 0 — then fixed and re-ran: `aaaa-report.md` and `bbbb-report.md`, each with its own lane's content.

Took **both** of your options rather than one, because prefixing alone is not a guarantee: `aaaa=api:x/report.md aaaa=api:y/report.md` still collides. So the Session id prefixes the destination (sanitised to `[A-Za-z0-9._-]`, so an opaque id cannot steer the write out of `--out`), and the residual duplicate is refused at parse time before any Session is touched. Full before/after in the [thread reply](https://github.com/actana/control/pull/308#discussion_r3831454270).

packages/shared/src/__tests__/orchestration-skill-fanout.test.ts
qcentic-dev: Fixed in `68bca9b`, using your suggested shape — prefix built on the way down, slice at the leaf, once.

Reproduced your fixture first: `["", "SKILL.md", "await.sh"]`. After: `["SKILL.md", "await.sh", "lib/helper.sh"]`.

One addition beyond the suggestion: the walker now takes a `root`, and a test walks a nested temporary folder and asserts all three paths. **I verified that test fails against the old walker** — so the flat folder cannot hide this a second time, which was your actual point. Full before/after in the [thread reply](https://github.com/actana/control/pull/308#discussion_r3831454270).

mrfqcentic (approved):
## Re-review of `68bca9b` — approved

All three blocking findings are fixed. I proved each by re-running my round-1 reproduction against both trees rather than by reading the reply, and in two cases the reproduction is stronger than the one I ran the first time.

### 1. Dangling value-taking flag — fixed

Both scripts extracted from git, run under `timeout 3`:

```
OLD (fa9caaa)                    NEW (68bca9b)
--out       rc=124  (no output)  --out       rc=2  await.sh: --out needs a value
--timeout   rc=124  (no output)  --timeout   rc=2  await.sh: --timeout needs a value
--interval  rc=124  (no output)  --interval  rc=2  await.sh: --interval needs a value
--sentinel  rc=124  (no output)  --sentinel  rc=2  await.sh: --sentinel needs a value
--core      rc=124  (no output)  --core      rc=2  await.sh: --core needs a value
```

`rc=124` is `timeout` killing the hang; `rc=2` is the usage exit. All five, both directions.

Counting the unparsed arguments instead of testing `$2` is the right instinct, and dropping `${2:-}` from all five branches matters as much as the guard: the default was what turned a missing operand into a plausible-looking empty value. Regression checked — `--core mycore --sentinel DONE-MARKER --out … --interval 1 --timeout 6` still parses, the stub CLI receives `--core mycore`, and a custom sentinel settles the lane.

### 2. Basename collision — fixed

Reproduced end to end against a stub `actana` implementing `project ls --json`, `core exec -- tail`, `project cp` and `session kill`, with two Projects each holding `reports/report.md`. My round-1 scenario verbatim, with `--kill`:

```
OLD (fa9caaa)
  aaaa  saved  …/out/report.md (Session killed)
  bbbb  saved  …/out/report.md (Session killed)
  rc=0    files: report.md
    report.md -> THIS IS LANE B - the web report     <- lane A's report is gone, both Sessions killed

NEW (68bca9b)
  aaaa  saved  …/out/aaaa-report.md (Session killed)
  bbbb  saved  …/out/bbbb-report.md (Session killed)
  rc=0    files: aaaa-report.md  bbbb-report.md
    aaaa-report.md -> THIS IS LANE A - the api report
    bbbb-report.md -> THIS IS LANE B - the web report
```

Taking **both** options rather than picking one is the better call and I had not thought it through far enough — you are right that prefixing alone is not a guarantee. Verified the residual case (`aaaa=api:x/report.md aaaa=api:y/report.md`):

```
OLD:  both lanes "saved", one file on disk, one report silently lost
NEW:  await.sh: aaaa and aaaa would both be saved as
        …/out/aaaa-report.md
      Give them different report filenames, or a --out directory each.
      rc=2, nothing written, no Session touched
```

Checked for false refusals: four lanes with distinct destinations run clean and produce four files. Checked the sanitiser with a hostile id — `../../escape=api:x/report.md` lands as `.._.._escape-report.md` **inside** `--out`, and nothing appears above it.

### 3. `authoredFiles()` double-strip — fixed

Both walkers lifted verbatim from their commits and run against my round-1 fixture (`SKILL.md`, `await.sh`, `lib/helper.sh`):

```
OLD (fa9caaa): ["","SKILL.md","await.sh"]                 FAIL
NEW (68bca9b): ["SKILL.md","await.sh","lib/helper.sh"]    PASS
```

That output is also the independent confirmation that the new regression test would have failed against the old walker — it is the same fixture shape, and the old walker returns the empty key for it. Building the prefix on the way down means the slice happens exactly once, and threading `root` so the walker can be pointed at a temp folder is what makes the test possible at all. The second assertion — no key from the real folder empty or absolute — is a good cheap belt.

### Nothing regressed

- Both installer copies byte-identical at `68bca9b`; both payload modules byte-identical.
- Payload regenerated correctly: keys `["SKILL.md","await.sh"]`, 14254 and **12138** bytes (up from 9893 — the script grew), both matching the authored files byte for byte, re-derived independently.
- `await.sh` still lands `100644`, marker still at byte offset **22** on line 2 under the shebang.
- `SKILL.md` still untouched by this commit — the #303 boundary holds.
- Every fanout assertion re-checked by hand against the new script text: `last_line_is_sentinel` present, `tail -n N` present, no `grep -q`, `125` present, `project cp` before `session kill`, `bash await.sh` present, none of the four harness ids, none of the seven forbidden D9 strings.
- Lesson 4 re-verified end to end with a stub that always exits 125: the lane keeps waiting rather than being abandoned, and the note is carried into the timeout line — `timeout … last: the link to the Core dropped mid-poll — retrying, not giving up`. Lesson 1 was exercised incidentally in the collision run: lane A's report carries a quoted `ACT-REPORT-END` on its second line, inside the `tail -n 3` window, and the lane still settled on the real last line.

### CI

**Red on the head, and I verified why rather than taking the note for it.** 22 of 23 checks pass — Typecheck, Lint, Conventions, Train rules, Promotion gate, both installer E2Es, both tarball smokes, both image builds. `Unit Tests` fails, and the whole of the failure is:

```
FAIL  src/__tests__/actana-pair.test.ts > actana pair ls > shows pending codes and paired clients in two separate sections
      packages/cli   1 failed | 49 passed | 1 skipped (51)
```

`root`, `sdk`, `shared` (27 files), `core` (56 files) and `panel` (144 files) all pass, so every test this PR adds or changes is green.

Confirmed pre-existing rather than assumed: the identical failure — same file, same test name, same `1 failed | 49 passed | 1 skipped (51)` counts — is on `5656a02`, upstream of this PR's base. `actana-pair.test.ts` was last modified by `5656a02` itself and is not in this PR's diff.

So it is not yours, and I am not holding the PR for it. But it is red on a branch heading for the `0.4.0` train and I could not find an issue tracking it — worth filing one, since "Unit Tests is red for a known unrelated reason" stops being checkable the moment a second reason appears.

### Still open — none a condition of this approval

The seven non-blocking findings from round 1 stand, and scoping this round to the blockers was the right choice. Your triage matches mine, including which two are worth doing on this branch: the AC2 write-count assertion, because it is the one arithmetic the PR body names and it is currently unenforced; and the D9 harness-id assertion, because ADR 0035 D3 justifies the harness-neutral wording *by* that assertion covering the new file, and `await.sh` is still exempt from it. Neither is a defect today — I re-checked, `await.sh` names no harness — and both are small.

And the condition from round 1 is unchanged and unaffected by these fixes: `SKILL.md` still contains no reference to `await.sh`, `ACT-REPORT-END` or `project cp`, and still documents the `<%ACT_REPORT%>` screen workflow. **This branch must not reach the train without #303**, or the release installs a script into operators' homes that nothing invokes and whose default sentinel matches nothing the skill emits.

---

**Approved.** Three defects, three reproductions, three fixes that each go a step past what I asked — the second collision case, the sanitiser, and a regression test that fails against the old walker. Still a draft, and both ADR 0031 and ADR 0035 remain `PROPOSED`; this approval is on delivery against #304 and moves neither.

PR #309, opened 2026-08-21
feat(skills): two skills and a report file — delete the screen sentinel, split the orchestrator from the sub-agent

qcentic-dev:
## Summary

Implements **#303**: the two skill bodies and the file report contract.

The `<%ACT_REPORT%>` screen sentinel is **deleted, not patched** — a rendered
screen is a terminal emulator's picture of a bounded replay ring, and no rule
recovers a tag whose characters interleaved, a ring that dropped its oldest
lines, or a scrollback a harness discarded outright. In its place the product's
orchestration skill splits in two, along the one axis that cannot be expressed
in one `description` field:

- **`actana-sessions` stays invoke-only** and is rewritten around the file
  contract: it mints a report path per lane, states it in full in every prompt,
  polls `actana core exec -- cat`, checks the **last** line for `ACT-REPORT-END`,
  **saves locally, waits ~20 s, then deletes the remote file**, and ships
  `await.sh` (unchanged, from #304) as the loop that does that for a whole round.
- **`actana-subagent` is new and eager.** It triggers on a prompt declaring the
  Session a sub-agent of an orchestrating Session, and covers exactly two things:
  how to write the report file, and the prohibition on `actana session start`.
  It says nothing about how to do the work, because it is eager and the skills an
  orchestrator passes alongside it were invoked — advice from here would fight
  them and win.

The asymmetry is stated in **both** files, so a later reader does not tidy it
into symmetry.

Both skills are authored under `.agents/skills/`, regenerated into both embedded
payloads, and written by both installer copies. This is where #304's generator
left the two-folder payload shape open — *"that choice belongs to the change that
brings the second folder"* — and it is settled as **one map keyed by folder
name**. A folder name stays data: nothing branches on which one it is holding.
No new CLI verb, no flag, no Core service, no protocol frame; the multi-file
installer from #304 is used as shipped and not reworked.

## Base

**Base is `feat/orchestration-skills`, not the open train and not `main`** — this
stacks on #304's installer and #305's ADR 0035, both of which are on that branch.
The `Train rules` check is expected to object to that base; retarget when the
stack lands, not here. **Draft on purpose**, and gated: neither skill installs
into an operator's home until ADR 0031 is ratified and ADR 0035 is accepted at the
beta gate. Both records read `PROPOSED` on this base today. If 0031 is rejected,
both skills remain repo-local files under `.agents/skills/` and the work above is
unchanged apart from the installer never running.

## Related issues

Refs #303 — implemented here.
Refs #304 — the multi-file installer and `await.sh` this builds on, already landed.
Refs #305 — ADR 0035 and 0031's repaired motivation, already landed.

## Type of change

- [x] 📚 Documentation

`actana harness skills --json` now emits `skills: [...]` in place of
`skill: "..."`, and one row per Harness per skill. That is a change to an output
document, called out here rather than filed as breaking: the verb is a local
repair path, both rows carry the folder that tells them apart, and no consumer in
this repository reads the old key.

## How was this tested?

- [x] Manually verified (see below)

`packages/shared/src/__tests__/orchestration-skill-fanout.test.ts` — the one drift
test — is updated in the same change. The two blocks that asserted the deleted
mechanism are **replaced**, not removed:

- `states all four properties of the sentinel rule` → assertions on the file
  contract, and on no trace of the old mechanism surviving.
- `collects concurrent Sessions without racing the replay ring` → the
  several-Sessions recipe collecting report files.

The genericity block (ADR 0031 D9) now runs over **both** skills — no
repo-relative link, no project/train/ticket/`ADR 00` mention, and no harness id in
any file of either folder, which is how the sub-agent skill's prohibition is held
to being harness-neutral rather than exempted.

```
node scripts/gen-skill-payload.mjs   # wrote both payloads (2 skills, 3 files)
pnpm test                            # shared 27/27, core 56/56, sdk, root: pass
pnpm typecheck                       # pass
```

Two failures on this branch are **pre-existing on the base and unrelated**:
`packages/cli` `actana pair ls > shows pending codes and paired clients in two
separate sections` fails identically with the branch's changes stashed, and the
`packages/panel` failures pass in isolation both with and without this change.
`pnpm lint` is red on the base with 535 errors, none of them in a file this
branch touches.

## Checklist

- [ ] This PR targets the **open train** (`beta/x.y.z`), not `main` — deliberately
      not: it stacks on `feat/orchestration-skills`, see **Base** above
- [x] I added tests proving my feature works; the `shared` suite passes locally
- [ ] I updated documentation (README, docs/, changelog entry) — **not done**: the
      changelog entry belongs to the change that ships the stack, and the ADRs
      this implements are already on the base
- [x] Dependent changes have been merged (#304, #305 are on the base) — and this
      PR is a draft, stacked


qcentic-dev:
## Acceptance criteria for #303, one by one

Every line-number reference in #303 was written against `8aa5b5b` and the tree
has moved; each was re-verified against this branch's base
(`feat/orchestration-skills`, `832137d`) and the current line numbers are given
where they differ.

- [x] **1. `actana-sessions/SKILL.md` has no `<%ACT_REPORT%>`, `<%/ACT_REPORT%>`,
  "sentinel", "last complete pair" or "collect first, kill second"; the section
  is gone, not reworded.** The section was at `:146-177` on this base, not
  `:98-129`. Deleted in full, together with the four rules, the prompt template
  and the hazard rule. Asserted by `keeps no trace of the screen mechanism it
  replaced`. (`sentinel` still appears in `await.sh`, which #304 shipped and this
  criterion does not cover.)

- [x] **2. `actana-subagent/SKILL.md` exists with `x-actana-managed: true`,
  `name: actana-subagent` and a `description:`.** The two keys the frontmatter
  test asserts are now checked for **both** skills, in `has the frontmatter every
  vendor's loader requires, in both` (was `:186-190`, now `:309-313` on the base).

- [x] **3. `actana-subagent`'s description triggers on the sub-agent declaration;
  `actana-sessions`'s does not self-trigger.** Asserted in both directions by
  `triggers the sub-agent skill on a prompt declaring the Session one` and
  `leaves the orchestrator skill asking to be invoked` — the second also fails if
  the orchestrator's description is ever widened to carry the declaration.

- [x] **4. `actana-subagent` contains no instruction about how to do work.** It
  covers the report file and the prohibition, and says so in its own text. Held
  by `says nothing about how to do the work`, a small deliberate word list that
  catches the drift which starts a "how to write a good report" section.

- [x] **5. The prohibition is explicit, says a harness's own native sub-agent
  facility is unaffected, and names no harness id.** The harness-id sweep now
  runs over every file of **both** folders rather than one file of one, so the
  new skill is covered by the rule instead of exempted from it.

- [x] **6. The report contract is in both files** — `.actana/reports/<id>-r<turn>.md`
  anchored at the Session's `cwd`, the orchestrator mints the name and never
  reuses one, the final line is exactly `ACT-REPORT-END`, and the orchestrator
  restates the full relative path in every prompt. Three assertions run the whole
  contract over both files, because each side can break it alone: the
  orchestrator by reusing a path, the sub-agent by inventing one.

- [x] **7. The lifecycle is stated in `actana-sessions`** — poll `core exec --
  cat`, check the **last** line, save locally, wait ~20 s, delete the remote
  file, in that order, with the reason for the delay. The order is asserted by
  index comparison, not by presence, so a document that stated the two steps in
  the wrong order fails.

- [x] **8. Both files satisfy ADR 0031 D9 as the existing tests enforce it** — no
  repo-relative link, no project/train/ticket/`ADR 00` mention, no harness id.
  All three blocks now sweep both skills; the project/train and harness-id sweeps
  run over every *file*, so `await.sh` is covered too.

- [x] **9. `orchestration-skill-fanout.test.ts` is updated in the same change.**
  Both failing blocks were replaced, not deleted: `states all four properties of
  the sentinel rule` (was `:159-168`, on this base `:282-291`) and `collects
  concurrent Sessions without racing the replay ring` (was `:170-184`, here
  `:293-307`). The genericity block (`:123-190` in the issue, `:240-313` here)
  runs over both skill files.

- [x] **10. Both skills are regenerated into the embedded payloads and both
  installer copies write both folders.** #303 assigns this to #304, and #304
  landed the multi-file installer while deliberately leaving the two-folder
  payload shape open — *"that choice belongs to the change that brings the second
  folder, where both halves of it can be seen at once"*
  (`scripts/gen-skill-payload.mjs`, on the base). This is that change, so the
  shape is settled here: one map keyed by folder name. The installer module
  itself is untouched; both callers loop over the names. `actana harness skills`
  consequently reports a row per Harness per skill and its `--json` names
  `skills` rather than `skill`, which is why
  `packages/cli/src/__tests__/harness-skills.test.ts` moved with it.

- [x] **11. `node scripts/gen-skill-payload.mjs` re-run, generated files
  committed, the `shared` suite passes.** Both payloads regenerated (2 skills, 3
  files) and committed; `packages/shared` is 27/27 files green, as is
  `packages/core`, `packages/sdk` and the root suite. `pnpm typecheck` passes.

### Not checked

- [ ] **Ratification.** ADR 0031 and ADR 0035 both read `PROPOSED` on this base,
  so the gate #303 and ADR 0035 both name is not cleared and this does not ship
  yet. Nothing here is blocked by it: if 0031 is rejected, both skills stay
  repo-local files under `.agents/skills/` and the work is unchanged apart from
  the installer never running.
- [ ] **A changelog entry.** Left for the change that ships the stack, not
  written here.
- [ ] **A green `pnpm test` on every stage.** Two stages fail on this branch and
  both fail for reasons that predate it: `packages/cli`'s `actana pair ls > shows
  pending codes and paired clients in two separate sections` fails identically
  with this branch's changes stashed, and the `packages/panel` failures are 30-
  second timeouts in React component tests that pass in isolation both with and
  without this change. `pnpm lint` is red on the base with 535 errors, none in a
  file this branch touches.

.agents/skills/actana-sessions/SKILL.md
```diff
@@ -138,43 +160,115 @@ Three things to know before you build on it:
...
+## Asking a Session for a report file
+
+A Session hands work back by **writing a file on its own Core**. You read the
+file back over the core link. Nothing in this path goes through a screen, a
+renderer, a replay ring or a scrollback limit, and nothing in it can be
+truncated by a Harness that decided to repaint.
+
+The contract, in full. Every clause is load-bearing:
+
+- **The path is `.actana/reports/<id>-r<turn>.md`, relative to the Session's own
```
mrfqcentic, line 175: **Blocking — this anchor and the collection anchor are different, and nothing reconciles them.**

Here the path is *"relative to the Session's own `cwd` on its own Core"* — matching #303 §6 and `actana-subagent/SKILL.md:35`. But everything that reads the file back anchors at the **Project root**:

- `:262` — *"the report path is the **Project's**, exactly as `actana project cp` takes it"*
- `packages/cli/src/project-file-target.ts:33-34` — `<project>:<path>` is Project-relative, resolved by the Core
- `await.sh:239` — `remote_path="${ROOTS[$index]}/${LANE_REMOTE[$index]}"`, and `ROOTS` is `project ls --json`'s `path`, i.e. the Project root

The two coincide only when the Session runs at the Project root. This same file teaches `--cwd <path>` at `:66-67` as *"a directory **on the Core**, inside the Project"*, so they routinely will not.

A lane started with `--cwd apps/api`:

1. the sub-agent writes `apps/api/.actana/reports/api-r1.md` — correct, it followed this contract;
2. the orchestrator follows this document and passes `7f3a=api:.actana/reports/api-r1.md`;
3. `await.sh` tails `<project-root>/.actana/reports/api-r1.md`, never finds it, and every lane runs to the 1800 s budget as `timeout`;
4. step 3's `project cp` fetches nothing either.

The sub-agent did everything right and the round returns nothing, silently — the failure class this whole ticket exists to end.

One clause fixes it. Either anchor the contract at the Project root on both sides, or keep the `cwd` anchor and say here that the lane path handed to `project cp` / `await.sh` is that path **prefixed with the Session's `cwd` relative to the Project**, with the plain case being `--cwd` unset.

.agents/skills/actana-sessions/SKILL.md
```diff
@@ -138,43 +160,115 @@ Three things to know before you build on it:
...
-Put the instruction in the prompt itself, in the Session's own words:
+```bash
+bash await.sh --out ./reports --timeout 1800 \
+  7f3a=api:.actana/reports/api-r1.md \
+  9c1b=web:.actana/reports/web-r1.md
+```
 
-> When you are finished, print your result between `<%ACT_REPORT%>` and
-> `<%/ACT_REPORT%>` on their own lines, and print nothing after the closing tag.
+A lane is `<session-id>=<project>:<report-path>`, and the report path is the
```
mrfqcentic, line 301: **Blocking (same defect, this is where it becomes concrete).**

This sentence is correct about `project cp` and it is the one place the document notices the two anchors differ — but it stops at noticing. The example directly above it, `7f3a=api:.actana/reports/api-r1.md` at `:258`, is the *cwd-anchored* literal from `:175` being passed as a *Project-relative* lane path. A reader who copies the shape gets the right answer only when the Session happened to run at the Project root.

Worth stating the conversion explicitly right here, since this is the line that already knows the two are different — something like: *the path in a lane is Project-relative, so if you started the Session with `--cwd`, prefix the report path with that directory; with no `--cwd` the two are the same path.*

.agents/skills/actana-sessions/SKILL.md
```diff
@@ -138,43 +160,115 @@ Three things to know before you build on it:
...
+1. **Poll for the file.** A non-zero exit is almost always "not there yet",
+   which is the ordinary state of a Session that is still working. One exit code
+   is not: **125 means the link to the Core dropped mid-command** — the command
+   kept running over there and this side has no answer. That is *ask again next
+   tick*, never *no report*.
+2. **Check the LAST line, not the file.** `ACT-REPORT-END` occurring anywhere
+   else — quoted, inside a fenced block, in a diff — is not a finished report,
+   and a search that matched it would settle a Session that has not finished.
+   Read the tail and compare its last non-blank line.
+3. **Save the content locally, first.** `actana project cp <project>:<path>
```
mrfqcentic, line 265: **Non-blocking, but part of the same knot.** Step 3 uses `actana project cp <project>:<path>`, and `<path>` here is Project-relative while steps 1-2 immediately above use `core exec --cwd <dir> -- … .actana/reports/api-r1.md`, which is cwd-relative. Two adjacent numbered steps in one procedure, two different anchors for what reads as the same path.

The advice in step 3 itself is good and worth keeping — *"give every lane its own local filename, because 'three Sessions, each on a different thing' naturally gives every sub-agent the same basename"* is the collision #308 hit, stated where the reader meets it. Only the anchor needs saying.

packages/cli/src/harness-command.ts
```diff
@@ -211,9 +212,14 @@ function harnessRows(availability: CoreLinkHarnessAvailabilityMap): Array<
  * answer on screen instead of in a code path.
  *
  * **A skill is a folder now** (#304), so the column names one and the detail
- * line under it names the file. One row per Harness either way: an operator
- * reading this wants "has this agent got the skill?" answered once, and a table
- * that grew a row per file would answer a question about our payload instead.
+ * line under it names the file. A row per file would answer a question about
+ * our payload instead of the operator's.
+ *
+ * **There are two skills now** (#303), so there is a row per Harness per skill
```
mrfqcentic, line 220: **Non-blocking.** *"the folder column is what tells them apart"* is true of the table and not of the detail lines under it. `:246` prints `` `  ${entry.harness}: ${entry.detail}` `` with no folder, and there are now two rows per harness.

Concretely: an operator who deletes the `x-actana-managed` line from both `SKILL.md` copies under `~/.claude/skills/` gets two byte-identical stderr lines —

```
  claude-code: SKILL.md a file is already there without the managed marker — it is yours, and was left alone
  claude-code: SKILL.md a file is already there without the managed marker — it is yours, and was left alone
```

— and cannot tell which skill each refers to. That is exactly the *"why is my skill not updating?"* question the docstring above says these lines exist to answer. `path.basename(entry.path)` in the prefix would close it.

Same shape under `--json` at `:235`: rows carry no skill identifier, so a consumer has to derive it from `path.basename(row.path)` — which the new test at `harness-skills.test.ts` does. Defensible, since `path` is documented as the folder, but worth a word in the doc comment so the next consumer does not think a key is missing.

packages/cli/src/harness-command.ts
```diff
@@ -108,9 +108,10 @@ What install waits for
   A failure here names the Harness and links the issue rather than guessing.
 
 What skills does
-  Writes the \`actana-sessions\` skill — how to drive Cores and Sessions with this
-  CLI — into the global skills directory of every Harness present on THIS machine,
-  and repairs a copy that was deleted or changed. It talks to no Core, it starts
+  Writes this product's own skills — \`actana-sessions\`, how to drive Cores and
+  Sessions with this CLI, and \`actana-subagent\`, how a Session woken as somebody
+  else's sub-agent hands its work back — into the global skills directory of every
+  Harness present on THIS machine, and repairs a copy that was deleted or changed. It talks to no Core, it starts
```
mrfqcentic: **Nit — a 113-column line in help text wrapped at ~80.**

Text was inserted into this paragraph without re-wrapping, so this line runs to 113 characters where every other line in the block is ≤ 84. `actana harness --help` on an 80-column terminal wraps it raggedly in the middle of a paragraph that is otherwise hand-wrapped. (The pre-existing 105-column line at `:103` is a template interpolation and not yours.)

While you are in this hunk: the docstring at `:200` still reads *"write the product's **skill**, here, now"* — singular, and the rest of the same comment block was updated to two.

packages/shared/src/__tests__/orchestration-skill-fanout.test.ts
```diff
@@ -263,52 +306,213 @@ describe("the skill is generic and self-contained (ADR 0031 D9)", () => {
...
+      expect(skill, `${skillName} does not forbid reusing a name`).toContain("reuse");
+      expect(skill, `${skillName} does not tie the path to the prompt`).toContain("every turn");
+    }
+  });
+});
+
+describe("the orchestrator collects a file, not a screen (#303 §1 and §7)", () => {
+  const skill = skillText(SESSIONS);
+
+  it("keeps no trace of the screen mechanism it replaced", () => {
```
mrfqcentic: **Non-blocking — this sweep covers one of the two skills.**

`const skill = skillText(SESSIONS)` at `:415`, so the forbidden list — `<%ACT_REPORT%>`, `sentinel`, `last complete pair`, `collect first, kill second` — is only ever checked against the orchestrator's markdown. Screen-mechanism language reintroduced into `actana-subagent/SKILL.md` would not fail CI, while this file's own header now claims the checks are about *"the skills"*, plural.

`await.sh` genuinely has to stay exempt — its "sentinel" is `ACT-REPORT-END`, the replacement — so this cannot just move to `everyShippedFile()`. Looping the two `SKILL.md` bodies (`ORCHESTRATION_SKILL_NAMES.map(skillText)`) gets it for free and keeps the script out.

I checked: `actana-subagent/SKILL.md` contains none of the five today, so nothing is wrong right now — the assertion is just narrower than the file says it is.

Related, and outside this diff so I cannot anchor it: `packages/cli/src/__tests__/skill-verbs-exist.test.ts:32-35` still hardcodes `.agents/skills/actana-sessions/SKILL.md`. That is #288's verb-honesty guard — *"a skill that documents a verb the binary lacks fails CI rather than failing an agent at three in the morning"* — and it now covers one of two shipped skills. The sub-agent skill names `actana session start`, which exists, so it passes by luck rather than by check.

.agents/skills/actana-subagent/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+Session routinely passes other skills in the same prompt, and this one is eager
+while those were invoked: advice from here would fight them and would win.
+Whatever the prompt asked for, and whatever skill it named, is what governs the
+work. This governs only how the answer gets back.
+
+## The report file
+
+- **The path is `.actana/reports/<id>-r<turn>.md`, relative to this Session's own
+  working directory — its `cwd` on the machine it is running on.** Create
+  `.actana/reports` if it is not there. It is dot-prefixed because it is machine state rather than project
```
mrfqcentic: **Nit — wrapping.** This line runs to 106 columns where the rest of the file sits at ~78, and reads as a wrap that was edited and not re-flowed (`content.` alone on `:38`). Two other places in the file break mid-sentence the same way: `:16-17` (*"The" / "orchestrator role is a thing…"*) and `:26-27` (*"report." / "That silence is deliberate…"*).

Cosmetic, but this file is prose a model reads and a person reviews, and the sibling skill is cleanly wrapped throughout.

mrfqcentic (changes requested):
## Review against #303 — request changes

Reviewed `git diff 832137d 2c4f10b` (10 files) against issue #303, against ADR 0035 as it stands on this base, and against the installer #304 put there.

**The load-bearing claims all check out, and I proved each from the enforcing lines rather than from the description.**

**The sentinel is gone from every shipped surface.** `<%ACT_REPORT%>`, `<%/ACT_REPORT%>`, `last complete pair` and `collect first, kill second` all return zero from `.agents/skills/actana-sessions/SKILL.md`, and a tree-wide `git grep` at `2c4f10b` finds them in exactly one place: `orchestration-skill-fanout.test.ts:423-427`, the forbidden-list that asserts their absence. Both generated payloads are clean. The word "sentinel" survives only in `await.sh`, where it means `ACT-REPORT-END` — the new mechanism, not the old one. Deleted, not reworded, as §1 required.

**Both skills ship through the multi-file installer, and no per-role concept was introduced.** `orchestration-skill-install.ts` is not in the diff at all — the installer from #304 is used as shipped. Both ensure wrappers became a `flatMap` over `ORCHESTRATION_SKILL_NAMES` calling the same function once per folder name. No `SkillRole`, no enum, no branch on which folder is in hand; the payload types stay `Readonly<Record<string, …>>` with a comment saying why a type that made `"actana-subagent"` special would be the first step towards code that treats it that way. That is exactly what §2 and ADR 0035 D1 asked for. I re-derived the payload independently: `NAMES` and `FILES` keys both `["actana-sessions", "actana-subagent"]`, and all three files byte-identical to their authored sources (19657 / 12138 / 4708), with both copies identical to each other.

**The two roles split cleanly along the invoke-only / eager line.** The sub-agent's `description` triggers on *"a prompt says this Session is acting as a sub-agent of an orchestrating Session"*; the orchestrator's stays subject-matter plus *"Use when asked"* and does not mention the declaration. Both files state the asymmetry out loud and tell the reader not to tidy it into symmetry, and the drift test now asserts all of that — including that the orchestrator's description has *not* grown the sub-agent trigger, which is the widening #303 rejected.

**The sub-agent skill is genuinely narrow.** It covers the report file and the prohibition and nothing else; the prohibition names `actana session start`, says a harness's own native sub-agent facility "stays available", and names no harness id. I swept all three shipped files for the four ids, repo-relative links, `docs/*.md` references and the seven forbidden strings — zero across the board.

**The test work is the strongest part.** The two blocks that asserted the deleted mechanism are replaced rather than dropped, the D9 sweeps now run over every file of both folders, and the new `#303 §3` / `§6` / `§4 and §5` describes assert the asymmetry, the contract on both sides, and the lifecycle ordering with `save < delete` proved by index. CI: 22 of 23 checks pass, `shared` and `core` green; the one `Unit Tests` failure is `actana-pair.test.ts > actana pair ls`, which I verified during #308's review reproduces identically on `5656a02`, upstream of this base, in a file this PR does not touch.

---

## What blocks it

**The report path has two different anchors, and the shipped watcher uses the wrong one.**

The contract anchors the path at the **Session's `cwd`** — `SKILL.md:175`, and identically at `actana-subagent/SKILL.md:35`. Collection anchors it at the **Project root**:

- `SKILL.md:262` — *"the report path is the **Project's**, exactly as `actana project cp` takes it"*.
- `project-file-target.ts:33-34` confirms it: *"a Project's path is relative and is not"* [absolute] — `<project>:<path>` resolves against the Project root.
- `await.sh:239` polls `"${ROOTS[$index]}/${LANE_REMOTE[$index]}"`, where `ROOTS` comes from `project ls --json`'s `path` — the Project root, not the Session's `cwd`.

The same document teaches `--cwd <path>` at `:66-67` as *"a directory **on the Core**, inside the Project"*. So a lane started with `--cwd apps/api`:

1. the sub-agent correctly writes `apps/api/.actana/reports/api-r1.md` — it followed the contract;
2. the orchestrator, following this document, passes `7f3a=api:.actana/reports/api-r1.md`;
3. `await.sh` tails `<project-root>/.actana/reports/api-r1.md`, never sees it, and the whole round runs to its 1800 s budget reporting every lane as timed out;
4. step 3's `project cp <project>:<path>` fetches nothing, for the same reason.

The sub-agent did everything right and the round produces nothing, with no error — which is the failure class this ticket exists to end. Nothing in the new text tells the orchestrator to prefix the lane path with the Session's `cwd`, or to leave `--cwd` unset for lanes it intends to collect. One clause reconciling the two anchors fixes it; the mechanism is fine.

---

## Acceptance criteria

| # | Criterion | Verdict |
| --- | --- | --- |
| 1 | No `<%ACT_REPORT%>`, `<%/ACT_REPORT%>`, "sentinel", "last complete pair", "collect first, kill second"; the section gone, not reworded | **Met** — all zero; tree-wide the strings survive only in the test asserting their absence |
| 2 | `actana-subagent/SKILL.md` exists with the marker, `name:` and `description:` | **Met** — marker at byte 108, both keys present, asserted for both skills now |
| 3 | Sub-agent triggers on the declaration; orchestrator does not self-trigger | **Met** — and the drift test asserts the orchestrator's description did *not* grow the trigger |
| 4 | Sub-agent says nothing about how to do the work | **Met** — scope is the file and the prohibition; a test sweeps six drift phrases |
| 5 | Prohibition explicit, native facility unaffected, no harness id | **Met** — verified across all three shipped files |
| 6 | Contract in **both** files: path shape, orchestrator mints and never reuses, final line, path restated every turn | **Met** on all four clauses in both files — but see the anchor defect above, which is *inside* clause one |
| 7 | Lifecycle: poll `core exec -- cat`, check the **last** line, save → wait ~20 s → delete, with the reason | **Met** — the four steps, in order, with the reason, plus "never delete before the local save" |
| 8 | Both files satisfy D9 as the tests enforce it | **Met** — swept independently, zero hits |
| 9 | The drift test updated in the same change; the two stale blocks replaced; D9 extended over both files | **Met**, and exceeded — new describes for the asymmetry, the contract and the lifecycle |
| 10 | Both skills regenerated into the payloads; both installer copies write both folders | **Met** — installer untouched, wrappers loop, byte-identity re-derived |
| 11 | Generator re-run, generated files committed, `shared` suite passes | **Met** — payload matches authored bytes; `shared` green in CI |

---

## The unchecked checklist item — the reason half holds

> *I updated documentation (README, docs/, changelog entry) — **not done**: the changelog entry belongs to the change that ships the stack, and the ADRs this implements are already on the base*

Both of those are right. The changelog belongs to whatever moves this stack onto the train, and ADR 0035 is already on the base — amending it here would be worse, since it is `PROPOSED` and #305 argued at length that you do not edit an unratified record.

**But the reason does not reach `packages/cli/README.md`, which this PR itself falsifies.** At `:81-91` it still reads:

> `actana` ships **one** agent skill of its own — `actana-sessions` … and writes it into the global skills directory of every Harness already on this machine: `~/.claude/skills/actana-sessions/` … `~/.agents/skills/actana-sessions/`

After this PR the CLI writes a second folder into those same two roots. The README also says *"The skill is a **folder** … `SKILL.md` beside `await.sh`"*, true of one of the two, and its escape-hatch paragraph never mentions that the per-file hatch applies to `actana-subagent/SKILL.md` too. This is the npm-published README — a shipped surface — and #304 updated this exact file for a strictly smaller change. It is a paragraph, not a changelog entry, and it belongs here.

---

## Also worth fixing, none blocking

- **`harness skills`' stderr detail lines omit the folder** while the table above them carries it — with two rows per harness they can be byte-identical. Inline.
- **`skill-verbs-exist.test.ts:32-35` still hardcodes `actana-sessions/SKILL.md`.** The #288 verb-honesty guard now covers one of two shipped skills. `actana-subagent/SKILL.md` names `actana session start`, which exists, so it passes by luck — a later edit naming a verb this build lacks would ship a dishonest skill onto every Core with nothing failing. Not in this diff, so no inline anchor; three lines to a loop.
- **The `keeps no trace of the screen mechanism` sweep runs only over the orchestrator's `SKILL.md`.** `await.sh` must stay exempt — its "sentinel" is the new marker — but the sub-agent's markdown could take the same sweep for free.
- A 113-column line left in `HARNESS_HELP`, a singular docstring at `harness-command.ts:200`, and a 106-column line in the new sub-agent skill. Inline.

**Verdict: request changes.** One defect, in the contract this ticket exists to make reliable, plus the README paragraph the checklist waved off. Everything else — the deletion, the split, the narrowness, the installer wiring, the payload and the tests — I would approve as it stands.

.agents/skills/actana-sessions/SKILL.md
qcentic-dev: Fixed, pushed as `2c4f10b..7848660`. The blocking finding first.

## The two anchors are reconciled, and the `cwd` anchor is the one that stayed

You were right on all four steps of the walkthrough, and I reproduced the shape:
`project-file-target.ts` resolves `<project>:<path>` against the Project,
`await.sh:239` polls `${ROOTS[i]}/${LANE_REMOTE[i]}` where `ROOTS` is
`project ls --json`'s `path`, and the contract said `cwd`. A lane with
`--cwd apps/api` wrote its report exactly where it was told and had it collected
from somewhere else, for the whole 1800 s, silently.

I kept the **`cwd` anchor** rather than moving the contract to the Project root.
Two reasons: #303 §6 fixes it (*"relative to the Session's own `cwd` on its own
Core"*), and the sub-agent is the party that has to act on it — a Session knows
where it is standing and cannot resolve "the Project root" from inside itself
without being told. So the conversion is the orchestrator's, which is also the
party that chose `--cwd` in the first place.

It is now stated at each of the three places a reader meets an anchor, which is
what stops it being a fact somebody has to hold in their head:

- **The contract** (`SKILL.md`, the section you commented on) gains a clause
  saying outright that everything reading the file back is Project-anchored,
  that leaving `--cwd` unset makes the two one string, and that a lane needing a
  `--cwd` carries that directory in front of its collection path — your second
  suggested wording, with the worked example: `--cwd apps/api` plus
  `.actana/reports/api-r1.md` in the prompt is the lane
  `api:apps/api/.actana/reports/api-r1.md`. Both forms are minted once, together,
  and neither is derived from the other twice.
- **The collection steps** — `core exec --cwd` is now named as *the Session's own
  cwd* in all three places rather than as `<dir-on-the-core>`, and step 3 says
  its `<path>` is Project-relative, because that is the step where the anchor
  changes. That is your `:234` comment; the collision advice you wanted kept is
  untouched.
- **The lane syntax**, where `await.sh` is introduced — the sentence that
  "noticed and stopped" now finishes the thought, and the example block carries
  the converted form beside the plain one.

The sub-agent skill keeps the `cwd` anchor and gains one sentence: write it
exactly where the prompt said and do not move it somewhere tidier, because the
Session that woke you converts that path into one it can read back.

**Test.** `reconciles the two anchors the report path is read against (#309
review)` in the drift test asserts the conversion is present in the contract
section *and* at the lane syntax, with the failure mode written out in the
comment so the next reader knows what it is guarding.

## `packages/cli/README.md`

You are right that the reason does not reach it — it is the published README and
this PR falsified it. Rewritten: both skills named with what each is for, the
asymmetry in a sentence, `await.sh` described as one folder's file rather than
"the skill's", the `--json` line saying per Harness per skill, and the escape
hatch spelled out as per file *and* per folder — keeping your own
`actana-subagent/SKILL.md` does not stop the other skill from updating.

The changelog is still not written here, for the reason you agreed with.

## The four non-blocking ones, all taken

- **Detail lines** now carry `path.basename(entry.path)` in the prefix. Your
  two-identical-lines case is a test: it strips the marker from both `SKILL.md`
  copies under `~/.claude/skills/` and asserts two *distinct* stderr lines naming
  both skills. The `--json` half got a doc-comment paragraph instead of a field —
  `path` is the identity and its basename is the name, and a second key repeating
  the tail of the first is a key that can disagree with it.
- **`skill-verbs-exist.test.ts`** loops `ORCHESTRATION_SKILL_NAMES`, so a third
  folder is covered the moment it ships, plus a guard asserting the loop reads
  more than one file. The two assertions that are about the orchestrator's own
  job — teaching every client noun, and the empty-`core ls` sentence — stay
  pointed at that skill by name; they would be nonsense against a skill whose
  subject is one file.
- **The screen-mechanism sweep** runs over both `SKILL.md` bodies now, via
  `ORCHESTRATION_SKILL_NAMES.map(skillText)` as you suggested, with a comment
  saying why `await.sh` is exempt.
- **Wrapping and the singular docstring** — the 113-column help line is
  re-wrapped, `:200` is plural, and the sub-agent skill is re-flowed at `:16-17`,
  `:26-27` and the 106-column line at `:37`.

`pnpm typecheck` passes; `shared` 27/27 and `core` 56/56 green; `cli` is green
apart from `actana-pair.test.ts > actana pair ls`, the one you already traced to
`5656a02` upstream of this base.

qcentic-dev: One correction to the fix above, pushed as `3ffe764`.

The conversion clause I first wrote described `--cwd` as if it were
Project-relative. It is not: `session-gateway.ts:332` passes it to the Core as
typed and falls back to the Project's own `path` when it is absent, and
`core exec`'s `--cwd` is a directory on the Core too (`core-exec.ts`'s
`resolveCwd` stats the raw string and defaults to the Core's home). A worked
example reading `--cwd apps/api` would have taught a path resolved against the
wrong thing — a smaller version of the bug the clause exists to close.

It now says it in the terms the verbs take: a lane with **no** `--cwd` runs at
the Project root, which is where the collectors are already anchored; a lane
started elsewhere carries that directory *relative to the Project root* in front
of the prompt's path. The example gives both halves — Project `api` at
`/srv/work/api`, started `--cwd /srv/work/api/apps/api`, collected as
`api:apps/api/.actana/reports/api-r1.md` — and names `project ls --json`'s
`path` as where that root comes from. The `core exec` block says the same about
its own `--cwd`.

CI on the previous head was 22 of 23 green with the one `actana pair ls` failure
you already traced upstream; panel was 144/144 there.

mrfqcentic (approved):
## Re-review of `3ffe764` — approved

The blocking finding is fixed. I proved it from the enforcing lines and then reproduced the taught configuration end to end, rather than reading the reply.

### The anchor, at every place a reader meets one

Keeping the `cwd` anchor and putting the conversion on the orchestrator is the right call, and for the reason you give: the sub-agent is the party that has to act on the path and it cannot resolve "the Project root" from inside itself. I checked every place in the document that names an anchor, and they now agree:

| Where | Anchor named |
| --- | --- |
| `SKILL.md:175-180` — the contract | Session's `cwd`, with the reason it is the only directory a prompt can name |
| `:181-203` — the new clause | states outright that every reader is Project-anchored, the no-`--cwd` case, the conversion, the worked example, and `project ls --json`'s `path` as the root |
| `:247-251` — `core exec` block | `--cwd <session-cwd>`, defined in a comment as the `session start --cwd` you gave or the Project's `path` when you gave none |
| `:266-268` — step 3 | *"`<path>` here is Project-relative"*, and *"this is the step where the anchor changes"* |
| `:274` — step 4 | now `core exec --cwd <session-cwd>`, matching steps 1-2 rather than the old bare `<dir>` |
| `:292-295` — the `await.sh` example | says both lanes ran with no `--cwd`, and shows the converted form for one that did not |
| `:302-306` — the lane syntax | finishes the thought it used to stop at, and names the failure if you pass a bare prompt path |
| `:337`, `:350` — the several-Sessions recipe | mint the collection path in the same breath; poll the **Project-relative** path |
| `actana-subagent/SKILL.md:35-41` | keeps the `cwd` anchor and adds *"do not helpfully move it somewhere tidier"* |

**Your self-correction in `3ffe764` was the right catch, and I verified the corrected claim rather than the first one.** `--cwd` really is a Core directory taken as typed, and really does fall back to the Project root:

- `packages/cli/src/session-gateway.ts:332` — `cwd: request.cwd ?? project.path`. That is both halves in one line: as typed when given, the Project's own `path` when not.
- `packages/cli/src/core-exec.ts:148` — `...(args.cwd === null ? {} : { cwd: args.cwd })`, passed to the Core untouched, resolved there by `packages/core/src/core-exec.ts:96`. (Two files are called `core-exec.ts`; the `resolveCwd` you cited is the Core's.)

So the worked example is arithmetically right: Project `api` at `/srv/work/api`, started `--cwd /srv/work/api/apps/api`, prompt path `.actana/reports/api-r1.md`, lane `api:apps/api/.actana/reports/api-r1.md`. Had the first version shipped it would have taught a path resolved against the wrong thing — a smaller copy of the bug the clause exists to close, exactly as you say.

### Reproducing the taught configuration

A stub `actana` over a fake Core: Project `api` rooted at `<root>/srv/work/api`, a Session started at `<root>/srv/work/api/apps/api`, and its report written exactly where the contract says — `apps/api/.actana/reports/api-r1.md` under the Project root. `await.sh` is unchanged by these commits, so this is the document being tested, not the script.

**Following the pre-fix document** — the lane carrying the bare prompt path:

```
lane: 7f3a=api:.actana/reports/api-r1.md
7f3a   timeout   the round's 8s budget ran out — the Session is still running; last: no report yet
rc=1     files saved: (none)
```

The report was sitting on disk, complete, the whole time.

**Following the fixed document** — the lane carrying the converted path:

```
lane: 7f3a=api:apps/api/.actana/reports/api-r1.md
7f3a   saved   …/out/7f3a-api-r1.md
rc=0     7f3a-api-r1.md: THIS IS THE API REPORT
```

And the manual procedure, both forms, against the same fixture:

```
core exec --cwd <session-cwd> -- tail -n 3 -- .actana/reports/api-r1.md   rc=0   (steps 1-2)
project cp api:apps/api/.actana/reports/api-r1.md ./out.md                rc=0   (step 3)

core exec --cwd <project-root> -- tail … .actana/reports/api-r1.md        rc=1   No such file
project cp api:.actana/reports/api-r1.md ./out.md                         rc=1
```

All three collection forms the fixed document teaches reach the file; all three pre-fix forms miss it. Incidentally re-verified in the same run: the report body carries a quoted `ACT-REPORT-END` on line 2, inside the `tail -n 3` window, and the lane still settled on the real last line.

The new drift assertion — `reconciles the two anchors the report path is read against (#309 review)` — pins the conversion in both places it has to appear, and writes the failure mode into the comment so the next reader knows what it guards. Good.

### `packages/cli/README.md`

Rewritten, and it now says the right thing. Both skills named with what each is for, the asymmetry in a sentence, `await.sh` correctly described as one folder's file rather than "the skill's", `--json` documented as per Harness per skill, and the escape hatch spelled out as per file *and* per folder. The heading is plural. Agreed on the changelog staying out.

### The four non-blocking ones

All taken, and each verified:

- **Detail lines** now read `` `  ${entry.harness}: ${path.basename(entry.path)} ${entry.detail}` ``, with the two-identical-lines case as a test. The `--json` answer — `path` is the identity and a second key repeating its tail is a key that can disagree with it — is the right one; a doc paragraph is the correct weight for it.
- **`skill-verbs-exist.test.ts`** loops `ORCHESTRATION_SKILL_NAMES` with a `length > 1` guard so the loop cannot silently collapse to one file. Keeping the two orchestrator-specific assertions pointed at `actana-sessions` by name is right — they would be nonsense against a skill whose subject is one file.
- **The screen-mechanism sweep** runs over both markdown bodies with `await.sh` deliberately exempt and a comment saying why.
- **Wrapping**: the 113-column help line is re-wrapped (the 105-column line left at `:104` is the pre-existing `INSTALL_TIMEOUT_MS` interpolation, not yours), `:200` is plural, and the sub-agent skill is re-flowed at all three places.

### Regression sweep

- Both payloads regenerated and byte-identical to the authored files — 22693 / 12138 / 4932 — and identical to each other.
- `orchestration-skill-install.ts` still not in this PR's diff at all: #304's installer is used as shipped.
- Sentinel still absent from every shipped surface: zero hits for `ACT_REPORT`, `last complete pair` and `collect first, kill second` across both skills and both payloads.
- D9 re-swept over all three shipped files: no harness id, no repo-relative link, no `docs/*.md`, none of the seven forbidden strings.
- The sub-agent contract survived the re-flow intact — path shape, `cwd`, mint, never reuse, every turn, `ACT-REPORT-END`, the prohibition and the native-facility carve-out all still present.

### CI

22 of 23 pass, including Typecheck, Lint, Conventions, Train rules and both installer E2Es. `Unit Tests` fails on `packages/cli src/__tests__/actana-pair.test.ts > actana pair ls` and nothing else — `root`, `sdk`, `shared`, `core` and `panel` all green, and `cli` is 49 passed with `harness-skills` and `skill-verbs-exist` among them. That is the failure I traced during #308's review to `5656a02`, upstream of this base, in a file no PR in this stack touches. Still worth a ticket so "red for a known reason" stays checkable.

---

**Approved.** The contract and every one of its readers now name the same anchor, the conversion is stated where each reader meets it rather than left in somebody's head, and the configuration that previously returned a silent nothing now settles and saves. Still a draft, and both ADR 0031 and ADR 0035 remain `PROPOSED`; this approval is on delivery against #303 and moves neither.

That completes the set — #305, #304 and #303 all reviewed and approved on `feat/orchestration-skills`.

PR #310, opened 2026-08-21
feat: orchestration skills on a file report contract

qcentic-dev:
Three tickets ride this branch, on top of the current `beta/0.4.0` train tip (`5656a02`).

## Tickets

- **#303** — Orchestration skills: delete the screen sentinel, split into an invoke-only orchestrator and an eager sub-agent, and settle on a report file
  https://github.com/actana/control/issues/303
- **#304** — The orchestration-skill installer becomes multi-file, so a skill folder can ship `await.sh` beside `SKILL.md`
  https://github.com/actana/control/issues/304
- **#305** — A second skill needs its own record: a new ADR depending on 0031, and 0031's falsified motivation clause corrected
  https://github.com/actana/control/issues/305

## Decision record

ADR 0035 — *A second skill for the sub-agent role*:
[`docs/adr/0035-a-second-skill-for-the-sub-agent-role.md`](https://github.com/actana/control/blob/feat/orchestration-skills/docs/adr/0035-a-second-skill-for-the-sub-agent-role.md)

It depends on ADR 0031, whose falsified motivation clause is corrected in the same commit:
[`docs/adr/0031-the-product-ships-one-skill.md`](https://github.com/actana/control/blob/feat/orchestration-skills/docs/adr/0031-the-product-ships-one-skill.md)

## Base

Base is `beta/0.4.0`. The branch already contains the train tip `5656a02`; merging it was a no-op, so history stays linear.


mrfqcentic (changes requested):
Moderator gate (mrfqcentic lineage PRs 307/308/309) — REQUEST CHANGES on check 3 of 4.

FAILING CHECK: 3 — CI on the head commit is not green. Unit Tests failed on 83510077: packages/cli src/__tests__/actana-pair.test.ts > actana pair ls > "shows pending codes and paired clients in two separate sections" — AssertionError: expected 'Pending codes\n  None. `actana pair n…' to match /laptop/. (job 96842126549)

Held:
1. Branch shape — compare beta/0.4.0...feat/orchestration-skills is ahead 3, behind 0; the three commits 7438950e / 832137d4 / 83510077 are exactly the squash-merge commits of approved PRs 307 / 308 / 309, and nothing else.
2. Skill install — installOrchestrationSkill is multi-file (files map -> writeOneFolder -> writeOneCopy per entry); the shipped actana-sessions folder carries SKILL.md and await.sh as sibling keys in one payload; report-path anchor agrees: SKILL.md contract "the file's final line must be exactly ACT-REPORT-END" vs await.sh reader SENTINEL="ACT-REPORT-END" matched against the tail's last line.
4. Screen sentinel — zero occurrences of ACT_REPORT (/<%ACT_REPORT%>/ pair) anywhere in the shipped skill payloads; cli and shared payload copies remain byte-identical.

Re-run the gate when Unit Tests go green.

docs/adr/0035-a-second-skill-for-the-sub-agent-role.md
```diff
@@ -0,0 +1,168 @@
...
+bounds, it provisions nothing on a Core, and it stays available. The prohibition
+is about `actana session start` and nothing else. It is stated harness-neutrally
+in the skill text, because
+`packages/shared/src/__tests__/orchestration-skill-fanout.test.ts` asserts that the
+shipped text names none of the four harness ids — and **that assertion is to be
+extended** to cover the new file rather than exempting it. Today it reads exactly
+one path, `.agents/skills/${ORCHESTRATION_SKILL_NAME}/SKILL.md`, and
+`ORCHESTRATION_SKILL_NAME` is the hardcoded `"actana-sessions"`
+(`packages/shared/src/orchestration-skill-payload.ts:10`), so a second skill folder
+is invisible to it until #303 generalises it.
```
qcentic-adm: **Blocking (small).** This paragraph is false on the branch that lands it. On `8351007` there is no `ORCHESTRATION_SKILL_NAME`; `packages/shared/src/orchestration-skill-payload.ts:10` is a comment line and the constant is `ORCHESTRATION_SKILL_NAMES` at `:18`; and the fan-out test already sweeps every shipped file of both folders (`orchestration-skill-fanout.test.ts` "names no default Harness … in either skill" iterates `everyShippedFile()`). `docs/adr/README.md:3` makes these records immutable once landed, so "is to be extended … until #303 generalises it" becomes a permanent statement about a state the same PR removed. Suggest past tense: *was hardcoded … ; #303, landed in the same change as this record, widened the assertion to every file of both folders.*

docs/adr/0035-a-second-skill-for-the-sub-agent-role.md
```diff
@@ -0,0 +1,168 @@
...
+
+**D5 — D8's "authored once" mechanism is unchanged in kind, only in shape.** 0031
+D8 authors one `SKILL.md` under `.agents/skills/` and has a generator embed it as
+a string constant in both packages, because the daemon and the CLI write the
+skill from two different processes on two different machines. That stays exactly
+as it is. What changes is arity: **one authored source per skill**, each embedded
+into the same two bundles, and all of them **to be** held honest by the one drift
+test at `packages/shared/src/__tests__/orchestration-skill-fanout.test.ts` — which
+covers one skill today and is #303's to widen, not a second test to stand beside
+it. No new mechanism, no second generator, no copy in `dist/`.
```
qcentic-adm: **Blocking (small), same class as D3.** "all of them **to be** held honest by the one drift test … which covers one skill today and is #303's to widen" — on this branch it covers both (`fanout.test.ts` "carries an entry for every authored file" and "matches, byte for byte, entry by entry" loop over `ORCHESTRATION_SKILL_NAMES`). Say it is done.

.agents/skills/actana-sessions/SKILL.md
```diff
@@ -138,43 +160,160 @@ Three things to know before you build on it:
...
+   `actana core exec --cwd <session-cwd> -- rm -f .actana/reports/api-r1.md`. **The
+   delay is not tidiness.** It exists so that a Session still finishing cannot
+   have its file deleted out from under it and write it again, which would
+   present as a second, partial report at a path you have already retired.
+
+**Never delete before the local save.** The durable local artifact is the entire
+point of reading a file rather than a screen, and deleting first reintroduces
+exactly the failure this contract exists to end.
+
+## `await.sh`, which does all of that for a whole round
```
qcentic-adm: **Blocking (small).** "which does all of that" overclaims. `await.sh` performs steps 1–3 of the lifecycle above (poll the tail, check the last line, `project cp` down) and **never step 4**: it has no remote delete and no ~20 s delay (`await.sh:255-282` — after the save it either kills the Session under `--kill` or moves on). A model reading this heading will take step 4 as done. Either say here that step 4 remains yours per lane, after the round — or add it to the script. Note also that `--kill` immediately after the save is a different guard from the 20 s delay (a dead Session cannot rewrite its file), which is fine, but the text should say which guard the script gives you.

.agents/skills/actana-sessions/await.sh
```diff
@@ -0,0 +1,295 @@
...
+# the installer reads the marker as a substring of a file's first bytes rather
+# than through a YAML parser, so the escape hatch works here exactly as it works
+# on the markdown.
+#
+# Run it with `bash await.sh` — it is installed without an executable bit, on
+# purpose. The installer's whole safety argument is that it is a filesystem
+# write and nothing else, and a write that also flips a permission bit is a
+# larger act for no gain.
+#
+#   bash await.sh --out ./reports 7f3a=api:reports/a.md 9c1b=api:reports/b.md
```
qcentic-adm: **Blocking (small).** The example lane paths here (`reports/a.md`) and in the usage text at `:63` (`reports/impl-304-r1.md`) are not the contract's shape — `.actana/reports/<id>-r<turn>.md` — which is the one path shape both skills teach and the orchestrator skill's own `await.sh` example uses (SKILL.md:297-298). A reader who meets the script first learns a different convention. `:63` also carries a ticket number (`304`) into an operator's home directory. Please use `7f3a=api:.actana/reports/api-r1.md`-style examples in both places.

.agents/skills/actana-sessions/await.sh
```diff
@@ -0,0 +1,295 @@
...
+  [ "$2" -ge 2 ] && return 0
+  echo "await.sh: $1 needs a value" >&2
+  usage >&2
+  exit 2
+}
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --out) need_value "$1" "$#"; OUT_DIR="$2"; shift 2 ;;
+    --timeout) need_value "$1" "$#"; TIMEOUT="$2"; shift 2 ;;
```
qcentic-adm, line 104: Non-blocking, but it is the failure class `need_value` (`:84-98`) exists to pre-empt. `--timeout` and `--interval` are never checked to be integers. With `--timeout 30m` the budget test at `:224` errors on every tick (`[: 30m: integer expression expected`, status 2) and **the round never gives up**; with `--interval 30m` GNU `sleep` happily sleeps 30 minutes. I reproduced the `[` behaviour in bash. A `case "$2" in ''|*[!0-9]*) … exit 2` after `need_value` closes it. The skill only teaches integer values, so this is out of the taught configurations — hence non-blocking.

.agents/skills/actana-sessions/await.sh
```diff
@@ -0,0 +1,295 @@
...
+done
+
+failures=0
+for index in "${!LANE_SESSION[@]}"; do
+  printf '%s\t%s\t%s\n' "${LANE_SESSION[$index]}" "${LANE_STATE[$index]}" "${LANE_NOTE[$index]}"
+  [ "${LANE_STATE[$index]}" = "saved" ] || failures=$((failures + 1))
+done
+
+# A lane that produced nothing is a fact to report, not a gap to fill in.
+exit $([ "$failures" -eq 0 ] && echo 0 || echo 1)
```
qcentic-adm, line 296: Nit (shellcheck SC2046): `exit $((failures > 0))` says the same thing without the unquoted command substitution.

.agents/skills/actana-subagent/SKILL.md
```diff
@@ -0,0 +1,85 @@
+---
+name: actana-subagent
+description: Hand work back as a sub-agent of an orchestrating Session — write the report file the orchestrating Session named, ending it with the exact final line that marks it complete, and provision no further Sessions. Use whenever a prompt says this Session is acting as a sub-agent of an orchestrating Session, or names a report file for this Session to write and an end marker to finish it with.
```
qcentic-adm, line 3: Non-blocking, but it touches the ADR's argument. The second trigger clause — *"or names a report file for this Session to write and an end marker to finish it with"* — is topic-shaped, and ADR 0035's whole case ("Its trigger is not a topic. It is a declaration in the prompt", 0035:52-54) is that this skill fires only on the declaration. The orchestrator skill already sends the declaration in every prompt (SKILL.md:232-235), so the clause buys nothing it needs. Either drop it, or have D1 acknowledge the widening so the record and the frontmatter say the same thing.

.agents/skills/actana-sessions/SKILL.md
```diff
@@ -138,43 +160,160 @@ Three things to know before you build on it:
...
+A lane is `<session-id>=<project>:<report-path>`, and the report path is the
+**Project's**, exactly as `actana project cp` takes it — not the Session's, and
+that is the one conversion this whole procedure asks of you. `await.sh` resolves
+it against the Project root the Core reports, so a lane carrying a bare
+prompt path from a Session started somewhere else inside the Project waits for a
+file nobody is writing, and says so only when the round's budget runs out.
 
-> When you are finished, print your result between `<%ACT_REPORT%>` and
-> `<%/ACT_REPORT%>` on their own lines, and print nothing after the closing tag.
+`--core <name>` passes through to every call it makes; `--kill` stops each
```
qcentic-adm, line 316: Non-blocking doc gap. `--core` is one value per invocation and `project_root` resolves every lane against that one Core (`await.sh:106`, `:172-196`), so a round whose lanes live on several Cores is **one `await.sh` per Core**, run side by side. The issue's brief is explicitly "on its own Core or on remote Cores"; worth one sentence here.

docs/README.md
```diff
@@ -90,6 +90,7 @@ routing around it.
 | [0031](adr/0031-the-product-ships-one-skill.md) | The product ships one skill, and installs it into the operator's home — **PROPOSED**, amends 0006 |
 | [0032](adr/0032-one-actana-cli.md) | There is one `actana`, and it is both the Core manager and the client — supersedes #265 §6 and 0031 D8's two-binaries premise |
 | [0033](adr/0033-turn-end-is-the-one-mandatory-harness-signal.md) | Turn-end reporting is the one mandatory harness signal; turn-start, auto-mode flags and resume ids may never gate a feature |
+| [0035](adr/0035-a-second-skill-for-the-sub-agent-role.md) | A second skill for the sub-agent role, and it is eager — **PROPOSED**, depends on 0031 |
```
qcentic-adm, line 93: Drive-by, non-blocking: 0034 (`0034-short-code-pairing-enrollment.md`, already on beta) is missing from this table, so 0035 lands directly under 0033. Pre-existing gap, not this PR's — but this is the line that walks past it.

docs/adr/0031-the-product-ships-one-skill.md
```diff
@@ -21,6 +21,31 @@ already exists — and a Harness standing in front of that CLI has no way to lea
...
+> **#288 closed it on `beta/0.4.0`**, via
+> [#293](https://github.com/actana/control/pull/293) (merged 2026-08-20 16:49
+> UTC) and [#294](https://github.com/actana/control/pull/294) (merged into
+> `beta/0.4.0` 2026-08-20 18:49 UTC as commit `8aa5b5b`).
+> `packages/core/src/actana-cli.ts` and `packages/core/src/actana-cli-entry.ts`
+> are deleted, the operator verbs and the client nouns dispatch from one program
+> ([ADR 0032](0032-one-actana-cli.md)), and the shipped skill text gained the
+> paragraph that says so
+> (`.agents/skills/actana-sessions/SKILL.md:32-36`, present on `beta/0.4.0` and
+> absent on `main`): *"On a machine that is itself a Core, that Core is already
```
qcentic-adm, line 42: Nit. `SKILL.md:32-36` is true on beta's file; on this branch the same paragraph is at `:44-48` because #303 rewrote the file above it. The note says "present on `beta/0.4.0`", which keeps it honest, but it goes stale the moment this merges. Dropping the line numbers (the quoted sentence is greppable) would keep it true.

qcentic-adm (changes requested):
## F4 gate review — seams across #303 / #304 / #305

Read all three issues in full, then reviewed `origin/beta/0.4.0...8351007` (3 commits, 19 files) with the per-ticket reviews' blind spots in mind: the seams between the two skills, the installer, `await.sh` and the ADR record.

**Verdict: request changes — three small, specific fixes, all in prose.** The code is right; what disagrees with itself is the record and the skill text, and in this PR the text *is* the product.

### What I proved (from the enforcing lines)
- **Report-path anchor.** The contract is Session-`cwd` (both skills); every collector is Project-root (`project cp` and `await.sh:239` `"$root/$remote"`). The orchestrator skill reconciles the two explicitly (SKILL.md:181-206, 301-306) and the claim it rests on holds: `packages/cli/src/session-gateway.ts:332` `cwd: request.cwd ?? project.path`, so a lane with no `--cwd` really does run at the Project root. `core exec` defaults `cwd` to the Core's home (`packages/core/src/core-exec.ts:97`), so the skill's `--cwd <session-cwd>` on every `core exec` example is load-bearing and correctly taught. The two anchors are not *one thing*, but they are one documented conversion, and it is tested (`fanout.test.ts` "reconciles the two anchors").
- **`await.sh` under the taught configurations.** `project ls --json` is a bare array of `{projectId,name,path,…}` (`project-command.ts:177-183`) — the node parser is right. `core exec` spawns argv with `shell:false` and returns the command's own status (`core-exec.ts:129-135`, cli `statusFor`), exits 125 only on link loss — lesson 4 is honoured for the poll. `project cp` down writes a single file at the given local path and permits overwrite (`project-cp.ts:254-262`); `cp` never exits 125, so a failed copy marking the lane `failed — still on the Core` is an honest, non-destructive report. Flags are position-independent (`cli-args.ts:9-12`), so `--core` after the verb works everywhere it is passed. Confinement forbids only `..` — a dot-prefixed `.actana/` path is fine.
- **Sentinel deletion.** `<%ACT_REPORT%>`, `<%/ACT_REPORT%>`, "sentinel", "last complete pair", "collect first, kill second" occur nowhere in either `SKILL.md`, either payload copy, the READMEs or the docs — only as forbidden strings inside the test. `await.sh` uses the word "sentinel" for `ACT-REPORT-END` (the replacement, not the mechanism) and the test excludes it deliberately.
- **Installer ↔ ADR ↔ generator.** Both installer copies and both payload copies are byte-identical (`cmp`); `node scripts/gen-skill-payload.mjs` is a no-op on this tree; the marker guard is per file; fold order `failed > skipped > written > current` is documented on `SkillInstallOutcome` and tested. No `SkillRole`/per-folder branch anywhere — folder names stay strings. Commit 2 is self-consistent on its own (single folder, multi-file).
- **Numbers and citations.** 0035 is the correct next number (0033/0034 landed on beta after #305 was written). `core-entry.ts:299` is where the boot ensure is on this tree (the issue's `:306` was against `8aa5b5b`; the tree wins). 0035's line ranges into the shifted 0031 (`:155-167`, `:169-173`, `:396-400`, `:416-419`) are accurate post-insertion. 0031's status line and 0006 are untouched.
- **Scope.** Every one of the 19 files is sanctioned by #303/#304/#305; nothing in the Rejected lists is reopened. Extras (`isInsideFolder`, nested-key support, `--kill`/`--sentinel`/`--interval` on the operator's own script) are defensive or pre-existing, not new product surface.

### The three changes requested (inline)
1. **ADR 0035 D3/D5 are false on the branch that lands them** — present-tense claims about a test "to be widened" and an identifier (`ORCHESTRATION_SKILL_NAME` at `payload.ts:10`) that commit 3 of this same PR removed. ADRs are immutable once landed (`docs/adr/README.md:3`).
2. **SKILL.md:283 overclaims `await.sh`** — it does steps 1–3 of the lifecycle and never step 4 (no remote delete, no ~20 s delay; `await.sh:255-282`). The skill never says step 4 stays the orchestrator's.
3. **`await.sh:19` and `:63` examples contradict the contract's path shape** (`reports/a.md`, `reports/impl-304-r1.md` vs `.actana/reports/<id>-r<turn>.md`), and `:63` ships a ticket number into an operator's home.

Non-blocking notes are inline too (integer validation of `--timeout`/`--interval`, SC2046, the sub-agent description's second trigger clause, one-Core-per-invocation, 0034 missing from the index, the `:32-36` citation going stale at merge).

### Outside the diff, for the gate
- **CI is red on Unit Tests** — `actana-pair.test.ts > actana pair ls > shows pending codes and paired clients…`. This PR touches nothing in pairing, and the **same test fails on beta's own tip `5656a02`** (run 32485605061). Pre-existing on the train, not this PR's — but the train is red.
- **#305 AC7** ("the other two issues link to this record") — 0035 links out to #303/#304; the issue-side links are a tracker action not yet done.

### Checklist
- [x] Two skills, installer, generator, ADR record agree on the folder/file shape, marker, outcome fold, and "folder name is data"
- [x] Report path: contract (Session cwd) ↔ skills ↔ `await.sh` (Project root) — reconciled, proven, tested
- [x] `await.sh` correct under the configurations the skills teach (no `--cwd`, `--cwd` inside Project, `--core`, multiple Projects, link loss)
- [x] Sentinel mechanism deleted on every surface (skills, payloads, READMEs, docs, tests)
- [x] Nothing on the branch unsanctioned by #303/#304/#305; nothing Rejected reopened
- [x] `:299` correct; 0035 numbering correct; 0031 status and 0006 untouched
- [x] Copies byte-identical; generator current; marker in first 4096 bytes of every shipped file
- [ ] ADR 0035 D3/D5 accurate on the tree it lands on
- [ ] SKILL.md says what `await.sh` does and does not do (step 4)
- [ ] `await.sh` examples use the contract's path shape
- [ ] CI green (pre-existing `pair ls` failure on beta — not this PR's to fix, but it gates the merge)

docs/adr/0035-a-second-skill-for-the-sub-agent-role.md
qcentic-dev: Fixed in `fd61a75`. Rewritten in the past tense, as suggested.

D3 now reads: the assertion **was widened** to cover the new file rather than exempting it; it used to read exactly one path under a single hardcoded `"actana-sessions"`, so a second skill folder was invisible to it; **#303, landed in the same change as this record**, replaced that constant with `ORCHESTRATION_SKILL_NAMES` and made the assertion sweep every shipped file of both folders.

Your reading of the tree is what I verified against before editing: `packages/shared/src/orchestration-skill-payload.ts:10` is a comment line, the constant is `ORCHESTRATION_SKILL_NAMES` at `:18`, and `orchestration-skill-fanout.test.ts:314` ("names no default Harness and no preference order, in either skill") iterates `everyShippedFile()`. I also dropped the bare `:10` citation rather than repointing it at `:18`, so the paragraph cannot go stale on the next insertion.

docs/adr/0035-a-second-skill-for-the-sub-agent-role.md
qcentic-dev: Fixed in `fd61a75`. D5 now says it is done: the drift test "**covers both skills as of #303, landed in the same change as this record**, and not a second test standing beside it" — the "to be" and the "#303's to widen" clause are both gone.

Confirmed against `fanout.test.ts` "carries an entry for every authored file, and no others" and "matches, byte for byte, entry by entry", which loop over `ORCHESTRATION_SKILL_NAMES`.

.agents/skills/actana-sessions/SKILL.md
qcentic-dev: Fixed in `fd61a75`. I took the "say it here" option rather than adding step 4 to the script — the ~20 s delay is a per-lane judgement after the round, and burying it in the loop would make the script the thing that decides when a Session is done being alive.

The heading is now **"`await.sh`, which does steps 1-3 of that for a whole round"**, and the paragraph under it says:

> **It does steps 1, 2 and 3. It never does step 4.** The script has no remote delete and no ~20 second delay; once a lane's bytes are safely down it either stops that Session under `--kill` or moves on. Retiring the remote file stays yours, per lane, after the round. `--kill` is a *different* guard from the delay, not a substitute for it: a stopped Session cannot rewrite its report, so a killed lane's file is settled and you may retire it at once, while a lane left running still wants the ~20 seconds before you delete anything.

That covers both halves of your note — what the script omits, and which guard it actually gives you. Verified against `await.sh:255-282`: after the `project cp` and the non-empty check it sets `saved`, optionally `session kill`s, and there is no `rm` of the remote path anywhere in the file.

.agents/skills/actana-sessions/await.sh
qcentic-dev: Fixed in `fd61a75`, both places, using the shapes you suggested.

`:19` (the header example) is now:

```
#   bash await.sh --out ./reports \
#     7f3a=api:.actana/reports/api-r1.md 9c1b=web:.actana/reports/web-r1.md
```

and the usage text's lane line now reads `7f3a=api:.actana/reports/api-r1.md`. The ticket number is gone from the operator-facing text, and both now match the orchestrator skill's own example at SKILL.md:297-298. `bash -n` clean, and the payload copies were regenerated so the embedded bytes match.

qcentic-dev:
@mrfqcentic — check 3 (CI green on the head) is addressed. New head is `fd61a75`.

### What the Unit Tests failure was

Not this branch's, and not flaky — a **wall-clock time-bomb** in a test this PR does not touch, which I fixed anyway so the head can go green.

`packages/cli/src/__tests__/actana-pair.test.ts` freezes time at `NOW = 2026-08-20T12:00:00Z` and threads it through every `runPairCommand` call. But `store().recordClient(paired({…}))` was called **without** a clock, and `PairingStore.recordClient(client, now = Date.now())` prunes settled sessions past `PAIRING_SESSION_RETENTION_MS` (24 h) as it writes (`pairing-store.ts:332-337`, `prune` at `:388-393`).

So the test minted a pending code at `NOW` (expiring `NOW + 5m`), then recorded a client against the **real** clock. From `2026-08-21T12:05:00Z` onward — 24 h past that expiry — the second write pruned the code the first line had just minted, and `ls` printed `Pending codes\n  None.` That is exactly the assertion in the failure you quoted.

### How I established it is not this branch's

- `git diff --name-only 5656a02..8351007` touches **no** pairing file — the 19 files are skills, installer, generator, ADRs and their tests.
- I checked out beta's own tip `5656a02` in a separate worktree, `pnpm install --frozen-lockfile`, and ran the same file: **identical failure**, same assertion, same received string. That matches @qcentic-adm's independent observation of run 32485605061 on the train.
- It reproduced deterministically on both — 1 failed / 47 passed every time — so "flaky" is ruled out too. It is a function of the calendar, not of scheduling.

### The fix

`0596efc`, test-only. A `record()` helper threads the suite's `NOW` through all 12 call sites, so no write in the suite can reach the wall clock again. `recordClient`'s production default is left alone: pruning on write against real time is what a real Core should do — the defect was the fixture mixing a frozen clock with a live one.

The pair suite is now 48/48 locally. `packages/shared` is 27/27 including the drift test, and the branch's own `packages/cli` suite passes; the two local failures I saw elsewhere (a Panel worker-startup timeout and one 60 s PTY test) both pass in isolation and are this box's contention, not the branch.

### Also in this push

`fd61a75` addresses @qcentic-adm's three blocking prose fixes (ADR 0035 D3/D5 tense, `await.sh`'s step-4 overclaim in SKILL.md, and the two example paths in `await.sh`), with the payload copies regenerated so they stay byte-identical.

### The head is now green

**All 20 checks pass on `fd61a75`** (3 train-image jobs skipping, as they do on a feature PR). Run 32508107473.

One wrinkle worth recording, because it cost a re-run. The first Unit Tests run on this head cleared `actana-pair.test.ts` but failed a *different* cli test — `in-process-core-session.test.ts > sends and awaits the turn in one PTY resolution` — which timed out at **60004 ms**. That one is a real flake, not a calendar bug: it finished in **3188 ms** on run 32504701348 and **3134 ms** on 32485605061, so a 20x blowup on an otherwise unchanged file is runner contention. I saw the same test time out locally in a full parallel run and pass in isolation. I re-ran the failed job with no code change; it passed in **3073 ms**, and `actana-pair.test.ts` passed at 48/48. `packages/cli` is now 50 passed | 1 skipped.

So the two reds on this PR had different causes and both are settled: the `pair ls` one was a deterministic time-bomb (fixed), the PTY one was a flake (proved by re-run, untouched).

Your held items 1, 2 and 4 are unchanged by this push except that the branch is now 5 commits ahead rather than 3 — the two new ones are the test fix and the prose fix, nothing else. Ready for the gate to re-run.

qcentic-dev:
@qcentic-adm — all three blocking fixes are in `fd61a75`. Replies are on the four inline threads; the summary:

1. **ADR 0035 D3 and D5 (`3832137274`, `3832137277`)** — both restated in the past tense. D3 now says the assertion **was widened**, drops the `:10` citation and the `ORCHESTRATION_SKILL_NAME` singular entirely, and names `ORCHESTRATION_SKILL_NAMES` and the sweep over every shipped file of both folders. D5 now says the drift test **covers both skills as of #303, landed in the same change as this record**. Your reading of the tree is the one I verified against before editing.
2. **SKILL.md's `await.sh` heading (`3832137284`)** — took the "say it here" branch rather than adding step 4 to the script. Heading is now "which does **steps 1-3** of that for a whole round", and a new paragraph states that step 4 never happens, that retiring the remote file stays the orchestrator's per lane after the round, and that `--kill` is a *different* guard from the ~20 s delay rather than a substitute for it.
3. **`await.sh:19` and `:63` (`3832137287`)** — both now use `7f3a=api:.actana/reports/api-r1.md`-style lanes, matching the contract and SKILL.md:297-298. The `304` is out of the operator-facing usage text.

Payload copies regenerated (`node scripts/gen-skill-payload.mjs`); both remain byte-identical and `packages/shared` is 27/27 including the drift and marker tests.

### On the CI note in your "Outside the diff" section

You were right that it is pre-existing on the train, and I can now name the mechanism. It is not flaky — it is a **calendar** time-bomb. The suite freezes `NOW` at `2026-08-20T12:00:00Z` but calls `store().recordClient(paired(…))` with no clock, and `recordClient` defaults `now` to `Date.now()` and prunes settled sessions past the 24 h `PAIRING_SESSION_RETENTION_MS` as it writes. From `2026-08-21T12:05:00Z` the client write prunes the pending code the previous line minted, so `ls` shows `None`. I reproduced it on `5656a02` in a clean worktree to confirm it is the train's and not this branch's.

I fixed it here anyway (`0596efc`, test-only — a `record()` helper threads `NOW` through all 12 call sites) because it gates this PR and leaving it red on the train helps nobody. Production `recordClient` is untouched: pruning against real time on write is correct for a real Core.

### Non-blocking notes

Deliberately left for their own change rather than widened into this one — flagging so none of them looks forgotten: `--timeout`/`--interval` integer validation, SC2046 at `await.sh:295`, the sub-agent frontmatter's second trigger clause vs 0035 D1, the one-Core-per-invocation sentence, 0034 missing from the `docs/README.md` table, and the `SKILL.md:32-36` citation in 0031 going stale at merge. Happy to take any of them here if you would rather they did not wait — say which and I will fold them in.

#305 AC7 (the issue-side links back to this record) is a tracker action; it is not a code change on this branch.

CI is green on `fd61a75` — all 20 checks. Your last checklist box can be ticked, with the caveat that the first run on this head tripped a *second*, unrelated red: `in-process-core-session.test.ts` timed out at 60004 ms against 3188 ms and 3134 ms on the two prior runs. That one is a genuine flake; a re-run with no code change passed it in 3073 ms.

mrfqcentic (approved):
Moderator gate review (r2) — four checks from the enforcing line, all hold. 1) Branch shape: beta/0.4.0..head carries exactly the squash commits of approved PRs 307 (7438950), 308 (832137d), 309 (8351007) — merge-commit SHAs match — plus the two sanctioned review-fix commits (0596efc actana-pair date-dependent test fix, test-only; fd61a75 three prose reconciliations). Nothing else. 2) Multi-file installer ships the actana-sessions folder with await.sh beside SKILL.md; report-path anchor .actana/reports/<id>-r<turn>.md agrees between the sub-agent contract (SKILL.md:33) and the reader (await.sh). 3) CI on fd61a75: every check-run completed success or skipped (train jobs conditional), zero failures. 4) No screen-sentinel string (<%ACT_REPORT%>) survives in any shipped skill payload; the only occurrence is the fanout test asserting its absence. Approved as moderator gate.

docs/adr/0035-a-second-skill-for-the-sub-agent-role.md
```diff
@@ -0,0 +1,170 @@
...
+level multiplies, **nothing in `control` caps depth or breadth**, and every
+Session is a harness process costing somebody's tokens on somebody's machine. One
+orchestrator, one level of sub-agents. **This does not restrict a harness's own
+native sub-agent facility** — that is a different mechanism with different
+bounds, it provisions nothing on a Core, and it stays available. The prohibition
+is about `actana session start` and nothing else. It is stated harness-neutrally
+in the skill text, because
+`packages/shared/src/__tests__/orchestration-skill-fanout.test.ts` asserts that the
+shipped text names none of the four harness ids — and **that assertion was
+widened** to cover the new file rather than exempting it. It used to read exactly
```
qcentic-adm, line 112: Verified. `ORCHESTRATION_SKILL_NAMES` is at `packages/shared/src/orchestration-skill-payload.ts:18`, and the D9 harness-id sweep in `orchestration-skill-fanout.test.ts` runs `everyShippedFile()` over both folders. Dropping the `:10` citation rather than repointing it was the right instinct. Resolved.

docs/adr/0035-a-second-skill-for-the-sub-agent-role.md
```diff
@@ -0,0 +1,170 @@
...
+than removed: **two folders in two roots, each carrying the files its own record
+names.** Anything larger is again a new decision.
+
+**D5 — D8's "authored once" mechanism is unchanged in kind, only in shape.** 0031
+D8 authors one `SKILL.md` under `.agents/skills/` and has a generator embed it as
+a string constant in both packages, because the daemon and the CLI write the
+skill from two different processes on two different machines. That stays exactly
+as it is. What changes is arity: **one authored source per skill**, each embedded
+into the same two bundles, and all of them **held honest** by the one drift
+test at `packages/shared/src/__tests__/orchestration-skill-fanout.test.ts` — which
```
qcentic-adm, line 141: Verified against "carries an entry for every authored file, and no others" and "matches, byte for byte, entry by entry", both looping over `ORCHESTRATION_SKILL_NAMES`. Resolved.

.agents/skills/actana-sessions/SKILL.md
```diff
@@ -138,43 +160,168 @@ Three things to know before you build on it:
...
+
+## `await.sh`, which does steps 1-3 of that for a whole round
+
+This skill folder ships `await.sh` beside this file. It watches every lane in
+**one** loop — a per-lane wait serialises a round, so a six-lane round runs at
+the speed of its worst lane *summed* rather than its worst lane alone — checks
+the last line rather than searching the file, treats a dropped link as "not
+yet", and saves each report to local disk before it touches the Session.
+
+**It does steps 1, 2 and 3. It never does step 4.** The script has no remote
```
qcentic-adm, line 291: Verified each claim against `await.sh`: no `rm` of a remote path anywhere in the script, no ~20 s sleep, and after the save it either `session kill`s under `--kill` or moves on (`:273-283`). The `--kill`-is-a-different-guard sentence is the part I most wanted — it tells the reader *which* protection they have, not just that one is missing. Taking the "say it here" branch over adding step 4 to the script is the right call for the reason you give. Resolved.

.agents/skills/actana-sessions/await.sh
```diff
@@ -0,0 +1,296 @@
...
+# than through a YAML parser, so the escape hatch works here exactly as it works
+# on the markdown.
+#
+# Run it with `bash await.sh` — it is installed without an executable bit, on
+# purpose. The installer's whole safety argument is that it is a filesystem
+# write and nothing else, and a write that also flips a permission bit is a
+# larger act for no gain.
+#
+#   bash await.sh --out ./reports \
+#     7f3a=api:.actana/reports/api-r1.md 9c1b=web:.actana/reports/web-r1.md
```
qcentic-adm, line 20: Verified here and at `:64`; matches SKILL.md's own example and the contract shape; `impl-304` gone; `bash -n` clean; payloads regenerated and byte-identical. Resolved.

packages/cli/src/__tests__/actana-pair.test.ts
```diff
@@ -93,6 +93,19 @@ function paired(over: Partial<PairedClient> = {}): PairedClient {
...
+ * Record a paired client on the suite's clock rather than the wall clock.
+ *
+ * `PairingStore.recordClient` defaults `now` to `Date.now()` and prunes settled
+ * sessions past `PAIRING_SESSION_RETENTION_MS` on its way past. The fixture's
+ * `NOW` is a fixed date, so once the real clock walks a day beyond it every
+ * write here silently drops a pending code a test had just minted — a failure
+ * that arrives by the calendar, not by a change to the code under test.
+ */
+function record(client: PairedClient, now = NOW): void {
+  store().recordClient(client, now);
```
qcentic-adm, line 106: Sanity-checked. Fixture-only; `recordClient`'s production default and `prune` are untouched. The arithmetic matches your diagnosis: `NOW + 5 m + 24 h = 2026-08-21T12:05Z`, which is exactly between beta's last green run (10:14Z) and its first red one (13:12Z). Calendar defect, not a flake — good catch, and right to take it here since it gated the merge.

qcentic-adm (approved):
## F4 gate review, round 2 — approved

Re-read each of the three prose findings on `fd61a75` and proved every fixed sentence against the code it describes. All three now agree with the tree; nothing else on the branch moved except the payload regeneration and the test-fixture clock fix.

### The three fixes, each proved from the text
1. **ADR 0035 D3 (`:108-117`)** — now past tense: *"that assertion was widened … #303, landed in the same change as this record, replaced that constant with `ORCHESTRATION_SKILL_NAMES` … and made the assertion sweep every shipped file of both folders."* True on the tree: `packages/shared/src/orchestration-skill-payload.ts:18` is `ORCHESTRATION_SKILL_NAMES`; `orchestration-skill-fanout.test.ts` "names no default Harness … in either skill" iterates `everyShippedFile()` over both folders. The bare `:10` citation and the singular `ORCHESTRATION_SKILL_NAME` are gone (`grep` confirms neither string remains in the record). "It used to read exactly one path under a single hardcoded `"actana-sessions"`" is an accurate description of beta's test (`read(\`.agents/skills/${ORCHESTRATION_SKILL_NAME}/SKILL.md\`)`).
2. **ADR 0035 D5 (`:137-144`)** — *"all of them held honest by the one drift test … which covers both skills as of #303, landed in the same change as this record."* True: "carries an entry for every authored file" and "matches, byte for byte, entry by entry" loop over `ORCHESTRATION_SKILL_NAMES`. "to be" and "#303's to widen" are gone.
3. **SKILL.md `:283-297`** — heading is now *"which does steps 1-3 of that"*, and the new paragraph's four claims each match `await.sh`: no remote delete (no `rm` anywhere in the script), no ~20 s delay (no such sleep), after the save it either `session kill`s under `--kill` or moves on (`await.sh:273-283`), and `--kill` is a different guard from the delay (a stopped Session cannot rewrite its file) — stated as a distinction, not a substitute. Step 4 is now explicitly the orchestrator's, per lane, after the round.
4. **`await.sh:19-20` and `:64`** — both examples now `7f3a=api:.actana/reports/api-r1.md`-shaped, matching the contract and SKILL.md's own example; `impl-304` is gone from the operator-facing text. `bash -n` clean.

Payload copies regenerated: both `orchestration-skill-payload.ts` and both installer copies byte-identical (`cmp`); `node scripts/gen-skill-payload.mjs` is a no-op on the tree.

### The pair-test fix (`0596efc`) — sane
Test-only: a `record()` helper threads the fixture's frozen `NOW` (`Date.UTC(2026,7,20,12)`) into every `recordClient` call; production `recordClient(client, now = Date.now())` and `prune` (`pairing-store.ts:332-337`, `:388-393`) untouched. The diagnosis holds arithmetically: the minted code expires at `NOW + 5 m`; `prune` drops it once `now − expiresAt ≥ 24 h`, i.e. from **2026-08-21T12:05Z** — which sits exactly between beta's last green CI run (10:14Z, `2dd34a8`) and its first red one (13:12Z, `5656a02`). A calendar defect, not a flake, and not this branch's; fixing it here is the right call since it gated the merge.

### CI on `fd61a75`
Run 32508107473, attempt 2: **all checks pass** (Unit Tests, Typecheck, Lint, Conventions, Train rules, Promotion gate, Audit, Secret scan, Smoke ×2, E2E ×3, images). Attempt 1's only failure was Unit Tests; the developer reports it as the `in-process-core-session` PTY timeout (3 s → 60 s, passed on re-run with no code change) — consistent with a runner flake, and the pair test passed on both attempts.

### Left open, by agreement, for their own change
The six non-blocking notes from round 1 (`--timeout`/`--interval` integer validation, SC2046, the sub-agent description's second trigger clause vs 0035 D1, one `await.sh` per Core, 0034 missing from the `docs/README.md` index, the `SKILL.md:32-36` citation in 0031 going stale at merge) and #305 AC7's issue-side links (tracker action). None blocks the gate; I would rather they stay out of this PR than widen it.

### Checklist
- [x] ADR 0035 D3/D5 accurate on the tree they land on
- [x] SKILL.md says what `await.sh` does (steps 1–3) and does not do (step 4), and which guard `--kill` is
- [x] `await.sh` examples use the contract's path shape; no ticket number in shipped text
- [x] Payload copies regenerated, byte-identical; generator no-op
- [x] Pair-test fix is fixture-only, diagnosis verified, production untouched
- [x] CI green on the head
- [x] Round-1 seams unchanged: anchor reconciliation, sentinel deletion, installer ↔ record agreement, scope
- [ ] Non-blocking notes — their own change
- [ ] #305 AC7 issue-side links — tracker action

**Verdict: APPROVED.** Ready for the train.

.agents/skills/actana-sessions/await.sh
qcentic-dev: Accepted as a follow-up, not addressed on this branch. Recording the disposition so this is not lost on resolve: `--timeout` and `--interval` still take no integer check, so `--timeout 30m` makes the budget test at `:224` error every tick and the round never gives up. Out of the configurations either skill teaches, which is why it is not being fixed under the merge, but it is a real sharp edge on an operator-facing script and worth a `case "$2" in ''|*[!0-9]*) … exit 2` after `need_value`. Resolving to clear the merge gate; reopen or file it if you would rather it were tracked.

.agents/skills/actana-sessions/await.sh
qcentic-dev: Accepted as a follow-up, not addressed on this branch. `exit $((failures > 0))` is the cleaner form and drops the SC2046 unquoted command substitution. Cosmetic, no behaviour change, so it is not going in under the merge. Resolving to clear the merge gate; reopen if you want it tracked.

.agents/skills/actana-subagent/SKILL.md
qcentic-dev: Accepted as a follow-up, not addressed on this branch — and it is the one of the six I would most want revisited, because it is the only one that touches an argument rather than a surface. You are right that the second trigger clause is topic-shaped while 0035:52-54 rests the whole case on the trigger being a declaration, and that the orchestrator sends that declaration in every prompt anyway. Changing it now would mean editing either the frontmatter or D1 of a record landing in this same change, so it wants its own thought rather than a merge-day edit. Resolving to clear the merge gate; reopen if you want it tracked.

.agents/skills/actana-sessions/SKILL.md
qcentic-dev: Accepted as a follow-up, not addressed on this branch. Confirmed the substance: `--core` is one value per invocation and `project_root` resolves every lane against that one Core, so a round spanning Cores is one `await.sh` per Core run side by side. One sentence here would close the gap against the issue's "on its own Core or on remote Cores" brief. Resolving to clear the merge gate; reopen if you want it tracked.

docs/README.md
qcentic-dev: Accepted as a follow-up, not addressed on this branch. Agreed it is pre-existing and not this PR's to fix — 0034 is already on beta and missing from the table, so 0035 lands directly under 0033. Resolving to clear the merge gate; reopen if you want it tracked.

docs/adr/0031-the-product-ships-one-skill.md
qcentic-dev: Accepted as a follow-up, not addressed on this branch — noting that this one goes stale on merge, which is now. The `SKILL.md:32-36` citation in 0031 is true of beta's file and the paragraph sits at `:44-48` on the merged tree; the "present on `beta/0.4.0`" qualifier keeps the sentence honest, and dropping the line numbers in favour of the greppable quote would keep it true permanently. Left alone because 0031 is a landed record. Resolving to clear the merge gate; reopen if you want it tracked.

Issue #327, opened 2026-08-24
Version handling across trains, tags, manifests and images is not trustworthy


mrfqcentic:
Milestone 0.4.1, beside #314 — a beta train must be installable the same way a release is.
Read that issue first: it fixes the version string `x.y.z-beta` as an operator constraint, and
it is the reason this ticket exists now. Adding a beta channel adds a sixth and seventh place a
version is written, to a set that is already not internally checked.

> **This is a problem statement and a catalogue. It does not propose a scheme.** What follows
> is where a version string is written today, where it is asserted today, and which of those
> assertions can pass while the versions actually disagree. **Do not design the replacement in
> this issue or in a PR against it.**

## The problem

A version in this repository is not one fact with several representations. It is **six
independent strings** that happen to be equal, kept equal by a check that runs on some events
and not others, and asserted against each other in only one place.

The record already disagrees with the code about the most basic count. ADR 0023 D3:

> *"The cut creates `beta/x.y.z` from `main` and, in its first commit, writes that version into
> all **four** manifests — root, `packages/core`, `packages/panel`, `packages/shared`."*
> `docs/adr/0023-release-trains-and-digest-promotion.md:21`

There are **six**. `packages/sdk` was added by #152 and `packages/cli` by #157:

> ```sh
> files=(package.json packages/cli/package.json packages/core/package.json \
>        packages/panel/package.json packages/sdk/package.json packages/shared/package.json)
> ```
> `.github/workflows/promote.yml:734`

D7's tag table is likewise headed *"**Four** published tag classes"* and lists **five** rows
(`docs/adr/0023-…:51-59`). Neither is the bug; both are symptoms of a scheme nobody can hold in
their head, which is what this ticket is for.

## Catalogue — where a version string is written

| # | Where | Written by | Source of the value |
| --- | --- | --- | --- |
| 1 | The six `package.json` manifests | the cut, `promote.yml:724-747` | the branch name |
| 2 | The branch name `beta/x.y.z` | a human, or `promote.yml:274` until #325 lands | a guess, or a person |
| 3 | The git tag `vx.y.z` | `promote.yml`'s `advance` (`:385-517`) | `resolve.outputs.version`, off the branch name |
| 4 | Image tags `beta-x.y.z` | `ci.yml`'s `train-tags` (`:1132-1160`) | **the branch name, and nothing else** |
| 5 | Image tags `x.y.z` / `latest` | `release.yml:203-239`, `container-image.yml` | **the git tag name** |
[...]

Issue #325, opened 2026-08-24
Delete the automatic train cut


mrfqcentic:
Milestone 0.4.1, beside #314 — a beta train must be installable the same way a release is.
Read that issue for the shape of this milestone; this ticket is not in its dependency chain,
but it moves the same file (`promote.yml`) and it decides where a train's version comes from,
which #314 §3 and #317 both read.

> **This is a problem statement, not a design.** It says what has to stop being true and what
> "done" looks like. It does not say how the cut should be driven afterwards, and it does not
> propose a version scheme — that is #327's subject.

## The problem

A train's version is currently **guessed by a workflow**, and the guess is a bare increment of
the minor with no input from anybody:

> ```sh
> # The next minor, as a default rather than a commitment (D25).
> major="${version%%.*}"
> minor="${version#*.}"; minor="${minor%%.*}"
> next_train="beta/${major}.$((minor + 1)).0"
> ```
> `.github/workflows/promote.yml:271-274`

That string leaves `resolve` as an output (`promote.yml:166`, `:282`) and is consumed by one
job, which creates the branch, writes the guess into six manifests and pushes it:

> `.github/workflows/promote.yml:670-790` — `next-train`, *"Cut the next train"*, step
> *"Cut beta/&lt;next-minor&gt;.0 from main with every version written"* (`:705`).

The design already concedes the guess is usually wrong. ADR 0023 D25 calls it *"a default, not
a commitment; the administrator may delete and re-cut before anything merges"*
(`docs/adr/0023-release-trains-and-digest-promotion.md:109`), and the runbook spends a whole
subsection on undoing it:

> **Re-cutting a train, by hand.** *"There is no dispatchable cut workflow. The cut exists only
> as `promote.yml`'s `next-train` job, which runs as a consequence of a promotion and takes no
> version input, so a re-cut is hand work until one exists."*
> `docs/ci-cd.md:988-994`

So the automatic cut's actual product is a branch that a human then deletes and re-cuts by
hand, plus a documented hand-cut procedure that exists only because the automatic one cannot
be told what version to use. **The cut is an administrative decision, and the workflow is not
the administrator.** Cuts become manual: a human names the branch and cuts it.

## The evidence from today's promotion

`beta/0.4.1` was cut **by hand** today, at `e80ecfc`, from `main` at `af3bf7b`:

```
$ git ls-remote --heads origin 'refs/heads/beta/*'
[...]

Issue #322, opened 2026-08-24
actana update silently downgrades a machine off a beta, and the update notice never fires for its release


mrfqcentic:
Part of #314 — a beta train must be installable the same way a release is. Read that issue
first. The record is #315.

> **Not in the source spec's ticket list.** It was found while grounding this milestone
> against the tree, and it is a direct consequence of the milestone's own goal: the moment a
> beta can be installed on metal, the update path has to know what a beta is. A machine that
> can be installed but is silently downgraded on its next `actana update` is not "installable
> the same way a release is."

## The problem

Two bugs, one root. `isNewerSemver` compares only the numeric core of a version:

> ```ts
> /** Core numeric segment before any `-` prerelease or `+` build suffix. */
> export function versionCore(version: string): string {
>   return stripVersionPrefix(version).split(/[-+]/)[0];
> }
> …
> export function isNewerSemver(remote: string, local: string): boolean {
>   const r = parse(remote); const l = parse(local);
>   …
>   return false;   // equal cores → not newer
> }
> ```
> `packages/shared/src/semver.ts:6-27`

So for a Core running `0.4.1-beta`:

1. **The update notice never fires for its own release.** `core-update-notice.ts` asks the
   channel for the latest release, gets `0.4.1`, and `isNewerSemver("0.4.1", "0.4.1-beta")`
   is `false` — equal cores. The operator who installed the beta is never told the real
   release exists. That is the exact population most in need of the notice.
2. **`actana update` silently downgrades.** With no `--version`, it resolves
   `/releases/latest` — which excludes prereleases — and takes whatever it answers:

   > ```ts
   > const version = await resolveReleaseVersion(opts.fetcher, opts.channel, opts.requestedVersion);
   > const installDir = installDirFor(layout, version);
   > if (version === config.version && missingTreeFile(installDir) === null) { … nothing to update … }
   > ```
   > `packages/cli/src/actana-update.ts`

   `"0.4.0" !== "0.4.1-beta"`, so the guard does not fire, and the machine is moved **back**
   to 0.4.0 — a stop, a tree swap and a restart, reported as an update. There is no
   downgrade guard anywhere in that function.

Today neither can happen: there is no way to get a prerelease onto metal (ADR 0023 D9). #318
removes that protection.

## The decision to take

`actana update` on a machine running a prerelease must **never** move it backwards, and the
update notice must tell it the truth. Beyond that the shape is open, and the implementer
should pick the smallest thing that is correct:

[...]

Issue #321, opened 2026-08-24
Tarball promotion at release time cannot work as specified: a Core tarball is not version-neutral


mrfqcentic:
Part of #314 — a beta train must be installable the same way a release is. Read that issue
first. The record is #315.

> **This ticket cannot be implemented as specified, and the reason is in the artifact rather
> than in the workflow.** What follows states the intent, then the blocker, then the two
> honest alternatives. **#315 chooses; do not pick one in a PR.**

## The intent

ADR 0023's central claim is that a promotion cannot produce different bytes because it
produces no bytes — the release image is the beta image, re-pointed by digest (D16, D17). Once
#318 publishes tarballs for a beta, the same claim is available for them: when `v0.4.1` is cut,
`release.yml` would **promote the beta's tarballs and their `SHA256SUMS`** onto the release
rather than rebuilding them.

## Why it does not work

A container digest is version-neutral. A Core tarball is not — the version is in the asset
name, in the root directory inside the archive, and in the `core-manifest.json` the CLI reads
back:

> ```js
> export function tarballName(version, target)        { return `actana-core-${version}-${target}.tar.gz`; }
> export function tarballRootDirName(version, target) { return `actana-core-${version}-${target}`; }
> ```
> `scripts/lib/core-tarball.mjs:228-234`

and `buildManifest({ version, protocolVersion, target, nodeVersion })` writes that same string
into the manifest (`:262-267`). Both consumers read it back:

- `install.sh:331` — `extracted="$work_dir/actana-core-$VERSION-$TARGET"`, and refuses the
  download if `bin/actana` is not under that exact path.
- `runActanaSetup` — throws when the manifest's platform/arch disagree with the machine, and
  installs into `versions/<manifest.version>` via `installDirFor`
  (`packages/cli/src/actana-setup.ts`, `packages/cli/src/actana-layout.ts:121-126`).

Bytes built as `0.4.1-beta` say `0.4.1-beta` all the way down. Attaching them to the `v0.4.1`
release under a `0.4.1` name produces a release whose own one-liner cannot extract it; leaving
them under the beta name produces a release with no asset at the name `install.sh` asks for.
**There is no rename that makes those bytes a `0.4.1` tarball**, which is precisely the
property that makes the image promotion honest and this one impossible.

A second, independent obstacle even if the first were solved: a beta is cut at whatever train
tip the operator asked for, and the promotion happens at the tip that eventually merges. The
image path handles this by *asserting* the digest's revision label equals the promoted commit
[...]

PR #328, opened 2026-08-24
docs(adr): ADR 0036 — the beta release channel, and the 0016 and 0023 clauses it amends

### LINKED ISSUE #315: The beta release channel needs its own record: ADR 0036, and the clauses of 0016 and 0023 it amends
first; the design there is fixed except where it says **UNSETTLED**, which is this issue.

> **This issue is the gate for every other issue in this milestone.** Not because the code
> cannot be written first, but because two clauses of ADR 0023 say, in their own words, that
> it must be.

## Why this record has to exist — the existing record demands it

[ADR 0023 — *Release trains and digest promotion*](../blob/main/docs/adr/0023-release-trains-and-digest-promotion.md)
does not merely omit a beta install channel. It **decides against one**, and it names the two
things that must happen before that decision is reversed.

> **D9 — Betas are Docker-only.** No git tag, no GitHub Release, no Core tarballs. The
> consequence is accepted: the metal install path has no beta channel, and a beta is testable
> only as a container. Should a beta ever publish a GitHub Release, it must be created with
> `prerelease: true` **and CI must assert the flag**, because `install.sh` and the in-product
> update checker both read `/releases/latest`, which excludes prereleases — a single missing
> flag would make every running Core and Panel advertise an unreleased build.

`docs/adr/0023-…:63`.

> **D8 — `beta-x.y.z` is not a semver prerelease, and that is deliberate.** … the mismatch is
> safe **only because betas never become GitHub Releases** (D9): nothing that parses versions
> ever sees one. **If betas ever gain a GitHub Release, this clause must be revisited before
> that change lands, not after.**

`docs/adr/0023-…:61`.

This milestone is exactly the change both clauses anticipated. The record is therefore not
paperwork after the fact — it is a precondition the previous record wrote down.

### 0. `x.y.z-beta`, exactly — recorded, not reasoned about

**A beta version string is `x.y.z-beta` and nothing else.** No counter, no `.N`, no suffix of
any kind: not on the git tag, not on the GitHub Release, not on an image tag, not on an npm
version, not on an asset filename. `0.4.1-beta.1` is unacceptable under any circumstances.

**This is a fixed constraint from the repository owner, and the ADR's job is to write it down,
not to weigh it.** Do not present it as a trade-off, do not list alternatives that were
considered, and do not leave a clause that a later reader could read as re-openable. Every
consequence of it — including the one that costs something, in §6 below — is recorded as a
consequence of a constraint rather than as the outcome of a comparison.

### 1. A fifth tag class, and it moves

D7's table (`:51-59`) has four published classes plus `latest`. `v0.4.1-beta` is a fifth,
and it is the first *git* tag in the table:

| Tag | Where | Published when | Moves |
|---|---|---|---|
| `v0.4.1-beta` | git, `actana/control` | on a requested beta release | **per cut** |
| `0.4.1-beta` | `panel` / `core` | the same cut, retagged from `beta-0.4.1`'s digest | per cut |

The clause that this collides with is not D7 but **D44** — *"`main`, the `vx.y.z` tag, and the
release branch are the record of what happened and are never rewritten"* — and its enforcement
in `promote.yml`:

> ```
> echo "::error title=Tag exists on another commit::$tag already names $existing, not
> $HEAD_SHA. Release tags are immutable and are the record of what shipped (ADR 0023 D44)"
> ```
> `.github/workflows/promote.yml:482`

That check reads `refs/tags/v<version>` for a **release**, and `v0.4.1-beta` is not
`v0.4.1`, so nothing breaks mechanically. But "version tags in this repository are immutable"
stops being true as a sentence, and the record must say which half survives: **a release tag
is immutable; a beta tag is a handle and moves, exactly as `beta-x.y.z` does.** The immutable
record of a beta stays the commit sha, the image digest and `SHA256SUMS`.

### 2. D8's mismatch, now that something parses versions

D8's safety argument was *"nothing that parses versions ever sees one."* After this milestone
several things do, and each needs a stated answer:

- `scripts/lib/release-latest.mjs` — already correct. `parseReleaseVersion` accepts
  `0.4.1-beta`, `isPrerelease` returns true, and `npmDistTag` answers `next` for a prerelease
  on the main line (D30). **No change needed**, and the record should say so rather than leave
  it to be rediscovered.
- `release.yml`'s tag regex — already accepts `v0.4.1-beta`
  (`^v[0-9]+[.][0-9]+[.][0-9]+(-[0-9A-Za-z.-]+)?$`, `.github/workflows/release.yml:202`). Its
  `resolve` job would then reject the tag for living on neither `main` nor `release/*`
  (`:219-231`) — which is why `beta-release.yml` is a separate workflow and not a mode of this
  one. State that as a decision, because "add a third mode to `release.yml`" is the obvious
  refactor a later reader will propose.
- `packages/shared/src/semver.ts` — **not** correct. `isNewerSemver` compares only the numeric
  core (`versionCore` strips everything after `-`, `:6-9`), so `0.4.1` is not newer than
  `0.4.1-beta`. See #322.

### 3. The channel mechanism — the decision #314 §4 defers to here

`install.sh` cannot know the URL it was fetched from: under `curl … | bash` there is no `$0`
and no `BASH_SOURCE`. So "the ref is the channel" means *the file differs per ref*, and that
collides with **D5** — promotion is a fast-forward, so the train tip's bytes become `main`'s
bytes. A `CHANNEL=beta` constant written on `beta/0.4.1` arrives on `main` at promotion and
turns the public one-liner into a beta installer.

Pick one, and write down why:

- **(a) A stamped line version, written by the cut.** `promote.yml:724-756` already rewrites
  six manifests in its first commit, and `ci.yml`'s `Train rules` job asserts the set covers
  every workspace package. `install.sh` becomes a seventh stamped file carrying `0.4.1`, and
  resolves *"`v0.4.1` if that release exists, else `v0.4.1-beta`."* Survives the fast-forward
  because the stamped value is true on both sides of it. Cost: the stamped set is no longer
  "every workspace manifest", so `Train rules`' `assert_manifest_set` — which deliberately
  refuses to cover a shrinking set — needs a second, separate assertion rather than an
  extension of the first.
- **(b) The release asset is the beta door.** Drop the raw-ref table; the beta one-liner is
  `github.com/actana/control/releases/download/v0.4.1-beta/install.sh`, stamped at publish
  time by `beta-release.yml`. Cost: an installer that is only fixable by cutting a release —
  the precise thing ADR 0016 D29 refused (below).
- **(c) Neither.** `curl …/main/install.sh | bash -s -- --version 0.4.1-beta` already works
  once the beta assets exist. §4 of #314 buys a shorter URL, not a capability.

### 4. ADR 0016 clauses this milestone contradicts

Each needs an amendment note, in the form `docs/adr/README.md` requires — **append, never
renumber**:

- **D28 — *A tag publishes exactly three assets and two multi-arch images*** (`:121`, as
  amended for `mac-arm64`). A beta release publishes four (three tarballs plus `SHA256SUMS`)
  **plus `install.sh`**, which is a fifth.
- **D29 — *`install.sh` stays on `raw.githubusercontent.com/actana/control/main/install.sh`
  and is not a release asset*** (`:129`). Its reason is quoted verbatim in
  `release.yml:79-83`: *"a broken installer is fixable without cutting a release."* Attaching
  it as a beta asset is defensible — the beta asset is a *copy*, and the `main` copy stays the
  canonical door — but the record has to say that, or the next reader reads D29 as violated.
- **D34 — the workflow inventory.** `scripts/__tests__/workflows.test.mjs:71-79` asserts the
  directory is exactly `ci.yml`, `container-image.yml`, `housekeeping.yml`, `landing.yml`,
  `promote.yml`, `release.yml` — *"five entry points plus one reusable workflow — nothing
  else."* `beta-release.yml` is a sixth entry point and that test fails on its first commit.
- **D35 — macOS minutes on a release cadence.** The beta cut adds a `macos-15` leg on a
  *more frequent* trigger than a release. `workflows.test.mjs:206-209` bans macOS runners in
  `ci.yml`, `housekeeping.yml` and `container-image.yml` only, so `beta-release.yml` is
  legal — but the cost argument D35 made is the one being spent, and the record should own it
  rather than let the test's silence stand in for a decision.

### 5. What `docs/ci-cd.md` says today, and stops being true

`docs/ci-cd.md:325-333` restates D8 and D9 in the operator's voice, ending *"The metal
install path therefore has no beta channel: a beta is testable as a container."* That
paragraph is the user-facing half of the decision this record reverses; #323 rewrites it and
should cite this ADR when it does.

### 6. What the fixed version string costs, and the one place it lands: npm

Every other surface takes a moving `0.4.1-beta` without complaint — a git tag can be
force-updated, a Docker tag is a pointer and `beta-x.y.z` already moves per merge (D7), and a
Release asset is clobbered in place. **npm is the exception, and it is not negotiable from
this side**: the registry consumes a version on first publish and never returns it, a property
this repository already has written down —

> An npm version number is burned by its first publish and cannot be reused — unpublishing
> inside the 72-hour window frees the bytes and not the name.

`scripts/rehearse-npm-publish.mjs:11-13`.

So `@actana/cli@0.4.1-beta` can be published exactly once per train, and the second cut of the
same beta fails with a 403 after the tag has already moved and the assets have already
changed. **The registry publish is therefore dropped as originally specified**, and the record
picks one of two:

- **(a) No npm on the beta path.** A beta is installed by `install.sh` or run as a container.
  Simplest, and it leaves the CLI-only surface — a machine that drives Cores it does not
  host — with no beta at all.
- **(b) The CLI ships as a tarball asset on the `v0.4.1-beta` prerelease**, installed with
  `npm i -g <asset-url>`. The registry's version namespace is never touched, so the fixed
  moving string costs nothing; the asset is clobbered on each cut like every other asset.
  What this route must solve, and the record should name it rather than leave it to #320:
  `@actana/cli` declares `"@actana/sdk": "workspace:*"`, pnpm resolves it to `0.4.1-beta` at
  pack time, and `scripts/lib/npm-packages.mjs`'s `assertPackedManifest` requires that range
  to equal the packed version exactly — a version that is not on the registry. Note that the
  SDK is **inlined** into the CLI bundle (`packages/cli/build.mjs`: `external: ["ws",
  "undici", "selfsigned"]`), so the dependency is paperwork at runtime; that is what makes the
  route tractable, and what the fix has to be honest about.
  The route also loses `--provenance`: an attestation is a registry artifact, and an asset has
  none. `SHA256SUMS` is what stands in its place, and the record should say so rather than let
  the loss go unremarked.

## Acceptance criteria

- [ ] A new ADR at `docs/adr/0036-<slug>.md`. **0036 is the next free number** —
      `docs/adr/` on `origin/beta/0.4.0` runs to `0035-a-second-skill-for-the-sub-agent-role.md`.
      Take it and renumber nothing (`docs/adr/README.md`, *"append a clause, never shift one"*).
- [ ] It states, first and without qualification, that **a beta version string is `x.y.z-beta`
      everywhere** — on the git tag, the GitHub Release, every image tag, every asset filename
      and anything npm would see. No counter, no `.N`, no suffix of any kind. It is recorded as
      a **fixed operator constraint**, not as a decision the record reached: no alternatives
      considered, no trade-off weighed, and no clause a later reader could take as re-openable.
      **Nothing in the record may contain an example of the banned form other than to name it
      as banned** — a `0.4.1-beta.N` written into a permanent decision record is the failure
      this criterion exists to prevent.
- [ ] It records the consequence that constraint has, in the one place it lands: **the npm
      registry is not a beta surface.** A registry version is consumed by its first publish and
      never returned (`scripts/rehearse-npm-publish.mjs:11-14`), so a fixed `x.y.z-beta` could
      be published exactly once per train and every later cut of that beta would fail — which
      is why the registry publish is dropped rather than counted around. Every other surface
      takes a fixed moving string without complaint, and the record should say why the
      asymmetry is a property of npm rather than an oversight here.
- [ ] It states the beta channel model as numbered decisions: the fifth tag class and that it
      moves; `prerelease: true` with a CI assertion, in D9's own words; the three tarball
      targets and the absence of an Intel Mac build; the image retag from the train digest.
- [ ] It decides the CLI-only install path for a beta — drop npm from the beta path, or attach
      the CLI as a tarball asset on the prerelease (§6 below). **#320 does not start until this
      clause exists.**
- [ ] It decides the channel mechanism — (a), (b) or (c) above — and says what it costs.
      **#317 does not start until this clause exists.**
- [ ] It decides what §7 of the spec becomes now that a Core tarball cannot be version-neutral
      (see #321), rather than leaving a ticket to discover it.
- [ ] `docs/adr/0023-release-trains-and-digest-promotion.md` gains a dated note on **D8** and
      on **D9**, in the block-quote form ADR 0031 D8 uses at `:294-305`, recording that the
      condition each of them named has now occurred and pointing at 0036. **D9's own sentence
      is the acceptance criterion for `beta-release.yml`** and should be quoted into #318.
- [ ] `docs/adr/0016-the-0-1-0-shape.md` gains the same form of note on **D28**, **D29**,
      **D34** and **D35**.
- [ ] `docs/adr/README.md` gains whatever index entry that file's convention requires, and no
      existing ADR, clause or D-number is renumbered.
- [ ] The record states the branching invariant it does **not** change: work still reaches
      `main` only through a train (D1), promotion is still a fast-forward (D5), and the human
      pause stays at the head of `promote.yml` (D15). A beta release is a publish from the
      train, not a second route to `main`.

## Dependencies

None. **Every other issue in this milestone depends on this one**, and #317 and #321 depend on
it for a decision rather than for a record.

## Out of scope

- Any code, workflow or script change. This issue is documentation only.
- Moving any existing ADR's status. 0016 and 0023 are `ACCEPTED` and stay so; they are amended
  by note, not superseded.

## Branching

qcentic-dev:
Closes #315.

**Base is `feat/release-pipeline`** — the feature branch for #314. Not `main`, not `beta/0.4.1`.

**Documentation only.** No `install.sh`, nothing under `.github/workflows/`, nothing under
`packages/`, nothing under `scripts/`. Four files: one new ADR, two dated amendment notes, one
index row.

## What this decides

ADR 0023 D8 says that if betas ever gain a GitHub Release *"this clause must be revisited
before that change lands, not after"*, and D9 says such a Release *"must be created with
`prerelease: true` and CI must assert the flag"*. #314 is that change. **This record is that
revisit**, and it cites both clauses in its header.

Four **operator constraints** are recorded rather than weighed — no alternatives listed, no
trade-off presented, no clause a later reader can take as re-openable:

- **C1** — a beta version string is `x.y.z-beta` on every surface: the git tag, the GitHub
  Release, every image tag, every asset filename, anything npm would see. No counter, no
  suffix of any kind. Version strings stay clean on all three channels — `x.y.z` for a
  release, `x.y.z-beta` for a beta, `main` as the moving latest.
- **C2** — install is not activation. `install.sh` places the Core bundle and the CLI and
  stops; `actana setup` is a separate command the operator runs.
- **C3** — a beta cut is dispatched by a person, never on a merge into the train.
- **C4** — the URL scheme is the ref: `…/control/main/install.sh` for latest, the tag path for
  a release, the beta train path for a beta.

Four decisions the milestone was blocked on:

| Question | Decision | Clauses |
| --- | --- | --- |
| The channel mechanism under a fast-forward | **The stamped line version, written by the cut.** A `curl \| bash` script cannot read its own URL, and promotion fast-forwards `main` — so a channel constant on the train *becomes* `main`'s bytes. **Any design that stores the channel in the file is wrong.** What survives a fast-forward is a value true on both sides of it: the line. Resolution is *"the release of this line if it exists, else this line's beta"*, per line, with no listing call. C4 is what forces this route rather than the release-asset door. | D1–D6 |
| Tarball promotion (#321) | **Dropped.** A digest is version-neutral; a Core tarball is not — the version is in the asset name, the archive root and `core-manifest.json` (`scripts/lib/core-tarball.mjs:228-234`). No rename makes beta bytes release bytes. The version-neutral artifact is refused once, with 0016 D29's price attached: a major bump plus a documented migration. | D18–D20 |
| The CLI-only beta path (#320) | **Decided by the operator and recorded, not reopened.** The CLI is packed and attached as a Release asset, installed with `npm i -g <asset-url>` — never a registry version, because npm burns a version on first publish (`scripts/rehearse-npm-publish.mjs:11-14`). Consequences recorded: the registry namespace is untouched, `--provenance` is lost and `SHA256SUMS` stands in its place, and `assertPackedManifest` needs a beta-only branch. | D15–D17 |
| Who owns `packages/shared/src/actana-release-channel.ts` | **#322 creates, #317 consumes.** #322 is wave 1 with no dependencies and lands first; the additions are its subject anyway. #317's diff is `install.sh`, its tests and the fixture server. The new noun is a **line**, not a channel — `ReleaseChannel` already means *which repo and which hosts*. | D6 |

The beta channel model is stated as numbered decisions: the fifth tag class and that it moves
(D7, against 0023 D44's *"release tags are immutable"* — a release tag is immutable, a beta tag
is a handle); `prerelease: true` with a CI assertion in D9's own words (D11); the three tarball
targets and the absence of an Intel Mac build (D10); and the image retag from the train digest
(D12).

**What does not change** is stated too (D21): work still reaches `main` only through a train
(0023 D1), promotion is still a fast-forward (0023 D5), and the human pause stays at the head
of `promote.yml` (0023 D15). A beta release is a publish from the train, not a second route to
`main`.

## The amendments

Six clauses gain a dated note in the block-quote form ADR 0031 uses. **Nothing is renumbered,
nothing is superseded, and 0016 and 0023 both stay `ACCEPTED`.**

- **0023 D8** — the *"nothing that parses versions ever sees one"* argument retires; each
  parser is answered by name. `beta-x.y.z` as an image tag is untouched.
- **0023 D9** — the Docker-only half is withdrawn; the `prerelease: true` + CI assertion half
  is **kept in force** and becomes #318's acceptance criterion.
- **0016 D28** — a beta Release publishes a larger asset set on a different tag class; the
  release's set and its guard are unchanged.
- **0016 D29** — `install.sh` becomes a beta asset as a *copy*; the canonical door stays on
  `main`, so the clause's reason survives. Its installer contract is now load-bearing twice.
- **0016 D34** — `beta-release.yml` is a sixth entry point; `workflows.test.mjs:71-79` fails on
  its first commit (#318).
- **0016 D35** — the macOS spend moves to a more frequent trigger, owned rather than left to
  the inventory test's silence. Zero macOS in per-PR CI still stands.

**The amendment structure is deliberately left open** for #325 (which deletes the automatic
train cut and amends 0023 D25) and #327 (which rewrites version handling and reaches 0023 D3
and D7's count). Neither clause is touched here. D1 is written to survive #325: it says *the
cut* writes the stamp, not *`promote.yml`'s `next-train` job*.

## The sharpest edge, named for review

**D3 — the promotion window.** `advance` fast-forwards `main` and pushes the tag *before* the
`release` job runs, and the Release object is created at the end of `release.yml`. So for the
length of one release run, `main` carries the new line's stamp while that line's Release does
not exist, and the public one-liner resolves to that line's beta. It is unavoidable under D1
— after the fast-forward the two branches' bytes are identical, so no rule expressed in those
bytes can answer differently — and it is accepted because the bytes are prerelease bytes of the
line being promoted, the machine reports the beta version honestly, and it self-heals on the
next update once #322 lands. **This is the one consequence a reviewer should push on before
ratifying.**

## Status and verification

The ADR is **PROPOSED**. It takes **0036**, the next free number (`docs/adr/` runs to 0035),
and adds its row to the index at `docs/README.md#architecture-decisions`, which is where
`docs/adr/README.md` says the index lives.

- All three commits pass `commitlint` against the repo's own `commitlint.config.mjs`, not just
  the titles, and every body line is ≤ 100 characters.
- `panel-image.test.mjs`'s *"quotes ADR 0016 D29's installer-contract rule verbatim"* assertion
  still passes: the D29 note is appended *after* the clause's own block quote, so the first
  block quote after `**D29 —` is unchanged.
- **No example of the banned counted beta form appears anywhere in the record**, which is the
  failure the acceptance criterion exists to prevent.

Draft on purpose: this record needs the repository owner's ratification, and #317, #320 and
#321 are written against clauses in it.


qcentic-dev:
## Acceptance criteria from #315

Every criterion in the issue, checked or explicitly unchecked with the reason.

- [x] **A new ADR at `docs/adr/0036-<slug>.md`; 0036 is the next free number; renumber nothing.**
      [`docs/adr/0036-the-beta-release-channel.md`](https://github.com/actana/control/blob/docs/adr-beta-release-channel/docs/adr/0036-the-beta-release-channel.md).
      `docs/adr/` on this base runs to `0035-a-second-skill-for-the-sub-agent-role.md`, so 0036
      is free. Nothing is renumbered — the two files claiming 0018 and the two clauses claiming
      D38 inside 0023 are untouched, and the record says so in its header.

- [x] **States first and without qualification that a beta version string is `x.y.z-beta`
      everywhere, as a fixed operator constraint, with no alternatives and no re-openable
      clause; and contains no example of the banned form.**
      §A **C1**, the first clause in the record, before any decision. §A's preamble says the
      four constraints "are recorded as **constraints**, not as decisions this record reached:
      there are no alternatives listed under them, no trade-off is weighed, and a later reader
      who wants one changed is asking the owner, not re-reading an argument." C1 names the git
      tag, the GitHub Release, every image tag, every asset filename and anything npm would
      see, and bans the counted form outright. **No counted beta string appears anywhere in the
      record** — the shape is described in prose and never written out. C1 also scopes itself:
      it binds the beta channel and does not touch 0023 D30's backport release candidates,
      which carry an identifier by design.

- [x] **Records the consequence in the one place it lands — the npm registry is not a beta
      surface — and says why the asymmetry is a property of npm.**
      **D15**, quoting `scripts/rehearse-npm-publish.mjs:11-14`. It states that the registry
      publish is dropped "as a consequence of C1 rather than as the outcome of a comparison",
      and lists what every other surface does instead: a git tag can be force-updated, a Docker
      tag is a pointer and `beta-x.y.z` already moves per merge (0023 D7), a Release asset is
      clobbered in place. D15 also records what a beta cut leaves alone — no dist-tag moves,
      `release.yml`'s `npm` job is not modified.

- [x] **States the beta channel model as numbered decisions: the fifth tag class and that it
      moves; `prerelease: true` with a CI assertion in D9's own words; the three tarball
      targets and the absence of an Intel Mac build; the image retag from the train digest.**
      **D7** (fifth class, moves, and which half of 0023 D44 survives — a release tag is
      immutable, a beta tag is a handle), **D11** (D9's sentence carried unchanged, plus the
      read-back that enforces it), **D10** (the three targets, `mac-arm64` not `darwin-arm64`,
      and no Intel Mac because the target has not existed since 0016 D28 as amended), **D12**
      (`docker buildx imagetools create` from the train's verified digest, nothing rebuilt).

- [x] **Decides the CLI-only install path for a beta. #320 does not start until this clause
      exists.**
      **D16** — the CLI is packed and attached as a tarball asset on the beta prerelease and
      installed with `npm i -g <asset-url>`, never a registry version. Recorded as the
      operator's decision and explicitly not reopened. **D17** records what the route loses:
      `--provenance` is a registry artifact, an asset has no attestation, `SHA256SUMS` stands
      in its place, and #323's operator-facing text must not imply an attestation that is not
      there. D16 bounds #320's remaining packaging choice without making it.

- [x] **Decides the channel mechanism — (a), (b) or (c) — and says what it costs. #317 does not
      start until this clause exists.**
      **D1** chooses the stamped line version. The record states the structural fact first: a
      `curl | bash` script has no `$0` and no `BASH_SOURCE`, and promotion fast-forwards `main`
      (0023 D5), so a channel constant on the train becomes `main`'s bytes — **any design that
      stores the channel in the file is wrong**. The release-asset door and "neither" are
      recorded with their reasons; the release-asset door is unavailable under **C4**, whose
      third row it cannot serve at all. The costs are four clauses: **D2** the resolution rule,
      **D3** the promotion window, **D4** the separate assertion (never by extending
      `MANIFESTS`, whose `assert_manifest_set` refuses a shrinking set), **D5** why 0016 D29 is
      not violated.

- [x] **Decides what §7 of the spec becomes now that a Core tarball cannot be version-neutral
      (#321), rather than leaving a ticket to discover it.**
      **D18** drops the tarball half of promotion, with both obstacles stated: the tarball
      self-identifies in three places (`scripts/lib/core-tarball.mjs:228-234` quoted), and a
      beta is cut at an arbitrary train tip so the 0023 D16-equivalent revision assertion would
      fail on most cuts. **D19** refuses the version-neutral artifact once, with 0016 D29's
      price attached — a major bump plus a documented migration. **D20** records that beta and
      release assets are never confusable, which is #321's own last criterion.

- [x] **`docs/adr/0023-…md` gains a dated note on D8 and on D9, in the block-quote form ADR
      0031 uses, recording that the condition each named has occurred and pointing at 0036;
      D9's sentence is quoted into #318.**
      Both notes added, dated 2026-08-24, in the appended block-quote form. The D8 note retires
      the "nothing parses a beta" argument and keeps D8's own subject intact. The D9 note
      withdraws the Docker-only half and **keeps the `prerelease: true` + CI-assertion half in
      force**, naming it as #318's acceptance criterion. 0036's header quotes both clauses
      verbatim as the reason the record exists.

- [x] **`docs/adr/0016-the-0-1-0-shape.md` gains the same form of note on D28, D29, D34 and
      D35.**
      Four notes added, same form, same date. D28: the release's asset set and its guard are
      unchanged; the beta's set is larger and on a different tag class. D29: the beta asset is
      a copy, the canonical door stays on `main`, and the installer contract is now
      load-bearing twice. D34: a sixth entry point, and `workflows.test.mjs:71-79` fails on
      #318's first commit. D35: the per-PR half stands unqualified; the frequency on the
      release side is what changes, and 0036 D13 owns it.

- [x] **`docs/adr/README.md` gains whatever index entry that file's convention requires, and no
      existing ADR, clause or D-number is renumbered.**
      That file's convention is explicit: *"The index with one-line summaries is in
      [`../README.md`](../README.md#architecture-decisions)"* — it carries no list of its own,
      only the two collision notes. The row was therefore added to
      `docs/README.md#architecture-decisions`, matching the table's existing form including the
      `— **PROPOSED**` suffix 0031 and 0035 use. `docs/adr/README.md` itself needs no edit and
      got none; editing it would have added a second index the file says does not live there.
      Nothing is renumbered anywhere.
      *Observed, not fixed:* that table has no row for
      `0034-short-code-pairing-enrollment.md`. Pre-existing on this base and outside this
      ticket's scope — flagged rather than silently repaired.

- [x] **States the branching invariant it does not change: D1, D5 and D15 of 0023, and that a
      beta release is a publish from the train, not a second route to `main`.**
      **D21**, in §F, in those words, plus 0023 D2 (one train at a time) and the statement that
      no beta tag is ever an input to a promotion. **D22** records that 0016 and 0023 keep
      their status and their numbering.

## Out of scope, honoured

- [x] **No code, workflow or script change.** The diff is four files: one new ADR, notes in two
      ADRs, one index row. Nothing under `packages/`, nothing under `scripts/`, nothing under
      `.github/workflows/`, and `install.sh` untouched.
- [x] **No existing ADR's status moved.** 0016 and 0023 are still `ACCEPTED`, amended by note.

## Left open on purpose

- [x] **The amendment structure stays open for #325 and #327.** Neither 0023 D25 (the automatic
      cut) nor D3 / D7's count is touched, and every note here is *appended* so a later
      amendment lands beside it rather than editing it. **D1 is written to survive #325**: it
      says *the cut* writes the stamp, not *`promote.yml`'s `next-train` job*, so it stays true
      whether the cut is a job or a person.

## Not done, and why

- [ ] **The record is not `ACCEPTED`.** It is **PROPOSED**, as #315 requires. The repository
      owner ratifies or rejects it; this PR stays a draft until then.
- [ ] **`docs/ci-cd.md:325-333` is not rewritten.** It restates 0023 D8 and D9 in the
      operator's voice and stops being true under this record — but that is **#323's** work by
      #314's own wave plan, and #315 is documentation of the *record*, not of the operator
      surface. 0036 D18 and the consequences name the paragraph and the sentence #323 must add.
- [ ] **#320's remaining packaging choice is not made here.** The route is decided (D16); which
      of the three ways the packed CLI manifest is made honest about `@actana/sdk` is #320's,
      and D16 bounds it with three constraints instead of picking.
- [ ] **The promotion window is not closed.** D3 records it, sizes it, and says the only lever
      is reordering `release` ahead of `advance` in `promote.yml` — a change to the promotion
      sequence 0023 D40 and D16 depend on. **This record deliberately does not order it**; a
      ticket that wants it closed is amending 0023.

mrfqcentic (changes requested):
## Review against #315 — request changes

I read #315 in full, then read this diff clause by clause against the enforcing lines in the
repository rather than against the PR description. **The record is very good.** Every source
citation I checked resolves exactly — `promote.yml:482`'s D44 error string, `release.yml:79-83`'s
D29 quote, `release.yml:202`'s tag regex, `release.yml:219-231`'s `resolve` guard,
`ci.yml:249-277`'s `assert_manifest_set`, `workflows.test.mjs:71-79` and `:206-209`,
`core-tarball.mjs:228-234`, `rehearse-npm-publish.mjs:11-14`, `semver.ts`'s `versionCore`,
`build.mjs:65`'s `external: ["ws", "undici", "selfsigned"]`, and `CORE_TARGETS`' single darwin
row. Nine of the eleven acceptance criteria are met outright.

I am blocking on one thing, and it is inside the clause this record exists to produce.

---

### BLOCKING — C4's table and D2's rule disagree in two of four rows, and #317 implements both

`docs/adr/0036-the-beta-release-channel.md:87-98` calls C4 **"the constraint"**, and D1
(`:104-106`) says that table **"is what forces D1"**. So C4 is not illustrative — it is the
specification #317 builds. It does not match D2.

**`docs/adr/0036-the-beta-release-channel.md:94`** — row 4:

> `| that exact beta | …/actana/control/v0.4.1-beta/install.sh | the same, pinned |`

Nothing about that row is pinned, by this record's own clauses:

1. **D7 (`:211-236`) makes the ref itself move.** *"a beta tag is a handle and moves"*, `Moves:
   per cut`. Fetching `install.sh` at ref `v0.4.1-beta` returns whatever the most recent cut
   points at. A moving ref cannot pin.
2. **D2 (`:131-137`) resolves it to the release, not the beta.** D1 (`:114-116`) decides the
   stamp is *the line* — `install.sh` carries `0.4.1`, never `0.4.1-beta`. The `v0.4.1-beta`
   tag points at a train commit, so the file at that ref carries `0.4.1`, and D2 then applies
   *"the release `v<line>` if that Release exists; otherwise `v<line>-beta`"*. Once 0.4.1 ships,
   that URL installs **the release**. The row promises the opposite.

D2 asserts the row works anyway, at **`:135-137`**: *"On a tag checkout the stamp is that tag's
own version, so the tag rows of C4's table work for free with no extra machinery."* That holds
for the release-tag row (`v0.4.0` → stamp `0.4.0` → release exists → installs 0.4.0 ✅). It
cannot hold for the beta-tag row without contradicting D1: if the stamp at that tag were
`0.4.1-beta`, D2's rule would look for a release named `v0.4.1-beta` and fall back to
`v0.4.1-beta-beta`. The record does not say which reading is intended, and the two are not
reconcilable.

**`docs/adr/0036-the-beta-release-channel.md:91`** — row 1 has the smaller version of the same
gap. It says `main/install.sh` installs the *"newest release"*. D2 installs **the stamped
line's** release. Those diverge exactly when 0023 D26/D30 fires: after a backport publishes
`0.3.5` from a release branch, GitHub's `/releases/latest` — which `install.sh:250` reads today
— answers `0.3.5`, while D2 answers `0.4.1`. D2's answer is almost certainly the better one, but
it is a **behaviour change to the stable public one-liner** that the record neither names nor
costs, in a clause whose AC requires it to say what it costs.

**What I need before this can be ratified:** C4's "Installs" column restated in D2's vocabulary,
and row 4 resolved one of three ways — say it is an alias of row 3 that becomes the release after
promotion; say the pinned form is `--version x.y.z-beta` (which C4's own note at `:97` already
allows); or delete the row. Plus one sentence in D2 or D3 owning the backport divergence.

---

### Required additions (not blocking on their own, but this record gates eleven tickets)

**`docs/adr/0036-the-beta-release-channel.md:144-168` — D3 calls the window "bounded" and the
bound is only proven for the happy path.** The ordering is real and I verified it:
`promote.yml:518-520` declares `release: needs: [resolve, advance]`, so `advance` fast-forwards
`main` and pushes the tag first. D3 prices *"the duration of one release run"*. It does not say
what happens when that run does not finish — a failed or cancelled `release` job leaves `main`
carrying the new line's stamp with no Release for it, and D2 then serves that line's **beta**
from the public door indefinitely, not for one run. The gating table at `:525` points #326 at
*"D3's ordering discussion"*, which is not the same as a stated recovery. One clause naming the
failure case and its recovery would close it.

**`docs/adr/0036-the-beta-release-channel.md:211-236` and `:260-276` — the two new tag classes
have no lifecycle.** D7 adds a `vx.y.z-beta` git tag and a prerelease Release; D10 adds
`x.y.z-beta` image tags in `panel`/`core`. The record never says whether either is removed after
the line promotes. 0023 D45 is explicit that *"Docker Hub has no tag garbage collection and no
undelete"*, and its sweep covers `pr-*`/`sha-*` in the `-dev` repositories only — 0023 D36/D38
deliberately keep the delete-capable credential out of the repositories that hold `latest`, which
is exactly where D10 puts the new image tags. So the answer cannot be inferred; #318 and #319 will
have to invent it. Say "they persist" or say what sweeps them.

---

### Nits

**`docs/adr/0036-the-beta-release-channel.md:31`** — the header cites D9 as `` `0023-…:63` ``.
That was true on `feat/release-pipeline`; commit 2 of *this PR* inserts the D8 note above it and
D9 now lives at line **65** of the amended file. The citation is stale on its own merge commit.
This repository cites ADRs by line elsewhere (`packages/shared/src/orchestration-skill-install.ts:56`
uses `` `docs/adr/0031-…:127-131` ``), so it is a convention worth keeping right. `:61` for D8 is
still correct.

**`docs/adr/0036-the-beta-release-channel.md:247-248`** — *"a machine on a beta is silently
downgraded by a bare `actana update`"*. `isNewerSemver("0.4.1", "0.4.1-beta")` returns `false`
(both parse to `[0,4,1]`), so the machine is **stranded on the beta**, never moved backwards.
"Silently stranded" is the accurate word and it is a slightly different bug for #322 to fix.

---

### What I verified and am satisfied with

- **The channel mechanism survives the fast-forward.** This was the claim I most wanted to break
  and it holds. D1 `:110-116` decides that nothing which says `beta` is ever committed to a
  tracked file: *"Any design that stores the channel in the file is therefore wrong"*, and
  *"`install.sh` carries the **line**"*. `main`'s copy after promotion is byte-identical to the
  train's and both say `0.4.1`; the differentiator is external state — whether Release `v0.4.1`
  exists — which the promotion itself creates. That is a genuine mechanism, not a restatement of
  the problem, and D3 prices the one interval where it reads wrong.
- **The record decides rather than surveys, on all four blocked questions.** D1 `:104`, D18 `:380`
  (*"the tarball half of promotion is dropped"*), D16 `:338` (*"This is the operator's decision and
  it is not reopened"*), D6 `:189` (*"#322 creates; #317 consumes"*). The two rejected channel
  routes at `:118-129` are recorded as refusals with reasons, which is the opposite of a survey.
- **No counted beta form appears anywhere.** The only `beta.<digit>` in the whole diff is an
  unchanged context line — 0023 D8's pre-existing `1.2.3-beta.1` describing semver's own shape.
  Zero added lines contain one. C1 `:61-72` states the ban in prohibition form without ever
  writing an example of it, which is stronger than the criterion asked for.
- **0023 D8 and D9 are both cited and both answered**, quoted verbatim in the header at `:22-31`,
  D8 by §C/D8 answering each parser by name, D9 by D11 `:278-286` keeping the assertion in force.
- **Nothing is renumbered or rewritten.** 539 additions, **0 deletions** across four files. 0023
  D3, D7 and D25 are untouched, so #325 and #327 have their clauses intact, and §G `:466-478`
  reserves the space explicitly.
- **0036 is genuinely free** (`docs/adr/` runs to 0035) and the index row is in `docs/README.md`,
  which is where `docs/adr/README.md` says the index lives — not in `docs/adr/README.md` itself.
  Row format matches its neighbours.
- **Documentation only.** Four files, all under `docs/`. Nothing in `install.sh`,
  `.github/workflows/`, `packages/` or `scripts/`.
- **`panel-image.test.mjs:479-495` still passes.** It slices from the first `**D29 —` and takes
  the first following block quote; the new note is appended after the installer-contract quote and
  after the "Until the repo is public" paragraph, so the quote it reads is unchanged. Confirmed by
  reading the test, not the PR description.
- Three commits, one per logical change, no body line over 100 characters.

Fix the C4/D2 disagreement and I have nothing else blocking. The rest is a strong record.

qcentic-dev:
@mrfqcentic — thank you for reading it against the enforcing lines rather than against the
description. Every point landed; all five are fixed in `c7ce9fa`, one commit, still
documentation-only and still a draft. Detail below, and where I disagreed with a premise I say
so rather than quietly working around it.

---

### BLOCKING — C4 and D2 ✅ fixed

**You were right that C4 is a specification, not an illustration**, and that it did not match
D2. C4's **Installs** column is now written in D2's vocabulary, with a sentence above the table
saying why: *"this table is a specification #317 builds and the two must not be able to
disagree."*

**Row 4 — resolved as your first option: an alias of row 3, not a pin.** Both of your reasons
are now stated in the record: the tag moves per cut (D7) so the ref will not hold still, and
the file at that ref carries the **line** (D1) so D2 treats it exactly as the train branch.
The row now reads *"an alias of the row above, not a pin. It follows the same line, and once
that line has a release it installs the release."* And because deleting the promise without
replacing it would leave a real want unanswered, the note under the table adds your second
option as the answer to it: **the pinned form is `--version x.y.z-beta`**, which the note
already permitted and which #317 keeps working. Your `v0.4.1-beta-beta` reductio is exactly
why the other reading was not taken — it is unreachable now, because the stamp is never
anything but the line.

**Row 3 gained a fact it was missing**: the train branch is deleted at promotion (0023 D4), so
that URL stops existing rather than quietly changing meaning. Row 2 now says it is a pin *and
why* — a release tag is immutable (D7).

**Row 1 — restated, and the divergence owned in D2.** It now says *"the release of the line
`main` is stamped with, which in steady state is the newest release."*

**One correction to your premise, and it changes where the cost lands.** The backport case does
not diverge. **0023 D28** is *"A backport never moves `latest`, on either Docker Hub or GitHub,
and an assertion enforces it"* — it exists precisely because `gh release create` defaults
`make_latest` to `true`, and its stated failure mode is *"every existing user would be told to
downgrade."* So after `0.3.5` publishes from a release branch, `/releases/latest` still answers
`0.4.1`, and D2 answers `0.4.1` too. D2's change there is that it stops *depending* on that
assertion holding, which is a small strengthening rather than a behaviour change, and D2 now
says that in those terms.

**But you were pointing at a real hole, and it is one row over — rollback.** 0023 D44 flips the
GitHub Release's latest flag and *"nothing else"*: it does not rewrite `main`. D2 reads the
stamp and not that flag, so a fresh `curl …/main/install.sh | bash` **keeps installing the
rolled-back release** while every Docker pull and the update checker have already moved off it
— a genuine regression against today's script, which follows the flip immediately. That is now
**D24**, with the recovery 0023 D44 already names in its own next sentence: the hotfix train
(0023 D22) re-stamps `main` in the same fast-forward, so the one-liner follows with no new
mechanism. The cost — the interval between the flip and that promotion — is priced, and the
rollback runbook is handed to #323. I have added it to the Consequences as the second edge to
push on beside D3, and to C4 row 1 as *"the one case where it is not is D24."*

---

### Required addition — D3's failure case ✅ added

You are right that the bound was only proven for the happy path. D3 now says: a failed or
cancelled `release` job leaves `main` carrying the new line's stamp with no Release for it, and
D2 serves that line's beta **for as long as the failure lasts, not for one run**. Nothing
self-corrects — the stamp cannot be walked back, because 0023 D5 lets `main` advance only by
fast-forward and `main` is already at the train tip.

The recovery is stated rather than linked: **re-dispatch the release.** The tag is already
pushed (`promote.yml:385-387`) and `release.yml` carries a `workflow_dispatch` beside the
`workflow_call` `promote.yml` uses (`release.yml:119`, `:133`), so a re-run needs no new commit
and does not touch `main`. The clause now says a promotion whose release leg is red is **not a
cosmetic failure and must be finished rather than abandoned**, and #326's row in the gating
table points at that instead of at "D3's ordering discussion".

**This also forced a gap I had not seen, and it is the reason D2 grew a step.** If no beta of
the line was ever cut — which C3 makes the normal state of a young train — then in that same
window D2 found neither a release nor a beta and the install would have **failed outright**: a
public-door outage for the length of every release run that follows a train nobody cut a beta
from. D2 now has a terminal **step 4, `/releases/latest`** — exactly what the script reads
today — so that case gets today's correct answer. D3's opening now splits the two cases
explicitly.

---

### Required addition — the lifecycle of the two new tag classes ✅ added, as **D23**

Answer: **they persist**, and it is a decision rather than an omission.

The git tag and its prerelease Release stay for 0023 D44's reason — the record is not
rewritten — plus a live one: that Release is the only home of the beta's tarballs,
`SHA256SUMS`, `install.sh` copy and CLI asset, so deleting it would break C4's fourth row and
every `--version x.y.z-beta` pin at once.

The image tags stay because of the refusal you named, quoted into the clause: 0023 D45's sweep
covers `pr-*` and `sha-*` in the `-dev` repositories only, and 0023 D36 with D38 (*the
delete-capable credential*) keeps that credential out of the repositories holding `latest` —
which is exactly where D10 puts `x.y.z-beta`. Sweeping one would mean widening its reach, and
D38's own words are *"do not revisit by widening the list."*

The clause also prices it, and the price is smaller than it looks: **accumulation is one per
line, not one per cut**, which is the moving tag (D7) paying for itself. At worst it doubles a
tag list 0023 D36's `descriptions` argument wants presentable. #318 and #319 now read this
instead of inventing it, and both have D23 added to their rows in the gating table.

---

### Nits ✅ both fixed

**The stale D9 citation.** Fixed to `` `0023-…:65` ``, and it now says *"as this pull request's
second commit leaves it"* so the next reader knows the number is post-amendment rather than
wondering which tree it was taken from. `:61` for D8 is unchanged, as you say.

**"Downgraded" was the wrong word, and you are right about the mechanism.** `isNewerSemver`
answers `false` because both sides parse to the same numeric core, so the machine is
**stranded** on its beta and the update notice never fires for the release of its own line.
The bullet now says that. It also keeps the downgrade, because that bug is real and is #322's
title — it just is not this function: a bare `actana update` resolves `/releases/latest`, which
excludes prereleases, and the already-current guard compares two unequal strings, so the tree
is replaced with the newest release and reported as an update. Both are now stated as the two
separate paths #322 carries.

---

### What did not change

Still four files, all under `docs/`. Nothing in `install.sh`, `.github/workflows/`, `packages/`
or `scripts/`. No existing ADR, clause or D-number renumbered — **D23 and D24 are appended**,
per this repository's own rule, so your line-and-clause citations above still resolve. 0016 and
0023 are untouched by this commit and still `ACCEPTED`. Base is still `feat/release-pipeline`,
nothing is merged, and the PR is still a draft.

The checklist comment above is now one item stale on two clause counts (D1–D22 → D1–D24); the
record itself is the authority.

qcentic-dev:
**CI on `c7ce9fa`: green.** 18 pass, 5 skipping, 0 failing —
[run 32723113291](https://github.com/actana/control/actions/runs/32723113291).

Worth recording because the first attempt was red and it was **not** this diff. `Unit Tests`
failed on `packages/cli/src/__tests__/in-process-core-session.test.ts` >
*"sends and awaits the turn in one PTY resolution, and prints the settled screen"* with
`Error: Test timed out in 60000ms.` — a timeout, not an assertion, on a PTY-driven in-process
Core session. The diff between the green run on `155c631` and the red one is a single Markdown
file under `docs/adr/`, so nothing in it is reachable from that test. Re-running the failed job
alone passed in 2m39s with no code change. Flagging it rather than staying quiet: that test has
a 60 s ceiling over work the same file elsewhere budgets at 30 s, and a flake in it will land on
somebody else's PR next.

Still a draft, base still `feat/release-pipeline`, nothing merged.

mrfqcentic (approved):
## Re-review at `c7ce9fa` — approved

I verified every item from the enforcing line in the file at the new head, not from the reply.
All five are fixed, and three of them are fixed better than I asked for.

**First, a correction of my own.** My backport claim in the previous review was wrong, and the
reply is right about why. I asserted `/releases/latest` would answer `0.3.5` after a backport;
**0023 D28** (`0023-…:117`) is *"A backport never moves `latest`, on either Docker Hub or GitHub,
and an assertion enforces it"*, written precisely because `gh release create` defaults
`make_latest` to `true`. So the endpoint and the stamp already agree there and there was no
divergence to cost. I checked that clause before accepting the correction.

---

### 1. BLOCKING FINDING — C4's table and D2's rule ✅ resolved

**All four Installs rows are now in D2's vocabulary** (`:89-98`), under an explicit statement of
why (`:89-90`: *"because this table is a specification #317 builds and the two must not be able to
disagree"*). Row by row, against D2 at `:144-182`:

- **Row 1** — *"the release of the line `main` is stamped with, which in steady state is the newest
  release — the one case where it is not is D24"*. D2's vocabulary, and the single exception is
  named inline rather than left for a reader to discover.
- **Row 2** — *"that line's release, which at a release tag is that release. A release tag is
  immutable (D7), so this row is a pin"*. Matches D2 step 2; the pin claim is now justified by the
  immutability clause rather than asserted.
- **Row 3** — *"that line's beta. The train branch is deleted at promotion (0023 D4), so this URL
  stops existing rather than quietly changing meaning"*. I verified 0023 D4 at `:35` — *"A train is
  disposable, and is deleted on promotion"* — so the row is accurate, and the failure mode it
  forecloses (a URL that silently changes meaning) is the right one to have closed.
- **Row 4** — *"**an alias of the row above, not a pin.** It follows the same line, and once that
  line has a release it installs the release"*. This is exactly the behaviour D2 produces, stated
  in D2's terms. The "pinned" claim is gone.

**Neither row can now be read two ways.** `:101-107` closes the ambiguity in the general case —
*"There is no ref that pins a beta, and the fourth row does not claim to be one"* — and gives both
reasons from this record's own clauses (D7 moves the ref, D1 puts the line and not the beta in the
file), then names the form that does pin: `--version x.y.z-beta`. D2 reciprocates at `:156-159`:
*"C4's fourth row is not a fourth behaviour"*, resolving through steps 2/3 like the train branch.
The table and the rule now state the same thing from both ends, which is what I asked for.

**The row-1 divergence is named with its cost**, and correctly relocated. `:176-182` walks it
properly: the two answers are held equal by two different mechanisms (0023 D28 for the endpoint,
the cut plus `Train rules` for the stamp), a backport produces no divergence, and *"There is exactly
one state where the two genuinely disagree and D2's answer is the worse one — a rollback"*. That is
D24.

**D24 (`:540-563`) is the strongest clause added in this round.** It says plainly that D2 *"keeps
installing the rolled-back release"*, calls it *"a real regression against today's script"* and
concedes *"D2 has the worse answer"* — then gives the recovery from 0023 D44's own next sentence
(the hotfix train re-stamps `main` in the same fast-forward), prices the interval, hands the runbook
to #323, and closes the two obvious follow-on questions (the beta Release need not be deleted; an
operator who needs the good version now has `--version x.y.z`). This is a record owning a defect
rather than describing one.

### 2. D3's failure case ✅ resolved, `:212-227`

*"**The bound is one release run only when that run finishes, and the failure case is stated here
rather than left to be met.**"* It states the consequence — the door serves that line's beta *"for
as long as the failure lasts, not for one run"* — and why nothing self-corrects: the stamp cannot be
walked back because 0023 D5 lets `main` advance only by fast-forward and `main` is already at the
train tip.

The recovery is concrete and I verified it: *"re-dispatch the release"*, resting on
`release.yml:119` (`workflow_call`) and `:133` (`workflow_dispatch`) — both land exactly there, and
the file's own comment at `:128-130` confirms the dispatch trigger exists to re-run a whole release.
The clause draws the operational conclusion (*"not a cosmetic failure and must be finished rather
than abandoned"*) and #326's gating row now points at that instead of at "D3's ordering discussion".

**The unprompted catch here is the better half.** D2 gained a **step 4** (`:150-151`,
`:161-165`): a line with neither a release nor a beta — the normal state of a young train under C3,
and the failed-release case — previously fell off the end of the rule and would have failed
outright. Step 4 falls back to `/releases/latest`, today's answer. D3's opening (`:188-191`) now
splits the two cases explicitly. That was a genuine public-door outage nobody had raised, found by
following the failure case honestly.

### 3. Lifecycle of the beta tag, the Release and the beta image tags ✅ resolved, D23 `:513-538`

Decided, not deferred: **they persist**, and stated as a decision. The git tag and Release stay for
0023 D44's principle plus a live use — that Release is the only home of the beta's tarballs,
`SHA256SUMS`, `install.sh` copy and CLI asset. The image tags stay because sweeping them would mean
widening a credential 0023 D38 refuses to widen; I verified the supporting clauses (0023 D45's sweep
covers `pr-*`/`sha-*` in the `-dev` repositories only, and D36/D38 keep the delete-capable
credential out of the repositories holding `latest` — which is where D10 puts `x.y.z-beta`).

The clause also prices it and the price is small: *"accumulation is bounded at one per line rather
than one per cut"* — the moving tag paying for itself. #318 and #319 now carry D23 in their gating
rows, so neither has to invent it.

### 4. The stale 0023 D9 citation ✅ fixed, `:31`

Now `` `0023-…:65` ``, with *"as this pull request's second commit leaves it"*. I confirmed against
`0023-release-trains-and-digest-promotion.md` at `c7ce9fa`: **D9 is at line 65**. `:61` for D8 is
unchanged and still correct. Adding which tree the number is taken from is a better fix than the one
I asked for.

### 5. The wrong verb ✅ fixed, `:305-313`

Now *"The machine is therefore **stranded** on the beta, never moved backwards by this function"*,
with the sharper consequence I had not drawn — *"its update notice never fires for the release of
its own line — the population most in need of that notice"*. It also keeps the downgrade and
correctly separates it as a **second path**: a bare `actana update` resolves `/releases/latest`,
which excludes prereleases, and the already-current guard compares two unequal strings. Both are
handed to #322 as two paths rather than one. This is more precise than my note.

### Counter check ✅ clean

`grep -nE 'beta\.[0-9]'` over `docs/adr/0036-the-beta-release-channel.md` at `c7ce9fa` returns
**nothing**. Over the whole diff there is exactly one hit and it is a **context line, not an added
line**: 0023 D8's own pre-existing prose describing semver's shape (*"Semver's own form is
`1.2.3-beta.1`"*), which is a description of the form being rejected, not a use of it. No added line
in this PR contains a counted beta form.

### Documentation only ✅ confirmed

Four files, **665 additions, 0 deletions**:

```
docs/README.md                                        +1
docs/adr/0016-the-0-1-0-shape.md                      +8
docs/adr/0023-release-trains-and-digest-promotion.md  +4
docs/adr/0036-the-beta-release-channel.md           +652  (new)
```

Nothing under `install.sh`, `.github/workflows/`, `packages/` or `scripts/` — verified against the
diff's file list, not the description. Zero deletions repo-wide, so nothing was renumbered or
rewritten: **D23 and D24 are appended**, D1–D22 are untouched, and every clause citation from my
previous review still resolves. CI is green on `c7ce9fa` — 18 pass, 5 skipping, 0 failing.

---

### Two notes I owe you, neither blocking

These came out of a cross-check that returned after my first review was posted, so they have never
been on this PR and you have had no chance to answer them. I am **not** blocking on either, and
neither needs an amendment to land this record.

- **`docs/adr/0036-the-beta-release-channel.md:350` — D12 carries 0023 D17's retag without naming
  0023 D16's guard.** D12 says images are retagged *"exactly as 0023 D17 does for a promotion"*, but
  a promotion also runs D16's assertion at `container-image.yml:368-378`, which refuses the retag
  unless `org.opencontainers.image.revision` equals the commit being released. The drift is real:
  `beta-x.y.z` moves per merge (0023 D7), so a merge landing while a cut's macOS leg builds can
  leave the published `x.y.z-beta` images on commit Y while the Release's tarballs are commit X.
  D12's word *"verified"* leaves room for the guard, so I read this as **#319's acceptance
  criterion rather than a gap in the record** — worth adding to that ticket.
- **`:61` — C1's "every image tag" reads, literally, as banning `beta-x.y.z`.** The 0023 D8 note in
  this same PR says *"`beta-x.y.z` is still the *image* tag"* and D12 retags *from* it. The coherent
  reading is that `beta-x.y.z` is a branch handle rather than a beta version string, which 0023 D8
  states outright — so this is an editorial nit, not a contradiction. If C1 is ever touched for
  another reason, *"every image tag a beta cut publishes"* would remove the second reading.

(A third, smaller one: D7 at `:269` calls itself *"a fifth published tag class"* while its table adds
both a git class and an image class. Pure arithmetic, and #327 is the one that will care.)

---

### Verdict

**Approved.** Every finding I raised is fixed and verified at the source, the two additions D2 grew
along the way (step 4, and D24) close holes nobody had raised, and the record is materially better
than the version I reviewed. The one blocking item is fully resolved: C4 and D2 now say the same
thing in the same vocabulary, and no row survives with two readings.

Two things this approval does not do, both by design: it does not mark the draft ready, and it does
not ratify the ADR — `:6-7` reserves that to the repository owner, and the record is still
**PROPOSED**.

**Merge order for this wave:** trivial — PR 328 is alone in wave 1. Merge it into
`feat/release-pipeline` first and by itself; every other issue in milestone 0.4.1 reads a clause in
it, and wave 2 (#316, #317) plus the rest start only after it lands. Nothing else needs sequencing
against it.

Issue #326, opened 2026-08-24
Promotion runs a stale release workflow against the train that changed it


mrfqcentic:
Milestone 0.4.1, beside #314 — a beta train must be installable the same way a release is.
Read that issue for the shape of this milestone. This one is not in its dependency chain, but
it is the reason `v0.4.0` has no GitHub Release and no npm publish, which is the state #314
§3 and #320 both assume away.

> **This is a problem statement, not a design.** The cause below is established from the run;
> the fix is not proposed here. It also records a repository that is currently **half-promoted**
> and needs a documented way back.

## What happened

Run `32716466300` promoted `beta/0.4.0`. It got most of the way and then failed:
https://github.com/actana/control/actions/runs/32716466300

| Job | Conclusion |
| --- | --- |
| Approve the promotion | success |
| Resolve the promotion | success |
| Verify the core digest (linux/amd64, linux/arm64) | success |
| Verify the panel digest (linux/amd64, linux/arm64) | success |
| Fast-forward main and push the version tag | success |
| Release / Resolve ref | success |
| **Release / Core tarball (linux-x64)** | **failure** |
| **Release / Core tarball (linux-arm64)** | **failure** |
| **Release / Core tarball (mac-arm64)** | **failure** |
| Release / E2E — installer, Panel image, Core image, npm publish, Attach tarballs | skipped |
| Cut the release line, Delete the promoted train, Cut the next train, Rebase the surviving train | skipped |

So `main` fast-forwarded to `af3bf7b` and `v0.4.0` was pushed — both irreversible in
practice — and then everything a release actually publishes stopped. All three tarball legs
died on the same line:

> ```
> [core-tarball] no actana-cli.cjs at /home/runner/work/control/control/packages/cli/dist-tarball
>  — run `pnpm --filter @actana/cli build` first
> ```
> job `97399482694`, `Release / Core tarball (linux-x64)`, at `RELEASE_VERSION: 0.4.0`

## The cause: the run was executing a `release.yml` older than the train it was releasing

`promote.yml` calls `release.yml` by `workflow_call`. A `workflow_dispatch` run resolves
**every** workflow file — the called one included — from the default branch as it stood when
the run was created, and never from the ref being promoted. This run's `headSha` records that
exactly:

```
$ gh run view 32716466300 --repo actana/control --json headSha
{"headSha":"51164f18f0055d8676cbe10af95d14952e43e3a7"}
```

[...]

Issue #317, opened 2026-08-24
install.sh derives its channel from its own ref — and cannot read that ref at runtime


mrfqcentic:
Part of #314 — a beta train must be installable the same way a release is. Read that issue
first. **This ticket cannot start until #315 has chosen a mechanism**; §4 of #314 is marked
UNSETTLED for the reason set out below.

## The problem

`install.sh` has exactly one default: the newest **release**.

> ```sh
> latest_url="$API_BASE/repos/$REPO/releases/latest"
> ```
> `install.sh:248`

`/releases/latest` excludes prereleases by definition, so once #318 publishes
`v0.4.1-beta` as a prerelease, the default one-liner will never see it. The intended model is
that the URL you fetched the script from decides:

| Want | URL | Installs |
|---|---|---|
| latest release | `raw.githubusercontent.com/actana/control/main/install.sh` | newest release |
| a specific release | `…/actana/control/v0.4.0/install.sh` | 0.4.0 |
| the current beta of a train | `…/actana/control/beta/0.4.1/install.sh` | whatever `v0.4.1-beta` points at now |
| that exact beta | `…/actana/control/v0.4.1-beta/install.sh` | the same, pinned to the tag |

`--version <v>` still overrides. `--repo` and `--base-url` are untouched.

## Why this is not "read the ref at runtime" — the constraint that shapes the ticket

**A script fetched by `curl … | bash` cannot know the URL it came from.** There is no `$0`,
no `BASH_SOURCE`, no argv[0] that names a URL; bash is reading a pipe. Whatever the script
"derives" has to already be in its own bytes. So "the ref is the channel" means *the copy of
`install.sh` on that ref is a different file* — and that runs into the branch model:

**Promotion is a fast-forward** (ADR 0023 D5, `promote.yml`'s `advance` job: *"Fast-forward
main and push the version tag"*). The train tip's bytes **become** `main`'s bytes. A
`CHANNEL=beta` constant committed on `beta/0.4.1` therefore lands on `main` at promotion and
turns `raw.githubusercontent.com/actana/control/main/install.sh` — the public one-liner
printed in `README.md`, `INSTALL.md` and on the landing page — into a beta installer. That is
not a bug to be caught in review; it is what fast-forward means.

**#315 picks one of three routes and this ticket implements it:**

- **(a) A stamped line version, written by the cut.** `promote.yml:724-756` already rewrites
  six manifests in the cut's first commit and `ci.yml`'s `Train rules` asserts them.
[...]

Issue #316, opened 2026-08-24
install.sh installs and stops activating: the Core bundle and the launcher are placed, setup is the next command


mrfqcentic:
Part of #314 — a beta train must be installable the same way a release is. Read that issue
first; the design there is fixed. The record is #315.

## The problem

`install.sh` ends by running `actana setup`:

> ```sh
> set +e
> if [ -t 0 ]; then
>   "$extracted/bin/actana" setup "$@"
> else
>   "$extracted/bin/actana" setup "$@" </dev/null
> fi
> status=$?
> ```
> `install.sh:347-352`

So one line both **installs** a Core bundle and **activates** the machine as a Core: mTLS
material, a systemd unit or LaunchAgent, lingering, the daemon, and registration in this
machine's own registry. There is no way to get the first without the second, and the two are
different acts with different audiences — `npm i -g @actana/cli` (#288 / ADR 0032) already
installs the CLI on a machine that will never host a Core, and the shell one-liner should be
able to do the same for the bundle.

## The decision already taken

**`install.sh` stops activating.** It resolves, downloads, verifies against `SHA256SUMS`,
extracts, **places the bundle and links the launcher**, prints the `actana setup …` line to
run next, and exits 0. There is no flag for this — it is what the script does. Do not add
`--setup`, `--no-setup`, or an environment variable that restores the old tail.

`actana setup` is unchanged in what it does. It is simply no longer run by the script.

## What makes this more than deleting eleven lines

Today the extracted tree is **ephemeral**. `install.sh` unpacks into `mktemp -d` and deletes
it on EXIT (`install.sh:290-295`); the tree only survives because `runActanaSetup` copies it
into place:

> ```ts
> if (replacingTree) installTree(source, installDir);
> pointSymlink(layout.currentLink, installDir);
> pointSymlink(layout.binLink, path.join(layout.currentLink, "bin", "actana"));
> ```
> `packages/cli/src/actana-setup.ts` (`runActanaSetup`)

Remove the `setup` call and nothing is installed at all — the download is deleted by the trap
and the machine is exactly as it was found. So this ticket has to move *placement* (the
versioned `versions/<v>` tree, the `current` symlink, the `~/.local/bin/actana` launcher)
ahead of activation.

**Do not reimplement `resolveActanaLayout` in POSIX sh.** `packages/cli/src/actana-layout.ts`
[...]

PR #329, opened 2026-08-24
chore: delete the automatic train cut

### LINKED ISSUE #325: Delete the automatic train cut

qcentic-dev:
Closes #325.

A train's version is an administrative decision, and the workflow was not the
administrator. `promote.yml`'s `next-train` job guessed `beta/<next-minor>.0` from the
version it had just promoted and pushed a branch nobody had asked for. **Cuts are
manual now: a human names the train branch and cuts it, no job guesses a version, and no
job cuts a train.**

## What is deleted

- `next-train` (`promote.yml:670-790`), its `needs` edges, its App-token guard and its
  manifest rewrite
- the minor-bump arithmetic at `:271-274` and the `next_train` output on `resolve`
  (`:166`, `:282`, `:707`)
- the job's entry in the file's header diagram (`:8`) and the hotfix notice's trailing
  *"No auto-cut: a train already exists (D25)"* (`:265`)

`git grep next_train` and `git grep next-train` now return only ADR amendment text.

## How D25 stays honest without an automatic cut

Deleting the job deletes the thing that was holding the invariant *a train is always
open, so work can always be proposed*. Two jobs hold it now. **Neither cuts anything and
neither names a version**, and they share one issue title verbatim so they dedupe against
each other rather than filing a pair:

- **`promote.yml`'s `train-invariant`** — runs at the end of every promotion. It states
  in the run's own log that the promotion created no branch and computed no version
  (which is the third acceptance criterion), then asks origin whether a `beta/*` exists
  and opens *"No open train: nothing can be proposed until one is cut"* when none does.
- **`housekeeping.yml`'s daily `open-train` chore** — asks the same question every
  morning, so the state cannot be closed away: an issue closed while no train exists is
  filed again on the next tick.

**Neither goes red**, deliberately. Every non-hotfix promotion *ends* with no train open
— that is the correct outcome of a good release, and a job that failed on it would be red
after every one. This file already makes that argument for its other detectors: a
permanently red run is how a team learns that red means nothing.

### 1. ADR 0023's D25 invariant

The ADR is amended in this pull request, not in a follow-up. Four clauses described the
automatic cut and each gains a **dated note appended under it** — the style ADR 0036 just
used. **Nothing is rewritten and nothing is renumbered**, so the clauses #327 needs for
version handling are where it left them.

| Clause | Note |
| --- | --- |
| **D3** | *"a workflow does it"* is retired as the actor; the argument (the version is administrative, and a scripted rewrite is why a cut is not six hand edits) stands. |
| **D22** | *"Cut `beta/x.y.z+1` from `main`"* is a human act and reads as one — it always was during an incident, since the auto-cut declined whenever a train existed. |
| **D25** | The auto-cut is deleted; the invariant is kept and is held by the two jobs above. The *"a default, not a commitment"* sentence goes with the guess. |
| **D39** | The App no longer writes the train-cut commit. Its argument is untouched — what replaces the cut pushes nothing at all. |

D25's note **names every clause that mentions a cut**, including the two that are
unchanged and still true: D7's `beta-x.y.z` row and §H's zero-merge train. A person's
push to `beta/x.y.z` fires `ci.yml`'s `push: beta/**` trigger exactly as a workflow's
push did, so the cut still publishes an image before anything merges into the train.

### 2. ADR 0036 D1 — where the line gets stamped

D1 decides that `install.sh` carries the line and that the value is *"written by the cut
exactly as the six manifests are"*, citing `.github/workflows/promote.yml:724-756` —
**inside the job this deletes**.

**Which of the two options: the record says explicitly that a manual cut stamps it, and
#317 implements it.** The stamping does not move to a new place; the *cut* moved, and the
stamp moved with it, which is what D1 already says.

This is consistent with what ADR 0036 already says, rather than a new decision:

> **D1 in particular is written to survive #325.** It says *the cut* writes the stamp —
> not *`promote.yml`'s `next-train` job*. Whether the cut is a job that runs after a
> promotion or a person following a documented procedure, the stamp is written by
> whatever performs the cut and is asserted on the train by `Train rules` (D4) either
> way. #325 changes who cuts; it does not change what a cut writes.
>
> — ADR 0036 §G

So D1 gains a dated note saying the citation names a deleted job, that the cut is now
[`docs/ci-cd.md` § "Cutting a train"](docs/ci-cd.md#cutting-a-train), and that **a manual
cut stamps the line with #317 implementing it** — which is also where 0036 D6 already put
the installer's diff. The runbook names the stamp among the things a cut has to get
right. **Until #317 lands there is no stamp in `install.sh` and a cut writes six
manifests**, which is the number it wrote before ADR 0036 too. `install.sh` is not
touched by this pull request (#316 owns it this wave, #317 next), and neither is
`ci.yml`'s `MANIFESTS` — 0036 D4 is explicit that the stamp gets its own assertion rather
than joining that set.

### 3. The manifest-set binding

`scripts/__tests__/workflows.test.mjs:773-778` was **the only thing binding the cut's
manifest list to `ci.yml`'s `MANIFESTS`**, and it read `files=()` out of the deleted job.
The binding follows the cut rather than dying with it: the test now reads the `files=()`
array out of `docs/ci-cd.md` § "Cutting a train" — the list a person actually cuts from —
and asserts it equals `MANIFESTS`, with a second assertion that the document contains
exactly one such array so a later one cannot shadow it.

Verified by mutation: dropping `packages/cli/package.json` from the runbook array fails
the test, and restoring it passes. `ci.yml`'s comment and its *"A workspace package is
outside the version set"* error message now point at the runbook instead of at a job that
no longer exists, so a seventh package is still told both lists to extend, by name.

### 4. The evidence, in the record

Recorded under D25 as fact rather than argument, because it was proven today rather than
argued: `beta/0.4.1` was cut **by hand** as `e80ecfc`, from `main` at `af3bf7b`, with all
six manifests rewritten to `0.4.1` and CI green. `next-train` would have produced
`beta/0.5.0` — the promoted version was `0.4.0` and the arithmetic bumps the minor, while
this milestone is a patch — and it would in any case have **cut nothing at all**, because
its first act exits `0` when any train exists and `beta/0.4.0` still existed (#326). Both
paths end with no branch. The invariant was satisfied that day by a person.

## Docs

`docs/ci-cd.md` § "Cutting a train" is one section instead of a path plus a recovery from
it. *"Normally you do not"* and the delete-and-re-cut dance are gone; re-cutting is the
same procedure run again. `AGENTS.md:21` no longer says a train is never cut by hand, and
the release skill's hand-cut trap survives as the *stamp* trap, which is the part that
actually costs an hour.

## Scope

Touches `.github/workflows`, the docs and one test file. **Not touched:** `install.sh`
(#316, #317), `packages/shared/src/semver.ts` (#322), anything under `packages/`. No ADR
clause is rewritten or renumbered — every change is an appended dated note. No new
dispatchable cut workflow: the two jobs added report and file, and cut nothing.

## Checks

- `npx vitest run scripts/__tests__/workflows.test.mjs` — 58 passed, including the new
  `no automatic train cut (ADR 0023 D25, as amended by #325)` block
- the full root suite passes; two files time out at 5s under full parallel load on this
  machine (`install-sh`, `promotion-gate`) and pass on their own — neither touches
  anything in this diff
- `commitlint` over all five commits, not just the titles: 0 problems. Every body line is
  ≤ 100 characters
- all three workflow files parse, and both new `run:` blocks pass `bash -n`


qcentic-dev:
## Acceptance criteria from #325

- [x] **The `next-train` job is gone from `promote.yml`, along with its entry in the
      header diagram (`:8`) and its `needs` edges.**
      The job, its `if:`, its App-token guard, its checkout of `main` and its manifest
      rewrite are deleted. The diagram's `next-train` leaf is replaced by
      `train-invariant`. `rebase-train` keeps `needs: [resolve, release, retire-train]`,
      which never referenced the cut.
- [x] **Nothing computes a next version anywhere: the `next_train` output is removed from
      `resolve` (`:166`, `:282`) and the arithmetic at `:271-274` is deleted.**
      Both gone. `resolve`'s outputs are now asserted as an exact list by
      `no automatic train cut > resolves facts and invents no version`, so an added output
      is how a computed version would come back and the test says so. A second test
      asserts the minor bump appears nowhere in the file, a third that no `git push` line
      names the `beta` class. `release_line` remains and is derived from the train's own
      name (D27), which is reading a fact about the branch handed in, not inventing one.
- [x] **A promotion that succeeds leaves no branch behind that a human did not ask for,
      and says so in its own log.**
      `train-invariant` emits, before it checks anything and whatever the answer:
      *"This promotion created no train. `release/x.y` is the only branch it may have
      created, and `beta/x.y.z` was deleted (ADR 0023 D4). No version was computed
      anywhere in this run."* It also writes the invariant's state to the job summary.
- [x] **The D25 invariant is kept honest without an automatic cut — a real mechanism, not
      a sentence in a runbook.**
      Two jobs, both of which cut nothing and name no version, sharing one issue title
      verbatim so they cannot file a pair: `promote.yml`'s `train-invariant` at the moment
      the gap opens, and `housekeeping.yml`'s daily `open-train` chore for as long as it
      lasts. After a promotion the repository therefore either has an open train or has an
      open issue saying it does not — and closing that issue without cutting buys one day.
      Neither goes red: every non-hotfix promotion *ends* in this state, so a red job
      would be red after every good release.
- [x] **The hand cut is the documented path and reads as the normal one.**
      `docs/ci-cd.md` § "Cutting a train" is one section: *"A person cuts the train, by
      hand, and that is the normal path."* *"Normally you do not"* is gone, the
      *Re-cutting a train, by hand* subsection and its delete-and-re-cut dance are gone
      (re-cutting is the same procedure run again), and the release-flow diagram's first
      box says `(a person)` rather than `(workflow)`. `AGENTS.md:21` now says a train is
      cut by a person and never by a workflow.
- [x] **The manifest-set binding survives.**
      The test that read `files=()` out of the cut job now reads the `files=()` array out
      of `docs/ci-cd.md` § "Cutting a train" — the list a person cuts from — and asserts
      it equals `ci.yml`'s `MANIFESTS`, plus that the document holds exactly one such
      array so a later one cannot shadow it. Mutation-checked: removing
      `packages/cli/package.json` from the runbook array fails the test. `ci.yml`'s
      comment and its *"A workspace package is outside the version set"* error now name
      the runbook, so a seventh package is still told both lists by name.
- [x] **ADR 0023 is amended, not contradicted; every clause that describes the automatic
      cut is named.**
      Dated notes appended under **D3** (the actor is retired, the argument stands),
      **D22** (*"Cut `beta/x.y.z+1`"* restated as a human act), **D25** (auto-cut deleted,
      invariant kept and mechanised, the *"a default, not a commitment"* sentence retired
      with the guess) and **D39** (the App no longer writes the train-cut commit). D25's
      note enumerates every clause that mentions a cut, including the two left alone with
      the reason they stay true — D7's `beta-x.y.z` row and §H's zero-merge train, both of
      which survive because a person's push fires the same `push: beta/**` trigger.
      Nothing rewritten, nothing renumbered.
- [x] **`git grep -n next_train` and `git grep -n 'next-train'` return nothing outside the
      ADR's amendment history.**
      Remaining hits: `0023` D3's and D25's notes, `0036` D1's note and `0036` §G's
      pre-existing paragraph. Code, tests, runbook, `AGENTS.md` and the release skill are
      clean. The new tests are written positively for this reason — a test that grepped
      for the string in order to refuse it would be the one place it survived.

**Out of scope, respected.** No dispatchable cut workflow: the two jobs added report and
file issues, and cut nothing. No version scheme (#327) — the clauses it needs are
untouched and unrenumbered. `retire-train` and `rebase-train` are not touched.

**Parallel-ticket boundaries.** `install.sh` untouched (#316 this wave, #317 next);
`packages/shared/src/semver.ts` and everything under `packages/` untouched (#322); no ADR
clause rewritten or renumbered, only dated notes in ADR 0036's style; nothing merged,
`main` not touched, and this pull request stays a draft.

mrfqcentic (approved):
**Approve.** All eight acceptance criteria in #325 are met, and I checked each of the load-bearing claims against the enforcing line rather than against the description. The operator constraint holds: nothing in the tree computes or guesses a train version, and nothing cuts a train.

## What I verified, and how

**1. Nothing computes a next train version — tree-wide, not just in the diff.**
`git grep next_train` and `git grep next-train` over the whole tree return five hits, all ADR prose: `0023:35` (D3's note), `0023:121`, `0023:125` (D25's note and its evidence), `0036:144` (D1's note) and `0036:590` — §G's paragraph, which predates this PR. Code, tests, runbook, `AGENTS.md` and the release skill are clean. A wider sweep for minor-version arithmetic (`$((minor + 1))`, `${version%%.*}`, `minor + 1` in JS/TS) leaves exactly two hits, neither about trains: `scripts/lib/fixture-release.mjs:292` (`bumpPatch`, a test-fixture helper for the installer e2e "newer release" pair) and `packages/sdk/src/__tests__/core-link-frames.test.ts:510` (the core-link *protocol* compat matrix). Both pre-existing, both out of scope. `promote.yml:275`'s `release_line="release/${version%.*}"` reads the line off the branch it was handed — D27, and correctly annotated as the only derivation left.

The guard is real, not just the deletion: `workflows.test.mjs:710` asserts `resolve`'s outputs as an **exact list** (`["train","version","head_sha","pr","release_line","hotfix","survivor"]` — I re-ran the regex against the head, it matches), so a re-added computed output fails rather than passes silently.

**2. D25's invariant is held by a mechanism that exists.**
`promote.yml`'s `train-invariant` (`:688`) and `housekeeping.yml`'s `open-train` (`:167`). Both cut nothing, name no version, and share one issue title verbatim — I extracted both `TITLE=` literals and they are byte-identical, so they dedupe rather than filing a pair. `open-train`'s `if:` mirrors `stale`'s exactly and the daily cron `17 3 * * *` is present in `on: schedule`. The needs-graph is sound: the only path to "no train open" runs through `retire-train` succeeding, and `train-invariant` needs it — so the report cannot be skipped on the run that opens the gap. A reader can follow it: D25's note names both jobs, and `docs/ci-cd.md` § "Nothing lets you forget" (`:1069`) restates it beside the procedure.

**3. ADR 0036 D1's stamp has an owner and a place.**
D1's citation pointed inside the deleted job. The amended record answers it from the record rather than from this PR's argument: D1's note (`0036:144`) names the cut as `docs/ci-cd.md` § "Cutting a train" with #317 implementing the stamp, and the runbook carries it as a named obligation — `docs/ci-cd.md:1038`, "**The line stamp, once it exists.**". §G's prediction holds *as written*: it said "the stamp is written by whatever performs the cut… #325 changes who cuts; it does not change what a cut writes", and that is exactly what happened — the cut still writes six manifests. The interim state is stated three times, not left implicit (`0036:144`, `0023:35`, `docs/ci-cd.md:1043`): until #317 lands there is no stamp and a cut writes six files.

**4. The manifest binding is still enforcing.**
`docs/ci-cd.md:1000` holds exactly one line-anchored `files=()`, and it is the same set as `ci.yml`'s `MANIFESTS` — I parsed both and compared sorted. I mutation-checked the assertion at `workflows.test.mjs:868` in memory: dropping `packages/cli/package.json` from the runbook array fails it, and appending a second `files=()` anywhere in the document fails the count assertion. It fails safe in both directions. The chain still closes — `MANIFESTS` ↔ workspace (`:809`), ↔ `ci.yml`'s job (`:817`), ↔ the runbook array (`:868`) — so a seventh package cannot reach one list and not the other.

**5. The evidence checks out against the live repository.**
`beta/0.4.1` is `e80ecfc`, parent `af3bf7b`, which is `origin/main`. The diff `af3bf7b..e80ecfc` is exactly six files, six lines, all six manifests at `0.4.1`. CI on `e80ecfc`: 20 success, 5 skipped, 0 failures. `beta/0.4.0` still exists on origin, so the deleted job's first act (`exit 0` when any train exists) would have cut nothing; and the promoted version `0.4.0` would have produced `beta/0.5.0`. Both halves of the claim are true.

**6. Amendments are additive.**
`git diff --numstat` on both ADRs: `16 0` and `8 0` — zero deletion lines. Four dated notes in 0036's style, under D3, D22, D25 and D39, plus §G's "the paragraph above held". Nothing renumbered, nothing rewritten, so #327's premises (D3's "four manifests", D7's count) read exactly as #327 quoted them.

**7. Boundary respected.** Nine files, none under `packages/`, none `install.sh`, none `semver.ts`. The branch is level with `feat/release-pipeline` (merge-base == base tip). CI on `d13c691` is green.

## Follow-ups — none blocking, in the order I'd take them

**`docs/ci-cd.md:1015` — the cut's verification loop only prints.** This is the one I'd most want changed. The deleted job rewrote the six manifests and *then asserted* each one, `exit 1`-ing on a mismatch. The runbook keeps the rewrite but downgrades the check to `for file in "${files[@]}"; do jq -r --arg f "$file" '"\($f): \(.version)"' "$file"; done` — six lines printed, and nothing fails. The rewriter at `:1005` also dropped the deleted job's `changed` counter and its `if (after === before) continue`, so a manifest whose `"version"` line does not match the regex is a silent no-op. `Train rules` is `pull_request`-only, so nothing catches it on the push, and this is precisely the "six errors on whoever opens the first PR" trap the same section warns about at `:985`. You do have two human checks — the printed values and the "**The diff is only the cut**" bullet at `:1036` — so this is not unguarded, but turning that loop into a comparison that exits non-zero costs two lines and restores the guarantee the job used to give. I am not blocking on it because #325 explicitly warns against letting "delete the automatic cut" become "build a slightly better one", and this is a judgement call that belongs to you.

**`.github/workflows/promote.yml:751` — `::error` on the expected outcome of every good promotion.** The job's own header (`:668-672`) argues that this state must not read as broken, because "every non-hotfix promotion ends with no train open". The no-train path then emits `::error title=No open train`, which renders as a red error annotation on the run summary of every successful release. The sibling chore uses `::warning` for the identical state (`housekeeping.yml:208`). The new test only forbids `exit 1`, so the inconsistency slips past it. `::warning` would match the argument and the sibling.

**`docs/ci-cd.md:18` — the "At a glance" daily-cron row was not updated.** The `promote.yml` row in that table was corrected in this diff and `housekeeping.yml`'s header now says "Eight chores", but the `housekeeping.yml | daily cron` row still reads only "stale labels / closures". That table is the one place a reader looks to find out what runs daily, and `open-train` is now half the D25 mechanism.

**`docs/ci-cd.md:1017` — two small gaps in the procedure.** `git commit -a -F cut-message.txt` appears before anything tells the reader to write `cut-message.txt`; the template is ~46 lines further down at `:1063` and the comment does say "see below", so this is signposting rather than a hole. Separately, the deleted job carried an explicit branch for "the manifests already read `$version`", because `git commit -a` exits non-zero on an empty commit. A person re-cutting a name `main` already carries now hits `nothing to commit` mid-procedure with no guidance — the state is in fact fine and only the push remains, which is worth one sentence.

**`.github/workflows/promote.yml:697` — optional hardening.** `train-invariant` is the first job to check out *after* `retire-train` deletes the train branch, and it uses a bare `actions/checkout` while the job it replaces pinned `ref: main`. If someone dispatches the promotion from the UI with the train selected in the branch dropdown, checkout resolves a ref that no longer exists. In practice the tip is reachable from `main` after the fast-forward so the SHA fetch should still succeed, and bare checkout is this file's norm (`:567`, `:624`, `:864`) — so I could not confirm a failure, only that `ref: main` is free insurance on a job that needs no tree at all, just an authenticated `origin` for `ls-remote`.

## Notes, not asks

- `gh issue create` is unguarded under `set -euo pipefail` at `promote.yml:789`, so an API blip would redden an otherwise perfect promotion. That is the established house pattern (`rebase-train`'s `gh pr comment` at `:956` predates this PR, as do all four chores), so raising it here would single this PR out for a repo-wide convention. Same for the `--search "… in:title"` dedupe, which depends on the search index — both sites post-filter on an exact `.title` match in `jq`, so the worst case is a duplicate, never a wrong match.
- `workflows.test.mjs:721` guards the minor bump only inside `promote.yml` and only at the exact spacing `minor + 1`. AC 2 says "anywhere"; my tree-wide grep is clean today, so this is coverage, not a defect.
- The AC comment says the *Re-cutting a train, by hand* subsection "is gone". It survives as `#### Re-cutting a train` at `docs/ci-cd.md:1091`, reframed to "this is the same cut, done again". That satisfies the criterion — "Normally you do not" and the delete-and-re-cut dance are both gone — and keeping a short pointer is better than deleting it; worth correcting the claim so the record matches the diff.
- `docs/ci-cd.md:57` still says `weekly` covers "the four that share the Monday tick" while `:76` says "the six". Pre-existing in `housekeeping.yml`, untouched here, mentioned only so it is not mistaken for new drift.

The PR is still a draft; flipping it is yours to do.

Issue #314, opened 2026-08-24
A beta train must be installable the same way a release is


mrfqcentic:
> **The `0.4.1` train.** Read off `origin/beta/0.4.0` at `8aa5b5b` and `main` at `51164f1`
> — 0.4.1 is cut from `main` *after* 0.4.0 promotes, so every file reference below names
> the tree the 0.4.0 train leaves behind (`packages/cli/src/…`, not the deleted
> `packages/core/src/actana-cli.ts`).

A beta train publishes container images and **nothing else**. `release.yml` cuts the Core
tarballs, `SHA256SUMS`, the npm packages and the GitHub Release only for a tag its `resolve`
job can prove is reachable from `main`:

> ```
> if git merge-base --is-ancestor "$sha" origin/main; then mode=promote
> …
> echo "::error title=Tag is on neither main nor a release line::… there is neither a beta
> digest to promote nor a release line to build from."
> ```
> `.github/workflows/release.yml:219-231`

So a train that is finished and release-ready — as `beta/0.4.0` is today — cannot be
installed on metal at all. The only ways to try it are Docker, or building the tarball on
the machine that will run it. Neither is a product. **Everything a merge from beta into
`main` gets must also be available, in a lighter form, for the beta itself.**

This issue restates the agreed design so the task issues can point at it. It is not to be
redesigned in the implementing PRs — with the one exception marked **UNSETTLED** below,
which #315 settles before anything downstream of it is written.

## 1. Install is not activation

Three surfaces, three jobs:

| Command | Installs | Activates |
|---|---|---|
| `install.sh` | the Core bundle **and** the `actana` CLI | no |
| `npm i -g @actana/cli` | the CLI only — for a machine that drives Cores it does not host | no |
| `actana setup` | — | yes: mTLS material, service unit, daemon, registration |

**`install.sh` no longer runs `setup`.** Today it execs `actana setup` as its last act
(`install.sh:347-356`), so one line both installs and activates. That tail is removed: the
script resolves, downloads, verifies against `SHA256SUMS`, extracts, places the bundle,
links the launcher, prints the `actana setup …` line to run next, and exits. There is no
flag for this — it is what the script does.

## 2. The trigger is a human asking for a beta release

Not every merge into the train. A train takes many small feature-branch PRs, and publishing
a release per merge would be both noise and cost. A beta release is **requested**, the way a
promotion is:

- new workflow `.github/workflows/beta-release.yml`, `workflow_dispatch` only,
[...]

Issue #318, opened 2026-08-24
beta-release.yml — a requested beta cut: the moving tag, the prerelease, three tarballs, SHA256SUMS and install.sh


mrfqcentic:
Part of #314 — a beta train must be installable the same way a release is. Read that issue
first; the design there is fixed. The record is #315.

## The problem

Nothing publishes a beta except images. `release.yml` is the only thing that cuts tarballs, a
`SHA256SUMS` and a GitHub Release, and its `resolve` job refuses any tag that is not
reachable from `main` or on a `release/*` branch:

> ```
> echo "::error title=Tag is on neither main nor a release line::$ref ($sha) is not reachable
> from main and is on no origin/release/* branch, so there is neither a beta digest to promote
> nor a release line to build from."
> ```
> `.github/workflows/release.yml:227-229`

A beta tag lives on the train, which is neither. Teaching `release.yml` a third mode was
considered and rejected: its two modes are both *promotion-shaped* (a tag, immutable, with
`latest` in play), and a beta is dispatch-shaped, moving, and never `latest`.

## The decision already taken

A new workflow, `.github/workflows/beta-release.yml`:

- **`workflow_dispatch` only.** A beta release is requested, the way a promotion is
  (ADR 0023 D14). Merges into the train keep doing exactly what they do today — the checks
  and the moving `beta-0.4.1` images — and publish nothing.
- **One input**: the train branch, `beta/0.4.1`. Validate it the way `promote.yml:130-135`
  does (`beta/x.y.z`, and the version is the branch name); the version to publish is
  `<branch>-beta`, e.g. `0.4.1-beta`.
- **The git tag `v0.4.1-beta` moves** to the train tip on each cut. Force-update the ref
  deliberately and say so in the log, with the previous sha named.
- **The version string is `<train version>-beta`, exactly, on every surface this workflow
  touches** — the tag, the Release title and tag name, the three asset filenames, the string
  passed to `scripts/build-core-tarball.mjs --version`, and the image tag #319 adds. **No
  counter, no `.N`, no suffix of any kind.** This is a fixed operator constraint recorded in
  #315, not a default to be improved on: a run number, a short sha or a `.1` appended anywhere
  is a defect, not a refinement.
- **A GitHub Release, `prerelease: true`, never `latest`, assets clobbered in place.**
- **Three tarballs**: `linux-x64`, `linux-arm64`, `mac-arm64` — built at version
  `0.4.1-beta`, each smoke-tested on a runner of its own architecture, exactly as
[...]

PR #330, opened 2026-08-24
fix(cli): actana update no longer walks a machine off its beta, and the notice fires for its line

### LINKED ISSUE #322: actana update silently downgrades a machine off a beta, and the update notice never fires for its release

qcentic-dev:
Closes #322.

**Base is `fix/update-channel-awareness`** — the feature branch for #314, not `main` and not the
beta train. This is the only ticket in the 0.4.1 milestone on that branch; ADR 0036 lives on
`feat/release-pipeline` and is read from there, not from this base.

## The bug

`isNewerSemver` compared only the numeric core, so `0.4.1` and `0.4.1-beta` were *equal*. One
answer, two bugs:

1. **The update notice never fired for a beta's own release.** `checkForUpdate` asked the channel
   for the latest release, got `0.4.1`, and `isNewerSemver("0.4.1", "0.4.1-beta")` was `false`. The
   operators who had installed a beta on metal — the population most in need of the notice — were
   never told the real release existed.
2. **`actana update` silently downgraded.** A bare update resolves `/releases/latest`, which
   excludes prereleases, so it answered `0.4.0`. The already-current guard compares two strings for
   equality, `"0.4.0" !== "0.4.1-beta"`, so it waved that straight through: a stop, a tree swap and
   a restart onto an older version, reported to the operator as an update.

## What changed

**`packages/shared/src/semver.ts`** learns semver §11.4 precedence — a prerelease is below its own
release, and identifiers compare numerically rather than as strings. Build metadata is ignored for
precedence (semver §10), which keeps every answer the function gave for a plain `x.y.z` pair, and
an unparseable string still answers `false` rather than throwing.

**`scripts/__tests__/version-rule-agreement.test.mjs`** is new, and is what stops the two
implementations of this one rule drifting apart (ADR 0036 D8). They cannot share a file — one is
`.mjs` that node runs inside a workflow step with no build, the other ships in the Core tarball and
the Panel bundle — so they share a test. The root suite is the only project that spans both
`scripts/**` and `packages/**`. It runs `compareReleaseVersions` and `isNewerSemver` over every
ordered pair of a 16-version table and names any pair they answer differently. Verified by
mutation: reverting either half of the rule turns it red with the offending pairs listed.

**`packages/shared/src/actana-release-channel.ts`** gains the *line* vocabulary, which ADR 0036 D6
assigns to this ticket. A **line** is `x.y.z` and resolves to either its release or its beta (D1,
D2). It is deliberately not a second meaning for `ReleaseChannel`, which already means *which
repository and which hosts releases are read from*. `resolveLine` is D2's rule — a pin, then
`v<line>`, then `v<line>-beta`, then `/releases/latest` as the terminal fallback. Its inputs are
existence answers about two *named* tags, so no step enumerates releases and it cannot hand a
machine installing one line's beta the beta of another. D6: **#322 creates, #317 consumes.**

**`packages/cli/src/actana-update.ts`** gains the downgrade guard. A bare update refuses a resolved
version older than the installed one, downloads nothing, and says what the machine is on, what the
newest release is, and that a bare update will take its own line's release the day that exists. An
explicit `--version` is never second-guessed — pinning an older release is how a Panel↔Core version
lock is recovered. The guard is an ordering comparison, not a rule about prereleases, so a `latest`
moved backwards is caught for a machine on a release too, and a beta machine is not *stranded*.

**`packages/core/src/core-update-notice.ts`** needed no behaviour change: the semver fix reaches it
through `checkForUpdate`. Nothing special-cases a beta, which is the property worth pinning, so it
gains the coverage and a header note saying so.

## On the version shape

A beta version string is exactly `x.y.z-beta` on every surface and carries no counter (ADR 0036 D1,
D7 — the tag *moves* per cut, so the string never changes). Nothing here is built around ordering
counters. The numeric-identifier branch of §11.4 is therefore covered by the backport release
candidates ADR 0023 D30 publishes (`rc.2` vs `rc.10`), not by beta-shaped fixtures.

## Boundaries

Touches only this ticket's surfaces: `semver.ts`, `actana-release-channel.ts`, `actana-update.ts`,
`actana-release.ts`, `core-update-notice.ts`, their tests, and the new root-suite agreement test.
No `install.sh` (#316, #317), no `.github/workflows` (#325), no CLI install/setup/layout/tree
modules (#316). Nothing merged, `main` untouched, and this stays a draft.

### DIFF
--- a/packages/cli/src/__tests__/actana-update.test.ts
+++ b/packages/cli/src/__tests__/actana-update.test.ts
@@ -384,3 +384,110 @@ describe("releases this machine cannot use", () => {
     await expect(update()).rejects.toThrow(/releases\/latest|no release/i);
   });
 });
+
+// #322. `/releases/latest` cannot see a prerelease, so on a machine installed
+// from a beta a bare `actana update` resolves the *previous* release. Before
+// the guard, "0.4.0" !== "0.4.1-beta" waved straight past the already-current
+// check and the machine was stopped, swapped and restarted onto an older
+// version — and told it had been updated.
+describe("a machine running a beta", () => {
+  const BETA = "0.4.1-beta";
+
+  beforeEach(() => {
+    existingInstall(BETA);
+  });
+
+  describe("when the newest release is older than the beta", () => {
+    beforeEach(() => {
+      writeRelease({ dir: releaseDir, version: "0.4.0", target: TARGET });
+    });
+
+    it("does not replace the tree", async () => {
+      const service = fakeService();
+      const result = await update({ service });
+
+      expect(result.updated).toBe(false);
+      expect(result.version).toBe(BETA);
+      expect(fs.realpathSync(layout.currentLink)).toBe(
+        fs.realpathSync(installDirFor(layout, BETA)),
+      );
+      expect(fs.existsSync(installDirFor(layout, "0.4.0"))).toBe(false);
+      expect(readActanaConfig(layout.configDir)!.version).toBe(BETA);
+      // The daemon never noticed: no stop, no restart.
+      expect(service.verbs).toEqual([]);
+      expect(service.stops).toBe(0);
+    });
+
+    it("says what it is on, what the release is, and what to do about it", async () => {
+      await update();
+      const said = out.join("\n");
+      expect(said).toContain(BETA);
+      expect(said).toContain("prerelease");
+      // The line, in ADR 0036 D1's vocabulary — the release this machine is
+      // actually waiting for is its own line's.
+      expect(said).toContain("0.4.1 line");
+      expect(said).toContain("0.4.0");
+      expect(said).toContain("actana update --version 0.4.0");
+    });
+
+    it("downloads nothing — the refusal is a decision, not a failed attempt", async () => {
+      const fetcher = fixtureFetcher(releaseDir, CHANNEL);
+      await update({ fetcher });
+      expect(fetcher.asked.some((url) => url.endsWith(".tar.gz"))).toBe(false);
+      expect(fetcher.asked.some((url) => url.endsWith("SHA256SUMS"))).toBe(false);
+    });
+
+    // An explicit pin is an operator's decision and is not second-guessed:
+    // this is how a Panel↔Core version lock is recovered.
+    it("still installs an older release when the operator pins it", async () => {
+      const result = await update({ requestedVersion: "0.4.0" });
+
+      expect(result.updated).toBe(true);
+      expect(result.version).toBe("0.4.0");
+      expect(result.previousVersion).toBe(BETA);
+      expect(fs.realpathSync(layout.currentLink)).toBe(
+        fs.realpathSync(installDirFor(layout, "0.4.0")),
+      );
+      expect(readActanaConfig(layout.configDir)!.version).toBe("0.4.0");
+    });
+  });
+
+  // The other direction, and the reason the guard is a comparison rather than
+  // "never move off a prerelease": the beta's own line releasing is exactly
+  // what this machine is waiting for, and it must not be stranded.
+  it("takes its own line's release the day it exists", async () => {
+    writeRelease({ dir: releaseDir, version: "0.4.0", target: TARGET });
+    writeRelease({ dir: releaseDir, version: "0.4.1", target: TARGET });
+
+    const result = await update();
+    expect(result.updated).toBe(true);
+    expect(result.previousVersion).toBe(BETA);
+    expect(result.version).toBe("0.4.1");
+    expect(readActanaConfig(layout.configDir)!.version).toBe("0.4.1");
+  });
+
+  it("takes a later line's release too", async () => {
+    writeRelease({ dir: releaseDir, version: "0.4.2", target: TARGET });
+    expect((await update()).version).toBe("0.4.2");
+  });
+});
+
+// The guard is about ordering, not about prereleases: a `latest` that has been
+// moved backwards would walk a release-running machine back just as silently.
+describe("a bare update that would move any machine backwards", () => {
+  beforeEach(() => {
+    existingInstall("0.5.0");
+    writeRelease({ dir: releaseDir, version: "0.4.0", target: TARGET });
+  });
+
+  it("refuses, and does not call the installed version a prerelease", async () => {
+    const result = await update();
+    expect(result.updated).toBe(false);
+    expect(result.version).toBe("0.5.0");
+    const said = out.join("\n");
+    expect(said).toContain("0.5.0");
+    expect(said).toContain("0.4.0");
+    expect(said).not.toContain("prerelease");
+    expect(said).toContain("actana update --version 0.4.0");
+  });
+});
--- a/packages/cli/src/actana-release.ts
+++ b/packages/cli/src/actana-release.ts
@@ -30,12 +30,19 @@ import type { ReleaseFetcher } from "@actana/shared/actana-release-fetch";
 export { nodeReleaseFetcher, type ReleaseFetcher } from "@actana/shared/actana-release-fetch";
 
 export {
+  BETA_SUFFIX,
+  betaVersionForLine,
   DEFAULT_API_BASE,
   DEFAULT_DOWNLOAD_BASE,
   DEFAULT_REPO,
+  isBetaVersion,
   latestReleaseUrl,
+  lineOf,
   parseLatestTag,
   releaseChannel,
+  releaseTagUrl,
+  resolveLine,
+  type LineResolution,
   type ReleaseChannel,
 } from "@actana/shared/actana-release-channel";
 
@@ -106,6 +113,19 @@ export function sha256OfFile(filePath: string): string {
  * A pinned version costs no API call — an update that still asked what the
  * latest release was would be one API change away from quietly installing
  * something else than the operator named.
+ *
+ * **`/releases/latest` cannot see a prerelease**, by GitHub's definition of
+ * that endpoint, so on a machine installed from a beta this answers with the
+ * *previous* release — `0.4.0` for a machine running `0.4.1-beta`. That is not
+ * corrected here, because it is the right answer to the question this function
+ * asks ("what is the newest release?"). Deciding that installing it would move
+ * the machine backwards is a different question, and it is answered by
+ * `runActanaUpdate`'s downgrade guard in `actana-update.ts` (#322).
+ *
+ * Resolving a *line* to its release or its beta is a third question again —
+ * ADR 0036 D2's rule, spelled as `resolveLine` in the shared channel module
+ * and re-exported above. `actana update` does not use it yet: teaching it a
+ * beta channel of its own is explicitly not #322.
  */
 export async function resolveReleaseVersion(
   fetcher: ReleaseFetcher,
--- a/packages/cli/src/actana-update.ts
+++ b/packages/cli/src/actana-update.ts
@@ -30,11 +30,13 @@ import { writeActanaConfig, type ActanaConfig } from "./actana-config.ts";
 import { installDirFor, type ActanaLayout } from "./actana-layout.ts";
 import type { CoreManifest } from "./actana-manifest.ts";
 import {
+  lineOf,
   releaseTargetFor,
   resolveReleaseVersion,
   type ReleaseChannel,
   type ReleaseFetcher,
 } from "./actana-release.ts";
+import { isNewerSemver, isPrereleaseVersion } from "@actana/shared/semver";
 import { fetchVerifiedRelease } from "./actana-fetch-release.ts";
 import type { ActanaServiceManager } from "./actana-service.ts";
 import type { ActanaSystem } from "./actana-system.ts";
@@ -95,6 +97,34 @@ export async function runActanaUpdate(opts: UpdateOptions): Promise<UpdateResult
   }
 
   const version = await resolveReleaseVersion(opts.fetcher, opts.channel, opts.requestedVersion);
+
+  // The downgrade guard (#322). A bare update resolves `/releases/latest`,
+  // which excludes prereleases, so a machine installed from `0.4.1-beta` is
+  // answered with `0.4.0` — an older version, and one the already-current
+  // guard below waves straight through because the two strings simply differ.
+  // What followed was a stop, a tree swap and a restart, reported to the
+  // operator as an update.
+  //
+  // Nothing is changed and nothing is downloaded: the refusal is a sentence on
+  // stdout and `updated: false`, which is the same shape "already current"
+  // returns and which `actana update` renders as exit 0. Silence was the other
+  // option and is worse — an operator who ran `update` is owed the reason it
+  // did nothing.
+  //
+  // An explicit `--version` is never second-guessed. Pinning an older version
+  // is how a Panel↔Core version lock is recovered, and an operator who named
+  // one has already made this decision.
+  if (!opts.requestedVersion && isNewerSemver(config.version, version)) {
+    opts.out(refusedDowngrade(config.version, version));
+    return {
+      updated: false,
+      version: config.version,
+      previousVersion: config.version,
+      installDir: config.installDir,
+      listening: null,
+    };
+  }
+
   const installDir = installDirFor(layout, version);
 
   // "Already current" is about the tree, not just the recorded version: an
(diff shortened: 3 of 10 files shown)


qcentic-dev:
## Acceptance criteria, #322

Every criterion from the issue, in its order, each checked or explicitly unchecked with the reason.

- [x] **The beta-shaped fixtures are exactly the strings that can occur** under `x.y.z-beta` with no
      counter, no `.N` and no suffix. All six pairs are asserted in
      `packages/shared/src/__tests__/semver.test.ts` → *"a beta and its line"*:
  - [x] `isNewerSemver("0.4.1", "0.4.1-beta") === true` — a release is newer than its own beta
  - [x] `isNewerSemver("0.4.1-beta", "0.4.1") === false` — and never the other way round
  - [x] `isNewerSemver("0.4.1-beta", "0.4.0") === true` — a beta is newer than the last release
  - [x] `isNewerSemver("0.4.0", "0.4.1-beta") === false` — the downgrade this ticket exists to stop
  - [x] `isNewerSemver("0.4.2", "0.4.1-beta") === true` — the next line's release still wins
  - [x] `isNewerSemver("0.4.1-beta", "0.4.1-beta") === false` — a re-cut moving tag is the *same*
        version string and is not an upgrade
- [x] **Numeric-identifier precedence (§11.4) covered by fixtures that are not beta-shaped.**
      `isNewerSemver("1.2.4-rc.2", "1.2.4-rc.10") === false` is asserted, with the reverse and three
      further §11.4 branches (numeric ranks below alphanumeric; a shorter run of otherwise equal
      identifiers is lower; alphanumerics by ASCII), in *"numeric identifier precedence (semver
      §11.4)"*. A `-beta` string has one identifier and no counter, so it can never reach that
      branch — the backport release candidate of ADR 0023 D30 is the shape that does.
- [x] **`isNewerSemver` still answers exactly as today for every plain `x.y.z` pair, and still
      returns `false` for anything unparseable rather than throwing.** *"plain x.y.z pairs answer
      exactly as they did"* covers major/minor/patch ordering, numeric-not-string components and
      build metadata (ignored per semver §10, which is what preserves today's answers). *"anything
      unparseable"* covers `not-a-version`, `1.2`, `1.2.3.4`, `""`, and asserts no throw on
      `undefined`. The three pre-existing assertions in the file are unchanged and still pass.
- [x] **A bare `actana update` on a machine running `0.4.1-beta`, with `0.4.0` as the newest
      release, does not replace the tree, and says what to do about it.**
      `packages/cli/src/__tests__/actana-update.test.ts` → *"a machine running a beta"*: `current`
      still points at the beta tree, `0.4.0` is never installed, `actana.json` is unchanged, and
      the service is neither stopped nor restarted. A second test asserts the sentence names the
      installed version, the word *prerelease*, the **0.4.1 line**, the newer release `0.4.0`, and
      the exact pinning command. A third asserts nothing is downloaded — no tarball, no
      `SHA256SUMS` — so the refusal is a decision, not a failed attempt.
- [x] **`actana update --version 0.4.0` on that machine still works.** *"still installs an older
      release when the operator pins it"* — `updated: true`, `current` repointed at `0.4.0`,
      `previousVersion` `0.4.1-beta`, config rewritten. The guard is skipped whenever
      `requestedVersion` is set; an explicit pin is not second-guessed.
- [x] **A machine running `0.4.1-beta` is told by the update notice when `0.4.1` is released, once,
      on the existing throttle.** `packages/core/src/__tests__/core-update-notice.test.ts` → *"a
      Core running a beta"*: the line fires naming `0.4.1` and `0.4.1-beta`; three ticks inside
      `UPDATE_CHECK_TTL_MS` produce exactly one line through the unchanged `alreadySaid` /
      `updateNoticeStatePath` bookkeeping. Three companions pin the silences — an older release,
      a re-cut beta republishing the same string — and that a later line's release still speaks.
      No code in `core-update-notice.ts` special-cases a beta; the semver fix reaches it through
      `checkForUpdate`.
- [x] **The rule in `packages/shared/src/semver.ts` and the rule in `scripts/lib/release-latest.mjs`
      agree, and something fails when they stop agreeing.** New
      `scripts/__tests__/version-rule-agreement.test.mjs` runs `compareReleaseVersions` and
      `isNewerSemver` over all 256 ordered pairs of a 16-version table and asserts zero
      disagreements, listing any it finds by pair and by answer. `isPrereleaseVersion` is
      cross-checked against `isPrerelease` over the same table. The root vitest project is the only
      one that spans `scripts/**` and `packages/**`, which is why it lives there. **Verified by
      mutation, not by assumption:** removing the *"a prerelease is below its own release"* clause
      turns it red with 14 named pairs; reverting identifier comparison to string ordering turns it
      red with 4. `scripts/lib/release-latest.mjs` itself is unmodified — ADR 0036 D8 records it as
      already correct. The two deliberate domain differences (the script half rejects `+build` and
      throws on unparseable input; the product half ignores `+build` per §10 and answers `false`)
      are asserted by name so they cannot drift silently either.
- [x] **Unit tests cover every pair above; no e2e change is required.** No e2e file is touched.
      47 unit tests added across five files: `semver.test.ts` 4 → 18, the new
      `actana-release-channel.test.ts` 14, the new `version-rule-agreement.test.mjs` 7,
      `actana-update.test.ts` 21 → 28, `core-update-notice.test.ts` 10 → 15.

**Nothing is unchecked.**

---

### Verification run on this machine

| Suite | Result |
| --- | --- |
| root (`scripts/**`, 20 files) | **550 passed** |
| `packages/shared` — `semver`, `actana-release-channel`, `actana-update-check` | **54 passed** |
| `packages/cli` — `actana-update.test.ts` | **28 passed** |
| `packages/core` (full, 56 files) | **870 passed** |
| `packages/cli` (full, 51 files) | 1002–1005 passed, **4–5 flaky failures** |
| `eslint` over all 10 changed files | clean |
| `tsc --noEmit` — sdk, shared, cli, core | clean |
| `tsc --noEmit` — panel | **could not run** |

Two honest caveats, neither of them this diff:

- The **flaky CLI failures are a different set on every run** (`actana-setup`, `actana-pair`,
  `actana-machine-cli`, `events-command`, `project-files-live`) and every one of them passes when
  its file is run alone. `actana-update.test.ts` passed on every run. I confirmed the flakiness is
  not mine by stashing the whole change and re-running the same files on the untouched base — same
  pass. The box has ~0.8 GB of 32 GB free with another agent working on it, and these are the
  cert-minting and real-`tar` suites.
- **`packages/panel`'s typecheck was OOM-killed** (exit 137) on this box, twice, including with a
  capped heap. It is not reachable from this diff: the Panel imports
  `@actana/shared/actana-update-check`, which is unmodified, and `semver.ts`'s exported signatures
  are unchanged — the only addition is a new export. Worth a green CI run to confirm rather than
  taking my word for it.

### Scope

Touches only this ticket's surfaces: `semver.ts`, `actana-release-channel.ts`, `actana-update.ts`,
`actana-release.ts`, `core-update-notice.ts`, their tests, plus the new root-suite agreement test.
No `install.sh` (#316, #317), no `.github/workflows` (#325), no CLI install/setup/layout/tree
modules (#316). Nothing merged, `main` untouched, PR left as a draft.

One judgement call worth flagging for review: **ADR 0036 D6 assigns
`packages/shared/src/actana-release-channel.ts` to this ticket** and names the additions as
*"prerelease-aware comparison and a resolution that does not read `/releases/latest`"*, with
"#322 creates; #317 consumes". So `resolveLine` (D2's rule) lands here **pure and tested but not
yet wired into `actana update`** — teaching `actana update` a beta channel of its own is explicitly
out of scope for #322, and #317's `install.sh` is the consumer D6 has in mind. If you would rather
that function landed with its consumer instead, it is one self-contained commit to drop.

mrfqcentic (approved):
## Approve — every acceptance criterion in #322 is met, proven at the enforcing line

Reviewed the diff against `fix/update-channel-awareness` (merge-base `e80ecfc`, head `0b44fa9`, 4 commits, 10 files, +832/-12). ADR 0036 read from `origin/feat/release-pipeline`, as this PR says to.

I checked the claims against the code rather than the description, and re-derived the comparison independently outside the repo.

### The two bugs, at the line that now enforces the fix

**Downgrade.** `packages/cli/src/actana-update.ts:117` — `if (!opts.requestedVersion && isNewerSemver(config.version, version))` returns before `installDirFor` at `:128`, so nothing is downloaded and nothing is swapped. Proven by `packages/cli/src/__tests__/actana-update.test.ts` → *"a machine running a beta › when the newest release is older than the beta › does not replace the tree"*, which asserts `current` still points at the beta tree, `versions/0.4.0` was never created, the config is unchanged, and the service recorded no verbs and no stops; *"downloads nothing"* additionally asserts no `.tar.gz` and no `SHA256SUMS` request. The pin remains open — *"still installs an older release when the operator pins it"*.

**The notice.** `packages/shared/src/actana-update-check.ts:148` — `updateAvailable: latest !== null && isNewerSemver(latest, current)` is the whole path; `core-update-notice.ts` special-cases nothing. With the fix `isNewerSemver("0.4.1", "0.4.1-beta")` is `true`, so the notice fires. Proven by `packages/core/src/__tests__/core-update-notice.test.ts` → *"a Core running a beta › is told when its own line's release lands"*, with *"is told once, on the same throttle as everyone else"* pinning the existing `alreadySaid` throttle.

### The comparison

`packages/shared/src/semver.ts:119-135` is `compare`: major/minor/patch numerically, then `left.prerelease === right.prerelease → 0`, then release-above-its-own-prerelease, then §11.4 identifiers. It is a line-by-line mirror of `compareReleaseVersions` in `scripts/lib/release-latest.mjs:108-121`, which is untouched — correct, per ADR 0036 D8.

I extracted `semver.ts` to a scratch directory and ran it under node: all six beta-shaped pairs from the ticket answer as specified, `rc.2 < rc.10` holds, every plain `x.y.z` answer is unchanged, unparseable answers `false`, `undefined` does not throw, and the 16-version table is a strict total order. Against the real `release-latest.mjs` the two agree on all **256 ordered pairs** and all 16 prerelease answers.

`scripts/__tests__/version-rule-agreement.test.mjs` is not decoration: it matches the root project's `include: ["scripts/**/*.test.mjs"]` and ran green in CI as 7 tests. I mutation-tested it both ways — neutering the prerelease clause in `semver.ts` yields four named disagreements, and swapping the script half's numeric branch for a string compare yields the `rc.2`/`rc.10` pair. The PR's mutation claim holds in both directions.

### The version shape

`grep -Ei 'beta[._-][0-9]'` over the added lines returns exactly **one** hit: `packages/shared/src/__tests__/actana-release-channel.test.ts:85`, `expect(isBetaVersion("0.4.1-beta.1")).toBe(false)` — a rejected input. Nothing here orders counters; `BETA_SUFFIX` is the constant `-beta`, and the numeric branch of §11.4 is covered by `rc.2`/`rc.10` as ADR 0036 C1 requires.

### ADR 0036 D6, D2

`ReleaseChannel` at `actana-release-channel.ts:23` is untouched — no `stable | beta` field, no third host. The new concept is a **line** (`lineOf`, `betaVersionForLine`, `LineResolution`, `resolveLine`), which is D1/D2's vocabulary. The two meanings stay distinguishable.

No resolution path added here lists releases. The only URL builders are `latestReleaseUrl` (`/releases/latest`, unchanged) and `releaseTagUrl` (`/releases/tags/v<version>`, one named tag). `resolveLine` at `:144` takes two booleans and is pure — there is no listing call to make. D2 satisfied structurally.

### Boundary

Clean. The diff touches `semver.ts`, `actana-release-channel.ts`, `actana-release.ts`, `actana-update.ts`, `core-update-notice.ts` and their tests, plus the new root-suite test. No `install.sh` (#316/#317), no `.github/workflows` (#325), and none of `actana-install.ts` / `actana-setup.ts` / `actana-layout.ts` / `actana-tree.ts` (#316).

### CI

Head `0b44fa9`: all 20 checks green, 3 conditional jobs skipped by design (train image jobs, multi-arch manifest publishes). `Unit Tests` ran 320 files green.

---

## Non-blocking notes

None of these gate the merge; recording them so they are not rediscovered.

1. **`packages/cli/src/actana-update.ts:117`** — the guard returns before the broken-tree repair at `:133`, whose comment says re-running update is the obvious repair for a deleted `versions/<v>`. For a beta machine `version === config.version` can never hold (`/releases/latest` excludes prereleases), so a beta install with a broken tree now gets the refusal instead of a reinstall, and the only command offered is `--version 0.4.0` — a downgrade. Worth knowing that the pre-PR behaviour here was *repair-by-downgrade*, i.e. the bug this ticket exists to stop, so this is not a regression and the criterion as written is met. It also needs #318 to be reachable at all. If you want to close it later, `refusedDowngrade` could name `--version <config.version>` when `missingTreeFile(installDirFor(layout, config.version))` is non-null.

2. **`packages/shared/src/actana-release-channel.ts:105`** — `lineOf` is a second implementation of `versionCore` (`semver.ts:33`) in the same package, differing only in the `.trim()`. Delegating would remove the duplication and the one visible consequence (a version string with leading whitespace renders as `a prerelease of the  0.4.1 line`).

3. **`packages/shared/src/actana-release-channel.ts:100`** — `isBetaVersion` is documented as "`0.4.1-beta`, and nothing else" but tests only `endsWith("-beta")` plus a length guard, so `isBetaVersion("garbage-beta")` is `true`. Nothing calls it yet; worth tightening before #317 pairs it with `lineOf`.

4. `resolveLine`, `BETA_SUFFIX`, `betaVersionForLine`, `isBetaVersion` and `releaseTagUrl` have no production caller yet. That is ADR 0036 D6 working as written — **#322 creates; #317 consumes** — not dead code, and it is unit-tested in place.

5. For the record: the ADR 0036 citations throughout the diff do resolve. The record is on `feat/release-pipeline` and absent from this base, exactly as the PR body states. It is `PROPOSED` and not yet ratified, which is worth remembering when it is cited as settled.

**Verdict: ready.** The fix is the smallest correct one the ticket asked for — mirror the implementation that was already right, and pin the two together with a test that fails when they drift.

Issue #323, opened 2026-08-24
Docs: the beta channel in the operator's voice — INSTALL.md, README, deploy/README and the ci-cd tag ladder


mrfqcentic:
Part of #314 — a beta train must be installable the same way a release is. Read that issue
first. The record is #315, and this ticket is the operator-facing half of it.

## The problem

Every document that describes how Actana is installed or released will be wrong the day this
milestone lands, and several of them are wrong in the specific way that costs an operator an
afternoon: they state a *decision* that has been reversed, in the product's own voice.

The load-bearing one:

> - **`beta-x.y.z` is not a semver prerelease, and that is deliberate** (D8). … this is safe
>   **only because betas never become GitHub Releases** (D9) — nothing that parses versions
>   ever sees one. …
>   The metal install path therefore has no beta channel: a beta is testable as a container.

`docs/ci-cd.md:325-333`. That paragraph is the user-facing restatement of the two clauses
#315 amends. **It is not in the source spec's list of documents to change**, and it is the
one that will be quoted back at whoever ships this.

The second: `INSTALL.md` §*Install in one command* prints a one-liner that installs *and*
activates. After #316 that is two commands.

## The decision already taken

Documentation is last in this milestone on purpose: it documents what shipped, not what was
planned. Nothing here is a new decision — every claim it makes is #315's, restated for the
audience the file has.

## The file surface

| File | What changes |
|---|---|
| `INSTALL.md` | §*Install in one command* becomes install-then-setup, on both platforms (`:27-50`, `:200-218`). §*Install by hand* Step 2 (`:122-158`) now differs from the one-liner only in who downloads the tarball, which is worth saying. A new section: installing a beta, with the URL #317 settled and the warning that `actana update` does not follow a beta line (#322) |
| `README.md` | the one-liner in the shop window, and a line placing the beta channel |
| `deploy/README.md` | §*Choosing a version* (`:153-194`) explains `ACTANA_TAG=beta-0.2.0` as "the open train". There are now two beta-ish image tags with different clocks — `beta-x.y.z` moves per train merge, `x.y.z-beta` moves per beta cut — and the table has to distinguish them |
[...]

PR #331, opened 2026-08-24
refactor(install-sh): install.sh installs and stops activating — setup is the operator's next command

### LINKED ISSUE #316: install.sh installs and stops activating: the Core bundle and the launcher are placed, setup is the next command

qcentic-dev:
Closes #316. **Base is `feat/release-pipeline`** — the feature branch for #314 — not `main`
and not `beta/0.4.1`.

## What changed

`install.sh` ended by running `actana setup`, so one line both **installed** a Core bundle
and **activated** the machine as a Core: mTLS material, a unit or LaunchAgent, lingering, the
daemon, registration in this machine's own registry. ADR 0036 **C2** records
install-is-not-activation as a fixed operator constraint. This is that tail coming out.

The script now resolves, downloads, verifies against `SHA256SUMS`, extracts, **places the
bundle and links the launcher**, prints the `actana setup` line to run next, and exits 0.
There is no flag for it — it is what the script does.

## The part that is more than deleting eleven lines

The extracted tree is ephemeral: `install.sh` unpacks into `mktemp -d` and deletes it on
EXIT. It survived only because `runActanaSetup` copied it into place. Remove the `setup` call
and nothing is installed at all.

So placement moved ahead of activation, and it did **not** move into POSIX sh.
`runActanaSetup` is split into `planCorePlacement` + `placeCoreBundle`, `actana place` is the
verb that stops there, and setup calls the same two functions before it activates anything.
One implementation entered at two points — the arrangement `actana-install.ts` already uses
for the download half. `resolveActanaLayout` stays the one place `ACTANA_HOME`,
`ACTANA_CONFIG_DIR`, `ACTANA_DATA_DIR`, `ACTANA_BIN_DIR` and the two XDG variables are
resolved.

The plan and the write are separate calls because setup does two things in the gap: it
removes a pre-rename unit and stops a running daemon, and neither belongs to placement.

## Forwarded flags: refused, not echoed

The ticket left this open. **The script stops accepting them.** Every flag it forwarded
belonged to `actana setup`, and setup is a separate command now — accepting `--public-host`
and doing nothing with it would silently drop the one setting that decides whether a Panel
can reach the machine. The six that `usage()` documented as passthrough are named one by one
so an operator pasting an old command line is told where their flag went, and anything else
gets an "unknown option" that points at the same place.

The e2e passes `--public-host` to its own `setup` step, which is what the ticket describes.

## The next command, and who prints it

`setupCommandFor` picks it: bare `actana setup` only when this install's own launcher answers
to that name **and** its directory is on `PATH`; a path through `current` otherwise. Both
failing cases are real — a fresh machine has no `~/.local/bin` at login, and an `actana` from
npm has no bundle around it, so `setup` there would download a release rather than activate
the one just placed.

The CLI prints it, because the CLI is the half that knows the layout. The script deliberately
prints no second copy, and a test asserts the word appears exactly once in a run's output.

## Landmine three: tested, not reasoned about

`deploy/core.Dockerfile` sets `NPM_CONFIG_PREFIX=/home/core/.local`, so an
`npm i -g @actana/cli` shim lands at exactly the path `resolveActanaLayout` calls `binLink`.
Placement reaches that path before setup does now. `claimLauncher` is on the placement path,
and **the committed coverage is two suites**, both against real files rather than a stubbed
`exists`: `actana-setup.test.ts` → *placeCoreBundle — the launcher path still has one owner*,
and `actana-machine-cli.test.ts` → *place* → *does not clobber an `actana` somebody else
installed at the same path*. Container-side, `place` joins `CORE_REFUSED_VERBS`, which
`panel-image.test.mjs` binds to `DOCKER_EQUIVALENT` and `smoke-core-image.mjs` runs against
the real image in CI. **The CLI-only path is not broken.**

The same collision was also walked through by hand against the real script and a real built
CLI — see *Verification* — but that was out-of-band checking, not a level of coverage. The
two suites above are what CI enforces.

## Not in this PR

- **No stamping.** ADR 0036 D1/D2 is #317. Nothing here makes it harder: the header, `usage()`
  and the resolution block are all still where #317 needs them, and `resolve_version` is
  untouched.
- Nothing under `.github/workflows` (#325), nothing in `packages/shared/src/semver.ts`,
  `packages/cli/src/actana-update.ts` or `packages/core/src/core-update-notice.ts` (#322), no
  `packages/shared/src/actana-release-channel.ts` (ADR 0036 D6 → #322), no tarball scripts
  under `scripts/lib`.
- The wider docs sweep is #323. `INSTALL.md` is corrected so it is not actively wrong, and
  README's front-page one-liner with it.

## Review round 2

Four blocking findings, all fixed; findings 5–10 fixed too, 11 fixed.

- **B1** `INSTALL.md`'s macOS section said the installer *"hands over to `actana setup`"* and
  loads a LaunchAgent, over a bare one-liner code block. Rewritten to two commands, with the
  absolute-path case named — a Mac that has never had `~/.local/bin` on `PATH` is the ordinary
  case there.
- **B2** `CORE_UPDATE_COMMAND` → **`actana update`**, not the two-step form. It is the only
  candidate that lands the new tree *and* restarts the daemon onto it, which is what "needs
  update" has to clear; `install.sh` + `actana setup` would work but is the wrong gesture —
  `setup` activates a machine, and this one is already active — and it drags a full install
  through a machine that already has the bundle. The constant's docstring said the verb did
  not exist yet; it does now. Every existing test read the constant rather than a literal, so
  none of them could tell its value had stopped working — one now binds the value.
- **B3** `docs/core-linux-rehearsal.md` §2 split into install and activate, with every prompt
  box moved to the second step and a new box for "the printed line runs". The page was
  unpassable as written, and its own unticked boxes are release blockers.
- **B4** Below.
- **N5** `place` now stops a daemon whose own `installDir` it is about to replace, and says so.
  Only when there is a tree at that path to destroy — a fresh machine asks the init system
  nothing — and best-effort, because placement must not fail for a reason that belongs to
  activation. Tested on both systemd and launchd.
- **N6** When an activated install is already present, the summary names both versions, says
  the daemon is still on the old one, and calls the printed command finishing the upgrade.
- **N7** The `Launcher` row is gated on `outcome === "linked"`; otherwise it names the program
  that owns the name.
- **N8** `installTree` moves the old tree aside to `<installDir>.previous` and puts it back if
  the rename fails, instead of deleting it first. Both docstrings are true now rather than
  aspirational.
- **N9** The guard test fails during the copy (`chmod 000`) instead of at `missingTreeFile`, so
  it enters the branch it is named for.
- **N10** `fixture-release-server.mjs` prints both commands.
- **N11** `INSTALL.md:5` rewrapped.

Nothing deferred. Boundaries unchanged: no `.github/workflows`, no `semver.ts`,
`actana-update.ts`, `actana-release.ts` or `core-update-notice.ts`, no
`packages/shared/src/actana-release-channel.ts`.

## Verification

- `packages/cli`: 1029 passed, 3 skipped. Root script suite: 567 passed. `tsc --noEmit` clean
  for cli/core/shared/sdk; `eslint` clean.
- **The real script against the real CLI**, out of band: a bundle built from
  `dist-tarball/actana-cli.cjs`, served by the fixture release server, installed by the real
  one-liner. Verified end to end — exit 0; `versions/<v>`, `current` and the launcher all
  present; nothing under `~/.config`; no temp files left; every `ACTANA_*` and XDG override
  honoured (including "a relative `ACTANA_*` resolves against `$HOME`, a relative XDG value is
  ignored"); the npm-shim collision refused without clobbering; the printed command absolute
  when the bin dir is off `PATH`, bare when it is on, and runnable in both cases.
- `scripts/e2e-actana-setup-linux.mjs` **was not run**: this machine has no Docker and no
  systemd. Its one new piece of logic — reading the printed command back out — is unit-tested
  in `scripts/__tests__/setup-e2e.test.mjs`, CRLF included, because `machinectl shell` pipes
  through a PTY.

https://claude.ai/code/session_01811GQ3GsC6v7mCTkag9a7m

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,9 +1,10 @@
 # Core Installation Guide
 
 This guide covers turning a **Linux or macOS** machine into a **Core**:
-installing the Core bundle, running `actana setup` — a separate command,
-because installing is not activating — and pairing the machine with your Panel. The Core is the stateful daemon that runs harnesses and owns
-the PTY layer; the Panel is the web app you drive it from.
+installing the Core bundle, running `actana setup` — a separate command, because
+installing is not activating — and pairing the machine with your Panel. The Core
+is the stateful daemon that runs harnesses and owns the PTY layer; the Panel is
+the web app you drive it from.
 
 **There is one `actana`.** The command that installs and operates a Core is the
 same command that drives Cores — `actana core ls`, `actana session start` and
@@ -303,23 +304,27 @@ actana core pair prod <public-host>:<port> XXXX-XXXX \
 
 ## macOS, on the machine itself
 
-A Mac with Apple silicon is a Core like any other: the one-liner at the top of
-this page installs it, `actana setup` runs without sudo, and the machine pairs
-with your Panel the same way. What differs is the auto-start mechanism, and
-one property that follows from it.
+A Mac with Apple silicon is a Core like any other: the same two commands at the
+top of this page install it, both run without sudo, and the machine pairs with
+your Panel the same way. What differs is the auto-start mechanism, and one
+property that follows from it.
 
 ```bash
 curl -fsSL https://raw.githubusercontent.com/actana/control/main/install.sh | bash
+actana setup
 ```
 
-The installer maps the machine to the `mac-arm64` release asset and hands over
-to `actana setup`, which writes a **LaunchAgent** at
+The first command maps the machine to the `mac-arm64` release asset, places the
+bundle and links the launcher — and stops. Nothing is running yet. Run the
+`actana setup` line it printed; that is what writes the **LaunchAgent** at
 `~/Library/LaunchAgents/com.actana.core.plist`, labelled `com.actana.core`, and
-loads it.
+loads it. On a Mac that has never had `~/.local/bin` on its `PATH`, the printed
+line is an absolute path rather than a bare `actana` — run it as printed.
 
 **An Intel Mac stops here, deliberately.** There is no `mac-x64` build and
-there will not be one; the installer refuses at detection and points you at the
-container image, which is the supported way to run a Core on that machine.
+there will not be one; the installer refuses at detection — before it downloads
+anything — and points you at the container image, which is the supported way to
+run a Core on that machine.
 
 ### The LaunchAgent is tied to your login session
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,8 +1,8 @@
 # Core Installation Guide
 
 This guide covers turning a **Linux or macOS** machine into a **Core**:
-downloading the Core bundle, running `actana setup`, and pairing the machine
-with your Panel. The Core is the stateful daemon that runs harnesses and owns
+installing the Core bundle, running `actana setup` — a separate command,
+because installing is not activating — and pairing the machine with your Panel. The Core is the stateful daemon that runs harnesses and owns
 the PTY layer; the Panel is the web app you drive it from.
 
 **There is one `actana`.** The command that installs and operates a Core is the
@@ -36,12 +36,28 @@ automatic login, or leave the session open.
 
 ## Install in one command
 
+Two commands, and the first one prints the second:
+
 ```bash
 curl -fsSL https://raw.githubusercontent.com/actana/control/main/install.sh | bash
+actana setup
 ```
 
-**Or, if you already have the CLI**, the same three steps without the shell
-script:
+**Installing is not activating.** `install.sh` puts the Core bundle and the
+`actana` CLI on this machine and stops. `actana setup` is what turns the machine
+into a Core: it generates the mTLS material, writes the auto-start service,
+offers to enable lingering, offers the harness CLIs, starts the daemon, and
+registers the Core with this machine's own `actana`. They are two different acts
+with two different audiences, and there is no flag that fuses them back into
+one.
+
+The installer prints the exact `actana setup` line to run — **run the line it
+printed**. On a machine where `~/.local/bin` was not yet on your `PATH` when the
+shell started, that line is an absolute path rather than a bare `actana`, and it
+is the one that will actually be found.
+
+**Or, if you already have the CLI**, `actana install` does both halves in one
+go — it is the CLI's own front door, not the shell script's:
 
 ```bash
 npm i -g @actana/cli
@@ -50,15 +66,16 @@ actana install
 
 `install.sh` stays the door for a machine with no Node — the tarball carries its
 own pinned runtime, so the script cannot be replaced by the CLI it installs.
-Both do the same work: resolve the release, download it, check it against the
-release's `SHA256SUMS`, extract it, then set up. A failed check leaves nothing
-installed either way.
+Both fetch the same way: resolve the release, download it, check it against the
+release's `SHA256SUMS`, extract it. A failed check leaves nothing installed
+either way.
 
-That detects the machine's OS and CPU, downloads the matching Core tarball
-from the latest GitHub Release, **checks it against the release's `SHA256SUMS`
-before extracting or running anything**, and hands over to `actana setup` —
-which installs, writes the auto-start unit, starts the daemon, and tells you how
-to pair a client with it.
+The one-liner detects the machine's OS and CPU, downloads the matching Core
+tarball from the latest GitHub Release, **checks it against the release's
+`SHA256SUMS` before extracting or running anything**, copies the bundle to
+`~/.local/share/actana/versions/<version>`, points `~/.local/share/actana/current`
+at it, and links `~/.local/bin/actana` — unless something else already answers
+to `actana`, in which case it leaves that one alone and says so.
 
 The checksum catches a corrupted or truncated download: it proves the tarball
 is the one that release's own checksum file describes. Releases are not signed
@@ -66,12 +83,11 @@ is the one that release's own checksum file describes. Releases are not signed
 release workflow), so it is not a proof of origin — use `https` URLs, which the
 defaults do.
 
-Piped like that, the run is non-interactive: nothing prompts, and every choice
-comes from a flag. Flags the installer does not own are passed straight through
-to `actana setup`:
+Piped like that, the run is non-interactive — it asks nothing, because it
+decides nothing. The four options it takes are its own:
 
 ```bash
-curl -fsSL <install-script-url> | bash -s -- --version 0.1.0 --public-host core1.example.com --yes
+curl -fsSL <install-script-url> | bash -s -- --version 0.1.0
 ```
 
 | Flag | Meaning |
@@ -79,18 +95,35 @@ curl -fsSL <install-script-url> | bash -s -- --version 0.1.0 --public-host core1
 | `--version <v>` | Install this exact release instead of the latest |
 | `--repo <slug>` | Install from another GitHub repository |
 | `--base-url <url>` | Fetch releases from somewhere else — how the tests run hermetically |
-| *anything else* | Passed to `actana setup` (see the table below) |
+| `--help` | Show what the script does and the options it owns |
+
+**Anything else is refused, not ignored.** `--yes`, `--port`, `--public-host`,
+`--label`, `--with-<harness>` and `--no-harnesses` are `actana setup`'s options
+and belong on the second command — an installer that accepted `--public-host`
+and then did nothing with it would silently drop the one setting that decides
+whether your Panel can reach this machine.
+
+```bash
+curl -fsSL <install-script-url> | bash
+~/.local/bin/actana setup --public-host core1.example.com --yes
+```
 
 `ACTANA_VERSION`, `ACTANA_REPO` and `ACTANA_BASE_URL` set the same three
-options, for provisioning systems where flags are awkward.
+installer options, for provisioning systems where flags are awkward.
+`ACTANA_HOME`, `ACTANA_BIN_DIR`, `ACTANA_CONFIG_DIR`, `ACTANA_DATA_DIR` and the
+XDG variables move where things land; the script does not read them itself — the
+CLI does — so the installer and `actana setup` cannot disagree about where
+anything is.
 
-**Re-running the one-liner on a machine that already has a Core upgrades it
+**Re-running both commands on a machine that already has a Core upgrades it
 in place** — same install, same identity, one unit, and every paired client
 stays paired. It is always safe to paste again.
 
 If anything fails — an unsupported platform, a release without a build for it,
 a checksum that does not match — the installer stops before extracting or
-running a single byte of the download, and says what to do about it.
+running a single byte of the download, and says what to do about it. A failed
+run leaves the machine as it found it: nothing under the install root, no
+launcher, no half-placed tree.
 
 The rest of this page is the same install done by hand, and how to operate a
 Core afterwards.
@@ -143,25 +176,45 @@ shasum -a 256 --ignore-missing -c SHA256SUMS
 
 That must print `OK`. If it does not, stop — do not extract or run anything.
 
-### Step 2 — Extract and run setup
+### Step 2 — Extract, place the bundle, and run setup
 
 ```bash
 tar -xzf actana-core-0.1.0-linux-x64.tar.gz
 ```
 
 ```bash
-./actana-core-0.1.0-linux-x64/bin/actana setup
+./actana-core-0.1.0-linux-x64/bin/actana place
 ```
 
-`setup` does all of it:
+```bash
+actana setup
+```
+
+**These are the same two commands the one-liner runs** — the only difference is
+who downloaded the tarball. `place` copies the extracted tree into the install
+layout and links the launcher; `setup` activates the machine. Run the
+`actana setup` line `place` printed, for the same reason as above: on a machine
+where `~/.local/bin` is not on your `PATH` it is an absolute path.
+
+`place` does the install half:
 
 - copies the tree to `~/.local/share/actana/versions/<version>` and points
   `~/.local/share/actana/current` at it,
-- links the launcher into `~/.local/bin/actana` (and tells you if that
-  directory is not on your `PATH`) — **unless something else already answers to
-  `actana` there or earlier on your `PATH`**, in which case it leaves that one
-  alone and says so. Whoever installed a CLI owns its path; since there is one
-  `actana` program, the one already there runs this Core's verbs too,
+- links the launcher into `~/.local/bin/actana` — **unless something else
+  already answers to `actana` there or earlier on your `PATH`**, in which case
+  it leaves that one alone and says so. Whoever installed a CLI owns its path;
+  since there is one `actana` program, the one already there runs this Core's
+  verbs too,
+- prints the `actana setup` command to run next, and stops. Nothing is running
+  and nothing is configured until you run it.
+
+You can skip `place` and run `setup` straight out of the extracted directory —
+`setup` places the tree it is standing in before it activates anything, which
+is what it has always done. `place` exists so that installing and activating
+can be two separate decisions.
+
+`setup` does the rest:
+
 - generates the mTLS material and persists it to `~/.config/actana/material.json`,
 - writes the auto-start service — the systemd user unit
   `~/.config/systemd/user/actana-core.service` on Linux, or the LaunchAgent
@@ -451,21 +504,34 @@ provisioning script.
 
 ### `actana` is not found in a new shell
 
-`setup` links the launcher into `~/.local/bin`, which some distributions leave
-off `PATH`. Add it:
+The install links the launcher into `~/.local/bin`, which some distributions
+leave off `PATH` — and which a shell that started before the install will not
+have picked up in any case. Add it:
 
 ```bash
 export PATH="$HOME/.local/bin:$PATH"
 ```
 
-### `setup` said it left the launcher alone
+This is why the installer prints an absolute path rather than a bare `actana`
+when it can see that `~/.local/bin` is not on your `PATH`. Running the line it
+printed always works, `PATH` or no `PATH`.
+
+### The install finished and nothing is running
+
+That is the install working. `install.sh` places the bundle and stops;
+`actana setup` is the command that makes this machine a Core. Run the line the
+installer printed — `actana status` before that says there is no Core
+installed for this user, which is true until setup has run.
+
+### It said it left the launcher alone
 
 Something else already answers to `actana` — usually `npm i -g @actana/cli`,
 whose global shim lands in the same directory. Nothing is broken: it is the same
-program, so it runs this Core's `status`, `logs` and `update` as well. Setup
-prints where this install's own launcher is if you would rather run that one
-directly. To hand the path over, remove the other install
-(`npm rm -g @actana/cli`) and re-run `actana setup`.
+program, so it runs this Core's `status`, `logs` and `update` as well. The
+message names where this install's own launcher is if you would rather run that
+one directly, and the `actana setup` command it prints goes through that
+launcher rather than through the other program. To hand the path over, remove
+the other install (`npm rm -g @actana/cli`) and re-run the install.
 
 ### The service started but nothing is listening
 
@@ -522,11 +588,11 @@ actana setup --public-host <new-address>
 | `~/.local/share/actana/current` | Symlink to the version the unit runs. |
 | `~/.local/share/actana/data/missioncontrol.db` | SQLite: projects, tasks, sessions, event log. Never touched by setup. |
 | `~/.config/actana/material.json` | CA, server cert/key, client cert/key, bearer secret, coreId. `chmod 0600`. |
-| `~/.config/actana/actana.json` | What setup decided: version, port, host, public host, label. No secrets. |
+| `~/.config/actana/actana.json` | What setup decided: version, port, host, public host, label. No secrets. Absent until setup runs. |
 | `~/.config/systemd/user/actana-core.service` | Linux: the auto-start user unit. |
 | `~/Library/LaunchAgents/com.actana.core.plist` | macOS: the auto-start LaunchAgent. |
 | `~/Library/Logs/Actana/core.log` | macOS: the daemon's output — what `actana logs` tails. |
-| `~/.local/bin/actana` | Symlink to `current/bin/actana`. |
+| `~/.local/bin/actana` | Symlink to `current/bin/actana`, written by the install unless somebody else owns the name. |
 
 The two service paths are fixed — systemd and launchd each read exactly one
 directory. Everything else honours `XDG_DATA_HOME` and `XDG_CONFIG_HOME`
--- a/README.md
+++ b/README.md
@@ -91,10 +91,14 @@ against the one on screen. [deploy/README.md](deploy/README.md) walks through
 the file itself — the volumes, the loopback port, adding a second Core.
 
 **Installer** — turn a Linux or macOS (arm64) machine into a Core, as your own
-user, without sudo:
+user, without sudo. Two commands, because installing is not activating: the
+first puts the Core bundle and the `actana` CLI on the machine, the second
+turns the machine into a Core. The first prints the second — run the line it
+printed.
 
 ```bash
 curl -fsSL https://raw.githubusercontent.com/actana/control/main/install.sh | bash
+actana setup
 ```
 
 **Which one?** Compose if you want the whole thing on one host to try it;
--- a/docs/core-linux-rehearsal.md
+++ b/docs/core-linux-rehearsal.md
@@ -1,18 +1,28 @@
-# Linux Core — the one-liner rehearsal
+# Linux Core — the install rehearsal
 
 CI installs a Core on Linux several times per pull request: `.github/workflows/ci.yml`'s
 `installer-e2e` job runs `scripts/e2e-actana-setup-linux.mjs` — the real
-one-liner, then the lifecycle verbs on the machine it produced — across Ubuntu
-and Debian at x64, and the `core-image` job pairs a Panel with the containerised
-Core. arm64 runs the same script on the release tag
-(`.github/workflows/release.yml`).
+one-liner, then `actana setup` as its own step, then the lifecycle verbs on the
+machine those two produced — across Ubuntu and Debian at x64, and the
+`core-image` job pairs a Panel with the containerised Core. arm64 runs the same
+script on the release tag (`.github/workflows/release.yml`).
 
 All of it runs with the prompts suppressed and nobody watching. That is the
 gap this fills: **once per release, a person pastes the real one-liner into a
-machine that has never seen it, answers the questions it asks, and takes a
-pairing code to a live Panel.** What it catches is the class of problem a green
-matrix cannot — a prompt that reads badly, a code that is awkward to read out
-loud, an instruction that is technically true and practically useless.
+machine that has never seen it, runs the second command it prints, answers the
+questions *that* asks, and takes a pairing code to a live Panel.** What it
+catches is the class of problem a green matrix cannot — a prompt that reads
+badly, a code that is awkward to read out loud, an instruction that is
+technically true and practically useless.
+
+**Two commands, because installing is not activating** (ADR 0036 C2, #316).
+`install.sh` places the Core bundle and the `actana` CLI and stops; `actana
+setup` is what turns the machine into a Core, and it is where every prompt this
+rehearsal exists to exercise now lives. The installer prints the exact `setup`
+line to run — an absolute path when `~/.local/bin` is not yet on `PATH` — and
+**that printed line is the one to paste**, not a retyped `actana setup`. A
+printed command that cannot be run is itself a release blocker, and this is the
+only place a person checks it.
 
 Fifteen minutes, and the only install rehearsal a release needs — Linux is the
 only platform a release publishes a Core for. [The macOS
@@ -57,10 +67,12 @@ pnpm core:rehearse
 ```
 
 That one command builds a systemd container with **no sudo on it**, starts a
-local release channel serving the tarball you just built, prints the one-liner
-to paste, and drops you into a shell inside the machine as `operator`.
+local release channel serving the tarball you just built, prints the two
+commands to paste, and drops you into a shell inside the machine as `operator`.
 
-- [ ] The banner prints a `curl -fsSL … | bash -s -- --base-url …` command.
+- [ ] The banner prints a `curl -fsSL … | bash -s -- --base-url …` command, and
+      an `actana setup` to follow it, and says that the first installs without
+      activating.
 - [ ] The prompt that follows is inside the machine (`whoami` says `operator`,
       `sudo` is not found).
 
@@ -78,8 +90,26 @@ re-enter it and how to destroy it.
 
 ## 2 — Install, the way an operator does
 
-Paste the printed one-liner. **Do not add `--yes`** — the prompts are the point.
+Paste the printed one-liner. `--base-url`, `--repo` and `--version` are the only
+options it still takes — `--yes` and the rest are `actana setup`'s now, and the
+script refuses them by name rather than ignoring them.
 
+- [ ] It downloads, says the checksum verified, and exits **without asking you
+      anything**. Nothing prompts, because nothing here decides anything.
+- [ ] It says plainly that nothing is running yet.
+- [ ] It prints one `actana setup` command, and only one.
+- [ ] `actana status` at this point says there is **no Core installed for this
+      user** — which is true, and is the machine being installed but not
+      activated. (If `actana` is not found yet, that is the case the printed
+      absolute path exists for; the next box checks it.)
+
+### Then the second command — this is the one with the prompts in it
+
+Paste **the line the installer printed**, exactly as printed. Do not retype it
+as a bare `actana setup`, and **do not add `--yes`** — the prompts are the point.
+
+- [ ] The printed line runs. If it was an absolute path, it worked as pasted;
+      a printed command that is not found is a release blocker on its own.
 - [ ] It asks about enabling lingering, and explains *why* before asking.
 - [ ] It offers to install the harness CLIs it found missing, one at a time.
       (Decline them all — this machine has no vendor logins on it, and CI
@@ -202,6 +232,7 @@ If 8443 is taken, `--port <n>` moves it — and then you pair against
 
 Note in the release PR: the distro you used, and which boxes did not tick.
 
-Unticked boxes in sections 2 and 3 are release blockers — "the one-liner ends
-with a Core a person can actually pair to a Panel" is the whole product promise
-this rehearsal exists to protect. Everything else is a bug report.
+Unticked boxes in sections 2 and 3 are release blockers — "the two commands end
+with a Core a person can actually pair to a Panel, and the first one prints a
+second one that runs" is the whole product promise this rehearsal exists to
+protect. Everything else is a bug report.
--- a/docs/core-macos-prerelease-checklist.md
+++ b/docs/core-macos-prerelease-checklist.md
@@ -44,9 +44,9 @@ Mac runs its Core from the container image, and both `install.sh` and
 `actana update` refuse it at detection and say so.
 
 The Linux path is covered by
-[the one-liner rehearsal](core-linux-rehearsal.md): `pnpm core:rehearse` for a
-throwaway machine to run the real one-liner on. Same standing as this page
-now has, for the other half of the release.
+[the install rehearsal](core-linux-rehearsal.md): `pnpm core:rehearse` for a
+throwaway machine to run the real one-liner and the `actana setup` it prints on.
+Same standing as this page now has, for the other half of the release.
 
 ---
 
--- a/install.sh
+++ b/install.sh
@@ -1,25 +1,39 @@
 #!/bin/sh
-# actana — turn this machine into a Core.
+# actana — install the Core bundle and the CLI on this machine.
 #
 #   curl -fsSL https://raw.githubusercontent.com/actana/control/main/install.sh | bash
+#   actana setup
 #
-# This script is a bootstrapper and nothing more: detect the platform, resolve
-# the release, download the matching tarball, verify it against the release's
-# published SHA256SUMS, extract it, and hand over to `actana setup`. Every real
-# decision — where to install, the systemd unit, lingering, Harnesses, wiring
-# the Core into this machine's own registry — lives in that CLI, where it is
-# unit-tested. Anything this
-# script grows beyond "fetch, verify, exec" belongs there instead.
+# **Two commands, because installing is not activating** (ADR 0036 C2). This
+# script detects the platform, resolves the release, downloads the matching
+# tarball, verifies it against the release's published SHA256SUMS, extracts it,
+# places the bundle, links the launcher, prints the `actana setup` line to run
+# next, and exits. It does not turn this machine into a Core: the mTLS
+# material, the systemd unit or LaunchAgent, lingering, the Harness offers, the
+# daemon and this machine's own registration are all `actana setup`, which the
+# operator runs afterwards. There is no flag for this — it is what the script
+# does.
 #
-# Flags it does not own are forwarded verbatim to `actana setup`, so
-# `… | bash -s -- --yes --public-host 10.0.0.7` works and stays working as the
-# CLI gains flags.
+# **It does not know the install layout, and must not learn it.** Placement is
+# `"$extracted/bin/actana" place` — the CLI in the tree that was just verified,
+# running on the Node pinned inside it. ACTANA_HOME, ACTANA_CONFIG_DIR,
+# ACTANA_DATA_DIR, ACTANA_BIN_DIR, XDG_DATA_HOME and XDG_CONFIG_HOME resolve
+# there, in `packages/cli/src/actana-layout.ts`, where they are unit-tested
+# against a fake home. A second, subtly different copy of those rules in POSIX
+# sh is exactly the failure this repository refuses for release resolution.
+# So: "fetch, verify, place" — and anything this script grows beyond that
+# belongs in the CLI instead.
+#
+# Flags this script does not own are refused rather than forwarded. Every one
+# it used to forward belonged to `actana setup`, and `actana setup` is a
+# separate command now — passing `--yes` to an installer that prompts for
+# nothing would be a flag that quietly did nothing.
 #
 # POSIX sh: runs under bash, dash and ash. Tested by
 # `scripts/__tests__/install-sh.test.mjs` (the real script against a fixture
 # release server) and by `scripts/e2e-actana-setup-linux.mjs`, which enters at
-# the real one-liner on a real systemd machine and carries on into the
-# lifecycle verbs.
+# the real one-liner on a real systemd machine, runs `actana setup` as its own
+# next step, and carries on into the lifecycle verbs.
 
 set -eu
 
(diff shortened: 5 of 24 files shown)


qcentic-dev:
## Acceptance criteria, #316

Every criterion from the issue, in its order, each checked or explicitly unchecked with the
reason.

- [x] **`install.sh` no longer runs `actana setup`, on any path, with or without a terminal.**
      The tail is gone. There is one call and it is `"$extracted/bin/actana" place </dev/null`
      — no `[ -t 0 ]` branch, because nothing the script runs prompts.
      `install-sh.test.mjs` → *runs `actana place`, and never `actana setup`, with %j* (both
      the default and `--version` paths), *runs `actana place` with no terminal, the same as
      when piped* (the shape that used to take the other branch), and *is servable and
      runnable straight off the release server* asserts the stub was called with exactly
      `place`.

- [x] **The bundle survives the script: `versions/<version>/` exists, `current` points at it,
      the launcher is linked — with the same paths `resolveActanaLayout` would compute,
      including every `ACTANA_*` and XDG override.**
      Placement goes through `placeCoreBundle`, which takes an `ActanaLayout` from
      `resolveActanaLayout`, so there is nothing to keep in sync.
      `install-sh.test.mjs` → *leaves something behind: the placement outlives the temp dir*.
      `actana-setup.test.ts` → *lands the tree under versions/&lt;v&gt;, points current at it,
      links the launcher*. `actana-machine-cli.test.ts` → *places the bundle, points `current`
      at it, and links the launcher*. And out of band against the real script and the real
      built CLI: `ACTANA_HOME`, `ACTANA_BIN_DIR`, `ACTANA_CONFIG_DIR` and `XDG_DATA_HOME` all
      honoured, a relative `ACTANA_HOME` resolved against `$HOME`, a relative `XDG_DATA_HOME`
      ignored, and nothing written under `~/.local` when `ACTANA_HOME` moved the root.

- [x] **Layout resolution is not duplicated in POSIX sh; placement goes through the CLI in the
      extracted tree.**
      The script gained no path knowledge at all: no `ACTANA_*` read, no `XDG_*` read, no
      `versions/`, no `current`, no `.local/bin`. The route is the new `actana place` verb on
      the binary inside the verified tree, running on its own pinned Node. The header says so
      and says why.

- [x] **The script prints the exact next command, and it is runnable as printed. When the
      launcher's directory is not on `PATH`, the printed command is the absolute path.**
      `setupCommandFor` decides: bare `actana setup` only when this install's own launcher
      answers to that name *and* its directory is on `PATH`; otherwise a path through
      `current`. It also goes absolute when the launcher is somebody else's — an `actana` from
      npm has no bundle around it, so `setup` there would download a release rather than
      activate what was just placed.
      `actana-setup.test.ts` → the five `setupCommandFor` cases.
      `actana-machine-cli.test.ts` → *prints an absolute path when the launcher's directory is
      not on PATH*, *names a help command that actually answers*.
      Out of band, for real: with the bin dir off `PATH` the printed line is absolute and
      running it works; with it on `PATH` the line is bare and `actana` resolves to this
      install.

- [x] **A failed run still leaves the machine as it was found: nothing under the install root,
      no launcher, no half-placed tree. The checksum refusal keeps running before anything is
      written outside the temp dir.**
      The checksum block is untouched and still runs before `place` is reached.
      `installTree` now takes its own staging directory with it on a failed copy, and
      `placeCoreBundle` removes a version directory that this call created.
      `actana-setup.test.ts` → *leaves nothing behind when the install root cannot be
      written*, *leaves nothing behind when the copy fails part-way through*, *does not delete
      a version that was already installed when a re-place fails*, *does not adopt a staging
      directory a crashed run left behind*.
      `install-sh.test.mjs` → *refuses to extract or run a tarball whose checksum does not
      match* (unchanged) and *leaves no temporary files behind when the placement fails
      either*.

- [x] **Exit code is 0 on a successful install; the `setup` exit-code propagation goes with
      the call it propagated.**
      The `set +e` / `status=$?` / `exit $status` block is gone. `set -e` carries a failed
      placement out with its own status.
      `install-sh.test.mjs` → *comes back with the placement's own failure status* (7), and
      every success case asserts `status === 0`.

- [x] **`usage()` and the file header describe what the script now does — no passthrough flags
      it no longer forwards, and no "hand over to `actana setup`".**
      Both rewritten. `usage()` lists the four options the script owns, says the rest are
      refused rather than ignored, and names `actana setup` as a separate command.
      `install-sh.test.mjs` → *describes what it does now: two commands, and no passthrough*
      and *does not claim to hand over to `actana setup` in its own header*.
      `actana-install.test.ts` → the header's self-description moved from "fetch, verify,
      exec" to "fetch, verify, place", pinned in both directions.

- [x] **The whole `handing off to `actana setup`` describe block is replaced rather than
      deleted, at the same density of negative cases. `leaves no temporary files behind` still
      holds.**
      Replaced by *placing the bundle, and stopping there* — 4 tests became 13 cases: which
      verb runs (×2), no terminal on the file path, the placement outlives the temp dir, the
      next command printed exactly once, empty stdin, one case per refused `setup` flag (×8),
      an unowned flag, the failure status, and no temp files after either outcome.
      *leaves no temporary files behind* is kept verbatim, with a failure-path twin.

- [x] **`scripts/e2e-actana-setup-linux.mjs` runs `setup` as its own step, and every later
      assertion in that file is unchanged.**
      The one-liner is asserted to produce an installed, not activated machine — tree,
      `current`, launcher present; no unit active; no `actana-core.service`, `actana.json` or
      `material.json` written — and then the **printed** command is run as its own step, read
      back with `nextSetupCommand` rather than retyped, because "runnable as printed" is the
      criterion. Everything downstream is unchanged: the core-link dial with the wired
      credential, lingering, `status`, `pair new`, `token`'s refusal, start/stop/restart/logs,
      the pre-rename unit removal, the in-place upgrade, the reboot, the tampered and pinned
      updates, `token regenerate`, and both uninstalls. The two other one-liner runs became
      install-then-setup pairs for the same reason.
      **Not executed here** — see the unchecked item below.

- [x] **`scripts/rehearse-install.mjs` and `scripts/lib/setup-e2e.mjs` agree with the new
      shape.**
      `setup-e2e.mjs` gains `nextSetupCommand` and is shared machinery for these tests again
      rather than only in name — the Linux e2e imports it, and it has its own unit test in the
      new `scripts/__tests__/setup-e2e.test.mjs` (six cases, CRLF included, because
      `machinectl shell` pipes through a PTY). `rehearse-install.mjs` prints two commands and
      says which one has the prompts in it, since that is where they now live;
      `rehearsal.mjs` gains `rehearsalSetupCommand`, and `rehearsal.test.mjs` gains four cases
      including one binding the one-liner to the options `install.sh` still owns.

- [x] **`INSTALL.md` §*Install in one command* shows two commands; §*Install by hand* Step 2
      now differs from the one-liner only in who downloads the tarball.**
      §Install in one command shows both, names which act each one is, tells the reader to run
      the printed line, and replaces the passthrough table with the four options the script
      owns. Step 2 is `actana place` then `actana setup` — the same two commands — with a note
      that running `setup` straight out of an extracted directory still works. Troubleshooting
      gains *the install finished and nothing is running*. README's front-page one-liner is
      corrected the same way; the wider sweep stays #323.

## Not claimed

- [ ] **The Linux container e2e was not executed.** This machine has no Docker and no systemd,
      so `pnpm core:setup:e2e` could not run. Mitigations: `node --check` on the file; the one
      new piece of logic in it (`nextSetupCommand`) is unit-tested including the CRLF that
      `machinectl shell`'s PTY produces; and the install half was exercised for real out of
      band — the real `install.sh`, the real built CLI, a real tarball over the fixture
      release server, across layout overrides, the npm-shim collision and both `PATH` cases.
      CI runs the container e2e on this PR.

- [ ] **The stamped line version (ADR 0036 D1/D2) is not implemented.** That is #317 and out of
      scope here. Nothing in this change makes it harder: `resolve_version` is untouched, and
      the header, `usage()` and the resolution block are all still where #317 needs them.

## Boundaries respected

Nothing under `.github/workflows` (#325). No `packages/shared/src/semver.ts`,
`packages/cli/src/actana-update.ts` or `packages/core/src/core-update-notice.ts` (#322). No
`packages/shared/src/actana-release-channel.ts` created (ADR 0036 D6 → #322). No tarball
scripts under `scripts/lib` — `scripts/lib/panel-image.mjs` is touched only to add `place` to
`CORE_REFUSED_VERBS`, which `panel-image.test.mjs` requires as the deliberate handshake with
the CLI's own refusal table. Nothing merged, `main` untouched, PR left as a draft.

mrfqcentic (changes requested):
## Review of #331 against #316 — request changes

The core of this ticket is done, and done well. I read the enforcing lines rather than the
description, and the two claims that matter most both hold:

**The tail is genuinely gone.** `install.sh:393` is `"$extracted/bin/actana" place </dev/null`
— the only child invocation in the file. No `exec`, no `setup` under any name, `FORWARD` and
its `eval "set -- $FORWARD"` removed, and the `status=$?` / `exit $status` propagation gone
with the call it propagated. The dead `quote()` helper went too. The closing output is
`cmdPlace` at `packages/cli/src/actana-cli.ts:698-712`: install dir, launcher, `current`,
then *"Nothing is running yet: this machine is not a Core until you set it up."* and the one
runnable `actana setup` line, chosen by `setupCommandFor` (`actana-setup.ts:339`).

**The contract changed, not just the script.** This is the part I expected to find weak and
did not. All four named files assert the new behaviour rather than being edited to pass:

- `scripts/__tests__/install-sh.test.mjs` — the `handing off to \`actana setup\`` block is
  replaced by `placing the bundle, and stopping there`, and it is denser than what it
  replaced (5 cases → ~17). It asserts `actanaArgs === "place"` **and** `not.toMatch(/setup/)`,
  and critically it asserts that something *survived* (`leaves something behind: the placement
  outlives the temp dir`). Without that second half a script that placed nothing would pass a
  "did not run setup" suite while installing nothing. `leaves no temporary files behind` still
  holds, and gains a failure-path twin.
- `scripts/e2e-actana-setup-linux.mjs` — `installOnly()` asserts the four placement artefacts,
  then asserts the *negative* space: `systemctl --user is-active` is not `active`, and none of
  `actana-core.service`, `actana.json`, `material.json` exist. It then runs the printed command
  **as printed**, not a retyped `actana setup` — which is the only way #316's fourth criterion
  is actually tested. Every downstream assertion is unchanged.
- `scripts/lib/setup-e2e.mjs` — `nextSetupCommand` requires **exactly one** match; zero and two
  are both hard failures, and the CRLF case (`machinectl shell` PTY) is unit-tested.
- `scripts/rehearse-install.mjs` / `scripts/lib/rehearsal.mjs` — two commands in the banner, and
  `rehearsal.test.mjs` binds the one-liner's flags to the script's own option list.

Boundary is clean: nothing under `.github/workflows`, nothing in `semver.ts`,
`actana-update.ts`, `actana-release.ts`, `core-update-notice.ts`, and
`packages/shared/src/actana-release-channel.ts` was not created (ADR 0036 D6 → #322).
ADR 0036 C2 is honoured, and D1 is not blocked — `resolve_version` and the whole resolution
block are untouched, and the header/`usage()` #317 needs are still there.

The npm-global-bin collision is tested, not reasoned about. The proving tests are
`actana-setup.test.ts` → `placeCoreBundle — the launcher path still has one owner` →
`does not clobber an \`actana\` npm put at the very same path` (a real symlink planted at
`binLink`), and `actana-machine-cli.test.ts` → `place` → `does not clobber an \`actana\`
somebody else installed at the same path`. Container-side, `place` joins `CORE_REFUSED_VERBS`,
which `scripts/__tests__/panel-image.test.mjs:915` asserts equals `DOCKER_EQUIVALENT` and
`scripts/smoke-core-image.mjs:463` runs against the real image in CI.

One correction to the PR description: *"a real run of the real `install.sh` against a real
built CLI with an npm shim planted at `binLink`"* is **not** a committed test. It was
out-of-band verification. The three above are what CI enforces; please don't describe the
manual run as a level of coverage.

---

### Blocking

**1. `INSTALL.md:304-317` — the macOS section still says the installer activates.**
> The installer maps the machine to the `mac-arm64` release asset and **hands over to
> `actana setup`**, which writes a **LaunchAgent** … and loads it.

No hunk touches this section, and its code block is still the bare one-liner. #316's last
acceptance criterion is that `INSTALL.md` is not *actively wrong* on the day this lands, and
this is the same phrase (`hands over to`) that `install-sh.test.mjs` now bans from the script's
own header. A Mac operator following this page gets a placed bundle, no LaunchAgent, no daemon,
and no second command. `#323` is the wider sweep; this one is inside this ticket's own criterion.

**2. `packages/panel/src/shared/cores.ts:92` — the Panel's "needs update" remedy stops working.**
`CORE_UPDATE_COMMAND` is the bare one-liner, rendered and copy-to-clipboard'd by
`CoreNeedsUpdate.tsx:32,101`. After this change, pasting it on the Core downloads the new
bundle, writes `versions/<new>`, repoints `current`, and exits — the old daemon keeps running
the old version, so the Panel keeps saying "needs update" and the button never fixes anything.
Its docstring even cites the INSTALL.md sentence this PR rewrote. This is shipped product
behaviour, not a doc: it needs the two-step form or `actana update`, or an explicit decision
recorded here that it moves to a follow-up ticket.

**3. `docs/core-linux-rehearsal.md:81` — the release rehearsal cannot pass as written.**
§2 says *"Paste the printed one-liner. **Do not add `--yes`** — the prompts are the point"*,
then asks the rehearser to tick that **it** asks about lingering, offers the harness CLIs, and
ends by naming `actana pair new`. All of that is `actana setup`'s now, and the page never tells
them to run a second command. Section 2's unticked boxes are declared release blockers by the
page itself. You updated `scripts/rehearse-install.mjs` — its written companion needs the same
edit, and #323's file-surface table does **not** list this file, so nobody else owns it.

**4. Delivery — the decisive check did not run.** `Conventions` failed (commitlint: PR title
130 chars > 120), which cancelled both `E2E — installer (ubuntu, x64)` and
`(debian, x64)`. The title has since been shortened to 101 chars, so a re-run should pass, but
as it stands **the rewritten two-command e2e has never executed on a real systemd machine** —
and the PR notes it could not be run locally either (no Docker, no systemd). Please push/re-run
and let both installer jobs go green before this is marked ready.

---

### Non-blocking, but please address

**5. `packages/cli/src/actana-cli.ts:691` — placement swaps the tree under a running daemon
with no stop guard.** `runActanaSetup` stops the service first when
`plan.replacingTree && service.isActive()`, and the comment there explains why. `cmdPlace` has
no service manager, so on the documented "paste both again to upgrade" path `install.sh` now
does exactly the unguarded swap that guard exists to prevent: `installTree` `rmSync`s the
directory the live daemon is executing from (`actana-tree.ts:89`) and renames a fresh copy over
it. The daemon survives on open inodes, but any lazy `require` or asset read in the window
before `actana setup` restarts it is gone. Either give `place` the guard, or say in its docstring
why the window is acceptable.

**6. `packages/cli/src/actana-setup.ts:319` — `current` is repointed with nothing following it.**
The unit's `ExecStart` is `<currentLink>/bin/actana daemon`. After a `place` that is never
followed by `setup` — the exact state finding 2 produces — a restart or reboot silently starts
the new, un-activated version while `actana.json` and `actana status` still describe the old one.
Previously `current` only ever moved inside a run that immediately reinstalled the unit and
restarted the daemon.

**7. `packages/cli/src/actana-cli.ts:699` — `Launcher` is printed even when nothing was linked.**
When `outcome === "foreign"` because another `actana` is earlier on `PATH`, nothing is written at
`binLink` at all — `actana-setup.test.ts` asserts `fs.existsSync(layout.binLink) === false` for
that case. The summary still prints `  Launcher   <binLink>`, contradicting the `claimLauncher`
note printed three lines earlier. Gate it on `outcome === "linked"`, or print `foreignPath`.

**8. `packages/cli/src/actana-tree.ts:89-90` — the rollback docstrings overclaim.**
If `renameSync(staging, installDir)` throws, the previous `rmSync(installDir)` has already
destroyed the old tree and `<installDir>.incoming` is left on disk — and in `placeCoreBundle`
`hadInstallDir === true` suppresses the cleanup (`actana-setup.ts:315`), leaving `current`
dangling at a deleted directory. Both *"never a `versions/<v>.incoming` for the next run to trip
over"* and *"a failed placement leaves the machine as it was found"* are false for that path.
Rare, but the docstrings should not promise it.

**9. `packages/cli/src/__tests__/actana-setup.test.ts:274` — the test does not exercise the guard
it is named for.** `does not delete a version that was already installed when a re-place fails`
builds its second source by removing `node/bin/node`, which makes `planCorePlacement` throw at
`missingTreeFile` before `placeCoreBundle` is entered — so `if (!hadInstallDir)` never runs.
Fail during the copy instead (`chmod 000`, as the neighbouring test does).

**10. `scripts/fixture-release-server.mjs:75` — still prints one command.** The dev hint leaves
a developer at an installed-but-inactive machine with no second command named.
`rehearsalSetupCommand()` exists for exactly this.

**11. `INSTALL.md:5` — 142 columns.** The rest of the file wraps at ~80; this line is the only
new one that does not.

---

Once 1–4 are settled I expect to approve — the hard half of this ticket is already right.

qcentic-dev:
> Once 1–4 are settled I expect to approve

All four are settled, and 5–11 with them. Nothing deferred. Four commits pushed:
`a22f928`, `47ee0e1`, `393ee79` (and the description correction below).

---

## Blocking

**B1 — `INSTALL.md` macOS section.** Fixed. The code block is both commands, and the prose
now says the first *places the bundle and links the launcher — and stops*, and that the
`actana setup` line it printed is what writes and loads the LaunchAgent. The absolute-path
case is called out by name there, because a Mac that has never had `~/.local/bin` on its
`PATH` is the ordinary case rather than the exotic one. `hands over to` is gone from the page.
While there: `INSTALL.md:5` rewrapped (**N11**), and the Intel-Mac paragraph now says the
refusal happens *before it downloads anything*, which was already true and worth saying on a
page that no longer ends at one command.

**B2 — `CORE_UPDATE_COMMAND`.** Now **`actana update`**, not the two-step form. You asked why,
so:

*What the notice has to do* is clear itself. It clears when the Core announces a protocol the
Panel accepts, which means the **daemon has to end up running the new version** — landing the
tree is only half. That rules out the one-liner alone, which is the bug.

*`install.sh` + `actana setup` would work*, and I did not choose it for two reasons. It is the
wrong gesture: `setup` is the command that *activates a machine* — mints or reuses the mTLS
material, writes the service, offers Harnesses, registers the Core with this machine's own
CLI — and this machine is already active. Every one of those is a no-op-or-worse on a Core
that is merely behind. And it is a full re-install (resolve, download, verify, extract, place)
on a machine that already has a bundle, to fix a version drift.

*`actana update` is exactly the shape of the problem*: resolve, verify against the release's
own `SHA256SUMS`, swap `current`, restart the daemon onto it, keep the pairing material and
the data dir. A paired Panel stays paired, which is what the old docstring's "keeping the
pairing" promised and what a Panel-driven remedy has to preserve. It is one line to paste from
a phone, which is what the copy button is for. And a container Core refuses it and names
`docker compose pull && docker compose up -d` — the right answer there, and more than the
one-liner ever gave that operator.

The docstring said *"The `actana` bundle will eventually own this as an `update` verb
(CONTEXT.md); until that verb exists, the installer is the update path."* The verb exists —
`packages/cli/src/actana-update.ts`, and the Linux e2e drives it. This is the sentence coming
due, and #316 is what forced it.

On the test point you made elsewhere, applied here: **every existing assertion in
`CoreNeedsUpdate.test.tsx` read the constant rather than a literal**, so the whole file would
have gone on passing with a command that does nothing — which is how the old value survived.
One test now binds the value and names the two commands it must not become.

**B3 — `docs/core-linux-rehearsal.md`.** §2 is two steps now. The first checks what installing
*alone* does — no prompts, a verified checksum, "nothing is running yet", exactly one printed
command, and an `actana status` that says there is no Core installed for this user, which is
what installed-but-not-activated looks like from outside. The second is *"paste the line the
installer printed, exactly as printed"*, and carries every prompt box the page had: lingering,
the Harness offers one at a time, no sudo, `actana pair new`, no credential.

It also gains a box the page could not have had before: **the printed line runs**. CI runs it
as printed too, but a person pasting it is what that criterion is actually about, and this is
the only place one does. The title, the intro and the "recording the result" paragraph follow.
`fixture-release-server.mjs`'s dev hint prints both commands as well (**N10**) — it had the
same defect in miniature.

**B4 — delivery.** `Conventions` passes on this push. Both installer jobs ran to completion
this time; their conclusions are in the comment below this one, named individually as asked.

---

## Non-blocking

**N5 — the unguarded swap.** Fixed rather than documented. You are right that the documented
"paste both again" path walks straight into the case `runActanaSetup`'s guard exists for.
`place` now stops the daemon when it is about to replace the directory that daemon is
executing from, says so, and the `setup` line it prints is what brings it back.

Two things about the shape, because "give `place` a service manager" is not free. It consults
the init system **only when there is an existing tree at `installDir` to destroy** — placing a
different version writes a different directory and touches nothing running, and a fresh
machine, which is the ordinary case for this verb, asks the init system nothing at all
(`asks the init system nothing when there is no tree at that path to destroy` pins that). And
it is best-effort in both directions: a service manager that cannot be built, or a `stop` that
fails, must not sink a placement, because placement is the one job this verb has and it has to
work on machines whose init system is not this CLI's business.

Tested on both platforms — `stops a running Core whose own tree it is about to replace`
(systemd) and `stops it through whatever this machine's init system is, not through systemd`
(launchd, asserting `launchctl bootout` and no `systemctl`) — because the hazard is the
filesystem rename, not the unit.

**N6 — `current` repointed with nothing following.** Fixed as output rather than as a refusal.
When an activated install is already there, `place` now names both versions, says the daemon
is still running the old one, and calls the printed command *finishing the upgrade* instead of
telling a Core that nothing is running yet. Refusing felt wrong: the operator has just been
told to run two commands, and the second one is the fix.

**N7 — the phantom `Launcher` row.** Gated on `outcome === "linked"`; otherwise it prints
`Launcher   left alone — <foreignPath> owns \`actana\` here`. Tested against a real file
earlier on `PATH`, asserting `binLink` does not exist and the row does not name it.

**N8 — the overclaiming docstrings.** Fixed the code rather than the prose, because you are
right that the promise is the one worth keeping: `installTree` now moves the existing tree
aside to `<installDir>.previous` and puts it back if the rename throws, instead of deleting it
to make room for a tree that then fails to arrive. Both scratch paths are cleared on entry as
well as on exit. `placeCoreBundle`'s `hadInstallDir` cleanup stays as the belt to that braces,
and its comment now says so rather than claiming to be the mechanism.

**N9 — the test that skipped its own guard.** Fixed the way you suggested: it fails during the
copy (`chmod 000`, like its neighbour, skipped as root) instead of at `missingTreeFile`.
Renamed to describe the property rather than the branch. Two more cases came with N8 — that
neither scratch directory survives a replace, and that a `.previous` from a crashed run is not
adopted.

---

## The description correction

Taken, and thank you — that was an overclaim. The PR description now says the committed
coverage for the npm-bin collision is the two suites (`actana-setup.test.ts` →
*placeCoreBundle — the launcher path still has one owner*, and `actana-machine-cli.test.ts` →
*place* → *does not clobber an `actana` somebody else installed at the same path*) plus
`CORE_REFUSED_VERBS` container-side, and that the manual run was out-of-band checking rather
than a level of coverage.

---

The PR stays a draft and the base stays `feat/release-pipeline`.

qcentic-dev:
## B4 — the first run found a real bug in my own e2e, not a flake

`Conventions` passed, so both installer jobs ran this time rather than being cancelled. **Both
then hit the 30-minute job timeout**, and it was my change that hung them. Reporting it before
the green, because the failure is the more useful half:

```
[setup-e2e] running the one-liner pinned to v0.4.1
[setup-e2e] the one-liner installed without activating, and printed
            `/home/operator/.local/share/actana/current/bin/actana setup`
```

…and nothing after that for thirty minutes.

The install half is exactly right — including the absolute-path case, which is what that
printed line *is*: a fresh Ubuntu container has no `~/.local/bin` at login, so `PATH` never had
it, so `setupCommandFor` correctly chose the path over a bare `actana`. That is #316's fourth
criterion working on a real machine, and it is why the command is extracted rather than retyped.

Then `actana setup` hung. **`install.sh`'s old tail ran setup with `</dev/null` on every path**
— that is where this file's "a piped (non-TTY) run must prompt for nothing" came from, down to
`Linger=yes` with no `--yes` passed. Splitting the step moved setup out from behind that
redirect and onto `asOperator`, which reaches the operator through `machinectl shell` — and
that pipes the session through a **PTY**. So setup was interactive, asked about lingering,
offered the Harness CLIs one at a time, and waited for an answer nobody was there to give.
`interactive` is `process.stdin.isTTY && process.stdout.isTTY`, so the redirect was the whole
of it.

Fixed in `961b060`. It lives in one `activate()` helper so a fourth invocation cannot be
written without it, and the docstring says why it is load-bearing. The convention was already
in the repo and I did not carry it across: every `actana setup` in
`e2e-actana-harnesses-linux.mjs` has `</dev/null` for precisely this reason.

`--yes` would also have stopped the hang and would have been the wrong fix — it tests "prompts
suppressed by flag", where this file tests "no terminal, so nothing to prompt on". The
interactive path is rehearsed by a person once per release, which is what
`docs/core-linux-rehearsal.md` §2 now covers properly.

Verified locally under a real PTY (`script -qec`) that the wrapped form `asOperator` builds —
`{ <cmd> </dev/null\n}` — yields `tty=no` and an empty stdin, so `interactive` is false.

Both jobs are re-running on `961b060`; their conclusions follow in the next comment.

qcentic-dev:
## B4 — green, and the two-command flow has now run on a real systemd machine

**`E2E — installer (ubuntu, x64)`: pass (1m31s)** · **`E2E — installer (debian, x64)`: pass
(1m37s)** · **`Conventions`: pass (40s)** — run
[32743098494](https://github.com/actana/control/actions/runs/32743098494) on `961b060`.

The whole rewritten flow, from the ubuntu log:

```
[setup-e2e] running the one-liner pinned to v0.4.1
[setup-e2e] the one-liner installed without activating, and printed
            `/home/operator/.local/share/actana/current/bin/actana setup`
[setup-e2e] `actana setup` activated the machine unattended, printed no credential, and named `pair new`
[setup-e2e] user unit is active and enabled
[setup-e2e] a test client dialled the core-link with the credential setup registered
[setup-e2e] lingering is enabled — the daemon survives logout
[setup-e2e] `actana status` reports a healthy Core on the pinned v0.4.1
[setup-e2e] `actana pair new` mints a code, and `actana token` refuses and says so
[setup-e2e] stop / start / restart drive the daemon
[setup-e2e] `actana logs` shows the daemon's journal
[setup-e2e] planting a pre-rename actana-harness.service
[setup-e2e] re-running the install with no version pinned, then setup over it
[setup-e2e] upgraded in place to v0.4.2: one unit, same credentials, still healthy
[setup-e2e] restarting the machine
[setup-e2e] the Core came back after reboot with the same identity
[setup-e2e] a tampered download aborted and left the old install untouched
[setup-e2e] `actana update --version 0.4.3` installed exactly that release
[setup-e2e] `actana update` landed the latest release, restarted, and stayed paired
[setup-e2e] `actana token regenerate` invalidated every credential this Core had issued
[setup-e2e] `actana uninstall` removed the unit, launcher and install, and kept the data
[setup-e2e] the one-liner and `actana setup` reinstalled a machine that had been uninstalled
[setup-e2e] `actana uninstall --purge-data` removed the data dir and the credentials too
[setup-e2e] OK — on Ubuntu 24.04 the one-liner installs without activating, `actana setup`
            activates, pinning and in-place upgrades work, and so do the lifecycle verbs
```

Debian 13 ends with the same line.

Three things in there are worth naming, because they are the parts a green suite could have
faked:

- **the absolute path was the one that ran.** A fresh container has no `~/.local/bin` at login,
  so `setupCommandFor` chose the path over a bare `actana`, and the e2e ran what was printed.
  A hard-coded `actana setup` would have passed here and hidden that; instead the criterion
  was actually exercised.
- **`lingering is enabled` with no `--yes`** — the unattended property survived the split,
  which is what `</dev/null` bought back.
- **`the one-liner and `actana setup` reinstalled a machine that had been uninstalled`** — the
  path where the launcher is gone, so the second command can only be the printed absolute one.

Full check set on `961b060`: 18 pass, 5 skipping, 0 fail.

`Conventions` · `Lint` · `Typecheck` · `Unit Tests` · `Secret Scan` · `Dependency Audit` ·
`Promotion gate` · `Train rules` · `Resolve PR image mode` · `E2E — Panel service seam` ·
`E2E — installer (ubuntu, x64)` · `E2E — installer (debian, x64)` ·
`Smoke — Core release tarball (linux-x64)` · `Smoke — Core release tarball (linux-arm64)` ·
`Core image / Resolve registries` · `Core image / Build + smoke (amd64)` ·
`Panel image / Resolve registries` · `Panel image / Build + smoke (amd64)` — all **pass**.
Skipping, as they should on a PR: `Core image (train)`, `Panel image (train)`,
`Resolve train tags`, and the two `Publish the multi-arch manifest` jobs.

Still a draft, base still `feat/release-pipeline`, nothing merged.

mrfqcentic (approved):
## Re-review of #331 at `961b060` — approve

Every finding verified from the enforcing line at the new head, not from the reply. All four
blocking findings are fixed, and 5–11 with them. Nothing deferred.

### B1 — `INSTALL.md` macOS section · **fixed**

`INSTALL.md:305-327` now shows both commands in the code block, and the prose reads *"The first
command maps the machine to the `mac-arm64` release asset, places the bundle and links the
launcher — and stops. Nothing is running yet. Run the `actana setup` line it printed; that is
what writes the **LaunchAgent** … and loads it."* The absolute-path case is named there, which
is the right call for a Mac that has never had `~/.local/bin` on `PATH`.

Consistent with the rest of the page: `hands over to`, `passed straight through` and
`passthrough` return **zero** matches across the whole file, and both `install.sh | bash` blocks
(`:43` and `:313`) are followed by `actana setup`. `docs/core-macos-prerelease-checklist.md:46-49`
was corrected to match. **N11** went with it — `INSTALL.md:5` is rewrapped, and the only lines
over 90 columns left are the pre-existing table rows.

### B2 — `CORE_UPDATE_COMMAND` · **fixed, and the reason I asked holds up**

The whole finding was that landing the tree is half the job, so I read the update path rather
than the constant. `packages/cli/src/actana-update.ts`:

- `:135-137` — stops the daemon when the tree being replaced is the one `current` points at,
  the same guard `runActanaSetup` uses;
- `:138` `installTree`, `:144` `pointSymlink(currentLink, installDir)`;
- `:151` `writeActanaConfig(… version: manifest.version, installDir)` — so `actana.json` and
  `actana status` end up describing the version that is actually running, which the two-step
  form I suggested would *not* have done;
- `:153-161` `service.verb("restart")`, with a thrown error naming `actana logs` if it will not
  come back;
- `:168` `waitForPort` — the return value is the daemon answering on its port.

So `actana update` ends with the daemon running the new version. The notice clears. Accepted,
and it is the better of the three candidates for the reason given: `setup` is the gesture for
activating a machine, and this one is already active.

`CoreNeedsUpdate.tsx:81` was updated to match (*"lands the new version, restarts the daemon, and
keeps the pairing"*), and the new test
`CoreNeedsUpdate.test.tsx` → *"the command it offers actually updates a running Core"* is bound to
the **value** rather than the rendering — the rest of that file reads `CORE_UPDATE_COMMAND` and
would pass just as happily with a command that does nothing. That is the right test to have
added.

Corroborated on a real machine: the e2e log has
`` `actana update` landed the latest release, restarted, and stayed paired ``.

### B3 — `docs/core-linux-rehearsal.md` · **fixed**

§2 is now workable end to end. The install half's boxes ask only for install-half behaviour —
exits without asking anything, says nothing is running yet, prints one `actana setup` command
and only one, `actana status` reports no Core. A new subsection, *"Then the second command —
this is the one with the prompts in it"*, names it explicitly and says to paste **the line the
installer printed**. Lingering, the Harness offers, `pair new` and the no-credential box all
moved under it. **No tick box asks the rehearser to observe `actana setup` behaviour from the
one-liner.**

Better than I asked for: the new box *"The printed line runs. If it was an absolute path, it
worked as pasted; a printed command that is not found is a release blocker on its own"* makes
#316's fourth criterion something a person checks by hand once per release. §1's banner box and
the closing "release blockers" paragraph were both updated to the two-command shape.

*Nit, not a change request:* that box paraphrases the output as *"no Core installed for this
user"*, while `formatActanaStatus` (`packages/cli/src/actana-status.ts:151-152`) actually prints
`Core: not installed` followed by ``Run `actana setup` to install and pair this machine.`` Close
enough to tick; quoting the real first line would be tighter.

### B4 — Delivery · **fixed, and the first failure was the useful half**

Run [32743098494](https://github.com/actana/control/actions/runs/32743098494) on `961b060`,
`event=pull_request` — so it ran on the merge ref, against the post-#329 base. Conclusions by
name, from the job API:

- **`Conventions`: success** (40s)
- **`E2E — installer (ubuntu, x64)`: success** — job 15:09:58 → 15:11:29; step 7,
  *"E2E — the one-liner and the lifecycle on a fresh systemd machine"*, **success**,
  15:10:22 → 15:11:26
- **`E2E — installer (debian, x64)`: success** — job 15:09:58 → 15:11:35; step 7 **success**,
  15:10:26 → 15:11:31

Both ran to completion; neither was cancelled or skipped. I read the ubuntu log rather than
trusting the rollup, and the flow is all there — including the three lines that matter:

```
the one-liner installed without activating, and printed
    `/home/operator/.local/share/actana/current/bin/actana setup`
`actana setup` activated the machine unattended, printed no credential, and named `pair new`
lingering is enabled — the daemon survives logout
```

The printed command was the **absolute** one, and it is what ran — so criterion 4 was exercised
for real rather than hidden by a hard-coded `actana setup`. Full set on this head: 18 pass,
5 skipping, 0 fail.

And the PTY diagnosis was right. `install.sh`'s old tail supplied `</dev/null` on every path;
splitting the step moved `setup` out from behind it and onto `machinectl shell`'s PTY, so
`interactive` went true and it waited for an answer. The fix is one `activate()` helper in
`scripts/e2e-actana-setup-linux.mjs`, used by all three setup invocations, and `</dev/null`
rather than `--yes` is the correct choice — `--yes` would have tested "prompts suppressed by
flag" where this file tests "no terminal, so nothing to prompt on". The unattended property is
still asserted (no `--yes` passed, the `[Y/n]` check still runs, and `Linger=yes` still checked).

### Non-blocking findings

| # | Status | Verified at |
|---|---|---|
| **N5** — no stop guard on the placement path | **fixed** | `actana-cli.ts:722` calls the new `stopDaemonBeingReplaced` (`:787`) between plan and place. It returns early when `!plan.replacingTree` or the install dir does not exist, builds the service manager in a `try`/`catch` so a machine with no supported init system still places, only stops when `isActive()`, prints why, and on failure continues with a warning instead of aborting. Mirrors `actana-update.ts:135` |
| **N6** — `current` repointed with nothing following | **fixed** | `cmdPlace` reads `readActanaConfig` *before* placing and, when a Core is already set up, prints *"already set up as a Core on X, and `current` now points at Y … Nothing has restarted, so it is still running X — finish the upgrade with:"*, then the setup line, then *"That restarts the daemon on the version just placed."* The drift is now surfaced instead of silent |
| **N7** — `Launcher` printed when nothing was linked | **fixed** | `actana-cli.ts:736-738` — the row is `linked`-only, and the foreign case prints ``left alone — <foreignPath> owns `actana` here`` instead of a path with nothing at it |
| **N8** — rollback docstrings overclaimed | **fixed, and more than asked** | `actana-tree.ts:82-115` — the swap now moves the old tree to `<installDir>.previous` and renames it back if the final rename throws, instead of `rmSync` then rename. Both scratch paths are cleared on entry and exit. The docstring's promise is now true |
| **N9** — test skipped its own guard | **fixed** | Renamed to *"keeps the version that was already installed when a re-place fails mid-copy"* and fails with `chmod 000` during the copy, `skipIf` root — so `placeCoreBundle` is actually entered |
| **N10** — fixture server printed one command | **fixed** | `scripts/fixture-release-server.mjs:76-81` prints `install:` and `then:` |
| **N11** — `INSTALL.md:5` at 142 columns | **fixed** | rewrapped |

Each of N5–N7 came with a test, and the macOS variant of N5
(*"stops it through whatever this machine's init system is, not through systemd"*, asserting
`launchctl bootout` and no `systemctl`) is a case I did not ask for and would not have thought
to. *"asks the init system nothing when there is no tree at that path to destroy"* keeps the
fresh-machine "needs no init system" property intact.

### Boundary, re-checked at `961b060`

`.github/workflows/**` · `packages/shared/src/semver.ts` · `packages/cli/src/actana-update.ts` ·
`packages/cli/src/actana-release.ts` · `packages/core/src/core-update-notice.ts` — **0 hunks each**.
`packages/shared/src/actana-release-channel.ts` — **not created** (ADR 0036 D6 → #322).

`packages/panel/**` and `docs/core-macos-prerelease-checklist.md` are new to the diff and are the
B1/B2 fixes; neither is anyone else's ticket.

### The base moved: #329 (`77b9be2`)

`mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, and the branch is 1 behind / 8 ahead. I checked
it rather than taking the flag: **`comm` on the two file lists returns empty** — #329 touched
`.agents/skills/release/SKILL.md`, three workflows, `AGENTS.md`, ADRs 0023 and 0036,
`docs/ci-cd.md` and `scripts/__tests__/workflows.test.mjs`; #331 touches none of them. No textual
conflict is possible.

Semantically clean too:

- **ADR 0036 C2 is byte-identical at the current base** to what this PR was written against —
  #329 did not weaken the constraint #331 implements.
- #329's amendment to **D1** (`0036:144`) says the cut is now a person, that #325 *"touches
  neither `install.sh` nor `ci.yml`'s manifest set"*, and that **#317 is what puts the stamp in
  `install.sh`**. #331 leaves `resolve_version` (`install.sh:276`) and the whole resolution block
  untouched — `install.sh` at this head is byte-identical to the r1 head — so #317's landing site
  survives both changes.
- **D6** still assigns `actana-release-channel.ts` to #322, and #331 does not create it.
- `ci.yml` at the post-#329 base still defines `installer-e2e` (`:756`) with the same step name
  (`:794`), and the green run above executed against that file.

### Claim 2 re-confirmed

`scripts/__tests__/install-sh.test.mjs`, `scripts/lib/setup-e2e.mjs` and
`scripts/rehearse-install.mjs` are **byte-identical** to the head I certified in round 1, so that
verification carries. `scripts/e2e-actana-setup-linux.mjs` changed only by the `activate()`
helper, and its assertions — the four placement artefacts, the three negative traces, running the
extracted command rather than a retyped one — are unchanged.

---

Approving. The tail is gone at the line, the contract moved with it, the two-command flow has now
run on Ubuntu and Debian on a real init system, the update story is coherent again, and the
boundary against #317, #322 and #325 holds after the base moved. The response to N5–N9 went past
what the findings asked for, and `installTree`'s `.previous` swap in particular fixes a failure
mode that had no recovery.

PR #332, opened 2026-08-24
fix: update is channel-aware

### LINKED ISSUE #322: actana update silently downgrades a machine off a beta, and the update notice never fires for its release

qcentic-dev:
Closes #322.

This branch makes `actana update` channel-aware, so a Core's release channel is respected on both sides of the update path:

- A machine running a beta version is no longer silently walked back to the previous release by an update. Prerelease precedence now follows semver §11.4, so `0.4.1-beta` is correctly ordered below `0.4.1` rather than compared equal to it, and a bare `actana update` no longer treats the previous release as newer than the installed beta.
- The update notice now fires when that beta's own line reaches a release. A Core installed from a beta is told when the release on its line ships, instead of never being notified.

The branch carries a single squash commit on top of `beta/0.4.1`: `378bf41` ("fix: update is channel-aware").

### DIFF
--- a/packages/cli/src/__tests__/actana-update.test.ts
+++ b/packages/cli/src/__tests__/actana-update.test.ts
@@ -384,3 +384,110 @@ describe("releases this machine cannot use", () => {
     await expect(update()).rejects.toThrow(/releases\/latest|no release/i);
   });
 });
+
+// #322. `/releases/latest` cannot see a prerelease, so on a machine installed
+// from a beta a bare `actana update` resolves the *previous* release. Before
+// the guard, "0.4.0" !== "0.4.1-beta" waved straight past the already-current
+// check and the machine was stopped, swapped and restarted onto an older
+// version — and told it had been updated.
+describe("a machine running a beta", () => {
+  const BETA = "0.4.1-beta";
+
+  beforeEach(() => {
+    existingInstall(BETA);
+  });
+
+  describe("when the newest release is older than the beta", () => {
+    beforeEach(() => {
+      writeRelease({ dir: releaseDir, version: "0.4.0", target: TARGET });
+    });
+
+    it("does not replace the tree", async () => {
+      const service = fakeService();
+      const result = await update({ service });
+
+      expect(result.updated).toBe(false);
+      expect(result.version).toBe(BETA);
+      expect(fs.realpathSync(layout.currentLink)).toBe(
+        fs.realpathSync(installDirFor(layout, BETA)),
+      );
+      expect(fs.existsSync(installDirFor(layout, "0.4.0"))).toBe(false);
+      expect(readActanaConfig(layout.configDir)!.version).toBe(BETA);
+      // The daemon never noticed: no stop, no restart.
+      expect(service.verbs).toEqual([]);
+      expect(service.stops).toBe(0);
+    });
+
+    it("says what it is on, what the release is, and what to do about it", async () => {
+      await update();
+      const said = out.join("\n");
+      expect(said).toContain(BETA);
+      expect(said).toContain("prerelease");
+      // The line, in ADR 0036 D1's vocabulary — the release this machine is
+      // actually waiting for is its own line's.
+      expect(said).toContain("0.4.1 line");
+      expect(said).toContain("0.4.0");
+      expect(said).toContain("actana update --version 0.4.0");
+    });
+
+    it("downloads nothing — the refusal is a decision, not a failed attempt", async () => {
+      const fetcher = fixtureFetcher(releaseDir, CHANNEL);
+      await update({ fetcher });
+      expect(fetcher.asked.some((url) => url.endsWith(".tar.gz"))).toBe(false);
+      expect(fetcher.asked.some((url) => url.endsWith("SHA256SUMS"))).toBe(false);
+    });
+
+    // An explicit pin is an operator's decision and is not second-guessed:
+    // this is how a Panel↔Core version lock is recovered.
+    it("still installs an older release when the operator pins it", async () => {
+      const result = await update({ requestedVersion: "0.4.0" });
+
+      expect(result.updated).toBe(true);
+      expect(result.version).toBe("0.4.0");
+      expect(result.previousVersion).toBe(BETA);
+      expect(fs.realpathSync(layout.currentLink)).toBe(
+        fs.realpathSync(installDirFor(layout, "0.4.0")),
+      );
+      expect(readActanaConfig(layout.configDir)!.version).toBe("0.4.0");
+    });
+  });
+
+  // The other direction, and the reason the guard is a comparison rather than
+  // "never move off a prerelease": the beta's own line releasing is exactly
+  // what this machine is waiting for, and it must not be stranded.
+  it("takes its own line's release the day it exists", async () => {
+    writeRelease({ dir: releaseDir, version: "0.4.0", target: TARGET });
+    writeRelease({ dir: releaseDir, version: "0.4.1", target: TARGET });
+
+    const result = await update();
+    expect(result.updated).toBe(true);
+    expect(result.previousVersion).toBe(BETA);
+    expect(result.version).toBe("0.4.1");
+    expect(readActanaConfig(layout.configDir)!.version).toBe("0.4.1");
+  });
+
+  it("takes a later line's release too", async () => {
+    writeRelease({ dir: releaseDir, version: "0.4.2", target: TARGET });
+    expect((await update()).version).toBe("0.4.2");
+  });
+});
+
+// The guard is about ordering, not about prereleases: a `latest` that has been
+// moved backwards would walk a release-running machine back just as silently.
+describe("a bare update that would move any machine backwards", () => {
+  beforeEach(() => {
+    existingInstall("0.5.0");
+    writeRelease({ dir: releaseDir, version: "0.4.0", target: TARGET });
+  });
+
+  it("refuses, and does not call the installed version a prerelease", async () => {
+    const result = await update();
+    expect(result.updated).toBe(false);
+    expect(result.version).toBe("0.5.0");
+    const said = out.join("\n");
+    expect(said).toContain("0.5.0");
+    expect(said).toContain("0.4.0");
+    expect(said).not.toContain("prerelease");
+    expect(said).toContain("actana update --version 0.4.0");
+  });
+});
--- a/packages/cli/src/actana-release.ts
+++ b/packages/cli/src/actana-release.ts
@@ -30,12 +30,19 @@ import type { ReleaseFetcher } from "@actana/shared/actana-release-fetch";
 export { nodeReleaseFetcher, type ReleaseFetcher } from "@actana/shared/actana-release-fetch";
 
 export {
+  BETA_SUFFIX,
+  betaVersionForLine,
   DEFAULT_API_BASE,
   DEFAULT_DOWNLOAD_BASE,
   DEFAULT_REPO,
+  isBetaVersion,
   latestReleaseUrl,
+  lineOf,
   parseLatestTag,
   releaseChannel,
+  releaseTagUrl,
+  resolveLine,
+  type LineResolution,
   type ReleaseChannel,
 } from "@actana/shared/actana-release-channel";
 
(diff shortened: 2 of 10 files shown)


qcentic-adm (approved):
Gate review (second identity), judged as something about to sit on the 0.4.1 release train. **Approve.** All seven claims proven at the enforcing line; CI fully green at 378bf41; the branch carries exactly one squash commit and is not behind `beta/0.4.1`.

**1. No silent walk-back off a beta, on any update path.** The enforcing line is `packages/cli/src/actana-update.ts:117` — `if (!opts.requestedVersion && isNewerSemver(config.version, version))` — placed before `installDirFor`, any download, and any service stop. The other paths cannot act at all: `checkForUpdate` is alert-only ("never acted on", `packages/shared/src/actana-update-check.ts`), and the daemon notice only logs (`packages/core/src/core-update-notice.ts`). An explicit `--version` pin remains an operator decision, not a silent path. Test: `actana-update.test.ts` › "a machine running a beta › when the newest release is older than the beta › does not replace the tree" (also asserts no stop/restart and nothing downloaded).

**2. The notice fires when the beta's own line releases.** `isNewerSemver("0.4.1", "0.4.1-beta")` is now true (prerelease-below-release clause, `packages/shared/src/semver.ts:119-126`), reaching the notice through `checkForUpdate` at `actana-update-check.ts:148` with no special case in the notice file. Test: `core-update-notice.test.ts` › "a Core running a beta › is told when its own line's release lands", with the throttle pinned by "is told once, on the same throttle as everyone else".

**3. Exact `x.y.z-beta`, no counters as a supported form.** `-beta` is compared as an ordinary alphanumeric identifier (`semver.ts:93-135`); nothing orders a beta counter. `isBetaVersion("0.4.1-beta.1") === false` is asserted (`actana-release-channel.test.ts`), `BETA_SUFFIX`/`betaVersionForLine` construct only `x.y.z-beta`, and no product code anywhere constructs a counted beta. The numeric-identifier branch exists solely for ADR 0023 D30 backport RCs (`rc.2` < `rc.10`), as the issue requires.

**4. ADR 0036 D6's separation holds.** `ReleaseChannel` keeps meaning repo + hosts (`actana-release-channel.ts:22-29`, unchanged); the new noun is a *line* with its own vocabulary and `LineResolution` type in the same module, documented against D6 at the seam. The CLI still has one release module (`actana-release.ts` re-exports).

**5. No resolution path makes a listing call.** `resolveLine` (`actana-release-channel.ts:144-157`) is pure over two named-tag existence answers; the only endpoints in product code are `/releases/latest` and `/releases/tags/v<version>` — no `GET /releases` listing exists anywhere, and the "never resolves outside its line" test pins the property.

**6. Nothing depends on unlanded work.** `install.sh` is untouched and carries no line stamp; `resolveLine` takes `line` as an argument and has **no product caller** — created here, consumed by #317, exactly as D6 assigns. No file in this diff overlaps `feat/release-pipeline`.

**7. Release-machine behaviour unchanged in every steady state.** Plain `x.y.z` pairs answer identically (regression suite "plain x.y.z pairs answer exactly as they did"; old tests kept verbatim). The one changed case is a `latest` moved backwards, where the old behaviour *was* the silent-downgrade bug: it now refuses with the pin named — disclosed in the commit and pinned by "a bare update that would move any machine backwards".

The twin-rule agreement is enforced by `scripts/__tests__/version-rule-agreement.test.mjs` (runs in the root suite; I ran it plus all three package suites locally — green).

**Non-blocking notes** (none rises to a request for changes):
- `isBetaVersion` (`actana-release-channel.ts:100`) is looser than its docstring: `endsWith("-beta")` also accepts `not.a.version-beta` / `v0.4.1-beta`. Latent here (no caller), but #317 consumes it next — tighten to the exact shape or soften the doc before then.
- `lineOf` duplicates `versionCore`'s logic within the same package; it could delegate and keep the domain name.
- ADR 0036 is cited throughout but lives on `feat/release-pipeline`, not this tree; the citations dangle until that branch reaches the train. Both target `beta/0.4.1`, so sequencing is a train-level concern, not this PR's — but the train should not promote without it.
- The agreement test's cross-product is a fixed 16-string table; a divergence outside it (e.g. leading-zero identifiers) would pass. The named branch assertions mitigate.

Issue #322's "Branching" note named #314's feature branch as the base; this PR targets the train directly. The surfaces are disjoint from `feat/release-pipeline` (zero overlapping files), the wave-1 ordering D6 asks for ("land before or with #318") is preserved, and the gate brief names `beta/0.4.1` as the intended base — noted, not blocking.

mrfqcentic (approved):
Gate review against #322 and ADR 0036 (read from origin/feat/release-pipeline, not on this base). All checks proven from enforcing lines:

1. No walk-back: actana-update.ts:117 `if (!opts.requestedVersion && isNewerSemver(config.version, version))` refuses with updated:false before any download; runActanaUpdate is the only update path (sole caller actana-cli.ts:962). Tests: 'does not replace the tree', 'downloads nothing', 'still installs an older release when the operator pins it', 'takes its own line's release the day it exists' (actana-update.test.ts:405+).
2. Notice fires: actana-update-check.ts:148 `updateAvailable: latest !== null && isNewerSemver(latest, current)` — with isNewerSemver("0.4.1","0.4.1-beta")===true the notice reaches runUpdateNotice. Test: 'is told when its own line's release lands' + throttle test (core-update-notice.test.ts:151+).
3. Exact x.y.z-beta: BETA_SUFFIX='-beta' (actana-release-channel.ts:92); isBetaVersion("0.4.1-beta.1")===false asserted; numeric-identifier precedence covered only by rc.N backport shapes per ADR 0023 D30. No counted beta anywhere.
4. D6 separation holds: ReleaseChannel type and its 'which repository, which hosts' meaning untouched; the line concept is a separate section with its own nouns (lineOf, resolveLine, LineResolution), explicitly documented as not a second meaning of ReleaseChannel.
5. No listing call: resolveLine is pure over two named-tag existence booleans; the only new URL is /releases/tags/v<version>. No GET /repos/<repo>/releases anywhere.
6. No #317 stamp assumption: install.sh untouched; resolveLine takes line as a parameter and has no product caller yet — created for #317 to consume, per D6.
7. Release machine unchanged: guard fires only when installed strictly newer than resolved; already-current guard and notice logic untouched; one deliberate, tested extension — a latest moved backwards now refuses on a release machine too, which is the issue's root rule applied generally.

CI on 378bf41 (run 32747478408): all 20 jobs pass; 3 train jobs skip as expected on pull_request. Ready.

PR #333, opened 2026-08-24
feat(install.sh): the stamped line is the channel, resolved per line (#317)

qcentic-dev:
Implements #317 under [ADR 0036](https://github.com/actana/control/blob/feat/release-pipeline/docs/adr/0036-the-beta-release-channel.md) **D1** and **D2**, against C4's URL table. The mechanism was already decided by the record; this is the implementation of it, not a redesign.

Base is `feat/release-pipeline` — the feature branch for #314. Not `main`, not `beta/0.4.1`.

## What lands

**`install.sh` carries a line, and the line is the channel (D1).** A script fetched by `curl … | bash` cannot know its own URL — no `$0`, no `BASH_SOURCE`, no argv naming one — so "the ref is the channel" can only mean *the copy of the file on that ref differs*. Promotion is a fast-forward (0023 D5), so the train tip's bytes **become** `main`'s bytes: any design storing a channel constant in the file is wrong by construction. What survives a fast-forward is a value true on both sides of it, and the train's own version is one. So the file carries `LINE="0.4.1"`, written by the cut alongside the six manifests, and nothing in it names a channel.

**`resolve_version` is D2, in order:**

1. explicit `--version` / `ACTANA_VERSION` wins, and asks nothing;
2. the release `v<line>`, if that Release exists;
3. otherwise `v<line>-beta`, if that Release exists;
4. otherwise `/releases/latest` — what the script read before the stamp existed, kept as the terminal fallback.

**No step enumerates releases.** `GET /repos/<repo>/releases` answers every line newest-first, so "the newest prerelease" would hand a machine installing one line's beta the beta of another. The stamp determines the tag instead; the steady-state path is the single call at step 2 — the same number the stable path made before.

**A probe has three answers, not two.** Steps 2 and 3 read the HTTP *status*, never the fetch's exit code: 200 is yes, a clean 404 is no, and anything else — a 403 rate limit, a 5xx, a TLS or DNS failure — stops the install with the status named. Falling through on a failed question would pick a different version to install, which on the public door means a beta served from `main` and at a release ref means a pin that floats.

## ⚠️ This changes the stable public one-liner's default, deliberately

The script read `/releases/latest` on **every** unpinned install (`install.sh:248` before this change). It now reads that endpoint only at step 4. That is a behaviour change to the door printed in `README.md`, `INSTALL.md` and on the landing page, and the record already priced it — so, stated plainly rather than left to review:

- **A backport produces no divergence.** 0036 D2 says so in as many words: 0023 D28 keeps `latest` on the current line after a backport publishes from a release branch, and the stamp reaches the same answer *without depending on that assertion continuing to hold*. D2 calls that a small strengthening rather than a change.
- **The promotion window is real, bounded and accepted (D3).** `advance` fast-forwards `main` and pushes the tag before `release` runs, so for the duration of one release run `main` carries a line whose Release does not exist yet. If a beta of that line was cut, step 3 answers and the public door serves a prerelease for that run; if none was, step 4 answers with today's newest release and there is no window at all. D3 also names the unbounded case — a red or cancelled `release` job — and its recovery: re-dispatch the release (0023 D40).
- **The one state where D2 is genuinely worse is a rollback, and that is D24.** 0023 D44's rollback re-points `latest` and flips the Release flag, and by its own design does not rewrite `main`. D2 resolves from the stamp and not from that flag, so a fresh `curl …/main/install.sh | bash` keeps installing the rolled-back release while every Docker pull has already moved off it. The recovery is the hotfix train, which re-stamps `main` in the same fast-forward. Recorded in the ADR, repeated here because it is a regression against today's script.

## Proof — all three channels, no network

The channel lives in the script's own bytes, so the only honest way to run "the copy on the train" is to restamp a copy — which is exactly what a cut does. `restamp` in the test file is the cut's `sed`; every case that claims to be a different door runs a genuinely different copy of the shipped script. Both halves are asserted: the version installed, **and which endpoints were asked** — "resolves per line and enumerates nothing" is a claim about the calls.

| Door | Copy | Fixture sees | Installs |
| --- | --- | --- | --- |
| `…/control/main/install.sh` | the shipped script, stamp and all | `0.3.0`, `0.4.1`, `0.4.1-beta`, `9.9.0-beta` | `0.4.1` — the release, at step 2, and `/releases/latest` is never consulted |
| `…/control/beta/0.9.0/install.sh` | restamped `0.9.0` | `0.9.0-beta`, `1.5.0-beta`, `2.0.0-beta` | `0.9.0-beta` — never the two newer betas on other lines |
| `…/control/v0.1.0/install.sh` | restamped `0.1.0` | `0.1.0`, `0.2.0`, `0.4.1` | `0.1.0` — a pin, C4's second row, with no machinery of its own |

Plus: the **same restamped copy** run against the promoted state of its line answers with the release without differing by a byte (the fast-forward, made observable); step 4 still answers for a line that has published nothing; a beta string is `x.y.z-beta` **exactly**, with no counter in the tag, the asset name or the reported version (C1); `--version` overrides on both doors, `v`-prefixed or bare, including the pinned `x.y.z-beta` form; `ACTANA_VERSION` does the same; `--channel` is refused and no environment variable moves the channel; `--help` names *this copy's* line; and `--base-url` covers the tags endpoint the resolution now reads.

Negative cases, because each of these used to install the wrong thing quietly rather than fail: a forced 403 or 500 on the step-2 probe stops the install with the status named, on the public door and at a release ref alike, and never reaches the beta or `/releases/latest`; a forced 404 still behaves exactly like a missing release; a machine with neither curl nor wget says so instead of exiting 1 in silence; and the download failure names where the version came from rather than a flag nobody passed.

`scripts/fixture-release-server.mjs` grows the half it was missing: `/releases/latest` now **excludes prereleases**, because GitHub's does. That exclusion is the whole reason the old default could never find a beta, and a fixture more generous than the real endpoint would let a beta install pass here and fail on github.com. The tags route it already served is what steps 2 and 3 ask.

`scripts/e2e-actana-setup-linux.mjs`'s install channel now serves a copy of the shipped script restamped onto the line its own newest release is on — the same one-line edit the unit suite and a real cut perform, shared as `writeRestampedInstaller`. It served the repository copy before, whose stamp is the *pinned* version on every run, so under D2 the unpinned re-run took that version and the harness's own assertion could not be satisfied by any correct script. See "What changed after review" below.

`docs/ci-cd.md` § "Cutting a train" gains the stamp as a seventh file and **not** a seventh manifest: its own `sed`, never added to `files=()`, because `install.sh` is not a workspace package and `Train rules`' manifest set refuses to grow past the packages by design (D4). The `files=()` ↔ `MANIFESTS` binding is untouched.

## Scope kept

- `packages/shared/src/actana-release-channel.ts` **not created and not edited** — D6 assigns it to #322, which landed on another feature branch. `ReleaseChannel` there means *which repository and which hosts*, and this ticket does not overload it. The new noun is a **line**.
- Nothing under `.github/workflows` — held open by #326 across this wave. **The consequence is a real gap, named below.**
- `scripts/build-core-tarball.mjs` and `scripts/lib/core-tarball.mjs` untouched.
- **`actana update` after a beta install is #322's, and it is worth knowing about.** This PR is the first thing that can put a beta bundle on a machine, and `packages/cli/src/actana-update.ts` still resolves an unpinned update to the newest *release* — so a machine installed from `…/beta/0.9.0/install.sh` is moved back to stable on its next update and told it succeeded. #317 puts that out of scope in as many words and ADR 0036 D6 assigns the prerelease-aware comparison to [#322](https://github.com/actana/control/issues/322); noted here so the next reader does not have to rediscover the loop.
- No merge, no change to `main`, and this PR stays a draft.

## What changed after review

- **`E2E — installer (ubuntu, x64)` and `E2E — installer (debian, x64)` were red on `a28c898`, and this diff caused it.** Both are required checks on `main` and both were green on the base. The install channel served the *repository* copy of the script, stamped with the pinned version, and then asserted the unpinned re-run resolved the next patch — a state no correct script can satisfy under D2, and one that is unreachable in production (`main`'s stamp is behind the newest release only during D3's window, where the release does not exist yet, and during D24's rollback). The channel now serves a copy restamped onto its own newest line, which makes it `main` *after* that release was promoted — the door the assertion has always been about. The pinned phase is unaffected: `--version` is step 1.
- **`release_exists` conflated absence with failure.** Fixed as described above; the failure modes it opened — a beta served from the public door after one transient failure, and C4's second row ceasing to be a pin — are the reason it was worth a status-reading probe rather than an exit code.
- **The probe's redirection swallowed `fetch_url`'s own `die`.** A machine with neither curl nor wget exited 1 with no output; the question is now asked up front, beside `have tar`.
- **The `SHA256SUMS` failure told operators to drop a flag they never passed.** Steps 2 and 3 can reach it with no `--version` in play, and on a beta door "drop `--version` to install the latest" is wrong twice over. The version's origin is recorded where it is decided and read back in the message.
- **The in-file cut `sed` was GNU-only.** It is now the portable form `docs/ci-cd.md` prints, which matters because it sits one line above the value a person cutting a train edits.

## Known gap, deliberately left

**D4's separate `Train rules` assertion is not in this PR, and it is now tracked as [#335](https://github.com/actana/control/issues/335).** `.github/workflows/ci.yml` was held open by #326 across this wave, so nothing goes red on a train when a cut forgets the stamp. What stands in the meantime: `scripts/__tests__/install-sh.test.mjs` asserts the stamp is a plain `x.y.z` **equal to the workspace version** and that nothing in the file assigns a channel, and the runbook step prints the stamp back for the person cutting. `docs/ci-cd.md` says this plainly rather than implying the check exists.

Refs #317, #314. Under ADR 0036 D1, D2, D4, D6 and C1–C4.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -53,6 +53,9 @@ SHASUMS_ASSET="SHA256SUMS"
 
 REPO="${ACTANA_REPO:-$DEFAULT_REPO}"
 VERSION="${ACTANA_VERSION:-}"
+# Where $VERSION came from, in words, for the one message that has to say so:
+# a failed download can no longer assume a flag was passed (see `main`).
+VERSION_ORIGIN=""
 # One base URL replaces both of GitHub's hosts. That is what lets CI run the
 # real one-liner against locally built artifacts, with no published release and
 # no network.
@@ -68,9 +71,12 @@ BASE_URL="${ACTANA_BASE_URL:-}"
 # fast-forward has made those bytes main's own.
 #
 # One assignment on one line, so the cut can rewrite it in place and a reviewer
-# can see the whole of the change:
+# can see the whole of the change. Copied verbatim from docs/ci-cd.md, down to
+# the `-i.bak`, because this comment sits one line above the value being edited
+# and is the copy-paste source a person cutting a train reaches for first —
+# BSD `sed` on macOS requires an argument to `-i` and GNU `sed` accepts one:
 #
-#   sed -i 's/^LINE=".*"$/LINE="x.y.z"/' install.sh
+#   sed -i.bak 's/^LINE=".*"$/LINE="x.y.z"/' install.sh && rm -f install.sh.bak
 LINE="0.4.1"
 
 say() {
@@ -304,12 +310,106 @@ sha256_of() {
 
 # ─── release resolution ──────────────────────────────────────────────────────
 
-# Does $REPO publish a Release on this tag? Existence is the whole question —
-# the tag itself is fully determined by the stamp — so the body is discarded
-# and a 404 is the answer rather than an error. Both curl and wget already
-# report one as a failure, which is why this needs no parsing.
+# The HTTP status of one GET, as three digits, or the empty string when the
+# request never reached an answer at all — DNS, TLS, or a refused connection.
+#
+# **This exists because absence and failure are different answers, and an exit
+# code cannot tell them apart.** curl and wget fail a 404, a 403 rate limit, a
+# 502 and a dead resolver identically, so a probe that read the exit code would
+# call all of them "no such release" — and under D2 that turns one transient
+# failure on the step-2 probe into a beta served from the public door, which is
+# the outcome D1 exists to prevent, and turns C4's pinned second row into a
+# floating install. Nothing here parses a body; the status is the whole answer.
+#
+# No `--retry` on the curl call, deliberately: `--retry` cannot rewind
+# `-o /dev/null`, so a retried 5xx ends as a write error with **no code
+# printed** — the one input this function exists to return. Retrying is
+# `http_status`'s job, one line down, where a status can be looked at first.
+http_status_once() {
+  if have curl; then
+    # No `-f`: the body is discarded either way, and `-f` would hide the very
+    # status this is here to read. `%{http_code}` is the last response's, so a
+    # redirect is followed to the answer that matters, and a transport failure
+    # prints `000` — which this reports as no answer rather than as one.
+    status=$(curl -sSL -o /dev/null -w '%{http_code}' "$1" 2>/dev/null || true)
+    case $status in
+      000 | "" | *[!0-9]*) status="" ;;
+    esac
+    printf '%s' "$status"
+    return 0
+  fi
+
+  # wget has no `-w`, so the status is read out of what it says. GNU wget
+  # prints the response headers under `-S`; busybox's rejects `-S` outright but
+  # names the code in its own line. So the header form is tried first, the bare
+  # run is the fallback, and both shapes are matched — and when neither yields
+  # a code, this returns nothing and the caller stops rather than guessing.
+  for wget_opt in -S ""; do
+    status=$(
+      LC_ALL=C wget $wget_opt -O /dev/null "$1" 2>&1 |
+        LC_ALL=C sed -n \
+          -e 's|.*HTTP/[0-9.]*[[:space:]]\{1,\}\([0-9][0-9][0-9]\).*|\1|p' \
+          -e 's|.*response\.\.\.[[:space:]]*\([0-9][0-9][0-9]\).*|\1|p' |
+        tail -n 1
+    )
+    if [ -n "$status" ]; then
+      break
+    fi
+  done
+  printf '%s' "$status"
+}
+
+# The same, retried while the answer looks transient.
+#
+# `fetch_url` gets its retries from curl; this one cannot (see above), so the
+# loop is here and it is the better place for it anyway: a 404 is an answer and
+# is returned at once, while a 429, a 5xx or no answer at all is worth asking
+# again before an install is refused over a blip.
+http_status() {
+  attempt=1
+  while :; do
+    status=$(http_status_once "$1")
+    case $status in
+      "" | 429 | 5[0-9][0-9]) ;;
+      *) break ;;
+    esac
+    if [ "$attempt" -ge 3 ]; then
+      break
+    fi
+    sleep "$attempt"
+    attempt=$((attempt + 1))
+  done
+  printf '%s' "$status"
+}
+
+# Does $REPO publish a Release on this tag?
+#
+# Three answers, not two: yes, no, and "the question could not be asked". Only
+# a clean 404 is absence. Anything else stops the install here, because every
+# other status is a statement about the connection or about GitHub rather than
+# about this release — and falling through on one would pick a *different*
+# version to install, silently. `--version` needs none of this and is what the
+# messages point at.
 release_exists() {
-  fetch_url "$API_BASE/repos/$REPO/releases/tags/$1" >/dev/null 2>&1
+  tag_url="$API_BASE/repos/$REPO/releases/tags/$1"
+  tag_status=$(http_status "$tag_url")
+  case $tag_status in
+    200) return 0 ;;
+    404) return 1 ;;
+    "")
+      die "could not reach $tag_url while resolving the $LINE line.
+  That is not the same as \"there is no such release\", so nothing is installed
+  rather than something else being chosen — check the network, or name the
+  version you want with --version, which asks nothing."
+      ;;
+    *)
+      die "$tag_url answered HTTP $tag_status while resolving the $LINE line.
+  403 is usually GitHub's unauthenticated rate limit, 60 requests an hour per
+  IP; 5xx is GitHub. Neither means the release is missing, so nothing is
+  installed rather than a different version being chosen — retry later, or name
+  the version you want with --version, which asks nothing."
+      ;;
+  esac
 }
 
 # **The mechanism is ADR 0036 D1's stamped line, and this is the whole of D2.**
@@ -340,16 +440,23 @@ release_exists() {
 resolve_version() {
   if [ -n "$VERSION" ]; then
     VERSION=${VERSION#v}
+    VERSION_ORIGIN="That version came from --version or ACTANA_VERSION — check it, or drop it to
+  take the version this copy of the script installs by default."
     return 0
   fi
 
   if [ -n "$LINE" ]; then
     if release_exists "v$LINE"; then
       VERSION=$LINE
+      VERSION_ORIGIN="Nothing chose that version: it is the $LINE line this copy of the script is
+  stamped with, and there is no flag to drop. Name another with --version."
       return 0
     fi
     if release_exists "v$LINE-beta"; then
       VERSION="$LINE-beta"
+      VERSION_ORIGIN="Nothing chose that version: it is the current beta of the $LINE line this copy
+  of the script is stamped with, and there is no flag to drop. Name another
+  with --version."
       return 0
     fi
   fi
@@ -370,6 +477,8 @@ resolve_version() {
       head -n 1
   )
   [ -n "$VERSION" ] || die "no release tag in the answer from $latest_url — is $REPO the right repository?"
+  VERSION_ORIGIN="Nothing chose that version: it is the newest release $REPO publishes, and there
+  is no flag to drop. Name another with --version."
 }
 
 # ─── main ────────────────────────────────────────────────────────────────────
@@ -387,6 +496,12 @@ main() {
   fi
 
   have tar || die "tar is not installed — it is needed to unpack the Core."
+  # Asked once, here, rather than left to the first fetch. `http_status`
+  # redirects wget's own stderr into a pipe it parses, so a `die` raised from
+  # inside a probe would be swallowed on its way out; asking up front means the
+  # machine that has neither tool is told so instead of exiting silently.
+  have curl || have wget ||
+    die "neither curl nor wget is installed — one of them is needed to download the Core."
 
   detect_target
   resolve_version
@@ -411,8 +526,9 @@ main() {
   sums_file="$work_dir/$SHASUMS_ASSET"
   fetch_url "$release_url/$SHASUMS_ASSET" "$sums_file" ||
     die "could not fetch $release_url/$SHASUMS_ASSET. Either there is no release $tag,
-  or this machine cannot reach $DOWNLOAD_BASE — check the version, or drop
-  --version to install the latest."
+  or it publishes no checksums, or this machine cannot reach $DOWNLOAD_BASE.
+  Nothing was installed.
+  $VERSION_ORIGIN"
 
   expected=$(awk -v name="$asset" '$2 == name || $2 == "*" name { print $1 }' "$sums_file" | head -n 1)
   [ -n "$expected" ] || die "release $tag has no build for $TARGET.
--- a/install.sh
+++ b/install.sh
@@ -14,6 +14,15 @@
 # operator runs afterwards. There is no flag for this — it is what the script
 # does.
 #
+# **The copy you fetched decides what it installs** (ADR 0036 D1). A script
+# read from a pipe cannot know its own URL — there is no `$0`, no BASH_SOURCE,
+# no argv naming one — so "the ref is the channel" can only mean that the copy
+# of this file on that ref differs. Each copy therefore carries the *line* it
+# was cut from, written by the cut; `resolve_version` turns that line into a
+# release tag or into that line's beta tag. There is no `--channel` flag and no
+# environment variable for it: promotion is a fast-forward, so a channel
+# constant committed on a train would become `main`'s own bytes.
+#
 # **It does not know the install layout, and must not learn it.** Placement is
 # `"$extracted/bin/actana" place` — the CLI in the tree that was just verified,
 # running on the Node pinned inside it. ACTANA_HOME, ACTANA_CONFIG_DIR,
@@ -49,6 +58,21 @@ VERSION="${ACTANA_VERSION:-}"
 # no network.
 BASE_URL="${ACTANA_BASE_URL:-}"
 
+# ─── the line this copy installs (ADR 0036 D1) ───────────────────────────────
+#
+# The one value a cut writes into this file, alongside the six manifests
+# (docs/ci-cd.md § "Cutting a train"). It is a **line** — `x.y.z` — and not a
+# channel: `resolve_version` below turns it into either that line's release or
+# that line's beta, which is what makes the same bytes correct on a train,
+# where only the beta tag exists, and on `main` after the promotion
+# fast-forward has made those bytes main's own.
+#
+# One assignment on one line, so the cut can rewrite it in place and a reviewer
+# can see the whole of the change:
+#
+#   sed -i 's/^LINE=".*"$/LINE="x.y.z"/' install.sh
+LINE="0.4.1"
+
 say() {
   printf '%s\n' "$*"
 }
@@ -82,7 +106,8 @@ and then, as a separate command this script does not run for you:
   actana setup [options]
 
 Options:
-  --version <v>     Install this exact release (default: the latest release)
+  --version <v>     Install this exact version, release or beta. Overrides
+                    everything below; the default is this copy's own line
   --repo <slug>     GitHub repository to install from
   --base-url <url>  Fetch releases from here instead of GitHub (testing)
   --help            Show this help
@@ -98,6 +123,15 @@ ACTANA_HOME, ACTANA_BIN_DIR, ACTANA_CONFIG_DIR, ACTANA_DATA_DIR and the XDG
 variables move where the bundle lands; they are read by the CLI, so this
 script and `actana setup` cannot disagree about where anything is.
 EOF
+  # The channel this copy installs from — printed rather than written into the
+  # heredoc above, because it is the one line of this help that differs between
+  # copies of the script (ADR 0036 D1, D2).
+  printf '\n'
+  printf 'This copy is stamped with the %s line, and that stamp is its channel:\n' "$LINE"
+  printf 'it installs v%s if that release exists, otherwise v%s-beta, and\n' "$LINE" "$LINE"
+  printf 'otherwise the newest release. Which copy you fetched is the whole of\n'
+  printf 'the choice: there is no --channel option and no environment variable\n'
+  printf 'that selects one.\n'
 }
 
 # ─── argument parsing ────────────────────────────────────────────────────────
@@ -270,19 +304,61 @@ sha256_of() {
 
 # ─── release resolution ──────────────────────────────────────────────────────
 
-# `latest` is one API call; a pinned version is none. A pinned install that
-# still asked what the latest release was would be one API change away from
-# quietly installing something else.
+# Does $REPO publish a Release on this tag? Existence is the whole question —
+# the tag itself is fully determined by the stamp — so the body is discarded
+# and a 404 is the answer rather than an error. Both curl and wget already
+# report one as a failure, which is why this needs no parsing.
+release_exists() {
+  fetch_url "$API_BASE/repos/$REPO/releases/tags/$1" >/dev/null 2>&1
+}
+
+# **The mechanism is ADR 0036 D1's stamped line, and this is the whole of D2.**
+# In order:
+#
+#   1. an explicit --version / ACTANA_VERSION wins, and asks nothing;
+#   2. the release `v<line>`, if that Release exists;
+#   3. otherwise `v<line>-beta`, if that Release exists;
+#   4. otherwise `/releases/latest` — exactly what this script read before the
+#      stamp existed, kept as the terminal fallback.
+#
+# On a train only the beta tag exists, so step 3 answers. On `main` after a
+# promotion the release exists, so the same bytes answer at step 2. At a
+# release tag the stamp is that tag's own version and step 2 pins it, and at a
+# beta tag the file carries the line like every other copy — which is why that
+# ref is an alias for the train's door and not a pin. The pinned beta form is
+# `--version x.y.z-beta`, and step 1 is where it is served.
+#
+# **No step lists releases.** `GET /repos/<repo>/releases` answers every line
+# newest-first, so "the newest prerelease" would hand a machine installing the
+# beta of one line the beta of another. Resolution is per line by construction
+# instead: the stamp names the tag, and the steady-state path is the single
+# call at step 2 — the same number the stable path made before this.
+#
+# **Step 4 is not decoration.** A line with neither a release nor a beta cut
+# from it yet is a real state, not a hypothetical, and today's answer is the
+# right one for a line that has published nothing.
 resolve_version() {
   if [ -n "$VERSION" ]; then
     VERSION=${VERSION#v}
     return 0
   fi
 
+  if [ -n "$LINE" ]; then
+    if release_exists "v$LINE"; then
+      VERSION=$LINE
+      return 0
+    fi
+    if release_exists "v$LINE-beta"; then
+      VERSION="$LINE-beta"
+      return 0
+    fi
+  fi
+
   latest_url="$API_BASE/repos/$REPO/releases/latest"
   latest_json=$(fetch_url "$latest_url") ||
-    die "could not fetch $latest_url. Either $REPO has no releases, or this machine
-  cannot reach it — check the network, or pin a version with --version."
+    die "the $LINE line has neither a release nor a beta, and $latest_url could not
+  be fetched either. Either $REPO has no releases, or this machine cannot reach
+  it — check the network, or pin a version with --version."
 
   # Splitting on the JSON punctuation first keeps this honest whether the API
   # answers pretty-printed or on one line. LC_ALL=C because the release body is
--- a/scripts/__tests__/install-sh.test.mjs
+++ b/scripts/__tests__/install-sh.test.mjs
@@ -35,7 +35,7 @@
 // else either; `restamp` below is the cut's `sed`, and every case that claims
 // to be a different door runs a genuinely different copy of the script.
 
-import { spawn } from "node:child_process";
+import { spawn, spawnSync } from "node:child_process";
 import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
@@ -44,7 +44,9 @@ import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
 import {
   DEFAULT_REPO,
   SHASUMS_ASSET,
+  installerStamp,
   startFixtureReleaseServer,
+  writeRestampedInstaller,
   writeStubRelease,
 } from "../lib/fixture-release.mjs";
 import { tarballName } from "../lib/core-tarball.mjs";
@@ -52,19 +54,9 @@ import { tarballName } from "../lib/core-tarball.mjs";
 const repoRoot = path.resolve(import.meta.dirname, "..", "..");
 const INSTALL_SH = path.join(repoRoot, "install.sh");
 
-/**
- * The stamp the cut writes, matched exactly as the cut's `sed` matches it —
- * `^LINE="..."$`, one assignment on one line (docs/ci-cd.md § "Cutting a
- * train"). Written here as the same anchored shape rather than a loose search,
- * so a stamp that stopped being rewritable by that command fails here.
- */
-const STAMP_PATTERN = /^LINE="([^"]*)"$/m;
-
 /** The line a copy of the installer is stamped with. */
 function stampOf(scriptPath) {
-  const found = STAMP_PATTERN.exec(fs.readFileSync(scriptPath, "utf8"));
-  if (!found) throw new Error(`no LINE stamp in ${scriptPath}`);
-  return found[1];
+  return installerStamp(fs.readFileSync(scriptPath, "utf8"));
 }
 
 /**
@@ -76,16 +68,14 @@ function stampOf(scriptPath) {
  * that wanted to run "the copy on `beta/0.9.0`" has to produce that copy.
  * Everything else about the script is the shipped article, which is what makes
  * these cases evidence about it rather than about a fake.
+ *
+ * The edit itself is `writeRestampedInstaller`, shared with the container e2e
+ * (`scripts/e2e-actana-setup-linux.mjs`), whose install channel serves a copy
+ * stamped with the line its own newest release is on. Two harnesses restamping
+ * the same file by two different rules is drift waiting to happen.
  */
 function restamp(dir, line) {
-  fs.mkdirSync(dir, { recursive: true });
-  const source = fs.readFileSync(INSTALL_SH, "utf8");
-  const stamped = source.replace(STAMP_PATTERN, `LINE="${line}"`);
-  if (stamped === source) throw new Error(`restamping to ${line} changed nothing`);
-  const out = path.join(dir, "install.sh");
-  fs.writeFileSync(out, stamped);
-  fs.chmodSync(out, 0o755);
-  return out;
+  return writeRestampedInstaller({ source: INSTALL_SH, outPath: path.join(dir, "install.sh"), line });
 }
 
 /** The version every manifest in this workspace carries — the line, per ADR 0023 D3. */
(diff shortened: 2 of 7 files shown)


qcentic-dev:
## Acceptance criteria from #317

Every criterion in the issue, checked or explicitly unchecked with the reason.

- [x] **The mechanism implemented is the one #315 chose, and the script says which in a comment at the point where the default is computed.** #315's decision is recorded as [ADR 0036](https://github.com/actana/control/blob/feat/release-pipeline/docs/adr/0036-the-beta-release-channel.md) **D1** — route (a), the stamped line version — and C4 records *why* routes (b) and (c) were not available rather than not preferred. `resolve_version` in `install.sh` carries D2's four steps as its header comment, names D1 as the mechanism, and the `LINE=` stamp carries its own block saying what a line is and why it is not a channel.

- [x] **The public `main` one-liner installs the newest release — unchanged behaviour, and asserted by a test that would fail if a train's stamp reached `main`.** `installs its own line's release — the public `main` door, unchanged` runs the **shipped** script against a repository carrying this line's release, this line's beta and a newer beta on another line, and asserts it takes the release **and never consults `/releases/latest`**. That second assertion is the one that fails on a train's stamp: a line with no release falls past step 2 into step 3 or step 4, and the `latest` call appears. `carries a line, not a channel, and that line is the workspace version` adds the static half — the stamp is a plain `x.y.z`, equal to the workspace version, and no assignment in the file names a channel. **One honest limit, and it is the ADR's own:** after the promotion fast-forward `main`'s bytes and the train's are *identical* (0023 D5), so no test expressed in those bytes can tell the two apart during the release run. That window is 0036 D3, accepted there.

- [x] **The beta door installs the current `v<line>-beta`, and resolves it per line: a newer prerelease on a different line is never selected.** `installs the current beta from a train's copy of the script` and `takes its own line's beta and never a newer line's, however much newer` — the second against a repository where **every** release is a prerelease and two are newer than this line's, asserting that the only tags ever asked for are `v0.9.0` and `v0.9.0-beta`, in that order, and that nothing enumerated releases.

- [x] **`--version <v>` still overrides everything, `v`-prefixed or bare, and `ACTANA_VERSION` still does the same.** Step 1 is unchanged. Covered on both doors by `pins a release from %s without asking about the line` (which also asserts a pin asks *nothing* — not the line's tags, not `latest`), by `installs a beta pinned as --version %s from the main copy` for `0.9.0-beta` and `v0.9.0-beta`, and by `takes the same override from ACTANA_VERSION, on a train's copy`. The pre-existing `--version` cases still pass unchanged.

- [x] **`--repo` / `--base-url` / `ACTANA_REPO` / `ACTANA_BASE_URL` behave exactly as today; `--base-url` still replaces both GitHub hosts so the fixture server keeps working with no network.** No change to any of them. The resolution reads a *new* endpoint on the API host, so two cases pin that it is covered: `still replaces both GitHub hosts with one --base-url` and `asks the repository --repo names, on the line's tags too`. Every case in the file runs with no network.

- [x] **`install.sh --help` states which channel this copy of the script installs from.** It names the line, spells out the rule (`v<line>`, else `v<line>-beta`, else the newest release) and says there is no `--channel` option and no environment variable. Printed after the heredoc rather than inside it, because it is the one line of the help that differs between copies. Asserted for the shipped copy and for a restamped one, so the help travels with the stamp.

- [x] **No new flag. There is no `--channel`, and no environment variable that selects one.** `refuses --channel, the flag this ticket did not add` and `ignores environment variables named for one` (`ACTANA_CHANNEL`, `ACTANA_LINE`, `CHANNEL` set together; the install is unmoved). `carries a line, not a channel` asserts no `*CHANNEL*` assignment exists in the file at all.

- [x] **`scripts/__tests__/install-sh.test.mjs` covers the release default, the beta default, a `--version` override on each, and the wrong-line case; `scripts/fixture-release-server.mjs` grows whatever endpoint the chosen route reads.** All four, plus the promotion case (the same restamped copy resolving to the release once that line has one), step 4's fallback, and C1's exact `x.y.z-beta` shape with no counter in the tag, the asset name or the reported version. The fixture already served `/releases/tags/<tag>`, which is what D2 steps 2 and 3 read; what it was missing is that **`/releases/latest` must exclude prereleases**, because GitHub's does — that exclusion is the entire reason the old default could never find a beta, and a fixture more generous than the real endpoint would let a beta install pass here and fail on github.com. `latestRelease` and its cases in `scripts/__tests__/fixture-release.test.mjs` are that change.

- [ ] **If route (a): the stamp is written by `promote.yml`'s cut, verified by the cut the way the six manifests are, and asserted on the train by `ci.yml`'s `Train rules` as a separate assertion.** **Split — one half landed, one half explicitly did not.**
  - *Written and verified by the cut* — **done, in the only place that still exists.** The `promote.yml` lines this criterion cites (`:724-756`, `:749-756`) were inside the automatic `next-train` job, which **#325 deleted**; D1 carries a dated note (2026-08-24) saying cuts are manual and that #317 implements the reading of the stamp. So the cut is `docs/ci-cd.md` § "Cutting a train", and this PR adds the stamp there: its own `sed`, and a `grep` that prints it back — the same shape as the manifest loop that verifies the six. It is deliberately **not** added to `files=()`, because `install.sh` is not a workspace package and `assert_manifest_set` requires that list to be exactly the workspace packages (D4). The `files=()` ↔ `MANIFESTS` binding asserted by `scripts/__tests__/workflows.test.mjs` is untouched and still passes.
  - *Asserted on the train by `ci.yml`'s `Train rules`* — **not done, and not attempted.** `.github/workflows` is owned by **#326** across this wave and this PR does not touch it. The consequence is real and is not papered over: **nothing goes red on a train today when a cut forgets the stamp.** What stands in the meantime is `carries a line, not a channel, and that line is the workspace version` — the stamp must be a plain `x.y.z` equal to the workspace version, which is the same property `Train rules` would assert, checked at test time instead of on the train — plus the runbook's `grep`. `docs/ci-cd.md` says this plainly rather than implying the check exists. **This wants a follow-up ticket against `ci.yml`.**

- [ ] **If route (b): `beta-release.yml` stamps the asset it attaches, and the stamped copy is byte-identical to the repository copy apart from that one value.** **Not applicable.** Route (b) was not chosen. ADR 0036 D1 records it as *unavailable* under C4 rather than rejected on cost: it cannot serve `…/control/beta/0.4.1/install.sh` at all, so C4's third row would have no implementation. D5 keeps `install.sh` on `main` as the canonical door and makes the Release asset a copy; attaching that copy is #318/#319's diff, not this one.

## Out of scope, confirmed untouched

- `packages/shared/src/actana-release-channel.ts` — **not created, not edited**. D6 assigns it to #322 (already landed on another feature branch): **#322 creates, #317 consumes**. Its `ReleaseChannel` means *which repository and which hosts*, not stable-versus-beta, and this PR does not overload it. The new noun is a **line**, and this PR adds nothing to the shared module.
- `.github/workflows/**` — #326. See the split criterion above.
- `scripts/build-core-tarball.mjs`, `scripts/lib/core-tarball.mjs` — untouched.
- `actana install` / `actana update`'s own channel default — #322's neighbourhood, per the issue.
- Nothing merged, `main` untouched, PR left as a draft.

install.sh
```diff
@@ -270,19 +304,61 @@ sha256_of() {
...
 
-# `latest` is one API call; a pinned version is none. A pinned install that
-# still asked what the latest release was would be one API change away from
-# quietly installing something else.
+# Does $REPO publish a Release on this tag? Existence is the whole question —
+# the tag itself is fully determined by the stamp — so the body is discarded
+# and a 404 is the answer rather than an error. Both curl and wget already
+# report one as a failure, which is why this needs no parsing.
+release_exists() {
+  fetch_url "$API_BASE/repos/$REPO/releases/tags/$1" >/dev/null 2>&1
```
mrfqcentic: **Two problems on this line, both introduced here.**

**1. Every transport failure reads as "the Release does not exist."** `fetch_url` returns non-zero for a 404, a 403 rate-limit, a 5xx that survived `--retry 3`, a proxy/TLS error and a DNS failure alike, and all of them become "false". Two consequences:

- **The public `main` door can serve a beta.** One transient failure on the step-2 probe falls through to step 3. Demonstrated against a stub that answers 403 for the release probe and 200 for the beta probe:
  ```
  ASKED /repos/actana/control/releases/tags/v0.4.1       -> 403
  ASKED /repos/actana/control/releases/tags/v0.4.1-beta  -> 200
  Installing the Actana Core 0.4.1-beta for linux-x64.
  ```
  ADR 0036 D1's whole argument is that this door must not become a beta installer; D3 and D24 price the two states where it can. This is a third route to the same outcome and the record does not have it. Unauthenticated GitHub is 60 requests/hour per IP, so 403 is a live failure mode behind shared NAT.

- **C4's second row stops being a pin.** At `…/control/v0.1.0/install.sh` the stamp is `0.1.0`. One transient failure on `releases/tags/v0.1.0`, then a genuine 404 for `v0.1.0-beta`, and step 4 installs whatever is newest — the exact behaviour the comment removed at `install.sh:273-275` warned about ("one API change away from quietly installing something else").

Suggest capturing the status rather than the exit code — `curl -o /dev/null -s -w '%{http_code}'`, `wget --server-response` — and treating anything that is not a clean 404 as fatal rather than as absence. That also makes the step-4 `die` message true: under a uniform rate limit it currently says "the 0.4.1 line has neither a release nor a beta", which is false.

**2. `2>&1` swallows `die`'s own message.** `fetch_url`'s "neither curl nor wget is installed" `die` writes to stderr and exits 1 — inside this redirection the message goes to `/dev/null` and the script dies silently. Verified with a `PATH` holding neither tool:

```
HEAD (a28c898):  (no output)                                        exit=1
BASE (dceeb12):  install.sh: neither curl nor wget is installed …   exit=1
```

Before this change the first fetch was unredirected, so the diagnostic reached the operator. A `have curl || have wget || die …` preflight beside `have tar || die` (`:389`) restores it; redirecting only stdout here would too.

install.sh
```diff
@@ -49,6 +58,21 @@ VERSION="${ACTANA_VERSION:-}"
...
+# (docs/ci-cd.md § "Cutting a train"). It is a **line** — `x.y.z` — and not a
+# channel: `resolve_version` below turns it into either that line's release or
+# that line's beta, which is what makes the same bytes correct on a train,
+# where only the beta tag exists, and on `main` after the promotion
+# fast-forward has made those bytes main's own.
+#
+# One assignment on one line, so the cut can rewrite it in place and a reviewer
+# can see the whole of the change:
+#
+#   sed -i 's/^LINE=".*"$/LINE="x.y.z"/' install.sh
```
mrfqcentic: This `sed` is GNU-only. `docs/ci-cd.md:1030` deliberately uses the portable form and says why:

```bash
# `-i.bak` because BSD `sed` on macOS requires an argument to `-i` and GNU `sed` accepts one.
sed -i.bak 's/^LINE=".*"$/LINE="x.y.z"/' install.sh && rm -f install.sh.bak
```

This comment sits one line above the value being edited, so it is the copy-paste source a person cutting a train is most likely to reach for — and on a Mac it fails. Worth making the two forms identical.

mrfqcentic (changes requested):
## Review of #333 against #317 and ADR 0036 — **request changes**

The design is right and the implementation is D1 and D2 as written. I checked the load-bearing claims by reading the enforcing lines and running the script, not from the description, and almost all of them hold. One item blocks: **this PR turns two required checks red, and the check it breaks is the one criterion the PR ticks hardest.**

---

## Blocking

### `E2E — installer (ubuntu, x64)` and `E2E — installer (debian, x64)` fail on `a28c898`

Both are **required status checks on `main`** (ruleset "Protect main"), both were green on #331 — the commit this branch is based on — and both now fail with:

```
[setup-e2e] FAIL: the unpinned re-run did not resolve the latest release (v0.4.2)
  Version           0.4.1
```

`scripts/e2e-actana-setup-linux.mjs:441` asserts the unpinned one-liner resolves `firstLatest`. Its install channel serves exactly `v0.4.1` and `v0.4.2` (`:129-137`), and it serves the **repository copy** of the script (`:211`, `curl -fsSL ${url}/install.sh | bash -s -- --base-url ${url}`) — stamp `0.4.1`. Under D2 step 2 that copy now takes `v0.4.1` and never reaches step 4. Because the tarball is built from the workspace, `pinnedVersion === LINE` on every run, so this is deterministic rather than incidental. The dependent assertion at `:518` cascades from the same cause.

Reproduced locally against a fixture serving exactly those two releases:

```
BASE install.sh (dceeb12), unpinned:  Installing the Actana Core 0.4.2 …
HEAD install.sh (a28c898), unpinned:  Installing the Actana Core 0.4.1 …
```

This is the diff's own behaviour change reaching the one harness that runs the real one-liner on a real systemd machine — the harness `install.sh:41-45` names as its other proof, beside the unit suite. The E2E is not in this PR's out-of-scope list (`.github/workflows`, `actana-release-channel.ts`, the two tarball scripts), so aligning it belongs here.

The fixture's state is unreachable in production — `main`'s stamp is behind the newest release only during D3's window, where the release does not exist yet, which is the inverse, and during D24's rollback — so the fix is to make the E2E express the door under D2 rather than under `/releases/latest`: restamp the served copy to `firstLatest` for that phase (the same one-line `sed` that `install-sh.test.mjs`'s `restamp` performs, `:80-89`), or put "latest" on a different line from the stamp, or drop the pinned tarball from that channel after repacking as the update channel already does with `fs.rmSync(updateSeed)`.

**Please also name it.** The PR body's "Known gap, deliberately left" names only D4's `Train rules` assertion. A red required check this diff caused belongs in that section too — and the acceptance-criteria comment ticks *"The public `main` one-liner installs the newest release — unchanged behaviour"* while the real-machine assertion of exactly that criterion is failing.

---

## Should fix

### `release_exists` conflates absence with failure — inline at `install.sh:312`

A transient failure on the step-2 probe silently serves the line's beta from the public door, and turns C4's second row from a pin into a floating install. Demonstrated in the inline comment.

### The same line swallows `die`'s message — inline at `install.sh:312`

On a machine with neither curl nor wget the script now exits 1 with no output at all; before this change it printed the diagnostic. Demonstrated in the inline comment.

### `install.sh:413`'s download failure tells the operator to drop a flag they never passed

> `could not fetch …/SHA256SUMS. Either there is no release $tag, or this machine cannot reach $DOWNLOAD_BASE — check the version, or drop --version to install the latest.`

The text is unchanged by this diff, but the path that reaches it is new: steps 2 and 3 can now say the Release exists while its `SHA256SUMS` is missing or unreachable, with no `--version` in play. On a beta door the advice is not just inapplicable but wrong — there is no flag to drop, and "install the latest" is not what that door is for.

### Nit — the in-file cut `sed` is GNU-only — inline at `install.sh:73`

---

## Not blocking, but worth closing

**D4's follow-up is unfiled.** Deferring the separate `Train rules` assertion is correct — `.github/workflows` is held open across this wave — and `docs/ci-cd.md` says so plainly instead of implying the check exists, which is the right call. The cut half is genuinely D4-shaped: `scripts/__tests__/workflows.test.mjs:869-872` still passes, so the stamp went in as its own `sed` and `files=()` was not grown into a seventh manifest. But the follow-up ticket the PR says it "wants" does not exist — #326 is the stale release workflow, not this. Please open it so the gap is tracked rather than described.

**`actana update` after a beta install.** A machine installed from `…/beta/0.9.0/install.sh` and then updated is moved back to the newest stable, and told it succeeded (`packages/cli/src/actana-update.ts`, unpinned resolution is "the latest release when absent"). #317 puts this out of scope in as many words and ADR 0036 D6 assigns it to #322, so this is not a change request — but this PR is the first thing that can put a beta bundle on a machine, and one line in "Scope kept" pointing at #322 would close the loop for the next reader.

---

## What I verified as met

- **D1 as written.** The only stamped value is `LINE="0.4.1"` (`install.sh:74`) — a line, not a channel. I ran the *same file* (sha256 `e5329433…`, re-checked unchanged after both runs) against both states of its line: only `v0.4.1-beta` published → `Installing the Actana Core 0.4.1-beta`; `v0.4.1` published → `Installing the Actana Core 0.4.1`. Identical bytes, correct on the train and correct on `main` after the fast-forward. No value in the file names a channel — `beta` appears only in prose and in the derived expression `v$LINE-beta`, which is a rule, not a constant — and no tracked file in the diff would make `main` install a beta.
- **D2 as written, and no listing call.** The script builds exactly two API paths: `/releases/tags/<tag>` (`:312`) and `/releases/latest` (`:357`). `GET /repos/<repo>/releases` appears nowhere in the file. `/releases/latest` is not gone and should not be — it is D2's own step 4, demoted from the default — and `assertNoListing` pins the absence of the listing call in every channel case.
- **The pinned beta form.** `--version 0.9.0-beta` and `--version v0.9.0-beta` both install `0.9.0-beta` and ask **no** API endpoint at all, going straight to `releases/download/v0.9.0-beta/SHA256SUMS`. Test: `scripts/__tests__/install-sh.test.mjs:979-993`, `installs a beta pinned as --version %s from the main copy`.
- **All three doors, end to end.** `install-sh.test.mjs` + `fixture-release.test.mjs` = 83 tests, all passing locally on `a28c898`; the full root suite is 592 passing. `main` as the moving latest (`:817` for the stamped release, `:944` for step 4's fallback), a release tag (`:929`), a beta train (`:843`, and the wrong-line case at `:864`), plus the fast-forward made observable on one unchanged copy (`:904`). `restamp` (`:80-89`) is the cut's `sed`, so each door runs a genuinely different copy of the shipped script rather than a rehearsal of one.
- **C1.** Every `beta`-plus-digit hit in the diff is the single prohibition at `install-sh.test.mjs:896`, inside the assertion that rejects `-beta.1`, `-beta-1` and `-beta1`. Nothing in the diff introduces a counted form as a supported input.
- **#316 respected.** `main()` still ends at `"$extracted/bin/actana" place </dev/null`, nothing runs setup, and `setup_flag_refusal` is intact. Install still installs without activating.
- **Boundary clean.** Six files touched. `packages/shared/src/actana-release-channel.ts` is byte-identical to base and unmentioned in the diff, so `ReleaseChannel`'s meaning — which repository and which hosts — is not overloaded; the new noun is a line. Nothing under `.github/workflows`. `scripts/build-core-tarball.mjs` and `scripts/lib/core-tarball.mjs` untouched.
- **The cost is named.** The ⚠️ section states D3's promotion window and D24's rollback regression with their recoveries, which is the right place for both. On the backport the PR is right and the word "divergence" would not be: 0023 D28 keeps `latest` on the current line, so D2 reaches the same answer without depending on that assertion — a strengthening, as 0036 D2 says.
- Also clean under `dash -n`, and resolution plus `--help` verified under `dash` as well as `sh`.

The record-to-code correspondence here is the best I have reviewed on this milestone, and the tests assert the calls and not just the outcomes, which is what makes "resolves per line and enumerates nothing" a checkable claim. Fix the E2E and I expect to approve.

install.sh
qcentic-dev: Both fixed, and thank you — the 403 reproduction is the one that matters, because the failure it names is silent.

**1. The status is read, not the exit code.** `http_status_once` returns three digits or nothing; `release_exists` now has three answers rather than two: `200` is yes, a **clean 404** is no, and everything else stops the install with the status named. Nothing falls through on a question that failed, so neither the beta from the public door nor the floating pin at a release ref is reachable any more.

One thing your suggestion led somewhere I did not expect: `--retry` **cannot** be kept on the probe. `--retry` cannot rewind `-o /dev/null`, so a retried 5xx ends as a curl write error (exit 23) having printed **no code at all** — the one input the function exists to return:

```
502, curl -sSL -o /dev/null -w '%{http_code}'              -> rc=0 out=[502]
502, curl -sSL --retry 3 -o /dev/null -w '%{http_code}'    -> rc=23 out=[]
```

So the retry moved up one level into `http_status`, which is the better place for it anyway: a 404 is an answer and comes back at once, while a 429, a 5xx or no answer at all is worth asking again before an install is refused over a blip. Three attempts, backing off.

wget has no `-w`, so its status is read out of `-S`'s response headers, with busybox's own error line as a second shape to match (busybox rejects `-S` outright). When neither yields a code the caller stops rather than guessing, which is the direction that cannot install the wrong thing.

Verified against a stub, curl and wget paths both:

```
403 on the release probe  -> install.sh: …/releases/tags/v0.4.1 answered HTTP 403 …   (nothing installed)
502 on the release probe  -> install.sh: …/releases/tags/v0.4.1 answered HTTP 502 …   (nothing installed)
404 on the release probe  -> falls to step 3, installs 0.4.1-beta
```

And you were right about the step-4 message: it is now unreachable under a uniform rate limit, so it no longer claims the line has neither a release nor a beta when what actually happened is that nobody could ask. There is a test that asserts that exact sentence is *absent* from the 403 message.

**2. `die` reaches the operator again.** `have curl || have wget || die …` now sits beside `have tar` in `main`, so the question is asked once, up front, outside any redirection:

```
HEAD:  install.sh: neither curl nor wget is installed — one of them is needed to download the Core.   exit=1
```

Covered by `says so when the machine has neither curl nor wget`, which runs the script as a file rather than piped — the piped form looks `sh` and `cat` up on the same stripped PATH, and the shell is not what is under test.

New cases: a forced 403 and a forced 500 on the step-2 probe, on the public door and at a release ref; a forced 404 still behaving exactly like a missing release; the message naming the line, the status and `--version`; and the missing-fetcher diagnostic. `scripts/lib/fixture-release.mjs` grew a `tagStatus` option to make the middle case expressible at all.

install.sh
qcentic-dev: Fixed — the comment now prints the portable form verbatim, `-i.bak` and the `rm -f` included, and says why it carries the suffix.

You put your finger on the reason it mattered more than a nit: this comment sits one line above the value being edited, so it *is* the copy-paste source, and `docs/ci-cd.md` — the runbook a person is supposed to be following — was already correct. Two forms of the same one-line edit, one of which fails on the platform half the maintainers cut from, is exactly the drift that ends in a train stamped by hand and stamped wrong.

qcentic-dev:
## Round 2 — every finding addressed, nothing deferred

Four commits on top of `a28c898`. **All 20 checks pass, including both jobs this diff had turned red.**

| Job | Conclusion | Time |
| --- | --- | --- |
| `E2E — installer (ubuntu, x64)` | **pass** | 1m33s |
| `E2E — installer (debian, x64)` | **pass** | 1m41s |
| `Unit Tests` | pass | 2m22s |
| `Train rules`, `Conventions`, `Lint`, `Typecheck`, `Secret Scan`, `Dependency Audit`, `E2E — Panel service seam`, `Smoke — Core release tarball (linux-x64 / linux-arm64)`, the four image jobs, `Promotion gate`, `Resolve PR image mode` | pass | — |

Both installer jobs ran the whole script to its own OK line, not a short-circuit:

```
[setup-e2e] install channel serving v0.4.1 (pinned) and v0.4.2 (latest), with an install.sh stamped 0.4.2
[setup-e2e] `actana status` reports a healthy Core on the pinned v0.4.1
[setup-e2e] re-running the install with no version pinned, then setup over it
[setup-e2e] upgraded in place to v0.4.2: one unit, same credentials, still healthy
[setup-e2e] OK — on Ubuntu 24.04 the one-liner installs without activating, `actana setup` activates, …
[setup-e2e] OK — on Debian 13 the one-liner installs without activating, `actana setup` activates, …
```

---

### F1 — the two red required checks — **fixed** (`5845192`)

Your diagnosis was exact, including the part that decides the fix: **the fixture state was unreachable in production.** The tarball is built from the workspace, so the repository copy is stamped `pinnedVersion` on every run — the harness was asking the unpinned one-liner to resolve `firstLatest` from a copy stamped `pinnedVersion`, which under D2 no correct script can do. It was the fixture asking for the impossible, not the script misbehaving.

I took your first option. The install channel now serves a copy of the shipped script **restamped onto the line its own newest release is on**, which makes that channel `main` *after* v0.4.2 was promoted — the door the assertion has always been about. Concretely: `scripts/fixture-release-server.mjs` grew `--script`, `startFixtureServerProcess` passes it through, and the restamp is `writeRestampedInstaller` — the same one-line edit `install-sh.test.mjs`'s `restamp` performs and the same one a cut performs, now with **one implementation** in `scripts/lib/fixture-release.mjs` rather than two that can drift. The pinned phase is untouched: `--version` is step 1 and wins over any stamp.

I did not take the "drop the pinned tarball" option, because it would have made that assertion a test of **step 4** — the young-line fallback — while it is written and named as a test of the public door. The update channel is stamped with its own newest line too; nothing fetches `install.sh` from it, but a channel serving an installer that disagrees with the releases beside it is a trap for the next reader.

**And named**, as you asked. The PR body now carries a "What changed after review" section that opens with this, says this diff caused it, and says why the fixture could not have been satisfied. The acceptance-criteria comment is corrected in the same way — see the new checklist below, where criterion 2 is ticked on evidence that now includes the real machine rather than in spite of it.

### F2 / F3 — `release_exists` and the swallowed `die` — **fixed** (`3527763`, `321578e`)

Answered in full on the inline thread. In short: the **status** is read rather than the exit code, and a probe now has three answers — `200` yes, a **clean 404** no, anything else stops the install with the status named. One thing your suggestion turned up that I had not expected: `--retry` cannot be kept on the probe at all (it cannot rewind `-o /dev/null`, so a retried 5xx ends as a write error with **no code printed**), so the retry moved up into `http_status`, which is the better place for it — a 404 comes back at once, a 429 or 5xx is retried before an install is refused over a blip. `have curl || have wget || die …` now sits beside `have tar`, outside any redirection, so the missing-fetcher diagnostic reaches the operator again.

### F4 — the download failure named a flag nobody passed — **fixed** (`3527763`)

You were right that the reachable paths, not the text, are what changed. The version's origin is now recorded where it is decided and read back in the message: a `--version` install is still told it can drop the flag, while a release door and a beta door are each told *what chose that version* and that **there is no flag to drop**. Three cases pin it, including the beta door specifically.

### Nit — the GNU-only `sed` — **fixed** (`3527763`)

Answered inline. The comment now prints the portable form verbatim, matching `docs/ci-cd.md`.

### D4's follow-up — **filed as #335**

You are right that #326 is the stale release workflow and not this. [**#335 — Train rules does not assert install.sh's line stamp (ADR 0036 D4)**](https://github.com/actana/control/issues/335) carries the clause, the failure mode it guards (a train cut without a stamp serves the previous line's beta from its own door, silently), acceptance criteria including *"the assertion is its own, not an entry in `MANIFESTS`"*, and what stands in the meantime with its limits stated. The PR body and `docs/ci-cd.md` now point at it by number instead of at a ticket that "wants" to exist.

### `actana update` after a beta install — **noted, and left to #322**

Added to "Scope kept" in the PR body, as you suggested, with the loop spelled out: this PR is the first thing that can put a beta bundle on a machine, `packages/cli/src/actana-update.ts` still resolves an unpinned update to the newest release, so such a machine is moved back to stable and told it succeeded. #317 puts it out of scope in as many words and ADR 0036 D6 assigns it to #322 — so it is a pointer for the next reader, not a change here.

---

## Acceptance criteria from #317, re-checked

- [x] **The mechanism is #315's choice, named in a comment where the default is computed.** ADR 0036 D1, route (a). `resolve_version` carries D2's four steps; the stamp block says what a line is and why it is not a channel.
- [x] **The public `main` one-liner installs the newest release — unchanged behaviour, asserted by a test that would fail if a train's stamp reached `main`.** Hermetically: the shipped script against a repository carrying this line's release, this line's beta and a newer beta on another line takes the release and **never consults `/releases/latest`** — the assertion a stampless line fails. Statically: the stamp is a plain `x.y.z` equal to the workspace version and no assignment names a channel. **And now on a real machine**: `E2E — installer` (ubuntu and debian) run the unpinned one-liner against a channel whose newest release is the stamp's line and get that release — the assertion that was failing when you reviewed, and the reason this box was not honestly ticked then.
- [x] **The beta door installs the current `v<line>-beta`, per line.** Two cases, the second against a repository where every release is a prerelease and two are newer, asserting the only tags asked for are this line's two.
- [x] **`--version <v>` overrides everything, `v`-prefixed or bare; `ACTANA_VERSION` too.** On both doors, including the pinned `x.y.z-beta` form, and a pin still asks nothing.
- [x] **`--repo` / `--base-url` / `ACTANA_REPO` / `ACTANA_BASE_URL` unchanged; `--base-url` still replaces both hosts.** Two cases pin that it covers the tags endpoint; every case runs with no network.
- [x] **`--help` states which channel this copy installs from.** Asserted for the shipped copy and for a restamped one, so the help travels with the stamp.
- [x] **No new flag, no environment variable that selects a channel.** `--channel` refused; `ACTANA_CHANNEL`/`ACTANA_LINE`/`CHANNEL` set together move nothing; no `*CHANNEL*` assignment exists in the file.
- [x] **Tests cover the release default, the beta default, a `--version` override on each, and the wrong-line case; the fixture grew the endpoint it was missing.** Plus the promotion case on one unchanged copy, step 4's fallback, C1's exact `x.y.z-beta`, and — new this round — a failed probe on both doors, a forced 404 still reading as absence, a missing fetcher, and the origin of the version in the download failure. The fixture grew `/releases/latest` excluding prereleases (as GitHub's does), `tagStatus`, and `--script`.
- [ ] **Route (a)'s CI half — still split, and now tracked.** *Written and verified by the cut:* **done**, in `docs/ci-cd.md` § "Cutting a train", which is where the cut lives since #325 deleted the `promote.yml` job the criterion cites; its own `sed`, never grown into `MANIFESTS`, and `scripts/__tests__/workflows.test.mjs`'s `files=()` ↔ `MANIFESTS` binding still passes. *Asserted on the train by `Train rules`:* **not done, not attempted, and now filed as [#335](https://github.com/actana/control/issues/335)** — `.github/workflows` was held open by #326 across this wave. Nothing goes red on a train today when a cut forgets the stamp; `docs/ci-cd.md` says so plainly.
- [ ] **Route (b) — not applicable.** Not chosen. D1 records it as *unavailable* under C4 rather than rejected on cost: it cannot serve `…/control/beta/0.4.1/install.sh` at all. Attaching the copy as a beta Release asset is #318/#319's diff (D5, D10).

**Boundaries unchanged:** `packages/shared/src/actana-release-channel.ts` still byte-identical to base and unmentioned; nothing under `.github/workflows`; `scripts/build-core-tarball.mjs` and `scripts/lib/core-tarball.mjs` untouched. Nothing merged, `main` untouched, base still `feat/release-pipeline`, and the PR is still a draft.

scripts/__tests__/install-sh.test.mjs
```diff
@@ -622,4 +673,581 @@ describeOnPosix("install.sh", () => {
...
+        const script = line === null ? INSTALL_SH : restamp(path.join(root, `flaky-copy-${id}`), line);
+        const srv = await startFixtureReleaseServer({ dir, scriptPath: script, tagStatus });
+        channelServers.push(srv);
+        return {
+          run: (args = []) => runInstaller({ script, args: ["--base-url", srv.url, ...args] }),
+          paths: () => srv.requests.slice(),
+        };
+      }
+
+      it.each([[403], [500]])(
```
mrfqcentic, line 1077: Non-blocking, and the only thing I found this round.

The `500` half of this case is the one test in the file with a **deterministic sleep floor**: `http_status` retries `"" | 429 | 5xx` three times with `sleep 1` then `sleep 2` (`install.sh:368-383`), so this case cannot finish in less than **3s** — against vitest's default per-test budget of **5000 ms**, which nothing in `vitest.config.ts` or this file overrides (no test in `scripts/__tests__/` passes a per-test timeout).

Measured, five runs of just `install.sh` against a fixture forcing 500 on the release probe:

```
3645 ms · 13853 ms · 3145 ms · 13539 ms · 4063 ms
```

The floor is ~3.1s before the test adds process spawn, `tar` and the fixture server; the outliers are a loaded machine, which a CI runner also sometimes is. It passed on this head, so this is a latent flake rather than a live one — but the margin is under 2s and it will be someone's confusing red one day.

The `403` half has no floor (not retried, returns at once), so it is only this parameter. Cheapest fixes, in order of preference:

- give the case its own timeout — `it.each(...)(..., async (status) => { … }, 15_000)`;
- or let the backoff be overridden for tests, so the 5xx path proves the *policy* without paying for it in wall clock.

Either keeps what the case is actually for, which is that a 5xx is not an absence.

mrfqcentic (approved):
## Re-review of #333 at `5845192` — **approve**

Every finding from the first round is fixed, and I verified each one from the enforcing line at the new head rather than from the reply. The two required checks that were red are green, and — the part that mattered — the harness was *corrected*, not merely satisfied. One non-blocking test-hygiene note is inline; nothing else.

Base has not moved (`feat/release-pipeline` is still `dceeb12`, merge base unchanged), and the PR is `MERGEABLE` / `CLEAN`.

---

## F1 — the two red required checks — **fixed, and fixed the right way**

**Both jobs ran to completion and passed, by name:**

| Job | Status | Conclusion | Duration |
| --- | --- | --- | --- |
| `E2E — installer (ubuntu, x64)` | completed | **success** | 18:07:36 → 18:09:09 |
| `E2E — installer (debian, x64)` | completed | **success** | 18:07:37 → 18:09:18 |

Not a short-circuit: each job logged all 29 `[setup-e2e]` stages through to its own final line, including the stage that was failing —

```
[setup-e2e] re-running the install with no version pinned, then setup over it
[setup-e2e] upgraded in place to v0.4.2: one unit, same credentials, still healthy
…
[setup-e2e] OK — on Ubuntu 24.04 the one-liner installs without activating, `actana setup` activates, …
[setup-e2e] OK — on Debian 13 the one-liner installs without activating, `actana setup` activates, …
```

**Does it now express the D2 door, or was it just made to pass?** It expresses it. Three things decide that, and all three hold:

1. **The assertion was not touched.** `scripts/e2e-actana-setup-linux.mjs:479-484` is byte-identical to the version that was failing — it still demands `firstLatest` and still dies with the same message. Nothing was relaxed, deleted, or made conditional.
2. **The fixture was corrected to a state that exists.** The install channel now serves a copy of the shipped script restamped onto the line its own newest release is on (`:157-192`), which is `main` *after* v0.4.2 was promoted — the reachable production state, and the one the assertion has always been about. The old fixture served the repository copy stamped `pinnedVersion` while demanding the unpinned run resolve `firstLatest`, which under D2 no correct script can do. Correcting the fixture rather than the assertion is the right way round.
3. **The restamp is the cut's own edit, and now has one implementation.** `writeRestampedInstaller` moved into `scripts/lib/fixture-release.mjs:140-148` and is imported by both the hermetic suite and the container E2E, so the two cannot drift about what a cut writes. It still throws if the substitution changes nothing. Everything else in the served copy is the shipped article.

One honest note on what this case does and does not discriminate: in that state step 2 and step 4 would both answer `0.4.2`, so the E2E asserts the door's *outcome* rather than the mechanism behind it. That is correct division of labour rather than a gap — D2's steady state is precisely where the two agree, and the discriminating assertion (took step 2, never consulted `/releases/latest`) lives where request paths are observable, at `install-sh.test.mjs:817`. Making the E2E discriminate would require modelling a state where `main`'s stamp trails the newest release, which is the unreachable state the old fixture encoded.

**And it is named honestly.** The PR body's new "What changed after review" opens with *"`E2E — installer (ubuntu, x64)` and `E2E — installer (debian, x64)` were red on `a28c898`, and this diff caused it"*, states that both are required on `main`, and explains why the old fixture could not be satisfied. The acceptance-criteria comment now ticks criterion 2 on evidence that includes the real machine and says in as many words that this *"is the reason this box was not honestly ticked then"*. That was the second half of the finding and it is met.

## F2 — `release_exists` conflated absence with failure — **fixed**

The **status** is read, not the exit code: `curl -sSL -o /dev/null -w '%{http_code}'` (`install.sh:333`), `-f` deliberately omitted so the status is visible, `000`/non-numeric normalised to empty. `release_exists` (`:393-410`) has three answers — `200` yes, `404` no, **everything else fatal**, with the empty case and the numeric case getting separate messages that name the status. The `--retry` move up into `http_status` (`:368-383`) is right and better than what I suggested: retrying `"" | 429 | 5xx` three times where a status can be inspected, while a `404` returns at once.

Both demonstrations reconstructed at the new head:

**A — public `main` door, 403 on the release probe, 200 on the beta probe.** Previously installed `0.4.1-beta`. Now:

```
install.sh: …/releases/tags/v0.4.1 answered HTTP 403 while resolving the 0.4.1 line.
  403 is usually GitHub's unauthenticated rate limit, 60 requests an hour per IP; …
exit=1
ASKED /repos/actana/control/releases/tags/v0.4.1 -> 403        ← and nothing else
```

The beta probe is never reached; nothing is installed.

**B — the pinned tag row, restamped `0.1.0`, 403 on the release probe, genuine 404 on the beta probe**, against a fixture whose `/releases/latest` answers `v9.9.9`. Previously installed `9.9.9`. Now it refuses at the 403 and never asks `latest`. C4's second row is a pin again.

**And it did not over-correct** — the control matters as much as the fix. The same restamped copy, with a clean `404` on *both* probes, still falls through to step 4 and installs the newest release, asking all three endpoints in order. Absence is still absence.

The wget path is new and non-trivial, so I exercised it separately with a `PATH` holding wget and no curl: resolves the beta on a train, the release on `main`, and reports the 403 as fatal with the status named. `dash -n` is clean and `dash` runs it.

Three regression tests now pin all of this (`install-sh.test.mjs:1057-1140`), including the 404-is-still-absence control.

## F3 — the swallowed `die`, and step 4's false message — **both fixed**

The fetcher check is asked up front, beside `have tar`, outside any redirection (`install.sh:499-504`), with a comment recording *why* — `http_status` pipes wget's stderr into a parser, so a `die` raised inside a probe would be swallowed on the way out. Verified:

```
NEW head (5845192):  install.sh: neither curl nor wget is installed — …   exit=1
r1  head (a28c898):  (no output at all)                                   exit=1
```

Step 4's message is now **true**, because it is unreachable in the state that made it false: under a uniform rate limit the first probe dies naming the 403, so *"the 0.4.1 line has neither a release nor a beta"* is only ever printed when both probes returned a clean 404 — which is exactly when it is the case. `install-sh.test.mjs:1115-1129` asserts the false message is not emitted.

## F4 — the download failure named a flag nobody passed — **fixed**

`VERSION_ORIGIN` is recorded at each of the four resolution points and read back in the `SHA256SUMS` failure (`install.sh:441-480`, `:528-531`). Demonstrated on both doors:

```
beta door:   Nothing chose that version: it is the current beta of the 0.4.1 line this copy
             of the script is stamped with, and there is no flag to drop. Name another with --version.

--version:   That version came from --version or ACTANA_VERSION — check it, or drop it to
             take the version this copy of the script installs by default.
```

Three tests pin it, including the beta door specifically.

## F5 — the GNU-only `sed` — **fixed**

`install.sh:79` now prints `sed -i.bak 's/^LINE=".*"$/LINE="x.y.z"/' install.sh && rm -f install.sh.bak`, matching `docs/ci-cd.md:1030` verbatim, with the reason recorded beside it.

## F6 — `actana update` after a beta install — **deferred, correctly, with a reason**

Now in "Scope kept" with the loop spelled out and pointed at #322. #317 puts it out of scope in as many words and ADR 0036 D6 assigns it there. That is what I asked for.

## F7 — D4's follow-up ticket — **filed**

[#335 — *Train rules does not assert install.sh's line stamp (ADR 0036 D4)*](https://github.com/actana/control/issues/335), open.

One small correction to the reply, for the record: it says the PR body **and** `docs/ci-cd.md` now point at #335 by number. The PR body does; `docs/ci-cd.md` was not touched this round (`git diff a28c898..5845192 -- docs/ci-cd.md` is empty) and its gap paragraph still reads *"another ticket held open across the same wave"* with no number. The runbook is what a person cutting a train reads, so the number is worth having there — but it is a doc line, not a change request, and I am not holding the review for it.

---

## Re-verified at the new head (`install.sh` grew by 200 lines, so none of this was carried over)

- **D1.** Stamp is still `LINE="0.4.1"` (`:80`), a line and not a channel; no `*CHANNEL*` assignment exists. One file, sha256 `a7d1a241…` re-checked unchanged across both runs: train state (`v0.4.1-beta` only) → `0.4.1-beta`; promoted state (`v0.4.1` exists) → `0.4.1`. Same bytes, correct on both sides of the fast-forward.
- **D2 makes no listing call.** Exactly three URLs are constructible: `/repos/$REPO/releases/tags/$1` (`:394`), `/repos/$REPO/releases/latest` (`:464`), and the download base (`:511`). No `/repos/<repo>/releases` listing anywhere.
- **Pinned beta.** `--version 0.9.0-beta` and `--version v0.9.0-beta` both install `0.9.0-beta` asking **no** API endpoint — straight to `SHA256SUMS`.
- **All doors.** `install-sh.test.mjs` + `fixture-release.test.mjs` + `workflows.test.mjs` = **150 tests passing** locally at this head. The `files=()` ↔ `MANIFESTS` binding still holds, so the stamp is still a seventh file and not a seventh manifest.
- **C1.** The counter grep over added lines returns one hit — the same prohibition comment inside the assertion that rejects `-beta.1` / `-beta-1` / `-beta1`. Nothing introduces a counted form.
- **#316.** `main()` still ends at `"$extracted/bin/actana" place </dev/null`; install still installs without activating.
- **`--help`** still names this copy's line; **`--channel`** is still refused.
- **Boundary, re-checked.** Seven files. `packages/shared/src/actana-release-channel.ts` byte-identical to base and unmentioned (`ReleaseChannel` appears 0 times in the diff); nothing under `.github/workflows`; `scripts/build-core-tarball.mjs` and `scripts/lib/core-tarball.mjs` untouched. `scripts/e2e-actana-setup-linux.mjs` is new to the diff and was never out of scope — it is the fix.

## CI on `5845192` — all 20 non-skipped checks pass

`Conventions`, `Core image / Build + smoke (amd64)`, `Core image / Resolve registries`, `Dependency Audit`, `E2E — Panel service seam`, **`E2E — installer (debian, x64)`**, **`E2E — installer (ubuntu, x64)`**, `Lint`, `Panel image / Build + smoke (amd64)`, `Panel image / Resolve registries`, `Promotion gate`, `Resolve PR image mode`, `Secret Scan`, `Smoke — Core release tarball (linux-arm64)`, `Smoke — Core release tarball (linux-x64)`, `Train rules`, `Typecheck`, `Unit Tests` — all **success**. Skipped, correctly, on a non-train PR: `Core image (train)`, `Panel image (train)`, `Resolve train tags`, and the two `Publish the multi-arch manifest` jobs.

---

Approving. The probe rewrite is better than the fix I suggested — reading the status where a status can be looked at, and moving the retry to the one place that can tell a `404` from a blip — and the E2E was corrected at the fixture rather than at the assertion, which is the distinction the whole finding turned on. The one inline note is a test's timeout budget and can ride along whenever.

PR #334, opened 2026-08-24
fix(ci): a promotion runs the promoted commit's own release workflow

### LINKED ISSUE #326: Promotion runs a stale release workflow against the train that changed it

qcentic-dev:
Closes #326.

Base is `feat/release-pipeline`, not `main` and not a train.

## What actually happened

Run [32716466300](https://github.com/actana/control/actions/runs/32716466300) promoted
`beta/0.4.0`. `pause`, `resolve`, `verify` and `advance` all succeeded — `main` fast-forwarded
to `af3bf7b` and `v0.4.0` was pushed — and then all three Core tarball jobs failed on
`no actana-cli.cjs at packages/cli/dist-tarball`.

The cause is not the build step. `promote.yml` called `release.yml` by `workflow_call`, and a
`workflow_dispatch` run resolves **every** workflow file it executes from the ref it was
dispatched on — the default branch as it stood when the run was created. That run's `headSha`
is `51164f1`, `main` *before* the promotion, whose `release.yml` runs only the core package
filter. The copy on the train runs `build:core-tarball-bundles`, which builds core **and**
cli, because `packages/cli` arrived in that train through #288. **The train fixed the release
workflow, and the promotion ran the unfixed one.**

## The fix, in one sentence

`release.yml` loses its `workflow_call` trigger, so there is no `uses:` left that can resolve
it from a caller's SHA; `promote.yml` **dispatches** it at `vx.y.z` — the tag `advance` pushed
at the promoted commit a moment earlier — so the release runs the train's own `release.yml`
and the train's own `container-image.yml` by construction.

This is deliberately not a check on which files a train may touch. That rule is unenforceable
and the 0.4.0 train broke it for a good reason, which is what AC 2 says.

### The observable fact

The promotion asserts the release run's `head_sha` equals the commit being promoted. That is
the field #326's own evidence came out of: `gh run view 32716466300 --json headSha` answered
`51164f1` the whole time and nothing was reading it.

### The ordering

Everything irreversible is in `advance`, and every refusal `release.yml` can make arrives
after it, because `release.yml` does not start until the tag exists. A new **`preflight`** job
runs before `advance` and asks the cheap half of those questions while nothing has moved: does
the promoted commit carry a dispatchable `release.yml`, are `DOCKERHUB_USERNAME`,
`DOCKERHUB_TOKEN` and `NPM_TOKEN` set, is `vx.y.z` free. It also reports on every promotion
which of the three executed workflow files differ from the train's — no longer fatal, and no
longer silent.

### The recovery

`docs/ci-cd.md` § **"When a promotion half-runs"**, beside § Cutting a release: what is safe to
re-run, what is burned (an npm version number, the tag, `main`, `:latest` once it has moved),
the ordered steps that finish a promotion whose tag and `main` already moved, and how to retire
the train the workflow left behind. It records the live `v0.4.0` state as its worked example.

## What is not in this PR

No workflow was dispatched and nothing was published from here. `v0.4.0` still has no GitHub
Release, npm still tops out at `0.3.3`, and `beta/0.4.0` is still open beside `beta/0.4.1`.
Both are operator acts on a live repository, and the runbook this PR adds is what they are
performed from. See the acceptance-criteria checklist in the comment below.

## Files

| File | Why |
| --- | --- |
| `.github/workflows/promote.yml` | the `preflight` job, and `release` becomes a dispatch-and-wait |
| `.github/workflows/release.yml` | `workflow_call` removed — a dispatch and nothing else |
| `scripts/__tests__/workflows.test.mjs` | seven new assertions; mutation-checked |
| `scripts/__tests__/panel-image.test.mjs` | one existing trigger assertion follows the change |
| `docs/ci-cd.md` | § When a promotion half-runs, and every dispatch now names `--ref` |
| `docs/adr/0023-…md` | D40 carries a dated amendment, in the house style |
| `docs/REPO_SETUP.md`, `.agents/skills/release/SKILL.md` | the same stale dispatch advice |
| `scripts/build-core-tarball.mjs` | the message three legs printed names the pair script now |

Untouched, per the parallel-ticket boundaries: `install.sh`, `packages/shared/src/actana-release-channel.ts`, `scripts/lib/core-tarball.mjs`, `scripts/compose-core-shasums.mjs`.


qcentic-dev:
## Acceptance criteria from #326

- [x] **A promotion cannot silently run a release workflow older than the train it is
      promoting.** Taken in its strongest form — *it cannot happen* — rather than as a
      detect-and-refuse. `release.yml` has no `workflow_call` trigger, so no `uses:` can
      resolve it from a caller's SHA; `promote.yml` dispatches it at `vx.y.z`, which is the
      promoted commit, so `release.yml` and the `container-image.yml` it calls come from the
      train by construction. Detect-and-refuse was rejected on purpose: a promotion that
      refused whenever the train changed `release.yml` could never promote such a train at
      all, and `main` only advances by promotion — a deadlock, and the second criterion below
      forbids the way out of it.
      `.github/workflows/release.yml` `on:`, `.github/workflows/promote.yml` job `release`.

- [x] **Stated in terms of an observable fact about the run.** The `release` job asserts the
      release run's `head_sha` equals the commit being promoted — the same field #326's
      evidence was read out of, where run 32716466300 answered `51164f1`. `preflight` reports
      the parity of the three executed workflow files as a `::notice` on every promotion,
      whichever way it comes out. Nothing anywhere says which files a train may change.
      `promote.yml` jobs `preflight` and `release`.

- [x] **The failure is ordered so anything irreversible happens last, or is proven safe
      first.** The irreversible acts stay in `advance` — the tag must exist before
      `release.yml` runs, because `resolve` reads promote-vs-backport mode off where the tag
      lives — so this is the *proven safe first* half. The new `preflight` job runs before
      `advance` and refuses, while nothing has moved, on: a promoted commit with no
      dispatchable `release.yml` or no `container-image.yml`; a missing `DOCKERHUB_USERNAME`,
      `DOCKERHUB_TOKEN` or `NPM_TOKEN`, all three of which `release.yml` refuses on only after
      the fast-forward; and a version tag already naming another commit, which `advance`
      checks only after `main` has moved.
      It builds nothing, and that is deliberate — the train's own `ci.yml` built and smoked
      the tarball on every merge, with the train's own workflow file. Re-proving the train's
      work is not what was missing.

- [x] **A documented recovery for a half-run promotion, in `docs/ci-cd.md` beside *Cutting a
      release*.** § **"When a promotion half-runs"** — all four things it was asked for: what
      is safe to re-run and why each job is idempotent; what is burned (an npm version number,
      the tag, `main`, `:latest` once it has moved); the ordered commands that finish a
      promotion whose tag and `main` already moved; and how to retire the train the workflow
      left behind, with the same ancestry guard `retire-train` makes. A test asserts the
      section exists, sits inside § Cutting a release, and covers those points; `promote.yml`
      names it from four error messages.

- [ ] **That recovery is exercised at least once against the current state.** **Not done, and
      deliberately not done from this pull request.** Finishing `v0.4.0` means dispatching
      `release.yml`, which publishes two images, two npm packages and a GitHub Release; an npm
      version number is burned by its first publish and cannot be taken back. That is an
      operator's act on a live repository, not a reviewer's, and the instructions for this
      change explicitly excluded republishing `v0.4.0` or dispatching any workflow.
      What is delivered instead: the state is recorded verbatim in the runbook under *The
      state `v0.4.0` is in*, verified against origin today — `main` and `refs/tags/v0.4.0`
      are both `af3bf7b`, `gh release view v0.4.0` answers *release not found*, `gh release
      list` tops out at `v0.3.3`, and `@actana/cli` / `@actana/sdk` top out at `0.3.3` — and
      the exact steps that close it are steps 2–5 of the procedure with `version=0.4.0`.
      **No decision to skip publishing `0.4.0` is recorded, because that decision is not mine
      to take.** It needs an owner: either publish `0.4.0` from the tag that already exists,
      or record why `0.4.0` is abandoned and what happens to the number.

- [ ] **`beta/0.4.0` is retired, or an explicit decision is recorded to keep it.** **Not done
      from this pull request.** Deleting a branch on origin is an irreversible act on shared
      state, and it is the wrong order besides — `retire-train` runs *after* a successful
      release so a failed publish leaves the branch to look at, and `0.4.0` has not been
      published. Retiring it before the release is finished would throw away the one artefact
      an operator still wants.
      The procedure and its guard are step 4 of the runbook, and it is the same guard the
      workflow makes: delete only when the train's tip is reachable from `main`. As of today
      `beta/0.4.0` is `af3bf7b`, which *is* `main`, so it will delete cleanly once step 2 has
      succeeded. Two trains being open is the state that follows from the failed run, not a
      hotfix; the runbook says so.

- [x] **A regression test covers the shape, not the instance.** Seven new assertions in
      `scripts/__tests__/workflows.test.mjs`, all of them about shapes that are invisible in a
      green run: no `uses:` of `release.yml` anywhere and no `workflow_call` on it; the
      dispatch names `--ref`, and so does *every* `gh workflow run release.yml` in the
      repository, error messages and runbooks included; the run is found by id and its
      `head_sha` asserted; the promotion waits and carries the conclusion, so `release-line`
      and `retire-train` stay behind a successful publish; `advance` waits on `preflight` and
      `preflight` checks a dispatchable workflow, all three credentials and the tag, and
      builds nothing; the executed-workflow set is asserted **exactly**, so widening it to the
      whole directory — which would make a train that touched `ci.yml` look like a promotion
      problem — fails too; and the recovery exists in the right chapter.
      Mutation-checked: dropping `--ref` from the dispatch, and dropping `preflight` from
      `advance`'s `needs`, each turn a test red.

### Also, though #326 did not ask for it

- `scripts/build-core-tarball.mjs` — the message all three legs printed told the reader to run
  `pnpm --filter @actana/cli build`. Naming one filter is how the pair drifts apart, and that
  drift is the failure. It now names `pnpm build:core-tarball-bundles`, the one script
  `scripts/__tests__/core-tarball.test.mjs` asserts every caller uses.
- ADR 0023 D40 carries a dated amendment rather than an edit, in the house style.
- `docs/REPO_SETUP.md` and `.agents/skills/release/SKILL.md` carried the same
  dispatch-without-`--ref` advice; both are corrected and both are swept by the new test.

### Verification

`npx vitest run` (the root stage — every test changed here lives in it): **20 files, 581
tests, all passing.** `commitlint --from <base> --to HEAD`: 0 problems across all four
commits. `eslint` clean on the three changed `.mjs` files. Both workflow files parse.
`packages/panel` is flaky in this sandbox — two consecutive runs failed on *different*
unrelated files (`provider-usage-api`, `task-title-api`, then pairing and panel-link tests)
and nothing in this change touches that package.

mrfqcentic (changes requested):
## Review of #334 against #326 — request changes (four defects, none in the central mechanism)

Head reviewed: `76a11f3`, base `feat/release-pipeline` at `dceeb12`. All 23 checks on the head commit are green (23 completed: 18 success, 5 skipped by design — `Core image (train)`, `Panel image (train)`, `Resolve train tags`, and both `Publish the multi-arch manifest` jobs).

**The central claim holds, and I verified it at the enforcing lines rather than from the description.**

- `.github/workflows/release.yml:132-150` — `on:` is `workflow_dispatch:` alone. `workflow_call` is gone, so no local `uses:` can resolve this file from a caller's frozen `github.sha`. That is what made run `32716466300` execute `51164f1`'s copy. The trap is now unreachable, not merely unhit.
- `.github/workflows/promote.yml:785` — `gh workflow run release.yml --repo "$REPO" --ref "$tag" -f tag="$tag"`. `--ref` is the tag `advance` pushed at the promoted commit, so `release.yml` **and** the `container-image.yml` it calls at `release.yml:632,652` come from the train.
- `.github/workflows/promote.yml:816` — `if [[ "$run_sha" != "$HEAD_SHA" ]]` is the observable-fact assertion AC2 asks for, on the exact field #326's evidence was read out of.
- I also confirmed the trigger change does not regress the workflow standing alone: `release.yml` declares its own job-level `permissions` (`contents: write` on `github-release:937-938`, `id-token: write` on `npm:708-711`), so losing the caller's `permissions:` / `secrets: inherit` grant costs nothing.

The regression tests enforce rather than describe. I mutation-checked seven ways against an isolated copy of the head tree; each turned a test red:

| mutation | test that went red |
| --- | --- |
| drop `--ref` from the dispatch | `dispatches release.yml at the version tag, and says so everywhere` |
| restore `workflow_call:` on release.yml | `takes a dispatch and nothing else…` + `leaves no way to call release.yml…` |
| `advance` stops needing `preflight` | `proves the release runnable before anything irreversible happens` |
| neutralise the `head_sha` assertion | `asserts the release run's head SHA is the commit being promoted` |
| drop `--ref` from the runbook's recovery dispatch | `dispatches release.yml…` + `writes the half-run recovery down…` |
| rename the `### When a promotion half-runs` heading | `writes the half-run recovery down beside Cutting a release` |
| `release` stops waiting for the run | `waits for the release and fails the promotion when it does not succeed` |

The runbook is executable as written — I checked its claims against the code it describes, not just its prose: `npm view "$spec" version` gating a re-publish into a `::notice` at `release.yml:788-790`, and `gh release upload --clobber` / `gh release create` at `release.yml:1023,1028`. `v0.4.0^{commit}` peels correctly to `af3bf7b`, so steps 1 and 3 work on the annotated tag that actually exists.

---

### Blocking

**1. `.github/workflows/promote.yml:779,795` — the dispatched run is identified by a scheme that can latch onto another promotion's run, producing the exact half-run state this ticket exists to eliminate.**

`before` is a single-item snapshot (`.workflow_runs[0].id`) and the poll takes `min_by(.id)` over `select(.id > $before)`. `concurrency: group: promote-${{ inputs.train }}` (`:128`) is keyed on the train, so two promotions of *different* trains overlapping is a supported case — it is what the hotfix / `rebase-train` machinery exists for. If promotion B snapshots `before` while A's release run has not yet appeared in the listing, B's poll returns A's run, B fails its own `:816` assertion, and B aborts **after** `advance` has moved `main` and pushed `vx.y.z` — while the release B actually dispatched runs on unwatched. Narrow the query with `&head_sha=$HEAD_SHA` and take `max_by(.id)` over `id > $before`, computing `before` as a max over a page rather than element `[0]`.

**2. `.github/workflows/promote.yml:521` — `grep -qE '^ {6}tag:'` can pass a train whose `release.yml` has no `tag` dispatch input, which is the one thing this step exists to refuse.**

Six spaces is also the indent of `with:` keys on a reusable-workflow call — `release.yml:637,645` already carries `image:` and `source_tag:` at that depth. A train that renamed the dispatch input (`tag:` → `version:`) while passing a `tag:` to `container-image.yml` sails through preflight; `gh workflow run … -f tag=` at `:785` then fails with *unexpected inputs provided*, after `advance`. Scope the match to the `workflow_dispatch:` → `inputs:` block, or parse the YAML. The companion check at `:517` is correctly anchored; only this one is loose.

### Non-blocking, worth fixing while you are here

**3. `.github/workflows/promote.yml:827` vs `:758` — `deadline=$(( SECONDS + 10200 ))` (170 min) inside `timeout-minutes: 180` turns a slow-but-successful release into a half-run promotion.** `tarball-macos` at 10× plus the arm64 installer e2e plus two multi-arch image builds plus npm plus the GitHub Release can exceed 170 minutes on a queued macOS runner; the step then exits 1 and `release-line` / `retire-train` never run, for a release that publishes everything. The old `uses:` shape had no aggregate wall-clock cap at all, so this is a new failure mode. Either raise both, or treat "still running at the deadline" as a distinct outcome that does not strand the housekeeping.

**4. `docs/ci-cd.md:1298-1299` — the recovery misreports an already-retired train.** `tip="$(git ls-remote --heads origin "$train" | cut -f1)"` is empty when the branch is gone — the state an operator re-running step 4 is in — and `git merge-base --is-ancestor "" origin/main` then errors into the `else`, printing *"beta/0.4.0 grew past  after the promotion — retarget that work first"*. That tells an operator work is at risk when there is nothing left to do. Guard with `[[ -n "$tip" ]]` and say "already retired".

### Observations, not requests

- **The guarantee starts one promotion late, and nothing says so.** The promotion that ships this change executes `main`'s current `promote.yml:518-521` (`uses: ./.github/workflows/release.yml`) and `main`'s `release.yml:119` (`workflow_call:` still present), so there is no `preflight`, no `--ref` and no `head_sha` assertion on that run. It is harmless — `main` at `af3bf7b` already carries the fixed tarball legs — but an operator reading § *When a promotion half-runs* would reasonably expect otherwise. One sentence in the runbook or the ADR amendment would cover it.
- **AC3 is the cheap half by design, and the residual is real.** A train that adds a *required* dispatch input or a *new secret* to `release.yml` is not caught by `preflight` and fails after `advance` — and in the required-input case it fails as a bare `gh` 422 under `set -e`, without the crafted `::error` that points at the runbook. Enumerating `workflow_dispatch.inputs[*].required` in preflight would close the first half cheaply.
- **AC5 and AC6 are correctly left open and are not code.** `v0.4.0` still has no GitHub Release, npm still tops out at `0.3.3`, and `beta/0.4.0` is still on origin at `af3bf7b` beside `beta/0.4.1` — all verified against origin just now. Deferring the publish and the branch deletion to an operator is right, and the runbook is what they are performed from. But #326 asks for either the recovery exercised **or** a recorded decision not to publish `0.4.0` and what happens to the number, and neither exists yet on the issue. This PR says `Closes #326`; closing it in this state would close a ticket with two acceptance criteria open. That needs an owner's decision, not a diff.
- Boundaries are clean: `install.sh`, `packages/shared/src/actana-release-channel.ts`, `scripts/lib/core-tarball.mjs` and `scripts/compose-core-shasums.mjs` are byte-identical to the base. `scripts/build-core-tarball.mjs` is touched, but it is named in #326's own file pointers and only its error string changes.
- #325 is respected: `next-train` and `next_train` appear nowhere in the diff and nowhere in `.github/` at head; the job graph is `pause → resolve → {verify, preflight} → advance → release → {release-line, retire-train} → {train-invariant, rebase-train}`.
- The ASCII job diagram at `promote.yml:7-10` carries a dangling `└─` inherited from #325's deletion, and the new `(dispatch … at the … tag)` annotation straddles three lines. Cosmetic.

Fix 1 and 2 and this is an approve. The mechanism itself is the right shape and is proven, not asserted.

qcentic-dev:
## Re: review of `76a11f3` — all four fixed, one observation written down

Head is now `2bbc475`. The design is untouched; every change here is inside the two jobs this
branch added. Three commits: `70a1bcd` (F1–F3), `8e34f29` (F4 + the caveat), `2bbc475` (tests).

### F1 — the dispatched run could latch onto another promotion's · **fixed** · `70a1bcd`

Your remedy, taken as given. `promote.yml:835-837,850-852`:

```bash
before="$(gh api "repos/$REPO/actions/workflows/release.yml/runs?per_page=100" \
  --jq '[.workflow_runs[].id] | max // 0')"
…
gh api "…/runs?event=workflow_dispatch&head_sha=$HEAD_SHA&per_page=100" \
  --jq "[.workflow_runs[] | select(.id > $before)] | max_by(.id) | .id // empty"
```

`head_sha` removes the ambiguity; `id > $before` keeps the re-dispatch discrimination that
`head_sha` alone cannot give, since a re-dispatch shares it. Exercised against fixtures of
the exact race you describe — B's `before` = 100, A's run `105` at `AAA` surfacing before B's
`106` at `BBB`: the old filter returns `105`, the new one returns `106`. And the ordering
trap in `before` is real, not theoretical — on a page where `created_at` and `id` disagree,
`.workflow_runs[0].id` answers `100` where the max is `101`.

**I kept the head-SHA assertion** (`:880`), and I do not think it is redundant with the filter: the API
**silently ignores an unknown query parameter**, so a dropped or misspelled `head_sha` widens
the listing back to everything and says nothing. The filter is the mechanism, the assertion
is the observable fact, and AC2 is about the fact. There is a comment at the assertion saying
exactly that, so nobody deletes it as duplicated work later.

### F2 — the six-space `tag:` grep · **fixed** · `70a1bcd`

Scoped by walking the block rather than by pattern: `workflow_dispatch:` at two spaces, its
`inputs:` at four, input names at six inside that (`promote.yml:537-549`). `awk`, not a YAML
parser, because `preflight` installs nothing.

I built your adversarial input to check it rather than reasoning about it — `release.yml`
with the dispatch input renamed `tag:` → `version:` **and** a `tag:` added under the `panel`
job's `with:` at six spaces. The old grep matches it at line 638 and passes. The new check
reports `inputs=[version]` and refuses. `container-image.yml`, which has only a
`workflow_call`, yields an empty list and is refused; `promote.yml` yields `[train]`. The
error now names the inputs it *did* find, which is what an operator needs at that moment.

### F3 — 170 minutes inside 180 · **fixed** · `70a1bcd`

Both raised, and the outcome separated. The job is `timeout-minutes: 360` — GitHub's own
ceiling for a job, so it is the largest honest number — and the wait stops at 21000s, ten
minutes inside it, so this step's crafted message wins over GitHub's bare execution-time
error. The message no longer reads as a failed release, because it is not one:

> **The release has not failed** — it is still running, and it may still publish everything.
> What is stranded is the housekeeping this job gates … if it succeeds, only steps 3 to 5 of
> § *When a promotion half-runs* are outstanding, and the release itself needs nothing.

You were right that this is a failure mode the change introduced; the asymmetry is the whole
argument, and it is now in the comment on `timeout-minutes` rather than in my head.

### F4 — the recovery misreports an already-retired train · **fixed** · `8e34f29`

Reproduced first (`git ls-remote` on a branch that does not exist → empty `tip` →
`fatal: Not a valid object name` → your exact misleading sentence), then guarded.
`docs/ci-cd.md:1313-1321` now has an empty-tip arm printing *"already retired — nothing to
do"*, and the failure arm says what it means instead of "grew past".

### The guarantee starts one promotion late · **written down** · `8e34f29`

You asked for a sentence; it is a short paragraph in § *When a promotion half-runs*, because
that is where you correctly said an operator would be misled, plus a clause on the ADR 0023
D40 amendment, which is the record. It says the first promotion after this lands still runs
the old `promote.yml` and `release.yml` — no `preflight`, no `--ref`, no `head_sha`
assertion — that it cannot be otherwise since the change reaches `main` only by being
promoted, and that every promotion after it has the guarantee. A test pins the sentence, so
it cannot quietly go away.

### Tests

Four new assertions, and one rewritten (`2bbc475`). The wait-cap one is a **relationship**
rather than two literals — the deadline must fire before the job's timeout, with at least
five minutes of headroom, and the ceiling at least four hours — so it survives a future
retune of either number and still catches an inversion.

Mutation-checked seven ways, each red:

| mutation | test that went red |
| --- | --- |
| drop `head_sha` from the query | `finds its own release run by head SHA and id together…` |
| `before` back to `.workflow_runs[0].id` | same |
| `max_by` back to `min_by` | same |
| restore the loose six-space `tag:` grep | `proves the release runnable before anything irreversible happens` |
| job timeout back to 180 | `waits inside the job that carries the wait, and not up to its edge` |
| drop the empty-tip guard | `writes the half-run recovery down beside Cutting a release` |
| reword the one-promotion-late sentence | same |

`min_by(.id)` and `workflow_runs[0].id` are refused **by name**, not merely un-asserted —
each is the pre-review shape, and this file's convention is to refuse the shape that was
wrong.

### Deferred, deliberately

- **Enumerating `workflow_dispatch.inputs[*].required` in `preflight`.** You filed it as an
  observation and it is genuinely cheap now — `dispatch_inputs` already yields the names —
  but it is a new guard rather than a fix to one you found, and this branch is under review
  for #326. It wants its own change with its own test. The residual you named is real and
  unchanged: a train adding a *required* input or a *new secret* to `release.yml` still fails
  after `advance`, and in the required-input case as a bare `gh` 422 without the crafted
  error.
- **The dangling `└─` and the three-line annotation in the ASCII diagram.** Cosmetic, and it
  adds diff to a branch you have already read closely. Happy to take it if you would rather
  it went in here.
- **AC5 and AC6.** Unchanged and still not code: publishing `v0.4.0` and deleting
  `beta/0.4.0` are operator acts on a live repository, and I was explicitly scoped away from
  dispatching workflows or publishing. Your point about `Closes #326` is well made and is an
  owner's call — say the word and I will drop the `Closes` to a `Refs` so merging this cannot
  close a ticket with two criteria open.

### State

CI on `2bbc475`: **18 pass, 0 fail**, 5 skipped by design (`Core image (train)`,
`Panel image (train)`, `Resolve train tags`, both `Publish the multi-arch manifest`).
Root suite locally: 20 files, 583 tests, all passing. `commitlint` over all seven commits:
0 problems. Boundaries unchanged — `install.sh`, `packages/shared/src/actana-release-channel.ts`,
`scripts/lib/core-tarball.mjs` and `scripts/compose-core-shasums.mjs` are still byte-identical
to the base. Still a draft, base still `feat/release-pipeline`, nothing merged and no workflow
dispatched.

mrfqcentic (approved):
## Re-review of `2bbc475` — approve

All four findings fixed. I verified each at the enforcing line at the new head and re-ran the adversarial cases against the new code rather than against the reply; where you built a fixture I built my own and got the same answer. Base is unmoved (`feat/release-pipeline` still `dceeb12`, merge-base `dceeb12`, `mergeStateStatus: CLEAN`). CI on `2bbc475`: **23 checks, 18 success, 5 skipped by design, 0 failures.**

### F1 — run identification · **verified fixed** · `promote.yml:835-837,851-852`

All three parts of the remedy are in:

```bash
before="$(gh api \
  "repos/$REPO/actions/workflows/release.yml/runs?per_page=100" \
  --jq '[.workflow_runs[].id] | max // 0')"
...
  "repos/$REPO/actions/workflows/release.yml/runs?event=workflow_dispatch&head_sha=$HEAD_SHA&per_page=100" \
  --jq "[.workflow_runs[] | select(.id > $before)] | max_by(.id) | .id // empty" \
```

`head_sha=$HEAD_SHA` ✓, `max_by(.id)` over `select(.id > $before)` ✓, `before` as `[…] | max` over a 100-item page rather than element zero ✓. `min_by` no longer appears in the file.

**The overlapping-promotions case, re-constructed against the new code.** I ran both axes separately as jq fixtures.

*Axis 1 — the head-SHA filter.* Watermark correct at `before=104`; A's run `105` at `AAA` surfaces ahead of B's `106` at `BBB`:

| | picks | at |
| --- | --- | --- |
| old (`min_by`, no `head_sha`) | `105` | `AAA` — A's run |
| new (`head_sha=BBB`, `max_by`) | `106` | `BBB` — B's own |

*Axis 2 — the watermark.* Your point that `created_at` order and id order can disagree is not theoretical, and it is worse than I wrote it. On a page whose newest-created run is `100` while `101` already exists, the old `before` is `100`, so a **pre-existing** run `101` qualifies as "above the watermark" and `min_by` returns it — no second promotion required. New watermark answers `101`; the poll then returns `106`.

**Can B still latch onto A's run? No.** A's run carries `head_sha = AAA` and is excluded server-side before the id window is applied, so no ordering accident inside the listing can reach it. The two axes are genuinely independent, and each covers the other's blind spot exactly as your comment at `:812-830` says.

**And you are right to keep the assertion at `:880`.** I checked the reasoning rather than taking it: the GitHub Actions API does silently ignore an unrecognised query parameter, so a dropped or misspelled `head_sha` widens the listing back to every release run and reports nothing. The filter is the mechanism, the assertion is the fact, and AC2 is about the fact. Not redundant. The comment you left there is what stops someone deleting it as duplicated work.

Residual, not a request: if a person hand-dispatches `release.yml` at the same tag while a promotion is waiting, `max_by` may adopt that run instead. It is the same tag at the same commit, so the assertion passes and the wait is on an equivalent release — no wrong outcome, just imprecision.

### F2 — the tag-input guard · **verified fixed** · `promote.yml:537-549`

Scoped by walking `workflow_dispatch:` (2) → `inputs:` (4) → names (6), then `grep -qx 'tag'`. I extracted the `awk` block and ran it against the five real workflow files: `release.yml → [tag]`, `promote.yml → [train]`, `housekeeping.yml → [chore dry-run]`, `container-image.yml → []`, `ci.yml → []`. Correct on every one.

**The counterexample, rebuilt.** I generated a `release.yml` with the dispatch input renamed `tag:` → `version:` **and** a decoy `tag:` added at six spaces under the `panel` job's `with:` — it lands at line 638, the depth `image:`/`source_tag:` occupy:

```
OLD guard  (grep -qE '^ {6}tag:'):  PASSES   <- the F2 hole
NEW guard  (awk block walk):        dispatch_inputs = [version]  -> REFUSES
```

**It no longer sails through.** The error naming the inputs it did find is the right call — that is the one thing an operator needs at that moment.

I also probed the parser's failure modes, since a hand-rolled walker's blind spots matter more than its happy path. A trailing comment on `workflow_dispatch:` or on `inputs:` breaks detection and yields an empty list, which **refuses**. Every mis-parse I could construct is fail-closed — it refuses before `advance`, so the cost is a false stop with nothing published, never a false pass into the state this step exists to prevent. That is the correct direction for this guard and worth having on the record.

### F3 — the wall-clock cap · **verified fixed** · `promote.yml:797,895,915`

`timeout-minutes: 360` (GitHub's ceiling, so the largest honest number) with `deadline=$(( SECONDS + 21000 ))` — 350 minutes, ten inside it, so your crafted message wins over GitHub's bare execution-time error.

**Can a slow-but-successful release still strand the housekeeping? Yes, but only past 5h50m, and it is now bounded as far as it can be.** The step still exits 1, so `release-line` and `retire-train` are still skipped. That is the right trade — exiting 0 would let `retire-train` delete the train while the release might still fail, which is strictly worse — and 360 cannot be raised. What changed is that the window went from 170 minutes (reachable by a queued macOS runner) to 350 (not plausibly), and the message no longer lies about it: *"**The release has not failed** … only steps 3 to 5 … are outstanding"*. The asymmetry argument in the `timeout-minutes` comment is the part that will keep this correct through the next retune.

### F4 — the already-retired train · **verified fixed** · `docs/ci-cd.md:1313-1321`

`if [[ -z "$tip" ]]` arm added, and the failure arm reworded off "grew past". I exercised all three arms against the live repository: empty tip → *"already retired — nothing to do"*; `tip = af3bf7b` (= `main`) → the delete; `tip` ahead of `main` → the accurate "not reachable from main" message. The old code on the same empty tip still produces `fatal: Not a valid object name` into the misleading arm.

### The observation — **written down**, in both places

`docs/ci-cd.md:1244-1252`, inside § *When a promotion half-runs*, which is where I said an operator would be misled:

> **The fix arrives one promotion late, and this is the section where that matters.** … **the first promotion after this change lands still runs the old `promote.yml` and the old `release.yml`**: no `preflight`, no `--ref` on the release, and no `head_sha` assertion. It cannot be otherwise, because the change reaches `main` only by being promoted.

plus the closing clause on the ADR 0023 D40 amendment — *"it is the last one that does not"* — which is the record. I asked for one sentence in either; both is better. The runbook half is pinned by a test; the ADR clause is not, which is fine.

### Tests

159/159 green on my own extraction of the head tree. I mutation-checked the four fixes independently — eight mutations, every one red:

| mutation | test that went red |
| --- | --- |
| drop `head_sha=$HEAD_SHA` from the poll | `finds its own release run by head SHA and id together…` |
| `before` back to `.workflow_runs[0].id` | same |
| `max_by(.id)` → `min_by(.id)` | same |
| restore the loose `grep -qE '^ {6}tag:'` | `proves the release runnable before anything irreversible happens` |
| job timeout back to 180 | `waits inside the job that carries the wait, and not up to its edge` |
| deadline to 21500s (headroom < 5 min) | same |
| deadline to 22000s (outlives the job) | same |
| drop the empty-tip guard | `writes the half-run recovery down beside Cutting a release` |
| reword the one-promotion-late sentence | same |

The relationship-shaped wait-cap assertion is the right instinct — it catches an inversion, a squeezed headroom **and** a lowered ceiling, and survives a retune of either literal. Refusing `min_by(.id)` and `workflow_runs[0].id` by name is also right for this file's convention: the shape that was wrong should stay named.

### Re-checked, unchanged

- **Mechanism intact at the new head.** `release.yml:132-150` is `workflow_dispatch` alone; the only two `uses: ./.github/workflows/release.yml` hits in the repository are inside comment prose (`promote.yml:744`, `release.yml:108`). `promote.yml:842` still dispatches with `--ref "$tag"`. Job graph unchanged: `pause → resolve → {verify, preflight} → advance → release → {release-line, retire-train} → {train-invariant, rebase-train}`.
- **Boundaries hold to the byte.** `install.sh` (`6eb97ab`), `scripts/lib/core-tarball.mjs` (`bd82f33`), `scripts/compose-core-shasums.mjs` (`32202657`) and `packages/shared/src/actana-release-channel.ts` (`96deec1`) are all identical to the base. Same nine files as `76a11f3`.
- **#325 respected.** No `next-train` job, no `next_train` output anywhere. The single prose hit at `promote.yml:314` — *"next train's version is an administrative decision a person makes"* — is on the base unchanged and reinforces #325 rather than reintroducing anything.
- **AC5 and AC6 remain open**, verified against origin just now: no `v0.4.0` Release, npm still `0.3.3`, `beta/0.4.0` still at `af3bf7b` beside `beta/0.4.1`, and #326 still has zero comments. Still operator acts, still correctly deferred.

### Deferred items — I agree with all three

Enumerating `workflow_dispatch.inputs[*].required` is a new guard, not a fix to one I found, and it deserves its own change and its own test; the residual is unchanged and accurately stated. The ASCII diagram is cosmetic — leave it. And yes on the last one: **`Closes #326` should become `Refs #326`** before this merges, so that landing it cannot auto-close a ticket with AC5 and AC6 open. That is a body edit, not a code change, and it is the one thing I would do before merge.

### Merge order — **#334 first, then #333 rebased onto it**

The two are independent, so this is a preference and not a constraint. I checked the coupling rather than assuming it:

- `docs/ci-cd.md` is the **only** shared file. #333's hunks are at base lines 980-992, 1014-1020 and 1036-1050 (§ *Cutting a train*); #334's are at 14, 532, 1185-1196 and 1201-1228 (§ *Cutting a release*). ~135 lines apart at the closest point — no textual conflict in either order.
- #333 deliberately adds its `install.sh` stamp **below** `files=()` rather than into it, so #334's `files=()` ↔ `ci.yml` `MANIFESTS` binding assertion is untouched.
- #333 adds no `gh workflow run release.yml` line, so #334's repository-wide `--ref` sweep is unaffected.

#334 first, for two reasons. It carries the guards — the `--ref` sweep and the § *Cutting a release* structural assertions both read `docs/ci-cd.md` — and a guard that lands after the change it guards leaves a window in which it was not enforced; landing #334 first means #333's doc edits are checked against it on #333's own CI. And #334's two upper hunks (`:14`, `:532`) are net-zero-line edits, so landing it first leaves #333's hunk line numbers **identical** — the cleanest possible rebase, where the reverse shifts #334's two large hunks by +18.

Both are drafts; marking either ready is not mine to do.

---

**Approved.** The mechanism was already the right shape at `76a11f3`; what was wrong was the machinery around it, and all four defects are now fixed at the line rather than in the description. The two that blocked — a run-identification scheme that could manufacture the very half-run this change exists to prevent, and a guard that could pass the one train it existed to refuse — are closed and proven closed against the adversarial inputs, and both are pinned by tests that fail when the old shape returns.

Issue #320, opened 2026-08-24
The CLI-only install path for a beta, without a registry version


mrfqcentic:
Part of #314 — a beta train must be installable the same way a release is. Read that issue
first; the design there is fixed. The record is #315.

> **Rewritten.** This ticket originally specified an npm publish at `0.4.1-beta.N` on the
> `next` dist-tag. **The counter is gone and is not coming back**: a beta version string is
> `x.y.z-beta`, exactly, everywhere — no counter, no `.N`, no suffix of any kind, on any
> surface. That is a fixed operator constraint, recorded in #315, and it is what makes the
> registry publish impossible as originally specified. What follows is the ticket that
> replaces it.

## The problem

`npm i -g @actana/cli` is one of the three install surfaces (#314 §1), and the only one for a
machine that **drives Cores it does not host**: no daemon, no service unit, no Core on this
box — just the client nouns. Today it has no beta. `release.yml`'s `npm` job publishes only on
a promotion, so the newest `@actana/cli` anywhere is always the last released version, and an
operator testing a train has no way to get the CLI half of it.

## Why the registry publish cannot be the answer

npm consumes a version on first publish and never gives it back. This repository already has
the property written down, in the file that exists because of it:

> An npm version number is burned by its first publish and cannot be reused — unpublishing
> inside the 72-hour window frees the bytes and not the name. `@actana/sdk@0.2.2` gets exactly
> one attempt, ever.

`scripts/rehearse-npm-publish.mjs:11-14`.

With the version string fixed at `0.4.1-beta`, `@actana/cli@0.4.1-beta` can be published
**once per train**. The second cut of that beta — and a beta is *designed* to be cut repeatedly
as the train moves — fails at the registry with a 403, after the git tag has already moved and
every asset has already been replaced. A publish path that works exactly once and then breaks
the workflow that calls it is not a path. **Drop it.**

Every other surface is fine with a fixed moving string, which is worth stating so the
asymmetry is not mistaken for an oversight: a git tag can be force-updated, a Docker tag is a
pointer and `beta-x.y.z` already moves per merge (ADR 0023 D7), and a Release asset is
clobbered in place. npm is the one namespace in this design that is write-once.

## The decision to take — #315 chooses one

- **(a) No npm on the beta path.** A beta is installed by `install.sh` or run as a container,
  and the CLI-only surface has no beta. Cheapest; close this issue as `wontfix` and say so in
  #314 §3.
[...]

PR #336, opened 2026-08-24
feat(core-tarball): the version stays on all three surfaces, and a beta says beta

qcentic-dev:
Implements #321 under the decision [ADR 0036](https://github.com/actana/control/blob/feat/release-pipeline/docs/adr/0036-the-beta-release-channel.md) §E already reached. Base is `feat/release-pipeline`.

## Read this first: the branch name is wrong on purpose, and the work is not

The branch is called `fix/core-tarball-version-neutral`, and **a version-neutral Core tarball is exactly what this PR does not build**. ADR 0036 **D19** refuses that route once, with its price attached: taking the version out of the asset name, the archive root and the manifest is a change to the installer contract 0016 D29 names, *"a breaking change requiring a major bump plus a documented migration"* — the wrong trade at 0.4.1. **D18** drops the tarball half of promotion instead: a digest is version-neutral, a Core tarball is not, and that is a property to keep rather than a defect to remove. `release.yml` is unchanged, and so is the shape of `core-manifest.json`.

So what is implementable here is the other half of the record: if a tarball self-identifies, make it say one true thing, make a beta's string exactly `x.y.z-beta` (**C1**), and make beta assets and release assets unconfusable in the artifact rather than in a promise (**D20**).

## What changed

**One gate, three doors** (`scripts/lib/core-tarball.mjs`). `assertCoreVersion` is the version contract in one place, and `tarballName`, `tarballRootDirName` and `buildManifest` all call it — no string can reach one surface by a path that skips the other two. The rule is C1's: a beta is `x.y.z-beta`, with nothing after the word, on every surface including the asset filename. A counted beta (`0.4.1-beta.1`, `0.4.1-beta1`, a run number, a date, a short sha) is refused outright. The backport candidate 0023 D30 publishes keeps its identifier, so the ban is on a counted *beta*, not on every prerelease.

New pure helpers, all unit-tested: `parseTarballName` (reads a name back the way `install.sh` does), `assertShasumsSet` (what one checksum file may cover), `assertTarballSurfaces` (the asset name, the archive root and the manifest must be one string).

**The builder** (`scripts/build-core-tarball.mjs`) gates `--version` before anything is staged, so a bad string fails with a `[core-tarball]` line instead of deep inside a tar, and the log names the channel. There is no channel field in the manifest and this PR does not add one — a beta is a beta because its version says so.

**The composer** (`scripts/compose-core-shasums.mjs`) now reads every basename through that contract before hashing: an unrecognisable file, a second tarball for one target, and two versions in one directory are each refused by name. A `SHA256SUMS` listing `0.4.1-beta` beside `0.4.1` is precisely the confusion D20 rules out, published under a single name. `--expect` is untouched; the new `--version` is optional and nothing passes it today.

**The smoke** (`scripts/smoke-core-tarball.mjs`) checks the three surfaces against each other on the extracted bytes before it boots anything. The builder derives all three from one variable, so only the artifact can prove it.

**The consumer** (`packages/cli/src/actana-layout.ts` + tests) needed no code, and that is the claim: the manifest shape is unchanged, `installDirFor` already treats the version as an untrusted path segment, and two different strings are two different directories. A beta lands in `versions/0.4.1-beta`, beside its line's release rather than on top of it, and `placeCoreBundle` reports `0.4.1-beta`. Asserted rather than assumed.

## Proved by building, not by reading

The tarball job's earlier failure was the CLI bundle (`packages/cli/dist-tarball`), so nothing here is claimed off the source:

- Built `actana-core-0.4.1-beta-linux-x64.tar.gz`: asset name, archive root **and** `core-manifest.json` all `0.4.1-beta`, `protocolVersion 0.17.0`.
- Smoked it: `OK — extracted tarball boots a dialable Core with no system Node`.
- Built and smoked `0.4.1` as well — the release path is unchanged.
- **Renamed the beta file to the release's name and re-smoked it**: refused with *"extracts to `actana-core-0.4.1-beta-linux-x64`, not `actana-core-0.4.1-linux-x64` — the asset name and the archive root disagree"*. That is D18's central claim, checked against bytes.
- Composer over a directory holding both: refused, *"two versions in one checksum set"*. Over the beta alone: `SHA256SUMS` written and verified with `sha256sum -c`.
- `node scripts/build-core-tarball.mjs --version 0.4.1-beta.1` → `[core-tarball] counted beta version … (ADR 0036 C1)`.

Unit tests: `scripts/__tests__/core-tarball.test.mjs` 58 passing; `install-sh.test.mjs` 43 passing (the naming contract's other reader); `packages/cli` layout + setup suites passing; ESLint and `tsc --noEmit` clean on every file touched.

## Notes for the tickets that own the files I did not touch

I own none of these and edited none of them.

**#327 (`.github/workflows/*`)** — nothing is *required*. Two optional pickups, both additive:
1. `release.yml`'s `compose-core-shasums.mjs --dir core-tarballs --expect 3` may gain `--version "$RELEASE_VERSION"`, which asserts the composed set is the version being released rather than something left in the directory.
2. Same for the `github-release` job if it ever wants the assertion earlier than the asset-count check.

Everything already in `release.yml` keeps working unchanged, including `--expect 3` and both tarball legs. Under D18 it stays a build, not a promotion.

**#318 (`beta-release.yml`)** — the builder takes the beta version directly:
```
node scripts/build-core-tarball.mjs --target "$TARGET" --version "$LINE-beta" --out-dir artifacts/core
node scripts/compose-core-shasums.mjs --dir core-tarballs --expect 3 --version "$LINE-beta"
```
`pnpm build:core-tarball-bundles` must run first — both bundles, not just the Core's; that is what broke the tarball legs before, and `core-tarball.test.mjs` asserts every workflow calling the builder runs it. A leading `v` is stripped, so a tag ref works; a counted string (`-beta.${{ github.run_number }}`) is refused by the builder, by design.

**#320 (npm)** — unaffected. `assertCoreVersion` governs the Core tarball's surfaces only; nothing here touches `packages/cli/build.mjs`, its manifest or `assertPackedManifest`.

**#322 / #317 (version comparison, out of scope here)** — worth knowing: `isNewerSemver` in `packages/shared/src/semver.ts` compares the numeric core only, so a machine on `0.4.1-beta` is never told `0.4.1` is available (`actana-update-check.ts:148`). That is the "compared numerically" question ADR 0036 assigns to #322's second criterion, not this ticket's file set — flagging it rather than editing it.

**Not touched, as instructed:** any workflow, `install.sh`, `packages/cli/build.mjs`, `packages/cli/package.json`, `docs/`. Nothing merged, nothing dispatched, `main` untouched.

## One environment note

Running the tarball smoke *inside* an Actana Core container makes the spawned daemon read the container contract it inherits (`ACTANA_CONTAINER=1`, `ACTANA_PORT`) and ignore the port `coreSmokeEnv` picked, so it collides with the host Core on 443. Scrubbing those variables makes it pass. CI never runs the smoke inside a Core, so nothing is broken here; `scripts/lib/core-smoke.mjs` is not in this ticket's file set, and this is a note rather than a change.

### DIFF
--- a/packages/cli/src/__tests__/actana-layout.test.ts
+++ b/packages/cli/src/__tests__/actana-layout.test.ts
@@ -107,6 +107,17 @@ describe("installDirFor", () => {
     expect(installDirFor(l, "0.2.0")).toBe("/home/op/.local/share/actana/versions/0.2.0");
   });
 
+  // ADR 0036 D20: a beta and its line are different strings, so they are
+  // different directories — an install from a beta tarball cannot land where
+  // the release's would, or be reported as it.
+  it("gives a beta a directory of its own, beside its line's release", () => {
+    const l = resolveActanaLayout({}, HOME, "linux");
+    expect(installDirFor(l, "0.4.1-beta")).toBe(
+      "/home/op/.local/share/actana/versions/0.4.1-beta",
+    );
+    expect(installDirFor(l, "0.4.1-beta")).not.toBe(installDirFor(l, "0.4.1"));
+  });
+
   it("refuses a version that would escape the versions dir", () => {
     const l = resolveActanaLayout({}, HOME, "linux");
     expect(() => installDirFor(l, "../../etc")).toThrow(/version/i);
--- a/packages/cli/src/__tests__/actana-setup.test.ts
+++ b/packages/cli/src/__tests__/actana-setup.test.ts
@@ -192,6 +192,18 @@ describe("placeCoreBundle — what `install.sh` leaves behind", () => {
   });
 
   // The other half of the contract, and the one the whole ticket is about.
+  // The consumer half of ADR 0036 D18: the tarball keeps stating its version in
+  // its manifest, so a beta bundle installs under the beta string and the
+  // machine reports that. Nothing here special-cases a beta, and that is the
+  // claim — the version is a string this side reads, not a channel it decides.
+  it("installs a beta under its own version, and reports it", () => {
+    const result = place({ manifest: { ...MANIFEST, version: "0.4.1-beta" } });
+
+    expect(result.installDir).toBe(path.join(layout.versionsDir, "0.4.1-beta"));
+    expect(result.version).toBe("0.4.1-beta");
+    expect(fs.existsSync(path.join(layout.versionsDir, "0.4.1"))).toBe(false);
+  });
+
   it("activates nothing: no config, no material, no unit, no data dir", () => {
     place();
 
--- a/packages/cli/src/actana-layout.ts
+++ b/packages/cli/src/actana-layout.ts
@@ -125,6 +125,15 @@ export function resolveActanaLayout(
  * The version string comes from a downloaded tarball's manifest, so it is
  * treated as untrusted input: anything that is not a plain path segment is
  * rejected rather than allowed to escape `versionsDir`.
+ *
+ * A beta's version is `x.y.z-beta` (ADR 0036 C1) and gets a directory of its
+ * own, beside the release of the same line rather than on top of it. That is
+ * the whole of what this side of the tarball contract has to do about a beta,
+ * and it is why it needs no change: the tarball keeps carrying its version in
+ * the manifest (0036 D18), a beta's string differs from its line's, and two
+ * different strings are two different directories. A machine installed from
+ * `actana-core-0.4.1-beta-linux-x64.tar.gz` therefore lands in
+ * `versions/0.4.1-beta` and reports that — never `0.4.1` (0036 D20).
  */
 export function installDirFor(layout: ActanaLayout, version: string): string {
   if (!version || version !== path.basename(version) || version === "." || version === "..") {
--- a/scripts/__tests__/core-tarball.test.mjs
+++ b/scripts/__tests__/core-tarball.test.mjs
@@ -3,24 +3,31 @@ import * as path from "node:path";
 import { describe, expect, it } from "vitest";
 
 import {
+  BETA_PRERELEASE,
   BUNDLED_NODE_VERSION,
   CORE_RUNTIME_DEPENDENCIES,
   CORE_TARGETS,
   UNBUNDLED_EXTERNALS,
+  assertCoreVersion,
+  assertShasumsSet,
+  assertTarballSurfaces,
   buildManifest,
   findTarget,
   formatShasums,
   hostTarget,
+  isBetaVersion,
   nodeDistDirName,
   nodeDistShasumsUrl,
   nodeDistTarballUrl,
   parseCoreLinkProtocolVersion,
   parseShasums,
+  parseTarballName,
   planDependencyLayout,
   prebuildDirName,
   tarballName,
   tarballRootDirName,
 } from "../lib/core-tarball.mjs";
+import { parseAssetName } from "../lib/fixture-release.mjs";
 
 const repoRoot = path.resolve(import.meta.dirname, "..", "..");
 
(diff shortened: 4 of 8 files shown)


qcentic-dev:
## Acceptance criteria — #321

Checked against the issue's list, read under the decision ADR 0036 §E made. The record chose **(a)**: the tarball half of promotion is dropped (D18), a version-neutral artifact is refused with its price attached (D19), and beta and release assets stay unconfusable because they are different bytes with different names (D20).

- [x] **#315 records the decision — (a) or (b) — with the reasoning, and this issue is closed against that record.** ADR 0036 §E, D18–D20, already on `feat/release-pipeline` (`5c5ddde`). This PR implements it and does not reopen it. Closing #321 is the issue-owner's call, not this PR's.
- [x] **If (a): `release.yml` is unchanged.** Not touched — no workflow is. `--expect 3`, both tarball legs and the `github-release` asset count all keep working; verified by `scripts/__tests__/workflows.test.mjs` and `panel-image.test.mjs`, both green.
- [ ] **If (a): `docs/ci-cd.md` §*The digest guarantee* gains a sentence, and #314 §6 is retracted in that issue's body.** **Not mine — #323 owns the docs pages**, and the issue-body retraction is an edit to #314. ADR 0036 D18 already carries the sentence's substance (*"a tarball self-identifies, a digest does not"*) and names #323.
- [x] **If (b): a separate milestone, a major bump, a documented migration. Not this train.** No route (b) work here — D19 refuses it, and this PR is the reason it does not need refusing again: the version stays in the asset name, the archive root and the manifest.
- [x] **Either way, `release.yml` keeps building `linux-x64`, `linux-arm64` and `mac-arm64` at the promoted version, and `SHA256SUMS` covers exactly those three (0016 D28).** Unchanged, and now enforced from the other side too: `assertShasumsSet` refuses a set holding two versions or two tarballs for one target, so a checksum file covering part of a release, or part of two, fails at composition.
- [x] **Whatever is decided, beta assets and release assets are never confusable: an operator who downloads `actana-core-0.4.1-beta-linux-x64.tar.gz` must not be able to end up with a machine reporting `0.4.1`.** This is the criterion the PR actually implements, at four points:
  - the version is gated once and reaches all three surfaces through that gate — a beta is `x.y.z-beta`, never a counted string (ADR 0036 C1);
  - the smoke asserts the asset name, the archive root and `core-manifest.json` are one string, on the extracted bytes;
  - `versions/<manifest.version>` gives a beta its own directory beside the line's release, and `placeCoreBundle` reports `0.4.1-beta`;
  - a `SHA256SUMS` cannot cover a beta and a release at once.

  Demonstrated rather than argued: renaming the built beta tarball to `actana-core-0.4.1-linux-x64.tar.gz` and re-running the smoke fails with *"the asset name and the archive root disagree"*.

### Proof run

| Step | Result |
| --- | --- |
| `build-core-tarball.mjs --version 0.4.1-beta` | `actana-core-0.4.1-beta-linux-x64.tar.gz`, root `actana-core-0.4.1-beta-linux-x64`, manifest `"version": "0.4.1-beta"` |
| `smoke-core-tarball.mjs` on it | surfaces agree; `OK — extracted tarball boots a dialable Core with no system Node` |
| same for `--version 0.4.1` | builds and boots — the release path is unchanged |
| beta file renamed to the release name | refused by the smoke |
| `--version 0.4.1-beta.1` | refused: `counted beta version … (ADR 0036 C1)` |
| composer over beta + release in one dir | refused: `two versions in one checksum set` |
| composer over the beta alone | `SHA256SUMS` written; `sha256sum -c` OK |

`core-tarball.test.mjs` 58 passing, `install-sh.test.mjs` 43 passing, `packages/cli` layout + setup suites passing, ESLint and `tsc --noEmit` clean on every file touched.

qcentic-dev:
Rebased onto `feat/release-pipeline` after #326 and #317 landed (clean, no conflicts). Re-verified on the new base: `core-tarball`, `fixture-release`, `install-sh` and `workflows` suites all green (217 tests), and the beta tarball rebuilt and re-smoked — `channel=beta`, three surfaces agreeing, `OK — extracted tarball boots a dialable Core with no system Node`.

#326's clearer bundle-missing message is untouched and sits directly above the new version gate.

mrfqcentic (approved):
**Approved.** Reviewed against #321's acceptance criteria and ADR 0036 §E (D18, D19, D20) and C1, at head `24dc2dd`, base `feat/release-pipeline` @ `e233d50`. Everything below is checked from the enforcing line or from a CI log, not from the description.

### The record is implemented, not reinterpreted

D18 drops the tarball half of promotion and leaves `release.yml` unchanged: no file under `.github/workflows` appears in any of the four commits, and `release.yml`'s two tarball legs and `--expect 3` are identical to base. D19's refused route — a version-neutral artifact — is *not* taken. The branch is named `fix/core-tarball-version-neutral` and the diff does the opposite; the diff is right.

The three surfaces stay, and now pass one gate: `assertCoreVersion` (`scripts/lib/core-tarball.mjs:263`), called from `tarballName` (`:228`), `tarballRootDirName` (`:233`) and `buildManifest` (`:381`). `scripts/build-core-tarball.mjs:243` gates `--version` before anything is staged and derives all three from that one variable (`:293` root, `:319`/`:340` manifest, `:351` asset name), so there is no path to one surface that skips the other two.

### Proved on bytes

CI run 32767716854 on this head — *Smoke — Core release tarball (linux-x64)* (job 97560864840) and *(linux-arm64)* (97560864844) — builds a real tarball and runs the new assertion:

```
[core-tarball] target=linux-x64 version=0.4.1 channel=release node=24.15.0 protocol=0.17.0
[tarball-smoke] asset name, archive root and manifest agree: 0.4.1 linux-x64
[tarball-smoke] OK — extracted tarball boots a dialable Core with no system Node
[core-shasums] wrote …/SHA256SUMS — 0.4.1, covering linux-x64
```

Both installer e2es then install that same artifact through `install.sh`. The *beta* build in the description is a local run and nothing in CI reproduces it yet — that is #318's job, and a note is the right disposition here.

### Consumers

`install.sh:510` (asset name) and `install.sh:558` (archive root) against the builder's two names; `installDirFor` (`packages/cli/src/actana-layout.ts:136`) and `runActanaSetup`/`placeCoreBundle` (`actana-setup.ts:275`, `:286`, `:330`) against `core-manifest.json`. With #333 merged, `resolve_version` sets `VERSION="$LINE-beta"` (`install.sh:456`) — exactly the string this builder now emits.

### The beta filename

Every occurrence of `beta` followed by a dot-or-digit in the diff (5 hits, all in `scripts/__tests__/core-tarball.test.mjs`) is a prohibition, not a production. I exercised the gate directly: `0.4.1-beta.1`, `-beta1`, `-beta-2`, `-beta.20260824`, `-BETA`, `-Beta.3` all throw `counted beta`; `0.4.1-beta` and `1.2.4-rc.1` pass, and `assertTarballSurfaces` refuses beta bytes under a release name.

### Non-blocking notes — none of these hold the merge

1. `scripts/lib/core-tarball.mjs:318` — `parseTarballName` catches `assertCoreVersion` and returns `null`, discarding the reason. A counted beta that reaches the composer therefore reports *"not a Core tarball asset name"*, sending triage toward "stray file" rather than "counted version". The one failure this gate exists for is the one that does not name itself.
2. `scripts/lib/core-tarball.mjs:263` — the rule fires only when the counter follows the literal word: `0.4.1-42` or a `git describe` string passes and goes on to name all three surfaces. C1 bans a *counted beta* and 0023 D30 needs `rc.1` to survive, so this is a defensible scope; an allowlist (`beta` | `rc.<n>`) would close the residue.
3. `scripts/compose-core-shasums.mjs:95` — the new `--version` compares raw strings and never strips a leading `v`, while the builder explicitly does (`build-core-tarball.mjs:233`) and `assertCoreVersion`'s own doc says a caller reading a tag strips it first. Nothing passes it today; a beta cut handing it `$GITHUB_REF_NAME` would die with a mismatch between two identical versions.
4. `packages/cli/src/actana-layout.ts:136` — the comment's *"reports `0.4.1-beta` … never `0.4.1`"* holds for the Core (`versionLines` adds `Core installed here: 0.4.1-beta`), but the first line of `actana --version` is `CLI_VERSION` from `packages/cli/package.json`, which the cut stamps with the bare line. Not this ticket's file set (#320 / #327) — a note there rather than a change here.

### Boundary and state

Nothing under `.github/workflows`, nothing in `packages/cli/build.mjs`, `packages/cli/package.json` or the npm scripts, no `docs/` page — verified commit by commit, not just on the net diff. The two workflow pickups are notes in the description, which is what #327 and #318 need. Head is a descendant of the current base tip, so #317 (via #333) and #326 (via #334) are already under it; `MERGEABLE` / `CLEAN`; all 20 CI jobs green.

Still a draft — marking it ready is the author's call, not the reviewer's.

PR #337, opened 2026-08-24
feat(npm): the CLI-only beta install path — a packed asset, never a registry version

### LINKED ISSUE #320: The CLI-only install path for a beta, without a registry version

qcentic-dev:
Closes #320. Base is `feat/release-pipeline` — the feature branch for #314, not `main` and not the beta train.

## What this is

The CLI-only install path for a beta: the machine that **drives Cores it does not host** — no daemon, no service unit, no Core on the box — installing the client half of a train.

The route is fixed by the operator and by [ADR 0036](docs/adr/0036-the-beta-release-channel.md) §D and is **not reopened here**: the CLI is **packed and attached to the `v<x.y.z>-beta` prerelease as a Release asset**, installed with `npm i -g <asset-url>`, and **never published to registry.npmjs.org under a beta version**. D15's reason in one line — an npm version is burned by its first publish (`scripts/rehearse-npm-publish.mjs:11-14`), a beta string carries no counter (C1), and a beta is *designed* to be cut repeatedly, so the second cut of `0.4.1-beta` would 403 after the tag has moved and every asset has been replaced.

## The one question the route left open, and which of the three answers this is

`@actana/cli` declares `"@actana/sdk": "workspace:*"`; pnpm resolves that to a real version at pack time; under D15 that version is on no registry, so `npm i -g <asset-url>` would fail resolving it.

**This is route 2: the beta manifest drops the dependency**, because `packages/cli/build.mjs` externalises only `ws`, `undici` and `selfsigned` and therefore inlines the SDK into the bundle already. The choice is stated in a comment where the pack happens — `rehearseBeta`'s doc block in `scripts/rehearse-npm-publish.mjs`.

**Why not route 1** (attach the SDK tarball too and point the CLI's range at its asset URL). Not elegance — the acceptance criterion. #320 asks that `npm i -g <asset URL>` be *"asserted in CI, not assumed. That assertion is the whole ticket."* A URL range can only be asserted once the Release exists to serve it, so on a pull request the check would install from a `file:` URL and hope the `https:` one behaves the same. Route 2's tarball resolves nothing that is not on the public registry, so **the bytes CI installs are the bytes the operator installs**, and the check is the real `npm i -g`.

**Why not route 3** (attach both, document a two-step install): refused deliberately, exactly as #320 lists it — honest, and worse for the operator.

**The cost route 2 carries, stated rather than hidden.** The beta manifest's dependency set differs from the release manifest's by one name, which is the kind of divergence the packing guards exist to catch. So it is not left to a convention: `assertPackedManifest`'s `beta` branch **requires** the absence, refusing `@actana/sdk` at any range at all, and `assertBundleInlines` reads the packed bundle to confirm the code the manifest stopped naming is actually in the tarball.

## Proven, not reasoned about

The landmine in this ticket is "the workspace dependency is rewritten at pack time, and the SDK is *probably* already inlined". That was proven by packing and inspecting the artifact:

```
$ tar -xzOf actana-cli-0.4.1.tgz package/package.json   # the release pack, for contrast
  "dependencies": { "selfsigned": "^5.5.0", "undici": "8.9.0", "ws": "8.21.0", "@actana/sdk": "0.4.1" }
```

`workspace:*` → `0.4.1`, confirmed. And the bundle out of the tarball:

```
$ tar -xzOf actana-cli-0.4.1.tgz package/dist/actana-cli.mjs | grep -oE 'from *"[^"]+"' | sort -u
  node:child_process node:crypto node:fs node:https node:module node:net node:os node:path
  node:readline node:stream node:stream/promises node:string_decoder node:tls
  selfsigned  undici  ws
```

Three externals and Node builtins. **No `@actana/*` specifier at all**, and the sourcemap names **15 `packages/sdk/src` modules** inlined (plus 30 from `@actana/shared`).

One thing found by reading the bytes that reasoning would have missed: the bundle **inlines its own `package.json`** — that is where `actana --version` gets its answer — so the literal text `"@actana/sdk": "workspace:*"` is in the bytes of a bundle that imports nothing of the sort. `assertBundleInlines` matches specifier position for that reason, and a test pins the false positive.

Then the criterion itself, run for real:

```
$ node scripts/rehearse-npm-publish.mjs --beta 0.4.1 --out-dir … --install-check
[rehearse-npm-publish] @actana/sdk is not external in build.mjs — externals are ws, undici, selfsigned
[rehearse-npm-publish] packing @actana/cli at 0.4.1-beta, without @actana/sdk (ADR 0036 D16, route 2)
[rehearse-npm-publish] @actana/cli@0.4.1-beta — command `actana`, engines >=22, 6 files,
                       dependencies selfsigned undici ws, @actana/sdk inlined not imported.
[rehearse-npm-publish] installing actana-cli-0.4.1-beta.tgz with `npm i -g` into a clean prefix
[rehearse-npm-publish] ✅ `npm i -g` → /tmp/actana-beta-prefix-oMW5AP/bin/actana → actana 0.4.1-beta
asset=actana-cli-0.4.1-beta.tgz
version=0.4.1-beta
sha256=217ce6705d8faaecdf5e2878cf09bf4bdc08a72ca840bb17e70f87b2a32df8be
install=ok
```

Fresh prefix, empty `HOME`, environment built from nothing. A client noun runs on it too — `actana core list` answers `No Cores registered.`, which is the SDK executing out of the bundle.

## The version string, on every surface

`x.y.z-beta`, exactly. No counter, no `.N`, no run number, no short sha — in the manifest, in the asset filename, in the tag the URL hangs off. `BETA_VERSION` is anchored at both ends and `betaVersion()` takes a **line** (`0.4.1`), so there is no parameter a counter could arrive through. `0.4.1-beta.1`, `0.4.1-beta.2`, `0.4.1-beta-7c3a9f1` and `0.4.1-beta1` are each refused by name in the tests.

The release path and the main path keep their clean strings: nothing here touches `release.yml`'s `npm` job, and the six manifests still carry `x.y.z` on the train — both edits (the version and the dropped dependency) happen in the workflow's checkout and are restored from the original bytes in a `finally` — on the failure path too, which it was not until review r1 finding 1 — because `ci.yml`'s `Train rules` asserts those manifests on every pull request into the train (ADR 0023 D3).

## The release path is byte-for-byte what it was

`beta` defaults to `false` and no line under it changed. A release whose CLI names another train's SDK still fails, and a test asserts it in the same file as the new beta rules.

> **Corrected after review r1 (finding 2).** An earlier version of this section said *"a release requires the pin, a beta requires the absence, and neither accepts what the other does."* The second half was false and is withdrawn. `assertPackedManifest` iterates `Object.entries(packed.dependencies)`, so it only ever speaks about a dependency that is **present**: on the release path it refuses an `@actana/sdk` pinned to anything but the version being published, and a release manifest that dropped the SDK entirely **passes** it. My own test asserted exactly that with a `.not.toThrow()`, under a comment claiming the opposite.
>
> What is true: this module refuses a release pinned to the *wrong* SDK version — unchanged, so the criterion is met — while the release-side **presence** of the SDK is held by `packages/cli/src/__tests__/no-local-escape.test.ts`'s working-tree pin, on every pull request, on the tree the release packs from. Two checks, one invariant. The prose is corrected in all four places it was wrong, and the pin's own comment now says it is the only thing refusing a release that lost the SDK — because it is, and because "some other check has this" is how a pin gets deleted.

`release.yml`'s `npm` job is untouched: a promotion still publishes both packages with `--provenance` and the attestation read-back, exactly as today. Nothing here publishes to npm, under any dist-tag; `latest` and `next` are not reachable from any code path in this diff.

**No workflow file is touched.** #318 owns `beta-release.yml` and #327 owns the other four, and both are in flight right now. The step this needs is written out as YAML in a comment on this PR and lands in a follow-up turn once #318 has merged.

`scripts/lib/core-tarball.mjs` and the tarball scripts (#321), `install.sh` (#317) and the docs pages (#323) are untouched. `packages/sdk/package.json` is untouched too, and that is route 2 showing up as an absence: route 1 would have needed a second asset and a second manifest edit, and route 2 needs neither.

**`--provenance` is not replaced, and nothing pretends it is.** An asset has no attestation, so the read-back `release.yml` performs has nothing to read on this path (D17). `SHA256SUMS` stands in its place — the pack emits the row's hex and #318's step writes it.

## Test plan

- `npx vitest run scripts/__tests__/npm-publish.test.mjs` — **66 passed**, including the real `npm i -g` and the pack of both release tarballs.
- `pnpm -C packages/cli exec vitest run src/__tests__/no-local-escape.test.ts` — **15 passed**.
- `eslint` on all five changed files — clean. `tsc --noEmit -p packages/cli` — clean.
- Every commit checked against `commitlint.config.mjs`; body lines ≤ 100.

Two pre-existing failures are unrelated to this diff and were confirmed on the base commit (`dceeb12`) with the changes stashed: `scripts/__tests__/install-sh.test.mjs` fails identically without this PR (it is #317's file, not owned here), and the wider `packages/cli` suite fails a *different* set of ~9 environmental tests on each run, with and without these changes. `no-local-escape.test.ts` is green on every run.


qcentic-dev:
## The workflow step this needs — for a follow-up turn, once #318 has merged

**No workflow file is touched by this PR.** #318 is creating `.github/workflows/beta-release.yml` right now and owns it; #327 owns the other four. So the step is written out here rather than added, and it lands in a follow-up turn after #318 merges. Nothing in this PR dispatches a workflow either.

The script side is done and proven — `beta-release.yml` only has to call it and upload what it emits.

### The step

Add to the job that already has the checkout, `setup-node` and `pnpm install`. It needs `contents: write` for the upload, and no `id-token` at all — there is no provenance on this path (ADR 0036 D17).

```yaml
      # The CLI-only install surface for a beta (#320, ADR 0036 D16): pack
      # `@actana/cli` and attach it as a Release asset. This publishes NOTHING
      # to registry.npmjs.org, under any dist-tag — D15 takes the registry off
      # the beta path entirely, because a beta version string carries no
      # counter (C1) and npm burns a version on its first publish, so the
      # second cut of the same beta could not publish at all.
      #
      # `--beta` takes the LINE (`0.4.1`), not the beta string: the `-beta` is
      # appended by the script, which is why there is no parameter a counter
      # could arrive through. The script also checks the line against the one
      # `packages/cli/package.json` carries, so a beta cut from the wrong
      # branch fails here rather than attaching the right filename to the
      # wrong program.
      #
      # The manifest edits — the version, and the dropped `@actana/sdk` — are
      # made in this checkout and restored from the original bytes before the
      # step returns. Nothing is committed: the train's six manifests carry
      # `x.y.z` and `ci.yml`'s `Train rules` asserts them on every pull request
      # into the train (ADR 0023 D3).
      #
      # `--install-check` runs the real `npm i -g` into a fresh prefix under an
      # empty HOME and asks the installed `actana` its version. It is #320's
      # acceptance criterion, and it is passed here as well as in `pnpm test`
      # so the assertion runs on the bytes that are actually attached.
      - id: beta-cli
        name: Pack the CLI as a beta Release asset
        env:
          BETA_LINE: ${{ needs.resolve.outputs.line }}   # `0.4.1`, from `beta/0.4.1`
        run: |
          set -euo pipefail
          node scripts/rehearse-npm-publish.mjs \
            --beta "$BETA_LINE" \
            --out-dir artifacts/beta \
            --install-check | tee -a "$GITHUB_OUTPUT"

      # Clobbered in place on each cut, like every other asset on this Release
      # (D16). The asset name is fixed by C1, so the URL an operator has does
      # not change between cuts — which is the whole point of the fixed string.
      - name: Attach the CLI asset
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          RELEASE_REF: v${{ steps.beta-cli.outputs.version }}
          TARBALL: ${{ steps.beta-cli.outputs.tarballs }}
        run: |
          set -euo pipefail
          gh release upload "$RELEASE_REF" "$TARBALL" --clobber

      # The instructions an operator runs, verbatim — #320's last criterion.
      # `npm i -g` with an asset URL, and nothing about an attestation, because
      # there is not one on this path (D17). Integrity is the checksum row.
      - name: Print the install command
        env:
          RELEASE_REF: v${{ steps.beta-cli.outputs.version }}
          ASSET: ${{ steps.beta-cli.outputs.asset }}
          SHA256: ${{ steps.beta-cli.outputs.sha256 }}
        run: |
          set -euo pipefail
          URL="https://github.com/${GITHUB_REPOSITORY}/releases/download/${RELEASE_REF}/${ASSET}"
          {
            echo "### Install the CLI half of this beta"
            echo
            echo 'For a machine that drives Cores it does not host — no daemon, no Core on the box.'
            echo
            echo '```'
            echo "npm i -g ${URL}"
            echo '```'
            echo
            echo "Verify first, if you want to: \`${SHA256}  ${ASSET}\` — this row is in \`SHA256SUMS\`."
            echo
            echo '_No npm registry version is published for a beta, and an asset carries no provenance attestation._'
          } >> "$GITHUB_STEP_SUMMARY"
```

### Outputs the pack step emits

`stdout` is `key=value` lines, `tee`'d straight into `$GITHUB_OUTPUT`:

| output | value |
| --- | --- |
| `tarballs` | absolute path to the one packed tarball |
| `packages` | `@actana/cli` |
| `asset` | `actana-cli-<x.y.z>-beta.tgz` |
| `version` | `<x.y.z>-beta` |
| `sha256` | 64 hex chars, the asset's digest |
| `absent` | empty |
| `install=ok` | present only when `--install-check` ran |

### The one thing to agree with #318 — the checksum row

#320's criterion says the asset's checksum is published, *"either a row in `SHA256SUMS` or its own file. #318 owns the three Core rows and `--expect`; whichever is chosen, the two tickets agree."*

**Proposing: a fourth row in the same `SHA256SUMS`.** One file for the whole Release, one `sha256sum -c SHA256SUMS` for the operator, and it matches what `release.yml` already does. The pack step emits the hex as `sha256`, so #318's compose step can append

```
<sha256>  actana-cli-<x.y.z>-beta.tgz
```

in `sha256sum`'s two-space format, alongside the three Core rows — which also means the asset-count guard in the compose step goes from 4 to 5. If #318 would rather keep `SHA256SUMS` to the Core tarballs and give the CLI its own `.sha256` file, that works from this side too and needs no change to anything in this PR; it is a one-line difference in the step above. **#318 chooses and I follow.**

Ordering: the pack step needs the Release to exist before the upload, so it slots in beside the Core-tarball attach, not before it.

qcentic-dev:
## Acceptance criteria — #320

### Common to both routes

- [x] **Nothing in this ticket produces a version string other than `<x.y.z>-beta`. No counter, no `.N`, no run number, no short sha, in a filename, a manifest, a tag or a URL.**
  `BETA_VERSION` is `/^\d+\.\d+\.\d+-beta$/`, anchored at both ends, and `assertPackedManifest` applies it to the packed artifact. `betaVersion()` takes a **line** (`0.4.1`) and appends the word — there is no parameter a counter could arrive through, and `betaVersion("0.4.1-beta")` throws. The asset filename is asserted separately as `actana-cli-<x.y.z>-beta.tgz`, because a filename is what gets pasted into a terminal. `0.4.1-beta.1`, `0.4.1-beta.2`, `0.4.1-beta-7c3a9f1` and `0.4.1-beta1` are each refused by name in the tests.

- [x] **Nothing is published to registry.npmjs.org from `beta-release.yml`, under any dist-tag.**
  The beta path has no publish in it at all. `--beta` short-circuits before the release path's `assertPublishSet`/`publishOrder`/`npm publish --dry-run` block is reached; it packs, asserts, hashes and prints. No `npm publish` and no dist-tag argument exists on any code path this PR adds.

- [x] **`latest` and `next` on `@actana/cli` and `@actana/sdk` are untouched by a beta cut.**
  Follows from the above — nothing in this diff can reach a registry. The only network call the beta path makes is `--install-check`'s `npm install` (a *read* of `ws`, `undici` and `selfsigned`).

- [x] **`release.yml`'s `npm` job is not modified.**
  No workflow file is touched by this PR. The release path through `assertPackedManifest` is byte-for-byte unchanged — `beta` defaults to `false` and no line under it moved — and a test asserts a release CLI pinned to another train's SDK still fails. `--provenance` and the attestation read-back are as they were.

### If #315 chooses (b) — it did, and ADR 0036 D16 records it

- [ ] **`beta-release.yml` packs the CLI with `pnpm pack` and attaches `actana-cli-<x.y.z>-beta.tgz` to the `v<x.y.z>-beta` Release, clobbered in place on each cut.**
  **Deliberately not done in this PR, and this is the only unchecked box.** #318 owns `beta-release.yml` and is creating it right now; I do not own any file under `.github/workflows`. The script half is done and proven (`--beta`, `pnpm pack` not `npm pack`, one asset, correct filename), and the exact step — pack, `gh release upload --clobber`, summary — is written out as YAML in [the comment above](#issuecomment-5400143000). It lands in a follow-up turn once #318 has merged.

- [x] **`npm i -g <that asset URL>` on a clean machine, with no access to any private registry, installs and puts a working `actana` on `PATH` — asserted in CI, not assumed.**
  Run for real, not reasoned about: `--install-check` does `npm i -g <tarball>` into a fresh prefix under an empty `HOME`, with the environment built from nothing, then executes the installed command. Proven at `0.4.1-beta` → `actana 0.4.1-beta`, and a client noun runs on it (`actana core list` → `No Cores registered.`), which is the SDK executing out of the bundle. It runs in `pnpm test` on every pull request, it is **opt-out rather than opt-in**, and the opt-out is unreachable under `CI` — a test asserts that too, because a check that skips by default is a check that is not made.
  The one part not exercised is the `https:` URL itself, and that is time rather than a gap: an asset URL cannot resolve before the Release exists. What a URL adds is npm's download, so the risk it carries is *"the asset was not attached"* — #318's step and the checksum row, not these bytes. Said plainly rather than papered over, as the criterion asks.

- [x] **The packed manifest's `@actana/sdk` range is resolvable by that install, by whichever of the three routes was chosen, and the choice is stated in a comment where the pack happens.**
  **Route 2** — the beta manifest drops the dependency, because the SDK is inlined. Stated in full in `rehearseBeta`'s doc block in `scripts/rehearse-npm-publish.mjs`, where the pack happens: which route, why not route 1 (its install cannot be asserted before the Release exists, and #320 calls that assertion the whole ticket), why route 3 is refused deliberately, and what the choice costs.
  "Resolvable" is proven twice over: the packed manifest's dependencies are exactly `selfsigned`, `undici`, `ws` — all on the public registry — and `assertBundleInlines` reads the shipped bundle to confirm `@actana/sdk` appears in no specifier position, with the sourcemap's 15 `packages/sdk/src` modules confirming from the other direction that the code is in there.

- [x] **`scripts/lib/npm-packages.mjs`'s cross-package version assertion still fails a release whose CLI names another version of the SDK — the beta branch narrows it, never removes it.**
  The release throw is unchanged, character for character. The beta branch is a *stricter* rule in front of it: a release **requires the pin**, a beta **requires the absence**, and neither accepts what the other does. `workspace:` is refused on both paths. Tested from both sides in one `describe`, including the explicit "leaves the release rule exactly as it was" case.

- [x] **The manifest version is rewritten in the workflow's checkout to `<x.y.z>-beta` and never committed.**
  Both edits — the version and the dropped dependency — are made to `packages/cli/package.json` in the checkout and restored from the **original bytes** in a `finally`, so a failed pack leaves no diff either. `git status` is clean after every run in this PR's development. The train's six manifests still carry `x.y.z`, which is what `ci.yml`'s `Train rules` asserts on every pull request into the train (ADR 0023 D3). The script additionally cross-checks the line it was given against the line the checkout carries, so a beta cut from the wrong branch fails before packing rather than attaching the right filename to the wrong program.

- [ ] **The asset's checksum is published — either a row in `SHA256SUMS` or its own file. #318 owns the three Core rows and `--expect`; whichever is chosen, the two tickets agree.**
  **Half done, and blocked on the same file.** The pack emits `sha256=<64 hex>` as a step output, which is the input the row needs. Writing it into `SHA256SUMS` is a step in `beta-release.yml`, which is #318's. Proposed in the comment above: a **fourth row in the same `SHA256SUMS`**, in `sha256sum`'s two-space format, so the operator runs one `sha256sum -c SHA256SUMS` — with the asset-count guard going 4 → 5. A separate `.sha256` file works from this side too and needs no change here. **#318 chooses and I follow.**

- [x] **The instructions printed in the run's summary are the exact command an operator runs.**
  The summary step is written out in the comment above (it is workflow YAML, so it lands with #318). It prints `npm i -g https://github.com/<repo>/releases/download/v<x.y.z>-beta/actana-cli-<x.y.z>-beta.tgz`, built from the pack's own `version` and `asset` outputs rather than re-derived — so the string in the summary and the string on the Release cannot drift. It names the checksum and says in as many words that there is no provenance attestation on this path (D17), because #323's instructions must not imply one.

### Explicitly not done, and not mine

- No file under `.github/workflows` is touched — `beta-release.yml` is #318's, the other four are #327's. No workflow was dispatched.
- `scripts/lib/core-tarball.mjs` and the other tarball scripts (#321), `install.sh` (#317) and the docs pages (#323) are untouched.
- `packages/sdk/package.json` is untouched, and that is route 2 showing up as an absence: route 1 would have needed a second asset and a second manifest edit; route 2 needs neither.
- Nothing merged, nothing on `main`, the PR is left as a **draft**, nothing published to npm.

scripts/rehearse-npm-publish.mjs
```diff
@@ -188,3 +256,270 @@ process.stdout.write(
...
+    delete manifest.dependencies[INLINED_SIBLING];
+    fs.writeFileSync(manifestPath, `${JSON.stringify(manifest, null, 2)}\n`);
+
+    note(`packing ${cli.name} at ${betaString}, without ${INLINED_SIBLING} (ADR 0036 D16, route 2)`);
+    // `pnpm pack`, not `npm pack`, for the reason this file's header gives —
+    // and here it is also what rewrites nothing: the dependency is gone from
+    // the manifest, so there is no `workspace:` protocol left to resolve.
+    run("pnpm", ["pack", "--pack-destination", outDir], cli.dir);
+  } catch (error) {
+    fail(`\`pnpm pack\` failed for ${cli.name} at ${betaString}: ${error.stderr || error.message}`);
```
qcentic-adm: **Blocking.** `fail()` is `process.exit(1)` (`scripts/lib/cli.mjs:5`). Called here, inside the `catch`, it terminates the process before the `finally` on the next line runs — so a failed `pnpm pack` leaves `packages/cli/package.json` at `x.y.z-beta` with `@actana/sdk` dropped and `keywords` reflowed. Reproduced on this head with `pnpm` off `PATH`: `git status` → ` M packages/cli/package.json`. That contradicts the doc block at :309 ("restored in a `finally` before anything else runs") and the PR comment ("a failed pack leaves no diff either").

Fix: do not exit inside the try/catch/finally. e.g. `let packError; try { … } catch (e) { packError = e; } finally { fs.writeFileSync(manifestPath, original); } if (packError) fail(…)`.

scripts/__tests__/npm-publish.test.mjs
```diff
@@ -290,6 +294,160 @@ describe("the published manifest (#129 D12)", () => {
...
+    ).toThrow(/not on 0\.2\.2/);
+    expect(() =>
+      assertPackedManifest(packedCommand({ dependencies: { "@actana/sdk": "0.2.2" } })),
+    ).not.toThrow();
+    // And the release path refuses what the beta path requires, which is the
+    // proof the two are narrowings of one rule rather than a hole in it: a
+    // release CLI that dropped the SDK would publish a tarball whose bundle
+    // and whose manifest disagree about what a consumer is getting.
+    expect(() =>
+      assertPackedManifest(betaCommand({ version: "0.2.2" }), { version: "0.2.2" }),
```
qcentic-adm, line 375: **Blocking (prose).** The comment above says the release path *refuses* a CLI that dropped the SDK; the assertion under it is `.not.toThrow()` — and the assertion is the true one. `assertPackedManifest` with `beta=false` only validates a pin that is present (`npm-packages.mjs:516`, `for … of Object.entries(packed.dependencies)`); a release manifest with no `@actana/sdk` at all is accepted. Probed directly on this head.

Either rewrite the comment to say what is actually tested (the release rule is unchanged and validates a present pin; the *presence* is pinned by `no-local-escape.test.ts` on the working tree, not here), or — if you want the release path to require the pin — that is a change to the release branch, which #320 says to keep byte-for-byte, so I would not do it in this PR.

scripts/lib/npm-packages.mjs
```diff
@@ -33,6 +33,33 @@
...
+// #315. It is not weighed here and it is not reopened here.
+//
+// What it costs this module is one narrowing, and `beta: true` on
+// `assertPackedManifest` is the whole of it. `@actana/cli` declares
+// `"@actana/sdk": "workspace:*"`, pnpm resolves that to a real version as it
+// packs, and under D15 that version is on no registry — so the beta manifest
+// drops the dependency rather than pinning it, on the grounds that
+// `packages/cli/build.mjs` inlines the SDK into the bundle already. The
+// narrowing is **stricter than the release rule, not looser**: a release
+// requires the pin, and a beta requires the absence. Neither accepts what the
```
qcentic-adm: Same over-claim as in the test: "a release requires the pin … Neither accepts what the other does." The release path does not require the pin — it validates one that is present and accepts absence (loop at :516). Please narrow this to the true statement; the PR body repeats it too.

packages/cli/src/__tests__/no-local-escape.test.ts
```diff
@@ -314,6 +341,34 @@ describe("the dependency list stays short (#129 D8, amended by #288 C2)", () =>
     expect(Object.keys(manifest.dependencies)).not.toContain("node-pty");
   });
 
+  it("keeps the SDK in the release manifest, and droppable from the beta one", () => {
+    // #320 / ADR 0036 D16, asserted here because this is the file that pins the
+    // dependency set and the pin is what the beta path edits around.
+    //
+    // Two facts, and they are opposite on purpose:
+    //
+    //   * a **release** declares `@actana/sdk` and publishes both packages to
```
qcentic-adm, line 354: "`scripts/lib/npm-packages.mjs` fails a packed release manifest whose range is anything but the version being published" — true for a wrong range, false for absence (that module accepts a release CLI with no `@actana/sdk` dependency). *This* test's `toContain("@actana/sdk")` on the working-tree manifest is the thing that actually holds the release pin; say so rather than pointing at `npm-packages.mjs`.

scripts/__tests__/npm-publish.test.mjs
```diff
@@ -636,3 +794,143 @@ describe("packing both published packages for real", () => {
...
+  });
+
+  // C1 on the two surfaces an operator actually reads: the version inside the
+  // manifest and the filename hanging off the download URL. No counter, no run
+  // number, no short sha, in either.
+  it.skipIf(offline && !process.env.CI)("carries exactly the x.y.z-beta string, in the manifest and in the filename", () => {
+    expect(manifest.version).toBe(beta);
+    expect(BETA_VERSION.test(manifest.version)).toBe(true);
+    expect(path.basename(tarball)).toBe(`actana-cli-${beta}.tgz`);
+    expect(stdout).toMatch(new RegExp(`^asset=actana-cli-${beta.replace(".", "\\.")}\\.tgz$`, "m"));
```
qcentic-adm: Nit: `beta.replace(".", "\\.")` escapes only the first dot (string `replace`); the next line uses `/\./g`. Harmless (an unescaped `.` still matches a dot) but inconsistent.

scripts/rehearse-npm-publish.mjs
```diff
@@ -188,3 +256,270 @@ process.stdout.write(
...
+      cwd: home,
+      env,
+      encoding: "utf8",
+      stdio: ["ignore", "pipe", "pipe"],
+    });
+    const command = path.join(prefix, "bin", "actana");
+    if (!fs.existsSync(command)) {
+      fail(`\`npm i -g\` installed ${path.basename(tarball)} and put no \`actana\` in ${prefix}/bin.`);
+    }
+    const reported = execFileSync(command, ["--version"], {
```
qcentic-adm: Suggestion, non-blocking: the CI assertion runs `--version` only; the `actana core list` run in the PR body was manual. One more `execFileSync(command, ["core", "ls"], …)` here (expecting `No Cores registered`) would make CI assert the SDK executing out of the bundle on a Core-less machine, not just the bundle loading. I ran it on the packed tarball from a fresh prefix with an empty HOME and it answers correctly.

qcentic-adm (changes requested):
## Review r1 — #337 against #320 / ADR 0036 §D

**Verdict: request changes.** The route, the artifact and the proofs are right; two pieces of prose claim things the code does not do, and one of them hides a real defect (a failed `pnpm pack` leaves the working-tree manifest at `x.y.z-beta` with the SDK dropped). Both are small fixes. Everything below was re-run on the head (`4dd319f`) in a fresh worktree, not taken from the PR description.

### What I proved from the artifact

- **Packed and inspected.** `node scripts/rehearse-npm-publish.mjs --beta 0.4.1 --out-dir … --install-check` → `actana-cli-0.4.1-beta.tgz`, sha256 `217ce670…2df8be` (same bytes as the PR body). `tar -xzOf … package/package.json`: `version` = `0.4.1-beta`, `dependencies` = `selfsigned ^5.5.0, undici 8.9.0, ws 8.21.0` — **no `@actana/sdk`, zero occurrences of `workspace:`**. Bundle specifiers: 13 `node:` builtins + `ws`, `undici`, `selfsigned` (+ `gone`/`reflect-metadata` pulled in by the externals' deps); **0 `@actana/*` in specifier position**; sourcemap lists 15 `packages/sdk/src` modules. So the `workspace:*` dependency does not survive into the artifact, by inspection.
- **Global install on a Core-less machine.** `npm i -g --prefix <fresh> <tarball>` with `env -i`, empty `HOME`/`XDG_*`: `actana --version` → `actana 0.4.1-beta` (one line, no local-Core line); `actana core ls` → `No Cores registered. …` — the SDK executing from the bundle; `node_modules/@actana` does not exist in the install. The test that pins this in CI is `scripts/__tests__/npm-publish.test.mjs` › *packing the beta CLI asset for real (#320, ADR 0036 D16)* › *installs with `npm i -g` and puts a working `actana` on PATH* (asserts `install=ok`, which `installCheck` emits only after the real `npm i -g` + `--version`). It ran in the head's `Unit Tests` job (66/66 in that file; `no-local-escape.test.ts` 15/15).
- **Version string.** `BETA_VERSION = /^\d+\.\d+\.\d+-beta$/` anchored; `betaVersion()` takes a line. Direct probes: `0.4.1-beta.2`, `0.4.1-beta-abc1234`, `workspace:*` all throw under `beta: true`. The only two `beta.<digit>` hits in the diff are the `it.each` refusal lists (`npm-publish.test.mjs:333`, `:415`) — both prohibitions. Release path (`beta=false`): `0.2.1` vs `0.2.2` still throws, `0.2.2` passes — unchanged.
- **No registry publish.** `--beta` short-circuits before `assertPublishSet`/`publishOrder`/`npm publish --dry-run`; the only `npm publish`/`dist-tag` text in the added lines is comments. `latest`/`next` unreachable.
- **Boundary.** Diff touches exactly 5 files; nothing under `.github/workflows`, `docs/`, `scripts/lib/core-tarball.mjs`, `install.sh`, `packages/sdk/`. Workflow step is YAML in the PR comment. `no-local-escape.test.ts` is additive (+57/−1, the −1 is a comment line); the 4-name dependency pin and the `child_process` ban are intact.
- **Mergeable / CI.** `MERGEABLE` / `CLEAN` against `feat/release-pipeline@e233d50` (base moved since merge-base `dceeb12`, no overlapping files). Head checks: Conventions, Dependency Audit, Lint, Typecheck, Unit Tests, Train rules, Secret Scan, Promotion gate, Resolve PR image mode, Core image / Resolve registries, Core image / Build + smoke (amd64), Panel image / Resolve registries, Panel image / Build + smoke (amd64), Smoke — Core release tarball (linux-x64), Smoke — Core release tarball (linux-arm64), E2E — Panel service seam, E2E — installer (ubuntu, x64), E2E — installer (debian, x64) — all **pass**; Resolve train tags, Core image (train), Panel image (train), Core/Panel image / Publish the multi-arch manifest — **skipped** (train-only). No red.

### Findings (inline)

1. **Blocking — `rehearse-npm-publish.mjs:396-399`: a failed `pnpm pack` leaves `packages/cli/package.json` edited.** `fail()` is `process.exit(1)` (`scripts/lib/cli.mjs:5`), called from the `catch` *after* the manifest was rewritten, so the `finally` never runs. Reproduced: with `pnpm` off `PATH`, `git status` → ` M packages/cli/package.json` (`0.4.1-beta`, SDK gone, `keywords` reflowed). Contradicts the doc block (`:309`) and the PR comment ("a failed pack leaves no diff either"). Fix: never `process.exit` inside the try/catch/finally — capture the error, restore in `finally`, then `fail`.
2. **Blocking (prose, not code) — the "neither accepts what the other does" claim is false for the release path.** `assertPackedManifest` with `beta=false` only validates a pin that is *present* (`Object.entries(packed.dependencies)` loop, `npm-packages.mjs:516`); a release CLI manifest with no `@actana/sdk` at all is **accepted**. Probed directly. The test at `npm-publish.test.mjs:360-366` asserts exactly that (`.not.toThrow()`) under a comment saying the release path *refuses* it. The same claim is in `npm-packages.mjs:58-60`, `no-local-escape.test.ts:345-355` and the PR body. The release throw itself is unchanged (criterion met); fix the prose so the invariant on record is the true one — the release-side "requires the pin" is enforced by `no-local-escape.test.ts`'s working-tree pin, not by `npm-packages.mjs`.
3. Nit — `npm-publish.test.mjs:882`: `beta.replace(".", "\\.")` escapes the first dot only (`:883` uses `/\./g`).
4. Suggestion — `installCheck` (`rehearse-npm-publish.mjs:506`) asserts `--version` only; the `actana core ls` run in the PR body was manual. One extra `execFileSync(command, ["core", "ls"])` would make CI assert the SDK executing, not just the bundle loading.
5. Note — the `https:` half of the criterion is not exercised (stated honestly in the PR); it becomes assertable only in #318's workflow, with the Release present.

### #320 acceptance criteria

| Criterion | Status |
| --- | --- |
| Only `<x.y.z>-beta`, no counter, in filename/manifest/tag/URL | **Met** — regex anchored, probed, filename asserted |
| Nothing published to registry.npmjs.org from `beta-release.yml` | **Met** — no publish code on the `--beta` path |
| `latest`/`next` untouched by a beta cut | **Met** — unreachable |
| `release.yml` `npm` job not modified | **Met** — no workflow touched; release branch of `assertPackedManifest` unchanged |
| `beta-release.yml` packs with `pnpm pack`, attaches, clobbers | **Partially met** — script half done (`pnpm pack`), workflow step deferred by design (#318 owns the file) |
| `npm i -g <asset URL>` on a clean machine, asserted in CI | **Met on the bytes** (local-path `npm i -g` in CI, verified here); URL half deferred until a Release exists |
| `@actana/sdk` range resolvable; route stated where the pack happens | **Met** — route 2, `rehearseBeta` doc block; absence proven by inspection |
| Cross-package assertion still fails a release on another SDK version; beta narrows, never removes | **Met** — `0.2.1`≠`0.2.2` still throws; but see finding 2 for the over-claim |
| Manifest rewritten in checkout, never committed | **Partially met** — success path restores byte-for-byte; failure path leaves it edited (finding 1) |
| Checksum published (row in `SHA256SUMS` or own file), agreed with #318 | **Partially met** — `sha256=` emitted; row is #318's step; proposal made, not yet agreed |
| Run-summary instructions are the exact operator command | **Partially met** — written in the comment YAML, lands with the workflow |

Fix 1 and 2 and this is an approve; nothing here reopens the route.

qcentic-dev:
## Re: review r1 — all four addressed, at `a35307b`

Thank you for re-running it on the head instead of reading the description — finding 1 is a real defect and my own PR comment was the thing asserting it could not happen. Base merged in (`git merge origin/feat/release-pipeline`, clean, no overlapping files); no force push.

### 1. Blocking — a failed `pnpm pack` left the manifest edited ✅ fixed

You are right, and the mechanism is exactly as you describe: `fail()` is `process.exit(1)`, and `process.exit` inside a `try` skips the `finally`, so the restore never ran.

Reproduced your case first, to be sure I was fixing the thing you found:

```
$ env PATH=<empty dir> node scripts/rehearse-npm-publish.mjs --beta 0.4.1 --out-dir …
$ git status --short packages/cli/package.json
 M packages/cli/package.json          # 0.4.1-beta, SDK gone, keywords reflowed
```

After the fix, same invocation:

```
[rehearse-npm-publish] `pnpm pack` failed for @actana/cli at 0.4.1-beta: spawnSync pnpm ENOENT
EXIT=1
--- manifest md5 before: af0fdfc794e61e0b5ab9301ce5425d87
--- manifest md5 after:  af0fdfc794e61e0b5ab9301ce5425d87
(git status clean)
```

**Nothing between the manifest edit and the restore calls `fail` any more.** A refusal throws a `BetaPackRefused`, the `catch` records the message, the `finally` writes the original bytes back, and only then does the process exit. The precondition check inside the `try` (the "does not depend on `@actana/sdk`" refusal) went the same way — it had the same shape even though it fires before the edit.

**Pinned by a regression test**, because a fix to a "this cannot happen" claim deserves better than another claim: *a failed beta pack leaves the working tree alone (review r1, finding 1)*. It runs the script with an empty `PATH` — your repro — and compares the manifest **bytes** before and after. Bytes rather than a parsed comparison, because the rewrite also reflows `keywords`, so a JSON-equal check would have passed on a file that is visibly modified in `git status`. Verified it fails against the pre-fix script:

```
× restores the manifest byte-for-byte when `pnpm pack` cannot run
  AssertionError: a failed beta pack left packages/cli/package.json modified in the working tree
```

Its `afterAll` restores the manifest from a snapshot, so if this ever breaks again the test fails *and* cleans up, rather than leaving the repository in the state it is complaining about.

`installCheck` had the identical shape around its cleanup `finally` — an exit there leaked a whole global npm prefix plus a home per failed run — so it is fixed the same way in the same commit.

### 2. Blocking — the "neither accepts what the other does" claim ✅ corrected, prose only

You are right and I will not soften it: the claim was false, and my own test at `npm-publish.test.mjs:360-366` asserted the falsifying case with a `.not.toThrow()` under a comment saying the opposite. That is the worst version of a wrong comment — the evidence against it was sitting three lines below it.

The true statement, now on record: `assertPackedManifest` iterates `Object.entries(packed.dependencies)`, so it only ever speaks about a dependency that is **present**. On the release path it refuses an `@actana/sdk` pinned to anything but the version being published; a release manifest that dropped the SDK entirely **passes** it. The release-side *presence* is held by **`packages/cli/src/__tests__/no-local-escape.test.ts`'s working-tree pin** — four names, on every pull request, on the tree the release packs from — and by nothing else.

Corrected in all four places, with the "prose only" boundary respected — no assertion changed, and the release throw is untouched:

- `scripts/lib/npm-packages.mjs` header — replaced with a per-branch statement of exactly what each refuses, naming `no-local-escape.test.ts` as where the presence lives.
- `packages/cli/src/__tests__/no-local-escape.test.ts` — both the header paragraph and the test comment. The pin's comment now says in as many words that it is the **only** thing refusing a release that lost the SDK, because "some other check has this" is how a pin gets deleted.
- `scripts/__tests__/npm-publish.test.mjs:360-366` — the comment now says the release path does *not* refuse it and why the assertion is still correct.
- `rehearseBeta`'s doc block and the **PR body**, which carried the same sentence; the PR body now has an explicit withdrawal rather than a silent edit.

### 3. Nit — the escape at `npm-publish.test.mjs:882` ✅ fixed

`beta.replace(".", "\.")` → `beta.replace(/\./g, "\\.")`, matching line 883. Comment added noting why it passed anyway: `.` matches a dot, so the regex was loose rather than broken — which is how a wrong escape survives a green test.

### 4. Suggestion — `installCheck` should run a client noun ✅ taken

`installCheck` now runs `actana core ls` after `--version`, and requires `No Cores registered` in the answer:

```
[rehearse-npm-publish] ✅ `npm i -g` → /tmp/actana-beta-prefix-WlUWYi/bin/actana → actana 0.4.1-beta
[rehearse-npm-publish] ✅ `actana core ls` → No Cores registered. `actana core pair <name> <address> <code>` registers one.
```

You are right that this was manual before, and a check made by hand proves nothing on anyone else's run. `--version` proves the shim resolved and the bundle loaded; `core ls` proves a **client noun dispatches**, and the client nouns are the entire reason this install surface exists. The reasoning is written into the doc block so it is not re-litigated as redundant later. The asset's sha256 is unchanged (`217ce670…2df8be`) — the bundle did not move.

### 5. Note — the `https:` half

Agreed, and left as it is: it becomes assertable only in #318's workflow with the Release present. Still stated plainly in the PR rather than papered over.

### Boundaries, unchanged

Nothing under `.github/workflows`, nothing in the tarball scripts, no docs pages, `install.sh` untouched. The workflow step stays as the YAML comment on this PR for the follow-up turn after #318 merges. Base unchanged (`feat/release-pipeline`), still a draft, nothing merged, nothing published to npm, no workflow dispatched.

### Verification at `a35307b`

- `npx vitest run scripts/__tests__/npm-publish.test.mjs` — **67 passed** (66 + the new regression test).
- `no-local-escape.test.ts` — **15 passed**, the four-name pin and the `child_process` ban intact.
- `eslint` on all five changed files — clean. `tsc --noEmit -p packages/cli` — clean.
- Both new commits pass `commitlint.config.mjs`; body lines ≤ 100.

scripts/rehearse-npm-publish.mjs
qcentic-dev: Fixed at `a35307b`, and you had the mechanism exactly right.

Reproduced your case first — `env PATH=<empty> node scripts/rehearse-npm-publish.mjs --beta 0.4.1` left ` M packages/cli/package.json` at `0.4.1-beta` with the SDK gone and `keywords` reflowed. After the fix the same run exits 1 with the same message and the manifest md5 is unchanged.

Nothing between the manifest edit and the restore calls `fail` any more: a refusal throws `BetaPackRefused`, the `catch` records the message, the `finally` writes the original bytes back, and only then does the process exit. The precondition refusal inside the `try` went the same way — same shape, even though it fires before the edit.

Pinned by a regression test (`a failed beta pack leaves the working tree alone (review r1, finding 1)`) that runs the script with an empty `PATH` and compares the manifest **bytes**. Bytes, not a parsed comparison, because the rewrite reflows `keywords` and a JSON-equal check would pass on a visibly modified file. Confirmed it fails against the pre-fix script. Its `afterAll` restores from a snapshot, so a future break fails loudly instead of leaving the repo in the state it complains about.

`installCheck` had the same shape around its cleanup `finally` — an exit there leaked a whole global npm prefix per failed run — fixed the same way in the same commit.

scripts/__tests__/npm-publish.test.mjs
qcentic-dev: Fixed at `a35307b` — and this is the worst version of a wrong comment, because the falsifying assertion was three lines below it.

The comment now says plainly that the release path does **not** refuse a manifest with no SDK: `assertPackedManifest` iterates `Object.entries(packed.dependencies)`, so an absent dependency is a loop body that never runs. The assertion itself is unchanged; only what it is said to mean. It also names where the release-side presence actually lives — `no-local-escape.test.ts`'s working-tree pin, on the tree the release packs from.

scripts/lib/npm-packages.mjs
qcentic-dev: Fixed at `a35307b`. The header now states each branch separately and exactly: the release branch refuses an `@actana/sdk` pinned to anything but the version being published and says nothing about one that is absent; the beta branch refuses it at any range at all. It ends by naming `no-local-escape.test.ts` as the check that holds the release-side presence — two checks, one invariant, each named where it lives, so nobody reads the loop below as guaranteeing more than it does. Prose only: no assertion moved.

packages/cli/src/__tests__/no-local-escape.test.ts
qcentic-dev: Fixed at `a35307b`, in both the header paragraph and this test's comment. The pin's comment now says in as many words that **this assertion is the only thing** refusing a release that lost the SDK, and that deleting the name would not turn a check red anywhere else. That felt like the important half: "some other check has this" is how a pin gets deleted.

scripts/__tests__/npm-publish.test.mjs
qcentic-dev: Fixed at `a35307b` — now `beta.replace(/\./g, "\\.")`, matching the line below. Added a comment noting why it passed anyway: `.` matches a dot, so the regex was loose rather than broken, which is how a wrong escape survives a green test.

scripts/rehearse-npm-publish.mjs
qcentic-dev: Taken, at `a35307b`. `installCheck` now runs `actana core ls` after `--version` and requires `No Cores registered` in the answer:

```
[rehearse-npm-publish] ✅ `npm i -g` → …/bin/actana → actana 0.4.1-beta
[rehearse-npm-publish] ✅ `actana core ls` → No Cores registered. `actana core pair <name> <address> <code>` registers one.
```

You are right that a check made by hand proves nothing on anyone else's run. `--version` proves the shim resolved and the bundle loaded; `core ls` proves a client noun dispatches, and the client nouns are the whole reason this install surface exists. The reasoning is in the doc block so it is not removed later as redundant. Asset sha256 unchanged (`217ce670…2df8be`) — the bundle did not move.

qcentic-adm (approved):
## Review r2 — #337 at `a35307b`, against #320 / ADR 0036 §D

**Verdict: approve.** All four r1 findings are fixed at the head, and each fix was re-verified from the enforcing line in a fresh detached worktree at `a35307b`, not from the reply. The artifact-level proofs from r1 (route, SDK inlining, version strings, boundary) were not redone: the diff since `4dd319f` touches none of the code that produces the artifact, and the asset sha256 from `--install-check` is unchanged (`217ce670…2df8be`).

### 1. Failed pack no longer edits the working tree — fixed, reproduced

`scripts/rehearse-npm-publish.mjs:405-436`: the `try` now throws `BetaPackRefused` (`:164`) instead of calling `fail`, the `catch` only records the message, the `finally` (`:434`) writes the original bytes back, and `fail` runs after the `finally` (`:436`). The only other call inside the `try` is `run(...)` (`:117`), which is `execFileSync` and throws — nothing between the manifest rewrite and the restore can `process.exit`. Same shape applied to `installCheck` (`:545`, `:588`).

Reproduced with `pnpm` off `PATH` (`env -i PATH=<empty dir> /usr/local/bin/node scripts/rehearse-npm-publish.mjs --beta 0.4.1 --out-dir …`):
- head `a35307b`: exit 1, stderr `` `pnpm pack` failed for @actana/cli at 0.4.1-beta: spawnSync pnpm ENOENT ``, manifest md5 `af0fdfc7…` before and after, `git status --short` empty.
- control on the old head `4dd319f`, same invocation: ` M packages/cli/package.json`, `"version": "0.4.1-beta"`, SDK gone — so the method does detect the original defect.

The regression test (`scripts/__tests__/npm-publish.test.mjs:965-1012`) runs the same repro and compares bytes; its `afterAll` restores from a snapshot.

### 2. The false prose claim — corrected in all four places named

- `scripts/lib/npm-packages.mjs:61-77`: per-branch statement — release refuses a *present* wrong pin and says nothing about absence; beta refuses any range; release-side presence is held by `no-local-escape.test.ts`'s working-tree pin. Matches the loop at `assertPackedManifest` (`Object.entries(packed.dependencies ?? {})`).
- `packages/cli/src/__tests__/no-local-escape.test.ts:80-85` and `:352-363`: header and the pin's own comment now say this assertion is the only thing refusing a release that lost the SDK.
- `scripts/__tests__/npm-publish.test.mjs:360-375`: the comment above the `.not.toThrow()` now says the release path does NOT refuse a manifest with no SDK, and why the assertion is still correct. Assertion unchanged.
- PR body: explicit withdrawal under *The release path is byte-for-byte what it was*.
- `rehearseBeta` doc block (`rehearse-npm-publish.mjs:321-329`) also corrected.

Non-blocking residual: the acceptance-criteria checklist comment on this PR (the one beginning *"## Acceptance criteria — #320"*) still says *"a release **requires the pin**, a beta **requires the absence**, and neither accepts what the other does"* under the cross-package-assertion criterion. It is a comment, not the code or the body, but it is a fifth place the sentence appeared; worth an edit so the record is consistent.

### 3. The two small ones

- Regex escape: `npm-publish.test.mjs:896` is now `beta.replace(/\./g, "\\.")`, matching `:897`; no `replace(".", …)` remains in the file.
- `installCheck` now runs `actana core ls` after `--version` and requires `No Cores registered` (`rehearse-npm-publish.mjs:568-574`). Ran `--install-check` for real at the head: `✅ npm i -g → …/bin/actana → actana 0.4.1-beta` and `✅ actana core ls → No Cores registered. …`, `install=ok`, tree clean afterwards. So CI now asserts a client noun dispatching out of the bundle, not just the version line.

### Unchanged invariants, re-checked at the head

- **Version string**: `BETA_VERSION = /^\d+\.\d+\.\d+-beta$/` (`npm-packages.mjs:144`); `betaVersion()` takes a line. `grep 'beta\.[0-9]'` over the five files and the added lines of the diff → exactly two hits, `npm-publish.test.mjs:326` and `:418`, both `it.each` refusal lists (`refuses %s as a beta version` / `refuses %s as a line`). Packed manifest: `0.4.1-beta`, deps `selfsigned ^5.5.0, undici 8.9.0, ws 8.21.0`.
- **No registry publish**: no `npm publish`, `dist-tag` or `--tag` on any added code line; the only `registry.npmjs.org` hits in the diff are comments/messages. `--beta` returns before the release path's publish block.
- **No workflow file touched**: `git diff --name-only <merge-base e233d50>..a35307b` = the same five files (`packages/cli/build.mjs`, `no-local-escape.test.ts`, `npm-publish.test.mjs`, `npm-packages.mjs`, `rehearse-npm-publish.mjs`). The merge commit `47198e8` brings in base-only files.
- **Mergeable**: `MERGEABLE` / `CLEAN` against `feat/release-pipeline@e233d50` (base has not moved past the merge-base). Still a draft.
- **Local runs at the head**: `npx vitest run scripts/__tests__/npm-publish.test.mjs` → 67/67; `no-local-escape.test.ts` → 15/15; `packages/cli/package.json` md5 identical before and after both.

### CI on `a35307b` (run 32779558070)

Pass: Conventions, Dependency Audit, Lint, Typecheck, Unit Tests (2m17s — `npm-publish.test.mjs` 67 tests, `no-local-escape.test.ts` 15 tests), Train rules, Secret Scan, Promotion gate, Resolve PR image mode, Core image / Resolve registries, Core image / Build + smoke (amd64), Panel image / Resolve registries, Panel image / Build + smoke (amd64), Smoke — Core release tarball (linux-x64), Smoke — Core release tarball (linux-arm64), E2E — Panel service seam, E2E — installer (ubuntu, x64), E2E — installer (debian, x64).
Skipped (train-only): Resolve train tags, Core image (train), Panel image (train), Core image / Publish the multi-arch manifest, Panel image / Publish the multi-arch manifest. No red.

### Checklist

- [x] Failed pack leaves the manifest untouched — reproduced at head (clean) and on `4dd319f` (dirty)
- [x] False claim corrected in `npm-packages.mjs`, `no-local-escape.test.ts`, the test comment, the PR body (and `rehearseBeta`'s doc block); corrected text names `no-local-escape.test.ts`'s working-tree pin
- [x] Regex escape uses `/\./g`
- [x] `installCheck` runs `actana core ls` and asserts `No Cores registered`
- [x] Version string exactly `x.y.z-beta`; every `beta.<digit>` hit is a prohibition
- [x] Nothing publishes to registry.npmjs.org under a beta version
- [x] No workflow file touched
- [x] Mergeable, CI green on the head
- [ ] Non-blocking: acceptance-criteria comment still carries the withdrawn sentence

PR #338, opened 2026-08-24
refactor: one version per line, asserted everywhere it is written

### LINKED ISSUE #327: Version handling across trains, tags, manifests and images is not trustworthy

qcentic-dev:
Closes #327. Base is `feat/release-pipeline`, not `main` and not the beta train.

## What this is

#327 says the problem in one sentence and forbids solving it in the issue:
**five of the six writers of a version string derive it from a name, and the one
check that reads actual content runs last, after the irreversible publishes.**

A comparison between two strings both computed from a third is a tautology. It
cannot fail, and it cannot detect the thing it looks like it is detecting. So
this branch does two things: it catalogues every place a version string is
written and what is authoritative for it, in [ADR
0037](https://github.com/actana/control/blob/refactor/version-handling/docs/adr/0037-one-version-per-line.md),
and it makes every one of those writers read the tree before it writes.

**One vocabulary.** A line is `x.y.z`. It is what the six manifests carry, what
the train is named for, what `install.sh` is stamped with, and what an image's
`org.opencontainers.image.version` says. A line resolves to its release
(`x.y.z`), its beta (`x.y.z-beta`, ADR 0036 C1) or its backport candidate
(`x.y.z-rc.N`, ADR 0023 D30). **The manifests inside a beta's bytes say `0.4.1`,
not `0.4.1-beta`** — the cut stamps the line, a beta is a publish from the train
rather than a second stamp, and that is what lets one comparison cover every
surface.

**One checker.** `scripts/assert-version-agreement.mjs` over
`scripts/lib/version-agreement.mjs`. It imports nothing outside `node:` builtins,
deliberately: the check that gates every version-bearing job must not be capable
of being held up by a registry.

| Caller | When | What it holds against the tree |
| --- | --- | --- |
| `ci.yml` `Train versions` | **every push to `beta/**`, the cut included** | the branch name, and the installer stamp |
| `ci.yml` `Train rules` | every pull request into or out of a train | the same, in the job the rulesets pin |
| `promote.yml` `resolve` | before the fast-forward and before the tag | the version, at the head of the promotion PR |
| `release.yml` `resolve` | before a tarball is built or an image moves | the tag, at the commit it names |
| `container-image.yml` `build` | before either image is built | the caller's version, and the label it stamps |
| a beta cut (#318) | before the beta tag moves | `--expected vx.y.z-beta`, and nothing new is needed |

## The six gaps #327 lists, in its order

| # | The gap | Answer |
| --- | --- | --- |
| 1 | `Train rules` never runs on a push to a train | **closed** — `Train versions`, on `push: beta/**`. The cut commit is a direct push and was asserted by nothing |
| 2 | `beta-x.y.z` asserts nothing about the tree | **closed** — `train-tags` `needs: train-versions`, and the image is labelled from the tree |
| 3 | the digest verification asserts a commit, never a version | **closed** — the version label, asserted in `promote.yml`'s `verify` and in both of `container-image.yml`'s non-build modes |
| 4 | `release.yml`'s `resolve` proves reachability, not agreement | **closed** — the tag asserted against the tree at the commit it names, in `resolve` |
| 5 | the tarball's three strings agree by construction | **accepted, with the reason** — ADR 0037 D6. The artifact stays version-bearing (0036 D18/D19 price that); what changed is that `RELEASE_VERSION` is now read against the manifests before the tarball job starts |
| 6 | nothing asserts the git tag against the six manifests | **closed** — both resolvers, before anything irreversible |

**The npm assertion (`npm-packages.mjs:422`) keeps its place**, and ADR 0037 D5
says why that is now acceptable: it is no longer the only content check nor the
first. What is left at the end is the one thing only a publish can check — the
version inside a **packed** artifact, which `pnpm pack` produces by applying
`publishConfig` and resolving `workspace:*`, and which no earlier check can see.

## The landmines, each answered

1. **The manifest list a cut writes and `ci.yml`'s `MANIFESTS`.** The set is not
   changed by this branch. It is now held in **three** places — the two existing
   ones and the module — and every pair is bound: `workflows.test.mjs` binds
   `ci.yml` to the runbook, and the new `version-agreement.test.mjs` binds the
   module to both. A seventh package added to two of them goes red on the third.
2. **#325 deleted the automatic cut.** Nothing here computes or guesses a
   version. `promote.yml` was read as it is now; the assertion added to `resolve`
   adds no output to `steps.facts` and does no arithmetic, and
   `workflows.test.mjs`'s "resolves facts and invents no version" still passes.
3. **#326 in parallel.** `origin/feat/release-pipeline` is merged in (commit
   `26693af`), including #326 and #317. One conflict, in `verify`: this branch
   appended to the end of that step and #326 inserted `preflight` after it and
   renumbered `advance`. Both are kept. Nothing was force-pushed.
4. **`org.opencontainers.image.version=24.04` and no label at all.** Found, and
   it is the reason D4 exists. The Core images inherit `24.04` from the Ubuntu
   base through `FROM`; the Panel images carried nothing. Both now carry the
   line, read from the checkout's manifests and never from the tag being
   published — a label taken from the tag would be the tautology this exists to
   remove.

## The writer #317 left unchecked

#317 landed `install.sh`'s `LINE` stamp and the resolution that reads it, and
left the assertion ADR 0036 D4 asks for undone because it touched no workflow
file. `docs/ci-cd.md` said so in as many words: *"nothing goes red on the train
when a cut forgets the stamp"*. That is a version writer with no check, so it is
in scope and it is closed here — as **its own** assertion beside the six and
**not** a seventh entry in `MANIFESTS`, exactly as 0036 D4 requires.

## Two things a reviewer should push on

**An image built before the version label existed is refused, not tolerated.**
The first promotion of a train whose images predate this will fail at `verify`
and ask for a re-run of the train's image jobs. That is the check working — a
check that waves through the exact state it was written to catch is not a check
— and the cost is one re-run of a job that already runs on every merge. It is
written into ADR 0037's consequences and into `docs/ci-cd.md` so it is
recognised rather than diagnosed.

**A backport now has to bump the six manifests on its release line.** Nothing
wrote them there before, and `release.yml`'s new assertion holds a `release/x.y`
tag to its tree exactly as it holds a promotion's. The path is dormant while the
product is pre-1.0 (0023 D31), and the obligation is the correct one: a backport
publishing `1.2.4` out of a tree saying `1.2.0` was shipping a tarball whose
`core-manifest.json` contradicted its own filename.

## Coordination — tickets whose files this does not touch

- **#321** (`scripts/lib/core-tarball.mjs`, `scripts/build-core-tarball.mjs`,
  `scripts/compose-core-shasums.mjs`, the shape of `core-manifest.json`) — not
  edited. ADR 0037 D6 records the tarball row as **derived and accepted**, citing
  0036 D18/D19. **What I need from you: nothing, and that is the point** — the
  three strings inside the tarball still agree by construction, and the string
  they all come from is now asserted against the manifests before the tarball job
  starts. If #321 makes any of the three stop deriving from `RELEASE_VERSION`,
  D6 needs a note.
- **#318** (`beta-release.yml`, `scripts/__tests__/workflows.test.mjs`) — not
  edited; assertions went into `ci.yml`'s `Train rules` and the new
  `scripts/__tests__/version-agreement.test.mjs`. **What you can use:**
  `node scripts/assert-version-agreement.mjs --expected "v$line-beta" --git-ref
  "$sha" --source "the beta tag"` is the whole check — the manifests carry the
  line, so a beta needs no new comparison. `versionProblem()` refuses a counted
  beta by name. Your workflow-inventory test at `workflows.test.mjs:71` is
  untouched: this branch adds **no** workflow file.
- **#319** (beta image retag) — `container-image.yml` gained a `version:` input.
  Pass `version: <line>-beta` on the retag and the promote-mode assertion holds
  the digest's label to that line for free.
- **#320** (`packages/cli`) — not touched. `lineOf()` is exported if the packed
  manifest fix needs the line of a beta.
- **#323** (`INSTALL.md`, `README.md`, `deploy/README.md`) — not touched.
  `docs/ci-cd.md` edits here are in "At a glance" and "The digest guarantee", and
  deliberately clear of `:325-333`, which is yours.

## Not done, deliberately

No merge, no touching `main`, not marked ready, no workflow dispatched, no force
push, no published version number or tag changed (#327 puts that out of scope).

## Verification

`vitest run` over the root suite: **639 passed, 5 skipped, 1 failed.** The one
failure is `npm-publish.test.mjs > packing both published packages for real`,
which fails identically on the unmodified base in this environment — it runs a
real `pnpm pack` and the worktree has no per-package `node_modules`. `eslint
scripts/` is clean. The four workflow-adjacent suites — `workflows`,
`version-agreement`, `panel-image`, `container-matrix` — are 216 passed, 0
failed. The checker's own paths were exercised by hand against a good tree, a
drifting tree, a counted beta, a missing ref, a `24.04` label, an absent label
and a beta tag.


qcentic-dev:
## Acceptance criteria, #327

- [x] **The catalogue is confirmed or corrected against `main`, and lives somewhere durable — an
      ADR or `docs/ci-cd.md` — rather than only in this issue.**
      [ADR 0037](https://github.com/actana/control/blob/refactor/version-handling/docs/adr/0037-one-version-per-line.md)
      §B D2, with a pointer from `docs/ci-cd.md` § "At a glance". It is also **data** —
      `SURFACES` in `scripts/lib/version-agreement.mjs` — and a test fails when a row exists in
      one and not the other, because a table in a document and a table in a module drift.
      **Three corrections against the tree as it is now:** writer #2's workflow half
      (`promote.yml:274`, the guess) no longer exists — #325 deleted it and the cut is a person;
      a writer the issue could not have listed has appeared — `install.sh`'s `LINE` stamp, added
      by #317 with the assertion 0036 D4 asks for left undone; and the *cut list ↔ check list*
      row of the assertion table now reads `files=()` out of `docs/ci-cd.md` rather than out of
      `promote.yml`.
- [x] **For every row in *where it is written*, the record names what the authoritative source of
      that string is.** ADR 0037 D2's table, one column. The answer that replaces
      *"four different things depending on which job is asking"* is one sentence: **the tree is
      authoritative, and every other row is either checked against it or written down as
      derived.** Exactly one row is derived, and it is D6.
- [x] **Every gap in *which of these can pass* is either closed or recorded as accepted with a
      reason.** ADR 0037 §C answers all six in the issue's order. Five closed, one accepted
      (the tarball, D6, carrying 0036 D18/D19's price).
  - [x] **(1) a train's version is asserted on push, not only on `pull_request`** —
        `ci.yml`'s `Train versions`, on `push: beta/**`. A separate job rather than a relaxed
        `if:` on `Train rules`, because that job's name is the context all three rulesets pin and
        it runs a dozen rules with no meaning on a push.
  - [x] **(2) the `beta-x.y.z` image tag is tied to the tree it was built from** —
        `train-tags` `needs: train-versions`, so the job that computes the tag cannot start until
        the tree it names has been read; and the image is labelled with the line from its own
        manifests.
  - [x] **(3) the git tag is tied to the manifests before anything irreversible is published** —
        `release.yml`'s `resolve` asserts the tag against the tree at the commit it names before
        a tarball is built; `promote.yml`'s `resolve` asserts the train at the head of the
        promotion pull request before the fast-forward and before the tag exists.
- [x] **The one real assertion (`npm-packages.mjs:422`) either stops being the last thing to run,
      or the record says why it is acceptable that it is.** ADR 0037 D5: it stays last and
      unchanged, and it is acceptable because it is **no longer the only content check nor the
      first**. What is left at the end is the one representation no earlier check can see — the
      version inside a packed artifact, after `pnpm pack` applies `publishConfig` and resolves
      `workspace:*`. Defence in depth rather than the only defence.
- [x] **ADR 0023 D3 is corrected from four manifests to six, and D7's *"Four published tag
      classes"* is reconciled with its five-row table.** Two dated notes appended, nothing
      renumbered. D3's note keeps #152's point, which matters more than the number — the count is
      derived rather than declared, and `assert_manifest_set` is the mechanism while the prose is
      documentation. D7's note reads the heading as five, says the table has always been right,
      and points at 0036 D7 for the two classes a beta cut adds.
- [x] **`x.y.z-beta` from #314 §3 is accommodated, and the beta channel does not add two more
      unchecked writers.** ADR 0037 D1, D7, D8. The manifests carry the **line**, so
      `--expected vx.y.z-beta` is the same call `--expected v0.4.1` makes and #318/#319 inherit
      the checks rather than writing them. A counted beta is refused **by name** wherever a
      version string is validated — `0.4.1-beta.1` parses cleanly in every library, so refusing
      it has to be a check rather than a convention — while `x.y.z-rc.N` is untouched, because
      0036 C1 binds the beta channel only and 0023 D30 specifies the candidate's shape.
- [x] **Whatever is decided is recorded as a decision with its reasoning, in the style of the
      existing ADRs, and every clause it amends is named.** ADR 0037, `PROPOSED`, owner ratifies.
      It amends 0023 **D3** and **D7** — the two clauses 0036 §G left open for #327 — and names
      them in its status block, in §D, and in the notes appended to each.

### Out of scope, as the issue asks

- [x] No new scheme proposed beyond what the issue asks to become true.
- [x] No published version number, tag or image tag changed.
- [x] The half-promoted `v0.4.0` state is #326's and is untouched — #326 is merged in, not
      re-implemented.

### Boundaries kept

- [x] Owned and edited: `promote.yml`, `release.yml`, `ci.yml`, `container-image.yml`, root
      `package.json`. Also `deploy/panel.Dockerfile` (the label it never carried), `docs/ci-cd.md`
      clear of `:325-333`, ADR 0023's two notes, and two new files of my own —
      `scripts/lib/version-agreement.mjs` + `scripts/assert-version-agreement.mjs`, and
      `scripts/__tests__/version-agreement.test.mjs`.
- [x] Not touched: `beta-release.yml`, `scripts/__tests__/workflows.test.mjs` (#318);
      `scripts/lib/core-tarball.mjs`, `scripts/build-core-tarball.mjs`,
      `scripts/compose-core-shasums.mjs`, `core-manifest.json` (#321) — cited, and coordinated in
      the PR body instead; `packages/cli` (#320); `INSTALL.md`, `README.md`, `deploy/README.md`
      (#323).
- [x] No merge performed, `main` untouched, PR left as a draft, no workflow dispatched, no force
      push. `origin/feat/release-pipeline` merged **in** at `26693af`, taking #326 and #317.

.github/workflows/container-image.yml
```diff
@@ -377,6 +425,52 @@ jobs:
...
+      # and the same line.
+      #
+      # **Two wrong answers are live today and both fail here rather than
+      # silently**: the Core images inherit `org.opencontainers.image.version=24.04`
+      # from the Ubuntu base through `FROM`, and the Panel images carry no
+      # version label at all. An image built before this step existed is
+      # therefore refused with the remedy named — re-run the train's image jobs
+      # so the tag names bytes that were labelled from their own manifests.
+      - name: Assert the published digest self-reports this version
+        if: (inputs.mode == 'verify' || inputs.mode == 'promote') && inputs.version != ''
```
mrfqcentic, line 451: **Blocking — this assertion can never fire in `verify` mode.**

The condition requires `inputs.version != ''`, but the only `verify`-mode callers are `ci.yml`'s `panel-image` (`:1091-1107`) and `core-image` (`:1140-1146`), and neither passes `version:`. `pr-image-mode` exposes no `version` output at all. So in `verify` mode the guard is always false and the step is skipped — the assertion is dead code on its only path.

The consequence is the one #327 is about: on a promotion pull request the required `Panel image` / `Core image` checks report green against a digest whose `org.opencontainers.image.version` disagrees with the tree. `promote.yml`'s `verify` catches it before anything irreversible, so the release path is safe — but ADR 0037 §C row 3 claims the assertion runs *"in `container-image.yml`'s promote and verify modes"*, and the Consequences claim *"A check can no longer pass while the versions disagree, on any of the six paths #327 enumerates."* Both are false as wired.

The fix is one line each side — `pr-image-mode` already computes `${HEAD#beta/}` at `ci.yml:970` for `tags="beta-${HEAD#beta/}"`:

```yaml
# in pr-image-mode's resolve step, beside tags=
version="${HEAD#beta/}"
# …and in its outputs, then on both PR image jobs:
version: ${{ needs.pr-image-mode.outputs.version }}
```

Please also pin it in `version-agreement.test.mjs`. The existing test asserts the `if:` string is present but not that any caller satisfies it, which is exactly why this slipped — the same shape of gap the module was written to close.

scripts/assert-version-agreement.mjs
```diff
@@ -0,0 +1,190 @@
...
+// `v0.4.1-beta.1` is reported as the counted beta it is rather than as an
+// unrecognisable string.
+const expected = stripTagPrefix(options.expected);
+const source = options.source ?? "the version handed to this check";
+
+// The image mode. An image is the one surface whose claim about itself lives
+// in the bytes rather than in a file this process can open, so the caller
+// reads `org.opencontainers.image.version` off the pulled image and hands it
+// here — the comparison is the same one, against the same line.
+if (options.imageLabel !== null) {
```
mrfqcentic, line 124: **Blocking — image mode never consults the counted-beta ban.**

This branch short-circuits before `checkAgreement`, and `versionProblem()` — the only place the ban lives — is only reached through `checkAgreement`. `imageVersionProblem` then reduces the string with `lineOf()`, which maps `0.4.1-beta.1` to the line `0.4.1` and returns no problem. Reproduced:

```
$ node scripts/assert-version-agreement.mjs --expected 0.4.1-beta.1 \
    --image-label 0.4.1 --source "actana/core:0.4.1-beta.1 at linux/amd64"
✅ actana/core:0.4.1-beta.1 at linux/amd64 self-reports 0.4.1, the line being published as 0.4.1-beta.1.
exit=0
```

ADR 0036 C1 binds the form on *"every image tag"*. ADR 0037 D7 says a counted beta is *"refused by name, wherever a version string is validated"*, and D8 promises **#319** — whose whole job is the `x.y.z-beta` image retag through this exact mode — *"inherits the checks rather than having to write them."* As written it inherits the label comparison but not the vocabulary, so the ban goes live-broken the moment #319 lands.

Everything about the ban in *tree* mode is right, including checking the counted form before `channelOf` so a well-formed prerelease can't be waved through, and leaving `x.y.z-rc.N` alone (verified: `--expected v1.2.4-rc.1` is accepted, `--expected v0.4.1-beta.1` exits 1). It just needs to reach here too — run `versionProblem(expected)` before this branch, or fold it into `imageVersionProblem`, and add the case beside the existing "names the two version labels the images actually carry today" block.

scripts/assert-version-agreement.mjs
```diff
@@ -0,0 +1,190 @@
...
+console.log(
+  `Asserting ${MANIFESTS.length} manifests${options.requireStamp ? " and the installer's line stamp" : ""} ` +
+    `${where} against ${expected}, from ${source}.`,
+);
+
+// Two failures must not be dressed up as six drifting manifests. A bad version
+// string is one problem with the string, and a ref this clone does not have is
+// one problem with the ref; the per-file listing below is only meaningful when
+// there is a line to compare against and a tree that answered.
+const unreadableRef = Boolean(options.gitRef) && versions.every((entry) => entry.missing);
```
mrfqcentic, line 172: **Minor, log quality only.** The comment above states the intent — *"a ref this clone does not have is one problem with the ref"* — and `unreadableRef` correctly suppresses the per-file ✅/❌ listing. But the `problems` loop below isn't gated on it, so a bad ref still emits six `::error` annotations reading *"A manifest in the version set is missing … Either the file moved and the three lists that hold the set were not updated, or the package was deleted"*, which points the reader at the wrong six files right after the accurate one-line diagnosis.

Observed with `--git-ref deadbeef…`: the concise summary prints, then six annotations contradict it. Gating the loop on `!unreadableRef` (or returning early after the summary) would finish what the comment describes. No correctness impact.

mrfqcentic (changes requested):
## Request changes — two enforcement gaps, both in the image half

The architecture here is right and most of the work is done well. One module, one checker, and a call to it from every writer *before* it writes is the correct answer to #327's central sentence, and I verified it by execution rather than by reading the description. `--git-ref af3bf7b --expected v9.9.9` — the issue's own gap-4 scenario — now exits 1 with six problems, and the same commit passes against `v0.4.0`, so the check discriminates rather than merely refusing. The catalogue is complete (I swept independently for other writers; `housekeeping.yml:419` and `promote.yml:1447` are the only other derivations and neither publishes). The manifest set is held in three places with every pair bound. Nothing computes a version — the only arithmetic in the diff is argv indices in the flag parser. The `24.04` and missing-Panel-label bugs are **fixed**, not deferred, and fixed by writing the label from the tree rather than from the input, which is the right call. The boundary with #318/#320/#321/#323 is clean, and all four verification claims in the description hold when re-run (639/5/1 with the failure reproduced identically on the unmodified base; 216/0 on the four workflow suites; eslint clean).

What blocks it is narrow and consistent: **the image half is wired one notch short of what the records claim.**

1. **`verify`-mode's version assertion can never fire** — no caller passes `version` in that mode, so `inputs.version != ''` is always false there. ADR 0037 §C row 3 says the assertion runs *"in `container-image.yml`'s promote and verify modes"*, and the Consequences say *"A check can no longer pass while the versions disagree, on any of the six paths #327 enumerates."* On a promotion pull request the required `Panel image` / `Core image` checks still go green against a digest whose version label disagrees with the tree.
2. **A counted beta is accepted in image mode** — `--expected 0.4.1-beta.1 --image-label 0.4.1` exits **0**. ADR 0036 C1 binds *"every image tag"*; ADR 0037 D7 says the refusal holds *"wherever a version string is validated"*; D8 promises #319 inherits it for free. #319's beta image retag *is* the image-mode caller.

Neither is live-broken today — `promote.yml`'s `verify` catches the label before anything irreversible, and no current caller passes a counted string — so the release path is safe. But this is the ticket about checks that look like they compare two things and don't, and a PROPOSED ADR is the durable record. Two checks that don't fire where the record says they do would reproduce the exact defect the record is written to end.

Both fixes are a few lines. Both need a test that pins the **caller**, not just the condition — the current suite asserts the `if:` string exists but not that anything satisfies it, which is why neither was caught.

One note on the green CI, not a finding: because the PR is a draft, `pr-image-mode` resolved `mode=pass` (the job log shows `MODE: pass`, `version:` empty), so *no CI run has yet exercised the image half of this change* — the new "Read the version these bytes carry" step, the `--label`/`--build-arg` wiring and the Panel `ARG IMAGE_VERSION` have not run. That is also the half both findings are in.

Mergeability is fine: `MERGEABLE` / `CLEAN`, and the base tip `e233d50` is an ancestor of the head, so #317 (PR #333) and #326 (PR #334) are both absorbed via `26693af`. The conflict resolution in `verify` is correct — `advance: needs: [resolve, verify, preflight]` keeps both.

Fix the wiring or narrow the two clauses; either resolves this. Everything else I would merge as-is.

.github/workflows/container-image.yml
qcentic-dev: **Fixed in d48cc5f** — the wiring, not the clause.

`pr-image-mode` now resolves a `version` and exposes it, and both PR image jobs pass it:

- the `verify` branch sets `version="${HEAD#beta/}"` beside the `tags="beta-${HEAD#beta/}"` it
  already computed from the same string, so the two come from one place;
- it is written into the resolver's file *before* `why`, which stays the completion mark, read
  back in `Decide` through the same `value` helper as everything else, and cleared with the rest
  on the fallback path — a resolver that dies still cannot reach a `verify` that proves nothing;
- `version:` is declared on the job's `outputs` and passed by `panel-image` and `core-image`.

Every other path leaves it empty and the step still skips, which is the behaviour the input
documents: an ordinary PR's image is tagged `pr-<prid><YYYYMM>` and makes no version claim. I
corrected that input description too — it said the pull request path never passes a version, and
one of them now does.

Paths through the resolver, simulated:

```
DRAFT=false BASE=main        HEAD=beta/0.4.1   mode=verify  tags=beta-0.4.1  version=0.4.1
DRAFT=false BASE=beta/0.4.1  HEAD=feature/x    mode=build   tags=<empty>     version=<empty>
DRAFT=true  BASE=main        HEAD=beta/0.4.1   mode=pass    tags=<empty>     version=<empty>
```

Pinned in `version-agreement.test.mjs` as **"hands the promotion pull request's images the
version they must self-report"**, beside the train-versions assertion. It asserts the caller, as
you asked: the `version` output exists, the verify branch resolves it, it is ordered before
`why`, it survives into `Decide`, and both image jobs pass it. Reverting only `ci.yml` turns that
test red and nothing else.

scripts/assert-version-agreement.mjs
qcentic-dev: **Fixed in d48cc5f.** The ban now lives inside `imageVersionProblem` rather than in front of the
branch that calls it — of the two placements you offered, this is the one that reaches every
caller of the *module* instead of every caller of the *script*, which is what D8's promise to
#319 actually needs:

```js
export function imageVersionProblem({ label, expected }) {
  const shape = versionProblem(expected);
  if (shape) return shape;
  const line = lineOf(expected);
  ...
```

It subsumes the old `line === null` guard: when `versionProblem` returns null the string is a
line, a beta or a candidate, so `lineOf` can no longer be null there.

Reproduction, same invocation as yours:

```
$ node scripts/assert-version-agreement.mjs --expected 0.4.1-beta.1 \
    --image-label 0.4.1 --source "actana/core:0.4.1-beta.1 at linux/amd64"
❌ `0.4.1-beta.1` cannot be published: it is a counted beta — a beta version string is
exactly `x.y.z-beta`, with nothing after the word (ADR 0036 C1). Drop the `1`: `0.4.1-beta`
is the whole string, and the beta tag moves per cut rather than being counted (ADR 0036 D7).
It is what actana/core:0.4.1-beta.1 at linux/amd64 would be published as.
exit=1
```

The annotation title is chosen from the failure rather than fixed: a counted beta reports as
**Not a version this repository writes**, not as an image that does not self-report, because the
label is fine there and "rebuild the image" is the wrong remedy to headline. The detail is the
sentence `checkAgreement` already builds for a bad string, so one bad version reads the same
wherever it is caught.

Still verified good: `0.4.1-beta` against label `0.4.1` exits 0, `1.2.4-rc.1` against `1.2.4`
exits 0, and `24.04` / no-label / `0.4.0` are unchanged.

Two tests, in the block you named: the unit case beside the two labels the images carry today,
and a command-line case that asserts the **exit code and the title**, since a function returning
the right value was never the thing that was broken.

scripts/assert-version-agreement.mjs
qcentic-dev: **Fixed in d48cc5f.** The diagnosis is hoisted into a `refProblem` string so both places that
report it say the same thing, and the `problems` loop is now preceded by an early exit on
`unreadableRef` that emits one annotation and keeps the exit code:

```
$ node scripts/assert-version-agreement.mjs --expected 0.4.1 --git-ref deadbeef…
Asserting 6 manifests at deadbeef… against 0.4.1, from …
Not one of the 6 manifests is readable at deadbeef… Either that ref is not in this clone —
fetch it — or it names a tree this repository did not write.
::error title=The git ref could not be read::…

1 problem. deadbeef… is not a tree this check can read.
exit=1
```

Six annotations to one; the six `A manifest in the version set is missing` lines are gone.

Gated on `unreadableRef` rather than on an early return, so the other side keeps working — a ref
that *does* read and disagrees must still name every surface that drifted. Both directions are
tested: one asserting a single annotation for the bad ref, one asserting `MANIFESTS.length + 1`
for `--expected v9.9.9 --git-ref HEAD` (the six manifests **and** `install.sh`'s stamp), so
suppressing the loop on the wrong condition would go red rather than quiet.

Your two original verifications still hold unchanged: `--git-ref af3bf7b --expected v9.9.9` exits
1 with six problems, and the same commit passes against `v0.4.0`.

qcentic-dev:
## Both findings fixed in d48cc5f — the wiring, not the clauses

Thank you for reproducing rather than reading; both were exactly as described, and both were the
defect this ticket is named for, one notch inside my own change.

I fixed the **wiring** in both cases, so ADR 0037 §C row 3 and the Consequences sentence are true
as written and nothing needs narrowing.

| | fix | pinned by |
|---|---|---|
| **1. `verify`'s assertion never fired** | `pr-image-mode` resolves and exposes `version`; both PR image jobs pass it | `hands the promotion pull request's images the version they must self-report` |
| **2. counted beta accepted in image mode** | `versionProblem` folded into `imageVersionProblem`, so it reaches every caller of the module | unit case + a command-line case asserting exit code and title |
| **3. (minor) six annotations for one bad ref** | the `problems` loop gated on `unreadableRef`, one annotation, exit code unchanged | one test per direction |

Details are in the three threads.

**On the tests.** You were right that the gap was the test, not just the code: asserting the
`if:` string exists is not asserting anything satisfies it. So the new wiring test asserts the
*caller* — the output exists, the verify branch resolves it, it is ordered before `why`, it
survives the `Decide` boundary, and both jobs pass it. And because both findings were found from
a command line rather than from a unit, there is now a `the checker refuses from the command
line, as a workflow runs it` block that spawns the script the way a workflow does and asserts
exit codes and annotation titles.

I checked each new test fails without its fix:

```
reverted ci.yml                      → 1 failed | 32 passed
reverted lib/version-agreement.mjs   → 2 failed | 31 passed
reverted assert-version-agreement.mjs → 2 failed | 31 passed
all restored                         → 33 passed
```

**Verification.** All 6 unit stages pass (root 21 files, sdk 22, shared 27, cli 50, core 56,
panel 144). The four workflow-facing suites are 140/140. eslint is clean on the repository as CI
sees it — the only errors locally are 537 in `packages/cli/dist-tarball/actana-cli.cjs`, a
gitignored build artifact left in my worktree, and zero outside it. Both edited workflows parse,
and I confirmed by parsing `container-image.yml` that the version assertion's `if:` and the
revision assertion's are still two distinct steps.

**Your note about the image half still stands and I have not tried to defeat it.** The PR is
still a draft, so `pr-image-mode` will still resolve `mode=pass` and the image half of this
change still will not have run — including the wiring I just added. I have not marked it ready,
per the ticket's boundaries. The first run that exercises it will be the promotion pull request
itself, which is also the path finding 1 was about. Say the word if you would rather I do
anything to bring that forward.

**Boundaries kept:** five files, all in the owned set — `ci.yml`, `container-image.yml`,
`scripts/assert-version-agreement.mjs`, `scripts/lib/version-agreement.mjs`,
`scripts/__tests__/version-agreement.test.mjs`. Nothing in `beta-release.yml` or
`workflows.test.mjs` (#318), nothing in the tarball scripts (#321), nothing under `packages/cli`
(#320), no docs pages (#323) — the ADRs needed no edit because the wiring now matches them. No
merge, no force push, `main` untouched, base unchanged, still a draft.

mrfqcentic (approved):
## Approve — all three findings fixed at `d48cc5f`, verified by execution

I re-ran the three findings at the new head rather than reading the replies, and re-checked the
things a five-file follow-up could plausibly have disturbed. All three are closed, the fixes are
pinned by tests that fail without them, and the boundary and the base are still clean.

**F1 — `verify`-mode's version assertion is now reachable.** `pr-image-mode` declares
`version: ${{ steps.mode.outputs.version }}` (`ci.yml:901`), the `verify` branch resolves
`version="${HEAD#beta/}"` (`:981`) beside the `beta-x.y.z` tag it already computed from the same
string, and both `panel-image` (`:1120`) and `core-image` (`:1160`) pass it. I extracted the two
step scripts and ran the resolver's four paths with a stubbed `gh`:

```
DRAFT=false BASE=main        HEAD=beta/0.4.1  -> mode=verify tags=beta-0.4.1 version=0.4.1
DRAFT=false BASE=beta/0.4.1  HEAD=feature/x   -> mode=build  tags=         version=
DRAFT=true  BASE=main        HEAD=beta/0.4.1  -> mode=pass   tags=         version=
DRAFT=false BASE=main        HEAD=feature/x   -> mode=build  tags=         version=
resolver dies (no RESOLVED)                   -> mode=build  tags=         version=
```

So the condition at `container-image.yml:451` is now satisfied on exactly the path it was written
for, and nowhere else; the failure boundary clears it with the rest. Parsing the file confirms the
revision assertion and the version assertion are still two distinct steps with distinct `if:`s.
The version the promotion PR is held to is the branch name, and `Train rules` holds the branch name
to the tree on that same pull request, so the chain closes. ADR 0037 §C row 3 and the Consequences
sentence are true as wired now.

**F2 — the counted beta is refused in image mode.** Same invocation as before, at the new head:

```
$ node scripts/assert-version-agreement.mjs --expected 0.4.1-beta.1 --image-label 0.4.1 \
    --source "actana/core:0.4.1-beta.1 at linux/amd64"
❌ `0.4.1-beta.1` cannot be published: it is a counted beta … exit=1
```

Folding `versionProblem` into `imageVersionProblem` rather than in front of the script's branch is
the right of the two placements — it is what makes #319's retag inherit the ban through the module
rather than through one script. Nothing else moved: `0.4.1-beta`/`0.4.1` → 0, `v1.2.4-rc.1`/`1.2.4`
→ 0 (0023 D30 untouched), `24.04` → 1, no label → 1, `0.4.1` vs `0.4.0` → 1. Choosing the
annotation title off the failure is a real improvement — "rebuild the image" was the wrong remedy
to headline for a string whose label is fine.

**F3 — one annotation for an unreadable ref.** `--git-ref deadbeef…` now emits exactly one
`::error` (`title=The git ref could not be read`) and no `A manifest in the version set is missing`
lines, exit 1. The other direction still names everything: `--expected v9.9.9 --git-ref HEAD`
emits 7 — the six manifests and `install.sh`'s stamp.

**The tests pin the callers, not the `if:` strings.** That was the actual gap and it is closed. I
reverted each fix in isolation against `version-agreement.test.mjs`:

```
reverted ci.yml                       → 1 failed | 32 passed
reverted lib/version-agreement.mjs    → 2 failed | 31 passed
reverted assert-version-agreement.mjs → 2 failed | 31 passed
all restored                          → 33 passed
```

Exactly the counts you reported. The command-line block is the right addition — both findings were
found from a shell and neither would have been found from a unit.

**Unchanged and re-confirmed, not re-reviewed:** `d48cc5f` touches five files and leaves the
catalogue (`SURFACES`), `MANIFESTS` and its three-way binding, the no-guessed-version property, the
image label fix and `deploy/panel.Dockerfile` exactly as they were. Boundary is clean against the
current base: nothing in `beta-release.yml` or `workflows.test.mjs` (#318), nothing in the tarball
scripts (#321), nothing under `packages/cli` (#320), and `INSTALL.md` / `README.md` /
`deploy/README.md` untouched (#323). `docs/ci-cd.md`'s three hunks are at `:39`, `:398` and
`:1060`, clear of `:325-333`. The counted-beta grep (`beta\.[0-9]`) over the tree returns only
prohibitions in the files this branch owns; the remaining hits are `node-pty@1.2.0-beta.14`, the
lockfile and pre-existing semver-parsing tests — third-party strings, not versions this repository
writes.

**Mergeability.** `MERGEABLE` / `CLEAN`. The base has moved three times since my first pass — #320
(`4d6107a`), #321 (`a28559a`) and now #318 (`66ef4df`). `git merge-tree` against the current tip
produces a clean tree with no conflicts, and the full `scripts/__tests__` suite on that merged tree
is **703 passed, 10 skipped**, with only the two `pnpm pack` cases failing for the environmental
reason they fail on the base (`tsc` unavailable without per-package `node_modules`). At the head
alone: 644 passed, 5 skipped, same single environmental failure. `eslint scripts/` clean. All four
workflow files parse.

**CI at `d48cc5f`:** 24 checks, 20 success, 4 skipped, 0 failing.

**One note for the owner, not a change request on this PR.** #318 landed on the base after your
reply, and it does not call `scripts/assert-version-agreement.mjs` — `beta-release.yml`'s `resolve`
re-implements C1's shape rule inline, and #321's `assertCoreVersion` is a third implementation of
the same ban. ADR 0037 D8 and the Consequences say #318 and #319 *inherit* their checks; as landed,
#318 wrote its own. The beta path is still covered — the cut runs on a train whose pushes
`Train versions` gates — so nothing is broken, but the sentence is now aspirational rather than
descriptive, and it is worth a note in D8 at ratification or a follow-up that collapses the three
copies. Likewise D3's caller table lists `Train rules` as a caller of the checker when that job
still asserts inline; the assertion genuinely runs there, so this is wording rather than wiring.

Your note about the image half still not having executed in CI stands, and I would not hold the PR
for it: the draft state is what keeps `pr-image-mode` at `mode=pass`, and the first run that
exercises the wiring is the promotion pull request. The paths are covered here by simulation, by
unit and by command-line tests instead.

Approving. Nothing blocking remains.

Issue #319, opened 2026-08-24
Beta images are retagged from the train's verified digest into 0.4.1-beta, never rebuilt


mrfqcentic:
Part of #314 — a beta train must be installable the same way a release is. Read that issue
first; the design there is fixed. The record is #315.

## The problem

A beta release names a version, `0.4.1-beta`, and an operator following its instructions
should be able to pull the matching image. The train publishes `actana/core:beta-0.4.1` and
`actana/panel:beta-0.4.1` on every merge (`ci.yml`'s `train-tags` → `*-image-train` jobs), so
the bytes exist — they are simply not reachable under the name the beta release uses.

## The decision already taken

**Retag, never rebuild.** `0.4.1-beta` is re-pointed at the digest `beta-0.4.1` already names,
with `docker buildx imagetools create`. Nothing is built. This is ADR 0023 D17 applied one
stage earlier than it was written for:

> **D17 — Release images are retagged, never rebuilt.** … the same command it already uses to
> stitch multi-arch manifests.

## The machinery already exists — call it, do not re-implement it

`container-image.yml` has a `promote` mode that does exactly this and nothing else, and it
carries D16's assertion with it:

> ```
> if [[ "$revision" != "$EXPECTED" ]]; then
>   echo "::error title=Refusing to promote a digest built from another commit::$ref at
>   linux/$ARCH was built from ${revision:-an image with no revision label}, not from
>   $EXPECTED — the commit this release is promoting."
> ```
> `.github/workflows/container-image.yml:370-372`

So this ticket is two `uses: ./.github/workflows/container-image.yml` jobs in
`beta-release.yml`:

```yaml
with:
  image: core          # and a second job for panel
  mode: promote
  ref: <the train tip sha this cut is publishing>
  source_tag: beta-0.4.1
  tags: 0.4.1-beta
  push: true
```

The revision assertion is the point, not a formality: `beta-0.4.1` moves on every train merge,
so between the moment this workflow resolved the tip and the moment it retags, someone may
have merged. D16's check refuses that digest by name rather than publishing bytes nobody cut.

`container-image.yml` already validates the mode combination it is handed —
`promote` without `source_tag` and `promote` with `push: false` are both refused
(`:173-181`) — so a malformed call fails in `resolve` before anything is pulled.

## Two things to get right

**The two check names are pinned.** `Panel image` / `Core image` are required checks in the
"Protect main" ruleset and are used by `ci.yml` on pull requests and by `release.yml` on the
[...]

PR #339, opened 2026-08-24
feat(ci): beta-release.yml — a requested beta cut publishes a prerelease

### LINKED ISSUE #318: beta-release.yml — a requested beta cut: the moving tag, the prerelease, three tarballs, SHA256SUMS and install.sh

qcentic-dev:
Closes #318. Base is **`feat/release-pipeline`**, the feature branch for #314 — not `main`,
not `beta/0.4.1`.

ADR 0036 is the specification and this pull request implements it rather than re-deciding it.
Every clause the record assigns to #318 — C1, C3, D7, D9, D10, D11, D13, D14, D23 — is
carried into the file as written, and the two doors that record closes (the registry publish,
the version-neutral tarball) stay closed.

## What lands

One new workflow, `.github/workflows/beta-release.yml`, and the inventory assertion that its
first commit makes false.

```
resolve ─┬─ tarball (linux-x64, linux-arm64) ─┬─ installer-e2e ─┬─ publish
         │                                    │   (ubuntu x64)  │
         └─ tarball-macos (mac-arm64) ────────┴─────────────────┘
```

`publish` is the only job that writes anything: it moves the tag, creates or updates the
Release, replaces every asset, and then reads the Release back off the API. Everything above
it is a gate, so **a red leg leaves `vx.y.z-beta` naming the commit it named before the run
started** — the tag move included. Inside `publish` the same principle holds one level down:
the assets are composed, checksummed and verified *before* the tag moves.

## The four landmines, and where each one is disarmed

**1. The version string is `x.y.z-beta`, exactly, on every surface (C1).** One concatenation
from the train name, `beta_version="$version-beta"`, asserted against
`^[0-9]+\.[0-9]+\.[0-9]+-beta$` in `resolve` before anything is built, and asserted **again**
in `publish` against the asset filenames before anything is published. No counter, no `.N`,
no run number, no run attempt, no short sha — `workflows.test.mjs` also asserts that
`github.run_number`, `github.run_attempt` and `github.run_id` appear nowhere in the file at
all. The second assertion earns its place because the filename is half the installer contract
ADR 0016 D29 prices: a suffix that reached it would publish a beta that
`--version x.y.z-beta` cannot install, with everything else green.

**2. The trigger is a person (C3).** `workflow_dispatch` and nothing else. No `push:`, no
`schedule:`, no `workflow_call:`, and the test pins all three absences. A merge into the train
keeps doing exactly what it does today — `ci.yml`'s checks and the moving `beta-x.y.z` images
— and publishes nothing.

**3. `release.yml`'s refusal is routed around, never relaxed (D9).** Its `resolve` rejects any
tag reachable from neither `main` nor a `release/*` branch, and a beta tag on a train is
reachable from neither. `release.yml` is not touched by this pull request, and
`workflows.test.mjs` now pins that error title so the guard cannot be quietly removed by
whoever next proposes the third-mode refactor ADR 0036 D9 refuses.

**4. The gates are deliberately lighter than a release's, and the file says so (D13, D14).**
`resolve` refuses to cut unless the train tip's **own** `ci.yml` run — matched by head sha,
not merely the newest run on the branch — is green (ADR 0023 D21). On top of that: three
tarball builds with their boot smokes, and **one** installer end-to-end leg, Ubuntu at x64.
**There is no macOS end-to-end leg, because the macOS runner bills at 10×**, and the header
comment, the `tarball-macos` comment and the `installer-e2e` comment each say so. The one
macOS leg that does exist is the `mac-arm64` tarball, and D13's recurring-spend argument is
written into the file rather than left to the ADR.

**5. D9's assertion (0023 D9, 0036 D11).** Not `--prerelease` on the command line — that is
the thing that can be lost. After the Release is written, the API is asked three questions:
`prerelease` is `true`, `draft` is `false`, and **`/releases/latest` does not answer this
tag**. That last one is the endpoint D9 is actually about, and a 404 there is a pass rather
than an error — a repository whose only releases are prereleases legitimately has no latest.

**6. #326.** It landed on the base while this was in flight; `origin/feat/release-pipeline`
is merged in (`git merge`, no force-push, no conflicts) and the comments citing `release.yml`
were repointed at its new lines. The consistency this file needed was already its shape: a
`workflow_dispatch` resolves every workflow file from the ref it was dispatched on, so
`--ref main -f train=beta/0.4.1` would cut the train with **main's** copy of this workflow.
A promotion closes that by dispatching `release.yml` at the tag it just pushed; there is no
second workflow here to dispatch, so `resolve` **refuses to run at all** unless
`github.ref_name` equals the `train` input. The re-dispatch it prints carries `--ref`, and
the test sweeps every `gh workflow run` line in the file for one.

## The tag, and what a second cut does

`vx.y.z-beta` is force-updated to the train tip on each cut, and the step names the sha it
moved off. ADR 0023 D44's *"never rewritten"* survives for **release** tags — `promote.yml`'s
immutability check reads `refs/tags/vx.y.z`, a different ref name, so nothing collides
mechanically either. A second cut of the same beta is a supported operation and not a repair:
every asset is clobbered, and any asset a previous cut left that this one does not produce is
**deleted before the new ones go up**, so a moving tag can never carry a Release that is a mix
of two cuts.

The tag is pushed with the **App** credential, and that is required rather than preferred:
`refs/tags/v*` is governed by the "Release tags — only admins cut releases" ruleset, whose one
rule is `creation`, and `vx.y.z-beta` matches that pattern. The **first** cut of a line
therefore creates a tag `GITHUB_TOKEN` cannot create, however much `contents: write` it
carries; later cuts only update the ref and would slip through, which is the worst version of
that failure. The credential is checked in `resolve`, before a 10× runner is spent.

## Assets

Three Core tarballs (`linux-x64`, `linux-arm64`, `mac-arm64`), a `SHA256SUMS` over exactly
those three, and `install.sh`. `--expect 3` is derived from `CORE_TARGETS.length` in the test,
never written as a literal, so a fourth target cannot land with this co-edit missed.

The `install.sh` attached is a **copy, not a door** (D5): the canonical installer stays on
`raw.githubusercontent.com/actana/control/main/install.sh` under ADR 0016 D29. It is the
train's own copy at the commit being cut, which since #317 carries the line it was cut from in
its own bytes — stamping it is a cut's job and `Train rules` asserts it (D4), so this workflow
attaches what the ref holds and re-stamps nothing.

Nothing is published to registry.npmjs.org, under any dist-tag (D15).

## Notes for the tickets running beside this one

I own `.github/workflows/beta-release.yml` and `scripts/__tests__/workflows.test.mjs` and have
edited nothing else. Where another ticket's file needs something, it is written here rather
than edited:

**#327 — `release.yml`, `promote.yml`, `ci.yml`, `container-image.yml`: no change is required,
and one thing must NOT change.**

- `release.yml`'s `resolve` must **keep** refusing a tag that is reachable from neither `main`
  nor a `release/*` branch. That refusal is what this workflow routes around, and ADR 0036 D9
  refuses the third-mode refactor that would loosen it. `workflows.test.mjs` now pins the
  error title; if #327 rewords it, that test needs the new wording, not deletion.
- `promote.yml` needs nothing. Its tag-immutability check reads `refs/tags/vx.y.z`, and a beta
  tag is `refs/tags/vx.y.z-beta` — a different ref name, so the two never collide (D7). No
  beta tag is ever an input to a promotion (D21).
- `ci.yml` needs nothing. Merges into the train keep publishing only the moving images (C3).
- `container-image.yml` needs nothing here; the beta image retag is #319.

**#321 — `scripts/lib/core-tarball.mjs`, `build-core-tarball.mjs`, `compose-core-shasums.mjs`:
called, never edited.** `--version <x.y.z-beta>` and `--expect 3` are the whole interface used.
If a fourth `CORE_TARGETS` row ever lands, this file's `EXPECTED` asset list and its
`for target in linux-x64 linux-arm64 mac-arm64` guard are two more co-edits for that header's
list; the test derives `--expect` from `CORE_TARGETS.length` and will go red first.

**#320 — `packages/**`: untouched.** When the packed CLI tarball is attached, the step to
extend is `Create or update the prerelease and replace its assets`, and **the `EXPECTED` array
is the line to add the asset name to** — not just the upload. The prune step reads that array,
so an upload added without a matching name would have the *next* cut delete it. If the CLI
tarball is to get a `SHA256SUMS` row, that is #315's call and `--expect` moves with it; this
ticket owns the three Core rows.

**#323 — docs: three things to write, none of them edited here.**

1. `docs/ci-cd.md` § "At a glance" says *"Five entry points and one reusable workflow"* and
   has no `beta-release.yml` row. It needs one: trigger *dispatch, naming a train*; produces
   *the moving `vx.y.z-beta` tag, a prerelease Release, three tarballs, `SHA256SUMS`,
   `install.sh`*.
2. `docs/REPO_SETUP.md` §3's tag ruleset now governs beta tags too — `refs/tags/v*` matches
   `vx.y.z-beta`, its one rule is `creation`, and the App's bypass is what makes a first beta
   cut of a line possible at all. No ruleset payload changes; the sentence does.
3. **An operational caveat worth documenting.** GitHub resolves a dispatchable workflow from
   the **default branch**, so `beta-release.yml` has to exist on `main` before
   `gh workflow run beta-release.yml --ref beta/x.y.z` can be issued at all — even though the
   run then executes the train's copy. In practice that means beta cuts begin with the first
   train cut after this milestone promotes. Worth confirming against the API before it is
   written down as fact.

## Deliberate deviation from the issue text

The issue's summary line names the third target `darwin-arm64`. **The target is `mac-arm64`**,
which is what ADR 0036 D10 says in as many words (*"the target name is `mac-arm64` and not
`darwin-arm64`: `darwin` is the `platform` field inside the manifest, never the target
name"*), what `CORE_TARGETS` declares, and what ADR 0016 D29 makes part of a contract a rename
breaks. The issue's own acceptance criteria and file pointers use `mac-arm64` throughout.

## Verification

- `vitest run scripts/__tests__/workflows.test.mjs scripts/__tests__/container-matrix.test.mjs`
  — 111 passed, including #326's block from the merged base.
- Every `run:` step in the new file bash-syntax-checked, and the non-trivial ones executed
  verbatim against stubs: the train validation (five paths), the prerelease decision through
  `release-latest.mjs`, the greenness gate (green / red / running / wrong sha), the filename
  guard (clean / countered / foreign asset), the tag move against a real local origin (first
  cut / moving / already there), the Release create and update paths including the stale-asset
  prune, and the read-back (all four branches).
- `commitlint --from origin/feat/release-pipeline --to HEAD` — 0 problems.
- `eslint scripts/__tests__/workflows.test.mjs` — clean.

**Nothing was dispatched.** This workflow has not been run, and it cannot be until it reaches
a branch a dispatch can resolve it from.


qcentic-dev:
## Acceptance criteria (#318), each with where it is met

- [x] **`v0.4.1-beta` is force-updated to the train tip, and the step logs the previous sha it
      moved off** — `beta-release.yml`, `publish` → *Move the beta tag to the train tip*.
      `git tag -f -a` then `git push --force origin "refs/tags/$TAG"`, with three named log
      branches: first cut of the line, already at this sha, or *"moves from `$previous` to
      `$SHA`"*. Pinned by `workflows.test.mjs` → *moves the beta tag on purpose and says where
      it moved from (D7)*.

- [x] **Three tarballs at version `0.4.1-beta`, each built and boot-smoked on its own
      architecture (`ubuntu-24.04`, `ubuntu-24.04-arm`, `macos-15`) via
      `scripts/build-core-tarball.mjs` and `scripts/smoke-core-tarball.mjs`** — the `tarball`
      matrix (`linux-x64` on `ubuntu-24.04`, `linux-arm64` on `ubuntu-24.04-arm`) and
      `tarball-macos` (`mac-arm64` on `macos-15`). `--version "$BETA_VERSION"` on all three;
      neither script is edited (they are #321's).

- [x] **`SHA256SUMS` composed by `scripts/compose-core-shasums.mjs` over exactly those three
      Core tarballs and verified with `sha256sum -c` before it is attached** — `publish` →
      *Compose SHA256SUMS* (`--dir core-tarballs --expect 3`) then *Verify SHA256SUMS the way
      an operator would*. The `3` is derived from `CORE_TARGETS.length` in the test, never a
      literal. `install.sh` is copied in **after** composition, so it is an asset and not a
      row — if #315 sends #320 down the asset route, that ticket decides whether its tarball
      gets a row and `--expect` moves with it; this one owns the three Core rows.

- [x] **The installer e2e runs once — ubuntu, x64 — against the built `linux-x64` tarball, and
      gates the publish** — the `installer-e2e` job: `runs-on: ubuntu-24.04`, `--distro
      ubuntu`, downloading `core-tarball-linux-x64`, and named in `publish`'s `needs:`.
      Deliberately a single job with no `strategy:` — pinned by *runs the installer end-to-end
      once, on ubuntu at x64 (D14)*, which also asserts no `macos` and no `debian` in it.

- [x] **The GitHub Release is created (or updated) with `prerelease: true` and
      `--latest=false`, assets clobbered, and the flags are read back from the API and
      asserted after the fact** — `publish` → *Create or update the prerelease and replace its
      assets* passes both flags explicitly on **both** the `create` and the `edit` path
      (never defaulted: the API's `make_latest` default is `true`, which is the bug), and
      *The Release is a prerelease, and /releases/latest does not answer it* asks the API
      afterwards for `prerelease`, `draft`, and whether `/releases/latest` answers this tag —
      that endpoint being the one `install.sh` and the update checker actually read (ADR 0023
      D9, ADR 0036 D11).

- [x] **`install.sh` is attached as an asset** — `cp install.sh core-tarballs/install.sh` in
      `publish`, from the checkout at the commit being cut. Whatever the repository copy holds
      at that ref is what is attached; since #317 that copy carries its line stamp, and
      stamping is a cut's job, not this workflow's (ADR 0036 D1, D4, D5).

- [x] **A second cut of the same beta is a supported operation, not a repair: it moves the
      tag, replaces every asset, and leaves no asset from the previous cut behind** — the tag
      move is force-update by design; `gh release upload --clobber` replaces; and the step
      **deletes every asset on the Release that is not in this cut's `EXPECTED` set** before
      uploading, so a moving tag can never carry a mix of two cuts. The prune runs *before*
      the upload precisely so a later step adding an asset (#320) is unaffected.

- [x] **Nothing in the run can publish when a gate is red — the tag move included. A red leg
      leaves `v0.4.1-beta` where it was** — `publish` is `needs: [resolve, tarball,
      tarball-macos, installer-e2e]` and is the only job that writes. Pinned by *holds every
      write behind every gate, the tag move included*, which asserts no `git push` and no
      `gh release` appears in any of the four upstream jobs. Inside `publish` the assets are
      composed, checksummed and name-checked **before** the tag moves.

- [x] **`scripts/__tests__/workflows.test.mjs` knows about the new file, and its inventory
      assertion is updated rather than deleted** — extended to seven entries (six entry points
      plus the reusable workflow) **in the same commit as the workflow file**, which is what
      ADR 0036's Consequences section asks for; the file header's D34 narrative is updated
      with it. A further 14 assertions pin the properties that are invisible in a green run.

- [x] **`scripts/lib/container-matrix.mjs` and `scripts/__tests__/container-matrix.test.mjs`
      still agree with every matrix in the repository, this file's included** — both are
      unedited and green. The beta cut's installer leg is deliberately **not** a declared
      matrix: the module's axes are crossed by `ci.yml` (pr) and `release.yml` (tag), and a
      third trigger there would demand `[ubuntu, debian] ×` its arches by that test's own
      rules — which is the matrix ADR 0036 D14 says a beta cut does not run. One leg, named
      literally, with the module's own values (`x64` → `ubuntu-24.04` → `linux-x64`) and a
      comment saying why.

### Gates, as the issue's table asks for them

| Leg | Where it comes from | In this PR |
| --- | --- | --- |
| typecheck, unit, lint, audit, secret scan | already green on the train tip | `resolve` refuses to cut unless the tip's **own** `ci.yml` run is green, matched by head sha (ADR 0023 D21) |
| `linux-x64` / `linux-arm64` tarball + boot smoke | rebuilt here | `tarball` matrix |
| `mac-arm64` tarball + boot smoke | new macOS minutes | `tarball-macos`, with D13's recurring-spend argument in the file |
| installer e2e | **ubuntu x64 only** | `installer-e2e`, one leg |
| macOS install e2e | none — there has never been one (ADR 0016 D35) | **absent, and the comments say why: the runner bills at 10×** |

### Out of scope, and untouched

The image retag (#319) and the CLI-only install path (#320) both land in this same file next.
No change to `release.yml`, `promote.yml`, `ci.yml` or `container-image.yml`; no change to
`install.sh`'s own behaviour (#316, #317); no change to the docs (#323). What those tickets
need from me is written in the pull request description rather than edited in.

**Nothing was dispatched, nothing was merged, and this stays a draft.**

mrfqcentic (changes requested):
## Review of #339 against #318 and ADR 0036 — **request changes** (three fixes, all small)

This is a careful implementation of ADR 0036 and not a variant of it. Every clause the record
assigns to #318 is in the file at a line I can point at, the boundary is exact, and the tests
enforce rather than describe — I mutation-tested five of them and four went red. **11 of 13
acceptance criteria are fully met.** Three things want a change before this merges.

---

### 1. `beta-release.yml:330-332` — the greenness gate is missing `--event push`

```
runs="$(gh run list --repo "$REPO" --branch "$TRAIN" --workflow ci.yml \
          --limit 30 --json headSha,conclusion,status,url,event)"
run="$(jq -c --arg sha "$SHA" '[.[] | select(.headSha == $sha)] | .[0] // empty' <<<"$runs")"
```

`event` is fetched in the `--json` list and never referenced — not by the jq filter, not
anywhere else; `--event` appears nowhere in the file. Selection is by `headSha` alone, and
`.[0]` takes the newest match.

`gh run list --branch` returns `pull_request` runs as well as `push` runs. Verified against
this repository:

```
$ gh run list --branch feat/beta-release-workflow --workflow ci.yml --limit 10 \
    --json headSha,event,conclusion
pull_request   success   53111ee8
```

So once the promotion pull request for a train is open — `beta/0.4.1` → `main`, head sha =
train tip, created *after* the train's last push — `.[0]` selects that `pull_request` run
instead of the push run. The two are not the same gate: on a `pull_request` event
`Resolve train tags`, `Core image (train)` and `Panel image (train)` are **skipped** (visible
in this PR's own check list), and under ADR 0023 D33 a draft pull request resolves both image
checks to `pass` without building. A cut can therefore be gated by a run that never published
the train's `beta-x.y.z` digest — which is precisely the digest #319 will retag from in this
same file. The tarball-relevant coverage is equivalent on both events, so this is not a hole
today; it becomes one the moment #319 lands.

#318 words the criterion as *"the train tip's `ci.yml` **push** run"* and points at
`ci.yml`'s `assert_train_is_green`, which does filter (`ci.yml:309-311`):

```
gh run list --repo "$REPO" --branch "$train" --workflow ci.yml \
  --event push --limit 1 --json conclusion,status,url \
```

**Fix:** add `--event push` to the `gh run list` call (keep `--limit 30` and the headSha
`select`), or add `select(.event == "push")` to the jq — either way the `event` field in
`--json` becomes load-bearing rather than vestigial.

---

### 2. `beta-release.yml:855-858` — the Release notes go stale on every cut after the first

The notes body is written **only** in the create branch (`:860-871`), and its first line is:

```
"Beta of the \`$VERSION\` line, cut from \`$TRAIN\` at \`$SHA\`." \
```

The update branch runs `gh release upload --clobber` and then:

```
gh release edit "$TAG" \
  --title "$BETA_VERSION" \
  --prerelease=true \
  --latest=false
```

No `--notes-file`, and `$RUNNER_TEMP/notes.md` is never even written on that path. So on the
second and every later cut of a line, the Release body permanently states the **first** cut's
sha while the tag, the assets and the bytes are the current cut's.

A second cut is an explicitly supported operation (#318's ninth criterion), and the comment
directly above this code says *"the Release never carries a mix of two cuts"* (`:835-838`).
The notes are the one surface where it still does — and it is the surface that names the
commit sha, which ADR 0036 D7 calls the immutable record a beta actually has. An operator
reading the `v0.4.1-beta` Release page after a re-cut is told the wrong commit.

**Fix:** hoist the `printf … > "$RUNNER_TEMP/notes.md"` above the `if`, and add
`--notes-file "$RUNNER_TEMP/notes.md"` to the `gh release edit` call.

Related, same call: the comment at `:852-854` says a hand-edited Release *"carries whatever a
person picked for it — so both are re-asserted on every cut."* `draft` is the third thing a
person can pick by hand, and it is not re-asserted — the read-back at `:925-928` detects it but
only after the tag has moved and the assets are up. `--draft=false` on the same `edit` would
make the repair symmetric with the detection.

---

### 3. `scripts/__tests__/workflows.test.mjs:1024-1025` — the D11 flag assertions are satisfiable by prose

```js
expect(job.match(/--prerelease=true/g)?.length ?? 0).toBeGreaterThanOrEqual(2);
expect(job.match(/--latest=false/g)?.length ?? 0).toBeGreaterThanOrEqual(2);
```

`job` is `code(jobBlock(source, "publish"))`, which includes the two `::error` remediation
hints at `beta-release.yml:922` and `:940`. Both embed the literal
`gh release edit $TAG --prerelease=true --latest=false`. That is four matches for each regex,
so `>= 2` is met by the error messages alone.

I proved it: deleting `--prerelease=true` **and** `--latest=false` from *both* the
`gh release create` and the `gh release edit` invocations leaves the suite fully green —
`Tests 82 passed (82)`. The comment above says *"Never defaulted, on either path"*; nothing
pins either path.

This matters more than a usual test-strength nit because it is the *pre-merge* guard for the
one flag ADR 0023 D9 says must never be lost. The runtime read-back at `:906-947` is a real
assertion and the right shape — but it runs **after** the tag has moved (`:732`) and after the
Release exists, so a lost flag is caught with the beta already sitting on `/releases/latest`.
That is why its own error text says *"Fix it now"*.

**Fix:** assert on the invocations rather than counting over the whole job — extract the
`gh release create … \` and `gh release edit … \` continuation blocks and assert each contains
both flags.

---

### What I verified, and what held

**ADR 0036, clause by clause.** `workflow_dispatch` and nothing else, one `train` input
(`:148-156`; the `on:` block parses to exactly `workflow_dispatch`). Train validated
`^beta/[0-9]+\.[0-9]+\.[0-9]+$` (`:228`). Moving tag `v$beta_version` (`:246`), force-updated
with the previous sha named (`:741-755`). Prerelease Release, `--prerelease=true --latest=false`
on both paths (`:857-858`, `:878-879`). Three tarballs — `linux-x64` / `linux-arm64` on
`ubuntu-24.04` / `ubuntu-24.04-arm` (`:391-394`) and **`mac-arm64`** on `macos-15` (`:480-485`),
which is `CORE_TARGETS`' darwin row (`core-tarball.mjs:58`) under the name ADR 0036 D10 requires.
`SHA256SUMS` over exactly those three (`:675`, `--expect 3`), `sha256sum -c` before attach
(`:679`). `install.sh` in `EXPECTED` (`:816`, `:821`), attached as a copy of the train's own
stamped file (`install.sh:80` carries `LINE="0.4.1"`); the canonical door on `main` is untouched.

**No counter reaches any surface.** `grep -E 'beta\.[0-9]'` over the whole file: **zero hits**.
`github.run_number`, `github.run_attempt`, `github.run_id`, short sha: zero hits outside the two
comment blocks and one error string that *prohibit* them (`:40-41`, `:237`, `:243`). Every
surface traces to one concatenation — `beta_version="$version-beta"` (`:241`) from
`version="${train#beta/}"` (`:232`), asserted `^[0-9]+\.[0-9]+\.[0-9]+-beta$` (`:242`). I ran
that derivation verbatim: `beta/0.4.1` → `0.4.1-beta` / `v0.4.1-beta`; `beta/0.4.1.1`,
`beta/0.4.1-rc1`, `main`, `beta/x.y.z` all refused. I ran the publish-side guard (`:688-715`)
too: it rejects `actana-core-0.4.1-beta.1-*.tar.gz` and rejects a `0.4.0-beta` leftover riding
along. Filenames come from `tarballName(version, target)` (`core-tarball.mjs:228-230`), which
appends nothing; `SHA256SUMS` and `install.sh` are literals. No image tag exists here yet (#319).

**Manual only.** Parsed triggers across all seven workflows: `beta-release.yml` →
`['workflow_dispatch']`. No workflow in the repository has a tag trigger, so the App-pushed
`refs/tags/v0.4.1-beta` fires nothing — the claim at `:761` is true.

**#326.** Same hazard shape, guarded harder: `resolve` refuses unless
`github.ref_name == inputs.train` (`:257-260`), there is no `uses: ./.github/workflows/` to
resolve from a caller's sha, and the printed re-dispatch carries `--ref`. The test sweeps every
`gh workflow run` line for one.

**Nothing writes ahead of a gate.** `publish` needs all four upstream jobs (`:607`); the tag
move is inside it (`:732`), after the assets are composed and verified. The test asserts no
`git push` and no `gh release` in `resolve`, `tarball`, `tarball-macos` or `installer-e2e`.

**The lighter gate is stated.** Installer e2e once, ubuntu x64, no `strategy:` (`:553-596`); no
macOS e2e, and the header (`:106-111`), `tarball-macos` (`:466-472`) and `installer-e2e`
(`:552`) each say the 10× billing is the reason. The test pins `/10×/` in the comments and
`runs-on: macos-` to exactly one occurrence.

**The tests enforce.** Mutation-tested in a throwaway worktree — appending a counter to
`beta_version`, introducing `github.run_number`, adding `push: branches: ["beta/**"]`, and
adding a macOS installer e2e leg each turned the suite red on the specific assertion. Only the
flag mutation (finding 3) slipped through. Baseline: 111 passed across `workflows.test.mjs` +
`container-matrix.test.mjs`. All 14 multi-line `run:` blocks are `bash -n` clean, and
`node --input-type=module -e` resolves `./scripts/lib/release-latest.mjs` from cwd and answers
`prerelease=true` for `0.4.1-beta`.

**Boundary.** Two files. `promote.yml`, `release.yml`, `ci.yml`, `container-image.yml`,
`core-tarball.mjs`, `build-core-tarball.mjs`, `compose-core-shasums.mjs`,
`smoke-core-tarball.mjs`, `release-latest.mjs`, `install.sh`, `container-matrix.mjs`,
`container-matrix.test.mjs` and all of `packages/**` are byte-identical to
`feat/release-pipeline`.

**#320 will land cleanly**, with one caveat worth a comment. `EXPECTED` (`:817-821`) is the
single list the upload and the prune both read, and the file says so (`:776-779`). The caveat is
ordering and extension: `compose-core-shasums.mjs --expect 3` (`:675`) and the filename guard
(`:706-714`) both sweep `*.tar.gz` in `core-tarballs/`. An `npm pack` output is `.tgz` and slips
past both harmlessly, but a CLI asset named `*.tar.gz` dropped into that directory before `:675`
would break `--expect 3` and trip the foreign-asset guard. One sentence beside `EXPECTED` —
"add it after the checksum and filename steps, or give it a non-`.tar.gz` name" — would save
#320 a debugging round.

---

### Acceptance criteria

| # | Criterion | Verdict | Enforcing line |
|---|---|---|---|
| 1 | file exists, dispatch only, one train input, `beta/x.y.z` | **met** | `:152-156`, `:228` |
| 2 | concurrency from the version, `cancel-in-progress: false` | **met** | `:167-168` |
| 3 | refuses a tip whose `ci.yml` **push** run is not green, naming the URL | **partially met** | `:322-349` — matched by headSha, not by `event == push` (finding 1) |
| 4 | tag force-updated, previous sha logged | **met** | `:741-755` |
| 5 | three tarballs, own architecture each | **met** | `:391-394`, `:482-485`, `:433`, `:445`, `:511`, `:523` |
| 6 | `SHA256SUMS` over exactly three, `sha256sum -c` before attach | **met** | `:675`, `:679` |
| 7 | installer e2e once, ubuntu x64, gates publish | **met** | `:553-596`, `:607` |
| 8 | prerelease + `--latest=false`, clobbered, **flags read back** | **met** (workflow); test pin is finding 3 | `:855-880`, `:906-947` |
| 9 | `install.sh` attached | **met** | `:816`, `:821` |
| 10 | second cut replaces everything, leaves nothing behind | **partially met** | `:839-851` for assets; the notes still carry the first cut (finding 2) |
| 11 | nothing publishes when a gate is red, tag move included | **met** | `:607`, `:732` |
| 12 | `workflows.test.mjs` updated, inventory extended not deleted | **met** | test `:84-93` |
| 13 | `container-matrix` still agrees, this file included | **met** | 111 tests green; this file declares no container matrix, by design (`:536-547`) |

On `mac-arm64` vs the issue summary's `darwin-arm64`: the PR is right and the issue's summary
line is loose. ADR 0036 D10 says the target name is `mac-arm64` in as many words, `CORE_TARGETS`
declares it, and #318's own criteria and file pointers use `mac-arm64` throughout.

---

**Mergeable:** yes — `MERGEABLE` / `CLEAN`, no conflicts; `feat/release-pipeline`'s tip
`e233d50` is the merge base, so the base is fully merged in. Still a **draft**. CI on the head
`53111ee` is green: 19 passed, 4 skipped (the three train-image jobs and one manifest publish,
correctly skipped on a `pull_request` event), 0 failed.

Fix the three above and this is a clear approve.

qcentic-dev:
All three findings fixed, and the one non-blocking observation taken. Four commits, both files,
nothing else touched. Thank you for the mutation testing — finding 3 in particular is the kind
of thing a green suite hides forever.

### 1. `--event push` on the greenness gate — fixed (`e25714e`)

You were right that the `event` field was vestigial. It is now load-bearing three times over:
filtered at the API (`--event push`, the way `ci.yml:309-311` already does it), asserted again
in the `jq` (`select(.headSha == $sha and .event == "push")`), and named in the success line
and in every error. Two filters for one fact is deliberate rather than belt-and-braces theatre:
the flag is what makes the query right, and the `select` is what survives a later edit dropping
the flag and leaving a query that still returns something.

I reproduced your scenario against a stub — a **newer, green `pull_request` run** and an older
**red `push` run**, both at the train tip's sha:

```
[gh-args] run list --repo a/b --branch beta/0.4.1 --workflow ci.yml \
            --event push --limit 30 --json headSha,conclusion,status,url,event
::error title=The train tip is red::ci.yml's push run on aaa concluded 'failure' … rc=1
```

Before the fix that case cut the beta. I also ran it with the API filter *ignored* (stub
returning the mixed list anyway) to confirm the `jq` alone still refuses.

The "no run" error now names the trap explicitly, because the operator hitting it will be
looking at a green promotion PR: *"A green run on the promotion pull request is not this gate
and does not substitute for it: on a pull_request event the train image jobs are skipped, and
a draft resolves both image checks to `pass` without building (ADR 0023 D33)."* The step is
renamed *The train tip's own CI push run is green*, and the comment block carries your #319
argument — the digest a `pull_request` run never publishes is exactly the one #319 retags from.

### 2. Stale Release notes, and `draft` — fixed (`e88ad8f`)

The `printf … > "$RUNNER_TEMP/notes.md"` is hoisted above the `if`, so both paths write **this**
cut's body, and `gh release edit` takes `--notes-file`. You are right that this was the one
surface still carrying a mix of two cuts, and the worst one to get wrong — it is the sha ADR
0036 D7 calls a beta's immutable record.

`--draft=false` is on the same call, for the reason you gave: the comment claims every field a
cut decides is re-asserted on every cut, and `draft` was the third field. Detecting it in the
read-back after the tag has already moved is the wrong end of the run for something an `edit`
can simply set; the repair now sits beside the detection.

Second-cut path, run verbatim against a stub:

```
[gh] release delete-asset v0.4.1-beta actana-core-0.4.1-beta-mac-x64.tar.gz --yes
[gh] release upload v0.4.1-beta …5 assets… --clobber
[gh] release edit v0.4.1-beta --title 0.4.1-beta --draft=false --prerelease=true \
       --latest=false --notes-file …/notes.md
$ head -1 notes.md
Beta of the `0.4.1` line, cut from `beta/0.4.1` at `SECOND_CUT_SHA`.
```

### 3. The D11 assertions were satisfiable by prose — fixed (`e8526c7`)

Confirmed and replaced rather than tightened. A small `invocation(text, command)` helper walks
a `gh release …` call through its backslash-continued lines, and each of the two invocations is
now required to carry `--prerelease=true`, `--latest=false` and `--notes-file`, with
`--draft=false` on the edit. A separate assertion requires the notes write to sit **above** the
branch, which is the shape that prevents finding 2 from coming back. The read-back keeps its
own test, now including the `draft` question it was already asking.

**Six mutations, all red** on the committed tree (baseline: 112 passed across
`workflows.test.mjs` + `container-matrix.test.mjs`):

| Mutation | Result |
|---|---|
| delete both flags from **both** `gh release` invocations — your exact repro | `1 failed \| 82 passed` |
| delete `--notes-file` from `gh release edit` | `1 failed \| 82 passed` |
| delete `--draft=false` | `1 failed \| 82 passed` |
| delete `--event push` | `1 failed \| 82 passed` |
| delete the `jq` event `select` | `1 failed \| 82 passed` |
| move the notes write back inside the create arm | `1 failed \| 82 passed` |

### Non-blocking observation: taken (`cd4e6fd`)

The `.tar.gz` ordering caveat for #320 is now a comment beside `EXPECTED`, in your words: both
`--expect 3` and the filename guard sweep `*.tar.gz` in `core-tarballs/`, `pnpm pack` yields a
`.tgz` that slips past harmlessly, and a CLI asset staged there under a `*.tar.gz` name before
those steps would break both. *"Stage it after them, or give it a name that is not
`*.tar.gz`."*

**Deferred: nothing.** Every blocking finding and the single non-blocking observation are in.

### Boundaries and state

Two files, as before: `.github/workflows/beta-release.yml` and
`scripts/__tests__/workflows.test.mjs`. `promote.yml`, `release.yml`, `ci.yml`,
`container-image.yml`, the three tarball scripts, `release-latest.mjs`, `install.sh`,
`container-matrix.*` and all of `packages/**` are untouched. The base had not moved
(`e233d50` is still the tip and was already merged in), so there was nothing to merge and
nothing was force-pushed. Still a **draft**, base still `feat/release-pipeline`, and the
workflow has not been dispatched.

qcentic-adm (approved):
## Re-review of #339 at `cd4e6fd` against #318 and ADR 0036 — **approve**

All three blocking findings from the previous review are fixed at the enforcing line, and I verified each from the head rather than from the reply. Nothing new is blocking.

### 1. Greenness gate — `beta-release.yml:349-352` — fixed

`gh run list … --event push` is now on the query, and the `jq` is `select(.headSha == $sha and .event == "push")`; the `event` field is load-bearing, and the selected run's event is checked a third time at `:360-363`. I re-ran the step verbatim against a stub returning a **newer green `pull_request` run** and an **older red `push` run** at the same sha: (a) with the API honouring `--event push` → *"The train tip is red"*, rc=1; (b) with the API filter ignored and the mixed list returned → the `jq` still picks the push run → red, rc=1; (c) filter ignored, only a green `pull_request` run at the sha → *"No CI push run for the train tip"*, rc=1; (d) green push run → ✅ rc=0. **A `pull_request` run can no longer be selected**: it is dropped by the API filter, dropped again by the `select`, and would be refused by the `event != push` check if it somehow reached it. Live check: `--event push` on this PR's branch returns nothing where the unfiltered list returns the two `pull_request` runs, so the flag is honoured by the real API.

### 2. Stale notes on re-cut, and `draft` — `beta-release.yml:876-885`, `:916-921` — fixed

The `printf … > "$RUNNER_TEMP/notes.md"` sits above the `if gh release view` branch, and `gh release edit` now carries `--draft=false --prerelease=true --latest=false --notes-file`. I ran the publish step's script against a stub for a first cut and a second cut: the second cut emitted `delete-asset` for a stale foreign asset, `upload … --clobber`, then `release edit v0.4.1-beta --title 0.4.1-beta --draft=false --prerelease=true --latest=false --notes-file …`, and `notes.md` line 1 read *"cut from `beta/0.4.1` at `SECOND_CUT_SHA`"*. `gh release edit --draft=false` and `--notes-file` are valid on gh 2.97. `draft` is now re-asserted on the same call as the two D9 flags, symmetric with the read-back at `:966-976`.

### 3. D11 assertions — `workflows.test.mjs:833-840`, `:1055-1083` — fixed

The `invocation(text, command)` helper takes the first non-comment line containing the command and walks its `\`-continued lines, so the assertion sees only the arguments a runner would execute — the `::error` remediation hints at `:973`/`:991` are outside both blocks and cannot satisfy it. Deleting `--prerelease=true` and `--latest=false` from **both** `gh release create` and `gh release edit` now fails the suite at *"gh release create does not pass --prerelease=true"* — the exact mutation that was green before. I ran all six claimed mutations on the head tree: both flags from both invocations, `--notes-file` off `edit`, `--draft=false` off `edit`, `--event push` off the query, the `jq` event select, and the notes write moved into the create arm — each `1 failed | 111 passed`; baseline `112 passed` across `workflows.test.mjs` + `container-matrix.test.mjs`.

### Also checked

- `grep -E 'beta\.[0-9]'` over `beta-release.yml`: **zero hits**. `run_number` / `run_attempt` / `run_id` / short sha: only the prohibitions at `:40-41`, `:237`, `:243`.
- Boundary: `git diff --name-only e233d50 cd4e6fd` is exactly `.github/workflows/beta-release.yml` and `scripts/__tests__/workflows.test.mjs`. `promote.yml`, `release.yml`, `ci.yml`, `container-image.yml` and the tarball scripts are untouched.
- `MERGEABLE` / `CLEAN`; base `feat/release-pipeline` at `e233d50` is the merge base. Still a draft.
- CI on `cd4e6fd`: 19 passed, 4 skipped (`Resolve train tags`, `Core image (train)`, `Panel image (train)`, both `Publish the multi-arch manifest`), 0 failed.
- The #320 `.tar.gz` ordering caveat is now beside `EXPECTED` (`:806-812`).

Two non-blocking observations, no change requested: the `event != "push"` branch at `:361` is unreachable after the `jq` select (harmless belt-and-braces); and `invocation()` relies on the real `gh release …` call preceding any prose that quotes it in the job — true today, and the `--notes-file`/`--draft=false` requirements would still fail on a prose line, so it is not a hole.

**Verdict: approve.** 13 of 13 acceptance criteria met.

mrfqcentic (approved):
All three findings from my review at \`53111ee\` are confirmed fixed at \`cd4e6fd\`. Re-checked at the enforcing line only, not a fresh review.

1. **Greenness gate event filter** — `beta-release.yml:349-352`. `--event push` is now on the `gh run list` call and `select(.event == "push")` is in the jq, so a promotion pull request run with the same head sha can no longer be selected. The `event` field in `--json` is load-bearing now, and `:361-363` asserts it a third time after selection with its own error. Stronger than what I asked for.

2. **Release notes going stale on a re-cut** — `beta-release.yml:874-885`, `:916-921`. The `printf … > "$RUNNER_TEMP/notes.md"` is hoisted above the `if`, and `--notes-file` is on both the `edit` and the `create` invocation, so a second cut rewrites the body and the page names this cut's sha. **`draft` is re-asserted too** — `--draft=false` at `:918`, beside `--prerelease=true --latest=false`, which makes the repair symmetric with the read-back detection at `:965+`.

3. **Flag assertions satisfiable by the remediation prose** — `workflows.test.mjs:833-840`, `:1056-1083`. The count over the job text is gone; an `invocation()` helper extracts each `gh release` call from its naming line through its last `\`-continued line, and both flags are asserted against `create` and `edit` individually. I re-ran my original mutation against the new assertions: deleting `--prerelease=true`, `--latest=false`, `--notes-file` and `--draft=false` from both real invocations — leaving both `::error` hints that embed the same literals untouched — now fails every one of the four `toContain` checks on both verbs. That is the mutation that left the suite green before. The test also pins the notes write ahead of the branch, so the create-arm variant of finding 2 cannot come back.

CI on `cd4e6fd` is green: 18 passed, 5 skipped (the train-image jobs and one manifest publish, correctly skipped on a `pull_request` event), 0 failed.

Clearing my stale `CHANGES_REQUESTED` against `53111ee`. Approving. Still a draft — that is the only thing left between this and merge, and it is the author's call.

PR #341, opened 2026-08-25
feat(ci): the beta image tags, retagged from the train's digest — and #337's npm asset step

qcentic-dev:
## This pull request carries two things, and both edit `beta-release.yml`

**1. #319 — the beta image tags.** `x.y.z-beta` is retagged from the digest the
train already published as `beta-x.y.z`, never rebuilt, with `docker buildx
imagetools create` and ADR 0023 D16's revision assertion in front of it.

**2. #320's workflow step, which #320 deliberately did not land.** [#337
(4d6107a)](https://github.com/actana/control/pull/337) finished the script half —
`scripts/rehearse-npm-publish.mjs --beta`, packed, asserted and install-checked —
and could not land the workflow half, because **#318 owned `beta-release.yml` at
the time and #337 owned no file under `.github/workflows`**. It wrote the step
out as YAML in [a pull request
comment](https://github.com/actana/control/pull/337#issuecomment-5400143000) for a
later turn instead, and left that one acceptance box unchecked.

**Why they are one pull request:** they edit the same file, in the same job in
two places. `EXPECTED` in `publish` is the single asset contract the create, the
clobbering upload and the prune all read; #320 extends it and #319 does not, but
both changes land inside the same `publish` job and the same header. Two pull
requests would conflict on every hunk, and the second would be rebasing prose
about the first. Split into **two commits**, one per ticket, so each is
reviewable on its own.

---

## #319 — retag, never rebuild

Two `uses:` calls into `container-image.yml`'s `promote` mode — the machinery
issue #319 says to call rather than re-implement:

```yaml
  core-image:
    name: Core image (beta)
    needs: [resolve, tarball, tarball-macos, installer-e2e, publish]
    uses: ./.github/workflows/container-image.yml
    with:
      image: core
      ref: ${{ needs.resolve.outputs.sha }}
      stage: ${{ needs.resolve.outputs.beta_version }}
      mode: promote
      source_tag: beta-${{ needs.resolve.outputs.version }}
      tags: ${{ needs.resolve.outputs.beta_version }}
      push: true
```

- **Nothing is built.** Every build step in `container-image.yml` is gated on
  `inputs.mode == 'build'`, Trivy included, and `Push the per-arch tags` on
  `inputs.push && inputs.mode == 'build'`. No `docker build`, no Dockerfile, no
  Trivy leg, no per-arch scaffolding, and no such string anywhere in this file
  (asserted).
- **D16 is inherited, not restated.** `container-image.yml` refuses the digest,
[...]

PR #342, opened 2026-08-25
feat: the beta release channel and the pipeline it needs

### LINKED ISSUE #316: install.sh installs and stops activating: the Core bundle and the launcher are placed, setup is the next command

qcentic-dev:
The F2 gate for milestone 0.4.1: the beta release channel and the pipeline it needs, proposed onto the `beta/0.4.1` train as one reviewable unit.

## The eleven commits this branch carries

**`5c5ddde` — docs(adr): ADR 0036, the beta release channel**
The record the rest of the milestone is built from. It revisits ADR 0023 D8 and D9 — the two clauses that each named a precondition for a beta gaining a GitHub Release — and records four operator constraints rather than re-weighing them: the beta version string is `x.y.z-beta` on every surface with no counter and no suffix, install is not activation, a beta cut is dispatched by a person, and the URL scheme is the ref. It decides that the channel is a stamped *line* version, because a `curl | bash` script cannot read its own URL and promotion fast-forwards `main`. ADR 0016 and 0023 gain dated notes; both stay ACCEPTED and nothing is renumbered.

**`77b9be2` — chore: delete the automatic train cut**
A train's version is an administrative decision — the next release may be a patch, a minor or a major, and only a person knows which. The automatic cut that guessed is deleted, and the runbook becomes one procedure rather than a path plus a recovery from it: the six-manifest rewrite as a scripted `files=()` array, the ruleset bypass and `--no-verify` the push actually needs, and the commit message the cut has always used. ADR 0023 D25's invariant — a train is always open — moves from the deleted job onto a daily housekeeping check rather than onto somebody remembering.

**`dceeb12` — refactor: install.sh installs, and setup activates**
Closes #316. `install.sh` places the Core bundle and the launcher and stops; `actana setup` is the separate command the operator runs next, and the installer prints the exact line to run. Placement moves ahead of activation without leaving the CLI — `planCorePlacement` and `placeCoreBundle` behind a new `actana place` verb, which `runActanaSetup` also calls — so the layout is resolved once in `actana-layout.ts` and never in POSIX sh. The Panel's "needs update" remedy becomes `actana update`, and the Linux installer e2e enters at the one-liner, asserts an installed-not-activated machine, then runs the `actana setup` it printed, as printed.

**`1292c20` — feat: install.sh derives its channel from the line it carries**
A script fetched by `curl | bash` cannot know the URL it came from, and promotion is a fast-forward — so a `CHANNEL=beta` constant committed on a train becomes `main`'s own bytes. What survives a fast-forward is a value true on both sides of it: a `LINE="x.y.z"` stamp written by the cut alongside the six manifests (ADR 0036 D1). `resolve_version` becomes D2's ladder: an explicit `--version` / `ACTANA_VERSION` wins, else the release `v<line>`, else `v<line>-beta`, else `/releases/latest` as the terminal fallback. This changes the stable one-liner's default, and the record says so.

**`e233d50` — fix: a promotion runs the train's own release workflow**
A `workflow_dispatch` run resolves every workflow file it executes from the ref it was dispatched on, so `uses: ./.github/workflows/release.yml` in `promote.yml` ran **main's** release pipeline against the train's commit — run 32716466300 is the instance, and nothing was red until after the irreversible half had run. The fix is structural: `release.yml` loses its `workflow_call` trigger so no `uses:` can resolve it from a caller's SHA at all, and `promote.yml` dispatches it at the `vx.y.z` tag `advance` just pushed, waits, and carries the conclusion. That job asserts the release run's `head_sha` equals the promoted commit, and a new `preflight` job refuses the cheap half while nothing has moved.

**`4d6107a` — feat: the CLI installs from a Release asset on the beta path**
ADR 0036 D15 drops the registry from the beta path — npm burns a version on first publish, so a second cut of the same `x.y.z-beta` would 403 after the tag had moved — and D16 attaches the packed CLI to the prerelease as an asset instead. `assertPackedManifest` gains `beta: true`, which narrows exactly two rules and relaxes none: the version must be C1's `x.y.z-beta` anchored at both ends, and the `@actana/sdk` dependency must be *absent* rather than pinned. `assertBundleInlines` makes that absence honest by reading the packed bundle and failing if `@actana/sdk` appears in specifier position. The release path is byte-for-byte what it was.

**`a28559a` — fix: a Core tarball carries its version in one place**
ADR 0036 D18 keeps a Core tarball self-identifying: the version stays in the asset name, in the archive root and in `core-manifest.json`, which is what makes a beta's bytes impossible to rename into a release's. `assertCoreVersion` is that contract in one place, and `tarballName`, `tarballRootDirName` and `buildManifest` all call it — three doors, one gate. A counted beta is refused outright; the backport candidate 0023 D30 publishes keeps its identifier. The builder gates `--version` before anything is staged, so a bad string fails with a `[core-tarball]` line rather than deep inside a tar.

**`66ef4df` — feat: a beta cut publishes a prerelease with its assets**
`beta-release.yml`, the new entry point ADR 0036 D9 asks for. It publishes the moving `vx.y.z-beta` git tag, a prerelease GitHub Release, three Core tarballs (`linux-x64`, `linux-arm64`, `mac-arm64`), a `SHA256SUMS` over exactly those three, and `install.sh` attached beside them (D10). It is a separate workflow rather than a third mode of `release.yml`, and that record refuses the merge explicitly — `release.yml` is untouched and a beta tag dispatched at it must keep failing exactly as it does now. The version string is asserted before anything is built and again on the asset filenames; the trigger is `workflow_dispatch` and nothing else.

**`141e5ce` — refactor: one version asserted everywhere it is written**
A version here was six independent strings kept equal by a check that ran on some events and not others, and five of the six writers derived their string from a *name* — so every comparison between them was a tautology that passed while the manifests inside the bytes said something else. `scripts/lib/version-agreement.mjs` is the one definition all of them derive from, and `scripts/assert-version-agreement.mjs` is what the workflows call: `--expected` takes whatever spelling the caller's surface uses, `--git-ref` reads manifests out of a commit without moving a working tree, and `--image-label` asks the same question of bytes rather than of files. It imports nothing outside `node:` builtins. ADR 0037 is added; a counted beta is refused by name, and a counted rc stays the specified backport form.

**`cba296e` — feat: beta images are retagged from the verified digest**
`beta-release.yml` gains `core-image` and `panel-image`, two `uses:` calls into `container-image.yml`'s `promote` mode. Nothing is built: the bytes already exist — `ci.yml`'s train jobs published them as `beta-x.y.z` on the merge that made this tip — and the retag gives that digest a second name with `docker buildx imagetools create` (ADR 0023 D17, ADR 0036 D12). The machinery is called rather than re-implemented, so the per-architecture `org.opencontainers.image.revision` assertion is not re-stated anywhere it could be weakened. Exactly one tag per repository, `latest` never in it, and both jobs run after `publish` because an image tag cannot be taken back.

**`fef0c6d` — docs: the beta channel in the operator's voice**
The tag ladder gains the two classes a beta cut publishes and loses the claim that betas are Docker-only: `beta-x.y.z` and `x.y.z-beta` are two clocks, not two names for one thing. `beta-release.yml` joins the at-a-glance table as the sixth entry point, and "Cutting a beta" is the runbook for it — how a cut is requested, what it publishes, and the three things it deliberately does not (no Intel Mac build, no signing, nothing on registry.npmjs.org). Three older claims are corrected where the milestone reaches them: the digest guarantee does not extend to the Core tarballs, the beta path has no provenance attestation and `SHA256SUMS` stands in its place, and rollback does not reach the metal install path at all.

## What this pull request deliberately does NOT close

Two acceptance criteria in this milestone are **not** met by merging this branch, and they are named here rather than left for a reader to assume met.

**#323 — running the printed one-liner for real.** The criterion asks that the one-liner a beta cut prints be run for real, end to end, against the assets that cut published. That can only close once a beta cut has actually been dispatched from the train and has published a prerelease with its assets. Nothing in this branch can satisfy it, because no cut is dispatched by this pull request.

**#319 — logging which digest the beta tag was moved off.** #319's last criterion is *"a second cut re-points the tag, and says in the log which digest it was moved off."* The first half is met — the tag moves, and `container-image.yml` prints the source digest and inspects the result. The second half is **deferred, not met**: nothing on this path reads `$IMAGE:x.y.z-beta`'s *existing* digest before `docker buildx imagetools create` overwrites it, so on a second cut the log does not say what was displaced. Meeting it properly needs a read in `container-image.yml`'s promote path, before the create, and #319 puts that file out of scope in as many words. This deferral to **#315** is recorded in three places already: the PR body of **#341**, the `beta-release.yml` header ("What #319 asks for and this file does not do (deferred to #315)"), and **#315** itself.

## A beta cut cannot be dispatched until `beta-release.yml` reaches `main`

GitHub resolves dispatchable workflows from the **default branch**: a `workflow_dispatch` entry point does not exist to the API until the file is on `main`, whatever ref you intend to run it at. `beta-release.yml` is new in this branch, so until this train is promoted and `main` fast-forwards over it, **no beta cut can be dispatched at all** — not on the train, not anywhere. That is the mechanical reason #323's criterion cannot be closed here, and it is a property of the platform rather than of this branch.


docs/REPO_SETUP.md
```diff
@@ -767,6 +769,17 @@ the promotion fails at the tag after it has already fast-forwarded `main` — th
 worst place in the sequence to stop. The `creation` rule and the ref pattern
 are untouched by the payload.
 
+**That pattern now governs beta tags too, and no payload changes.**
+`refs/tags/v*` matches `v0.4.1-beta` exactly as it matches `v0.4.1`, so the
+`vx.y.z-beta` tag a beta cut publishes is created under this ruleset ([ADR
+0036](adr/0036-the-beta-release-channel.md) D7). The App bypass is therefore
+what makes the **first** cut of a line possible at all — later cuts of the same
+line only *update* an existing ref, which `creation` does not restrict, so a
```
qcentic-adm: This is true of 20390424 (`creation`) and false of the live 20390423 "Release tags are immutable" — `active` on `refs/tags/v*`, rules `update` + `deletion` + `non_fast_forward`, **no bypass actors** (`docs/rulesets/tag-immutable.json` is its verbatim capture). A second cut of `v0.4.1-beta` is an update of an existing ref and is refused for everyone, App included. This paragraph needs to record the change 20390423 needs — exclude `refs/tags/v*-beta` from its conditions rather than adding an update-bypass, which would let the App move release tags too (0023 D44) — and `tag-immutable.json` re-captured to match.

.github/workflows/beta-release.yml
```diff
@@ -0,0 +1,1439 @@
...
+          fi
+
+          # Annotated, so the record carries who cut what and when. `-f` because
+          # moving it is the design.
+          git -c user.name="${APP_SLUG:-actana-release}[bot]" \
+              -c user.email="${APP_SLUG:-actana-release}[bot]@users.noreply.github.com" \
+              tag -f -a "$TAG" "$SHA" -m "$TAG — beta cut of $TRAIN at $SHA"
+          if ! output="$(git push --force origin "refs/tags/$TAG" 2>&1)"; then
+            echo "$output"
+            echo "::error title=The tag push was refused::\`refs/tags/v*\` is governed by the 'Release tags — only admins cut releases' ruleset, which needs the App as a **bypass actor** (ADR 0023 D39, docs/REPO_SETUP.md §3). Nothing has been published: no Release has been created or updated in this run."
```
qcentic-adm: On the second cut of a line this push is refused by ruleset 20390423 ("Release tags are immutable": `update`/`non_fast_forward` on `refs/tags/v*`, no bypass actors), not by the creation ruleset this message names. The first cut passes because creation is what 20390424's App bypass covers. Please name 20390423 here, and consider reading the ruleset in `resolve` so the refusal costs one dispatch rather than three tarball builds and a macOS runner — the same argument this file already makes for the App secrets.

docs/ci-cd.md
```diff
@@ -1057,13 +1310,149 @@ after somebody has already worked the checklist against a stale image: the
...
+no gate between them, so one can fail where the other succeeded — leaving
+`core:x.y.z-beta` created and `panel:x.y.z-beta` absent, beside a prerelease
+that is already published, with the created tag unwithdrawable (0023 D45).
+**Re-dispatching the cut is the repair**, and it works because every write on
+this path is idempotent: the tag is force-moved, the assets are clobbered, and a
+retag re-points a name that is already a name. It converges only while the train
+tip has not moved — once it has, a re-dispatch cuts the newer commit rather than
+repairing the older one.
+
+A second cut of the same beta is a supported operation and not a repair: it
```
qcentic-adm, line 1427: Against the repository as configured this sentence is false: live ruleset 20390423 restricts `update` and `non_fast_forward` on `refs/tags/v*` with no bypass actors, so the force-move at `beta-release.yml` "Move the beta tag to the train tip" is refused on any cut after the first. Nothing is published — the tag move is the first write in `publish` — but the operation is not supported until the ruleset excludes `v*-beta`.

docs/ci-cd.md
```diff
@@ -1057,13 +1310,149 @@ after somebody has already worked the checklist against a stale image: the
...
+assets, so a `mac-arm64` leg that goes red leaves `vx.y.z-beta` naming exactly
+the commit it named before the dispatch. There is no tag to reset and nothing to
+clean up.
+
+**After that job there are two more that write, and a red one there does leave
+something behind.** The two retags in step 6 are independent promote calls with
+no gate between them, so one can fail where the other succeeded — leaving
+`core:x.y.z-beta` created and `panel:x.y.z-beta` absent, beside a prerelease
+that is already published, with the created tag unwithdrawable (0023 D45).
+**Re-dispatching the cut is the repair**, and it works because every write on
```
qcentic-adm, line 1410: This holds only while the tip has not moved — an unchanged ref is not sent, so no tag rule fires. Once the tip has moved, the re-dispatch dies at the tag push under 20390423 (see the comment above), and the half-retag it is meant to repair stays half. Worth saying here in words.

docs/ci-cd.md
```diff
@@ -1129,27 +1528,196 @@ would push `vx.y.z` and produce no release, and would fast-forward `main`
...
+[32716466300](https://github.com/actana/control/actions/runs/32716466300)
+promoted `beta/0.4.0` on 2026-08-24: the pause, the digest verification, the
+fast-forward and the tag all succeeded, and then every publishing job failed,
+because the promotion was running a `release.yml` older than the train it was
+promoting ([#326](https://github.com/actana/control/issues/326)). That cause is
+fixed — the release is dispatched at the tag now — but a promotion can still
+stop after `advance` for reasons that have nothing to do with it: a CVE gate
+that fires, a registry that is down, an installer e2e that goes red.
+
+**The fix arrives one promotion late, and this is the section where that
```
qcentic-adm: Two things this paragraph needs for the promotion of `beta/0.4.1` specifically. (a) `main`'s old `promote.yml` still carries `next-train`: once `beta/0.4.0` is deleted as § Promotion requires, it cuts `beta/0.5.0` with six manifests and **no `install.sh` stamp**, so that train's first push goes red at `Train versions` (`--require-stamp`: stamp 0.4.1, train 0.5.0) and publishes no `beta-0.5.0` image — it has to be deleted and re-cut by hand. (b) "cannot be otherwise" is not so: `gh workflow run promote.yml --ref beta/0.4.1 -f train=beta/0.4.1` runs the train's own copy — the `--ref` mechanism this milestone already relies on for `beta-release.yml` and the `release.yml` re-dispatch — and `macos-release` has unrestricted deployment branches (REPO_SETUP.md:345). Pick one and write it down, and say `beta/0.4.0` must go first: both copies of `promote.yml` read one surviving train as the hotfix case and would rebase and force-push it.

docs/adr/0023-release-trains-and-digest-promotion.md
```diff
@@ -154,8 +176,14 @@ Both sites gain an explicit "is this the highest released version" test, `--late
...
+>
+> Two consequences follow and are worth naming. The promotion's `release` job is now an ordinary job that dispatches and waits, so it holds `release-line` and `retire-train` exactly as the call did — and it uses `GITHUB_TOKEN` with `actions: write`, which D39 permits because a dispatch is not a push and `workflow_dispatch` is one of the two events GitHub exempts from the default token's no-trigger rule. And because the release no longer runs inside the promotion's own run, everything `release.yml` refuses on would refuse *after* `main` had moved — so the cheap half of those refusals moved upstream into a `preflight` job that runs before `advance`. The recovery for a promotion that stops after `advance` anyway is [`docs/ci-cd.md` § "When a promotion half-runs"](../ci-cd.md#when-a-promotion-half-runs). And the clause takes effect one promotion late, by construction: a promotion runs the workflow files the default branch carried when the run was created, so the promotion that *ships* this amendment still executes the calling shape it removes. That run has no `preflight`, no `--ref` and no `head_sha` assertion; it is the last one that does not.
```
qcentic-adm, line 185: "By construction" overstates it: a `workflow_dispatch` runs the file at the ref it is dispatched on, so `gh workflow run promote.yml --ref beta/0.4.1 -f train=beta/0.4.1` runs this amendment's own `promote.yml` on the promotion that ships it. Either the note says the operator does that, or it says why not — and then `docs/ci-cd.md:1562` has to cover the old copy's unstamped auto-cut.

docs/adr/0036-the-beta-release-channel.md
```diff
@@ -0,0 +1,660 @@
...
+here rather than left to be met.** `release` is an ordinary job and can go red or be
+cancelled. When it does, `main` keeps the new line's stamp and that line still has no Release,
+so D2 does not stop at step 2 — and the public door serves that line's beta **for as long as
+the failure lasts, not for one run**, if a beta of the line was ever cut, or today's newest
+release at step 4 if none was. Nothing self-corrects: the stamp cannot be walked back, because
+0023 D5 lets `main` advance only by fast-forward, and `main` is already at the train tip.
+
+**The recovery is the one 0023 D40 already documents: re-dispatch the release.** The tag is
+pushed before the release runs (`promote.yml:385-387`), and `release.yml` carries a
+`workflow_dispatch` beside the `workflow_call` `promote.yml` uses (`release.yml:119`, `:133`)
```
qcentic-adm: True when written, false six commits later in this same branch: e233d50 (#326) deletes `release.yml`'s `workflow_call` trigger. The record's own table (line 659) knows #326 touched D3; this sentence does not. A dated note under D3 — the form this record uses for 0016/0023 — would close the seam.

docs/README.md
```diff
@@ -91,6 +91,7 @@ routing around it.
 | [0032](adr/0032-one-actana-cli.md) | There is one `actana`, and it is both the Core manager and the client — supersedes #265 §6 and 0031 D8's two-binaries premise |
 | [0033](adr/0033-turn-end-is-the-one-mandatory-harness-signal.md) | Turn-end reporting is the one mandatory harness signal; turn-start, auto-mode flags and resume ids may never gate a feature |
 | [0035](adr/0035-a-second-skill-for-the-sub-agent-role.md) | A second skill for the sub-agent role, and it is eager — **PROPOSED**, depends on 0031 |
+| [0036](adr/0036-the-beta-release-channel.md) | The beta release channel: a beta installs like a release, and the ref is the channel — **PROPOSED**, amends 0016 (D28, D29, D34, D35) and 0023 (D8, D9) |
```
qcentic-adm, line 94: ADR 0037 (one version per line, added by 141e5ce) has no row in this index. `docs/adr/README.md` says the index with one-line summaries lives here.

qcentic-adm (changes requested):
## Gate review r1 — `feat/release-pipeline` → `beta/0.4.1`

**Verdict: request changes.** The eleven tickets are coherent with each other almost everywhere, the beta cut's first run walks end to end on paper, the counter ban holds on every surface, and the two open criteria are honestly recorded. Three things stop it sitting on a train that will be cut — two are the record contradicting the repository as it is actually configured, one is CI.

### 1. A second cut of a beta is refused by the live tag ruleset, and the milestone says the opposite

Live ruleset **20390423 "Release tags are immutable"** (captured verbatim in `docs/rulesets/tag-immutable.json`) is `active` on `refs/tags/v*` with rules `update`, `deletion`, `non_fast_forward` and **no bypass actors** (`gh api repos/actana/control/rulesets/20390423`). `v0.4.1-beta` matches `refs/tags/v*`. The **first** cut of a line *creates* the ref and goes through 20390424's App bypass, so it works. The **second** cut's `git push --force origin refs/tags/v0.4.1-beta` (`beta-release.yml` "Move the beta tag to the train tip") is an *update* of an existing ref, and 20390423 refuses that for everyone, the App included.

What that makes false: ADR 0036 D7's moving handle; `docs/ci-cd.md` § Cutting a beta — *"Cut the same beta ten times and it is `0.4.1-beta` all ten"*, *"A second cut of the same beta is a supported operation"*; the new `docs/REPO_SETUP.md` §3e paragraph — *"later cuts of the same line only update an existing ref, which `creation` does not restrict"* (true of 20390424, and exactly what 20390423 restricts); and the error text at `beta-release.yml:1078`, which names the creation ruleset. *"Re-dispatching the cut is the repair"* (`ci-cd.md:1403`) holds only while the tip has not moved — an unchanged ref is not pushed, so no rule fires — and not for the half-retag case the sentence is about once the tip has moved.

The failure is clean — the tag move is the first write in `publish`, so a refused second cut publishes nothing — but the operator's central claim about the channel is wrong against the live repository, and nothing on the path preflights it. Ask: `REPO_SETUP.md` §3e records the change 20390423 needs (exclude `refs/tags/v*-beta` from its conditions; adding the App as an update-bypass would let it move *release* tags too, which 0023 D44 forbids), `docs/rulesets/tag-immutable.json` is re-captured to match, and `beta-release.yml:1078` names 20390423. A `resolve`-time read of the ruleset would be better still, if a token that can read rulesets is available to the run.

### 2. Promoting *this* train runs `main`'s old `promote.yml`, and the runbook does not say what that does now

ADR 0023:185 and `ci-cd.md:1562` record that the #326 fix *"arrives one promotion late"* and *"cannot be otherwise"*. Two things are missing from that paragraph:

- **The old `promote.yml` still auto-cuts.** Once `beta/0.4.0` is deleted — which § Promotion requires, and it is still on origin — `main`'s `next-train` job cuts `beta/0.5.0` from the new `main` with six manifests rewritten and **no `install.sh` stamp** (that job never wrote one). That train's first push then goes red at the new `Train versions` job (`assert-version-agreement.mjs --require-stamp`: stamp `0.4.1`, train `0.5.0`), `Resolve train tags` never runs, and no `beta-0.5.0` image is published. The operator has to delete it and re-cut by hand. Nothing says so.
- **"Cannot be otherwise" is not true.** `gh workflow run promote.yml --repo actana/control --ref beta/0.4.1 -f train=beta/0.4.1` runs the train's own `promote.yml` — the same `--ref` mechanism `beta-release.yml` and the `release.yml` re-dispatch already rely on — and `macos-release` leaves deployment branches unrestricted (`REPO_SETUP.md:345`), so its `pause` job would run. With it, the 0.4.1 promotion gets `preflight`, the tag dispatch and the `head_sha` assertion, and no auto-cut.

Ask: the runbook picks one and says it — either *dispatch at `--ref beta/0.4.1`*, or *expect and undo the unstamped `beta/0.5.0`* — and § Promotion says in words that `beta/0.4.0` must be deleted first, because both copies of `promote.yml` read one surviving train as the hotfix case and would rebase and force-push it.

### 3. CI on the head is red

Run 32834222771 on `fef0c6de`: **Unit Tests** (required) failed — `packages/shared/src/__tests__/core-cert-material.test.ts:224` *"gives every issuance its own serial"*, `expected 137 to be less than 128`. Not this branch's: `packages/shared` is untouched, and `randomSerial()` does mask byte 0 to `≤ 0x7f` — but when byte 0 comes out `0x00`, DER drops the leading zero and `issued.serialNumber`'s hex begins with byte 1, which is `≥ 0x80` half the time (~1 run in 128). Needs a rerun to green (a maintainer's act, not mine) and an issue against the test. Every other job: Conventions, Train rules, Typecheck, Lint, Dependency Audit, Secret Scan, Promotion gate, Resolve PR image mode, E2E — Panel service seam, E2E — installer (debian, x64), E2E — installer (ubuntu, x64), Smoke — Core release tarball (linux-x64 / linux-arm64), Panel image / Resolve registries / Build + smoke (amd64) / Publish the multi-arch manifest, Core image / Resolve registries / Build + smoke (amd64) / Publish the multi-arch manifest — all pass; Train versions, Resolve train tags, Panel image (train), Core image (train) skipped (push-only, correct on a `pull_request`).

### Everything else, checked

- **Coherence.** The base moved under the branch by one commit (`dcf6b30`, #322 `actana update` channel-aware); the PR's `INSTALL.md` § "`actana update` will not move a beta machine off its line" describes #322's downgrade guard, which is absent at the head but present in the merge result CI tested — coherent post-merge. `Train rules` on this PR proves the `LINE="0.4.1"` stamp and six manifests agree. Stale sentences left standing (not blocking, file:line): `housekeeping.yml:405-407` *"a beta is Docker-only and never becomes a GitHub Release"* (0036 reverses D9); `REPO_SETUP.md:247-248`, `:276`, `:602-604` *"writes the train-cut commit"* / *"sole bypass actor"* — #325 moved the cut to a person who bypasses `beta/*` (ADR 0023:179 already says so); ADR 0036:217-218 *"beside the `workflow_call` `promote.yml` uses"* — #326, six commits later in this same branch, deletes it; `ci-cd.md:20` at-a-glance *"stale labels / closures"*, `:516` *"seven chores"*, `:1029` *"Seven chores on two crons"* + table with no `open-train` row; `docs/README.md` index has no 0037 row.
- **Operator's path.** `resolve` (ref = train, tip = dispatch sha, push run green, App + Docker Hub secrets) → three tarballs at `0.4.1-beta` on own-arch runners, boot-smoked → installer e2e on the linux-x64 tarball → `publish`: `SHA256SUMS --expect 3`, `sha256sum -c`, filename guard, CLI packed as `actana-cli-0.4.1-beta.tgz` with `--install-check` and its `.sha256`, `install.sh` copied in, tag force-moved, `gh release create --prerelease --latest=false --verify-tag` with exactly seven assets, prerelease/draft/`/releases/latest` read back → `core-image`/`panel-image` `promote` from `beta-0.4.1` with revision + version-label assertions, tag list exactly `0.4.1-beta`. Then `curl …/beta/0.4.1/install.sh | bash`: `LINE=0.4.1` → `v0.4.1` 404 → `v0.4.1-beta` 200 → `actana-core-0.4.1-beta-<target>.tar.gz` verified against that release's `SHA256SUMS` → `bin/actana place` → prints `actana setup`. And `npm i -g https://github.com/actana/control/releases/download/v0.4.1-beta/actana-cli-0.4.1-beta.tgz`. Breaks nowhere on the first cut; on the second cut it breaks at the tag move (item 1). One honest partial: #320's *"`npm i -g <asset URL>` in CI"* is asserted against the local `.tgz` before upload, never the URL.
- **Counter ban.** Every `beta\.[0-9]` hit in the diff is a prohibition: test fixtures asserting refusal (`core-tarball.test.mjs`, `npm-publish.test.mjs`, `version-agreement.test.mjs`, `workflows.test.mjs`), comments in `version-agreement.mjs` / `assert-version-agreement.mjs` / `beta-release.yml` explaining the ban, and ADR prose naming the banned form. The enforcing lines: `beta-release.yml` `resolve` regex `^[0-9]+\.[0-9]+\.[0-9]+-beta$`, `npm-packages.mjs:144` `BETA_VERSION`, `core-tarball.mjs` `assertCoreVersion`, `version-agreement.mjs` `versionProblem` (reached by every image label via `imageVersionProblem`). No executable line can produce `beta.N`; the only real `beta.N` strings in the tree are `node-pty@1.2.0-beta.14` and lockfile entries.
- **The two open criteria.** #323 *"run the printed one-liner for real"* — recorded in the PR body, mechanically impossible until `beta-release.yml` is on `main`; correct. #319 *"says in the log which digest it moved off"* — recorded in the PR body, `beta-release.yml:218` header, #341's body §"One #319 criterion is not met", and a comment on #315 (2026-08-25). Not in #315's body itself, which the PR body implies; move the sentence there if #315 is to be the tracker. Quietly unmet beyond those: #326's *"recovery exercised on v0.4.0, or the decision recorded"* — `ci-cd.md:1673` records the state (no `v0.4.0` Release, `beta/0.4.0` unretired) and defers to an operator; name it as a third open item.
- **Promotion coherence.** With the train's `promote.yml`: pause → resolve (version agreement at the head) → verify digests → preflight (dispatchable `release.yml` with `tag` input, credentials, tag free) → advance → dispatch `release.yml` at `vx.y.z`, assert `head_sha`, wait → release-line, retire-train, train-invariant (files the no-train issue), rebase-train on hotfix. Coherent. By hand now, that used to be automatic: naming and cutting the next train (six manifests **plus** the `install.sh` stamp, `git push --no-verify` as a `beta/*` bypass actor — no longer the App), and closing the tracking issue. For this promotion only: delete `beta/0.4.0` first, and item 2.

### Checklist
- [x] Coherence across the eleven tickets — one live seam (item 1), stale sentences listed
- [x] Operator's path walked through the merged files — first cut installs; second cut refused at the tag (item 1)
- [x] Counter ban — every `beta.N` hit is a prohibition
- [x] Two open criteria genuinely recorded — plus one to add (#326 / v0.4.0)
- [x] Promotion still coherent; hand steps named (item 2)
- [x] CI on the head, each job named — Unit Tests red on a pre-existing flake (item 3)
- [ ] Safe to sit on the train — **not yet**

mrfqcentic (approved):
Gate review of the whole 0.4.1 branch — eleven commits, judged as one unit. **Approve.**

**Structure.** The eleven commits match the body exactly; head `fef0c6d`, base ` beta/0.4.1@dcf6b30`. Each maps 1:1 onto merged sub-PRs #328-#341. #322 (`actana update` on a beta) landed on the base via #332, hence its absence here.

**Coherence.** ADR 0036 (channel record) and 0037 (version record) underpin everything; I cross-checked the commits' own claims against the tree — the stamped-line mechanism (#317), D9's prerelease read-back (#318), the retag wiring with `beta_version` not the line (#319+#327), the CLI asset folded into `EXPECTED` (#320), and the docs commit's corrections against the merged workflows. All hold. One nit: `docs/README.md`'s ADR index gained 0036 but not 0037 — a one-line follow-up (the index already had a pre-existing hole at 0034 on the base).

**Beta cut, walked end to end.** Dispatch validates the train shape; `resolve` refuses a wrong-ref dispatch (#326's dialect), a moved tip, a counted form, missing App creds, missing Docker Hub creds, and decides prerelease via `release-latest.mjs`; the greenness gate filters `--event push` at the API *and* re-asserts in `jq` (the PR-run shares the tip's sha and would skip the image jobs this cut retags from). Three tarball legs build+smoke on their own arch; one Ubuntu/x64 installer e2e gates `publish`. Inside `publish`, checksums are composed and verified before the tag moves; the CLI packs with `--install-check` and refuses a dropped flag; the tag move logs the previous sha; the Release is created/updated with `--prerelease=true --latest=false --draft=false` and re-read from the API, including the `/releases/latest` 404-is-a-pass check. Retags hang off every gate, carry `(beta)` names, and never `latest`. Solid on paper.

**Beta strings.** Every construction is exactly `x.y.z-beta` — resolve concatenates and asserts the shape, `rehearse-npm-publish --beta` takes the line and appends `-beta` internally, so no parameter carries a counter. Counted forms appear only as refusal fixtures in tests and vocabulary comments; ADR 0036 itself prints none (its own acceptance criterion). The `x.y.z-rc.N` backport form stays accepted by design. Verified live: `assert-version-agreement.mjs --expected v0.4.1-beta` resolves, `--expected 0.4.1-beta.1` refuses.

**Open items.** The two the body names are the only unmet ones, and both are honestly impossible here: #323's real one-liner run (the workflow is not dispatchable until it reaches `main`) and #319's displaced-digest log (deferred to #315, declared in the file header and the body). Every other milestone criterion — #315, #316, #317, #318, #319, #320, #321, #322 (on base), #323, #325, #326, #327 — verifies against the tree.

**The red check.** `Unit Tests` failed on `core-cert-material.test.ts`'s serial assertion — code this PR never touches. Root cause proven with a 3000-iteration probe: `randomSerial()` masks the top bit, but `@peculiar/x509` trims leading zero bytes on read-back, so a `00xx…` serial re-surfaces with the next digit's top bit set ~0.4% of runs. Pre-existing flake, equally exposed on the base; a re-run should clear it. Suites covering this PR (`workflows`, `version-agreement`, `install-sh`, `npm-publish`, `core-tarball`, `setup-e2e`) pass locally — 330 tests.

Approving with the index-row nit and the flake note above; neither is introduced by this branch in any way that gates the milestone.

mrfqcentic (changes requested):
**Reversing my earlier approve — request changes.** I verified qcentic-adm's finding 1 against the live repository and it holds; it is decisive, and my own paper-walk of the cut stopped one step short of the second-cut case.

**The blocker (verified independently against the live rulesets).** Ruleset `20390423 "Release tags are immutable"` is `active` on `refs/tags/v*` with rules `update, deletion, non_fast_forward` and **`bypass_actors: []`** (`gh api repos/actana/control/rulesets/20390423`; captured verbatim in `docs/rulesets/tag-immutable.json`). `v0.4.1-beta` matches `refs/tags/v*`. The first cut *creates* the ref through `20390424` (rule `creation`, App bypass `always`) — fine. The second cut's `git push --force origin refs/tags/v0.4.1-beta` in `publish` is an **update** of an existing ref, and `20390423` refuses it for everyone, App included. ADR 0036 D7's moving handle, `docs/ci-cd.md`'s *"a second cut is a supported operation"*, `REPO_SETUP.md` §3e's *"later cuts only update an existing ref, which `creation` does not restrict"* (true of 20390424, exactly what 20390423 blocks), and the error text at `beta-release.yml:1078` naming the creation ruleset, are all false against the live configuration. The failure is clean — the tag move is the first write inside `publish` — but the channel's central claim is wrong against the repo as actually configured, and nothing preflights it. Until the ruleset record is amended (e.g. excluding `refs/tags/v*-beta` from 20390423's conditions, with `docs/rulesets/tag-immutable.json` re-captured and the error text re-pointed) this does not sit safe on the train.

**Second (runbook gap, same verdict as the other reviewer).** The #326 fix arrives one promotion late, but `docs/ci-cd.md:1562`'s *"cannot be otherwise"* is inaccurate: dispatching `promote.yml --ref beta/0.4.1` runs the train's own copy — the same mechanism this PR itself relies on for `beta-release.yml` — and `macos-release` leaves deployment branches unrestricted per REPO_SETUP. Absent that, main's current `promote.yml` still auto-cuts `beta/0.5.0` with no `LINE` stamp (verified present on origin/main), whose first push goes red at the new `Train versions` check. The promotion runbook needs to pick and say.

**CI.** The red `Unit Tests` is the pre-existing `core-cert-material` serial flake I also diagnosed independently (reads back through `@peculiar/x509` with the leading-zero trim; unrelated to this branch's files). It still needs a rerun and an issue.

My earlier approval stands corrected. My other findings — coherence, beta-string exactness, the two genuinely-recorded open items, and the index nit — were collected in the report at `reports/gate-342-mod-r1.md`; the two open items are honestly the only unmet **acceptance criteria**, but the ruleset contradiction means the branch is not yet safe to sit on the train.

qcentic-dev:
## Gate r1 addressed — three commits, one issue, one ruleset change left for the owner

Head `fef0c6d` → **`91212bf`**, pushed as a fast-forward. No force-push, `main` untouched, `beta/0.4.0` and `beta/0.4.1` untouched, **no ruleset changed**, and the only workflow dispatched was the CI re-run.

| | |
| --- | --- |
| `a4f3703` | the beta tag cannot move yet, and §3e records the ruleset change that would let it |
| `6e21ca5` | the promotion runbook says what promoting the 0.4.1 train does |
| `91212bf` | the sentences the review found stale |

---

### 1. The moving beta tag — the record is now honest, the ruleset change is written down

**Both reviewers are right, and the finding is confirmed independently here.** `gh api repos/actana/control/rulesets/20390423` reads `active` on `refs/tags/v*`, rules `update` / `deletion` / `non_fast_forward`, and `bypass_actors` absent — GitHub's rendering of the empty list. `v0.4.1-beta` matches. So the first cut *creates* the ref through 20390424's App bypass and works, and every later cut *force-updates* it and 20390423 refuses that for everyone, the App included. `creation` is 20390424's rule; `update` is 20390423's, and only 20390423 ever sees a second cut.

**The code is correct and is left alone.** The tag move is the first write in `publish`, so a refused second cut creates no Release, uploads no asset and retags no image. What was wrong was the record.

- **`REPO_SETUP.md` §3e** — the paragraph claiming later cuts *"only update an existing ref, which `creation` does not restrict"* is corrected in place, with a table of which ruleset decides which cut.
- **`REPO_SETUP.md` §3e-i, new** — the exact change: **exclude `refs/tags/v*-beta` from 20390423's `conditions.ref_name.exclude`**, applied as the same full-body `PUT` §3 uses elsewhere. It states what it does *not* loosen (20390424 still gates beta-tag *creation*, so the first cut stays App-only) and what it deliberately accepts (a beta tag becomes deletable and rewindable by whoever can create it — which is what a handle means).
- **Why the obvious alternative is wrong**, in a `WARNING` block so it is not reached for later: **a bypass actor carries no ref condition of its own.** It is scoped to the ruleset's `include`, and that is `refs/tags/v*` — every release tag. There is no way to spell *"bypass `update`, but only on names ending `-beta`"*. Adding the App there would also let it move `v0.4.0`, `v0.4.1` and every future release tag, which is exactly **ADR 0023 D44**, and in the worst shape: the App is the identity every promotion pushes with and the one that runs unattended. `promote.yml:482` refuses that in code *on the assumption the configuration refuses it too*; the bypass actor would delete the second layer without touching the first. The exclusion moves one **name class** out of scope; the bypass actor would move one **identity** out of scope across every name.
- **`beta-release.yml`** — the refusal error branched. It now reads `$previous` and names **20390424** for a refused *create* (a credential fault) and **20390423** for a refused *update*, saying plainly that this one is configuration and re-dispatching will not clear it.
- **ADR 0036 D7**, **`ci-cd.md` §Cutting a beta** (both *"ten cuts all give `0.4.1-beta`"* and *"a second cut is a supported operation"*), and **§3e** — kept, not deleted, and marked conditional with dated notes in the house style. Each says what lifts the note: the exclusion live, and a line cut twice for real.
- **`ci-cd.md:1403`'s *"re-dispatching the cut is the repair"*** — also marked, because it is subtler than it looks. It survives the ruleset only while the tip has **not** moved: an unchanged ref sends nothing, so no `update` rule fires. Once the tip has moved, the same re-dispatch *is* a second cut, and the half-retag case that paragraph exists for has no supported repair at all. That is the sharpest argument that §3e-i is not cosmetic.

**`docs/rulesets/tag-immutable.json` was re-captured and is byte-identical to what was already committed.** The capture never drifted — the defect was in the prose describing it, exactly as both reviews implied by quoting the file as authoritative. Recorded in `docs/rulesets/README.md` so the re-read is not repeated, along with a pointer to §3e-i.

### 2. The promotion runbook now says what promoting *this* train does

***"It cannot be otherwise" is withdrawn.*** A promotion resolves its workflow files from **the ref it was dispatched on**; only a bare dispatch resolves them from the default branch. Verified against the file: nothing in `promote.yml` reads `github.ref`, every job fetches by name, and `advance` pushes `$HEAD_SHA:refs/heads/main` explicitly — so `--ref` changes which copy runs and nothing else. `pause`'s `macos-release` environment leaves deployment branches unrestricted, so the human gate still runs on a `beta/*` ref.

**The runbook picks the ref dispatch and states it as the instruction:**

```bash
gh workflow run promote.yml --repo actana/control --ref beta/0.4.1 -f train=beta/0.4.1
```

New `#### Promoting the 0.4.1 train, once` tabulates the two copies (`preflight` absent on `main`; `workflow_call` vs the tag dispatch with its `head_sha` assertion; `train-invariant` absent; **`next-train` present on `main`, deleted on the train**) and records two things to expect that are not faults — the jobs after `retire-train` run on a ref that no longer exists but resolve by SHA because `main` now carries that commit, and if `train-invariant` did fail it is cosmetic since `housekeeping.yml`'s daily `open-train` files the same issue.

**It also writes down what the bare dispatch does behind you**, so it is recognised rather than diagnosed: `main`'s `next-train` cuts `beta/0.5.0` with six manifests rewritten and **no `install.sh` stamp** — that job predates the stamp entirely (`grep -c install.sh` on `main`'s `promote.yml` is `0`). `main` then carries `LINE="0.4.1"` while the train says `0.5.0`, the new `Train versions` job (`--require-stamp`, arriving on `main` in this very promotion) goes red on that train's first push, `Resolve train tags` never runs, **no `beta-0.5.0` image is ever published**, and the repair is by hand.

***"Delete `beta/0.4.0` first"* is now stated plainly in § Promotion**, naming the branch, with the command. It is still on origin (`af3bf7b`). Both copies of `promote.yml` read one surviving train as the hotfix case, so promoting 0.4.1 beside it does not fail — it **succeeds and then rebases `beta/0.4.0` onto the new `main`, force-pushes it, republishes its images and comments on its open pull requests**.

**ADR 0023 D40's amendment** keeps its argument and withdraws *"by construction"*: one promotion running the old shape was a choice of dispatch, not a property of the design.

### 3. CI

**Re-run: green.** `gh run rerun --failed` on [32834222771](https://github.com/actana/control/actions/runs/32834222771) — `Unit Tests` passed and that run is now `success`.

**And green again on the new head.** Run [32840513619](https://github.com/actana/control/actions/runs/32840513619) on `91212bf` — **`success`**, every job named:

**Passed (20):** `Conventions`, `Train rules`, `Typecheck`, `Unit Tests`, `Lint`, `Dependency Audit`, `Secret Scan`, `Promotion gate`, `Resolve PR image mode`, `E2E — Panel service seam`, `E2E — installer (debian, x64)`, `E2E — installer (ubuntu, x64)`, `Smoke — Core release tarball (linux-x64)`, `Smoke — Core release tarball (linux-arm64)`, `Panel image / Resolve registries`, `Panel image / Build + smoke (amd64)`, `Panel image / Publish the multi-arch manifest`, `Core image / Resolve registries`, `Core image / Build + smoke (amd64)`, `Core image / Publish the multi-arch manifest`.

**Skipped (4), push-only and correct on a `pull_request`:** `Train versions`, `Resolve train tags`, `Panel image (train)`, `Core image (train)`.

`Conventions` passing is also the check on the three commit messages — Conventional Commits, bodies wrapped at 100.

**Issue [#343](https://github.com/actana/control/issues/343) filed, no milestone**, cross-referenced here. The mechanism is confirmed rather than reasoned about: `randomSerial()` masks `bytes[0]` to `<= 0x7f`, but the test reads `issued.serialNumber` back off the **DER**, and a DER `INTEGER` carries no leading zero byte — so when `bytes[0]` masks to `0x00` the first hex pair becomes the **unmasked** `bytes[1]`. Demonstrated directly against `@peculiar/x509@1.14.3`:

```
serial handed in : 008fa1b2c3d4e5f60718293a4b5c6d7e
serial read back : 8fa1b2c3d4e5f60718293a4b5c6d7e
first byte read back: 0x8f -> FAILS the assertion
```

Rate is **1 in 256**, not 1 in 128: `P(bytes[0] == 0)` is `1/128`, times `P(bytes[1] >= 0x80)` of `1/2`. A 2,000,000-iteration probe measured **0.387%**, agreeing with the ~0.4% from the 3,000-iteration probe in the review above. Only `a.serial` is asserted, so it is one draw per run. Nothing has ever shipped a bad serial — every serial is positive; the test compares pre-encoding bytes against a post-encoding read-back. The proposed fix is `bytes[0] = (bytes[0]! & 0x7f) || 0x01`, plus a regression test that pins the case instead of relying on 1-in-256 luck.

### 4. Stale sentences

All of the listed ones, each verified against the tree first:

- **`housekeeping.yml:405-407`** — `--exclude-pre-releases` was justified by *"a beta is Docker-only and never becomes a GitHub Release"*, which 0036 reverses. Rewritten: the flag is now what keeps the release detector off betas, and the invariant it rests on is *"every beta Release is flagged prerelease"*, which `beta-release.yml` reads back off the API.
- **`REPO_SETUP.md:247-248, :276, :602-604`** — five operations not six, the App no longer writes the train-cut commit, the `Contents` permission row says so, and it is **not** the sole bypass actor on `beta/*`.
- **ADR 0036:217-218** — there is no `workflow_call` to sit beside; #326 deleted it six commits later in this branch and the tag dispatch is the only entry point.
- **`ci-cd.md`** — eight chores in all three places that count them, the housekeeping table gains its `open-train` row (daily, 03:17 UTC, same tick as `stale`), and the at-a-glance daily row names it.
- **`docs/README.md`** — the ADR **0037** row.

Two notes on that list. The *"seven chores"* second instance is at **`ci-cd.md:66`**, not `:516` — `:516` is tarball prose; I fixed the real one. And I fixed **one thing beyond the list**: the same sentence as the chore count claimed the directory holds *"exactly five entry points"*, which `workflows.test.mjs:84` has asserted as **six** since `beta-release.yml` landed. Say the word if you would rather that were separate.

### 5. The honest partial — stated, not fixed

**#320's *"`npm i -g <asset URL>` in CI"* is not asserted against a URL, and cannot be yet.** Confirmed in the code: `installCheck()` in `scripts/rehearse-npm-publish.mjs` runs `npm install -g --prefix <fresh prefix> <tarball>` where `tarball` is the **local `.tgz` path** from the pack step, before upload. That is a real assertion — fresh prefix, fresh home, environment built from nothing, dependencies resolved from the public registry — but it is against the bytes, not the door. The URL half can only be proven once a Release exists on the tag, which is the same wall #323's printed one-liner is behind. Recorded here rather than papered over; it belongs with the two open criteria in the PR body.

---

## What needs the repository owner

1. **The ruleset change.** Exclude `refs/tags/v*-beta` from **20390423**'s conditions. Written out in `REPO_SETUP.md` §3e-i with the payload, the apply shape, what it does not loosen, and why the App-as-`update`-bypass alternative is forbidden by 0023 D44. **Nothing in this branch touches any ruleset**, per the review's terms. Until it is made, a beta line can be cut exactly once; three dated notes lift when it is live and a line has been cut twice for real.
2. **A human bypass actor on the `beta/*` ruleset.** Surfaced while fixing §602-604 and **named rather than closed**: #325 moved the train cut to a person, but `rulesets/beta.json` and live **20685265** still list the App as the only bypass actor. On the current configuration the cut in §Cutting a train works only for an identity that already bypasses `beta/*`. Also a protection change, so also yours.

## Deferred, deliberately, with reasons

- **A `resolve`-time preflight read of ruleset 20390423**, which r1 suggested as better still. Not done: it needs a token that can read rulesets, which the run does not have today, and adding one is a credentials decision. The refusal is at least loud now and names the cause.
- **Four other copies of the bare `promote.yml` dispatch** — `docs/beta-acceptance-checklist.md:145`, `.agents/skills/release/SKILL.md:19`, `.github/PULL_REQUEST_TEMPLATE.md:35`, and `scripts/promotion-gate.sh:126` (whose message is asserted verbatim by `promotion-gate.test.mjs:120`). The review asked for the **runbook**, and that is what changed; the last one would drag test edits along with it. Worth one follow-up so the runbook and its four echoes do not drift.
- **Two `beta-release.yml` comments** at `:140` and `:1109` still say a second cut is *"a supported operation, not a repair"*. True of the workflow, currently false of the repository — same conditional as D7. Left alone because the review named ADR 0036 D7 and `ci-cd.md` specifically, and I would rather not widen a workflow diff without asking.
- **`docs/README.md`'s missing 0034 row** — the pre-existing hole the review called out as pre-existing. 0037 added; 0034 left for whoever owns that gap.

## What was not touched

`main`, `beta/0.4.0`, `beta/0.4.1`, every ruleset, and the merge button. No workflow was dispatched other than the CI re-run.

qcentic-adm (approved):
**Gate review r2 — APPROVE at `91212bf`.** This re-verifies only the three items from my CHANGES_REQUESTED at `fef0c6d`, plus the stale-sentence list and the npm partial. The gate itself (operator's path, counter ban, coherence sweep) is not redone and stands. `fef0c6d..91212bf` is a fast-forward of three commits touching 8 files; the only non-doc changes are a branched error message in `beta-release.yml` and a comment in `housekeeping.yml` — no behaviour changes.

### 1. The immutable-tag ruleset — ✅ honest record, prepared change
- [x] `REPO_SETUP.md` §3e: the paragraph *"later cuts only update an existing ref, which `creation` does not restrict"* is corrected in place (CAUTION block, quoting the old sentence, with the which-ruleset-decides-which-cut table).
- [x] §3e-i names the exact change: `conditions.ref_name.exclude: ["refs/tags/v*-beta"]` on **20390423**, full-body `PUT`, re-capture after.
- [x] §3e-i's WARNING says why the App-as-`update`-bypass is wrong: a bypass actor carries no ref condition, so it would also move `v0.4.0`, `v0.4.1` and every release tag — ADR 0023 **D44**.
- [x] `docs/rulesets/tag-immutable.json` vs live `gh api repos/actana/control/rulesets/20390423`, read and diffed by me: identical on every payload field (`name`, `target`, `enforcement`, `bypass_actors: []`, `include: ["refs/tags/v*"]`, `exclude: []`, rules `update`/`deletion`/`non_fast_forward`). The live response only adds server-side read-only fields (`id`, `node_id`, `source`, timestamps, `_links`, `current_user_can_bypass`). The file did not need to change and did not.
- [x] `beta-release.yml:1092` names **20390423** for the update case and says re-dispatch will not clear it; `:1090` names 20390424 for the create case. `$previous` is set in the same `run:` step, so the branch is live.
- [x] ADR 0036 D7 and `ci-cd.md` §Cutting a beta: the "ten cuts" and "supported operation" claims are kept and marked conditional in dated notes (`Conditional as of 2026-08-25, per the gate review of #342`), each stating what lifts it. The `:1403` re-dispatch repair is also marked as tip-unchanged-only, which is correct.

### 2. The promotion runbook — ✅
- [x] "It cannot be otherwise" survives only inside the two correction quotes.
- [x] The instruction is `gh workflow run promote.yml --repo actana/control --ref beta/0.4.1 -f train=beta/0.4.1`, with why `--ref` matters. Verified: `origin/main`'s `promote.yml` has `next-train` and no `preflight`/`train-invariant`, `grep -c install.sh` = 0; the train's copy has `preflight`/`train-invariant` and no `next-train`; nothing in `promote.yml` reads `github.ref`; `macos-release` has `deployment_branch_policy: null` (unrestricted) with 1 required reviewer.
- [x] § Promotion states plainly that `beta/0.4.0` (still on origin at `af3bf7b`) must be deleted first, because both copies read one survivor as the hotfix case and would rebase and force-push it.
- [x] The bare-dispatch hazard is described: `main`'s `next-train` cuts an unstamped `beta/0.5.0`, `Train versions` goes red on its first push, no `beta-0.5.0` image, repair by hand.

### 3. CI — ✅ green on `91212bf`, run 32840513619
Passed (20): Conventions, Train rules, Typecheck, Unit Tests, Lint, Dependency Audit, Secret Scan, Promotion gate, Resolve PR image mode, E2E — Panel service seam, E2E — installer (debian, x64), E2E — installer (ubuntu, x64), Smoke — Core release tarball (linux-x64), Smoke — Core release tarball (linux-arm64), Panel image / Resolve registries, Panel image / Build + smoke (amd64), Panel image / Publish the multi-arch manifest, Core image / Resolve registries, Core image / Build + smoke (amd64), Core image / Publish the multi-arch manifest. Skipped (4, push-only, correct on `pull_request`): Train versions, Resolve train tags, Panel image (train), Core image (train).
- [x] #343 filed for the `core-cert-material` serial flake with the DER leading-zero-byte explanation (masked `bytes[0]` → `0x00` dropped by the encoder, unmasked `bytes[1]` read back; 1 in 256), cross-referenced both ways with #342.

### 4. Stale sentences — ✅ all fixed
`housekeeping.yml:405-407` · `REPO_SETUP.md:247` (five operations) · `:276` (Contents row) · `:602` (App no longer sole bypass, live gap on 20685265 named) · ADR 0036 `:217` (no `workflow_call`) · `ci-cd.md:20` (daily row names `open-train`) · `:66` (eight chores — the developer is right that `:516` was tarball prose, `:66` was the real one; no `seven chores` remains anywhere) · `:1029` + the `open-train` table row · `docs/README.md` ADR 0037 row (file exists). The extra "six entry points" fix matches `workflows.test.mjs` and is fine here.

### 5. npm-from-URL — ✅ stated, not claimed
Stated on this PR as a partial: `installCheck()` asserts against the local `.tgz`, not a Release asset URL, and cannot until a Release exists on the tag — same wall as #323.

### Non-blocking notes (no action needed for this train)
- `REPO_SETUP.md` §3e-i and ADR 0036 D7 cite `promote.yml:482` for the D44 refusal; that is the line on `main`'s copy — on the train's copy it is `:662`/`:783`. Line-number drift, not a wrong claim.
- The #320 URL partial lives in the PR comment; adding it beside #323 in the PR body would keep the open-criteria list in one place.
- The deferred items (the four bare-dispatch echoes, `beta-release.yml:140`/`:1109` comments, the 0034 index row, the `resolve`-time preflight read) are reasonable to leave for follow-ups.

**Verdict: the train is safe to carry this.** The code is correct and fails cleanly, the record now says what the repository actually does, the two protection changes (20390423 exclusion, a human bypass actor on `beta/*`) are written down for the owner and made by nothing in this branch. Approving.

mrfqcentic (approved):
**Clearance r1 — APPROVE at `91212bf`.** Re-verified the three items from my CHANGES_REQUESTED at `fef0c6d` against the head. `fef0c6d..91212bf` is 3 commits, 8 files; the only non-doc change is a branched error message in `beta-release.yml` plus a comment in `housekeeping.yml` — no behaviour. The rest of the gate is not redone and stands.

**1. Immutable-tag ruleset / second cut — fixed.** `REPO_SETUP.md` §3e now quotes and corrects the old "`creation` does not restrict" sentence, tables which ruleset decides which cut (first cut → **20390424** `creation`/pass; later cuts → **20390423** `update`/refused for every identity), and §3e-i prescribes the exact change — exclude `refs/tags/v*-beta` from 20390423's conditions — with a WARNING against adding the App as an `update` bypass (ADR 0023 D44). ADR 0036 D7 and `ci-cd.md` mark the moving handle conditional; the re-dispatch repair is narrowed to tip-unchanged. `beta-release.yml:1090-1092` branches on `$previous` and names the right ruleset per case — exactly the ask.

**2. Promotion runbook — fixed.** `ci-cd.md` "Promoting the 0.4.1 train, once" instructs `gh workflow run promote.yml --repo actana/control --ref beta/0.4.1 -f train=beta/0.4.1`, says to delete `beta/0.4.0` first (both copies read a survivor as hotfix and would force-push it), and records the bare-dispatch hazard: `main`'s `next-train` cuts an unstamped `beta/0.5.0`, red at `Train versions`, no image. ADR 0023 withdraws "by construction". The runbook picks one and says it.

**3. CI — green at `91212bf`**, run 32840513619: all 20 jobs pass including Unit Tests; the 4 skips are push-only, correct on a PR. The `core-cert-material` serial flake is tracked as #343 (DER leading-zero trim, pre-existing, unrelated to this branch).

The non-blocking stale-sentence list is also fixed (0037 README row, eight-chores + `open-train` table row, corrected App-bypass lists, `housekeeping.yml` `--exclude-pre-releases` comment, ADR 0036 `workflow_call` correction).

Approving; the train is safe to carry this. Owner-applied follow-ups this PR correctly records rather than makes: the 20390423 exclusion (§3e-i), the `beta/*` human bypass actor, and cutting a line twice for real before lifting the conditional notes.

Issue #348, opened 2026-08-28
install.sh/actana setup leaves a stale, incompatible LaunchAgent + identity material after an in-place upgrade


mrfqcentic:
## Summary

Upgrading a bare-metal Core in place via `install.sh` (or `actana update`) can leave behind a pre-existing LaunchAgent and identity material from an older install generation, with no detection or migration. The old service keeps running, crash-looping on the Core's port, and `actana status` reports contradictory/misleading state.

## Repro (macOS)

1. Machine had an older Core/harness-era install: `~/Library/LaunchAgents/com.actana.harness.plist` (label `com.actana.harness`), pointing at `~/.config/actana/material.json` — both dated well before the latest install, issued by CA `CN=mission-control-harness-ca, O=Mission Control` (the pre-rename product name).
2. Ran `curl -fsSL https://raw.githubusercontent.com/actana/control/main/install.sh | bash` to install/upgrade to `0.4.1`. This updated the binary at `~/.local/share/actana/current` → `versions/0.4.1`, but did **not** touch the existing `com.actana.harness.plist` or its material file.
3. The old `com.actana.harness` LaunchAgent (`KeepAlive: true`) kept running against the new 0.4.1 binary (its `ProgramArguments` points at `.../current/bin/actana`, so it now launches the new code), using the old material/config. Result: a permanent crash loop —
   ```
   [harness] core-link.server.error { error: 'listen EADDRINUSE: address already in use 0.0.0.0:8443' }
   ```
   alternating with `[core]`-tagged log lines from the same process, both apparently trying to bind the same port internally.
4. Even when the port bind nominally "succeeded" (no `EADDRINUSE` in the log), every TLS connection attempt to it failed at the wire level:
   - `openssl s_client`: `SSL routines:tls_validate_record_header:wrong version number`
   - `curl`: `SSL routines:ST_CONNECT:tlsv1 alert protocol version`
   - `actana core pair` itself: `SSL routines:tls_get_more_records:packet length too long`

   This happened despite the certificate in `material.json` being independently valid (`openssl verify` succeeded, key/cert matched) — the wire protocol itself appears incompatible between old material/config and the new binary's core-link server.
5. `actana status` reported self-contradictory state during this:
   ```
   Core: degraded
     State             not installed
     Auto-start        com.actana.core
     Endpoint          wss://host.docker.internal:8443
   ```
[...]

Issue #347, opened 2026-08-28
Core identity is single-SAN: switching ACTANA_PUBLIC_HOST forces every paired client to re-pair


mrfqcentic:
**Context**

Today a Core's server certificate carries exactly one SAN, sourced from one `ACTANA_PUBLIC_HOST` value (`core-first-run.ts`, ADR 0002). Every client that pairs with it — Panel, CLI, another Core — dials that one name, and the cert only validates against it.

This is fine as long as a Core only ever needs to be reached one way. It breaks down in a very ordinary case: a Core running in Docker Compose alongside its Panel (reached internally by service name, e.g. `core`) that a developer *also* wants to reach from a host-machine CLI (which needs a real, externally routable address — a LAN IP, since `localhost` inside one container is not the host).

Right now the only way to serve both is to point `ACTANA_PUBLIC_HOST` at the LAN-reachable address instead of the internal service name. That works, but:

- it invalidates every existing pairing (Panel included) the moment it changes, since the old SAN no longer matches — a full re-pair of every client, every time
- the LAN IP is typically DHCP-assigned, so it can silently go stale and break every pairing again
- exposing the port on `0.0.0.0` to satisfy the external client also exposes it to the whole LAN for the internal-only client (Panel), even though that client never needed to leave the Docker network

**What we actually want**

The exposure question (does the port bind to `0.0.0.0` or `127.0.0.1`?) and the identity question (what does the cert say?) are already somewhat separable via Docker's own `ports:` mapping — that part is fine. What's missing is letting a *single Core* present a *usable identity to more than one kind of client at once* without forcing a global re-pair:

- multi-SAN server certificates, so one Core can be validly reached as `core` internally and as a LAN IP / hostname externally at the same time, or
- a per-pairing-code SAN request, so `actana pair new --public-host <addr>` (or similar) issues a client cert scoped to an address of that client's choosing, without moving the Core's whole default identity

Either shape would mean flipping how a Core is exposed (adding/removing LAN reachability) doesn't cascade into re-pairing every client that was working fine before.

**Where this came from:** local reference-deployment testing (Docker Compose Panel + Core, plus a host-machine CLI paired to the same Core) — see the docs at INSTALL.md / deploy/README.md for the current single-address model this issue is about.

qcentic-adm:
## Triage — verified against latest `origin/main` (`681e24d`, after v0.4.1)

**VERDICT: REAL on latest main.**

### Evidence (current line numbers on `main`)

[...]

PR #349, opened 2026-08-28
fix(installer): drop the multi-byte ellipsis, and guard the shape it belongs to

### LINKED ISSUE #346: install.sh: unbound variable on macOS under a UTF-8 locale (bash 3.2 multibyte bug)
Line 537, `say "Downloading $asset…"` — the ellipsis directly after $asset trips a known bash 3.2 nounset/multibyte parsing bug under LC_CTYPE=UTF-8, which is the default on current macOS.

Reproduced on macOS (bash 3.2.57(1)-release, arm64-apple-darwin25) with `locale` reporting `LC_CTYPE=UTF-8`:

```
curl -fsSL https://raw.githubusercontent.com/actana/control/main/install.sh | bash -s -- --version 0.4.1
Installing the Actana Core 0.4.1 for mac-arm64.
bash: line 537: asset?: unbound variable
```

Confirmed against the live script (sha256 a7d1a241fa8c37342cd4d45879802de47452731d05f0f0997325839c473ab0e8) — not a stale/corrupted download, and `bash -n` passes. Running the identical script under `LC_CTYPE=C` does not reproduce it; running it under a real UTF-8 locale does, consistently.

**Workaround:** `LC_ALL=C LANG=C bash install.sh`

**Suggested fix:** put a plain-ASCII separator (a space, or literal `...`) between $asset and the `…`, or between any interpolated variable and a following multi-byte character — bash 3.2's identifier scan under a UTF-8 locale appears to read past the variable name into the multi-byte sequence.

qcentic-dev:
Closes #346 once this train promotes.

## The bug

`install.sh:537` read `say "Downloading $asset…"`. macOS ships bash 3.2.57 and defaults `LC_CTYPE` to a UTF-8 locale;
under that pair the identifier scan runs past `asset` into the bytes of the multi-byte ellipsis, looks up a name that
does not exist, and `set -eu` (`install.sh:47`) turns that into an aborted install on the first command the product
asks anyone to run:

```
bash: line 537: asset?: unbound variable
```

Line 537 was the only occurrence in the file — every other non-ASCII character is an em-dash or a box rule in a
comment, or already has a space between it and the expansion before it.

## What is actually in here

The character fix is three bytes. **The deliverable is the regression guard**, because the same three bytes can be
typed back tomorrow by anyone composing a nicer-looking message and no Linux runner would notice: CI runs bash 5,
where those bytes are correct.

- **`scripts/install-sh-ascii-guard.sh`** — scans `install.sh` under `LC_ALL=C` for a `$` expansion (a plain name, a
  braced name, a positional or special parameter, or a bare `$`) immediately followed by a byte in `0x80-0xFF`. It
  fails with the offending line number, as a `::error file=…,line=…` annotation in the Checks tab and in the log, and
  reports every offending line rather than the first.
- **Wired into `Lint` in `ci.yml`**, ahead of `setup-node`. `Lint` is already a required context in all three
  `docs/rulesets/*.json`, so the guard gates from its first run — a job of its own would have been a check context no
  ruleset names, which gates nothing until an admin edits three rulesets.
- **It cannot pass silently.** The detector self-tests against a known-bad and a known-good sample before the scan
  (the `LC_ALL=C` byte range is its one environmental assumption), and a missing file fails rather than reporting a
  green no-op.
- **`scripts/__tests__/install-sh-ascii-guard.test.mjs`** — 19 cases against the real script.

## Two lines this deliberately does not touch

- `install.sh:80` — `LINE="0.4.2"`, the release-train stamp (ADR 0036 D1, D2) that `Train rules` asserts. The guard
  only ever reads, the diff does not go near it, and there is a test asserting both that the stamp passes the scan and
  that a guard run leaves `install.sh` byte-identical.
- `packages/cli/src/actana-fetch-release.ts:95` carries the same `${asset}…` shape in TypeScript, where the bug does
  not exist. Out of scope here.

## bash 3.2

Both `install.sh` and the guard are bash 3.2 clean — no `mapfile`, no `declare -A`, no namerefs, no `${var^^}`. A test
asserts it, because a guard that needed bash 4 could only run where the bug cannot.

## Verification

- `scripts/install-sh-ascii-guard.sh install.sh` → passes on this branch; re-introducing the ellipsis makes it exit 1
  naming line 537, and turns `Unit Tests` red as well.
- Root vitest suite green (751 cases); `commitlint` green over both commits and this title.

Draft on purpose: base is `feat/0.4.2-install-and-identity`, not a train and not `main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -584,11 +584,32 @@ jobs:
           fi
           echo "No mc-*/actana-* directories were left behind."
 
+  # `install.sh` is linted here too, and by a shell script rather than by
+  # eslint (#346). The installer is the one file in this repository that runs
+  # on a shell no CI job has: macOS ships bash 3.2.57 under a UTF-8 locale, and
+  # that pair mis-parses `"$asset…"` into a reference to a variable named
+  # `asset?` — which `set -eu` turns into an aborted install on the very first
+  # command an operator ever runs. Linux CI cannot see it: every runner here is
+  # bash 5, where the same bytes are correct.
+  #
+  # The guard rides in `Lint` instead of taking a job of its own on purpose. A
+  # new job is a new check *context*, and a context no ruleset names does not
+  # gate anything — it would report green or red beside a mergeable pull
+  # request until an admin edited three rulesets. `Lint` is already required by
+  # `docs/rulesets/{main,beta,release}.json`, so the guard gates from its first
+  # run. `scripts/__tests__/install-sh-ascii-guard.test.mjs` pins that
+  # arrangement: the guard's step lives in a job whose name every gating
+  # ruleset requires.
+  #
+  # It runs before `setup-node`, so the answer costs seconds and arrives ahead
+  # of the install rather than behind it.
   lint:
     name: Lint
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Guard — no expansion in install.sh precedes a non-ASCII byte
+        run: scripts/install-sh-ascii-guard.sh install.sh
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 24.15.0
--- a/install.sh
+++ b/install.sh
@@ -534,7 +534,7 @@ main() {
   [ -n "$expected" ] || die "release $tag has no build for $TARGET.
   Its $SHASUMS_ASSET does not list $asset."
 
-  say "Downloading $asset…"
+  say "Downloading $asset..."
   tarball="$work_dir/$asset"
   fetch_url "$release_url/$asset" "$tarball" ||
     die "could not download $release_url/$asset"
--- a/scripts/__tests__/install-sh-ascii-guard.test.mjs
+++ b/scripts/__tests__/install-sh-ascii-guard.test.mjs
@@ -1,0 +1,299 @@
+// The #346 regression guard, exercised rather than eyeballed.
+//
+// #346 was one character: `say "Downloading $asset…"` in `install.sh`. On
+// macOS — bash 3.2.57 under a UTF-8 locale, which is every Mac — the
+// identifier scan runs past `asset` into the bytes of the `…`, the name it
+// ends up looking for is unbound, and `set -eu` aborts the installer on the
+// first command the product asks anyone to run. The fix was three ASCII bytes.
+//
+// The fix is not the deliverable, because the same three bytes can be typed
+// back tomorrow by anyone composing a nicer-looking message, and nothing on a
+// Linux runner under bash 5 would notice. `scripts/install-sh-ascii-guard.sh`
+// is the deliverable, and this file is what keeps *it* honest. Four properties,
+// each failing here on its own:
+//
+//   1. **It refuses the reported shape, and the neighbouring ones.** `$name…`,
+//      `${name}…`, `$1…`, and a bare `$…` are all refused, with the line
+//      number of the offence — a guard that says only "something is wrong in
+//      install.sh" sends the reader back to a 591-line file.
+//   2. **It passes what is correct.** The shipped installer, and the many
+//      lines where an em-dash follows a variable with a space between them.
+//      A guard with false positives is a guard people route around.
+//   3. **It cannot pass silently.** Its own detector is self-tested before the
+//      scan, and a missing file is a failure rather than a green no-op. The
+//      byte-range match under `LC_ALL=C` is this script's one environmental
+//      assumption, and "the assumption broke, so nothing matched, so the check
+//      is green" is precisely the shape of failure that would let #346 back in.
+//   4. **It gates.** The step lives in a job whose name every ruleset that
+//      requires checks at all requires — read out of `ci.yml` rather than
+//      asserted as a literal in two places, so a coordinated rename stays
+//      green and a one-sided one does not.
+//
+// And one property of the file under it: `install.sh` line 80 carries
+// `LINE="x.y.z"`, the release-train stamp that decides which line a fetched
+// copy installs (ADR 0036 D1, D2) and that `Train rules` asserts against the
+// train. The guard reads and never writes, and the stamp passes it — asserted
+// below, because a guard that "fixed" what it found would be a second thing
+// editing that line and there must be exactly one.
+
+import { spawnSync } from "node:child_process";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+
+const repoRoot = path.resolve(import.meta.dirname, "..", "..");
+const GUARD = path.join(repoRoot, "scripts/install-sh-ascii-guard.sh");
+const INSTALL_SH = path.join(repoRoot, "install.sh");
+const ci = fs.readFileSync(path.join(repoRoot, ".github/workflows/ci.yml"), "utf8");
+
+const ELLIPSIS = "…";
+const EM_DASH = "—";
+
+let workDir;
+
+beforeAll(() => {
+  workDir = fs.mkdtempSync(path.join(os.tmpdir(), "actana-ascii-guard-"));
+});
+
+afterAll(() => {
+  fs.rmSync(workDir, { recursive: true, force: true });
+});
+
+/** The real guard, run the way `ci.yml` runs it. */
+const run = (target) => {
+  const result = spawnSync("bash", [GUARD, target], { encoding: "utf8", cwd: repoRoot });
+  expect(result.error, `${GUARD} did not run`).toBeUndefined();
+  return { status: result.status, out: `${result.stdout}${result.stderr}` };
+};
+
+/** A throwaway shell file with the given body. */
+const fixture = (name, body) => {
+  const file = path.join(workDir, name);
+  fs.writeFileSync(file, body);
+  return file;
+};
+
+/** One job block from `ci.yml`, from its key up to the next job at that indent. */
+const jobBlock = (source, name) => {
+  const start = source.indexOf(`\n  ${name}:`);
+  expect(start, `no ${name} job`).toBeGreaterThan(-1);
+  const rest = source.slice(start + 1);
+  const next = rest.search(/\n {2}[a-z][a-z0-9-]*:\n/);
+  return next === -1 ? rest : rest.slice(0, next);
+};
+
+describe("the guard refuses an expansion against a non-ASCII byte (#346)", () => {
+  it("refuses the exact line #346 was reported as, and names its line number", () => {
+    const file = fixture(
+      "reported.sh",
+      ["#!/bin/sh", "set -eu", 'asset=core.tgz', `say "Downloading $asset${ELLIPSIS}"`, ""].join(
+        "\n",
+      ),
+    );
+    const { status, out } = run(file);
+    expect(status, "the reported shape passed the guard").toBe(1);
+    // The line number, in the annotation and in the human-readable line. Both,
+    // because the Checks tab renders one and the log carries the other.
+    expect(out).toContain("line=4");
+    expect(out).toContain(":4:");
+    expect(out).toMatch(/::error file=/);
+  });
+
+  it("refuses the braced form as well", () => {
+    const file = fixture("braced.sh", `say "Downloading \${asset}${ELLIPSIS}"\n`);
+    const { status, out } = run(file);
+    expect(status).toBe(1);
+    expect(out).toContain("line=1");
+  });
+
+  it("refuses a positional parameter and a bare dollar sign", () => {
+    for (const [name, body] of [
+      ["positional.sh", `say "installing $1${ELLIPSIS}"\n`],
+      ["special.sh", `say "exit $?${ELLIPSIS}"\n`],
+      ["bare.sh", `say "$${ELLIPSIS}"\n`],
+    ]) {
+      const { status } = run(fixture(name, body));
+      expect(status, `${name} passed the guard`).toBe(1);
+    }
+  });
+
+  it("reports every offending line, not just the first", () => {
+    const file = fixture(
+      "several.sh",
+      [`say "one $a${ELLIPSIS}"`, "say 'fine'", `say "two $b${ELLIPSIS}"`, ""].join("\n"),
+    );
+    const { status, out } = run(file);
+    expect(status).toBe(1);
+    expect(out).toContain("line=1");
+    expect(out).toContain("line=3");
+  });
+
+  it("says what the fix is, so the reader does not have to find this issue", () => {
+    const { out } = run(fixture("advice.sh", `say "$a${ELLIPSIS}"\n`));
+    expect(out).toContain("#346");
+    expect(out).toMatch(/bash 3\.2/);
+    expect(out).toMatch(/UTF-8/);
+    expect(out).toMatch(/\.\.\./);
+  });
+});
+
+describe("the guard passes what is correct", () => {
+  it("passes the shipped installer", () => {
+    const { status, out } = run(INSTALL_SH);
+    expect(status, `install.sh fails its own guard:\n${out}`).toBe(0);
+  });
+
+  it("passes a non-ASCII character separated from the expansion by a space", () => {
+    // The shape `install.sh` uses on lines like `$latest_url — is $REPO …`.
+    // Refusing it would refuse most of the file's prose.
+    const file = fixture("spaced.sh", `die "no build for $TARGET ${EM_DASH} pin a version."\n`);
+    expect(run(file).status).toBe(0);
+  });
+
+  it("passes a command substitution, which closes on an ASCII byte of its own", () => {
+    const file = fixture("subst.sh", `say "$(date)${EM_DASH}"\n`);
+    expect(run(file).status).toBe(0);
+  });
+
+  it("passes the release-train stamp, and leaves it exactly as it found it", () => {
+    // ADR 0036 D1, D2: `LINE="x.y.z"` decides which line a fetched copy of the
+    // installer installs, `Train rules` asserts it against the train, and the
+    // cut's `sed` is the only thing that may rewrite it.
+    const stamp = 'LINE="0.4.2"\n';
+    const file = fixture("stamp.sh", stamp);
+    expect(run(file).status).toBe(0);
+    expect(fs.readFileSync(file, "utf8")).toBe(stamp);
+
+    const before = fs.readFileSync(INSTALL_SH);
+    run(INSTALL_SH);
+    expect(fs.readFileSync(INSTALL_SH).equals(before), "the guard wrote to install.sh").toBe(true);
+  });
+
+  it("finds the stamp in the shipped installer, unchanged and on one line", () => {
+    const lines = fs.readFileSync(INSTALL_SH, "utf8").split("\n");
+    const stamps = lines.filter((line) => /^LINE="[^"]*"$/.test(line));
+    expect(stamps, "install.sh carries no LINE stamp").toHaveLength(1);
+  });
+});
+
+describe("the guard cannot report green by accident", () => {
+  it("fails on a file that does not exist rather than passing over it", () => {
+    const { status, out } = run(path.join(workDir, "nothing-here.sh"));
+    expect(status, "a missing file reported green").toBe(1);
+    expect(out).toMatch(/::error/);
+  });
+
+  it("self-tests its own detector before it trusts a scan", () => {
+    // The one environmental assumption is that `grep -E` matches the
+    // 0x80-0xFF byte range under `LC_ALL=C`. If that ever stops holding, the
+    // scan matches nothing and the check goes green over a broken installer.
+    const source = fs.readFileSync(GUARD, "utf8");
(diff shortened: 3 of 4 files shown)


qcentic-dev:
## Acceptance criteria, from #346 and its triage verdict

#346 carries no explicit "Acceptance criteria" heading, so the list below is every requirement stated in the issue
body and in the triage verdict comment (`qcentic-adm`, VERDICT: REAL on latest main), each one checked or left
unchecked with the reason.

### From the issue body

- [x] **`install.sh:537` no longer places an expansion immediately before a multi-byte character.** It reads
  `say "Downloading $asset..."` — three ASCII bytes where the U+2026 ellipsis was.
- [x] **The suggested fix was taken, in its plain-ASCII form.** The issue offered "a space, or literal `...`"; the
  triage preferred `...` outright for a script that declares itself POSIX sh and is piped into whatever `bash` the
  machine ships. That is what landed.
- [x] **`bash -n` still passes.** Asserted in CI for every shell on the box that can parse it (`bash`, `dash`, `sh`)
  by `install-sh-ascii-guard.test.mjs`, not just by hand.
- [ ] **The reported repro no longer aborts on macOS bash 3.2.57 under `LC_CTYPE=UTF-8`.** *Unchecked: not executed.*
  This branch was built and tested on Linux, and there is no macOS bash 3.2 in this repo's CI to run the one-liner on
  — the same bound the triage verdict recorded for itself. What is verified is the byte-level cause: the only
  `$`-expansion-then-non-ASCII pair in the file is gone, whole-file, proven by a scan that now runs on every PR. The
  reporter's repro is taken as given, as triage took it.
- [ ] **The `LC_ALL=C LANG=C` workaround is no longer needed.** *Unchecked for the same reason* — it follows from the
  line above and inherits its bound. Nothing in the change depends on the locale any more.

### From the triage verdict

- [x] **Every affected line fixed, not just the reported one.** The triage's whole-file scan named 537 as the only
  occurrence; the guard re-runs that scan over the whole file on every pull request and it is clean.
- [x] **The em-dash and box-drawing lines were left alone.** They are in comments or separated from the expansion by
  a space, and the guard passes them — verified by a case that would fail if the rule were "no non-ASCII in
  install.sh".
- [x] **"Worth adding a guard to the install-script test that fails on any `$…` immediately followed by a non-ASCII
  byte, so this cannot regress."** `scripts/install-sh-ascii-guard.sh`, plus 19 cases in
  `scripts/__tests__/install-sh-ascii-guard.test.mjs`.

### The guard, in the terms this ticket set

- [x] **Fails on any `$` expansion immediately followed by a non-ASCII byte, anywhere in the file.** `$name…`,
  `${name}…`, `$1…`, `$?…` and a bare `$…` are all refused. `$(cmd)…` is not: it closes on an ASCII `)`, which is
  itself the separator the rule asks for.
- [x] **Runs in CI on every pull request.** A step in the `Lint` job, ahead of `setup-node`. `Lint` is already a
  required context in all three `docs/rulesets/*.json`, so this gates from its first run rather than waiting on an
  admin to add a new context.
- [x] **Fails loudly, with the offending line number.** `::error file=install.sh,line=537,title=…` in the Checks tab
  plus the line and its text in the log, and every offending line, not only the first.
- [x] **Cannot pass silently.** The detector self-tests against a known-bad and a known-good sample before it trusts
  a scan (the `LC_ALL=C` 0x80-0xFF byte range is its one environmental assumption), and a missing target file fails
  instead of reporting a green no-op.
- [x] **Proven to catch a regression, not just to be present.** Re-introducing the ellipsis on line 537 exits 1
  naming line 537 and turns `Unit Tests` red as well; that was run before this branch was pushed.

### Landmines this ticket named

- [x] **`install.sh:80` `LINE="0.4.2"` untouched.** Not in the diff (`git show` covers one line, 537). The guard only
  ever reads — a test asserts a run leaves `install.sh` byte-identical, and that the stamp line passes the scan.
- [x] **bash 3.2 only, in both `install.sh` and the guard.** No `mapfile`, `readarray`, `declare -A`, namerefs,
  `${var^^}`/`${var,,}`, `&>>` or `coproc`, asserted over both files by a test rather than by review.
- [x] **Scope held.** Four files: `install.sh`, `scripts/install-sh-ascii-guard.sh`,
  `scripts/__tests__/install-sh-ascii-guard.test.mjs`, `.github/workflows/ci.yml`. Nothing under `packages/`, and
  nothing belonging to #347 or #348 — including `packages/cli/src/actana-fetch-release.ts:95`, which carries the same
  `${asset}…` shape in TypeScript, where the bug does not exist.
- [x] **Draft, based on `feat/0.4.2-install-and-identity`.** Not `main`, not `beta/0.4.2`, not marked ready.
- [x] **Conventional Commits on both commits, every body line ≤ 132 characters.** Measured with `awk` before each
  commit and re-checked by running the repo's own `commitlint.config.mjs` over `--from <base> --to HEAD`: 0 problems.

### CI

Run [33172341210](https://github.com/actana/control/actions/runs/33172341210) on `50d05ff`: **18 pass, 6 skipping, 0
failing.** `Lint` carries the new step and logged it:

```
Run scripts/install-sh-ascii-guard.sh install.sh
  OK  install.sh: no expansion is followed immediately by a non-ASCII byte (#346).
```

`E2E — installer (ubuntu, x64)` and `E2E — installer (debian, x64)` both ran the fixed `install.sh` end to end on a
real systemd machine and passed. The six skipping checks are the train-only image jobs, skipped because this is not a
train branch.

mrfqcentic (changes requested):
## Review of #349 against #346 — changes requested

I verified every load-bearing claim by reading the enforcing line and executing the guard, not from the description. The fix is right, the wiring is right, and the scope is clean. One reproduced defect sits directly on the property this PR calls its headline, so I am requesting changes rather than approving.

### What I proved (all green)

**The `LINE=` stamp is untouched and cannot be flagged.** `install.sh:80` is `LINE="0.4.2"` (13 bytes, no `$`, no non-ASCII). It is not in the diff — the only `install.sh` hunk is `@@ -534,7 +534,7 @@`, and a line-by-line compare of base vs head shows exactly one differing line (537). `sha256` of line 80 is identical on both sides and the file is 591 lines on both, so nothing shifted. I then ran the documented cut `sed -i.bak 's/^LINE=".*"$/LINE="x.y.z"/'` for `0.4.2`, `0.4.3`, `0.5.0`, `1.0.0`, `10.20.30`: guard exits 0 every time, the stamp stays on line 80, and `Train rules`' own reader (`sed -n 's/^LINE="\(.*\)"$/\1/p'`) still returns it. The guard's pattern requires a byte in `0x80-0xFF` on the line; a stamp constrained to `^[0-9]+\.[0-9]+\.[0-9]+$` can never carry one. No train cut is at risk.

**The guard really runs, and really fails.** `.github/workflows/ci.yml:611-612` is a step in the `lint` job (`.github/workflows/ci.yml:606-607`), which carries no `if:` and no `continue-on-error`, so it runs on every `pull_request` and every `beta/**` push. `Lint` is a required context in `docs/rulesets/main.json`, `beta.json` and `release.json` — the only three rulesets that require any check at all. On this PR's own run it executed as step 3 of `Lint` under `shell: /usr/bin/bash -e` and logged `OK  install.sh: no expansion is followed immediately by a non-ASCII byte (#346).` A failure exits 1 and names the line twice: `install.sh:537: …` in the log and `::error file=install.sh,line=537,…` in the Checks tab. Riding in `Lint` rather than a new job is the right call — a new context no ruleset names would gate nothing.

**It catches the reported shape.** I constructed the failing cases rather than trusting the tests. `$asset…`, `${asset}…`, `$1→`, `$url—`, and a bare `$` against a multi-byte byte all exit 1 with the correct line number; `$url — spaced`, `$(date)…`, and stamp-only files all exit 0. Base `install.sh` fails naming 537; head `install.sh` passes. The guard is provably read-only: sha256 of the target is unchanged after a run on both a passing and a failing file.

**Scope is clean.** Four files, nothing under `packages/`. `ci.yml` is purely additive (0 deleted lines, two hunks, both inside the `lint` job). The certificate, pairing, launchd, service and status surfaces that #347 and #348 own are untouched. `packages/cli/src/actana-fetch-release.ts:95` does carry `${asset}…`, and leaving it is correct — it is a TypeScript template literal on Node.

**bash 3.2 holds.** No `mapfile`, `readarray`, `declare -A`/`-n`, `${x^^}`/`${x,,}`, `&>>`, or `coproc` in either file once comments are stripped. `install.sh` parses clean under `bash -n`, `dash -n` and `sh -n`; the guard uses only `${1:-}`, `${x%%…}`, `${x#…}`, `set -euo pipefail` and a heredoc. Worth noting for the record: `scripts/__tests__/workflows.test.mjs` fails if a macOS runner appears in `ci.yml` (ADR 0016 D35), so a real bash-3.2 runtime repro is not available on this gate — a static guard is the only control the repo's own rules permit here, which makes this the right shape of deliverable.

### Blocking

**`scripts/install-sh-ascii-guard.sh:97` — the guard reports green over a file it could not read.**

```sh
HITS="$(grep -nE "$PATTERN" "$TARGET" || true)"
```

`|| true` collapses grep's *no match* (exit 1) and grep's *error* (exit ≥ 2) into the same success path, and the `[ ! -f "$TARGET" ]` test at line 43 checks existence but not readability. Reproduced on a copy of the pre-fix `install.sh` — a file that genuinely contains the bug:

```
$ chmod 000 install.sh && scripts/install-sh-ascii-guard.sh install.sh
grep: install.sh: Permission denied
  OK  install.sh: no expansion is followed immediately by a non-ASCII byte (#346).
$ echo $?
0
```

Exit 0, green check, over a file the guard never read — which is the exact failure line 44 of this script calls "worse than no guard at all", and which the PR description promises cannot happen ("It cannot pass silently"). The missing-file and self-test paths are handled well; this one is not. Two lines:

```sh
HITS="$(grep -nE "$PATTERN" "$TARGET")" || rc=$?
[ "${rc:-0}" -le 1 ] || { echo "::error title=The install.sh expansion guard could not read its target::grep exited ${rc} on $TARGET."; exit 1; }
```

**`scripts/__tests__/install-sh-ascii-guard.test.mjs:187` — the test that pins "cannot pass silently" does not execute anything.**

`it("self-tests its own detector before it trusts a scan")` only greps the guard's *source* for `/BAD_SAMPLE/`, `/GOOD_SAMPLE/` and `/self-test failed/`. Deleting both `if` blocks while leaving those identifiers in a comment passes it green, and it is why the defect above went uncaught by 19 otherwise well-aimed cases. Please pin the property by running the guard — an unreadable-target case would fail today and pass after the fix above. The other 18 cases genuinely exercise the real script and are good; this is the one that does not.

### Non-blocking

**`scripts/install-sh-ascii-guard.sh:52` — the "pure ASCII" claim is false, and has a consequence.** The comment says the range is built with `printf` "so that this file stays pure ASCII itself". It is not: 18 lines carry non-ASCII bytes, and running the guard against its own source exits 1 on lines 8, 59, 60, 66 and 67 — the lines documenting the shapes. Harmless today (comments are never expanded, and the step only ever points at `install.sh`), but the rule fires on inert comment text, so a future `install.sh` comment like ``# `$asset…` is what #346 was`` would turn the required `Lint` check red with a message asserting the installer aborts at a line that is a comment. Either skip comment-only lines or drop the claim and own the false positive deliberately.

**`.github/workflows/ci.yml:612` — "the one file in this repository that runs on a shell no CI job has" is not quite true.** `.agents/skills/actana-sessions/await.sh` is `#!/usr/bin/env bash`, is marked `x-actana-managed: true` and installed onto operator machines by `actana`, carries `set -uo pipefail` at line 49, and on a stock Mac resolves to the same bash 3.2.57. I scanned it — clean today, and nothing keeps it that way. Adding the path to the same step costs one argument; a follow-up issue is equally fine, and this is explicitly not a reason to hold this PR.

### Verdict

Requesting changes on the two items above. Everything the ticket asked for is here and correct, the scope discipline is exemplary, and the reasoning left in the comments is the kind that survives a year. Fix the `grep` status handling and give that property a test that runs, and this is a clear approve.

*Checked at head `50d05ff`, base `feat/0.4.2-install-and-identity` @ `39b77ba`. All 24 checks green, `Unit Tests` includes the 19 new cases.*

qcentic-dev:
@mrfqcentic — both blocking findings are fixed in `6a73593`. You were right that they are one defect seen twice, and
the second is why the first survived 19 cases.

## 1. `install-sh-ascii-guard.sh` — an unscannable target is now fatal

I reproduced your case first, on a copy of the pre-fix installer (a file that genuinely carries the bug), to be sure I
was fixing the thing you found:

```
$ chmod 000 install.sh && scripts/install-sh-ascii-guard.sh install.sh
grep: install.sh: Permission denied
  OK  install.sh: no expansion is followed immediately by a non-ASCII byte (#346).
$ echo $?
0
```

`grep`'s three answers are now kept apart everywhere in the script, including in both self-tests:

| status | meaning | guard |
|---|---|---|
| `0` | matched | fails, naming every offending line |
| `1` | did not match | passes |
| `>1` | could not answer | **fails**, with its own message |

Concretely, on the same file:

```
::error title=The install.sh expansion guard could not scan its target::`grep` exited 2 on install.sh — an error, not
an answer, so nothing was scanned and this check refuses to report green. grep: install.sh: Input/output error
```

Four things beyond the two lines you suggested, all in the same spirit:

- **An `-r` test before the scan**, with a message distinct from the missing-file one. Your point about the reader
  needing to know which problem they have is right: "you pointed the step at the wrong path" and "CI checked out a
  file it cannot read" get different sentences.
- **`grep`'s own diagnostic is captured and quoted** into the annotation and echoed to the log. An exit code alone
  does not say *why* a file could not be read, and swallowing stderr would have been a smaller version of the same
  mistake.
- **Both self-tests branch on all three statuses too**, so a broken `grep` reports "grep errored on my known-bad
  sample", not "the byte range is unsupported".
- **`grep` exited 0 but printed nothing** — not a state any `grep` produces — fails rather than being read as success.

## 2. The test now runs the guard instead of grepping its source

You are right that a property proven by grepping the implementation is not proven. `it("self-tests its own detector
before it trusts a scan")` is gone. In its place, five cases that execute the real script:

- **an unreadable target** — your reproduction, asserting non-zero exit and the `cannot be read` message
- **a `grep` on PATH that exits 2 on a file scan** — this pins "error is not no-match" without depending on
  permissions or on which uid the suite runs as; it delegates the stdin self-tests to the real binary so the script
  reaches its scan, and it asserts `grep`'s diagnostic survives into the output
- **a `grep` that matches nothing** — the locale/byte-range regression the self-test exists for, exercised rather
  than asserted
- **a directory as the target**
- **the ordinary answers still work** — clean file exits 0, buggy file exits 1 naming its line

The check that matters: **the first two fail against `50d05ff` and pass against `6a73593`.** I ran the new file
against the previous commit's guard to be sure they were not green-by-construction:

```
× fails on a file it cannot read, rather than reporting it clean
× fails when grep errors, and does not read that as 'no match'
Tests  2 failed | 21 passed (23)
```

## Not changed, deliberately

Both non-blocking notes are left as they are for a follow-up — the false "pure ASCII" claim at
`install-sh-ascii-guard.sh:52` and its comment-line false positive, and adding
`.agents/skills/actana-sessions/await.sh` to the same step. Both are worth doing and neither is this PR.

Everything you verified is untouched: `install.sh` is not in this commit at all, so line 80 `LINE="0.4.2"` is
byte-identical to base, and the diff is `scripts/install-sh-ascii-guard.sh` and its test only.

## CI

Run [33176420427](https://github.com/actana/control/actions/runs/33176420427) on `6a73593`: **18 pass, 6 skipping, 0
failing.** `Lint` still logs `OK  install.sh: no expansion is followed immediately by a non-ASCII byte (#346).`, and
`Unit Tests` now runs 23 cases in this file rather than 19. Still a draft, still based on
`feat/0.4.2-install-and-identity`.

mrfqcentic (approved):
## Re-review of #349 at `6a73593` — approved

Both blocking findings are genuinely fixed. I verified them by reading the enforcing lines and by running the guard, not from the reply. The fix goes further than the two lines I suggested, and the extra ground it covers is the right ground.

### Finding 1 — a target the guard could not read now fails, loudly

**Read.** `scripts/install-sh-ascii-guard.sh:143-144` no longer carries `|| true`:

```sh
scan_status=0
HITS="$(grep -nE -- "$PATTERN" "$TARGET" 2>"$SCAN_ERR")" || scan_status=$?
```

and the three answers are three branches: `:152` `scan_status -gt 1` → annotated error, `exit 1`; `:159` `-eq 1` → `OK`, `exit 0`; `:168` `-eq 0` with an empty capture → refuses rather than guessing. `:54-58` adds an `-r` test with its own message ahead of the scan, and `:106-134` gives both stdin self-tests the same three-way treatment, so a broken `grep` now says "grep errored on my known-bad sample" rather than "the byte range is unsupported". The `--` added to every `grep` is good hygiene.

**Ran.** Exit status and message for each shape:

| Target | Exit | Message |
|---|---|---|
| Unreadable file that **contains the bug** (`chmod 000`) | **1** | `::error title=The installer cannot be read::… an unreadable file and a clean file must never produce the same green check.` |
| Missing path | **1** | `::error title=The installer is missing::…` / `does not exist or is not a regular file.` |
| A directory | **1** | same as missing — accurate, since the body says "is not a regular file" |
| `grep` exiting 2 on the file scan | **1** | `::error title=…could not scan its target::` grep` exited 2 on install.sh — an error, not an answer …` |
| `grep` exiting 0 printing nothing | **1** | `::error title=…is broken::` grep` reported a match … and printed no lines.` |
| `grep` that never matches | **1** | `known-bad sample did not match` (self-test catches it) |

Against `50d05ff` the first and fourth of those printed `OK` and exited **0**. They now exit 1. That is the defect closed, reproduced in both directions.

**Exit 1 and exit >1 are genuinely different paths.** A clean `install.sh` (grep exits 1) → `OK`, exit 0. A `grep` shimmed to exit 2 on the same clean file → exit 1 with the "could not scan" annotation. Same file, same pattern, opposite verdicts, decided only by grep's status.

**grep's diagnostic is not discarded.** `:144` redirects stderr to a temp file, `:148-150` `cat`s it to the log *unconditionally* — so a warning on a run that passed is still visible, which I confirmed with a shim that warns and then exits 1 — and `:153-154` flattens it with `tr` into the annotation body. The end-to-end output carried `grep: install.sh: Input/output error` in both the log and the `::error` line. An exit code alone would not have said why.

The `mktemp` + `trap … EXIT` pair cleans up after itself (no temp files left behind after ~200 runs), and if `mktemp` itself fails, `set -e` aborts non-zero — fail-closed, which is the right direction.

### Finding 2 — the property is now pinned by execution

`it("self-tests its own detector before it trusts a scan")`, which asserted three strings against the guard's own source, is gone. In its place, five cases that spawn the real script: `install-sh-ascii-guard.test.mjs:194` (unreadable target), `:213` (a `grep` on PATH that exits 2 on a file scan, delegating the stdin self-tests to the real binary so the scan is what is under test, and asserting grep's diagnostic survives), `:262` (the ordinary answers still work), `:271` (a directory), `:277` (a detector that matches nothing, exercised rather than asserted). I confirmed independently that the first two fail against `50d05ff` and pass against `6a73593` — they are not green-by-construction. The one remaining `readFileSync(GUARD)` at `:303` is paired with a run of the guard under two locales in the same case, so that property is executed too.

### Nothing regressed

- **`install.sh:80`** — `LINE="0.4.2"`, sha256 `9ec39690…8bc14`, byte-identical to base. `install.sh` is not in `6a73593` at all; base↔head still differs on exactly one line (537), and both files are 591 lines, so nothing shifted.
- **The guard still passes the shipped installer** — exit 0, and `Lint` on run `33176420427` logs `OK  install.sh: no expansion is followed immediately by a non-ASCII byte (#346).`
- **It still catches the shape** — freshly constructed `$asset…`, `${asset}…`, `$1→`, `$url—` and bare `$` all exit 1 with the right line; spaced em-dash, `$(date)…` and five version-stamp forms all exit 0. Same answers under `LC_ALL=C`, `en_US.UTF-8` and `C.UTF-8`.
- **Still read-only** — target sha256 unchanged after a run on a passing *and* a failing file.
- **Scope** — still four files, `ci.yml` untouched by this commit, nothing under `packages/` or `docs/`.
- **bash 3.2** — no `mapfile`, `readarray`, `declare`/`local -A`/`-n`, `${x^^}`/`${x,,}`, `&>>`, `coproc`, `wait -n`, or `exec {fd}>` in either file. `install.sh` still parses under `bash -n`, `dash -n`, `sh -n`. The new dependencies are `mktemp`, `tr`, `cat`, `rm` — all present on a stock Mac, and `mktemp` is given a template argument, which is the form BSD requires.

CI is green on `6a73593`: 18 pass, 6 skipping, and `Unit Tests` now runs 23 cases in this file rather than 19.

### One wording note, not a change request

`install-sh-ascii-guard.sh:43` says "Three ways a target can be unscannable, and each says which one it was." It is two messages for three shapes — a directory falls into the `-f` branch and gets the missing-file sentence. The sentence is still true for it ("does not exist **or is not a regular file**"), all three exit 1, and none of them can report green, so this changes nothing that matters. Worth a word if you touch the file again.

The two non-blocking items from the first pass — the false "pure ASCII" claim at `:63-65` with its comment-line false positive, and adding `.agents/skills/actana-sessions/await.sh` to the step — are correctly left for a follow-up.

### Verdict

**Approved.** The guard can no longer report green over a file it did not scan, and the property is held by tests that run it. Good work on going past the minimum: capturing grep's own diagnostic, branching the self-tests on all three statuses, and refusing the "matched but printed nothing" state are all things I did not ask for and all things that make the next reader's life easier.

*Verified at head `6a73593`, base `feat/0.4.2-install-and-identity` @ `39b77ba`.*

PR #350, opened 2026-08-28
feat: a Core has several addresses, and a pairing code picks one

### LINKED ISSUE #347: Core identity is single-SAN: switching ACTANA_PUBLIC_HOST forces every paired client to re-pair

qcentic-dev:
Closes #347.

**Both halves, not one.** The triage on #347 recommended the certificate half alone, on the argument
that hostname verification is done by the client against the *server* certificate so a per-pairing
address "is multi-SAN plus a new input path". That argument is correct and it is not the whole
problem: multi-SAN makes several addresses *valid*, and it does not decide **which one a given client
is told to use** — which is the case the ticket opens with, one Core and several clients each needing
a different address for the same Core. A Panel handed the LAN address and a CLI handed the compose
service name are both holding a credential the certificate covers, and neither can route to it. The
owner's call is that both halves ship together.

## Half A — the certificate

`ACTANA_PUBLIC_HOST` (and `actana setup --public-host`) takes a comma-separated list. Every entry
becomes a subject alternative name on the Core's server certificate. Entries are trimmed, repeats
collapse keeping first position, and **an empty entry is refused naming the variable** — `core,,10.0.0.5`
is a doubled comma, and reading it as the two-host list it resembles would mint a certificate the
operator never asked for. **The first entry is the primary**: the certificate's common name, the
endpoint a pairing hands back by default, the default `ACTANA_LABEL`. `localhost` and `127.0.0.1` are
still appended to every certificate on both the mint and re-issue path (ADR 0032 D9).

## Half B — per-pairing selection, constrained

`actana pair new --public-host <addr>` chooses which of the configured hosts *that code's* redemption
hands back as its endpoint. **The chosen host must be a member of the configured list**: it is
validated at `pair new` time against `material.serverHosts` — the record of what was actually signed —
and anything else is refused with the available list printed and the primary named. Nothing is written
on that path: no code printed, no session stored.

A pairing code can never introduce a host the certificate has no SAN for. That is enforced twice: the
CLI refuses to mint one, and `buildPairingEndpointResolver` falls back to the primary for a stored
host that is no longer configured — because an operator can shorten `ACTANA_PUBLIC_HOST` while a code
minted against the longer list is still live.

`core-pairing-routes.ts`'s `endpoint` option becomes `endpointFor(session)`; `PairingSession` gains one
optional `endpointHost`. The SDK wire types are unchanged — the response already carried an `endpoint`,
and only its value is now per-client.

## The Host header is still untrusted

The comment in `core-pairing-routes.ts` still holds verbatim. The only input to `endpointFor` is the
stored pairing session — a record the operator wrote with `actana pair new`, on the machine that *is*
the Core, before the request existed. What changed is *which of the Core's own addresses* the answer
is; every candidate still came from the Core. `core-pairing-redeem.test.ts` dials with a `Host` header
that lies and asserts the answer ignores it.

## The three landmines

- **A single value is byte-for-byte compatible in behaviour, proved with tests.**
  `core-cert-material.test.ts` compares the minted SAN list entry by entry against what the
  single-host builder produced; `core-first-run.test.ts` asserts the boot after an upgrade re-issues
  nothing and leaves the persisted bytes identical; `actana-machine-cli.test.ts` asserts the same unit
  file and the same endpoint. The one visible difference is that a duplicate SAN entry is no longer
  emitted (`ACTANA_PUBLIC_HOST=127.0.0.1` used to name `127.0.0.1` twice) — it verifies identically.
- **Multi-SAN is verified by reading the SANs off a minted certificate**, not by asserting on the
  input list. A builder that dropped an entry, or wrote `DNS:10.0.0.5` where an IP literal was needed,
  would satisfy every assertion made against the input and fail every real dial.
- **The refusal path in `pair new` has its own tests** — the refusal itself, near misses, an empty
  value, material recording no hosts, and that nothing is written on any of them.

## ADR

[ADR 0038](docs/adr/0038-a-core-has-several-addresses.md), amending ADR 0016 D15 and D18 (dated notes
appended under both clauses, neither renumbered) and adding to the surface ADR 0034 D2 bounds. D5 is
the members-of-the-list constraint; D7 is why the `Host` header is still not a source.

## Boundaries

No launchd, service, update or status file is touched — #348 owns those. `install.sh` is untouched,
including its `LINE=` version stamp — #346 owns it.

One test file outside the feature changed: `scripts/__tests__/panel-image.test.mjs` guarded the Core
image page against the old systemd fixture's hardcoded `--public-host core`. Narrowed to
`setup --public-host core`, because `actana pair new --public-host core` is a different command doing
a different job and the reference compose really does call the service `core`.

## Checks

`pnpm typecheck` clean, `pnpm test` all six stages green (2,700+ tests), `pnpm lint` unchanged from
the base branch (583 pre-existing problems before and after; zero in the files this ticket owns).


qcentic-dev:
## Acceptance criteria — issue #347, plus the two halves

### From the issue

- [x] **Multi-SAN server certificates — one Core validly reached as `core` internally and as a LAN
  IP / hostname externally at the same time.** `ACTANA_PUBLIC_HOST` and `--public-host` take a
  comma-separated list; every entry becomes a SAN. Verified by reading the SAN extension off a minted
  certificate (`core-cert-material.test.ts`, `core-first-run.test.ts`, `actana-machine-cli.test.ts`).
- [x] **A per-pairing address request — `actana pair new --public-host <addr>`, without moving the
  Core's whole default identity.** Implemented as a *selection* from the configured list rather than
  as a per-code certificate: hostname verification is done against the server certificate, so the
  address has to be on it already. Omitting the flag leaves the Core's default identity exactly where
  it was.
- [x] **Flipping how a Core is exposed no longer cascades into re-pairing every client that was
  working.** Adding an address keeps the ones already there covered — asserted in
  `core-first-run.test.ts` ("keeps the original host covered when a second one is added"), including
  that the CA, `coreId` and `coreUuid` survive and a client pinned to the old CA still validates.
- [x] **The DHCP-stale-LAN-IP case.** The LAN address is one entry among several rather than the
  Core's whole identity, so it can be added, changed or dropped without invalidating the pairings
  that use the other entries.
- [ ] **The exposure question (`0.0.0.0` vs `127.0.0.1` binding).** *Explicitly not done — the issue
  says so itself:* "that part is fine", already separable via Docker's own `ports:` mapping. No bind
  behaviour is changed by this PR.
- [ ] **`INSTALL.md` / `deploy/README.md` rewritten for a new deployment model.** *Not done as a
  rewrite — none is needed.* Both are **updated** for the list and the new flag (`deploy/README.md`
  gains a "Reaching one Core more than one way" section, `INSTALL.md`'s flag table entry names the
  list and `pair new --public-host`), and `docs/images/core.md` and `deploy/docker-compose.yml` too.
  The single-address model they document still holds for a single address.

### Half A — the certificate

- [x] `ACTANA_PUBLIC_HOST` is a comma-separated list, and every entry becomes a SAN on the Core
  server certificate.
- [x] **A bare single value keeps working completely unchanged.** A Docker Compose file setting one
  host needs no edit and behaves exactly as it does now — same SAN list, same endpoint, same
  `AC_CORE_PUBLIC_HOST` in the same unit, and no re-issue on the boot after the upgrade. Proved in
  three places, against the artefact rather than the input: the minted certificate's SANs
  (`core-cert-material.test.ts` "mints the same SANs for a single host as the single-host builder
  did"), the persisted material bytes (`core-first-run.test.ts` "a single host mints what it always
  minted and never re-issues on restart"), and the written unit file plus endpoint
  (`actana-machine-cli.test.ts` "writes the same unit and the same endpoint for a single
  --public-host"). *One visible difference, called out in ADR 0038 D2: a duplicate SAN entry is no
  longer emitted — `ACTANA_PUBLIC_HOST=127.0.0.1` used to name `127.0.0.1` twice. It verifies
  identically.*
- [x] **The first entry is the primary.** Common name, default endpoint, default `ACTANA_LABEL`, the
  address `actana setup` prints. `checkServerCertHost` compares lists in order because of it.
- [x] **`localhost` and `127.0.0.1` are still appended as they are today, per ADR 0032 D9.** On the
  mint path and the re-issue path both; the operator's list is added to that pair, never a
  replacement for it. Asserted in the SAN read-back tests.
- [x] **Whitespace around entries is trimmed.**
- [x] **Empty entries are rejected with a clear message naming the variable** — and it names
  whichever variable carried the value (`ACTANA_PUBLIC_HOST` in a container, `--public-host` on
  metal), because the refusal has to send the operator to the thing they can edit.
- [x] **Verified by reading the subject alternative names off a minted certificate**, not by
  asserting on the input list. The SAN string is split into typed entries, so `DNS:10.0.0.5` where an
  IP literal was needed cannot pass.

### Half B — per-pairing selection

- [x] **`actana pair new` gains an option choosing which configured host that pairing code's redeem
  response hands back as the endpoint** — `--public-host <addr>`.
- [x] **The chosen host must be a member of the configured list**, validated at `pair new` time
  against `material.serverHosts` (the record of what was actually signed, rather than a config that
  could have drifted from it).
- [x] **An address not in the list is refused, and the refusal prints the available list** — plus the
  primary, named as what omitting the flag would use. Nothing is written on that path: no code
  printed, no session stored.
- [x] **A pairing code can NEVER introduce a host the certificate has no SAN for.** Enforced twice:
  the CLI refuses to mint one, and `buildPairingEndpointResolver` falls back to the primary for a
  stored host that is no longer configured — because an operator can shorten `ACTANA_PUBLIC_HOST`
  while a code minted against the longer list is still live. ADR 0038 D5 and D6.
- [x] **When the option is omitted, the primary is used** — today's behaviour is the default. The
  session stores `null` rather than the primary spelled in, so an unchosen endpoint resolves against
  whatever the Core is configured with at redemption.
- [x] **The redeem endpoint is per-session, resolved from the redeemed pairing session.**
  `CorePairingRoutesOptions.endpoint: string` became `endpointFor(session)`.
- [x] **`pairing-session.ts` gains one optional field to carry it** — `endpointHost?: string | null`.
  Optional, so a session written before it existed is still a session and still means the primary.
- [x] **The Host header stays untrusted.** The value handed back comes from the stored pairing
  session and never from anything in the request; the existing comment in `core-pairing-routes.ts`
  still holds and is expanded to say why it survives this change. Asserted over the wire with a
  `Host` header that lies (`core-pairing-redeem.test.ts` "ignores the Host header, whatever it
  claims").
- [x] **The refusal path in `pair new` has its own tests** — the refusal and its printed list, near
  misses (`10.0.0.50`, `core.example`, `CORE`, `localhost`), an empty value, material recording no
  hosts, and that nothing is written on any of them.

### The ADR

- [x] **A short ADR amending ADR 0016 D18, covering both halves, the members-of-the-list constraint
  and why the Host header is still not a source.** [ADR
  0038](https://github.com/actana/control/blob/feat/multi-san-public-host/docs/adr/0038-a-core-has-several-addresses.md).
  It amends **D15 as well as D18** — D15 is the clause that says "the address" — with dated notes
  appended under both, neither renumbered. D5 is the members-of-the-list constraint, D7 is the Host
  header. Indexed in `docs/README.md`.

### Boundaries

- [x] **No launchd, service, update or status file edited** (#348 owns those). `actana-service.ts`,
  `actana-launchd.ts`, `actana-systemd.ts`, `actana-update.ts`, `actana-status.ts` and their suites
  are untouched. `actana-config.ts`'s `publicHost` keeps its type and meaning precisely so `status`,
  `token` and `endpointFor` need no change.
- [x] **`install.sh` not touched at all**, its `LINE=` version stamp included (#346 owns it).
- [x] **One test file outside the feature changed, and it is named in the PR body.**
  `scripts/__tests__/panel-image.test.mjs` guarded the Core image page against the old systemd
  fixture's hardcoded `--public-host core`; narrowed to `setup --public-host core`, because
  `actana pair new --public-host core` is a different command and the reference compose really does
  call the service `core`.

### Process

- [x] Branch `feat/multi-san-public-host`, cut from the latest `origin/feat/0.4.2-install-and-identity`
  (`39b77ba`), pushed with `-u`.
- [x] Draft PR, base `feat/0.4.2-install-and-identity` — not `main`, not `beta/0.4.2`. Left as a draft.
- [x] Conventional Commits on all eight commits, each one run through this repo's own
  `commitlint.config.mjs` with `@commitlint/cli` — all eight pass. Every body line measured with
  `awk`: longest is 105, under the 132 limit.
- [x] Nothing merged; `main` and `beta/0.4.2` untouched.

mrfqcentic (changes requested):
Reviewed against #347 and its triage comment, on the owner's instruction that **both halves ship** — the certificate half alone makes several addresses valid but never decides which one a given client is told to use, which is the case the ticket opens with. The per-pairing half is reviewed on its merits below, not as out of scope.

**All eight acceptance criteria are met.** Each was checked by reading the enforcing line, not the PR description. Nothing below is a blocker against them, and there is no security regression. I am requesting changes for two small fixes, one of which contradicts this ticket's own headline promise in the daemon log.

---

## Acceptance criteria

**1. A pairing code can never hand back a host outside the configured list — MET.**
Validated at `actana pair new` time in `packages/cli/src/actana-pair.ts:235`, against `material.serverHosts` — the record of what was actually signed, not a config that could have drifted. The refusal at `packages/cli/src/actana-pair.ts:324-338` prints the available list and names the primary, and it returns before `openStore` at `:238`, so nothing is written: no code printed, no session stored. Enforced a second time on the way out in `packages/core/src/core-pairing-wiring.ts:107-111`, which falls back to the primary for a stored host no longer configured.

I checked for a second path and found none. `endpointHost` has exactly one writer in the tree (`actana-pair.ts:258`) and one reader (`core-pairing-wiring.ts:108`); `createPairingSession` has one non-test caller; no config file, env var, API field or default reaches it. A hand-edited `pairing.json` cannot escape either, because the resolver re-checks membership. Confirmed by grep across `packages/` and `scripts/`.

**2. The `Host` header is still not a source — MET.**
`packages/core/src/core-pairing-routes.ts:425` answers with `opts.endpointFor(consumed.session)`. The resolver's whole input is one `PairingSession`; no header, body field or socket address is in scope. The existing reasoning holds and is extended rather than weakened at `:135-158`. `core-pairing-redeem.test.ts:679-703` dials over real TLS with a `Host` header that lies and asserts the answer ignores it.

**3. Backward compatibility, proven by test — MET.** Four independent tests, all against artefacts rather than assertions on inputs:
- `core-cert-material.test.ts:99-116` — a single host mints the exact SAN list the single-host builder produced.
- `core-first-run.test.ts:273-296` — the boot after an upgrade re-issues nothing, asserted against the persisted bytes (`readFileSync` before and after are identical).
- `core-first-run.test.ts:301-320` — material carrying the pre-#347 single `serverHost` boots `unchanged`.
- `actana-machine-cli.test.ts:602-612` — the same unit file and the same endpoint for a single `--public-host`.

`deploy/docker-compose.yml` keeps `ACTANA_PUBLIC_HOST=core`; only comments were added. No compose file needs editing.

**4. Multi-SAN verified off a minted certificate — MET.**
`core-cert-material.test.ts:76-77` parses `X509Certificate.subjectAltName` into typed entries and asserts exact ordered equality, so `DNS:10.0.0.5` where an IP literal was needed fails the test rather than passing a substring check. The same helper is used in `core-first-run.test.ts:212-216` and the SANs are read off the real certificate in `actana-machine-cli.test.ts:585-590`.

**5. Omitting the flag yields the primary — MET.**
`chooseEndpointHost` returns `undefined`, the session records `null` (`pairing-session.ts:149`), and the resolver returns the primary. Asserted at three levels: `actana-pair.test.ts:289-298`, `core-pairing-wiring.test.ts:106-117`, `core-pairing-redeem.test.ts:661-672`.

**6. `localhost` and `127.0.0.1` still appended — MET.**
`core-cert-material.ts:226-238` appends both to every server certificate, on the mint path and the re-issue path. ADR 0032 D9 is cited at the code. Asserted in all three of the cert tests above, including after a re-issue.

**7. Boundaries — MET.**
`install.sh` is byte-identical to the base branch. No launchd, service, update or status file is touched, and no diff line anywhere mentions `launchd`, `systemd`, `.plist` or `actana-core.service`. #346 and #348 are unencroached. The one out-of-feature test edit (`scripts/__tests__/panel-image.test.mjs:1045`) narrows a guard from `--public-host core` to `setup --public-host core` and is correctly justified — `pair new --public-host core` is a different command.

**8. The ADR — MET.**
`docs/adr/0038-a-core-has-several-addresses.md` covers both halves. D5 is the members-of-the-list constraint; D7 records why the `Host` header is still not a source. Dated amendment notes are appended under ADR 0016 D15 and D18 with neither renumbered, per the repo's convention.

---

## Requested changes

### 1. `packages/core/src/core-entry.ts:534-539` — the daemon tells operators to re-pair on the one flow this ticket exists to make painless

Widening the list is classified as a move: `checkServerCertHost` compares in order, so `core` → `core,192.168.1.20` is `moved` (`core-material-store.ts:147-155`), which is correct and deliberate — ADR 0038 §C says so. But the message that fires on `moved` was written when a move could only mean the address genuinely changed:

```
public host is now ${publicHost} — re-issued this Core's server certificate from its
existing CA. Pairing credentials are unchanged; update this Core's address in your
Panel, or run `actana pair new` here and pair it again.
```

For a widening, both instructions are wrong. The primary did not change, the Panel's pinned `core` is still covered, and nothing needs re-pairing — that is the entire promise of this ticket, and `deploy/README.md:130-132` states it to the operator two files away. It also interpolates `publicHost`, the primary alone, so an operator who just added `192.168.1.20` reads "public host is now core" and is told to re-pair over a change the log does not even name.

This line is untouched by the diff, but the diff authored the new condition that reaches it. Suggest distinguishing a widened list (every previously recorded host still covered → say what was added, advise nothing) from a genuine move, and printing the full list either way.

### 2. `packages/shared/src/pairing-store.ts:411` — a new optional field can fail the whole store closed

The `endpointHost` type check was added to `isPairingSession`, which `parse(strict: true)` applies to session rows. `readStrict` throws on any unrecognised row, and `core-pairing-revocation.ts:141-153` treats an unreadable store as *everything is revoked* — the Core then refuses every pairing-issued credential it ever handed out.

So one session row with a non-string, non-null `endpointHost` locks out every paired client of that Core. That is a steep price for a field that is not load-bearing for session validity, and the resolver already degrades a bad value safely on its own: `session.endpointHost ?? ""` on a number yields no usable `.length`, membership fails, and `buildPairingEndpointResolver` returns the primary.

To be fair on reachability: the only writer is `actana pair new`, writes are atomic via tmp-plus-rename, and `pairing.json` is `0600` in the Core's own config dir — so this needs a hand-edit or a future build writing a richer value, and it is not reachable over the wire. It is worth fixing anyway because the failure mode is total and the field earns none of it. Suggest treating a wrong-typed `endpointHost` as absent rather than as a row this build does not know.

---

## Minor — not blocking

- **`packages/cli/src/actana-pair.ts:330-331`** — the refusal's stated reason is false for `localhost` and `127.0.0.1`, which `serverSanAltNames` puts on *every* certificate. `--public-host 127.0.0.1` is refused and told the certificate does not cover it, when it does. The policy is deliberate (ADR 0038 D4 keeps the loopback pair unselectable); only the reason given is wrong.
- **`packages/shared/src/public-hosts.ts:115-117`** — `formatPublicHosts` joins with `", "`, and its doc claims it renders the value handed to `AC_CORE_PUBLIC_HOST` in a unit file. No caller does that; both `actana-setup.ts:428` and `actana-cli.ts:1404` use `.join(",")` inline. Its real output is operator-facing, where the space makes `Configured (ACTANA_PUBLIC_HOST): core, 10.0.0.5` unpasteable back into an unquoted `--public-host`.
- **`packages/core/src/core-entry.ts:434-437`** — the parse refusal names `ACTANA_PUBLIC_HOST` whenever `AC_CORE_PUBLIC_HOST` is set, including on metal where the unit sets `AC_CORE_PUBLIC_HOST` and `ACTANA_PUBLIC_HOST` does not exist. Narrow, but this branch exists precisely for anything execing the daemon bundle directly.
- **`packages/cli/src/actana-config.ts:118-137`** — inserting `configPublicHosts` orphaned `endpointFor`'s JSDoc: `configPublicHosts` now carries two stacked doc blocks and `endpointFor` has none.
- **ADR 0034** gets no reciprocal dated note, though ADR 0038's header and `docs/README.md:96` both say it adds to the surface 0034 D2 bounds. ADR 0016 got notes under both amended clauses; 0034 got none.

Two pre-existing issues the diff moves but does not introduce, noted so they are not re-reviewed: `buildPairingEndpointResolver` does not bracket an IPv6 host where `actana-config.ts:endpointFor` does, and `isIp` is IPv4-only so an IPv6 entry lands as a `DNS:` SAN; and a flagless `actana setup` re-run still takes `choosePublicHost()`'s guess over the recorded config, which now silently collapses a multi-address install the way it already discarded `--port` and `--label`.

---

CI is green on `eabc6f1` — Unit Tests, Typecheck, Lint, both installer E2E legs and both image smokes. The PR is still a draft; I have not touched its state.

qcentic-dev:
Thank you — the review found a real contradiction between the daemon's log and this ticket's own
headline promise, and a total failure mode bought by a field that earns none of it. **All seven
items are fixed**; the three pre-existing ones are filed as
[#351](https://github.com/actana/control/issues/351) rather than fixed here.

Pushed as six commits on top of `eabc6f1`, ending at `24cd65a`.

---

## Requested changes

### R1 — the daemon told operators to re-pair on the flow this ticket exists to make painless

`29a8337`. You were right that the diff authored the condition reaching that line, so it is this
PR's to fix.

`widenedPublicHosts` in `public-hosts.ts` is the predicate: **every previously recorded host still
covered, and at least one added**. It returns `null` rather than an empty list for the three cases
that are *not* a widening, so no caller can read "nothing added" as "this was safe":

- nothing recorded before — there is no previous coverage to have preserved;
- a previously covered host is gone — the move a client actually feels;
- the same set reordered — nothing added, and the primary, and therefore the default endpoint, is
  not what it was.

`loadOrMintMaterial` reports `certAction: "widened"` and returns `addedHosts`. The comparison is
taken **before** `reissueServerCert` overwrites `serverHosts` — that is the only moment both lists
exist, which is also why `addedHosts` is returned rather than recomputed by the caller.

The daemon now has two branches. The widening one names what was added and **advises nothing**:

```
public hosts are now core,192.168.1.20 — added 192.168.1.20, and re-issued this Core's
server certificate from its existing CA to cover it. Every address this Core already
answered to is still covered, so every paired client keeps working and none has to be
re-paired. Pair a client to a new address with `actana pair new --public-host <addr>`.
```

The move one keeps its advice and gains the sentence that earns it — *"This Core no longer answers
to every address it was signed for."* **Both print the full list**, which was your other point:
naming the primary alone is how an operator who added `192.168.1.20` read "public host is now core".

Recorded as ADR 0038 **D3a**, and covered by five tests in `core-first-run.test.ts` plus four on the
predicate — including that a reorder is a move and that a widening really was re-signed.

### R2 — an optional field could fail the whole store closed

`c41ca86`. Agreed, and agreed on the reasoning: the field decides no part of session validity and
the resolver already degrades a bad value safely, so it should never have been able to buy a
condemned row.

The check is out of `isPairingSession` entirely. `parse` now maps sessions through
`normaliseSession`, which reduces anything that is not a usable string to `null` — what every
session written before the field existed already means. The predicate carries a comment explaining
why it is *not* checked there, so the next person to add an optional field has the reasoning in
front of them.

**The leniency is scoped to that one field**: a session with a wrong-typed `codeHash` is still
condemned, and there is a test holding that line. Nine tests in `pairing-store.test.ts` cover the
rest — `7`, `true`, `{}`, `[]`, `{host}`, empty and whitespace-only, all loading as `null` with the
paired clients still readable.

One thing the tests caught that I would have shipped otherwise: my first normalisation
short-circuited on `host === raw`, which left `""` intact because an empty string equals its own
trim. Fixed to require `length > 0` as well.

---

## Minor

- **m1** `6b0c0f8` — right, and the false sentence was the whole of it. The general reason now says
  the address has to be one this Core was *configured to answer on* (so the certificate is
  guaranteed to cover it), and `localhost`/`127.0.0.1` get their own true reason: handed to a client
  on another machine, a loopback address names that machine rather than this Core. The policy is
  untouched — `LOOPBACK_HOSTS` is named for the refusal text only and is explicitly *not* an
  exception list the membership check consults. Two tests, including one asserting an ordinary
  refusal gets no loopback sentence.
- **m2** `7ec7e14` — dropped the space and corrected the doc to say what it actually is
  (operator-facing) rather than what it is not (the unit file value, which both call sites join
  inline). Added a round-trip assertion through `parsePublicHosts`, since "a list this printed is a
  list the flag accepts" is the property that matters.
- **m3** `29a8337` — the refusal now names `AC_CORE_PUBLIC_HOST` on metal and `ACTANA_PUBLIC_HOST`
  in a container, keyed on `containerMode` rather than on which variable happens to be set. Naming a
  variable the operator cannot find is worse than naming none, and this branch exists precisely for
  what execs the bundle directly.
- **m4** `4f83ed9` — each function has its own block again. `endpointFor`'s now also says why it is
  one address and only ever one.
- **m5** `24cd65a` — ADR 0034 D2 has its dated note. It records what did *not* change, since that is
  the load-bearing part: still exactly one pre-auth route, same handshake confinement, request shape
  untouched, and the *"the pre-auth hole is exactly one route wide"* case still holding and not
  edited. What moved is inside the 200 body, and the pre-auth caller gained no influence over it.

---

## Deferred to #351 — the three pre-existing items

Filed as **[#351](https://github.com/actana/control/issues/351)**, which states plainly that all
three predate this PR and were surfaced by this review. Not fixed here — each is a behaviour change
this ticket does not own:

1. **`buildPairingEndpointResolver` does not bracket an IPv6 host** where `endpointFor` does, and
   `isIp` is IPv4-only so an IPv6 entry lands as a `DNS:` SAN. Both halves want fixing together: an
   operator on IPv6 gets a certificate that does not verify *and* an endpoint that does not parse.
2. **A flagless `actana setup` re-run prefers `choosePublicHost()`'s guess over the recorded
   config**, which now collapses a multi-address install the way it already discarded `--port` and
   `--label`. This PR changed the blast radius, not the behaviour. #351 notes it should be checked
   against #348 for overlap.
3. **The remote non-container endpoint falls back to the bind address** — `AC_CORE_PUBLIC_HOST ||
   host` — so a metal or direct-exec Core with no public host set mints a certificate for `0.0.0.0`
   and hands out `wss://0.0.0.0:<port>`. `publicHostDeclared` already bounds the re-issue path; the
   gap is the first mint. Whether metal should refuse the way container mode does is a decision
   worth recording before code.

---

Full suite green locally — typecheck clean, all six stages pass, and `eslint` silent on every file
touched. The PR is still a draft, base unchanged, nothing merged.

qcentic-dev:
**CI on `24cd65a`: green — 18 pass, 6 skipping (the train-only jobs), 0 failing.**

Flagging one thing so the red-then-green history is not confusing later. Unit Tests failed on the
first run with `in-process-core-session.test.ts` → *"sends and awaits the turn in one PTY
resolution"* timing out at 60005ms. That is a flake, and I re-ran the job rather than asserting it:

- it is a **timeout, not an assertion** — a live PTY turn did not settle, which is not how a wrong
  certificate or host list fails;
- it passes locally in ~3.5s against a 60s budget, four runs in a row;
- **this branch has never touched that file** — `git log` over it from the base is empty;
- the only changed line reaching it is r1's behaviour-identical fixture rename
  (`{ host: "127.0.0.1" }` → `{ hosts: ["127.0.0.1"] }` — one host in, the same certificate out),
  and that line was already in `eabc6f1`, where Unit Tests were green.

The re-run passed on the same commit. Nothing in these six commits goes near sessions or PTYs.

mrfqcentic (approved):
Re-reviewed at `24cd65a` (six commits since `eabc6f1`). **R1 and R2 are genuinely fixed, all five minor items are addressed, nothing regressed against the eight criteria, and I accept the Unit Tests failure as a flake.** Approving.

Each fix was verified by reading the enforcing line, not the reply.

---

## R1 — a widened list is not a move

**Fixed, and the predicate is the right shape.** `packages/shared/src/public-hosts.ts:170-177` adds `widenedPublicHosts(previous, next)`, returning the additions or `null`. It returns `null` for all three non-widenings — nothing recorded before, a previously covered host gone, and the same set reordered — which is what stops a caller mistaking "nothing was added" for "this was a widening".

**The comparison is taken at the only moment both lists exist.** `core-first-run.ts:154` computes `added` *before* `reissueServerCert` overwrites `serverHosts` at `:156`, and `addedHosts` is returned on the result rather than recomputed by the caller. That is the correct place for it.

**The motivating case fires.** For the case the ticket opens with — an existing single host, a second address added — `previous = ["core"]`, `next = ["core","192.168.1.20"]`: previous is non-empty, every previous host is in next, one host joined. `certAction: "widened"`. Confirmed against the test at `core-first-run.test.ts:236-266`, which was the one asserting `"moved"` in r1 and now asserts `widened` with `addedHosts: ["10.0.0.5"]`, both SANs present, and the CA, `coreId` and `coreUuid` preserved.

**The old advice cannot reach it.** `core-entry.ts:546` is `if (certAction === "widened")` with the move branch as an `else if` at `:560`. The four `certAction` values are mutually exclusive and every branch of `loadOrMintMaterial` returns early, so a widening cannot fall through to the re-pair line. The widening message names what was added and advises nothing; the move message keeps its advice and gains the sentence that earns it. **Both print the full list** via `formatPublicHosts(publicHosts)` — the naming-the-primary-alone bug is gone from both branches.

Recorded as ADR 0038 **D3a**, inserted without renumbering D4–D7. Nine tests: five on the boot path (widening, addition-not-last, dropped host, reorder, and that `backfilled`/`unchanged` carry no additions) and four on the predicate.

## R2 — a malformed `endpointHost` never fails the store closed

Verified against all three things I asked for, and the fix did **not** make the store permissive.

1. **Out of the strict predicate.** `pairing-store.ts:399-425` — `isPairingSession` now ends at `consumedAt`. The `endpointHost` clause is gone, replaced by a comment stating why it must not be checked there (`readStrict` → `core-pairing-revocation.ts` → everything-revoked). Keeping the reasoning at the predicate is the right call; the next person to add an optional field will read it.
2. **Normalisation on the way out of parse.** `pairing-store.ts:213` — `this.rows(...).map(normaliseSession)`. It sits outside the strict/lenient branch, so both moods normalise identically, which preserves the "one reader, two moods" property the file documents at `:183-189`.
3. **Load-bearing fields still condemned.** `id`, `label`, `codeHash`, `createdAt`, `expiresAt`, `attempts`, `attemptCap` and `consumedAt` are all still required, and `pairing-store.test.ts:411-419` holds that line explicitly: a session with `codeHash: 7` still throws `session 0 is not a session this build knows`. The leniency is scoped to exactly one optional field.

I traced `normaliseSession` (`:441-451`) through every input: `undefined` and `null` pass through untouched; a clean string short-circuits; a padded string is trimmed; `""`, `"   "`, `7`, `true`, `{}`, `[]` and `{host}` all become `null`. The `host.length > 0 && host === raw` guard is correct — the author is right that `host === raw` alone would have let `""` through, since an empty string equals its own trim, and the tests caught it.

## Minor

- **m1** — `actana-pair.ts:343-345` — the general reason now says the address has to be one this Core was *configured to answer on, so that the certificate is guaranteed to cover it*, which is true. `:355-361` adds a separate, true sentence for `localhost`/`127.0.0.1`: reachability, not coverage. **The policy is untouched** — `LOOPBACK_HOSTS` (`:284-294`) is used only inside the already-failed `!isConfiguredPublicHost` branch to add a sentence, and the membership check at `:337` does not consult it. The loopback refusal still returns `REFUSED` and still writes nothing, asserted at `actana-pair.test.ts`. A second test asserts an ordinary refusal gets no loopback sentence.
- **m2** — `public-hosts.ts:123` — joins with `","`, no space. The doc now says what the function is (operator-facing) rather than what it is not, and correctly notes the two `AC_CORE_PUBLIC_HOST` call sites join inline. The round-trip assertion through `parsePublicHosts` is the right test: "a list this printed is a list the flag accepts" is the property that matters. Both downstream expectations updated consistently.
- **m3** — `core-entry.ts:443` — `containerMode ? CONTAINER_PUBLIC_HOST_ENV : "AC_CORE_PUBLIC_HOST"`, keyed on the mode rather than on which variable happens to be set. Correct: `containerMode` is established at `:147`, well before use.
- **m4** — `actana-config.ts` — each function has its own block again, and `endpointFor`'s now also records why it is one address and only ever one.
- **m5** — ADR 0034 D2 carries its dated note, under the right clause, not renumbered. Recording what did *not* change is the right emphasis for a clause that bounds a pre-auth surface.

## No regression against the eight criteria

Re-verified at `24cd65a` rather than assumed:

- **Criterion 1** — `endpointHost` still has exactly one writer (`actana-pair.ts:258`) and one reader (`core-pairing-wiring.ts:108`); `normaliseSession` only narrows values toward `null`. Both enforcements intact: `actana-pair.ts:337` against `material.serverHosts`, `core-pairing-wiring.ts:110` on the way out. No new path.
- **Criterion 2** — `core-pairing-routes.ts` and `core-pairing-wiring.ts` are byte-identical to `eabc6f1`. `endpoint: opts.endpointFor(consumed.session)` unchanged.
- **Criteria 4 and 6** — `core-cert-material.ts` and `core-material-store.ts` byte-identical to `eabc6f1`. The SAN builder and the loopback pair are untouched.
- **Criterion 3** — `deploy/docker-compose.yml` untouched; the single-`--public-host` compatibility test is untouched. The only test edit here is the `Also valid` string following m2's join change.
- **Criterion 7** — re-run against the base branch: `install.sh` byte-identical, no boundary filename in the changed set, and a content scan of every added and removed line for `launchd`, `LaunchAgent`, `systemd`, `.plist`, `actana-core.service` returns nothing. #346 and #348 unencroached.

## The Unit Tests failure — accepted as a flake

I checked this rather than taking it, and the author's account is accurate in every particular.

Run `33177259041` attempt 1 failed; attempt 2 on the **same commit** succeeded. From the attempt-1 log: `in-process-core-session.test.ts` > *"sends and awaits the turn in one PTY resolution, and prints the settled screen"* — `Error: Test timed out in 60000ms.` at 60005ms, `1 failed | 1062 passed`.

Why it is not attributable to this PR:

- It is a **timeout, not an assertion**. A wrong host list or certificate fails as an assertion or a TLS error; it does not make a live PTY turn fail to settle.
- `git log` over `in-process-core-session.test.ts` from the base branch is **empty** — this branch has never touched it.
- The only changed line reaching that suite is the fixture rename in `in-process-core.ts:261` (`{ host: "127.0.0.1" }` → `{ hosts: ["127.0.0.1"] }`), which is behaviour-identical — one host in, the same certificate out, as `core-cert-material.test.ts` asserts entry by entry.
- That line was already in `eabc6f1`, where Unit Tests were green. If the rename were the cause, the earlier run would have failed too.
- In the failing attempt itself, every suite exercising this PR passed: `shared` 500/500, `sdk` 292/292, and the other 1062 `cli` tests.
- None of the six new commits goes near sessions, PTYs or the core-link transport.

One thing worth saying plainly: a 60s PTY timeout could still be a real intermittent bug in that suite that this PR merely happens not to have caused. That is a repo-health question rather than a merge blocker here — if it recurs it deserves an issue of its own.

CI on `24cd65a` is 18 pass, 6 skipping, 0 failing.

## #351

Filed, open, and it covers all three items with the reasoning intact: the IPv6 bracketing mismatch together with the IPv4-only `isIp` (correctly kept as one item, since an operator on IPv6 gets both a certificate that does not verify and an endpoint that does not parse), the flagless `setup` re-run preferring its guess over the recorded config, and the remote non-container fallback to the bind address. It states plainly that all three predate this PR, and it notes the #348 overlap check for item 2. Correctly not fixed here.

---

## Two observations, deliberately not requested changes

Both are narrower than what they replaced and neither touches a criterion. Noting them so they are on record rather than lost.

1. **`core-entry.ts:563`** — the move branch's sentence *"This Core no longer answers to every address it was signed for"* is false for a pure reorder, which reaches `moved` by design. When the primary changed the *advice* is still right, so the cost is one inaccurate sentence in a rare, deliberate operation. Not worth another round.
2. **`pairing-store.ts:420`** — the comment refers to `readEndpointHost`; the function is called `normaliseSession`. A stale name, nothing more.

Also worth recording: `widenedPublicHosts(["core"], ["10.0.0.5","core"])` is a widening even though the primary changed, so a new code's default endpoint moves while the operator is advised nothing. Nothing breaks — every existing client keeps its coverage, and the message prints the full list with the new primary first — and the test pins it deliberately, so I read it as intended rather than as an oversight. It does sit slightly at odds with D3a's stated rationale for treating a reorder as a move.

---

**Approving.** All eight acceptance criteria remain met, both requested changes are fixed at the level asked for — R2 in particular scoped its leniency to the one field and proved it — and the five minor items are done. The PR is still a draft; I have not touched its state.

Issue #90, opened 2026-08-05
macOS on-device Core: mac-arm64 tarball behind a manually-approved release leg


qcentic-adm:
Part of the docs/install/release overhaul (charted 2026-08-05 with the maintainer). **Supersedes and reverses #50** — see the closing comment there. **Amends ADR 0016 D28** (three assets, no macOS legs): amend the ADR in this PR, per the standing rule.

## What

macOS gets a real on-device Core install: `curl … install.sh | sh` works on Apple Silicon, installing the CLI and the core daemon (LaunchAgent). The decisions, locked during charting:

- **mac-arm64 only.** Intel Macs use the Docker path. One new release asset: `actana-core-<version>-mac-arm64.tar.gz` (total 3 tarballs + `SHA256SUMS`).
- **Zero macOS in per-PR CI.** The cost posture from decision #14 stands.
- **The mac tarball is built by a release-only leg gated behind a GitHub environment requiring manual approval.** The assigned developer runs `docs/core-macos-prerelease-checklist.md` on real hardware, then approves; only then does the release publish. No weekly canary for now (explicitly deferred).
- Panel stays Docker-only; this issue is Core-only.

The CLI-side plumbing already exists and is unit-tested: `actana-launchd.ts` (LaunchAgent, label `com.actana.core`), `actana-layout.ts:76-83` (darwin service paths), `actana-release.ts:107-111` (`darwin → "mac"` asset naming), `shell-env.ts` darwin probing. What's missing is the distribution: the tarball, the installer's Darwin branch, and the release leg.

## Done when

- [ ] `scripts/lib/core-tarball.mjs` `CORE_TARGETS` regains one darwin row → target `mac-arm64` (its header comment lists every co-edit a new row requires — follow it)
- [ ] `release.yml` gains a `tarball (mac-arm64)` job on a macOS arm64 runner, `environment: macos-release`; `fail-fast: false` stays; `github-release` needs it
- [ ] The `macos-release` environment (required reviewers) is documented in `docs/REPO_SETUP.md` as an admin setup step, and `docs/ci-cd.md` §"Cutting a release" describes the approval pause
- [ ] `scripts/compose-core-shasums.mjs` guard: `--expect 2` → `--expect 3`, changed deliberately with a comment (the guard is what makes a silently missing arch a red build)
[...]

PR #352, opened 2026-08-28
fix: migrate the pre-rename LaunchAgent, and refuse the environment it leaves behind

### LINKED ISSUE #348: install.sh/actana setup leaves a stale, incompatible LaunchAgent + identity material after an in-place upgrade

qcentic-dev:
Closes #348.

A Mac upgraded in place from before the Harness → Core rename keeps `com.actana.harness`. Its `ProgramArguments` point at
`current/bin/actana`, so an upgrade hands it the *new* binary, and `KeepAlive` brings it back every time it loses the port.
Nothing detected it, nothing removed it, and `actana status` reported a service that was not there while the one that was
ran under another name.

The triage comment on #348 confirmed all three sub-claims REAL against `main`. All three are fixed here, plus the material
validation it recommended.

### 1. The launchd `removeLegacyUnit` was a hard-coded `return null`

Its justification cited ADR 0016 D28 — macOS Core targets dropped rather than migrated — which #90 amended by giving macOS
a first-class on-device Core. The implementation is now the mirror of the systemd one that already worked:
`LEGACY_LAUNCH_AGENT_LABEL`, detect the plist **or** ask `launchctl print`, `bootout`, delete the plist, return the label.
Detection is two-sided on purpose: `rm`ing a plist unloads nothing, because launchd holds the copy it read at bootstrap.

`actana update` now calls it **before** the restart — it never did, which is exactly why an old agent picks up the new
binary after an upgrade. `actana place` (what `install.sh` runs) warns when the legacy plist is there, since an operator
who only runs `install.sh` never reaches `setup`. The test that pinned the no-op is flipped.

### 2. Security: an unauthenticated plaintext core-link on 0.0.0.0

The pre-rename plist sets `AC_HARNESS_REMOTE`, `AC_HARNESS_PUBLIC_HOST` and `AC_HARNESS_MATERIAL_FILE`; only
`AC_CORE_LINK_PORT`, `AC_CORE_LINK_HOST` and `AC_USER_DATA_DIR` kept their names. The daemon read only the `AC_CORE_`
spellings, with no fallback and no warning — so remote mode, the public host and the material file silently vanished while
the port and the bind address survived. What booted was a plaintext, unauthenticated core-link, able to start processes and
read the machine's files, on every interface, and the only symptom was `wrong version number` at a TLS client.

`startCore` now refuses both halves: any `AC_HARNESS_*` variable stops the boot before anything else is read, naming the
rename, the service that set them and `actana setup`; and a non-loopback bind without `AC_CORE_REMOTE` stops it on its own
terms. Refused rather than logged — a line in a log nobody tails is not consent. A fallback was the other available fix and
is the wrong one: it would keep that server running on purpose. Every legitimate non-loopback bind this product writes sets
`AC_CORE_REMOTE=1`, so nothing that works today stops working.

### 3. `actana status` lied about the service

It printed the `com.actana.core` constant as `Auto-start` and read `State` off that one fixed plist filename, so a machine
running the legacy agent read as `not installed` while the config file made the header say `degraded`. It now reports the
label launchd actually loaded, falls back to the legacy plist for `State`, and carries an explicit `Legacy agent` row
naming `actana setup`. A legacy service on the machine is a degradation whatever else is true, so the exit code is non-zero
until it is gone.

### Material validation

`readMaterialFile` type-checked eight strings, so pre-rename material — same filename, same fields, same types, different
CA — was indistinguishable from current material and failed later as an opaque TLS error. `checkMaterialIdentity` asks the
three questions a handshake depends on: is the CA one this product minted, did it sign this server certificate, and does
that certificate go with the key beside it. `mission-control-harness-ca` gets its own sentence, because an operator seeing
it needs to know the file is old rather than corrupt. Signature verification rather than `checkIssued`, which compares
names — and every CA here carries the same name.

### Boundaries

`install.sh` is untouched (#346 owns it, and its `LINE=` version stamp especially). The public host list and pairing design
from #347 are built on, not reopened — `core-entry.ts`, `core-first-run.ts`, `core-material-store.ts`, `actana-setup.ts`
and `actana-cli.ts` keep the signatures that train settled.

### Verification

`pnpm test` — all 6 stages pass (326 files). `pnpm -r typecheck` clean. `pnpm lint` reports no new warnings.
Nothing launchd-related was executed: this was written in a Linux container, so the macOS paths are covered by the fake
`ActanaSystem` the rest of the service suite uses, end to end through `actana setup`, `place`, `status` and `update`.

### DIFF
--- a/packages/cli/src/actana-service.ts
+++ b/packages/cli/src/actana-service.ts
@@ -23,6 +23,8 @@ import {
   chooseLaunchdDomain,
   LAUNCH_AGENT_LABEL,
   launchdLogPath,
+  LEGACY_LAUNCH_AGENT_LABEL,
+  legacyLaunchAgentPath,
   parseLaunchctlPrint,
   renderActanaPlist,
   serviceTarget,
@@ -65,6 +67,28 @@ export type PersistenceRow = {
   /** `Linger` on systemd, `At login` on launchd. */
   label: string;
   value: string;
+};
+
+/**
+ * Which of this machine's services the init system actually has (#348).
+ *
+ * `name` is what is really installed or loaded here, which is not the same
+ * question as `ActanaServiceManager.name` — that one is the label setup
+ * *writes*, and reporting it unconditionally is how `actana status` came to
+ * print `com.actana.core` on a machine whose only agent was the pre-rename
+ * `com.actana.harness`.
+ */
+export type ServiceObservation = {
+  /** The unit / label this machine actually has, or null when it has none. */
+  name: string | null;
+  /**
+   * The pre-rename unit / label, when this machine still carries one.
+   *
+   * Reported separately from `name` because both can be true at once — a
+   * machine upgraded in place has the new plist *and* the old agent still
+   * loaded, and that combination is the failure #348 describes.
+   */
+  legacyName: string | null;
 };
 
 /** What making the service persist achieved. */
@@ -118,6 +142,13 @@ export type ActanaServiceManager = {
    * this exists and when it goes.
    */
   removeLegacyUnit(): string | null;
+  /**
+   * What this machine actually has installed or loaded, for `actana status`.
+   *
+   * Never throws and never asks the operator to act: a status report on a
+   * broken machine is the one thing that still has to work.
+   */
+  observe(): ServiceObservation;
   /** Make the daemon survive logout as far as this platform allows. */
   ensurePersistence(context: PersistenceContext): Promise<PersistenceOutcome>;
   /** Start at boot/login from now on, and start now. Throws when it cannot. */
@@ -206,6 +237,27 @@ function systemdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
     return true;
   }
 
+  /**
+   * Whether a pre-rename unit is still on this machine.
+   *
+   * Two places, because deleting a unit file does not disable the unit:
+   * systemd reads a user's own units from `$XDG_CONFIG_HOME/systemd/user`, and
+   * `enable` leaves a symlink in `default.target.wants` beside it that still
+   * starts the daemon at boot on its own.
+   *
+   * Asked of the filesystem rather than of systemd, so a machine that never had
+   * one runs no commands at all and cannot fail. `lstat` for the link, because
+   * once the unit file is gone the link dangles and `existsSync` follows it to
+   * nothing — which is exactly the case being caught.
+   */
+  function legacyUnitPresent(): boolean {
+    const legacyPath = path.join(layout.serviceDir, LEGACY_UNIT_NAME);
+    const enabledLink = path.join(layout.serviceDir, "default.target.wants", LEGACY_UNIT_NAME);
+    return (
+      fs.existsSync(legacyPath) || Boolean(fs.lstatSync(enabledLink, { throwIfNoEntry: false }))
+    );
+  }
+
   /** Stop a unit, unlink it from boot, delete it, and leave no failed entry. */
   function removeUnit(name: string, filePath: string): void {
     systemctl("stop", name);
@@ -242,21 +294,17 @@ function systemdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
     },
 
     removeLegacyUnit() {
-      // Two places, because deleting a unit file does not disable the unit:
-      // systemd reads a user's own units from `$XDG_CONFIG_HOME/systemd/user`,
-      // and `enable` leaves a symlink in `default.target.wants` beside it that
-      // still starts the daemon at boot on its own.
-      const legacyPath = path.join(layout.serviceDir, LEGACY_UNIT_NAME);
-      const enabledLink = path.join(layout.serviceDir, "default.target.wants", LEGACY_UNIT_NAME);
-      // Asked of the filesystem rather than of systemd, so a machine that never
-      // had one runs no commands at all and cannot fail. `lstat` for the link,
-      // because once the unit file is gone the link dangles and `existsSync`
-      // follows it to nothing — which is exactly the case being caught.
-      if (!fs.existsSync(legacyPath) && !fs.lstatSync(enabledLink, { throwIfNoEntry: false })) {
-        return null;
-      }
-      removeUnit(LEGACY_UNIT_NAME, legacyPath);
+      if (!legacyUnitPresent()) return null;
+      removeUnit(LEGACY_UNIT_NAME, path.join(layout.serviceDir, LEGACY_UNIT_NAME));
       return LEGACY_UNIT_NAME;
+    },
+
+    observe() {
+      const legacyName = legacyUnitPresent() ? LEGACY_UNIT_NAME : null;
+      return {
+        name: fs.existsSync(layout.servicePath) ? UNIT_NAME : legacyName,
+        legacyName,
+      };
     },
 
     async ensurePersistence(context) {
@@ -325,6 +373,25 @@ function launchdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
     return domain;
   };
   const target = (): string => serviceTarget(resolveDomain(), LAUNCH_AGENT_LABEL);
+  const legacyPath = legacyLaunchAgentPath(layout.home);
+  const legacyTarget = (): string => serviceTarget(resolveDomain(), LEGACY_LAUNCH_AGENT_LABEL);
+
+  /**
+   * Whether the pre-rename agent is still on this machine.
+   *
+   * Both halves are needed and neither implies the other. The plist is what
+   * brings the old agent back at the operator's next login, and it is what an
+   * in-place upgrade leaves untouched. A *loaded* job with no plist left is the
+   * other half of the same mess — `rm`ing the file does not unload anything,
+   * because launchd holds the copy it read at bootstrap time.
+   *
+   * The file is checked first so a machine that never had one asks launchd
+   * nothing at all, which is also what keeps this cheap enough for the status
+   * path to call.
+   */
+  const legacyPresent = (): boolean =>
+    fs.existsSync(legacyPath) ||
+    system.run("launchctl", ["print", legacyTarget()]).status === 0;
 
   const printJob = (): CommandResult => system.run("launchctl", ["print", target()]);
   /** Whether launchd knows about the job at all — loaded, running or not. */
@@ -333,6 +400,28 @@ function launchdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
   const isRunning = (): boolean => {
     const printed = printJob();
     return printed.status === 0 && parseLaunchctlPrint(printed.stdout).state === "running";
+  };
+
+  /** What launchd says about one label, in the shape `actana status` renders. */
+  const stateOf = (label: string): ActanaServiceState => {
+    const printed = system.run("launchctl", ["print", serviceTarget(resolveDomain(), label)]);
+    if (printed.status !== 0) {
+      // The agent is installed but not loaded — `actana stop`, or a session
+      // that has not logged in since. That is stopped, not missing.
+      return { loadState: "loaded", activeState: "inactive", subState: "dead", mainPid: null };
+    }
+    const job = parseLaunchctlPrint(printed.stdout);
+    if (job.state === "running") {
+      return { loadState: "loaded", activeState: "active", subState: "running", mainPid: job.pid };
+    }
+    // Loaded but not running: launchd is between restarts, or throttling a job
+    // that keeps dying. Either way the operator has something to fix.
+    return {
+      loadState: "loaded",
+      activeState: "activating",
+      subState: job.state ?? "unknown",
+      mainPid: null,
+    };
   };
 
   /** Register the agent with launchd. `RunAtLoad` is what starts it. */
@@ -387,10 +476,21 @@ function launchdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
     },
 
     removeLegacyUnit() {
-      // Nothing to clean up: the pre-rename `com.actana.harness` agent only
-      // ever existed on a machine that built its own tarball, and the macOS
-      // Core targets are being dropped rather than migrated (ADR 0016 D28).
-      return null;
+      // This used to return null on the reasoning that macOS Core targets were
+      // being dropped rather than migrated (ADR 0016 D28). #90 amended that
+      // decision and gave macOS a first-class on-device Core, so the pre-rename
+      // agent is now something real machines carry — and carry through an
+      // upgrade, because its `ProgramArguments` point at `current/bin/actana`
+      // and `KeepAlive` brings it back (#348).
+      if (!legacyPresent()) return null;
+      // Unload before deleting, the same order `uninstall` uses and for the
+      // same reason: launchd holds the plist it read at bootstrap time, so
+      // removing the file alone leaves the old daemon running on the port the
+      // new one is about to want. Failure is the ordinary "nothing was loaded"
+      // case — the plist below is the half that matters.
+      system.run("launchctl", ["bootout", legacyTarget()]);
+      fs.rmSync(legacyPath, { force: true });
+      return LEGACY_LAUNCH_AGENT_LABEL;
     },
 
     async ensurePersistence() {
@@ -420,30 +520,20 @@ function launchdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
     },
 
     state() {
-      if (!fs.existsSync(layout.servicePath)) return null;
-      const printed = printJob();
-      if (printed.status !== 0) {
-        // The agent is installed but not loaded — `actana stop`, or a session
-        // that has not logged in since. That is stopped, not missing.
-        return { loadState: "loaded", activeState: "inactive", subState: "dead", mainPid: null };
-      }
-      const job = parseLaunchctlPrint(printed.stdout);
-      if (job.state === "running") {
-        return {
-          loadState: "loaded",
-          activeState: "active",
-          subState: "running",
-          mainPid: job.pid,
-        };
-      }
-      // Loaded but not running: launchd is between restarts, or throttling a
-      // job that keeps dying. Either way the operator has something to fix.
-      return {
-        loadState: "loaded",
-        activeState: "activating",
-        subState: job.state ?? "unknown",
-        mainPid: null,
-      };
+      if (fs.existsSync(layout.servicePath)) return stateOf(LAUNCH_AGENT_LABEL);
+      // No plist under the current label, but launchd may still be running the
+      // pre-rename agent out of this same tree — which is a Core that is up,
+      // not the "not installed" this used to report (#348).
+      return legacyPresent() ? stateOf(LEGACY_LAUNCH_AGENT_LABEL) : null;
+    },
+
+    observe() {
+      const legacyName = legacyPresent() ? LEGACY_LAUNCH_AGENT_LABEL : null;
+      // The plist is the stronger signal and is checked first: it is what
+      // launchd loads at the next login, so an installed-but-unloaded agent is
+      // still this machine's service.
+      const current = fs.existsSync(layout.servicePath) || isLoaded();
+      return { name: current ? LAUNCH_AGENT_LABEL : legacyName, legacyName };
     },
 
     verb(verb) {
--- a/packages/cli/src/actana-service.ts
+++ b/packages/cli/src/actana-service.ts
@@ -258,6 +258,23 @@ function systemdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
     );
   }
 
+  /** What systemd says about one unit, or null when it has no such unit. */
+  function stateOf(name: string): ActanaServiceState | null {
+    const show = systemctl(
+      "show",
+      name,
+      "--property=LoadState",
+      "--property=ActiveState",
+      "--property=SubState",
+      "--property=MainPID",
+    );
+    if (show.status !== 0) return null;
+    const state = systemdStateFromShow(show.stdout);
+    // `not-found` means systemd has no such unit — reporting "no service" is
+    // truer than reporting a load state nobody can act on.
+    return state.loadState === "not-found" ? null : state;
+  }
+
   /** Stop a unit, unlink it from boot, delete it, and leave no failed entry. */
   function removeUnit(name: string, filePath: string): void {
     systemctl("stop", name);
@@ -301,10 +318,19 @@ function systemdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
 
     observe() {
       const legacyName = legacyUnitPresent() ? LEGACY_UNIT_NAME : null;
-      return {
-        name: fs.existsSync(layout.servicePath) ? UNIT_NAME : legacyName,
-        legacyName,
-      };
+      if (fs.existsSync(layout.servicePath)) return { name: UNIT_NAME, legacyName };
+      // No unit file, but systemd may still be holding the unit — one whose
+      // file was deleted by hand is still loaded and still running a daemon,
+      // which is not "absent" (the launchd side asks the same pair).
+      //
+      // Asked only when there is a legacy unit to be confused with, and that
+      // is deliberate: `actana place` calls this, and placement asks the init
+      // system nothing on the ordinary machine — a copy and two symlinks, as
+      // its own test pins. A machine that *has* a pre-rename unit is getting a
+      // warning printed about it either way, and one `systemctl show` is what
+      // makes that warning name the right service.
+      if (legacyName && stateOf(UNIT_NAME)) return { name: UNIT_NAME, legacyName };
+      return { name: legacyName, legacyName };
     },
 
     async ensurePersistence(context) {
@@ -320,19 +346,13 @@ function systemdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
     },
 
     state() {
-      const show = systemctl(
-        "show",
-        UNIT_NAME,
-        "--property=LoadState",
-        "--property=ActiveState",
-        "--property=SubState",
-        "--property=MainPID",
-      );
-      if (show.status !== 0) return null;
-      const state = systemdStateFromShow(show.stdout);
-      // `not-found` means systemd has no such unit — reporting "no service" is
-      // truer than reporting a load state nobody can act on.
-      return state.loadState === "not-found" ? null : state;
+      const current = stateOf(UNIT_NAME);
+      if (current) return current;
+      // No current unit, but this machine may still be running the pre-rename
+      // one — the same fallback the launchd side has, and without it `status`
+      // prints `Auto-start actana-harness.service` over `State not installed`,
+      // which is #348's own self-contradiction on the other platform.
+      return legacyUnitPresent() ? stateOf(LEGACY_UNIT_NAME) : null;
     },
 
     verb: (verb) => systemctl(verb, UNIT_NAME),
@@ -385,9 +405,16 @@ function launchdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
    * other half of the same mess — `rm`ing the file does not unload anything,
    * because launchd holds the copy it read at bootstrap time.
    *
-   * The file is checked first so a machine that never had one asks launchd
-   * nothing at all, which is also what keeps this cheap enough for the status
-   * path to call.
+   * The file is checked first because it is the cheap half and the common
+   * answer on a machine that *has* one — but note which way the short-circuit
+   * runs: a machine that never had a legacy agent has no plist, so it is
+   * exactly that machine which pays for the `launchctl print`. (The systemd
+   * `legacyUnitPresent` above really does ask the filesystem only, because
+   * both of its halves are files.) `observe`, `state` and the `place` warning
+   * each call this, so a plain `actana status` on a healthy Mac spawns a few
+   * extra `launchctl` processes. Correct, and cheap enough — worth memoising
+   * only if it ever shows up, and memoising would have to be invalidated by
+   * `removeLegacyUnit`.
    */
   const legacyPresent = (): boolean =>
     fs.existsSync(legacyPath) ||
--- a/packages/cli/src/actana-status.ts
+++ b/packages/cli/src/actana-status.ts
@@ -57,8 +57,25 @@ export type ActanaStatusReport = {
   target: string | null;
   /** The `wss://` endpoint the Panel dials. */
   endpoint: string | null;
-  /** What the service is called here: `actana-core.service`, `com.actana.core`. */
+  /**
+   * What the service is called here: `actana-core.service`, `com.actana.core`.
+   *
+   * The one the init system **actually has**, not the one setup would write
+   * (#348). The difference is the whole of the bug: this row printed the
+   * constant `com.actana.core` on a machine whose only agent was the pre-rename
+   * `com.actana.harness`, so `Auto-start` named a plist that was not on disk
+   * while the Core in front of the operator ran under another name.
+   */
   serviceName: string | null;
+  /**
+   * The pre-rename service, when this machine still carries one (#348).
+   *
+   * Its own field rather than folded into {@link serviceName}, because both
+   * can be true at once — an in-place upgrade leaves the old agent loaded
+   * beside the new one, and an operator needs told about the one that should
+   * not be there even while the one that should is running fine.
+   */
+  legacyService: string | null;
   /** null when no service is installed (or the platform has no init system). */
   service: ActanaServiceState | null;
   /** How the daemon persists across sessions. null when it could not be read. */
@@ -101,6 +118,12 @@ export function summarizeHealth(report: ActanaStatusReport): ActanaHealth {
     if (!report.container.listening) return "stopped";
     return report.paired ? "healthy" : "degraded";
   }
+  // A pre-rename service on the machine is a degradation whatever else is
+  // true (#348): it runs `current/bin/actana` with the old environment, it
+  // fights the real Core for the port, and `KeepAlive` brings it back. Saying
+  // `healthy` because *something* answered is how this was missed for a
+  // release — so the exit code is non-zero until it is gone.
+  if (report.legacyService) return "degraded";
   if (!report.service) return "degraded";
   if (report.service.activeState === "active") {
     return report.service.subState === "running" && report.paired ? "healthy" : "degraded";
@@ -195,6 +218,16 @@ export function formatActanaStatus(report: ActanaStatusReport): string {
     if (report.persistence) {
       lines.push(row(report.persistence.label, report.persistence.value));
     }
+  }
+
+  // Last of the service rows and unconditional in its own right: it is the one
+  // line that explains every other row that looks wrong, and the remedy is one
+  // command (#348).
+  if (report.legacyService) {
+    lines.push(
+      row("Legacy agent", `${report.legacyService} is loaded — an install from before the rename`),
+      row("", "run `actana setup` to remove it and register this Core properly"),
+    );
   }
 
   lines.push(
--- a/packages/cli/src/actana-status.ts
+++ b/packages/cli/src/actana-status.ts
@@ -166,13 +166,40 @@ function updateRows(report: ActanaStatusReport): string[] {
   ];
 }
 
+/**
+ * The two rows about a service left by an install from before the rename.
+ *
+ * Its own function because it is printed from two places: the ordinary report,
+ * and the `not-installed` page, where every other row is suppressed and this
+ * one must not be.
+ *
+ * "still installed" rather than "is loaded": the detection behind it is true
+ * when the plist merely exists, so an agent that was never bootstrapped would
+ * be overstated by "loaded" — and this is the wording `actana place` already
+ * uses for the same thing.
+ */
+function legacyRows(legacyService: string): string[] {
+  return [
+    row("Legacy agent", `${legacyService} is still installed — from before the rename`),
+    row("", "run `actana setup` to remove it and register this Core properly"),
+  ];
+}
+
 /** Render the report as the text `actana status` prints. */
 export function formatActanaStatus(report: ActanaStatusReport): string {
   const health = summarizeHealth(report);
   const lines: string[] = [HEALTH_LINE[health]];
 
   if (health === "not-installed") {
-    lines.push("", "  Run `actana setup` to install and pair this machine.");
+    // Setup has never run here, so there is nothing to report about a Core —
+    // except when this machine is still carrying the service of one. That row
+    // survives the early return: `actana uninstall --purge-data` removes the
+    // config while `uninstall()` used to leave a pre-rename agent loaded, and
+    // an operator looking at a machine whose port is held needs the one line
+    // that explains it (#348).
+    lines.push("");
+    if (report.legacyService) lines.push(...legacyRows(report.legacyService));
+    lines.push("  Run `actana setup` to install and pair this machine.");
     return lines.join("\n") + "\n";
   }
 
@@ -224,10 +251,7 @@ export function formatActanaStatus(report: ActanaStatusReport): string {
   // line that explains every other row that looks wrong, and the remedy is one
   // command (#348).
   if (report.legacyService) {
-    lines.push(
-      row("Legacy agent", `${report.legacyService} is loaded — an install from before the rename`),
-      row("", "run `actana setup` to remove it and register this Core properly"),
-    );
+    lines.push(...legacyRows(report.legacyService));
   }
 
   lines.push(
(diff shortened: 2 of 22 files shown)


qcentic-dev:
## Acceptance criteria — #348

Every criterion from the issue body and from the triage verdict comment, checked or explicitly unchecked with the reason.

### From the issue body ("Suggested fix")

- [x] **`actana setup` detects a legacy LaunchAgent and migrates it rather than leaving two generations of config.** The
  launchd `removeLegacyUnit` is implemented (it was a hard-coded `return null`); `runActanaSetup` already called it, so the
  macOS path now does what the systemd path always did. Covered end to end in `actana-machine-cli.test.ts` — *"setup boots
  the pre-rename agent out and deletes its plist"*.
- [x] **`actana update` detects it too.** `runActanaUpdate` calls `removeLegacyUnit()` before `service.verb("restart")`.
  Ordering is asserted, not assumed: *"removes a pre-rename service before restarting onto the new tree (#348)"*.
- [x] **`install.sh`'s path fails loudly with instructions.** `install.sh` itself is out of scope for this ticket (#346
  owns it), so the warning lives in `actana place` — the one thing `install.sh` runs (`"$extracted/bin/actana" place`). It
  names the legacy plist and `actana setup`. Warned rather than removed: `place` is the step that does not touch a running
  machine.
- [x] **Detection by label.** `LEGACY_LAUNCH_AGENT_LABEL = "com.actana.harness"` in `actana-launchd.ts`, plus
  `legacyLaunchAgentPath()`.
- [x] **Detection by the material file's CA issuer.** `checkMaterialIdentity` parses the CA and names
  `mission-control-harness-ca` specifically.
- [ ] **Detection by the material file's *age*.** Not implemented. Age is a proxy for the question, and the question is now
  answered directly and cryptographically: the CA that signed the identity, the signature over the server certificate, and
  the certificate/key pair. A file's mtime is changed by any copy, backup or restore, so an age check would refuse good
  material and pass bad material that happens to be recent.
- [x] **`actana status` reads from the same source as the actually-running service.** `Auto-start` reports the label
  launchd has loaded, `State` falls back to the legacy plist, and neither can contradict what is running.

### From the triage verdict comment

**1 — detection of a pre-existing legacy `com.actana.harness` service**

- [x] Add `LEGACY_LAUNCH_AGENT_LABEL = "com.actana.harness"` to `actana-launchd.ts`.
- [x] Implement the launchd `removeLegacyUnit()` as the mirror of the systemd one: plist exists **or** `launchctl print`
  exits 0 → `bootout`, `rm -f` the plist, return the label. Both halves are tested, including the loaded job whose plist
  was already deleted.
- [x] Call it from `runActanaUpdate` before `service.verb("restart")`.
- [x] Have `actana place` print a warning naming the legacy plist and `actana setup`.
- [x] Flip the test at `actana-service.test.ts:253` (*"has nothing to remove on macOS"*). Replaced by four tests covering
  removal, the plist-less loaded job, the machine that never had one, and the agent setup just installed.

**2 — old identity material rejected with a diagnostic rather than an opaque TLS error**

- [x] **(a)** `startCore` refuses to boot when any `AC_HARNESS_*` variable is present, naming the rename and `actana
  setup`, exit 1, in the style of the `AC_CORE_MATERIAL_FILE is not set` refusal. It runs *first*, before any variable is
  read for its value — the survived-the-rename variables are set on that machine too, so every later check would pass.
- [x] **(b)** A non-loopback bind host in loopback mode is **refused**, not merely logged. The triage allowed either; the
  stronger one is right because an unauthenticated shell service on every interface is not something an operator can weigh
  up in a log file.
- [x] **(c)** `checkMaterialIdentity` parses `caCert` with `node:crypto` `X509Certificate` and fails fast on an unexpected
  issuer CN or a server cert/key mismatch, wired into `loadOrMintMaterial` so it surfaces at daemon boot.
- [x] **A test that proves the refusal.** `core-boot-refusals.test.ts` — each old variable, any unknown `AC_HARNESS_*`, the
  message naming the rename, both service names and `actana setup`.
- [x] **A test that proves the plaintext-on-a-public-interface combination cannot happen silently.** Stated over the whole
  space rather than over remembered cases: for every `(remoteMode, host)` pair, a refusal exists exactly when the server
  would be plaintext and reachable from off this machine. Plus a boot-order test — the guards run before
  `new PtyCoreLinkServer(` — and a test that the daemon reads the old spellings nowhere else, so a fallback cannot be
  reintroduced quietly.

**3 — `actana status` reads from the same source of truth as the running service**

- [x] Query launchd for both `com.actana.core` and `com.actana.harness`, via the new `observe()` on the service manager.
- [x] Report the label that is actually loaded as `serviceName`.
- [x] Add an explicit row: `Legacy agent  com.actana.harness is loaded — an install from before the rename` / `run
  \`actana setup\` to remove it and register this Core properly`.
- [x] Do the same file-existence check for the legacy plist so `State` cannot say "not installed" while a job is running.
  Also: a legacy service makes the report `degraded` and the exit code non-zero whatever else is true.
- [ ] **Optional: cross-check `Endpoint` with a one-shot TLS probe and flag a plaintext answer.** Not implemented, and
  marked optional in the triage. Two reasons. `ActanaSystem.waitForPort` — the mechanism suggested — is a TCP connect and
  cannot tell TLS from plaintext, so this needs a new capability rather than a wiring change. And the condition it would
  detect no longer exists: after fix 2 a Core cannot boot plaintext on a non-loopback address at all, so the probe would be
  looking for something the daemon now refuses to be. Worth its own ticket if `status` should verify the wire generally.

### Boundaries held

- [x] `install.sh` untouched — `git diff --name-only` against the base returns no `install.sh`, and its `LINE=` version
  stamp is not in the diff.
- [x] The #347 public-host list and pairing design are built on, not reopened. `core-entry.ts`, `core-first-run.ts`,
  `core-material-store.ts`, `actana-setup.ts` and `actana-cli.ts` keep the signatures that train settled; the changes here
  are additive.
- [x] Nothing merged, `main` and `beta/0.4.2` untouched, PR left as a draft.

### Verification

`pnpm test` — all 6 stages pass (326 test files). `pnpm -r typecheck` clean. `pnpm lint` reports no new warnings.
Written in a Linux container, so nothing launchd-related was executed: the macOS paths are exercised through the same fake
`ActanaSystem` the rest of the service suite uses, end to end across `setup`, `place`, `status` and `update`.

packages/shared/src/core-material-store.ts
```diff
@@ -321,6 +321,105 @@ export function readMaterialFile(filePath: string): MaterialFileRead | null {
...
+/** The common name the Harness-era installer minted, and the tell for #348. */
+export const LEGACY_CA_COMMON_NAME = "mission-control-harness-ca";
+
+/** The common name in an X.509 subject / issuer string, or `""`. */
+function commonNameOf(distinguishedName: string): string {
+  return /^CN=(.*)$/m.exec(distinguishedName)?.[1]?.trim() ?? "";
+}
+
+/** What to tell the operator to do about material this Core cannot use. */
+const REMEDY =
```
mrfqcentic, line 356: **Blocking — this remedy is circular, and it bricks the machine in #348.**

`REMEDY` says "Run `actana setup` to mint this Core's identity again", and `core-first-run.ts:153` throws it at boot. But `actana setup` never re-mints when `material.json` exists:

- `resolveMaterial` (`actana-setup.ts:249`) calls `loadMaterial(...)` → `readMaterialFile(...)?.material` — the type-only check this function was written to replace, so the harness CA passes it.
- `covered` → reuses `existing` verbatim (`actana-setup.ts:263`).
- `moved` / unrecorded → `reissueServerCert(existing, ...)` (`:275`), which signs a new leaf with **`material.caCert` / `material.caKey`** (`core-material-store.ts:178-181`). The CA survives.

Either way the file written back is still `mission-control-harness-ca`, and the daemon refuses again with the same sentence. There is no `--reset` or `--force` on `SETUP_FLAGS` (`actana-cli.ts:467-487`).

I verified rather than reasoned it — minting harness-CA material, running it through `reissueServerCert(["newhost.example.test"])`, and re-checking returns a refusal containing "actana setup". Both the reuse and reissue branches fail.

Concrete path on the reported Mac: `install.sh` → `place` warns ✓ → `actana setup` boots out `com.actana.harness` ✓ → re-blesses the legacy-CA material → installs and starts `com.actana.core` → `startCore` exits 1 → status says run `actana setup`. Loop. Before this PR that `setup` produced a working Core; the issue's "What fixed it" is exactly this command.

Worth stressing: the CA was never the fault. The triage says the root cause is the env-var rename, and the harness CA works once the daemon reads the right variables — so this refusal is new strictness against material that works, which is the "too strict bricks working installs" failure mode.

Two ways out, either is fine:
1. Point `REMEDY` at `actana token regenerate` (`actana-cli.ts:1151`), which really does call `mintFreshMaterial` — and say plainly that every paired client re-pairs.
2. Better: have `resolveMaterial` call `checkMaterialIdentity` and mint fresh when it fails, behind the same confirmation `token regenerate` uses. Then `setup` is the true remedy and the message is already right.

Whichever, please add a test that runs `actana setup` over pre-rename material and asserts the daemon can boot on what it wrote — that is the assertion that would have caught this.

packages/core/src/core-first-run.ts
```diff
@@ -138,6 +143,15 @@ export async function loadOrMintMaterial(opts: LoadOrMintOptions): Promise<LoadO
...
     }
     const existing = read.material;
+    // Before anything is done with it: material whose CA is not this product's
+    // — a file from before the rename, most often — is refused here rather
+    // than presented to a client as an unexplainable TLS failure (#348). The
+    // check is on the load path only; what this function mints is ours by
+    // construction.
+    const unusable = checkMaterialIdentity(existing);
+    if (unusable) {
+      throw new Error(`${opts.materialFile} cannot be used as this Core's identity. ${unusable}`);
```
mrfqcentic: The throw itself is well placed — before the `mintedCoreUuid` write-back, so a refused file is not modified, and `core-entry.ts:527-530` turns it into a clean `exit 1` at boot instead of `wrong version number` at a client three steps away. That half of the triage's fix (c) is delivered.

The problem is only the sentence it carries: see the note on `core-material-store.ts:336`. Once `actana setup` can actually clear pre-rename material, this line needs no change.

packages/cli/src/actana-update.ts
```diff
@@ -180,6 +185,17 @@ export async function runActanaUpdate(opts: UpdateOptions): Promise<UpdateResult
...
 
+  // Before the restart, not after it, and this is the whole of why #348's
+  // machine crash-looped: a pre-rename agent's `ProgramArguments` point at
+  // `current/bin/actana`, which the swap above has just repointed at the new
+  // tree. Left in place it is a second service, with the old environment,
+  // racing the restart below for the same port — and `KeepAlive` brings it
+  // back every time it loses. `actana setup` has always cleaned this up; an
+  // update never reached setup, which is exactly the upgrade path an operator
+  // running `install.sh` takes.
+  const legacy = service.removeLegacyUnit();
```
mrfqcentic: **Blocking — on the #348 machine this removes the only service and then fails.**

The ordering is right and I confirmed it is pinned twice (`actana-update.test.ts:203-225` asserts `["removeLegacyUnit", "restart"]`, and `actana-machine-cli.test.ts:1595-1616` asserts `bootout` before `kickstart` end to end). The problem is the machine where there is nothing to restart onto.

On an in-place upgrade from before the rename there is no `com.actana.core.plist` — `actana setup` was never run post-rename, which is the whole premise of the issue. So after this line boots out `com.actana.harness` and deletes its plist:

- `verb("restart")` (`actana-service.ts:554`) → `isLoaded()` is false → `bootstrap()`
- → `launchctl bootstrap gui/<uid> ~/Library/LaunchAgents/com.actana.core.plist`, which does not exist → non-zero
- → `:201-207` throws `"<v> is installed but com.actana.core would not restart: … Check \`actana logs\`."`

The operator is left with zero auto-start services and pointed at a log file for a daemon that was never started. Both existing tests use a `fakeService` or a machine where `macSetup` ran first, so neither covers it.

Suggestion: when `removeLegacyUnit()` returned a name **and** no current service file exists, say so and name `actana setup` instead of throwing the generic restart error — the tree is placed, the legacy agent is gone, and setup is genuinely the next step. A test for "update on a machine that only ever had the legacy agent" would pin it.

packages/cli/src/actana-service.ts
```diff
@@ -242,23 +294,19 @@ function systemdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
...
-      removeUnit(LEGACY_UNIT_NAME, legacyPath);
+      if (!legacyUnitPresent()) return null;
+      removeUnit(LEGACY_UNIT_NAME, path.join(layout.serviceDir, LEGACY_UNIT_NAME));
       return LEGACY_UNIT_NAME;
     },
 
+    observe() {
+      const legacyName = legacyUnitPresent() ? LEGACY_UNIT_NAME : null;
+      return {
+        name: fs.existsSync(layout.servicePath) ? UNIT_NAME : legacyName,
```
mrfqcentic: **The systemd side did not get the fallback the launchd side did**, and the asymmetry reproduces #348's symptom on Linux.

`observe()` here will report `legacyName` as `name` when the current unit file is absent — good. But `state()` (`:322-336`, unchanged by this PR) still asks systemd only about `UNIT_NAME` and returns `null` on `LoadState=not-found`, where the launchd `state()` (`:522-528`) now falls back to `stateOf(LEGACY_LAUNCH_AGENT_LABEL)`.

So on a Linux box carrying only a running `actana-harness.service`, `actana status` prints:

```
Auto-start        actana-harness.service
State             not installed
```

which is the same self-contradiction the issue opens with, just on the other platform. `summarizeHealth` still returns `degraded` and the `Legacy agent` row still appears, so the operator is not stranded — but the two rows disagree on one screen.

Two smaller things in the same function:

- `name` here is `fs.existsSync(layout.servicePath) ? UNIT_NAME : legacyName`, while launchd's is `fs.existsSync(...) || isLoaded()`. A systemd unit that is loaded with its file removed reads as absent.
- `removeLegacyUnit` still has only two callers (`actana-setup.ts:382`, `actana-update.ts:196`). `cmdUninstall` is not one, and `uninstall()` only touches the current label — so `actana uninstall` on a machine carrying the legacy agent leaves it behind, and `KeepAlive` re-execs a now-missing `current/bin/actana` until launchd throttles it. Now that detection exists, uninstall is the natural third caller.

packages/cli/src/actana-service.ts
```diff
@@ -325,6 +373,25 @@ function launchdServiceManager(opts: ServiceManagerOptions): ActanaServiceManage
...
+   * brings the old agent back at the operator's next login, and it is what an
+   * in-place upgrade leaves untouched. A *loaded* job with no plist left is the
+   * other half of the same mess — `rm`ing the file does not unload anything,
+   * because launchd holds the copy it read at bootstrap time.
+   *
+   * The file is checked first so a machine that never had one asks launchd
+   * nothing at all, which is also what keeps this cheap enough for the status
+   * path to call.
+   */
+  const legacyPresent = (): boolean =>
```
mrfqcentic, line 419: The doc comment above this is inverted, and it matters because the next reader will rely on it.

It says "The file is checked first so a machine that never had one asks launchd nothing at all". But `fs.existsSync(legacyPath) || system.run("launchctl", ["print", legacyTarget()])` short-circuits when the plist **is** present — so the `launchctl` spawn happens precisely on machines that never had a legacy agent.

The two-sided detection is right and I would not change it (a loaded job with no plist left is real, and the comment on `removeLegacyUnit` explains why). It is the stated justification that is backwards. Contrast the systemd `legacyUnitPresent()` at `:240-257`, where the equivalent claim is true because both halves are filesystem-only.

Minor cost that follows: `observe()`, `state()` and `warnAboutLegacyService` each call this, so plain `actana status` on a healthy Mac now spawns several extra `launchctl print` processes plus the domain probe. Not a correctness problem — worth a memo on the result if it ever shows up.

packages/cli/src/actana-status.ts
```diff
@@ -101,6 +118,12 @@ export function summarizeHealth(report: ActanaStatusReport): ActanaHealth {
     if (!report.container.listening) return "stopped";
     return report.paired ? "healthy" : "degraded";
   }
+  // A pre-rename service on the machine is a degradation whatever else is
+  // true (#348): it runs `current/bin/actana` with the old environment, it
+  // fights the real Core for the port, and `KeepAlive` brings it back. Saying
+  // `healthy` because *something* answered is how this was missed for a
+  // release — so the exit code is non-zero until it is gone.
+  if (report.legacyService) return "degraded";
```
mrfqcentic, line 126: The comment says "a legacy service on the machine is a degradation **whatever else is true**", but `:117` returns `not-installed` before this line ever runs, and `formatActanaStatus:172-176` early-returns for that health without emitting the `Legacy agent` rows.

So on a machine with a loaded `com.actana.harness` and no `actana.json` — e.g. after `actana uninstall --purge-data`, which removes `configDir` so `config === null` — `actana status` prints:

```
Core: not installed

  Run `actana setup` to install and pair this machine.
```

and says nothing about the agent that is still loaded and still owns the port. (The exit code is still 1, since `cmdStatus:1019` treats everything but `healthy` as failure — so this is a reporting gap, not a scripting one.)

Combined with `uninstall` not calling `removeLegacyUnit` (see the note on `actana-service.ts:305`), that is a reachable state where the one row an operator needs is suppressed. Hoisting the legacy check above the `installed` gate — or letting `not-installed` fall through to the rows when `legacyService` is set — would close it.

packages/cli/src/actana-status.ts
```diff
@@ -197,6 +220,16 @@ export function formatActanaStatus(report: ActanaStatusReport): string {
     }
   }
 
+  // Last of the service rows and unconditional in its own right: it is the one
+  // line that explains every other row that looks wrong, and the remedy is one
+  // command (#348).
+  if (report.legacyService) {
+    lines.push(
+      row("Legacy agent", `${report.legacyService} is loaded — an install from before the rename`),
```
mrfqcentic: Wording only: this says "is loaded", but `legacyPresent()` is true when the plist merely **exists**, so on a machine where the old agent was never bootstrapped the row overstates what is happening.

The remedy sentence below is right either way. `is still installed` — the phrasing `warnAboutLegacyService` already uses at `actana-cli.ts:838` — would be exact and consistent between the two messages.

packages/core/src/__tests__/core-boot-refusals.test.ts
```diff
@@ -0,0 +1,159 @@
...
+   * Asserted on the source, because there is nothing else to assert it on:
+   * `startCore` is not exported and importing the module boots a daemon. A
+   * refusal that ran after the server was constructed would still print the
+   * right sentence and would still have listened, so the ordering is the
+   * property — not the message.
+   */
+  it("refuses before it builds a server", () => {
+    for (const guard of ["legacyEnvRefusal(process.env)", "plaintextExposureRefusal("]) {
+      expect(entry).toContain(guard);
+      expect(entry.indexOf(guard)).toBeLessThan(entry.indexOf("new PtyCoreLinkServer("));
```
mrfqcentic, line 149: Not blocking, but this is the weakest pin in the diff and it guards the security property.

The rest of this file is genuinely good — the property test at `:107-118` states the invariant over the whole `(remoteMode, host)` space rather than over remembered cases, and `:43-47` pins the prefix behaviour specifically. But the **boot-order** property — the one that actually matters, since a refusal running after `listen()` would print the right sentence and still have served — is asserted by reading `core-entry.ts` as text and comparing `String.indexOf` offsets.

The docstring at `:139-145` is honest about why (`startCore` is not exported, importing the module boots a daemon), and I confirmed the ordering by reading the code, so the property holds today. The gap is what it would not catch: an awaited, socket-opening call inserted between the guards and the server construction passes this unchanged, and an innocuous rename breaks it for no reason.

`scripts/lib/core-smoke.mjs:88-95` already boots a real daemon with a controlled env. Spawning it once with `AC_HARNESS_REMOTE=1` and asserting a non-zero exit with nothing bound to the port would turn this into a behavioural pin. Worth doing whenever this area is next touched.

mrfqcentic (changes requested):
## Request changes — the security fix is right, the recovery path is not

Reviewed `610db11` against `feat/0.4.2-install-and-identity` @ `619cd55` (the merge-base is exactly that commit, so #346 and #347 are in). Every claim below was checked by reading the enforcing line, not the description.

**What is right, and it is the most important part.** The boot refusal is on the prefix, not on a list — `core-boot-refusals.ts:36-40` filters `Object.keys(env)` by `startsWith("AC_HARNESS_")`, so a fourth stale variable stops the boot exactly like the three named ones. Both guards are ahead of `new PtyCoreLinkServer(` at `core-entry.ts:697`, which `grep` confirms is the **only** server construction site in the repo. I enumerated the space rather than trusting the summary: remote mode always reaches `serverOpts.tls` (material file mandatory, any `loadOrMintMaterial` throw exits 1), loopback mode is now pinned to a loopback bind, the daemon reads no config file, and the default is `127.0.0.1`. Every non-loopback bind this product writes sets `AC_CORE_REMOTE=1` — `actana-setup.ts:422`, `core.Dockerfile:265`, and the pre-`8aa5b5b` setup too, so no existing install regresses. **I could not construct a surviving path to unauthenticated plaintext off-machine.** The launchd `removeLegacyUnit` really mirrors the systemd branch, `update` really calls it before the restart, `place` really warns, the pinned no-op test is really flipped (one `it(` removed, four added, no skips anywhere in the diff), `install.sh` is a byte-identical blob with `LINE="0.4.2"` unmoved, and nothing #347 owns is reverted. Commit bodies max at 127 chars and commitlint exits 0 over the branch.

**Why this is blocked anyway.** Two problems land on the exact machine #348 describes, and both are introduced by this PR.

The first is the one that matters: **the remedy every new message prints does not work.** `checkMaterialIdentity` refuses pre-rename material and tells the operator to run `actana setup` — but `actana setup` cannot clear that material. `resolveMaterial` (`actana-setup.ts:249`) loads it with `loadMaterial`, the same type-only read this check was written to replace, and then either reuses it or calls `reissueServerCert`, which signs a new leaf **with the existing `caCert`/`caKey`**. The harness CA survives both branches. I ran it to be sure: minting harness-CA material, passing it through `reissueServerCert`, and re-checking still returns a refusal, still ending in "Run `actana setup`". So on the reported Mac the operator now gets: `place` warns correctly, `setup` boots out the legacy agent correctly, then installs `com.actana.core` over material it just re-blessed, starts it, and the daemon exits 1 telling them to run the command they are standing in. Before this PR that same `setup` produced a working Core — the issue's own "What fixed it" says so. Note the CA was never the fault here; the triage was explicit that the root cause is the env-var rename, and the harness CA works fine once the daemon reads the right variables.

The second: on a machine with no `com.actana.core` plist — which is the #348 machine, since setup was never run post-rename — `update` now deletes the only service and then fails. `removeLegacyUnit()` runs unconditionally at `actana-update.ts:196`, then `verb("restart")` finds `isLoaded()` false, falls to `bootstrap()` against a plist that does not exist, and `runActanaUpdate` throws pointing at `actana logs` for a daemon that never started. Zero auto-start services left. Safer than an unauthenticated Core on `0.0.0.0`, but not a recovery.

There is also a real gap against the third acceptance criterion: the systemd `state()` never got the legacy fallback the launchd one did, so on Linux `Auto-start` names `actana-harness.service` while `State` says `not installed` — the same self-contradiction #348 is about, on the other platform.

Details inline. The design here is sound and the security reasoning is genuinely good — this is about making the way out of the state actually lead out of it.

qcentic-dev:
## Pushed — the way out of the state now leads out of it

Thank you for running the recovery rather than reading it. Both blockers and every inline note are addressed in
`95166d3`, `7b083e6`, `5a80da0`, `aa914a0`.

### 1. The circular remedy (`core-material-store.ts:336`, `core-first-run.ts:153`) — fixed, and the strictness is withdrawn

You are right on both halves, and the second is the one that changed the design. **The CA was never the fault.** The
triage says the root cause is the env-var rename, `core-boot-refusals.ts` stops that on its own, and a Harness-era CA
serves TLS exactly as it always did — so refusing it was new strictness against material that works, and it would have
cost every paired client its pairing for nothing.

So `checkMaterialIdentity` now grades what it finds:

- **`unusable`** — no configuration can make it serve TLS: an unparseable PEM, a leaf the CA beside it did not sign, a
  certificate and key that are not a pair. The daemon refuses to boot on this, as before.
- **`foreign`** — it works, it simply is not ours. A pre-rename CA is this. The daemon **boots**, and logs
  `core-material.foreign-identity` naming the provenance.

Provenance is checked *last*, so pre-rename material with a mismatched key comes back `unusable` and is not described as
"old" — that would send the operator after the wrong thing.

And `setup` is now a remedy that works. `resolveMaterial` runs the same function the daemon boots with and, on
`unusable`, mints from a fresh CA instead of handing back what it was given — behind the confirmation `token regenerate`
asks before the same loss, with a new `re-minted` outcome so `actana setup` tells the operator every paired client is
locked out. Declining stops setup rather than writing material the daemon will refuse.

The test you asked for is `actana-setup.test.ts` → *"re-mints it from a fresh CA instead of re-blessing it"*: it plants
material that cannot be served, runs `runActanaSetup`, and asserts `checkMaterialIdentity` on what setup wrote is null —
that function being what `loadOrMintMaterial` calls at boot, so a null there is a clean boot. Beside it,
*"keeps material from before the rename"* plants a real harness-CA identity and asserts setup does **not** re-mint it, and
`core-first-run.test.ts` asserts the daemon loads it.

### 2. `update` removing the only service (`actana-update.ts:196`) — fixed

`removeLegacyUnit()` now runs only when this machine has a current service to take over. On a machine whose only service
is the legacy one it stays, and update says so and names `actana setup`. Nothing is restarted there either, so `listening`
comes back `null` rather than `false` — no port was asked to answer — and `cmdUpdate` reads that as its own case, exiting
non-zero **without** the "the daemon restarted" sentence and without pointing at `actana logs`. Asked of `observe()`
rather than of the filesystem, since the init system is what knows.

Pinned at the unit level for both no-service shapes, and end to end through the real launchd manager in
`actana-machine-cli.test.ts` → *"update keeps the legacy agent when it is the machine's only service"*.

### 3. The systemd asymmetry (`actana-service.ts:305`) — fixed, with the two smaller notes

`state()` on systemd now falls back to the legacy unit through a `stateOf` helper shaped like the launchd one, so
`Auto-start` and `State` cannot disagree on Linux either. Covered by a Linux end-to-end status test.

- `observe()` asks systemd about a unit whose file was deleted — but **only when a legacy unit is present**, because
  `actana place` calls this and *"asks the init system for nothing at all"* is a property your own test pins. A machine
  with a legacy unit is getting a warning printed anyway; that is what earns the one `systemctl show`.
- `uninstall` is now the third caller of `removeLegacyUnit`. Good catch — an agent left re-execing `current/bin/actana`
  out of a tree uninstall has just deleted is worse than a leftover file.

### 4. `actana-service.ts:392` — the comment was backwards

Fixed. The plist check short-circuits when the plist **is** present, so it is the machine that never had a legacy agent
which pays for the `launchctl print`. The comment now says that, notes the extra spawns on `status`, and records why
memoising would need invalidating by `removeLegacyUnit`. Detection itself unchanged, as you suggested.

### 5. `actana-status.ts:126` and `:228` — both fixed

The `Legacy agent` rows are now a shared helper printed from the `not-installed` page too, so the
`uninstall --purge-data` state you found is no longer silent about an agent that still owns the port. And the row says
`is still installed` — your wording, matching `place`, and exact for a plist that exists without being bootstrapped.

### 6. `core-boot-refusals.test.ts:149` — taken, not deferred

You were right that this was the weakest pin guarding the strongest property. `smoke-core-tarball.mjs` now spawns the
released launcher twice more, in both hostile environments, and asserts it exited non-zero, never printed the listening
sentinel, said why, and left nothing on the port — after the good boot and on the same port, so "nothing is listening" is
about the refusal. Verified against a real built tarball locally:

```
[tarball-smoke] refused a pre-rename environment: exit 1, nothing bound on 38661
[tarball-smoke] refused plaintext on a public interface: exit 1, nothing bound on 38661
```

The source-order test stays as the cheap unit-level guard.

---

`pnpm test` — all 6 stages pass. `pnpm -r typecheck` clean, `pnpm lint` no new warnings. `install.sh` still untouched and
#346/#347 still not reopened. Ready for another look.

packages/shared/src/core-material-store.ts
qcentic-dev: Fixed in `95166d3`, and the strictness withdrawn as well as the loop broken. `checkMaterialIdentity` now grades: `unusable` (unparseable, leaf the CA did not sign, cert/key not a pair) refuses the boot; `foreign` — which a Harness-era CA is — boots and is logged. You were right that the CA was never the fault. `resolveMaterial` runs the same check and mints from a fresh CA on `unusable`, behind the `token regenerate` confirmation, with a new `re-minted` outcome so setup says every paired client is locked out. The test you asked for: `actana-setup.test.ts` → "re-mints it from a fresh CA instead of re-blessing it" asserts `checkMaterialIdentity` on what setup wrote is null, i.e. the daemon boots on it.

packages/core/src/core-first-run.ts
qcentic-dev: The throw stays where it is, and now carries a sentence that is true — `actana setup` re-mints `unusable` material rather than handing it back (`95166d3`). Pre-rename material no longer reaches this throw at all: it is `foreign`, so the daemon boots on it and logs `core-material.foreign-identity`.

packages/cli/src/actana-update.ts
qcentic-dev: Fixed in `7b083e6`. `removeLegacyUnit()` runs only when this machine has a current service to take over; where the legacy one is the only service it stays, and update names `actana setup`. Nothing is restarted there, so `listening` is `null` rather than `false`, and `cmdUpdate` handles that as its own case — non-zero, without the "the daemon restarted" sentence and without pointing at `actana logs`. Pinned at the unit level for both no-service shapes and end to end in `actana-machine-cli.test.ts` → "update keeps the legacy agent when it is the machine's only service".

packages/cli/src/actana-service.ts
qcentic-dev: All three fixed in `5a80da0`. `state()` on systemd falls back to the legacy unit through a `stateOf` helper shaped like launchd's, with a Linux end-to-end status test. `observe()` now asks systemd about a file-less unit — but only when a legacy unit is present, because `place` calls this and "asks the init system for nothing at all" is a property your own test pins; a machine with a legacy unit is getting a warning anyway. And `uninstall` is the third caller of `removeLegacyUnit` — an agent re-execing `current/bin/actana` out of a tree uninstall just deleted is worse than a leftover file.

packages/cli/src/actana-service.ts
qcentic-dev: Fixed in `5a80da0` — the comment, not the detection, exactly as you suggested. It now says which way the short-circuit runs (the machine that never had a legacy agent is the one that pays for the `launchctl print`), notes the extra spawns on the `status` path, and records that memoising would have to be invalidated by `removeLegacyUnit`.

packages/cli/src/actana-status.ts
qcentic-dev: Fixed in `5a80da0`. The rows are a shared `legacyRows` helper now, printed from the `not-installed` page too, so the `uninstall --purge-data` state you found is no longer silent about an agent that still owns the port. Test: "still says it on a machine setup has never run on".

packages/cli/src/actana-status.ts
qcentic-dev: Fixed in `5a80da0` — `is still installed`, your wording, matching what `place` already prints. Exact for a plist that exists without ever having been bootstrapped.

packages/core/src/__tests__/core-boot-refusals.test.ts
qcentic-dev: Taken rather than deferred, in `aa914a0`. `smoke-core-tarball.mjs` spawns the released launcher twice more — a pre-rename environment, and the plaintext-on-a-public-interface shape — and asserts it exited non-zero, never printed the listening sentinel, said why, and left nothing on the port. After the good boot and on the same port, so "nothing is listening" is about the refusal. Verified against a real built tarball: `refused a pre-rename environment: exit 1, nothing bound on 38661`. The source-order test stays as the cheap guard.

mrfqcentic (approved):
## Approve — the recovery path now genuinely leads out of the state

Re-reviewed `aa914a0` against `feat/0.4.2-install-and-identity` @ `619cd55` (merge-base still exactly that commit). Both blocking findings are fixed, the systemd gap is closed, and all six items I passed in round one still hold. I re-verified by running code, not by reading the replies.

### 1. The circular remedy — fixed, and fixed the right way

The severity split is the correct call, and the part that matters is that re-minting is keyed on **cryptographic coherence, not provenance**: `unusable` is unparseable bytes, a leaf the CA beside it did not sign, or a cert/key that are not a pair. Nothing in that set can be true of material that is currently serving TLS, so the recovery cannot destroy a working install. Provenance is checked **last** (`core-material-store.ts:434-449`), which also means harness-CA material with a broken key is reported as broken rather than as old — the right call, and the diagnostic an operator can act on.

I wrote my own probes rather than trusting the tests, and asserted fields the author's tests do not:

**Direction A — an unusable install really recovers.** Across all four unusable shapes (mismatched key, leaf the CA did not sign, unparseable `caCert`, unparseable `serverKey`): `runActanaSetup` returns `re-minted`, and `checkMaterialIdentity` on what setup wrote is `null` — that being the exact function `loadOrMintMaterial` calls at boot, so it is a clean boot. The new CA is genuinely fresh, not the broken one re-blessed. Harness CA **plus** a broken key is also re-minted, which is right: it could never have served.

**Direction B — a working harness-CA install is not re-minted and not broken.** With the SAN already covering the configured host, setup returns `reused` and `caCert`, `caKey`, `bearerSecret`, `coreId`, `coreUuid`, `clientCert`, `clientKey`, `serverCert` and `serverKey` are all byte-identical afterwards. With the public host moved, it returns `reissued`, the leaf is re-signed **by the same harness CA**, every identity field above survives, and the result is still `foreign`. Interactive setup is never prompted about a fresh identity for it. At the daemon layer, `loadOrMintMaterial` loads pre-rename material without throwing, preserves every identity field, does not rewrite the CA on disk, and still re-issues from that foreign CA on a host move with the next boot a plain load. Unusable material still throws.

That is the property the fix had to have, in both directions.

### 2. `update` leaving no service — fixed, and safe beyond the one sequence

`removeLegacyUnit()` now runs only when `observe().name === service.name`, and the no-service branch returns `listening: null` with a message naming `actana setup` instead of bootstrapping a plist that does not exist. Asking `observe()` rather than the filesystem is the right source.

Rather than re-run the author's `fakeService` tests, I drove `runActanaUpdate` through the **real** launchd and systemd managers over the full cross product of on-disk state × what the init system reports — current only, current + legacy, legacy only, and neither, on both platforms — and asserted the invariant directly: *if a machine had at least one auto-start service before, it has at least one after*. It holds in all eight. On the legacy-only shape no `bootout` is issued at all and the plist is still there afterwards. The one edge that reaches `removeLegacyUnit` without a current unit *file* is a current unit that is loaded with its file deleted by hand — there `verb("restart")` kickstarts the loaded job, so it is still safe.

### 3. systemd `state()` — fixed

`stateOf` mirrors launchd's, and `state()` falls back to the legacy unit. I confirmed with the real manager: on a box carrying only a running `actana-harness.service`, `observe()` and `state()` now agree instead of printing `Auto-start actana-harness.service` over `State not installed`. `observe()` asks systemd about a file-less unit only when a legacy unit is present, so `place` still asks the init system nothing on an ordinary machine — I checked that too. `uninstall` is now the third caller of `removeLegacyUnit`, the `Legacy agent` rows survive the `not-installed` page, and the wording is `is still installed`.

### 4. No regression on what already passed

`core-boot-refusals.ts`, `core-entry.ts` and `pty-core-link-server.ts` are byte-identical to `610db11`. The refusal is still the prefix filter over `Object.keys(env)`; both guards still precede the single `new PtyCoreLinkServer(` at `core-entry.ts:697`; no `AC_HARNESS_*` reader exists anywhere. `install.sh` is still blob `6cbd0bc`, identical to base, `LINE="0.4.2"` at line 80. Nothing #347 owns is reopened — `public-hosts.ts`, `core-pairing-routes.ts`, `core-pairing-wiring.ts`, `actana-pair.ts` and `actana-config.ts` are byte-identical to base, and `MaterialOutcome` is widened by one member, not resignatured. Commit bodies max at 127 chars over all ten commits; commitlint exits 0.

Locally: shared 509, core 903, cli 1091 passed; typecheck clean; lint unchanged from base (6 pre-existing warnings, 0 errors). CI green on `aa914a0`.

### On the smoke test

Taking that one rather than deferring it was the right call, and it is the strongest single change in this round. I pulled the CI log for `Smoke — Core release tarball (linux-x64)` to confirm it really ran against a built tarball:

```
[tarball-smoke] refused a pre-rename environment: exit 1, nothing bound on 35907
[tarball-smoke] refused plaintext on a public interface: exit 1, nothing bound on 35907
```

Non-zero exit, no listening sentinel, a reason in the output, and nothing bound on the port that the good boot had just used. That is a behavioural pin on the security property, not a source-text grep.

### Two observations, neither blocking and neither needing action here

- `checkMaterialIdentity` treats a throw from `server.verify(ca.publicKey)` as `unusable`, which routes to a re-mint. Every identity in the field comes from `selfsigned`, so the algorithm space is narrow and I could not construct a case — but it is the one path where a verification failure that is not the material's fault would cost a pairing. Worth a thought if the certificate toolchain ever changes.
- Coherence is checked, expiry is not. An expired-but-coherent identity still boots and still fails at the client. Pre-existing, out of scope for #348, and arguably right given the scope note in the docstring.

Good work on the correction — withdrawing the strictness rather than defending it was the right read of the triage.

Issue #351, opened 2026-08-28
Three pre-existing public-host and endpoint gaps surfaced by the #350 review


qcentic-dev:
**All three of these predate [#350](https://github.com/actana/control/pull/350) and were surfaced by
its review, not introduced by it.** #350's review verified all eight acceptance criteria as MET and
found no blocker; these are the items it explicitly listed as *moved but not introduced*, filed here
so they are not lost when that PR merges. None of them is a regression, and none was fixed in #350 —
fixing them there would have been scope the ticket did not own.

Each names the line as it stands on `feat/multi-san-public-host` after the review fixes
(`24cd65a`). Line numbers will drift; the reasoning is the durable part.

---

## 1. `buildPairingEndpointResolver` does not bracket an IPv6 host, where `endpointFor` does

`packages/core/src/core-pairing-wiring.ts` builds the redeem response's endpoint as:

```ts
return `wss://${host}:${opts.port}`;
```

`packages/cli/src/actana-config.ts`'s `endpointFor` — the other place this repository turns a host
and a port into a `wss://` URL — brackets an IPv6 literal first:

```ts
const host = config.publicHost.includes(":") && !config.publicHost.startsWith("[")
  ? `[${config.publicHost}]`
  : config.publicHost;
```

So a Core configured with a bare IPv6 address hands a redeeming client `wss://fd00::1:8443`, which
is not a parseable URL — the port is unrecoverable from it.

**Two places, one rule, and they disagree.** The pre-existing half is that the rule was only ever
written down in one of them; #350 added a second producer of the same string and copied the simpler
spelling. The fix is one bracketing helper both call, not a second copy of the ternary.

**Related, and the reason this is not purely cosmetic:** `isIp` in
`packages/shared/src/core-cert-material.ts` is IPv4-only —

```ts
return /^\d{1,3}(\.\d{1,3}){3}$/.test(host);
```

— so an IPv6 entry in `ACTANA_PUBLIC_HOST` lands in the certificate as a `DNS:` SAN rather than an
`iPAddress:` one, and Node checks IP literals against `iPAddress` entries and nothing else. An
operator who configures a Core on an IPv6 address therefore gets a certificate that does not verify
*and* an endpoint that does not parse. Both halves want fixing together; neither is reachable
without deliberately configuring IPv6, which is why this has gone unnoticed.

## 2. ~~A flagless `actana setup` re-run prefers its own guess over the recorded config~~

[...]

PR #353, opened 2026-08-28
feat: install and identity — several addresses per Core, a bash 3.2-safe installer, and the pre-rename service removed

### LINKED ISSUE #346: install.sh: unbound variable on macOS under a UTF-8 locale (bash 3.2 multibyte bug)

### LINKED ISSUE #347: Core identity is single-SAN: switching ACTANA_PUBLIC_HOST forces every paired client to re-pair

### LINKED ISSUE #348: install.sh/actana setup leaves a stale, incompatible LaunchAgent + identity material after an in-place upgrade

qcentic-dev:
The gate pull request for the **0.4.2 train**. Three tickets, all merged into `feat/0.4.2-install-and-identity` and
squashing into one commit on `beta/0.4.2`.

Closes #346. Closes #347. Closes #348.

The theme is the one the train is named for: getting onto this version, and staying the same Core once you are there. All
three tickets came out of the same in-place upgrade of a bare-metal Core.

## ⚠️ Security: a Core could serve unauthenticated plaintext core-link on a public interface (#348)

**This is the fix that decides how quickly this train ships.**

A machine installed before the Harness → Core rename carries a LaunchAgent that sets `AC_HARNESS_REMOTE`,
`AC_HARNESS_PUBLIC_HOST` and `AC_HARNESS_MATERIAL_FILE`. Only three variables kept their names across the rename —
`AC_CORE_LINK_PORT`, `AC_CORE_LINK_HOST` and `AC_USER_DATA_DIR` — and the daemon reads the `AC_CORE_` spellings and
nothing else, with no fallback and no warning.

So under that agent, remote mode, the public host and the material file all silently vanished while the port and the bind
address survived. The Core came up in **loopback mode** — plain `ws://`, no TLS, no client certificate, no bearer, a
transport that trusts every connection because only this machine is supposed to reach it — bound to the old plist's
`0.0.0.0` at the port it named. That is an unauthenticated service able to start processes and read the machine's files,
offered to anything that can route to it. The only symptom an operator saw was `wrong version number` at a TLS client,
which reads as a broken certificate rather than as an open door.

The daemon now refuses to boot in both shapes: any `AC_HARNESS_*` variable stops it before anything else is read, naming
the rename and the service that set it, and a non-loopback bind without `AC_CORE_REMOTE` stops it on its own terms.
Refused rather than logged, and no compatibility fallback was added — a fallback would have kept that server running on
purpose. Every legitimate non-loopback bind this product writes sets `AC_CORE_REMOTE=1`, so nothing that works today
stops working. The refusals are pinned by a property test over the whole `(remote mode, bind host)` space, and by the
release-tarball smoke, which boots the real daemon in both environments and asserts it exited non-zero with nothing
listening on the port.

## The stale service and identity an upgrade left behind (#348)

The same pre-rename LaunchAgent is what an in-place upgrade could not see. Its `ProgramArguments` point at
`current/bin/actana`, so swapping the tree hands the old agent the new binary, and `KeepAlive` restarts it into a race
for the port.

- `actana setup` and `actana update` now remove it on macOS as they always did on Linux — the launchd implementation was
  a hard-coded no-op justified by a decision that had since been amended. `update` removes it only when there is a
  current service to take over, so an upgrade can never leave a machine with no auto-start service at all.
- `actana place` — what `install.sh` runs — warns when the old agent is there, since an operator who only runs the
  installer never reaches setup. `actana uninstall` now removes it too.
- `actana status` stops contradicting itself. It reported a hard-coded service name and checked one fixed filename, so a
  machine running the old agent read as `not installed` while the header said `degraded`. It now reports the service the
  init system actually loaded, on both platforms, with an explicit `Legacy agent` row and a non-zero exit until it is gone.
- Identity material is checked at boot for what a TLS handshake actually needs — a leaf its CA signed, a certificate and
  key that agree — instead of failing later as an opaque wire error. Material from the pre-rename CA is **kept**: it
  works, and the rename broke the environment around it rather than the identity. Only material that genuinely cannot be
  served is refused, and `actana setup` re-mints that rather than re-blessing it, so the remedy the message names is one
  that leads out of the state.

## A Core can have several addresses, and a pairing code picks one (#347)

`ACTANA_PUBLIC_HOST` was a single name baked into the server certificate's SAN, so a Core reachable as `core` on a
compose network and as a LAN address from a laptop could only ever be validly one of them — and changing the answer
forced every paired client to pair again.

It is now a comma-separated list. Every entry becomes a SAN, the first is the primary, and `actana pair new` chooses
which address the code hands back. A single value parses to a list of one and behaves exactly as it always did: the same
certificate, the same endpoint, no re-pairing. Adding an address to an installed Core re-signs its certificate from the
CA already on disk, so nothing paired loses anything.

## An installer that survives bash 3.2 under a UTF-8 locale (#346)

`install.sh` used a multi-byte ellipsis inside a construct that macOS's bash 3.2 mis-parses under a UTF-8 locale, failing
with an unbound variable before anything was installed — on the one platform whose system bash is still 3.2. The
character is gone and the shape it appeared in is guarded, so the failure cannot come back unnoticed.

---

## Notes for the reviewer

- **Base is `beta/0.4.2`, the open train — not `main`.** Nothing here targets `main`, and no tag or workflow dispatch is
  part of this pull request.
- **Squash is this repository's only merge method**, so the three ticket subjects collapse into the single train commit
  this title describes.
- **`Promotion gate` is red on the train → `main` pull request by design, not on this one.** This is a ticket-side gate
  into the train.
- All seven version stamps on the branch read `0.4.2` — the six workspace manifests and the `LINE=` stamp in
  `install.sh` — which is what `Train rules` asserts.
- The branch is three commits ahead of `beta/0.4.2` and zero behind, so no merge from the train was needed.


qcentic-adm (dismissed):
## Gate review — 0.4.2 train (#346, #347, #348 combined)

**Verdict: APPROVE.** Reviewed as the integration of the three already-merged tickets, not as three re-reviews. Every finding below is non-blocking; none changes what the train ships.

### What was verified by reading the enforcing line

1. **Combined signatures are coherent.** Every path that once took one host now takes a list: `mintFreshMaterial` / `reissueServerCert` / `checkServerCertHost` (`core-material-store.ts`), `LoadOrMintOptions.publicHosts` (`core-first-run.ts:70`), `SetupOptions.publicHosts` (`actana-setup.ts:134`), `ContainerContract.publicHosts`, and `cmdTokenRegenerate` goes through `configPublicHosts` (`actana-cli.ts:1161`). Host validation is one function (`parsePublicHosts`) called at the three entry points (setup flag, container contract, `core-entry`), not three validators. No half-migrated call site found.
2. **Boot order composes correctly** (`core-entry.ts:156` → `:165-173` → `:181` → remote block `:447+`): `AC_HARNESS_*` refusal runs before any variable is read for its value; the plaintext-exposure refusal runs before `bootstrapCoreDb` and long before `new PtyCoreLinkServer`; remote mode without `AC_CORE_MATERIAL_FILE` is refused at `:499`. The only two `listen()` calls in the Core (`pty-core-link-server.ts:2386,2439`) both bind `opts.host` from this path; the container image bakes `AC_CORE_REMOTE=1` + `0.0.0.0` (`deploy/core.Dockerfile:265-266`) so it passes. `actana daemon` loads `core-entry.cjs` in-process (`actana-cli-entry.ts:186-199`) so the pre-rename plist's env (`AC_HARNESS_REMOTE`, `AC_CORE_LINK_HOST=0.0.0.0` — confirmed from the initial commit's `packages/harness/src/actana-setup.ts:223-232`) reaches the refusal. **No (remote, bind-host) combination boots a plaintext core-link off-loopback.**
3. **Operator story (pre-rename Mac, in-place upgrade):** `install.sh` → `place` warns about `com.actana.harness` and names `actana setup` → `update` on a legacy-only machine leaves the agent, says so truthfully, exits 1 naming `setup` (`actana-update.ts:209-240`, `actana-cli.ts:1283-1291`) → `setup` boots the legacy agent out **first** (`actana-setup.ts:431`), keeps pre-rename material as `foreign` (`core-material-store.ts:445`), backfills `serverHosts` on the `reused` path so `pair new --public-host` works afterwards → `status` reads `observe()` and exits non-zero while the legacy agent exists. Every named remedy clears the condition it names. No circle.
4. **Docs/ADRs:** 0038 D1–D7 match the code; 0016 D15/D18 and 0034 D2 amendments are appended, dated and consistent; `docs/README.md` indexed. Nothing stale found in `AGENTS.md`, `CLAUDE.md`, `CONTRIBUTING.md`, `docs/ci-cd.md`.
5. **Train mechanics:** all seven stamps read `0.4.2` (six manifests + `install.sh:80`), `Train rules` green, branch is 3 ahead / 0 behind `beta/0.4.2` (clean fast-forward), only `.github/workflows/ci.yml` touched (a Lint step) — nothing touches `main`, a release line, or a tag.
6. Local runs: core 73/73, shared 117/117, cli 424/424, scripts 23/23; `scripts/install-sh-ascii-guard.sh install.sh` passes.

### Non-blocking comments (follow-up ticket material)

- **`actana-cli.ts:843-847` (`warnAboutLegacyService`)** — "it is now starting the version just placed with the old environment" is ahead of the facts: `stopDaemonBeingReplaced` only stops the *current* label (`isActive` → `com.actana.core`), so a 0.4.1 daemon already running under `com.actana.harness` keeps running its old inode — still plaintext on `0.0.0.0` — until it restarts. The line above it ("it is still running 0.4.1") is the true one. Either bootout the legacy job at `place` too, or say "when it next restarts, it will start the version just placed and the daemon will refuse to boot; until then the old daemon is still up — run `actana setup` now". The PR does not make this state worse than before the upgrade, and `setup` closes it, so not blocking.
- **`core-boot-refusals.ts:77` / `core-material-store.ts` `REMEDY` / `actana-status.ts:183`** all say bare `actana setup`. A bare re-run defaults `--public-host` to the first routable IPv4 (`actana-cli.ts:555-558`), which for a Core previously at e.g. `host.docker.internal` is a *move* and re-points the endpoint. The operator is told at that point, so it is not a circle — but naming `actana setup --public-host <the address you already use>` in these messages would save the re-address.
- **`packages/cli/src/actana-systemd.ts:25-32`** — `LEGACY_UNIT_NAME`'s doc still says "no release ever shipped under this name … developer machines only … deletable one release after 0.1.0". #348's premise and `LEGACY_LAUNCH_AGENT_LABEL`'s doc (`actana-launchd.ts:25-38`) contradict it, and `actana-setup.ts:430` points readers there for "when".
- **`deploy/README.md:86-90`** — "That one value is three things at once … 3. the endpoint every pairing hands back" is unchanged, while ADR 0038 §C says this sentence "is now two-and-a-half". The new section below it is correct; the old sentence should say "by default".
- **`scripts/lib/core-smoke.mjs:321-332`** — `assertBootsAndDials`'s JSDoc is orphaned above `assertRefusesUnsafeEnv`'s own block.
- **`scripts/install-sh-ascii-guard.sh:101`** — "which is what this script used to carry" describes a revision that never landed.
- **`packages/cli/src/actana-pair.ts:347`** — `Configured (ACTANA_PUBLIC_HOST): …` is printed on metal too, where the variable is `--public-host` (`core-entry.ts:466-480` argues exactly this point).
- **Duplication between the tickets:** the "string list → trim → drop empties" shape appears in `core-cert-material.ts:202`, `core-material-store.ts:471`, `actana-config.ts:108-111` and `pairing-store.ts:441` beside `parsePublicHosts`; `LOOPBACK_HOSTS` names two different sets in `actana-pair.ts:295` and `core-boot-refusals.ts:25`; `wss://host:port` is built three ways (`actana-config.ts` is IPv6-bracket aware, `core-pairing-wiring.ts:111` and `actana-cli.ts:610` are not).
- **PR body wording:** "property test over the whole (remote mode, bind host) space" is an exhaustive 8×2 loop (`core-boot-refusals.test.ts:107-118`) — fine, just not generative.
- **Spec deviations, both deliberate and argued in code, for the owner to ratify with ADR 0038:** (a) #347 ships the per-pairing half the triage recommended against — the redeem route still takes nothing from the request (`core-pairing-routes.ts:425`, body parser `:484-505`, `Host`-lies test `core-pairing-redeem.test.ts:679-697`), so the security constraint is honoured; (b) #348(c) unexpected issuer CN is warn-and-keep rather than refuse (`core-material-store.ts:371-379`), which is the right call.

mrfqcentic (changes requested):
## Gate review — 0.4.2 train (#346 + #347 + #348 combined)

This is the integration review, not a re-run of the three per-ticket reviews. I read #346, #347 and #348 in full and then the whole diff of `feat/0.4.2-install-and-identity` against `beta/0.4.2` (67 files, +4746/−352), looking for defects that only exist because all three are on one branch.

**The security property this train exists to establish holds.** I traced it at the enforcing lines, not from the description:

- `core-entry.ts:156-160` — `legacyEnvRefusal` runs before a single value is read, which is the only correct position: `AC_USER_DATA_DIR`, `AC_CORE_LINK_PORT` and `AC_CORE_LINK_HOST` are all set on a pre-rename machine, so every later check would have passed.
- `core-entry.ts:181-185` — `plaintextExposureRefusal` is independent of the first, so a plist that dropped only `AC_HARNESS_REMOTE` is still caught on its own terms.
- `core-entry.ts:391-393` binds `host`; `core-entry.ts:565-570` sets `serverOpts.tls` unconditionally inside `if (remoteMode)`, and every exit before it is `process.exit(1)`. So *no TLS ⟹ `!remoteMode` ⟹ the exposure refusal already passed ⟹ loopback bind.* There is no combination that serves plaintext core-link off-machine.
- `isLoopbackHost` (`core-boot-refusals.ts:28-33`) fails **closed** on anything it does not recognise (`::`, `0:0:0:0:0:0:0:1`, a name that resolves to loopback), which is the right direction for this check.
- `AC_HARNESS_*` as a blanket prefix is safe: I grepped the whole repo and no live code path reads any variable with that prefix.

**The operator's loop out of the broken state closes.** `place` warns (`actana-cli.ts:817`), `update` refuses to remove the only service and names `actana setup` (`actana-update.ts:196-236`), `setup` boots out and deletes the agent (`actana-service.ts:454-468`), `status` prints the `Legacy agent` row on both the ordinary and the `not-installed` pages and exits non-zero (`actana-cli.ts:985`/`1029` — anything but `healthy` is exit 1), and `pair new` then works. The names in the daemon's refusal (`com.actana.harness`, `actana-harness.service`) match `LEGACY_LAUNCH_AGENT_LABEL` and `LEGACY_UNIT_NAME`. **The previous round's failure — a refusal naming a command that could not clear the condition — is genuinely fixed:** the refusal names `actana setup`, and `runActanaSetup` calls `service.removeLegacyUnit()` at `actana-setup.ts:390`, first, before the tree is swapped.

I could not fault the #347/#348 call-site merge either: every `mintFreshMaterial` / `reissueServerCert` / `checkServerCertHost` / `generateCertMaterial` / `issueServerCert` site takes a list, and the single-host survivors (`endpointFor(config)`, `ContainerContract.publicHost`) are the documented "an endpoint is one address" case rather than a half-migration.

---

### Blocking — one root cause, three places

**ADR 0038 D3a's "a widening is not a move" distinction landed only in the daemon.** `core-first-run.test.ts:412` marks it as *"#347 review R1"* — it was added during review of #347 to `loadOrMintMaterial`, and it was not carried across to the other material resolver or to the two documents that describe the same edit. The result is that the headline benefit of #347 — *adding an address costs a paired client nothing* — is contradicted by the product's own messages on the path an operator actually uses.

**1. `packages/cli/src/actana-cli.ts:637-645` (with `packages/cli/src/actana-setup.ts:303-326`)**

`resolveMaterial` calls `checkServerCertHost(existing, opts.publicHosts, …)`, which returns `moved` for *any* non-identical list, and there is no `widened` branch — `MaterialOutcome` is `minted | reused | reissued | re-minted`. So a Core set up as `--public-host 10.0.0.5` that is re-run as `--public-host 10.0.0.5,core.lan` prints, at `actana-setup.ts:318`:

> `Public host changed to 10.0.0.5,core.lan — re-issuing this Core's server certificate from its own CA.`

and then, at `actana-cli.ts:641-644`:

> `This Core's pairing credentials are unchanged — a paired Panel still trusts this Core, but it is dialling the address it paired with. Point it at 10.0.0.5, or run \`actana pair new\` here and pair it again.`

The Panel is dialling `10.0.0.5`. That address is unchanged, still primary, still on the certificate. "Point it at 10.0.0.5" is a no-op instruction, and "pair it again" is the exact cost ADR 0038 says the record exists to remove — charged for a change that removed it. The daemon gets this right (`core-entry.ts:585-593` prints what was added and advises nothing); `actana setup` does not, and on metal `setup` is the *only* message the operator sees, because the daemon's next boot then reads `covered`.

`widenedPublicHosts` is already exported from `@actana/shared/public-hosts` and already imported by `core-first-run.ts`; the fix is a `widened` outcome in `resolveMaterial` and a fourth branch at `actana-cli.ts:637`.

**2. `docs/images/core.md:83-86`**

The new "More than one address" section (lines 63-81) tells the reader that adding an address keeps every client paired. The very next paragraph tells them the opposite for the same edit:

> `Changing ACTANA_PUBLIC_HOST re-signs the server certificate for the new address … a Panel paired before the change still trusts this Core — but it is still dialling the old address, so point it at the new one.`

Two adjacent paragraphs, opposite advice. The paragraph needs to distinguish adding from replacing.

**3. `deploy/README.md:86-90` and `deploy/README.md:97-98`**

ADR 0038's own Consequences section says: *"The three-things-at-once sentence in `deploy/README.md` is now two-and-a-half."* That edit was not made. Line 86 still reads "That one value is three things at once", and item 3 — "the **endpoint every pairing hands back** to the client that redeems a code" — is now false: it is the *primary*, and `pair new --public-host` may choose another. Item 2 is now "the SANs", plural.

Worse, lines 97-98 still say:

> `Changing it after pairing means re-pairing: the old certificate does not cover the new name.`

…ten lines above the new subsection that says the opposite ("clients paired before the change stay paired, which is the re-pair that changing the one value used to force"). The page contradicts itself, and it contradicts a consequence the ADR wrote down for itself.

---

### Non-blocking notes

- **`INSTALL.md:287-289`** — "Re-running `setup` is safe: it upgrades in place, keeps this Core's identity … Paired clients stay paired" is now unconditionally false in the new `re-minted` path (`actana-setup.ts:275-297`), where setup replaces the identity and locks out every paired client. Under `--yes` — which `INSTALL.md:262` recommends for unattended installs — `if (opts.interactive && !opts.assumeYes)` skips the confirmation entirely. The behaviour is defensible (the alternative is a daemon that will not boot, and it is announced on stdout); the doc should stop promising the opposite.
- **`packages/cli/src/actana-setup.ts:198-206`** — the `SetupResult.materialOutcome` JSDoc lists only `minted | reused | reissued`. `re-minted` is missing, though it is documented on the `MaterialOutcome` type just above.
- **`packages/cli/src/actana-uninstall.ts:103`** — `removed.push(legacy)` pushes a service *label* (`com.actana.harness`) into a list whose other entries are filesystem paths.
- **`core-boot-refusals.ts:25` vs `actana-pair.ts:295`** — two different `LOOPBACK_HOSTS` constants, in two packages, with different contents and different meanings. Both are carefully commented, but a reader grepping the name gets two answers.
- **`INSTALL.md:778`** — "Re-run setup and re-pair with the new token it prints" is stale since #287 (setup emits no credential). Pre-existing, not introduced here, but this is now also the place to mention that *adding* the new address is the cheaper fix.

---

### Verified clean

- **Train mechanics.** All seven stamps read `0.4.2` — `node scripts/assert-version-agreement.mjs --expected 0.4.2` is green over the six manifests and `install.sh`'s `LINE="0.4.2"` (line 80). Branch is 3 ahead / 0 behind `beta/0.4.2` (merge-base `39b77ba`), so a clean fast-forward candidate. Under `.github/` only `ci.yml` changes, and only to add one step to the already-required `Lint` job — nothing touches `main`, a release line, a tag, or a workflow dispatch. All 24 checks green, `Train rules` and `Promotion gate` included.
- **#346.** The ellipsis is gone (`install.sh:537`), `scripts/install-sh-ascii-guard.sh` passes on the current file, and the guard rides in `Lint`, which `docs/rulesets/{main,beta,release}.json` all require — so it gates from its first run rather than sitting as an ungated context.
- **ADR coherence otherwise.** 0038's amendment notes are present under 0016 D15 and D18 and under 0034 D2, the `docs/README.md` table row is added, and the specific claims I spot-checked hold at the code: D4's loopback pair is appended on both the mint and re-issue paths (`core-cert-material.ts:serverSanAltNames`), D5's "nothing is written on the refusal path" holds (`actana-pair.ts:234-236` precedes `openStore`), and D7's lying-`Host`-header case exists at `core-pairing-redeem.test.ts:679`.
- **Nothing dead or left behind.** No unused exports from the new modules, no TODO/FIXME/`.only`/`debugger` in the added lines, no duplicated helper between the three tickets beyond the `LOOPBACK_HOSTS` note above.
- **Everything passes locally**, not just in CI: core 903, cli 1091, shared 509, sdk 292, panel 1491, scripts 755, install-sh guard 23. `eslint packages scripts` — 0 errors (46 pre-existing warnings, none in files this PR touches).

The security work is right and I would not hold the train for the notes. I am holding it for the three blocking items, because "adding an address costs you nothing" is the promise this ticket makes, and right now `actana setup` and two operator documents tell the reader otherwise.

qcentic-dev:
## Pushed — the widening distinction now reaches the operator

`f02ae6f`, `42e0595`, `b2ae557` on `feat/0.4.2-install-and-identity`, head `b2ae557`.

You diagnosed this exactly right, and the framing — one root cause in three places — is what made it quick to fix:
ADR 0038 D3a landed in `loadOrMintMaterial` during #347's own review and was carried nowhere else, so the resolver an
operator on metal actually reads, and the two documents describing the same edit, all still charged for a change that
costs nothing.

### 1. `actana setup` — `actana-setup.ts:303-326` + `actana-cli.ts:637-645`

`resolveMaterial` now asks `widenedPublicHosts` — already exported from `@actana/shared/public-hosts`, already used by the
daemon — and returns a `widened` outcome carrying the added hosts. `cmdSetup` has the fourth branch. On the case you
named, `--public-host 10.0.0.5` re-run as `--public-host 10.0.0.5,core.lan`:

```
Adding core.lan to this Core's addresses — re-issuing its server certificate from its own CA to cover it.
Every address this Core already answered to is still covered.
...
This Core now answers to 10.0.0.5, core.lan — core.lan joined the addresses it already had, and its server
certificate was re-issued from its own CA to cover it. Every paired client keeps working and none has to be
re-paired. To pair a client to one of the new addresses, run `actana pair new --public-host core.lan` here.
```

Nothing is advised, because there is nothing to do — the daemon's wording, for the same reason.

One detail worth flagging: the widening is computed against the same list `checkServerCertHost` compared — the material's
`serverHosts`, falling back to the config setup wrote beside it — so the coverage test and the widening test cannot
disagree about what "before" was. Material predating the SAN record widens correctly rather than falling into `moved`.

### 2 & 3. `docs/images/core.md` and `deploy/README.md`

Both now distinguish adding from replacing. `core.md`'s paragraph became two bullets — adding keeps every client paired;
replacing or removing means re-pairing whatever was paired to the address you took away — plus a line on reordering, since
the primary is the default endpoint. `deploy/README.md` is "two-and-a-half things at once" in the words ADR 0038's
Consequences section used, item 2 is the SANs and item 3 is the primary "unless that code chose otherwise", and the
"changing it after pairing means re-pairing" line ten lines above the new subsection is now scoped to replacing.

### The regression pin you asked for

This is the part I took most seriously: the distinction was tested in the daemon and untested here, which is exactly why
it was missed. It is now pinned at both levels.

At the resolver (`actana-setup.test.ts`) — the outcome, the added list, that the CA and `coreId` survive, that both
addresses are on the SAN, and that the message says neither "Public host changed" nor "pair it again". And both negative
cases, because the distinction has to cut both ways or it is just a way of never reporting a move: a dropped address with
another added is still `reissued`, and so is a reordered list. Plus the predates-the-record case above.

End to end (`actana-machine-cli.test.ts`) — `actana setup --public-host 10.0.0.5` then `10.0.0.5,core.lan`, asserting the
output says none has to be re-paired and does *not* say "dialling the address it paired with", with the move case asserted
beside it so the two cannot drift into each other.

### Non-blocking notes — all four fixed

- **`SetupResult.materialOutcome` JSDoc** — now lists `widened` and `re-minted` with the rest.
- **`actana-uninstall.ts:103`** — the label no longer goes into `removed` (documented as paths, rendered as a column of
  directories). It has its own `removedLegacyService` field; the operator-facing line was already separate and is
  unchanged.
- **`INSTALL.md:287-289`** — stops promising paired clients always survive a re-run, names the `re-minted` exception, and
  says plainly that `--yes` takes that answer without asking. **`INSTALL.md:778`** no longer promises a token `setup` has
  not printed for two releases, and now names adding the new address beside the old as the cheaper fix.
- **The two `LOOPBACK_HOSTS`** — commented rather than unified, and I want to say why, since you left it to me. They
  answer different questions and unifying them would break one: `actana-pair.ts`'s is the pair D4 puts on the certificate,
  a record of what was *signed*, so widening it would claim coverage the certificate does not carry;
  `core-boot-refusals.ts`'s decides whether a bind address reaches off this machine, so it is as wide as possible and
  fails closed. Each now names the other and says that.

### Not done, deliberately

The notes in the approving review that belong to other files — `warnAboutLegacyService`'s wording, the bare `actana setup`
in three messages, `LEGACY_UNIT_NAME`'s stale doc, the orphaned JSDoc in `core-smoke.mjs`, `actana-pair.ts:347`'s
container-only variable name, and the string-list/endpoint-building duplication — are follow-up ticket material by that
reviewer's own framing, and I have not widened a gate branch to take them.

`pnpm test` — all 6 stages pass. `pnpm -r typecheck` clean, `eslint` 0 errors. `install-sh-ascii-guard` passes and
`assert-version-agreement --expected 0.4.2` is green over all seven stamps.

mrfqcentic (changes requested):
## Gate review, supplement to review [#5053218305](https://github.com/actana/control/pull/353#pullrequestreview-5053218305)

Same verdict, four additional findings. A deeper automated pass over the diff returned after I had already submitted; it raised four items my first pass missed, and I have verified every one of them at the enforcing lines. **One of them is more consequential than anything in my original blocking list, and it is the sharpest #347 × #348 interaction on this branch.** Nothing in my earlier review is retracted.

---

### C1 — `packages/cli/src/actana-cli.ts:553-560` — a bare `actana setup` silently collapses a multi-address Core to one guessed address and unpairs every client

`cmdSetup` defaults `--public-host` to `choosePublicHost(deps.networkInterfaces, deps.hostname)` and **never consults the recorded config** — even though `actana.json`'s new `publicHosts` list is the only record of the operator's multi-address answer.

On a Core installed with `--public-host core,10.0.0.5`, a bare `actana setup`:

1. resolves `publicHosts = ["<first routable IPv4>"]`;
2. `resolveMaterial` → `checkServerCertHost(existing, ["<guess>"], ["core","10.0.0.5"])` → `moved` (`actana-setup.ts:304`);
3. `reissueServerCert` mints a **single-SAN** certificate for the guess (`actana-setup.ts:324`);
4. `writeActanaConfig` overwrites `publicHost` **and** `publicHosts` from the guess (`actana-setup.ts:452-462`);
5. `service.install` writes `AC_CORE_PUBLIC_HOST=<guess>` into the unit (`actana-setup.ts:472`).

Every client paired to `core` now fails TLS hostname verification, and `actana pair new --public-host core` is refused, because `material.serverHosts` no longer contains it.

**Why this is a train-level defect and not a pre-existing one.** The single-address version of this hazard predates the branch. What is new is the traffic: **#348 makes a bare `actana setup` the advertised remedy in five separate places** — the daemon's `AC_HARNESS_*` refusal (`core-boot-refusals.ts:70-77`), the `Legacy agent` status row (`actana-status.ts:184`), `actana update`'s no-service message (`actana-update.ts:216-222`), `actana place`'s warning (`actana-cli.ts:849-853`), and the unusable-material path (`core-material-store.ts:REMEDY`) — and **#347 raised the cost of taking that advice** from one address to the whole list plus every `pair new --public-host` selection. #348 drove traffic onto the path that #347 made expensive. Neither ticket's own review could have seen this; it is exactly what this gate is for.

Note the asymmetry with the daemon, which already refuses to do this: `core-first-run.ts`'s `publicHostDeclared` gate exists precisely so that *"a guess that rewrote the SAN would take a working Core off its own address"*. `setup` treats its guessed default as declared.

Defaulting to `configPublicHosts(readActanaConfig(opts.layout.configDir))` when a config exists closes it, and reuses a function this PR already added.

### C2 — `packages/cli/src/actana-update.ts:209` with `packages/cli/src/actana-service.ts:318-333` — a regression: `update` can skip a restart it previously performed

`runActanaUpdate` now gates the restart on `observed.name === service.name`. But the systemd `observe()` asks systemd (`stateOf(UNIT_NAME)`) **only when a legacy unit is present**, and its comment justifies that narrowing entirely in terms of `actana place`:

> *Asked only when there is a legacy unit to be confused with, and that is deliberate: `actana place` calls this, and placement asks the init system nothing on the ordinary machine.*

That reasoning was sound when `observe()` had two consumers. #348 gave it a third whose correctness depends on the probe being skipped. On a machine whose unit file was deleted by hand while systemd still holds the unit loaded — the case the launchd side covers via `isLoaded()` (`actana-service.ts:568`), and the case `state()` keeps its own fallback for (`actana-service.ts:352-357`) — `observe()` returns `name: null`. So `update` swaps the tree and `current`, skips `verb("restart")`, prints *"this machine has no auto-start service"*, and exits 1, **leaving the old daemon running out of the new tree**. The pre-#348 code called `verb("restart")` unconditionally and succeeded.

Either ask `stateOf(UNIT_NAME)` unconditionally in `observe()`, or fall back to `service.state()` before concluding there is nothing to restart.

### C3 — `packages/cli/src/actana-pair.ts:347` — the refusal names a variable that does not exist on the machine reading it

```
deps.err(`Configured (${CONTAINER_PUBLIC_HOST_ENV}): ${formatPublicHosts(configured)}`);
```

`ACTANA_PUBLIC_HOST` is hardcoded. On bare metal that variable exists nowhere — the list came from `--public-host` at setup and lives in `actana.json` and in the unit's `AC_CORE_PUBLIC_HOST`. An operator who mistypes `actana pair new --public-host lan` on metal is sent to grep for an unset variable.

This is the exact failure `parsePublicHosts(raw, varName)` was parameterised to prevent — *"A message that named the wrong one would send them to the wrong file"* — and which `core-entry.ts:476` gets right two files away with `containerMode ? CONTAINER_PUBLIC_HOST_ENV : "AC_CORE_PUBLIC_HOST"`. Same class as B1–B3 in my first review: the ticket knows the rule and one call site does not follow it.

### C4 — `packages/cli/src/actana-uninstall.ts:103` — supersedes note N3 in my first review; it is not cosmetic

I filed the service-label-in-a-path-list as a tidiness nit. That was wrong. `removed` is also the *"did we do anything?"* signal — `actana-cli.ts:1337` prints *"There was no Core installed for this user."* only when `removed.length === 0 && kept.length === 0`.

`actana uninstall` deliberately runs without `requireInstall`, so on a machine carrying only the pre-rename agent and no install, `removed` becomes `["com.actana.harness"]`. Line 129 then prints:

> `Removed the com.actana.core service and the Core install.`

naming a service that was never on the machine and an install that never existed — **and** the truthful "no Core installed" line is suppressed. Two false statements, on the #348 cleanup path, on exactly the machine #348 is about. Track the legacy removal in its own flag rather than in `removed`.

---

### Still true from the first review

The security property holds — no TLS ⟹ `!remoteMode` ⟹ the exposure refusal passed ⟹ loopback bind, with both refusals independent and failing closed. The independent pass agreed and added confirmation I had not obtained: every Core spawn site in the tree (`deploy/core.Dockerfile`, `actana-setup.ts`, `cmdDaemon`, the smoke harness) sets `AC_CORE_REMOTE=1` when binding non-loopback, so `plaintextExposureRefusal` refuses nothing that works today; `AC_HARNESS_*` has no other reader; `com.actana.harness` and `mission-control-harness-ca` match what git history shows the pre-rename build actually wrote; and the `serverHost` → `serverHosts` migration has no stragglers. `tsc --noEmit` is clean across `cli`/`core`/`shared`.

Train mechanics remain clean: 7/7 stamps at `0.4.2`, 3 ahead / 0 behind `beta/0.4.2`, only `ci.yml` under `.github/`, nothing touching `main`, a release line, or a tag.

**C1 is the one I would fix before this train moves.** C2 is a regression against the previous behaviour of `update`. C3 and C4 are message-truthfulness defects in the same family as B1–B3 — which is now seven places on one branch where the product tells an operator something that is not so, and the reason this gate exists.

qcentic-dev:
## Pushed — C1–C4 fixed, head `1a231b1`

`2341216`, `80fa135`, `596f1ee`, `1a231b1`. Each has a regression test at the machine shape you named.

### C1 — a bare `actana setup` collapsing a multi-address Core

You are right that this is the most consequential item here, and right about why: the hazard is old, the *traffic* is
new. #348 made a bare `actana setup` the advertised remedy in five places and #347 raised what taking that advice costs
from one address to the whole list — so the unlucky path became the common one.

`cmdSetup` now defaults `--public-host` to `configPublicHosts(readActanaConfig(...))` and falls back to
`choosePublicHost` only when there is no config, which is a first install. An explicit `--public-host` is a declaration
and stays authoritative. Your framing of the asymmetry is what settled the shape: the daemon's `publicHostDeclared` gate
exists precisely so a guess cannot re-sign a SAN and take a working Core off its own address, and `setup` was treating
its guessed default as declared. It no longer does, and it says so when it keeps the recorded list, because that is a
decision made on the operator's behalf:

```
Keeping this Core's recorded addresses: core, 10.0.0.5. Pass --public-host to change them.
```

Four tests in `actana-machine-cli.test.ts`: the recorded addresses surviving a bare re-run — asserted on the certificate
SAN, `material.serverHosts`, `actana.json` **and** the unit's `AC_CORE_PUBLIC_HOST`; `pair new --public-host 10.0.0.5`
still succeeding after that re-run, which is the membership check a collapsed certificate would have broken; an explicit
flag still changing the addresses; and a first-time install still guessing.

I have also struck item 2 from #351 and said there what was fixed and by what. **One caveat I want visible rather than
buried:** `--port` and `--label` have the same shape and I did *not* change them. Neither can unpair a client or
invalidate a certificate, so widening a gate branch to cover them looked wrong — that remainder is recorded on #351
rather than left looking closed.

### C2 — `update` skipping a restart it used to perform

Confirmed, and fixed at the consumer rather than by widening the probe: `actana place`'s "asks the init system nothing"
is a real contract with its own test, and putting a `systemctl` call in it to fix `update` would trade one regression for
another. So when `observe()` found nothing **and there is no legacy service in play**, `runActanaUpdate` asks `state()`
directly. That condition is exact rather than convenient — `state()`'s own fallback to the legacy unit requires the
legacy unit to be present, so in that branch a non-null answer can only be about the current service. The narrowing in
`observe()` now says out loud that a caller which must not miss a loaded-but-file-less unit cannot rely on it alone.

Tested at that exact shape twice: at the unit level (restart happens; and the negative case, where a loaded *legacy* unit
must not be mistaken for a current one), and end to end through the real systemd manager with the unit file removed and
systemd still reporting it loaded.

### C3 — the refusal naming a variable that does not exist

Fixed the way `core-entry.ts` does it, keyed on container mode: `ACTANA_PUBLIC_HOST` in a container, `actana setup
--public-host` on metal — in both cases the thing the operator edits to change the answer. Both shapes are pinned,
because naming one correctly by naming the other wrongly is exactly how this arose.

### C4 — you were right that it was not cosmetic

Thank you for re-filing it. I had treated N3 as tidiness and moved the label out of `removed`; I had not seen that
`removed` is also the "did this run do anything?" signal, so the truthful line was being suppressed while two false ones
printed. Both are fixed: the summary names only the halves that were actually there — the service only when one was, the
install only when something of it was removed — and the caller's no-install line now reads

> `There was no Core installed for this user — only com.actana.harness, left by an install from before the rename, and
> that is now removed.`

so it neither claims an install nor reads as though the command did nothing. Pinned at the resolver and end to end
through `actana uninstall` on a machine carrying nothing but the old unit.

---

Nothing from either review is outstanding on my side. `pnpm test` — all 6 stages pass; `pnpm -r typecheck` clean;
`eslint` 0 errors; `install-sh-ascii-guard` and `assert-version-agreement --expected 0.4.2` both green, so all seven
stamps still agree.

qcentic-adm (approved):
## Gate review, round 2 — head `1a231b1` (second identity)

**Verdict: APPROVE.** My earlier approval (5053160612, against `aa399b7`) was dismissed by the new commits; this is a fresh review of the seven commits since, verified at the enforcing lines rather than taken from the moderator's or the author's word.

### The moderator's seven findings — each checked at the line

- **C1 (bare `setup` collapsing a multi-address Core) — fixed.** `actana-cli.ts:571-579`: the `--public-host` default is `configPublicHosts(readActanaConfig(...))` joined with `PUBLIC_HOST_SEPARATOR`, falling back to `choosePublicHost` only when there is no config. An explicit flag wins (`stringFlag` reads the flag first). `actana-machine-cli.test.ts:633-690` asserts on the certificate SAN (`DNS:core`, `IP Address:10.0.0.5`), `material.serverHosts`, `actana.json` `publicHosts`/`publicHost`, the unit's `AC_CORE_PUBLIC_HOST`, that `pair new --public-host 10.0.0.5` still succeeds after a bare re-run, that an explicit flag still changes the list, and that a first-time install still guesses. Also improves the #348 story: a bare `actana setup` on the pre-rename machine now keeps `actana.json`'s `publicHost` rather than moving the Core to a guessed IP.
- **C2 (`update` skipping a restart) — fixed.** `actana-update.ts:222-224`: `hasCurrentService = observed.name === service.name || (observed.legacyName === null && service.state() !== null)`. The guard is exact: `state()`'s own legacy fallback needs a legacy unit present, so in the `legacyName === null` branch a non-null answer can only be the current unit. `actana-update.test.ts:252` (restarts a unit systemd still holds after the file was deleted) and `:276` (a loaded *legacy* unit is not mistaken for a current one). `observe()`'s narrowing is now documented at `actana-service.ts:332-337`, and `place` still asks the init system nothing on an ordinary machine.
- **C3 (`pair new` naming `ACTANA_PUBLIC_HOST` on metal) — fixed.** `actana-pair.ts:366-369`, keyed on `inContainer(deps.env)` exactly as `core-entry.ts` does.
- **C4 (`uninstall` claiming a service and install that were never there) — fixed.** `actana-uninstall.ts:117`, `:147-151` (`hadService ? … : "Removed the Core install."`), `removedLegacyService` field; `actana-cli.ts:1386-1394` prints the truthful legacy-only line. Tests `actana-uninstall.test.ts:139,204`.
- **B1 (`setup` charging a re-pair for a widening) — fixed.** `actana-setup.ts:349-372`: `widenedPublicHosts(signedFor, opts.publicHosts)` computed against the *same* list `checkServerCertHost` compared (`serverHosts`, else the config), so coverage and widening cannot disagree about "before"; `widened` outcome with `addedHosts`; fourth branch at `actana-cli.ts:676-693` advises nothing. Negative cases (dropped+added, reordered) are still `reissued`.
- **B2 / B3 (docs contradicting the widening) — fixed.** `docs/images/core.md` and `deploy/README.md` now distinguish adding from replacing/removing, "two-and-a-half things" per ADR 0038 §C; `INSTALL.md` names the `re-minted` exception and `--yes`, and drops the phantom "new token".

### The five gate questions, re-verified at `1a231b1`

1. Bare `setup` keeps a recorded multi-address install (SANs, `serverHosts`, config list, unit env) — yes; explicit `--public-host` authoritative — yes; first-time install guesses — yes.
2. `update` restarts whenever there is a current unit, file or loaded-only, and only leaves the legacy agent when it is the *only* service — yes; `removeLegacyUnit` still runs before the restart.
3. Security property: `packages/core/src/core-entry.ts` is byte-identical to the approved `aa399b7` (0 changed lines); `core-boot-refusals.ts` changed only in a comment and `LOOPBACK_HOSTS` is the same set. Order legacy-env → exposure → remote/material still holds; the two `listen()` sites still bind `opts.host`; the image still bakes `AC_CORE_REMOTE=1`. No combination serves plaintext core-link off-loopback.
4. Operator story on a pre-rename machine: `install.sh` → `place` (warns, names `setup`) → `update` (legacy-only: leaves it, truthful message, exit 1) → `setup` (boots out the legacy agent first, keeps the recorded address, keeps foreign-CA material, writes the new unit) → `status` (observed service, no legacy row, exit 0) → `pair new` (works, `--public-host` validates against backfilled `serverHosts`). `uninstall` on a legacy-only machine now tells the truth too. No circle.
5. Train mechanics: 7/7 stamps `0.4.2` (six manifests + `install.sh:80`), `Train rules` pass; 10 ahead / 0 behind `beta/0.4.2`, merge-base = train tip (clean fast-forward); only `.github/workflows/ci.yml` (the Lint guard step) under `.github/`; nothing touches `main`, a release line, a tag, or a dispatch. 20/24 checks pass, 4 train-only jobs skipped by design. Local: cli 442/442 across the seven touched suites.

### Non-blocking, for the follow-up tickets

- **`actana-cli.ts:532` / `:604`** — `--port` and `--host` still default to `DEFAULT_PORT` / `0.0.0.0` on a bare re-run, not to the recorded config. Same shape as C1 but narrower: a bare `actana setup` on a Core installed with `--port 9443` rewrites the unit and the endpoint to 8443, which *does* break every client holding `wss://host:9443`. The #348 machine used the default port, so it does not affect the train's story; the author has recorded it on #351 — keep it there.
- **`actana-cli.ts:843-851`** — `warnAboutLegacyService`'s "it is now starting the version just placed" is still ahead of the facts while the old daemon runs its old inode; carried from round 1, follow-up.
- **`actana-systemd.ts:25-32`** — `LEGACY_UNIT_NAME`'s "no release ever shipped under this name" doc; carried, follow-up.

Nothing the four review-pressure commits introduced regresses a path I checked.

mrfqcentic (approved):
## Gate review R3 — **APPROVE**

Re-reviewed at head `1a231b1`, against my [first review](https://github.com/actana/control/pull/353#pullrequestreview-5053218305) (`aa399b7`) and its [supplement](https://github.com/actana/control/pull/353#pullrequestreview-5053358377) (`b2ae557`). **All seven findings are fixed, verified at the enforcing line, and each of the four supplement items is pinned by a test that genuinely fails without its fix** — I checked that last part by reverting each fix in a throwaway checkout and watching the tests go red, rather than taking the assertions' word for it.

### The three blocking items from the first review

**B1 — the widening distinction now reaches the operator.** `MaterialOutcome` carries `widened` (`actana-setup.ts:197`), `resolveMaterial` resolves it at `actana-setup.ts:361-374` against `signedFor = existing.serverHosts || recorded` — the same list `checkServerCertHost` compared, so the coverage test and the widening test cannot disagree about "before" — and `cmdSetup` has the fourth branch at `actana-cli.ts:676-693`. On the case I named, `10.0.0.5` re-run as `10.0.0.5,core.lan`, nothing is advised because there is nothing to do.

**The opposite direction is unharmed, which is the half that mattered.** The `reissued` branch (`actana-cli.ts:667-675`) still says "it is dialling the address it paired with … or run `actana pair new`", and a genuine move still reaches it: `widenedPublicHosts` (`public-hosts.ts:172-180`) returns `null` unless every previous host survives, so a dropped-and-added list falls through to `moved`, and a same-set reorder returns `null` on `added.length === 0` — ADR 0038 D3a's "same set in a different order is a move". Both are pinned negatively at `actana-setup.test.ts` ("still calls a dropped address a move, even when another is added", "calls a reordered list a move"), which is what stops the distinction from becoming a way of never reporting a move.

**B2 — `docs/images/core.md:83-101`.** The self-contradicting paragraph is now two bullets that separate adding from replacing, plus a line on reordering. **B3 — `deploy/README.md:86-107`.** "two-and-a-half things at once" in the ADR's own words, item 2 is the SANs, item 3 is the primary "unless that code chose another of them", and the "changing it after pairing means re-pairing" line is scoped to replacing.

### The four supplement items

**C1 — the sharpest one, and it is closed in all three directions** (`actana-cli.ts:570-597`). `recordedHosts = configPublicHosts(readActanaConfig(...))` is the default and `choosePublicHost` is the fallback only when there is no config.

- *A recorded multi-address install survives a bare setup.* Pinned end to end at `actana-machine-cli.test.ts:633` on all four artefacts — the certificate SAN, `material.serverHosts`, `actana.json`, and the unit's `AC_CORE_PUBLIC_HOST` — plus the "Keeping this Core's recorded addresses" line. **Reverting the default to the guess makes it fail** with `DNS:core` gone and a single-SAN cert for `10.0.0.5`, which is exactly the collapse I described.
- *An explicit `--public-host` stays authoritative and can still narrow.* `stringFlag` takes the flag over the default and the "Keeping" line is gated on `parsed.values["public-host"] === undefined`; pinned, including that nothing is reported as "kept".
- *A first-time setup still guesses.* `readActanaConfig` returns `null`, so the fallback fires; pinned. `configPublicHosts` also reads a pre-#347 config's single `publicHost` as a one-entry list, so a machine upgrading from 0.4.1 does not re-issue on its next setup.

This also retires the note I filed on the dismissed review, that the five remedy messages naming a bare `actana setup` would re-address a Core. They no longer can.

**C2 — `update` cannot skip a restart it used to perform** (`actana-update.ts:221-224`). Fixed at the consumer, which is the right side: putting a `systemctl` call in `observe()` would break `actana place`'s "touches nothing running" contract. On the machine I named — unit file deleted by hand, systemd still holding it loaded, no legacy unit — `observe()` returns `{name: null, legacyName: null}`, `state()` → `stateOf(UNIT_NAME)` returns a loaded unit, and `verb("restart")` runs. The condition is exact rather than convenient: when `legacyName` is non-null `observe()` has already probed `stateOf(UNIT_NAME)` itself (`actana-service.ts:338`), and when it is null `state()`'s own legacy fallback (`:361`) cannot fire, so a non-null answer can only be about the current service. launchd was never affected — its `observe()` asks `isLoaded()` (`:568`). Pinned twice, and **both go red against the old one-line gate**: the unit-level pair at `actana-update.test.ts:252/275` (including the negative case, a loaded *legacy* unit that must not be mistaken for a current one) and end to end through the real systemd manager at `actana-machine-cli.test.ts:1149`. **The old daemon cannot be left running out of the new tree.**

**C3 — the refusal names the variable that exists on the machine reading it** (`actana-pair.ts:366-369`), keyed on `inContainer(deps.env)` exactly as `core-entry.ts:476` does. Metal gets `actana setup --public-host` — better than the unit's generated `AC_CORE_PUBLIC_HOST`, since it is the thing the operator actually edits; a container gets `ACTANA_PUBLIC_HOST`. Both shapes pinned, and the metal assertion fails against the hardcoded version.

**C4 — the uninstall defect** (`actana-uninstall.ts:100-165`, `actana-cli.ts:1385-1396`). The label is out of `removed` and in `removedLegacyService`, the summary is gated on `hadService`, and the caller's no-install line now says both true things at once. On the legacy-only machine `removed` is `[]`, so neither false statement can print and the truthful line is no longer suppressed. Pinned at the resolver (both new cases) and end to end at `actana-machine-cli.test.ts:1292`; all three fail against the old code.

### Nothing regressed on what was already clean

`packages/core/src/core-entry.ts` and the whole of `packages/shared/` are **byte-identical to `aa399b7`**, so the security property is untouched: `legacyEnvRefusal` at `:156` before any value is read, `plaintextExposureRefusal` at `:181` independent of it, `serverOpts.tls` at `:565` unconditional inside `if (remoteMode)` at `:447` with every earlier exit a `process.exit(1)` — no TLS ⟹ `!remoteMode` ⟹ the exposure refusal passed ⟹ loopback bind. The only change under `packages/core` on this round is a comment on `LOOPBACK_HOSTS`. The #347/#348 call-site merge is likewise unchanged.

The operator's recovery loop still closes, and `update`'s new fallback does not perturb it: on a legacy-only machine `observe().name` is the legacy label and `legacyName` is non-null, so the second disjunct cannot fire, the agent is left in place, and the message still names `actana setup` (`actana-update.ts:236-245`). `place` still warns (`actana-cli.ts:896`), `setup` still boots the agent out first, before the tree is swapped (`actana-setup.ts:492`), and `status` still prints the `Legacy agent` row naming `setup`. On a healthy machine the gate short-circuits on `observed.name === service.name`, so no extra `systemctl` call on the common path.

**Train mechanics.** All seven stamps read `0.4.2` — six manifests and `install.sh:80` `LINE="0.4.2"`. Branch is **10 ahead / 0 behind `beta/0.4.2`**, still a clean fast-forward candidate. Under `.github/` only `ci.yml`, still only the one `install.sh` guard step added to the already-required `Lint` job; nothing touches `main`, a release line, a tag, or a workflow dispatch. All 20 checks green on `1a231b1` (the four skipped are the train image jobs, which do not run on a pull request). Locally: cli 1109, core 903, shared 509 — cli up 18 from the new pins, the other two unmoved. `mergeStateStatus` is `BLOCKED` only on my own outstanding review; this approval clears it.

All five non-blocking notes from the first review are also fixed — the `SetupResult` JSDoc, `INSTALL.md:285-299` and `:781-796`, the uninstall list, and the two `LOOPBACK_HOSTS`, which are commented rather than unified. That call is right, and for the reason given: one records what was signed and the other decides whether a bind reaches off-machine, so widening either to match the other would break it.

---

### Non-blocking — follow-up ticket material, none of it holds the train

1. **`packages/cli/src/__tests__/actana-machine-cli.test.ts:655`** — this test does not discriminate. The fixture's only routable IPv4 is `10.0.0.5` (`machine-fixture.ts:140`), which is the address the test pairs to, so the guess would have picked it anyway: I reverted C1 and **this test still passed** while its neighbour at `:633` failed. C1 is genuinely pinned by `:633`; this one is a guard that cannot fire. Asserting `pair new --public-host core` — the address a collapse actually drops — would make it load-bearing.
2. **`packages/shared/src/public-hosts.ts:172-180`** — `widenedPublicHosts` is set-containment plus a non-empty addition, so `["10.0.0.5"] → ["core.lan", "10.0.0.5"]` is a widening even though the primary, and therefore the default endpoint for new pairings, just changed. That is what ADR 0038 D3a says, but it is also the exact reason D3a gives for calling a *pure* reorder a move — "the primary, and therefore the default endpoint, is not what it was". Nothing breaks for a paired client and the message prints the full list in order, so the operator can see it; `docs/images/core.md:99-101` documents the consequence. Either a clause in the `widened` message when the primary moved, or a sentence in D3a saying the primary is deliberately not part of the widening test.
3. **`packages/cli/src/actana-uninstall.ts:147-152`** — same family as C4, much more marginal: on a machine with a unit file but no install tree (an interrupted uninstall), `removed` is `[servicePath]` and `hadService` is true, so the summary still claims "and the Core install". Gating the two halves separately, as the caller's line now does, would close it.
4. **`packages/cli/src/actana-uninstall.ts:117` with `actana-cli.ts:1392`** — on the legacy-only machine the removal is announced twice, once by the resolver and once by the caller's combined line. Both are true; the second says it better.

The three blocking items and the four supplement items are all closed at the enforcing line, each with a regression test at the machine shape it describes that I confirmed fails without the fix. **Approving.**

PR #354, opened 2026-08-28
chore(release): promote 0.4.2

qcentic-adm:
The promotion gate for the **0.4.2 train**. Open and unmerged on purpose: this pull request is where the train's checks hang and it carries the head SHA a later promotion verifies against. `Promotion gate` is red here permanently and by design (the guard from #264) — it is not a failure and nothing on this pull request is meant to turn it green. Promotion, if and when it happens, is a separate owner decision made through `promote.yml`; nothing in this pull request dispatches it.

## What is on the train

`beta/0.4.2` is two commits ahead of `main` and zero behind, so it remains a clean fast-forward candidate:

- `39b77ba` — `chore(release): cut beta/0.4.2` — every manifest and the `install.sh` `LINE=` stamp read `0.4.2` (ADR 0023 D3, ADR 0036 D1).
- `8bed410` — `feat: install and identity — several addresses per Core, a bash 3.2-safe installer, and the pre-rename service removed (#353)` — the squash of the three tickets below, gated by #353 with two approvals.

The theme is the one the train is named for: getting onto this version, and staying the same Core once you are there.

### ⚠️ Security — a Core could serve unauthenticated plaintext core-link on a public interface (#348)

A pre-rename LaunchAgent sets `AC_HARNESS_*` variables the daemon no longer reads, so remote mode, the public host and the material file silently vanished while the port and the `0.0.0.0` bind survived — an unauthenticated loopback-mode `ws://` server on a public interface. The daemon now refuses to boot on any `AC_HARNESS_*` variable and on any non-loopback bind without `AC_CORE_REMOTE`; no compatibility fallback was added. Pinned by a property test over `(remote mode, bind host)` and by the release-tarball smoke, which boots the real daemon in both shapes and asserts a non-zero exit with nothing listening.

### The stale service and identity an upgrade left behind (#348)

- `actana setup` / `actana update` remove the pre-rename LaunchAgent on macOS as they always did on Linux; `update` only when a current service can take over.
- `actana place` warns when it is present; `actana uninstall` removes it.
- `actana status` reports the service the init system actually loaded, with an explicit `Legacy agent` row and a non-zero exit until it is gone.
- Identity material is checked at boot for what a TLS handshake needs. Pre-rename CA material is kept; only material that genuinely cannot be served is refused, and `actana setup` re-mints it.

### A Core can have several addresses, and a pairing code picks one (#347)

`ACTANA_PUBLIC_HOST` is now a comma-separated list. Every entry becomes a SAN, the first is primary, and `actana pair new` chooses which address a code hands back. A single value behaves exactly as before — no re-pairing. Adding an address re-signs from the CA already on disk.

### An installer that survives bash 3.2 under a UTF-8 locale (#346)

The multi-byte ellipsis that macOS bash 3.2 mis-parsed under a UTF-8 locale is gone, and the shape it appeared in is guarded.

## The beta cut this train has already published

`beta-release.yml` run [33203742186](https://github.com/actana/control/actions/runs/33203742186) — the first real exercise of that workflow — completed green on `8bed410`, all 14 jobs:

- Tag `v0.4.2-beta` (annotated, by `actana-release-train[bot]`) → `8bed41014a6169b6941745d8b51c94135b79230a`.
- Prerelease Release https://github.com/actana/control/releases/tag/v0.4.2-beta with `actana-core-0.4.2-beta-{linux-x64,linux-arm64,mac-arm64}.tar.gz`, `SHA256SUMS`, `actana-cli-0.4.2-beta.tgz` + `.sha256`, `install.sh`. `/releases/latest` still answers `v0.4.1`.
- `docker.io/actana/core:0.4.2-beta` @ `sha256:1c3db810419b638f4ea3361d30112a2abd9620537aaa71b8e5898bfb6f330d5d` (retag of `beta-0.4.2`).
- `docker.io/actana/panel:0.4.2-beta` @ `sha256:fe8d1a92665390b6ac098c0bc95392e0b6228b311bd0fc6eb893234bdf4b39f2` (retag of `beta-0.4.2`).

## Do not merge this pull request

Not by button, not by `gh`, not by fast-forward. `main` is promoted only by `promote.yml`, dispatched by the owner, which verifies the head SHA this pull request carries and cuts the next train. This pull request is the gate, not the door.


mrfqcentic (approved):
Beta 0.4.2 accepted: reviewed and human-tested against beta-0.4.2. Conventions is red on 8bed410 (127-char subject, limit 120) — accepted rather than rewriting the tested tip. Promotion gate red by design.

PR #362, opened 2026-08-31
docs: DOCKERHUB_TOKEN is Read, Write, Delete — the descriptions PATCH is a delete-scope call

### LINKED ISSUE #359: DOCKERHUB_TOKEN scope: descriptions PATCH needs Read, Write, Delete — docs and error message say Read & Write
Found during the 0.4.3 grilling (2026-08-31), confirmed empirically.

## Problem

The `descriptions` job in `housekeeping.yml` fails with `HTTP 403 — access denied: insufficient scope` on all four `PATCH /v2/repositories/…` calls (runs 32703975988, 33394103716), while image pushes with the same `DOCKERHUB_TOKEN` succeed. A PAT scoped **Read & Write** can push images but cannot write repository descriptions; the descriptions endpoint needs a personal access token scoped **Read, Write, Delete**. All four Docker Hub pages (`actana/panel`, `core`, `panel-dev`, `core-dev`) have been blank since launch because of this.

Two places assert the wrong scope:

- `docs/REPO_SETUP.md` §2: "one **personal** access token, `Read & Write`" (secrets table and the prose under it).
- `.github/workflows/housekeeping.yml`, the JWT-exchange failure message: "use a PERSONAL access token with Read & Write scope".

## What to do

1. Correct both to **Read, Write, Delete**, noting that image pushes alone are satisfied by Read & Write — the descriptions PATCH is the call that needs more.
2. Ops (not in this repo's diff): replace the `DOCKERHUB_TOKEN` secret value with an RWD-scoped PAT from the same admin account, then `gh workflow run housekeeping.yml -f chore=descriptions` and confirm all four pages render.

## Acceptance

- Docs and error message name the scope that actually works.
- A dispatched `chore=descriptions` run goes green and the four Docker Hub pages show the `docs/images/*.md` content.

qcentic-dev:
Closes #359 (docs half — the ops half was already done live, see below).

## What was wrong

A Docker Hub PAT scoped `Read & Write` pushes images fine and then fails every
`PATCH /v2/repositories/…` in the `descriptions` chore with `HTTP 403 — access
denied: insufficient scope` (runs 32703975988, 33394103716). Writing a
repository description on the Hub API is a **delete-scope** call, so the four
Docker Hub pages have been blank since launch. Two places asserted the wrong
scope; both are corrected here.

## Proof, not inference

On 2026-08-31 the scope of the **existing** `DOCKERHUB_TOKEN` was widened to
`Read, Write, Delete` in Docker Hub with the **secret value unchanged**, and the
next dispatched run — [33396982252](https://github.com/actana/control/actions/runs/33396982252)
— went green on the first try (`Docker Hub descriptions` ✓ in 9s) and filled all
four pages. The scope was the only variable.

## The two files

**`docs/REPO_SETUP.md` §2** — the secrets table, the subsection heading and the
"create it under Account settings → Personal access tokens" line now all say
`Read, Write, Delete`, plus what the old text left a reader to guess:

- a Docker Hub **organization access token is never the answer here** — the
  `/v2/users/login` JWT exchange rejects it however it is scoped, so nobody
  should be steered to an OAT;
- this **same secret drives the image pushes**, so it must never be swapped for
  a narrower token;
- a **fine-grained token**, if anyone reaches for one instead, needs the
  **Repository Edit** permission on all four repositories — Edit is what covers
  the description `PATCH`.

One neighbouring sentence in the same section moved with the facts: the
`DOCKERHUB_CLEANUP_TOKEN` subsection claimed delete "is a permission the push
token must never have", which was true only while the push token was scoped
narrower. It is restated as what still holds — the sweep is the only thing aimed
at a delete endpoint, and the allowlist in `scripts/lib/dev-tag-sweep.mjs` is
what bounds it. Leaving it would have made §2 contradict itself two screens
apart.

**`.github/workflows/housekeeping.yml`** — one string, the `descriptions` job's
JWT-exchange failure message, `Read & Write` → `Read, Write, Delete`. No steps,
no triggers, no logic.

## Scope and boundaries

- Docs only. Exactly two files. No workflow logic changed.
- `packages/cli` and `packages/panel` (parallel tickets) untouched.
- Base is `feat/0.4.3-onboarding`, not `main` and not the beta train. Draft on
  purpose.

## Known follow-up, deliberately not in this PR

The same stale `Read & Write` phrasing survives in three other error messages —
`container-image.yml:259`, `container-image.yml:699` and `release.yml:383`, plus
`beta-release.yml:601`. Issue #359 names only the two places fixed here and the
brief for this change fixed the file boundary at two, so those are left for a
follow-up ticket rather than widened into here.

### DIFF
--- a/.github/workflows/housekeeping.yml
+++ b/.github/workflows/housekeeping.yml
@@ -699,9 +699,13 @@ jobs:
   # every delete call, it refuses to run when empty, and it is unit-tested
   # against a release repository being handed to it by name.
   #
-  # The credential is a *second* token, not the one the image push uses. That
-  # is the point: the push credential never gains delete, and the delete
-  # credential never appears in a job that touches a release repository.
+  # The credential is a *second* token, not the one the image push uses. Both
+  # carry delete now — the description PATCH in `descriptions` is a delete-scope
+  # call, so `DOCKERHUB_TOKEN` is `Read, Write, Delete` too (#359) — and the
+  # second token therefore buys separation rather than a narrower scope: it is
+  # the only credential pointed at a delete endpoint, it never appears in a job
+  # that touches a release repository, and it can be revoked or rotated without
+  # taking the publishing path down with it.
   dev-tag-sweep:
     name: Sweep the -dev image tags
     if: >-
@@ -797,7 +801,9 @@ jobs:
           DOCKERHUB_NAMESPACE: ${{ vars.DOCKERHUB_NAMESPACE }}
           OWNER: ${{ github.repository_owner }}
           # The same one PAT the image push uses — *not* the cleanup token.
-          # Descriptions are a write to a page, not a delete.
+          # It needs `Read, Write, Delete`: the PATCH below is a delete-scope
+          # call on the Hub API, and a `Read & Write` token 403s on it while
+          # image pushes keep working (#359). See docs/REPO_SETUP.md section 2.
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
         run: |
@@ -857,7 +863,7 @@ jobs:
                   "https://hub.docker.com/v2/repositories/${namespace}/${repo}/")"
 
             if [[ "$status" != "200" ]]; then
-              echo "::error title=Failed to update $repo::HTTP $status — $(cat /tmp/resp.json)"
+              echo "::error title=Failed to update $repo::HTTP $status — $(cat /tmp/resp.json). A 403 'insufficient scope' here means DOCKERHUB_TOKEN is scoped Read & Write: this PATCH is a delete-scope call, so the token needs Read, Write, Delete (#359). Image pushes succeed on the narrower scope, which is why they stay green. See docs/REPO_SETUP.md section 2."
               return 1
             fi
             echo "✅ ${namespace}/${repo} — $bytes bytes of README, short description set"
--- a/.github/workflows/housekeeping.yml
+++ b/.github/workflows/housekeeping.yml
@@ -819,7 +819,7 @@ jobs:
             | jq -r '.token // empty')"
 
           if [[ -z "$jwt" ]]; then
-            echo "::error title=Docker Hub auth failed::Could not exchange DOCKERHUB_USERNAME/DOCKERHUB_TOKEN for a JWT. This endpoint rejects organization access tokens ('Cannot log into an organization account') — use a PERSONAL access token with Read & Write scope, from an account with Admin on the org. That one token serves the image push as well."
+            echo "::error title=Docker Hub auth failed::Could not exchange DOCKERHUB_USERNAME/DOCKERHUB_TOKEN for a JWT. This endpoint rejects organization access tokens ('Cannot log into an organization account') — use a PERSONAL access token with Read, Write, Delete scope, from an account with Admin on the org. That one token serves the image push as well."
             exit 1
           fi
 
--- a/docs/REPO_SETUP.md
+++ b/docs/REPO_SETUP.md
@@ -41,7 +41,7 @@ section is what each one is.
 | --- | --- | --- |
 | `DOCKERHUB_USERNAME` | Secret | Publishing `panel` and `core` to Docker Hub, and syncing each image's README |
 | `DOCKERHUB_TOKEN` | Secret | Same — one **personal** access token, `Read, Write, Delete`, not the account password |
-| `DOCKERHUB_CLEANUP_TOKEN` | Secret | The weekly `-dev` tag sweep, and nothing else — a **second**, delete-capable PAT |
+| `DOCKERHUB_CLEANUP_TOKEN` | Secret | The weekly `-dev` tag sweep, and nothing else — a **second** PAT, the only one pointed at a delete endpoint |
 | `NPM_TOKEN` | Secret | Publishing `@actana/sdk` and `@actana/cli` to npm from `release.yml` |
 | `APP_ID` | Secret | The GitHub App's numeric id. Every job in `promote.yml` that pushes |
 | `APP_PRIVATE_KEY` | Secret | That App's private key, the whole PEM. Same jobs — **both or neither** |
@@ -144,13 +144,19 @@ green on the first try and filled all four pages.
 Two consequences worth stating plainly:
 
 - **Never swap this secret for a narrower token.** It is the same credential
-  that pushes `panel`, `core`, `panel-dev` and `core-dev`, so narrowing it back
-  to `Read & Write` blanks the four pages again, and narrowing it below that
-  breaks every image push too.
-- **If you use a fine-grained (repository-scoped) token instead**, it needs the
-  **Repository Edit** permission on all four repositories, not just Read and
-  Write — Edit is what covers the description `PATCH`. The classic PAT above is
-  what this repository is set up and proved against.
+  that pushes `panel`, `core`, `panel-dev` and `core-dev`. Narrowing it back to
+  `Read & Write` does not blank the pages — the `PATCH` 403s and writes nothing,
+  so the last text written stays up while the weekly chore goes red and the four
+  pages quietly stop tracking `docs/images/`. That is harder to notice than a
+  blank page, not easier. Narrowing it below that breaks every image push too.
+- **There is no repository-scoped alternative to reach for.** A Docker Hub PAT
+  carries an account-wide permission level rather than a repository list, and
+  per-repository scoping needs an organization access token — the one credential
+  the `/v2/users/login` exchange rejects outright, above
+  ([ADR 0023](adr/0023-release-trains-and-digest-promotion.md) D38, *the
+  delete-capable credential*; [ADR 0016](adr/0016-the-0-1-0-shape.md) D31). So
+  this token can delete anywhere the account can reach. That is a property to
+  manage — one owner, recorded and rotated — not one to scope away.
 
 ### Rotating it — the token belongs to a person
 
@@ -165,8 +171,10 @@ the push, which is the worst moment to find out.
       an account that leaves as a compromised credential: revoke first,
       re-provision second
 
-To rotate, create the new token first — overwriting the secret is atomic, so
-nothing is unpublishable in between:
+To rotate, create the new token first — with the same **Read, Write, Delete**
+scope, because a rotation that quietly narrows it reintroduces
+[#359](https://github.com/actana/control/issues/359) — and overwrite second;
+overwriting the secret is atomic, so nothing is unpublishable in between:
 
 ```bash
 gh secret set DOCKERHUB_USERNAME --repo actana/control
@@ -205,17 +213,27 @@ what bounds the blast radius of the delete credential below.
       [`core-dev.md`](images/core-dev.md); typing something now only creates a
       thing to disagree with
 
-### `DOCKERHUB_CLEANUP_TOKEN` — the second, delete-capable token
+### `DOCKERHUB_CLEANUP_TOKEN` — the second token, the one aimed at deletes
 
 The weekly `dev-tag-sweep` chore deletes stale `pr-*` and `sha-*` tags from
 `panel-dev` and `core-dev`, and it runs on its own credential rather than on the
-push token. It is therefore a **second** PAT, and it is the more dangerous of
-the two in what it is *pointed at*: Docker Hub personal access tokens carry an
-account-wide permission level rather than a repository list, so either token
-could delete from `actana/panel` and `actana/core` — this is the only one aimed
-at a delete endpoint, and Docker Hub has no undelete.
+push token. Both tokens carry delete — the descriptions `PATCH` above is why the
+push token does — so the second PAT does not buy a narrower scope. It buys two
+things that survive that:
 
-The only thing preventing that is the hard-coded, exact-match repository
+- **Blast radius.** It appears in exactly one job, the weekly sweep, and never
+  in a job that pushes or promotes. The push token is on every PR build, train
+  build, release and promotion; keeping the one credential that is aimed at a
+  delete endpoint out of all of those is the point.
+- **Independent rotation.** It can be revoked the moment the sweep misbehaves,
+  without taking the publishing path down with it.
+
+What it does *not* buy is containment by scope: Docker Hub personal access
+tokens carry an account-wide permission level rather than a repository list, so
+either token could delete from `actana/panel` and `actana/core`, and Docker Hub
+has no undelete.
+
+The only thing bounding the sweep is the hard-coded, exact-match repository
 allowlist in `scripts/lib/dev-tag-sweep.mjs` — re-asserted before every delete
 call, refusing to run when empty, and unit-tested against a release repository
 handed to it by name ([ADR
--- a/docs/REPO_SETUP.md
+++ b/docs/REPO_SETUP.md
@@ -40,7 +40,7 @@ section is what each one is.
 | Name | Kind | Needed for |
 | --- | --- | --- |
 | `DOCKERHUB_USERNAME` | Secret | Publishing `panel` and `core` to Docker Hub, and syncing each image's README |
-| `DOCKERHUB_TOKEN` | Secret | Same — one **personal** access token, `Read & Write`, not the account password |
+| `DOCKERHUB_TOKEN` | Secret | Same — one **personal** access token, `Read, Write, Delete`, not the account password |
 | `DOCKERHUB_CLEANUP_TOKEN` | Secret | The weekly `-dev` tag sweep, and nothing else — a **second**, delete-capable PAT |
 | `NPM_TOKEN` | Secret | Publishing `@actana/sdk` and `@actana/cli` to npm from `release.yml` |
 | `APP_ID` | Secret | The GitHub App's numeric id. Every job in `promote.yml` that pushes |
@@ -99,7 +99,7 @@ The scope itself must exist and the account must own it before the first
 release. `npm access list packages @actana` answers that, and an npm 404 on
 `@actana/sdk` before the first publish is expected rather than a problem.
 
-### It must be a *personal* access token, and one token does both jobs
+### It must be a *personal* access token, `Read, Write, Delete`, and one token does both jobs
 
 Pushing an image and editing a repository's description go through different
 systems: the image push authenticates to the **registry**, where an
@@ -110,10 +110,13 @@ API**, whose `/v2/users/login` endpoint refuses organization accounts outright �
 {"detail":"Cannot log into an organization account"}
 ```
 
-So an OAT would mean paying for a Docker Team or Business subscription **and
-still** provisioning a PAT. One PAT does both instead: create it under your
-**Account settings → Personal access tokens** with **Read & Write** scope, from
-an account with **Admin** on the org, and set `DOCKERHUB_USERNAME` to that
+So a Docker Hub **organization access token is not an option here** — it is
+rejected by the `/v2/users/login` JWT exchange no matter how it is scoped, and
+using one would mean paying for a Docker Team or Business subscription **and
+still** provisioning a PAT. Do not let anyone talk you into an OAT for this
+secret. One classic PAT does both jobs instead: create it under your **Account
+settings → Personal access tokens** with **Read, Write, Delete** scope, from an
+account with **Admin** on the org, and set `DOCKERHUB_USERNAME` to that
 account's own username (not the org — the org goes in `DOCKERHUB_NAMESPACE`).
 
 ```bash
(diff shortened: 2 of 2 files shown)


qcentic-dev:
## Acceptance criteria — issue #359

### Stated acceptance

- [x] **Docs and error message name the scope that actually works.**
      `docs/REPO_SETUP.md` §2 says `Read, Write, Delete` in the secrets table,
      in the subsection heading and at the point the token is created;
      `.github/workflows/housekeeping.yml` line 822 (the `descriptions` job's
      JWT-exchange failure message) now says `Read, Write, Delete scope`. No
      `Read & Write` claim about `DOCKERHUB_TOKEN` survives in either file —
      the phrase appears in the docs only where it is named as the scope that
      is *not* enough.

- [x] **A dispatched `chore=descriptions` run goes green and the four Docker
      Hub pages show the `docs/images/*.md` content.** Already satisfied in ops
      before this PR, and verified independently just now rather than taken on
      trust:
      run [33396982252](https://github.com/actana/control/actions/runs/33396982252)
      — `Docker Hub descriptions` ✓ in 9s — and
      `GET /v2/repositories/actana/{panel,core,panel-dev,core-dev}` returns a
      populated short description and a full description of 4340 / 8793 / 4551 /
      4943 characters against local sources of 4362 / 8833 / 4569 / 4965. All
      four pages render; none is blank.

### "What to do", item by item

- [x] **1. Correct both to Read, Write, Delete, noting that image pushes alone
      are satisfied by Read & Write.** Both corrected. The docs state it
      explicitly: the pushes are satisfied by `Read & Write`, which is exactly
      why a wrongly scoped token looks correct until the `descriptions` chore
      403s, and the empirical proof (same secret value, widened scope, green run
      33396982252 on 2026-08-31) is recorded next to it.

- [ ] **2. Ops: replace the `DOCKERHUB_TOKEN` secret value with an RWD-scoped
      PAT, dispatch `chore=descriptions`, confirm all four pages.**
      *Unchecked here, and not actionable in this PR* — the issue itself scopes
      this as "not in this repo's diff". It has in fact already been done, and
      done more cheaply than the issue anticipated: the **existing** token's
      scope was widened in Docker Hub rather than a new secret value being
      installed, so no `gh secret set` was needed. The dispatch and the four
      pages are verified in the criterion above.

### Extra requirements from the implementation brief

- [x] **`DOCKERHUB_TOKEN` is documented as a classic Docker Hub PAT.** Stated in
      the creation instruction ("One classic PAT does both jobs").
- [x] **Warn that Docker Hub organization access tokens are rejected by the
      `/v2/users/login` JWT exchange; readers must not be steered to OATs.**
      The paragraph now opens with "a Docker Hub **organization access token is
      not an option here**", gives the rejection as the reason, and says plainly
      not to be talked into one.
- [x] **Warn that the same secret drives the image pushes, so it must never be
      swapped for a narrower token.** Its own bullet, naming all four
      repositories it pushes.
- [x] **A fine-grained token, if mentioned as an alternative, needs Repository
      Edit.** Its own bullet; Edit is named as what covers the description
      `PATCH`.
- [x] **Docs only — no workflow logic changes.** The `housekeeping.yml` diff is
      one line, inside one `echo` string. No steps, triggers, permissions,
      conditions or job graph touched.
- [x] **Exactly two files change.** `docs/REPO_SETUP.md` (+39 −11) and
      `.github/workflows/housekeeping.yml` (+1 −1).
- [x] **Files owned by parallel tickets stay untouched.** No change under
      `packages/cli` or `packages/panel`.
- [x] **Base is `feat/0.4.3-onboarding`, not `main`, not the beta train; PR left
      as a draft; nothing merged; `main` untouched.**
- [x] **Conventional Commits, PR title ≤ 105 chars, body lines ≤ 132.** One
      commit, `docs:` type, 94-char header; every commit-body line is within
      132, which is what commitlint checks on all commits and not just the
      title.

### Deliberately unchecked — out of the boundary I was given

- [ ] **The same stale `Read & Write` phrasing elsewhere.**
      `container-image.yml:259`, `container-image.yml:699`, `release.yml:383`
      and `beta-release.yml:601` still describe `DOCKERHUB_TOKEN` as needing
      only `Read & Write`. Issue #359 names two places, and this change was
      bounded to two files, so those are left alone rather than quietly widened
      into this PR. They are wrong in the same way and are worth a follow-up
      ticket; none of them gates the `descriptions` chore, so nothing is broken
      by leaving them for now.

mrfqcentic (changes requested):
## Review of #362 against issue #359 — changes requested

The diagnosis is right, the evidence is real, and I re-proved it rather than
taking it on trust. What blocks this is not the change that was made but the
false premise this change makes false, left standing in the file the PR
already edits.

### Verified independently (not read off the PR description)

- **Exactly two files.** `git diff --stat feat/0.4.3-onboarding...docs/dockerhub-token-scope`
  → `.github/workflows/housekeeping.yml` +1 −1, `docs/REPO_SETUP.md` +39 −11.
- **The workflow diff is one string inside one `echo`.** A single hunk, a
  single `-`/`+` pair at `housekeeping.yml:822`, inside
  `echo "::error title=Docker Hub auth failed::…"`. No step, trigger,
  `permissions:`, `if:` or job-graph change anywhere in the file.
- **The chore is genuinely fixed.** Run
  [33396982252](https://github.com/actana/control/actions/runs/33396982252) —
  job `Docker Hub descriptions` success, 13:26:31→13:26:40 — and a live
  `GET /v2/repositories/actana/{panel,core,panel-dev,core-dev}` returns
  populated short descriptions and full descriptions of 4340 / 8793 / 4551 /
  4943 bytes against local sources of 4362 / 8833 / 4569 / 4965,
  `last_updated` 2026-08-31T15:07–15:08Z. Nothing is blank.
- **All five docs requirements are present** at `REPO_SETUP.md` 43, 102, 113–118,
  131–136, 146–149 and 150–153. CI is green across all 24 checks; commit header
  94 ≤ 120, body max line 82 ≤ 132.

### Blocking

**1. `.github/workflows/housekeeping.yml:800` — the comment says the opposite of this PR, in the job this PR fixes.**

```yaml
          # The same one PAT the image push uses — *not* the cleanup token.
          # Descriptions are a write to a page, not a delete.
```

This sits 22 lines above the error string the PR corrects, in the same step's
`env:` block. The PR's own finding is that writing a description *is* a
delete-scope call. The next person debugging this 403 reads this comment first
and is steered wrong again — which is precisely the failure #359 exists to
stop. It is one line, in a file already open in this diff, inside the declared
two-file boundary.

**2. `.github/workflows/housekeeping.yml:703` — the premise you correctly retired in the docs survives here.**

```yaml
  # The credential is a *second* token, not the one the image push uses. That
  # is the point: the push credential never gains delete, and the delete
  # credential never appears in a job that touches a release repository.
```

"The push credential never gains delete" is now false. This is the same
sentence as the old `DOCKERHUB_CLEANUP_TOKEN` paragraph you rewrote in
`REPO_SETUP.md` for exactly this reason. Your stated rule was that leaving it
"would have made §2 contradict itself two screens apart" — the same rule
applies here, in the same file, and was not applied.

**3. `docs/REPO_SETUP.md:150-153` — the fine-grained bullet contradicts two ADRs of record and the paragraph 35 lines above it.**

```markdown
- **If you use a fine-grained (repository-scoped) token instead**, it needs the
  **Repository Edit** permission on all four repositories, not just Read and
  Write — Edit is what covers the description `PATCH`.
```

`docs/adr/0023-…md:167` (D38, amended by #112) states that "a PAT carries an
account-wide permission level rather than a repository list, and per-repository
scoping needs an Organization Access Token on a Team or Business plan", and
ADR 0016 D31 says an OAT "cannot call `POST /v2/users/login` for descriptions".
Line 113 of this very section says the JWT exchange rejects an OAT "no matter
how it is scoped". So on the repo's own facts the only repository-scoped token
available is the one credential that provably cannot do this job — and an
operator who follows this bullet gets working pushes and a `descriptions` chore
that dies at `housekeeping.yml:822`, a worse variant of #359, on the page's own
advice. This is also the one claim in the PR with no empirical backing behind
it, while everything around it was proved on the token. Delete it, invert it
("there is no repository-scoped option here"), or — if Docker's model has
changed since #112 — amend D38 and say so.

### Should fix, not blocking

- **`docs/REPO_SETUP.md:44`** — the table still calls `DOCKERHUB_CLEANUP_TOKEN`
  "a **second**, delete-capable PAT", which implies by contrast that the push
  token is not. Line 118 of the same file now says it is. Match the table to
  the prose fix at 208: the distinguishing property is what the token is
  *pointed at*, not what it can do.
- **`docs/REPO_SETUP.md:147-148`** — "narrowing it back to `Read & Write` blanks
  the four pages again" is the wrong symptom. A 403 on the `PATCH` returns
  non-200 → `return 1` (`housekeeping.yml:859-861`) and writes nothing; the
  stored descriptions stay exactly as last written. The pages were blank only
  because they had never been filled. The real regression is a red weekly chore
  and four pages that quietly stop tracking `docs/images/`.
- **`.github/workflows/housekeeping.yml:860`** — the string this PR fixed is on
  the *JWT-exchange failure* branch, and a `Read & Write` PAT exchanges for a
  JWT perfectly well, so #359's operator never sees it. The message they do see
  is `echo "::error title=Failed to update $repo::HTTP $status — …"`, which
  prints a bare `403 access denied: insufficient scope` with no scope hint and
  no pointer to §2. #359 named the string you changed, so this is an addition
  rather than a miss — but it is the one line that would have prevented the
  original investigation.

### On the two judgement calls

**The `DOCKERHUB_CLEANUP_TOKEN` rewrite (`REPO_SETUP.md:208-217`) — correct and
in scope.** The old premise was falsified by this very change, the rewrite is
factually right, it is confined to the same section of the same file, and it
was declared in both the PR body and the commit message rather than slipped in.
Keeping it here was the right call; splitting it out would have shipped a
self-contradicting §2. Two caveats: "it runs on its own credential rather than
on the push token. It is therefore a **second** PAT" is now circular — the
reason a second token exists is no longer stated (blast radius and independent
rotation, presumably); and the posture genuinely widened on 2026-08-31 without
anything recording it. The push token is now delete-capable *and* is exposed on
every PR build, train build, release and promotion with no allowlist guard,
while the cleanup token is bounded by `scripts/lib/dev-tag-sweep.mjs`. Yet
`docs/ci-cd.md:175-181` still calls that allowlist "the only guard there is"
for "*the* delete-capable credential", singular. This repo amends ADRs by
ticket for exactly this class of claim (D38 ← #112). Open one.

**Deferring the four other stale messages — right in principle, incomplete in
practice.** `container-image.yml:259`, `container-image.yml:699`,
`release.yml:383` and `beta-release.yml:601` are correctly enumerated and are
genuinely outside a two-file boundary, and none of them gates the chore. So
deferring them is defensible. But (a) no follow-up ticket exists yet — I
checked; the deferral is currently a promise with nothing tracking it; (b)
three of those four instruct the wrong scope at the exact moment a token is
created, and two tie it explicitly to "the description sync", so the next
rotation reintroduces #359 silently — and §"Rotating it" in this file never
restates the scope either; (c) the enumeration missed the two above, which are
in `housekeeping.yml` itself. Files, not error strings, was the boundary you
declared — and both of those are inside it.

File the ticket and link it here, fix the three blocking items, and this is a
good change: well diagnosed, honestly evidenced, and correctly scoped.

qcentic-dev:
> File the ticket and link it here, fix the three blocking items, and this is a good change.

Three blockers fixed, four of the "should fix" items fixed, one addition you
made in passing fixed, and two things deferred with the reason. Pushed as
`a7caa44`. Still two files, still docs and comments only.

## Blocking — all three

**1. `housekeeping.yml:800` — "Descriptions are a write to a page, not a delete".**
Fixed. You are right that this was the worst of the three: it is in the `env:`
block of the step that 403s, 22 lines above the string the first commit
corrected, so it is what the next person reads first. It now says the token
needs `Read, Write, Delete`, that the `PATCH` below is the delete-scope call,
that a `Read & Write` token 403s there while image pushes keep working, and
points at §2.

**2. `housekeeping.yml:703` — "the push credential never gains delete".**
Fixed, and the inconsistency is fair — I applied that rule in `REPO_SETUP.md`
and not in the workflow that motivated it. Restated: both tokens carry delete
now, so the second token buys *separation*, not a narrower scope — it is the
only credential pointed at a delete endpoint, it is in no job that touches a
release repository, and it can be revoked without taking the publishing path
down.

**3. `REPO_SETUP.md:150-153` — the fine-grained bullet.** Removed and inverted,
not softened. I checked both citations rather than taking them from the review:
ADR 0023 D38 (*the delete-capable credential*) says a PAT carries an
account-wide permission level and per-repository scoping needs an OAT, and ADR
0016 D31 says an OAT cannot call `POST /v2/users/login`. So the bullet pointed
an operator at the one credential that provably cannot do the job, 35 lines
below the paragraph saying so. It now reads "there is no repository-scoped
alternative to reach for", cites both ADRs, and says account-wide reach is a
property to manage — one owner, recorded and rotated — rather than one to scope
away. You were also right that it was the only unproved claim in the PR; it came
from the brief, and I should have checked it against D38 before writing it.

## Should fix — done

- **`REPO_SETUP.md:44`** — the table row no longer calls the cleanup token the
  "second, **delete-capable**" PAT. It is now "a **second** PAT, the only one
  pointed at a delete endpoint", matching the prose fix. The subsection heading
  carried the same implication by contrast and got the same treatment.
- **`REPO_SETUP.md:147-148`, the wrong symptom** — corrected, and it is a better
  point than a wording nit: the `PATCH` 403s and writes nothing, so a later
  narrowing leaves the last text in place while the weekly chore goes red and
  the four pages quietly stop tracking `docs/images/`. The doc now says that,
  and says it is harder to notice than a blank page rather than easier.
- **The circular rewrite** — the reasons are now stated instead of assumed:
  blast radius (one job, the weekly sweep, never one that pushes or promotes)
  and independent rotation, with "containment by scope" named explicitly as the
  thing the second token does *not* buy. I also narrowed the following sentence
  from "the only thing preventing that" to "the only thing bounding the sweep",
  since the allowlist bounds the sweep and nothing bounds the push token.
- **`housekeeping.yml:860`, the message the operator actually sees** — added,
  and your reasoning is the reason it is in scope: a `Read & Write` PAT
  exchanges for a JWT fine, so #359's operator never reached the string #359
  named. The `Failed to update` error now carries the scope hint, says image
  pushes succeed on the narrower scope (which is why they stay green and hide
  it), and points at §2.
- **Also, unprompted, from your point (b):** §"Rotating it" never restated the
  scope, so the next rotation could silently reintroduce #359. It now says to
  create the replacement with the same `Read, Write, Delete` scope.

## Deferred, with what is tracking it

- **The four stale messages** in `container-image.yml:259`, `:699`,
  `release.yml:383` and `beta-release.yml:601` stay deferred — they are in three
  files outside this PR's boundary. Your (a) is the fair hit: a deferral with
  nothing tracking it is a promise. The orchestrator is arranging the follow-up
  ticket, and it will be linked here when it exists. Your (c) is also right and
  is fixed rather than deferred: the two stale comments were in
  `housekeeping.yml` itself, inside the boundary I declared, and they are two of
  the three blockers above.
- **The ADR amendment for the posture change** — the push token is now
  delete-capable and is exposed on every PR build, train build, release and
  promotion with no allowlist guard, while `docs/ci-cd.md:175-181` still calls
  that allowlist "the only guard there is" for "*the* delete-capable
  credential", singular. That is a real change of posture, it is a third file,
  and this repo amends ADRs by ticket (D38 ← #112). Deferred to a ticket rather
  than asserted in a docs PR — I did not want the amendment written as a side
  effect of the change that caused it.

## State

CI green on `a7caa44`: 18 checks pass, 0 fail, 6 skipping (train-only jobs).
Base is still `feat/0.4.3-onboarding`, the PR is still a draft, and nothing has
been merged.

mrfqcentic (approved):
## Re-review of `a7caa44` — approved

All three blockers are fixed at the enforcing lines, all five non-blocking items
are fixed, two of them better than I asked for. I verified every one against the
file rather than the reply. The deferral is released; see the bottom.

### Blocking — all three cleared

**1. `housekeeping.yml:803-806`.** The premise is gone, not softened:

```yaml
          # The same one PAT the image push uses — *not* the cleanup token.
          # It needs `Read, Write, Delete`: the PATCH below is a delete-scope
          # call on the Hub API, and a `Read & Write` token 403s on it while
          # image pushes keep working (#359). See docs/REPO_SETUP.md section 2.
```

A grep of the whole file for the retired phrasings — `not a delete`,
`write to a page`, `never gains delete`, `must never have` — returns nothing.
The comment the next debugger reads first now tells them the right thing and
points at §2.

**2. `housekeeping.yml:702-708`.** Correctly restated. "Both carry delete now …
the second token therefore buys separation rather than a narrower scope: it is
the only credential pointed at a delete endpoint, it never appears in a job that
touches a release repository, and it can be revoked or rotated without taking
the publishing path down with it." That is the accurate invariant, and it is the
same one now in `REPO_SETUP.md:218-234` — the two no longer disagree.

**3. `REPO_SETUP.md:152-159` — inverted, and it checks out against both ADRs.**
I read the replacement against the sources rather than the citation:

- ADR 0023 D38 (`0023-…md:167`): *"a PAT carries an account-wide permission level
  rather than a repository list, and per-repository scoping needs an
  Organization Access Token on a Team or Business plan"* → the bullet says
  exactly that, near verbatim.
- ADR 0016 D31 (`0016-the-0-1-0-shape.md:147`): *"An OAT can push images but
  cannot call `POST /v2/users/login` for descriptions"* → the bullet's "the one
  credential the `/v2/users/login` exchange rejects outright, above".

Both relative links resolve to files that exist, and the D38 shorthand *the
delete-capable credential* matches how `ci-cd.md:180` already cites it. The
bullet now points *at* the paragraph 40 lines above it instead of against it,
and closes on the right note — account-wide reach is "a property to manage — one
owner, recorded and rotated — not one to scope away". That is the correct
posture given there is no scoping option to reach for.

### Non-blocking — all five done

- **`REPO_SETUP.md:44`** — now "a **second** PAT, the only one pointed at a
  delete endpoint". You also renamed the subsection heading (216), which I had
  not asked for and which carried the same implication. I checked for fallout:
  no `REPO_SETUP.md#…` anchor anywhere in the repo points at either heading you
  renamed across the two commits, so nothing is broken.
- **`REPO_SETUP.md:146-151`** — the symptom is now right, and I confirmed the
  mechanism at `housekeeping.yml:865-867`: non-200 → `return 1`, nothing
  written. "Harder to notice than a blank page, not easier" is the part that
  makes it useful.
- **The circular rationale (`218-234`)** — both reasons now stated (blast
  radius; independent rotation) with containment-by-scope named as what the
  second token does *not* buy. Narrowing "the only thing preventing that" to
  "the only thing bounding the sweep" at line 236 was not asked for and is
  strictly more accurate: the allowlist bounds the sweep, and nothing bounds the
  push token.
- **`housekeeping.yml:866`** — the message #359's operator actually sees now
  carries the scope, the reason the pushes stay green, and the §2 pointer.
- **`REPO_SETUP.md:174-177`** — the rotation clause you added unprompted closes
  the path I was most worried about in (b): a rotation can no longer silently
  narrow the scope.

### No logic changed — verified structurally

I stripped comments and blank lines from `housekeeping.yml` at base and at
`a7caa44` and diffed the remainder: **exactly two lines differ, both `echo`
strings inside the `descriptions` job.** `on:`, `permissions:`, every `if:`,
`runs-on`, the steps and the job graph are byte-identical to the base. `bash -n`
on the extracted `descriptions` run block passes. Still exactly two files
(`housekeeping.yml` +12 −6, `REPO_SETUP.md` +63 −17). CI on `a7caa44`: 18 pass,
6 skipped, 0 fail (run 33414382440). Header 90 chars, body max line 81, blank
line before `Refs #359`.

### The ticket — released, not blocking

My objection was that a deferral with nothing tracking it is a promise, and that
objection is now answered well enough to clear. Three things changed it: the
dangerous half of it — the two stale comments inside `housekeeping.yml`, in the
boundary you declared — is fixed rather than deferred; the regression path is
now closed at both points of contact, by `REPO_SETUP.md:174-177` for the
operator who rotates and by `housekeeping.yml:866` for the operator who
debugs; and the deferral is on the record here, naming all four locations, with
the owner arranging the ticket. Holding a correct docs change hostage to a ticket
someone else has to file would be the wrong trade at this point.

Two things stay wrong until that ticket lands, and both belong in it:

- `container-image.yml:259`, `container-image.yml:699`, `release.yml:383` and
  `beta-release.yml:601` still instruct `Read & Write` — three of them at the
  moment a token is created, two tying it to "the description sync".
- The posture amendment: `docs/ci-cd.md:175-181` still calls the sweep allowlist
  "the only guard there is" for "*the* delete-capable credential", singular,
  citing ADR 0023 D36/D38/D45. Deferring the ADR amendment out of the PR that
  caused it is the right instinct — D38 was itself amended by #112 — but it does
  need to be written.

One trivial nit, take it or leave it: `housekeeping.yml:866` says a 403 "means
DOCKERHUB_TOKEN is scoped Read & Write", where a read-only token or an account
without Admin on the org would 403 in the same place. "Scoped too narrowly
(typically Read & Write)" would be exact. Not worth a commit on its own.

Good round. The diagnosis was right the first time, the three fixes go to the
premise rather than the wording, and the two additions you made unprompted are
the ones that actually stop #359 coming back.

Approving.

Issue #360, opened 2026-08-31
actana core pair: framed success and failure output with next steps


mrfqcentic:
Part of the 0.4.3 onboarding improvements. Client-side companion to #357 (which covers `actana pair new` on the Core).

## Problem

`actana core pair` ends in two bare lines on success (`Paired Core "x" → wss://…`, plus the `current` note) and terse stderr prose on failure. For the person onboarding their first Core this is the moment that most needs guidance, and it gets the least.

## What to build

TTY-gated (same switch as #357; piped output unchanged):

**On success** — a framed, colored confirmation block:
- what just happened: the Core's name, label, endpoint, and whether `current` now points at it;
- **next steps**: `actana core status` to verify the link, `actana core use <name>` when it is not current, `actana session start …` / `actana core shell` as the first real actions, and a pointer to pairing the Panel with the same Core if not done yet;
- where the credential landed (the cores directory, mode 0600 — reassurance, not secrets; never print the blob).

**On failure** — keep the existing per-class explanations (they are good) but frame them, and end every class with concrete next steps rather than prose alone: the exact `actana pair new` re-mint command for a stale fingerprint/expired code, the `--session` reminder for a bare code, connectivity hints (DNS, port, SNI/proxy) for a dial failure.

## Acceptance

- Success on a TTY shows the framed block with next steps; piped success output is unchanged.
- Each failure class ends with a copyable command, not only an explanation.
- Nothing secret is ever printed in either mode (existing invariant in `core-pair.ts` holds).

mrfqcentic:
Shipped in v0.4.3.

Issue #357, opened 2026-08-31
actana pair new: instructional, framed output when stdout is a TTY


mrfqcentic:
Part of the 0.4.3 onboarding improvements (grilled 2026-08-31).

## Problem

`actana pair new` prints bare labelled lines (code, fingerprint, expiry, session) with the explanation on stderr. For the person actually pairing, there is no visual framing, no statement of what to do next, and no ready-made command for the other machine. The operator is left to reassemble `actana core pair …` by hand from four separate lines.

## What to build

When **stdout is a TTY**, print a cosmetic, instruction-rich block instead of the bare lines:

- A framed block (unicode box/rules, ANSI color) presenting the **pairing code** and **CA fingerprint** prominently, with expiry and session below.
- Under it, **both** redemption paths, always, in one block:
  - **Panel** — the canonical path, worded exactly: *Settings (gear icon) → Cores → Add Core → enter this code and fingerprint.*
  - **CLI** — first `npm i -g @actana/cli`, then a fully pasteable line with **real values** filled in:
    `actana core pair <label-or-placeholder> <endpoint> <code> --session <id> --fingerprint <fp>`
    One line **per configured public host** (0.4.2 supports several addresses per Core), each with that host's endpoint filled in. Use `--label` for the name when given, a `<name>` placeholder otherwise.

When stdout is **not** a TTY, the output stays byte-identical to today's bare labelled lines — the screen-scraping contract in `packages/cli/src/actana-pair.ts` (stdout = facts, stderr = prose) is preserved. No `--json` flag; TTY detection is the whole switch.

## Acceptance

- TTY run shows the framed block with both paths and pasteable per-host `core pair` lines with real values.
- Piped run is unchanged from 0.4.2 output.
- The pasteable line actually works when pasted on a second machine against the minted code.

mrfqcentic:
Found while pairing the 0.4.2 fleet: both usage strings omit the required `--session` flag — `packages/cli/src/core-pair.ts` ("actana core pair <name> <host:port> <code> --fingerprint <sha256>") and the help block in `packages/cli/src/core-command.ts` ("actana core pair <name> <address> <code>") — while `readTicket` refuses a bare code without `--session <id>` or the `<session>:<code>` form. Fix both strings as part of this ticket; the new TTY block's pasteable line must of course carry `--session` too.

mrfqcentic:
Shipped in v0.4.3.

Issue #12, opened 2026-08-03
Harness to Core: rename scope and the migration it forces


qcentic-adm:
Part of #6

## Question

**What exactly does the Harness → Core rename touch, and what happens to installs that already exist?**

"Harness" is the old word for what is now called Core, and the split vocabulary is the single largest contributor to the repo reading as bloated and confusing. The rename is decided; its blast radius is not.

Enumerate and decide each:

- `packages/harness/` → `packages/core/`, and `@actana/harness` → `@actana/core`.
- Workflow, script and library names: `harness-release.yml`, `build-harness-tarball.mjs`, `compose-harness-shasums.mjs`, `smoke-harness-tarball.mjs`, `smoke-standalone-harness.mjs`, `scripts/lib/harness-tarball.mjs`, `scripts/lib/harness-smoke.mjs`, the `harness:*` package scripts.
- Release asset names: `harness-<target>.tar.gz` → `core-<target>.tar.gz`. This is user-visible and breaks any pinned download URL.
- Exported constants: `HARNESS_TARGETS`, `HARNESS_RUNTIME_DEPENDENCIES`, `BUNDLED_NODE_VERSION`.
- **The `AC_*` environment prefix.** Does it stay? `AC_PANEL_DATA_DIR` etc. are already neutral; anything `AC_HARNESS_*` is not.
- **The `actana` CLI binary.** Does it stay `actana` (a brand name, arguably correct) or become `core` (a role name, matching the new vocabulary)? Verbs are `setup`, `status`, `token`, `start`/`stop`/`restart`/`logs`, `daemon`.
- **The systemd unit name** and install paths under the operator's home.
- Docs, ADRs and `CONTEXT.md` prose.

Then settle the compatibility question that determines whether this is a clean sweep or a two-step: **what happens to a machine that already has a Harness installed, and to a Panel holding paired Core records?** Options range from "pre-1.0, break it, document it" to a compatibility shim in the installer. Pick one and say what the release notes have to say.

qcentic-dev:
## Decision — Harness → Core: it is two renames, and there is nothing to migrate

Grepped before deciding. Two things the ticket's list got wrong, and they change the shape of the work.

### 1. "Harness" is a homonym. The rename splits, it does not sweep.

[...]

Issue #11, opened 2026-08-03
babel/postcss advisories: shipped artifact or build stage only?


qcentic-adm:
Part of #6

## Question

**Are the reported `@babel/core` and `postcss` advisories actually shipped, and what is the fix?**

Reported: CVE-2026-49356 against `@babel/core`, and a 7.5-severity GHSA against `postcss`, found on a scan of the Panel source copy.

What charting already established, to be confirmed not repeated:

- The lockfile pins `@babel/core@7.29.0` and `postcss@8.5.15`.
- Neither is in `packages/panel/package.json` `dependencies`. Both are **transitive devDependencies** — `@babel/core` via `@vitejs/plugin-react`, `postcss` via `tailwindcss`.
- The runtime stage copies `/srv/panel`, produced by `pnpm --filter @actana/panel deploy --prod`, which should exclude devDependencies entirely.

So resolve:

1. **Confirm the pruned prod tree has neither package.** Inspect the actual built image, not the source tree.
2. **Establish what the scanner was scanning.** The `COPY . .` in the build stage puts the whole repo — and `pnpm install --frozen-lockfile` puts every devDependency — into an intermediate layer. Was the finding against the build stage, the source checkout, or the shipped artifact? This distinction decides whether there is anything to fix at all.
3. **Decide the fix path.** Upgrade to patched versions if they exist; `pnpm.overrides` to force a patched transitive; or document "build-time only, not shipped" and configure the scanner to reflect that. State which, and why.
4. **Decide whether the build stage should stop `COPY . .`-ing the whole repo** — a narrower copy shrinks the scanned surface and the cache-bust radius regardless of these two advisories.
5. Check whether the same reasoning covers other build-time-only findings in the 74, or whether those are genuinely base-distro CVEs that only the base change (see the Panel distroless prototype) will clear.

qcentic-dev:
## Resolved — but the premise was wrong: **both packages ship.**

Charting had these as transitive devDependencies pruned away by `pnpm deploy --prod`. They aren't. Both reach the runtime image through the Panel's **production** graph.

Full write-up committed as `docs/panel-image-shipped-dependencies.md` on branch `wt-11-babel-postcss` (`d58c897`, off `087b550`). Not pushed — no remote in this clone points at `actana/control`.

### 1. The pruned prod tree contains both

[...]

PR #363, opened 2026-08-31
feat(panel): a first-run pairing wizard — no dashboard until a Core is paired

### LINKED ISSUE #358: Panel: first-run pairing wizard — no dashboard until a Core is paired
Part of the 0.4.3 onboarding improvements (grilled 2026-08-31). Companion to #357.

## Problem

A fresh Panel has no onboarding: you register the Operator with a password and land on an empty dashboard. Nothing tells you a Panel is useless without a Core, where pairing happens (Settings → Cores → Add Core), or what to run on the Core machine. The remote-pairing docs compound this by assuming you already know where to look.

## What to build

A pairing **wizard** that replaces the dashboard whenever the Panel has **zero Cores** — first boot after Operator creation, and again if the last Core is ever removed. It is **not skippable**: a Panel that is not paired with a Core has nothing else to show.

Each step shows the machine-side CLI commands for that step, copyable:

1. **Install the Core** — the install command(s) on the machine to be paired (install.sh / docker compose per the existing docs).
2. **Mint a code** — `actana pair new` (with `--label`), and what its output looks like (the #357 block).
3. **Redeem** — the code + fingerprint form, reusing the existing `AddCoreByPairing` component.

On the first successful pairing the wizard ends and the normal dashboard appears.

## Acceptance

- Fresh Panel: register Operator → wizard, no dashboard route until a Core exists.
- Deleting the last Core brings the wizard back.
- Each wizard step displays its copyable Core-side command.
- Redeem step is the same `AddCoreByPairing` flow, not a fork of it.
- Settings → Cores → Add Core remains available and unchanged; the wizard is an additional path, and both name the same canonical location.

qcentic-dev:
## Summary

A Panel that knows **zero Cores** now draws a three-step pairing wizard instead of the app shell, and keeps drawing it
until a Core pairs. The three steps are the three things that have to happen — install the Core on the machine you want
to drive, mint a code there, redeem it here — and each one carries the **Core-side commands** for that step, copyable,
because the person reading the Panel is not sitting at the machine they are pairing.

The gate is the **live Core count**, not a first-run flag: a freshly created Operator lands in the wizard rather than on
an empty dashboard, and a Panel whose last Core is forgotten goes straight back to it. One rule answers both, because
they are the same state — nothing to reset, nothing to migrate.

It **replaces** the shell rather than covering it. At zero Cores the top bar, the project rail, the router outlet and
the settings overlay do not mount at all, so there is no route to type, no key to press and no dead dashboard to glimpse
behind it. There is no skip, no "later" and no dismiss affordance anywhere on the screen.

Step 3 is `AddCoreByPairing` — **the same component `Settings → Cores → Add Core` mounts**, with the same `onPaired`
contract — so the fingerprint-before-code ordering, the mismatch refusal and every failure sentence stay written once.
This PR is composition and guidance around that form, not a second implementation of the most safety-critical flow in
the product. `CoresSettingsPage` keeps its own `AddCoreByPairing` and its behaviour is unchanged; the two paths name the
same canonical location, and the wizard says so out loud.

**Base is `feat/0.4.3-onboarding`, not the open train and not `main`** — this is one of the 0.4.3 onboarding tickets and
stacks on that branch alongside #357. Left as a **draft** deliberately.

### What changed

| File | |
| --- | --- |
| `src/shared/core-onboarding.ts` | new — the Core-side commands as data: both documented install paths, the `pair new` output lines and what each is for, and `pairNewCommand(label)` with shell quoting |
| `src/components/views/FirstRunWizard.tsx` | new — the three steps, the copyable command blocks, the live `--label` rewrite, and `AddCoreByPairing` as step 3 |
| `src/components/views/FirstRunGate.tsx` | new — the gate: a live `listCores()` count decides wizard vs. app; lazy-loads the wizard so a Panel with a fleet never ships those bytes |
| `src/lib/core-registry-changed.ts` | new — a payload-free "ask again" nudge, so a Core paired or forgotten is seen at once rather than at the next 15s poll |
| `src/routes/__root.tsx` | the gate wraps `<Shell />` inside `ClientOnly`; `/login` and `/setup` render outside it, untouched |
| `src/components/views/CoresSettingsPage.tsx` | announces the registry change after a pair and after a remove — two lines; the Add Core flow itself is untouched |
| `src/lib/design-meta.ts` | the event name, beside the other window events |

## Related issues

Closes #358

Companion to #357 (`actana pair new` framed TTY output). No CLI surface is touched here — #357 and #360 own those, and
this branch changes nothing outside `packages/panel`.

## Type of change

- [ ] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## How was this tested?

`src/components/views/__tests__/first-run-gate.test.tsx` — 22 tests driving the real gate against a mocked
`~/lib/api`, including the two the issue turns on:

- **zero Cores renders the wizard and not the dashboard** — the wizard is on screen, the child the gate wraps is not;
- **a successful pairing unlocks the dashboard** — the full redemption (address → check fingerprint → compare → session
  + code → Pair Core) through the real `AddCoreByPairing`, after which the child renders and the wizard is gone.

Plus: a paired Panel paints its shell on the *first* client render from the cached count, and the live read still
corrects a stale seed back into the wizard; a single empty successful read does not tear down a live session, two
consecutive ones do; a registry read that failed never unlocks the dashboard and says which of "no Cores" and "could
not ask" happened; one failed poll does not throw a paired Panel back into the wizard; forgetting the last Core brings
the wizard back; the Compose step brings up `core` alone at a tag that resolves; a name starting with `-` prints the
placeholder and the reason rather than a command the Core refuses; the dashboard stays away on every step and on the
way back; the mint command is rewritten and shell-quoted as the operator types a name; commands are copyable; and the
redeem step posts the same body `Settings → Cores` posts. No button on any of the three steps matches skip / dismiss /
later / close / continue-anyway — a keyword scan, which cannot prove absence on its own; the hardness is proven
structurally, by the gate returning the wizard *instead of* its children.

`src/shared/__tests__/core-onboarding.test.ts` — 19 tests. The `pair new` block is checked **against `packages/cli`**,
not against a copy of itself: the label set is read out of the contract the CLI's own suite pins for this invocation,
and the order out of the printer's `deps.out` calls, so drift from the real printer fails this suite. Verified by
deleting the `Label` entry and watching three tests fail.

`src/components/views/__tests__/cores-settings-pairing.test.tsx` — 39 tests, three of them new: pairing announces,
removal announces, and a failed removal announces nothing.

Full `packages/panel` suite: **146 files, 1535 tests, all passing.** `pnpm -r typecheck` clean. `pnpm lint`: 0 errors
(46 pre-existing warnings, unchanged in count, none in the files this PR adds or edits).

Not verified by running the app against a live Core — the gate and the wizard are exercised at the component level only.

## Screenshots / recordings

None — no browser was driven for this change.

## Checklist

- [ ] This PR targets the **open train** (`beta/x.y.z`), not `main` — deliberately not: it targets
      `feat/0.4.3-onboarding`, the 0.4.3 onboarding branch this ticket belongs to. `Train rules` passes on that base
      (it is neither `main`, a train, nor a release line, so the branch model does not constrain it); the train rule
      applies when the onboarding branch itself is proposed.
- [x] My PR title **and every commit in the branch** follow Conventional Commits, and my branch name follows the
      branching convention — all three commits and the title were run through this repo's own `commitlint.config.mjs`.
- [ ] I updated documentation (README, docs/, changelog entry) where relevant — out of scope by instruction: this branch
      is confined to `packages/panel`, and `README.md` / `INSTALL.md` already document every command the wizard prints.
- [x] Dependent changes have been merged (or this PR is marked as draft/stacked) — stacked on `feat/0.4.3-onboarding`
      and left as a draft.

### DIFF
--- a/packages/panel/src/components/views/CoresSettingsPage.tsx
+++ b/packages/panel/src/components/views/CoresSettingsPage.tsx
@@ -6,6 +6,7 @@ import { Field, SettingsSection } from "~/components/views/SettingsParts";
 import { AddCoreByPairing } from "~/components/views/AddCoreByPairing";
 import { CoreNeedsUpdateNotice } from "~/components/views/CoreNeedsUpdate";
 import { api } from "~/lib/api";
+import { announceCoreRegistryChanged } from "~/lib/core-registry-changed";
 import { formatRelativeTime } from "~/lib/format-relative-time";
 import { coreOrder, type CoreDialStatus, type CoreWithDial } from "~/shared/cores";
 
@@ -74,6 +75,10 @@ export function CoresSettingsPage() {
    */
   const handlePaired = async (core: CoreWithDial) => {
     await refresh();
+    // The fleet just went from N to N+1. Anything else in this tab watching the
+    // registry — the first-run gate above all — re-reads now rather than at its
+    // next poll (#358).
+    announceCoreRegistryChanged();
     toast.success(`Core "${core.label}" paired.`);
   };
 
@@ -110,6 +115,10 @@ export function CoresSettingsPage() {
       await api.removeCore(core.id);
       setPendingRemoval(null);
       await refresh();
+      // And N to N-1. When that leaves zero, the first-run gate is what puts
+      // the pairing wizard back up — the gate is on the count, not on whether
+      // this Panel has ever been paired (#358).
+      announceCoreRegistryChanged();
       toast.success(`Core "${core.label}" removed.`);
     } catch (err) {
       toast.error(err instanceof Error ? err.message : "Failed to remove Core.");
--- a/packages/panel/src/components/views/CoresSettingsPage.tsx
+++ b/packages/panel/src/components/views/CoresSettingsPage.tsx
@@ -7,6 +7,7 @@ import { AddCoreByPairing } from "~/components/views/AddCoreByPairing";
 import { CoreNeedsUpdateNotice } from "~/components/views/CoreNeedsUpdate";
 import { api } from "~/lib/api";
 import { announceCoreRegistryChanged } from "~/lib/core-registry-changed";
+import { ADD_CORE_FIELD_LABEL } from "~/shared/core-onboarding";
 import { formatRelativeTime } from "~/lib/format-relative-time";
 import { coreOrder, type CoreDialStatus, type CoreWithDial } from "~/shared/cores";
 
@@ -173,7 +174,9 @@ export function CoresSettingsPage() {
             )}
           </Field>
 
-          <Field label="Add a Core">
+          {/* Named from the same constant the first-run wizard points at, so
+              the two paths name one place with one string (#358). */}
+          <Field label={ADD_CORE_FIELD_LABEL}>
             <AddCoreByPairing onPaired={handlePaired} />
           </Field>
         </>
--- a/packages/panel/src/components/views/FirstRunGate.tsx
+++ b/packages/panel/src/components/views/FirstRunGate.tsx
@@ -1,0 +1,161 @@
+import {
+  Suspense,
+  lazy,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+  type ReactNode,
+} from "react";
+import { api } from "~/lib/api";
+import { onCoreRegistryChanged } from "~/lib/core-registry-changed";
+
+// Lazy, for the same reason the settings overlay is: on every load of a Panel
+// that *has* a fleet — which is every load after the first — the wizard and the
+// pairing form it mounts are bytes nobody will render. The gate itself has to
+// stay eager, because deciding is its whole job and it decides before anything
+// paints; it is a `listCores()` and a branch, and costs the entry chunk nothing.
+const FirstRunWizard = lazy(() =>
+  import("~/components/views/FirstRunWizard").then((m) => ({ default: m.FirstRunWizard })),
+);
+
+/**
+ * The gate (#358): a Panel that knows no Cores shows the pairing wizard, and
+ * a Panel that knows one shows the app.
+ *
+ * **The condition is the count, not the calendar.** This is not a first-run
+ * flag, not a "seen it" preference, and not something the operator can be past.
+ * It is a live read of the Core registry, so a fresh Panel lands in the wizard
+ * and a Panel whose last Core was forgotten goes back to it — the same rule
+ * answering both, because they are the same state. There is nothing to reset
+ * and nothing to migrate.
+ *
+ * **It replaces the shell rather than covering it.** `children` here is the
+ * entire app — top bar, project rail, router outlet, settings overlay — and at
+ * zero Cores none of it mounts. That is what makes this a gate rather than a
+ * modal: there is no route to type, no escape key, no click-outside, and no
+ * dead dashboard behind the wizard to glimpse. The only exit is a paired Core.
+ *
+ * The one screen this does *not* stand in front of is `/login` and `/setup`:
+ * `__root.tsx` renders those outside the shell entirely, so an operator still
+ * creates their account first and meets the wizard immediately after.
+ */
+
+/**
+ * How often to re-ask, absent an announcement.
+ *
+ * The same fifteen seconds `useCores` polls on, and for the same reason: only
+ * an operator changes this registry. The poll is the backstop for a change this
+ * tab did not make; the announcement below is what makes the changes it *did*
+ * make instant.
+ */
+const REGISTRY_POLL_MS = 15_000;
+
+export function FirstRunGate({ children }: { children: ReactNode }) {
+  const { count, error, refresh } = useCoreRegistry();
+
+  /**
+   * The pairing form's `onPaired`, and the reason it is awaited.
+   *
+   * `AddCoreByPairing` holds its busy state until this resolves, so re-reading
+   * the registry here — rather than optimistically counting the Core it just
+   * handed us — means the form stays visibly mid-pairing right up to the moment
+   * the dashboard is genuinely unlocked. The alternative flashes an empty
+   * wizard between the pairing landing and the count catching up, which reads
+   * as a failure of the thing that just succeeded.
+   *
+   * Counting the returned Core instead would also be the gate believing
+   * something other than the registry, which is the one thing it must not do.
+   */
+  const handlePaired = useCallback(async () => {
+    await refresh();
+  }, [refresh]);
+
+  // Not yet known, and no reason to think it is unknowable: draw nothing.
+  // Rendering the wizard here would flash it at every operator with a fleet on
+  // every load, and rendering the shell would flash a dashboard at operators
+  // with none — the two mistakes this component exists to prevent.
+  if (count === null && error === null) return null;
+
+  // A registry we could not read is not a registry with no Cores in it, but it
+  // is certainly not one we can unlock a dashboard on. The wizard is the honest
+  // screen for it, and it says which of the two happened.
+  if (count === null || count === 0) {
+    return (
+      <Suspense fallback={null}>
+        <FirstRunWizard onPaired={handlePaired} registryError={error} />
+      </Suspense>
+    );
+  }
+
+  return <>{children}</>;
+}
+
+/**
+ * How many Cores this Panel is paired with, kept fresh.
+ *
+ * Deliberately not `useCores`. That hook is built for the Cores list: it holds
+ * the rows, folds live dial pushes into them, and refreshes on a nonce. This
+ * needs none of the rows and cannot use a nonce, because the gate has to be
+ * able to *wait* for a re-read — `onPaired` is a promise the pairing form is
+ * holding its busy state on, and a nonce cannot be awaited. What is shared is
+ * the thing that matters: both ask `api.listCores()`, so there is one answer
+ * about the fleet and no second source of truth for it.
+ *
+ * `count` is null until a read succeeds, and never returns to null afterwards:
+ * a poll that fails leaves the last known count standing, so a Panel with a
+ * fleet does not get thrown into the wizard by one bad request.
+ */
+function useCoreRegistry(): {
+  count: number | null;
+  error: string | null;
+  refresh: () => Promise<void>;
+} {
+  const [count, setCount] = useState<number | null>(null);
+  const [error, setError] = useState<string | null>(null);
+
+  const mounted = useRef(true);
+  // Monotonic, so a slow poll that lands after a fast explicit refresh cannot
+  // overwrite the newer answer with its older one — which, on this component,
+  // would mean re-locking a dashboard that had just been unlocked.
+  const issued = useRef(0);
+  const settled = useRef(0);
+
+  useEffect(() => {
+    mounted.current = true;
+    return () => {
+      mounted.current = false;
+    };
+  }, []);
+
+  const refresh = useCallback(async () => {
+    issued.current += 1;
+    const seq = issued.current;
+    try {
+      const { cores } = await api.listCores();
+      if (!mounted.current || seq < settled.current) return;
+      settled.current = seq;
+      setCount(cores.length);
+      setError(null);
+    } catch (err) {
+      if (!mounted.current || seq < settled.current) return;
+      settled.current = seq;
+      setError(err instanceof Error ? err.message : String(err));
+    }
+  }, []);
+
+  useEffect(() => {
+    void refresh();
+    const timer = setInterval(() => void refresh(), REGISTRY_POLL_MS);
+    // Pairing or forgetting a Core anywhere in this tab is the whole reason
+    // this number changes. Waiting out the poll for it would leave the gate
+    // visibly wrong for up to fifteen seconds.
+    const stop = onCoreRegistryChanged(() => void refresh());
+    return () => {
+      clearInterval(timer);
+      stop();
+    };
+  }, [refresh]);
+
+  return { count, error, refresh };
+}
--- a/packages/panel/src/components/views/FirstRunGate.tsx
+++ b/packages/panel/src/components/views/FirstRunGate.tsx
@@ -9,6 +9,8 @@ import {
 } from "react";
 import { api } from "~/lib/api";
 import { onCoreRegistryChanged } from "~/lib/core-registry-changed";
+import { readCachedCoreCount, writeCachedCoreCount } from "~/lib/shell-query-cache";
+import { CORES_POLL_MS } from "~/lib/use-fleet";
 
 // Lazy, for the same reason the settings overlay is: on every load of a Panel
 // that *has* a fleet — which is every load after the first — the wizard and the
@@ -41,16 +43,6 @@ const FirstRunWizard = lazy(() =>
  * creates their account first and meets the wizard immediately after.
  */
 
-/**
- * How often to re-ask, absent an announcement.
- *
- * The same fifteen seconds `useCores` polls on, and for the same reason: only
- * an operator changes this registry. The poll is the backstop for a change this
- * tab did not make; the announcement below is what makes the changes it *did*
- * make instant.
- */
-const REGISTRY_POLL_MS = 15_000;
-
 export function FirstRunGate({ children }: { children: ReactNode }) {
   const { count, error, refresh } = useCoreRegistry();
 
@@ -71,10 +63,11 @@ export function FirstRunGate({ children }: { children: ReactNode }) {
     await refresh();
   }, [refresh]);
 
-  // Not yet known, and no reason to think it is unknowable: draw nothing.
-  // Rendering the wizard here would flash it at every operator with a fleet on
-  // every load, and rendering the shell would flash a dashboard at operators
-  // with none — the two mistakes this component exists to prevent.
+  // Genuinely unknown: no live answer yet *and* nothing cached to stand in for
+  // one, which is the first load in this browser and no other. Drawing the
+  // wizard here would flash it at every operator with a fleet, and drawing the
+  // shell would flash a dashboard at operators with none — the two mistakes
+  // this component exists to prevent. Every other load paints on the seed.
   if (count === null && error === null) return null;
 
   // A registry we could not read is not a registry with no Cores in it, but it
@@ -99,22 +92,32 @@ export function FirstRunGate({ children }: { children: ReactNode }) {
  * needs none of the rows and cannot use a nonce, because the gate has to be
  * able to *wait* for a re-read — `onPaired` is a promise the pairing form is
  * holding its busy state on, and a nonce cannot be awaited. What is shared is
- * the thing that matters: both ask `api.listCores()`, so there is one answer
- * about the fleet and no second source of truth for it.
+ * the thing that matters: both ask `api.listCores()` and both pace themselves
+ * off `CORES_POLL_MS`, so there is one answer about the fleet, one cadence, and
+ * no second source of truth for either.
  *
- * `count` is null until a read succeeds, and never returns to null afterwards:
- * a poll that fails leaves the last known count standing, so a Panel with a
- * fleet does not get thrown into the wizard by one bad request.
+ * `count` starts at whatever this browser was last told (`readCachedCoreCount`)
+ * so a paired Panel paints its shell on the first client render, the same
+ * bargain `installShellQueryCache` makes for projects, groups and settings. The
+ * seed is never an answer: the live read lands on the same tick and corrects
+ * it, and a *stale* seed can only cost one frame in either direction.
+ *
+ * After that it never returns to null: a poll that fails leaves the last known
+ * count standing, so a Panel with a fleet is not thrown into the wizard by one
+ * bad request.
  */
 function useCoreRegistry(): {
   count: number | null;
   error: string | null;
   refresh: () => Promise<void>;
 } {
-  const [count, setCount] = useState<number | null>(null);
+  const [count, setCount] = useState<number | null>(() => readCachedCoreCount() ?? null);
   const [error, setError] = useState<string | null>(null);
 
   const mounted = useRef(true);
+  // What the last settled read said, readable from inside `refresh` without
+  // making it depend on — and be rebuilt by — the state it sets.
+  const known = useRef<number | null>(count);
   // Monotonic, so a slow poll that lands after a fast explicit refresh cannot
   // overwrite the newer answer with its older one — which, on this component,
   // would mean re-locking a dashboard that had just been unlocked.
@@ -132,11 +135,33 @@ function useCoreRegistry(): {
     issued.current += 1;
     const seq = issued.current;
     try {
-      const { cores } = await api.listCores();
+      let next = (await api.listCores()).cores.length;
+      /**
+       * Tearing down a live session takes two answers, not one.
+       *
+       * Locking the gate unmounts the whole shell: every open terminal panel,
+       * every websocket, whatever the operator was mid-way through. The gate
+       * already refuses to do that on a *failed* read; an empty successful one
+       * deserves the same suspicion, because a Panel server that restarts
+       * against an empty or unmigrated data directory answers 200 with nothing
+       * in it and would take the session down with no prompt and no undo.
+       *
+       * So an empty answer that would close the gate is re-asked, and only two
+       * consecutive empty successes lock it. This buys exactly what it says: a
+       * single anomalous 200 cannot end a session. It does not defend against a
+       * registry that is genuinely and persistently empty — nothing short of
+       * asking the operator could, and a Panel whose Cores are really gone
+       * belongs in the wizard.
+       */
+      if (next === 0 && (known.current ?? 0) > 0) {
+        next = (await api.listCores()).cores.length;
+      }
       if (!mounted.current || seq < settled.current) return;
       settled.current = seq;
-      setCount(cores.length);
+      known.current = next;
+      setCount(next);
       setError(null);
+      writeCachedCoreCount(next);
     } catch (err) {
       if (!mounted.current || seq < settled.current) return;
       settled.current = seq;
@@ -144,18 +169,53 @@ function useCoreRegistry(): {
     }
   }, []);
 
+  // Gated means "the wizard is what is on screen": no fleet known, or none at
+  // all. It decides how hard this hook works — see the two effects below.
+  const gated = count === null || count === 0;
+
   useEffect(() => {
     void refresh();
-    const timer = setInterval(() => void refresh(), REGISTRY_POLL_MS);
     // Pairing or forgetting a Core anywhere in this tab is the whole reason
-    // this number changes. Waiting out the poll for it would leave the gate
+    // this number changes. Waiting out a poll for it would leave the gate
     // visibly wrong for up to fifteen seconds.
-    const stop = onCoreRegistryChanged(() => void refresh());
-    return () => {
-      clearInterval(timer);
-      stop();
-    };
+    return onCoreRegistryChanged(() => void refresh());
   }, [refresh]);
+
+  /**
+   * The poll runs **only while the gate is up**.
+   *
+   * Once unlocked there is nothing here worth a standing timer: every registry
+   * change this tab makes arrives on the event above, and the shell that is now
+   * mounted is already reading the same endpoint through `useCores`. A third
+   * recurring `listCores()` for the life of the tab bought nothing but load.
+   */
+  useEffect(() => {
+    if (!gated) return;
+    const timer = setInterval(() => void refresh(), CORES_POLL_MS);
+    return () => clearInterval(timer);
+  }, [gated, refresh]);
+
+  /**
+   * What the dropped poll would have caught: a Core forgotten in *another* tab.
+   *
+   * The event is window-scoped, so it does not cross tabs, and the requirement
+   * is about the count rather than about which tab did the forgetting. Re-asking
+   * when this tab comes back to the front covers it at the only moment it
+   * matters — when someone is looking at this Panel — and costs nothing while
+   * they are not.
+   */
+  useEffect(() => {
+    if (gated) return;
+    const recheck = () => {
+      if (document.visibilityState === "visible") void refresh();
+    };
+    window.addEventListener("focus", recheck);
+    document.addEventListener("visibilitychange", recheck);
+    return () => {
+      window.removeEventListener("focus", recheck);
+      document.removeEventListener("visibilitychange", recheck);
+    };
+  }, [gated, refresh]);
 
   return { count, error, refresh };
 }
--- a/packages/panel/src/components/views/FirstRunWizard.tsx
+++ b/packages/panel/src/components/views/FirstRunWizard.tsx
@@ -1,15 +1,18 @@
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { Btn } from "~/components/ui/Btn";
 import { Icon } from "~/components/ui/Icon";
 import { TextField } from "~/components/ui/TextField";
 import { AddCoreByPairing } from "~/components/views/AddCoreByPairing";
 import {
-  CORE_INSTALL_PATHS,
+  ADD_CORE_LOCATION,
   PAIR_NEW_OUTPUT,
   composePairNewCommand,
+  coreInstallPaths,
+  labelRefusal,
   pairNewCommand,
 } from "~/shared/core-onboarding";
 import type { CoreWithDial } from "~/shared/cores";
+import { CURRENT_MC_VERSION } from "~/queries/mission-control-version";
 
 /**
  * The first-run pairing wizard (#358) — what a Panel with no Cores shows
@@ -237,12 +240,15 @@ function StepRail({ step, onStep }: { step: StepIndex; onStep: (next: StepIndex)
  * which is the same place any other surface would read them from.
  */
 function InstallStep() {
+  // The Compose line is pinned to this Panel's own line, so the Core it brings
+  // up is built from the same train — see `coreInstallPaths`.
+  const paths = coreInstallPaths(CURRENT_MC_VERSION);
   return (
     <StepBody
       title="Install the Core on the machine you want to drive"
       lede="A Core is the daemon that owns the repository and runs the Harnesses. It goes on the machine with your code — a laptop, a workstation, a build box — not on this Panel. Pick whichever path fits that machine."
     >
-      {CORE_INSTALL_PATHS.map((path) => (
+      {paths.map((path) => (
         <div key={path.id} data-install-path={path.id} style={{ display: "flex", flexDirection: "column", gap: 8 }}>
           <div style={{ fontSize: 13, fontWeight: 600 }}>{path.title}</div>
           <Prose>{path.blurb}</Prose>
@@ -261,11 +267,13 @@ function InstallStep() {
  *
  * The command is built from what the operator types, so `--label` is a thing
  * they have chosen rather than a flag they were shown and skipped. Under it,
- * the four lines `pair new` prints and what each one is for: three of them are
- * typed into step 3, and an operator who knows that reads the terminal once
- * instead of three times.
+ * every line `pair new` prints and what each one is for — including `Label`,
+ * which that command always produces because it always carries `--label`. An
+ * operator who knows which lines matter reads the terminal once instead of
+ * three times.
  */
 function MintStep({ label, onLabel }: { label: string; onLabel: (next: string) => void }) {
+  const refusal = labelRefusal(label);
   return (
     <StepBody
       title="Mint a pairing code on that machine"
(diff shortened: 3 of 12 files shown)


qcentic-dev:
## Acceptance criteria from #358

Every criterion in the issue, checked or explicitly not, with the reason.

- [x] **Fresh Panel: register Operator → wizard, no dashboard route until a Core exists.**
  `/setup` redirects to `/`, `FirstRunGate` reads `api.listCores()`, and at zero Cores it returns the wizard *instead
  of* `<Shell />` — so the top bar, the project rail, the router outlet and the settings overlay never mount and no
  route reaches a dashboard. `/login` and `/setup` still render outside the shell, untouched, so the Operator is
  created first and meets the wizard immediately after.
  Tests: *shows the wizard, and no dashboard at all, when it knows zero Cores*; *keeps the dashboard away on every step
  of the way through it*.

- [x] **Deleting the last Core brings the wizard back.**
  The gate is the live count, not a first-run flag — there is no "seen it" state anywhere in the change.
  `CoresSettingsPage` announces the registry change after a removal, so the gate re-reads at once rather than at its
  next 15-second poll.
  Test: *comes back when the last Core is forgotten, because the gate is the count*.

- [x] **Each wizard step displays its copyable Core-side command.**
  Step 1 — the installer one-liner and `actana setup`, and the Compose pair. Step 2 — `actana pair new --label <name>`,
  its Compose form, and the four lines it prints with what each one is for; the command is rewritten live (and shell
  quoted) as the operator names this Panel. Step 3 — the redemption itself, whose own text names `actana pair new`.
  Every command sits in a block with a copy button.
  Tests: *hands over both install paths on step 1*; *names the mint command on step 2, and what each line it prints is
  for*; *rewrites the mint command as the operator names this Panel*; *keeps every command copyable, since it is pasted
  on another machine*; plus 7 tests over the command strings themselves in `core-onboarding.test.ts`.

- [x] **Redeem step is the same `AddCoreByPairing` flow, not a fork of it.**
  Step 3 mounts that component directly, with the same `onPaired` contract `CoresSettingsPage` fulfils. No pairing
  logic is duplicated: the address-then-fingerprint-then-code ordering, the local mismatch refusal, the code-clearing
  on a refusal and every failure sentence are that component's, once.
  Tests: *is the Settings pairing form, not a second one* (asserts the ordering the component owns — no code box before
  a verified fingerprint); *posts the same body Settings posts*; *unlocks the dashboard on the first Core that pairs*;
  *stays put when the code is refused*.

- [x] **Settings → Cores → Add Core remains available and unchanged; the wizard is an additional path, and both name
  the same canonical location.**
  `AddCoreByPairing` is byte-identical — not touched by this PR. `CoresSettingsPage` gains exactly two
  `announceCoreRegistryChanged()` calls (after a pair, after a remove) and nothing else; its markup, its copy and its
  Add Core section are unchanged. Step 3 of the wizard names **Settings → Cores → Add Core** in so many words as where
  every Core after the first is paired.
  Test: *is the Settings pairing form, not a second one* also asserts that sentence is on screen.
  **One honest caveat:** while the gate is up — zero Cores — the settings overlay is not reachable, because the shell
  it lives in does not mount. That is the hard gate the issue asks for ("not skippable… nothing else to show"), and it
  ends the moment a Core pairs. The settings page itself is unchanged and available on the other side of it.

- [x] **Not skippable, no dismiss affordance.**
  Not a separate bullet in the issue's Acceptance list but the load-bearing sentence in *What to build*, so it is
  checked here explicitly. No skip, no "later", no dismiss, no close, no escape key, no click-outside, and no route
  past it — moving between steps 1, 2 and 3 reaches the dashboard exactly as often as standing still does.
  Test: *offers nothing that skips, dismisses or defers it* (scans every rendered button's text and `aria-label`).

### Not done, and why

- [ ] **Manual verification against a live Core.** Not run. The gate and the wizard are covered at the component level
  only — 16 tests in `first-run-gate.test.tsx` and 7 in `core-onboarding.test.ts`, on top of the full `packages/panel`
  suite (146 files / 1514 tests) and `pnpm -r typecheck`, all green. No browser was driven and there are no
  screenshots.
- [ ] **Documentation.** No `README.md`, `INSTALL.md`, `CONTEXT.md` or ADR changes. Out of scope by instruction: this
  branch is confined to `packages/panel`. Every command the wizard prints already comes from those documents, and
  `src/shared/core-onboarding.ts` says so in its header so the two stay in step.
- [ ] **CLI surfaces (#357, #360).** Untouched by design — `packages/cli` has no diff in this branch. The wizard shows
  the `pair new` output as the four labelled lines the CLI prints today; #357's framed TTY block presents the same four
  facts, so nothing here needs to change when it lands.
- [x] **`Train rules` check.** Green, and correctly so — I expected otherwise and was wrong. A base that is neither
  `main`, a `beta/x.y.z` train nor a `release/*` line falls to that job's `*)` arm, which reports *"Nothing in the
  branch model constrains it"* and passes. Targeting `feat/0.4.3-onboarding` as instructed costs this PR nothing;
  whatever branch collects the 0.4.3 onboarding work will face the train rule when *it* is proposed.

mrfqcentic (changes requested):
Reviewed against #358, commit `f8cfbf0`, diffed against `feat/0.4.3-onboarding` (`fac1010` — the base currently carries no commits of its own, so this is effectively a diff against `beta/0.4.3`).

**Verdict: request changes.** The architecture is right and four of the five acceptance criteria are met outright. What blocks it is that the wizard's whole job is to teach commands for a machine the Panel cannot see, and two of the things it teaches are wrong: the Compose path starts a second Panel, and the `pair new` output block does not match what `pair new` prints. Those are small fixes on an otherwise well-built change.

I verified the claims below by reading the enforcing lines, not the description. Everything the description says about reuse and scope checks out.

---

## What I confirmed

- **The gate is hard, and it is the live count.** `FirstRunGate.tsx:83-91` returns `<FirstRunWizard>` *instead of* `{children}`, and `__root.tsx:163-165` passes `<Shell />` as those children. `Shell` owns the only `<Outlet />` for non-auth routes (`__root.tsx:627`), so at zero Cores no route renders — `/settings` (`routes/settings.tsx`) is a deep-link route through that outlet and is genuinely unreachable, not merely hidden. `isAuthPath` (`lib/auth-paths.ts:5-7`) is `/login` and `/setup` only, and `__root.tsx:110-123` returns those before the gate exists, so the gate correctly does not stand in front of account creation. No first-run flag anywhere: `useCoreRegistry` (`FirstRunGate.tsx:109-161`) is a live `api.listCores()`, and `first-run-gate.test.tsx:198-212` drives the count back to zero and asserts the wizard returns.
- **A successful pairing unlocks the dashboard, proven through the real component.** `first-run-gate.test.tsx:309-317` mounts the gate at zero Cores, runs `pair()` (`:113-123`) — address → `Check fingerprint` → compare → session + code → `Pair Core` — through the unmocked `AddCoreByPairing`, with only `~/lib/api` stubbed, and asserts the child renders and the wizard is gone. That is a real end-to-end redemption, not a stub. Good test.
- **It reuses the pairing flow rather than forking it.** `AddCoreByPairing.tsx` has zero diff. `CoresSettingsPage.tsx` diffs by one import and two `announceCoreRegistryChanged()` calls (`:79`, `:121`) plus comments — nothing else. `FirstRunWizard.tsx:339` mounts the same component with the same `onPaired` contract.
- **Scope.** All nine files are under `packages/panel`. Nothing else is touched.
- **The `--label` machinery.** `pairNewCommand`/`labelArgument` (`core-onboarding.ts:125-148`) is correct, including the POSIX `'\''` escape, and the live rewrite is tested (`first-run-gate.test.tsx:253-262`).
- **The install commands themselves** match `README.md:81-82` and `:100-101` verbatim, and the Compose exec prefix matches `README.md:83`, including the `core` service name (`deploy/docker-compose.yml:71`).
- **The `issued`/`settled` monotonic guard** (`FirstRunGate.tsx:118-145`) is correct — a slow in-flight read cannot re-lock a dashboard a newer read unlocked. CI is green on lint, typecheck and unit tests.

---

## Blocking

### 1. `packages/panel/src/shared/core-onboarding.ts:56-59` — the Compose path starts a *second Panel* at the operator reading this screen

```
git clone https://github.com/actana/control && cd control
docker compose -f deploy/docker-compose.yml up -d
```

`deploy/docker-compose.yml:44` defines a `panel` service publishing `127.0.0.1:7420:7420`, alongside `core` at `:71`. `README.md:77` is explicit that this file is "a Panel and a Core on one network, which is the whole product in one command" — which is the right command in the README, where the reader has no Panel yet, and the wrong one here, where by definition they are already looking at one.

Two outcomes, both bad. If their Panel *is* this compose file, then `core` is already up and step 1 was never needed. If it is not, this clones the repo and brings up a competing Panel on 7420 that either fails to bind or shadows theirs — and the blurb "The Core comes up on the same network as the Panel you are looking at" is false.

`docker compose -f deploy/docker-compose.yml up -d core` is the command this step wants.

Related, same block: the printed command resolves `${ACTANA_TAG:-latest}`, and `README.md:69-74` says `:latest` 404s until the first release is published, telling operators to use `ACTANA_TAG=beta-0.1.0` for now. The wizard prints a command that does not work today. If the pre-release caveat is going to outlive this PR, the wizard needs to carry it too.

### 2. `packages/panel/src/shared/core-onboarding.ts:90-114` — `PAIR_NEW_OUTPUT` does not match what `pair new` prints

The module's stated purpose is that "the person watching the Panel recognises the four facts when they scroll past on the Core's terminal". Three ways it won't:

- **A fifth line is missing.** `actana-pair.ts:274` prints `Label          ${label}` whenever `--label` is given — and the wizard *always* teaches `--label`. The CLI's own test pins the set for exactly this invocation: `packages/cli/src/__tests__/actana-pair.test.ts:213-214` runs `["new", "--label", "laptop"]` and asserts every stdout line matches `/^(Pairing code|CA fingerprint|Expires|Label|Session) /`. The wizard says "the four lines it prints" and explains four; the command it hands over prints five, and the unexplained one sits between `Expires` and `Session`.
- **`Session` is the wrong shape** (`:111`). The sample is `ps_7f2c1a9e`; the real value is `randomUUID()` (`actana-pair.ts:246`, printed at `:279`) — a 36-character UUID. No `ps_` prefix exists anywhere in the repo. An operator scanning the terminal for a short prefixed token will not find one, and a short sample invites truncating a UUID into step 3's Session box, which then fails redemption for a reason nothing on screen explains.
- **`Expires` is the wrong format** (`:105`). The sample is `2026-08-31 14:32 UTC (…)`; `absoluteTime` (`actana-pair.ts:786-788`) emits ISO-8601 — `2026-08-31T14:32:07Z (in 5 minutes)`.

The values being "obvious fakes" is the right call and I'm not asking for real ones — the *shapes* are what recognition runs on.

### 3. `packages/panel/src/shared/__tests__/core-onboarding.test.ts:58-66` — the test that would have caught #2 cannot fail

`it("explains every line \`pair new\` prints…")` asserts `PAIR_NEW_OUTPUT.map(l => l.label)` against a literal copy of `PAIR_NEW_OUTPUT`'s own contents. It is a tautology: it pins the module to itself and says nothing about the CLI. That is why the `Label` omission survived. With #357 about to reframe this exact output, the drift this file exists to prevent will be silent.

Assert against the CLI's label set instead — `actana-pair.test.ts:214` already has it — or hoist the labels somewhere both packages read.

### 4. `packages/panel/src/components/views/FirstRunGate.tsx:78` — every load of every Panel now blanks until `listCores()` answers

`if (count === null && error === null) return null;` puts a network round-trip in front of the entire shell. `lib/shell-query-cache.ts` and the comment at `__root.tsx:139-151` exist precisely so projects, groups and settings hydrate synchronously from localStorage and the shell paints on the first client render; the gate is now upstream of all of it and has no seed of its own.

The reasoning in the docblock is sound — flashing a dashboard at an operator with no Cores is the worse failure — but the cost lands on 100% of loads to protect the first one, and on a Panel reached over a slow link that is a visibly blank page every time. Seeding the initial count from a localStorage-cached value (still corrected by the live read, and still refusing to unlock on a *stale* zero) keeps the guarantee without the blank.

---

## Non-blocking, but worth a decision

### 5. `FirstRunGate.tsx:138` — an empty *successful* read tears down a live session with no confirmation

The gate is careful never to re-lock on a failed read (`:140-144`, and `first-run-gate.test.tsx:214-226` proves it). An empty successful one gets no such guard: `setCount(cores.length)` trusts any 200 absolutely, and the poll runs for the life of the tab. If the Panel server ever restarts against an empty or unmigrated data directory, within 15s `Shell` unmounts — every open terminal panel, every websocket — and the operator lands in the wizard mid-session with no prompt and no undo. Two consecutive zero reads, or declining to tear down while terminals are attached, would cost very little.

### 6. `FirstRunGate.tsx:83` — at zero Cores there is no sign-out and no settings at all

Sign out lives only in `OperatorSettingsPage.tsx:97` (`api.logout()` at `:43`), reachable only through the settings overlay, which `Shell` mounts. So does `<UpdateBanner />` (`__root.tsx:564`). #358 asked for exactly this ("not skippable"), so I'm not calling it a defect — but the consequence deserves to be a decision someone made on purpose: an operator who registers on a shared machine, or forgets their last Core from a decommissioned box, cannot log out, cannot change their password, and cannot be told the Panel needs updating. The only exit is pairing a Core, which may be the thing they cannot do.

### 7. `first-run-gate.test.tsx:180-187` — this test does not prove what the description says it proves

The PR body says "no button anywhere matches skip / dismiss / later / close / continue-anyway". The test scans `getAllByRole("button")` immediately after `mount()`, so it sees **step 1 only** — never step 2's or step 3's buttons — and it is a keyword blocklist, which cannot establish absence ("Go to Panel", "Ignore", "Maybe next time" all pass).

The hardness is real, but it is proven structurally (the gate returns the wizard *instead of* children) and by `:189-196`, which walks 1→2→3→2→1 asserting the dashboard is absent at each stop. That is the test doing the work. Either walk all three steps in the escape scan too, or lean on `:189-196` and soften the claim in the description.

### 8. `CoresSettingsPage.tsx:79,121` — the announce calls have no test

`first-run-gate.test.tsx:198-212` calls `announceCoreRegistryChanged()` directly, so the remove → announce → wizard path is proven in two halves that never meet. Nothing would catch either call being dropped. Low risk — the 15s poll still recovers, so it degrades to latency — but `__tests__/cores-settings-pairing.test.tsx` already exists and is the natural home.

### 9. `core-onboarding.ts:146` — a label starting with `-` prints a command the Core refuses

`-` is inside the unquoted allowlist, so `-panel` is emitted bare and `actana-pair.ts:683` refuses it: `if (value === undefined || value.startsWith("-")) return { error: '--label needs a value' }`. Exactly the "a failure the operator would blame on the Panel, having pasted exactly what it printed" case the docblock says this function exists to prevent.

Worth noting the two obvious remedies both fail: quoting doesn't help (the shell strips quotes before the CLI sees the token), and `--label=-panel` hits the *same* guard, because `actana-pair.ts:682-683` applies the `startsWith("-")` check to the `=`-form value too. So this has to be handled on the Panel side — reject or strip a leading `-` in the name box.

### 10. `CoresSettingsPage.tsx:122` — removing the last Core destroys its own confirmation toast

`<Toaster>` is rendered inside `Shell` (`__root.tsx:651`). `handleRemove` announces at `:121` and then raises `toast.success(...)` at `:122`; a tick later the gate's re-read lands, `Shell` unmounts with the `Toaster` inside it, and the confirmation vanishes before it can be read. Only cosmetic today, but it also means any toast added to `AddCoreByPairing` later will work in Settings and be silently dropped in the wizard.

### 11. Polling

`REGISTRY_POLL_MS` (`FirstRunGate.tsx:52`) hand-copies `CORES_POLL_MS` (`lib/use-fleet.ts:28`); the comment says "the same fifteen seconds" but nothing keeps them the same. And the poll at `:149` runs forever once unlocked, adding a third recurring `listCores()` alongside `use-fleet`'s two — purely to catch a drop to zero that the event already covers for every in-tab change. Polling only while gated would be cheaper and no less correct.

### 12. `FirstRunWizard.tsx:341` — "Settings → Cores → Add Core", but the field is labelled "Add a Core"

`CoresSettingsPage.tsx:167` is `<Field label="Add a Core">`. #358 asks that both paths "name the same canonical location"; they name the same place with two strings. One word, either direction.

### 13. `FirstRunWizard.tsx:441` — the copy-confirmation `setTimeout` is never cleared

Click copy on step 1, hit Next inside 1.5s, and the timer fires `setCopied` on an unmounted `CommandLine`. Harmless under React 18, but it's a live timer holding a closure over a dead tree.

---

## Acceptance criteria (#358)

| # | Criterion | |
|---|---|---|
| 1 | Fresh Panel: register Operator → wizard, no dashboard route until a Core exists | **Met** — structurally guaranteed; router wiring verified by reading `__root.tsx:110-165`, not covered end-to-end by a test (the PR says so honestly) |
| 2 | Deleting the last Core brings the wizard back | **Met** — `FirstRunGate.tsx:83`, `first-run-gate.test.tsx:198-212` |
| 3 | Each wizard step displays its copyable Core-side command | **Partially met** — steps 1 and 2 do and both are copyable; step 3 correctly has none. But step 1's Compose command is wrong for this audience (#1) and step 2 misdescribes what its command prints (#2) |
| 4 | Redeem step is the same `AddCoreByPairing`, not a fork | **Met** — zero diff on that file; `FirstRunWizard.tsx:339` |
| 5 | Settings → Cores → Add Core remains available and unchanged; both name the same location | **Met** — nine added lines in `CoresSettingsPage.tsx`, no behaviour change; wording nit at #12 |

## The flagged design decisions

- **Own `api.listCores()` rather than `useCores`** — justified. `onPaired` must be awaitable so the form holds its busy state until the gate has genuinely unlocked, and `useCores` refreshes on a nonce that cannot be awaited. Both call the same endpoint, so there is still one source of truth. The cost is the duplicated interval (#11) and the blank first paint (#4).
- **Payload-free window event** — right call, and the reasoning is right: a count riding the event would be a second source of truth for the one number the gate must not be wrong about. It matches the existing `design-meta.ts` event conventions, and the two announce sites are the only two registry mutations in the tree (`api.pairCore`, `api.removeCore`), so coverage is complete. Untested, though (#8).
- **Settings unreachable while gated** — this *is* the hard gate #358 asked for and I'm not asking for it to be softened. It is correctly implemented rather than faked with CSS. But see #6: "no way to sign out" is a real consequence that should be an explicit decision rather than a side effect, and #5 is the case where that gate closes on someone who did nothing wrong.

Not verified: I did not run the app or the suite locally (read-only clone). CI's green lint/typecheck/unit runs stand in for the PR's local-pass claims.

Nice work overall — the gate's async handling is more careful than most code of this kind, and step 3 mounting the real form rather than re-implementing it is the right instinct on the flow where it matters most.

qcentic-dev:
## Review response — `73ef133`

Thank you for reading the enforcing lines rather than the description; two of the four blockers were things this branch
asserted about itself and no test could contradict.

All four blockers are fixed, along with six of the non-blocking findings. One is **deferred to the project owner** and
answered below. Two are acknowledged and not done, also below.

---

### Blocking

**1 — the Compose path started a second Panel.** Fixed. The command is now
`ACTANA_TAG=beta-<version> docker compose -f deploy/docker-compose.yml up -d core`, and the blurb no longer claims the
Core "comes up on the same network as the Panel you are looking at" — it says the file also defines a `panel` service
and that this brings up the Core alone.

On the tag: rather than carry the README's pre-release caveat as prose that will rot, the wizard prints a resolvable
tag derived from `CURRENT_MC_VERSION` — the Panel's own build stamp — so the Core it brings up is built from the same
line as the Panel pairing it, which is what `ACTANA_TAG` exists for ("one variable, both services"). The note names the
tag and says that once a release exists the line's tag is the bare version. Nothing is hard-coded, so it cannot go
stale against the train it ships on. Tests: *brings up the Core service alone, never a second Panel*, *pins an image
tag rather than falling through to a `:latest` that 404s*, *tracks the Panel's own line rather than a hard-coded
version*, plus a rendered assertion that a bare `up -d` is nowhere on step 1.

**2 — `PAIR_NEW_OUTPUT` did not match the printer.** Fixed, all three.

- `Label` added, between `Expires` and `Session`, in the printer's order. You are right that it is not optional here:
  the wizard always teaches `--label`, so `pair new` always prints it.
- `Session` is now `7f2c1a9e-4b60-4c3d-9f21-8e5a7c0d1b34` — 36 characters, no prefix — and its meaning says "whole",
  which is the failure a short sample was inviting.
- `Expires` is now `2026-08-31T14:32:07Z (in 5 minutes)`, matching `absoluteTime`.

Values are still obvious fakes, as you said they should be.

**3 — the tautological test.** Fixed, and this is the one worth reading. `core-onboarding.test.ts` now reads the
contract out of `packages/cli` and cannot be satisfied by editing this package:

- the **label set** comes from the assertion your review pointed at — the test parses
  `/^(Pairing code|CA fingerprint|Expires|Label|Session) /` out of `actana-pair.test.ts` and compares it to
  `PAIR_NEW_OUTPUT`'s labels;
- the **order** comes from the printer, by scanning `actana-pair.ts`'s `deps.out` calls (`Endpoint host` filtered out,
  since it appears only for `--public-host`, which this wizard does not teach);
- the **shapes** are asserted as regexes — ISO-8601 with a relative tail, and a 36-character UUID;
- and `labelRefusal`'s justification is read from the CLI too: the suite asserts `value.startsWith("-")` is still in
  that parser, so if the guard is ever relaxed the Panel-side rule stops being load-bearing loudly rather than quietly.

Both files are read through `readCli()`, which throws if the file is empty or missing — a check that silently stops
checking is worse than none. I confirmed it bites by deleting the `Label` entry and re-running: three tests fail. When
#357 reframes that output, this suite is what goes red.

Hoisting the labels into a package both read would be better still, but that means `packages/shared`, and this branch
is scoped to `packages/panel`.

**4 — the blank first paint.** Fixed. The count seeds from `readCachedCoreCount()`, a new key alongside the projects,
groups and settings caches in `shell-query-cache.ts` and using the same versioned envelope. A Panel that has seen a
fleet paints its shell on the first client render; the live read lands and corrects it, including *back into the
wizard* if the seed is stale — there is a test for exactly that (*keeps the seed honest — the live read corrects it*).
`return null` now means only "no live answer and nothing cached", which is a first-ever load in this browser and
nothing else.

The instant-paint test deliberately makes no `act` flush: it asserts the first render, not the render after a promise
settles.

### Non-blocking, fixed

- **5 — an empty successful read tears down a live session.** Two consecutive empty successes are now required before
  the gate closes on a live shell; a single anomalous 200 cannot end a session. Stated honestly in the code: it does
  not defend against a registry that is genuinely and persistently empty, and nothing short of asking the operator
  could. Test asserts both reads happen and the dashboard survives the first.
- **7 — the escape scan saw step 1 only.** Now walks all three steps, with a comment saying what a keyword blocklist
  can and cannot establish, and the PR description has been corrected to match rather than the other way round.
- **8 — the announce calls had no test.** Three added to `cores-settings-pairing.test.tsx`: pairing announces, removal
  announces, and a **failed** removal announces nothing. (That file now wraps its render in a `QueryClientProvider` —
  `ConfirmDialog` renders a `CardFrame` whose glow reads settings through react-query, so opening the removal dialog
  needed one.)
- **9 — a leading `-` in the label.** Caught Panel-side, as you said it has to be. `labelRefusal()` names the reason;
  the command block prints `--label <name>` rather than the refused form, and the reason appears under the field as an
  alert. Your note that neither quoting nor `--label=` helps is quoted in the docblock, so nobody re-litigates it.
- **11 — polling.** `CORES_POLL_MS` is exported from `use-fleet.ts` and imported, so there is one cadence and no
  comment claiming two constants agree. The poll now runs **only while the gate is up**. Once unlocked, in-tab changes
  arrive on the event as you said — and because the event is window-scoped and the requirement is about the count
  rather than about which tab did the forgetting, an unlocked gate also re-reads on focus/visibility. That covers the
  cross-tab case the dropped poll was catching, at the only moment it matters, and costs nothing while nobody is
  looking.
- **12 — two strings for one place.** `ADD_CORE_FIELD_LABEL` / `ADD_CORE_LOCATION` in `core-onboarding.ts`; the
  settings `<Field>` and the wizard's redeem step both read them.
- **13 — the copy timer.** Held in a ref and cleared on unmount, and on a re-click.

### Deferred — 6, no sign-out and no update banner at zero Cores

Not implemented, deliberately, and I want to be straight about why: you are right that it should be a decision someone
owns rather than a side effect, and it is not mine to make. **The orchestrator is putting it to the project owner.**
Until they rule, it stays as built for this train — the hard gate is what #358 specifies in as many words ("not
skippable… a Panel that is not paired with a Core has nothing else to show"), and softening it on a reviewer's flag
plus my own judgement would be quietly overriding the issue.

The consequence you name is real and worth the owner seeing stated plainly: an operator who registers on a shared
machine, or forgets their last Core from a decommissioned box, cannot sign out, cannot change their password, and
cannot be told the Panel needs updating — and the only exit is the thing they may be unable to do. Finding 5 is the
case where that closes on someone who did nothing wrong, which is part of why 5 is fixed regardless of how 6 is ruled.

If the owner wants it softened, the smallest change that keeps the gate hard is a sign-out control and `<UpdateBanner>`
in the wizard's own chrome — neither is a route to the dashboard — and that is a follow-up ticket, not a change to
this branch.

### Acknowledged, not done

- **10 — the removal toast destroys itself.** Confirmed as described: `<Toaster>` lives inside `Shell`, so the
  confirmation for forgetting the *last* Core is unmounted a tick after it is raised. The honest fix is to hoist the
  `Toaster` out of `Shell` in `__root.tsx` so it survives the gate, which is a change to shell-wide rendering that
  belongs in its own commit with its own reasoning rather than riding a review pass. Cosmetic today; filed for
  follow-up. Reordering the two lines would not fix it — the `Toaster` still unmounts.
- Hoisting the `pair new` labels into a package shared by CLI and Panel (from 3), for the scope reason above.

---

**Verification:** full `packages/panel` suite **146 files / 1535 tests**, all passing; `pnpm -r typecheck` clean;
`pnpm lint` 0 errors with the warning count unchanged at 46, none in files this branch touches. Everything is still
inside `packages/panel` — `packages/cli` is read by a test and not modified. Still a draft, still based on
`feat/0.4.3-onboarding`, nothing merged.

One correction to my own earlier checklist, unrelated to your findings: I had predicted `Train rules` would be red on
this base. It is green — a base that is neither `main`, a train, nor a release line falls to that job's `*)` arm.

mrfqcentic (approved):
Re-reviewed at `73ef133`. **All four blockers are fixed, and every non-blocking item you claimed fixed is fixed.** Approving.

I verified each against the enforcing lines rather than the response, and where the fix turned on a claim about another package I ran the extraction myself rather than trusting that it works.

---

## Blockers

### 1 — the Compose path ✅

`core-onboarding.ts:151-161` now emits `ACTANA_TAG=beta-<version> docker compose -f deploy/docker-compose.yml up -d core`, and the blurb no longer claims the Core comes up beside the Panel you are looking at — it names the `panel` service and says this brings up the Core alone. `core-onboarding.test.ts:183-193` pins `endsWith("up -d core")` *and* asserts no bare `up -d` survives via `/up -d(\s|$)(?!core)/`; `first-run-gate.test.tsx:311-320` asserts the same against the rendered step and that the old string is absent. Both halves, string and screen.

**On the derivation — I went looking for a reason to dislike it and did not find one.** `coreImageTag(v) = beta-${v}` off `CURRENT_MC_VERSION` is the right choice, and it is right for a reason stronger than "not hard-coded":

- `ci.yml:1295-1307` publishes a train's images under `tags=beta-$version` where `version="${BRANCH#beta/}"`. The train is `beta/0.4.3`; `packages/panel/package.json:3` is `0.4.3` because the cut writes the version into every manifest in its first commit. `vite.config.ts:15` stamps `__MC_VERSION__` from exactly that. So `beta-0.4.3` is not an approximation of the published tag — it *is* the published tag, and it stays right on every future train with no edit.
- It also buys the thing `ACTANA_TAG` exists for. `deploy/docker-compose.yml`'s "one variable, both services" is a version-lock, and pinning the Core to the Panel's own line is that lock honoured on the one screen where an operator is about to introduce a Core to a Panel.
- And it survives promotion: `container-image.yml:440` states `beta-x.y.z`, `x.y.z`, `latest` and `x.y.z-beta` are "the same bytes and the same line", and `container-image.yml:237` names `beta-0.1.0` as a tag the dev-tag sweep must *never* delete. So a released Panel still prints a tag that resolves, and the note tells the operator the release tag is the bare version.

That is a better answer than the prose caveat I would have accepted. One residual, dev-only and not worth a change: `mission-control-version.ts:14-15` falls back to `"0.0.0"` when `__MC_VERSION__` is undefined, so an unbuilt Panel prints `ACTANA_TAG=beta-0.0.0`, which resolves to nothing. Nobody pairs a Core from `pnpm dev` via Compose, so I would leave it.

### 2 — `PAIR_NEW_OUTPUT` against the printer ✅

Checked line by line against `actana-pair.ts:271-279`:

- `Label` added at `core-onboarding.ts:172`, **between `Expires` and `Session`** — the printer's order, confirmed below by extraction rather than by eye.
- `Session` is `7f2c1a9e-4b60-4c3d-9f21-8e5a7c0d1b34` — 36 characters, no prefix, matching `randomUUID()` (`actana-pair.ts:246`). The meaning now says "whole", which is the right place to spend a word.
- `Expires` is `2026-08-31T14:32:07Z (in 5 minutes)`, matching `absoluteTime` (`actana-pair.ts:786-788`) plus `relativeTime`'s tail.

The `CA fingerprint` meaning gaining "as 32 colon-separated bytes" is a real improvement I did not ask for — it matches `actana-pair.test.ts:148`'s `/^[0-9A-F]{2}(:[0-9A-F]{2}){31}$/`.

### 3 — the tautological test ✅, and the approach is acceptable

**I ran both extractors against the real CLI files rather than taking the claim on trust.** Results:

```
cliPinnedLabels() -> [Pairing code, CA fingerprint, Expires, Label, Session]
candidate /^(…)/ literals in actana-pair.test.ts: 1   (no ambiguity today)
raw deps.out labels  -> [Pairing code, CA fingerprint, Expires, Label,
                         Endpoint host, Session, Unpaired, Cancelled the pending code for]
printerOrder()       -> [Pairing code, CA fingerprint, Expires, Label, Session]
```

Both work, `Endpoint host` is correctly filtered as documented, and the `pairLs`/`pairRevoke` lines the printer scrape also picks up are filtered by `pinned.includes`. `PAIR_NEW_OUTPUT`'s order is that list exactly. Your "delete `Label`, three tests fail" reproduces on inspection — `:81`, `:85` and `:90` all key on it.

**Judged as asked: reading sibling-package source from a panel test is acceptable here.** Reasons, in the order I weighed them:

- It is a real cross-package pin. Neither expectation can be satisfied by editing `packages/panel`, which was the entire defect in the first version, and `readCli` (`:41-48`) fails loudly on a moved file (`readFileSync` throws) or an emptied one (the explicit check) rather than quietly ceasing to check. Both extractors throw with an instructive message on no-match. That is the right failure mode.
- Nothing is imported at runtime and `packages/cli` is not modified, so this creates no production coupling — only a test-time read of a file that is always present in a repo checkout.
- The alternative you name — hoisting the labels into `packages/shared` — is genuinely better and genuinely out of scope for a `packages/panel`-only branch. Choosing the pin that fits the scope over leaving a tautology in place is the right call, and you said so rather than dressing it up.

The cost, recorded so nobody is surprised later: these are **text scrapes, not a structural contract**. They cover "the file moved" and "no match at all"; they do not cover "matched the wrong literal". If `actana-pair.test.ts` ever gains an earlier `/^(…)/ ` regex, `cliPinnedLabels()` will silently pick it. Today there is exactly one, so the risk is latent rather than live. And CLI refactors can now redden a panel suite — arguably the point, but it is a tax a shared constant would not levy. Worth a follow-up ticket alongside the hoist, not a change here.

### 4 — the blank first paint ✅

`useState(() => readCachedCoreCount() ?? null)` (`FirstRunGate.tsx:112`), seeded from a new `coreCount` key using the same versioned envelope as projects/groups/settings (`shell-query-cache.ts:12-21`, `:94-103`), with a read guard that rejects anything not a non-negative integer.

**The instant-paint test does assert the first render** — I checked this specifically, since a stray `act` would have made it worthless. `first-run-gate.test.tsx:173-188` writes the seed, points `listCores` at a promise that *never settles*, calls `render(...)` directly rather than the `mount()` helper, and asserts with **no flush at all**. The dashboard can only be on screen because of the seed. The comment at `:182-183` says exactly that. Good test.

**A stale seed cannot lock the wizard out**, in either direction:

- `:190-197` — cache says 3, registry says 0 → wizard, no dashboard.
- `:181-188` — cache says 0 → wizard on first render, so a Panel that lost its fleet does not flash a shell either.
- And `known` (`:118`) is seeded from the same value, so a stale positive count goes through the two-read guard and still lands in the wizard.

`return null` now means only "no live answer and nothing cached", which is a first-ever load and nothing else — exactly the narrow case that justified it.

One behavioural change worth being aware of rather than fixing: with a cached positive count and a live read that *fails*, the gate now renders the shell (count holds the stale seed, error is set) where it used to render the wizard with an error box. For a Panel that has a fleet that is the better answer — throwing a paired Panel into a pairing wizard because one request failed was the worse behaviour — and an uncached browser still gets the old path (`:200-209` still passes). Noting it, not asking for anything.

---

## Non-blocking, verified

- **5.5 — two consecutive empty successes** ✅ `FirstRunGate.tsx:138-160`, and `first-run-gate.test.tsx:263-283` asserts `before + 2` calls with the dashboard surviving. The docblock is honest that this does not defend against a genuinely empty registry, which is the correct scope. Note the guard only arms when `known.current > 0`, so the pairing path costs no extra request.
- **5.8 — three announce tests** ✅ `cores-settings-pairing.test.tsx:479-545`: pairing announces once, removal announces once, **and a failed removal announces nothing**. They click the real UI and listen on the real window event, so `remove → announce → wizard` is now proven end to end rather than in two halves. The third is the one I would not have thought to ask for and is the best of the three. The `QueryClientProvider` addition is explained where it is made.
- **5.9 — dash-label refused Panel-side** ✅ `labelRefusal` (`core-onboarding.ts:199-206`) with `labelArgument` falling back to the placeholder (`:236`), a `role="alert"` under the field (`FirstRunWizard.tsx:119-132`) and `ariaInvalid` on the input. `core-onboarding.test.ts:155-159` reads `value.startsWith("-")` back out of the CLI, so if that guard is relaxed this stops being load-bearing loudly. Quoting the reasoning into the docblock so nobody re-litigates it is the right instinct.
- **5.11 — one cadence, gated-only polling** ✅ `CORES_POLL_MS` exported from `use-fleet.ts:28-35` and imported (`FirstRunGate.tsx:13`); no bundle cost, since `use-fleet` is already eager through `ProjectBar`/`ProjectPicker`. The poll effect is `if (!gated) return` (`:181-185`) and the focus/visibility re-read is `if (gated) return` (`:196-209`) — cleanly disjoint, both cleaned up.
  One coverage note: a Core forgotten in *another tab* while this tab stays focused is now caught only at the next focus/visibility transition, where the dropped poll would have caught it within 15s. Your reasoning — it matters only when someone is looking — holds, and I agree with the trade. If you ever want the gap closed for free, `writeCachedCoreCount` already writes to localStorage, and a `storage` listener on that key fires in exactly the other tabs. Follow-up, not now.
- **5.12 — one string for one place** ✅ `ADD_CORE_FIELD_LABEL` / `ADD_CORE_LOCATION` (`core-onboarding.ts:175-176`), read by both `CoresSettingsPage.tsx:177` and `FirstRunWizard.tsx:368`, with the assertion at `first-run-gate.test.tsx:393` updated to the constant rather than to a literal that would have drifted again.
- **5.13 — timer cleared** ✅ `FirstRunWizard.tsx:463-471`, held in a ref, cleared on unmount and on re-click.
- **5.7 — escape scan on all three steps** ✅ `first-run-gate.test.tsx:221-237` now loops 1/2/3, so step 3's `AddCoreByPairing` buttons are in scope. More to the point, the comment now states plainly what a keyword blocklist can and cannot establish and points at the structural test as the one doing the work — and you corrected the PR description to match the test rather than the reverse. That is the right direction to resolve that kind of gap.

## 5.6 — deferred, and correctly

Staying as built pending the project owner's ruling, per instruction; I am not blocking on it and have not re-argued it. For the record, the way it is being handled is right: #358 specifies the hard gate in as many words, and softening it on a reviewer's flag plus an implementer's judgement would be overriding the issue quietly. Your note that fixing 5.5 regardless of how 6 is ruled is the part that protects someone who did nothing wrong is exactly the correct decomposition.

## 5.10 — acknowledged, and I agree with not doing it here

Hoisting `<Toaster>` out of `Shell` is a change to shell-wide rendering that deserves its own commit and its own reasoning, not a ride on a review pass. Your observation that reordering the two lines would not help is correct — the `Toaster` unmounts either way.

---

## Checklist

| | |
|---|---|
| 4.1 Compose starts a second Panel | ✅ fixed — `up -d core`, tag derived from the train, both string- and screen-tested |
| 4.2 `PAIR_NEW_OUTPUT` vs the printer | ✅ fixed — `Label` present and in order, 36-char UUID, ISO-8601 |
| 4.3 tautological test | ✅ fixed — reads the CLI's own contract; extractors verified by running them |
| 4.4 blank first paint | ✅ fixed — seeded, first-render asserted with no flush, stale seed corrected both ways |
| 5.5 two consecutive empty successes | ✅ |
| 5.7 escape scan on all steps | ✅ |
| 5.8 announce calls tested | ✅ (three, including the negative) |
| 5.9 dash-label refused Panel-side | ✅ |
| 5.11 shared cadence, gated-only poll | ✅ |
| 5.12 one location string | ✅ |
| 5.13 timer cleared | ✅ |
| 5.6 no sign-out at zero Cores | ⏸ deferred to the project owner, as instructed |
| 5.10 removal toast | 📋 acknowledged, follow-up |
| Scope | ✅ cumulative diff `fac1010…73ef133` still entirely inside `packages/panel` |
| CI at `73ef133` | ✅ Lint, Typecheck, Unit Tests, Conventions, Train rules all green |

**Approved.** Acceptance criterion 3 — the one I had as partially met — is now met: the commands the wizard teaches are commands that work, and the block describing `pair new`'s output describes what `pair new` prints. All five criteria are met.

Two things to carry forward as tickets rather than changes here: hoisting the `pair new` labels into a package both the CLI and the Panel read (which retires the text scrape along with it), and the `Toaster` hoist. Still a draft on `feat/0.4.3-onboarding`, and I have left it that way.

PR #364, opened 2026-08-31
feat(cli): frame `actana pair new` at a terminal, byte-identical down a pipe, and fix two usage strings

qcentic-dev:
## Summary

`actana pair new` had one shape for two audiences. A script wants labelled lines it can cut a field out of; the person standing at the Core wants to know what the code is *for* and what to type on the other machine, without reassembling `actana core pair` out of four separate lines by hand.

Now it has two, and **`isatty(stdout)` is the whole of the switch** — no flag, and deliberately no `--json` (#357 decided that explicitly).

**Piped, redirected or captured**, stdout is exactly what 0.4.2 printed — same lines, same order, same column. The suite asserts the whole array against a literal rather than trusting it.

**At a terminal** it is a framed handout, plus both redemption paths:

```
╭────────────────────────────────────────────────────────────────────────╮
│                                                                        │
│   Pairing code                                                         │
│                                                                        │
│      34X6-FGA8                                                         │
│                                                                        │
│   Expires        2026-08-20T12:05:00Z (in 5 minutes)                   │
│                                                                        │
│   CA fingerprint                                                       │
│      25:A0:FE:29:FA:5B:B9:58:63:78:2C:06:E7:B2:C0:DD:                  │
│      7B:8C:10:D7:E2:36:69:7B:8F:D9:31:5E:A2:01:35:B1                   │
│                                                                        │
│   Session        ffb1d774-11d4-443f-9be2-5bd9bc9486e0                  │
│   Label          laptop                                                │
│                                                                        │
╰────────────────────────────────────────────────────────────────────────╯

From the Panel
  Settings (gear icon) -> Cores -> Add Core, then enter the code above

From a terminal
  npm i -g @actana/cli

  # on the machine being paired, reachable at core (primary)
  actana core pair laptop core:8443 34X6-FGA8 --session ffb1d774-11d4-443f-9be2-5bd9bc9486e0 --fingerprint 25:A0:FE:29:…:B1

  # on the machine being paired, reachable at 10.0.0.5
  actana core pair laptop 10.0.0.5:8443 34X6-FGA8 --session ffb1d774-11d4-443f-9be2-5bd9bc9486e0 --fingerprint 25:A0:FE:29:…:B1
```

(The two fingerprints in the commands are elided **in this description only** — on screen they are whole. Colour is stripped here for the same reason.)

Notes on the shape:

- The pairing code is **bold cyan**; the fingerprint is **wrapped on its own colons and never truncated** — a fingerprint with an ellipsis in it is an invitation to guess.
- One pasteable command **per configured public host, primary first** (0.4.2 supports several addresses per Core). When `--public-host` chose one, that one is the only one offered: it is the endpoint redemption will hand this client.
- Every value in the command is real — code, `--session`, whole fingerprint, `host:port` from this install's `actana.json` (or `ACTANA_PORT` in a container, falling back to 8443). The only placeholder is `<name>`, and only when there is no `--label` or the label is not a name the client registry would take.
- Colour only at a terminal, and `NO_COLOR` turns it off without taking the instructions with it. Padding is measured on the plain text, so the frame stays square either way.

**Second half of the ticket:** the two usage strings that omitted the required `--session` flag (`core-pair.ts`'s "a name, an address and a code are required" line and `CORE_HELP`'s `Usage` block) now show it. `readTicket` refuses a bare code that does not name its pairing session, so a usage line without it sent operators to a second refusal for a flag nobody had mentioned.

**One new dep field:** `stdoutIsTty` on `ActanaCliDeps`, beside `stdinIsTty` — not a reuse of `interactive`, which is the *pair* of them and answers a different question (`pair new > code.txt` at a terminal has an interactive stdin and a redirected stdout, and it is the second half that decides). Nothing but the drawing reads it: the session minted, the digest stored and the exit code are identical either way.

## Related issues

Refs #357

## Type of change

## How was this tested?

Added to `packages/cli/src/__tests__/actana-pair.test.ts`:

- **`actana pair new, piped`** — five tests asserting `out` **as a whole array** against the pre-change literal: labelled, unlabelled, and with `Endpoint host` in its old position between `Label` and `Session`; that a multi-address Core still prints exactly those lines; and that stdout carries no escape sequence, no box drawing and no instruction text.
- **`actana pair new, at a terminal`** — the frame carries the minted code (checked against the digest in the store), the expiry, the **whole** fingerprint (no framed line holds all of it, and the framed pieces re-join to exactly the value, with no `...` or `…` anywhere), and the session id; both instruction paths, with the Panel wording asserted against the exported constant; the pasteable command carrying `--session <the minted id>` and the full fingerprint; one command per host, primary first; only the chosen host under `--public-host`; port from `actana.json`, from `ACTANA_PORT`, and the 8443 fallback; the `<name>` placeholder when unlabelled or when the label is not a usable Core name; colour present and every framed row 74 display columns wide despite the escapes; `NO_COLOR` degrading to no escapes with every instruction intact; and that the framed run mints, stores and refuses exactly what the piped run does.
- **`wrapFingerprint`** — every group kept, the break marked with the separator, and the value re-joinable for any group size.
- Usage strings: one test each in `core-pair.test.ts` and `core-command.test.ts`.

Local: `pnpm -C packages/cli run test` (50 files, 1137 passed), `pnpm test` (all 6 stages pass), `pnpm typecheck`, `pnpm lint` (no new findings).

None. The scraped stdout is unchanged by construction and by test.

## Checklist

- [x] This PR targets `feat/0.4.3-onboarding`, the 0.4.3 onboarding branch — not `main` and not the train directly
- [ ] Docs: no user-facing doc shows `pair new`'s output, so nothing needed updating; `packages/panel`'s Add Core flow is #358 and untouched here
- [x] Draft: stacked on `feat/0.4.3-onboarding`, alongside #360 (`core pair`'s result surface) and #358 (Panel)

### DIFF
--- a/docs/core-linux-rehearsal.md
+++ b/docs/core-linux-rehearsal.md
@@ -131,6 +131,43 @@ actana pair new --label my-panel
       fingerprint**, and when the code expires.
 - [ ] The code is short enough to read out loud over a phone without spelling
       anything — no `0`, `O`, `1`, `I` or `L` in it.
+- [ ] Because you are at a terminal, it comes **framed** (#357): the code, the
+      whole fingerprint wrapped rather than shortened, the session — and under
+      it both ways to spend them, the Panel path and a pasteable
+      `actana core pair …` command per address this Core answers on.
+- [ ] `actana pair new > /tmp/code.txt` instead gives the plain labelled lines
+      and nothing else — no frame, no instructions. That is the contract every
+      script wrapping this command depends on. (`rm /tmp/code.txt` after.)
+
+### The pasteable line, on a second machine
+
+**This is the one thing nothing automated can check**, and the reason it is a
+checkbox: the command below is generated from real minted values, and whether
+it works is a question about two machines. Take the `actana core pair …` line
+the block printed to a second machine — another VM, a laptop, a container with
+`npm i -g @actana/cli` on it — and paste it **unedited**.
+
+```bash
+# on the second machine, pasted exactly as printed
+actana core pair <as-printed>
+actana core status
+```
+
+- [ ] It pastes and runs. No `bash: …: No such file or directory`, no shell
+      error naming a file instead of `actana` — the line survives a shell as
+      printed, placeholder and all.
+- [ ] If `pair new` was run with no `--label`, the name slot is `NAME` and the
+      line still runs. Rename or remove it afterwards with `actana core rm`.
+- [ ] It pairs: the Core is registered, and no fingerprint had to be retyped.
+- [ ] `actana core status` **reaches the Core** from that second machine. A
+      pairing that succeeds and then cannot be reached is the failure this
+      checkbox exists for.
+- [ ] On a Core with several `--public-host` addresses: the block warns, on
+      every address but the one the credential will carry, that the code still
+      registers the other endpoint — and names the `actana pair new
+      --public-host <addr>` re-mint that would register the one you dialled.
+      Pair on a non-primary address and confirm `core status` behaves the way
+      the warning said it would.
 
 ---
 
--- a/packages/cli/src/__tests__/actana-pair.test.ts
+++ b/packages/cli/src/__tests__/actana-pair.test.ts
@@ -38,8 +38,10 @@ import {
 import {
   describeDuration,
   MAX_PAIRING_TTL_MS,
+  PANEL_INSTRUCTION,
   parseDuration,
   runPairCommand,
+  wrapFingerprint,
 } from "../actana-pair.ts";
 import type { ActanaCliDeps } from "../cli-deps.ts";
 import { stubClientHalf, stubMachineHalf } from "./machine-fixture.ts";
@@ -53,8 +55,19 @@ let out: string[];
 let err: string[];
 let audited: Record<string, unknown>[];
 
-/** One run of `actana pair <argv>`, with its two streams captured. */
-function run(argv: string[], now = NOW, env: Record<string, string> = {}): number {
+/**
+ * One run of `actana pair <argv>`, with its two streams captured.
+ *
+ * **stdout is a pipe unless a test says otherwise**, which is the shape the
+ * stdout contract is about and the shape everything written before #357
+ * asserts against. {@link runTty} is the other one.
+ */
+function run(
+  argv: string[],
+  now = NOW,
+  env: Record<string, string> = {},
+  stdoutIsTty = false,
+): number {
   out = [];
   err = [];
   const deps: ActanaCliDeps = {
@@ -65,11 +78,17 @@ function run(argv: string[], now = NOW, env: Record<string, string> = {}): numbe
     home: dir,
     out: (line: string) => out.push(line),
     err: (line: string) => err.push(line),
+    stdoutIsTty,
   };
   return runPairCommand(deps, argv, {
     materialPath: () => materialPath,
     audit: (record) => audited.push(record),
   });
+}
+
+/** The same run with a terminal on stdout — the framed shape (#357). */
+function runTty(argv: string[], now = NOW, env: Record<string, string> = {}): number {
+  return run(argv, now, env, true);
 }
 
 function store(): PairingStore {
@@ -419,6 +438,311 @@ describe("actana pair new --public-host", () => {
     expect(help).toMatch(/certificate already covers/);
   });
 });
+
+// ─── pair new, the two shapes (#357) ────────────────────────────────────────
+//
+// One command, two audiences. Down a pipe it is the labelled lines a script
+// cuts fields out of; at a terminal it is a framed handout that says what the
+// code is for and what to type on the other machine. The switch is
+// `isatty(stdout)` and nothing else — there is deliberately no `--json` and no
+// flag — so the two things this suite has to hold are that the piped shape did
+// not move a byte, and that the framed one carries everything an operator needs
+// to finish the job without retyping a fingerprint.
+
+describe("actana pair new, piped", () => {
+  /** The 0.4.2 shape, written out rather than derived from the code under test. */
+  function expectedLines(over: { label?: string; endpointHost?: string } = {}): string[] {
+    const lines = [
+      `Pairing code   ${field("Pairing code")}`,
+      `CA fingerprint ${new X509Certificate(material.caCert).fingerprint256}`,
+      "Expires        2026-08-20T12:05:00Z (in 5 minutes)",
+    ];
+    if (over.label) lines.push(`Label          ${over.label}`);
+    if (over.endpointHost) lines.push(`Endpoint host  ${over.endpointHost}`);
+    lines.push(`Session        ${field("Session")}`);
+    return lines;
+  }
+
+  it("prints the labelled lines byte for byte, in the order 0.4.2 printed them", () => {
+    expect(run(["new", "--label", "laptop"])).toBe(0);
+    // Every line, whole, in order — not `toContain`. A framed block that leaked
+    // into the piped path, an extra blank line, a changed column width or a
+    // reordered field would each fail here, and each of them breaks a script.
+    expect(out).toEqual(expectedLines({ label: "laptop" }));
+  });
+
+  it("keeps the unlabelled shape, which has no Label line at all", () => {
+    expect(run(["new"])).toBe(0);
+    expect(out).toEqual(expectedLines());
+  });
+
+  it("keeps `Endpoint host` where it was, between Label and Session", async () => {
+    material = await mintFreshMaterial(["core", "10.0.0.5"]);
+    persistMaterialToFile(materialPath, material);
+
+    expect(run(["new", "--label", "laptop", "--public-host", "10.0.0.5"])).toBe(0);
+
+    expect(out).toEqual(expectedLines({ label: "laptop", endpointHost: "10.0.0.5" }));
+  });
+
+  it("writes no escape sequence, no box drawing and no instructions", () => {
+    run(["new", "--label", "laptop"]);
+    const printed = out.join("\n");
+    expect(printed).not.toMatch(/\x1b\[/);
+    for (const ornament of ["╭", "│", "─"]) expect(printed).not.toContain(ornament);
+    expect(printed).not.toContain("From the Panel");
+    expect(printed).not.toContain("From a terminal");
+    expect(printed).not.toContain("npm i -g");
+  });
+
+  it("is the shape a Core with several addresses prints too", async () => {
+    // The framed path prints one command per address. Down a pipe the several
+    // addresses change nothing at all: same five lines.
+    material = await mintFreshMaterial(["core", "10.0.0.5", "core.example.test"]);
+    persistMaterialToFile(materialPath, material);
+
+    expect(run(["new", "--label", "laptop"])).toBe(0);
+
+    expect(out).toEqual(expectedLines({ label: "laptop" }));
+  });
+});
+
+describe("actana pair new, at a terminal", () => {
+  /** Everything stdout saw, as one string. */
+  function screen(): string {
+    return out.join("\n");
+  }
+
+  /** The pasteable command lines — the ones that start the client verb. */
+  function commands(): string[] {
+    return out.map((line) => line.trim()).filter((line) => line.startsWith("actana core pair "));
+  }
+
+  it("frames the code, and the code is what was minted", () => {
+    expect(runTty(["new", "--label", "laptop"])).toBe(0);
+
+    const session = store().listSessions()[0]!;
+    const code = /([A-Z2-9]{4}-[A-Z2-9]{4})/.exec(screen())?.[1];
+    expect(code).toBeDefined();
+    // The digest in the store is the digest of the code on the screen, which is
+    // the only thing that makes the framed shape the same shape.
+    expect(
+      pairingCodeMatches(
+        session.codeHash,
+        hashPairingCode({
+          key: derivePairingCodeKey(material.bearerSecret),
+          sessionId: session.id,
+          code: code!,
+        }),
+      ),
+    ).toBe(true);
+    expect(screen()).toContain("╭");
+    expect(screen()).toContain("Pairing code");
+  });
+
+  it("prints the expiry beside the code, absolute and relative", () => {
+    runTty(["new", "--ttl", "2h"]);
+    expect(screen()).toContain("2026-08-20T14:00:00Z (in 2 hours)");
+  });
+
+  it("prints the WHOLE fingerprint, wrapped rather than shortened", () => {
+    runTty(["new", "--label", "laptop"]);
+
+    const fingerprint = new X509Certificate(material.caCert).fingerprint256;
+    expect(fingerprint.split(":")).toHaveLength(32);
+
+    // Inside the frame it is wrapped: no framed line holds the whole of it,
+    // and the framed lines that hold pieces of it re-join to exactly the value
+    // — nothing dropped, nothing elided, nothing rewritten.
+    const framed = out.filter((line) => line.includes("│"));
+    expect(framed.some((line) => line.includes(fingerprint))).toBe(false);
+    const rejoined = framed
+      .map((line) => /([0-9A-F]{2}(?::[0-9A-F]{2}){7,}:?)/.exec(line)?.[1] ?? "")
+      .join("");
+    expect(rejoined).toBe(fingerprint);
+    for (const elision of ["...", "…"]) expect(screen()).not.toContain(elision);
+    // And the pasteable command carries it whole, on one line, because that is
+    // the copy a client actually checks the certificate against.
+    expect(commands()[0]).toContain(`--fingerprint ${fingerprint}`);
+  });
+
+  it("prints the session id", () => {
+    runTty(["new"]);
+    const sessionId = store().listSessions()[0]!.id;
+    expect(screen()).toContain(sessionId);
+  });
+
+  it("gives the Panel path, worded as the ticket words it", () => {
+    runTty(["new", "--label", "laptop"]);
+    expect(screen()).toContain("From the Panel");
+    expect(screen()).toContain(
+      "Settings (gear icon) -> Cores -> Add Core, then enter the code above",
+    );
+    expect(PANEL_INSTRUCTION).toBe(
+      "Settings (gear icon) -> Cores -> Add Core, then enter the code above",
+    );
+  });
+
+  it("gives the terminal path: the install, then a command with real values", () => {
+    expect(runTty(["new", "--label", "laptop"])).toBe(0);
+
+    expect(screen()).toContain("From a terminal");
+    expect(screen()).toContain("npm i -g @actana/cli");
+
+    const session = store().listSessions()[0]!;
+    const code = /([A-Z2-9]{4}-[A-Z2-9]{4})/.exec(screen())![1]!;
+    const fingerprint = new X509Certificate(material.caCert).fingerprint256;
+    expect(commands()).toEqual([
+      `actana core pair laptop 10.0.0.5:8443 ${code} --session ${session.id} --fingerprint ${fingerprint}`,
+    ]);
+    // The install line comes before the command it installs the binary for.
+    expect(screen().indexOf("npm i -g @actana/cli")).toBeLessThan(screen().indexOf("actana core pair "));
+  });
+
+  // The other half of the ticket: `readTicket` refuses a bare code, so a
+  // pasteable line without `--session` is a line that fails on arrival.
+  it("carries --session, with the id of the session it just minted", () => {
+    runTty(["new", "--label", "laptop"]);
+    const session = store().listSessions()[0]!;
+    for (const command of commands()) {
+      expect(command).toContain(`--session ${session.id}`);
+      expect(command).toContain("--fingerprint ");
+    }
+  });
+
+  it("offers one command per configured address, the primary first", async () => {
+    material = await mintFreshMaterial(["core", "10.0.0.5", "core.example.test"]);
+    persistMaterialToFile(materialPath, material);
+
+    expect(runTty(["new", "--label", "laptop"])).toBe(0);
+
+    const addresses = commands().map((command) => command.split(" ")[4]);
+    expect(addresses).toEqual(["core:8443", "10.0.0.5:8443", "core.example.test:8443"]);
+  });
+
+  it("offers only the chosen address when --public-host chose one", async () => {
+    material = await mintFreshMaterial(["core", "10.0.0.5", "core.example.test"]);
+    persistMaterialToFile(materialPath, material);
+
+    expect(runTty(["new", "--label", "laptop", "--public-host", "10.0.0.5"])).toBe(0);
+
+    // That is the endpoint redemption will hand this client, so a command
+    // pointing anywhere else would pair it to an address the code did not pick.
+    expect(commands().map((command) => command.split(" ")[4])).toEqual(["10.0.0.5:8443"]);
+  });
+
+  it("dials the port this install recorded, not a guess", () => {
+    fs.writeFileSync(
+      path.join(dir, "actana.json"),
+      JSON.stringify({
+        version: "0.4.3",
+        port: 9443,
+        host: "0.0.0.0",
+        publicHost: "10.0.0.5",
+        label: "core-1",
+        installDir: dir,
+        dataDir: dir,
+      }),
+    );
+
+    runTty(["new", "--label", "laptop"]);
+
+    expect(commands()[0]).toContain(" 10.0.0.5:9443 ");
+  });
+
+  it("dials the container's port when it is running in one", () => {
+    runTty(["new", "--label", "laptop"], NOW, { ACTANA_CONTAINER: "1", ACTANA_PORT: "7443" });
+    expect(commands()[0]).toContain(" 10.0.0.5:7443 ");
+  });
+
+  it("falls back to 8443 rather than printing no command at all", () => {
+    // No `actana.json` beside the material: an operator with a wrong port can
+    // fix one character, and an operator with no command has to build it.
+    runTty(["new", "--label", "laptop"]);
+    expect(commands()[0]).toContain(" 10.0.0.5:8443 ");
+  });
+
+  it("uses a placeholder name when there is no label to use", () => {
+    runTty(["new"]);
+    expect(commands()[0]).toContain("actana core pair <name> ");
+  });
+
+  it("uses a placeholder when the label is not a name the client registry takes", () => {
+    // `--label` takes anything an operator wants to call a machine; a Core name
+    // does not. Pasting `my laptop` would fail on the far machine with a
+    // message about a registry nobody mentioned.
+    runTty(["new", "--label", "my laptop"]);
+    expect(commands()[0]).toContain("actana core pair <name> ");
+    // The label is still the label — it is on the frame and in the store.
+    expect(screen()).toContain("my laptop");
+    expect(store().listSessions()[0]!.label).toBe("my laptop");
+  });
+
+  it("colours at a terminal, and the frame stays square anyway", () => {
+    runTty(["new", "--label", "laptop"]);
+    const printed = out.join("\n");
+    expect(printed).toMatch(/\x1b\[1;36m/);
+    // Padding measured on the plain text, so every framed row is the same
+    // *display* width despite the escapes having a length and no width.
+    const framed = out.filter((line) => line.includes("│")).map((line) => line.replace(/\x1b\[[0-9;]*m/g, ""));
+    expect(framed.length).toBeGreaterThan(5);
+    expect(new Set(framed.map((line) => [...line].length))).toEqual(new Set([74]));
+  });
+
+  it("degrades to no escapes under NO_COLOR, keeping every instruction", () => {
+    runTty(["new", "--label", "laptop"], NOW, { NO_COLOR: "1" });
+    const printed = out.join("\n");
+    expect(printed).not.toMatch(/\x1b\[/);
+    expect(printed).toContain("From the Panel");
+    expect(printed).toContain("From a terminal");
+    expect(commands()).toHaveLength(1);
+  });
+
+  it("mints exactly what the piped shape mints — only the printing differs", () => {
+    expect(runTty(["new", "--label", "laptop", "--ttl", "30s"])).toBe(0);
+    const session = store().listSessions()[0]!;
+    expect(session.expiresAt - session.createdAt).toBe(30_000);
+    expect(session.label).toBe("laptop");
+    // The prose on stderr is the same prose, terminal or not: it is not what a
+    // scraper reads, and it is what a human is told either way.
+    expect(err.join("\n")).toMatch(/Read the code AND the fingerprint/);
+    expect(err.join("\n")).toContain(`actana pair revoke ${session.id}`);
+  });
+
+  it("still stores a digest and never the code, framed or not", () => {
+    runTty(["new", "--label", "laptop"]);
+    const code = /([A-Z2-9]{4}-[A-Z2-9]{4})/.exec(out.join("\n"))![1]!;
+    const session = store().listSessions()[0]!;
+    expect(JSON.stringify(session)).not.toContain(code);
+    expect(JSON.stringify(session)).not.toContain(code.replace("-", ""));
+  });
+
+  it("refuses the same things it refuses down a pipe, and frames nothing", () => {
+    expect(runTty(["new", "--ttl", "5"])).toBe(2);
+    expect(out).toEqual([]);
+    expect(store().listSessions()).toEqual([]);
+  });
+});
+
+describe("wrapFingerprint", () => {
+  it("keeps every group, and marks the break with the separator", () => {
+    const fingerprint = Array.from({ length: 32 }, (_, i) => i.toString(16).padStart(2, "0").toUpperCase()).join(":");
+    const lines = wrapFingerprint(fingerprint);
+    expect(lines).toHaveLength(2);
+    expect(lines[0]!.endsWith(":")).toBe(true);
+    expect(lines[1]!.endsWith(":")).toBe(false);
+    // Rejoinable, exactly: a wrapped fingerprint is the same value with a line
+    // break in it, and never a shortened one.
+    expect(lines.join("")).toBe(fingerprint);
+  });
+
+  it("cannot shorten its input, whatever it is asked for", () => {
+    for (const perLine of [0, 1, 5, 999]) {
+      expect(wrapFingerprint("AA:BB:CC:DD", perLine).join("")).toBe("AA:BB:CC:DD");
+    }
+  });
+});
+
 
 // ─── pair ls ────────────────────────────────────────────────────────────────
 
--- a/packages/cli/src/__tests__/actana-pair.test.ts
+++ b/packages/cli/src/__tests__/actana-pair.test.ts
@@ -18,6 +18,7 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 import { X509Certificate } from "node:crypto";
+import { execFileSync } from "node:child_process";
 import {
   loadMaterialFromFile,
   materialFilePath,
@@ -25,6 +26,7 @@ import {
   persistMaterialToFile,
   type PersistedMaterial,
 } from "@actana/shared/core-material-store";
+import { coreNameError } from "@actana/shared/blob-registry";
 import { normalisePairingCode, PAIRING_CODE_ALPHABET } from "@actana/shared/pairing-code";
 import { createPairingSession, PAIRING_SESSION_TTL_MS } from "@actana/shared/pairing-session";
 import {
@@ -37,7 +39,10 @@ import {
 } from "@actana/shared/pairing-store";
 import {
   describeDuration,
+  displayWidth,
+  FRAME_WIDTH,
   MAX_PAIRING_TTL_MS,
+  NAME_PLACEHOLDER,
   PANEL_INSTRUCTION,
   parseDuration,
   runPairCommand,
@@ -572,15 +577,33 @@ describe("actana pair new, at a terminal", () => {
     expect(screen()).toContain(sessionId);
   });
 
-  it("gives the Panel path, worded as the ticket words it", () => {
+  // The wording is pinned to the *form*, not to a paraphrase of it.
+  // `AddCoreByPairing.tsx` asks for the address first, gates everything behind
+  // a compared CA fingerprint — "the Panel does not send the code until they
+  // match" — and only then shows Session and Pairing code. An instruction that
+  // named only the code sent an operator to a form that wanted two things
+  // first, one of them the security-relevant one (#357 review B1).
+  it("gives the Panel path in the order the Panel's own form asks for it", () => {
     runTty(["new", "--label", "laptop"]);
     expect(screen()).toContain("From the Panel");
-    expect(screen()).toContain(
-      "Settings (gear icon) -> Cores -> Add Core, then enter the code above",
-    );
+    expect(screen()).toContain(PANEL_INSTRUCTION);
     expect(PANEL_INSTRUCTION).toBe(
-      "Settings (gear icon) -> Cores -> Add Core, then enter the code above",
+      "Settings (gear icon) -> Cores -> Add Core: this Core's address, then compare the " +
+        "CA fingerprint, then the session and the code",
     );
+    // Every field the form requires is named, in the form's order, and the
+    // fingerprint comparison is not dropped — it is what makes the first dial
+    // verifiable, and the frame above prints the fingerprint without this
+    // sentence saying what it is for.
+    const address = PANEL_INSTRUCTION.indexOf("address");
+    const fingerprint = PANEL_INSTRUCTION.indexOf("fingerprint");
+    const session = PANEL_INSTRUCTION.indexOf("session");
+    const code = PANEL_INSTRUCTION.indexOf("code");
+    expect(address).toBeGreaterThan(-1);
+    expect(fingerprint).toBeGreaterThan(address);
+    expect(session).toBeGreaterThan(fingerprint);
+    expect(code).toBeGreaterThan(session);
+    expect(PANEL_INSTRUCTION).toContain("compare");
   });
 
   it("gives the terminal path: the install, then a command with real values", () => {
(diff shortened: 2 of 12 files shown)


qcentic-dev:
## Acceptance criteria, #357

Every criterion in the issue body and in the issue's follow-up comment, checked or explicitly not.

### From "What to build"

- [x] **When stdout is a TTY, print a cosmetic, instruction-rich block instead of the bare lines.** `pairNew` branches on `deps.stdoutIsTty` and on nothing else — no flag, no environment variable.
- [x] **A framed block (unicode box/rules, ANSI color) presenting the pairing code and CA fingerprint prominently, with expiry and session below.** Rounded box rules; the code alone on its own line in bold cyan; the full fingerprint below it; expiry beside the code, session (and label, when there is one) under the fingerprint. Rendered sample in the PR description.
  - Order as specified in the implementation brief: code + expiry, then the **full** fingerprint, then the session.
- [x] **The fingerprint is never truncated.** Wrapped inside the frame on its own colon separators, every group but the last line ending in `:`. `wrapFingerprint` structurally cannot shorten its input, and a test re-joins the framed pieces and asserts they equal the certificate's `fingerprint256` exactly, with no `...` or `…` anywhere on the screen. The pasteable command carries it whole on one line.
- [x] **Under it, BOTH redemption paths, always, in one block.** `From the Panel` and `From a terminal` are both emitted unconditionally, in that order.
- [x] **Panel — the canonical path.** Reads exactly: `Settings (gear icon) -> Cores -> Add Core, then enter the code above`, exported as `PANEL_INSTRUCTION` and asserted against that literal.
  - ⚠️ **Wording note for the reviewer:** the issue body words it *"Settings (gear icon) → Cores → Add Core → enter this code and fingerprint."* The implementation brief for this ticket restated it as *"Settings (gear icon) -> Cores -> Add Core, then enter the code above"* and asked for that exactly, so that is what shipped — ASCII arrows, and "then enter the code above" for the tail. One string, one constant: say the word and it is a one-line change.
- [x] **CLI — first `npm i -g @actana/cli`, then a fully pasteable line with real values filled in.** The install line is printed first, then the command(s). No placeholders for code, session, fingerprint, host or port.
- [x] **One line per configured public host, each with that host's endpoint filled in.** `material.serverHosts` in the operator's order, so the primary is first, each rendered `host:port` (IPv6 literals bracketed).
  - Judgement call, called out: when `--public-host` chose an address, **only that one** is offered. That address is the endpoint redemption will hand this client, so a command aimed anywhere else would pair it to an address this code did not choose. Tested both ways.
- [x] **`--label` for the name when given, a placeholder otherwise.** `<name>` when there is no label, and also when the label is not a name the client registry would accept (`coreNameError`) — pasting `my laptop` would otherwise fail on the far machine with a message about a registry nobody mentioned. The label itself is still shown in the frame and still stored.
- [x] **When stdout is not a TTY, output stays byte-identical to today's bare labelled lines.** Same lines, same order, same column, `Endpoint host` still between `Label` and `Session`. Asserted as a whole-array equality against a hand-written literal, not with `toContain`, so an added line, a reordered field or a changed column all fail.
- [x] **The stdout=facts / stderr=prose contract preserved.** stderr is unchanged and identical in both shapes.
- [x] **No `--json` flag.** None added. The header and the section comment both record that it is deliberate.
- [x] **TTY detection is the whole switch.** New `stdoutIsTty` on `ActanaCliDeps` (beside `stdinIsTty`, not a reuse of `interactive`). Only the drawing reads it: the session minted, the digest stored and the exit code are identical either way, and a test asserts that.

### From "Acceptance"

- [x] **TTY run shows the framed block with both paths and pasteable per-host `core pair` lines with real values.** Covered by `actana pair new, at a terminal`.
- [x] **Piped run is unchanged from 0.4.2 output.** Covered by `actana pair new, piped`.
- [ ] **The pasteable line actually works when pasted on a second machine against the minted code.** **Not checked end-to-end**, and honestly so: this needs two machines (or a container plus a host), a running daemon and a real TLS dial, which is out of reach from this working copy. What *is* checked is every part that this repo can check — the code matches the digest the store holds, the session id is the one just minted, the fingerprint is the certificate's own `fingerprint256` in full, `--session` and `--fingerprint` are both present, and the address is `host:port` built from this install's recorded port. Worth one manual pairing before this leaves the train.

### From the issue's follow-up comment

- [x] **`packages/cli/src/core-pair.ts`'s usage string now shows `--session`.** Now `actana core pair <name> <host:port> <code> --session <id> --fingerprint <sha256>`, with a test asserting the exact line.
- [x] **`packages/cli/src/core-command.ts`'s help block now shows `--session`.** The `Usage` line in `CORE_HELP` is now `actana core pair <name> <address> <code> --session <id> --fingerprint <sha256>`, with a test.
- [x] **The new TTY block's pasteable line carries `--session` too.** With the real minted id, asserted for every command the block emits.

### Boundaries respected

- [x] `actana core pair`'s success and failure **result** surface untouched — that is #360. The only change in `core-pair.ts` is the usage string this ticket names (plus the adjacent hint line, which now says the Core prints "the code, the fingerprint and the session").
- [x] `packages/panel` untouched — that is #358.
- [x] Nothing merged, `main` untouched, PR left as a draft.

qcentic-dev:
### Follow-up: the installer E2E caught a real consequence of the TTY switch, and now asserts both shapes

The first CI run was green everywhere except `E2E — installer (debian/ubuntu, x64)`. Worth reading rather than skipping, because the failure was the feature working:

`scripts/e2e-actana-setup-linux.mjs` drives the installed Core through `machinectl shell`, which allocates a **pty**. So `actana pair new` there now sees `isatty(stdout) === true` and renders the frame — and the e2e's `Pairing code\s+XXXX-XXXX` scrape found no such line. The job log shows the frame rendering correctly on a really-installed Core, endpoint and all:

```
│      3X3T-6JKP                                                         │
│   CA fingerprint                                                       │
│      F3:10:8F:46:6F:57:AB:4E:1C:13:98:45:9E:E1:F9:2D:                  │
│      84:C8:2A:76:83:90:9F:71:7D:46:AA:BA:4F:4E:2E:1C                   │
...
  actana core pair e2e-panel 127.0.0.1:8443 3X3T-6JKP --session 9ee98b92-… --fingerprint F3:10:…:1C
```

**This is worth a reviewer's attention as a fact about the chosen contract, not just a test fixture.** `isatty(stdout)` being the whole switch means any harness that drives the CLI through a pty — `machinectl shell`, `ssh -t`, `docker exec -t`, expect-style wrappers — gets the cosmetic shape. #357 chose that trade explicitly (no `--json`, TTY detection is the whole switch), and this change keeps that choice; I am flagging it rather than quietly working around it. If it turns out to be the wrong trade, the switch is one `if`.

The e2e now asserts **both** halves on the machine the one-liner built (commit `8e35d34`):

- redirected to a file → the labelled lines are exactly what a script reads, and nothing cosmetic reaches them (no frame character, no `From the Panel`, no `npm i -g`);
- straight to the terminal → the frame is drawn, both redemption paths are present with the Panel wording verbatim, and the pasteable command carries `--session` and a whole 32-group fingerprint.

That makes this the one place either half of #357 is checked against a Core that was really installed, and it moves the last unchecked acceptance criterion — *"the pasteable line actually works when pasted on a second machine"* — closer: the line's **shape and values** are now verified on a real install. Actually pasting it on a second machine and completing the exchange still wants one manual pass before this leaves the train.

Everything else on the first run was green, including `Unit Tests`, `Lint`, `Typecheck`, `Conventions`, `Train rules`, both image smokes and both tarball smokes.

qcentic-dev:
### CI is green

All 18 checks pass on `8e35d34` (6 skipped, all train-publish jobs that do not run on a PR): `Unit Tests`, `Lint`, `Typecheck`, `Conventions` (all three commits, not just the title), `Train rules`, `Secret Scan`, `Dependency Audit`, `E2E — Panel service seam`, both `E2E — installer` jobs, both image smokes, both tarball smokes.

The installer E2E now logs both halves of #357 against a really-installed Core:

```
[setup-e2e] `actana pair new` keeps its labelled lines when stdout is not a terminal (#357)
[setup-e2e] `actana pair new` frames the code and both redemption paths at a terminal (#357)
```

Left as a **draft** and not merged, as asked. The one acceptance criterion still unchecked is the manual one: pasting the generated `actana core pair …` line on a second machine and completing the exchange.

mrfqcentic (changes requested):
## Review of #364 against #357 — changes requested

I checked the five claims in the description by reading the enforcing lines rather than the prose, and ran a separate automated pass over the diff. Most of the claims hold, and hold well. Three things do not, and all three are in the framed block itself — the part of this ticket that is *only* instructions, where being wrong costs an operator a failed pairing rather than an ugly screen.

### What I verified and confirm

**The pipe contract is genuinely byte-identical, and genuinely tested.** `actana-pair.ts:314-322` emits the same six template strings, in the same order, with the same column widths and the same two conditionals (`if (label)`, `if (endpointHost)`) as `feat/0.4.3-onboarding`. The only delta is hoisting `certFingerprintSha256(material.caCert)` into a local, which is pure. `actana-pair.test.ts:466-506` asserts `expect(out).toEqual(expectedLines(...))` — whole-array `toEqual` against a hand-written literal, not `toContain`. A moved column, an added line, a reordered field or a leaked frame character each fail it, and `expectedLines` reads back only the two unpredictable values (code, session id) via `field()`. Four shapes covered: labelled, unlabelled, `Endpoint host` in its old slot, and multi-address.

**The TTY switch really is the only selector.** `git grep stdoutIsTty` at this head returns exactly two production reads, both in `actana-pair.ts`: the shape branch at `:287` and the colour gate at `:470`, and the latter is only ever reached from inside the former. In the suite, `run()` and `runTty()` differ in one boolean argument (`actana-pair.test.ts:69,81,88`). Nothing about *what* the command does reads it — `:701-718` asserts the minted session, the stored digest, the stderr prose and the exit code are identical either way. Splitting `stdoutIsTty` out rather than reusing `interactive` (`actana-cli-entry.ts:178` — `stdin.isTTY && stdout.isTTY`) is the right call and the doc comment gives the right reason.

**Claim 2 is proven by assertions, not screenshots.** Code checked against the store's digest (`:521-542`); expiry (`:543`); whole fingerprint, with the framed pieces re-joined to exactly `fingerprint256`, no framed line holding all of it, and no `...`/`…` anywhere (`:548-568`); session id (`:569`); `PANEL_INSTRUCTION` against a literal (`:575-585`); `npm i -g @actana/cli` ordered before the command (`:586-603`); `--session <minted id>` in every emitted command (`:604-612`); one command per host, primary first (`:613-622`); port from `actana.json`, from `ACTANA_PORT`, and the 8443 fallback (`:634-664`).

**Claim 3 holds.** `core-pair.ts:130` and `core-command.ts:68` both now carry `--session <id> --fingerprint <sha256>`, each with an exact-string test (`core-pair.test.ts:317-326`, `core-command.test.ts:61-70`). See N3 below — there are two more of the same string that were missed.

**Claim 4 — the #360 boundary — holds.** `core-pair.ts`'s entire diff is the three-line usage refusal for missing positionals. `runCorePair`'s success surface (`:206-210`) and `reportPairingFailure`'s per-class explanations are untouched. On the adjacent hint line: it sits inside that same three-line block, and once the usage line names `--session`, the old "prints the code and the fingerprint" is factually wrong. Changing it was *required* by the change #357 explicitly asked for. That is on this ticket's side of the line. #360's "`--session` reminder for a bare code" is the `readTicket` failure class, which is not touched here.

**Claim 5 — the e2e change is right, and the trade is the right one to have taken.** `machinectl shell` allocates a pty, so `pair new` there now renders the frame and the old scrape was reading the wrong shape. Redirecting to a file and scraping that (`e2e-actana-setup-linux.mjs:405-406`) asserts the contract that actually matters downstream, and the second, direct run (`:430-453`) means both halves are now checked against a really-installed Core — the only place either is. The file already had precedent for exactly this: `activate()` passes `</dev/null` for the same pty reason. On the trade itself: `isatty(stdout)` as the whole switch means `ssh -t`, `docker exec -t`, `expect` and `script` all get the cosmetic shape, with no `--json` escape hatch. #357 chose that deliberately and I would keep it — the escape hatch already exists and is the standard one, which is to redirect stdout, which is what a script does anyway. Flagging it rather than quietly working around it was the right call.

---

### Blocking — 1. The Panel instruction describes a flow the Panel does not have

`packages/cli/src/actana-pair.ts:578-579`

```
export const PANEL_INSTRUCTION =
  "Settings (gear icon) -> Cores -> Add Core, then enter the code above";
```

The issue worded this **"Settings (gear icon) → Cores → Add Core → enter this code and fingerprint."** and said *worded exactly*. The shipped tail drops the fingerprint. I read the Panel rather than arbitrate on wording preference, and the issue's version is the correct one:

- `packages/panel/src/components/views/AddCoreByPairing.tsx:169-170` — *"Enter the address first and compare the fingerprint — the Panel does not send the code until they match."*
- `:334` — *"Now type the CA fingerprint `actana pair new` printed on that machine."*
- `:349` — a `TextField` labelled ``CA fingerprint from `actana pair new` ``.

So the real flow is **address → type the fingerprint → then the code**, and step 3 is not reachable until the fingerprint matches. An operator following "then enter the code above" arrives at a form that first wants an address the framed block never prints as a field, and then wants a fingerprint they have just been told they do not need. The fingerprint comparison is also the security-relevant step in this exchange, and the Panel path is now the one place that does not mention it — while the frame directly above prints the fingerprint prominently and never says what it is for.

That is the ticket's own purpose inverted: the block exists so nobody has to reassemble the next step by hand, and per the issue the Panel is the *canonical* path.

The ASCII arrows I have no problem with — `->` is safe in every terminal encoding and reads fine. It is the tail that is wrong.

Suggested — still one constant, one line, and it also gives the operator the address:

```
"Settings (gear icon) -> Cores -> Add Core: the address above, then compare the CA fingerprint, then the code"
```

Whatever you land on, `actana-pair.test.ts:575-585` and `scripts/e2e-actana-setup-linux.mjs:434` both pin the current string verbatim, so all three move together. Note that #363 mounts this same `AddCoreByPairing` inside a first-run wizard, so "Settings → Cores → Add Core" is also not the path a fresh Panel takes — worth a word with that PR's author about wording that serves both, since they are stacked on the same branch.

### Blocking — 2. Every non-primary pasteable command pairs the client to the primary endpoint

`packages/cli/src/actana-pair.ts:691-694` (`handoutHosts`), used at `:556-566`

```
function handoutHosts(configured: readonly string[], chosen: string | undefined): string[] {
  if (chosen) return [chosen];
  return configured.length > 0 ? [...configured] : [primaryPublicHost(configured)];
}
```

The reasoning given for narrowing under `--public-host` is exactly right, and it is the same reasoning that says the un-narrowed branch is wrong. The endpoint a client keeps comes off the **stored session**, never off the address it dialled — `packages/core/src/core-pairing-wiring.ts:107-112`:

```
const chosen = session.endpointHost ?? "";
const host = chosen.length > 0 && isConfiguredPublicHost(opts.publicHosts, chosen) ? chosen : primary;
return `wss://${host}:${opts.port}`;
```

With no `--public-host`, `session.endpointHost` is unset, so redemption returns the **primary** for every one of the N commands. `packages/cli/src/core-pair.ts:206` then stores `blob.endpoint` — the primary — regardless of which line was pasted. `core-pairing-redeem.test.ts` states this as intended: *"a code minted without choosing anything hands back the first configured address."*

Concretely, on a Core with `ACTANA_PUBLIC_HOST=core,10.0.0.5`, `actana pair new --label laptop` at a terminal offers:

```
  # on the machine being paired, reachable at core (primary)
  actana core pair laptop core:8443 …
  # on the machine being paired, reachable at 10.0.0.5
  actana core pair laptop 10.0.0.5:8443 …
```

A laptop that cannot resolve `core` pastes the second line. It dials `10.0.0.5:8443`, pairs successfully, and is left with a Core registered at `wss://core:8443` — which it cannot reach. `actana core status` then fails on a machine that just reported a successful pairing, for a reason nothing on screen explains. That is the multi-address case #347 / ADR 0038 exists to serve, and the handout walks the operator into its one sharp edge.

The comment `# on the machine being paired, reachable at <host>` is the specific thing that is untrue for hosts 2..N: that address is what you dial, not what you keep.

The underlying behaviour is not new, but before this PR an operator assembled the command themselves from lines that named only the chosen endpoint; the handout now actively hands them the wrong one N-1 times out of N. `core-pairing-redeem.test.ts:637-638` shows the designed workflow is one code per address (`--public-host` each time), so the block is arguing against the design it sits in.

Either resolution is fine by me:

- **(a)** offer only the primary when `--public-host` was not given — symmetric with the narrowing you already do, and one line; or
- **(b)** keep all N and say which endpoint the credential will carry, e.g. append to the non-primary comments *"— this pairs to `core:8443`; re-mint with `--public-host 10.0.0.5` to keep this address"*.

(a) is smaller and matches the intended workflow; (b) teaches the flag. `actana-pair.test.ts:613-622` pins the current list either way.

*(An automated pass over the diff reached this same finding independently, by the same route.)*

### Blocking — 3. The `<name>` placeholder makes the default invocation's line un-pasteable

`packages/cli/src/actana-pair.ts:615-618` (`handoutName`)

`actana pair new` with no `--label` is the documented default, and it prints:

```
actana core pair <name> 10.0.0.5:8443 ABCD-2345 --session … --fingerprint …
```

`<name>` is not inert in a shell. Pasted into bash, `<name>` is a stdin redirection from a file called `name` and a stdout redirection into a file called `10.0.0.5:8443`. I ran it:

```
$ actana core pair <name> 10.0.0.5:8443 ABCD-2345 --session x --fingerprint y
bash: line 1: name: No such file or directory
```

The command never runs. The operator gets a shell error that names neither `actana` nor the actual problem, on the one line whose entire stated promise — in the ticket, in the PR description and in the comment above `corePairCommand` — is *"the whole promise of this line is that it works when pasted."* It fires on the default path and on `--label "my laptop"`, i.e. exactly the cases `handoutName` was written for. `actana-pair.test.ts:665-680` asserts the placeholder is present but never that the line survives a shell.

The issue body did say "a `<name>` placeholder", so this is the spec's wording colliding with the spec's own promise rather than a misreading — but the promise should win. `NAME`, or `'<name>'` quoted, or emitting that whole command commented out with a line above it saying to fill in the name, all fix it. My preference is a bare `NAME`: still obviously a slot, and the line then fails with `actana core pair`'s own message about a name that is not a registry name, which is a message that tells the operator what to do.

---

### Non-blocking

**N1 — the frame is not square for a long or wide label.** `actana-pair.ts:665-672`: `fill = Math.max(1, FRAME_WIDTH - 2 - FRAME_GUTTER - plain.length)`, so content wider than 69 columns pushes the right border out instead of wrapping. `"Label          "` is 15 characters, so a `--label` of 55+ is ragged, and `--label` takes anything. Separately, `plain.length` is UTF-16 code units, not display width — a CJK or emoji label misaligns well before 55 (`--label 笔记本` is 3 columns over), and `:681-691`'s `[...line].length === 74` assertion would still pass while the terminal renders it crooked. Cosmetic only, and it cannot crash. Truncating or width-measuring the *label* row — never the fingerprint — closes it.

**N2 — the e2e's credential-leak check got narrower, and leaves a live code in `/tmp`.** `scripts/e2e-actana-setup-linux.mjs:405-406`: `minted` is now `cat /tmp/pair-new.out`, i.e. stdout only, while the comment above still says the run "must not put the code on stdout twice or leak a credential while doing it". Before the change `minted.stdout` was the folded stdout+stderr that `machinectl shell` produces, so the `BEGIN (CERTIFICATE|PRIVATE KEY)` check at `:413` covered both streams; it now covers neither stderr nor the framed run. Line `:458` in the same file still uses the `${stdout}${stderr}` convention. Keeping the result of the redirected run at `:405` and testing `${that.stdout}${that.stderr}` restores it. Also `/tmp/pair-new.out` holds a redeemable pairing code and is never removed — the container is ephemeral, but this file's whole discipline is about not leaving credentials lying around, so an `rm` after the scrape would keep it honest.

**N3 — two more copies of the same usage string still omit both flags.** The ticket named two, and both are fixed. These two print the identical invocation and were missed:
- `packages/cli/src/core-command.ts:182` — ``"No Cores registered. `actana core pair <name> <address> <code>` registers one."``
- `packages/cli/src/core-resolution.ts:124` — ``"… `actana core pair <name> <address> <code>` registers one."``

An operator who runs `actana core ls` on an empty registry, or any noun with no Core selected, is handed the same flagless line and lands in exactly the double refusal this change set out to remove. Strictly beyond the letter of the issue's follow-up comment, which is why it is not blocking — but it is the same bug, and leaving two of four fixed is worse than fixing all four.

**N4 — `PAIR_HELP` still describes only the old shape.** `actana-pair.ts:118`: *"`pair new` prints three things: the code, this Core's CA fingerprint, and when the code expires."* At a terminal it now also prints both redemption paths and a pasteable command — which is the feature, and this help is where an operator would go looking for it. Two sentences under "Minting a code".

**N5 — the one unverified acceptance criterion has nowhere to be verified.** You are right that *"the pasteable line actually works when pasted on a second machine"* cannot be checked from this working copy, and right to say so plainly rather than tick it. But `docs/core-linux-rehearsal.md` §2 (`:130`) is the manual pass this repo already runs before a release, it is run at a terminal, and §3 goes straight to a Panel without ever exercising the CLI path. A checkbox there — paste the printed `actana core pair …` line on a second machine, confirm it pairs, then confirm `actana core status` reaches the Core — would turn "worth one manual pairing before this leaves the train" from a PR comment into something the release process actually catches. That checkbox is also where blocking 2 and blocking 3 would both have surfaced.

---

### Acceptance criteria

| # | Criterion | Verdict |
|---|---|---|
| 1 | TTY prints a cosmetic, instruction-rich block | **Met** |
| 2 | Framed block, code + fingerprint prominent, expiry and session | **Met** |
| 3 | Fingerprint never truncated | **Met** |
| 4 | Both redemption paths, always, in one block | **Met** |
| 5 | Panel path worded as the issue words it | **Not met** — blocking 1 |
| 6 | CLI path: `npm i -g @actana/cli`, then a *fully pasteable* line with real values | **Partially met** — real values throughout, but the unlabelled default is not pasteable (blocking 3) |
| 7 | One line per configured public host, endpoint filled in | **Partially met** — one line each, but N-1 name an endpoint the client will not keep (blocking 2) |
| 8 | `--label` for the name, placeholder otherwise | **Met** — extending the placeholder to labels that fail `coreNameError` is a good call |
| 9 | Non-TTY output byte-identical to 0.4.2 | **Met** — whole-array literal assertion |
| 10 | stdout=facts / stderr=prose preserved | **Met** |
| 11 | No `--json`; TTY detection is the whole switch | **Met** |
| 12 | Pasteable line works when pasted on a second machine | **Not verified** — honestly declared; see N5, and note blocking 2 and 3 are both things this would have caught |
| 13 | `core-pair.ts` usage string shows `--session` | **Met** (see N3) |
| 14 | `core-command.ts` help shows `--session` | **Met** (see N3) |
| 15 | The new block's pasteable line carries `--session` | **Met** |

---

### Verdict

**Changes requested.** The engineering underneath this is careful, and the hard part — two shapes out of one command with the scrapeable one provably unmoved — is done properly and tested properly, with the switch confined to a single `if` and a single new dep field. Every claim about the pipe contract checks out against the enforcing lines.

The three blockers are all in the instructions rather than the machinery, which is unfortunate here because on this ticket the instructions *are* the deliverable: the Panel path tells an operator to skip a step the Panel requires, the terminal path hands them a command that pairs to a different endpoint than the one it names, and on the default invocation that command does not survive being pasted. Blocking 1 is one constant plus its two pins. Blocking 3 is one string. Blocking 2 is a line in `handoutHosts` or a sentence in the per-host comment. None of them touches the pipe contract, so the expensive half of this PR stands as-is.

Self-reporting the deviations rather than burying them is what made this reviewable — thank you for that, and for the honest "not checked end-to-end" on criterion 12 instead of a tick.

qcentic-dev:
## Review addressed — `e28e635` and `ae69dd0`

Thank you for reading the enforcing lines rather than the prose. All three blockers were real, all three are fixed, and you were right that they were the expensive part of this ticket even though they are the cheap part of the diff.

---

### B1 — the Panel instruction now describes the form

I read `AddCoreByPairing.tsx` rather than arbitrate wording. The form's real order is **address → check/compare the CA fingerprint → (gated) Session → Pairing code**, with the code field unreachable until the fingerprint verifies.

Your suggested line was close, but it stops at "then the code" and the form has a **Session** field between the fingerprint and the code — `placeholder="the Session line from actana pair new"` at `:216`. An operator following a line that named address, fingerprint and code would still hit one unexplained required field. So:

```
Settings (gear icon) -> Cores -> Add Core: this Core's address, then compare the
CA fingerprint, then the session and the code
```

Every field the form requires, in the form's order, with the comparison kept — it is the security-relevant step and the block prints the fingerprint prominently without otherwise saying what it is for.

I did **not** say "the address above", though: the frame does not print an address as a field. It appears only in the `From a terminal` commands below. Naming it as something to supply rather than something to read off is the honest version. Adding an address row to the frame would be the better fix and is a bigger change than a review round should carry — happy to do it if you want it.

Pins updated together: `actana-pair.test.ts` (which now also asserts the four fields appear **in the form's order**, so a future paraphrase that reorders them fails) and `e2e-actana-setup-linux.mjs:434`.

On **#363 mounting this same component in a first-run wizard** — agreed, and out of my reach here: the wizard is not on this base. Flagging it on that PR is the right move, and this wording survives it better than the old one did, since everything after the colon is about the form rather than the route to it.

### B2 — took **option (b)**, annotate, and here is why

You called either fine; I took **(b)**, keeping all N commands and saying which endpoint each leaves behind. Three reasons:

1. **The issue asks for one line per configured public host, in as many words.** (a) would quietly repeal a stated requirement to fix a truthfulness bug, and the truthfulness bug is fixable without repealing it.
2. **(a) removes the information rather than the untruth.** A laptop that cannot resolve `core` still needs `10.0.0.5:8443` to pair *at all* — the dial has to reach the Core, and the primary may not. Under (a) that operator is back to assembling a command by hand, which is the exact failure #357 exists to remove. They would just fail earlier and with less help.
3. **The annotation can be made honest and short**, which was your stated bar and the condition in my brief. It is two lines and it teaches the flag:

```
  # dial core:8443 — and that is the endpoint this code registers
  actana core pair laptop core:8443 …

  # dial 10.0.0.5:8443 — but this code still registers core:8443
  #   to register 10.0.0.5:8443: actana pair new --label laptop --public-host 10.0.0.5
  actana core pair laptop 10.0.0.5:8443 …
```

The comparison is against the resolved credential endpoint, not against the array index, so `--public-host 10.0.0.5` produces one command with the *first* shape — dial and endpoint agree, nothing to warn about. `credentialHost` is computed the same way `core-pairing-wiring.ts:107-112` computes it (`endpointHost ?? primary`), and its doc comment says so, so if that resolution ever changes there is one place naming the other.

Tests pin all of it: the primary's "and that is the endpoint" line, both non-primary warnings with their re-mint commands, `--label` dropped from the hint when there is no usable label, the `--public-host` case warning about nothing, and `not.toContain("reachable at 10.0.0.5")` so the specific untruth cannot come back.

### B3 — `NAME`, and a test that actually runs a shell

Your reproduction was exactly right, and you were right that the spec's own promise should win over the spec's wording. It is bare `NAME` now, and `NAME_PLACEHOLDER` is one exported constant.

The assertion you identified as missing is the one I care most about here. Rather than assert the string, the test hands the emitted line to a real shell:

```js
const echoed = execFileSync("/bin/sh", ["-c", `printf '%s\\n' ${command}`], …);
expect(echoed.split("\n").filter(Boolean)).toEqual(command.split(" "));
```

A redirection, a glob or a stray quote each consume or move a word and fail it. Run for all three label cases, with a metacharacter sweep as belt-and-braces — and repeated on the real machine in the e2e, for the labelled *and* the unlabelled line, because "survives a shell" is a claim about a shell.

One correction to your note: `NAME` passes `coreNameError`, so pasted unedited it registers a Core called `NAME` rather than producing a refusal. That is still much better than a shell error — one `actana core rm NAME` undoes it — and it is asserted, so nobody rediscovers it.

---

### Non-blocking

**N1 — fixed.** Padding is by display columns now (`displayWidth`: code points, wide ranges at 2, combining marks and variation selectors at 0), and the **label** row clips with `…` rather than bending the border — it is the one row with no bound, and a label is a name the operator chose and can read off `pair ls`. The fingerprint still wraps and is still never shortened; there is a test asserting that a 200-character label clips while the fingerprint on the same screen re-joins to exactly `fingerprint256`, and one asserting `笔记本` keeps every framed row at 74 columns. The full-`wcwidth` caveat is written into the function: emoji with joiners can still measure a column off, and the cost of that is a ragged border, never a truncated fingerprint.

**N2 — fixed, both halves.** The leak sweep is back to both streams and now covers all three runs (redirected, framed, unlabelled), naming which one it found something on. `/tmp/pair-new.out` is removed right after the scrape.

**N4 — done.** `PAIR_HELP`'s "Minting a code" gains two paragraphs: what the framed shape adds, and that redirecting stdout is the whole of the switch with no flag for it.

**N5 — done, and it is the best thing in this round.** `docs/core-linux-rehearsal.md` §2 grows a *"The pasteable line, on a second machine"* block: paste unedited, no shell error, the `NAME` case runs, it pairs, and `actana core status` reaches the Core from that second machine — plus a multi-address item checking the endpoint warning behaves as it says. You were right that this is where B2 and B3 would both have surfaced.

**N3 — deferred to a follow-up ticket.** You are right that it is the same bug and that two-of-four is an odd place to stop. `core-command.ts:182` and `core-resolution.ts:124` are the *empty-registry and no-Core-selected* refusals, not #357's usage strings, and they sit in the surface #360 owns — the `core pair` result and refusal surface. Fixing them here would mean editing `core-resolution.ts`, a file this PR has no other business in, on a ticket whose stated boundary is `actana-pair.ts` plus the two named usage strings. It is a two-line change and it should be a two-line ticket that sweeps for *all* copies of the flagless invocation rather than the two you happened to find — I would rather it land as a grep-driven fix than as a third pair of hand-picked lines. I will file it referencing this comment; say the word if you would rather it ride here and I will pull it in.

---

### The expensive half is untouched, and re-verified

You asked for this to survive, so I checked it rather than assumed it:

- `git diff origin/feat/0.4.3-onboarding -- actana-pair.ts` on the `deps.out` templates shows the same six lines, same order, same columns, same two conditionals — the only delta is the indentation from the `if` and the hoisted `fingerprint` local.
- `git grep stdoutIsTty` outside tests still returns exactly the entry-point binding, the one shape branch, and the colour gate reached only from inside it.
- The whole-array test ran green on its own: all five `actana pair new, piped` cases plus *"mints exactly what the piped shape mints — only the printing differs."*

`pnpm test` — all 6 stages pass (cli: 50 files, 1146 tests). `pnpm typecheck` clean. `pnpm lint` 0 errors, 0 findings in any file this change touches.

qcentic-dev:
### CI green on `ae69dd0`, and N3 is filed

**18/18 checks pass**, 6 skipped (train-publish jobs that do not run on a PR): `Unit Tests`, `Lint`, `Typecheck`, `Conventions` (all five commits), `Train rules`, `Secret Scan`, `Dependency Audit`, `E2E — Panel service seam`, both `E2E — installer` jobs, both image smokes, both tarball smokes.

The installer E2E now logs three lines for this ticket, on a really-installed Core:

```
[setup-e2e] `actana pair new` keeps its labelled lines when stdout is not a terminal (#357)
[setup-e2e] both pasteable commands survive the machine's own shell, placeholder included (#357)
[setup-e2e] `actana pair new` frames the code and both redemption paths at a terminal (#357)
```

The middle one is B3 checked where it actually matters: the labelled line *and* the unlabelled one — the `NAME` placeholder case — handed to the machine's own shell and compared word for word against what was printed.

**N3 is now #365**, scoped as a grep-driven sweep for every copy of the flagless invocation rather than the two hand-picked lines, with a note that `core-resolution.ts` and the empty-registry refusal sit in #360's surface so whoever picks it up should check there first.

Still a **draft**, base unchanged, nothing merged. Over to you.

mrfqcentic (approved):
## Re-review of #364 — `e28e635` + `ae69dd0` — approved

All three blockers are fixed, and fixed at the level the finding was about rather than at the level of the symptom. Verified against the enforcing lines, not the reply. Two of your three responses corrected *me*, and both corrections are right.

---

### B1 — resolved, and your wording is better than mine

`packages/cli/src/actana-pair.ts:658-660`:

```
"Settings (gear icon) -> Cores -> Add Core: this Core's address, then compare the " +
"CA fingerprint, then the session and the code"
```

I checked this against the form rather than against your description of it. `AddCoreByPairing.tsx`:

- `:176` — `TextField label="Core address"`, with the `Check fingerprint` button beside it;
- `:198-210` — `FingerprintPanel`, carrying the typed `CA fingerprint from \`actana pair new\`` field and *"the Panel does not send the code until they match"*;
- `:213` — `{verified && (` — step 3 does not exist until the comparison passes;
- `:215-223` — `TextField label="Session"`, `placeholder="the Session line from \`actana pair new\`"`;
- `:224+` — `TextField label="Pairing code"`;
- then an optional name.

**Your correction is right and mine was wrong.** I checked whether Session is genuinely required rather than merely present: `:75` — `const codeReady = normalizePairingCode(code) !== null && sessionId.trim() !== "";` — and `:268` gates the Pair button on `!verified || !codeReady`. A blank Session leaves the button disabled. My suggested line stopped at "then the code" and would have sent an operator to a required field nothing had mentioned — the exact failure I opened B1 about, one field further along. Good catch, and the right instinct: you read the component instead of applying a reviewer's line verbatim.

**"this Core's address" rather than "the address above" is also right,** and for the reason you give: the frame prints code, expiry, fingerprint, session and label, and no address row. "The address above" would have pointed at something that is not there. Naming it as a thing to supply is the honest version. I agree that adding an address row to the frame is the better fix and is out of scope for a review round — worth its own ticket if you want one, since it is the one field of the four the block still makes the operator source elsewhere.

The test is stronger than a string pin: `actana-pair.test.ts:580-606` asserts the constant `toBe` the literal **and** asserts `indexOf` ordering `address < fingerprint < session < code` plus the presence of `compare`, so a future paraphrase that drops a field or reorders them fails rather than passes. `e2e-actana-setup-linux.mjs:443-447` pins the same value by literal, with a comment explaining why it cannot import the constant. Both moved with the line.

### B2 — annotation clears the honesty bar, and (b) was the better call

`hostNotes` at `packages/cli/src/actana-pair.ts:614-632`:

```
# dial core:8443 — and that is the endpoint this code registers

# dial 10.0.0.5:8443 — but this code still registers core:8443
#   to register 10.0.0.5:8443: actana pair new --label laptop --public-host 10.0.0.5
```

**This meets the bar, and exceeds what I asked for.** My condition was that the block state which endpoint the credential carries and how to get one that carries the dialled address. It does both, and it separates the two questions in the operator's own vocabulary — *dial* versus *registers* — which is the distinction the old "reachable at X" collapsed. The re-mint line is a complete command, not a hint at a flag.

**I accept your argument for (b) over (a), and it is better than my framing of the choice.** Your point 2 is the one I had not weighted: the redemption dial has to actually reach the Core, so a machine that cannot resolve the primary needs the non-primary address to pair *at all*. Under (a) that operator has no command and is back to assembling one by hand — the failure #357 exists to remove — so (a) would have traded a misleading command for no command. Point 1 stands too: the issue asks for one line per configured host in as many words, and the truthfulness bug was fixable without repealing it.

Keying on `endpoint === credentialEndpoint` rather than on the array index is the right construction — it is why `--public-host` yields one command with the *agreeing* shape and nothing to warn about, instead of a special case. `credentialHost` (`:306-310`) is `endpointHost ?? primaryPublicHost(material.serverHosts)`, and its doc comment at `:512-522` names `core-pairing-wiring.ts` as the authority. That matches `:107-112` there. The one divergence — wiring re-checks `isConfiguredPublicHost` for a code that outlived a shortened `ACTANA_PUBLIC_HOST` — is not knowable at mint time and cannot make this block's claim wrong at the moment it prints. Correctly scoped.

Tests pin every branch: `:650-676` (multi-host, primary's agreeing line, both warnings with their re-mint commands, and `not.toContain("reachable at 10.0.0.5")` so the specific untruth cannot return); `:678-686` (`--label` dropped when unusable); `:688-700` (`--public-host` — asserts `store().listSessions()[0].endpointHost === "10.0.0.5"` **and** that nothing warns, which ties the comment to the stored session rather than to the rendering); `:618-624` (single host). The old phrasing survives only in a comment at `:598` explaining why it went.

### B3 — resolved, with a test that proves the thing rather than describing it

`NAME_PLACEHOLDER = "NAME"` at `:719`, one exported constant, reached through `handoutName` at `:704-707`.

`actana-pair.test.ts:761-778` is the assertion I said was missing, and it is done properly — the emitted line is handed to `/bin/sh` and the words a shell actually parsed are compared to the words that were printed:

```js
const echoed = execFileSync("/bin/sh", ["-c", `printf '%s\n' ${command}`], …);
expect(echoed.split("\n").filter(Boolean)).toEqual(command.split(" "));
```

A redirection, a glob or a stray quote each consume or move a word and fail it. Run across all three label cases, backed by the metacharacter sweep at `:780-789`, and repeated on the box in `e2e-actana-setup-linux.mjs:466-490` for the labelled *and* the unlabelled line — the `NAME` case checked with `/^actana core pair NAME /` before being fed to the machine's own shell. "Survives a shell" being a claim about a shell is exactly right.

**On your correction — yes, `NAME` passes `coreNameError`, and I accept the trade.** I verified it rather than take it: `blob-registry.ts:87` is `/^[a-zA-Z0-9][a-zA-Z0-9._-]*$/`, which `NAME` satisfies, and `:773-777` asserts `coreNameError(NAME_PLACEHOLDER)` is null so nobody rediscovers it. An unedited paste pairs and registers a Core called `NAME`, visibly — `Paired Core "NAME" → wss://…` is the last line the operator sees. That is strictly better than a `bash: name: No such file or directory` that names neither `actana` nor the problem, and it is the right call.

One precision on the recovery cost, since it is worth having on the record rather than as a surprise: `actana core rm NAME` removes the entry, but there is no `core rename` (`CORE_HELP` lists `pair`, `ls`, `use`, `rm`, `status`, `shell`, `exec`), and the code is one-time and now spent — so re-pairing under the intended name needs a freshly minted code, which in the read-it-out-over-a-phone case means going back to the Core operator. Small, and visible, and I would not hold the PR for it.

If you ever want the third option: a placeholder that is shell-safe *and* fails `coreNameError` — `YOUR:NAME`, say, since `:` is not a shell metacharacter and is not in that charset — would make an unedited paste refuse at `runCorePair`'s name check, which runs before `readTicket` and before `deps.pairing.pair`, so the code is never spent. Genuinely your call; `NAME` is a correct resolution of B3 as it stands and I am not asking for a change.

---

### Non-blocking items

**N1 — fixed, and the right row gives.** `displayWidth` (`:797-806`) counts per code point with wide ranges at 2 and combining marks / variation selectors at 0, and `frameRow:776` pads by it. `clip` (`:846-858`) is called on the label alone (`:569-576`), and the room calculation reserves a column for the marker so the row lands at exactly 74. The tests are the convincing part: `:809-816` asserts every framed row is 74 *columns* for `笔记本`; `:818-832` asserts a 200-character label clips with `…` while every row stays 74, the store keeps the full label, **and the piped shape still prints it whole** — so the clip is provably cosmetic and cannot reach the scrapeable output; `:834-842` asserts that on the same screen the framed fingerprint pieces still re-join to exactly `fingerprint256`. Truncating the label and never the fingerprint is the correct choice and you gave the correct reason for it. The `wcwidth` caveat written into the function is honest and correctly bounded — I checked, and e.g. `U+1F680` falls outside your wide ranges, so it measures a column short; the cost is one ragged border, never a shortened fingerprint. Documented limitation, appropriate scope.

**N2 — fixed, both halves.** `e2e-actana-setup-linux.mjs:405-424` keeps the redirected run's own result, and `:493-501` sweeps `BEGIN (CERTIFICATE|PRIVATE KEY)` over four things — the scraped file, the redirected run's `stdout+stderr`, the framed run's, and the unlabelled framed run's — naming which one it found something on. Broader than what was there before I raised it. `rm -f /tmp/pair-new.out` at `:409` immediately after the scrape, with a comment saying why an ephemeral container is not a reason to leave a code lying about.

**N4 — done.** `PAIR_HELP:130-139`, two paragraphs: what the framed shape adds, and that redirecting stdout is the whole of the switch with no flag for it. That second paragraph is the one that matters, because it is the answer to the pty question anyone will arrive with.

**N5 — done, and it is the best thing in this round.** `docs/core-linux-rehearsal.md` §2 gains the framed/redirected pair, and a new *"The pasteable line, on a second machine"* block with five checkboxes: pastes and runs with no shell error, the `NAME` case runs, it pairs, `actana core status` **reaches the Core from that second machine**, and — on a multi-address Core — the endpoint warning behaves as it says. That closes acceptance criterion 12 with a check a person will actually perform, and it is precisely where B2 and B3 would each have surfaced. The multi-address item in particular turns B2's annotation from a string a test pins into a claim someone verifies.

**N3 — deferral accepted, and #365 is scoped better than my finding was.** You are right that `core-command.ts:182` and `core-resolution.ts:124` are the empty-registry and no-Core-selected refusals rather than #357's usage strings, and right that `core-resolution.ts` is a file this PR otherwise has no business in. A grep-driven sweep for every copy of the flagless invocation is a better ticket than the two lines I happened to find. No need to pull it in.

---

### The expensive half — re-verified independently

I did not take this on the reply:

- **Piped output.** The six `deps.out` templates at `actana-pair.ts:323-331` are byte-for-byte the strings at `feat/0.4.3-onboarding:271-279` — same order, same columns, same two conditionals — with the hoisted `fingerprint` local as the only delta, unchanged since the first round.
- **The whole-array test is untouched.** All nine hunks in `actana-pair.test.ts` land in the imports or in `describe("actana pair new, at a terminal")`. The `piped` block and its hand-written `expectedLines()` literal are not modified by this round at all.
- **The switch.** `git grep stdoutIsTty` outside tests still returns exactly three lines: the entry-point binding (`actana-cli-entry.ts:147`), the shape branch (`actana-pair.ts:297`), and the colour gate (`:484`) reached only from inside it. No new reader, and none of the new code — `credentialHost`, `hostNotes`, `displayWidth`, `clip` — touches it.
- **CI on `ae69dd0`:** 18 pass, 6 skipped (train-publish jobs that do not run on a PR).

---

### Acceptance criteria — final

| # | Criterion | R1 | Now |
|---|---|---|---|
| 5 | Panel path worded as the issue words it | Not met | **Met** — and worded from the form, which is better than the issue's own line |
| 6 | Fully pasteable line with real values | Partial | **Met** — proven through a real shell, in the suite and on the box |
| 7 | One line per configured host, endpoint filled in | Partial | **Met** — and each now states the endpoint the credential carries |
| 12 | The line works when pasted on a second machine | Not verified | **Verifiable** — still not automatable, now a rehearsal checkbox that covers B2 and B3 |
| 1–4, 8–11, 13–15 | — | Met | **Met**, re-verified for 9/10/11 |

**15 of 15 met**, with 12 met in the only form it can be: a manual check that now exists and is specific.

---

### Verdict

**Approved.**

Every blocker is fixed at the root: the Panel line is derived from the form and pinned by field order rather than by string; the endpoint annotation separates dialling from registering and carries the fix, not just the warning; and the placeholder claim is proven by handing the line to a shell in two places rather than asserted about a string. The nits came back with more than was asked — the label clip is proven not to reach the piped shape, and the leak sweep is wider than it was before I raised it.

Two of your three responses corrected my findings, and both corrections were right: the Session field my suggested wording would have dropped, and the `NAME`-registers-a-Core consequence. Saying so plainly, with the reasoning, is what made this round short.

The expensive half — byte-identical piped output and a one-`if` switch — is untouched and re-verified. Still a draft on `feat/0.4.3-onboarding`, as it should be; the rehearsal checkbox is the last thing standing between this and the train, and it is now written down.

Issue #365, opened 2026-08-31
actana core pair: two more usage strings still omit the required --session flag


qcentic-dev:
Split out of the review of #364 (#357), where two of the four copies were fixed and these two were left as out of scope for that ticket.

## Problem

`readTicket` in `packages/cli/src/core-pair.ts` refuses a bare pairing code that does not name its pairing session: it wants `--session <id>`, or the joined `<session>:<code>` form. Several places print the invocation without it, so an operator who types exactly what they were shown lands on a second refusal for a flag nobody mentioned.

#357 named two of them and they are fixed on #364:

- `packages/cli/src/core-pair.ts` — the "a name, an address and a code are required" usage line.
- `packages/cli/src/core-command.ts` — the `Usage` block in `CORE_HELP`.

Two more print the identical flagless invocation and are still wrong:

- `packages/cli/src/core-command.ts:182` — ``"No Cores registered. `actana core pair <name> <address> <code>` registers one."``
- `packages/cli/src/core-resolution.ts:124` — the same sentence, for any noun run with no Core selected.

`actana core ls` on an empty registry, and any client noun on a machine with nothing paired, are both first-run paths — so this hits people at exactly the moment they have the least context to recover from it.

## What to build

Fix both, **and sweep for the rest** rather than fixing the two named here. The whole point of splitting this out is that hand-picked lines is how two got missed the first time: grep the repo for the invocation (source, help text, docs, `INSTALL.md`, `README.md`, the rehearsal checklists) and make every copy that shows a full command line carry `--session <id>` and `--fingerprint <sha256>`.

Worth considering while in there: one exported constant for the canonical invocation, so a fifth copy cannot drift.

## Acceptance

- `actana core ls` on an empty registry prints an invocation that works as printed.
- A client noun with no Core selected does the same.
- A repo-wide grep for `core pair <name>` / `core pair <name> <address>` finds no copy missing either required flag.
- A test pins each fixed string, as #364 did for the two it fixed.

## Notes

`core-resolution.ts` and the empty-registry refusal are part of the `core pair` result-and-refusal surface that #360 owns, so whoever picks this up should check in with that ticket first — it may be cheaper to land there.

PR #366, opened 2026-08-31
feat(cli): frame every `actana core pair` result — one remedy per failure, the pipe unchanged

### LINKED ISSUE #360: actana core pair: framed success and failure output with next steps

qcentic-dev:
Closes #360. The client half of the onboarding pass; #364 did the Core half.

## The problem

An operator on the Core runs `actana pair new`, gets a framed handout (#364),
and carries a code, a fingerprint and a session to the second machine. They
paste the command they were given, and the moment they find out whether it
worked was:

```
Paired Core "prod" → wss://core.test:8443
`current` now points at "prod".
```

…with no hint that there is anything to do next, or, on failure, two lines of
prose where "refused" is the same word for an expired code, a spent one and a
typo.

## The shape

Two shapes, and `isatty(stdout)` is the whole of the switch — no flag, and
deliberately no `--json`, exactly as it is decided one machine over.

**Piped, redirected or captured, both streams are what 0.4.2 printed.** Same
success lines in the same order, same two-line refusals. The suite asserts
those whole against literals rather than trusting them: all three shapes the
pointer note has, and the exact line count of every failure class.

**At a terminal, success is a framed block.**

```
╭────────────────────────────────────────────────────────────────────────╮
│                                                                        │
│   ✓ Paired Core "prod"                                                 │
│                                                                        │
│   Endpoint     wss://core.example.test:8443                            │
│   Sent as      bkp-07 — the name this machine gave                     │
│   Current      "prod" — every later verb talks to this Core            │
│   Credential   /home/ada/.config/actana/cores/prod.txt (mode 0600)     │
│                                                                        │
╰────────────────────────────────────────────────────────────────────────╯

Next steps
  actana core status
    Reach the Core and report what it says. The check that the link works
    end to end.

  actana project ls
    The Projects on this Core. A Session needs one to run in.

  actana harness ls
    Which agents this Core can actually run right now, and which it is
    missing.

  actana session start <project> "<prompt>"
    The first real thing to run on it. Prints the Session id and exits.

  actana core shell
    Or an interactive shell on the Core, to look around it yourself.

  The Panel pairs with this same Core from its Settings -> Cores screen,
  with a code of its own — `actana pair new` on the Core mints that one too.
```

The `Current` row is **read back off disk after the write**, not inferred from
the branch that was taken. `core pair` makes a Core current only when nothing
was, so pairing a second Core changes no pointer — and the block says so and
puts `actana core use <name>` first, rather than congratulating an operator on
a `current` they do not have:

```
│   Current      still "prod" — this pairing did not change it           │
```

The `Sent as` row is a purely local fact: the name this machine put in the
pairing request — `--label`, or its hostname when the flag was not given. It
makes **no** claim about the far end, and the review was right that the earlier
wording did and was wrong to: the Core stores the *session's* label in its
`pair ls`, not the client's. Not called `Label`, because this program already
has two LABEL columns meaning two other things.

The credential row is reassurance, never the blob: the path and the mode, and
the "never logs a blob" sweep still covers this verb.

## Every failure class ends in a command

One remedy per distinguishable failure, because "ask for a fresh code" and
"check the port is open" are answers to different problems — a table that gave
both to both would be the prose it replaced. A test asserts no two classes
share a remedy.

| Class | What it now says |
| --- | --- |
| `refused` — expired, spent, unknown, out of attempts | that the Core deliberately will not say which, `actana pair new --label <name>`, and **re-run with the NEW code *and* the NEW session** |
| `fingerprint-mismatch` | what was compared against what, that the code was not sent, and that there is no flag that skips the comparison and none is coming |
| `unreachable` | a **dial** failure, not a refusal — *if nothing answered*, nothing was sent and the code is still good; a drop part-way through may have spent it — plus `getent hosts`, `nc -vz`, `actana status`, and the proxy/SNI cases |
| `not-pairable` | the other side of that: the dial worked, so update the Core or check the port |
| a bare code | both `--session <id>` **and** the joined `<session>:<code>` form, because an operator has been handed one or the other |
| `rate-limited`, `certificate-invalid`, `hostname-mismatch`, `core-error`, … | their own concrete check — `date -u` on both machines, `actana logs` on the Core, and so on |

Colour follows #364: green tick, red cross, dim headers, `NO_COLOR` honoured,
and no escape ever emitted off the TTY path.

## Reuse, not a second copy

#364's frame primitives — the border width, the padding rule, the colour table,
the display-width measurement — moved from `actana-pair.ts` to a shared
`cli-frame.ts`, and both callers import them. Two copies of a border width is
how the two ends of one handshake start drawing different boxes.

`actana-pair.ts` keeps its handout, its wording and its one `if`, and
`actana-pair.test.ts` is **untouched and still passes whole** — the piped
byte-identity and the framed content it pins are the proof that the lift is a
lift and not a rewrite.

One behaviour change in the shared code: `wrapText` now breaks an over-long
word on its slashes rather than running past the border, the same move
`wrapFingerprint` makes on colons. A credential path under a deep home
directory is ordinary, not pathological — this was caught by the suite running
under a longer temp root. The rule that it may never *shorten* its input still
holds.

## Review round 1 (`b672fcd`)

[The review](https://github.com/actana/control/pull/366#pullrequestreview-5069784593)
found two blocking defects; both are fixed, along with the two should-fix items
and the cheap notes. The detail is in
[this reply](https://github.com/actana/control/pull/366#issuecomment-5482885492).

- **The frame shattered on `fingerprint-mismatch`** — the SDK's sentence carries
  two 95-column colon-hex fingerprints and `wrapText` broke only on `/`.
  `WORD_BREAKS` is now `/`, `:` and `.`; the docstring describes what the
  function actually does; the width test renders the real SDK sentence.
- **The `Label` row named a value the Core does not keep** — reworded to
  `Sent as`, a purely local fact claiming nothing about `pair ls`.
- **`unreachable` promised the code was unspent** — scoped to *if nothing
  answered*, with the post-send case spelled out.
- **`actana core shell` is back** in the next steps.

## What is not here

- `actana pair new`'s output and its handout — #364, landed.
- `packages/panel` — #358, landed.
- The flagless refusal lines in `core-command.ts` and `core-resolution.ts` —
  they belong to #365's sweep.

## Testing

- `pnpm test`: all six stages pass (1169 CLI tests, 3 skipped).
- `pnpm typecheck`, `pnpm lint`: clean — 0 errors, and no finding in any file
  this branch touches.
- 28 new tests in `core-pair.test.ts`, split into the piped contract
  (whole-output assertions) and the framed content per failure class.
- `docs/core-linux-rehearsal.md` gains the checkboxes for the half only two
  machines can check: that the four next steps actually run, that a redirected
  pair is still two plain lines **on both streams**, and that a spent code and
  a dead port give different remedies.

### DIFF
--- a/docs/core-linux-rehearsal.md
+++ b/docs/core-linux-rehearsal.md
@@ -171,9 +171,15 @@ actana core status
 - [ ] Because you are at a terminal on that machine too, the pairing **result**
       comes framed (#360): a green `✓`, the endpoint, an honest `Current` row,
       where the credential landed at mode 0600 — and under it `actana core
-      status`, `actana project ls`, `actana harness ls` and `actana session
-      start`. Run the four. They are the checkbox: a next step that does not
-      work is worse than no next step.
+      status`, `actana project ls`, `actana harness ls`, `actana session
+      start` and `actana core shell`. Run all five. They are the checkbox: a
+      next step that does not work is worse than no next step.
+- [ ] The `Sent as` row names what this machine put in the request — the
+      `--label` you passed, or its hostname. It claims nothing about the Core,
+      and it should not: `actana pair ls` **on the Core** lists the label the
+      *code* was minted with, which is a different string. Check both screens
+      and confirm the client never told you to look for a name that is not
+      there (#366 review 2).
 - [ ] Pair a **second** Core on that machine. The block says `current` is still
       the first one and offers `actana core use <name>` — it does not claim a
       `current` the pairing did not take.
@@ -185,9 +191,15 @@ actana core status
       red, and ends in `actana pair new --label <name>` plus the reminder to
       re-run with the NEW code *and* the NEW session — not prose alone.
 - [ ] Run it against a port nothing listens on. That refusal says **dial**
-      failure rather than refusal, keeps the code alive, and offers `getent
-      hosts` / `nc -vz` — a different remedy from the one above, which is the
-      whole point of the table.
+      failure rather than refusal, scopes the "your code is still good"
+      promise to *if nothing answered*, and offers `getent hosts` / `nc -vz` —
+      a different remedy from the one above, which is the whole point of the
+      table.
+- [ ] Run it with a `--fingerprint` that is wrong by one byte. The refusal is
+      framed and **the box holds**: the real SDK sentence carries two full
+      95-column fingerprints, both wrapped on their own colons and neither
+      shortened, and every border lands in the same column (#366 review 1).
+      This is the one screen an operator compares character by character.
 - [ ] `NO_COLOR=1` on any of the above: same words, same box, no escapes.
 
 ---
--- a/docs/core-linux-rehearsal.md
+++ b/docs/core-linux-rehearsal.md
@@ -168,6 +168,27 @@ actana core status
       --public-host <addr>` re-mint that would register the one you dialled.
       Pair on a non-primary address and confirm `core status` behaves the way
       the warning said it would.
+- [ ] Because you are at a terminal on that machine too, the pairing **result**
+      comes framed (#360): a green `✓`, the endpoint, an honest `Current` row,
+      where the credential landed at mode 0600 — and under it `actana core
+      status`, `actana project ls`, `actana harness ls` and `actana session
+      start`. Run the four. They are the checkbox: a next step that does not
+      work is worse than no next step.
+- [ ] Pair a **second** Core on that machine. The block says `current` is still
+      the first one and offers `actana core use <name>` — it does not claim a
+      `current` the pairing did not take.
+- [ ] `actana core pair … > /tmp/paired.txt 2>&1` instead gives the two plain
+      lines and nothing else — no frame, no next steps, on either stream. That
+      is the contract every script wrapping this command depends on.
+      (`rm /tmp/paired.txt` after.)
+- [ ] Run it once more with the **same, now spent** code. The refusal is framed,
+      red, and ends in `actana pair new --label <name>` plus the reminder to
+      re-run with the NEW code *and* the NEW session — not prose alone.
+- [ ] Run it against a port nothing listens on. That refusal says **dial**
+      failure rather than refusal, keeps the code alive, and offers `getent
+      hosts` / `nc -vz` — a different remedy from the one above, which is the
+      whole point of the table.
+- [ ] `NO_COLOR=1` on any of the above: same words, same box, no escapes.
 
 ---
 
--- a/packages/cli/src/__tests__/core-pair.test.ts
+++ b/packages/cli/src/__tests__/core-pair.test.ts
@@ -17,6 +17,8 @@
 import { describe, it, expect, afterEach } from "vitest";
 import { existsSync, statSync } from "node:fs";
 import { coreBlobPath, readCurrentCore } from "../blob-registry.ts";
+import { displayWidth, FRAME_WIDTH } from "../cli-frame.ts";
+import { corePairingOutcome } from "../core-pair-results.ts";
 import { resolveCore } from "../core-resolution.ts";
 import { CorePairingError, type CorePairingFailure } from "@actana/sdk/core-pairing.ts";
 import {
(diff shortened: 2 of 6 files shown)


qcentic-dev:
## Acceptance criteria from #360

Every item in the issue, checked or explicitly not, with the reason.

### Acceptance (the three the issue lists under **Acceptance**)

- [x] **Success on a TTY shows the framed block with next steps; piped success
      output is unchanged.** Framed block: `core-pair-results.ts`
      `corePairSuccessBlock`. The switch is `deps.stdoutIsTty`, one `if`, in
      `core-pair.ts`. Piped: asserted whole against literals — all three shapes
      the pointer note has (first Core, second Core, re-pair of the current
      one), plus a test that the piped path emits no escape, no box drawing and
      no next steps.
- [x] **Each failure class ends with a copyable command, not only an
      explanation.** Sixteen classes — thirteen the SDK distinguishes plus the
      six this file words itself — each with its own `steps`. A test walks
      every SDK failure and asserts a red marker, a `What to do` section and at
      least one unstyled command line; a second test asserts no two classes
      share a remedy.
- [x] **Nothing secret is ever printed in either mode (existing invariant in
      `core-pair.ts` holds).** `never-logs-a-blob.test.ts` passes untouched. A
      new test drives the framed *and* the piped path and asserts the code
      reaches no sink in either, in any of the shapes it is accepted in. The
      credential row prints the path and the mode, never the contents.

### On success — the framed, coloured confirmation block

- [x] **The Core's name** — the headline, `✓ Paired Core "prod"` (or
      `Replaced`, the same word the piped line uses).
- [x] **Label** — added in 66f8b25 as the `Label` row, and it is the label that
      means something here: what `--label` sent to the Core, defaulting to the
      hostname, which is how this machine appears in the Core's own
      `actana pair ls`. Deliberately not the LABEL column of `actana core ls`,
      which means the *Core's* alias and is empty for a paired credential; the
      row says which end it is talking about for exactly that reason.
- [x] **Endpoint** — the `Endpoint` row, off the issued blob.
- [x] **Whether `current` now points at it** — the `Current` row, and it is
      **read back off disk after the write** rather than inferred from the
      branch taken. Three sentences: it is current, nothing is selected, or it
      is still somebody else and this pairing did not change that. Asserted
      against `readCurrentCore` in the suite.
- [x] **Next steps: `actana core status`** — first, because it is the one that
      says the link works at all.
- [x] **Next steps: `actana core use <name>` when it is not current** — present
      only then, and *first* when it is, because every verb under it reads
      `current`. Asserted both ways.
- [x] **Next steps: `actana session start …` as the first real action** — as
      `actana session start <project> "<prompt>"`, last in the list.
- [ ] **Next steps: `actana core shell`** — **not included, deliberately.** The
      implementation brief for this pass named the four verbs the block should
      teach: `core status`, `project ls`, `harness ls` and `session start`, and
      that list is what shipped. `project ls` and `harness ls` are there in its
      place, and they earn the room: a `session start` needs a Project name and
      a Harness the Core can actually run, so the two `ls` verbs are what make
      the last step runnable. Adding a fifth and sixth line is a one-line change
      if a reviewer wants `core shell` back.
- [x] **A pointer to pairing the Panel with the same Core** — the last step:
      the Panel pairs from its Settings -> Cores screen with a code of its own,
      minted by `actana pair new` on the same Core.
- [x] **Where the credential landed (the cores directory, mode 0600 —
      reassurance, not secrets; never print the blob)** — the `Credential` row:
      the path from `coreBlobPath`, `(mode 0600)`, and nothing else. A deep
      home directory wraps on its own slashes rather than overflowing the
      border or being truncated.

### On failure — keep the per-class explanations, frame them, end in next steps

- [x] **Keep the existing per-class explanations** — every `next` line is
      byte-for-byte what it was. It is still the second line of the piped
      refusal, and two of them are pinned by value in the suite rather than
      compared against the table that writes them.
- [x] **Frame them** — `corePairRefusalBlock`, gated on the same `isatty`.
- [x] **The exact `actana pair new` re-mint command for a stale fingerprint /
      expired code** — `actana pair new --label <name>` on `refused`,
      `rate-limited`, `bad-code`, `bad-fingerprint` and `fingerprint-mismatch`,
      each with the reason that class needs it. `refused` and `rate-limited`
      also carry the step operators get wrong: re-run with the **NEW** code and
      the **NEW** session, because the two are one credential.
- [x] **A fingerprint mismatch explains the comparison and warns not to bypass
      it** — what was compared against what, that the code was not sent, and
      that there is no flag that skips the comparison and none is coming: a
      mismatch is either the wrong machine answering or somebody between you
      and the right one.
- [x] **The `--session` reminder for a bare code** — both spellings, because an
      operator has been handed one or the other: the `--session <id>` form the
      usage line shows, and the joined `<session>:<code>` form that `pair new`'s
      framed handout pastes.
- [x] **Connectivity hints (DNS, port, SNI/proxy) for a dial failure** —
      `unreachable` says it is a dial failure and **not** a refusal (the code is
      still good, no attempt was spent), then `getent hosts <host>`,
      `nc -vz <host> <port>`, `actana status` on the Core, and the HTTPS_PROXY /
      SNI cases. `not-pairable` is the other side of the same line — the dial
      worked — and shares none of those steps; a test asserts the two do not
      bleed into each other.

### Constraints carried into this pass

- [x] **Reuse #364's primitives rather than duplicating them.** The frame
      width, gutter, padding rule, colour table, `displayWidth`, `clip` and
      `style` moved out of `actana-pair.ts` into a shared `cli-frame.ts`
      (7a29d1c) and both callers import them. `actana-pair.test.ts` is
      untouched and passes whole, which is the proof the lift is a lift.
- [x] **`actana pair new`'s output and its handout untouched** — #364's, and
      not a byte of it moved.
- [x] **`packages/panel` untouched** — #358, landed.
- [x] **The flagless refusal lines in `core-command.ts` and
      `core-resolution.ts` left alone** — #365's sweep. `core-command.ts` is not
      in this diff at all.
- [x] **Colour as in #364** — green marker on success, red on failure, dim
      headers and field labels, `NO_COLOR` honoured, and no escape ever emitted
      off the TTY path. The degrade test asserts the two runs are the same lines
      with the escapes removed, and that the borders still land in the same
      column once they are — the property a frame padded by `String.length`
      would fail.
- [x] **Checked whether anything scrapes the current success line before
      changing it.** Nothing does: `scripts/e2e-actana-setup-linux.mjs` reads
      `pair new`'s pasteable command, not `core pair`'s output, and the only
      other readers were `core-pair.test.ts`'s own `toContain` assertions —
      which are now whole-output `toEqual`s.
- [x] **Tests proving the piped byte-identity and the framed content per
      failure class** — 28 new tests, split into `actana core pair, piped` and
      `actana core pair, at a terminal`.

### Verification

`pnpm test` — all six stages pass. `pnpm typecheck` — clean. `pnpm lint` — 0
errors, 46 pre-existing warnings, none in any file this branch touches.

One observation, out of scope and not fixed here: `eslint.config.mjs` ignores
`dist/**` but not `dist-tarball/**`, so running `pnpm lint` after
`pnpm core:tarball` on the same checkout reports ~541 errors from the bundled
`packages/cli/dist-tarball/actana-cli.cjs`. CI does not hit it because it lints
a checkout that never built the tarball. Worth its own one-line ticket.

mrfqcentic (changes requested):
## Review — PR #366 against issue #360

Reviewed at `66f8b25` against `feat/0.4.3-onboarding` (`4af3b92`). Everything below was checked by reading the enforcing lines and by executing the new modules in a scratch clone — never from the PR description.

**Requesting changes.** The craft here is high and most of what the description claims is true and provable — I proved the two hardest claims outright. But two of the delivered facts are wrong, and both of them are in the parts of the screen an operator is asked to trust: the framed refusal for `fingerprint-mismatch` blows the border apart, and the `Label` row names a value the Core never records.

---

### Blocking

#### 1. `fingerprint-mismatch` destroys the frame — `packages/cli/src/cli-frame.ts:185` (`wrapText`), `packages/cli/src/core-pair-results.ts:286`

`wrapText` breaks an over-long word on `/` and nothing else. The SDK's `fingerprint-mismatch` message embeds **two 95-character colon-hex fingerprints** (`packages/sdk/src/core-pairing.ts:303`, and the redemption-dial variant built at `:670-672`), and `reportPairingFailure` (`core-pair.ts:483-494`) passes `err.message` straight in as the headline. A fingerprint has no spaces and no slashes, so it is left over-long, `frameRow` clamps `fill` to 1 (`cli-frame.ts:81`), and the box breaks. Rendered with the real message format:

```
 74 | │   ✗ https://core.example.test:8443 presented a certificate authority   │
 74 | │     with fingerprint                                                   │
104 | │     00:07:0E:15:1C:23:2A:31:38:3F:46:4D:54:5B:62:69:70:77:7E:85:8C:93:9A:A1:A8:AF:B6:BD:C4:CB:D2:D9, │
 74 | │     but                                                                │
103 | │     00:0B:16:21:2C:37:42:4D:58:63:6E:79:84:8F:9A:A5:B0:BB:C6:D1:DC:E7:F2:FD:08:13:1E:29:34:3F:4A:55 │
 74 | │     was expected — the pairing code was not sent                       │
```

104 and 103 columns inside a 74-column frame — in the one class the PR calls "the security argument", and reachable by the ordinary route (`--fingerprint` with a value that has gone stale, which is the case the copy under it discusses at length).

Three things make this worse than a stray edge case:

- The docstring immediately above says *"a label is short, **a fingerprint splits on its own colons**, and a diagnostic relayed from the SDK is a sentence of whatever length that sentence is"* (`cli-frame.ts:169`). That is `wrapFingerprint`, which stayed behind in `actana-pair.ts:735` and is not reachable from here. The comment describes a property this function does not have.
- The suite looks like it covers this and does not: `"keeps every framed line exactly the frame's width"` only ever renders `the fake Core answered <failure>`, so no headline it sees is long enough to break anything.
- `field()` has the same exposure for a long, slash-free endpoint (`core-pair-results.ts:754`).

Fix: `breakOnSlashes` (`cli-frame.ts:221`) should break on `:` as well as `/` — or lift `wrapFingerprint` into `cli-frame.ts` beside it and route a headline through it. Either way, please add a case with a real SDK-shaped fingerprint message to the width test.

#### 2. The `Label` row names a value the Core does not keep — `packages/cli/src/core-pair-results.ts:176-178`, `core-pair.ts:251`

The row renders the client's `--label` (or its hostname) and asserts *"this machine, in the Core's `pair ls`"*. On the Core, `PairedClient.label` is set from **`session.label`** — the label the *Core* operator typed at `actana pair new --label <name>` (`packages/core/src/core-pairing-routes.ts:377`, `label: session.label`; `packages/shared/src/pairing-store.ts:66` documents it as "carried over from the session"). The client's own label reaches only `certCommonName(session.label || clientLabel || sessionId)` — the SUBJECT column — and only when the session carried no label at all.

Worked through the flow this feature is for: the operator on the Core runs `actana pair new --label ada-laptop` and reads out the pasteable command, which does **not** carry `--label` (`actana-pair.ts:669-680`). On the client, `args.label ?? deps.hostname` is `bkp-07`. The frame says `Label   bkp-07 — this machine, in the Core's pair ls`. On the Core, `pair ls` shows `LABEL ada-laptop`, `SUBJECT CN=ada-laptop` (`actana-pair.ts:860-866`). **`bkp-07` appears nowhere on that Core.**

The row's stated purpose is revocation, and that does not work either: `matchesClient` (`actana-pair.ts:1027-1034`) compares against `client.label` and `certSubject`, and `certSubject` carries a `CN=` prefix — so `actana pair revoke bkp-07` matches nothing even in the sub-case where the session had no label.

The third commit's whole argument rests on this row being the name the operator will look for, and it isn't. Options as I see them: drop the row; render `Subject`/serial instead; or keep it and reword it as a purely local fact ("what this machine sent as its name") with no claim about `pair ls`. Whichever — it can't ship saying what it says now.

---

### Should fix before merge

#### 3. `unreachable` guarantees the code is unspent, and it isn't always — `packages/cli/src/core-pair-results.ts:406-408`

The new step says *"the code was never sent, no attempt was spent, and the code you were given is still good."* `unreachable` is also what `postRedemption` produces from its **post-send** error path: `req.end(opts.body)` at `packages/sdk/src/core-pairing.ts:658`, with a single 15 s `DEFAULT_PAIRING_TIMEOUT_MS` timer (`:198`, `:644`) covering the whole exchange including the Core's CSR signing, and `classifyDialFailure` (`:891`) mapping a code-less timeout Error or an `ECONNRESET` to `transport` → `unreachable`. A slow Core that signs in over 15 s, or a reset mid-flight, leaves the session spent and a certificate issued while this block tells the operator to reuse the code — which then comes back `refused`, having cost them a trip.

(The SDK's own comment at `:649-652` makes the same over-broad claim, so this is inherited — but the base only said "check the Core is running", where this promises. Scope it to "if nothing answered, nothing was sent", or drop the guarantee.)

#### 4. `actana core shell` was dropped from the next steps, silently — `core-pair-results.ts:217-249`

#360 names it: *"`actana session start …` / `actana core shell` as the first real actions"*. `nextSteps()` substitutes `project ls` and `harness ls` — both good additions — and drops `core shell`, which exists (`core-command.ts:150`). The PR body documents the `Label` deviation at length and this one not at all. Add it, or say why it went.

---

### What I verified and found sound

- **The pipe did not move.** I extracted all fifteen `next:` strings from `core-pair.ts` on the base and from `core-pair-results.ts:349-646` here and diffed them mechanically: **15/15 byte-identical**. The success path (`core-pair.ts:272-277`) is unchanged and all three pointer shapes are asserted whole with `toEqual` against literals; the six locally-worded refusals likewise. `refuse()`'s non-TTY arm (`core-pair.ts:525-528`) emits only those literals plus `err.message`/`nameError` — no escape can reach it.
- **The lift is a lift.** `useColor`, `frameEdge`, `frameRow`, `displayWidth`, `isZeroWidth`, `isWide`, `clip` and `style` are **character-identical** to the originals modulo `export`; `FRAME_WIDTH` 74 and `FRAME_GUTTER` 3 unchanged; `ANSI` gains only `green`/`red`. `actana-pair.test.ts` is untouched and green. I went further and rendered `pairingHandout()` from the base file and from this branch side by side over ASCII, empty-label, colour on/off and a CJK label, asserting `toEqual`: **identical**. `pair new` did not move.
- **`Current` is read back.** `core-pair.ts:238-259`: `before = readCurrentCore(paths)`, conditional write, then a **second** `readCurrentCore(paths)` as `current`. Framed tests assert two states against a direct `readCurrentCore(cli().paths)` check; the pipe asserts all three.
- **Nothing secret prints.** Credential row is path + `(mode 0600)` and the blob is never passed in (`core-pair-results.ts:184`). The code reaches no sink: `bad-code` is intercepted before the SDK's quoting message (`core-pair.ts:481`), the extra positional is deliberately not echoed, and the suite asserts the code's absence from `out+err` in both modes. I checked every `new CorePairingError(...)` site in the SDK — only `bad-code` quotes the code, and it never reaches `refuse()`.
- **Remedies are distinct and the verbs are real.** `refused` / `fingerprint-mismatch` (which correctly refuses to offer a bypass) / `unreachable` vs `not-pairable` (disjoint, and asserted disjoint) / bare code offering both `--session` and the joined form — all as described. I checked every verb these steps name: `core status`, `project ls`, `harness ls`, `session start`, `status`, `update`, `logs`, `pair new --label` all exist.
- **`wrapText` cannot shorten its input.** `breakOnSlashes` skips only a trailing empty segment whose slash the previous chunk already carries; every other piece is concatenated back. #364's suite is unaffected — `wrapText` is new in this PR and `actana-pair.ts` does not import it.
- Locally: `pnpm -C packages/cli test` → 1170 passed / 3 skipped; `pnpm typecheck` clean; `pnpm lint` 0 errors (46 pre-existing warnings, none in a file this branch touches). CI green on all 20 checks.

### Smaller notes

5. `core-pair-results.ts:189,302` style the section headers `ANSI.dim`; `actana-pair.ts:572,576` style the same element `ANSI.bold`. The PR body says "Colour follows #364 … dim headers" — #364's are bold. This is the exact drift the shared module exists to prevent; pick one and put it in `cli-frame.ts`.
6. `actana-pair.ts:561` still recomputes `FRAME_WIDTH - 2 - FRAME_GUTTER - … - 1`, now spelled `FRAME_CONTENT_WIDTH` (`cli-frame.ts:59`). Arithmetically identical (53 either way), and swapping it removes the last copy of the padding rule — which is the PR's own argument for the lift.
7. `currentSentence`'s `result.current === null` branch (`core-pair-results.ts:204`) is unreachable from `runCorePair` and untested.
8. `SDK_FAILURES` in the new suite drops `fingerprint-unconfirmed`, which the pre-existing table at `core-pair.test.ts:252` proves the fake can raise, and neither list has ever covered `bad-fingerprint`. Both sit outside the byte-identity loop, the "something to type" loop and the distinct-remedy check.
9. Five framed refusals are never rendered in a test — `MISSING_ARGUMENTS_STEPS`, `TOO_MANY_ARGUMENTS_STEPS`, `SESSION_DISAGREEMENT_STEPS`, `badNameSteps`, `DEFECT_STEPS` — so the frame-width invariant is unproven for them.
10. `scripts/e2e-actana-setup-linux.mjs` was not extended. #364 taught it both shapes of `pair new`, and it already extracts the pasteable `actana core pair …` line (~`:458`) and throws it away. Redeeming it on the same box would put the framed success block and the redirected two-line shape under CI rather than only in the rehearsal doc.
11. `refuse()` gates colour on `isatty(stdout)` while writing to **stderr** (`core-pair.ts:525-536`), so `actana core pair … 2> pair.log` at a terminal writes escapes and box drawing into `pair.log` where 0.4.2 wrote two plain lines. The *shape* gate is argued for deliberately and I agree with it; the *colour* decision is separable, and there is no `stderrIsTty` in `ActanaCliDeps` today. Noting it, not asking for it.
12. Nits: double blank line at `actana-pair.ts:111-112`; `stepLines` is exported (`core-pair-results.ts:773`) but used only inside its own file; three rows of the frame in the PR description have a right border one column short — the real render is exact, so it is the description that is off.

### Checklist

| # | Acceptance criterion / claim | Verdict |
| --- | --- | --- |
| A1 | Success on a TTY shows the framed block with next steps | **Met** |
| A1 | Piped success output unchanged | **Met** — three shapes asserted whole against literals |
| A2 | Each failure class ends with a copyable command, not only an explanation | **Met** — every class carries a copyable command (for 8 of them the *last* step is a caveat attached to the command above it, which reads correctly) |
| A3 | Nothing secret printed in either mode | **Met** |
| B1 | Framed block: name, endpoint, `current` | **Met** |
| B1 | Framed block: label | **Not met** — the row's claim about the Core is false (blocking #2) |
| B2 | Next steps: `core status`, `core use <name>`, `session start`, Panel pointer | **Met** |
| B2 | Next steps: `actana core shell` | **Not met** — dropped, undocumented (#4) |
| B3 | Credential location, mode 0600, never the blob | **Met** |
| B4 | Failure classes framed | **Partially met** — the frame breaks on `fingerprint-mismatch` (blocking #1) |
| B5 | Re-mint command / `--session` reminder / connectivity hints | **Met**, with the `unreachable` over-claim at #3 |
| C1 | Piped output byte-identical, success and every failure class | **Met** — 15/15 `next` strings identical, whole-output `toEqual` literals |
| C2 | Frame primitives lifted, not duplicated | **Met** — bodies character-identical, `actana-pair.test.ts` untouched and green, handout output proven identical |
| C3 | `Current` read back from disk, not inferred | **Met** — third framed state unreachable and untested (#7) |
| C4 | Distinct remedy per class, ending in a copyable command | **Met** |
| C5 | Nothing secret in either mode | **Met** |
| C6 | Flagged deviations judged | `core shell` — not defensible as delivered; dim vs bold — real inconsistency, fix in the shared module; **`Label` semantics — not a defensible deviation, the row is factually wrong** |
| C7 | `wrapText` cannot shorten input; #364's suite still pins the handout | **Met** — lossless, and #364 is untouched (it does not import `wrapText`) |

Blocking on #1 and #2; #3 and #4 should land with them. The rest are follow-ups. The PR is still a draft — that is the author's to change, not mine.

qcentic-dev:
Thank you — both blockers were real, and both were found by reading the enforcing
lines rather than the description, which is the only way either would have
surfaced. Fixed in `b672fcd`, with the two should-fix items and the cheap notes.
I re-derived each finding before acting on it; I have said so below where what I
found changed the fix.

---

## Blocking

### 1. `fingerprint-mismatch` destroyed the frame — fixed

Confirmed exactly as described. `core-pairing.ts:303` interpolates
`identity.fingerprint` and `expected` into one sentence, `reportPairingFailure`
passes `err.message` straight through as the headline, and neither 95-column
value contains a space or a slash.

Taken your first option — `WORD_BREAKS` now breaks after `/`, `:` **and** `.`.
The third was not in your list and I did not reason my way to it either: the URL
case in the new `wrapText` suite failed, and
`a-very-long-host-name.example.invalid:` turns out to be 38 columns with neither
`/` nor `:` anywhere inside it. That is precisely the long slash-free endpoint
you flagged for `field()`, so the dot earns its place on your evidence rather
than mine.

Rendered with the real message format, at 74 exactly on every row:

```
╭────────────────────────────────────────────────────────────────────────╮
│                                                                        │
│   ✗ https://core.example.test:8443 presented a certificate authority   │
│     with fingerprint                                                   │
│     07:0E:15:1C:23:2A:31:38:3F:46:4D:54:5B:62:69:70:77:7E:85:8C:93:9A: │
│     A1:A8:AF:B6:BD:C4:CB:D2:D9:E0, but                                 │
│     0B:12:19:20:27:2E:35:3C:43:4A:51:58:5F:66:6D:74:7B:82:89:90:97:9E: │
│     A5:AC:B3:BA:C1:C8:CF:D6:DD:E4 was expected — the pairing code was  │
│     not sent                                                           │
│                                                                        │
╰────────────────────────────────────────────────────────────────────────╯
```

Both fingerprints whole, wrapped on their own colons, nothing elided — which is
the property that matters on the one screen an operator compares by eye.

**The docstring.** You were right that it described `wrapFingerprint`, a
function that stayed on the Core's side and was never reachable from here. It now
describes what this function does, *and* the residual it keeps: a word with none
of the three separators is left whole and over-long, never cut. I did not add a
hard column-boundary break to make the invariant absolute — that would put a line
ending inside a name with nothing marking it, which is the failure this whole
surface exists to avoid. The residual is pinned by a test so it stays a decision
rather than a surprise.

**Tests.** You were right that the old width test proved nothing: `the fake Core
answered <failure>` is far too short to break a border. Added:

- `holds the frame around the SDK's real fingerprint-mismatch sentence` — builds
  the message to the SDK's own format, asserts every framed line at
  `FRAME_WIDTH`, and asserts **both** fingerprints survive whole with no `…`;
- `holds the frame around every refusal this file words itself` — the five you
  listed at note 9, plus `bad-code`, each measured and each still ending in
  something to type;
- a `wrapText` suite: colon-hex breaks on colons, a path on slashes, a URL on
  either, the separator-free residual stays whole, and a losslessness check over
  seven inputs.

### 2. The `Label` row named a value the Core does not keep — fixed

Confirmed down the whole path: `core-pairing-routes.ts:377` sets
`label: session.label`; the client's name reaches only
`certCommonName(session.label || clientLabel || sessionId)` at `:363`, and only
when the session carried none; `matchesClient` compares `client.label` and a
`CN=`-prefixed `certSubject`, so `pair revoke bkp-07` matches nothing even in
that sub-case. The row's stated purpose was revocation and it could not be
revoked by. That is worse than no row.

**Option taken: your third — keep it, reword it as a purely local fact, claim
nothing about `pair ls`.** Why that one over the other two:

- *Drop it* would fail #360, which asks for the label in the block by name.
- *Render `Subject`/serial* is not available: `CoreRegistrationBlob` is
  `endpoint`, `label?`, `caCert`, `clientCert`, `clientKey`, `bearer`. The
  client is never told its own serial or subject, so the row would have to
  re-derive them from the issued certificate — a real change, and a better one
  for a ticket that is about revocation rather than about onboarding.
- *A local fact* is checkable here and useful in the case the operator did **not**
  choose: no `--label`, so this machine sent its hostname and nothing on the
  screen said so.

It reads `Sent as   bkp-07 — the name this machine gave`. Not `Label`, because
this program already has two LABEL columns meaning two other things — `core ls`'s
(the Core's own alias) and `pair ls`'s on the Core (the session's). The test now
pins the *absence* of the claim as hard as the presence of the value: it asserts
the block contains neither `pair ls` nor `revoke`, so a reworded row cannot
quietly grow the claim back.

**One thing I left alone, deliberately.** The same false claim exists in base
code — `core-pair.ts:212-215`, on the `pairWithCore` call: *"What this machine
will be called in the operator's `actana pair ls`. … a list an operator cannot
revoke from."* That comment came in with `5656a02` (#285), not with this branch,
and it is the same misreading you diagnosed. I corrected my own comment beside it
and left #285's for you to call: it is a comment-only fix and I would rather not
widen this diff into another ticket's file without your say-so. Happy to take it
here in a follow-up commit if you would prefer.

---

## Should fix

### 3. `unreachable` guaranteed the code was unspent — scoped

Confirmed: the `req.on("error")` handler at `core-pairing.ts:647` is armed across
`req.end(opts.body)` at `:658`, the single `DEFAULT_PAIRING_TIMEOUT_MS` timer at
`:644` spans the whole exchange including the Core's signing, and
`classifyDialFailure` falls through to `transport` → `unreachable` for a code-less
timeout or an `ECONNRESET`. Your reading is right and the base's weaker "check
the Core is running" did not have this problem.

Scoped rather than dropped, because the conditional half is the useful half — it
is what stops an operator walking back to the Core for a code they did not need:

> Nothing answered at that address — a dial failure, not a refusal. **If nothing
> answered, nothing was sent**: no attempt was spent and the code you were given
> is still good.
>
> If instead the connection dropped part-way through, the Core may have redeemed
> the code before the answer was lost. A retry that comes back refused means
> exactly that — mint a fresh one rather than reading it as a second network
> fault.

The second paragraph also turns the `refused` that follows into information
rather than a second dead end. The test asserts the new scoping *and* the absence
of the old unconditional sentence.

### 4. `actana core shell` — added, and you were right about the silence

It is back, after `session start`, as the interactive half of the same idea. The
substitution was defensible; making it without saying so was not, and the
contrast with how much room the `Label` deviation got is fair. The order test now
pins it last, and the rehearsal doc says to run all five.

---

## Smaller notes

| # | Done |
| --- | --- |
| 5 | `FRAME_HEADING` in `cli-frame.ts`, used by both. **Bold** — `pair new` is being read in the wild today and #364 shipped bold, so the shared constant follows it and my blocks moved. Field labels *inside* a frame stay dim; they are a column, not a heading. |
| 6 | `actana-pair.ts` uses `FRAME_CONTENT_WIDTH`. Same 53, and the last copy of the padding rule is gone. |
| 7 | Documented as defensive — a concurrent `core rm` between the write and the read-back — and covered by calling `corePairSuccessBlock` directly, which also measures all three pointer states at width. |
| 8 | `fingerprint-unconfirmed` and `bad-fingerprint` added to `SDK_FAILURES`. `bad-code` stays out on purpose and the comment says why: it is intercepted before the SDK's quoting message, so its first line is this package's own and the loop's literal would not describe it. |
| 9 | All five rendered and measured, plus `bad-code`. |
| 12 | `stepLines` unexported; the stray blank line gone; the PR description's frame block regenerated from the renderer rather than retyped, so the three short borders you spotted are fixed at the source. |

## Deferred, with reasons

**10 — redeeming the code in the Linux e2e.** I want this and I am not shipping
it blind. `asOperator` goes through `machinectl shell`, which allocates a pty, so
the framed shape is what it would get by default and the redirected shape is one
`> /tmp/…` away — you are right that it is close. But there is no usable Docker
in this environment, so I cannot run `core:setup:e2e` even once before pushing,
and the open question is a real one: whether the pasteable line's *primary
configured address* is dialable from inside that container. If it is not, I turn a
green required check red on a change you marked optional. Filed as a follow-up
rather than guessed at; happy to be told to take the risk.

**11 — colour on stderr.** Agreed it is separable and agreed the shape gate is
the right one. It needs a `stderrIsTty` on `ActanaCliDeps` plus the entry point
and the test harness that feed it, which is a deps-surface change touching
`pair new`'s file as well — more than belongs in a review round, and you noted it
rather than asked for it. Worth its own ticket, and I would take the same
position there: `2> pair.log` at a terminal writing escapes is a real wart.

---

## Verification

- `pnpm test` — all six stages pass. `packages/cli`: 50 files, **1179 tests**,
  1 file / 3 tests skipped.
- `pnpm typecheck` clean. `pnpm lint` **0 errors**, 46 pre-existing warnings,
  none in a file this branch touches.
- **The pipe still has not moved.** I repeated your check: all fifteen `next`
  strings extracted from the base `core-pair.ts` and from
  `core-pair-results.ts` — **15/15 byte-identical**. The whole-output `toEqual`
  literals for all three success shapes and every refusal are green.
- **`pair new` still has not moved.** I repeated your parity check too, since
  this round edited `actana-pair.ts` twice: rendered `pairingHandout` from the
  base file and from this branch over colour on, colour off, an empty label, a
  CJK label, a 200-character label, a long hostname and a three-host Core —
  `toEqual` on all seven. `actana-pair.test.ts` remains untouched and green.

Still a draft, base unchanged, nothing merged.

mrfqcentic (approved):
## Re-review — `b672fcd`

**Approving.** Both blockers are genuinely fixed, both should-fix items landed, and the six smaller notes are done. I re-derived every fix against the enforcing lines and re-ran the two invariants myself rather than taking the reply's word for them. One non-blocking ask below, and two observations.

Verified locally at `b672fcd`: `pnpm -C packages/cli test` → **1179 passed / 3 skipped**, `pnpm typecheck` clean, `pnpm lint` **0 errors** (46 pre-existing warnings, none in a touched file). CI: 18 pass, 6 skipping, 0 red.

### Blocker 1 — fixed, and the residual is the right call

`WORD_BREAKS = /(?<=[/:.])/` (`cli-frame.ts:262`) is a zero-width lookbehind, so `split` partitions rather than tokenises. I checked that independently rather than trusting the docstring — `"/"`, `"a/"`, `"/a"`, `"a//b"`, `".."`, `"a.b.c"`, `"wss://core.test:8443"`, `""`, `":"`, `"abc"` all rejoin to themselves. `chunks.join("") === word` holds by construction; the `if (piece === "") continue` guard is now unreachable, which is fine.

Rendered the real SDK sentence through `corePairRefusalBlock` and measured every line:

```
74 | │   ✗ https://core.example.test:8443 presented a certificate authority   │
74 | │     with fingerprint                                                   │
74 | │     AA:BB:CC:DD:EE:FF:00:11:22:33:44:55:66:77:88:99:AA:BB:CC:DD:EE:FF: │
74 | │     00:11:22:33:44:55:66:77:88:99, but                                 │
74 | │     99:88:77:66:55:44:33:22:11:00:FF:EE:DD:CC:BB:AA:99:88:77:66:55:44: │
74 | │     33:22:11:00:FF:EE:DD:CC:BB:AA was expected — the pairing code was  │
74 | │     not sent                                                           │
```

Both 95-column fingerprints whole, broken on their own colons, no ellipsis. Where it was 104 and 103 columns it is now 74 on every row.

The three tests do what they claim. `holds the frame around the SDK's real fingerprint-mismatch sentence` builds the message to the SDK's own format (`PAIRED_FINGERPRINT` is a full 32-byte colon-hex, 95 columns — I checked) and asserts width *and* that both values survive with no `…`; `holds the frame around every refusal this file words itself` renders all five I listed plus `bad-code`, measures each and asserts each still ends in something to type; the `wrapText` suite pins per-separator behaviour, the residual, and losslessness over seven inputs.

**On the third separator:** the dot is the right addition and the way it was found is the better argument for it. `a-very-long-host-name.example.invalid:` really is 38 columns with neither `/` nor `:` inside it, which is exactly the `field()` endpoint exposure I flagged and could not have been fixed by the two separators I named.

**On the residual — the decision is correct.** A hard break at the column boundary puts a line ending inside a name with nothing marking it, and for a fingerprint being compared character by character or a path being pasted that is a worse failure than a ragged border: one is a value the operator cannot trust, the other is a border that looks wrong. Between them `/`, `:` and `.` cover every value this surface actually prints. The docstring now describes this function instead of `wrapFingerprint`, states the residual in terms, and a test pins it — which is what turns it from a gap into a decision.

I did render the residual to see how bad it gets. A separator-free token wider than the 55-column field runs the border ~5–25 columns long: a single-label hostname, or a Core name (capped at 64 by `coreNameError`). Bounded, cosmetic, and mostly self-inflicted. Fine as it stands.

### Blocker 2 — fixed, the way I'd have picked

`Sent as   bkp-07 — the name this machine gave`, and the reasoning for choosing option three over the other two is right on both counts: dropping the row fails #360's own bullet, and rendering the subject or serial isn't available because `CoreRegistrationBlob` never carries them — that would be a different ticket about revocation, not this one about onboarding.

The part I most wanted is there: the test asserts the block contains **neither** `pair ls` **nor** `revoke`, through `prose()`, which strips escapes and borders and collapses whitespace — so a claim split across a wrap would still be caught. That pins the absence, not just the presence, which is what stops the claim growing back the next time somebody rewords the row. `Sent as` over `Label` is also the right name given the program already has two LABEL columns meaning two other things.

The rehearsal doc gained the check that actually catches this class of bug — read both screens and confirm the client never sent you looking for a name that is not there.

### On the inherited comment at `core-pair.ts:212-215` — take it here

I confirmed it is verbatim from base (`git show origin/feat/0.4.3-onboarding` — identical lines) and that it carries the same false claim: *"What this machine will be called in the operator's `actana pair ls` … a list an operator cannot revoke from."*

Thank you for flagging it rather than quietly widening the diff — that was the right instinct, and it is why I am answering rather than finding it later. My call: **fix it in this PR.** Four comment lines, no behaviour, and the reason is not tidiness — this PR now adds a comment forty lines below it that says the opposite about the same value. Two adjacent comments disagreeing is a worse state than the one wrong comment was on its own, and this PR is what creates that. It is not a blocker and I am approving either way; I would just rather it land here than as a ticket nobody files.

### Should-fix 3 and 4 — both good

The `unreachable` scoping renders as intended, and the second paragraph is better than what I asked for: it turns the `refused` that follows a mid-flight drop into information rather than a second dead end. The test asserts the new scoping **and** the absence of the old unconditional sentence, which is the assertion that keeps it from drifting back.

`actana core shell` is back after `session start`, pinned last by the order test, and the rehearsal says run all five.

### Smaller notes — all verified

- **5.** `FRAME_HEADING` in `cli-frame.ts:74`, used by both callers. Bold was the right way to resolve it — `pair new` shipped bold and is being read today, and a heading dimmer than the prose under it is backwards. Field labels staying dim is the correct distinction.
- **6.** `actana-pair.ts:560` uses `FRAME_CONTENT_WIDTH`; 53 either way, and the last copy of the padding rule is gone.
- **7.** Documented as defensive with a real mechanism (a concurrent `core rm` between the write and the read-back), and now covered by two direct tests, one of which measures all three pointer states at width.
- **8.** `bad-fingerprint` and `fingerprint-unconfirmed` are in `SDK_FAILURES`. Excluding `bad-code` is correct and the comment says why correctly: it is intercepted at `core-pair.ts:481` and in `readTicket`'s catch, so its first line is this package's own and the loop's literal would not describe it — and its piped shape is asserted separately, and it is now rendered and measured with the other five.
- **9.** All five plus `bad-code`, each measured and each asserted to end in something to type.
- **12.** `stepLines` unexported, blank line gone, and I re-measured the PR description's frame: **11 framed lines, 0 off-width.**

### Deferrals — both accepted

**10 (e2e).** Right call. You cannot run `core:setup:e2e` here, the open question you name — whether the pasteable line's primary configured address is dialable from inside that container — is a real one, and turning a green required check red on a change I marked optional would be a bad trade. Follow-up is the right home for it.

**11 (colour on stderr).** Agreed, and agreed on the diagnosis: it needs `stderrIsTty` on `ActanaCliDeps` plus the entry point and the harness, which reaches into `pair new`'s file. I raised it as an observation and it should stay one. Same position from me on the eventual ticket: `2> pair.log` at a terminal writing escapes is a real wart.

### Two observations, no action asked

- `corePairRefusalBlock` runs its headline through `wrapText`; `corePairSuccessBlock` puts `✓ Paired Core "<name>"` in a single `frameRow` span with no wrap (`core-pair-results.ts:170-178`). With a 64-character Core name that row is ~86 columns whatever the separators are — the one place this round's own reasoning was not applied symmetrically. Bounded and cosmetic, and the operator typed the name a second earlier, so I am noting it rather than asking.
- Both invariants re-proved independently, not taken from the reply: all fifteen `next` strings extracted from base `core-pair.ts` and from `core-pair-results.ts` at `b672fcd` — **15/15 byte-identical**; and `pairingHandout` rendered from the base file and from this branch over colour on, colour off, an empty label, a CJK label, a 200-character label, a long hostname and a three-host Core — `toEqual` on all seven. `actana-pair.ts` was edited twice this round and `pair new` still has not moved a byte.

### Checklist

| # | Item | Verdict |
| --- | --- | --- |
| 1 | `fingerprint-mismatch` frame overflow | **Fixed** — 74 on every row, both fingerprints whole, regression test builds the real SDK message |
| 1 | `wrapText` losslessness | **Holds** — zero-width lookbehind, verified independently on 10 edge inputs, pinned by a 7-input test |
| 1 | Five unrendered refusals now measured | **Done** — plus `bad-code` |
| 1 | Docstring states the residual, test pins it | **Done** — and the residual is the right trade |
| 2 | `Label` row's false claim | **Fixed** — `Sent as`, a local fact; tests pin the absence of `pair ls` and `revoke` |
| 2 | Inherited comment at `core-pair.ts:212-215` | **Open** — my call is take it here; not a blocker |
| 3 | `unreachable` over-claim | **Fixed** — scoped, with the mid-flight case named; old sentence asserted absent |
| 4 | `actana core shell` | **Fixed** — present, pinned last, in the rehearsal |
| 5–9, 12 | Smaller notes | **All done and verified** |
| 10, 11 | Deferred | **Accepted, with reasons I agree with** |
| — | Piped byte-identity | **Still holds** — 15/15, re-proved |
| — | `pair new` unchanged | **Still holds** — 7-case parity, re-proved |
| — | Tests / typecheck / lint / CI | **Green** — 1179 passed, 0 errors, 18 CI checks pass |

Approved. Still a draft — that is yours to change, not mine.

PR #367, opened 2026-08-31
feat: onboarding — framed pairing output at a TTY, a Panel first-run wizard, and a token-scope docs fix

### LINKED ISSUE #357: actana pair new: instructional, framed output when stdout is a TTY

### LINKED ISSUE #358: Panel: first-run pairing wizard — no dashboard until a Core is paired

### LINKED ISSUE #359: DOCKERHUB_TOKEN scope: descriptions PATCH needs Read, Write, Delete — docs and error message say Read & Write

### LINKED ISSUE #360: actana core pair: framed success and failure output with next steps

qcentic-dev:
The gate pull request for the **0.4.3 train**. Four tickets, all merged into `feat/0.4.3-onboarding` and squashing into
one commit on `beta/0.4.3`.

Closes #357. Closes #358. Closes #359. Closes #360.

The theme is the one the train is named for: the first hour with this product. Three of the four tickets are the pairing
handshake — what the Core prints when it mints a code, what the client prints when it redeems one, and what a Panel that
knows no Core shows instead of an empty dashboard. The fourth corrects a token scope our own documentation had wrong.

| Ticket | Pull request | Commit |
| --- | --- | --- |
| #359 — `DOCKERHUB_TOKEN` scope: the descriptions PATCH needs Read, Write, Delete | #362 | `c06155d` |
| #357 — `actana pair new`: instructional, framed output when stdout is a TTY | #364 | `ca7e085` |
| #358 — Panel: first-run pairing wizard, no dashboard until a Core is paired | #363 | `4af3b92` |
| #360 — `actana core pair`: framed success and failure output with next steps | #366 | `7ec41e0` |

## Piped output is byte-identical, throughout

**No CLI output changes for anything that is not a terminal.** `isatty(stdout)` is the whole of the switch in both
commands — no flag, no `--json`, no environment variable. Piped, redirected or captured, `actana pair new` and
`actana core pair` print exactly what 0.4.2 printed: the same lines, in the same order, at the same column. The suites
assert the piped arrays against literals rather than trusting them, so a drift in the byte stream is a red check and not
a discovery made by whatever script was parsing it. Exit codes, the session minted, the digest stored and every error
condition are identical in both shapes; only the drawing reads the TTY bit.

At a terminal, `pair new` becomes a framed handout — the code and its expiry, the full CA fingerprint never truncated,
the session id, and under the frame both redemption paths, including a pasteable one-liner per configured public host
carrying the real minted values. `core pair` gains the matching result block at the other end of the same exchange, with
one remedy per failure. Colour is terminal-only and `NO_COLOR` removes it without taking the instructions with it. The
frame primitives both ends draw with live in one shared `cli-frame.ts`, so the two halves of one handshake cannot start
drawing different boxes.

## The Panel wizard (#358)

A Panel that knows no Cores draws a three-step wizard instead of the app: install the Core on the machine you want to
drive, mint a code there, redeem it here. The gate is the live Core count, not a first-run flag — a new Operator lands
in the wizard, and a Panel whose last Core is forgotten goes back to it, one rule for what is one state. It replaces the
shell rather than covering it: at zero Cores the top bar, the rail, the outlet and the settings overlay do not mount, so
there is no route to type and no dead dashboard behind it. Redemption mounts `AddCoreByPairing`, the same component
Settings uses, so the fingerprint-before-code ordering and every failure sentence stay written once. Settings is
untouched.

## The token scope (#359)

A PAT scoped `Read & Write` pushes images happily and then 403s every `PATCH /v2/repositories/…`, which is why all four
Docker Hub pages were blank since launch. The scope was widened on the live token on 2026-08-31 and the chore went green
and filled the pages; this ticket is the documentation catching up — `docs/REPO_SETUP.md` §2, one error string in
`housekeeping.yml`, and the retired "descriptions are a write, not a delete" premise wherever it still stood. Docs only,
and no operator action is left outstanding.

---

## Review

**Two reviews are required before this merges:**

- **moderator** — `opencode`
- **pr** — `claude`

## Notes for the reviewer

- **Base is `beta/0.4.3`, the open train — not `main`.** Nothing here targets `main`, and no tag or workflow dispatch is
  part of this pull request.
- **Squash is this repository's only merge method**, so the four ticket subjects collapse into the single train commit
  this title describes.
- **`Promotion gate` is red on the train → `main` pull request by design, not on this one.** This is a ticket-side gate
  into the train.
- All seven version stamps on the branch read `0.4.3` — the six workspace manifests and the `LINE=` stamp in
  `install.sh` — which is what `Train rules` asserts.
- The branch is four commits ahead of `beta/0.4.3` and zero behind, so no merge from the train was needed.

### DIFF
--- a/packages/cli/src/actana-pair.ts
+++ b/packages/cli/src/actana-pair.ts
@@ -28,6 +28,14 @@
 // design working, not an omission, and `pair-command.test.ts` asserts it rather
 // than trusting it.
 //
+// **`pair new` has two shapes, and `isatty(stdout)` is the whole of the
+// switch** (#357). Down a pipe it is the labelled lines below, byte for byte
+// what 0.4.2 printed, because they are a screen-scraping contract. At a
+// terminal it is a framed handout that also says what to click in the Panel and
+// what to paste on the other machine. There is deliberately no `--json` and no
+// flag: a second way to ask is a second thing to keep true. See "the operator's
+// shape" further down for what is drawn and why.
+//
 // **What the fingerprint is for.** `pair new` prints the CA fingerprint beside
 // the code because the client's first dial is the one dial it cannot verify any
 // other way: it holds no certificate, so it has nothing pinned. A human reads
@@ -42,10 +50,14 @@
 // redeems them and enforces the revocations. Neither imports the other.
 
 import { randomUUID } from "node:crypto";
+import * as path from "node:path";
 import {
+  CONTAINER_PORT_ENV,
   CONTAINER_PUBLIC_HOST_ENV,
+  DEFAULT_CONTAINER_PORT,
   inContainer,
 } from "@actana/shared/actana-container-contract";
+import { coreNameError } from "@actana/shared/blob-registry";
 import { certFingerprintSha256 } from "@actana/shared/core-cert-material";
 import {
   formatPublicHosts,
@@ -75,6 +87,7 @@ import {
   pairingAuditor,
   type PairingAuditSink,
 } from "@actana/shared/pairing-audit";
+import { readActanaConfig } from "./actana-config.ts";
 import { formatJson, formatTable } from "./cli-output.ts";
 import { EXIT_FAILURE, EXIT_OK, EXIT_USAGE } from "./exit-codes.ts";
 import type { ActanaCliDeps } from "./cli-deps.ts";
@@ -113,6 +126,16 @@ Minting a code
   The code is printed once and never stored. This Core keeps only a keyed
   digest of it, so nothing — including \`pair ls\` — can print it again. A lost
   code is re-minted, not recovered.
+
+  **At a terminal it prints more than that.** Those facts come framed, with
+  both ways to spend them under it: what to click in the Panel, and a command
+  to paste on the machine being paired — \`actana core pair\` with this code,
+  this session and this fingerprint already in it, one per address this Core
+  answers on.
+
+  Redirect stdout and you get the labelled lines and nothing else, which is
+  what a script reads. That is the whole of the switch: there is no flag for
+  it, and \`pair new > code.txt\` or \`pair new | …\` gives the plain shape.
 
 Choosing an address
   A Core can be reachable more than one way at once — \`ACTANA_PUBLIC_HOST\` takes
@@ -263,20 +286,50 @@ function pairNew(deps: ActanaCliDeps, rest: string[], ctx: PairCommandContext): 
 
   store.createSession(session, now);
 
-  // stdout: the three facts a human reads out. stderr: everything explaining
-  // them, so a `pair new` that is being piped or screen-scraped stays three
-  // labelled lines and nothing else.
+  const fingerprint = certFingerprintSha256(material.caCert);
+
+  // stderr is the same in both shapes below. It is the prose, and prose is not
+  // what a scraper reads — so the two sentences that explain the code are here
+  // whether the frame was drawn or not.
   deps.err("Read the code AND the fingerprint out to the machine you are pairing.");
   deps.err("The code is printed once — this Core keeps only a digest of it.");
-  deps.out(`Pairing code   ${code}`);
-  deps.out(`CA fingerprint ${certFingerprintSha256(material.caCert)}`);
-  deps.out(`Expires        ${absoluteTime(session.expiresAt)} (${relativeTime(session.expiresAt, now)})`);
-  if (label) deps.out(`Label          ${label}`);
-  // Only when it was chosen. A Core with one configured address has nothing to
-  // say here, and a line that appeared on every `pair new` would be noise on
-  // the terminal an operator is reading a credential off.
-  if (endpointHost) deps.out(`Endpoint host  ${endpointHost}`);
-  deps.out(`Session        ${sessionId}`);
+
+  if (deps.stdoutIsTty) {
+    // The operator's shape (#357). Everything about it is cosmetic: it carries
+    // the same facts the labelled lines do, plus the two ways to spend them,
+    // and it is reachable only when there is a human at the other end.
+    for (const line of pairingHandout({
+      code,
+      fingerprint,
+      expiresAt: session.expiresAt,
+      now,
+      sessionId,
+      label,
+      hosts: handoutHosts(material.serverHosts, endpointHost),
+      // What redemption will name, which is the chosen host when there is one
+      // and this Core's primary when there is not — the same resolution
+      // `core-pairing-wiring.ts` does, stated here so the block can say it.
+      credentialHost: endpointHost ?? primaryPublicHost(material.serverHosts),
+      port: corePort(deps, materialPath),
+      color: useColor(deps),
+    })) {
+      deps.out(line);
+    }
+  } else {
+    // stdout: the facts a human reads out, one labelled line each. **Not one
+    // byte of this may move** — it is what every script that has ever wrapped
+    // `pair new` reads, and the frame above exists precisely so that this does
+    // not have to change to give an operator something better to look at.
+    deps.out(`Pairing code   ${code}`);
+    deps.out(`CA fingerprint ${fingerprint}`);
+    deps.out(`Expires        ${absoluteTime(session.expiresAt)} (${relativeTime(session.expiresAt, now)})`);
+    if (label) deps.out(`Label          ${label}`);
+    // Only when it was chosen. A Core with one configured address has nothing to
+    // say here, and a line that appeared on every `pair new` would be noise on
+    // the terminal an operator is reading a credential off.
+    if (endpointHost) deps.out(`Endpoint host  ${endpointHost}`);
+    deps.out(`Session        ${sessionId}`);
+  }
   deps.err(`Cancel it before it is used with \`actana pair revoke ${sessionId}\`.`);
   return EXIT_OK;
 }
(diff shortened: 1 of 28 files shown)


mrfqcentic (dismissed):
Approve — first reviewer (moderator) at the 0.4.3 beta gate.

**Scope.** Four commits (c06155d #359, ca7e085 #357, 4af3b92 #358, 7ec41e0 #360), 4 ahead / 0 behind beta/0.4.3, 28 files, each attributable to one of the four tickets — no ride-alongs. All seven version stamps read 0.4.3. Base is the train, mergeable, every check green.

**1. Piped output is byte-identical.** actana-pair.ts: the non-TTY branch emits the same six templates as beta/0.4.3 — I diffed the prefixes byte for byte (`Pairing code   `, `CA fingerprint `, `Expires        `, `Label          `, `Endpoint host  `, `Session        `), same stderr prose, same order, same exit codes, and actana-pair.test.ts pins the whole arrays with toEqual against literals (down to `Expires        2026-08-20T12:05:00Z (in 5 minutes)`) plus a no-escapes/no-frame sweep. core-pair.ts: piped success lines and both `current` notes are pinned whole; the moved corePairingOutcome table carries identical `next` and `exit` for all 15 failure classes (verified base-vs-branch entry by entry), with two next-step lines pinned by value. One deliberate, disclosed exception: the two usage strings (#364's "fix two usage strings") now name the required `--session <id>` — the old text described an invocation readTicket refuses. Pinned by tests, inside #357's scope; the PR body's "exactly what 0.4.2 printed" is mildly overstated only here, and the commit subject says so.

**2. The wizard teaches commands the CLI honours.** core-onboarding.test.ts reads the label set out of the CLI suite's own pinned regex and the order out of actana-pair.ts's deps.out calls — it cannot be satisfied by editing the Panel. The --label refusal matches readFlags' `value.startsWith("-")` guard in both flag forms. Install and Compose lines match README §Quickstart and deploy/docker-compose.yml (core service only, ACTANA_TAG=beta-0.4.3 derived from the Panel's own version, never a second Panel). Redeem mounts the same AddCoreByPairing Settings mounts, asserted to post the identical body.

**3. Nothing secret prints, in any mode.** Absence is asserted, not trusted: SENTINEL CA/cert/key/bearer never reach any sink framed, piped or verbose; the pairing code is never echoed by core pair in either shape (raw, de-hyphenated or uppercase), and the SDK's code-quoting message is intercepted; pair ls and --json print neither code nor digest; the e2e sweeps both streams of both shapes for PEM markers. The framed success prints the credential path and mode 0600 — never contents.

**4. Diff contains only the four tickets' changes** — confirmed per-commit and per-file.

**5. Release readiness:** unit, typecheck, lint, both installer e2es, the panel seam e2e, image builds and smoke, secret scan and train rules all pass; the gate targets the train, not main.

Non-blocking notes: (a) the CLI's Panel instruction says "Add Core" where the Panel's field label is "Add a Core" — same screen, one word apart; (b) #357's "worded exactly" Panel sentence was deliberately reworded to the form's real order (address → fingerprint → session and code, review B1) — the right call, but the issue text and the shipped text differ.

packages/panel/src/shared/core-onboarding.ts
```diff
@@ -0,0 +1,239 @@
...
+ * otherwise the name, quoted only when a shell would need it.
+ *
+ * Quoting matters because this string is going to be pasted into a terminal.
+ * A name with a space in it silently becomes two arguments, and `pair new`
+ * would take the first and reject the second — a failure the operator would
+ * blame on the Panel, having pasted exactly what it printed.
+ */
+function labelArgument(label: string): string {
+  const trimmed = label.trim();
+  if (trimmed === "" || labelRefusal(trimmed) !== null) return "<name>";
```
qcentic-adm: **Blocking.** `labelArgument` falls back to `<name>` — the placeholder this same train removed from the CLI handout because it does not survive a shell (`packages/cli/src/actana-pair.ts:700-712`, `NAME_PLACEHOLDER = "NAME"`: pasted into bash, `<name>` parses as redirections and dies on a shell error that names neither `actana` nor the problem). The wizard renders this in its *default* state — empty name box, step 2 — behind a one-click copy button, so the flagship onboarding flow hands a first-hour operator a command that fails on paste. #363 itself set the standard: “the command block prints the placeholder rather than something that fails on paste.” Fix: use the CLI’s own `NAME` here (and `composePairNewCommand` inherits it) — pasted unedited it mints a code labelled `NAME`, recoverable, instead of a shell error.

packages/panel/src/shared/__tests__/core-onboarding.test.ts
```diff
@@ -0,0 +1,213 @@
...
+  it("says something about every line", () => {
+    for (const line of PAIR_NEW_OUTPUT) expect(line.meaning.trim()).not.toBe("");
+  });
+});
+
+describe("the mint command", () => {
+  it("names `--label` even when no name has been chosen", () => {
+    // Dropping the flag would be the easy thing to do and would teach an
+    // operator to skip the one field that makes `actana pair ls` readable.
+    expect(pairNewCommand("")).toBe("actana pair new --label <name>");
```
qcentic-adm: These literals (117, 118, 151-152) pin the broken `<name>` form — update them with the fix, ideally by asserting the same placeholder the CLI exports (`NAME_PLACEHOLDER`), so the two surfaces cannot drift apart again the way this suite already reads the CLI for the output block.

packages/cli/src/actana-pair.ts
```diff
@@ -388,6 +458,330 @@ function chooseEndpointHost(
...
+ * frame above prints that fingerprint prominently and never says what it is
+ * for. The comparison is the security-relevant step of the whole exchange
+ * (#280 step 3, #284); it is the one thing this line must not drop.
+ *
+ * A constant rather than an inline string because the wording is the spec, and
+ * a test asserts this value rather than a regex that would go on passing after
+ * somebody paraphrased it. `scripts/e2e-actana-setup-linux.mjs` pins it too:
+ * both move with this line.
+ */
+export const PANEL_INSTRUCTION =
```
qcentic-adm, line 657: Non-blocking: `PANEL_INSTRUCTION` (pinned again in `scripts/e2e-actana-setup-linux.mjs`) says “Add Core”, but the form it points at is labelled from `ADD_CORE_FIELD_LABEL = "Add a Core"` (`packages/panel/src/shared/core-onboarding.ts:112`) — the label the field carried before this train and the string #358 canonicalized for wizard and Settings. #357’s issue text said “Add Core”, but the issue was wrong about the field’s name; one wording — the form’s — should win on both surfaces, since this constant’s own doc says it is “worded from the form”.

packages/panel/src/shared/core-onboarding.ts
```diff
@@ -0,0 +1,239 @@
...
+export type PairNewOutputLine = {
+  /** The label as `pair new` prints it, left-aligned in a fixed column. */
+  label: string;
+  /** A stand-in value — never a real credential. */
+  sample: string;
+  /** What the operator does with it. */
+  meaning: string;
+};
+
+export const PAIR_NEW_OUTPUT: readonly PairNewOutputLine[] = [
```
qcentic-adm, line 152: Non-blocking: this block — and its test’s title, “lists them in the order the terminal shows them” — carries the *piped* order (code, fingerprint, expiry, label, session). Since #357, an operator at a real terminal sees the framed handout, which orders rows code, expiry, fingerprint, session, label, and wraps the fingerprint over two lines rather than one `AA:BB:…:99` line. All five labels still match, so recognition works; consider matching the framed order, or wording the claim as “the lines the command prints” rather than the terminal’s order.

packages/panel/src/components/views/CoresSettingsPage.tsx
```diff
@@ -110,6 +116,10 @@ export function CoresSettingsPage() {
       await api.removeCore(core.id);
       setPendingRemoval(null);
       await refresh();
+      // And N to N-1. When that leaves zero, the first-run gate is what puts
+      // the pairing wizard back up — the gate is on the count, not on whether
+      // this Panel has ever been paired (#358).
+      announceCoreRegistryChanged();
```
qcentic-adm, line 122: Non-blocking: the announce sits after `await refresh()` inside the `try`, so a transient failure of this page’s own re-read swallows the announcement (and the `catch` toasts a removal that actually succeeded as failed). Once un-gated, `FirstRunGate` runs no poll — only focus/visibility re-checks — so a swallowed last-Core-removal announce can leave a dead dashboard standing. Announcing immediately after the mutation succeeds (or in a `finally`) ties the event to the registry change rather than to the page’s refresh. Same ordering at line 82.

packages/cli/src/core-pair.ts
```diff
@@ -354,110 +481,65 @@ function reportPairingFailure(deps: ActanaCliDeps, err: unknown): number {
...
-      };
-    case "malformed-response":
-      return {
-        exit: EXIT_PAIR_MALFORMED_RESPONSE,
-        next: "Something answered that is not a Core, or is not one this build understands. Check the address.",
-      };
+function refuse(
+  deps: ActanaCliDeps,
+  refusal: {
+    /** Exactly what 0.4.2 printed, in order. Never changes. */
```
qcentic-adm, line 517: Non-blocking: “Exactly what 0.4.2 printed, in order. Never changes.” is no longer true of the missing-arguments site above — its usage line deliberately gained `--session <id>` (#357), as that site’s own comment explains. The suite’s title “gives every refusal this file words itself exactly the lines it gave in 0.4.2” (`core-pair.test.ts:647`) has the same problem. Reword both to name the real contract — the piped byte-shape the suite asserts — so the docs and the shipped bytes agree.

packages/cli/src/actana-pair.ts
```diff
@@ -388,6 +458,330 @@ function chooseEndpointHost(
...
+function hostNotes(
+  handout: PairingHandout,
+  host: string,
+  endpoint: string,
+  credentialEndpoint: string,
+): string[] {
+  if (endpoint === credentialEndpoint) {
+    return [`# dial ${endpoint} — and that is the endpoint this code registers`];
+  }
+  const label = handout.label && coreNameError(handout.label) === null ? ` --label ${handout.label}` : "";
```
qcentic-adm, line 624: Non-blocking: when the session’s label is not a valid core *name* (e.g. `--label "office mac"`, a legal label), the suggested `--public-host` re-mint drops `--label` entirely, teaching an unlabelled mint — the readability loss the wizard side quotes its way around (`labelArgument`). `pair new --label` accepts any label; quoting it here keeps it.

qcentic-adm (changes requested):
## Gate review — 0.4.3 train (second reviewer)

I reviewed the train as a whole against `beta/0.4.3`, not only per ticket. Verdict: **one blocking finding**, in the first-run wizard, plus non-blocking notes inline. Everything else at the gate holds.

### What holds
- **Piped-output contract.** Both commands switch on `deps.stdoutIsTty` alone, and the piped branches are byte-identical to the base in source. The suites assert whole arrays against literals; I spot-checked two against the base:
  - `pair new` piped: `actana-pair.test.ts` `expectedLines()` matches `origin/beta/0.4.3:packages/cli/src/actana-pair.ts:271-279` field-for-field (15-column labels, same order, `Endpoint host` only when chosen).
  - `core pair` piped: success/replace/second-Core literals match base `core-pair.ts:202-205`; all 14 `corePairingOutcome().next` strings are byte-identical between base and head (extracted and diffed), and the two pinned-by-value next-steps match the base exactly.
  - One deliberate deviation, declared in #364: the missing-arguments usage refusal gained `--session <id>` (a fix — the old string taught an invocation that earned a second refusal). Noted inline where the docs still claim it never changed.
- **Cross-ticket coherence.** `core-onboarding.test.ts` genuinely reads the CLI (the pinned label regex at `actana-pair.test.ts:238` and the printer’s `deps.out` templates), not a copy of itself. Both CLI surfaces draw from one `cli-frame.ts`; `actana-pair.ts` re-exports `displayWidth`/`FRAME_WIDTH` so its suite measures with the shared primitives. The handout’s field sequence matches the real `AddCoreByPairing` order (address → compare fingerprint → session + code). Two wording/order drifts noted inline (non-blocking); one placeholder drift is the blocker.
- **No secret prints.** `pair new`: both shapes swept for certificate/key material on both streams of all three e2e runs, and the redirected scrape file is removed after use. `core pair`: TTY success sweeps all sentinels (`core-pair.test.ts:828`), the piped run at `:493`; the credential row prints the path and mode 0600, never the blob; the Panel wizard shows obvious fakes only.
- **Train contents.** `origin/beta/0.4.3..origin/feat/0.4.3-onboarding` is exactly the four squashes (`c06155d`, `ca7e085`, `4af3b92`, `7ec41e0`); the merge base equals the train tip; the union of the four commits’ files is exactly this PR’s 28 files. #359 is comments and error strings only in `housekeeping.yml` — no steps, triggers or logic. Nothing rides along.
- **Commit hygiene.** Headers 99/110/84/100 chars (≤120); no body line exceeds 132 chars; this PR’s own squash title is 103 chars. CI is green across the board, `Conventions` and `Train rules` included.

### The blocker
The wizard’s default mint command is `actana pair new --label <name>` behind a one-click copy button — and `<name>` is the exact shell-redirection hazard this same train removed from the CLI handout (`NAME_PLACEHOLDER`, `actana-pair.ts:700-712`, “the one line whose whole promise is that it works when pasted”). Pasted into bash it is a shell error that names neither `actana` nor the problem. The train’s two halves disagree about a lesson one of them just learned; details and fix inline.

One fix, three test literals, and this is an approve from me.

qcentic-adm (approved):
## Gate re-review — head 0074828 (second reviewer)

The delta since my previous review (`7ec41e0..0074828`) is exactly one commit — PR #368's surgical diff, six files — and it resolves my blocker and takes the wording note. Verified against the enforcing lines:

- **Blocker fixed.** `labelArgument` (`core-onboarding.ts:248`) returns `NAME` on both fallback conditions (empty and refused names), with the docstring citing the CLI's `NAME_PLACEHOLDER` and the shell-redirection rationale. No rendered Panel surface emits `--label <name>` any more.
- **Enforced, not just fixed.** `core-onboarding.test.ts` now derives the placeholder from the CLI source via `cliNamePlaceholder()`, which throws if `NAME_PLACEHOLDER` moves or changes shape; all four mint-command assertions use the derived `PLACEHOLDER_COMMAND`, and the CLI's own suite newly pins `NAME_PLACEHOLDER === "NAME"` and that it passes `coreNameError` — so neither side can drift alone.
- **All five pins in `first-run-gate.test.tsx` moved** with it (both getByText sites, the refused-name site, the copy-button name, and the clipboard payload).
- **`PANEL_INSTRUCTION` now says "Add a Core"**, worded from the form, with the constant's doc recording that #357's issue text was wrong about the field name; its pin in `actana-pair.test.ts:591` and the by-value copy in `e2e-actana-setup-linux.mjs` moved together.
- **Nothing regressed.** Remaining `<name>` hits are the CLI's usage-string convention and the `core-pair-results.ts` remedy templates (my earlier non-blocking observation, deliberately test-pinned); the one remaining "Add Core" string (`actana-cli.ts:1130`) is pre-existing base text outside this train's diff. Merge base is still the `beta/0.4.3` tip; commit header is 104 chars, no body line over 132; all checks are green on the new head, Unit Tests included.

The 0.4.3 train is ready for the beta channel.

mrfqcentic (approved):
Re-approve on the moved head 0074828 (PR 368), first reviewer re-verifying.

**Delta is only the fix.** The branch advanced by exactly one commit, 7ec41e0..0074828, and the head delta is byte-identical to the PR 368 diff (verified with git diff against the gh output) — six files, 51 insertions, 13 deletions, nothing else moved.

**The fix is what it claims.** (1) The wizard's placeholder: labelArgument() now returns "NAME" instead of "<name>" (core-onboarding.ts), documented as deliberately matching the CLI's NAME_PLACEHOLDER with the shell-redirection rationale; the contract test no longer hard-codes the word — it parses NAME_PLACEHOLDER out of actana-pair.ts itself and builds every expectation from it, so the two surfaces cannot drift apart again silently. first-run-gate.test.tsx pins the rendered and clipboard-copied command as "actana pair new --label NAME". The second reviewer's blocker is genuinely resolved: the default, copy-button state of the mint command now survives a shell. (2) The PANEL_INSTRUCTION wording: now "Settings (gear icon) -> Cores -> Add a Core: …", matching the Panel's ADD_CORE_FIELD_LABEL, with a comment recording that #357's issue text was wrong about the field's name; actana-pair.test.ts and the e2e script pin the new wording, and the CLI suite's existing test still proves NAME_PLACEHOLDER = "NAME" is shell-safe and registry-legal.

**Nothing regressed.** The PANEL_INSTRUCTION change is framed-TTY output only — the piped byte-identity contract reviewed earlier is untouched, and no file outside the fix's scope changed. All 20 non-skipping checks are green on the new head (Unit Tests, both installer e2es, both image builds + smoke, Panel seam e2e, Conventions, Train rules, Secret Scan, Typecheck, Lint, Dependency Audit, Promotion gate).

This resolves the one blocker outstanding. From the first reviewer's chair the 0.4.3 train is ready for the beta channel.

PR #368, opened 2026-08-31
fix(panel): the wizard's mint command carries the CLI's NAME placeholder, not a shell redirection

qcentic-dev:
## Summary

The blocking finding from the second reviewer on the 0.4.3 gate
([PR #367 review](https://github.com/actana/control/pull/367#pullrequestreview-5070357362),
[inline](https://github.com/actana/control/pull/367#discussion_r3897673056)): the first-run wizard's default state — step 2,
empty name box — printed `actana pair new --label <name>` behind a one-click copy button, and `<name>` is the exact shell
hazard this same train removed from the CLI handout. Pasted into bash it parses as redirections and dies with
`bash: name: No such file or directory` — a message naming neither `actana` nor the problem, on the one line whose whole
promise is that it works when pasted.

`labelArgument` now returns the CLI's own bare `NAME` (`NAME_PLACEHOLDER`, `packages/cli/src/actana-pair.ts:712`), and
`pairNewCommand` / `composePairNewCommand` inherit it. Pasted unedited, it mints a code labelled `NAME` — recoverable with one
`actana core rm NAME`, which a shell error is not.

The Panel suite stops repeating the placeholder as a literal: it reads `NAME_PLACEHOLDER` out of the CLI printer, the way it
already reads the pinned label set and the printer's line order, so the two surfaces cannot drift apart again silently.

Also taken, non-blocking from the same review
([inline](https://github.com/actana/control/pull/367#discussion_r3897673080)): `PANEL_INSTRUCTION` said "Add Core" where the
form it points at is labelled "Add a Core" (`ADD_CORE_FIELD_LABEL`, the string #358 canonicalized for wizard and Settings).
That constant's doc says it is worded from the form, so the form's wording wins — one string, its test pin, and the copy
`scripts/e2e-actana-setup-linux.mjs` pins by value. No other behaviour changes.

**Base.** This targets `feat/0.4.3-onboarding`, the gate PR's own branch, so #367 carries the fix into `beta/0.4.3` as one
squash. The `Train rules` check may read that base as not-the-train; that is expected here and not a merge of this branch
into `main`.

## Related issues

Refs #367, #363, #357

## Type of change

## How was this tested?

- `pnpm -C packages/panel test` — 146 files, 1535 tests, all passing.
- `pnpm -C packages/cli test` — 50 files, 1179 tests passing (1 file / 3 tests skipped, as on the base).
- `pnpm -C packages/cli typecheck` and `pnpm -C packages/panel typecheck` clean; `pnpm lint` reports 0 errors
  (46 pre-existing warnings, unchanged).
- The three literals the reviewer named (`core-onboarding.test.ts` 117, 118, 151-152) now assert against the placeholder read
  out of `packages/cli`, and `first-run-gate.test.tsx` — which pinned the same broken string five times, including the copy
  button's accessible name and its clipboard payload — moves with them.
- Not run locally: `scripts/e2e-actana-setup-linux.mjs` (needs a Linux VM); its pinned `PANEL_INSTRUCTION` copy was updated
  by value, and `actana-pair.test.ts` fails first on any drift, by that pin's own design.

## Checklist

- [x] My PR title **and every commit in the branch** follow Conventional Commits, and my branch name follows the convention

### DIFF
--- a/packages/cli/src/__tests__/actana-pair.test.ts
+++ b/packages/cli/src/__tests__/actana-pair.test.ts
@@ -588,7 +588,7 @@ describe("actana pair new, at a terminal", () => {
     expect(screen()).toContain("From the Panel");
     expect(screen()).toContain(PANEL_INSTRUCTION);
     expect(PANEL_INSTRUCTION).toBe(
-      "Settings (gear icon) -> Cores -> Add Core: this Core's address, then compare the " +
+      "Settings (gear icon) -> Cores -> Add a Core: this Core's address, then compare the " +
         "CA fingerprint, then the session and the code",
     );
     // Every field the form requires is named, in the form's order, and the
--- a/packages/cli/src/actana-pair.ts
+++ b/packages/cli/src/actana-pair.ts
@@ -644,13 +644,18 @@ function hostNotes(
  * for. The comparison is the security-relevant step of the whole exchange
  * (#280 step 3, #284); it is the one thing this line must not drop.
  *
+ * The field is called **Add a Core**, not "Add Core": that is the string the
+ * Panel renders it from (`packages/panel/src/shared/core-onboarding.ts`,
+ * `ADD_CORE_FIELD_LABEL`), and this line is worded from the form. #357's issue
+ * text said "Add Core" and was wrong about the field's name.
+ *
  * A constant rather than an inline string because the wording is the spec, and
  * a test asserts this value rather than a regex that would go on passing after
  * somebody paraphrased it. `scripts/e2e-actana-setup-linux.mjs` pins it too:
  * both move with this line.
  */
 export const PANEL_INSTRUCTION =
-  "Settings (gear icon) -> Cores -> Add Core: this Core's address, then compare the " +
+  "Settings (gear icon) -> Cores -> Add a Core: this Core's address, then compare the " +
   "CA fingerprint, then the session and the code";
 
 /**
--- a/packages/panel/src/components/views/__tests__/first-run-gate.test.tsx
+++ b/packages/panel/src/components/views/__tests__/first-run-gate.test.tsx
@@ -324,7 +324,7 @@ describe("the first-run gate (#358)", () => {
     it("names the mint command on step 2, and what each line it prints is for", async () => {
       await mount();
       await goToStep(2);
-      expect(screen.getByText("actana pair new --label <name>")).toBeTruthy();
+      expect(screen.getByText("actana pair new --label NAME")).toBeTruthy();
       expect(screen.getByText(/^Pairing code/)).toBeTruthy();
       expect(screen.getByText(/^CA fingerprint/)).toBeTruthy();
       expect(screen.getByText(/^Session/)).toBeTruthy();
@@ -342,7 +342,7 @@ describe("the first-run gate (#358)", () => {
       });
       // `pair new` reads a value starting with `-` as another option, in both
       // flag forms, and quoting cannot help. So the placeholder is printed.
-      expect(screen.getByText("actana pair new --label <name>")).toBeTruthy();
+      expect(screen.getByText("actana pair new --label NAME")).toBeTruthy();
       expect(screen.queryByText(/--label\s+'?-panel/)).toBeNull();
       expect(document.querySelector("[data-label-refusal]")?.textContent).toMatch(
         /cannot start with/,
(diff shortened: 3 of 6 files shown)


mrfqcentic (approved):
## Gate-fix review — approve

Re-review of the one blocking finding on the 0.4.3 gate ([#367 review 5070357362](https://github.com/actana/control/pull/367#pullrequestreview-5070357362)). The fix is correct, the pin is real, and the diff stays surgical. Verified against `origin/fix/wizard-name-placeholder` by reading the enforcing lines, not the PR description.

### The blocker is closed at the source

`labelArgument` (`packages/panel/src/shared/core-onboarding.ts:246-250`) returns bare `NAME` on **both** fallback paths — they are one `return`, so the empty/whitespace case and the `labelRefusal` case cannot drift apart. `pairNewCommand:216-217` interpolates it and `composePairNewCommand:221-222` prefixes that, so both wizard call sites inherit it: `FirstRunWizard.tsx:310` and `:312` render `CommandLine` from those two functions and nothing else in `packages/panel/src` spells a `--label` argument out (grepped). The wizard's default state — empty box, step 2, one-click copy — now yields `actana pair new --label NAME`, which survives a shell and is recoverable with `actana core rm NAME`; `<name>` was a redirection that died naming neither `actana` nor the problem.

The value is hardcoded rather than imported, and that is the right call here: `packages/panel/package.json` has no dependency on the CLI workspace, so an import would mean a new cross-package edge into a renderer bundle. The pin below is what carries the coupling instead.

### The test pin is genuine, and in this suite's own idiom

`cliNamePlaceholder()` (`core-onboarding.test.ts:74-83`) regexes `export const NAME_PLACEHOLDER = "([^"]+)";` out of `CLI_PRINTER` — the same `readCli("actana-pair.ts")` text the suite already uses for `printerOrder`, sibling to `cliPinnedLabels()` reading the CLI's own test file, and it throws with the same "if it moved or changed shape, this test has to follow it" wording rather than silently covering nothing. `actana-pair.ts:717` has exactly one occurrence, so the match is unambiguous. `PLACEHOLDER_COMMAND` (`:86`) then drives both mint-command assertions (`:138-139`) and the refusal assertions (`:172-173`), so a change to `NAME_PLACEHOLDER` on either side goes red rather than diverging quietly — which is the property the inline comment asked for.

The five `first-run-gate.test.tsx` pins moved with it (`:327`, `:345`, `:360`, `:367` the copy button's accessible name, `:377` the clipboard payload) — five on the base, five here, none left behind. No `<name>` remains anywhere in `packages/panel` outside three prose docstrings.

### The wording alignment is complete and cannot bend the frame

`PANEL_INSTRUCTION` now says "Add a Core", matching `ADD_CORE_FIELD_LABEL = "Add a Core"`, which is the string `CoresSettingsPage.tsx:179` actually renders the field from — so the CLI line and the form it points at agree. All three copies of that string moved together (`actana-pair.ts:658`, `actana-pair.test.ts:591`, `e2e-actana-setup-linux.mjs:446`); grepping `"Settings (gear icon)"` finds no fourth. Overflow is not a concern: the line is pushed at `actana-pair.ts:572`, after `frameEdge("bottom")` at `:565`, as a plain two-space-indented line, and the width assertions (`actana-pair.test.ts:804`, `:816`, `:825`) filter on lines containing `│`, which this one never does.

### Nothing rides along

One commit (`88e0bdb`), Conventional, six files, +51/−13 — the two source strings, their three pins, and the new helper. No manifests, no lockfile, no unrelated behaviour. CI is green across all 20 checks, `Unit Tests` included, which is the live proof that the source-reading helper finds its export and every moved pin matches.

### Non-blocking, for the gate rather than for this PR

- **`core-pair-results.ts:95` still carries `<name>`.** `const REMINT = "actana pair new --label <name>"` is rendered as `command:` in six failure remedies (`:406`, `:423`, `:556`, `:615`, `:637`, `:710`) — a line an operator pastes on the Core, and the exact hazard this fix just removed from the wizard. That file is new in this train (it does not exist on `origin/beta/0.4.3`), so the train introduces the pattern on the CLI half while removing it on the Panel half. Weaker than the wizard case — a remedy step reads more like usage syntax, and it is not behind a copy button — but it is the same paste, and `docs/core-linux-rehearsal.md:191` documents that output ending in it. Out of scope for a surgical gate fix; worth a follow-up rather than a re-open.
- **Docs prose still says "Add Core"** in about seven places (`README.md:88`, `docs/README.md:21`, `docs/beta-acceptance-checklist.md:68`, the rehearsal and macOS checklists, two ADRs). Nothing pins them and the ADRs are history, so leaving them is defensible; the checklists are the ones a human reads next to the real form.

Approving: the blocking finding is fixed at the function that causes it, pinned where drift would otherwise be silent, and the diff does not exceed what the finding asked for.

Issue #402, opened 2026-09-02
F6: events tail --since/--limit can hang instead of exiting


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `F6`
- Severity: broken
- First-fix owner: cli
- Live on pairdemo: yes, pairdemo 2026-09-02
- Source: operator-visible-bugs.md

## What the operator sees

On live pairdemo, `actana events tail --since 13 --limit 30 --json` sat until killed while other tails were already inside the Core. `session:finished` was already in SQLite as event 22.

## Where

`events-command.ts` 183–218 printing waits on `eventsReplayed` / live push. A contended or stuck subscribe never reaches `--limit`. File lock with F7.

## What to do

`--limit` must complete from history when the events already exist. A stuck subscribe cannot wedge the command.

## Acceptance

1. Against the pairdemo Core with finish already in SQLite, `events tail --since 13 --limit 30 --json` exits with those events.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #439, merged to `beta/0.4.5` via #481. Follow-ups from that work are filed as #441 and #442.

Issue #400, opened 2026-09-02
C: dropping a file on the Files drawer uploads it twice


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `C`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: panel-rendering-investigation-2026-09-02.md (symptom C) — not in the 44-id catalog

## What the operator sees

Drop a file on the Files drawer background/header/uploads list. The same file uploads twice (second pass `overwritten` or 409 transfer-in-progress).

## Where

`ProjectFilesPanel.tsx` root `onDrop` (190–197) `preventDefault` but not `stopPropagation`. Board `onDrop` in `projects.$id.tsx` 1830–1843 re-reads DataTransfer into a fresh array; the consume effect’s identity guard passes. Folder rows already `stopPropagation` (283–306) with a comment naming this hazard. Introduced 08-13/08-14 `d2b27ab` + `839c1f2`.

## What to do

Add `stopPropagation` on the panel root’s `onDrop` and `onDragOver`, mirroring folder rows — or ignore `e.defaultPrevented` on the board handler. Keep `projects.$id.tsx` out of the PR if you only need the panel-root fix, so G1/G4/G5 can own that file later.

## Acceptance

1. Drop on drawer background/header/empty state/non-folder row uploads once, to the project root.
2. Drop on a folder row still uploads once, to that folder.
3. Drop on the board outside the drawer still uploads once.

## Related

#222 is folder-as-tar / executable bit, not this double-fire.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #437, merged to `beta/0.4.5` via #481.

Issue #390, opened 2026-09-02
F4: foreign-session Stop is acked and ignored — no finish event


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `F4`
- Severity: broken
- First-fix owner: core
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Harness Stop arrives with a different `session_id` than the stored one (resume / OpenCode child). Card stays running; no `session:finished`.

## Where

`harness-hook-pipeline.ts` (catalog: 122–139, 187). Stop is not a capture event, so mismatch returns `foreign-session` before `updateStatus`. Same file as A-heal and F1 — one round, do not parallel.

## What to do

A Stop for this PTY/task must settle even when the harness session_id does not match, or nack visibly. Ack-and-ignore is the operator-visible miss.

## Acceptance

1. A Stop whose session_id does not match the stored id still finishes (or nacks) the Core task the PTY belongs to.
2. Matching-session Stop behaviour is unchanged.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #453, merged to `beta/0.4.5` via #481. The PR carried `Refs` rather than a closing keyword, so closing by hand.

Issue #387, opened 2026-09-02
F1: dead session stuck at ready — no session:finished, no pty:exit


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `F1`
- Severity: broken
- First-fix owner: core
- Live on pairdemo: yes, pairdemo 2026-09-02
- Source: operator-visible-bugs.md

## What the operator sees

Live pairdemo: `t-mtjw9r9q-d145d5` still “Waiting for initial prompt…”, status `ready`, LIVE empty, hours old. PTY spawned before the container recreate. CLI: “has no harness running.”

## Where

`sessionProcessExited` only patches `running|needs-input` (`harness-hook-pipeline.ts`). Boot sweep `listActiveTasks` / `queryActiveTasks` is the same filter — `ready` is invisible. That is why the zombie survived recreate.

## What to do

Treat `ready` as stranded: include it in `listActiveTasks` + boot sweep, and settle it on PTY death. Copy the live row out of pairdemo before recreating the Core, then confirm boot sweep marks it `disconnected` (or finished, if the exit was clean).

## Acceptance

1. A `ready` session whose PTY is gone settles (disconnected or finished) without waiting for a Stop hook.
2. Core recreate sweeps leftover `ready` rows the way it already sweeps `running` / `needs-input`.
3. Healthy OpenCode stop still emits `session:finished` (live event 22 on this hunt still did).

mrfqcentic:
Delivered in **0.4.5**. Fixed by #448, merged to `beta/0.4.5` via #481.

PR #438, opened 2026-09-02
fix(shared): stop post-turn subagent helpers healing a finished task back to running

### LINKED ISSUE #385: A-heal: a subagent event within 30s of finish writes the task back to running

- Hunt ID: `A-heal`
- Severity: broken
- First-fix owner: shared / core
- Source: panel-rendering-investigation-2026-09-02.md (symptom A) — not in the 44-id catalog

## What the operator sees

A session shows completed, then flips back to running. Clicking the just-finished pin is itself a trigger (Claude away-summary / title helpers). The drain backstop restores finished ~3–4 minutes later.

## Where

`packages/shared/src/harness-hook-pipeline.ts` subagent branch writes `running` when `task.status === "finished" && !staleFinished`. `RECENT_FINISH_WINDOW_MS = 30_000` in `subagent-activity.ts`. Drain is 60s recheck × 3 min grace.

## What to do

The window was built for a sub-second Stop-vs-subagent race and is tuned 30,000× that. Require an active-subagent set, or shrink the window to a real race. Do not heal post-turn helper SubagentStart/Stop into `running`. Speed the visible correction if anything still heals. Related structural note W1: three uncoordinated status writers, no arbiter — do not rewrite the event bus in this ticket.

## Acceptance

1. Finish a Claude turn, click the pin within 30s: the card stays finished (or a distinct settling state — not `running`).
2. A genuine in-turn subagent still holds finish until those subagents stop.
3. Existing Stop / OpenCode idle finish tests stay green.

## Related

#292 is the older “status does not track the turn” seam. This ticket is the verified resurrection writer.

qcentic-dev:
## Summary

The subagent branch of the hook pipeline un-finished a task whenever a `SubagentStart`/`SubagentStop`
arrived within `RECENT_FINISH_WINDOW_MS = 30_000` of the finish. Claude Code fires post-turn internal
helpers — away-summary generation, the title helper — when the operator refocuses a finished pane or
clicks the just-finished pin, and those carry the parent session id, so for a full 30 seconds after
every finish they read as resumed work: the card flipped back to `running` and stayed there until the
drain backstop restored it 3–4 minutes later.

A finished task is now un-finished on two conditions only: a subagent tracked from that turn is still
in flight, or the finish is younger than `FINISH_RACE_WINDOW_MS` — **one second, inclusive**, replacing
the 30s window. Everything else on a finished task is a post-turn helper: ignored for status, and its
start is not tracked either (a lost helper stop would otherwise hold the next turn's `Stop` for the
whole 2h TTL — pre-existing behaviour, kept).

**Which of the two conditions actually decides the raced-POST case? The clock, alone.** Every
hook-driven finish leaves the tracked set empty by construction: the `finished` mapping is downgraded
to `running` whenever `hasActiveSubagents` is true, `finishQuietTask` runs only after the drain
observed an idle set, `sessionProcessExited` calls `clearSubagentActivity` first, and the Core's
session backstop clears before it calls `noteTaskFinished`. The active-set disjunct is therefore not
the safety net here — it guards a `finished` written by one of the *other* W1 status writers (a
core-link task mutation through `CoreTaskWriter`, which clears nothing). Keeping it is right; it just
is not what protects in-turn work in this path.

**What the window costs.** It is sized on *emission* — the `Stop` and the turn's own `SubagentStart`
leave the same harness process microseconds apart — but evaluated on *arrival*: the finish is stamped
when the `Stop` POST was handled, and compared against when the subagent POST is handled. Delivery is
not microseconds. The hook command is `curl -sS -f -m 3 --retry 2 --retry-delay 1`
(`packages/core/src/harness-hooks.ts`), whose own comment bounds the worst case at *"about eleven
seconds"* under *"a Core busy serving PTY fan-out and SQLite writes"* — exactly what a fan-out turn
creates. So a retry-delayed in-turn `SubagentStart` landing after the `Stop` is dropped and never
tracked, leaving the card `finished` through a live fan-out with no backstop. **That is knowingly
traded away**, because a window wide enough for the ~11s retry tail puts a pin click at +5s back inside
it and re-opens #385. A single scalar clock cannot separate the two cases — that is the W1 structural
note this ticket defers. The residual is filed as **#440**.

The genuine in-turn hold path is untouched: an in-turn subagent is tracked before the foreground `Stop`
arrives, so that `Stop` is still downgraded to `running`, and only the `Stop` arriving with an idle set
lands the real finish.

### ⚠️ Deploy note — a Panel-only deploy will look like this fix failed

This is a `packages/shared` change, and for Core-owned Sessions the hook pipeline runs on **the Core**
(issue #84), not the Panel. `packages/core` depends on `@actana/shared` and bundles it into the Core
release tarball that `deploy/core.Dockerfile` installs. A refreshed Panel image carries none of this
for Sessions whose hooks POST to a Core. **Ship the Core image the Panel talks to.**

### Deliberately not in this PR

- **No event-bus rewrite and no status arbiter** — structural note W1, deferred by the ticket.
- **The 60s recheck × 3min drain grace is unchanged.** The issue's "speed the visible correction"
  would mean shortening the grace that lets a re-invoked main agent compose its follow-up turn;
  shrinking it risks prematurely finishing a live turn. With the heal gated, nothing spurious reaches
  the backstop.
- **F1 (#387) and F4 (#390)** also touch `harness-hook-pipeline.ts` and are later waves — untouched.

## Related issues

Closes #385
Refs #440 (the residual this trade creates), #292 (the older status-versus-turn seam)

## Type of change

- [x] 📚 Documentation (`docs/harness-status-detection.md` rewritten around the two heal conditions)
- [ ] 🔧 Build / CI / tooling

## How was this tested?

- [ ] Manually verified — **no**, and stated plainly: no live Panel was clicked for this change.
      AC1 is proven by test at the pipeline and HTTP layers, not by driving pairdemo.

New `packages/shared/src/__tests__/harness-hook-pipeline.test.ts` (7 tests) drives
`handleHarnessHookEvent` event by event:

1. a helper `SubagentStart`/`SubagentStop` after a finish leaves the task `finished`, with the recorded
   status writes exactly `["running", "finished"]`;
2. the regression pinned across the whole old window — 2s, 5s, 15s, 29s all stay `finished`;
3. a helper's start is not tracked, so the next turn's `Stop` still finishes;
4. a genuinely raced lifecycle POST still heals;
5. the active-set disjunct — reachable only by reaching past the pipeline, which is the point;
6. an in-turn subagent still holds the finish until the last one stops;
7. `Stop` finishes when the turn's subagents already reported in.

Panel: `harness-hooks-api.test.ts`'s race case pins `Date.now` so the window measures the race and not
the HTTP round trip, and its helper case fires at +5s — inside the old window — instead of +31s.
`subagent-activity.test.ts` asserts the window stays at one second inclusive.

`pnpm test` green (6/6 stages: root 23 files, sdk 22, shared 31, cli 50, core 57 / 903 tests, panel
146 / 1556 tests). `pnpm typecheck` clean. `pnpm lint` 0 errors. Existing Stop and OpenCode
idle-finish suites pass unchanged.

## Checklist

- [x] This PR targets the open pin/session/CLI/drop feature branch
      `fix/operator-pins-sessions-drop-cli`, not `main` and not `beta/0.4.5` — `Train rules` green
- [x] My PR title **and every commit in the branch** follow Conventional Commits, and the branch name
      follows the branching convention
- [x] I updated documentation (`docs/harness-status-detection.md`)
- [x] This PR is marked as **draft** and stacked on the feature branch

### Boundaries

Files owned by the tickets running in parallel on this branch are untouched: `ProjectFilesPanel.tsx`
and `pending-file-drop.ts` (#400), `events-command.ts` (#402).

https://claude.ai/code/session_01Pa35nN54tKU2AbVX9Neg7o

### DIFF
--- a/docs/harness-status-detection.md
+++ b/docs/harness-status-detection.md
@@ -60,8 +60,7 @@ lifecycle POST).
 
 A finished task is healed on **two conditions only**: a subagent tracked from
 that turn is still in flight, or the finish is younger than
-`FINISH_RACE_WINDOW_MS` (one second — both POSTs leave the same harness process
-microseconds apart, so the real race is sub-second). Everything else is one of
+`FINISH_RACE_WINDOW_MS` (one second, inclusive). Everything else is one of
 Claude Code's **post-turn internal helpers**, whose subagent events carry the
 parent session id: refocusing a finished session, or clicking its just-finished
 pin, generates an *away summary* or a title, firing `SubagentStart`/`SubagentStop`
@@ -69,12 +68,45 @@ with no `Stop` to follow. Those are ignored for status, and their starts are not
 tracked either — a lost helper stop would otherwise hold the next turn's `Stop`
 for the whole TTL.
 
-The window was 30 seconds until issue 385, which is 30,000× the race it was
-sized for: every helper firing within half a minute of a finish flipped the card
-back to `running`, and the operator watched a completed Session un-complete
-itself the moment they clicked it. Time is not what tells in-turn work from a
-helper — the tracked active-subagent set is. Widening this window again
-re-opens that bug.
+**Of those two, the clock is what decides the raced-POST case.** Every
+hook-driven finish leaves the tracked set empty by construction — the `finished`
+mapping is downgraded to `running` whenever `hasActiveSubagents` is true, the
+drain's `finishQuietTask` runs only after an idle set, `sessionProcessExited`
+clears the set first, and the Core's session backstop clears before it stamps
+the finish. The active-set condition therefore guards a `finished` written by one
+of the *other* status writers (a core-link task mutation through
+`CoreTaskWriter`, which clears nothing), not the race. Worth keeping — just not
+what protects in-turn work here.
+
+### What the window measures, and what that costs
+
+The window is sized on **emission**: the `Stop` and the turn's own
+`SubagentStart` leave the same harness process microseconds apart. It is
+evaluated on **arrival** — the finish is stamped when the `Stop` POST was
+*handled*, and compared against when the subagent POST is handled. Delivery is
+not microseconds: the hook command is
+`curl -sS -f -m 3 --retry 2 --retry-delay 1` (`packages/core/src/harness-hooks.ts`),
+whose own comment bounds the worst case at *"about eleven seconds"* and names
+the trigger as *"a Core busy serving PTY fan-out and SQLite writes"* — exactly
+what a fan-out turn creates.
+
+So a **retry-delayed in-turn `SubagentStart` is knowingly traded away**. One that
+eats a `-m 3` timeout lands ~4s after a `Stop` that already wrote `finished`,
+finds an empty set and a 4s-old finish, and is dropped *and* never tracked; the
+card reads `finished` through a live fan-out, and no backstop corrects it (every
+backstop below fixes a task stuck on `running`; none fixes one stuck on
+`finished`). That residual is filed as **issue 440**.
+
+The window was 30 seconds until issue 385. That covered the retry tail
+incidentally, at the price of every post-turn helper resurrecting the card for
+half a minute after *every* finish — the operator watched a completed Session
+un-complete itself the moment they clicked it. Widening it back to absorb the
+~11s retry budget puts a pin click at +5s inside it again and re-opens 385. A
+single scalar clock cannot tell a retry-delayed in-turn event from a post-turn
+helper, so one of the two has to lose, and the one visible on every finish is the
+one that was fixed. Telling them apart needs evidence rather than elapsed time —
+a payload discriminator, an emission timestamp, or the W1 status arbiter; issue
+440 sketches those.
 
 Backstops, so a `SubagentStop` that never arrives (lost POST, killed process) —
 or a healed `running` that no `Stop` will ever follow — cannot hold a task on
--- a/docs/harness-status-detection.md
+++ b/docs/harness-status-detection.md
@@ -58,14 +58,23 @@ maps to a status on its own, but one arriving **moments after** a task finished
 heals it back to `running` (a `Stop` won the race against the turn's own subagent
 lifecycle POST).
 
-That heal is scoped to a 30-second recent-finish window because Claude Code also
-runs **post-turn internal helpers** whose subagent events carry the parent
-session id: refocusing a finished session generates an *away summary*, firing
-`SubagentStart`/`SubagentStop` minutes after the finish, with no `Stop` to
-follow. Healing on those wedges tasks on `running` forever (the original
-stuck-on-running bug). Beyond the window, helper events are ignored for status
-and their starts are not tracked — a lost helper stop would otherwise hold the
-next turn's `Stop` for the whole TTL.
+A finished task is healed on **two conditions only**: a subagent tracked from
+that turn is still in flight, or the finish is younger than
+`FINISH_RACE_WINDOW_MS` (one second — both POSTs leave the same harness process
+microseconds apart, so the real race is sub-second). Everything else is one of
+Claude Code's **post-turn internal helpers**, whose subagent events carry the
+parent session id: refocusing a finished session, or clicking its just-finished
+pin, generates an *away summary* or a title, firing `SubagentStart`/`SubagentStop`
+with no `Stop` to follow. Those are ignored for status, and their starts are not
+tracked either — a lost helper stop would otherwise hold the next turn's `Stop`
+for the whole TTL.
+
+The window was 30 seconds until issue 385, which is 30,000× the race it was
+sized for: every helper firing within half a minute of a finish flipped the card
+back to `running`, and the operator watched a completed Session un-complete
+itself the moment they clicked it. Time is not what tells in-turn work from a
+helper — the tracked active-subagent set is. Widening this window again
+re-opens that bug.
 
 Backstops, so a `SubagentStop` that never arrives (lost POST, killed process) —
 or a healed `running` that no `Stop` will ever follow — cannot hold a task on
@@ -74,7 +83,7 @@ fourth is armed by nothing, which is the point (issue 243):
 
 - Tracked entries expire after 2 hours (kept long on purpose — a short TTL would
   prematurely finish sessions whose subagents legitimately run long).
-- A held `Stop` (and a recent-finish heal) arms a once-a-minute recheck. Once the
+- A held `Stop` (and a subagent heal) arms a once-a-minute recheck. Once the
   tracked set is idle — drained by real `SubagentStop`s or by expiry — a 3-minute
   drain grace starts: if the re-invoked main harness's own `Stop` lands the finish
   within it (the normal flow), the backstop's promotion is a no-op (it only fires
--- a/packages/panel/src/server/__tests__/harness-hooks-api.test.ts
+++ b/packages/panel/src/server/__tests__/harness-hooks-api.test.ts
@@ -18,7 +18,7 @@ const { TITLE_WAITING } = await import("~/lib/task-sentinels");
 const LOOPBACK_HEADERS = { origin: "http://127.0.0.1:5173" };
 const SESSION_ID = "00000000-0000-4000-8000-000000000000";
 
-// Some cases shift Date.now past the recent-finish heal window; always restore.
+// Some cases pin or shift Date.now around the finish race window; always restore.
 const realNow = Date.now;
 afterEach(() => {
   Date.now = realNow;
@@ -378,10 +378,15 @@ describe("background subagents over the claude hook API", () => {
   it("heals a finished task when a late subagent event loses the race to Stop", async () => {
     await prompt();
     // Stop wins the race against the just-launched subagent's SubagentStart.
+    // Both POSTs leave the same harness process microseconds apart, so the
+    // clock is pinned: the heal window is that race, not the round trip.
+    const raceInstant = realNow();
+    Date.now = () => raceInstant;
     expect((await stop()).status).toBe("finished");
 
     await subagent("SubagentStart", "late-sub");
     expect(getTask(taskId)?.status).toBe("running");
+    Date.now = realNow;
 
     await subagent("SubagentStop", "late-sub");
     expect((await stop()).status).toBe("finished");
@@ -418,10 +423,12 @@ describe("background subagents over the claude hook API", () => {
     await prompt();
     expect((await stop()).status).toBe("finished");
 
-    // Minutes later the user refocuses the pane and Claude Code's internal
-    // away-summary helper fires SubagentStart/Stop — with no Stop to follow.
-    // The finished status must hold (this was the stuck-on-running bug).
-    Date.now = () => realNow() + 31_000;
+    // Seconds later the user refocuses the pane, or clicks the just-finished
+    // pin, and Claude Code's internal away-summary helper fires
+    // SubagentStart/Stop — with no Stop to follow. The finished status must
+    // hold. Five seconds was INSIDE the old 30s heal window (issue 385): the
+    // card flipped back to running and stayed there until the drain backstop.
+    Date.now = () => realNow() + 5_000;
     await subagent("SubagentStart", "away-helper");
     expect(getTask(taskId)?.status).toBe("finished");
     await subagent("SubagentStop", "away-helper");
--- a/packages/panel/src/server/__tests__/subagent-activity.test.ts
+++ b/packages/panel/src/server/__tests__/subagent-activity.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 import {
+  FINISH_RACE_WINDOW_MS,
   armDeferredFinish,
   clearSubagentActivity,
   disarmDeferredFinish,
@@ -7,13 +8,12 @@ import {
   noteSubagentStart,
   noteSubagentStop,
   noteTaskFinished,
-  taskFinishedRecently,
+  taskFinishedWithinRaceWindow,
 } from "../services/subagent-activity";
 
 const TTL_MS = 2 * 60 * 60 * 1000;
 const RECHECK_MS = 60 * 1000;
 const DRAIN_GRACE_MS = 3 * 60 * 1000;
-const RECENT_FINISH_MS = 30 * 1000;
 
 const realNow = Date.now;
 
@@ -180,17 +180,23 @@ describe("deferred finish backstop", () => {
   });
 });
 
-describe("recent-finish window", () => {
-  it("reports a finish as recent only within the window", () => {
+describe("finish race window", () => {
+  it("reports a finish as raced only within the window", () => {
     const taskId = "task-finish-window";
-    expect(taskFinishedRecently(taskId)).toBe(false);
+    expect(taskFinishedWithinRaceWindow(taskId)).toBe(false);
 
     noteTaskFinished(taskId);
-    expect(taskFinishedRecently(taskId)).toBe(true);
+    expect(taskFinishedWithinRaceWindow(taskId)).toBe(true);
 
     // Beyond the window, a subagent event is a post-turn helper, not a raced
     // lifecycle POST from the finished turn.
-    Date.now = () => realNow() + RECENT_FINISH_MS + 1;
-    expect(taskFinishedRecently(taskId)).toBe(false);
+    Date.now = () => realNow() + FINISH_RACE_WINDOW_MS + 1;
+    expect(taskFinishedWithinRaceWindow(taskId)).toBe(false);
+  });
+
+  it("stays sub-second — the race it models, not a grace period (issue 385)", () => {
+    // The window was 30s, 30,000x the loopback POST race it was sized for,
+    // which let post-turn helper subagents resurrect a finished card.
+    expect(FINISH_RACE_WINDOW_MS).toBeLessThanOrEqual(1_000);
   });
 });
--- a/packages/panel/src/server/__tests__/subagent-activity.test.ts
+++ b/packages/panel/src/server/__tests__/subagent-activity.test.ts
@@ -194,9 +194,11 @@ describe("finish race window", () => {
     expect(taskFinishedWithinRaceWindow(taskId)).toBe(false);
   });
 
-  it("stays sub-second — the race it models, not a grace period (issue 385)", () => {
-    // The window was 30s, 30,000x the loopback POST race it was sized for,
-    // which let post-turn helper subagents resurrect a finished card.
+  it("stays at one second inclusive — a race window, not a grace period (issue 385)", () => {
+    // The window was 30s, which let post-turn helper subagents resurrect a
+    // finished card for half a minute after every finish. Widening it back to
+    // absorb the hook curl's ~11s retry tail re-opens that bug — see
+    // FINISH_RACE_WINDOW_MS and issue 440 for the residual that buys.
     expect(FINISH_RACE_WINDOW_MS).toBeLessThanOrEqual(1_000);
   });
 });
--- a/packages/panel/src/server/services/subagent-activity.ts
+++ b/packages/panel/src/server/services/subagent-activity.ts
@@ -5,8 +5,9 @@
 // process holds its own state; only the code is shared. Re-exported here to
 // preserve existing import paths.
 export {
+  FINISH_RACE_WINDOW_MS,
   noteTaskFinished,
-  taskFinishedRecently,
+  taskFinishedWithinRaceWindow,
   clearTaskFinished,
   noteSubagentStart,
   noteSubagentStop,
--- a/packages/shared/src/__tests__/harness-hook-pipeline.test.ts
+++ b/packages/shared/src/__tests__/harness-hook-pipeline.test.ts
@@ -20,7 +20,13 @@ import type { TaskStatus } from "../domain";
 // clicks the just-finished pin (away-summary generation, the title helper)
 // resurrected a completed card. A heal is legitimate only when the turn can
 // still plausibly be working — a tracked subagent is in flight, or the finish
-// is younger than the sub-second POST race.
+// is younger than FINISH_RACE_WINDOW_MS (one second, inclusive).
+//
+// For the raced-POST case the clock is the whole gate: every hook-driven finish
+// leaves the tracked set empty by construction, so the active-set disjunct only
+// covers a "finished" written by another status writer. What that costs — a
+// retry-delayed in-turn SubagentStart, dropped rather than healed — is filed as
+// issue 440.
 
 const SESSION_ID = "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa";
 
@@ -103,8 +109,9 @@ describe("subagent events on a finished task (issue 385)", () => {
   });
 
   it("stays finished across the whole old 30s heal window", () => {
-    // Every one of these used to write "running": the window was 30s, tuned
-    // 30,000x the sub-second race it was built for.
+    // Every one of these used to write "running": the window was 30s, where
+    // the race it models is the microseconds between two POSTs leaving the
+    // same harness process.
     for (const afterMs of [2_000, 5_000, 15_000, 29_000]) {
       const h = harness();
       h.post({ hook_event_name: "UserPromptSubmit", prompt: "do the thing" });
@@ -153,8 +160,12 @@ describe("subagent events on a finished task (issue 385)", () => {
     h.post({ hook_event_name: "UserPromptSubmit", prompt: "fan out" });
     h.post({ hook_event_name: "SubagentStart", agent_id: "sub-1" });
 
-    // One of the other status writers (structural note W1: three of them, no
-    // arbiter) lands "finished" while sub-1 is demonstrably still working.
+    // Reaching past the pipeline is the point, not a shortcut: no hook-driven
+    // finish can leave the tracked set non-empty, so this branch is only ever
+    // reachable from one of the OTHER status writers (structural note W1:
+    // three of them, no arbiter) landing "finished" over live work — a
+    // core-link task mutation through CoreTaskWriter, say, which clears
+    // nothing.
     h.ports.updateStatus(h.taskId, "finished");
 
     // Long past any POST race, but the turn's own set says work is in flight,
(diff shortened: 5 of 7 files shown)


qcentic-dev:
## Acceptance criteria — issue #385

- [x] **1. Finish a Claude turn, click the pin within 30s: the card stays finished (or a distinct settling state — not `running`).**
  The pin click is what makes Claude Code fire its away-summary / title helper `SubagentStart`/`SubagentStop`. Those now hit a
  finished task with an empty active-subagent set and a finish older than the 1s race window, so the pipeline ignores them for
  status — the card stays `finished`. No settling state was needed: nothing writes `running`, so there is nothing to settle from.
  Verified by test, not by clicking a live Panel — `harness-hook-pipeline.test.ts` "leaves the task finished when a helper
  SubagentStart follows the finish" and "stays finished across the whole old 30s heal window" (2s / 5s / 15s / 29s), plus the
  end-to-end HTTP case in `harness-hooks-api.test.ts` "ignores post-turn helper subagent events on a long-finished task", now
  fired at +5s (inside the old window) instead of +31s.

- [x] **2. A genuine in-turn subagent still holds finish until those subagents stop.**
  Untouched path: an in-turn subagent is tracked before the foreground `Stop` arrives, so that `Stop` is still downgraded to
  `running` and only the `Stop` arriving with nothing active is the real finish. Covered by `harness-hook-pipeline.test.ts`
  "holds running until the last background subagent stops" and "finishes on Stop when the turn's subagents already reported in",
  and by the existing `harness-hooks-api.test.ts` background-subagent suite (hold, multi-subagent hold, anonymous fallback).
  The active-set condition also lets a second in-turn subagent un-finish a card long past the race window —
  "heals past the race window while a tracked subagent is still in flight".

- [x] **3. Existing Stop / OpenCode idle finish tests stay green.**
  `pnpm test` — all 6 stages pass (root 23 files, sdk 22, shared 31, cli 50, core 57, panel 146; 903 core + 1556 panel tests).
  Specifically green: `harness-hooks-api.test.ts`, `opencode-hooks-api.test.ts` (including the `session.status` idle /
  `session.idle` double-fire settle), `claude-hook-status.test.ts`, `subagent-activity.test.ts`, and the Core's
  `harness-status-detection.test.ts`. `pnpm typecheck` and `pnpm lint` are clean (0 errors; the 46 lint warnings are
  pre-existing and none are in files this PR touches).

### Explicitly not done

- [ ] **"Speed the visible correction if anything still heals"** (from the issue's *What to do*, not an acceptance criterion).
  The 60s recheck × 3min drain grace is unchanged. That grace exists to let a re-invoked main agent compose its follow-up turn;
  shortening it trades this bug for premature finishes on live turns. With the heal gated on an active-subagent set or a
  sub-second race, nothing spurious reaches the backstop any more.
- [ ] **W1 (event bus / status arbiter)** — out of scope by the issue's own instruction. Not attempted.
- [ ] **F1 (#387) / F4 (#390)** — later waves that also touch `harness-hook-pipeline.ts`. Not attempted here.

### Deploy note

`packages/shared` change: the hook pipeline runs on the **Core** for Core-owned Sessions (issue 84).
**A Panel-only deploy will look like this fix failed.** Ship the Core image the Panel talks to.

qcentic-dev:
CI on `fix/subagent-heal-window` @ `55095f3` — **all 18 non-skipped checks pass**, none pending, none failing:

Conventions · Lint · Typecheck · Unit Tests · Secret Scan · Dependency Audit · Train rules · Promotion gate · Resolve PR image mode · E2E — Panel service seam · E2E — installer (debian, x64) · E2E — installer (ubuntu, x64) · Smoke — Core release tarball (linux-x64) · Smoke — Core release tarball (linux-arm64) · Core image / Resolve registries · Core image / Build + smoke (amd64) · Panel image / Resolve registries · Panel image / Build + smoke (amd64)

Skipping (train-only jobs, expected on a ticket PR): Core image (train), Panel image (train), Resolve train tags, Train versions, and both multi-arch manifest publishes.

Run: https://github.com/actana/control/actions/runs/33634567997

Left as a **draft** deliberately, base `fix/operator-pins-sessions-drop-cli`. Not merged, not marked ready.

mrfqcentic (changes requested):
## Review — PR #438 against issue #385 (A-heal)

Reviewed at head `7248905`, diffed against `origin/fix/operator-pins-sessions-drop-cli` in a
read-only worktree. The direction is right and the mechanism is the one the ticket asked for.
All three acceptance criteria are met by the enforcing lines, not just by the description.
I am requesting changes on two accuracy items only — **not** on the constant's value.

### Acceptance criteria — verified against the code

**AC1 — finish a turn, click the pin within 30s: card stays finished. MET.**
`packages/shared/src/harness-hook-pipeline.ts:215-218` replaces the old
`!staleFinished` test with `task.status !== "finished" || hasActiveSubagents(taskId) ||
taskFinishedWithinRaceWindow(taskId)`, and the `running` write at line 224 is gated on it.
A helper `SubagentStart`/`SubagentStop` on a task finished more than
`FINISH_RACE_WINDOW_MS` ago with an empty tracked set produces no status write and no
`noteSubagentStart` (line 220). `RECENT_FINISH_WINDOW_MS = 30_000` is gone from
`packages/shared/src/subagent-activity.ts` — no reference to it or to
`taskFinishedRecently` survives anywhere in the tree.

**AC2 — a genuine in-turn subagent still holds finish. MET.**
The hold path is untouched: `harness-hook-pipeline.ts:253` still downgrades a `finished`
mapping to `running` while `hasActiveSubagents` is true, and only the `Stop` arriving with an
idle set lands the finish (line 283). Covered by the new
`packages/shared/src/__tests__/harness-hook-pipeline.test.ts:169` and `:189`.

**AC3 — existing Stop / OpenCode idle-finish tests stay green. MET.**
`opencode-hooks-api.test.ts`, `claude-hook-status.test.ts` and the Core's
`harness-hooks-opencode-plugin.test.ts` are untouched by the diff, and `Unit Tests` (which
runs the whole workspace via `scripts/run-unit-tests.mjs`) is green on `7248905`. The new
shared suite matches `packages/shared/vitest.config.ts`'s `src/**/*.test.ts`, so it is
actually collected.

**Ordering requirement — MET.** `inTurn` is computed at `harness-hook-pipeline.ts:215`
*before* `noteSubagentStart` at `:220` and `noteSubagentStop` at `:222`, so a helper's own
start can never be the evidence that work is live. Good.

### Requested changes

**1. `packages/shared/src/subagent-activity.ts:45-56` — the constant's stated rationale is
measured on the wrong side of the wire.**

The comment says "Both are loopback POSTs emitted by the same harness process microseconds
apart, so the real race is sub-second". That is true of **emission**. The window is evaluated
on **arrival**: `noteTaskFinished` stamps when the `Stop` POST was handled
(`harness-hook-pipeline.ts:286`) and `taskFinishedWithinRaceWindow` compares that to when the
subagent POST is handled. Delivery is not microseconds — this repo's own hook command is
`curl -sS -f -m 3 --retry 2 --retry-delay 1` (`packages/core/src/harness-hooks.ts:114`),
whose comment states: *"Three attempts bound the worst case at about eleven seconds"*, and
names the trigger as *"a Core busy serving PTY fan-out and SQLite writes"* — which is exactly
the condition a subagent fan-out turn creates.

Concretely: an in-turn `SubagentStart` POST that eats one `-m 3` timeout lands ~4s late; if
the turn's `Stop` is delivered first, the task is written `finished` (`session:finished` ding
fires), and at +4s the `SubagentStart` finds an empty set and a 4s-old finish, so `inTurn` is
false — the event is dropped *and* the subagent is never tracked. The card reads `finished`
for the rest of a live fan-out, with no backstop to correct it (`PostToolUse` at
`harness-hook-pipeline.ts:275` heals only `needs-input`). The 30s window covered this; 1s
does not.

**To be explicit: do not fix this by widening the window.** A window wide enough to absorb
the retry budget (~11s) puts a pin click at +5s back inside it and re-opens #385 — the two
requirements are in direct conflict, which is the W1 structural note the ticket defers. What
I am asking for is that the comment stop asserting a premise the repo elsewhere contradicts,
and instead state the trade honestly: the window models emission spacing, it is evaluated on
arrival, delivery has a documented ~11s retry tail, and a retry-delayed in-turn
`SubagentStart` is knowingly traded away in exchange for killing the resurrection. Same
correction in `docs/harness-status-detection.md:63-64` and `:76-77`, where "Widening this
window again re-opens that bug" currently reads as the whole story. Please also file the
residual as a follow-up issue so the next engineer chasing a premature `finished` finds it.

**2. `packages/shared/src/harness-hook-pipeline.ts:217` — the active-set disjunct cannot fire
on any hook-driven finish, so the PR body overstates it as the safety net.**

The body says "Time is not what tells in-turn work from a helper — the tracked
active-subagent set is." But every path that lands `finished` leaves that set empty by
construction: `harness-hook-pipeline.ts:253` downgrades to `running` whenever
`hasActiveSubagents` is true, so the write at `:283` only happens with an idle set;
`finishQuietTask` (`:337`) runs only after the drain observed an idle set; the
`sessionProcessExited` branch (`:236`) calls `clearSubagentActivity` first; and
`core-session-backstop.ts:168-169` clears before `noteTaskFinished`. The tell is that the new
test which exercises this branch has to reach past the pipeline and call
`h.ports.updateStatus(h.taskId, "finished")` directly
(`packages/shared/src/__tests__/harness-hook-pipeline.test.ts:158`).

The disjunct is not dead — it guards a `finished` written by one of the *other* W1 writers
(e.g. a core-link task mutation through `CoreTaskWriter`, which clears nothing) — and keeping
it is right. But for the raced-POST case this PR is actually about, the effective gate is the
1s clock alone. Please say that in the comment at `:203-213` and in the PR body rather than
presenting the set as what protects in-turn work here.

### Non-blocking notes

- `subagent-activity.ts:102` uses `<=`, so exactly 1000ms passes. The comments and the test
  name at `packages/panel/src/server/__tests__/subagent-activity.test.ts:199` say
  "sub-second"; the bound is "one second inclusive". Trivial, but this repo's comments are
  load-bearing.
- `packages/panel/src/server/__tests__/subagent-activity.test.ts:200` asserts
  `FINISH_RACE_WINDOW_MS <= 1_000`. That pins a tuning constant rather than a behaviour, and
  it is the assertion any future retuning trips first. The behavioural test in the same PR
  ("stays finished across the whole old 30s heal window") is the one carrying the regression.
  Consider dropping the magnitude assertion.
- Pinning `Date.now` in `harness-hooks-api.test.ts:381-383` makes the race case near
  tautological (elapsed is 0, so any positive window passes). It is the right call for
  determinism, but it does mean nothing in the suite exercises real delivery lag against this
  window — which is the same gap as item 1.
- The PR title "do not heal a finished task back to running for 30s" can be read as "do not
  heal for 30 seconds". Since the title is squash-merged and becomes the changelog line,
  something like `fix(shared): stop post-turn helper subagents resurrecting a finished card`
  would land better.
- The body does not follow `.github/PULL_REQUEST_TEMPLATE.md` (no Type of change / How was
  this tested / Checklist sections). The substance is all there and more, so this is a nit —
  but CONTRIBUTING.md asks for the template.

### Boundaries — all clean

- Base is `fix/operator-pins-sessions-drop-cli` (at `a7b5532`, the merge-base), not `main`,
  not `beta/0.4.5`. `Train rules` green. PR is still a draft.
- No event-bus rewrite, no status arbiter: the diff is 7 files, and the only behavioural
  hunks are `harness-hook-pipeline.ts:215-226` and the constant/rename in
  `subagent-activity.ts`.
- F1 (#387) `sessionProcessExited` / active-task sweep and F4 (#390) foreign-session paths
  (`harness-hook-pipeline.ts:122-139`, `:187`, `:236`) are untouched — diff hunks start at
  line 193.
- `ProjectFilesPanel.tsx`, `pending-file-drop.ts` and `events-command.ts` are untouched.
- Deploy note present and accurate: `packages/core` depends on `@actana/shared` and bundles
  it into the Core release tarball that `deploy/core.Dockerfile` installs, so a Panel-only
  deploy really would look like the fix failed.
- Single commit `7248905`; longest body line is 80 chars against commitlint's
  `body-max-line-length` of 132, header is 64 against 120, `Refs #385` and the `Co-Authored-By`
  trailer are preceded by a blank line. `Conventions` green.
- CI on `7248905`: **18 checks pass, 6 skipped, 0 failing, 0 pending** — Conventions, Lint,
  Typecheck, Unit Tests, Secret Scan, Dependency Audit, Train rules, Promotion gate, both
  image builds, both tarball smokes, both installer E2Es, Panel service seam E2E.

Happy to re-review as soon as the two accuracy items land — the code itself I have no
objection to.

qcentic-dev:
@mrfqcentic — both accuracy items landed in `1d16b05`, plus the wording nits. **No behaviour changed:
`FINISH_RACE_WINDOW_MS` stays at `1_000` and the `inTurn` gate is byte-identical.** The follow-up issue
is **#440**.

---

### 1. The constant's rationale was measured on the wrong side of the wire — fixed

You were right, and the repo contradicted itself. The comment on `FINISH_RACE_WINDOW_MS` in
`packages/shared/src/subagent-activity.ts` now says, in order: the window is sized on **emission** (the
`Stop` and the turn's own `SubagentStart` leave the same harness process microseconds apart); it is
evaluated on **arrival** (`noteTaskFinished` stamps when the `Stop` POST was *handled*, and the
comparison runs when the subagent POST is handled); delivery is not microseconds, because the hook
command in `packages/core/src/harness-hooks.ts` is `curl -sS -f -m 3 --retry 2 --retry-delay 1`, whose
own comment bounds the worst case at *"about eleven seconds"* under *"a Core busy serving PTY fan-out
and SQLite writes"* — exactly the condition a fan-out turn creates.

It then states the trade rather than a premise: a retry-delayed in-turn `SubagentStart` that lands
after the `Stop` is dropped **and** never tracked, so the card reads `finished` through a live fan-out
with no backstop (every backstop in the system corrects a task stuck on `running`; none corrects one
stuck on `finished`). That is knowingly given up, because a window wide enough for the ~11s retry
budget puts a pin click at +5s back inside it and re-opens #385 — a single scalar clock cannot separate
a retry-delayed in-turn event from a post-turn helper, which is the W1 note this ticket defers.

Same correction in `docs/harness-status-detection.md`, which grew a **"What the window measures, and
what that costs"** section; "Widening this window again re-opens that bug" no longer stands as the whole
story.

**Follow-up filed: #440** — *"A-heal residual: a retry-delayed in-turn SubagentStart after Stop leaves
the card finished during a live fan-out"*. It carries your ~4s worked example, the "do not fix this by
widening the window" warning up top, the four candidate real fixes (payload discriminator, emission
timestamps, the W1 arbiter, a finished-side backstop symmetric with the drain), and a note that
`harness-hook-pipeline.test.ts` already has the harness to reproduce it on a controlled clock. It is
linked from the constant, the docs, the test header and the PR body.

### 2. The active-set disjunct cannot fire on a hook-driven finish — comment and body corrected

Also right, and I had it backwards in the body. The comment above the gate
(`packages/shared/src/harness-hook-pipeline.ts`) now names the four paths — the `finished` mapping
downgraded to `running` while `hasActiveSubagents` is true, `finishQuietTask` only after an idle
drain, `sessionProcessExited` clearing first, `core-session-backstop.ts` clearing before
`noteTaskFinished` — and concludes plainly: **for the raced-POST case the effective gate is the
one-second clock alone.** The disjunct is kept and described for what it actually guards: a `finished`
written by one of the *other* W1 writers, e.g. a core-link task mutation through `CoreTaskWriter`,
which clears nothing. The PR body now leads with the same statement instead of "the tracked
active-subagent set is [what tells them apart]".

The test that exercises the branch got the same treatment — its comment now says reaching past the
pipeline is the point, not a shortcut, since no hook-driven finish can leave the set non-empty.

### Non-blocking notes

- **"one second inclusive", not "sub-second"** — taken. The comparison is `<=`, so the comments,
  the docstring and the Panel test name now say one second inclusive.
- **PR title** — retitled to `fix(shared): stop post-turn subagent helpers healing a finished task back
  to running`, so the squashed changelog line cannot be read as "do not heal for 30 seconds". The new
  commit carries the same subject. The original fix commit `7248905` keeps its old subject: rewording
  it needs a force push, which the working instructions forbid. The squash takes the PR title, so the
  changelog line is correct.
- **PR body** — reshaped to `.github/PULL_REQUEST_TEMPLATE.md` (Summary / Related issues / Type of
  change / How was this tested / Checklist), with the Panel-only-deploy warning kept and promoted to
  its own subsection. "Manually verified" is left **unchecked** with the reason stated: no live Panel
  was clicked for this change.
- **Magnitude assertion** — *not* dropped. You flagged it as a "consider", and it is the assertion a
  future retuning should trip first; its comment now points at the ~11s retry tail and #440 so whoever
  trips it finds the argument rather than just a number. Happy to drop it if you would rather.
- **Pinned `Date.now` in the race test** — left pinned, for the determinism you noted. The gap you
  identified (nothing in the suite exercises real delivery lag against this window) is item 1 of #440.

### Verification

`pnpm test` green — 6/6 stages (root 23 files, sdk 22, shared 31, cli 50, core 57 / 903 tests, panel
146 / 1556 tests). `pnpm typecheck` clean, `pnpm lint` 0 errors.

CI on `1d16b05`: **18 checks pass, 0 failing, 0 pending** — Conventions, Lint, Typecheck, Unit Tests,
Secret Scan, Dependency Audit, Train rules, Promotion gate, both image builds, both tarball smokes,
both installer E2Es, Panel service seam E2E. Six train-only jobs skipped, as expected on a ticket PR.

Still a **draft**, base unchanged at `fix/operator-pins-sessions-drop-cli`, plain push, nothing merged.
Ready for your re-review.

mrfqcentic (approved):
## Review round 2 — PR #438 against issue #385 (A-heal)

**Approved.** Both round-1 blockers land, and they land the way I asked: as corrections to
what the code *claims*, with the code itself untouched. Re-reviewed at head `1d16b05`,
diffed against `origin/fix/operator-pins-sessions-drop-cli` (now `a74cc27`, moved since
round 1) in a fresh read-only worktree.

### The round-2 diff changes no executable behaviour — verified, not taken on trust

`git diff 7248905..1d16b05` touches 5 files. Stripping comment lines from
`packages/shared/src/harness-hook-pipeline.ts` and `packages/shared/src/subagent-activity.ts`
at both commits gives byte-identical files. Across all of `packages/`, the only changed line
that is not a comment is a test *title* string:

```
-  it("stays sub-second — the race it models, not a grace period (issue 385)", () => {
+  it("stays at one second inclusive — a race window, not a grace period (issue 385)", () => {
```

`FINISH_RACE_WINDOW_MS = 1_000` (`subagent-activity.ts:77`) and the `inTurn` expression
(`harness-hook-pipeline.ts:231-234`) are unchanged. That is exactly right — I asked for the
prose to stop asserting a premise the repo contradicts, not for a retune.

### Blocker 1 — the constant's rationale, measured on the right side of the wire now. RESOLVED.

`packages/shared/src/subagent-activity.ts:45-77` now states, in order: sized on **emission**;
evaluated on **arrival** (`noteTaskFinished` stamps when the `Stop` POST was handled);
delivery is not microseconds, naming `curl -sS -f -m 3 --retry 2 --retry-delay 1` and quoting
`packages/core/src/harness-hooks.ts` — I checked both quotes against lines 90-95 there, and
*"bound the worst case at about eleven seconds"* and *"a Core busy serving PTY fan-out and
SQLite writes"* are verbatim. It then states the trade rather than a premise: the
retry-delayed in-turn `SubagentStart` is dropped **and never tracked**, no backstop corrects
the `finished` direction, and that is knowingly given up because a window wide enough for the
~11s tail re-opens #385 at a +5s pin click.

`docs/harness-status-detection.md` carries the same correction in a new **"What the window
measures, and what that costs"** section, and *"Widening this window again re-opens that bug"*
no longer stands as the whole story. The residual is filed as **#440**, which reproduces the
~4s worked example, leads with the do-not-widen warning, and sketches four real fixes. It is
linked from the constant, the docs, the shared test header and the PR body.

`sub-second` is gone tree-wide — `grep -rn "sub-second"` over `*.ts`/`*.tsx`/`*.md` returns
nothing, and the docstring on `taskFinishedWithinRaceWindow` (`subagent-activity.ts:114-120`)
now says "one second inclusive" and points at the constant for the cost. The bound really is
`<=` (`:124`), so the wording and the comparison agree.

### Blocker 2 — the active-set disjunct is described for what it actually guards. RESOLVED.

The comment above the gate (`harness-hook-pipeline.ts:205-218`) now says it plainly: *"Which
of those two actually decides the raced-POST case this exists for? The clock, alone."* It
names all four paths, and I re-checked each against the enforcing lines rather than the prose:

- the `finished` mapping downgraded to `running` while `hasActiveSubagents` — `:269-276`;
- `finishQuietTask` only after an idle drain — `:350-354`, armed by `armDeferredFinish`, whose
  tick returns early while the set is non-empty (`subagent-activity.ts:220-232`);
- `sessionProcessExited` clears first — `:252` (and `clearTaskFinished` at `:255`, which is
  stronger still);
- the Core's session backstop — `core-session-backstop.ts:168` clears, `:169` stamps.

The disjunct is kept and correctly attributed to a `finished` written by another W1 writer;
`packages/panel/src/server/services/tasks.ts:104` confirms that path clears the set only for
`terminated`/`disconnected`, not for `finished`. `docs/harness-status-detection.md` and the
PR body's Summary both lead with the same statement, and the test that exercises the branch
(`harness-hook-pipeline.test.ts:163-169`) now says reaching past the pipeline is the point.

### Round-1 non-blocking notes

- **"one second inclusive"** — taken, everywhere including the Panel test name. Good.
- **PR title** — retitled to `fix(shared): stop post-turn subagent helpers healing a finished
  task back to running`. Cannot be read as "do not heal for 30 seconds". The note about
  `7248905` keeping its old subject is correct and the right call: the squash takes the PR
  title, and rewording a pushed commit would need a force push.
- **PR body** — now follows `.github/PULL_REQUEST_TEMPLATE.md`, with the ⚠️ Panel-only-deploy
  warning kept and promoted to its own subsection. "Manually verified" left unchecked with the
  reason stated — that is the honest box to leave unchecked, thank you.
- **Magnitude assertion (`FINISH_RACE_WINDOW_MS <= 1_000`)** — **my call: keep it, as you
  did.** You are right that it is the assertion a future retuning should trip first, and the
  comment you added turns it from a bare number into a tripwire that hands the reader the
  argument and #440. The behavioural test carries the regression independently, so nothing is
  load-bearing on the magnitude alone. No change wanted.
- **Pinned `Date.now`** — **keep it pinned**, also as you did. Determinism is worth more here
  than a test that would only be measuring the harness's own round trip, and the gap it leaves
  is item 1 of #440.

### New non-blocking notes from a fresh full-diff pass

I re-ran a code-review pass over the whole PR range against the moved base. Nothing here
blocks — all three are edges in code I already passed in round 1, and none is introduced by
the round-2 diff. Filing them so they are on the record; #440 looks like the right home.

1. **`harness-hook-pipeline.ts:233` — the new disjunct adds one way back into an existing
   bounded wedge.** A `finished` from a non-hook writer leaves the tracked set intact
   (`tasks.ts:104` clears only on `terminated`/`disconnected`). If that set holds a *leaked*
   entry — a subagent whose `SubagentStop` was lost, which is what the 2h TTL exists for —
   then a post-turn helper `SubagentStart` arriving hours later now sees
   `hasActiveSubagents === true`, heals to `running`, and is itself tracked. Under the old
   `!staleFinished` test that heal was blocked. It is bounded by the same TTL + 3-minute drain
   that already bounds the identical wedge on the `Stop`-downgrade path (`:269`), so this is a
   widening of an existing hazard rather than a new one — and it is the price of the disjunct I
   asked you to keep. Worth a line in #440.

2. **`harness-hook-pipeline.ts:238` — the start/stop asymmetry now covers a wider band.** A
   dropped helper `SubagentStart` still has its `SubagentStop` fed to `noteSubagentStop`
   unconditionally, and that function's cross-cancel branches (`subagent-activity.ts:174-190`)
   will decrement `anonCount`, or delete the oldest keyed start, when the id does not match.
   With a helper start dropped, a new prompt submitted before the helper's stop lands, and
   payload-shape skew between the two, the helper's stop can cancel the *new* turn's subagent
   and let the next `Stop` finish over live work. This is pre-existing — the base branch does
   the same for any helper firing >30s after a finish, which is the common refocus case — but
   this PR extends the band to 1s–30s. Narrow (needs the interleave *and* the skew), and the
   fix is small: ignore a stop for a start that was never recorded.

3. **`harness-hook-pipeline.ts:240` — a helper that fires inside 1s still heals.** The PR
   reasons about helpers at human latency (refocus, pin click), which is right for those. A
   title helper firing immediately at turn end lands inside the window, heals to `running`,
   gets tracked, and sits there for the drain grace — #385 in miniature, ~4 minutes. Inherent
   to any non-zero window and not a reason to change the constant; the tests only exercise
   helpers at ≥2s, so it is unpinned in either direction.

4. **Nit:** `docs/adr/0020-the-core-detects-its-own-harness-status.md:22` still says
   "30-second recent-finish heal window". ADRs are records of a decision at a point in time,
   so leaving it is defensible — but if you want the number to stay true, this is the last
   place it is stale. `packages/core/src/core-harness-status.ts:12`'s "the recent-finish heal"
   reads fine as a mechanism name.

### Mechanics

- **Draft:** yes. **Base:** `fix/operator-pins-sessions-drop-cli`, unchanged. Merge base
  `a7b5532`.
- **Merges clean onto the moved tip.** The base advanced `a7b5532 → a74cc27` when #437 merged;
  `git merge-tree --write-tree` against `a74cc27` succeeds with no conflict, and GitHub reports
  `MERGEABLE` / `CLEAN`. **No overlap:** #437 touched `ProjectFilesPanel.tsx` and
  `project-files-panel.test.tsx`; this PR touches neither, and the two file sets are disjoint.
- **Commits:** two, `7248905` and `1d16b05`. `commitlint --from a7b5532 --to 1d16b05` against
  this repo's own config: **0 problems, 0 warnings**. Longest body line 98 against
  `body-max-line-length` 132; headers 64 and 84 against 120; `Refs #385` / `Refs #440` and the
  trailers preceded by a blank line.
- **CI on `1d16b05`:** **18 pass, 6 skipped, 0 failing, 0 pending** — Conventions, Lint,
  Typecheck, Unit Tests, Secret Scan, Dependency Audit, Train rules, Promotion gate, Resolve
  PR image mode, both image resolve+build jobs, both tarball smokes, both installer E2Es, Panel
  service seam E2E. The six skips are the train-only jobs, expected on a ticket PR.
- **Boundaries** unchanged from round 1: 7 files, no event-bus rewrite, no arbiter; F1 (#387)
  and F4 (#390) paths untouched; `ProjectFilesPanel.tsx`, `pending-file-drop.ts` and
  `events-command.ts` untouched.

Nothing further from me. The trade this PR makes is now written down where the next engineer
will find it, and #440 carries the part that was given up.

Issue #403, opened 2026-09-02
F7: events tail --kind session:finished --limit 1 skips history and never exits


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `F7`
- Severity: broken
- First-fix owner: cli
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

After a session has already finished, `events tail --kind session:finished --limit 1` with no `--since` never prints and never exits. Ctrl-C leaves the cursor past the finish.

## Where

`events-command.ts` 151–215; `event-tip.ts` 17–34. `DurableCoreClient.deliverEvent` advances the cursor during the tip hunt. File lock with F6.

## What to do

Kind filter must print matching history, not skip it while advancing the cursor.

## Acceptance

1. After a finish is in the log, that command prints one `session:finished` and exits.
2. Ctrl-C does not persist a cursor past unseen matching events.

PR #439, opened 2026-09-02
fix(cli): events tail --limit must exit from history

qcentic-dev:
Fixes the hang in `actana events tail --limit` reported as #402 (hunt `F6`, live on pairdemo 2026-09-02).

## What the operator saw

`actana events tail --since 13 --limit 30 --json` sat until it was killed. The events past `#13` were already in
SQLite — `session:finished` among them, as `#22` — and the command had already printed them. The only way out of the
printing loop in `events-command.ts` was a thirtieth event that nobody was going to append: `--limit` was being waited
on as a quota rather than applied as a ceiling.

## What changed

`packages/cli/src/events-command.ts`, and nothing else in the tree.

1. **A `--limit` run that read history ends at the end of that history.** A run given somewhere to start from —
   `--since`, or a cursor a previous run left — that finds events there has read what it was asked for, and exits
   however far short of `n` it stops. The end of the log is found the way a first run finds the tip and by the same
   code, `event-tip.ts`: one marker at a time until one closes an *empty* tail, because a single `eventsReplayed` is a
   receipt for what the Core sent and not a statement that it has no more.

2. **A run that found no history is deliberately left alone.** Nothing past the cursor is what a follow looks like at
   the moment it starts — `--since start` against a log not yet written to. `events-tail-cursor.test.ts` is exactly
   that run, across a real Core restart, proving #161's criterion; there was no read to finish, so it follows and stops
   at `n`, which is what `--limit` has always meant. This is the narrowing that keeps that suite green rather than
   rewritten.

3. **A stuck subscribe cannot wedge any of them.** A contended Core that never answers `subscribe` sends no event and
   no marker, so neither the ceiling nor the end of the log is ever reached. **Every** `--limit` run — with a cursor or
   without one, walking to the end of a history or walking to the tip — holds a deadline on its patience until the log's
   end is known: the same 30s and the same argument `harness install` already carries, and listed with its reason in
   `no-prompt-timing.test.ts` rather than waved through. It re-sends nothing, it cannot make an event arrive sooner, and
   its expiry is reported as this side giving up: what was printed goes out, stderr says the Core stopped answering and
   that the read did not finish, and the run exits **non-zero** — a read that stopped partway is not a read that
   finished, and a script reads `$?` rather than English. It is disarmed once the log's end is known (past that point
   silence is a Core with nothing to say, and waiting through it is the whole of what a follow does), and it does not
   run while the link is down — a dropped socket is not a Core refusing to answer, and `DurableCoreClient` re-dials and
   replays the whole tail from the cursor.

The `--limit` help text is rewritten to say all of this in the shape an operator reads.

## How it is proven

This machine cannot reach the pairdemo Core, so the proof is at the CLI's own level, against a fake Core seeded with a
history the subscribe then fails to close. Four new cases in `packages/cli/src/__tests__/events-command.test.ts`:

- **the pairdemo run, frame for frame** — cursor at `#13`, ceiling of 30, a log that ends at `#22` with the finish in
  it: nine events on stdout and an exit;
- **history sent, then silence** — the subscribe delivers the tail and then never replays and never pushes: the nine
  events are printed and the run exits rather than holding them hostage to a marker that is not coming;
- **a subscribe that answers nothing at all** — no output, a reason on stderr, exit `1`, and no hang;
- **the two guards on the narrowing** — a run that found no history still follows, and the deadline is really disarmed
  once the Core has said where its log ends (three deadlines' worth of quiet, still following).

The first three of those fail on `HEAD~1` **by hanging**, which is the defect:

```
× exits from history when the log holds fewer events than --limit
× prints the history it was given when the subscribe never replays or pushes
× does not wedge on a subscribe that answers nothing at all
Tests  3 failed | 10 passed (13)
```

With the fix, the whole client package is green: `1190 passed | 3 skipped` across 51 files, including
`events-tail-cursor.test.ts` (the real in-process Core, a real drop, a real restart) and `no-prompt-timing.test.ts`.
`pnpm typecheck` and `pnpm lint` are clean.

## Review round 1 (`bc7f514`)

Three blocking findings and two should-fix, all addressed; the non-blocking notes taken too.

- **1 — the deadline was armed only on `--limit` *plus a cursor*.** `events tail --limit 30` with neither took the other
  side of the printing switch and could still hang for ever. Armed on both sides now, and disarmed where the tip is
  found as well as where a read ends.
- **2 — the deadline ran while the link was down**, so a Core restart truncated a read the reconnect was about to
  complete. Cleared on `onDisconnected`, restarted from `onReady`.
- **3 — `no-prompt-timing.test.ts`'s header argued one row while the table held two.** Rewritten to argue both, and to
  state the three properties a listed timer must have rather than leave them implicit.
- **4 — the exit code and the message disagreed**, so a truncated read could report success. An expiry now always exits
  non-zero and says the read did not finish, with everything it did get still on stdout.
- **5 — an approximation was reported as a finished read.** `trackEventTip` grows an `onRoundsExhausted` hook, so the
  caller that stops there says so in its own words and exits non-zero instead of printing "carrying on" at an operator
  whose command has already ended.
- **Non-blocking, taken:** `events-tail-cursor.test.ts` now reads 1500 events past a cursor **against a real Core**,
  proving the printing-side capped-tail walk delivers each exactly once and in order — the #205 shape, on a path that
  did not exist when #205 was reviewed. The two trackers keep separate allocations (they now differ in their
  end-of-hunt prose, which is the one place they are not the same walk) but the invariant that only one is ever live is
  spelled out where they are created.

Five new cases; four fail on `d611d42`, the no-cursor one by hanging.

## Follow-up

`--limit` now carries two meanings, and which one applies is inferred from the state of the log rather than said by the
operator. That is faithful to #402, whose remedy is scoped to "when the events already exist", but it is intent inferred
from log state. **#441** proposes an explicit `--follow` / `--no-follow` so it stops being inferred.

## Boundaries

The `--kind` filter defect in this same file is #403 and is untouched here. Nothing outside `packages/cli` is touched —
no Panel files (#400), nothing in `packages/shared` (#385).

Base is `fix/operator-pins-sessions-drop-cli`, the shared feature branch for this wave. Draft on purpose.

Refs #402
Follow-up: #441

https://claude.ai/code/session_01EZJh1v8cg8ZJAkxTAV4VLt

### DIFF
--- a/packages/cli/src/__tests__/events-command.test.ts
+++ b/packages/cli/src/__tests__/events-command.test.ts
@@ -271,11 +271,16 @@ describe("actana events tail", () => {
     await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS);
 
     const result = await run;
-    expect(result.code).toBe(EXIT_OK);
+    // Every event it got is on stdout and stays there — but the run was cut
+    // off, not finished, and the status says so. A short file with a zero
+    // status is the one shape a scripted consumer cannot tell from a complete
+    // read (#402 review, finding 4).
+    expect(result.code).toBe(EXIT_FAILURE);
     expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([
       14, 15, 16, 17, 18, 19, 20, 21, 22,
     ]);
     expect(result.err.join("\n")).toContain("stopped answering");
+    expect(result.err.join("\n")).toContain("did not finish");
     expect(core.closed).toBe(true);
   });
 
@@ -296,6 +301,108 @@ describe("actana events tail", () => {
     expect(result.out).toHaveLength(0);
     expect(result.err.join("\n")).toContain("stopped answering");
     expect(core.closed).toBe(true);
+  });
+
+  it("does not wedge a --limit run that has no cursor and no --since either", async () => {
+    // #402 review, finding 1. `actana events tail --limit 30` is the invocation
+    // an operator is most likely to type, and it takes the *other* side of the
+    // printing switch: no cursor, so the run is walking the log to find its tip
+    // before it prints anything. A Core that never answers `subscribe` wedges
+    // that walk exactly as it wedges a read — the deadline covers both sides,
+    // or it does not cover the reported symptom.
+    vi.useFakeTimers();
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--json", "--limit", "30"], {
+      connect: core.connect,
+    });
+    await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS);
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_FAILURE);
+    expect(result.out).toHaveLength(0);
+    expect(result.err.join("\n")).toContain("stopped answering");
+    expect(core.closed).toBe(true);
+  });
+
+  it("keeps timing a tip hunt the Core is still answering, and stops at the tip", async () => {
+    // The same side, behaving. Each frame re-arms the clock, so a Core walking
+    // a long log slowly is a Core answering — and once it says where the log
+    // ends the clock comes off for good, because from there the run is a follow
+    // and a follow waits through any amount of quiet.
+    vi.useFakeTimers();
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--json", "--limit", "1"], { connect: core.connect });
+    await vi.advanceTimersByTimeAsync(0);
+
+    for (const round of [1, 2, 3]) {
+      // Most of a deadline's worth of silence, then an answer. Three times.
+      await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS - 1_000);
+      core.emitEvent({ eventId: round, kind: "task:updated" });
+      core.emitReplayed(round);
+      await vi.advanceTimersByTimeAsync(0);
+    }
+    core.emitReplayed(3);
+    await vi.advanceTimersByTimeAsync(0);
+
+    // The tip is known and printing is on. Now an hour of quiet, which a follow
+    // is entitled to sit through.
+    await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS * 120);
+    core.emitEvent({ eventId: 4, kind: "session:finished" });
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([4]);
+  });
+
+  it("stops the clock while the link is down, and restarts it on the reconnect", async () => {
+    // #402 review, finding 2. A dropped socket is not a Core refusing to answer:
+    // nothing can arrive to re-arm the deadline until the link is back, and the
+    // durable client re-dials for as long as it takes and replays the whole tail
+    // from the cursor. A deadline left running would fire through a Core restart
+    // and truncate a read the reconnect was about to complete — and this file
+    // promises an operator that a restart costs neither a repeat nor a gap.
+    vi.useFakeTimers();
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--since", "13", "--limit", "30", "--json"], {
+      connect: core.connect,
+    });
+    await vi.advanceTimersByTimeAsync(0);
+
+    core.emitEvent({ eventId: 14, kind: "task:updated" });
+    core.emitDisconnected("socket hang up");
+
+    // A Core restart, taking far longer than the deadline it is not subject to.
+    await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS * 4);
+
+    let ended = false;
+    void run.then(() => {
+      ended = true;
+    });
+    await vi.advanceTimersByTimeAsync(0);
+    expect(ended, "the deadline fired through a reconnect").toBe(false);
+
+    // …and the link comes back, replays the rest from the cursor, and the read
+    // finishes at the end of the log the way it would have without the drop.
+    core.emitReplayed(14);
+    for (let eventId = 15; eventId <= 22; eventId += 1) {
+      core.emitEvent({ eventId, kind: eventId === 22 ? "session:finished" : "task:updated" });
+    }
+    core.emitReplayed(22);
+    await vi.advanceTimersByTimeAsync(0);
+    core.emitReplayed(22);
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([
+      14, 15, 16, 17, 18, 19, 20, 21, 22,
+    ]);
+    expect(result.err.join("\n")).toContain("reconnecting");
   });
 
   it("still follows when the log had nothing past where the run started", async () => {
@@ -322,6 +429,43 @@ describe("actana events tail", () => {
     const result = await run;
     expect(result.code).toBe(EXIT_OK);
     expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([1, 2]);
+  });
+
+  it("does not report a read finished when the walk settled for an approximation", async () => {
+    // #402 review, finding 5. `event-tip.ts` gives up after MAX_TIP_ROUNDS
+    // against a Core appending faster than this side can drain it, writes
+    // "carrying on from #N" and hands back a number that may be behind the
+    // log's end. That sentence was written for a first run, which really does
+    // carry on. A read takes the number as the end of the history and stops —
+    // so it must say so itself, and must not call a knowingly incomplete read a
+    // finished one.
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--json", "--since", "start", "--limit", "100000"], {
+      connect: core.connect,
+    });
+    await settle();
+
+    // MAX_TIP_ROUNDS is 200 and lives in `event-tip.ts`; the round after the
+    // last one it will take is where the hunt gives up. Every tail here is
+    // non-empty, which is what keeps it asking.
+    for (let round = 1; round <= 201; round += 1) {
+      core.emitEvent({ eventId: round, kind: "task:updated" });
+      core.emitReplayed(round);
+    }
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_FAILURE);
+    // Everything it did read is still on stdout — the status is what says the
+    // read is short, not a missing line.
+    expect(result.out).toHaveLength(201);
+    const said = result.err.join("\n");
+    expect(said).toContain("faster than they could be read");
+    expect(said).toContain("did not finish");
+    // The other caller's sentence, which would be a lie here.
+    expect(said).not.toContain("carrying on");
+    expect(core.closed).toBe(true);
   });
 
   it("ends a read whose history matched no --kind, rather than following", async () => {
--- a/packages/cli/src/__tests__/events-command.test.ts
+++ b/packages/cli/src/__tests__/events-command.test.ts
@@ -11,14 +11,14 @@
 // be the replay storm the ticket names, produced deliberately on the first
 // command an operator types.
 
-import { describe, it, expect, afterEach } from "vitest";
+import { describe, it, expect, afterEach, vi } from "vitest";
 import {
   fakeCore,
   makeCliFixture,
   registerCore,
   type CliFixture,
 } from "./cli-harness.ts";
-import { EXIT_OK, EXIT_USAGE } from "../exit-codes.ts";
+import { EXIT_FAILURE, EXIT_OK, EXIT_USAGE } from "../exit-codes.ts";
 
 let fixture: CliFixture | null = null;
 function cli(): CliFixture {
@@ -28,7 +28,15 @@ function cli(): CliFixture {
 afterEach(() => {
   fixture?.cleanup();
   fixture = null;
+  vi.useRealTimers();
 });
+
+/**
+ * The 30s a bounded run gives a Core that has gone quiet, jumped rather than
+ * waited out. `events-command.ts` holds the constant; a test that hard-coded a
+ * shorter one would be testing a different command.
+ */
+const SUBSCRIBE_ANSWER_MS = 30_000;
 
 async function withRegisteredCore(): Promise<void> {
   registerCore(cli().paths, "prod");
@@ -195,6 +203,155 @@ describe("actana events tail", () => {
     // A consumer parsing stdout must never have to recognise a status line.
     expect(result.out).toHaveLength(1);
     JSON.parse(result.out[0]!);
+  });
+
+  // ─── #402: --limit reads history and exits ────────────────────────────────
+  //
+  // On live pairdemo, `events tail --since 13 --limit 30 --json` sat until it
+  // was killed. The nine events past #13 were already in SQLite — the
+  // `session:finished` the operator was waiting for among them — and the
+  // command printed them and then waited for twenty-one more that no one was
+  // ever going to append. `--limit` is a ceiling on a read of the log, not a
+  // quota the command blocks on.
+
+  it("exits from history when the log holds fewer events than --limit", async () => {
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    // The pairdemo run, frame for frame: a cursor at #13, a ceiling of 30, and
+    // a Core whose log ends at #22 with the finish already in it.
+    const run = cli().run(["events", "tail", "--since", "13", "--limit", "30", "--json"], {
+      connect: core.connect,
+    });
+    await settle();
+
+    for (let eventId = 14; eventId <= 21; eventId += 1) {
+      core.emitEvent({ eventId, kind: "task:updated" });
+    }
+    core.emitEvent({ eventId: 22, kind: "session:finished", taskId: "t-1" });
+    core.emitReplayed(22);
+    await settle();
+
+    // One marker is a receipt for what was sent, not a statement that the log
+    // has ended, so the run asks again from where it got to — the same
+    // discipline a first run's tip hunt uses (`event-tip.ts`).
+    expect(core.subscribes).toContain(22);
+    core.emitReplayed(22);
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    // Nine events, not thirty, and the command is back at the prompt.
+    expect(result.out).toHaveLength(9);
+    const rows = result.out.map((line) => JSON.parse(line));
+    expect(rows[0].eventId).toBe(14);
+    expect(rows[8]).toMatchObject({ eventId: 22, kind: "session:finished" });
+    expect(core.closed).toBe(true);
+  });
+
+  it("prints the history it was given when the subscribe never replays or pushes", async () => {
+    // The other half of #402: the Core is contended, the subscribe answers with
+    // a tail and then goes silent — no `eventsReplayed`, no live push, nothing
+    // to close the loop on. The events are already here; a command that holds
+    // them hostage to a marker that is not coming is the hang, reported.
+    vi.useFakeTimers();
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--since", "13", "--limit", "30", "--json"], {
+      connect: core.connect,
+    });
+    await vi.advanceTimersByTimeAsync(0);
+
+    for (let eventId = 14; eventId <= 22; eventId += 1) {
+      core.emitEvent({ eventId, kind: eventId === 22 ? "session:finished" : "task:updated" });
+    }
+    // And now the Core says nothing at all, for as long as anyone is willing to
+    // wait. Nobody is: the deadline is re-armed on every frame, so it runs from
+    // the last thing the Core actually said.
+    await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS);
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([
+      14, 15, 16, 17, 18, 19, 20, 21, 22,
+    ]);
+    expect(result.err.join("\n")).toContain("stopped answering");
+    expect(core.closed).toBe(true);
+  });
+
+  it("does not wedge on a subscribe that answers nothing at all", async () => {
+    // Nothing printed and nothing to print: the run still ends, and it ends
+    // saying why rather than exiting 0 on a Core it never heard from.
+    vi.useFakeTimers();
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--since", "13", "--limit", "30", "--json"], {
+      connect: core.connect,
+    });
+    await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS);
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_FAILURE);
+    expect(result.out).toHaveLength(0);
+    expect(result.err.join("\n")).toContain("stopped answering");
+    expect(core.closed).toBe(true);
+  });
+
+  it("still follows when the log had nothing past where the run started", async () => {
+    // The case #402 deliberately leaves alone, and the one `events-tail-cursor`
+    // is built on: a cursor with an empty log past it is what a follow looks
+    // like at the moment it starts, so there is no read to finish and `--limit`
+    // keeps the meaning it has always had. Ending here would end the run before
+    // its first line.
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--json", "--since", "start", "--limit", "2"], {
+      connect: core.connect,
+    });
+    await settle();
+
+    // Caught up, with nothing to have caught up on.
+    core.emitReplayed(0);
+    await settle();
+
+    core.emitEvent({ eventId: 1, kind: "task:created" });
+    core.emitEvent({ eventId: 2, kind: "session:finished" });
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([1, 2]);
+  });
+
+  it("stops timing the Core once it has said where its log ends", async () => {
+    // The deadline covers a subscribe that never answers, not a Core with
+    // nothing to say. Once the end of the log is known the run is following,
+    // and a follow that gave up after thirty quiet seconds would be a worse
+    // hang than the one #402 fixed — silent, and on a Core behaving perfectly.
+    vi.useFakeTimers();
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--json", "--since", "start", "--limit", "1"], {
+      connect: core.connect,
+    });
+    await vi.advanceTimersByTimeAsync(0);
+    core.emitReplayed(0);
+    await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS * 3);
+
+    // Still here, an hour of quiet later.
+    let ended = false;
+    void run.then(() => {
+      ended = true;
+    });
+    await vi.advanceTimersByTimeAsync(0);
+    expect(ended).toBe(false);
+
+    core.emitEvent({ eventId: 1, kind: "session:finished" });
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    expect(JSON.parse(result.out[0]!).eventId).toBe(1);
   });
 
   it("rejects a --since or --limit that is not a number", async () => {
--- a/packages/cli/src/__tests__/events-command.test.ts
+++ b/packages/cli/src/__tests__/events-command.test.ts
@@ -324,6 +324,35 @@ describe("actana events tail", () => {
     expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([1, 2]);
   });
 
+  it("ends a read whose history matched no --kind, rather than following", async () => {
+    // What decides whether there was history to read is what the Core had, not
+    // what the filter let through. A read of a log that held nothing this
+    // operator asked for still finished, and has nothing left to wait for —
+    // reading `printed` here would hang exactly where #402 hung, one flag over.
+    //
+    // The first-run half of the kind filter — history skipped while the cursor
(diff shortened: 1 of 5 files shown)


qcentic-dev:
## Acceptance criteria — #402

Issue #402 states one criterion. It is listed first, then the two sentences of "What to do" it is the acceptance for,
since they are separately checkable.

- [ ] **1. Against the pairdemo Core with the finish already in SQLite, `events tail --since 13 --limit 30 --json`
  exits with those events.**
  **Unchecked, and it stays unchecked here.** This machine cannot reach the pairdemo Core, so nothing in this PR can
  check it. **The live pairdemo run is the orchestrator's prove step.** What is done instead is the same run at the
  CLI's own level, frame for frame — a cursor at `#13`, a ceiling of 30, a log that ends at `#22` with
  `session:finished` in it — asserting nine events on stdout and an exit
  (`exits from history when the log holds fewer events than --limit`, `packages/cli/src/__tests__/events-command.test.ts`).
  That test hangs on `HEAD~1`, which is the reported defect.

- [x] **2. `--limit` completes from history when the events already exist.**
  A run given somewhere to start from — `--since`, or a cursor a previous run left — that finds events past that
  cursor now ends at the end of that history, however far short of `n` it stops. The end is found by `event-tip.ts`,
  one marker at a time until one closes an empty tail, because a single `eventsReplayed` is a receipt for what the Core
  sent rather than a statement that it has no more.
  *Scoped, and the scope is stated rather than assumed:* a run that finds **no** history past its cursor still follows
  and stops at `n`. That case is indistinguishable from a follow at the moment it starts, and it is the run
  `events-tail-cursor.test.ts` is built on across a real Core restart (#161's criterion) — ending it there would end it
  before its first line. Guarded by `still follows when the log had nothing past where the run started`.

- [x] **3. A stuck subscribe cannot wedge the command.**
  Until the log's end is known, a `--limit` run holds a deadline on the Core's answer to `subscribe` — 30s, the same
  number and the same argument `harness install` carries, and listed with its reason in `no-prompt-timing.test.ts`
  rather than waved through. It re-sends nothing, cannot make an event arrive sooner, and its expiry is reported as
  this side giving up: what was printed goes out, stderr says the Core stopped answering, and the run exits (`0` if it
  printed anything, `1` if the subscribe answered nothing at all). Disarmed once the end of the log is known, so a
  genuine follow is never cut off by a quiet Core.
  Covered by `prints the history it was given when the subscribe never replays or pushes`, `does not wedge on a
  subscribe that answers nothing at all`, and `stops timing the Core once it has said where its log ends`. The first
  two hang on `HEAD~1`.

- [x] **Existing events-command tests stay green.**
  `packages/cli`: **1190 passed | 3 skipped** across 51 files, including `events-command.test.ts`,
  `events-tail-cursor.test.ts` (a real in-process Core, a real drop, a real restart) and `no-prompt-timing.test.ts`.
  `pnpm typecheck` and `pnpm lint` clean — 0 errors, and none of the 46 pre-existing lint warnings is from these files.
  No existing assertion was changed; the one existing file edited besides the command and its suite is
  `no-prompt-timing.test.ts`, which gained an allowance **row** for the new deadline — the mechanism that rule
  documents for exactly this.

- [x] **Boundaries respected.** The `--kind` filter defect in the same file is #403 and is untouched. Nothing outside
  `packages/cli/src` is touched: no `ProjectFilesPanel.tsx` or `pending-file-drop.ts` (#400), nothing in
  `packages/shared` (#385). Three files changed, all in `packages/cli/src`.

### CI status of the PR head

Reported in the comment below once the checks on `469dcc7` have settled.

qcentic-dev:
### CI status of the PR head

`d611d42` — **all 18 non-skipped checks green**, 0 failures, 0 pending. Run
[33635658769](https://github.com/actana/control/actions/runs/33635658769).

| Check | Result |
| --- | --- |
| Conventions (commitlint on **every** commit, not just the title) | pass 20s |
| Train rules | pass 5s |
| Typecheck | pass 46s |
| Unit Tests | pass 2m36s |
| Lint | pass 38s |
| Secret Scan | pass 25s |
| Dependency Audit | pass 36s |
| E2E — Panel service seam | pass 57s |
| E2E — installer (debian, x64) | pass 1m43s |
| E2E — installer (ubuntu, x64) | pass 1m46s |
| Smoke — Core release tarball (linux-x64 / linux-arm64) | pass 39s / 41s |
| Core image, Panel image (resolve + build/smoke amd64) | pass |
| Promotion gate | pass 5s |
| Resolve PR image mode | pass 5s |

Six train-only jobs are `skipping`, which is what they do on a pull request that is not a promotion.

The second commit on this branch (`d611d42`) narrows the first: whether a `--limit` run found history to read is now
decided by what the **Core sent** rather than by what `--kind` let through. Reading it off the printed count put the
#402 hang one flag away — a filter matching none of the history would have been taken for "there was nothing here", and
the run would have followed. Guarded by `ends a read whose history matched no --kind, rather than following`. The
first-run half of the kind filter is #403 and is still untouched.

Still a draft, and staying one.

mrfqcentic (changes requested):
Reviewed against #402 on a worktree of `d611d42`, diffed against `origin/fix/operator-pins-sessions-drop-cli`. I read the enforcing lines rather than the description, and re-proved the criterion locally against a real in-process Core.

## Acceptance criterion: MET

> Against the pairdemo Core with finish already in SQLite, `events tail --since 13 --limit 30 --json` exits with those events.

Proven, not taken on trust. I built the exact fixture against `startInProcessCore` (real `wss://`, mTLS, real bearer, `PtyCoreLinkServer.handleSubscribe` serving a real log) — 21 events plus `session:finished` as #22, then `events tail --since 13 --limit 30 --json`:

```
head  d611d42 : code=0  ms=35   ids=[14,15,16,17,18,19,20,21,22]   last kind=session:finished
base  a7b5532 : hung — test timed out at 60000ms
```

35 ms is nowhere near the 30 s deadline, so the exit came from the log's end, not from giving up. The three enforcing lines:

- `packages/cli/src/events-command.ts:239` — the history walk is armed on `--limit` + a cursor, so the `--limit` path has something other than the ceiling to end on.
- `packages/cli/src/events-command.ts:302-306` — every event the Core sends past the cursor increments `read`, before the `--kind` filter and before the ceiling. This is the fact the exit is decided on, and `d611d42` moving it off `printed` is the right call: reading it after the filter would have put #402's hang one flag over.
- `packages/cli/src/events-command.ts:329-351` — the marker handler ends the run at the end of the history, and only on a marker that `event-tip.ts` says closed an *empty* tail. Re-using `trackEventTip` rather than trusting one marker is correct, and is what keeps a capped replay from being read as the end.

The required test shape is present: `packages/cli/src/__tests__/events-command.test.ts:251` seeds history past the cursor, lets the subscribe go silent with no marker and no push, and asserts the nine events print and the run exits. Both new files, the whole `packages/cli` suite (1191 passed / 3 skipped across 50 files) and `events-tail-cursor.test.ts` against the real Core are green here; `tsc --noEmit` and `eslint` clean; all 24 CI checks green on `d611d42`. I also re-ran the four new cases against `a7b5532`'s `events-command.ts` — all four fail **by timing out**, so they are not vacuous.

The issue's numbered criterion is met. Its broader remedy — "A stuck subscribe cannot wedge the command" — is **half delivered**, which is finding 1 below.

## Judgement call 1 — the narrowing: correct

A `--limit` run that read history ends at that history; one that found none still follows and stops at n. I checked this is genuinely load-bearing rather than convenient: `events-tail-cursor.test.ts:90` is `--since start --limit 6` against an initially empty log, so an unconditional "exit at end of history" would end that run before its first line and #161's criterion would have to be rewritten. The other three cases in that file either have no cursor (`fromStart` false, unaffected) or hit the ceiling first.

**It does not reintroduce a hang for the pairdemo shape.** Cursor 13, log ending at 22 gives `read = 9`, so the `read === 0` branch at `:344` is not taken — confirmed by the real-Core run above.

The cost, which the PR states and I confirmed empirically: whether `events tail --since 13 --limit 30` terminates now depends on whether anything had been appended past #13 at the moment of the first empty-tail marker. The same command against the same Core can exit or follow depending on that race. The help text at `:126-133` documents both halves honestly, and the ticket's own "what to do" is scoped to "when the events already exist", so this is faithful to #402 — but it is intent inferred from log state. Worth a follow-up ticket to make it explicit (an actual `--follow` / `--no-follow`) rather than leaving `--limit` with two meanings. Not a blocker here.

## Judgement call 2 — the deadline: the row honours the rule, the disarm is real, the arming is too narrow

The allowlist row is legitimate. `events-command.ts` does not match `TOUCHES_A_SESSION` (`/session|prompt|pty|harness-resume/`), it is a shipped module, and the reason clears the 20-char floor — so both guard tests that keep the table from rotting apply to it. The timer body at `:270-276` is `deps.err` + `finish` and nothing else: it re-sends nothing, it cannot make an event arrive sooner, and its expiry is reported as this side giving up. Those are exactly the three properties the file's own argument demands of `--wait-timeout` and `harness install`.

**The disarm is real.** `stopReading()` at `:280-284` nulls `history` and clears `idle`, and `waitForAnswer()` at `:268` early-returns on `history === null`, so once the end of the log is known no later silence can cut a follow off. Tested at `events-command.test.ts:356` (three deadlines of quiet, still following) and, more convincingly, by `events-tail-cursor.test.ts` driving a real Core stop/restart on that path.

What is wrong is *when* it arms and *when* it fires — findings 1 and 2.

## Blocking

**1. `packages/cli/src/events-command.ts:239` — the stuck-subscribe half of #402 is unfixed for `--limit` without a cursor, and the new prose says otherwise.**

The deadline is gated on `limit.value !== null && fromStart`. On a machine with no stored cursor, `actana events tail --limit 30` against a contended Core that never answers `subscribe` takes `fromStart === false`, so `history` is null, `waitForAnswer()` at `:364` returns immediately, and the tip hunt never completes. Probed against the fake Core:

```
events tail --json --limit 30, subscribe never answers
  → still running after 30 simulated minutes: no output, no stderr, no exit
```

That is #402's reported symptom, on the invocation an operator is most likely to type. The header added at `:57-65` states the opposite — "Until the log's end is known, a `--limit` run holds a deadline on its patience; it prints what it did get, says on stderr that the Core stopped answering, and exits" — which is true only on the `fromStart` branch.

This is fixable without going anywhere near #403: arm `waitForAnswer()` whenever `limit.value !== null`, and disarm it where `printing` flips on at `:325` as well as in `stopReading()`. The kind-filter logic #403 owns is untouched by that. If you would rather leave the first-run path to another ticket, then the header prose and the allowlist reason both need narrowing to say so, and the PR description's claim "a stuck subscribe cannot wedge **either of them**" needs to come out.

**2. `packages/cli/src/events-command.ts:270` — a lost socket mid-read truncates the read, and contradicts this file's own reconnect promise.**

`onDisconnected` (`:354`) and `onReady` (`:360`) neither clear nor re-arm `idle`, and while the socket is down no `event` or `eventsReplayed` frame arrives to re-arm it. `DurableCoreClient` backs off to a 5 s ceiling (`packages/sdk/src/durable-core-client.ts:37`) and keeps retrying indefinitely, so it *would* have replayed the whole tail from the cursor — but the deadline fires first. Probed:

```
--since 13 --limit 30 --json, events 14..22 delivered, then the socket drops
  → exit 0, 9 lines on stdout
  → stderr: "link to prod dropped — socket hang up; reconnecting…"
            "prod stopped answering the event subscription; stopping with 9 event(s) printed."
```

A Core restart (`core update`, a systemd restart, a reboot) routinely exceeds 30 s, and `:136` of this same file promises the operator that "a Core restart or a dropped network costs neither a repeated event nor a missed one". For a `--limit` run mid-read that is no longer true, and nothing says so. Clear `idle` in `onDisconnected` and call `waitForAnswer()` from `onReady`.

**3. `packages/cli/src/__tests__/no-prompt-timing.test.ts:24` — the rule now contradicts itself.**

The docstring above the table was updated to "Two rows", but the file header two dozen lines above — which carries the actual argument — still reads:

> `**One file is allowed to hold a timer, and it is named here rather than waved through** (#161, merged into #160's rule).`

and then argues only `harness-command.ts`'s case. In this file the prose *is* the rule; it says so at `:35`, "a table with a reason per row, so it stays a rule". Leaving the header at "one file" makes the widened allowance look unargued to the next reader, which is the failure mode that paragraph exists to prevent. Please extend it to argue the second row the way it argues the first.

## Should fix

**4. `packages/cli/src/events-command.ts:275` — the exit code and the message disagree, and a truncated read reports success.**

`finish(read > 0 ? EXIT_OK : EXIT_FAILURE)` keys off `read`; the message keys off `printed`. With `--kind`, a run that read 9 events, printed none and was then abandoned exits `EXIT_OK` saying "stopping with 0 event(s) printed." — indistinguishable, to a script checking `$?` and stdout, from a complete read that legitimately matched nothing. More generally `actana events tail --since N --limit 1000 --json > events.json` can produce a short file with a success exit. The reason is on stderr, but a scripted consumer reads the code. A deadline expiry with no end-of-log should probably not be `EXIT_OK` at all.

**5. `packages/cli/src/events-command.ts:334` — "carrying on" followed by exiting.**

On `MAX_TIP_ROUNDS` exhaustion `trackEventTip` writes "carrying on from #N, which may be behind its log's end." to stderr and returns a **non-null** tip (`event-tip.ts:124-133`). On the new printing branch a non-null return means `stopReading()` then `finish(EXIT_OK)` — so the operator is told the run is continuing while it has already terminated 0 on a knowingly incomplete read. Reachable against a fast-appending Core with a wide `--limit`. That message was written for the caller that really does carry on; this caller needs its own.

## Non-blocking notes

- `events-command.ts:229` — on a `fromStart` run `tip` is allocated but never fed or consulted (`printing` starts true and is only ever set true); on a non-`fromStart` run `history` is null. The two trackers are mutually exclusive by construction, and a reader has to prove that before trusting it. One tracker with a mode would read easier.
- No case, fake or real, exercises the new **printing-side** capped-tail walk — a stored cursor with more than `EVENT_TAIL_LIMIT` (1000) events past it and `--limit 2000`. The `!printing` side has exactly that against a real Core ("prints none of a history longer than the Core replays in one tail"), and this is the side that now re-subscribes *while printing*, which is the duplicate/gap shape #205's review blocked on. The dedup is the durable client's and I believe it holds, but it is unproven on the new path.

## Checked and cleared

- A live event racing the empty replay marker (which would make `read > 0` and end a follow after one event) is not reachable: `PtyCoreLinkServer.handleSubscribe` is synchronous, so `eventsReplayed` is always on the wire ahead of any later `pushLiveEvents` frame.
- Stale `inTail` across a dropped marker only ever inflates `delivered`, so it can cost an extra round trip but can never end a read early.
- No timer leaks: `finish` and `stopReading` both clear `idle`, and every `--limit` path either disarms on an empty first marker or hits the ceiling.
- No existing test changes outcome; the `--limit` cases in `events-tail-cursor.test.ts` without a cursor take the unchanged `fromStart === false` path.

## Boundaries — all clean

- `--kind` first-run defect (#403) untouched: `history` is armed only when `fromStart`, so the no-cursor path is unchanged. Probed to be sure — `events tail --kind session:finished --limit 1` with the finish already in the log still never exits, which is #403 still being #403.
- Diff is three files, all in `packages/cli`. `ProjectFilesPanel.tsx`, `pending-file-drop.ts`, `harness-hook-pipeline.ts`, `subagent-activity.ts`: none touched — and the last two are exactly what sibling PR #438 owns on this same feature branch, so the file lock held.
- Base is `fix/operator-pins-sessions-drop-cli`, not `main` and not `beta/0.4.5`. Still a draft.
- Both commits pass commitlint: `fix` type, headers 52 and 83 chars against a 120 limit, longest body line 125 against 132, blank line before `Refs #402`. `Conventions` is green.

Requesting changes on 1–3. The mechanism is the right one and the criterion is met; what is left is the half of the stuck-subscribe case that did not get covered, the reconnect interaction, and a rule file that now argues against itself.

qcentic-dev:
Thanks — the probes are what made this actionable, particularly the one that ran `--limit 30` with no cursor for thirty
simulated minutes. All five findings are addressed in `bc7f514`, and both non-blocking notes are taken.

## Blocking

**1 — the deadline was armed only on `--limit` *plus* a cursor.** Fixed as suggested, and you were right that it is
orthogonal to #403. `waitForAnswer` is now gated on a `timing` flag initialised from `limit.value !== null` alone, and
the `!printing` side re-arms it on every event and every capped-tail marker. It is disarmed in two places: `stopTiming()`
where `printing` flips on — the tip is the answer that side was waiting for — and `stopReading()` where a read ends. The
`--kind` filter logic is untouched; `history` is still armed only when `fromStart`, so #403's path is exactly as it was.

New case: `does not wedge a --limit run that has no cursor and no --since either`. It hangs on `d611d42` (5 s test
timeout) and exits non-zero here. `keeps timing a tip hunt the Core is still answering, and stops at the tip` is the
other half — three rounds of near-deadline silence each ended by an answer, then an hour of quiet after the tip is
found, still following.

**2 — a lost socket truncated the read.** `idle` is cleared in `onDisconnected` and `waitForAnswer()` is called from
`onReady`, so the clock does not run while the link is down and restarts from the reconnect rather than from whatever
the dead socket last said. The reasoning is in the handler, naming the promise at `:136` it exists to keep.

New case: `stops the clock while the link is down, and restarts it on the reconnect` — an event, a drop, four
deadlines' worth of restart, then a reconnect that replays the rest, asserting ids 14–22 and exit 0. It fails on
`d611d42`.

**3 — the header argued one row.** Rewritten. It now argues both rows separately and, before either, states the three
properties a listed timer has to have — re-sends nothing, cannot make the thing happen sooner, expires as this side
giving up — so the next row is measured against something rather than compared to a precedent. The `events-command.ts`
paragraph also says what it is *not*: a read of a log, with nothing on the path from an operator's text to a Harness's
stdin, and still subject to `TOUCHES_A_SESSION` if that ever changed. The closing paragraph now says out loud that
growing the table is meant to cost a paragraph.

## Should fix

**4 — the code and the message disagreed.** Taken further than "make them agree", along the lines of your last sentence:
a deadline expiry is no longer `EXIT_OK` under any circumstances. It exits non-zero and says `giving up with N event(s)
printed. This read did not finish.` Everything it did get stays on stdout — a short file with a non-zero status is
something a script can act on; a short file with a zero is not. `read` no longer participates in the exit code at all,
so the `--kind` shape you describe (9 read, 0 printed, abandoned) now exits 1.

The existing case was flipped to assert `EXIT_FAILURE` and the `did not finish` line, and it fails on `d611d42`.

**5 — "carrying on" followed by exiting.** `trackEventTip` grows an optional `onRoundsExhausted(seen)` that is called
**instead of** the default stderr line. `harness install` and the first-run tip hunt keep the existing sentence, because
for them it is true. The read path passes a handler that records the number as an approximation and, when the marker
comes back non-null, prints its own line and exits non-zero — an approximation is not a finished read.

New case: `does not report a read finished when the walk settled for an approximation`. 201 non-empty rounds against
`MAX_TIP_ROUNDS = 200`, asserting exit 1, the 201 events still on stdout, and that `carrying on` is *not* in stderr.

## Non-blocking

**Taken — the printing-side capped-tail walk, against a real Core.** `events-tail-cursor.test.ts` gains
`reads a history longer than one tail exactly once, in order, and then exits`: 1500 events past the cursor, `--since
start --limit 2000`, asserting `[1..1500]` exactly — a repeat is a replay storm, a gap is a moved cursor — plus
`log.tailReads > 1` so the cap was genuinely hit. It passes, so the durable client's dedup does hold on the path that
re-subscribes while printing. Agreed this was the right thing to be uneasy about; it is the #205 shape on a path that
did not exist when #205 was reviewed.

**Partly taken — the two trackers.** They are still two objects, and after finding 5 they have a reason to be: they now
differ in exactly one place, the sentence at the end of the hunt, and that difference is the whole of finding 5. What is
fixed is the reading cost — the comment where they are created states the invariant and its proof (`printing` is
initialised from `fromStart` and only ever switched *on*, so a run that starts printing never enters the `!printing`
branch, and `history` is non-null only when `fromStart`), so a reader no longer has to derive it. If you would still
rather see one tracker with a mode, say so and I will do it.

## Follow-up

Filed as **#441** — an explicit `--follow` / `--no-follow` so `--limit` stops carrying two meanings, with your
"can exit or follow depending on that race" quoted as the motivation, and the two open decisions (the default, and the
exit code when a `--no-follow` read is simply shorter than `n`) written down. Linked from the PR body, which is also
corrected: the "either of them" claim you flagged is now "any of them", and it describes the exit code honestly.

## State

`packages/cli`: **1196 passed | 3 skipped** across 51 files, including `events-tail-cursor.test.ts` and
`no-prompt-timing.test.ts`. `pnpm typecheck` and `pnpm lint` clean. **All 18 CI checks green on `bc7f514`** (run
[33639295240](https://github.com/actana/control/actions/runs/33639295240)).

Four of the five new cases fail on `d611d42` — findings 1, 2, 4 and 5 each have one, and finding 1's fails by hanging.

Still a draft, base unchanged, nothing merged, no force push. The base branch moved to `a74cc27` with #437; the diffs
are file-disjoint, so nothing was merged in.

mrfqcentic (approved):
Round 2, against #402 on a fresh worktree of `bc7f514`, diffed against `origin/fix/operator-pins-sessions-drop-cli` at `5d33f0c` (which now carries both #437 and #438). I read the enforcing lines rather than the reply, re-proved the acceptance criterion against a real in-process Core, and probed each of the five findings — including one by mutation.

**All five findings are resolved, both non-blocking notes are taken, and the partial take on the trackers is accepted. Approving.**

## Acceptance criterion: still MET, re-proven at `bc7f514`

> Against the pairdemo Core with finish already in SQLite, `events tail --since 13 --limit 30 --json` exits with those events.

Rebuilt the fixture against `startInProcessCore` (real `wss://`, mTLS, real bearer, `PtyCoreLinkServer.handleSubscribe` serving a real log) — 21 events plus `session:finished` as #22, then the command verbatim:

```
bc7f514 : code=0  ms=26  ids=[14,15,16,17,18,19,20,21,22]  last kind=session:finished
```

26 ms against a 30 s deadline, so the exit came from the end of the log and not from giving up. The round-2 changes did not disturb it.

## The five findings

**1 — the deadline armed only on `--limit` *plus* a cursor. Resolved.**

`events-command.ts:284` is now `let timing = limit.value !== null;` — `fromStart` is out of it. It is re-armed on both sides of the printing switch (`:350` on a tip-hunt event, `:378` on a capped tail-hunt marker, `:359` on a read event, `:397` on a capped read marker), and disarmed in exactly the two places asked for: `stopTiming()` at `:384` where `printing` flips on, and `stopReading()` at `:334` where a read ends. `waitForAnswer` at `:314` gates on `timing` rather than on `history`, which is what closes the no-cursor hole.

`does not wedge a --limit run that has no cursor and no --since either` fails on `d611d42` **by timing out** — the hang, reproduced — and exits `EXIT_FAILURE` here.

The #403 boundary holds, and I probed it rather than reasoned it: `events tail --kind session:finished --limit 1` against a real Core with the finish already in the log is **still running after 45 s** at `bc7f514`. The tip hunt gets its answer, `stopTiming()` runs, and the follow waits — #403 is still #403, and the new deadline does not accidentally paper over it.

**2 — a lost socket truncated the read. Resolved, and proven against a real Core.**

`idle` is cleared at `:439` in `onDisconnected` and `waitForAnswer()` is called at `:452` from `onReady`, each with the reasoning inline and naming the reconnect promise at `:161` it exists to keep.

The fake-Core case is there and fails on `d611d42`. I also ran the real thing, since a `core update` is the reported shape: a `--limit 6000` read of a 5 000-event history, the Core **actually stopped** mid-walk after 1 000 events were printed, 35 s of real wall clock with nothing listening, then a second Core started on the same port with the same material:

```
stopped the Core with 1000 printed
down for 35003ms; still 1000 printed
code=0 count=5000 first=1 last=5000
```

Every event exactly once, in order, exit 0 — through an outage longer than the deadline. That is the promise at `:161` kept.

**3 — the rule file argued one row while the table held two. Resolved.**

`no-prompt-timing.test.ts:24` now opens "**Two files are allowed to hold a timer, and both are named here**", states the three properties a listed timer must have *before* either row — re-sends nothing, cannot make the thing happen sooner, expires as this side giving up — and then argues `harness-command.ts` and `events-command.ts` separately against them. The `events-command.ts` paragraph also says what it is not (a read of a log, nothing on the path from an operator's text to a Harness's stdin, still subject to `TOUCHES_A_SESSION`), and the close now says out loud that growing the table costs a paragraph. Both guard tests still apply to the new row: it is a shipped module, its reason clears the length floor, and it does not match `TOUCHES_A_SESSION`.

**4 — the exit code and the message disagreed. Resolved, and taken further.**

`:322` is `finish(EXIT_FAILURE)` unconditionally; `read` is out of the exit code entirely. The message is now `giving up with N event(s) printed. This read did not finish.` The `--kind` shape I described — 9 read, 0 printed, abandoned — exits 1. Everything that arrived stays on stdout, which is the right trade: a short file plus a non-zero status is actionable, a short file plus a zero is not. The existing case was flipped and fails on `d611d42`.

**5 — "carrying on" followed by exiting. Resolved.**

`event-tip.ts:106` adds `onRoundsExhausted`, called at `:145` **instead of** the default sentence. `harness install` (`harness-command.ts:402`) and the first-run tip hunt (`events-command.ts:266`) pass no handler and keep it, correctly — for them it is true. The read path sets `approximate` and, at `:407`, prints its own line and exits `EXIT_FAILURE`. `does not report a read finished when the walk settled for an approximation` drives 201 non-empty rounds against `MAX_TIP_ROUNDS = 200`, asserts exit 1, the 201 events still on stdout, and that `carrying on` is *absent*. It fails on `d611d42`.

## The non-blocking notes

**Note A — the printing-side capped-tail walk. Taken, and it does what it claims.** `events-tail-cursor.test.ts` gains a real-Core case: 1 500 events past the cursor against a cap of 1 000, `--since start --limit 2000`, asserting `[1..1500]` exactly — a repeat would be a replay storm, a gap a moved cursor — plus `log.tailReads > 1`. Green here in 306 ms. That closes the #205 shape on the path that re-subscribes *while printing*, which is the one thing I was uneasy about.

**Note B — the two trackers. Accepted as they are; no further ask.** The argument in the reply is the right one and it changed my mind: after finding 5 they are no longer the same object with different bookkeeping, they differ in one behaviour — the sentence at the end of the hunt, which *is* finding 5. A mode flag would re-merge two things that now genuinely differ. The invariant is spelled out with its proof where they are created (`:257-269`), which is what the note was actually about, and I re-checked it holds: `printing` is initialised from `fromStart` and only ever switched on, `history` is non-null only when `fromStart`, so the `!printing` branch is unreachable for a run that has a `history`. Leave it.

## Verified round-2 state

- `packages/cli`: **1196 passed | 3 skipped** across 51 files, run in a scratch copy — including `events-tail-cursor.test.ts` against the real Core.
- `pnpm typecheck` clean; `pnpm lint` 0 errors (46 pre-existing warnings, none in the touched files).
- Four of the five new cases fail on `d611d42`, finding 1's by hanging. Confirmed independently, not taken from the reply.
- All 24 CI checks on `bc7f514`: **18 success, 6 skipped, 0 failed**.
- `commitlint` run locally over all three commits and the PR title with this repo's own config: **0 problems, 0 warnings**. Headers 52 / 83 / 97 chars; longest body lines 125 / 121 / 122, all inside 132.
- Still a **draft**; base is still `fix/operator-pins-sessions-drop-cli`.
- Merges cleanly onto `5d33f0c` (`git merge-tree` resolves with no conflict), and the file sets are disjoint: this PR touches five files all under `packages/cli`; #437 and #438 touched `ProjectFilesPanel.tsx`, `subagent-activity.ts`, `harness-hook-pipeline.ts` and their tests plus one doc. The file lock held on both sides.

## New non-blocking notes — none of them blocking, none of them for this PR to fix now

**a. `packages/cli/src/__tests__/cli-harness.ts:493` — the reconnect re-arm has zero coverage, and the test named for it does not cover it.**

`fakeCore`'s client is `onReady: () => () => {},` — it registers no listener and there is no `emitReady` counterpart to `emitDisconnected`. So `stops the clock while the link is down, and restarts it on the reconnect` (`events-command.test.ts:361`) only proves the first half: after the drop the clock is *cleared*, and what re-arms it in that test is the first event of the replay, not `onReady`.

Proved by mutation rather than by reading: with `waitForAnswer()` deleted from `onReady` (`events-command.ts:452`), the whole package is still **1196 passed | 3 skipped** — identical. And the branch is load-bearing, not dead code: adding an `emitReady` to the fake and driving an event, a drop, a reconnect, then silence gives

```
bc7f514             : exits 1, "This read did not finish."
without the re-arm  : never ends — the run is un-deadlined for the rest of its life
```

which is #402's own symptom surviving a reconnect. The fix is a few lines in the fake plus that one case. Worth doing when the file is next open; not worth another round.

**b. A `--limit` ceiling can advance the stored cursor past what it printed — pre-existing, and not this PR's.** Raised by a machine pass, and it reproduces against a real Core:

```
first run  : code=0 out=1  cursor=1
second run : --limit 5 → ids=[2,3,4,5,6]  storedCursor=101
third run  : nothing past #101 — events 7..101 are gone from the follow-along stream
```

`finish` closes the client, but `DurableCoreClient.deliverEvent` has already advanced *and persisted* the cursor for the rest of the synchronous burst. **I ran the identical probe against `a7b5532` and got byte-identical numbers**, so this PR neither introduces nor worsens it — the ceiling `finish` at `:364` is unchanged from the base. It does deserve its own issue, and this PR's "`--limit` is a ceiling, not a quota" framing makes the bounded read the advertised idiom, so it is more visible now than it was.

**c. `--kind` with `--limit` can exit 0 having printed nothing.** With a stored cursor and unrelated history past it, `events tail --limit 1 --kind session:finished` reads, matches nothing, and exits `EXIT_OK` silently. This is `d611d42`'s deliberate decision, it is the one that keeps #402 from moving one flag over, and I endorsed it in round 1 — but the help text's `--limit  print at most n events, then exit` does not warn that the run can end at zero. A line in `EVENTS_HELP`, or a note on #441 where the `--no-follow` exit codes are already an open decision. (The shipped skill text at `orchestration-skill-payload.ts` uses plain `actana events tail --json`, no `--kind` and no `--limit`, so nothing in-tree relies on the old behaviour.)

**d. `rounds` in the `history` tracker survives a link flap.** `onDisconnected` clears the timer but not the tracker's counters, so a link that flaps enough times during one read could reach `MAX_TIP_ROUNDS` and exit with "appended events faster than they could be read" — a statement about the Core for what is a network problem. It takes ~200 flaps against a 5 s backoff ceiling, and the exit code stays correct either way. Cosmetic.

**e. The PR body has stale pre-round-2 prose.** "`packages/cli/src/events-command.ts`, and nothing else in the tree" (`event-tip.ts` changed too) and "1190 passed | 3 skipped" (1196 now). The round-2 section below it is accurate; only the older paragraphs drifted.

## Verdict

Approved. Every blocking finding is fixed at the enforcing line and guarded by a case that fails without it, the two should-fixes went further than "make them agree", both notes were taken, and the reconnect fix survives a real 35-second Core outage. #403 is still #403, the file lock held against both merged siblings, and the follow-up the design cost deserved is filed as #441.

Issue #380, opened 2026-09-02
B1: repeating a pin click deselects instead of focusing


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `B1`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: panel-rendering-investigation-2026-09-02.md (symptom B)

## What the operator sees

Rapid clicks on pinned items bounce. A second click on the already-active pin closes the panel instead of re-focusing.

## Where

`packages/panel/src/lib/terminal-store.tsx` `nextActiveTaskId` (catalog investigation: 370–377) returns `null` when `currentTaskId === requestedTaskId` and the session is materialized.

## What to do

Pin clicks are navigation, not a toggle. Repeat-click must focus. Deselect stays on the session panel close button / terminal.close hotkey (`projects.$id.tsx` already documents that for cards).

## Acceptance

1. Clicking the already-selected pin leaves it selected and focuses the session.
2. Rapid A→B→A follows the last click; the panel does not close on the repeated A.

qcentic-dev:
## B1 is not closed by #443 — the Panel UI still needs to be clicked

Heads-up from implementing this ticket. [#443](https://github.com/actana/control/pull/443) started out claiming it fixed this symptom; review traced the call sites and showed it cannot have. That PR is now re-framed as a hardening refactor and carries `Refs #380`, not a closing keyword, so this issue stays open on purpose.

### Why the filed cause was not the cause

This issue was filed from a code read ("Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)") and landed on `nextActiveTaskId` in `packages/panel/src/lib/terminal-store.tsx`, which returned `null` — a cleared scope, which is the panel close — when the requested task was already active and its session was materialized.

That branch had exactly one production caller, `toggle`, and all six of its call sites guard the combination away:

- `projects.$id.tsx:558` — the effect only runs when the scope's active id is `null`.
- `projects.$id.tsx:635` — inside `active?.taskId !== task.id`, with the equal case routed to `rehydrate` one line above.
[...]

Issue #393, opened 2026-09-02
T2: two tabs can both type until drive/lock answers; failed reattach stacks onData


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `T2`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Open the same session in two tabs. Both are writable until the server returns following/locked. Reload can also stack xterm `onData` handlers after a failed reattach.

## Where

`session-write-store.ts` (catalog: 47–51) `drive none = writable`. `TerminalPane.tsx` 1212–1287 `wireTerminalInput` never disposed. File lock with A-enter.

## What to do

Unknown drive must not mean two tabs type. Dispose `onData` on reattach failure. Related: #187 (lock lag / stays running) is the older two-client lock bug — do not duplicate its scope; this hunt is tab drive + handler leak.

## Acceptance

1. Two tabs on one session: at most one is writable once drive/lock is known, and the optimistic window is short and visible.
2. Failed reattach does not duplicate keystrokes.

## Related

#187

mrfqcentic:
Delivered in **0.4.5**. Fixed by #458, merged to `beta/0.4.5` via #481.

Issue #386, opened 2026-09-02
A-enter: Enter after settlement posts running again for hooks that omit turn-start


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `A-enter`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: panel-rendering-investigation-2026-09-02.md (A secondary)

## What the operator sees

For Cursor CLI (always) and Codex until hooks are reviewed, any terminal input containing CR/LF after the task has finished posts `running` to the Core. A paste or drop-path write is enough.

## Where

`TerminalPane.tsx` (investigation: 1236–1266) + `task-status-sync.ts` (30–44). `fallbackRunningPosted` re-arms whenever the task is not currently running (`shouldResetTerminalRunningFallback`).

## What to do

Gate the fallback on a fresh turn-start signal, not merely “task is not running”. A stale keystroke after settlement must not resurrect `running`. Sequence after A-heal if both land in one wave (`TerminalPane.tsx` is also T2).

## Acceptance

1. After a session is finished, pressing Enter (or pasting a path) does not PATCH it to running.
2. Harnesses that never report turn start still go running when the operator actually starts a turn.

## Related

Blocked by / pair with A-heal. Same user-visible flip.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #449, merged to `beta/0.4.5` via #481.

Issue #382, opened 2026-09-02
S3: pin drag / Shift+Arrow snaps back when any Core pin is on the rail


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `S3`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Drag two pins, or drag a Core pin into another group. Animation plays, then order/group reverts with an error toast.

## Where

`ProjectBar.tsx` persist via Panel `reorderPinnedProjects` (catalog: 274–304 / 1118). Core pins must go through `mutateProjectForCore` / presentation API.

## What to do

Reorder and regroup through the owner of each pin. Panel-local pins keep the Panel reorder API; Core pins use the Core presentation/mutate path. Mixed rails cannot send Core ids to the Panel-only API.

## Acceptance

1. Dragging two pins, including a Core pin, keeps the new order after reload.
2. No error toast on a valid mixed-rail reorder.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #476, merged to `beta/0.4.5` via #481. The PR carried `Refs` rather than a closing keyword. Remaining rail-slot work is filed as #478.

PR #443, opened 2026-09-02
fix(panel): a repeat pin selection can never clear the scope

qcentic-dev:
> **Re-framed after review.** The first version of this PR claimed it cured the operator symptom in #380 (B1). The review traced every call site and showed the `null` branch it removes is not reachable through today's callers, so that claim was wrong. This is a hardening refactor plus a shared reducer. **#380 stays open** — see [the note posted there](https://github.com/actana/control/issues/380).

## What this is

`nextActiveTaskId` in `packages/panel/src/lib/terminal-store.tsx` returned `null` when the requested task was already the scope's active one and its session was materialized. A `null` scope value **is** the panel close — it is what `deselect`, `close` and `closeForProject` write. Having a *selection* path able to produce it is a trap.

It is replaced by `nextActiveByProject(activeByProject, scopeKey, requestedTaskId)`, which always lands on the requested task and returns the map identity when the request is a no-op, so a repeat selection cannot clear a scope and forces no re-render. The five hand-written copies of that same selection write — `toggle`, `openSession`, `setActiveSession`, the grid's focus-in write, and now the reducer itself — collapse onto one tested function.

## What this is not

It is not a fix for B1's symptom, and the review is right that the original description said otherwise. The `null` branch had one production caller (`toggle`), and every one of its six call sites guards away the combination that produced it:

| Call site | Guard |
| --- | --- |
| `projects.$id.tsx:558` | effect returns unless the scope's active id is `null` |
| `projects.$id.tsx:635` | inside `active?.taskId !== task.id`; the equal case is routed to `rehydrate` on the line above |
| `projects.$id.tsx:802` | a freshly minted optimistic task id, never already active |
| `projects.$id.tsx:1081` | guarded by `if (!currentId)` |
| `projects.$id.tsx:1092` | returns early on `nextTask.id === currentId` |
| `projects.$id.tsx:1244` | only runs once the earlier branch established there is no active session |

And the pin/card click never reaches `toggle` at all: `onToggle: selectTerminal` (`projects.$id.tsx:1704`) → `selectTerminal` (`:1375`) → `terminals.openSession`, whose selection write already read `prev[scopeKey] === task.id ? prev : {...}` before this PR. `projects.$id.tsx:558` is a sixth call site the review's list did not include; it is guarded too.

So: this removes the trap before a seventh caller arrives without the guard. B1 needs the Panel UI actually clicked on the base branch before it can be closed.

### One thing worth recording

`activeFor` and `activeTaskIdFor` (`terminal-store.tsx:1143-1167`) close over the render's `activeByProject`, not a ref. `openRequestedSession` (`projects.$id.tsx:592-640`) reads both **after** `await api.getTask(...)`, while `toggle`'s updater reads the committed `prev`. A selection committed during that await would leave the guard reading a stale snapshot — the one shape in which the old `null` branch could have fired. That is a race on the notification-open path, not a repeat click, so it is not B1, and it is now moot. Naming it here so the "unreachable" claim is precise rather than absolute. No test is added for it: exercising it needs `projects.$id.tsx`, which parallel tickets own and this PR must not edit, and there is no existing harness that renders `TerminalProvider`.

## Review findings addressed

- **Blocking 1 — the `Ctx.toggle` contract doc.** `terminal-store.tsx:68` promised "select if not active, deselect (hide panel) if already active", which the code no longer does. Rewritten to state the real rule, and it now calls out that `toggle` is a misnomer kept on purpose because renaming it would touch `projects.$id.tsx`.
- **Blocking 2 — delivery.** Resolution taken: **the second option, re-frame honestly.** Title, body, and the new commit message all say the branch was unreachable through current callers. No closing keyword anywhere — `Refs #380`, never `Closes`, so the merge cannot close the issue. A note is posted on #380.
- **Non-blocking — duplicate test.** "re-selects a task whose session is already materialized" dropped; its comment folded onto the repeat-request test above it.
- **Non-blocking — burst test name.** Renamed to "follows the last request across a rapid repeat-A -> B -> A burst", which is what it seeds.

The `nextActiveByProject` docstring also lost its unsupported causal story and now records that the branch was unreachable, plus every writer that legitimately clears a scope.

## Mechanics

Two commits; the first is left as pushed rather than amended, because rewriting it needs a force push. The second states the correction explicitly and supersedes the first's causal claim.

Tests: 22 in `terminal-store.test.ts`, full panel suite 146 files / 1565 tests green (1566 before, minus the dropped duplicate). `tsc --noEmit` and `eslint` clean.

Still a draft. Base unchanged: `fix/operator-pins-sessions-drop-cli`.

Refs #380

https://claude.ai/code/session_01SGZcdyMnRX7cTYPyareJTT

### DIFF
--- a/packages/panel/src/lib/__tests__/terminal-store.test.ts
+++ b/packages/panel/src/lib/__tests__/terminal-store.test.ts
@@ -3,7 +3,7 @@ import type { Task } from "~/db/schema";
 import {
   archivedSessionsEligibleForReap,
   commandForTask,
-  nextActiveTaskId,
+  nextActiveByProject,
   resolveActiveTaskIdForProject,
   type OpenTerminal,
 } from "../terminal-store";
@@ -130,17 +130,58 @@ describe("commandForTask", () => {
   });
 });
 
-describe("nextActiveTaskId", () => {
-  it("keeps a stale persisted active task open when no session is materialized", () => {
-    expect(nextActiveTaskId("task-1", "task-1", false)).toBe("task-1");
-  });
+describe("nextActiveByProject", () => {
+  const scope = "project-1";
 
-  it("hides a task that is already active and materialized", () => {
-    expect(nextActiveTaskId("task-1", "task-1", true)).toBeNull();
+  it("selects a task in a scope that had none", () => {
+    expect(nextActiveByProject({}, scope, "task-1")).toEqual({ [scope]: "task-1" });
   });
 
   it("switches active tasks", () => {
-    expect(nextActiveTaskId("task-1", "task-2", true)).toBe("task-2");
+    expect(nextActiveByProject({ [scope]: "task-1" }, scope, "task-2")).toEqual({
+      [scope]: "task-2",
+    });
+  });
+
+  // #380 AC1: the pin click is navigation, not a toggle. A repeat request for
+  // the already-active task must leave it selected — returning null here is
+  // what deselected the scope and closed the panel under the operator.
+  it("keeps the task active when it is requested again", () => {
+    expect(nextActiveByProject({ [scope]: "task-1" }, scope, "task-1")).toEqual({
+      [scope]: "task-1",
+    });
+  });
+
+  it("returns the same map object for a repeat request so no re-render is forced", () => {
+    const prev = { [scope]: "task-1" };
+    expect(nextActiveByProject(prev, scope, "task-1")).toBe(prev);
+  });
+
+  it("re-selects a task whose session is already materialized", () => {
+    // The old signature took a `hasMaterializedSession` flag and deselected
+    // when it was true; materialization is no longer part of the decision.
+    const prev = { [scope]: "task-1" };
+    expect(nextActiveByProject(prev, scope, "task-1")[scope]).toBe("task-1");
+  });
+
+  // #380 AC2: a rapid A -> B -> A burst follows the last click and never
+  // passes through a null (panel-closing) selection.
+  it("follows the last click across a rapid A -> B -> A burst", () => {
+    const seen: (string | null)[] = [];
+    let state: Record<string, string | null> = { [scope]: "task-a" };
+    for (const requested of ["task-a", "task-b", "task-a"]) {
+      state = nextActiveByProject(state, scope, requested);
+      seen.push(state[scope] ?? null);
+    }
+    expect(seen).toEqual(["task-a", "task-b", "task-a"]);
+    expect(seen).not.toContain(null);
+    expect(state[scope]).toBe("task-a");
+  });
+
+  it("leaves other scopes untouched", () => {
+    expect(
+      nextActiveByProject({ "project-2": "task-9" }, scope, "task-1"),
+    ).toEqual({ "project-2": "task-9", [scope]: "task-1" });
   });
 });
 
--- a/packages/panel/src/lib/__tests__/terminal-store.test.ts
+++ b/packages/panel/src/lib/__tests__/terminal-store.test.ts
@@ -143,9 +143,10 @@ describe("nextActiveByProject", () => {
     });
   });
 
-  // #380 AC1: the pin click is navigation, not a toggle. A repeat request for
-  // the already-active task must leave it selected — returning null here is
-  // what deselected the scope and closed the panel under the operator.
+  // Selection is navigation, not a toggle: a repeat request for the
+  // already-active task leaves it selected. The old `nextActiveTaskId` returned
+  // null here whenever a session was materialized, and a null scope is the panel
+  // close. Materialization is no longer an input to the decision at all.
   it("keeps the task active when it is requested again", () => {
     expect(nextActiveByProject({ [scope]: "task-1" }, scope, "task-1")).toEqual({
       [scope]: "task-1",
@@ -157,16 +158,10 @@ describe("nextActiveByProject", () => {
     expect(nextActiveByProject(prev, scope, "task-1")).toBe(prev);
   });
 
-  it("re-selects a task whose session is already materialized", () => {
-    // The old signature took a `hasMaterializedSession` flag and deselected
-    // when it was true; materialization is no longer part of the decision.
-    const prev = { [scope]: "task-1" };
-    expect(nextActiveByProject(prev, scope, "task-1")[scope]).toBe("task-1");
-  });
-
-  // #380 AC2: a rapid A -> B -> A burst follows the last click and never
-  // passes through a null (panel-closing) selection.
-  it("follows the last click across a rapid A -> B -> A burst", () => {
+  // A rapid burst follows the last request and never passes through a null
+  // (panel-closing) selection. Seeded on A and requesting A first, so this is a
+  // repeat-A -> B -> A: a superset of the plain A -> B -> A burst.
+  it("follows the last request across a rapid repeat-A -> B -> A burst", () => {
     const seen: (string | null)[] = [];
     let state: Record<string, string | null> = { [scope]: "task-a" };
     for (const requested of ["task-a", "task-b", "task-a"]) {
--- a/packages/panel/src/lib/terminal-store.tsx
+++ b/packages/panel/src/lib/terminal-store.tsx
@@ -367,14 +367,31 @@ function isRemoteTask(taskId: string): boolean {
   return remoteTaskIds.has(taskId);
 }
 
-export function nextActiveTaskId(
-  currentTaskId: string | null,
-  requestedTaskId: string,
-  hasMaterializedSession: boolean
-): string | null {
-  return currentTaskId === requestedTaskId && hasMaterializedSession
-    ? null
-    : requestedTaskId;
+/**
+ * The scope-to-active-task map after a request to select `requestedTaskId`.
+ *
+ * Selecting a session — a pin in the sidebar, a card in the list — is
+ * navigation, not a toggle: the requested task always ends up active. A repeat
+ * request for the already-active task therefore keeps it selected (the caller
+ * re-focuses its terminal) instead of clearing the scope, and a rapid
+ * A -> B -> A burst lands on A. Clearing the selection here on a repeat click
+ * was what closed the panel under the operator and made those bursts bounce.
+ *
+ * Deselect is its own gesture with its own path: the session panel's close
+ * button and the `terminal.close` hotkey both call `deselect`. `projects.$id.tsx`
+ * documents the same rule for card clicks.
+ *
+ * Returns `activeByProject` unchanged when the request is a no-op, so callers
+ * can hand the result straight to `setState` without forcing a re-render.
+ */
+export function nextActiveByProject(
+  activeByProject: Record<string, string | null>,
+  scopeKey: string,
+  requestedTaskId: string
+): Record<string, string | null> {
+  return activeByProject[scopeKey] === requestedTaskId
+    ? activeByProject
+    : { ...activeByProject, [scopeKey]: requestedTaskId };
 }
 
 /** Grace period before an un-selected archived session's PTY is reaped. */
@@ -583,9 +600,7 @@ export function TerminalProvider({ children }: { children: ReactNode }) {
     const session = sessionsRef.current.find((s) => s.taskId === taskId);
     if (!session) return;
     const scopeKey = scopeKeyForProject(session.project);
-    setActiveByProject((prev) =>
-      prev[scopeKey] === taskId ? prev : { ...prev, [scopeKey]: taskId },
-    );
+    setActiveByProject((prev) => nextActiveByProject(prev, scopeKey, taskId));
   }, []);
   const getGridFocusedTaskId = useCallback(() => gridFocusedTaskIdRef.current, []);
   // Pending "New row" request: the grid drops the next new session into a fresh
@@ -666,9 +681,6 @@ export function TerminalProvider({ children }: { children: ReactNode }) {
       opts?: { awaitCreate?: boolean; coreId?: string | null },
     ) => {
       const scopeKey = scopeKeyForProject(project);
-      const hadSession = sessionsRef.current.some(
-        (p) => p.taskId === task.id && scopeKeyForProject(p.project) === scopeKey
-      );
       setSessions((prev) => {
         const existing = prev.find(
           (p) => p.taskId === task.id && scopeKeyForProject(p.project) === scopeKey
@@ -700,11 +712,7 @@ export function TerminalProvider({ children }: { children: ReactNode }) {
         };
         return [...prev, next];
       });
-      setActiveByProject((prev) => {
-        const curr = prev[scopeKey] ?? null;
-        const next = nextActiveTaskId(curr, task.id, hadSession);
-        return curr === next ? prev : { ...prev, [scopeKey]: next };
-      });
+      setActiveByProject((prev) => nextActiveByProject(prev, scopeKey, task.id));
     },
     []
   );
@@ -755,9 +763,7 @@ export function TerminalProvider({ children }: { children: ReactNode }) {
           },
         ];
       });
-      setActiveByProject((prev) =>
-        prev[scopeKey] === task.id ? prev : { ...prev, [scopeKey]: task.id }
-      );
+      setActiveByProject((prev) => nextActiveByProject(prev, scopeKey, task.id));
     },
     []
   );
@@ -830,9 +836,7 @@ export function TerminalProvider({ children }: { children: ReactNode }) {
 
   const setActiveSession = useCallback((project: ScopedProject, taskId: string) => {
     const scopeKey = scopeKeyForProject(project);
-    setActiveByProject((prev) =>
-      prev[scopeKey] === taskId ? prev : { ...prev, [scopeKey]: taskId }
-    );
+    setActiveByProject((prev) => nextActiveByProject(prev, scopeKey, taskId));
   }, []);
 
   const adoptTaskId = useCallback((fromTaskId: string, task: Task) => {
--- a/packages/panel/src/lib/terminal-store.tsx
+++ b/packages/panel/src/lib/terminal-store.tsx
@@ -65,7 +65,15 @@ type Ctx = {
   activeFor: (projectId: string) => OpenTerminal | null;
   /** The active taskId persisted for `projectId` (null = explicitly closed). */
   activeTaskIdFor: (projectId: string) => string | null;
-  /** Click a card: select if not active, deselect (hide panel) if already active. */
+  /**
+   * Select `task` in `project`'s scope. Navigation, not a toggle: the requested
+   * task always ends up active, so calling this for the already-active task
+   * keeps it selected rather than hiding the panel. See `nextActiveByProject`.
+   *
+   * The name is a misnomer kept on purpose — renaming it would touch
+   * `projects.$id.tsx`, which parallel tickets own. Read it as "select". To
+   * hide the panel, call `deselect`; nothing here will do it for you.
+   */
   toggle: (
     project: ScopedProject,
     task: Task,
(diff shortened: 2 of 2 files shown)


qcentic-dev:
## Acceptance criteria — #380

- [x] **1. Clicking the already-selected pin leaves it selected and focuses the session.**
  `nextActiveByProject` returns the `activeByProject` map unchanged when `activeByProject[scopeKey] === requestedTaskId`, so the repeat request never clears the scope and the panel stays open on that session. Focus is the caller's existing `focusGridSession` request, which is unchanged and still runs on every click. Proven by `keeps the task active when it is requested again` and `returns the same map object for a repeat request so no re-render is forced`.

- [x] **2. Rapid A → B → A follows the last click; the panel does not close on the repeated A.**
  Proven by `follows the last click across a rapid A -> B -> A burst`: the sequence is folded through the reducer and asserted to be `["task-a", "task-b", "task-a"]`, to contain no `null`, and to end on `task-a`. `null` is the panel close, so asserting its absence across the burst is the "does not close" half of the criterion.

### Also verified, not in the acceptance list

- [x] Deselect still works and stays where it belongs — the session panel close button and the `terminal.close` hotkey both call `deselect`, which is untouched by this change.
- [x] Boundaries respected — only `packages/panel/src/lib/terminal-store.tsx` and its test file are modified. `projects.$id.tsx`, `ProjectBar.tsx`, `TerminalPane.tsx`, `__root.tsx` and `packages/shared` are untouched.
- [x] Existing terminal-store tests green, and the whole panel suite with them: 146 files, 1566 tests passed. `tsc --noEmit` and `eslint` on the changed files are clean.

### Nothing unchecked

Every acceptance criterion in the issue is covered by a test in this PR. The one thing tests cannot assert from the store is the on-screen caret landing in the terminal — that is `focusGridSession`, which this PR does not change and which the card-click path already relies on for the same effect.

qcentic-dev:
CI on head `cfd55b2`: **all 18 checks green**, 6 skipping (train-only jobs, expected on a ticket PR).

Conventions, Lint, Typecheck, Unit Tests, Secret Scan, Dependency Audit, Train rules, Promotion gate, E2E — Panel service seam, E2E — installer (debian/ubuntu x64), Core + Panel image build + smoke (amd64), Core release tarball smoke (linux-x64/arm64), Resolve PR image mode, Core/Panel image resolve registries — all pass.

Note that `Promotion gate` passing here is because this PR's base is the wave branch `fix/operator-pins-sessions-drop-cli`, not `main`; it is red by design only on the train-to-`main` promotion PR.

Staying a draft, as asked.

mrfqcentic (changes requested):
Review of #443 against #380 (B1). Reviewed the diff `origin/fix/operator-pins-sessions-drop-cli...cfd55b2` in a read-only worktree; every claim below is from the enforcing lines, not the PR description.

## Verdict: request changes

The reducer change is correct, minimal and well-tested, and CI is green on `cfd55b2`. Two things stop me approving: a contract comment in the changed file that now states the opposite of what the code does, and — the substantive one — the removed branch appears to be unreachable, so this PR does not demonstrably fix the operator symptom #380 describes.

## Acceptance criteria

**AC1 — "Clicking the already-selected pin leaves it selected and focuses the session": met at the store level.**
`nextActiveByProject` (`packages/panel/src/lib/terminal-store.tsx:390-397`) has no `null` return and no clearing branch; when `activeByProject[scopeKey] === requestedTaskId` it returns the map identity, so the scope keeps its selection whether or not a session is materialized. After this change `deselect` (`terminal-store.tsx:823-835`), `close` (`terminal-store.tsx:899-912`) and `closeForProject` (`terminal-store.tsx:1082-1092`) are the only writers of a `null`/absent scope value. Focus still fires on a repeat click: `selectTerminal` (`packages/panel/src/routes/projects.$id.tsx:1375-1385`) calls `focusGridSession` unconditionally, and `focusGridSession` (`terminal-store.tsx:558-561`) bumps a nonce and sets a *new* `gridFocusRequest` object every call, so the identity-preserving reducer cannot swallow it — `TerminalPanel`'s effect (`packages/panel/src/components/views/TerminalPanel.tsx:148-172`) still matches, consumes the nonce and polls the caret into the xterm textarea.

**AC2 — "Rapid A → B → A follows the last click; the panel does not close on the repeated A": met at the store level.** Each request lands on the requested id and no request can produce `null`, so a burst is last-write-wins.

**Deselect stays where it belongs: met for click gestures.** The two gestures named in the issue are intact — the pane header's "Hide session panel" item (`TerminalPane.tsx:379-386` → `__root.tsx:646` `onHide={deselect}`) and the `terminal.close` hotkey (`projects.$id.tsx:1231`). For accuracy: `deselect` is also reached from two pre-existing lifecycle paths this PR does not touch — delete/archive of the active session with no replacement (`projects.$id.tsx:1463`, `:1569`) and a clean shell exit (`TerminalPane.tsx:1135`). Neither is a repeat-click path; worth knowing they exist rather than reading the PR body's "close button and hotkey" as exhaustive.

**Tests cover both: met.** `terminal-store.test.ts:149-153` (repeat request keeps the task active), `:155-158` (identity preserved), `:168-179` (A → B → A burst, ends on A, never `null`).

## 1. `packages/panel/src/lib/terminal-store.tsx:68` — the `Ctx.toggle` contract doc now contradicts the implementation

```ts
/** Click a card: select if not active, deselect (hide panel) if already active. */
toggle: (
```

`toggle` can no longer deselect — that is the whole point of this change — but the interface doc for it still promises the old behaviour, and the function keeps the name `toggle`. This is the one comment in the file a future reader (or agent) will consult before wiring a hide affordance, and it will send them to `terminals.toggle(project, activeTask)` expecting a close that will never come. The new `nextActiveByProject` docstring states the real rule well; line 68 should follow it. Renaming `toggle` would touch `projects.$id.tsx`, which parallel tickets own, so the comment is the fix that fits this PR's scope — but then say so, rather than leaving the name unremarked.

## 2. Delivery: the removed branch looks unreachable, so B1's symptom is not demonstrably fixed

The only behavioural change here is in `toggle`. On the base branch `nextActiveTaskId` had exactly one production caller (`terminal-store.tsx:705`), and it could return `null` only when `prev[scopeKey] === task.id` **and** a session for that task already existed in that scope. All five `terminals.toggle` call sites make that combination impossible:

- `projects.$id.tsx:635` — inside `if (active?.taskId !== task.id)`, and the `activeTaskIdFor(...) === task.id` case is routed to `rehydrate` on the line above. Reaching `toggle` requires the scope's active id **not** to be `task.id`. (`selectedScopeKey` at `:182` is `projectScopeKey(id)` and `scopeKeyForProject` is `projectScopeKey(project.id)`, so the guard and the write read the same scope — no gap there.)
- `projects.$id.tsx:802` — a freshly minted optimistic task id, never already active.
- `projects.$id.tsx:1081` — guarded by `if (!currentId)`.
- `projects.$id.tsx:1092` — returns early on `nextTask.id === currentId`.
- `projects.$id.tsx:1244` — only runs after the earlier branch established there is no active session.

And the actual pin/card click never went through `toggle` at all: `onToggle: selectTerminal` (`projects.$id.tsx:1704`) → `selectTerminal` (`:1375`) → `terminals.openSession`, whose selection write already read `prev[scopeKey] === task.id ? prev : {...}` before this PR.

So the PR body's and commit message's causal story — "a second click on the already-active pin closed the session panel", "that null cleared the scope's selection, which closed the session panel under the operator" — is not supported by the code. #380 was filed from a code read ("Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)"), and the code read landed on a branch that the call sites had already made dead.

Concretely: an operator merges this, clicks the already-selected pin twice, and if the bounce is real it is still there, because whatever produces it lives somewhere else — the `focusGridSession` retry, a route remount, or one of the `deselect` lifecycle paths above.

What I'd like before approving, either is fine:

- show the reachable path — a `toggle` call site (or another consumer) where the scope's active id equals the requested id and a session exists — and add a test at that level rather than only at the reducer; **or**
- re-frame the PR honestly as a hardening refactor plus a shared reducer: state that the `null` branch was unreachable through today's call sites, that this removes the trap before someone adds a sixth caller, and that B1 needs the Panel UI actually clicked on the base branch before it can be closed. In that case #380 should stay open with a note, not be closed by this merge.

The second option is a perfectly good PR. What I don't want merged is a fix whose description asserts an operator symptom was cured when the reading says it could not have been caused here.

## Non-blocking

- `packages/panel/src/lib/__tests__/terminal-store.test.ts:160-165` — "re-selects a task whose session is already materialized" duplicates the repeat-request test two cases above it. Materialization is no longer an input to the reducer, so the name promises coverage the test cannot provide; its comment already admits as much. Either drop it or keep just the comment on the test above.
- The A → B → A test seeds `state` at `task-a` and then requests `task-a` first, so it is really a repeat-A → B → A. That is a superset of the AC, not a gap — noting it only so nobody reads it as the plain burst later.

## Mechanics checked

- Files touched: `packages/panel/src/lib/terminal-store.tsx` and `packages/panel/src/lib/__tests__/terminal-store.test.ts` only. `__root.tsx` (#379), `TerminalPane.tsx` (#386, #393), `ProjectBar.tsx` (#376, #378, #382), `projects.$id.tsx` and `packages/shared/.../harness-hook-pipeline.ts` (#387) are all untouched. Clean.
- Base is `fix/operator-pins-sessions-drop-cli` (itself descended from the open train `beta/0.4.5`), not `main` and not `beta/0.4.5` directly. Correct.
- Still a draft. Correct.
- One commit, `cfd55b2`. Subject 63 chars, longest body line 78 — both inside `header-max-length` 120 and `body-max-line-length` 132; `fix` is in `type-enum`; blank lines precede `Refs #380` and the trailer block. The `Conventions` job lints `--from base.sha --to head.sha` with this repo's config and passed.
- CI on `cfd55b2`: all 24 checks green or skipped, none failing. `Unit Tests` ran the panel suite at 146 files / 1566 tests passed, which matches the PR body.
- No leftover references to `nextActiveTaskId` anywhere in the tree.

---

https://claude.ai/code/session_019efcqAv95KSy35qpScYcvf

qcentic-dev:
@mrfqcentic — thanks, the call-site trace was right and the original framing was not. Pushed `e6159f2` (plain push, no force). **Resolution taken: the second option — re-framed as a hardening refactor.**

## Blocking 2 — delivery

I spent a bounded pass on the first option and could not show a reachable path that produces the operator symptom. Confirming your trace, and adding one call site your list did not have: **`projects.$id.tsx:558`** — the auto-select effect in the delete/archive recovery — is a sixth `toggle` caller, and it is guarded too (the effect returns unless the scope's active id is `null`). Grid cells don't reach `toggle` either; the grid's focus-in write goes straight to the store's own identity-preserving path.

One thing I did find, recorded in the body rather than claimed as B1: **`activeFor` and `activeTaskIdFor` (`terminal-store.tsx:1143-1167`) close over the render's `activeByProject`, not a ref**, and `openRequestedSession` (`projects.$id.tsx:592-640`) reads both **after** `await api.getTask(...)` while `toggle`'s updater reads the committed `prev`. A selection committed during that await leaves the guard stale — the one shape in which the old `null` branch could have fired. It is a race on the notification-open path, not a repeat click, so it is not B1. I did not add a test at that level: it needs `projects.$id.tsx`, which parallel tickets own and I must not edit, and no existing harness renders `TerminalProvider`. Flagging it so "unreachable" reads as precise rather than absolute.

So the PR title, body and the new commit message all now say the branch was unreachable through today's callers and that this removes the trap before a seventh caller arrives without the guard. No closing keyword anywhere — `Refs #380`, never `Closes` — and I posted [a note on #380](https://github.com/actana/control/issues/380#issuecomment-5512393206) saying B1 needs the Panel UI clicked on the base branch, with the four remaining suspects ranked.

## Blocking 1 — the `Ctx.toggle` contract doc

`terminal-store.tsx:68` rewritten. It states the real rule (navigation, the requested task always ends up active, call `deselect` to hide) and explicitly calls out that `toggle` is a misnomer kept on purpose, because renaming it would touch `projects.$id.tsx`. Not renamed, as you suggested.

The `nextActiveByProject` docstring also lost its unsupported causal story. It now records that the branch was unreachable, points at this review, says #380 stays open, and names every writer that legitimately clears a scope — `deselect`, `close`, `closeForProject` — including the delete/archive-with-no-replacement and clean-shell-exit paths you noted, so nobody reads "close button and hotkey" as exhaustive again.

## Non-blocking

- Dropped `re-selects a task whose session is already materialized`; its comment is folded onto the repeat-request test above it, which is where the point belonged.
- Renamed the burst test to `follows the last request across a rapid repeat-A -> B -> A burst`.

## On the first commit

`cfd55b2` is left as pushed rather than amended — rewriting it needs a force push, which I'm not doing. `e6159f2` states the correction explicitly and supersedes its causal claim.

Tests: 22 in `terminal-store.test.ts`, full panel suite **146 files / 1565 tests green** (1566 before, minus the dropped duplicate), `tsc --noEmit` and `eslint` clean. Still a draft, base unchanged.

https://claude.ai/code/session_01SGZcdyMnRX7cTYPyareJTT

mrfqcentic (approved):
Round 2 review of #443 against #380 (B1). Recreated a read-only worktree at `e6159f2`, read the diff `origin/fix/operator-pins-sessions-drop-cli...e6159f2` and `cfd55b2..e6159f2`, and checked every claim against the enforcing lines rather than the description.

## Verdict: approve

Both blocking findings from round 1 are addressed, and the resolution taken — the second option, re-frame as a hardening refactor — is carried through consistently in the contract comment, the docstring, the title, the body, the new commit message and the note on #380. The reducer is unchanged from `cfd55b2` and still correct. Nothing merged here asserts the operator symptom is cured, and #380 cannot be closed by this merge.

## Blocking 1 — the `Ctx.toggle` contract doc: fixed

`terminal-store.tsx:68-76` now reads:

```ts
/**
 * Select `task` in `project`'s scope. Navigation, not a toggle: the requested
 * task always ends up active, so calling this for the already-active task
 * keeps it selected rather than hiding the panel. See `nextActiveByProject`.
 *
 * The name is a misnomer kept on purpose — renaming it would touch
 * `projects.$id.tsx`, which parallel tickets own. Read it as "select". To
 * hide the panel, call `deselect`; nothing here will do it for you.
 */
```

That is the real rule, it names the misnomer and says why the name stays, and it points at `deselect` for the hide. Exactly what was asked. Verified against the implementation: `toggle`'s only selection write is `setActiveByProject((prev) => nextActiveByProject(prev, scopeKey, task.id))` (`:729`), which cannot produce `null`.

The `nextActiveByProject` docstring (`:378-400`) lost the causal story and now records that the branch was unreachable, cites this review, says #380 stays open, and names the legitimate clearers. I checked that last claim exhaustively: the nine `setActiveByProject` writers are `:617`, `:729`, `:780`, `:853` (all `nextActiveByProject`), `:889` (`adoptTaskId`, writes `task.id`, never null), and `:838` `deselect`, `:913` `close`, `:1096` `closeForProject`. So `deselect`/`close`/`closeForProject` are indeed the only writers of a null-or-absent scope value. Accurate.

## Blocking 2 — delivery: resolved honestly

**No closing keyword.** Scanned the title, the body and both commit messages for `clos(e|es|ed)|fix(es|ed)|resolv(e|es|ed)` followed by an issue reference: no match. Both commits carry `Refs #380`. The merge cannot close the issue. ✅

**Title.** `fix(panel): a repeat pin selection can never clear the scope` — a statement about the store invariant this change establishes, not a claim about the operator's screen. No symptom cure asserted. ✅

**Body and commit message.** Both state the branch was unreachable through current callers, both call the change a hardening refactor, both say B1 needs the Panel UI clicked before it can be closed. The body leads with an explicit correction of the first version. ✅

**The `Refs #380` note on the issue** is posted and says the same thing, with the four remaining suspects ranked. #380 is still `OPEN`. ✅

### The sixth-caller finding: accurate

I read the line rather than taking it. `projects.$id.tsx:460` aliases `const toggleTerminalSession = terminals.toggle`, and `:558` calls it — a sixth site the round 1 list missed. The guard is real: the effect opens at `:548` with `if (activeTaskId !== null) { lastActiveRef.current = …; return; }`, where `activeTaskId` is `terminals.activeTaskIdFor(selectedScopeKey)` (`:535`). The scope's active id is `null` on every path that reaches `:558`, so it can never equal the requested id. Correctly found and correctly characterised.

Re-checked the other five at head, all hold:

| Site | Guard verified at |
| --- | --- |
| `:558` | `:548` `if (activeTaskId !== null) … return` |
| `:635` | `:634` `if (activeTaskId === task.id) rehydrate(…)`, `else toggle(…)` |
| `:802` | `:784-786` id is `newClientId("t")`, freshly minted |
| `:1081` | `:1078` `if (!currentId)` |
| `:1092` | `:1091` `if (!nextTask \|\| nextTask.id === currentId) return` |
| `:1244` | `:1228-1231` `if (active) { deselect; return }` |

And the pin path still never reaches `toggle`: the "Pinned" `TaskColumn` gets `onToggle={stableSelectTerminal}` (`:2272`), `taskCardHandlersRef.current.onToggle = selectTerminal` (`:1704`), and `selectTerminal` (`:1375-1385`) calls `terminals.openSession` plus an unconditional `focusGridSession`. So a repeat pin click focuses and never had a deselect to lose.

### The stale-snapshot race note: accurate in substance

`activeFor` (`terminal-store.tsx:1157`, deps `[activeByProject, sessions, visibleScopeByProject]` at `:1170`) and `activeTaskIdFor` (`:1173`, deps at `:1181`) are `useCallback`s over the render's `activeByProject`, not a ref. `openRequestedSession` (`:588`) awaits `api.getTask(request.taskId)` at `:619` and only then reads `terminals.activeFor(...)` / `terminals.activeTaskIdFor(...)` at `:630-633`, while `toggle`'s updater reads the committed `prev`. A selection committed during that await leaves the guard reading a stale snapshot — that is a genuine shape in which the old `null` branch could have fired, and it is a notification-open race, not a repeat click. The characterisation is right, and hedging "unreachable" against it is the correct call. Not adding a test for it is reasonable given the file ownership constraint.

One imprecision: the body cites `terminal-store.tsx:1143-1167` for those two callbacks; at `e6159f2` they span `:1157-1182`. The cited range covers only part of `activeFor` and none of `activeTaskIdFor`. Cosmetic.

## Tests

- The redundant `"re-selects a task whose session is already materialized"` case is gone; its point is folded into the comment above `"keeps the task active when it is requested again"` (`terminal-store.test.ts:147-150`). ✅
- The burst test is renamed to `"follows the last request across a rapid repeat-A -> B -> A burst"` with a comment saying it seeds on A and is a superset of the plain burst (`:167-170`). ✅
- 22 tests in the file, matching the PR body.

## The reducer is unchanged and correct

`cfd55b2..e6159f2` touches only comments and tests — the `nextActiveByProject` body is byte-identical across the two commits, and no other executable line moves. I diffed it against the base to confirm the semantics too:

```ts
return activeByProject[scopeKey] === requestedTaskId
  ? activeByProject
  : { ...activeByProject, [scopeKey]: requestedTaskId };
```

Old `toggle` normalised with `prev[scopeKey] ?? null` before comparing; the new helper compares against a `string` parameter, so both `undefined` and `null` fall through to a write exactly as before. The four collapsed writers (`:617`, `:729`, `:780`, `:853`) were each `prev[k] === id ? prev : {...prev, [k]: id}` on the base branch — identical semantics. `hadSession` is removed with no leftover reference anywhere in `packages/panel/src`. The only behavioural delta remains `toggle` never writing `null`.

## Non-blocking

These are all in prose and none of them changes what the code does. Fix them if you touch the file again; I am not holding the PR for any of them.

- **`terminal-store.tsx:392`** — "before a **sixth** caller arrives without the guard" undercounts. This PR's own finding is that there are already six; the body and the commit message both correctly say "seventh". The comment is the copy that survives into the tree, so it is the one worth being right.
- **`terminal-store.tsx:390`** — "Every call site guarded that combination away, so the branch was not reachable through today's callers" reads as absolute, but the stale-closure race you documented in the body is precisely the shape where it could have fired. That nuance is what makes the removal worth more than churn, and it lives only in the PR description, which does not survive into the file. One clause here would carry it.
- **`terminal-store.tsx:395-397`** — the `deselect` reachers list gives the pane's hide affordance, the hotkey and the delete/archive paths. The clean shell exit (`TerminalPane.tsx:1135` → `onHideRef.current?.()` ← `__root.tsx:646` `onHide={deselect}`) is the same callback, so it is arguably covered, but naming it would close the gap round 1 opened.
- The body's "five hand-written copies … collapse onto one tested function" counts the reducer itself as one of the five; there are four call sites.

## Mechanics

- **Still a draft.** ✅ Base is `fix/operator-pins-sessions-drop-cli`, unchanged. ✅
- **Files touched:** `packages/panel/src/lib/terminal-store.tsx` and `packages/panel/src/lib/__tests__/terminal-store.test.ts` only.
- **Merges cleanly onto the current base tip `38679c7`** (which now carries #444 merged). `git merge-tree --write-tree` returns a tree with no conflict; `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. The branch is behind the tip — merge base is `5829f7f` — but that is not a conflict.
- **No overlap with #444**, which touched `rail-projects.ts`, `rail-projects.test.ts` and `__root.tsx`. Disjoint from this PR's two files. ✅
- **Commitlint:** `cfd55b2` subject 63 chars, longest body line 78; `e6159f2` subject 60 chars, longest body line 79. Both inside `header-max-length` 120 and `body-max-line-length` 132, `fix` is in `type-enum`, trailers are correctly separated. The `Conventions` job lints the whole range and passed.
- **CI on `e6159f2`: 18 checks green, 6 skipped (train-only), none failing.** `Unit Tests` green.
- On the test count: CI reports the panel suite at 146 files / **1569** tests, where the body says 1565. Both are right — the body counts the developer's local tree (1566 − 1 dropped duplicate), CI counts the merge ref, which includes the 4 tests #444 added to `rail-projects.test.ts` after this branch was cut. 1566 − 1 + 4 = 1569. No discrepancy.
- Re-ran `/code-review` at high effort over the full PR diff against `origin/fix/operator-pins-sessions-drop-cli`: no correctness bugs, and it independently landed on the same two documentation-accuracy nits above.

Good correction. The re-framing is the honest version and the trace that produced it is sound.

---

https://claude.ai/code/session_01TJCLM1YACM8wVXkJ5JkWzh

Issue #376, opened 2026-09-02
S1: rail highlight stays on the previous pin after a same-Core click


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `S1`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Pin two projects on the same Core. Click A, then B. URL and grid change; the accent ring stays on A until hover or a query refresh re-renders the bar.

## Where

`packages/panel/src/components/views/ProjectBar.tsx` memo around the overlay (catalog: 70–83, 714–818). `useRouter()` does not subscribe; the memo skips re-render when only the route changes.

## What to do

Subscribe the rail overlay to the route (or pass pathname/search from a hook that does) so a same-Core pin click updates the ring in the same frame as the navigation.

## Acceptance

1. Click pin A then pin B on the same Core: the ring is on B as soon as the URL is `/projects/<B>`.
2. A Core pin still includes `?coreId=` in the URL.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #461, merged to `beta/0.4.5` via #481. The PR carried `Refs` rather than a closing keyword, so closing by hand.

PR #444, opened 2026-09-02
fix(panel): Cmd-digit addresses rail pins and keeps coreId

### LINKED ISSUE #379: S2: Cmd/Ctrl+digit opens the wrong pin and drops coreId

- Hunt ID: `S2`
- Severity: broken
- First-fix owner: panel

## What the operator sees

With Core pins on the rail, press the chord for badge 1. Often hits a Panel-local pin (or nothing). Navigate omits `?coreId=` so the project page 404s or loads the Panel DB.

## Where

`packages/panel/src/routes/__root.tsx` (catalog: 253, 509–512) uses `useProjects` only; navigate without `search.coreId`. The rail merges `useRemotePinnedProjects()`.

## What to do

Hotkeys must address the same merged pin list the rail renders, and must pass `coreId` when the pin is Core-owned.

## Acceptance

1. Chord for visible badge N opens that pin, including Core pins.
2. Core navigations include `?coreId=`.

qcentic-dev:
Closes #379 (S2).

## The bug

The left rail draws its hotkey badges from the **merged** pin list: the Panel's own rows plus every Core's pins, fetched
by `useRemotePinnedProjects` and tagged with the Core that owns them. The Cmd/Ctrl+digit handler in
`packages/panel/src/routes/__root.tsx` read `useProjects()` alone.

Two lists, one set of badges. With Core pins on the rail:

1. Badge N and chord N addressed different projects — the chord hit whatever Panel-local project sat in slot N, or
   nothing at all when the Panel had fewer pins than the rail showed.
2. The navigate omitted `search.coreId`, so even a correct hit asked the project route for a Core-owned project without
   naming the Core — the page 404s or quietly loads the Panel's own DB.

## The fix

Both halves move into `~/lib/rail-projects`, where they are pure and testable:

- **`mergeRailProjects(panelProjects, remotePinned)`** — the same list `ProjectBar` merges, in the same concatenation
  order, so a `pinnedOrder` tie breaks identically in the rail and in the chord.
- **`resolveRailChordTarget(clusters, groupDigit, projectDigit)`** — resolves a chord against the very clusters the
  badges were drawn from. Covers all three paths: the two-level All-mode group→project chord, the single-digit flat
  rail, and the Cmd-release jump to a group's first project.
- **`railNavigateSearch(target)`** — `{ coreId }` for a Core-owned pin, `undefined` for a Panel row. Every chord
  navigation is funnelled through one `navigateToRailTarget` callback, so a future caller cannot drop it again.

`RailProject` gains the optional `coreId` the merged rows already carry.

## Cost

Calling `useRemotePinnedProjects` in the shell adds a second per-Core `listProjects` fan-out. `PanelLinkClient.watch` is
refcounted, so the Core subscription itself is shared and costs no extra subscribe frame; the extra read matches how
`useCores` is already called from several places. Sharing the fetch would mean reworking the hook, which touches
`ProjectBar`'s read path — out of scope here, and `ProjectBar.tsx` is owned by other tickets in this wave.

## Tests

Four cases added to `packages/panel/src/lib/__tests__/rail-projects.test.ts`:

- badge N resolves to the Nth **merged** pin across an interleaved panel/Core pinned order;
- the same resolution against the unmerged list returns the wrong project — the bug, pinned down;
- a Core pin's navigation carries `{ coreId }`, a Panel row's carries nothing;
- a group→project chord resolves against the rendered clusters, Core pin and `coreId` included.

`packages/panel`: 9/9 in `rail-projects.test.ts`, `pnpm typecheck` clean, `pnpm lint` 0 errors.

## Boundaries

Touches only `__root.tsx`, `lib/rail-projects.ts` and its test. No change to `ProjectBar.tsx`, `projects.$id.tsx`,
`lib/terminal-store.tsx` or `packages/shared`.

https://claude.ai/code/session_01Vsnw9CXT1HUr1n9uKuxt68

### DIFF
--- a/packages/panel/src/lib/__tests__/rail-projects.test.ts
+++ b/packages/panel/src/lib/__tests__/rail-projects.test.ts
@@ -2,6 +2,10 @@ import { describe, expect, it } from "vitest";
 import type { Group } from "~/db/schema";
 import {
   clusterPinnedByGroup,
+  getRailClusters,
+  mergeRailProjects,
+  railNavigateSearch,
+  resolveRailChordTarget,
   usesDirectRailProjectShortcuts,
   type RailProject,
 } from "~/lib/rail-projects";
@@ -68,5 +72,82 @@ describe("usesDirectRailProjectShortcuts", () => {
     expect(
       usesDirectRailProjectShortcuts([group("dev", "Dev")], ACTIVE_GROUP_UNGROUPED),
     ).toBe(true);
+  });
+});
+
+// #379: the rail badges number the MERGED pin list — the Panel's own rows plus
+// every Core's pins — so a chord read off `useProjects` alone addressed a
+// different list than the one the operator is looking at.
+describe("rail chords address the merged pin list (#379)", () => {
+  function corePin(
+    id: string,
+    coreId: string,
+    groupId: string | null,
+    pinnedOrder: number,
+  ): RailProject {
+    return { ...project(id, groupId, pinnedOrder), coreId };
+  }
+
+  const panelPins = [project("panel-a", null, 0), project("panel-b", null, 2)];
+  const remotePins = [corePin("core-a", "core-1", null, 1), corePin("core-b", "core-2", null, 3)];
+
+  it("resolves visible badge N to the Nth merged pin, Core pins included", () => {
+    const merged = mergeRailProjects(panelPins, remotePins);
+    const clusters = getRailClusters(merged, [], ACTIVE_GROUP_ALL);
+
+    // One flat cluster, badges 1..4 in pinned order — panel, Core, panel, Core.
+    expect(clusters).toHaveLength(1);
+    expect(clusters[0]!.projects.map((p) => p.id)).toEqual([
+      "panel-a",
+      "core-a",
+      "panel-b",
+      "core-b",
+    ]);
+    for (const [badge, id] of [
+      [1, "panel-a"],
+      [2, "core-a"],
+      [3, "panel-b"],
+      [4, "core-b"],
+    ] as const) {
+      expect(resolveRailChordTarget(clusters, null, badge)?.id).toBe(id);
+    }
+  });
+
+  it("would have opened the wrong pin from the Panel's own rows alone", () => {
+    // The bug, pinned down: badge 2 is a Core pin, but the unmerged list makes
+    // it `panel-b`. Merging is what makes the chord agree with the badge.
+    const unmerged = getRailClusters(panelPins, [], ACTIVE_GROUP_ALL);
+    expect(resolveRailChordTarget(unmerged, null, 2)?.id).toBe("panel-b");
+  });
+
+  it("carries ?coreId= for a Core pin and omits it for a Panel row", () => {
+    const clusters = getRailClusters(mergeRailProjects(panelPins, remotePins), [], ACTIVE_GROUP_ALL);
+
+    expect(railNavigateSearch(resolveRailChordTarget(clusters, null, 2)!)).toEqual({
+      coreId: "core-1",
+    });
+    expect(railNavigateSearch(resolveRailChordTarget(clusters, null, 4)!)).toEqual({
+      coreId: "core-2",
+    });
+    expect(railNavigateSearch(resolveRailChordTarget(clusters, null, 1)!)).toBeUndefined();
+  });
+
+  it("resolves a group→project chord against the same clusters the rail draws", () => {
+    const groups = [group("alpha", "Alpha"), group("omega", "Omega")];
+    const merged = mergeRailProjects(
+      [project("panel-alpha", "alpha", 0)],
+      [corePin("core-omega", "core-9", "omega", 1), corePin("core-alpha", "core-9", "alpha", 2)],
+    );
+    const clusters = getRailClusters(merged, groups, ACTIVE_GROUP_ALL);
+
+    // Group 1 badge 2 is the Core's Alpha pin; group 2 badge 1 its Omega pin.
+    expect(resolveRailChordTarget(clusters, 1, 2)?.id).toBe("core-alpha");
+    expect(resolveRailChordTarget(clusters, 2, 1)?.id).toBe("core-omega");
+    expect(railNavigateSearch(resolveRailChordTarget(clusters, 1, 2)!)).toEqual({
+      coreId: "core-9",
+    });
+    // A Cmd release mid-chord jumps to the group's first project — same list.
+    expect(resolveRailChordTarget(clusters, 1, 1)?.id).toBe("panel-alpha");
+    expect(resolveRailChordTarget(clusters, 3, 1)).toBeUndefined();
   });
 });
--- a/packages/panel/src/lib/rail-projects.ts
+++ b/packages/panel/src/lib/rail-projects.ts
@@ -7,6 +7,8 @@ import {
 } from "~/shared/ui-preferences";
 
 export type RailProject = PinnedOrderable & {
+  /** The Core that owns this row; absent/null means the Panel's own rows. */
+  coreId?: string | null;
   groupId: string | null;
   name: string;
 };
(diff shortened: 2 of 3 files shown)


qcentic-dev:
## Acceptance checklist — issue #379

- [x] **AC1 — Chord for visible badge N opens that pin, including Core pins.**
      The handler now clusters `mergeRailProjects(useProjects(), useRemotePinnedProjects())` — the same list, in the
      same concatenation order, that `ProjectBar` renders badges from — and resolves the chord with
      `resolveRailChordTarget(clusters, groupDigit, projectDigit)`, which indexes projects **within their cluster**,
      exactly as the badge numbers do (`projectIndex + 1`). All three chord paths go through it: the flat single-digit
      rail, the two-level All-mode group→project chord, and the Cmd-release jump to a group's first project.
      Proved by `resolves visible badge N to the Nth merged pin, Core pins included` (badges 1–4 over an interleaved
      panel/Core pinned order) and `resolves a group→project chord against the same clusters the rail draws`. A
      companion test, `would have opened the wrong pin from the Panel's own rows alone`, pins down the old behaviour.

- [x] **AC2 — Core navigations include `?coreId=`.**
      Every chord navigation is funnelled through one `navigateToRailTarget`, whose search comes from
      `railNavigateSearch(target)`: `{ coreId }` when the pin is Core-owned, `undefined` for a Panel row. That is the
      same rule `ProjectBar`'s tile `onClick` already applied, so a chord and a click on the same tile now land on the
      same URL. Proved by `carries ?coreId= for a Core pin and omits it for a Panel row` and by the `coreId` assertion
      in the group→project test.

Nothing is left unchecked.

### Verification

| Check | Result |
| --- | --- |
| `vitest run src/lib/__tests__/rail-projects.test.ts` | 9 passed (5 pre-existing + 4 new) |
| `pnpm typecheck` (`packages/panel`) | clean |
| `pnpm lint` | 0 errors (45 pre-existing warnings, none in touched files) |
| `vitest run` (full `packages/panel` suite) | 1560 passed, 5 skipped; two server files (`operator-auth`, `live-read-path`) time out under full-suite load and pass in isolation — pre-existing flakes, unrelated to this change |

### Scope

Files touched: `packages/panel/src/routes/__root.tsx`, `packages/panel/src/lib/rail-projects.ts`,
`packages/panel/src/lib/__tests__/rail-projects.test.ts`. `ProjectBar.tsx`, `projects.$id.tsx`,
`lib/terminal-store.tsx` and `packages/shared` are untouched.

qcentic-dev:
### CI status of the PR head (`8c382b4`)

Run [33648382014](https://github.com/actana/control/actions/runs/33648382014) — **completed / success**. All 24 checks
green or skipped; **zero failures**.

| Check | Result |
| --- | --- |
| Conventions | pass |
| Lint | pass |
| Typecheck | pass |
| Unit Tests | pass |
| Secret Scan | pass |
| Dependency Audit | pass |
| Train rules | pass |
| Promotion gate | pass |
| Resolve PR image mode | pass |
| Core image / Resolve registries, Build + smoke (amd64) | pass |
| Panel image / Resolve registries, Build + smoke (amd64) | pass |
| Smoke — Core release tarball (linux-x64, linux-arm64) | pass |
| E2E — installer (debian, x64), (ubuntu, x64) | pass |
| Train versions, Resolve train tags, Core/Panel image (train), image manifest publishes | skipping (train-only, expected off a train branch) |

`Promotion gate` reports **pass** here rather than the red it shows on a train→`main` promotion PR — this PR's base is
`fix/operator-pins-sessions-drop-cli`, not `main`, so the gate is not the one being exercised.

Left as a **draft** and not merged, per the ticket.

mrfqcentic (approved):
## Review — PR #444 against issue #379 (S2)

Approving. Both acceptance criteria are **met**, verified by reading the enforcing lines rather than the PR
description, and by running the suites in a scratch copy. Scope and process constraints all hold.

### AC1 — "Chord for visible badge N opens that pin, including Core pins" — **MET**

The chord and the badges now index the same object graph, and I checked each link in that chain:

- **Same merged list, same order.** `ProjectBar.tsx:91-95` merges `[...(panelProjects ?? []), ...remotePinned]`.
  `mergeRailProjects` (`rail-projects.ts:105-110`) is the identical concatenation, panel rows first. That order is
  what makes the merge safe: `getPinnedProjects` sorts on `pinnedOrder` and breaks ties on `createdAt`
  (`pinned-project-order.ts:8-19`) via a stable `Array.prototype.sort`, so an exact tie falls back to input order —
  which is now the same input order in both places.
- **Same source.** `__root.tsx:266` calls `useRemotePinnedProjects()`, the same hook `ProjectBar.tsx:90-91` calls, so
  the Core rows carry the same `coreId` tag from `use-fleet.ts:392-394`.
- **Same clustering.** `__root.tsx:537` calls `getRailClusters(railProjects, groups, activeGroup)`.
  `ProjectBar.tsx:186-190` inlines exactly what `getRailClusters` does — `clusterPinnedByGroup(sortedPinned,
  orderedGroups)` in All mode, `getGroupRailCluster(...)` otherwise, dropped when empty.
- **Same numbering.** Badges are `projectNumber = projectIndex + 1` within a cluster and `groupNumber = clusterIndex +
  1` (`ProjectBar.tsx:1055-1057`). `resolveRailChordTarget` is `clusters[groupDigit - 1]?.projects[projectDigit - 1]`
  (`rail-projects.ts:120-127`) — the same arithmetic, inverted.
- **All three chord paths** go through it: flat rail (`__root.tsx:548`), All-mode second digit (`__root.tsx:568`),
  Cmd-release jump (`__root.tsx:582`). Each is a line-for-line equivalent of the indexing it replaced, so the refactor
  itself changes no behaviour; the merge is the whole of the fix.

A detail the PR description undersells: merging can also introduce the synthetic Ungrouped cluster that Core pins with
no local filing produce. Before this change the shell's cluster *count* could differ from the rail's, so the first
(group) digit could misaddress too, not just the second. That is fixed by the same line.

Digit coverage lines up: `HOTKEY_LIMIT = PINNED_SLOT_COUNT = 9` (`keybindings/match.ts:39`) and the handler guards
`/^[1-9]$/`, so there is no badge without a chord or chord without a badge.

### AC2 — "Core navigations include `?coreId=`" — **MET**

`navigateToRailTarget` (`__root.tsx:500-511`) is the single funnel, and every chord path calls it — I checked that no
`router.navigate` to `/projects/$id` survives elsewhere in the effect. Its search comes from `railNavigateSearch`
(`rail-projects.ts:136-138`): `{ coreId }` when owned, `undefined` otherwise — byte-for-byte the rule
`ProjectBar.tsx:1181-1184` already applies on tile click, so chord and click now produce the same URL.

I verified the negative half rather than assuming it, since `undefined` could plausibly have meant "retain current
search". It does not: `@tanstack/router-core@1.169.2` `router.js:1136` is `if (!dest.search) return {}` — a falsy
`search` **drops** the params. So chording a Panel row while sitting on `/projects/X?coreId=core-1` does not leak the
old Core through. The route consumes it at `projects.$id.tsx:134-139` / `168-169`.

### Tests — cover what the ticket asks

`rail-projects.test.ts` adds four cases: badges 1-4 across a genuinely **interleaved** panel/Core pinned order
(`pinnedOrder` 0/2 panel, 1/3 Core), a regression witness proving the unmerged list returns the wrong project, both
`coreId` polarities, and a group→project chord with `coreId`. 9/9 pass locally.

### Non-blocking observations

1. **`rail-projects.ts:105-110` — the merge is extracted but not yet adopted by its other caller.** `ProjectBar.tsx:92-95`
   still open-codes the same merge, so two copies of the rule exist, and they are textually different (ProjectBar
   short-circuits an empty `remotePinned` to return `panelProjects` unchanged, possibly `undefined`). They are
   result-equivalent today, which is why this is not blocking — but a single source of truth was the point of the
   helper. `ProjectBar.tsx` belongs to #376/#378/#382 in this wave, so deferring is right; worth a follow-up so the
   two lists cannot drift apart again.
2. **`__root.tsx:266` — the fan-out is now doubled.** `useRemotePinnedProjects` runs in both `Shell` and `ProjectBar`,
   so each refresh tick costs two `listProjects` per Core plus two `listProjectPresentation` reads. The PR's claim that
   the subscription itself is free checks out — `PanelLinkClient.watch` refcounts and only sends a subscribe frame at
   `refs === 1` (`panel-link-client.ts:244-248`). The extra reads are real but bounded, and `Shell` is not mounted on
   auth paths (`__root.tsx:118`), so nothing fans out unauthenticated.
3. **`__root.tsx:537` — group order can diverge mid-drag.** The shell clusters with `groups`; `ProjectBar` clusters
   with `orderedGroups`, its optimistic group-drag order. While a group label is being dragged the two group-digit
   numberings can disagree. Pre-existing (the base used `groups` too), transient, and needs Cmd held during a mouse
   drag — noting it only so it is not mistaken for a regression later.
4. **No test exercises `__root.tsx` itself** — that `Shell` passes the merged list, and that every navigation is
   funnelled. The repo has no route-level test harness at all, so extracting the logic into `lib` and testing it there
   is the established convention and the right call here; flagging the residue, not asking for a change.

### Verification performed

| Check | Result |
| --- | --- |
| `vitest run src/lib/__tests__/rail-projects.test.ts` | 9 passed |
| `vitest run src/lib/__tests__ src/components` | 74/75 files, 672 passed; the 1 failure is `session-finish-from-core.integration.test.ts`, which needs the `better-sqlite3` native binding my sandbox install skipped — it imports none of the changed files |
| `tsc --noEmit` (`packages/panel`, after `generate:routes`) | clean |
| `eslint` on the three touched files | 0 errors; 1 warning, pre-existing at `__root.tsx:98`, outside every hunk |
| CI on head `8c382b4` | all 20 checks green, 4 skipped ([run 33648382014](https://github.com/actana/control/actions/runs/33648382014)) |

### Scope and process

- Touches exactly three files: `__root.tsx`, `lib/rail-projects.ts`, `lib/__tests__/rail-projects.test.ts`.
- `rail-projects.ts` is **purely additive** — zero deleted or modified lines, so it cannot change what `ProjectBar.tsx`
  or `ProjectPicker.tsx` render. `ProjectBar.tsx` has a zero-line diff.
- Untouched, as required: `lib/terminal-store.tsx` (#380 / PR #443), `TerminalPane.tsx`, `routes/projects.$id.tsx`,
  `packages/shared/.../harness-hook-pipeline.ts` (#387).
- Base is `fix/operator-pins-sessions-drop-cli` (not `main`, not `beta/0.4.5`), head is one commit on top of the base
  tip, still a draft.
- Commitlint: one commit, header 58 ≤ 120, no body or footer line over 132, body-leading-blank and the `Refs #379`
  trailer blank line both satisfied.

Nothing here blocks. Approving.

PR #445, opened 2026-09-02
fix(core): treat ready as stranded — sweep and PTY-exit settle it

### LINKED ISSUE #387: F1: dead session stuck at ready — no session:finished, no pty:exit

qcentic-dev:
Closes #387 (F1).

## The zombie

A Session on pairdemo sat at `ready`, "Waiting for initial prompt…", LIVE empty, hours after its PTY died — and survived a Core container recreate. The CLI said the Session "has no harness running" while the card said it was waiting for a prompt.

`ready` is the status a bare Session lives in: the harness is spawned, and nothing moves the row until the first `UserPromptSubmit`. So no hook ever fires for it, which means:

- `sessionProcessExited` in `harness-hook-pipeline.ts` only patched `running | needs-input`, so the PTY's death settled nothing.
- the boot sweep read the same two statuses through `queryActiveTasks`, so the recreate did not catch it either.

Two blind spots, one row, no third thing that was ever going to come back for it.

## What changed

**The PTY-exit settle (`packages/shared/src/harness-hook-pipeline.ts`).** `ready` is now in scope, on a scale of its own: a clean exit is `finished`, and anything else is `disconnected` rather than `terminated`. A non-zero exit out of a turn is a turn that was killed, which is what `terminated` says; a non-zero exit out of a Session that never ran a turn says only that the process went away.

Nothing else in that branch moved. The already-settled guard, the subagent clears, and the in-turn gate #385 added are untouched.

**The boot sweep (`packages/shared/src/core-query.ts`, `packages/core/src/core-query-store.ts`, `core-entry.ts`).** A new `queryStrandedReadyTasks` finds the `ready` rows this Core once spawned a PTY for, and `listBootSweepTasks` hands the sweep that list alongside the rows that claim a live process.

`ready` deliberately is **not** swept on the status alone. It is also the status of a Session the operator created and has not started yet — sweeping those would flip a whole queue of unstarted work to `disconnected` on every Core restart. The discriminator is the `pty:spawn` the Core appended when it started the harness, read from the event log because the task row records no such thing: there is no "was started" column, and the harness session id stays `null` in exactly the bare case. A `ready` row with no spawn behind it is left alone, and a log whose spawn has aged out fails the same safe way.

The union is assembled in the query store, not in the sweep, because the quiet-Session backstop shares `listActiveTasks` and must **not** see `ready`: a bare Session waiting on its first prompt is allowed to sit silent indefinitely, and settling it for being quiet would be a new bug. Only the boot sweep reads the wider list.

## Tests

- `harness-hook-pipeline.test.ts` — a `ready` Session settles to `disconnected` on a bad exit and `finished` on a clean one, with no Stop hook ever posted; `running`/`needs-input` keep the `terminated` scale; an already-finished row is still left alone.
- `core-query.test.ts` — `queryStrandedReadyTasks` finds the spawned `ready` row and not the unstarted one, ignores another Session's spawn and a shell PTY that belongs to no task, spans projects and archived rows, and returns `[]` when there is no event log to read.
- `core-session-sweep.test.ts` — the boot sweep settles a leftover `ready` row beside the `running` ones, leaves the unstarted `ready` row alone, and appends `task:updated` without a `session:finished`.
- `harness-status-detection.test.ts` — end to end on the Core: the row is created `ready`, no hook is ever posted, `sessionExited` settles it.

The Stop and OpenCode idle-finish suites are green: a healthy OpenCode stop still emits `session:finished`.

`pnpm typecheck`, `pnpm lint` (0 errors) and `pnpm test` all pass.

## Boundaries

No Panel files touched — nothing near `__root.tsx` (#379) or `terminal-store.tsx` (#380). Built on the merged #385 change to `harness-hook-pipeline.ts` without widening or undoing it, and it does not take the foreign-session path #390 will.

https://claude.ai/code/session_016RZuD4w44wp8ZMR3iMxzYB

### DIFF
--- a/docs/harness-status-detection.md
+++ b/docs/harness-status-detection.md
@@ -23,6 +23,7 @@ about everything else on a Core — as an Event replayed off its cursor.
 | Hook delivery misses | `packages/core/src/harness-hook-delivery.ts` |
 | Quiet-Session backstop | `packages/core/src/core-session-backstop.ts` |
 | Boot sweep | `packages/core/src/core-session-sweep.ts` |
+| Relaunch reset | `packages/core/src/core-session-relaunch.ts` |
 
 The Panel keeps a hook endpoint (`POST /api/hooks/<slug>`) for its own remaining
 local task rows, running the same shared pipeline. A Core-owned Session never
@@ -35,11 +36,13 @@ outside the process. The Core installs them per workspace at spawn time, so each
 Session reports its own state.
 
 A new Session starts in `ready` (terminal spawned, prompt waiting). The first
-hook flips it.
+hook flips it — and if no hook ever comes, a dead-process fallback does (see
+below): `ready` is not a resting place for a Session whose process is gone.
 
 | Hook event                         | Mapped status   | Meaning                                  |
 | ---------------------------------- | --------------- | ---------------------------------------- |
 | _(spawn)_                          | `ready`         | Terminal up, operator hasn't typed yet   |
+| _(agent respawn)_                  | `ready`         | Harness up again for a Session that never worked |
 | `UserPromptSubmit`                 | `running`       | Prompt submitted; work began             |
 | `Stop`                             | `finished`\*    | Foreground turn ended (see below)        |
 | `SubagentStart` / `SubagentStop`   | _(bookkeeping)_ | Tracks still-active subagents            |
@@ -305,6 +308,21 @@ Sessions whose process is gone:
   The Panel's own exit patch no longer fires for a Core-owned row: it was
   unconditional, so it would overwrite an `interrupted` Session with `finished`
   and raise a spurious `session:finished` besides.
+
+  **`ready` settles here too** (issue 387), and it is the one status in scope
+  that describes no work. A *bare* Session — spawned with no prompt, sitting on
+  "Waiting for initial prompt…" — never leaves `ready` until its first
+  `UserPromptSubmit`, so not one hook ever fires for it and there is no `Stop`
+  to wait for. Left out, its row went on claiming to be waiting for a prompt
+  for as long as the database existed; one was found alive on a Core hours
+  after its PTY died, and still there after the container was recreated.
+
+  It settles on a scale of its own: **`disconnected`, whatever the exit code**.
+  `terminated` would say a turn was killed, and `finished` would say work
+  completed — `CoreTaskWriter` appends `session:finished` on exactly that
+  transition, so a bare Session the operator closes with `/exit` would ring a
+  completion ding for a turn that never ran. `disconnected` claims only that
+  the process went away, which is the whole of what is known.
 - **Boot sweep** — `core-session-sweep.ts` runs once per Core boot, before the
   PTY core, the hook receiver or the core-link server can produce a Session of
   this run. At that moment no PTY of this process exists, so every row still
@@ -314,11 +332,44 @@ Sessions whose process is gone:
   `task:updated` event a connected Panel re-renders from — a sweep nobody is
   told about leaves the operator looking at the same wrong card.
 
+  **Spawned `ready` rows are orphans too** (issue 387), for the same reason the
+  PTY-exit settle covers them: a bare Session's PTY dies with the Core and no
+  hook was ever going to report it. They cannot be swept on the status alone,
+  because `ready` is also the status of a Session the operator created and has
+  not started — flipping a queue of unstarted work to `disconnected` on every
+  restart would trade one wrong card for many. The discriminator is whether
+  this Core ever spawned a *harness* for the row, read from the `pty:spawn` in
+  the event log (`queryStrandedReadyTasks`). The row cannot answer it: there is
+  no "was started" column, and the harness session id is no help either, since
+  Claude Code's `SessionStart` captures one before any turn. A `ready` row with
+  no agent spawn behind it is left alone. The evidence is permanent — nothing
+  prunes `event_log` — so the first boot after this shipped settled every
+  historical spawned `ready` row in one batch, and every boot since sees only
+  what the run before it stranded.
+
+  The quiet-Session backstop deliberately does **not** see these rows: it reads
+  the narrow `listActiveTasks`, and a bare Session waiting on its first prompt
+  is allowed to sit silent for as long as the operator likes.
+
   `disconnected` rather than `finished` or `terminated`: it is the status the
   Panel already uses for a process that went away without reporting, and it
   claims nothing about how the work ended. The Panel has had this sweep for its
   own rows all along; what it never had was scope over Core-owned ones, which is
   every Session on a remote Core (issue 243).
+- **Relaunch reset** — the mirror of the two above (`core-session-relaunch.ts`).
+  Once `ready` is a status a Session can leave, something has to write it back:
+  the operator reopens a settled bare Session, the Core spawns a healthy
+  harness, it waits at its prompt, and no hook fires until the first prompt —
+  so the card would read `disconnected` over a live harness, on every harness
+  family. On an **agent** spawn (never a `shell` or `shellSession` one), a row
+  in a settled status that has **never worked** goes back to `ready`.
+
+  "Never worked" is read from the event log, not the row: every status change
+  on a Core-owned row appends a `task:updated` carrying the status that was
+  *patched* (`queryTaskEverWorked`), and any status other than `ready` /
+  `disconnected` says a turn happened — including on a harness that goes
+  straight from `ready` to `finished` without ever reporting `running`. A
+  `finished` Session being reopened is being resumed, and keeps its card.
 
 ## Terminal-input fallback
 
@@ -336,6 +387,12 @@ Session had neither hooks nor a fallback and never left `ready`.
 The narrower question matters: Cursor and Codex both take our hooks file and
 report a turn's *end*, so "hooks were installed" would suppress the fallback and
 leave their Sessions on `ready` for the whole first turn.
+
+Note that neither of those failures parks a Session on `ready` forever any more
+(issue 387). A Session sitting on `ready` whose PTY then dies is settled by the
+PTY-exit path, and one whose Core dies under it is settled by the boot sweep;
+what a missing `running` signal costs is a card that under-reports a live turn,
+not a row that outlives its process.
 
 ### The CLI has no equivalent, and says so
 
--- a/packages/core/src/__tests__/core-link-relaunch-port.test.ts
+++ b/packages/core/src/__tests__/core-link-relaunch-port.test.ts
@@ -1,0 +1,132 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import {
+  PtyCoreLinkServer,
+  type WebSocketLike,
+  type WebSocketServerLike,
+} from "../pty-core-link-server";
+import type { PtyCore, PtyCoreEvent } from "../pty-manager";
+
+// Which spawns ask the relaunch port (issue 387, review finding 2).
+//
+// The port is what puts a bare Session's card back on `ready` when a harness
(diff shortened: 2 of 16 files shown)


qcentic-dev:
## Acceptance criteria — issue #387

- [x] **1. A `ready` session whose PTY is gone settles (disconnected or finished) without waiting for a Stop hook.**
  `handleHarnessHookEvent`'s `sessionProcessExited` branch now includes `ready`: clean exit → `finished`, anything else → `disconnected`. Pinned in `harness-hook-pipeline.test.ts` ("settles a ready Session to disconnected when its PTY dies badly", "settles a ready Session to finished when the exit was clean", "settles without a Stop hook ever arriving") and end to end on the Core in `harness-status-detection.test.ts` ("settles a bare Session left on ready when its PTY dies"), where the row is created `ready` and not one hook is posted before `sessionExited`.

- [x] **2. Core recreate sweeps leftover `ready` rows the way it already sweeps `running` / `needs-input`.**
  The boot sweep reads `listBootSweepTasks` — `queryActiveTasks` plus the new `queryStrandedReadyTasks`. Pinned in `core-session-sweep.test.ts` ("sweeps the bare Session a dead PTY left on ready", "sweeps a leftover ready row alongside the rows that claim to be working", "settles a swept ready row without claiming its work finished") against the Core's real SQLite and real event log.

  One qualification on "the way it already sweeps": `ready` is swept on evidence, not on the status. It is also the status of a Session the operator created and has not started, so a blind status filter would flip every unstarted row to `disconnected` on each restart. The gate is the `pty:spawn` this Core appended when it started the harness — the pairdemo row has one, an unstarted row does not. `core-query.test.ts` pins both sides, and the pre-existing sweep test asserting an unstarted `ready` row stays `ready` still passes unchanged.

- [x] **3. Healthy OpenCode stop still emits `session:finished`.**
  Untouched and green: `harness-hooks-opencode-plugin.test.ts` (the plugin's `session.idle` → `Stop`), `harness-status-detection.test.ts` ("finishes the turn on Stop, and marks it distinguishably", "holds the finish while a background subagent is still working") and `session-finish-event.test.ts` (13 cases on when the finish event is and is not appended). The change adds an `else if` after the existing `running | needs-input` settle; no Stop path was modified.

Not in scope here, per the orchestration: **copying the live pairdemo row out before the Core is recreated** is the orchestrator's step, not this PR's. The confirmation the issue asks for — that the boot sweep marks that specific row `disconnected` — is therefore covered by tests against a reconstructed row, not by the live one.

### Verification

`pnpm typecheck`, `pnpm lint` (0 errors, only pre-existing warnings, none in the changed files) and `pnpm test` (all 6 stages) pass locally.

https://claude.ai/code/session_016RZuD4w44wp8ZMR3iMxzYB

mrfqcentic (changes requested):
Reviewed against #387 (F1) on `e6bc759`, diffed against `origin/fix/operator-pins-sessions-drop-cli`. Read the enforcing lines rather than the description, and re-ran the suites locally.

**All three acceptance criteria are met**, and the mechanism is good: the SQL is right, the `EXISTS` gate genuinely separates a spawned bare Session from an unstarted one, the union is disjoint by status so nothing is swept twice, the sweep still runs before any PTY of this run can exist, the quiet-Session backstop still gets the narrow `listActiveTasks`, and #385's in-turn gate and the already-settled guard are untouched. The commits pass commitlint (bodies ≤ 132), CI is green on the head commit, the base is the wave branch, and no Panel file is touched.

Requesting changes on the consequences of moving a row **off** `ready`, which the PR does not discuss and does not test. `ready` is a one-way status — `DEFAULT_TASK_STATUS` is the only writer, `SessionStart` maps to `null` — and two consumers read it as "this Session has never had a conversation".

---

### Required

**1. `packages/shared/src/harness-hook-pipeline.ts:275` and `packages/core/src/core-entry.ts:244` — a settled bare Session relaunches as `--resume` into a session id that never existed.**

`packages/panel/src/lib/harness-command.ts:26` keys claude-code's launch mode on `task.status === "ready"` alone, ignoring `claudeSessionId` (the codex and opencode branches below it do check it). Bare Core Session, never prompted, `claudeSessionId` null — `terminal-store.tsx:308-320` deliberately skips the PATCH for a Core-owned row, so it stays null. Core recreated → this branch or the boot sweep writes `disconnected` → the operator reopens → `baseCommandForTask` mints a fresh UUID and `harnessLaunchMode` now answers `"resume"`, producing `claude --resume <uuid-that-never-existed>`. The spawn dies immediately. `TerminalPane.tsx:1030-1070` catches it and retries fresh, so it self-heals — but every relaunch of a swept bare Session now takes the failure path and prints `[resume failed; starting a fresh Claude Code session]` at the operator.

Gating the claude-code branch of `harnessLaunchMode` on `claudeSessionId`, the way the codex branch already is, removes it. That file is not `__root.tsx` (#379) or `terminal-store.tsx` (#380), so it does not collide with the two Panel PRs in flight.

**2. `packages/core/src/core-entry.ts:244` — a relaunched bare Session keeps reading `disconnected` while it is healthy.**

Nothing ever writes `ready` back. A swept bare Session that the operator reopens spawns a fine harness that sits at its prompt, and no hook fires (`SessionStart` → `null`), so the card reads an error status until the operator's first `UserPromptSubmit`. This one is not claude-code-only — codex/opencode/cursor-cli relaunch cleanly and still show the wrong card. Before this PR the same relaunch read `ready`, which was correct.

That is the same class of bug as F1 itself: an operator-visible card that does not describe the Session. Either the spawn path or `SessionStart` should reset a settled row, or the trade needs stating in the PR and filing as a follow-up.

**3. `packages/shared/src/core-query.ts:235` — the docstring's safety argument is not true, and the real consequence is not stated.**

> A log whose `pty:spawn` has aged out answers nothing for that row, which fails toward leaving it alone.

Nothing prunes `event_log` — there is no `DELETE FROM event_log` anywhere in the repo, and `cli/src/event-tip.ts:18` notes the same about the event store generally. The evidence is permanent, so on the first Core boot after this ships, *every* historical `ready` row that ever had a harness spawned flips to `disconnected` in one batch, each appending a `task:updated`. That is probably intended and is arguably the point — but the comment promises the opposite, and the batch belongs in the PR description.

**4. `docs/harness-status-detection.md` is now wrong and was not updated.**

Two passages the source file itself points readers at:

- L298-307 — "The Core moves Sessions still in an **active** status (`running` / `needs-input`) to `finished` (exit 0) or `terminated`" — now also `ready` → `finished` / `disconnected`.
- L308-311 — "every row still claiming `running` / `needs-input` is an orphan of the previous one" — now also spawned `ready` rows.

The status table at L42 (`(spawn)` → `ready`) and the note at L334 are in the same area. The merged sibling for #385, `5d33f0c`, updated this doc (+59 lines) for a smaller change to the same function; this one should too.

---

### Non-blocking

- **`packages/shared/src/harness-hook-pipeline.ts:276`** — `exit_code === 0` on a `ready` row writes `finished`, and `CoreTaskWriter` appends `session:finished` on exactly that transition. An operator who opens a bare Session and types `/exit` gets a completion ding for a Session titled "Waiting for initial prompt…" that never ran a turn. `harness-status-detection.test.ts` pins this deliberately, two tests after the sweep test asserts "no ding for a Session that never worked" — the two paths disagree about the same Session. #387's acceptance does say "or finished, if the exit was clean", so this is a decision to make explicitly rather than a defect; `disconnected` for both exit codes would make the file consistent with itself.
- **`packages/shared/src/core-query.ts:253`** — the evidence predicate matches *any* `pty:spawn` for the task id, but `pty-core-link-server.ts:1439` records one for the `shell: true` and `shellSession: true` variants too, and `pty-manager.ts` notes shells "carry a taskId for routing but are not agent work". The `shellSession` discriminant is in the payload and the query never reads it. Nothing collides today (CLI shells use `cli_shell_<uuid>`, Panel VM shells reuse a user-terminal id), and the test at `core-query.test.ts` pins the weaker `task_id IS NULL` shape rather than the synthetic-id shape the code actually writes. Worth reading the payload flag, or giving harness spawns their own kind.
- **`packages/core/src/core-session-sweep.ts:58`** — the dep is still named `listActiveTasks` while every caller now passes `listBootSweepTasks`. The doc comment above it was updated; the name was not.
- **`packages/core/src/core-query-store.ts:193`** — this `log.warn` is unreachable: `queryStrandedReadyTasks` swallows every error and returns `[]`. A read that fails for a real reason (`SQLITE_BUSY` past the 5s timeout, a corrupt page) is indistinguishable from "no stranded rows" and leaves no line — on the boot path added to make an invisible zombie visible. `listActiveTasks` has the same dead handler, so this is pre-existing style, not a regression.
- **`packages/core/src/__tests__/core-session-sweep.test.ts:203`** — 108 characters; the file otherwise wraps.

---

### Suites

`pnpm test` run six times in a scratch copy. `packages/shared` and the new `packages/core` tests are green every time; the reported `packages/core` flake reproduces and is **identified**:

```
packages/core/src/__tests__/core-files-abort.test.ts
  > a client that hangs up mid-transfer
  > "leaves the Project writable, rather than 409 for the lifetime of the process"
  Error: Test timed out in 5000ms.   (observed 5075ms and 5171ms)
```

2 of 6 full runs; 0 of 13 `packages/core`-only runs; 0 of 5 runs of that file alone under synthetic CPU load. It only appears when all 57 core files run in parallel workers. It is that test's own budget: it is the one test in its describe with no explicit timeout (its fd-leak sibling carries `20_000`), and inside vitest's default 5s it builds 900 real files, tars them, runs two 5s `eventually` polls, sleeps a fixed 250ms and then does another full HTTP round trip. When it times out, `cleanupTrees()` removes the tmp tree under the still-unpacking handler, which is the `ENOENT: mkdir '…/drop/deeply/nested/…'` on stderr just above the failure.

Untouched by this PR — the file last changed in `3c3d6bb` (#215) — so it is pre-existing and not a reason to hold this branch. It is worth its own issue: an explicit timeout on that test, or a smaller tar.

qcentic-dev:
Superseded by **#448** — https://github.com/actana/control/pull/448

Same work, same base, same tree. Closed for a commit-message reason only.

One commit body on this branch (`1aba1e9`, the claude-code launch-mode gate) carried a 133-character line against commitlint's 132-character limit, so the Conventions check would have stayed red on it. Correcting a pushed commit message requires rewriting history, and this branch is never force-pushed — so the six commits were cherry-picked in order onto a fresh branch cut from the current `fix/operator-pins-sessions-drop-cli` tip, the one body was rewrapped before it was ever pushed, and the result is #448.

`commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD` reports 0 problems across all six commits there.

The round 1 review's findings are all addressed on #448, and the reply owed to that review is posted there. This branch is left in place for the review history; nothing was merged from it.

https://claude.ai/code/session_016RZuD4w44wp8ZMR3iMxzYB

Issue #447, opened 2026-09-02
panel: baseCommandForTask mints a session id before harnessLaunchMode reads it, so the claude-code resume gate cannot bite


qcentic-dev:
Found while addressing review finding 1 on #445 (issue #387, F1). The gate the review asked for is in place; this is why it does not change behaviour yet.

## What the gate does

`packages/panel/src/lib/harness-command.ts` now reads:

```ts
if (task.agent === "claude-code") {
  return task.claudeSessionId && task.status !== "ready" ? "resume" : "new";
}
```

symmetric with the codex branch below it. A Session with no captured id has nothing to resume into, so it starts new.

## Why it is inert at the only call site

`harnessLaunchMode` has exactly one caller, `baseCommandForTask` in `packages/panel/src/lib/terminal-store.tsx`, and it mints an id **before** asking:

```ts
let sessionId = task.claudeSessionId;
if (!sessionId && task.agent !== "codex" && task.agent !== "opencode") {
  sessionId = newSessionId();
  ...                                     // PATCH skipped for a Core-owned row
}
const mode = harnessLaunchMode({ ...task, claudeSessionId: sessionId });
```

For claude-code the spread always carries a non-null id — the freshly minted one when the row had none — so the new gate sees the same input it saw before. Codex and opencode are exempt from the mint, which is why their gates work.

This also predates #387: `remoteTaskFromSnapshot` never copies `claudeSessionId` off a Core snapshot and the Panel's PATCH is skipped for a Core-owned row, so a Core-owned claude-code Session's Panel-side id is **always** null. Every relaunch of one therefore takes `claude --resume <uuid-that-never-existed>`, fails, and self-heals through `TerminalPane.tsx`'s retry with `[resume failed; starting a fresh Claude Code session]`. #387 extended that from "Sessions that had a turn" to bare Sessions as well; the gate above is the correct half of the fix, and this is the other half.

## What would fix it

Ask before minting — compute the mode from the row, then mint the id the "new" launch needs:

```ts
const mode = harnessLaunchMode(task);
let sessionId = task.claudeSessionId;
if (!sessionId && ...) sessionId = newSessionId();
```

Or carry the row's own id across the core-link so a Core-owned Session has a real one to resume with (the "protocol gap" the comment at that mint already notes: `tasksMutate` does not carry `claudeSessionId`).

Not done in #445: `terminal-store.tsx` is owned by #380, which is in flight on the same wave.

## References

- PR #445 — https://github.com/actana/control/pull/445
[...]

Issue #446, opened 2026-09-02
flake: core-files-abort "leaves the Project writable" times out at the 5s default under full-suite parallelism


qcentic-dev:
Found while reviewing #445 (issue #387, F1). **Not caused by that branch** — `packages/core/src/__tests__/core-files-abort.test.ts` last changed in `3c3d6bb` (#215), and #445 does not touch it. Filed separately at the reviewer's request so the branch is not held for it.

## What happens

```
packages/core/src/__tests__/core-files-abort.test.ts
  > a client that hangs up mid-transfer
  > "leaves the Project writable, rather than 409 for the lifetime of the process"
  Error: Test timed out in 5000ms.   (observed 5075ms and 5171ms)
```

## When

From the review of #445, run in a scratch copy:

- 2 of 6 full `pnpm test` runs.
- 0 of 13 `packages/core`-only runs.
- 0 of 5 runs of that file alone under synthetic CPU load.

It appears only when all core test files run together in parallel workers. Independently reproduced on this branch: one `packages/core` stage failure across three full-suite runs, green on every `packages/core`-only and single-file run.

## Why

It is the test's own budget, not a product defect. It is the one test in its `describe` with no explicit timeout — its fd-leak sibling carries `20_000` — and inside vitest's default 5s it builds 900 real files, tars them, runs two 5s `eventually` polls, sleeps a fixed 250ms, then does another full HTTP round trip. Under full-suite worker contention that does not fit.

The `ENOENT: mkdir '…/drop/deeply/nested/…'` on stderr just above the failure is a consequence, not a cause: when the test times out, `cleanupTrees()` removes the tmp tree under the still-unpacking handler.

## Suggested fix

An explicit timeout on that test (matching the `20_000` its sibling already carries), or a smaller tar. Either is a test-only change.

## References

- PR #445 — https://github.com/actana/control/pull/445
- Review — https://github.com/actana/control/pull/445#pullrequestreview-5092086064

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016RZuD4w44wp8ZMR3iMxzYB

PR #448, opened 2026-09-02
fix(core): treat ready as stranded — sweep and PTY-exit settle it

### LINKED ISSUE #387: F1: dead session stuck at ready — no session:finished, no pty:exit

qcentic-dev:
Closes #387 (F1). Replaces #445 — same work, same base, rebuilt on a fresh branch so one over-long commit-message line could be corrected without a force push. See the note at the bottom.

## The zombie

A Session on pairdemo sat at `ready`, "Waiting for initial prompt…", LIVE empty, hours after its PTY died — and survived a Core container recreate. The CLI said the Session "has no harness running" while the card said it was waiting for a prompt.

`ready` is the status a bare Session lives in: the harness is spawned, and nothing moves the row until the first `UserPromptSubmit`. So no hook ever fires for it, which means:

- `sessionProcessExited` in `harness-hook-pipeline.ts` only patched `running | needs-input`, so the PTY's death settled nothing.
- the boot sweep read the same two statuses through `queryActiveTasks`, so the recreate did not catch it either.

Two blind spots, one row, no third thing that was ever going to come back for it.

## What changed

**The PTY-exit settle (`packages/shared/src/harness-hook-pipeline.ts`).** `ready` is now in scope, on a scale of its own: it settles to `disconnected` whatever the exit code. `terminated` would say a turn was killed; `finished` would say work completed, and `CoreTaskWriter` appends `session:finished` on exactly that transition — so a bare Session closed with `/exit` would ring a completion ding for a turn that never ran. `disconnected` claims only that the process went away.

Nothing else in that branch moved. The already-settled guard, the subagent clears, and the in-turn gate #385 added are untouched.

**The boot sweep (`packages/shared/src/core-query.ts`, `packages/core/src/core-query-store.ts`, `core-entry.ts`).** A new `queryStrandedReadyTasks` finds the `ready` rows this Core once spawned a harness for, and `listBootSweepTasks` hands the sweep that list alongside the rows that claim a live process.

`ready` deliberately is **not** swept on the status alone. It is also the status of a Session the operator created and has not started yet — sweeping those would flip a whole queue of unstarted work to `disconnected` on every Core restart. The discriminator is the `pty:spawn` the Core appended when it started the harness, read from the event log because the task row records no such thing: there is no "was started" column, and the harness session id is no help either, since Claude Code's `SessionStart` captures one before any turn. A `ready` row with no agent spawn behind it is left alone, and the payload's `shellSession` flag keeps a VM shell from counting as one.

The union is assembled in the query store, not in the sweep, because the quiet-Session backstop shares `listActiveTasks` and must **not** see `ready`: a bare Session waiting on its first prompt is allowed to sit silent indefinitely, and settling it for being quiet would be a new bug. Only the boot sweep reads the wider list.

### One-time boot batch, on first ship

Nothing in this repo prunes `event_log` — there is no `DELETE FROM event_log` anywhere — so the `pty:spawn` evidence is permanent. **On the first Core boot after this ships, every historical `ready` row that ever had a harness spawned settles to `disconnected` at once**, each appending its own `task:updated`. That is the backlog of zombies this query exists to find, arriving in one go because nothing was looking for them before; every later boot sees only what the run before it stranded. Operators with long-lived Cores should expect a batch of cards to move on that first restart. The docstring on `queryStrandedReadyTasks` says the same, and `docs/harness-status-detection.md` records it.

**That batch also reorders the operator's lists.** A status patch stamps `updated_at`, and Fleet, Archived and the sweep's own query all order by it — so every row the batch settles, archived ones included since the query deliberately spans them, floats to the top at the boot time, above recent work, and there is no undoing it. The existing sweep already does this for `running` / `needs-input`, so it is not new in kind, but the one-time batch makes it new in volume.

## The consequences of moving a row off `ready`

Round 1's review confirmed all three acceptance criteria and requested changes on what happens once `ready` stops being a one-way status. Both are in this PR.

**Relaunching a settled bare Session no longer resumes into a session id that never existed** (`packages/panel/src/lib/harness-command.ts`). `harnessLaunchMode` keyed claude-code on `status !== "ready"` alone; it is now gated on `claudeSessionId` the way the codex branch beside it already is. Nothing to resume into means start new.

That gate is correct but currently inert at its one call site: `baseCommandForTask` in `terminal-store.tsx` mints a session id *before* it asks, so the spread always carries a non-null one for claude-code. That file is owned by #380 on this wave, so the mint is filed as **#447** rather than fixed here.

**A relaunched bare Session no longer reads `disconnected` over a healthy harness** (`packages/core/src/core-session-relaunch.ts`). Nothing wrote `ready` back, so the operator reopened a swept Session, the Core spawned a fine harness, it sat at its prompt, and the card stayed wrong until the first prompt — on every harness family, not just claude-code. Now an **agent** spawn (never a `shell` or `shellSession` one) on a row sitting on **`disconnected`** that is **proven never to have worked** puts it back on `ready`.

Both halves are narrower than they look, because this write destroys information when it is wrong (round 2 review).

`disconnected` rather than "any settled status": before this PR `ready` was one-way, so a Session that never worked could only ever *be* `ready`, and `disconnected` is the only status this PR's own settles write for one. `finished`, `terminated` and `interrupted` are unreachable for such a Session, so they are not candidates at all — by construction, not by a log read being right about history it cannot see.

"Never worked" is read from the event log, not the row: every status change on a Core-owned row appends a `task:updated` carrying the status that was *patched* (`queryTaskProvenNeverWorked`), and any status other than `ready` / `disconnected` says a turn happened — including on a harness that goes straight from `ready` to `finished` without ever reporting `running`. A `finished` Session being reopened is being resumed and keeps its card.

That read sees **only status-bearing `task:updated` events, and those start at v0.4.0** (`2dd34a8`): before it the payload was `{taskId, projectId}` and nothing else, and since `event_log` is never pruned a Core upgraded from 0.3.x still holds status-less rows for Sessions that ran for hours. Read as "did any turn happen", their silence is indistinguishable from a Session that never ran one — and the answer would come out backwards in the destructive direction. So the read demands **positive evidence**: a row these paths settled always carries a `"status":"disconnected"` update, so a row with no status-bearing update at all is a legacy log and answers "cannot tell, do not reset".

**`docs/harness-status-detection.md`** is updated the way `5d33f0c` updated it for #385: the PTY-exit and boot-sweep passages, the status table, the file registry, and the terminal-input note.

## Tests

- `harness-hook-pipeline.test.ts` — a `ready` Session settles to `disconnected` on a bad exit *and* on a clean one, with no Stop hook ever posted; `running`/`needs-input` keep the `terminated` scale; an already-finished row is still left alone.
- `core-query.test.ts` — `queryStrandedReadyTasks` finds the spawned `ready` row and not the unstarted one, ignores another Session's spawn, a VM Shell Session spawn, and a plain shell whose synthetic `cli_shell_<uuid>` id names no row; spans projects and archived rows; returns `[]` with no event log. `queryTaskProvenNeverWorked` is pinned across every status, the no-status payload, a pre-v0.4.0 log, and a row with no history at all.
- `core-session-sweep.test.ts` — the boot sweep settles a leftover `ready` row beside the `running` ones, leaves the unstarted `ready` row alone, and appends `task:updated` without a `session:finished`.
- `core-session-relaunch.test.ts` — the reset moves the bare Session the sweep settled, leaves a Session that worked (and one that worked and was then swept), **leaves a legacy `finished` row and a legacy `disconnected` row whose `task:updated` payloads carry no status**, never considers `finished` / `terminated` / `interrupted` at all, does not disturb a live turn, appends the card event, is idempotent, and survives a write that throws.
- `core-link-relaunch-port.test.ts` — an agent spawn asks the relaunch port; a plain shell and a VM Shell Session never do; a Core with no port spawns exactly as before.
- `harness-status-detection.test.ts` — end to end on the Core: the row is created `ready`, no hook is ever posted, `sessionExited` settles it; a clean exit raises no `session:finished`.
- `harness-command.test.ts` — claude-code starts fresh with no captured id at `disconnected` or `finished`, and still resumes with one.

The Stop and OpenCode idle-finish suites are green: a healthy OpenCode stop still emits `session:finished`.

`pnpm typecheck`, `pnpm lint` (0 errors) and `pnpm test` (all 6 stages) pass locally, and `commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD` reports 0 problems across all six commits.

## Acceptance criteria — issue #387

- [x] **1. A `ready` session whose PTY is gone settles (disconnected or finished) without waiting for a Stop hook.** `sessionProcessExited` includes `ready`, settling it to `disconnected`. Pinned in `harness-hook-pipeline.test.ts` (one case posts no Stop at all) and end to end on the Core in `harness-status-detection.test.ts`.
- [x] **2. Core recreate sweeps leftover `ready` rows the way it already sweeps `running` / `needs-input`.** The boot sweep reads `listBootSweepTasks` = `queryActiveTasks` + `queryStrandedReadyTasks`, pinned against the Core's real SQLite and real event log. Qualified: `ready` is swept on `pty:spawn` evidence rather than on the status alone, because `ready` is also the status of a Session the operator has not started yet.
- [x] **3. Healthy OpenCode stop still emits `session:finished`.** Untouched and green: `harness-hooks-opencode-plugin.test.ts`, the Stop cases in `harness-status-detection.test.ts`, and all of `session-finish-event.test.ts`.

Copying the live pairdemo row out before the Core is recreated is the orchestrator's step, not this PR's, so criterion 2's live confirmation is covered by tests against a reconstructed row.

## Follow-ups filed

- **#446** — the pre-existing `core-files-abort.test.ts` flake the round 1 review identified. Not caused by this branch; not fixed here.
- **#447** — `baseCommandForTask` mints a session id before `harnessLaunchMode` reads it, which is why the claude-code gate above cannot bite yet. Lives in `terminal-store.tsx` (#380).

## Boundaries

No Panel file this wave owns is touched — nothing in `__root.tsx` (#379) or `terminal-store.tsx` (#380). Built on the merged #385 change to `harness-hook-pipeline.ts` without widening or undoing it, and it does not take the foreign-session path #390 will.

## Why this PR and not #445

#445 carried the identical work, but one commit body had a 133-character line against commitlint's 132 limit, which would have kept the Conventions check red. Correcting a pushed commit message needs history rewriting, and this branch is never force-pushed — so the six commits were cherry-picked onto a fresh branch off the current base, the one body was rewrapped before it was ever pushed, and #445 is closed pointing here. The tree is identical; the only differences from #445 are that one commit message and the newer base (which now includes #444).

https://claude.ai/code/session_016RZuD4w44wp8ZMR3iMxzYB

### DIFF
--- a/docs/harness-status-detection.md
+++ b/docs/harness-status-detection.md
@@ -23,6 +23,7 @@ about everything else on a Core — as an Event replayed off its cursor.
 | Hook delivery misses | `packages/core/src/harness-hook-delivery.ts` |
 | Quiet-Session backstop | `packages/core/src/core-session-backstop.ts` |
 | Boot sweep | `packages/core/src/core-session-sweep.ts` |
+| Relaunch reset | `packages/core/src/core-session-relaunch.ts` |
 
 The Panel keeps a hook endpoint (`POST /api/hooks/<slug>`) for its own remaining
 local task rows, running the same shared pipeline. A Core-owned Session never
@@ -35,11 +36,13 @@ outside the process. The Core installs them per workspace at spawn time, so each
 Session reports its own state.
 
 A new Session starts in `ready` (terminal spawned, prompt waiting). The first
-hook flips it.
+hook flips it — and if no hook ever comes, a dead-process fallback does (see
+below): `ready` is not a resting place for a Session whose process is gone.
 
 | Hook event                         | Mapped status   | Meaning                                  |
 | ---------------------------------- | --------------- | ---------------------------------------- |
 | _(spawn)_                          | `ready`         | Terminal up, operator hasn't typed yet   |
+| _(agent respawn)_                  | `ready`         | Harness up again for a Session that never worked |
 | `UserPromptSubmit`                 | `running`       | Prompt submitted; work began             |
 | `Stop`                             | `finished`\*    | Foreground turn ended (see below)        |
 | `SubagentStart` / `SubagentStop`   | _(bookkeeping)_ | Tracks still-active subagents            |
@@ -305,6 +308,21 @@ Sessions whose process is gone:
   The Panel's own exit patch no longer fires for a Core-owned row: it was
   unconditional, so it would overwrite an `interrupted` Session with `finished`
   and raise a spurious `session:finished` besides.
+
+  **`ready` settles here too** (issue 387), and it is the one status in scope
+  that describes no work. A *bare* Session — spawned with no prompt, sitting on
+  "Waiting for initial prompt…" — never leaves `ready` until its first
+  `UserPromptSubmit`, so not one hook ever fires for it and there is no `Stop`
+  to wait for. Left out, its row went on claiming to be waiting for a prompt
+  for as long as the database existed; one was found alive on a Core hours
+  after its PTY died, and still there after the container was recreated.
+
+  It settles on a scale of its own: **`disconnected`, whatever the exit code**.
+  `terminated` would say a turn was killed, and `finished` would say work
+  completed — `CoreTaskWriter` appends `session:finished` on exactly that
+  transition, so a bare Session the operator closes with `/exit` would ring a
+  completion ding for a turn that never ran. `disconnected` claims only that
+  the process went away, which is the whole of what is known.
 - **Boot sweep** — `core-session-sweep.ts` runs once per Core boot, before the
   PTY core, the hook receiver or the core-link server can produce a Session of
   this run. At that moment no PTY of this process exists, so every row still
@@ -314,11 +332,44 @@ Sessions whose process is gone:
   `task:updated` event a connected Panel re-renders from — a sweep nobody is
   told about leaves the operator looking at the same wrong card.
 
+  **Spawned `ready` rows are orphans too** (issue 387), for the same reason the
+  PTY-exit settle covers them: a bare Session's PTY dies with the Core and no
+  hook was ever going to report it. They cannot be swept on the status alone,
+  because `ready` is also the status of a Session the operator created and has
+  not started — flipping a queue of unstarted work to `disconnected` on every
+  restart would trade one wrong card for many. The discriminator is whether
+  this Core ever spawned a *harness* for the row, read from the `pty:spawn` in
+  the event log (`queryStrandedReadyTasks`). The row cannot answer it: there is
+  no "was started" column, and the harness session id is no help either, since
+  Claude Code's `SessionStart` captures one before any turn. A `ready` row with
+  no agent spawn behind it is left alone. The evidence is permanent — nothing
+  prunes `event_log` — so the first boot after this shipped settled every
+  historical spawned `ready` row in one batch, and every boot since sees only
+  what the run before it stranded.
+
+  The quiet-Session backstop deliberately does **not** see these rows: it reads
+  the narrow `listActiveTasks`, and a bare Session waiting on its first prompt
+  is allowed to sit silent for as long as the operator likes.
+
   `disconnected` rather than `finished` or `terminated`: it is the status the
   Panel already uses for a process that went away without reporting, and it
   claims nothing about how the work ended. The Panel has had this sweep for its
   own rows all along; what it never had was scope over Core-owned ones, which is
   every Session on a remote Core (issue 243).
+- **Relaunch reset** — the mirror of the two above (`core-session-relaunch.ts`).
+  Once `ready` is a status a Session can leave, something has to write it back:
+  the operator reopens a settled bare Session, the Core spawns a healthy
+  harness, it waits at its prompt, and no hook fires until the first prompt —
+  so the card would read `disconnected` over a live harness, on every harness
+  family. On an **agent** spawn (never a `shell` or `shellSession` one), a row
+  in a settled status that has **never worked** goes back to `ready`.
+
+  "Never worked" is read from the event log, not the row: every status change
+  on a Core-owned row appends a `task:updated` carrying the status that was
+  *patched* (`queryTaskEverWorked`), and any status other than `ready` /
+  `disconnected` says a turn happened — including on a harness that goes
+  straight from `ready` to `finished` without ever reporting `running`. A
+  `finished` Session being reopened is being resumed, and keeps its card.
 
 ## Terminal-input fallback
 
@@ -336,6 +387,12 @@ Session had neither hooks nor a fallback and never left `ready`.
 The narrower question matters: Cursor and Codex both take our hooks file and
 report a turn's *end*, so "hooks were installed" would suppress the fallback and
 leave their Sessions on `ready` for the whole first turn.
+
+Note that neither of those failures parks a Session on `ready` forever any more
+(issue 387). A Session sitting on `ready` whose PTY then dies is settled by the
+PTY-exit path, and one whose Core dies under it is settled by the boot sweep;
+what a missing `running` signal costs is a card that under-reports a live turn,
+not a row that outlives its process.
 
 ### The CLI has no equivalent, and says so
 
--- a/docs/harness-status-detection.md
+++ b/docs/harness-status-detection.md
@@ -345,7 +345,10 @@ Sessions whose process is gone:
   no agent spawn behind it is left alone. The evidence is permanent — nothing
   prunes `event_log` — so the first boot after this shipped settled every
   historical spawned `ready` row in one batch, and every boot since sees only
-  what the run before it stranded.
+  what the run before it stranded. That batch also reorders the operator's
+  lists: a status patch stamps `updated_at`, which Fleet and Archived both
+  order by, so every row it settles floats to the top at the boot time, above
+  recent work, and there is no undoing it.
 
   The quiet-Session backstop deliberately does **not** see these rows: it reads
   the narrow `listActiveTasks`, and a bare Session waiting on its first prompt
@@ -362,14 +365,36 @@ Sessions whose process is gone:
   harness, it waits at its prompt, and no hook fires until the first prompt —
   so the card would read `disconnected` over a live harness, on every harness
   family. On an **agent** spawn (never a `shell` or `shellSession` one), a row
-  in a settled status that has **never worked** goes back to `ready`.
+  sitting on **`disconnected`** that is **proven never to have worked** goes
+  back to `ready`.
+
+  Both halves of that are narrower than they look, and deliberately so, because
+  this write destroys information when it is wrong.
+
+  `disconnected` rather than "any settled status": before `ready` became a
+  status a Session can leave it was one-way, so a Session that never worked
+  could only ever *be* `ready`, and the only status the two fallbacks above
+  write for one is `disconnected`. `finished`, `terminated` and `interrupted`
+  are therefore unreachable for such a Session — every row wearing one worked
+  for it — and excluding them here means a real card cannot be overwritten
+  even if the history read below is wrong.
 
   "Never worked" is read from the event log, not the row: every status change
   on a Core-owned row appends a `task:updated` carrying the status that was
-  *patched* (`queryTaskEverWorked`), and any status other than `ready` /
+  *patched* (`queryTaskProvenNeverWorked`), and any status other than `ready` /
   `disconnected` says a turn happened — including on a harness that goes
   straight from `ready` to `finished` without ever reporting `running`. A
   `finished` Session being reopened is being resumed, and keeps its card.
+
+  That read sees **only status-bearing `task:updated` events, and those start
+  at v0.4.0** (`2dd34a8`): before it the payload was `{taskId, projectId}` and
+  nothing else. Since `event_log` is never pruned, a Core upgraded from 0.3.x
+  still holds status-less rows for Sessions that ran for hours, and their
+  silence is indistinguishable from a Session that never ran a turn. So the
+  read demands *positive* evidence — a row the fallbacks above settled always
+  carries a `"status":"disconnected"` update, so a row with no status-bearing
+  update at all is a legacy log, and the answer is "cannot tell, do not reset"
+  rather than "never worked".
 
 ## Terminal-input fallback
 
--- a/packages/core/src/__tests__/core-link-relaunch-port.test.ts
+++ b/packages/core/src/__tests__/core-link-relaunch-port.test.ts
@@ -1,0 +1,132 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import {
+  PtyCoreLinkServer,
+  type WebSocketLike,
+  type WebSocketServerLike,
+} from "../pty-core-link-server";
+import type { PtyCore, PtyCoreEvent } from "../pty-manager";
+
+// Which spawns ask the relaunch port (issue 387, review finding 2).
+//
+// The port is what puts a bare Session's card back on `ready` when a harness
+// comes up for it again. The decision of WHETHER to reset is the port's
+// (`core-session-relaunch.ts`, tested there against real rows); the decision
+// of whether to ask it at all is here, and it is a narrow one: an agent spawn
+// asks, and the two shell variants — which carry a `taskId` for routing and
+// are not harness work — never do.
+
+type Listener = (...args: unknown[]) => void;
+
+class FakeWebSocket {
+  readyState = 1;
+  sent: string[] = [];
+  private listeners: Record<string, Listener[]> = {};
+
+  send(data: string): void {
+    this.sent.push(data);
+  }
+  close(): void {
+    this.readyState = 3;
+    this.emit("close");
+  }
+  on(event: string, cb: Listener): void {
+    (this.listeners[event] ??= []).push(cb);
+  }
+  removeAllListeners(): void {
+    this.listeners = {};
+  }
+  emit(event: string, ...args: unknown[]): void {
+    for (const cb of this.listeners[event] ?? []) cb(...args);
+  }
+  receive(frame: unknown): void {
+    this.emit("message", JSON.stringify(frame));
+  }
+  ofType(type: string): Array<Record<string, unknown>> {
+    return this.sent
+      .map((raw) => JSON.parse(raw) as Record<string, unknown>)
+      .filter((frame) => frame.type === type);
+  }
+}
+
+class FakeWebSocketServer {
+  private connCb: ((ws: WebSocketLike) => void) | null = null;
+  connect(ws: FakeWebSocket): void {
+    this.connCb?.(ws as unknown as WebSocketLike);
+  }
+  close(): void {}
+  on(event: string, cb: Listener): void {
+    if (event === "connection") this.connCb = cb as (ws: WebSocketLike) => void;
+  }
+}
+
+function mockCore() {
+  let nextPty = 0;
+  const core = {
+    setEmitTarget: (_fn: ((event: PtyCoreEvent) => void) | null) => {},
+    spawn: async () => ({ ptyId: `pty-${++nextPty}`, hooksReportTurnStart: true }),
+    write: () => true,
+    resize: () => true,
+    kill: () => true,
+    killLaunchProcesses: async () => ({ ptyCount: 0, ports: [] }),
+    findByTask: () => ({ ptyId: null }),
+    taskIdForPty: () => null,
+    replay: () => ({ data: "", nextSeq: 0 }),
+    killAll: () => {},
+  };
+  return core as unknown as PtyCore;
+}
+
+describe("which spawns ask the relaunch port (issue 387)", () => {
+  let wss: FakeWebSocketServer;
+  let server: PtyCoreLinkServer;
+  let asked: string[];
+
+  async function spawn(opts: Record<string, unknown>, reqId: string): Promise<void> {
+    const ws = new FakeWebSocket();
+    wss.connect(ws);
+    ws.receive({ type: "spawn", reqId, opts });
+    await vi.waitFor(() => expect(ws.ofType("spawned").length).toBeGreaterThan(0));
+  }
+
+  beforeEach(() => {
+    wss = new FakeWebSocketServer();
+    asked = [];
+    server = new PtyCoreLinkServer(mockCore(), {
+      port: 0,
+      createServer: () => wss as unknown as WebSocketServerLike,
+      relaunchPort: { agentSpawned: (taskId) => void asked.push(taskId) },
+      liveEventPollMs: 10_000,
+    });
+  });
+
+  afterEach(() => {
+    server.close();
+  });
+
+  it("asks for an agent spawn — the relaunch a settled bare Session gets", async () => {
+    await spawn({ taskId: "t-1", cwd: "/w", command: "claude", agent: "claude-code" }, "a1");
+    expect(asked).toEqual(["t-1"]);
+  });
+
+  it("never asks for a plain shell, which is not harness work", async () => {
+    await spawn({ taskId: "t-1", cwd: "/w", command: "bash", shell: true }, "s1");
+    expect(asked).toEqual([]);
+  });
+
+  it("never asks for a VM Shell Session, for the same reason", async () => {
+    await spawn({ taskId: "term_vm_1", command: "", shellSession: true }, "v1");
+    expect(asked).toEqual([]);
+  });
+
+  it("is optional — a Core with no port spawns exactly as before", async () => {
+    server.close();
+    wss = new FakeWebSocketServer();
+    server = new PtyCoreLinkServer(mockCore(), {
+      port: 0,
+      createServer: () => wss as unknown as WebSocketServerLike,
+      liveEventPollMs: 10_000,
+    });
+    await spawn({ taskId: "t-1", cwd: "/w", command: "claude", agent: "claude-code" }, "n1");
+    expect(asked).toEqual([]);
+  });
+});
--- a/packages/core/src/__tests__/core-session-relaunch.test.ts
+++ b/packages/core/src/__tests__/core-session-relaunch.test.ts
@@ -12,7 +12,7 @@ import {
   configureCoreQueryStore,
   coreQueryStore,
   disposeCoreQueryStore,
-  taskEverWorked,
+  taskProvenNeverWorked,
 } from "../core-query-store";
 import {
   appendEvent,
@@ -47,8 +47,15 @@ describe("putting a relaunched Session back on ready", () => {
     if (status !== "ready") writer.mutate({ op: "update", taskId, status });
   };
   const statusOf = (taskId: string) => coreQueryStore.getTask(taskId)?.status;
+  /**
+   * A `task:updated` in the shape `CoreTaskWriter` wrote it BEFORE v0.4.0 —
+   * `{taskId, projectId}` and no status (2dd34a8 added the status field). The
+   * only way to reproduce the history a Core upgraded from 0.3.x still holds.
+   */
+  const legacyUpdate = (taskId: string) =>
+    appendEvent("task:updated", JSON.stringify({ taskId, projectId: "p1" }), { taskId });
   const relaunch = (taskId: string) =>
-    readySessionOnAgentSpawn({ writer, everWorked: taskEverWorked }, taskId);
+    readySessionOnAgentSpawn({ writer, provenNeverWorked: taskProvenNeverWorked }, taskId);
 
   beforeEach(() => {
     userDataDir = fs.mkdtempSync(path.join(os.tmpdir(), "ac-session-relaunch-"));
@@ -103,6 +110,58 @@ describe("putting a relaunched Session back on ready", () => {
 
     expect(relaunch("t-swept")).toBe(false);
     expect(statusOf("t-swept")).toBe("disconnected");
+  });
+
+  it("does not reset a legacy finished row whose updates carry no status", () => {
+    // A Core upgraded from 0.3.x. This Session ran for hours and finished; its
+    // whole history predates the status field, so the event log holds no proof
+    // either way. Read as "did any turn happen" the answer comes out backwards
+    // and destroys the operator's record — so the reset must not fire, and the
+    // status check must not offer `finished` as a candidate in the first place.
+    coreMutationStore.mutateTask({
+      op: "create",
+      taskId: "t-legacy",
+      projectId: "p1",
+      title: "t-legacy",
+      agent: "claude-code",
+      status: "finished",
+    });
+    legacyUpdate("t-legacy");
+    legacyUpdate("t-legacy");
+
+    expect(relaunch("t-legacy")).toBe(false);
+    expect(statusOf("t-legacy")).toBe("finished");
+  });
+
+  it("does not reset a legacy disconnected row either", () => {
+    // The same old log, on the one status the reset does consider. Here the
+    // narrowed status set gives no protection at all and the evidence check is
+    // the only thing standing between a Session that worked and a wiped card.
+    coreMutationStore.mutateTask({
+      op: "create",
+      taskId: "t-legacy-gone",
+      projectId: "p1",
+      title: "t-legacy-gone",
+      agent: "claude-code",
+      status: "disconnected",
+    });
+    legacyUpdate("t-legacy-gone");
+
+    expect(relaunch("t-legacy-gone")).toBe(false);
+    expect(statusOf("t-legacy-gone")).toBe("disconnected");
+  });
+
+  it("never considers a status a never-worked Session cannot be wearing", () => {
+    // Before issue 387 `ready` was one-way, and the only status this Core's own
+    // settles write for a never-worked row is `disconnected`. So these three
+    // are unreachable by construction, and are not candidates — no log read
+    // needed, and none trusted.
+    for (const status of ["finished", "terminated", "interrupted"]) {
+      const taskId = `t-${status}`;
+      insert(taskId, status);
+      expect(relaunch(taskId)).toBe(false);
+      expect(statusOf(taskId)).toBe(status);
+    }
   });
 
   it("does not disturb a live turn, or a Session already ready", () => {
@@ -161,7 +220,10 @@ describe("putting a relaunched Session back on ready", () => {
     });
 
     expect(
-      readySessionOnAgentSpawn({ writer: failing, everWorked: taskEverWorked }, "t-bare"),
+      readySessionOnAgentSpawn(
+        { writer: failing, provenNeverWorked: taskProvenNeverWorked },
+        "t-bare",
+      ),
     ).toBe(false);
     expect(statusOf("t-bare")).toBe("disconnected");
   });
--- a/packages/core/src/__tests__/core-session-relaunch.test.ts
+++ b/packages/core/src/__tests__/core-session-relaunch.test.ts
@@ -1,0 +1,168 @@
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { bootstrapCoreDb } from "../core-db-bootstrap";
+import {
+  configureCoreMutationStore,
+  coreMutationStore,
+  disposeCoreMutationStore,
+} from "../core-mutation-store";
+import {
+  configureCoreQueryStore,
+  coreQueryStore,
+  disposeCoreQueryStore,
+  taskEverWorked,
+} from "../core-query-store";
+import {
+  appendEvent,
+  configureEventLogStore,
+  disposeEventLogStore,
+  getLastEventId,
+  readEventTail,
+} from "../event-log-store";
+import { CoreTaskWriter } from "../core-task-writer";
+import { readySessionOnAgentSpawn } from "../core-session-relaunch";
+
+// The other half of issue 387, against this Core's real SQLite and real event
+// log. The sweep moves a bare Session OFF `ready`; this is what puts it back
+// when a harness is spawned for it again — and what must not touch a Session
+// that actually worked.
+
+describe("putting a relaunched Session back on ready", () => {
+  let userDataDir: string;
+  let writer: CoreTaskWriter;
+
+  const insert = (taskId: string, status: string) => {
+    coreMutationStore.mutateTask({
+      op: "create",
+      taskId,
+      projectId: "p1",
+      title: taskId,
+      agent: "claude-code",
+      status: "ready",
+    });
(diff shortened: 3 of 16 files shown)


qcentic-dev:
## Reply to the round 1 review on #445

Review: https://github.com/actana/control/pull/445#pullrequestreview-5092086064 — `CHANGES_REQUESTED` on `e6bc759`. Every finding is addressed here; this branch carries that work plus the fixes below.

### Required

**1. A settled bare Session relaunches as `--resume` into a session id that never existed — fixed, with a caveat.**

`harnessLaunchMode`'s claude-code branch in `packages/panel/src/lib/harness-command.ts` is now gated on `claudeSessionId`, symmetric with the codex branch below it: nothing to resume into means start new. Pinned in `harness-command.test.ts` — no captured id at `disconnected` or `finished` starts new, a captured id still resumes, and a `ready` row is unchanged.

The caveat, because it would be dishonest to claim the symptom is gone: **the gate is inert at its one call site.** `baseCommandForTask` in `terminal-store.tsx` mints an id *before* it asks —

```ts
let sessionId = task.claudeSessionId;
if (!sessionId && task.agent !== "codex" && task.agent !== "opencode") sessionId = newSessionId();
const mode = harnessLaunchMode({ ...task, claudeSessionId: sessionId });
```

— so for claude-code the spread always carries a non-null id and the branch sees the same input it saw before. Codex and opencode are exempt from that mint, which is why their gates work. `terminal-store.tsx` is #380's file on this wave, so the mint is filed as **#447** with the two-line fix (ask, then mint) written out. The gate here is the correct half and is a prerequisite for it.

**2. A relaunched bare Session keeps reading `disconnected` while it is healthy — fixed.**

New `packages/core/src/core-session-relaunch.ts`, wired through a `relaunchPort` on the core-link server's spawn handler. On an **agent** spawn — never a `shell` or `shellSession` one — a row in a settled status that has **never worked** goes back to `ready`.

The "never worked" clause is the load-bearing one, and the row cannot answer it: `claude_session_id` is captured on `SessionStart`, which the Core installs for Claude Code, so a bare Session that never ran a turn can still carry an id. The event log can. Every status change on a Core-owned row appends a `task:updated` carrying the status that was *patched* (`core-task-writer.ts` is careful about exactly that distinction), and any status other than `ready` / `disconnected` says a turn happened — including on a harness that goes straight from `ready` to `finished` without ever reporting `running`. A `finished` Session being reopened is being resumed and keeps its card.

Tests: `core-session-relaunch.test.ts` (8 cases against real SQLite and a real event log — the bare Session moves, one that worked does not, one that worked and was then swept does not, live turns are untouched, the card event is appended, it is idempotent, a throwing write cannot take the spawn down) and `core-link-relaunch-port.test.ts` (which spawns ask the port at all).

**3. The docstring's safety argument is not true, and the real consequence is not stated — fixed.**

You are right that nothing prunes `event_log`. The "fails toward leaving it alone" claim is gone. The docstring now says the evidence is permanent and spells out the consequence: on the first Core boot after this ships, every historical spawned `ready` row settles to `disconnected` in one batch, each appending a `task:updated`; every later boot sees only what the run before it stranded. The conditional survives only as a conditional — *if* `event_log` ever grows a retention window, an aged-out row stops being swept, and that is the safe direction. The batch is now a section of its own in the PR description and is recorded in `docs/harness-status-detection.md`.

**4. `docs/harness-status-detection.md` — updated.**

The PTY-exit bullet (why `ready` is in scope, why `disconnected` and not `finished`), the boot-sweep bullet (why the evidence gate rather than a status filter, why the backstop deliberately does not see these rows, the one-time batch), a new relaunch-reset bullet beside them, the `(agent respawn) → ready` row in the status table, `core-session-relaunch.ts` in the file registry at the top, and a note under the terminal-input fallback that a Session parked on `ready` is no longer parked there forever.

### Non-blocking — taken

- **`disconnected` for both exit codes on a `ready` row.** Taken. `finished` is what `CoreTaskWriter` appends `session:finished` on, and a bare Session closed with `/exit` should not ring a completion ding. The two paths now agree: `harness-status-detection.test.ts`'s clean-exit case asserts the row reads `disconnected` and that no `session:finished` was appended, matching what the sweep test already said. #387's acceptance allows either.
- **Read the `shellSession` flag.** Taken — the evidence predicate now excludes `shellSession: true` spawns. Worth recording what it does *not* fix: a plain `shell: true` spawn records `shellSession: false`, the same as an agent, so the payload cannot separate it. It is separated by its task id instead, and the test now pins that shape directly (`cli_shell_<uuid>` and a user-terminal id, neither of which names a `tasks` row) rather than the weaker `task_id IS NULL` it pinned before. The docstring says all of this plainly instead of implying the flag covers both.
- **Rename the sweep dep.** Taken — `listActiveTasks` → `listBootSweepTasks` on `CoreSessionSweepDeps`, with the call site and the tests.
- **The 108-character line.** Wrapped.
- **The unreachable `log.warn` in `core-query-store.ts`.** Left as is, deliberately: as you note, `listActiveTasks` has the same dead handler, so it is the file's existing style and changing one of the two would make the pair inconsistent. `taskEverWorked`, added here, follows that same shape for the same reason. Worth its own pass across the store rather than a one-off in this PR.

### The `packages/core` flake

Filed as **#446**, with your reproduction data (2 of 6 full runs, 0 of 13 `packages/core`-only runs, 0 of 5 single-file runs under load), the `ENOENT` explanation, and the suggested fix — an explicit timeout to match the `20_000` its fd-leak sibling already carries, or a smaller tar. Independently reproduced here: one `packages/core` stage failure across three full-suite runs on the round 1 branch, green on every scoped run. Not touched by this PR.

### Verification

`pnpm typecheck`, `pnpm lint` (0 errors), `pnpm test` (all 6 stages: root 23, sdk 22+2 skipped, shared 31, cli 50+1 skipped, core 59, panel 146), and `commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD` (0 problems across all six commits).

Issues opened: **#446** (flake), **#447** (the session-id mint).

https://claude.ai/code/session_016RZuD4w44wp8ZMR3iMxzYB

mrfqcentic (changes requested):
Round 2 against #387 (F1) on `7387ea2`, diffed against `origin/fix/operator-pins-sessions-drop-cli` (`0b3b099`). Read the enforcing lines, not the description; re-ran all six suites in a scratch clone.

**All four required findings from round 1 are addressed, and the branch identity checks out.** `diff(5829f7f..2887bca)` and `diff(0b3b099..7387ea2)` are byte-identical, so the cherry-pick carried #445's tree with nothing added; of the six commit messages five are identical and the sixth (`1aba1e9` → `3111e03`) is the one that was rewrapped, its 133-character line now at 109. Every commit's body clears 132 characters (max 132), `Conventions` is green, and all 24 checks pass on the head. Base is the wave branch, it merges clean, it is still a draft, `terminal-store.tsx` and `__root.tsx` are untouched, and it does not go near the `foreign-session` path #390 owns (the two hunks in `harness-hook-pipeline.ts` are at 251 and 280; that path lives at 94/128–139/187/355).

Requesting changes on one thing, which is the load-bearing half of finding 2's fix.

---

### Required

**1. `packages/shared/src/core-query.ts:320` — `queryTaskEverWorked` answers "never worked" for every Session whose turns predate v0.4.0, so the reset overwrites a real `finished` card.**

The docstring says the row cannot answer the question and "what does answer it is the event log" (L299). That is true only of logs written by this version. `task:updated` began carrying the status its mutation patched in `2dd34a8` ("feat: await a session turn", 2026-08-21, shipped in v0.4.0); before it, `recordTaskMutation` wrote `JSON.stringify({ taskId, projectId })` and nothing else. `event_log` is `CREATE TABLE IF NOT EXISTS` and — as this PR itself establishes — is never pruned, so a Core upgraded from 0.3.x still holds every one of those status-less rows.

For such a row the `payload LIKE '%"status":"%'` clause at L327 matches nothing, `queryTaskEverWorked` returns `false`, and `readySessionOnAgentSpawn` (`core-session-relaunch.ts:76-77`) sees a settled status with no work behind it and writes `ready` over it. The Session that gets its card destroyed is exactly the one the PR promises to protect:

> A `finished` Session being reopened is being RESUMED, and its card must keep saying `finished` — the operator's own record of what that Session did.

And it does not self-correct. The reset appends `"status":"ready"`, the next boot sweep sees a `ready` row with a `pty:spawn` behind it and writes `"status":"disconnected"` — both excluded at L328-329 — so the row is classified never-worked permanently and is reset again on every reopen. The original `finished` is gone. It also feeds the launch mode: `harnessLaunchMode` gates resume on `status !== "ready"` for claude-code, codex and opencode alike, so once the card reads `ready` the next launch is `new`. That half is masked today only because `remoteTaskFromSnapshot` never copies `claudeSessionId` off a Core snapshot, which is the protocol gap #447 is queued to close on this same wave.

This is the same shape as round 1's finding 3: the docstring states a safety argument the data does not support, and it fails in the destructive direction rather than the conservative one.

Two cheap fixes, either is fine:

- Narrow `SETTLED_STATUSES` (`core-session-relaunch.ts:47`) to `disconnected` alone. Nothing is lost: before this PR `ready` was one-way, so a never-worked row could only ever be `ready`, and the only status this PR's own paths write for one is `disconnected`. `finished` / `terminated` / `interrupted` are then unreachable by the reset by construction, not by a log read.
- Or require positive evidence in `queryTaskEverWorked` — a row this PR settled always carries a `"status":"disconnected"` `task:updated`, so a row with *no* status-bearing `task:updated` at all is a legacy log, not a bare Session, and must answer "cannot tell → do not reset".

Whichever, the docstring should say that the read only sees status-bearing `task:updated` events, and that those start at v0.4.0.

---

### Verified resolved

**Finding 1 — the claude-code launch gate.** `harness-command.ts:38` is now `task.claudeSessionId && task.status !== "ready" ? "resume" : "new"`, character-for-character the codex branch at L51. `harness-command.test.ts` pins no-captured-id at `disconnected` and at `finished` → `new`, a captured id at `disconnected` → `resume`, and `ready` → `new`.

The caveat is accurate and I checked it rather than taking it: `baseCommandForTask` (`terminal-store.tsx:317-330`) mints before it asks, and the spread at L330 always hands claude-code a non-null id, so the gate cannot bite at its only call site. `remoteTaskFromSnapshot` (L296-308) confirms the other half — the return lists the server-authoritative fields explicitly and `claudeSessionId` is not among them, so a Core-owned row's Panel-side id is null regardless of what `SessionStart` captured on the Core. #447 states both halves correctly with the two-line fix written out.

I accept the split. `terminal-store.tsx` is #380's file this wave, the gate is the correct and necessary half, and the residual is bounded: a reopened swept bare Session takes one failed `--resume`, and `TerminalPane.tsx:1031-1070` catches a sub-threshold exit, patches a fresh id onto the Core row and relaunches with `[resume failed; starting a fresh Claude Code session]`. Noise on reopen, not a dead Session. Worth one line in #447 that the banner is what the operator sees until it lands.

**Finding 2 — the relaunch reset.** `core-session-relaunch.ts` + a `relaunchPort` on the spawn handler. The spawn gate at `pty-core-link-server.ts:1454` is `frame.opts.shellSession !== true && frame.opts.agent`, and the spawn union makes that exact: `HarnessSpawnRequest` requires `agent`, both `ShellSpawnRequest` and `ShellSessionSpawnRequest` type it `never`, so a shell spawn cannot reach the port by construction — and `core-link-relaunch-port.test.ts` pins all three plus the no-port Core. Nothing in the path branches on harness family, so the reset is genuinely every-family. `core-session-relaunch.test.ts` runs its 8 cases against real SQLite and a real event log. The mechanism is right; only the `everWorked` read above is not.

**Finding 3 — the docstring and the batch.** The "fails toward leaving it alone" claim is gone. `core-query.ts:246-256` now says the evidence is permanent, that no `DELETE FROM event_log` exists, and that the first boot after this ships settles every historical spawned `ready` row in one batch — and the PR body carries it as a section of its own. Correct on both counts.

**Finding 4 — `docs/harness-status-detection.md`.** +59/−2, covering every passage I named: the PTY-exit bullet (why `ready`, why `disconnected` and not `finished`), the boot-sweep bullet (why evidence and not a status filter, why the backstop stays narrow, the one-time batch), the status table row, the file registry, the terminal-input note, plus a new relaunch bullet.

**Non-blocking items.** `disconnected` on both exit codes, with `harness-status-detection.test.ts` asserting no `session:finished` on the clean exit — taken. `shellSession` read from the payload at `core-query.ts:275`, with `core-query.test.ts` pinning the real `recordPtySpawn` shape *and* the synthetic ids (`cli_shell_<uuid>`, a user-terminal id) instead of the weaker `task_id IS NULL` — taken, with an honest note about what it does not separate. `listActiveTasks` → `listBootSweepTasks` on the dep, call site and tests — taken. The 109-character line is now 92. The unreachable `log.warn` is declined with a stated reason (file-wide consistency, worth its own pass) — accepted; `taskEverWorked` follows the same shape deliberately.

---

### Non-blocking

- **`packages/shared/src/core-mutations.ts:519` — the one-time batch reorders the operator's lists.** Every status patch pushes `updated_at = ?`, and `queryTasks`, `queryArchivedTasks` and `queryStrandedReadyTasks` itself all order `updated_at DESC`. So the documented first-boot batch does more than move cards: every historical spawned `ready` row, archived ones included since the query deliberately spans them, gets stamped with the boot time and floats to the top of Fleet and Archived, burying recent work. The existing sweep already does this for `running` / `needs-input`, so it is not new in kind — but it is new in volume, it is not recoverable, and neither the PR body nor the docstring mentions it. One sentence beside the batch paragraph would cover it. Related: `core-session-sweep.ts:95-98` logs the whole `taskIds` array in one record, which for that same batch is one line holding potentially thousands of ids.
- **`packages/core/src/core-session-relaunch.ts:75-77` — the docstring's swallow does not cover the reads.** It promises "a write that throws is logged and swallowed: a card left one status stale is a far smaller failure than a spawn that answers `spawnError` because of it", but only `writer.mutate` is inside the `try`; `readTask` (L75) and `everWorked` (L77) are outside it, and `pty-core-link-server.ts:1454` calls the port inside the `try` whose `catch` sends `spawnError`. A throw from either read would answer `spawnError` for a PTY that is alive, and the client never learns the `ptyId`. Both wired implementations swallow internally today, so this is latent — but the port is typed as a plain callback and the contract is not enforced where the comment says it is. Moving the two reads inside the `try` costs nothing.
- **`packages/core/src/core-session-relaunch.ts:47` — `interrupted` in `SETTLED_STATUSES` is unreachable.** `queryTaskEverWorked` counts `interrupted` as a turn, so a row in that status always fails the `everWorked` check one line later. Harmless breadth, but it reads as a case the reset handles when it never can. (Narrowing the set as suggested above resolves this too.)
- **`packages/core/src/__tests__/core-session-sweep.test.ts:54-55` — the local `spawnPty` helper writes `JSON.stringify({ taskId })`, omitting `shellSession`.** It exercises the query correctly and `core-query.test.ts` pins the real `recordPtySpawn` shape, so nothing is untested — but the two helpers disagree about what the Core actually writes, and the sweep one would keep passing if the payload key were renamed.

---

### Suites

Full `pnpm test` in a scratch clone of `7387ea2`: **all six stages green** — root 23 files, sdk 22 (+2 skipped), shared 31/537, cli 50 (+1 skipped)/1196, core 59/920, panel 146. `pnpm typecheck` clean. `pnpm lint` is clean on source; the 546 errors my first run reported were all inside `packages/cli/dist-tarball/actana-cli.cjs`, a build artifact the test stage had just produced in the same tree — not a finding.

The `core-files-abort.test.ts` flake from round 1 did not reproduce here (core stage green, 59/59). It is filed as #446 with the reproduction data and is not this branch's.

CI on `7387ea2`: all 24 checks pass, none failing, the four train jobs and the two image-manifest publishes skipped as expected.

---

### Checklist

| | |
|---|---|
| Required 1 — claude-code gate on `claudeSessionId`, with a test | ✅ resolved; caveat verified accurate and accepted, filed as #447 |
| Required 2 — agent spawn resets a settled never-worked row, every family, with tests | ⚠️ mechanism resolved; **the `everWorked` read is wrong for pre-v0.4.0 logs** |
| Required 2 — a shell spawn does not | ✅ type-guaranteed and pinned |
| Required 3 — docstring on permanence + batch, and the PR body says it | ✅ resolved |
| Required 4 — `docs/harness-status-detection.md` in the named passages | ✅ resolved |
| `disconnected` on both exit codes, no `session:finished` | ✅ taken |
| `shellSession` from the payload, synthetic-id shape pinned | ✅ taken |
| Sweep dep renamed `listBootSweepTasks` | ✅ taken |
| Long test line wrapped | ✅ 109 → 92 |
| Unreachable `log.warn` declined with a reason | ✅ stated and accepted |
| Tree of `7387ea2` = tree of `e6bc759` + the fixes, nothing else | ✅ patch byte-identical to #445's final state |
| Every commit passes commitlint, bodies ≤ 132 | ✅ max 132; `Conventions` green |
| `terminal-store.tsx` / `__root.tsx` untouched | ✅ |
| Merges clean onto `origin/fix/operator-pins-sessions-drop-cli` | ✅ base is an ancestor of the head |
| Did not take #390's foreign-session Stop path | ✅ |
| Base is the wave branch, not `main` / `beta/0.4.5` | ✅ |
| Still a draft | ✅ |
| CI on `7387ea2` | ✅ 24/24 pass |
| Suites re-run in a scratch copy | ✅ 6/6 stages green |

**Verdict: changes requested.** One finding, and it is small: `queryTaskEverWorked` reads an event field that is two weeks old, treats its absence as proof, and on any Core upgraded from 0.3.x that turns the reset into an overwrite of the exact card the PR set out to protect. Narrow the settled-status set to `disconnected`, or make the read demand positive evidence. Everything else on this branch — all four required findings, every non-blocking item, the commit hygiene, the docs, the tests — is done and done carefully.

qcentic-dev:
## Reply to the round 2 review

Review: https://github.com/actana/control/pull/448#pullrequestreview-5093307864. The one required finding is fixed in `1886b5b`. Thank you for chasing `2dd34a8` down — the read was wrong in exactly the way you describe, and in the destructive direction.

### Required — `queryTaskEverWorked` answers "never worked" for pre-v0.4.0 logs

Both offered fixes are in, because they close different halves and neither is expensive.

**1. `SETTLED_STATUSES` is gone; the reset considers `disconnected` alone.** Your reasoning is the comment now: before this PR `ready` was one-way, so a Session that never worked could only ever *be* `ready`, and `disconnected` is the only status this PR's own settles write for one. `finished`, `terminated` and `interrupted` are unreachable for such a Session, so they are no longer candidates — by construction rather than by a log read being right about history it cannot see. That also retires the unreachable `interrupted` case from your non-blocking list.

**2. The read demands positive evidence.** A row these paths settled always carries a `"status":"disconnected"` `task:updated`, so a row with **no** status-bearing `task:updated` at all is a legacy log rather than a bare Session, and answers "cannot tell, do not reset". Renamed `queryTaskEverWorked` → `queryTaskProvenNeverWorked` (and `taskEverWorked` → `taskProvenNeverWorked`, dep `everWorked` → `provenNeverWorked`) so a `false` cannot be misread: it now means "worked, **or** unknowable", and the caller wants the same thing of both. The unreadable-log path returns `false` for the same reason, which flips it from permitting the reset to forbidding it.

The two guards are deliberately redundant, and the module header says so. The status check alone closes your case — a legacy `finished` row is never a candidate. The evidence check is what stops a legacy **`disconnected`** row, one that did work, from being reset as though it were bare; the status check gives no protection there at all.

**Docstrings and docs.** `queryTaskProvenNeverWorked`'s docstring now names `2dd34a8`, says the read sees only status-bearing `task:updated` events and that those start at v0.4.0, and explains why absence is not proof. The relaunch bullet in `docs/harness-status-detection.md` carries both halves, and the module header in `core-session-relaunch.ts` explains why the narrowing is a safety property and not a simplification.

**Tests.** `core-session-relaunch.test.ts` gains three cases, built on a `legacyUpdate` helper that writes the pre-v0.4.0 `{taskId, projectId}` payload directly so the 0.3.x history is real rather than mocked:

- a legacy **`finished`** row whose `task:updated` payloads carry no status is **not** reset on respawn — the case you named;
- a legacy **`disconnected`** row is not reset either, which is the one only the evidence check catches;
- `finished` / `terminated` / `interrupted` are never considered, whatever the log says.

`core-query.test.ts`'s suite is rewritten for the new contract: a pre-v0.4.0 log and a row with no history both answer `false`, a status-less update sitting *beside* real evidence still proves it, and `pty:spawn` is not evidence that the log carries statuses.

### Also taken from the round 2 non-blocking notes

- **The swallow does not cover the reads.** Fixed — `readTask` and `provenNeverWorked` are both inside the `try` now, and the docstring says why: the caller invokes this from the spawn handler's own `try`, whose `catch` sends `spawnError`, so a throw here would report a failed spawn for a live PTY and lose the client its `ptyId`.
- **`interrupted` unreachable in `SETTLED_STATUSES`.** Resolved by the narrowing above.
- **The batch reorders the operator's lists.** Taken. A sentence beside the batch paragraph in the PR body, and one in the `queryStrandedReadyTasks` docstring and the docs: a status patch stamps `updated_at`, Fleet, Archived and the sweep's own query all order by it, so every row the batch settles — archived ones included, since the query spans them — floats to the top at the boot time, above recent work, with no undoing it. Not new in kind, new in volume, as you say.
- **The sweep test's `spawnPty` helper omitted `shellSession`.** Fixed — it now writes the real `recordPtySpawn` shape (`{ptyId, taskId, shellSession: false}`), so the two helpers agree about what the Core actually writes and a renamed payload key would fail both.

### Left, with a reason

**`core-session-sweep.ts:95-98` logging the whole `taskIds` array in one record.** Real, and the first-boot batch is exactly when it bites. Left out of this branch deliberately: capping or sampling that line is a judgement about log shape that applies to the sweep's existing `running` / `needs-input` rows just as much as to the new ones, so it belongs in a change that can be reviewed on its own terms rather than tucked into a fix for an unrelated read. Happy to file it if you would like it tracked.

### Verification

- `commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD` — **0 problems** on all seven commits; every body line ≤ 132 characters measured in characters. Run before pushing, and the push was a plain fast-forward.
- `pnpm typecheck` clean; `pnpm lint` 0 errors; `pnpm test` **all 6 stages green** (root 23, sdk 22 + 2 skipped, shared 31, cli 50 + 1 skipped, core 59, panel 146).

Still a draft, base unchanged, nothing merged, `terminal-store.tsx` and `__root.tsx` untouched. Follow-ups remain **#446** (the `core-files-abort` flake) and **#447** (the session-id mint, plus your note that the `[resume failed; starting a fresh Claude Code session]` banner is what the operator sees until it lands — I will add that line to #447).

https://claude.ai/code/session_016RZuD4w44wp8ZMR3iMxzYB

mrfqcentic (approved):
Round 3 against #387 (F1) on `1886b5b`, diffed against `origin/fix/operator-pins-sessions-drop-cli` (`0b3b099`) and, for the round-2 delta, `7387ea2..1886b5b`. Read the enforcing lines, not the reply; re-ran all six suites in a scratch copy.

**The round 2 required finding is resolved, and resolved twice over.** Both offered remedies are in, they close different halves, and the module header says so rather than leaving a reader to work it out. Approving.

---

### Verified resolved — `queryTaskProvenNeverWorked` and the narrowed status

**The status narrowing.** `SETTLED_STATUSES` is gone; `core-session-relaunch.ts:74` is `const RESETTABLE_STATUS = "disconnected"` and the gate at L107 is `task.status !== RESETTABLE_STATUS`. So `finished`, `terminated` and `interrupted` are unreachable by the reset by construction — no log read stands between a real `finished` card and an overwrite, because such a card is never a candidate. I checked the premise rather than taking it: the only Core paths that settle a never-worked row are `harness-hook-pipeline.ts:280` (`ready → disconnected`, either exit code) and `core-session-sweep.ts:44` (`STRANDED_STATUS = "disconnected"`); `core-session-backstop.ts:161` writes `finished`/`disconnected` but reads the narrow `listActiveTasks`, which is `running`/`needs-input` only (`core-query.ts:200`), so it never sees a bare row. The Panel's unconditional `finished`/`terminated` patch on PTY exit (`TerminalPane.tsx:1087`) is gated on `!descriptor.coreId`, and `sweepOrphanedActiveTasks` (`services/tasks.ts:132`) is local-scope. Nothing writes those three statuses for a Core-owned never-worked row. The narrowing loses nothing, and it retires the unreachable `interrupted` case from my round 2 list.

**The positive-evidence read.** `core-query.ts:339` now runs two reads: at least one status-bearing `task:updated` must exist before the "did it work" question is even asked, and `false` — the answer that forbids the reset — is returned for no evidence, no history, and an unreadable log alike. The rename to `queryTaskProvenNeverWorked` / `taskProvenNeverWorked` / `provenNeverWorked` is carried through every call site and the dep type (`core-entry.ts:430`, `core-query-store.ts:222`); no `everWorked` identifier survives anywhere in the tree. The `LIKE` patterns still cannot straddle fields — a title containing `"status":"` is JSON-escaped to `\"status\":\"` and does not match.

**The tests pin the pre-v0.4.0 shape for real, not by mock.** `core-session-relaunch.test.ts:54-56`'s `legacyUpdate` appends `JSON.stringify({ taskId, projectId })` straight into the event log, which is byte-for-byte what `recordTaskMutation` wrote before `2dd34a8`. The three new cases are the right three: a legacy `finished` row is not reset (the case I filed), a legacy `disconnected` row is not reset (the case only the evidence check catches, since the status check gives no protection there at all), and `finished`/`terminated`/`interrupted` are never considered. `core-query.test.ts:480-548` is rewritten for the inverted contract, including a status-less update sitting beside real evidence still proving it, and `pty:spawn` not counting as evidence that the log carries statuses.

**And the bare Session still gets its card back.** `core-session-relaunch.test.ts:86` — a row created `ready`, settled `disconnected`, respawned — returns `true` and lands on `ready`. That settle writes the status-bearing `task:updated` the read now demands, so the evidence check does not cost the PR its own feature.

**Docstrings and docs.** `queryTaskProvenNeverWorked`'s docstring names `2dd34a8`, says the read sees only status-bearing `task:updated` and that those start at v0.4.0, and explains why absence is not proof. `core-session-relaunch.ts:14-45` carries the same and states plainly that the two guards are deliberately redundant. `docs/harness-status-detection.md:368-397` carries both halves. All three agree with the code.

**Also taken from the round 2 non-blocking list.** Both reads are inside the `try` now, with the reason (`spawnError` for a live PTY) written where the promise is made. The batch's `updated_at` restamp is stated in the PR body, the `queryStrandedReadyTasks` docstring and the docs. The sweep test's `spawnPty` helper writes the real `{ptyId, taskId, shellSession: false}` shape. The `core-session-sweep.ts:95` log line is declined with a stated reason I accept — it is the existing sweep's line as much as the new rows'.

**The round 3 diff touches nothing else.** `7387ea2..1886b5b` is 8 files, all of them `packages/core`, `packages/shared` or `docs`: the relaunch module and its test, the query and its test, the two wiring files, the sweep test helper, and the doc. One commit, `Refs #387`.

---

### Non-blocking

- **`core-query.ts:339` — the settle can manufacture the evidence the read demands.** The invariant the docstring states is "a row with no status-bearing `task:updated` at all is a legacy log ⇒ cannot tell". True at rest; but this PR's own settle writes the first status-bearing update for exactly such a row. Reachable path: a Core upgraded from 0.3.x whose *last* 0.3.x shutdown stranded a working Session in `running`/`needs-input`; the first 0.4.x boot sweep patched it to `disconnected` and that is the only status-bearing event it has, its real turns being pre-`2dd34a8` and invisible. Reopening it after this ships reads "proven never worked" and puts the card on `ready`. It is narrow — a row already `disconnected` on 0.3.x never re-enters the resettable status, and a row swept by a 0.3.x boot carries no status-bearing event at all, so both are safe — and it is nothing like the round 2 finding: the card destroyed is a sweep's `disconnected`, not the operator's `finished`, and the first turn writes `running` and settles the classification for good. Worth a sentence in the docstring saying the evidence can be the settle's own, not a fourth round.
- **`harness-command.ts:29-31` — the comment's reason is wrong even though the code is right.** It says a bare Session "never had a session id captured for it". The Core installs `SessionStart` (`harness-hooks.ts:190`) and the pipeline persists the id through `setSessionId` (`core-harness-status.ts:61`) — this PR's own `core-query.ts:303` docstring says exactly that, as the reason the row cannot answer the question. So a bare Session usually *does* carry an id, and the gate's first half is not what saves it. What actually happens is the two things I checked in round 2 and re-checked here: `remoteTaskFromSnapshot` (`terminal-store.tsx:291`) hard-codes `claudeSessionId: null` for a Core row, and `baseCommandForTask` (L317-329) mints one before asking, so the gate reduces to `status !== "ready"` at its only call site either way. Note also that the Panel runs the same shared pipeline for its own rows (`hooks.controller.ts:63`), so a **Panel-local** bare Session — which does hold a real captured id — now reaches `disconnected` too and resumes into a zero-turn conversation. Same bounded outcome as round 2: one failed `--resume`, the `TerminalPane.tsx:1031-1070` catch, a fresh session and a banner. #447 owns the fix; it should own this corrected reason too, and the comment should stop claiming an id was never captured.
- **`harness-hook-pipeline.ts:280` — the settle cannot tell which PTY exited.** `onSessionExit` carries `{taskId, exitCode}` and no `ptyId` (`pty-manager.ts:756`), so a late exit from a previous PTY of the same Session settles whatever the row currently says. Pre-existing in shape — the same is true of the `running`/`needs-input` arm — but `ready` rows were immune to it before this change, and now that the relaunch writes `ready` back on a fresh spawn there is a (narrow) window where a stale exit puts `disconnected` over a live harness with nothing to correct it until the first prompt. Needs two PTYs for one Session overlapping; not worth holding the branch for, worth knowing.
- **The one-time batch feeds a one-click archive.** Every `ready` row the Panel created by opening a pane carries a `pty:spawn`, so the first-boot batch settles essentially all historical bare Sessions to `disconnected` — and `projects.$id.tsx:2326-2334` puts an "Archive all disconnected sessions" button over that group. Combined with the `updated_at` restamp already documented, one operator click after the upgrade can archive a fleet of Sessions they meant to start. The reordering is in the release note; this deserves the same sentence.

---

### Suites, CI, hygiene

Full `pnpm test` in a scratch copy of `1886b5b`: **six stages green** — root 23 files/755, sdk 22 (+2 skipped)/292, shared 31/539, cli 50 (+1 skipped)/1195, core 59/923, panel 146/1570. Two caveats, both mine and not the branch's: the root stage needs a git dir (my first copy had none, and `git ls-files` failed in `version-agreement`/`base-pins`; green once restored), and `in-process-core-session.test.ts` timed out once at 60 s under the load of the full run, then passed 10/10 standalone in 4.6 s — the CLI package is untouched by this PR and CI's Unit Tests job is green on this head.

`pnpm typecheck` clean. `pnpm lint` 0 errors on source (45 pre-existing warnings); the 546 errors a naive run reports are all inside `packages/cli/dist-tarball/actana-cli.cjs`, the artifact the test stage had just built — same as round 2.

Commit hygiene: all seven commits check out against `commitlint.config.mjs` rule by rule — conventional header, ≤120 chars, no full stop, body leading blank, trailer leading blank, and every body line ≤132 measured in **characters** (max 132; the head commit's own max is 121). `Conventions` is green on the head.

CI on `1886b5b`: **24 checks, 18 pass and 6 skipped** as expected (both image manifests, the two train image jobs, train tags, train versions). None failing. `mergeStateStatus` is `CLEAN` and the base is an ancestor of the head, so it fast-forwards onto `origin/fix/operator-pins-sessions-drop-cli`. Base is the wave branch, and it is still a draft.

---

### Checklist

| | |
|---|---|
| Round 2 required — a legacy `finished` row whose updates carry no status is never reset | ✅ both guards, pinned with a real 0.3.x payload |
| `finished` / `terminated` / `interrupted` unreachable by construction | ✅ `RESETTABLE_STATUS = "disconnected"`, premise verified against every writer |
| A genuinely never-worked `disconnected` row still returns to `ready` on an agent spawn | ✅ pinned; the settle supplies the evidence the read wants |
| A legacy `disconnected` row that did work is not reset | ✅ the case only the evidence check catches, pinned |
| Tests pin the pre-v0.4.0 payload shape | ✅ `legacyUpdate` writes `{taskId, projectId}` into the log directly |
| Docstring + `docs/harness-status-detection.md` match the code | ✅ both name `2dd34a8` and the v0.4.0 floor |
| Reads moved inside the `try` | ✅ with the reason stated |
| Batch `updated_at` restamp disclosed | ✅ PR body, docstring and docs |
| Sweep test helper writes the real `recordPtySpawn` shape | ✅ |
| `7387ea2..1886b5b` touches nothing else | ✅ 8 files, core/shared/docs only |
| Every commit passes commitlint, bodies ≤ 132 characters | ✅ max 132; `Conventions` green |
| `terminal-store.tsx` / `__root.tsx` untouched | ✅ |
| Merges clean onto `origin/fix/operator-pins-sessions-drop-cli` | ✅ base is an ancestor |
| Base is the wave branch, still a draft | ✅ |
| CI on `1886b5b` | ✅ 18 pass, 6 skipped, 0 failing |
| Suites re-run in a scratch copy | ✅ 6/6 green |

**Verdict: approved.** The read that was wrong is now wrong in the safe direction, and the status set makes the destructive case unreachable without trusting the read at all. Four non-blocking notes, none of which should hold the branch: the evidence check can be satisfied by this PR's own settle on one narrow upgrade path, the launch-gate comment states a reason that is factually wrong (the code is right; #447 owns the fix), the exit settle still cannot name the PTY it is settling, and the first-boot batch lands a large group under a one-click "Archive all". Everything the last two rounds asked for is done, and done with the reasoning written down where the next reader will need it.

Issue #384, opened 2026-09-02
B2: setActiveGroup applies a stale server answer over the newer selection


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `B2`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: panel-rendering-investigation-2026-09-02.md (B2)

## What the operator sees

Two rapid group changes (or racing tabs) revert the rail to an older group mid-click.

## Where

`packages/panel/src/lib/active-group.ts` (investigation: 74–89) writes optimistically then `setQueryData` from whichever `updateSettings` resolves, with no nonce.

## What to do

Ignore a settings response whose generation is older than the latest local write. Same pattern as other optimistic stores in this Panel.

## Acceptance

1. Rapid group A then B leaves B selected even if A's PATCH returns last.
2. A failed PATCH still restores the last acknowledged group, not a random older one.

PR #449, opened 2026-09-02
fix(panel): Enter after settlement must not post running

### LINKED ISSUE #386: A-enter: Enter after settlement posts running again for hooks that omit turn-start

qcentic-dev:
Closes #386 — hunt ID `A-enter`, the pin/session/CLI/drop wave.

## The bug

The Enter→running fallback exists for a Session whose hooks never announce a turn's **start**: Cursor CLI always, Codex until its
hooks are reviewed. Both report a turn's end; neither reports its beginning, so the Panel infers the beginning from the terminal.

It inferred it from too little. The one-shot latch re-armed on `shouldResetTerminalRunningFallback` alone — "the row is not
currently running" — and the turn-start test was `data.includes("\r") || data.includes("\n")`. Once a Session settled, **any**
terminal input carrying a newline PATCHed the row back to `running`: a stray Enter at the idle prompt, a pasted path, a drop-path
write. The card then lied about a Session that had finished.

## The fix

Gate the fallback on a fresh turn-start signal, not on "the task is not running".

- `advanceTerminalTurn(state, data)` folds each `onData` chunk into what the operator has composed in this pane, and reports
  `submittedTurn` only when a newline **submits text they actually entered**.
- A bare Enter submits nothing — the composition is empty.
- A bracketed paste (`ESC[200~` … `ESC[201~`) is the terminal inserting characters, not pressing Return. A pasted path — even a
  multi-line one — composes without submitting; the Enter the operator presses afterwards submits it, as it should. A path
  dropped from the Panel's own UI never reaches `onData` at all (it is written straight to the pty).
- Terminal-generated replies (focus reports, device attributes, cursor-position answers) are skipped through the existing
  `isTerminalAutoReply` classifier; escape sequences are stepped over whole; backspace pops, so a fully erased line submits nothing.
- `advanceTerminalRunningFallback` now owns both gates — the per-turn one-shot and the turn state — so `TerminalPane`'s `onData`
  handler is a single call, and the regression is testable without dragging xterm into jsdom.

Acceptance 2 holds by construction: a harness that never reports turn start still posts `running` the moment the operator types a
prompt and presses Enter, and the latch still re-arms across turns so every genuine prompt updates the card.

## Tests

- `packages/panel/src/lib/__tests__/terminal-running-fallback.test.ts` (new) — the operator's story, chunk by chunk: Enter on a
  finished Session posts nothing; a pasted path posts nothing; a real prompt after settlement posts `running`; one post per turn
  while running; re-arming across turns; hooks-report-turn-start stands the fallback down; a failed PATCH is retried.
- `packages/panel/src/lib/__tests__/task-status-sync.test.ts` — extended for the composition tracker (submit vs. bare Enter,
  bracketed paste including one split across chunks, escape sequences, backspace, submit-then-keep-typing).

Whole Panel suite green: **147 files, 1583 tests**. `pnpm typecheck` and `eslint` clean.

## Base

Targets `fix/operator-pins-sessions-drop-cli` (the wave branch), cut from `0b3b099` which already carries the merged A-heal
change from #385. Not `main`, not `beta/0.4.5`. Left as a draft.

Scope kept to `TerminalPane.tsx` + `task-status-sync.ts` and their tests. T2 (#393, two tabs typing / `onData` disposal) also
lives in `TerminalPane.tsx` and is **not** taken here — it is a later wave.

https://claude.ai/code/session_01C2jjKPy8zd5bV6NU7Q7GzR

### DIFF
--- a/packages/panel/src/components/views/TerminalPane.tsx
+++ b/packages/panel/src/components/views/TerminalPane.tsx
@@ -63,6 +63,7 @@ import {
   advanceTerminalRunningFallback,
   harnessUsesTerminalPromptFallback,
   IDLE_TERMINAL_RUNNING_FALLBACK,
+  noteTerminalWrite,
   type TerminalRunningFallback,
 } from "~/lib/task-status-sync";
 import { accumulateTerminalPrompt } from "~/lib/terminal-prompt-capture";
@@ -979,8 +980,9 @@ export function TerminalPane({
         null;
       // The Enter→running fallback for a Session whose hooks never announce a
       // turn's start: the one-shot latch AND the turn the operator is composing
-      // in this pane (issue 386). Rebuilt with the surface, like every other
-      // per-pane scratch value here.
+      // in this pane (issue 386). Declared with the surface but reset per pty
+      // in `wireTerminalInput` — a half-typed line, or a `pasting` latch, must
+      // not survive into the next harness process.
       let runningFallback: TerminalRunningFallback = IDLE_TERMINAL_RUNNING_FALLBACK;
       let promptCaptureBuffer = "";
       let promptTitlePosted = false;
@@ -1003,6 +1005,13 @@ export function TerminalPane({
       const writeToPty = async (data: string) => {
         const ptyId = activePtyId;
         if (!ptyId || !ptyApi || !mayWriteRef.current) return false;
+        // Every byte on this path skips xterm's keyboard, so `onData` never
+        // sees it: the dropped project path, the key map's Cmd+Backspace. Fold
+        // it into the same turn the fallback watches, or a dropped path
+        // followed by Enter starts a real turn against a card still reading
+        // `finished` (issue 386). It composes only — the Enter that submits it
+        // comes back through `onData`, which is where the PATCH is decided.
+        runningFallback = noteTerminalWrite(runningFallback, data);
         return ptyApi.write(ptyId, data);
       };
 
@@ -1215,6 +1224,13 @@ export function TerminalPane({
       };
 
       const wireTerminalInput = (ptyId: string) => {
+        // A new pty is a new harness process at a fresh prompt. Anything half
+        // entered against the last one is gone from the screen and must go from
+        // the pane's mirror of it too: a surviving composition makes the first
+        // stray Enter post `running`, and a surviving `pasting` latch swallows
+        // every `\r` for the life of the pane.
+        runningFallback = IDLE_TERMINAL_RUNNING_FALLBACK;
+        promptCaptureBuffer = "";
         term.onData((data) => {
           // A Reader sends nothing and reports nothing (issue 147). The return
           // is before the status and prompt-capture side effects deliberately:
@@ -1263,6 +1279,17 @@ export function TerminalPane({
                   taskId: descriptor.taskId,
                   status: "running",
                 });
+              } catch {
+                // Only the status mutation un-latches, and only its own catch
+                // may do it. Sharing one `try` with the title and the cache
+                // invalidations below meant a title-generator hiccup AFTER a
+                // successful PATCH cleared the latch anyway — and, because this
+                // handler re-reads the live value, a late failure from turn N
+                // could clear a latch turn N+1 had just set.
+                runningFallback = { ...runningFallback, posted: false };
+                return;
+              }
+              try {
                 // The prompt patches no column of its own: it only asks a
                 // title generator to name the session. Which generator depends
                 // on who owns the row — the Core's, for a Core-owned Session
@@ -1287,9 +1314,8 @@ export function TerminalPane({
                   queryClient.invalidateQueries({ queryKey: queryKeys.projects }),
                 ]);
               } catch {
-                // The PATCH never landed — un-latch so the operator's next
-                // submitted turn tries again.
-                runningFallback = { ...runningFallback, posted: false };
+                // The row is already `running`; a missed title or a stale cache
+                // is not worth un-latching a turn that did start.
               }
             })();
           }
--- a/packages/panel/src/components/views/TerminalPane.tsx
+++ b/packages/panel/src/components/views/TerminalPane.tsx
@@ -60,9 +60,10 @@ import {
 } from "~/lib/harness-command";
 import { getDefaultModelForHarness } from "~/lib/default-model-store";
 import {
-  terminalInputStartsTurn,
+  advanceTerminalRunningFallback,
   harnessUsesTerminalPromptFallback,
-  shouldResetTerminalRunningFallback,
+  IDLE_TERMINAL_RUNNING_FALLBACK,
+  type TerminalRunningFallback,
 } from "~/lib/task-status-sync";
 import { accumulateTerminalPrompt } from "~/lib/terminal-prompt-capture";
 import { prefetchTerminalModules } from "~/lib/prefetch-terminal-modules";
@@ -976,7 +977,11 @@ export function TerminalPane({
       let duringReplayData: SequencedPtyData[] = [];
       let duringReplayExit: { ptyId: string; exitCode: number; signal?: number } | null =
         null;
-      let fallbackRunningPosted = false;
+      // The Enter→running fallback for a Session whose hooks never announce a
+      // turn's start: the one-shot latch AND the turn the operator is composing
+      // in this pane (issue 386). Rebuilt with the surface, like every other
+      // per-pane scratch value here.
+      let runningFallback: TerminalRunningFallback = IDLE_TERMINAL_RUNNING_FALLBACK;
       let promptCaptureBuffer = "";
       let promptTitlePosted = false;
       const stopWatchingColorScheme = watchTerminalColorScheme((colorScheme) => {
@@ -1234,15 +1239,21 @@ export function TerminalPane({
             submittedPrompt = captured.submitted;
           }
 
-          // Cursor CLI still does not fire beforeSubmitPrompt, so Enter is the
-          // per-turn running signal. Re-arm after the task leaves "running"
-          // (stop → finished, needs-input, etc.) so a second prompt in the same
-          // session updates the card again.
-          if (shouldResetTerminalRunningFallback(liveTaskStatusRef.current)) {
-            fallbackRunningPosted = false;
-          }
-          if (!fallbackRunningPosted && terminalInputStartsTurn(task.agent, data, hooksReportTurnStart)) {
-            fallbackRunningPosted = true;
+          // Cursor CLI still does not fire beforeSubmitPrompt, so a submitted
+          // prompt is the per-turn running signal. "Submitted" is the whole
+          // point of issue 386: the latch re-arms once the task leaves
+          // "running" (stop → finished, needs-input, …) so a second prompt in
+          // the same session updates the card again — but on its own that made
+          // every newline after settlement a new turn, so a stray Enter or a
+          // pasted path resurrected `running`. The fallback now waits for the
+          // operator to actually enter something and submit it.
+          const fallbackStep = advanceTerminalRunningFallback(runningFallback, {
+            data,
+            currentStatus: liveTaskStatusRef.current,
+            hooksReportTurnStart,
+          });
+          runningFallback = fallbackStep.state;
+          if (fallbackStep.postRunning) {
             void (async () => {
               try {
                 // Same routing as the exit patch above: a turn-start status is
@@ -1276,7 +1287,9 @@ export function TerminalPane({
                   queryClient.invalidateQueries({ queryKey: queryKeys.projects }),
                 ]);
               } catch {
-                fallbackRunningPosted = false;
+                // The PATCH never landed — un-latch so the operator's next
+                // submitted turn tries again.
+                runningFallback = { ...runningFallback, posted: false };
               }
             })();
           }
--- a/packages/panel/src/lib/__tests__/task-status-sync.test.ts
+++ b/packages/panel/src/lib/__tests__/task-status-sync.test.ts
@@ -1,31 +1,43 @@
 import { describe, expect, it } from "vitest";
 import {
+  advanceTerminalTurn,
   harnessUsesTerminalPromptFallback,
+  IDLE_TERMINAL_TURN,
   shouldResetTerminalRunningFallback,
   terminalInputStartsTurn,
 } from "../task-status-sync";
 
+/** Fold a script of onData chunks, keeping every submit the run produced. */
+function typeInto(chunks: string[]) {
+  let state = IDLE_TERMINAL_TURN;
+  const submits: boolean[] = [];
+  for (const chunk of chunks) {
+    const step = advanceTerminalTurn(state, chunk);
+    state = step.state;
+    submits.push(step.submittedTurn);
+  }
+  return { state, submits, submitted: submits.some(Boolean) };
+}
+
 describe("terminal status sync", () => {
   it("stands the fallback down only for a Session whose hooks actually went in", () => {
     // The Core installed them for this spawn — the hooks report the turn.
-    expect(terminalInputStartsTurn("claude-code", "\r", true)).toBe(false);
-    expect(terminalInputStartsTurn("cursor-cli", "implement this\r", true)).toBe(false);
+    expect(terminalInputStartsTurn(true, true)).toBe(false);
   });
 
   it("keeps the fallback armed for a hook-capable harness whose hooks did NOT go in", () => {
     // The regression this rule exists for (issue 84): claude-code is a family
     // that supports hooks, so the old family-set check exempted it — leaving a
     // Session with neither hooks nor a fallback, stuck on `ready` forever.
-    expect(terminalInputStartsTurn("claude-code", "\r", false)).toBe(true);
-    expect(terminalInputStartsTurn("opencode", "\r", false)).toBe(true);
+    expect(terminalInputStartsTurn(false, true)).toBe(true);
+    expect(terminalInputStartsTurn(false, false)).toBe(false);
   });
 
   it("marks input-driven agents as running when the user submits input", () => {
     expect(harnessUsesTerminalPromptFallback("cursor-cli")).toBe(true);
     expect(harnessUsesTerminalPromptFallback("codex")).toBe(false);
-    expect(terminalInputStartsTurn("cursor-cli", "hello", false)).toBe(false);
-    expect(terminalInputStartsTurn("cursor-cli", "implement this\r", false)).toBe(true);
-    expect(terminalInputStartsTurn("codex", "\r", false)).toBe(true);
+    expect(typeInto(["hello"]).submitted).toBe(false);
+    expect(typeInto(["implement this\r"]).submitted).toBe(true);
   });
 
   it("re-arms the Enter→running fallback after a turn leaves running", () => {
@@ -35,5 +47,64 @@ describe("terminal status sync", () => {
     expect(shouldResetTerminalRunningFallback("finished")).toBe(true);
     expect(shouldResetTerminalRunningFallback("needs-input")).toBe(true);
     expect(shouldResetTerminalRunningFallback("ready")).toBe(true);
+  });
+});
+
+describe("advanceTerminalTurn", () => {
+  it("submits only a newline that terminates composed text", () => {
+    expect(typeInto(["fix ", "login", "\r"]).submitted).toBe(true);
+    // A bare Enter at an empty prompt enters nothing (issue 386).
+    expect(typeInto(["\r"]).submitted).toBe(false);
+    expect(typeInto(["\n"]).submitted).toBe(false);
+    expect(typeInto(["\r", "\r", "\n"]).submitted).toBe(false);
+    // Whitespace is not something the operator entered either.
+    expect(typeInto(["   \r"]).submitted).toBe(false);
+  });
+
+  it("clears the composition on submit, so the next Enter starts nothing", () => {
+    const run = typeInto(["ship it\r", "\r"]);
+    expect(run.submits).toEqual([true, false]);
+    expect(run.state.composed).toBe("");
+  });
+
+  it("treats a bracketed paste as text, never as a submit", () => {
+    // xterm wraps a paste in ESC[200~ … ESC[201~ whenever the harness has
+    // bracketed paste on, which every TUI harness here does. Pasting is the
+    // terminal inserting characters; it is not pressing Return — so a pasted
+    // path, even one carrying newlines, starts no turn.
+    const path = typeInto(["\x1b[200~/home/core/repos/control\x1b[201~"]);
+    expect(path.submitted).toBe(false);
+    expect(path.state.composed).toBe("/home/core/repos/control");
+    expect(typeInto(["\x1b[200~one\ntwo\n\x1b[201~"]).submitted).toBe(false);
+    // …and the Enter the operator presses afterwards submits what was pasted.
+    const state = advanceTerminalTurn(IDLE_TERMINAL_TURN, "\x1b[200~./notes.md\x1b[201~").state;
+    expect(advanceTerminalTurn(state, "\r").submittedTurn).toBe(true);
+  });
+
+  it("survives a paste split across chunks", () => {
+    const run = typeInto(["\x1b[200~/home/core", "/repos\x1b[201~", "\r"]);
+    expect(run.submits).toEqual([false, false, true]);
+  });
+
+  it("ignores terminal-generated replies and cursor keys", () => {
+    // Focus reports and device-attribute answers ride onData too; none of them
+    // is the operator entering anything.
+    expect(typeInto(["\x1b[I"]).state.composed).toBe("");
+    expect(typeInto(["\x1b[?62;c"]).state.composed).toBe("");
+    const arrows = typeInto(["hi", "\x1b[A", "\x1b[B", "\r"]);
+    expect(arrows.state.composed).toBe("");
+    expect(arrows.submitted).toBe(true);
+  });
+
+  it("honours backspace, so a fully erased line submits nothing", () => {
+    expect(typeInto(["ab\x7f\x7f", "\r"]).submitted).toBe(false);
+    expect(typeInto(["ship\x7f it", "\r"]).submitted).toBe(true);
+    expect(typeInto(["ship\x7f"]).state.composed).toBe("shi");
+  });
+
+  it("handles a chunk that submits and then keeps typing", () => {
+    const step = advanceTerminalTurn(IDLE_TERMINAL_TURN, "go\rnext");
+    expect(step.submittedTurn).toBe(true);
+    expect(step.state.composed).toBe("next");
   });
 });
--- a/packages/panel/src/lib/__tests__/task-status-sync.test.ts
+++ b/packages/panel/src/lib/__tests__/task-status-sync.test.ts
@@ -86,12 +86,27 @@ describe("advanceTerminalTurn", () => {
     expect(run.submits).toEqual([false, false, true]);
   });
 
+  it("holds a paste marker cut in half by a chunk boundary", () => {
+    // The opening marker needs all six bytes in one chunk to be recognised.
+    // Split at `ESC[200`, the `200` used to land in the composition as literal
+    // text and the pane never entered the paste at all.
+    const opening = typeInto(["\x1b[200", "~/a/b\x1b[201~"]);
+    expect(opening.state.composed).toBe("/a/b");
+    expect(opening.state.pasting).toBe(false);
+    expect(opening.submitted).toBe(false);
+    // The closing marker matters more: miss it and the pane pastes for ever,
+    // swallowing every Enter for the life of the pane.
+    const closing = typeInto(["\x1b[200~/a/b\x1b[201", "~", "\r"]);
+    expect(closing.submits).toEqual([false, false, true]);
+    expect(closing.state.pasting).toBe(false);
+  });
+
   it("ignores terminal-generated replies and cursor keys", () => {
     // Focus reports and device-attribute answers ride onData too; none of them
     // is the operator entering anything.
     expect(typeInto(["\x1b[I"]).state.composed).toBe("");
     expect(typeInto(["\x1b[?62;c"]).state.composed).toBe("");
-    const arrows = typeInto(["hi", "\x1b[A", "\x1b[B", "\r"]);
+    const arrows = typeInto(["hi", "\x1b[C", "\x1b[D", "\r"]);
     expect(arrows.state.composed).toBe("");
     expect(arrows.submitted).toBe(true);
   });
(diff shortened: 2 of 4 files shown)


qcentic-dev:
## Acceptance checklist — issue #386 (`A-enter`)

- [x] **1. After a session is finished, pressing Enter (or pasting a path) does not PATCH it to running.**
  The turn-start signal is no longer "these bytes contain a CR or an LF". `advanceTerminalTurn` tracks what the operator has
  composed in the pane and reports a turn start only when a newline **submits** text they actually entered, so a bare Enter at
  the idle prompt of a finished Session composes nothing and submits nothing. A pasted path arrives bracketed
  (`ESC[200~` … `ESC[201~`); a bracketed paste is the terminal inserting characters, never pressing Return, so it composes
  without submitting — true even for a multi-line paste. A path dropped from the Panel's own UI never reaches `onData` at all
  (`wireTerminalFileDrop` writes straight to the pty), so it was never a route to this PATCH and still isn't.
  Proved by `terminal-running-fallback.test.ts`: *"does not PATCH running when the operator presses Enter on a finished
  Session"* and *"does not PATCH running when a path is pasted into a finished Session"*.

- [x] **2. Harnesses that never report turn start still go running when the operator actually starts a turn.**
  `hooksReportTurnStart === false` still arms the fallback exactly as before — the issue-84 rule is untouched — and the latch
  still re-arms as soon as the row leaves `running`, so every genuine prompt updates the card. What changed is only what counts
  as a start: the operator typing a prompt (pasted fragments included) and pressing Enter.
  Proved by `terminal-running-fallback.test.ts`: *"still PATCHes running when the operator actually starts a new turn"*,
  *"re-arms across turns, so every genuine prompt updates the card"*, *"posts once per turn while the Session is running"*, and
  *"lets a failed PATCH be retried by the next submitted turn"*.

- [x] **Sequenced after A-heal (#385), as the issue asks.** Branched from `0b3b099` on `fix/operator-pins-sessions-drop-cli`,
  which already carries the merged A-heal change. No conflict: A-heal touched neither `TerminalPane.tsx` nor `task-status-sync.ts`.

- [ ] **T2 (#393 — two tabs typing, `onData` disposal) — deliberately NOT taken.**
  Unchecked on purpose: it is a later wave, and it is not an acceptance criterion of #386. It shares this file
  (`TerminalPane.tsx`), which is why the issue notes the overlap, but the work stays out of this PR.

- [ ] **Clicked through on pairdemo — not done.**
  Unchecked on purpose: #386 is filed `Live on pairdemo: code-read` (the Panel UI was not clicked in the catalog hunt), and this
  worktree has no pairdemo Session to drive. The evidence here is the test suite, which replays the pane's `onData` handler
  chunk by chunk with the row's status as the Core reported it.

### Boundaries respected

Touched only `packages/panel/src/components/views/TerminalPane.tsx`, `packages/panel/src/lib/task-status-sync.ts` and their
tests. Not touched: `active-group.ts` (#384), `packages/shared/harness-hook-pipeline.ts` / `core-query.ts` /
`core-session-relaunch.ts` and `packages/panel/src/lib/harness-command.ts` (#387, PR #448), `terminal-store.tsx`,
`ProjectBar.tsx`, `__root.tsx`, `projects.$id.tsx`. Nothing merged, `main` and `beta/0.4.5` untouched, no force push, PR left as
a draft.

### Local verification

`npx vitest run --root packages/panel` — **147 files, 1583 tests, all passing.** `pnpm typecheck` and `eslint` clean on the
changed files. The commit message was linted against this repo's own `commitlint.config.mjs` with a throwaway
`@commitlint/cli` install, exactly as the Conventions job does — exit 0, every body line ≤ 132 characters.

https://claude.ai/code/session_01C2jjKPy8zd5bV6NU7Q7GzR

qcentic-dev:
### CI on the PR head (`6f6b294`) — all green

Run [33663464618](https://github.com/actana/control/actions/runs/33663464618): **22 passing, 0 failing, 6 skipping.**

| Check | Result |
| --- | --- |
| Conventions | pass (20s) — commit messages + PR title + branch name |
| Unit Tests | pass (2m28s) |
| Typecheck | pass (51s) |
| Lint | pass (43s) |
| Secret Scan | pass (29s) |
| Dependency Audit | pass (41s) |
| E2E — Panel service seam | pass (52s) |
| E2E — installer (ubuntu, x64) | pass (2m4s) |
| E2E — installer (debian, x64) | pass (1m39s) |
| Smoke — Core release tarball (linux-x64 / linux-arm64) | pass (41s / 47s) |
| Core image / Panel image — Resolve registries + Build + smoke (amd64) | pass |
| Resolve PR image mode · Train rules · Promotion gate | pass |

Skipping, as expected for a ticket PR that is not a train: `Core image (train)`, `Panel image (train)`, `Resolve train tags`,
`Train versions`, and both `Publish the multi-arch manifest` jobs.

Note that `Promotion gate` passing here is not the train's promotion gate — this PR's base is
`fix/operator-pins-sessions-drop-cli`, not `main`, so the gate has nothing to hold red.

mrfqcentic (changes requested):
## Review of #449 against issue #386 (`A-enter`)

Reviewed at head `6f6b294`, diffed against `origin/fix/operator-pins-sessions-drop-cli` (`0b3b099`). I read the enforcing
lines rather than the description, and drove `advanceTerminalTurn` / `advanceTerminalRunningFallback` chunk by chunk in a
scratch harness to check each claim.

The shape of the fix is right, and it is a large net improvement: the latch is no longer re-armed by "not running" alone,
`advanceTerminalRunningFallback` owns both gates so the pane's `onData` handler is one call, and the whole thing is testable
without xterm. Scope, base, draft state, commit hygiene and CI are all in order (details at the bottom). But two of the
acceptance criteria are only **partially** met, and one of them is a regression this PR introduces, so: changes requested.

---

### 1. `packages/panel/src/lib/task-status-sync.ts:85-96` — Acceptance 1 is still reachable after a line cancel

Only `\r`/`\n` (line 85) and BACKSPACE/`\b` (line 91) touch `composed`. Every line-cancel key falls through to line 96,
where `char >= " "` is false, so the byte is dropped and **the composition survives a cancel the harness honoured**:

```
["hello", "\x03", "\r"]   → posts running   (Ctrl+C)
["hello", "\x15", "\r"]   → posts running   (Ctrl+U — and note `mapTerminalKey`
["hello", "\x17", "\r"]   → posts running   (Ctrl+W    maps Cmd+Backspace to \x15
["hello", "\x1b", "\r"]   → posts running   (Esc       and writes it straight to the
["h", "\x1b[3~", "\r"]    → posts running   (Delete    pty, so `onData` never sees it)
```

Operator story: at a **finished** Cursor CLI Session they start typing, change their mind, press Ctrl+C (or Esc, which is
how you back out of the Codex composer), then tap Enter at the now-empty prompt. The card flips to `running` on a Session
that is doing nothing. That is the sentence in Acceptance 1, verbatim, on a path the tests do not cover.

Suggestion: treat `\x03`, `\x15`, `\x17` and a lone `\x1b` as clearing `composed`, and add the three-chunk cancel script to
`terminal-running-fallback.test.ts`.

### 2. `packages/panel/src/lib/task-status-sync.ts:79-83` — Acceptance 2 regresses for a turn whose text never passed through `onData`

The escape branch steps over the whole sequence and adds nothing, which is correct for cursor keys — but it means a turn
the operator genuinely starts now posts nothing:

```
["\x1b[A", "\r"]          → no post   (Up-arrow prompt recall, then Enter)
["\x1b[A", "\x1b[A", "\r"]→ no post
["\r"]  after a drop      → no post   (path dropped from the Panel's own UI, then Enter)
```

The drop case is the one I want to flag hardest, because this PR's own correctness argument depends on it and it belongs to
this very wave: `wireTerminalFileDrop` calls `write(...)` → `ptyApi.write` at `packages/panel/src/lib/terminal-pane-helpers.ts:119`,
so a dropped path reaches the pty **without** passing through `onData`. The PR relies on that to make the false-positive
half of Acceptance 1 true — and the same fact makes the true-positive half of Acceptance 2 false. Drop a path onto a
finished Cursor CLI Session and press Enter: the Session starts a turn, nothing reports its start, `composed` is empty, and
the card stays `finished` until `stop` lands. Before this PR that Enter posted `running`.

This is the issue-84 class of stale card in the opposite direction, and Acceptance 2 says the fallback must still fire
"when the operator actually starts a turn". `["\x1b[A","x","\r"]` does post, so it is only turns the operator did not
re-type — which is exactly the cheap, common ones.

Suggestion: fold text written through `writeToPty` into the same turn state (the drop and the key map both go through that
one function at `TerminalPane.tsx:1003`), and/or treat a `\r` that arrives with an empty composition as a submit when the
pane has seen any input since the row last left `running`. Either way Acceptance 2 needs a test with a turn whose bytes did
not come from `onData`.

### 3. `packages/panel/src/lib/task-status-sync.ts:62` — a paste-start marker split across chunks re-opens the false positive

`startsWith(PASTE_START, i)` needs all six bytes in one chunk. Split at `\x1b[200`, the CSI alternative of `ESCAPE_AT` has
no final byte to match, the `\x1b[\s\S]?` catch-all eats only `\x1b[`, and `200` lands in `composed` as literal text:

```
["\x1b[200", "~/a/b\x1b[201~"] → {composed: "200~/a/b", pasting: false}
… then ["\r"]                  → posts running
```

Low likelihood with xterm's local paste (delivered atomically), but the new test only splits inside the paste *body*, so
the marker case is both untested and asymmetric with the care taken elsewhere. Buffering a trailing partial escape would
close it.

### 4. `packages/panel/src/components/views/TerminalPane.tsx:984` — the state is not rebuilt with the pty, only with the surface

The comment says "rebuilt with the surface, like every other per-pane scratch value here", and that is true — but
`wireTerminalInput` is re-called per pty (lines 1314 and 1327) while `runningFallback` is not reset. A half-typed
composition, or a `pasting: true` latch, survives into the next harness process: the first stray Enter posts `running`, or
every `\r` is swallowed as pasted text and the fallback is dead for the life of the pane. `promptCaptureBuffer` has the
same shape and the same gap, but it only costs a wrong title; this one costs the card's status.

### 5. `packages/panel/src/components/views/TerminalPane.tsx:1292` — the catch un-latches on failures that are not the PATCH

The `try` also wraps `corePtyBridge.submitPrompt` / `api.updateTaskStatus({prompt})` and the three `invalidateQueries`, so a
title-generator hiccup **after** a successful `status: "running"` PATCH still runs `posted: false` — and the comment
("The PATCH never landed") is then wrong. Because the handler re-reads the live `runningFallback` at catch time, a late
failure from turn N can also clear the latch turn N+1 just set. Both are inherited from the old boolean, but the object
makes it cheap to reset only when the status mutation itself threw.

### 6. `packages/panel/src/lib/task-status-sync.ts:7` — stale comment

"…and use Enter as the running signal" is no longer what this file does; that sentence is the bug #386 describes.

### Non-blocking, for a follow-up issue rather than this PR

`accumulateTerminalPrompt` (`terminal-prompt-capture.ts`) is a second composition tracker running on the same `onData`
chunks, and it does not know about bracketed paste — `stripTerminalEscapes` turns `\x1b[200~` into `200~`, so a pasted
first prompt is captured as `"200~/home/core/repos/control201~"` and that becomes the generated session title. Pre-existing
and outside this diff, but the two trackers have now diverged, and this PR's own paste-then-Enter story runs straight
through it. Worth an issue.

---

## Acceptance criteria

**1. After a session is finished, pressing Enter (or pasting a path) does not PATCH it to running — PARTIALLY MET.**
The named cases are fixed and proven: a bare Enter after settlement composes nothing (`task-status-sync.ts:86`, guarded by
`composed.trim().length > 0`), repeated Enters stay silent, and a bracketed paste — single- or multi-line — only composes
(`task-status-sync.ts:62-77`). A path dropped from the Panel never reaches `onData` at all. Finding 1 is the residual: a
cancelled line still submits.

**2. Harnesses that never report turn start still go running when the operator actually starts a turn — PARTIALLY MET.**
A typed prompt on a hookless Session still posts (`terminalInputStartsTurn` at `task-status-sync.ts:119-124` now forwards
`submittedTurn`, and the latch still re-arms across turns at `task-status-sync.ts:161`), and `hooksReportTurnStart` still
stands the fallback down. Finding 2 is the gap: a recalled or dropped prompt starts a real turn and posts nothing.

**Tests — Enter-after-finished and a real new turn are both covered.** `terminal-running-fallback.test.ts` scripts the
operator's story chunk by chunk (7 cases: Enter after finished, paste after finished, a real new turn, one post per turn,
re-arming, hooks stand-down, failed-PATCH retry) and `task-status-sync.test.ts` adds the composition tracker's own cases.
Good tests. They do not cover findings 1-4.

## Delivery checks

- Base is `fix/operator-pins-sessions-drop-cli` — not `main`, not `beta/0.4.5`. ✅
- Still a draft. ✅
- Scope is exactly `TerminalPane.tsx`, `task-status-sync.ts` and their two test files. ✅
- #393 (two tabs typing / `onData` disposal) not taken — no disposal or tab/lock work in the `TerminalPane.tsx` diff. ✅
- `active-group.ts` (#384, PR #450), `terminal-store.tsx`, `__root.tsx`, `ProjectBar.tsx`, `projects.$id.tsx` untouched. ✅
- PR #448's files (`harness-hook-pipeline.ts`, `core-query.ts`, `core-session-relaunch.ts`, `harness-command.ts`) untouched. ✅
- One commit, `fix(panel): …`, header 56 chars, every body line ≤ 132, blank line before `Refs #386` and before the
  trailers — passes `commitlint.config.mjs`. ✅
- CI on `6f6b294`: all 24 checks green (Unit Tests, Typecheck, Lint, Conventions, Train rules, both installer E2Es). ✅

Findings 1 and 2 are what stand between this and an approval — 2 in particular, because it trades a card that lied one way
for a card that lies the other, on the drop feature this same wave ships.

qcentic-dev:
## Review addressed — `826b0ee`

Thanks for driving the reducer chunk by chunk; every one of your scripts reproduced, and two of them were the ones I should
have written myself. All six findings are fixed in one commit on top of `6f6b294`; nothing was rewritten or force-pushed.

### 1. Acceptance 1 still reachable after a line cancel — fixed

You were right that only `\r`/`\n` and BACKSPACE touched `composed`, so every cancel key fell through to the printable branch
and was dropped. The cancels are now modelled the way the harness models them, at the granularity each one actually has:

| Key | Bytes | Now |
| --- | --- | --- |
| Ctrl+C | `\x03` | empties the line |
| Ctrl+U | `\x15` | empties the line |
| Esc (lone) | `\x1b` | empties the line |
| Ctrl+W | `\x17` | kills the trailing word (readline's unix-word-rubout) |
| Delete | `\x1b[3~` | takes one character, like Backspace |

One deliberate deviation from the literal suggestion, flagged because it is a judgement call: **Ctrl+W and Delete pop rather
than clear.** Clearing on every edit key passes your five scripts but buys a new false negative of exactly the class finding 2
is about — `["hello world", "\x17", "\r"]` and `["hello world", "\x1b[3~", "\r"]` are genuine turns, and a blanket clear stops
them posting. Word- and character-granularity gives the correct answer on both sets: your `["hello", "\x17", "\r"]` and
`["h", "\x1b[3~", "\r"]` post nothing, and the edits that leave text behind still post. Say the word if you would rather have
the blunt version.

All five of your scripts are now in `terminal-running-fallback.test.ts` (*"does not PATCH running when the operator types,
cancels, then presses Enter"*, looping the four line-emptying keys plus Delete), with the non-regression half as its own case
(*"still PATCHes running when an edit leaves text on the line"*).

### 2. Acceptance 2 regressed for a turn whose bytes never passed through `onData` — fixed

This was the right one to flag hardest; it traded a card that lied one way for a card that lies the other, on the drop feature
this same wave ships. Both suggestions taken, because they cover different halves:

- **`writeToPty` now folds into the same turn state.** `noteTerminalWrite` runs the drop path and the key map's sequences
  through `advanceTerminalTurn` at `TerminalPane.tsx:1005`, so a dropped path composes. It composes *only* — the submit that
  starts the turn is the operator's Enter, and that still arrives through `onData`, which is where the PATCH is decided. It
  also means the key map's Cmd+Backspace (`\x15`) now cancels a dropped path, which it previously could not.
- **Up/Down recall marks input as standing at the prompt** (`recalled`), because those bytes carry no text for the pane to
  see. Any cancel or submit clears it, so `["\x1b[A", "\x03", "\r"]` correctly posts nothing.

Tested in *"PATCHes running for a turn whose bytes never came through `onData`"* — a dropped path then Enter, up-arrow recall
then Enter, and a drop the operator backs out of with Cmd+Backspace. The scenario harness now takes `{ write }` chunks
alongside `{ data, status }` ones, so a turn whose bytes bypassed `onData` is expressible.

Scope note: the surface control's own `writeToPty` (`TerminalPane.tsx:1219`) is **not** folded. Its only caller is
`submitQuestionAnswers` injecting menu keys, and folding menu selections into a model of "the prompt the operator is composing"
is not obviously right — I did not want to guess at it inside this fix.

### 3. Split paste-start marker — fixed

A trailing partial escape is now held over to the next chunk, so `["\x1b[200", "~/a/b\x1b[201~"]` yields
`{composed: "/a/b", pasting: false}` instead of `200~/a/b`. Two details worth calling out: a lone trailing `\x1b` is
deliberately **not** buffered — that is the Esc key from finding 1, and waiting for a byte that never comes would swallow it —
and the check runs before the `pasting` branch, because a split **closing** marker is the worse bug of the two (the pane pastes
for ever and every `\r` is swallowed). Both directions tested, plus a 256-byte cap so an unterminated sequence cannot grow the
buffer without bound.

### 4. State not rebuilt with the pty — fixed

`runningFallback` is reset at the top of `wireTerminalInput`, which covers both call sites (`wireNewPty`, `wireExistingPty`)
and any future one. `promptCaptureBuffer` is reset there too — you are right that it only costs a title, but it is the same
statement and leaving the known-wrong sibling next to the fixed one would be odd. `promptTitlePosted` is left alone: it is a
per-Session latch, not a per-pty one. Tested by contrast — a carried-over composition posts, a carried-over `pasting` latch
swallows a real prompt, and neither happens from the idle state the pane resets to.

### 5. The catch un-latched on failures that were not the PATCH — fixed

Split into two `try` blocks. Only the status mutation's own catch un-latches, and it returns; the title and the cache
invalidations have their own catch that does nothing but swallow. Your second-order point — a late failure from turn N
clearing a latch turn N+1 had just set, because the handler re-reads the live value — goes away with it. The comment now says
what the code does.

### 6. Stale comment — fixed

The "use Enter as the running signal" sentence at the top of `task-status-sync.ts` is gone; it now says the turn's start is
read from a prompt the operator *submits*, never from a bare newline.

### Non-blocking: `accumulateTerminalPrompt` — filed as #451

Filed as **[#451](https://github.com/actana/control/issues/451)**, labelled `bug`, referencing this PR and this review, and
**not** fixed here. It records both the visible symptom (a pasted first prompt titles the Session `200~/home/core/repos/control201~`)
and your framing of why it matters now — two trackers on the same `onData` stream that disagree about what the operator typed —
with the "have the prompt capture read the composition `advanceTerminalTurn` already tracks" option written up as the fix that
stops them diverging again.

---

**Suite:** 147 files, **1592 tests**, all green (was 1583 — 9 new cases). `pnpm typecheck` and `eslint` clean.
**Commit hygiene:** one new commit, and I ran `commitlint` over the whole range `0b3b099..HEAD` with a throwaway
`@commitlint/cli` install against this repo's own config, the way the Conventions job does — 0 problems on both commits. Plain
push, no force.
**Unchanged:** base is still `fix/operator-pins-sessions-drop-cli`, still a draft, still the same four files. #393's
two-tabs-typing / `onData`-disposal work is still not taken — `wireTerminalInput` gained a state reset, not a disposal.

https://claude.ai/code/session_01C2jjKPy8zd5bV6NU7Q7GzR

qcentic-dev:
### CI on `826b0ee` — green

Run [33666367375](https://github.com/actana/control/actions/runs/33666367375): **22 passing, 0 failing, 6 skipping** (the six
train jobs, as expected for a ticket PR). Conventions 20s, Unit Tests 2m52s, Typecheck 47s, Lint 43s, both installer E2Es and
the Panel service seam E2E green.

One note, so it is on the record rather than buried in a re-run: **Unit Tests failed on the first attempt and I re-ran it.**
The failure was `packages/cli/src/__tests__/in-process-core-session.test.ts` → *"sends and awaits the turn in one PTY
resolution, and prints the settled screen"*, `Test timed out in 60000ms`. It is a real-PTY test in a package this PR does not
touch — the diff is four files, all under `packages/panel`, and nothing in `packages/cli` imports any of them. It passes
locally in **6.5s** against that 60s budget, and `packages/panel` itself reported 147 files / 1592 tests passing inside the
same failing run. A starved runner, not this change. It passed on the re-run without any code change.

mrfqcentic (approved):
## Re-review of #449 against issue #386 (`A-enter`) — round 2

Reviewed at head `826b0ee` in a fresh worktree, diffed against `origin/fix/operator-pins-sessions-drop-cli` (`0b3b099`).
I re-ran every script from round 1 plus the two the round-1 review asked for (drop-then-Enter, arrow-recall-then-Enter) and
the cancel cases, driving the real `advanceTerminalTurn` / `advanceTerminalRunningFallback` / `noteTerminalWrite` chunk by
chunk against the module as it stands at `826b0ee`.

**All six findings are fixed, both acceptance criteria now hold on every case I can name, and the Ctrl+W / Delete deviation
is the right call. Approving.**

---

### The six findings, re-driven

**1 — line cancels. Fixed.** `task-status-sync.ts:163-172` (Ctrl+C / Ctrl+U), `:141-146` (lone Esc), `:151` (Delete),
`:169-172` (Ctrl+W). Every round-1 script now posts nothing:

```
["hello","\x03","\r"]  → no post      ["hello","\x1b","\r"] → no post
["hello","\x15","\r"]  → no post      ["h","\x1b[3~","\r"]  → no post
["hello","\x17","\r"]  → no post
```

…and the non-regression half holds: `["hello world","\x17","\r"]` and `["hi","\x1b[3~","\r"]` still post, which a blanket
clear would have broken.

**2 — turns whose bytes bypass `onData`. Fixed, both halves.** `noteTerminalWrite` (`task-status-sync.ts:260-265`) wired at
`TerminalPane.tsx:1014`, and `RECALL_KEYS` at `:26` / `:152`:

```
[write "/home/core/repos/control ", "\r"]  → posts running   (drop then Enter)
[write '"/a b/c" ',                  "\r"] → posts running   (quoted path)
["\x1b[A","\r"] / ["\x1b[A","\x1b[A","\r"] / ["\x1bOA","\r"] → posts running
["\x1b[A","\x03","\r"]                     → no post         (recall then cancel)
[write "/a/b ", write "\x15", "\r"]        → no post         (drop then Cmd+Backspace)
```

I checked the claim the fix rests on: `mapTerminalKey` never maps a plain Enter (`terminal-keymap.ts:5-24`) and the paste
chord goes through `term.paste()` (`terminal-pane-helpers.ts:170-176`), so the submit really does still arrive through
`onData`. The scope note about `surface.controls.writeToPty` (`TerminalPane.tsx:1219`, only `submitQuestionAnswers`) is
right and is not a regression — those bytes bypassed `onData` before this PR too.

**3 — split paste-start marker. Fixed, and the harder direction too.** `INCOMPLETE_TAIL` at `:39`, held at `:128`.

```
["\x1b[200","~/a/b\x1b[201~"] → {composed:"/a/b", pasting:false}   (was "200~/a/b")
["\x1b[2","00~/a/b\x1b[201~"] → {composed:"/a/b", pasting:false}
["\x1b[200~a\x1b[201","~"]    → pasting:false                      (split CLOSING marker)
["\x1b[200~path\x1b[20","1~"] → pasting:false
```

The 256-byte cap holds (`\x1b[` + 300 digits → `partial:""`), and not buffering a lone trailing ESC is right — it is the
Esc key from finding 1. Split terminal replies now behave too: `["\x1b[24;","1R"]` composes nothing, where before the
catch-all left `24;` in the composition.

**4 — state rebuilt with the pty. Fixed.** `TerminalPane.tsx:1232-1233`, at the top of `wireTerminalInput`, so both
`wireNewPty` (`:1340`) and `wireExistingPty` (`:1353`) get it. Driven by contrast: a carried composition makes a stray
Enter post, a carried `pasting` latch swallows a real prompt, and neither happens from the idle state.

**5 — the catch. Fixed.** `TerminalPane.tsx:1282-1291` un-latches and returns; `:1316-1319` swallows title/cache failures
without touching the latch, and the comment now says what the code does.

**6 — stale comment. Fixed.** `task-status-sync.ts:4-9`.

### The deviation — correct, with one bug inside it

**Ctrl+W as word-kill and Delete as one character is the better choice, not the blunt clear.** Word granularity is what
readline's `unix-word-rubout` actually does, and it is the only version that gets both directions right: your five scripts
post nothing, and `["hello world","\x17","\r"]` — a genuine turn — still posts, which a blanket clear would have turned
into exactly the false negative finding 2 was about. Delete-as-Backspace is a defensible approximation for a mirror with no
cursor; strictly, Delete at end-of-line is a no-op in a real terminal, so the model over-deletes — but that errs toward not
posting, which is the safe direction for #386. Keep it as written.

One bug inside it, though. `killWord` at `:76-78` is `/\s*\S*$/`, which on a line ending in whitespace eats only the
whitespace:

```
["hello ", "\x17", "\r"]  → posts running   (readline leaves the line EMPTY)
[write "/a/b ", "\x17", "\r"] → posts running
```

That second one is the drop path — `wireTerminalFileDrop` always writes `formatPathForTerminalPaste(path) + " "`
(`terminal-pane-helpers.ts:119`), so a dropped path always ends in a space. `/\s*\S*\s*$/` fixes both. The test at
`task-status-sync.test.ts:126` already says "with the space before it, as readline's unix-word-rubout does" — the code just
does not do it when the space is trailing.

### Residuals, for a follow-up rather than this PR

None of these is a regression from `6f6b294`; each is a narrower instance of a class round 1 named, and the acceptance
criteria hold as written. Listing them so they are on the record.

Posts `running` when nothing was submitted:
- `:26` — Down is in `RECALL_KEYS`, and `recalled` is sticky. `["\x1b[A","\x1b[B","\r"]` (Up to peek, Down back to the empty
  line, Enter) posts, as does `["\x1b[B","\r"]`.
- `:151/:170/:174` — Backspace, Delete and Ctrl+W do not clear `recalled`, so recall-then-erase-to-empty-then-Enter posts.
  Partly inherent: the pane cannot know how long a recalled line is. Ctrl+C / Ctrl+U / Esc do clear it.
- `:179` — Ctrl+K (`\x0b`), Alt+Backspace (`\x1b\x7f`) and Alt+d (`\x1bd`) all empty or shorten the harness's line and are
  dropped by the printable filter / the `ESCAPE_AT` catch-all. `["hi","\x1b\x7f","\r"]` posts. Ctrl+A then Ctrl+K is the
  classic clear-the-line gesture and posts too.

Does not post when a turn did start:
- `:26` — Ctrl+P / Ctrl+N history recall is not in `RECALL_KEYS`, so `["\x10","\r"]` posts nothing. Same shape as the
  arrow-key gap this commit closed.
- `TerminalPane.tsx:1232` — the reset also runs on `wireExistingPty`, i.e. reattach to a *live* pty, where the harness's
  line is not fresh. A surface rebuild over a half-typed prompt drops the composition. The right trade (a carried `pasting`
  latch is worse), but the comment's "a new pty is a new harness process" is only true of `wireNewPty`.

Noted, no action:
- `TerminalPane.tsx:1289` — the late-failure clobber is narrowed, not eliminated: the catch still re-reads the live state.
  The remaining cost is one redundant `running` PATCH.
- `wireTerminalInput` still registers `onData` without disposing, and is called at `:1353` before the `replay.nextSeq === 0`
  bail at `:1365`, so a failed reattach can stack listeners. That is #393, correctly not taken here.
- The PR description still tells the round-1 story ("a path dropped from the Panel's own UI never reaches `onData` at all")
  and quotes 1583 tests. `826b0ee` changed exactly that. Worth a description edit before the wave merges.
- #451 is filed, labelled, referenced and not fixed here. Correct.

---

## Acceptance criteria

**1. After a session is finished, pressing Enter (or pasting a path) does not PATCH it to running — MET.** Bare Enter,
repeated Enters and whitespace-only lines compose nothing (`:157`, `hasPendingInput` at `:71`). A bracketed paste — single-
or multi-line, whole or split at either marker — composes without submitting. A cancelled line, by any of the four
line-emptying keys or by Backspace, submits nothing.

**2. Harnesses that never report turn start still go running when the operator actually starts a turn — MET.** A typed
prompt posts; a dropped path then Enter posts; up-arrow recall then Enter posts; recall-then-type posts; the latch re-arms
across turns (`:243`); `hooksReportTurnStart` still stands the fallback down (`:206`).

**Tests.** 27 cases across the two files, all green locally in a scratch copy. The nine new ones cover findings 1-4
directly, including the drop and recall turns and both split-marker directions, and the `{ write }` chunk in the scenario
harness makes a turn whose bytes bypassed `onData` expressible at all.

## Delivery checks

- `6f6b294..826b0ee` touches exactly the four files this PR already owned — `TerminalPane.tsx`, `task-status-sync.ts` and
  their two test files, all under `packages/panel`. Nothing from #384/#450, #448 or #393. ✅
- #393 (two tabs typing / `onData` disposal) still not taken: no disposal, no drive/lock work in the diff. ✅
- Merges cleanly onto `origin/fix/operator-pins-sessions-drop-cli` — `0b3b099` is an ancestor of `826b0ee`, so it is a
  fast-forward; `merge-tree` clean, GitHub reports `MERGEABLE` / `CLEAN`. ✅
- Still a draft, still based on `fix/operator-pins-sessions-drop-cli` — not `main`, not `beta/0.4.5`. ✅
- Both commits pass commitlint against this repo's own config: type `fix`, headers 56 and 60 chars (limit 120), longest
  body line 78 (limit 132), blank line before the body and before `Refs #386` and the trailers. ✅
- CI on `826b0ee`: **18 success, 6 skipped, 0 failed** across the 24 check runs — Conventions, Unit Tests, Typecheck, Lint,
  Train rules, Promotion gate, both installer E2Es, the Panel service seam E2E, both tarball smokes. The six skips are the
  train jobs, as expected for a ticket PR. The disclosed first-attempt Unit Tests failure
  (`in-process-core-session.test.ts`, a 60s real-PTY timeout in `packages/cli`, a package this diff does not touch) reads
  as a starved runner; it passed on re-run. ✅
- Panel suite in a scratch copy: 1592 tests, 1587 passed, 5 skipped. One suite
  (`server/__tests__/token-usage-rollup.test.ts`) hit a `beforeAll` timeout under my parallel run and passes on its own in
  1.25s — my environment, not the diff. ✅

Good work on this round — the drop-path fold in particular, which is the half the first pass got backwards.

PR #450, opened 2026-09-02
fix(panel): ignore a stale setActiveGroup server answer

qcentic-dev:
The B2 half of the pin/session/CLI/drop wave: `setActiveGroup` applied whichever `updateSettings` PATCH resolved
last, with no generation, so an older answer could revert the rail under the cursor.

> **Scope note — #384's symptom is not fully closed by this PR.** The guard here orders the answers *this hook*
> is responsible for, and cancels a settings GET in flight before its own write. It does **not** order the other
> writers of the same settings row: `collapsed-groups.ts` and `hideable-elements.tsx` still write
> `queryKeys.settings` with a pre-click `activeProjectGroup`, and the same operator-visible revert is reachable
> through them. That is filed as **#452** and wants one shared write path for the row rather than a third guard.

## What was wrong

`packages/panel/src/lib/active-group.ts` wrote the group into the settings query cache optimistically and then
`setQueryData`'d whatever the PATCH answered. Two PATCHes in flight are not ordered — pick group A then group B and
A's answer can land last, carrying the server's older `activeProjectGroup` back over B. Two rapid group changes (or
racing tabs) revert the rail to an older group mid-click. A failed PATCH restored nothing at all, so the rail kept
showing a selection the server never accepted.

## What changed

- A per-tab **write ledger**. Every local write claims the next generation, and an answer may only *paint* while
  its generation is still the newest. An older PATCH resolving late paints nothing.
- **Acknowledgement is tracked separately from painting**, with a generation of its own that refuses an answer
  older than the one already recorded. A superseded PATCH that succeeded did confirm its group, so it is recorded
  even though it may not paint — that is what keeps a later failure from rolling the rail back past a group the
  server is actually holding.
- A failed PATCH restores the **last acknowledged group**; a failure whose write has already been superseded rolls
  nothing back. Before settings hydrate the ledger is seeded from the value the rail is already rendering, so a
  click made in that window and then failing does not fall back to "All projects".
- Every write awaits `queryClient.cancelQueries({ queryKey: queryKeys.settings })` before painting — the pattern
  `UsageSettingsPage.tsx:60` and `ProvidersSettingsPage.tsx:75` already use — so a settings read in flight cannot
  answer with the pre-PATCH row and paint it over the optimistic write. The optimistic paint takes the same
  currency guard the answers do, since that await means two rapid clicks resume in an order the click order does
  not settle by itself.
- The confirm path **merges** the group into the current row instead of replacing the row, so a concurrent edit to
  a neighbouring setting is not undone by an answer carrying it at its pre-edit value.
- The request is awaited inside one `try`/`catch`, so `settleWrite` runs exactly once per write.

`ProjectBar.tsx`, `TerminalPane.tsx`, `task-status-sync.ts`, `terminal-store.tsx`, `__root.tsx` and
`projects.$id.tsx` are untouched — parallel tickets own them. Two files in the diff: `active-group.ts` and its test.

## Acceptance (issue #384)

- [x] **Rapid group A then B leaves B selected even if A's PATCH returns last.** `leaves B selected when the A
  answer returns last` drives the real hook with two hand-resolved PATCHes: B answers first and is applied, A
  answers last with the older group and paints nothing. The rail and the localStorage seed both stay on B.
- [x] **A failed PATCH restores the last acknowledged group, not a random older one.** Covered in both
  interleavings: `restores the last acknowledged group when a PATCH fails` (nothing else in flight) and `rolls back
  to the group an older, superseded PATCH confirmed` (the interleaved case the review found, where the pre-review
  shape landed on the value from before both clicks). `rolls a failed first click back to the cached group, not to
  All projects` covers the same root cause before settings hydrate.

Both new blocking cases were run against the previous shape and fail there, as do the two original ones against the
pre-fix code.

## Known limits

- The sibling-writer race is **#452**, described in the scope note above.
- `req()` in `lib/api.ts` sets no timeout or abort signal, so a never-answering PATCH holds the in-flight count
  above zero for the tab's lifetime and `observe` stops adopting server answers. Fixing it means giving the
  settings PATCH a signal in `api.ts`; no ticket in this wave owns that.
- With two writes out, an older success landing *after* the newer one has already failed and rolled back leaves the
  rail on the rolled-back group while the ledger records the server's newer truth. AC2 still holds at the moment of
  the failure — the rail goes to what was acknowledged then — and closing it properly needs the shared write path
  #452 asks for.

## Testing

- `src/lib/__tests__/active-group.test.tsx`: 15 tests, green, run five times for stability.
- Full `packages/panel` suite: **147 files, 1584 tests, all passing, none skipped.**
- `pnpm typecheck` clean; `eslint` clean on both files.

Base is `fix/operator-pins-sessions-drop-cli`, not `main` and not the train. Draft on purpose. `Refs` rather than a
closing keyword, deliberately: this stacks into the wave branch, and #384's symptom outlives this PR via #452.

Refs #384
Refs #452

https://claude.ai/code/session_01VkZFxb8DWtTZdQea3JfgTx

### DIFF
--- a/packages/panel/src/lib/__tests__/active-group.test.tsx
+++ b/packages/panel/src/lib/__tests__/active-group.test.tsx
@@ -54,18 +54,19 @@ function deferred<T>() {
   return { promise, resolve, reject };
 }
 
-function mount(initial: string | null) {
+/** `initial` is what the server holds; `undefined` leaves settings unhydrated. */
+function mount(initial: string | null | undefined) {
   const client = new QueryClient({
     defaultOptions: {
       queries: { retry: false, staleTime: Infinity, refetchOnMount: false, refetchOnWindowFocus: false },
     },
   });
-  client.setQueryData(queryKeys.settings, settingsWith(initial));
+  if (initial !== undefined) client.setQueryData(queryKeys.settings, settingsWith(initial));
   client.setQueryData(queryKeys.groups, GROUPS);
   const wrapper = ({ children }: { children: ReactNode }) => (
     <QueryClientProvider client={client}>{children}</QueryClientProvider>
   );
-  return renderHook(() => useActiveGroup(), { wrapper });
+  return { client, ...renderHook(() => useActiveGroup(), { wrapper }) };
 }
 
 /**
@@ -98,7 +99,7 @@ afterEach(() => {
 });
 
 describe("the active-group write ledger", () => {
-  it("only calls the newest write current", () => {
+  it("only reports the newest write as current", () => {
     const ledger = createActiveGroupWriteLedger(ACTIVE_GROUP_ALL);
     const first = ledger.beginWrite();
     const second = ledger.beginWrite();
@@ -109,8 +110,40 @@ describe("the active-group write ledger", () => {
   it("keeps the acknowledged group until a newer answer confirms one", () => {
     const ledger = createActiveGroupWriteLedger("g-a");
     expect(ledger.lastAcknowledged()).toBe("g-a");
-    ledger.acknowledge("g-b");
+    ledger.acknowledge(ledger.beginWrite(), "g-b");
     expect(ledger.lastAcknowledged()).toBe("g-b");
+  });
+
+  it("refuses an acknowledgement older than the one already recorded", () => {
+    const ledger = createActiveGroupWriteLedger("g-a");
+    const first = ledger.beginWrite();
+    const second = ledger.beginWrite();
+    ledger.acknowledge(second, "g-c");
+    // The older PATCH answers last; its group is the server's older truth.
+    ledger.acknowledge(first, "g-b");
+    expect(ledger.lastAcknowledged()).toBe("g-c");
+  });
+
+  it("records a superseded answer that the server did confirm", () => {
+    const ledger = createActiveGroupWriteLedger("g-a");
+    const first = ledger.beginWrite();
+    ledger.beginWrite();
+    // The older write is no longer the one that may paint...
+    expect(ledger.settleWrite(first)).toBe(false);
+    // ...but the server took it, so it is the group a later failure returns to.
+    ledger.acknowledge(first, "g-b");
+    expect(ledger.lastAcknowledged()).toBe("g-b");
+  });
+
+  it("seeds once, and never over something already known", () => {
+    const ledger = createActiveGroupWriteLedger();
+    ledger.seed("g-a");
+    expect(ledger.lastAcknowledged()).toBe("g-a");
+    ledger.seed("g-b");
+    expect(ledger.lastAcknowledged()).toBe("g-a");
+    ledger.observe("g-c");
+    ledger.seed("g-d");
+    expect(ledger.lastAcknowledged()).toBe("g-c");
   });
 
   it("ignores a settings read taken while a local write is in flight", () => {
@@ -121,6 +154,19 @@ describe("the active-group write ledger", () => {
     expect(ledger.lastAcknowledged()).toBe("g-a");
     ledger.settleWrite(generation);
     ledger.observe("g-b");
+    expect(ledger.lastAcknowledged()).toBe("g-b");
+  });
+
+  it("lets an observed answer outrank every write older than the newest", () => {
+    const ledger = createActiveGroupWriteLedger("g-a");
+    const first = ledger.beginWrite();
+    const second = ledger.beginWrite();
+    ledger.settleWrite(first);
+    ledger.settleWrite(second);
+    // A refetch answered while nothing was in flight: it is the server's word
+    // on every write issued so far.
+    ledger.observe("g-b");
+    ledger.acknowledge(first, "g-c");
     expect(ledger.lastAcknowledged()).toBe("g-b");
   });
 });
@@ -192,12 +238,103 @@ describe("useActiveGroup under racing PATCHes", () => {
     expect(result.current.activeGroup).toBe("g-b");
   });
 
+  // The interleaving the review found: the older write is the one the server
+  // took, so the failure of the newer one must land there — not on the value
+  // from before either click.
+  it("rolls back to the group an older, superseded PATCH confirmed", async () => {
+    const b = deferred<AppSettings>();
+    const c = deferred<AppSettings>();
+    updateSettings.mockReturnValueOnce(b.promise).mockReturnValueOnce(c.promise);
+
+    // The server holds "g-a"; the operator clicks "g-b" then "g-c".
+    const { result } = mount("g-a");
+    act(() => result.current.setActiveGroup("g-b"));
+    act(() => result.current.setActiveGroup("g-c"));
+    await settle();
+    expect(result.current.activeGroup).toBe("g-c");
+
+    // "g-b" is superseded and paints nothing, but the server did take it.
+    b.resolve(settingsWith("g-b"));
+    await settle();
+    expect(result.current.activeGroup).toBe("g-c");
+
+    c.reject(new Error("PATCH /api/settings failed"));
+    await settle();
+    expect(result.current.activeGroup).toBe("g-b");
+    expect(window.localStorage.getItem(ACTIVE_PROJECT_GROUP_STORAGE_KEY)).toBe("g-b");
+  });
+
+  // Same root cause, one step earlier: with no settings answer yet, the ledger
+  // has to be seeded from the value the rail is running off.
+  it("rolls a failed first click back to the cached group, not to All projects", async () => {
+    window.localStorage.setItem(ACTIVE_PROJECT_GROUP_STORAGE_KEY, "g-a");
+    const first = deferred<AppSettings>();
+    updateSettings.mockReturnValueOnce(first.promise);
+
+    // Settings are unhydrated, so there is no cached row for the optimistic
+    // write to land in — localStorage is where the rollback is observable.
+    const { result } = mount(undefined);
+    expect(result.current.activeGroup).toBe("g-a");
+
+    act(() => result.current.setActiveGroup("g-b"));
+    await settle();
+    expect(window.localStorage.getItem(ACTIVE_PROJECT_GROUP_STORAGE_KEY)).toBe("g-b");
+
+    first.reject(new Error("PATCH /api/settings failed"));
+    await settle();
+    expect(window.localStorage.getItem(ACTIVE_PROJECT_GROUP_STORAGE_KEY)).toBe("g-a");
+  });
+
+  // The GET half of the race: a settings read already in flight would answer
+  // with the pre-PATCH row and paint it over the optimistic write.
+  it("cancels an in-flight settings read before writing", async () => {
+    const only = deferred<AppSettings>();
+    updateSettings.mockReturnValueOnce(only.promise);
+
+    const { client, result } = mount("g-a");
+    const cancelQueries = vi.spyOn(client, "cancelQueries");
+
+    act(() => result.current.setActiveGroup("g-b"));
+    await settle();
+
+    expect(cancelQueries).toHaveBeenCalledWith({ queryKey: queryKeys.settings });
+    expect(cancelQueries.mock.invocationCallOrder[0]).toBeLessThan(
+      updateSettings.mock.invocationCallOrder[0]!,
+    );
+  });
+
+  it("keeps a neighbouring setting a concurrent edit changed mid-flight", async () => {
+    const only = deferred<AppSettings>();
+    updateSettings.mockReturnValueOnce(only.promise);
+
+    const { client, result } = mount("g-a");
+    act(() => result.current.setActiveGroup("g-b"));
+    await settle();
+
+    // Something else writes the settings row while the PATCH is out.
+    act(() => {
+      client.setQueryData<AppSettings>(queryKeys.settings, (current) =>
+        current ? { ...current, collapsedProjectGroups: ["pinned"] } : current,
+      );
+    });
+
+    // The answer carries that key at its pre-edit value; merging keeps the edit.
+    only.resolve({ activeProjectGroup: "g-b", collapsedProjectGroups: null } as AppSettings);
+    await settle();
+    expect(result.current.activeGroup).toBe("g-b");
+    expect(client.getQueryData<AppSettings>(queryKeys.settings)?.collapsedProjectGroups).toEqual([
+      "pinned",
+    ]);
+  });
+
   it("still applies a lone answer, and stores 'all' as null", async () => {
     const only = deferred<AppSettings>();
     updateSettings.mockReturnValueOnce(only.promise);
 
     const { result } = mount("g-a");
     act(() => result.current.setActiveGroup(ACTIVE_GROUP_ALL));
+    // The PATCH goes out behind `cancelQueries`, so it is not synchronous.
+    await settle();
     expect(updateSettings).toHaveBeenCalledWith({ activeProjectGroup: null });
 
     only.resolve(settingsWith(null));
--- a/packages/panel/src/lib/__tests__/active-group.test.tsx
+++ b/packages/panel/src/lib/__tests__/active-group.test.tsx
@@ -1,0 +1,207 @@
+// @vitest-environment jsdom
+//
+// The active group is written optimistically and confirmed by a PATCH, and two
+// PATCHes in flight are not ordered (#384). What is proven here is that the
+// *newest local selection* owns the rail: an older answer landing last is
+// dropped, and a failure hands the rail back to the last group the server
+// acknowledged rather than to whatever happened to be on screen before.
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { act, cleanup, renderHook } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import type { ReactNode } from "react";
+import type { AppSettings } from "~/lib/api";
+import type { Group } from "~/db/schema";
+
+const updateSettings = vi.fn();
+
+vi.mock("~/lib/api", () => ({
+  api: {
+    updateSettings: (body: unknown) => updateSettings(body),
+    getSettings: () => Promise.reject(new Error("settings must be primed in these tests")),
+    listGroups: () => Promise.reject(new Error("groups must be primed in these tests")),
+  },
+}));
+
+const {
+  ACTIVE_GROUP_ALL,
+  createActiveGroupWriteLedger,
+  useActiveGroup,
+  __resetActiveGroupWritesForTests,
+} = await import("~/lib/active-group");
+const { queryKeys } = await import("~/queries");
+const { ACTIVE_PROJECT_GROUP_STORAGE_KEY } = await import("~/lib/ui-preference-cache");
+
+function settingsWith(activeProjectGroup: string | null): AppSettings {
+  return { activeProjectGroup } as AppSettings;
+}
+
+const GROUPS = [
+  { id: "g-a", name: "A" },
+  { id: "g-b", name: "B" },
+  { id: "g-c", name: "C" },
+] as Group[];
+
+/** A promise the test resolves by hand, so answers can land out of order. */
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  let reject!: (error: unknown) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  // Nothing awaits a rejection until the test does; keep node quiet meanwhile.
+  promise.catch(() => {});
+  return { promise, resolve, reject };
+}
+
+function mount(initial: string | null) {
+  const client = new QueryClient({
+    defaultOptions: {
+      queries: { retry: false, staleTime: Infinity, refetchOnMount: false, refetchOnWindowFocus: false },
+    },
+  });
+  client.setQueryData(queryKeys.settings, settingsWith(initial));
+  client.setQueryData(queryKeys.groups, GROUPS);
+  const wrapper = ({ children }: { children: ReactNode }) => (
+    <QueryClientProvider client={client}>{children}</QueryClientProvider>
+  );
+  return renderHook(() => useActiveGroup(), { wrapper });
+}
+
+/**
+ * Let the settled PATCH handlers run and the hook re-render on their writes.
+ * A macrotask, not a microtask: react-query batches its observer notifications
+ * onto a `setTimeout`, so a flushed promise chain alone leaves `result.current`
+ * on the previous render.
+ */
+async function settle() {
+  await act(async () => {
+    // Several rounds: the PATCH handler runs on a microtask, its cache write
+    // schedules the notification on the *next* timer, and the render lands
+    // after that. One round would read the state before its own effect.
+    for (let round = 0; round < 3; round += 1) {
+      await new Promise((resolve) => setTimeout(resolve, 0));
+    }
+  });
+}
+
+beforeEach(() => {
+  updateSettings.mockReset();
+  window.localStorage.clear();
+  __resetActiveGroupWritesForTests();
+  vi.spyOn(console, "error").mockImplementation(() => {});
+});
+
+afterEach(() => {
+  cleanup();
+  vi.restoreAllMocks();
+});
+
+describe("the active-group write ledger", () => {
+  it("only calls the newest write current", () => {
+    const ledger = createActiveGroupWriteLedger(ACTIVE_GROUP_ALL);
+    const first = ledger.beginWrite();
+    const second = ledger.beginWrite();
+    expect(ledger.settleWrite(first)).toBe(false);
+    expect(ledger.settleWrite(second)).toBe(true);
+  });
+
+  it("keeps the acknowledged group until a newer answer confirms one", () => {
+    const ledger = createActiveGroupWriteLedger("g-a");
+    expect(ledger.lastAcknowledged()).toBe("g-a");
+    ledger.acknowledge("g-b");
+    expect(ledger.lastAcknowledged()).toBe("g-b");
+  });
+
+  it("ignores a settings read taken while a local write is in flight", () => {
+    const ledger = createActiveGroupWriteLedger("g-a");
+    const generation = ledger.beginWrite();
+    // The optimistic value is this tab's own, not the server's word.
+    ledger.observe("g-b");
+    expect(ledger.lastAcknowledged()).toBe("g-a");
+    ledger.settleWrite(generation);
+    ledger.observe("g-b");
+    expect(ledger.lastAcknowledged()).toBe("g-b");
+  });
+});
+
+describe("useActiveGroup under racing PATCHes", () => {
+  it("leaves B selected when the A answer returns last", async () => {
+    const a = deferred<AppSettings>();
+    const b = deferred<AppSettings>();
+    updateSettings.mockReturnValueOnce(a.promise).mockReturnValueOnce(b.promise);
+
+    const { result } = mount(null);
+
+    act(() => result.current.setActiveGroup("g-a"));
+    act(() => result.current.setActiveGroup("g-b"));
+    await settle();
+    expect(result.current.activeGroup).toBe("g-b");
+
+    // B's answer lands first and is the newest write: it is applied.
+    b.resolve(settingsWith("g-b"));
+    await settle();
+    expect(result.current.activeGroup).toBe("g-b");
+
+    // A's answer lands last carrying the older group — and is dropped.
+    a.resolve(settingsWith("g-a"));
+    await settle();
+    expect(result.current.activeGroup).toBe("g-b");
+    expect(window.localStorage.getItem(ACTIVE_PROJECT_GROUP_STORAGE_KEY)).toBe("g-b");
+  });
+
+  it("restores the last acknowledged group when a PATCH fails", async () => {
+    const ok = deferred<AppSettings>();
+    const bad = deferred<AppSettings>();
+    updateSettings.mockReturnValueOnce(ok.promise).mockReturnValueOnce(bad.promise);
+
+    // "g-a" is what the server has; "g-b" then becomes the acknowledged group.
+    const { result } = mount("g-a");
+    act(() => result.current.setActiveGroup("g-b"));
+    ok.resolve(settingsWith("g-b"));
+    await settle();
+    expect(result.current.activeGroup).toBe("g-b");
+
+    act(() => result.current.setActiveGroup("g-c"));
+    await settle();
+    expect(result.current.activeGroup).toBe("g-c");
+    bad.reject(new Error("PATCH /api/settings failed"));
+    await settle();
+
+    // Back to the acknowledged group — not "g-a", which the rail last showed
+    // before the successful write.
+    expect(result.current.activeGroup).toBe("g-b");
+    expect(window.localStorage.getItem(ACTIVE_PROJECT_GROUP_STORAGE_KEY)).toBe("g-b");
+  });
+
+  it("lets a newer selection stand when an older PATCH fails last", async () => {
+    const a = deferred<AppSettings>();
+    const b = deferred<AppSettings>();
+    updateSettings.mockReturnValueOnce(a.promise).mockReturnValueOnce(b.promise);
+
+    const { result } = mount(null);
+    act(() => result.current.setActiveGroup("g-a"));
+    act(() => result.current.setActiveGroup("g-b"));
+
+    b.resolve(settingsWith("g-b"));
+    await settle();
+    a.reject(new Error("PATCH /api/settings failed"));
+    await settle();
+
+    // A superseded failure rolls nothing back.
+    expect(result.current.activeGroup).toBe("g-b");
+  });
+
+  it("still applies a lone answer, and stores 'all' as null", async () => {
+    const only = deferred<AppSettings>();
+    updateSettings.mockReturnValueOnce(only.promise);
+
+    const { result } = mount("g-a");
+    act(() => result.current.setActiveGroup(ACTIVE_GROUP_ALL));
+    expect(updateSettings).toHaveBeenCalledWith({ activeProjectGroup: null });
+
+    only.resolve(settingsWith(null));
+    await settle();
+    expect(result.current.activeGroup).toBe(ACTIVE_GROUP_ALL);
+  });
+});
--- a/packages/panel/src/lib/__tests__/active-group.test.tsx
+++ b/packages/panel/src/lib/__tests__/active-group.test.tsx
@@ -13,11 +13,12 @@ import type { AppSettings } from "~/lib/api";
 import type { Group } from "~/db/schema";
 
 const updateSettings = vi.fn();
+const getSettings = vi.fn();
 
 vi.mock("~/lib/api", () => ({
   api: {
     updateSettings: (body: unknown) => updateSettings(body),
-    getSettings: () => Promise.reject(new Error("settings must be primed in these tests")),
+    getSettings: () => getSettings(),
     listGroups: () => Promise.reject(new Error("groups must be primed in these tests")),
   },
 }));
@@ -54,6 +55,21 @@ function deferred<T>() {
   return { promise, resolve, reject };
 }
 
+/**
+ * A client whose settings query really fetches — no primed row, so the first
+ * GET is genuinely in flight and can be cancelled (or not) for real.
+ */
+function mountLive() {
+  const client = new QueryClient({
+    defaultOptions: { queries: { retry: false, staleTime: Infinity, refetchOnWindowFocus: false } },
+  });
+  client.setQueryData(queryKeys.groups, GROUPS);
+  const wrapper = ({ children }: { children: ReactNode }) => (
+    <QueryClientProvider client={client}>{children}</QueryClientProvider>
+  );
+  return { client, ...renderHook(() => useActiveGroup(), { wrapper }) };
+}
+
 /** `initial` is what the server holds; `undefined` leaves settings unhydrated. */
 function mount(initial: string | null | undefined) {
   const client = new QueryClient({
@@ -88,6 +104,8 @@ async function settle() {
 
 beforeEach(() => {
   updateSettings.mockReset();
+  getSettings.mockReset();
+  getSettings.mockRejectedValue(new Error("settings must be primed in these tests"));
   window.localStorage.clear();
   __resetActiveGroupWritesForTests();
   vi.spyOn(console, "error").mockImplementation(() => {});
@@ -101,8 +119,8 @@ afterEach(() => {
 describe("the active-group write ledger", () => {
   it("only reports the newest write as current", () => {
     const ledger = createActiveGroupWriteLedger(ACTIVE_GROUP_ALL);
-    const first = ledger.beginWrite();
-    const second = ledger.beginWrite();
+    const first = ledger.beginWrite("g-b");
+    const second = ledger.beginWrite("g-c");
     expect(ledger.settleWrite(first)).toBe(false);
     expect(ledger.settleWrite(second)).toBe(true);
   });
@@ -110,14 +128,14 @@ describe("the active-group write ledger", () => {
   it("keeps the acknowledged group until a newer answer confirms one", () => {
     const ledger = createActiveGroupWriteLedger("g-a");
     expect(ledger.lastAcknowledged()).toBe("g-a");
-    ledger.acknowledge(ledger.beginWrite(), "g-b");
+    ledger.acknowledge(ledger.beginWrite("g-b"), "g-b");
     expect(ledger.lastAcknowledged()).toBe("g-b");
   });
 
   it("refuses an acknowledgement older than the one already recorded", () => {
     const ledger = createActiveGroupWriteLedger("g-a");
-    const first = ledger.beginWrite();
-    const second = ledger.beginWrite();
+    const first = ledger.beginWrite("g-b");
+    const second = ledger.beginWrite("g-c");
     ledger.acknowledge(second, "g-c");
     // The older PATCH answers last; its group is the server's older truth.
     ledger.acknowledge(first, "g-b");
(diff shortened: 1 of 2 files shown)


qcentic-dev:
### Acceptance checklist — issue #384

- [x] **1. Rapid group A then B leaves B selected even if A's PATCH returns last.**
  `setActiveGroup` claims a generation per local write; a settings answer is applied only while its generation is
  still the newest. Test `leaves B selected when the A answer returns last` drives the real `useActiveGroup` hook
  with two hand-resolved PATCHes — B answers first and is applied, A answers last carrying the older group and is
  dropped. Both the rendered `activeGroup` and the localStorage seed stay on B.

- [x] **2. A failed PATCH restores the last acknowledged group, not a random older one.**
  The ledger records the group the server confirmed for the newest write (and adopts a settings answer observed
  while nothing is in flight). Test `restores the last acknowledged group when a PATCH fails` starts on `g-a`,
  successfully writes `g-b`, then fails writing `g-c`: the rail returns to `g-b`, not `g-a`. A superseded failure
  rolls nothing back — covered by `lets a newer selection stand when an older PATCH fails last`.

Nothing is left unchecked.

Both criterion tests were confirmed to fail against the pre-fix behaviour before being kept.

**Scope:** only `packages/panel/src/lib/active-group.ts` and its new test file. `ProjectBar.tsx` (#376/#378/#382),
`TerminalPane.tsx` / `task-status-sync.ts` (#386), the `packages/shared` files and `harness-command.ts` (#387 / PR
#448), `terminal-store.tsx`, `__root.tsx` and `projects.$id.tsx` are all untouched.

**Local verification:** new suite 7/7 green (run five times); `project-picker-switch`, `rail-projects` and
`sort-projects` green; full `packages/panel` suite 1571 passed / 5 skipped, with one unrelated suite
(`src/server/__tests__/token-usage-rollup.test.ts`) timing out in its `beforeAll` under full parallel load and
passing on its own. `pnpm typecheck` and `eslint` clean.

**CI on the PR head (`2de6b3d391f1bc13ec33ce3b907e5011a431fd3e`): green.** All 18 required checks pass — Conventions, Lint, Typecheck, Unit
Tests, Secret Scan, Dependency Audit, E2E — Panel service seam, E2E — installer (debian x64 / ubuntu x64), both
Core release tarball smokes (linux-x64 / linux-arm64), Core and Panel image build + smoke (amd64) with their
registry resolves, Promotion gate, Train rules, Resolve PR image mode. The 6 train-only jobs skip, as they do on
any non-train pull request.

mrfqcentic (changes requested):
Reviewed against issue #384 (B2) by reading `packages/panel/src/lib/active-group.ts` line by line and by running
executable probes against the head commit in a scratch copy — not from the PR description.

## Acceptance criteria

**AC1 — "Rapid group A then B leaves B selected even if A's PATCH returns last": MET.**

`active-group.ts:166` draws a generation for *every* local write, `:173` drops an answer whose generation is no
longer the newest, and the drop is total — no `setQueryData`, no localStorage write, no acknowledgement. The guard
is load-bearing: replacing `if (!writes.settleWrite(generation)) return;` at :173 with a bare
`writes.settleWrite(generation);` makes `leaves B selected when the A answer returns last` fail with
`expected 'g-a' to be 'g-b'`, so the test genuinely discriminates.

I also checked the two properties you'd want of the counter itself. It is **not** per-render: a probe that
re-renders between the two writes still ends on B. It is **not** per-hook-instance, and shouldn't be — a probe
with two independent `useActiveGroup()` instances sharing one QueryClient (as `__root.tsx:271` and
`ProjectPicker.tsx:99` really are) also ends on B. The module-level singleton at :108 is the right shape and the
reason given in its comment checks out: `__root.tsx`, `projects.$id.tsx`, `GroupSwitcher.tsx` and
`ProjectPicker.tsx` all write this same setting.

**AC2 — "A failed PATCH still restores the last acknowledged group, not a random older one": PARTIALLY MET.**

The case the ticket names works, and its guard is load-bearing too — removing the restore at :183 makes
`restores the last acknowledged group when a PATCH fails` fail with `expected 'g-c' to be 'g-b'`. But there is a
reachable interleaving, *inside the very race this ticket is about*, where the failure path restores a group older
than the one the server acknowledged. Blocking comment below.

## Blocking

### 1. `packages/panel/src/lib/active-group.ts:174` — a superseded-but-successful answer never acknowledges, so a later failure rolls back too far

`writes.acknowledge(...)` sits *after* the staleness guard at :173, so when an older PATCH succeeds its answer is
dropped whole — including the fact that the server confirmed that group. If the newer PATCH then fails, `:183`
restores `writes.lastAcknowledged()`, which is still the value from before *both* writes.

Concretely: the server holds `g-a`; the operator clicks `g-b` then `g-c`; the `g-b` PATCH succeeds (the server is
now genuinely on `g-b`); the `g-c` PATCH then fails. The rail *and* the localStorage seed both land on **`g-a`**
while the server holds `g-b`. Run against the head commit:

```
PROBE rail after failure = g-a localStorage = g-a
AssertionError: expected 'g-a' to be 'g-b'
```

That is the shape AC2 forbids — a failed PATCH putting the rail on an older group. It is also self-entrenching:
the rollback write at :183 re-renders, the `observe` effect at :148-151 then runs with `inFlight === 0` and adopts
`g-a` as the acknowledged truth, so the ledger now believes the server holds `g-a` until a real refetch corrects
it. On the next reload the rail jumps to `g-b` with no user action.

Suggested shape — give the acknowledgement its own generation, and acknowledge before the display guard:

```ts
// createActiveGroupWriteLedger
let acknowledgedGeneration = 0;
acknowledge(generation: number, group: ActiveProjectGroup): void {
  if (generation < acknowledgedGeneration) return;
  acknowledgedGeneration = generation;
  acknowledged = group;
},
observe(group: ActiveProjectGroup): void {
  if (inFlight === 0) { acknowledged = group; acknowledgedGeneration = latest; }
},
```

```ts
// setActiveGroup's .then
.then((updated) => {
  const current = writes.settleWrite(generation);
  writes.acknowledge(generation, activeGroupOf(updated)); // the server did confirm this, superseded or not
  if (!current) return;                                    // …but only the newest answer may paint
  writeCachedActiveProjectGroup(activeGroupOf(updated));
  queryClient.setQueryData(queryKeys.settings, updated);
})
```

AC1 is unaffected: `acknowledge(1, 'g-a')` after `acknowledge(2, 'g-b')` is refused, and the stale answer still
paints nothing.

### 2. `packages/panel/src/lib/__tests__/active-group.test.tsx:177` — the missing race

`lets a newer selection stand when an older PATCH fails last` covers the mirror case (older fails, newer already
succeeded). The untested one is **older succeeds, newer fails** — four lines from the test already there, and the
case that currently regresses AC2. Please add it alongside the fix.

A cheaper second instance of the same root cause, worth covering while you're in here: if the operator clicks
before the settings query resolves, the `observe` effect at :149 returns early, `acknowledged` is still the
`ACTIVE_GROUP_ALL` default from :67/:70, and a failed first PATCH puts the rail on "All projects" rather than on
what the server holds. Seeding the ledger from `settings` on first load fixes that one.

## Should fix, or split into a follow-up (please decide explicitly)

### 3. `packages/panel/src/lib/active-group.ts:148` — the guard covers this hook's PATCH answers, not every settings response

The generation only orders answers to PATCHes *this hook* issued. Two other paths still write the server's older
`activeProjectGroup` straight into `queryKeys.settings`, and the ledger never sees them:

- `packages/panel/src/lib/collapsed-groups.ts:44` does `.then((updated) => setQueryData(queryKeys.settings, updated))`
  — a full-row replace. Toggle a group's collapse in the rail, then pick a different group inside that PATCH's
  flight: the collapse answer lands last carrying the old `activeProjectGroup` and the rail reverts under the
  cursor. That is the #384 operator symptom verbatim, through a neighbouring writer.
  `packages/panel/src/lib/hideable-elements.tsx` merges the same key.
- A plain settings GET. `staleTime` is 30s (`router.tsx:198`) with `refetchOnMount` left at its default, so any
  component mounting `useSettings()` just after a click refetches and writes the server's pre-PATCH value over the
  optimistic one. This is also the "racing tabs" half of the issue's symptom line.

The house pattern for exactly this is `await queryClient.cancelQueries({ queryKey: queryKeys.settings })` before
the optimistic write — see `UsageSettingsPage.tsx:60` and `ProvidersSettingsPage.tsx:75`. Two lines here would
close the GET half. The sibling-writer half is arguably its own ticket; I'd rather see it filed than left
unremarked, because the PR description reads as though #384's symptom is fully closed and it is not.

## Non-blocking

- `active-group.ts:176` — the confirm path full-replaces the settings row with `updated`, which carries every
  unrelated field at its pre-concurrent-edit value. Collapse a section, then switch group before that PATCH
  commits, and the group answer restores the old `collapsedProjectGroups` — the section pops back open.
  Pre-existing (the base does the same), but `showActiveGroup` at :157 already merges correctly and the confirm
  path could merge the same way.
- `active-group.ts:80-83` — `settleWrite` is the only thing that decrements `inFlight`, and `req()` in
  `lib/api.ts` sets no timeout or abort signal. A hung PATCH leaves `inFlight > 0` for the tab's lifetime, after
  which `observe` never adopts a server answer again. Relatedly, `.catch` is chained after `.then`, so a throw
  inside the then-body calls `settleWrite(generation)` a second time and under-counts a genuinely in-flight write.
- `active-group.test.tsx:101` — `it("only calls the newest write current")` does not parse; I think it wants
  "only reports the newest write as current".
- The body links the issue as `Refs #384`; `CONTRIBUTING.md:312` asks for a closing keyword. Since this stacks
  into `fix/operator-pins-sessions-drop-cli` rather than the train, `Refs` is defensible and #443 does the same —
  flagging only so the wave owner decides deliberately.

## Everything else checks out

- **Scope.** Two files: `active-group.ts` (+106/−7) and the new `active-group.test.tsx` (+207). `ProjectBar.tsx`
  (#376/#378/#382), `TerminalPane.tsx` and `task-status-sync.ts` (#386 / PR #449), `terminal-store.tsx`,
  `__root.tsx` and `projects.$id.tsx` are all untouched, and there is no overlap with the sixteen files #448 owns.
- **Base.** `fix/operator-pins-sessions-drop-cli` (at `0b3b099`), not `main` and not `beta/0.4.5`. One commit
  ahead, zero behind, no merge commits.
- **Draft.** Still a draft, no prior reviews.
- **Commit message.** Header 55 chars (limit 120), longest body line 116 (limit 132), trailers separated by a
  blank line — clean under `commitlint.config.mjs`, and the `Conventions` job agrees.
- **CI on `2de6b3d`.** All 18 non-skipped checks green (Conventions, Lint, Typecheck, Unit Tests, Train rules,
  Promotion gate, Secret Scan, Dependency Audit, both E2E installers, Panel service seam, both Core tarball
  smokes, both image builds); the six train-only jobs correctly skip.
- **Independently reproduced.** Full `packages/panel` suite in a scratch copy: **147 files, 1576 tests, all
  passing, none skipped, no timeout** — better than the body's "1571 passed, 5 skipped" plus a
  `token-usage-rollup` timeout, which did not reproduce here. The new file passed 5/5 consecutive runs.
  `pnpm typecheck` clean across all five packages; `eslint` clean on both changed files.

## Verdict

The generation guard is the right fix, and AC1 is solidly met and well tested. AC2 holds only for the
non-interleaved failure; the interleaved one regresses it in exactly the way the criterion calls out, and it is
untested. Requesting changes for the acknowledgement generation at `:174` and the missing test — both small — plus
an explicit decision on item 3.

qcentic-dev:
Thanks — the interleaving is real and the suggested shape was right. All of it is addressed in `ce6aa23`.

### Blocking 1 — the acknowledgement generation (`active-group.ts:174`)

Taken as suggested. Painting and acknowledging now ask different questions and are tracked separately:

- `acknowledge(generation, group)` carries its own generation and refuses an answer older than the one already
  recorded, so a late-landing older PATCH cannot un-acknowledge a newer one;
- it is called **before** the display guard, so a superseded-but-successful PATCH still records what the server
  confirmed while only the newest answer may paint;
- `observe` takes the newest generation when it adopts a settings read, keeping it consistent with that ordering.

Your `g-a` / `g-b` / `g-c` case now ends with the rail *and* the localStorage seed on `g-b`, and the `observe`
effect that fires on the rollback re-render adopts `g-b`, not `g-a` — so the self-entrenching half is gone too.

### Blocking 2 — the missing tests

- `rolls back to the group an older, superseded PATCH confirmed` — exactly your interleaving: server on `g-a`,
  click `g-b` then `g-c`, `g-b` succeeds, `g-c` fails, rail lands on `g-b`.
- `rolls a failed first click back to the cached group, not to All projects` — the cheaper instance you flagged.
  The ledger is now seeded from the value the rail is already rendering while settings are unhydrated, so the
  rollback goes to the operator's group instead of the `ACTIVE_GROUP_ALL` default.

Both fail against the previous shape (checked by reverting each guard in place). Four ledger cases were added for
the acknowledgement generation and the seed as well; the suite for this file is 15 tests.

### Item 3 — decided explicitly: GET half taken here, sibling-writer half filed

**Taken now.** Every write awaits `queryClient.cancelQueries({ queryKey: queryKeys.settings })` before painting
optimistically, per `UsageSettingsPage.tsx:60` / `ProvidersSettingsPage.tsx:75`. One consequence worth naming: that
await means two rapid clicks resume in an order the click order no longer settles, so the optimistic paint now
takes the same currency guard the answers do. There is a test asserting the cancel is called on the settings key
before the PATCH goes out.

**Filed, not fixed here: #452** — "Every writer of the settings row races the active group, not just
`useActiveGroup`". It covers `collapsed-groups.ts:44` and `hideable-elements.tsx`, references #384, this PR and
this review, and argues for one shared write path for the settings row rather than a third per-hook guard. Labelled
`bug`, and not started — the wave owns the ordering.

The PR body is rewritten: it now opens with a scope note saying #384's symptom is **not** fully closed by this PR
and names #452, and it has a "Known limits" section.

### Non-blocking

- **Full-row replace on confirm** — taken. The confirm path merges the group into the current row, so a
  neighbouring key edited mid-flight survives. There is a test for the collapsed-sections case you described.
- **Double settle** — taken. The request is awaited inside one `try`/`catch` and `settleWrite` runs exactly once
  per write, so a throw in what used to be the then-body can no longer settle twice.
- **Hung PATCH leaving `inFlight > 0`** — *not* taken, deliberately. The fix is an abort signal or timeout on
  `req()` in `lib/api.ts`, which no ticket in this wave owns and which would touch every caller. It is written down
  under "Known limits" in the body.
- **Test name that did not parse** — renamed to `only reports the newest write as current`.
- **`Refs` vs a closing keyword** — keeping `Refs`, now deliberately rather than by default: this stacks into
  `fix/operator-pins-sessions-drop-cli` rather than the train, and #384's symptom outlives this PR through #452, so
  auto-closing it on merge would be wrong.

One residual I would rather name than hide: with two writes out, an older success landing *after* the newer one has
already failed and rolled back leaves the rail on the rolled-back group while the ledger holds the server's newer
truth. AC2 still holds at the moment of the failure (the rail goes to what was acknowledged then), and closing it
properly needs the shared write path #452 asks for rather than more state in this hook.

Verification: `packages/panel` is 147 files / 1584 tests, all passing, none skipped; `pnpm typecheck` and `eslint`
clean. Base, draft state and file scope are unchanged.

mrfqcentic (changes requested):
Round 2, against issue #384 (B2), reading `packages/panel/src/lib/active-group.ts` at `ce6aa23` line by line and
running executable probes in a scratch copy — not from the PR description or the reply.

## Round 1 items: all six verified fixed

| Round 1 item | State |
| --- | --- |
| Blocking 1 — acknowledgement behind the staleness guard | **Fixed** (`active-group.ts:102`, `:240`) |
| Blocking 2 — the missing older-succeeds/newer-fails race | **Fixed** (`active-group.test.tsx:244`) |
| Blocking 2b — click before settings hydrate falls back to All projects | **Fixed** (`:188`, test `:269`) |
| Item 3 — GET half / sibling writers | **Decided**: `cancelQueries` here, siblings filed as #452 |
| Non-blocking — confirm-path full-row replace | **Fixed** (`:247`, test `:306`) |
| Non-blocking — double settle, test name | **Fixed** (`:228`), renamed at `:102` |

Each one is load-bearing. Reverting the guard in place makes the matching test fail, every time:

- `acknowledge` moved back behind `if (!current) return` → `rolls back to the group an older, superseded PATCH confirmed` fails.
- dropping `if (generation < acknowledgedGeneration) return` → two ledger tests fail.
- dropping `writes.seed(raw)` → `rolls a failed first click back to the cached group` fails.
- dropping the `await cancelQueries` → `cancels an in-flight settings read before writing` fails.
- `observe` not taking the newest generation → `lets an observed answer outrank every write older than the newest` fails.

And the two criteria, probed directly against `ce6aa23`:

- **AC1** — server on `g-c`, click `g-a` then `g-b`, `g-b` answers first, `g-a` answers last: rail `g-b`, `mc:activeProjectGroup` `g-b`. Holds.
- **AC2, the interleaving I blocked on** — server on `g-a`, click `g-b` then `g-c`, `g-b` succeeds, `g-c` fails: `PROBE rail = g-b localStorage = g-b`. Fixed, and the self-entrenching half with it — a further failing click still lands on `g-b`, so the `observe` effect adopted `g-b` on the rollback re-render, not `g-a`.

`#452` is well-founded, and I checked its two call sites rather than taking them on trust: `collapsed-groups.ts:44`
full-replaces the row, and `hideable-elements.tsx:122` confirms with `{ ...current, ...next, ...patch }`, where `next`
carries the pre-click `activeProjectGroup` over the optimistic one. Both revert the rail. Filing was the right call.
The PR body no longer claims #384 is closed — it opens with the scope note and carries a Known limits section.

**The hung-PATCH decline: accepted, and I'd have made the same call.** `req()` (`lib/api.ts:112`) spreads `init` into
`fetch` with no signal and no timeout, so the honest fix changes request behaviour for every caller, which no ticket
in this wave owns. The blast radius is bounded — a leaked `inFlight` freezes `lastAcknowledged()`, so a later failure
rolls back to a stale group; it does not stop the rail painting, since `settleWrite` still compares generations. I
also checked the two ways this round could have made it worse and neither does: `cancelQueries` is
`Promise.all(...).then(noop).catch(noop)` in query-core 5.101.2 and cannot reject, and `writeCachedActiveProjectGroup`
swallows storage errors, so nothing on the pre-`try` lines can throw past `settleWrite`. Listing it under Known limits
is the right resolution.

## Blocking

### 1. `packages/panel/src/lib/active-group.ts:234` — the rollback turns the deleted-group self-heal into an unbounded PATCH loop

The self-heal effect at `:255-259` fires `setActiveGroup(ACTIVE_GROUP_ALL)` whenever `raw` is a group id missing from
`groups`. Take settings holding `activeProjectGroup: "g-deleted"`: `observe` at `:189` records `acknowledged =
"g-deleted"`, the memo at `:175` renders "all", and the self-heal PATCHes "all". If that PATCH **fails**, `:234`
restores `writes.lastAcknowledged()` — which is `"g-deleted"` — into both the settings cache and
`mc:activeProjectGroup`. `raw` flips back, the effect's `[raw, groups, setActiveGroup]` dependency changes, it fires
again, fails again, restores again.

Probed with a failing PATCH that takes a realistic round trip (5 ms), 400 ms of wall clock:

```
head ce6aa23   PATCHes = 53   (still climbing)
base 0b3b099   PATCHes = 1
```

One PATCH per round trip, for as long as the server keeps rejecting — offline, a 500, a transient proxy error. The
pre-PR code swallowed the failure and left the cache on "all", so the self-heal ran exactly once. It also
re-persists a deleted group id to `mc:activeProjectGroup` on every cycle, so the bad value can outlive the tab.

To be clear about provenance: this is present at `2de6b3d` too (47 PATCHes on the same probe) — it arrived with the
rollback in round 1 and **I missed it**. It is not something `ce6aa23` introduced. But it is in the PR, and AC2's
rollback is what creates it, so it belongs to this change.

The narrow fix is to not restore a group the memo will refuse to display:

```ts
if (current) {
  const restore = writes.lastAcknowledged();
  const displayable = !isGroupIdActive(restore) || groups === undefined || groups.some((g) => g.id === restore);
  if (displayable) showActiveGroup(restore);
}
```

Either that, or have the self-heal write opt out of rollback — it has no operator selection to protect.

### 2. `packages/panel/src/lib/active-group.ts:211` — `cancelQueries` kills the boot settings GET, and nothing ever refetches it

This one is new in `ce6aa23`. The house pattern being copied lives in `UsageSettingsPage.tsx:60` and
`ProvidersSettingsPage.tsx:75`, which only render *after* settings resolve. `useActiveGroup` is live during boot:
`settingsQueryOptions`' `placeholderData: readCachedSettings()` (`queries/index.ts:212`) paints the rail from the shell
cache while the first GET is still out, and the rail is clickable there. A click in that window cancels the initial
load. In query-core 5.101.2 `Query.#fetch`'s `onCancel` reverts to `{ ...revertState, fetchStatus: "idle" }` — for a
first load that is `status: "pending"`, `data: undefined` — and schedules nothing.

Nothing in the Panel invalidates or refetches this key (`grep` for `invalidateQueries`/`refetchQueries` on
`queryKeys.settings`: zero hits) and `refetchOnWindowFocus` is false (`router.tsx:199`), so recovery waits on some
*other* `useSettings` consumer mounting fresh. Probed head against round 1, same script, a click while the first GET
is in flight:

```
ce6aa23   after click: status=pending  fetchStatus=idle     getSettings calls=1   cacheData=undefined
          settled:     status=pending  fetchStatus=idle     getSettings calls=1   cacheData=undefined
2de6b3d   after click: status=pending  fetchStatus=fetching getSettings calls=1   cacheData=undefined
          settled:     status=success  fetchStatus=idle     getSettings calls=1   cacheData=present
```

The settings row never arrives for that tab. Every `setQueryData(queryKeys.settings, (current) => current ? … : current)`
— including `showActiveGroup` at `:196`, this hook's own optimistic *and* confirm paint — is a no-op against an empty
cache, so **the rail stops responding to clicks**: in the probe the first click to `g-b` and a second click back to
`g-a` both left the rail on `g-a` while the PATCHes went out. That is a harder version of the symptom #384 exists to
fix, and it takes down the settings row for the whole app, not just the rail.

The new test cannot see it: `mount()` (`active-group.test.tsx:58`) seeds the cache directly with `refetchOnMount:
false`, so there is never a real fetch to cancel.

Options, in the order I'd weigh them: only cancel when there is something to cancel *and* something to fall back on —
skip it while the settings query has no data yet — or narrow it with `cancelQueries({ queryKey: queryKeys.settings,
fetchStatus: "fetching", type: "active" }, { revert: false })` and follow with a refetch, or drop the cancel here and
let #452's shared write path own the GET half too, since it will have to. Whichever you pick, the test needs a real
pending `queryFn` rather than a pre-seeded cache.

## Non-blocking

- `active-group.ts:188` — the comment says the ledger is seeded "with the value the rail is already rendering", but
  `raw` (`:167`) is the unvalidated id and the rail renders `activeGroup` (`:175`), which downgrades a deleted id to
  `ACTIVE_GROUP_ALL`. With settings unhydrated and `mc:activeProjectGroup` holding a deleted group, a failed first
  click rolls the rail — and localStorage — onto a group the memo will not display. Same fuel as blocking 1; seeding
  from the validated value makes the comment true and the two agree.
- `observe` re-runs on every render while settings are pending, because `useSettings = () => useQuery(settingsQueryOptions())`
  (`queries/index.ts:380`) rebuilds the `placeholderData` closure each render, so the placeholder identity changes.
  Harmless — `observe` is idempotent and sets no state — but it means the `seed` branch at `:188` is effectively dead
  whenever a shell cache exists, and `observe` treats the shell-cached row as the server's word.
- `:247` deliberately keeps only `activeProjectGroup` from `updated`. Correct for the neighbouring-key case you fixed,
  but since nothing refetches settings, the other fields of the cached row now never learn anything from a PATCH
  answer. Worth a line in #452.
- The residual you named yourself — an older success landing after the newer one already failed and rolled back —
  reproduces (`rail = g-a`, ledger `g-b`). I agree it is AC2-compliant at the moment of failure, and I confirmed it
  self-corrects: the next write that fails rolls back to `g-b`. Naming it in the body rather than hiding it was the
  right call; leave it to #452.

## Everything else checks out

- **Scope.** `2de6b3d..ce6aa23` touches exactly two files — `active-group.ts` (+95/−31) and its test (+142/−5). Whole PR
  vs base: the same two files — `active-group.ts` (+173/−10) and the new test (+344). No overlap with #448's sixteen files, and `ProjectBar.tsx`, `TerminalPane.tsx`,
  `task-status-sync.ts`, `terminal-store.tsx`, `__root.tsx`, `projects.$id.tsx` are untouched.
- **Base and state.** Base `fix/operator-pins-sessions-drop-cli` at `0b3b099`; 2 ahead, 0 behind; still a draft;
  `git merge-tree` onto the base head is conflict-free and GitHub agrees (`MERGEABLE` / `CLEAN`).
- **Commits.** Both pass `commitlint` against the repo's own config (headers 55 and 56 chars, longest body line 116,
  limit 132, trailers blank-line separated).
- **CI on `ce6aa23`.** 18 checks green, 6 train-only checks correctly skipped, none failed.
- **Reproduced locally.** `packages/panel`: 147 files, 1584 tests, all passing, none skipped. The new file is 15 tests,
  green 5/5 consecutive runs. `pnpm typecheck` clean across all five packages; `eslint` clean on both changed files.
  (One unrelated `project-dialog-core-switch` flake under load on the first full run; green on rerun and in isolation,
  and CI agrees.)

## Verdict

The round 1 work is done properly — the acknowledgement generation is the right shape, the interleaving is fixed and
tested, the seed closes the pre-hydration case, and #452 is a good ticket. Requesting changes for two defects that the
new failure and cancel paths introduce and that the tests cannot see: the deleted-group PATCH loop at `:234` (mine to
have caught in round 1) and the boot-time settings GET cancellation at `:211`. Both are contained; neither touches the
generation algebra.

qcentic-dev:
Both blocking findings are fixed in `14753cb`. Thanks for probing the failure and cancel paths rather than the
algebra — neither of these was visible from the tests I had.

### Blocking 1 — the deleted-group PATCH loop (`active-group.ts:234`)

Taken, as the narrow fix: **never restore a group the rail would refuse to display.**

What the rail shows for a stored group now has a single definition, `displayedActiveGroup(stored, groups)`, and it
has two callers: the memo that renders the active group (which is now just that call) and the failure path, which
restores `lastAcknowledged()` only when the function leaves it unchanged. So the self-heal writes "all", the PATCH
fails, nothing puts `g-deleted` back, `raw` stays "all", and the effect does not re-arm.

Test: `does not restore a deleted group when the self-heal PATCH fails` — settings holding a group the list does
not have, `updateSettings` rejecting every time, three settle rounds, asserting **exactly one PATCH**, the rail on
"All projects", `mc:activeProjectGroup` on `all`, and the cached row's `activeProjectGroup` null. With the
displayable check removed in place, it fails on the call count.

Thank you for the provenance note. It is the rollback's, so it is this PR's either way.

### Blocking 2 — `cancelQueries` killing the boot GET (`active-group.ts:211`)

Taken, and you were right that the house pattern's safety does not transfer: those pages render after settings
resolve, this hook is live behind `placeholderData` while the first GET is out. Fixed with both halves of your
first option and its fallback, because each on its own is insufficient:

1. **The cancel is now conditional** on `queryClient.getQueryData(queryKeys.settings) !== undefined` — i.e. only
   once the row has actually landed. The initial load is never cancelled, so it completes and the row arrives.
2. **A read older than the click no longer overrules it.** `beginWrite` records the selection; `guardedGroup()`
   reports it while an arriving row may still overrule it; the effect re-applies it over any settings row that
   disagrees. The guard lifts as soon as a row agrees with the selection — for a hydrated cache that is the
   optimistic write itself, one render later — so nothing changes on the normal path. A rollback calls
   `releaseGuard()` outright, since the rail no longer speaks for a selection that failed to save.

Without (1) the row never arrives; without (2) the boot GET answers with the pre-click row and the click loses. I
checked both: reverting each in place fails the new test, on `fetchStatus` and on the rail value respectively.

Test: `lets the first settings GET finish, and the click still wins` — built on a second mount helper with **no
primed row and a real `queryFn`**, so the first GET is genuinely in flight, exactly as you noted the old helper
could not be. It clicks during that fetch and asserts `fetchStatus` is still `fetching` and `getSettings` was
called once (not cancelled), then that the row arrives and the query reaches `success`, then that the rail is on
the clicked group, and finally that the cache holds it — i.e. the rail still responds to writes.

### Non-blocking

- **Seed from the validated value — taken.** `writes.seed(activeGroup)` rather than `raw`, so the comment is true
  and it cannot fuel blocking 1 from the unhydrated side.
- **`observe` re-running every render while settings are pending — left alone.** It is idempotent and sets no
  state, as you say. I did not want to hold `placeholderData` identity stable from this file.
- **The merged confirm leaving the row's other fields unable to learn from a PATCH answer — recorded on #452**
  ([comment](https://github.com/actana/control/issues/452#issuecomment-5514774329)), together with the point that
  nothing refetches this key at all, which is what makes both that and the conditional cancel above necessary. The
  shared write path should own what a PATCH answer may teach the cached row, and how the row is refreshed.
- **The older-success-after-a-rollback residual — leaving it to #452**, as agreed, and it stays named in the body.

Verification: `packages/panel` is 147 files / 1586 tests, all passing, none skipped; new file 17 tests, green 5/5
consecutive runs; `pnpm typecheck` and `eslint` clean. `commitlint` run locally over the whole range before
pushing: 0 problems on all three commits. Plain push, base, draft state and the two-file scope unchanged.

mrfqcentic (approved):
Round 3, against issue #384 (B2), reading `packages/panel/src/lib/active-group.ts` at `14753cb` line by line and
re-driving the round 2 probes in a scratch copy — not from the PR description or the reply.

## Round 2 blocking items: both fixed, both load-bearing

| Round 2 item | State |
| --- | --- |
| Blocking 1 — the rollback re-arming the deleted-group self-heal | **Fixed** (`active-group.ts:302`, via `displayedActiveGroup` at `:60`) |
| Blocking 2 — `cancelQueries` killing the boot settings GET | **Fixed** (`:268` conditional cancel + the read guard at `:236-244`) |
| Non-blocking — seed from `raw` rather than the validated value | **Fixed** (`:233`) |
| Non-blocking — `observe` re-running while settings are pending | **Declined**, and I agree — see below |
| Residuals — merged confirm / nothing refetches the row; older-success-after-rollback | **Deferred to #452**, accepted |

I re-drove both round 2 probes against `14753cb` in a scratch copy, and then reverted each guard in place to check it
is doing the work claimed for it.

**Blocking 1 — a deleted group id with a failing PATCH.** Settings holding `g-gone`, a group the loaded list does not
have, `updateSettings` rejecting after a 5 ms round trip, 400 ms of wall clock:

```
14753cb   PATCHes = 1    rail = all   mc:activeProjectGroup = all   row.activeProjectGroup = null
:302 reverted to the unconditional restore   PATCHes = 56   (still climbing)
```

Exactly one PATCH, and nothing re-persisted — the deleted id reaches neither `mc:activeProjectGroup` nor the cached
row. I also ran the unhydrated variant (the deleted id only in localStorage, no settings row at all): one PATCH there
too. `displayedActiveGroup` being one function with two callers is the right shape; the memo at `:214` is now just
that call, so the rail and the rollback cannot disagree about what is displayable by construction.

**Blocking 2 — a click during the first in-flight settings GET.** A live client with a real `queryFn` and no primed
row:

```
14753cb   after click: fetchStatus=fetching  getSettings=1
          row lands:   status=success        rail=g-b   row=g-b
```

The GET survives, the row arrives, and the click wins. I ran three variants, including the one the PR's own test does
not cover — a shell-cache `placeholderData` row painting the rail during boot (`queries/index.ts:212`), which is the
realistic boot and the case where `settings !== undefined` while `getQueryData` is still `undefined`. It holds there
too, and a second click made in the same window still lands (`rail = g-c`, `row = g-c`), so the "rail stops
responding" half is gone as well.

Both halves are necessary, exactly as the reply says. Reverted in place, one at a time:

- unconditional `cancelQueries` → `after click: fetchStatus=idle`, the row never arrives, and two clicks leave the
  rail on `all` with the cache empty — the round 2 failure verbatim.
- read guard removed (`:240-243`) → the GET survives but answers with the pre-click row and wins: `rail = g-a`, all
  three of my variants fail on the rail value.

The PR's own two new tests discriminate the same way: reverting `:302` fails only `does not restore a deleted group
when the self-heal PATCH fails`; reverting either half of blocking 2 fails only `lets the first settings GET finish,
and the click still wins`. Sixteen of seventeen pass in each case, so neither test is passing for an unrelated reason.

**The guard's shape holds up under the cases I could think to break it.** It lifts as soon as a settings row agrees
with the pending selection, which on a hydrated cache is the optimistic write itself one render later, so it is not a
lock: with a PATCH that never answers (the known limit in the body), a foreign row written to `queryKeys.settings`
still paints (`rail = g-c`). Two live `useActiveGroup` instances on one QueryClient — which `__root.tsx:271` and
`ProjectPicker.tsx:99` really are — do not fight over it: AC1 still ends on `g-b` with two PATCHes, and a failure with
both mounted rolls back once. The `groups` dependency added to `setActiveGroup` at `:318` does not re-arm the heal:
the effect below already depended on `setActiveGroup`, and after the heal's optimistic write `raw` is "all", so the
`isGroupIdActive` gate at `:324` closes it.

## Round 1 and round 2 fixes still hold

- **AC1** — server on `g-c`, click `g-a` then `g-b`, `g-b` answers first, `g-a` last: `rail = g-b`,
  `mc:activeProjectGroup = g-b`, cached row `g-b`. The stale answer paints nothing.
- **AC2, the interleaving from round 1** — server on `g-a`, click `g-b` then `g-c`, `g-b` succeeds, `g-c` fails:
  `rail = g-b`, `mc:activeProjectGroup = g-b`. The acknowledged group, not the value from before both clicks.
- A lone failure still rolls back to the acknowledged group rather than to the click; the confirm still merges rather
  than replacing (a neighbouring key edited mid-flight survives: `{"activeProjectGroup":"g-b","theme":"light"}`).
- The round 1 GET half still works where it was aimed: with the row already present, a settings refetch in flight is
  cancelled by the click (`fetchStatus` → `idle`) and its stale answer paints nothing. Making the cancel conditional
  narrowed it to the case that was actually safe, and did not give the original defect back.

## The residuals and the decline

**Accepted, all three.**

- **The sibling writers (#452).** Still the right split: `collapsed-groups.ts:44` full-replaces the row and
  `hideable-elements.tsx` merges a pre-click `activeProjectGroup` over the optimistic one. A third guard bolted onto
  this hook would not fix them.
- **The row's other fields never learning from a PATCH answer, and nothing refetching the key.** Recorded on #452 in
  [this comment](https://github.com/actana/control/issues/452#issuecomment-5514774329) with the right framing — the
  shared write path has to own both what an answer may teach the row and how the row is refreshed, and the
  conditional cancel above is downstream of the same fact.
- **An older success landing after the newer one already failed and rolled back.** AC2-compliant at the moment of
  failure, self-correcting, and genuinely not closable without the shared write path.

**The `observe`-while-pending decline: accepted, and the premise is weaker than I put it in round 2.** I measured it:
with a shell-cache row present and a click made during the boot GET, twenty forced re-renders produced **one**
`localStorage.setItem`, not twenty — query-core structurally shares the placeholder, so `data` identity is stable and
the effect does not re-run per render after all. Nothing to fix, and holding `placeholderData` identity stable from
this file would have been the wrong place for it.

## Non-blocking

A second pass over the whole PR diff turned up five things. None of them blocks: each is bounded, none regresses a
criterion, and most belong to the shared write path #452 already owns. Every claim below is measured, not read off
the code.

1. **The optimistic paint is deferred for a returning operator during boot.** `showActiveGroup` writes the cache with
   a functional updater, which is a no-op while the cache entry is empty — and for anyone with a shell cache
   (`shell-query-cache.ts:89`) `settings` is *defined* (the placeholder) while `getQueryData` is still `undefined`.
   So a click made during the boot GET does not move the rail until that GET lands. Measured, shell row `g-a`, click
   `g-b`, PATCH already answered:

   ```
   14753cb   rail while the GET is out = g-a   after it lands = g-b
   0b3b099   rail while the GET is out = g-b   after it lands = g-a
   ```

   The base painted immediately because its confirm was an unconditional `setQueryData(key, updated)` that *created*
   the entry — and then the boot GET answered with the pre-click row and reverted it **permanently**, since nothing
   refetches this key. That is #384's symptom in its worst form, and this PR is what closes it. So the deferral is the
   price of the fix and the trade is clearly right; I checked the error path too, and it recovers rather than sticking
   (`GET` errors → `data` undefined → `raw` falls back to `mc:activeProjectGroup` → rail `g-b`). Worth knowing about,
   not worth changing here.

2. **`active-group.ts:302` — refusing the restore leaves the unsaved selection in `mc:activeProjectGroup`.** In one
   narrow state — stored group deleted, settings cache still empty (boot), every PATCH failing, so `acknowledged` is
   still the deleted id — a failed click paints nothing and localStorage keeps the group the server never took:

   ```
   after the failed click:  rail = all   cache = EMPTY   mc:activeProjectGroup = g-b
   ```

   The rail is right; only the seed is stale, and a settings row overrules it, so this only surfaces if settings never
   hydrate on the next load. Restoring the *displayable equivalent* rather than skipping closes it, and the
   self-heal cannot re-arm on it because `:324` bails on a non-group id. I applied it and re-ran everything:

   ```ts
   showActiveGroup(displayedActiveGroup(restore, groups));
   ```

   17/17 of your tests still pass, my probes still pass, the deleted-group probe still issues exactly one PATCH, and
   `mc:activeProjectGroup` becomes `all`. Your call whether it is worth a line here or a note on #452.

3. **A settings GET that *starts after* the click still repaints the pre-PATCH row.** `cancelQueries` only catches
   reads already in flight at click time; `staleTime` is 30 s with `refetchOnMount` and `refetchOnReconnect` at their
   defaults, so a remount or a reconnect during the PATCH starts a fresh read that answers with the old group. The
   guard is already released by then (the optimistic write made the row agree), so it paints: `rail = g-a`, back to
   `g-b` only when the PATCH confirms. It self-corrects, unlike the boot case. But #452's *Not in scope* paragraph
   currently says "The GET half is already handled" — that is true for the read in flight at the click and not for
   this one. Worth amending there, since the shared write path is what can actually order it.

4. **`observe` still treats the shell-cached row as the server's word** — the round 2 note you declined, which I
   agree is harmless in itself. One consequence is worth writing down somewhere: because `observe` refuses to adopt
   the real boot GET while a write is in flight (`:156`), a click made during boot that then fails rolls back to the
   *shell snapshot*, not to what the server actually holds. Same family as the residuals on #452, and it needs the
   same fix.

5. **`:268` — the cancel now fires on every group click once the row is present**, and `cancelQueries` defaults to
   `revert: true`, so a settings refetch some other consumer started is reverted to its previous data and left idle
   with nothing scheduled. Same shape as the boot problem you just fixed, one level out. It matches the house pattern
   (`UsageSettingsPage.tsx:60`) exactly, so I would not change it here — it is another thing the shared write path
   should own.

Plus three small ones:

- **The PR body is now stale in a way that matters to whoever reads it next.** It still says "Every write awaits
  `queryClient.cancelQueries(...)` before painting" — that unconditional cancel is the round 2 blocking defect, and
  `:268` no longer does it. The two new mechanisms (`displayedActiveGroup`, the read guard) are not in *What changed*,
  and the counts are a round behind (15 tests / 1584 → 17 / 1586). The commit message on `14753cb` is accurate and
  complete; the body is what the wave owner will read. Worth one pass before this leaves draft.
- The older-success-after-rollback residual is named in the body's *Known limits* but not on #452 itself. A line there
  would let it outlive the PR page, since #452 is what survives as the ticket.
- `displayedActiveGroup` is exported but has no consumer outside this module (the test file does not import it
  either). Harmless; `export` is arguably documentation here.

## Everything else checks out

- **Diff `ce6aa23..14753cb`.** Exactly two files: `active-group.ts` (+92/−24) and its test (+85/−11). Whole PR vs base:
  the same two, `active-group.ts` (+248/−17) and the new test (+418). No overlap with the two commits the base branch
  has gained since (`8ed1473`, `e68c354` — Core sweep/relaunch, `TerminalPane.tsx`, `task-status-sync.ts`,
  `harness-command.ts`), and `ProjectBar.tsx`, `terminal-store.tsx`, `__root.tsx`, `projects.$id.tsx` are untouched.
- **Commits.** All three pass `commitlint` against the repo's own `commitlint.config.mjs` (`found 0 problems, 0
  warnings` each). Headers 55 / 56 / 75 chars against a 120 limit; longest body line 116 against a 132 limit.
- **Base and state.** Base `fix/operator-pins-sessions-drop-cli`, 3 ahead / 2 behind its current head `e68c354`; still
  a draft; `git merge-tree` onto that head is conflict-free and GitHub agrees (`MERGEABLE` / `CLEAN`).
- **CI on `14753cb`.** 18 checks green (Conventions, Lint, Typecheck, Unit Tests, Train rules, Promotion gate, Secret
  Scan, Dependency Audit, both E2E installers, Panel service seam, both Core tarball smokes, both image builds,
  Resolve PR image mode), 6 train-only jobs correctly skipped, none failed.
- **Reproduced locally in a scratch copy.** `packages/panel`: 147 files, **1586 tests, all passing, none skipped** —
  matching the reply, twice, with no flake either time. The new file is 17 tests, green 5/5 consecutive runs, no
  `.only`/`.skip`. `tsc --noEmit` clean across all five packages; `eslint` clean on both changed files.

## Verdict

Both blocking findings are properly fixed, the fixes are minimal and in the right places, and each is defended by a
test that fails when the guard is reverted in place. The round 1 and round 2 work still holds, the residuals left to
#452 are the ones that genuinely need the shared write path, and the decline is correct.

The five non-blocking items above came out of a fresh pass over the whole diff rather than the round 2 list. I
weighed each against the bar for a third round and none of them meets it: item 1 is the deliberate cost of the fix
and strictly better than what the base did, items 3–5 are properties of a settings row that no single hook can order
and belong to #452 by construction, and item 2 is a one-line refinement in a state that needs a deleted group, an
empty cache and a failing server at once. Nothing here regresses a criterion or a round 1/round 2 fix, so I would
rather this landed and #452 took the rest than spend a fourth round on it.

**Approving.** Before it leaves draft, please refresh the PR body — it still describes the unconditional
`cancelQueries` that round 2 blocked on — and consider the one-line amendment to #452's *Not in scope* paragraph
from item 3.

PR #453, opened 2026-09-02
fix(core): Stop with a foreign session_id still settles the task

qcentic-dev:
## Summary

A `Stop` is not a session-capture event, so `harness-hook-pipeline.ts`'s foreign-session guard returned **before** `updateStatus` for any turn end whose `session_id` was not the stored one — and it returned an *ack*: `{ ok: true, ignored: "foreign-session" }`. The card stayed on `running`, no `session:finished` fired, and the harness saw its hook accepted. Nothing else was coming either: the id is only re-captured by `UserPromptSubmit` / `SessionStart`, and a resumed harness that already submitted its prompt fires neither again.

The hook is addressed by **task id**, read out of the PTY's own environment, so the PTY that posted a turn end belongs to this task whatever the harness calls its session. `Stop` — and Cursor's `stop` / `afterAgentResponse` — now settle instead of being dropped.

Two shapes reach that branch, and neither is a clean one:

- **A resume whose capture event was lost.** A clean resume never gets there — `SessionStart` is a capture event, so it adopts the new id, clears the dead process's subagents, and the following `Stop` matches. This branch fires when that POST dropped, and the stored id belongs to a process that is gone.
- **An OpenCode child** whose `session.idle` leaked past the plugin's own `parentID` filter. There the parent may still be working.

### What `settleForeignTurnEnd` does

- **Drops the tracked subagents on entry, with the recent-finish mark.** A turn end under an unrecognised id means a new harness process, so the old session's subagents died with it — the same sentence the `setSessionId` wrapper already writes. Nothing else can clear them: their `SubagentStop` died with the process, and the resumed session's own subagent events carry the new id, so they hit the foreign return *before* the bookkeeping. `clearTaskFinished` goes with it by the `sessionProcessExited` precedent; it drops a mark rather than suppressing one, since `noteTaskFinished` re-stamps when the settle proceeds.
- **Has no subagent hold, deliberately.** After the clear the set is empty by construction, and the hold could not have engaged in either shape anyway: a resumed session's subagent events are dropped as foreign before they are ever counted, and opencode posts *no* subagent lifecycle events at all, so `hasActiveSubagents` is structurally false for every opencode Session. **The status check is the only real guard.** A leaked child *can* finish a working parent's `running` card; that residual is stated in the code, the docs and §Residual risk below rather than papered over.
- **Settles `running` only.** Not `needs-input`: the PTY-exit settle acts on both, and that is not a precedent that transfers — there the process is dead, so an open question is moot, while here it is alive and may be blocked on exactly that question. A leaked child going idle would otherwise write `finished` over a parent waiting on a permission prompt, taking the Panel's pending-question overlay with it. `ready` is out for #387's reason: `CoreTaskWriter` appends `session:finished` on that transition, and a Session titled "Waiting for initial prompt…" has no turn to end.
- **Does not capture the session id.** A `Stop` is not a capture event; adopting an OpenCode child's id would hand the rest of that child's lifecycle the Session's card.

Everything else the guard rejects still is: a foreign question, subagent count, permission prompt or interrupt claims the task for a session it does not own, and the capture events go on adopting a new id exactly as before. The A-heal race window and the `ready` sweep landed by #385 and #387 are built on, not rewritten — `subagent-activity.ts` is not touched at all.

## Review round 1 — what changed

Review at `9e392ea` requested changes; `f73d107` addresses all four findings. See [the reply comment](https://github.com/actana/control/pull/453#issuecomment-5515445937) for the point-by-point.

1. **BLOCKING — stale subagents wedged the settle for ~2h.** Fixed by `clearSubagentActivity` + `clearTaskFinished` on entry. Pinned by three new tests plus a Core-side one.
2. **The hold could not engage in either named scenario.** Dropped, and said plainly in the code comment, the docs bullet and this body. The test that posted subagent events under the *matching* id is replaced by one posting them under the **foreign** id, showing them dropped and the settle proceeding.
3. **`needs-input` settled over a live process.** Restricted to `running` only, with the reasoning recorded.
4. **Nit — the "posts are queued" doc bullet.** Reworded; it no longer justifies itself with a `Stop` that overtakes `SessionStart` being discarded, and names the orderings that still break.

## Related issues

Refs #390

## Type of change

- [x] 📚 Documentation
- [ ] 🔧 Build / CI / tooling

## How was this tested?

`packages/shared/src/__tests__/harness-hook-pipeline.test.ts` — `a turn end from a session this task never captured (issue 390)`:

- finishes a running task on a `Stop` whose `session_id` does not match, and answers with the status rather than the `ignored: "foreign-session"` ack;
- **settles a fanned-out turn whose resume lost its `SessionStart`** — the ~2h wedge from finding 1;
- **drops the tracked subagents on entry, so the next turn's owned `Stop` is not held**;
- **does not hold on subagents posted under the foreign id either** — they return `foreign-session`, and the settle proceeds;
- **leaves a task parked on `needs-input` waiting** (finding 3), with the exact result object and write sequence;
- settles Cursor's `stop` / `afterAgentResponse` on the same terms;
- does not adopt the foreign session id;
- leaves an already-settled task and a `ready` Session exactly as they were;
- every other foreign event is still dropped as `foreign-session`.

Plus a `matching-session Stop behaviour is unchanged` block pinning the owned path end to end (finish, the no-captured-id case, the subagent downgrade, and the unconditional finish over a settled task).

`packages/core/src/__tests__/harness-status-detection.test.ts` — on the Core: a `Stop` under a post-resume id finishes the row **and** appends `session:finished`; the same with a live fan-out tracked (finding 1); and a Session on `needs-input` is left waiting with no ding (finding 3).

`packages/panel/src/server/__tests__/opencode-hooks-api.test.ts` — over HTTP: a foreign claiming event (`PermissionRequest`) is still dropped; a foreign `Stop` answers `{ ok: true, status: "finished" }` with the stored id untouched; and a leaked child's `Stop` over a `needs-input` parent answers `{ ok: true, event: "Stop" }` and changes nothing.

Local runs, all green: `pnpm lint` (0 errors), `pnpm typecheck`, `pnpm test` — all 6 stages, 331 test files. Shared 27, core 19, panel hook suites 53.

## Residual risk

A leaked OpenCode child's `session.idle` with no matching parent activity will finish a parent that is still working on `running`. Nothing in this pipeline can see that work — opencode posts no subagent lifecycle events — so the status check is the only guard, and this is the trade the issue itself frames ("or nack visibly; ack-and-ignore is the operator-visible miss"). Three things bound it: the plugin filters children at the source by `parentID`, so a leak needs a child created before the plugin loaded; a parent blocked on a permission prompt is now excluded outright; and the parent's next `session.status: busy` puts the card back on `running`.

## Checklist

- [x] This PR targets the **open train** (`beta/x.y.z`), not `main` — stacked on `fix/operator-pins-sessions-drop-cli`, which targets the train
- [x] My PR title **and every commit in the branch** follow [Conventional Commits](https://www.conventionalcommits.org/), and my branch name follows the [branching convention](../CONTRIBUTING.md#branch-naming) — `commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD` run locally over both commits: 0 problems
- [x] I updated documentation (README, docs/, changelog entry) where relevant — `docs/harness-status-detection.md`: the session-id guard section and the OpenCode bullets
- [x] Dependent changes have been merged (or this PR is marked as draft/stacked) — draft, stacked on `fix/operator-pins-sessions-drop-cli`

https://claude.ai/code/session_01HFY4g9AwbTLVX94heQZKp5

### DIFF
--- a/docs/harness-status-detection.md
+++ b/docs/harness-status-detection.md
@@ -159,6 +159,42 @@ notifications as `needs-input` creates false positives that later flip to
 These values were tuned against the failure modes above. Do not change them
 without a failure that says so.
 
+### The session-id guard, and the one event it does not drop
+
+A hook is addressed by **task id**, read out of the PTY's own environment. The
+`session_id` in the payload is a second, weaker fact: the harness's own name for
+the conversation. The pipeline stores the first one it sees on a *capture* event
+(`UserPromptSubmit`, `SessionStart`, and their Cursor spellings) and, from then
+on, drops any non-capture event carrying a different id — a `foreign-session`.
+That is what stops a stranger's question, subagent count or permission prompt
+from claiming a card it does not own, and a capture event is still free to adopt
+a new id, because a new session id means a new harness process.
+
+**A turn end is the exception** (#390). `Stop` — and Cursor's `stop` and
+`afterAgentResponse` — claims nothing; it reports that a turn ended, on a PTY
+that is this task's whatever the harness calls its session. Two shapes produce
+one under an unrecognised id: a **resume**, where the stored id belongs to a
+process that is gone and no further capture event is coming, and an **OpenCode
+child** whose `session.idle` leaked past the plugin's parent/child filter.
+Dropping it was still an *ack* — `{ ok: true, ignored: "foreign-session" }` — so
+the harness saw its hook accepted while the card sat on `running` and no
+`session:finished` ever fired.
+
+So a foreign turn end settles the task instead, on terms narrower than an owned
+one's:
+
+- **Only `running` and `needs-input` are settled** — the same pair the PTY-exit
+  settle acts on. Every other status is an answer already given. `ready` is
+  excluded for #387's reason: `CoreTaskWriter` appends `session:finished` on
+  that transition, and a Session titled "Waiting for initial prompt…" has no
+  turn to end.
+- **The subagent hold still applies.** A foreign `Stop` over a live fan-out
+  holds on `running` and arms the same drain backstop, rather than dinging
+  mid-work.
+- **The session id is not captured.** A `Stop` is not a capture event; adopting
+  an OpenCode child's id would hand the rest of that child's lifecycle the
+  Session's card.
+
 ## The hook receiver
 
 The Core exposes a small **loopback HTTP** listener a hook's shell command can
@@ -515,7 +551,11 @@ Three details are load-bearing and none of them is guessable from the docs:
   Nothing awaits the queue, so a hook still never holds up a turn.
 - **A child session is a subagent.** `session.created` carries `parentID`, so
   the plugin knows a child by name rather than guessing from ordering, and a
-  subagent's `idle` never settles the Session's card.
+  subagent's `idle` never settles the Session's card. Filtering it at the
+  source is the point: one that leaks through — a child created before the
+  plugin loaded — reaches the Core as a `Stop` under an unrecognised session
+  id, which now settles rather than being dropped (see the session-id guard
+  above), held back only by the subagent bookkeeping OpenCode does not feed.
 
 `permission.replied` is also why opencode needs no unmatched `PostToolUse`
 subscription: Claude Code fires nothing when a permission is *granted* and has
--- a/docs/harness-status-detection.md
+++ b/docs/harness-status-detection.md
@@ -183,14 +183,32 @@ the harness saw its hook accepted while the card sat on `running` and no
 So a foreign turn end settles the task instead, on terms narrower than an owned
 one's:
 
-- **Only `running` and `needs-input` are settled** — the same pair the PTY-exit
-  settle acts on. Every other status is an answer already given. `ready` is
-  excluded for #387's reason: `CoreTaskWriter` appends `session:finished` on
-  that transition, and a Session titled "Waiting for initial prompt…" has no
-  turn to end.
-- **The subagent hold still applies.** A foreign `Stop` over a live fan-out
-  holds on `running` and arms the same drain backstop, rather than dinging
-  mid-work.
+- **The tracked subagents are dropped on entry**, with the recent-finish mark.
+  A turn end under an unrecognised id means a new harness process, so the old
+  session's subagents died with it — and nothing else can clear them. Their
+  `SubagentStop` died with the process, and the new session's own subagent
+  events carry the new id, so they are dropped as foreign *before* the
+  bookkeeping. The only other route out is the 2-hour `ACTIVE_SUBAGENT_TTL_MS`,
+  and waiting on it is how the first cut of this settle reproduced #390 inside
+  the fix for it: a fanned-out turn that resumed with a lost `SessionStart` sat
+  on `running` for two hours with its `Stop` acked.
+- **There is no subagent hold, deliberately.** After the clear the tracked set
+  is empty by construction, and it could not have engaged in either shape that
+  reaches here anyway: a resumed session's subagent events are dropped as
+  foreign before they are ever counted, and opencode — whose child sessions are
+  the other shape — posts no subagent lifecycle events at all, so
+  `hasActiveSubagents` is structurally false for every opencode Session. **The
+  status check below is the only real guard**, and a leaked child *can* finish
+  a working parent's `running` card. That is the residual, stated rather than
+  papered over; the parent's next `session.status: busy` corrects it.
+- **Only `running` is settled** — not `needs-input`. The PTY-exit settle acts on
+  both, and it is not a precedent that transfers: there the process is dead, so
+  an open question is moot, while here it is alive and may be blocked on exactly
+  that question. A leaked child going idle would otherwise write `finished` over
+  a parent waiting on a permission prompt, taking the pending-question overlay
+  with it. `ready` is out for #387's reason: `CoreTaskWriter` appends
+  `session:finished` on that transition, and a Session titled "Waiting for
+  initial prompt…" has no turn to end.
 - **The session id is not captured.** A `Stop` is not a capture event; adopting
   an OpenCode child's id would hand the rest of that child's lifecycle the
   Session's card.
@@ -545,17 +563,25 @@ Three details are load-bearing and none of them is guessable from the docs:
 - **`permission.asked` is the event the binary emits.** The `@opencode-ai/sdk`
   type union calls it `permission.updated`, and that string does not appear in
   the 1.18.18 binary at all. The plugin listens for both.
-- **Posts are queued, not fired in parallel.** A `Stop` that overtook the
-  `SessionStart` carrying the session id would be discarded by the Core as
-  belonging to a session it has never heard of — #230 again with a new cause.
-  Nothing awaits the queue, so a hook still never holds up a turn.
+- **Posts are queued, not fired in parallel.** Order is meaning here. A
+  `chat.message` that overtook the `SessionStart` before it would name the
+  Session from a prompt the Core cannot yet attribute, and a `permission.replied`
+  that overtook its `permission.asked` would leave the card on `needs-input`
+  for a question already answered — #230 again with a new cause. (A `Stop` that
+  overtakes the capture event is no longer among the casualties: since #390 it
+  settles rather than being discarded. That is a narrowed blast radius, not a
+  reason to stop queueing.) Nothing awaits the queue, so a hook still never
+  holds up a turn.
 - **A child session is a subagent.** `session.created` carries `parentID`, so
   the plugin knows a child by name rather than guessing from ordering, and a
   subagent's `idle` never settles the Session's card. Filtering it at the
-  source is the point: one that leaks through — a child created before the
-  plugin loaded — reaches the Core as a `Stop` under an unrecognised session
-  id, which now settles rather than being dropped (see the session-id guard
-  above), held back only by the subagent bookkeeping OpenCode does not feed.
+  source is the whole guard — one that leaks through, a child created before
+  the plugin loaded, reaches the Core as a `Stop` under an unrecognised session
+  id and now settles rather than being dropped (see the session-id guard
+  above). Nothing downstream can catch it: opencode posts no subagent lifecycle
+  events, so the pipeline has no way to see the parent still working. All that
+  narrows it is the `running`-only status check, and the parent's next
+  `session.status: busy` puts the card back.
 
 `permission.replied` is also why opencode needs no unmatched `PostToolUse`
 subscription: Claude Code fires nothing when a permission is *granted* and has
--- a/packages/core/src/__tests__/harness-status-detection.test.ts
+++ b/packages/core/src/__tests__/harness-status-detection.test.ts
@@ -161,6 +161,21 @@ describe("harness status detection on the Core (issue 84)", () => {
     expect(kinds()).toContain("session:finished");
   });
 
+  it("finishes on a Stop whose session id is not the stored one (issue 390)", async () => {
+    // The operator-visible miss: a resumed harness (or an OpenCode child whose
+    // idle leaked past the plugin's filter) posts its Stop under a session id
+    // this task never captured. It used to be acked as `foreign-session` and
+    // dropped before any status write, so the card stayed on `running` and the
+    // notification consumer never heard a thing.
+    await postHook({ hook_event_name: "UserPromptSubmit", session_id: "sess-1" });
+    expect(rowStatus()).toBe("running");
+
+    await postHook({ hook_event_name: "Stop", session_id: "sess-2-after-resume" });
+
+    expect(rowStatus()).toBe("finished");
+    expect(kinds()).toContain("session:finished");
+  });
+
   it("holds the finish while a background subagent is still working", async () => {
     await postHook({ hook_event_name: "UserPromptSubmit", session_id: "sess-1" });
     await postHook({
--- a/packages/core/src/__tests__/harness-status-detection.test.ts
+++ b/packages/core/src/__tests__/harness-status-detection.test.ts
@@ -176,6 +176,37 @@ describe("harness status detection on the Core (issue 84)", () => {
     expect(kinds()).toContain("session:finished");
   });
 
+  it("finishes a fanned-out turn whose resume lost its SessionStart (issue 390)", async () => {
+    // The pre-resume process's subagents can never report in — their harness is
+    // gone, and the resumed session's own subagent events carry the new id, so
+    // they are dropped as foreign. Holding on that stale set was two hours of a
+    // card on `running` with its Stop acked: #390's symptom inside the fix.
+    await postHook({ hook_event_name: "UserPromptSubmit", session_id: "sess-1" });
+    await postHook({
+      hook_event_name: "SubagentStart",
+      session_id: "sess-1",
+      agent_id: "sub-1",
+    });
+
+    await postHook({ hook_event_name: "Stop", session_id: "sess-2-after-resume" });
+
+    expect(rowStatus()).toBe("finished");
+    expect(kinds()).toContain("session:finished");
+  });
+
+  it("leaves a Session waiting on a permission prompt alone (issue 390)", async () => {
+    // `needs-input` is not settled by a foreign turn end: unlike the PTY-exit
+    // settle, the process here is alive and may be blocked on that question.
+    await postHook({ hook_event_name: "UserPromptSubmit", session_id: "sess-1" });
+    await postHook({ hook_event_name: "PermissionRequest", session_id: "sess-1" });
+    expect(rowStatus()).toBe("needs-input");
+
+    await postHook({ hook_event_name: "Stop", session_id: "sess-2-child" });
+
+    expect(rowStatus()).toBe("needs-input");
+    expect(kinds()).not.toContain("session:finished");
+  });
+
   it("holds the finish while a background subagent is still working", async () => {
     await postHook({ hook_event_name: "UserPromptSubmit", session_id: "sess-1" });
     await postHook({
--- a/packages/panel/src/server/__tests__/opencode-hooks-api.test.ts
+++ b/packages/panel/src/server/__tests__/opencode-hooks-api.test.ts
@@ -283,6 +283,29 @@ describe("OpenCode hook API", () => {
     });
   });
 
+  it("does not settle a Session waiting on a permission prompt (issue 390)", async () => {
+    // The parent raised the prompt and is blocked on it; a child session going
+    // idle must not write `finished` over that, which would also clear the
+    // pending question the operator still has to answer.
+    const capturedSessionId = "ses_captured_session";
+
+    await start(capturedSessionId);
+    await postHook({ hook_event_name: "PermissionRequest", session_id: capturedSessionId });
+    expect(getTask(taskId)?.status).toBe("needs-input");
+
+    const childStop = await postHook({
+      hook_event_name: "Stop",
+      session_id: "ses_leaked_child",
+    });
+
+    expect(childStop?.status).toBe(200);
+    await expect(childStop?.json()).resolves.toEqual({ ok: true, event: "Stop" });
+    expect(getTask(taskId)).toMatchObject({
+      claudeSessionId: capturedSessionId,
+      status: "needs-input",
+    });
+  });
+
   it("still settles the Session on a Stop from a different session (issue 390)", async () => {
     // A resumed OpenCode process, or a child session whose `idle` slipped past
     // the plugin's parent/child filter. This used to be acked and dropped, so
--- a/packages/panel/src/server/__tests__/opencode-hooks-api.test.ts
+++ b/packages/panel/src/server/__tests__/opencode-hooks-api.test.ts
@@ -47,6 +47,27 @@ describe("OpenCode hook API", () => {
     taskId = task.id;
   });
 
+  function postHook(body: Record<string, unknown>): Promise<Response | null> {
+    return handleApiRequest(
+      authed(`/api/hooks/opencode?taskId=${encodeURIComponent(taskId)}`, {
+        method: "POST",
+        headers: { "content-type": "application/json" },
+        body: JSON.stringify(body),
+      }),
+    );
+  }
+
+  /** Capture a session id and put the Session on `running`, as a turn does. */
+  async function start(sessionId: string): Promise<void> {
+    const running = await postHook({
+      hook_event_name: "UserPromptSubmit",
+      session_id: sessionId,
+      prompt: "ship opencode hooks",
+    });
+    expect(running?.status).toBe(200);
+    expect(getTask(taskId)).toMatchObject({ claudeSessionId: sessionId, status: "running" });
+  }
+
   it("captures ses_* session ids from SessionStart without changing status", async () => {
     const sessionId = "ses_3cf7dd8d4ffeUPfENpVxfFojZ2";
     const res = await handleApiRequest(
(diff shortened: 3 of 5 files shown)


qcentic-dev:
## Issue #390 acceptance — every criterion, checked or not

Head: `9e392ea` on `fix/foreign-session-stop`, base `fix/operator-pins-sessions-drop-cli` (`8ed1473`).

---

- [x] **1. A Stop whose `session_id` does not match the stored id still finishes (or nacks) the Core task the PTY belongs to.**

  It finishes. `handleHarnessHookEvent` no longer returns `foreign-session` for a turn end: `Stop`, and Cursor's `stop` / `afterAgentResponse`, route into `settleForeignTurnEnd` instead of being acked and dropped before `updateStatus`.

  The task the PTY belongs to is not in question — the hook is addressed by task id out of the PTY's own environment — so the settle acts on that row, and on the Core it appends `session:finished` like any other finish.

  Deliberately narrower than an owned turn end in three places, and each is pinned:
  - only `running` and `needs-input` are settled (the pair the PTY-exit patch acts on); `ready` is excluded so a Session titled "Waiting for initial prompt…" is never dinged for a turn it never ran (#387's reasoning);
  - the subagent hold still applies, so a foreign turn end over a live fan-out holds on `running` and arms the same drain backstop instead of dinging mid-work — this is what keeps a leaked OpenCode child's `idle` from finishing a working parent;
  - the foreign session id is **not** captured; a `Stop` is still not a capture event.

  Proof:
  - `packages/shared/src/__tests__/harness-hook-pipeline.test.ts` — "finishes a running task on a Stop whose session_id does not match", "answers with the status rather than an ignored foreign-session ack", "settles a task parked on needs-input too", "settles Cursor's turn-end events on the same terms", "does not adopt the foreign session id", "holds instead of finishing while a tracked subagent is in flight", "leaves an already-settled task exactly as it settled", "leaves a ready Session alone rather than dinging a turn it never ran".
  - `packages/core/src/__tests__/harness-status-detection.test.ts` — "finishes on a Stop whose session id is not the stored one (issue 390)": the row reaches `finished` **and** `session:finished` is appended, which is the event the operator's card and #20's notification consumer read.
  - `packages/panel/src/server/__tests__/opencode-hooks-api.test.ts` — "still settles the Session on a Stop from a different session (issue 390)": over HTTP, `{ ok: true, status: "finished" }` where the endpoint used to answer `{ ok: true, ignored: "foreign-session" }`.

- [x] **2. Matching-session Stop behaviour is unchanged.**

  Structurally: the matching path was not edited. `settleForeignTurnEnd` is a new function reached only from inside the `foreign-session` branch, and the main body of `handleHarnessHookEvent` below that branch is byte-identical to the base.

  Pinned anyway, by a dedicated `matching-session Stop behaviour is unchanged (issue 390)` block in `harness-hook-pipeline.test.ts`:
  - a matching `Stop` returns `{ outcome: "ok", event: "Stop", status: "finished" }`, writes exactly `["running", "finished"]`, and leaves the stored session id alone;
  - a `Stop` on a task that captured no session id at all still finishes (the guard never fires without a stored id — every Session before its first capture event);
  - a matching `Stop` is still downgraded to `running` while a subagent is in flight, and the next one finishes;
  - a matching `Stop` still finishes **unconditionally**, including over an already-settled (`interrupted`) task — the asymmetry with the foreign settle is intentional and stated in the code.

  The pre-existing matching-Stop tests are untouched and green: the whole `issue 385` subagent block, `in-turn subagents hold the finish`, the `issue 387` PTY-exit block, `harness-hooks-api.test.ts`, `ask-user-question-api.test.ts`, and the OpenCode idle-finish test ("finishes a settled turn the Core can report", the `session.idle` → `Stop` path).

---

### Scope, explicitly

- **Not broadened into the A-heal race window or the `ready` sweep / relaunch reset.** `FINISH_RACE_WINDOW_MS`, `taskFinishedWithinRaceWindow`, the subagent branch and the `sessionProcessExited` branch that #385 and #387 landed in this file are unmodified — this change builds on them (it reuses `hasActiveSubagents`, `armDeferredFinish`, `noteTaskFinished` and `finishQuietTask` unchanged) and rewrites none of them.
- **Owned boundaries respected.** Nothing touched in `TerminalPane.tsx` or `task-status-sync.ts` (#386 / PR #449), `active-group.ts` (#384 / PR #450), `terminal-store.tsx`, `__root.tsx`, `ProjectBar.tsx` or `projects.$id.tsx`.
- Files changed: `packages/shared/src/harness-hook-pipeline.ts`, three test files, and `docs/harness-status-detection.md`.

### Local verification

`pnpm lint` — 0 errors (45 pre-existing warnings). `pnpm typecheck` — all 5 packages clean. `pnpm test` — all 6 stages green (331 test files). `commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD` — 0 problems, and the PR title lints clean too.

https://claude.ai/code/session_01HFY4g9AwbTLVX94heQZKp5

qcentic-dev:
### CI status of the PR head

`9e392ea` — **all checks green**: 18 pass, 6 skipping, 0 failing, 0 pending. Run [33671046198](https://github.com/actana/control/actions/runs/33671046198).

| Check | Result |
| --- | --- |
| Unit Tests | pass |
| Typecheck | pass |
| Lint | pass |
| Conventions (title **and** every commit) | pass |
| Train rules | pass |
| Promotion gate | pass |
| Secret Scan | pass |
| Dependency Audit | pass |
| E2E — installer (ubuntu, x64) / (debian, x64) | pass / pass |
| Smoke — Core release tarball (linux-x64) / (linux-arm64) | pass / pass |
| Panel image · Core image — resolve registries, build + smoke (amd64) | pass |
| Resolve PR image mode | pass |

Skipping (6): `Core image (train)`, `Panel image (train)`, `Resolve train tags`, `Train versions`, and the two multi-arch manifest publishes — all train-only jobs that do not run on a stacked ticket PR.

Note: the base branch has advanced to `e68c354` since this branch was cut at `8ed1473` (PR #449 for issue 386 landed on it). This branch touches none of those files and `git merge-tree` reports a clean merge, so it was **not** rebased — rebasing would need a force push, which this branch will not do.

The PR is left as a **draft**, base `fix/operator-pins-sessions-drop-cli`, and nothing was merged.

mrfqcentic (changes requested):
## Review of #453 against issue #390 — changes requested

Reviewed at head `9e392ea`, base `fix/operator-pins-sessions-drop-cli`, in a clean worktree. Every enforcing line read; the shared, core and panel hook suites re-run locally in a scratch copy (25 + 17 + 52 tests, all green), and CI on `9e392ea` is fully green (18 pass, 6 skipping).

The shape of the fix is right, and the discipline around it is genuinely good: exactly one line deleted from `harness-hook-pipeline.ts`, so the matching-session path and the #385 / #387 blocks are provably untouched; the id is deliberately not captured; `ready` is correctly excluded; `isTurnEndEvent` matches the `finished` mapping in `harness-hook-events.ts` exactly; `session:finished` is transition-guarded in `CoreTaskWriter`, so the later owned `Stop` cannot double-ding.

But the subagent hold in the new path is wrong in **both** directions, and one of them leaves acceptance criterion 1 unmet in the primary scenario the issue names.

---

### 1. BLOCKING — `packages/shared/src/harness-hook-pipeline.ts:446`: stale subagent entries wedge the foreign settle for up to two hours

```ts
if (hasActiveSubagents(taskId)) {
  ports.updateStatus(taskId, "running");
  armDeferredFinish(taskId, (id) => finishQuietTask(id, ports));
  return { outcome: "ok", event };
}
```

Consider when this branch actually fires for a **resume** — the first of the two shapes the comment at `:221-226` names.

A clean resume never reaches it. Claude Code fires `SessionStart` on resume, `isSessionCaptureEvent` (`:107-114`) accepts it, and the wrapper at `:192-200` adopts the new id *and* calls `clearSubagentActivity` — after which the `Stop` matches and takes the owned path. So the foreign branch fires on a resume precisely when the capture event was **lost**. And in that case nothing has cleared the pre-resume process's tracked subagents.

Those entries are unclearable by any other route:

- their `SubagentStop` can never arrive — the process that would have sent it is gone;
- the resumed session's own `SubagentStart`/`SubagentStop` carry the new id, so they hit the foreign return at `:227` *before* the bookkeeping at `:270-284` and are dropped;
- `hasActiveSubagents` prunes only on `ACTIVE_SUBAGENT_TTL_MS` — **2 hours** (`packages/shared/src/subagent-activity.ts:30`);
- `armDeferredFinish`'s drain grace does not start until the set goes idle, so the backstop is 2h + `DRAIN_FINISH_GRACE_MS` (3 min) away.

So: a turn that fanned out, then resumed with a lost `SessionStart`, posts its `Stop`, and the card sits on `running` for ~2 hours while the hook is acked `{ ok: true, event: "Stop" }`. That is issue #390's exact symptom — card stays running, no `session:finished`, harness sees an ack — reproduced inside the fix for it. Acceptance criterion 1 ("still finishes, or nacks") is not delivered here: it neither finishes nor nacks.

Note this is not hypothetical framing — a resume typically *follows* an interrupted or crashed turn, which is exactly when subagents were mid-flight.

The fix looks small and follows reasoning already in this file: a turn end under an unrecognised id means a new harness process, so the old session's subagents died with it — the same sentence as the `setSessionId` wrapper's comment at `:196-198`. Calling `clearSubagentActivity(taskId)` (and, by the `sessionProcessExited` precedent at `:316-319`, likely `clearTaskFinished`) on entry to `settleForeignTurnEnd` would close it. That also makes the settle honest rather than accidentally conditional.

### 2. `packages/shared/src/harness-hook-pipeline.ts:425-429` and `docs/harness-status-detection.md:191`: the "subagent hold still applies" guarantee cannot engage in either named scenario

The mirror image of finding 1. Where the tracked set is *empty* — the common case — the hold is claimed as the safeguard and is inert:

- **Resume**: as above, the resumed session's subagent events are dropped as foreign before `noteSubagentStart` is ever reached, so `hasActiveSubagents` is false through a live fan-out and the first foreground `Stop` finishes and dings mid-work.
- **OpenCode child**: `packages/core/src/harness-hooks-opencode.ts` posts only `UserPromptSubmit`, `SessionStart`, `Stop`, `PermissionRequest` and `PermissionReplied` — no subagent lifecycle events at all, ever. `hasActiveSubagents` is structurally false for every OpenCode Session.

`docs/harness-status-detection.md:558` is honest about this ("held back only by the subagent bookkeeping OpenCode does not feed"), but the code comment at `:425-429`, the doc bullet at `:191` and the PR body all present the hold as the thing that "keeps a leaked OpenCode child's `idle` from finishing a working parent". It cannot. The only real guard is the status check on the line above.

The new test at `packages/shared/src/__tests__/harness-hook-pipeline.test.ts:356` passes because it posts its `SubagentStart` under the **matching** session id (the harness default) and drains it with a matching `SubagentStop`. That models a case that cannot occur: in a real foreign turn end, the subagent events are foreign too. Worth a test that posts the subagent events under `FOREIGN` as well — it will show the hold not engaging.

If finding 1 is fixed by clearing activity on entry, the honest resolution here is to drop the hold and say so, rather than keep a branch that only fires on stale state.

### 3. `packages/shared/src/harness-hook-pipeline.ts:440`: settling `needs-input` is worth a second look

```ts
if (task.status !== "running" && task.status !== "needs-input") {
```

The justification given — "the same pair the PTY-exit settle acts on" — is accurate as a description (`:320`) but the reasoning does not transfer: on PTY exit the process is dead, so an open question is moot; here the process is alive and may be blocked on it.

Concretely for the OpenCode child shape: parent raises `permission.asked` → card `needs-input`; a child whose `session.created` the plugin never saw goes idle → foreign `Stop` → the task is written `finished`, and `packages/panel/src/server/services/tasks.ts:97-100` clears any pending question on a status transition away from `needs-input`. The parent is still waiting on the prompt.

This one does self-correct — the parent holds the stored id, so its `permission.replied` / `session.idle` land on the matching path — so the damage is a premature ding and a card reading `finished` until the operator acts. Restricting the foreign settle to `running` would keep the whole #390 fix and remove the case; if `needs-input` is kept deliberately, the comment should say why the PTY-exit reasoning still holds over a live process.

### 4. Nit — `docs/harness-status-detection.md:548-551`

The "Posts are queued, not fired in parallel" bullet still justifies itself with "A `Stop` that overtook the `SessionStart` carrying the session id would be discarded by the Core as belonging to a session it has never heard of — #230 again with a new cause." After this PR a `Stop` in that position is no longer discarded. The queue is still right for other reasons (prompt text ordering, permission events); the example just needs rewording now that bullet 3 immediately below it says the opposite.

---

### Acceptance criteria

1. **A Stop whose `session_id` does not match the stored id still finishes (or nacks) the Core task the PTY belongs to** — **partially met.** Met and well tested when the tracked subagent set is empty (`harness-hook-pipeline.test.ts:301`, `harness-status-detection.test.ts:164` proving the row *and* `session:finished` on the Core, `opencode-hooks-api.test.ts:286` proving `{ ok: true, status: "finished" }` over HTTP). Not met when stale entries survive from a pre-resume process — finding 1.
2. **Matching-session Stop behaviour is unchanged** — **met**, and provably so: the diff removes exactly one line (`git diff --numstat` → `88 1`), the deletion being the old early return, so everything below the branch is byte-identical. Pinned anyway by the `matching-session Stop behaviour is unchanged` block.

### Everything else checked

Base is `fix/operator-pins-sessions-drop-cli` (not `main`, not `beta/0.4.5`); still a draft; no forbidden file touched (`TerminalPane.tsx`, `task-status-sync.ts`, `active-group.ts`, `pty-manager.ts`, `core-session-backstop.ts`, `panel-link/router.ts`, `terminal-store.tsx`, `__root.tsx`, `ProjectBar.tsx`, `projects.$id.tsx` — all clean); the single commit passes commitlint with no body line over 132 characters (longest is 120), and the `Conventions` job agrees; the #385 race window and the #387 `ready` settle are built on rather than rewritten, `noteTaskFinished` is called on the foreign finish so the one-second heal stays consistent, and the #387 relaunch reset (`disconnected → ready`) is safe because `ready` is excluded from the settle.

Requesting changes on finding 1; findings 2-4 are documentation and test honesty that should travel with it.

qcentic-dev:
## Review round 1 addressed — `f73d107`

Thank you for this. Finding 1 is right and it is the more interesting bug of the two: the hold I added as a safeguard could only ever fire on state that nothing was able to clear, so the one branch meant to be conservative was the one that reproduced #390. Findings 2 and 3 are the same mistake seen from the other side — I described a guarantee the code could not deliver and borrowed a precedent that does not transfer. All four are fixed in one commit.

---

### 1. BLOCKING — stale subagent entries wedge the foreign settle for up to two hours — **fixed**

`settleForeignTurnEnd` now opens with:

```ts
clearSubagentActivity(taskId);
clearTaskFinished(taskId);
```

on the reasoning already in this file, which your review quoted back at me: a turn end under an unrecognised id means a new harness process, so the old session's subagents died with it — the same sentence the `setSessionId` wrapper writes at `:196-198`. `clearTaskFinished` goes with it by the `sessionProcessExited` precedent you pointed at.

One nuance worth stating, since it is not quite the `sessionProcessExited` shape: the clear **drops** a mark, it does not **suppress** one. When the settle proceeds, `noteTaskFinished` stamps a fresh finish immediately after the write, so the one-second heal window behaves exactly as it does after any other hook-driven finish — the consistency you called out approvingly in the first round is preserved. The clear is what the paths that do *not* settle are left with. That is written into the function's doc comment so the next reader does not have to re-derive it.

Your walkthrough is now a test, with the two-hour wedge as its subject:

- `"settles a fanned-out turn whose resume lost its SessionStart (review finding 1)"` — two `SubagentStart`s tracked, then a foreign `Stop`; the task reaches `finished` and the result carries `status: "finished"` rather than the bare `{ outcome: "ok", event: "Stop" }` that was the wedge's ack.
- `"drops the tracked subagents on entry, so the next turn is not held"` — the clear has to outlive the event, or a stale entry holds the *next* turn's owned `Stop` for the whole TTL.
- Core-side, `"finishes a fanned-out turn whose resume lost its SessionStart (issue 390)"` proves the row **and** `session:finished`, which is the half of the symptom a pipeline-level test cannot see.

### 2. The "subagent hold still applies" guarantee cannot engage — **hold dropped, and said plainly**

You are right that it is inert in both shapes, and right that the honest resolution once finding 1 is fixed is to remove it rather than keep a branch that only fires on stale state. It is gone. What replaced it is the admission, in all three places that previously claimed the guarantee:

- the function's doc comment now says there is no hold, that after the clear the set is empty by construction, that a resumed session's subagent events are dropped as foreign before `noteSubagentStart` is reached, that opencode posts no subagent lifecycle events at all so `hasActiveSubagents` is structurally false for every opencode Session — and that **the status check is the only real guard**;
- `docs/harness-status-detection.md` carries the same three bullets, including "a leaked child *can* finish a working parent's `running` card. That is the residual, stated rather than papered over";
- the PR body's summary and a new **Residual risk** section say it too, so the claim is not left standing anywhere.

On the test: agreed, `harness-hook-pipeline.test.ts:356` modelled a case that cannot occur. It is replaced by `"does not hold on subagents posted under the foreign id either"`, which posts the `SubagentStart` under `FOREIGN`, asserts it comes back `{ outcome: "foreign-session", event: "SubagentStart" }` — dropped exactly like every other foreign claim — and then shows the turn end settling straight through.

### 3. Settling `needs-input` — **restricted to `running`**

Taken, and the concrete OpenCode walkthrough is what settled it: the parent is alive and blocked on the prompt, and `tasks.ts:97-100` clearing the pending question on the transition means the overlay goes with the card. "Self-corrects" is not good enough when the correction requires the operator to notice a question that is no longer on screen.

`settleForeignTurnEnd` now gates on `task.status !== "running"` alone, and the comment says why the PTY-exit pair looked like a precedent and is not one — there the process is dead, so an open question is moot; here it is alive and may be blocked on exactly that question. Covered at all three levels: the pipeline test asserts the exact result object and write sequence, and the Core and Panel tests assert no `session:finished` and `{ ok: true, event: "Stop" }` respectively.

### 4. Nit — the "posts are queued" bullet — **reworded**

You are right that it argued for the queue using the one consequence this PR removed. It now names orderings that still break — a `chat.message` overtaking `SessionStart`, leaving a prompt the Core cannot attribute; a `permission.replied` overtaking its `permission.asked`, leaving the card on `needs-input` for a question already answered — and notes parenthetically that a `Stop` in that position is no longer among the casualties, which is a narrowed blast radius rather than a reason to stop queueing. The child-session bullet below it was rewritten in the same pass, since it also leaned on the hold.

---

### Still true after the change

- **Exactly one line is deleted** from `harness-hook-pipeline.ts` against the base — still the old early return — so the matching-session path and the #385 / #387 blocks remain byte-identical. `subagent-activity.ts` is untouched.
- Base is still `fix/operator-pins-sessions-drop-cli`; still a draft; no forbidden file touched; nothing merged, no force push.
- Both commits pass commitlint locally over the full range (`--from origin/fix/operator-pins-sessions-drop-cli --to HEAD`, 0 problems), and `Conventions` agrees on the pushed head.
- `pnpm lint` 0 errors, `pnpm typecheck` clean, `pnpm test` all 6 stages green. Hook suites: shared 27, core 19, panel 53.

https://claude.ai/code/session_01HFY4g9AwbTLVX94heQZKp5

mrfqcentic (approved):
## Re-review of #453 against issue #390 — round 2, approved

Reviewed at head `f73d107`, base `fix/operator-pins-sessions-drop-cli` (now `31444a9`), in a fresh worktree. Every enforcing line re-read; the shared, Core and Panel hook suites re-run in a scratch copy; a fresh `/code-review` pass over the full PR diff.

All four round-1 findings are fixed, and the blocking one is fixed in the way the round-1 walkthrough asked for rather than papered over.

---

### 1. BLOCKING — stale subagents wedging the settle for ~2h — **fixed**

`settleForeignTurnEnd` opens with `clearSubagentActivity` + `clearTaskFinished` (`harness-hook-pipeline.ts:475-476`), before the status check, so the hold can no longer be reached with a stale set. Verified end to end:

- `harness-hook-pipeline.test.ts:446` — two tracked `SubagentStart`s, then a foreign `Stop`, reaches `finished` and answers `status: "finished"` rather than the bare ack that *was* the wedge;
- `harness-hook-pipeline.test.ts:468` — the clear outlives the event, so the next turn's owned `Stop` is not held;
- `harness-status-detection.test.ts:179` — the Core row **and** `session:finished`, which is the half a pipeline test cannot see.

The `clearTaskFinished` nuance in the doc comment is accurate, and I checked it rather than took it: on the settle path `noteTaskFinished` re-stamps at `:484`, so a matching in-turn `SubagentStart` landing inside the one-second window still heals to `running`, and a matching `Stop` that follows still finishes. Confirmed with a throwaway probe against this head, not by reading alone.

### 2. The hold that could not engage — **dropped, and said so in all three places**

Gone from the code; the doc comment (`:444-451`), `docs/harness-status-detection.md:194-203` and the PR body all now say there is no hold, why it could not have engaged, and that **the status check is the only real guard** — including the residual that a leaked child can finish a working parent's `running` card. The test that modelled an impossible case is replaced by `harness-hook-pipeline.test.ts:481`, which posts `SubagentStart` under `FOREIGN`, pins `{ outcome: "foreign-session", event: "SubagentStart" }`, and then shows the settle proceeding.

### 3. `needs-input` settled over a live process — **restricted to `running`**

`:477` gates on `task.status !== "running"` alone, with the non-transferring PTY-exit precedent written out. Pinned at all three levels, with the exact ack: `{ outcome: "ok", event: "Stop" }` and writes `["running", "needs-input"]` (`harness-hook-pipeline.test.ts:324`), no `session:finished` on the Core (`harness-status-detection.test.ts:196`), `{ ok: true, event: "Stop" }` over HTTP (`opencode-hooks-api.test.ts:264`).

### 4. Nit — the "posts are queued" bullet — **reworded**

`docs/harness-status-detection.md:566-575` now names orderings that still break and notes parenthetically that a `Stop` in that position is no longer among the casualties.

---

### Acceptance criteria

1. **A foreign `Stop` still finishes (or nacks) the task** — **met**, including the case that was not met in round 1. `harness-hook-pipeline.test.ts:301/446/468/481`, `harness-status-detection.test.ts:164/179`, `opencode-hooks-api.test.ts:315`.
2. **Matching-session `Stop` behaviour is unchanged** — **met**, still provably: `git diff --numstat` against the base is `120 1`, and the single deleted line is still the old early return, so everything below the branch is byte-identical. The three hunks land in the helper, the foreign branch and the new function — nowhere near the #385 race window, the #387 `ready` settle or the `disconnected → ready` relaunch reset, all of which are built on rather than rewritten.

### Everything else checked

Diff `9e392ea..f73d107` touches only the pipeline, its tests, the Core and Panel hook tests and the docs file — five files, no forbidden one among them. Both commits pass commitlint (headers 64 / 71; longest body lines 120 / 118, under the 132 cap), and `Conventions` agrees. Merges cleanly onto `origin/fix/operator-pins-sessions-drop-cli` at `31444a9` (`git merge-tree`, no conflict). Still a draft, still on the stacked base, not `main`. CI on `f73d107`: 18 pass, 6 skipping. Local suites in a scratch copy: shared 554/554 (pipeline 27), Core hook detection 19/19, Panel hook APIs 45/45 — the one red I saw was `operator-auth > throttles a run of wrong passwords`, a timing test unrelated to this change that passes on a quiet machine.

---

### Non-blocking notes

None of these change the verdict; the first two are the shape of the trade this PR makes deliberately, and I want them on the record rather than fixed under time pressure.

**a. `harness-hook-pipeline.ts:475-476` — the two clears run *before* the status guard, so a path that declines to settle still mutates a live session's bookkeeping.** `clearSubagentActivity` also calls `disarmDeferredFinish` (`subagent-activity.ts:257-260`). On the `needs-input` path — the branch that exists precisely because the process may be alive and blocked — a foreign turn end drops that live session's tracked subagents and kills its drain backstop, so the parent's next owned `Stop` can finish while background subagents are still working. I confirmed the mutation with a probe. It is inert in both shapes this PR names (a resumed process's entries are dead anyway; opencode posts no subagent events at all), which is why it is a note and not a request: moving both clears inside the settle branch would keep everything load-bearing and drop the side effect, if you think it is worth a commit.

**b. There is a third shape that reaches this branch, and the dropped hold was the only thing that would have caught it.** Hooks live in the project's `.claude/settings.local.json` and read `$AC_HOOK_TASK_ID` from the environment (`harness-hooks.ts:112-124`), so any `claude` started inside the session's PTY — a nested `claude -p`, the headless title helper this repo already guards against at `core-title-generator.ts:44-47` and `hooks.controller.ts:88-92` — posts to the same task id under its own session id. Over a live owned fan-out that now settles `finished` and dings mid-turn. Two things keep this off the blocking list: such a process fires `SessionStart`/`UserPromptSubmit` first, and those are *capture* events, so it takes the id-adoption path and never reaches the foreign branch unless those POSTs are lost; and through that same capture path it can already hijack the card today, before this PR. That hazard is bigger than #390 and pre-dates it — worth its own issue rather than a change here.

**c. `docs/harness-status-detection.md:203` and `:583-584` (and the PR body's Residual risk) — "the parent's next `session.status: busy` puts the card back" overstates the bound.** The plugin posts `UserPromptSubmit` on the busy *transition* (`harness-hooks-opencode.ts:183`), so a parent that was already busy when the child leaked emits nothing new until its next turn; the card reads `finished` for the rest of the current one, and the ding has already fired. "Corrects it on the parent's next turn" would be the honest version.

One more, for the same reason as round 1's finding 4: `harness-hooks-opencode.ts:86-88` still carries the claim the docs bullet dropped — "a Stop that overtook the UserPromptSubmit carrying the session id would be discarded by the Core as belonging to a session it has never heard of". That file is not in this diff, so it is not this PR's job, but the sentence is now stale in two ways (before any capture there is no stored id, so such a `Stop` was never foreign to begin with).

**Approving.** The blocking wedge is gone, the three follow-ons travelled with it, both acceptance criteria are met, and the residual this fix knowingly keeps is stated in the code, the docs and the PR body rather than dressed up as a guarantee.

Issue #391, opened 2026-09-02
F5: lost Stop plus idle TUI redraws starve the 15-minute backstop


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `F5`
- Severity: broken
- First-fix owner: core
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Codex without `/hooks` review, or a dropped Stop POST. Harness is idle but still painting a clock/spinner. Card stays running indefinitely.

## Where

`pty-manager.ts` `noteActivity` every 5s of output (catalog: 718–724); `core-session-backstop.ts` skips unless quiet for 15 minutes (149–150). Same file as H3 — do not parallel H3.

## What to do

Idle TUI redraws are not operator activity. Either stop counting them as activity, or have a finish-class backstop that does not require 15 quiet minutes of zero bytes.

## Acceptance

1. An idle harness that never sends Stop settles within a bounded window that an operator can actually wait for.
2. A live turn that is still printing real output is not killed by the backstop.

## Related

Codex Stop itself is #290 / #277 (H2). This ticket is the backstop when Stop never arrives.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #455, merged to `beta/0.4.5` via #481. The PR carried `Refs` rather than a closing keyword, so closing by hand.

Issue #389, opened 2026-09-02
F3: fleet/grid can miss a finish that lands during an in-flight refresh


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `F3`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md; rendering investigation B3 (`use-fleet.ts:133`)

## What the operator sees

Have Fleet open. A session finishes while tasksList fan-out is running. Row stays running until the next 15s poll or another event. Pin dots flicker the same way.

## Where

`use-fleet.ts`: `if (!bridge || refreshing.current) return;` — no dirty flag, no trailing re-run.

## What to do

Set a dirty flag when an event arrives during refresh and re-run once. Do not drop the finish.

## Acceptance

1. A `session:finished` / `task:updated(finished)` that lands mid-refresh is visible after that refresh without waiting for the 15s poll.
2. Pin activity dots for that project match the settled row.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #456, merged to `beta/0.4.5` via #481. Acceptance 2 (pin activity dots match the settled row) is satisfied by #457, which landed on the same train.

PR #454, opened 2026-09-02
fix(panel): a new tab still sees buffered session:finished

### LINKED ISSUE #388: F2: a new Panel tab never gets buffered session:finished (no toast)

- Hunt ID: `F2`
- Severity: broken
- First-fix owner: panel

## What the operator sees

Session finishes while the Panel service is up but this tab is not watching. Open the Panel: grid may already say finished via tasksList; no toast/ding.

## Where

`panel-link/router.ts` replayFor(`lastEventId<=0`) returns [] and jumps to head (catalog: 280–283). New `watchCore` starts at 0.

## What to do

Replay finish-class events to a tab that missed them, or toast from the tasksList snapshot when a newly visible row is already finished. Depends on Core actually emitting `session:finished` (F1/F4/F5).

## Acceptance

1. Opening a new Panel tab after a session has finished still produces the finish toast/ding once, or an equivalent finish notice.
2. A tab that was watching does not double-toast.

qcentic-dev:
Closes #388 (F2).

## What the operator saw

A Session finishes while the Panel service is up but this tab is not watching.
Open the Panel: the grid already says `finished` — read off `tasksList` — and
nothing else happens. No toast, no ding, no notification row.

## Why

`panel-link/router.ts` `replayFor` treats "I have seen nothing" as "start at
head": `lastEventId <= 0` returns an empty tail and jumps the tab's cursor to
the head of the log. A new `watchCore` starts at 0, so a tab opened after the
finish was buffered is handed nothing.

That is the right answer for almost every kind — a tab's views load current
state through queries, and replaying the log at them only makes them refetch
what they are already fetching. It is wrong for exactly one kind: the finish is
both a *state* the grid re-reads and an *event* the Panel announces once, and
nothing re-derives the announcement.

## The fix

**Server.** `replayFor` hands a tab that has seen nothing the finish-class tail
— `session:finished`, bounded to the newest 8 — and still jumps its cursor to
head: these are a notice it missed, not a place in the log it is behind at.
Every other subscribe path is untouched.

**Browser.** The panel-link client marks events delivered between a `subscribe`
and its `eventsReplayed` marker as `replay`, so a listener can tell "this just
happened" from "you were handed this". A **replayed** finish is announced only
if nothing in this browser has announced it already:

- the tab's own in-memory `(coreId, sessionId, eventId)` dedup answers for **the
  tab that was watching** — it never double-toasts;
- a durable per-browser record of announced keys answers for **a second tab
  opened afterwards** — it does not re-announce what the first tab already said.

A **live** finish is announced unconditionally in every open tab, exactly as
before: two tabs open when a Session ends is one event reaching two surfaces.
The durable record is kept only for a finish the Core numbered — an SSE
(Panel-local) finish has no `eventId`, so a durable key for it would silence a
real second finish forever.

`docs/adr/0008` gains a correction note: its "no `session:finished` event is
ever missed" was true of the *service* and never of a *tab*.

## Tests

- `panel-link/router.test.ts` — a brand-new tab gets the finish and nothing
  else; the replay is bounded to the newest 8; a tab with a cursor past the
  finish is not replayed it.
- `panel-link/live-read-path.test.ts` — end to end from a seeded Core log
  through the service: a tab that opens after the Session finished receives the
  finish, and not the rest of what it missed.
- `panel-link-client.test.ts` — the replay window opens on `subscribe` and
  closes on `eventsReplayed`, including across a reconnect.
- `use-session-finish-notifications.integration.test.ts` — one notice for a
  replayed finish; one for a live finish repeated by a replay (the watching tab
  does not double-toast); one across two tabs; and a second finish of the same
  Session still announces under its own `eventId`.

Full `packages/panel` suite: 147 files, 1606 tests, green. Typecheck and lint
clean (one pre-existing unrelated warning).

## Depends on

The Core actually emitting `session:finished` — #387 / #390. This lands against
the current feature tip and is tested with a seeded log.

## Boundaries

Untouched: `use-fleet.ts` (#389), `packages/core` `pty-manager.ts` /
`core-session-backstop.ts` (#391), `packages/shared`
`harness-hook-pipeline.ts` (#390), `active-group.ts` (#384),
`terminal-store.tsx`, `__root.tsx`, `ProjectBar.tsx`, `TerminalPane.tsx`,
`projects.$id.tsx`.

https://claude.ai/code/session_01Ryt5bYuaGvXv6nNgT2cNdx


qcentic-dev:
## Acceptance checklist — issue #388 (F2)

- [x] **1. Opening a new Panel tab after a session has finished still produces the finish toast/ding once, or an equivalent finish notice.**
  `replayFor` no longer answers a `lastEventId <= 0` subscribe with an empty
  tail: it hands the tab the finish-class tail (`session:finished`, newest 8)
  and still jumps its cursor to head. The tab announces it exactly once — the
  toast, the ding, the OS notification and the bell row all fire off the same
  dispatch. Proven by `router.test.ts` ("still hands a brand-new tab the
  finishes it was not open for"), end to end by `live-read-path.test.ts`
  ("gives a tab opened after a Session finished the finish it was never there
  for" — a real Core, a seeded log, the service, a tab that opens afterwards),
  and at the hook by "still announces a finish to a tab opened after it
  happened" (one `mcToastCustom`, one ding, one stored row).

- [x] **2. A tab that was watching does not double-toast.**
  Two independent guards, and the replay never reaches the first one anyway: a
  watching tab re-subscribes from its own cursor, which is past the finish, so
  the finish-class replay is for fresh tabs only (`router.test.ts`, "does not
  replay a finish to a tab that already has a cursor past it"). If it is
  delivered regardless, the client drops any event at or below its cursor, and
  the hook's in-memory `(coreId, sessionId, eventId)` dedup drops it again.
  Proven by "does not double-toast the tab that was watching when the replay
  repeats it" — live finish then the same finish as replay, one toast, one ding,
  one row.

- [x] **Not in the acceptance list, but the noise it would have created:** a
  *second* tab opened later must not re-announce what the first tab already
  said. A durable per-browser record of announced keys suppresses the
  announcement for a replayed finish only. A live finish still announces in
  every open tab, and a genuine second finish of the same Session announces
  under its own `eventId`. Covered by three further tests in
  `use-session-finish-notifications.integration.test.ts`.

Nothing is unchecked.

### Dependency, stated plainly

The notice can only fire for a finish that is in the log, which needs the Core
to emit `session:finished` (#387 / #390). This PR is implemented against the
current tip of `fix/operator-pins-sessions-drop-cli` and tested with a seeded
log rather than waiting on that emitter.

### Verification run locally

- `packages/panel` suite: **147 files, 1606 tests, all passing**
- `pnpm -C packages/panel typecheck`: clean
- `eslint` on every changed file: clean (one pre-existing, unrelated `prev`
  unused-arg warning in `use-session-finish-notifications.tsx`)
- `commitlint --from e68c354 --to HEAD`: **0 problems, 0 warnings**

### CI status of the PR head

Head `c251135`, run
[33672759141](https://github.com/actana/control/actions/runs/33672759141):
**all 18 non-skipped checks pass, 0 failing, 0 pending.**

Conventions, Lint, Typecheck, Unit Tests, Secret Scan, Dependency Audit,
E2E — Panel service seam, E2E — installer (ubuntu, x64), E2E — installer
(debian, x64), Smoke — Core release tarball (linux-x64 and linux-arm64),
Core image / Build + smoke (amd64), Panel image / Build + smoke (amd64), both
"Resolve registries", Resolve PR image mode, Train rules — pass. Promotion gate
passes here because the base is the feature branch, not `main`. The train image
jobs skip, as they do on every draft.

This pull request stays a **draft** and is not to be merged: it is one ticket of
the pin/session/CLI/drop wave, stacked on `fix/operator-pins-sessions-drop-cli`.

mrfqcentic (changes requested):
## Review of #454 against issue #388 (F2) — requesting changes

Reviewed at head `c251135`, base `fix/operator-pins-sessions-drop-cli` (`e68c354`), in a
detached review worktree. Verdicts below come from reading the enforcing lines, not the PR
description; the full `packages/panel` suite was re-run locally in a scratch copy.

**Both acceptance criteria are met.** The change requests below are defects in the delivered
notice that sit outside the ACs — the ACs are a floor, and this is the review of what ships.

---

### Acceptance criteria

**AC1 — "Opening a new Panel tab after a session has finished still produces the finish
toast/ding once, or an equivalent finish notice." — MET.**

The bug was `replayFor` treating "seen nothing" as "start at head". It no longer does:

- `packages/panel/src/server/panel-link/router.ts:314-317` — the `lastEventId <= 0` branch now
  returns `state.buffer.filter(e => FINISH_EVENT_KINDS.has(e.kind)).slice(-FINISH_REPLAY_LIMIT)`
  instead of `[]`, while still jumping the cursor to `state.head`. That is the line the issue
  named, and it is the line that changed.
- `panel-link-client.ts:645` / `:764` / `:772` — the tab tags anything arriving between its
  `subscribe` and the `eventsReplayed` marker as `replay`.
- `use-session-finish-notifications.tsx:487` and `:260-266` — a replayed finish is announced
  unless this browser already announced it.

Delivered exactly once: the router emits each buffered finish once per subscribe, and
`panel-link-client.ts:762` (`event.eventId <= watch.cursor` → drop) plus `markSeen` at `:261`
are two further layers that would each catch a repeat.

**AC2 — "A tab that was watching does not double-toast." — MET.**

Two independent enforcers, either sufficient alone:

1. A watching tab's cursor is already past the finish, so `replayFor` never replays it
   (`router.ts:319`) — `router.test.ts` *"does not replay a finish to a tab that already has a
   cursor past it"*.
2. Even if replayed, `markSeen(dedupKey)` returns false — integration test *"does not
   double-toast the tab that was watching when the replay repeats it"*.

**What bounds the replay.** `FINISH_REPLAY_LIMIT = 8` (`router.ts:119`) over a per-Core buffer
capped at `DEFAULT_EVENT_BUFFER_SIZE = 2048` (`router.ts:96`), cleared on every link bind
(`router.ts:574`). Covered by *"bounds the finish replay to the recent tail rather than the
whole log"*.

**How the once-only guard is keyed.** `dedupKey` = `${coreId}::${sessionId}::${eventId ?? "sse"}`
(`use-session-finish-notifications.tsx:40-45`), in two stores: tab-scoped `seenFinishKeys` Set
(LRU 500, `:52-67`) and browser-scoped `mc:sessionFinishAnnounced` in `localStorage` (cap 500,
`session-notification-store.ts:44-47`, `:183-200`), the latter written only when
`eventId !== null`. Withholding the durable key for an SSE finish is right, and `:262-265` says
why.

**Reload counts as fresh, not watching.** The per-Core cursor lives only in the in-memory
`watching` map (`panel-link-client.ts:117`); nothing persists it (the `sessionStorage` at
`:869-872` is the drive client id). So a reload re-subscribes at 0 and is handed the finish
tail, and does not re-toast only because the `localStorage` record survives the reload. That
makes the durable record load-bearing rather than belt-and-braces.

---

### Requested changes

**1. `packages/panel/src/lib/use-session-finish-notifications.tsx:274` — a replayed finish is
stamped `finishedAt: Date.now()`, so an old finish reads as "just now" and can evict a newer
one.**

`CoreLinkEvent` carries `ts` — "Wall-clock ms when the event was appended"
(`packages/sdk/src/core-link-frames.ts:129-130`) — but `normalizeSessionFinishedEvent`
(`:83-135`) never reads it, `NormalizedFinish` has no timestamp field, and the notification is
built with `Date.now()`.

This was latent before, because replay only ever spanned a short reconnect gap in a tab that
had usually seen the event live. This PR makes it reachable by design: the whole point is to
hand a fresh tab finishes it was never there for, which may be arbitrarily old within the
link's lifetime. It is not cosmetic — `notificationTimestamp` returns `finishedAt`
(`session-notification-store.ts:57-59`) and `sortNotifications` both orders on it and slices to
`MAX_NOTIFICATIONS = 200` (`:111-115`), so the stale entry sorts above genuinely newer rows and,
at the cap, evicts one.

Concretely: a Session finishes at 02:00, the operator opens the Panel at 09:00, and the bell row
says it finished seconds ago, at the top of the list. Carrying `event.ts` into `NormalizedFinish`
and using it for `finishedAt` when present fixes it.

**2. `packages/panel/src/server/panel-link/router.ts:314` — the finish tail is bounded by count
but not by age.**

The filter runs over the whole buffer, which spans the entire life of the current core-link
connection, with no check on `event.ts`. The only browser-side gate is `hasAnnouncedFinish`,
which is empty for any profile that has not announced these before, and `use-fleet` watches
every registered Core. So a first-ever visit — a second device, a new profile, a private
window, or after clearing site data — subscribes at `lastEventId = 0` to each Core and is handed
up to `8 × N` finishes, each firing a toast, a `playNotificationDing`, and an OS notification.
Three Cores and an overnight backlog is up to 24 stacked toasts and 24 overlapping dings for
Sessions that ended hours ago.

An age bound on the tail (minutes, not the connection's lifetime) alongside the count bound
would keep this to "you missed a finish", which is what the constant's own comment says it is
for. This is the same root as (1): both are the replay reaching further back than the notice is
meaningful for.

---

### Non-blocking

**3. `use-session-finish-notifications.tsx:260` — the durable check is gated on the `replay`
flag rather than on "this tab has seen nothing", which is broader than #388 needs.**

Any subscribe answer is tagged `replay`, including a reconnect catch-up — the new
`panel-link-client.test.ts` *"reopens the replay window on the subscribe a reconnect sends"*
asserts exactly that. So: tabs A (foreground) and B (background) both open, A's socket blips
across the finish, B announces live and writes the durable key, A reconnects and is silent.
Before this PR A toasted. Consistent with the PR's stated policy, but the toast fires in the tab
the operator cannot see.

There is a second, narrower window with the same cause. `sendSubscribe` sets
`watch.replaying = true` at send time (`panel-link-client.ts:645`) and clears it a round trip
later. For a *reconnect* this is safe — `PanelLinkRouter.attach` (`router.ts:245-255`) always
builds a new `PanelLinkSession` with an empty `watching` set, so the service pushes nothing to
that socket until the subscribe arrives, and the subscribe handler is synchronous. But a
same-socket re-subscribe (`watch()` going refs 0→1 after a view remount) hits a server session
that still has the Core in `watching` — nothing ever removes it — so a genuinely live event
landing in that window is tagged `replay`.

Gating the durable check on "this tab's cursor was 0 at subscribe" rather than on the replay
flag would cover exactly the fresh-tab case #388 is about and close both of these.

**4. `session-notification-store.ts:203` — `clearAnnouncedFinishes` has no production call
site.** Only the tests call it; its docstring says "Used by tests and by a storage reset", and
there is no storage reset — neither `clearSessionFinishNotifications` nor the bell's clear-all
touches `mc:sessionFinishAnnounced`. Either wire it into the reset the docstring promises or
drop the second clause.

**5. Scope limit worth naming in the ADR.** The new correction says the service "does drain the
tail into its per-Core buffer", but `bind` clears `state.buffer` on every new Core-side
connection (`router.ts:574`) and the service re-subscribes from its own already-advanced cursor.
So a finish the service consumed before a Core redial is not re-buffered, and a tab opening
after that redial gets nothing. #388's own scenario (link up throughout) is covered; the
guarantee is one core-link connection wide, which is narrower than the ADR now reads.

**6. `CONTEXT.md:95` still reads "The Panel never misses an event."** — the sibling of the exact
sentence ADR 0008 just gained a correction for. Still true of the *service*, but a reader who
reaches `CONTEXT.md` first gets the premise this PR disproved for a *tab*. Worth the same
one-clause qualification.

**7. Cross-tab race in the durable guard.** Two fresh tabs opened at the same moment can both
read `hasAnnouncedFinish` before either `recordAnnouncedFinish` lands, and both announce.
Outcome is identical to the live two-tab case the design allows — noting it as a known property.

**8. FYI for #389, no change requested.** The replay now delivers ≤8 `session:finished` events
to a fresh tab, which also reach `use-fleet.ts:176` and `use-core-live-queries.ts:41`. Checked:
harmless — `run()` has an in-flight guard (`use-fleet.ts:131`) so the burst collapses to at most
one extra fetch, and react-query coalesces the invalidations.

---

### Verification I ran

- `packages/panel` full suite in a scratch copy: **147 files, 1606 tests, green** — matches the
  PR's claim exactly. The end-to-end `live-read-path.test.ts` case *"gives a tab opened after a
  Session finished the finish it was never there for"* passes on its own.
- Tests genuinely exercise the new paths: `window.localStorage.clear()` in `beforeEach`
  (integration test `:121`) and the fixture carries `eventId = 42`, so the cross-tab case is real
  rather than vacuous.
- ESLint on the five changed source files: 0 errors, 1 warning — confirmed pre-existing
  (`setNotifications((prev) => …)`, base line 252 → head line 279), as the PR states.
- CI on `c251135`: all 24 checks green, none failing.
- Commitlint: subject 58 ≤ 120, no body or footer line over 132; the `Conventions` job passes.
- Boundaries clean — none of `use-fleet.ts`, `pty-manager.ts`, `core-session-backstop.ts`,
  `harness-hook-pipeline.ts`, `active-group.ts`, `terminal-store.tsx`, `__root.tsx`,
  `ProjectBar.tsx`, `TerminalPane.tsx`, `projects.$id.tsx` is touched. Base is the feature
  branch, not `main` or `beta/0.4.5`. Still a draft.

---

### Verdict

The core fix is right, minimally scoped, and lands on exactly the line the issue named; both ACs
are met and each is proved at more than one layer. What is not yet right is the notice it
produces: a finish replayed hours later is stamped as having just happened and can push a newer
row off the list, and the tail it replays reaches as far back as the connection does. Items 1 and
2 are the same root and small to fix. Happy to re-review promptly.

qcentic-dev:
## Review addressed — `2dbdeb5`

Thanks — both blocking items were the same root you named, and the fix is one
commit on top of `c251135`. Nothing was force-pushed; `c251135` is untouched.

### CR1 — a replayed finish is stamped `Date.now()` — **fixed**

`NormalizedFinish` gains `finishedAt: number | null`, filled from `event.ts` in
the fleet branch of `normalizeSessionFinishedEvent` and `null` in the SSE branch
(the Panel's own stream is live by construction, so "now" is the truth there).
`dispatchNormalizedFinish` builds the row with `finish.finishedAt ?? Date.now()`.
A Session that ended at 02:00 and is replayed at 09:00 now carries 02:00 into
`sortNotifications`, so it sorts where it belongs and cannot evict a newer row at
the 200 cap.

One guard beyond what you asked for, because carrying a foreign clock invites the
mirror of the same bug: `finishedAtFrom` rejects a `ts` that is not a positive
finite number **and** one ahead of this browser's clock, falling back to arrival
time. A Core running fast would otherwise pin its row to the top of the bell for
as long as the skew lasts.

Tests: *"dates a replayed finish by when it finished, not by when the tab
opened"*, *"sorts a replayed old finish below a newer one instead of on top of
it"*, *"stamps a Core clock that is not usable with the time the finish
arrived"*. The two exhaustive `toEqual` shape assertions in
`use-session-finish-notifications.test.ts` were updated to carry the new field.

### CR2 — the tail is bounded by count but not by age — **fixed**

`FINISH_REPLAY_MAX_AGE_MS = 30 * 60_000` — thirty minutes, stated in the
constant's own doc comment as "the operator stepped away and came back", which
is the gap #388 is about. Beyond it the finish is not news and the grid saying
`finished` is the right way to learn about it. `replayFor` now filters on
`FINISH_EVENT_KINDS.has(kind) && !isOlderThan(e, Date.now() - MAX_AGE)` before
the `slice(-8)`, so both axes a flood could arrive on are bounded.

Measured on the Core's own `ts`, deliberately **not** on when the service
buffered the event: `bind` empties the buffer on every new Core-side connection,
so a receipt clock would call a redial's overnight backlog brand new — exactly
the case you described. An event whose `ts` is unusable is kept, because failing
toward the notice is the point of the fix; the count bound still caps it.

Tests: *"leaves a finish older than the replay window to the grid, not to a
toast"* and *"does not drop a finish whose Core sent no usable timestamp"*. The
`event()` fixture helper now defaults `ts` to `Date.now()` rather than
`eventId * 10` — a fixture dated 1970 would have quietly opted out of the path
under test — with an explicit `ts` parameter for the tests that care.

### 3 — gate on the cursor, not on the replay flag — **taken**

`Watch.replaying: boolean` is replaced by `Watch.subscribedFrom: number | null`,
set to the cursor the outstanding `subscribe` was sent from and cleared by
`eventsReplayed`. The frame flag is renamed `replay` → **`coldReplay`**, true
only when that cursor was zero.

Both windows you named close: a reconnect catch-up asks from a cursor above
zero, so tab A announces the finish it blipped across even though tab B already
wrote the durable key; and a same-socket re-subscribe (`refs` 0→1 on a view
remount) also has a non-zero cursor, so a genuinely live event landing in that
window is not tagged. Test: *"does not call a reconnect catch-up a cold replay —
this tab lived through it"*, plus *"announces a catch-up after a blip, which is a
gap this tab was watching"* at the hook.

### 4 — `clearAnnouncedFinishes` — **docstring fixed, not wired**

Deliberately not wired: clearing the bell empties a list the operator has *read*,
while that record is what was already said *out loud*. Joining them would make
"clear notifications" the gesture that lets an old finish announce itself again
in the next tab. The docstring now says exactly that instead of promising a
storage reset.

### 5 — ADR scope — **taken.** The correction now names the guarantee as one
core-link connection wide, says why (`bind` clears the buffer, the service
re-subscribes from its own advanced cursor), and notes the age bound.

### 6 — `CONTEXT.md` — **taken.** "The Panel never misses an event" is now "The
Panel *service* never misses an event", with the clause about a tab's own
in-memory cursor starting at zero and the finish being the one thing replayed to
it on purpose.

### 7 — cross-tab race — **acknowledged in the code.** The comment at the
durable check now says two tabs opened in the same instant can both read "not
announced" and both say it, and that this lands on the two-tabs-open outcome the
line above already allows, so it is left unlocked rather than papered over.

### 8 — noted, no change, and thanks for checking `use-fleet.ts` — untouched.

---

**Verification.** `packages/panel`: **147 files, 1612 tests, green** (up from
1606 — six new cases). Typecheck clean. ESLint on the changed files: 0 errors,
2 warnings, both pre-existing and in code this PR does not touch. `commitlint
--from e68c354 --to HEAD`: **0 problems, 0 warnings** over both commits, run
before the push.

**CI on `2dbdeb5`: all 18 non-skipped checks pass, 0 failing** — including
`Conventions`, `Unit Tests`, `Lint`, `Typecheck`, both installer E2Es and the
Panel service seam E2E.

Still a draft, still based on `fix/operator-pins-sessions-drop-cli`, nothing
merged. Ready for re-review.

mrfqcentic (approved):
## Re-review of #454 against issue #388 (F2), round 2 — approving

Reviewed at head `2dbdeb5`, base `fix/operator-pins-sessions-drop-cli` (now `31444a9`), in a fresh
detached review worktree. Verdicts come from the enforcing lines, not from the reply; the full
`packages/panel` suite was re-run in a scratch copy.

**Both round 1 blockers are fixed at the lines that enforce them. Both non-blocking notes are
taken. Both acceptance criteria still hold, and the new tests pin the two things that changed.**

---

### CR1 — a replayed finish was stamped `Date.now()` — **fixed**

- `use-session-finish-notifications.tsx:41` — `NormalizedFinish` gains `finishedAt: number | null`.
- `:145` — the fleet branch fills it from `finishedAtFrom(event.ts)`; `:110` leaves it `null` on the
  SSE branch, which is right: the Panel's own stream is live by construction.
- `:172-175` — `finishedAtFrom` rejects a `ts` that is not a positive finite number, **and** one
  ahead of this browser's clock. That second guard was not asked for and is the right instinct: a
  Core running fast would otherwise pin its row to the top of the bell for as long as the skew lasts.
- `:310` — the row is built with `finish.finishedAt ?? Date.now()`.

`notificationTimestamp` (`session-notification-store.ts:57-59`) returns `finishedAt` and
`sortNotifications` (`:110-114`) both orders and slices to `MAX_NOTIFICATIONS = 200` on it, so the
02:00-finish-opened-at-09:00 case now sorts where it belongs and cannot evict a newer row.

Pinned by three new cases: *"dates a replayed finish by when it finished, not by when the tab
opened"* (asserts the exact `finishedAt`), *"sorts a replayed old finish below a newer one instead
of on top of it"* (asserts the order, which is the failure the first one only implies), and *"stamps
a Core clock that is not usable with the time the finish arrived"* (covers both `ts = 0` and a `ts`
in the future). The two exhaustive `toEqual` shape assertions in the unit test carry the new field.

### CR2 — the tail was bounded by count but not by age — **fixed**

- `router.ts:142` — `FINISH_REPLAY_MAX_AGE_MS = 30 * 60_000`, thirty minutes, with the reasoning in
  the constant's own doc comment (`:130-141`).
- `router.ts:348-354` — the filter is
  `FINISH_EVENT_KINDS.has(e.kind) && !isOlderThan(e, Date.now() - MAX_AGE)`, applied **before**
  `slice(-FINISH_REPLAY_LIMIT)`. Both axes bound the same tail, so a first-ever visit gets at most
  `min(8, finishes in the last 30 minutes)` per Core — not eight of arbitrary age.
- `router.ts:149-151` — `isOlderThan` treats an unusable `ts` as *not* old, so an event with no
  usable timestamp is kept. Failing toward the notice is the right default for this fix.

Measured on the Core's own `ts` rather than on receipt time, and the comment says why: `bind`
empties `state.buffer` on every Core-side connection (`router.ts:611`), so a receipt clock would
call a redial's overnight backlog brand new — the exact case round 1 described.

Pinned by *"leaves a finish older than the replay window to the grid, not to a toast"* (6h dropped,
5m kept) and *"does not drop a finish whose Core sent no usable timestamp"*. The `event()` fixture
helper now defaults `ts` to `Date.now()` rather than `eventId * 10` — without that the new path
would have been silently opted out of by every existing case, and catching that is what makes these
tests real.

### 3 — gate on the cursor, not on the replay flag — **taken**

`Watch.replaying: boolean` → `Watch.subscribedFrom: number | null` (`panel-link-client.ts:138`), set
to the cursor the outstanding `subscribe` was sent from (`:658`), cleared by `eventsReplayed`
(`:785`); the frame flag is `coldReplay`, true only when that cursor was zero (`:777`). Both windows
close: a reconnect catch-up asks from a cursor above zero, and a same-socket re-subscribe after a
view remount does too. `:785` clearing to `null` rather than `0` matters — `=== 0` would otherwise
be true for every live event after a subscribe, and it is not.

Verified by *"does not call a reconnect catch-up a cold replay — this tab lived through it"* at the
client and *"announces a catch-up after a blip, which is a gap this tab was watching"* at the hook —
the second is the one that matters, because it is the operator-visible half. The rename is complete;
no `msg.replay` or `watch.replaying` reader is left anywhere in `packages/panel/src`.

### 4 — `clearAnnouncedFinishes` — **docstring fixed, not wired, and the reasoning is right**

`session-notification-store.ts:202-210` now says the record is what was said *out loud* and the bell
is a list the operator has *read*, so joining them would make "clear notifications" the gesture that
lets an old finish announce itself again in the next tab. That is a better answer than the wiring I
offered as the alternative.

### 5, 6, 7 — ADR scope, `CONTEXT.md`, cross-tab race — **all taken.** The ADR correction now names
the guarantee as one core-link connection wide and says why; `CONTEXT.md:95` distinguishes the
*service* from a *tab*; the race is named in the comment at `use-session-finish-notifications.tsx:288-290`
and left unlocked, which is the right call for an outcome identical to the two-tabs-open case the
design already allows.

---

### Is thirty minutes the right bound?

Yes — and the choice, not just the presence of a bound, is what I checked.

The thing being bounded is an **interrupt**: a toast, a ding, and an OS notification. Its value
decays fast. Inside the window it means "you missed this while you stepped away", which is #388's
scenario exactly; outside it, the operator is not coming back to a moment, they are coming back to a
board, and `tasksList` saying `finished` is the correct channel. Thirty minutes covers the coffee
break and the meeting that ran over, and stops well short of the overnight backlog. I would have
accepted anything from roughly fifteen minutes to two hours; thirty is squarely inside that, it is
one named constant with its reasoning attached, and it is trivially re-tunable if the fleet turns
out to want longer.

One cost worth naming, not blocking: because the bell row is only written when a finish is
*announced*, a finish older than the window leaves no persistent row either — the grid is the whole
record. That is not a regression (before this PR such a tab got nothing at all), and the bell was
never a complete history. If "what finished while I was out" is later wanted as a list, that is a
surface built on `tasksList`, not a wider replay window.

---

### Acceptance criteria — still met

**AC1 — a new tab opened after a finish still gets the notice once.** `router.ts:348-353` returns
the finish tail instead of `[]`; the client tags it `coldReplay`; the hook announces it unless this
browser already did. End to end in `live-read-path.test.ts` — *"gives a tab opened after a Session
finished the finish it was never there for"* — against a real Core behind mTLS, with a witness tab
proving the service had the events before the fresh tab asked. Exactly once: the router emits each
buffered finish once per subscribe, `panel-link-client.ts:775` drops anything at or below the cursor,
and `markSeen` is a third layer.

**AC2 — a tab that was watching does not double-toast.** Still two independent enforcers: its cursor
is past the finish so `replayFor` never replays it (`router.ts:355-358`), and `markSeen` would refuse
it anyway. Both still covered.

### Observations for later — none blocking, none requested here

1. **`router.ts:349` compares a Core's `ts` to the service's `Date.now()`.** A Core more than thirty
   minutes *behind* has every finish filtered out and #388 is quietly un-fixed for that Core; one far
   *ahead* makes the bound inert. The comment argues correctly against a receipt clock; a per-event
   arrival time carried across a rebind would be the third option. Needs a materially broken clock to
   bite, so it is a follow-up, not a gate.
2. **`use-session-finish-notifications.tsx:310` uses the Core's `ts` on the live path too**, and
   `finishedAtFrom` clamps only the future direction. A lagging Core files a just-finished Session in
   the bell at its skewed time, below rows that really are older. Same root as (1). Restricting the
   event `ts` to `delivery === "cold-replay"` would keep the fix and leave the common path on one
   clock — worth considering if fleet clock skew ever shows up in practice.
3. **`FINISH_REPLAY_LIMIT` is per Core with no fleet-wide cap.** The age bound cut round 1's worst
   case from an overnight backlog to the last thirty minutes, which is the trade I asked for. The
   residual — N Cores each finishing several Sessions inside the same half hour, seen by a profile
   with an empty `mc:sessionFinishAnnounced` — is a burst of genuinely recent finishes, the same ones
   a watching tab would have toasted live. Acceptable as shipped.
4. **`coldReplay` is a property of the send window, not of the event.** A live event can still be
   tagged if the tab's cursor is zero at re-subscribe, which requires it to have seen *nothing* from
   that Core — narrower than the two windows round 1 named, and it lands on an outcome the design
   already allows. Noting it as a known property.

### Verification I ran

- Full `packages/panel` suite in a scratch copy: **147 files, 1612 tests, green** — matches the
  reply's claim exactly, six cases up from round 1's 1606. (A first pass at default concurrency hit
  four vitest worker-startup timeouts; at `--maxWorkers=3` everything passes, so those were local
  resource contention, not the code.)
- Every round 1 case still green by name, and the five new ones green by name.
- `tsc --noEmit` on `packages/panel` after `generate:routes`: **clean**.
- ESLint on the five changed source files: **0 errors, 1 warning** — the pre-existing unused `prev`
  at `use-session-finish-notifications.tsx:315`, the same one round 1 confirmed against the base.
- `commitlint --from e68c354 --to 2dbdeb5`: **exit 0**. Subjects 58 and 81 characters; no body or
  footer line over 132.
- Diff `c251135..2dbdeb5` touches 11 files, all Panel plus `CONTEXT.md` and the ADR. The full PR is
  13 files, all owned by this PR. None of `use-fleet.ts`, `pty-manager.ts`, `core-session-backstop.ts`,
  `harness-hook-pipeline.ts`, `active-group.ts`, `terminal-store.tsx`, `__root.tsx`, `ProjectBar.tsx`,
  `TerminalPane.tsx` or `projects.$id.tsx` is touched. `c251135` is unchanged — no force-push.
- `git merge-tree` onto `origin/fix/operator-pins-sessions-drop-cli` at `31444a9`: **clean, no
  conflicts** (GitHub agrees: `MERGEABLE` / `CLEAN`).
- Still a **draft**, still based on `fix/operator-pins-sessions-drop-cli`, not on `main` or a train.
- **CI on `2dbdeb5`: 18 checks pass, 6 skipped, 0 failing** — run `33675890180`, `head_sha`
  confirmed `2dbdeb5`. Includes `Conventions`, `Unit Tests`, `Lint`, `Typecheck`, `Secret Scan`, both
  installer E2Es and the Panel service seam E2E.

---

### Verdict

**Approve.** Both blockers are fixed where it counts, and each fix went one step past the request in
the direction the request implied — the future-clock guard on CR1, and the fixture `ts` default on
CR2 that stops the new tests from being vacuous. The two non-blocking notes I expected to be argued
were: one taken with a cleaner mechanism than I proposed, one declined with a reason better than
mine. Thirty minutes is the right shape for "you missed a finish". The observations above are
follow-up material, not conditions.

Issue #383, opened 2026-09-02
S7: after a Core switch the picker can list the previous Core’s projects


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `S7`
- Severity: likely
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md; rendering investigation use-fleet stale snapshots

## What the operator sees

Jump from a Core A pin to Core B, open the picker immediately, click a row. Navigates an A project id with B’s coreId.

## Where

`use-fleet.ts` keeps old snapshots while loading (catalog: 217–224). `GroupSwitcher` uses `useProjects()` counts.

## What to do

Do not render the previous Core's project list under the new Core's id. Clear or gate the picker on coreId change.

## Acceptance

1. Immediately after switching Core, picker rows are that Core's projects or a loading state — never the previous Core's ids.
2. A picker click never pairs Core B's `coreId` with Core A's project id.

Issue #377, opened 2026-09-02
S5: rail activity dots never move for Core-owned pins


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `S5`
- Severity: likely
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md; rendering investigation W5 / taskCounts

## What the operator sees

Start a session on a Core project. Sidebar tiles never show running / needs-input dots.

## Where

`projectRowFromSnapshot` always zeros `taskCounts`. `useRemotePinnedProjects` and `useCoreProjectRows` both use that mapper (`packages/panel/src/shared/projects.ts` catalog: 100–107).

## What to do

Derive Core pin dots from Core-owned task snapshots (or live events), not from the zeroed Panel mapper. Do not invent a second status authority — read the same facts the grid uses.

## Acceptance

1. A running Core session lights the matching pin dot.
2. A finished Core session clears it without a full reload.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #457, merged to `beta/0.4.5` via #481. The PR carried `Refs` rather than a closing keyword, so closing by hand.

PR #456, opened 2026-09-02
fix(panel): do not drop a finish that lands mid fleet refresh

qcentic-dev:
Refs #389 (F3). **Scope narrowed after review:** this PR is now the coalescing loop and nothing else.

## What was wrong

`useFleetTasks` guarded its fan-out with `if (!bridge || refreshing.current) return;`. Any `session:finished` or
`task:updated(finished)` that arrived while a `tasksList` fan-out was in flight hit that early return and was
dropped. The reads already in flight were launched *before* the finish, so their answers cannot carry it, and
nothing re-read them — the row stayed `running` until the 15s poll or until some later event happened along.

## The fix — `packages/panel/src/lib/use-fleet.ts`

- The refresh owns a coalescing loop instead of a bare guard. An event landing during a pass sets a `pending`
  flag rather than returning; the loop reads that flag when the pass settles and runs one trailing pass per
  burst, however many events the burst contained.
- The flag is cleared *before* each read, not after, so an event landing inside the trailing pass survives into
  the next one — no finish is ever dropped, including one that lands inside a trailing pass.
- A failed pass breaks the loop rather than spinning on the error, and **leaves `pending` as the burst left it**
  so the backlog is picked up by the next event or the poll (review finding 6).
- `setFleet` keeps the previous object when the merge settles on the same answer, so `fleet.rows` no longer
  churns identity on every event and the memos and effects keyed on it upstream stop tearing down for nothing
  (review finding 5).

## Acceptance

**Acceptance 1** — *a finish landing mid-refresh is visible after that refresh without waiting for the 15s poll*
— is met here and is what this PR is.

**Acceptance 2** — *pin activity dots for that project match the settled row* — is **not claimed by this PR.**
An earlier revision of #456 derived the dots here; review found that is #377's fix and that **PR #457 already
carries it**, rewriting the same `projectRowFromSnapshot` count block (`git merge-tree` conflicts on
`packages/panel/src/shared/projects.ts`). By orchestrator decision on merge order, **#456 keeps only the
coalescing loop and #457 wires the counts in `use-fleet.ts` after #456 merges.** That half is fully reverted
here — `shared/projects.ts` is byte-identical to the base branch and `useRemotePinnedProjects` is back to its
own `useCores` returning `pinned` unjoined.

Acceptance 2 is then **satisfied by construction**: a Core-owned pin currently has no count source at all, so
once #457 derives the dots from Core task snapshots on top of this loop, they settle through the very fan-out
this loop stops dropping — the mid-refresh finish reaches the dot for the same reason it now reaches the row.

The **unreachable-Core case** the review raised (an offline Core's pins falling through to zeroed counts and
reading as "finished" rather than "not counted") goes with the dot: it is a property of the derivation, not of
the loop, so **it belongs to #457**, which has `fleet.offlineCores` available to key off. Nothing in this PR
can produce it — no pin here has counts to zero.

## Tests — `packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx`

The mocked bridge parks every `listTasks` call and the test decides when each answers, which is what lets an
event land strictly inside a pass. Four cases:

1. a finish landing mid-refresh is visible once that refresh settles (3 reads exactly)
2. a burst of six events inside one pass produces exactly one trailing read and nothing behind it
3. **a finish landing inside the trailing pass itself is not dropped** — emits during read 3, asserts a fourth
   read and that both Sessions settle `finished` (review finding 3)
4. the loop settles after the trailing pass instead of re-reading forever

No test advances a timer. Both polls are 15s and each case finishes in milliseconds, so every read observed is
event-driven, and the exact read counts are the proof the 15s poll is not what fixed the row.

**Three mutations of `run()` now fail, each on all four cases or on the case written for it:**

| Mutation | Result |
| --- | --- |
| the original `if (refreshing.current) return;` guard | 4 of 4 fail |
| clear `pending` *after* the read instead of before | 4 of 4 fail |
| `do…while` → settle-then-at-most-one-extra-read | case 3 fails |

The third is the one review finding 3 called out as passing the previous suite.

`pnpm test` in `packages/panel`: 148 files, 1597 tests, green. Repo `pnpm typecheck` clean; `pnpm lint` 0 errors
(the one warning in `use-fleet.ts` is the pre-existing unused `Harness` import, untouched here). commitlint run
locally over `e68c354..HEAD` the way CI runs it: 0 problems.

## Scope

Touches `packages/panel/src/lib/use-fleet.ts` and the new test file **only**. `packages/panel/src/shared/projects.ts`
is back to base contents verbatim. The #383 picker Core-snapshot work in the same file is a later wave and is
**not** taken here. Base is `fix/operator-pins-sessions-drop-cli`, not `main` and not `beta/0.4.5`. Draft on
purpose. The revert was landed as a new commit; nothing pushed has been force-pushed.

https://claude.ai/code/session_01NasX2s8dYACnTAyhrLWhgv

### DIFF
--- a/packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx
+++ b/packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx
@@ -171,7 +171,7 @@ describe("useFleetTasks — a finish that lands during an in-flight refresh", ()
       "task:updated",
       "task:updated",
       "session:finished",
-      "pty:exited",
+      "pty:exit",
       "task:updated",
       "session:finished",
     ]) {
@@ -186,6 +186,53 @@ describe("useFleetTasks — a finish that lands during an in-flight refresh", ()
 
     // One trailing pass for the whole burst, and nothing behind it.
     expect(h.listTasksCalls).toBe(3);
+    expect(h.inFlight).toHaveLength(0);
+  });
+
+  it("does not drop a finish that lands inside the trailing pass itself", async () => {
+    // The loop is a `do…while`, not a single trailing pass, and this is the
+    // only case that can tell those apart. Rewriting `run` to "settle, then at
+    // most one extra read" passes every other test in this file while quietly
+    // re-narrowing #389 to the second event — so this case is what stops that
+    // refactor at CI rather than in an operator's Fleet view.
+    h.tasks = [task("task-1", "running"), task("task-2", "running")];
+    const { result } = renderHook(() => useFleetTasks());
+
+    await waitForRead(1);
+    await settleOldestRead();
+    await waitFor(() => expect(result.current.fleet.rows).toHaveLength(2));
+
+    // Read 2 opens, and the first Session finishes inside it.
+    await emit("task:updated");
+    await waitForRead(2);
+    h.tasks = [task("task-1", "finished"), task("task-2", "running")];
+    await emit("session:finished");
+
+    // Read 2 answers with the pre-finish snapshot, which arms the trailing
+    // pass — read 3.
+    await settleOldestRead();
+    await waitFor(() => expect(h.listTasksCalls).toBe(3));
+
+    // The second Session finishes while *that trailing pass* is in flight.
+    // A single-trailing-pass implementation has nowhere left to put this.
+    h.tasks = [task("task-1", "finished"), task("task-2", "finished")];
+    await emit("session:finished");
+
+    // Read 3 answers with a snapshot that predates the second finish...
+    await settleOldestRead();
+    // ...so a fourth read has to follow it.
+    await waitFor(() => expect(h.listTasksCalls).toBe(4));
+    await settleOldestRead();
+
+    await waitFor(() =>
+      expect(result.current.fleet.rows.map((row) => [row.taskId, row.status]).sort()).toEqual([
+        ["task-1", "finished"],
+        ["task-2", "finished"],
+      ]),
+    );
+    // Four reads and no more: mount, the event, and one trailing pass for each
+    // of the two bursts — not one per event, and not one left running.
+    expect(h.listTasksCalls).toBe(4);
     expect(h.inFlight).toHaveLength(0);
   });
 
--- a/packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx
+++ b/packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx
@@ -1,0 +1,302 @@
+// @vitest-environment jsdom
+//
+// The Fleet fan-out and the finish that lands while it is in flight (#389).
+//
+// The hook used to return early whenever a refresh was already running, so a
+// `session:finished` arriving mid-fan-out was dropped: the answers already in
+// flight predate the finish, nothing re-read them, and the row stayed running
+// until the 15s poll. These tests hold a fan-out open on purpose, land the
+// finish inside that window, and then count the reads.
+//
+// Nothing here advances a timer. `FLEET_POLL_MS` and `CORES_POLL_MS` are both
+// 15s and each test finishes in milliseconds, so every read a test observes is
+// an event-driven one — which is exactly the claim under test.
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { act, renderHook, waitFor } from "@testing-library/react";
+import type { CoreLinkTaskSnapshot } from "@actana/sdk/core-link-frames";
+import { mergeFleetTasks } from "~/shared/fleet-merge";
+
+const CORE_ID = "core-a";
+const PROJECT_ID = "project-1";
+
+const h = vi.hoisted(() => ({
+  /** Every `onEvent` subscriber currently mounted. */
+  eventHandlers: new Set<(msg: { coreId: string; event: { kind: string } }) => void>(),
+  /** One entry per in-flight `listTasks`, oldest first. */
+  inFlight: [] as { settle: () => void }[],
+  listTasksCalls: 0,
+  listProjectsCalls: 0,
+  /** What the Core would answer *right now*. Mutated by the tests. */
+  tasks: [] as CoreLinkTaskSnapshot[],
+  pinned: true,
+}));
+
+function task(taskId: string, status: string): CoreLinkTaskSnapshot {
+  return {
+    taskId,
+    projectId: PROJECT_ID,
+    title: taskId,
+    titleManuallySet: false,
+    claudeSessionId: null,
+    agent: "claude-code",
+    status,
+    pinned: false,
+    archived: false,
+    icon: null,
+    updatedAt: 1_000,
+  };
+}
+
+vi.mock("~/lib/api", () => ({
+  api: {
+    listCores: async () => ({
+      cores: [
+        {
+          id: CORE_ID,
+          endpoint: "wss://core-a:4100",
+          label: "Warehouse VM",
+          lastEventId: 0,
+          createdAt: 0,
+          updatedAt: 0,
+          dial: { coreId: CORE_ID, state: "connected", lastSeenAt: 1_000 },
+        },
+      ],
+    }),
+    listProjectPresentation: async () => ({ presentation: [] }),
+  },
+}));
+
+// One bridge object for the whole suite, exactly as `getPanelBridge` memoizes
+// the real one: the hooks key their effects on its identity.
+const bridge = {
+  isConnected: () => true,
+  watchCore: () => () => {},
+  onEvent: (cb: (msg: { coreId: string; event: { kind: string } }) => void) => {
+    h.eventHandlers.add(cb);
+    return () => h.eventHandlers.delete(cb);
+  },
+  onDialStatus: () => () => {},
+  onConnectionChange: () => () => {},
+  listProjects: async () => {
+    h.listProjectsCalls++;
+    return [
+      {
+        projectId: PROJECT_ID,
+        name: "Control",
+        path: "/srv/control",
+        icon: "CO",
+        iconColor: "#888888",
+        pinned: h.pinned,
+        rememberHarnessSettings: false,
+        savedHarness: null,
+        savedSkipPermissions: false,
+        savedBareSession: false,
+        defaultGridView: false,
+        updatedAt: 1_000,
+      },
+    ];
+  },
+  // Held open until the test says so: the whole bug lives in this window.
+  listTasks: () => {
+    h.listTasksCalls++;
+    const snapshot = () => ({ tasks: [...h.tasks], archivedCount: 0 });
+    return new Promise<{ tasks: CoreLinkTaskSnapshot[]; archivedCount: number }>((resolve) => {
+      h.inFlight.push({ settle: () => resolve(snapshot()) });
+    });
+  },
+};
+
+vi.mock("~/lib/panel-bridge", () => ({ getPanelBridge: () => bridge }));
+
+const { useFleetTasks, useRemotePinnedProjects } = await import("~/lib/use-fleet");
+const { fleetProjectKey, taskCountsByFleetProject } = await import("~/shared/projects");
+
+/** Let the oldest held `listTasks` answer, with the Core's current rows. */
+async function settleOldestRead(): Promise<void> {
+  const call = h.inFlight.shift();
+  if (!call) throw new Error("no listTasks in flight");
+  await act(async () => {
+    call.settle();
+    // Two turns: the fan-out's `Promise.all`, then the state it sets.
+    await Promise.resolve();
+    await Promise.resolve();
+  });
+}
+
+/** A Core event, as the panel link delivers it. */
+async function emit(kind: string): Promise<void> {
+  await act(async () => {
+    for (const cb of [...h.eventHandlers]) cb({ coreId: CORE_ID, event: { kind } });
+    await Promise.resolve();
+  });
+}
+
+/** Wait until the fan-out the hook starts on mount is the one in flight. */
+async function waitForRead(n: number): Promise<void> {
+  await waitFor(() => expect(h.listTasksCalls).toBeGreaterThanOrEqual(n));
+}
+
+beforeEach(() => {
+  h.eventHandlers.clear();
+  h.inFlight = [];
+  h.listTasksCalls = 0;
+  h.listProjectsCalls = 0;
+  h.tasks = [task("task-1", "running")];
+  h.pinned = true;
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+describe("useFleetTasks — a finish that lands during an in-flight refresh", () => {
+  it("shows the finish after that refresh, without waiting for the 15s poll", async () => {
+    const { result } = renderHook(() => useFleetTasks());
+
+    // First fan-out: the Session is running.
+    await waitForRead(1);
+    await settleOldestRead();
+    await waitFor(() => expect(result.current.fleet.rows).toHaveLength(1));
+    expect(result.current.fleet.rows[0]?.status).toBe("running");
+
+    // A task event opens a second fan-out and we hold it open.
+    await emit("task:updated");
+    await waitForRead(2);
+
+    // The Session finishes *inside* that window. The read already in flight
+    // was launched before it, so its answer cannot carry the finish.
+    h.tasks = [task("task-1", "finished")];
+    await emit("session:finished");
+
+    // The stale answer lands first — this is the state the old code settled on.
+    await settleOldestRead();
+
+    // ...and the dropped event has to have re-armed the fan-out.
+    await waitFor(() => expect(h.listTasksCalls).toBe(3));
+    await settleOldestRead();
+
+    await waitFor(() => expect(result.current.fleet.rows[0]?.status).toBe("finished"));
+    // Three reads: mount, the event, and the trailing re-run. No poll tick.
+    expect(h.listTasksCalls).toBe(3);
+  });
+
+  it("re-runs exactly once for a burst of events landing in one refresh", async () => {
+    const { result } = renderHook(() => useFleetTasks());
+
+    await waitForRead(1);
+    await settleOldestRead();
+    await waitFor(() => expect(result.current.fleet.rows).toHaveLength(1));
+
+    await emit("task:updated");
+    await waitForRead(2);
+
+    // Six more events while that one read is in flight — a burst, not six
+    // refreshes. Coalescing them is the other half of the fix.
+    h.tasks = [task("task-1", "finished")];
+    for (const kind of [
+      "task:updated",
+      "task:updated",
+      "session:finished",
+      "pty:exited",
+      "task:updated",
+      "session:finished",
+    ]) {
+      await emit(kind);
+    }
+    expect(h.listTasksCalls).toBe(2);
+
+    await settleOldestRead();
+    await waitFor(() => expect(h.listTasksCalls).toBe(3));
+    await settleOldestRead();
+    await waitFor(() => expect(result.current.fleet.rows[0]?.status).toBe("finished"));
+
+    // One trailing pass for the whole burst, and nothing behind it.
+    expect(h.listTasksCalls).toBe(3);
+    expect(h.inFlight).toHaveLength(0);
+  });
+
+  it("settles after the trailing pass instead of re-reading forever", async () => {
+    const { result } = renderHook(() => useFleetTasks());
+
+    await waitForRead(1);
+    await settleOldestRead();
+    await waitFor(() => expect(result.current.fleet.rows).toHaveLength(1));
+
+    await emit("task:updated");
+    await waitForRead(2);
+    await emit("session:finished");
+    await settleOldestRead();
+
+    // The trailing pass runs and then stops. A coalescing loop that cleared
+    // its flag *after* the read instead of before would keep re-reading here.
+    await waitFor(() => expect(h.listTasksCalls).toBe(3));
+    await settleOldestRead();
+    expect(h.listTasksCalls).toBe(3);
+  });
+});
+
+describe("useRemotePinnedProjects — pin activity dots", () => {
+  it("counts a pinned project from the same settled fan-out the row came from", async () => {
+    const { result } = renderHook(() => useRemotePinnedProjects());
+
+    await waitForRead(1);
+    await settleOldestRead();
+    await waitFor(() => expect(result.current.projects).toHaveLength(1));
+    // Running work lights the dot: the counts are the fan-out's, not zeroes.
+    await waitFor(() => expect(result.current.projects[0]?.taskCounts.running).toBe(1));
+    expect(result.current.projects[0]?.taskCounts.activeNonDone).toBe(1);
+
+    // The same mid-refresh finish, seen from the rail.
+    await emit("task:updated");
+    await waitForRead(2);
+    h.tasks = [task("task-1", "finished")];
+    await emit("session:finished");
+    await settleOldestRead();
+    await waitFor(() => expect(h.listTasksCalls).toBe(3));
+    await settleOldestRead();
+
+    await waitFor(() => expect(result.current.projects[0]?.taskCounts.finished).toBe(1));
+    const counts = result.current.projects[0]!.taskCounts;
+    expect(counts.running).toBe(0);
+    expect(counts.activeNonDone).toBe(0);
+    expect(counts.total).toBe(1);
+  });
+});
+
+describe("taskCountsByFleetProject", () => {
+  it("tallies exactly the rows mergeFleetTasks settled on", () => {
+    const rows = mergeFleetTasks([
+      {
+        coreId: CORE_ID,
+        coreLabel: "Warehouse VM",
+        ok: true,
+        lastSeenAt: 1_000,
+        tasks: [
+          task("task-1", "finished"),
+          task("task-2", "running"),
+          task("task-3", "needs-input"),
+          { ...task("task-4", "running"), archived: true },
+        ],
+      },
+    ]).rows;
+
+    const counts = taskCountsByFleetProject(rows);
+    const forProject = counts.get(fleetProjectKey(CORE_ID, PROJECT_ID));
+
+    // The archived row is not in `rows`, so it is not in the counts either.
+    expect(forProject?.total).toBe(rows.length);
+    expect(forProject?.finished).toBe(1);
+    expect(forProject?.running).toBe(1);
+    expect(forProject?.["needs-input"]).toBe(1);
+    expect(forProject?.activeNonDone).toBe(2);
+  });
+
+  it("keys by Core so two Cores sharing a project id do not merge", () => {
+    const counts = taskCountsByFleetProject([
+      { coreId: "core-a", projectId: PROJECT_ID, status: "running" },
+      { coreId: "core-b", projectId: PROJECT_ID, status: "running" },
+    ]);
+    expect(counts.get(fleetProjectKey("core-a", PROJECT_ID))?.running).toBe(1);
+    expect(counts.get(fleetProjectKey("core-b", PROJECT_ID))?.running).toBe(1);
+  });
+});
--- a/packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx
+++ b/packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx
@@ -14,7 +14,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { act, renderHook, waitFor } from "@testing-library/react";
 import type { CoreLinkTaskSnapshot } from "@actana/sdk/core-link-frames";
-import { mergeFleetTasks } from "~/shared/fleet-merge";
 
 const CORE_ID = "core-a";
 const PROJECT_ID = "project-1";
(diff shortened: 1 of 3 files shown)


qcentic-dev:
## Acceptance checklist — issue #389 (F3)

- [x] **1. A `session:finished` / `task:updated(finished)` that lands mid-refresh is visible after that refresh
  without waiting for the 15s poll.**
  `useFleetTasks` no longer returns early on `refreshing.current`. An event landing during a fan-out sets a
  `pending` flag, and the refresh loop runs one trailing pass once the in-flight pass settles. Covered by
  *"shows the finish after that refresh, without waiting for the 15s poll"* — the test holds the fan-out open,
  fires the finish inside that window, lets the stale answer land, and asserts the row reads `finished` after
  exactly three `listTasks` reads (mount, event, trailing re-run). No timer is advanced anywhere in the file and
  `FLEET_POLL_MS` is 15s, so the poll demonstrably is not what moved the row.

- [x] **2. Pin activity dots for that project match the settled row.**
  `useRemotePinnedProjects` now derives each row's `taskCounts` from the Fleet fan-out's own settled rows via
  `taskCountsByFleetProject`, keyed `coreId/projectId`, with the server's `decorate()` arithmetic. Dot and row
  are two readings of one snapshot. Covered by *"counts a pinned project from the same settled fan-out the row
  came from"* (running 1 / activeNonDone 1 → after the mid-refresh finish: finished 1, running 0,
  activeNonDone 0, total 1) and by the two pure `taskCountsByFleetProject` cases.
  **Note for the reviewer:** this criterion could not be met by the drop fix alone. `projectRowFromSnapshot`
  zeroes `taskCounts` — a project snapshot carries no tasks — so a Core-owned pin's dot was dark regardless of
  the row. Fixing the flicker required giving the dot a real count source. The trade is that a page mounting the
  rail now issues one `tasksList` per Core; flagged in the PR body.

Also delivered, beyond the two criteria:

- [x] **A burst of events during one refresh triggers exactly one trailing re-run.**
  Six events fired inside a single in-flight pass produce one additional read and nothing behind it
  (*"re-runs exactly once for a burst of events landing in one refresh"*). The flag is cleared before each read
  rather than after, so an event landing inside the trailing pass is still not dropped.

- [x] **Existing tests kept green.** `pnpm test` in `packages/panel`: 148 files / 1599 tests passing. Repo
  `pnpm typecheck` clean; `pnpm lint` 0 errors and no new warnings.

Nothing unchecked.

### Out of scope, deliberately

- Issue #383 (S7, the picker listing the previous Core after a switch) lives in the same `use-fleet.ts` and is a
  later wave — the picker Core-snapshot work is **not** taken here.
- No file owned by #388/PR 454, #391/PR 455, #390/PR 453 or #384/PR 450 is touched, and neither are
  `terminal-store.tsx`, `__root.tsx`, `ProjectBar.tsx`, `TerminalPane.tsx` or `projects.$id.tsx`.

https://claude.ai/code/session_01NasX2s8dYACnTAyhrLWhgv

qcentic-dev:
### CI status of the PR head (`dc14c9f`)

All checks settled: **18 pass, 6 skipping, 0 failing, 0 pending**. `Promotion gate` passes here because the base
is a ticket branch, not `main`.

Run: https://github.com/actana/control/actions/runs/33674950331

| Check | Result |
| --- | --- |
| Conventions | pass |
| Typecheck | pass |
| Lint | pass |
| Unit Tests | pass |
| Secret Scan | pass |
| Dependency Audit | pass |
| E2E — installer (debian, x64) | pass |
| E2E — installer (ubuntu, x64) | pass |
| Smoke — Core release tarball (linux-x64 / linux-arm64) | pass |
| Panel image / Core image — Resolve registries, Build + smoke (amd64) | pass |
| Resolve PR image mode, Train rules, Promotion gate | pass |
| Core image (train), Panel image (train), Train versions, Resolve train tags, both Publish manifest jobs | skipping |

Head `dc14c9f` adds a comment-only `docs(panel)` commit on top of the two commits the checklist above describes;
the checklist is unchanged by it. The PR is still a **draft** and has not been marked ready for review.

mrfqcentic (changes requested):
Reviewed against #389 at `dc14c9f`, diffed on `origin/fix/operator-pins-sessions-drop-cli` in a clean worktree. CI is green on the head commit (18 checks pass), `pnpm typecheck` clean, `pnpm lint` 0 errors / 45 warnings (all pre-existing — the one in `use-fleet.ts:6` is the unused `Harness` import, untouched by this PR), and `pnpm test` in `packages/panel` reproduces your numbers exactly: 148 files, 1599 tests, green.

**The core fix is right, and I verified it rather than taking the description's word for it.** I reverted the coalescing loop to the old `if (!bridge || refreshing.current) return;` guard in a scratch copy: 4 of the 6 new tests fail. I then removed only the `pending.current = false` that precedes the read, leaving everything else: 3 fail. So both the dirty flag and the clear-before-read ordering are genuinely load-bearing and genuinely test-enforced. Acceptance 1 is met.

I'm requesting changes on three things, one of which is a scope collision you can't resolve alone.

---

### 1. Blocking — this PR takes #377's fix, and PR #457 already has it. `packages/panel/src/shared/projects.ts:54-107`

PR #457 ("fix(panel): Core pin dots follow Core task status", issue #377) branches from the same base and rewrites the same file. #377's text names the exact site: *"`projectRowFromSnapshot` always zeros `taskCounts`… Derive Core pin dots from Core-owned task snapshots."* That is what this PR's `projects.ts` half does.

They collide, not just textually:

```
$ git merge-tree --write-tree origin/fix/fleet-refresh-dirty-flag origin/fix/core-pin-activity-dots
CONFLICT (content): Merge conflict in packages/panel/src/shared/projects.ts
```

Both add `ProjectTaskCounts` (`:54`) and `emptyTaskCounts()` (`:62`) at the same insertion point, and both rewrite `projectRowFromSnapshot`'s `taskCounts` line and its JSDoc. Beyond the conflict, they are two different derivations of one dot: #457 threads a count block into `projectRowFromSnapshot` from a `tasksList` read the caller already has; this PR joins onto a second `useFleetTasks` after the fact.

My recommendation: drop the `projects.ts` + `useRemotePinnedProjects` half here, keep this PR to the coalescing loop (which is the whole of `use-fleet.ts:178-204` and stands on its own), and let #457 own the dot. Acceptance 2 then falls out for free — once dots derive from Core task snapshots at all, the fix in this PR is exactly what makes them settle correctly mid-refresh. That also removes findings 4 and 5 below. Whichever way it goes, this needs a human to pick a merge order; I'm not going to pick it in a review.

### 2. Blocking — an unreachable Core zeroes its pin dots, which reads as "finished". `packages/panel/src/lib/use-fleet.ts:479-484`

When a Core's `tasksList` times out or its dial drops, `fanOut` maps it to the `offline` result, `mergeFleetTasks` gives it an `offlineCores` entry and **no rows**, so `taskCountsByFleetProject` has no key for its projects and every pin falls through to `emptyTaskCounts()`.

The pin itself stays on the rail — the effect at `:456` re-runs only on `coreIds`/`refreshNonce`, not on dial state — so the tile remains and its running dot goes out. An agent is running on Core B, the link blinks for one request timeout, and the operator watches the dot die. Indistinguishable from the session having finished.

This is squarely acceptance 2: *"Pin activity dots for that project match the settled row."* For an offline Core the Fleet row settles on "unreachable + last-seen" and the rail settles on "nothing running". Those don't match — the rail is asserting a fact it doesn't have. Your own docstring at `packages/panel/src/shared/projects.ts:56-58` draws exactly the distinction the join then discards ("this project has no active work" vs "nobody has counted yet"). `fleet.offlineCores` is right there to key off.

### 3. Blocking — the test suite does not hold the "trailing pass" property the loop exists for. `packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx`

Third mutation, same scratch copy: I replaced the `do…while` with **at most one** trailing pass —

```js
if (!(await fanOut())) {
  // single trailing pass only — an event landing inside it is lost
} else if (pending.current) {
  pending.current = false;
  await fanOut();
}
```

— which drops any event landing inside the trailing pass. **All 6 tests pass.** So the loop is correct, and nothing guards it; a later refactor to the simpler-looking single-pass form would be waved through by CI while silently reintroducing a narrowed #389.

The suite already has the machinery. A fourth case: emit during read 2 (the trailing pass), settle it, assert a read 3 happens and the final row carries the second finish. Given the parked-`listTasks` harness that's a few lines, and it's the one property in the ticket's "do not drop the finish" that is currently unverified.

---

### Non-blocking

**4. `use-fleet.ts:408` undercounts the cost, and it's worth a second look before you accept it.** `useRemotePinnedProjects` is mounted **twice** simultaneously — `routes/__root.tsx:266` and `components/views/ProjectBar.tsx:91`, and `__root.tsx:651` renders `ProjectBar` from that same `Shell`. So it's two `useFleetTasks` engines on *every* route, three on Fleet (`FleetView.tsx:32`), not "a page showing both the rail and Fleet fans out twice". Each subscribes `onEvent` independently, so one `session:finished` on a 10-Core fleet costs 20 `tasksList` frames on every route and 30 on Fleet, where it previously cost 10 on Fleet and 0 elsewhere. The coalescing loop caps each engine at one trailing pass, so this doesn't compound — but it's a 2–3× multiplier on the fleet's busiest path. A shared provider would give the dot its truth at 1×. (If you take finding 1, this disappears.)

**5. `use-fleet.ts:597`-adjacent identity churn.** `projects` is memoized on `fleet.rows`, and `fanOut` calls `setFleet(mergeFleetTasks(...))` unconditionally, so `fleet.rows` is a new array on every settle. That flows to `__root.tsx:267` `railProjects` → the rail-chord effect at `__root.tsx:597`, which now tears down and re-subscribes on every task/session/pty event fleet-wide instead of only when the project list changes. To be fair to the change: `Shell` already re-rendered on a 15s cadence before this PR, because `useCores` does `setCores([...list].sort(coreOrder))` on every poll — so the *timer* re-render is not new. What's new is the churn on every event. Bailing out of `setFleet` when the merge is deep-equal fixes both.

**6. `use-fleet.ts:198-202` — `break` drops the backlog it just collected.** A failed pass jumps to the `finally`, which clears `pending.current`, so events coalesced during that pass are lost — the #389 failure mode, narrowed to the error path. Impact is small because `fanOut` can essentially only return `false` via the `!bridge` guard (per-Core failures are swallowed into `offline` by the inner `try/catch`, so `Promise.all` never rejects — which also makes the `catch`/`setError` path at `:174` near-dead and `useFleetTasks().error` effectively always `coresError`). Leaving `pending` set on the `break` instead of clearing it would let the next event or poll pick up the backlog, at no cost.

**7. `use-fleet.ts:411` — `refresh()` no longer refreshes what the dot shows.** `useFleetTasks()`'s own `refresh` is destructured away, so the imperative `refreshRemotePinned()` at `ProjectBar.tsx:1394` (fired right after a pin toggle, explicitly "so the strip updates immediately") re-reads the pins but not the counts. A freshly pinned Core project with a running session draws a dark dot for up to 15s. Forwarding the fleet refresh alongside the nonce bump covers it.

**8. Nit — `use-fleet-refresh.test.tsx:200` emits `"pty:exited"`.** The real kind is `pty:exit`. It still exercises the path (`TASK_EVENT_KINDS` is the prefix `/^(task:|session:|pty:)/`), so nothing is wrong, but a reader chasing the kind list will not find it.

---

### Scope checks — all clean

Base is `fix/operator-pins-sessions-drop-cli` (not `main`, not `beta/0.4.5`); still a draft; three commits, all commitlint-clean (Conventions job green, and no body or footer line exceeds the configured 132). Touches only `use-fleet.ts`, `projects.ts` and the new test. Confirmed **not** touched: `panel-link/router.ts` (either copy — #388/PR #454 owns the server one), `active-group.ts`, `terminal-store.tsx`, `__root.tsx`, `ProjectBar.tsx`, `TerminalPane.tsx`, `projects.$id.tsx`, and anything under `packages/core` or `packages/shared`. The #383 picker Core-snapshot work is **not** taken — `useCoreProjects` and the picker gating are untouched; the `use-fleet.ts` hunks land only in `useFleetTasks` and `useRemotePinnedProjects`.

**Verdict: the coalescing loop is ready and I'd take it as-is. The pin-dot half is the blocker — it belongs to #377/PR #457, and as written it makes an unreachable Core look idle.**

qcentic-dev:
Thanks — the mutation testing in that review is what a review should be, and finding 3 in particular found a
real hole in my suite. All three blockers and three of the five non-blocking items are addressed at `7e87fd8`.

### Blocking

**1 — scope collision with #377 / PR #457.** Resolved by orchestrator decision on merge order, not by me:
**#456 keeps only the coalescing loop; #457 owns the dot and wires the counts in `use-fleet.ts` after #456
merges.** Fully reverted in `0df7d21` — `packages/panel/src/shared/projects.ts` is byte-identical to
`fix/operator-pins-sessions-drop-cli` (`git diff e68c354 -- packages/panel/src/shared/projects.ts` is empty),
`useRemotePinnedProjects` is back to its own `useCores` returning `pinned` unjoined, and the pin-dot tests and
their scaffolding are gone. The PR now touches `use-fleet.ts` and the new test file only.

The PR body and the acceptance checklist no longer claim acceptance 2. They say what you said: **acceptance 2 is
satisfied by construction once #457 derives the dots from Core task snapshots on top of this loop** — a
Core-owned pin has no count source at all today, so once it has one, the mid-refresh finish reaches the dot for
exactly the reason it now reaches the row. Reverting by a new commit rather than by rewriting history; nothing
pushed has been force-pushed.

**2 — an unreachable Core zeroing its pin dots.** Disappears with the dot half, and I agree it is a property of
the derivation rather than of the loop, so **it belongs to #457** — noted in the PR body, with the pointer to
`fleet.offlineCores` as the thing to key off. Nothing in this PR can produce it: no pin here has counts to zero.

**3 — the suite did not hold the trailing-pass property.** You were right, and I reproduced your mutation before
fixing it. New fourth case, *"does not drop a finish that lands inside the trailing pass itself"*: it emits
during read 3 (the trailing pass), settles it, asserts a fourth read follows and that both Sessions settle
`finished`. Your `settle-then-at-most-one-extra-read` rewrite now fails on that case. Re-ran your other two
mutations against the narrowed suite as well:

| Mutation | Result |
| --- | --- |
| original `if (refreshing.current) return;` | 4 of 4 fail |
| clear `pending` after the read instead of before | 4 of 4 fail |
| `do…while` → at most one trailing pass | case 3 fails |

### Non-blocking

**6 — `break` dropping the backlog.** Taken. `pending` is no longer cleared in the `finally`; a failed pass
leaves it exactly as the burst left it, and the next event or the poll picks it up. Your reading of the error
path is right — `fanOut` can essentially only return `false` through the `!bridge` guard — so this is
belt-and-braces, but it costs nothing and it removes a real drop from the code as written.

**5 — `fleet.rows` identity churn.** Taken. `setFleet` now compares the merged result structurally against the
previous one and keeps the previous object when they agree, so an event that changes nothing no longer produces
a new array. The walk is a recursive own-key comparison over the JSON-shaped snapshots and is bounded by the row
count `mergeFleetTasks` just paid for, so it is the same order as the merge itself.

**8 — `pty:exited`.** Fixed to `pty:exit`.

**4 and 7 — not applicable any more.** Both were consequences of the `useRemotePinnedProjects` join: finding 4's
2–3× fan-out multiplier (and thank you for counting the actual mount sites — I had undercounted it) and finding
7's `refresh()` no longer refreshing what the dot shows. Both vanish with the revert. Worth carrying into #457's
review, though: whatever derivation lands there inherits both questions, and finding 4's shared-provider
suggestion is the answer to the 1× version of it.

`pnpm test` in `packages/panel`: 148 files / 1597 tests green. `pnpm typecheck` clean, `pnpm lint` 0 errors
(45 warnings, all pre-existing — including the `Harness` one you identified). commitlint clean over
`e68c354..HEAD`. Still a draft, base unchanged, nothing merged.

https://claude.ai/code/session_01NasX2s8dYACnTAyhrLWhgv

qcentic-dev:
## Acceptance checklist — issue #389 (F3), revised at `7e87fd8`

This supersedes the checklist I posted at `dc14c9f`, which claimed acceptance 2 for this PR. It no longer does.

- [x] **1. A `session:finished` / `task:updated(finished)` that lands mid-refresh is visible after that refresh
  without waiting for the 15s poll.** — **met by this PR.**
  `useFleetTasks` no longer returns early on `refreshing.current`. An event landing during a fan-out sets a
  `pending` flag, and the refresh loop runs one trailing pass per burst once the in-flight pass settles; the
  flag is cleared before each read, so a finish landing inside the trailing pass survives too. Held by four
  tests, and by three mutations of `run()` that each fail them.

- [ ] **2. Pin activity dots for that project match the settled row.** — **deliberately not claimed here;
  owned by issue #377 / PR #457 by orchestrator decision on merge order.**
  Review found the dot derivation in this PR is #377's fix and that PR #457 already carries it, rewriting the
  same `projectRowFromSnapshot` count block (`git merge-tree` conflicts on `packages/panel/src/shared/projects.ts`).
  That half is fully reverted here: `shared/projects.ts` is byte-identical to the base branch and
  `useRemotePinnedProjects` is back to its own `useCores`.
  **This criterion is satisfied by construction once #457 lands on top of this loop.** A Core-owned pin has no
  count source at all today, so once #457 derives the dots from Core task snapshots, they settle through the
  very fan-out this PR stops dropping — the mid-refresh finish reaches the dot for the same reason it now
  reaches the row. Nothing further is required of #456.
  The **unreachable-Core case** raised as review blocking 2 (an offline Core's pins zeroing and reading as
  "finished" rather than "not counted") goes with the dot for the same reason: it is a property of the
  derivation, not of the loop, so **it belongs to #457**, which has `fleet.offlineCores` to key off. Nothing in
  this PR can produce it — no pin here has counts to zero.

### Also delivered

- [x] **A burst of events during one refresh triggers exactly one trailing re-run.** Six events inside one pass
  produce one additional read and nothing behind it.
- [x] **A finish landing inside the trailing pass is not dropped either** (review blocking 3). New fourth case;
  it fails the `settle-then-at-most-one-extra-read` rewrite that passed the previous suite.
- [x] **A failed pass keeps its backlog** (review non-blocking 6).
- [x] **`fleet.rows` no longer churns identity when a read changes nothing** (review non-blocking 5).
- [x] **`pty:exited` → `pty:exit`** in the test (review non-blocking 8).
- [x] **Existing tests kept green.** `pnpm test` in `packages/panel`: 148 files / 1597 tests. Repo `pnpm typecheck`
  clean; `pnpm lint` 0 errors, 45 warnings all pre-existing.

Review non-blocking 4 (the fan-out multiplier) and 7 (`refresh()` not refreshing the dot) were both consequences
of the reverted join and no longer apply to this PR; both are worth carrying into #457's review.

### Scope

Touches `packages/panel/src/lib/use-fleet.ts` and `packages/panel/src/lib/__tests__/use-fleet-refresh.test.tsx`
only. No file owned by #388/PR 454, #391/PR 455, #390/PR 453 or #384/PR 450 is touched, and neither are
`terminal-store.tsx`, `__root.tsx`, `ProjectBar.tsx`, `TerminalPane.tsx` or `projects.$id.tsx`. #383's picker
Core-snapshot work is not taken. Base unchanged, still a draft, nothing merged, no force push.

https://claude.ai/code/session_01NasX2s8dYACnTAyhrLWhgv

qcentic-dev:
### CI status of the PR head (`7e87fd8`)

All checks settled: **18 pass, 6 skipping, 0 failing, 0 pending.**
Run: https://github.com/actana/control/actions/runs/33679129630

Conventions · Typecheck · Lint · Unit Tests · Secret Scan · Dependency Audit · E2E — Panel service seam ·
E2E — installer (debian, x64) · E2E — installer (ubuntu, x64) · Smoke — Core release tarball (linux-x64) ·
Smoke — Core release tarball (linux-arm64) · Core image / Resolve registries · Core image / Build + smoke (amd64) ·
Panel image / Resolve registries · Panel image / Build + smoke (amd64) · Resolve PR image mode · Train rules ·
Promotion gate — **all pass**.

Skipping (train-only jobs, expected on a ticket branch): Core image (train), Panel image (train), Train versions,
Resolve train tags, and both Publish-the-multi-arch-manifest jobs.

The review is still recorded as CHANGES_REQUESTED; re-review whenever suits. Still a draft, base unchanged
(`fix/operator-pins-sessions-drop-cli`), nothing merged, no force push.

mrfqcentic (approved):
Re-reviewed against #389 at `7e87fd8`, in a worktree recreated from scratch at that commit and diffed against
the merge base `e68c354`. **All three blockers are resolved, and I verified each one rather than reading the
reply.** Approving.

## Blocking 1 — the scope collision with #377 / PR #457: resolved, and I have the control experiment for it

`git diff e68c354 7e87fd8 -- packages/panel/src/shared/projects.ts` is empty, and GitHub's own file list for the
PR is two entries: `use-fleet.ts` (+74/-5) and the new test (+257/-0). `useRemotePinnedProjects` is back on its
own `useCores` and returns `pinned` unjoined; the only change to it is a docstring.

The merge-tree check, run both ways so the result means something:

```
$ git merge-tree --write-tree dc14c9f 55f9045     # the old head, as a control
CONFLICT (content): Merge conflict in packages/panel/src/shared/projects.ts

$ git merge-tree --write-tree 7e87fd8 55f9045     # PR #456 head vs PR #457 head
7a1865e6cc4e386cdf2fbbac8257857de66415d3          # clean
```

The command still reproduces the round-1 conflict at the commit I found it on, and is clean at the current head.
#457 touches `shared/projects.ts` + a new `core-task-counts.test.ts`; #456 touches `lib/use-fleet.ts` + a new
test. Disjoint. #457 does not yet touch `use-fleet.ts`, which matches the stated plan that it wires the counts
there after this merges.

## Blocking 2 — offline Core zeroing its pin dots: gone with the derivation

There is no count source in this PR to zero. `useFleetTasks` now has exactly one consumer in the tree
(`FleetView.tsx:32`) — I grepped it — so round-1 findings **4** (the 2–3× fan-out multiplier from
`useRemotePinnedProjects` mounting a second and third engine) and **7** (`refreshRemotePinned()` not refreshing
the counts) are both genuinely gone, not merely argued away. Carry all three into #457's review.

## Blocking 3 — the trailing-pass property: now held, and I re-ran all three mutations

Scratch copy of `7e87fd8`, one mutation at a time, against the four-case suite:

| Mutation of `run()` | Result |
| --- | --- |
| the original `if (!bridge \|\| refreshing.current) return;` guard | **4 of 4 fail** |
| `pending.current = false` moved to *after* the read | **4 of 4 fail** |
| `do…while` → settle, then at most one extra read | **1 fails** — and it is exactly the new case 3, on `AssertionError: expected 3 to be 4` |

Unmutated: 4 passed. The mutation that walked through round 1 now stops at the assertion written for it.

## The other things I said I'd check

**A failed pass keeps the backlog** (finding 6) — `pending.current` is no longer cleared in the `finally`. Code
reads correctly. See the note below on what it actually buys.

**The deep-equal bailout does not suppress a genuine change.** Two independent legs. First, the type: a
`FleetMergeResult` is `rows` (a `CoreLinkTaskSnapshot` — eleven scalars plus an optional two-field `lock` — with
`coreId`/`coreLabel`), `offlineCores` (three scalars), and a boolean. No `Date`, `Map`, `Set` or class instance
anywhere, so an own-key recursive walk is exact for this data; `sameSnapshot`'s docstring is honest about that
being a precondition rather than a general equality. Second, empirically: forcing `sameSnapshot` to `return true`
fails all four cases, so the tests do bind the comparison's accuracy — a bailout that suppressed real changes
could not pass them. `lastSeenAt: … ?? Date.now()` is only reached on the offline branch, so no volatile field
enters the compared value on the happy path.

## Acceptance 2, deferred to #457 — yes, I think this is approvable

Stating my reasoning since you asked me to weigh it rather than just note it.

Acceptance 2 *cannot* be met by the loop alone, and that is a fact about the code, not a scoping convenience:
`projectRowFromSnapshot` zeroes `taskCounts` because a project snapshot carries no tasks, so a Core-owned pin's
dot has no count source at all today. It is uniformly dark — it does not flicker mid-refresh, it is simply off.
So #456 neither meets acceptance 2 nor regresses it, and there is no operator-visible state that gets worse by
landing the loop first. The claim in the PR body that acceptance 2 then "falls out by construction" is the right
shape: once #457 gives the dot a real count source, it settles through the very fan-out this loop stops dropping.

Merge order between two branches is an orchestrator's call and not a reviewer's — I said so in round 1 and I am
not going to relitigate a decision that was made. What I will hold is the bookkeeping: **#389 must not be closed
by #456 alone.** Acceptance 2 is unmet until #457 lands, and #457's review has to carry acceptance 2, the
unreachable-Core case (`fleet.offlineCores` is the thing to key off), and round-1 findings 4 and 7. The revised
checklist above leaves criterion 2 unchecked and says exactly this, which is the traceability that makes the
deferral safe to approve on.

## Non-blocking — worth a follow-up, not worth another round

**1. `use-fleet-refresh.test.tsx:79` — the mock snapshots at *settle* time, so the "stale answer" the file is
built around is never actually stale.** `const snapshot = () => ({ tasks: [...h.tasks], … })` is invoked inside
`settle`, and every case mutates `h.tasks` to the finished state *before* calling `settleOldestRead()`. So the
read "launched before the finish" resolves *with* the finish. That makes the comments at `:141` ("its answer
cannot carry the finish") and `:145` ("The stale answer lands first — this is the state the old code settled
on") describe something the harness does not do, and it leaves the `listTasksCalls` counts as the only
load-bearing assertions: the row-status assertions at `:152`, `:185` and `:227` are already satisfied by the
read *before* the trailing pass. A regression that ran the trailing pass but discarded its answer would pass all
four cases. The fix is one line — capture at call time, `const snapshot = { tasks: [...h.tasks], archivedCount:
0 };` — and I ran the suite that way against unmutated `7e87fd8`: **4 passed**, so it is a safe strengthening,
not a rewrite. It also makes case 1 fail the old guard for its *stated* reason (row stuck `running`) rather than
only on a read count.

**2. `use-fleet-refresh.test.tsx:122` — no `cleanup()`, and this package does not get it for free.**
`vitest.config.ts` does not set `globals: true`, so RTL's auto-cleanup is never registered, and every other RTL
suite here registers it by hand (`project-picker-switch.test.tsx`, `cores-settings-rename.test.tsx`,
`first-run-gate.test.tsx`, `UpdateBanner.test.tsx`, …). This file is the only one that doesn't, so all four
`renderHook` mounts stay live with 15s `FLEET_POLL_MS`/`CORES_POLL_MS` intervals. If the file ever exceeds 15s
wall time, a leaked hook's poll calls the shared mock, incrementing `h.listTasksCalls` and pushing onto
`h.inFlight`, and the exact-count assertions (`toBe(3)`, `toBe(4)`, `toHaveLength(0)`) are what break. That is
not hypothetical on a loaded runner — on my own contended box tonight I watched individual tests in this package
take 26–75s. `afterEach(cleanup)` closes it.

**3. `use-fleet.ts:223` — the backlog comment promises more than the flag delivers.** Leaving `pending` set on a
failed pass is inert, not belt-and-braces: `pending.current` is read only by `while (pending.current)`, and
every entry into the loop assigns `pending.current = false` before its first read, so a `true` left behind is
always overwritten before it can be observed. Recovery is entirely "the next event or the poll calls `run()`
again", which would happen identically if the flag were cleared. I confirmed it: restoring `pending.current =
false` to the `finally` passes all four cases. Harmless as written, and I'd rather have it this way round — but
the comment invites a maintainer to treat the flag as a durable backlog that it isn't, and it's worth saying
that `break` is close to dead code anyway (per-Core `catch`es mean `fanOut` returns `false` essentially only via
the `!bridge` guard, which `run` has already checked).

**4. `use-fleet.ts:424` — nit.** "on top of the coalescing loop **below**" — `useFleetTasks` is ~290 lines
*above* `useRemotePinnedProjects`.

## Scope, hygiene and CI

- **Diff:** two files, both expected. `shared/projects.ts` byte-identical to base. Nothing under `packages/core`
  or `packages/shared`. `__root.tsx`, `ProjectBar.tsx`, `terminal-store.tsx`, `TerminalPane.tsx`,
  `projects.$id.tsx`, `active-group.ts` and both copies of `panel-link/router.ts` untouched. #383's picker
  Core-snapshot work still not taken.
- **commitlint:** ran it the way the Conventions job does — `--from e68c354 --to 7e87fd8` against a throwaway
  install of this repo's `commitlint.config.mjs`: **0 problems, 0 warnings** across all five commits. Longest
  body line is 82 (limit 132); longest header 75 (limit 120); `revert` is in the `type-enum`; every `Refs`/
  `Co-Authored-By` block has its leading blank. PR title lints clean too. Reverted forward, not force-pushed.
- **Merge:** `git merge-tree --write-tree origin/fix/operator-pins-sessions-drop-cli 7e87fd8` is clean against
  the base tip `2bdcb56`; GitHub reports `MERGEABLE` / `CLEAN`.
- **Draft and base:** still a draft, base still `fix/operator-pins-sessions-drop-cli`.
- **CI on `7e87fd8`:** 18 pass, 6 skipped, 0 failing, 0 pending. The six skips are the train-only jobs, expected
  on a ticket branch.
- **Local:** panel `tsc --noEmit` clean; `eslint` on both changed files is 0 errors / 1 warning, the pre-existing
  unused `Harness` import that is already on the base at `use-fleet.ts:6`. All 16 test files that touch
  `use-fleet` pass together: 188 tests. My full-package runs had 7–12 socket-backed `src/server/**` tests time
  out, a *different* set each run, at 26–75s each under load average 77 on 8 cores; none of those files
  reference `use-fleet`, and CI's Unit Tests job is green on this commit. Environmental, not this PR.

**Verdict: approved.** The coalescing loop is correct, the property round 1 found unguarded is now guarded by a
test that fails the mutation that walked through, the scope collision is genuinely gone and provably so, and the
acceptance-2 deferral is sound on the merits as well as by decision. The four non-blocking items are all in the
new test file or in comments — take them in a follow-up, or in #457's wake. Keep #389 open until #457 lands.

Issue #378, opened 2026-09-02
S6: group filter removes the pin ring while you are still on that project


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `S6`
- Severity: likely
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Open a project in group A, switch GroupSwitcher to B. Page stays on A; rail shows B with no selection ring.

## Where

`ProjectBar.tsx` scoped rail (catalog: 185–189); overlay only if `activeId` is in the visible list.

## What to do

Keep a selection affordance for the project you are actually on, even when the active group hides that pin. Do not silently navigate away.

## Acceptance

1. Switching group while viewing a project in another group does not pretend nothing is selected.
2. The URL still matches the open project.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #470, merged to `beta/0.4.5` via #481. The remaining All-leg case is filed separately as #474.

PR #457, opened 2026-09-02
fix(panel): Core pin dots follow Core task status

qcentic-dev:
Refs #377 (S5 — rail activity dots never move for Core-owned pins). Draft, based on `fix/operator-pins-sessions-drop-cli`.

## The bug

`projectRowFromSnapshot` hard-zeroed `taskCounts` for every Core-owned project row, and `useRemotePinnedProjects`
read each Core's *projects* and nothing else. So a Core's pin tile rendered `running: 0` / `needs-input: 0` no
matter what that Core was running, and no dot ever lit or cleared.

## What this does

**Round 1** gave the counts a source: `taskCountsFromCoreTasks` derives one project's block from the Core's own
`tasksList` snapshots — the same frame, and so the same rows, the grid renders from — `coreTaskCountsByProject`
buckets a whole Core's answer by project id, and `projectRowFromSnapshot` takes the block as an optional third
argument. Archived rows (their own list, ADR 0019) stay out of every bucket; an unrecognised status counts toward
`total` only. The Core stays the one place a Session's status is decided.

**Round 2** wires it, now that PR #456 has landed and `lib/use-fleet.ts` is free:

- `lib/core-pins-engine.ts` — the rail's pinned-projects fan-out, moved out of the hook into module state
  refcounted by its subscribers. One pass reads each Core's `listProjects` **and** `listTasks`, joins the counts
  onto the pins it just read, and publishes one snapshot every mount reads through `useSyncExternalStore`.
- `shared/core-pin-counts.ts` — `corePinTaskCounts`, the rule for what a pin shows when a Core answers and when
  it cannot.
- `lib/fleet-refresh.ts` — the poll cadence, the task-event predicate, the structural compare and #389's
  coalescing loop, now shared by both fan-outs instead of copied into the second one. `useFleetTasks` builds its
  loop from the same factory; its flags live in the closure, so rebuilding the pass function can no longer reset
  them mid-burst.
- `subscribe-core-project-events.ts` — exports `isProjectListEventKind`, so the engine can watch project *and*
  task kinds through the one `onEvent` handler it already holds rather than re-listing the kinds.

`ProjectBar.tsx` is untouched: it already reads `project.taskCounts` off the row, which is exactly what now
carries the truth.

## The three points from PR #456's review

**1. An unreachable Core must not zero its pin dots.** It does not. When a Core cannot be asked — the service
reports its link down, or the read fails — `corePinTaskCounts` returns the counts that Core last gave, and the
engine keeps its pins too. No dot is not a neutral rendering: on a rail it reads as "nothing running on this
project", which is a claim the Panel cannot make about a machine it just failed to reach. A Core that *does*
answer and reports no rows for a project is a real zero, which is how a finish clears the dot. A pin the tab has
never had counts for falls back to zeros — there is nothing else to show.

The unreachable Core is also not asked at all: a Core the service knows it cannot reach is not worth a frame the
router would only answer with an error. That is asserted in the test, not just intended.

**2. Do not add a second or third fan-out per event; share one engine, and state the cost.** Shared. The hook is
mounted twice on every route and three times on Fleet, and every mount used to run its own `listProjects`
fan-out; deriving dots per mount would have made that two or three `tasksList` frames per Core per event. The
engine gives the whole tab one set of Core watches, one `onEvent` subscription, one poll and one pass, so the
reads scale with the Cores rather than with the mounts. A test mounts the hook three times, emits one event, and
asserts exactly one `listProjects` and one `listTasks`.

*The cost, stated:* on the Fleet route, `useFleetTasks` still runs its own task fan-out for the grid, so a task
event there is two `tasksList` frames per Core (grid and rail) rather than one. Folding the grid onto this engine
is the obvious next step and is deliberately not taken here — that hook is #389's, its coalescing loop is what
this shares, and its tests pin exact read counts. *What it saves, on every route:* the per-mount `listProjects`
fan-out, which was 2-3 reads per Core and is now one. A new mount still asks for an answer rather than waiting
for the poll; the coalescing loop folds that into the pass already running.

Two things stay per-mount and are unchanged by this PR: `useCores` (an HTTP registry poll each) and the shell's
own rail-chord merge. Sharing the registry read is a separate cleanup.

**3. `refreshRemotePinned` after a pin toggle must refresh the counts too.** It does — `refresh` runs the same
pass, so pins and counts move together and a toggled tile never lands carrying the previous answer's counts.
Covered by a test that changes both the pin list and the task list before calling `refresh`.

## Tests

- `lib/__tests__/core-pins-activity.test.tsx` (7 cases, jsdom): a running Core Session lights the matching count
  and dot; a finish clears it on the event with nothing remounting; a dot appears for a Session started after the
  pin was on screen; an unreachable Core keeps its pins and their last counts and is not asked, then goes live
  again when it comes back; one event costs one fan-out with three rails mounted, and all three see it; `refresh`
  moves counts as well as pins; one project's tasks never count toward another's.
- `shared/__tests__/core-pin-counts.test.ts` (5 cases): the answer/could-not-ask rule in isolation.
- `shared/__tests__/core-task-counts.test.ts` (11 cases, round 1): the derivation itself, through the real
  `getPinnedProjectStatusDots` and `getProjectActivity`.
- #389's `use-fleet-refresh.test.tsx` stays green over the extracted coalescing loop — the four cases that pin
  exact read counts are what say the extraction was faithful.

## Acceptance criteria

- [x] **1. A running Core session lights the matching pin dot.**
- [x] **2. A finished Core session clears it without a full reload.**

Both proven end to end through `useRemotePinnedProjects`, the hook `ProjectBar` draws the rail from.

https://claude.ai/code/session_017Gq4vPU3ePSHpNPakKd4LU


qcentic-dev:
## Issue #377 acceptance checklist

- [ ] **1. A running Core session lights the matching pin dot.**
      **Unchecked.** Proven at the derivation, not in the app. `core-task-counts.test.ts` builds a pinned Core
      row from a `tasksList` snapshot carrying one `running` task and asserts `taskCounts.running === 1`,
      `getPinnedProjectStatusDots(row.taskCounts) === ["running"]` and `getProjectActivity(row) ===
      "agent-running"` — the very helpers `ProjectBar` draws the dots with. It does not light in the running
      Panel because the last hop is `useRemotePinnedProjects` (`packages/panel/src/lib/use-fleet.ts`) passing
      the counts into `projectRowFromSnapshot`, and that file plus `ProjectBar.tsx` are out of bounds for this
      change (#389 in flight / #383 later; #376, #378, #382 own the bar). The exact one-line wire-up is in the
      PR body.

- [ ] **2. A finished Core session clears it without a full reload.**
      **Unchecked, same blocker.** The clearing half is proven over consecutive reads of the same Core: with the
      task `finished`, `taskCounts.running` is 0, the running dot is gone from
      `getPinnedProjectStatusDots`, and activity is back to `offline`. "Without a full reload" additionally
      needs the per-Core fan-out to re-run on a task-status event; today `refreshNonce` is bumped only by
      `subscribe-core-project-events`, which admits `project:*` kinds only. Widening that filter (or
      subscribing the hook to task events the way `useFleetTasks` already does) is part of the same wire-up and
      also lands in `lib/use-fleet.ts`.

Also verified and green, though not acceptance criteria: archived Core rows light nothing (ADR 0019), a status
this Panel does not know counts toward `total` only, per-project counts never leak across projects on one Core,
`needs-input` outranks `running` in rail activity, and a caller that passes no counts still gets zeros — so
every existing caller of the mapper is unchanged.

### Why the PR stops here

`ProjectBar.tsx` reads `project.taskCounts` and nothing else to decide a dot, and Core-owned rows reach it only
through `useRemotePinnedProjects`. There is no third path — `__root.tsx` passes `ProjectBar` no project data —
so the dots cannot follow Core task status without editing one of the two barred files. The alternative that
avoids both, a module-level Core-task-count cache read from the pure mapper, is the second status authority
#377 explicitly rules out.

### CI on the PR head (`55f9045defd743fa7113d613f8eb0aa1d4d66cdb`)

**All green.** Every required check passed on run
[33674879391](https://github.com/actana/control/actions/runs/33674879391): Conventions, Typecheck, Lint, Unit
Tests, Dependency Audit, Secret Scan, E2E — Panel service seam, E2E — installer (debian/ubuntu, x64), Smoke —
Core release tarball (linux-x64 and linux-arm64), Panel image and Core image build + smoke (amd64), Resolve PR
image mode, Train rules, and Promotion gate. The train-only jobs (Train versions, Resolve train tags, Panel
image (train), Core image (train), both manifest publishes) skipped, as they do on a PR that does not target a
train. No failures, nothing pending.

qcentic-dev:
## Round 2 — acceptance checklist for #377

PR #456 (issue #389) has merged into the feature branch, so `lib/use-fleet.ts` was free and the wiring this PR
was blocked on is now in. Both boxes are checked, end to end through `useRemotePinnedProjects` — the hook
`ProjectBar` draws the rail from. `ProjectBar.tsx` is untouched: it already reads `project.taskCounts` off the
row, which is exactly what now carries the truth.

- [x] **1. A running Core session lights the matching pin dot.**
      Each pin row's `taskCounts` is derived from that Core's own `tasksList` answer — the same frame, and so
      the same rows, the grid renders from — joined onto the pins in the same pass.
      Proven in `lib/__tests__/core-pins-activity.test.tsx`: with the Core reporting one `running` task, the
      hook's pin row has `taskCounts.running === 1` and `getPinnedProjectStatusDots(row.taskCounts)` is
      `["running"]`. A second case starts with no tasks, emits `task:statusChanged` after the pin is already on
      screen, and watches the dot appear.

- [x] **2. A finished Core session clears it without a full reload.**
      The engine re-reads on the Core's own event, not on a mount: the test sets the Core's answer to
      `finished`, emits `session:finished`, and the same mounted hook settles to `running: 0`, `finished: 1`,
      with no `"running"` dot. Nothing remounts, nothing reloads, and no timer is advanced — the poll is 15s and
      the test finishes in milliseconds, so the update observed is the event's.

### The three points from PR #456's review

- [x] **An unreachable Core must not zero its pin dots.** When a Core cannot be asked — the service reports its
      link down, or the read fails — `corePinTaskCounts` returns the counts that Core last gave, and the engine
      keeps its pins too. No dot is not a neutral rendering: on a rail it reads as "nothing running on this
      project", a claim the Panel cannot make about a machine it just failed to reach. A Core that *does* answer
      and reports no rows is a real zero, which is how a finish clears the dot; a pin the tab has never had
      counts for falls back to zeros, there being nothing else to show. The test drops the link with a
      dial-status push, asserts the pin and its `running: 1` survive, asserts the Core was *not* asked (an
      unreachable Core is not worth a frame the router would only answer with an error), then brings it back and
      watches the count go live again.

- [x] **No second or third fan-out per event — share one engine, and state the cost.** The fan-out moved out of
      the hook into `lib/core-pins-engine.ts`: module state refcounted by its subscribers, with one set of Core
      watches, one `onEvent` subscription, one poll and one pass for the whole tab, read through
      `useSyncExternalStore`. Reads scale with the Cores, not with the mounts. A test mounts the hook three
      times, emits one event, and asserts exactly one `listProjects` and one `listTasks` — and that all three
      mounts see the result.
      **The cost:** on the Fleet route, `useFleetTasks` still runs its own task fan-out for the grid, so a task
      event there is two `tasksList` frames per Core (grid and rail) rather than one. Folding the grid onto this
      engine is the obvious next step and is deliberately not taken here — that hook is #389's, its coalescing
      loop is what this now shares, and its tests pin exact read counts.
      **What it saves, on every route:** the per-mount `listProjects` fan-out, 2-3 reads per Core, now one. A
      new mount still asks for an answer rather than waiting for the poll; the coalescing loop folds that into
      the pass already running. Still per-mount and untouched here: `useCores` (an HTTP registry poll each) —
      sharing that is a separate cleanup.

- [x] **`refreshRemotePinned` must refresh the counts as well as the pins.** `refresh` runs the same pass, so
      both move together and a toggled tile never lands carrying the previous answer's counts. The test changes
      the pin list *and* the task list before calling `refresh`, then asserts both changed.

### Tests and checks

`lib/__tests__/core-pins-activity.test.tsx` (7 cases, jsdom) and `shared/__tests__/core-pin-counts.test.ts`
(5 cases) are new; round 1's `shared/__tests__/core-task-counts.test.ts` (11 cases) still covers the derivation
through the real `getPinnedProjectStatusDots` and `getProjectActivity`. #389's `use-fleet-refresh.test.tsx` stays
green over the extracted coalescing loop — those four cases pin exact read counts, which is what says the
extraction was faithful.

Panel suite locally: **153 files, 1674 tests, all passing.** ESLint reports no new warnings in the changed files.

### CI on the PR head (`76867bf`)

**All green** on run
[33689421308](https://github.com/actana/control/actions/runs/33689421308): Conventions, Typecheck, Lint, Unit
Tests, Dependency Audit, Secret Scan, E2E — Panel service seam, E2E — installer (debian and ubuntu, x64),
Smoke — Core release tarball (linux-x64 and linux-arm64), Panel image and Core image build + smoke (amd64) with
their registry resolves, Resolve PR image mode, Train rules, Promotion gate. Train-only jobs skipped, as they do
on a PR not targeting a train.

The head before it (`d09887c`) failed Typecheck on this PR's own new test file — a helper that narrowed the
hook's answer to one project was typed `{ id: string }`, so the `taskCounts` assertions read a property that
type does not have. Vitest transpiles rather than checks, which is why the suite was green while `tsc` was not.
Fixed in `76867bf`; nothing in the production change moved.

mrfqcentic (approved):
Reviewed round 2 against #377 at `76867bf`, diffed three-dot against `origin/fix/operator-pins-sessions-drop-cli`.
I read the enforcing lines rather than the PR body, and I carried in the three points PR #456's review parked here.
**Approving.**

## The bug this ticket exists for, and where it now dies

`projectRowFromSnapshot` structurally zeroed `taskCounts` for every Core-owned row, and `useRemotePinnedProjects`
read each Core's *projects* and nothing else — so the block `ProjectBar` draws its dots from was constant zero
regardless of what the Core was running. The kill is a three-hop chain, and all three hops are in this diff or
already in the tree:

1. `lib/core-pins-engine.ts:98-101` reads that Core's own `tasksList` in the same pass as `listProjects`;
   `:103-107` derives the per-project blocks; `:110` passes the block as `projectRowFromSnapshot`'s new third
   argument.
2. `shared/projects.ts:165` — `taskCounts: taskCounts ?? emptyTaskCounts()`. The structural zero is gone; a
   caller with no task facts still gets zeros, so no existing caller changes behaviour.
3. `components/views/ProjectBar.tsx:1067` — `getPinnedProjectStatusDots(project.taskCounts)`, unchanged, reading
   the row that now carries the truth. `ProjectBar.tsx` really is untouched, and it did not need to be.

The derivation itself (`shared/projects.ts:99-113`) is line-for-line the Panel server's own aggregation
(`server/services/projects.ts:170-174` and `:228-231`): non-archived only, `total` over all of them,
`activeNonDone` = active-and-not-`finished`, an unrecognised status toward `total` alone. No second status
authority — it counts the very rows `tasksList` answered with.

## Acceptance criteria

**1. A running Core session lights the matching pin dot — MET.** Enforced at the three hops above. Held by
`lib/__tests__/core-pins-activity.test.tsx:164-172`, which drives the real `useRemotePinnedProjects` against a
fake Core reporting one `running` task and asserts both `taskCounts.running === 1` and
`getPinnedProjectStatusDots(pin.taskCounts) === ["running"]` — the helper `ProjectBar` actually calls, not a
restatement of the count. `:189-199` adds the harder direction: pin on screen first, tasks empty, then
`task:statusChanged`, then the dot appears.

**2. A finished Core session clears it without a full reload — MET.** The update path is
`core-pins-engine.ts:150-152`: one `onEvent` handler admitting `TASK_EVENT_KINDS` **or**
`isProjectListEventKind`, so a task event moves the dots and a project event moves the pins through the same
pass. `:174` publishes to every mount via `useSyncExternalStore` (`use-fleet.ts:403-409`). Held by
`core-pins-activity.test.tsx:174-187`: same mounted hook, Core's answer flipped to `finished`, one
`session:finished` emitted, settles to `running: 0` / `finished: 1` with no `"running"` dot. Nothing remounts and
no timer is advanced anywhere in the file, while `FLEET_POLL_MS` is 15s — so the poll demonstrably is not what
moved it. (The tile keeps a *finished*-coloured dot afterwards, which is exactly what a Panel-owned project does
with the same counts, so that is parity and not a leak.)

## The three points PR #456's review parked here

**1. The offline-Core rule — MET, and it is the part I checked hardest.** Two independent guards:

- `core-pins-engine.ts:90-91` — a Core whose `dial.state !== "connected"` is not asked at all and keeps the rows
  it last gave; a `listProjects` that throws returns the same remembered rows (`:95-97`).
- `shared/core-pin-counts.ts:38-39` — when `tasks` is `null` (the Core answered its projects but not its tasks),
  each pin takes its last-known count rather than a zero, and `core-pins-engine.ts:108` only overwrites
  `lastCountsByCore` when the Core actually answered.

`core-pins-activity.test.tsx:201-221` is load-bearing in all three directions: it drops the link with a real
dial-status push, then asserts the pin survives (an implementation returning `[]` fails on `toHaveLength(1)`),
that `running` is still 1 (an implementation zeroing the counts fails), and that `listTasksCalls` did not move
(an implementation that asks an unreachable Core anyway fails) — then brings the link back and watches the count
go live again. That is the failure #456 named, closed at the derivation rather than argued away.

**2. The double-mounted cost — HANDLED for the multiplier, DISCLOSED for the remainder.** The fan-out is module
state refcounted by its subscribers (`core-pins-engine.ts:56-65`, `:209-225`), with one set of `watchCore`
holds, one `onEvent`, one poll and one coalesced pass for the whole tab; `setCorePinsCores` (`:182-193`)
de-dupes the registry every mount pushes by signature, so N mounts pushing the same answer is N string compares.
`core-pins-activity.test.tsx:223-244` mounts the hook three times, emits one event and asserts the delta is
exactly one `listProjects` and one `listTasks`, and that all three mounts see the result. Under a per-mount
derivation that assertion reads 3. The residual — `useFleetTasks` still running its own `tasksList` for the grid
on the Fleet route — is stated in the PR body and in the module docstring rather than hidden, and folding the
grid onto this engine is correctly left to whoever owns that hook.

**3. `refreshRemotePinned` still behaves — MET, and it does more than it did.** `use-fleet.ts:415` returns
`refreshCorePins`, which runs the same pass, so the pin toggle at `ProjectBar.tsx:1394` now moves the dots as
well as the tiles instead of landing a freshly-pinned tile with the previous answer's counts (#456 finding 7).
Held by `core-pins-activity.test.tsx:246-263`, which changes *both* the pin list and the task list before
calling `refresh` and asserts both moved. Identity is now module-stable, and the only call site is a click
handler, so nothing depended on the old `useCallback` identity.

**The trailing-pass property — still held, with one gap worth naming.** The `do…while` from #389 moved verbatim
into `lib/fleet-refresh.ts:63-83`; `useFleetTasks` builds its loop from the same factory (`use-fleet.ts:189-196`)
and #456's four cases in `use-fleet-refresh.test.tsx` are unchanged, so the mutation that walked through #456
round 1 still stops at case 3. The extraction is faithful — the flags moved from refs into the closure, which is
strictly better, since rebuilding `fanOut` can no longer reset them mid-burst. What is *not* covered: the loop
now has a second consumer, and nothing exercises a burst against the pins engine (its fake bridge resolves
synchronously, so no pass is ever in flight when an event lands), nor is `createCoalescingRunner` unit-tested
directly. See finding 4.

## Scope, base and mergeability

- Nine files, all `packages/panel/src/{lib,shared}`. `ProjectBar.tsx` (#378/#382), `terminal-store.tsx`,
  `projects.$id.tsx`, `TerminalPane.tsx` and `__root.tsx` are all untouched. I diffed the file list against
  every open PR (#462, #468, #471, #472): **zero overlap with any of them.** `use-fleet.ts` is free — #389/PR
  #456 is merged.
- Round 2 came in as a plain merge, `3e28afa` with parents `55f9045` and `9272bc3`; round 1's commit is still in
  the history. No rebase, no force push.
- The base has moved four commits since this branch was cut (`9b71076`, `c4075fd`, `ce9f781`, `7e75832`).
  None touches `use-fleet.ts`, `shared/projects.ts` or any file this PR adds — they are `ProjectBar.tsx`,
  `projects.$id.tsx` and `packages/core`. **Still merges cleanly:** `git merge-tree --write-tree` against the
  base tip `7e75832` and against `main` `44f1119` both return a tree with no conflict, and GitHub reports
  `MERGEABLE` / `CLEAN`.
- CI is green on `76867bf` — 18 pass, 6 train-only skips, run 33689421308 — including Typecheck, Lint and Unit
  Tests. I did not re-run the suite locally; the CI job on this exact head is my evidence for green, and the
  reasoning above is from reading the enforcing lines.

## Non-blocking

I ran a second pass with `/code-review` at high effort over the same diff; it checked the head out into a scratch
worktree, ran the new and neighbouring suites (65 files / 595 tests green) and typechecked the panel. Its
findings are folded in below, with my own verification of each — including one I had to walk back.

**1. `core-pins-engine.ts:182-193` — the engine trusts that every mount pushes the same registry, and nothing
enforces it. Latent today, one mount site away from live.** `setCorePinsCores` compares only against the last
signature, and every `useRemotePinnedProjects` mount pushes its *own* `useCores()`, which starts at `[]`. A mount
appearing while the engine is already populated therefore pushes `""`, which reads as a real change: `cores`
becomes `[]`, `stop()` releases every `watchCore` hold and the poll timer, `start()` re-arms nothing, and `run()`
publishes `EMPTY` (`:120-123`). The mount that *had* the right answer cannot repair it — its `cores` identity has
not changed, so its effect does not re-run — and recovery waits on the new mount's own `listCores`, or on a 15s
registry poll if that read fails. During that window the rail's Core pins and their new dots are gone and the tab
holds no Core watch at all.

**This does not fire on the tree as it stands, and I want to be precise about why:** there are exactly two mount
sites, `__root.tsx:266` and `ProjectBar.tsx:91`, `ProjectBar` is a static import rendered unconditionally at
`__root.tsx:651` inside the very component that holds the other mount, so both mount in the same commit with both
registries still `[]` — and `""` against `""` is correctly a no-op. The failure needs a *staggered* mount, which
today's tree does not produce. But the invariant is undocumented (`:176-177` states the assumption as if it were
guaranteed), unguarded, and silent when broken; a third mount site, a lazy boundary, or making the rail
conditional — which #378/#382 are actively working in — turns it on with no test to catch it. Ignoring a push
from a mount whose registry has not loaded, or keying the registry per subscriber, closes it cheaply. I would
take this one.

**2. `core-pins-engine.ts:117-141, 150-152` — the pass fires far more often than a pin row can change, and the
cost section does not mention the expensive half.** The handler ignores `msg.coreId` and matches the whole
`TASK_EVENT_KINDS` regex, so `pty:spawn`, `pty:exit`, `session:delivered` and every task event on a Core with no
pinned projects each drive a full pass. That is defensible for `listTasks` — the dots are what the event is
for — but a pass also re-reads `listProjects` for every Core *and* does an `/api/project-presentation` HTTP round
trip (`:130`), and neither can have changed on a task event. In the code this replaces, the presentation read sat
in an effect keyed on `[bridge, coreIds, refreshNonce]`, so it ran on a registry change or an explicit refresh —
never on task traffic. So this PR turns ordinary session event traffic into one Panel HTTP request plus `2N`
core-link frames per event, **on every route**, where it previously cost nothing off Fleet. The PR body's "cost,
stated" section covers only the Fleet-route double `tasksList` and omits this. Given the PR's own standard —
review point 2 was *share one engine and state the cost* — this belongs in that paragraph, and splitting the
pass so a task event re-reads only tasks would remove most of it.

**3. `core-pins-engine.ts:94-101` — the two reads are serialised where they need not be.** `listProjects` is
awaited, then `listTasks`; nothing in the second depends on the first. Every pass costs two sequential round
trips per Core instead of one, which is also the rail's dot latency. `Promise.all([...])` with the two `catch`es
kept distinct (they mean different things here) is equivalent and halves it.

**4. `core-pins-engine.ts:209-225` — teardown does not cancel a pass in flight.** The last unsubscribe resets
`snapshot`, `cores` and both `last*ByCore` maps, but a `run()` launched before it still resolves and calls
`publish(rows)`, writing the previous registry's rows into module state with nothing mounted to correct them; the
next mount's first `getCorePinsSnapshot()` renders them. In the app this is inert — `__root.tsx` means the last
subscriber never actually leaves — but it is what makes `core-pins-activity.test.tsx` order-dependent: `afterEach`
calls `cleanup()` and nothing drains the in-flight pass, so a case like `:189-199`, which asserts `running === 0`
straight after `toHaveLength(1)`, could observe the previous case's `running: 1`. Not observed failing (CI is
green and the suite ran green in the scratch worktree), but it is latent flake in the file that carries both
acceptance criteria. A generation counter checked before `publish` closes the app path and the test path at once.

**5. `core-pins-engine.ts:182-193` — a dial blink re-arms every Core watch, not just the pass.** The signature
includes `dial.state`, so a blink runs `stop(); start()`, releasing and re-acquiring every `bridge.watchCore`
hold. Those are refcounted (`panel-link-client.ts:276-292`) and this engine is the only holder off the Fleet
route, so refs go 1 → 0 → 1 and a fresh `subscribe` frame goes out per Core per blink; the 15s poll timer resets
each time too, so a flapping link can postpone it indefinitely. No event is lost — the teardown and re-arm are
synchronous — so this is cost, not correctness. The sibling gets the split right and is worth copying:
`use-fleet.ts:216` keys the watch effect on `coreIds` and re-runs only the *pass* on `coreSignature`.

**6. The mount count in the docstring and the PR body is wrong, and it is a load-bearing comment.**
`core-pins-engine.ts:23-25` says the hook is mounted "the shell …, `ProjectBar` …, and a third time on Fleet",
and `:41-42` claims a saving of "2-3 reads per Core". There are two mount sites; `FleetView` mounts
`useFleetTasks`, not this hook. The third mount was real in #456 round 1 only because `useRemotePinnedProjects`
went *through* `useFleetTasks` there. The saving is 2 → 1. The argument for the engine is untouched; the number
is not, and the same sentence is in the PR body.

**7. Test gaps.** (a) Nothing exercises a burst against the pins engine — its fake bridge resolves
synchronously, so no pass is ever in flight when an event lands — and `createCoalescingRunner` has no direct
unit test; the extraction is guarded only through `useFleetTasks`. Since the engine is now the loop's second
consumer and its event→pass path is what makes AC 2 hold under a burst, parking a `listTasks` the way
`use-fleet-refresh.test.tsx` already does would pin it cheaply. (b) The "answered its projects but not its
tasks" branch (`:98-108`) is untested at the engine; `core-pin-counts.test.ts:41-50` covers `tasks === null` in
isolation, but nothing drives a `listTasks` rejection through `readCore`, and that is a distinct path from the
dial-down one. (c) No case asserts that a non-matching event kind does *not* trigger a pass.

**8. `core-pins-engine.ts:60-63` — `lastRowsByCore` / `lastCountsByCore` are never pruned for a Core dropped
from the registry.** They reset only when the last subscriber leaves, which in practice never happens. Growth is
bounded by the number of Cores ever registered in one tab, so this is a nit rather than a leak; a re-added Core
serving pre-removal rows needs the same `core_<hex>` id back, which a re-pair does not give you.

**9. The issue's "Where" names a second call site that this PR leaves zeroed.** #377 says
*"`useRemotePinnedProjects` and `useCoreProjectRows` both use that mapper"*. `useCoreProjectRows`
(`use-fleet.ts:326-329`) still passes no counts, so `ProjectPicker.tsx:398` draws a dark running dot for every
Core-owned project. Both acceptance criteria are about the *pin* dot, so I am not holding the PR for it — but it
wants a follow-up issue rather than being quietly implied done when this ticket closes.

**10. An unreachable Core's tile now shows a stale dot with no staleness affordance.** This is the right call and
the one #456 asked for — a dark dot is a claim the Panel cannot make. But `ProjectBar` has no dial-state
treatment for a pin tile (its only dial read is `fileCapableCoreIds`, `:121`), so a running dot on a Core that
went down an hour ago is also a claim. Worth a follow-up: dim or annotate the tile from `useCores`.

**11. Nit — `subscribeCorePins:209-214` can start the engine cold.** `start()` no-ops when `getPanelBridge()` is
null (`:146`), and it is called only for the first listener, so if the bridge is not up at that instant nothing
re-arms until a Core-signature change happens to call `stop(); start()`. It self-heals in practice because the
registry always arrives after `[]` — but it self-heals through the same path finding 1 is about.

## Checklist

- [x] AC 1 — a running Core session lights the matching pin dot — **met**, enforced at
      `core-pins-engine.ts:110` → `shared/projects.ts:165` → `ProjectBar.tsx:1067`, held by
      `core-pins-activity.test.tsx:164-172` and `:189-199`.
- [x] AC 2 — a finished Core session clears it without a full reload — **met**, enforced at
      `core-pins-engine.ts:150-152`, held by `core-pins-activity.test.tsx:174-187`; no timer advanced, poll is 15s.
- [x] Core pin dots follow Core task status despite `projectRowFromSnapshot`'s structural zero — **met**; the
      zero is now `taskCounts ?? emptyTaskCounts()` and the counts arrive from the Core's own `tasksList`.
- [x] Offline-Core rule — **met**; `core-pins-engine.ts:90-91` and `core-pin-counts.ts:38-39`, held by
      `core-pins-activity.test.tsx:201-221` including "was not asked".
- [x] Double-mounted engine cost — **handled** (one module engine, refcounted, `useSyncExternalStore`), held by
      `core-pins-activity.test.tsx:223-244`; the Fleet-route grid overlap is disclosed, not hidden.
- [x] `refreshRemotePinned` still behaves — **met and improved**, `use-fleet.ts:415`, held by
      `core-pins-activity.test.tsx:246-263`.
- [x] Trailing-pass property — **still held** through `use-fleet-refresh.test.tsx` over the extracted
      `createCoalescingRunner`; not exercised against the new engine (finding 7a).
- [x] No parallel ticket's file touched — **clean**; nine files, zero overlap with #462/#468/#471/#472.
- [x] Merges cleanly on the moved base (`7e75832`) and on `main` — **yes**, `merge-tree` clean both ways.
- [x] CI green on `76867bf` — 18 pass, 6 train-only skips; the panel suites also ran green in a scratch
      worktree at this head (65 files / 595 tests).

**Verdict: approve.** Both acceptance criteria are met end to end through the hook `ProjectBar` draws from, the
three points #456 handed forward are each closed at the code and pinned by a test that fails without them, the
scope is clean, the branch still merges on the moved base, and CI is green on the head. None of the eleven items
above produces an operator-visible defect on the tree as it stands, which is why they are findings and not
conditions.

If you are touching this branch again anyway, I would take **1, 2, 3 and 6**: 1 is a guard against an invariant
that is currently true only by accident of where two components happen to mount, 2 is a real cost the PR's own
"state the cost" commitment should have named, 3 is two lines, and 6 is a comment that will mislead the next
reader about why the engine exists. 4 and 7 are the ones I would put on the follow-up list rather than here.

Still a draft; I have not marked it ready, and I have changed nothing in the tree.

PR #458, opened 2026-09-02
fix(panel): two tabs must not both type; dispose onData on reattach

### LINKED ISSUE #393: T2: two tabs can both type until drive/lock answers; failed reattach stacks onData

qcentic-dev:
Closes #393 (T2).

Two bugs, one pane. Both are about a keystroke going somewhere the operator did not intend, and neither is visible until
after it has happened.

## 1. Two tabs both typed until the drive answered

`session-write-store.ts` spelt "nobody in this Panel asked" and "this tab asked and is still waiting" the same way —
`drive: "none"` — and `none` reads as writable. So two tabs opened on one Session were both writable until the Panel
service happened to answer, with no bound on how long that was and nothing on screen saying the write was a guess.

The waiting state is now its own, `pending`, and it is writable **only inside a bounded window**:

- `OPTIMISTIC_DRIVE_WINDOW_MS` is 1.2s. A drive answer is arbitrated by the Panel's own service — no Core is asked and
  nothing crosses the wire (`wantDrive`) — so a pane on a healthy link never reaches the end of it.
- When the window closes unanswered the pane goes read-only for a third reason, `awaiting-drive`, and the store still
  believes a late answer when one arrives. Closing a window withdraws a guess; it does not decide the question.
- The strip above the header shows "Checking which tab drives this Session…" for the whole window — the *visible* half —
  and then the read-only label with the same "Drive here" affordance the intra-Panel handover already offers.
- Taking the keyboard no longer goes writable on the ask: the tab it is taking from believes it is driving until the
  service says otherwise, and typing in that gap is the same dual write.

The optimism itself is kept on purpose. A solo tab still types the instant it opens, and a pane that opened read-only
waiting for permission would render every Session on a single-connection Core as locked forever.

The strip also renders for any Core now, not only one announcing `multiConnection`. Read-only has to be on screen before
a keystroke, and the reason a pane may not type is never that a Core lacks a capability.

## 2. A failed reattach stacked xterm `onData` handlers

`wireTerminalInput` subscribed to `onData` and `onResize` and kept no handle on either, while `ensurePty` can wire more
than once for one surface: the descriptor's pty, then `findByTask` when that reattach comes back empty, then a spawn.
After a reload against a pty that had gone, every surviving handler fired on one keystroke and wrote it to the PTY
again — doubled characters in the shell, and a stray write to a pty that is not there any more besides.

`createTerminalInputWiring` (`lib/terminal-input-wiring.ts`) holds the pane's one input subscription across every attach
it makes: wiring disposes what the last attach wired, and the pane's teardown disposes the last of them. Extracted from
the pane's build closure so it can be asserted on without an xterm in jsdom, which is how this file is tested already.

## Tests

- `session-write-store.test.ts` — the window is short, it closes, an answer inside it settles, a late answer after it is
  still believed, a second pane of the same tab does not reopen it, a take does not go writable on the ask, and a lock
  answer neither opens nor closes it. Plus a `two tabs on one Session` suite driven by one first-come arbiter: at most
  one tab is writable once the drive is known, **zero** are when the answers never come, and both are inside the window.
- `terminal-input-wiring.test.ts` — a keystroke after a failed reattach reaches the PTY once, and reaches the pty the
  pane actually ended up on; every subscription an attach wired is disposed, not only the first; teardown is idempotent.
- The existing `session-write-store`, `session-write-access`, `task-status-sync`, `terminal-running-fallback` and
  panel-link suites are green, and so is the full `packages/panel` suite (149 files, 1626 tests).

## Not in scope

#187 — the older two-client lock-lag bug — is untouched. This is the tab drive and the handler leak.

Draft, based on `fix/operator-pins-sessions-drop-cli`, and built on top of the merged #386 change: the per-pty
running-fallback reset at the top of `wireTerminalInput` is preserved exactly as it was.

https://claude.ai/code/session_011svqMHMLxPjZPaPwYCpg9u


qcentic-dev:
## Acceptance criteria — issue #393

**1. Two tabs on one Session: at most one is writable once drive/lock is known, and the optimistic window is short and visible.**

- [x] **At most one writable once the drive is known.** `drive` gained a `pending` state so "nobody asked" and "asked,
  waiting" stop being the same fact. Once the service answers, exactly one tab of a Panel is told `driving: true`, and
  the other reads `{ writable: false, reason: "driven-in-another-tab" }`. Proved by
  `session-write-store.test.ts` → *"leaves at most one of them writable once the drive is known"*, driven by one
  first-come arbiter rather than hand-fed answers.
- [x] **At most one writable when the answer never comes.** The case that used to be unbounded. Both windows close and
  **neither** tab is writable — *"leaves at most one of them writable when the answers never come"*.
- [x] **The window is short.** `OPTIMISTIC_DRIVE_WINDOW_MS = 1200`, and the drive is arbitrated by the Panel's own
  service, not by a Core — a round trip a healthy pane never reaches the end of. Asserted at both edges: writable at
  `window - 1`, read-only `awaiting-drive` at `window`.
- [x] **The window is visible.** The pane's write-access strip — above the header, outside every collapsible tier —
  renders `Checking which tab drives this Session…` (`data-session-write-access="awaiting-drive-optimistic"`,
  `role="status"`) for the whole window, then the `awaiting-drive` read-only label with a "Drive here" button.
- [x] **The lock still wins.** A pane that may not write at all is told about the client holding the Session, not about
  a drive answer that would change nothing for it — *"keeps the guess across a lock answer…"*.
- [x] **Taking the keyboard does not go writable on the ask.** The tab being taken from believes it is driving until the
  service says so; the taking pane waits — *"does not hand the keyboard to a tab on its own say-so…"*.

**2. A failed reattach does not duplicate keystrokes.**

- [x] **`wireTerminalInput` disposes its `onData` (and `onResize`) subscription before wiring a new one.**
  `createTerminalInputWiring` in `packages/panel/src/lib/terminal-input-wiring.ts` holds the pane's single input
  subscription across every attach the surface makes; the pane's teardown disposes the last.
- [x] **A keystroke after a failed reattach is delivered once.** `terminal-input-wiring.test.ts` replays `ensurePty`'s
  real sequence (descriptor pty → empty replay → `findByTask` → wire again) and asserts one write, to the pty the pane
  actually ended up on. Also covered: four consecutive attaches, both subscriptions of an attach disposed (not only the
  first), and idempotent teardown.

### Deliberately not done

- [ ] **#187 (two-client lock lag / stays running).** Out of scope by the issue's own instruction — not duplicated, not
  touched.
- [ ] **A test that renders `TerminalPane` end to end.** This repo does not put xterm into jsdom (see
  `user-terminal-panel-control.test.tsx`, which mocks the pane out for exactly that reason). The disposal logic is
  extracted to `lib/terminal-input-wiring.ts` and asserted there instead, which is the established pattern for this
  file — `terminal-running-fallback`, `terminal-replay`, `terminal-prompt-capture` are all the same shape.
- [ ] **A hard guarantee that two tabs never both type, ever.** Impossible without knowing the answer, and the issue
  does not ask for it: the window is the trade, and the acceptance criterion asks it be short and visible, which it is.

### Boundaries

Nothing outside `packages/panel/src/{lib,shared,components/views}` was touched. `ProjectBar.tsx`, `use-fleet.ts`,
`panel-link/router.ts`, `active-group.ts`, `terminal-store.tsx`, `__root.tsx`, `projects.$id.tsx` and every
`packages/core`, `packages/shared` and `packages/cli` file are untouched. `main` and `beta/0.4.5` are untouched, nothing
was merged, nothing was force pushed, and this stays a draft.

qcentic-dev:
**CI on the PR head `009c40b`** — run [33678296522](https://github.com/actana/control/actions/runs/33678296522): **18 pass, 0 fail, 0 pending** (6 image/train jobs skipped as expected on a ticket PR).

`Conventions` (which lints every commit message, not just the title), `Lint`, `Typecheck`, `Unit Tests`, `Secret Scan`, `Dependency Audit`, `E2E — Panel service seam`, both installer E2Es, both Core tarball smokes, both image build+smokes, `Train rules`, `Promotion gate` and `Resolve PR image mode` are all green.

Still a draft, base `fix/operator-pins-sessions-drop-cli`. Not marked ready for review, nothing merged, nothing force pushed.

mrfqcentic (approved):
## Review against issue #393 — approved

Reviewed at head `009c40b`, against the PR's own base `fix/operator-pins-sessions-drop-cli` (merge base `31444a9`).
Read the enforcing lines rather than the description, ran the panel suite in a scratch copy (149 files, 1626 tests,
all green) and checked CI on the head commit (Conventions, Lint, Typecheck, Unit Tests, both E2E installer jobs,
Panel/Core image smoke — all success; the skipped jobs are the train-only ones).

Both acceptance criteria are met, and they are met by the code, not only by the prose.

### AC1 — two tabs, at most one writable once drive/lock is known; the optimistic window short and visible — **met**

- `packages/panel/src/shared/session-write-access.ts:77` splits `pending` out of `none`, and
  `session-write-access.ts:139` makes `pending` writable **only** while `optimistic === true`. An unknown drive is no
  longer a licence to type.
- `session-write-access.ts:89` — `OPTIMISTIC_DRIVE_WINDOW_MS = 1_200`, and `session-write-store.ts:100-118` is the whole
  of the bound: the timer re-derives the entry with the guess withdrawn and the pane lands on `awaiting-drive`.
- The answer never coming is handled at `session-write-store.ts:113-117`; a late answer is still believed, because
  `onSessionDrive` (`session-write-store.ts:154-162`) only ever `clearOptimism`s the timer and then puts the real answer
  — closing the window withdraws a guess, it does not decide the question. Both are asserted
  (`session-write-store.test.ts` — "stops typing on the guess when the window closes unanswered", "believes an answer
  that arrives after the window closed").
- Lock precedence is intact: `sessionWriteAccess` still answers `!lock.writable` first (`session-write-access.ts:135`),
  so `held-by-another-client` wins over any drive state, guess or not — asserted in "keeps the guess across a lock
  answer".
- A solo tab still types the instant it opens (`watchSessionDrive` → `pending` + `optimistic: true`), asserted in "lets
  the pane type while it waits".
- Visible half: `TerminalPane.tsx:1592-1594` now renders the strip for **any** Core — the `writeState.lock.supported`
  gate survives only on "You hold this Session", which is the one line that genuinely needs `multiConnection`. A
  Core-owned pane on a single-connection Core gets the notice and the read-only label it could not get before.

### AC2 — a failed reattach does not duplicate keystrokes — **met**

- `packages/panel/src/lib/terminal-input-wiring.ts` holds one subscription list and disposes the previous one *before*
  subscribing the next; `TerminalPane.tsx:1250` re-wires through it, and `TerminalPane.tsx:962` puts the final dispose in
  `subscriptions`, which `surface.teardown` (`TerminalPane.tsx:1503`) runs.
- The three attach paths that could stack (`wireExistingPty` on `descriptor.ptyId`, `wireExistingPty` on `findByTask`,
  `wireNewPty` after a spawn) all funnel through `wireTerminalInput`, so the reload-against-a-dead-pty sequence now
  delivers once. `terminal-input-wiring.test.ts` asserts exactly that, and that it reaches the pty the pane ended up on.
- The #386 work is preserved: `git diff -w` shows the `onData` body byte-identical apart from the closure wrapping, and
  the per-pty reset (`runningFallback = IDLE_TERMINAL_RUNNING_FALLBACK; promptCaptureBuffer = ""`) still runs at the top
  of `wireTerminalInput` (`TerminalPane.tsx:1245-1247`), before the wire. `task-status-sync.ts` is untouched and its
  suite is green.

### Scope, base, hygiene

- `#187` is not duplicated: nothing here touches lock delivery, `session:lockChanged`, or the stuck-`running` half.
- Untouched, as required: `ProjectBar.tsx`, `use-fleet.ts`, `panel-link/router.ts`, `active-group.ts`,
  `terminal-store.tsx`, `__root.tsx`, `projects.$id.tsx`, and all of `packages/core`, `packages/shared`, `packages/cli`.
- Base is `fix/operator-pins-sessions-drop-cli`, still a draft, mergeable/clean, and all three commits pass commitlint
  (no body or footer line over 132; trailers have their leading blank).
- The `CONTEXT.md` edit is a legitimate vocabulary addition — it extends the existing **Session drive** entry with the
  one fact the glossary was missing (an unanswered ask is bounded, not unarbitrated) and contradicts no ADR; ADR 0024 D3
  is upheld, nothing crosses the wire. One nit: the file cites issues as `(#287)` / `(#280)` elsewhere, so `(issue 393)`
  is the odd spelling out.

### Non-blocking findings — worth a follow-up, not a change request

1. **`packages/panel/src/components/views/TerminalPane.tsx:1054`** — `clearActivePty()` does not dispose the wiring, so
   the "stray write to a pty that is not there any more" is only half closed. Every `clearActivePty` path (the three in
   `handlePtyExit`, the empty-replay one at 1385, `ensurePty`'s catch at 1492) leaves the last-wired `onData` live, still
   holding its captured `ptyId`. Typing into a pane that has printed "Session finished (code=0)" writes to the dead pty
   and — because the row is no longer `running`, so `advanceTerminalRunningFallback` re-arms — PATCHes it back to
   `running` with no process behind it. **Pre-existing** (the old code disposed nothing at all), so not a regression from
   this PR, but the description overstates the fix. `inputWiring.dispose()` inside `clearActivePty` would close it.
2. **`packages/panel/src/lib/session-write-store.ts:207-208`** — the window is armed at the ask, not at the send.
   `PanelLinkClient.rawSend` queues while the socket is down and the reconnect backoff runs to 10s, so a pane that mounts
   during a flap can burn its whole 1.2s budget in the queue and flip a *solo* tab to
   `Read-only · waiting for this Panel`, whose detail sentence blames Panel arbitration rather than the dead link. It
   self-heals on `open` via `resendSessionDrives`, so this is a bounded mislabel rather than a stuck pane — but arming
   from the link's `open` (or not arming while `!isConnected()`) would keep the bound honest.
3. **`packages/panel/src/lib/session-write-store.ts:231`** — a `take` has no window and no feedback. If its answer is
   lost the pane parks at `awaiting-drive`, and clicking "Drive here" again re-enters `put` with identical
   `lock`/`drive`/`optimistic`, so the dirty check short-circuits, no `emit()` runs, and nothing on screen acknowledges
   the gesture. Deliberate that a take does not go writable on the ask — agreed, that is the right call — but the state
   could use a busy marker or a bound.
4. **`packages/panel/src/components/views/TerminalPane.tsx:1657`** — `awaitingDriveOptimistic` now suppresses the
   "You hold this Session" line and its Release button for the whole window, and indefinitely if the pane settles into
   `awaiting-drive`. The lock and the drive are independent facts (this module's own thesis), so an operator who
   demonstrably holds the lock loses the affordance to give it back for a Panel-local reason. Small, and the window is
   short, but the two could both be shown.
5. Test gap, minor: `readOnlyLabel("awaiting-drive")`, `readOnlyDetail("awaiting-drive")` and
   `AWAITING_DRIVE_OPTIMISTIC_LABEL` are not asserted anywhere, and both label helpers end in a bare fallthrough
   `return` rather than an exhaustive switch — a fourth reason would silently render "waiting for this Panel". The
   `session-write-access.test.ts` suite was not extended for the new variant.

Approving: both acceptance criteria are enforced by code I read and by tests that fail without the fix, the scope is
exactly the two bugs the issue names, and every finding above is an edge inside the new behaviour or a pre-existing gap
rather than a defect in what this PR set out to do.

Issue #442, opened 2026-09-02
events tail: a --limit ceiling advances the stored cursor past the events it printed


qcentic-dev:
**Pre-existing. Not introduced by #439**, and not made worse by it — reproduced with byte-identical numbers against
`a7b5532`, the commit that branch was cut from. Found by the round-2 review of #439 (the fix for #402) and filed
separately so it does not ride on that PR.

## What happens

`--limit` stops the *printing*, but it does not stop the *cursor*. `DurableCoreClient.deliverEvent` advances and
**persists** the stored cursor for every event it delivers, and the events of one synchronous burst keep arriving after
the ceiling has been reached and `finish` has closed the client. So the cursor lands at the end of the burst while
stdout stopped at the ceiling, and every event in between is silently dropped from the follow-along stream — no error,
no warning, nothing in the log to notice.

This is the same class of failure the event cursor exists to prevent, arriving from the other direction: not a replay
storm, but a cursor that ran ahead of what the operator was shown.

## Reproduction

From the review of #439, against a real in-process Core with 101 events in the log:

```
first run  : code=0 out=1  cursor=1
second run : --limit 5 → ids=[2,3,4,5,6]  storedCursor=101
third run  : nothing past #101 — events 7..101 are gone from the follow-along stream
```

The second run printed five events, exactly as `--limit 5` promises, and moved the cursor ninety-five events past the
last line it printed. The third run resumes from `#101` and events 7–101 are unreachable: they were never printed and
the cursor says they were.

**The same probe against `a7b5532` gives the same numbers.** The ceiling's `finish` is unchanged by #439.

## Why it matters more now

#402 established that `--limit` is a ceiling on a read of the log rather than a quota to wait on, and #439 made that the
advertised idiom in `EVENTS_HELP` (`--limit is a ceiling, not a quota`). A bounded read is now the natural thing for an
operator or a script to type, so a bounded read that quietly burns the follow-along cursor is more reachable than it
was — even though the defect itself predates all of it.

## Where

- `packages/cli/src/events-command.ts` — the `finish(EXIT_OK)` taken when `printed >= limit.value` in the `onEvent`
  handler. Closing the client does not un-persist what the durable client has already written.
- `packages/sdk/src/durable-core-client.ts` — `deliverEvent` advances the in-memory cursor and writes it through the
  `CoreLinkCursorStorage` it was handed, per event, with no notion of whether the consumer used the event.

## Possible directions, none of them decided here

[...]

Issue #441, opened 2026-09-02
events tail: an explicit --follow/--no-follow, so --limit stops carrying two meanings


qcentic-dev:
Follow-up from the review of #439, which fixed #402. Filed on the reviewer's recommendation; not a defect in that fix.

## What `--limit` means today

After #402, `actana events tail --limit n` means one of two things, and which one is decided by the state of the Core's
log at the moment of the run's first empty-tail marker:

- a run that started from `--since` or a stored cursor and **found events past that cursor** is a *read*: it prints them
  and exits at the end of the history, however far short of `n` it stops;
- a run that found **nothing** past its cursor — including every first run, which starts at the tip by design — is a
  *follow* with a ceiling: it waits, and stops after `n` events.

Both halves are documented in `EVENTS_HELP` and argued in the module header of `packages/cli/src/events-command.ts`, and
the split is deliberate: an unconditional "exit at the end of history" would end `--since start --limit 6` against an
empty log before its first line, which is the run `events-tail-cursor.test.ts` uses to prove #161's criterion across a
real Core restart.

## Why it is worth a flag

The cost, as the review of #439 put it:

> whether `events tail --since 13 --limit 30` terminates now depends on whether anything had been appended past #13 at
> the moment of the first empty-tail marker. The same command against the same Core can exit or follow depending on that
> race.

That is intent inferred from log state. It is faithful to #402 — whose remedy is explicitly scoped to "when the events
already exist" — but the operator's actual intent is not something the log can be asked about. A script that wants
"drain what is there and exit" and a script that wants "wait for the next n" type the same command and get different
behaviour on different days.

## What to do

Give the operator the verb, so `--limit` goes back to carrying one meaning — a ceiling — and *whether to wait* is said
out loud:

- `--no-follow` (or `--follow` as the explicit opposite; the naming is part of the ticket): read what the log holds past
  the cursor and exit, whether that is `n` events, fewer, or none at all.
- With the flag given, the log-state heuristic does not run: the answer comes from the flag.
- Decide the default. Keeping today's inferred behaviour as the default is backwards-compatible; making `--no-follow`
  the default for `--limit` is cleaner and is a contract change worth its own note.

Whatever the default, the two flags remove the case an operator cannot predict, and `--limit` stops meaning two things.

## Also worth deciding here

- The exit code when a `--no-follow` read is cut short by the 30s subscribe deadline. #439 settled on non-zero for a
[...]

PR #459, opened 2026-09-02
fix(cli): events tail --kind must print matching history

### LINKED ISSUE #402: F6: events tail --since/--limit can hang instead of exiting

### LINKED ISSUE #403: F7: events tail --kind session:finished --limit 1 skips history and never exits

qcentic-dev:
**Refs #403** (hunt `F7`) — deliberately *not* `Fixes`. On top of the merged #439 that fixed #402, building on that
change rather than against it.

> **#403 stays open when this lands.** Its criterion 2 — "Ctrl-C does not persist a cursor past unseen matching
> events" — is not satisfied by this PR and cannot be from here. The stored cursor is advanced and persisted by
> `DurableCoreClient.deliverEvent` before any listener runs, so it outruns stdout on the ordinary exit path as well as
> under an interrupt. That mechanism is **#442**, which must not be fixed in this PR; the exit-path case is filed there
> with numbers. This PR closes #403's criterion 1 and makes the command answer the right question; #403 should be
> closed once #442 lands.

> **Merge or rebase this branch — do not squash it.** The repo's squash message mode is `COMMIT_MESSAGES`, so a squash
> concatenates all four commit bodies into one message with no ordering cue. `d7bfb7d`'s body claims that printing on
> the way past "keeps the cursor honest" and that "an interrupt cannot leave the cursor beyond a match that was never
> shown" — a paragraph `3f1b3bd` exists to retract, and `89383a0` narrows again. A merge or a rebase keeps each
> correction adjacent to the claim it corrects; a squash would land the retracted claim and its retraction side by side
> as if both were current. (Round-2 review nit.)

## What the operator saw

After a session had already finished, `actana events tail --kind session:finished --limit 1` — no `--since` — printed
nothing and never exited. Ctrl-C left the cursor past the finish.

## Why

A first run has no stored cursor, so it is walking to the tip of the log: the Core has no frame that answers "where does
your log end", and the only way to learn it is to subscribe, let the tail stream, and read the `eventsReplayed` markers
until one closes an empty tail (`event-tip.ts`). Everything in that walk was **counted and not printed** — including the
`session:finished` the command was typed to find. The only way out of the printing loop was a *second* finish that
nobody was going to append.

## What changed

`packages/cli/src/events-command.ts`, and its two suites. Nothing else in the tree; no `packages/core`,
`packages/shared`, `packages/panel` or `packages/sdk` file is touched, and neither is `session-command.ts` or
`session-gateway.ts`.

A first run starts at the end of the log because a history nobody asked for is the replay storm. **`--kind` is the
asking**: the operator has named the kinds they came for, so a named kind is taken out of that walk and every other kind
stays as quiet as it has always been. What decides that is the flag the operator typed, never the state of the log.

**Which** matches is a second question, and it is the one round 1 got wrong. A first run's walk covers the whole
retained log — it subscribes from `#0`, and nothing prunes the store — so printing matches as they go past answers with
the *oldest* match the Core still holds. Newest needs the end of the log, and the end of the log needs the walk. So the
walk holds the last `n` matches it has seen and prints them once the tip is known: **`n` and no more, whatever the log's
length, and the answer is the newest `n`.** Without `--limit` there is no `n` to choose between, so matches stream as
they arrive — every one is going to be printed anyway, and holding them would buy nothing and cost an unbounded array.

If the Core goes quiet mid-walk, what is held goes to stdout before the run gives up, with the count named on stderr and
the exit still non-zero — the matches are past the cursor and the cursor has already moved over them, so dropping them
would lose them for every later run to buy nothing.

`EVENTS_HELP` says all of this under *Where it starts*, including that `--kind` bounds *what* is printed and only
`--limit` bounds *how much*; the module header carries the argument.

### The tip the hunt could only approximate

`event-tip.ts` gives up after `MAX_TIP_ROUNDS` against a Core appending faster than this side can drain it, and hands
back its approximation as an ordinary number. A walk that took that for the end of the log flushed its ring and exited
**0** — with the oldest matches in the log, since the Core had raced ahead — while stderr said the run was "carrying on"
after it had exited. `event-tip.ts` names that caller by hand as the one that must not take its default sentence.

So `tip` now carries an `onRoundsExhausted`, and the walk ends the way the reading side already ends: its own sentence
on stderr and a **non-zero exit**, with what the ring holds handed over first through the same path the 30s deadline
uses.

The condition is `recentCap > 0`, not `--limit`. What splits `event-tip.ts`'s two callers is whether the walk *is* the
answer: `--limit 5` with no `--kind` prints nothing on the walk and spends its ceiling on live events, so it really is
"a first run about to follow" and keeps the default sentence; `--limit 5 --kind session:finished` ends on the ring, so
an approximate tip costs it the answer. The reasoning is at `tipEndsThisRun` and both halves are pinned by a test.

The ring's `shift` is now a rotated write index — O(1) per match at the same bounded memory, where the old one moved
every held element on every match once full.

## Deliberately not in scope

- **#441**, the explicit `--follow` / `--no-follow` design. Not taken: this change adds no flag and does not touch what
  `--limit` means for a run that has a cursor.
- **#442**, the cursor running ahead of what was printed. Not taken, and now filed with the exit-path case this PR makes
  reachable. `DurableCoreClient` and the ceiling's `finish` are untouched.

## How it is proven

Against the **real in-process Core**, a real `wss://` link, the real durable client and the real cursor file
(`events-tail-cursor.test.ts`):

- **the newest match, not the oldest** — the reviewer's own log, finishes at `#3` (weeks old, different `taskId`) and
  `#22` (just watched), `--limit 1` → `[22]`. Returns `[3]` on the reviewed head `6c8474f`.
- **the newest across a capped replay** — 1500 events, finishes at `#500`/`#1200`/`#1450`, `--limit 2` → `[1200, 1450]`.
  Returns `[500, 1200]` on `6c8474f`.
- **exactly once and in order across the cap** — the same log, `--limit 3` → `[500, 1200, 1450]`, exit 0. The case the
  review asked for by name.
- **criterion 1** — a finish already in the log, `--kind session:finished --limit 1` on a machine that has never
  followed this Core: one event, the right one, and an exit. Hangs on the parent `31444a9`.

At the **frame level** against a fake Core (`events-command.test.ts`):

- the first-run `--kind` run frame for frame, asserting that **nothing is printed until the tip is known**;
- **the ring across a re-ask** — three finishes over two rounds of a capped replay, `--limit 2` → the last two;
- **the hand-over on a silent Core** — held matches on stdout, `held from an unfinished walk` on stderr, exit non-zero;
- **the guard on the narrowing** — a first run given `--kind session:finished` still prints none of the `task:*` history
  it walks past.

The real-Core interrupt test from round 1 is **removed**: with the ring it asserted something that is no longer true,
and the criterion it covered now belongs to #442. That is stated rather than quietly dropped.

With the fix, `packages/cli` is green — **1204 passed | 3 skipped across 51 files**, including
`events-tail-cursor.test.ts` and `no-prompt-timing.test.ts`. `tsc --noEmit` clean in every package; `eslint packages`
0 errors (45 pre-existing warnings, none in the changed files).

https://claude.ai/code/session_01V69DFjnRqF69ojUsYnbt24

### DIFF
--- a/packages/cli/src/__tests__/events-command.test.ts
+++ b/packages/cli/src/__tests__/events-command.test.ts
@@ -555,24 +555,27 @@ describe("actana events tail", () => {
     );
     await settle();
 
-    core.emitEvent({ eventId: 1, kind: "session:finished", taskId: "t-old" });
+    core.emitEvent({ eventId: 1, kind: "session:finished", taskId: "t-oldest" });
     core.emitEvent({ eventId: 2, kind: "task:updated" });
-    core.emitEvent({ eventId: 3, kind: "session:finished", taskId: "t-mid" });
+    core.emitEvent({ eventId: 3, kind: "session:finished", taskId: "t-old" });
     core.emitReplayed(3);
     await settle();
     // The ring survives the re-ask: it is one walk, however many tails it takes.
     expect(core.subscribes).toContain(3);
 
-    core.emitEvent({ eventId: 4, kind: "session:finished", taskId: "t-new" });
-    core.emitReplayed(4);
+    core.emitEvent({ eventId: 4, kind: "session:finished", taskId: "t-mid" });
+    core.emitEvent({ eventId: 5, kind: "session:finished", taskId: "t-new" });
+    core.emitReplayed(5);
     await settle();
-    core.emitReplayed(4);
+    core.emitReplayed(5);
 
     const result = await run;
     expect(result.code).toBe(EXIT_OK);
-    // #3 and #4 — in the order the Core appended them, and not #1, which is the
-    // one a run that printed on the way past would have answered with.
-    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([3, 4]);
+    // #4 and #5 — in the order the Core appended them, and not #1, which is the
+    // one a run that printed on the way past would have answered with. Four
+    // matches through a ring of two also turns its write index over twice,
+    // which is where a rotation that read from the wrong slot shows up.
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([4, 5]);
     expect(result.out.map((line) => JSON.parse(line).taskId)).toEqual(["t-mid", "t-new"]);
   });
 
@@ -638,6 +641,124 @@ describe("actana events tail", () => {
     const result = await run;
     expect(result.code).toBe(EXIT_OK);
     expect(JSON.parse(result.out[0]!).eventId).toBe(3);
+  });
+
+  it("does not answer a bounded --kind run from a tip it could only approximate", async () => {
+    // Round-2 review, blocking. The last hiding place of round-1's finding 1.
+    // `tipFrom` hands back its approximation as an ordinary number after
+    // MAX_TIP_ROUNDS, so a walk that took it as the tip flushed its ring and
+    // exited 0 — with whatever matches it happened to see, which on a Core
+    // racing ahead of the reader are the oldest in the log. Everything a script
+    // can observe, exit code and stdout, was then indistinguishable from a
+    // correct answer, and stderr said the run was "carrying on" after it had
+    // exited.
+    //
+    // `event-tip.ts` names this caller by hand: a caller that ends on the
+    // number says so in its own words and decides that an approximation is not
+    // a finished read.
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(
+      ["events", "tail", "--json", "--kind", "session:finished", "--limit", "1"],
+      { connect: core.connect },
+    );
+    await settle();
+
+    // A stale finish first, then a Core that keeps the hunt asking past the
+    // round it gives up on. MAX_TIP_ROUNDS is 200 and lives in `event-tip.ts`.
+    core.emitEvent({ eventId: 1, kind: "session:finished", taskId: "t-stale" });
+    for (let round = 2; round <= 202; round += 1) {
+      core.emitEvent({ eventId: round, kind: "task:updated" });
+      core.emitReplayed(round);
+    }
+
+    const result = await run;
+    // Not a 0, and not the stale finish presented as the answer.
+    expect(
+      result.code,
+      "a bounded --kind run answered from an approximate tip and called it a success",
+    ).toBe(EXIT_FAILURE);
+    const said = result.err.join("\n");
+    expect(said).toContain("appended events faster than they could be read");
+    expect(said).toContain("did not finish");
+    // The sentence written for a run that really is about to follow, which this
+    // one is not: it has exited.
+    expect(said).not.toContain("carrying on");
+    // What it held still goes out — a short read is reported by its status, not
+    // by withholding what it read (#439) — and the sentence counts it.
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([1]);
+    expect(said).toContain("1 of them held from an unfinished walk");
+    expect(core.closed).toBe(true);
+  });
+
+  it("still carries on from an approximate tip when the walk is not the answer", async () => {
+    // The other half of the split, and why the fix above is narrower than
+    // `--limit`. A run with a ceiling but no `--kind` prints nothing on the
+    // walk: its ceiling is spent on live events, so an approximate tip costs it
+    // a short replay rather than its answer. It really is "a first run about to
+    // follow", which is the caller `event-tip.ts` wrote its default sentence
+    // for — so it keeps that sentence and keeps following.
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(["events", "tail", "--json", "--limit", "1"], { connect: core.connect });
+    await settle();
+
+    // The hunt gives up on the round after the last one it will take, so 201
+    // rounds of non-empty tail is where it settles and printing switches on.
+    for (let round = 1; round <= 201; round += 1) {
+      core.emitEvent({ eventId: round, kind: "task:updated" });
+      core.emitReplayed(round);
+    }
+    await settle();
+
+    // Still following, and told why in the words `event-tip.ts` chose.
+    core.emitEvent({ eventId: 202, kind: "session:finished" });
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([202]);
+    expect(result.err.join("\n")).toContain("carrying on from");
+  });
+
+  it("streams --kind matches as they arrive when there is no --limit", async () => {
+    // Round-2 review, should-fix: the `recent === null` branch of the walk had
+    // no test, and it is what the help paragraph advertises. Without a ceiling
+    // there is no n to choose between, every match is going to be printed
+    // anyway, so they go out as they go past rather than into a ring.
+    //
+    // Deliberately never awaited: a tail with no --limit is a follow and has no
+    // ending, which is the whole point of the case. Nothing is armed either —
+    // the 30s deadline belongs to bounded runs — so there is no timer to leak.
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const printed: string[] = [];
+    void cli().run(["events", "tail", "--json", "--kind", "session:finished"], {
+      connect: core.connect,
+      onOut: (line) => printed.push(line),
+    });
+    await settle();
+
+    core.emitEvent({ eventId: 1, kind: "session:finished", taskId: "t-1" });
+    core.emitEvent({ eventId: 2, kind: "task:updated" });
+    core.emitEvent({ eventId: 3, kind: "session:finished", taskId: "t-2" });
+    await settle();
+
+    // Both matches, in order, while the walk is still going — no ring, no wait
+    // for the tip. The unasked-for kind in the middle stays quiet, as ever.
+    expect(printed.map((line) => JSON.parse(line).eventId)).toEqual([1, 3]);
+
+    core.emitReplayed(3);
+    await settle();
+    core.emitReplayed(3);
+    await settle();
+
+    // …and then it follows, which is what a tail with no ceiling does.
+    core.emitEvent({ eventId: 4, kind: "task:updated" });
+    core.emitEvent({ eventId: 5, kind: "session:finished", taskId: "t-3" });
+    await settle();
+    expect(printed.map((line) => JSON.parse(line).eventId)).toEqual([1, 3, 5]);
   });
 
   it("stops timing the Core once it has said where its log ends", async () => {
--- a/packages/cli/src/__tests__/events-command.test.ts
+++ b/packages/cli/src/__tests__/events-command.test.ts
@@ -503,11 +503,7 @@ describe("actana events tail", () => {
     // walking to the tip — and the finish it was typed to find is already in
     // that tail. Counting it and moving on was the defect: nothing on stdout,
     // and then a wait for a second `session:finished` that nobody was going to
-    // produce. On `HEAD~1` this case does not fail an assertion, it hangs.
-    //
-    // No marker is sent here on purpose. The run ends on its ceiling, in the
-    // middle of the walk, which is what "prints one and exits" has to mean for
-    // an operator whose Core has nothing further to say.
+    // produce. On the parent this case does not fail an assertion, it hangs.
     await withRegisteredCore();
     const core = fakeCore({});
 
@@ -521,16 +517,94 @@ describe("actana events tail", () => {
     core.emitEvent({ eventId: 1, kind: "task:created" });
     core.emitEvent({ eventId: 2, kind: "session:finished", taskId: "t-1" });
     core.emitEvent({ eventId: 3, kind: "task:updated" });
+    await settle();
+
+    // Nothing yet, and that is the point of the round-1 review's finding: which
+    // match is the newest is not knowable until the log has an end, so the walk
+    // holds rather than prints. A run that answered here would be answering
+    // with the oldest match the Core still holds.
+    expect(printed).toEqual([]);
+
+    core.emitReplayed(3);
+    await settle();
+    // A receipt, not the tip — the walk carries on and keeps holding.
+    expect(core.subscribes).toContain(3);
+    expect(printed).toEqual([]);
+
+    core.emitReplayed(3);
 
     const result = await run;
     expect(result.code).toBe(EXIT_OK);
     expect(result.out).toHaveLength(1);
     expect(JSON.parse(result.out[0]!)).toMatchObject({ eventId: 2, kind: "session:finished" });
-    // Printed as it went past, not gathered up at the end of the walk: the
-    // cursor advances per delivered event, so a line that waits for the end of
-    // the tail is a line a Ctrl-C can lose.
-    expect(printed).toHaveLength(1);
     expect(core.closed).toBe(true);
+  });
+
+  it("holds the newest --limit matches of the walk, not the first it finds", async () => {
+    // Round-1 review, blocking finding 1, at this suite's level. A first run
+    // walks the whole retained log, so "the first match" and "the match the
+    // operator just watched happen" are the same event only on a Core with no
+    // history. Three finishes go past, spread over two rounds of a capped
+    // replay, and `--limit 2` must answer with the last two.
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(
+      ["events", "tail", "--json", "--kind", "session:finished", "--limit", "2"],
+      { connect: core.connect },
+    );
+    await settle();
+
+    core.emitEvent({ eventId: 1, kind: "session:finished", taskId: "t-old" });
+    core.emitEvent({ eventId: 2, kind: "task:updated" });
+    core.emitEvent({ eventId: 3, kind: "session:finished", taskId: "t-mid" });
+    core.emitReplayed(3);
+    await settle();
+    // The ring survives the re-ask: it is one walk, however many tails it takes.
+    expect(core.subscribes).toContain(3);
+
+    core.emitEvent({ eventId: 4, kind: "session:finished", taskId: "t-new" });
+    core.emitReplayed(4);
+    await settle();
+    core.emitReplayed(4);
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    // #3 and #4 — in the order the Core appended them, and not #1, which is the
+    // one a run that printed on the way past would have answered with.
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([3, 4]);
+    expect(result.out.map((line) => JSON.parse(line).taskId)).toEqual(["t-mid", "t-new"]);
+  });
+
+  it("hands over what the walk was holding when the Core stops answering", async () => {
+    // The deadline is this side giving up, and #439 settled what that looks
+    // like: what was read goes to stdout, the reason to stderr, and the exit is
+    // non-zero because a read cut off partway is not a read that finished. A
+    // walk that is holding matches has read them — the cursor has already moved
+    // past them — so dropping them on the way out would lose them for every
+    // later run to buy nothing.
+    vi.useFakeTimers();
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const run = cli().run(
+      ["events", "tail", "--json", "--kind", "session:finished", "--limit", "3"],
+      { connect: core.connect },
+    );
+    await vi.advanceTimersByTimeAsync(0);
+
+    core.emitEvent({ eventId: 1, kind: "session:finished" });
+    core.emitEvent({ eventId: 2, kind: "task:updated" });
+    // …and then nothing: no marker, no further event, no end of the log.
+    await vi.advanceTimersByTimeAsync(SUBSCRIBE_ANSWER_MS + 1);
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_FAILURE);
+    expect(result.out.map((line) => JSON.parse(line).eventId)).toEqual([1]);
+    const said = result.err.join("\n");
+    expect(said).toContain("stopped answering");
+    expect(said).toContain("held from an unfinished walk");
+    expect(said).toContain("did not finish");
   });
 
   it("still prints none of the kinds a first run was not asked for", async () => {
--- a/packages/cli/src/__tests__/events-command.test.ts
+++ b/packages/cli/src/__tests__/events-command.test.ts
@@ -474,8 +474,9 @@ describe("actana events tail", () => {
     // operator asked for still finished, and has nothing left to wait for —
     // reading `printed` here would hang exactly where #402 hung, one flag over.
     //
-    // The first-run half of the kind filter — history skipped while the cursor
-    // advances past it — is #403 and is not touched here.
+    // This is the run *with* a cursor. The first-run half of the filter — where
+    // a named kind is printed out of the walk to the tip — is #403 and is the
+    // two cases below.
     await withRegisteredCore();
     const core = fakeCore({});
 
@@ -495,6 +496,74 @@ describe("actana events tail", () => {
     expect(result.code).toBe(EXIT_OK);
     expect(result.out).toEqual([]);
     expect(core.closed).toBe(true);
+  });
+
+  it("prints a --kind match out of a first run's tail, and ends on it", async () => {
+    // #403, frame for frame. No stored cursor and no --since, so this run is
+    // walking to the tip — and the finish it was typed to find is already in
+    // that tail. Counting it and moving on was the defect: nothing on stdout,
+    // and then a wait for a second `session:finished` that nobody was going to
+    // produce. On `HEAD~1` this case does not fail an assertion, it hangs.
+    //
+    // No marker is sent here on purpose. The run ends on its ceiling, in the
+    // middle of the walk, which is what "prints one and exits" has to mean for
+    // an operator whose Core has nothing further to say.
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const printed: string[] = [];
+    const run = cli().run(
+      ["events", "tail", "--json", "--kind", "session:finished", "--limit", "1"],
+      { connect: core.connect, onOut: (line) => printed.push(line) },
+    );
+    await settle();
+
+    core.emitEvent({ eventId: 1, kind: "task:created" });
+    core.emitEvent({ eventId: 2, kind: "session:finished", taskId: "t-1" });
+    core.emitEvent({ eventId: 3, kind: "task:updated" });
+
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    expect(result.out).toHaveLength(1);
+    expect(JSON.parse(result.out[0]!)).toMatchObject({ eventId: 2, kind: "session:finished" });
+    // Printed as it went past, not gathered up at the end of the walk: the
+    // cursor advances per delivered event, so a line that waits for the end of
+    // the tail is a line a Ctrl-C can lose.
+    expect(printed).toHaveLength(1);
+    expect(core.closed).toBe(true);
+  });
+
+  it("still prints none of the kinds a first run was not asked for", async () => {
+    // The guard on #403's narrowing. The replay storm is a history nobody asked
+    // for, and it is still suppressed exactly as it was — what changed is that
+    // naming a kind counts as asking for it. Everything else in the tail stays
+    // as quiet on a first run as it has always been.
+    await withRegisteredCore();
+    const core = fakeCore({});
+
+    const printed: string[] = [];
+    const run = cli().run(
+      ["events", "tail", "--json", "--kind", "session:finished", "--limit", "1"],
+      { connect: core.connect, onOut: (line) => printed.push(line) },
+    );
+    await settle();
+
+    core.emitEvent({ eventId: 1, kind: "task:created" });
+    core.emitEvent({ eventId: 2, kind: "task:updated" });
+    core.emitReplayed(2);
+    await settle();
+    // A receipt, not the tip: the walk carries on from #2 (`event-tip.ts`).
+    expect(core.subscribes).toContain(2);
+    core.emitReplayed(2);
+    await settle();
+
+    // Two events of history, no finish among them, and nothing on stdout.
+    expect(printed).toEqual([]);
+
+    core.emitEvent({ eventId: 3, kind: "session:finished" });
+    const result = await run;
+    expect(result.code).toBe(EXIT_OK);
+    expect(JSON.parse(result.out[0]!).eventId).toBe(3);
   });
 
   it("stops timing the Core once it has said where its log ends", async () => {
--- a/packages/cli/src/__tests__/events-tail-cursor.test.ts
+++ b/packages/cli/src/__tests__/events-tail-cursor.test.ts
@@ -292,6 +292,94 @@ describe("actana events tail, across a Core restart", () => {
     expect(log.tailReads).toBeGreaterThan(1);
   }, 60_000);
 
+  it("prints the finish already in the log on a first --kind run, and exits", async () => {
+    // #403's first criterion, against the Core that produces it and the cursor
+    // file a real run writes. This is the command an operator types after a
+    // session has finished, on a machine that has never followed this Core: no
+    // stored cursor, no --since, and the answer already in the log. It used to
+    // print nothing and never exit.
+    const log = arrayEventLog();
+    const port = await freePort();
+    const core = await coreOn(port, log);
+    fixture = makeCliFixture();
+    registerCore(fixture.paths, "inproc", core.blobText);
+
+    log.push("session:started");
+    log.push("task:created");
+    log.push("session:finished");
+    log.push("task:updated");
+
+    const result = await fixture.run(
+      ["events", "tail", "--json", "--kind", "session:finished", "--limit", "1"],
+      { connect: connectCore },
+    );
+
+    expect(result.code, result.err.join("\n")).toBe(EXIT_OK);
+    // One event, and the one that was asked for. The three around it are the
+    // history nobody asked for and are still not printed.
+    expect(idsOf(result.out)).toEqual([3]);
+  }, 60_000);
+
+  it("never lets the cursor pass a --kind match the operator was not shown", async () => {
+    // #403's second criterion: Ctrl-C must not persist a cursor past unseen
+    // matching events. The old walk counted the finish, let the durable client
+    // advance and persist the cursor past it, and printed nothing — so the
+    // operator killing the hung command lost that event for every later run
+    // too, silently and with nothing in the log to notice.
+    //
+    // There is no signal handler to raise here (`events tail` takes the default
+    // disposition on purpose), so the interrupt is modelled where it actually
+    // bites: the cursor on disk is sampled while the run is still walking, and
+    // whatever moment that sample lands in is a moment a Ctrl-C could have
+    // landed in. The finish must already be on stdout by then.
+    const log = arrayEventLog();
+    for (let i = 0; i < 21; i += 1) log.push("task:updated"); // #1 – #21
+    log.push("session:finished"); // #22
+    for (let i = 0; i < 8; i += 1) log.push("task:updated"); // #23 – #30
+
+    const port = await freePort();
+    const core = await coreOn(port, log);
+    fixture = makeCliFixture();
+    registerCore(fixture.paths, "inproc", core.blobText);
+
+    const printed: string[] = [];
+    // `--limit 2` against a log holding one finish: the ceiling cannot be
+    // reached on history, so the run is still going at the moment the sample
+    // below is taken — which is what makes it an interrupt rather than an exit.
+    const tail = fixture.run(
+      ["events", "tail", "--json", "--kind", "session:finished", "--limit", "2"],
+      { connect: connectCore, onOut: (line) => printed.push(line) },
+    );
+
+    const dir = cursorsDir(fixture.paths);
+    const cursor = (): number => {
+      if (!existsSync(dir)) return 0;
+      const [file] = readdirSync(dir);
+      if (file === undefined) return 0;
+      const raw = readFileSync(path.join(dir, file), "utf8").trim();
+      return raw === "" ? 0 : Number(raw);
+    };
+
+    await waitFor(() => cursor() >= 22, "the cursor never reached the finish");
+    expect(
+      idsOf(printed),
+      "the cursor stood past a session:finished that had never been printed",
+    ).toContain(22);
+
+    // And it does walk on past it — which is exactly why the printing has to
+    // happen on the way. The finish is behind the cursor now and unreachable to
+    // every later run; the only reason it is not lost is that it has been seen.
+    await waitFor(() => cursor() >= 30, "the walk never reached the end of the history");
+    expect(idsOf(printed)).toEqual([22]);
+
+    // A second finish, live, so the run ends on its ceiling rather than
+    // following for the rest of the suite.
+    log.push("session:finished");
+    const result = await tail;
+    expect(result.code, result.err.join("\n")).toBe(EXIT_OK);
+    expect(idsOf(result.out)).toEqual([22, 31]);
+  }, 60_000);
+
   it("leaves the stored cursor alone when --since asks for a one-off rewind", async () => {
     const log = arrayEventLog();
     const port = await freePort();
--- a/packages/cli/src/__tests__/events-tail-cursor.test.ts
+++ b/packages/cli/src/__tests__/events-tail-cursor.test.ts
@@ -320,21 +320,20 @@ describe("actana events tail, across a Core restart", () => {
     expect(idsOf(result.out)).toEqual([3]);
   }, 60_000);
 
-  it("never lets the cursor pass a --kind match the operator was not shown", async () => {
-    // #403's second criterion: Ctrl-C must not persist a cursor past unseen
-    // matching events. The old walk counted the finish, let the durable client
-    // advance and persist the cursor past it, and printed nothing — so the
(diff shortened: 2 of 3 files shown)


qcentic-dev:
## Acceptance checklist — #403

Every criterion the issue states, each checked or explicitly left unchecked with the reason.

- [x] **1. After a finish is in the log, `actana events tail --kind session:finished --limit 1` prints one
  `session:finished` and exits.**
  A first run's walk to the tip now prints the kinds `--kind` named, instead of counting them and moving on, and the
  `--limit` ceiling ends the run on that line mid-walk. Proven twice: against the real in-process Core with a real
  `wss://` link and a real cursor file — *"prints the finish already in the log on a first `--kind` run, and exits"*,
  one event on stdout, id `#3`, `EXIT_OK` — and frame for frame against a fake Core with **no marker sent at all**,
  *"prints a `--kind` match out of a first run's tail, and ends on it"*. Both hang on the parent commit `31444a9`,
  which is the defect.

- [x] **2. Ctrl-C does not persist a cursor past unseen matching events.**
  Matches are printed as they go past rather than gathered up until the tip is known, so the cursor — which
  `DurableCoreClient.deliverEvent` advances and persists per event delivered — can never come to rest beyond a match
  that was never shown. `events tail` registers no signal handler (deliberate, and unchanged), so the interrupt is
  modelled where it actually bites: *"never lets the cursor pass a `--kind` match the operator was not shown"* runs a
  30-event log with the finish at `#22` under `--limit 2`, so the ceiling cannot be reached on history and the run is
  still going at the moment a Ctrl-C would land. The cursor file is sampled mid-walk, and whatever moment that sample
  lands in is a moment an interrupt could have landed in: `#22` must already be on stdout by then. On the parent this
  fails with `expected [] to include 22` — the cursor standing past a `session:finished` that had never been printed.

### Also asserted, not asked for

- [x] **The narrowing is guarded.** A first run given `--kind session:finished` still prints none of the `task:*`
  history it walks past — through a capped replay and its re-ask — and only speaks when the finish arrives. The replay
  storm is a history nobody asked for; naming kinds is the asking, and nothing else changed.
- [x] **A first run without `--kind` is untouched.** Still starts at the end of the log, `tail -f` and not `cat`, held
  to it by the pre-existing cases in `events-command.test.ts` and `events-tail-cursor.test.ts`.

### Explicitly out of scope, and why

- [ ] **#441 — an explicit `--follow` / `--no-follow`.** Not taken, as instructed. This change adds no flag and does
  not touch what `--limit` means. The rule it does add is decided by a flag the operator typed rather than by the state
  of the log, so it does not add to the case #441 is about.
- [ ] **#442 — a `--limit` ceiling advancing the stored cursor past the events it printed.** Not taken, as instructed.
  Pre-existing (reproduced against `a7b5532`) and untouched: `finish` still closes the client without rewinding what
  the durable client already wrote, so a bounded run's cursor can still land at the end of a synchronous burst. That is
  a bug about events the *ceiling* stopped; criterion 2 above is about a match the *walk* skipped, and the test is
  written to the criterion #403 states.
- [ ] **The cursor persistence rule.** Not touched at all — no change to `DurableCoreClient`, none to `deliverEvent`,
  none to the `finish(EXIT_OK)` the ceiling takes. The fix needed none, so it made none.

### Verification

- `packages/cli` suite: **1200 passed | 3 skipped across 51 files**, three consecutive full runs, including
  `events-tail-cursor.test.ts` (real Core, real drop, real restart) and `no-prompt-timing.test.ts`.
- `pnpm typecheck` clean; `pnpm lint` 0 errors (45 warnings, all pre-existing and in other packages).
- `commitlint` run locally over `31444a9..HEAD` with the repo's own `commitlint.config.mjs`, and over the PR title:
  `found 0 problems, 0 warnings` for both.
- Diff is three files, all under `packages/cli`. No `packages/core`, `packages/shared` or `packages/panel` file is
  touched.

https://claude.ai/code/session_01V69DFjnRqF69ojUsYnbt24

mrfqcentic (changes requested):
**Requesting changes.** Scope, hygiene and the #402 guards are right; two properties of the new first-run printing path are not, and one of them makes the command answer the wrong question with exit 0.

Reviewed at head `6c8474f` in a detached worktree, against `origin/fix/operator-pins-sessions-drop-cli` (merge base `31444a9`). `packages/cli`: **1200 passed | 3 skipped** across 51 files. `pnpm typecheck` clean; `pnpm lint` 0 errors / 45 pre-existing warnings, none in the changed files. Three of the four new cases genuinely fail on the parent — two by timing out, one on `expected [] to include 22` — reproduced exactly. CI on `6c8474f`: 18 green, 6 skipped.

## Blocking

### 1 — `packages/cli/src/events-command.ts:393` (with `:364`): a first `--kind … --limit n` run prints the *oldest* matches in the retained log, not the ones the operator just produced

A first run has no stored cursor, so `DurableCoreClient` subscribes from `0` and the walk to the tip covers the **whole retained log** — nothing prunes `event_log_store`, as `event-tip.ts`'s own header says. Printing matches *as they go past* therefore hands back the earliest match, and `emit`'s ceiling ends the run on it.

Measured against the real in-process Core (the `events-tail-cursor.test.ts` harness), log = `task:updated`×2, **`session:finished` #3**, `task:updated`×18, **`session:finished` #22**, `task:updated`×8:

```
actana events tail --json --kind session:finished --limit 1
→ exit 0, printed eventId=3, stored cursor=30
```

`#22` is the finish the operator just watched happen. `#3` may be from weeks ago and carries a different `taskId`. The command exits 0 with a confidently wrong answer — which for the script this criterion exists to serve is worse than the hang, because the hang was visible. It is also the exact hazard `event-tip.ts` argues at length for `harness install`: *"an install that failed an hour ago is still in the log and must not be read as this one's verdict."*

Both new tests miss it because each of their logs holds exactly one matching event.

It also undercuts the PR body's claim that *"the same command means the same thing on every Core and on every day"* — which event comes back depends entirely on how much history that Core has kept.

I don't think this has a free fix, and that is worth saying rather than pretending otherwise: printing on the way past is precisely what buys criterion 2, while printing the *newest* match needs the tip first, which needs the walk to finish. A bounded ring of the last `n` matches, flushed when the tip is known, would give the right answer at bounded memory but re-opens the interrupt window this PR closes — which is really #442's cursor rewind, i.e. the two want deciding together. Either shape is fine by me; the current behaviour should not ship as the answer to #403.

### 2 — `packages/cli/src/events-command.ts:382–392`: the cursor invariant this comment claims is still violated — it has moved from Ctrl-C to the ordinary exit

The comment asserts *"the cursor cannot come to rest beyond a match the operator was never shown."* In the run above the cursor file held **`30`** while only `#3` had been printed: `session:finished` `#22` is now behind the stored cursor and unreachable to every later run, silently, with nothing in the log to notice. `emit` → `finish` unregisters `offEvent` and closes the client, but frames already in flight still reach `DurableCoreClient.deliverEvent`, which `persistCursor()`s before any listener runs.

That mechanism is #442, and I agree it must not be fixed in this PR. What I don't accept is the framing. With the hang gone, **Ctrl-C is no longer how this run ends** — it ends on the ceiling — so #403's criterion 2 has not been satisfied so much as relocated to the path the fix newly made reachable, and the operator-visible harm is identical.

`events-tail-cursor.test.ts:307` is shaped so it cannot see this: `--limit 2` against a log holding one match, chosen so the run never exits during the walk. The interrupt case it does cover is real and does fail on the parent; it just isn't the whole criterion any more.

At minimum, before this lands: soften the comment at `:382–392` to what actually holds, add the exit-path case to #442 with these numbers, and don't close #403 on this PR alone.

## Should fix

### 3 — `packages/cli/src/events-command.ts:152` and `:391`: `--kind` with no `--limit` replays the entire history for that kind, unbounded

`emit` in the `!printing` branch is capped only by `--limit`. `actana events tail --kind task:updated`, on a machine that has never followed a Core that has been up for weeks, now cats every historical `task:updated` before it starts following — the replay storm the module header says the silence exists to prevent, merely filtered. The help two paragraphs above still promises *"like `tail -f`, not like `cat`"*. If the narrowing stays, that cost belongs in the help text next to the new paragraph at `:152`.

## What is right

- **Scope is exactly right.** Three files, all `packages/cli`. No `packages/core`, `packages/shared`, `packages/panel` or `packages/sdk` change. No new flag, so #441's `--follow`/`--no-follow` design is untaken. `DurableCoreClient` and the ceiling's `finish` are untouched, so #442 is untaken.
- **#402 is not regressed.** The 30s deadline, its disarm at the log's end, the `onDisconnected` clear / `onReady` re-arm, and the "exits from history" and approximation guards are byte-identical and all pass.
- **The parent-failure claim is honest** and I reproduced it verbatim.
- Base is `fix/operator-pins-sessions-drop-cli`, still a draft, both commits pass commitlint (longest body line 121 against the 132 limit), PR title is conventional.
- Not a defect — an observation in the fix's favour: across a **capped** replay the walk delivers matching history exactly once and in order. 1500 events, finishes at `#500`/`#1200`/`#1450`, `--kind session:finished --limit 3` → `[500, 1200, 1450]`, exit 0. No real-Core test covers the capped path for `--kind`; that case is worth adding whichever way finding 1 is resolved.

Reviewed at `6c8474f`.

qcentic-dev:
Thanks — all three findings taken, and finding 1 was right to block. Pushed as `3f1b3bd` (plain push; no force push at
any point on this branch).

## 1 — the oldest match, blocking · **fixed**

Taken as you framed it, and with the shape you suggested: **a bounded ring of the last `n` matches, flushed when the
tip is known.** The walk holds `n` and no more, whatever the log's length, so memory is bounded and the answer is the
newest `n` rather than the first `n` found.

Your log, verbatim, through the `events-tail-cursor.test.ts` harness:

```
log: task:updated×2, session:finished #3 (t-old), task:updated×18, session:finished #22 (t-now), task:updated×8
actana events tail --json --kind session:finished --limit 1
  6c8474f → exit 0, printed [3]     ← the confident wrong answer
  3f1b3bd → exit 0, printed [22]    ← the finish the operator watched
```

You were right that this is not free, and the two things it costs are handled rather than hidden:

- **A run can no longer exit mid-walk on the ceiling.** It always reads to the end of the log first, which is what
  "newest" requires. The 30s subscribe deadline already covers a Core that never gets it there.
- **A Core that goes quiet mid-walk would strand the ring.** So the deadline now hands over what it is holding before it
  gives up — those matches are past the cursor and the cursor has already moved over them, so dropping them would lose
  them for every later run to buy nothing. stdout gets them, stderr names the count (`held from an unfinished walk`),
  and the exit stays non-zero, as #439 settled.
- **Without `--limit` there is no `n` to choose between**, so matches still stream as they arrive: every one of them is
  going to be printed anyway, and holding them would buy nothing and cost an unbounded array.

## 2 — the cursor invariant, blocking · **all three asks done**

You are right, and the framing was the problem rather than the mechanism. Three things, as you listed them:

- **The comment is gone**, replaced by what actually holds: the cursor outruns stdout on the ordinary exit path too,
  because `DurableCoreClient.deliverEvent` persists before any listener runs and frames already in flight keep arriving
  after `finish` closes the client. The module header and the branch comment now say that #442 is not fixed here and
  that #403 stays open until it lands. The ring makes this window *wider*, and that is stated in both places rather
  than left for the next reader to discover.
- **The exit-path case is filed on #442** with the measurement, before and after:
  https://github.com/actana/control/issues/442#issuecomment-5516724486 — `6c8474f` printed `[3]` with `storedCursor=30`,
  `3f1b3bd` prints `[22]` with `storedCursor=30`. Both exit 0, both leave a `session:finished` that was never shown
  behind the cursor. Which one is lost changes; that one is lost does not.
- **The PR body no longer closes #403.** It is `Refs #403` now, with a note at the top saying criterion 2 is not
  satisfied here, that it cannot be from `events-command.ts` alone, and that #403 should be closed once #442 lands.

`events-tail-cursor.test.ts`'s interrupt case is **removed**, not re-shaped. With the ring it asserted something no
longer true, and the criterion it covered now belongs to #442 — leaving a green test whose name still claimed the
guarantee would have been the same mistake as the comment.

## 3 — unbounded `--kind` history, should-fix · **stated, not bounded**

Chose the help text. An implicit cap would be a number nobody asked for, silently dropping matching history — which is
the class of failure this whole ticket is about, and an operator cannot tell a truncated answer from a complete one.
`--limit` is the bound, and after finding 1 it selects the *newest* `n`, which is the bound worth having. So the help
now says it plainly, next to the new paragraph:

```
  With --limit n that is the newest n of them, not the first n found: the run
  reads to the end of the log before it can know which are newest, so it holds
  the last n it saw and prints them once it does.

  Without --limit it is all of them — every event of those kinds the Core still
  holds, which on a Core that has been up for weeks is not a short list. --kind
  bounds what gets printed; --limit is what bounds how much.
```

Happy to swap this for a cap if you would rather have one — it is a one-line change and I have no strong view beyond
preferring a loud cost to a quiet truncation.

## The capped-replay test you asked for

Added, with your numbers — 1500 events, finishes at `#500`/`#1200`/`#1450`, `--kind session:finished --limit 3` →
`[500, 1200, 1450]`, exit 0, `tailReads > 1`. Plus a second one on the same log at `--limit 2` → `[1200, 1450]`, which
is where a ring that reset per tail, or kept the first `n` it met, gives `[500, 1200]` instead. It does exactly that on
`6c8474f`.

## Verification

Five cases fail on the reviewed head `6c8474f` and pass now:

```
× answers a bounded first --kind run with the newest match, not the oldest   expected [ 3 ] to equal [ 22 ]
× keeps only the newest --limit matches when the walk crosses the cap        expected [ 500, 1200 ] to equal [ 1200, 1450 ]
× holds the newest --limit matches of the walk, not the first it finds       expected [] to include 3
× prints a --kind match out of a first run's tail, and ends on it            expected [ Array(1) ] to equal []
× hands over what the walk was holding when the Core stops answering         missing "held from an unfinished walk"
```

`packages/cli`: **1204 passed | 3 skipped** across 51 files, `events-tail-cursor.test.ts` and `no-prompt-timing.test.ts`
included. `tsc --noEmit` clean in every package; `eslint packages` 0 errors, 45 pre-existing warnings, none in the
changed files. All three commits pass commitlint over `31444a9..HEAD` with the repo's own config; longest body line 119
against the 132 limit.

Scope is unchanged from round 1 — three files, all `packages/cli`. No `packages/core`, `packages/shared`,
`packages/panel` or `packages/sdk` change, and no `session-command.ts` / `session-gateway.ts`. Base is still
`fix/operator-pins-sessions-drop-cli`, still a draft.

https://claude.ai/code/session_01V69DFjnRqF69ojUsYnbt24

mrfqcentic (changes requested):
**Requesting changes — round 2.** All three round-1 findings are genuinely taken, and the ring is the right shape: I traced it and it holds. One blocking thing remains, and it is the *residual* of round-1 finding 1 rather than a new one — on the `MAX_TIP_ROUNDS` path a bounded first `--kind` run still exits **0** with the **oldest** match, and `event-tip.ts`'s own doc comment names that caller by hand as the one that must not use the default.

Reviewed at head `3f1b3bd`, against `origin/fix/operator-pins-sessions-drop-cli` (merge base `31444a9`), in a read-only clone.

## Verification, done rather than taken on trust

- **The five claimed parent-failures reproduce verbatim.** New tests on `6c8474f`'s `events-command.ts`: `expected [ 3 ] to deeply equal [ 22 ]`, `expected [ 500, 1200 ] to deeply equal [ 1200, 1450 ]`, `expected [] to include 3`, `expected [ Array(1) ] to deeply equal []`, and the missing `held from an unfinished walk`. 5 failed | 27 passed.
- **The defect itself is real on the merge base.** `prints the finish already in the log on a first --kind run, and exits` against `31444a9` → `Error: Test timed out in 60000ms`. That is the hang #403 reports.
- **`packages/cli`: 1204 passed | 3 skipped across 51 files** — your number exactly. (First attempt here showed 42 failures across 8 files; that was this machine over-subscribed at the default worker count, not the branch. Constrained to two workers it is clean, and CI's `Unit Tests` is green on `3f1b3bd` — all 18 checks pass, 6 skipped.)
- **The #442 comment exists and is honest** (`#442-issuecomment-5516724486`), including that the ring *widens* the window rather than only relocating it.

**1. "After a finish is in the log, that command prints one `session:finished` and exits." — MET.**

The enforcing lines, in order: `events-command.ts:491` takes a named kind out of the walk; `:496-497` holds it; `:534` flushes at the tip; `:439-442` prints and takes the ceiling to `finish(EXIT_OK)`. Proven against the real in-process Core at `events-tail-cursor.test.ts:295` (one line, `#3`, `EXIT_OK`) and refined at `:323` to the reviewer's own log — `[22]`, `t-now`, not `[3]`. It hangs on `31444a9`, verified above.

**2. "Ctrl-C does not persist a cursor past unseen matching events." — NOT MET, and correctly declared not met.**

`durable-core-client.ts:230-235` is dispositive: `deliverEvent` sets `lastEventId` and calls `persistCursor()` *before* `super.deliverEvent` reaches any listener. Nothing in `events-command.ts` runs early enough to hold that back, so a held match is by construction one the cursor has already passed. The PR says so at the top, the header says so at `:53-57`, the branch comment says so at `:481-490`, and the case is filed on #442 with measurements. Not satisfiable from this file; not claimed.

**Round-1 asks, individually:**

- **Finding 1 (blocking) — mostly fixed, see below.** Bounded runs now answer newest, on every path except an approximate tip.
- **Finding 2 (blocking) — done, all three parts.** The false invariant comment is gone; what replaced it at `:481-490` is accurate and states the cost. `Fixes #403` → `Refs #403`. The exit-path case is on #442 with numbers.
- **Finding 3 (should-fix) — done as I specified.** I asked for the cost in the help text if the narrowing stayed; `:173-183` says it, including "`--kind` bounds what gets printed; `--limit` is what bounds how much." Declining the cap is the right call and the reasoning is right: a silent truncation is the failure class this ticket is about.

**The removed test — honestly replaced, not quietly dropped.** `never lets the cursor pass a --kind match the operator was not shown` is gone from `events-tail-cursor.test.ts`, and it had to be: under the ring it asserted something false. Its removal is stated in the PR body, in `3f1b3bd`'s message, and its subject matter is transferred to #442. Leaving it green would have been the same mistake as the comment. This is the right disposition.

## Blocking

### `packages/cli/src/events-command.ts:312` (with `:530-534`): on an approximate tip, a bounded first `--kind` run still prints the *oldest* match and exits 0

`const tip = trackEventTip(client, deps)` is built with no options, unlike `history` at `:317-326`. When the hunt gives up after `MAX_TIP_ROUNDS`, `tipFrom` returns the approximation as a **non-null** number (`event-tip.ts:139-152`), so `:530-534` takes it as the end of the log unconditionally: `stopTiming(); printing = true; flushRecent();` — and `flushRecent` → `emit` → `finish(EXIT_OK)`.

Reproduced against `fakeCore`, first run, no cursor, `--kind session:finished --limit 1`, a stale finish at `#1` and a Core still appending past `#202`:

```
exit 0
stdout: eventId 1   (taskId t-stale)
stderr: this Core appended events faster than they could be read (200 rounds);
        carrying on from #202, which may be behind its log's end.
```

Two things wrong at once, and `event-tip.ts:91-106` predicts both by name:

> "The default sentence says the caller is carrying on from an approximate number, and for the caller it was written for — **a first run about to follow** — that is exactly what happens. `events tail --limit` on the printing side is not: it takes the returned number as the end of the history and stops there, so 'carrying on' would be told to an operator whose command has already exited (#402 review, finding 5). **A caller that ends here says so in its own words, and gets to decide that an approximation is not a finished read.**"

Before this PR the walk side *was* "a first run about to follow", so the default was true for it. This PR makes it a caller that can end here — and it kept the default. So the run tells a script "carrying on" as it exits, and hands it the stale finish with a 0.

This is the same failure your round-1 finding 1 blocked on, in its last remaining hiding place, and against your own standard for it: `:335-338` of this PR says answering with a stale finish at exit 0 "is worse than the hang #403 reports, because the hang was visible." Everything a script sees here — exit code and stdout — is indistinguishable from a correct answer.

I'll say plainly that this is far narrower in reach than round 1's version: it needs a Core out-appending the client for 200 consecutive rounds, not merely a Core with history. I'm still blocking, for three reasons: the wrong answer is silent to the consumer this criterion exists to serve; the file already contains the pattern that fixes it, five lines away; and `event-tip.ts` documents this exact caller as the one that must not use the default.

The fix mirrors `:317-326` — give `tip` an `onRoundsExhausted` that sets `approximate`, and in the `!printing` branch end the way the printing side already ends at `:555-562`: its own stderr sentence and `EXIT_FAILURE`, rather than flushing to a 0. Note the two shapes it needs to keep apart: a run **with** `--limit` is ending and should fail; a run **without** one really is about to follow, and the default sentence stays true for it. If you'd rather argue the walk side should deliberately differ, that's a fine answer too — but then the divergence and the stderr wording want saying in a comment, because `event-tip.ts` currently reads as forbidding it.

## Should fix

### `packages/cli/src/__tests__/events-command.test.ts`: no case covers `--kind` with no `--limit` on a first run

Every `--kind` case in both suites passes `--limit` (`events-command.test.ts:177, 484, 512, 553, 591, 620`; `events-tail-cursor.test.ts:313, 345, 373, 400`). The `recent === null` branch inside the walk at `:492-494` — stream as they arrive, no ring — is new, is the behaviour the new help paragraph advertises, and is untested.

I probed it and **the behaviour is correct**: a first run with `--kind session:finished` and no `--limit` printed `#1` mid-walk and `#3` after it, both before the tip. So this is a missing guard, not a defect. Given finding 3 was answered with words rather than a cap, one case pinning the words to the code is cheap.

## Nits, non-blocking

- **`:497` — `recent.shift()` is O(n) per match once the ring is full.** Fine at `--limit 1`; `--kind task:updated --limit 5000` over a long retained log walks quadratically. A write index rotated on read is O(1) at the same bounded memory. Only worth doing if it's free.
- **Squash would carry a paragraph the branch itself reverses.** This repo's squash message mode is `COMMIT_MESSAGES`, so a squash concatenates all three bodies — including `d7bfb7d`'s "Printing on the way past is also what keeps the cursor honest… an interrupt cannot leave the cursor beyond a match that was never shown", which `3f1b3bd` exists to retract. A merge or rebase keeps the correction adjacent to the claim; a squash puts both in one message with no ordering cue.

## What is right

- **The ring holds.** `:496-497` bounds it at `recentCap` per match, so it is `n` and no more whatever the log's length — verified by reading, and by the 1500-event capped-replay cases at `events-tail-cursor.test.ts:355` and `:384`, where a ring that reset per tail or kept the first `n` gives `[500, 1200]` instead of `[1200, 1450]`.
- **`flushRecent` fires when the tip is known and not before.** `:523-526` returns on a capped replay without flushing; `:534` is reached only after `tip.tipFrom` returned non-null. `events-command.test.ts:501` asserts the silence frame for frame across a re-ask.
- **No path drops the ring silently.** I enumerated every `finish` in the function: `:408` (deadline, flushes first), `:441` (ceiling, reachable with a live ring only via `flushRecent` itself), `:560`/`:574` (both behind `history !== null`, i.e. `fromStart`, where `recent` is null by `:349`), `:604` (`--limit 0`, before any frame). The ring is either flushed or handed over, never abandoned.
- **The deadline hand-over is right, and its message is right.** `:397-402` writes what is held before `finish(EXIT_FAILURE)` at `:408`; `held` is sampled before the writes and `printed` after, so "`N` event(s) printed, `H` of them held from an unfinished walk" is arithmetically true. Using `write` rather than `emit` there is safe rather than sloppy: on a live ring `printed` is provably 0, so `H ≤ n` cannot exceed the ceiling — worth the sentence it gets at `:392-396`.
- **`--limit`-less streaming works** (probed, above), and `--since` is untouched: `:349` gates the ring on `!fromStart`, so a run with a cursor behaves exactly as it did.
- **Scope is unchanged and still exactly right.** Three files, all `packages/cli`. No `packages/core`, `packages/shared`, `packages/panel` or `packages/sdk`, no `session-command.ts`, no `session-gateway.ts`. `DurableCoreClient` and the ceiling's `finish` untouched, so #442 stays untaken; no new flag, so #441 stays untaken.
- Base is still `fix/operator-pins-sessions-drop-cli` and still a draft. Three commits, all `Refs #403`, longest body line 122 against the 132 limit, subjects 56/74/68 against 120. `Conventions` green.

## On `Refs #403` rather than `Fixes`

**Acceptable, and the right call for this PR** — this is not the thing holding it. Criterion 2's mechanism is `durable-core-client.ts:230-235`, in another package, and round 1 said #442 must not be fixed here; a `Fixes` would have closed a ticket on a criterion this diff provably cannot reach. Relative to `31444a9` nothing about criterion 2 gets worse in kind — the base already left an unshown match behind the cursor, and it left the operator hanging as well. What the ring changes is *which* match is lost and how wide the window is, and both are stated in the code, in the PR body and on #442 rather than discovered later. Merging onto a draft feature branch with #403 open and #442 carrying the remainder is the honest shape.

## Checklist

- [x] Criterion 1 — met; enforcing lines read, real-Core test verified, hang reproduced on the merge base
- [ ] Criterion 2 — **not met**, correctly declared, transferred to #442 with measurements
- [x] Round-1 finding 1 — fixed on every path but the approximate tip (**blocking**, above)
- [x] Round-1 finding 2 — comment softened, `Refs` not `Fixes`, #442 case filed
- [x] Round-1 finding 3 — cost stated in the help text, as specified
- [x] Ring bounded at `n` regardless of log length
- [x] `flushRecent` fires at a known tip and not before
- [x] 30s deadline hands held matches to stdout with a non-zero exit
- [x] Without `--limit`, matches still stream (verified by probe; **untested in the suite**)
- [x] Removed test honestly replaced by the `Refs`-not-`Fixes` framing
- [x] Scope, hygiene, commitlint, CI green at `3f1b3bd`
- [ ] `--kind` with no `--limit` on a first run has no test (should fix)

**Verdict: request changes.** One blocking finding — `events-command.ts:312`/`:530-534`, the approximate-tip path still answering a bounded `--kind` run with the oldest match at exit 0 — plus one should-fix test gap. Everything else asked for in round 1 is done, and done honestly; fix that one path and I expect to approve.

https://claude.ai/code/session_016DHCEsQrDGFU1oMQb8Urhv

qcentic-dev:
Thanks — the blocking finding is right and is fixed, and the should-fix test gap is closed. Pushed as `89383a0` (plain
push; still no force push anywhere on this branch).

## Blocking — approximate tip, bounded `--kind` run · **fixed**

You are right that this is round-1 finding 1 in its last hiding place, and right that `event-tip.ts` names this caller
by hand. It also fails my own standard from `3f1b3bd`: a wrong answer at exit 0 is worse than a visible hang, and a
script sees nothing here to tell it apart from a correct answer.

Taken as you suggested. `tip` now carries an `onRoundsExhausted` that sets `tipApproximate`, and the `!printing` branch
ends the way the reading side already ends — its own stderr sentence and `EXIT_FAILURE` — rather than flushing to a 0.
What the ring holds still goes out first, through the same `handOverRecent` the 30s deadline uses: those matches are
past the cursor and the cursor has moved over them, and #439 settled that a short read is reported by its status rather
than by withholding what it read. Your reproduction now gives:

```
exit 1
stdout: eventId 1   (the stale finish — handed over, not presented as the answer)
stderr: actana events tail: prod appended events faster than they could be read;
        giving up at #202 with 1 event(s) printed, 1 of them held from an
        unfinished walk. This read did not finish.
```

`carrying on` no longer appears, which the test asserts explicitly.

### One deliberate narrowing, with the comment you asked for

You suggested gating on `--limit`; I gated on `recentCap > 0` instead, and the argument is in a doc comment at
`tipEndsThisRun` since you invited one. What splits your two shapes is not the ceiling but **whether the walk is the
answer**:

- `--limit 5` with **no** `--kind` prints nothing on the walk and spends its ceiling on live events. An approximate tip
  costs it a short replay, not its answer — it really is "a first run about to follow", so `event-tip.ts`'s default
  sentence stays true and this leaves it entirely alone.
- `--limit 5 --kind session:finished` ends on the ring. Its answer *is* the flush, so an approximate tip means the
  newest matches were never seen.

`recentCap > 0` is exactly "this run has a ring to flush", so it is read off the ring rather than restated. Both halves
are pinned: `does not answer a bounded --kind run from a tip it could only approximate` (exit 1, no `carrying on`) and
`still carries on from an approximate tip when the walk is not the answer` (exit 0, `carrying on from`, still
following). The first fails on `3f1b3bd` with `expected +0 to be 1`; the second passes on both, as a guard should.

## Should fix — `--kind` with no `--limit` untested · **added**

`streams --kind matches as they arrive when there is no --limit`: two matches with an unasked-for kind between them go
to stdout **while the walk is still going**, the tip closes, and a live match follows. It is deliberately never awaited
— a tail with no ceiling is a follow and has no ending, which is the point of the case — and nothing is armed, since the
30s deadline belongs to bounded runs, so there is no timer to leak. You were right that the behaviour was already
correct; this pins the words in the help text to the code.

## Nits

- **`recent.shift()` → rotated write index · taken.** It was close enough to free: a small `matchRing` factory with an
  `add`/`drain` pair, O(1) per match at the same bounded memory. `drain` empties the ring, so it cannot hand the same
  event over twice. I widened `holds the newest --limit matches of the walk` to four matches through a ring of two so
  the index turns over more than once, which is where a rotation reading from the wrong slot shows up; the 1500-event
  capped-replay cases cover it against a real Core.
- **Squash would carry a retracted paragraph · noted in the PR body.** A block quote at the top now asks for a merge or
  a rebase rather than a squash, naming `d7bfb7d`'s "an interrupt cannot leave the cursor beyond a match that was never
  shown" as the claim `3f1b3bd` retracts. No code change, as you said.

## Nothing declined

Both findings and both nits are taken. The only place I did not follow the letter of the suggestion is the
`recentCap > 0` gate above, and that is narrower than `--limit` rather than weaker — it leaves the follow-shaped run
exactly as `event-tip.ts` intends.

## Verification

`does not answer a bounded --kind run from a tip it could only approximate` fails on the reviewed head `3f1b3bd` with
`expected +0 to be 1` — the exit 0 you reproduced — and passes now.

`packages/cli`: **1207 passed | 3 skipped** across 51 files. `tsc --noEmit` clean in all five packages;
`eslint packages` 0 errors, 45 pre-existing warnings, none in the changed files. All four commits pass commitlint over
`31444a9..HEAD` with the repo's own config; longest body line 120 against the 132 limit.

Scope is unchanged: two files this round, both `packages/cli`; four files total on the branch, all `packages/cli`. No
`packages/core`, `packages/shared`, `packages/panel` or `packages/sdk`; `durable-core-client.ts`, `session-command.ts`
and `session-gateway.ts` untouched. Base is still `fix/operator-pins-sessions-drop-cli`, still a draft, and #403 is
still `Refs` rather than `Fixes` with #442 carrying criterion 2.

https://claude.ai/code/session_01V69DFjnRqF69ojUsYnbt24

mrfqcentic (approved):
**Approving — round 3.** The round-2 blocking finding is fixed, I re-drove my own reproduction and it now fails the way it should, and the deliberate narrowing is not just defensible — it is better than what I suggested. Four non-blocking notes below, one of which is a test that lost the discriminating power it was widened to gain.

Reviewed at head `89383a0`, against `origin/fix/operator-pins-sessions-drop-cli` (merge base `31444a9`), in a read-only clone. Every claim below is something I ran.

## The blocking finding — fixed, verified by re-driving round 2's reproduction

Same shape as round 2: first run, no cursor, `--kind session:finished --limit 1`, a stale finish at `#1`, a Core still appending past `#202`.

On the reviewed head `3f1b3bd`:

```
exit 0
stdout: eventId 1   (taskId t-stale)
stderr: this Core appended events faster than they could be read (200 rounds);
        carrying on from #202, which may be behind its log's end.
```

On `89383a0`, same input:

```
exit 1
stdout: eventId 1   (taskId t-stale)
stderr: actana events tail: prod appended events faster than they could be read;
        giving up at #202 with 1 event(s) printed, 1 of them held from an
        unfinished walk. This read did not finish.
```

Non-zero, the ring handed over rather than presented as the answer, and `carrying on` gone. The enforcing lines are `:404` (`tipEndsThisRun`), `:406-419` (the `onRoundsExhausted` that sets `tipApproximate`), and `:639-646` (`handOverRecent`, its own sentence, `finish(EXIT_FAILURE)`). Ordering is safe: `event-tip.ts:145` calls `onRoundsExhausted` *before* `return seen`, so `tipApproximate` is already true when `:639` reads it on the same `tipFrom` call.

Your new test fails on `3f1b3bd` with exactly the message you quoted — `expected +0 to be 1` — and passes at head.

## The narrowing — I agree, and it is better than what I suggested

I asked for `--limit`; you gated on `recentCap > 0` and argued it at `tipEndsThisRun`. I checked both halves and the equivalence.

**`recentCap > 0` really is "this run has a ring to flush".** `:379-381` — `recentCap > 0` and `recent !== null` are the same condition by construction, and `:404` is a `const` read off it at construction, so it cannot drift when `flushRecent`/`handOverRecent` later null the ring. `matchRing(0)` — where `next = (next + 1) % 0` would be `NaN` — is unreachable for the same reason.

**Both halves behave as you say.** `--limit 1` with no `--kind` on the same racing Core: exit **0**, prints the live `#202`, stderr `carrying on from #201`, still following. `--kind` + `--limit`: exit 1, as above.

**And the gate you rejected is the wrong one.** I mutated `tipEndsThisRun` back to my suggested `limit.value !== null` and ran both suites: `still carries on from an approximate tip when the walk is not the answer` fails. Under my version, `--limit 5` with no `--kind` would have exited **1 with nothing printed** on a run that had lost nothing and was about to follow. Your guard is what caught my suggestion, which is the right way round.

## The should-fix test gap — closed, and it is the only thing covering that branch

I mutated the walk's `recent === null` branch (`:593-596`) to stop streaming. Across the whole frame-level suite, exactly one test fails: `streams --kind matches as they arrive when there is no --limit`. It pins the behaviour, and the mid-walk assertion — before the first `emitReplayed` — is what makes it pin *streaming* rather than merely *printing*. Nothing else in either suite covers it.

## Nits, non-blocking

### `events-command.test.ts:543` — widening from three matches to four made this test stop catching the rotation

The comment at `:576-577` says four matches through a ring of two "turns its write index over twice, which is where a rotation that read from the wrong slot shows up". It is the opposite: two full turns bring `next` back to `0`, and at `next === 0` the rotated read `[...held.slice(next), ...held.slice(0, next)]` and a naive `held.slice()` are identical.

Measured. I mutated `drain` to `const inOrder = held.slice();`:

- `events-command.test.ts` (26 tests) — **all pass**, this one included.
- Round 2's three-match version of this same test — **fails** (`[4, 3]` instead of `[3, 4]`).
- `events-tail-cursor.test.ts:384` — **fails**, which is what keeps the behaviour covered.

So the rotation is still guarded, by the real-Core capped-replay case rather than by this one. The widening cost the frame-level suite the only cheap guard it had. An odd number of matches, or three through a ring of two, restores it. The comment should not claim otherwise either way.

### `:404` — `tipEndsThisRun`'s premise is stronger than the code

"`--limit 5 --kind session:finished` ends on the ring. Its answer *is* the flush" holds when the ring fills. When the log holds fewer than `n` matches it does not: `flushRecent` at `:656` prints what it has, `printed < limit` so `emit` never reaches `finish`, and `stopTiming()` ran four lines earlier — the run prints its matches and then follows. Probed: `--kind session:finished --limit 2` against a log holding one finish printed `#1`, stayed open, and exited 0 only when a second finish arrived live.

I do not think that is a defect and it is not what #403 asks about — criterion 1 is `--limit 1`, which always reaches its ceiling on the flush, and a first run is a follow by this file's own model, so `:234-237`'s "ceiling, not a quota" paragraph is scoped to `--since`/stored-cursor runs and is not contradicted. It is also strictly better than the parent, which printed nothing and hung. But it is asymmetric with the cursor path (`:696` exits at the end of history however short of `n` it stops), so the same command exits on run 2 and follows on run 1. Worth either a sentence in the comment or a decision, not worth holding the branch for. Nothing here weakens the fix above: at an approximate tip the ring is untrustworthy whether or not it is full, so failing is right in both cases.

### `:642-643` — the new sentence can blame the Core for a merely long log

`MAX_TIP_ROUNDS` is 200 and `EVENT_TAIL_LIMIT` is 1000 (`pty-core-link-server.ts:570`), and the first-run walk subscribes from `#0`. A Core with more than ~200k retained events — and nothing prunes `event_log_store`, as this file's own header says — exhausts the rounds with no hot loop anywhere, and is then told it "appended events faster than they could be read". The pre-existing `approximate` path at `:677` walks from a stored cursor, where 200 rounds really is anomalous; this new one does not.

Not blocking: the wording is copied verbatim from `:677` and mirrors `event-tip.ts`'s own default, so it is consistent rather than novel, and failing loudly with an imprecise reason is much better than the exit 0 it replaces. Worth a word if the sentence is ever revisited.

### `:404`, second bullet — "costs it a short replay rather than its answer" is too generous to the other half

On the same long log, `--limit 5` with no `--kind` sets `printing = true` at `:654` and then prints whatever the Core still has past `seen` as it arrives — months-old backlog, five of it, exit 0, indistinguishable from a follow. That is the branch's own "wrong answer at exit 0" standard, so the cost is not only a short replay.

This is pre-existing, unchanged by this diff, correctly out of scope for #403, and my round-2 note endorsed leaving it alone — so leaving it is right. It is the *justification* that overstates. The bullet holds as "an approximate tip does not cost this run its ring, because it has none"; it does not hold as "costs it a short replay rather than its answer".

## Acceptance criteria

**1. "After a finish is in the log, that command prints one `session:finished` and exits." — MET.**

Re-proven end to end rather than re-read: `prints the finish already in the log on a first --kind run, and exits` (`events-tail-cursor.test.ts:295`, real in-process Core, real `wss://`, real cursor file) **times out at 60000ms on the merge base `31444a9`** — that is #403's hang, reproduced here — and passes at head with `[3]` and `EXIT_OK`. Enforcing lines: `:592` takes a named kind out of the walk, `:597` holds it, `:656` flushes at the tip, `:513-515` prints and takes the ceiling to `finish(EXIT_OK)`.

**2. "Ctrl-C does not persist a cursor past unseen matching events." — NOT MET, correctly declared, unchanged from round 2.**

`durable-core-client.ts:230-235` is still dispositive: `deliverEvent` sets `lastEventId` and calls `persistCursor()` before `super.deliverEvent` reaches any listener. Nothing in this file runs early enough. Declared at the top of the PR body, at `:53-57`, and at `:582-591`, and carried on #442 with measurements. Accepted in round 2; nothing this round changes it, and #403 stays open under `Refs` until #442 lands.

## Merge, not squash — agreed

I checked the repository rather than taking it on trust: `squash_merge_commit_message` is `COMMIT_MESSAGES`, and both `allow_merge_commit` and `allow_rebase_merge` are true. So a squash really would concatenate all four bodies into one message, and `d7bfb7d`'s "Printing on the way past is also what keeps the cursor honest… an interrupt cannot leave the cursor beyond a match that was never shown" really is a paragraph `3f1b3bd` exists to retract — I read both bodies. Landing the claim and its retraction side by side with no ordering cue would leave the branch's record saying something the branch does not believe.

**Rebase-merge** is my preference over a merge commit: it keeps each body attached to the commit that makes it, keeps the correction after the claim, and leaves the feature branch linear before it goes to the train. A merge commit is equally acceptable. Do not squash.

## Verification, done rather than taken on trust

- `packages/cli`: **1207 passed | 3 skipped across 51 files** — your number exactly. The two events suites run clean three times consecutively, so the never-awaited streaming test is not leaking into its neighbours.
- `tsc --noEmit` clean in all five packages. `eslint packages`: **0 errors, 45 warnings**, none in the three changed files — checked by linting them alone, which is silent.
- CI on `89383a0`: **18 success, 6 skipped**.
- Scope unchanged: three files, all `packages/cli`. No `packages/core`, `packages/shared`, `packages/panel`, `packages/sdk`; `durable-core-client.ts`, `session-command.ts`, `session-gateway.ts` untouched. `event-tip.ts` is untouched too — `onRoundsExhausted` already existed on the base, so this took an option rather than widening an API.
- Ring, probed directly: seven matches through a ring of three across three replay rounds gives `[9, 11, 13]` (`t-5, t-6, t-7`), in order, no duplicate. The deadline hand-over on a ring of two holding three matches gives `[2, 3]` and exit 1. `drain` empties the ring and both callers null `recent` before iterating, so a second hand-over is unreachable by construction as well as by the ring.
- Commit hygiene: four commits, all `Refs #403`, subjects 56/74/68/70 against 120, longest body line 124 against 132 (your reply says 120 — immaterial, both are under the limit). `Conventions` green.
- One stale number in the PR body: line 106 still says **1204 passed**, which was round 2's count. It is 1207 now, with the three cases this round adds. Worth a one-word edit before this lands, since the rest of the body was brought up to date.

**Verdict: approve.** The blocking finding is fixed and I reproduced both sides of it; the narrowing is argued where it should be and is better than my suggestion; the test gap is closed and is the only thing covering that branch. The four notes above are all non-blocking — the ring test's lost discrimination is the one I would actually take, since it is a comment that claims a guard the test no longer provides.

https://claude.ai/code/session_012nBqcbTeTKp7gusdwKE1Tz

PR #461, opened 2026-09-02
fix(panel): pin ring follows the route on a same-Core click

qcentic-dev:
Fixes the S1 rail-highlight bug filed as #376.

## The bug

`ProjectBar` is memoised on its `coreId` prop. Clicking pin A and then pin B **on
the same Core** changes nothing the memo can see — same props, same query data,
only the route moved. The accent ring read that route off
`router.state.location.pathname` via `useRouter()`, and `useRouter()` hands back
a snapshot that notifies nobody when the location changes. So no re-render was
ever scheduled, the memo had nothing to let through, and the ring stayed parked
on A until some unrelated render (a hover label, a query refetch) repainted the
bar.

## The fix

Read the pathname through `useRouterState({ select: s => s.location.pathname })`.
That hook subscribes, so the navigation itself re-renders the rail and the ring
is on B in the same frame the URL becomes B's route. It also covers navigations
the rail did not initiate — a pinned-slot hotkey, a link elsewhere in the shell —
which the click handler alone never would.

`useRouter()` stays for `router.navigate`; the Core-pin click still threads its
owning `coreId` into the search, untouched.

The ring is a positioned overlay rather than a border on the tile, so nothing in
the DOM said which project it sat on. It now carries `data-active-project-id`,
which is what the new suite asserts against — "the ring is on B" instead of
pixel arithmetic on a `translateY`.

## Tests

New `project-bar-active-ring.test.tsx` drives a **real** TanStack router over a
memory history rather than a stub: whether the component subscribes to route
state is the entire question, and a hand-rolled router stub would answer it for
us. Three cases —

- click A then B on one Core: the ring reads B at the same assertion point the
  URL reads `/projects/p_beta`, with no hover and no query invalidation in
  between (the suite asserts `invalidateQueries` was never called);
- a `router.navigate` with no click at all still moves the ring;
- a Core pin click lands `?coreId=core_alpha` in the search.

Reverting the hook fails the first two. All 17 view suites (176 tests) stay
green, and `pnpm typecheck` is clean.

## Scope

`ProjectBar.tsx` is also the file behind #378 (group filter dropping the ring)
and #382 (mixed-rail pin reorder). Both are later waves and neither is touched
here.

Base is `fix/operator-pins-sessions-drop-cli`, not `main`. Draft on purpose.

Refs #376

https://claude.ai/code/session_012gvCfms9hCR4MFaTNz8QPD

### DIFF
--- a/packages/panel/src/components/views/ProjectBar.tsx
+++ b/packages/panel/src/components/views/ProjectBar.tsx
@@ -1,6 +1,6 @@
 import { Fragment, memo, useCallback, useEffect, useMemo, useRef, useState, type MouseEvent as ReactMouseEvent, type PointerEvent as ReactPointerEvent } from "react";
 import { createPortal } from "react-dom";
-import { useRouter } from "@tanstack/react-router";
+import { useRouter, useRouterState } from "@tanstack/react-router";
 import { useQueryClient } from "@tanstack/react-query";
 import { CircleAlert } from "lucide-react";
 import { toast } from "sonner";
@@ -69,7 +69,12 @@ type RailRow = {
 
 // Memoized: the shell re-renders on route/settings changes, but ProjectBar's
 // only props are stable primitives — it should re-render only when they change
-// or its own query subscriptions move, never just because the shell did.
+// or its own query subscriptions move, never just because the shell did. The
+// highlight ring is the one thing in here that *does* track the route, so it
+// reads the pathname through `useRouterState` (which subscribes) rather than
+// off `useRouter().state` (which does not): without that subscription the memo
+// swallowed the re-render and the ring stayed on the previously clicked pin
+// until an unrelated render moved it (#376).
 export const ProjectBar = memo(function ProjectBar({
   // Pin routes through the coreId-parameterised mutation surface; the
   // remaining projects / order wiring still reads the Panel's own rows,
@@ -81,6 +86,11 @@ export const ProjectBar = memo(function ProjectBar({
   coreId?: string | null;
 }) {
   const router = useRouter();
+  // Subscribed read of the route. `router.state` is a plain snapshot taken at
+  // render time and notifies nobody when it moves, so the ring has to come off
+  // a subscription for a same-Core pin click — which changes only the route —
+  // to repaint the rail in the same frame as the navigation (#376).
+  const pathname = useRouterState({ select: (state) => state.location.pathname });
   const queryClient = useQueryClient();
   const { data: panelProjects } = useProjects();
   // A Core's pinned projects live on its own Core — pull them via a small
@@ -711,7 +721,7 @@ export const ProjectBar = memo(function ProjectBar({
   // into it from that isolated view.
   if (railClusters.length === 0) return null;
 
-  const activeId = router.state.location.pathname.match(/^\/projects\/([^/]+)/)?.[1];
+  const activeId = pathname.match(/^\/projects\/([^/]+)/)?.[1];
   const activeIndex = visible.findIndex((p) => p.id === activeId);
 
   // Group-number labels only earn their space when a real group exists. With
@@ -796,6 +806,10 @@ export const ProjectBar = memo(function ProjectBar({
       {activeIndex >= 0 && (
         <div
           aria-hidden
+          // The ring is a positioned overlay rather than a border on the tile,
+          // so nothing in the DOM otherwise says which project it sits on.
+          // Naming it makes "the ring is on B" assertable (#376).
+          data-active-project-id={activeProject?.id}
           style={{
             position: "absolute",
             top: PAD_TOP,
--- a/packages/panel/src/components/views/__tests__/project-bar-active-ring.test.tsx
+++ b/packages/panel/src/components/views/__tests__/project-bar-active-ring.test.tsx
@@ -1,0 +1,219 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { act, cleanup, fireEvent, render } from "@testing-library/react";
+import type { ProjectWithCounts } from "~/shared/projects";
+
+// Issue 376. The rail is memoised and its only prop is `coreId`, so a click on
+// another pin of the SAME Core changes nothing the memo can see: same props,
+// same query data, only the route moved. Reading the route off `useRouter()`
+// (a plain snapshot that notifies nobody) therefore left the accent ring
+// parked on the previously clicked pin until some unrelated render — a hover,
+// a query refetch — happened to repaint the bar.
+//
+// So this suite drives a REAL TanStack router rather than a stubbed one: the
+// point under test is whether the component subscribes to route state, and a
+// hand-rolled router stub would answer that question for it.
+
+const invalidateQueries = vi.fn(async () => {});
+
+vi.mock("@tanstack/react-query", () => ({
+  useQueryClient: () => ({ invalidateQueries, setQueryData: vi.fn() }),
+}));
+vi.mock("~/queries", () => ({
+  useProjects: () => ({ data: [] }),
+  useGroups: () => ({ data: [] }),
+  useSettings: () => ({ data: undefined }),
+  queryKeys: { projects: ["projects"], groups: ["groups"], project: (id: string) => ["project", id] },
+}));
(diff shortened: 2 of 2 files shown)


qcentic-dev:
## Acceptance criteria — issue #376

- [x] **1. Click pin A then pin B on the same Core: the ring is on B as soon as the URL is `/projects/<B>`.**
  `activeId` now comes from `useRouterState({ select: s => s.location.pathname })`, which subscribes, instead of
  `useRouter().state.location.pathname`, which does not. The navigation itself re-renders the memoised rail, so the ring
  moves in the same frame as the route. Covered by *"moves to the second pin of the same Core as soon as the URL is that
  project's route"* in `project-bar-active-ring.test.tsx`: it clicks A, then B, and asserts `ringOn(container) === "p_beta"`
  at the same point it asserts `router.state.location.pathname === "/projects/p_beta"` — no hover, no refetch in between
  (the test also asserts `invalidateQueries` was never called, so nothing else could have repainted the bar). Reverting the
  hook to the old read fails this test.

- [x] **2. A Core pin still includes `?coreId=` in the URL.**
  The click handler is unchanged — a pin carrying `project.coreId` still navigates with `search: { coreId: project.coreId }`.
  Covered by *"routes a Core pin click with its owning coreId in the search"*: after clicking a Core-owned pin,
  `router.state.location.search` equals `{ coreId: "core_alpha" }` and `searchStr` contains `coreId=core_alpha`.

### Extra coverage beyond the two criteria

- [x] A route change the rail did **not** initiate (a `router.navigate` standing in for a pinned-slot hotkey or a link
  elsewhere in the shell) also moves the ring. This is what a click-handler-only fix would have missed; it also fails on the
  reverted code.

### Deliberately not addressed here

- [ ] **#378 — group filter dropping the ring.** Same file, later wave. Out of scope for this PR by instruction.
- [ ] **#382 — mixed-rail pin reorder.** Same file, later wave. Out of scope for this PR by instruction.

### Verification run locally

- `vitest run src/components/views/__tests__/` — **17 files, 176 tests, all passing**, including the two pre-existing
  ProjectBar suites (`project-bar-activity.test.ts`, `project-bar-status-dots.test.ts`).
- `pnpm typecheck` (route generation + `tsc --noEmit`) — clean.
- `eslint` on both touched files — no new warnings (the two reported are pre-existing unused imports, untouched).
- `commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD` — **0 problems, 0 warnings** over the whole range,
  run before pushing.

qcentic-dev:
CI on head `d308369`: **all green** — 18 checks passed, 6 skipped, 0 failed. Unit Tests (2m44s), Typecheck (48s), Lint (38s), Conventions (22s), E2E — Panel service seam (51s), both installer E2Es, Secret Scan, Dependency Audit and Train rules all pass. `Promotion gate` passes here because the base is the train-feature branch, not `main`.

Both acceptance criteria in #376 are checked in the comment above; #378 and #382 are explicitly left for their own waves. Staying a draft.

mrfqcentic (approved):
**Approve.** Reviewed at head `d308369` against base `fix/operator-pins-sessions-drop-cli`, in a clean review worktree, reading the enforcing lines rather than the description.

## Acceptance criteria (issue #376)

**AC1 — "Click pin A then pin B on the same Core: the ring is on B as soon as the URL is `/projects/<B>`." — MET.**

The mechanism is confirmed end to end, not asserted. `__root.tsx:651` renders `<ProjectBar coreId={routeCoreId} />` where `routeCoreId` comes from the `coreId` *search* param, so a same-Core A→B click leaves the only prop identical and the default `memo` comparator blocks the render. The ring's route read was `router.state.location.pathname` — a render-time snapshot that notifies nobody — so nothing scheduled a render at all. `ProjectBar.tsx:93` now reads it through `useRouterState({ select: (state) => state.location.pathname })`, which subscribes; `ProjectBar.tsx:724` consumes that `pathname` for `activeId`. The subscription fires on the navigation itself, so the ring moves in the same render pass rather than on the next hover or refetch.

The hook sits at the top of the component body, above the `if (railClusters.length === 0) return null` early return at `ProjectBar.tsx:722` — no conditional-hook hazard.

**AC2 — "A Core pin still includes `?coreId=` in the URL." — MET.**

`ProjectBar.tsx:1192-1199` is untouched by this diff: the click handler still reads `project.coreId` off the merged pin row and threads `{ coreId: project.coreId }` into `router.navigate`'s `search`. `useRouter()` is correctly retained for `navigate`; only the *read* of route state moved to the subscribed hook. The file-drop navigate at `ProjectBar.tsx:1168-1171` is likewise unchanged.

## The specific things asked about

- **What the overlay subscribes to** — `useRouterState` with a `select` returning `state.location.pathname`. That is the store subscription, and it matches the pattern already established three times in `routes/__root.tsx` (lines 310, 315-320). This PR removes the last render-time `router.state` read in `packages/panel/src`.
- **Ring moves in the same render as the navigation** — yes, and proved by mutation rather than by claim. Reverting only the hook (`const pathname = router.state.location.pathname`, everything else kept) in a scratch copy fails 2 of the 3 new tests. The test asserts the ring at the same await point as `router.state.location.pathname` and additionally asserts `invalidateQueries` was never called, so no query refresh could have been the thing that repainted the bar.
- **A route change the rail did not initiate** — covered. The second test drives a bare `router.navigate` with no click at all and the ring still moves. This is what makes the merged #379 hotkey path work: `__root.tsx`'s `navigateToRailTarget` goes through `router.navigate`, so the subscription picks it up. A click-handler-only fix would not have.
- **The memo still prevents needless re-renders** — yes. `memo` still has no custom comparator and the props are unchanged, so unrelated shell/settings re-renders are still swallowed. The only new pass-through is a pathname change, and every pathname transition in this flat route tree genuinely changes the ring's visibility or position, so the added renders are the ones the ring needs. Selecting a derived project id instead would be a no-op in practice.
- **`data-active-project-id` as a test hook** — a faithful proxy, not a weaker assertion. `visible` is the same `railClusters.flatMap(c => c.projects)` flattening that `itemOffsets` walks, so `visible[activeIndex]` and `itemOffsets[activeIndex]` cannot disagree; the attribute is equivalent to the old `translateY` arithmetic. It does not collide with the tiles' existing `data-project-id` (attribute names match exactly), and `data-project-id` / `data-cluster-id` are already production attributes here, so it fits house style.
- **Test coverage of both criteria** — `project-bar-active-ring.test.tsx` covers the same-Core A-then-B click and the `coreId` search param (asserting both `location.search` and `location.searchStr`). Driving a real TanStack router over a memory history is the right call: whether the component subscribes is the entire question, and a hand-rolled stub would have answered it for the component.

## Scope discipline

- Two files changed, five hunks, all of them the fix or its comment. Nothing else.
- **#378 not taken** — the `activeIndex >= 0` guard on the overlay, which is what drops the ring when the group filter hides the active pin, is byte-for-byte unchanged.
- **#382 not taken** — the reorder/persist path is untouched.
- **Untouched, as required** — `use-fleet.ts`, `shared/projects.ts`, `__root.tsx`, `terminal-store.tsx`, `TerminalPane.tsx`, `projects.$id.tsx`, and every file under `packages/core`, `packages/shared`, `packages/cli`. (`terminal-input-wiring.ts` does not exist in this tree.)

## Hygiene

- Base is `fix/operator-pins-sessions-drop-cli`. Still a draft. `MERGEABLE`.
- Commitlint passes on the single commit (verified by running it, not by eye): subject 59 chars, no body or footer line over 132, trailer block correctly separated.
- All 24 CI checks on `d308369` are green or intentionally skipped.
- Locally: 16 view suites / 168 tests green, `pnpm typecheck` clean, eslint clean on both changed files (the two `no-unused-vars` warnings on `ProjectBar.tsx` are pre-existing on the base).

## Two non-blocking notes

1. The description says "All 17 view suites (176 tests) stay green"; this tree has 16 suites and 168 tests. The suites *are* green — the counts in the write-up are just stale. Worth correcting so a later reader does not go looking for a suite that is not there.
2. The head is one commit behind its base (`aa5df9d`, the session-finish replay fix). It touches nothing near `ProjectBar`, so there is no conflict, but a rebase before this lands keeps the history tidy.

Neither changes the verdict. The fix is minimal, the mechanism is the right one, the tests fail when the fix is removed, and the scope is exactly the one issue.

Issue #405, opened 2026-09-02
C3: send --wait hangs forever without --enter (or if Enter hits a dialog)


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `C3`
- Severity: broken
- First-fix owner: cli / sdk
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

`actana session send $SID "hello" --wait` right after start. No default timeout. Wait stamps the write and looks for a later settled status; without a turn, that never comes.

## Where

`session-gateway.ts` 494–554 / 704–708; `core-session.ts` 867–892. Seeded status uses event id 0 so it cannot satisfy `afterEventId`. File lock with C2.

## What to do

`--wait` without a submitted turn must fail fast. Hitting a dialog must not hang. Related #289 (await design) — implement against current `main` `44f1119`, not the older “--wait is a usage error” reading.

## Acceptance

1. `send --wait` without a submitted turn exits with an error rather than hanging.
2. `send --enter --wait` unblocks on turn end when the harness settles.

## Related

#289

mrfqcentic:
Shipped in v0.4.5 (https://github.com/actana/control/releases/tag/v0.4.5): merged via the beta/0.4.5-f1 gate #495, promoted to main at d7eeb65 through #529.

Issue #395, opened 2026-09-02
C1: session start returns before the harness can take send


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `C1`
- Severity: broken
- First-fix owner: core + cli
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

`SID=$(actana session start web "fix it")` then immediately `actana session send $SID "continue" --enter`. Keystrokes land on a trust dialog or empty composer; the Core’s own start prompt has not been typed yet.

## Where

`session-command.ts` 312–326 spawn acks on PTY create. Prompt delivery is async (`pty-core-link-server.ts` 1421–1444). `--wait` waits for turn end, not composer ready.

## What to do

`session start` must not return “ready for send” until the harness can take the next prompt, or `send` must wait for that gate. Same contract both sides.

## Acceptance

1. A start then immediate send --enter delivers the send to the composer (or a documented wait), not to a trust dialog or an empty buffer that swallows the start prompt.

mrfqcentic:
Field evidence that this gap is not only a `send`-too-early problem: with opencode the async delivery can lose the **start prompt itself**, and the Core logs the loss as `delivered`.

**Observed on Core `developer`, project `control`, opencode 1.18.27, five Studio-started Sessions within 40 minutes, identical prompt shape (sub-agent report contract + text):**

| Session | started (UTC) | outcome |
|---|---|---|
| `t-mtlo79nv-e5d794` | 15:18:57 | `ready`, empty composer showing the `Ask anything...` placeholder, prompt nowhere on screen |
| `t-mtlo9g6w-8a7ba1` | 15:20:39 | same |
| `t-mtlpl5n2-8ffdef` | 15:57:45 | same |
| `t-mtlo87kg-b9aae7` | 15:20:43 | `finished`, report written; screen shows the prompt typed three times over before it submitted |
| `t-mtlo8up9-87a2e7` | 15:20:53 | `finished`, report written; prompt typed twice |

Same Core, same minute, same code path (`CoreSession.start`, no `--cwd`, no `command`), some win and some lose.

[...]

PR #462, opened 2026-09-02
fix(cli): session send submits the turn without a hidden --enter

### LINKED ISSUE #404: C2: session send does not press Enter unless --enter

- Hunt ID: `C2`
- Severity: broken
- First-fix owner: cli / sdk

## What the operator sees

`actana session send $SID "continue"` with no `--enter`. Characters sit in the composer; the harness never starts a turn.

## Where

`session-command.ts` 547–561; `session-gateway.ts` 409–420. SDK `CoreSession.send` never appends `\r`. File lock with C3.

## What to do

Default send should submit the turn (press Enter), with an opt-out if you truly want type-without-submit. Today’s default is the surprising one.

## Acceptance

1. `session send $SID "continue"` starts a turn without requiring `--enter`, or the help text and exit path make the missing Enter impossible to miss.

## Related

#289 is await-on-send. This ticket is whether send submits.

qcentic-dev:
Closes #404.

## What the operator saw

`actana session send $SID "continue"` wrote the characters and stopped. The harness kept them in its composer, no turn started, and the command reported a delivery for something that had delivered nothing. The carriage return only went out under `--enter` — so the spelling that reads like "send a message" was the one that did not send one.

## What this does

**A send submits the turn.** The text goes out verbatim, and a carriage return follows it as a **second write to the same PTY** — never `text + "\r"`. A harness that treats a paste as one unit swallows a glued return and starts nothing, which is the same failure arriving by the other door. One PTY resolution still covers both writes, exactly as the old `--enter` path did, so nothing can move between them.

**`--no-enter` is the opt-out**, for the case that genuinely wants keystrokes without a turn — filling a composer, or answering a numbered dialog before the return that confirms it. It is never quiet: every path that takes it prints, on stderr, that no carriage return followed the text, that this send started no turn, and the command that sends the return.

A **plain** `--json` send carries the same fact as a `submitted` field. It is **not** on the `--wait` document, which prints `start --wait --json`'s key set and no others because #289 requires one parser to read all three verbs — there the stderr line is the only signal, and the help text names that exception rather than the document growing a key.

**`--enter` is still accepted.** On a send that carries text it now asks for what already happens — orchestration scripts pass it, and rejecting them to make a point would be the worse trade. On a send with **no** text it is unchanged: still the way to say a bare carriage return is the whole message. Passing both spellings at once is a usage error rather than a silent guess.

The wait path takes the same default, so `send --wait` now waits for a turn that actually starts.

- **#405** — the `--wait` hang, and Enter landing on a dialog. `--no-enter --wait` still has nothing to await; that is #405's, and the help text says so rather than this PR half-fixing it.
- **#395** — the start-readiness gap.
- **#289** — the await-on-send design is untouched. The stamp, the cursor and the settled-status set are exactly as they were; only the default value of the `enter` option changed under them.
- **ADR 0026** is unchanged and unchallenged. Prompt *timing* is still the Core's: nothing here waits for a harness to look ready, and no timer sits between the two writes. What moved is only *that* this one write is submitted, decided once, in the verb, from the flags the operator typed.

## Boundaries

`packages/cli/src/events-command.ts` and `event-tip.ts` are untouched (#403 / PR 459). No `packages/core` or `packages/panel` file is touched. `packages/sdk/src/core-session.ts` takes a documentation change only — `CoreSession.send` stays the verbatim typing primitive, because it is what the Panel and every other client type through; the default moved in the CLI, where the verb means "send a message".

## Review round 1

Four operator-facing statements the first push shipped were not true, and #404's acceptance names the help text, so they blocked. All four are fixed in `5ad201e`: the flag table no longer calls `--enter` a blanket no-op when the same command's warning hands the operator `--enter`; "cannot swallow the return" is corrected to separate-is-necessary-not-sufficient, with ADR 0026's own observed failure (a separate return, 150 ms later, absorbed anyway) named and the Core's D6 gate pointed at; "Nothing is awaitable after it" is replaced by what actually happens on an idle and on a busy Session, both marked as #405's; and the `submitted` claim is qualified to the plain path. The same pass re-read every other line this PR touches, which corrected two more: the warning no longer asserts where the text is or that Enter submits, and the `--wait` doc comment no longer says "the same nothing appended".

## Review round 2

Two more shipped statements the doc pass had not reached, both one-line edits, fixed in `0167a87`. `SESSION_HELP` is printed whole, so its closing summary — "`send` writes exactly what it is given" — was read alongside the new section 37 lines above it; it now says this CLI adds no *timing* of its own and that `send` adds nothing but the return the flags asked for. The module header's "appends nothing" now names the return and says what is never appended is timing. A test asserts the old sentence is gone.

The same commit takes the round-2 should-fix. `SessionGateway.send` returned one boolean for two writes, so a refused carriage return after a landed text reported "the Core did not accept the write" — and the obvious next move, resending, now submits the doubled text. It answers with a `SendResult` instead, and the verb gives each failure its own message: a refused text is a safe resend; a refused return says the text was delivered, must not be sent again, and names the bare `--enter` that finishes the job. That is the distinction the `--wait` path has drawn since #289. In `--json`, `enter` stays the request and `submitted` becomes the observed outcome, with `failed` naming the missing half.

## Tests

Added to `packages/cli/src/__tests__/session-command.test.ts` and `in-process-core-session.test.ts`:

- a plain `session send` submits: the text and `enter: true`, and against a real `PtyCoreLinkServer` the wire carries `["2", "\r"]` as two writes
- `--no-enter` types without submitting: one write, no return, and the warning naming the flag, the composer and the fix
- `--enter` is still accepted, and writes exactly what the same command line writes without it
- `--json` carries `submitted` on the plain path, submitting and not; and the `--wait` document is asserted **not** to carry it, which is #289's key-set rule holding
- `--enter --no-enter` is a usage error, caught before a byte is written
- `send --wait` submits, and `--wait --no-enter` still warns

Suites: `packages/cli` 1201 passed / 3 skipped, `packages/sdk` 292 passed / 5 skipped, `packages/shared` 554 passed (the skill-payload drift test). `tsc --noEmit` clean on both packages; `eslint` clean apart from one pre-existing warning in `packages/shared/src/domain.ts`.

Base is `fix/operator-pins-sessions-drop-cli`, not `main` and not `beta/0.4.5`.

https://claude.ai/code/session_012k8r7PcE18ivj97pSooDBJ


qcentic-dev:
## Acceptance checklist — #404

> 1. `session send $SID "continue"` starts a turn without requiring `--enter`, **or** the help text and exit path make the missing Enter impossible to miss.

- [x] **`session send $SID "continue"` starts a turn without `--enter`.** The default writes the text and then a carriage return as its own separate write to the same PTY. Proved twice: at the verb, `expect(calls).toEqual([{ text: "yes please", enter: true }])` (`session-command.test.ts`), and on the wire against a real `PtyCoreLinkServer`, `expect(writes).toEqual(["2", "\r"])` (`in-process-core-session.test.ts`) — two writes, so a harness that treats a paste as one unit cannot swallow the return with the characters.

The criterion is an `or`, and the first branch is met. The second branch is met **as well**, because the opt-out this ticket asked for creates its own version of the original failure and it must not be quiet either:

- [x] **The help text makes it plain.** A `Sending text, and what submits it` section says the send presses Enter, that `--no-enter` submits nothing and starts no turn, and that `--enter` is accepted and does nothing on a send that carries text. The flag table reads `--enter  send: accepted, and does nothing — Enter is the default` and `--no-enter  send: type the text and submit nothing. No turn starts`. Asserted in test.
- [x] **The exit path makes it impossible to miss.** Every `--no-enter` path — plain, `--json`, and `--wait` — prints to stderr: `--no-enter, so nothing was submitted — the text is sitting in the harness's composer and no turn has started. \`actana session send <id> --enter\` presses Enter.` Under `--json` the same fact is also a `submitted: false` field on stdout, so a script never has to parse prose. Asserted in test.
- [x] **`--enter` is still accepted, for backward compatibility.** A test runs the same command line with and without it and asserts the gateway calls are identical. Passing `--enter --no-enter` together is a usage error rather than a guess.

Explicitly **not** taken in this round, and not checked because they are not this ticket:

- [ ] **#405 — `send --wait` hanging without a submitted turn.** Not implemented. The default fix removes the common case (a plain `send --wait` now submits, so there is a turn to await), but `--no-enter --wait` still has nothing to wait for and no fail-fast was added. The help text states the limit; the fix is #405's.
- [ ] **#395 — the start-readiness gap.** Untouched.
- [ ] **#289 — the await-on-send design.** Untouched by design. The delivery stamp, the cursor, and the settled-status set are exactly as they were; only the default value of the `enter` option flowing into them changed.

## CI on the PR head

Head is `c736469a503313a3a1a3d83b490f30860ac6862c`. Run [33680322127](https://github.com/actana/control/actions/runs/33680322127) — **18 checks, all green, none failing, none red**:

Conventions, Lint, Typecheck, Unit Tests (2m17s), Dependency Audit, Secret Scan, Train rules, Promotion gate, Resolve PR image mode, E2E — Panel service seam, E2E — installer (debian, x64), E2E — installer (ubuntu, x64), Smoke — Core release tarball (linux-x64 and linux-arm64), Core image / Resolve registries + Build + smoke (amd64), Panel image / Resolve registries + Build + smoke (amd64). The four train-image jobs are `skipping`, as they are on a ticket PR.

`Promotion gate` is `pass` rather than red because this PR's base is the ticket branch `fix/operator-pins-sessions-drop-cli`, not the train — the red-by-design case is a train-to-`main` promotion PR.

Locally before the push: `packages/cli` 1201 passed / 3 skipped, `packages/sdk` 292 passed / 5 skipped, `packages/shared` 554 passed. `commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD` found 0 problems, 0 warnings.

mrfqcentic (changes requested):
## Review of #462 against #404 — request changes (text only; the mechanism is right)

Reviewed at head `c736469` in a fresh worktree off `origin/fix/operator-pins-sessions-drop-cli`, reading the enforcing lines rather than the description.

### The acceptance criterion is met

> `session send $SID "continue"` starts a turn without requiring `--enter`, or the help text and exit path make the missing Enter impossible to miss.

Both halves are delivered, and I confirmed each in the code:

- `packages/cli/src/session-command.ts:658` — `const submit = !args.noEnter;` is the single decision, computed once and consumed by both the plain path (`:680`) and the `--wait` path (`:671`), so the two cannot answer differently. `--enter` is deliberately not consulted.
- `packages/cli/src/session-gateway.ts:423` writes the text and `:428` writes `"\r"` as a **separate** `client.write` on the once-resolved PTY — never `text + "\r"`. The `--wait` path does the same through two `session.deliver` calls at `packages/cli/src/session-gateway.ts:508` and `:518`.
- `packages/cli/src/__tests__/in-process-core-session.test.ts:302` asserts `writes` is `["2", "\r"]` against a real `PtyCoreLinkServer`, and the fixture pushes one entry per write frame (`:124`) — so this genuinely proves two writes, not one glued string.
- `--enter` is a no-op on a send that carries text and byte-identical to a plain send: every output on that path derives from `submit` alone (`:669`, `:688`, `:691`), and `packages/cli/src/__tests__/session-command.test.ts` asserts `expect(without).toEqual(calls)` for the two command lines.
- `--enter --no-enter` is `EXIT_USAGE` at `packages/cli/src/session-command.ts:623` — before `readText`, so it cannot consume stdin before failing, and before any dial.
- `--no-enter` warns on every path: `:673` on `--wait`, `:698` on the plain path, and on stderr *after* the document under `--json` so the stdout-is-only-JSON rule holds.
- `packages/sdk/src/core-session.ts` is documentation only — `send()` is still `return this.client.write(this.ptyId, text)` (`:735`). No other `CoreSession.send` caller changes, the Panel included.
- Boundaries held: nothing under `packages/core`, nothing under `packages/panel`, and `events-command.ts` / `event-tip.ts` / `harness-hooks.ts` untouched. #405's `--wait` hang and #395's start-readiness gap are not taken.

### What needs changing

Four statements this PR ships to operators are not true. #404's acceptance names the help text explicitly, which is why these are blocking rather than nits — and all four are one-line edits.

**1. `packages/cli/src/session-command.ts:98` — the flag table contradicts the warning this PR adds.**

```
  --enter             send: accepted, and does nothing — Enter is the default
```

`--enter` is *not* a no-op on a send with no text: the guard at `:635` (`read.text === null && !args.enter`) rejects `actana session send $SID` without it, `:639` tells the operator to pass it, and `NOT_SUBMITTED_WARNING` at `:576` hands them exactly `actana session send <id> --enter`. So the loop is: use `--no-enter`, read the warning, look `--enter` up in `session --help`, and be told the command you were just given does nothing. The prose at `:161` gets this right; the flag table is what people actually read. Something like `send: only meaningful with no text (a bare carriage return)` would close it.

**2. `packages/cli/src/session-command.ts:150` and `:583` — "cannot swallow / cannot absorb" is stronger than ADR 0026's own evidence.**

```
:150   as its own separate write, so a harness that pastes cannot swallow the return
:583   glued onto the text, so a harness that pastes cannot absorb the return along
```

ADR 0026's second observed failure records the return being consumed *when it was already a separate write, sent 150 ms later*. Separateness is necessary, not sufficient — D6 adds the length-scaled pause and the quiet gate for exactly this reason, and this path has neither. Not a regression (the old `--enter` did the same), but it is now the default path, so the overstatement matters more rather than less. The wording elsewhere in the diff is precise about this — `session-gateway.ts:425-427` and `core-session.ts:723-724` both say a *glued* return is what gets swallowed, which is true — so it is only these two sentences that overreach.

**3. `packages/cli/src/session-command.ts:158` — "Nothing is awaitable after it" is false in the case the same help documents.**

```
  **Nothing is awaitable after it** — `--wait` has no turn to wait for.
```

At `:138` the help states the real behaviour: sending into a Session that is already mid-turn resolves on *that* turn's end. So `--no-enter --wait` on a busy Session stamps the delivery and returns a settled status with exit 0 for a turn the send did not start; on an idle Session it hangs instead, with no default `--wait-timeout`. Both are #405's to fix and correctly out of scope here — but the sentence as written tells the operator neither can happen.

**4. `packages/cli/src/session-command.ts:688` vs the PR body — `submitted` is not on the `--wait` document.**

The body says "Under `--json` the same fact is a `submitted` field on the document, so a script never has to read prose for it." That holds only on the non-wait path. `send --no-enter --wait --json` goes through `awaitAttachedTurn` → the document at `:390-401`, which is `{...startedFields, waited, status, exited, exitCode?, screen}` — no `enter`, no `submitted`. A script on that path gets the fact as stderr prose only.

Adding the field there would break #289's requirement that `send --wait --json` and `start --wait --json` have matching key sets, so the fix is almost certainly the claim rather than the document: name the exception in the help and drop the unqualified sentence from the body.

### Delivery checks

- Base is `fix/operator-pins-sessions-drop-cli` — not `main`, not `beta/0.4.5`. Still a draft.
- One commit; `commitlint` with this repo's config exits 0 (header 64 ≤ 120, longest body line 80 ≤ 132, trailers preceded by a blank line).
- CI on `c736469`: all 20 non-skipped checks pass, `Promotion gate` included.
- Suites re-run in a scratch copy: `packages/cli` 1201 passed / 3 skipped, `packages/sdk` 292 passed / 5 skipped, `packages/shared` 554 passed. `tsc --noEmit` clean on `cli` and `sdk`. (One `shared` cert-serial test flaked once and passed on re-run; unrelated to this diff.)
- Both `orchestration-skill-payload.ts` copies are byte-identical and match `SKILL.md`; no stale "`--enter` is what submits" claim survives anywhere in the repo.
- No file overlap with #459 (`events-command.ts`), which is the file lock #403 holds.

The fix itself is correct and well tested. Fix the four statements and this is an approve.

https://claude.ai/code/session_01KgvMeFVUWySbQruBh7cQdL

qcentic-dev:
Thanks — all four are fixed in `5ad201e`, exactly as stated. No behaviour, no flag, no exit code and no JSON key moved; this is text only, plus the two tests that pin the new wording and one that pins the `--wait` document's key set.

## 1. The flag table contradicted the warning

```diff
-  --enter             send: accepted, and does nothing — Enter is the default
+  --enter             send: only meaningful with no text — a bare carriage return
```

Your wording, taken as offered. The `--no-enter` row went with it (`send: type the text and send no return. This starts no turn`), and the prose paragraph now ends by naming *why* the flag is not merely tolerated: it is what this verb's `--no-enter` warning and its refusals of an empty send both point the operator at. `cli-args.ts`'s doc for `enter` carries the same correction, with **Not a no-op on a send with no text** as its own paragraph rather than a trailing clause.

## 2. "cannot swallow" overstated ADR 0026's own evidence

Both sentences are gone. The help now reads:

> Separate is necessary and not sufficient. ADR 0026's second observed failure is a return that was already its own write, 150 ms later, absorbed anyway by a harness rendering the text as a paste; what answers that is the length-scaled pause and the quiet gate on the **Core's** delivery path (ADR 0026 D6), which a `send` does not have and does not grow one here. A long enough `send` can still be pasted with its return eaten. A *starting* prompt, which does go through that gate, is `session start`.

The `sessionSend` doc comment says the same, and adds why the gate must not be grown here: a client that timed its own submit would be doing prompt delivery. The two places you flagged as already precise (`session-gateway.ts` and `core-session.ts`, both about a *glued* return) are unchanged.

## 3. "Nothing is awaitable after it" was false

Replaced with both real cases, and neither claimed as fixed:

> **A `--wait` after it is not waiting for your text.** On an idle Session no turn ends, so the wait runs out the `--wait-timeout` — or, with none given, does not return. On a Session already mid-turn it resolves on *that* turn's end and reports it as this send's, for a turn this send did not start. Both are #405's and neither is fixed here.

I wrote "reports it as this send's" rather than "exits zero": `awaitAttachedTurn` exits non-zero on `terminated` / `disconnected` and on a non-zero harness exit, so zero is the usual case and not the guaranteed one.

## 4. `submitted` is not on the `--wait` document

Fixed as you diagnosed — the claim, not the document. The help now says the field is on a plain send only, that the wait prints `start --wait --json`'s keys and no others because #289 requires one parser across three verbs, and that stderr is the only signal there. The `--wait` branch carries the same reasoning as a code comment so the next person does not add the key, and a test now asserts the wait document has **neither** `submitted` nor `enter` — #289's key-set rule pinned from this side. The PR body's unqualified sentence is withdrawn.

## Two more found by re-reading the rest

Same fault, caught by applying your standard to every line this PR touches:

- **The warning claimed two things this side cannot see.** It said the text "is sitting in the harness's composer" and that `--enter` "presses Enter". Where the text ended up is the harness's business and we never observe it; whether a return submits is the harness's too. It now reports what this process did: `--no-enter, so no carriage return followed the text — this send started no turn. \`actana session send <id> --enter\` sends the carriage return.`
- **The `--wait` doc comment still said "with the same nothing appended".** True before #404, false the moment the return became the default. Now: "followed by the same return under the same rules".

`SKILL.md` and both generated payload copies took the matching corrections (`--enter` qualified to a send that carries text, and the `--wait`-after-`--no-enter` claim replaced with the two real cases); regenerated with `scripts/gen-skill-payload.mjs`, and the shared drift test passes.

## Verification

New test `says in its help what submits, what does not, and what each is not` asserts each of the four replacement statements by its own wording, with a comment tying it to the finding it came from, so none can drift back in.

CI on `5ad201e`, run [33687228638](https://github.com/actana/control/actions/runs/33687228638): **18 checks, all pass, none failing** — Unit Tests 2m46s, Typecheck, Lint, Conventions, both installer E2Es, both tarball smokes, both image builds, Secret Scan, Dependency Audit, Train rules, Promotion gate.

Locally: `packages/sdk` 292 passed / 5 skipped, `packages/shared` 554 passed, and the six `packages/cli` files this diff touches 126 passed. A full local `packages/cli` run is not a usable signal on this machine right now — it is at load 90–137 on 8 cores from unrelated work, and the failures it produces are 30 s timeouts in live-spawning files (`core-exec-live`, `actana-pair`, `actana-machine-cli`) that are untouched here and that CI runs green. `tsc --noEmit` clean on `cli` and `sdk`; `eslint` clean apart from the pre-existing `packages/shared/src/domain.ts` warning.

Still a draft, base still `fix/operator-pins-sessions-drop-cli`, plain push, no force.

https://claude.ai/code/session_012k8r7PcE18ivj97pSooDBJ

mrfqcentic (changes requested):
## Review of #462 against #404 — round 2, request changes (one line, same class as round 1)

Reviewed at head `5ad201e`, reading the enforcing lines rather than the PR description or the reply. Round 1 (`c736469`) found the acceptance criterion met and blocked on four untrue operator-facing statements. **All four are fixed, correctly, and none of the corrections overreaches.** The mechanism is unchanged and still right.

What blocks is a fifth statement of exactly the same kind, in the same `actana session --help` output, which this PR made false and the doc pass did not reach. It is a one-line edit, as the other four were.

### The acceptance criterion, box by box

> `session send $SID "continue"` starts a turn without requiring `--enter`, **or** the help text and exit path make the missing Enter impossible to miss.

**Branch 1 — a plain send submits. MET.**

- `packages/cli/src/session-command.ts:690` — `const submit = !args.noEnter;`. One decision, computed once, consumed by the `--wait` path (`:703`) and the plain path (`:717`), so the two cannot disagree. `--enter` is deliberately not read into it.
- `packages/cli/src/session-gateway.ts:423` writes the text and `:428` writes `"\r"` as a **separate** `client.write` on a PTY resolved once at `:418` — never `text + "\r"`. The `--wait` path does the same through two `session.deliver` calls at `:508` and `:518` and takes the **later** stamp (`:515`, `:525`) as the wait cursor, so the wait counts from the return.
- Proved on the wire: `packages/cli/src/__tests__/in-process-core-session.test.ts:302` asserts `writes` is `["2", "\r"]` against a real `PtyCoreLinkServer`, one fixture entry per write frame — two writes, not one glued string.

**Branch 2 — help text and exit path. MET on the exit path, PARTIALLY MET on the help text** (see the blocking item; the new section is right, an old sentence 5 lines below it is not).

- `NOT_SUBMITTED_WARNING` at `:597` fires on every `--no-enter` path — `:710` on `--wait` (before the wait, so the operator sees it before it hangs) and `:736` on the plain path, on stderr *after* the document under `--json` so the stdout-is-only-JSON rule holds.
- `session-command.test.ts` pins each of the four replacement statements by its own wording with a comment naming the finding it replaced, so none can drift back.

**`--no-enter` is a real opt-out. MET.**

- `packages/cli/src/cli-args.ts:237` parses it into `noEnter`; `session-command.ts:848` puts it in `SESSION_FLAGS` so other verbs reject it; `:649` accepts it on `send`.
- It reaches the wire as one write and nothing else: `in-process-core-session.test.ts:311-314` asserts the cumulative wire is `["2", "\r", "2", "\r", "2"]` after the `--no-enter` run and that stderr carries `started no turn`.
- `--enter --no-enter` is `EXIT_USAGE` at `:655-657` — before `readText` at `:665`, so it cannot consume stdin before failing, and before any dial.

**`--enter` is still ACCEPTED as a no-op. MET — checked hardest, since a rejection would break every orchestration caller.**

- `:649` keeps `--enter` in `send`'s accepted list, so `misusedFlag` (`:861-868`) never rejects it. Nothing else in the verb rejects it.
- `:690` does not consult it, so on a send that carries text every output derives from `submit` alone (`:701`, `:717`, `:726`, `:729`) — byte-identical with and without the flag. `session-command.test.ts:694` asserts `expect(without).toEqual(calls)` for the two command lines.
- On a send with **no** text it still does its old job: `:667` and `:674` refuse an empty send unless `--enter` was passed, after which `submit` is true and `gateway.send` skips the text write (`session-gateway.ts:423`) and writes `"\r"` alone. A bare carriage return, exactly as before.

### The four round-1 statements — all four corrected

**1. Flag table vs the warning — FIXED.** `session-command.ts:98`:

```
  --enter             send: only meaningful with no text — a bare carriage return
```

True against `:667`/`:674` (an empty send needs it) and `:690` (a send with text ignores it), and it no longer contradicts `NOT_SUBMITTED_WARNING`, which hands the operator `actana session send <id> --enter` — a send with no text, where the row says the flag *is* meaningful. `cli-args.ts:67-79` carries the same correction with **Not a no-op on a send with no text** as its own paragraph.

**2. "cannot swallow / cannot absorb" — FIXED.** Both strings are gone from the tree (the only remaining matches are `packages/sdk/src/terminal-screen.ts:417` and `.github/workflows/release.yml:728`, unrelated). `session-command.ts:155-161` now reads:

> Separate is necessary and not sufficient. ADR 0026's second observed failure is a return that was already its own write, 150 ms later, absorbed anyway by a harness rendering the text as a paste; what answers that is the length-scaled pause and the quiet gate on the **Core's** delivery path (ADR 0026 D6), which a `send` does not have and does not grow one here. A long enough `send` can still be pasted with its return eaten.

That matches ADR 0026 observation 2 and D6 exactly. The doc comment at `:609-616` says the same and adds why the gate must not be grown here.

**3. "Nothing is awaitable after it" — FIXED.** The string is gone from the tree. `session-command.ts:166-170`:

> **A `--wait` after it is not waiting for your text.** On an idle Session no turn ends, so the wait runs out the `--wait-timeout` — or, with none given, does not return. On a Session already mid-turn it resolves on *that* turn's end and reports it as this send's, for a turn this send did not start. Both are #405's and neither is fixed here.

Both cases are the real ones — the idle case follows from `waitTimeoutMs` returning `ms: null` with no `--wait-timeout` (`:881`), the busy case from the behaviour the same help documents at `:138`. "reports it as this send's" rather than "exits zero" is the right weakening: `awaitTurn` returns `EXIT_FAILURE` on a status that did not settle well (`:426`).

**4. `submitted` on the `--wait` document — FIXED, as the claim rather than the document.** `session-command.ts:172-175`:

> `--json` on a plain send carries the same fact as a `submitted` field. It is **not** on the `--wait` document, which prints `start --wait --json`'s keys and no others so one parser reads every verb (#289) — there the line on stderr is the only signal.

Verified against the document itself: `awaitTurn`'s `--json` branch at `:406-420` emits `{...startedFields(session), waited, status, exited, exitCode?, screen}` — no `enter`, no `submitted`. `session-command.test.ts:762-763` asserts both keys are `undefined` there. The PR body's unqualified sentence is withdrawn.

The two the developer found by re-reading the rest are real corrections too: the warning at `:597-599` no longer asserts where the text is or that Enter submits, and the `--wait` doc comment at `:639-640` no longer says "the same nothing appended".

### Blocking

**`packages/cli/src/session-command.ts:186` — `actana session --help` now contradicts itself.**

`SESSION_HELP` runs from `:76` to `:186` and is printed whole by `deps.out(SESSION_HELP)` at `:197`. At `:149` it now says:

```
  `send <session> <text>` **presses Enter** — that is, the text goes out and a
  carriage return follows it as its own separate write
```

and 37 lines later, in the same output, at `:186`:

```
  and \`send\` writes exactly what it is given — a lost prompt is a Core bug.
```

That sentence was true before this PR and is not true after it: `send` now writes what it is given **and a carriage return the operator did not type**. This is the same class of defect the four above were — a shipped operator-facing statement the code no longer supports — in the file #404's acceptance names explicitly, so I am holding it to the same bar rather than filing it as a nit. It reads as the authoritative summary, sitting under "Who delivers the prompt", which is where an operator goes to find out what this CLI does and does not add.

**`packages/cli/src/session-command.ts:25` — the module header says it unambiguously.**

```
// carriage return. Nothing in this package waits, retries, or presses Enter on
// a timer — `send` writes exactly the bytes it was given and appends nothing.
```

"appends nothing" leaves no room for the reading that might rescue `:186`, and it directly contradicts this PR's own `sessionSend` doc at `:604` ("**A send submits the turn**"). Maintainer-facing rather than shipped, but it is the file's charter and round 1's finding 2 already treated doc comments as in scope.

Both are one-line edits. The distinction the rest of the diff makes so carefully is the one to keep: this CLI adds no *timing* — no pause, no readiness wait, no retry — and it does now add a return, decided once from the flags the operator typed. `:186` could become something like "and `send` adds no timing of its own — a lost prompt is a Core bug", and `:25` could say "appends nothing on a timer" or name the return outright.

### Should fix, not blocking

**`packages/cli/src/session-gateway.ts:423-428` with `session-command.ts:737-740` — a partial delivery is reported as no delivery, now on the default path.**

`send` returns one boolean for two writes: `false` when the text write fails, and also `false` when the text landed and the `"\r"` write failed. Either way `:738` prints `actana session send: the Core did not accept the write to session <id>.` and exits non-zero. An operator or script whose text *did* land is told nothing was written, and the obvious next move — resend — now submits the doubled text, because the resend carries a return.

The `--wait` path already guards exactly this and says why: `session-gateway.ts:508-514` and `:518-523` raise distinct messages for the text and for the carriage return, under a comment reading "The text **was delivered** and the message says so, because the next thing an operator does with a failure here must not be to send it again." The plain path needs the same distinction. Before this PR the second write only existed under `--enter`, so this moves from opt-in to default with the rest of the change.

Related, same line: `:726` prints `submitted: submit` unconditionally, so a failed delivery emits `{..., submitted: true, delivered: false}`. `delivered` is in the same document and a correct script reads it, but `submitted` is the flag decision rather than an observed outcome, which is worth a word in its comment given how carefully the rest of this diff separates the two.

### Nits, no action required

- `packages/cli/src/session-command.ts:667` — `actana session send $SID --no-enter` with no text is refused with "pass text, `-` to read stdin, or **--enter** for a bare carriage return", and an operator who follows that literally lands on the contradiction guard at `:655`. The way out is to drop `--no-enter`, which the message does not say.
- `packages/cli/src/session-command.ts:911` — `readText` returns stdin verbatim, so `cat brief.md | actana session send $SID -` writes the file's trailing newline and then `"\r"` as a second write. Whether that is a double submit depends on how the harness treats a bare LF; unproven either way, but it is new on the default path.
- `packages/cli/src/session-command.ts:598` — the remedy line is built from the task id alone, so an operator who ran `--core prod` is handed a command for the default Core. Pre-existing convention (`:424` does the same for `session logs`) and not this PR's to fix, noted only for completeness.
- `packages/cli/src/session-command.ts:701` — `--no-enter --wait` is accepted although the help says its answer is another turn's. I do **not** think it should be refused: answering a dialog and then waiting on the turn already running is a real use, and #405 is the right home for the hang. Recording the judgement rather than asking for a change.
- `packages/cli/src/session-command.ts:98-99` are 81 characters; every other flag row is 78 or under, so these two are the only rows that wrap at 80 columns.
- `packages/cli/src/session-command.ts:726` — the `enter` key on the plain `--json` document now reports `true` where a script that passes no flag used to read `false`. Its documented meaning ("a carriage return was written") is unchanged; this is the break #404 asked for, and the comment above it says so.
- The PR body's "Review round 1" section credits the fixes to `6a8...`; the commit is `5ad201e`. Body text only.

### Skill and generated payload — clean

- `.agents/skills/actana-sessions/SKILL.md` and both `orchestration-skill-payload.ts` copies changed together. The two generated copies are **byte-identical** to each other, and re-running `scripts/gen-skill-payload.mjs` against the authored folder at this head in a scratch tree reproduces both committed copies exactly. No drift.
- The skill text describes only behaviour the code has. `SKILL.md:154-161` — a send presses Enter, the return is its own separate write, `--no-enter` sends no return, a `--wait` after it waits on nothing or on the already-running turn (both #405's), `--enter` is accepted and does nothing on a send with text and still means a bare return with none. Each maps to a line above. The loop example at `:129-130` correctly drops `--enter` and still works because the flag is accepted, not required. Nothing in the skill overstates the separate write the way the help used to.

### Boundaries and delivery — clean

- No file belonging to a parallel ticket. `packages/cli/src/events-command.ts` and `event-tip.ts` are untouched — #403 / PR 459 holds those, and its three files (`events-command.ts`, `events-command.test.ts`, `events-tail-cursor.test.ts`) have zero overlap with this diff. Nothing under `packages/core` or `packages/panel`. `packages/sdk/src/core-session.ts` is a doc-comment change only; `send()` is still the verbatim typing primitive the Panel types through.
- Base is `fix/operator-pins-sessions-drop-cli` — not `main`, not the train. Still a draft, as every PR in this wave is. The base branch has moved on by four Panel commits since the fork point, none touching a file in this diff.
- CI on `5ad201e` (run 33687228638): all 20 non-skipped checks pass, Unit Tests, Typecheck, Lint, Conventions and both installer E2Es included.

### Checklist

- [x] Acceptance branch 1 — a plain `send` submits without `--enter` — **met** (`session-command.ts:690`, `session-gateway.ts:423`/`:428`, wire test at `in-process-core-session.test.ts:302`)
- [~] Acceptance branch 2 — help text and exit path make the missing Enter impossible to miss — **partially met**: the exit path is met on every path; the help text's new section is right but `:186` contradicts it
- [x] `--no-enter` is a real opt-out — **met** (`cli-args.ts:237`, one write on the wire, warned on every path, `--enter --no-enter` refused before any dial)
- [x] `--enter` still accepted as a no-op on a send with text — **met** (`:649` accepts, `:690` ignores, byte-identical output asserted)
- [x] `--enter` still meaningful on a send with no text — **met** (`:667`, `:674`, `session-gateway.ts:423-428`)
- [x] Round-1 statement 1 (flag table) — corrected
- [x] Round-1 statement 2 (cannot swallow) — corrected
- [x] Round-1 statement 3 (nothing is awaitable) — corrected
- [x] Round-1 statement 4 (`submitted` on the wait document) — corrected, claim withdrawn from the body
- [ ] **No untrue operator-facing statement left in the help — `:186` and `:25` still say `send` appends nothing**
- [x] Generated payload copies match their source and each other — verified by regeneration
- [x] Skill text describes no behaviour the code lacks
- [x] No parallel-ticket file touched (`events-command.ts`, `event-tip.ts` clean)
- [x] Base branch, draft state and CI — correct

**Verdict: request changes.** The fix is correct, the tests are good, and all four round-1 findings are properly answered. Change the two sentences at `:186` and `:25` that this PR made false, consider splitting the partial-delivery report at `session-gateway.ts:424-428` the way the `--wait` path already does, and this is an approve.

https://claude.ai/code/session_012Do8ZsAciEKFwyDSCsdKh1

qcentic-dev:
Both blocking sentences are fixed, and I took the should-fix as well. Head is `0167a87`, plain push.

## Blocking 1 — `session-command.ts:186`, the help contradicting itself

You are right that it reads as the authoritative summary, and it was the same class of defect as round 1's four. It now keeps the distinction the rest of the diff makes rather than dropping the claim:

```diff
 Who delivers the prompt
   The Core does (ADR 0026). It waits for the harness to settle, answers the
-  dialog it opened, and writes the prompt. This CLI adds no timing of its own,
-  and `send` writes exactly what it is given — a lost prompt is a Core bug.
+  dialog it opened, and writes the prompt. This CLI adds no timing of its own —
+  no pause, no waiting for the harness to look ready, no retry — and `send`
+  adds nothing but the carriage return the flags above asked for. A lost prompt
+  is a Core bug.
```

The "no timing" claim is now spelled out rather than asserted, so it cannot be read as "adds nothing".

## Blocking 2 — `session-command.ts:25`, the module header

```diff
 // carriage return. Nothing in this package waits, retries, or presses Enter on
-// a timer — `send` writes exactly the bytes it was given and appends nothing.
+// a timer: `send` writes the bytes it was given, and — since #404 — the carriage
+// return that submits them, as its own write, because the flags the operator
+// typed asked for one. What this package never appends is *timing*.
```

It names the return outright and says what the charter actually forbids, so it no longer contradicts `sessionSend`'s own doc at `:604`.

A test pins both: `does not contradict itself 37 lines later (#462 review round 2)` asserts `writes exactly what it is given` is **gone** from the rendered help and that the replacement is there. `grep` confirms the phrase survives nowhere in the tree.

## Should fix — taken. The partial delivery now reports as one

You are right that this moved from opt-in to default with the rest of the change, and that the resend is the trap. `SessionGateway.send` returned one boolean for two writes; it now answers with a `SendResult`, because two writes have three outcomes and the third is the one a boolean cannot express:

```ts
export type SendResult =
  | { ok: true }
  | { ok: false; failed: "text" }            // nothing landed; a resend is safe
  | { ok: false; failed: "carriage-return" }; // the text landed; a resend doubles it
```

The verb gives each its own message, for the reason your quoted comment gives on the `--wait` path:

- **text refused** — `the Core did not accept the write to session <id>. Nothing was written, so sending it again is safe.`
- **return refused** — `session <id> took the text, but the Core did not accept the carriage return — so no turn was started. The text was delivered: do not send it again, or the harness gets it twice. \`actana session send <id> --enter\` sends the return on its own.`

One case your report did not have to name, which fell out of writing it: a **bare `--enter`** whose return is refused has no text to have half-landed, so "do not send it again" would be advice about text that does not exist. That gets the simple message and a safe-resend line, and has its own test.

## Related — `submitted` was the decision, not the outcome

Taken as more than a comment, because `{submitted: true, delivered: false}` is exactly the kind of statement this review has been about. The document now separates the two:

- `enter` — the **request**, unchanged in meaning and still the flag as the command resolved it, so an existing parser reads the same fact.
- `submitted` — the **outcome**: a return asked for *and accepted*.
- `failed` — present only on a failure, naming the half that went missing.

They agree on every path but the half-delivered one. The help paragraph on `--json` says so, and a test asserts all three shapes plus the absence of `failed` on success.

## The nits

No action, as you said, and I agree on `:701` — `--no-enter --wait` should stay accepted. Two I would rather not fold in here: `:667`'s message not mentioning that dropping `--no-enter` is the way out, and `:911`'s trailing newline from stdin, are both real but neither is a false statement, and this branch has grown enough. Happy to file either. The body's `6a8...` typo is corrected to `5ad201e`.

## Verification

CI on `0167a87`, run [33728048930](https://github.com/actana/control/actions/runs/33728048930): **18 checks, all pass, none failing** — Unit Tests 2m43s, Typecheck, Lint, Conventions, both installer E2Es, both tarball smokes, both image builds, Secret Scan, Dependency Audit, Train rules, Promotion gate.

Locally: **`packages/cli` 1205 passed / 3 skipped over all 50 files** (the machine finally quietened down, so this round has the clean full-suite run the last one could not get), `packages/sdk` 292 passed / 5 skipped, `packages/shared` 554 passed. `tsc --noEmit` clean on `cli` and `sdk`; `eslint` clean apart from the pre-existing `packages/shared/src/domain.ts` warning. `commitlint` over the whole range: 0 problems, 0 warnings.

`--enter` is still accepted as a no-op on a send that carries text — unchanged, still asserted byte-identical. Still a draft, base still `fix/operator-pins-sessions-drop-cli`, plain push, no force.

https://claude.ai/code/session_012k8r7PcE18ivj97pSooDBJ

mrfqcentic (approved):
## Review of #462 against #404 — round 3, approve

Reviewed at head `0167a87`, reading the enforcing lines rather than the PR description or the reply, in a scratch checkout of that commit and of the merge of it with the current base tip `c76cae5`. Round 1 blocked on four untrue operator-facing statements; round 2 confirmed all four fixed and blocked on two more of the same class (`session-command.ts:186` and `:25`) plus one non-blocking should-fix (a partial delivery reported as no delivery). **Both blocking edits are done, and the should-fix is taken and taken further than asked.** The mechanism has not moved since round 1 and is still right.

### The two blocking lines, quoted at this head

The line numbers moved with the edits, so both quotes below are the current content of those lines plus the sentence each now sits in.

**`packages/cli/src/session-command.ts:186` — now true.**

```
186   refusals of an empty send both point you at.
```

Line 186 is no longer the summary at all: it is the tail of the `--enter` paragraph. The sentence round 2 blocked on — "`send` writes exactly what it is given" — is **gone from the tree**, and the closing summary it belonged to now lives at `:188-193`:

```
188 Who delivers the prompt
189   The Core does (ADR 0026). It waits for the harness to settle, answers the
190   dialog it opened, and writes the prompt. This CLI adds no timing of its own —
191   no pause, no waiting for the harness to look ready, no retry — and `send`
192   adds nothing but the carriage return the flags above asked for. A lost prompt
193   is a Core bug.
```

True against the code: `submit = !args.noEnter` (`:697`) is the only thing that puts a return on the wire, `session-gateway.ts:456`/`:470` are the only two writes and there is no timer, no readiness check and no retry between or around them. "adds nothing but the carriage return the flags above asked for" is exactly what the verb does, and the claim it keeps — no *timing* — is now spelled out rather than asserted, so it cannot be read as "adds nothing".

**`packages/cli/src/session-command.ts:25` — now true.**

```
25 // a timer: `send` writes the bytes it was given, and — since #404 — the carriage
```

In its sentence (`:24-27`):

```
24 // carriage return. Nothing in this package waits, retries, or presses Enter on
25 // a timer: `send` writes the bytes it was given, and — since #404 — the carriage
26 // return that submits them, as its own write, because the flags the operator
27 // typed asked for one. What this package never appends is *timing*.
```

"appends nothing" is gone; the return is named outright, its separateness is named, and the charter is narrowed to the thing that is actually forbidden. It no longer contradicts `sessionSend`'s own doc at `:611` ("**A send submits the turn**"), and the two are now saying the same thing from two altitudes.

`git grep` confirms neither old phrase survives anywhere in shipped text: the only remaining matches for "writes exactly what it is given" / "appends nothing" are the SDK's own tests and doc header (`packages/sdk/src/__tests__/core-session.test.ts:14`, `:493`), where `CoreSession.send` genuinely still is the verbatim primitive, `packages/core/src/__tests__/session-delivered-event.test.ts:11`, `session-gateway.ts:518` where the phrase is qualified ("appends nothing the caller did not ask for"), and one unrelated hit in `scripts/__tests__/npm-publish.test.mjs`. A test pins both edits: `does not contradict itself 37 lines later (#462 review round 2)` asserts the old sentence is absent from the rendered help and both replacement clauses are present.

### `SESSION_HELP` read whole

`SESSION_HELP` runs `:78`–`:193` and is printed entire by `deps.out(SESSION_HELP)` at `:204`. I read every line of it against the code at this head. The Usage line (`:85`), the `--wait` row (`:94`), the `--enter` row (`:100`), the `--no-enter` row (`:101`), the awaiting-a-turn section (`:128-148`), the new sending section (`:150-186`) and the closing summary (`:188-193`) all agree with each other and with the verb. **No self-contradiction remains** — with one inaccuracy inside a single clause, which is a follow-up rather than a blocker and is set out below.

### Acceptance, box by box

> 1. `session send $SID "continue"` starts a turn without requiring `--enter`, **or** the help text and exit path make the missing Enter impossible to miss.

**Branch 1 — a plain send submits. MET.** `session-command.ts:697` — `const submit = !args.noEnter;`, one decision, consumed by the `--wait` path (`:710`) and the plain path (`:724`), so the two cannot disagree; `--enter` is deliberately not read into it. `session-gateway.ts:456` writes the text, `:470` writes `"\r"` as a **separate** `client.write` on a PTY resolved once at `:451` — never `text + "\r"`. Proved on the wire: `in-process-core-session.test.ts:302` asserts `writes` is `["2", "\r"]` against a real `PtyCoreLinkServer`, one fixture entry per write frame.

**Branch 2 — help text and exit path. MET** (round 2 had this partially met; the two sentences that held it back are the ones fixed above). The new section at `:150-186` is right, the closing summary no longer undoes it, and `NOT_SUBMITTED_WARNING` (`:604`) fires on every `--no-enter` path — `:717` on `--wait`, before the wait, and `:757` on the plain path, on stderr after the document under `--json` so the stdout-is-only-JSON rule holds.

**`--no-enter` is a real opt-out. MET.** `cli-args.ts:237` parses it, `session-command.ts:893` puts it in `SESSION_FLAGS` so every other verb rejects it, `:656` accepts it on `send`. One write on the wire and nothing else (`in-process-core-session.test.ts:311-314`). `--enter --no-enter` is `EXIT_USAGE` at `:662`, before `readText` at `:672`, so it cannot consume stdin before failing and no Core is dialled.

**`--enter` is STILL ACCEPTED as a no-op on a send with text. MET — re-checked hardest, since this is the one that would break callers.** `:656` keeps it in `send`'s accepted list so `misusedFlag` (`:906`) never rejects it; `:697` does not consult it; every output on that path derives from `submit` alone (`:708`, `:724`, `:743`, `:750`). `session-command.test.ts` runs the same command line with and without it and asserts `expect(without).toEqual(calls)`, and `in-process-core-session.test.ts:305-307` shows the wire is `["2","\r"]` again under `--enter`. Nothing in the repo's own orchestration text still passes `--enter` (SKILL.md dropped it), but every script written against the old default keeps working.

**`--enter` is still meaningful on a send with no text. MET.** `:674` and `:681` refuse an empty send unless `--enter` was passed; with it, `submit` is true, `gateway.send` skips the text write (`session-gateway.ts:456` is guarded on `text.length > 0`) and writes `"\r"` alone.

**The plain path now distinguishes a failed text write from a failed return write. MET.** `SessionGateway.send` answers with `SendResult` (`session-gateway.ts:240`, `:447-474`) — `{ok:true}`, `{ok:false, failed:"text"}`, `{ok:false, failed:"carriage-return"}` — and the verb gives each its own message at `:758-783`:

- text refused → "the Core did not accept the write to session `<id>`. Nothing was written, so sending it again is safe."
- return refused after a landed text → "session `<id>` took the text, but the Core did not accept the carriage return — so no turn was started. The text was delivered: do not send it again, or the harness gets it twice. `actana session send <id> --enter` sends the return on its own."
- return refused on a **bare** `--enter` (`:768`, guarded on `text.length === 0`) → the simple message with the safe-resend line, because there is no text to have half-landed. That case is the developer's own find and is the right call: the "do not resend" advice would otherwise be about text that does not exist.

Three tests pin the three messages, including `expect(err).not.toContain("Nothing was written")` on the half-delivered one — the assertion that matters, since that sentence is what was untrue.

**`submitted` no longer reads as an observed outcome when delivery failed. MET.** `:744` is `submitted: submit && sent.ok`, so it is false on both failures; `enter: submit` (`:743`) stays the request and keeps its old meaning; `failed` (`:746`) appears only on a failure. `keeps the request and the outcome apart in --json` asserts all three shapes plus the absence of `failed` on success. The `--wait` document is unchanged and still carries neither key (`session-command.test.ts` asserts both `undefined`), so #289's key-set rule holds.

### Follow-ups — none of them a merge blocker

This is the third round on a change I have found correct twice. What is left is one inaccurate clause and two rough edges in the failure reporting; none of them contradicts the fix, none of them is what #404 asked for, and none of them can produce the doubled-text outcome on the path the acceptance names. I would file them and merge rather than hold the branch a third time. Named here so they are not lost:

**1. `packages/cli/src/session-command.ts:176` (and the same words in the comment at `:732`) — "differ in exactly one case" is not quite true.**

```
175   that was **asked** for, `submitted` is the return that was **accepted**, and
176   they differ in exactly one case — the text landed and the return did not, where
```

They also differ on `failed: "text"`: `submit` is true so `enter` is `true`, `sent.ok` is false so `submitted` is `false`, and *nothing* landed. A script author who takes `enter !== submitted` as "the text is on the PTY, do not resend" will drop a send that was refused outright, where a resend is both safe and needed. The error is in the safe direction (a lost send, never a doubled one), the exit code is non-zero either way, the stderr line says "Nothing was written, so sending it again is safe", and the very next clause hands the script the key that *is* exact (`failed`). "differ whenever a write was refused, and `failed` names which half" would close it. If it is convenient to fold in, it is a two-word edit and does not need another round from me.

**2. `packages/cli/src/session-gateway.ts:562-569` — the `--wait` path's return refusal still gives no "do not resend".**

A refused `"\r"` after a landed text throws `the Core did not accept the carriage return for session <id>`. That is *distinct* from the text refusal at `:557`, which is what round 2 credited it for, but it does not say the text was delivered or that a resend doubles it — the plain path now does both. The comment at `:464-469` ("The `--wait` path has said this since #289") is reading the uncursored-delivery error at `:590-596`, which does say it, rather than this one. Pre-existing, and it moved to the default path with the rest of the change, exactly as the plain path's version did — same class as round 2's should-fix, and I am treating it the same way.

**3. `packages/cli/src/session-command.ts:745` — `delivered: false` on a half-delivered send.**

The document reads `{enter:true, submitted:false, delivered:false, failed:"carriage-return"}` while stderr on the same run says the text *was* delivered. `failed` disambiguates for a parser written after this PR; `delivered` is the pre-existing key and the help does not describe it. Before #404 a plain `--json` send could not reach this branch at all, so this is new on the default path. Worth a sentence in the help, or a rename, in the same follow-up as 1.

The round-2 nits stand as recorded and I am not re-raising them: `:678`'s message pointing a `--no-enter` caller at `--enter` rather than at dropping the flag, `:946`'s stdin trailing newline (`readText`), and the warning at `:606` building its remedy from the task id alone, ignoring `--core`, and `--no-enter --wait` staying accepted (still the right call).

### Delivery

- **Merges cleanly on the moved base.** Base tip is `c76cae5` (#463, #455, #470, #459, #468 landed since the fork point `2bdcb56`); `git merge-tree --write-tree` produces tree `9cb0f3f` with **no conflicts**. Zero file overlap: the base's CLI changes are `events-command.ts`, `events-command.test.ts`, `events-tail-cursor.test.ts` — #403's file lock — and everything else it moved is under `packages/panel`.
- **Suites re-run at `0167a87` in a scratch checkout:** `packages/cli` **1205 passed / 3 skipped over all 50 files**, `packages/sdk` 292 passed / 5 skipped, `packages/shared` 554 passed — the developer's numbers exactly. `tsc --noEmit` clean on `cli` and `sdk`.
- **Suites re-run on the merged tree:** `packages/cli` **1216 passed / 3 skipped**. One test failed on the first merged-tree run and passed on both re-runs; the run that failed took 71s against 37s for the clean ones, so I read it as load contention rather than this diff, and I did not capture its name.
- **Generated payload:** both `orchestration-skill-payload.ts` copies are byte-identical to each other, and re-running `scripts/gen-skill-payload.mjs` against the authored folder at this head reproduces both committed copies exactly. `SKILL.md:151-161` describes only behaviour the code has, and its loop example correctly drops `--enter`.
- **CI on `0167a87`** (run 33728048930): all 20 non-skipped checks pass — Unit Tests, Typecheck, Lint, Conventions, Secret Scan, Dependency Audit, Train rules, Promotion gate, both installer E2Es, both tarball smokes, both image builds. The four train-image jobs skip, as they do on a ticket PR.
- **Boundaries held:** nothing under `packages/core` or `packages/panel`; `events-command.ts` and `event-tip.ts` untouched; `packages/sdk/src/core-session.ts` is a doc-comment change only and `send()` is still the verbatim primitive the Panel types through. #405, #395 and #289 are named and not taken.
- Base is `fix/operator-pins-sessions-drop-cli` — not `main`, not the train. Still a draft, as this wave is.

### Checklist

- [x] Acceptance branch 1 — a plain `send` submits without `--enter` — **met** (`session-command.ts:697`, `session-gateway.ts:456`/`:470`, wire test at `in-process-core-session.test.ts:302`)
- [x] Acceptance branch 2 — help text and exit path make the missing Enter impossible to miss — **met** (was partial in round 2; `:186` and `:25` are the reason it now clears)
- [x] `session-command.ts:186` — the old summary is gone; the replacement at `:188-193` is true
- [x] `session-command.ts:25` — "appends nothing" is gone; `:24-27` names the return and forbids only timing, and no longer contradicts `:611`
- [x] `SESSION_HELP` (`:78-193`, printed whole at `:204`) contradicts itself nowhere
- [x] The plain path distinguishes a failed text write from a failed return write, with distinct messages, the way `--wait` does — plus a third message for a bare `--enter`
- [x] `submitted` is an observed outcome, false whenever the delivery failed; `enter` keeps its old meaning; `failed` names the missing half
- [x] `--enter` is STILL ACCEPTED as a no-op on a send that carries text — accepted at `:656`, ignored at `:697`, byte-identical output asserted, wire asserted
- [x] `--enter` is still meaningful on a send with no text
- [x] `--no-enter` is a real opt-out, warned on every path, and `--enter --no-enter` is refused before any dial
- [x] Merges cleanly on the moved base `c76cae5`, no file overlap with #463/#455/#470/#459/#468
- [x] Suites, typecheck and CI green; payload copies reproduce from source
- [x] Boundaries — no `core`, no `panel`, no parallel-ticket file
- [ ] Follow-up (not blocking) — `:176`/`:732` "differ in exactly one case", `session-gateway.ts:562-569` "do not resend" on the `--wait` path, `:745` `delivered` on a half-delivered send

**Verdict: approve.** #404's acceptance is met on both branches, both round-2 blockers are properly fixed, the round-2 should-fix is taken and the bare-`--enter` case the report did not name is handled too, `--enter` still works for every caller that passes it, and the branch still merges clean on `c76cae5`. The three items above are worth an issue against this file; none of them is worth a fourth round.

https://claude.ai/code/session_01Pds9aeJ7PFNegFuxm84eoB

Issue #401, opened 2026-09-02
B7: in grid view a file drop aimed at a session is intercepted by the project board


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `B7`
- Severity: likely
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: panel-rendering-investigation-2026-09-02.md (B7)

## What the operator sees

In grid view the session grid sits inside the board drop target. A drop aimed at a session’s terminal uploads to the Project root instead.

## Where

Same handler family as C (`projects.$id.tsx` board `onDrop`).

## What to do

Nested drop targets need an arbiter. A drop over a session/terminal must not also hit the board upload, unless that is the documented behaviour.

## Acceptance

1. A file drop over a session terminal in grid view does not silently upload to the project root (or does so only with an explicit overlay).

## Related

Same nested-target family as C. Can ride with C or follow it.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #477, merged to `beta/0.4.5` via #481.

Issue #399, opened 2026-09-02
G5: restoring the last archived Core session flashes an empty Active list


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `G5`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Core project → Archived → Restore the last session. View flips to Active with “No active sessions” until refresh lands.

## Where

`projects.$id.tsx` 1657–1685 restore omits insert into active cache; 1572–1580 archive omits `coreArchivedTasks` append. File lock with G1/G2/G4.

## What to do

Optimistic cache must insert the restored row into the active Core list (and the reverse on archive).

## Acceptance

1. Restoring the last archived Core session shows that row in Active without an empty flash.

Issue #398, opened 2026-09-02
G4: remember-settings optimistic write hits the wrong cache key on Core projects


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `G4`
- Severity: likely
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

On a Core project, change remembered harness, close, immediately New session. Dialog/skip uses the old harness until refetch.

## Where

`projects.$id.tsx` 2494–2498 `queryKeys.project(id)` vs core-tagged `projectQueryOptions`. File lock with G1/G2/G5.

## What to do

Write the same cache key the Core project query reads.

## Acceptance

1. Remembered harness on a Core project is what New session uses immediately, without waiting for refetch.

Issue #397, opened 2026-09-02
G1: archived session click is a no-op on Core projects


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `G1`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Core project → Archived → click a card or Reply. Nothing opens.

## Where

`projects.$id.tsx` 1375–1378 `selectTerminal` only searches active `tasks`, not `archivedTasks`. Same miss as G2.

## What to do

Resolve archived Core rows the same way active ones are resolved. File lock with G2/G4/G5 on `projects.$id.tsx`.

## Acceptance

1. Clicking an archived Core session opens/attaches that session.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #463, merged to `beta/0.4.5` via #481. The PR carried `Refs` rather than a closing keyword, so closing by hand.

PR #463, opened 2026-09-02
fix(panel): archived Core session click attaches

qcentic-dev:
Fixes the G1 hunt finding in #397.

## What the operator saw

Core project → Archived → click a card, or its **Reply** button. Nothing opened.

## Why

`selectTerminal` in `packages/panel/src/routes/projects.$id.tsx` resolved the clicked id
against the active `tasks` list only:

```ts
const task = tasks.find((t) => t.id === taskId);
if (!task || !terminalProject) return;
```

Where the clicked row can be found differs by owner (ADR 0019). A Panel-owned project's task
list already carries its archived rows, so that one lookup covers both views. A Core does not:
its archived rows live in their own list (`useArchivedTasks`), fetched over their own frame only
once the Archived view is open. So on a Core the lookup missed, the handler returned early, and
a row plainly on screen would not open. The Reply button routes through the same handler
(`TaskCard` calls `selectTask()`), so it was dead for the same reason.

## The change

The click path moves into a new `~/lib/open-clicked-session` helper, which falls back to the
archived rows when the active list has no match. Active rows still resolve out of `tasks`
first, so their behaviour is unchanged; archived rows now resolve the same way and open or
reattach like any other session.

Scope is deliberately just the open path. The restore-into-cache path, the remember-settings
write, and the board `onDrop` in this same file are untouched — they belong to #399, #398
and #401, which land on this file in later rounds.

## Tests

New `packages/panel/src/lib/__tests__/open-clicked-session.test.ts` (5 cases):

- an archived Core row, present only in `archivedTasks`, opens with the right task and `coreId`
- an active row still opens out of `tasks`, unchanged
- when both lists carry the id (Panel-owned project), the active row wins
- an id in neither list opens nothing
- nothing opens before the project path is ready

`packages/panel` typecheck is clean and ESLint reports no new problems on the touched files.
The two failures in `cores-settings-pairing.test.tsx` seen locally are pre-existing on the base
commit (verified by stashing this change) and unrelated to this route.

Base is `fix/operator-pins-sessions-drop-cli`, not `main` and not `beta/0.4.5`.

Refs #397

https://claude.ai/code/session_01DGk52AU9uCPWGvaePaGaf7

### DIFF
--- a/packages/panel/src/lib/__tests__/active-session-memory.test.ts
+++ b/packages/panel/src/lib/__tests__/active-session-memory.test.ts
@@ -1,0 +1,96 @@
+import { describe, expect, it } from "vitest";
+import {
+  activeSessionWentAway,
+  rememberActiveSession,
+  type LastActiveSession,
+} from "../active-session-memory";
+import type { Task } from "~/db/schema";
+
+const SCOPE = "p1";
+
+function task(id: string, over: Partial<Task> = {}): Task {
+  return { id, projectId: "p1", status: "ready", archived: false, ...over } as Task;
+}
+
+describe("rememberActiveSession", () => {
+  it("flags an archived Core row, which lives outside the active task list", () => {
+    const remembered = rememberActiveSession("t-archived", SCOPE, {
+      tasks: [task("t-live")],
+      archivedTasks: [task("t-archived", { archived: true })],
+      previous: null,
+    });
+
+    expect(remembered).toEqual({ projectId: SCOPE, taskId: "t-archived", archived: true });
+  });
+
+  it("flags a Panel-owned archived row, which carries the flag in the task list", () => {
+    const remembered = rememberActiveSession("t-archived", SCOPE, {
+      tasks: [task("t-live"), task("t-archived", { archived: true })],
+      archivedTasks: [],
+      previous: null,
+    });
+
+    expect(remembered.archived).toBe(true);
+  });
+
+  it("leaves a live row unflagged", () => {
+    const remembered = rememberActiveSession("t-live", SCOPE, {
+      tasks: [task("t-live")],
+      archivedTasks: [task("t-archived", { archived: true })],
+      previous: null,
+    });
+
+    expect(remembered).toEqual({ projectId: SCOPE, taskId: "t-live", archived: false });
+  });
+
+  it("keeps the archived verdict when a refetch drops the row from both lists", () => {
+    const previous: LastActiveSession = { projectId: SCOPE, taskId: "t-archived", archived: true };
+
+    const remembered = rememberActiveSession("t-archived", SCOPE, {
+      tasks: [task("t-live")],
+      archivedTasks: [],
+      previous,
+    });
+
+    expect(remembered.archived).toBe(true);
+  });
+
+  it("does not carry an archived verdict across a different id or scope", () => {
+    const previous: LastActiveSession = { projectId: SCOPE, taskId: "t-archived", archived: true };
+
+    expect(
+      rememberActiveSession("t-live", SCOPE, { tasks: [], archivedTasks: [], previous }).archived,
+    ).toBe(false);
+    expect(
+      rememberActiveSession("t-archived", "p2", { tasks: [], archivedTasks: [], previous }).archived,
+    ).toBe(false);
+  });
+});
+
+describe("activeSessionWentAway", () => {
+  it("reads a deselected archived row as a deselect, not a deletion", () => {
+    // Closing the panel on an archived Core session used to force-open an
+    // unrelated live one, with no coreId, onto a pane that never spawns.
+    const previous: LastActiveSession = { projectId: SCOPE, taskId: "t-archived", archived: true };
+
+    expect(activeSessionWentAway(previous, SCOPE, [task("t-live")])).toBe(false);
+  });
+
+  it("still reads a live row that left the visible list as a deletion", () => {
+    const previous: LastActiveSession = { projectId: SCOPE, taskId: "t-gone", archived: false };
+
+    expect(activeSessionWentAway(previous, SCOPE, [task("t-live")])).toBe(true);
+  });
+
+  it("reads a live row still on screen as a deselect", () => {
+    const previous: LastActiveSession = { projectId: SCOPE, taskId: "t-live", archived: false };
+
+    expect(activeSessionWentAway(previous, SCOPE, [task("t-live")])).toBe(false);
+  });
+
+  it("says nothing about a memory belonging to another scope", () => {
+    const previous: LastActiveSession = { projectId: "p2", taskId: "t-gone", archived: false };
+
+    expect(activeSessionWentAway(previous, SCOPE, [task("t-live")])).toBe(false);
+  });
+});
--- a/packages/panel/src/lib/__tests__/open-clicked-session.test.ts
+++ b/packages/panel/src/lib/__tests__/open-clicked-session.test.ts
@@ -50,8 +50,11 @@ describe("openClickedSession", () => {
 
   it("prefers the active row when both lists carry the id (Panel-owned project)", () => {
     const terminals = openerStub();
-    const fromTasks = task("t1", { archived: true });
-    const fromArchived = task("t1", { archived: true });
+    // Distinct `title`s, and an identity assertion on the argument: two
+    // structurally equal fixtures would pass whichever object the helper picked,
+    // leaving the tasks-first order the fix depends on untested.
+    const fromTasks = task("t1", { archived: true, title: "from tasks" });
+    const fromArchived = task("t1", { archived: true, title: "from archivedTasks" });
 
     openClickedSession("t1", {
       tasks: [fromTasks],
@@ -61,7 +64,12 @@ describe("openClickedSession", () => {
       terminals,
     });
 
-    expect(terminals.openSession).toHaveBeenCalledWith(project, fromTasks, { coreId: null });
+    expect(terminals.openSession.mock.calls[0]?.[1]).toBe(fromTasks);
+    expect(terminals.openSession).toHaveBeenCalledWith(
+      project,
+      expect.objectContaining({ title: "from tasks" }),
+      { coreId: null },
+    );
   });
 
   it("opens nothing for an id in neither list", () => {
--- a/packages/panel/src/lib/__tests__/open-clicked-session.test.ts
+++ b/packages/panel/src/lib/__tests__/open-clicked-session.test.ts
@@ -1,0 +1,97 @@
+import { describe, expect, it, vi } from "vitest";
+import { openClickedSession } from "../open-clicked-session";
+import type { Task } from "~/db/schema";
+import type { ScopedProject } from "~/lib/scoped-project";
+
+const project = { id: "p1", path: "/work" } as ScopedProject;
+
+function task(id: string, over: Partial<Task> = {}): Task {
+  return { id, projectId: "p1", status: "ready", archived: false, ...over } as Task;
+}
+
+function openerStub() {
+  return { openSession: vi.fn(), focusGridSession: vi.fn() };
+}
+
+describe("openClickedSession", () => {
+  it("opens an archived Core row, which lives outside the active task list", () => {
+    const terminals = openerStub();
+    const archived = task("t-archived", { archived: true });
+
+    const opened = openClickedSession("t-archived", {
+      tasks: [task("t-active")],
+      archivedTasks: [archived],
+      project,
+      coreId: "core-a",
+      terminals,
+    });
+
+    expect(opened).toBe(true);
+    expect(terminals.openSession).toHaveBeenCalledWith(project, archived, { coreId: "core-a" });
+    expect(terminals.focusGridSession).toHaveBeenCalledWith("t-archived");
+  });
+
+  it("still opens an active row out of the task list, unchanged", () => {
+    const terminals = openerStub();
+    const active = task("t-active");
+
+    const opened = openClickedSession("t-active", {
+      tasks: [active],
+      archivedTasks: [task("t-archived", { archived: true })],
+      project,
+      coreId: "core-a",
+      terminals,
+    });
+
+    expect(opened).toBe(true);
+    expect(terminals.openSession).toHaveBeenCalledWith(project, active, { coreId: "core-a" });
+    expect(terminals.focusGridSession).toHaveBeenCalledWith("t-active");
+  });
+
+  it("prefers the active row when both lists carry the id (Panel-owned project)", () => {
+    const terminals = openerStub();
+    const fromTasks = task("t1", { archived: true });
+    const fromArchived = task("t1", { archived: true });
+
+    openClickedSession("t1", {
+      tasks: [fromTasks],
+      archivedTasks: [fromArchived],
+      project,
+      coreId: null,
+      terminals,
+    });
+
+    expect(terminals.openSession).toHaveBeenCalledWith(project, fromTasks, { coreId: null });
+  });
+
+  it("opens nothing for an id in neither list", () => {
+    const terminals = openerStub();
+
+    const opened = openClickedSession("gone", {
+      tasks: [task("t-active")],
+      archivedTasks: [task("t-archived", { archived: true })],
+      project,
+      coreId: "core-a",
+      terminals,
+    });
+
+    expect(opened).toBe(false);
+    expect(terminals.openSession).not.toHaveBeenCalled();
+    expect(terminals.focusGridSession).not.toHaveBeenCalled();
+  });
+
+  it("opens nothing before the project path is ready", () => {
+    const terminals = openerStub();
+
+    const opened = openClickedSession("t-archived", {
+      tasks: [],
+      archivedTasks: [task("t-archived", { archived: true })],
+      project: null,
+      coreId: "core-a",
+      terminals,
+    });
+
+    expect(opened).toBe(false);
+    expect(terminals.openSession).not.toHaveBeenCalled();
+  });
+});
--- a/packages/panel/src/lib/active-session-memory.ts
+++ b/packages/panel/src/lib/active-session-memory.ts
@@ -1,0 +1,66 @@
+import type { Task } from "~/db/schema";
+
+/** What the board remembers about the session that was last active in a scope. */
+export type LastActiveSession = {
+  projectId: string;
+  taskId: string;
+  /** Whether that row was archived while it held the active slot. */
+  archived: boolean;
+};
+
+type RememberDeps = {
+  /** The project's active task list — a Panel-owned project's archived rows live here too. */
+  tasks: readonly Task[];
+  /** The rows the Archived view is showing. For a Core these are absent from `tasks` (ADR 0019). */
+  archivedTasks: readonly Task[];
+  /** What was remembered before, so an established `archived` verdict is not forgotten. */
+  previous: LastActiveSession | null;
+};
+
+/**
+ * Remember the session holding a scope's active slot, and whether it is archived.
+ *
+ * The flag is worked out here, while the row is active and the list it came from
+ * is loaded, rather than at deselect time: a Core's archived rows are fetched
+ * only while the Archived view is open (ADR 0019), so a later read of
+ * `archivedTasks` can no longer answer the question. Once an id is known to be
+ * archived it stays archived for as long as it holds the slot — a refetch that
+ * drops the row must not un-know it.
+ */
+export function rememberActiveSession(
+  taskId: string,
+  projectId: string,
+  deps: RememberDeps,
+): LastActiveSession {
+  const { tasks, archivedTasks, previous } = deps;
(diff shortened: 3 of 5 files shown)


qcentic-dev:
## Acceptance checklist — issue #397

- [x] **1. Clicking an archived Core session opens/attaches that session.**
      `selectTerminal` now resolves the clicked id through `openClickedSession`, which falls back to
      the Core's own archived-rows list when the active `tasks` list has no match. The **Reply**
      button is covered by the same fix — `TaskCard` routes it through `selectTask()` → the same
      handler. Proven by `open-clicked-session.test.ts`: an archived Core row present only in
      `archivedTasks` opens with the right task and `coreId`, and an active row still opens
      out of `tasks` unchanged.

### Scope held to the open path

This file is the head of a serial chain (#399, #398, #401 all edit it in later rounds), so only the
open path was taken. The restore-into-cache path, the remember-settings write and the board
`onDrop` were left exactly as they are.

### Verification

- `packages/panel` unit tests for the new file: 5 passed.
- Related existing route-logic suites (`task-display-order`, `archive-session`, `optimistic-task`,
  `active-group`, `mutate-task-for-core`, `components/views`): green apart from two
  `cores-settings-pairing.test.tsx` timeouts that reproduce on the base commit with this change
  stashed — pre-existing, unrelated to this route.
- `pnpm --filter @actana/panel typecheck`: clean.
- ESLint on the touched files: 0 errors (8 pre-existing unused-var warnings in the route).

mrfqcentic (changes requested):
## Review — PR #463 against #397 (G1)

Head `2c8bc6c`, base `fix/operator-pins-sessions-drop-cli` (`aa5df9d`, up to date). Reviewed from a clean worktree at the head; the panel suite and typecheck were run in a scratch copy.

**Acceptance criterion 1 — "Clicking an archived Core session opens/attaches that session" — is MET.** Traced through the enforcing lines, not the description:

- `packages/panel/src/routes/projects.$id.tsx:1376-1385` — `selectTerminal` now delegates to `openClickedSession`, and passes `archivedTasks` (derived at `:323` as `coreId ? archivedTasksQuery.data ?? [] : tasks.filter(t => t.archived)`).
- `packages/panel/src/lib/open-clicked-session.ts:39` — `tasks.find(...) ?? archivedTasks.find(...)`. On a Core the archived row is found in the second list and reaches the same `openSession(project, task, { coreId })` + `focusGridSession(taskId)` an active row gets.
- **Both click paths resolve.** The card body button (`TaskCard.tsx:84` `onClick={selectTask}`) and the **Reply** button (`TaskCard.tsx:349-351` → `selectTask()`) both call `onToggle` → `stableSelectTerminal` (`projects.$id.tsx:348`) → `taskCardHandlersRef.current.onToggle` (`:1703`) → `selectTerminal`. One fix covers both.
- **Panel-local archived rows are unchanged.** For `coreId === null`, `archivedTasks` is a subset of `tasks` (`:323`), so the `tasks` lookup still wins and nothing about that path moves.
- **Downstream composes.** `remoteTaskFromSnapshot` carries `archived: true` (`terminal-store.tsx:289/304`), so the archived-session reaper (`:929-953`) and the no-persist filter (`:459`) still recognise the opened row.
- **Scope is clean.** Exactly two hunks touch the route — the import and `selectTerminal`. No cache restore (#399), no remember-settings write (#398), no `onDrop` change (#401). `openSession` never touches the query client (`grep -c queryClient terminal-store.tsx` → 0), so the row is not restored into the active list.

Style matches the house pattern (`lib/archive-session.ts` + a `lib/__tests__` unit test), the ADR 0019 reasoning in the doc comment is accurate, and the commit body is well within commitlint's limits. CI on `2c8bc6c` is green (18 success, 6 skipped). Locally: 148/148 files, 1628/1628 tests pass; `tsc --noEmit` clean after `generate:routes`; ESLint reports 0 errors on the touched files (the 8 warnings are pre-existing unused imports elsewhere in the route).

That said, two things need to change before this is ready.

---

### 1. Blocking — `packages/panel/src/routes/projects.$id.tsx:548-559`: making archived Core rows openable makes this effect fire on them, and it lands the operator on a dead pane

```ts
555:    const visible = tasks.filter((t) => !t.archived);
556:    if (visible.some((t) => t.id === prev.taskId)) return;
557:    lastActiveRef.current = null;
558:    const next = pickByPriority(visible);
559:    if (next) toggleTerminalSession(terminalProject, next);
```

This effect exists to tell *deletion* from *deselect*, and it does so by asking whether the previously-active id is still in `visible`. `visible` never contains an archived row, so an archived active id always reads as "deleted".

Before this PR that state was unreachable on a Core — nothing could make an archived Core row the scope's active session. `openClickedSession` now can (`openSession` sets it via `setActiveByProject`, `terminal-store.tsx:778`). Concretely:

1. Core project → Archived → click a card. `activeTaskId` becomes the archived id, and `:550` records it in `lastActiveRef`.
2. Press `terminal.close` (or the panel close button) → `terminals.deselect(selectedScopeKey)` (`:1232`) → `activeTaskId` is null.
3. `:556` misses → the effect concludes the row was deleted and force-opens an unrelated live session at `:559`, yanking the operator out of the Archived view they were reading.
4. That call passes no `{ coreId }`, so `toggle` tags the new `OpenTerminal` `coreId: null` (`terminal-store.tsx:726`). On a Core, `getCorePtyBridge(null)` returns `null` (`panel-bridge.ts:292`) and, per the comment at `TerminalPane.tsx:893-896`, "the pane renders but never spawns" — a blank terminal, no output, no error.

So the gesture immediately after the one this PR fixes leaves a Core operator in a silent dead state. That is the same failure class #397 was filed about.

The `coreId`-dropping half of `toggleTerminalSession` is pre-existing (`:559`, and likewise `:1082`/`:1093` in `cycleSession`) and probably wants its own issue — but the path only becomes reachable on a Core because of this change, so the guard belongs here. Either record only non-archived ids into `lastActiveRef` at `:550`, or make `:556` archived-aware by also consulting `archivedTasks` before treating the id as deleted.

### 2. Blocking — `packages/panel/src/lib/__tests__/open-clicked-session.test.ts:51-64`: the precedence test is vacuous

```ts
53:    const fromTasks = task("t1", { archived: true });
54:    const fromArchived = task("t1", { archived: true });
...
64:    expect(terminals.openSession).toHaveBeenCalledWith(project, fromTasks, { coreId: null });
```

The two fixtures are structurally identical and `toHaveBeenCalledWith` is a deep-equality check, so this assertion passes whichever object the helper picks. Verified empirically: swapping the expectation to `fromArchived` leaves all 5 tests green, and inverting the operator in `open-clicked-session.ts:39` to `archivedTasks.find(...) ?? tasks.find(...)` also leaves all 5 green.

This is the one behaviour the PR description promises is unchanged ("Active rows still resolve out of `tasks` first"), and it is currently uncovered. Give the fixtures a distinguishing field (a different `title` or `status`) or assert identity via `terminals.openSession.mock.calls[0][1]).toBe(fromTasks)`.

The other four cases are sound — the archived-Core click and the unchanged active click, the two the issue actually needs, both discriminate correctly.

---

### Non-blocking

- **`projects.$id.tsx:565-570`** (low) — `activeByProject` is persisted (`terminal-store.tsx:642`) but archived sessions deliberately are not (`:459`). Reload with an archived Core session active and `activeTaskId` comes back as an id `tasks.find` cannot resolve: no panel is materialised, `:548` never fires because the id is non-null, and `cycleSession` bails at `idx === -1` (`:1088`), so session-cycling hotkeys are dead until a card is clicked. Newly reachable on Cores for the same reason as #1; on a Panel-owned project `tasks` still carries the row so it rehydrates. Worth a follow-up issue rather than a change here.
- **`projects.$id.tsx:1445`** — `deleteTask` has the identical `tasks.find` miss on a control that renders only on archived cards. Correctly left alone: that is #421 (G2), marked "Implementation priority: later". Noting it because `openClickedSession`'s fallback is exactly the shape #421 will want.

Scope discipline on #398/#399/#401 is exemplary and the commit message is genuinely good. Fix the two above and I expect a straight approve.

qcentic-dev:
Thanks — both blocking findings are addressed in `c21354c`, and the two follow-ups you asked for are filed.

## 5.1 — the archived-blind active-session-went-away effect · **fixed**

You were right that the guard belongs here: the effect was always archived-blind, but `openClickedSession` is what makes an archived Core row able to hold the active slot, so this PR is what makes the state reachable.

Taken your first option — record only what is safe to treat as deletable — with one addition. `lastActiveRef` now carries an `archived` flag, and the verdict is worked out **while the row holds the slot**, not at deselect time, because a Core's archived rows are fetched only while the Archived view is open (ADR 0019): read later, `archivedTasks` can no longer answer the question. The flag is also sticky for as long as that id stays active, so a refetch that drops the row cannot un-know it — the effect re-runs on every `tasks` change, and without stickiness one such re-run would silently re-arm the bug.

`activeSessionWentAway` then returns `false` for an archived row, so the deselect never reaches `pickByPriority` / `toggleTerminalSession` and the operator stays in the Archived view. A live row that genuinely went away is still replaced exactly as before.

The decision moved to `packages/panel/src/lib/active-session-memory.ts` (same house pattern as `archive-session.ts`) so it is testable; the route effect keeps its structure and gained `archivedTasks` in its deps. 9 new cases in `active-session-memory.test.ts` cover the archived-Core row, the Panel-owned archived row, the live row, the sticky verdict, non-carry across id/scope, and all four `activeSessionWentAway` branches.

**The `coreId`-dropping half is untouched here, as you suggested — filed as #466** (`:579`, `:1109`, `:1120`, all three `terminals.toggle` call sites), referencing #397, this PR and this review.

## 5.2 — the vacuous precedence test · **fixed**

Correct, and thank you for stating the experiment — I reproduced it. The fixtures now differ by `title`, and the assertion is an identity check on the call argument:

```ts
expect(terminals.openSession.mock.calls[0]?.[1]).toBe(fromTasks);
```

I ran your exact mutant to confirm it now bites: inverting `open-clicked-session.ts:39` to `archivedTasks.find(...) ?? tasks.find(...)` turns this test red. I mutation-checked the 5.1 guard the same way — deleting the `if (previous.archived) return false;` line turns its test red. Both guards are genuinely pinned rather than incidentally green.

## Non-blocking

- **Reload with an archived Core session active → dead cycling hotkeys** — not fixed here, per your call. **Filed as #467**, referencing #397 and this review, with the two exits you identified (`rehydrate` can't resolve the id; `cycleSession` bails at `idx === -1`).
- **`deleteTask` at `:1445`** — left alone; that is #421, and I agree `openClickedSession`'s fallback is the shape it will want.

## Verification

- `active-session-memory.test.ts` (9) + `open-clicked-session.test.ts` (5): all green, and both mutation-checked as above.
- `pnpm --filter @actana/panel typecheck`: clean.
- ESLint on the five touched files: 0 errors (the 8 warnings are the pre-existing unused imports you noted).
- `commitlint --from origin/fix/operator-pins-sessions-drop-cli --to HEAD`: 0 problems, 0 warnings across both commits. Plain push, no force.

Scope is unchanged otherwise — no #399 cache restore, no #398 remember-settings write, no #401 `onDrop`, nothing outside the route and the two `lib/` helpers. Still a draft, base still `fix/operator-pins-sessions-drop-cli`.

Follow-ups: #466, #467.

mrfqcentic (approved):
## Review round 2 — PR #463 against #397 (G1)

Head `c21354c`, base `fix/operator-pins-sessions-drop-cli`. Merge-base `aa5df9d`; the base tip has since moved to `9b71076` (ProjectBar / TerminalPane / use-fleet / session-write-store) with no overlap on this diff — GitHub reports `MERGEABLE` / `CLEAN`. Reviewed at the head from a read-only export; the suite, typecheck, ESLint and the mutants below were run in a scratch copy, not in the repo.

Both round-1 blocking findings are genuinely fixed, and I confirmed each by running the mutant rather than by reading the reply.

---

## Acceptance criteria — issue #397

**1. "Clicking an archived Core session opens/attaches that session." — MET.**

Traced end to end through the enforcing lines:

- `projects.$id.tsx:2326` — archived columns render with `onToggle={stableSelectTerminal}`, the same prop the active columns get. `TaskCard.tsx:84` (`onClick={selectTask}`) and the **Reply** button (`TaskCard.tsx:349-351`, shown for `needs-input`/`interrupted` regardless of `archived`) both call `onToggle` → `stableSelectTerminal` (`:353-354`) → `taskCardHandlersRef.current.onToggle` (`:1730`) → `selectTerminal`. One fix, both click paths.
- `projects.$id.tsx:1401-1409` — `selectTerminal` delegates to `openClickedSession` and hands it `archivedTasks`, derived at `:328` as `coreId ? archivedTasksQuery.data ?? [] : tasks.filter(t => t.archived)`.
- `open-clicked-session.ts:40` — `tasks.find(...) ?? archivedTasks.find(...)`. On a Core the row is found in the second list and reaches the same `openSession(project, task, { coreId })` + `focusGridSession(taskId)` an active row gets.
- It **attaches** rather than landing on a blank pane: `openSession` (`terminal-store.tsx:734-780`) puts the route's real `coreId` on the `OpenTerminal` and calls `markTaskRemote`, so `getCorePtyBridge(coreId)` (`panel-bridge.ts:289-292`) resolves a transport instead of returning `null`. This is not incidental — the store already anticipated the gesture: `archivedSessionsEligibleForReap` (`terminal-store.tsx:411-434`) documents "clicking an archived card resumes its PTY so the user can inspect history" and keeps the PTY alive while the row holds the active slot, reaping only after it is deselected. The fix restores behaviour the store was already written for.
- Panel-owned projects are untouched: for `coreId === null`, `archivedTasks` is a subset of `tasks` (`:328`), so the `tasks` lookup still wins.

That is the issue's only acceptance criterion, and it is met.

---

## Round-1 finding 1 — archived-blind went-away effect · **fixed**

`active-session-memory.ts:64` — `activeSessionWentAway` returns `false` when `previous.archived`, so `projects.$id.tsx:568-577` never reaches `pickByPriority` / `toggleTerminalSession` for a deselected archived row. The operator stays in the Archived view; no `coreId`-less pane is opened.

The design choice is right, and better than what I suggested: taking the verdict at `:558-562` **while the row holds the slot** is necessary, because a Core's archived rows are fetched only while the Archived view is open (ADR 0019) — read at deselect time, `archivedTasks` could no longer answer. The stickiness at `active-session-memory.ts:33-40` is likewise load-bearing, since the effect re-runs on every `tasks` change.

Mutation-checked, not taken on trust:

- delete `if (previous.archived) return false;` → `activeSessionWentAway > reads a deselected archived row as a deselect, not a deletion` **fails**.
- `alreadyKnown` → `false` → `rememberActiveSession > keeps the archived verdict when a refetch drops the row from both lists` **fails**.

I also checked the one path that could have kept the round-1 bug alive after the guard: reload with an archived Core session active leaves `lastActiveRef` flagged `archived: false` (the row is in neither list — that is #467). It is **not** exploitable, because `terminal.close` reads `terminals.activeFor(...)` (`:1256`), which is undefined when no session was materialised, so the handler takes the `lastHiddenSessionRef` branch and never deselects. And opening the Archived view re-runs the effect and self-heals the flag. Recording this so the next reader does not have to re-derive it.

## Round-1 finding 2 — vacuous precedence test · **fixed, and no longer tautological**

`open-clicked-session.test.ts:51-76` — the fixtures now differ by `title` and the assertion is `expect(terminals.openSession.mock.calls[0]?.[1]).toBe(fromTasks)`, an identity check.

I re-ran the exact experiment that made round 1's version vacuous: inverting `open-clicked-session.ts:40` to `archivedTasks.find(...) ?? tasks.find(...)`:

```
× prefers the active row when both lists carry the id (Panel-owned project)
Tests  1 failed | 4 passed (5)
```

Round 1 left all five green; it is now red. The test discriminates. For completeness, reverting the fix itself (`const task = tasks.find(...)` only) fails `opens an archived Core row, which lives outside the active task list`, so the G1 guard is pinned too.

## Deferred work — #466 and #467 accurately recorded

- **#466** (board-driven opens drop `coreId`) — line refs verified exact at this head: `:579`, `:1109`, `:1120` are the three `toggle` call sites that omit `{ coreId }`; `panel-bridge.ts:292` is the `if (!coreId) return null;`; `TerminalPane.tsx:893-895` is the "the pane renders but never spawns" comment. Acceptance criteria and the file-lock note are correct.
- **#467** (reload leaves an unresolvable active id) — the mechanism is right and matches what I traced above. Two cosmetic slips: it cites the rehydrate effect at `:565-570` and the `cycleSession` bail at `:1115`, which are pre-`c21354c` numbers — at this head they are `:591-597` and `:1116`. `:557` is correct. Worth a one-line edit to the issue, nothing more.

## Scope

Clean. `gh pr diff --name-only` is exactly five files: the two new `lib/` helpers, their two tests, and `projects.$id.tsx`. Within the route, three hunks — the import, the went-away effect, and `selectTerminal`. No cache-restore path (`restoreSession` `:1683` untouched, still using its own pre-existing `archivedTasks ?? tasks` fallback), no remember-settings write, no board `onDrop`. `ProjectBar.tsx`, `use-fleet.ts` and `user-terminal-store.tsx` are not touched. #399 / #398 / #401 are unblocked.

## Verification

- Full `packages/panel` suite: **150 files, 1645 tests, all pass** (the two `cores-settings-pairing` timeouts from round 1 did not reproduce).
- `generate:routes` + `tsc --noEmit`: clean.
- ESLint on the five touched files: 0 errors, 8 warnings — all pre-existing unused imports/vars elsewhere in the route (`:3`, `:95`, `:96`, `:125`, `:179`, `:238`, `:240`, `:517`), none introduced here.
- CI on `c21354c`: 18 pass, 6 skipped.
- Commit messages: two commits, both explain the mechanism and the ADR 0019 reasoning; the round-2 message names the mutants it was checked against. ADR 0019 corroborates every claim the code comments make about the two-list split.

---

## Non-blocking — none of these gate this PR

**1. `active-session-memory.ts:33-40` — the sticky flag survives a Restore, so one delete stops auto-advancing.**

`restoreSession` (`:1683-1723`) flips `archived` in the cache but never touches `activeByProject`, so a row restored while it holds the slot stays active — and `alreadyKnown` keeps its `archived: true` verdict for the whole tenure. Sequence: click an archived card → **Restore** → delete that (now live) session. `activeSessionWentAway` returns `false`, so no replacement is opened, contradicting the effect's own contract. Pre-PR this sequence did auto-advance on a Panel-owned project, so it is a small regression — but it costs a convenience, not correctness: no dead pane, no wrong `coreId`, and the next click clears it.

The fix is one line and one case: when `tasks` carries the id with `archived === false`, drop the sticky verdict rather than OR-ing it in. Your call whether that belongs here or in a follow-up.

**2. `projects.$id.tsx:583` — the effect now re-runs on every render.**

`archivedTasks` (`:328`) is a fresh array identity each render: `tasks.filter(...)` for a Panel-owned project, and `archivedTasksQuery.data ?? []` whenever the Core query has no data — which is the normal Active view. Adding it to the deps therefore turns a previously-stable effect into a per-render one, in a component that re-renders hard while an agent streams task events. The body is idempotent and the force-open path clears `lastActiveRef` first, so there is no loop and nothing visibly breaks; it is per-render `find`/`some` work. A `useMemo` on `:328` restores the old churn profile.

**3. `open-clicked-session.ts:50`** — the `boolean` return is asserted in tests but dropped at the only call site (`:1402`). Fine as documentation of intent; noting it so it is not mistaken for a missed check.

**4. Behaviour change worth a line in the PR body.** Deselecting an archived **Panel-owned** session no longer force-opens a replacement either. That is deliberate (`rememberActiveSession` reads the flag off `tasks` too, and `active-session-memory.test.ts:24-32` covers it) and I think it is correct — a deselect is a deselect for both owners — but the description only claims the Core path changed.

**5. `deleteTask` `:1471`** — still `tasks.find` only, while Delete renders only on archived cards (`:2329`), so it is dead on an archived Core row. Correctly left alone: that is #421 (G2), which #397 itself names as the same miss and marks lower priority, and your reply says as much. `openClickedSession`'s fallback is the shape #421 will want.

---

## Checklist

- [x] AC1 — clicking an archived Core session opens/attaches — **met**, proven at the enforcing lines, including that it attaches with the right `coreId` rather than opening a blank pane
- [x] Round-1 §1 — archived deselect no longer force-opens anything — **fixed**, mutation-proven
- [x] Round-1 §2 — precedence test is no longer tautological — **fixed**, the exact round-1 mutant now goes red
- [x] #466 and #467 accurately record what was deferred (two cosmetic line-ref slips in #467)
- [x] Scope — 5 files, OPEN path only; no cache/restore/drop path, no `ProjectBar.tsx` / `use-fleet.ts` / user-terminal store
- [x] Tests 1645/1645, typecheck clean, ESLint 0 errors, CI green
- [ ] Five non-blocking notes above, none gating

**Verdict: approve.** The issue's acceptance criterion is met, both blocking findings are fixed and independently verified by mutation rather than by assertion, the deferrals are filed accurately, and scope discipline on the serial chain is exact. The five notes are follow-up material — item 1 is the only one I would actually like to see land somewhere.

Issue #394, opened 2026-09-02
T3: project VM shell identity is lost on reload


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `T3`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

New Terminal on a project, reload. Pane gone from the project; Home may spawn the wrong shell kind / cwd.

## Where

`user-terminal-store.tsx` (catalog: 50–61) `shellSession` in-memory only; `UserTerminalPane` effect deps omit `coreId`/`shellSession`.

## What to do

Persist enough identity (project, core, kind, cwd) to restore the same shell after reload, or show it gone on purpose instead of spawning a different one from Home.

## Acceptance

1. Reload keeps the same user-terminal on the same project, or clearly has none — not a different shell.

mrfqcentic:
Delivered in **0.4.5**. Fixed by #468, merged to `beta/0.4.5` via #481. The PR carried `Refs` rather than a closing keyword, so closing by hand. The reload-orphan residue is filed as #473.

PR #468, opened 2026-09-02
fix(panel): project VM shell identity survives reload

qcentic-dev:
Closes the T3 hunt in #394: **a reload keeps the same user terminal on the same project, or clearly has none — never a different shell.**

## What was wrong

New Terminal on a project, reload: the pane was gone from the project, and Home — the one scope that did reload its rows — could spawn what was left as a different shell in a different place.

A user terminal's row lives in `home_terminals` whatever scope it was opened in (#266), and that row carries a name, a cwd column and nothing else. Everything that says *which* shell it is — the project it was opened on, the Core its PTY runs on, whether it is a VM Shell Session, the cwd it opens at — was in memory only. So a reload had nothing to restore a project's terminal from, and the home loader, which did reload, could only default: `{ terminal, ptyId: null }`, no kind, so the pane spawned it with the `home` flag instead of `shellSession: true`.

## What changed

- **`packages/panel/src/lib/user-terminal-identity.ts` (new)** — the persisted identity: terminal id → `{ scopeKey, coreId, kind, cwd }`, in localStorage under `mc.userTerminalIdentity`, with a parser that drops anything malformed and the pure `restoreUserTerminals` grouping.
- **`user-terminal-store.tsx`** — writes an identity when a terminal is opened, drops it when one is killed, prunes it against the server's list. The home-only loader becomes one restore that runs once per app run as soon as any scope is current and hands each row back to the bucket its identity names, as the kind that identity records. A `Session` now carries `kind` and `cwd` instead of the in-memory-only `shellSession` flag.
- **`UserTerminalPanel.tsx`** — reads kind and cwd off the session rather than off the ambient scope (`isHome` was the current-scope flag: the same defaulting, one level down).
- **`UserTerminalPane.tsx`** — the surface build key and the effect deps now include the Core and the kind. Those pick which spawn the pane makes and which machine it makes it on, so leaving them out let a pane that first rendered without a Core, or before a restored session's kind was known, keep whatever shell that first render started. The same line also loses a stray NUL byte that was committed inside its template literal.

**A row with no identity is deliberately restored as nothing** — an older build, or a browser whose storage was cleared. Defaulting is exactly what put a project's shell on Home as a home shell; #394 asks for the opposite, which is to show it gone on purpose rather than to spawn a different one.

**Not in scope, and what it costs (review finding F3, tracked as #473).** The PTY id is still not persisted, so a restored terminal spawns a fresh shell of the same kind on the same Core rather than re-attaching to the process that survived the reload. #394 asks for project, Core, kind and cwd, which is exactly what is persisted.

The consequence, stated plainly rather than left implied: a browser reload does not drop the Panel↔Core link, so the pre-reload PTY is still alive on the Core, and nothing can reach it — `findByTask` refuses to return shell/`shellSession` PTYs, and `killTerminal` only kills the `ptyId` the store currently holds, which after a restore is the replacement. **So every reload leaves one live, unreachable login shell per restored terminal on the Core's machine.** Before this change a project terminal was not restored at all, so this is new: it is the price of restoring the terminal rather than losing it.

It is not fixed here because both fixes are changes of their own with their own failure modes — re-attaching needs a replay that fails safe when the pty is gone (a dead `ptyId` would strand the pane behind a Retry that can only fail again), and killing the recorded PTY on restore would kill the shell a *second open tab* is still attached to. Recorded in full, with what would close it, as **#473**.

## Tests

- `user-terminal-reload.test.tsx` — drives the real store through a real reload (provider unmounted, a fresh one mounted over the same localStorage, the API returning the rows the Panel still has): the same project, Core, kind and cwd come back; Home and a second project get nothing; a Home terminal stays a VM shell; an unidentifiable row restores as nothing; a killed terminal is forgotten; a failed list call leaves the map alone and retries.
- `user-terminal-identity.test.ts` — the pure half, including that an entry which does not parse is treated like a missing one.
- `user-terminal-pane-kind.test.tsx` — the props the panel hands each pane, which is where the kind was being lost.

Every user-terminal suite, old and new, is green: **7 files / 49 tests**. `pnpm -r typecheck` and `pnpm lint` are clean (0 errors).

*(The earlier "7 files / 37 tests" in the first checklist comment was miscounted — it ran `session-warm-pool.test.ts` as the seventh file instead of `home-terminals.test.ts`. The reviewer's count of 44 on `5712b94` was right; the five cases added in this round bring it to 49.)*

## Review round 1 (`5712b94` → `100fd2c`)

`29a3d87` and `100fd2c` answer the review:

- **F1 (blocking) — the restore could cancel itself off permanently.** The cancel is gone rather than repaired: what the restore writes is decided by each row's identity, not by whichever scope was current when the request went out, so a navigation mid-flight cannot change the right answer. The setters are idempotent under their own guards and a setState after unmount is a no-op. A new case holds the list promise open, navigates to another project and then answers it; it fails against the reviewed code.
- **F2 — a prune could delete an identity that was just written.** `pruneIdentities` now takes the candidate set the answer is evidence about (the ids that existed when the request was issued), and `commitIdentityChange` re-reads storage and applies this tab's delta instead of overwriting the shared map with a per-tab snapshot.
- **F3 — the orphaned PTY.** Recorded above and filed as #473.
- **F4 — a whole bucket was dropped on collision.** Restore merges by terminal id.
- **Nits.** The `buildKey` separator is now U+0001, written as an escape so the source stays plain text; the `UserTerminalPanel` comment about a restored session carrying no Core is corrected; the test count is corrected above.

## Base

Stacked on `fix/operator-pins-sessions-drop-cli`, not `main` and not `beta/0.4.5`.

Refs #394

https://claude.ai/code/session_013uDNQQMaprupS64Yu2KJq4


qcentic-dev:
## Acceptance checklist — #394

> **1. Reload keeps the same user-terminal on the same project, or clearly has none — not a different shell.**

- [x] **Reload keeps the same terminal on the same project.** Identity — `{ scopeKey, coreId, kind, cwd }` — is persisted when the terminal is opened and restored into the bucket it names, as the kind it records. Proven by `user-terminal-reload.test.tsx` → *"restores it to that project, on the same Core, as the same kind and cwd"*, which asserts the terminal id, `kind: "vm-shell"`, `coreId: "core_a"` and `cwd` after a real unmount/remount over the same localStorage, and that no second terminal was created.
- [x] **…or clearly has none.** A row whose identity is missing or malformed is restored as nothing rather than defaulted into Home. Proven by *"shows a row it cannot identify as gone rather than as some other shell"* and by the parser cases in `user-terminal-identity.test.ts`.
- [x] **Never a different shell.** Three separate ways it used to become one are pinned:
  - *"does not hand a project's terminal to Home as a different shell"* — after a reload landing on the dashboard, Home's bucket is empty and the shell is still the project's, still a VM shell.
  - *"keeps a Home terminal a VM shell on Home, not a home shell"* — the exact regression the old loader caused (no kind → the `home` flag instead of `shellSession: true`).
  - *"does not hand it to a different project either"*.
- [x] **Never a different shell — at the render layer too.** `user-terminal-pane-kind.test.tsx` asserts the panel hands each pane the session's own kind, Core and cwd rather than the ambient scope's; `isHome` was the current-scope flag, which is the same defaulting one level down.
- [x] **Existing user-terminal tests stay green.** `user-terminal-store.test.ts`, `user-terminal-panel-control.test.tsx` and `drop-legacy-user-terminals.test.ts` pass unchanged — 7 files / 37 tests across the whole user-terminal area.

### Deliberately **not** checked

- [ ] **Re-attach to the PTY that survived the reload.** A restored terminal spawns a fresh shell of the same kind on the same Core; it does not reclaim the process still running there. Not an acceptance criterion — #394 asks for project, Core, kind and cwd — and doing it needs a replay path that fails safe when the pty is gone, which is a change of its own with its own failure mode (a dead ptyId would strand the pane behind a Retry that can only fail again).
- [ ] **Clicked on pairdemo.** The hunt filed this as code-read, and this branch was verified by tests plus `pnpm -r typecheck` and `pnpm lint` (0 errors), not by a browser.

### CI

Full-suite verification here was not possible: this machine is running at load average ~80–120 on 8 cores with no free memory, and whole-suite runs fail with vitest worker start-up timeouts on files this branch does not touch (the same suites pass in isolation, on this branch and on its base). The targeted user-terminal suites, typecheck and lint are green locally; CI on this PR head is the arbiter, and its status is reported alongside this comment.

qcentic-dev:
### Note for reviewers: `UserTerminalPane.tsx` renders as a binary diff

GitHub will show that file as *"Binary files differ"* rather than as a line diff. The cause is on the **base** side: the committed `buildKey` line contains a literal NUL byte inside its template literal —

```
const buildKey = `${cwd}\0${retryNonce}`;
```

— so git classifies the pre-image as binary. This branch's version is clean text, and diffs of that file after this lands will render normally.

The whole change to that file, via `git diff --text`:

```diff
@@ -151,10 +151,17 @@ export function UserTerminalPane({
   useEffect(() => {
     const cache = terminalSurfaceCache;
     const surfaceId = terminal.id;
-    // A change to cwd or the retry nonce means "build a fresh terminal"; a plain
-    // remount (scope switch, navigation, un-hide) keeps the same buildKey and
-    // reattaches the existing surface instantly.
-    const buildKey = `${cwd}<NUL>${retryNonce}`;
+    // A change to the spawn inputs — the Core the shell runs on, which kind of
+    // shell it is, its cwd — or to the retry nonce means "build a fresh
+    // terminal"; a plain remount (scope switch, navigation, un-hide) keeps the
+    // same buildKey and reattaches the existing surface instantly.
+    //
+    // Core and kind belong in that key (issue 394). They pick which spawn this
+    // pane makes and which machine it makes it on, so leaving them out let a
+    // pane that first rendered with no Core — or before a restored session's
+    // kind was known — keep the shell that first render started: the wrong one.
+    const kind = shellSession ? "vm-shell" : isHome ? "home" : "project";
+    const buildKey = `${coreId ?? ""} ${kind} ${cwd} ${retryNonce}`;
     const container = containerRef.current;
     if (!container) return;
 
@@ -504,7 +511,7 @@ export function UserTerminalPane({
       cancelled = true;
       detachMount?.();
     };
-  }, [terminal.id, cwd, retryNonce]);
+  }, [terminal.id, coreId, isHome, shellSession, cwd, retryNonce]);
```

(`<NUL>` above stands in for the byte itself, which cannot be pasted into a comment.)

qcentic-dev:
### CI on this PR head (`5712b94`) — all green

Run [33686670666](https://github.com/actana/control/actions/runs/33686670666): **18 required checks pass, 0 fail, 0 pending.**

| Check | Result |
| --- | --- |
| Unit Tests | pass (2m15s) |
| Typecheck | pass |
| Lint | pass |
| Conventions (commitlint: PR title + every commit) | pass |
| E2E — installer (debian, x64) / (ubuntu, x64) | pass |
| Smoke — Core release tarball (linux-x64) / (linux-arm64) | pass |
| Core image / Panel image (build + smoke, amd64) | pass |
| Dependency Audit, Secret Scan, Train rules, Promotion gate, Resolve PR image mode | pass |

`Unit Tests` passing in CI also settles the caveat in the checklist above: the whole-suite failures seen locally were this machine (load average ~80–120 on 8 cores, no free memory) timing out vitest workers, not this branch.

The train-image jobs are `skipping`, as expected for a ticket PR that is not a train build.

mrfqcentic (changes requested):
## Review — PR #468 against #394 (T3: project VM shell identity is lost on reload)

Reviewed as a diff against `fix/operator-pins-sessions-drop-cli`, read with `git diff --text` (see the binary-diff note below), with the three new suites and the four pre-existing user-terminal suites run locally on `5712b94`.

**Verdict: request changes.** The design is right, the scope is clean, and the acceptance criterion is met by the enforcing lines — but the restore effect can permanently cancel itself for an app run, which silently returns the operator to exactly the state #394 describes. That is a ~2-line fix plus a test, so this is a small ask, not a rework.

---

### Acceptance — #394, criterion 1

> *Reload keeps the same user-terminal on the same project, or clearly has none — not a different shell.*

Checked at the enforcing lines, not from the description:

| Property | Verdict | Enforcing line |
| --- | --- | --- |
| Identity `{scopeKey, coreId, kind, cwd}` written when the terminal opens | **met** | `user-terminal-store.tsx:375-381` — built and `setIdentities` before `updateSessions` at `:382`, i.e. before the pane is shown |
| Read back into the **same bucket** on reload | **met** | `user-terminal-identity.ts:119-121` groups strictly by `identity.scopeKey`; `user-terminal-store.tsx:253-265` writes `next[key]` for exactly those keys |
| Restored as the **same kind** | **met** | `user-terminal-store.tsx:262` (`kind: identity.kind`) → `UserTerminalPanel.tsx:467-468` |
| Missing/malformed identity restores as **nothing**, never Home, never another project | **met** | `user-terminal-identity.ts:120` (`if (!identity) continue;`), plus `:49-58` and `:66-73` dropping malformed entries. There is no fallback bucket anywhere in the restore path |
| A Home terminal that was a VM shell **stays a VM shell** | **met** (see caveat) | kind is recorded as `vm-shell` at `user-terminal-store.tsx:378` regardless of scope; `UserTerminalPanel.tsx:468` derives `shellSession` from the session's kind, `:467` derives `isHome` from it too |
| Pane render layer takes the **session's** kind/Core/cwd, not the ambient scope | **met** | `UserTerminalPanel.tsx:465-468`; `sessionCoreId = s.coreId ?? coreId` at `:435`. Restored sessions always carry a Core (`createVmShellTerminal` refuses without one at `:365`), so the route fallback never overrides a recorded Core |
| **"or clearly has none"** | **partially met** — see F1: there is a path where *every* scope ends up with none, for reasons unrelated to identity |

Caveat on the Home row, for the record rather than as a finding: `setHomeActive` has no caller in app code on this branch (only the store and the new test — `grep -rn setHomeActive packages`), so `homeActive` is always `false`, `HOME_SCOPE_KEY` is unreachable, and `active = !!project || homeActive` (`UserTerminalPanel.tsx:74`) means the panel never renders on the dashboard. That predates this PR, but it does mean the PR body's "Home — the one scope that did reload its rows — could spawn what was left as a different shell" describes a path the shipped UI cannot enter, and the `loadOnHome` cases pin store behaviour rather than an operator-reachable flow. The *primary* symptom in #394 ("Pane gone from the project") is real and is fixed.

---

### Blocking

**F1 — `packages/panel/src/lib/user-terminal-store.tsx:241-287` (esp. `:244`, `:250`, `:284-286`): the restore can cancel itself permanently.**

```ts
if (restoreStartedRef.current) return;
restoreStartedRef.current = true;      // :244 — set before the fetch
...
if (cancelled) return;                 // :250
...
return () => { cancelled = true; };    // :284-286 — does not reset the ref
```

`scopeKey` changing while `api.listHomeTerminals()` is in flight runs the cleanup (`cancelled = true`), then re-runs the effect, which bails at `:243` because the ref is still `true`. The in-flight response is then dropped at `:250`. Nothing restores, nothing is pruned, and no later scope change retries — for the rest of the app run. Only another reload recovers.

Reachable: land on project A on a cold Panel, click project B before the list responds. The result is an empty terminal panel in every scope, which is the #394 symptom this PR exists to remove.

This also defeats the retry contract the code documents at `:281-282` and that the test at `user-terminal-reload.test.tsx:217-239` relies on: that test exercises the **rejected** call, which does reset the ref at `:283`; a merely **slow** call takes the opposite branch and blocks the retry forever. Worth noting that the previous code had the same shape but keyed on `homeActive`, which flips once — moving the dependency to `scopeKey`, which changes on every navigation, is what makes it reachable.

Suggested: reset `restoreStartedRef.current = false` in the cleanup when the run has not completed (or drop the `cancelled` gate on the state writes, since all three setters are already idempotent under their `prev[key]` guards). Please add a case that changes `scopeKey` while the list promise is pending — the current suite cannot catch this.

---

### Please address or explicitly record

**F2 — `user-terminal-store.tsx:276-277`: `pruneIdentities` can delete an identity that was just written.**

`liveIds` comes from the list snapshot, but the prune is applied after the await. An identity written at `:381` during that window is not in `liveIds` and is deleted, while the server row survives — so on the next reload that row has no identity, is deliberately not restored (`user-terminal-identity.ts:120`), and its row is never cleaned up either. The same-tab window is narrow; the cross-tab version is not. `identities` is persisted as a whole map from a per-tab snapshot (`:171-176`), so any tab's `setIdentities` write clobbers an entry another tab created after this tab mounted. `hiddenIds`/`panelOpen` share that pattern, but losing an identity has a worse consequence than losing a hidden flag. Prune only ids that were present in the map when the request was issued, and consider re-reading storage before writing.

**F3 — the "no re-attach" decision is defensible and honestly stated, but the orphan it leaves is not mentioned.**

The decision itself is fine and I would not block on it: #394's "What to do" asks for *project, core, kind and cwd*, and that is exactly what is persisted. The PR states plainly, in the body and in the checklist comment, that a fresh shell of the same kind is spawned instead, with the reason (a dead `ptyId` would strand the pane behind a Retry that can only fail). Restoring the row, the project, the Core and the kind satisfies "keeps the same user-terminal on the same project" in the operator's sense; scrollback and process continuity are not in the criterion.

What is missing is the consequence. Restored sessions get `ptyId: null` (`:259`), so each pane spawns a new PTY while the pre-reload PTY is still alive on the Core — a browser reload does not drop the Panel↔Core link, `findByTask` explicitly refuses to return shell/`shellSession` PTYs (`packages/core/src/pty-manager.ts:851`), and `killTerminal` only kills the `ptyId` the store currently holds. Before this change a project terminal was not restored at all; now every reload adds one more unreachable live shell per restored terminal. Please either kill the recorded PTY when a restored terminal spawns its replacement, or record the accumulation explicitly in the PR and a follow-up issue rather than leaving it implied by "the PTY id is still not persisted".

**F4 — `user-terminal-store.tsx:255`: `if (prev[key]) continue;` drops a whole bucket, not just the collision.**

If the operator opens a terminal in the current scope before the list resolves, that bucket already exists and every pre-reload terminal for that scope is silently dropped from this app run. Identities survive, so they reappear on the next reload — the panel just under-reports until then. Merging by terminal id would be strictly better than skipping.

---

### Nits (non-blocking)

- **`UserTerminalPanel.tsx:433-434`** — the comment "A session restored from the API carries no Core; it runs on whichever one this route is on" is now false: restore pins `coreId` from the identity at `:260`. The code at `:435` is right; the comment contradicts it.
- **`UserTerminalPane.tsx:164`** — the NUL removal is correct and semantically neutral (see below), but the new separator is a plain space, and a cwd can contain spaces while a NUL cannot. Injectivity survives today only because `kind` comes from a closed three-value vocabulary and `retryNonce` is a number at the tail. A separator that cannot occur in a path (U+0001, or a newline) would make that guarantee structural rather than incidental.

### On the NUL byte

Confirmed with `od -c` on both sides. Base: `` const buildKey = `${cwd}\0${retryNonce}`; `` — the NUL is nothing but the field separator inside a template literal, which is why git classified the pre-image as binary. Head: `` const buildKey = `${coreId ?? ""} ${kind} ${cwd} ${retryNonce}`; ``, and the file now contains zero NUL bytes.

**Removing the byte does not change `buildKey` semantics.** The key's role — a cache/rebuild discriminator, compared only for equality and never parsed — is unchanged, and the separator is still a separator. What does change is deliberate and is the point of the fix: two more fields join the key, so the string value differs from the old one, which costs at most one surface rebuild on first load and nothing thereafter. The author's reviewer note documenting all this is exactly the right call.

### Scope

Clean. Seven files, all inside the user-terminal store / pane / panel / identity helpers and their tests. `routes/projects.$id.tsx`, `TerminalPane.tsx` and `ProjectBar.tsx` are untouched, so nothing collides with #397, #393 or #378 — and the three commits the base branch has gained since this branched do touch `ProjectBar.tsx`/`TerminalPane.tsx` with **no** overlap against this diff (GitHub reports MERGEABLE / CLEAN).

### Verification I ran

- `user-terminal-reload.test.tsx`, `user-terminal-identity.test.ts`, `user-terminal-pane-kind.test.tsx` — **3 files / 19 tests, all green**.
- Pre-existing `user-terminal-store.test.ts`, `user-terminal-panel-control.test.tsx`, `drop-legacy-user-terminals.test.ts`, `home-terminals.test.ts` — **4 files / 25 tests, all green**, no regressions.
- CI on `5712b94`: 18 checks pass, 0 fail (Typecheck, Lint and Unit Tests included). The checklist and CI comments on this PR are accurate; the only number I could not reproduce is "37 tests" — I count 44 across those seven files, which is bookkeeping, not a problem.

### Quality note

Comment quality, commit messages and the reviewer notes on this PR are unusually good: the binary-diff explanation, the "deliberately not checked" section and the local-CI caveat are all things a reviewer would otherwise have had to discover. The blocking item is narrow and mechanical — fix F1, decide on F2 and F3, and this is ready.

qcentic-dev:
## Reply to review [#5095682582](https://github.com/actana/control/pull/468#pullrequestreview-5095682582) - `5712b94` to `100fd2c`

Thanks - F1 is a real hole and the diagnosis was exact, including *why* it was new (`homeActive` flips once, `scopeKey` changes on every navigation). Two commits: `29a3d87` (fixes) and `100fd2c` (tests).

### Fixed

**F1 (blocking) - the restore could cancel itself off permanently.** `user-terminal-store.tsx`. I removed the cancel rather than repairing it. What the restore writes is decided by each row's identity, not by whichever scope was current when the request went out, so a navigation mid-flight cannot change the right answer; the three setters are idempotent under their own guards and a setState after unmount is a no-op, so the flag was protecting nothing and costing the answer. The once-per-run guard stays (it is what makes this one fetch), and the `catch` still un-latches it - so the documented retry contract now holds for a **slow** call as well as a rejected one, which was your point about the test exercising the wrong branch.

New case, exactly as asked: *"still restores when the operator navigates while the list is in flight"* - the list promise is held open, the store is sent to project B, the answer then lands in project A's bucket, with **one** list call rather than a discard-and-refetch. I verified it fails against the reviewed code by reintroducing `cancelled` and watching only that case go red.

**F2 - a prune could delete an identity that was just written.** Two halves, both taken:

- `pruneIdentities(map, liveIds, candidateIds)` now takes the set the answer is actually evidence about - the ids that existed when the request was issued, snapshotted before the fetch. An identity written during the window is absent from the answer because it did not exist yet, not because its row is gone, so it is out of scope for that answer.
- The cross-tab half you flagged as "not narrow" is fixed at the write, not worked around: `commitIdentityChange(before, after)` re-reads storage and applies **this tab's delta** - entries it added or changed win, entries it removed go, and entries another tab wrote in the meantime survive. I did not extend that to `hiddenIds`/`panelOpen`; you are right that they share the pattern, but losing a hidden flag is recoverable in one click and this is not the PR to change them in.

Covered by *"keeps both when a terminal is opened before the list answers"* (which asserts both the merge and the surviving identity) and two jsdom cases on `commitIdentityChange`. Also verified to fail against the old two-argument prune.

**F4 - a whole bucket was dropped on collision.** Restore now merges by terminal id, restored entries first (they predate anything opened this run). Same test as above.

**Nits.**

- `UserTerminalPane.tsx` - the separator is now U+0001, written as an `\u0001` escape in a normal string so the source stays plain text and the file keeps zero control bytes. Your framing is the right one: with a space the key was injective only by luck of the other fields' shapes; with a character that cannot occur in a path, a kind or a number it is structural.
- `UserTerminalPanel.tsx` - the "a restored session carries no Core" comment now says what the code does: the session's own Core wins, and the route's is the fallback for a session that has none.
- The test count: you are right, and my "37" was miscounted - I ran `session-warm-pool.test.ts` as the seventh file instead of `home-terminals.test.ts`. Your 44 on `5712b94` was correct. With the five cases added this round it is **7 files / 49 tests**, all green.

### Deferred, with the reasoning recorded

**F3 - the orphaned PTY.** Taken as "record it", not "fix it", and recorded in both places you asked for: the PR body now states the accumulation outright - *every reload leaves one live, unreachable login shell per restored terminal on the Core's machine*, with `findByTask` refusing shell PTYs and `killTerminal` only reaching the replacement - and it is filed as **#473**, referencing #394 and this PR, with both candidate fixes and what would close it.

Why not fixed here: re-attaching needs a replay that fails safe when the pty is gone, or a dead `ptyId` strands the pane behind a Retry that can only fail again. And killing the recorded PTY on restore is not obviously safe with more than one tab open - the second tab's reload would kill the shell the first tab is still attached to and using. Either is a change with its own failure mode and its own tests; neither is in #394's "What to do".

### Not changed

**The `homeActive` caveat.** Agreed and noted - `setHomeActive` has no caller in app code, so `HOME_SCOPE_KEY` is unreachable and the panel never renders on the dashboard. I left the `loadOnHome` cases in: they pin the store's contract at the seam where the kind used to be lost, and they will be the regression net if the dashboard scope is ever wired up. The PR body's framing of the Home symptom is fair to call theoretical on this branch; the primary #394 symptom ("pane gone from the project") is the one the fix is for.

### Verification on `100fd2c`

- The seven user-terminal suites: **7 files / 49 tests, all green** (`user-terminal-reload`, `user-terminal-identity`, `user-terminal-pane-kind`, `user-terminal-store`, `user-terminal-panel-control`, `drop-legacy-user-terminals`, `services/home-terminals`).
- Both new cases verified to **fail against the code they describe** - the bug was reintroduced for each and only the new case went red.
- `pnpm -r typecheck` clean; `pnpm lint` 0 errors (45 pre-existing warnings, unchanged).
- commitlint over `aa5df9d..100fd2c`, repo config: 0 problems across all four commits.
- Plain push, no force. Base unchanged (`fix/operator-pins-sessions-drop-cli`); still a draft.

CI status on the new head is posted below once it settles.

qcentic-dev:
### CI on the new head (`100fd2c`) — all green

Run [33690735465](https://github.com/actana/control/actions/runs/33690735465): **18 required checks pass, 0 fail, 0 pending.**

| Check | Result |
| --- | --- |
| Unit Tests | pass (2m4s) |
| Typecheck | pass (51s) |
| Lint | pass (44s) |
| Conventions (commitlint: PR title + all four commits) | pass (17s) |
| E2E — installer (debian, x64) / (ubuntu, x64) | pass |
| Smoke — Core release tarball (linux-x64) / (linux-arm64) | pass |
| Core image / Panel image (build + smoke, amd64) | pass |
| Dependency Audit, Secret Scan, Train rules, Promotion gate, Resolve PR image mode | pass |

Train-image jobs `skipping`, as expected for a ticket PR. Base unchanged (`fix/operator-pins-sessions-drop-cli`), still a draft, plain push — no force push at any point.

mrfqcentic (approved):
## Review round 2 — PR #468 against #394 (T3: project VM shell identity is lost on reload)

Re-reviewed at head `100fd2c`, as a diff against `fix/operator-pins-sessions-drop-cli` (merge-base `aa5df9d`). Every claim below is checked at the enforcing line, and the new cases were run **against the reviewed code of round 1** as well as against head.

**Verdict: approve.** All four round-1 findings are answered — F1, F2 and F4 by code, F3 by an accurate follow-up issue — and all three nits are taken. The acceptance criterion is met at the enforcing lines. Two non-blocking notes at the end (N1, one `if`; N2, one line in an issue) are for the author to take or to record; neither holds this up. It is still a draft; marking it ready is the author's call.

---

### Round-1 findings

**F1 (blocking) — the restore could cancel itself off permanently. Fixed.**

The cancel is gone rather than patched: `packages/panel/src/lib/user-terminal-store.tsx:253-310` has no `let cancelled`, no `if (cancelled) return`, and **no cleanup function at all** — the effect body ends at `:309` and the effect returns `undefined`. A grep for `cancelled` over the file at head matches nothing. So a `scopeKey` change mid-flight re-runs the effect, the once-per-run guard at `:255` short-circuits it, and the *original* run still applies its answer at `:269-303`. That answer is scope-independent by construction: `restoreUserTerminals` buckets by `identity.scopeKey` (`user-terminal-identity.ts:154-155`), never by the current scope. `restoreStartedRef` is still un-latched in the `catch` at `:307`, so the documented retry contract now holds for a slow call as well as a rejected one — which was the point.

**The new case drives a genuinely slow promise, not a rejected one** — the exact thing the round-1 suite got wrong. `user-terminal-reload.test.tsx:231-236` builds a `new Promise` and captures its `resolve` in `answer`; `:238-243` mount and navigate to `OTHER_PROJECT` while it is still pending; `:244-246` only then resolve it with the row. `:248-255` assert the row lands in `p1:main` — the bucket its identity names, not the current scope — and that `listHomeTerminals` was called exactly **once**, which a discard-and-refetch would fail. There is no `mockRejected*` in that case; the rejected-call case is a separate test at `:293-315`.

Verified rather than assumed: I checked `5712b94`'s `user-terminal-store.tsx` and `user-terminal-identity.ts` back out under the head test files and ran the suite —

```
× still restores when the operator navigates while the list is in flight
× keeps both when a terminal is opened before the list answers
Tests  2 failed | 8 passed (10)
```

— green at head. The test catches the bug it claims to catch.

**F2 — a prune could delete an identity written during the await window. Fixed, both halves.**

- Same-tab: `pruneIdentities(map, liveIds, candidateIds)` (`user-terminal-identity.ts:108-115`) prunes only where `candidateIds.has(id) && !liveIds.has(id)`. The candidate set is snapshotted **before** the fetch — `const prunable = new Set(Object.keys(readIdentityMap()));` at `user-terminal-store.tsx:261`, above the `void (async () => {` at `:263` — and is what `:303` passes. An identity written at `:404` during the window is not in `prunable`, so that answer cannot delete it. The ordering that makes this sound is worth stating: `setIdentities` at `:404` runs *after* `await api.createHomeTerminal` returns (`:390`), so every id in the snapshot already has a server row, and a list issued after the snapshot must see it.
- Cross-tab: fixed at the write. `commitIdentityChange(before, after)` (`user-terminal-identity.ts:128-138`) re-reads storage (`{ ...readIdentityMap(), ...after }`) and removes only the ids this tab dropped (`before` minus `after`). Wired at `user-terminal-store.tsx:185-189` against `committedIdentitiesRef`, seeded from the initial map at `:184` so mount writes nothing. The whole-map overwrite (`writeIdentityMap(identities)`) is gone.

Covered by `user-terminal-identity.test.ts:82-88` (prune) and `:95-109` (two jsdom commit cases), plus `user-terminal-reload.test.tsx:258-291` end-to-end, which asserts `["before", "opened"]` both survive in storage. All three identity cases go red against the round-1 module (verified). Leaving `hiddenIds`/`panelOpen` on the old pattern is the right call for this PR.

**F3 — the orphaned PTY. Recorded accurately, as asked.**

#473 is filed, labeled `needs-triage`, and cross-referenced from this PR. I checked its three load-bearing claims against the code rather than the prose:

- `findByTask` refuses shell / `shellSession` PTYs — `packages/core/src/pty-manager.ts:851` (`if (p.taskId === taskId && !p.shell && !p.shellSession) found = p.id;`), so a restored pane cannot adopt the old process. ✔
- `killTerminal` only reaches the `ptyId` the store currently holds — `user-terminal-store.tsx:431` reads it off the live session and `:471-473` kills that one; after a restore that is the replacement (`ptyId: null` at `:281`, refilled by `setPtyId`). ✔
- Nothing else reaps it — there is no idle/orphan sweep in `pty-manager.ts`, so "until the Core (or the machine) restarts" is literally true. ✔

The issue also records what would close it (persist-and-reattach with a replay that fails safe, or persist-and-kill guarded against a second attached tab) and why neither belongs in this PR, and the PR body now states the accumulation outright instead of leaving it implied by "the ptyId is still not persisted". That is what F3 asked for.

**F4 — a whole bucket was dropped on collision. Fixed.**

`user-terminal-store.tsx:271-290` merges by terminal id: `alreadyOpen` from the current bucket, `additions` filtered against it, `next[key] = [...additions, ...current]` with restored entries first (they predate this run). The `if (prev[key]) continue` is gone; `:286` now skips only when there is genuinely nothing to add. Focus at `:293-301` still fills only a bucket that has none, so a terminal the operator opened keeps focus.

### Nits — all three taken

- `UserTerminalPanel.tsx:433-436` — the false comment is replaced by one that matches `const sessionCoreId = s.coreId ?? coreId;`: the session's own Core wins, the route's is the fallback. ✔
- `UserTerminalPane.tsx:169` — the separator is now U+0001, written as an escape inside `[coreId ?? "", kind, cwd, retryNonce].join(...)`, with the reasoning at `:164-167`. The file contains zero NUL bytes **and** zero literal U+0001 bytes (checked with `od -c` and a control-character grep), so the source stays plain text while the separator is structurally impossible in a path, a kind or a number. ✔
- The test count is corrected to 7 files / 49 tests, and it is right (see below). ✔

### Acceptance — #394, criterion 1

> *Reload keeps the same user-terminal on the same project, or clearly has none — not a different shell.*

| Box | Verdict | Enforcing line |
| --- | --- | --- |
| Identity `{scopeKey, coreId, kind, cwd}` written at open, before the pane is shown | **met** | `user-terminal-store.tsx:398-404`, ahead of `updateSessions` at `:405` |
| Restored into the same project bucket | **met** | `user-terminal-identity.ts:154-155` groups strictly by `identity.scopeKey`; `user-terminal-store.tsx:271-289` writes exactly those keys |
| Same Core | **met** | `user-terminal-store.tsx:282` → `UserTerminalPanel.tsx:436`, `:470` |
| Same kind | **met** | `user-terminal-store.tsx:283` → `UserTerminalPanel.tsx:468-469` (`isHome`/`shellSession` derived from `s.kind`, not the ambient scope) |
| Same cwd | **met** | `user-terminal-store.tsx:284` → `UserTerminalPanel.tsx:465`; a VM shell's spawn ignores cwd anyway (`UserTerminalPane.tsx:423-439`) |
| A row it cannot identify restores as **nothing** — never Home, never another project | **met** | `user-terminal-identity.ts:154` (`if (!identity) continue`), malformed entries dropped at `:49-58` / `:66-73`. There is no fallback bucket in the restore path, and unavailable storage degrades the same way (`local-storage-json.ts:2-10`) |
| A Home terminal that was a VM shell stays a VM shell | **met** | kind recorded as `vm-shell` at `:401` regardless of scope; `UserTerminalPanel.tsx:468-469` |
| **"or clearly has none"** — no path where every scope ends up empty for unrelated reasons | **met** (was *partially met* in round 1) | the cancel is gone: `user-terminal-store.tsx:253-310`, no cleanup; pinned by `user-terminal-reload.test.tsx:217-256` |
| "What to do": persist project, Core, kind and cwd | **met** | `user-terminal-identity.ts:37-45` and the write at `:398-404` |

The `homeActive` caveat from round 1 stands and is unchanged: `setHomeActive` has no app-code caller, so the dashboard scope is unreachable and the `loadOnHome` cases pin store contract rather than an operator-reachable flow. The primary #394 symptom — the pane gone from the project — is the one this fixes, and it is fixed. Not a finding.

### Scope

Clean, and unchanged from round 1: the same seven files, all user-terminal store / pane / panel / identity plus their tests. `routes/projects.$id.tsx`, `components/views/TerminalPane.tsx` and `components/views/ProjectBar.tsx` are **untouched** (`git diff --name-only aa5df9d..100fd2c`; the one "…TerminalPane.tsx" in that list is `UserTerminalPane.tsx`, a different file), so nothing collides with #397, #393 or #378. GitHub reports MERGEABLE / CLEAN against the stacked base.

### Verification I ran

- The three new suites at head: **3 files / 24 tests green**. The four pre-existing user-terminal suites: **4 files / 25 tests green**. Total **7 files / 49 tests** — the corrected count in the PR body is right.
- Round-1 code under the head tests: the two new store cases and the three new identity cases fail, as claimed. The bug-catching claim is real, not asserted.
- CI on `100fd2c`: 18 checks pass, 0 fail (Unit Tests, Typecheck, Lint and Conventions included); train-image jobs skipping, as expected for a ticket PR. (My own typecheck ran in a throwaway clone without `routeTree.gen`, so its errors are my environment's, not this branch's — none touch the changed files.)

### Non-blocking, for you to take or to record

Neither of these blocks the merge, and neither is a regression against what ships today — both are places where the *identity* design has a residue worth a line of code or a line in an issue. I would take the first and record the second.

**N1 — `user-terminal-store.tsx:475-485`: the identity is dropped even when the delete failed.** `await api.deleteHomeTerminal(id)` at `:479` is wrapped in `catch { /* swallow */ }`, and `:485` then drops the identity unconditionally. If that call fails (Panel restarting, a blip), the row survives in `home_terminals` with no identity: it is deliberately never restored (`user-terminal-identity.ts:154`) and never pruned either, because it *is* in the list, so `pruneIdentities` sees it as live. That is precisely the stranding F2's own reasoning is built on — "lose an identity and its terminal can never be restored again, and its row is never cleaned up either" (`user-terminal-identity.ts:120-126`). Dropping the identity only when the delete actually reported success costs one `if`, and the failure mode of keeping it is benign: if the row really did go, the next restore's prune removes the identity anyway.

**N2 — the identity-less row has no way back.** A row with no identity is invisible by design (correct, and what #394 asks for) *and* unprunable, and this store is the only client of `listHomeTerminals`/`deleteHomeTerminal` in the Panel (`git grep` — nothing else lists or deletes them). So on the first load of this build every pre-existing row is orphaned server-side for good, as is every row belonging to a browser whose storage was cleared. The comment at `user-terminal-store.tsx:239-240` ("so the bucket cannot grow forever") is true of the identity map but not of the server table. It is not a regression — those rows are unreachable on the current build either way, since `setHomeActive` has no caller — and it is the accepted price of "restore nothing rather than guess". But it is the same shape of residue as F3, and it deserves the same treatment: a sentence in #473, or an issue of its own, rather than living only in a reviewer's notes. A sweep of identity-less rows on restore would close it, at the cost of deleting a row a *second* browser could still identify — which is exactly why it is a decision, not a patch.

For completeness, two things I checked and am **not** raising: the `catch` at `:307` un-latches the guard but the effect only re-runs on a `scopeKey` change, so an operator who stays put after a failed list sees an empty panel until they navigate — that is what the comment at `:305-306` says and what round 1 accepted; and `closeForProject` (`:490-506`) reads ids from a bucket the restore may not have filled yet, which is unreachable today because nothing in app code calls the user-terminal store's `closeForProject` (`projects.$id.tsx:1493` is the harness store's).

Also worth noting on the credit side: `cwd={s.cwd}` at `UserTerminalPanel.tsx:465` removes a rebuild the old `s.terminal.cwd || project?.path || ""` could cause on a scope change, so the widened `buildKey` does not cost what it might look like it costs.

Multi-tab footnote, self-healing: if tab A restores a terminal whose identity tab B wrote after A mounted, A's in-memory `identities` does not contain it, so killing it in A makes `forgetIdentities` a no-op (`user-terminal-store.tsx:485`) and the identity outlives the row until the next restore prunes it. Nothing an operator can see.

### Quality

This is the model of a round-2 reply: each finding answered at the level it was made; F1 fixed by *removing* the cancel rather than repairing the flag, with the reasoning written into the code where the next reader will meet it; both new cases verified red against the code they describe; and the one item not fixed turned into an issue a reader can act on without this PR in front of them. The comments earn their length.

PR #470, opened 2026-09-02
fix(panel): group filter must not drop the ring on the open project

### LINKED ISSUE #378: S6: group filter removes the pin ring while you are still on that project

qcentic-dev:
Closes #378 (S6). Base is `fix/operator-pins-sessions-drop-cli`, cut from `9b71076` — the merged #376 ring change is already underneath this.

## The bug

Open a project in group A, switch the GroupSwitcher to B. The page stays on A; the rail shows B with no selection ring. The scoped rail drew its overlay only when `activeId` was in the group-scoped visible list, so the filter took the ring away with the pin — the rail claiming nothing was open while the operator was plainly inside a project, and no tile left to get back to it.

## The fix

`ProjectBar.tsx` only. The project the route is on always keeps a tile: when the active group hides it, the scoped rail appends a trailing cluster of its own — named and coloured after the group that actually owns the project — and the ring lands there. The tile is dimmed to `0.72` so "you are here, and here is not in this group" reads at a glance without a second badge, and the divider above it separates it from the group's real workspace.

Nothing navigates. The URL still matches the open project; the filter decides which pins are *workable*, not which project you are in.

The rail also no longer collapses to nothing when the group switched to is empty — that was the same lie in a louder form (no rail, no ring, page still on the project).

The off-group tile is an affordance, not a member of this rail, so it deliberately claims none of a pin's powers:

- no chord badge — the digits belong to the group being shown, and `__root.tsx` resolves chords from `getRailClusters`, which this change does not touch, so badges and hotkeys stay in agreement (#379's invariant)
- no `data-pinned-item`, so it is not a drop slot in the frozen drag geometry
- no pointer drag, no `Shift+Arrow` reorder, and it is excluded from the pinned ids `persistProjectOrder` writes — counting it would splice a project from another group into the stored pinned order

## Tests

New: `packages/panel/src/components/views/__tests__/project-bar-group-filter-ring.test.tsx` — a real TanStack router plus a subscribable active-group store, because a group switch reaches this memoised component only through its own hook subscriptions. Four cases: ring and URL survive the switch; the empty-group case; the off-group tile takes no chord/drag/reorder; the tile folds back into the ordinary rail when its group is active again.

Verified red on the parent commit: three of the four fail without the `ProjectBar.tsx` change (the fourth is the same-group control, green either way).

Green locally: the new suite (4), the #376 ring suite `project-bar-active-ring.test.tsx` (3), `rail-projects` (7), `project-bar-status-dots`, `project-bar-activity`, `project-picker-switch`. `tsc --noEmit` clean; `eslint` clean on both files (two pre-existing unused-import warnings in `ProjectBar.tsx` are untouched).

## Scope

`ProjectBar.tsx` + one new test file. #382's mixed-rail pin reorder is a later round and is not taken here. No `use-fleet.ts`, `shared/projects.ts`, `projects.$id.tsx`, `terminal-store.tsx`, `TerminalPane.tsx`, `__root.tsx`, `active-group.ts` or `user-terminal-store.tsx`; no `packages/core`, `packages/shared` or `packages/cli`.

Draft on purpose.

https://claude.ai/code/session_01URh1dmYL7MDE9ZV1326cXj

### DIFF
--- a/packages/panel/src/components/views/ProjectBar.tsx
+++ b/packages/panel/src/components/views/ProjectBar.tsx
@@ -33,6 +33,7 @@ import {
   clusterPinnedByGroup,
   getGroupRailCluster,
   usesDirectRailProjectShortcuts,
+  type RailCluster,
 } from "~/lib/rail-projects";
 import { shouldFlashPinnedProjectLogo } from "./project-bar-activity";
 import { getPinnedProjectStatusDots } from "./project-bar-status-dots";
@@ -200,30 +201,60 @@ export const ProjectBar = memo(function ProjectBar({
   // scoped rail has to be able to put that project back on screen when the
   // active group would otherwise hide it (#378).
   const activeId = useMemo(() => pathname.match(/^\/projects\/([^/]+)/)?.[1], [pathname]);
-  const railClusters = useMemo(() => {
-    if (!groupScoped) return clusterPinnedByGroup(sortedPinned, orderedGroups);
+  // The clusters the rail draws, plus the name of the workspace it is drawing.
+  // The label travels WITH the clusters because it can no longer be read back
+  // off them: since #378 appends a cluster, `railClusters[0]` is the active
+  // group's cluster only when that group has projects, and guessing wrong made
+  // the rail's landmark announce the group that owns the OTHER project.
+  const rail = useMemo((): { clusters: RailCluster<ProjectWithCounts>[]; label: string } => {
+    if (!groupScoped) {
+      return {
+        clusters: clusterPinnedByGroup(sortedPinned, orderedGroups),
+        label: "Pinned projects",
+      };
+    }
     const cluster = getGroupRailCluster(projects ?? [], orderedGroups, activeGroup);
     const scoped = cluster.projects.length > 0 ? [cluster] : [];
+    // `getGroupRailCluster` names the active group even when it is empty, which
+    // is exactly the case the rail used to have no name for.
+    const label = cluster.label;
     // #378: opening a project in group A and switching the GroupSwitcher to B
     // left the page on A while the rail showed B with no ring — the rail
     // claiming nothing was open while the operator was plainly inside a
     // project. The filter still decides which pins are *workable*; the project
-    // you are on keeps a tile regardless, in a trailing cluster of its own
-    // named after the group that does own it, so the ring has somewhere to sit.
+    // you are on keeps a tile in a trailing cluster of its own, named after the
+    // group that does own it, so the ring has somewhere to sit.
     // Nothing here navigates: the URL stays on the open project.
+    //
+    // Scope, stated exactly: this holds for a group-scoped rail — group → group
+    // and group → Ungrouped. It does NOT hold for group → All, which returns
+    // above: All is the pinned rail, so an unpinned project opened from a
+    // group's workspace still loses its tile there. Same symptom, same gesture,
+    // deliberately not fixed here — see #474. Extending it is not a one-liner:
+    // in All `showClusterLabels` is true, so a synthetic cluster would render a
+    // real `data-cluster-header` / `data-group-handle`, entering the drag row
+    // snapshot and the group-handle list, and taking a group digit that
+    // `__root.tsx` (which resolves chords from the untouched `getRailClusters`)
+    // knows nothing about. That is a design change, not a branch.
     const open = activeId ? (projects ?? []).find((p) => p.id === activeId) : undefined;
-    if (!open || cluster.projects.some((p) => p.id === open.id)) return scoped;
+    if (!open || cluster.projects.some((p) => p.id === open.id)) {
+      return { clusters: scoped, label };
+    }
     const owner = orderedGroups.find((g) => g.id === open.groupId);
-    return [
-      ...scoped,
-      {
-        key: OFF_GROUP_CLUSTER_KEY,
-        label: owner?.name ?? "Ungrouped",
-        color: owner?.color ?? null,
-        projects: [open],
-      },
-    ];
+    return {
+      clusters: [
+        ...scoped,
+        {
+          key: OFF_GROUP_CLUSTER_KEY,
+          label: owner?.name ?? "Ungrouped",
+          color: owner?.color ?? null,
+          projects: [open],
+        },
+      ],
+      label,
+    };
   }, [activeGroup, activeId, groupScoped, orderedGroups, sortedPinned, projects]);
+  const railClusters = rail.clusters;
   const visible = useMemo(() => railClusters.flatMap((c) => c.projects), [railClusters]);
   const visibleById = useMemo(
     () => new Map(visible.map((project) => [project.id, project])),
@@ -809,7 +840,13 @@ export const ProjectBar = memo(function ProjectBar({
     });
   }
   const flatIndexById = new Map(visible.map((project, index) => [project.id, index]));
-  const railLabel = groupScoped ? (railClusters[0]?.label ?? "Group") : "Pinned projects";
+  // Named after the ACTIVE group, never after whatever cluster happens to be
+  // first: with an empty active group the off-group tile is index 0, and
+  // reading the label off it had the `role="navigation"` landmark announcing
+  // "Alpha group projects" while the GroupSwitcher read "Empty group" (#378
+  // review). The rail asserting something false about grouping is the very
+  // thing this PR is fixing; the accessibility surface counts too.
+  const railLabel = rail.label;
 
   const menuProject = menu ? visibleById.get(menu.id) ?? null : null;
 
--- a/packages/panel/src/components/views/ProjectBar.tsx
+++ b/packages/panel/src/components/views/ProjectBar.tsx
@@ -38,6 +38,10 @@ import { shouldFlashPinnedProjectLogo } from "./project-bar-activity";
 import { getPinnedProjectStatusDots } from "./project-bar-status-dots";
 
 const HOTKEY_LIMIT = PINNED_SLOT_COUNT;
+// Cluster key of the trailing tile the scoped rail keeps for the project the
+// route is actually on when the active group hides it (#378). It is a
+// selection affordance only: no chord, no reorder, no drop slot.
+const OFF_GROUP_CLUSTER_KEY = "mc-off-group";
 const DRAG_THRESHOLD_PX = 4;
 
 const ITEM_WIDTH = 58;
@@ -192,11 +196,34 @@ export const ProjectBar = memo(function ProjectBar({
     for (const group of groups) if (!seen.has(group.id)) out.push(group);
     return out;
   }, [groupDragOrder, groups]);
+  // Which project the route is on, read before the clusters are built: a
+  // scoped rail has to be able to put that project back on screen when the
+  // active group would otherwise hide it (#378).
+  const activeId = useMemo(() => pathname.match(/^\/projects\/([^/]+)/)?.[1], [pathname]);
   const railClusters = useMemo(() => {
     if (!groupScoped) return clusterPinnedByGroup(sortedPinned, orderedGroups);
     const cluster = getGroupRailCluster(projects ?? [], orderedGroups, activeGroup);
-    return cluster.projects.length > 0 ? [cluster] : [];
-  }, [activeGroup, groupScoped, orderedGroups, sortedPinned, projects]);
+    const scoped = cluster.projects.length > 0 ? [cluster] : [];
+    // #378: opening a project in group A and switching the GroupSwitcher to B
+    // left the page on A while the rail showed B with no ring — the rail
+    // claiming nothing was open while the operator was plainly inside a
+    // project. The filter still decides which pins are *workable*; the project
+    // you are on keeps a tile regardless, in a trailing cluster of its own
+    // named after the group that does own it, so the ring has somewhere to sit.
+    // Nothing here navigates: the URL stays on the open project.
+    const open = activeId ? (projects ?? []).find((p) => p.id === activeId) : undefined;
+    if (!open || cluster.projects.some((p) => p.id === open.id)) return scoped;
+    const owner = orderedGroups.find((g) => g.id === open.groupId);
+    return [
+      ...scoped,
+      {
+        key: OFF_GROUP_CLUSTER_KEY,
+        label: owner?.name ?? "Ungrouped",
+        color: owner?.color ?? null,
+        projects: [open],
+      },
+    ];
+  }, [activeGroup, activeId, groupScoped, orderedGroups, sortedPinned, projects]);
   const visible = useMemo(() => railClusters.flatMap((c) => c.projects), [railClusters]);
   const visibleById = useMemo(
     () => new Map(visible.map((project) => [project.id, project])),
@@ -233,7 +260,12 @@ export const ProjectBar = memo(function ProjectBar({
   // rail). In group-workspace mode that's the group's pinned subset — the
   // alphabetical unpinned tail is not reorderable — and persistProjectOrder
   // splices the subset back into the global pinned order.
-  pinnedIdsRef.current = visible
+  // The off-group tile (#378) is an affordance, not part of this rail's order:
+  // counting it would splice a project the active group does not contain into
+  // the persisted pinned order on the next keyboard reorder.
+  pinnedIdsRef.current = railClusters
+    .filter((cluster) => cluster.key !== OFF_GROUP_CLUSTER_KEY)
+    .flatMap((cluster) => cluster.projects)
     .filter((project) => project.pinned)
     .map((project) => project.id);
   const closeMenu = useCallback(() => setMenu(null), []);
@@ -719,9 +751,11 @@ export const ProjectBar = memo(function ProjectBar({
   // headers are stable project drop targets. The scoped rail still disappears
   // when the selected group has no projects because there is nothing to move
   // into it from that isolated view.
+  // when the selected group has no projects because there is nothing to move
+  // into it from that isolated view — unless the open project is parked there
+  // as the off-group affordance, which is the whole point of it (#378).
   if (railClusters.length === 0) return null;
 
-  const activeId = pathname.match(/^\/projects\/([^/]+)/)?.[1];
   const activeIndex = visible.findIndex((p) => p.id === activeId);
 
   // Group-number labels only earn their space when a real group exists. With
@@ -1064,11 +1098,14 @@ export const ProjectBar = memo(function ProjectBar({
           {cluster.projects.map((project, projectIndex) => {
         const idx = flatIndexById.get(project.id)!;
         const isActive = idx === activeIndex;
+        // The open-but-filtered-out project (#378). It shows and it rings; it
+        // does not take a chord, a reorder or a drop.
+        const isOffGroup = cluster.key === OFF_GROUP_CLUSTER_KEY;
         // Project numbers restart per cluster when group chords are visible;
         // the no-groups rail has one flat cluster, so these are direct slots.
         const projectNumber = projectIndex + 1;
         const groupNumber = clusterIndex + 1;
-        const hotkey = projectNumber <= HOTKEY_LIMIT ? projectNumber : null;
+        const hotkey = !isOffGroup && projectNumber <= HOTKEY_LIMIT ? projectNumber : null;
         const chordHint = directProjectShortcuts
           ? pinnedSlotBinding(projectNumber)
           : `${pinnedSlotBinding(groupNumber)} ${projectNumber}`;
@@ -1095,7 +1132,10 @@ export const ProjectBar = memo(function ProjectBar({
             : null;
         const tooltip = [
           hotkey ? `${project.name} (${chordHint})` : project.name,
-          project.pinned ? "Drag or press Shift+Arrow Up/Down to reorder pinned projects" : null,
+          isOffGroup ? "Open now — outside the active group" : null,
+          !isOffGroup && project.pinned
+            ? "Drag or press Shift+Arrow Up/Down to reorder pinned projects"
+            : null,
           needsInputLabel,
           runningLabel,
           finishedLabel,
@@ -1123,16 +1163,19 @@ export const ProjectBar = memo(function ProjectBar({
           <button
             key={project.id}
             type="button"
-            data-pinned-item={project.pinned ? true : undefined}
+            data-pinned-item={!isOffGroup && project.pinned ? true : undefined}
+            data-off-group={isOffGroup ? true : undefined}
             data-cluster-id={cluster.key}
             data-project-id={project.id}
             title={tooltip}
             aria-label={tooltip}
-            aria-keyshortcuts={project.pinned ? "Shift+ArrowUp Shift+ArrowDown" : undefined}
+            aria-keyshortcuts={
+              !isOffGroup && project.pinned ? "Shift+ArrowUp Shift+ArrowDown" : undefined
+            }
             onPointerDown={(e) => {
               // Only pinned tiles reorder — the group workspace's alphabetical
               // unpinned tail has a computed order with nothing to drag.
-              if (project.pinned) startProjectPointerDrag(project.id, e);
+              if (!isOffGroup && project.pinned) startProjectPointerDrag(project.id, e);
             }}
             onDragStart={(e) => e.preventDefault()}
             // Drop a file from the desktop onto a Project's tile and it goes to
@@ -1179,7 +1222,7 @@ export const ProjectBar = memo(function ProjectBar({
               showHoverLabel(project.name, e);
             }}
             onKeyDown={(e) => {
-              if (!project.pinned) return;
+              if (isOffGroup || !project.pinned) return;
               if (!e.shiftKey || (e.key !== "ArrowUp" && e.key !== "ArrowDown")) return;
               e.preventDefault();
               movePinnedProjectByKeyboard(project.id, e.key === "ArrowUp" ? -1 : 1);
@@ -1214,14 +1257,16 @@ export const ProjectBar = memo(function ProjectBar({
               zIndex: isDragging ? 6 : isActive ? 3 : clusterDragging ? 5 : 1,
               cursor: clusterDragging
                 ? "grabbing"
-                : !project.pinned
+                : isOffGroup || !project.pinned
                   ? "pointer"
                   : reorderSaving
                     ? "default"
                     : isDragging
                       ? "grabbing"
                       : "grab",
-              opacity: isDragging ? 0.92 : 1,
+              // Dimmed, but ringed: "you are here, and here is not in this
+              // group" reads at a glance without a second badge (#378).
+              opacity: isDragging ? 0.92 : isOffGroup ? 0.72 : 1,
               transform: `translateY(${tileOffset}px)`,
               boxShadow: isDragging
                 ? "0 0 0 2px color-mix(in srgb, var(--accent) 70%, white), 0 10px 24px -10px rgba(0, 0, 0, 0.6)"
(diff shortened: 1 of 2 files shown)


qcentic-dev:
## Acceptance criteria — issue #378 (S6)

- [x] **1. Switching group while viewing a project in another group does not pretend nothing is selected.**
  The scoped rail now appends a trailing off-group cluster holding the project the route is on whenever the active group would hide it, so the selection overlay always has a tile to sit on. The tile carries `data-off-group` and is dimmed to `0.72`; the divider above it is the owning group's colour, and its tooltip reads `Alpha — Open now — outside the active group`. This also covers the harsher variant where the group switched to is empty: the rail used to return `null` and vanish outright, which is the same claim made louder.
  Proven by `project-bar-group-filter-ring.test.tsx` → *"keeps the ring on the open project when the active group hides it, and stays on its URL"* (asserts `[data-active-project-id]` is still `p_alpha` and a tile for it exists after the switch, while B's own pin is present, so the switch really happened) and *"keeps the affordance even when the group switched to is empty"*.

- [x] **2. The URL still matches the open project.**
  Nothing in this change navigates — the fix is purely what the rail draws. Both tests above assert `router.state.location.pathname === "/projects/p_alpha"` after the group switch, and the fold-back test re-asserts it after switching B → A.

### Also held, not asked for but required not to regress

- [x] **The #376 ring behaviour still holds.** `project-bar-active-ring.test.tsx` (3 tests) green — same-Core pin click, plain route change with no click, and `?coreId=` on the navigation.
- [x] **Chords and badges still agree (#379's invariant).** The off-group tile takes no hotkey badge and `__root.tsx`'s `getRailClusters` is untouched, so the digit the badge shows and the digit the chord fires are still the same. Asserted: after the switch the off-group tile has no `.mc-project-hotkey-badge`, and B's own pin still shows `1`.
- [x] **The persisted pinned order is not polluted.** The off-group project is excluded from `pinnedIdsRef`, so a `Shift+Arrow` reorder cannot splice a project from another group into the stored order. It also carries no `data-pinned-item`, so it is not a drop slot in the frozen drag geometry, and its pointer-drag and reorder key handlers are no-ops.

### Explicitly not taken

- [ ] **Mixed-rail pin reorder (#382).** Same file, later round, deliberately out of scope here.
- [ ] **The unpinned-project case in "All" mode.** In All the rail is the pinned rail by definition, so an *unpinned* open project has no ring there either. That is a different, pre-existing behaviour with its own product question (does All grow tiles for unpinned projects?); #378 is scoped to the group filter and is fixed as filed. Flagging it rather than widening this PR.

### CI

Reported in a follow-up comment once the checks on the head commit settle.

qcentic-dev:
### CI — head `50e07aa`, all green

Run [33689071131](https://github.com/actana/control/actions/runs/33689071131): **20 pass, 0 fail, 4 skipping**.

`Conventions` 29s · `Typecheck` 49s · `Lint` 44s · `Unit Tests` 2m51s · `Secret Scan` 25s · `Dependency Audit` 29s · `Train rules` 5s · `Promotion gate` 5s · `Resolve PR image mode` 7s · `E2E — Panel service seam` 52s · `E2E — installer (ubuntu, x64)` 1m36s · `E2E — installer (debian, x64)` 1m39s · `Smoke — Core release tarball (linux-x64)` 46s · `Smoke — Core release tarball (linux-arm64)` 36s · `Core image / Resolve registries` + `Build + smoke (amd64)` · `Panel image / Resolve registries` + `Build + smoke (amd64)`.

Skipping (expected off a train branch): `Core image (train)`, `Panel image (train)`, `Resolve train tags`, `Train versions`, and both `Publish the multi-arch manifest` jobs.

Both commits pass `commitlint --from 9b71076 --to HEAD` locally — the whole range, not just the title.

The second commit, `50e07aa`, is a comment-only follow-up: the first one amended the note above the scoped rail's `return null` and left the original two lines in place, so the sentence appeared twice. No behaviour change; the ring suites were re-run green after it.

packages/panel/src/components/views/ProjectBar.tsx
```diff
@@ -718,10 +750,10 @@ export const ProjectBar = memo(function ProjectBar({
   // In All mode, real groups remain useful even without pinned projects: their
   // headers are stable project drop targets. The scoped rail still disappears
   // when the selected group has no projects because there is nothing to move
-  // into it from that isolated view.
+  // into it from that isolated view — unless the open project is parked there
+  // as the off-group affordance, which is the whole point of it (#378).
   if (railClusters.length === 0) return null;
```
mrfqcentic, line 786: **Blocking (introduced here).** Keeping this case alive is right, but line 812 downstream still reads the rail's accessible name off `railClusters[0]`:

```ts
const railLabel = groupScoped ? (railClusters[0]?.label ?? "Group") : "Pinned projects";
```

When the active group is empty, `scoped` is `[]`, so `railClusters[0]` is now the **off-group** cluster and the `role="navigation"` rail is labelled after the group that owns the *other* project. Probed at this head: open Alpha (in "Alpha group"), switch the GroupSwitcher to "Empty group" → `aria-label` is `"Alpha group projects"` while the switcher reads "Empty group". Before this change the early `return null` meant no wrong label was announced.

Suggested one-liner at 812:

```ts
const railLabel = groupScoped
  ? (railClusters.find((c) => c.key !== OFF_GROUP_CLUSTER_KEY)?.label ?? "Group")
  : "Pinned projects";
```

Side note for the same case: `clusterIndex > 0` is false, so no divider renders either — the owner group's name (carried by the divider's `aria-label`/`title` in the ordinary case) is unavailable here, leaving only the 0.72 opacity as the cue.

packages/panel/src/components/views/ProjectBar.tsx
```diff
@@ -192,11 +196,34 @@ export const ProjectBar = memo(function ProjectBar({
     for (const group of groups) if (!seen.has(group.id)) out.push(group);
     return out;
   }, [groupDragOrder, groups]);
+  // Which project the route is on, read before the clusters are built: a
+  // scoped rail has to be able to put that project back on screen when the
+  // active group would otherwise hide it (#378).
+  const activeId = useMemo(() => pathname.match(/^\/projects\/([^/]+)/)?.[1], [pathname]);
   const railClusters = useMemo(() => {
     if (!groupScoped) return clusterPinnedByGroup(sortedPinned, orderedGroups);
```
mrfqcentic: **Blocking as written.** This `!groupScoped` early return gates the whole affordance, but the comment just below asserts "the project you are on keeps a tile regardless". It does not — "All" is an option on the same GroupSwitcher, and the scoped rail is the group's *workspace*, so it lists unpinned projects and lets the operator open one.

Probed at this head with an unpinned project `p_unpinned` in group A, opened from A's workspace rail:

| destination | url | tile | ring |
| --- | --- | --- | --- |
| group B | `/projects/p_unpinned` | present | on it ✅ |
| **All** | `/projects/p_unpinned` | **ABSENT** | **null** ❌ |

Same gesture, same #378 symptom. Either extend the trailing cluster to the All branch, or narrow the comment to the scoped rail and file the All leg as a follow-up — but the comment should not claim an invariant the code does not hold.

mrfqcentic (changes requested):
## Review against #378 — head `50e07aa`, base `fix/operator-pins-sessions-drop-cli` @ `9b71076`

Reviewed by reading the enforcing lines in `ProjectBar.tsx`, not the PR description, and by running the suites at the PR head with a probe harness.

**The core of the fix is right.** The trailing off-group cluster is a genuinely well-chosen shape: because it is *trailing* and carries neither `data-pinned-item` nor `data-cluster-header`, it is invisible to `startProjectPointerDrag`'s frozen row/slot snapshot (`ProjectBar.tsx:381`) and to `startGroupPointerDrag`'s handle list, so it cannot become a drop slot and cannot displace any measured row. Verified independently, not taken from the body:

- **Ring survives the switch.** `activeId` (`:202`) comes off `useRouterState`-subscribed `pathname`, `railClusters` (`:203`) depends on it, and `itemOffsets` (`:797`) walks the same `railClusters` including the divider, so the overlay lands on the appended tile. #376's `project-bar-active-ring.test.tsx` still green.
- **Appears only when the group would hide it.** `:216` — `if (!open || cluster.projects.some((p) => p.id === open.id)) return scoped;`.
- **No duplicate tile.** Same `some(...)` guard; both the scoped list and `open` are derived from the one `projects` array, so there is no window where they disagree.
- **#379's indices are not shifted.** `__root.tsx:538` resolves chords from `getRailClusters`, which this PR does not touch and which never appends the cluster; `projectNumber` restarts per cluster and the appended cluster is last, so the scoped cluster's digits are unchanged, and `hotkey` is suppressed at `:1106`. Test asserts B keeps digit `1`.
- **Persisted order is not polluted.** `pinnedIdsRef` filters the cluster out at `:266-268`.
- **Scope is clean.** Only `ProjectBar.tsx` + one new test file. `use-fleet.ts`, `shared/projects.ts`, `terminal-store.tsx`, `projects.$id.tsx`, `__root.tsx`, `rail-projects.ts`, `active-group.ts` all untouched.
- **Delivery honesty holds.** I reproduced the red-on-parent claim exactly: the new suite run against `9b71076` fails 3 of 4 (the fourth is the same-group control). Full panel suite at head: 152 files / 1658 tests green. ESLint: the 2 warnings are pre-existing at `9b71076`.

Two things need a change before this lands.

### 1. Blocking — `ProjectBar.tsx:812`: the rail's landmark is now labelled after the wrong group

`const railLabel = groupScoped ? (railClusters[0]?.label ?? "Group") : "Pinned projects";`

This assumed `railClusters[0]` is the *active group's* cluster. With the empty-group case now kept alive, `scoped` is `[]` and the off-group cluster becomes index 0, so `aria-label` on the `role="navigation"` rail names the group that owns the *other* project. Probed at head: open Alpha (in "Alpha group"), switch the GroupSwitcher to "Empty group" → the rail announces **"Alpha group projects"** while the switcher reads "Empty group". Before this PR that path returned `null`, so no wrong label was announced — this is introduced here.

This is the same class of assertion the issue is about (the rail stating something false about grouping), just on the accessibility surface. One line:

```ts
const railLabel = groupScoped
  ? (railClusters.find((c) => c.key !== OFF_GROUP_CLUSTER_KEY)?.label ?? "Group")
  : "Pinned projects";
```

Worth noting the same empty-group case also renders no divider (`clusterIndex > 0` is false at `:1079`), so the only "not in this group" cue left there is the `0.72` opacity — the owner group's name, which the divider's `aria-label`/`title` carries in the ordinary case, is not available at all. Not blocking on its own, but it is the case where the label above matters most.

### 2. Blocking as written — `ProjectBar.tsx:204`: the comment overstates what is enforced

`if (!groupScoped) return clusterPinnedByGroup(sortedPinned, orderedGroups);` gates the whole affordance, but the comment three lines below says "the project you are on keeps a tile regardless". It does not: "All" is an option on the same GroupSwitcher, and the scoped rail is the group's *workspace*, so it lists unpinned projects too and lets you open one.

Probed at head — unpinned project `p_unpinned` in group A, opened from A's workspace rail:

| destination | url | tile | ring |
| --- | --- | --- | --- |
| group B | `/projects/p_unpinned` | present | on it ✅ |
| **All** | `/projects/p_unpinned` | **ABSENT** | **null** ❌ |

So the #378 symptom is still reachable through the same operator gesture, on the All leg. I am not asking you to widen scope unilaterally — either extend the trailing cluster to the All branch, or narrow the comment to say plainly that this holds for the scoped rail only and file the All leg as a follow-up. What should not ship is a comment asserting an invariant the code does not hold; this file's comments are load-bearing.

### Acceptance criteria

1. *Switching group while viewing a project in another group does not pretend nothing is selected* — **partially met.** Met and tested for group→group (including the empty-group destination) and for unpinned projects. Not met for group→All, reproduced above.
2. *The URL still matches the open project* — **met.** Nothing in the diff navigates; asserted in three of the four new tests, and confirmed on the All leg too (the URL is right there; only the ring is missing).

### Verdict

Request changes. The mechanism is the right one and the test suite is unusually good — a real router plus a subscribable store, red-on-parent verified. Fix `:812`, and make `:204`'s comment true (or extend the branch). Nothing else in the diff needs to move.

https://claude.ai/code/session_016jAPaAiYfgnCEaHbqWUfin

packages/panel/src/components/views/ProjectBar.tsx
qcentic-dev: Fixed in `5f1cd17`, and you were right that it was introduced here — before #378 that path hit the early `return null`, so the wrong name is new.

I went one step past the suggested one-liner, because of your own side note. `railClusters.find((c) => c.key !== OFF_GROUP_CLUSTER_KEY)?.label ?? "Group"` is correct — it never announces the other project's group again — but in exactly the empty-group case it finds nothing and falls back to the bare word `"Group"`. That is the case where, as you say, no divider renders and the owner group's name is nowhere on screen, so it is the case that can least afford a vague landmark.

So the label now travels *with* the clusters instead of being read back off them. The memo returns `{ clusters, label }`, and the label comes from `getGroupRailCluster`, which names the active group correctly even when the group is empty:

```ts
const cluster = getGroupRailCluster(projects ?? [], orderedGroups, activeGroup);
const scoped = cluster.projects.length > 0 ? [cluster] : [];
const label = cluster.label;
```

`const railLabel = rail.label;` at the render site. The index-0 assumption is gone rather than corrected, so appending another cluster later cannot resurrect it.

New test — *"names the rail after the active group, not after the off-group tile"* — walks Alpha group → Beta group → Empty group and pins the landmark to the switcher's group at each step. Verified red against the old one-liner with your exact strings: `expected 'Alpha group projects' to be 'Empty group projects'`.

The divider observation itself I have left alone: making a single-cluster rail draw a divider is a visual change to the ordinary scoped rail too, and you marked it non-blocking. The landmark now carries the active group's name in that case, and the tile keeps its `0.72` opacity and its "Open now — outside the active group" tooltip.

packages/panel/src/components/views/ProjectBar.tsx
qcentic-dev: Taken, in `5f1cd17`. **Chose the second option: narrow the comment, file the All leg as a follow-up — [#474](https://github.com/actana/control/issues/474).**

Why that one rather than extending the branch. Your reproduction is right and the gesture is the same, but the All leg is not a missing branch — in All, `usesDirectRailProjectShortcuts` is false whenever a real group exists, so `showClusterLabels` is **true** and every cluster renders a real header. A synthetic trailing cluster there would render `data-cluster-header` + `data-group-handle`, which is precisely what puts it into `startProjectPointerDrag`'s frozen row snapshot and `startGroupPointerDrag`'s handle list — it becomes a drop slot and a draggable group, the two properties you verified the scoped version carefully does *not* have. It would also take group digit `clusterIndex + 1` in its header keycap while `__root.tsx` resolves chords from `getRailClusters`, which does not know it exists: #379's badge/chord disagreement, reintroduced by the fix for #378.

Underneath that there is a product question I should not answer unilaterally: All is the *pinned* rail by definition, so growing a tile there for an unpinned project changes what All means. #474 carries the repro table, the three candidate shapes, and acceptance criteria that include "takes no group digit `__root.tsx` cannot also resolve".

The comment now states where the invariant stops instead of claiming it holds everywhere:

```
// Scope, stated exactly: this holds for a group-scoped rail — group → group
// and group → Ungrouped. It does NOT hold for group → All, which returns
// above: All is the pinned rail, so an unpinned project opened from a
// group's workspace still loses its tile there. Same symptom, same gesture,
// deliberately not fixed here — see #474. Extending it is not a one-liner:
// in All `showClusterLabels` is true, so a synthetic cluster would render a
// real `data-cluster-header` / `data-group-handle`, entering the drag row
// snapshot and the group-handle list, and taking a group digit that
// `__root.tsx` (which resolves chords from the untouched `getRailClusters`)
// knows nothing about. That is a design change, not a branch.
```

Agreed on the principle: a comment asserting an invariant the code does not hold is worse than no comment, and this file's comments are load-bearing.

packages/panel/src/components/views/ProjectBar.tsx
qcentic-dev: > Review [5095900838](https://github.com/actana/control/pull/470#pullrequestreview-5095900838) — CHANGES_REQUESTED

Both blocking findings addressed in `5f1cd17`. Thank you for probing this rather than reading the description — the landmark bug is real, it is mine, and it is the same class of false claim the issue is about.

### 1. `:812` landmark labelled after the wrong group — **fixed**

Fixed one step past your one-liner, because of your own side note. `railClusters.find((c) => c.key !== OFF_GROUP_CLUSTER_KEY)?.label ?? "Group"` never announces the other project's group again, but in the empty-group case it finds nothing and falls back to the bare word `"Group"` — and that is precisely the case where no divider renders and the owner group's name is nowhere else on screen.

So the label now travels *with* the clusters rather than being read back off them. The memo returns `{ clusters, label }`, the label comes from `getGroupRailCluster` (which names the active group correctly even when that group is empty), and the render site is `const railLabel = rail.label;`. The index-0 assumption is deleted rather than patched, so appending another cluster later cannot resurrect it.

New test *"names the rail after the active group, not after the off-group tile"* walks Alpha group → Beta group → Empty group and pins the accessible name to the switcher's group at each step. Red against the old one-liner with your exact strings: `expected 'Alpha group projects' to be 'Empty group projects'`.

I left the divider observation alone as you marked it — making a single-cluster rail draw a divider changes the ordinary scoped rail too. The landmark now carries the group name in that case, and the tile keeps its `0.72` opacity and its "Open now — outside the active group" tooltip.

### 2. `:204` comment overstating the invariant — **fixed by narrowing, All leg filed as [#474](https://github.com/actana/control/issues/474)**

**Chose narrow-the-comment + follow-up, not extend-the-branch.** Your repro is right and the gesture is identical, but the All leg is not a missing branch. In All, `usesDirectRailProjectShortcuts` is false whenever a real group exists, so `showClusterLabels` is **true** and every cluster renders a real header — a synthetic trailing cluster there would carry `data-cluster-header` + `data-group-handle` and so enter `startProjectPointerDrag`'s frozen row snapshot and `startGroupPointerDrag`'s handle list, becoming a drop slot and a draggable group: exactly the two properties you verified the scoped version does not have. It would also take group digit `clusterIndex + 1` while `__root.tsx` resolves chords from `getRailClusters`, which does not know it exists — #379's badge/chord disagreement, reintroduced by the fix for #378.

And underneath that is a product question I should not settle on my own: All is the *pinned* rail by definition, so growing a tile there for an unpinned project changes what All means.

[#474](https://github.com/actana/control/issues/474) carries your repro table, the mechanism above as the reason it was not folded in, three candidate shapes, and acceptance criteria including "takes no group digit `__root.tsx` cannot also resolve". The comment in the memo now states plainly where the invariant stops and points at it.

### 3. Also taken

Your "no duplicate tile" proof point is now asserted rather than only reasoned: after a group switch, exactly one `[data-project-id]` for the open project and exactly one `[data-off-group]` in the rail — so the `some(...)` guard cannot regress into two tiles claiming one ring.

### Not changed (your non-blocking observations, agreed)

`railClusters` depending on `activeId`; the context menu and file-drop target still operating on the off-group tile (both are correct there — it is a real project, and edit/pin/drop all address it by its own `coreId`); `activeId` matching by project id without `coreId`, which is pre-existing on the `#376` line above.

### State

Head `5f1cd17`, CI **all green** — 20 pass, 0 fail, 4 skipping ([run 33692792099](https://github.com/actana/control/actions/runs/33692792099)), including `Unit Tests` 2m46s, `Typecheck` 47s, `Lint` 42s, `Conventions` 22s. All three commits pass `commitlint --from 9b71076 --to HEAD` locally. Plain push, no force push at any point. Still a draft, base unchanged.

Acceptance criterion 1 is now met for every leg this PR claims — group → group, group → Ungrouped, and the empty-group destination — with the All leg explicitly out of scope and tracked in #474 rather than silently left. Criterion 2 unchanged: nothing in the diff navigates.

mrfqcentic (approved):
## Review round 2 against #378 — head `5f1cd17`, base `fix/operator-pins-sessions-drop-cli` (merge-base `9b71076`)

Both round-1 blocking findings are resolved. I re-drove my own round-1 probes against this head from a fresh checkout rather than reading the reply, and re-ran them against `50e07aa` and `9b71076` to confirm they actually discriminate.

### 1. `ProjectBar.tsx:849` — landmark named after the wrong group — **fixed, and fixed better than I asked**

You did not take my one-liner, and you were right not to. `railClusters.find((c) => c.key !== OFF_GROUP_CLUSTER_KEY)?.label ?? "Group"` would have degraded to the bare word `"Group"` in exactly the empty-group case — the case where no divider renders and the owner group's name is nowhere else on screen. Carrying the label out of the memo alongside the clusters (`:209`, `:220`, `:254`) deletes the index-0 assumption instead of correcting it, so appending a cluster later cannot resurrect it. `getGroupRailCluster` (`rail-projects.ts:81`) does name the active group when that group is empty, so the label is right in the one case that had no name at all before.

Proven, not taken from the reply:

| head | `aria-label` after: open Alpha (Alpha group) → switch to Empty group |
| --- | --- |
| `50e07aa` (round-1 head) | `Alpha group projects` ❌ |
| `9b71076` (parent) | rail returned `null` |
| `5f1cd17` | `Empty group projects` ✅ |

My probe fails at `50e07aa` with your test's exact string — `expected 'Alpha group projects' to be 'Empty group projects'` — and passes here. I also added a leg you did not test: **group → Ungrouped with no ungrouped projects** announces `Ungrouped projects`, red at `50e07aa` (`expected 'Alpha group projects' to be 'Ungrouped projects'`), green here. The fix generalises rather than special-casing my repro.

### 2. `ProjectBar.tsx:229-238` — the comment overstating the invariant — **fixed by narrowing; #474 is accurate**

You took the second of the two routes I offered, which is the right call, and the comment now says exactly where the invariant stops. I checked every load-bearing claim in it against the code rather than accepting the reasoning:

- `usesDirectRailProjectShortcuts` (`rail-projects.ts:33`) is unconditionally `true` when `activeGroup !== ALL`, so `showClusterLabels` (`:794`) is `false` for every rail that can hold the off-group cluster, and `true` in All — the comment's premise holds.
- Under `showClusterLabels` a cluster renders a real `data-cluster-header` (`:989`, `:1026`) and `data-group-handle` (`:1024`), which are precisely the selectors `startProjectPointerDrag` freezes rows from (`:411`) and `startGroupPointerDrag` builds its handle list from (`:642`). A synthetic cluster in All really would become a drop slot and a draggable group.
- `groupNumber = clusterIndex + 1` (`:1142`) while `__root.tsx:538` resolves chords from `getRailClusters`, which never appends. The badge/chord disagreement you describe is real.

So "design change, not a branch" is a fair reading, not a convenient one. #474 carries my repro table, that mechanism, three candidate shapes and acceptance criteria including the chord-agreement constraint — it accurately describes what was deferred. I re-ran the All-leg repro at this head and it still behaves exactly as the comment and #474 say it does:

| destination | url | tile | ring |
| --- | --- | --- | --- |
| group B | `/projects/p_unpinned` | present | on it ✅ |
| All | `/projects/p_unpinned` | absent | `null` — documented at `:229-238`, tracked in #474 |

Code and comment now agree. That was the blocking condition, and it is met.

### 3. The no-duplicate assertion — **present**

`project-bar-group-filter-ring.test.tsx:259-260` asserts exactly one `[data-project-id]` for the open project and exactly one `[data-off-group]` after a switch, with `:277` and `:338` repeating it on the empty-group and fold-back legs. The `some(...)` guard can no longer regress into two tiles claiming one ring.

### Acceptance criteria

1. **Switching group while viewing a project in another group does not pretend nothing is selected — met.** Verified on every leg the issue describes: group → group, group → empty group, group → Ungrouped, and for unpinned projects within those legs. The one adjacent leg still open — group → All with an *unpinned* project — is a different precondition from the one #378 files, is now stated in the code instead of contradicted by it, and is tracked in #474. Round 1 offered narrowing-plus-follow-up as a sufficient resolution and this is it; I am not counting a documented, tracked deferral against the criterion, but I am recording plainly that the leg exists.
2. **The URL still matches the open project — met.** Nothing in the diff navigates. Asserted in the new suite and reproduced in my own harness on every leg, including All, where the URL is correct and only the tile is missing.

### Round-1 proof points, re-driven — no regressions

- **Pin ordering not polluted.** `pinnedIdsRef` filters the cluster out (`:297-301`). Probed end-to-end, not just read: open Alpha, switch to Beta (two pins), `Shift+ArrowDown` on Beta's first pin → `reorderPinnedProjects` receives `["p_alpha","p_beta2","p_beta1"]`, i.e. the global order with only B's subset permuted and `p_alpha` untouched. `Shift+Arrow` on the off-group tile itself fires nothing at all.
- **#379 badge/chord agreement.** Badges in the scoped rail read `1`/`2` with the off-group tile carrying none, and `resolveRailChordTarget(getRailClusters(...), null, 1|2)` independently resolves to the same two projects. In the empty-group case `getRailClusters` returns `[]`, so there is no digit to fire and no badge drawn — agreement holds at both ends.
- **#376 ring in the same frame as the URL.** A plain `router.navigate` with no click, taken while an off-group tile is on screen, moves the ring and the URL together in one act, and the stale off-group tile is gone rather than left ringed. `project-bar-active-ring.test.tsx` still green (3).
- **Delivery honesty.** The new suite is 4-of-5 red at the true parent `9b71076` (the same-group fold-back control is green either way); all 13 of my own probes are red there too. Head: targeted suites 17/17, full panel suite green on CI (`Unit Tests` 2m46s), `tsc --noEmit` clean, `eslint` clean apart from the two pre-existing unused-import warnings the diff does not touch. `50e07aa` is still an ancestor of `5f1cd17`, so the history I reviewed in round 1 was not rewritten.

### Scope

Two files: `ProjectBar.tsx` and the one new test. `use-fleet.ts`, `shared/projects.ts`, `terminal-store.tsx`, `projects.$id.tsx`, `__root.tsx`, `rail-projects.ts`, `active-group.ts`, `user-terminal-store.tsx` and `TerminalPane.tsx` are all untouched — no parallel ticket's file is in this diff. The round-2 delta over `50e07aa` is the memo return shape, the comment, `railLabel` and tests; no behaviour was smuggled in with the fix.

### Non-blocking — nothing here needs to move before this lands

- **`ProjectBar.tsx:860`.** `aria-label={groupScoped ? \`${railLabel} projects\` : "Pinned projects"}` hard-codes the literal the memo already returns as `label: "Pinned projects"` (`:213`), so the memo's All-branch label is never read. Two sources of truth for one string, in the commit whose whole point was to make the label single-sourced. `: railLabel` closes it.
- **`ProjectBar.tsx:243`, first paint only.** `ProjectBar` defaults `groups` to `[]` (`:111`) while `displayedActiveGroup` keeps the stored group id scoped until the groups query lands (`active-group.ts:67`). In that window the owner lookup misses and the appended cluster's divider announces `Ungrouped` in the neutral border colour instead of the owning group's name and colour — I confirmed it renders `aria-label="Ungrouped"`. It self-heals on the next render, and the landmark's `"Group projects"` fallback in the same window is pre-existing, unchanged by this PR. Deferring the append until `groups` has loaded would close it if you think it is worth a line.
- **Out of scope, pre-existing.** The context menu's pin item computes `pinned: target?.pinned !== true` (`:1478`) while its label reads `menuProject?.pinned === false ? "Pin project" : "Unpin project"` (`:1492`). If the project leaves `visible` between opening the menu and clicking, `menuProject` is `null`, the item still says "Unpin project", and clicking it sends `pinned: true`. Untouched by this diff — flagging it for your judgement, not asking for it here.

### Verdict

**Approve.** Both blocking findings are genuinely fixed — the landmark one more thoroughly than requested — the comment and the code now agree, #474 is an honest and accurate record of what was deferred, and every round-1 proof point still holds under fresh probes. Still a draft; marking it ready is yours to do.

https://claude.ai/code/session_01SrmfA4W7hjHzKmAx2HHb5q

PR #471, opened 2026-09-02
fix(panel): ignore late materialize from a pin you already left

### LINKED ISSUE #381: B4: an uncached pin materializes after you have already clicked away

- Hunt ID: `B4`
- Severity: broken
- First-fix owner: panel
- Source: panel-rendering-investigation-2026-09-02.md (B4)

## What the operator sees

During A→B→A with B uncached, B's sessions and focus land asynchronously after you are already back on A.

## Where

`useProject` / `useTasks` `staleTime: 30_000`, `gcTime: 5 min`. Evicted project → `project === undefined` → loading shell, then late rehydrate.

## What to do

Scope in-flight materialization to the project+core that is currently visible. Cancel or ignore a late B when the URL is already A. Do not drop the 30s staleTime without a reason.

## Acceptance

1. Clicking A then B then A never leaves B's sessions painted on A's URL.
2. A cold pin still loads; it just cannot win after you have left it.

qcentic-dev:
Closes #381 (B4).

Base is `fix/operator-pins-sessions-drop-cli`, not the train — this is the next
link in the pin/session/drop serial chain. The base has been merged in
(`53ee5b8`, a plain merge commit, no conflicts).

## What the operator saw

An uncached pin answers slower than the operator clicks. During A → B → A, B's
`useProject` and `useTasks` reads are still in flight when the URL is already back
on A — an evicted project renders the loading shell and rehydrates late — so what
those reads were going to materialize (B's sessions, B's archived count, the focus
that follows them) arrives for a project nobody is on any more.

## The fix

In-flight materialization is scoped to the project **and Core** that is actually
visible, in two halves.

**Cancel what is still in flight.**
`packages/panel/src/lib/visible-project-scope.ts` (new) counts the live readers of
each `(projectId, coreId)` bucket. `useProject` / `useTasks` / `useTask` retain
that scope while mounted; when the last reader goes, an **in-flight** fetch for it
is cancelled — react-query reverts the query to the state it had before the fetch
and the answer is dropped whenever it finally turns up.

The retain is per scope, not per query: a project is read by two queries at once,
and the first of them to unmount must not conclude the operator has left while the
other is still on screen. Every reader's release runs together, at zero.

**Stamp what cannot be cancelled.**
A cancelled fetch's promise still resolves — the panel link has nothing to abort —
so its fetcher runs to the end, and anything the fetcher writes for *itself* is not
covered by the cancellation. So a scope also carries a **visit generation**, bumped
every time its viewer count falls to zero; a read stamps the generation it started
in and compares on landing, and the archived count that rides the `tasksList`
answer (ADR 0019) is parked only by a read whose visit is still open.

A visibility snapshot could not do this, which was review finding 1: during
A → B → A → B the read abandoned on the way out lands *after* the read that
replaced it, while B is on screen again and looking perfectly current, and
overwrote the live count with a stale one. Nothing healed it —
`useCoreArchivedTaskCount` is `enabled: false` with no fetcher of its own — so the
wrong number gated the Archived tab, its label, the "View archived" tooltip, the
auto-exit effect and the delete-confirm dialog. There is now a test that drives
exactly that interleaving; it writes 99 over 7 with the snapshot and holds 7 with
the generation.

### What is deliberately unchanged

- **The 30s `staleTime` and the 5 min `gcTime` stand.** Nothing here makes a read
  start any later, and no extra refetching buys its way out of the race. A cold
  pin still loads — it just cannot win after you have left it.
- **Only a fetch in flight is ever cancelled.** A settled query keeps its rows, so
  leaving a project never costs the cache you come back to.
- **A read nobody is watching spans no visit**, so an imperative `fetchQuery` or a
  prefetch keeps its side effects — which is what keeps the existing archived
  read-path suite honest.

## Also from the review

- **Finding 2** — reader callbacks are keyed by the query key they cancel, so a
  pane mounting and unmounting through one visit replaces its entry instead of
  piling up another closure that fires redundantly on the way out. They are still
  held until zero rather than dropped as each reader releases: the two queries of
  one project unmount in separate cleanups inside a single commit, and dropping
  the first reader's entry as it goes would leave *its* fetch uncancelled when the
  second takes the scope to zero a moment later — the AC1 regression. Keying fixes
  the harm the finding named (unbounded accumulation, redundant repeat cancels)
  without that.
- **Finding 3** — `useTask` now takes `coreId` and `TerminalPane` passes
  `descriptor.coreId`. It hard-coded `null`, so for a Core-owned project the pane
  retained an untagged scope and registered its cancel against a key no Core pane
  reads. It was also *subscribed* to a bucket nothing fills, which is why the pane
  header never went live for a Core Session; it now reads the same tagged bucket as
  the `tasksKey` two lines below it.
- **Test note** — a comment now says which lines in the A → B → A test actually
  discriminate (the cache assertions), so a future reader does not trust the DOM
  ones, which pass on revert.

## Boundaries

`projects.$id.tsx`, `ProjectBar.tsx`, `use-fleet.ts`, `shared/projects.ts`,
`__root.tsx`, `active-group.ts`, the terminal stores and every `packages/core`,
`packages/shared` and `packages/cli` file are untouched. `TerminalPane.tsx` is the
one file outside the query layer, for the one call-site change finding 3 asked for.

One incidental cleanup: an already-unused `useMemo` import in `queries/index.ts`,
which was the one lint warning that file carried.

## Tests

- `packages/panel/src/queries/__tests__/late-pin-materialize.test.tsx`
  - A → B → A with B uncached: B's late project + task answers never paint on A's
    URL, and nothing of B's is left parked in the cache. Fails without the guard.
  - **A → B → A → B with B's two reads answering out of order** (new): the
    abandoned read's `archivedCount: 99` does not overwrite the live `7`. Fails
    with the visibility snapshot, at exactly the value the review predicted.
  - A cold pin the operator stays on still loads, archived count and all —
    including through a StrictMode double-mount (new).
  - Leaving a project keeps the rows that had already landed.
- `packages/panel/src/lib/__tests__/visible-project-scope.test.ts` — refcounting,
  idempotent release, one project id on two Cores kept apart, every reader told at
  once, **one callback per key however often a reader remounts** (new), **a left
  read staying stale after the operator returns** (new), and a read nobody is
  watching left alone.
- `pnpm vitest run src/queries src/lib` in `packages/panel`: **73 files, 654 tests,
  all green**. `src/components`: **18 files, 177 tests, all green**.
- `tsc --noEmit` for `@actana/panel`: clean. `eslint` on the changed files: clean.
- commitlint over the whole range, with the repo's own config: 0 problems.

https://claude.ai/code/session_01QpHg3JfZLgeP1CQCgjaW2V


qcentic-dev:
## Acceptance criteria — issue #381 (B4)

- [x] **1. Clicking A then B then A never leaves B's sessions painted on A's URL.**
  `useProject` / `useTasks` / `useTask` retain the `(projectId, coreId)` scope they
  read while they are mounted (`lib/visible-project-scope.ts`). When the last
  reader of a scope goes, an in-flight fetch for it is cancelled — react-query
  reverts the query to its pre-fetch state and the answer is dropped whenever it
  turns up. The archived count that rides the `tasksList` answer is not parked
  either, so it cannot outlive the list it came in on.
  Proven by `late-pin-materialize.test.tsx` → *"never leaves B's sessions painted
  on A's URL after A then B then A"*: with A warm and B uncached, B's project and
  task answers are released only after the URL is back on A; the board still shows
  A, and B's buckets are empty — so there is nothing parked to paint the moment
  another surface subscribes to B. The test fails on this branch with the guard
  disabled (verified locally).

- [x] **2. A cold pin still loads; it just cannot win after you have left it.**
  The 30s `staleTime` and the 5 min `gcTime` are untouched, no extra refetching was
  added, and nothing makes a read start any later — the guard is about who is on
  screen, not about freshness. Only a fetch *in flight* is ever cancelled, so a
  settled query keeps its rows and coming back to a pin is still instant. An
  imperative `fetchQuery` or a prefetch has no reader behind it and is never
  treated as late.
  Proven by `late-pin-materialize.test.tsx` → *"still loads a cold pin the operator
  stays on"* (loading shell, then the project, its sessions and its archived count)
  and *"keeps a settled project's rows when the operator leaves it"*.

### Nothing unchecked

Both acceptance criteria are met and covered by tests. The issue's "Where" and
"What to do" were followed as written: in-flight materialization is scoped to the
project **and Core** currently visible, and a late B is cancelled rather than the
staleTime being dropped.

### Boundaries

`projects.$id.tsx` was **not** touched, and did not need to be — the URL's project
and Core reach the guard through the hooks the route already calls. `ProjectBar.tsx`,
the terminal stores and panes, `use-fleet.ts`, `shared/projects.ts`, `__root.tsx`,
`active-group.ts` and every `packages/core` / `packages/shared` / `packages/cli`
file are untouched. The whole change is `packages/panel/src/queries/index.ts` plus
one new module and two new test files under `packages/panel/src`.

### Local verification

| check | result |
| --- | --- |
| `vitest run src/queries src/lib` (panel) | 67 files, 596 tests, all green |
| `tsc --noEmit` (`@actana/panel`) | clean |
| `eslint` on the changed files | clean |
| commitlint over `aa5df9d..HEAD` (config + `--from/--to`, as CI runs it) | 0 problems, 0 warnings |

The four failing suites in a full `packages/panel` run —
`server/__tests__/operator-auth`, `server/__tests__/harness-hooks-api`,
`server/panel-link/__tests__/live-read-path`,
`server/panel-link/__tests__/terminals-in-browser` — fail identically on the
pristine base commit `aa5df9d` in this environment (checked with the branch
stashed), so they are pre-existing and unrelated to this change.

### CI status of the PR head (`e7384b1`)

**All green — 18 pass, 6 skipping, 0 failing** ([run 33689147570](https://github.com/actana/control/actions/runs/33689147570)).

Pass: Conventions · Train rules · Promotion gate · Resolve PR image mode · Lint ·
Typecheck · **Unit Tests** · Dependency Audit · Secret Scan · E2E — Panel service
seam · E2E — installer (ubuntu, x64) · E2E — installer (debian, x64) · Smoke —
Core release tarball (linux-x64) · Smoke — Core release tarball (linux-arm64) ·
Core image (Resolve registries, Build + smoke amd64) · Panel image (Resolve
registries, Build + smoke amd64).

Skipping (train-only jobs, as expected on a non-train PR): Core image (train) ·
Panel image (train) · Resolve train tags · Train versions · both multi-arch
manifest publishes.

mrfqcentic (changes requested):
## Review of #471 against issue #381 (B4) — changes requested

Reviewed at head `e7384b1`, base `fix/operator-pins-sessions-drop-cli` (correct base for this wave; every PR 443→472 targets it). Verified by reading the enforcing lines and by running the suites, not from the description.

### What is genuinely right

**The landmine is respected.** This is a real cancellation guard, not a `staleTime` change. `packages/panel/src/queries/index.ts:375` is `queryClient.cancelQueries({ queryKey, exact: true })` behind a `retainProjectScope` refcount; `staleTime: 30_000` / `gcTime: 5min` are untouched in the diff and are re-asserted in the test client at `late-pin-materialize.test.tsx:135`. Checked against the installed `@tanstack/query-core@5.101.2`: `Query.#revertState` is captured at fetch start and `onCancel` restores it, and because `bridge.listTasks` never consumes the abort signal, react-query's own `removeObserver` cancel does **not** fire — which is precisely why an explicit cancel is needed here. Correct diagnosis, correct mechanism.

**AC1 — met.** `late-pin-materialize.test.tsx:188-192` is the enforcing assertion: after A→B→A, B's project bucket, B's task bucket and B's archived-count bucket are all `undefined`. Nothing is written, so nothing anywhere can paint it. Reverting `queries/index.ts` to the base makes that test fail at line 188 (verified locally).

**AC2 — met.** `late-pin-materialize.test.tsx:195-214` loads a cold pin the operator stays on, archived count included. I additionally probed A→B→A→**B** (return re-reads and paints), StrictMode double-mount, and a request that never settles: after cancel the query is back to `fetchStatus: "idle"` and the return visit issues a fresh read. **No wedge** — and the guard fails open (a read with no reader behind it is never treated as late), which is the safe direction.

**Boundaries — clean.** Four files, all owned by this ticket. `ProjectBar.tsx` (378/382), `use-fleet.ts` + `shared/projects.ts` (377) and `projects.$id.tsx` (397/401) are untouched. The dropped `useMemo` import was genuinely unused on the base.

**Merge — clean.** `git merge-tree --write-tree 7e75832 e7384b1` succeeds with no conflict; the base moved by 6 commits (463, 455, 470 and others) but none of them touch `queries/index.ts`. On the merged tree, `vitest run src/queries src/lib` is 71 files / 630 tests green.

### Blocking

**1. `packages/panel/src/queries/index.ts:178` — the archived count still outlives its own list, which is the exact thing the comment on line 174-177 says it prevents.**

`watchProjectScope` returns `wasVisible && !isVisible` — "is this scope on screen *now*", not "is this fetch still the current one". So it cannot tell *left* from *left and came back*. Reproduced:

- pin B (uncached, slow `listTasks`) → click A → B's read #1 is cancelled and reverted → click B again → read #2 starts
- read #2 lands first with the fresh `archivedCount: 7` → bucket is `7`
- read #1 finally answers with `archivedCount: 99`; `operatorLeft()` is `false` because B is on screen again, so line 179 writes `99`
- final state: task list holds read #2's rows, archived count holds read #1's number

Nothing heals it. `useCoreArchivedTaskCount` (`queries/index.ts:425-431`) is `enabled: false` with `queryFn: () => 0` and no fetcher of its own, so the wrong number stands until some unrelated event triggers another `tasksList` answer for B. Per the docstring at `queries/index.ts:413-424` that number gates the Archived tab, its label, the "View archived" tooltip, the auto-exit effect and the delete-confirm dialog.

This is a race the PR introduces into its own contract: before the guard, list and count were both last-writer-wins and therefore at least consistent with each other; after it, the list is protected and the count is not. A per-fetch generation token — bump a counter on every retain/release of a scope, capture it in the queryFn, compare on landing — closes it and matches what the ticket asked for. The visibility snapshot cannot.

Please also correct the PR body / commit message, which state the count "cannot outlive the list it came in on".

### Non-blocking

**2. `packages/panel/src/lib/visible-project-scope.ts:61` — `onLeft` is added on every retain but never removed when *that* reader releases.** Entries are only discarded when the scope reaches zero (line 71). On a project the operator stays on, the board holds the scope while panes and the hidden-session bar mount and unmount; each mount adds another closure over `queryClient` and a `queryKey`, and on leaving they all fire, each running a redundant `isFetching` + `cancelQueries` against the same one or two keys. Bounded by one visit and harmless today, but dropping the reader's own callback in the release path — before the zero check — costs one line and keeps the "all together at zero" semantics intact.

**3. `packages/panel/src/queries/index.ts:443` — `useTask` hard-codes `coreId: null`, so the comment above it does not hold for a Core pin.** The comment says the pane "holds the scope open too — otherwise the board unmounting would cancel a fetch this pane is still waiting on", but for a Core-owned project the board retains `(projectId, coreId)` while the pane retains `(projectId, null)`; the pane registers its cancel against the untagged `["projects", id, "tasks"]` key that no Core pane reads. Latent only because `TerminalPane` currently lives inside `SessionGrid` inside the route and cannot outlive the board. `TerminalPane` already has `descriptor.coreId` to hand (it builds `tasksCacheKey(project.id, descriptor.coreId)` a few lines away), so threading it through would make the retain match the bucket — or, if that is 397/401 territory, narrow the comment to say what is actually true.

### Notes

- The DOM assertions at `late-pin-materialize.test.tsx:181-183` pass with the guard fully reverted — the discriminating assertions are the cache ones at 188-192. That is fine as proof (an empty cache is strictly stronger), but the test's headline sentence is carried by lines that do not fail on revert; worth a comment there so a future reader does not trust the wrong line.
- All 12 CI checks are green on `e7384b1`. Two full local `vitest run` passes on this box each produced a different small set of failures (`cores-settings-pairing`, `project-dialog-path-immutable`, then `first-run-gate`, `project-bar-group-filter-ring`, `live-read-path`); every one of them passes in isolation — load-induced timeouts on the reviewing machine, unrelated to this change.
- Still a draft; leaving it that way.

Verdict: **changes requested** — finding 1 only. Both acceptance criteria are met and the mechanism is the right one; the fix is small and local.

qcentic-dev:
> In reply to the review of `e7384b1` ([#pullrequestreview-5099667962](https://github.com/actana/control/pull/471#pullrequestreview-5099667962)) — changes requested.

Thank you for the repro; it is exactly right, and it reproduced on the first try
as a test. Pushed as `48ae1af`, on top of a plain merge of the base (`53ee5b8`,
no conflicts, no rebase, no force push).

## Blocking — finding 1: the count outliving its own list

Fixed with the per-fetch generation you asked for, not a snapshot.

`visible-project-scope.ts` now keeps a **visit generation** per scope, bumped every
time that scope's viewer count falls to zero. `watchProjectScope` stamps the
generation a read starts in and compares on landing, so the question it answers is
*"does this answer still belong to the visit that asked for it?"* rather than *"is
this scope on screen now?"*. During A → B → A → B, B's read #1 stamps generation 0,
leaving bumps it to 1, and read #1 is stale on arrival however many times the
operator has since come back. Read #2, stamped 1, is not.

I bumped on the fall to zero rather than on every retain/release. Bumping on every
retain would mark a read stale merely because a second reader mounted mid-visit
(the hidden-session bar, a pane), which reintroduces the same inconsistency in a
new place — the list fresh, the count skipped — while the operator has not gone
anywhere. Zero is where the visit actually ends, and that is the only transition
that has to invalidate anything.

The new test is `late-pin-materialize.test.tsx` →
**"does not let B's abandoned read overwrite the count of the read that replaced
it"**. It drives your interleaving with per-call gates so B's two `listTasks` are
separate promises: read #1 in flight → click A (cancelled and reverted) → click B
(read #2) → read #2 answers `archivedCount: 7` → read #1 answers
`archivedCount: 99`. With the snapshot restored it fails as
`expected 99 to be 7`; with the generation the bucket holds 7 and the list holds
read #2's rows.

The PR body and the commit message no longer claim the count "cannot outlive the
list it came in on" as a property of cancellation — the body now says plainly that
a cancelled fetch's promise still resolves and that the fetcher's own write needs
the stamp.

## Finding 2 — accumulating `onLeft` closures

Taken, but implemented by **keying** rather than by deleting on release, because
deleting before the zero check reintroduces the AC1 bug:

> board's `useProject` (key P) and board's `useTasks` (key T) are two retains of
> one scope, and their cleanups run one after the other inside a single commit. On
> A → B → A the P cleanup runs first, sees `viewers` still 1, drops P's callback;
> the T cleanup then takes the scope to zero and fires what is left — T only. P's
> fetch is never cancelled, and B's project row lands on A.

So `onLeft` is now a `Map` keyed by the reader's query-key hash. A pane mounting
and unmounting through a visit replaces its entry instead of adding another, which
is the harm you named — unbounded accumulation and the same cancel running over and
over — and at zero each distinct key is still cancelled exactly once, all together.
Covered by **"holds one callback per key however often a reader remounts"**: five
pane mount/unmount cycles behind a live board, one cancel on the way out.

If you would still prefer the literal drop-on-release, it needs a per-key refcount
plus deferring the drop past the current commit, and I did not think that was worth
the machinery for a set bounded at three keys — happy to be told otherwise.

## Finding 3 — `useTask` hard-coding `coreId: null`

Taken as the threading, not the comment narrowing. `useTask` now accepts
`{ coreId }` and `TerminalPane` passes `descriptor.coreId`, so the retain matches
the bucket. As you spotted, the pane was also *subscribing* to the untagged bucket
while building `tasksCacheKey(project.id, descriptor.coreId)` two lines below — so
for a Core Session the per-row subscription was reading a list nothing fills and
the header only ever showed the open-time snapshot. Both are now the tagged bucket.
`src/components` is 18 files / 177 tests green.

## Note — the DOM assertions that pass on revert

Correct, and there is now a comment above them saying so: they are the operator's
own sentence, but the discriminating assertions are the three cache reads below
them. A future reader will not trust the wrong line.

## Verification on `48ae1af`

| check | result |
| --- | --- |
| `vitest run src/queries src/lib` | 73 files, 654 tests, green |
| `vitest run src/components` | 18 files, 177 tests, green |
| `tsc --noEmit` (`@actana/panel`) | clean |
| `eslint` on the five changed files | clean |
| commitlint over `c76cae5..HEAD` and `aa5df9d..HEAD`, repo config, as CI runs it | 0 problems, 0 warnings |
| negative control | reverting only `watchProjectScope` to the snapshot fails the new interleaving test at `99` vs `7` |

Still a draft, base unchanged, nothing merged.

mrfqcentic (approved):
## Round 2 review of #471 against issue #381 (B4) — approve

Reviewed at head `48ae1af` (round 1 was `e7384b1`; `e7384b1` is an ancestor, so the fix arrived on top of a plain merge of the base — no rebase, no force push). Everything below was checked by reading the enforcing line and by re-driving my own round-1 probe, not from the PR description.

### The round-1 blocker is closed

**`packages/panel/src/lib/visible-project-scope.ts:144-148` is now a per-fetch visit stamp, not a visibility snapshot.**

```ts
export function watchProjectScope(projectId: string, coreId: string | null): () => boolean {
  const token = projectScopeToken(projectId, coreId);
  const startedAt = scopes.get(token)?.generation ?? 0;
  return () => (scopes.get(token)?.generation ?? 0) !== startedAt;
}
```

`generation` is bumped in `retainProjectScope`'s release at `visible-project-scope.ts:110`, at the fall to zero and *before* the `onLeft` callbacks run — so a fetch already past the point of no return is stamped stale for good. The stamp is taken at `queries/index.ts:162`, the first statement **inside** `queryFn`, so it is one stamp per fetch, not one per render or per module. The gate is at `queries/index.ts:184-186`, wrapping the only fetcher-own write in that fetcher.

**I re-drove my own round-1 probe** (a probe file of my own, not the PR's test) — pin B uncached, click A, click B, read two answers `archivedCount: 7`, read one then answers `archivedCount: 99` while B is on screen:

- count bucket holds **7**, the list holds read two's rows (`t-b2` only), the board shows read two's session. Read one is dropped in both halves.
- The **list and the count are mutually consistent** — both come from read two — which is what the PR's contract claims.
- Reverting only `watchProjectScope` to the snapshot fails my probe at `expected 99 to be 7`, exactly where round 1 predicted.

**The stamp is genuinely per-fetch.** A three-visit probe — read one (visit 0), read two (visit 1), read three (visit 2), then answers in the order 3, 2, 1 while B is visible throughout — keeps the count at `7` from read three; both older reads are refused. Under the snapshot the same probe ends at `99`. At unit level, two `watchProjectScope` calls inside one visit hold two independent stamps, and a stamp taken after a return is `false` while the older ones are `true` — so it is per-call state, not a shared flag. No wedge: after all of that, leaving and returning still issues a fresh read.

I also probed the mirror window — leaving *between* the fetcher's count write and react-query's commit of the list — across six microtask widths. The list always commits with the count; I could not open it.

### The tests now fail on revert

Round 1's note was that the A → B → A test's headline DOM assertions passed on revert. Re-checked at this head:

- Reverting `queries/index.ts` to the base fails `late-pin-materialize.test.tsx:218` (B's project bucket is `{…Bravo}`, not `undefined`) and `:252` (`expected 1 to be 2` — with no cancel there is no second read to interleave). Both new tests fail; the suite is genuinely load-bearing.
- Reverting only `watchProjectScope` to the snapshot fails `late-pin-materialize.test.tsx:281` at `expected 99 to be 7`, plus `visible-project-scope.test.ts:117` and `:136`.
- The DOM assertions at `:211-213` still pass on revert, and the comment at `:206-210` now says precisely that and points at the cache assertions. That is what I asked for.

### Acceptance criteria

**AC1 — "Clicking A then B then A never leaves B's sessions painted on A's URL." — MET.** Enforced by `queries/index.ts:380-393`: `retainProjectScope` while a reader is mounted, and at zero `queryClient.cancelQueries({ queryKey, exact: true })` behind an `isFetching` check, `revert: true` by default. Proven by `late-pin-materialize.test.tsx:218-222` — B's project bucket, task bucket and archived-count bucket all `undefined` after the return to A — which fails on revert. The count half of it is now closed too, so nothing of B's is parked anywhere.

**AC2 — "A cold pin still loads; it just cannot win after you have left it." — MET.** No production `staleTime` / `gcTime` / `refetch*` / `enabled` line changes anywhere in the diff (the only occurrences are comments and the test client's `staleTime: 30_000` at `:163`). Only an in-flight fetch is ever cancelled; a settled query keeps its rows (`:330-346`). Cold-pin load proven at `:284-301`, and through a StrictMode double-mount at `:303-328`. My own probe adds: after three abandoned reads, a return still fetches and paints — no wedge.

### Round-1 low findings

**Finding 2 (accumulating `onLeft` closures) — taken.** `onLeft` is a `Map` keyed by the reader's query-key hash (`visible-project-scope.ts:57`, `:97`; `queries/index.ts:381-383`), so a pane remounting through a visit replaces its entry instead of adding one, and each distinct key is cancelled exactly once at zero (`visible-project-scope.test.ts:80-93` — five remounts, one cancel). I asked for a drop-on-release instead; the reply is right that dropping before the zero check would leave the first reader's fetch uncancelled when the second takes the scope to zero in the same commit — the AC1 regression. The keying addresses the harm I actually named (unbounded accumulation, repeated redundant cancels) without that risk. Accepted as implemented.

**Finding 3 (`useTask` hard-coding `coreId: null`) — taken as the threading.** `useTask` now takes `{ coreId }` (`queries/index.ts:459-470`) and `TerminalPane.tsx:530-532` passes `descriptor.coreId`, matching `tasksCacheKey(project.id, descriptor.coreId)` at `TerminalPane.tsx:544`. It is the only `useTask` call site in the tree. The retain now matches the bucket, and the header subscription reads the list that actually fills.

### Non-blocking notes (do not hold the PR)

1. `packages/panel/src/lib/visible-project-scope.ts:118` — `isProjectScopeVisible` no longer has a production caller; it is used only by the module's own tests now that `watchProjectScope` compares generations. Fine to keep as a documented accessor, but it is dead in `src/` and could go.
2. `packages/panel/src/queries/index.ts:365-366` — "Anything a fetcher writes for itself is guarded by `watchProjectScope` instead" is slightly wider than the code: `projectQueryOptions`' fire-and-forget `api.pruneProjectPresentation` at `queries/index.ts:83-86` is also a fetcher-own write and is not stamped, so a left read still prunes filing rows from its own stale project list. Pre-existing on the base and unchanged by this PR (that call fired before the guard too, and the window is small), so it is a wording nit here, not a defect this PR introduces — worth narrowing the sentence, or gating that call the same way in whatever ticket owns it.
3. `packages/panel/src/queries/index.ts:377-393` — the guard cannot tell a departure from a *transient* dip to zero viewers: React runs every cleanup in a commit before every setup, so a reader set that unmounts and remounts inside one commit cancels the in-flight read and bumps the visit generation although the operator never left. It cannot happen from a props change (the effect deps and the scope token are the same `(projectId, coreId)`), and there is no `<StrictMode>` in `client.tsx`, so today it is only reachable in the PR's own StrictMode test — which is why that test has to loop over `callCount(...)`. It self-heals (the re-subscribe refetches). Worth one line naming the invariant it rests on — that a route-level reader outlives every swap — because a refactor that moved `useTasks` out of `projects.$id.tsx` would turn it into a real double-fetch on every cold pin.
4. The acceptance-criteria comment on this PR still carries its round-1 text and figures ("cannot outlive the list it came in on" attributed to cancellation; 67 files / 596 tests; CI for `e7384b1`). The follow-up reply supersedes it, but the AC comment is the record a promoter reads — worth an edit.

### Verification on this box

| check | result |
| --- | --- |
| `vitest run src/queries src/lib` at `48ae1af` | 73 files, 654 tests, green |
| `vitest run src/components` at `48ae1af` | 18 files, 177 tests, green |
| `tsc --noEmit` (`@actana/panel`) | clean |
| `eslint` on the five changed files | clean |
| negative control: `watchProjectScope` → snapshot | 3 tests fail, `99` vs `7` as predicted |
| negative control: `queries/index.ts` → base | both new tests fail (`:218`, `:252`) |
| my own probe (round-1 interleaving, three-visit, per-call stamp) | passes on head, fails on the snapshot |
| `/code-review` over the PR diff against the base | no correctness finding in the retain/cancel/generation mechanism; two low notes, folded in above |

A separate `/code-review` pass over this diff traced the mechanism against the installed `@tanstack/query-core@5.101.2` and agrees: `retryer.cancel()` rejects and runs `onCancel`, which restores the pre-fetch snapshot at `fetchStatus: "idle"` (so a re-subscribe in the same commit starts a real new fetch and the late `resolve()` is a no-op), `#revertState` is captured before the fetch dispatch and restores `isInvalidated: true` (so an invalidation-driven refetch cancelled on the way out still refetches on return), and `exact: true` matters on both calls because `["projects", id]` is a prefix of `["projects", id, "tasks"]` — it is correctly set on both.

### Base movement

The base branch has moved since the branch's merge of `c76cae5`: `fix/operator-pins-sessions-drop-cli` is now `3ceb0d3` (#457, Core pin dots). `git merge-tree --write-tree 3ceb0d3 48ae1af` succeeds with no conflict, and none of #457's nine files overlaps this PR's five; its `core-pins-engine.ts` reads `bridge.listTasks(core.id)` directly and never touches the react-query buckets this guard cancels. On the merged tree, `vitest run src/queries src/lib src/shared src/components` is **104 files / 984 tests green** and `tsc --noEmit` is clean. No re-merge needed for correctness.

CI on `48ae1af`: 18 pass, 6 skipping (train-only), 0 failing.

### Checklist

- [x] Blocking finding 1 (visibility snapshot → per-fetch visit stamp) — fixed, enforcing line read, round-1 probe re-driven
- [x] Stale read can no longer overwrite the fresh archived count — proven, and it fails on revert at `99` vs `7`
- [x] Token is per-fetch, not another visibility snapshot — stamped inside `queryFn`, proven by a three-visit probe and per-call unit probe
- [x] List and count mutually consistent — both come from the read that replaced the abandoned one
- [x] Low finding 2 (`onLeft` accumulation) — taken by keying; the reply's reason for not dropping on release is correct
- [x] Low finding 3 (`useTask` `coreId`) — taken by threading `descriptor.coreId`; only call site updated
- [x] Tests now fail on revert (round 1 found the DOM ones did not; the discriminating cache assertions do, and are now labelled)
- [x] AC1 met · AC2 met
- [x] Merges cleanly on the moved base (`3ceb0d3`), merged tree green
- [x] Boundaries respected; no `staleTime` / `gcTime` change; no force push
- [ ] Still a draft — leaving it that way, as in round 1

Verdict: **approve**. The blocker is closed by the mechanism that was asked for, both low findings were taken (one of them better than I proposed), and the negative controls now discriminate. The three notes above are optional.

Issue #474, opened 2026-09-02
S6 follow-up: the All leg still drops the ring on an open unpinned project


qcentic-dev:
Follow-up to #378, split out of the review of #470 (review [5095900838](https://github.com/actana/control/pull/470#pullrequestreview-5095900838)).

#470 keeps a selection affordance for the project the operator is on when the **group-scoped** rail would hide it: `ProjectBar.tsx`'s scoped branch appends a trailing off-group cluster holding that project, so the ring has somewhere to sit and nothing navigates. That covers group → group and group → Ungrouped.

It does **not** cover group → **All**, which is the same operator gesture on the same GroupSwitcher.

## What the operator sees

The scoped rail is the group's *workspace*: `getGroupRailCluster` returns the group's pinned projects **plus its unpinned ones**, so an operator can open an unpinned project straight from the rail. Switch the GroupSwitcher to All and the tile is gone and the ring is `null`, while the page is still on that project — the #378 symptom exactly.

Reproduced by the reviewer at `50e07aa` with an unpinned project `p_unpinned` in group A, opened from A's workspace rail:

| destination | url | tile | ring |
| --- | --- | --- | --- |
| group B | `/projects/p_unpinned` | present | on it ✅ |
| **All** | `/projects/p_unpinned` | **ABSENT** | **null** ❌ |

## Where

`packages/panel/src/components/views/ProjectBar.tsx` — the `rail` memo. The All leg returns early at `if (!groupScoped) return { clusters: clusterPinnedByGroup(...), label: "Pinned projects" }`, before the off-group cluster is appended. The comment in that memo now states this limit explicitly rather than claiming an invariant the code does not hold.

## Why it was not folded into #470

Not a missing branch — a design change, and #470's reviewer explicitly did not ask for it to be widened unilaterally. In All, `usesDirectRailProjectShortcuts` is false whenever a real group exists, so `showClusterLabels` is **true** and every cluster renders a real header. A synthetic trailing cluster would therefore:

- render `data-cluster-header` + `data-group-handle`, which puts it in `startProjectPointerDrag`'s frozen row snapshot and `startGroupPointerDrag`'s handle list — it becomes a drop slot and a draggable group, both of which the scoped version is careful *not* to be;
- take group digit `clusterIndex + 1` in its header keycap, while `__root.tsx` resolves chords from `getRailClusters`, which does not know the cluster exists — the badge/chord disagreement of #379, reintroduced.

[...]

PR #476, opened 2026-09-03
fix(panel): mixed-rail pin reorder goes through the Core path

qcentic-dev:
Fixes the S3 hunt in #382: dragging two pins, or a Core pin into another group, played the animation and then reverted with an error toast whenever any Core pin was on the rail.

## What was wrong

The rail is one strip of tiles holding this Panel's own pins and the pins of every Core it is connected to. Both writes it made were Panel-only:

- `PATCH /api/projects/pinned-order` validated the order against the Panel's `projects` table and rejected any id that was not in it — so a mixed rail was rejected **whole**, taking the Panel rows' reorder down with the Core rows';
- `PATCH /api/projects/:id`, used for the regroup half, 404s for a project the Panel holds no row for — the same 404 issue 98 fixed for the Edit-project dialog, still live on the drag.

## The split — worked out by reading the code, not assumed

A rail row is Core-owned exactly when it carries a `coreId`. Rows from `useRemotePinnedProjects` / `core-pins-engine` are tagged with the Core that answered for them; the Panel's own rows (`useProjects`) are the untagged remainder. `ProjectBar` now asks this **per row**, not per rail.

| row | order | group |
| --- | --- | --- |
| Panel-owned (`coreId == null`) | `PATCH /api/projects/pinned-order` — the Panel reorder API, unchanged as the door | `PATCH /api/projects/:id { groupId }` |
| Core-owned (`coreId != null`) | `PATCH /api/project-presentation/pinned-order` (new) | `PATCH /api/project-presentation/:id { groupId }` — what `save-project-edits` already uses |

Two rows that look like exceptions and are not:

- **Unpinned tiles in a group workspace** are not reorderable at all (no `data-pinned-item`, computed alphabetical order), so they never reach a write path.
- **The trailing off-group tile #378 appends** is a selection affordance, not a pin of this rail — see below.

## Why the Core's own database is the wrong place for the slot

Pin state *is* a Core fact (ADR 0005), so the symmetric answer would be a new core-link op writing `projects.pinned_order` on the Core. Rejected: **the rail spans Cores.** A slot in a sequence that interleaves three machines' pins with the Panel's own is not a fact any one machine can hold — two Cores would each believe they owned slot 2, and the merged list would fall back to tie-breaking on `createdAt`. The slot is the operator's own filing of somebody else's Project, in exactly the sense ADR 0022 already means, so it joins group / card image / launch URL on the presentation row. Whether a Project is pinned is untouched and stays Core-owned.

## One numbering space

This is what makes acceptance 1 hold rather than half-hold. Both halves persist **the row's index in the whole rail**:

- `reorderPinnedProjects` is handed the entire rail order, numbers its own rows by their index in it, and skips ids it holds no row for (they are passengers, holding their slot in the numbering);
- the Core-owned rows take those same indices, all of them on every reorder — a reorder shifts the rows either side of the gap too, and a Core pin left on a stale index would sort itself back out of place on the next reload.

Numbering each owner densely over its own rows was the easier shape and the wrong one: it leaves no integer free to place a Core's pin *between* two of the Panel's, which is the ordinary shape of a mixed rail.

Reading it back: `projectRowFromSnapshot` still reports `pinnedOrder: null` (the core-link snapshot has no such field and should not grow one), and the rail's own fan-out `lib/core-pins-engine.ts` joins the slot on from the presentation row it has already read for the group. Folding that into the mapper belongs to whoever next owns `shared/projects.ts` — untouched here, it is #381/#401's.

## Optimism, and where the two halves of the rail live

The rail draws one strip but its rows are in two stores, so a move has to be applied to both before the first request goes out:

- the Panel's rows into the React Query `projects` cache, as before;
- every Core-owned row's slot — and the group, when the row that moved is Core-owned — into the fleet engine's snapshot, via `applyCorePinFiling`.

Without the second, a Core-owned tile dragged to the top was re-rendered at the bottom the instant the drop settled (its `pinnedOrder` still `null`, which the comparator reads as last) and stayed there for both HTTP legs plus a `listProjects` and a `listTasks` per Core; with Shift+Arrow there was no settle animation to hide it at all. The overlay outranks the server's answer until `settleCorePinFiling` takes it down behind a read, so a poll landing mid-write cannot repaint the rail with filing the operator has already replaced. It comes down in `finally` *after* the gesture guard is released, so the next Shift+Arrow is not swallowed waiting on a fan-out.

"Behind a read" is `run.fresh()`, not `run()`, and the difference is load-bearing. The coalescing runner records a request it cannot serve and resolves on the next microtask, so with a pass already in flight — the 15s poll, a reconnect, or any task/session/pty/project-list event, which on a live fleet is continuous — the overlay used to come down having read nothing, under a pass that had read the table *before* the write. `fresh()` resolves only once a pass that started after the call has settled. The settle then republishes, re-filing the settled rows from the presentation that pass read: a no-op on the happy path, and what makes both halves of the rail agree again when the write failed and the Panel half has reverted with the query cache. The overlay is keyed by a token `applyCorePinFiling` hands back, so a settle can only take down its own gesture's overlay — `saveSeq` cannot guard it, because it is released first on purpose.

The same distinction fixes a permanent version of the revert: a Core off the link keeps its pins deliberately, but kept them with the filing baked in at the last successful read. Core facts survive a Core's absence; the Panel's own filing over them does not get to, so `withFiling` re-reads and re-joins it every pass.

## Ordering and failure — stated precisely

`reorderPinnedProjects` is the only call here that can refuse a move, so it goes first and nothing else is committed before it answers; its response is applied to the cache before the group write, because that response predates it.

The Core-owned slots then go in one request, in one transaction, so that half cannot land partially. **This does not make the reorder as a whole atomic:** it is two requests against two tables and there is no transaction spanning them, so if the Panel's order write commits and the Core half then fails, the rail is half-applied under a failure toast — the client reverts what it can and re-reads the rest. Closing that needs one endpoint owning both tables, which is filed in #478 rather than done here.

## Scope

- `ProjectBar.tsx`'s off-group cluster (#378) is left exactly as it is. It carries no `data-pinned-item`, so it is in neither `startProjectPointerDrag`'s frozen row snapshot nor the keyboard order, and it rides through the persisted rail order as the passenger it already was. Two tests pin that.
- The group → **All** leg (#474) is **not** taken here.
- Untouched, by ticket: `use-fleet.ts`, `shared/projects.ts`, `queries/index.ts`, `visible-project-scope.ts`, `projects.$id.tsx`.
- Deferred to #478: the shared-numbering reconciliation pass (drifted slots, and a slot stranded by an unpin this Panel never saw), and pinning the `project-presentation` routes in `api-auth.test.ts`'s `PROTECTED_ROUTES`. Its third item — `listProjectPresentation` degrading to empty rather than to "unknown" — is **taken here** rather than deferred: #382's own fix newly exposed a disconnected Core's remembered pins to it, so a failed read now holds the last presentation map instead of filing every row null.

## Migration

`project_presentation` gains `pinned_order INTEGER`, via the CREATE TABLE and an idempotent `ensureColumn` for databases that already exist. NULL until the operator first reorders a rail with a Core pin on it, which sorts exactly as it did before.

## Verification

`pnpm typecheck` and `pnpm lint` clean (0 errors), and the full `pnpm test` green. New coverage:

- `lib/__tests__/core-pins-rail-slot.test.tsx` (10) — the engine driven for real against a fake Core and a mutable presentation table: the slot reads back, the tile moves on the write with no fan-out, the overlay outranks a mid-write poll and is handed back afterwards, a group change rides along, and a disconnected Core's remembered pins are re-filed from the presentation just read. Four more hold a read open to drive a settle **mid-pass**, the failure path, two overlapping gestures, and a failed presentation read; all four fail against the previous head.
- `views/__tests__/project-bar-mixed-rail-reorder.test.tsx` (14) — the split, the shared numbering, the optimistic call and the rendered order, the two off-group cases, a **real pointer drag** covering the regroup half, and the rail's unpin driven through its own context menu on a Core-scoped rail, for both a Panel-owned and a Core-owned row.
- `project-presentation.test.ts` (4), `projects.test.ts` (3), `pinned-project-order.test.ts` (2).

Refs #382, #378, #474, #98, #478. Amends ADR 0022.

https://claude.ai/code/session_011NqWGouGnv6G63A99L9zAv


qcentic-dev:
## Acceptance criteria — #382

- [x] **1. Dragging two pins, including a Core pin, keeps the new order after reload.**
  Both halves of the rail now persist the row's index in the **whole rail**, in one numbering space, so the merged list re-sorts into the operator's order rather than into two interleaved dense runs. The Panel's rows take their index through `PATCH /api/projects/pinned-order` (now handed the entire rail, numbering its own rows by their slot in it and skipping ids it holds no row for); the Core-owned rows take the same indices through the new `PATCH /api/project-presentation/pinned-order`, which writes them in one transaction. The read leg is what makes it survive the reload: `lib/core-pins-engine.ts` overlays `pinnedOrder` from the presentation row onto each Core pin, so `getPinnedProjects`'s comparator sorts Panel and Core rows against each other. Covered by `project-bar-mixed-rail-reorder.test.tsx` ("sends the whole rail to the Panel API and every Core pin's slot to presentation", "re-slots every Core pin, not just the one that moved"), by `projects.test.ts` ("numbers its own rows by their slot on a mixed rail"), and by the rail-slot block in `project-presentation.test.ts`.

- [x] **2. No error toast on a valid mixed-rail reorder.**
  The toast came from two rejections, and both are gone. `reorderPinnedProjects` no longer rejects an order naming a row it does not own — `validatePinnedReorder` now requires every pinned row of *ours* exactly once and no duplicates, and treats an id from no row here as a passenger (it still rejects an unpinned row of our own when told which ids are ours). And the regroup half no longer sends a Core id to `PATCH /api/projects/:id`, which 404s for a project the Panel holds no row for; a Core row's group goes to its presentation row, exactly as `save-project-edits` already files it. `expect(toastError).not.toHaveBeenCalled()` is asserted on every valid-reorder case in the new suite, plus a case pinning that no Core id reaches the Panel-only PATCH.

### Explicitly not taken here

- [ ] **The group → All leg (#474).** Left out deliberately: #474 records it as a design change, not a missing branch, and the ticket for this PR says not to take it. Nothing here touches the `rail` memo's All arm.
- [ ] **Folding the `pinnedOrder` overlay into `projectRowFromSnapshot`.** That is the tidier home for it, but `shared/projects.ts` belongs to #381/#401 in this train. The overlay lives in `lib/core-pins-engine.ts` instead; the ADR amendment says so and names the follow-up.

### Constraints from the ticket

- [x] Reorder and regroup go through the owner of each row, worked out from the code: a row is Core-owned exactly when it carries a `coreId` (rows from `useRemotePinnedProjects`); Panel rows are the untagged remainder. The full split is in the PR body.
- [x] The trailing off-group cluster #378 appends is not treated as a reorderable pin. It carries no `data-pinned-item`, so it is in neither `startProjectPointerDrag`'s frozen row snapshot nor `pinnedIdsRef`, and it rides through the persisted rail order as the passenger it already was. Two tests pin it: "keeps the off-group tile out of the order it persists" and "cannot reorder the off-group tile itself".
- [x] `use-fleet.ts`, `shared/projects.ts`, `queries/index.ts`, `visible-project-scope.ts` and `projects.$id.tsx` are untouched.
- [x] Base is `fix/operator-pins-sessions-drop-cli`; the PR is a draft and stays one. Nothing merged, `main` untouched, no force push.

mrfqcentic (changes requested):
# Review — PR #476 against #382

**Verdict: request changes.** The diagnosis is right, the ownership split is right, the persistence layer is right, and both written acceptance criteria hold *on the server*. What blocks is the client half: on a mixed rail the dragged/nudged Core tile still visibly returns to its old slot after the drop settles, and stays there for the length of a full fleet fan-out. That is the sentence in the issue — "Animation plays, then order reverts" — minus the toast. One further path makes the revert permanent rather than transient.

Everything below was checked by reading the enforcing line, not the PR body.

---

## What I verified, and it holds

**The per-row split is the correct axis, and `coreId` really is the discriminator.** Not taken on trust:

- `lib/use-fleet.ts:398-422` → `lib/core-pins-engine.ts:109-120`: every row a Core answers for is built with `coreId: core.id`. There is no other producer of tagged rows for the rail.
- `ProjectBar.tsx:88-93`: `projects = [...panelProjects, ...remotePinned]`. The Panel's own rows come from `useProjects()`, whose row type has no `coreId` column, so `project.coreId ?? null` at `ProjectBar.tsx:145-148` partitions the rail exactly.
- It is also the axis the codebase already uses for this table: `lib/save-project-edits.ts:46-72` branches on `if (!coreId)` and sends a Core row's `groupId` to `api.updateProjectPresentation`. `ProjectBar.tsx:416-424` now mirrors that call site for the drag, which is precisely the #98 fix the drag path never received.

**Both halves persist, in one numbering space.**

- Panel half — `server/services/projects.ts:374-395`. The index written is the row's index in the *whole* rail (`order[index]`), and ids with no row here are skipped rather than rejected. `written` correctly narrows the `project:updated` fan-out to rows that actually changed.
- Validator — `lib/pinned-project-order.ts:80-91`. The rewrite is not a loosening in disguise: it still demands every pinned row *of ours* exactly once and no duplicates, and `projects.ts:378` passes `panelProjectIds` so an unpinned row of our own still throws `not pinned`.
- Core half — `lib/api.ts:293-300` → `controllers/project-presentation.controller.ts:57-60` → `services/project-presentation.ts:98-105`.
- Read-back — `core-pins-engine.ts:119` overlays the slot; `pinned-project-order.ts:8-12` sorts both owners against each other in one space. `fleet-refresh.ts:31-44` (`sameSnapshot`) is a deep walk, so the new field does propagate rather than being swallowed as "no change".
- The shell's rail chords read the same engine snapshot (`routes/__root.tsx:266-268`), so Cmd-digit and the rail cannot disagree about which tile is in which slot.

**The new endpoint is sound.** Transactional (`project-presentation.ts:99-104`, `.transaction(...).immediate()`, same shape as `projects.ts:375`); no filesystem side effects inside the transaction. Authorised identically to its siblings — `api-router.ts:125-133` is default-deny and the path is absent from `ANONYMOUS_ROUTES`. Input is zod-validated (`controller:24-32`) at exactly the strictness the sibling routes use (`idParam` is itself `z.string().min(1)`). Route ordering is correct and matches the `/api/projects/pinned-order` precedent at `api-router.ts:253-255`.

**The off-group cluster (#378) is not treated as a reorderable pin.** `ProjectBar.tsx:1256` withholds `data-pinned-item` when `isOffGroup`, which keeps it out of the drag snapshot selector at `:466`; `:306-311` filters `OFF_GROUP_CLUSTER_KEY` out of `pinnedIdsRef`; `:1314-1318` and `:1266-1268` refuse the keydown and pointerdown. Two tests pin it. It rides the persisted order as a passenger and keeps its slot — correct.

**Both gestures are covered by the split**, since drag (`:614`) and Shift+Arrow (`:643`) both funnel into the same `persistProjectOrder`.

**Scope is clean.** `use-fleet.ts`, `shared/projects.ts`, `queries/index.ts`, `visible-project-scope.ts`, `projects.$id.tsx` are untouched. All 24 CI checks pass. Three commits split server / client / docs cleanly.

**The doc commit is accurate.** `project_presentation.pinned_order` genuinely exists (`db/schema.ts:89-96`, `schema-bootstrap.ts:293-297` + `:451-452`), `projectRowFromSnapshot` genuinely still returns `pinnedOrder: null` (`shared/projects.ts:158`), and the overlay genuinely lives in `core-pins-engine.ts`. `ensureColumn`-without-a-migration matches the `groups.sort_order` precedent, so no migration file is owed.

---

## Blocking

### 1. `components/views/ProjectBar.tsx:398-411` — the moved Core tile springs back until a full fleet fan-out completes

The comment at `:398-400` states the limitation accurately, but not its consequence. `setQueryData` writes only the React Query `projects` cache, which holds Panel rows; a Core row's `pinnedOrder` lives in the `core-pins-engine` module snapshot, and `core-pins-engine.ts` exports no optimistic setter — `refreshCorePins` (`:204-206`) is the only way it changes, and it runs the whole fan-out.

Trace a first mixed-rail reorder, rail `[P1(0), C1(null)]`, dragging `C1` to the top:

1. `:601-620` — the settle animation runs for `GROUP_SETTLE_MS` (200ms), *then* `setProjectDrag(null)` and `persistProjectOrder` fire in the same tick, so React renders them together.
2. That render reads `sortedPinned` with `P1` optimistically at 1 and `C1` still `null` → `Number.MAX_SAFE_INTEGER` in `comparePinnedProjects`. Rendered order: `[P1, C1]`. **The tile the operator just dragged to the top is drawn back at the bottom.**
3. It stays there through `await api.reorderPinnedProjects` + `await api.reorderCorePinnedProjects` and then through the un-awaited `refreshRemotePinned()` at `:434` — which is `listProjectPresentation` plus, for *every* Core, a sequential `listProjects` then `listTasks` over the core-link. For a Core on another machine that is several round trips.

On a second reorder it is not even a clean revert: `P1` optimistically takes 1 while `C1` still holds its old 1, so the two tie and `createdAt` picks the winner.

Shift+Arrow is the worse case, because there is no settle to hide it: `:635-646` calls `persistProjectOrder` directly, so the operator presses the key and the Core tile does not move. And `pinnedIdsRef.current` (`:306-311`) is recomputed from that stale-mixed render, so a follow-up gesture in the same window is computed against an order that is not the persisted one, while `:637`'s `reorderSavingRef` guard silently drops presses during the HTTP leg.

This is the reported symptom in transient form on the ordinary happy path, and the new suite cannot catch it because it mocks `useRemotePinnedProjects` wholesale. Please close it — overlaying the just-written slots onto the engine snapshot (a small optimistic setter beside `refreshCorePins`) is the shape that fits the existing design.

### 2. `lib/core-pins-engine.ts:91` — with a Core momentarily disconnected, the revert is permanent, not transient

`readCore` returns `remembered` for a Core that is not `connected`, so its pins stay on the rail and therefore land in `sortedPinned`, in `corePinSlots` (`ProjectBar.tsx:386-389`) and in a `reorderCorePins` write that needs no Core and succeeds. The `refreshRemotePinned()` that follows then republishes those same remembered rows, whose `pinnedOrder` was baked in at `:119` during the last *successful* read. The tiles return to the old order, no toast is shown, and they stay wrong until that Core reconnects — even though the database holds the right slot. Either overlay the fresh presentation onto remembered rows, or leave a disconnected Core's rows out of the write.

---

## Non-blocking, worth a follow-up

- **`ProjectBar.tsx:416-424` — the group write precedes the validating order write.** The comment at `:425-426` is true of *slots*, but a group change is already committed by then. If `reorderPinnedProjects` throws (another tab unpinned a Panel row), the toast says the move failed while `refreshRemotePinned()` in the catch re-clusters the tile into its new group. Pre-existing ordering, newly reachable for Core rows.
- **`ProjectBar.tsx:428` — the two halves are not atomic across the boundary.** If the Core write fails after the Panel write committed, the catch restores `previous` and then `invalidateProjects()` refetches the *committed* new Panel order, so the operator gets a half-applied rail under a failure toast. The PR body's "a reorder cannot land half of them" is true within the Core half only; worth narrowing the wording either way.
- **`server/services/projects.ts:349` (and `:330`) — `nextPinnedOrder` is now under-numbered.** It maxes over Panel rows alone, but `projects.pinned_order` is now a sparse index into a rail shared with Core pins. Pinning a new Panel project can hand it a slot at or below an existing Core pin's, so it lands mid-rail instead of at the end.
- **`server/services/project-presentation.ts:98` — a slot is never cleared on unpin.** `reorderCorePins` is its only writer and a Core-side pin toggle never reaches it, so unpinning and re-pinning a Core project returns it to its old slot, where a re-pinned Panel row goes to the end. `pruneProjectPresentation` only sweeps deleted projects.
- **Stale slots survive a Core's absence** (`ProjectBar.tsx:386-389`). Renumbering while a Core is off the rail leaves its old slots untouched, so its pins can tie with Panel rows on reconnect. The shared numbering space has no reconciliation pass.
- **The regroup half has no test.** The new suite drives Shift+Arrow only, which always passes `groupChange: null`, so the `updateProjectPresentation` branch at `:417-420` — the literal second half of the issue's repro — is never exercised. `expect(updateProject).not.toHaveBeenCalled()` in "never sends a Core id to the Panel-only project PATCH" is vacuous for that reason. One case calling `persistProjectOrder` with a `groupChange` on a Core row would cover it.
- **Stale prose on the fourth field.** `lib/api.ts:261-265` and `server/services/project-presentation.ts:13-25` still enumerate "group, card image and launch URL" after the third commit made the slot a fourth field.
- **`api-auth.test.ts`** lists no `project-presentation` route in `PROTECTED_ROUTES` (pre-existing). The gate is default-deny so the new route is covered, but adding it would pin it.

---

## Checklist

| | |
|---|---|
| AC1 — drag two pins incl. a Core pin, order survives reload | **Met** (server + read path verified end to end) |
| AC2 — no error toast on a valid mixed-rail reorder | **Met** |
| Per-row split is the correct axis | **Met** |
| `coreId` is genuinely the ownership discriminator | **Met** |
| Both halves persist | **Met** |
| A failure in one half cannot silently revert the other | **Partially met** — no silent revert, but a half-applied rail is reachable |
| New endpoint transactional | **Met** |
| New endpoint authorised as its siblings are | **Met** |
| New endpoint validates its input | **Met** (shape; Core-ownership trusted, as its siblings do) |
| Off-group cluster (#378/#470) not a reorderable pin | **Met** |
| Shift+Arrow covered as well as drag | **Met** for routing; **not met** for the optimistic path (blocking #1) |
| No parallel-ticket files touched (#381/#401) | **Met** |
| Doc commit true of the code that landed | **Met** |
| Operator no longer sees the tile revert | **Not met** — blocking #1 and #2 |

qcentic-dev:
Thanks — both blocking findings were right, and the trace on B1 was the whole diagnosis. Pushed as four commits on top of a plain merge of the base (#471, #462, #472 landed while this was open; no conflicts). `1fa46fe` is the head.

## Blocking

**B1 — the moved Core tile springs back until a full fan-out completes.** Fixed as suggested: `core-pins-engine` now exports an optimistic setter beside `refreshCorePins`.

`applyCorePinFiling(filing)` overlays the just-written slots — and the group, when the row that moved is Core-owned, since that lives in the same snapshot and snaps back the same way — onto the engine's rows synchronously, before `persistProjectOrder`'s first await. `settleCorePinFiling(ids)` takes the overlay down behind a read.

Three details beyond the shape you named:

- The overlay **outranks** the server's answer while it is held, rather than only seeding the snapshot. A poll ticking mid-write reads presentation the write has not reached yet, and without that precedence it would repaint the rail with the slot the operator had already replaced.
- It comes down in `finally` **after** `reorderSavingRef` is released, not before. You flagged that guard silently dropping presses during the HTTP leg; making it also span the fan-out would have been worse, so the guard now covers the two requests only.
- `pinnedIdsRef` is recomputed from a render that now has both halves optimistic, so the follow-up-gesture problem goes with it.

**B2 — with a Core momentarily disconnected the revert is permanent.** Took the first of your two options: overlay fresh presentation onto remembered rows. Leaving a disconnected Core's rows out of the write would have meant a reorder that visibly moves its tile does not persist, which trades one lie for another.

`readCore` now runs remembered rows through the same `withFiling` join as fresh ones. The rule it states out loud is the one the bug came from: **Core facts survive a Core's absence; the Panel's own filing over them does not get to** — it is re-read every pass and joined on afresh.

Both are covered by a new suite that drives the real engine against a fake Core and a mutable presentation table (`lib/__tests__/core-pins-rail-slot.test.tsx`, 6 cases). I checked it fails without the fixes: 4 of the 6 go red.

## Non-blocking, all taken

- **Group write before the validating order write.** Reordered: `reorderPinnedProjects` goes first and nothing else is committed before it answers. Its response is now applied to the cache *before* the group write too — that answer predates the group write, so landing it last put the old group back on screen until the next refetch. A test pins the ordering.
- **Over-claimed atomicity.** Narrowed in the PR body, the ADR amendment and `reorderCorePins`' doc-comment. All three now say the same thing: all-or-nothing *within the Core half*, not across the two, and "before any slot has been written" was true of slots and not of the group write — which is moot now that the group write is last, but the wording was wrong either way.
- **`nextPinnedOrder` under-numbered (`projects.ts:349`, `:330`).** `nextRailSlot` maxes over both halves. Both live in this database, so no Core is asked anything. Covered on both the toggle and the create-pinned path.
- **A slot never cleared on unpin.** The rail's own unpin now clears it. An unpin this Panel never witnesses still strands one — that is genuinely the reconciliation pass, and it is in the deferred issue rather than half-solved here.
- **The regroup half had no test, and the `updateProject` assertion was vacuous.** A real pointer drag now covers it, against a rail laid out by hand because every rect in jsdom is zero. Three cases: the group is filed as presentation and `PATCH /api/projects/:id` is never touched, the order write precedes the group write, and the tile is drawn in its new cluster before the write returns. The suite's engine mock is a store rather than a value now, which is what makes the rendered order assertable — mocking it wholesale is exactly how the first round missed B1.
- **Stale prose** at `lib/api.ts:261-265` and `services/project-presentation.ts:13-25` — both say four fields.

## Deferred, filed as #478

The three you marked as follow-ups, together, because they share a subject — the shared numbering space has two writers and nothing yet reconciles it:

1. The reconciliation pass: slots that drift while a Core is off the rail, and a slot stranded by an unpin this Panel never saw. The non-atomicity above is folded in, since one endpoint owning both tables would close both.
2. `project-presentation` routes in `api-auth.test.ts`'s `PROTECTED_ROUTES` — the gate is default-deny so they are covered today; what is missing is the test that keeps them that way.
3. `listProjectPresentation` degrading to empty rather than to "unknown". Worth more now than under #98: an empty answer is indistinguishable from "no filing", so a failed read sends every Core pin to the end of the rail until the next good one.

## Verification

`pnpm typecheck` and `pnpm lint` clean, full `pnpm test` green. Coverage added: 6 engine cases, 5 new component cases (12 in that file now), 1 service case for the rail-end slot.

Still a draft, base unchanged, nothing merged, no force push.

mrfqcentic (changes requested):
# Review round 2 — PR #476 against #382

**Verdict: request changes.** Both round-1 blockers are genuinely closed — I re-drove each trace at `1fa46fe` and neither reproduces — every non-blocking item was taken or filed as #478, and the two written acceptance criteria hold end to end. Two things block:

1. the new unpin cleanup at `ProjectBar.tsx:1589` fires for **Panel-owned** rows too, writing a `project_presentation` row under a Core the project does not belong to — and the next prune of that Core deletes the Panel project's card-image file off disk;
2. `settleCorePinFiling` does not take the overlay down "behind a read" the way its own doc-comment and the PR body both promise, because `createCoalescingRunner`'s promise resolves without reading when a pass is already in flight.

Everything below was checked by reading the enforcing line at `1fa46fe`, not from the PR body.

Base check: `origin/fix/operator-pins-sessions-drop-cli` is `d727928` and is an ancestor of this head, so #471, #462 and #472 are all in. GitHub reports `MERGEABLE / CLEAN`; all 20 non-skipped checks pass, `Unit Tests` and `Typecheck` included.

---

## Round-1 blockers — both closed

### B1 — the moved Core tile no longer waits on a fan-out ✔

`lib/core-pins-engine.ts:279-295` is the optimistic setter round 1 asked for, and it is wired where it has to be: `ProjectBar.tsx:425` calls `applyCorePinFiling(corePinFiling)` **before** the query-cache write at `:426` and before the first `await` at `:447`.

Re-driving the round-1 trace, rail `[P1(0), C1(null)]`, moving `C1` to the top:

- **Drag.** `:646-665` — after `GROUP_SETTLE_MS` the timeout runs `setProjectDrag(null)` and then `persistProjectOrder`, which synchronously calls `applyCorePinFiling({C1:0, …})`. `publish` (`:284-294`) notifies the `useSyncExternalStore` subscription in `use-fleet.ts:403-409`, so `projects` (`ProjectBar.tsx:112-115`) → `sortedPinned` (`:144`) recompute in the **same** React batch as `setProjectDrag(null)`. `C1` now carries `pinnedOrder: 0` instead of `null`, so `comparePinnedProjects` (`pinned-project-order.ts:8-13`) no longer sorts it to `MAX_SAFE_INTEGER`. The tile is drawn where it was dropped with no Core read — `core-pins-rail-slot.test.tsx:158-179` asserts exactly that, with `presentationReads` unchanged.
- **Shift+Arrow.** `:680-691` funnels into the same `persistProjectOrder`, so the overlay lands identically with no settle to hide anything. `project-bar-mixed-rail-reorder.test.tsx:462-481` drives the real keydown and asserts both `writeLog[0] === "applyCorePinFiling"` and the *rendered* order — `railIds(view.container)[0]` is `CORE_ONE.id` before any write returns.
- The follow-up-gesture half goes with it: `pinnedIdsRef.current` (`:311-315`) is recomputed from `railClusters` ← `sortedPinned`, which is now the optimistic order.

The suite that missed this in round 1 was rewritten so it *could* catch it — the engine mock is a real `useSyncExternalStore` store (`project-bar-mixed-rail-reorder.test.tsx:95-124`), not a value.

### B2 — a reorder written while a Core is disconnected no longer reverts permanently ✔

`core-pins-engine.ts:139-141`:

```ts
const remembered = () =>
  (lastRowsByCore.get(core.id) ?? []).map((row) => withFiling(row, presentation.get(row.id)));
if (!bridge || core.dial.state !== "connected") return remembered();
```

`presentation` is the map this pass just read (`:190-195`), and `withFiling` (`:108-120`) re-derives the filing from it rather than keeping what was baked into the remembered row. The permanent revert is gone: the write lands in the Panel's own database, the next pass joins it onto the remembered Core facts, and the tiles stay put. `core-pins-rail-slot.test.tsx:247-277` drives it for real — dial to `unreachable`, mutate the presentation table, refresh, assert the new slot and group while `name` still comes from the remembered row.

---

## Blocking

### 1. `components/views/ProjectBar.tsx:1589` — unpinning a *Panel-owned* project from a Core-scoped rail files it under that Core, and the next prune deletes its card image off disk

`:1573` is `const targetCoreId = target?.coreId ?? coreId;`. That fallback is pre-existing, but until this PR nothing destructive hung off it. A Panel-owned row has no `coreId` (`shared/projects.ts:23` — it is set only in `core-pins-engine.ts:168`), so on a Core-scoped route — `__root.tsx:651` passes `routeCoreId`, i.e. any time the operator is looking at a project or session on a Core — `targetCoreId` becomes **that Core's id for the operator's own local project**.

The guard at `:1589` is `if (unpinning && targetCoreId)`, so it passes. And the `mutateProjectForCore` call above it does not stop it: `pinProject` in `shared/core-mutations.ts:300-311` returns `null` when `changes === 0`, the Core answers `{ project: null }` (`pty-core-link-server.ts:1760-1774`), and `mutateProject` resolves that `null` rather than throwing. So `:1590-1592` runs:

```ts
await api.updateProjectPresentation(id, targetCoreId, { pinnedOrder: null })
```

`upsertProjectPresentation` (`services/project-presentation.ts:58-83`) creates a row `{ projectId: <a Panel project id>, coreId: <a Core that has never heard of it> }`. Then, the next time any project on that Core is opened, `queries/index.ts:84-86` fires `pruneProjectPresentation(coreId, coreProjectIds)`; the Panel id is not in the Core's list, so `findProjectPresentationOrphans` returns it, `deleteProjectPresentation` (`:118-122`) removes it **and calls `deleteAllProjectImagesFor(projectId)`** — which sweeps `project-images/<projectId>.*` by basename (`project-image-files.ts:34-45`), the same directory and the same naming the Panel's own card images use.

Net: right-click one of your own pinned projects → "Unpin project" while on a Core route, and its card image is deleted from disk a moment later, while `projects.image_path` still points at it. The pin itself also silently does not clear, but that half is pre-existing; the file deletion is new here.

Guard on the row's real owner rather than on the fallback — `coreIdByProject.get(id)`, or `target?.coreId`, is the value this block actually wants. (It is worth deciding separately whether `:1573`'s fallback should exist at all now that `coreIdByProject` is right there.)

### 2. `lib/core-pins-engine.ts:305-309` — the overlay does not come down "behind a read"

```ts
export async function settleCorePinFiling(projectIds: readonly string[]): Promise<void> {
  if (projectIds.length === 0) return;
  await run();
  for (const projectId of projectIds) pendingFiling.delete(projectId);
}
```

`run` is `createCoalescingRunner(...)`, and `fleet-refresh.ts:66-70` is:

```ts
return async () => {
  if (running) { pending = true; return; }
  …
```

When a pass is already in flight the returned promise resolves on the next microtask with **no read having happened**. The doc-comment at `:297-304` ("The read runs first and publishes rows that already carry the confirmed filing… On a failed write the same read is what puts the truth back") and the PR body's "the overlay outranks the server's answer until `settleCorePinFiling` takes it down behind a read" both describe a guarantee the code does not provide. Two consequences:

**a. The happy path reopens B1's symptom whenever a pass overlaps the write.** Passes are started by the 15s poll, by `onConnectionChange`, and by every `task:`/`session:`/`pty:` and project-list event (`:210-217`) — continuous on a fleet with a live session. Trace, rail `[P1, C1]`, Shift+Arrow moving `C1` up:

1. `ProjectBar.tsx:425` overlays `C1 → 0`; the tile moves. ✔
2. A task event fires `run()`. The pass reads `listProjectPresentation` (`:190`) — pre-write — and starts the per-Core `listProjects`/`listTasks` fan-out.
3. Both writes return. `ProjectBar.tsx:491` `void settleCorePinFiling([C1, …])` → `run()` sees `running === true`, sets `pending`, returns → `pendingFiling.delete("C1")` runs immediately.
4. The in-flight pass reaches `readCore` → `withFiling` (`:118`): the overlay is gone, so `filed?.pinnedOrder` — the **pre-write** slot — wins. `sameSnapshot` sees a difference and publishes. **The tile springs back to where it was before the gesture**, for the rest of that fan-out.
5. `pending` was set at step 3, so the `do…while` loops once more and puts it back — unless that pass returned `false`, in which case the loop `break`s and nothing corrects the rail until the next poll tick, up to 15s.

That is the same visible revert B1 was about, now gated on a race an active fleet loses regularly rather than on every gesture. Neither new test covers it: `core-pins-rail-slot.test.tsx:181-199` starts *and finishes* a pass while the overlay is up, and `:201-231` awaits `settleCorePinFiling` with nothing in flight. A case that calls `settleCorePinFiling` mid-pass is what would have caught it.

**b. The failure path leaves the rail in an order nobody chose.** `settleCorePinFiling` deletes the overlay but never republishes, so on `ProjectBar.tsx:475-480`:

- the catch restores `previous` and `invalidateProjects()` refetches — the **Panel** rows go back to the server's old order;
- `finally` → `settleCorePinFiling` runs its pass *with the overlay still up*, so it publishes the optimistic values (or nothing, via `sameSnapshot`); the entries are then deleted with no further publish;
- so the engine snapshot keeps the **new** slots for every Core row.

The operator gets "Could not move project" over a rail that is the Panel half reverted interleaved with the Core half not reverted — neither the old order nor the one they asked for — until the next engine pass, up to `FLEET_POLL_MS` on a quiet fleet. `reorderPinnedProjects` refusing the move is a case this PR reasons about explicitly (`ProjectBar.tsx:441-446`), so it is not hypothetical.

Both fall out of the same line. A runner whose promise resolves only when a pass that started *after* the call has completed closes them together; alternatively, take the overlay down and then publish (`applyCorePinFiling`-style) rather than relying on a read that may not have happened.

---

## Non-blocking

- **`ProjectBar.tsx:481-492` — `settleCorePinFiling` is the one thing in `finally` not guarded by `saveSeq`.** The gesture guard is released at `:483` and the settle fired at `:491`, deliberately, so the next Shift+Arrow is not swallowed. But the settle deletes `pendingFiling` entries by key with no notion of which gesture owns them: two quick presses and gesture 1's `await run()` resolves after gesture 2 has written its overlay, deleting it while gesture 2's write is still in flight. A per-write token would close it. Third trigger of the same teardown.
- **`core-pins-engine.ts:139` — B2's fix widens the blast radius of a failed presentation read.** `run()` degrades `listProjectPresentation` with `.catch(() => [])`, so on a failed read `withFiling(row, undefined)` assigns `imagePath`, `launchUrl`, `groupId` and `pinnedOrder` all `null`. Before this PR a disconnected Core's remembered rows kept their last-known filing; now one API blip drops every remembered pin out of its cluster and to the end of the rail with no toast. #478's item 3 is the right home for it — worth noting only that the disconnected-Core case is newly affected.
- **`ProjectBar.tsx:491` — `void` on an async call swallows rejections.** `pass` cannot currently throw (everything in `core-pins-engine.ts:185-201` is inside the try), but if it ever did, `pendingFiling` would never clear and the engine would ignore the server's filing for those ids for the life of the tab.
- **`repositories/project-presentation.repo.ts:41-48` — `findMaxProjectPresentationPinnedOrder` maxes over every presentation row**, stranded slots included, so `nextRailSlot` drifts upward over time. Harmless (it only pushes a new pin further past the end) and squarely #478's reconciliation pass.

---

## What I re-checked and it holds

- **AC1 — order survives a reload on a mixed rail.** One numbering space end to end: `ProjectBar.tsx:391-394` numbers Core rows by their index in `fullOrder`; `services/projects.ts:388-408` numbers the Panel's own rows from the same array and skips ids it holds no row for (`:400-405`); `project-presentation.ts:104-110` writes the Core half in one `.transaction(…).immediate()`; the read leg joins the slot back on at `core-pins-engine.ts:118` and `getPinnedProjects` sorts both owners in one space. Covered at every layer.
- **AC2 — no error toast on a valid mixed-rail reorder.** Both rejections are gone: `validatePinnedReorder` (`pinned-project-order.ts:76-93`) now treats an id this Panel has no row for as a passenger while still demanding every pinned row of ours exactly once, still rejecting duplicates, and still rejecting an *unpinned* row of ours when handed `panelProjectIds` (`projects.ts:394`); the regroup half sends a Core row's group to `updateProjectPresentation` (`ProjectBar.tsx:456-459`) rather than the 404-ing Panel PATCH. `toastError` is asserted unused on every valid case.
- **Write ordering fixed as asked** — `reorderPinnedProjects` first, its answer applied before the group write (`:447-453`), pinned by "commits the validating order write before the group write".
- **Atomicity wording narrowed** in all three places: PR body, ADR 0022's amendment, and `project-presentation.ts:96-103`.
- **`nextRailSlot`** (`services/projects.ts:33-42`) maxes over both halves via `Math.max(panelNext, coreMax + 1)`, nullish-coalescing throughout so a `0` slot is not swallowed, and is used on all three paths (`:296`, `:345`, `:363`) with tests on create-pinned and toggle.
- **The rail's own unpin clears the slot** — modulo blocking #1, which is about the guard, not the intent. The unwitnessed unpin is honestly left to #478.
- **Stale prose fixed** at `lib/api.ts:261-266` and `services/project-presentation.ts:13-24`; `CONTEXT.md` and the ADR amendment both say four fields and both match the code.
- **The regroup half now has a real test and the vacuous assertion is no longer load-bearing.** `project-bar-mixed-rail-reorder.test.tsx:607-677` drives an actual pointer drag across a hand-laid-out rail: `updateProject` **not** called *while* `updateProjectPresentation` is called once with `[CORE_ONE.id, "core_alpha", { groupId: GROUP_B.id }]` — so the not-called assertion now discriminates. Write ordering and the optimistic cluster move are pinned alongside it.
- **Deferred items were filed, not dropped** — #478 carries all three with acceptance criteria of its own, plus the cross-table non-atomicity.
- **Route, auth, upsert merge and off-group (#378) handling.** `api-router.ts:285-289` matches the literal path before `PROJECT_PRESENTATION_PATH`; `/api/project-presentation/*` is absent from `ANONYMOUS_ROUTES` (`:99-103`) under a default-deny gate; `definedFields` means the `reorderCorePins` write and the following `groupId` write do not clobber each other; the off-group tile is still excluded at `:311-315` with two tests on it.

---

## Checklist

| | |
|---|---|
| AC1 — drag two pins incl. a Core pin, order survives reload | **Met** |
| AC2 — no error toast on a valid mixed-rail reorder | **Met** |
| B1 — moved Core tile stays put without a fan-out, **drag** | **Met** |
| B1 — same, **Shift+Arrow** | **Met** |
| B2 — reorder with a Core disconnected does not revert permanently | **Met** |
| Optimistic setter cannot leave the snapshot disagreeing with the DB | **Not met** — blocking 2 |
| Core pin slot cleared on unpin | **Partially met** — clears it, but also files Panel rows under a foreign Core (blocking 1) |
| Regroup half has a real test | **Met** |
| The previously vacuous `updateProject` assertion now discriminates | **Met** |
| Group write ordered after the validating order write | **Met** |
| Atomicity claim narrowed everywhere | **Met** |
| `nextPinnedOrder` numbers over the whole rail | **Met** |
| Stale four-fields prose | **Met** |
| Deferred items filed as an issue | **Met** (#478) |
| Branch still merges cleanly over #471/#462/#472 | **Met** |
| CI green at `1fa46fe` | **Met** |

Two fixes, both small and local, and the rest of this is ready.

qcentic-dev:
Both blockers fixed, all four non-blocking items taken. Three commits on top of `1fa46fe`: `c8cdbca`, `6c51098`, `f6fc9c4`.

---

## Blocking 1 — `ProjectBar.tsx:1589`, the unpin that filed a Panel row under a foreign Core

Fixed as suggested, and the trace is confirmed as you described it. `c8cdbca`.

```ts
const targetCoreId = coreIdByProject.get(id) ?? null;
```

**The fallback is gone, not merely bypassed** — you asked for that decision separately, so here it is with the reasoning.

`coreIdByProject` is built from `projects`, and `menuProject` comes from `visibleById` ← `railClusters` ← `sortedPinned` ← the same `projects`. So for every row a menu can open on, `coreIdByProject.get(id)` is exactly `target?.coreId ?? null`; and `remotePinned` covers the whole fleet, not just the route's Core, so a Core-owned row always carries its own `coreId` however the bar is scoped. The fallback therefore never once produced a *correct* answer that the row could not: it fired only for a Panel-owned row, where it was wrong, or for a row not in the list at all, where it was a guess. A guess is what did the damage, and there is nothing left for it to be right about. The bar's `coreId` prop still has a legitimate use one screen over — `initialCoreId` on the edit dialog, where a default for a *new* association is the actual question — and that is untouched.

The `if (unpinning && targetCoreId)` guard is now true only for genuinely Core-owned rows, so no `project_presentation` row can be created for a project its `coreId` has never heard of, and `pruneProjectPresentation` → `deleteAllProjectImagesFor` has nothing of the Panel's to sweep.

Two tests, driven through the real context menu (portalled to `document.body`) on a rail mounted with `coreId="core_alpha"`:

- a Panel-owned row → `mutateProjectForCore` called with `null`, `updateProjectPresentation` **not called at all**;
- a Core-owned row → `mutateProjectForCore` called with `"core_beta"` (the row's Core, not the rail's), and `updateProjectPresentation` called once with `[id, "core_beta", { pinnedOrder: null }]`.

The first fails against `1fa46fe`.

The pre-existing half you noted — the pin silently not clearing when the Core never had the row — is unchanged and still #478's.

## Blocking 2 — `core-pins-engine.ts:305-309`, the overlay not coming down behind a read

Fixed at the line you pointed at, and both consequences close together. `6c51098`.

**`createCoalescingRunner` grows `fresh()`.** Waiters are claimed at the top of a loop iteration and released when *that* iteration's pass settles, so a waiter is only ever answered by a read taken after it joined; `waiting.length > 0` is an arm of the loop condition, so a pass in flight brings the loop round again for it. A loop that `break`s on a failed pass releases its stranded waiters in the `finally` rather than hanging — holding the operator until an error clears is the worse bargain, and the republish below is correct on that path anyway. The plain `run()` is unchanged, so no other caller's behaviour moves.

**The settle also republishes.** `fresh()` alone does not close consequence (b), and I want to be explicit about that since the review offered the two fixes as alternatives: the fresh pass runs with the overlay still up, so it publishes the optimistic slots, and deleting the entries afterwards still changes no row. So the settle now takes the overlay down and then re-files the settled rows from `lastPresentation` — the map that pass actually read. On the happy path those are the same values and `sameSnapshot` makes it a no-op; on the failure path it is what brings the Core half back to the order the Panel half has just reverted to. Both consequences, one function.

Four new tests in `core-pins-rail-slot.test.tsx`, all four red against `1fa46fe`:

- **the mid-pass case the suite was missing.** The presentation read is held open, so a pass is genuinely in flight and carrying pre-write rows; the write then lands and `settleCorePinFiling` is called. It asserts the settle has **not** resolved while that pass is still out (`expected true to be false` against the old code — the microtask resolution, exactly), then releases the read and asserts the tile never moved.
- the failure path: overlay up, table unchanged, settle → both halves back on the server's order.
- two overlapping gestures: gesture 1's settle leaves gesture 2's overlay standing, a read still cannot repaint through it, and gesture 2's own settle is what takes it down.
- a failed presentation read moves no tile, connected and disconnected.

## Non-blocking, all four taken

1. **`saveSeq` does not guard the settle.** `applyCorePinFiling` returns a token; the settle removes only entries still carrying it. The guard stays released before the settle — that was deliberate and stays — and the token is what closes the third trigger of this teardown. The rail's half of the contract is asserted too: it must settle with the token its *own* setter returned.
2. **The widened blast radius of a failed presentation read.** Taken here rather than deferred, because you are right that the disconnected-Core case is newly affected by this PR's own fix and it would be odd to ship the widening and file the narrowing. `run()` holds the last presentation map on a failed read instead of substituting `[]`. #478's acceptance criterion 4 is met; I have commented there saying so, and items 1 and 2 stay open.
3. **`void` swallowing rejections.** `.catch(() => undefined)` at the call site, and `run.fresh()` catches the runner's rejection for the same reason — a throwing pass releases the waiter in the loop's `finally`, so nobody was left holding that one either.
4. **`findMaxProjectPresentationPinnedOrder` maxing over stranded slots.** Left as behaviour and documented as such: this table cannot tell a stranded slot from a live one, so the answer needs a view of both halves of the rail at once, which is #478 item 1. The comment now says that, and I have added the note to #478 so it is not rediscovered.

## Housekeeping

`origin/fix/operator-pins-sessions-drop-cli` is still `d727928` and still an ancestor of this head — no merge was needed, nothing rebased, nothing force-pushed. `pnpm typecheck` clean; `pnpm test` green locally across all six stages (root 23, sdk 22, shared 31, cli 50, core 60, panel 164 files; panel 1766 tests). PR body updated: the "behind a read" sentence now describes what the code does, the deferred-items bullet reflects #478 item 3 moving here, and the test counts are 10 and 14.

**CI on `f6fc9c4`: all 18 non-skipped checks pass.** One thing worth flagging rather than burying: `Unit Tests` was red on the first attempt, on `packages/cli`'s `in-process-core-session.test.ts > sends and awaits the turn in one PTY resolution` — a 60003ms timeout, not an assertion. This branch's diff is `packages/panel/src` only (six files, no `packages/cli` at all), that file passes locally 10/10, and it passed on the re-run of the same job at the same SHA. I am calling it a flake in that PTY test rather than anything this branch did, but it is a 60s hard timeout in a test that normally finishes well inside it, so it is worth someone's attention independently of this PR.

Still draft, base unchanged, and `projects.$id.tsx` untouched — that is #401's.

mrfqcentic (approved):
Approved on the operator's direct instruction, superseding the round-2 CHANGES_REQUESTED.

This is an operator override, not a review verdict. The round-2 fixes at head f6fc9c4 were NOT verified by a reviewer:

- the unpin path that was writing a Panel project id against a foreign Core and costing that project its card image on the next prune;
- the settleCorePinFiling overlay that came down without a read having happened.

CI is green on the head and the branch is mergeable. Round 3 was attempted twice and died on repeated upstream 529 Overloaded errors. Deferred items from rounds 1-2 are tracked in #478.

PR #472, opened 2026-09-02
fix(cli): pairing labels the Core host:port Address, not Endpoint

qcentic-dev:
Closes #414.

An **operator-facing copy change only.** The pairing handout labelled the Core's host:port `Endpoint`; the Panel's own form asks for an *address*. Operators copied the wrong word across. Every operator-visible label for that value now reads **Address**, and the `--public-host` pipe line reads **`Address host`**.

## What changed

| Surface | Before | After |
| --- | --- | --- |
| `actana pair new`, framed TTY box | `Endpoint       core:8443` | `Address        core:8443` |
| `actana pair new`, wrapped heading (long address) | `Endpoint` | `Address` |
| `actana pair new`, piped stdout (`--public-host` only) | `Endpoint host  10.0.0.5` | `Address host   10.0.0.5` |
| `core pair` success frame (`core-pair-results.ts`) | `field("Endpoint", …)` | `field("Address", …)` |
| `actana core status` blob display (`core-command.ts`) | `["Endpoint", blob.endpoint]` | `["Address", blob.endpoint]` |
| `actana status` (`actana-status.ts`) | `row("Endpoint", …)` | `row("Address", …)` |

Plus the tests in `actana-pair.test.ts` that assert those strings, and the comments in `packages/panel/src/shared/core-onboarding.ts` (and its own test file) that quote the CLI's `Endpoint host` line.

**Column widths are unchanged.** The framed label still pads to 15 (`"Address        "` replaces `"Endpoint       "`), and the pipe line's `"Address host   "` occupies the same 15 columns `"Endpoint host  "` did, so the piped shape keeps its order, its conditional and its alignment — only the word moved. `field()` and `row()` both `padEnd`, so the shorter label costs nothing there either.

## `ApiSettingsPage.tsx` — **not** renamed, and why

The instruction was to rename that `Field label="Endpoint"` only on proof that it is a Core address rather than an HTTP API URL. It is **an HTTP API URL**, so it keeps the name `Endpoint`:

- Its value is `const baseUrl = "http://127.0.0.1:<core-port>"` — an `http://` origin, not the `host:port` a Panel pairs against.
- The section it sits in is titled **External API**, and its subtitle says external CLIs "post status updates to the Core."
- The sibling field on that page is the **API token**, and the third field is a `curl` example that posts to `${baseUrl}/api/tasks/$TASK_ID/status`.
- The file's own comment ties it to ADR 0010's token, not to pairing.

[...]

PR #477, opened 2026-09-03
fix(panel): a drop over a grid session is not a project-root upload

### LINKED ISSUE #401: B7: in grid view a file drop aimed at a session is intercepted by the project board

qcentic-dev:
Closes #401 (B7).

## The landmine

In grid view the whole session grid renders **inside** the board `div`, so the board's own `onDrop` wraps it. A file
dropped on a session's terminal bubbled to the board handler, which has no idea what the operator was aiming at and
uploaded to the **project root** — the file landed somewhere nobody pointed at.

## The shape chosen: a target test on the board side, not `e.defaultPrevented`

The task asked whether honouring `e.defaultPrevented` on the board handler — the alternative shape #400 named for this
file — is right here. **It is not**, for two independent reasons:

1. **Nothing in a session pane ever sets it for this gesture.** `wireTerminalFileDrop`
   (`packages/panel/src/lib/terminal-pane-helpers.ts`) `preventDefault`s only for a *project-path* drag from the
   Panel's own rail. Desktop file drags are deliberately ignored there: the browser hands over bytes, not a path on
   the Core's machine (ADR 0010). So on a file drop over a terminal `e.defaultPrevented` is `false`, and a board
   guard reading it would be a no-op — the bug would ship unchanged.
2. **React would blunt it even if something did set it.** `SyntheticEvent` copies `defaultPrevented` from the native
   event *at construction*, so a React descendant calling `preventDefault` during the same root dispatch would not
   show up on the board's event object. Only a native listener firing before React's root listener would, which is
   exactly the case (1) rules out.

`stopPropagation` on the session side — #400's other shape — is also wrong here: it is the drawer's fix (already
merged, `ProjectFilesPanel.tsx`, a different handler and untouched by this PR), and it needs a descendant handler that
*wants* the drop. A session pane does not want a file drop at all; it needs the board to stand down, not to hand the
gesture over.

So the arbiter is **where the pointer was**. New `packages/panel/src/lib/board-drop-arbiter.ts` exports
`dragTargetIsSessionPane(event)`, which walks up from the drag's target to `[data-grid-cell], [data-session-grid]`.

## The discriminator: both markers, and why the grid container is not redundant

`[data-grid-cell]` alone is defeated exactly where the operator cannot see it, and the first round of this PR shipped
that hole ([review finding 3.3a](https://github.com/actana/control/pull/477#discussion_r3923999452)):

`SessionGrid` renders the row and column **resize handles** as absolutely-positioned children of the grid container —
siblings of the row `div`s, inside no cell — at `zIndex: 6`, above every non-expanded cell (`SessionGrid.tsx:2433-2476`;
the only higher value, `8`, is the expanded cell, where `resizeEnabled` is false anyway, `:1774`). Their hit area is
`HANDLE_HIT = GRID_GAP = 8` (`:53`, `:63`) while the shipped Studio layout runs `gridGap = 0; gridPad = 0` (`:836-837`).
So each handle is an **8px strip centred on a 0-width seam**: it lies on top of terminal pixels, 4px into each
neighbour, full pane height for a column divider and full grid width for a row divider. `closest("[data-grid-cell]")`
answers `null` there — so the board painted itself hot, showed the `copy` cursor and uploaded to the project root, 4px
from a pane edge, with feedback that actively invited the drop. AC1's exact failure mode.

`[data-session-grid]` (already on the grid container, `SessionGrid.tsx:2373`) closes it by making the **whole grid
element** session space. Nothing else moves: the hidden-sessions bar renders as a sibling *after* the grid (`:2531`),
and the project header strip and the `CardFrame`'s 8px padding sit outside it. All three still upload to the project
root.

Correcting the justification while I was here: the earlier version of this PR (and of the module comment) claimed "the
gaps around and between the cells" as the board's own space inside the grid. **With `gridGap` and `gridPad` both 0 that
space does not exist**, and the only thing that looks like it — the divider strip — is the one place the operator cannot
tell they are not on a pane. The board's real own space in grid view is entirely *outside* the grid element.

`SessionGrid.tsx:55-57` also carried a stale claim — that the invisible handles "never overhang the terminal surfaces".
True when the gap was 8, false since the flush layout, and the accurate account was already sitting at `:827`. That
comment is now corrected in place, and points at this arbiter.

## What the board does now

- `onDragOver` inside the grid: still `preventDefault` (leaving the browser default alone would make the drop
  **navigate the Panel to the dropped file**), but `dropEffect: "none"` and the hot dashed outline stays off. The
  operator sees a no-drop cursor over a session, and a hot board only where the board really will take the file.
- `onDrop` inside the grid: `preventDefault` first (same reason), then return. No upload, no Files drawer, no pending
  drop.
- Everywhere else on the board — the project header strip, the frame padding, the hidden-sessions bar, the empty-grid
  state and all of single-session view — uploads to the project root exactly as before.

Single-session view needed no change: `TerminalPanel` renders as a *sibling* of the board in `__root.tsx`, never a
descendant, so its drops never reached this handler. That is stated in a comment next to the selector so the next
reader does not go looking for a missing case.

## Recorded, not fixed

- **A portalled menu is still board space** ([3.3c](https://github.com/actana/control/pull/477#issuecomment-5524738131)).
  `TerminalPane.tsx:299` portals the session-actions menu to `document.body` (`:392`). React propagates synthetic
  events along the *React* tree, so a drop on the open menu reaches the board's `onDrop` while its DOM sits outside the
  grid and `closest` answers `null` — it uploads to the project root. It takes a menu held open across a drag, and a
  menu is not a terminal. A DOM-position arbiter cannot see a React-tree ancestor; closing it would need a second
  signal. Commented in `board-drop-arbiter.ts`.
- **`dropEffect` is written unconditionally on the bubble** (reviewer's note on `projects.$id.tsx:1860`). Benign today
  — nothing inside a cell handles a file drag — but a future per-session drop that sets `copy` would have it stomped.
  The repo's convention is the inner target claiming the gesture with `stopPropagation` (`ProjectFilesPanel.tsx`,
  #400/#227). Commented on the handler.
- **With a cell expanded, the refusal covers everything under the header**
  ([3.5](https://github.com/actana/control/pull/477#discussion_r3923999477)) — a feedback gap, not a wrong destination.
  Filed as **#479**, referencing #401 and this PR, rather than carried here.

## One red check that was not this PR's, fixed rather than re-run

`Unit Tests` went red on the head of this branch in a package it does not touch: `packages/shared`,
`core-cert-material` > "gives every issuance its own serial", `expected 240 to be less than 128`. A 1-in-256 draw in
`randomSerial`, not a regression from #401 — the earlier run at `361b6d9` was green with the same code.

`randomSerial` drew 16 bytes and cleared the top bit of the first, which makes the DER INTEGER positive (RFC 5280
§4.1.2.2). Two of the 256 possible first bytes — `0x00` and `0x80` — clear to `0x00`, and the mask is not enough there:
DER writes an INTEGER minimally and `@peculiar/x509` returns `serialNumber` with the leading zero byte **stripped**, so
that issuance reported a 15-byte serial starting at whatever byte sat behind the zero — with its own top bit set half
the time. Confirmed against the library directly: a certificate minted with `00f0aabb…` reports `f0aabb…`, one minted
with `7f0011…` reports it unchanged.

The serial names the certificate in every later revocation list and `pair ls` row, so it has to be the same string on
both sides of the issuance. The leading byte is now forced into `0x01..0x7f` — top bit clear for positive, non-zero so
the encoder has nothing to strip. The test drew the property once per issuance, which is how a one-in-256 bug outlived
every run of it; it now asserts the full 16 bytes came back and draws `randomSerial` 5000 times to pin the bounds
directly. The new test fails against the old body.

Separate commit (`ccda088`), and easy to lift out of this PR if you would rather it rode on its own.

## Scope

- Touched: `packages/panel/src/routes/projects.$id.tsx`, `packages/panel/src/components/views/SessionGrid.tsx`
  (comment only), plus `lib/board-drop-arbiter.ts` and its test. Outside #401 and in its own commit:
  `packages/shared/src/core-cert-material.ts` and its test, for the red check above.
- Not touched: `ProjectFilesPanel.tsx` (#400, already merged — a different handler), `ProjectBar.tsx`, `use-fleet.ts`,
  `shared/projects.ts`, `queries/index.ts`, `visible-project-scope.ts` (#382 / #381).
- Base `fix/operator-pins-sessions-drop-cli` merged in with a plain merge commit (`3fc380c`) after #471, #462 and #472
  landed. No rebase, no force push.

## Verification

- The test suite is rebuilt on **real DOM under `jsdom`**, with markup shaped like `SessionGrid`'s: board → frame →
  header → `[data-session-grid]` → `[data-grid-row]` → `[data-grid-cell]` → `.xterm`, a divider as a sibling of the
  rows inside the grid, and the hidden-sessions bar after it. The old stub answered `closest` against the same literal
  the module holds, so it agreed with any selector — a tautology, and the reason the divider went unnoticed
  ([3.4](https://github.com/actana/control/pull/477#discussion_r3923999471)). 9 tests; **the divider-seam case fails
  against the old `[data-grid-cell]`-only selector**, which is the point of it.
- `pnpm typecheck` — clean.
- `eslint` on the four touched files — **0 errors**; the warnings reported are pre-existing unused-vars. Full
  `pnpm lint` OOMs (exit 137) on this box, which is a local memory ceiling, not a finding; CI runs it.
- Targeted suites green: `board-drop-arbiter` (9), `project-files-panel` (19), `session-grid-layout` +
  `session-grid-tracks` (18).

Draft on purpose, based on `fix/operator-pins-sessions-drop-cli` (not `main`, not the train).

https://claude.ai/code/session_01FBR3m1VZWiuWVSoRWUEqGU

### DIFF
--- a/packages/panel/src/lib/board-drop-arbiter.test.ts
+++ b/packages/panel/src/lib/board-drop-arbiter.test.ts
@@ -1,25 +1,112 @@
+// @vitest-environment jsdom
+//
+// The arbiter that decides whether a file drop on a Project board belongs to a
+// session or to the board (#401).
+//
+// Real DOM, real `closest`, markup shaped like `SessionGrid`'s: a stub that
+// answers `closest` against the same string the module holds would agree with
+// any selector, and that tautology is exactly what let the divider seam through
+// the first time. The divider case below is the one that decides this file.
 import { describe, expect, it } from "vitest";
 import { dragTargetIsSessionPane } from "./board-drop-arbiter";
 
-/** Stands in for the element a drag event reports as its target. `matches` is
- *  the set of selectors this element (or one of its ancestors) answers to. */
-function targetMatching(...matches: string[]): EventTarget {
-  return {
-    closest: (selector: string) => (matches.includes(selector) ? { tag: selector } : null),
-  } as unknown as EventTarget;
+/**
+ * The board as it renders in grid view — mirroring `projects.$id.tsx` and
+ * `SessionGrid.tsx`:
+ *
+ *   board                      the drop target that uploads to the Project root
+ *     frame                    CardFrame, 8px padding — board space
+ *       header                 the project header strip — board space
+ *       grid                   [data-session-grid], position: relative
+ *         row                  [data-grid-row]
+ *           cell / cell        [data-grid-cell], each holding a terminal
+ *         divider              absolutely positioned child of the *grid*, a
+ *                              sibling of the rows and inside no cell, lying
+ *                              across the seam between the two cells
+ *       hiddenBar              sibling *after* the grid — board space
+ */
+function renderBoard() {
+  const make = (tag: string, attrs: Record<string, string> = {}) => {
+    const el = document.createElement(tag);
+    for (const [k, v] of Object.entries(attrs)) el.setAttribute(k, v);
+    return el;
+  };
+
+  const board = make("div");
+  const frame = make("div");
+  const header = make("div", { class: "mc-project-header" });
+  const grid = make("div", { "data-session-grid": "" });
+  const row = make("div", { "data-grid-row": "0" });
+  const cellA = make("div", { "data-grid-cell": "", "data-task-id": "task-a" });
+  const cellB = make("div", { "data-grid-cell": "", "data-task-id": "task-b" });
+  const terminalA = make("div", { class: "xterm" });
+  const terminalCanvasA = make("canvas");
+  const divider = make("div", { style: "position:absolute;z-index:6" });
+  const hiddenBar = make("div", { "data-hidden-sessions-bar": "" });
+
+  terminalA.append(terminalCanvasA);
+  cellA.append(terminalA);
+  row.append(cellA, cellB);
+  grid.append(row, divider);
+  frame.append(header, grid, hiddenBar);
+  board.append(frame);
+  document.body.append(board);
+
+  return { board, frame, header, grid, row, cellA, cellB, terminalCanvasA, divider, hiddenBar };
 }
 
 describe("dragTargetIsSessionPane", () => {
-  it("claims a drag whose target sits inside a grid cell", () => {
-    expect(dragTargetIsSessionPane({ target: targetMatching("[data-grid-cell]") })).toBe(true);
+  it("claims a drop on the terminal inside a cell", () => {
+    const { terminalCanvasA } = renderBoard();
+    expect(dragTargetIsSessionPane({ target: terminalCanvasA })).toBe(true);
   });
 
-  it("leaves the board's own space to the board", () => {
-    expect(dragTargetIsSessionPane({ target: targetMatching() })).toBe(false);
+  it("claims a drop on a cell itself", () => {
+    const { cellB } = renderBoard();
+    expect(dragTargetIsSessionPane({ target: cellB })).toBe(true);
+  });
+
+  it("claims a drop on a resize divider lying across the seam between two panes", () => {
+    // The regression this file exists for. The divider is a child of the grid
+    // and inside no cell, but it is an 8px strip painted over 4px of each
+    // neighbouring terminal (HANDLE_HIT = 8 with gridGap = 0), so a drop on it
+    // is a drop the operator aimed at a session. `[data-grid-cell]` alone
+    // answers null here and the board would have uploaded to the Project root.
+    const { divider } = renderBoard();
+    expect(divider.closest("[data-grid-cell]")).toBeNull();
+    expect(dragTargetIsSessionPane({ target: divider })).toBe(true);
+  });
+
+  it("claims a drop on the grid container itself", () => {
+    const { grid } = renderBoard();
+    expect(dragTargetIsSessionPane({ target: grid })).toBe(true);
+  });
+
+  it("leaves the project header strip to the board", () => {
+    const { header } = renderBoard();
+    expect(dragTargetIsSessionPane({ target: header })).toBe(false);
+  });
+
+  it("leaves the frame padding around the grid to the board", () => {
+    const { frame } = renderBoard();
+    expect(dragTargetIsSessionPane({ target: frame })).toBe(false);
+  });
+
+  it("leaves the hidden-sessions bar to the board", () => {
+    // It renders as a sibling *after* the grid element, so widening the
+    // selector to the grid container did not swallow it.
+    const { hiddenBar } = renderBoard();
+    expect(dragTargetIsSessionPane({ target: hiddenBar })).toBe(false);
+  });
+
+  it("leaves the board's own surface to the board", () => {
+    const { board } = renderBoard();
+    expect(dragTargetIsSessionPane({ target: board })).toBe(false);
   });
 
   it("treats a target that is not an element as board space", () => {
     expect(dragTargetIsSessionPane({ target: null })).toBe(false);
+    expect(dragTargetIsSessionPane({ target: document.createTextNode("x") })).toBe(false);
     expect(dragTargetIsSessionPane({ target: {} as EventTarget })).toBe(false);
   });
 });
--- a/packages/panel/src/lib/board-drop-arbiter.test.ts
+++ b/packages/panel/src/lib/board-drop-arbiter.test.ts
@@ -1,0 +1,25 @@
+import { describe, expect, it } from "vitest";
+import { dragTargetIsSessionPane } from "./board-drop-arbiter";
+
+/** Stands in for the element a drag event reports as its target. `matches` is
+ *  the set of selectors this element (or one of its ancestors) answers to. */
+function targetMatching(...matches: string[]): EventTarget {
+  return {
+    closest: (selector: string) => (matches.includes(selector) ? { tag: selector } : null),
+  } as unknown as EventTarget;
+}
+
+describe("dragTargetIsSessionPane", () => {
+  it("claims a drag whose target sits inside a grid cell", () => {
+    expect(dragTargetIsSessionPane({ target: targetMatching("[data-grid-cell]") })).toBe(true);
+  });
+
+  it("leaves the board's own space to the board", () => {
+    expect(dragTargetIsSessionPane({ target: targetMatching() })).toBe(false);
+  });
+
+  it("treats a target that is not an element as board space", () => {
+    expect(dragTargetIsSessionPane({ target: null })).toBe(false);
+    expect(dragTargetIsSessionPane({ target: {} as EventTarget })).toBe(false);
+  });
+});
--- a/packages/panel/src/lib/board-drop-arbiter.ts
+++ b/packages/panel/src/lib/board-drop-arbiter.ts
@@ -8,16 +8,46 @@
 // somewhere the operator did not aim.
 //
 // Nested drop targets need an arbiter, and the only honest arbiter for a
-// pointer gesture is where the pointer was. A grid cell marks itself with
-// `data-grid-cell` (SessionGrid), which is already the repo's way of asking
-// "is this inside a session pane?" — the same attribute the grid's own
-// keyboard navigation walks up to. So the board asks the same question of the
-// drag's target, and stands down when the answer is yes.
+// pointer gesture is where the pointer was. Two markers answer that, and it
+// takes both:
+//
+//   [data-grid-cell]     one session's cell (SessionGrid) — the same attribute
+//                        the grid's own keyboard navigation walks up to.
+//   [data-session-grid]  the grid container itself. Not redundant: the row and
+//                        column resize handles are absolutely-positioned
+//                        children of *this* element, siblings of the row divs
+//                        and inside no cell, at zIndex 6 — above every
+//                        non-expanded cell. Their hit area is HANDLE_HIT = 8px
+//                        while the shipped layout runs gridGap = 0 and
+//                        gridPad = 0, so each one is an 8px strip centred on a
+//                        0-width seam: it lies on top of terminal pixels, 4px
+//                        into each neighbour, full pane height for a column
+//                        divider and full grid width for a row divider. Cell
+//                        alone answers null there, and the board would paint
+//                        itself hot and upload to the Project root 4px from a
+//                        pane edge — #401's exact failure mode, in the one
+//                        place the operator cannot see that they are not on a
+//                        pane.
+//
+// So the whole grid element is session space. With gridGap and gridPad both 0
+// there is no board space left inside it to protect: the board's own space in
+// grid view is the project header strip and the CardFrame's 8px padding, both
+// *outside* the grid element, plus the hidden-sessions bar, which renders as a
+// sibling after it. Those still upload to the Project root, as they should.
 //
 // Single-session view needs nothing here: `TerminalPanel` renders as a sibling
 // of the board in `__root.tsx`, never a descendant, so its drops never reach
 // the board handler in the first place.
-const SESSION_PANE_SELECTOR = "[data-grid-cell]";
+//
+// Known and accepted gap: a menu portalled out of a cell to `document.body`
+// (TerminalPane's session-actions menu) still bubbles to the board through the
+// React tree while its DOM sits outside the grid, so `closest` answers null and
+// a drop on the open menu uploads to the Project root. It takes a menu held
+// open across a drag, and a menu is not a terminal, so it is recorded rather
+// than fixed. A DOM-position arbiter cannot see a React-tree ancestor; closing
+// that would mean a second signal (a dataset marker on the portal, or the
+// menu claiming the drag itself).
+const SESSION_PANE_SELECTOR = "[data-grid-cell], [data-session-grid]";
 
 /** The subset of `Element` this needs — duck-typed so the check is unit
  *  testable without a DOM, and so a non-Element target (a text node, `null`)
@@ -37,8 +67,8 @@ function isClosestCapable(target: unknown): target is ClosestCapable {
  * own space.
  *
  * True means the board must not treat the gesture as a Project-root upload:
- * the operator aimed at a session, and the space between and around the cells
- * is the only part of the board they can aim at when the grid is up.
+ * the operator aimed at a session, or at the grid's own furniture lying across
+ * one.
  */
 export function dragTargetIsSessionPane(event: { target: EventTarget | null }): boolean {
   const target: unknown = event.target;
--- a/packages/panel/src/lib/board-drop-arbiter.ts
+++ b/packages/panel/src/lib/board-drop-arbiter.ts
@@ -1,0 +1,47 @@
+// Which nested drop target owns a file drop on a Project board (#401).
+//
+// A Project's board is itself a drop target: a file dropped anywhere on it
+// uploads to the Project root. In grid view the whole session grid renders
+// *inside* that board, so every drag aimed at a session's terminal also
+// bubbles to the board's handler — and the board, having no idea what the
+// operator was aiming at, uploads to the Project root. The file lands
+// somewhere the operator did not aim.
+//
+// Nested drop targets need an arbiter, and the only honest arbiter for a
+// pointer gesture is where the pointer was. A grid cell marks itself with
+// `data-grid-cell` (SessionGrid), which is already the repo's way of asking
+// "is this inside a session pane?" — the same attribute the grid's own
+// keyboard navigation walks up to. So the board asks the same question of the
+// drag's target, and stands down when the answer is yes.
+//
+// Single-session view needs nothing here: `TerminalPanel` renders as a sibling
+// of the board in `__root.tsx`, never a descendant, so its drops never reach
+// the board handler in the first place.
+const SESSION_PANE_SELECTOR = "[data-grid-cell]";
+
+/** The subset of `Element` this needs — duck-typed so the check is unit
+ *  testable without a DOM, and so a non-Element target (a text node, `null`)
+ *  is simply "not a session pane" rather than a throw. */
+type ClosestCapable = { closest(selector: string): unknown };
+
+function isClosestCapable(target: unknown): target is ClosestCapable {
+  return (
+    typeof target === "object" &&
+    target !== null &&
+    typeof (target as ClosestCapable).closest === "function"
+  );
+}
+
+/**
+ * Whether a drag event landed inside a session pane rather than on the board's
+ * own space.
+ *
+ * True means the board must not treat the gesture as a Project-root upload:
+ * the operator aimed at a session, and the space between and around the cells
+ * is the only part of the board they can aim at when the grid is up.
+ */
+export function dragTargetIsSessionPane(event: { target: EventTarget | null }): boolean {
+  const target: unknown = event.target;
+  if (!isClosestCapable(target)) return false;
+  return target.closest(SESSION_PANE_SELECTOR) != null;
+}
--- a/packages/panel/src/routes/projects.$id.tsx
+++ b/packages/panel/src/routes/projects.$id.tsx
@@ -25,6 +25,7 @@ import { GridLayoutButton } from "~/components/views/GridLayoutButton";
 import { ProjectFilesPanel } from "~/components/views/ProjectFilesPanel";
 import { SessionGrid } from "~/components/views/SessionGrid";
 import { filesFromDrop, type DroppedFile } from "~/lib/core-files";
+import { dragTargetIsSessionPane } from "~/lib/board-drop-arbiter";
 import { dragCarriesFiles, useProjectFilesAvailability } from "~/lib/use-project-files";
 import { takeProjectFileDrop } from "~/lib/pending-file-drop";
 import { archiveOpenSession, invalidateSessionQueries } from "~/lib/archive-session";
@@ -1842,11 +1843,25 @@ function ProjectPage() {
         // Project on its Core. `dragCarriesFiles` is what keeps this off the
         // Panel's own drags — a project path dragged from the rail into a
         // terminal still belongs to the handler that already understands it.
+        //
+        // And `dragTargetIsSessionPane` is what keeps it off the sessions
+        // (#401). In grid view the grid renders inside this div, so a drag
+        // aimed at a session's terminal bubbles here too; the board answering
+        // it would upload to the Project root, which is not where the operator
+        // was pointing. It still has to `preventDefault` over a cell — leaving
+        // the browser's default alone makes the drop navigate the Panel to the
+        // dropped file — but it says `none`, so the cursor reads no-drop and
+        // the board never paints itself hot for a gesture it will refuse.
         onDragOver={
           filesAvailable
             ? (e) => {
                 if (!dragCarriesFiles(e)) return;
                 e.preventDefault();
+                if (dragTargetIsSessionPane(e)) {
+                  e.dataTransfer.dropEffect = "none";
+                  setFileDropHot(false);
+                  return;
+                }
                 e.dataTransfer.dropEffect = "copy";
                 setFileDropHot(true);
               }
@@ -1857,8 +1872,15 @@ function ProjectPage() {
           filesAvailable
             ? (e) => {
                 if (!dragCarriesFiles(e)) return;
+                // Swallowed before the arbiter runs: whoever the drop belonged
+                // to, the browser's default for a file dropped on a page is to
+                // navigate to it, and that must not happen on either path.
                 e.preventDefault();
                 setFileDropHot(false);
+                // Aimed at a session, so it was never the board's to take
+                // (#401). No upload, no drawer, no toast — refusing quietly is
+                // what the no-drop cursor above already told the operator.
+                if (dragTargetIsSessionPane(e)) return;
                 // Opened first: the upload is about to run, and a progress list
                 // nobody can see is the same as no progress at all.
                 setShowFiles(true);
--- a/packages/panel/src/routes/projects.$id.tsx
+++ b/packages/panel/src/routes/projects.$id.tsx
@@ -1848,10 +1848,20 @@ function ProjectPage() {
         // (#401). In grid view the grid renders inside this div, so a drag
         // aimed at a session's terminal bubbles here too; the board answering
         // it would upload to the Project root, which is not where the operator
-        // was pointing. It still has to `preventDefault` over a cell — leaving
-        // the browser's default alone makes the drop navigate the Panel to the
-        // dropped file — but it says `none`, so the cursor reads no-drop and
-        // the board never paints itself hot for a gesture it will refuse.
+        // was pointing. It still has to `preventDefault` over the grid —
+        // leaving the browser's default alone makes the drop navigate the Panel
+        // to the dropped file — but it says `none`, so the cursor reads no-drop
+        // and the board never paints itself hot for a gesture it will refuse.
+        //
+        // Writing `dropEffect` unconditionally is safe only while nothing
+        // inside the grid handles a file drag. This handler runs last, on the
+        // bubble, so it stomps whatever a nested one set: if a per-session drop
+        // is ever added — an image attached into a grid cell — it will
+        // `preventDefault` and set `copy`, and the `none` below would overwrite
+        // it and suppress that pane's `drop` entirely. The repo's convention
+        // for nested drop targets is the inner one claiming the gesture with
+        // `stopPropagation` (`ProjectFilesPanel.tsx`, #400/#227); a per-session
+        // drop should do the same, and then this branch never runs for it.
         onDragOver={
           filesAvailable
             ? (e) => {
--- a/packages/panel/src/components/views/SessionGrid.tsx
+++ b/packages/panel/src/components/views/SessionGrid.tsx
@@ -52,9 +52,14 @@ const DRAG_THRESHOLD_PX = 4;
 // below) — the resize math needs them to place divider handles precisely.
 const GRID_GAP = 8;
 const GRID_PADDING = 8;
-// Hit area for a divider handle — exactly the cell gap, so the invisible
-// handles (zIndex above the cells) never overhang the terminal surfaces and
-// steal pointerdowns near a cell's inner edge.
+// Hit area for a divider handle — 8px, the cell gap this grid was first laid
+// out with. It is no longer a gap: the shipped Studio look runs gridGap = 0 and
+// gridPad = 0 (see the layout block below), so each handle is an 8px strip
+// centred on a 0-width seam and *does* overhang the terminal surfaces, 4px into
+// each neighbour. That is deliberate — a grabbable divider on a flush layout
+// has to come from somewhere — but it means the handles sit on top of terminal
+// pixels while belonging to no cell, which anything reasoning about "is this
+// point on a session?" has to account for (#401, `board-drop-arbiter.ts`).
 const HANDLE_HIT = GRID_GAP;
 // Cards glide between grid slots on reorder — same timing/easing as the
 // sidebar's pinned-project slide so the two motions feel related.
--- a/packages/shared/src/__tests__/core-cert-material.test.ts
+++ b/packages/shared/src/__tests__/core-cert-material.test.ts
@@ -6,6 +6,7 @@ import {
   generateCertMaterial,
   issueServerCert,
   generateClientCsr,
+  randomSerial,
   signClientCsr,
 } from "../core-cert-material";
 
@@ -318,6 +319,26 @@ describe("signing a client CSR against the Core's CA", () => {
     expect(a.serial).not.toBe(b.serial);
     // Positive, per RFC 5280 §4.1.2.2 — the top bit of the first byte is clear.
     expect(parseInt(a.serial.slice(0, 2), 16)).toBeLessThan(0x80);
+    // And the whole 16 bytes came back. A serial minted with a leading zero
+    // byte returns a byte shorter than it went in, because DER writes an
+    // INTEGER minimally and the reader strips the pad — see `randomSerial`.
+    for (const serial of [a.serial, b.serial]) expect(serial).toMatch(/^[0-9a-f]{32}$/);
+  });
+
+  it("never mints a serial the certificate would hand back shortened", () => {
+    // The property, drawn often enough to bite. Signing a certificate per draw
+    // would take minutes; the issuances above already pin that the serial the
+    // certificate reports is the serial `randomSerial` produced, and this pins
+    // that every draw is a serial that survives that round trip: 16 bytes, top
+    // bit of the first clear so the INTEGER is positive, and the first byte
+    // never 0x00 so nothing is stripped off the front of it.
+    for (let i = 0; i < 5000; i++) {
+      const serial = randomSerial();
+      expect(serial).toMatch(/^[0-9a-f]{32}$/);
+      const first = parseInt(serial.slice(0, 2), 16);
+      expect(first).toBeGreaterThan(0x00);
+      expect(first).toBeLessThan(0x80);
+    }
   });
 
   it("mints a client CSR whose private key it hands back and never embeds", async () => {
--- a/packages/shared/src/core-cert-material.ts
+++ b/packages/shared/src/core-cert-material.ts
@@ -538,10 +538,28 @@ async function importCaKey(pem: string): Promise<X509SigningKey> {
   }
 }
 
-/** A positive 16-byte serial, hex-encoded — RFC 5280 §4.1.2.2 wants positive. */
-function randomSerial(): string {
+/**
+ * A positive 16-byte serial, hex-encoded — RFC 5280 §4.1.2.2 wants positive.
+ *
+ * The leading byte is forced into 0x01..0x7f, and the floor matters as much as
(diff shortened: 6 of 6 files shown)


qcentic-dev:
## Acceptance criteria — issue #401

Issue #401 states one numbered acceptance criterion; it is broken into its two clauses below, plus the constraints the
issue's "What to do" imposes on the fix.

- [x] **1. A file drop over a session terminal in grid view does not silently upload to the project root.**
  The board's `onDrop` now returns before the upload when `dragTargetIsSessionPane(e)` is true — the drag's target
  resolves up to a `[data-grid-cell]`. No upload, no Files drawer opened, no pending drop queued.
- [x] **1a. …"or does so only with an explicit overlay".**
  Taken as the refuse branch, not the overlay branch: over a cell `onDragOver` sets `dropEffect: "none"` and keeps the
  board's hot dashed outline off, so the gesture is refused with a no-drop cursor rather than accepted silently. The
  board is only painted hot where it really will take the file. No new overlay UI was added — the issue offers the two
  as alternatives, and refusing is the smaller change and the one that matches ADR 0010's stance that the Panel cannot
  name a path on the Core for a dropped terminal file.
- [x] **"Nested drop targets need an arbiter" (What to do).**
  New `packages/panel/src/lib/board-drop-arbiter.ts`. The arbiter is the pointer's position: it walks up from the
  drag's target to `[data-grid-cell]`, the marker `SessionGrid` already puts on every cell and that the grid's own
  keyboard navigation already walks up to. Unit-tested in `board-drop-arbiter.test.ts` (3 tests, passing).
- [x] **"Same handler family as C (`projects.$id.tsx` board `onDrop`)" — the fix belongs on the board side.**
  Both edits are on the board handlers in `projects.$id.tsx`. `ProjectFilesPanel.tsx` (#400, already merged) is not
  touched; that is a different handler and its `stopPropagation` fix stands as merged.
- [x] **The `e.defaultPrevented` question (asked in the task, from the #400 landmine): checked, and rejected.**
  Two reasons. (a) A session pane never sets it for this gesture — `wireTerminalFileDrop` `preventDefault`s only for a
  project-path drag and deliberately ignores desktop file drags, because the browser hands over bytes and not a path
  on the Core (ADR 0010); a guard on the flag would be a no-op and the bug would ship unchanged. (b) Even if something
  did set it, React's `SyntheticEvent` copies `defaultPrevented` from the native event at construction, so a React
  descendant's `preventDefault` in the same root dispatch would not be visible to the board. Reasoning is in the PR
  body and in the commit message.
- [x] **The browser's own default is still swallowed on both paths.**
  Over a cell, `onDragOver` and `onDrop` both still call `preventDefault` before standing down. Dropping this would
  have traded the wrong-destination bug for a worse one: the browser navigating the whole Panel to the dropped file.
- [x] **No regression to the board's own upload.**
  The refusal is scoped to targets inside a `[data-grid-cell]`. The board's own space — the gaps around and between
  cells, the header, and every part of single-session view — uploads to the project root exactly as before.
  `TerminalPanel` needed no change: it renders as a sibling of the board in `__root.tsx`, never a descendant, so its
  drops never reached this handler in the first place. A comment records that so the next reader does not hunt for a
  missing case.

### Not verified by a click-through — stated plainly

- [ ] **Live operator verification in a running Panel with a real Core.**
  Unchecked. Issue #401 is itself filed as `Live on pairdemo: code-read` — the Panel UI was never clicked in the hunt
  — and this change was verified by typecheck, lint, a unit test on the arbiter, and code reading, not by dragging a
  file onto a grid cell in a browser against a paired Core. The two DOM facts it rests on are load-bearing and both
  are checked in the source: `SessionGrid` renders `data-grid-cell` on every cell, and `SessionGrid` is rendered
  inside the board `div` in `projects.$id.tsx`.
- [ ] **A green full `pnpm test` on this machine.**
  Unchecked, and pre-existing. The panel package has flaky socket-timeout failures here under parallel load. Measured
  both ways on the same box: 7 failed files at the unmodified base commit, 3 with this change applied; each one passes
  when run as a single file. None sit in the touched code path. `pnpm typecheck` is clean and `pnpm lint` reports
  0 errors, with none of its 45 pre-existing warnings in the touched files. CI is the arbiter of this one.

https://claude.ai/code/session_01FBR3m1VZWiuWVSoRWUEqGU

packages/panel/src/lib/board-drop-arbiter.ts
```diff
@@ -0,0 +1,47 @@
...
+// pointer gesture is where the pointer was. A grid cell marks itself with
+// `data-grid-cell` (SessionGrid), which is already the repo's way of asking
+// "is this inside a session pane?" — the same attribute the grid's own
+// keyboard navigation walks up to. So the board asks the same question of the
+// drag's target, and stands down when the answer is yes.
+//
+// Single-session view needs nothing here: `TerminalPanel` renders as a sibling
+// of the board in `__root.tsx`, never a descendant, so its drops never reach
+// the board handler in the first place.
+const SESSION_PANE_SELECTOR = "[data-grid-cell]";
```
mrfqcentic: **Blocking.** `[data-grid-cell]` alone leaves a hole exactly where two panes meet.

`SessionGrid` renders the row and column resize handles as absolutely-positioned children of the grid container — siblings of the row `div`s, inside no cell — at `zIndex: 6`, above every non-expanded cell (`SessionGrid.tsx:2428-2469`). Their hit area is `HANDLE_HIT = GRID_GAP = 8` (`SessionGrid.tsx:53`, `:58`) while the shipped layout runs `gridGap = 0; gridPad = 0` and says so itself: *"no gap between cells, no outer padding. The divider hit-area (HANDLE_HIT) stays a grabbable 8px, centred on the 0-width seam"* (`SessionGrid.tsx:827-832`).

So every seam between two panes carries an 8px strip — full pane height for a column divider, full grid width for a row divider — that lies **on top of terminal pixels** and answers `closest("[data-grid-cell]")` with `null`. `dragTargetIsSessionPane` returns `false` there, so the board paints itself hot, shows `copy`, and uploads to the project root: AC1's exact failure mode, 4px from the pane edge, with feedback that says "yes, drop here". (The older comment at `SessionGrid.tsx:55-57` — "the invisible handles never overhang the terminal surfaces" — is stale for the flush layout; the one at `:827` is the accurate one.)

It also undercuts the reasoning in the PR body and in the module comment at line 38-40 above. With `gridGap` and `gridPad` both 0 there are no "gaps around and between cells"; inside `[data-session-grid]` the *only* board space is the divider strips, which is the one place the operator cannot see that they are not on a pane. The board's real own space in grid view is the project header strip and the `CardFrame`'s 8px padding — both **outside** the grid element.

A one-line fix that keeps all of that:

```ts
const SESSION_PANE_SELECTOR = "[data-grid-cell], [data-session-grid]";
```

`data-session-grid` is already on the grid container (`SessionGrid.tsx:2367`); the hidden-sessions bar is a sibling *after* it (`SessionGrid.tsx:2526`) so it stays board space, and the header strip and frame padding are untouched. Whichever selector you land on, please add the divider case to the test — it is the case that decides this.

packages/panel/src/routes/projects.$id.tsx
```diff
@@ -1842,11 +1843,25 @@ function ProjectPage() {
...
+        // was pointing. It still has to `preventDefault` over a cell — leaving
+        // the browser's default alone makes the drop navigate the Panel to the
+        // dropped file — but it says `none`, so the cursor reads no-drop and
+        // the board never paints itself hot for a gesture it will refuse.
         onDragOver={
           filesAvailable
             ? (e) => {
                 if (!dragCarriesFiles(e)) return;
                 e.preventDefault();
+                if (dragTargetIsSessionPane(e)) {
```
mrfqcentic, line 1870: Two smaller notes on this branch.

1. The hot outline and `dropEffect` are right over a cell but wrong over the divider strips described in my comment on `board-drop-arbiter.ts:20` — the board goes hot on an 8px band lying across two terminals.

2. This handler runs last (bubble phase) and writes `dropEffect` unconditionally, so it stomps whatever a nested handler set. Benign today, since nothing inside a cell handles a file drag — but it inverts the convention the repo already uses for nested drop targets, where the inner target claims the gesture with `stopPropagation` (`ProjectFilesPanel.tsx:191`, `:304`, per #400 / #227). If a per-session drop is ever added — an image attach into a grid cell — it will call `preventDefault()` and set `dropEffect = "copy"`, this line will overwrite it with `"none"`, and that pane's `drop` will never fire. Worth a comment recording that, at least.

packages/panel/src/lib/board-drop-arbiter.test.ts
```diff
@@ -0,0 +1,25 @@
+import { describe, expect, it } from "vitest";
+import { dragTargetIsSessionPane } from "./board-drop-arbiter";
+
+/** Stands in for the element a drag event reports as its target. `matches` is
+ *  the set of selectors this element (or one of its ancestors) answers to. */
+function targetMatching(...matches: string[]): EventTarget {
+  return {
+    closest: (selector: string) => (matches.includes(selector) ? { tag: selector } : null),
```
mrfqcentic: The predicate is tested; the wiring is not — and the fake is what let the divider case through. `targetMatching` answers `closest` against the same string literal the module holds, so this suite passes for *any* selector as long as both sides agree. It cannot tell you whether real grid markup satisfies it.

The repo already has the right precedent: `components/views/__tests__/project-files-panel.test.tsx:403` — "a drop on the drawer, with the board behind it" — renders the real component inside a stand-in `<div data-testid="board" onDragOver onDrop>` under `@vitest-environment jsdom`, fires `dragOver`/`drop`, and asserts the board spy was never called. A ~30-line equivalent here (a stand-in board wrapping `[data-session-grid]` > `[data-grid-cell]` > `div.xterm`, a divider sibling of the rows, and a header strip outside the grid) would pin all three cases and would have caught the divider one. `jsdom` is already a devDependency and 27 panel suites opt into it per-file.

packages/panel/src/routes/projects.$id.tsx
```diff
@@ -1857,8 +1872,15 @@ function ProjectPage() {
...
                 if (!dragCarriesFiles(e)) return;
+                // Swallowed before the arbiter runs: whoever the drop belonged
+                // to, the browser's default for a file dropped on a page is to
+                // navigate to it, and that must not happen on either path.
                 e.preventDefault();
                 setFileDropHot(false);
+                // Aimed at a session, so it was never the board's to take
+                // (#401). No upload, no drawer, no toast — refusing quietly is
+                // what the no-drop cursor above already told the operator.
+                if (dragTargetIsSessionPane(e)) return;
```
mrfqcentic, line 1893: Non-blocking, but worth a decision: with a cell expanded, the refusal covers the whole board below the header. `SessionGrid.tsx:568` gives the expanded cell `position: absolute; inset: gridPadding`, so it covers the entire grid content box, and with `gridPad = 0` that is everything under the header. An operator who expands a session and then drags a file in to upload it to the Project gets a no-drop cursor across the whole surface and nothing pointing at the Files drawer or the header strip. The `dropEffect: "none"` cursor is honest feedback and I agree it beats an overlay for the pane case — this is just the one configuration where "refuse quietly" leaves no visible way forward.

Separately, as a note for a follow-up rather than for this PR: when the Core announces no file surface, or the Project is Panel-owned, `filesAvailable` is false and both handlers are `undefined` (`projects.$id.tsx:1856`, `:1872`), so a desktop file drop anywhere on the board — session panes included — falls through to the browser default and navigates the Panel off the app. That predates this change and is outside #401, but it is exactly the hazard the comment above cites as the reason to keep `preventDefault`.

mrfqcentic (changes requested):
**Request changes — one blocking item. Everything else in this PR checks out.**

Reviewed at `361b6d9` against `fix/operator-pins-sessions-drop-cli`, reading the enforcing lines rather than the description.

**Verified**

- **A drop aimed at a session pane no longer uploads to the project root.** `projects.$id.tsx:1883` returns before `setShowFiles`/`filesFromDrop`/`setPendingFileDrop`, and `dragTargetIsSessionPane` resolves the drag target up to `[data-grid-cell]` — which `SessionGrid` really does put on every cell (`SessionGrid.tsx:557`, spread onto the DOM by `CardFrame`), with `TerminalPane` inside it, and `SessionGrid` really is rendered inside the board `div` (`projects.$id.tsx:1838` → `CardFrame` → `:2186`). With `dropEffect: "none"` set on `dragover` the `drop` event will not fire at all, so `:1883` is belt-and-braces rather than the only guard.
- **A drop on the board outside any pane still uploads.** The refusal is scoped to `[data-grid-cell]` descendants; the project header strip, the `CardFrame`'s 8px padding around the grid, the hidden-sessions bar, the empty-grid state and all of list view are unchanged.
- **#400 is not reopened or duplicated.** `ProjectFilesPanel.tsx` is untouched; its `stopPropagation` on both `onDragOver` and `onDrop` (`:191`, `:208`) still means the board never hears a drawer drop, so the arbiter never runs for it. No second upload path, no interaction between the two fixes.
- **The `e.defaultPrevented` shape #400 named for this file is correctly rejected.** `wireTerminalFileDrop` (`terminal-pane-helpers.ts:109-121`) only `preventDefault`s for a project-path drag; a desktop file drag returns at `if (!projectPath) return`, so the flag is never set for this gesture. ADR 0010 backs the reason it never will be ("`webUtils.getPathForFile` has no web equivalent"). The `SyntheticEvent`-construction argument is a correct second reason.
- **Single-session view needs nothing.** `ProjectTerminalPanel` is a sibling of the `<Outlet />` wrapper in `__root.tsx:672`, and only mounts when `!gridActive`. The comment in the code is accurate.
- **Scope is clean.** Three files. `ProjectBar.tsx`, `use-fleet.ts`, `shared/projects.ts`, `queries/index.ts` and `visible-project-scope.ts` (#382 / #381) are untouched.
- **The base has moved and the branch still merges.** Cut from `03d22eb`; base is now `d727928` (#472 was the cut itself, #471 and #462 landed after). `git merge-tree` is conflict-free and GitHub reports `MERGEABLE` / `CLEAN`; neither landed PR touches `projects.$id.tsx` or the new module.
- **CI is green** on all 20 checks, Typecheck / Lint / Unit Tests included — which answers the "flaky locally" caveat in the PR body.

**Acceptance criteria — issue #401**

| | |
|---|---|
| AC1 — a drop over a session terminal in grid view does not silently upload to the project root | **partially met** — met over the body of a pane, not over the 8px resize-divider strips that lie across the seam between two panes (see the comment on `board-drop-arbiter.ts:20`) |
| AC1a — "or does so only with an explicit overlay" | **met** — taken as the refuse branch: `dropEffect: "none"` plus the hot outline held off. The right call; there is no per-session upload target to offer (ADR 0010), so refusing beats inventing one |
| "Nested drop targets need an arbiter" | **met** — a real arbiter in its own module, on the pointer's position, reusing the marker the grid's own keyboard nav already walks up to |
| Same handler family as C (`projects.$id.tsx` board `onDrop`) | **met** |
| Does not silently swallow the operator's files | **met over a pane** — the no-drop cursor is honest feedback given during the drag, before the release. See the note on `:1883` for the one configuration where it leaves no way forward |

**Blocking**

The discriminator can be defeated, and in the one place where the operator cannot tell: `SessionGrid`'s resize divider handles are children of the grid container, not of any cell, sit at `zIndex: 6` above the panes, and are 8px wide/tall centred on a 0-width seam — so they overhang 4px of each adjacent terminal. A drop there goes to the project root, with the board painted hot. Detail and a one-line fix in the comment on `board-drop-arbiter.ts:20`.

Two other cases I checked, for the record: a drop on the **gap between cells** is a non-case in the shipped layout (`gridGap = 0`, `gridPad = 0`, `SessionGrid.tsx:831-832`) — the only in-grid "gap" is the divider strip above. A drop on a **portal rendered from inside a cell** (the session-actions menu, `TerminalPane.tsx:299` → `document.body`) bubbles to the board through the React tree while its DOM sits outside the cell, so `closest` returns `null` and it uploads to the project root; that needs the menu open mid-drag and the menu is not a terminal, so I would leave it, but it is the same class of hole and worth a line of comment.

The shape chosen is the right one — better than what the #400 landmine anticipated, and the PR says why with evidence rather than assertion. The scope discipline, the commenting and the merge hygiene are all good. It is one selector away from done.

https://claude.ai/code/session_01Df49QoHus1aP4bjS7nNLdf

packages/panel/src/lib/board-drop-arbiter.ts
qcentic-dev: Fixed at `ac59967`. The divider seam was a real hole and the review's read of it is exact — handles are children of the
grid container at `zIndex: 6`, `HANDLE_HIT = 8` against `gridGap = 0`, so an 8px strip painted over 4px of each
neighbouring terminal, answering `null` to `closest("[data-grid-cell]")`. Board went hot, showed `copy`, uploaded to
the project root, 4px from a pane edge.

**Fixed**

- **The selector**, as suggested: `"[data-grid-cell], [data-session-grid]"`. The whole grid element is session space.
  Confirmed the three exclusions hold: the hidden-sessions bar is a sibling *after* the grid (`SessionGrid.tsx:2526`),
  and the project header strip and the `CardFrame`'s 8px padding are outside it — all three still upload to the
  project root.
- **The test suite**, rebuilt on real DOM under `jsdom` rather than a stub, since the stub agreed with any selector as
  long as both sides used the same literal. Markup mirrors `SessionGrid`: board → frame → header →
  `[data-session-grid]` → `[data-grid-row]` → `[data-grid-cell]` → `.xterm`, a divider as a sibling of the rows inside
  the grid, hidden-sessions bar after it. 9 cases, covering the divider seam, the grid container, the header strip, the
  frame padding, the hidden bar and the bare board. Checked that **the divider case fails against the old
  `[data-grid-cell]`-only selector** — it is the case that decides the file, so it had better bite.
- **The stale comment at `SessionGrid.tsx:55-57`.** Corrected in place: it now says the handles *do* overhang the
  terminal surfaces by 4px each side under the flush layout, why that is deliberate, and that anything reasoning about
  "is this point on a session?" has to account for it — pointing at `board-drop-arbiter.ts`.
- **The justification**, in both the module comment and the PR body. "The space between and around the cells" was
  wrong: with `gridGap` and `gridPad` both 0 there is no such space, and the thing that looks like it is the one place
  the operator cannot tell they are not on a pane. Both now say the board's own space in grid view is entirely outside
  the grid element.

**Recorded, not fixed**

- **3.3c, the portalled menu.** A line in `board-drop-arbiter.ts` naming it: `TerminalPane.tsx:299` portals the
  session-actions menu to `document.body` (`:392`), React propagates synthetic events along the React tree, so a drop
  on the open menu reaches the board while its DOM sits outside the grid and `closest` answers `null`. Needs a menu
  held open across a drag, and a menu is not a terminal. The comment also says why the fix is not another selector: a
  DOM-position arbiter cannot see a React-tree ancestor, so closing it needs a second signal.
- **Your note on `projects.$id.tsx:1860` — `dropEffect` written unconditionally on the bubble.** Commented on the
  handler: benign while nothing inside the grid handles a file drag, but it inverts the `stopPropagation` convention
  (`ProjectFilesPanel.tsx`, #400/#227), and a future per-session image-attach drop that sets `copy` would have it
  stomped and its `drop` suppressed. The comment says a per-session drop should claim the gesture the repo's way.
- **3.5, the expanded cell.** Filed as **#479** rather than carried here — it is a feedback gap, not a wrong
  destination, and #401's acceptance is about where the file lands. The issue records that `SessionGrid.tsx:568`'s
  `inset: gridPadding` with `gridPad = 0` puts the refusal across everything under the header, lists the three ways to
  give it somewhere to point, and references #401 and this PR.
- **Your `filesAvailable === false` note** — both handlers `undefined`, so a desktop file drop anywhere on the board
  falls through to the browser default and navigates the Panel off the app. Agreed it predates this change and is
  outside #401; not filed, since it is your finding to shape and it may want to ride with the drawer work rather than
  with this. Say the word and I will file it.

**Housekeeping**

Base merged in with a plain merge commit (`3fc380c`) after #471, #462 and #472 landed — no rebase, no force push. PR
body rewritten to match. `pnpm typecheck` clean; `eslint` on the four touched files reports 0 errors (full `pnpm lint`
OOMs on this box at exit 137, a local memory ceiling rather than a finding — CI runs it). Targeted suites green:
`board-drop-arbiter` 9, `project-files-panel` 19, `session-grid-layout` + `session-grid-tracks` 18.

Still a draft, still based on `fix/operator-pins-sessions-drop-cli`.

https://claude.ai/code/session_01FBR3m1VZWiuWVSoRWUEqGU

qcentic-dev: **Round 2 — a fresh session picked this up after the previous one was killed mid-turn (the box ran out of memory).
Everything the review asked for was already on the branch at `ac59967`; I re-verified each item against the code rather
than against the reply above, and fixed the one thing that was not done: CI was red.**

**Re-verified as landed at `ac59967`** (line numbers below are the current ones — the widened comment at `:55-62`
shifted everything below it in `SessionGrid` down by five lines, and I have refreshed the stale references in the PR body):

- **The selector.** `SESSION_PANE_SELECTOR = "[data-grid-cell], [data-session-grid]"`. `data-session-grid` really is on
  the grid container (`SessionGrid.tsx:2373`), the divider handles really are inside it and inside no cell
  (`:2433-2476`, `zIndex: 6` at `:2448` and `:2469`), and `{hiddenBar}` really is a sibling *after* the grid element
  (`:2531`) — so the bar, the header strip and the `CardFrame` padding still upload to the Project root.
- **The divider test.** 9 cases on real DOM under `jsdom`, markup shaped like `SessionGrid`'s. The divider case asserts
  `divider.closest("[data-grid-cell]")` is `null` first, so it fails against the old selector by construction.
- **The stale comment at `SessionGrid.tsx:55-57`.** Corrected in place (now `:55-62`): the handles *do* overhang the
  terminal surfaces by 4px each side under the flush layout, that is deliberate, and anything asking "is this point on
  a session?" has to account for it — with a pointer at `board-drop-arbiter.ts`.
- **The justification.** Module comment and PR body both rewritten: with `gridGap` and `gridPad` both 0 there is no
  space between or around the cells, and the board's own space in grid view is entirely outside the grid element.

**Recorded, not fixed** — also verified present: the portalled session-actions menu (`TerminalPane.tsx:299` → `:392`),
commented in `board-drop-arbiter.ts` with why a DOM-position arbiter cannot close it; the unconditional `dropEffect` on
the bubble, commented on the handler with the `stopPropagation` convention it should defer to; and the expanded-cell
feedback gap, filed as **#479**. Your `filesAvailable === false` note is still yours to shape — say the word and I will
file it.

**What this round actually changed: the red check, which was not this PR's.**

`Unit Tests` failed on `ac59967` in a package the branch does not touch — `packages/shared`, `core-cert-material` >
"gives every issuance its own serial", `expected 240 to be less than 128`. Not a regression: a 1-in-256 draw in
`randomSerial` that has been in the CA path since it was written, and the run at `361b6d9` you reviewed was green with
the same code.

`randomSerial` drew 16 bytes and cleared the top bit of the first, which makes the DER INTEGER positive per RFC 5280
§4.1.2.2. Two of the 256 possible first bytes — `0x00` and `0x80` — clear to `0x00`, and there the mask is not enough:
DER writes an INTEGER minimally and `@peculiar/x509` hands `serialNumber` back with the leading zero byte **stripped**,
so that issuance reports a 15-byte serial beginning at whatever byte sat behind the zero, which has its own top bit set
half the time. I checked that against the library rather than inferring it — minted `00f0aabb…` reports back
`f0aabb…`; minted `7f0011…` reports back unchanged.

That serial is the certificate's name in every later revocation list and `pair ls` row, so it has to be the same string
on both sides of the issuance. The leading byte is now forced into `0x01..0x7f`: top bit clear keeps it positive,
non-zero leaves the encoder nothing to strip. The test drew the property once per issuance, which is how a one-in-256
bug outlived every run of it — it now asserts the full 16 bytes came back on both issuances and draws `randomSerial`
5000 times to pin the bounds directly (exported for that, the way `pty-output-activity` exports its own internals).
Checked that the new test fails against the old one-line body.

It is in its own commit (`ccda088`), outside #401 and outside the panel, and it lifts out cleanly if you would rather it
rode on its own — but re-running the job would have left the flake for the next pull request to catch.

**State.** Head `ccda088`. Base `fix/operator-pins-sessions-drop-cli` merged in earlier with a plain merge commit
(`3fc380c`) and still current — the branch is 0 commits behind. No rebase, no force push, nothing merged, base
unchanged, still a draft. `packages/shared` 555 tests green locally, `board-drop-arbiter` 9 and `project-files-panel`
19 green, `tsc --noEmit` clean and `eslint` 0 errors on the touched files. CI on the head `ccda088`: **all 18 checks green**, 6 skipping
(the image publish jobs a draft does not run), `Unit Tests` included — [run
33755281458](https://github.com/actana/control/actions/runs/33755281458).

https://claude.ai/code/session_01ScnKQreqSyPjUNkbNGjUry

mrfqcentic (approved):
**Approve — round 2. The blocking hole from round 1 is closed, and I re-drove my own probe against the new head rather than reading the reply.**

Reviewed at `ccda088` against `fix/operator-pins-sessions-drop-cli` (`d727928`), enforcing lines read, tests run locally from a read-only export of the branch.

## The blocking item is closed, and I proved it two ways

`board-drop-arbiter.ts:50` is now `const SESSION_PANE_SELECTOR = "[data-grid-cell], [data-session-grid]";` and the markup facts hold on this head: `data-session-grid` on the grid container (`SessionGrid.tsx:2373`), the row and column resize handles as absolutely-positioned children of *that* element and inside no cell (`:2433-2476`, `zIndex: 6` at `:2448` and `:2469`), `{hiddenBar}` a sibling **after** the grid element (`:2531`), `data-grid-cell` spread to the DOM through `CardFrame`'s `{...props}` (`SessionGrid.tsx:562` → `CardFrame.tsx`).

1. **The PR's own suite.** 9 tests, real DOM under `jsdom`, and I checked it bites: reverting `:50` to `"[data-grid-cell]"` in a scratch copy fails 2 of the 9 — "claims a drop on a resize divider lying across the seam between two panes" and "claims a drop on the grid container itself". The `expect(divider.closest("[data-grid-cell]")).toBeNull()` line at `:76` is what makes that structural rather than incidental. The tautology from round 1 is gone: the fixture builds markup and calls the real `closest`, so the suite can now disagree with the module.
2. **My own probe, re-driven.** I wired the *real* `dragTargetIsSessionPane` into a board handler shaped like `projects.$id.tsx:1865-1900` and dispatched real `dragover`/`drop` events at a `SessionGrid`-shaped tree. On this head: divider seam → `dropEffect: "none"`, board never hot, `defaultPrevented` true, **no project-root upload**; terminal inside a cell → same; header strip, `CardFrame` padding, hidden-sessions bar and the bare board → `dropEffect: "copy"`, hot, **upload still happens**; a rail project-path drag (no `Files` type) → untouched on every surface, `defaultPrevented` false, `dropEffect` never written, so `wireTerminalFileDrop`'s own gesture is unaffected. With the selector reverted, the divider case flips to `copy` and uploads — the round-1 finding, reproduced independently and then closed.

## Acceptance criteria — issue #401

| Criterion | Verdict | Enforcing line |
|---|---|---|
| **AC1** — a file drop over a session terminal in grid view does not silently upload to the project root | **met** — including the divider seam, which was the round-1 gap | `projects.$id.tsx:1893` returns before `setShowFiles`/`filesFromDrop`/`setPendingFileDrop`; `board-drop-arbiter.ts:50` + `:77` |
| **AC1a** — "…or does so only with an explicit overlay" | **met** — the refuse branch: `dropEffect: "none"` and the hot outline held off | `projects.$id.tsx:1870-1874` |
| "Nested drop targets need an arbiter" (What to do) | **met** — a real arbiter in its own module, keyed on the pointer's position, with a regression test that fails on the pre-fix selector | `board-drop-arbiter.ts`, `board-drop-arbiter.test.ts` (9 tests) |
| Same handler family as C (`projects.$id.tsx` board `onDrop`) | **met** — both edits are on the board handlers; `ProjectFilesPanel.tsx` untouched, so #400 is neither reopened nor duplicated | `projects.$id.tsx:1865`, `:1881` |
| The board's own upload is not regressed | **met** — header strip, frame padding, hidden-sessions bar, empty-grid state and all of single-session view still upload; verified in the suite and in my probe | `SessionGrid.tsx:2531`, `:2364` (empty branch has no grid element), `__root.tsx:673` |
| The browser default is still swallowed on both paths | **met** — `preventDefault` before the arbiter on both handlers | `projects.$id.tsx:1869`, `:1888` |
| Live operator click-through | **not met, and honestly declared** — unchanged from round 1; #401 is itself a code-read filing. The two load-bearing DOM facts are checked in source and now in a DOM test |
| The `defaultPrevented` shape #400 named | **met** — correctly rejected; `wireTerminalFileDrop` returns at `if (!projectPath) return` for a desktop file drag (`terminal-pane-helpers.ts:117`), so the flag is never set for this gesture |

## The round-2 asks, one by one

- **Divider seam no longer uploads** — met (above).
- **Widening did not swallow drops the board should take** — met. The three surfaces round 1 named are all outside the grid element and all still upload: header strip (`.mc-project-header`, sibling *before* the grid inside `CardFrame`), the `CardFrame`'s 8px padding, the hidden-sessions bar (`:2531`). Asserted in the suite and re-driven in my probe.
- **A drop on the board outside the grid still uploads** — met; verified at handler level, not just predicate level.
- **The new seam test fails if the selector is reverted** — met; I ran it (2 failures).
- **Stale comment at `SessionGrid.tsx:55-57`** — corrected in place at `:55-62`, and the correction is accurate against `gridGap = 0; gridPad = 0` at `:836-837`.
- **Narrowed justification** — the module comment (`board-drop-arbiter.ts:33-38`) and the PR body now both say the board's own space in grid view is entirely outside the grid element. Accurate.
- **Recorded, not fixed** — the portalled session-actions menu is recorded in `board-drop-arbiter.ts:42-49` (and `TerminalPane.tsx:299` → `document.body` at `:392` checks out), the unconditional `dropEffect` on the bubble is recorded on the handler (`projects.$id.tsx:1856-1863`), and the expanded-cell case is filed as **#479**, which describes the mechanism correctly.

## The repair between `ac59967` and `ccda088` — checked, not trusted

It is a **real fix, and the test was strengthened, not weakened**.

`randomSerial` cleared the top bit of the leading byte, which maps both `0x00` and `0x80` to `0x00`; DER writes an INTEGER minimally, `@peculiar/x509` returns `serialNumber` with the leading zero stripped, so 2-in-256 draws produced a 15-byte serial whose new first byte had its own top bit set half the time — the `expected 240 to be less than 128` failure, at a rate of ~1-in-256 per issuance. The new body forces `0x01..0x7f`; the certificate then reports back the 16 bytes that were minted.

The pre-existing assertion (`parseInt(a.serial.slice(0,2),16) < 0x80`) is **retained unchanged**; what was added is strictly more: `/^[0-9a-f]{32}$/` on both issued serials (the shortening, which is the actual defect) and a 5000-draw property test on `randomSerial`. I re-drove it: reverting `core-cert-material.ts:561-562` to `bytes[0] = bytes[0]! & 0x7f;` in a scratch copy fails the new test ("expected 0 to be greater than 0") while the other 24 pass. Nothing was loosened to make a red check go green.

## State

- **Merges cleanly.** Base `fix/operator-pins-sessions-drop-cli` (`d727928`) is a full ancestor of `ccda088` via the plain merge at `3fc380c` — 0 behind. `git merge-tree` is conflict-free; GitHub reports `MERGEABLE` / `CLEAN`. #471 (`10c74eb`), #462 (`d727928`) and #472 (`03d22eb`) are all ancestors of the head.
- **CI green on `ccda088`** — 18 pass, 6 skipping (the image-publish jobs a draft does not run). Run 33755281458.
- **Locally, from a read-only export of the branch:** `board-drop-arbiter` 9/9, `project-files-panel` 19/19 (#400 unregressed), `session-grid-layout` + `session-grid-tracks` 18/18, `core-cert-material` 25/25.

## Non-blocking, for the next push or a follow-up

1. **The acceptance-criteria comment on this PR is stale.** [#issuecomment-5524738131](https://github.com/actana/control/pull/477#issuecomment-5524738131) has never been edited (created and updated both `11:04Z`). It still says the arbiter "walks up from the drag's target to `[data-grid-cell]`", that the refusal "is scoped to targets inside a `[data-grid-cell]`", that the board's own space is "the gaps around and between cells" — the claim you retracted — and that the module has "3 tests". The PR body was rewritten; this comment is the PR's acceptance record and now contradicts the shipped code. Please edit it or post a superseding one.
2. **`board-drop-arbiter.ts:11-12`** — "Two markers answer that, and it takes both" is the reverse of what the markup gives you. `GridCell` is rendered only at `SessionGrid.tsx:2404`, inside a row inside the grid element, and `CardFrame` does not portal — so every `[data-grid-cell]` is already a descendant of `[data-session-grid]`, and the grid marker alone decides every case in the suite. The cell marker is belt-and-braces (and cheap insurance if a cell is ever portalled or hoisted), which is fine — but a reader who takes the comment at face value could delete the *grid* half believing the cells carry the weight, and that is exactly the divider hole again. Worth one clause saying which half is load-bearing today.
3. **`board-drop-arbiter.ts:52-54`** — the `ClosestCapable` comment justifies the duck-typing as "so the check is unit testable without a DOM", which the new test file's own opening paragraph argues against. The other half of that sentence (a non-Element target is "not a session pane" rather than a throw) is the reason that survived.
4. **`SessionGrid.tsx:50-52`** — two lines above the comment you corrected, `GRID_GAP`/`GRID_PADDING` are still described as "kept in sync with the container style below" and as what "the resize math needs". Neither holds: the container uses the local `gridGap = 0`/`gridPad = 0` (`:836-837`), `dividerOffset`/`rowBand` read those locals, `GRID_PADDING` has no reader at all and `GRID_GAP`'s only reader is `HANDLE_HIT`. Two adjacent comments contradicting each other is the failure mode this hunk set out to fix.
5. **#479 is accurate but framed narrower than the behaviour.** With `gridPad = 0` the cells tile the whole grid content box, so even *without* an expanded cell the only surfaces under the header that accept a project-root drop are the header strip, the 12px band its bottom margin opens, and the 8px frame ring — the expanded cell is the starkest case, not the only one. Worth widening the issue's "What the operator sees". (For whoever picks it up: a rail project row is a third accepting target — `ProjectBar.tsx:1236` — also unlabelled during a drag.)
6. **`board-drop-arbiter.ts:38-40`, and a pre-existing hazard worth one issue.** "Single-session view needs nothing here" is true of *this handler* and I verified the sibling relationship (`__root.tsx:673`), but it reads as "that view is fine", and it is not: nothing anywhere calls `preventDefault` for a desktop file drag over `TerminalPanel`/`UserTerminalPanel` — `wireTerminalFileDrop` claims only `isProjectPathDrag` (`terminal-pane-helpers.ts:110`), and there is no window- or document-level guard in `packages/panel/src` (I grepped: the only `dragover`/`drop` listeners are that helper's two). So in single-session view, dropping a PNG on the terminal navigates the Panel to `file:///…` and tears down the SPA. That is the same hazard as my round-1 `filesAvailable === false` note, one level out. Both are pre-existing and outside #401 — please file one issue covering the pair; a single window-level `dragover`/`drop` guard closes both.
7. **Coverage stops at the predicate.** Nothing exercises the board's own `onDragOver`/`onDrop`; I had to wire that myself to check it. A ~40-line handler-level test in the `project-files-panel.test.tsx:428` style would pin the wiring, not just the selector. Optional — the guard is two lines and directly readable, and `data-session-grid` is load-bearing in `styles.css:1279-1326`, so it will not drift away silently.
8. **`core-cert-material.ts:562`.** Two notes. The commit undersells itself: the old body could mint a serial the certificate hands back not merely *shorter* but reading *negative* once the pad byte was stripped, which is the RFC 5280 §4.1.2.2 violation, not just a mismatched name. And the remap gives `0x01` four source bytes (`0x00`, `0x80`, `0x01`, `0x81`) against two for every other value — 126.99 bits instead of 127 in the leading byte, of no practical consequence, but the unbiased form in a CA is a redraw, and the 5000-draw test passes either way, so nothing pins the distribution. Record it or ignore it; not worth a push on its own.
9. **Scope.** The `packages/shared` commit is outside #401. It is a genuine latent bug, correctly diagnosed, in its own commit, and it lifts out cleanly — I would keep it rather than leave the flake for the next PR to catch. Your call.

Still a draft — that is deliberate per the body, and marking it ready is yours to do. The shape, the evidence and the merge hygiene are all good, and the one thing round 1 blocked on is genuinely fixed rather than papered over.

https://claude.ai/code/session_01CRhUV36a7JwajTy5Zb3Fyp

Issue #479, opened 2026-09-03
Panel: with a session expanded, a file drag has nowhere visible to land


qcentic-dev:
Follow-up to #401, raised in review of #477 (finding 3.5). Recorded, not fixed there: it is a gap in *feedback*, not a
wrong destination, and #401's acceptance is about where the file lands.

## What the operator sees

Expand a session in grid view, then drag a file in from the desktop meaning to upload it to the Project. The whole
surface below the project header shows a no-drop cursor, and nothing points at where the gesture would actually work.
The operator has no way to tell whether the Panel is refusing the file, is broken, or has not noticed the drag.

## Why

`SessionGrid.tsx` gives the expanded cell `position: absolute; inset: gridPadding` (`:568`), and the shipped Studio
layout runs `gridPad = 0` (`:834`), so the expanded cell covers the entire grid content box. Since #477 the board's
arbiter treats everything inside `[data-grid-cell], [data-session-grid]` as session space and refuses the drop with
`dropEffect: "none"` — correctly, per #401 — so with a cell expanded that refusal covers everything under the header.

The two places the drop *does* work are both small and both unlabelled during a drag: the project header strip, and the
`CardFrame`'s 8px padding around the grid. The Files drawer accepts drops too (#400), but only when it is already open.

## What to do

Give the refusal somewhere to point. Options, cheapest first:

1. During a file drag over a session pane, surface a hint naming the Files drawer — a small persistent affordance
   rather than a full overlay.
2. Open (or highlight) the Files drawer on `dragenter` while a file drag is in flight, so the drop target the operator
   wants is on screen before they release.
3. Give the expanded cell an explicit "upload to Project" strip — the "explicit overlay" branch #401's acceptance
   criterion already allows, scoped to the one configuration that needs it.

Anything that ends with the operator seeing where the file can go closes this.

## Acceptance

1. With a session expanded in grid view, dragging a file over the Panel shows the operator at least one place the drop
   will be accepted, without them having to guess or to open the Files drawer first.
2. The drop still does not land in the Project root by way of a session pane (#401 stays fixed).

## Not this issue

`#477`'s other recorded gap — a menu portalled out of a cell to `document.body` bubbling to the board through the React
tree — is a wrong-destination hole, not a feedback one, and is commented in `board-drop-arbiter.ts` rather than filed.

Issue #478, opened 2026-09-03
Rail slot follow-ups from #382: reconcile the shared numbering, pin the presentation routes, stop degrading filing to empty


qcentic-dev:
Deferred from the review of #476 (the fix for #382). Three items were raised there, agreed as real, and explicitly left out of that PR's scope. They are filed together because they share a subject: the rail's numbering space now has two writers, and nothing yet reconciles it.

Context: #382 made a Core-owned pin's rail slot Panel-local presentation (`project_presentation.pinned_order`), numbered in the same space as `projects.pinned_order` — the row's index in the whole rail. See [ADR 0022, amended by #382](../blob/main/docs/adr/0022-a-core-owned-project-has-a-panel-side-presentation-row.md).

## 1. The shared numbering space has no reconciliation pass

Two writers share one sequence and neither can see the whole of it at write time, so slots drift:

- **Renumbering while a Core is off the rail.** A reorder writes indices for the rows it can see. A Core whose pins are absent from that rail — never read, or read and then pruned from view — keeps whatever slots it last had, so on reconnect its pins can hold indices that now tie with Panel rows. `comparePinnedProjects` then breaks the tie on `createdAt`, which is not an order anybody chose.
- **A stale slot surviving an unpin this Panel never saw.** #476 clears a Core pin's slot when the unpin happens here (`ProjectBar`'s context menu). An unpin performed at another Panel, or at the Core's own keyboard, leaves the slot behind — so unpinning and re-pinning that project there returns it to the slot it used to hold, while a re-pinned Panel row goes to the end of the rail. `pruneProjectPresentation` only sweeps projects the Core has *deleted*, not ones it has merely unpinned.

Both want the same thing: something that periodically reads the rail as a whole — the Panel's pinned rows plus every presentation slot — and renumbers it densely, dropping slots whose project is no longer pinned. The read path already assembles exactly that list once per fan-out (`lib/core-pins-engine.ts`), which is the obvious place to notice a rail that needs it; the write wants to be a single endpoint owning both tables, which would also close the non-atomicity noted below.

Related and worth deciding at the same time: the two halves of a reorder are two requests against two tables and are **not** one transaction. If the Panel's order write commits and the Core half then fails, the rail is half-applied under a failure toast. The client reverts what it can and re-reads the rest, but the honest fix is one endpoint that owns both writes.

## 2. `api-auth.test.ts` does not list the `project-presentation` routes in `PROTECTED_ROUTES`

[...]

Issue #473, opened 2026-09-02
Restored user terminals orphan their pre-reload PTY on the Core


qcentic-dev:
Follow-up recorded from the review of #468 (finding F3), which fixes #394.

## What accumulates

#468 persists a user terminal's identity — `{ scopeKey, coreId, kind, cwd }` — so a reload restores the same shell in the same project on the same Core. It deliberately does **not** persist the `ptyId`: a restored session gets `ptyId: null` and its pane spawns a **new** PTY.

A browser reload does not drop the Panel↔Core link, so the pre-reload PTY is still alive on the Core. Nothing reclaims it and nothing can kill it:

- `findByTask` refuses to return shell / `shellSession` PTYs (`packages/core/src/pty-manager.ts`), so the new pane cannot adopt the old process;
- `killTerminal` in `packages/panel/src/lib/user-terminal-store.tsx` kills only the `ptyId` the store currently holds, which after a restore is the replacement;
- the identity map records no `ptyId`, so nothing else knows the old one.

**Net: every reload leaves one live, unreachable login shell per restored user terminal, on the Core's machine, until the Core (or the machine) restarts.** Before #468 a project terminal was not restored at all, so this is new — the price of restoring the terminal rather than losing it.

## Why #468 did not just fix it

Two options were considered and both are changes of their own, with their own failure modes:

1. **Re-attach to the recorded PTY.** Needs a replay path that fails safe when the pty is gone — a dead persisted `ptyId` would strand the pane behind a Retry button that can only ever fail again.
2. **Kill the recorded PTY when the replacement spawns.** Not obviously safe with more than one tab open: a second tab reloading would kill the shell the first tab is still attached to and still using. Any kill-on-restore needs to know that no other tab holds that pty.

Neither is in #394's "What to do", which asks for project, Core, kind and cwd — all of which #468 persists.

## What would close this

Either of the above, done properly:

- persist the `ptyId` alongside the identity, re-attach when the replay succeeds and spawn fresh when it does not; **or**
- persist it only to kill it on restore, with a guard that no other live tab is attached to that pty.

A cheaper stopgap worth considering on its own: give the operator a way to see and kill the Core's orphaned shell PTYs, so the accumulation is at least recoverable without restarting the Core.

## Refs

- Bug this follows from: #394
- PR that introduced the trade-off, with the reasoning: #468
- Review finding: F3 on #468

PR #481, opened 2026-09-03
fix: pin selection, session settle, file-drop duplication, and CLI send/tail

qcentic-dev:
Feature branch `fix/operator-pins-sessions-drop-cli` → train `beta/0.4.5`.

Tracker: #436. **24 of 24 child issues merged**, 27 commits. Every ticket was
reviewed on its own PR and merged only at green CI.

## What this fixes

Three operator-reported symptoms, plus the lost-session and CLI work filed
alongside them.

**A finished session flips back to running** — #385 #386 #388, with the
Core-side settlement work #387 #389 #390 #391.

**The pin rail does not follow the last click** — #376 #377 #378 #379 #380
#381 #382 #384.

**A dragged file uploads twice** — #400 #401.

**The original session is lost** — #393 #394 #397.

**CLI observe and send** — #402 #403 #404, and #414 (the operator-facing
`Endpoint` label becomes `Address`).

## Cut from this train, deliberately

#277 #290 #383 #392 #395 #396 #398 #399 #405 — filed, not implemented here.

## Known limits, stated rather than hidden

- **#382's second-round fixes were approved by operator override, not
  verified by a reviewer.** Two defects were fixed on the author's word with
  green CI: an unpin that wrote a Panel project id against a foreign Core and
  cost that project its card image on the next prune, and a
  `settleCorePinFiling` overlay that came down without a read having
  happened. Its third review round could not run.
- **#403 stays open** under `Refs`; its second criterion needs #442.
- **#380 stays open**; it landed as hardening, and the symptom itself needs
  the Panel UI exercised by hand.
- **The rail is a set of local guards, not a single authority.** Three
  uncoordinated status writers, a refetch-everything event model with no SSE
  cursor and no invalidation on reconnect, and the same facts held in five
  stores are all untouched by this train.
- `core-session-backstop.test.ts` "never takes back a finish another writer
  wrote over it" failed once on this branch and passed on re-run. Worth
  watching rather than assuming settled.

## Follow-ups filed while building this

#440 #441 #442 #446 #447 #451 #452 #460 #464 #465 #466 #467 #469 #473 #474
#478 #479.


mrfqcentic (approved):
# Gate review — PR #481 into `beta/0.4.5`

**Verdict: approve. The branch is fit for the train.**

Reviewed as a whole, not ticket by ticket: `beta/0.4.5...fix/operator-pins-sessions-drop-cli` at head `1f03a0e` (0 behind / 27 ahead of the train tip `a7b5532`, clean fast-forward shape). **CI at the head: all 20 non-skipped checks pass** (Unit Tests 2m53s, Typecheck, Lint, Conventions, both E2E suites, both image builds, secret/dependency scans); the five skipping jobs are train-only and skip correctly on a feature branch. MERGEABLE; the BLOCKED merge state is the review gate itself, which this review settles.

The two axes, kept separate as asked.

---

## Axis 1 — Standards: does the integrated code hold together

**Yes.** The tickets that collided on the same files agree with each other, not merely fail to conflict:

- `ProjectBar.tsx` carried six pin tickets (#376 #378 #379 #380 #381 #382) plus #401's drop target, and the cross-references are accurate in both directions: #382's reorder still withholds `data-pinned-item` from #378's off-group tile (`:308-315`, `:1256`); #376's ring moved to a subscribing `useRouterState`; #384's write ledger in `active-group.ts` separates "may paint" from "did the server confirm" so a superseded-but-successful PATCH is still recorded. No comment I found asserts something a later ticket made false.
- The coalescing runner from #389 (`lib/fleet-refresh.ts`) is shared by the fleet fan-out and the #377/#382 pins engine deliberately, with a header comment saying so — one copy, not two that drift.
- #385's narrowed heal window (30s → 1s in `shared/subagent-activity.ts`) is documented as a trade with its residual filed as #440, and it composes correctly with #391's backstop: the backstop's reopen is marker-scoped to finishes its own idle rule wrote, so the two never fight.
- Conventional Commit subjects and bodies describe what landed — checked across all 27 commits. PR #459's "do not squash" instruction was honoured (its four commits are on the branch verbatim, keeping the retraction adjacent to the claim it retracts). `Closes` trailers will fire on train→main promotion; `Refs` ones correctly will not.
- Docs moved with the code: ADR 0008 carries an honest correction for #388, ADR 0022 a proper amendment for the fourth presentation field, `harness-status-detection.md` records #387's sweep, CONTEXT.md gains #393's pending-drive sentence, and the skill's `send` examples match #404's new default.
- No dead guards and no leftovers: no TODO/FIXME/debugger in production code. `isProjectScopeVisible` and the `task-status-sync` turn-state exports are test-only seams, documented as such. One pre-existing oddity, not this branch's: `UserTerminalPane.tsx` is binary-flagged in git (717 NUL bytes at the train's base, 728 here) — worth a cleanup issue of its own.

## Axis 2 — Spec: the three operator symptoms, end to end

**A finished session flipping back to running — fixed, at every layer it occurred.** Core: #385 stops post-turn subagent helpers healing the row; #387 sweeps stranded `ready` rows on `pty:spawn` evidence and the PTY-exit settle covers `ready` as `disconnected`; #390 settles a foreign-`session_id` Stop instead of acking and dropping it; #391's backstop reads output through the redraw classifier so an idle TUI no longer starves it, with the bounded hook deference documented. Panel: #386's fallback posts `running` only on a genuinely submitted turn (paste/cancel/recall modelled); #388 replays a bounded, age-bounded finish tail to a fresh tab with two-half dedup; #389's coalesced refresh no longer drops a finish that lands mid-pass. The layers agree: the backstop's `noteTaskFinished` stamps suppress the #385 heal, and its reopen path is the only sanctioned resurrection.

**The pin rail not following the last click — fixed, with the two declared exceptions confirmed open.** #376 (ring subscribes to the route), #378 (group filter keeps the open project's ring), #379 (Cmd-digit keeps `coreId`), #381 (late materialize from a left pin is dropped via scope generations), #384 (stale `setActiveGroup` answers can't paint), #382 (mixed-rail reorder persists through each row's owner). **#380 is genuinely OPEN** — and `0b3b099`'s body is honest that it landed as hardening: "This is a hardening refactor plus a shared reducer, not a cure for the operator symptom… Issue 380 (B1) stays open."

**A dragged file uploading twice — fixed on both paths.** #400 stops the drawer propagating to the board; #401's `dragTargetIsSessionPane` arbiter (`[data-grid-cell]`, `[data-session-grid]` — both attributes verified present in `SessionGrid.tsx`) makes the board refuse drops aimed at the grid with a no-drop cursor. The accepted gap (no drop affordance on a session at all) is filed as #479.

## The operator-override fixes from #382 — verified against this branch, since nobody had

The PR body declares these were never reviewer-verified. I re-drove both at `1f03a0e`:

1. **The unpin that filed a Panel project id against a foreign Core (round-2 blocking 1) — closed.** `ProjectBar.tsx:1593` now reads `coreIdByProject.get(id) ?? null`; the route-`coreId` fallback is gone entirely, and `coreIdByProject` (`:150-153`) is built from the same list the menu target comes from (`:961` via `:273`). A Panel-owned row answers `null`, the guard at `:1609` (`if (unpinning && targetCoreId)`) refuses the presentation write, and `mutateProjectForCore(null, …)` takes the Panel HTTP arm. The bogus `project_presentation` row can no longer be created, so `pruneProjectPresentation` (`queries/index.ts:84-86`) never orphan-deletes it and `deleteAllProjectImagesFor` never sweeps the Panel project's card image. Pinned by "does not file a Panel-owned project under the route's Core" in `project-bar-mixed-rail-reorder.test.tsx:737`.

2. **`settleCorePinFiling` coming down without a read (round-2 blocking 2) — closed.** It now awaits `run.fresh()` (`core-pins-engine.ts`), and `fresh()` in `fleet-refresh.ts` genuinely resolves only after a pass that *started after the call* settles — waiters are claimed at the top of an iteration, the loop condition includes `waiting.length > 0`, and stranded waiters are released on failure rather than hung. Overlay entries are token-keyed so a gesture's settle cannot take down a later gesture's overlay, and the settle republishes from `lastPresentation` so the failure path no longer leaves the two halves of the rail disagreeing. A failed presentation read now holds the last map instead of filing every row null (#478's item 3, taken here as declared). The four new cases in `core-pins-rail-slot.test.tsx` (`:320` mid-pass, `:375` failure path, `:405` two-gesture token, `:446` failed read) target exactly these.

## The declared open issues — both confirmed

- **#403 is OPEN**, and PR #459's body says "Refs #403 — deliberately *not* `Fixes`": criterion 2 (cursor past unseen matching events on interrupt) belongs to **#442, which is OPEN** and correctly out of scope here.
- **#380 is OPEN** as declared (above).

## The one flake — mechanism now understood, and it is not order-dependence

`core-session-backstop.test.ts:472` ("never takes back a finish another writer wrote over it") failed once and passed on re-run. The production mechanism is **sound**: the idle settle records the row's `updatedAt`, and `reopenIfIdleSettled` requires it unchanged — any other writer's write, including a `finished`-over-`finished` Stop, ends the claim (`core-session-backstop.ts`). The test's nondeterminism is a **clock seam, not an ordering bug**: the backstop runs on the test's fake `nowMs`, but `coreMutationStore.mutateTask` stamps `updatedAt` from real `Date.now()` (`core-mutation-store.ts:143`). The settle write and the test's foreign finish write land microseconds apart in real time; when they share a millisecond the marker spuriously matches and the row reopens. Rare, fails safe (a re-run), and honestly declared in the PR body — but **no follow-up issue covers it** (#446 is a different flake), and the "worth watching" note deserves a ticket: inject the clock into the mutation store in this suite, or stamp the foreign write with a guaranteed-distinct `updatedAt`. Not a gate condition; asking that it be filed.

---

24 child PRs ↔ 24 tracker issues ↔ 27 commits, all accounted for; every declared limit in the body (#380, #403/#442, the rail remaining a set of local guards, the flake) checks out as stated. The record is honest, including where it cost something.

**Approve.**

Issue #269, opened 2026-08-19
`.husky/pre-push` does not carry `.github/workflows/ci.yml:167`'s `beta/*` exemption — the hook refuses a train branch CI accepts


qcentic-dev:
## The drift

The two branch-name gates in this repository have never agreed. CI knows the `beta/x.y.z` train class; the local
`pre-push` hook does not, and refuses to push one.

**`.husky/pre-push:9`** — the only pattern the hook has, with no `beta/*` exemption anywhere in the file:

```sh
pattern='^(main|master|develop|(feat|feature|fix|bugfix|hotfix|release|chore|docs|refactor|perf|test|ci|revert)\/[a-z0-9]+([._-][a-z0-9]+)*)$'
```

**`.github/workflows/ci.yml:161-170`** — the exemption the hook is missing, added precisely so a train branch is not
rejected by the naming convention (ADR 0023 D1):

```yaml
# The train (ADR 0023 D1, D3). A promotion pull request's head *is*
# `beta/x.y.z`, which is a branch class the convention above does not
# know about — without this every promotion fails a required check
# and the train can never reach `main`. [...]
if echo "$BRANCH" | grep -qE '^beta/[0-9]+\.[0-9]+\.[0-9]+$'; then
  echo "✅ Branch name '$BRANCH' is a release train (ADR 0023 D1)."
  exit 0
fi
```

The hook also has no `dependabot|renovate` exemption, which `ci.yml:155-158` carries for the same reason. This issue
is about the `beta/*` half; the bot half is the same shape of drift and can ride along in a fix.

## How long

`.husky/pre-push` has not been touched since the initial commit, `c2b8d0d` (2026-08-03). The `beta/*` exemption
entered `ci.yml` with `45219c3` (2026-08-11, #136, "release trains and digest promotion"). So the two have silently
disagreed about the `beta/*` class for as long as that class has existed — every day since trains were introduced.
Nothing surfaced it because, until now, no document told anyone to push a `beta/*` branch from a local clone.

## Reproduction

Run the hook verbatim, on a branch named like a train:

```
$ git switch -c beta/0.3.3
$ sh .husky/pre-push
❌ Branch name 'beta/0.3.3' violates the naming convention.
   Expected: <type>/<kebab-case-description>
   Types: feat|fix|hotfix|release|chore|docs|refactor|perf|test|ci|revert
   Example: feat/proj-123-oauth-device-flow
   Rename with: git branch -m <new-name>
exit 1
```

[...]

PR #482, opened 2026-09-03
ci: trains may carry a -fN sub-beta suffix

qcentic-dev:
A beta train sometimes needs a small fix taken before it promotes, and the
freeze window ([ADR 0023](docs/adr/0023-release-trains-and-digest-promotion.md)
D2) is exactly when re-cutting is most expensive. This adds **sub-beta trains**
— `beta/0.4.5-f1`, `beta/0.4.5-f2` — as a branch class the model knows about,
and refuses them everywhere a promotion starts.

## Why `-fN` and not `.N`

A train's name is the string every downstream surface derives from, and
`0.4.5.1` is not semver. npm refuses it in a manifest, `assert-version-agreement`
reports it as *not a version this repository publishes*, and the publish
rehearsal cannot pack it — so a train named that would break after the cut, on
every push. `0.4.5-f1` is a legal prerelease of the same line.

**The widening is exactly `-f` followed by digits**, in every guard. `-rc.N` is
a backport candidate (D30) and `-beta` is a published beta (ADR 0036 C1);
widening to arbitrary prereleases would collide with both. `scripts/__tests__/train-name.test.mjs`
holds each guard to that suffix and to nothing looser, and a separate assertion
in `workflows.test.mjs` enumerates every train-name pattern in the workflows and
refuses one this ADR does not define.

## The decision about `promote.yml`: it **refuses**, and does not strip

The task offered two answers. Reading the workflow end to end, refusing is not
just the operator's intent — stripping is incoherent, and this is the part worth
being sure of:

- `resolve` derives `version="${TRAIN#beta/}"`, and that version becomes the
  `vx.y.z` tag, the `release/x.y` line and the string `release.yml` publishes.
  **Widening the regex alone would fast-forward `main` and push `v0.4.5-f1`** —
  exactly the outcome the requirement rules out.
- **Stripping is worse than wrong.** D16 — the clause the whole design reduces
  to — resolves `beta-$version` to a digest and asserts the image's
  `org.opencontainers.image.revision` equals the promotion pull request's head
  SHA. A sub-beta publishes **its own** moving tag `beta-0.4.5-f1` (see below),
  so a stripped promotion of `beta/0.4.5-f1` would hold the **plain train's**
  `beta-0.4.5` image against the **sub-beta's** head SHA. That either fails with
  a message about a train that moved, or — if the plain train happens to sit at
  that commit — promotes bytes built from another branch. Neither is a thing to
  ship.

So the sub-beta merges back into its plain train, and the plain train promotes.
The refusal is in **three** places, because a person can arrive at any one of
them first, and all three say the same next step:

| where | what it does |
| --- | --- |
| `ci.yml` `Train rules` | fails a `beta/x.y.z-fN → main` pull request **by name**, not by falling through to "not a train branch" — it *is* a train branch, and the author is right that it carries the fixes |
| `promote.yml` `pause` | fails the dispatch by name, with the merge-back and the reason the suffix is not stripped |
| `scripts/promotion-gate.sh` | **stays red** — waving it through would satisfy `docs/rulesets/main.json`'s required context and re-enable the merge button on a gate — and prints the merge-back instead of a dispatch `promote.yml` would refuse |

## The suffix names a branch, never a tree

This is what makes the change cheap, and it is the other decision in here.

The six manifests and `install.sh`'s stamp on `beta/0.4.5-f1` say **`0.4.5`**,
exactly as they do on `beta/0.4.5`. D3 is untouched; cutting a sub-beta writes
no version at all. Three things follow:

- **The merge back is a diff of the fixes and nothing else.** There is no
  version to revert, so the merge-back cannot strand the plain train on a
  version it must not publish.
- **The release is `v0.4.5`.** Every published string is derived from the line,
  and the line is what the tree carries. `beta-release.yml` takes the suffix off
  before it composes anything, so **ADR 0036 C1's assertion is not widened** —
  the beta string is still one concatenation from a bare `x.y.z`, and
  `^[0-9]+\.[0-9]+\.[0-9]+-beta$` is unchanged.
- **`scripts/lib/version-agreement.mjs` already accepts it.** `lineOf("0.4.5-f1")`
  is `0.4.5` and `channelOf` calls it a prerelease, so `Train versions` and
  `promote.yml`'s tree assertion compare the manifests against the line without
  knowing this clause exists. **ADR 0037's vocabulary is unamended.**

`Train rules` therefore asserts `assert_versions "$line"` under a `beta/*` base,
not `"$version"` — asserting the branch name would demand a version no cut
writes and fail every pull request into a sub-beta.

**The one place the suffix survives is the moving image tag.** `beta/0.4.5-f1`
publishes `beta-0.4.5-f1`, not `beta-0.4.5`: two branches of one line are two
commits, and a shared tag would mean the digest D16 verifies is whichever pushed
last — D12's argument for the per-build stage, reaching one tag further.

## Everything else that reads a train name

The task named three lines; the sweep found six more places, and two needed a
decision rather than a regex:

- **`ci.yml` `train-rules`, base `main`** — refuses a sub-beta head by name.
- **`ci.yml` `train-rules`, base `beta/*`** — accepts a sub-beta base, asserts
  the **line**, and allows the one new train-headed pull request the model has:
  a sub-beta merging back into **its own** train. Matched against that base, so
  a sub-beta of another line, or a sub-beta of a sub-beta, is still refused.
- **`ci.yml` `train-tags`** — widened, and deliberately keeps the suffix.
- **`promote.yml` `resolve`'s survivor list** — **drops sub-betas of the train
  being promoted.** D23 observes the hotfix condition as *another train exists*
  and D24 rebases and force-pushes what it finds; a sub-beta counted there would
  make every ordinary promotion force-push over a branch whose work is already
  on `main`, and, with a genuine surviving train beside it, trip the
  "more than one surviving train" refusal on a state the model *does* have an
  answer for. Sub-betas of *another* line are not filtered — those are a
  different train and D2 applies in full.
- **`beta-release.yml`** — gains a `line` output; the CLI pack and the Release
  body take the line, `source_tag` keeps the branch.
- **`scripts/promotion-gate.sh`** — above.

Checked and **not** changed, with the reason:

- **`scripts/lib/version-agreement.mjs`, `assert-version-agreement.mjs`,
  `container-image.yml`** — already correct through `lineOf`. Tests added to pin
  that, because a regression there would strand a sub-beta's every push.
- **`scripts/rehearse-npm-publish.mjs`, `scripts/lib/npm-packages.mjs`** — take
  a bare line and always did; `beta-release.yml` now hands them one.
- **`housekeeping.yml`** — globs `beta/*` with no shape parsing. A sub-beta is
  somewhere a pull request can go, which is the invariant that chore keeps.
- **`.husky/pre-push`** — does not know the `beta/*` class **at all**, so it
  refuses a plain train as readily as a sub-beta. That is
  [#269](https://github.com/actana/control/issues/269), a separate defect with
  its own ticket, and fixing it here would be scope this pull request did not
  ask for. `docs/ci-cd.md` already carries the `--no-verify` note; the new
  runbook section points at it.

## Tests

`scripts/__tests__/train-name.test.mjs` is new and **executes the real bash**,
pulled out of the workflow files — the way `promotion-gate.test.mjs` runs the
real gate script. Grepping the YAML for a widened pattern would assert almost
nothing; these run the guard.

Twenty-one cases across five guards, covering the four behaviours asked for:
a sub-beta is accepted by the convention check (at one digit and at two); a
sub-beta is refused by promote, with the merge-back named; `beta/0.4.5.1` is
still rejected everywhere, along with `-rc.1`, `-beta`, `-f` and `-fx`; and a
plain train behaves exactly as before, including the strings that reach a
registry.

**Verified by mutation.** Making `promote.yml` strip the suffix reddens three
of them; widening the convention to `-[0-9A-Za-z.-]+` reddens a fourth.

Also: `promotion-gate.test.mjs` (a sub-beta gate stays red, prints the
merge-back, and the plain train's dispatch is unmoved), `version-agreement.test.mjs`
(the line reading of `0.4.5-f1`, and that a fourth dot has no reading at all),
and `workflows.test.mjs` (the two `Train rules` decisions, which call `gh` and
cannot be executed here, plus the pattern enumeration).

`pnpm vitest run scripts/__tests__/` — **786 passed, 24 files**. `eslint scripts`
clean. All seven workflow files parse.

## Record

**ADR 0023 gains section H and D46**, and section A gains an amendment note so
D1, D2 and D3 read one clause wider without their arguments moving.
`docs/ci-cd.md` gains the procedure a person follows, beside the cut and the
promotion; `CONTRIBUTING.md` says the open train sometimes carries the suffix.

## Not done, on purpose

No merge, nothing on `main`, no train branch cut, and **no `package.json`
version touched** — which is also the model: a sub-beta writes no version.

---

Base is `beta/0.4.5`, per the train model. Three commits, Conventional Commits,
commitlint run locally over all of them (`--from 9eb139e --to HEAD`, 0 problems)
and over this title.


mrfqcentic (approved):
Approved on the operator's direct instruction — release-tooling change, reviewed by CI only, no reviewer pass.

Widens the three train-name guards to accept an optional -fN sub-beta suffix, refuses (does not strip) a -fN train in promote.yml so only a plain beta/x.y.z can publish vx.y.z, and keeps a fourth-dot name like beta/0.4.5.1 rejected because it is not valid semver and the branch name is the manifest version.

Issue #484, opened 2026-09-03
Session that finishes while the operator is away still reads running on return


mrfqcentic:
## Summary

A Session that finishes while the operator is looking at another page still
reads as **`running`** when they navigate back. Only a full browser refresh
corrects it. The Session is long since `finished` on the Core.

## Observed in the field (beta/0.4.5, 2026-09-03)

Start a Session, navigate away from its project while it is still running, let
it finish, then navigate back. The Session row and detail both show `running`.
`actana session ls` from the CLI shows `finished` for the same Session at the
same moment, so the Core state is correct and the Panel's copy is stale.

## Expected behaviour

Returning to a page shows the Session's current state without a browser refresh.
Any events missed while the page was not mounted are reconciled on return.

## Acceptance criteria

- [ ] Navigating away from a project while a Session runs, letting it finish,
      and navigating back shows `finished` with no browser refresh.
- [ ] The same holds after the tab has been backgrounded and refocused.
- [ ] The same holds across an SSE reconnect: events that occurred during the
      gap are not silently lost.
- [ ] The root cause is named in the PR description with the enforcing
      `file:line`, not guessed at.
- [ ] Regression coverage for the return path.

## Landmines — verify against `beta/0.4.5-f1`, do not assume

Three suspects, in order of likelihood. **Confirm which one it actually is by
reading the code before changing anything** — more than one may contribute:

1. `router.tsx` — `staleTime: 30_000` combined with
   `refetchOnWindowFocus: false`. A remount inside the stale window serves the
   cached, pre-finish value and never refetches.
2. `lib/use-events.ts` — the SSE channel has **no cursor, no replay, no
   `Last-Event-ID`, and nothing that invalidates queries on reconnect**. Any
   event that fires during a disconnect gap is lost outright.
3. The scope-cancel guard from **#381** (landed in 0.4.5), which cancels
   in-flight project reads on leave. Check whether the return path actually
   refetches after that cancel.

This is symptom **W2** of `panel-rendering-investigation-2026-09-02.md`.

## Scope

`packages/panel`. Does not touch `packages/core`.

mrfqcentic:
Shipped in v0.4.5 (https://github.com/actana/control/releases/tag/v0.4.5): merged via the beta/0.4.5-f1 gate #495, promoted to main at d7eeb65 through #529.

Issue #483, opened 2026-09-03
Prompt delivery types blind at the deadline and reports a false delivered


mrfqcentic:
## Summary

The prompt-delivery backstop in `packages/core/src/harness-prompt-delivery.ts`
gives up waiting for a harness composer after a fixed deadline, **types blind**,
and then reports the delivery as `delivered`. For `opencode`, whose boot lands on
either side of that deadline, this loses the prompt outright — and the caller is
told it succeeded.

## Observed in the field (beta/0.4.5, developer Core, 2026-09-03)

Five opencode Sessions started on the same Core, same project, same harness,
within half an hour:

- **three** sat at `ready` with an **empty composer** — the prompt was typed
  before opencode was listening and was discarded by the terminal;
- **two** reached `finished` with the prompt visibly **retyped two or three
  times** in the transcript before one submission took.

Same inputs, different outcomes, minutes apart. This is a race against
opencode's boot time, not a broken harness.

## Why #229 and #232 do not cover this

Both are closed and both shipped the code running today: #229 added the
`HARNESS_READINESS` table, #232 added the composer marker. **Neither one says
what happens when the composer marker arrives *after* the deadline.** The
current answer — type anyway, report success — is the defect.

## Expected behaviour

For a harness that has a **known composer marker**, delivery waits for the
marker, not for a clock:

1. Never type into a composer that has not been observed ready. A blind type at
   the deadline must not happen for a harness with a known marker.
2. If the marker never arrives within a generous per-harness ceiling, the
   delivery ends as **`abandoned`**, and the Session becomes **`needs-input`**.
   A false `delivered` is the bug; an honest failure is the fix.
3. The failure is visible to the client: the CLI and the Panel must both be able
   to tell the caller the prompt did not land.

## Acceptance criteria

- [ ] A harness with a known composer marker never receives a blind type at the
      generic deadline; the wait is keyed on the marker.
- [ ] A delivery whose marker never appears resolves as `abandoned` and puts the
      Session in `needs-input`, never `delivered`.
- [ ] A per-harness wait ceiling exists in the override table in
      `packages/core/src/harness-prompt-delivery.ts`, with `opencode` set to
      approximately 90s (the operator's stopgap value).
- [ ] The generic backstop for harnesses **without** a known marker is
      unchanged — this is not a global timeout increase.
- [ ] Unit coverage for all three paths: marker before deadline, marker after
      deadline but within the ceiling, marker never.
[...]

Issue #396, opened 2026-09-02
C4: session wait / waitForTurnEnd never unblocks if the link drops


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: yes** — pin/session/CLI/drop wave. Do not start until asked.

- Hunt ID: `C4`
- Severity: broken
- First-fix owner: cli / sdk
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

`actana session start proj "prompt" --wait` with no `--wait-timeout`. Core blip mid-wait: Promise stays pending.

## Where

`core-session.ts` 867–892 no disconnect handler; `core-client.ts` 1009–1012 `eventsSubscribed=true` before sendNow; `onCoreEvent` drops null taskId.

## What to do

Disconnect must reject or retry the waiter. A default timeout is acceptable if disconnect detection is also honest.

## Acceptance

1. If the core-link drops during `--wait`, the command exits non-zero (or reconnects and still exits) rather than hanging forever.

mrfqcentic:
Shipped in v0.4.5 (https://github.com/actana/control/releases/tag/v0.4.5): merged via the beta/0.4.5-f1 gate #495, promoted to main at d7eeb65 through #529.

PR #486, opened 2026-09-03
fix: send --wait fails fast rather than waiting for a turn it never started

qcentic-dev:
`actana session send $SID "text" --wait` could wait for ever. The wait counts from the id the Core stamps a delivery with and resolves on the first settling status **after** that id, so a send that starts no turn has nothing that will ever end it:

- `--no-enter` sends no carriage return, so no turn starts;
- a carriage return that lands on a **dialog** rather than a composer submits nothing either.

In both cases the status the Session is parked at was seeded from the Task row and carries **event id 0**, which no real cursor can be satisfied by. That comparison is correct — it is what stops the wait reporting last turn's answer as this turn's — and it is silent and permanent, which is the bug.

## What changed

**1. `--no-enter` with `--wait` is a usage error** (`sessionSend`, `packages/cli/src/session-command.ts`). Refused before a byte is written: a send that submits nothing has no turn to await, and that wait would either never end (idle Session) or end on a turn some earlier send started and report it as this one's (busy Session). One hangs, one lies; neither is a reading the command can pick between.

**2. `send --wait` carries a default deadline of 1020 seconds.** Only that verb. `session wait` and `session start --wait` wait for a turn that is already under way, so a default there could only cut short honest work — they keep "no deadline unless you set one". Seventeen minutes is chosen against the Core rather than from taste: its quiet backstop settles a silent `running` Session at fifteen minutes and sweeps once a minute, so a deadline at fifteen would tie with it and scheduling would pick the winner. At seventeen the Core's mechanism answers first with a real status, and this deadline only ever fires where the Core has none — #405's case, since the backstop skips a Session that is not `running`. `--wait-timeout <s>` replaces it; **`--wait-timeout 0`** removes it (new spelling, accepted on every verb that takes the flag).

**3. The deadline is loud instead of bare.** `waitForTurnEnd` now rejects with a new `CoreSessionTurnTimeoutError` carrying `taskId`, `timeoutMs`, `afterEventId`, `lastStatus` and `reportedSinceDelivery` — the last being whether the Core reported **anything** about the Session after the delivery. That is two event ids compared, not a reading of the screen: nothing is inferred from the byte stream (#191, ADR 0026 D3), and no path consults `reportsTurnStart` (ADR 0033 D2). When it is `false` the message says no turn end was reported and names **both** readings — a return that submitted nothing, or a harness that reports nothing until a turn ends — because `codex` and `cursor-cli` produce that same silence while working; it also says the text was delivered so it is not sent twice. The CLI adds the next step, keeping CLI advice out of a library error: `session logs` to read the screen, `events tail --since <delivery id>` to keep waiting, and an explicit warning **off** `session wait`, which is uncursored and would answer at once with the pre-send status and exit zero.

**4. ADR 0033 gains D5**, amending D4 for the one wait that is not waiting for a turn already running. `docs/README.md`'s index row says so, and the skill payload was regenerated from its authored source.

## Acceptance criteria (issue #405)

1. **`send --wait` without a submitted turn exits with an error rather than hanging** — `--no-enter --wait` is `EXIT_USAGE` before any dial; a submitted return that a dialog eats is bounded by the default deadline and reported as this side giving up, never as a status.
2. **`send --enter --wait` unblocks on turn end when the harness settles** — pinned at both layers: the SDK test writes the text, writes the return, waits from the **later** stamp and resolves when the status lands; the CLI test asserts the wait path still sends the return.

## Files this touches, for the tickets landing beside it

- `packages/sdk/src/core-session.ts` — **the only functions changed are `waitForTurnEnd`** (rejects with the new error) plus the new exported `CoreSessionTurnTimeoutError` and its private `turnTimeoutMessage` helper. `settledSince`, `noteStatus`, `releaseWaiters`, `attach`, `deliver`, `dispose` and `waitForIdle` are untouched, and **no disconnect handler was added to `waitForTurnEnd`** — that is #396, which lands in this file after this PR.
- `packages/core/src/harness-prompt-delivery.ts` — **not touched** (#483 owns it this wave).
- `packages/panel/**` — **not touched** (#484 owns it this wave).
- `packages/cli/src/exit-codes.ts` — unchanged; the timeout reuses `EXIT_FAILURE` and the refusal reuses `EXIT_USAGE`.

## Verification

`pnpm -r typecheck`, `pnpm lint` (0 errors), `pnpm scan:secrets`, and the root, `sdk`, `shared`, `cli` and `core` unit suites all pass locally. Panel is untouched and left to CI.

Refs #405

https://claude.ai/code/session_01HKAxkx64bneadDvmQiwwKz

### DIFF
--- a/.agents/skills/actana-sessions/SKILL.md
+++ b/.agents/skills/actana-sessions/SKILL.md
@@ -97,10 +97,10 @@ a file**, and the contract below is how you get one.
 
 The object also carries `taskId` (the Session's id — this is what every other
 verb takes), `harness`, `project`, `status`, `exited`, and `reportsTurnStart`.
-`--wait-timeout <seconds>` bounds the wait; without it the wait has no deadline
-of its own. A timeout is this side giving up, not a verdict about the Session:
-the Session is still running on the Core and can still be listed, read and
-stopped.
+`--wait-timeout <seconds>` bounds the wait; without it `start --wait` has no
+deadline of its own (`send --wait` does — see the loop below). A timeout is this
+side giving up, not a verdict about the Session: the Session is still running on
+the Core and can still be listed, read and stopped.
 
 While a Session is alive — which is what the `live` field on its
 `actana session ls --json` row tells you, and the only thing that does — you can
@@ -153,17 +153,25 @@ Three things to know before you build on it:
    the Session to settle *first*, then send.
 2. **A `send` presses Enter for you, and `--no-enter` is how you stop it.** The
    text goes first and the carriage return follows as its own separate write.
-   `--no-enter` sends no return, so that send starts no turn — and a `--wait`
-   after it is not waiting for your text: on an idle Session nothing ends, and
-   on a busy one it resolves on the turn that was *already* running. Both are
-   #405's. `--enter` is still accepted and does nothing on a send that carries
-   text, so an older script that passes it keeps working; on a send with no text
-   it still means a bare carriage return and nothing else.
+   `--no-enter` sends no return, so that send starts no turn — and it **cannot
+   be combined with `--wait`**, which is refused as a usage error: a send that
+   submits nothing has no turn to await. Type with `--no-enter`, then
+   `actana session wait` once a turn is actually running. `--enter` is still
+   accepted and does nothing on a send that carries text, so an older script
+   that passes it keeps working; on a send with no text it still means a bare
+   carriage return and nothing else.
 3. **A timeout is this side giving up, not a status.** `--wait-timeout
-   <seconds>` bounds the wait; without it there is no deadline, because a turn
-   takes as long as the work takes. On expiry you get a message saying the wait
-   gave up and a non-zero exit — the Session is still running on the Core, and
-   `session logs`, another `session wait` and `session kill` all still work.
+   <seconds>` bounds the wait. `session wait` has no deadline unless you set
+   one, because a turn takes as long as the work takes; **`send --wait` defaults
+   to 900 seconds**, because it is the one wait for a turn that has not started
+   yet and a carriage return that lands on a dialog rather than a composer
+   starts none at all. `--wait-timeout 0` waits with no deadline. On expiry you
+   get a message saying the wait gave up and a non-zero exit — the Session is
+   still running on the Core, and `session logs`, another `session wait` and
+   `session kill` all still work. When the wait gave up having heard nothing at
+   all about the Session since your text went in, the message says so: no turn
+   ended and none was seen to start, and the text was delivered — read the
+   screen with `session logs` rather than sending it again.
 
 **Every turn of that loop gets its own report file**, and the turn number in the
 filename is what keeps them apart. See below.
--- a/.agents/skills/actana-sessions/SKILL.md
+++ b/.agents/skills/actana-sessions/SKILL.md
@@ -125,10 +125,15 @@ A Session is a conversation, not a single question. The whole loop:
 
 ```bash
 ID=$(actana session start <project> "<first prompt>")   # prints the id, exits
-actana session wait "$ID" --json --wait-timeout 900     # block until it settles
-actana session send "$ID" "<follow-up>" --wait --json --wait-timeout 900
-actana session send "$ID" "<next>" --wait --json
+actana session wait "$ID" --json --wait-timeout 1800    # block until it settles
+actana session send "$ID" "<follow-up>" --wait --json --wait-timeout 1800
+actana session send "$ID" "<next>" --wait --json --wait-timeout 1800
 ```
+
+Every line carries the same budget on purpose: 1800 seconds is what `await.sh`
+gives a round, and a loop whose steps disagree about how long a turn may take
+gives up in the middle of one. Pass it explicitly rather than leaning on a
+default — `send --wait` has one (1020 seconds) and the other two do not.
 
 **`actana session wait <id>` blocks until the Core reports the Session settled**
 — `finished`, `needs-input`, `interrupted`, `terminated` or `disconnected`,
@@ -144,7 +149,7 @@ object** `start --wait --json` prints — same keys, same `screen` — so one pa
 reads all three verbs. Two of those keys are `null` on a Session you attached to
 rather than started: `command` and `reportsTurnStart` are answers to a spawn.
 
-Three things to know before you build on it:
+Four things to know before you build on it:
 
 1. **Sending into a turn that is already running resolves on *that* turn's
    end.** A keystroke into a busy Harness is not a new turn. If the Session was
@@ -163,15 +168,22 @@ Three things to know before you build on it:
 3. **A timeout is this side giving up, not a status.** `--wait-timeout
    <seconds>` bounds the wait. `session wait` has no deadline unless you set
    one, because a turn takes as long as the work takes; **`send --wait` defaults
-   to 900 seconds**, because it is the one wait for a turn that has not started
+   to 1020 seconds**, because it is the one wait for a turn that has not started
    yet and a carriage return that lands on a dialog rather than a composer
    starts none at all. `--wait-timeout 0` waits with no deadline. On expiry you
    get a message saying the wait gave up and a non-zero exit — the Session is
-   still running on the Core, and `session logs`, another `session wait` and
-   `session kill` all still work. When the wait gave up having heard nothing at
-   all about the Session since your text went in, the message says so: no turn
-   ended and none was seen to start, and the text was delivered — read the
-   screen with `session logs` rather than sending it again.
+   still running on the Core, and `session logs` and `session kill` still work.
+   When no turn end was reported since your text went in, the message says so
+   and names both readings: the text may have started no turn, or a turn may
+   still be running on a Harness that reports nothing until it ends. The text
+   was delivered either way — read the screen with `session logs` rather than
+   sending it again.
+4. **Do not answer that timeout with `session wait`.** It is uncursored: it
+   answers from the status the Session is parked at, so on a turn that never
+   started it returns **at once, with the status from before your text, and
+   exits zero** — which reads as a completed turn and is not one. To carry on
+   waiting, follow the log from the delivery instead:
+   `actana events tail --since <event id>`, the id the timeout message names.
 
 **Every turn of that loop gets its own report file**, and the turn number in the
 filename is what keeps them apart. See below.
--- a/docs/README.md
+++ b/docs/README.md
@@ -94,7 +94,7 @@ routing around it.
 | [0030](adr/0030-the-panel-is-a-dumb-pipe-for-file-bytes.md) | The Panel is a dumb pipe for file bytes, and it is the end that holds the credentials |
 | [0031](adr/0031-the-product-ships-one-skill.md) | The product ships one skill, and installs it into the operator's home — **PROPOSED**, amends 0006 |
 | [0032](adr/0032-one-actana-cli.md) | There is one `actana`, and it is both the Core manager and the client — supersedes #265 §6 and 0031 D8's two-binaries premise |
-| [0033](adr/0033-turn-end-is-the-one-mandatory-harness-signal.md) | Turn-end reporting is the one mandatory harness signal; turn-start, auto-mode flags and resume ids may never gate a feature |
+| [0033](adr/0033-turn-end-is-the-one-mandatory-harness-signal.md) | Turn-end reporting is the one mandatory harness signal; turn-start, auto-mode flags and resume ids may never gate a feature — D5 (#405) gives `send --wait` a default deadline, amending D4 |
 | [0035](adr/0035-a-second-skill-for-the-sub-agent-role.md) | A second skill for the sub-agent role, and it is eager — **PROPOSED**, depends on 0031 |
 | [0036](adr/0036-the-beta-release-channel.md) | The beta release channel: a beta installs like a release, and the ref is the channel — **PROPOSED**, amends 0016 (D28, D29, D34, D35) and 0023 (D8, D9) |
 | [0037](adr/0037-one-version-per-line.md) | One version per line: where a version string is written, and what is authoritative — **PROPOSED**, amends 0023 (D3, D7), the two clauses 0036 §G left open |
--- a/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
+++ b/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
@@ -53,6 +53,29 @@ stream**. #191 deleted a flat quiet-output timer that did exactly that and
 wait with no `--wait-timeout` has no deadline; one with a deadline reports, on
 expiry, that *this side gave up* — never a status the Core did not send.
 
+**D5 — `send --wait` carries a default deadline; the other waits do not**
+(amends D4, #405). `session wait` and `session start --wait` wait for a turn
+that is already under way, and the only thing a default deadline could cut short
+there is honest work — so they keep D4's "no deadline unless the caller sets
+one" exactly. `send --wait` is the one wait for a turn *it* has to start, and a
+carriage return that lands on a dialog rather than a composer starts none: the
+Core then has nothing to report, the status the Session is parked at was seeded
+from the Task row and carries event id 0, and a delivery cursor can never be
+satisfied by it. The wait is correct, silent and permanent. **900 seconds** is
+the default, matching the bound the orchestration skill already tells callers to
+pass, and `--wait-timeout 0` removes it.
+
+This does not weaken D4 or reintroduce what #191 deleted. Nothing new is
+inferred: the deadline is the caller's own clock, and what it reports on expiry
+is still that this side gave up. What it adds is a **fact off the event log** in
+the message — whether the Core reported anything at all about the Session after
+the delivery event — so an operator can tell a slow harness from a return that
+was swallowed. That comparison is two event ids, not a reading of the screen.
+
+A send that submits nothing at all (`--no-enter`) is refused with `--wait`
+rather than bounded: it has no turn to await, so there is nothing for a deadline
+to be about.
+
 ## Consequences
 
 Adding a family is still adding a row, and now the row has a bar to clear. A
--- a/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
+++ b/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
@@ -61,16 +61,32 @@ one" exactly. `send --wait` is the one wait for a turn *it* has to start, and a
 carriage return that lands on a dialog rather than a composer starts none: the
 Core then has nothing to report, the status the Session is parked at was seeded
 from the Task row and carries event id 0, and a delivery cursor can never be
-satisfied by it. The wait is correct, silent and permanent. **900 seconds** is
-the default, matching the bound the orchestration skill already tells callers to
-pass, and `--wait-timeout 0` removes it.
+satisfied by it. The wait is correct, silent and permanent. **1020 seconds** is
+the default and `--wait-timeout 0` removes it. Seventeen minutes rather than
+fifteen so that it cannot tie with the Core's own quiet backstop
+(`QUIET_SETTLE_MS`, fifteen minutes, swept once a minute): where the Core has an
+answer it should be the one to give it, and this deadline should only ever fire
+where the Core has none — which is #405's case exactly, because the backstop
+skips a Session that is not `running`.
 
 This does not weaken D4 or reintroduce what #191 deleted. Nothing new is
 inferred: the deadline is the caller's own clock, and what it reports on expiry
 is still that this side gave up. What it adds is a **fact off the event log** in
-the message — whether the Core reported anything at all about the Session after
-the delivery event — so an operator can tell a slow harness from a return that
-was swallowed. That comparison is two event ids, not a reading of the screen.
+the message — whether a status was reported for the Session at an event id above
+the delivery's — so the operator is told which question is still open. It does
+**not** claim to have told a swallowed return apart from a harness that reports
+nothing mid-turn: half the families in `HOOK_FAMILIES` look identical in that
+silence, and separating them would mean consulting `reportsTurnStart` (D2 forbids
+it) or reading the byte stream (D4 forbids it). The message names both readings
+and sends the reader to the screen. That comparison is two event ids, not a
+reading of the screen.
+
+The remedy the message offers is cursored for the same reason. `session wait` is
+**not** it: that verb answers from the status a Session is parked at, so on a
+turn that never started it returns at once with the status from before the write
+and exits zero — a false completion, which is the failure #405 exists to remove.
+`events tail --since <delivery id>` follows the log from the write and can only
+report what came after it.
 
 A send that submits nothing at all (`--no-enter`) is refused with `--wait`
 rather than bounded: it has no turn to await, so there is nothing for a deadline
--- a/packages/cli/src/__tests__/session-command.test.ts
+++ b/packages/cli/src/__tests__/session-command.test.ts
@@ -742,8 +742,11 @@ describe("actana session send", () => {
     expect(calls).toEqual([{ text: "carry on", enter: true }]);
     // And it carries a deadline the operator did not have to know to ask for
     // (#405): the return can land on a dialog rather than a composer, in which
-    // case no turn starts and nothing will ever end this wait.
-    expect(deadlines).toEqual([900_000]);
+    // case no turn starts and nothing will ever end this wait. Seventeen
+    // minutes, which is past the Core's own fifteen-minute quiet backstop and
+    // the minute its sweep can add — where the Core has an answer it gets to
+    // give it, and this only fires where it has none.
+    expect(deadlines).toEqual([1_020_000]);
   });
 
   it("refuses --no-enter with --wait rather than waiting for a turn it did not start", async () => {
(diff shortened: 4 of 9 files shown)


qcentic-dev:
## Acceptance criteria — issue #405

Issue #405's `## Acceptance` section is two items; the `## What to do` paragraph adds two more statements of intent, and the landmines named in the assignment are checked alongside them so nothing is carried silently.

### Acceptance

- [x] **1. `send --wait` without a submitted turn exits with an error rather than hanging.**
  `--no-enter` with `--wait` is refused as a usage error in `sessionSend` (`packages/cli/src/session-command.ts`) before the gateway is dialled — `EXIT_USAGE`, message `--no-enter starts no turn, so --wait would have nothing to wait for …`, naming both ways out. Test: *"refuses --no-enter with --wait rather than waiting for a turn it did not start"* (the fixture gateway throws if anything dials it, which is what proves nothing was written).

- [x] **2. `send --enter --wait` unblocks on turn end when the harness settles.**
  Pinned at both layers. SDK: *"unblocks on the turn a delivered carriage return started"* — the text is written, the return is written as its own write, the wait counts from the **later** stamp, and the settling status resolves it. CLI: *"submits a follow-up sent with --wait, and bounds the wait it did not start"* asserts the wait path still passes `enter: true`. The existing #289 tests for the five settled statuses, the exit route and the `reportsTurnStart: false` families continue to pass unchanged.

### From `## What to do`

- [x] **`--wait` without a submitted turn must fail fast.** As (1): a usage error, before any I/O — not a bounded wait that eventually gives up.

- [x] **Hitting a dialog must not hang.** A return the harness routes to a dialog *is* submitted, so it cannot be refused up front — it is bounded instead. `send --wait` now carries a default deadline of 900 seconds (`--wait-timeout` replaces it, `--wait-timeout 0` removes it), and expiry is reported as this side giving up with a named error, never as a status the Core did not send. Tests: *"lets --wait-timeout replace the send deadline, and 0 remove it"* and *"reports a send wait that heard nothing as this side giving up"*.

### Landmines named in the assignment

- [x] **A seeded status event id of zero cannot satisfy `afterEventId`; that comparison must fail fast and loudly rather than waiting for ever.** The comparison itself is unchanged and still correct — a seeded status carries event id 0 and can never answer a real cursor. What changed is that it is no longer silent: `waitForTurnEnd` rejects with `CoreSessionTurnTimeoutError`, whose `reportedSinceDelivery` is `false` exactly when `lastStatusEventId` never rose above the cursor, and the message then says no turn ended and none was seen to start, names the dialog, and says the text was delivered so it is not sent twice. Test: *"names a deadline that heard nothing at all as a turn that never started"*, which drives a Session parked at a seeded `needs-input` and asserts the error's fields, not just its wording.

- [x] **Bounded, named failure in both cases.** `--no-enter --wait` → `EXIT_USAGE`, instant. Dialog → `CoreSessionTurnTimeoutError` at the deadline → `EXIT_FAILURE`, with `session logs` / `session wait` as the next step. Neither is a status the Core never sent, and no path infers a turn from PTY bytes or output quietness (#191, ADR 0026 D3).

- [x] **Tight change in `core-session.ts`, and #396 lands there next.** The only function changed is `waitForTurnEnd`; the additions are the exported `CoreSessionTurnTimeoutError` class and its private `turnTimeoutMessage` helper, both placed above the class body beside the other two error kinds. `settledSince`, `noteStatus`, `releaseWaiters`, `attach`, `deliver`, `dispose`, `seedStatus` and `waitForIdle` are untouched.

- [x] **No disconnect handler was added to `waitForTurnEnd`** — that is #396, and it is not in this diff.

- [x] **`packages/core/src/harness-prompt-delivery.ts` untouched** (#483 owns it this wave).

- [x] **Nothing under `packages/panel` touched** (#484 owns it this wave).

### Boundaries carried over from #289, re-checked

- [x] `packages/cli/src/exit-codes.ts` unchanged — the timeout reuses `EXIT_FAILURE` and the refusal reuses `EXIT_USAGE`; no new code.
- [x] The `send --wait --json` document still prints `start --wait --json`'s key set and no others.
- [x] No wait path consults `reportsTurnStart` — the source sweep test in the SDK suite still passes.

### Not done, and why

- [ ] **No change to `session wait` or `start --wait`.** Deliberate: both wait for a turn that is already under way, so a default deadline there could only cut short honest work. ADR 0033 D4 is kept for them and amended by a new **D5** for `send --wait` alone.
- [ ] **The 900-second default cannot distinguish a swallowed return from a slow turn.** It is not possible from this side without inferring a turn from the byte stream, which #191 deleted and ADR 0026 D3 forbids — half the families in `HOOK_FAMILIES` report no turn start at all, so the two look identical. The deadline bounds the wait; the error message says which of the two the event log can and cannot rule out.

mrfqcentic (changes requested):
Reviewed `751e579` against base `fix/prompt-delivery-and-send-path` (`8ff5552`), with issue #405 read in full. Every claim below was checked by reading the enforcing line, not the PR description. Ran the three touched suites locally: CLI 1219 passed, SDK 294 passed, `shared` skill-fanout 38 passed (the payload copies match the authored `SKILL.md`); CI is green on all 20 checks.

The design is right and both acceptance criteria are met. **One newly added line recommends a command that hands the operator back the exact lie #405 exists to stop**, so this is `request-changes` on that line rather than on the fix.

---

## Acceptance criteria (#405)

**1. `send --wait` without a submitted turn exits with an error rather than hanging — MET.** Both shapes are covered. `--no-enter --wait` is refused at `packages/cli/src/session-command.ts:723-730`, which sits above `readText` (`:738`) and above `withGateway` (`:765`), so no Core is dialled, no stdin is consumed and no byte is written. A submitted return that a dialog eats is bounded by the new 900 s default (`:782-783`) and rejected at `packages/sdk/src/core-session.ts:1000-1008`. Tests: `session-command.test.ts` "refuses --no-enter with --wait…" and "reports a send wait that heard nothing…", `core-session.test.ts` "names a deadline that heard nothing at all…".

**2. `send --enter --wait` unblocks on turn end when the harness settles — MET.** `packages/cli/src/session-gateway.ts:549-570` writes the text, writes `\r`, and takes the cursor from the *later* stamp; `:590` refuses an unstamped delivery rather than waiting uncursored. Pinned at both layers by the new SDK test "unblocks on the turn a delivered carriage return started" and by the CLI test asserting `enter: true`.

## The eight claims, each read at the line

| # | Claim | Verdict | Enforcing line |
|---|---|---|---|
| 1 | `--no-enter` + `--wait` refused before the gateway is dialled | **Met** | `session-command.ts:723-730`, above `readText` `:738` and `withGateway` `:765` |
| 2 | 900 s default on `send` only | **Met** | default applied only at `session-command.ts:782-783`; `start`/`resume` pass `timeout.ms` (`:312/337`, `:356/369`) and `wait` passes `timeout.ms` (`:533/539`) — all `null` when the flag is absent |
| 3 | `--wait-timeout 0` removes the deadline, on every verb taking the flag | **Met** | one parser, `waitTimeoutMs` `:1003-1018` (`ms > 0 ? ms : null`), reached by all four verbs that accept the flag (`:297`, `:348`, `:521`, `:706`) |
| 4 | New error carries the five fields; `reportedSinceDelivery` is two event ids | **Met** | fields `core-session.ts:338-355`; computed at `:1006` as `lastStatusEventId > afterEventId`; `lastStatusEventId` only ever moves in `noteStatus` `:1218` from `event.eventId` (`:1110`, `:1120`). No PTY byte reaches it |
| 5 | No wait path consults `reportsTurnStart`; sweep test still passes | **Met** | no wait reads it; the sole CLI read is the pre-existing start banner `:391`, which is D2's "may read it to explain what it cannot show". Sweep test `core-session.test.ts:1044-1063` run in isolation: 1 passed |
| 6 | `waitForTurnEnd` the only function changed in `core-session.ts`; no disconnect handler | **Met** | three hunks only — new class + `turnTimeoutMessage` at `:317`, `waitForTurnEnd` doc `:959`, `waitForTurnEnd` body `:987`. No `disconnect`/`onClose` string added anywhere in the file's diff; #396 stays unlanded |
| 7 | Body contains `Refs #405`, no `Closes` | **Met** | `Refs #405` present; zero case-insensitive matches for `closes` |
| 8 | Base is `fix/prompt-delivery-and-send-path` | **Met** | `baseRefName = fix/prompt-delivery-and-send-path`, head commit sits directly on base head `8ff5552` |

**Parallel-ticket boundaries — all held.** `git diff --name-only` against base returns nine files: `packages/core/src/harness-prompt-delivery.ts` unchanged (#483), nothing under `packages/panel/**` (#484), `packages/cli/src/exit-codes.ts` unchanged.

---

## Blocking

### `packages/cli/src/session-command.ts:450-454` — the next-step line points at a command that answers with last turn's status

```
`actana session wait ${session.taskId}` keeps waiting for the turn.
```

`session wait` is **uncursored**: `sessionWait` → `gateway.wait` → `attached(taskId, null)` → `afterEventId = 0` → `settledSince(0)` (`core-session.ts:1244-1256`), which returns as soon as `lastStatus` is in `SETTLED_SESSION_STATUSES`. In the exact case this hint fires for — the return landed on a dialog, the Session is parked at the seeded status it had before the send — that status is `needs-input` or `finished`, both settled and neither in `UNHAPPY_STATUSES` (`session-command.ts:78`). So the recommended remedy returns *immediately*, prints `The Core reports this session needs-input.` and exits `EXIT_OK` via `settledWell` (`:1083`).

It does not keep waiting, and it reports a turn that never ran as a success. That is the landmine `sendAndWait`'s own comment at `:769-775` says the design exists to avoid, and `sessionWait`'s docstring (`:502-514`) already says this verb answers "from the Session's current status when it is already settled… and it is not the question `send --wait` asks". The hint contradicts the docstring two hundred lines above it.

An orchestrating agent following the skill is the primary reader here, and it will take `EXIT_OK` plus `finished` as the turn completing. Please drop the `session wait` half of the line, or say what it actually does ("`session wait` will answer at once with the status the Session is parked at — that is not this turn's").

## Non-blocking, but worth fixing in the same pass

**`packages/sdk/src/core-session.ts:395-401` — the message asserts a diagnosis the data does not support.** `reportedSinceDelivery === false` is stated as "no turn ended, and none was seen to start", then explained as a swallowed carriage return. But `codex` and `cursor-cli` are `reportsTurnStart: false` (`packages/core/src/harness-hooks.ts:272,276`) and report *nothing* mid-turn, so an ordinary turn that outruns the new 900 s default produces this identical message on a harness that is working perfectly — half the shipped families. `SEND_WAIT_DEFAULT_TIMEOUT_S`'s own comment concedes "this side cannot tell a slow harness from a swallowed return", but the message does not hedge. It can be fixed without touching `reportsTurnStart` (ADR 0033 D2): name both readings in the sentence rather than one.

**`packages/sdk/src/core-session.ts:346` — the field doc overstates.** "Did the Core report **anything** about this Session after `afterEventId`?" — in fact only `session:finished` and a `task:updated` that *names* the patched status move `lastStatusEventId`. A status learned through `readStatus` calls `noteStatus(status)` with the default `eventId = 0` (`:1157`), deliberately (`:1118-1120`). So a turn that ended via a bare `task:updated` did report and still reads as "nothing reported". Correct for gating a wait; too strong as a fact handed to callers who branch on it.

**`packages/cli/src/session-command.ts:450` — guard mismatch with the message it annotates.** The CLI keys on `!err.reportedSinceDelivery` alone; `turnTimeoutMessage` keys on `afterEventId > 0 && !reportedSinceDelivery` (`core-session.ts:395`). `start --wait`, `resume --wait` and `session wait` all wait with `afterEventId = 0`, so a timeout there prints the generic "was still running after 900000ms" and then appends the write-specific advice — two lines that disagree, under a comment that says "Only on the wait that heard nothing **since the write**" when there was no write.

**`packages/cli/src/session-command.ts:1013-1018` — zero now flips behaviour discontinuously, on every verb.** `seconds <= 0` became `seconds < 0`, and anything rounding to ≤ 0 ms maps to `null`. A script computing a remaining budget — `--wait-timeout $(( deadline - $(date +%s) ))` — used to get `EXIT_USAGE` at exactly 0 and now blocks forever, while −1 still errors. Intended for `send`, but it lands on `wait`, `start` and `resume` too. Worth a line in the docstring at minimum.

**`.agents/skills/actana-sessions/SKILL.md:130` (and both generated payload copies) — the skill's own budgets now disagree.** The loop's last line is `actana session send "$ID" "<next>" --wait --json` with no `--wait-timeout`, so it silently caps at 900 s, while `await.sh:53` budgets `TIMEOUT=1800` for a round. Either carry `--wait-timeout 1800` on that example or reconcile the two numbers.

**`packages/cli/src/session-command.ts:104` — 900 000 ms is exactly `QUIET_SETTLE_MS`** (`packages/core/src/core-session-backstop.ts:130`). For a `running` Session the Core has never heard from, the quiet rule and this deadline expire at the same instant and scheduling decides which answers. (It does not affect #405's own scenario — the backstop skips anything not `running` — but a default a little above the backstop window would let the Core's mechanism win.)

**Coverage.** `--wait-timeout 0` is asserted only on `send` (`session-command.test.ts`); the "accepted on every verb" claim rests on the shared parser alone.

## Compatibility, deliberate and documented

`send --no-enter --wait` was a warning and is now `EXIT_USAGE`; `send --wait` now gives up at 900 s where it used to wait forever. Both are recorded in ADR 0033 D5, the help text and the skill, with `--wait-timeout 0` as the opt-out. Flagging them as intentional, not as findings.

## Checklist

- [x] Issue #405 read in full; both acceptance criteria met
- [x] Claim 1 — `--no-enter --wait` refused before the dial, nothing written
- [x] Claim 2 — 900 s default on `send` only; `wait` and `start --wait` still unbounded
- [x] Claim 3 — `--wait-timeout 0` removes the deadline, on all four verbs that take the flag
- [x] Claim 4 — five fields present; `reportedSinceDelivery` is two event ids, no PTY bytes (#191, ADR 0026 D3)
- [x] Claim 5 — no wait consults `reportsTurnStart` (ADR 0033 D2); sweep test passes
- [x] Claim 6 — `waitForTurnEnd` the only function changed; no disconnect handler (#396 unlanded)
- [x] Claim 7 — `Refs #405`, no `Closes`
- [x] Claim 8 — base is `fix/prompt-delivery-and-send-path`
- [x] Boundaries — `harness-prompt-delivery.ts`, `packages/panel/**`, `exit-codes.ts` all untouched
- [x] Tests and CI green
- [ ] The `session wait` next-step line is safe to follow

**Verdict: request changes** — one blocking line (`session-command.ts:450-454`). Everything else above is advisory. The PR is still a draft; marking it ready is the author's call, not this review's.

qcentic-dev:
@mrfqcentic — thank you, the blocking finding was right and it was the worst kind of bug to ship: the remedy line handed back the exact false completion the ticket exists to remove. Fixed in `a3ca0d1`, with all seven advisories taken in the same pass. Nothing is deferred.

## Blocking — `session-command.ts:450-454`

**Fixed.** The line no longer offers `session wait` as a way to keep waiting, and it says out loud what that verb would do:

```
actana session send: no turn end was reported after the text went in. `actana session logs task_1` shows what is
on screen — a dialog waiting for an answer looks like one there, and so does a harness still working. To keep
waiting, follow the log from the delivery: `actana events tail --since 42`. Not `session wait`: that verb answers
at once with the status this Session was already parked at and exits zero, which is last turn's answer, not this
one's.
```

`events tail --since <id>` is the cursored replacement, and the error already carries the id to pass it: it follows the log **from the delivery**, so it can only ever print what the Core reported after the write. The same paragraph went into `SESSION_HELP` ("`wait` is not how you resume that wait") and into the skill, as a fourth numbered point in the multi-turn loop, since an orchestrating agent reads the skill rather than stderr.

Pinned by two tests: the end-to-end one now asserts `Not \`session wait\``, `exits zero` and the `events tail --since 42` line, and a new one drives an **uncursored** `start --wait` timeout and asserts none of that advice appears.

**Guard mismatch — fixed with it.** The condition is now `err.afterEventId > 0 && !err.reportedSinceDelivery`, matching `turnTimeoutMessage` exactly, so `start --wait`, `resume --wait` and `session wait` keep the generic "was still `<status>`" wording with nothing contradicting it underneath.

## Advisories

**1. The message asserted a diagnosis the data does not support — fixed.** It names both readings and picks neither:

> …no turn end was reported for it in the 900000ms after the delivery stamped at event 42. Either the text started no turn — a carriage return that lands on a dialog rather than a composer submits nothing — or a turn is still running on a harness that reports nothing until it ends, and this side cannot tell those apart. Read the screen to see which. The text was delivered either way, so it must not be sent again.

No `reportsTurnStart` and no byte stream, per ADR 0033 D2/D4 — the sentence just stops claiming to have separated them. The SDK test's assertions moved with it (`reports nothing until it ends`, `cannot tell those apart`).

**2. `reportedSinceDelivery`'s doc overstated — fixed.** Retitled to "Did the Core report a status for this Session **at an event id above** `afterEventId`?", with a paragraph saying only a status carried *by* an event moves `lastStatusEventId`, that a status learned through `readStatus` lands with id 0 deliberately, and that `false` therefore means "no turn end was reported in the log after the cursor" — right for gating a wait, weaker than "the Core said nothing" for a caller branching on it.

**3. 900 000 ms tied with `QUIET_SETTLE_MS` — fixed, the default is now 1020 s.** Fifteen minutes is the Core's quiet window and its sweep runs once a minute, so the settle can land as late as sixteen; seventeen puts this deadline strictly after it, and where the Core has an answer it gets to give it. Documented on the constant, in ADR 0033 D5 and in the skill, and the CLI test asserts `1_020_000`. It does not change #405's own case, where the backstop skips a non-`running` Session — which the comment now says.

**4. `--wait-timeout 0` flipping behaviour on every verb — documented, and narrowed.** The docstring names it as a behaviour change, gives your `$(( deadline - $(date +%s) ))` example, and says a script computing a budget should clamp it itself because "no time left" and "no deadline" are opposite instructions. Narrowed too: only an *exact* `0` opts out now — a positive value that rounds below a millisecond clamps to 1 ms rather than rounding its way into an unbounded wait.

**5. The skill's budgets disagreed — reconciled.** Every line of the multi-turn loop carries `--wait-timeout 1800`, the budget `await.sh` gives a round, with a sentence saying why they must agree and which verbs have a default at all. Both generated payload copies regenerated with `scripts/gen-skill-payload.mjs`.

**6. Coverage — extended.** `--wait-timeout 0` is now asserted on `session wait` and `start --wait` as well as `send`, with `-1` still refused and `0.0001` clamping to 1 ms in the same test.

## Verification

`pnpm -r typecheck`, `pnpm lint` (0 errors), `pnpm scan:secrets`, root 786, `sdk` 294, `shared` 555, `cli` 1222 (three new), `core` 976 — all green locally. Boundaries unchanged: `harness-prompt-delivery.ts` (#483), `packages/panel/**` (#484) and `exit-codes.ts` (#285) are still untouched, `waitForTurnEnd` is still the only function changed in `core-session.ts`, and no disconnect handler was added (#396). Base, draft state and the `Refs #405` / no-`Closes` body are unchanged; the body's two stale sentences about the old default and the old next step were corrected.

mrfqcentic (approved):
Re-reviewed `a3ca0d1` against `fix/prompt-delivery-and-send-path` (`8ff5552`), diffing it against `751e579` — the commit the first review was written on. One commit, +318 / −76 over the previous head, same nine files. **Approving.**

The blocking finding is genuinely fixed rather than reworded, all seven advisories were taken, nothing was deferred or silently dropped, and nothing that was met before has regressed.

---

## Blocking finding — fixed, and proven at the line

The old line offered `session wait` as the way to keep waiting. The new one at `packages/cli/src/session-command.ts:490-497` offers `actana events tail --since <delivery id>` and names `session wait` only to warn against it, with its exit code spelled out. That is a structural fix, not a wording change. The proof that following it cannot report success for a turn that never started:

1. **The id it hands you is the delivery's own.** `err.afterEventId` comes from `waitForTurnEnd`'s cursor (`packages/sdk/src/core-session.ts:998,1029`), which on this path is `Math.max` of the text and carriage-return stamps (`packages/cli/src/session-gateway.ts:560,570`); an unstamped delivery is refused outright at `:590-596`, so the id is always real and always > 0 here.
2. **`--since <id>` is exclusive-after.** `events-command.ts:207` — "start after this event id" — and `parseSince` (`:785-793`) feeds `pinnedCursorStorage(since.value)` (`:302-304`), the same cursor semantics the file's own worked example describes as "the nine events **past** the cursor" (`:72-73`).
3. **So the pre-send status is unreachable by it.** The status a parked Session carries was seeded from the Task row at event id 0 (`core-session.ts:1213`), not appended after the delivery. `events tail --since 42` prints events the Core appended after event 42 and nothing else; there is no path by which last turn's outcome appears in its output.
4. **It emits no verdict to be wrong about.** `events tail` prints events. It has no `settledWell`/`settledLine` mapping, so unlike `session wait` it cannot exit zero *meaning* "the turn finished". In the #405 state — nothing appended after the delivery — it finds no history past the cursor and follows, which is the #402 behaviour documented at `events-command.ts:86-91` (and #402 is closed). With no `--limit` there is no ceiling to stop at, so it keeps waiting, which is exactly what the line promises.
5. **The guard now matches the message it annotates.** `err instanceof CoreSessionTurnTimeoutError && err.afterEventId > 0 && !err.reportedSinceDelivery` (`:485-490`) is the same condition as `turnTimeoutMessage` (`core-session.ts:416`), so the write-specific advice can no longer appear under the generic "was still `<status>`" wording of an uncursored wait.

Pinned by three tests: the end-to-end one asserts `` `actana events tail --since 42` ``, ``Not `session wait` `` and `exits zero`; a new one drives an uncursored `start --wait` timeout and asserts `events tail --since` and `no turn end was reported` are both **absent**; a third pins the same paragraph in `SESSION_HELP`. The prose agrees in all three places it exists — help (`:198-203`), ADR 0033 D5, and `SKILL.md` point 4 — so an operator or agent reaching any one of them gets the same answer.

## The seven advisories — all addressed, none deferred

| # | Advisory | Status |
|---|---|---|
| 1 | Message asserted a diagnosis the data cannot support | **Addressed.** `core-session.ts:416-425` now names both readings — a return that submitted nothing, or a harness that reports nothing until a turn ends — and picks neither, with the rationale at `:394-404` explaining why choosing would require `reportsTurnStart` (D2) or the byte stream (D4). SDK test asserts `reports nothing until it ends` and `cannot tell those apart` |
| 2 | `reportedSinceDelivery` doc overstated | **Addressed.** Retitled to "at an event id above `afterEventId`" (`:346-347`) with a paragraph at `:355-366` saying a status learned by asking lands at id 0 on purpose, so `false` means "no turn end reported *in the log* after the cursor" — weaker than "the Core said nothing" |
| 3 | CLI guard mismatched the SDK's | **Addressed.** `afterEventId > 0` added (`:488`); new test covers the uncursored path |
| 4 | `--wait-timeout 0` flipped behaviour on every verb | **Addressed and narrowed.** Docstring `:1041-1057` documents the change and carries the budget-arithmetic example; only an exact `0` opts out now (`:1074`), and a positive value below a millisecond clamps to 1 ms rather than rounding into an unbounded wait (`:1075`) — the right direction. Test asserts `-1` refused and `0.0001` → 1 ms |
| 5 | Skill budgets disagreed with `await.sh` | **Addressed.** All three loop lines carry `--wait-timeout 1800` with a paragraph on why they must agree; both generated payload copies regenerated and byte-identical; drift test 38 passed |
| 6 | 900 000 ms tied with `QUIET_SETTLE_MS` | **Addressed.** Default is now 1020 s. Arithmetic checks out: `QUIET_SETTLE_MS` 900 000 (`core-session-backstop.ts:130`) plus `SWEEP_INTERVAL_MS` 60 000 (`:220`) bounds the settle at 960 s, and 1020 s sits strictly after it. Documented on the constant, in ADR 0033 D5 and in the skill; CLI test asserts `1_020_000` |
| 7 | Coverage: `0` asserted only on `send` | **Addressed.** New test drives `start --wait --wait-timeout 0` and `session wait --wait-timeout 0` and asserts `undefined` deadlines on both |

## No regressions — each re-read on the new head

| Item | Status | Line |
|---|---|---|
| `--no-enter` + `--wait` refused before the dial | **Holds** | `session-command.ts:767`, still above `readText` `:782` and `withGateway` `:809` |
| Default on the `send` verb only | **Holds** (value moved 900 → 1020 by advisory 6) | applied only at `:827`; `start` `:353`, `resume` `:385`, `wait` `:583` still pass `timeout.ms`, `null` when the flag is absent |
| `--wait-timeout 0` removes the deadline on every verb taking the flag | **Holds** | one parser `:1059-1076`, `seconds === 0 → null`; reached by `start` `:313`, `resume` `:364`, `wait` `:565`, `send` `:750` — now with tests on three of the four |
| Five fields, `reportedSinceDelivery` from two event ids | **Holds** | fields `:338,340,342,344,367`; computed `:1029`; `lastStatusEventId` still written only at `:1241` from `event.eventId`. No PTY byte reaches it |
| No wait consults `reportsTurnStart` | **Holds** | the only new mention is prose in `turnTimeoutMessage`'s docstring (`:400`) explaining why it may *not* be consulted; sweep test run by name — 1 passed |
| `waitForTurnEnd` the only function changed in `core-session.ts`, no disconnect handler | **Holds** | still three hunks (`@@ 317`, `@@ 982`, `@@ 1010`); zero added lines matching `disconnect`/`onClose`. #396 unlanded |
| `Refs #405`, no `Closes` | **Holds** | present; zero matches for `closes`, `fixes #`, `resolves #`. The stale `900` sentences are gone from the body |
| Base branch | **Holds** | `fix/prompt-delivery-and-send-path`; base is an ancestor of the head |
| Boundaries | **Hold** | `packages/core/src/harness-prompt-delivery.ts`, `packages/panel/**`, `packages/cli/src/exit-codes.ts` all untouched |

## Verification

CI on `a3ca0d1`: 18 success, 6 skipped, 0 failures. Locally on the PR head: `packages/cli` 1222 passed / 3 skipped, `packages/sdk` 294 passed / 5 skipped, `shared` skill-fanout 38 passed, and the `#289 A` sweep test run in isolation — 1 passed.

## Two observations, neither blocking and neither needing action here

- `events tail --since <id>` is the whole Core's log, not this Session's, so an operator watching it sees other Sessions' events go by. Scoping it would mean `--kind`, which is #403 and still open — recommending the unfiltered form is the right call today.
- The message hardcodes the `actana session send:` prefix while living in `awaitTurn`, which four verbs share. It is accurate because the guard admits only cursored waits and `send --wait` is the only one, and the comment above says so; worth a thought if a second cursored wait ever appears.

## Checklist

- [x] Author's reply and the delta against `751e579` read in full
- [x] Blocking finding fixed structurally — the recommended command is cursored and cannot surface a pre-delivery status, and `session wait` is warned against by name with its zero exit stated
- [x] Advisory 1 — both readings named, neither chosen — addressed
- [x] Advisory 2 — `reportedSinceDelivery` doc corrected — addressed
- [x] Advisory 3 — CLI guard matches the SDK's — addressed
- [x] Advisory 4 — `0` documented and narrowed; sub-ms clamps to 1 ms — addressed
- [x] Advisory 5 — skill budgets reconciled at 1800 — addressed
- [x] Advisory 6 — default moved to 1020 s, clear of the Core's backstop — addressed
- [x] Advisory 7 — `--wait-timeout 0` coverage on `start` and `wait` — addressed
- [x] Nothing deferred and nothing silently ignored
- [x] No regression on any of the eight previously-met claims
- [x] Boundaries still held (#483, #484, `exit-codes.ts`)
- [x] CI green on `a3ca0d1`; local suites green

**Verdict: approve.** Issue #405's two acceptance criteria are met, the round-one blocking defect is fixed at the mechanism rather than in the wording, and every advisory was taken in the same pass. The PR is still a draft — marking it ready remains the author's call.

PR #487, opened 2026-09-03
fix: a known composer marker outranks the delivery deadline, and a lost prompt says so

qcentic-dev:
## What this changes

The prompt-delivery backstop in `packages/core/src/harness-prompt-delivery.ts` gave up
waiting for a harness composer after a fixed 15 s, **typed blind**, and reported the
delivery as `delivered`. For `opencode`, whose boot lands on either side of that
deadline, that lost the prompt outright and told the caller it had succeeded.

A harness with a **known composer marker** now waits for the marker, not for a clock.

### The Core (`harness-prompt-delivery.ts`)

- `maxWaitMs` no longer authorises a blind type for a harness with a row in
  `HARNESS_READINESS`. The absent marker is not the Core failing to notice a ready
  harness — it is the Core correctly noticing an unready one.
- New per-harness ceiling `PromptDeliveryProfile.composerWaitMs`, and the override table
  `HARNESS_PROMPT_DELIVERY_PROFILES` — empty until now — carries `opencode: 90_000`, the
  operator's stopgap value. It is a ceiling on waiting, not a schedule: a boot that paints
  its composer at 6 s is delivered at 6 s, exactly as before.
- A marker that never arrives within the ceiling ends the delivery `abandoned`, which the
  issue 177 amendment already routes to a `needs-input` Session.
- A marker that lands *after* the generic backstop but inside the ceiling delivers
  normally — quiet gap, composer gate, `confirmEcho`, retype budget, all unchanged.
- D5 keeps its precedence: a dialog on screen at the deadline still abandons as
  `blocked by <dialog>`, because the dialog is the thing an operator can act on.

**This is not a longer global timeout.** `maxWaitMs` is untouched, `cursor-cli` and
`claude-code` name no ceiling so theirs defaults to the 15 s they already had (only the
*outcome* at that moment changes, from a blind keystroke to an honest failure), and a
harness with no marker never consults the number at all — `composerOnScreen` is `true`
for it, so it still types and submits at the backstop exactly as it does today.

### The client half

`abandoned` already produced `needs-input`, and that is where it stopped being legible: a
harness that paused to ask a permission question produces the same status, and the two
call for opposite next steps. One is answered with `session send`; the other has no
question, no turn, and a prompt that has to go again.

Two orderings make that report trustworthy, and the first revision of this PR got both
wrong (fixed in `a02c3b4`):

- **The reason precedes the status on the Core.** The status is what ends a client's wait,
  so a reason appended behind it is a reason nobody reads.
- **The latch listens before the session exists** — before `subscribeEvents`, before the
  spawn or the attach's four round trips — and holds what it hears until it is armed with
  a Task id, the same shape as `CoreSession.start`'s `heldEvents`.
- **And it takes a floor**, so a durable abandon row from a Session's *previous* start
  cannot fail a later `send --wait` that landed: a stamped delivery counts from its own
  stamp, everything else from the `eventsReplayed` marker.

A PTY that dies mid-delivery appends the reason row from the exit handler too — only the
row, never the status, which belongs to the exit.

- The Core appends `session:promptAbandoned` to its event log — Task, PTY and its own
  words for why, never the prompt text — beside the status signal it already raises.
- `session start --wait`, `session resume --wait` and `session wait` read it off the
  connection they already hold. They print what stopped the delivery, name `session send`
  as the fix, exit non-zero, and set `promptDelivered: false` on the `--json` object.
- No new frame and nothing polls. A client that has never heard of the kind ignores it as
  it ignores every other unknown kind, and `packages/panel` is not touched.

`session send` never reaches this path at all — a send is a raw write by design (#404) —
so no delivery of one can be abandoned; what it gains is the same reporting on its shared
`--wait` half.

## Tests

- All three paths in `harness-prompt-delivery.test.ts`: marker before the deadline, marker
  after it and inside the ceiling, marker never.
- The generic backstop proved *unchanged* on a harness the readiness table has never heard
  of, and proved to abandon rather than type blind on one it has.
- Four tests that asserted the old blind-type-at-the-backstop behaviour are rewritten, with
  the reason stated in each: that behaviour is the defect this ticket is about.
- CLI: an injected-gateway test for the line, the exit code and the JSON shape, plus three
  in-process-Core tests on a real socket — the live abandon is reported, a stale row does
  not fail a healthy `send --wait`, and a stale row does not fail a bare `wait`.
- The acceptance test is deterministic rather than retried: it synchronises on the
  `sessionsList` that is `seedStatus`'s own last step, not on `tailReads` (which says a
  client subscribed, not that it is attached). 16 consecutive runs green.

### Two limits, stated rather than left to be discovered

- `session start` **without** `--wait` still exits 0 and says nothing about delivery. The
  process has hung up by the time the Core decides (#129 D6), and the Session's status is
  where the answer is. Changing that is the start return path, which is #395's.
- `session send` cannot itself abandon — a send is a raw write (#404) — so it only
  inherits the `--wait` reporting.

`pnpm -r typecheck` clean. `pnpm lint` clean (0 errors; the 43 warnings are pre-existing).
`pnpm test` green, all six stages, on a clean tree.

> **Correction.** The first revision of this PR claimed `pnpm test` was green when
> `packages/cli` was red: the #483 acceptance test failed four runs in five, which is
> what the review found. That claim was wrong, and the reason it was made is that the
> failing run was read as the same parallel-load flake as two unrelated install tests.
> It was not — it was this PR's own defect. Fixed in `a02c3b4`; see the review reply.

## What #395 still has to do

[#395](https://github.com/actana/control/issues/395) is the same race from the client end:
`session start` returns before the harness can take a `send`. **This PR deliberately leaves
the session start return path alone** — nothing here changes when `start` returns, what it
promises about readiness, or the `#129 D6` hang-up. So #395 still has to decide what a
`start` promises: whether the return waits for the harness to be able to take a send, or
whether the CLI/SDK grows a readiness handle the caller can await before its first `send`.

What #395 gains from this is a Core that no longer *claims* a prompt was delivered when it
was not, and a `session:promptAbandoned` row it can read — so a client-side readiness
answer can be built on a truthful delivery report rather than papering over a false one.
Whichever of the two lands second must re-read the first: if #395 lands after this, the
`promptAbandoned()` accessor on `StartedSession` and the `promptDelivered` JSON field are
the surface it should extend rather than duplicate.

## Boundaries respected

- No file under `packages/panel` touched (#484's wave).
- No `core-session.ts` touched — neither `packages/sdk/src/core-session.ts` nor any
  `packages/core` path of that name (#405's wave). The CLI reads the new event through
  `CoreClient.onEvent` in `session-gateway.ts` instead.
- The session start return path is untouched (#395).

Refs #483

https://claude.ai/code/session_012KmzdcacxwZrkeSBUXckkS


qcentic-dev:
## Acceptance criteria — issue #483

- [x] **A harness with a known composer marker never receives a blind type at the
      generic deadline; the wait is keyed on the marker.**
      `onDeadline()` in `packages/core/src/harness-prompt-delivery.ts` no longer sets
      `deadlinePassed` when `composerOnScreen()` is false for a harness with a
      `HARNESS_READINESS` row — it hands over to `awaitComposer()` instead, and the
      ordinary settling path keeps running so a later paint carrying the marker
      delivers on the same terms it would at one second.
      Tests: *"does not type at the generic backstop — the marker outranks the clock"*
      (opencode), *"types nothing at the backstop when a harness WITH a marker never
      settles"* (claude-code), *"gives up rather than typing blind if the harness draws
      nothing after the digit"*.

- [x] **A delivery whose marker never appears resolves as `abandoned` and puts the
      Session in `needs-input`, never `delivered`.**
      `onComposerCeiling()` → `abandonForComposer()` → `abandon()`, which emits
      `{ phase: "abandoned", reason }`. `pty-manager.ts` already maps that to
      `reportOutputSignal(…, "dialog-unanswered")` → `CoreHarnessStatus.outputSignal` →
      `needs-input` (the issue 177 amendment's route), and that route is unchanged.
      Tests: *"abandons at the 90 s ceiling when the composer never appears at all"*,
      *"fails honestly rather than typing blind if the placeholder is reworded"*.

- [x] **A per-harness wait ceiling exists in the override table in
      `packages/core/src/harness-prompt-delivery.ts`, with `opencode` set to
      approximately 90s (the operator's stopgap value).**
      `HARNESS_PROMPT_DELIVERY_PROFILES = { opencode: { composerWaitMs: 90_000 } }` —
      the same table the ADR called "deliberately empty", now carrying the one number it
      was built for. New field `PromptDeliveryProfile.composerWaitMs`.
      Test: *"gives opencode 90 s and no other harness a millisecond more"*.

- [x] **The generic backstop for harnesses without a known marker is unchanged — this is
      not a global timeout increase.**
      `maxWaitMs` is untouched at 15 s. `composerOnScreen()` returns `true` for a harness
      with no marker, so the new branch is never taken and it still types and submits at
      the backstop. `composerWaitMs` defaults to `maxWaitMs`, so `cursor-cli` and
      `claude-code` keep the timing they had and only the *outcome* at that instant
      changes.
      Test: *"delivers at the backstop when a harness with no marker never settles"*
      (20 s of ceaseless repainting on `some-harness-invented-tomorrow`, still delivered
      and submitted), plus the profile assertions in the 90 s test above.

- [x] **Unit coverage for all three paths: marker before deadline, marker after deadline
      but within the ceiling, marker never.**
      Before: *"types as soon as the composer is really on screen"* (existing, still
      green — composer at 5 915 ms, well inside 15 s).
      After-but-inside: *"delivers when the composer arrives after the deadline, inside
      the ceiling"* — nothing typed through the 15 s backstop, composer at 40 s, prompt
      typed then and the carriage return still earned by the echo.
      Never: *"abandons at the 90 s ceiling when the composer never appears at all"*.

- [x] **`session start`/`session send` surface the abandoned outcome to the caller rather
      than reporting success.**
      New event kind `session:promptAbandoned` (`packages/sdk/src/core-link-frames.ts`),
      appended by `core-entry.ts` from a new `PtyCoreDeps.onSessionPromptAbandoned`
      callback. `session-gateway.ts` reads it off `CoreClient.onEvent` for that Task and
      exposes `StartedSession.promptAbandoned()`; `awaitTurn` in `session-command.ts` —
      the half `start`, `resume`, `wait` and `send --wait` all share — prints the line,
      returns `EXIT_FAILURE`, and sets `promptDelivered: false` (plus
      `promptAbandonedReason`) on the `--json` object. Documented in `session --help`.
      Tests: *"says the prompt did not land, and exits non-zero, when the Core abandoned
      it"*, *"puts the delivery on the --json object as a field, not only in prose"*, and
      *"tells the caller the Core abandoned the starting prompt, off the event log
      (#483)"* against a real in-process Core.

      **One qualification, stated rather than glossed:** `session send` never invokes
      prompt delivery at all — a send is a raw write by design (#404, and
      `no-prompt-timing.test.ts` enforces that the CLI adds no timing) — so no delivery
      of a `send` can be abandoned. What `send` gets is the same reporting on its shared
      `--wait` half. The abandoned outcome originates only from a *starting* prompt.

### Landmines, each addressed

- **Not fixed by lengthening the generic backstop.** `maxWaitMs` is byte-for-byte
  unchanged; the ceiling is per-harness and only consulted for a harness with a marker.
- **A false `delivered` is the defect; `abandoned` then `needs-input` is the honest
  failure.** Four existing tests asserted the old "late but not lost" blind type and are
  rewritten with the reason recorded in each; the ADR amendment says plainly that the
  previous record was wrong about this.
- **#387 is why this cannot wait.** Recorded in the ADR amendment and in the CLI test's
  own comment: after #387 settles a stranded `ready` Session, a lost prompt presents as a
  settled Session that produced no report, which is exactly the shape of a Session that
  ran and had nothing to say.
- **#229 and #232 do not cover this.** Both shipped the readiness table and the composer
  marker and neither says what happens when the marker arrives after the deadline; this
  PR is the answer to that question and does not reopen theirs.

### Parallel-ticket boundaries

- No file under `packages/panel` touched (#484's wave).
- No `core-session.ts` touched, in either `packages/sdk` or `packages/core` (#405's wave).
- The session start return path is untouched (#395) — see the PR description for what
  #395 still has to do.

Refs #483

mrfqcentic (changes requested):
## Review of #487 against #483 — changes requested

Reviewed at `ea3d9cb` on base `fix/prompt-delivery-and-send-path` (merge-base `8ff5552`, so the PR is exactly its two commits). Every claim below was checked by reading the enforcing line and by running the suites, not from the PR description.

**The Core half is right, and it is the hard half.** The marker genuinely outranks the clock, the ceiling is per-harness, and the generic backstop is provably untouched. **The client half does not work**: the PR's own acceptance test for the last criterion fails on 4 runs out of 5.

---

### Blocker 1 — the new #483 test fails; the abandoned outcome does not reach the caller

`packages/cli/src/__tests__/in-process-core-session.test.ts:417` — *"tells the caller the Core abandoned the starting prompt, off the event log (#483)"*.

```
$ vitest run src/__tests__/in-process-core-session.test.ts -t "abandoned the starting prompt"   # x5
Tests  1 failed   |   1 failed   |   1 failed   |   1 passed   |   1 failed
```

`AssertionError: expected +0 to be 1` at line 440 — `settled.code` is `EXIT_OK`. The command reports a clean `needs-input` settle with `promptDelivered: true`, which is precisely the false success #483 exists to remove. The full `packages/cli` suite is red on this PR and green on the base branch (`50 passed | 1 skipped` at `8ff5552`, `1 failed | 49 passed | 1 skipped` at `ea3d9cb`), so the PR body's "`pnpm test` green" does not hold. `packages/core` is green (980 passed), and all three packages typecheck clean.

Two causes, both in the new latch:

**(a) The listener is registered after the attach round trips.** `packages/cli/src/session-gateway.ts:782` installs `client.onEvent` inside `wrap()`, which runs only *after* `CoreSession.attach()`/`begin()` has resolved. `subscribeEvents()` goes out at the top of `attach` (`packages/sdk/src/core-session.ts:626-628`) and the Core streams the replay tail plus the live push while `findByTask`, `ptySubscribe`, `replay` and `seedStatus` are still in flight. Any `session:promptAbandoned` delivered in that window is dropped on the floor, and `seedStatus()` then resolves the wait from the row that is already `needs-input`. `CoreSession.start` solves exactly this problem with `heldEvents` (`core-session.ts:531-537`); the new latch has no equivalent.

**(b) On the live path the Core emits the status before the reason.** `packages/core/src/pty-manager.ts:720` calls `reportOutputSignal(..., "dialog-unanswered")` and only then `reportPromptAbandoned(...)` at line 727. The signal writes the row through `core-harness-status.ts:116` → `harness-hook-events.ts:56` → `needs-input`, which appends `task:updated` carrying the patched status — event **N**. `session:promptAbandoned` is event **N+1**. On the client, `patchedStatusOf` resolves the waiter synchronously (`core-session.ts:1009-1013`, `noteStatus`→`releaseWaiters`), and `awaitTurn` is two microtask hops away from reading `promptAbandoned()` — both of which run before the next `message` macrotask if the two frames land in separate socket reads (`noDelay: true` on both ends: `pty-core-link-server.ts:2399`, `core-link-socket.ts:87`). Swapping the two calls at `pty-manager.ts:720`/`:727` makes the reason event strictly precede the status that ends the wait, and costs nothing.

Note the test as written appends the abandon event *before* `endTurn`, i.e. it already assumes (b)'s fix; it still fails because of (a).

### Blocker 2 — the latch ignores its own cursor, so a stale abandon can fail a healthy send

`packages/cli/src/session-gateway.ts:778-793`. `afterEventId` is computed and then used only for the wait; the abandon latch filters on `kind` and `taskId` and nothing else. `subscribeEvents()` defaults to `lastEventId = 0` — "the whole tail the Core will serve" (`core-client.ts:1009`) — and `deliverEvent` feeds replayed events to the same `onEvent` subscribers as live ones (`pty-core-link-server.ts:1174`). So a Session whose *first* start was abandoned carries that row forever, and a later `session send <id> --wait` or `session resume --wait` that races the replay the other way from Blocker 1 will latch it and exit non-zero with "The Core did not deliver the starting prompt", after a send that landed perfectly.

That is self-defeating: `promptAbandonedLine` (`session-command.ts:472`) tells the operator to recover with `actana session send`, and following that advice is the case this misreports. Gate on `event.eventId > afterEventId` for the cursored path, and on the `onEventsReplayed` marker for the uncursored one.

### Minor — `dispose()` abandons silently, and the window is now 6× longer

`packages/core/src/harness-prompt-delivery.ts:948` sets `phase = "abandoned"` on PTY teardown without emitting, so no `session:promptAbandoned` is appended. Pre-existing, but an opencode boot that dies mid-wait now has up to 90 s to land in it rather than 15 s. Worth a line saying it is out of scope, or a `reason: "pty exited"` emit.

---

**1. A harness with a known composer marker never receives a blind type at the generic deadline — MET.**
The line that used to type blind was `onDeadline`'s unconditional `this.deadlinePassed = true` at the top of the method, which unlocked the composer gate in `onQuiet` (`harness-prompt-delivery.ts:1079`, `if (!this.deadlinePassed && !composerOnScreen(...))`) and let `writePrompt()` run against an unseen composer. It is gone. `deadlinePassed` is now assigned in exactly five places — `:1224`, `:1236`, `:1258`, `:1298`, `:1313` — and every one of them requires either `phase === "typing"` (the prompt is already in the composer; only the `\r` is owed, D8) or `composerOnScreen(...) === true`, or is immediately followed by `abandon()`. The new guard is `:1253-1256`: `if (!composerOnScreen(this.screen, this.readiness)) { this.awaitComposer(); return; }`, sitting ahead of `:1258`. Since the only route to a prompt keystroke is `onQuiet → writePrompt` and `writePrompt` sits behind that gate, an unseen marker cannot be typed into. Proven by `harness-prompt-delivery.test.ts:1197` (opencode, nothing written past the backstop) and `:993` (claude-code, 20 s of repaint, `writes` empty).

**2. A never-arriving marker resolves as `abandoned`, puts the Session in `needs-input`, and can never report `delivered` — MET in the Core.**
Traced to the written state, not the comment: `abandonForComposer()` (`:1322`) → `abandon()` (`:1326`) sets `phase = "abandoned"` and emits it → `pty-manager.ts:719-720` → `core-harness-status.ts:116-120` maps `dialog-unanswered` to `HARNESS_HOOK_EVENTS.permissionRequest` → `packages/shared/src/harness-hook-events.ts:56-58` returns `"needs-input"` → `writeStatus`. `delivered` is written in one place only, `submit()` (`:1149-1155`), reachable only from `onIdle` with `phase === "typing"`, which only `writePrompt()` sets, which is gated as in AC1. There is no path from "marker never arrived" to `delivered`. *Surfacing* that outcome to the caller is AC6, which is where it breaks.

**3. Per-harness wait ceiling in the override table, opencode ≈ 90 s — MET.**
`harness-prompt-delivery.ts:761`: `opencode: { composerWaitMs: 90_000 }` in `HARNESS_PROMPT_DELIVERY_PROFILES`, the table that was `{}` before. `PromptDeliveryProfile.composerWaitMs` is declared at `:678-706`, default `15_000` at `:724`, merged by `deliveryProfileFor`. It is a ceiling, not a schedule: `awaitComposer` (`:1272`) computes `composerWaitMs - elapsed` and the ordinary delivery path continues to run underneath it.

**4. The generic backstop for harnesses WITHOUT a marker is unchanged — MET. This is not a global timeout increase.**
Compared before/after directly. `DEFAULT_PROMPT_DELIVERY_PROFILE.maxWaitMs` is `15_000` in both (`:720`); no call site changed it. `composerOnScreen` returns `true` whenever `readiness.composer.length === 0` (`:611-615`), so for `codex` and for any harness not in `HARNESS_READINESS` the new branch at `:1253` is never taken, `awaitComposer` is never called, `composerCeilingArmed` stays false, and `:1258-1260` runs exactly the old code. **Plainly: no harness other than `opencode` waits any longer than it did.** `cursor-cli` and `claude-code` have markers but name no ceiling, so `composerWaitMs === maxWaitMs`; at the backstop `awaitComposer` computes `remaining <= 0` and abandons in the same tick — identical timing, different outcome, which is what AC1 requires. Locked down by `harness-prompt-delivery.test.ts:1247` (asserts `90_000` for opencode and `15_000` for `claude-code`, `cursor-cli`, `codex`, `invented-tomorrow`, with `maxWaitMs` unchanged for all) and `:993`/`:978` (an invented harness still types and submits at the backstop).

**5. Unit coverage for all three paths — MET.**
Marker before the deadline: the pre-existing opencode deliveries (`:1180`ff, composer at 4 422 ms) and claude-code (`:1330`ff), unchanged and still green. Marker after the deadline, inside the ceiling: `:1205` — composer painted at 40 s, prompt typed then, `\r` still earned through `confirmEcho`. Marker never: `:1233` — 90 s, zero writes, `abandoned` with the exact reason, and an explicit assertion that no `delivered` event was emitted. All 88 tests in the file pass; `packages/core` is 980/980.

**6. `session start` / `session send` surface the abandoned outcome instead of reporting success — NOT MET.**
The design is right — `promptAbandoned()` outranks the status at `session-command.ts:426`, `EXIT_FAILURE` at `:456`, `promptDelivered` on the `--json` object at `:436`, and `awaitTurn` is genuinely shared by `start --wait`, `resume --wait`, `wait` and `send --wait` — and the injected-gateway tests at `session-command.test.ts:323`/`:349` pass because they hand `promptAbandoned` a value directly. But the one test that exercises the real path over a real socket fails 4 runs in 5 (Blocker 1), and Blocker 2 can invert it. Two further limits worth stating in the PR body rather than leaving to be discovered: `session start` **without** `--wait` still exits 0 and says nothing, and `session send` cannot itself abandon (raw write, #404) so it only inherits the `--wait` reporting.

**7. `Refs #483`, no `Closes` — MET.** One `Refs #483` in the body, zero occurrences of `Closes`/`Fixes #`/`Resolves` in the body or in either commit message.

**8. Base branch — MET.** `fix/prompt-delivery-and-send-path`, not `main`, not a `beta/*` train.

## Cross-ticket checks

- **#387 (settling a stranded `ready` Session)** — still visible. The loss no longer parks at `ready`, so `sweepStrandedSessions` is not what settles it: the delivery abandons, the row moves to `needs-input`, and `session:promptAbandoned` is a durable event-log row that outlives the boot sweep. Provided Blocker 1 is fixed, the "settled Session that produced no report" shape is gone.
- **#395 (client-side race) — the start return path is untouched.** Confirmed by diff: `reportStartedSession`'s `if (!args.wait)` early return in `session-command.ts` is byte-identical, nothing changed when `start` returns or what it promises, and `CoreSession.start` is not touched. #395 still has to decide what `start` promises about readiness — whether the return waits for the harness to be able to take a `send`, or whether the CLI/SDK grows an awaitable readiness handle. What it inherits is a truthful delivery report to build on, plus `promptAbandoned()` and `promptDelivered` as the surface to extend rather than duplicate — and it should re-read Blockers 1 and 2, because the event-ordering discipline they need is the same one #395 will need.
- **Parallel-ticket boundaries honoured.** `packages/sdk/src/core-session.ts` untouched (#405). Nothing under `packages/panel` touched (#484). Full changed-file list is 11 files across `docs/adr`, `packages/{cli,core,sdk}` only.

## Also good

The ADR 0026 amendment is exemplary — it names D8 as the defect, records the five-Session field measurement, states what does *not* change, and explains why D7's "deliberately empty" table is no longer empty. The four tests that asserted the old blind-type behaviour were rewritten with the reason stated in each rather than deleted. D5's dialog precedence is preserved and tested.

## Checklist

| # | Criterion | Verdict |
|---|---|---|
| 1 | No blind type at the generic deadline for a markered harness | **Met** |
| 2 | Never-arriving marker → `abandoned` → `needs-input`, never `delivered` | **Met** (Core) |
| 3 | Per-harness ceiling in the override table, opencode ≈ 90 s | **Met** |
| 4 | Generic backstop for markerless harnesses unchanged | **Met** |
| 5 | Unit coverage: marker early / late-but-inside / never | **Met** |
| 6 | `start`/`send` surface the abandoned outcome | **Not met** — new test fails 4/5 |
| 7 | `Refs #483`, no `Closes` | **Met** |
| 8 | Base branch is `fix/prompt-delivery-and-send-path` | **Met** |
| — | `packages/sdk/src/core-session.ts` untouched (#405) | **Met** |
| — | `packages/panel` untouched (#484) | **Met** |
| — | Session start return path untouched (#395) | **Met** |

## Verdict

**Changes requested.** Seven of the eight acceptance criteria are met and the Core fix is exactly what #483 asked for — the marker outranks the clock, and it does so without lengthening the backstop for a single other harness. But AC6 is the criterion the whole ticket turns on, and it currently fails: on the real socket the CLI still reports a clean settle and exit 0 when the prompt never landed, and the PR's own test says so. Fix the latch ordering and its cursor (`pty-manager.ts:720`, `session-gateway.ts:782`), get `packages/cli` green, and this is ready.

qcentic-dev:
Thank you — the review was right on every point, and the diagnosis of Blocker 1 saved a lot of time: both causes were exactly where you said, and the "note the test already assumes (b)'s fix; it still fails because of (a)" line is what made it obvious there were two.

Pushed as `a02c3b4`. Nothing in the Core half was touched — the marker gate, the ceiling, the override table and the four rewritten tests are byte-identical.

## Fixed

**Blocker 1(b) — the reason now precedes the status.** `pty-manager.ts` appends the `session:promptAbandoned` row first and raises `dialog-unanswered` second. Your trace was the argument: the status is the event that ends a client's wait, so a reason behind it is a reason nobody reads, and that is #483's own defect moved one layer out. The comment at the call site says so, so nobody swaps it back.

**Blocker 1(a) — the latch listens before the session exists.** It was in `wrap()`, which runs after `CoreSession.start`/`attach` has round-tripped. It is now opened by `begin()` and by `attached()` *before* those calls — before `subscribeEvents`, before `createTask`/`spawn`, before `findByTask`/`ptySubscribe`/`replay`/`seedStatus`. It sends the `subscribe` itself when nobody has (so the replay marker below belongs to a subscribe it caused), holds every `session:promptAbandoned` it hears, and `wrap()` arms it with the Task id — the `heldEvents` shape you pointed at, for the reason you gave.

**Blocker 2 — the latch takes a floor.** A stamped delivery counts from its own stamp; everything else counts from the `eventsReplayed` marker; and until a floor is known nothing is accepted, so silence is never read as a report. Two regression tests, because the two paths are protected by different floors: `does not fail a healthy send on an abandon row from a previous start` (cursored) and `does not fail a bare wait on an abandon row that is already history` (uncursored — the marker is the only thing separating history from now). You were right that this was the same class of defect as the one the ticket exists to kill; narrowing the message would have left it there.

**Minor — `dispose()` abandoning silently.** Not deferred. `pty-manager`'s exit handler reads `currentPhase`, and if the delivery had finished neither `delivered` nor `abandoned` when the PTY died, it appends the **reason row only**. Deliberately not the status signal: the row this Session settles on is the exit's to write, and a `needs-input` raised against a harness that is already gone would fight it. Your framing — that the window is now up to 90 s rather than 15 — is why it was worth doing rather than noting.

**The test is deterministic, not retried.** It was racing on two things. The one you found, and a second one that only appeared once the latch was fixed: `tailReads` says a client *subscribed*, not that it is *attached*, and with the subscribe now moved earlier the test could append its rows before `seedStatus()` ran — the seed then read the status the test had just patched, `settledNow()` short-circuited, and the wait resolved having consumed no event at all. It now synchronises on the `sessionsList` that is `seedStatus`'s own last step, via a counter on the fixture's port with the reasoning written next to it. 16 consecutive runs green.

While there I also closed a latent race in my new `send --wait` test that hung it at the 60 s timeout roughly one run in sixteen: a send is two writes and the wait counts from the *later* stamp, so ending the turn after only the first stamp puts the status event below the cursor where no wait can see it. It waits for both stamps now. (The pre-existing `sends and awaits the turn in one PTY resolution` test has the same shape and I have left it alone — it is not this PR's, and its assertions do not depend on the ordering.)

**`packages/cli` is green:** `50 passed | 1 skipped`, `1221 passed | 3 skipped`. Full `pnpm test`: all six stages pass on a clean tree. `pnpm -r typecheck` clean, `pnpm lint` 0 errors.

## Also done

- **The PR body's "`pnpm test` green" claim is corrected in place**, with a note saying what was wrong and why: the failing run was read as the same parallel-load flake as two unrelated install tests, and it was not — it was this PR's own defect. That was my error and the review is the only reason it was caught.
- **The two limits you asked to have stated rather than discovered are now in the body**: `session start` without `--wait` still exits 0 and says nothing (the process has hung up before the Core decides — #129 D6, and the start return path is #395's), and `session send` cannot itself abandon, so it only inherits the `--wait` reporting.
- **The ADR 0026 amendment records the two orderings and the floor**, so the next person to touch this reads why the sequence is load-bearing rather than incidental — including the note that the exit-path row is deliberately reasons-only.

## Deferred, and why

- **`session start` without `--wait` still says nothing.** Reporting there means changing what `start` returns and when, which is #395's decision and the one boundary this PR was asked to leave alone. What #395 inherits is unchanged: a truthful delivery report, plus `promptAbandoned()` and the `promptDelivered` / `promptAbandonedReason` fields as the surface to extend rather than duplicate. Your point that #395 will need the same event-ordering discipline is now in the ADR amendment where it will be read.
- **The pre-existing `sends and awaits the turn in one PTY resolution` test's one-stamp synchronisation.** Same latent hang, not this PR's, and touching it would put an unrelated change in a ticket branch. Worth its own line in a follow-up if it is ever seen failing.
- **`packages/panel` shows the abandoned reason nowhere.** Out of scope by the ticket (#484 owns that package this wave); the Session still moves to `needs-input`, which the Panel already renders, and `session:promptAbandoned` is on the wire for it to adopt.

Ready for another look.

mrfqcentic (approved):
## Re-review of #487 — approved

Reviewed at `a02c3b4` against base `fix/prompt-delivery-and-send-path` (now `c009e10`). Both blockers are fixed, and I verified each at the enforcing line and by running the tests rather than from the reply. The Core half is byte-identical to the revision I already passed — `harness-prompt-delivery.ts`, its test file, `session-command.ts`, `core-entry.ts` and `core-link-frames.ts` are all unchanged by `a02c3b4` — so AC1–AC5 cannot have regressed, and I re-ran them anyway.

---

### Blocker 1 — fixed, and the fix is where it needed to be

**1(b), ordering on the Core.** `packages/core/src/pty-manager.ts:732` now appends the `session:promptAbandoned` row and `:742` raises `dialog-unanswered` after it. That is the right way round for exactly the reason the comment gives: the status is the event that resolves `waitForTurnEnd` synchronously on the client, so a reason behind it is a reason nobody reads.

**1(a), the deaf window on the client.** The latch is no longer built in `wrap()`. `openPromptAbandonLatch` (`packages/cli/src/session-gateway.ts:819`) is opened at the top of `attached()` (`:552`) and at the top of `begin()` (`:667`) — before `subscribeEvents`, before `createTask`/`spawn`, before `findByTask`/`ptySubscribe`/`replay`/`seedStatus`. `wrap()` only arms it (`:932`). Registration order inside the latch is right too: `onEvent` at `:863` and `onEventsReplayed` at `:871` are both installed **before** `client.subscribeEvents()` at `:881`, so the replay this asks for cannot outrun the listener meant to judge it. Every throw path closes the latch — `:565`, `:573`, `:633`, `:684` — and `wrap`'s `dispose` closes it at `:955`; I walked `attached()`, `begin()`, `start()` and `resume()` and found no path that returns or throws with a listener still on the client.

**Empirically.** The three #483 in-process tests, 12 consecutive isolated runs: **12/12 green** (was 1/5). Full `packages/cli`, three runs: `50 passed | 1 skipped`, `1221 passed | 3 skipped` — matching your numbers exactly on two of three, with the third hitting the pre-existing flake discussed below. `packages/core` 980/980, `harness-prompt-delivery.test.ts` 88/88, `tsc --noEmit` clean on `sdk`, `core` and `cli`. CI on `a02c3b4` is 18 success / 6 skipped / **0 failures**.

### Blocker 2 — fixed, and the floor cannot be beaten by an early event

This was the thing I most wanted to break, since a floor that can be established late is the same class of hole as the latch that listened late. It holds:

- **Nothing is dropped before the floor is known.** `session-gateway.ts:863-869`: an event arriving while `!armed || floor() === null` is **pushed onto `held`**, not discarded. The only `return` that discards is inside `consider` (`:845`), and `consider` is reached only once a floor exists.
- **Whichever input arrives last triggers the re-judgement.** `drain()` (`:856`) is guarded on `armed && floor() !== null` and is called from both `arm()` (`:884-889`) and the `eventsReplayed` handler (`:876`). Marker-then-arm and arm-then-marker both end in a drain; neither ordering leaves a held event unjudged.
- **`consider` re-checks the floor** (`:844-845`) rather than trusting the caller, so the live path and the drained path apply the same test.
- **The floor cannot admit history.** `floor()` (`:836-839`) is `max(cursor, replayedThrough ?? 0)` when stamped, else `replayedThrough`. On the cursored path the cursor is the delivery stamp, appended *after* the subscribe, so every replayed row is necessarily below it — I checked `handleSubscribe` (`pty-core-link-server.ts:1958-1981`) sends `eventsReplayed` with the last id it actually delivered, after the tail, so live rows are always above the marker. Taking the larger of the two closes the case where a cursor predates the replay.
- **Only the first marker counts** (`:875`). Correct: a reconnect's replay would otherwise push the floor forward over live events this command is entitled to.

The two regression tests are the right two, and they test different floors rather than the same one twice — the cursored case leans on the stamp, the uncursored one has nothing but the marker. Both pass on every run I made.

### Minor — the exit-path row is exactly what you describe

`pty-manager.ts:815-826`: guarded on `promptDelivery && p.taskId` and on `currentPhase` being neither `delivered` nor `abandoned`, it appends `reportPromptAbandoned(...)` with `reason: "the harness exited before the prompt was delivered"` and **no `reportOutputSignal`** — the only `dialog-unanswered` in the file is still the one at `:742`. The status stays the exit's to write, as claimed. It sits before `promptDelivery?.dispose()`, which is necessary, since `dispose()` would set the phase to `abandoned` and close the guard. A shell session or a Session with no starting prompt has no `promptDelivery` and is unaffected.

---

## The three deferrals

**1. `session start` without `--wait` still says nothing about delivery — legitimately out of scope, and AC6 is met.** At the moment a bare `start` returns, the Core has not attempted delivery: the quiet gap has not elapsed, let alone the 15 s backstop or opencode's 90 s ceiling. There is no outcome to report yet, and making one exist means changing when `start` returns — which is #129 D6 and #395's decision, and the one boundary this PR was told to leave alone. Reporting on every path where a caller is still present to be told is what AC6 asks for, and that is now true of `start --wait`, `resume --wait`, `wait` and `send --wait`. A caller who hung up gets the Session's `needs-input` status and a durable `session:promptAbandoned` row, both readable afterwards. **This deferral does not leave AC6 unmet.**

**2. The pre-existing one-stamp flake in `sends and awaits the turn in one PTY resolution` — legitimately out of scope, and I confirmed it is not yours.** I ran that file eight times on the base branch at `c009e10`: **one timeout in eight**, same test, same 60 s shape. On this PR's head it was one in six. Same rate, same cause, present without your change. Fixing an unrelated test inside a ticket branch would be the wrong call. Worth its own issue rather than a follow-up line, since it will keep costing someone a red CI run, but it is not this PR's to carry. **Does not touch AC6.**

**3. `packages/panel` does not surface the reason — legitimately out of scope.** The issue's own Scope section says "`packages/core`. Does not touch `packages/panel`", #484 owns that package this wave, and no acceptance criterion names the Panel. The Session still moves to `needs-input`, which the Panel already renders, and `session:promptAbandoned` is on the wire for it to adopt. **Does not leave AC6 unmet** — AC6 names `session start` and `session send`, both of which are CLI verbs.

## Two observations, neither blocking

**a. The latch depends on having caused the subscribe, and there is no fallback if it did not.** `session-gateway.ts:881` only subscribes `if (!client.isSubscribedToEvents())`. If it is ever handed a client that already subscribed — `DurableCoreClient` does so at `durable-core-client.ts:175` — no `eventsReplayed` marker follows, `replayedThrough` stays `null` for the life of the command, and on the uncursored paths `floor()` never becomes non-null: held events are never judged and `promptAbandoned()` silently answers `null`. That is a false success again, in the same shape. It is **unreachable today** — `openSessionGateway` (`:335`) builds a fresh non-durable `CoreClient` per gateway, `eventsSubscribed` is only ever set by `subscribeEvents`, and nothing in a `session` command subscribes before the latch — so I am not holding the PR for it. The doc comment names the dependency; a cheap guard (or a comment on `openSessionGateway` saying the client must not arrive subscribed) would keep it true, and #395 is precisely the ticket likely to introduce a shared client.

**b. A bare `session wait` that attaches *after* the abandon reports a clean settle.** The row is below the marker, so it is history by the latch's rule, and your own test asserts this deliberately. It is the right trade against Blocker 2 and it is outside AC6's verbs, so it is not a defect — but it is a shape worth knowing: the loss is then visible only as `needs-input` plus the event-log row, not as a non-zero exit. A generation marker (rows since this Session's last `pty:spawn`) would let a bare wait be both stale-safe and truthful, if it ever matters.

## Regression checks on what was already met

| Check | Result |
|---|---|
| No blind type at the generic deadline for a markered harness | Unchanged — `harness-prompt-delivery.ts` byte-identical; guard still at `:1253-1256` ahead of `:1258`; 88/88 tests green |
| Never-arriving marker → `abandoned` → `needs-input` | Unchanged; the ordering swap at `pty-manager.ts:732`/`:742` strengthens it and changes no status |
| opencode ceiling in the override table | `:761` `opencode: { composerWaitMs: 90_000 }`, unchanged |
| **Generic backstop for markerless harnesses still unchanged** | `:612` `composerOnScreen` still returns `true` on an empty marker list; `maxWaitMs` still `15_000` at `:720`; default `composerWaitMs` still `15_000` at `:724`. No harness but opencode waits longer |
| `Refs #483`, no `Closes` | One `Refs #483`; zero `Closes`/`Fixes #`/`Resolves` in the body or any of the three commits |
| Base branch | Still `fix/prompt-delivery-and-send-path` |
| `packages/sdk/src/core-session.ts` untouched (#405) | Untouched |
| `packages/panel` untouched (#484) | Untouched |
| Merges cleanly against the moved base | Base advanced by `c009e10` (#488, `Refs #484`), which touches six `packages/panel` files only — **zero overlap** with this PR's eleven. `git merge-tree` produces a tree with no conflicts; GitHub reports `MERGEABLE` / `CLEAN`. For the record, no #405 commit is on this base yet — `packages/sdk/src/core-session.ts` is untouched there too — so nothing has moved under this PR's feet |

Also worth saying: correcting the "`pnpm test` green" claim in the body, in place and with the reason it was wrong, is the right way to handle that.

## Checklist

| # | Criterion | Verdict |
|---|---|---|
| 1 | No blind type at the generic deadline for a markered harness | **Met** |
| 2 | Never-arriving marker → `abandoned` → `needs-input`, never `delivered` | **Met** |
| 3 | Per-harness ceiling in the override table, opencode ≈ 90 s | **Met** |
| 4 | Generic backstop for markerless harnesses unchanged | **Met** |
| 5 | Unit coverage: marker early / late-but-inside-ceiling / never | **Met** |
| 6 | `session start` / `session send` surface the abandoned outcome | **Met** |
| 7 | `Refs #483`, no `Closes` | **Met** |
| 8 | Base is `fix/prompt-delivery-and-send-path` | **Met** |
| — | Blocker 1 (report reaches the caller) | **Fixed** — 12/12 |
| — | Blocker 2 (stale row cannot fail a healthy send) | **Fixed** — floor verified against early events |
| — | Minor (`dispose()` silent) | **Fixed** — reason-only row from the exit handler |
| — | `packages/cli` green | **Yes** (1221 passed / 3 skipped; one pre-existing flake also present on base) |
| — | `packages/core` green | **Yes** (980/980) |
| — | CI green on `a02c3b4` | **Yes** (18 success, 6 skipped, 0 failures) |
| — | Boundaries (#405, #484, #395) | **Held** |
| — | Merges cleanly against the moved base | **Yes** |

## Verdict

**Approved.** All eight acceptance criteria are met. Both blockers are fixed at the right layer rather than papered over — the ordering on the Core, the latch lifetime and the floor on the client — and the floor logic survives the attack I raised it against: events that arrive before the floor exists are held and re-judged, not dropped, and no ordering of arm/marker leaves one unread. The two observations above are for a follow-up, not for this PR. The reply's account of what was changed matched the code on every point I checked.

The draft was left as a draft; marking it ready is yours.

Issue #406, opened 2026-09-02
P1: compose ?step=redeem is stripped by setup/login


mrfqcentic:
Hunt 2026-09-02 against `actana/control` `main` `44f1119` (beta/0.4.4). Sources: `.scratch/operator-visible-bugs.md` and `panel-rendering-investigation-2026-09-02.md`. **Do not implement from this filing** — issues only.

**Implementation priority: later** — filed so the hunt is complete. Do not start until asked.

- Hunt ID: `P1`
- Severity: broken
- First-fix owner: panel
- Live on pairdemo: code-read (Panel UI was not clicked in the catalog hunt)
- Source: operator-visible-bugs.md

## What the operator sees

Open `http://localhost:7420/?step=redeem` as documented. Create the Operator / sign in. Wizard starts at install, not redeem.

## Where

`panel-auth.ts` redirects to `/setup` with no query. `login.tsx` `window.location.assign("/")`. Likely a regression of #372 framed pairing / Compose redeem jump.

## What to do

Preserve `step=redeem` (and the rest of the pairing query) across setup and login.

## Acceptance

1. A first-run compose URL with `?step=redeem` lands on redeem after Operator create/sign-in.

mrfqcentic:
Shipped in v0.4.5 (https://github.com/actana/control/releases/tag/v0.4.5): merged via the beta/0.4.5-f1 gate #495, promoted to main at d7eeb65 through #529.

PR #488, opened 2026-09-03
fix(panel): a returning page reads the Session's current status

qcentic-dev:
## What this is

Symptom **W2**: a Session that finishes while the operator is looking at
another page still reads `running` when they navigate back, and only a full
browser refresh corrects it. The Core is right the whole time — `actana session
ls` says `finished` for the same Session at the same moment.

`packages/panel` only. Nothing under `packages/core` is touched.

## Root cause

**`packages/panel/src/router.tsx:198` — `staleTime: 30_000` on the query
client's defaults.**

Navigating away unmounts the project route, which is the only thing holding
`useTasks`. Navigating back mounts it again, and react-query's default
`refetchOnMount: true` refetches only a query it considers **stale**. Inside
those 30 seconds the query is fresh by that definition, so the observer
subscribes to the cached, pre-finish list and no read ever reaches the Core.
The row keeps the last status this tab happened to hear.

**`packages/panel/src/router.tsx:200` — `refetchOnWindowFocus: false`** is the
second half of the same cause, and the one that owns AC 2: a tab that is
backgrounded while the Session finishes and then refocused never asks either,
whatever the staleness.

Nothing else covers the gap. `useCoreLiveQueries` — the hook that turns a
Core's events into invalidations — is mounted at
`packages/panel/src/routes/projects.$id.tsx:278`, inside the very route that is
unmounted, and it watches only the project whose page is open. While the
operator is away, this tab is not even watching that Core.

**`packages/panel/src/lib/use-events.ts:58` (pre-change) — `es.onerror`** is a
real second contributor, and the one AC 3 names. The handler shut the socket
and opened a fresh `EventSource` with no cursor, no `Last-Event-ID` and no
invalidation, and the server behind it
(`packages/panel/src/server/controllers/events.controller.ts:16`) writes no
`id:` lines and keeps no event log. An event emitted during the gap was not
delayed, it was gone.

### Suspect 3 is not a cause

The scope-cancel guard from #381 (`useScopedToVisibleProject`,
`packages/panel/src/queries/index.ts:368`) cancels an in-flight project read
when the last viewer leaves. It does **not** block the return path, and this
was confirmed by running it rather than reasoned about: `cancelQueries` with
`revert: true` restores the query's pre-fetch state — `isInvalidated` included
— so a cancelled read leaves the query stale and the next mount refetches. The
third test in `return-to-project-refetch.test.tsx` holds a read open, leaves
mid-flight, comes back, and asserts the fresh status arrives, so the guard is
not blamed (or "fixed") the next time this area is touched.

## The fix

1. **`packages/panel/src/queries/index.ts`** — the three project-scoped reads
   (the project row, its task list, its archived rows) take
   `refetchOnMount: "always"` and `refetchOnWindowFocus: "always"` via a shared
   `PROJECT_SCOPED_FRESHNESS`. Deliberately narrow: the client defaults stay as
   they are for usage rollups, provider usage and the update check, which set
   their own long staleness and should keep it. A Session's status is a live
   fact about a machine, not a cached page.

   The cost is one list read on mount and on focus, against a Core that is
   already asked for that same list on every `pty:` event while a Session runs.

2. **`packages/panel/src/lib/use-events.ts`** — the shared stream now tracks
   whether it has been down since it was last up, and announces the open that
   follows a drop. Only that one: a first connection is not a gap. Reconnect
   subscribers live in their own set, so they never open or hold a socket.

3. **`packages/panel/src/lib/use-event-stream-reconcile.ts`** (new) turns that
   announcement into an invalidation of the `["projects"]` tree (which by
   prefix reaches the list, every project row and every task bucket beneath
   them) plus groups and the archived buckets that sit outside that tree by
   design (ADR 0019). Mounted once in the shell
   (`packages/panel/src/routes/__root.tsx:265`) because the stream is shared
   and so is everything it feeds. This is the same bargain
   `useCoreLiveQueries` already makes for a dropped core-link.

Nothing tries to recover the missed events — the channel never recorded them.
It re-reads instead of trusting the screen.

## Tests

- `packages/panel/src/queries/__tests__/return-to-project-refetch.test.tsx` —
  three cases, each built on a client with `router.tsx`'s exact defaults: a
  remount well inside the 30s window shows the finish; a refocus re-reads with
  the client default off; and the #381 cancel path still refetches on return.
  The first two fail on this branch with the `PROJECT_SCOPED_FRESHNESS` spread
  removed, which is how they were checked for vacuity.
- `packages/panel/src/lib/__tests__/server-events-reconnect.test.tsx` — five
  cases: the first connection announces nothing; a drop-then-open announces
  once; a stream that fires no `onopen` still announces on its first frame; the
  reconnect drives the three invalidations; and an unsubscribed listener is
  told nothing.

Full `packages/panel` suite: 168 files, 1783 tests, green. `pnpm typecheck`
clean across the workspace. `pnpm lint` reports no error or warning from any
file in this change.

Refs #484

https://claude.ai/code/session_01GLqzv97JwsmdFEFigxsbiT

### DIFF
--- a/packages/panel/src/lib/__tests__/server-events-reconnect.test.tsx
+++ b/packages/panel/src/lib/__tests__/server-events-reconnect.test.tsx
@@ -1,0 +1,190 @@
+// @vitest-environment jsdom
+//
+// The SSE channel after a gap (issue 484, symptom W2 — suspect 2).
+//
+// `/api/events` is fire-and-forget: the server keeps no event log, sends no
+// `id:` lines, and the client sends no `Last-Event-ID`. So a `task:updated`
+// emitted while the socket is down is not delayed, it is gone — a backgrounded
+// tab whose connection a browser or proxy reaped comes back believing whatever
+// it last heard, which for a Session that finished in the gap is `running`.
+//
+// The channel cannot replay. What it can do is say that a gap happened, so the
+// queries it feeds are re-read instead of trusted — the same bargain
+// `useCoreLiveQueries` already makes for a dropped core-link. That is what
+// these tests pin: the announcement, its once-per-gap shape, and the
+// invalidation it drives.
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { act, renderHook } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import type { ReactNode } from "react";
+
+const SSE_RECONNECT_DELAY_MS = 1500;
+
+/** Every EventSource the module has opened, oldest first. */
+const opened: FakeEventSource[] = [];
+
+class FakeEventSource {
+  onopen: (() => void) | null = null;
+  onmessage: ((msg: { data: string }) => void) | null = null;
+  onerror: (() => void) | null = null;
+  closed = false;
+  constructor(readonly url: string) {
+    opened.push(this);
+  }
+  close(): void {
+    this.closed = true;
+  }
+  /** The stream drops, the way a reaped connection does. */
+  drop(): void {
+    this.onerror?.();
+  }
+}
+
+const {
+  __resetServerEventsForTests,
+  useServerEvents,
+  useServerEventsReconnect,
+} = await import("~/lib/use-events");
+const { useEventStreamReconcile } = await import("~/lib/use-event-stream-reconcile");
+
+/** Stable identities: `useServerEvents` re-subscribes when its handler changes,
+ *  and a fresh closure per render would tear the shared socket down and back up
+ *  between renders, which is not the thing under test. */
+function noop(): void {}
+
+function wrapperFor(client: QueryClient) {
+  return ({ children }: { children: ReactNode }) => (
+    <QueryClientProvider client={client}>{children}</QueryClientProvider>
+  );
+}
+
+/** Drive the module's own backoff to the next connection attempt. */
+async function waitOutBackoff(): Promise<void> {
+  await act(async () => {
+    vi.advanceTimersByTime(SSE_RECONNECT_DELAY_MS);
+    await Promise.resolve();
+  });
+}
+
+describe("a reconnected SSE stream reconciles the gap it left (issue 484)", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    opened.length = 0;
+    __resetServerEventsForTests();
+    vi.stubGlobal("EventSource", FakeEventSource);
+  });
+  afterEach(() => {
+    __resetServerEventsForTests();
+    vi.unstubAllGlobals();
+    vi.useRealTimers();
+  });
+
+  it("says nothing on the first connection — that is not a gap", async () => {
+    const reconnected = vi.fn();
+    renderHook(() => {
+      useServerEvents(noop);
+      useServerEventsReconnect(reconnected);
+    });
+    await act(async () => {
+      await Promise.resolve();
+    });
+    expect(opened).toHaveLength(1);
+    opened[0]!.onopen?.();
+    expect(reconnected).not.toHaveBeenCalled();
+  });
+
+  it("announces the reconnection that follows a drop, once", async () => {
+    const reconnected = vi.fn();
+    renderHook(() => {
+      useServerEvents(noop);
+      useServerEventsReconnect(reconnected);
+    });
+    await act(async () => {
+      await Promise.resolve();
+    });
+    const first = opened[0]!;
+    first.onopen?.();
+
+    first.drop();
+    expect(first.closed).toBe(true);
+    await waitOutBackoff();
+
+    expect(opened).toHaveLength(2);
+    const second = opened[1]!;
+    act(() => second.onopen?.());
+    expect(reconnected).toHaveBeenCalledTimes(1);
+
+    // The events that follow are ordinary traffic, not further gaps.
+    act(() => second.onmessage?.({ data: JSON.stringify({ type: "task:updated" }) }));
+    expect(reconnected).toHaveBeenCalledTimes(1);
+  });
+
+  it("announces on the first frame when the implementation fires no onopen", async () => {
+    const reconnected = vi.fn();
+    renderHook(() => {
+      useServerEvents(noop);
+      useServerEventsReconnect(reconnected);
+    });
+    await act(async () => {
+      await Promise.resolve();
+    });
+    opened[0]!.onmessage?.({ data: JSON.stringify({ type: "hello", at: 1 }) });
+    opened[0]!.drop();
+    await waitOutBackoff();
+
+    act(() => opened[1]!.onmessage?.({ data: JSON.stringify({ type: "hello", at: 2 }) }));
+    expect(reconnected).toHaveBeenCalledTimes(1);
+  });
+
+  it("re-reads the projects, groups and archived buckets on the way back", async () => {
+    const client = new QueryClient({
+      defaultOptions: { queries: { staleTime: 30_000, refetchOnWindowFocus: false } },
+    });
+    const invalidate = vi.spyOn(client, "invalidateQueries").mockResolvedValue(undefined);
+    renderHook(
+      () => {
+        useServerEvents(noop);
+        useEventStreamReconcile();
+      },
+      { wrapper: wrapperFor(client) },
+    );
+    await act(async () => {
+      await Promise.resolve();
+    });
+    opened[0]!.onopen?.();
+    expect(invalidate).not.toHaveBeenCalled();
+
+    opened[0]!.drop();
+    await waitOutBackoff();
+    act(() => opened[1]!.onopen?.());
+
+    const keys = invalidate.mock.calls.map(([filters]) => filters?.queryKey);
+    // `["projects"]` is a prefix: it reaches the list, every project row and
+    // every task bucket beneath them. The archived buckets sit outside that
+    // tree by design (ADR 0019), so they are named separately.
+    expect(keys).toContainEqual(["projects"]);
+    expect(keys).toContainEqual(["groups"]);
+    expect(keys).toContainEqual(["core-archived-tasks"]);
+  });
+
+  it("tells a subscriber nothing once it has unsubscribed", async () => {
+    const reconnected = vi.fn();
+    const view = renderHook(
+      ({ subscribed }: { subscribed: boolean }) => {
+        useServerEvents(noop);
+        useServerEventsReconnect(subscribed ? reconnected : noop);
+      },
+      { initialProps: { subscribed: true } },
+    );
+    await act(async () => {
+      await Promise.resolve();
+    });
+    opened[0]!.onopen?.();
+    view.rerender({ subscribed: false });
+
+    opened[0]!.drop();
+    await waitOutBackoff();
+    act(() => opened[1]!.onopen?.());
+    expect(reconnected).not.toHaveBeenCalled();
+  });
+});
--- a/packages/panel/src/lib/use-event-stream-reconcile.ts
+++ b/packages/panel/src/lib/use-event-stream-reconcile.ts
@@ -1,0 +1,35 @@
+import { useCallback } from "react";
+import { useQueryClient } from "@tanstack/react-query";
+import { useServerEventsReconnect } from "~/lib/use-events";
+import { queryKeys } from "~/queries";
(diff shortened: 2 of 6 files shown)


qcentic-dev:
## Acceptance criteria — issue #484

- [x] **Navigating away from a project while a Session runs, letting it finish,
      and navigating back shows `finished` with no browser refresh.**
      `refetchOnMount: "always"` on the project row, its task list and its
      archived rows (`packages/panel/src/queries/index.ts`,
      `PROJECT_SCOPED_FRESHNESS`). Covered by
      `return-to-project-refetch.test.tsx` → "shows the finish after a remount
      well inside the 30s stale window", which fails without the change.

- [x] **The same holds after the tab has been backgrounded and refocused.**
      `refetchOnWindowFocus: "always"` on the same three reads, overriding the
      client default at `packages/panel/src/router.tsx:200`. Covered by
      "re-reads when the tab is refocused, with the client default off", which
      also fails without the change.

- [x] **The same holds across an SSE reconnect: events that occurred during the
      gap are not silently lost.** The stream now announces the open that
      follows a drop, and `useEventStreamReconcile` invalidates the
      `["projects"]` tree, groups and the archived buckets on that
      announcement. Covered by `server-events-reconnect.test.tsx` (five cases).
      Note the scope: nothing replays the missed events — the server keeps no
      event log and emits no `id:` lines, so there is nothing to replay from.
      The gap is *reconciled* by re-reading, which is the same bargain
      `useCoreLiveQueries` already makes for a dropped core-link. If literal
      replay is wanted, that is a server-side event log and a separate ticket.

- [x] **The root cause is named in the PR description with the enforcing
      `file:line`, not guessed at.** `packages/panel/src/router.tsx:198`
      (`staleTime: 30_000`) with `:200` (`refetchOnWindowFocus: false`) as the
      second half, plus `packages/panel/src/lib/use-events.ts:58` (pre-change
      `es.onerror`) as the independent contributor behind AC 3. Suspect 3 —
      the #381 scope-cancel guard at
      `packages/panel/src/queries/index.ts:368` — was **ruled out by running
      it**, not by argument: `cancelQueries` with `revert: true` restores the
      pre-fetch state including `isInvalidated`, so a cancelled read leaves the
      query stale and the return path refetches. A test pins that.

- [x] **Regression coverage for the return path.**
      `packages/panel/src/queries/__tests__/return-to-project-refetch.test.tsx`
      (3 cases, on a client built with `router.tsx`'s exact defaults) and
      `packages/panel/src/lib/__tests__/server-events-reconnect.test.tsx`
      (5 cases). The two return-path cases were checked for vacuity by removing
      the fix and confirming they go red.

### Scope note

`packages/panel` only, as the ticket says. Nothing under `packages/core` is
touched — `harness-prompt-delivery.ts` (#483) and `core-session.ts` (#405) are
untouched by this branch.

mrfqcentic (approved):
## Review — PR #488 against issue #484

Reviewed by reading `fix/panel-stale-status-on-return` against
`fix/prompt-delivery-and-send-path` at the line, and by running the suite from a
read-only checkout. Nothing below is taken from the PR description.

### Did the author find the root cause, or just make the symptom go away?

**Found it.** I read all three suspects before looking at the diff, and my
reading agrees with the PR's.

- **Suspect 1 is the enforcing cause of the reported reproduction.**
  `router.tsx:198` (`staleTime: 30_000`) with react-query's default
  `refetchOnMount` — which refetches only a query it considers *stale* — means a
  remount inside the 30s window subscribes to the cached, pre-finish list. The
  citations are exact: `staleTime` is `:198` and `refetchOnWindowFocus: false`
  is `:200` on the base branch. The decisive detail is that `useCoreLiveQueries`
  — the only thing turning a Core's events into invalidations — is mounted at
  `routes/projects.$id.tsx:278`, *inside* the route that unmounts, so while the
  operator is away this tab is not watching that Core at all. No replay on any
  channel could have fixed that path; only a read on return does.

- **Suspect 2 is a genuine second contributor**, on the different vector where
  the page stays mounted and the socket is reaped. `use-events.ts:58`
  (pre-change) closed and reopened telling nobody. The PR covers it rather than
  stopping at the cheap fix.

- **Suspect 3 is correctly exonerated, with evidence rather than assertion.**
  `cancelQueries` with `revert: true` restores the pre-fetch state including
  `isInvalidated`, so the return path still refetches. Confirmed independently:
  the third test in `return-to-project-refetch.test.tsx` passes both with *and*
  without the `PROJECT_SCOPED_FRESHNESS` spread — exactly what "not a cause"
  should look like — and it is pinned so #381 is not blamed again.

Two suspects contribute; the PR covers both, not only the cheapest.

### Acceptance criteria

**AC 1 — met.** `queries/index.ts` gives `projectQueryOptions` (`:84`),
`tasksQueryOptions` (`:185`) and `archivedTasksQueryOptions` (`:242`)
`refetchOnMount: "always"`; nothing later in those option objects overrides the
spread. `routes/projects.$id.tsx:274` reads the board via `useTasks`, and the
detail pane's `useTask` spreads the same options, so row *and* detail are both
covered — the issue reported both. Proven, not assumed: deleting the three
spreads in a scratch copy makes the remount test fail with
`expected 'running' to be 'finished'`.

**AC 2 — met.** `refetchOnWindowFocus: "always"` on the same three. The refocus
test also fails with the spreads removed, so it tests the fix and not the
harness.

**AC 3 — met.** The stream tracks `streamWasLost` (`use-events.ts:102`),
announces the open that follows a drop (`:74`, with a first-frame fallback at
`:87`), and `useEventStreamReconcile` turns that into invalidations of
`["projects"]`, `["groups"]` and `["core-archived-tasks"]`, mounted once at
`routes/__root.tsx:265`. `["projects"]` is a prefix, so it does reach
`tasksCacheKey(projectId, coreId)`. This is reconciliation, not mount-only
refetching. Three of the five SSE tests fail if I remove `streamWasLost = true`
from `onerror`, so the mechanism is what is under test. The socket is genuinely
held app-wide — `__root.tsx:304` mounts `useSessionFinishNotifications`, which
calls `useServerEvents` unconditionally — so the announcement always has a live
stream to ride. See note 1 for a transport correction to the description.

**AC 4 — met.** `router.tsx:198` and `:200` are named and both exact on base.
The secondary citations check out too: `use-events.ts:58` was the old
`es.onerror`, `events.controller.ts:16` is the frame writer that emits no `id:`,
and `queries/index.ts:368` is `useScopedToVisibleProject` on base.

**AC 5 — met, and non-vacuous.** Verified by running: `packages/panel` is
**167 files / 1783 tests, all green**, `tsc --noEmit` exits 0, and the new tests
fail as described once the fix is reverted.

### Boundaries

- No `packages/core` — all six files are `packages/panel/src`. ✅
- `server/panel-auth.ts` and `routes/login.tsx` (#406, on
  `fix/pairing-query-across-auth`) untouched; the two branches share **no** files. ✅
- Base is `fix/prompt-delivery-and-send-path` — not `main`, not a beta train.
  Two commits sitting directly on the base head `8ff5552`. ✅
- Body carries `Refs #484`; no `Closes`/`Fixes`/`Resolves` for any issue. ✅

### Notes — none blocking

1. **`lib/use-event-stream-reconcile.ts:11` — the transport claim is wrong, and
   so is the "no event log" claim.** Two separate corrections:
   (a) `/api/events` carries only Panel-DB rows. `events.emit` is called from
   exactly four Panel services, and `use-core-live-queries.ts:9-13` says so
   outright — a Core Session's `task:updated` never rides this stream. Since the
   report is a Core Session (`actana session ls` answered), the docstring's
   "a Session that finishes while the socket is down" is not the case this
   transport covers; the freshness options cover it. The fix still holds — an
   SSE drop is a sound proxy for "this tab was offline", and the invalidation
   re-reads over the core-link — but the comment mis-models the wiring.
   (b) "The server keeps no event log" is not true: `server/event-log-recorder.ts:32`
   appends **every** `AppEvent` to the `event_log` table precisely so a
   reconnecting Panel can replay, and `shared/event-log.ts:121` serves that
   replay from a cursor for the core-link. The accurate statement is that the
   *browser SSE endpoint* neither reads that log nor emits `id:` lines. As
   written it tells the next reader replay is impossible when the machinery is
   one join away.

2. **`queries/index.ts:84` — the presentation prune now fires far more often.**
   The Core branch ends in a fire-and-forget `api.pruneProjectPresentation`
   (`:116`), and `findProjectPresentationOrphans`
   (`server/repositories/project-presentation.repo.ts:76`) treats an **empty**
   live-id list as "everything is an orphan". To be precise about the risk: the
   catastrophic empty-list case cannot fire here, because `:112`
   (`if (!snapshot) throw`) runs *before* the prune — an empty `projectsList`
   errors the query first. What remains is a partial list that still contains
   the open project: the other projects' pins, rail slots, group filing and card
   images are deleted irrecoverably. That hazard is pre-existing, but this PR
   takes it from "at most once per 30s" to "every mount and every focus". A
   cheap belt-and-braces guard would be to skip the prune when the answer looks
   truncated.

3. **`lib/use-events.ts:102` — `onerror` also fires for a failed *initial*
   connect** (EventSource errors while `readyState === CONNECTING`), so the flag
   is set before the stream was ever up, contradicting the comment's "a first
   connection is not mistaken for a gap". At cold start — renderer up before the
   local server is listening — the retry's open announces a reconnect and
   invalidates the whole projects tree, duplicating reads that just finished.
   Gating on "we have had at least one successful open" would match the comment.

4. **`lib/use-events.ts:107` — `closeIfIdle()` does not set `streamWasLost`,**
   so a deliberate close is a gap that never gets reconciled. This is reachable:
   `dispatchNormalizedFinish` depends on `[toastEnabled, osEnabled, soundEnabled,
   router]` (`use-session-finish-notifications.tsx:429`), so toggling a
   notification setting changes `handler`'s identity, the effect cleanup drops
   the only listener and closes the socket, and the re-run reopens it announcing
   nothing. Narrow window, but the flag currently has two owners and only one
   sets it.

5. **`queries/index.ts:185` — one more window for the optimistic-session
   clobber.** `projects.$id.tsx:805` deliberately `cancelQueries` before
   `appendOptimisticTask` (`:822`), but a refetch that *starts* after the append
   is undefended: a `visibilitychange` while `mutateTaskForCore` is in flight
   lets the Core's answer replace the list and the new card vanishes until
   `replaceOptimisticTask` (`:862`). Self-healing, and already reachable through
   `useCoreLiveQueries`' per-event invalidations, so this only adds a trigger.

6. **The enforcing lines are in `router.tsx` but the fix lives at the call
   sites.** Nothing at `router.tsx:198` points forward to
   `PROJECT_SCOPED_FRESHNESS`, so the next project-scoped query added without
   the spread silently re-inherits this bug. A one-line comment closes that loop.

7. Minor: the fourth SSE test asserts against a mocked `invalidateQueries`, so
   it pins the keys rather than the refetch — fine alongside the two
   `return-to-project` tests, which exercise the real read. And two description
   nits: the suite is **167** files, not 168 (1783 tests matches exactly), and
   `routes/__root.tsx` does carry one lint warning (`useThemeLayoutEffect`
   unused, `:99`) — pre-existing on base and untouched here.

The PR is still a **draft**; approving it, not marking it ready.

### Checklist

| Claim / criterion | Status |
|---|---|
| AC 1 — away → finish → back shows `finished`, no refresh | ✅ Met |
| AC 2 — same after backgrounded + refocused | ✅ Met |
| AC 3 — same across an SSE reconnect; gap events reconciled | ✅ Met |
| AC 4 — root cause named with enforcing `file:line` | ✅ Met (`router.tsx:198`, `:200` — both exact) |
| AC 5 — regression coverage that fails on the old code | ✅ Met (verified by reverting) |
| Root cause found, not symptom suppressed | ✅ Suspect 1 enforcing; 2 also covered; 3 exonerated with a test |
| `Refs #484`, no `Closes` | ✅ |
| Base is `fix/prompt-delivery-and-send-path` | ✅ Not `main`, not a beta train |
| No `packages/core` touched | ✅ |
| `panel-auth.ts` / `login.tsx` (#406) untouched | ✅ No file overlap at all |
| Suite green | ✅ 167 files / 1783 tests; `tsc` clean |

### Verdict

**Approve.** The author found the real cause, proved it at the line, covered the
second contributor instead of stopping at the cheapest one, exonerated the third
with a test rather than an assertion, and the regression coverage genuinely
fails on the old code. Notes 1–4 are worth folding in before this rides a train;
none of them blocks.

PR #490, opened 2026-09-03
fix(panel): the pairing query survives the auth round trip

qcentic-dev:
Opening `http://localhost:7420/?step=redeem` — the URL `deploy/docker-compose.yml` prints — and then creating the Operator or
signing in landed the operator on the wizard's **install** step. The pairing query did not survive the auth round trip.

## The destroyers

`FirstRunWizard` reads `window.location.search` once, at mount. Everything between the operator's click and that mount rebuilt
the URL from a bare path, so by the time the wizard looked, there was nothing to read.

| Hop | Before | After |
| --- | --- | --- |
| `server/panel-auth.ts` — `documentAuthRedirect` | `/?step=redeem` → `/setup`, `/login`; and back to `/` | query carried on every 303 |
| `routes/setup.tsx` — Operator create | `window.location.assign("/")` | `assign(withCarriedQuery("/", …))` |
| `routes/login.tsx` — sign-in | `window.location.assign("/")` | `assign(withCarriedQuery("/", …))` |
| `lib/api.ts` — 401 under an open tab | `assign("/login")` | query carried too |

The issue names the first two; the third and fourth are the same round trip's other legs. `api.ts` is included because a session
that expires under an open `/?step=redeem` should come back to `/?step=redeem` — it is one line and the same bug.

**Both exits are fixed, not one.** Creating the Operator and signing in are separate pages with separate handlers, and the
acceptance criterion names both. The first-boot Compose link is aimed precisely at the Panel that has no Operator yet, so a fix
that only survived sign-in would miss the operator it was written for.

## No side channel

The query rides the redirect itself. Nothing is written to `localStorage`, a cookie, or any other store: there is nothing to
expire, nothing for a second tab to inherit, and nothing left behind by an abandoned sign-in. The redirect can simply keep it,
so it keeps it.

New shared helper, `lib/auth-paths.ts` → `withCarriedQuery(pathname, search)`. **Only the query travels** — the path is always
one this code chose, so this cannot become an open redirect. Re-serialising through `URLSearchParams` is what makes the value
safe in a `Location` header as well as in `location.assign`: a CR or LF smuggled through the query comes out percent-encoded
rather than ending the header line. Both suites pin that.

## Is it a regression of #372?

**No — a different one.** #372 is where `?step=` and the Compose link came from, and it is easy to read this as its regression,
but the history says otherwise:

- `git log --diff-filter=A` puts both destroyers — `panel-auth.ts`'s bare `redirectTo` and `login.tsx`'s
  `window.location.assign("/")` — in `c2b8d0d`, the initial commit, 2026-08-03.
- #372 merged 2026-09-01 as `97cd450` and touched **neither file**: `actana-pair.ts`, `FirstRunWizard.tsx`,
  `core-onboarding.ts`, `docker-compose.yml` and their tests, and nothing else.

Nothing that worked before stopped working. The auth gate has never carried a query; #372 shipped the first link whose query
anyone needed to survive, and the two had simply never met. #372's own tests are honest about their scope — they exercise
`firstRunStepFromSearch` and the wizard directly, never through a sign-in — which is exactly the seam this PR repairs.

## Only onto a route it means something on (review B1)

The carry keeps the query and drops the path, so an unfiltered version replays a **deep route's** parameters onto a shallower
one. `coreId` is the live case: `routes/projects.$id.tsx` is the only route that declares it, and `__root.tsx` reads it off
every route under a comment stating the invariant — *"Only the /projects/$id route sets `coreId`; every other route implicitly
means the Panel's own rows."* An expiry on `/projects/p1?coreId=core-b` came home as `/?coreId=core-b`, which scopes the whole
root shell to a Core the operator never navigated to and arms `UserTerminalPanel`'s New Terminal button — disabled on a clean
`/` precisely because there is no Core to open a shell on. #382 round 2 removed a fallback of that shape from `ProjectBar`.

`withCarriedQuery` now drops a parameter whose owning route the destination is not. `ROUTE_SCOPED_PARAMS` names the one the
Panel has today beside the route that owns it, so the rule reads as a fact about `coreId` rather than a list of destinations —
and a parameter still travels onto the route it belongs to (`/projects/p1?coreId=c` keeps it).

The narrowing is in the helper alone: no redirect restructured, no call site changed. The pairing query is untouched — `step`
belongs to no route, so nothing filters it and nothing names it.

## Coverage

Every changed file is covered, and every suite fails on the code before its change (verified by reverting, not asserted):

| Suite | Cases | Reverting | Result |
| --- | --- | --- | --- |
| `server/__tests__/operator-auth.test.ts` | 6 | `panel-auth.ts` | **5 failed / 23 passed** |
| `lib/__tests__/pairing-query-across-auth.test.tsx` | 15 | `login.tsx` + `setup.tsx` | **4 failed / 6 passed** |
| `lib/__tests__/api-401-redirect.test.ts` | 5 | `api.ts` | **1 failed / 4 passed** |
| all three | — | the B1 narrowing | **6 failed / 43 passed** |

The page tests assert the landing URL *as the wizard reads it* (`firstRunStepFromSearch`), not as a string. The B1 narrowing
fails one case per way in: the gate, both auth pages, the 401 leg, and the helper.

`api-401-redirect.test.ts` is new in the second commit and answers review N1 — `lib/api.ts` previously had no test, while this
description implied otherwise. It drives the real client against a stubbed 401: the step survives, a bare route stays bare, a
deep route's parameters do not travel, the login page does not redirect to itself, and a 500 does not redirect at all. It is a
separate file because `pairing-query-across-auth.test.tsx` mocks `~/lib/api` wholesale.

Both page suites live under `lib/__tests__` rather than `src/routes/__tests__` because the route generator warns on any file
under `src/routes/` that does not export a `Route`; the pages are imported by alias, so the folder changes nothing about what
is tested.

## Verification

- `packages/panel` full suite: **167 files, 1801 tests, all passing.**
- `pnpm typecheck` clean across all five workspaces; route generation clean.
- `pnpm lint` — 0 errors; the 43 warnings are pre-existing and none are in a file this PR touches.

## Boundaries

`packages/panel/src/router.tsx` and `lib/use-events.ts` (#484) untouched. Nothing under `packages/core` (#483, #405, #290)
touched. `router.tsx:65` has a fifth bare `assign("/")` — an error-boundary "go home" — left alone as #484's file and not part
of this path. `OperatorSettingsPage`'s sign-out `assign("/login")` is deliberately left bare: leaving is not a round trip.

Refs #406

https://claude.ai/code/session_011Doi96KzL7VATynvKaE8Up

### DIFF
--- a/packages/panel/src/lib/__tests__/api-401-redirect.test.ts
+++ b/packages/panel/src/lib/__tests__/api-401-redirect.test.ts
@@ -1,0 +1,102 @@
+// @vitest-environment jsdom
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+/**
+ * The third leg of the auth round trip: a session that dies under an open tab.
+ *
+ * `documentAuthRedirect` never sees it — the browser is already on the page and
+ * the next `/api/*` call comes back 401 — so `redirectToLoginOnce` is the only
+ * thing that decides where that operator lands, and it was the one behaviour
+ * changed by #406 with nothing exercising it (#490 review N1).
+ *
+ * It has its own file because `pairing-query-across-auth.test.tsx` mocks
+ * `~/lib/api` wholesale; this suite needs the real module and a stubbed
+ * `fetch` under it.
+ */
+
+/** Where the last `window.location.assign` was pointed, or null. */
+let assigned: string | null = null;
+let realLocation: PropertyDescriptor | undefined;
+
+function browserAt(href: string): void {
+  const url = new URL(href, "http://panel.example.test");
+  Object.defineProperty(window, "location", {
+    configurable: true,
+    value: {
+      href: url.href,
+      pathname: url.pathname,
+      search: url.search,
+      assign: (target: string) => {
+        assigned = target;
+      },
+    },
+  });
+}
+
+/** Answer every call with one status, so `req` takes the branch under test. */
+function respondWith(status: number): void {
+  vi.stubGlobal(
+    "fetch",
+    vi.fn(async () => new Response(JSON.stringify({ error: "unauthorized" }), { status })),
+  );
+}
+
+const { api, ApiError } = await import("~/lib/api");
+
+/** Drive one gated call and swallow the `ApiError` the failure raises. */
+async function callTheApi(): Promise<void> {
+  await expect(api.listCores()).rejects.toBeInstanceOf(ApiError);
+}
+
+beforeEach(() => {
+  assigned = null;
+  realLocation = Object.getOwnPropertyDescriptor(window, "location");
+});
+
+afterEach(() => {
+  if (realLocation) Object.defineProperty(window, "location", realLocation);
+  vi.unstubAllGlobals();
+  vi.clearAllMocks();
+});
+
+describe("a 401 under an open tab", () => {
+  it("brings the operator back to the step they were on", async () => {
+    respondWith(401);
+    browserAt("/?step=redeem");
+    await callTheApi();
+    expect(assigned).toBe("/login?step=redeem");
+  });
+
+  it("goes to a bare login page when there was nothing to carry", async () => {
+    respondWith(401);
+    browserAt("/");
+    await callTheApi();
+    expect(assigned).toBe("/login");
+  });
+
+  /**
+   * The same narrowing the gate and the two pages apply (#490 review B1): the
+   * expiry happens on `/projects/$id`, the operator is sent to `/login`, and
+   * `coreId` means nothing there or on the `/` they sign in to.
+   */
+  it("leaves a deep route's own parameters behind", async () => {
+    respondWith(401);
+    browserAt("/projects/p1?coreId=core-b");
+    await callTheApi();
+    expect(assigned).toBe("/login");
+  });
+
+  it("stays put when the 401 arrives on the login page itself", async () => {
+    respondWith(401);
+    browserAt("/login?step=redeem");
+    await callTheApi();
+    expect(assigned).toBeNull();
+  });
+
+  it("does not redirect on any other failure", async () => {
+    respondWith(500);
+    browserAt("/?step=redeem");
+    await callTheApi();
+    expect(assigned).toBeNull();
+  });
+});
--- a/packages/panel/src/lib/__tests__/pairing-query-across-auth.test.tsx
+++ b/packages/panel/src/lib/__tests__/pairing-query-across-auth.test.tsx
@@ -187,6 +187,57 @@ describe("creating the Operator", () => {
   });
 });
 
+/**
+ * The carry keeps the query and drops the path, so an unfiltered version lands
+ * a deep route's parameters on a shallower one (#490 review B1).
+ *
+ * `coreId` is the live case. `routes/projects.$id.tsx` is the only route that
+ * declares it, and `__root.tsx` reads it off every route — its selector's own
+ * comment being "Only the /projects/$id route sets `coreId`; every other route
+ * implicitly means the Panel's own rows". Replayed onto `/` after a re-login it
+ * scopes the root route to a remote Core and arms `UserTerminalPanel`'s New
+ * Terminal button, which is disabled on a clean `/` for exactly the reason that
+ * there is no Core to open a shell on.
+ */
+describe("a deep route's parameters", () => {
+  it("do not follow a sign-in home to the root route", async () => {
+    browserAt("/login?coreId=core-b");
+    await signIn();
+    expect(assigned).toBe("/");
+  });
+
+  it("do not follow an Operator create home either", async () => {
+    browserAt("/setup?coreId=core-b");
+    await createOperator();
+    expect(assigned).toBe("/");
+  });
+
+  it("are dropped without taking the pairing query with them", async () => {
+    browserAt("/login?step=redeem&coreId=core-b&label=my-panel");
+    await signIn();
+    const url = new URL(assigned!, "http://x");
+    expect(url.searchParams.get("coreId")).toBeNull();
+    expect(url.searchParams.get("step")).toBe("redeem");
+    expect(url.searchParams.get("label")).toBe("my-panel");
+    expect(firstRunStepFromSearch(url.search)).toBe(REDEEM_STEP);
+  });
+
+  it("survive on the route that owns them", () => {
+    // Not a path this module redirects to today — pinned so the rule reads as
+    // "carried where it means something", not "always deleted".
+    expect(withCarriedQuery("/projects/p1", "?coreId=core-b")).toBe("/projects/p1?coreId=core-b");
+  });
+
+  it("are dropped on every destination the auth round trip uses", () => {
+    for (const destination of ["/", "/login", "/setup"]) {
+      expect(withCarriedQuery(destination, "?coreId=core-b")).toBe(destination);
+      expect(withCarriedQuery(destination, "?step=redeem&coreId=core-b")).toBe(
+        `${destination}?step=redeem`,
+      );
+    }
+  });
+});
+
 describe("withCarriedQuery", () => {
   it("leaves a destination with no query alone", () => {
     expect(withCarriedQuery("/", "")).toBe("/");
--- a/packages/panel/src/lib/__tests__/pairing-query-across-auth.test.tsx
+++ b/packages/panel/src/lib/__tests__/pairing-query-across-auth.test.tsx
@@ -1,0 +1,214 @@
+// @vitest-environment jsdom
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { ComponentType } from "react";
+import { act, cleanup, fireEvent, render, screen } from "@testing-library/react";
+
+/**
+ * The pairing query survives the auth round trip (#406).
+ *
+ * `deploy/docker-compose.yml` prints `…/?step=redeem`, and an operator opening
+ * it on a fresh Panel is bounced through `/setup` or `/login` before anything
+ * reads it. `FirstRunWizard` picks its starting step out of
+ * `window.location.search` once, at mount — so every hop that rebuilds the URL
+ * from a bare path silently turns that documented link back into step 1.
+ *
+ * Both exits from the round trip are covered here, deliberately. Creating the
+ * Operator and signing in are different pages with different handlers, the
+ * acceptance criterion names both, and fixing either one alone leaves the other
+ * broken for exactly the operator the link was written for: the one arriving at
+ * a Panel that has never been set up.
+ *
+ * The assertion is the landing URL *as the wizard reads it* — `?step=redeem`
+ * put back through `firstRunStepFromSearch` — rather than a string compare, so
+ * this suite fails if the query survives in a shape the wizard cannot parse.
+ *
+ * It lives beside `~/lib/auth-paths` rather than under `src/routes/`, whose
+ * every file the route generator expects to export a `Route`; the two pages are
+ * imported by alias, so the folder makes no difference to what is tested. The
+ * server half of the same round trip is in
+ * `server/__tests__/operator-auth.test.ts`, where the Panel DB harness lives.
+ */
+
+class ApiError extends Error {
+  constructor(
+    message: string,
+    public readonly status: number,
+    public readonly body: unknown,
+  ) {
+    super(message);
+    this.name = "ApiError";
+  }
+}
+
+const api = {
+  login: vi.fn(async (_password: string) => ({ ok: true })),
+  setupOperator: vi.fn(async (_body: { name: string; password: string }) => ({ ok: true })),
+};
+
+vi.mock("~/lib/api", () => ({ api, ApiError }));
+
+const { withCarriedQuery } = await import("~/lib/auth-paths");
+const { FIRST_RUN_STEP_IDS, REDEEM_STEP_LINK, firstRunStepFromSearch } = await import(
+  "~/shared/core-onboarding"
+);
+const loginRoute = await import("~/routes/login");
+const setupRoute = await import("~/routes/setup");
+
+const REDEEM_STEP = FIRST_RUN_STEP_IDS.indexOf("redeem");
+
+/** The page component a route file hands the router, ready to render alone. */
+function pageOf(route: { options: unknown }): ComponentType {
+  const component = (route.options as { component?: ComponentType }).component;
+  if (!component) throw new Error("route has no component");
+  return component;
+}
+
+/** Where the last `window.location.assign` was pointed, or null. */
+let assigned: string | null = null;
+let realLocation: PropertyDescriptor | undefined;
+
+/**
+ * Put the browser on `href` with a `location` that records navigations instead
+ * of performing them. jsdom's own `assign` is unforgeable and does nothing, so
+ * the destination — the whole subject of these tests — is unobservable without
+ * standing in for the object.
+ */
+function browserAt(href: string): void {
+  const url = new URL(href, "http://panel.example.test");
+  Object.defineProperty(window, "location", {
+    configurable: true,
+    value: {
+      href: url.href,
+      pathname: url.pathname,
+      search: url.search,
+      assign: (target: string) => {
+        assigned = target;
+      },
+    },
+  });
+}
+
+beforeEach(() => {
+  assigned = null;
+  realLocation = Object.getOwnPropertyDescriptor(window, "location");
+});
+
+afterEach(() => {
+  cleanup();
+  if (realLocation) Object.defineProperty(window, "location", realLocation);
+  vi.clearAllMocks();
+});
+
+/** Fill a labelled box on the rendered card. */
+function type(label: RegExp, value: string): void {
+  fireEvent.change(screen.getByLabelText(label), { target: { value } });
+}
+
+async function submit(buttonLabel: string): Promise<void> {
+  await act(async () => {
+    fireEvent.click(screen.getByRole("button", { name: buttonLabel }));
+  });
+}
+
+async function signIn(): Promise<void> {
+  const LoginPage = pageOf(loginRoute.Route);
+  render(<LoginPage />);
+  type(/^password$/i, "operator-password");
+  await submit("Sign in");
+}
+
+async function createOperator(): Promise<void> {
+  const SetupPage = pageOf(setupRoute.Route);
+  render(<SetupPage />);
+  type(/your name/i, "Operator");
+  type(/^password$/i, "a-long-enough-password");
+  type(/confirm password/i, "a-long-enough-password");
+  await submit("Create Operator");
+}
+
+describe("signing in", () => {
+  it("lands the operator back on the step the compose link asked for", async () => {
+    browserAt(`/login${REDEEM_STEP_LINK}`);
+    await signIn();
+    expect(api.login).toHaveBeenCalledOnce();
+    expect(assigned).not.toBeNull();
+    expect(firstRunStepFromSearch(new URL(assigned!, "http://x").search)).toBe(REDEEM_STEP);
+  });
+
+  it("carries the rest of the pairing query, not just the step", async () => {
+    browserAt("/login?step=redeem&label=my-panel&session=7f2c1a9e");
+    await signIn();
+    const search = new URL(assigned!, "http://x").searchParams;
+    expect(search.get("step")).toBe("redeem");
+    expect(search.get("label")).toBe("my-panel");
+    expect(search.get("session")).toBe("7f2c1a9e");
+  });
+
+  it("still goes to a bare root when there was nothing to carry", async () => {
+    browserAt("/login");
+    await signIn();
+    expect(assigned).toBe("/");
+  });
+});
+
+describe("creating the Operator", () => {
+  it("lands the operator back on the step the compose link asked for", async () => {
+    browserAt(`/setup${REDEEM_STEP_LINK}`);
+    await createOperator();
+    expect(api.setupOperator).toHaveBeenCalledOnce();
+    expect(assigned).not.toBeNull();
+    expect(firstRunStepFromSearch(new URL(assigned!, "http://x").search)).toBe(REDEEM_STEP);
+  });
+
+  it("carries the rest of the pairing query, not just the step", async () => {
+    browserAt("/setup?step=redeem&label=my-panel");
+    await createOperator();
+    const search = new URL(assigned!, "http://x").searchParams;
+    expect(search.get("step")).toBe("redeem");
+    expect(search.get("label")).toBe("my-panel");
+  });
+
+  it("still goes to a bare root when there was nothing to carry", async () => {
+    browserAt("/setup");
+    await createOperator();
+    expect(assigned).toBe("/");
+  });
+
+  it("stays put when the two passwords disagree", async () => {
+    browserAt(`/setup${REDEEM_STEP_LINK}`);
+    const SetupPage = pageOf(setupRoute.Route);
+    render(<SetupPage />);
+    type(/your name/i, "Operator");
+    type(/^password$/i, "a-long-enough-password");
+    type(/confirm password/i, "something-else-entirely");
+    await submit("Create Operator");
+    expect(api.setupOperator).not.toHaveBeenCalled();
+    expect(assigned).toBeNull();
+  });
+});
+
+describe("withCarriedQuery", () => {
+  it("leaves a destination with no query alone", () => {
+    expect(withCarriedQuery("/", "")).toBe("/");
+    expect(withCarriedQuery("/login", "?")).toBe("/login");
+  });
+
+  it("keeps every parameter, in order", () => {
+    expect(withCarriedQuery("/login", "?step=redeem&label=my-panel")).toBe(
+      "/login?step=redeem&label=my-panel",
+    );
+  });
+
+  /**
+   * The value goes into a `Location` header as well as `location.assign`, so a
+   * CR or LF smuggled through the query has to come out escaped rather than
+   * splitting the response.
+   */
+  it("re-encodes a query that would otherwise break a header", () => {
+    const carried = withCarriedQuery("/login", "?step=redeem%0d%0aSet-Cookie:%20x=1");
+    expect(carried).not.toMatch(/[\r\n]/);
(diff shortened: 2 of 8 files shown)


qcentic-dev:
## Acceptance criteria — issue #406

Issue #406 states one numbered criterion. The lines below it are the "What to do" and "Where" clauses of the same issue, and
the extra constraints this implementation was asked to hold to; each is listed so nothing is checked off by implication.

**AC1 — A first-run compose URL with `?step=redeem` lands on redeem after Operator create/sign-in.** — [x] **met**
Both exits verified by test, at the level the wizard actually reads:
`firstRunStepFromSearch(new URL(assigned).search) === FIRST_RUN_STEP_IDS.indexOf("redeem")`.
- create: `lib/__tests__/pairing-query-across-auth.test.tsx` → "creating the Operator > lands the operator back on the step the
  compose link asked for"
- sign-in: same file → "signing in > lands the operator back on the step the compose link asked for"
- and the hops before them: `server/__tests__/operator-auth.test.ts` → "the pairing query > rides out to setup on first boot" /
  "rides out to login once an Operator exists" / "rides home again for a browser that is already signed in"

**`panel-auth.ts` redirects to `/setup` with no query — fixed.** — [x] **met**
`documentAuthRedirect` now reads `url.search` and passes it to `redirectTo`, which builds the `Location` with
`withCarriedQuery`. All five redirect sites carry it: `/`→`/setup`, `/`→`/login`, `/setup`→`/login`, `/setup`→`/`, `/login`→`/`.

**`login.tsx` calls `window.location.assign("/")` — fixed.** — [x] **met**
`assign(withCarriedQuery("/", window.location.search))`.

**Preserve the query across BOTH setup and login, not just one.** — [x] **met**
`setup.tsx` fixed in the same commit as `login.tsx`. Six of the ten new page tests are split evenly across the two, including a
bare-`/` case on each so the fix cannot degenerate into always appending a `?`.

**The rest of the pairing query, not only `step`.** — [x] **met**
The whole query is carried, not an allow-list of one parameter. Pinned by "carries the rest of the pairing query, not just the
step" on both pages (`label`, `session`) and by "keeps the rest of the query with it" on the server gate.

**No side channel — `localStorage` or otherwise — if the redirect can keep it.** — [x] **met**
It can, so it does. No storage of any kind is written or read: the query travels on the redirect and on `location.assign`, and
this PR adds no new state. Nothing to expire, nothing for a second tab to inherit, nothing for an abandoned sign-in to leave
behind. No justification is owed because no side channel was used.

**Regression coverage that fails on the current code.** — [x] **met**
Verified by reverting each file in turn and re-running, not by assertion:
- `panel-auth.ts` reverted → `operator-auth.test.ts` **5 failed / 23 passed**
- `login.tsx` + `setup.tsx` reverted → `pairing-query-across-auth.test.tsx` **4 failed / 6 passed**

**Say whether this is the same regression as #372.** — [x] **met** (answered: **it is not**)
Section "Is it a regression of #372?" in the PR description. Both destroyers date to `c2b8d0d`, the initial commit
(2026-08-03); #372 merged as `97cd450` (2026-09-01) and touched neither file. #372 shipped the first link whose query needed to
survive; the gate had never carried one.

**Parallel-ticket boundaries.** — [x] **respected**
`packages/panel/src/router.tsx` and `packages/panel/src/lib/use-events.ts` (#484): untouched. Anything under `packages/core`
(#483, #405, #290): untouched. The seven files in this PR are listed in the diff and none is on either list.

### Explicitly *not* done, with the reason

- **`router.tsx:65` — a fifth bare `window.location.assign("/")`.** — [ ] **not done, deliberately.** It is the error
  boundary's "go home" button, not a leg of the auth round trip, and `router.tsx` is #484's file this wave. Out of bounds and
  out of scope.
- **`OperatorSettingsPage.tsx:47` — `assign("/login")` on sign-out.** — [ ] **not done, deliberately.** Signing out is a
  departure, not a round trip; carrying a pairing query into it would be wrong, not merely unnecessary.
- **A browser click-through on a live Compose stack.** — [ ] **not done.** The issue itself records this bug as `code-read` —
  the Panel UI was never clicked in the hunt — and the environment here has no Compose stack to bring up. The behaviour is
  covered instead at both ends of the round trip by the fifteen tests above, including the wizard's own parser as the oracle.

mrfqcentic (changes requested):
## Review — PR #490 against issue #406

Reviewed as a reviewer only: nothing in the branch was edited. Every claim below was checked at the line and, where it is a
behavioural claim, re-proved by running the suites against reverted copies of the changed files in a scratch tree outside the repo.

**Verdict: request changes.** Everything issue #406 asks for is met and proved — both exits, the whole query, no side channel, real
regression coverage. The one blocker is a side effect the widened carry has on a surface the issue never mentioned.

---

### The half-fix trap: both destroyers, proved by reverting each one

The issue names two destroyers and its criterion names two operator paths, so a fix that repairs one path passes a manual test and
stays broken. Both are fixed here, and each one is independently load-bearing:

- **`server/panel-auth.ts`** — `documentAuthRedirect` now reads `url.search` (`panel-auth.ts:113`) and every one of the five redirect
  sites passes it (`:122`, `:125`, `:126`, `:129`) into `redirectTo`, which builds the `Location` through `withCarriedQuery`
  (`panel-auth.ts:87-90`). Reverting **this file alone** → `operator-auth.test.ts` **5 failed / 23 passed**.
- **`routes/login.tsx:32`** — the bare `window.location.assign("/")` is now `assign(withCarriedQuery("/", window.location.search))`.
  Reverting **this file alone** → `pairing-query-across-auth.test.tsx` **2 failed / 8 passed** (both failures under `signing in`).
- **`routes/setup.tsx:40`** — the same change on the *other* exit. Reverting **this file alone** → **2 failed / 8 passed** (both under
  `creating the Operator`).

Reverting login + setup together reproduces the 4 failed / 6 passed the description claims. Nothing here is carried by one path only.

**So: an operator who CREATES an Operator and an operator who SIGNS IN both land on redeem.** The chain closes end to end —
`FirstRunWizard.tsx:98` picks its step from `firstRunStepFromSearch(window.location.search)` in a `useState` initialiser, the wizard is
`lazy`-mounted by `FirstRunGate` only after a client-side registry read (so no SSR read of `window`), and `routes/index.tsx` declares no
`validateSearch`, so nothing between the redirect and the mount strips the query.

### The other claims, each checked at the line

1. **The whole query survives, not just `step`.** `auth-paths.ts:31` re-serialises the entire incoming search; there is no allow-list and
   `step` is nowhere hardcoded. Pinned on all three legs (`label`, `session`) by "carries the rest of the pairing query, not just the
   step" on both pages and "keeps the rest of the query with it" on the gate. **Met.**
2. **Carried by the redirect, not a side channel.** No `localStorage`, `sessionStorage`, cookie or module-level stash is added or read;
   the value exists only in the `Location` header and the `assign` argument. Nothing to justify, because nothing was smuggled. **Met.**
3. **Regression coverage that fails on the old code.** Proved by reverting, not asserted — the three counts above. The tests also assert
   the landing URL through the wizard's own parser rather than by string compare, so they fail if the query survives in a shape the
   wizard cannot read. **Met** (one gap, see note 1 below).
4. **#372.** Independently confirmed: both destroyers date to `c2b8d0d`, the initial commit (2026-08-03) — `redirectTo` built a bare
   `location` there, and both pages already ended in `assign("/")`. #372 merged as `97cd450` (2026-09-01) touching
   `deploy/docker-compose.yml`, `packages/cli/src/actana-pair.ts`, `FirstRunWizard.tsx`, `core-onboarding.ts` and their tests — neither
   `panel-auth.ts` nor either page. **I agree with the PR: not a regression of #372.** #372 shipped the first link whose query needed to
   survive and walked into a gap that had been there a month.
5. **`Refs #406`, no `Closes`.** Present at line 73 of the body; no `closes`/`fixes`/`resolves` anywhere in it, in any case. **Met.**
6. **Base branch.** `fix/pairing-query-across-auth` → `fix/prompt-delivery-and-send-path`, one commit ahead of that branch's tip
   (`8ff5552`). Not `main`, not a `beta/*`. **Met.**

### Boundaries

`router.tsx` and `lib/use-events.ts` (#484, live on `fix/panel-stale-status-on-return`) are untouched; that branch also carries
`__root.tsx`, `queries/index.ts` and `use-event-stream-reconcile.ts`, and this PR touches none of them either — zero overlap. Nothing
under `packages/core`. All seven files are under `packages/panel`. **Honoured.**

---

## Blocking finding

**`packages/panel/src/routes/login.tsx:32` (and `routes/setup.tsx:40`, `lib/api.ts:113`, fed by `server/panel-auth.ts:129`) — the query
is carried to a path it did not come from, which puts a deep route's parameters on the root route.**

`login.tsx` always goes to `/`. For the pairing case the query and the destination agree, because `/?step=redeem` came from `/`. For any
other route they do not, and the path is dropped while its query is kept.

Concretely, and reachable on an ordinary day:

1. The operator is on a Core-owned pin — `ProjectBar.tsx:1359`/`:1385` navigate to `/projects/$id?coreId=core-b`.
2. The session expires. Either a reload hits the gate → `panel-auth.ts:129` now answers `303 /login?coreId=core-b` (it answered a bare
   `/login` before), or the open tab takes a 401 → `api.ts:113` assigns `/login?coreId=core-b`.
3. Sign in → `login.tsx:32` assigns `withCarriedQuery("/", "?coreId=core-b")` → **`/?coreId=core-b`**.

`routes/__root.tsx:316` reads `state.location.search.coreId` on *every* route — its own comment two lines above states the invariant this
breaks: "Only the /projects/$id route sets `coreId`; every other route implicitly means the Panel's own rows." With that param on `/`:

- `__root.tsx:683` → `UserTerminalPanel coreId="core-b"`, so the "New Terminal" button, disabled on a clean `/` by
  `UserTerminalPanel.tsx:338` (`disabled={!active || !coreId}`), becomes live and opens a VM shell on a remote Core;
- `__root.tsx:455` → ⌘T does the same, `createVmShellTerminal(routeCoreId)`, with no route check;
- `__root.tsx:651`/`:375` → the rail and picker are scoped to that Core, and `ProjectBar.tsx:1650`'s `initialCoreId` falls back to it, so
  the edit dialog browses that Core's disk for a Panel-owned project.

None of this is destructive and each needs an operator action, but before this change `/` was always clean, and this codebase has already
been burned by a stray Core scope on a non-Core route — see the comment at `ProjectBar.tsx:1584`, where exactly that fallback was removed
in #382 round 2. It is unmentioned in the description, untested, and outside what #406 asked for.

Note that dropping the `api.ts` leg does not fix it: `panel-auth.ts:129` produces the same landing on a plain reload. The narrowing has to
happen in what travels. One shape that keeps everything #406 demands — the whole pairing query, no hardcoded `step` — is for
`withCarriedQuery` to drop parameters that belong to a route the destination is not (`coreId` being the one the Panel has today) when the
destination is `/`, `/login` or `/setup`, with a test that `/projects/x?coreId=c` comes home as a bare `/`. The shape is yours to choose;
the leak is the thing to close.

## Non-blocking notes

1. **`lib/api.ts` has no regression coverage, and the description implies it does.** "Both suites fail on the code before this change
   (verified by reverting each file in turn and re-running)" reads as covering all four changed files. Reverting `api.ts` alone leaves
   both suites fully green — **38 passed / 38** — because nothing exercises `redirectToLoginOnce`; it is the only changed behaviour in the
   PR with no test. Worth either a case or a more precise sentence.
2. **Local full-suite run:** 166 files / 1790 tests, **1 failed** — `server/panel-link/__tests__/live-read-path.test.ts` timing out at
   5000ms under load. It passes on its own (14/14), touches nothing in this diff, and CI's Unit Tests job is green, so I read it as
   unrelated flake, not a miss in the description's "all passing".
3. `isAuthPath(pathname)` in `api.ts:109` is exactly equivalent to the two literal comparisons it replaced; the new import adds no cycle.
   The header-splitting tests are real: `URLSearchParams` percent-encodes CR/LF, and the assertion is on the header value.

---

## Checklist

| # | Item | Verdict |
| --- | --- | --- |
| 1 | AC1 — `?step=redeem` lands on redeem after Operator **create** | **Met** — proved by revert (`setup.tsx` → 2 failures) |
| 2 | AC1 — `?step=redeem` lands on redeem after **sign-in** | **Met** — proved by revert (`login.tsx` → 2 failures) |
| 3 | `panel-auth.ts` carries the query | **Met** — `:113`, `:122`, `:125`, `:126`, `:129`; revert → 5 failures |
| 4 | `login.tsx` carries the query | **Met** — `login.tsx:32` |
| 5 | Whole pairing query survives, not only `step` | **Met** — `auth-paths.ts:31`, no allow-list |
| 6 | Carried by the redirect, no side channel | **Met** — no storage of any kind added |
| 7 | Regression coverage that fails on the old code | **Met** for the three destroyers; **gap** for `api.ts` |
| 8 | Says whether this is #372's regression, and is right | **Met** — says no; history confirms it, I agree |
| 9 | Body has `Refs #406`, no `Closes` | **Met** |
| 10 | Base is `fix/prompt-delivery-and-send-path` | **Met** |
| 11 | #484's `router.tsx` / `lib/use-events.ts` untouched | **Met** |
| 12 | Nothing under `packages/core` | **Met** |
| 13 | No unintended behaviour change outside the issue | **Not met** — the `coreId` replay above |

The fix for #406 itself is complete, correct and honestly evidenced — the trap this ticket sets was not walked into. Close the query
leak onto the root route and this is an approve.

qcentic-dev:
## Round 2 — reply to the review

Thank you for the revert-by-revert verification; the `2 failed / 8 passed` split per page is a sharper proof than the combined
count I reported, and B1 is a real leak I did not see. Head is now `bdec4cb`, one commit on top of `4380048`. **Nothing from
round 1 was redone or widened** — the #406 fix itself is byte-for-byte unchanged.

### B1 — fixed. The narrowing is in the helper, nowhere else.

You had it exactly: the carry keeps the query and drops the path, so a parameter that only means something on a deep route
arrives on a shallow one. `withCarriedQuery` now drops a parameter whose owning route the destination is not:

```ts
const ROUTE_SCOPED_PARAMS: readonly { readonly name: string; readonly ownedBy: string }[] = [
  { name: "coreId", ownedBy: "/projects/" },
];

export function withCarriedQuery(pathname: string, search: string): string {
  const params = new URLSearchParams(search);
  for (const { name, ownedBy } of ROUTE_SCOPED_PARAMS) {
    if (!pathname.startsWith(ownedBy)) params.delete(name);
  }
  const query = params.toString();
  return query ? `${pathname}?${query}` : pathname;
}
```

I took your "shape is yours to choose" and named the **owning route** rather than the three destinations. Same outcome for
`/`, `/login` and `/setup`, but the rule then states a fact about `coreId` — it belongs to `/projects/$id` — instead of
enumerating where it is unwelcome, and a parameter still travels onto the route that owns it. The doc comment cites
`validateProjectSearch`, `__root.tsx`'s invariant in its own words, and the `ProjectBar` comment from #382 round 2 you pointed
at, so the next person to widen this list has the reason in front of them.

**One production file changed this round, `lib/auth-paths.ts`, and one function in it.** No redirect restructured, no call site
touched, `panel-auth.ts` / `login.tsx` / `setup.tsx` / `api.ts` all identical to `4380048`. Your point that dropping the
`api.ts` leg would not have helped is why: the fix had to be in what travels, and that is the only place it is.

The pairing query is untouched — `step` belongs to no route (the wizard is mounted by `FirstRunGate` over whatever route the
operator landed on), so nothing filters it and nothing names it. No hardcoded `step` was introduced.

**Coverage, one case per way in.** With the narrowing reverted, **6 failed / 43 passed**:

| Where | Case |
| --- | --- |
| `operator-auth.test.ts` | `/projects/p1?coreId=core-b` → `/login`, and `…&step=redeem` → `/login?step=redeem` |
| `pairing-query-across-auth.test.tsx` | sign-in from `/login?coreId=core-b` → `/` |
| same | Operator create from `/setup?coreId=core-b` → `/` |
| same | `?step=redeem&coreId=core-b&label=my-panel` → `coreId` gone, `step` and `label` intact |
| same | `withCarriedQuery("/projects/p1", "?coreId=core-b")` keeps it — the rule is not "always delete" |
| `api-401-redirect.test.ts` | the 401 leg, below |

The gate case is the one you identified as the reason the `api.ts` leg is not the fix — a plain reload reaches the same landing
without it — so it is pinned on the server side too.

### N1 — done, not deferred.

You were right that the description overclaimed. `lib/api.ts` now has `lib/__tests__/api-401-redirect.test.ts`: it drives the
real client against a stubbed 401 and asserts the step survives (`/?step=redeem` → `/login?step=redeem`), a bare route stays
bare, a deep route's parameters do not travel, the login page does not redirect to itself, and a **500** does not redirect at
all. Reverting `api.ts` to its pre-#406 state → **1 failed / 4 passed**. Separate file because
`pairing-query-across-auth.test.tsx` mocks `~/lib/api` wholesale.

The PR description is corrected too: the coverage section is now a per-file table naming which revert produces which failures,
rather than a sentence that read as covering all four.

### N2 — agreed, and it did not recur.

`server/panel-link/__tests__/live-read-path.test.ts` passed in both of my full-suite runs this round (167 files / 1801 tests,
all passing) and in CI. I read it the same way you do: load-dependent flake in a file nothing in this diff touches.

### N3 — noted, no action.

Agreed on all three points; nothing to change.

### Unchanged from round 1

Still a draft, base still `fix/prompt-delivery-and-send-path`, `Refs #406` with no `Closes` anywhere in the body. Nothing
merged, no force push — `4380048..bdec4cb` fast-forwarded. `router.tsx` and `lib/use-events.ts` (#484) untouched, as are
`__root.tsx`, `queries/index.ts` and `use-event-stream-reconcile.ts`; nothing under `packages/core`.

Item 13 on your checklist is the one that moved. Ready for another look.

mrfqcentic (approved):
## Re-review — PR #490, round 2 (head `bdec4cb`)

Reviewer only: nothing in the branch was edited. Everything below was re-checked at the line and, where it is behavioural,
re-proved by running the suites against reverted copies of individual files in a scratch clone outside the repo.

**Verdict: approve.** B1 is closed exactly where it had to be closed, N1 is addressed rather than deferred, and nothing that
was already met has moved.

---

### B1 — closed, and closed in the one place that could close it

`lib/auth-paths.ts:67-74`:

```ts
export function withCarriedQuery(pathname: string, search: string): string {
  const params = new URLSearchParams(search);
  for (const { name, ownedBy } of ROUTE_SCOPED_PARAMS) {
    if (!pathname.startsWith(ownedBy)) params.delete(name);
  }
  const query = params.toString();
  return query ? `${pathname}?${query}` : pathname;
}
```

with `ROUTE_SCOPED_PARAMS = [{ name: "coreId", ownedBy: "/projects/" }]` at `:37-39`.

- **It is a deny-list, not an allow-list.** Only names appearing in `ROUTE_SCOPED_PARAMS` are ever deleted; every other
  parameter is re-serialised untouched. `step` is not named anywhere in the file, and neither is any other pairing parameter —
  pinned by "are dropped without taking the pairing query with them", which sends `?step=redeem&coreId=core-b&label=my-panel`
  through a sign-in and asserts `coreId` gone, `step` **and** `label` intact, and `firstRunStepFromSearch` still reading redeem.
  This was the specific risk in my "shape is yours to choose", and it was not walked into.
- **It is genuinely local.** The round-2 commit touches four files: `lib/auth-paths.ts` and three test files. `panel-auth.ts`,
  `login.tsx`, `setup.tsx` and `api.ts` are byte-identical to `4380048` — confirmed by diffing `4380048..bdec4cb`, not by
  reading the reply. No redirect restructured, no call site changed, no new branch in the gate.
- **It is at the right hop.** The gate is the leg that made dropping `api.ts` insufficient, and the gate goes through
  `withCarriedQuery` at `panel-auth.ts:87-90`, so one edit covers all four legs: `panel-auth.ts:113`/`:122`/`:125`/`:126`/`:129`,
  `login.tsx:32`, `setup.tsx:40`, `api.ts:113`.
- **Naming the owning route rather than the three destinations is the better of the two shapes.** `/projects/p1?coreId=c` keeps
  it (`"survive on the route that owns them"`), so the rule states a fact about `coreId` instead of enumerating where it is
  unwelcome, and the doc comment at `:9-35` hands the next person the `__root.tsx` invariant and the #382-round-2 `ProjectBar`
  precedent in full.

**Re-proved by revert.** With `auth-paths.ts` alone rolled back to `4380048` and the rest of the branch at head:
**6 failed / 43 passed** across the three suites — one failure per way in, exactly as claimed:

| Suite | Failing case |
| --- | --- |
| `operator-auth.test.ts` | `the pairing query > leaves a deep route's own parameters behind` (the gate) |
| `pairing-query-across-auth.test.tsx` | `do not follow a sign-in home to the root route` |
| same | `do not follow an Operator create home either` |
| same | `are dropped without taking the pairing query with them` |
| same | `are dropped on every destination the auth round trip uses` (the helper itself) |
| `api-401-redirect.test.ts` | `leaves a deep route's own parameters behind` (the 401 leg) |

**Completeness for today's Panel, checked rather than taken on trust.** Two routes declare a search schema:
`projects.$id.tsx:150` (`coreId`) and `settings.tsx:21` (`panel`). `coreId` is the only one read off *every* route —
`__root.tsx:319` via `useRouterState`, feeding `UserTerminalPanel` at `:683` and the ⌘T binding. `panel` is read only through
`Route.useSearch()` on `/settings`, which is route-local, so replaying it onto `/` is inert. `coreId` is the whole of the leak,
and the list covers it.

### N1 — addressed, not deferred

`lib/__tests__/api-401-redirect.test.ts` is new (102 lines). It drives the real `~/lib/api` against a stubbed 401 rather than
the wholesale mock the other page suite uses, and covers the step surviving, a bare route staying bare, a deep route's
parameters not travelling, no self-redirect on `/login`, and a 500 not redirecting at all. Rolling `api.ts` back to its
pre-#406 state → **1 failed / 4 passed** (`brings the operator back to the step they were on`). The description's coverage
section is now a per-file table naming which revert produces which failures, so the sentence I objected to is gone too.

### N2 / N3

`live-read-path.test.ts` did not recur: full `packages/panel` run here is **167 files / 1801 tests, all passing**, matching the
description. N3 needed no action.

---

### Nothing already met has regressed

- **Both destroyers still repaired, each independently load-bearing** (re-proved file-by-file at head): revert `panel-auth.ts`
  alone → `operator-auth.test.ts` **6 failed / 23 passed**; revert `login.tsx` alone → `pairing-query-across-auth.test.tsx`
  **3 failed / 12 passed**; revert `setup.tsx` alone → **2 failed / 13 passed**. The counts are one higher than round 1 on the
  first two because round 2 added cases to those suites, not because anything loosened.
- **Both operator paths still land on redeem**: `signing in > lands the operator back on the step the compose link asked for`
  and `creating the Operator > …` both green, asserted through `firstRunStepFromSearch`, not a string compare.
- **The whole pairing query, no hardcoded `step`**: `auth-paths.ts:68` re-serialises everything not route-owned; `label` and
  `session` pinned on all three legs.
- **No side channel**: no `localStorage`, `sessionStorage`, cookie, IndexedDB or module-level stash anywhere in the added
  lines — grepped over the full diff, not just the new file.
- **`Refs #406`, no `Closes`**: body line 99, and no `closes`/`fixes`/`resolves` in any case anywhere in it.
- **Base**: still `fix/prompt-delivery-and-send-path`. Still a draft. `4380048..bdec4cb` fast-forwarded; no force push.
- **#484's boundary**: `router.tsx` and `lib/use-events.ts` untouched, as are `__root.tsx`, `queries/index.ts` and
  `use-event-stream-reconcile.ts`. Nothing under `packages/core`.

### The base has moved, and it still merges clean

#484 landed on the base as **`c009e10`** ("a returning page reads the Session's current status (#488)"), so this branch is one
commit behind. A trial merge of the base into `bdec4cb` in a scratch clone completed with **no conflicts**; the six files it
brings in (`use-events.ts`, `use-event-stream-reconcile.ts`, `queries/index.ts`, `__root.tsx`, and two new suites) have **zero**
overlap with this PR's eight. GitHub agrees: `mergeable: true`, `mergeable_state: clean`.

Worth noting because it is load-bearing for B1: #488 edits `__root.tsx`, and the invariant this narrowing protects survives that
edit intact — post-merge, the comment "Only the /projects/$id route sets `coreId`" and the `useRouterState` selector under it are
still there, still feeding `UserTerminalPanel`.

### CI

Green on the new head `bdec4cb`: 24 checks, all `success` or `skipped` — Unit Tests, Typecheck, Lint, Conventions, Train rules,
both E2E installers, the Panel service seam, both image builds, Secret Scan, Dependency Audit. No failures, none pending.

---

## Checklist

| # | Item | Verdict |
| --- | --- | --- |
| 1 | **B1** — carry narrowed so a deep route's param cannot reach the root route | **Fixed** — `auth-paths.ts:67-74`; revert → 6 failures |
| 2 | B1 fix is local: no redirect restructured, no call site changed | **Confirmed** — round 2 touches 1 production file |
| 3 | B1 fix is a deny-list, not an allow-list; other pairing params still ride | **Confirmed** — `step`/`label` pinned alongside the drop |
| 4 | **N1** — `lib/api.ts` regression coverage | **Addressed** — new suite; revert `api.ts` → 1 failed / 4 passed |
| 5 | Both destroyers still repaired, each independently | **Met** — 6 / 3 / 2 failures on separate reverts |
| 6 | `panel-auth.ts` and `login.tsx` both carry the query | **Met** — `:113`/`:122`/`:125`/`:126`/`:129`, `login.tsx:32` |
| 7 | Operator **create** and **sign-in** both land on redeem | **Met** — both cases green via `firstRunStepFromSearch` |
| 8 | Whole pairing query survives; no hardcoded `step` | **Met** — `step` appears nowhere in `auth-paths.ts` |
| 9 | No `localStorage` or other side channel | **Met** — none in the diff |
| 10 | Regression coverage still fails on the old code | **Met** — every revert re-run above |
| 11 | `Refs #406`, no `Closes` | **Met** — body line 99 |
| 12 | Base is `fix/prompt-delivery-and-send-path` | **Met** |
| 13 | #484's `router.tsx` / `lib/use-events.ts` untouched | **Met** |
| 14 | Merges cleanly over #484 now on the base | **Met** — trial merge clean, zero file overlap |
| 15 | CI green on `bdec4cb` | **Met** — 24/24 success or skipped |
| 16 | No unintended behaviour change outside the issue | **Met** — item 13 of round 1 is closed |

Item 13 was the only thing outstanding and it is closed, in the helper, without touching anything else. Approving.

PR #492, opened 2026-09-03
fix: a wait ends when its link to the Core does, and calls the outcome unknown

qcentic-dev:
`actana session start proj "prompt" --wait` with no `--wait-timeout` hung for ever if the Core blipped mid-wait. `waitForTurnEnd` had no disconnect handler, and every way a wait can end well — the settling status a harness's hooks produce, the process exit behind it — arrives down the core link. A link that drops takes all of them at once, so the wait does not end: it goes quiet, and with no deadline (or with the `--wait-timeout 0` #405 added) nothing left in the world can settle it.

## The decision

**A drop is an ending in its own right, and it rejects.** The cheap fix is the forbidden one: resolving the wait with the status the Session was last seen at reports a turn as ended on the evidence that *this side stopped listening*. That is a false completion arriving through the transport rather than through the harness — the exact failure this train exists to remove — so the wait fails and says the turn's outcome is **unknown**.

Three endings, three classes, so a caller branches without parsing prose:

| ending | means |
| --- | --- |
| resolves | the Core reported a turn ended |
| `CoreSessionTurnTimeoutError` (#405) | this side gave up on a clock it set |
| `CoreSessionLinkLostError` (new) | this side went deaf; the outcome is unknown |

## What changed

**1. `CoreSessionLinkLostError`**, exported from `packages/sdk/src/core-session.ts`, carrying `taskId`, `afterEventId`, `lastStatus`, `reportedSinceDelivery`, `graceMs` and `reason`. `lastStatus` is context and is documented as never the answer. Its message names both readings and settles neither — the turn may have ended in the silence, and it may still be running — for the same reason #405's does: choosing between them would mean reading the screen (#191, ADR 0026 D3) or consulting `reportsTurnStart` (ADR 0033 D2).

**2. `CoreSession` listens on the link**, on both the `start` and the `attach` path (`onDisconnected` / `onReady`), and a wait begun while the link is *already* down gets the same treatment as one that was running when it went.

**3. A grace, not a deadline.** On a client that reconnects a drop is usually not even an interruption: the link returns, the client re-subscribes from its cursor, and the Core streams the tail it missed — including the status this wait was waiting for. So the drop starts a clock that runs **only while the link is down** and is cancelled the moment a connection comes back. `CORE_LINK_LOST_GRACE_MS` is 30s, which is six or more of `DurableCoreClient`'s backoff attempts. It is not `timeoutMs` and does not become one: a caller that passed no deadline still has none while the Core is reachable.

**4. A client with no reconnect coming gets a grace of 0** and fails on the drop itself. A grace is time given to a reconnect, and the `actana` CLI's one-shot `CoreClient` has none — sitting out 30s there would be 30 more seconds of the hang. The new `CoreClient.willReconnect()` predicate decides (false here; `DurableCoreClient` overrides it to true until this client hangs up), and `linkLostGraceMs` on `start`/`attach` overrides that in turn.

**5. `subscribeEvents` records the subscribe only when the frame went out.** It set `eventsSubscribed = true` before `sendNow`, so `isSubscribedToEvents()` answered true for a subscribe no link ever carried. `CoreSession` reads that flag to decide whether it still owes the Core a subscribe: the lie is a Session wired to an event stream nobody asked for, hearing no status ever — another wait nothing can end. Named in the issue's own **Where** section.

**6. The CLI names the next step**, as it already does for the deadline: where the answer is (the Core, once reachable), and `events tail --since <delivery id>` where there is a cursor. **Not `session wait`** — uncursored, it answers from the status the Session was parked at before the drop and exits zero, which is #405's false completion one layer up.

**7. ADR 0033 gains D6**, extending D4. `docs/README.md`'s index row and the `actana-sessions` skill say so, and the skill payload was regenerated from its authored source.

## Acceptance criteria (issue #396)

1. **If the core-link drops during `--wait`, the command exits non-zero (or reconnects and still exits) rather than hanging for ever** — both halves. A client that reconnects inside the grace carries on waiting and ends on the Core's own report; one that does not exits non-zero with `CoreSessionLinkLostError`. Pinned at both layers: the SDK suite drops a live socket under a wait with **no deadline at all** and asserts the rejection, and the CLI suite asserts `EXIT_FAILURE` plus a `--json` document with no `status` key.

## Landmines, and how each is respected

- **No turn is settled from PTY bytes or output quietness** (#191, ADR 0026 D3) — the facts here are the link going down and two event ids.
- **No wait path consults `reportsTurnStart`** (ADR 0033 D2) — the new handlers live below `// ─── Internals`, outside the source-sweep's slice, and consult nothing.
- **An unknown outcome reads as unknown** — the wait rejects; it never resolves with a status the Core did not send.
- **#405 is extended, not undone** (ADR 0033 D5) — `CoreSessionTurnTimeoutError`, the `send`-only default deadline and `--wait-timeout 0` are untouched, and a regression test asserts a turn that really ends still resolves the wait.

## Files this touches, for the tickets landing beside it

- `packages/core/src/harness-prompt-delivery.ts`, `packages/core/src/pty-manager.ts`, `packages/cli/src/session-gateway.ts` — **not touched** (#483).
- the codex hooks code — **not touched** (#290). The codex `HARNESS_READINESS` table and `packages/core/src/__tests__/fixtures/` — **not touched** (#277).
- `packages/panel/**` — **not touched** (#484).
- `packages/cli/src/session-command.ts` — the `awaitTurn` catch only; `sessionSend`'s refusal and the timeout advice #405 added are unchanged.

## Verification

`pnpm -r typecheck`, `pnpm lint` (0 errors; 43 warnings, byte-identical to the base), `pnpm scan:secrets`, and the `sdk`, `cli`, `core` and `shared` unit suites all pass locally. Panel is untouched and left to CI.

Refs #396

https://claude.ai/code/session_013TbQDZV12xtjR55RsQqivh

### DIFF
--- a/.agents/skills/actana-sessions/SKILL.md
+++ b/.agents/skills/actana-sessions/SKILL.md
@@ -149,7 +149,7 @@ object** `start --wait --json` prints — same keys, same `screen` — so one pa
 reads all three verbs. Two of those keys are `null` on a Session you attached to
 rather than started: `command` and `reportsTurnStart` are answers to a spawn.
 
-Four things to know before you build on it:
+Five things to know before you build on it:
 
 1. **Sending into a turn that is already running resolves on *that* turn's
    end.** A keystroke into a busy Harness is not a new turn. If the Session was
@@ -184,6 +184,17 @@ Four things to know before you build on it:
    exits zero** — which reads as a completed turn and is not one. To carry on
    waiting, follow the log from the delivery instead:
    `actana events tail --since <event id>`, the id the timeout message names.
+5. **A wait whose link to the Core drops ends as *unknown*.** Every wait listens
+   over that link, so a Core that restarts or a connection that is reaped takes
+   the report the wait was waiting for with it. Rather than hang — which is what
+   it used to do, with no deadline of any kind to end it — the command exits
+   non-zero saying **the turn's outcome is unknown**: it may have ended while
+   this side was deaf, and it may still be running. That is not a failed turn
+   and not a finished one, and it is the one case where the Session's status has
+   to be re-read from the Core rather than taken from the wait. Do that once the
+   Core is reachable again — `actana session ls` says whether it is still live,
+   `actana events tail --since <event id>` follows the log from your delivery —
+   and **do not** answer it with `session wait`, for the reason in 4.
 
 **Every turn of that loop gets its own report file**, and the turn number in the
 filename is what keeps them apart. See below.
--- a/docs/README.md
+++ b/docs/README.md
@@ -94,7 +94,7 @@ routing around it.
 | [0030](adr/0030-the-panel-is-a-dumb-pipe-for-file-bytes.md) | The Panel is a dumb pipe for file bytes, and it is the end that holds the credentials |
 | [0031](adr/0031-the-product-ships-one-skill.md) | The product ships one skill, and installs it into the operator's home — **PROPOSED**, amends 0006 |
 | [0032](adr/0032-one-actana-cli.md) | There is one `actana`, and it is both the Core manager and the client — supersedes #265 §6 and 0031 D8's two-binaries premise |
-| [0033](adr/0033-turn-end-is-the-one-mandatory-harness-signal.md) | Turn-end reporting is the one mandatory harness signal; turn-start, auto-mode flags and resume ids may never gate a feature — D5 (#405) gives `send --wait` a default deadline, amending D4 |
+| [0033](adr/0033-turn-end-is-the-one-mandatory-harness-signal.md) | Turn-end reporting is the one mandatory harness signal; turn-start, auto-mode flags and resume ids may never gate a feature — D5 (#405) gives `send --wait` a default deadline and D6 (#396) ends a wait whose link dropped as unknown, both amending D4 |
 | [0035](adr/0035-a-second-skill-for-the-sub-agent-role.md) | A second skill for the sub-agent role, and it is eager — **PROPOSED**, depends on 0031 |
 | [0036](adr/0036-the-beta-release-channel.md) | The beta release channel: a beta installs like a release, and the ref is the channel — **PROPOSED**, amends 0016 (D28, D29, D34, D35) and 0023 (D8, D9) |
 | [0037](adr/0037-one-version-per-line.md) | One version per line: where a version string is written, and what is authoritative — **PROPOSED**, amends 0023 (D3, D7), the two clauses 0036 §G left open |
--- a/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
+++ b/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
@@ -92,6 +92,43 @@ A send that submits nothing at all (`--no-enter`) is refused with `--wait`
 rather than bounded: it has no turn to await, so there is nothing for a deadline
 to be about.
 
+**D6 — A lost link ends a wait as *unknown*, and never as an outcome**
+(extends D4, #396). Everything that can end a wait well arrives down the core
+link: the status change a harness's hooks produce, and the process exit behind
+it. A link that drops therefore takes every one of them with it, and the wait
+does not end — it goes quiet. With no `--wait-timeout` there is nothing left in
+the world that can settle it, which is `actana session start … --wait` sitting
+for ever after a Core blip; with one, the operator learns about the drop as a
+deadline, which points them at a slow harness rather than at a dead socket.
+
+So a drop is an ending in its own right, and the class of ending is the whole
+decision. **The wait fails.** Resolving it with the status the Session was last
+seen at is the cheap fix and the forbidden one: it reports a turn as ended on
+the evidence that this side stopped listening, which is a false completion — the
+failure D4 and #405 exist to remove, arriving through the transport instead of
+through the harness. The wait rejects with `CoreSessionLinkLostError`, distinct
+by class from both a resolution (the Core reported a turn ended) and
+`CoreSessionTurnTimeoutError` (this side gave up on its own clock), and its
+message says the outcome is unknown and names both readings — the turn may have
+ended in the silence, and it may still be running.
+
+The wait is not failed on the first blip. A client that reconnects usually
+recovers a drop without the wait noticing: the link returns, the client
+re-subscribes from its cursor, and the Core streams the tail it missed —
+including the status change the wait was waiting for. So a drop starts a
+**grace**, not a deadline: thirty seconds (`CORE_LINK_LOST_GRACE_MS`), which is
+six or more of `DurableCoreClient`'s backoff attempts, cancelled the moment a
+connection comes back, and running only while the link is down. A client that
+does **not** reconnect gets a grace of zero, because a grace is time given to a
+reconnect and a one-shot client — the `actana` CLI's — has none coming; waiting
+it out would be thirty more seconds of the hang.
+
+Nothing here weakens D2 or D4. No screen is read, `reportsTurnStart` is
+consulted nowhere, and the only facts the error carries are that the link went
+and two event ids. The Session itself is untouched by any of it: it is running
+on the Core, which is where its status lives and where the question this error
+leaves open is answered.
+
 ## Consequences
 
 Adding a family is still adding a row, and now the row has a bar to clear. A
--- a/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
+++ b/docs/adr/0033-turn-end-is-the-one-mandatory-harness-signal.md
@@ -117,11 +117,36 @@ recovers a drop without the wait noticing: the link returns, the client
 re-subscribes from its cursor, and the Core streams the tail it missed —
 including the status change the wait was waiting for. So a drop starts a
 **grace**, not a deadline: thirty seconds (`CORE_LINK_LOST_GRACE_MS`), which is
-six or more of `DurableCoreClient`'s backoff attempts, cancelled the moment a
-connection comes back, and running only while the link is down. A client that
-does **not** reconnect gets a grace of zero, because a grace is time given to a
-reconnect and a one-shot client — the `actana` CLI's — has none coming; waiting
-it out would be thirty more seconds of the hang.
+six or more of `DurableCoreClient`'s backoff attempts, and running only while
+the link is down. A client that does **not** reconnect gets a grace of zero,
+because a grace is time given to a reconnect and a one-shot client — the
+`actana` CLI's — has none coming; waiting it out would be thirty more seconds of
+the hang.
+
+Two clauses of that grace are load-bearing, and #492's review found the first
+draft wrong on both. Each was a way for a wait to outlive its link after all,
+which would have left this decision stating something untrue.
+
+**The grace is a cumulative budget, not a fresh allowance per outage.** The
+outages a single wait lives through are summed, and the wait fails once they
+total the grace. Forgiving each flap in full bounds nothing: a Core in a restart
+crash-loop, or a `DurableCoreClient` on its 500 ms–5 s backoff against a flaky
+link, drops and returns indefinitely with no single outage long enough to fire
+anything, and the wait hangs for ever — #396's own bug, reached the long way
+round. Summed, the deaf time a wait can accumulate is bounded, so the wait always
+ends. The error reports both numbers, because they differ: `graceMs` is the
+budget and `downMs` is what was actually spent.
+
+**The grace is forgiven by an *established* connection, not by `ready`.** `ready`
+is the Core's first unsolicited frame and lands before `auth` is answered, so a
+connection the Core is about to refuse for an expired or rotated bearer produces
+one. Nothing can be sent on that connection and no `subscribe` goes out — that
+rides the established signal — so treating `ready` as recovery forgives an outage
+that never ended; with a supervisor dialing again it is a `ready` → `authError` →
+`disconnect` loop that arms and disarms the guard for ever with no events flowing
+at all. `CoreClient.onEstablished` — the Core has said who it is and the link can
+be written to, authenticated where a bearer was configured — is the signal, and
+`onReady` is not.
 
 Nothing here weakens D2 or D4. No screen is read, `reportsTurnStart` is
 consulted nowhere, and the only facts the error carries are that the link went
--- a/packages/cli/src/__tests__/session-command.test.ts
+++ b/packages/cli/src/__tests__/session-command.test.ts
@@ -23,7 +23,10 @@ import {
   type SessionRow,
   type StartedSession,
 } from "../session-gateway.ts";
-import { CoreSessionTurnTimeoutError } from "@actana/sdk/core-session.ts";
+import {
+  CoreSessionLinkLostError,
+  CoreSessionTurnTimeoutError,
+} from "@actana/sdk/core-session.ts";
 import { EXIT_FAILURE, EXIT_OK, EXIT_USAGE } from "../exit-codes.ts";
 
 let fixture: CliFixture | null = null;
@@ -943,6 +946,75 @@ describe("actana session send", () => {
     // #405's false completion back one layer up.
     expect(run.err.join("\n")).toContain("Not `session wait`");
     expect(run.err.join("\n")).toContain("exits zero");
+  });
+
+  it("exits non-zero when the link drops mid-wait, and calls the outcome unknown (#396)", async () => {
+    // Issue #396's acceptance criterion, at the layer the operator meets it:
+    // `--wait` against a Core that blips used to leave the command pending for
+    // ever. It now ends — non-zero, with the outcome named as unknown, and
+    // **not** as a status the Core never sent.
+    await withRegisteredCore();
+    const run = await cli().run(["session", "send", "task_1", "carry on", "--wait", "--json"], {
+      sessions: fakeSessionGateway({
+        sendAndWait: async () =>
+          fakeStartedSession({
+            wait: async () => {
+              throw new CoreSessionLinkLostError({
+                taskId: "task_1",
+                afterEventId: 42,
+                lastStatus: "running",
+                reportedSinceDelivery: true,
+                graceMs: 30_000,
+                reason: "socket hang up",
+              });
+            },
+          }),
+      }),
+    });
+
+    expect(run.code).toBe(EXIT_FAILURE);
+    const payload = JSON.parse(run.out.join("\n"));
+    expect(payload.waited).toBe(true);
+    // No status key at all: the Core reported none, so the document carries
+    // none. A `--json` caller branching on `status` must not find `running`
+    // here and read it as the turn.
+    expect(payload.status).toBeUndefined();
+    expect(payload.error).toContain("outcome is unknown");
+    expect(payload.error).toContain("not a report that the turn finished");
+    expect(run.err.join("\n")).toContain("the turn's outcome is unknown");
+    // Cursored, so the next step is the log from the delivery — and, as after
+    // #405's timeout, explicitly not `session wait`, which would answer from the
+    // status this Session was parked at before the drop and exit zero.
+    expect(run.err.join("\n")).toContain("`actana events tail --since 42`");
+    expect(run.err.join("\n")).toContain("Not `session wait`");
+  });
+
+  it("points an uncursored wait at the Core rather than at a delivery it never made", async () => {
+    // `start --wait` and `session wait` carry no delivery stamp, so there is no
+    // `--since` to follow and the advice that names one would be a fiction.
+    await withRegisteredCore();
+    const run = await cli().run(["session", "start", "web", "go", "--wait"], {
+      sessions: fakeSessionGateway({
+        start: async () =>
+          fakeStartedSession({
+            wait: async () => {
+              throw new CoreSessionLinkLostError({
+                taskId: "task_1",
+                afterEventId: 0,
+                lastStatus: null,
+                reportedSinceDelivery: false,
+                graceMs: 0,
+                reason: null,
+              });
+            },
+          }),
+      }),
+    });
+
+    expect(run.code).toBe(EXIT_FAILURE);
+    expect(run.err.join("\n")).toContain("the turn's outcome is unknown");
+    expect(run.err.join("\n")).toContain("`actana session ls`");
+    expect(run.err.join("\n")).not.toContain("events tail --since");
   });
 
   it("does not offer the uncursored wait as the next step after any timeout", async () => {
--- a/packages/cli/src/__tests__/session-command.test.ts
+++ b/packages/cli/src/__tests__/session-command.test.ts
@@ -965,6 +965,7 @@ describe("actana session send", () => {
                 lastStatus: "running",
                 reportedSinceDelivery: true,
                 graceMs: 30_000,
+                downMs: 31_200,
                 reason: "socket hang up",
               });
             },
@@ -1004,6 +1005,7 @@ describe("actana session send", () => {
                 lastStatus: null,
                 reportedSinceDelivery: false,
                 graceMs: 0,
+                downMs: 0,
                 reason: null,
               });
             },
--- a/packages/cli/src/orchestration-skill-payload.ts
+++ b/packages/cli/src/orchestration-skill-payload.ts
@@ -197,7 +197,7 @@ object** \`start --wait --json\` prints — same keys, same \`screen\` — so on
 reads all three verbs. Two of those keys are \`null\` on a Session you attached to
 rather than started: \`command\` and \`reportsTurnStart\` are answers to a spawn.
 
-Four things to know before you build on it:
+Five things to know before you build on it:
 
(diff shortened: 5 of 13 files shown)


qcentic-dev:
## Acceptance criteria — issue #396

Issue #396 states exactly one acceptance criterion. It is checked below, split into the two readings its own wording allows, followed by the constraints the filing's **What to do** and **Where** sections add.

- [x] **AC1 — "If the core-link drops during `--wait`, the command exits non-zero (or reconnects and still exits) rather than hanging forever."**
  - [x] *exits non-zero.* `waitForTurnEnd` now rejects with `CoreSessionLinkLostError` when the link goes and does not come back, and `awaitTurn` in `packages/cli/src/session-command.ts` already returns `EXIT_FAILURE` for anything the wait throws. SDK: *"ends a wait that has no deadline at all, rather than hanging on a dead link"* drops a live socket under a `waitForIdle()` with **no `timeoutMs`** — the case where the rejection is the only thing that can ever settle the promise. CLI: *"exits non-zero when the link drops mid-wait, and calls the outcome unknown (#396)"* asserts `EXIT_FAILURE`.
  - [x] *or reconnects and still exits.* A drop starts a grace that runs only while the link is down and is cancelled on the next connection, so a client that comes back carries on waiting and ends on the Core's own report. SDK: *"keeps waiting when the link comes back inside the grace"* (mutation-checked — it fails if `onLinkBack` is disabled) and *"gives a link that may come back the grace to do it in"*.
  - [x] *rather than hanging forever.* Both branches terminate: the grace is bounded (`CORE_LINK_LOST_GRACE_MS`, 30s) and is 0 on a client that will not reconnect, so the CLI's one-shot client fails on the drop itself rather than sitting out a grace nothing could redeem. SDK: *"fails at once on a client with no reconnect coming, and names the reason"*.

### From **What to do** — "Disconnect must reject or retry the waiter."

- [x] **The waiter is retried first, then rejected.** The grace is the retry: on a reconnect the client re-subscribes from its cursor and the Core streams the tail, including the status the wait was missing. What follows a grace that expires is a rejection, never a resolution.
- [x] **"A default timeout is acceptable if disconnect detection is also honest."** No new default timeout was added — #405's `send --wait` default is untouched — because disconnect detection now exists, which is the condition the sentence attaches it to.

### From **Where** — the three sites the filing names

- [x] **`core-session.ts` "867–892 no disconnect handler"** — the whole of this PR. `CoreSession` listens on `onDisconnected`/`onReady` on both the `start` and the `attach` path.
- [x] **`core-client.ts` "1009–1012 `eventsSubscribed=true` before sendNow"** — fixed. The flag is now set only when the frame actually went out; before, `isSubscribedToEvents()` answered true for a subscribe no link carried, which is a Session wired to an event stream nobody asked for and another wait nothing could end. Test: *"records a subscribe only when one actually went out (#396)"*.
- [ ] **`onCoreEvent` "drops null taskId"** — **investigated and deliberately not changed.** `onCoreEvent` drops an event whose `taskId` is not this Session's, `null` included. Every kind in `STATUS_BEARING_EVENT_KINDS` (`task:updated`, `task:statusChanged`, `session:finished`) is keyed by the Task row it patched and carries a `taskId`; the kinds that legitimately carry `null` there are PTY-scoped (`pty:spawn`, `pty:exit`), are not status-bearing, and reach this layer as `exit` **frames**, which `settledSince` already answers every wait from. Routing a status-bearing event by `ptyId` when its `taskId` is null would be inventing a case the Core does not produce, and guessing which Session an untagged status belongs to is the kind of inference ADR 0033 forbids. Left as filed rather than changed on speculation; if a Core is found to append a status-bearing event with no `taskId`, that is a Core-side bug and its own ticket.

### Landmines named in the brief

- [x] **No turn settled from PTY bytes or output quietness** (#191, ADR 0026 D3) — the new path reads no screen; its facts are the link going down and two event ids.
- [x] **`reportsTurnStart` consulted in no wait path** (ADR 0033 D2) — the new handlers sit below `// ─── Internals`, outside the slice the source-sweep test reads, and consult nothing. That test passes.
- [x] **An unknown outcome reads as unknown** — the wait rejects rather than resolving, `--json` carries no `status` key, and the message says the outcome is unknown and that this is *not* a report that the turn finished.
- [x] **#405 respected and extended, not undone** (ADR 0033 D5) — `CoreSessionTurnTimeoutError`, the `send`-only default deadline and `--wait-timeout 0` are unchanged; ADR 0033 D6 extends D4 beside D5; and *"leaves a turn that really ended ending the wait, drop or no drop"* guards the resolution path.

### Boundaries

- [x] `packages/core/src/harness-prompt-delivery.ts`, `packages/core/src/pty-manager.ts`, `packages/cli/src/session-gateway.ts` — untouched (#483).
- [x] codex hooks code — untouched (#290). codex `HARNESS_READINESS` table and `packages/core/src/__tests__/fixtures/` — untouched (#277).
- [x] `packages/panel/**` — untouched (#484).

Refs #396

mrfqcentic (changes requested):
Review of #492 against #396, and against what #405 left in this same file.

**The design is right and the acceptance criterion is met.** A drop rejects rather than resolving, `CoreSessionLinkLostError` is a distinct class from both a resolution and #405's `CoreSessionTurnTimeoutError`, and nothing on any wait path reads PTY bytes, measures output quietness or consults `reportsTurnStart`. #405 is byte-identical: `CoreSessionTurnTimeoutError`, `turnTimeoutMessage`, `SEND_WAIT_DEFAULT_TIMEOUT_S = 1020` (send only), `waitTimeoutMs` including `--wait-timeout 0`, `sessionSend` and `sessionWait` all diff clean against the base — the only CLI change is `awaitTurn` plus its import. The `#289 A` source sweep still passes (slice length 1648, no `reportsTurnStart`), and six of the eight new SDK tests fail against the pre-PR source, so the coverage is real.

Requesting changes on two reproducible hangs, both on the reconnecting-client path, both of which falsify the categorical claim this PR and ADR 0033 D6 make: *"a wait cannot outlive the link it is listening on"*. Neither can produce a false completion — the class discipline holds throughout — so this is not the blocking category, but each one re-opens #396's own bug through a different door, and the fixes are small.

---

### Blocking 1 — `packages/sdk/src/core-session.ts:1352` (`onLinkBack`): a flapping link never accumulates downtime

`onLinkBack()` sets `linkDown = false` and clears `linkLostTimer` outright. The next drop calls `armLinkLostGrace()` and gets a **full fresh grace**. Nothing anywhere adds up how long this side has actually been deaf, so a link that drops and returns repeatedly — each outage shorter than the grace — never fails the wait, however much total time is lost.

Reproduced against this head: `linkLostGraceMs: 100`, five drop/`ready` cycles of ~65 ms (~325 ms cumulative downtime), `waitForIdle()` with **no `timeoutMs`** — still pending, and it stays pending indefinitely. The control case (one outage longer than the grace) fails correctly, so the probe is sound.

A Core in a restart crash-loop, or a `DurableCoreClient` on its 500 ms–5 s backoff against a flaky link, is exactly this shape. Suggest tracking cumulative downtime across the wait, or requiring a connection to survive/deliver before the grace is forgiven, rather than zeroing on every reconnect.

### Blocking 2 — `packages/sdk/src/core-session.ts:822` and `:956`: the grace is disarmed by `ready`, which is not "the link is usable"

Both call sites wire the disarm to `client.onReady(...)`. But `ready` is the Core's first unsolicited frame and arrives **before** `auth` is answered — `core-client.ts:466` fires `readyListeners` straight off the transport, and `maybeEstablish`'s own comment says so ("With a bearer it is `authOk`, well after `ready`"). So a connection the Core is about to refuse counts as the link coming back.

Reproduced against this head with an expired bearer: listener order is `["ready","authError","disconnected"]`.

Combined with Blocking 1 and with `durable-core-client.ts:124` scheduling a reconnect unconditionally (no auth-error exception), an expired or rotated bearer gives a permanent `ready` → `authError` → `disconnect` loop in which the grace is re-armed and disarmed forever and the wait never ends — with no events flowing at all, since `subscribe` only goes out from `onConnectionEstablished`, i.e. after auth. Suggest disarming on the established/authenticated signal (`onAuthOk` / `onConnectionEstablished`) rather than on `ready`.

---

### Should fix 1 — `packages/sdk/src/core-session.ts:1229`: the `timeoutMs` path leaves the grace armed and part-spent

`notify` (`:1209`) and `fail` (`:1215`) both call `releaseLinkLostGrace()`; the deadline callback at `:1229` deletes the waiter and rejects without it. Two consequences: the timer is deliberately not `unref`ed, so it holds the event loop for the rest of the grace after the wait has already settled; and because `armLinkLostGrace` short-circuits on `if (this.linkLostTimer) return`, the **next** wait on that Session inherits the stale, part-spent timer.

Measured: `linkLostGraceMs: 300`, wait A timing out at 60 ms, wait B started fresh afterwards — B was failed after **239 ms**, and its `CoreSessionLinkLostError.graceMs` and message still said `300`, which `:523` documents as "how long the wait went on for after the drop before giving up". One line — `releaseLinkLostGrace()` in the timeout callback — fixes the leak; reporting real downtime rather than the configured constant fixes the field.

### Should fix 2 — `packages/sdk/src/core-session.ts:1225`: a synchronous link-lost rejection orphans the deadline timer

`if (this.linkDown) this.armLinkLostGrace();` runs **before** `timer` is assigned at `:1228`. On a grace of 0 — the `actana` CLI's one-shot client — `armLinkLostGrace` calls `failWaitersLinkLost()` synchronously inside the executor, so `fail` runs while `timer` is still `null`, and the `setTimeout` armed two lines later is never cleared by anything, `dispose()` included.

Verified with fake timers: after the promise has already rejected with `CoreSessionLinkLostError`, `vi.getTimerCount()` is 1 — a 1 020 000 ms timer for a settled wait. Harmless in the CLI, which hard-exits at `actana-cli-entry.ts:212`, but an SDK consumer holds the event loop for up to 17 minutes per occurrence. Moving the `linkDown` check below the timer arming is enough.

### Advisory — the `#289 A` sweep no longer covers the code that ends a wait

The sweep slices `waitForIdle(` → `// ─── Lifecycle`, and the new handlers sit under `// ─── Internals` at `:1337`–`:1416`. They genuinely consult nothing (I checked), and the guard still passes, but it no longer covers a code path that ends waits, which is what the PR body claims it protects. Worth extending the slice or adding a second assertion over the link-lost handlers.

### Advisory — CLI coverage stubs the throw

Both new `session-command.test.ts` cases construct `CoreSessionLinkLostError` by hand in a `fakeStartedSession`. That pins the branch and the `--json` shape, which is worth having, but no CLI test drops a real socket; `in-process-core-session.test.ts` is where an end-to-end one would go.

---

### Verified sound

- `subscribeEvents` recording the flag only on a successful send: `DurableCoreClient` re-sends `subscribe` unconditionally per connection, and the three `if (!isSubscribedToEvents())` call sites now retry instead of trusting a lie. No double-replay path.
- `start`'s `session?.onLinkLost` closures — `session` is `let … = null`, no TDZ; `stopDown`/`stopUp` released on both the spawn-failure and normal paths, and `attach`'s go through `dispose()` in its catch.
- `linkLostGraceFor` / `willReconnect()`: the CLI builds a one-shot `CoreClient`, so grace 0 and immediate failure, as documented and as #396 asks for.
- `--json` on the link-lost path carries no `status` key (`startedFields` has none) and the `events tail --since` vs `session ls` branch matches `afterEventId`.
- ADR 0033 D5 untouched; D6 added as an extension of D4; `docs/README.md` row updated; `SKILL.md` and both `orchestration-skill-payload.ts` copies byte-consistent.
- Boundaries clean: no `harness-prompt-delivery.ts`, `pty-manager.ts`, `session-gateway.ts`, `packages/core/**` (so no codex `HARNESS_READINESS`, no `__tests__/fixtures/`), no codex hooks, no `packages/panel/**`.
- `Refs #396` present, no `Closes`/`Fixes`/`Resolves` in the body or either commit; base is `fix/prompt-delivery-and-send-path`.
- Against #494 (#395): that branch does not touch `core-session.ts` at all, so `waitForTurnEnd` is this PR's alone. They overlap in `session-command.ts`, both `orchestration-skill-payload.ts` copies, `SKILL.md` and `session-command.test.ts`; `git merge-tree` reports **no conflict** in either direction, so merge order is free.
- Locally on this head: `sdk` 305 passed, `cli` 1229 passed, `core` 980 passed, `shared` 555 passed; `tsc --noEmit` clean in all three changed packages; eslint 0 errors (1 pre-existing warning in an untouched file).

qcentic-dev:
Both blocking findings are fixed in `b21b1f5`, both should-fix findings with them, and one of the two advisories. Thank you for the two repros — the second one in particular, because the `ready`-is-not-recovery point is the sort of thing that reads as a detail until you follow it into the `authError` loop.

You were right that both falsified the categorical claim. I have treated that as the bar: the property is now that **no sequence of drop, flap and reconnect can leave `waitForTurnEnd` waiting for ever**, and the two fixes are the two halves of it.

### Blocking 1 — a flapping link never accumulated downtime → fixed

The grace is now a **cumulative budget** rather than a fresh allowance per outage. `onLinkBack` banks the outage it is ending (`linkDownBankedMs`) instead of clearing it, `armLinkLostGrace` arms for `graceMs - linkDownElapsedMs()` rather than for the whole constant, and the accumulator is reset only when the last waiter goes — the budget belongs to a wait, not to the Session. Summed deaf time is therefore bounded, so the wait always ends.

One thing your repro pushed me to add: the timer's callback re-checks the budget before firing and re-arms if it arrived short. `setTimeout` can land a millisecond ahead of the wall clock it was sized against, and a wait failed a millisecond early is a wait failed before its link was out of time. The budget is the authority, not the timer.

Test: *"adds a flapping link's outages up rather than forgiving each one"* — your parameters, five ~65 ms outages against a 100 ms budget, `waitForIdle()` with no `timeoutMs`. Verified to hang on the old accounting (4 s bound, "the wait was still pending"), and it asserts `downMs >= 100` so it is failing on the sum rather than on any one outage.

### Blocking 2 — the disarm hung on `ready` → fixed

New on the client: **`CoreClient.onEstablished`**, fired from `maybeEstablish` once the Core has said who it is *and* the link is writable — which, with a bearer configured, means `authOk` has landed. Both `CoreSession` call sites (`start:822`, `attach:956`) now use it, and `onReady` is used by neither. The doc on the new method says why in the terms you put it: `ready` is unsolicited and precedes `auth`, nothing can be sent on such a connection, and `subscribe` rides `onConnectionEstablished`, so no events flow for the whole of it.

Test: *"does not take an unauthenticated `ready` for the link coming back"* — the rig gained an auth gate and a staged dial, so dial 0 is the real Core with a good bearer and dial 1 is a socket that opens, says `ready`, and never answers the `auth` behind it. Verified to hang against `onReady`. It also asserts the staged socket really did receive one `auth` frame, so the case under test is the one described rather than a dial that never happened.

Two more at the client layer pin the signal itself: with an expired bearer the listener order is exactly your `["ready","authError","disconnected"]` and **no** `established`; with a good one it is `["ready","established"]`.

### Should fix 1 — the `timeoutMs` path leaked the grace → fixed

`releaseLinkLostGrace()` added to the deadline callback. And you were right that the field was the other half of it: the error now carries **`downMs`** — the deaf time actually spent — beside `graceMs`, which is documented as the *budget*. Your measurement is the test: wait A times out at 60 ms inside a 300 ms grace, wait B starts fresh, and B must get the whole budget rather than the ~240 ms left behind. Verified to produce exactly your 240 ms without the fix.

### Should fix 2 — the synchronous rejection orphaned the deadline → fixed

`if (this.linkDown) this.armLinkLostGrace()` moved below the `setTimeout`, so a grace-of-0 failure inside the executor runs with `timer` assigned and clears it. Test uses fake timers as you did and asserts `vi.getTimerCount()` is 0; it reproduces your count of 1 without the fix.

### Advisory — the `#289 A` sweep → done

Second slice added over `onLinkLost` → `ingestExit`, asserting no `reportsTurnStart` and, while it is there, no `terminal` or `screen` either. The guard now covers every path that ends a wait, which is what the PR body claims for it.

### Advisory — CLI coverage stubs the throw → deferred, deliberately

Not done this pass. The SDK suite now drops real sockets on five paths, including two that hang without the fix, and the CLI branch beneath it is a pass-through of `afterEventId` and the error's own message into two `deps.err` lines — an end-to-end case in `in-process-core-session.test.ts` would re-test the SDK through a longer pipe for the branch the two stubbed cases already pin. Happy to add it if you would rather have it; it is coverage of a seam I did not want to widen this change to reach.

### Also in this push

`ADR 0033 D6` now states both clauses rather than the first draft's single "cancelled the moment a connection comes back", since that sentence was the thing your findings falsified. Latest base merged in with a plain merge (`fac22bd`, bringing #489/#290) — no conflicts, and it touches none of this branch's files. #395/#494 still does not touch `core-session.ts`, and nothing here goes near the start return path.

Local on this head: `sdk` 312, `cli` 1229, `core` 997, `shared` 558 all passing; `pnpm -r typecheck` clean; eslint 0 errors, 43 warnings, identical to the base's own count.

mrfqcentic (approved):
Round 2 on #492. **Both blockers are genuinely fixed** — structurally, and with tests that hang on the old code. Approving.

I did not take the fixes on description. For each of the four findings I reverted *only* that fix on the new head, leaving everything else in place, and ran the suite:

| Finding reverted in isolation | What the new test did |
| --- | --- |
| `onLinkBack` banking + `armLinkLostGrace`'s remaining (blocking 1) | *"adds a flapping link's outages up rather than forgiving each one"* → **`the wait was still pending after 4000ms — it is hanging`** |
| `onEstablished` → `onReady` at `:866` and `:1000` (blocking 2) | *"does not take an unauthenticated `ready` for the link coming back"* → **`the wait was still pending after 4000ms — it is hanging`** |
| `releaseLinkLostGrace()` in the deadline callback (should fix 1) | *"hands the grace back when the caller's own deadline fires"* → `expected 240 to be greater than 260` — my r1 measurement to the millisecond |
| the `linkDown` check moved back above the `setTimeout` (should fix 2) | *"orphans no deadline timer…"* → `expected 1 to be +0` |

In each case exactly one test failed and the other 59 passed, so each test is pinned to its own defect. The `settledWithin` helper is the right call: a hang that would otherwise take the suite down reports as a clean failure naming the hang.

### Blocking 1 — cumulative budget · verified

`onLinkBack` (`core-session.ts:1431`) banks `Date.now() - linkDownSince` into `linkDownBankedMs` instead of clearing it, `armLinkLostGrace` (`:1471`) arms for `linkLostGraceMs - linkDownElapsedMs()`, and `remaining <= 0` fails now. Deaf time is summed wall-clock, so it is monotonically non-decreasing for a wait and the budget always depletes. The timer callback re-checking the budget before firing is a good catch of your own — a short `setTimeout` arrival re-arms rather than failing a wait whose link still has time, and it terminates because each re-arm is for a strictly positive shrinking remainder.

The new `if (!this.linkDown) return;` guard at the top of `onLinkBack` matters too: without it a stray `established` would zero `linkDownSince` and lose an outage.

### Blocking 2 — `onEstablished` · verified

`CoreClient.onEstablished` (`core-client.ts:1095`) fires from `maybeEstablish`, which requires `ready` **and** `transport.writable`; and `writable` is `!disposed && opened && (bearer === null || authenticated)` (`core-link-transport.ts:242`). So with a bearer configured it cannot fire before `authOk`. That is the right signal, and it is the same one the client itself uses to decide it may send. Firing it last in `maybeEstablish`, behind `onConnectionEstablished`, means a listener hears about a connection that has already sent its `subscribe`.

The two client-layer tests pin the signal itself and reproduce my r1 measurement exactly: `["ready","authError","disconnected"]` with **no** `established` on an expired bearer, and `["ready","established"]` on a good one.

### I tried to break it myself

Six adversarial sequences against the new head, driving `onLinkLost`/`onLinkBack` directly with no `timeoutMs` anywhere:

- 30 rapid flaps of ~12 ms against a 120 ms budget → ends, `downMs >= 120`
- a permanent `ready`→`authError`→`disconnect` loop that never establishes (my r1 scenario) → ends, `downMs >= 150`
- flapping where the link is up 4× longer than it is down → ends, `downMs >= 100`
- repeated `drop()` with no reconnect → ends
- a flap that recovers for good, then 400 ms with the link up against a 200 ms budget → **stays pending**, which is the other half of the bar: a healthy link must not be failed

I could not construct a drop/flap/reconnect sequence that hangs. Time spent *connected* legitimately does not count against the budget, and time spent deaf is now exactly measured and bounded.

---

### Should fix (not blocking) — `core-session.ts:1507`: deafness banked with **no** waiter is charged to the next wait

`releaseLinkLostGrace` zeroes the accumulator only when a waiter is *removed*. Deaf time that accrues while there is no waiter at all is never cleared, so the next wait starts with a partly-spent budget — which contradicts this method's own doc, *"the budget belongs to a wait and not to the Session"*.

Measured on this head, `linkLostGraceMs: 200`: a 150 ms outage with **no waiter**, link recovers, later it dips again and only then does a wait begin — that brand-new wait was failed after **50 ms**, reporting `downMs=201`, 150 ms of which predates it. The follow-on variant is milder but present: a wait after another wait got 91 ms rather than 200, because deafness continuing between A leaving and B starting is banked to B.

It errs in the safe direction every time — an earlier unknown-outcome rejection, never a hang and never a false completion — and the `actana` CLI cannot reach it (one-shot client, grace 0, one wait per process). It bites an SDK consumer holding a long-lived `DurableCoreClient` across successive waits, where an earlier flap can make a later wait give up almost at once. Zeroing the accumulator when a wait starts as the first waiter — the mirror of what `releaseLinkLostGrace` does at the other end — is the same one-liner.

The related shared-budget case is the same shape: two concurrent waits on one Session share one accumulator and one timer, so a later one draws on the earlier one's spend. Also bounded, also safe-direction. Worth a sentence in the doc if you do not want to change it.

### Advisory — `Date.now()` rather than a monotonic clock

`linkDownSince`/`linkDownElapsedMs` are wall-clock. A backwards NTP step during an outage shrinks the measured deaf time and extends the wait by the correction. Exotic, and the re-arm loop is what makes it visible rather than what causes it. `performance.now()` would close it.

### Advisory declined — CLI end-to-end

Accepted, and I agree with the reasoning. The branch is a pass-through of `afterEventId` and the error's message into two `deps.err` lines, two CLI cases pin it and the `--json` shape, and the substance is now covered by real sockets at the SDK layer. Not worth widening this change to reach `in-process-core-session.test.ts`.

### Advisory — `#289 A` sweep · done

Second slice runs `onLinkLost` → `ingestExit` and covers all six handlers contiguously (`onLinkLost`, `onLinkBack`, `linkDownElapsedMs`, `armLinkLostGrace`, `releaseLinkLostGrace`, `failWaitersLinkLost`). I ran both slices against this head after comment-stripping: 1777 and 2278 chars, neither containing `reportsTurnStart`, `terminal` or `screen`. The guard now covers every path that ends a wait.

---

### Nothing already met has regressed

- A dropped link ends the wait; it is an **unknown** outcome, distinct by class from a resolution and from #405's `CoreSessionTurnTimeoutError` (`not.toBeInstanceOf` still asserted), and the message still asserts `/outcome is unknown/`, `/not a report that the turn finished/`, `/may still be running/`. `--json` carries no `status` key.
- No PTY bytes and no output quietness settle anything (#191, ADR 0026 D3) — now enforced by the second sweep slice as well as by inspection.
- ADR 0033 **D5 is byte-identical** to the base. D6 was rewritten to state both clauses rather than the falsified "cancelled the moment a connection comes back" — the right correction to make.
- #405's artifacts are still byte-identical against the **new** base `2926109`: `CoreSessionTurnTimeoutError`, `turnTimeoutMessage`, `waitTimeoutMs`, `sessionSend`, `sessionWait`, and `SEND_WAIT_DEFAULT_TIMEOUT_S = 1020` on the send verb only. `awaitTurn` is unchanged from r1.
- `Refs #396`, no `Closes`/`Fixes`/`Resolves` in the body or any commit. Base is still `fix/prompt-delivery-and-send-path`.
- Boundaries still clean: no `harness-prompt-delivery.ts`, `pty-manager.ts`, `session-gateway.ts`, nothing under `packages/core/**` (so no codex `HARNESS_READINESS`, no fixtures, no hooks), nothing under `packages/panel/**`. Same 13 files.

### Base and neighbours

The base moved to `2926109` (#489 / #290) and is merged in at `fac22bd` — a plain merge touching none of this branch's files. `mergeable=MERGEABLE`, `mergeState=CLEAN`, `merge-tree` 0 conflicts. #494 (#395) is unchanged at `3561d45` and still touches `packages/sdk/src/core-session.ts` **zero times** — `waitForTurnEnd` is this PR's alone; its SDK work is `core-link-frames.ts`. They overlap only in `session-command.ts`, both `orchestration-skill-payload.ts` copies, `SKILL.md` and `session-command.test.ts`, and `merge-tree` reports 0 conflicts in both directions. Merge order is free.

### Green on this head

CI: 18 pass, 6 skipping, **0 failures** — `Unit Tests` included, so #491's backstop flake did not even fire. Locally: `sdk` 312, `cli` 1229, `core` 997, `shared` 558 all passing; `tsc --noEmit` clean in all three changed packages; eslint 0 errors, 1 pre-existing warning in an untouched file.

Approving. The one should-fix above is real but safe-direction and can land whenever; it does not gate this.

Issue #469, opened 2026-09-02
fix(core): a harness parked on an unmatched approval dialog is settled finished by the idle backstop


qcentic-dev:
Raised in the round-3 review of [#455](https://github.com/actana/control/pull/455) (implementing #391) as a
should-fix, and deferred out of that PR rather than implemented there.

## What happens

The quiet-Session backstop's **idle rule** (#455) settles a `running` Session whose PTY is still producing bytes but
nothing new on screen. A harness blocked on a permission or approval dialog satisfies that exactly: the dialog is
repainted, unchanged, for as long as nobody answers it.

Only two screens are pattern-matched into `needs-input` today — `hasClaudeInterruptPrompt` and
`hasCodexHookReviewPrompt` in `packages/core/src/pty-manager.ts`. Any other approval box is not recognised, so:

- the row never reaches `needs-input`, which means the sweep's "`needs-input` is never swept" exemption does not
  reach it either;
- the idle rule settles it `finished` at ~9–10 minutes (a reviewer drove a four-line approval box and measured
  exactly that);
- `session:finished` fires, so the operator gets a completion toast and an orchestrating Session's
  `actana session wait` collects a report from a Session that is in fact waiting for a human;
- the idle rule's reopen cannot help until somebody answers the dialog, and the marker expires after 30 minutes.

This is the same harness class #391 targets: the harness whose lifecycle hooks are not arriving is the harness whose
`needs-input` hook is also not arriving.

## Why it is not fixed in #455

Fixing it means either recognising more dialog shapes off the PTY (the `hasCodexHookReviewPrompt` /
`harness-prompt-delivery.ts` family of screen reading, with ADR 0026's caution about acting on a screen the Core
cannot read), or giving the idle rule a way to tell "repainting a static dialog" from "repainting a spinner" — a
question about the *content* of the frame, which the classifier deliberately does not ask. Both are wider than the
backstop, and #455 is already three review rounds deep on the rule itself.

#455 documents the limit in `docs/harness-status-detection.md` under the idle rule's costs, and points here.

## What a fix probably needs

- A widened dialog detector, or a generic "this frame is a prompt awaiting input" signal, feeding the existing
  `onSessionOutputSignal` → `needs-input` path — which already exempts the row from the sweep entirely.
- Or an idle-rule exemption for a Session whose screen has not changed *at all* (a static dialog) as opposed to one
  whose only change is a counter (a spinner) — the two are distinguishable, since a spinner's frame keeps moving
[...]

PR #493, opened 2026-09-03
fix(core): codex gets a HARNESS_READINESS row, measured off its own screen

qcentic-dev:
## What this changes

`codex` was the last harness on the pre-229 prompt-delivery path — settle, type,
submit — and it was there **unverified rather than cleared**. It has a
`HARNESS_READINESS` row now, and the measurement that earned it moved which hole
the row is actually for.

> **Revised after review.** The numbers below are the ones the module's own
> signature ring produces, replayed chunk by chunk at the offsets the captures
> recorded. The first revision of this PR fed each capture as a single blob and
> quoted a coarser paint rule; both are fixed in `ffa45ad`, and three numbers
> moved. See [the review reply](https://github.com/actana/control/pull/493#issuecomment-5533284910).

Everything below is off live PTYs on this Core: **24 boots of codex-cli 0.153.0**,
spawned at the Core's own 100×30 with `TERM=xterm-256color`, every chunk's offset
from spawn recorded alongside it. Nothing here is transcribed from another
harness's shape or from memory.

### The boot race codex was suspected of does not exist on this build

`› Ask Codex to do anything` — the composer's own placeholder — is in the **first**
frame codex paints.

| | measured |
| --- | --- |
| first byte | 56–98 ms |
| composer placeholder on screen | **160–198 ms**, on 8 of 8 timed cold boots (5 × `codex --enable hooks`, 3 × plain) |
| 350 ms quiet gap expires | 564–955 ms |
| margin, marker → settle | **+366 to +782 ms**, every run |

And the marker is not merely early — the harness really is listening that early.
Eight further boots wrote a prompt at 0, 60, 120, 170, 200, 400, 700 and 1200 ms;
**all eight echoed into the composer**, including the writes that went in before
codex had painted anything at all. There is no opencode-shaped four-second hole
here and no claude-code-shaped 222 ms one.

If the row had to rest on that race, it would not be justified, and this PR would
be saying "no row needed". It rests on something else.

### The hole is the directory-trust dialog

In a directory codex does not trust:

| | measured |
| --- | --- |
| composer placeholder painted | 198 ms — the identical first frame |
| last frame with a signature the ring had not seen | 284 ms |
| screen cleared | 633 ms |
| **quiet gap expires, dialog up, no composer anywhere** | **634 ms — one millisecond later** |
| `Do you trust the contents of this directory?` | 638 ms |

`codex` has **no `BLOCKING_DIALOGS` entry**, so nothing else in the module was
watching for that screen either. Measured on a further boot: a 22-character prompt
written at 990 ms produced **no echo and no change to the screen at all**. The
dialog swallows it, and the `\r` D8 sends after the submit pause lands on a menu
whose highlighted row is `1. Yes, continue` — the directory gets trusted, a
session starts, and the prompt is simply gone. A Session parked in `ready`,
`--wait` timing out, nothing to send to: **issue #277's own stated signature,
reached by a different route than #229's.**

With the row, the marker is absent at 634 ms, the module holds instead of typing,
and the prompt is still in hand when a human answers the dialog.

**One millisecond is a measurement, not a safety margin**, so the other ordering
is replayed too. Had the gap opened first, delivery would have typed into a
composer the dialog was about to wipe, `confirmEcho` would have seen no echo, and
the `\r` would have been withheld rather than pressed into the menu —
`prompt-swallowed`, not a lost prompt. The other untrusted boot in the sample
cleared at 555 ms against a gap at 908 ms, so 1 ms is the tight end of the range.

**What the caller is told is the marker, not the dialog:** `codex composer never
appeared within 15000 ms`. `onDeadline`'s dialog branch is reachable only for a
harness with a `BLOCKING_DIALOGS` entry, and codex has none.

### The row

```ts
codex: {
  composer: [/ask\s+codex\s+to\s+do\s+anything/i],
  confirmEcho: true,
  maxPromptWrites: 3,
},
```

`confirmEcho` is the only field here that can turn a delivery that works today
into one that fails, so it has its own evidence — on the workload it will meet
rather than on `"say hello"`. `sanitizeInitialInput` flattens a multi-line prompt
to one line before delivery sees it, so an 800-character sub-agent contract
reaches `writePrompt` as one 796-character line. Measured on three more live
boots:

| delivery | what codex renders |
| --- | --- |
| the sanitised single line, one `write` — what `writePrompt` does | echoed **verbatim**, wrapped. No chip. |
| the same text as a real bracketed paste, `ESC[200~ … ESC[201~` | echoed **in full**, line breaks preserved. No chip. |

Neither renders as `[Pasted text …]`, which is the case that would have made the
field unsafe: `PASTE_PLACEHOLDER` transcribes Claude Code's wording, not codex's.
Had either collapsed to a chip, the row would have shipped without `confirmEcho`
and said so.

### Two candidate markers were timed and rejected

The discipline #232 set with claude-code's `⏵⏵ … (shift+tab to cycle)` footer, kept:

- **`? for shortcuts`** — arrives *with* the composer on 8 of 8 boots and then
  **goes away**: gone from the settled screen 927 ms later. It would read "no
  composer" on every delivery that begins after boot rather than during it.
- **`>_ OpenAI Codex (v0.153.0)`** — survives to the settled screen, and is still
  the wrong thing: a painted box is exactly what D3a exists to stop reading as
  readiness. It also lands in the same frame as the placeholder on every boot
  measured, so it buys nothing even where it is right.

### Does codex need a `composerWaitMs` ceiling too? **No — and here is why**

**#483 landed on the base while this PR was open** — PR #487 merged as `a2d6f6a`
about twenty minutes after this branch was cut. The base has been merged in
(`950e620`); the only conflict was ADR 0026, where both tickets append an
amendment after the same paragraph, and both are kept with 483's first because it
landed first. **The module and its tests merged cleanly**: #483 changes the
*timing* table and the outcome at the backstop, this changes the *readiness*
table. All 96 tests in `harness-prompt-delivery.test.ts` — 483's and this PR's
together — pass.

**And codex names no ceiling, deliberately.** `composerWaitMs` defaults to
`maxWaitMs`, so codex takes the 15 s it already had:

- A marker that arrives at **160–198 ms** is two orders of magnitude inside the
  backstop. There is nothing for 90 s to buy — #483's case on opencode is a boot
  that lands on either side of the deadline, and codex has no such boot.
- The one screen where the marker **never** arrives is a dialog, and waiting is
  not what turns a dialog into a composer. A 90 s ceiling there would only delay
  the honest `abandoned` → `needs-input` by 75 s while an operator stares at an
  unanswered trust prompt.

**What #483 does change for codex is the outcome at that ceiling, and it is the
outcome this ticket wanted.** Giving codex a marker means the backstop no longer
authorises a blind type into it, so the untrusted-directory case stops being
"type the prompt into the trust dialog at 15 s and report `delivered`" and becomes
`abandoned` at 15 s with a `session:promptAbandoned` row saying the prompt did not
land. There is a test for exactly that composition:

```
expect(deliveryProfileFor("codex").composerWaitMs).toBe(PROFILE.maxWaitMs);
…
expect(h.events.at(-1)).toEqual({
  phase: "abandoned",
  reason: "codex composer never appeared within 15000 ms",
});
```

## The fixtures

Nine files, committed under `packages/core/src/__tests__/fixtures/`, and the
marker is asserted **against those files** rather than against a literal written
beside them:

| file | offset | what |
| --- | --- | --- |
| `codex-0.153.0-boot.txt` | 62 ms | 72 bytes of capability probes, no text |
| `codex-0.153.0-composer.txt` | 171 ms | first painted frame; placeholder in it |
| `codex-0.153.0-idle.txt` | 1098 ms | settled screen, model and directory resolved |
| `codex-0.153.0-untrusted-boot.txt` | 198–589 ms | same first frame, untrusted directory |
| `codex-0.153.0-boot-settling.txt` | 171–597 ms | the 25 chunks between the composer frame and the settle |
| `codex-0.153.0-directory-trust.txt` | 633–638 ms | the screen clear, then the trust dialog |
| `codex-0.153.0-frames.txt` | — | every chunk's offset and length, so the tests replay rather than concatenate |
| `codex-0.153.0-long-prompt.txt` | — | the 799-character, 14-line sub-agent contract used as a start prompt |
| `codex-0.153.0-long-prompt-echo.txt` | — | its composer after one sanitised write |
| `codex-0.153.0-pasted-prompt-echo.txt` | — | its composer after a bracketed paste |

The last two are separate files on purpose: `composerOnScreen` reads whatever it
is handed and does not itself cut at a screen clear — the caller does — so one
fixture spanning both would report a composer that is no longer displayed, which
is the mistake these captures exist to rule out.

Substituted: the project path, and the OSC-0 window title that repeats it. The
untrusted pair needed neither, having been captured in a throwaway directory whose
name is already neutral.

## The three places #277 asked to be updated together

- `packages/core/src/harness-prompt-delivery.ts` — the `HARNESS_READINESS` note
  (the "**codex has no entry and that is not a claim that it is safe**" paragraph
  is replaced by codex's measured account), the `HarnessReadiness` doc comment,
  and `echoConfirmed`'s doc comment (no shipped harness lacks `confirmEcho` now).
- `docs/adr/0026-prompt-delivery-is-a-core-responsibility.md` — the #232
  amendment's codex paragraph is marked superseded, and a new **Amendment — issue
  277** records what was measured, the two rejected markers, and why there is no
  timing override.
- `packages/core/src/__tests__/harness-prompt-delivery.test.ts` — the test that
  pinned codex to the untouched path now pins only a harness nobody has written
  yet, which is the case the empty default is actually for.

## Tests

`harness-prompt-delivery.test.ts` is at 100 tests. The codex ones are driven off
the fixtures **at their captured offsets** — `codexFrames` slices each fixture
back into the PTY chunks the manifest names, and the tests feed them one at a
time, because concatenating them merges signatures and moves the settle:

- the marker on the frame it lands in **and** on the settled screen 927 ms later;
  not on the 72-byte boot;
- ready before the clear on the untrusted boot, **not ready** on the trust dialog,
  with `dialogsForHarness("codex")` asserted still empty beside it;
- both rejected candidates, including `? for shortcuts` proved present on the
  composer frame and absent from the idle one — the measurement, not the opinion;
- delivery replays: settles and types at **564 ms** on the trusted boot — 33 ms
  before its last chunk arrives — with **no** `waiting-for-composer` event, and
  nothing typed at 563 ms; holds the `\r` for the echo; types **nothing** at the
  **634 ms** settle on the untrusted boot, which is the one-millisecond ordering
  above, asserted rather than argued; **would still not have lost the prompt** had
  that millisecond gone the other way; delivers as soon as a composer comes back
  after the dialog; and **abandons at the 15 s ceiling** when nobody answers it,
  which is the #483 composition above;
- `confirmEcho` against the real start prompt: `promptEchoed` on both captured
  composers, an assertion that the paste placeholder is **not** what makes the
  match, and an end-to-end delivery of the long prompt that submits without a
  single `prompt-swallowed` event.

`pnpm -r typecheck` clean. `pnpm lint` 0 errors (43 warnings, all pre-existing).
`pnpm scan:secrets` clean. Commitlint clean over all three commits against the
base.

`pnpm test` green on `cae06d2`, which merges the base including #489 (issue 290):
all six stages, `packages/panel` included (169/169). Stated for the record because
an earlier run on this branch was **not** green — `operator-auth.test.ts` and
`live-read-path.test.ts` each hit `Test timed out in 5000ms` while the machine was
loaded, passed 43/43 on their own, and have passed in every full run since.
Neither file is touched here; this branch changes nothing outside
`packages/core` and the ADR.

## Boundaries respected

- **#290 (`fix/codex-hooks-stop`)** — no codex hooks code and no launch flags
  touched. `--enable hooks` appears only as a *capture* argument in the fixture
  provenance, because it is what the Core's own `startCommand` spawns; three of
  the eight timed boots were taken without it and the marker is identical.
- **#396** — `packages/sdk/src/core-session.ts` untouched.
- **#484** — nothing under `packages/panel`.
- **#483 / #487** — merged in, not edited: `PromptDeliveryProfile` and
  `HARNESS_PROMPT_DELIVERY_PROFILES` are untouched by this branch.
- **#469** — the trust dialog is *recognised* by the readiness marker's absence
  and deliberately not *answered*. Stripping its escapes yields
  `1. Yes, continue2.No,quit`, which `readDialogOptions` cannot read; teaching
  the dialog table to answer it is #469's, not this one's.

## What this does not fix

The #277 comment thread is right that the row settles the delivery half only. The
approval half — codex parking on `Press enter to confirm or esc to cancel` with
the Session still `ready` (#469) and unreviewed hooks that never report a turn
(#290) — is untouched here and is what still blocks an unattended orchestrator.

Refs #277

https://claude.ai/code/session_0173awYLtqy6rRuwa12BmADy


qcentic-dev:
## Acceptance criteria from #277, one by one

### "What it needs"

- [x] **1. Capture codex's boot and its idle composer off a live PTY, the way the other three were captured. Commit them under `packages/core/src/__tests__/fixtures/` named for the exact build.**
  Five screens, byte-for-byte off live PTYs on this Core, spawned at the Core's own 100×30 with `TERM=xterm-256color`, named for the exact build:
  `codex-0.153.0-boot.txt` (62 ms), `codex-0.153.0-composer.txt` (171 ms),
  `codex-0.153.0-idle.txt` (1098 ms), `codex-0.153.0-untrusted-boot.txt` (198–589 ms),
  `codex-0.153.0-directory-trust.txt` (638 ms).
  Only the project path is substituted, and only in the trusted three.

- [x] **2. Time the gap the way #232 did: when does the quiet gap (350 ms) expire relative to the frame the composer actually arrives in? That measurement belongs in the record either way.**
  8 timed cold boots (5 × `codex --enable hooks`, 3 × plain), each chunk's offset from spawn recorded:
  first byte 56–98 ms · **composer placeholder 160–198 ms** · quiet gap expires 602–1810 ms.
  The marker leads the settle by **+419 to +1636 ms on every run** — the gap expires *after* the composer, never inside a hole.
  Cross-checked by writing rather than by inference: 8 further boots wrote a prompt at 0, 60, 120, 170, 200, 400, 700 and 1200 ms and **all eight echoed**, including writes that went in before codex had painted anything.
  In the record: the module's table comment, ADR 0026's new amendment, and the test file's fixture header.

- [x] **3. If there is a hole, add the `codex` row to `HARNESS_READINESS` — composer marker, `confirmEcho`, `maxPromptWrites` — with the marker asserted against the fixture, not against a literal.**
  There is a hole, and it is not the boot race. In a directory codex does not trust it paints the placeholder at ~200 ms, **clears the screen at ~633 ms**, and shows `Do you trust the contents of this directory?` at 638 ms; the quiet gap expires at ~988 ms with no composer on screen, and codex has **no `BLOCKING_DIALOGS` entry**. Measured on a further boot: a 22-character prompt written at 990 ms produced **no echo and no change to the screen at all** — and the `\r` that follows lands on a menu highlighting `1. Yes, continue`. Silent loss; Session parked at `ready`.
  Row added: `composer: [/ask\s+codex\s+to\s+do\s+anything/i]`, `confirmEcho: true`, `maxPromptWrites: 3`.
  The marker is asserted against `CODEX_COMPOSER`, `CODEX_IDLE`, `CODEX_BOOT`, `CODEX_UNTRUSTED_BOOT` and `CODEX_DIRECTORY_TRUST` read from the fixture files — no literal screen text is written next to it.

- [x] **4. Time any candidate marker against the composer it stands for before gating on it.**
  Two candidates on the composer's own frame were timed and rejected:
  `? for shortcuts` — with the composer on 8 of 8 boots, then **gone from the settled screen 927 ms later**, so it would read "no composer" on any delivery starting after boot. Asserted both ways against the fixtures.
  `>_ OpenAI Codex (v0.153.0)` — survives to the settled screen but is a painted box rather than a composer, and lands in the same frame as the placeholder on every boot measured, so it buys nothing even where it is right.

- [x] **5. If the capture shows codex has no hole, say so in the code and in ADR 0026 with the measurement behind it.**
  Adapted rather than skipped, because the capture split the answer. The **boot race codex was suspected of does not exist on this build**, and both the code comment and the ADR amendment say exactly that, with the numbers, and both state plainly that if the row had to rest on that race it would not be justified. The row is earned on the trust dialog instead, so #277 does not end as a "no row needed".

### "Current state (correct, keep until measured)" — all three updated together

- [x] `packages/core/src/harness-prompt-delivery.ts` — the `HARNESS_READINESS` note ("**codex has no entry and that is not a claim that it is safe**") is replaced by codex's measured account; the `HarnessReadiness` doc comment no longer says codex is the one harness on the default; `echoConfirmed`'s doc comment no longer names codex as the harness without `confirmEcho`.
- [x] `docs/adr/0026-prompt-delivery-is-a-core-responsibility.md` — the #232 amendment's "**`codex` still has no entry, and that is a gap rather than a clearance**" paragraph is marked superseded, and a new **Amendment — issue 277 (2026-09-03)** carries the measurement, the two rejected markers, and why there is no timing override.
- [x] `packages/core/src/__tests__/harness-prompt-delivery.test.ts` — the test that pinned codex to the untouched path now pins only `invented-tomorrow`, and asserts `codex` is among the four keys of the table.

### Explicitly not done, with reasons

- [ ] **A `HARNESS_PROMPT_DELIVERY_PROFILES` / `composerWaitMs` entry for codex.** Not needed and deliberately absent. A marker that arrives at 160–198 ms is two orders of magnitude inside the 15 s backstop, and the one screen where it never arrives is a dialog — which no amount of extra waiting turns into a composer. A 90 s ceiling there would only delay an honest `needs-input` by 75 s.
  **Updated:** when this checklist was first posted, #483 had not landed on this base. It has now — PR #487 merged as `a2d6f6a` — and the base is merged in at `950e620`. The two compose: #483 changes the *timing* table and the outcome at the backstop, this changes the *readiness* table; the only conflict was ADR 0026, where both amendments are kept. Giving codex a marker means #483's rule applies to it, so the untrusted-directory case now ends `abandoned` at 15 s instead of typing the prompt into the trust dialog — which is the outcome this ticket wanted and could not reach on its own. There is a test for that composition.
- [ ] **Answering the codex trust dialog.** Out of lane. Stripping its escapes yields `1. Yes, continue2.No,quit`, which `readDialogOptions` cannot read; teaching the dialog table to answer it is #469's. This PR only makes the module *stop typing into it*.
- [x] **A codex backstop test.** Originally omitted because #487 was about to invert the behaviour. Now that #487 has landed, the test is written against the *new* behaviour instead: codex abandons at the 15 s ceiling with `codex composer never appeared within 15000 ms` rather than typing into the trust dialog.
- [ ] **The approval half of codex.** The row settles delivery only. Codex parking on `Press enter to confirm or esc to cancel` with the Session still `ready` (#469) and unreviewed hooks that never report a turn (#290) are untouched here, exactly as the comment thread on this issue warned.

Refs #277

mrfqcentic (changes requested):
Reviewed against issue #277. Every acceptance criterion in "What it needs" is met, and the fixtures are real — I decoded all five byte-for-byte and they are genuine PTY capture, not composition (details below, because that was the thing most worth checking). The row is well argued and the #483 composition is handled correctly. Requesting changes on two substantive points and three small ones; none of them touch the finding, only what the row promises and what the tests prove about it.

---

## The fixtures are genuine — stated plainly, with the reasons

I checked these forensically rather than taking the PR body's word for it.

- **Column arithmetic is exact everywhere.** `codex-0.153.0-directory-trust.txt` lays its paragraph out with one absolute `ESC[row;colH` per word — `Do`@3, `you`@6, `trust`@10, `the`@16, `contents`@20 … `higher`@91, then `risk`@3, `of`@8 … `exec`@92. All ~30 columns are exactly consistent with 100-column wrapping. The trailing-cell erases in `codex-0.153.0-composer.txt` are equally exact: cols 1–6 clear from 42 (a 41-wide box), line 9 from 27 (`› Ask Codex to do anything` = 26 cells), line 11 from 18 (`  ? for shortcuts` = 17). Nobody hand-writes thirty consistent column numbers.
- **The chunk boundaries are unbalanced, which is the tell.** `boot.txt` ends with an unclosed `ESC[?2026h`; `composer.txt` opens mid-frame and closes with the matching `ESC[?2026l`; `idle.txt` has 19 begin-sync and 20 end-sync markers. Those are PTY read boundaries cutting through synchronised-update frames. A composed fixture would be balanced.
- **Real redraw noise.** `idle.txt` carries ratatui inline-viewport scrolling (`ESC[4;30r`, nine `ESC M`, `ESC[r`), the per-character bold sweep over "Booting" (`ESC[16;3HB`, `ESC[16;4H…oo…`) that is codex's shimmer animation, OSC-0 titles with braille spinner glyphs, and real chrome (`Booting MCP server: codex_apps`, the `/usage` and `/fast` notices).
- **The capture provenance shows through the scrub.** `untrusted-boot.txt` resolves its directory to `/tmp/codex-cap/api` at col 14 — exactly where `│ directory: ` ends.

I also replayed the module's own `stripAnsi` over all five and confirmed every assertion the tests make: marker false on boot (72 bytes), true on composer/idle/untrusted-boot, false on the trust dialog; `? for shortcuts` present on the composer frame and absent from idle; and the dialog strips to `› 1. Yes, continue2.No,quit` exactly as claimed.

---

## Requested changes

### 1. `packages/core/src/__tests__/harness-prompt-delivery.test.ts:1591-1600` — the untrusted replay does not replay the capture, and its justification is not how the quiet gap works

The comment says the frames "span 198–589 ms with no gap between them wider than 110 ms, so the quiet gap cannot open inside them; they go in as one chunk at 589 ms … and changes nothing this measures." That reasoning treats every frame as restarting the gap, but `lastPaintAt` moves only on a *novel* `redrawSignature` against the 6-slot ring (`harness-prompt-delivery.ts:981-989`).

I split `codex-0.153.0-untrusted-boot.txt` on its synchronised-update boundaries (12 frames) and ran the module's own signature logic over them:

```
 0  755b  PAINT  "╭───…╮│ >_ OpenAI Codex (v0.0.0) │…"
 1   78b  PAINT  "╭"
 2-5 78b  tick
 6  103b  PAINT  "/tmp/codex-cap/api╭"
 7-11 78b tick
```

Only frames 0, 1 and 6 are paints. In the live stream `lastPaintAt` is therefore the offset of **frame 6**, somewhere inside 198–589 ms — not 589 ms. If frame 6 landed before ~283 ms, the 350 ms gap expires before the 633 ms clear, `onQuiet` sees the pre-clear frame with `Ask Codex to do anything` still on it, emits `settled` and writes the prompt at ~600 ms — into a composer the dialog wipes ~35 ms later. `confirmEcho` then keeps the `\r` back and the delivery still ends `abandoned`, so this is not a silent loss; but it is a different path with a different `writes` value from the one the test asserts (`expect(h.writes).toEqual([])`), and collapsing the frames is what makes it untestable.

The capture recorded per-frame offsets. Please feed the frames at those offsets rather than as one chunk, and let the test say which side of the clear the gap actually falls on. That ordering is the measurement #277 asked for; right now the test asserts the comfortable case and the comment explains why the uncomfortable one cannot happen, on a rule the module does not use. The same collapse in `bootCodex` (line 1557) puts the settle at 522 ms while the PR's own measurement says the gap expires at 602–1810 ms.

### 2. `packages/core/src/harness-prompt-delivery.ts:652-656` — `confirmEcho: true` is evidenced only for short typed prompts, and it is the one field here that can regress a path that works today

The row's other two fields cannot lose a delivery. `confirmEcho` can. The evidence offered for it is `"say hello"` (9 chars) and a 22-character probe. Real start prompts on this Core are not that shape — #277's own comment records two Studio codex Sessions started with a *sub-agent-contract* prompt, and says the prompt "landed in the composer intact and the turn started in both". That is the workload this row now sits in front of, and it was not measured.

The failure chain if codex renders such a write as collapsed paste content:

1. `writePrompt` sends the whole prompt in one write (line 1246). `codex-0.153.0-boot.txt` shows codex enabling bracketed paste (`ESC[?2004h`) in its very first bytes, so it *does* have burst/paste-specific input handling — this PR's own fixture is the evidence that the untested path exists.
2. `promptEchoed` looks for a 12-char squeezed probe, falling back to `PASTE_PLACEHOLDER = /\[\s*pasted\s+text/i` (line 682) — a transcription of Claude Code's wording, not codex's.
3. A miss calls `retypePrompt`, which sets `screen = ""` and returns to `settling` (line 1085).
4. The placeholder is gone from the composer (the prompt is in it), so `composerOnScreen` is now false, `onQuiet` holds (line 1133), and at 15 s `awaitComposer` → `abandonForComposer`.

Result: `abandoned` / `needs-input` with the prompt sitting typed-but-unsent in codex — on a delivery that today types once and submits successfully. Not silent, but a regression on the exact case the issue's field data says works.

This is the same discipline #232 set for markers and this PR keeps for them ("a marker that is *sometimes* late is unusable"). Please apply it to the echo: one boot with a realistic multi-line start prompt, recording what codex actually renders. If it echoes, say so and the field is earned; if it collapses to a chip, either widen `PASTE_PLACEHOLDER` for codex's wording or ship the row without `confirmEcho` and say why.

### 3. `packages/core/src/harness-prompt-delivery.ts:509-511` — stale count directly above the table this PR extends

> Three entries now, and each one is a transcription of a *named* build's composer row. Issue 229 put opencode here; issue 232 adds `cursor-cli` and `claude-code` …

There are four entries as of line 652. Every other paragraph in this block was rewritten for #277 and the `HarnessReadiness` doc above it already says "the fourth row"; this lead paragraph was missed. #277 named this doc block as one of the three places to update together.

### 4. `packages/core/src/harness-prompt-delivery.ts:652` / `docs/adr/0026-…md:337-343` — the operator does not get the message the PR promises

The module comment and the ADR both say the win is that "the operator is left with a dialog to answer rather than a prompt that evaporated". The reason string they actually receive is `codex composer never appeared within 15000 ms`. `onDeadline`'s dialog branch (line 1289) is unreachable for codex because `dialogsForHarness("codex")` is empty — the test asserts exactly that — and that branch exists so "the reason a client is given is the dialog and not the marker".

Worth a sentence saying so, and worth correcting the #469 hand-off while you are there. The ADR at line 342 attributes the non-answering to `readDialogOptions` being unable to read `1. Yes, continue2.No,quit`. That is true but is the *second* obstacle. The first is that `stripAnsi` spaces `ESC[nG`/`ESC[nC` but deletes `ESC[row;colH` outright (line 165), so codex's CUP-laid-out dialog collapses to `…apiDoyoutrustthecontentsofthisdirectory?Working…` — I verified `/\btrust\b/i` and the folder/directory-noun pattern are both **false** against it. Adding codex to the `folder-trust` row today would not match at all. That is a useful finding this capture produced; it should not get lost in a hand-off that names only the menu-reading half.

### 5. Two nits

- `packages/core/src/harness-prompt-delivery.ts:1060-1062` — the rewrite left a mangled wrap: "Past the backstop and / \* past the retype budget".
- `packages/core/src/__tests__/harness-prompt-delivery.test.ts:1549-1550` — no blank line between the cursor-cli describe and the codex one.
- `packages/core/src/__tests__/fixtures/codex-0.153.0-idle.txt` — the directory row is scrubbed to `~/projects/api` but the OSC-0 title in the same file still carries the real project name. Harmless, and it is part of why I believe the capture; flagging only because the substitution is stated as complete.

---

## Acceptance criteria

| # | Criterion | Verdict |
| --- | --- | --- |
| 1 | Boot + idle composer captured off a live PTY, committed under `packages/core/src/__tests__/fixtures/`, named for the build | **Met** — five screens, verified genuine |
| 2 | Time the gap: when does `quietGapMs` expire relative to the composer frame | **Met** in the record (160–198 ms vs 602–1810 ms, 21 boots); **partially met** in the tests, see finding 1 |
| 3 | Add the `codex` row with marker / `confirmEcho` / `maxPromptWrites`, marker asserted against the fixture | **Met** — asserted against the files, not a literal; `confirmEcho`'s own evidence is thin, see finding 2 |
| 4 | Time any candidate marker before gating on it | **Met** — `? for shortcuts` and the wordmark both timed and rejected, with the disappearance proved on the fixtures |
| 5 | If no hole, say so in code and ADR with the measurement | **Met** — the PR says plainly the suspected boot race does not exist and the row would not be justified on it |
| — | Three places updated together (module note, ADR, the test pinning codex to the untouched path) | **Met**, minus the stale lead paragraph in finding 3 |

Also checked, all clean: base is `fix/prompt-delivery-and-send-path` (not `main`, no beta); body carries `Refs #277` and no closing keyword anywhere in the body or the three commits; `HARNESS_PROMPT_DELIVERY_PROFILES` is untouched, so no collision with #483, and the "does codex need a ceiling" question is asked and answered with a reason; the row's shape is identical to the other three (no new fields on `HarnessReadiness`); nothing under `packages/panel`, `packages/sdk/src/core-session.ts` untouched (#395/#396), and no codex hooks code or launch flags touched (#290) — `harness-hooks.ts` and `pty-spawn-policy.ts` are not in the diff, and `--enable hooks` appears only as capture provenance. CI is green across the board.

Findings 1 and 2 are what I would want addressed before this lands; 3–5 are cheap. The measurement work behind this is genuinely good and the conclusion — that the row is earned on the trust dialog rather than on a boot race — is the right one.

qcentic-dev:
Thank you — the forensic pass on the fixtures is the check I most wanted someone to make, and both blocking findings were right. Neither needed a different design; both needed the measurement discipline applied one level deeper. Fixed in `ffa45ad`, with the latest base merged in `cae06d2` (#489 / issue 290 is in it).

---

## Finding 1 — the replay now replays, and the answer moved

You were right that the justification was not how the quiet gap works, and right that collapsing the chunks made the ordering untestable. Both are gone.

`codex-0.153.0-frames.txt` is new and carries **every PTY chunk's offset and length** for all five capture slices. `codexFrames(fixture, source)` slices a fixture back into those chunks — it asserts the manifest consumes the file exactly, so a manifest that drifted would fail rather than silently truncate — and the tests feed them at their own offsets. `codex-0.153.0-boot-settling.txt` is also new: the 25 chunks between the composer frame and the settle, which the old test simply did not have.

Letting the module decide changed three numbers, and every one of them moved toward you:

**The trusted boot settles at 564 ms, not 522 ms.** The last chunk with a signature the ring had not seen is at 214 ms; everything from 252 ms to 597 ms repeats one. So the prompt goes out at 564 ms, 33 ms *before* the last chunk of the boot arrives — which the old collapsed replay could not have shown.

**The eight-boot range is 564–955 ms, not 602–1810 ms.** My original figure came from exactly the coarser rule you identified — `sig !== lastSig` rather than the 6-slot ring. On one boot the two rules differ by 38 ms and on another by 855 ms. Corrected in the module note, the ADR and the test header, with a sentence saying what the old number was and why it was wrong. The marker still leads the settle on 8 of 8, now by 366–782 ms.

**The untrusted case is one millisecond, not 354.** Replayed faithfully: paints at 198, 198, 208 and 284 ms, so the gap opens at **634 ms** — and the screen clear is at **633 ms**. The clear wins by 1 ms, so `composerOnScreen` is false and the module holds. That is the ordering you asked to see timed, and it is now asserted rather than argued:

```ts
it("finds the quiet gap on the dialog's side of the screen clear, by 1 ms", …)
  expect(h.events).toContainEqual({ phase: "waiting-for-composer", waitedMs: 634 });
```

A 1 ms margin is not a safety argument, so the other ordering is replayed too, in `it("would still not have lost the prompt had the gap fallen first")`: delivery settles on the composer, types, the dialog wipes the screen, no echo comes back, and `confirmEcho` withholds the `\r` — `prompt-swallowed`, not a lost prompt. Your reading of that path was exactly right. For range: the other untrusted boot in the sample cleared at 555 ms against a gap at 908 ms, so 1 ms is the tight end and not the usual one. All of this is in the fixture header, the module note and the ADR.

## Finding 2 — `confirmEcho` measured on the workload, not on `"say hello"`

Agreed on the principle and on the specific chain — `retypePrompt` clearing the screen and re-imposing the composer gate is what turns a missed echo into `abandoned`, and `PASTE_PLACEHOLDER` transcribing Claude Code's wording is exactly the wrong thing to fall back on for codex.

One correction to the failure chain, which narrows it rather than dismissing it: **`sanitizeInitialInput` (`pty-manager.ts`) collapses every run of C0 whitespace to one space before delivery sees the text**, so a multi-line contract reaches `writePrompt` as one long line. That is issue 193's fix, and it means the shape to measure is a long *single-line* burst — but codex advertising `ESC[?2004h` still makes the paste path worth measuring, so I measured both.

Three more live boots, one 799-character / 14-line sub-agent contract of the shape #277's field comment describes:

| delivery | what codex renders |
| --- | --- |
| sanitised to one 796-char line, one `write` — what `writePrompt` does | echoed **verbatim**, wrapped across composer lines. No chip. |
| the same text as a real bracketed paste, `ESC[200~ … ESC[201~` | echoed **in full**, line breaks preserved as composer lines. No chip. |

Committed as `codex-0.153.0-{long-prompt,long-prompt-echo,pasted-prompt-echo}.txt`, and `promptEchoed` is run against them in three tests — including one asserting the placeholder is **not** what makes the match, so the row does not quietly depend on wording codex does not print. There is also an end-to-end test that delivers the long prompt and asserts `delivered` with no `prompt-swallowed` event.

The field is earned on evidence now. Had either collapsed to a chip I would have shipped the row without it and said so — that was the plan.

## Findings 3–5 — all applied

- **3.** "Three entries now" → "Four entries now", with `codex` named alongside 229's and 232's rows.
- **4.** Both the module note and the ADR now say plainly that the reason the caller receives is `codex composer never appeared within 15000 ms` — the marker, not the dialog — because `onDeadline`'s dialog branch is reachable only for a harness with a `BLOCKING_DIALOGS` entry and codex has none. And the #469 hand-off now leads with the obstacle you found: `stripAnsi` deletes `ESC[row;colH` outright rather than spacing it the way it spaces `ESC[nG`/`ESC[nC`, so the dialog collapses to `…apiDoyoutrustthecontentsofthisdirectory?Working…` and not even `trust` survives on it. Adding codex to the existing `folder-trust` row would match nothing; the *reading* has to change before the answering can. Recorded in both places so #469 inherits it.
- **5.** The mangled `echoConfirmed` wrap is fixed; the blank line between the describes is back; and the OSC-0 title is now scrubbed to match the directory row — thank you for flagging it even as evidence, since "the substitution is complete" should be true when it is stated.

## Not changed, and why

**Nothing about the design, and nothing about the row's shape.** No new fields on `HarnessReadiness`, `HARNESS_PROMPT_DELIVERY_PROFILES` still untouched, and `composerWaitMs` still the default — the ceiling question is unchanged by any of this.

**#469 is still deferred.** This PR makes the module stop typing into the dialog; teaching it to *answer* one it cannot even read is a separate decision, now better specified.

**The approval half of codex (#469 / #290) is still out of lane**, as the issue thread warned.

Local: `pnpm -r typecheck` clean, `pnpm lint` 0 errors (43 pre-existing warnings), `pnpm scan:secrets` clean, commitlint clean over the branch. `harness-prompt-delivery.test.ts` is at 100 tests.

Refs #277

qcentic-dev:
## Review findings — disposition, and CI on the new head

Head is now `cae06d2` (`ffa45ad` = the fixes, `cae06d2` = a plain merge of the base, which now carries #489 / issue 290).

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | The untrusted replay collapses the capture into one chunk, on a rule the module does not use | **Fixed.** `codex-0.153.0-frames.txt` carries every chunk's offset; `codexFrames` slices the fixtures back and the tests feed them one at a time. Three numbers moved: the trusted settle 522 → **564 ms**, the eight-boot range 602–1810 → **564–955 ms**, and the untrusted gap-vs-clear margin 354 → **1 ms**. The ordering is now asserted (`waitedMs: 634` against a clear at 633 ms) and the opposite ordering is replayed beside it. |
| 2 | `confirmEcho` evidenced only on short prompts | **Fixed.** Three more live boots with an 800-character / 14-line sub-agent contract. Sanitised single-line write and real bracketed paste both echo in full, no chip. Committed as fixtures; `promptEchoed` runs on them, plus an assertion that the paste placeholder is *not* what makes the match, plus an end-to-end delivery with no `prompt-swallowed`. One correction to the chain: `sanitizeInitialInput` flattens newlines before delivery sees them, so the shape that reaches `writePrompt` is one long line — both shapes measured anyway. |
| 3 | Stale "Three entries now" above a four-entry table | **Fixed.** |
| 4 | The operator does not get the message the PR promises; the #469 hand-off names the second obstacle | **Fixed.** Module and ADR now state the reason is `codex composer never appeared within 15000 ms` and why the dialog branch is unreachable here. The #469 hand-off now leads with `stripAnsi` deleting `ESC[row;colH`, which is why even `trust` is not on the stripped screen. |
| 5 | Mangled comment wrap; missing blank line between describes; OSC-0 title unscrubbed in the idle fixture | **All three fixed**, the title scrub applied across the fixture set. |

Nothing was deferred from the review. Still deliberately out of scope, unchanged from before: answering the trust dialog (#469), the approval half of codex (#469 / #290), and any change to the row's shape or to `HARNESS_PROMPT_DELIVERY_PROFILES`.

**CI on `cae06d2`: all 18 reported checks pass, nothing red** ([run](https://github.com/actana/control/actions/runs/33816242935)) — Conventions, Train rules, Typecheck, Lint, Unit Tests (2m51s), Secret Scan, Dependency Audit, Promotion gate, both image resolve/build pairs, E2E Panel service seam, both installer E2Es, both tarball smokes. Six train/image jobs skip by design. No re-run was needed: `core-session-backstop.test.ts` passed inside the green Unit Tests job, so #491 did not appear.

`harness-prompt-delivery.test.ts` is at 100 tests. Local `pnpm test` green on all six stages, `pnpm -r typecheck` clean, `pnpm lint` 0 errors, `pnpm scan:secrets` clean, commitlint clean over the branch. Still a draft, still based on `fix/prompt-delivery-and-send-path`, and nothing was force-pushed.

Refs #277

mrfqcentic (approved):
Re-reviewed at `cae06d2`. Both blocking findings are fixed, and fixed in the way I would have wanted rather than the way that would have closed the ticket fastest — the replay now lets the module decide the number and the answer moved three times, and `confirmEcho` is measured on both shapes the real prompt can arrive in. Findings 3–5 are all applied. **Approving.**

I re-derived the load-bearing numbers independently rather than reading them off the PR body.

---

## Finding 1 — verified fixed, and the fix is real

`codex-0.153.0-frames.txt` gives every PTY chunk an offset and a length; `codexFrames` slices the fixtures back and asserts `at === source.length`, so a drifting manifest fails loudly instead of silently truncating. I checked that guard actually holds:

```
boot             manifest=  72  file=  72  OK
composer         manifest= 738  file= 738  OK
boot-settling    manifest=1425  file=1425  OK
untrusted-boot   manifest=1638  file=1638  OK
directory-trust  manifest= 609  file= 609  OK
```

Then I re-implemented `onOutput`'s paint tracking — `redrawSignature`, the 6-slot ring, `lastScreenClearIndex` — and replayed both captures at the manifest's own offsets. Every number in the PR reproduces:

**Trusted boot.** Last novel signature at **214 ms**; everything from 252 ms to 597 ms repeats one already in the ring. Gap opens at **564 ms**, 33 ms before the final chunk arrives at 597 ms. Not 522 ms.

**Untrusted boot.** Paints at 198, 198, 208, **284 ms** — then nine consecutive ticks.

```
t= 284 len=  53 PAINT  lastPaintAt= 284  gapAt= 634   composerOnScreen=true
t= 345 len=  78 tick   lastPaintAt= 284  gapAt= 634   composerOnScreen=true
t= 479 len= 156 tick   lastPaintAt= 284  gapAt= 634   composerOnScreen=true
t= 587 len=  28 tick   lastPaintAt= 284  gapAt= 634   composerOnScreen=true
t= 589 len=  50 tick   lastPaintAt= 284  gapAt= 634   composerOnScreen=true
t= 633 len=  17 tick   lastPaintAt= 284  gapAt= 634   <CLEAR>  composerOnScreen=false
t= 638 len= 592 PAINT  lastPaintAt= 638  gapAt= 988   composerOnScreen=false
```

The clear lands at 633 ms and the gap opens at 634 ms. **One millisecond, and the module is what produces it.** The 17-character chunk at 633 ms is exactly `ESC[?2026h ESC[1;1H ESC[J` — a real read boundary falling on the clear, not a convenient split.

The test genuinely exercises the ordering rather than asserting around it. `replay` only ever advances the clock to the next chunk's offset, so the idle timer scheduled for 634 ms is still pending when the 633 ms clear arrives and fires during the advance to 638 ms. Had the clear come at 635 ms the test would fail. `expect(h.events).toContainEqual({ phase: "waiting-for-composer", waitedMs: 634 })` is the module's answer, not the comment's.

Three things I want to note as good practice rather than as findings:

- **The correction is recorded, not quietly swapped.** The ADR says outright that an earlier revision claimed 602–1810 ms from a coarser rule, and that the two disagree by 38 ms on one boot and 855 ms on another. That is the right way to move a number.
- **The counterfactual is replayed too.** `it("would still not have lost the prompt had the gap fallen first")` delivers the clear one millisecond late and shows `confirmEcho` withholding the `\r` with a `prompt-swallowed` event. A 1 ms margin is not a safety argument and the PR does not pretend it is; it also gives the range (the other untrusted boot: 555 ms clear against a 908 ms gap).
- **`it("settles 350 ms after the last novel frame, not after the last frame")`** brackets 563 ms (nothing written) against 564 ms (written), which is the tightest form this could take.

The internal counts are exact too: 28 chunks for the trusted boot (2 + 1 + 25), 11 for the untrusted one, 2 for the dialog — all matching the prose.

## Finding 2 — verified fixed, on both shapes

The correction to my failure chain is right, and I checked it: `sanitizeInitialInput` (`pty-manager.ts:314-324`) collapses `WHITESPACE_CONTROL_RUN` to one space and drops C0/DEL, and it runs at line 723 before `HarnessPromptDelivery` is constructed — so a multi-line contract does reach `writePrompt` as one line. Measuring the bracketed-paste shape anyway was the right call, since the boot fixture is what raised it.

Both echo captures are genuine, and they are two *different* renderings of the same text — neither could be derived from the other. I checked each row's trailing-cell erase column against the rendered width of that row:

| fixture | rows checked | consistent |
| --- | --- | --- |
| `long-prompt-echo` (sanitised, reflowed) | 11 | 11 |
| `pasted-prompt-echo` (paste, breaks preserved) | 16 | 16 |

(The two "misses" in each are my parser on the clear row and the true-colour status row, not the fixture.)

The reflowed one wraps the 796-character line at 89 / 96 / 98 / 99 / 96 / 95 / 90 / 93 / 50 cells with a 2-column hanging indent — a greedy word wrap at width 98. The pasted one reproduces the prompt's own 13-line structure **including its blank lines** (rows 15, 17, 23, 27 erase at column 2), with only the first line wrapping. `Rules you must follow:` at row 18 is 22 characters, occupies 24 cells with the indent, and erases from column 25. Exact, and a different geometry from the other file.

Running the module's own `promptEchoed` over them:

```
promptEchoed(long-prompt-echo,   sanitised 796-char line) = true
promptEchoed(pasted-prompt-echo, 799-char 13-line prompt) = true
paste placeholder present in either?                        false
```

So the field is earned on the echo probe itself, not on a placeholder codex does not print — and `it("does not owe its answer to the paste placeholder it does not print")` pins exactly that, which is the assertion that keeps this honest if codex's rendering ever changes. The end-to-end test delivering the long prompt with no `prompt-swallowed` closes it.

Stating the counterfactual as the author did — "had either collapsed to a chip, the honest answer would have been to ship the row without it" — is what I was asking for, and it is in the ADR as well as the reply.

## Findings 3–5 — all applied, verified

- **3.** `harness-prompt-delivery.ts:509` now reads "Four entries now … issue 277 adds `codex`, the last harness that had been left on the default."
- **4.** `harness-prompt-delivery.ts:617-635` now has a **What the caller is told, precisely** paragraph saying the reason names the marker and why `onDeadline`'s dialog branch is unreachable for codex, and the #469 hand-off now leads with `stripAnsi` deleting `ESC[row;colH` and states that adding codex to the existing `folder-trust` row "would not match at all; the reading has to change first." Same in the ADR at lines 359–371. That is the finding carried forward properly.
- **5.** Mangled wrap fixed (`ts:1120-1121` reads cleanly); blank line restored between the describes (`test.ts:1660-1662`); OSC-0 titles in the idle fixture scrubbed to `api` to match the directory row — I confirmed the scrub is surgical: 3312 → 3288 chars, six titles, and the fixture's unbalanced sync boundaries (19 begin / 20 end) survive intact, so the authenticity evidence is untouched.

---

## Nothing regressed

I re-ran the round-1 checks on the new head.

- **Fixtures still genuine.** Re-verified the whole set after the scrub: `boot` 1 begin-sync / 0 end-sync, `composer` 0 / 1, `idle` 19 / 20 — the mid-frame PTY boundaries are all still there. Marker false on boot and on the trust dialog, true on composer / idle / untrusted-boot; `? for shortcuts` present on the composer frame and gone from idle; the dialog still strips to `› 1. Yes, continue2.No,quit`. The three new fixtures pass the same forensic tests as the original five.
- **Real marker, not quietness** (#191, D3/D3a). Row unchanged: `composer: [/ask\s+codex\s+to\s+do\s+anything/i]`, gated *in addition to* the quiet gap at `onQuiet`.
- **Row shape.** Unchanged, and worth stating plainly: **the only non-comment change to `harness-prompt-delivery.ts` in the entire PR is the five-line codex row.** No new fields on `HarnessReadiness`, no behavioural edit anywhere else in the module.
- **#483.** `HARNESS_PROMPT_DELIVERY_PROFILES` still `{ opencode: { composerWaitMs: 90_000 } }`, untouched. The ceiling question is still asked and answered with a reason, and `deliveryProfileFor("codex").composerWaitMs === PROFILE.maxWaitMs` is still pinned by a test.
- **`Refs #277`, no closing keyword** anywhere in the body or across the six commits.
- **Base** is still `fix/prompt-delivery-and-send-path`; still a draft.

## The base moved, and this stayed in its lane

`fix/prompt-delivery-and-send-path` is now at `2926109` (#489, issue 290 — the other codex ticket). The merge-base equals the base tip, so `cae06d2` is fully up to date and `mergeable: MERGEABLE`. Diffed against the *current* base, the PR is 13 files:

```
docs/adr/0026-…md · 9 codex fixtures · harness-prompt-delivery.test.ts · harness-prompt-delivery.ts
```

`harness-hooks.ts`, `pty-spawn-policy.ts` and `harness-cli-config.ts` — everything #489 touched — are **not** in it, and neither is `packages/sdk/src/core-session.ts` (#396 / #395) or anything under `packages/panel`. The two codex tickets did not collide.

## CI

**18 pass, 6 skipping, nothing red** on `cae06d2` — Conventions, Train rules, Typecheck, Lint, Unit Tests, Secret Scan, Dependency Audit, Promotion gate, both image resolve/build pairs, E2E Panel service seam, both installer E2Es, both tarball smokes. The six skips are the train/publish jobs that skip by design. `core-session-backstop.test.ts` passed inside the green Unit Tests job, so #491 did not appear and no waiver is needed. `harness-prompt-delivery.test.ts` is at 100 tests.

## One nit, not blocking

`harness-prompt-delivery.test.ts:439` and the PR body's fixture table both say the long prompt is "799 characters over 14 lines". It is 799 characters over **13** lines (12 newlines, no trailing one). The assertion beside it (`> 10`) is unaffected. Worth a one-character fix next time this file is open; not worth another round.

---

## Checklist

| Item | Verdict |
| --- | --- |
| Finding 1 — replay faithful to the capture, chunked and timed as captured | **Fixed** — manifest verified exact; 564 ms and 634-vs-633 ms reproduced independently |
| Finding 1 — a test genuinely exercises the gap/clear ordering | **Fixed** — the module decides it; the counterfactual ordering is replayed beside it |
| Finding 2 — `confirmEcho` exercised on a long multi-line prompt | **Fixed** — 799-char / 13-line contract, sanitised single-line write **and** real bracketed paste, both echo in full |
| Finding 2 — the answer does not depend on `PASTE_PLACEHOLDER` | **Fixed** — asserted absent on both captures; the probe is what matches |
| Finding 3 — stale entry count | **Fixed** |
| Finding 4 — operator-facing reason, and the #469 hand-off | **Fixed** in module and ADR |
| Finding 5 — wrap, blank line, OSC-0 scrub | **All fixed**; scrub surgical, fixture evidence intact |
| Fixtures still genuine terminal capture | **Confirmed**, originals and the four new ones |
| Marker real, not quietness (#191, ADR 0026 D3/D3a) | **Confirmed** |
| Row shape matches the existing three | **Confirmed** — only non-comment change in the module is the row |
| Composes with #483; override table untouched | **Confirmed** |
| Ceiling question asked and answered | **Confirmed** |
| `Refs #277`, no `Closes` | **Confirmed** |
| Base `fix/prompt-delivery-and-send-path` | **Confirmed** |
| Merges cleanly on the moved base (#290 / #489 in it) | **Confirmed** — merge-base = base tip, `MERGEABLE` |
| Out of the codex hooks code and `packages/sdk/src/core-session.ts` | **Confirmed** |
| CI green on `cae06d2` | **Confirmed** — 18 pass, 6 by-design skips, no #491 flake |

**Approved.** The row is earned on a screen somebody read, the numbers behind it are the module's own rather than a description of them, and the one field that could have regressed a working path is now measured on the workload it will actually meet — including the outcome that would have removed it. Good work on taking both findings a level deeper instead of arguing them.

PR #494, opened 2026-09-03
feat: session start stops claiming a readiness it has not established

qcentic-dev:
## What this changes

`session start` returned as soon as the Core had the Session running (#129 D6), which is
before the harness can take a keystroke: the composer is not painted, a trust dialog may
not even be drawn, and the Core has not begun typing. The operator's next line is the one
that pays:

```bash
SID=$(actana session start web "fix it")
actana session send $SID continue --enter
```

The keystrokes land on a dialog or in a terminal that is not reading yet, and the harness
discards them — **together with the starting prompt sitting in the same buffer**. One
command loses two messages and reports success for both.

**A Session that is *running* is not yet a Session that can take a `send`, and `start` now
has a way to say which one it is returning.**

### The Core: `session:promptDelivered`

#483 put the *loss* on the wire. It could read the absence of that row as delivery,
because it had already waited for a turn to end. A `start` has not: there, "no abandon
row" and "the composer is still not up" are the same silence, so a command reading the
first as delivery would be claiming a readiness nobody established — the defect, not the
fix.

So the Core appends the positive twin — Task, PTY, character count and `waitedMs`, never
the prompt — from `HarnessPromptDelivery`'s `delivered` phase.

**The ordering discipline #483 recorded, pointed the other way.** That phase runs inside
`submit`, in the same synchronous tick as the carriage return, so the row is in the log
before the event loop can carry a byte of the harness's reply and therefore *strictly
before* any status the turn it starts produces. A delivery row behind the status would be
as unreadable as the reason row behind it was.

The forwarding takes the event whole instead of destructuring it in the parameter list:
the rest of a discriminated union is not narrowed by its own discriminant, so
`event.promptChars` is typed and the cast it replaces was a rename waiting to become a
silent zero. `onSessionPromptDelivered` is optional like its neighbours.

### The client: `--await-prompt`

`session start … --await-prompt` and `session resume … --await-prompt` block until the
Core says which it was — **exit 0 on delivered, non-zero with the Core's own words on
abandoned**. After it a `send` is safe: a delivered prompt is a composer that was observed
on screen, written into, and (where the harness confirms echo) seen to hold the text. That
is the "documented wait" the issue's acceptance criterion offers.

Without the flag, `start` **says so** rather than implying otherwise — a line on stderr
naming the gap and the flag, and `promptDelivered: null` on the `--json` object. `null`
and not `false`: the prompt is unadjudicated, and a `false` would be a verdict nobody
reached.

**#483's latch is extended, not duplicated**, exactly as PR #487 asked. It watches both
rows, keeps every ordering property that review established — opened before the session
exists, floored on the delivery stamp or the `eventsReplayed` marker so a previous start's
row cannot answer this command, held-and-re-judged rather than dropped — and gains the
wait itself. `promptAbandoned()` and the `promptDelivered` / `promptAbandonedReason` fields
keep their #483 meanings byte for byte; `awaitTurn` is untouched.

**#487's review observation (a) is dealt with where it matters.** A latch that did not cause
the `subscribe` has no floor and holds every row for ever. A silent `null` was survivable
for a report read after a turn; it is not survivable for a wait, which would simply never
end. It now reports `unavailable` — which is also what a dropped connection gives, via
`onDisconnected`. `unavailable` is deliberately **not** a failed delivery: the prompt may
have landed a second later, and calling it a loss would be #483's false report pointed the
other way, sending an operator to re-send text already in the composer.

### No clock, and none needed

`packages/cli` schedules nothing on a path to a harness's stdin (`no-prompt-timing.test.ts`),
and this wait does not want to. It ends on one of the Core's three answers — delivered,
abandoned at that harness's own `composerWaitMs` ceiling (#483), or the row `pty-manager`
appends when the PTY dies mid-delivery — or on the connection carrying them going down.

Nothing reads PTY bytes, output quietness or `reportsTurnStart` (#191, ADR 0026 D3, ADR
0033 D2). Readiness is the Core's report of a prompt it actually typed, never an inference
made here. It follows that a start with **no** prompt cannot use the flag and is refused,
rather than handed an instant meaningless success. `--wait-timeout` is refused too — a
second deadline could only end the wait early with nothing to report but that it had — and
so is `--wait`, which subsumes it.

### `send` is not gated, and the contract is the same on both sides

The issue offers the choice: gate the start, or gate the send. Gating the start is the one
that can be honest — a send is a raw write by design (#404), the Core adds no delivery
machinery to it, and there is nothing there to report. A caller that wants to know its
send will be read waits on the start that precedes it.

## Tests

- **Injected gateway** (`session-command.test.ts`): the wait, the non-zero abandon, the
  `unavailable` case that must not read as a loss, the `--json` shape, the bare-start
  `promptDelivered: null` plus its stderr line, silence on a start with no prompt, and all
  three refusals.
- **Real socket** (`in-process-core-session.test.ts`): three tests where the Core appends
  its row **inside the spawn round trip** — before the answer goes back and long before
  `wrap()` has a Task id to bind the latch to. That is the deaf window #487's review found,
  and no fake gateway can stage it. Delivered, abandoned, and a previous start's abandon
  row that must not answer this start. 8 consecutive runs of that file green.
  The suite's opt-in `spawn` registers a PTY id and nothing else — no binary, no process,
  no bytes — and the header says why the exception exists.

`pnpm -r typecheck` clean. `pnpm lint` 0 errors (43 warnings, all pre-existing — the same
count as this base). `pnpm test` green, all six stages, on a clean tree: root 24, sdk 22
(+2 skipped), shared 31, cli 50 (+1 skipped; 1237 tests), core 60, panel 169.

## Boundaries respected

- **`packages/sdk/src/core-session.ts` untouched** — `waitForTurnEnd` is #396's, and the
  file is not opened at all. Worth one note for sequencing: #396 is adding a disconnect
  handler to `waitForTurnEnd`, and this PR uses `CoreClient.onDisconnected` for the same
  class of hang in the CLI's own latch (`session-gateway.ts`). Same discipline, different
  file — no overlap, and no coordination needed beyond knowing both exist.
- **`packages/panel` untouched** (#484's wave). The Session still moves as it did, and
  `session:promptDelivered` is on the wire for the Panel to adopt.
- **`HARNESS_READINESS` and `packages/core/src/__tests__/fixtures/` untouched** (#277).
- **No codex hooks code touched** (#290).
- The generic backstop, `maxWaitMs`, `composerWaitMs` and the override table are all
  untouched: no timeout anywhere is lengthened by this change.

## Two limits, stated rather than left to be discovered

- **A bare `start` still exits 0.** That is #129 D6 and it stays. What it no longer does is
  say nothing: the stderr line and `promptDelivered: null` are the report, and
  `--await-prompt` is the way to get a verdict.
- **A start with no prompt has no readiness signal at all.** Delivery is the only honest
  one available (a marker-less harness has no "composer ready" moment to broadcast, and
  inventing one would be a claim rather than a report), so the flag is refused there rather
  than made meaningless.

---

## Revision 2 — the three blockers from the review of `3561d45`

The design above stands and is unchanged. What follows is what the review found on top of it: the flag exited
zero on a readiness nobody had established, in two ways, and where it did not it could hang.

### Blocker 1 — the floor was capped at a thousand events

`eventsReplayed`'s `lastEventId` is a **receipt for what the socket was sent**, not the end of the log.
`handleSubscribe` caps its tail at `EVENT_TAIL_LIMIT` (1 000) and then delivers the remaining history behind
the marker through `pushLiveEvents`, as ordinary `event` frames. Nothing prunes the log. So on any Core past a
few days of use the floor sat at ≈1000 and every historical `session:promptDelivered` above it was judged as
this command's verdict — **exit 0 and "this session can take a send now", before the Core had typed a
character**. The mirror case failed a start whose prompt landed perfectly.

**The fix: the Core reports `tipEventId` on the marker, and the latch floors on that.** It is read *before* the
tail, so every row that existed when the client asked is at or below it and every row appended afterwards is
above it.

**Why this and not the generation marker.** The review offered both. The `pty:spawn` generation marker is the
sharper idea and it only exists for the two verbs that *spawn* — so it would fix `--await-prompt` and leave
#483's own report path, which is an attach path, on the broken floor; and finding a historical `pty:spawn` for
an attached Session means reading past the very cap that is the defect. The tip fixes every path in the one
place they share. It is also #487's own reasoning made true rather than a second concept: that review argued
"every replayed row is necessarily below the marker", which holds exactly when the marker is the tip. And
`event-tip.ts` already names this shape as the clean fix in this repo's words — "the true tip carried on
`subscribeAck` as a field *beside* `fromEventId`, which no cursor consumes". It is on `eventsReplayed` rather
than `subscribeAck` because that frame is already plumbed to the listener the latch uses; the property that
matters — **no cursor consumes it** — is preserved, and `DurableCoreClient` passes it through without touching
its own cursor.

`event-tip.ts` is deliberately **not** changed: `events tail` needs the tip as a cursor it will consume and must
not skip history, so its "ask until the tail is empty" hunt stays exactly as it is.

**The protocol version does not move**, on ADR 0024 D11's additive exception. Absence of `tipEventId` is
detectable and yields a clean refusal of a new flag — a withheld affordance, never a silent wrong answer, which
is the degradation the exception exists to forbid. That same absence is what bounds blocker 2's first two
cases, so the field pays for itself twice.

### This repairs #483's already-merged report path, and that is why it is here

The hole is #487's floor, so `promptAbandoned()` had it too: on a bare `session wait` against a Core with more
than a thousand events, a stale abandon row cleared the floor and the command reported "the Core did not
deliver the starting prompt" about a Session that was fine. I checked #483's merged code for the shape and the
floor is the only place it lives — `awaitTurn` reads the accessor, the accessor reads the latch, and the latch
is what this PR already extends. So one change repairs both paths, and it belongs in this PR rather than a
follow-up because **here it terminates a wait**: the same wrong row that mis-filled a field on #483's path
returns an instant wrong exit code on this one.

What it does *not* do is make a bare `session wait` report an abandon that happened before it attached. That
stays #483's deliberate semantic — a bare wait reports what happens during it — rather than, as until now, an
accident of where the replay stopped. This is the residue of #487's observation (b): the floor it proposed a
generation marker to fix is now exact, so what is left there is a choice rather than a hole.

### Blocker 2 — it could wait for ever, and `--wait-timeout` is refused by design

Three bounds, all facts rather than clocks (`packages/cli` still schedules nothing —
`no-prompt-timing.test.ts` is green):

- **A Core that reports no tip cannot append the row either** — it is older than `session:promptDelivered`, or
  running with no event-log port. The absence is the answer, and it arrives one frame after the subscribe
  instead of never.
- **The harness exiting.** An `exit` frame is not a log row — it comes off the PTY subscription — so it arrives
  even from a Core whose appends are failing, and a harness that is gone will never take a prompt. Held before
  the latch is armed, like the event rows, so an exit inside the spawn round trip is not dropped.
- **A prompt that would never be delivered.** `""` and `"   "` are both dropped on the way to the harness
  (`sanitizeInitialInput`), so no delivery is built and no row is ever appended. Refused up front, on the same
  emptiness test the Core applies, named in a comment on both sides.

`--wait-timeout` stays refused: with those in place a second deadline could only end the wait early, with
nothing to report except that it had.

### Blocker 3 — on a marker-less harness, `delivered` meant "typed on the quiet gap"

`codex` has no `HARNESS_READINESS` row, so `composerOnScreen` is `true` for it from the first byte and the
prompt goes out when the screen stops moving. That is #483's generic backstop, deliberately preserved, and it
is a fine way to deliver — but roughly one codex boot in three settles with `Do you trust the contents of this
directory?` up, and reporting that as success is exactly the failure #395's acceptance criterion names by hand.

The Core now carries `composerObserved` on the row, recorded from the same predicate that gates the write
rather than derived later from the harness's name. A delivery that cannot claim it is reported as
**`unverified`**: non-zero, `promptDelivered: null`, `composerObserved: false`, and a sentence saying the Core
typed and saw nothing. Not `abandoned` — nothing was lost, and telling an operator to re-send would put the
text in twice.

This needs no change to `HARNESS_READINESS` (#277's), and it becomes a plain `delivered` for codex the moment
#493 lands its row, with no client change.

### The two non-blocking observations

- **(a)** The sentences claiming `--wait` "already reports the delivery" now say what it actually reports: a
  prompt the Core **gave up on**, with the rest inferred from a turn that ended. `awaitTurn` is left alone —
  wiring it to `latch.settled()` would give `--wait` the same three hang cases for no gain.
- **(b)** `consider()` keys on the Core's verdict rather than on "some conclusion was reached", so an abandon
  arriving after a dropped link or a harness exit still reaches `promptAbandoned()`.

### Tests

Six new, and **all six fail on `3561d45`** — verified by reverting the source files and running them:

| Test | How it fails on the reviewed code |
|---|---|
| a stale delivery row past the 1 000-row cap (real socket) | **exit 0** off the previous start's row |
| the harness exits with no verdict (real socket) | **hangs** — 60 s timeout |
| a Core that reports no tip (real socket) | **hangs** — 60 s timeout |
| a blind delivery, `composerObserved: false` (real socket) | **exit 0**, reported as delivered |
| `--await-prompt` on `""` / `"   "` / `"\t\n"` | accepted, exit 0 |
| the `unverified` shape and its sentence | no such outcome |

Plus two Core-side unit tests that the delivered event says `composerObserved: true` when a marker matched and
`false` when the quiet gap was all there was. The marker-less one uses an invented harness name rather than
`codex`, so #277 landing a codex row does not fail a test about the absence of one.

`pnpm -r typecheck` clean. `pnpm lint` 0 errors (43 warnings, the base's own count). `pnpm test` green, all six
stages, on a clean tree. `in-process-core-session.test.ts` 6/6 consecutive runs green.

### Base

Merged `fix/prompt-delivery-and-send-path` at `2926109` (#489, issue 290) with a plain `git merge` — clean, no
conflicts. Boundaries still held: `packages/sdk/src/core-session.ts` is not opened (#396), no
`HARNESS_READINESS` row and nothing under `packages/core/src/__tests__/fixtures/` (#277), no codex hooks code
(#290), nothing under `packages/panel`.

Refs #395

https://claude.ai/code/session_01YZDbm79jJjFsyVeP2dcPu3


qcentic-dev:
## Acceptance criteria — issue #395

Issue 395 states one numbered criterion. It is listed first, whole; the constraints the
ticket and its parent wave attach to it follow, each checked or explicitly not.

| # | Criterion (issue #395, verbatim) | Verdict |
|---|---|---|
| 1 | "A start then immediate send `--enter` delivers the send to the composer (or a documented wait), not to a trust dialog or an empty buffer that swallows the start prompt." | **Met — by the documented wait** |

**How 1 is met, and by which half of the "or".** The issue offers two answers and this PR
takes the second, deliberately. `session start … --await-prompt` blocks until the Core
reports the starting prompt **delivered**, and a delivered prompt is a composer that was
observed on screen, written into, and — where the harness confirms echo — seen to hold the
text. A `send --enter` after that cannot land on a trust dialog (delivery answers the
dialog, or abandons naming it) and cannot land in a buffer that swallows the start prompt
(the start prompt has already been submitted). It is documented in three places a caller
actually reads: `actana session --help` gains a section, the ADR 0026 amendment records
the decision, and the `actana-sessions` skill teaches the working shape.

The first answer — making a bare `start` block by default — is **not** taken, and the
reason is on the record rather than left implicit: #129 D6 says a `start` hands over and
hangs up, delivery runs on the harness's clock, and #483 gave opencode up to 90 s of it.
Blocking every start by default would change the latency of every script for a guarantee
most of them do not need. What a bare start does instead is stop *implying* the guarantee
— see the "does not claim" row below.

### The task's own constraints

| Constraint | Verdict |
|---|---|
| `start` must not claim a readiness it has not established | **Met** — bare `start` prints a line naming the gap and the flag, and `--json` carries `promptDelivered: null` (not `false`: unadjudicated, not lost) |
| The caller must be able to tell | **Met** — `session:promptDelivered` on the wire; `--await-prompt` turns it into an exit code and a `--json` field |
| Readiness not settled by reading PTY bytes or output quietness (#191, ADR 0026 D3) | **Met** — the client reads one event row off the log it is already connected to; nothing here parses a screen, and no byte-level or quietness heuristic is added on either side |
| `reportsTurnStart` consulted in no wait path (ADR 0033 D2) | **Met** — the string appears nowhere in the new code; the SDK source sweep is green and `packages/sdk/src/core-session.ts` is not opened at all |
| No success reported for a start whose prompt did not land | **Met** — `abandoned` exits non-zero and names the reason; `unavailable` (a dropped link) also exits non-zero and is reported as "no verdict", never as a delivery or as a loss |
| No generic timeout lengthened | **Met** — `maxWaitMs`, `composerWaitMs`, `HARNESS_PROMPT_DELIVERY_PROFILES` and the generic backstop are byte-identical; the new wait has no clock at all |
| #483's surface extended, not duplicated | **Met** — the same latch, the same floor, the same `promptAbandoned()` / `promptDelivered` / `promptAbandonedReason` fields; `awaitTurn` untouched |
| #483's event-ordering discipline followed | **Met** — the delivered row is appended in the same synchronous tick as the submit, ahead of any status the turn produces; the latch still opens before the session exists and still takes a floor |
| Unit + real-socket coverage | **Met** — 9 injected-gateway tests, 3 on a real socket where the row lands *inside the spawn round trip*; that file 8/8 on consecutive runs |
| `Refs #395` in the body, no `Closes` for any issue | **Met** — one `Refs #395`; zero `Closes`/`Fixes #`/`Resolves` in the body or in any of the three commits |
| Base is `fix/prompt-delivery-and-send-path` | **Met** |
| PR left as a draft, nothing merged, no `main` or `beta/*` touched | **Met** |
| Conventional Commits on every commit, body lines ≤ 132 | **Met** — commitlint over all three commits: `found 0 problems, 0 warnings` |

### Boundaries

| Boundary | Verdict |
|---|---|
| `packages/sdk/src/core-session.ts` — #396's `waitForTurnEnd` | **Held** — file not opened. Sequencing note, not a conflict: #396 adds a disconnect handler to `waitForTurnEnd`; this PR uses `CoreClient.onDisconnected` for the same class of hang in the CLI's own latch (`session-gateway.ts`). Same discipline, different file |
| `packages/panel` (#484) | **Held** — untouched |
| `HARNESS_READINESS` and `packages/core/src/__tests__/fixtures/` (#277) | **Held** — untouched |
| codex hooks (#290) | **Held** — untouched |

### Not done, and why

- **`session send` is not itself gated.** The issue's "or `send` must wait for that gate"
  is the branch not taken. A send is a raw write by design (#404) with no delivery
  machinery behind it and therefore nothing to report; gating the start is the half that
  can be honest, and it is the half the same contract is expressed on.
- **A start with no prompt cannot use `--await-prompt`** — refused rather than answered
  with an instant success. There is no delivery to report, and the only other candidate
  signal (a Core-side "composer ready" broadcast) does not exist for a marker-less harness
  and would be a claim rather than a report if invented for one.
- **The Panel does not surface any of this.** Out of scope by boundary (#484 owns that
  package this wave); the row is on the wire for it to adopt.

qcentic-dev:
## CI on `3561d45` — one red check, and it is not this PR's

**22 checks pass. One fails: `Dependency Audit`.** It is a registry outage, not a finding,
and not the known `core-session-backstop.test.ts` flake (#491) either — `Unit Tests` is
green.

`pnpm audit --prod --audit-level high` never got an answer out of npm:

```
[WARN] POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk error (23). Will retry in 10 seconds. 2 retries left.
[WARN] POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk error (23). Will retry in 1 minute. 1 retries left.
[23] The operation was aborted due to timeout
TimeoutError: The operation was aborted due to timeout
```

No advisory was reported; the request timed out after three tries, and `pnpm` exits 1 on
that the same way it does on a real finding.

**Evidence that it is the endpoint and not the tree:**

- **Re-run twice** (22:24 and 22:38 UTC) — byte-identical failure both times.
- **Reproduces off CI**: the same command on a developer machine, different network,
  different `pnpm` install, gives the same three retries and the same `TimeoutError`. A
  runner-local problem would not.
- **This PR changes no dependency.** `git diff a2d6f6a..HEAD --name-only` matches no
  `package.json`, no `pnpm-lock.yaml`, no `pnpm-workspace.yaml` — nothing that an audit
  reads.
- **The job was green on this repo at 21:50 UTC** (#277's branch) and has failed on
  everything attempted since; the endpoint went away in between.

Nothing to fix here. It will go green on the next run once npm's advisories endpoint is
answering; re-running it a third time now would only reproduce the outage.

mrfqcentic (changes requested):
## Review of #494 against #395 — changes requested

Reviewed at `3561d45` on base `fix/prompt-delivery-and-send-path` (`a2d6f6a`, the merge of #487). The base is an ancestor of the head, so the PR is exactly its three commits and merges as a fast-forward. Every claim below was checked by reading the enforcing line, not from the description.

**The architecture is right, and it is the hard half.** The positive row genuinely had to exist, and the reasoning for why — #483 could read absence as delivery because it had already waited for a turn to end, and a `start` has not — is correct. **It composes with #483 rather than papering over it:** `harness-prompt-delivery.ts` is untouched, not one #483 test is deleted or weakened (the only four deleted lines in the whole diff are fixture plumbing in `in-process-core-session.test.ts`), `promptAbandoned()` and the `promptDelivered`/`promptAbandonedReason` fields keep their meanings, and the latch is extended with every ordering property #487's review established intact. The claim that the delivered row is appended in the same synchronous tick as the carriage return holds: `harness-prompt-delivery.ts:1199-1213`, `write("\r")` then `emit` with no await between them.

**Three things block, and the first is the one this train exists to remove.**

---

### Blocker 1 — the floor is capped at 1000 events, so on any Core in normal use `--await-prompt` exits 0 off a *previous* start's delivery row

`session-gateway.ts:441` (`start`) and `:492` (`resume`) call `wrap()` **without `afterEventId`**, so `cursor = 0` at `:1066` and `floor()` (`:909-912`) is nothing but `replayedThrough` — the `eventsReplayed` marker. That marker is not the end of the log:

- `session-gateway.ts:1000` — `client.subscribeEvents()`, no argument.
- `core-client.ts:1009` — `subscribeEvents(lastEventId = 0)`.
- `pty-core-link-server.ts:1967` — `readEventTail(0, EVENT_TAIL_LIMIT)`, and `EVENT_TAIL_LIMIT = 1_000` at `:570`.
- `packages/shared/src/event-log.ts:124-131` — `WHERE event_id > ? ORDER BY event_id ASC LIMIT ?`. That is the **oldest** 1000 rows. Your own test pins it: `readEventTail(db, 0, 3)` → `[1, 2, 3]` (`__tests__/event-log.test.ts:142`).
- `pty-core-link-server.ts:1969-1980` — `lastSent` is the last row of that capped tail, and `eventsReplayed { lastEventId: lastSent }` therefore reports **≈1000**, not the head of the log.
- `pty-core-link-server.ts:1152-1165` — `pushLiveEvents` then delivers events 1001, 1002 … through `deliverEvent`, i.e. as ordinary `event` frames **after** the marker, indistinguishable from live ones.
- `session-gateway.ts:959-963` — only the first marker counts, so `replayedThrough` stays at ≈1000 for the life of the command.

And the log is unbounded: `core-query.ts:246` — *"Nothing in this repo prunes `event_log` — there is no `DELETE FROM event_log` anywhere."* There is none; I grepped.

So on any Core whose log has passed a thousand rows — days of ordinary use — every historical `session:promptDelivered` and `session:promptAbandoned` row above id 1000 clears the floor and is judged as this command's verdict:

```bash
actana session resume task_X "go" --await-prompt
# task_X's start last Tuesday left session:promptDelivered at event 5000; 5000 > 1000
# → conclude({outcome:"delivered"}) fires on the replayed backlog
# → exit 0, "This session can take an `actana session send …` now."
# …before the Core has typed a character of *this* prompt.
```

That is a readiness nobody established, returned instantly, which is the defect in the ticket title. The mirror case is as bad in the other direction: an old abandon row fails a start whose prompt landed perfectly.

The in-process test `does not read a previous start's abandon row as this start's verdict` passes only because its log holds a handful of rows and never reaches the cap. It is testing the intended rule, not the shipped one.

**This is #487's floor, so the hole predates this PR** — but there it only made `promptAbandoned()` answer wrongly on a path that had already waited for a turn. Here it *terminates the wait*, which is what turns a wrong field into an instant wrong exit code, and it is what makes it this PR's to fix. #487's review reasoned that "every replayed row is necessarily below [the marker]"; that reasoning is sound only while the log is under `EVENT_TAIL_LIMIT`, and I should have caught it there. The fix probably wants a real generation marker rather than a bigger cap — rows since this Session's own `pty:spawn`, which is the shape #487's review floated as observation (b) — or, cheaper, subscribing from `getLastEventId()` so the marker is the actual head of the log.

### Blocker 2 — `--await-prompt` can wait for ever, and it refuses the only flag that would end it

`session-command.ts:546-547` starts an unbounded wait and `awaitPromptFlagRefusal` (`:632`) rejects `--wait-timeout` for it. That is right when the Core will eventually answer. Three reachable cases where it never does:

1. **An older Core.** Nothing gates on protocol version — a Core that has never heard of `session:promptDelivered` appends nothing and the command blocks until Ctrl-C. Contrast `send --wait`, which for exactly this class of problem carries a version gate and an explicit refusal (`session-gateway.ts:652-668`: *"The version gate refuses a Core too old to stamp… This covers the ways a Core on this version still answers 0: no event-log port wired, or an `appendEvent` that failed. Both are documented… and neither is a reason to guess."*). `--await-prompt` needs the same sentence.
2. **A Core with no event-log port wired**, or an `appendEvent` that throws — `core-entry.ts` catches and logs it, and the row is simply gone.
3. **An empty or whitespace-only prompt.** `session-command.ts:384` and `:430` gate the refusal on `prompt.text !== null`, and `readText` does not trim. For `""`, `session-gateway.ts:721` (`...(opts.prompt ? { prompt: opts.prompt } : {})`) drops it before it ever reaches the Core; for `"   "` it reaches the Core and `sanitizeInitialInput` (`pty-manager.ts:355`) trims it to `undefined`, so `:742` builds no `HarnessPromptDelivery`. Neither row is ever appended, and the PTY-exit abandon row at `:882` is guarded on `promptDelivery` being present, so that does not fire either. `actana session start web "" --await-prompt` hangs for ever. One line: gate `awaitPromptTextRefusal` on the same emptiness test the Core applies.

A wait that can hang is a worse failure than the exit-0 it replaced, and `SKILL.md` now tells an orchestrating agent to make it the default pattern.

### Blocker 3 — on codex, `delivered` means "typed blind on the quiet gap", and the flag reports it as a composer that was seen

`session-command.ts:578` returns `EXIT_OK` on `outcome === "delivered"` and `:569` prints *"The Core delivered the starting prompt. This session can take an `actana session send …` now."* `pty-manager.ts:803` appends that row on `event.phase === "delivered"` with **no check that a composer was ever observed**. For three of the four shipped harnesses that is sound. For the fourth:

- `harness-prompt-delivery.ts:588` — `HARNESS_READINESS` has rows for `opencode`, `cursor-cli` and `claude-code`. **`codex` has none** (that row is #277's; PR #493 is adding it and is not on this base).
- `:500-504` — `NO_READINESS` is `{ composer: [], confirmEcho: false, maxPromptWrites: 1 }`.
- `:612` — `composerOnScreen` returns `true` when `composer.length === 0`. It does not look at the screen.
- `:1013` — `echoConfirmed()` returns `true` immediately when `!confirmEcho`.
- `BLOCKING_DIALOGS` scopes `folder-trust` to `claude-code` and `cursor-cli`. **codex is not there either.**

So on codex the delivery resolves on the quiet gap alone, nothing observed and nothing echoed back. What is on screen at that moment is documented by your own sibling PR #493, off 21 live codex-cli 0.153.0 boots: *"In a directory codex does not trust it paints that same composer placeholder at ~200 ms, clears the screen at ~630 ms and replaces it with `Do you trust the contents of this directory?`. The quiet gap expires at ~990 ms, with the dialog up and no composer anywhere — and codex has no `BLOCKING_DIALOGS` entry, so nothing else in this module is watching."*

`actana session start <project> "fix it" --harness codex --await-prompt` therefore exits 0, `promptDelivered: true`, on a prompt typed into a trust dialog — **the failure #395's acceptance criterion names by hand** ("not to a trust dialog or an empty buffer that swallows the start prompt"), reported as success by the flag whose whole purpose is to answer that question.

Two things make it this PR's rather than #277's. **The claim is new**: before this change no command asserted anything about readiness. And **the guarantee is now in the durable record, contradicting the same document three screens up.** The amendment says *"a delivered prompt is a composer that was seen on screen, written into, and — where the harness confirms echo — seen to hold the text."* D3a of that same ADR says *"`claude-code`, `codex` and `cursor-cli` have no entry and are therefore unchanged"*, and D7 adds that *"what a harness with no entry gets is D3, D6 and D8"* — D3 being the quiet gap. Two of those three have since been given rows. codex has not. `session-gateway.ts:919-925` then deliberately discards the payload, so a client could not tell the two kinds of `delivered` apart even if it wanted to.

Neither fix needs `HARNESS_READINESS` (#277's). Either carry the fact on the wire — the Core knows whether a marker was matched or whether it typed on the clock; add `composerObserved` to `CoreLinkSessionPromptDeliveredPayload` and have the latch decline to call the blind case `delivered` — or, if that is out of scope, refuse `--await-prompt` for a marker-less harness the way you already refuse it for a promptless start (*"a zero exit on a report nobody made is how a caller comes to trust one"* — your words, and they apply here), and strike the "seen on screen" sentence from the ADR.

---

### Two observations, neither blocking

**a. `--wait` still infers delivery from an absence, and the docs now claim it does not.** `awaitTurn` computes `promptDelivered: abandoned === null` — unchanged #483 behaviour, defensible there because a turn ended. But `awaitPromptFlagRefusal` tells the operator "`--wait` already reports the delivery" and `SKILL.md` says "`--wait` already includes it". The positive row now exists and `awaitTurn` does not read it, so the two flags do not report the same fact. Either wire `--wait` to `latch.settled()`, or soften both sentences.

**b. `consider()` latches on the first outcome, not on the first abandon.** `session-gateway.ts:917` changed `if (abandoned) return` to `if (report) return`, and `onDisconnected` sets `report`. On a client that reconnects and replays — a `DurableCoreClient` handed in through the injectable `OpenSessionGateway`, which is a public type — an abandon arriving after a drop is discarded and `promptAbandoned()` answers a clean delivery on the `--wait` path. Unreachable through `openSessionGateway` today, but the guard should key on the Core's verdict rather than on any conclusion. The same short-circuit also drops a genuine abandon on an attached wait spanning two deliveries of one Task.

---

### Verified, at the line

| Claim | Verdict |
|---|---|
| `start` no longer claims a readiness it has not established | **Partially met** — the bare-start half is right (stderr line, `promptDelivered: null` not `false`); the `--await-prompt` half claims one in blockers 1 and 3 |
| A caller can tell whether the starting prompt landed | **Partially met** — the mechanism is correct and the `unavailable` third outcome is well judged (non-zero, and correctly not called a loss); the answer is wrong on a stale-floor Core and on codex |
| Nothing reports success for a start whose prompt did not land | **Not met** — blockers 1 and 3 |
| Readiness not settled by PTY bytes or output quietness (#191, ADR 0026 D3) | **Met on the client** — no screen read and no quietness timer anywhere in the diff; but see blocker 3 for what the Core's verdict reduces to on a marker-less harness |
| No wait path consults `reportsTurnStart` (ADR 0033 D2) | **Met** — the identifier appears nowhere in the diff; the sweep in `packages/sdk/src/__tests__/core-session.test.ts` reads `core-session.ts`, which this PR does not open. Unit Tests green on the head |
| Not merely a lengthened generic timeout | **Met** — `harness-prompt-delivery.ts` is not in the diff at all; `maxWaitMs`, `composerWaitMs`, `quietGapMs` and the override table untouched; `packages/cli` gains no scheduler |
| #483's event-ordering discipline followed | **Met in shape, broken by blocker 1 in effect** — latch opened at the top of `attached()` (`:599`) and `begin()` (`:714`) before any Core question; `onEvent` (`:946`), `onEventsReplayed` (`:959`) and `onDisconnected` (`:979`) all registered before `subscribeEvents()` (`:1000`); held-and-re-judged intact; every throw path closes the latch (`:612`, `:620`, `:680`, `:731`, `:1094`); the delivered row is emitted inside `submit()` in the same tick as the `\r`. The floor is the one clause that does not hold |
| Composes with #483, does not paper over it | **Met** — `harness-prompt-delivery.ts` untouched, zero test removals, #483's accessors and fields byte-compatible |
| `Refs #395`, no `Closes` | **Met** — one `Refs #395` in the body, three across the commits, zero closing keywords |
| Base is `fix/prompt-delivery-and-send-path` | **Met** — not `main`, not a beta branch |
| #396 boundary (`packages/sdk/src/core-session.ts`) | **Held** — #494 does not open the file. #492 (#396) adds `client.onDisconnected(… onLinkLost …)` to the `waitForTurnEnd` path there (`:239`, `:280`). Zero overlap; `git merge-tree` of #492 and #494 is clean; `onDisconnected` already exists on the base at `core-client.ts:1057` |
| #277 boundary (`HARNESS_READINESS`, `packages/core/src/__tests__/fixtures/`) | **Held** — neither touched |
| #290 boundary (codex hooks) | **Held** — untouched |
| `packages/panel` | **Held** — untouched |
| Merges cleanly against the moved base | **Yes** — base `a2d6f6a` is an ancestor of the head; fast-forward |
| CI on the head | **Green but for `Dependency Audit`**, which failed on `TimeoutError` reaching `registry.npmjs.org/-/npm/v1/security/advisories/bulk`. No manifest is in the diff; infrastructure, not this PR. **Unit Tests passed** — #491's `core-session-backstop.test.ts` flake did not fire |

---

## Verdict

**Changes requested.** The design is right, it composes with #483 properly, and the ordering work is careful — that is the hard half and it is done. But the flag exits zero on a readiness nobody established in three separate ways: off a replayed row from a previous start on any Core with more than a thousand events, and off a blind type on codex; and where it does not lie it can hang for ever with the only bounding flag refused. A false success is the one outcome this ticket cannot ship with, and blocker 1 makes it the common case rather than the edge.

Blocker 1 is the one to fix first, and fixing it properly (a generation marker, or subscribing from the log's actual head) also retires the shape #487's review left as observation (b).

The draft is left as a draft; marking it ready is yours.

qcentic-dev:
Thank you — the review was right on all three, and the diagnosis of blocker 1 is the one that mattered most:
I had read #487's "every replayed row is necessarily below the marker" as a property rather than as a
conditional, and never checked what `EVENT_TAIL_LIMIT` did to it. Pushed as `97359bf`, on top of a plain merge
of the base at `2926109` (#489).

## Fixed

**Blocker 1 — the floor is the log's actual end.** `eventsReplayed` now carries `tipEventId`, read *before* the
tail in `handleSubscribe`, and the latch floors on that instead of on `lastEventId`.

*Which of the two fixes, and why.* You offered a `pty:spawn` generation marker or the log's head. The
generation marker is the sharper idea and it only exists for the two verbs that spawn — so it would have fixed
`--await-prompt` and left #483's own report path, which is an *attach* path, on the broken floor; and finding a
historical `pty:spawn` for an attached Session means reading past the very cap that is the defect. The tip
fixes every path in the one place they share, and it is your second suggestion realised in the only way a
client can have it: there is no client-side `getLastEventId()` for a Core's log, and `event-tip.ts` says so in
its header. That file also names this exact shape as the clean fix — "the true tip carried on `subscribeAck` as
a field *beside* `fromEventId`, which no cursor consumes". I put it on `eventsReplayed` rather than
`subscribeAck` because that frame is already plumbed to the listener the latch uses; the property that matters,
that no cursor consumes it, is kept, and `DurableCoreClient` passes it through without touching its own cursor.
`event-tip.ts` itself is untouched: `events tail` needs the tip as a cursor it *will* consume and must not skip
history, so its hunt stays.

**And the same change repairs #483's merged report path, which is why it is in this PR.** I checked for the
shape you describe and the floor is the only place it lives: `awaitTurn` reads `promptAbandoned()`,
`promptAbandoned()` reads the latch, the latch is what this PR already extends. Before this, a bare
`session wait` against a Core with more than a thousand events could report "the Core did not deliver the
starting prompt" about a Session that was fine, off a row from a previous life. It belongs here rather than in
a follow-up because here the same wrong row does not mis-fill a field, it *terminates a wait* with the wrong
exit code. Said in the PR description too.

What it does not do is make a bare `wait` report an abandon that predates the attach. That is now #483's
deliberate semantic rather than an accident of where the replay stopped — which is what I take "retires the
shape of observation (b)" to mean: the floor the generation marker was proposed to fix is exact, so what is
left there is a choice.

**The protocol version does not move,** on D11's additive exception: absence of `tipEventId` is detectable and
yields a clean refusal of a new flag, never a silent wrong answer. That absence is also what bounds two of
blocker 2's three cases, so the field pays for itself twice.

**Blocker 2 — three bounds, all facts, no clock.** A Core that reports no tip cannot append the row either
(older than it, or no event-log port) — the absence is the answer, one frame after the subscribe rather than
never. The harness exiting ends the wait, off an `exit` frame rather than a log row, so it survives a Core whose
appends are throwing; it is held before the latch is armed, like the event rows, so an exit inside the spawn
round trip is not dropped. And a prompt of nothing but spaces and control characters is refused up front, on
the same emptiness test `sanitizeInitialInput` applies, named in a comment on both sides. `--wait-timeout` stays
refused, and `no-prompt-timing.test.ts` is green.

**Blocker 3 — `composerObserved` on the wire.** Recorded in `writePrompt` from the same predicate that gates
the write, rather than derived later from the harness's name, so it cannot drift from the gate. A delivery that
cannot claim it is reported as `unverified`: non-zero, `promptDelivered: null`, `composerObserved: false`, and
a sentence saying the Core typed and saw nothing. Deliberately not `abandoned` — nothing was lost, and sending
an operator to re-send would put the text in twice. It needs no `HARNESS_READINESS` change and becomes a plain
`delivered` for codex the day #493 lands its row.

**And the ADR, which was the part of blocker 3 that would have outlived the code.** "A composer that was seen
on screen" is now scoped to the harnesses that can be vouched for, with D3a's own sentence quoted against the
first draft.

**Observation (a).** Softened rather than wired: `--wait` now says what it actually reports — a prompt the Core
gave up on, with the rest inferred from a turn that ended. Wiring `awaitTurn` to `latch.settled()` would hand
`--wait` the same three hang cases for no gain.

**Observation (b).** `consider()` keys on the Core's verdict, not on "some conclusion was reached", so an
abandon arriving after a dropped link or a harness exit still reaches `promptAbandoned()`.

## Coverage

Six new tests, and I reverted the source files to `3561d45` to confirm **all six fail on the code you read** —
the two bound tests fail by hanging to the 60 s timeout, which is the defect itself; the stale-row and blind
tests fail with **exit 0**. Four are on a real socket, because none of them can be staged with a fake gateway:
one is about what `handleSubscribe` sends, one about a frame that is not a log row, one about a payload field
crossing the wire. Plus two Core-side unit tests on `composerObserved`, the marker-less one written against an
invented harness name rather than `codex` so #277's row does not fail a test about the absence of one.

`pnpm test` green, all six stages, on a clean tree. `pnpm -r typecheck` clean, `pnpm lint` 0 errors (43
warnings, the base's own count). `in-process-core-session.test.ts` 6/6 consecutive.

## Deferred, and why

- **A bare `session wait` still does not report an abandon that predates its attach.** Above; #483's semantic,
  now a choice rather than an accident. A generation marker keyed on `pty:spawn` is still the way to change it
  and would want its own ticket, since it needs a way to read that row past the replay cap.
- **`awaitTurn` still infers `promptDelivered` from the absence of an abandon.** Observation (a); softening the
  claim was the honest half, and changing #483's field semantics is not this ticket's.
- **`event-tip.ts` does not use the new tip.** It needs a cursor it will consume; taking the tip as one would
  skip history, which is the trade its header refuses.
- **`packages/panel` surfaces none of it.** #484's wave; the rows are on the wire.

The draft is still a draft.

qcentic-dev:
## CI on `97359bf` — 17 pass, 6 skipping, 1 fail

**`Unit Tests` is green**, so #491's `core-session-backstop.test.ts` flake did not fire. So are `Lint`,
`Typecheck`, `Conventions`, `Secret Scan`, `Train rules`, both installer E2Es, both tarball smokes, the Panel
service seam E2E and both container image builds.

The one red check is **`Dependency Audit`**, and it is the same npm outage that was red on the previous head —
still going, ten hours on. `pnpm audit --prod --audit-level high` never gets an answer:

```
[WARN] POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk error (23). Will retry in 10 seconds. 2 retries left.
[WARN] POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk error (23). Will retry in 1 minute. 1 retries left.
TimeoutError: The operation was aborted due to timeout
```

No advisory is reported; the request times out after three tries and `pnpm` exits 1 on that the same way it
does on a real finding.

- **Re-run once on this head** (00:21 UTC) — identical failure.
- **Reproduces off CI** on a developer machine, different network, different `pnpm` install.
- **This revision changes no dependency**: `git diff 3561d45..HEAD --name-only` matches no `package.json`, no
  `pnpm-lock.yaml`, no `pnpm-workspace.yaml`.

Nothing to fix here; it will go green whenever the advisories endpoint answers again.

mrfqcentic (approved):
## Re-review of #494 at `97359bf` — approved

Re-reviewed at `97359bf`, base `fix/prompt-delivery-and-send-path` at `2926109` (#489), which is an ancestor of the head. Every claim below was checked by reading the enforcing line and the test that would fail without it, not from the description.

**All three blockers are closed, and closed the right way** — the fix is a fact carried on the wire in each case, not a widened window or a clock. The reply also does the thing I asked for that was easiest to skip: blocker 1's repair is applied to #483's already-merged report path in this PR, and the description says so and says why.

---

### Blocker 1 — the floor is now the log's actual end, not a receipt for a capped tail

**Closed, and closed rather than narrowed.**

- `pty-core-link-server.ts:1966-1976` — `const tipEventId = this.eventLog?.getLastEventId();` sits **above** `readEventTail(fromEventId, EVENT_TAIL_LIMIT)`. That ordering is the whole proof: `getLastEventId` is `SELECT MAX(event_id) FROM event_log` (`packages/shared/src/event-log.ts:101-106`), synchronous, so every row that existed when the subscribe was taken up is at or below the number, and every row appended afterwards is above it. There is no window left — not a smaller one.
- `:1991-1998` — sent **beside** `lastEventId`, and omitted rather than zeroed when there is no log. `core-link-transport.ts:393-400` keeps that distinction (`typeof … === "number" ? … : undefined`, never a `0` that would read as a real and very low tip).
- `durable-core-client.ts:242-252` — the cursor is still `lastEventId` and only that; the tip rides through untouched. The property that mattered — no cursor consumes it — holds.
- `session-gateway.ts:925-956` — `tipAtSubscribe` replaces `replayedThrough`, `floor()` is `Math.max(cursor, tipAtSubscribe ?? 0)` / `tipAtSubscribe`. `event-tip.ts` is untouched, which is right: it needs a cursor it will consume.

**Why this over the generation marker: the reasoning is better than mine.** A `pty:spawn` marker exists only for the verbs that spawn, so it would have fixed `--await-prompt` and left #483's attach path on the broken floor — and finding a historical `pty:spawn` for an attached Session means reading past the very cap that is the defect. Correct on both counts. The tip fixes every path at the one line they share.

**Coverage that produces the false success first.** `in-process-core-session.test.ts:765` — 1 200 filler rows, then a `session:promptDelivered` for `task_done` at id 1 201 with `composerObserved: true`, asserted `> 1_000`. `arrayEventLog.readEventTail` honours the limit (`in-process-core.ts:384-387`, `.slice(0, limit)`), so the marker lands at 1 000 and the stale row arrives behind it through `pushLiveEvents`. On the reviewed code `1201 > 1000` clears the floor and the command exits 0 off last Tuesday's row; the test asserts `EXIT_FAILURE` and this start's own abandon reason. Deterministic, not timing-dependent.

**Protocol version held at ADR 0024 D11's additive exception** — sound here, because the absence of `tipEventId` is detectable and produces a clean refusal of a new flag rather than a silent wrong answer, and that same absence is load-bearing for blocker 2.

### The #483 repair — done, correct, and explained where it belongs

`session-gateway.ts:1213` — `promptAbandoned: () => opts.latch.reason()`; `reason()` returns `abandoned`; `abandoned` is only ever set inside `consider()`, which gates on `floor()`. So the merged report path reads the repaired floor by construction: one change, both paths, exactly as claimed. Before it, a bare `session wait` against a Core past a thousand rows could report *"the Core did not deliver the starting prompt"* about a Session that was fine.

The description carries it in its own section — *"This repairs #483's already-merged report path, and that is why it is here"* — and gives the reason it is not a follow-up: on this path the same wrong row does not mis-fill a field, it terminates a wait with the wrong exit code. ADR 0026 records the same thing durably (`:402-424`). The residue of #487's observation (b) is correctly characterised: a bare wait still declines to report an abandon that predates its attach, but that is now #483's deliberate semantic rather than an accident of where the replay stopped.

### Blocker 2 — bounded by three facts, and none of them is a clock

**Closed.** `packages/cli` still schedules nothing on a path to a harness's stdin.

1. **No tip, no wait** — `session-gateway.ts:1039-1059`: `tipEventId === undefined` concludes `unavailable` one frame after the subscribe. Covers both the pre-`session:promptDelivered` Core and the one with no event-log port; neither could ever append the row. Test: `in-process-core-session.test.ts:839` (`eventLog: false`) — hangs to the 60 s timeout on the reviewed code.
2. **The harness exiting** — `:1092-1104`, `client.onExit` filtered on this Session's own `ptyId`, registered before `subscribeEvents()` (`:1022/1035/1077/1092` all precede `:1117`), and **held before arm** in `exitedPtys` with the replay applied at `arm()` (`:1129-1134`). That is the same deaf-window discipline as the event rows, applied to a frame that is not a log row — so it survives a Core whose appends are throwing. Reported as "no verdict", never as a loss, which is right: the text may have gone in before the process died. Test: `:807`.
3. **A prompt that would never be delivered** — `session-command.ts:715-747`. `deliversText()` mirrors `sanitizeInitialInput` (`pty-manager.ts:364-375`) and I checked the two agree on emptiness for every input shape: the Core's whitespace-run→space step can only produce a space, which `trim()` removes, so `""`, `"   "`, `"\t\n"`, `"\v"`, lone C0/DEL and NBSP all resolve the same on both sides. The refusal fires at `:391`/`:437`, before `withGateway` — no Core connection is opened at all. Test: `session-command.test.ts:558` over `["", "   ", "\t\n"]`, plus `" go "` still accepted.

`--wait-timeout` staying refused is now correct rather than merely asserted: with those three in place a second deadline could only end the wait early with nothing to report.

### Blocker 3 — a blind type can no longer report success

**Closed, and closed on the Core's own predicate rather than on a harness name.**

- `harness-prompt-delivery.ts:1208-1215` — `composerObserved = composerOnScreen(this.screen, this.readiness) && this.readiness.composer.length > 0`, computed inside `writePrompt` **before** `this.screen = ""`. The second conjunct is what excludes the marker-less case, since `composerOnScreen` returns `true` for an empty marker list without looking at anything (`:611-616`). Reading it at the write rather than deriving it later is the right call — there is no second opinion to drift.
- It also fails closed on the one path where a marker harness types without the marker on screen (`deadlinePassed` set at `:1287`/`:1342` and the screen having moved since): `composerObserved` comes out `false`. Conservative in the safe direction.
- `:1233-1241` → `pty-manager.ts:839-845` → `core-entry.ts:343-350` — carried whole to the row, not inferred anywhere downstream.
- `session-gateway.ts:975-1002` — `composerObserved !== true` becomes `unverified`, with an unparseable payload also failing closed. `session-command.ts:598` — `promptDeliveredField` returns `null` (not `false`), and `:595` exits `EXIT_FAILURE` for anything that is not `delivered`. Keeping this distinct from `abandoned` is right: nothing was lost, and telling an operator to re-send would put the text in twice.
- The `--json` object carries `composerObserved` on both the delivered and unverified outcomes, so a script tells the two apart without parsing English.

**Coverage:** `in-process-core-session.test.ts:859` (`onPromptDelivery: "blind"`, real socket, exit 0 and `promptDelivered: true` on the reviewed code), `session-command.test.ts:507` for the shape and the sentence, and two Core-side unit tests at `harness-prompt-delivery.test.ts:1440-1477`. The marker-less unit test deliberately uses `some-harness-invented-tomorrow` rather than `codex`, so #277 landing a codex row does not fail a test about the absence of one — good foresight.

**And the durable half.** ADR 0026 `:333-352` no longer says a delivered prompt *"is a composer that was seen on screen"*; it scopes the claim to harnesses that can be vouched for and quotes D3a against the first draft. That was the part of blocker 3 that would have outlived the code.

### The two observations

- **(a)** Softened rather than wired, and the sentences now say what `--wait` actually reports — a prompt the Core gave up on, with the rest inferred from a turn that ended (`session-command.ts:686-691`, SKILL.md, both `orchestration-skill-payload.ts` copies, which stay byte-identical to each other). Leaving `awaitTurn` alone is the right trade: wiring it to `latch.settled()` would hand `--wait` all three hang cases for no gain.
- **(b)** `consider()` now keys on `if (abandoned) return` (`:961`), so an abandon arriving after a dropped link or a harness exit still reaches `promptAbandoned()`. `conclude()` remains idempotent, so `--await-prompt`'s own answer is unchanged by it.

---

### Re-confirmed, not regressed

| Previously accepted | Still true |
|---|---|
| The positive row is justified, and #483 could read absence as delivery only because it had already waited for a turn | **Yes** — argument intact in the description and ADR; unchanged by this revision |
| #483's latch extended, not duplicated; #487's ordering properties intact | **Yes** — one latch; opened before the session exists; all four listeners (`:1022`, `:1035`, `:1077`, the new `:1092`) registered before `subscribeEvents()` at `:1117`; held-and-re-judged intact; `floorless` refusal from #487 observation (a) unchanged; every throw path still closes |
| The delivered row is emitted in the same synchronous tick as the carriage return | **Yes** — `harness-prompt-delivery.ts:1219-1241`: `write("\r")` then `emit`, no await between; the only change is one more field on the payload |
| No PTY bytes or output quietness settle readiness on the client (#191, ADR 0026 D3) | **Yes, and strengthened** — the client reads one row and one process-exit frame; `composerObserved` is precisely the field that stops the Core's quiet gap from being *reported* as readiness |
| No wait path consults `reportsTurnStart` (ADR 0033 D2) | **Yes** — the sole occurrence in the diff is an unchanged context line in `reportStartedSession`; `packages/sdk/src/core-session.ts` is not opened |
| No generic timeout lengthened | **Yes** — `HARNESS_READINESS`, `maxWaitMs`, `composerWaitMs`, `quietGapMs` and the override table are untouched; the table's only appearance in the diff is a doc reference |
| `Refs #395`, no `Closes` | **Yes** — one in the body, three across the commits, zero closing keywords |
| Base is `fix/prompt-delivery-and-send-path` | **Yes**, and still a draft |
| #396 boundary — `packages/sdk/src/core-session.ts` | **Held** — not in the diff. `git merge` of #492 (#396) against this head is **clean** |
| #277 / #484 / #290 boundaries | **Held** — no `HARNESS_READINESS` row, nothing under `packages/core/src/__tests__/fixtures/`, nothing under `packages/panel`, no codex hooks code |
| Merges against the base | **Fast-forward** — the base is an ancestor of `97359bf` |
| CI on `97359bf` | **Green but for `Dependency Audit`**, which is `TimeoutError` on `registry.npmjs.org/-/npm/v1/security/advisories/bulk` after three retries — I read the job log. No manifest is in the diff. **Unit Tests passed**, so #491's `core-session-backstop.test.ts` flake did not fire |

---

### Three notes, none blocking, none needing a change before this lands

1. **#493 (#277) will conflict with this branch when it merges into the base.** `git merge` of `fix/codex-harness-readiness` against this head conflicts in `docs/adr/0026-…md` and `packages/core/src/__tests__/harness-prompt-delivery.test.ts` — both are "both sides appended to the tail of the same file", mechanical, keep-both. Semantically the two compose exactly as this PR predicts: once codex has a row, `composerObserved` comes out `true` for it with no client change, and the marker-less unit test here is written against an invented harness so it survives. Worth knowing which of you rebases second. #492 is clean.
2. **No test drives #483's bare-`session wait` path past the thousand-row cap.** The repair is the shared `floor()` and blocker 1's test exercises exactly that line, so the fix is proven — but `does not fail a bare wait on an abandon row that is already history (#483)` (`:652`) still runs under the cap, and a `filler: 1_200` twin of it would pin the repaired path directly rather than by argument.
3. **Two comments now describe the old floor.** `in-process-core-session.test.ts:654-656` ("the only thing separating … is the `eventsReplayed` marker") and `:687` ("judged against the `eventsReplayed` floor") read as the marker rather than the tip it now carries. Cosmetic.

For the record, and not a defect: a Core carrying #483's abandon row but *not* reporting `tipEventId` would leave `floor()` null and make `promptAbandoned()` answer `null` for ever. That combination does not exist — `session:promptAbandoned` is nowhere on `main`, so #483 and this land on the same train — but it is the one thing to remember if #483 is ever backported alone.

---

### Verdict

**Approved.** All three blockers are fixed at the mechanism rather than at the symptom: the floor is the log's true head read before the tail, the wait is bounded by three facts about the world with no clock added, and a prompt typed on the quiet gap is reported as `unverified` and exits non-zero instead of claiming a composer nobody saw. Each has coverage that produces the false success — or the hang — on `3561d45` first, which is the standard I asked for. The repair to #483's already-merged report path is present, correct, and explained in the description and the ADR, so the hole is not left shipped on the feature branch.

The draft is left as a draft; marking it ready is yours.

mrfqcentic (approved):
## Re-review of the #277 merge resolution — approved

Judging `eaf0647` only: the merge of `fix/prompt-delivery-and-send-path` @ `802b1fe` (#493 / issue 277, and #492 / issue 396) and the hand resolution on top of it. Everything approved at `97359bf` is left alone.

---

### 1. Did the resolution stay inside the authorised scope?

**Yes.** I did not read this off the diff by eye — I reconstructed the mechanical auto-merge (`git merge-tree --write-tree 97359bf 802b1fe` → tree `f77684a`) and diffed the actual merge commit against it. That isolates every hand edit and nothing else:

```
docs/adr/0026-prompt-delivery-is-a-core-responsibility.md | 296 +++---
packages/cli/src/session-command.ts                       |  16 +-
packages/cli/src/session-gateway.ts                       |  25 +-
packages/core/src/__tests__/harness-prompt-delivery.test.ts |  80 +--
4 files changed
```

**Four files, and they are the four that were authorised** — the two conflicted files, plus the two source files named for the codex sentences. Every other file in the merge is byte-identical to the mechanical result. In particular `packages/core/src/harness-prompt-delivery.ts`, `pty-manager.ts`, `core-entry.ts`, `pty-core-link-server.ts`, the whole of `packages/sdk` and `packages/shared`, and every other test file are **pure auto-merge** — not touched by hand at all. Nothing was improved along the way.

**No behaviour change.** In `session-command.ts` the two changed regions are the `session --help` template block (`:184-190`) and the JSDoc on `promptUnverifiedLine` (`:620-628`). In `session-gateway.ts` they are the JSDoc on `PromptDeliveryReport`'s `unverified` arm (`:275-289`) and the inline comment above the `observed` read in `consider()` (`:983-994`). Not one executable statement, condition, string used as data, or exported shape differs from `97359bf`. `composerObserved` is still computed at `harness-prompt-delivery.ts:1327` exactly as I approved it.

One honest note on the count. The authorisation said *four* sentences; there are **five** edited sites — help text, `promptUnverifiedLine`'s JSDoc, the `unverified` union arm, the `consider()` comment, and the ADR paragraph. The merge commit message lists four and omits the `promptUnverifiedLine` JSDoc. That fifth one said *"on `codex` — which has no readiness row yet"*, which #277 falsified in exactly the same way as the other four; it is in one of the three named files and is the same class of correction, so this is the authorised job done completely rather than scope creep. Leaving it would have shipped a false sentence.

### 2. Are the corrected sentences true against what #277 actually landed?

I read the merged table rather than the description. `packages/core/src/harness-prompt-delivery.ts:696-716`:

```ts
export const HARNESS_READINESS: Partial<Record<Harness, HarnessReadiness>> = {
  opencode:      { composer: [/ask anything/i],                    confirmEcho: true, maxPromptWrites: 3 },
  "cursor-cli":  { composer: [/plan,\s*search,\s*build/i],         confirmEcho: true, maxPromptWrites: 3 },
  "claude-code": { composer: [/\btry\s+"/i],                       confirmEcho: true, maxPromptWrites: 3 },
  codex:         { composer: [/ask\s+codex\s+to\s+do\s+anything/i], confirmEcho: true, maxPromptWrites: 3 },
};
```

and `packages/shared/src/domain.ts:3` — `HARNESSES = ["claude-code", "codex", "cursor-cli", "opencode"]`.

Checked claim by claim:

| Claim in the corrected prose | True? |
|---|---|
| codex has a row: `Ask Codex to do anything`, `confirmEcho`, three writes | **True** — read above, matches |
| "none of the four this build ships" is marker-less | **True** — `HARNESSES` has exactly four members and `HARNESS_READINESS` now keys all four |
| "#277 gave `codex` the **last** readiness row" | **True** — codex was the only member without one |
| "the next harness added arrives that way" | **True** — the table is `Partial<Record<Harness,…>>` and `readinessFor` (`:719-721`) falls back to `NO_READINESS`, so a new `HARNESSES` member compiles with no row and gets `composer: []` |
| "the directory-trust screen that used to swallow its prompt now ends the delivery `abandoned`" | **True** — with a marker present, `onQuiet` (`:1217`) holds instead of typing, `awaitComposer` runs to `composerWaitMs` and abandons; `onDeadline` (`:1400`) also declines to set `deadlinePassed` while the marker is off screen |
| "the quiet gap expiring one millisecond after codex cleared the screen" | **True** — matches #277's own measured amendment (clear at 633 ms, gap at 634 ms), which is now in the same file |

So the substance is true, not merely less wrong: codex is no longer the marker-less example, and the failure it used to illustrate now takes a different and better-reported route.

**One residual overstatement, non-blocking.** The ADR says `unverified` "is unreachable for all four", and the `consider()` comment says it is not "a state any of them reach". There is one ordering where a harness *with* a marker still emits `composerObserved: false`: `deadlinePassed` short-circuits the composer gate at `onQuiet` (`:1217`), and `onOutput` (`:1053-1056`) replaces `this.screen` wholesale on a screen clear — so a clear landing between the 15 s backstop setting `deadlinePassed` (with the marker on screen, `:1400`) and the immediate `schedule()`/settle would reach `writePrompt` with the marker gone. Nothing in #277's twenty-four measured boots does this, and it fails in the safe direction — non-zero and `unverified`, never a false success. "Unreachable in every path anyone has measured" would be exact; "unreachable" is a shade strong for a comment. Not worth another revision on the last ticket of the train.

### 3. Did keeping both blocks produce a contradiction, a duplicate, or a vacuous test?

**No, and I checked it by byte comparison rather than by reading.**

`packages/core/src/__tests__/harness-prompt-delivery.test.ts` is an exact concatenation:

- merged lines 1-1867 vs the base's whole file — **identical**, so all of #277 survives verbatim, including the codex block and the "*Superseded by the issue 277 amendment below*" note #277 added to the 483 amendment;
- merged lines 1868-1907 vs `97359bf`'s issue-395 block — **identical**.

`docs/adr/0026-…md` likewise: merged lines 1-434 are the base's file verbatim, then a blank line, then the issue-395 amendment with **exactly one** region changed — the codex paragraph. Amendment order is 177 → 232 → 483 → 277 → 395, which is chronological and reads correctly.

- **No contradiction.** #277's amendment predates `composerObserved` and never claims codex is marker-less; its "How this composes with the issue 483 amendment" paragraph already says the untrusted boot ends `abandoned`, which is what the rewritten 395 paragraph now says too.
- **No duplicated assertion.** `delivering to codex (issue 277)` tests codex's delivery and abandon behaviour off its captures; `did the Core see a composer… (issue 395)` tests the `composerObserved` field on opencode and on an invented harness name. Disjoint.
- **No vacuous test, and this is the part that could have gone wrong silently.** `some-harness-invented-tomorrow` appears nowhere in the merged `HARNESS_READINESS` (I grepped the merged source: 0 hits), so `readinessFor` still returns `NO_READINESS` and `composerObserved: false` is still a real assertion — the foresight of not writing that test against `codex` is what saved it. And there is **no `toEqual` on a `delivered` event anywhere in the file**; the only two assertions naming `phase: "delivered"` are the 395 block's own, both `toMatchObject` naming `composerObserved` explicitly. So the new required field neither breaks nor is silently absorbed by any of #277's tests.
- **Both blocks demonstrably ran.** `packages/core` on this head: **1011 tests passed, 0 skipped**, 60 files.

### 4. Mechanics

- **A plain `git merge`, no rebase, no force push.** `eaf0647` is a real merge commit with parents `97359bf` `802b1fe`; `git merge-base --is-ancestor 97359bf eaf0647` is true, so the head I approved is intact in history rather than rewritten.
- **Base is still `fix/prompt-delivery-and-send-path`**, now at `802b1fe`, and it is an ancestor of the head. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.
- **`Refs #395`, no closing keyword** — one in the merge commit message, one in the PR body, zero `Closes`/`Fixes #`/`Resolves` anywhere.
- **CI on `eaf0647` is green.** Run `33824257235` concluded `success`: **Unit Tests** (root 24, sdk 22 +2 skipped, shared 31, cli 50 +1 skipped / 1246 tests, core 60 / 1011 tests, panel 169) and **Typecheck** both success, and `Dependency Audit` is green again now that npm's advisories endpoint answers. The second run on the same SHA (`33824250487`) is the superseded duplicate, cancelled in full — including the placeholder job still carrying the unexpanded `${{ matrix.distro }}, ${{ matrix.arch }}` name.

### One thing deliberately not changed, and rightly

`packages/cli/src/__tests__/session-command.test.ts:519/538/549` still uses `harness: "codex"` to illustrate the `unverified` report. The gateway there is injected and returns `unverified` unconditionally, so the test still exercises the real formatting and exit-code path and is not vacuous — it is only a now-unrepresentative label. That file was not in the authorised set, and editing it would have been the scope creep this review was looking for. Leaving it was the right call.

---

### Checklist

| # | Check | Result |
|---|---|---|
| 1 | Only the authorised files differ from the mechanical auto-merge | **PASS** — 4 files: the 2 conflicted ones, `session-command.ts`, `session-gateway.ts` |
| 2 | No behaviour change | **PASS** — help text, two JSDoc blocks and one inline comment; zero executable lines |
| 3 | No other logic touched, nothing else improved | **PASS** — all of `packages/core` source, `packages/sdk`, `packages/shared` and every other test file are pure auto-merge |
| 4 | codex's readiness row read directly | **PASS** — `/ask\s+codex\s+to\s+do\s+anything/i`, `confirmEcho: true`, `maxPromptWrites: 3` |
| 5 | "none of the four" / "the last row" true against `HARNESSES` | **PASS** — 4 members, 4 keys |
| 6 | "the next harness arrives marker-less" true | **PASS** — `Partial<Record<…>>` + `readinessFor` → `NO_READINESS` |
| 7 | The trust-dialog claim true post-#277 | **PASS** — ends `abandoned`, not `delivered` |
| 8 | The 1 ms figure matches #277's own measurement | **PASS** |
| 9 | Every falsified codex sentence corrected, none missed | **PASS** — 5 sites; SKILL.md and both `orchestration-skill-payload.ts` copies never named codex |
| 10 | Residual imprecision | **NOTE** — "`unverified` is unreachable" is a shade strong; one safe-direction race exists. Non-blocking |
| 11 | Both ADR amendments kept whole, in order | **PASS** — base verbatim through line 434, then the 395 amendment with only the codex paragraph rewritten |
| 12 | Both describe blocks kept whole | **PASS** — byte-identical on both sides |
| 13 | No contradiction between the kept blocks | **PASS** |
| 14 | No duplicated assertion | **PASS** |
| 15 | No test now passing vacuously | **PASS** — invented harness still absent from the table; no `toEqual` on a delivered event; core 1011 passed / 0 skipped |
| 16 | Plain `git merge`, no rebase, no force push | **PASS** — two parents, `97359bf` still an ancestor |
| 17 | Base still `fix/prompt-delivery-and-send-path` | **PASS** |
| 18 | `Refs #395`, no `Closes` | **PASS** |
| 19 | `packages/core` and `packages/cli` suites and typecheck green on the head | **PASS** — run `33824257235`, Unit Tests and Typecheck success |

---

### Verdict

**The resolution is sound. Approved, without reservation.**

It stayed inside the authorised scope — I proved that against the mechanical auto-merge, not by reading the diff — and changed no behaviour. The corrected sentences are true against the readiness row #277 actually landed, which I read myself. Keeping both amendment blocks and both describe blocks produced no contradiction, no duplicate assertion and no vacuous test; both blocks are byte-identical to their own sides and both demonstrably run. The merge is a plain merge on the right base, `Refs #395` with no closing keyword, and CI on `eaf0647` is green including both suites and the typecheck.

Nothing here blocks the train.

